PR #31422, opened 2022-07-03
Package/py phylophlan

erinpnewcomer:
This is a pull request for the package py-phylophlan. I mistyped and called it py-phyloseq in one commit.


var/spack/repos/builtin/packages/py-phylophlan/package.py
```diff
@@ -0,0 +1,24 @@
...
+
+from spack.package import *
+
+
+class PyPhylophlan(PythonPackage):
+    """PhyloPhlAn 3.0 is an integrated pipeline for large-scale
+    phylogenetic profiling of genomes and metagenomes."""
+
+    homepage = "https://github.com/biobakery/phylophlan"
+    url      = "https://github.com/biobakery/phylophlan/archive/refs/tags/3.0.2.tar.gz"
```
adamjstewart: Can we instead download from PyPI? See https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#pypi-vs-github. This will also likely change the checksum. There is a `pypi` variable you can use to simplify the URL.

var/spack/repos/builtin/packages/py-phylophlan/package.py
```diff
@@ -0,0 +1,24 @@
...
+
+    version('3.0.2', sha256='c342116662bbfbb49f0665291fc7c0be5a0d04a02a7be2da81de0322eb2256b4')
+
+    depends_on('python@3:')
+    depends_on('py-numpy@1.12.1:')
+    depends_on('py-biopython@1.70:')
+    depends_on('py-dendropy@4.2.0:')
+    depends_on('muscle')
+    depends_on('blast-plus')
+    depends_on('diamond')
```
adamjstewart: Missing deps on matplotlib, pandas, and seaborn.

var/spack/repos/builtin/packages/py-phylophlan/package.py
```diff
@@ -0,0 +1,24 @@
...
+
+    homepage = "https://github.com/biobakery/phylophlan"
+    url      = "https://github.com/biobakery/phylophlan/archive/refs/tags/3.0.2.tar.gz"
+
+    version('3.0.2', sha256='c342116662bbfbb49f0665291fc7c0be5a0d04a02a7be2da81de0322eb2256b4')
+
+    depends_on('python@3:')
+    depends_on('py-numpy@1.12.1:')
+    depends_on('py-biopython@1.70:')
+    depends_on('py-dendropy@4.2.0:')
```
adamjstewart: All Python deps should be `type=('build', 'run')`, possibly the ones below too. Depends on whether or not the package links to them or just needs them to be in `PATH` during runtime.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-phylophlan/package.py
erinpnewcomer: Hi! I have tested out changing this to Pypi instead of Github, but there is an error with the setup.py file in the Pypi package and not the Github tar file. Line 39 of setup.py calls for the long description that is stored in "readme.md", which exists in the Github download but is called "PKG-INFO" in the Pypi download. What would you recommend I do to fix this, or should I leave it as the Github version?

adamjstewart: In that case you can leave your build recipe as is. If you want you can open an issue with the phylophlan developers to let them know that their sdists are broken.

adamjstewart (dismissed):

var/spack/repos/builtin/packages/py-phylophlan/package.py
```diff
@@ -0,0 +1,27 @@
...
+    depends_on('python@3:', type=('build', 'run'))
+    depends_on('py-numpy@1.12.1:', type=('build', 'run'))
+    depends_on('py-biopython@1.70:', type=('build', 'run'))
+    depends_on('py-dendropy@4.2.0:', type=('build', 'run'))
+    depends_on('muscle', type=('build', 'run'))
+    depends_on('blast-plus', type=('build', 'run'))
+    depends_on('diamond', type=('build', 'run'))
+    depends_on('py-matplotlib', type=('build','run'))
+    depends_on('py-pandas', type=('build','run'))
+    depends_on('py-seaborn', type=('build','run'))
```
adamjstewart:
```suggestion
    depends_on('py-matplotlib', type=('build', 'run'))
    depends_on('py-pandas', type=('build', 'run'))
    depends_on('py-seaborn', type=('build', 'run'))
```
This should fix the failing style tests

erinpnewcomer: Thanks! Fixed in the newest commit.

adamjstewart (approved):

Issue #987, opened 2016-05-25
Allow packages to extend multiple other packages


adamjstewart:
I'm working on a package for Cantera, which optionally builds a Python module and Matlab toolbox. I don't have a Matlab package just yet, but it appears that Spack doesn't allow packages to extend multiple other packages. Is there a reason for this?

tgamblin:
I'm not sure what that would mean.  Each package's install method is written to assume a spec *for that particular package*, which includes things like the package version.  There could be conditional logic in the `install` method for another package that relies on that, but what does the spec even mean when you've extended multiple other packages?

adamjstewart:
By "extend" I mean allow `spack activate` to work its magic. If a package builds a Python module and a Matlab toolbox, it needs to symlink both of those, right?

tgamblin:
Oh, duh.  Yes.  I was confused by the "optionally builds a Python module and a Matlab toolbox" part -- I thought you wanted those built all by the same package.

Packages *should* be able to extend multiple packages, but at the moment the functions on `Package` assume a single extendee.  There's no reason they have to.  The main reason for that is that we'd want a little syntax help on the command line so that you could say which extendee to activate.

Suppose `foo` extends `python` and `matlab`, but you only want to activate it for Python -- would the following syntax work for you?

```bash
    # activate foo for all extendees (both python and matlab)  (?)
    spack activate foo

    # activate just for python
    spack activate foo ^python

    # activate just for matlab
    spack activate foo ^matlab

    # activate for python and matlab
    spack activate foo ^python ^matlab
```

We may also want to add better control for what parts of the package get symlinked into the extendee.  Right now the package can provide a list of stuff to ignore as a kwarg to `extends()`; might want more.

Thoughts?

adamjstewart:
>  I was confused by the "optionally builds a Python module and a Matlab toolbox" part -- I thought you wanted those built all by the same package.

Well, the Cantera package will build the C++/Fortran stuff by default. And if a user specifies `+python +matlab`, it will also build a Cantera Python module and Matlab toolbox. This requires a dependency so that I know the installation path, and it requires activation later.

> Suppose foo extends python and matlab, but you only want to activate it for Python

I don't think this is really a concern. If a user builds `cantera +python +matlab`, they probably want to activate both. If they just want Python, they can build `cantera +python -matlab`. Right now it's setup like this:
```python
extends('python', when='+python')
extends('matlab', when='+matlab')
```

[...]

PR #31430, opened 2022-07-04
jellyfish: added variants for different bindings

teachers-uk-net:
jellyfish can make bindings available to perl, ruby and python.

This PR makes them available as variants in the jellyfish package.

### DIFF
--- a/var/spack/repos/builtin/packages/jellyfish/package.py
+++ b/var/spack/repos/builtin/packages/jellyfish/package.py
@@ -22,9 +22,9 @@ class Jellyfish(AutotoolsPackage):
     variant('rubybind', default=False, description='Enable ruby binding')
     variant('perlbind', default=False, description='Enable perl binding')
 
-    extends('perl', when='bindings=perlbind')
-    extends('python', when='bindings=pybind')
-    extends('ruby', when='bindings=rubybind')
+    extends('perl', when='+perlbind')
+    extends('python', when='+pybind')
+    extends('ruby', when='+rubybind')
 
     patch('dna_codes.patch', when='@1.1.11')
 
--- a/var/spack/repos/builtin/packages/jellyfish/package.py
+++ b/var/spack/repos/builtin/packages/jellyfish/package.py
@@ -28,6 +28,7 @@ class Jellyfish(AutotoolsPackage):
 
     patch('dna_codes.patch', when='@1.1.11')
 
+    # v1.1.11 does not support language bindings
     conflicts('+pybind', when='@1.1.11')
     conflicts('+rubybind', when='@1.1.11')
     conflicts('+perlbind', when='@1.1.11')
--- a/var/spack/repos/builtin/packages/jellyfish/package.py
+++ b/var/spack/repos/builtin/packages/jellyfish/package.py
@@ -18,27 +18,32 @@ class Jellyfish(AutotoolsPackage):
     version('1.1.11', sha256='496645d96b08ba35db1f856d857a159798c73cbc1eccb852ef1b253d1678c8e2',
             url='https://www.cbcb.umd.edu/software/jellyfish/jellyfish-1.1.11.tar.gz')
 
-    variant('pybind', default=False, description='Enable python binding')
-    variant('rubybind', default=False, description='Enable ruby binding')
-    variant('perlbind', default=False, description='Enable perl binding')
+    variant(
+        'bindings', default='none', description='Support for binding to Ruby, Python and Perl',
+        values=('pybind', 'rubybind', 'perlbind', 'none'), multi=True
+    )
 
-    depends_on('perl', type=('build', 'run'))
-    depends_on('python', type=('build', 'run'))
+    extends('perl', when='bindings=perlbind')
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/jellyfish/package.py
```diff
@@ -18,7 +18,27 @@ class Jellyfish(AutotoolsPackage):
...
+                '--enable-python-binding',
+            ]
+        elif '+rubybind' in spec:
+            args = [
+                '--enable-ruby-binding',
+            ]
+        elif '+perlbind' in spec:
+            args = [
+                '--enable-perl-binding'
+            ]
```
alalazo, line 53: Are the bindings exclusive, i.e. can't we activate them for each language?

var/spack/repos/builtin/packages/jellyfish/package.py
```diff
@@ -18,7 +18,27 @@ class Jellyfish(AutotoolsPackage):
     version('1.1.11', sha256='496645d96b08ba35db1f856d857a159798c73cbc1eccb852ef1b253d1678c8e2',
             url='https://www.cbcb.umd.edu/software/jellyfish/jellyfish-1.1.11.tar.gz')
 
+    variant('pybind', default=False, description='Enable python binding')
+    variant('rubybind', default=False, description='Enable ruby binding')
+    variant('perlbind', default=False, description='Enable perl binding')
```
alalazo, line 23: I think we might be able to turn this into a multi-valued variant e.g. `languages=python`. Also, I think we want to have:
```python
extends('python', when='languages=python')
```
ans similar for other bindings.

alalazo (changes requested):
I left a few comments, let me know what you think.

var/spack/repos/builtin/packages/jellyfish/package.py
teachers-uk-net: Ah, yes, I believe you are probably correct, I will set some time aside to have a look.

var/spack/repos/builtin/packages/jellyfish/package.py
teachers-uk-net: Hi

Yes, easy enough to turn into a multi-valued variant (used bindings rather than languages) 
```
variant(
        'bindings', default='none', description='Support for binding to Ruby, Python and Perl',
        values=('pybind', 'rubybind', 'perlbind', 'none'), multi=True
    )
```
although when I tested using extends instead of depends_on
```
extends('python', when='bindings=pybind')
```
received an AssertionError (quite possible I do not have something correct although the spec concretized)
```
==> Error: Class constructor failed for package 'builtin.jellyfish'.

Caused by:
AssertionError: 
  File "/home/stuart/code/teachers_github/spack/lib/spack/spack/repo.py", line 1125, in get
    return package_class(spec)
  File "/home/stuart/code/teachers_github/spack/lib/spack/spack/package_base.py", line 808, in __init__
    if self.is_extension:
  File "/home/stuart/code/teachers_github/spack/lib/spack/spack/package_base.py", line 1394, in is_extension
    return self.extendee_spec is not None
  File "/home/stuart/code/teachers_github/spack/lib/spack/spack/package_base.py", line 1363, in extendee_spec
    assert len(deps) == 1
```
Any suggestions welcome!

Kind regards

alalazo: Right. This is longstanding limitation of Spack, see https://github.com/spack/spack/issues/987, which I forgot about when writing the note above. 

Wondering if we can use a single-valued variant as a workaround for this. I assume that in your test above you specified more than one value for the `bindings` variant, right?

teachers-uk-net: Yes, I did a test with more than one value for `bindings`, I will do some testing with single-valued variants.

teachers-uk-net: Hi

So, subjected most aspects to some more testing and have reverted to the single-value variants with extends and only allowed one binding to be specified at a time.

Discovered that v1.1.11 does not support any bindings (would fail silently on build but I have added conflicts to make it more explicit) - started this because we had a package that depends on jellyfish@2.2.7 that requires a python binding and then another user wanted Kraken, which depends on jellyfish@1.1.11 (and I had lazily overridden the original jellyfish).

v2.2.7 only successfully built with python@2.7.x (our use case with the package that depends on jellyfish@2.2.7 already depended on python@:2.7.18 so it was already in the DAG most of the time I was using it).

Hope this helps!

Kind regards

bernhardkaindl:
@teachers-uk-net It appears that you denied maintainers write permission to this PR's branch.

I pushed my changes to a new branch and opened a new PR (sorry), see #33832 

- The making the perl bindings a variant would be a 2nd step.
- The perl and python bindings were likely built by default already because they were added to the DAG in simple `depends_on()` calls.

alalazo:
Closing since #33832 was merged, and this has been stale for a while. Feel free to submit follow up PRs to pursue the items in https://github.com/spack/spack/pull/31430#issuecomment-1311248950

PR #31388, opened 2022-06-30
[mpas-model] Add optional `make_target` and `precision` variants

stephenmsachs:
Flagging maintainer @t-brown.

`make_target` can be used to instruct Spack to build one of the pre-defined make
targets in the MPAS makefile. Spack will guess the targret based on the value of
`spack_fc`, but the user can overwrite the target as variant. E.g.
```
spack install mpas-model make_target=pgi-nersc
```

`precision` is used to set the `PRECISION` flag in the Makefile to {single,
double}. Default is double.

PR #31431, opened 2022-07-04
[mpas-model] Overwrite compiler args in Makefile

stephenmsachs:
Tagging maintainer @t-brown 

The pre-defined compilation targets do not follow the usual behavior of
Makefiles. Compiler flags are set as strings (not Makefile variables) and as
such are not able to be overridden by environment variables. This patch changes
the behavior to the expected behavior of a Makefile such that `fflags` etc have
the desired effect.

This is a necessary fix for https://github.com/spack/spack/pull/31388.

And another quick fix: Use 8 byte reals only when `precision=double` is set.

### DIFF
--- a/var/spack/repos/builtin/packages/mpas-model/package.py
+++ b/var/spack/repos/builtin/packages/mpas-model/package.py
@@ -59,10 +59,12 @@ class MpasModel(MakefilePackage):
             ])
         elif satisfies('%intel'):
             fflags.extend([
-                '-r8',
                 '-convert big_endian',
                 '-FR',
             ])
+            if satisfies('precision=double'):
+                fflags.extend(['-r8'])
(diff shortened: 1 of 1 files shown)


alalazo (approved):
Can we contribute a patch upstream too, so that we can restrict later the versions that needs patching? In any case, LGTM - just waiting sometime for @t-brown to chime in.

stephenmsachs:
> Can we contribute a patch upstream too

Well, I can try. But I believe it's unlikely they will change the structure of their Makefile. That structure is older than the project on github..

t-brown:
There was talk upstream about moving to CMake, however I don't know the status of that.
This update LGTM.

t-brown (approved):

PR #31433, opened 2022-07-05
Remove fetch from depfile

haampie:
The idea of the fetch target is to allow parallel download of sources, which is
great, except that `spack fetch` and `spack install` have races or don't handle
locks correctly as shown here: https://github.com/spack/spack/issues/31387

Also, there's a downside to always fetch sources, since it does not take into
account whether or not the package is already installed, so it may download a
copy of CUDA/MKL even if already installed.

Therefore, it seems easiest for now to just drop the fetch target, maybe later
we can re-evaluate if it's useful to have under a flag.

The only way in which this is breaking is we've advertised the `fetch-all` target
in the docs. But I doubt people actually rely on it, and since `env depfile` is brand new and
not without issues, I think it's OK to make a breaking change without deprecation?

### DIFF
--- a/lib/spack/docs/environments.rst
+++ b/lib/spack/docs/environments.rst
@@ -1013,7 +1013,7 @@ The following advanced example shows how generated targets can be used in a
 
    SPACK ?= spack
 
-   .PHONY: all clean fetch env
+   .PHONY: all clean env
 
    all: env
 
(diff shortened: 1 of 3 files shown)


alalazo (approved):

alalazo:
Should this one go into v0.18.1 ?

haampie:
I would say not because it's actually breaking a (minor) documented feature, namely the generated `make fetch-all` target

PR #31437, opened 2022-07-05
armpl: Add new versions and rename package to armpl-gcc

annop-w:
Add versions 22.0.2.
Rename package name from armpl to armpl-gcc.


tldahlgren (approved):
Confirmed a subset of the sha256.

PR #31438, opened 2022-07-05
adiak: add version 0.2.2

mplegendre:


tldahlgren (approved):
Confirmed the commit exists and is associated with a tag corresponding to that version.

PR #31440, opened 2022-07-05
llvm-openmp: add v14.0.6

adamjstewart:
Confirmed that both 12.0.1 and 14.0.6 build on macos 12.4 (Apple M1 Pro) with Apple Clang 13.1.6.


tldahlgren (approved):
Confirmed the sha256 (and the URL change did not affect the sha256 for older versions).

PR #31441, opened 2022-07-05
py-libclang: add v13.0.0

adamjstewart:
Successfully builds on macOS 12.4 (Apple M1 Pro) with Apple Clang 13.1.6.


tldahlgren (approved):
Confirmed the sha256.

PR #31445, opened 2022-07-05
py-tensorboard: install from wheel

adamjstewart:
### Before

Compile from source, requires Bazel, can take over an hour to build, does not yet support Apple M1: https://github.com/tensorflow/tensorboard/issues/5714

### After

No non-Python dependencies, builds in 1 second.

TensorBoard is a pure Python package, there's no reason not to use the wheel. See https://github.com/spack/spack/discussions/30795 for discussion.

### DIFF
--- a/var/spack/repos/builtin/packages/py-tensorboard/package.py
+++ b/var/spack/repos/builtin/packages/py-tensorboard/package.py
@@ -3,38 +3,33 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-import tempfile
-
 from spack.package import *
 
 
-class PyTensorboard(Package):
+class PyTensorboard(PythonPackage):
     """TensorBoard is a suite of web applications for
     inspecting and understanding your TensorFlow runs and
     graphs."""
 
     homepage = "https://github.com/tensorflow/tensorboard"
-    url      = "https://github.com/tensorflow/tensorboard/archive/2.2.0.tar.gz"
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-tensorboard/package.py
```diff
@@ -3,38 +3,33 @@
...
-class PyTensorboard(Package):
+class PyTensorboard(PythonPackage):
     """TensorBoard is a suite of web applications for
     inspecting and understanding your TensorFlow runs and
     graphs."""
 
     homepage = "https://github.com/tensorflow/tensorboard"
-    url      = "https://github.com/tensorflow/tensorboard/archive/2.2.0.tar.gz"
+    url = "https://files.pythonhosted.org/packages/py3/t/tensorboard/tensorboard-2.9.1-py3-none-any.whl"
+    list_url = "https://pypi.org/simple/tensorboard/"
```
glennpj, line 16: Does using wheels eliminate the `pypi` attribute?

glennpj (approved):
This looks fine to me.

var/spack/repos/builtin/packages/py-tensorboard/package.py
adamjstewart: In the current implementation, yes. It turns out that it's actually quite difficult to programmatically parse the full wheel filename format. For `.tar.gz` and `.zip`, the string after "packages" is always "source". But for wheels, it could be "py3", "py2.py3", "cp37", etc. I need to make `pypi` smarter to handle these properly.

PR #28527, opened 2022-01-20
PythonPackage: add default libs/headers attributes

adamjstewart:
This likely needs a bit more thought and testing, but this is a good draft of what I think we should do for Python libraries.

Note that we only search `platlib`, not `purelib`. For most packages (at least ones that use setuptools), all files are installed to `platlib` if the package is not pure-Python. If things are installed in purelib instead of platlib, there likely aren't any libraries/headers anyway.

Closes #14513 
Closes #22475

PR #31446, opened 2022-07-05
py-torch: re-add headers/libs properties

adamjstewart:
Installation of torchvision was broken by #28527. I have no idea why. This PR fixes things while we figure out why #28527 doesn't work. It adds the exact same headers/libs properties as are in the base class. Hoping someone can figure out why this is needed.


adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

alalazo (approved):

PR #30582, opened 2022-05-10
ROCmPackage: unify amdgpu_targets

albestro:
About ROCm packages, it is a good general practice to inherit from `ROCmPackage`, which provides `rocm` and `amdgpu_targets` variant. For the latter one it also provides a defined set of valid AMD GPU targets values.

In some packages like `rocblas` and `rocsolver`, the `rocm` variant does not make much sense, so they manually specify the other variant together with a set of valid AMD GPU targets.

This triggers a maintenance problem. Indeed, for these packages the set of valid architectures has to be maintained manually, which is costly and error-prone.

This PR is a proposal to do it in a slightly different way, so that the set of valid targets is defined in a single place. I started with just `rocblas` and `rocsolver`, but if you agree on this style, I can also adapt others

PR #29018, opened 2022-02-16
Add additional AMD GPU targets from LLVM documentation

chuckatkins:
Add additional AMD GPU targets from the AMD-forked LLVM documentation.  This also expands the feature-modified target values and moves the knowledge of them to the base ROCmPackage.

PR #31450, opened 2022-07-05
Ci rocm stack

kwryankrattiger:
Add a pipeline that tests building the ROCm stack. There have been increasingly frequent issues keeping this stack stable as more packages depend on it, so it seems now is a good time to add this pipeline.

@srekolam @chuckatkins

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/rocm/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/rocm/spack.yaml
@@ -1,0 +1,172 @@
+spack:
+  view: false
+
+  concretizer:
+    reuse: false
+    unify: true
+
+  config:
+    install_tree:
+      root: /home/software/spack
+      padded_length: 512
+      projections:
+        all: '{architecture}/{compiler.name}-{compiler.version}/{name}-{version}-{hash}'
+
+  definitions:
+    - default_specs:
+      - hip
+      - hipblas
+      - hipcub
+      - hipfft
+      - hipfort
+      - hipify-clang
+      # - hip-rocclr # Needs fix for install
+      - hipsolver
+      - hipsparse
+      # - hipsycl # Needs fixes for to use llvm-amdgpu clang
+      - hsa-rocr-dev
+      - mesa +llvm
+      - miopen-hip
+      - rocalution
+      - rocblas amdgpu_target=gfx1030
+      - rocfft amdgpu_target=gfx1030
+      - rocm-clang-ocl
+      - rocm-cmake
+      - rocm-dbgapi
+      - rocm-debug-agent
+      # - rocm-device-libs # built by llvm-amdgpu
+      - rocm-gdb
+      - rocminfo
+      - rocm-opencl
+      - rocm-openmp-extras
+      # - rocm-smi # Deprecated after 4.1.0 in favor of rocm-smi-lib
+      - rocm-smi-lib
+      # - rocm-tensile # Build issues
+      # - rocm-validation-suite # Needs hip-rocclr
+      - rocprim amdgpu_target=gfx1030
+      - rocprofiler-dev
+      - rocrand amdgpu_target=gfx1030
+      - rocsolver ~optimal amdgpu_target=gfx1030 # Disabling optimal to improve build time
+      - rocsparse amdgpu_target=gfx1030
+      - rocthrust
+      - roctracer-dev
+      - roctracer-dev-api
+      - llvm-amdgpu +rocm-device-libs+llvm_dylib+link_llvm_dylib
+    - arch:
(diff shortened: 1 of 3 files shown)


srekolam (dismissed):

share/spack/gitlab/cloud_pipelines/stacks/rocm/spack.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - rocm-opencl
+      - rocm-openmp-extras
+      # - rocm-smi # Deprecated after 4.1.0 in favor of rocm-smi-lib
+      - rocm-smi-lib
+      # - rocm-tensile #
+      # - rocm-validation-suite # Needs hip-rocclr
+      - rocprim
+      - rocprofiler-dev
+      - rocrand amdgpu_target=gfx1030
+      - rocsolver
```
cgmb: Unless you're checking the performance of these binaries, I would recommend `rocsolver ~optimal`. It omits all the loop-unrolled kernels that are templated for different problem sizes, so it builds much faster without changing what the library needs from its dependencies or what it provides to its dependents.

cgmb (commented):
You can get a pretty big reduction in your build time if you specify an `amdgpu_target` variant for the mathlibs, particularly rocblas, rocfft, rocsparse and rocsolver. The build time cost of each architecture is basically linear and the default `amdgpu_target=auto` builds for ~7 architectures. Though, perhaps setting it for rocrand has forced the rest to resolve to `amdgpu_target=gfx1030` as well?

kwryankrattiger:
@cgmb I wasn't sure how to set the `amdgpu_target` since it is inconsistent across packages, but it looks like #30582 fixes some of that. There was another PR by @chuckatkins that does something similar to split out the features as well #29018.

I think now that the #30582 is merged I can put that in more easily.

share/spack/gitlab/cloud_pipelines/stacks/rocm/spack.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - mesa +llvm
+      - miopen-hip
+      - rocalution
+      - rocblas amdgpu_target=gfx1030
+      - rocfft amdgpu_target=gfx1030
+      - rocm-clang-ocl
+      - rocm-cmake
+      - rocm-dbgapi
+      - rocm-debug-agent
+      # - rocm-device-libs # built by llvm-amdgpu
```
srekolam, line 37: include rocm-device-libs too .. rocm-openmp-extras is dependent on rocmdevicelibs .it does not use the devicelibs built thru the llvm-amdgpu.

share/spack/gitlab/cloud_pipelines/stacks/rocm/spack.yaml
```diff
@@ -0,0 +1,155 @@
...
+    signing-job-attributes:
+      image: { "name": "ghcr.io/spack/notary:latest", "entrypoint": [""] }
+      tags: ["spack", "aws"]
+      script:
+        - aws s3 sync --exclude "*" --include "*spec.json*" ${SPACK_REMOTE_MIRROR_OVERRIDE}/build_cache /tmp
+        - /sign.sh
+        - aws s3 sync --exclude "*" --include "*spec.json.sig*" /tmp ${SPACK_REMOTE_MIRROR_OVERRIDE}/build_cache
+
+  cdash:
+    build-group: Build tests for different build systems
```
scottwittenburg: @kwryankrattiger Little thing: can you change the build group name to be stack-specific?  This is the group we used for `build_systems` I think.  Maybe something with `rocm` in the name.

kwryankrattiger: Yeah, I can do that! I wasn't sure how that was being used so I didn't touch it :p

kwryankrattiger:
@scottwittenburg or @srekolam if you guys could give this an approval to get it merged, I think it is ready now!

srekolam (dismissed):

kwryankrattiger:
@srekolam I just rebased, hopefully that fixes the coverage CI.

cgmb:
Is this change waiting on something? A CI run for ROCm would be very useful.

kwryankrattiger:
Just waiting on a clean CI. My current plan is to disable the failing packages, but I _think_  there may just need to point at the correct `amdgpu_targets`.

cgmb:
The failures are because the build is passing `-DAMDGPU_TARGETS:STRING=gfx1030;gfx701` but `gfx701` is not a supported architecture. Where is that value coming from? The target list only specifies `amdgpu_target=gfx1030`.

kwryankrattiger:
Closing this for now.

PR #31452, opened 2022-07-06
py-panaroo, py-edlib, and py-intbitset: new packages

luke-dt:


tldahlgren:
There are some dependencies missing.  According to the unit tests, the following should be dependencies:  `py-edlib`, `py-intbitset`, and `py-sklearn`.

var/spack/repos/builtin/packages/py-sklearn/package.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+from spack.package import *
+
+
+class PySklearn(PythonPackage):
```
adamjstewart: `sklearn` has been renamed to `scikit-learn`: https://pypi.org/project/sklearn/

You should use `scikit-learn` instead.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-sklearn/package.py
luke-dt: Yes, the developer of Panaroo probably should do that. Unfortunately for me, the depreciated sklearn is listed as a requirement in the panaroo setup.py.

It looks like it's only used for supplemental post-processing visualization (making a MDS plot):
https://github.com/gtonkinhill/panaroo/search?q=sklearn

adamjstewart: I would still use `depends_on('py-scikit-learn')`. Even if the name is different, the imports are the same, so it should work properly with the new package dependency.

luke-dt: I didn't know if would just work, that's great! Re-built and confirmed that it functions on my local machine.

var/spack/repos/builtin/packages/py-intbitset/package.py
```diff
@@ -0,0 +1,24 @@
...
+    unsigned integers either 32-bits integers or an infinite range with fast
+    set operations implemented via bit vectors in a Python C extension for
+    speed and reduced memory usage."""
+
+    homepage = "https://github.com/inveniosoftware/intbitset"
+    pypi     = "intbitset/intbitset-3.0.1.tar.gz"
+
+    version('3.0.1', sha256='f1e6d03c6729922a223c51849df65b9e916e625aefb911784e7f9acd4c207d53')
+
+    depends_on('python@3.Z:',     type=('build', 'run'))
```
adamjstewart: This line can be removed, they seem to support Python 2 and the Z doesn't mean anything here

var/spack/repos/builtin/packages/py-intbitset/package.py
```diff
@@ -0,0 +1,24 @@
...
+    depends_on('py-setuptools',   type='build')
+    depends_on('py-pytest',       type=('build'))
+    depends_on('py-pytest-xdist', type=('build'))
```
adamjstewart: I don't see anything to suggest that pytest is required, these lines can be removed

adamjstewart (changes requested):

adamjstewart (approved):

PR #31459, opened 2022-07-06
lammps package: added fftw_precision variant

marcodelapierre:
Adds `fftw_precision` variant, to enable choosing between `single` and `double` precision, as per https://docs.lammps.org/Build_settings.html#fft-library .

This is an important build option when using the `kspace` package. In fact, the resulting trade-off between performance and accuracy can become a significant factor for many simulation setups.

### DIFF
--- a/var/spack/repos/builtin/packages/lammps/package.py
+++ b/var/spack/repos/builtin/packages/lammps/package.py
@@ -144,7 +144,7 @@ class Lammps(CMakePackage, CudaPackage):
         '64-bit #atoms #timesteps, bigbig: also 64-bit imageint, 64-bit atom ids)',
         values=('smallbig', 'bigbig', 'smallsmall'), multi=False
     )
-    variant('fftw_precision', default='double',
+    variant('fftw_precision', default='double', when='+kspace',
             description='Select FFTW precision (used by Kspace)',
             values=('single', 'double'), multi=False)
 
@@ -583,7 +583,7 @@ class Lammps(CMakePackage, CudaPackage):
             # Using the -DFFT_SINGLE setting trades off a little accuracy
             # for reduced memory use and parallel communication costs
             # for transposing 3d FFT data.
-            if spec.variants['fftw_precision'].value == 'single':
+            if spec.satisfies('fftw_precision=single'):
                 args.append('-DFFT_SINGLE=True')
             else:
                 args.append('-DFFT_SINGLE=False')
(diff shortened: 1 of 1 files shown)


marcodelapierre:
@alalazo one more commit for the Lammps recipe, with a new useful variant (for an old feature)

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -144,6 +144,9 @@ def url_for_version(self, version):
         '64-bit #atoms #timesteps, bigbig: also 64-bit imageint, 64-bit atom ids)',
         values=('smallbig', 'bigbig', 'smallsmall'), multi=False
     )
+    variant('fftw_precision', default='double',
```
alalazo:
```suggestion
    variant('fftw_precision', default='double', when='+kspace',
```

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -567,20 +570,23 @@ def cmake_args(self):
...
-                # Using the -DFFT_SINGLE setting trades off a little accuracy
-                # for reduced memory use and parallel communication costs
-                # for transposing 3d FFT data.
-                args.append(self.define('FFT_SINGLE', True))
             if '^cray-fftw' in spec:
                 args.append('-DFFT=FFTW3')
+            # Using the -DFFT_SINGLE setting trades off a little accuracy
+            # for reduced memory use and parallel communication costs
+            # for transposing 3d FFT data.
+            if spec.variants['fftw_precision'].value == 'single':
```
alalazo:
```suggestion
            if spec.satisfies('fftw_precision=single'):
```

alalazo (changes requested):
If I understand the package logic correctly, the new variant probably needs to be conditional

var/spack/repos/builtin/packages/lammps/package.py
marcodelapierre: Addressed and tested in new commit

var/spack/repos/builtin/packages/lammps/package.py
marcodelapierre: Addressed and tested in new commit

marcodelapierre:
Thanks for the review @alalazo , I have implemented (and tested) your comments in the new commit.

alalazo (approved):

PR #31460, opened 2022-07-06
Added a variant for ZeroMQ (zmq) tests

imjelde:


var/spack/repos/builtin/packages/libzmq/package.py
```diff
@@ -91,4 +94,6 @@ def configure_args(self):
         if 'clang' in self.compiler.cc:
             config_args.append("CFLAGS=-Wno-gnu")
             config_args.append("CXXFLAGS=-Wno-gnu")
+        if '+zmq' in self.spec:
+            config_args.append('ZMQ_BUILD_TESTS')
```
alalazo, line 98: Are the tests getting installed? Or are they just being executed at build time?

alalazo (changes requested):

bernhardkaindl:
@imjelde: `ZMQ_BUILD_TESTS` is a flag in the CMakeLists.txt of newer versions, but spack uses the `configure` instead.

You could work towards migrating libzmq to derive from `CMakePackage` but it would be besides the point and the `Makefile` build by `configure` has this:

The `Makefile` generated by `configure` provides `make check` which spack would try to run on `spack install --test=root`, but unfortunately that fails to build the tests due to `error: "crypto_secretbox_ZEROBYTES" redefined [-Werror]`.

If you want to enable tests, work towards make `make check` pass in in the spack environment (source `spack-build-env-mods.txt` to get the build env of spack interactively) and get the needed fixes into the next upstream version, to be used by spack then.

PR #31465, opened 2022-07-06
[py-distributed] Add version 2022.2.1

stephenmsachs:


var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -36,9 +37,11 @@ class PyDistributed(PythonPackage):
     depends_on('py-cloudpickle@0.2.2:', type=('build', 'run'), when='@:2.16.0')
```
adamjstewart, line 39: Python 3.8+ is required now

adamjstewart (changes requested):

adamjstewart (approved):

PR #31466, opened 2022-07-06
OpenMX: @3.9 and %gcc@10: support

marcosmazz:
Tested on AMD Rome with `openmx@3.9%gcc@11.2.0 ^amdblis ^amdfftw ^amdlibflame ^amdscalapack ^openmpi@4.1.4`


marcosmazz:
Apparently `scalapack`  yields better results in `3.8` and is mandatory in `3.9`.
http://www.openmx-square.org/openmx_man3.8/node9.html
I will make some test to check if dependency list can be reduced.

var/spack/repos/builtin/packages/openmx/package.py
```diff
@@ -41,6 +49,10 @@ def edit(self, spec, prefix):
         # Move contents to source/
         # http://www.openmx-square.org/bugfixed/18June12/README.txt
         copy_tree('patch', 'source')
+        # Move extra file for 3.9 patch
+        # http://www.openmx-square.org/bugfixed/21Oct17/README.txt
+        if spec.satisfies('@3.9'):
```
alalazo, line 52: Are these check really just for `@3.9` or for anything "greater or equal to" `@3.9:`?

var/spack/repos/builtin/packages/openmx/package.py
```diff
@@ -83,10 +105,18 @@ def common_arguments(self):
...
+                'openmx', 'DosMain', 'jx', 'analysis_example', 'esp', 'polB',
+                'calB', 'Z2FH', 'bandgnu13', 'bin2txt', 'cube2xsf', 'intensity_map',
+                'md2axsf', 'tp', 'kSpin', 'BandDispersion', 'ADenBand', 'FermiLoop',
+                'GridCalc', 'MulPOnly', 'MulPCalc', 'example_mpi_spawn', 'gcube2oned'
+            ] + self.common_arguments
+        else:
+            return [
+                'openmx', 'DosMain', 'jx', 'analysis_example', 'esp', 'polB',
+                'bandgnu13', 'bin2txt', 'cube2xsf', 'intensity_map', 'md2axsf'
+            ] + self.common_arguments
```
alalazo: This is a subset of the case above, so we can start with assigning it and adding more targets if the condition is met.

alalazo (changes requested):

var/spack/repos/builtin/packages/openmx/package.py
marcosmazz: Really don't know why `all` target is not used in the first place

alalazo: If that can be used, let's just:
```
build_targets = ['all']
```

marcosmazz (commented):
Adding scalapack removes some cluttered version checking logic

var/spack/repos/builtin/packages/openmx/package.py
marcosmazz: Needs to append `common_arguments()` call to overwrite compilers and libs vars. Testing now with 
```
    @property
    def build_targets(self):
        return ['all'] + self.common_arguments

```

var/spack/repos/builtin/packages/openmx/package.py
marcosmazz: Is the way current patch is applied to 3.9 version.
Can't tell for future patches, see release method:
http://www.openmx-square.org/download.html

alalazo: Thanks for the pointer. It's quite confusing to see 3.9.9 in the patch and 3.9 in the release number, but that's somehow a choice of the upstream project.

I see though that there are at least two patches for version 3.9. We might consider reworking this package in the future, if it seems sensible, to use 3.8.5 and 3.9.9 instead of 3.8 and 3.9. This would allow us to add new "patch" versions without removing the ones previously present.

alalazo (approved):

PR #31467, opened 2022-07-06
[py-ipyparallel] setuptools.config.StaticModule moved

stephenmsachs:
... in py-setuptools@61

### DIFF
--- a/var/spack/repos/builtin/packages/py-ipyparallel/package.py
+++ b/var/spack/repos/builtin/packages/py-ipyparallel/package.py
@@ -25,6 +25,7 @@ class PyIpyparallel(PythonPackage):
     depends_on('py-jupyterlab@3.0:3', type='build', when='@7.1:')
     depends_on('py-packaging', type='build', when='@7.1:')
     depends_on('py-setuptools@40.8:', type='build', when='@7.1:')
+    depends_on('py-setuptools@:60', type='build', when='@:8.0.0')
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-ipyparallel/package.py
```diff
@@ -25,6 +25,7 @@ class PyIpyparallel(PythonPackage):
     depends_on('py-jupyterlab@3.0:3', type='build', when='@7.1:')
     depends_on('py-packaging', type='build', when='@7.1:')
     depends_on('py-setuptools@40.8:', type='build', when='@7.1:')
+    depends_on('py-setuptools@:60', type='build', when='@:8.0.0')
```
adamjstewart: Has this been fixed in 8.0.1?

stephenmsachs: Actually in 8.2.1. Let me change that.

adamjstewart: If it was fixed in 8.2.1 then the range should be `@:8.2.0`, version ranges are inclusive in Spack.

stephenmsachs: Thanks!..... this is embarassing 😄

adamjstewart (approved):

stephenmsachs:
@spackbot run pipeline

PR #31469, opened 2022-07-06
[py-gevent] Fix: deprecated compiler flags

stephenmsachs:
Upstream PR: https://github.com/gevent/gevent/pull/1896

### DIFF
--- a/var/spack/repos/builtin/packages/py-gevent/package.py
+++ b/var/spack/repos/builtin/packages/py-gevent/package.py
@@ -36,4 +36,5 @@ class PyGevent(PythonPackage):
     depends_on('py-zope-event', when='@20.5.1:', type=('build', 'run'))
     depends_on('py-zope-interface', when='@20.5.1:', type=('build', 'run'))
 
+    # Deprecated compiler options. upstream PR: https://github.com/gevent/gevent/pull/1896
     patch('icc.patch', when='%intel')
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/py-gevent/package.py
```diff
@@ -35,3 +35,7 @@ class PyGevent(PythonPackage):
     depends_on('py-greenlet@0.4.13:', when='@:1.4', type=('build', 'run'))
     depends_on('py-zope-event', when='@20.5.1:', type=('build', 'run'))
     depends_on('py-zope-interface', when='@20.5.1:', type=('build', 'run'))
+
+    patch('icc.patch',
+          sha256='44a3aa7ec3a9f056145c25492012f494166dee4c816a3a998a7e3b2d85b57c77',
```
adamjstewart: No need for a sha256 if it's a local file. Also, does this patch apply to all versions of the package?

stephenmsachs: I double checked, it works with all versions.

I had to add the sha to make it work at first. It works without sha now on a fresh Spack installation.

var/spack/repos/builtin/packages/py-gevent/package.py
```diff
@@ -35,3 +35,5 @@ class PyGevent(PythonPackage):
     depends_on('py-greenlet@0.4.13:', when='@:1.4', type=('build', 'run'))
     depends_on('py-zope-event', when='@20.5.1:', type=('build', 'run'))
     depends_on('py-zope-interface', when='@20.5.1:', type=('build', 'run'))
+
+    patch('icc.patch', when='%intel')
```
alalazo, line 40: Can we add a brief description of what the patch does and where does it come from, just above the directive? Other than that LGTM.

alalazo (changes requested):

var/spack/repos/builtin/packages/py-gevent/package.py
stephenmsachs: Done

adamjstewart (approved):

alalazo (approved):

PR #31486, opened 2022-07-07
spiner: update dependencies

rbberger:
Needs https://github.com/spack/spack/pull/31485.

The previous versions of `ports-of-call` had a CMake bug

### DIFF
--- a/var/spack/repos/builtin/packages/spiner/package.py
+++ b/var/spack/repos/builtin/packages/spiner/package.py
@@ -35,8 +35,8 @@ class Spiner(CMakePackage, CudaPackage):
     variant("format", default=False, description="Clang-Format Support")
 
     depends_on("cmake@3.12:")
-    depends_on("catch2@2.13.4:2.13.6")
-    depends_on("ports-of-call@1.1.0:")
+    depends_on("catch2@2.13.4:2.13.9")
+    depends_on("ports-of-call@1.2.0:")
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/spiner/package.py
```diff
@@ -35,8 +35,8 @@ class Spiner(CMakePackage, CudaPackage):
     variant("format", default=False, description="Clang-Format Support")
 
     depends_on("cmake@3.12:")
-    depends_on("catch2@2.13.4:2.13.6")
-    depends_on("ports-of-call@1.1.0:")
+    depends_on("catch2@2.13.4:2.13.9")
+    depends_on("ports-of-call@1.2.0:")
```
tldahlgren, line 39: Is `1.1.0` still relevant for any of the (earlier) versions?

tldahlgren: I ask because, if it is, you can add a `when` argument (see https://spack.readthedocs.io/en/latest/packaging_guide.html#conditional-dependencies) whose value can be a version range.

rbberger: @tldahlgren no, the older version would break in some variants due to a cmake bug. the new version should be used instead.

tldahlgren: I should've paid more attention to the needed PR....

tldahlgren (approved):

PR #31488, opened 2022-07-08
Cppcheck: fix install and convert to cmake build system

white238:
I ran into a problem with building cppcheck.  It was installing the `cfg` directory at the same level as `bin` and got the following error:

```
[white238@rzgenie2:cppcheck-2.8]$ bin/cppcheck --check-config temp.cpp
cppcheck: Failed to load library configuration file 'std.cfg'. File not found
nofile:0:0: information: Failed to load std.cfg. Your Cppcheck installation is broken, please re-install. The Cppcheck binary was compiled without FILESDIR set. Either the std.cfg should be available in bin/cfg or the FILESDIR should be configured. [failedToLoadCfg]
```
I was able to move the `<install prefix>/cfg` to `<install prefix>/bin/cfg` manually and solve the problem.

I was going to just fix the package but after digging into their source realized they have a more maintained CMake build system. I also added a new variant for `rules`.

I tested this with a combination of the following specs on toss3:

- cppcheck
- cppcheck+rules
- cppcheck+htmlreport
- cppcheck+rules+htmlreport

### DIFF
--- a/var/spack/repos/builtin/packages/cppcheck/package.py
+++ b/var/spack/repos/builtin/packages/cppcheck/package.py
@@ -6,7 +6,7 @@
 from spack.package import *
 
 
-class Cppcheck(MakefilePackage):
+class Cppcheck(CMakePackage):
     """A tool for static C/C++ code analysis."""
     homepage = "http://cppcheck.sourceforge.net/"
     url      = "https://downloads.sourceforge.net/project/cppcheck/cppcheck/1.78/cppcheck-1.78.tar.bz2"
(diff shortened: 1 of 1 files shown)


white238:
Can someone with the proper powers restart the pipeline? I am pretty sure this wasn't my fault.

scheibelp:
@spackbot rerun pipeline

scheibelp:
@spackbot run pipeline

becker33 (approved):

PR #30031, opened 2022-04-12
man-db: fix install prefix and library runtime

vvolkl:
`man-db` installed some directories to `$prefix/$prefix` instead of `$prefix` (DESTDIR is needed for some other folders, like this everything seems to be in order)

Also adds a workaround for the `lib/man-db` libdir (while waiting for https://github.com/spack/spack/pull/29657).

PR #31492, opened 2022-07-08
Don't use DESTDIR in man-db

haampie:
Autotools installs to `$DESTDIR/$prefix`, so using `DESTDIR` is a recipe for
disaster... the `man` executable was installed to `<install prefix>/<install
prefix>/bin/man`.


haampie:
ping @vvolkl for next time (slipped in review)

vvolkl:
See also https://github.com/spack/spack/pull/30031, which is mostly obsolete after this is merged.

michaelkuhn (approved):

PR #31503, opened 2022-07-08
py-mypy: add v0.961

adamjstewart:
Successfully builds on macOS 12.4 (Apple M1 Pro) with Python 3.9.13 and Apple Clang 13.1.6.


tldahlgren (approved):
Confirmed all new sha256.

PR #31510, opened 2022-07-09
py-numpy: add v1.23.1

adamjstewart:
https://github.com/numpy/numpy/releases/tag/v1.23.1

### DIFF
--- a/var/spack/repos/builtin/packages/py-numpy/package.py
+++ b/var/spack/repos/builtin/packages/py-numpy/package.py
@@ -23,6 +23,7 @@ class PyNumpy(PythonPackage):
     maintainers = ['adamjstewart', 'rgommers']
 
     version('main', branch='main')
+    version('1.23.1', sha256='d748ef349bfef2e1194b59da37ed5a29c19ea8d7e6342019921ba2ba4fd8b624')
     version('1.23.0', sha256='bd3fa4fe2e38533d5336e1272fc4e765cabbbde144309ccee8675509d5cd7b05')
     version('1.22.4', sha256='425b390e4619f58d8526b3dcf656dde069133ae5c240229821f01b5f44ea07af')
     version('1.22.3', sha256='dbc7601a3b7472d559dc7b933b18b4b66f9aa7452c120e87dfb33d02008c8a18')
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-numpy/package.py
```diff
@@ -102,6 +103,7 @@ class PyNumpy(PythonPackage):
     depends_on('python@3.6:3.10', type=('build', 'link', 'run'), when='@1.19')
     depends_on('python@3.7:3.10', type=('build', 'link', 'run'), when='@1.20:1.21')
     depends_on('python@3.8:', type=('build', 'link', 'run'), when='@1.22:')
+    depends_on('py-packaging', when='@1.21: platform=darwin target=aarch64:', type='build')
```
rgommers: Where does this new requirement come from?

adamjstewart: From the pyproject.toml

rgommers: Ah, there's this:
```
"packaging==20.5; platform_machine=='arm64'",  # macos M1
```
That is only for that version, and while I forgot the reason for it, it should not be needed for newer releases.

adamjstewart: It looks like it's been there since 1.21 and it's still there in main

rgommers: The requirement is there, but it's `==20.5` which is different from `@1.21:`. The dependency is not needed for `>= 1.21`.

adamjstewart: Ah, here `@1.21:` refers to the version of numpy, not the version of packaging

rgommers: Argh, I misread that. Okay now I'm curious where it came from, `numpy` builds just fine without `packaging`. So your change here matches what's in numpy main, but I believe it shouldn't be there.

adamjstewart: Looks like https://github.com/numpy/numpy/pull/18538

rgommers: Indeed no longer needed, see https://github.com/numpy/numpy/pull/21963

adamjstewart: Ah, thanks for clearing that up. The reason I didn't need this in Spack on M1 is because we never pinned setuptools. We still don't support separate concretization of build-only dependencies so we won't be able to pin for a while. Not sure if we should keep this in here for the sake of matching the pyproject.toml in case we pin these in the future or if we should just remove. Any opinions?

rgommers: I'd remove this, because it is just a dependency that isn't needed at all.

rgommers (approved):
LGTM, and verified sha256

tldahlgren (approved):
Confirmed sha256.

PR #31512, opened 2022-07-11
py-hatchling: add new packages, document new build backend

adamjstewart:
Every 6 months or so, PyPA runs out of things to do and rewrites the entire Python build system from scratch. Most recently, they now recommend all new packages use hatchling instead of setuptools. This PR adds the hatchling package and documents it in our PythonPackage build systems guide.

Successfully installs on macOS 12.4 (Apple M1 Pro) with Python 3.9.13 and Apple Clang 13.1.6. Tested by building another package that uses the hatchling build backend (will add in a separate PR).

### DIFF
--- a/lib/spack/docs/build_systems/pythonpackage.rst
+++ b/lib/spack/docs/build_systems/pythonpackage.rst
@@ -48,7 +48,7 @@ important to understand.
 **build backend**
    Libraries used to define how to build a wheel. Examples
    include `setuptools <https://setuptools.pypa.io/>`__,
-   `flit <https://flit.readthedocs.io/>`_, and
+   `flit <https://flit.readthedocs.io/>`_,
    `poetry <https://python-poetry.org/>`_, and
    `hatchling <https://hatch.pypa.io/latest/>`_.
 
@@ -348,7 +348,7 @@ dependencies under the following keys:
 
   This section includes keys with lists of optional dependencies
   needed to enable those features. You should add a variant that
-  optionally adds these dependencies. This variant should be False
+  optionally adds these dependencies. This variant should be ``False``
   by default.
 
 See https://hatch.pypa.io/latest/config/dependency/ for more
--- a/lib/spack/docs/build_systems/pythonpackage.rst
+++ b/lib/spack/docs/build_systems/pythonpackage.rst
@@ -49,7 +49,8 @@ important to understand.
    Libraries used to define how to build a wheel. Examples
    include `setuptools <https://setuptools.pypa.io/>`__,
    `flit <https://flit.readthedocs.io/>`_, and
-   `poetry <https://python-poetry.org/>`_.
+   `poetry <https://python-poetry.org/>`_, and
+   `hatchling <https://hatch.pypa.io/latest/>`_.
 
 ^^^^^^^^^^^
 Downloading
@@ -325,6 +326,33 @@ listed in a ``[tool.poetry.dependencies]`` section, and use a
 for specifying the version requirements. Note that ``~=`` works
 differently in poetry than in setuptools and flit for versions that
 start with a zero.
+
+"""""""""
+hatchling
+"""""""""
+
+If the ``pyproject.toml`` lists ``hatchling.build`` as the
+``build-backend``, it uses the hatchling build system. Look for
+dependencies under the following keys:
+
+* ``requires-python``
+
+  This specifies the version of Python that is required
+
+* ``project.dependencies``
+
+  These packages are required for building and installation. You can
+  add them with ``type=('build', 'run')``.
+
+* ``project.optional-dependencies``
+
+  This section includes keys with lists of optional dependencies
+  needed to enable those features. You should add a variant that
+  optionally adds these dependencies. This variant should be False
+  by default.
+
+See https://hatch.pypa.io/latest/config/dependency/ for more
+information.
 
 """"""
 wheels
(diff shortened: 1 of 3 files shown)


lib/spack/docs/build_systems/pythonpackage.rst
```diff
@@ -49,7 +49,8 @@ important to understand.
    Libraries used to define how to build a wheel. Examples
    include `setuptools <https://setuptools.pypa.io/>`__,
    `flit <https://flit.readthedocs.io/>`_, and
```
tldahlgren:
```suggestion
   `flit <https://flit.readthedocs.io/>`_,
```

lib/spack/docs/build_systems/pythonpackage.rst
```diff
@@ -326,6 +327,33 @@ for specifying the version requirements. Note that ``~=`` works
...
+* ``project.dependencies``
+
+  These packages are required for building and installation. You can
+  add them with ``type=('build', 'run')``.
+
+* ``project.optional-dependencies``
+
+  This section includes keys with lists of optional dependencies
+  needed to enable those features. You should add a variant that
+  optionally adds these dependencies. This variant should be False
```
tldahlgren:
```suggestion
  optionally adds these dependencies. This variant should be ``False``
```

tldahlgren (commented):
Confirmed the sha256 for the two new packages.  Just have a couple of minor docs suggestions.

pradyunsg:
> Most recently, they now recommend all new packages use hatchling instead of setuptools.

Wait, when did this happen? :o

adamjstewart:
> Wait, when did this happen? :o

Less than a month ago: https://github.com/pypa/packaging.python.org/pull/1031

What's even more concerning to me is that hatchling didn't even exist 6 months ago: https://pypi.org/project/hatchling/#history. I can't imagine that it's stable...

pradyunsg:
Well, I wouldn't call that a recommendation but I empathize that I might be too close to the problem from someone else's perspective. :)

FWIW, the `hatchling` project is a supporting project for `hatch`, which was indeed pulled out of it ~6 months ago. Hatch itself has been in development for almost half a decade: https://github.com/pypa/hatch/tree/pre-v1 -- it only got unblocked on core pieces of functionality over the last couple of years (via multiple PEPs, each of which underwent years of design and discussion).

I promise that we're not doing things to primarily annoy redistributors, we're certainly not doing things particularly quickly and we *definitely* have huge backlogs (we ain't running of TODOs). We're just a bundle of volunteers, trying to make the UX of working with Python better.

In the spirit of "criticism is the expression of an unmet need", if you reckon it'd be useful to have a chat about how the Python packaging ecosystem is evolving, I'd be happy to have a chat over on the Spack Slack to understand what parts you're finding frustrating. :)

ofek:
1. Default does not mean explicit recommendation
2. I removed/yanked the pre-public releases e.g. https://discuss.python.org/t/pep-660-editable-installs-for-pep-517-style-build-backends/8510/235
3. It's quite stable and usage is growing https://hatch.pypa.io/latest/users/

tldahlgren (approved):
Thanks for the fixes.

I'm approving this PR based on the corrections but, given the 👎 's in the description, am hesitant to merge ATM.

ofek:
@tldahlgren I think those are just in response to a perceived disregard for all the unpaid volunteer work PyPA members do.

pradyunsg:
I wouldn't hesistate -- as I understand it, the 👎🏽 are based on being disapproving of the language used in the description, not the actual changes. :)

adamjstewart:
Yeah, apologies for the strong wording lol. I have nothing against hatchling, I just don't know anything about it and had never heard of it until it became the default tutorial backend.

My frustration is that every time I cut a new release of my own software, [the packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/) is completely different. Just within setuptools, we've moved from `setup.py` to `setup.cfg` to `pyproject.toml` in a very short time period. For building and uploading, we've moved from `python setup.py upload` to `tox` to `build/twine` in the same time period. At this point, I'm not sure if the right way to build and install wheels is with `pip` or `build/installer`. When converting these to Spack packages, I now need to read `pyproject.toml`, `setup.py`, `setup.cfg`, `requirements.txt`, and several other possible files to determine what dependencies a package has ([trying to automate this in the future](https://github.com/spack/spack/discussions/31413)). 

Once a month or so someone asks me why I'm still using setuptools and not poetry/flit/flavor-of-the-month-popular-new-build-backend-I've-never-heard-of and I ask them what the difference between them is and get no response. From the standpoint of someone who both reads build systems to write Spack build recipes and someone who uses build systems to package my own software, it would be very helpful to have an authoritative and opinionated guide on what build backends exist and which you should use for which project. For example, I've heard that flit is great for simple pure-Python packages but if you have any non-Python code you'll want to use poetry or setuptools instead.

P.S. I'm definitely not trying to discredit all of your hard work. I'm also just a volunteer almost singlehandedly maintaining the entire Python ecosystem in Spack. Recent changes like deprecation of `python setup.py install` have caused me a lot of headaches, but I think I've managed to get Spack back up to state-of-the-art Python packaging. Glad to see there are other core Python developers who even know that Spack exists! Happy to chat more on Slack, and let me know if you see anything out of date in our [build system docs](https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html).

PR #31535, opened 2022-07-12
Liggghts: support lib installation

iWangJiaxiang:
> Thanks for bringing order to HPC software chaos. My organization have been using Spack since 2019 and thanks to Spack we could make software management in order without pain. During this period, we have extended Spack and now we are bringing them back to community. May our contributions be a small step in the great leap of HPC softwawre development.

This PR adds support for lib installation, which is needed by CFDEM software. See https://github.com/spack/spack/pull/31536

### DIFF
--- a/var/spack/repos/builtin/packages/liggghts/package.py
+++ b/var/spack/repos/builtin/packages/liggghts/package.py
@@ -26,7 +26,11 @@ class Liggghts(MakefilePackage):
     variant('profile', default=False,
             description='Generate profiling code')
 
-    depends_on('vtk@6.1.0:8.2.0')
+    variant('libs', default='SHARED',
+            values=('ALL', 'SHARED', 'STATIC', 'NONE'),
+            description='Whether to build libraries of LIGGGHTS')
+
+    depends_on('vtk@6.1.0:8.2.0 +xdmf+mpi')
     depends_on('mpi', when='+mpi')
     depends_on('jpeg', when='+jpeg')
     depends_on('zlib', when="+gzip")
@@ -89,6 +93,10 @@ class Liggghts(MakefilePackage):
         if '+profile' in spec:
             makefile.filter(r'^(USE_PROFILE = ).*', r'\1"ON"')
 
+        if spec.variants['libs'].value != "NONE":
+            makefile.filter(r'^(BUILD_LIBRARIES = ).*',
+                            r'\1"{0}"'.format(spec.variants['libs'].value))
+
         # Enable debug output of Makefile.auto in the log file
         # src/Obj_auto/make_auto.log to quickly troubleshoot if
         # anything goes wrong.
@@ -97,3 +105,19 @@ class Liggghts(MakefilePackage):
     def install(self, spec, prefix):
         mkdir(prefix.bin)
         install(os.path.join('src', 'lmp_auto'), prefix.bin.liggghts)
+
+        if spec.variants['libs'].value != "NONE":
+            mkdir(prefix.lib)
+   
+        if spec.variants['libs'].value == "SHARED":
+            install(os.path.join('src', 'liblmp_auto.so'), prefix.lib)
+        elif spec.variants['libs'].value == "STATIC":
+            install(os.path.join('src', 'liblmp_auto.a'), prefix.lib)
+        elif spec.variants['libs'].value == "ALL":
+            install(os.path.join('src', 'liblmp_auto.so'), prefix.lib)
+            install(os.path.join('src', 'liblmp_auto.a'), prefix.lib)
+        else:
+            pass
+
+        install_tree('src', prefix.src, symlinks=True)
+
--- a/var/spack/repos/builtin/packages/liggghts/package.py
+++ b/var/spack/repos/builtin/packages/liggghts/package.py
@@ -118,7 +118,7 @@ class Liggghts(MakefilePackage):
         libs_in_spec = spec.variants['libs'].value
         if libs_in_spec != None:
             mkdir(prefix.lib)
-            if 'shared' in libs_in_spec:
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/liggghts/package.py
```diff
@@ -26,7 +26,11 @@ class Liggghts(MakefilePackage):
     variant('profile', default=False,
             description='Generate profiling code')
 
-    depends_on('vtk@6.1.0:8.2.0')
+    variant('libs', default='SHARED',
+            values=('ALL', 'SHARED', 'STATIC', 'NONE'),
```
alalazo: Can we use a multi-valued variant here? Also, we use lowercase letters for variant values in all the other packages

var/spack/repos/builtin/packages/liggghts/package.py
```diff
@@ -89,6 +93,10 @@ def edit(self, spec, prefix):
         if '+profile' in spec:
             makefile.filter(r'^(USE_PROFILE = ).*', r'\1"ON"')
 
+        if spec.variants['libs'].value != "NONE":
+            makefile.filter(r'^(BUILD_LIBRARIES = ).*',
+                            r'\1"{0}"'.format(spec.variants['libs'].value))
```
alalazo: Looking at this, I wonder if we can always build both shared and static, and avoid adding the variant completely.

alalazo (changes requested):

alalazo:
@spackbot fix style

alalazo:
@iWangJiaxiang Do you plan to continue working on this PR?

var/spack/repos/builtin/packages/liggghts/package.py
iWangJiaxiang: Thanks for your suggestion.

Technically yes. But the lib is requied by CFDEM package. By adding this variant, if there is existing Liggghts, Spack will install new Liggghts instance with libs ranther than using existing one. If we make lib installation implicit, the user might never know why the CFDEM not working when the have existing Liggghts and might give up using CFDEM with Spack.

Recommendation: Reserving lib variant to make it explicit and setting up the lib installation by default.

iWangJiaxiang:
Sure thing! Working on it.

var/spack/repos/builtin/packages/liggghts/package.py
iWangJiaxiang: Fixed with latest commit

var/spack/repos/builtin/packages/liggghts/package.py
```diff
@@ -26,7 +26,11 @@ class Liggghts(MakefilePackage):
     variant('profile', default=False,
             description='Generate profiling code')
 
-    depends_on('vtk@6.1.0:8.2.0')
+    variant('libs', default='shared, static',
+            description='Build shared or/and static libs',
+            values=('shared', 'static'), multi=True)
```
alalazo:
```suggestion
    variant('libs', default='shared,static',
            description='Build shared or/and static libs',
            values=('shared', 'static'), multi=True)
```

var/spack/repos/builtin/packages/liggghts/package.py
```diff
@@ -89,6 +93,18 @@ def edit(self, spec, prefix):
         if '+profile' in spec:
             makefile.filter(r'^(USE_PROFILE = ).*', r'\1"ON"')
 
+        # Determine how to build libs in LIGGGHTS way
+        libs_in_spec = spec.variants['libs'].value
+        if libs_in_spec != None:
```
alalazo: You can dedent by one level, since the variant cannot be `None`
```suggestion
```

var/spack/repos/builtin/packages/liggghts/package.py
```diff
@@ -97,3 +113,14 @@ def edit(self, spec, prefix):
     def install(self, spec, prefix):
         mkdir(prefix.bin)
         install(os.path.join('src', 'lmp_auto'), prefix.bin.liggghts)
+
+        # Install libs
+        libs_in_spec = spec.variants['libs'].value
+        if libs_in_spec != None:
```
alalazo: Same here

var/spack/repos/builtin/packages/liggghts/package.py
```diff
@@ -97,3 +113,14 @@ def edit(self, spec, prefix):
...
+            mkdir(prefix.lib)
+            if 'shared' in libs_in_spec:
```
alalazo:
```suggestion
            if spec.satisfies("libs=shared"):
```
similarly for `static` down below

alalazo (changes requested):

alalazo (approved):

alalazo:
Hmm, docs are failing. I'll check what is wrong there :thinking:

var/spack/repos/builtin/packages/liggghts/package.py
```diff
@@ -89,6 +93,16 @@ def edit(self, spec, prefix):
         if '+profile' in spec:
             makefile.filter(r'^(USE_PROFILE = ).*', r'\1"ON"')
 
+        # Determine how to build libs in LIGGGHTS way
+        if spec.satisfies("libs=shared") and spec.satisfies("libs=static"):
```
alalazo, line 97: Since docs are failing and we have to perform a few changes anyhow:
```suggestion
        if spec.satisfies("libs=shared,static"):
```
should be a shorter way to express what you already have.

alalazo (changes requested):
We need to fix docs. I'd try to push fixes myself, but don't have enough permissions. Will check what is wrong and possibly post a diff.

alalazo:

PR #17427, opened 2020-07-08
Add `--no-same-owner` flag to tar decompressor

haampie:
Fixes #6759. This should improve untarring files owned by users not available in Rootless Docker builds.

PR #31537, opened 2022-07-12
compression.py: do not extract absolute & relative .. paths

haampie:
Don't extract by absolute paths, and do not allow `..` path components.

Note that the latter is possibly more strict than `tar -x`, the alternative would be to add a depth counter and make sure it's never negative. Another way to handle the former would be to strip the root `/` and still extract, but I don't wanna think about windows drive prefixes etc. 

Fixes a regression caused by Windows support.

### DIFF
--- a/lib/spack/spack/util/compression.py
+++ b/lib/spack/spack/util/compression.py
@@ -60,6 +60,27 @@ def allowed_archive(path):
         any(path.endswith(t) for t in ALLOWED_ARCHIVE_TYPES)
 
 
+def is_trusted_tar_path(path):
+    """When extracting a tar file, make sure we drop absolute paths and
+    paths that go a dir up due to ".." components. This prevents overwriting
+    files outside the target directory."""
+    return not os.path.isabs(path) and ".." not in path.split(os.path.sep)
+
+
+def make_tarinfo_root_owned_when_root(tarinfo):
+    """Only when extracting as effective root user, Python's tarfile logic hits
+    a code path where it tries to chown files to the user/group as set in the
+    tarfile. Of course this is undesirable since uid/gid's are system specific
+    and irrelevant. Unfortunately Python does not offer the equivalent of
+    `tar -o` to prevent this. Therefore we always chown to root when the user
+    extracting is effectively root. In general Spack should not be run by root
+    users, but when using user namespaces to map an unprivileged user to root,
+    this code path is triggered too."""
+    tarinfo.uid = tarinfo.gid = 0
+    tarinfo.uname = tarinfo.gname = 'root'
+    return tarinfo
+
+
 def _untar(archive_file):
     """ Untar archive. Prefer native Python `tarfile`
     but fall back to system utility if there is a failure
@@ -82,12 +103,11 @@ def _untar(archive_file):
 
         # Extract all members but wipe ownership info. This ensures we
         # will not attempt to chown the files as superuser.
-        def filter(tarinfo):
-            tarinfo.uid = tarinfo.gid = 0
-            tarinfo.uname = tarinfo.gname = 'root'
-            return tarinfo
         with tarfile.open(archive_file) as tar:
-            tar.extractall(members=map(filter, tar.getmembers()))
+            members = tar.getmembers()
+            members = filter(lambda t: is_trusted_tar_path(t.path), members)
+            members = map(make_tarinfo_root_owned_when_root, members)
+            tar.extractall(members=members)
     else:
         tar = which('tar', required=True)
         tar.add_default_arg('-oxf')
--- a/lib/spack/spack/test/util/compression.py
+++ b/lib/spack/spack/test/util/compression.py
@@ -96,3 +96,9 @@ def test_get_bad_extension():
 @pytest.mark.parametrize('path', ext_archive.values())
 def test_allowed_archvie(path):
     assert scomp.allowed_archive(path)
+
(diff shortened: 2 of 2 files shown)


johnwparent:
As you mentioned, we could be stripping off leading `/` characters to then extract. IMO this approach is superior, as then we're not silently dropping files out of extracted archives and more closely respecting `tar -x` behavior. I wouldn't want consideration for Windows to be a reason to abandon this approach entirely. To handle on Windows, `os.path.splitdrive` can be used to remove the drive prefix, or some relatively straightforward manual parsing can be done. If this is something you consider unfeasible, then simply ignoring the Windows prefix case and leaving a todo note that Windows handling is still necessary is an approach I'd support as well.

 Edit: Feel free to call out my GH username explicitly in any TODOs relating to Windows.

lib/spack/spack/util/compression.py
```diff
@@ -60,6 +60,27 @@ def allowed_archive(path):
...
+    a code path where it tries to chown files to the user/group as set in the
+    tarfile. Of course this is undesirable since uid/gid's are system specific
+    and irrelevant. Unfortunately Python does not offer the equivalent of
+    `tar -o` to prevent this. Therefore we always chown to root when the user
+    extracting is effectively root. In general Spack should not be run by root
+    users, but when using user namespaces to map an unprivileged user to root,
+    this code path is triggered too."""
+    tarinfo.uid = tarinfo.gid = 0
+    tarinfo.uname = tarinfo.gname = 'root'
+    return tarinfo
```
scheibelp, line 81: (question) are you observing this issue in the final installed prefix? We have explicit permissions-setting logic to assign permissions and I would have thought it would address this (however, if your concern is the intermediate results of unpacking the archive, then that could be useful too - I'd want to think about that though).

lib/spack/spack/util/compression.py
```diff
@@ -60,6 +60,27 @@ def allowed_archive(path):
         any(path.endswith(t) for t in ALLOWED_ARCHIVE_TYPES)
 
 
+def is_trusted_tar_path(path):
+    """When extracting a tar file, make sure we drop absolute paths and
+    paths that go a dir up due to ".." components. This prevents overwriting
+    files outside the target directory."""
+    return not os.path.isabs(path) and ".." not in path.split(os.path.sep)
```
scheibelp, line 67: (No change necessary) Regarding

> Note that the latter is possibly more strict than tar -x, the alternative would be to add a depth counter and make sure it's never negative

I think if `..` is in `normpath`, then that would also handle it (and allow cases where `..` appears but "stays inside" the destination).

scheibelp (commented):
This looks good although I have a question about the permissions: I thought we had logic to set permissions explicitly on the installation prefix, so I'm curious if you're seeing an intermediate issue that is already addressed.

scheibelp:
`https://github.com/python/cpython/issues/94531` looks relevant to this (a recent PR to document the path-handling pitfalls of `tarfile`): that doesn't add anything actionable here but might be useful for folks looking through this later.

lib/spack/spack/util/compression.py
haampie: See also https://github.com/spack/spack/pull/31524 where this was introduced. In short: when extracting as (effective) root user, tarfile chown's to users listed in the tarfile, which makes no sense when extracting on a different system with different users. Before the introduction of tarfile I fixed this issue for `tar` by adding `-o` / `--no-same-owner` in https://github.com/spack/spack/pull/17427

haampie:
There's also a possible issue where a tar entry is a symlink to a dir outside the staging dir and has children of that symlinked dir (I haven't checked how `tar` behaves for this case...). I'm wondering whether it's better to temporarily revert the tarfile changes until these issues are figured out.

Note that of course during the build the build system can also execute arbitrary code and mess up your files, so it would make sense to always use overlayfs or a sandbox when building anything with Spack, but still it'd be nice if untarring and unzipping is guaranteed not to have side effects outside the stage dir.

lib/spack/spack/util/compression.py
scheibelp: My hypothesis (which on further examination I think is wrong) was that while the issue you pointed out exists, it is overruled by the logic at the end of the Spack package installation method (so the files in the stage might exist with strange ownership/permissions, but the files in the install prefix will have sensible ownership and permissions).

I want to confirm: have you seen these ownershipissues propagate to the install directory (based on examining the logic, I wouldn't be surprised if the answer is "yes", but I want to make sure I understand the nature of the problem).

scheibelp:
> I'm wondering whether it's better to temporarily revert the tarfile changes until these issues are figured out.

I think this has merit, and in fact more than that, from reading through https://github.com/spack/spack/pull/31537#issuecomment-1182193854 and associated links more:

* The `tar` exe on Linux has considerations for the original problem (and also I think for https://github.com/spack/spack/pull/31537#issuecomment-1182514785) 
* The `tarfile` module doesn't (either by default or as options) and whether/how to add options to it is a matter of ongoing discussion
* I'd like to avoid reimplementing those checks (although I think you've handled the issue mentioned in https://github.com/spack/spack/pull/31537#issue-1301822517, I think https://github.com/spack/spack/pull/31537#issuecomment-1182514785 would be more complex to handle)

I'll try to handle that tomorrow (which would close this PR, although I assume that would be OK with you)

PR #31539, opened 2022-07-12
Add/packages nextflow tower

marcodelapierre:
This PR adds recipes for two very useful, official utilities that complement the Nextflow workflow manager:
* Tower Agent: https://github.com/seqeralabs/tower-agent
* Tower CLI: https://github.com/seqeralabs/tower-cli

On the side, I also added the most recent stable version of Nextflow in the corresponding recipe.

### DIFF
--- a/var/spack/repos/builtin/packages/tower-agent/package.py
+++ b/var/spack/repos/builtin/packages/tower-agent/package.py
@@ -1,0 +1,23 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class TowerAgent(Package):
+    """Tower Agent allows Nextflow Tower to launch pipelines
+       on HPC clusters that do not allow direct access through
+       an SSH client.
+    """
+
+    homepage = "https://github.com/seqeralabs/tower-agent"
+    url      = "https://github.com/seqeralabs/tower-agent/releases/download/v0.4.3/tw-agent-linux-x86_64"
+
+    version('0.4.3', sha256='1125e64d4e3342e77fcf7f6827f045e421084654fe8faafd5389e356e0613cc0', expand=False)
+
+    def install(self, spec, prefix):
+        mkdirp(prefix.bin)
+        install(self.stage.archive_file, join_path(prefix.bin, "tw-agent"))
+        set_executable(join_path(prefix.bin, "tw-agent"))
--- a/var/spack/repos/builtin/packages/tower-agent/package.py
+++ b/var/spack/repos/builtin/packages/tower-agent/package.py
@@ -19,8 +19,10 @@ class TowerAgent(Package):
 
     if platform.machine().startswith('x86_64'):
         if sys.platform.startswith('linux'):
-            url      = "https://github.com/seqeralabs/tower-agent/releases/download/v0.4.3/tw-agent-linux-x86_64"
-            version('0.4.3', sha256='1125e64d4e3342e77fcf7f6827f045e421084654fe8faafd5389e356e0613cc0', expand=False)
+            version('0.4.3',
+                    sha256='1125e64d4e3342e77fcf7f6827f045e421084654fe8faafd5389e356e0613cc0',
+                    url='https://github.com/seqeralabs/tower-agent/releases/download/v0.4.3/tw-agent-linux-x86_64',
+                    expand=False)
 
     def install(self, spec, prefix):
         mkdirp(prefix.bin)
--- a/var/spack/repos/builtin/packages/tower-agent/package.py
+++ b/var/spack/repos/builtin/packages/tower-agent/package.py
@@ -13,9 +13,10 @@ class TowerAgent(Package):
     """
 
     homepage = "https://github.com/seqeralabs/tower-agent"
-    url      = "https://github.com/seqeralabs/tower-agent/releases/download/v0.4.3/tw-agent-linux-x86_64"
 
-    version('0.4.3', sha256='1125e64d4e3342e77fcf7f6827f045e421084654fe8faafd5389e356e0613cc0', expand=False)
+    if sys.platform.startswith('linux'):
+        url      = "https://github.com/seqeralabs/tower-agent/releases/download/v0.4.3/tw-agent-linux-x86_64"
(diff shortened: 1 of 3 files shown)


marcodelapierre:
@pditommaso any comment on the Spack recipes for Tower Agent and CLI? I hope they can be useful to the wider Nextflow community!

pditommaso:
Umm, we are making distribution releases via https://jreleaser.org/ which has no support for it, and I have to say I have no experience with Spack. 

Maybe you want to contribute it?

marcodelapierre:
Yes @pditommaso! So, the recipes I am providing in this PR make TW Agent and CLI available to install within the Spack package manager, which is quite popular across many HPC centres worldwide. It is very powerful for deploying both centralised and user-only installations of collections of scientific software.

The recipes I have made make use of the binary releases that are available on the Github repos of the two tools. Other than the binaries, do you think there are other components that need downloading, or general purposes configurations that need to be set? I don't think so, but I wanted to double-check with you.
At this stage, I don't think there are further integrations that are possible with Jreleaser (but thanks for mentioning the tool).

pditommaso:
Awesome! I got confused, I was thinking it was a ticket in the tower CLI repo! 

> Other than the binaries, do you think there are other components that need downloading, or general purposes configurations that need to be set?

The binaries are self-contained. Therefore should not be needed anything else.

dialvarezs:
Do tower-cli and tower-agent work on other architectures aside from x86_64? Because with these recipes, I guess spack will try to install the packages on any architecture (I don't have a non-x86 machine on hand to test it anyways).

marcodelapierre:
Good point @dialvarezs !
They indeed only work on `x86_64` archs. I have added a condition for this in the recipes.

var/spack/repos/builtin/packages/tower-agent/package.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import platform
+import sys
```
dialvarezs:
```suggestion
```

var/spack/repos/builtin/packages/tower-agent/package.py
```diff
@@ -0,0 +1,30 @@
...
+class TowerAgent(Package):
+    """Tower Agent allows Nextflow Tower to launch pipelines
+       on HPC clusters that do not allow direct access through
+       an SSH client.
+    """
+
+    homepage = "https://github.com/seqeralabs/tower-agent"
+
+    if platform.machine().startswith('x86_64'):
+        if sys.platform.startswith('linux'):
```
dialvarezs:
```suggestion
    if platform.machine() == 'x86_64':
        if platform.system() == 'Linux':
```

var/spack/repos/builtin/packages/tower-cli/package.py
```diff
@@ -0,0 +1,35 @@
...
+class TowerCli(Package):
+    """Tower on the Command Line brings Nextflow Tower concepts
+       including Pipelines, Actions and Compute Environments
+       to the terminal.
+    """
+
+    homepage = "https://github.com/seqeralabs/tower-cli"
+
+    if platform.machine().startswith('x86_64'):
+        if sys.platform == 'darwin':
```
dialvarezs:
```suggestion
    if platform.machine() == 'x86_64':
        if platform.system() == 'Darwin':
```

var/spack/repos/builtin/packages/tower-cli/package.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import platform
+import sys
```
dialvarezs:
```suggestion
```

var/spack/repos/builtin/packages/tower-cli/package.py
```diff
@@ -0,0 +1,35 @@
...
+
+    homepage = "https://github.com/seqeralabs/tower-cli"
+
+    if platform.machine().startswith('x86_64'):
+        if sys.platform == 'darwin':
+            version('0.6.2',
+                    sha256='2bcc17687d58d4c888e8d57b7f2f769a2940afb3266dc3c6c48b0af0cb490d91',
+                    url='https://github.com/seqeralabs/tower-cli/releases/download/v0.6.2/tw-0.6.2-osx-x86_64',
+                    expand=False)
+        elif sys.platform.startswith('linux'):
```
dialvarezs:
```suggestion
        elif platform.system() == 'Linux':
```

dialvarezs (changes requested):
A few last suggestions on platform checking

marcodelapierre:
Thanks, all addressed!

dialvarezs (approved):
Great, we're ready then. Thank you!

dialvarezs (approved):

alalazo (approved):

aalmiray:
FWIW JReleaser has an [open ticket](https://github.com/jreleaser/jreleaser/issues/834) for adding Spack support. Any tips would be greatly appreciated to get it off the ground.

cc: @pditommaso  @marcodelapierre

marcodelapierre:
@aalmiray @pditommaso asking a naive question, but apart from Nextflow I am out of the Java galaxy: I can think of JReleaser a bit like the Java equivalent of pip/pypi for Python, isn't it?

Spack supports various types of packages/build systems: https://spack.readthedocs.io/en/latest/build_systems.html
Would it make sense to target the design of a new build system for Java, using JReleaser under the hood?

Note I am not part of the core dev team of Spack, at  this stage I am just throwing ideas to brainstorm.

aalmiray:
@marcodelapierre Thanks for the quick reply. I'm new to Spack so I got lost pretty quickly. Your templates have proved useful to figure out the install procedure of a generic CLI tool packaged as a zip or tar. Still looking for options to get an archive such as

```
    app-1.0.0
    ├── bin
    │   ├── app
    │   └── app.bat
    └── lib
        └── app-1.0.0.jar
```

To be installed using a basic Spack template.

marcodelapierre:
Have a look at this section of the docs:
https://spack.readthedocs.io/en/latest/packaging_guide.html ,

and in particular:
https://spack.readthedocs.io/en/latest/packaging_guide.html#file-functions

You might find useful directions.

PR #31542, opened 2022-07-12
WarpX: Add sensei insitu variant

kwryankrattiger:
Add variant to enable sensei adapter in warpx

### DIFF
--- a/var/spack/repos/builtin/packages/warpx/package.py
+++ b/var/spack/repos/builtin/packages/warpx/package.py
@@ -87,6 +87,8 @@ class Warpx(CMakePackage):
             description='Enable tiny profiling features')
 
     depends_on('sensei@4.0.0:', when='@22.07: +sensei')
+    conflicts('+sensei', when='@:22.06',
+              msg='WarpX supports SENSEI 4.0+ with 22.07 and newer')
 
     depends_on('ascent', when='+ascent')
     depends_on('ascent +cuda', when='+ascent compute=cuda')
--- a/var/spack/repos/builtin/packages/warpx/package.py
+++ b/var/spack/repos/builtin/packages/warpx/package.py
@@ -47,7 +47,9 @@ class Warpx(CMakePackage):
     variant('app', default=True,
             description='Build the WarpX executable application')
     variant('ascent', default=False,
-            description='Enable Ascent in situ vis')
+            description='Enable Ascent in situ visualization')
+    variant('sensei', default=False,
+            description='Enable SENSEI in situ visualization')
     variant('compute',
             default='omp',
             values=('omp', 'cuda', 'hip', 'sycl', 'noacc'),
@@ -84,11 +86,10 @@ class Warpx(CMakePackage):
     variant('tprof', default=True,
             description='Enable tiny profiling features')
 
+    depends_on('sensei@4.0.0:', when='@22.07: +sensei')
+
     depends_on('ascent', when='+ascent')
-    # note: ~shared is only needed until the new concretizer is in and
-    #       honors the conflict inside the Ascent package to find this
-    #       automatically
-    depends_on('ascent +cuda ~shared', when='+ascent compute=cuda')
+    depends_on('ascent +cuda', when='+ascent compute=cuda')
     depends_on('ascent +mpi', when='+ascent +mpi')
     depends_on('boost@1.66.0: +math', when='+qedtablegen')
     depends_on('cmake@3.15.0:', type='build')
@@ -166,6 +167,7 @@ class Warpx(CMakePackage):
             # variants
             self.define_from_variant('WarpX_APP', 'app'),
             self.define_from_variant('WarpX_ASCENT', 'ascent'),
+            self.define_from_variant('WarpX_SENSEI', 'sensei'),
             '-DWarpX_COMPUTE={0}'.format(
                 spec.variants['compute'].value.upper()),
             '-DWarpX_DIMS={0}'.format(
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -83,11 +85,14 @@ class Warpx(CMakePackage):
...
             description='Enable tiny profiling features')
 
+    depends_on('sensei', when='+sensei')
+    depends_on('sensei +mpi', when='+sensei +mpi')
+
     depends_on('ascent', when='+ascent')
     # note: ~shared is only needed until the new concretizer is in and
     #       honors the conflict inside the Ascent package to find this
     #       automatically
-    depends_on('ascent +cuda ~shared', when='+ascent compute=cuda')
```
kwryankrattiger, line 91: This was fixed/not needed with ascent 0.8 anymore.

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -83,11 +85,13 @@ class Warpx(CMakePackage):
     variant('tprof', default=True,
             description='Enable tiny profiling features')
 
+    depends_on('sensei', when='+sensei')
```
ax3l: Let's push this to
```suggestion
    depends_on('sensei@4.0.0:', when='+sensei')
    conflicts('+sensei', when='@:22.06',
              msg='WarpX supports SENSEI 4.0+ with 22.07 and newer')
```
since this is the version range we require since https://github.com/ECP-WarpX/WarpX/pull/3211

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -184,6 +189,9 @@ def cmake_args(self):
         with when('+openpmd'):
             args.append('-DWarpX_openpmd_internal=OFF')
 
+        if '+sensei' in spec:
+            args.append(self.define('SENSEI_DIR', os.path.join(spec['sensei'].prefix.lib, 'cmake')))
```
ax3l: Why do we need to set this? Aren't the usual [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html) hints sufficient in Spack?

Or is the SENSEI CMake Config file installed in a non-default location of the SENSEI install prefix and needs to be fixed?
https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure

ax3l: Please use Spack's `join_path` instead of `os.path` (please see examples in other locations of the same file :) )

var/spack/repos/builtin/packages/warpx/package.py
ax3l: Thanks, can you please also remove the comment above that documents why we needed `~shared`?

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -83,11 +85,13 @@ class Warpx(CMakePackage):
     variant('tprof', default=True,
             description='Enable tiny profiling features')
 
+    depends_on('sensei', when='+sensei')
+
     depends_on('ascent', when='+ascent')
     # note: ~shared is only needed until the new concretizer is in and
     #       honors the conflict inside the Ascent package to find this
     #       automatically
```
ax3l:
```suggestion
```

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -47,6 +47,8 @@ class Warpx(CMakePackage):
             description='Build the WarpX executable application')
     variant('ascent', default=False,
             description='Enable Ascent in situ vis')
+    variant('sensei', default=False,
+            description='Enable SENSEI in situ vis')
```
ax3l: I know I added this, let's write the full `visualization` for readability :-)
```suggestion
            description='Enable Ascent in situ visualization')
    variant('sensei', default=False,
            description='Enable SENSEI in situ visualization')
```

var/spack/repos/builtin/packages/warpx/package.py
kwryankrattiger: I don't really understand why, but sometimes reading the CMAKE_PREFIX_PATH gets confused inside of the spack build environment. I have had to do this for a couple different projects but haven't spent too much time trying to understand why

kwryankrattiger:
@ax3l This should be ready now.

ax3l (dismissed):
Awesome :)

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -84,11 +86,10 @@ class Warpx(CMakePackage):
     variant('tprof', default=True,
             description='Enable tiny profiling features')
 
+    depends_on('sensei@4.0.0:', when='@22.07: +sensei')
```
ax3l, line 89:
```suggestion
    depends_on('sensei@4.0.0:', when='@22.07: +sensei')
    conflicts('+sensei', when='@:22.06',
              msg='WarpX supports SENSEI 4.0+ with 22.07 and newer')
```

var/spack/repos/builtin/packages/warpx/package.py
ax3l: added the conflict now

ax3l (approved):

var/spack/repos/builtin/packages/warpx/package.py
ax3l: ok, makes sense

ax3l: Found the reason why this is needed:
Sensei does not yet install in the right standardized cmake paths. With 4.0, the config files land in `lib/cmake/` instead of `lib/cmake/sensei` on my Linux machine.
https://cmake.org/cmake/help/v3.24/command/find_package.html#config-mode-search-procedure

Docs: #32276

PR #25823, opened 2021-09-07
trilinos: variant for libx11

wortiz:
SEACAS only uses X11 for blot/fastq as far as I'm aware.

blot/fastq are pretty niche/legacy programs, this re-adds a variant to avoid bringing in X11 by default

@sethrj

PR #31546, opened 2022-07-12
trilinos: modify fortran linking flags

kuberry:
Past solutions https://github.com/spack/spack/pull/25823 for linking to gfortran made use of the `Trilinos_EXTRA_LINK_FLAGS`. However, these flags are not propagated to all Trilinos packages that need them (such as KokkosKernels). 

`CMAKE_SHARED_LINKER_FLAGS` is always propagated to all Trilinos packages, so this fix switches to using `CMAKE_SHARED_LINKER_FLAGS` instead of `Trilinos_EXTRA_LINK_FLAGS` for Trilinos versions less than 14.0.

Even though 14.0 is not released yet, it will be soon and will have modern CMake targets and will fix the propagation of `Trilinos_EXTRA_LINK_FLAGS` so we can continue to use this variable from version 14.0 onward.

### DIFF
--- a/var/spack/repos/builtin/packages/trilinos/package.py
+++ b/var/spack/repos/builtin/packages/trilinos/package.py
@@ -832,11 +832,18 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
                 libgfortran = fc('--print-file-name',
                                  'libgfortran.a',
                                  output=str).strip()
-            # -L<libdir> -lgfortran required for OSX
-            # https://github.com/spack/spack/pull/25823#issuecomment-917231118
-            options.append(
-                define('Trilinos_EXTRA_LINK_FLAGS',
-                       '-L%s/ -lgfortran' % os.path.dirname(libgfortran)))
+            if spec.version < Version('14'):
+                # Trilinos_EXTRA_LINK_FLAGS do not correctly propagate to
+                # Kokkos-Kernels prior to version 14.0
+                options.append(
+                    define('CMAKE_SHARED_LINKER_FLAGS',
+                           '-L%s/ -lgfortran' % os.path.dirname(libgfortran)))
+            else:
+                # -L<libdir> -lgfortran required for OSX
+                # https://github.com/spack/spack/pull/25823#issuecomment-917231118
+                options.append(
+                    define('Trilinos_EXTRA_LINK_FLAGS',
+                           '-L%s/ -lgfortran' % os.path.dirname(libgfortran)))
 
         if sys.platform == 'darwin' and macos_version() >= Version('10.12'):
             # use @rpath on Sierra due to limit of dynamic loader


var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -832,11 +832,18 @@ def define_tpl(trilinos_name, spack_name, have_dep):
...
+                # Kokkos-Kernels prior to version 14.0
+                options.append(
+                    define('CMAKE_SHARED_LINKER_FLAGS',
+                           '-L%s/ -lgfortran' % os.path.dirname(libgfortran)))
+            else:
+                # -L<libdir> -lgfortran required for OSX
+                # https://github.com/spack/spack/pull/25823#issuecomment-917231118
+                options.append(
+                    define('Trilinos_EXTRA_LINK_FLAGS',
+                           '-L%s/ -lgfortran' % os.path.dirname(libgfortran)))
```
sethrj, line 846: Is there a reason we can't use the spack wrapper flags (`flag_handler`) instead? That should cover all cases that call the compilers. I know this code is very old so that might've been just an oversight.

kuberry: Hi @sethrj. This problem comes up when gfortran (not built by spack) is used to build a static library (in this case openblas) and the static library is linked against by another package with shared libraries. There are a few ways to address this:

1.) Setting 
```yaml
compilers:
- compiler:
    spec: apple-clang@13.1.6
    paths:
      cc: /usr/bin/clang
      cxx: /usr/bin/clang++
      f77: /usr/local/bin/gfortran
      fc: /usr/local/bin/gfortran
    flags:
      ldflags: ' -L/usr/local/Cellar/gcc/11.3.0_2/lib/gcc/11/libgfortran.dylib -lgfortran '
```
or
setting 
```yaml
compilers:
- compiler:
    spec: apple-clang@13.1.6
    paths:
      cc: /usr/bin/clang
      cxx: /usr/bin/clang++
      f77: /usr/local/bin/gfortran
      fc: /usr/local/bin/gfortran
    flags:
      ldflags: -lgfortran
    extra_rpaths:
    - /usr/local/Cellar/gcc/11.3.0_2/lib/gcc/11/
```
Since the issue stems from a misconfigured combination of Clang/gfortran, this seems like a reasonable fix (fixing the configuration). All of the following options are meant to handle the misconfigured compilers after the fact, right?

2.) The fix in this PR that uses `CMAKE_SHARED_LINKER_FLAGS` / `Trilinos_EXTRA_LINK_FLAGS`.

3.) Have openblas provide the extra fortran flags needed so that any package upstream of it will inherit the correct flags.

4.) Use the `flag_handler` to do an equivalent to 2

5.) Open for suggestion

sethrj: I think it's not exclusively due to misconfigured libraries: it might just be the build systems not propagating the implicit language flags. If an upstream static library Food that builds a Fortran-based library and a `FooConfig.cmake`, then CMake will use [`CMAKE_<LANG>_IMPLICIT_LINK_LIBRARIES`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_IMPLICIT_LINK_LIBRARIES.html) to propagate `-lgfortran` etc. when linking downstream C++-only code against that fortran code. Presumably because of tribits (which doesn't [at least as of 13.2] look for `Config.cmake` files) and possibly inadequate upstream static installations of fortran tools (such as netlib blas/lapack, openblas), Trilinos could fail when using the C++ linker.

I think the best (most Spack-y) option is number 4, because it's a problem with the combination of trilinos and upstream library.

kuberry: Hi @sethrj,  I'm working towards number 4, but handling it in the `flag_handler`, and I've run into an issue. 

```py
fc = Executable(spec["mpi"].mpifc) if ("+mpi" in spec) else Executable(spack_fc)
```
with 

```py
fc = Executable(spec["mpi"].mpifc)
``` 
producing

```out
==> [2022-10-03-09:38:15.241117] Error: AttributeError: 'Spec' object has no attribute 'mpifc'

/Users/pakuber/test_spack_xyce/spack/var/spack/repos/builtin/packages/trilinos/package.py:480, in flag_handler:
        477                flags.append("-Wl,-undefined,dynamic_lookup")
        478            # Fortran lib (assumes clang is built with gfortran!)
        479            if "+fortran" in spec and spec.compiler.name in ["gcc", "clang", "apple-clang"]:
  >>    480                fc = Executable(spec["mpi"].mpifc) if ("+mpi" in spec) else Executable(spack_fc)
        481                libgfortran = fc("--print-file-name", "libgfortran." + dso_suffix, output=str).strip()
        482                # if libgfortran is equal to "libgfortran.<dso_suffix>" then
        483                # print-file-name failed, use static library instead
```
and 
```py
fc = Executable(spack_fc)
```
results in

```out
==> [2022-10-03-11:48:53.405227] Error: ProcessError: Command exited with status 1:
    '/Users/pakuber/test_spack_xyce/spack/lib/spack/env/clang/gfortran' '--print-file-name' 'libgfortran.dylib'
[spack cc] ERROR: Spack compiler must be run from Spack! Input 'SPACK_ENV_PATH' is missing
```
so it seems like at the time of the call to `flag_handler`, the environment is not yet populated.

kuberry: On further inspection it looks like in

https://github.com/spack/spack/blob/develop/lib/spack/spack/build_environment.py#L807-L809

```py
    if need_compiler:
        set_compiler_environment_variables(pkg, env_mods)
        set_wrapper_variables(pkg, env_mods)
```
`set_wrapper_variables` populates `env_mods` which are required by `set_compiler_environment_variables`calling the `flag_handler`, to use `Executable()`. 

Later in `setup_package` env_mods is used to populate the environment, and it isn't until after that point that `Executable` can be used with success. This looks like it will require a workaround to move a call to `spack_fc` into the `flag_handler`.

PR #31557, opened 2022-07-13
xgboost: Add version 1.6.1 and make openmp variant compatible with clang

haralmha:


haralmha:
First build with clang 14.0.6 triggered the cmake error:
```
Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)
```
After assigning these:
```
Could NOT find OpenMP_C (missing: OpenMP_libomp_LIBRARY)
```
After setting this:
```
Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
```
After setting these the installation passed the cmake step, and this build error occurred:
```
cannot find -llibomp
```
So I assigned `OpenMP_libomp_LIBRARY` to its shared library, and got:
```
undefined reference to `hwloc_bitmap_zero'
undefined reference to `hwloc_topology_get_support'
etc...
```
And therefore I added a hwloc dependency for clang.

var/spack/repos/builtin/packages/xgboost/package.py
```diff
@@ -59,4 +62,24 @@ def cmake_args(self):
...
+            OpenMP_C_LIB_NAMES = "libomp"
+            args += [
+                self.define('OpenMP_C_FLAGS', OpenMP_C_FLAGS),
+                self.define('OpenMP_C_LIB_NAMES', OpenMP_C_LIB_NAMES),
+                self.define('OpenMP_CXX_FLAGS', OpenMP_C_FLAGS),
+                self.define('OpenMP_CXX_LIB_NAMES', OpenMP_C_LIB_NAMES),
+            ]
+            shared_lib_suffix = 'so'
+            if self.spec.satisfies('platform=darwin'):
+                shared_lib_suffix = 'dylib'
```
adamjstewart: Spack packages have a `dso_suffix` variable you can use, but see below

var/spack/repos/builtin/packages/xgboost/package.py
```diff
@@ -59,4 +62,24 @@ def cmake_args(self):
...
+            clang = self.compiler.cc
+            clang_bin = os.path.dirname(clang)
+            clang_root = os.path.dirname(clang_bin)
+            args += [
+                self.define('OpenMP_libomp_LIBRARY', clang_root +
+                            '/lib/libomp.' + shared_lib_suffix)
```
adamjstewart: Might be more reliable to use `find_library` here, then you just give it a root directory and a filename (without a suffix) and it will find the right file for you on any platform, even if it's in `lib64` instead of `lib`

adamjstewart (commented):
Confirmed that this new version builds on macOS with Apple Clang and doesn't require the Clang-specific hacks.

I wonder if it would be easier to just use `llvm-openmp` with Clang too. Or maybe that would break if the versions don't match.

adamjstewart:
Can you also update py-xgboost?

var/spack/repos/builtin/packages/py-xgboost/package.py
```diff
@@ -26,10 +27,10 @@ class PyXgboost(PythonPackage):
     variant('dask',         default=False, description='Enables Dask extensions for distributed training.')
     variant('plotting',     default=False, description='Enables tree and importance plotting.')
 
-    for ver in ['1.3.3', '1.5.2']:
+    for ver in ['1.3.3', '1.5.2', '1.6.1']:
         depends_on('xgboost@' + ver, when='@' + ver)
 
-    depends_on('python@3.6:',   type=('build', 'run'))
+    depends_on('python@3.7:',   type=('build', 'run'))
```
adamjstewart:
```suggestion
    depends_on('python@3.7:', when='@1.6:', type=('build', 'run'))
    depends_on('python@3.6:',   type=('build', 'run'))
```

adamjstewart (changes requested):

adamjstewart (approved):

PR #27987, opened 2021-12-14
Configure requirements for a package

scheibelp:
Spack doesn't have an easy way to say something like "If I build package X, then I ~want~ *need* version Y":

* If you specify something on the command line, then you ensure that the constraints are applied, but the package is always built
  * Likewise if you `spack add X...` to your environment, the constraints are guaranteed to hold, but the environment always builds the package
* Conversely, you can add preferences to `packages.yaml`, but these are not guaranteed (Spack can choose other settings)

For example, you might want to specify that when you build `openmpi`, that you build `@4.0.1`: you want to allow building with other specs, but always use `@4.0.1` if building `openmpi`.

This PR adds ASP rules (and logic to `asp.py`) which can generate conditional constraints. ~My current guess is that it should be "consolidated" with `asp_spec_clauses` (or rather, that function should be made more generic to support this case).~ (EDIT: this last part is now addressed)

(Update from Jan) The schema for expressing this is a new `packages.yaml` section:

```
packages:
  zlib:
    require: '@1.2.12'                             # Only zlib 1.2.12 can be used
  libpng:
    require:
    - one_of: ['%gcc@8.4.0', '%clang@9.0.0']       # libpng can only build with one of these two compilers
    - one_of: ["@versionx", "@versiony"]
```

For any given `<package>` subsection, you can now add a `require` subsection, where you can add demands for the spec. The details you add there *must* be respected by Spack (i.e. it will either match the spec there, or fail to concretize).

TODOs:

- [x] This will require updating the YAML schema in `packages.yaml` (or some config file). (EDIT 1/7/22: that is now done) ~For now, since I'm experimenting with the ASP rules, I am manually generating what would be parsed from the config~
- [x] Tests

PR #31558, opened 2022-07-13
ASP-based solver: allow to reuse installed externals

### LINKED ISSUE #31484: Concretizer: reuse broken for external packages
### Steps to reproduce

First, blow away `~/.spack`, clone a fresh copy of Spack, and check for external packages:
```console
$ git clone https://github.com/spack/spack.git
$ . spack/share/spack/setup-env.sh
$ spack external find --not-buildable
```
This will register all external packages it finds on the system in `packages.yaml`. We'll focus on git since it's guaranteed to be found, but this affects all external packages.

Next, install git:
```console
$ spack install git
[+] /usr (external git-2.32.1-lweosdwarje645t4pqta7becf7n5dwbc)
$ spack find -l git
==> 1 installed package
-- darwin-monterey-m1 / apple-clang@13.1.6 ----------------------
lweosdw git@2.32.1
[12:52:36] ajstewart@ajstewart-mbp:~
$ spack spec -Il --reuse git
Input spec
 -   git

Concretized
[+]  lweosdw  git@2.32.1%apple-clang@13.1.6+man+nls+perl+subtree~svn~tcltk arch=darwin-monterey-m1
```
Next, modify the git package. For example, add a new variable:
```diff
diff --git a/var/spack/repos/builtin/packages/git/package.py b/var/spack/repos/builtin/packages/git/package.py
index 2d9a7a06f1..71d8d3e6c3 100644
--- a/var/spack/repos/builtin/packages/git/package.py
+++ b/var/spack/repos/builtin/packages/git/package.py
@@ -18,6 +18,7 @@ class Git(AutotoolsPackage):
     homepage = "http://git-scm.com"
     url      = "https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.12.0.tar.gz"
     maintainers = ['jennfshr']
+    foo = 'bar'
 
     tags = ['build-tools']
```
Now, Spack will no longer be able to reuse the existing git installation:
```console
$ spack spec -Il --reuse git
Input spec
 -   git

Concretized
 -   634kih5  git@2.32.1%apple-clang@13.1.6+man+nls+perl+subtree~svn~tcltk arch=darwin-monterey-m1

$ spack install --reuse git
[+] /usr (external git-2.32.1-634kih5iv3oxsqem6mrsz75t3zkdwxml)
```
This introduces a lot of inconsistencies in the database. `spack find` claims to only find a single installation, but `spack uninstall` begs to differ:
```
$ spack find -l git
==> 1 installed package
-- darwin-monterey-m1 / apple-clang@13.1.6 ----------------------
lweosdw git@2.32.1
$ spack uninstall git
==> Error: git matches multiple packages:

    -- darwin-monterey-m1 / apple-clang@13.1.6 ----------------------
    lweosdw git@2.32.1  634kih5 git@2.32.1

==> Error: You can either:
    a) use a more specific spec, or
    b) specify the spec by its hash (e.g. `spack uninstall /hash`), or
    c) use `spack uninstall --all` to uninstall ALL matching specs.
```
This also has a recursive effect. If a single package in the DAG is modifed and another package in the DAG is external, Spack will reinstall the modified package and anything that depends on it. Effectively, this means that `--reuse` no longer works if you use external packages and update Spack on occasion.

This is particularly problematic on macOS because many packages depend on `unwind` and `uuid` and Spack ships with a `packages.yaml` that registers these as external. Since Python depends on `uuid`, Spack reinstalls all of my Python libraries any time any file is modified, just like in the days before `--reuse` was added, but worse because we now use a full hash.

### Error message

_No response_

* **Spack:** 0.19.0.dev0 (a12ce9fd59c32e650b43610925e454322995a81d)
* **Python:** 3.9.12
* **Platform:** darwin-monterey-m1
* **Concretizer:** clingo

@tgamblin @alalazo 

### General information

### LINKED ISSUE #31832: Attempted to use external for 'apple-libuuid' which does not satisfy any configured external spec
### Steps to reproduce

My pretty-fresh spack installation has ROOT installed:

<details><summary>Ugly ROOT spec</summary><pre>
53qeuen root@6.26.04~aqua~davix~dcache~emacs~examples~fftw~fits~fortran+gdml+gminimal~graphviz~gsl~http~ipo~jemalloc~math~memstat+minuit~mlp~mysql~opengl~oracle~postgres~pythia6~pythia8+python~qt4~r+roofit~root7+rpath~shadow~spectrum~sqlite~ssl~table~tbb+threads~tmva+unuran~vc+vdt~veccore~vmc~x+xml~xrootd build_type=RelWithDebInfo cxxstd=17 patches=22af347
jjddmrr     cmake@3.23.2~doc+ncurses+ownlibs~qt build_type=Release
muw42gz         ncurses@6.2~symlinks+termlib abi=none
6ge5mev             gnuconfig@2021-08-14
ett7ygi             pkgconf@1.8.0
373h7tj         openssl@1.1.1q~docs~shared certs=mozilla patches=3fdcf2d
u5bzvfg             ca-certificates-mozilla@2022-03-29
256wjmr             perl@5.34.1+cpanm+shared+threads
e5dqqjq                 berkeley-db@18.1.40+cxx~docs+stl patches=b231fcc
rjsj6wi                 bzip2@1.0.8~debug~pic+shared
mbrctz7                     diffutils@3.8
fh6dfpn                         libiconv@1.16 libs=shared,static
chvvbrt                 gdbm@1.19
ua734bl                     readline@8.1.2
ke4alug                 zlib@1.2.12+optimize+pic+shared patches=0d38234
3s3dqxu     freetype@2.11.1
6kpcuy2         libpng@1.6.37
oronjek     libice@1.0.9
sombxov         util-macros@1.19.3
3myktsy         xproto@7.0.31
raxqbvb         xtrans@1.3.5
abkohk4     libjpeg-turbo@2.1.3
3qt26b3         nasm@2.15.05
ozh3lkm     libxml2@2.9.13~python
gv3m23p         xz@5.2.5~pic libs=shared,static
bhvuert     lz4@1.9.3 libs=shared,static
eomd6ql     nlohmann-json@3.10.5~ipo~multiple_headers build_type=RelWithDebInfo
nhld2jy     openblas@0.3.20~bignuma~consistent_fpcsr~ilp64+locking+pic+shared symbol_suffix=none threads=none
o6y4zfu     pcre@8.45~jit+multibyte+utf
vq2jw2b     python@3.9.13+bz2+ctypes+dbm~debug+ensurepip+libxml2+lzma~nis~optimizations+pic+pyexpat+pythoncmd+readline+shared+sqlite3+ssl~tix~tkinter~ucs4+uuid+zlib patches=0d98e93,4c24573,f2fd060
jbbx4ar         apple-libuuid@1353.100.2
f5zknwn         expat@2.4.8~libbsd
fvtlcau         gettext@0.21+bzip2+curses+git~libunistring+libxml2+tar+xz
axle3yr             tar@1.34 zip=pigz
zzd6h23                 pigz@2.7
cx7gkqf                 zstd@1.5.2+programs compression=none libs=shared,static
rjvdm33         libffi@3.4.2
b567lts         sqlite@3.38.5+column_metadata+dynamic_extensions+fts~functions+rtree
lqe2aod     unuran@1.8.1~gsl+rngstreams+shared
fwwrbyt         rngstreams@1.0.1
acznmsy     vdt@0.4.3~ipo~preload build_type=RelWithDebInfo
g6qsrzy     xxhash@0.8.1
</pre></details>

Trying to install a package with ROOT as a dependency failed to reuse it, so I tried manually requiring it:
```
$ spack spec -I vecgeom+root ^root/53qeuen
==> Error: Attempted to use external for 'apple-libuuid' which does not satisfy any configured external spec
```

### Error message

```console
$ spack -vd spec -I vecgeom+root ^root/53qeuen
==> [2022-08-01-06:42:01.776455] Imported spec from built-in commands
==> [2022-08-01-06:42:01.776876] Imported spec from built-in commands
==> [2022-08-01-06:42:01.777346] Reading config file /opt/spack/etc/spack/defaults/config.yaml
==> [2022-08-01-06:42:01.785312] Reading config file /opt/spack/etc/spack/config.yaml
==> [2022-08-01-06:42:01.789661] Reading config file /opt/spack/etc/spack/defaults/bootstrap.yaml
==> [2022-08-01-06:42:01.792486] DATABASE LOCK TIMEOUT: 3s
==> [2022-08-01-06:42:01.792511] PACKAGE LOCK TIMEOUT: No timeout
==> [2022-08-01-06:42:01.814553] Reading config file /opt/spack/etc/spack/defaults/config.yaml
==> [2022-08-01-06:42:01.822481] Reading config file /opt/spack/etc/spack/config.yaml
==> [2022-08-01-06:42:01.823530] DATABASE LOCK TIMEOUT: 3s
==> [2022-08-01-06:42:01.823550] PACKAGE LOCK TIMEOUT: No timeout
==> [2022-08-01-06:42:01.823609] [BOOTSTRAP CONFIG SCOPE] name=_builtin
==> [2022-08-01-06:42:01.823768] Reading config file /opt/spack/etc/spack/defaults/bootstrap.yaml
==> [2022-08-01-06:42:01.826417] [BOOTSTRAP CONFIG SCOPE] name=defaults, path=/opt/spack/etc/spack/defaults
==> [2022-08-01-06:42:01.826436] [BOOTSTRAP CONFIG SCOPE] name=defaults/darwin, path=/opt/spack/etc/spack/defaults/darwin
==> [2022-08-01-06:42:01.826457] [BOOTSTRAP CONFIG SCOPE] name=bootstrap, path=/Users/seth/.spack/bootstrap/config
==> [2022-08-01-06:42:01.826472] [BOOTSTRAP CONFIG SCOPE] name=bootstrap/darwin, path=/Users/seth/.spack/bootstrap/config/darwin
==> [2022-08-01-06:42:01.826644] Reading config file /Users/seth/.spack/bootstrap/config/darwin/compilers.yaml
==> [2022-08-01-06:42:01.858056] [BOOTSTRAP ROOT SPEC] clingo-bootstrap@spack+python %apple-clang target=aarch64
==> [2022-08-01-06:42:01.858163] [BOOTSTRAP MODULE clingo] Try importing from Python
==> [2022-08-01-06:42:01.858390] Reading config file /opt/spack/etc/spack/defaults/bootstrap.yaml
==> [2022-08-01-06:42:01.956870] Reading config file /opt/spack/etc/spack/defaults/config.yaml
==> [2022-08-01-06:42:01.983873] '/opt/spack/var/spack/environments/celeritas/.spack-env/._view/r3gorybsmmaw2ewvo4gspqpcttebk5eb/bin/python3.9' '-c' '
import json
from sysconfig import (
    get_config_vars,
    get_config_h_filename,
    get_makefile_filename,
    get_paths,
)

config = get_config_vars()
config['"'"'config_h_filename'"'"'] = get_config_h_filename()
config['"'"'makefile_filename'"'"'] = get_makefile_filename()
config.update(get_paths())

print(json.dumps(config))
'
==> [2022-08-01-06:42:02.011243] [BOOTSTRAP MODULE clingo] The installed spec "clingo-bootstrap@spack+python %apple-clang target=aarch64 ^python@3.9 /cz4h2dpuycwo4w36toajumqbjkbpqkqo" provides the "clingo" Python module
==> [2022-08-01-06:42:02.011764] Reading config file /opt/spack/etc/spack/defaults/concretizer.yaml
==> [2022-08-01-06:42:02.014042] Reading config file /opt/spack/etc/spack/defaults/repos.yaml
==> [2022-08-01-06:42:02.082447] Reading config file /opt/spack/etc/spack/defaults/config.yaml
==> [2022-08-01-06:42:02.090300] Reading config file /opt/spack/etc/spack/config.yaml
==> [2022-08-01-06:42:02.091505] Reading config file /opt/spack/etc/spack/defaults/mirrors.yaml
==> [2022-08-01-06:42:03.952062] Reading config file /opt/spack/etc/spack/compilers.yaml
==> [2022-08-01-06:42:03.957584] Reading config file /opt/spack/etc/spack/defaults/packages.yaml
==> [2022-08-01-06:42:03.970025] Reading config file /opt/spack/etc/spack/defaults/darwin/packages.yaml
==> [2022-08-01-06:42:03.974048] Reading config file /Users/seth/.spack/packages.yaml
==> [2022-08-01-06:42:09.336478] SpackError: Attempted to use external for 'apple-libuuid' which does not satisfy any configured external spec
==> [2022-08-01-06:42:09.336546] Error: Attempted to use external for 'apple-libuuid' which does not satisfy any configured external spec
  File "/opt/spack/lib/spack/spack/cmd/__init__.py", line 227, in parse_specs
    spec.concretize(tests=tests)  # implies normalize
  File "/opt/spack/lib/spack/spack/spec.py", line 2914, in concretize
    self._new_concretize(tests)
  File "/opt/spack/lib/spack/spack/spec.py", line 2889, in _new_concretize
    result = solver.solve([self], tests=tests)
  File "/opt/spack/lib/spack/spack/solver/asp.py", line 2243, in solve
    return self.driver.solve(
  File "/opt/spack/lib/spack/spack/solver/asp.py", line 676, in solve
    answers = builder.build_specs(tuples)
  File "/opt/spack/lib/spack/spack/solver/asp.py", line 2100, in build_specs
    action(*args)
  File "/opt/spack/lib/spack/spack/solver/asp.py", line 1938, in error
    raise UnsatisfiableSpecError(msg)
spack.solver.asp.UnsatisfiableSpecError: Attempted to use external for 'apple-libuuid' which does not satisfy any configured external spec

  File "/opt/spack/lib/spack/spack/main.py", line 961, in main
    return _main(argv)
  File "/opt/spack/lib/spack/spack/main.py", line 916, in _main
  File "/opt/spack/lib/spack/spack/main.py", line 944, in finish_parse_and_run
  File "/opt/spack/lib/spack/spack/main.py", line 608, in _invoke_command
  File "/opt/spack/lib/spack/spack/cmd/spec.py", line 104, in spec
    concretized_specs = spack.cmd.parse_specs(args.specs, concretize=True)
  File "/opt/spack/lib/spack/spack/cmd/__init__.py", line 241, in parse_specs
    raise spack.error.SpackError(msg)
spack.error.SpackError: Attempted to use external for 'apple-libuuid' which does not satisfy any configured external spec
```

And the external does seem to be installed:
```
$ spack find -lvd apple-libuuid
==> 1 installed package
-- darwin-monterey-m1 / apple-clang@13.1.6 ----------------------
jbbx4ar apple-libuuid@1353.100.2
```

* **Spack:** 0.19.0.dev0 (98d686278762f6667b0119a7659bf53609a045cd)
* **Python:** 3.9.13
* **Platform:** darwin-monterey-m1
* **Concretizer:** clingo

### General information

alalazo:
fixes #31484
fixes #31832

Before this change if anything was matching an external condition, it was considered "external" and thus something to be "built". This was happening in particular to external packages that were re-read from the DB, which then couldn't be reused, causing the problems shown in #31484.

This PR fixes the issue by excluding specs with a "hash" from being considered "external"

### DIFF
--- a/lib/spack/docs/build_settings.rst
+++ b/lib/spack/docs/build_settings.rst
@@ -131,7 +131,7 @@ OpenMPI.
    If ``concretizer:reuse`` is on (see :ref:`concretizer-options` for more information on that flag)
    pre-built specs include specs already available from a local store, an upstream store, a registered
    buildcache or specs marked as externals in ``packages.yaml``. If ``concretizer:reuse`` is off, only
-   external specs in ``packages.yaml`` are include d in the list of pre-built specs.
+   external specs in ``packages.yaml`` are included in the list of pre-built specs.
 
 If an external module is specified as not buildable, then Spack will load the
 external module into the build environment which can be used for linking.
--- a/lib/spack/docs/build_settings.rst
+++ b/lib/spack/docs/build_settings.rst
@@ -49,9 +49,8 @@ packages rather than building its own packages. This may be desirable
 if machines ship with system packages, such as a customized MPI
 that should be used instead of Spack building its own MPI.
 
-External packages are configured through the ``packages.yaml`` file found
-in a Spack installation's ``etc/spack/`` or a user's ``~/.spack/``
-directory. Here's an example of an external configuration:
+External packages are configured through the ``packages.yaml`` file.
+Here's an example of an external configuration:
 
 .. code-block:: yaml
 
@@ -97,11 +96,14 @@ Each package version and compiler listed in an external should
 have entries in Spack's packages and compiler configuration, even
 though the package and compiler may not ever be built.
 
-The packages configuration can tell Spack to use an external location
-for certain package versions, but it does not restrict Spack to using
-external packages.  In the above example, since newer versions of OpenMPI
-are available, Spack will choose to start building and linking with the
-latest version rather than continue using the pre-installed OpenMPI versions.
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Prevent packages from being built from sources
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Adding an external spec in ``packages.yaml`` allows Spack to use an external location,
+but it does not prevent Spack from building packages from sources. In the above example,
+Spack might choose for many valid reasons to start building and linking with the
+latest version of OpenMPI rather than continue using the pre-installed OpenMPI versions.
 
 To prevent this, the ``packages.yaml`` configuration also allows packages
 to be flagged as non-buildable.  The previous example could be modified to
@@ -121,9 +123,15 @@ be:
        buildable: False
 
 The addition of the ``buildable`` flag tells Spack that it should never build
-its own version of OpenMPI, and it will instead always rely on a pre-built
-OpenMPI.  Similar to ``paths``, ``buildable`` is specified as a property under
-a package name.
+its own version of OpenMPI from sources, and it will instead always rely on a pre-built
+OpenMPI.
+
+.. note::
+
+   If ``concretizer:reuse`` is on (see :ref:`concretizer-options` for more information on that flag)
+   pre-built specs include specs already available from a local store, an upstream store, a registered
+   buildcache or specs marked as externals in ``packages.yaml``. If ``concretizer:reuse`` is off, only
+   external specs in ``packages.yaml`` are include d in the list of pre-built specs.
 
 If an external module is specified as not buildable, then Spack will load the
 external module into the build environment which can be used for linking.
@@ -131,6 +139,10 @@ external module into the build environment which can be used for linking.
 The ``buildable`` does not need to be paired with external packages.
 It could also be used alone to forbid packages that may be
 buggy or otherwise undesirable.
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Non-buildable virtual packages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Virtual packages in Spack can also be specified as not buildable, and
 external implementations can be provided. In the example above,
@@ -153,21 +165,37 @@ but more conveniently:
        - spec: "openmpi@1.6.5%intel@10.1 arch=linux-debian7-x86_64"
          prefix: /opt/openmpi-1.6.5-intel
 
-Implementations can also be listed immediately under the virtual they provide:
+Spack can then use any of the listed external implementations of MPI
+to satisfy a dependency, and will choose depending on the compiler and
+architecture.
+
+In cases where the concretizer is configured to reuse specs, and other ``mpi`` providers
+(available via stores or buildcaches) are not wanted, Spack can be configured to require
+specs matching only the available externals:
 
 .. code-block:: yaml
 
    packages:
      mpi:
        buildable: False
-         openmpi@1.4.3%gcc@4.4.7 arch=linux-debian7-x86_64: /opt/openmpi-1.4.3
-         openmpi@1.4.3%gcc@4.4.7 arch=linux-debian7-x86_64+debug: /opt/openmpi-1.4.3-debug
-         openmpi@1.6.5%intel@10.1 arch=linux-debian7-x86_64: /opt/openmpi-1.6.5-intel
-         mpich@3.3 %clang@9.0.0 arch=linux-debian7-x86_64: /opt/mpich-3.3-intel
+       require:
+       - one_of: [
+           "openmpi@1.4.3%gcc@4.4.7 arch=linux-debian7-x86_64",
+           "openmpi@1.4.3%gcc@4.4.7 arch=linux-debian7-x86_64+debug",
+           "openmpi@1.6.5%intel@10.1 arch=linux-debian7-x86_64"
+         ]
+     openmpi:
+       externals:
+       - spec: "openmpi@1.4.3%gcc@4.4.7 arch=linux-debian7-x86_64"
+         prefix: /opt/openmpi-1.4.3
+       - spec: "openmpi@1.4.3%gcc@4.4.7 arch=linux-debian7-x86_64+debug"
+         prefix: /opt/openmpi-1.4.3-debug
+       - spec: "openmpi@1.6.5%intel@10.1 arch=linux-debian7-x86_64"
+         prefix: /opt/openmpi-1.6.5-intel
 
-Spack can then use any of the listed external implementations of MPI
-to satisfy a dependency, and will choose depending on the compiler and
-architecture.
+This configuration prevents any spec using MPI and originating from stores or buildcaches to be reused,
+unless it matches the requirements under ``packages:mpi:require``. For more information on requirements see
+:ref:`package-requirements`.
 
 .. _cmd-spack-external-find:
 
@@ -194,11 +222,6 @@ Specific limitations include:
 * Packages are not discoverable by default: For a package to be
   discoverable with ``spack external find``, it needs to add special
   logic. See :ref:`here <make-package-findable>` for more details.
-* The current implementation only collects and examines executable files,
-  so it is typically only useful for build/run dependencies (in some cases
-  if a library package also provides an executable, it may be possible to
-  extract a meaningful Spec by running the executable - for example the
-  compiler wrappers in MPI implementations).
 * The logic does not search through module files, it can only detect
   packages with executables defined in ``PATH``; you can help Spack locate
   externals which use module files by loading any associated modules for
--- a/lib/spack/spack/solver/concretize.lp
+++ b/lib/spack/spack/solver/concretize.lp
@@ -464,7 +464,9 @@ version_weight(Package, Weight) :- external_version(Package, Version, Weight).
 version(Package, Version) :- external_version(Package, Version, Weight).
 
 % if a package is not buildable (external_only), only externals are allowed
-external(Package) :- external_only(Package), node(Package).
+external(Package) :- external_only(Package),
+                     node(Package),
+                     not hash(Package, _).
 
 % a package is a real_node if it is not external
 real_node(Package) :- node(Package), not external(Package).
@@ -483,7 +485,8 @@ external(Package) :- external_spec_selected(Package, _).
 % determine if an external spec has been selected
 external_spec_selected(Package, LocalIndex) :-
     external_conditions_hold(Package, LocalIndex),
-    node(Package).
+    node(Package),
+    not hash(Package, _).
 
 external_conditions_hold(Package, LocalIndex) :-
     possible_external(ID, Package, LocalIndex), condition_holds(ID).
--- a/lib/spack/spack/solver/concretize.lp
+++ b/lib/spack/spack/solver/concretize.lp
@@ -276,7 +276,8 @@ error(0, Msg) :- node(Package),
    conflict(Package, TriggerID, ConstraintID, Msg),
    condition_holds(TriggerID),
    condition_holds(ConstraintID),
-   not external(Package).  % ignore conflicts for externals
+   not external(Package),  % ignore conflicts for externals
+   not hash(Package, _).  % ignore conflicts for installed packages
 
 #defined conflict/4.
 
--- a/lib/spack/spack/solver/concretize.lp
+++ b/lib/spack/spack/solver/concretize.lp
@@ -437,7 +437,7 @@ attr("node_compiler_version_satisfies", Package, Compiler, Version)
 #defined external/1.
 #defined external_spec/2.
 #defined external_version_declared/4.
-#defined external_only/1.
+#defined buildable_false/1.
 #defined pkg_provider_preference/4.
 #defined default_provider_preference/3.
 #defined node_version_satisfies/2.
@@ -464,8 +464,8 @@ error(2, "Attempted to use external for '{0}' which does not satisfy any configu
 version_weight(Package, Weight) :- external_version(Package, Version, Weight).
 version(Package, Version) :- external_version(Package, Version, Weight).
 
-% if a package is not buildable (external_only), only externals are allowed
-external(Package) :- external_only(Package),
+% if a package is not buildable, only externals or hashed specs are allowed
+external(Package) :- buildable_false(Package),
                      node(Package),
                      not hash(Package, _).
 
--- a/lib/spack/spack/test/concretize.py
+++ b/lib/spack/spack/test/concretize.py
@@ -1826,13 +1826,7 @@ class TestConcretize(object):
             mpi_spec = Spec("mpi").concretized()
             assert mpi_spec.name != "multi-provider-mpi"
 
-        external_conf.update(
-            {
-                "mpich": {"require": "%xl"},
-                "mpich2": {"require": "%xl"},
-                "zmpi": {"require": "%xl"},
-            }
-        )
+        external_conf["mpi"]["require"] = "multi-provider-mpi"
         spack.config.set("packages", external_conf)
 
         with spack.config.override("concretizer:reuse", True):
--- a/lib/spack/spack/test/concretize.py
+++ b/lib/spack/spack/test/concretize.py
@@ -1768,40 +1768,73 @@ class TestConcretize(object):
         ):
             s.concretized()
 
-    @pytest.mark.regression('31484')
-    def test_installed_externals_are_reused(
-            self, mutable_database, repo_with_changing_recipe
-    ):
+    @pytest.mark.regression("31484")
+    def test_installed_externals_are_reused(self, mutable_database, repo_with_changing_recipe):
         """Test that external specs that are in the DB can be reused."""
-        if spack.config.get('config:concretizer') == 'original':
-            pytest.xfail('Use case not supported by the original concretizer')
+        if spack.config.get("config:concretizer") == "original":
+            pytest.xfail("Use case not supported by the original concretizer")
 
         # Configuration to be added to packages.yaml
         external_conf = {
-            'changing': {
-                'buildable': False,
-                'externals': [{
-                    'spec': 'changing@1.0',
-                    'prefix': '/usr'
-                }]
+            "changing": {
+                "buildable": False,
+                "externals": [{"spec": "changing@1.0", "prefix": "/usr"}],
             }
         }
-        spack.config.set('packages', external_conf)
+        spack.config.set("packages", external_conf)
 
         # Install the external spec
-        external1 = Spec('changing@1.0').concretized()
+        external1 = Spec("changing@1.0").concretized()
         external1.package.do_install(fake=True, explicit=True)
         assert external1.external
 
         # Modify the package.py file
-        repo_with_changing_recipe.change({'delete_variant': True})
+        repo_with_changing_recipe.change({"delete_variant": True})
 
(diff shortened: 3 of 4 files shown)


lib/spack/spack/solver/concretize.lp
```diff
@@ -459,7 +459,9 @@ version_weight(Package, Weight) :- external_version(Package, Version, Weight).
 version(Package, Version) :- external_version(Package, Version, Weight).
 
 % if a package is not buildable (external_only), only externals are allowed
-external(Package) :- external_only(Package), node(Package).
+external(Package) :- external_only(Package),
```
becker33: I think this will allow you to reuse a non-external even when `buildable: false` for that package. Can you add a test for this?

becker33 (changes requested):

lib/spack/spack/solver/concretize.lp
alalazo: I don't think this is wrong. It will allow what you say only if the non-external is already installed, so not buildable, but built. Disallowing that case causes the bug in this way:
1. Install an external that is not buildable
2. Can't reuse the same external at the next concretization

adamjstewart (dismissed):
Works like a charm, thanks for the quick fix!

Also tested that the tests do indeed fail without this change.

lib/spack/spack/test/concretize.py
```diff
@@ -1836,3 +1836,33 @@ def test_installed_externals_are_reused(
         with spack.config.override("concretizer:reuse", True):
             external3 = Spec('changing@1.0').concretized()
         assert external3.dag_hash() == external1.dag_hash()
+
+    @pytest.mark.regression('31484')
+    def test_installed_builds_not_reused_buildable_false(
+            self, mutable_database, repo_with_changing_recipe
+    ):
```
alalazo: @becker33 I have been getting back at this specific test, and I think this is behavior we don't want to add as it is coded here. The only things we  currently say for `buildable: false` in our documentation are:

> The addition of the buildable flag tells Spack that it should never build its own version of OpenMPI, and it will instead always rely on a pre-built OpenMPI.
>
> The buildable does not need to be paired with external packages. It could also be used alone to forbid packages that may be buggy or otherwise undesirable.
> [ ... ]

and they clearly stem from a world where `reuse` was not a considered at all. It seems thus a stretch to me to require that `buildable: false` prevents the reuse of software that is either installed locally or available from a buildcache.

That said, I concur, based on our discussion on Slack, that there's a gap in expressiveness that we need to fill. For instance, as you pointed out:
```yaml
packages:
  mpi:
    buildable: false
```
likely means that the user wants to install only from one of the externals that might be listed in the same file. But what if it is:
```yaml
packages:
  trilinos:
    buildable: false
```
Couldn't that be a way to say "Reuse a binary version of trilinos and never build that from sources"? I wouldn't find that request "crazy", if a binary cache is available. 

I think the point is: `buildable: false` was added before `reuse` and we didn't figure out how the two should interact together. That is possibly what needs to get fixed (allow the user to express in which of the two cases they are), not excluding specs from reuse based on `buildable: false`.

**TL;DR** I agree that here we miss ways to express a few requests, but would it be possible to get the bugfix in as is, and work on a more extensive feature in a following PR? lf not I'm tempted to move this out of the v0.18.1 project, since it likely means it won't be solved soon.

becker33:
I still don't see how someone using this is supposed to use system mpi, and I don't think that's a use case we can afford to break.

lib/spack/spack/test/concretize.py
alalazo: > I still don't see how someone using this is supposed to use system mpi, and I don't think that's a use case we can afford to break.

@becker33 Fair, I already moved this to the next bugfix release. Do you agree though that the test, as is, is not something we want to force implicitly?

sethrj:
@alalazo I think this will also fix #31832 . Interestingly my `libuuid` external is a build-only dependency of the spec that's failing... so it doesn't appear in `spack find -lvd`. However I guess it still gets used by the concretizer?

adamjstewart:
@becker33 @alalazo any updates on this? Spack is basically unusable on macOS if you update develop frequently. Not sure if there's a priority level higher than "impact-high" lol. I've been using this PR locally on a couple systems since this PR was first opened and I would like to no longer have to use that local modification.

alalazo:
@adamjstewart There is currently an issue, that @becker33 pointed out, that is hard to solve in this PR alone. My thoughts on this are:
1. The second test added is imho not what we want from the perspective of the semantics
2. I am trying to push forward the review of #27987 to have hard requirements in `packages.yaml`
3. When #27987 is merged, we can remove the second test here (maybe add another one to ensure we don't have regression on the issue with `mpi:buildable:false`) and merge this

Rationale is that, if we have requirements, we can specify there what needs to be used.

sethrj:
@alalazo @adamjstewart @becker33 is there a quick fix for getting this in for macos platforms?

alalazo:
@sethrj https://github.com/spack/spack/pull/27987 will be merged today or tomorrow. With that in I plan to:
1. Rebase this PR
2. Substitute the last test with a similar one that uses requirements
3. :crossed_fingers: that everything is :heavy_check_mark: 
4. Merge the PR asap

sethrj:
Amazing! Thanks @alalazo !

adamjstewart:
Is the test the only thing that was preventing this PR from being merged? I was under the impression that there was a problem with the solution itself.

alalazo:
@adamjstewart The issue that @becker33 was raising is that we don't have a way to discriminate between an external and a binary package when `reuse:true`. This bug (which gives an inconsistent internal solution in ASP) don't allow _anything_ matching an external to be reused, so a user can say:
```yaml
mpi:
  buildable: false
packages:
  openmpi:
    externals:
    - spec: "openmpi@4.1.4"
```
and be sure it is using the external spec. I think with #27987 we can use hard-requirements to pin the required `virtual` into specs.

alalazo:
I added a new test. Currently users can select an external, by imposing unsatisfiable requirements on other providers that could otherwise be reused. This is not ideal, but works. If people agree (@adamjstewart @sethrj @becker33) I would merge the bug fix like it is for the time being, and then start working on extending the `packages:<pkg>:require` attribute to virtual packages. Ideally I'd like to have:
```yaml
packages:
  mpi:
    buildable: false
    require: 'openmpi@3.0.1'
  openmpi:
    externals:
    - spec: 'openmpi@3.0.1'
      prefix: /usr
```
applied whenever `mpi` is in the DAG as a virtual spec. 

EDIT: Pinging @scheibelp as he may want to comment on the extension sketched above.

alalazo:
@becker33 Would https://github.com/spack/spack/pull/31558#issuecomment-1217867075 be fine with you?

psakievich:
@becker33 any updates on this?

alalazo:
For the record: I submitted https://github.com/spack/spack/pull/32369 today so I hope we'll converge to the syntax in https://github.com/spack/spack/pull/31558#issuecomment-1217867075 pretty soon (no matter whether this PR or #32369 gets merged first).

alalazo:
#32369 has just been merged, and the second unit test updated accordingly

lib/spack/spack/solver/concretize.lp
```diff
@@ -464,7 +464,9 @@ version_weight(Package, Weight) :- external_version(Package, Version, Weight).
...
```
becker33: I still think we shouldn't be lying to Spack about whether it's an external. If it's a previously installed external, we should be marking that appropriately, not pretending it's not an external.

lib/spack/spack/test/concretize.py
```diff
@@ -1767,3 +1767,68 @@ def test_git_ref_version_errors_if_unknown_version(self, git_ref):
...
+            external2 = Spec("changing@1.0").concretized()
+        assert external2.dag_hash() != external1.dag_hash()
+
+        # ... while with reuse we have the same hash
+        with spack.config.override("concretizer:reuse", True):
+            external3 = Spec("changing@1.0").concretized()
+        assert external3.dag_hash() == external1.dag_hash()
+
+    @pytest.mark.regression("31484")
+    def test_user_can_select_externals_with_require(self, mutable_database):
```
becker33, line 1805: If I understand correctly, this only ensures we can force the provider, not the specific spec.

```
$ spack install mvapich2@2.3.7
$ spack config add packages:mvapich2:buildable:false
$ spack config edit  # add external for mvapich2@2.3.7
$ spack install mvapich2
```
would reuse the `mvapich2@2.3.7` installed on the first line, and there is no recourse to force it to use the external as far as I understand.

alalazo: True. Then you can call your external `mvapich2@2.3.7.special`. I hope this will be a niche case

lib/spack/spack/solver/concretize.lp
alalazo: Imo the visible effect to users is consistent. The property is called `buildable` (not `external_only`), and the user can require constraints to be sure clingo selects a spec matching the external.

alalazo:
@becker33 The bug occurring when:
1. A spec is installed
2. Later, a conflict is added for which the spec at 1 is a match
3. The spec installed at 1 cannot be reused

has been fixed in 30e8edc, and a unit test has been included to avoid regressions. 

I checked the rules in which `external(Package)` is involved and it seems to me they all come from before we introduced `reuse`. This is to say that external in `concretize.lp` means "defined as external in packages.yaml" since the `external(Package)` atom was introduced. One example for that is for instance:

https://github.com/spack/spack/blob/1c22af8ef4947d920b4d86e3571acdc2cdc8e377/lib/spack/spack/solver/concretize.lp#L450-L461

If a package is `external` we can only choose versions that are declared in `packages.yaml`. Not accounting for this, caused  the bug in #31484[^1]  and is fixed here. In 27adf2a I renamed an internal atom to adhere better to where it originates. In e569ba0 I updated docs to add some details on `buildable: false` and its behavior. Let me know how these changes looks.[^2]


[^1]: What gets installed is "frozen" and can get out of sync with a changing `packages.yaml`
[^2]: In the long term I hope we can get rid of most of this custom logic for externals, and that Spack can automatically detect packages and pull them into the local store in a reserved namespace (e.g. `spack.pkg.externals`), without the need for users to declare them manually. I don't think though implementing ~all that~ anything more that is in this PR is necessary to fix this particular bug. Instead, I think this PR can solve a high-priority bug that has been open for a month and a half and give us time to consider how to design a better support for externals that blends with other requests to better tune what can be reused.

lib/spack/docs/build_settings.rst
```diff
@@ -121,9 +123,15 @@ be:
...
-a package name.
+its own version of OpenMPI from sources, and it will instead always rely on a pre-built
+OpenMPI.
+
+.. note::
+
+   If ``concretizer:reuse`` is on (see :ref:`concretizer-options` for more information on that flag)
+   pre-built specs include specs already available from a local store, an upstream store, a registered
+   buildcache or specs marked as externals in ``packages.yaml``. If ``concretizer:reuse`` is off, only
+   external specs in ``packages.yaml`` are include d in the list of pre-built specs.
```
becker33:
```suggestion
   external specs in ``packages.yaml`` are included in the list of pre-built specs.
```

becker33 (changes requested):

becker33 (approved):

adamjstewart (approved):

adamjstewart:
Thanks for the hard work on this @alalazo and @becker33! It's unfortunate that it ended up being more involved than expected, but glad to see this bug finally fixed!

becker33:
Give @alalazo all the credit, I just complained about solutions and he came up with new ones

adamjstewart:
Playing Devil's advocate and finding ways that a solution might break is just as important 😄

PR #31561, opened 2022-07-13
Make environment writing update the config scope

psakievich:
Updating the `spack.yaml` through python does not propagate changes that are affected by the config scope such as `Environment::unify`.  This PR addresses this issue


psakievich:
I'll revisit this later

PR #31565, opened 2022-07-13
interproscan: new version 5.56-89.0

snehring:
Version update. Also seems to resolve some build errors I was seeing.


tldahlgren (approved):
Confirmed sha256.

PR #31569, opened 2022-07-13
deprecate `blacklist`/`whitelist` in favor of `include`/`exclude`

tgamblin:
For a long time the module configuration has had a few settings that use `blacklist`/`whitelist` terminology. We've been asked by some of our users to replace this with more inclusive language. In addition to being non-inclusive, `blacklist` and `whitelist` are inconsistent with the rest of Spack, which uses `include` and `exclude` for the same concepts.

- [x] Deprecate `blacklist`, `whitelist`, `blacklist_implicits` and `environment_blacklist` in favor of `exclude`, `include`, `exclude_implicits` and `exclude_env_vars` in module configuration, to be removed in Spack v0.20.
- [x] Print deprecation warnings if any of the deprecated names are in module config.
- [x] Update tests to test old and new names.
- [x] Update docs.
- [x] Update `spack config update` to fix this automatically, and include a note in the error that you can use this command.

@hnamLANL FYI -- sorry this took so long


lib/spack/spack/test/cmd/module.py
```diff
@@ -149,24 +149,26 @@ def test_find_recursive():
 
 
 @pytest.mark.db
-def test_find_recursive_blacklisted(database, module_configuration):
-    module_configuration('blacklist')
+@pytest.mark.parametrize("config_name", ["exclude", "blacklist"])
```
tldahlgren, line 153: Is it worth adding a note here and below about the deprecation version?

lib/spack/docs/module_file_support.rst
```diff
@@ -589,7 +589,7 @@ Filter out environment modifications
 Modifications to certain environment variables in module files are there by
 default, for instance because they are generated by prefix inspections.
 If you want to prevent modifications to some environment variables, you can
-do so by using the environment blacklist:
+do so by using the ``exclude_environment``:
```
tldahlgren: Is this word ordering to be consistent with `exclude_implicits` (among others)?

lib/spack/spack/test/modules/lmod.py
```diff
@@ -110,10 +110,15 @@ def test_autoload_all(self, modulefile_content, module_configuration):
 
         assert len([x for x in content if 'depends_on(' in x]) == 5
 
-    def test_alter_environment(self, modulefile_content, module_configuration):
+    @pytest.mark.parametrize(
```
tldahlgren, line 114: Is it worth noting here and below the deprecation version?

lib/spack/spack/test/modules/tcl.py
```diff
@@ -97,10 +97,15 @@ def test_prerequisites_all(self, modulefile_content, module_configuration):
 
         assert len([x for x in content if 'prereq' in x]) == 5
 
-    def test_alter_environment(self, modulefile_content, module_configuration):
+    @pytest.mark.parametrize(
```
tldahlgren, line 101: Another file where it might be worth commenting on the deprecation version.

tldahlgren (commented):
Some preliminary feedback/questions.

lib/spack/docs/module_file_support.rst
tgamblin: Yes, and it reads better.

tgamblin:
Added notes on the deprecation version to all the `@pytest.mark.parametrize` decorators

lib/spack/spack/schema/modules.py
```diff
@@ -224,6 +245,41 @@ def deprecation_msg_default_module_set(instance, props):
...
+# deprecated keys and their replacements
+exclude_include_translations = {
+    "whitelist": "include",
+    "blacklist": "exclude",
+    "blacklist_implicits": "exclude_implicits",
+    "environment_blacklist": "exclude_environment",
+}
+
+
+def update_keys(data, key_translations, changed=False):
```
becker33: This `changed` argument looks unnecessary.

lib/spack/spack/schema/modules.py
```diff
@@ -251,10 +307,14 @@ def update(data):
     # Move deprecated top-level keys under "default" module set.
     for key in deprecated_top_level_keys:
         if key in data:
+            print(data)
```
becker33: This looks vestigial

becker33 (changes requested):

lib/spack/spack/schema/modules.py
tgamblin: why?

lib/spack/spack/schema/modules.py
tgamblin: fixed

lib/spack/spack/schema/modules.py
becker33: Because it is only used as a return value, not as an input. Everything that modifies it returns the value, and is `|=` into the final return. 

The sub-dict doesn't need to know about whether its parent changed to know what to return. If the parent changed, the parent flips the return code True, and if the child changed, it flips it.

tgamblin: Ah ok -- true -- I'll get rid of the arg

tgamblin: fixed

becker33 (dismissed):

lib/spack/spack/schema/modules.py
```diff
@@ -257,4 +320,7 @@ def update(data):
         changed = True
         data['default'] = default
 
+    # translate exclude/include to blacklist/whitelist
+    changed |= update_keys(data, exclude_include_translations)
```
alalazo, line 334: If we hit the early exit at line 310, we'll never reach this update.

lib/spack/spack/schema/modules.py
```diff
@@ -257,4 +320,7 @@ def update(data):
         changed = True
         data['default'] = default
 
+    # translate exclude/include to blacklist/whitelist
```
alalazo: I think this comment should be backward:
```suggestion
    # translate blacklist/whitelist to exclude/include
```
right?

alalazo (dismissed):
Only two minor comments. The one on the early exit is probably an edge case.

lib/spack/docs/module_file_support.rst
```diff
@@ -599,7 +599,7 @@ do so by using the environment blacklist:
          all:
            filter:
              # Exclude changes to any of these variables
-             environment_blacklist: ['CPATH', 'LIBRARY_PATH']
+             exclude_environment: ['CPATH', 'LIBRARY_PATH']
```
haampie: or `exclude_variables`?

tgamblin: I like it better -- would `exclude_env_vars` be even more clear?

lib/spack/spack/schema/modules.py
tgamblin: fixed

lib/spack/spack/schema/modules.py
tgamblin: fixed

lib/spack/docs/module_file_support.rst
tgamblin: Committed again with `exclude_env_vars`

lib/spack/spack/schema/modules.py
tgamblin: Making this work revealed that the `module_configuration` fixture needs to use `mutable_config`.

alalazo (approved):

PR #31314, opened 2022-06-27
Add libfabric versions 1.15.0 and 1.15.1

teojgo:

PR #31572, opened 2022-07-14
libfabric: add versions 1.15.0 and 1.15.1

snehring:
I was working on this earlier today to get opx support and then saw #31314 before submitting and that it seemed stalled. 

This is more or less identical but it adds a missing dependency (uuid),  marks another fabric removed (mlx), and should solve the gitlab ci error.

### DIFF
--- a/var/spack/repos/builtin/packages/libfabric/package.py
+++ b/var/spack/repos/builtin/packages/libfabric/package.py
@@ -17,6 +17,8 @@ class Libfabric(AutotoolsPackage):
     maintainers = ["rajachan"]
 
     version("master", branch="master")
+    version('1.15.1', sha256='cafa3005a9dc86064de179b0af4798ad30b46b2f862fe0268db03d13943e10cd')
+    version('1.15.0', sha256='70982c58eadeeb5b1ddb28413fd645e40b206618b56fbb2b18ab1e7f607c9bea')
     version("1.14.1", sha256="6cfabb94bca8e419d9015212506f5a367d077c5b11e94b9f57997ec6ca3d8aed")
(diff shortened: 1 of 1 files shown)


rajachan (dismissed):
Looks good, thanks.

var/spack/repos/builtin/packages/libfabric/package.py
```diff
@@ -117,7 +122,7 @@ def configure_args(self):
         else:
             args.append('--with-kdreg=no')
 
-        for fabric in self.fabrics:
+        for fabric in self.spec.variants['fabrics'].value:
```
haampie: this is not correct, the idea is to explicitly disable fabrics that were not selected, but you are looping over those that were selected only.

snehring: True, whoops. How would you like to handle this? Make all the elements in fabrics conditional? Or handle _ConditionalVariantValues separately in that loop? Make fabrics a list and append when conditions are met instead of using conditional?

snehring:
There's probably a less ham-fisted way of dealing with this particular use of conditional variants that I'm just not aware of.

snehring:
@haampie are there any changes you'd like made to this?

snehring:
Rebased to current develop to fix conflicts due to the changes from black.

snehring:
@haampie is there anything you'd like me to do to this in order to get it merged?

snehring:
Hmm actually, looking at this again, it still doesn't explicitly disable a conditional fabric if it's still relevant. Removed fabrics only generate warnings if they're specified so we can probably just ignore that for the sake of simplicity.

snehring:
ci failures look transient. @spackbot run pipeline

snehring:
superseded by #32083

PR #26809, opened 2021-10-18
Libtiff: improve compression support

adamjstewart:
Fixes #26798 

I fell pretty far down the rabbit hole on this one. GDAL allows you to compress files. From what I can tell, this is on a driver-by-driver basis, and support for this involves compiling the driver with these specific (de)compression libraries. The only reason that GDAL has flags for these dependencies is to allow libtiff to be compiled during the build of GDAL. If you instead build against an external libtiff, then these flags aren't used.

In this PR I added explicit variants for all external compression libraries that libtiff supports. All variants are off by default to minimize build times, at least until the new concretizer gets better at reusing already-installed deps. I also removed the `zstd` variant from GDAL since it doesn't seem to do anything. There may be many more variants that can be removed but I'm not sure which. Maybe @rouault would know which dependencies are GDAL deps and which are deps of deps?

Tested by compiling GDAL with `libtiff+lerc+zstd` and using `gdal_translate` with both `-co COMPRESS=LERC` and `-co COMPRESS=ZSTD` to compress a Landsat tile. Both work great now! Thanks to @steo85it for reporting this.

PR #31583, opened 2022-07-14
GDAL: external LERC only available in 3.3+

adamjstewart:
User reported a build issue when linking to external LERC for older GDAL. According to https://gdal.org/drivers/raster/gtiff.html:

> `LERC` and `LERC_DEFLATE` are available only when using internal libtiff for GDAL < 3.3.0. Since GDAL 3.3.0, LERC compression is also available when building GDAL against external libtiff >= 4.3.0, built itself against https://github.com/esri/lerc

Actually, now that I look at this, I'm not sure if it even makes sense to have a `+lerc` variant/dependency. @rouault is LERC actually used in GDAL or is it only used when building the internal copy of libtiff? Since we don't use internal copies of any dependencies, we may be able to remove the `+lerc` variant. Are there any other dependencies listed at https://gdal.org/build_hints.html that are secretly dependencies of dependencies?

### DIFF
--- a/var/spack/repos/builtin/packages/gdal/package.py
+++ b/var/spack/repos/builtin/packages/gdal/package.py
@@ -112,7 +112,7 @@ class Gdal(CMakePackage):
     variant('jxl', default=False, when='@3.4:', description='Required for JPEGXL driver')
     variant('kdu', default=False, description='Required for JP2KAK and JPIPKAK drivers')
     variant('kea', default=False, description='Required for KEA driver')
-    variant('lerc', default=True, when='@2.4:', description='Required for LERC compression')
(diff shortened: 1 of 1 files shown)


rouault:
> @rouault is LERC actually used in GDAL or is it only used when building the internal copy of libtiff?

the story depends on versions. LERC can be used by the MRF driver too. I'm not sure if it is from 3.5.0 or master, but previously only the internal copy of libLERC could be used by the MRF driver. Now it can be the external copy too.

adamjstewart:
Looks like that may also be the case with zstd: https://github.com/spack/spack/pull/26809#issuecomment-946486379

I would love to know which dependencies are actually used in GDAL and from which versions those deps are supported externally.

PR #31587, opened 2022-07-14
libfabric: Correct develop version's GitHub branch name.

hyoklee:
master doesn't exist.

### DIFF
--- a/var/spack/repos/builtin/packages/libfabric/package.py
+++ b/var/spack/repos/builtin/packages/libfabric/package.py
@@ -16,7 +16,7 @@ class Libfabric(AutotoolsPackage):
 
     maintainers = ['rajachan']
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/libfabric/package.py
```diff
@@ -16,7 +16,7 @@ class Libfabric(AutotoolsPackage):
 
     maintainers = ['rajachan']
 
-    version('master', branch='master')
+    version('develop', branch='main')
```
adamjstewart, line 19: Can we call it main instead?

rajachan: +1, but this has already been taken care of in https://github.com/spack/spack/pull/32083/

Issue #31499, opened 2022-07-08
Cannot extract archive, unrecognized file extension: 'None'


haampie:
### Steps to reproduce

On linux,

```
$ spack -d stage eccodes@2.25.0
==> [2022-07-08-17:43:03.348355] Staging archive: /tmp/harmen/spack-stage/spack-stage-eccodes-2.25.0-svaf535zwdloqoikjuskycdctfqnns7t/eccodes-2.25.0-Source.tar.gz?api=v2
==> [2022-07-08-17:43:03.348504] CommandNotFoundError: Cannot extract archive, unrecognized file extension: 'None'
==> [2022-07-08-17:43:03.348532] Error: Cannot extract archive, unrecognized file extension: 'None'
Traceback (most recent call last):
  File "/home/harmen/spack/lib/spack/spack/main.py", line 893, in main
    return _main(argv)
  File "/home/harmen/spack/lib/spack/spack/main.py", line 848, in _main
    return finish_parse_and_run(parser, cmd_name, env_format_error)
  File "/home/harmen/spack/lib/spack/spack/main.py", line 876, in finish_parse_and_run
    return _invoke_command(command, parser, args, unknown)
  File "/home/harmen/spack/lib/spack/spack/main.py", line 533, in _invoke_command
    return_val = command(parser, args)
  File "/home/harmen/spack/lib/spack/spack/cmd/stage.py", line 64, in stage
    package.do_stage()
  File "/home/harmen/spack/lib/spack/spack/package_base.py", line 1577, in do_stage
    self.stage.expand_archive()
  File "/home/harmen/spack/lib/spack/spack/util/pattern.py", line 18, in __call__
    return [getattr(item, self.name)(*args, **kwargs)
  File "/home/harmen/spack/lib/spack/spack/util/pattern.py", line 18, in <listcomp>
    return [getattr(item, self.name)(*args, **kwargs)
  File "/home/harmen/spack/lib/spack/spack/stage.py", line 610, in expand_archive
    self.fetcher.expand()
  File "/home/harmen/spack/lib/spack/spack/fetch_strategy.py", line 84, in wrapper
    return fun(self, *args, **kwargs)
  File "/home/harmen/spack/lib/spack/spack/fetch_strategy.py", line 539, in expand
    decompress = decompressor_for(self.archive_file, self.extension)
  File "/home/harmen/spack/lib/spack/spack/util/compression.py", line 258, in decompressor_for
[...]

PR #27021, opened 2021-10-28
Spack Windows support

johnwparent:
This branch adds basic Windows support for Spack.

This branch includes:

- Basic Windows support for most Spack components
- Getting started on Windows docs
- Spack scripts and shell support for Windows CMD and PWSH
- An updated Actions script to include Windows CI jobs
- Introduction of the MSVC compiler and Windows platform/OS
- Updated path searching and path detection for Windows
- Windows specific support for non Windows compatible unix operations performed by Spack (file system, locking, symlinks, etc)
- Updates to CMake build system
- Updated a number of packages to use Windows compliant build systems (moved away from Autotools when possible)
- Added Autotools conflicts with Windows
- Windows MSI Installer support and command line option
- CI generates a 'Spack Installer' for Windows as a build artifact, can be distributed to "just install" Spack on Windows 
- Test compatibility with Windows

PR #31141, opened 2022-06-15
Hdf5 For Windows V2

jpopelar:
Remake of PR #29826 due to branch conflicts and duplicated commits

PR #31589, opened 2022-07-14
Refactor archive ext parsing

johnwparent:
At the moment, Spack relies on string parsing to determine compression/archive file extension type when fetching/staging packages. Prior to #27021, if Spack was unable to successfully extract a Spack recognized archive/compression extension, it simply tried to use the `tar` system utility. This was generally successful, however with Windows support, a stricter requirement was needed, specifying that Spack be able to successfully parse the extension or raise and exception.  As has been pointed out by #31499 and a number of other issues, this can at times be fragile and occasionally breaks Spack installs. 

This PR resolves this issue in two ways:

- Improvements to the string parsing (minor)
- Magic number parsing

The magic number parsing is the focal point of this PR. Spack will, if file is on filesystem, read magic bytes from binary stream of file, determining filetype/extension from this information. For compressed file types i.e. `.gz` Spack will decompress and peek at the first bytes in order to determine if the underlying file is an archive type i.e. `.tar`. If Spack is unable to decompress the file in memory (if python version does not include LZMA support for example, or the compression is `LZW` or `LZH`), it falls back on simple filepath parsing to check if the file being compressed is an extractable archive.

This behavior is similar to the `file` command, or `libmagic` but is written in cross platform Python, and thus can be used on Windows. Other pure Python implementations of similar functionalities exist, but either have significant drawbacks that would limit their usage in Spack, are overkill, or produce results that are not consistent with what we need Spack to produce. 

Additional changes made in this PR:

- Python native compression utility importing has been refactored to minimize runtime impact of Python import system.
- Tarball handling and documentation updated to reflect transition away from native python module (see #31563)
- Compression test refactored to reflect impact of above changes
- Refactor `file` usage to a more accessible location within Spack
- `decompressor_for` no longer requires both an extension and a file argument as the extension was being derived from the path in all usages.

### DIFF
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -115,7 +115,7 @@ def path_contains_subdirectory(path, root):
 
 @memoized
 def file_command(*args):
-    file_cmd = which('file')
+    file_cmd = which("file")
     for arg in args:
         file_cmd.add_default_arg(arg)
     return file_cmd
@@ -123,7 +123,7 @@ def file_command(*args):
 
 @memoized
 def _get_mime_type():
-    return file_command('-b', '-h', '--mime-type')
+    return file_command("-b", "-h", "--mime-type")
 
 
 @memoized
@@ -132,11 +132,10 @@ def _get_mime_type_compressed():
     compression first
     """
     mime_uncompressed = _get_mime_type()
-    mime_uncompressed.add_default_arg('-Z')
+    mime_uncompressed.add_default_arg("-Z")
     return mime_uncompressed
 
 
-@memoized
 def mime_type(filename):
     """Returns the mime type and subtype of a file.
 
@@ -147,12 +146,11 @@ def mime_type(filename):
         Tuple containing the MIME type and subtype
     """
     output = _get_mime_type()(filename, output=str, error=str).strip()
-    tty.debug('==> ' + output)
-    type, _, subtype = output.partition('/')
+    tty.debug("==> " + output)
+    type, _, subtype = output.partition("/")
     return type, subtype
 
 
-@memoized
 def compressed_mime_type(filename):
     """Same as mime_type but checks for type that has been compressed
 
@@ -163,8 +161,8 @@ def compressed_mime_type(filename):
         Tuple containing the MIME type and subtype
     """
     output = _get_mime_type_compressed()(filename, output=str, error=str).strip()
-    tty.debug('==> ' + output)
-    type, _, subtype = output.partition('/')
+    tty.debug("==> " + output)
+    type, _, subtype = output.partition("/")
     return type, subtype
 
 
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -24,7 +24,7 @@ from llnl.util.compat import Sequence
 from llnl.util.lang import dedupe, memoized
 from llnl.util.symlink import islink, symlink
 
-from spack.util.executable import Executable
+from spack.util.executable import Executable, which
 from spack.util.path import path_to_os_path, system_path_filter
 
 is_windows = _platform == "win32"
@@ -111,6 +111,61 @@ def path_contains_subdirectory(path, root):
     norm_root = os.path.abspath(root).rstrip(os.path.sep) + os.path.sep
     norm_path = os.path.abspath(path).rstrip(os.path.sep) + os.path.sep
     return norm_path.startswith(norm_root)
+
+
+@memoized
+def file_command(*args):
+    file_cmd = which('file')
+    for arg in args:
+        file_cmd.add_default_arg(arg)
+    return file_cmd
+
+
+@memoized
+def _get_mime_type():
+    return file_command('-b', '-h', '--mime-type')
+
+
+@memoized
+def _get_mime_type_compressed():
+    """Same as _get_mime_type but attempts to check for
+    compression first
+    """
+    mime_uncompressed = _get_mime_type()
+    mime_uncompressed.add_default_arg('-Z')
+    return mime_uncompressed
+
+
+@memoized
+def mime_type(filename):
+    """Returns the mime type and subtype of a file.
+
+    Args:
+        filename: file to be analyzed
+
+    Returns:
+        Tuple containing the MIME type and subtype
+    """
+    output = _get_mime_type()(filename, output=str, error=str).strip()
+    tty.debug('==> ' + output)
+    type, _, subtype = output.partition('/')
+    return type, subtype
+
+
+@memoized
+def compressed_mime_type(filename):
+    """Same as mime_type but checks for type that has been compressed
+
+    Args:
+        filename (str): file to be analyzed
+
+    Returns:
+        Tuple containing the MIME type and subtype
+    """
+    output = _get_mime_type_compressed()(filename, output=str, error=str).strip()
+    tty.debug('==> ' + output)
+    type, _, subtype = output.partition('/')
+    return type, subtype
 
 
 #: This generates the library filenames that may appear on any OS.
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -24,7 +24,7 @@ from llnl.util.compat import Sequence
 from llnl.util.lang import dedupe, memoized
 from llnl.util.symlink import islink, symlink
 
-from spack.util.executable import Executable, which
+from spack.util.executable import CommandNotFoundError, Executable, which
 from spack.util.path import path_to_os_path, system_path_filter
 
 is_windows = _platform == "win32"
@@ -115,7 +115,14 @@ def path_contains_subdirectory(path, root):
 
 @memoized
 def file_command(*args):
-    file_cmd = which("file")
+    """Creates entry point to `file` system command with provided arguments"""
+    try:
+        file_cmd = which("file", required=True)
+    except CommandNotFoundError as e:
+        if is_windows:
+            raise CommandNotFoundError("`file` utility is not available on Windows")
+        else:
+            raise e
     for arg in args:
         file_cmd.add_default_arg(arg)
     return file_cmd
@@ -123,6 +130,9 @@ def file_command(*args):
 
 @memoized
 def _get_mime_type():
+    """Generate method to call `file` system command to aquire mime type
+    for a specified path
+    """
     return file_command("-b", "-h", "--mime-type")
 
 
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -54,7 +54,7 @@ import spack.util.pattern as pattern
 import spack.util.url as url_util
 import spack.util.web as web_util
 import spack.version
-from spack.util.compression import decompressor_for, extension
+from spack.util.compression import decompressor_for, extension_from_path
 from spack.util.executable import CommandNotFoundError, which
 from spack.util.string import comma_and, quote
 
@@ -613,7 +613,7 @@ class VCSFetchStrategy(FetchStrategy):
 
     @_needs_stage
     def archive(self, destination, **kwargs):
-        assert extension(destination) == "tar.gz"
+        assert extension_from_path(destination) == "tar.gz"
         assert self.stage.source_path.startswith(self.stage.path)
 
         tar = which("tar", required=True)
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -486,7 +486,7 @@ class URLFetchStrategy(FetchStrategy):
             tty.debug("Source already staged to %s" % self.stage.source_path)
             return
 
-        decompress = decompressor_for(self.archive_file, self.extension)
+        decompress = decompressor_for(self.archive_file)
 
         # Below we assume that the command to decompress expand the
         # archive in the current working directory
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -486,7 +486,7 @@ class URLFetchStrategy(FetchStrategy):
             tty.debug("Source already staged to %s" % self.stage.source_path)
             return
 
-        decompress = decompressor_for(self.archive_file)
+        decompress = decompressor_for(self.archive_file, self.extension)
 
         # Below we assume that the command to decompress expand the
         # archive in the current working directory
--- a/lib/spack/spack/test/util/compression.py
+++ b/lib/spack/spack/test/util/compression.py
@@ -22,9 +22,9 @@ ext_archive = {}
     for ext in scomp.ALLOWED_ARCHIVE_TYPES
     if "TAR" not in ext
 ]
-# Spack does not use Python native handling for tarballs
-# Don't test tarballs in native test
-native_archive_list = [key for key in ext_archive.keys() if 'tar' not in key]
+# Spack does not use Python native handling for tarballs or zip
+# Don't test tarballs or zip in native test
+native_archive_list = [key for key in ext_archive.keys() if "tar" not in key and "zip" not in key]
 
 
 def support_stub():
@@ -33,9 +33,9 @@ def support_stub():
 
 @pytest.fixture
 def compr_support_check(monkeypatch):
-    monkeypatch.setattr(scomp, 'is_lzma_supported', support_stub)
-    monkeypatch.setattr(scomp, 'is_gzip_supported', support_stub)
-    monkeypatch.setattr(scomp, 'is_bz2_supported', support_stub)
+    monkeypatch.setattr(scomp, "is_lzma_supported", support_stub)
+    monkeypatch.setattr(scomp, "is_gzip_supported", support_stub)
+    monkeypatch.setattr(scomp, "is_bz2_supported", support_stub)
 
 
 @pytest.fixture
@@ -48,10 +48,9 @@ def archive_file(tmpdir_factory, request):
     return os.path.join(str(tmpdir), "Foo.%s" % extension)
 
 
-@pytest.mark.parametrize('archive_file', native_archive_list, indirect=True)
+@pytest.mark.parametrize("archive_file", native_archive_list, indirect=True)
 def test_native_unpacking(tmpdir_factory, archive_file):
-    extension = scomp.extension(archive_file)
-    util = scomp.decompressor_for(archive_file, extension)
+    util = scomp.decompressor_for(archive_file)
     tmpdir = tmpdir_factory.mktemp("comp_test")
     with working_dir(str(tmpdir)):
         assert not os.listdir(os.getcwd())
@@ -65,9 +64,8 @@ def test_native_unpacking(tmpdir_factory, archive_file):
 
 @pytest.mark.parametrize("archive_file", ext_archive.keys(), indirect=True)
 def test_system_unpacking(tmpdir_factory, archive_file, compr_support_check):
-    extension = scomp.extension(archive_file)
     # actually run test
-    util = scomp.decompressor_for(archive_file, extension)
+    util = scomp.decompressor_for(archive_file)
     tmpdir = tmpdir_factory.mktemp("system_comp_test")
     with working_dir(str(tmpdir)):
         assert not os.listdir(os.getcwd())
@@ -80,23 +78,23 @@ def test_system_unpacking(tmpdir_factory, archive_file, compr_support_check):
 
 
 def test_unallowed_extension():
-    bad_ext_archive = "Foo.py"
+    bad_ext_archive = "Foo.cxx"
     with pytest.raises(CommandNotFoundError):
-        scomp.decompressor_for(bad_ext_archive, "py")
+        scomp.decompressor_for(bad_ext_archive)
 
 
 @pytest.mark.parametrize("archive", ext_archive.values())
 def test_get_extension(archive):
-    ext = scomp.extension(archive)
+    ext = scomp.extension_from_path(archive)
     assert ext_archive[ext] == archive
 
 
 def test_get_bad_extension():
-    archive = "Foo.py"
-    ext = scomp.extension(archive)
+    archive = "Foo.cxx"
+    ext = scomp.extension_from_path(archive)
     assert ext is None
 
 
 @pytest.mark.parametrize("path", ext_archive.values())
-def test_allowed_archvie(path):
+def test_allowed_archive(path):
     assert scomp.allowed_archive(path)
--- a/lib/spack/spack/test/util/compression.py
+++ b/lib/spack/spack/test/util/compression.py
@@ -22,6 +22,9 @@ ext_archive = {}
     for ext in scomp.ALLOWED_ARCHIVE_TYPES
     if "TAR" not in ext
 ]
+# Spack does not use Python native handling for tarballs
+# Don't test tarballs in native test
+native_archive_list = [key for key in ext_archive.keys() if 'tar' not in key]
 
 
 def support_stub():
@@ -30,10 +33,9 @@ def support_stub():
 
 @pytest.fixture
 def compr_support_check(monkeypatch):
-    monkeypatch.setattr(scomp, "lzma_support", support_stub)
-    monkeypatch.setattr(scomp, "tar_support", support_stub)
-    monkeypatch.setattr(scomp, "gzip_support", support_stub)
-    monkeypatch.setattr(scomp, "bz2_support", support_stub)
+    monkeypatch.setattr(scomp, 'is_lzma_supported', support_stub)
+    monkeypatch.setattr(scomp, 'is_gzip_supported', support_stub)
+    monkeypatch.setattr(scomp, 'is_bz2_supported', support_stub)
 
 
 @pytest.fixture
@@ -46,7 +48,7 @@ def archive_file(tmpdir_factory, request):
     return os.path.join(str(tmpdir), "Foo.%s" % extension)
 
 
-@pytest.mark.parametrize("archive_file", ext_archive.keys(), indirect=True)
+@pytest.mark.parametrize('archive_file', native_archive_list, indirect=True)
 def test_native_unpacking(tmpdir_factory, archive_file):
     extension = scomp.extension(archive_file)
     util = scomp.decompressor_for(archive_file, extension)
--- a/lib/spack/spack/test/util/compression.py
+++ b/lib/spack/spack/test/util/compression.py
@@ -78,6 +78,8 @@ def test_system_unpacking(tmpdir_factory, archive_file, compr_support_check):
 
 
 def test_unallowed_extension():
+    # use a cxx file as python files included for the test
+    # are picked up by the linter and break style checks
     bad_ext_archive = "Foo.cxx"
     with pytest.raises(CommandNotFoundError):
         scomp.decompressor_for(bad_ext_archive)
--- a/lib/spack/spack/util/compression.py
+++ b/lib/spack/spack/util/compression.py
@@ -26,9 +26,7 @@ ALLOWED_ARCHIVE_TYPES = (
     [".".join(ext) for ext in product(PRE_EXTS, EXTS)] + PRE_EXTS + EXTS + NOTAR_EXTS
 )
 
-ALLOWED_SINGLE_EXT_ARCHIVE_TYPES = (
-    PRE_EXTS + EXTS + NOTAR_EXTS
-)
+ALLOWED_SINGLE_EXT_ARCHIVE_TYPES = PRE_EXTS + EXTS + NOTAR_EXTS
 
 is_windows = sys.platform == "win32"
 
--- a/lib/spack/spack/util/compression.py
+++ b/lib/spack/spack/util/compression.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import inspect
 import io
 import os
 import re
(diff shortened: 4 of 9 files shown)


lib/spack/llnl/util/filesystem.py
```diff
@@ -110,6 +110,61 @@ def path_contains_subdirectory(path, root):
...
+def _get_mime_type_compressed():
+    """Same as _get_mime_type but attempts to check for
+    compression first
+    """
+    mime_uncompressed = _get_mime_type()
+    mime_uncompressed.add_default_arg('-Z')
+    return mime_uncompressed
+
+
+@memoized
```
scheibelp: (nitpick) while memoizing the functions that return the command seems fine. If the file "underneath" the filename was replaced, then IMO we should be able to detect the change (i.e. don't memoize this or `compressed_mime_type`).

lib/spack/spack/fetch_strategy.py
```diff
@@ -284,8 +284,6 @@ def __init__(self, url=None, checksum=None, **kwargs):
         self.extra_options = kwargs.get('fetch_options', {})
         self._curl = None
 
-        self.extension = kwargs.get('extension', None)
```
scheibelp: Note this should still be available: without this, users can't specify an `extension` argument on their `version()` declaration, which I think should still be possible.

lib/spack/spack/util/compression.py
```diff
@@ -14,45 +14,45 @@
...
+    _bz2_support = True
+except ImportError:
+    _bz2_support = False
 
-def bz2_support():
-    try:
-        import bz2 # noqa
-        return True
-    except ImportError:
-        return False
```
scheibelp, line 33: (question) It looks like these modules switch from trying an import as part of the function to doing an import as part of loading the module; I presume this is to avoid repeated import attempts, but is there another reason? I wasn't under the impression that repeated `import` would be expensive, but would memoize achieve the same thing? Overall it doesn't bother me to do things in this updated manner either, I just want to confirm that it isn't essential to the other parts of the PR.

lib/spack/spack/util/compression.py
```diff
@@ -278,6 +275,208 @@ def decompressor_for(path, ext):
     return _untar
 
 
+class FileType:
```
scheibelp: IMO it's worth documenting that this should never be called directly. IMO it's usually clearest to do this with "interface-like"-classes by raising `NotIImplemented` in their methods

lib/spack/spack/util/compression.py
```diff
@@ -288,17 +487,36 @@ def strip_extension(path):
...
+    return False
+
+
+def strip_sourceforge_suffix(path):
+    """remove sourceforge suffix from path if present
+    return stripped path"""
+    # Strip sourceforge suffix.
+    if re.search(r'((?:sourceforge.net|sf.net)/.*)/download$', path):
+        return os.path.dirname(path)
+    return path
```
scheibelp, line 645: This has overlap with `url.py` module: presumably this method should be consolidated.

lib/spack/spack/util/compression.py
```diff
@@ -278,6 +275,208 @@ def decompressor_for(path, ext):
...
+                if not decomp_stream:
+                    # we were unable to decompress, fallback on
+                    # filename parsing to derive extension
+                    return extension(stream.name)
+                prefix_ext = extension_from_stream(decomp_stream, decompress=decompress)
+            return suffix_ext if not prefix_ext else '.'.join([prefix_ext, suffix_ext])
+    return None
+
+
+def extension_from_file(file, decompress=False):
```
scheibelp, line 554: (minor) I'm not sure what "from" adds in terms of the function's intended purpose (i.e. "file_extension" would mean the same thing to me.

lib/spack/spack/util/compression.py
```diff
@@ -278,6 +275,208 @@ def decompressor_for(path, ext):
...
+            suffix_ext = arc_type.extension()
+            prefix_ext = ''
+            if arc_type.is_compression_type() and decompress:
+                # stream represents compressed file
+                # get decompressed stream (if possible)
+                decomp_stream = arc_type.decomp_in_memory(stream)
+                if not decomp_stream:
+                    # we were unable to decompress, fallback on
+                    # filename parsing to derive extension
+                    return extension(stream.name)
```
scheibelp: Now that we have two methods of extension detection, IMO it would be helpful to distinguish them more by name: i.e. "name_parsing_extension_detection" and "magic_number_extension_detection" (feel free to choose shorter names but FWIW you can make them as long as you wish and it won't bother me).

lib/spack/spack/util/compression.py
```diff
@@ -278,6 +275,208 @@ def decompressor_for(path, ext):
...
+        with open(file, 'rb') as f:
+            ext = extension_from_stream(f, decompress)
+            # based on magic number, file is compressed
+            # tar archive. Check to see if file is abbreviated as
+            # t[xz|gz|bz2|bz]
+            if ext and ext.startswith('tar.'):
+                suf = ext.split('.')[1]
+                abbr = 't' + suf
+                if check_extension(file, abbr):
+                    return abbr
```
scheibelp, line 584: This looks like it's trying to be careful to say `.tgz` if the file name has `.tgz` in it, but I'm not sure I understand why we're trying to be careful about that: I would think we could always return `.tar.gz` for files even if they are named `.tgz` (in other words, magic number parsing can dictate the extension type and doesn't have to worry about what the filename is).

scheibelp (changes requested):
I have a few requests and some questions. I didn't get a chance to examine the finer details of this so I will need to do that tomorrow, but I think there's some actionable things in the meantime.

lib/spack/spack/util/compression.py
johnwparent: You are correct that the primary motivation was to avoid performing a large number of imports. Specifically, the flow of calling `tar_support()` which would attempt to import `tarfile` immediately followed by the line `import tarfile` on return of a `True` value. This felt congested and wasteful. I did consider it initially but I don't believe that memoizing is the right approach, as that would still retain the function scope for the imported modules, precluding general module usage without retuning the module directly from the method. Unless I misunderstood, if so, I can revert this change.
I wouldn't call it essential to the PR, or any functionality.

lib/spack/spack/fetch_strategy.py
johnwparent: That makes sense. It was unused after I took it out of `decompressor_for` so I thought it was extraneous, I didn't realize where it was coming from, my mistake.

lib/spack/spack/util/compression.py
johnwparent: Good to know that longer names are acceptable. A rename is in order for certain though, I'll do that now.

lib/spack/spack/util/compression.py
johnwparent: The intent was to contrast with the `extension_from_stream` method and indicate from what type of source the extension is being derived. The `extension` method would be renamed to `extension_from_path` if this is an acceptable idiom to you.

lib/spack/spack/util/compression.py
johnwparent: I don't think this extra check detracts from the quality of the method (if I'm wrong here, please correct me and disregard the rest of this comment) and it allows behavior to be consistent between our `extension` and `extension_from_file` methods as well as before and after this change within Spack. For the purposes of `decompressor_for`, yes, this is a frivolous distinction as the decomp strategy will be the same for `.tgz` and `tar.gz`, but in other contexts, having the appropriate string returned could be very useful (url parsing for example).

lib/spack/spack/util/compression.py
johnwparent: The intent was to contrast with the `extension_from_stream` method and indicate from what type of source the extension is being derived. The `extension` method would be renamed to `extension_from_path` if this is an acceptable idiom to you.

lib/spack/spack/util/compression.py
johnwparent: In this case the base class is defining a fair amount of default implementations, so I could raise some `NotImplementedError`s however I would then need to either duplicate a fair amount of code or create a mixin class that essentially is just a copy of the current state of `FileType` which to me propagates the issue. If you had a better way in mind, I am absolutely all ears, because I agree, having `FileType` hanging around is confusing. Docs are updated.

lib/spack/spack/util/compression.py
scheibelp: This is fine, my main goal here was to understand which parts of the PR are essential.

lib/spack/spack/util/compression.py
scheibelp: > The extension method would be renamed to extension_from_path if this is an acceptable idiom to you.

 That makes sense and I agree with changing the name of `extension(`

lib/spack/spack/util/compression.py
```diff
@@ -288,17 +546,26 @@ def strip_extension(path):
     return path
 
 
+def check_extension(path, ext):
+    """Check if extension is present in path"""
+    # Strip sourceforge suffix.
```
scheibelp, line 631: I think this comment is out of date

scheibelp: Without that extra logic, this ends up being the same as `ext in path`

lib/spack/spack/util/compression.py
scheibelp: The primary issue is that I found this confusing: I think that could be addressed by updating the docstring to explain it will try to return `.tgz` if the path contains `.tgz` (and that this might be useful for functions which want to take advantage of magic number matching to aid their parsing).

I also mentioned in `check_extension` that right now that is the same as `abbr in file`: while I think that is generally safe if `abbr` were `".tgz"`, I'm concerned that `"tgz"` (i.e. without the `"."`) is too loose of a match.

scheibelp:
The style tests are failing in the bootstrap phase: I'm checking if this can be fixed by restarting the CI

scheibelp:
@johnwparent can you push an empty commit here? That restarts things "more-thoroughly". I haven't seen this error pop up in other PRs (except for https://github.com/spack/spack/pull/31141) but I agree there doesn't seem to be a reason these would cause that error. If the error persists, I will investigate it tomorrow.

johnwparent:
@scheibelp It looks like the pip errors have been resolved. The bootstrapping tests seem unrelated at first glace but if the refactor is causing things to be extracted improperly that could result in this issue. Mypy parsing the `Foo` file seems to remain an issue so I'll look into configuring mypy or maybe the spack style cmd to ignore extension-less files.

haampie:
Can you please fix the mirror/source cache code while you're at it for consistency? You'd expect fetching and caching to use the same code paths, but nope, they each have their own heuristics to determine the file type 🤦. In particular, the source cache business special cases `github.com/x/y/tarball/ref` type of URLs and it also apparently handles `version(..., extension='.tar.gz')` which we should immediately deprecate 😭.

See https://github.com/spack/spack/blob/0662bbdc4342031250e464b844f98ee7313d42ca/lib/spack/spack/mirror.py#L324

scheibelp:
> and it also apparently handles version(..., extension='.tar.gz') which we should immediately deprecate

To be clear: you are arguing that we should remove the user option to set `extension=.tar.gz` (or in general, to manually set the extension)? I think it's worth keeping this in case the logic to determine the extension fails. Admittedly, that should be rare with this PR, but it would give users unfamiliar with the core logic a means to temporarily patch issues like https://github.com/spack/spack/issues/31499 while a more-general fix is in progress.

lib/spack/spack/util/compression.py
johnwparent: I believe this has been addressed by the method doc string. If this is not the case, I can improve the docstring or we can discuss alternatives.

lib/spack/spack/util/compression.py
johnwparent: The methods have been consolidated in a way that I believe addresses this comment.

lib/spack/spack/util/compression.py
```diff
@@ -291,27 +286,326 @@ def decompressor_for(path, ext):
...
+                prefix_ext = extension_from_stream(decomp_stream, decompress=decompress)
+            return suffix_ext if not prefix_ext else ".".join([prefix_ext, suffix_ext])
+    return None
+
+
+def extension_from_file(file, decompress=False):
+    """Return extension from archive file path
+    Extension is derived based on magic number parsing similar
+    to the `file` utility. Attempts to return abbreviated file extensions
+    whenever relevant (i.e. compressed tarball as `.tgz`). This distinction
```
scheibelp: I think "whenever relevant" is vague: we return ".tgz" when the file name has ".tgz" (and moreover we return ".tar.gz" when the name has ".tar.gz" in it, even though ".tgz" would be fine.

scheibelp:
Unit tests for Python 2.7 are failing on:

```
def decomp_in_memory(cls, stream):
        if is_bz2_supported():
            return io.BytesIO(
                initial_bytes=bz2.BZ2Decompressor().decompress(
>                   stream.read(), max_length=TarFileType.offset() + TarFileType.header_size()
                )
            )
E           TypeError: decompress() takes no keyword arguments
```

(this might be a Python-2.7-only problem)

johnwparent:
> Unit tests for Python 2.7 are failing on:
> 
> ```
> def decomp_in_memory(cls, stream):
>         if is_bz2_supported():
>             return io.BytesIO(
>                 initial_bytes=bz2.BZ2Decompressor().decompress(
> >                   stream.read(), max_length=TarFileType.offset() + TarFileType.header_size()
>                 )
>             )
> E           TypeError: decompress() takes no keyword arguments
> ```
> 
> (this might be a Python-2.7-only problem)

You are correct in that this is a Python2 only problem, but I refactored so the method works similarly for both 2 and 3. Should be resolved now.

lib/spack/spack/fetch_strategy.py
```diff
@@ -485,7 +485,7 @@ def expand(self):
             tty.debug("Source already staged to %s" % self.stage.source_path)
             return
 
-        decompress = decompressor_for(self.archive_file, self.extension)
+        decompress = decompressor_for(self.archive_file)
```
scheibelp: Is `self.extension` no longer used? Without that, you can't have `version(..., extension=...)`

I think it would be very useful to allow setting this to handle cases that we discussed with old archives which may not have a magic number (but we should also continue to support packages that use `extension` in their `version()` declarations).

I don't see any `builtin.mock` packages defining `extension` for version, so it may not have been tested properly (apologies for that - that was my fault) - can you add such a test after restoring that support?

lib/spack/llnl/util/filesystem.py
```diff
@@ -113,6 +113,59 @@ def path_contains_subdirectory(path, root):
...
+def file_command(*args):
+    file_cmd = which("file")
+    for arg in args:
+        file_cmd.add_default_arg(arg)
+    return file_cmd
+
+
+@memoized
+def _get_mime_type():
```
scheibelp, line 132: Add short docstring like "Generate method that can determine mime type on a specified path"

lib/spack/llnl/util/filesystem.py
```diff
@@ -113,6 +113,59 @@ def path_contains_subdirectory(path, root):
     return norm_path.startswith(norm_root)
 
 
+@memoized
+def file_command(*args):
+    file_cmd = which("file")
+    for arg in args:
+        file_cmd.add_default_arg(arg)
+    return file_cmd
```
scheibelp, line 128: If `which` returns None (it would on Windows right?) then this function should raise an exception.

lib/spack/spack/test/util/compression.py
```diff
@@ -78,23 +78,23 @@ def test_system_unpacking(tmpdir_factory, archive_file, compr_support_check):
 
 
 def test_unallowed_extension():
-    bad_ext_archive = "Foo.py"
+    bad_ext_archive = "Foo.cxx"
     with pytest.raises(CommandNotFoundError):
-        scomp.decompressor_for(bad_ext_archive, "py")
+        scomp.decompressor_for(bad_ext_archive)
```
scheibelp, line 85: (question) IIRC this was changed because the CI complains if we add a test file called `Foo.py` right? That's fine if so, but it might be worth adding a comment here mentioning that.

lib/spack/spack/util/compression.py
```diff
@@ -201,6 +196,35 @@ def _lzma_decomp(archive_file):
             return _xz(archive_file)
 
 
+def _win_compressed_tarball_handler(archive_file):
+    """Decompress and extract compressed tarballs on Windows.
+    This method uses 7zips in conjunction with the tar utility
```
scheibelp: (typo) 7zips

lib/spack/spack/util/compression.py
```diff
@@ -201,6 +196,35 @@ def _lzma_decomp(archive_file):
...
+    to perform decompression and extraction in a two step process
+    first using 7zip to decompress, and tar to extract.
+
+    The motivation for this method is the inability of 7zip
+    to directly decompress and extract compressed archives
+    in a single shot without undocumented workarounds, and
+    inconsistent degrees of lzma and compress compression
```
scheibelp: (wording) I'm having trouble parsing "inconsistent degrees of lzma and compress compression"

lib/spack/spack/util/compression.py
```diff
@@ -285,33 +309,346 @@ def decompressor_for(path, ext):
...
+        return False
+
+
+class CompressedFileTypeInterface(FileTypeInterface):
+    """Interface class for FileTypes that include compression information"""
+
+    _compressed = True
+
+    @classmethod
+    def decomp_in_memory(cls, stream):
```
scheibelp: This should have a short docstring that makes it clear that this is for peering into the compressed archive and generally not used for expanding the whole archive.

lib/spack/spack/util/compression.py
```diff
@@ -285,33 +309,346 @@ def decompressor_for(path, ext):
...
+    def name(cls):
+        raise NotImplementedError
+
+    @classmethod
+    def is_compression_type(cls):
+        return cls._compressed
+
+    @classmethod
+    def offset(cls):
+        return cls._OFFSET
```
scheibelp: I think it would be more straightforward to refer to the underlying properties and "promote" their names. E.g. the class can just have a .offset attribute"

This also applies to `magic_number`, `extension`, and `compressed`.

To be clear, methods which derive new values (like `header_size`) are fine.

lib/spack/spack/util/compression.py
```diff
@@ -285,33 +309,346 @@ def decompressor_for(path, ext):
...
+            )
+        return None
+
+
+class LzmaFileType(CompressedFileTypeInterface):
+    _MAGIC_NUMBER = b"\xfd7zXZ"
+    _ext = "xz"
+
+    @classmethod
+    def name(cls):
```
scheibelp: Since this doesn't use `cls`, you should also be able to call this a `@staticmethod` (in which case no args are necessary).

Likewise for all methods not using the class.

lib/spack/spack/util/compression.py
```diff
@@ -285,33 +309,346 @@ def decompressor_for(path, ext):
...
+        return "gzip compressed data"
+
+    @classmethod
+    def decomp_in_memory(cls, stream):
+        if is_gzip_supported():
+            return io.BytesIO(
+                initial_bytes=gzip.GzipFile(fileobj=stream).read(
+                    TarFileType.offset() + TarFileType.header_size()
```
scheibelp: I'd find it slightly clearer if this (and other `FileType`s) set all attributes at the beginning like:

```
class GZipFileType(CompressedFileTypeInterface):
    offset = TarFileType.offset
    magic_number = b"\x1f\x8b\x08"
    extension = "gz"
```

lib/spack/spack/util/compression.py
```diff
@@ -285,33 +309,346 @@ def decompressor_for(path, ext):
...
+    def magic_number(cls):
+        return [cls._MAGIC_NUMBER_LZH, cls._MAGIC_NUMBER_LZW]
+
+    @classmethod
+    def header_size(cls):
+        return max(len(cls._MAGIC_NUMBER_LZW), len(cls._MAGIC_NUMBER_LZH))
+
+    @classmethod
+    def _bytes_check(cls, magic_bytes):
+        for magic in cls.magic_number():
```
scheibelp: It seems like you want to imply that `magic_number` is an interface method but it's return type is inconsistent (sometimes an int, sometimes a list). That discrepancy should be resolved.

lib/spack/spack/util/compression.py
```diff
@@ -285,33 +309,346 @@ def decompressor_for(path, ext):
...
+    def name(cls):
+        return "compress'd data"
+
+    @classmethod
+    def magic_number(cls):
+        return [cls._MAGIC_NUMBER_LZH, cls._MAGIC_NUMBER_LZW]
+
+    @classmethod
+    def header_size(cls):
+        return max(len(cls._MAGIC_NUMBER_LZW), len(cls._MAGIC_NUMBER_LZH))
```
scheibelp: You are prepped for these being different lengths, but if they are, one of them will be longer, in which case I assume we'd pull enough bytes for the longer one (which is a problem because we don't do prefix matching and just check if `==`).

scheibelp (changes requested):
Two main issues:

* We still need to support `version(..., extension=...)` declarations in `package.py`
* FileTypeInterface seems to generalize a number of functions in anticipation of future extension, but these generalizations lead to inconsistencies. I make a number of specific comments there about this.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/fetch_strategy.py
johnwparent: Addressed.

lib/spack/llnl/util/filesystem.py
johnwparent: Addressed.

lib/spack/llnl/util/filesystem.py
johnwparent: Addressed.

lib/spack/spack/test/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
johnwparent: Addressed.

lib/spack/spack/util/compression.py
```diff
@@ -201,6 +197,34 @@ def _lzma_decomp(archive_file):
...
+    # 7zip is able to one shot extract compressed archives
+    # that have been named .txz. If that is the case, there will
+    # be no intermediate archvie to extract.
+    if os.path.exists(decomped_tarball):
+        # run tar on newly decomped archive
+        outfile = _untar(decomped_tarball)
+        # clean intermediate archive to mimic end result
+        # produced by one shot decomp/extraction
+        os.remove(decomped_tarball)
+        return outfile
```
scheibelp, line 222: It looks like

* if `decomped_tarball` doesn't exist, we return that path (which doesn't exist)
* if it does exist, we return `outfile`, a path that does exist

This seems inconsistent. If this is expected, documenting the return type in the docstring would be useful.

lib/spack/spack/util/compression.py
scheibelp: FWIW this is now a static method which accepts a single parameter called `cls` but that is not the class: i.e. `LzmaFileType.name()` would fail. Does anything use this method?

lib/spack/spack/util/compression.py
```diff
@@ -248,70 +272,352 @@ def _7zip(archive_file):
...
+            # t[xz|gz|bz2|bz]
+            if ext and ext.startswith("tar."):
+                suf = ext.split(".")[1]
+                abbr = "t" + suf
+                if check_extension(file, abbr):
+                    return abbr
+            if not ext:
+                # If unable to parse extension from stream,
+                # attempt to fall back to string parsing
+                ext = extension_from_path(file)
```
scheibelp, line 588: Can you add a call to `tty.debug` here? Something like:

```
tty.debug("Cannot determine extension from magic number: fall back on regex name parsing")
```

IMO it would be useful if we run into decompression problems to know roughly how we had determined the extension.

For clarity, it might be worth adding a debug message for both paths (e.g. when magic number parsing succeeds add `tty.debug("Determined extension from magic number: ...")`)

lib/spack/llnl/util/filesystem.py
```diff
@@ -113,6 +113,69 @@ def path_contains_subdirectory(path, root):
...
+    Returns:
+        Tuple containing the MIME type and subtype
+    """
+    output = _get_mime_type()(filename, output=str, error=str).strip()
+    tty.debug("==> " + output)
+    type, _, subtype = output.partition("/")
+    return type, subtype
+
+
+def compressed_mime_type(filename):
```
scheibelp, line 164: (question) is this used anywhere? It looks like `mime_type` moved from `relocate` so that this could be defined (and presumably used by modules that should not otherwise care about `relocate.py`). Moving these functions here is fine (although it's worth knowing whether doing so is essential to this PR).

scheibelp (changes requested):
I have a few minor requests and a question.

lib/spack/spack/util/compression.py
johnwparent: Yes, this is a good catch. If the `decomped_tarball` path does not exist, we want to return the version of that with no extensions, so we should strip off the final extension. Amended locally, pushing shortly.

lib/spack/spack/util/compression.py
johnwparent: This method is not currently used (although it was helpful for debugging) but it was trivial to add and more directly mimics `file`'s behavior. Fixed the problem locally, pushing shortly.

lib/spack/spack/util/compression.py
johnwparent: Amended.

lib/spack/llnl/util/filesystem.py
johnwparent: the `mime_type` work was originally going to be our approach before further investigation revealed the magic number parsing was a superieior approach. This is leftover from my initial pass implementing `file` based type parsing that we abandoned. I left it in as it added (ostensibly useful) components and I felt that `relocate` was perhaps not the most logical/visible place for this to live. It is not essential to the PR.

lib/spack/spack/util/compression.py
```diff
@@ -201,6 +199,40 @@ def _lzma_decomp(archive_file):
...
+        # clean intermediate archive to mimic end result
+        # produced by one shot decomp/extraction
+        os.remove(decomped_tarball)
+        return outfile
+    else:
+        # File was a .txz, 7zip can directly extract/decompress these
+        # in one shot (this behavior is undocumented)
+        # so be sure we return the correct extensions in the path
+        _, ext = os.path.splitext(decomped_tarball)
+        decomped_tarball = os.path.basename(decomped_tarball.strip(ext))
```
scheibelp: To remove a string off the end of another string: don't use `.strip()`: it's semantics are weird and will lead to unexpected scenarios. Use something like `re.sub(` (as is done in `strip_extension` - in fact if it's possible to consolidate it that would be good). I apparently missed that when reviewing the additions of the decompression methods like 7zip (that can be left as a TODO for later but I think it would be good to fix this one).

Another refactor (for later - but if you can add a TODO): I think `_7zip()` returning a file path that doesn't exist is a confusing condition to depend on in terms of deciding that what we should actually be doing is untar-ing the result.

lib/spack/spack/util/compression.py
```diff
@@ -248,70 +280,357 @@ def _7zip(archive_file):
...
+                    # We were unable to decompress or unable to derive
+                    # a nested extension from decompressed file.
+                    # Try to use filename parsing to check for
+                    # potential nested extensions if there are any
+                    tty.debug(
+                        "Cannot derive file extension from magic number;"
+                        " falling back to regex path parsing."
+                    )
+                    return extension_from_path(stream.name)
+            tty.debug("File extension successfully dervived by magic number.")
```
scheibelp: Sorry it just occurred to me it would be good to also mention the type that was detected (so e.g. spack debug output would say "based on magic numbers I think it's a .tar.gz" - so if we run into trouble later, a user can manually check and confirm).

Also typo: dervived

lib/spack/spack/util/compression.py
johnwparent: Agreed on the first point, will refactor. As to the second point, _7zip returning a path that doesn't exist is not ideal, and that should be refactored (although it will include some string parsing to determine if a file has an abbreviated extension) but that will still require a check in this method (again involving some semi-fragile string parsing) as well.

scheibelp: I agree that regardless of how `_7zip` itself is factored that this will need a check. I just think that this arrangement should be less confusing than "the file path doesn't exist so it must be a .tar" (this part can be handled in a later PR though).

scheibelp (dismissed):

scheibelp (approved):

PR #31593, opened 2022-07-15
Don't restage develop specs when a patch fails

psakievich:
If a patch fails for a develop spec a restage is attempted which results in an error.  The error message is not clear, and this is not always a desired behavior for develop specs.  For example, if you are doing a bisect and have to go back in the git history before the patch is valid.

This PR removes this behavior and instead prints a warning to let the user know the patch failed and it __could__ affect their build.

### DIFF
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -1565,6 +1565,7 @@ class PackageBase(six.with_metaclass(PackageMeta, PackageViewMixin, object)):
                 msg = ('A patch failure was detected in %s.' % self.name +
                        ' Build errors may occur due to this.')
                 tty.warn(msg)
+                return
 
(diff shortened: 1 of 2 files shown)


psakievich:
@becker33 @tgamblin these failures are just for the unit test I added for the legacy code path. I'm not really sure I understand why this is having issues. Spack stage logic is not something I've really looked into yet. Ideally I would combine these two unit tests because they are essentially testing the same thing just for develop and standard specs.

psakievich:
Note that commit [cbf85faab](https://github.com/spack/spack/pull/31593/commits/cbf85faabc797144fc3b40119431e0597b98279d) was only failing for code coverage, and that was because the legacy branch of operation was untested.

becker33 (approved):

PR #31130, opened 2022-06-14
removing feature bloat: monitor and analyzers

vsoch:
We should try to optimize spack to work really well for 80% of use cases, and both of monitor/analyzers don't fall within that. Regardless of the work done they need to be removed! Since I added both of these I want to remove them too :)

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

PR #31594, opened 2022-07-15
containerize: fix concretization -> concretizer

haampie:


ChristianKniep:
That was fast... Thx! 👍

haampie:
@alalazo it's running

```
# Install the software, remove unnecessary deps
RUN spack install --fail-fast && \
    spack gc -y
```

but w/o active env? I don't see `SPACK_ENV` being set.

alalazo:
@haampie you're right, that's a bug introduced by #31130

becker33 (approved):

PR #31614, opened 2022-07-17
Bracken v2

luke-dt:
the bracken package was significantly out of date (5+ years), and has changed a lot since v1.0.0. Some of the functionalities have moved to KrakenTools, and it works in a really different way now. Because of that, I removed v1.0.0 to simplify things (also no one should be using it).

I copied a lot from the Kraken2 package and Bracken bioconda recipe:

https://github.com/spack/spack/blob/18c2f1a57a69224748e091e912851317d8a00eea/var/spack/repos/builtin/packages/kraken2/package.py

https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bracken/build.sh


alalazo (changes requested):
Can you rebase the branch and fix style (we started to require black for the entire code base)?

PR #28548, opened 2022-01-21
bazel: add patches to compile with gcc-11

glennpj:
Bazel does not currently build with GCC-11. There are patches available on the bazel project site for Alpine Linux that should work. The problem is that java-tools also needs patching.

PR #31615, opened 2022-07-17
TensorFlow/Keras: add versions 2.8–2.9

adamjstewart:
Our TF package has become woefully out-of-date. This PR is an attempt to add the latest versions of TF.

So far I just updated the dependency list, didn't have the energy to dig much deeper than that. Unfortunately, the installation fails almost immediately with:
```
ERROR: /private/var/folders/j1/68dlgpr91vlgs26vty2c8xk80000gn/T/ajstewart/spack-stage/spack-stage-py-tensorflow-2.9.1-fqze2toml7dh3kzf6ew7cnayqgidwvhy/spack-src/tensorflow/tools/build_info/BUILD:9:10: While resolving toolchains for target //tensorflow/tools/build_info:gen_build_info: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
```
Unfortunately, https://github.com/bazelbuild/bazel/issues/10134 provides neither details nor migration instructions. Would love help with this if anyone is more knowledgeable than I am.

### DIFF
--- a/var/spack/repos/builtin/packages/py-keras/package.py
+++ b/var/spack/repos/builtin/packages/py-keras/package.py
@@ -9,17 +9,30 @@ from spack.package import *
 
 
 class PyKeras(PythonPackage):
-    """Deep Learning library for Python. Convnets, recurrent neural networks,
-    and more. Runs on Theano or TensorFlow."""
+    """Deep Learning for humans.
+
+    Keras is a deep learning API written in Python, running on top of the machine
+    learning platform TensorFlow. It was developed with a focus on enabling fast
+    experimentation. Being able to go from idea to result as fast as possible is
+    key to doing good research.
+    """
 
     homepage = "https://keras.io"
-    git = "https://github.com/keras-team/keras"
+    git = "https://github.com/keras-team/keras.git"
     url = "https://github.com/keras-team/keras/archive/refs/tags/v2.7.0.tar.gz"
 
+    version("2.9.0", sha256="90226eaa0337573304f3e5ab44d4d9e3a65fe002776c5cbd0f65b738152c1084")
+    version("2.8.0", sha256="5e777b0101d8385d3a90fc9056f1b2f6313f2c830d2e8181828b300c9229ec0c")
     version("2.7.0", sha256="7502746467ab15184e2e267f13fbb2c3f33ba24f8e02a097d229ba376dabaa04")
     version("2.6.0", sha256="15586a3f3e1ed9182e6e0d4c0dbd052dfb7250e779ceb7e24f8839db5c63fcae")
-    version("2.5.0", commit="9c266106163390f173625c4e7b1ccb03ae145ffc")
+    version("2.5.0", commit="9c266106163390f173625c4e7b1ccb03ae145ffc", deprecated=True)
     version("2.4.3", sha256="fedd729b52572fb108a98e3d97e1bac10a81d3917d2103cc20ab2a5f03beb973")
+    version("2.4.2", sha256="e26bc51b7b8fb7add452cdf6fba77d6509e6c78b9d9ef5fd32fe132c6d9182d2")
+    version("2.4.1", sha256="e282cc9c5c996043b21d045765c0c5bf541c1879232a97a574c51af0ce132cb1")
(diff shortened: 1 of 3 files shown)


adamjstewart:
Successfully builds on Ubuntu 20.04 with Bazel 5.2.0 and GCC 8.4.0. Still doesn't build on aarch64, but neither does the develop version, so this is probably safe to merge. Will keep working on aarch64 support.

nicholas-sly (dismissed):
I don't see any issues with this PR.

var/spack/repos/builtin/packages/py-keras/package.py
```diff
@@ -9,17 +9,30 @@
...
-    git = "https://github.com/keras-team/keras"
+    git = "https://github.com/keras-team/keras.git"
     url = 'https://github.com/keras-team/keras/archive/refs/tags/v2.7.0.tar.gz'
 
+    version('2.9.0', sha256='90226eaa0337573304f3e5ab44d4d9e3a65fe002776c5cbd0f65b738152c1084')
+    version('2.8.0', sha256='5e777b0101d8385d3a90fc9056f1b2f6313f2c830d2e8181828b300c9229ec0c')
     version('2.7.0', sha256='7502746467ab15184e2e267f13fbb2c3f33ba24f8e02a097d229ba376dabaa04')
     version('2.6.0', sha256='15586a3f3e1ed9182e6e0d4c0dbd052dfb7250e779ceb7e24f8839db5c63fcae')
-    version('2.5.0', commit='9c266106163390f173625c4e7b1ccb03ae145ffc')
+    version('2.5.0', commit='9c266106163390f173625c4e7b1ccb03ae145ffc', deprecated=True)
```
glennpj: Why is this one deprecated?

adamjstewart: There never was an official 2.5.0 release, just a release candidate. I guess I don't see a good reason to keep it.

glennpj:
There is an issue with GCC and Cuda versions that will need to be accounted for. I was building with GCC-11.3 and CUDA-11.4.4 but seeing failures like:
```
error: parameter packs not expanded with ‘...’
```
I did some searching and this is being reported in a few places, for instance: https://github.com/NVlabs/instant-ngp/issues/119

In the end it looks like a `conflicts` statement will need to be added to the cuda build system:
```
conflicts('%gcc@11.2:', when='+cuda ^cuda@:11.5')
```

In addition, the following constraint in py-tensorflow will need to be adjusted to allow CUDA-11.6 to get picked up when GCC-11.2+ is being used.
```
depends_on('cuda@:11.4', when='+cuda @2.4:')
```
I will try a few more builds to try to pinpoint what that needs to be. Unfortunately, building this takes a while.

adamjstewart:
@ax3l @Rombur @davidbeckingsale should this first CUDA conflict go in the `CudaPackage` base class?

ax3l:
Yes, if it's not already there?

glennpj:
This PR will also need #28548 merged for older versions of bazel to build with gcc-11+. I just rebased that PR and tested it in tandem with this one.

glennpj (changes requested):
The cuda dependencies will need to look like the following:
```
    depends_on('cuda', when='+cuda')
    depends_on('cuda@:10.2', when='+cuda @:2.3')
    depends_on('cuda@:11.4', when='+cuda @2.4:2.7')
```

aweits (approved):
Looks good.

```
[aweits@skl-a-00 spack]$ spack find -lv py-tensorflow
==> 1 installed package
-- linux-rhel7-skylake_avx512 / gcc@9.3.0 -----------------------
oic7arz py-tensorflow@2.9.1~android~aws~computecpp+cuda~dynamic_kernels~gcp~gdr~hdfs~ignite~ios~jemalloc~kafka~mkl~monolithic~mpi+nccl~ngraph~numa~opencl~rocm~tensorrt~verbs~xla cuda_arch=60,61,70,75,80 patches=2017b3e
```

glennpj (approved):

PR #31620, opened 2022-07-18
Fix a typo in "os.rename" during fetch

alalazo:


haampie (approved):

haampie:
Hm, `fs.rename` is a thing though, yet another windows wrapper utility around `os.rename`.

alalazo:
Fair enough, this was failing in #31049 I thought it was a typo, but now I see we have `os.rename` and `fs.rename` :thinking:

PR #31623, opened 2022-07-18
python: drop 3.9 preference

haampie:
I would be a fan of dropping the conservative preference sooner. If it breaks packages, that means we're missing an upperbound, and adding those improves Spack. If instead we stick to old python by default, we get a bunch of package updates in the meantime fixing new python compat, so that ultimately we would still be missing correct upperbounds when the preference is dropped, because nobody hit any incompatibility.


adamjstewart:
In the past I've been sticking to what other PMs do:

* [Anaconda](https://www.anaconda.com/products/distribution): 3.9
* [Homebrew](https://formulae.brew.sh/formula/python@3.9): 3.9

In the last few years, it seems like the trend has been that other PMs will bump their default version once the next version is first released. So I'm guessing they will start using 3.10 once 3.11 is released (October).

One reason to be conservative here is due to the distributed nature of Python binaries. For users using pip, many packages won't see a Python 3.X wheel until several months after 3.X is first released.

haampie:
Ah ok, but aren't we always building binaries?

adamjstewart:
We are, but if you just use Spack to install pip and then use pip to install everything else, it isn't a great idea to default to a bleeding edge release. I've also seen platform support issues with brand new 3.X.0 releases, although 3.10.5 is probably pretty stable at this point.

PR #31624, opened 2022-07-18
root does not actually install libraries in /lib but in /lib/root

HadrienG2:
This may or may not be related to the recent switch to gnuinstall build mode.


vvolkl (approved):
Yes, indeed the library install path changed with gnuinstall, nice catch!

alalazo (approved):

alalazo:
@spackbot run pipeline

PR #31625, opened 2022-07-18
LLVM requires swig version > 3

Paul-Ferrell:
When building the `+lldb` variant llvm requires a swig newer than 3.0.


haampie:
For all LLVM versions?

Paul-Ferrell:
I was just wondering how broadly this applied. 
<= llvm@9 - swig required, but no version specified in the cmake files (2.0 is probably the min though)
@10, @11 - swig @2.0
@12 - present (last two years) - swig @3.0

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -221,7 +221,9 @@ class Llvm(CMakePackage, CudaPackage):
     depends_on("libxml2")
 
     # lldb dependencies
-    depends_on("swig", when="+lldb")
+    depends_on("swig", when="@:9 +lldb")
+    depends_on("swig@2:", when="@10:11 +lldb")
+    depends_on("swig@3:", when="@12: +lldb")
```
haampie:
```suggestion
    with when("+lldb"):
        depends_on("swig")
        depends_on("swig@2:", when="@10:")
        depends_on("swig@3:", when="@12:")
```

maybe a bit more concise, note that it's ok to have overlapping ranges in `when`, the effective constraint is the intersection

Paul-Ferrell:
There's another bug in the llvm package I'm dealing with at the moment. I'm testing the solution right now - I'll add it to this if it works.

chuckatkins:
Looking at the LLVM CMake code swig shouldn't be a hard requirement for lldb , only if lua or python nbindings are enabled for lldb.  Are you seeing something different?

Paul-Ferrell:
The other problem I was seeing was a Spack bug triggered by my local workaround for this problem.  Once this gets merged, that work around won't be needed, and the problem will go away.

Paul-Ferrell:
The error in the gitlab-ci tests is a timeout, not an actual problem.

trws:
I poked the paraview build, hopefully it will pass this time around.

haampie (approved):

Issue #31627, opened 2022-07-18
Spack mirror --all fails immediately


Paul-Ferrell:
### Steps to reproduce

```
spack --timestamp --debug mirror create --all -n all -d /storage/spack/mirror 
==> [2022-07-18-12:03:13.762045] Imported mirror from built-in commands
==> [2022-07-18-12:03:13.767743] Imported mirror from built-in commands
==> [2022-07-18-12:03:13.769097] Reading config file /storage/spack/spack/etc/spack/defaults/repos.yaml
==> [2022-07-18-12:03:13.898314] Reading config file /storage/spack/spack/etc/spack/defaults/config.yaml
Traceback (most recent call last):
  File "/storage/spack/spack/bin/spack", line 98, in <module>
    sys.exit(spack.main.main())
  File "/storage/spack/spack/lib/spack/spack/main.py", line 893, in main
    return _main(argv)
  File "/storage/spack/spack/lib/spack/spack/main.py", line 848, in _main
    return finish_parse_and_run(parser, cmd_name, env_format_error)
  File "/storage/spack/spack/lib/spack/spack/main.py", line 876, in finish_parse_and_run
    return _invoke_command(command, parser, args, unknown)
  File "/storage/spack/spack/lib/spack/spack/main.py", line 533, in _invoke_command
    return_val = command(parser, args)
  File "/storage/spack/spack/lib/spack/spack/cmd/mirror.py", line 380, in mirror
    action[args.mirror_command](args)
  File "/storage/spack/spack/lib/spack/spack/cmd/mirror.py", line 325, in mirror_create
    mirror_specs = _determine_specs_to_mirror(args)
  File "/storage/spack/spack/lib/spack/spack/cmd/mirror.py", line 301, in _determine_specs_to_mirror
    mirror_specs = spack.mirror.get_all_versions(specs)
  File "/storage/spack/spack/lib/spack/spack/mirror.py", line 427, in get_all_versions
    pkg = spec.package
  File "/storage/spack/spack/lib/spack/spack/spec.py", line 1520, in package
    assert self.concrete, "Spec.package can only be called on concrete specs"
AssertionError: Spec.package can only be called on concrete specs
```

### Error message

See above

### Information on your system

Our script pulls the latest develop. 

```
spack debug report
* **Spack:** 0.19.0.dev0
* **Python:** 3.6.8
* **Platform:** linux-rhel7-broadwell
* **Concretizer:** clingo
```

### General information

[...]

PR #31630, opened 2022-07-18
Fixes the issue in #31627

Paul-Ferrell:
Instead of using spec.package (which causes an exception because the
spec isn't concretized), I instantiate `spec.package_class(spec)` in
all the places that seem to matter.


alalazo:
Superseded by #31636

PR #31631, opened 2022-07-18
Fix CUB imports for CUDA and CMake

### LINKED ISSUE #31568: cmake build system needs a way to add additional CMAKE_PREFIX_PATH entries for things like CUDA
### Steps to reproduce

There needs to be a way to add extra cmake prefixes to certain proprietary packages where the odds of getting it fixed upstream is unlikely.

Create a package (foo) with the following

```python
    def setup_dependent_build_environment(self, env, dependent_spec):
        if self.spec.satisfies("target=x86_64:"):
            env.append_path("CMAKE_PREFIX_PATH", self.prefix.targets + "x86_64-linux/lib/cmake")
```

Then create a package (bar) that depends on that package (foo), and attempt to install that package (bar).

bar will not have the prefix path $foo_prefix/targets/x86_64-linux/lib/cmake provided which causes it to fail.  Providing this as an environment variable doesn't work because spack passes this flag as a command line flag which has higher precedence.

I think the fix is to check this environmental variable in `get_cmake_prefix_path(pkg)` in `build_environment.py` or in `build_systems/cmake.py`, but I wasn't sure how to do that.

### Error message

Example for a modified cuda package and a package that depends on CUB (a part of CUDA) [cusz](https://github.com/robertu94/spack_packages/blob/master/packages/cusz/package.py)

image

* **Spack:** 0.18.0.dev0 (a9a08cbd8af38d371c0a4d9c9fc95b5bf99c258a)
* **Python:** 3.10.5
* **Platform:** linux-fedora36-icelake
* **Concretizer:** clingo

### General information

robertu94:
tl;dr Cuda installs CUB's (a cuda library) cmake files into a place where spack doesn't know about which means it can't tell cmake about it.  We provide a mechanism to tell spack where to find this additional path, and then use it in the cuda package.

fixes #31568

@scheibelp you were the last person to work on the CMAKE_PREFIX_PATH logic.  Could you please review this patch?

@ax3l @Rombur you two are the Cuda maintainers.  Could you please review the change the the Cuda package?

### DIFF
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -1021,24 +1021,28 @@ def get_cmake_prefix_path(pkg):
     spack_built = []
     externals = []
 
-    env_mod_list = modifications_from_dependencies(pkg.spec, 'build', custom_mods_only=True)
-    env_mod = {}
-    env_mod_list.apply_modifications(env_mod)
-    extra_prefix_paths = env_mod.get("CMAKE_PREFIX_PATH", "").split(";")
-
     # modifications_from_dependencies updates CMAKE_PREFIX_PATH by first
     # prepending all externals and then all non-externals
     for dspec in pkg.spec.traverse(root=False, order="post"):
         if dspec in build_link_deps:
             if dspec.external:
-                externals.insert(0, dspec)
+                externals.insert(0, dspec.prefix)
             else:
-                spack_built.insert(0, dspec)
+                spack_built.insert(0, dspec.prefix)
 
-    ordered_build_link_deps = spack_built + externals
-    build_link_prefixes = filter_system_paths(
-        x.prefix for x in ordered_build_link_deps)
-    return build_link_prefixes + extra_prefix_paths
+    # CMAKE_PREFIX_PATH here cannot come from the outside environment under
+    # normal circumstances, it has to come from spack.  include these second
+    # after the spack_built dependencies but before externals
+    env_mod_list = modifications_from_dependencies(pkg.spec, 'build',
+                                                   custom_mods_only=True)
+    env_mod = {}
+    env_mod_list.apply_modifications(env_mod)
+    extra_prefix_paths = env_mod.get("CMAKE_PREFIX_PATH", "").split(";")
+
+    ordered_build_link_deps = spack_built + extra_prefix_paths + externals
+    build_link_prefixes = filter_system_paths(x for x in
+                                              ordered_build_link_deps)
+    return build_link_prefixes
 
 
 def _setup_pkg_and_run(
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -1019,8 +1019,8 @@ def get_cmake_prefix_path(pkg):
     The goal of the order here is put everything provided by spack before everything
     that comes from externals provided to spack.
     """
-    build_deps = set(pkg.spec.dependencies(deptype=('build', 'test')))
-    link_deps = set(pkg.spec.traverse(root=False, deptype=('link')))
+    build_deps = set(pkg.spec.dependencies(deptype=("build", "test")))
+    link_deps = set(pkg.spec.traverse(root=False, deptype=("link")))
     build_link_deps = build_deps | link_deps
     spack_built = []
     externals = []
@@ -1045,16 +1045,15 @@ def get_cmake_prefix_path(pkg):
     # Since we are going to explicitly pass the CMAKE_PREFIX_PATH as a command line
     # argument (to make sure we include the spack prefixs), passing these via the
     # environment won't occur because CMake prefers the CLI argument
-    env_mod_list = modifications_from_dependencies(pkg.spec, 'build',
-                                                   custom_mods_only=True,
-                                                   set_package_py_globals=False)
+    env_mod_list = modifications_from_dependencies(
+        pkg.spec, "build", custom_mods_only=True, set_package_py_globals=False
+    )
     env_mod = {}
     env_mod_list.apply_modifications(env_mod)
     extra_prefix_paths = env_mod.get("CMAKE_PREFIX_PATH", "").split(";")
 
     ordered_build_link_deps = spack_built + extra_prefix_paths + externals
-    build_link_prefixes = filter_system_paths(x for x in
-                                              ordered_build_link_deps)
+    build_link_prefixes = filter_system_paths(x for x in ordered_build_link_deps)
     return build_link_prefixes
 
 
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -1033,7 +1033,12 @@ def get_cmake_prefix_path(pkg):
                 externals.insert(0, dspec.prefix)
             else:
                 spack_built.insert(0, dspec.prefix)
-
+    # TODO: for the given arguments, modifications_from_dependencies has no
+    # side effects *other than* setup_dependent_package, which is expected
+    # to be idempotent (e.g. setting attributes on the module twice is not
+    # a problem). Currently the API doesn't strictly prevent that though, so
+    # it should be possible to all modifications_from_dependencies and ensure
+    # that there are *no* side effects.
     # CMAKE_PREFIX_PATH here cannot come from the outside spack under
     # normal circumstances, it has to come from per-package implementations of
     # `setup_build_environment` which modify the shell environment
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -1020,6 +1020,12 @@ def get_cmake_prefix_path(pkg):
     build_link_deps = build_deps | link_deps
     spack_built = []
     externals = []
+
+    env_mod_list = modifications_from_dependencies(pkg.spec, 'build', custom_mods_only=True)
+    env_mod = {}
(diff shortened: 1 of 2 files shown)


lib/spack/spack/build_environment.py
```diff
@@ -1034,7 +1040,7 @@ def get_cmake_prefix_path(pkg):
     ordered_build_link_deps = spack_built + externals
     build_link_prefixes = filter_system_paths(
         x.prefix for x in ordered_build_link_deps)
-    return build_link_prefixes
+    return build_link_prefixes + extra_prefix_paths
```
scheibelp: (EDIT: I originally said I could look at this Wednesday. I need to push that back a day)

I'm concerned that this isn't following the order promoted by the comment, but I don't immediately have any good suggestions for how to restore that (in a way that doesn't duplicate code - I think it could be accomplished with some refactoring). I won't have time to make suggestions about that until Wednesday.

Also either way, if this were taken as-is, the comment at the beginning of the method is no-longer accurate. I wrote that (and this) so I would also want to look into why I avoided doing that in the first place.

robertu94: @scheibelp I figured this proposed change would take some time to figure out: I'm assuming we stopped considering this variable for a reason (given the comment), but I looked at the commit it came from and it wasn't obvious to me.

I'll clean up the style issue tomorrow, and make the comment make sense at the very least.

scheibelp: To address my original concern:

> Also either way, if this were taken as-is, the comment at the beginning of the method is no-longer accurate. I wrote that (and this) so I would also want to look into why I avoided doing that in the first place.

It seems like the original implementation of `get_cmake_prefix_path` also had this feature when I wrote https://github.com/spack/spack/commit/916cdfbb56b2abfc380ce1b4cac0d6494b7f2628, so I think that comment was just saying I was also deciding not to handle it here.

lib/spack/spack/build_environment.py
```diff
@@ -1014,26 +1014,33 @@ def add_modifications_for_dep(dep):
...
-                spack_built.insert(0, dspec)
-
-    ordered_build_link_deps = spack_built + externals
-    build_link_prefixes = filter_system_paths(
-        x.prefix for x in ordered_build_link_deps)
+                spack_built.insert(0, dspec.prefix)
+
+    # CMAKE_PREFIX_PATH here cannot come from the outside environment under
+    # normal circumstances, it has to come from spack.  include these second
+    # after the spack_built dependencies but before externals
```
scheibelp: I think this is a good PR-level commentary on the reason for these changes, but in the context of the code I think the following would be clearer:

> Collect CMAKE_PREFIX_PATH shell variable modifications from per-package implementations of `setup_build_environment`; these can then be passed explicitly to the cmake invocation.

do you agree?

lib/spack/spack/build_environment.py
```diff
@@ -1014,26 +1014,33 @@ def add_modifications_for_dep(dep):
...
+    env_mod_list = modifications_from_dependencies(pkg.spec, 'build',
+                                                   custom_mods_only=True)
```
scheibelp: This should also set `set_package_py_globals=False`.

With that, and if `modifications_for_dep` skipped `dpkg.setup_dependent_package(spec.package.module, spec)`, this would have no side effects (but I'd need to think about what to suggest to skip the latter - maybe another parameter? That might be confusing for the interface).

lib/spack/spack/build_environment.py
```diff
@@ -1014,26 +1014,33 @@ def add_modifications_for_dep(dep):
...
+    env_mod = {}
+    env_mod_list.apply_modifications(env_mod)
+    extra_prefix_paths = env_mod.get("CMAKE_PREFIX_PATH", "").split(";")
+
+    ordered_build_link_deps = spack_built + extra_prefix_paths + externals
```
scheibelp, line 1055: This generally preserves the desired order, although I don't know if I have a concise way to describe what exactly that order is. My best summary is:

* Spack-built prefixes (i.e. the install prefix of each Spack-built package)
* Custom modifications from the DAG, with priority given to spack-built-packages (this is nebulous since those methods could prepend, append, or even override entirely, so It might be more accurate that spack-built packages have the "last say")
* External prefixes

Overall I agree that is fine. If you think it is useful, then perhaps it's worth adding a comment describing this ordering to the method's docstring.

scheibelp (changes requested):
This overall looks good: I have a couple minor requests about comments. I have to think more about what to suggest regarding my comment about the side effects of `modifications_for_dep` (ideally running that function to collect CMAKE_PREFIX_PATH modifications wouldn't also do `dpkg.setup_dependent_package(spec.package.module, spec)`)

lib/spack/spack/build_environment.py
robertu94: @scheibelp this will be my first (second? not sure) contribution to spack core. Other than adding the argument you suggested I don't have context to understand what you want me to do here.

lib/spack/spack/build_environment.py
robertu94: I'll add a comment soon

lib/spack/spack/build_environment.py
scheibelp: Sorry yes I need to think on it: my first naive suggestion is that we would need another parameter like `set_package_py_globals` which (when set `=False`) would skip `dpkg.setup_dependent_package(spec.package.module, spec)`, but I'm thinking that having so many parameters to control small parts of the function's behavior will make it confusing to understand, so I'm thinking if I should work on refactoring it (which I could as a commit here if that's alright with you, but I wouldn't have time to do that until next week).

robertu94:
@scheibelp Sorry for the delay on this.  I have pushed changes to the comments, and added a doc string.  Please let me know if you want any more changes.

robertu94:
@scheibelp ping

robertu94:
@scheibelp This is blocking a fix for cuSZ being submitted. Is there a timeline on this being merged?

lib/spack/spack/build_environment.py
```diff
@@ -1014,26 +1014,44 @@ def add_modifications_for_dep(dep):
...
-                externals.insert(0, dspec)
+                externals.insert(0, dspec.prefix)
             else:
-                spack_built.insert(0, dspec)
-
-    ordered_build_link_deps = spack_built + externals
-    build_link_prefixes = filter_system_paths(
-        x.prefix for x in ordered_build_link_deps)
+                spack_built.insert(0, dspec.prefix)
+
```
scheibelp:
```suggestion
    # TODO: for the given arguments, modifications_from_dependencies has no
    # side effects *other than* setup_dependent_package, which is expected
    # to be idempotent (e.g. setting attributes on the module twice is not
    # a problem). Currently the API doesn't strictly prevent that though, so
    # it should be possible to all modifications_from_dependencies and ensure
    # that there are *no* side effects.
```

scheibelp (changes requested):
I looked at this more and the side effects I am concerned about in https://github.com/spack/spack/pull/31631#discussion_r927170914 should not be a problem. But I still think a refactor is in order, so I have added a suggestion for a TODO comment here. If you choose to commit it and the tests pass (assuming I formatted this right in the Github UI) then I'll merge it.

robertu94:
@scheibelp i can’t tell from the build log if that failure was spurious. However since it depends on Cuda, I don’t want to take a chance. How can we debug this failure?

scheibelp:
> i can’t tell from the build log if that failure was spurious. However since it depends on Cuda, I don’t want to take a chance. How can we debug this failure?

It looks like the log for the failed package (mfem) is truncated - I'm checking if there's a way we can extend it.

I notice a couple of files are out of date - apologies for not noticing that before but FYI this will need to merge-from/rebase-on develop as well. It's possible that the errors will be resolved (if not though, we'll want to get the full log).

scheibelp:
I think I know how to get the full logs, but I suspect that since the latest run was a long time ago that they were discarded, so if the problem comes up again after a merge/rebase, I can see where the mfem build went wrong.

scheibelp:
(note you can run `spack style --fix` locally or the spackbot can do that here)

robertu94:
@spackbot fix style

robertu94:
@scheibelp at least some of these failures seem to have come from other parts of spack not change by the PR that couldn't have been caused by this PR (i.e. undefined variables in code that I didn't change).  What should I do next?

scheibelp:
@robertu94 unfortunately after some investigation this error looks legitimate:

* Most directly, it occurs because of my request to set `set_package_py_globals=False`
  * In more detail: our tests make sure that module globals are set when invoking functions like `setup_build_environment`
  * That function is used to compute the `cmake_prefix_path`
  * So for the time being in order to set variables like that, we must apply the module modifications
* I don't think undoing that is the right approach (at least without refactoring the surrounding code)
* I need to think a bit more on this, have an internal discussion on implied interfaces, and then circle back (ultimately that will likely involve me pushing additional changes to the PR)

That process will take at least a few days

robertu94:
@scheibelp following up to see what the status is on this.

scheibelp:
@robertu94 I spoke with @becker33 today and he proposed something I thought might address your needs in a simpler manner: could you let me know if https://github.com/spack/spack/pull/32547 would be an acceptable approach for you?

PR #31632, opened 2022-07-18
Add Caliper v2.8.0 and rocm support

daboehme:
Add Caliper v2.8.0. Also adds new optional rocm dependency, and some cleanup (use `define_from_variant`).


tldahlgren (commented):
Following up on verbal, if you don't want or don't support building with `+cuda+rocm`, then you want to add a conflict.

tldahlgren (approved):
LGTM

PR #31637, opened 2022-07-19
[py-numpy] Newer versions do not build with `icc`

stephenmsachs:
Workaround until https://github.com/numpy/numpy/issues/22011 is solved.

Flagging maintainers @adamjstewart , @rgommers


alalazo:
@spackbot maintainers

adamjstewart:
LGTM, will let @rgommers comment on whether there is a quick patch we can apply to fix this or whether we should wait for a new release.

rgommers:
There's no patch yet, but I expect it will be solved fairly soon (`1.23.1` or `1.23.2`). It's due to the introduction of C++ in NumPy in 1.23.0, and there's a bunch of issues related to that which need to be addressed.

rgommers (approved):
LGTM

adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

PR #31643, opened 2022-07-19
environment.py: reduce # locks further

haampie:
Small refactor of accumulated hacks, reducing the risk of timeouts due to locks further.

Note that previously `[s for s in all_specs if s.installed]` was not wrapped in a read lock.


haampie:
@spackbot run pipeline

haampie:
ping @becker33

becker33 (approved):

PR #31645, opened 2022-07-19
environment.py: only install root specs explicitly

haampie:
When doing a partial environment install (e.g. installing an individual package
`spack -e . install --no-add /<hash>`) spack marks this package as explicit.

This is a problem for `spack env depfile -o Makefile`, which generates a `spack`
`-e . install --no-add /<hash>` command for every spec in the environment, meaning one
ends up with all specs installed explicitly, which is undesirable.

Note that whenever `spack -e . install new-spec` is executed and `new-spec` is
not yet in the environment, it is added as a root spec, and therefore it will
still be marked explicit.

So, the environment install semantics are: root spec implies explicit,
and if you need a dependency to be installed explicitly, then you should add it
to the root specs and rely on `concretizer:unify:true`.

Requires #31646, cause installing a single package implicitly is broken right now.

### DIFF
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -131,8 +131,9 @@ which is useful for CI pipeline troubleshooting""")
         help='(with environment) only install already concretized specs')
     subparser.add_argument(
         '--no-add', action='store_true', default=False,
-        help="""(with environment) only install specs provided as argument
-if they are already in the concretized environment""")
+        help="""(with environment) partially install an environment, limiting
+to concrete specs in the environment matching the arguments.
+Non-roots remain installed implicitly.""")
     subparser.add_argument(
         '-f', '--file', action='append', default=[],
         dest='specfiles', metavar='SPEC_YAML_FILE',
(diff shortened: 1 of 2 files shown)


haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

becker33:
@haampie I think this is /probably/ the right course of action, but it's non-intuitive with the meaning of implicit/explicit, so it probably requires good documentation for users.

haampie:
We could document that `spack install x y z` without environment is equivalent to `spack -e env install` with environment containing `specs: [x, y, z]` and `concretizer:unify:false`?

Then it would be expected that root specs are the "explicit specs".

becker33:
I think we need to document that

```
spack -e env add hdf5
spack -e env concretize
spack -e env install zlib
```
does not mark zlib as explicit. That's kinda weird behavior.

haampie:
This should still mark it as explicit (with or without this PR), because zlib is added as a root spec. It's just when you do spack -e env install --no-add zlib it is not marked explicit

haampie:
Anything that needs to be documented then @becker33?

becker33 (dismissed):

becker33 (approved):

PR #31650, opened 2022-07-19
VisIt: Allow osmesa and gui.

brugger1:
I modified the VisIt spack package to allow "osmesa" with "gui". This is how VisIt is normally built. When it is built this way rendering is done in one of 2 ways.

1) The "gui" (or client) uses OpenGL for local rendering.
2) The compute engine (or server) uses osmesa for rendering. This is typically done in parallel, where each process renders a portion of the image and then the images are composited together into a single image. The composited image is then sent back to the "gui", which displays the image.

I made this way of building the default. I removed the "conflicts" between "osmesa" and "gui". I also modified it to always use "glu" when building on Linux systems rather than sometimes "glu" and other times "mesa-glu".

I tested this on "crusher.olcf.ornl.gov", an early access system for the Frontier exascale system at Oak Ridge National Lab. I successfully displayed images in the GUI and was able to save an offscreen generated image.

### DIFF
--- a/var/spack/repos/builtin/packages/visit/package.py
+++ b/var/spack/repos/builtin/packages/visit/package.py
@@ -78,7 +78,7 @@ class Visit(CMakePackage):
     generator = "Ninja"
 
     variant('gui',    default=True, description='Enable VisIt\'s GUI')
-    variant('osmesa', default=False, description='Use OSMesa for off-screen CPU rendering')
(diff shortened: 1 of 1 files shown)


brugger1:
The `ci/gitlab-ci` pipeline has been in "Pending" state for 2 days. Also no reviewers have been assigned.

brugger1:
@spackbot run pipeline

brugger1:
Out of date. Will submit a new request.

PR #31653, opened 2022-07-19
Fix r-rhtslib

glennpj:
There is a problem with the git repo for rhtslib that apparently led to
a bad version entry during the previous round of package updates. A git
checkout on the commit also fails so use branch for most recent version.


glennpj:
@spackbot fix style

alalazo (approved):
Do you know what is causing the isuue?

PR #31654, opened 2022-07-19
VisIt: Update to VisIt 3.3.0.

brugger1:
Updated the VisIt spack package for VisIt 3.3.0.

I restricted an existing patch since it was no longer needed with 3.3.0.
I added a 3.3.0 and later variant of the parallel-hdf5 patch and made the parallel-hdf5 patch only apply to versions prior to 3.3.0 since the code changed slightly.

I tested this on "crusher.olcf.ornl.gov", an early access system for the Frontier exascale system at Oak Ridge National Lab. I successfully displayed images in the GUI.

### DIFF
--- a/var/spack/repos/builtin/packages/visit/package.py
+++ b/var/spack/repos/builtin/packages/visit/package.py
@@ -69,6 +69,7 @@ class Visit(CMakePackage):
     executables = ['^visit$']
 
     version('develop', branch='develop')
+    version('3.3.0', sha256='1a7485146133ac5f1e330d9029697750046ef8d9e9de23a6c2a3685c1c5f4aac')
     version('3.2.2', sha256='d19ac24c622a3bc0a71bc9cd6e5c9860e43f39e3279672129278b6ebce8d0ead')
     version('3.2.1', sha256='779d59564c63f31fcbfeff24b14ddd6ac941b3bb7d671d31765a770d193f02e8')
     version('3.1.1', sha256='0b60ac52fd00aff3cf212a310e36e32e13ae3ca0ddd1ea3f54f75e4d9b6c6cf0')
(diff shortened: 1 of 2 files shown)


cyrush (approved):

brugger1:
The `ci/gitlab-ci` pipeline has been in "Pending" state for 2 days. Looking at the details for that test, it looks like everything was successful.

brugger1:
@cyrush The ci/gitlab-ci test has been in Pending for 3 days. Looking at the details it looks like it successfully finished. Thoughts on what to do?

iarspider:
@spackbot run pipeline

tldahlgren (approved):
Confirmed sha256, rest LGTM

haampie:
@brugger1 somehow the visit build is broken: https://github.com/spack/spack/commit/6c0ca932aa7064d57737ccc5327bc12c87017546. No clue why (a) it is rebuild only there and (b) it did not fail before. Maybe interesting to @scottwittenburg.

From the logs:

```
lib/libavtquery_ser.so: undefined reference to `shm_open'
```
Either visit or one of its deps is missing a dependency on `librt`?

scottwittenburg:
Just noticing that the `visit` that is currently broken in develop (`visit/j4zxcr4`) is different from the one that built successfully in the most recent pipeline of this PR (`visit/uqtwn63`).  That just makes me think this PR didn't break it, but maybe a change to a dependency?  I'll see if I can find other pipelines where the broken sha appears.

cyrush:
Yes, there is a bug in VisIt's build system related to new features in 3.3.0:
https://github.com/visit-dav/visit/issues/17569

I do I think its just missing link to librt.  We hope to fix in 3.3.1.

tgamblin:
I think we should deprecate `3.3.0` so that `3.2.2` is preferred in `develop` until the fix is in -- any objections?

scottwittenburg:
If no objections, the PR is [here](https://github.com/spack/spack/pull/31790).

cyrush:
I am fine with changing the default version until there is a fix. 

Note: I think this issue only happens with some (newer compilers) and the gitlab ci pipeline did fully pass 3 days ago. 

It would be good to understand why the pipeline passed then, but not now.  Does the problem happen at runtime and that is why it wasn't seen?

scottwittenburg:
The version of visit that passed 3 days ago has a different hash than the one that is failing now.  I don't yet understand how that broken hash was never built in a PR pipeline before appearing in a develop pipeline, but I haven't had a chance to investigate to the bottom of that yet.

PR #29030, opened 2022-02-17
llvm: fix build with Fujitsu compiler, and conflicts gcc@8.4 on aarch64

tkameyama:
- Add -lpthread to build OpenMP libraries with Fujitsu compiler
-  avoid build failed with Fujitsu compiler
- Add conflicts gcc@8.4 on aarch64 https://bugzilla.redhat.com/show_bug.cgi?id=1958295

PR #31661, opened 2022-07-20
Use `llvm13-thread.patch` for all compilers when building llvm

msimberg:
I was getting undefined references to `pthread` functions without enabling the patch that was introduced in https://github.com/spack/spack/pull/29030 for the fujitsu compiler. This happened with both `llvm@13` and `llvm@14`.

Marking this as a draft since there are some open questions. I've preemptively enabled the patch for all compilers now since I would suspect there not to be any harm in doing so, but I may be wrong. Do you think it's necessary to be conservative and only enable the patch additionally for the two cases that I saw fail, i.e. `@13:14 %gcc`? @tkameyama have you submitted the patch (`llvm13-thread.patch`) upstream to LLVM? In either case, I will still need to update comments and/or patch names based the answers.


albestro:
Independently from @msimberg I had the same problem and I arrived at the same ~conclusion~ workaround. I'm also building with GCC, both `llvm@13 % gcc` (last May, don't remember gcc version) and `llvm@14 % gcc@11.2.0` (now).

trws:
Have either of you tested with the 15 rc by any chance? Leaving it completely open ended as a patch isn't something I like to do if we can help it.

msimberg:
> Have either of you tested with the 15 rc by any chance? Leaving it completely open ended as a patch isn't something I like to do if we can help it.

I have not, but can do so. I'll let you know what happens.

msimberg:
> Have either of you tested with the 15 rc by any chance? Leaving it completely open ended as a patch isn't something I like to do if we can help it.

First try with GCC 11.2.0 on a Cray system (i.e. with Cray's GCC) ends up in GCC segfaulting, both on `llvm@main` and `llvm@15.0.0-rc2`. We're going to try a few different GCCs to see if we can pin down the problem a bit more precisely.

msimberg:
So, `llvm@main` does still need _some_ patch. The build doesn't segfault with a vanilla GCC, but fails with the same problem as 13 and 14 have while building OpenMP without a patch. However, the patch actually doesn't even apply anymore on `main` and 15, so I've updated this PR to reflect that. We can figure out the required patches for `15:` in a separate PR.

trws:
Excellent, thank you @msimberg!

trws (approved):

PR #31665, opened 2022-07-20
mark external as explicit only when installing explicitly

haampie:


alalazo (commented):
LGTM, but can we add a test to avoid regression in the future?

haampie:
Actually adding the test revealed yet another issue where the spec was simply not registered in the database if not explicit.

alalazo (approved):

PR #31591, opened 2022-07-15
Rocm 5.2 release integration

renjithravindrankannath:
This PR contains changes for ROCM 5.2.0 release.
Additional changes for hipify-clang and rocm-openmp-extras will be under different PR

PR #31667, opened 2022-07-20
Changes for ROCm-5.2.0 changes and new recipe rocwmma

srekolam:
Changes to rocm-5.2.0 version and recipe rocwmma.


srekolam:
@estewart08  , @cgmb  (for rocwmma ), let me know if you have any comments.

srekolam:
this PR is dependent on #31591 . I believe the failure with linux tests is due to that

var/spack/repos/builtin/packages/rocwmma/package.py
```diff
@@ -0,0 +1,77 @@
...
+    # releases
+    conflicts('amdgpu_target=gfx906:xnack-', when='@5.2.0')
+    conflicts('amdgpu_target=gfx908', when='@5.2.0')
+    conflicts('amdgpu_target=gfx1010', when='@5.2.0')
+    conflicts('amdgpu_target=gfx1011', when='@5.2.0')
+    conflicts('amdgpu_target=gfx1012', when='@5.2.0')
+    conflicts('amdgpu_target=gfx1030', when='@5.2.0')
+    conflicts('amdgpu_target=gfx1031', when='@5.2.0')
+
+    depends_on('cmake@3.16.8:', type='build', when='@5.2.0:')
```
cgmb: As far as I can tell, there's no good reason for any of the ROCm mathlibs to require `3.16.8`. That just happened to be the version that was on the CI server and some component owners made it the minimum requirement because that was the version they used for testing. However, I've built pretty much the whole stack with 3.16.3 (the system version on Ubuntu 20.04) and---after changing the `cmake_minimum_required` statements---they have all worked fine.

If rocWMMA needs something higher than its stated requirement of CMake 3.5 and you don't want to bother checking experimentally, maybe just specify `3.16:`. It's nicer for the folks on Ubuntu 20.04 if they have the option to use the external CMake.

var/spack/repos/builtin/packages/rocwmma/package.py
```diff
@@ -0,0 +1,77 @@
...
+    amdgpu_targets = ROCmPackage.amdgpu_targets
+
+    variant('amdgpu_target', values=auto_or_any_combination_of(*amdgpu_targets))
+    variant('build_type', default='Release', values=("Release", "Debug", "RelWithDebInfo"), description='CMake build type')
+
+    # gfx908:xnack-;gfx90a:xnack-;gfx90a:xnack+
+    # are only targets currently supported for @5.2.0
+    # releases
+    conflicts('amdgpu_target=gfx906:xnack-', when='@5.2.0')
+    conflicts('amdgpu_target=gfx908', when='@5.2.0')
```
cgmb: I don't think there's any problem with rocWMMA on gfx908.

var/spack/repos/builtin/packages/rocwmma/package.py
```diff
@@ -0,0 +1,77 @@
...
+    homepage = "https://github.com/ROCmSoftwarePlatform/rocWMMA"
+    git = "https://github.com/ROCmSoftwarePlatform/rocWMMA.git"
+    url      = "https://github.com/ROCmSoftwarePlatform/rocWMMA/archive/refs/tags/rocm-5.2.0.tar.gz"
+    tags     = ['rocm']
+
+    maintainers = ['srekolam']
+
+    version('5.2.0', sha256='257ccd1cf2bc1d8064e72e78d276ef7446b2cb7e2dec05ff8331bb44eff2b7cb')
+
+    amdgpu_targets = ROCmPackage.amdgpu_targets
```
cgmb: rocWMMA exists to help libraries use specialized instructions available only on MI100 (gfx908) and MI200 (gfx90a) hardware. This makes it a bit different from the mathlibs, which could reasonably be hardware agnostic.

I think in this case, it's fine to maintain a separate list of supported amdgpu_targets rather than generating a big list of `conflicts`. Maybe `'gfx908', 'gfx908:xnack-', 'gfx908:xnack+', 'gfx90a', 'gfx90a:xnack-', 'gfx90a:xnack+'`

var/spack/repos/builtin/packages/rocwmma/package.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def setup_build_environment(self, env):
+        env.set('CXX', self.spec['hip'].hipcc)
+
+    def cmake_args(self):
+        args = [
+            self.define('ROCWMMA_BUILD_TESTS', 'ON'),
+            self.define('ROCWMMA_BUILD_VALIDATION_TESTS', 'ON'),
+            self.define('ROCWMMA_BUILD_BENCHMARK_TESTS', 'ON'),
+            self.define('ROCWMMA_BUILD_SAMPLES', 'ON'),
```
cgmb: These are usually `OFF` by default for other libraries?

cgmb (commented):
I will be out on parental leave for the next few weeks, so I won't have much time to review. I left some comments. I hope you find them helpful!

var/spack/repos/builtin/packages/rocwmma/package.py
cgmb: (Incidentally, the version of CMake used on the ROCm CI is no longer 3.16.8 anyway, so the original justification is no longer even valid.)

var/spack/repos/builtin/packages/rocwmma/package.py
```diff
@@ -0,0 +1,77 @@
...
+    depends_on('cmake@3.16.8:', type='build', when='@5.2.0:')
+    depends_on('cmake@3.5:', type='build')
+
+    depends_on('googletest@1.10.0:', type='test')
+
+    for ver in ['5.2.0']:
+        depends_on('rocm-cmake@%s:' % ver, type='build', when='@' + ver)
+        depends_on('llvm-amdgpu@' + ver,  type='build',  when='@' + ver)
+        depends_on('hip@' + ver,                         when='@' + ver)
+        depends_on('rocblas@' + ver,  type='build',  when='@' + ver)
```
cgmb: You should propagate the amdgpu_target to the rocblas dependency. The rocsolver package is a good example of how to do that.

srekolam:
@estewart08  , can you review the rocm-openmp-extras

estewart08:
Looking at it more closely now. I did not receive a build error, but some bitcode files seemed to be installed in a different area which the compiler did not like.

estewart08:
The rocm-openmp-extras recipe LGTM. I noted offline that the llvm-amdgpu package would need an additional patch.

srekolam:
@cgmb can you provide any review comments ?

cgmb (dismissed):
Assuming that you've tested that it builds successfully, rocWMMA looks reasonable to me.

srekolam:
@cgmb i built it. for testing it is dependent on mi-100/200's .. i will plan to get it verified on that h/w

srekolam:
@haampie  , please let me know if you have any review comments

srekolam:
can i get some review comments

kwryankrattiger:
After a quick look, this seems okay to me. I would like to have this merged sooner rather than later as this is quite important for the DaV SDK and other packages in E4S (@eugeneswalker @wspear).

tldahlgren (approved):
Confirmed (most) new sha256.  LGTM.

PR #31672, opened 2022-07-21
py-torch: use immutable URLs for open-ce patches

nicholas-sly:
Also update magma dependency specs to be more explicit.

Previous patches were pointed at the main branch of the open-ce repository, which reorganized their directory layout and broke the links to these patches.  Also added a new patch for 1.12.0 (which was part of the reorganization).  The 1.12.0 build has been tested on Lassen at LLNL.

### DIFF
--- a/var/spack/repos/builtin/packages/py-torch/package.py
+++ b/var/spack/repos/builtin/packages/py-torch/package.py
@@ -156,7 +156,8 @@ class PyTorch(PythonPackage, CudaPackage):
     depends_on('cudnn@6:7', when='@:1.0+cudnn')
     depends_on('cudnn@7.0:7', when='@1.1:1.5+cudnn')
     depends_on('cudnn@7:', when='@1.6:+cudnn')
-    depends_on('magma', when='+magma')
+    depends_on('magma+cuda~rocm', when='+magma+cuda')
+    depends_on('magma+rocm~cuda', when='+magma+rocm')
     depends_on('nccl', when='+nccl')
     depends_on('numactl', when='+numa')
     depends_on('llvm-openmp', when='%apple-clang +openmp')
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-torch/package.py
```diff
@@ -253,17 +254,20 @@ class PyTorch(PythonPackage, CudaPackage):
 
     # Use patches from IBM's Open CE to enable building on Power systems
     # 03xx - patch temporary to fix a problem that when fixed upstream can be removed
-    patch('https://github.com/open-ce/pytorch-feedstock/raw/main/recipe/0302-cpp-extension.patch',
-          sha256='ecb3973fa7d0f4c8f8ae40433f3ca5622d730a7b16f6cb63325d1e95baff8aa2', when='@1.10.0: arch=ppc64le:')
+    patch('https://raw.githubusercontent.com/open-ce/pytorch-feedstock/0a145060ca8523314ec3893af935c3b140e2d0b0/pytorch-1.10/recipe/0302-cpp-extension.patch',
```
adamjstewart: These are some long URLs. We could replace "raw.githubusercontent.com" with "github.com". We could also use a release tag instead of a specific commit, although it is possible to change the tag (just uncommon).

var/spack/repos/builtin/packages/py-torch/package.py
```diff
@@ -253,17 +254,20 @@ class PyTorch(PythonPackage, CudaPackage):
...
+          sha256='ecb3973fa7d0f4c8f8ae40433f3ca5622d730a7b16f6cb63325d1e95baff8aa2', when='@1.10.0:1.11.0 arch=ppc64le:')
```
adamjstewart: This probably isn't going to be fixed in 1.11.1 (if it ever comes out). Would be safer to use `when='@1.10:1.11 arch=...'`

var/spack/repos/builtin/packages/py-torch/package.py
```diff
@@ -156,7 +156,8 @@ class PyTorch(PythonPackage, CudaPackage):
     depends_on('cudnn@6:7', when='@:1.0+cudnn')
     depends_on('cudnn@7.0:7', when='@1.1:1.5+cudnn')
     depends_on('cudnn@7:', when='@1.6:+cudnn')
-    depends_on('magma', when='+magma')
+    depends_on('magma+cuda~rocm', when='+magma+cuda')
+    depends_on('magma+rocm~cuda', when='+magma+rocm')
```
adamjstewart: cuda and rocm are mutually exclusive, so you can simplify this:
```suggestion
    depends_on('magma+cuda', when='+magma+cuda')
    depends_on('magma+rocm', when='+magma+rocm')
```

nicholas-sly: I thought so as well, but could swear Spack tried to build rocm on Lassen, despite it not really being appropriate.  I added the extra explicit removal of those conflicting variants just to be sure.

adamjstewart: If you can reproduce that, open a bug report. It might choose the wrong default, but it shouldn't be possible to concretize `+rocm+cuda` for magma or PyTorch

var/spack/repos/builtin/packages/py-torch/package.py
nicholas-sly: Unfortunately, the latest tag doesn't contain these changes, so there isn't a tag available that could do that.  If you can swap out 'raw.githubusercontent' for 'github' and it will guarantee the hash will remain unchanged, I'm fine with swapping them out.

adamjstewart: The latter maps to the former (you also need to add `/raw/` later in the url):
```console
$ wget https://github.com/open-ce/pytorch-feedstock/raw/0a145060ca8523314ec3893af935c3b140e2d0b0/pytorch-1.10/recipe/0302-cpp-extension.patch
--2022-07-21 16:01:25--  https://github.com/open-ce/pytorch-feedstock/raw/0a145060ca8523314ec3893af935c3b140e2d0b0/pytorch-1.10/recipe/0302-cpp-extension.patch
Resolving github.com... 64:ff9b::c01e:ff70, 192.30.255.112
Connecting to github.com|64:ff9b::c01e:ff70|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://raw.githubusercontent.com/open-ce/pytorch-feedstock/0a145060ca8523314ec3893af935c3b140e2d0b0/pytorch-1.10/recipe/0302-cpp-extension.patch [following]
--2022-07-21 16:01:25--  https://raw.githubusercontent.com/open-ce/pytorch-feedstock/0a145060ca8523314ec3893af935c3b140e2d0b0/pytorch-1.10/recipe/0302-cpp-extension.patch
Resolving raw.githubusercontent.com... 2606:50c0:8002::154, 2606:50c0:8001::154, 2606:50c0:8000::154, ...
Connecting to raw.githubusercontent.com|2606:50c0:8002::154|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 3200 (3.1K) [text/plain]
Saving to: '0302-cpp-extension.patch'

0302-cpp-extension.patch                100%[============================================================================>]   3.12K  --.-KB/s    in 0s      

2022-07-21 16:01:25 (10.0 MB/s) - '0302-cpp-extension.patch' saved [3200/3200]
```

var/spack/repos/builtin/packages/py-torch/package.py
nicholas-sly: That's fair.  I was really just gating it from affecting 1.12.0, which has a separate patch for the same thing, but with a different hash.

nicholas-sly:
@adamjstewart I believe I have addressed all of the comments.  Let me know if you find anything else that should be changed.

adamjstewart (approved):

PR #31679, opened 2022-07-21
coreutils: adding fix for compat with older glibcs and gcc11+

snehring:
The combination of a really new compiler on an aging OS causes build errors out of the box with coreutils. This just adds in the -std=gnu11 compiler flag if gcc11+ is used. 

Annoyingly the gnu extensions do seem to be required here, so I couldn't use the nice compiler property.

### DIFF
--- a/var/spack/repos/builtin/packages/coreutils/package.py
+++ b/var/spack/repos/builtin/packages/coreutils/package.py
@@ -35,6 +35,17 @@ class Coreutils(AutotoolsPackage, GNUMirrorPackage):
 
     build_directory = 'spack-build'
 
+    def flag_handler(self, name, flags):
+        # gcc11+ defaults to C17 which casues issues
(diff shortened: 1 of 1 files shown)


haampie:
The build system should add the right `-std=...` flag, since they know what language it is written in, this is what their configure script checks:

```
checking for /spack/lib/spack/env/gcc/gcc option to enable C11 features... none needed
```

So I don't think this is the right approach?

The issue is probably this: https://gitlab.com/cjwatson/man-db/-/issues/6? There is a bug in `gnulib` for the particular GCC and glibc combination. `gnulib` is universally vendored, so it's a pain :|. But you can in principle add the patch here: https://github.com/coreutils/gnulib/commit/84863a1c4dc8cca8fb0f6f670f67779cdd2d543b

haampie:
Closing in favor of #31693, but added you to the co-authors in the commit :upside_down_face:

PR #31685, opened 2022-07-21
Perl: set PERL5LIB for test deps

adamjstewart:
We set `PATH` and `PYTHONPATH` for test deps, we were missing `PERL5LIB`.


adamjstewart:
@spackbot run pipeline

alalazo (approved):

PR #31688, opened 2022-07-21
Added cmake3 binary name to search for external

OliverPerks:
Support for search for 'cmake' and 'cmake3' (as we see it in yum based OSs)


var/spack/repos/builtin/packages/cmake/package.py
```diff
@@ -24,7 +24,7 @@ class Cmake(Package):
 
     tags = ['build-tools']
 
-    executables = ['^cmake$']
+    executables = ['^cmake$', '^cmake3$']
```
adamjstewart:
```suggestion
    executables = ['^cmake[0-9]*$']
```
Wouldn't this be more robust (e.g., cmake 2.8, 4.0, etc.)?

adamjstewart (dismissed):

adamjstewart:
Can you resolve the merge conflicts?

alalazo:
Ping @OliverPerks

bernhardkaindl (dismissed):
Only needed for RHEL7/CentOS7 because RHEL8 make cmake3 the default. But as RHEL7 is still maintained, lets merge it.

bernhardkaindl:
@spackbut fix style

bernhardkaindl:
@spackbot fix style

bernhardkaindl (approved):
Suggestion by adamsteward was integrated and conflict with develop resolved.

alalazo:
@spackbot run pipeline

adamjstewart (approved):

alalazo:
Force merged, since this one cannot make a pipeline build fail that previously wasn't

Jordan474:
This actually broke bootstrap on rhel7 for me: because it detects `cmake3` as available external, but then the clingo package runs `cmake`, which is cmake 2.

Many other packages could be affected by this when one does `spack external find`.

I guess one could change `Cmake.setup_dependent_package` to set the the `cmake` module function to `bin/cmake3`, but I'm not sure how robust this will be, and if this will cover all cmake usages. I would argue it is safer to ignore `cmake3` and only detect "vanilla" cmake installation `bin/cmake` ?

sebrowne:
> This actually broke bootstrap on rhel7 for me: because it detects `cmake3` as available external, but then the clingo package runs `cmake`, which is cmake 2.
> 
> Many other packages could be affected by this when one does `spack external find`.
> 
> I guess one could change `Cmake.setup_dependent_package` to set the the `cmake` module function to `bin/cmake3`, but I'm not sure how robust this will be, and if this will cover all cmake usages. I would argue it is safer to ignore `cmake3` and only detect "vanilla" cmake installation `bin/cmake` ?

We encountered this as well at our site.

PR #31691, opened 2022-07-22
kim-api: Added paths for bash/zsh completion

hseara:
Without providing specific paths, it was prepending '/etc'
    to the package path ($prefix), which failed the installation.
    
 Without setting the path, it tries to copy the completion scripts
    in '/etc/$prefix' path that it is write-protected to the spack user.


ellio167:
@hseara 
This (prepending of `/etc`) should only happen if `$prefix` is of the form `/opt/...`.  This behavior is intended and conforms to the GNU Coding standard (https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) and the Filesystem Hierarchy Standard (https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s13.html).  

I have no problem with the completion script installation when building via spack in a non `/opt/...` location.  Am I correct in assuming you have your spack installation located somewhere under `/opt/`?  Perhaps this is not a recommended location for spack?  (Maybe spack has a convention/policy for this scenario?)

hseara:
As you mentioned, our spack installation resides in `/opt` with all other shared network infrastructure.
I was not aware of such a corner case. Thanks for sharing this information. We have always used `/opt` for sharing stuff over the network for historical reasons due to the usage of `ohpc`. Changing the directory where the software resides now will be a monumental task ( documentation, user scripts, etc..).

I do not contest that the observed behavior is intended when doing stand-alone installations. However, in the context of spack, it is clear that unless installations within the `/opt` folder are forbidden, the behavior observed contradicts the ultimate aim of spack to be self-contained. My fix will ensure that the package remains self-constrained.

Does my fix result in any side effects beyond not conforming to standards?

ellio167:
Yes, this is a weird (often unknown and/or ignored) corner case.  I cannot find any discussion of this in the spack space.

I agree with you; it is appropriate to change the kim-api install behavior for its spack package.  I'll comment further elsewhere.

var/spack/repos/builtin/packages/kim-api/package.py
```diff
@@ -45,3 +45,11 @@ def patch(self):
         if self.spec.satisfies('%nvhpc'):
             filter_file('-std=gnu', '',
                         'examples/simulators/simulator-model-example/CMakeLists.txt')
+
+    def cmake_args(self):
+        args = [
+            '-DZSH_COMPLETION_COMPLETIONSDIR={0}/zhs_completion.d/'.format(self.prefix),
+            '-DBASH_COMPLETION_COMPLETIONSDIR={0}/bash_completion.d/'.format(
+                self.prefix)
```
ellio167: I think this will be a cleaner solution:

```suggestion
            '-DCMAKE_INSTALL_SYSCONFDIR={0}/etc'.format(self.prefix)
```

BTW, note the typo in the original: `'-DZSH_COMPLETION_COMPLETIONSDIR={0}/zhs_completion.d/'.format(self.prefix),` which should be `'-DZSH_COMPLETION_COMPLETIONSDIR={0}/zsh_completion.d/'.format(self.prefix),` (transposition `zhs` instead of `zsh`)

hseara: I was not sure if SYSCONFDIR will do the job. But if it does, good for me.

ellio167 (approved):

var/spack/repos/builtin/packages/kim-api/package.py
```diff
@@ -45,3 +45,9 @@ def patch(self):
         if self.spec.satisfies('%nvhpc'):
             filter_file('-std=gnu', '',
                         'examples/simulators/simulator-model-example/CMakeLists.txt')
+
+    def cmake_args(self):
+        args = [
+            '-DCMAKE_INSTALL_SYSCONFDIR={0}/etc'.format(self.prefix)
```
alalazo:
```suggestion
            self.define("CMAKE_INSTALL_SYSCONFDIR", self.prefix.etc)
```

alalazo (commented):
If you rebase and ping me I'll take care of merging the PR asap

alalazo:
@spackbot fix style

alalazo (approved):

PR #30858, opened 2022-05-25
wget: fixing compilation issue with gcc11+ on rhel7

snehring:
gcc11 and 12 defaulting to c17 breaks compiling wget against older glibc versions with very helpful error messages such as:
```./string.h:1080:1: error: expected identifier or '(' before '__extension__'```. 

This is probably an issue with other older os versions beyond rhel/centos7, but I don't have any other hosts readily available to test against.

PR #31699, opened 2022-07-22
wget: replacing prior workaround with proper fix

snehring:
The flag hander workaround I put in #30858 isn't the best fix, but fortunately @haampie was able to find the proper fix when I hit the same issue in coreutils.


haampie:
Looks good, it's not yet fixed on their master branch, but restricting to 1.21.3 is fine for now

haampie (approved):

PR #31701, opened 2022-07-22
Add VASP 6.3.2; handle changes to makefile.include naming pattern

alansill:
Added the SHA256 for version 6.3.2 and added logic to eal with the change of naming pattern for the makefile.include files that now appears to leave out the "linux_" prefix. (Changes should be. backwards compatible.)


alansill:
I don't know what to do about the windows style tests since this package was not previously available in the build system for windows AFAICT, and this package.py was aimed specifically at Linux.

var/spack/repos/builtin/packages/vasp/package.py
```diff
@@ -115,7 +116,13 @@ def edit(self, spec, prefix):
...
                                          spec.compiler.name)
 
-        os.rename(make_include, 'makefile.include')
+        # Recent versions of vasp have renamed the makefile.include files
+        # to leave out the linux_ string
+        if os.path.exists(make_include):
+            os.rename(make_include, 'makefile.include')
+        else:
+            new_make_include = make_include.replace('linux_', '')
+            os.rename(new_make_include, 'makefile.include')
```
alalazo:
```suggestion
        if not os.path.exists(make_include):
            make_include = make_include.replace('linux_', '')
        os.rename(make_include, 'makefile.include')
```

alalazo (changes requested):
One minor request and a conflict to be fixed.

alansill:
I don't understand why it has flagged the updated version as a conflict. The whole point of this pull request was to update the package.py file in exactly those lines. What is flagging the additions as producing a conflict?

alalazo:
@alansill We started enforcing style with [black](https://github.com/psf/black) a few days ago, and that meant reformatting the entire code base. You need to rebase against develop to resolve the conflict, sorry.

alalazo:
@spackbot fix style

alalazo (approved):

PR #31703, opened 2022-07-23
new package: r-signac

glennpj:
Also includes new package depdendency for r-rcpproll.


tldahlgren (approved):
Confirmed the sha256 for both packages.

PR #31705, opened 2022-07-23
dd4hep: add v1.21

vvolkl:
Some small backwards incompatibilities introduced in this version, but nothing that should affect the packages in the spack builtin repo.


tldahlgren (approved):
Confirmed sha256.

PR #31739, opened 2022-07-26
bugfix: use cmake version from dependency

trws:
Ensure that build tools with module-level commands in spack use the version built as part of their build graph if one exists.  This is now also required for mesa, scons, cmake and ctest, out of graph versions of these tools in path will not be found unless added as an external.  

This bug appeared because a new version of rocprim needs cmake 3.16, while I have 3.14 in my path I had added an external for cmake 3.20 to the dag, but 3.14 was still used to configure rocprim causing it to fail.  As far as I can tell, all the build tools added in build_environment.py had this problem, despite the fact that they should have been resolving these tools by name with a path search and find the one in the dag that way.  I'm still investigating why the path searching and Executable logic didn't do it, but this makes three of the build systems much more explicit, and leaves only gmake and ninja as dependencies from out in the system while ensuring the version in the dag is used if there is one.

The additional sqlite version is to perturb the hash of python to work around a relocation bug which will be fixed in a subsequent PR.

### DIFF
--- a/lib/spack/spack/util/executable.py
+++ b/lib/spack/spack/util/executable.py
@@ -291,6 +291,13 @@ def which_string(*args, **kwargs):
             win_candidates = [name + ext for ext in [".exe", ".bat"]]
         candidate_names = [name] if not win_candidates else win_candidates
 
+        if sys.platform == "win32":
+            new_path = path[:]
+            for p in path:
+                if os.path.basename(p) == "bin":
+                    new_path.append(os.path.dirname(p))
+            path = new_path
+
         for candidate_name in candidate_names:
             if os.path.sep in candidate_name:
                 exe = os.path.abspath(candidate_name)
--- a/lib/spack/spack/util/executable.py
+++ b/lib/spack/spack/util/executable.py
@@ -328,14 +328,12 @@ def which(*args, **kwargs):
     Keyword Arguments:
         path (list or str): The path to search. Defaults to ``PATH``
         required (bool): If set to True, raise an error if executable not found
-        exe_class (callable): If not None, used instead of Executable
 
     Returns:
         Executable: The first executable that is found in the path
     """
     exe = which_string(*args, **kwargs)
-    exe_class = kwargs.get("exe_class", Executable)
-    return exe_class(exe) if exe else None
+    return Executable(exe) if exe else None
 
 
 class ProcessError(spack.error.SpackError):
--- a/lib/spack/spack/util/executable.py
+++ b/lib/spack/spack/util/executable.py
@@ -328,12 +328,14 @@ def which(*args, **kwargs):
     Keyword Arguments:
         path (list or str): The path to search. Defaults to ``PATH``
         required (bool): If set to True, raise an error if executable not found
+        exe_class (callable): If not None, used instead of Executable
 
     Returns:
         Executable: The first executable that is found in the path
     """
     exe = which_string(*args, **kwargs)
-    return Executable(exe) if exe else None
+    exe_class = kwargs.get("exe_class", Executable)
+    return exe_class(exe) if exe else None
 
 
 class ProcessError(spack.error.SpackError):
--- a/var/spack/repos/builtin.mock/packages/mpich/package.py
+++ b/var/spack/repos/builtin.mock/packages/mpich/package.py
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
+from spack.util.executable import Executable
+
+ninja = Executable('ninja')  # to appease ninja test
 
 
 class Mpich(Package):
--- a/var/spack/repos/builtin.mock/packages/mpich/package.py
+++ b/var/spack/repos/builtin.mock/packages/mpich/package.py
@@ -6,7 +6,7 @@
 from spack.package import *
 from spack.util.executable import Executable
 
-ninja = Executable('ninja')  # to appease ninja test
+ninja = Executable("ninja")  # to appease ninja test
 
 
 class Mpich(Package):
--- a/var/spack/repos/builtin.mock/packages/mpich/package.py
+++ b/var/spack/repos/builtin.mock/packages/mpich/package.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
-from spack.util.executable import Executable
 
 
 class Mpich(Package):
--- a/var/spack/repos/builtin.mock/packages/mpich/package.py
+++ b/var/spack/repos/builtin.mock/packages/mpich/package.py
@@ -6,8 +6,6 @@
 from spack.package import *
 from spack.util.executable import Executable
 
-ninja = Executable("ninja")  # to appease ninja test
-
 
 class Mpich(Package):
     homepage = "http://www.mpich.org"
--- a/var/spack/repos/builtin/packages/cmake/package.py
+++ b/var/spack/repos/builtin/packages/cmake/package.py
@@ -403,6 +403,7 @@ class Cmake(Package):
         """Called before cmake packages's install() methods."""
 
         module.cmake = Executable(self.spec.prefix.bin.cmake)
+        module.ctest = Executable(self.spec.prefix.bin.ctest)
 
     def test(self):
         """Perform smoke tests on the installed package."""
--- a/var/spack/repos/builtin/packages/cmake/package.py
+++ b/var/spack/repos/builtin/packages/cmake/package.py
@@ -399,6 +399,11 @@ class Cmake(Package):
                 filter_file("mpc++_r)", "mpc++_r mpiFCC)", f, string=True)
                 filter_file("mpifc)", "mpifc mpifrt)", f, string=True)
 
+    def setup_dependent_package(self, module, dependent_spec):
+        """Called before cmake packages's install() methods."""
+
+        module.cmake = Executable(self.spec.prefix.bin.cmake)
+
     def test(self):
         """Perform smoke tests on the installed package."""
         spec_vers_str = "version {0}".format(self.spec.version)
--- a/var/spack/repos/builtin/packages/ninja/package.py
+++ b/var/spack/repos/builtin/packages/ninja/package.py
@@ -6,6 +6,7 @@ import sys
 
 from spack.build_environment import MakeExecutable
 from spack.package import *
+from spack.util.executable import which_string
 
 
 class Ninja(Package):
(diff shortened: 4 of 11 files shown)


davidbeckingsale (dismissed):

var/spack/repos/builtin/packages/cmake/package.py
```diff
@@ -375,6 +375,11 @@ def install(self, spec, prefix):
                 filter_file('mpc++_r)', 'mpc++_r mpiFCC)', f, string=True)
                 filter_file('mpifc)', 'mpifc mpifrt)', f, string=True)
 
+    def setup_dependent_package(self, module, dependent_spec):
+        """Called before cmake packages's install() methods."""
+
+        module.cmake = Executable(self.spec.prefix.bin.cmake)
```
tgamblin, line 405: If we're doing this, would it not also make sense to just remove the cmake assignment from `build_enviroment.py`?

tgamblin (changes requested):

eugeneswalker:
I wonder if this would fix our issue wherein `raja +rocm` fails in cmake phase because it can't find `rocprim`.

I'm trying a build now, with this PR.

See comment on this line:
https://github.com/spack/spack/blob/develop/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml#L225

var/spack/repos/builtin/packages/cmake/package.py
trws: I hesitated on that lest it create fallout, but it's worth a shot.

trws:
> I wonder if this would fix our issue wherein `raja +rocm` fails in cmake phase because it can't find `rocprim`.
> 
> I'm trying a build now, with this PR.
> 
> See comment on this line: https://github.com/spack/spack/blob/develop/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml#L225

It should.  I added the dependency, and found the error with cmake because the rocprim package requires a newer cmake than is available by default on out clusters.  If not, definitely let me know.

eugeneswalker:
> I wonder if this would fix our issue wherein `raja +rocm` fails in cmake phase because it can't find `rocprim`.

Looks like this PR gets us past the cmake error. Now, we have a new error, but that is progress!
```
==> Installing raja-2022.03.0-qeg5yfid7aadpw4fspjbfacul5hbxmpc
==> No binary for raja-2022.03.0-qeg5yfid7aadpw4fspjbfacul5hbxmpc found: installing from source
==> No patches needed for raja
==> raja: Executing phase: 'initconfig'
==> raja: Executing phase: 'cmake'
==> raja: Executing phase: 'build'
...
  >> 131    g++: error: unrecognized command-line option '--offload-arch=gfx900'
  >> 132    make[2]: *** [CMakeFiles/RAJA.dir/build.make:79: CMakeFiles/RAJA.dir/src/AlignedRangeIndexSetBuilders.cpp.o] Error 1
     133    make[2]: *** Waiting for unfinished jobs....
  >> 134    g++: error: unrecognized command-line option '--offload-arch=gfx900'
  >> 135    make[2]: *** [CMakeFiles/RAJA.dir/build.make:135: CMakeFiles/RAJA.dir/
...
```

becker33 (changes requested):
I think we need to do the same thing for `ctest` in this PR.

Doesn't need to be this PR, but looks like we need to do the same thing for `make`/`gmake`/`scons`/`ninja`/`meson`. For `make` and `gmake` we probably should keep the ones in `build_environment` but overwrite them if it's in the DAG, the others we should treat like this.

trws:
I agree @becker33, will take a look at those others.

trws:
Ah… That’s a different problem, it’s because you’re telling raja 
to build rocm with gcc.  There’s nothing in HipPackage or anything 
like that which actually changes the C++ compiler to the correct value, 
most other rocm packages are just lying about what compiler they’re 
using right now.  If you switch it to build raja with the compiler 
installed by llvm-amdgpu it will work, still working on what the best 
way around that is for the spack package and raja cmake stuff.  No clue 
why we weren’t seeing this before.

eugeneswalker:
> Ah… That’s a different problem, it’s because you’re telling raja to build rocm with gcc. There’s nothing in HipPackage or anything like that which actually changes the C++ compiler to the correct value, most other rocm packages are just lying about what compiler they’re using right now. If you switch it to build raja with the compiler installed by llvm-amdgpu it will work, still working on what the best way around that is for the spack package and raja cmake stuff. No clue why we weren’t seeing this before.

What about doing it like this? Would this work?

* https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/exago/package.py#L133-L134
* https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/hiop/package.py#L195-L196
* https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/trilinos/package.py#L465-L471

trws:
Yes, I think it would.

trws:
I take that back @eugeneswalker, it would *almost* work.  RAJA needs the actual clang compiler from hip rather than hipcc, so it would need a bit of extra logic to grab that depending on whether the result is `amdclang++` or `clang++`.

eugeneswalker (dismissed):

trws:
@tgamblin, @becker33 I think I've updated everything that can reasonably be done here.  The gmake package overrides the version used, but leaves the base one there so we don't break the world.  I went to add a `bmake` to the bsd make package, but it currently doesn't build for me under spack (the exact same command outside spack works, so I have no idea why) but we didn't have that before so it's not terribly important.

All of: cmake, ctest, scons, meson and ninja are removed from build_environment.py.  It would not surprise me if we find out that ninja being missing breaks something, if that happens I intend to add it as a dependency in the CmakePackage class, that class needs some cleanup anyway to get rid of all the make/ninja crud, but it would almost certainly take care of all the packages that need updates.

tgamblin:
@trws can you do one more thing (sorry this is currently manual) -- can you do a rebuild like we did for your `-Werror` PR?  I feel like we need that for due diligence to make sure this does not break builds.

trws:
All raja/camp changes have been removed from here and will be part of a subsequent PR once umpire issues are resolved.  It was proving problematic for several stacks, so this will fix the cmake/build tool issue only.

trws:
This should be ready for a re-review if anyone has time.  The failure is a syntax error in one configuration of mesa, and it's the only error that comes up both for this *and* for the build-the-world PR.  It's here: https://gitlab.spack.io/spack/spack/-/jobs/2924593

I'll look into that, but the fact that it's only AWS aarch64 where it happens and that it only happens in one configuration, and has nothing to do with the build tool selected, makes me think it's probably not directly this.

Especially look over the changes in `which`, it was necessary either to special-case `ninja` or do that, but it seems Visual Studio installs quite a number of tools that way, including others found by this PR.

tgamblin (dismissed):

trws:
@becker33, if you get a chance could you take another look here? Mainly a look at the windows handling in Executable would be good.  It's finally passing builds (assuming the GHA tests don't time out this time).

lib/spack/spack/util/executable.py
```diff
@@ -291,6 +291,13 @@ def which_string(*args, **kwargs):
             win_candidates = [name + ext for ext in [".exe", ".bat"]]
         candidate_names = [name] if not win_candidates else win_candidates
 
+        if sys.platform == "win32":
+            new_path = path[:]
+            for p in path:
+                if os.path.basename(p) == "bin":
+                    new_path.append(os.path.dirname(p))
```
becker33, line 298: Doesn't this need an `else` case? If the path doesn't end in `bin`, I don't think we should just drop it.

var/spack/repos/builtin.mock/packages/mpich/package.py
```diff
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
+from spack.util.executable import Executable
+
+ninja = Executable("ninja")  # to appease ninja test
```
becker33: Can you explain what this is doing?

var/spack/repos/builtin/packages/ninja/package.py
```diff
@@ -69,3 +69,9 @@ def install(self, spec, prefix):
         # instead of 'ninja'. Install both for uniformity.
         with working_dir(prefix.bin):
             symlink("ninja", "ninja-build")
+
+    def setup_dependent_package(self, module, dspec):
+        name = "ninja"
+        if sys.platform == "win32":
+            name = name + ".exe"
+        module.ninja = which(name, path=[self.spec.prefix.bin], required=True)
```
becker33: Should we make `which` just automatically try with/without `.exe` on windows?

lib/spack/spack/util/executable.py
trws: The new_path list always contains the original contents of path, hence the init with `path[:]`, then if it ends in `bin` it gets duplicated without the `bin` on the end.

var/spack/repos/builtin.mock/packages/mpich/package.py
trws: All of the ninja build_system tests assume that a ninja `Executable` object exists in the module scope of this mock package.  Adding it as a dependency doesn't work, because it can't actually build it, so the build_system tests need us to find the ninja executable in the environment.  This was my initial stopgap for keeping them all working, but it doesn't make me particularly happy.

var/spack/repos/builtin/packages/ninja/package.py
trws: That's a really good point, I'll check that and add it if we aren't doing it.

trws: Apparently we are, so I'm removing this check.

var/spack/repos/builtin.mock/packages/mpich/package.py
trws: Had a bit more of a think about this and reworked the ninja handling, which took some tweaking of `which` and a couple other spots, but I think puts us in a much better place.  Apparently we weren't actually setting `-j` reliably on ninja unless it was being set by the caller, that's fixed, also put it back in the default environment if it's available because it really is a bottom dependency, especially on windows.  Bit more churn, but I like the result a whole lot more.

becker33 (dismissed):

becker33 (dismissed):

trws:
@tgamblin, @becker33, this is now passing everything except Kokkos 3.4 on oneapi, so it's for all intents and purposes passing.  The ninja/which code has been tweaked a bit once more, but mainly to remove things.

trws:
Yes... I think I remember how to set that up, will copy you from a new 
PR to poke all the builders.

PR #31744, opened 2022-07-26
herwig3, thepeg: correct lhapdfsets dependency, specify boost libs

vvolkl:


iarspider:
Doesn't herwig need lhapdf sets at runtime as well? I find it weird...

vvolkl:
@iarspider Yes it does, but discussions in  SPI resulted in  the suggestion not to depend on them because of the difficulties in versioning. If CMS publishes a properly versioned dataset I think we could use that though ...

haralmha (approved):

tldahlgren (approved):
LGTM

PR #31746, opened 2022-07-26
`spack -e x config --scope=y add z` add to scope

haampie:
Fixes a bug where `spack -e x config --scope=y add z` adds `z` to environment `x` instead of scope `y`.

### DIFF
--- a/lib/spack/spack/cmd/config.py
+++ b/lib/spack/spack/cmd/config.py
@@ -120,7 +120,7 @@ def _get_scope_and_section(args):
     path = getattr(args, "path", None)
 
     # w/no args and an active environment, point to env manifest
-    if not section and not path:
+    if not section and not scope:
         env = ev.active_environment()
         if env:
             scope = env.env_file_config_scope_name()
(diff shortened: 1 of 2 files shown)


lib/spack/spack/cmd/config.py
```diff
@@ -117,7 +117,7 @@ def _get_scope_and_section(args):
     path = getattr(args, 'path', None)
 
     # w/no args and an active environment, point to env manifest
-    if not section:
+    if not section and not path:
```
becker33: Shouldn't this be `and not scope`?

haampie: Nope. I think this is followed in `spack config edit` s.t. it opens the spack.yaml file.

haampie: changed that. not sure what `spack -e x config edit [path]` is supposed to do, but irrelevant for this pr.

haampie:
ping @becker33

becker33 (approved):

PR #31748, opened 2022-07-26
OpenGL restructure

chuckatkins:
This provides concrete `glx` and `osmesa` packages which delegate to virtual `libglx` and `libosmesa` packages provided by mesa. This was necessary because GLX and OSMesa both provide GL implementations but with `mesa` providing the virtual `gl` package there was no way to properly distinguish which of the two OpenGL implementations was being requested when querying the `spec['gl']` dependency.  The additional level of indirection this adds allows for that.

When updating the dependees to work with the changes to the gl packages, this also attempts to create a consistent variant for specifying `glx` or `osmesa`.


var/spack/repos/builtin/packages/catalyst/package.py
```diff
@@ -79,8 +79,8 @@ class Catalyst(CMakePackage):
     depends_on('py-mpi4py', when='+python3', type=('build', 'run'))
 
     depends_on('gl@3.2:', when='+rendering')
-    depends_on('osmesa', when='+rendering+osmesa')
-    depends_on('glx', when='+rendering~osmesa')
+    depends_on('osmesa', when='osmesa')
```
kwryankrattiger: missing a `+` here

adamjstewart:
> This was necessary because GLX and OSMesa both provide GL implementations but with `mesa` providing the virtual `gl` package there was no way to properly distinguish which of the two OpenGL implementations was being requested when querying the `spec['gl']` dependency.

Can't you just check the value of `spec['gl'].name` or `'^mesa' in spec`?

var/spack/repos/builtin/packages/mesa-demos/package.py
```diff
@@ -11,18 +11,21 @@ class MesaDemos(AutotoolsPackage):
     """This package provides some demo applications for testing Mesa."""
 
     homepage = "https://www.mesa3d.org"
-    url      = "https://github.com/freedesktop/mesa-demos/archive/mesa-demos-8.3.0.tar.gz"
+    url      = "https://gitlab.freedesktop.org/mesa/demos/-/archive/mesa-demos-8.4.0/demos-mesa-demos-8.4.0.tar.gz"
 
+    version('8.4.0', sha256='e9d235e6dad69d6b00877bf07e6d1859e368c0873e5401ec68a6ddb43375e900')
```
kwryankrattiger, line 16: Should this come in separately?

chuckatkins:
> > This was necessary because GLX and OSMesa both provide GL implementations but with `mesa` providing the virtual `gl` package there was no way to properly distinguish which of the two OpenGL implementations was being requested when querying the `spec['gl']` dependency.
> 
> Can't you just check the value of `spec['gl'].name` or `'^mesa' in spec`?

No because `mesa` is actually providing two distinct implementations of `gl`, just knowing that it's provided by `mesa` doesn't have the specificity for which one the package wants.  This is also forward looking to setup some of the indirection needed to enable glvnd where `glx` and `gl` are two distinctly separate libraries `libGLX.so` and `libOpenGL.so` and `mesa` doesn't provide any libraries directly used by applications.

adamjstewart:
But can't you use `spec['gl'].name.startswith('mesa')`? We really need to combine `mesa` and `mesa18` into a single package, there's no reason it should be split. We have many packages for libraries that changed build systems, and documentation for this too: https://spack.readthedocs.io/en/latest/build_systems/multiplepackage.html

If the libraries provided by mesa and opengl are not equivalent that's a much better reason for further subdividing the virtual dep though. Just saying that it isn't strictly necessary for knowing which gl provider is in use.

var/spack/repos/builtin/packages/visit/package.py
```diff
@@ -195,14 +195,14 @@ class Visit(CMakePackage):
     depends_on('vtk ~mpi', when='~mpi')
 
     # Necessary VTK patches
-    depends_on('vtk', patches=[patch('vtk_compiler_visibility.patch')])
+    depends_on('vtk', patches=[patch('vtk_compiler_visibility.patch')],
```
kwryankrattiger, line 101: Does this belong in this patch?

kwryankrattiger:
@adamjstewart This patch is more directed at getting spack to deduce `opengl` without forcing package maintainers to care about where it is coming from. Selecting a consistent GL provider across the stack is currently extremely difficult and error prone. This alleviates much of those issues.

chuckatkins:
> But can't you use `spec['gl'].name.startswith('mesa')`? We really need to combine `mesa` and `mesa18` into a single package, there's no reason it should be split. We have many packages for libraries that changed build systems, and documentation for this too: https://spack.readthedocs.io/en/latest/build_systems/multiplepackage.html
> 
> If the libraries provided by mesa and opengl are not equivalent that's a much better reason for further subdividing the virtual dep though. Just saying that it isn't strictly necessary for knowing which gl provider is in use.

Re separate `mesa` and `mesa18` packages: At the time it was implemented 2y ago there wasn't a recommended way of handling a build system switch and since nothing should be directly depending on the mesa package itself, only the virtuals it provides, splitting into two was much cleaner and simpler wrt the package being less complicated and more maintainable. 

As for the additional indirection, the issue is that the `mesa` package itself provides multiple `gl` implementations via `glx` and `osmesa`, not that multiple mesa implementations via `mesa` and `mesa18` exist.  The issues are orthogonal to one another.

adamjstewart:
We already have glx and osmesa virtuals, is there any reason to rename them to libglx and libosmesa? It's unclear to me what is actually different about this PR other than renaming existing virtuals.

chuckatkins:
> We already have glx and osmesa virtuals, is there any reason to rename them to libglx and libosmesa? It's unclear to me what is actually different about this PR other than renaming existing virtuals.

The goal is to have `gl` provided by `glx` and `osmesa` and not the `mesa` package but there's no way to have a virtual package provide another virtual package (just making the package files, setting `self.virtual=True` and addig a `provides` doesn't work). So this makes `glx` and `osmesa` concrete packages that provide the virtual `gl` package.

chuckatkins:
> > We already have glx and osmesa virtuals, is there any reason to rename them to libglx and libosmesa? It's unclear to me what is actually different about this PR other than renaming existing virtuals.
> 
> The goal is to have `gl` provided by `glx` and `osmesa` and not the `mesa` package but there's no way to have a virtual package provide another virtual package (just making the package files, setting `self.virtual=True` and adding a `provides` doesn't work). So this makes `glx` and `osmesa` concrete packages that provide the virtual `gl` package.

So, that's actually the whole point of this pr and why it's needed: `mesa` doesn't actually provide a single OpenGL implementation, it provides multiple implementations via GLX and OSMesa, both of which are their own complete OpenGL implementation.  So when `mesa` provides `gl` then `spec['gl'].libs` doesn't have the necessary specificity to identify which `gl` implementation is being used. Moving `provides('gl')` to concrete `glx` and `osmesa` packages provides the necessary specificity.

adamjstewart:
Interesting. So, to confirm my understanding:

* OpenGL is an API, of which there are many implementations (similar to BLAS/LAPACK/MPI/etc.)
* GLX and OSMesa are implementations of OpenGL, both provided by the `mesa` and `mesa18` packages in Spack
* Spack also has an `opengl` package for generic OpenGL externals, e.g., on macOS

Are there packages that only support `mesa` GLX or `mesa` OSMesa but not both? If not, wouldn't it make sense to have a single virtual (`gl`) that is provided by `mesa`, `mesa18`, and `opengl` and get rid of virtuals like `osmesa` and `glx`? Then the use of GLX or OSMesa could be controlled by a variant instead of a virtual. Just trying to better understand what should be a virtual vs. what should be a package.

chuckatkins:
> Interesting. So, to confirm my understanding:
> 
>     * OpenGL is an API, of which there are many implementations (similar to BLAS/LAPACK/MPI/etc.)
> 
>     * GLX and OSMesa are implementations of OpenGL, both provided by the `mesa` and `mesa18` packages in Spack
> 
>     * Spack also has an `opengl` package for generic OpenGL externals, e.g., on macOS

Close.  The OpenGL family is split up into frontend APIs and backend APIs.
  * Backend APIs - Actually run on GPUs and and do the heavy lifting.
    * OpenGL - The desktop rendering API we all know and love.
    * OpenGL ES - A reduced OpenGL API targeting embedded environments.
  * Frontend APIs - Interface the backend with the "presentation" layer of the system.
    * GLX - Interface GL with the X Window System
    * OSMesa - GL for CPU-only headless rendering; a mesa-specific implementation, i.e. OffScreen Mesa.
    * EGL - Interface GL with embedded UIs, typically used on devices but in the past few years also on the Linux desktop via Wayland; NVidia drivers also allow headless full OpenGL via EGL.
    * WGL - Interface GL with Microsoft Windows
    * AGL - Interface GL with Apple's native window system.

Traditionally the frontend and backend APIs are combined into a single library, `libGL.so` for GLX+OpenGL, `libOSMesa.so` for OSMesa+OpenGL, `opengl32.{lib,dll}` for WGL+OpenGL, and `OpenGL.framework` for AGL+OpenGL. EGL is a newer API and departs from this with `libEGL.so` for EGL combined with `libGLESv1_CM.so` for OpenGL ES v1 or `libGLESv2.so` fir OpenGL ES v2.  This continues to be the case on Apple and Windows but on Linux GLVND (GL Vendor Neutral Dispatch) has changed everything and has been they way do do OpenGL on linux for the past several years.  GLVND separates the frontend and backend APIs into distinct stand-alone libraries: `libGLX.so`, `libEGL.so`, `libOpenGL.so`, `libGLESv1_CM.so`, and `libGLESv2.so`. These are end-user / developer-facing libraries provided by the GLVND package are are little more that stubs or shims, with the actual implementation provided by something like Mesa or the NVIDIA driver and loaded at runtime from either environment variables or config files.

Now, as for spack.  The OpenGL infrastructure in spack doesn't handle GLVND yet but it needs to and creating this logical separation of the `gl` backend from the `glx` front end lays the foundation for that (there've been several pushes for this change by myself or my team but none of them have quite crossed the finish line).  So, for the "traditional" setup, i.e. non-glvnd,  we have the following:

The Mesa package, either via the `mesa` or `mesa18` spack packages, provides the GLX and OpenGL APIs together in `libGL.so`, as well as the OSMesa and OpenGL APIs together in `libOSMesa.so`.  The `opengl` package is intended to be only external for a system or vendor provided implementation of GLX+OpenGL via `libGL.so` on Linux, the combined WGL+OpenGL on Windows, or the `OpenGL.framework` on Apple.


> Are there packages that only support `mesa` GLX or `mesa` OSMesa but not both? If not, wouldn't it make sense to have a single virtual (`gl`) that is provided by `mesa`, `mesa18`, and `opengl` and get rid of virtuals like `osmesa` and `glx`? Then the use of GLX or OSMesa could be controlled by a variant instead of a virtual. Just trying to better understand what should be a virtual vs. what should be a package.

Most applications only support one at a time but that's entirely consumer driven.  `mesa` can provide both at the same time and it's particularly useful for it to do so since it can provide a shared infrastructure for many parts of the underlying implementations.  This is even more so when GLVND eventually gets enabled.

kwryankrattiger (approved):
lgtm

var/spack/repos/builtin/packages/glew/package.py
```diff
@@ -18,18 +19,41 @@ class Glew(CMakePackage):
...
     version('2.0.0',  sha256='c572c30a4e64689c342ba1624130ac98936d7af90c3103f9ce12b8a0c5736764')
 
+    variant('gl',
+            default='glx' if sys.platform.startswith('linux') else 'other',
+            values=('glx', 'osmesa', 'other'), multi=False,
+            description='The OpenGL provider to use')
+    conflicts('osmesa', when='gl=glx')
+    conflicts('osmesa', when='gl=other')
+    conflicts('glx', when='gl=osmesa')
+    conflicts('glx', when='gl=other')
```
dylex, line 29: I'm a bit confused by these conflicts directives. They seem to say that this package conflicts with being named (or providing a virtual) "osmesa", which will never happen. Are they supposed to say `'^osmesa'` meaning conflict with depending on osmesa or `'gl=osmesa'`?

kwryankrattiger: The variant `gl=<X>` is a variant for this package, it is not something special that spack interprets at the same level as `osmesa` or `^osmesa`. The point of the conflicts is to translate that variant correctly to the virtual packages so that an invalid combination does not get built.

dylex: But osmesa and glx are concrete packages, not virtual ones, and glew does not provide either of them, does it? If I'm interpreting the conflict checking correctly, saying `conflicts('osmesa', ...)` translates into checking whether the concretized spec for glew `glewSpec.satisfies('osmesa')` and I don't see how that can ever be the case.

kwryankrattiger: The conflict is just saying you cannot build a `glew gl=glx` if osmesa is also in the stack, because that isn't compatible. And that can happen because mesa can provide both glx and osmesa backends in the same build. You try removing these and building glew with osmesa and paraview with ~osmesa you can get a build with mixed backends which is not good. These conflicts prevent those cases.

adamjstewart:
I'm trying to use this on macOS 10.15.7 but I'm running into a few issues. I have two installations of OpenGL, a traditional Unix installation in `/opt/x11`, and a macOS framework version in `/System/Library/Frameworks/OpenGL.framework`. The former appears to follow the same layout as Linux, but the opengl recipe now uses a completely different layout for macOS. The framework version has no `glxinfo` or `gl.h`, only `libGL.dylib`. Any idea how I can get this working on macOS?

kwryankrattiger:
@adamjstewart Briefly reading it sounds like the headers should be in `/System/Library/Frameworks/OpenGL.framework/Headers` for the framework stuff. Is that not the case?

Based on how the opengl (the unix-like external interface in spack) works, you should be able to just set the prefix to `prefix: /System/Library/Frameworks/OpenGL.framework` in the externals and the default provider for libglx to opengl

```yaml
packages:
  all:
    providers:
      libglx:: [opengl]
  opengl:
    buildable: false
    externals:
    - spec: opengl@4.1
      prefix: /System/Library/Frameworks/OpenGL.framework
```

adamjstewart:
There is no `Headers` directory or any `*.h` files anywhere recursively within that prefix.

PR #31751, opened 2022-07-26
Tau must get GCC path from environment on Cray

wspear:
self.compiler doesn't provide the path to the gcc compiler when using cray cc and the spack internal compiler overrides the location in PATH. If possible get the location from the GCC_PATH variable instead.


tldahlgren (approved):
LGTM

PR #31764, opened 2022-07-27
itensor: new version, restrict hdf5 to :1.10

melven:
ITensor does not build with HDF 1.12, see e.g. https://portal.hdfgroup.org/display/support/HDF5%201.12.2#releasenotes for HDF5 API compatibility.


tldahlgren (approved):
Confirmed sha256.

PR #31766, opened 2022-07-27
py-absl-py: new version

melven:

PR #31767, opened 2022-07-27
py-tensorflow: new version (2.9.1)

melven:
Add Tensorflow 2.9.1 with versions of dependencies extracted from tensorflow/tools/pip_package/setup.py .

### DIFF
--- a/var/spack/repos/builtin/packages/py-keras/package.py
+++ b/var/spack/repos/builtin/packages/py-keras/package.py
@@ -16,6 +16,7 @@ class PyKeras(PythonPackage):
     git = "https://github.com/keras-team/keras"
     url = 'https://github.com/keras-team/keras/archive/refs/tags/v2.7.0.tar.gz'
 
+    version('2.9.0', sha256='90226eaa0337573304f3e5ab44d4d9e3a65fe002776c5cbd0f65b738152c1084')
     version('2.7.0', sha256='7502746467ab15184e2e267f13fbb2c3f33ba24f8e02a097d229ba376dabaa04')
     version('2.6.0', sha256='15586a3f3e1ed9182e6e0d4c0dbd052dfb7250e779ceb7e24f8839db5c63fcae')
     version('2.5.0', commit='9c266106163390f173625c4e7b1ccb03ae145ffc')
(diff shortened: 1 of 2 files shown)


melven:
This also requires newer py-absl-py, see https://github.com/spack/spack/pull/31766 .

melven:
I managed to compile py-tensorflow 2.9.1 with py-keras with CUDA support and GCC 11 on Ubuntu 20.04.

Here is a part of the log-file with all involved specs / versions: [spack_concretize_py_tensorflow.log](https://github.com/spack/spack/files/9196504/spack_concretize_py_tensorflow.log)

aweits:
@melven have you looked at this PR? #31615

melven:
> @melven have you looked at this PR? #31615

No, sorry, I missed that one. My changes are pretty much included in #31615, so I'll close this MR in favor of the much more extensive changes in #31615.

Only the support for newer CUDA is not included there (not sure what the documentation said, but it 11.7 worked for me in first tests):
```
-    depends_on('cuda@:11.4', when='+cuda @2.4:')
+    depends_on('cuda@:11.4', when='+cuda @2.4:2.7')
+    depends_on('cuda@:11.7', when='+cuda @2.9:')
```

melven:
Closing this MR in favor of #31615.

adamjstewart:
I'm hesitant to support a wider range of CUDA if the docs disagree. That's how we got into trouble with Bazel and Java. If you can find the docs on this let me know. We can also submit a PR to update the docs if that's needed.

melven:
Ok, beeing conservative here is probably better - the ML frameworks are really ugly to compile...

adamjstewart:
But still look for the docs, it may have changed in the latest release and I didn't find it yet.

melven:
[Docs](https://www.tensorflow.org/install/source?hl=en#gpu) and [tensorflow setup.py](https://github.com/tensorflow/tensorflow/blob/d8ce9f9c301d021a69953134185ab728c1c248d3/tensorflow/tools/pip_package/setup.py#L319) say both 11.2.

adamjstewart:
Hmm, we certainly don't want to lock it down to a single version if we know newer versions work. I'm open to changing the bounds to be `:11.7` or just removing the bounds altogether unless we know there is a specific incompatibility.

adamjstewart:
@aweits and @glennpj added our current CUDA constraints, do you remember where those came from or if newer versions didn't work?

glennpj:
I would have gotten the upper constraint from building/testing. I cannot remember now if I tested all the tensorflow versions that I added (2.5-2.7) with cuda >11.4 or not. The documented tested configurations only list a single version of cuda/cudnn/compiler/bazel, which is good as a reference but a little limiting for Spack's vast build matrix.

adamjstewart:
@melven if you remove the version constraint entirely, what's the oldest version of TF you can build with CUDA 11.7? I do remember there being backwards incompatible changes in CUDA 11.5 that [affected PyTorch](https://discuss.pytorch.org/t/compiling-1-10-1-from-source-with-gcc-11-and-cuda-11-5/140971), so I don't think it's safe to assume that ALL versions support CUDA 11.5+. If we can figure out the minimum version that works, we can document that instead.

glennpj:
I think I was seeing run time errors when I last looked.

adamjstewart:
Supported CUDA versions have been updated in #31615!

PR #31618, opened 2022-07-18
fix stage with uri params

haampie:
Closes #31499.
 
- don't determine extension of a url
- parse url to drop query params from filename

Can you take over from here @scheibelp / @johnwparent, this regressed in the Windows related work, and I can't invest more time in it right now adding tests.

Reproducer:

```console
$ spack clean -sd && spack -d stage eccodes@2.25.0
```

[As mentioned in #31499, it's better to rely on [Content-Dispostion](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) when provided, since URLs are not file paths. This can be obtained from [get_filename](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.get_filename) (note that `http.client.HTTPMessage` is a subclass of `email.message.Message` on recent enough Pythons). Either that, or just fix the filename to `your-favorite-name` and detect the file type through magic bytes]

PR #31783, opened 2022-07-28
Fix parsing of clean_url

stephenmsachs:
Fixes `IsADirectoryError` on https://github.com/spack/spack/pull/31618


haampie (approved):
Just to clarify: in #31618 the idea was to determine the file extension heuristically by dropping query params etc from a url and then consider it as a file path. That broke for URLs that only have query params like `http://example.com/?patch=x` as it would result in empty string as basename. This PR reverts to the old behavior of saving files as `?patch=x` in that case.

PR #31784, opened 2022-07-28
gcc: fix unstable patch urls

haampie:
the patch urls dynamically generate a diff, which includes metadata
about the git version used, meaning the shasum will change.

instead ship the patches with spack.

Closes #31759


michaelkuhn (approved):
Thanks!

PR #30038, opened 2022-04-13
Deprecate `spack:concretization` over `concretizer:unify` 

haampie:
Now that we have `concretizer` configuration, we can use it for environments.

1. Introduce `spack:concretizer:unify:true|false` in favor of `spack:concretization:together|separately`.
2. Add a `spack env update` path. This update makes environments incompatible with Spack 0.17, but forward compatible with 0.19 once we remove `spack:concretization`.
3. Add a deprecation warning for `concretization`.
4. Store the default value `spack:concretizer:unify` in new `spack.yaml` files to make it easier for us to change the default of `unify: false` to `unify: true` in 0.19. (Downside being that new environments in 0.18 are incompatible with 0.17 by default).
5. Turn `default_manifest_yaml` into a function (should be private API?) so we can use spack.config in it.

Further changes:
1. Silence `deprecatedProperties` by allowing `return None` from the callback to signal it's not necessary to print yet another deprecation warning.
2. Spack refused to save old format `spack.yaml` saying it's forward incompatible, ironically this error is backwards incompatible, so I've changed it into a warning.

We don't really have to introduce yet another flag `spack env create --unify=...` because `spack -e . config add concretizer:unify:...` exists.

PR #21262, opened 2021-01-25
allow multiple compatible deps from CLI

becker33:
Currently, Spack can fail for a valid spec if the spec is constructed from overlapping, but not conflicting, concrete specs via the hash.

For example, if `abcdef` and `ghijkl` are the hashes of specs that both depend on `zlib/mnopqr`, then `foo ^/abcdef ^/ghijkl` will fail to construct a spec, with the error message "Cannot depend on zlib... twice".

This PR changes this behavior to check whether the specs are identical before failing.

With this PR, `foo ^/abcdef ^/ghijkl` will concretize.

As a side-effect, so will `foo ^zlib ^zlib` and other specs that are redundant on their dependencies.

PR #31787, opened 2022-07-28
concretizer:unify:true as a default

haampie:
`spack env create` enables a view by default (in a weird hidden
directory, but well...). This is asking for trouble with the other
default of `concretizer:unify:false`, since having different flavors of
the same spec in an environment, leads to collision errors when
generating the view.

A change of defaults would improve user experience:

Either views should be disabled by default, or we should set
`concretizer:unify:true`.

I guess not everybody creating an environment really uses views,
especially when they're in this hidden `.spack-env/view` folder, so in
that sense disabling views could make sense.

However, `unify:true` makes even more sense, since any time the issue is
brought up in Slack, the user changes the concretization config, since
it wasn't the intention to have different flavors of the same spec, and
as a bonus it reduces install times.

`unify: when_possible` is another possible default, but I think it's better
to hit a concretization error than it is to hit an obscure view collision
error *after* waiting an hour for the environment to build. (to be fair,
we could do a better job warning the user after concretization that
the view contains the same package twice)

---

Note, this is a follow-up to #30038, which made this change of defaults less breaking going from 0.18 -> 0.19:

> Store the default value `spack:concretizer:unify` in new `spack.yaml` files to make it easier for us to change the default of unify: false to unify: true in 0.19.

### DIFF
--- a/etc/spack/defaults/concretizer.yaml
+++ b/etc/spack/defaults/concretizer.yaml
@@ -33,4 +33,4 @@ concretizer:
   # environments can always be activated. When "false" perform concretization separately
   # on each root spec, allowing different versions and variants of the same package in
   # an environment.
-  unify: false
+  unify: true
--- a/lib/spack/docs/environments.rst
+++ b/lib/spack/docs/environments.rst
@@ -519,27 +519,33 @@ available from the yaml file.
 ^^^^^^^^^^^^^^^^^^^
 Spec concretization
 ^^^^^^^^^^^^^^^^^^^
-An environment can be concretized in three different modes and the behavior active under any environment
-is determined by the ``concretizer:unify`` property. By default specs are concretized *separately*, one after the other:
+An environment can be concretized in three different modes and the behavior active under
+any environment is determined by the ``concretizer:unify`` configuration option.
+
+The *default* mode is to unify all specs:
 
 .. code-block:: yaml
 
    spack:
        specs:
-         - hdf5~mpi
          - hdf5+mpi
          - zlib@1.2.8
        concretizer:
-         unify: false
+         unify: true
 
-This mode of operation permits to deploy a full software stack where multiple configurations of the same package
-need to be installed alongside each other using the best possible selection of transitive dependencies. The downside
-is that redundancy of installations is disregarded completely, and thus environments might be more bloated than
-strictly needed. In the example above, for instance, if a version of ``zlib`` newer than ``1.2.8`` is known to Spack,
-then it will be used for both ``hdf5`` installations.
+This means that any package in the environment corresponds to a single concrete spec. In
+the above example, when ``hdf5`` depends down the line of ``zlib``, it is required to
+take ``zlib@1.2.8`` instead of a newer version. This mode of concretization is
+particularly useful when environment views are used: if every package occurs in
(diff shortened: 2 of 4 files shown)


tgamblin:
I agree with @haampie's rationale here -- how about:

* default to `unify:true` for envs created with a view
* default to `unify:when_possible` for envs created `--without-view`

The only decision I see remaining here is what the global default should be -- I guess it should be `unify:true` as long as the default for `spack env create` is to have a view.  Any thoughts on this?

haampie:
Not sure what the point of global config is if environments always set the config in their own scope?

psakievich:
IMO I think `unify:true` would be best to set as the default.  Whether it's through the global config or the default manifest makes no difference to me.  However, I think it would also make a lot of sense to make an option for env creation to set the unify status and give the users the freedom to determine this on a case by case basis.

We can try to anticipate all the needs and best use cases but given the flexibility environments provide I expect the percentage of cases where the default doesn't meet end user needs will continue to be non-trivial.
```console
spack env create --unify when_possible -d ~/my_when_possible
spack env create --unify true -d ~/my_true
spack env create --unify false -d ~/my_false
```

haampie:
Notice you can set that with `spack -e my_when_possible config add concretizer:unify:when_possible` too

psakievich:
Sure.  I think this syntax is fine, but it is just another command you need to execute in the environment creation/activation/utilization process.  If you know what you want at creation then I think it should be a creation argument.

psakievich:
In terms of the original argument problem of view clashing with `unify: false` I agree that `unify:true` is better than disabling views.  Personally my default is no views and `unify:true` but I've seen many users rely on binaries getting loaded into `PATH` through the view which is a nice feature.

becker33:
> IMO I think unify:true would be best to set as the default. Whether it's through the global config or the default manifest makes no difference to me. However, I think it would also make a lot of sense to make an option for env creation to set the unify status and give the users the freedom to determine this on a case by case basis.

I agree with @psakievich here.

haampie:
I'm still in the camp that thinks `spack config add concretizer:unify:when_possible` is fine :tm:

haampie:
I still didn't change my mind :p `view` is a prop of an env, `unify` is a special concretizer config property. Yes it applies to environments, but it's in a separate section. For config changes, we have a separate setting. Also you can just change the global default and it will apply to any new env

alalazo (dismissed):
I would be fine merging this as is, without a config option. We can add it later if there's request for that. @becker33 @tgamblin thoughts?

haampie:
This is now also less pedantic like #21262, it doesn't make sense not to allow `specs: [python, python]` or `[python@3, python@3.10]` with `unify:true`; should just work? Instead we catch the concretization error and improve it a little bit.

Now we get:

```yaml
spack:
  specs: [python@3.10, python@3.9]
  view: true
  concretizer:
    unify: true
```

```console
==> Error: No version for 'python' satisfies '@3.10.8' and '@3.9.15'. Consider setting `concretizer:unify` to `when_possible` or `false` to relax the concretizer strictness.
```

alalazo (approved):

haampie:
@spackbot run pipeline

PR #31790, opened 2022-07-28
Deprecate visit 3.3.0 because it is broken

scottwittenburg:
Based on this [comment](https://github.com/spack/spack/pull/31790), it seems like deprecating version `3.3.0` is the best way to fix the currently broken pipelines by making spack prefer the previous (working) version.  When the new version `3.3.1` is added, that should become the preferred version.

### DIFF
--- a/var/spack/repos/builtin/packages/visit/package.py
+++ b/var/spack/repos/builtin/packages/visit/package.py
@@ -69,7 +69,7 @@ class Visit(CMakePackage):
     executables = ['^visit$']
 
     version('develop', branch='develop')
-    version('3.3.0', sha256='1a7485146133ac5f1e330d9029697750046ef8d9e9de23a6c2a3685c1c5f4aac')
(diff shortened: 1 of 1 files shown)


cyrush:
Q: what does `deprecated` imply, vs changing the preferred version?

scottwittenburg:
> Q: what does `deprecated` imply, vs changing the preferred version?

I think the reason @tgamblin suggested deprecation was because of your comment that there was a known bug in the build system of that version, from which I inferred we would always want to avoid that version.  Also, if we actually marked the previous version as preferred, we'd just have to remove that once the next version comes out.  This way, when the next version is added, it will automatically become the one spack chooses (in the absence of conflicting constraints). 

But I may have misunderstood his intentions there.

tldahlgren:
FWIW.  The doc for the option is at https://spack.readthedocs.io/en/latest/packaging_guide.html?highlight=deprecate#deprecating-old-versions.

tgamblin (approved):

Issue #21744, opened 2021-02-17
Installation issue: perl


tepperly:
### Steps to reproduce the issue

```console
$ spack install gcc@7.3.0+bootstrap+binutils^libc+iconv
$ spack compiler find
==> Added 1 new compiler to /home/epperly2/tmp/spack/var/spack/environments/ol6/spack.yaml
    gcc@7.3.0
==> Compilers are defined in the following files:
    /home/epperly2/.spack/linux/compilers.yaml
    /home/epperly2/tmp/spack/var/spack/environments/ol6/spack.yaml
$ spack install emacs%gcc@7.3.0
[+] /home/epperly2/tmp/spack/opt/spack/linux-ol6-core2/gcc-4.4.7/pkgconf-1.7.3-ridnfohoheeikphfhwncij64mpys4hsr
[+] /home/epperly2/tmp/spack/opt/spack/linux-ol6-core2/gcc-4.4.7/berkeley-db-18.1.40-zrmu5cu2veekhethshmrdrj36ltfovf3
[+] /home/epperly2/tmp/spack/opt/spack/linux-ol6-core2/gcc-4.4.7/zlib-1.2.11-xjsck64mrww72obm747mseqxurjuod4u
==> Installing nasm-2.15.05-nkob4xmk3mq6ip32gm2xutx6mcpclhib
==> Warning: Spack will not check SSL certificates. You need to update your Python to enable certificate verification.
==> No binary for nasm-2.15.05-nkob4xmk3mq6ip32gm2xutx6mcpclhib found: installing from source
==> Using cached archive: /home/epperly2/tmp/spack/var/spack/cache/_source-cache/archive/3c/3caf6729c1073bf96629b57cee31eeb54f4f8129b01902c73428836550b30a3f.tar.xz
==> nasm: Executing phase: 'autoreconf'
==> nasm: Executing phase: 'configure'
==> nasm: Executing phase: 'build'
==> nasm: Executing phase: 'install'
[+] /home/epperly2/tmp/spack/opt/spack/linux-ol6-core2/gcc-4.4.7/nasm-2.15.05-nkob4xmk3mq6ip32gm2xutx6mcpclhib
==> Installing libiconv-1.16-7roszuprcilqm4oqgdtvklcqpir7jyg4
==> Warning: Spack will not check SSL certificates. You need to update your Python to enable certificate verification.
==> No binary for libiconv-1.16-7roszuprcilqm4oqgdtvklcqpir7jyg4 found: installing from source
[...]

PR #31796, opened 2022-07-29
gdbm: restrict patch to 1.11:1.18, add nls variant

iarspider:
File to patch missing in 1.10. Also restruct `readline` dependency to 1.13 and newer.


var/spack/repos/builtin/packages/gdbm/package.py
```diff
@@ -29,15 +29,17 @@ class Gdbm(AutotoolsPackage, GNUMirrorPackage):
     version('1.9.1',  sha256='6025852637772b0699f2294b5f14fd4a084bca3c8161d29d64d1f30d6d1a9aed')
     version('1.9',    sha256='f85324d7de3777db167581fd5d3493d2daa3e85e195a8ae9afc05b34551b6e57')
 
-    depends_on("readline")
+    variant('nls', default=True, description="Use Native Language Support")
+
+    depends_on("readline", when="+nls")
```
haampie: This looks really weird to me. What does translation have to do with command line stuff? Don't you mean to add a dependency on gettext when +nls?

iarspider: @haampie good catch.

iarspider:
I wish the error messages were a bit more detailed... :(

iarspider:
The problem with the pipeline is: perl build fails because of `undefined reference to 'libintl_textdomain'`. There is an open ticket in Perl bugtracker https://github.com/Perl/perl5/issues/18467, in Spack bugtracker https://github.com/spack/spack/issues/21744 . There is a possible fix here: https://github.com/buildroot/buildroot/commit/e0a291244002ab19f8fc8a0adb215b51888eb5aa . Should I apply the fix to Spack recipe?

haampie:
Not sure if that fix is correct. My guess is that Spack's implicit `-I` flags influence Perl's configure script, making it believe that certain things are in glibc even though they're coming from gettext (and require linking). In that case adding `-lintl` would just be symptom relief (and possibly just wrong?)...

iarspider:
@spackbot fix style

bernhardkaindl (dismissed):
@iarspider: gdbm's configure enabled NLS by default already:
```py
spack-stage-gdbm-1.20-jrtkagtq2vmk4focstlpqc7ul7rzof6p/spack-src/configure --help | grep nls
  --disable-nls           do not use Native Language Support
```
And gdbm prefers using GNU gettext from GLibc, not from the separate library:
```py
checking for GNU gettext in libc... yes
checking whether to use NLS... yes
checking where the gettext function comes from... libc
```
This means that it is better to not add the `gettext` package to the build because it's:
- Not used by `gdbm`
- It's new presence can only break dependents of `gdbm` like `perl`.

The resulting purpose of your that variant is to allow to disable NLS in `gdbm` when not wanted.
Maybe this could help the build of gdbm on some strange systems without a fully-working glibc.

Because you didn't explain what your motivation to add the variant was, for now I am assuming this.

I rebased your commits to the current develop, squashed them and pushed them with my fixup including explanatory comments to avoid more confusion in the future.

iarspider:
@spackbot run pipeline

iarspider:
@bernhardkaindl thanks for fixing this PR. I will check why I needed `gettext` dependency (I clearly don't need it now).

bernhardkaindl:
@iarspider Can you check if #34286 fixes your `perl` build problem?

The bug reports you refer to are quite old, what is your exact environment and what is your full issue description?

Can you describe in which environement (host OS, etc) your problem happen? (or e.g. open a bug report for your issue with all details)?

Unless you or gdbm need to use gettext, I'd prefer if we could just disable NLS and not even make it a variant.

A very common issue with packages which enable gettext is that they are apparently only built on systems using `glibc` as the system libc, and `glibc` includes e.g. the functions which gettext provides in `libintl` and glibc also provides `/usr/include/libintl.h`. From Rocky Linux 8:

```py
$ rpm -qf /usr/include/libintl.h 
glibc-headers-2.28-216.el8.aarch64
```

When a user then tries to build such package on a non-glibc system, they might be missing `libintl.h` or the `libintl` functions are not found because they are not included in the other system libc variants, e.g. such as `musl` of Alpine Linux, or on Darwin(MacOSX). Some spack packages addressed missing `libintl` functions by adding
```py
        depends_on("gettext")
```
and sometimes even need to add:
```py
            env.append_flags("LDFLAGS", "-lintl")
```
or
```py
            args.append("-DCMAKE_SHARED_LINKER_FLAGS=-lintl")
```
But adding `-lintl` should be done only if `gettext` provides `libintl`, which is not the case when spack uses the system's `gettext` after `spack external find gettext` on glibc systems because these don't have a `libintl` provided by their system libs (since the functions are already in glibc).

glibc really mad a mess there, sigh, but that ship has sailed a loooong time ago. Here is what such packages should do then:
```py
        # Append -lintl if building against a gettext which provides a separate libintl:
        # This prevents build failures when `spack external find gettext` was used on glibc systems.
        # setup spack to use the system-provided gettext and on glibc systems libintl is "in" glibc:
        if any("libintl." in filename.split("/")[-1] for filename in self.spec["gettext"].libs):
            env.append_flags("LDFLAGS", "-lintl")
```
See #34383 if you want to read up more on that.

PR #31798, opened 2022-07-29
iwyu: new version 0.18

sethrj:


sethrj:
It builds, but on my more recent builds (using M1/M2), there's a segfault when executing. It looks like on these machines, libc++ is being loaded via rpath:
```
          cmd LC_LOAD_DYLIB
      cmdsize 48
         name @rpath/libc++.1.dylib (offset 24)
```
which causes it to incorrectly point to the LLVM rather than system `libc++`.
```
          cmd LC_RPATH
      cmdsize 96
         path /opt/spack/opt/spack/monterey/llvm/14.0.6/ozdt5jsfgrstq4xcnys7cn4oat7ptgxj/lib (offset 12)
```

On my older Spack installation, where iwyu works, `libc++` is *not* an rpath. Not sure what's happening here but the behavior changed between February and July of this year...

tldahlgren (approved):
Confirmed sha256.

PR #31800, opened 2022-07-29
Improve "go-bootstrap" failure msg on macOS

sethrj:
This message shows up when trying to build (e.g.) git-lfs on a mac running the latest system or latest chipset.


bernhardkaindl (approved):
Due to the reformat of the recipe by black meanwhile, I re-applied the commit manually

PR #30374, opened 2022-04-28
Add BUILD_TESTING to standard CMake arguments

fsimonis:
CTest determines whether to enable tests using the [`BUILD_TESTING`](https://cmake.org/cmake/help/latest/module/CTest.html?highlight=build_testing) variable.
This should be used by projects to conditionally enable the compilation of tests.
Spack knowns which packages are requested to run tests and can thus automatically define this variable for CMake packages.

This addition should
* have no observable consequences, and
* reduce compile time for _every_ CMake project using `BUILD_TESTING`.

PR #31805, opened 2022-07-29
CMake: Override GNU Install paths for LIBDIR

kwryankrattiger:
There are a number of issues using GNU Install paths when it is lib64 inside of spack. This overrides that behavior to always point at lib.

### DIFF
--- a/lib/spack/spack/build_systems/cmake.py
+++ b/lib/spack/spack/build_systems/cmake.py
@@ -179,6 +179,7 @@ class CMakePackage(PackageBase):
             define("CMAKE_INSTALL_PREFIX", convert_to_posix_path(pkg.prefix)),
             define("CMAKE_BUILD_TYPE", build_type),
             define("BUILD_TESTING", pkg.run_tests),
+            define("CMAKE_INSTALL_LIBDIR", "lib"),
         ]
 
         # CMAKE_INTERPROCEDURAL_OPTIMIZATION only exists for CMake >= 3.9


kwryankrattiger:
I am not 100% on how this should behave on windows or mac, but for cray and other linux platforms this should work.

haampie:
Not a big fan of overriding non-standard variables. This is not a CMake builtin, it's defined by the GNUInstallDirs module. My hesitation is mostly that (a) some package will somehow depend on GNUInstallDirs behavior and (b) there may be variables derived from `CMAKE_INSTALL_LIBDIR` that are now incorrect.

([Same for BUILD_TESTING](https://github.com/spack/spack/pull/30374#issuecomment-1121538055) btw, I'd rather see that change reverted... in this case, some CMake projects rely on the documented `include(CTest)` setting `BUILD_TESTING=ON`, which is no longer true because spack may set it to OFF on the command line)

kwryankrattiger:
I think the way BUILD_TESTING is configured there is probably fine since it is turned on/off by `pkg.run_tests`

As far as the `GNUInstallDirs` stuff, I think that it is correct to say it is a standard part of CMake, but it is not part of the default environement. Similar to inlcuding `<iostream>` in a C++ project. There are deeper issues with using the deduced lib directory both in the projects in spack as well as the spack itself that  make it problematic to just leave. I think it is safe to override it, but it may require additional platform checking for Windows and MacOS that I haven't done here.

robertu94:
I'm also going to agree with @haampie here.  This change might possibly break things when multiple binary formats are supported by libraries, and goes against the Filesystem Hierarchy Standard.  Specifically, I've worked with several libraries that generate their own pkg-config or generated header  files, and while the CMake export functionality is "correct", they hard code the lib location in these generated files which can cause issues.

kwryankrattiger:
@robertu94 

It is worth noting, the primary reason for me moving to this more drastic patch rather than just fixing it on a per-package basis is I have been observing strange and inconsistent behavior with CMake inside of spack. One of the most frustrating issues I am finding is CMake will just ignore search paths that are not under `lib`, everything under `lib64` is just ignored. In some cases, this was the fault of a package hard coding paths for third-party packages, which I fixed and back ported for those projects. But in other cases it just ignored the path entirely. Tracing the cmake configuration manually did reveal anything obvious to me as to why. There are many oddities inside the spack build environment that I have encountered when using cmake, and I am ever edging closer to just figuring out why and fixing the core issue, but this was a much more economical fix in the short term.

> This change might possibly break things when multiple binary formats are supported by libraries

I am not sure how this is possible since I don't believe spack supports this. If there are cases where packages in spack are installing multiple formats under the same hash, I am happy to expose an option for those packages to disable this logic. Could you point me to some examples of this?

>  goes against the Filesystem Hierarchy Standard

AFAIK, the FHS states the `lib<##>` directories are optional, and only have additional rules on what may go in them, not that libraries _must_ go in them. I may be wrong, so please correct me if I am, I am only vaguely familiar with the standards rules.

> they hard code the lib location in these generated files which can cause issues.

Reading your comment I just want to clarify one thing. Setting this variable does not change the default location libraries are installed when using CMake, it is still up to the package to make the decision to use GNUInstallDirs paths, builtin-defaults, or some other hard coded option. If they are hard coding paths, this change will do nothing. If they are using `GNUInstallDirs`, this will simply say the LIBDIR is now `lib` always. The built-in default for CMake is always `lib`, regardless of platform.

@haampie I dug a little deeper on you comment on `BUILD_TESTING` and I am moved to agree with you.

robertu94:
@kwryankrattiger 

>One of the most frustrating issues I am finding is CMake will just ignore search paths that are not under lib, everything under lib64 is just ignored. In some cases, this was the fault of a package hard coding paths for third-party packages, which I fixed and back ported for those projects

This is interesting to me.  I have most of my libraries installed in lib64, and CMake seems to find them just fine.  There is also an option: [FIND_LIBRARY_USE_LIB64_PATHS](https://cmake.org/cmake/help/latest/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html#prop_gbl:FIND_LIBRARY_USE_LIB64_PATHS) which looks for lib64 variants with find_library.  I'd be curious as for the real root cause here.

> Could you point me to some examples of [multiple binary formats are supported by libraries]?

Good point.  I couldn't find an example of a package that does this with my current set of installed packages.  The ones that I thought did OpenBLAS and Cuda do something else.

> AFAIK, the FHS states the lib<##> directories are optional

You are correct, these directories are optional (didn't know that 😄 ; I thought that was meant to convey that you don't have to have the directory if it is going to be empty). However, it is worth noting the default behavior of other things.  Ubuntu, CentOS,  Fedora, and Gentoo all by default use lib64 as do the default behavior of CMake.  It appears that Anaconda and Julia's Binary Builder doesn't use it, but Python's Pip does.  In short we are probably going to have to live with both directories existing and different libraries putting them in different places.  However, I think there is an open question of how to do this.

kwryankrattiger:
`FIND_LIBRARY_USE_LIB64_PATHS` doesn't have an analogous version for the Cofig search, which is where the issue comes up. So cmake just fails to find things like `lib64/cmake/SENSEIConfig.cmake` but it will find it when it is in `lib/cmake`.  Then the library search stuff is all relative to that, handled by the target export stuff.

robertu94:
@kwryankrattiger According to [The CMake Config Search Procedure](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-) these are the paths searched for Config files for Unix systems
```
<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
<prefix>/(lib/<arch>|lib*|share)/<name>*/                       (U)
<prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/         (U)
<prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/         (W/U)
<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/               (W/U)
<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/ (W/U)
```

These are the paths searched.  Which includes `<prefix>/lib64/cmake` as `<prefix>/lib*/cmake`  My guess is you've either found a bug in CMake (certainly possible, I've found a few issues in CMake over the years), or there is a typo somewhere.

kwryankrattiger:
Yeah, it is supposed to. But running `cmake --debug-find` it doesn't print those as searched paths, only `lib` which is the core of that problem. But when I run similar examples on those systems it works. So there is something embedded in the spack environment that I am missing.

kwryankrattiger:
Dropping this for now.

PR #31809, opened 2022-07-29
Add meson Python build backend

adamjstewart:
This PR adds Yet Another Build Backend (YABB) to the Python ecosystem. In particular, the latest version of scipy switched from setuptools to meson for its builds. I also updated the documentation to describe this new build backend. Since almost all build backends share the same format for listing dependencies, I consolidated much of this information so I only list differences between each build system.


adamjstewart:
@spackbot fix style

alalazo (approved):

PR #28926, opened 2022-02-14
darwin: make sure MACOSX_DEPLOYMENT_TARGET has a minor component

sethrj:
Follow-on to #28797 . 

Older versions of GCC (8.3 and probably others) directly propagate `MACOSX_DEPLOYMENT_TARGET` to `ld` rather than padding with `.0.0` like newer ones. This leads to a failure:
```console
$ ld -macosx_version_min 11
ld: -macos_version_min value malformed: '11'
```
whereas `ld -macosx_version_min 11.0 ...` works.

So this commit tests for the macOS version having only a single component and adds `.0` if that's the case. This choice is consistent with what CMake does and ensures that libraries on the same major OS are all compatible.

PR #28797, opened 2022-02-05
macOS: always set `MACOSX_DEPLOYMENT_TARGET`

sethrj:
Closes #28090 . I'm leaving this in draft until I build out a new toolchain (GCC, HDF5, trilinos, ROOT, geant4, and more) on bigsur, and perhaps monterey? But this will hopefully resolve the crazy barrage of 
```
ld: warning: dylib (../libgfortran.dylib) was built for newer macOS version (11.5) than being linked (11.0)
```
warnings when using GCC+clang.

Issue #28090, opened 2021-12-20
Add MACOSX_DEPLOYMENT_TARGET on Darwin


sethrj:
### Summary

Define `MACOSX_DEPLOYMENT_TARGET` in the build environment on Darwin platforms based on the target OS (e.g. `bigsur`).

### Rationale

The `MACOSX_DEPLOYMENT_TARGET` environment variable provides a default `-mmacosx-version-min` argument for GCC and Clang compilers, as well as the default value of `CMAKE_OSX_DEPLOYMENT_TARGET` for CMake-based build systems. The default value for the deployment target is usually the major version (11, 10.16, ...) for CMake and Clang, but some versions of GCC specify a minor component as well (11.3), leading to numerous link warnings about inconsistent or incompatible target versions. Setting the environment variable ensures consistent versions for an install toolchain target, even when the host macOS version changes.

### Description

Add a `setup_platform_environment` to `spack/platforms/darwin.py` that sets the variable based on 

### Additional information

- Error caused by user environment's deployment version: #28014 
- Possibly related issue due to toolchain version and build environment variables: #26290
- Example of confusing warnings caused due to inconsistent gcc/clang treatment of min-macos-version: #21644
- Fallout from the GCC linker target in py-scipy: #20367

```
ld: warning: dylib (../libgfortran.dylib) was built for newer macOS version (11.1) than being linked (11.0)
```

```console
$ spack --version
0.17.0-588-1a576e732e
```

### General information

- [X] I have run `spack --version` and reported the version of Spack
- [X] I have searched the issues of this repo and believe this is not a duplicate

sethrj:
Looks like ROOT's fork of LLVM doesn't support contemporary versions of this variable: see #28161

trws:
What happened to #28797? I just ran into this actually killing a build due to the variable not being set while fixing up a newer version of luajit.

sethrj:
@trws That's my bad, sorry: thought the branch was on my fork, not the mainline. I'm actually working on it this very moment: I've got one patch in place for perl (which doesn't like 11+) and am hitting an unrelated issue with `go` since I'm trying to install on an `m1` Mac for the first time.

trws:
No worries, just happy to see you're already working on it.  It seems pre-10.5 versions didn't support `-rpath` on `ld`, and if it isn't specified in the environment `ld` just dies sometimes, I found this with luajit current so it's an easy repro.

sethrj:
[...]

PR #31811, opened 2022-07-29
MACOSX_DEPLOYMENT_TARGET: MAJOR.MINOR

adamjstewart:
Previously, Spack was setting `MACOSX_DEPLOYMENT_TARGET` to "11" or "12" instead of "11.6" or "12.4". Many packages like `py-scikit-build` expect this env var to have at least a major and minor version number. Without this change, `py-ninja` (which uses `py-scikit-build` to build) fails to build.

Related to:

* #28991
* #28926
* #28797
* #28090

It looks like #28926 was designed to fix this same issue but doesn't seem to be working. @sethrj thoughts on this approach and whether we should revert the hack in #28926 or fix it?

### DIFF
--- a/lib/spack/spack/operating_systems/mac_os.py
+++ b/lib/spack/spack/operating_systems/mac_os.py
@@ -150,9 +150,8 @@ class MacOs(OperatingSystem):
         # only used the minor component)
         part = 1 if version >= Version('11') else 2
 
-        mac_ver = str(version.up_to(part))
-        name = mac_releases.get(mac_ver, "macos")
-        super(MacOs, self).__init__(name, mac_ver)
+        name = mac_releases.get(str(version.up_to(part)), "macos")
+        super(MacOs, self).__init__(name, version.up_to(2))
 
     def __str__(self):
         return self.name
--- a/lib/spack/spack/operating_systems/mac_os.py
+++ b/lib/spack/spack/operating_systems/mac_os.py
@@ -150,8 +150,9 @@ class MacOs(OperatingSystem):
         # only used the minor component)
         part = 1 if version >= Version('11') else 2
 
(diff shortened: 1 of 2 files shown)


lib/spack/spack/operating_systems/mac_os.py
```diff
@@ -150,9 +150,8 @@ def __init__(self):
         # only used the minor component)
         part = 1 if version >= Version('11') else 2
 
-        mac_ver = str(version.up_to(part))
-        name = mac_releases.get(mac_ver, "macos")
-        super(MacOs, self).__init__(name, mac_ver)
+        name = mac_releases.get(str(version.up_to(part)), "macos")
+        super(MacOs, self).__init__(name, version.up_to(2))
```
sethrj: Actually this is **not** what we want. If we include the true minor version, then "monterey"-generated code on one machine may require a newer SDK than "monterey"-generated code on another machine. (Think what would happen if your binary cache–generating machine is 12.5 and you're running 12.4.) This kind of situation is what leads to thousands of warnings about mismatched SDK errors.

sethrj (changes requested):

sethrj:
Definitely try to fix by adding a `.0`. I know that my hack fixed the issue for GCC (I can't imagine *not* checking myself that the deployment target got the extra zero...) Do you know where the hack is breaking?

adamjstewart:
Thanks for the explanation. I believe the latest commit should also fix the bug. Your previous solution wasn't working because `version` was a string, so `len(version) == len("12") == 2`.

sethrj:
🤦 I now remember encountering that and fixing it... I wonder if i didn't commit and push. I couldn't think of a way to test this variable robustly; can you?

sethrj (approved):
Tada! Thanks for the fix.

PR #31812, opened 2022-07-29
Add support for macOS Ventura

adamjstewart:
Hasn't been released yet, but we know the official name now, so it's probably best to get this in sooner rather than later.

@alalazo do we need to update archspec too?


alalazo:
> do we need to update archspec too?

I don't think so, unless the new OS also changes something in the output of `sysctl` and we have to rework detection.

alalazo (approved):

PR #31346, opened 2022-06-28
Allow building octave w/ Intel compiler

davidedelvento:
Fixes https://github.com/spack/spack/issues/6647 (I know it's trivial, but for the poor souls building complex environments, each small thing like this contributes to a death by papercuts!!)

PR #31813, opened 2022-07-29
Allow building octave w/ Intel compiler

davidedelvento:
I think there was an issue for this, but I can't find it. This works for me, and it doesn't without it.


siko1056 (approved):
Looks identical to #31346 ?  Thus no further comments.  Did the previous merge not succeed @davidedelvento @tldahlgren ?

davidedelvento:
You are right the previous merge was successful, this was operator error because he was on the v0.18 branch which was behind the merge. I could partially blame COVID and a long week for having overlooked that detail.
Sorry.

PR #31814, opened 2022-07-29
making qtconsole optional

davidedelvento:
The console install, with its huge Qt and related dependencies, should be optional and off by default since most people would want to use the web, rather than Qt interface

### DIFF
--- a/var/spack/repos/builtin/packages/py-jupyter/package.py
+++ b/var/spack/repos/builtin/packages/py-jupyter/package.py
@@ -14,11 +14,13 @@ class PyJupyter(PythonPackage):
 
     version('1.0.0', sha256='d9dc4b3318f310e34c82951ea5d6683f67bed7def4b259fafbfe4f1beb1d8e5f')
 
+    variant('qtconsole', default=False, description='Enable Qt console')
+
     # pip silently replaces distutils with setuptools
     depends_on('py-setuptools', type='build')
     depends_on('py-notebook', type=('build', 'run'))
-    depends_on('py-qtconsole', type=('build', 'run'))
-    depends_on('py-jupyter-console', type=('build', 'run'))
+    depends_on('py-qtconsole', type=('build', 'run'), when='+qtconsole')
+    depends_on('py-jupyter-console', type=('build', 'run'), when='+qtconsole')
     depends_on('py-nbconvert', type=('build', 'run'))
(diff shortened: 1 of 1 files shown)


adamjstewart (commented):
Two comments here:

1. The package lists `py-jupyter-console` as required, not optional
2. This package is no longer used and hasn't had a new release in almost 7 years, I think you want one of the hundreds of other Jupyter packages

davidedelvento:
Thanks for getting back so quickly.

> Two comments here:
>     1. The package lists `py-jupyter-console` as required, not optional

Not sure what you mean. The whole purpose of the PR is to make it optional, am I missing a step?

>     2. This package is no longer used and hasn't had a new release in almost 7 years, I think you want one of the hundreds of other Jupyter packages

There aren't "hundreds" but only less than a dozen Jupyter packages. If you point me in the right direction (what to install instead for having the jupyter client to install alternative kernels) I'll be happy to use one of them. Moreover, if this is "no longer used" it should be deleted, or at least have a deprecation message, to avoid lots of suffering for poor souls like myself who are still learning how to use spack....

adamjstewart:
> Not sure what you mean. The whole purpose of the PR is to make it optional, am I missing a step?

What I mean is that the setuptools build system in the package (see `setup.py`) lists it as a required dependency. If it were optional, it would be listed under `extras_require` instead of `install_requires`.

> If you point me in the right direction (what to install instead for having the jupyter client to install alternative kernels) I'll be happy to use one of them.

Maybe `py-jupyter-client` or `py-nbclient`? Or `py-jupyter-server`, or `py-jupyter-core`, or `py-jupyter-console`... Basically, whatever you would normally `pip install`, just add `py-` to the beginning of that name. How did you install this before Spack?

davidedelvento:
I see. So you are just replicating PyPI packages in spack.

That is good to know because that was not clear to me before. You may close this PR (I am not doing it myself in case you think that removing the package altogether is better).

Thanks a lot!

adamjstewart:
> I see. So you are just replicating PyPI packages in spack.

Yep, this is somewhat documented here: https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html

> I am not doing it myself in case you think that removing the package altogether is better

Prob best to leave it in for reproducibility or old packages that used to depend on it.

Issue #31793, opened 2022-07-29
Installation issue: julia@1.7.3


bosmacs:
### Steps to reproduce the issue

```console
$ spack spec -I julia@1.7.3
==> Error: curl 'platform=darwin' conflicts with '+libssh2'
```

### Error message

<details><summary>Error message</summary><pre>
==> Error: curl 'platform=darwin' conflicts with '+libssh2'
</pre></details>

### Information on your system

* **Spack:** 0.19.0.dev0 (4f57430ef36dab4e84532ac705c6fefa38965257)
* **Python:** 3.9.13
* **Platform:** darwin-monterey-m1
* **Concretizer:** clingo

### Additional information

@glennpj @haampie @vchuravy Trying to install julia on an M1 Mac, I'm getting an error before even getting to the build stage:
```
==> Error: curl 'platform=darwin' conflicts with '+libssh2'
```
The curl conflict is well documented [in this issue](https://github.com/spack/spack/issues/7777).

In a (perhaps naïve) attempt to work around the conflict, I get the same error when attempting to specify a the `~libssh2` variant of curl:
```
spack spec -I julia@1.7.3 ^curl~libssh2
==> Error: curl 'platform=darwin' conflicts with '+libssh2'
```
`curl~libssh2` installs fine on its own. I assume the libssh2 dependency of julia is propagating to curl despite the dependency spec (I'm still a spack beginner, so may be specifying it incorrectly). 

This seems like a roadblock to installing julia with spack on darwin; are there any known workarounds?

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have run `spack maintainers <name-of-the-package>` and **@mentioned** any maintainers
- [X] I have uploaded the build log and environment files
- [X] I have searched the issues of this repo and believe this is not a duplicate

haampie:
Right, this would require some digging into how julia builds curl for macos. The spack recipe explicitly depends on `curl +libssh2`, that's why you get a concretization failure. Can you check if by magic `curl+libssh2` builds fine on macos now (you have to update the recipe: `spack edit curl` and remove the line `conflicts('platform=darwin', when='+libssh2')`)? I don't think anybody really tried to fix the issue reported in #7777.

(Btw, I don't think Julia 1.7 officially supports aarch64, @vchuravy knows this better than me)

Digging:

[...]

PR #31816, opened 2022-07-30
curl: --with-pkg=prefix instead of --with-pkg

### LINKED ISSUE #7777: spack install curl+libssh2 fails for Mac OS 10.13.4
Summary

The installation for curl for spack fails for Mac OS 10.13.4 (at least).  The default preferred version is 7.56.0. This package installed correctly when macports is used with default version 7.59.

### Expected Result
The package would be installed with no errors.

### Actual Result

The installation fails with the message

configure: error: one or more libs available at link-time are not available run-time. Libs used at link-time: -lssh2 -lssh2  -lssl -lcrypto -lssl -lcrypto  -lldap -lz

[LOG.txt](https://github.com/spack/spack/files/1917450/LOG.txt)

[spack-build.txt](https://github.com/spack/spack/files/1917459/spack-build.txt)

The above is really spack-build.out. I changed out to txt to make your interface happy.

run
spack install curl
from a terminal on a Macintosh system running OS 10.13.4/

This includes:

 MacPro Late 2013 OS 10.13.4 3.5 GHz 6-Core Xeon E5
This is a standard configuration. The compile set is clang@9.1.0-apple

haampie:
Make life a little bit easier for curl to locate libraries.

(Note: related to #7777, however, it does not fix that, because the issue there has to do with certain libraries setting incorrect install_names for their libraries, and Spack's `fix_darwin_install_name` is not enough to fix that.)

### DIFF
--- a/var/spack/repos/builtin/packages/curl/package.py
+++ b/var/spack/repos/builtin/packages/curl/package.py
@@ -81,7 +81,7 @@ class Curl(AutotoolsPackage):
     )
     variant("nghttp2", default=False, description="build nghttp2 library (requires C++11)")
     variant("libssh2", default=False, description="enable libssh2 support")
-    variant("libssh", default=False, description="enable libssh support")  # , when='7.58:')
+    variant("libssh", default=False, description="enable libssh support", when="@7.58:")
     variant("gssapi", default=False, description="enable Kerberos support")
     variant("librtmp", default=False, description="enable Rtmp support")
     variant("ldap", default=False, description="enable ldap support")
@@ -94,14 +94,6 @@ class Curl(AutotoolsPackage):
         description="Build shared libs, static libs or both",
     )
 
-    conflicts("+libssh", when="@:7.57")
-    # on OSX and --with-ssh the configure steps fails with
-    # one or more libs available at link-time are not available run-time
-    # unless the libssh are installed externally (e.g. via homebrew), even
-    # though spack isn't supposed to know about such a libssh installation.
-    # C.f. https://github.com/spack/spack/issues/7777
-    conflicts("platform=darwin", when="+libssh2")
-    conflicts("platform=darwin", when="+libssh")
     conflicts("platform=cray", when="tls=secure_transport", msg="Only supported on macOS")
     conflicts("platform=linux", when="tls=secure_transport", msg="Only supported on macOS")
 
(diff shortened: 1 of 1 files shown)


alalazo (dismissed):

alalazo:
@haampie LGTM and I trust you tested installations, so this requires only to be blackened before it gets merged.

haampie:
It worked for me (macOS Catalina 10.15.3), but @bosmacs is reporting it does not solve his issue https://github.com/spack/spack/issues/31793#issuecomment-1200210029.

Maybe @adamjstewart can check too? Would be good to know what triggers the problem exactly.

adamjstewart (commented):
Confirmed that `curl+libssh+libssh2` builds for me on macOS 12.4 (Apple M1 Pro) with Apple Clang 13.1.6 using this PR. But also confirmed that it builds for me on develop if I remove the conflicts. So I'm not in a position to properly test this PR since I'm unable to reproduce the previous build issues.

haampie:
I have a hard time reproducing the original issue even without the PR. However, with this PR at least this warning is silenced:

```
configure: using LIBS: -lssh2 -L/Users/harmenstoppels/spack/opt/spack/darwin-catalina-broadwell/apple-clang-11.0.0/libssh2-1.10.0-ayzs2zdu6nnsnxq64poynodxa2z3gtja/lib -lssh2 -lz
configure: LIBS note: LIBS should only be used to specify libraries (-lname).
```

bosmacs:
See my [updated comments](https://github.com/spack/spack/issues/31793#issuecomment-1203041818). It appears in my testing some issues remain with tls dependencies.

haampie:
@adamjstewart we can still merge this since it's a slight improvement.

The issue of #7777 is not solved though, because dependencies of curl (like `mbedtls libs=shared`) do not specify proper install names for dylibs. It can't really be fixed by `fix_darwin_install_name` either, because mbedtls produces 3 dylibs linked against each other, meaning that the install_name is already copied before we have a chance to fix it.

adamjstewart:
Should we re-add the conflicts?

haampie:
No, both libssh and libssh2 are cmake packages, and cmake sets proper install names. I think https://cmake.org/cmake/help/latest/policy/CMP0068.html and https://cmake.org/cmake/help/latest/policy/CMP0042.html could be relevant, but didn't check it in detail. I don't wanna add conflicts on a particular cmake version ;p since the install_name trouble is mostly a Spack issue imho.

alalazo (approved):

alalazo:
Merging, according to https://github.com/spack/spack/pull/31816#issuecomment-1205081502 We can add conflicts if we'll see issues later on this subject.

PR #31836, opened 2022-08-01
gromacs: add CP2K

w8jcik:
> CP2K is a quantum chemistry and solid state physics software package that can perform atomistic simulations of solid state, liquid, molecular, periodic, material, crystal, and biological systems. CP2K provides a general framework for different modeling methods such as [DFT](http://en.wikipedia.org/wiki/Density_functional_theory) using the mixed [Gaussian and plane waves approaches](https://www.cp2k.org/quickstep#gpw) GPW and GAPW.

Tested with

```
$ spack install gromacs@2022.2+cuda~mpi+cp2k ^cp2k~mpi ^dbcsr~mpi+openmp
$ spack install gromacs@2022.1+cuda~mpi+cp2k+plumed ^plumed@master ^cp2k~mpi+plumed ^dbcsr~mpi+openmp
```


alalazo:
@spackbot maintainers

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -732,10 +733,48 @@ def build(self, spec, prefix):
...
+        with open(join_path(self.prefix.lib.pkgconfig, "libcp2k.pc"), "r+") as handle:
+            content = handle.read().rstrip()
+
+            if "^fftw" in self.spec:
+                content += " -lfftw3"
+
+            if "^fftw+openmp" in self.spec:
+                content += " -lfftw3_omp"
+
+            content += " -lopenblas"
```
marvinbernhardt: I believe this line leads to problems when one wants to use something other than openblas, right?
I try to install `gromacs+cp2k ^netlib-lapack` and get `/usr/bin/ld: cannot find -lopenblas: No such file or directory`

marvinbernhardt (changes requested):

var/spack/repos/builtin/packages/cp2k/package.py
dev-zero: this has been addressed

dev-zero (approved):

var/spack/repos/builtin/packages/cp2k/package.py
w8jcik: Yes, but only `gromacs~mpi` builds so far. I am trying to make `gromacs+mpi` build as well. Not sure where is the issue, except that it fails during linking.

dev-zero: @w8jcik post the error and the full linker line leading up to the error here or on Slack, maybe someone has seen the error and can help?

marvinbernhardt: > Yes, but only `gromacs~mpi` builds so far. I am trying to make `gromacs+mpi` build as well. Not sure where is the issue, except that it fails during linking.

I see this too. First of many errors I get when building (linking) gromacs is `undefined reference to pilaenv_`. This seems to be some [scalapack function](https://netlib.org/scalapack/explore-html/d7/d8f/pilaenv_8f_source.html). This makes somehow sense, since scalapack is only used with `+mpi` in cp2k. Also I see that there is no `-lscalapack` flag in the command that trows the errors. Not sure how this should be fixed though.

marvinbernhardt (approved):
I managed to build and run a simple energy minimization. PR looks good. Thanks for your effort!

alalazo (approved):

var/spack/repos/builtin/packages/cp2k/package.py
dev-zero: maybe the `.pc` file needs some more fixing?

w8jcik: Yes, it is most likely a fault of the `.pc` file.

This `Dockerfile` builds CP2K in Spack-less way

```Dockerfile
FROM ubuntu:22.04

# https://serverfault.com/a/797318
ARG DEBIAN_FRONTEND=noninteractive

RUN apt-get update -y
RUN apt-get upgrade -y

# CUDA 11.5.1 and GCC 11.2
RUN apt-get install -y nvidia-cuda-toolkit nvidia-cuda-dev
RUN apt-get install -y build-essential cmake git

RUN apt-get install -qq --no-install-recommends \
  autoconf autogen automake autotools-dev bzip2 ca-certificates git less \
  libtool libtool-bin make nano patch pkg-config python3 unzip wget xxd zlib1g-dev

RUN apt-get install -y gcc g++ gfortran
RUN apt-get install -y gcc-10 g++-10 gfortran-10

RUN apt-get install -qq libopenblas-dev libfftw3-dev

RUN useradd -m user
USER user
WORKDIR /home/user
SHELL ["/bin/bash", "-c"]

RUN git clone --recursive -b v2022.1 https://github.com/cp2k/cp2k.git

RUN cd "${HOME}/cp2k/tools/toolchain/" \
    && ./install_cp2k_toolchain.sh --with-libxsmm=install \
        --with-openblas=system --with-fftw=system \
        --enable-cuda --gpu-ver=V100

RUN cp "${HOME}/cp2k/tools/toolchain/install/arch/"* "${HOME}/cp2k/arch/"

RUN source "${HOME}/cp2k/tools/toolchain/install/setup" \
    && cd cp2k \
    && make -j ARCH=local VERSION=ssmp \
    && make -j ARCH=local VERSION=ssmp libcp2k

# RUN source "${HOME}/cp2k/tools/toolchain/install/setup" \
#     && cd cp2k \
#     && make -j ARCH=local VERSION=ssmp test

RUN git clone -b v2022.2 https://github.com/gromacs/gromacs.git gromacs-cp2k-src

RUN source "${HOME}/cp2k/tools/toolchain/install/setup" \
    && DIR="gromacs-cp2k" \
    && cmake \
        -DBUILD_SHARED_LIBS=OFF -DGMXAPI=OFF -DGMX_INSTALL_NBLIB_API=OFF \
        -DGMX_EXTERNAL_BLAS=ON -DGMX_EXTERNAL_LAPACK=ON \
        -DGMX_CP2K=ON -DCP2K_DIR="${HOME}/cp2k/lib/local/ssmp" \
        -DGMX_GPU=cuda \
        -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 -DCMAKE_FORTRAN_COMPILER=gfortran-10 \
        -DCMAKE_INSTALL_PREFIX=`pwd`/${DIR} -S "${DIR}-src" -B "${DIR}-build" \
    && cmake --build "${DIR}-build" --target install -j $(nproc)

RUN source gromacs-cp2k/bin/GMXRC.bash \
    && gmx --version
```

My idea is to change it to build with MPI and see how the `.pc` file looks.

Other, maybe simpler option, is to analyse the script that generates `.pc`.

I will try in some time, otherwise you might find this `Dockerfile` useful.

PR #31837, opened 2022-08-01
Ascent package fixes

kwryankrattiger:
I have found issues with mixing `~btl_find_mpi` into builds on perlmutter. I am still needing to test this on crusher, but I found that this fix with `+blt_find_mpi` on cray systems works better when building downstream projects.

### DIFF
--- a/var/spack/repos/builtin/packages/ascent/package.py
+++ b/var/spack/repos/builtin/packages/ascent/package.py
@@ -107,6 +107,9 @@ class Ascent(CMakePackage, CudaPackage):
     # patch for finding Conduit python more reliably
     # https://github.com/Alpine-DAV/ascent/pull/935
     patch("ascent-find-conduit-python-pr935.patch", when="@0.8.0")
+    # patch for finding dependencies in configure file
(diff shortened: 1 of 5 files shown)


var/spack/repos/builtin/packages/ascent/package.py
```diff
@@ -232,11 +235,15 @@ def setup_build_environment(self, env):
     # Note: cmake, build, and install stages are handled by CMakePackage
     ####################################################################
 
+    @property
+    def root_cmakelists_dir(self):
```
kwryankrattiger, line 227: This fixes warnings from cmake and is the mechanism spack provides to point to a non-root source directory. The previous way seems like it would be prone to errors.

cyrush:
I am surprised that `~blt_find_mpi` had issues on cray b/c that simply avoids extra cmake logic and lets the compiler take care of MPI stuff implicitly . I guess it depends on which compiler you are using. 

At this point, I don't think there is any going back to cmake as old as 3.14, so maybe we should just remove the logic?

kwryankrattiger:
Current state seems to be working.

PR #31839, opened 2022-08-01
e4s oneapi stack: remove notes for now-fixed builds

eugeneswalker:
`parallel-netcdf` and `scr` are building now so remove the notes on why they were failing.

FYI @wspear


wspear (approved):
Looks good.

PR #31842, opened 2022-08-01
WarpX: Add hints for FindMPI

kwryankrattiger:
This seems to be needed on some cray systems and is safe on normal desktops.

### DIFF
--- a/var/spack/repos/builtin/packages/warpx/package.py
+++ b/var/spack/repos/builtin/packages/warpx/package.py
@@ -189,13 +189,18 @@ class Warpx(CMakePackage):
             self.define_from_variant("WarpX_QED_TABLE_GEN", "qedtablegen"),
         ]
 
-        with when("+openpmd"):
+        # FindMPI needs an extra hint sometimes, particularly on cray systems
+        if "+mpi" in spec:
+            args.append(self.define("MPI_C_COMPILER", spec["mpi"].mpicc))
+            args.append(self.define("MPI_CXX_COMPILER", spec["mpi"].mpicxx))
+
+        if "+openpmd" in spec:
             args.append("-DWarpX_openpmd_internal=OFF")
 
         # Work-around for SENSEI 4.0: wrong install location for CMake config
         #   https://github.com/SENSEI-insitu/SENSEI/issues/79
         if "+sensei" in spec:
-            args.append(self.define("SENSEI_DIR", join_path(spec["sensei"].prefix.lib, "cmake")))
+            args.append(self.define("SENSEI_DIR", spec["sensei"].prefix.lib.cmake))
 
         return args
 


alalazo:
@spackbot maintainers

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -187,11 +187,15 @@ def cmake_args(self):
             self.define_from_variant("WarpX_QED_TABLE_GEN", "qedtablegen"),
         ]
 
+        with when("+mpi"):
+            args.append(self.define("MPI_C_COMIPLER", spec["mpi"].mpicc))
```
ax3l: Typo in compiler

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -187,11 +187,15 @@ def cmake_args(self):
             self.define_from_variant("WarpX_QED_TABLE_GEN", "qedtablegen"),
         ]
 
+        with when("+mpi"):
```
ax3l: Please add a comment, this is odd and should read it's cray specific

kwryankrattiger: It is actually occasionally required on non-cray systems. It just happened to show up on a cray system this time.

var/spack/repos/builtin/packages/warpx/package.py
```diff
@@ -187,11 +187,16 @@ def cmake_args(self):
             self.define_from_variant("WarpX_QED_TABLE_GEN", "qedtablegen"),
         ]
 
+        # FindMPI needs an extra hint sometimes, particularly on cray systems
+        with when("+mpi"):
```
ax3l: Hi @kwryankrattiger,

I am sorry to be picky on this, but I need more information to accept this change please.

Regarding the problem:
- on which system did you see the problem?
- what was your compiler config in Spack, versions of compiler, PrgEnv+modules and environment hints?
- what was the CMake version?
- what was the exact problem and for which `spec`? Can you post the failing CMake configure and/or failing build output?

You might be aware of our configurations on Cray systems, such as:
- Perlmutter: https://warpx.readthedocs.io/en/latest/install/hpc/perlmutter.html
- Crusher: https://warpx.readthedocs.io/en/latest/install/hpc/crusher.html
- Frontier: https://warpx.readthedocs.io/en/latest/install/hpc/frontier.html

Spack related question:
- On all but the latest HPE machines, Cray does not provide `mpicc`/`mpicxx` wrappers - does Spack add its own wrappers for all systems?
  - Update from Slack:
    - https://spack.readthedocs.io/en/latest/packaging_guide.html#mpi-on-cray-machines
    - `we’ll be changing this but mpicc on crays right now is just cc`

CMake uber question: I just want to make you aware my works in CMake, which should make these quirks unnecessary and help us to compile with with the Cray compiler wrappers as well as underlying vendor compilers
- CMake 3.22+: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6264
- WIP: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7247

I am asking all this, because it's one out of multiple packages I maintain on the same software stack & build logic, and if there are underlying CMake or Spack issues, I would rather like to debug and fix them at the source than maintain custom hints.

Lastly, if we ultimately need this work-around line, let us limit it down, e.g., to platform cray.

ax3l: @kwryankrattiger please don't be discouraged by my curiosity. I am truly interested in the underlying issue :)

Of course we will go forward with this to unblock our planned workflows :)

kwryankrattiger: Hey @ax3l No problem, I appreciate the feedback, makes things better long term :) 

To answer some of you questions:
> on which system did you see the problem

This was on Perlmutter, but I have found similar issues in Crusher testing unrelated to WarpX, but other projects that use MPI such as SENSEI and Ascent.

> what was your compiler config in Spack, versions of compiler and environment hints

I am using GCC 10.3.0, I didn't try with the CCE compiler, but previously I was having issues with it.

> what was the CMake version

CMake 3.23.2

> what was the exact problem and for which spec? Can you post the failing CMake configure and/or failing build output?

I am building `warpx+ascent+sensei` concretized unified with `ecp-data-vis-sdk+sensei+ascent`. I don't have the logs anymore, when I get a chance I will re-produce them and attach them here.

> Cray does not provide mpicc/mpicxx wrappers - does Spack add them?

So cray-mpich actually does provide these compiler wrappers, they are located in `$CRAY_MPICH_PREFIX/bin`. That said, they are not used to compiler anything, although they can be, only to assist CMakes `FindMPI` in locating the include and library directories. In an ideal world, CMake should be able to find these automagically, and when you configure a cmake project outside of spack it works just fine without the hints, but for some reason in the spack environment this additional hint is needed. I have tried digging a couple times to ascertain the reason why but I couldn't find anything in particular that would cause the issue.

>  let us limit it down, e.g., to platform cray.

The reason I didn't do this is because I have seen this kind of issue on non-cray systems, I have an AMD test bed that I work on for development that occasionally demonstrates this behavior. By putting this in general I figure it at very least can future proof against configurations that are weird for some unknown reason.

I suspect this is related to why SENSEI needs to define `SENSEI_DIR` even though it is in the `CMAKE_PREFIX_PATH` environment variable, which really doesn't make any sense. Whenever I get some time, I am going to give this a proper dig because it would be awesome to clean a lot of this stuff up.

> You might be aware of our configurations on Cray systems

I haven't seen this before. I will look through it more closely, but at a glance it seems like it is more or less mimicking what spack is/should be doing behind the scenes.

tgamblin: Note that we would really like to move all Spack builds to use the `mpicc` wrappers with cray-mpich -- eventually `spec["mpi"].mpicc` will start returning those instead of `cc`, simply because there is so much garbage that `cc` puts on the compile line -- we can't control it effectively from within Spack.  This is part of moving away from modules and towards a build model for cray that looks much like every other platform.

So if you want to start using those *instead* of `cc` it would actually be great.  I would be interested to know if you run into issues.

ax3l (dismissed):
Looks like a great solution, thanks a lot for clarifying.

ax3l:
CI failed, let me restart.

@spackbot re-run pipeline

alalazo (changes requested):
This is completely broken I think. `with when` should be used with directives, not within functions where instead:
```python
self.spec.satisfies(constraint)
```
ought to be used.

kwryankrattiger:
rebased and converted the `with when()` -> `if in spec`

alalazo (approved):

Issue #31401, opened 2022-06-30
binary packages: symbolic links not rewritten to account for new locations


eugeneswalker:
### Steps to reproduce

Symbolic links are not being re-written to account for new locations when installing packages from build cache.

This is shown below when w/ `openssl` installed from build cache into a path that differs from the path where `openssl` was originally built.

Reproduced using:
* Docker container image `ecpe4s/ubuntu20.04-runner-x86_64:2022-05-01`
* `spack/develop` (861edcf996a1ce001aa20426d290ef950fd79870 from `Thu Jun 30 16:38:58 2022 -0600`)

```
# 1. Clone spack
$> git clone https://github.com/spack/spack /spack
$> (cd /spack && git checkout 861edcf996a1ce001aa20426d290ef950fd79870)
$> . /spack/share/spack/setup-env.sh

# 2. Build openssl in one path, push to cache
$> spack config add "config:install_tree:padded_length:256"
$> spack install --fresh openssl
$> spack buildcache create -af --unsigned -d / openssl

# 3. Uninstall openssl
$> spack uninstall --all

# 4. Install openssl from the cache, to a path that differs from the first path
$> spack config rm "config:install_tree:padded_length:256"
$> spack mirror add local /
$> spack install --fresh --cache-only --no-check-signature openssl
...
==> Installing openssl-1.1.1p-6i5af6cusk73iywespdfifrbqdbzq4br
==> Fetching file:///build_cache/linux-ubuntu20.04-cascadelake-gcc-11.1.0-openssl-1.1.1p-6i5af6cusk73iywespdfifrbqdbzq4br.spec.json
==> Fetching file:///build_cache/linux-ubuntu20.04-cascadelake/gcc-11.1.0/openssl-1.1.1p/linux-ubuntu20.04-cascadelake-gcc-11.1.0-openssl-1.1.1p-6i5af6cusk73iywespdfifrbqdbzq4br.spack
==> Extracting openssl-1.1.1p-6i5af6cusk73iywespdfifrbqdbzq4br from binary cache
[...]

PR #31843, opened 2022-08-01
e4s ci: packages: prefer openssl certs=system due to binary relocation issue #31401

eugeneswalker:
Prefer `openssl certs=system` since `openssl certs=mozilla` cannot be properly relocated due to https://github.com/spack/spack/issues/31401

FYI @wspear @scottwittenburg


haampie:
alternatively copy the pem file

eugeneswalker:
> alternatively copy the pem file

instead of symlinking to the mozilla certs install path, just copy the certificates so they become part of the openssl install? sounds good to me!

PR #31852, opened 2022-08-01
Help with finding hdf5/blt; set host compiler properly for +mpi+cuda

scheibelp:
* My build wanted to set the C++ standard. This PR sets `c++14` universally, but should that be conditioned to only apply for a specific version range?
* When building `dray+cuda+mpi` I had to set CUDAHOSTCXX: I was not able to set this using familiar CMake variables like `CUDA_HOST_COMPILER`
* `dray` was locating the wrong `blt`, so I pointed it explicitly to the Spack-built one
* `dray` was having trouble locating `hdf5` for my build, so I explicitly pointed it at the Spack-built one

### DIFF
--- a/var/spack/repos/builtin/packages/dray/package.py
+++ b/var/spack/repos/builtin/packages/dray/package.py
@@ -107,8 +107,8 @@ class Dray(Package, CudaPackage):
         env.set("CTEST_OUTPUT_ON_FAILURE", "1")
 
         spec = self.spec
-        if '+cuda' in spec and '+mpi' in spec:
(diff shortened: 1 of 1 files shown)


cyrush (dismissed):

cyrush:
HDF5 info should be fully imported from Conduit, so it's strange you had to add.

Also, FYI: Devil Ray has been moved into ascent, so the separate spack package won't be needed for future releases.

scheibelp:
@spackbot run pipeline

PR #31857, opened 2022-08-01
py-breathe: add new version and improve version constraints

sethrj:
Follow-on to some lovely discussions on https://github.com/spack/spack/pull/31828

### DIFF
--- a/var/spack/repos/builtin/packages/py-breathe/package.py
+++ b/var/spack/repos/builtin/packages/py-breathe/package.py
@@ -33,6 +33,8 @@ class PyBreathe(PythonPackage):
     depends_on("python@3.6:", type=("build", "run"), when="@4.33:")
     depends_on("py-docutils@0.5:", type=("build", "run"), when="@:4.20")
     depends_on("py-docutils@0.12:", type=("build", "run"), when="@4.21:")
+    # Note: Pygments is missing from the setup.py in 4.34.0 but is listed in
+    # the requirements file and used by breathe.filetypes.
     depends_on("py-pygments@1.6:", type=("build", "run"), when="@4.21:")
     depends_on("py-six@1.4:", type=("build", "run"), when="@:4.20")
     depends_on("py-six@1.9:", type=("build", "run"), when="@4.21:4.32")
--- a/var/spack/repos/builtin/packages/py-breathe/package.py
+++ b/var/spack/repos/builtin/packages/py-breathe/package.py
@@ -39,7 +39,7 @@ class PyBreathe(PythonPackage):
     depends_on("py-sphinx", type=("build", "run"))
     depends_on("py-sphinx@:3", type=("build", "run"), when="@:4.20")
     depends_on("py-sphinx@3:3.2", type=("build", "run"), when="@4.21:4.32")
-    depends_on("py-sphinx@3:4", type=("build", "run"), when="@4.33.0:4.33")
-    depends_on("py-sphinx@4.0:4,5.0.1:5", type=("build", "run"), when="@4.34:")
+    depends_on("py-sphinx@3:4", type=("build", "run"), when="@4.33")
+    depends_on("py-sphinx@4,5.0.1:5", type=("build", "run"), when="@4.34:")
 
     depends_on("doxygen@1.8.4:", type="run")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-breathe/package.py
```diff
@@ -38,6 +39,7 @@ class PyBreathe(PythonPackage):
     depends_on("py-sphinx", type=("build", "run"))
     depends_on("py-sphinx@:3", type=("build", "run"), when="@:4.20")
     depends_on("py-sphinx@3:3.2", type=("build", "run"), when="@4.21:4.32")
-    depends_on("py-sphinx@3:", type=("build", "run"), when="@4.33:")
+    depends_on("py-sphinx@3:4", type=("build", "run"), when="@4.33:")
```
adamjstewart:
```suggestion
    depends_on("py-sphinx@3:4", type=("build", "run"), when="@4.33")
```
Otherwise this constraint will also be applied to 4.34.0

var/spack/repos/builtin/packages/py-breathe/package.py
```diff
@@ -38,6 +39,7 @@ class PyBreathe(PythonPackage):
...
+    depends_on("py-sphinx@4.0:4,5.0.1:", type=("build", "run"), when="@4.34:")
```
adamjstewart:
```suggestion
    depends_on("py-sphinx@4.0:4,5.0.1:5", type=("build", "run"), when="@4.34:")
```
Sphinx 6 is not yet supported.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-breathe/package.py
sethrj: Good point... probably
```
depends_on("py-sphinx@3:4", type=("build", "run"), when="@4.33.0:4.33")
```
because 4.33.1 is not in `4.33` (oh spack versions, how I love you so)

adamjstewart: Nah, `@4.33` works. At least it works with clingo, don't care about the old concretizer.

sethrj: what seriously? I was convinced 4.33 was a specific version and `4.33.0:4.33` was a ranged version...

adamjstewart: It used to be in the old concretizer. I think it still is on the command line.

sethrj: I think the command line has scarred me for life :(

var/spack/repos/builtin/packages/py-breathe/package.py
```diff
@@ -38,6 +39,7 @@ class PyBreathe(PythonPackage):
     depends_on("py-sphinx", type=("build", "run"))
```
adamjstewart, line 41: pygments is no longer required

var/spack/repos/builtin/packages/py-breathe/package.py
```diff
@@ -38,6 +39,7 @@ class PyBreathe(PythonPackage):
...
     depends_on("py-sphinx@:3", type=("build", "run"), when="@:4.20")
     depends_on("py-sphinx@3:3.2", type=("build", "run"), when="@4.21:4.32")
-    depends_on("py-sphinx@3:", type=("build", "run"), when="@4.33:")
+    depends_on("py-sphinx@3:4", type=("build", "run"), when="@4.33.0:4.33")
+    depends_on("py-sphinx@4.0:4,5.0.1:5", type=("build", "run"), when="@4.34:")
```
adamjstewart:
```suggestion
    depends_on("py-sphinx@3:4", type=("build", "run"), when="@4.33")
    depends_on("py-sphinx@4,5.0.1:5", type=("build", "run"), when="@4.34:")
```

var/spack/repos/builtin/packages/py-breathe/package.py
sethrj: Maybe they just forgot to leave it in the setup.py file? The `breathe.filetypes` definitely has `import pygments`, and `Pygments>=1.6` is listed in the `requirements/production.txt` which is loaded by `pip` in the breathe CI scripts.

adamjstewart: Can you add a comment explaining that the dep is missing?

sethrj:
@adamjstewart I think I've finally addressed all the feedback, if you have a second to review. Thanks!

adamjstewart (approved):

sethrj:
Thanks!

PR #31862, opened 2022-08-02
openssl: copy mozilla certs to work around symlink relocation issue

haampie:
closes #31843

Alternative to #31843, make spack copy the cacerts.pem file from ca-certificates-mozilla instead of symlinking it.

This is necessary because relocation logic does not handle symlinks across spack packages.


alalazo (approved):

eugeneswalker:
Thank you @haampie for this

PR #31864, opened 2022-08-02
document common packages.yaml not respected due to --reuse issue

haampie:
Since this comes up twice a day... resurrect Known issues in the docs.

And also, this should probably enter the 0.18 docs, since I don't think we can backport any breaking changes w.r.t. packages.yaml behavior.

https://spack--31864.org.readthedocs.build/en/31864/known_issues.html

### DIFF
--- a/lib/spack/docs/known_issues.rst
+++ b/lib/spack/docs/known_issues.rst
@@ -24,10 +24,6 @@ packages, the concretizer values reuse of installed packages higher than prefere
 set in ``packages.yaml``. Note that ``packages.yaml`` specifies only preferences, not
 hard constraints.
 
-To verify you are affected by this problem, you can use
-``spack spec --install-status <spec>``. This should indicate with a ``[+]`` that
-certain "problematic" specs were picked because they were already installed.
-
 There are multiple workarounds:
 
 1. Disable reuse during concretization: ``spack install --fresh <spec>`` when installing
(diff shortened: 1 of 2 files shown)


lib/spack/docs/known_issues.rst
```diff
@@ -0,0 +1,38 @@
...
+micro-architectures.
+
+This is however not a bug. In order to reduce the number of builds of the same
+packages, the concretizer values reuse of installed packages higher than preferences
+set in ``packages.yaml``. Note that ``packages.yaml`` specifies only preferences, not
+hard constraints.
+
+To verify you are affected by this problem, you can use
+``spack spec --install-status <spec>``. This should indicate with a ``[+]`` that
+certain "problematic" specs were picked because they were already installed.
```
alalazo: This might also happen with buildcaches, in which case the user won't see a `[+]` :thinking:

alalazo: We should add (in another PR, of course) a new symbol, for instance `[B]`, to indicate that a spec has been selected from a buidcache.

haampie: I'll just remove that line

haampie:
```suggestion
```

alalazo (approved):
We'll probably have to add a line when requirements are added to `packages.yaml`

Issue #22866, opened 2021-04-08
Spack blame output incorrect for blank keys


sethrj:
As I mentioned in https://github.com/spack/spack/pull/22598#issuecomment-808896029 , the "empty" keys in the default modules.yaml file aren't rendered correctly.

### Steps to reproduce the issue

```console
$ spack config blame modules
# ...snip...
etc/spack/defaults/modules.yaml:32             share/pkgconfig:
etc/spack/defaults/modules.yaml:33             - PKG_CONFIG_PATH
                                               ? ''
                                               : - CMAKE_PREFIX_PATH
```

Prior to #22598 the blank lines had the correct line numbers corresponding to the values in `defaults/modules.yaml`:
```
34:    '':
35:      - CMAKE_PREFIX_PATH
```

Expected output:

```console
$ spack config blame modules
# ...snip...
etc/spack/defaults/modules.yaml:32             share/pkgconfig:
etc/spack/defaults/modules.yaml:33             - PKG_CONFIG_PATH
etc/spack/defaults/modules.yaml:34             '':
etc/spack/defaults/modules.yaml:35             - CMAKE_PREFIX_PATH
```
### Information on your system

* **Spack:** 0.16.1-2140-19b6d3589a
* **Python:** 3.8.5
* **Platform:** darwin-bigsur-skylake
* **Concretizer:** original

alalazo:
Bug still present on

* **Spack:** 0.17.1-735-f8d4b4746c
* **Python:** 3.9.7
* **Platform:** linux-ubuntu20.04-icelake
* **Concretizer:** clingo

sethrj:
@tgamblin @becker33 You may want to fix this before 0.18 is released, because with the warning message
```
==> Warning: Top-level properties "enable", "prefix_inspections", "lmod" in module config are ignored as of Spack v0.18. They should be set on the "default" module set. Run

	$ spack config update modules

to update the file to the new format
```

the modules file
```yaml
modules:
  enable::
    - lmod
  prefix_inspections:
    '':
    - CMAKE_PREFIX_PATH
    - '${PACKAGE}_ROOT'
```
creates a ~~broken~~ very odd `modules.yaml` file (though thankfully it leaves the backup):
```yaml
modules:
  prefix_inspections:
    ? ''
    : - CMAKE_PREFIX_PATH
      - ${PACKAGE}_ROOT
```

[...]

PR #31867, opened 2022-08-02
Make prefix_inspections start with `./`

haampie:
Not a proper fix to #22866, but at least an improvement...

1. This is a way to get rid of the weird `spack config blame modules` output involving a question mark (due to empty string as a key, which imho is a bit questionable anyways ~so in that sense a question mark _does_ make sense~):
   ```yaml
   $ spack config get modules
   # before
   modules:
     prefix_inspections:
       share/pkgconfig:
       - PKG_CONFIG_PATH
       ? ''
       : - CMAKE_PREFIX_PATH
   
   # after
   modules:
     prefix_inspections:
       /share/pkgconfig:
       - PKG_CONFIG_PATH
       /:
       - CMAKE_PREFIX_PATH
   ```
3. Might actually me more clear to users that this is a mapping from path -> environment variables, especially `/` vs `''`.

### DIFF
--- a/etc/spack/defaults/darwin/modules.yaml
+++ b/etc/spack/defaults/darwin/modules.yaml
@@ -15,7 +15,7 @@
 # -------------------------------------------------------------------------
 modules:
   prefix_inspections:
-    lib:
+    ./lib:
       - DYLD_FALLBACK_LIBRARY_PATH
-    lib64:
+    ./lib64:
       - DYLD_FALLBACK_LIBRARY_PATH
--- a/etc/spack/defaults/modules.yaml
+++ b/etc/spack/defaults/modules.yaml
@@ -14,23 +14,24 @@
 #   ~/.spack/modules.yaml
 # -------------------------------------------------------------------------
 modules:
-  # Paths to check when creating modules for all module sets
+  # This maps paths in the package install prefix to environment variables
+  # they should be added to. For example, <prefix>/bin should be in PATH.
   prefix_inspections:
-    bin:
+    ./bin:
       - PATH
-    man:
+    ./man:
       - MANPATH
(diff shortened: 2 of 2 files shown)


haampie:
@scheibelp a test fails on windows, which makes me wonder: (a) is there any use for `modules.yaml` on Windows at all, apart from the fact that `modules:prefix_inspections` happens to be used in the build environment. And (b) we should probably remove `modules:prefix_inspections` from the build environment, since it's configurable but does not influence the hash (so, not particularly contributing to reproducibility...).

tgamblin:
Can we just fix this by getting ruamel not to treat `''` as a complex mapping?  It looks like upstream ruamel doesn't: https://github.com/yaml/pyyaml/issues/401.

I wonder if we are doing something in `spack_yaml.py` (e.g. we use a `str` subclass so that we can add file/line metadata) that causes ruamel to think it's complex when it shouldn't.

haampie:
Greg mentioned we can do `./` too, which sounds good to me

becker33 (approved):

matz-e:
I think this change broke our `git` environment module (as in the TCL ones) on Spack 0.19, which has
```
prepend_path PATH /path/to/git/installation/./bin
```

Maybe it's my combination of `zsh` and `environment-modules` 3 on our cluster, but normalizing the relative paths before joining them with the `spec.prefix` helped, and removed the `./` indirection.

haampie:
Sounds OK to run normpath on the user input `./bin -> bin`, as long as you don't run it after joining. Can you make a PR?

matz-e:
Done (#35318), wasn't entirely sure if this was accounted for somewhere else, and/or intended.

PR #31869, opened 2022-08-02
CUDA support to quantum-espresso package

bellenlau:
Hello,

this is the extension of the quantum-espresso package to support CUDA.

Thank you,

Laura

### DIFF
--- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
+++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
@@ -74,7 +74,7 @@ class QuantumEspresso(CMakePackage, CudaPackage):
         conflicts("^openblas threads=none")
         conflicts("^openblas threads=pthreads")
 
-    # Maybe the following is not needed, check CMakeLists.txt
+    # Add cuda support 
     with when('+cuda'):
         # only nvhpcsdk compiler is supported
         conflicts("%aocc")
@@ -109,7 +109,6 @@ class QuantumEspresso(CMakePackage, CudaPackage):
     variant("nvtx", default=False, description="Enables NVTX markers for profiling")
     with when("+nvtx~cuda"):
         depends_on("cuda")
-
 
     # Apply upstream patches by default. Variant useful for 3rd party
     # patches which are incompatible with upstream patches
--- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
+++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
@@ -92,6 +92,7 @@ class QuantumEspresso(CMakePackage, CudaPackage):
         conflicts("%nag")
         conflicts("%xl")
         conflicts("%xl_r")
+        conflicts('@:6.6')
         # cuda version >= 10.1
         conflicts("cuda@:10.0.130")
         # bugs with following nvhpcsdk versions
@@ -387,22 +388,14 @@ class QuantumEspresso(CMakePackage, CudaPackage):
             self.define_from_variant('QE_ENABLE_CUDA', 'cuda'),
             self.define_from_variant('QE_ENABLE_PROFILE_NVTX', 'nvtx'),
             self.define_from_variant('QE_ENABLE_MPI_GPU_AWARE', 'mpigpu'),
-            #self.define_from_variant('NVFORTRAN_CUDA_CC', 'cuda_arch')
         ]
 
         if '+cuda' in self.spec:
             cmake_args.append(self.define('QE_ENABLE_OPENACC', True))
-            cuda_arch = spec.variants['cuda_arch'].value
-            print(cuda_arch)
-            if cuda_arch[0] != 'none':
-                cmake_args.append('-DNVFORTRAN_CUDA_CC={0}'.format(cuda_arch))
 
         # QE prefers taking MPI compiler wrappers as CMake compilers.
         if '+mpi' in spec:
             cmake_args.append(self.define('CMAKE_C_COMPILER', spec['mpi'].mpicc))
-            # Define pgi wrapper
-            #if '^spectrum-mpi' in spec:
-            #   spec['mpi'].mpifc = 'mpipgifort'
             cmake_args.append(self.define('CMAKE_Fortran_COMPILER', spec['mpi'].mpifc))
 
         if not spec.satisfies("hdf5=none"):
--- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
+++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
@@ -68,10 +68,10 @@ class QuantumEspresso(CMakePackage):
     # Need OpenMP threaded FFTW and BLAS libraries when configured
     # with OpenMP support
     with when("+openmp"):
-        conflicts("^fftw~openmp")
-        conflicts("^amdfftw~openmp")
-        conflicts("^openblas threads=none")
-        conflicts("^openblas threads=pthreads")
+        depends_on("fftw+openmp", when="^fftw")
+        depends_on("amdfftw+openmp", when="^amdfftw")
+        depends_on("openblas threads=openmp", when="^openblas")
+        depends_on("amdblis threads=openmp", when="^amdblis")
 
     # Add cuda support
     with when("%nvhpc"):
--- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
+++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
@@ -379,7 +379,7 @@ class QuantumEspresso(CMakePackage):
     @run_before("cmake")
     def caveats(self):
         if "+cuda~openmp" in self.spec:
-            tty.warn("CUDA without OpenMP could incur in severe performance degredation !!!")
+            tty.warn("CUDA without OpenMP could incur in severe performance degradation !!!")
 
     def cmake_args(self):
         spec = self.spec
--- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
+++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
@@ -7,7 +7,7 @@
 from spack.package import *
 
 
-class QuantumEspresso(CMakePackage):
+class QuantumEspresso(CMakePackage, CudaPackage):
     """Quantum ESPRESSO is an integrated suite of Open-Source computer codes
     for electronic-structure calculations and materials modeling at the
     nanoscale. It is based on density-functional theory, plane waves, and
@@ -70,7 +70,7 @@ class QuantumEspresso(CMakePackage):
     variant("libxc", default=False, description="Uses libxc")
     depends_on("libxc@5.1.2:", when="+libxc")
 
-    variant("openmp", default=False, description="Enables openMP support")
+    variant('openmp', default=False, description='Enables openMP support, auto-enabled by CMakeLists with +cuda')
     # Need OpenMP threaded FFTW and BLAS libraries when configured
     # with OpenMP support
     with when("+openmp"):
@@ -79,16 +79,46 @@ class QuantumEspresso(CMakePackage):
         conflicts("^openblas threads=none")
         conflicts("^openblas threads=pthreads")
 
+    # NVTX variant for profiling ; requires linking to CUDA library, handled by CMake
+    variant('nvtx', default=False, description='Enables NVTX markers for profiling')
+    # Maybe the following is not needed, check CMakeLists.txt
+    with when('+cuda'):
+        conflicts("%arm")
+        conflicts("%cce")
+        conflicts("%apple-clang")
+        conflicts("%clang")
+        conflicts("%gcc")
+        conflicts("%intel")
+        conflicts("%nag")
+        conflicts("%xl")
+        conflicts("%xl_r")
+        # cuda version >= 10.1
+        conflicts("cuda@:10.0.130")
+        # bugs with following nvhpcsdk versions
+        conflicts('%nvhpc@21.11:22.2', msg='Issues this with NVHPC compiler version')
+        # bugs with nvhpc@22.3 in versions > 7.1 still present
+        with when('@develop'):
+           conflicts('%nvhpc@22.3', msg='Issues this with NVHPC compiler version')
+        with when('@7.2:'):
+           conflicts('%nvhpc@22.3', msg='Issues this with NVHPC compiler version')
+        # only cmake is supported 
+        conflicts('~cmake', msg='Only CMake supported for GPU version')
+        # PGI support
+        conflicts('%pgi@:18.10', msg='PGI supported from v 19.10')
+
     # Apply upstream patches by default. Variant useful for 3rd party
     # patches which are incompatible with upstream patches
     desc = "Apply recommended upstream patches. May need to be set "
     desc = desc + "to False for third party patches or plugins"
     variant("patch", default=True, description=desc)
 
-    variant("mpi", default=True, description="Builds with mpi support")
-    with when("+mpi"):
-        depends_on("mpi")
-        variant("scalapack", default=True, description="Enables scalapack support")
+    variant('mpi', default=True, description='Builds with mpi support')
+    with when('+mpi'):
+        depends_on('mpi')
+        variant('scalapack', default=True, description='Enables scalapack support')
+        with when('+cuda'):
+            # add mpi_gpu_aware variant, False by default
+            variant('mpigpu', default=False, description='Enables __GPU_MPI')
 
     with when("+scalapack"):
         depends_on("scalapack")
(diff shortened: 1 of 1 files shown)


danielecesarini:
@ye-luo this PR is a work of Laura and mine, please revise this PR for us.

@bellenlau is a colleague of mine in CINECA and she works on QE. She will be in charge also of CMake build system for QE (not only of spack package). I will help Laura to address future issues on QE but she will be the primary contact for QE issues on SPACK package and CMake build system.

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -7,7 +7,7 @@
 from spack.package import *
 
 
-class QuantumEspresso(CMakePackage):
+class QuantumEspresso(CMakePackage, CudaPackage):
```
ye-luo: I don't see this necessary. QE doesn't use CUDA toolkit but relies on NVHPC.

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -79,6 +74,42 @@ class QuantumEspresso(CMakePackage):
...
+        conflicts("%arm")
+        conflicts("%cce")
+        conflicts("%clang")
+        conflicts("%fj")
+        conflicts("%gcc")
+        conflicts("%intel")
+        conflicts("%nag")
+        conflicts("%oneapi")
+        conflicts("%xl")
+        conflicts("%xl_r")
```
ye-luo: Can we enable `+cuda` variant only when NVHPC is used instead of having this long list of conflicts.

ye-luo (changes requested):
@bellenlau first welcome.
See my inline comments.

var/spack/repos/builtin/packages/quantum-espresso/package.py
bellenlau: CudaPackage has been removed in the last commit b846a23f487af9289e4274e530a7b4715edbbe05

var/spack/repos/builtin/packages/quantum-espresso/package.py
bellenlau: +cuda variant is now enabled only when NVHPC is used b846a23f487af9289e4274e530a7b4715edbbe05

bellenlau:
Hello @alalazo, we noticed that the warning message in f9d464d652964cfd73bcaee33bf0b093e38c9e70
```
# Warning for CUDA without OpenMP
@run_before("cmake")
def caveats(self):
    if "+cuda~openmp" in self.spec:
        tty.warn("CUDA without OpenMP could incur in severe performance degradation !!!")
```
does not pop out before _cmake_ stage, despite variants match +cuda~openmp.  
Any suggestion on this issue?

danielecesarini:
> Hello @alalazo, we noticed that the warning message in [f9d464d](https://github.com/spack/spack/commit/f9d464d652964cfd73bcaee33bf0b093e38c9e70)
> 
> ```
> # Warning for CUDA without OpenMP
> @run_before("cmake")
> def caveats(self):
>     if "+cuda~openmp" in self.spec:
>         tty.warn("CUDA without OpenMP could incur in severe performance degradation !!!")
> ```
> 
> does not pop out before _cmake_ stage, despite variants match +cuda~openmp. Any suggestion on this issue?

Actually, the tty.warn() prints on the standard output, which is redirected on the build file of SPACK at this time. We would wonder if there is some hook to print messages to warn the user about the configuration choice (that could not be the optimal one). We would do even in the concretization phase not only in the build phase.

ye-luo:
Right now the blocking issue is that we cannot set openmp variant default value depending on cuda on/off. Implementing a warning is cumbersome and almost invisible. So my suggestions is setting openmp on by default. Users should not stuck in MPI-only land and if they really need MPI-only, OMP_NUM_THREADS=1 can be set. I generally think enable OpenMP by default is a positive thing.

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -79,6 +75,37 @@ class QuantumEspresso(CMakePackage):
...
+            )
+            # bugs with nvhpc@22.3 in versions > 7.1 still present
+            with when("@develop"):
+                conflicts("%nvhpc@22.3", msg="Develop has issues with this NVHPC compiler version")
+            with when("@7.2:"):
+                conflicts("%nvhpc@22.3", msg="v > 7.1 has issues with this NVHPC compiler version")
+            # only cmake is supported
+            conflicts("~cmake", msg="Only CMake supported for GPU-enabled version")
+            # PGI support
+            conflicts("%pgi@:19.9", msg="PGI supported for v >= 19.10")
```
ye-luo: If nvhpc is required, isn't this conflict always false?

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -79,6 +75,37 @@ class QuantumEspresso(CMakePackage):
...
+            # conflicts("cuda@:10.0.130")
+            # bugs with following nvhpcsdk versions
+            conflicts(
+                "%nvhpc@21.11:22.2", msg="QE-GPU has issues with these NVHPC compiler versions"
+            )
+            # bugs with nvhpc@22.3 in versions > 7.1 still present
+            with when("@develop"):
+                conflicts("%nvhpc@22.3", msg="Develop has issues with this NVHPC compiler version")
+            with when("@7.2:"):
+                conflicts("%nvhpc@22.3", msg="v > 7.1 has issues with this NVHPC compiler version")
```
ye-luo: I think 22.3 works fine after 7.1 release based on https://gitlab.com/QEF/q-e/-/issues/479#note_927308910

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -79,6 +75,37 @@ class QuantumEspresso(CMakePackage):
...
+        )
+        with when("+cuda"):
+            # GPUs are enabled since v6.6
+            conflicts("@:6.5")
+            # cuda version >= 10.1
+            # conflicts("cuda@:10.0.130")
+            # bugs with following nvhpcsdk versions
+            conflicts(
+                "%nvhpc@21.11:22.2", msg="QE-GPU has issues with these NVHPC compiler versions"
+            )
```
ye-luo: Should be conflict with `+openmp` only

var/spack/repos/builtin/packages/quantum-espresso/package.py
bellenlau: I tested the develop at commit 77c166c6ebb13ea0bf13d846bcd6c812e2584825 with NVHPC/22.3  CUDA/11.5 OpenMPI/4.1.2 on juwels booster and it does not converge with Davidson diagonalization

var/spack/repos/builtin/packages/quantum-espresso/package.py
bellenlau: You are right. To be consistent with CMakeLists.txt we should enable also the PGI compiler, I suggest.

ye-luo: No need to waste time on pgi releases.

var/spack/repos/builtin/packages/quantum-espresso/package.py
ye-luo: Got it. Probably we should just black list 22.3 completely.

ye-luo:
> Right now the blocking issue is that we cannot set openmp variant default value depending on cuda on/off. Implementing a warning is cumbersome and almost invisible. So my suggestions is setting openmp on by default. Users should not stuck in MPI-only land and if they really need MPI-only, OMP_NUM_THREADS=1 can be set. I generally think enable OpenMP by default is a positive thing.

@danielecesarini @bellenlau what do you think?
Let us finalize this PR.

bellenlau:
> > Right now the blocking issue is that we cannot set openmp variant default value depending on cuda on/off. Implementing a warning is cumbersome and almost invisible. So my suggestions is setting openmp on by default. Users should not stuck in MPI-only land and if they really need MPI-only, OMP_NUM_THREADS=1 can be set. I generally think enable OpenMP by default is a positive thing.
> 
> @danielecesarini @bellenlau what do you think? Let us finalize this PR.

OpenMP as a default always is fine - even if it is not the same when using cmake only . I will push asap.

var/spack/repos/builtin/packages/quantum-espresso/package.py
bellenlau: Added in 487d76f99948010ddf0daf8cd1eaf35a2dd9929e.

var/spack/repos/builtin/packages/quantum-espresso/package.py
bellenlau: Done in 487d76f99948010ddf0daf8cd1eaf35a2dd9929e.

var/spack/repos/builtin/packages/quantum-espresso/package.py
bellenlau: Removed in 487d76f99948010ddf0daf8cd1eaf35a2dd9929e.

bellenlau:
@amd-toolchain-support apparently the pipeline to build QuantumEspresso fails when configuring amdfftw with OpenMP (https://gitlab.spack.io/spack/spack/-/jobs/3003806). Could you provide further insight on this?

Here is how the configuration fails:

```
checking for dummy main to link with Fortran 77 libraries... unknown
configure: WARNING: *** Couldn't figure out how to link C and Fortran; using default Fortran wrappers.
checking for OpenMP flag of C compiler... unknown
configure: error: don't know how to enable OpenMP
```

alalazo:
@bellenlau Probably you already saw the message, but in case you missed it, you can reproduce the failure locally following these instructions:
```
To reproduce this build locally, run:
    spack ci reproduce-build https://gitlab.spack.io/api/v4/projects/2/jobs/3003806/artifacts [--working-dir <dir>]
If this project does not have public pipelines, you will need to first:
    export GITLAB_PRIVATE_TOKEN=<generated_token>
... then follow the printed instructions.
```
printed at the end of the logs.

danielecesarini:
> @bellenlau Probably you already saw the message, but in case you missed it, you can reproduce the failure locally following these instructions:
> 
> ```
> To reproduce this build locally, run:
>     spack ci reproduce-build https://gitlab.spack.io/api/v4/projects/2/jobs/3003806/artifacts [--working-dir <dir>]
> If this project does not have public pipelines, you will need to first:
>     export GITLAB_PRIVATE_TOKEN=<generated_token>
> ... then follow the printed instructions.
> ```
> 
> printed at the end of the logs.

@alalazo we think that this is a bug of Spack (or about the CI), because the problem is related to the AMDFFW library on a container that runs on ARM/Graviton2 instance (we don't depend on AMDFFTW but only to the virtual provider FFTW-API@3). But this is look very weird, could you check the pipeline?

alalazo:
@danielecesarini This is not a bug in Spack, nor in the CI. The changes you make here, and the overuse of conflicts, lead clingo to add penalties to certain providers. The result is that with the changes in this PR the "optimal" configuration is when `amdfftw` is in the DAG[^1]. 

For complicated packages, like `quantum-espresso`, my experience is that directives that "tell Spack what to do" work better than directives that tell Spack what to do not. Can you check if this diff:
```diff
diff --git a/var/spack/repos/builtin/packages/quantum-espresso/package.py b/var/spack/repos/builtin/packages/quantum-espresso/package.py
index ff8efb6345..1e1007c775 100644
--- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
+++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
@@ -68,10 +68,9 @@ class QuantumEspresso(CMakePackage):
     # Need OpenMP threaded FFTW and BLAS libraries when configured
     # with OpenMP support
     with when("+openmp"):
-        conflicts("^fftw~openmp")
-        conflicts("^amdfftw~openmp")
-        conflicts("^openblas threads=none")
-        conflicts("^openblas threads=pthreads")
+        depends_on("fftw+openmp", when="^fftw")
+        depends_on("amdfftw+openmp", when="^amdfftw")
+        depends_on("openblas threads=openmp", when="^openblas")
 
     # Add cuda support
     with when("%nvhpc"):

```
works for you? With that I was able to make `quantum-espresso` concretize by default using `fftw` and `openblas`
[^1]: obviously `amdfftw` doesn't build neatly, so maybe it's worth pinging @amd-toolchain-support and @cgmb about that

cgmb:
> 1. obviously `amdfftw` doesn't build neatly, so maybe it's worth pinging @amd-toolchain-support and @cgmb about that

The AMD CPU and GPU math library teams are entirely separate. I'm a part of the GPU group, so I'm afraid I don't have any contacts with the `amdfftw` developers. I'm an occasional user of the library, but I know basically nothing about autotools. AOCL is the domain of AMD Toolchain Support.

bellenlau:
@alalazo Thank you for the hint! Changes are included in commit 4cb6bc4ee2fc64ce717675f843b88f1228e2bc5b. 

> For complicated packages, like `quantum-espresso`, my experience is that directives that "tell Spack what to do" work better than directives that tell Spack what to do not. Can you check if this diff:
> 
> ```diff
> diff --git a/var/spack/repos/builtin/packages/quantum-espresso/package.py b/var/spack/repos/builtin/packages/quantum-espresso/package.py
> index ff8efb6345..1e1007c775 100644
> --- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
> +++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
> @@ -68,10 +68,9 @@ class QuantumEspresso(CMakePackage):
>      # Need OpenMP threaded FFTW and BLAS libraries when configured
>      # with OpenMP support
>      with when("+openmp"):
> -        conflicts("^fftw~openmp")
> -        conflicts("^amdfftw~openmp")
> -        conflicts("^openblas threads=none")
> -        conflicts("^openblas threads=pthreads")
> +        depends_on("fftw+openmp", when="^fftw")
> +        depends_on("amdfftw+openmp", when="^amdfftw")
> +        depends_on("openblas threads=openmp", when="^openblas")
>  
>      # Add cuda support
>      with when("%nvhpc"):
> ```
> 
> works for you? With that I was able to make `quantum-espresso` concretize by default using `fftw` and `openblas`
> 

We further added  `depends_on("amdblis threads=openmp", when="^amdblis")` to improve the default concretization of amd blas.

danielecesarini:
@ye-luo I think that we are ready to finalize this PR.

ye-luo:
> @ye-luo I think that we are ready to finalize this PR.

I will take a look on Friday. @bellenlau could you rebase all the commits on top of the develop?

alalazo (dismissed):
LGTM. I'll merge as soon as @ye-luo approves too

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -61,23 +60,44 @@ class QuantumEspresso(CMakePackage):
...
-        conflicts("^fftw~openmp")
-        conflicts("^amdfftw~openmp")
-        conflicts("^openblas threads=none")
-        conflicts("^openblas threads=pthreads")
+        depends_on("fftw+openmp", when="^fftw")
+        depends_on("amdfftw+openmp", when="^amdfftw")
+        depends_on("openblas threads=openmp", when="^openblas")
+        depends_on("amdblis threads=openmp", when="^amdblis")
+
+    # Add cuda support
```
ye-luo: Let us mention "CUDA Fortran" instead of CUDA. Use upper case.
Also add a note that it depends on NVHPC compiler but not directly on CUDA toolkit.

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -61,23 +60,44 @@ class QuantumEspresso(CMakePackage):
...
+    with when("%nvhpc"):
+        variant(
+            "cuda",
+            default=False,
+            description="Build with CUDA",
```
ye-luo: Let us mention CUDA Fortran instead of CUDA.

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -61,23 +60,44 @@ class QuantumEspresso(CMakePackage):
...
+            # bugs with following nvhpcsdk versions and +openmp
+            with when("+openmp"):
+                conflicts(
+                    "%nvhpc@21.11:22.3",
+                    msg="bugs with NVHPCSDK from v21.11 to v22.3, OpenMP and GPU",
+                )
+            # only cmake is supported
+            conflicts("~cmake", msg="Only CMake supported for GPU-enabled version")
+
+    # NVTX variant for profiling ; requires linking to CUDA library, handled by CMake
```
ye-luo: `CUDA library` which library? Better to be specific. I guess CUDA runtime APIs?

ye-luo (commented):
Only need some updates in comments to keep our up-to-date knowledge close to the source before accepting this PR.

Later we could explore having an option nvidia_gpu_arch to restrict compilation to specific GPU architectures. Current default in NVHPC compiles for every known GPU architectures and that is slow.

bellenlau:
@ye-luo comments have been updated in ff4d4c672fdf47bab3e3547a302157a3c103912a

ye-luo (approved):
LGTM. @alalazo please merge

alalazo (approved):

alalazo:
Thanks!

danielecesarini (approved):

bellenlau:
@alalazo  Apparently, merging is blocked because of a lack of permit. Do you have any hints on this issue?

alalazo:
@bellenlau We are reworking CI on GitHub, and had issues running pipelines for a couple of days. That part is solved now.

PR #31808, opened 2022-07-29
ci: Expand the DAV SDK stack to include both osmesa and glx

chuckatkins:

PR #31870, opened 2022-08-02
libedit: Make sure libedit linkes to the correct terminfo lib from spack

chuckatkins:
This makes sure that libedit links to the terminfo provided by spack and not one it finds elsewhere on the system.  Fixes the llvm build failure #31808


kwryankrattiger (approved):
Configurations with/without termlib configure and build

PR #31872, opened 2022-08-02
black: fix bootstrap for Python 3.6

greenc-FNAL:
- py-platformdirs: sanitize UTF-8 setup.cfg on Python 3.6
- py-platformdirs: black compliance

When Spack is using Python 3.6 (_e.g._ on RHEL7-based platforms such as
scientific7), the package `py-platformdirs` as required by black causes
the bootstrapping process to fail when the command:

```console
spack style -t black
```

is executed for the first time.

The `setup.cfg` configuration file for `py-platformdirs` contains the
line:

```
maintainer = Bernát Gábor, Julian Berman, Ofek Lev, Ronny Pfannschmidt
```

which is not handled correctly by Python 3.6. This PR replaces the
non-ASCII characters with Python escape sequences via the
`errors="backslashreplace"` feature of Python's builtin
`open()` function.

### DIFF
--- a/var/spack/repos/builtin/packages/py-platformdirs/package.py
+++ b/var/spack/repos/builtin/packages/py-platformdirs/package.py
@@ -23,12 +23,9 @@ class PyPlatformdirs(PythonPackage):
     depends_on("py-setuptools-scm@5:+toml", type="build")
 
     @when("^python@:3.6")
-    @run_before("install")
-    def handle_utf8_setup_cfg(self):
-        """Sanitize setup.cfg to resolve issues with Python 3.6 and UTF-8"""
-        with open("setup.cfg", mode="r", encoding="utf-8") as input_file:
-            cfg = input_file.readlines()
-        with open(
-            "setup.cfg", mode="w", encoding="ascii", errors="backslashreplace"
-        ) as output_file:
-            output_file.writelines(cfg)
+    def setup_build_environment(self, env):
+        """Python 3.6 needs a little help with setup.cfg re:
+
+        https://github.com/platformdirs/platformdirs/issues/50
+        """
+        env.set("LC_ALL", "en_US.UTF-8")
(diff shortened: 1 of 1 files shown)


adamjstewart:
Is this the issue you were seeing? https://github.com/platformdirs/platformdirs/issues/50

Maybe changing `LANG` would be a quicker workaround.

greenc-FNAL:
> Is this the issue you were seeing? [platformdirs/platformdirs#50](https://github.com/platformdirs/platformdirs/issues/50)
> 
> Maybe changing `LANG` would be a quicker workaround.

I'm not sure where `LANG` is being set if that's the cause. Output of `locale`:

```console
$ locale
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE=POSIX
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```

adamjstewart:
We set `LC_ALL` [here](https://github.com/spack/spack/blob/v0.18.1/lib/spack/spack/build_environment.py#L238,L243) based on the setting [here](https://github.com/spack/spack/blob/v0.18.1/etc/spack/defaults/config.yaml#L129). You can try overriding this in `setup_build_environment` for just that package. We could change the default, but that won't work if someone overrides that setting.

greenc-FNAL:
The following works:

```python
    @when("^python@:3.6")
        if not os.environ.get("LC_ALL", "").endswith(r"UTF-8"):
            env.set("LC_ALL", "en_US.UTF-8")
```

Shall I change the PR, or are you planning to change the `config.yaml` default?

adamjstewart:
I would remove the if-statement and add a comment with a link to https://github.com/platformdirs/platformdirs/issues/50, but yes, I think that's the best solution for this PR.

adamjstewart (approved):

PR #31880, opened 2022-08-02
ExaGO: Disable Python with rocm+ipopt. HiOp: fix umpire+raja costraints. RAJA: fix option naming for 0.14.0

cameronrutherford:


cameronrutherford:
@pelesh perhaps we should get some other reviewers for ExaSGD spack packages. I needed these changes in order to build and configure ExaGO on crusher with HiOp develop. I have opened a PR for HiOp and ExaGO with changes as well.

cameronrutherford:
I think these changes should be merged, but perhaps there is a way to compile ipopt with `-fPIC`, removing the conflict that I have added.

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -109,8 +113,6 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("coinhsl+blas", when="+sparse")
     depends_on("metis", when="+sparse")
 
-    depends_on("ginkgo@glu+cuda", when="+ginkgo")
```
pelesh, line 112: We need to keep Ginkgo dependency.

pelesh:
> I think these changes should be merged, but perhaps there is a way to compile ipopt with `-fPIC`, removing the conflict that I have added.

The conflict you added is very much a fringe case as far as I can see. Few people will need Ipopt and ROCm at the same time as Ipopt has no GPU support whatsoever. Afaik Ipopt can be built with -fPIC flag enabled. We can troubleshoot that offline.

pelesh (changes requested):
We need to keep Ginkgo dependency.

var/spack/repos/builtin/packages/hiop/package.py
cameronrutherford: I have re-added this. Thanks for catching

cameronrutherford:
> > I think these changes should be merged, but perhaps there is a way to compile ipopt with `-fPIC`, removing the conflict that I have added.
> 
> The conflict you added is very much a fringe case as far as I can see. Few people will need Ipopt and ROCm at the same time as Ipopt has no GPU support whatsoever. Afaik Ipopt can be built with -fPIC flag enabled. We can troubleshoot that offline.

I was unable to build ipopt with `-fPIC` through spack, and this was a block for building with the exago python wrapper enabled. While it is a fringe case, I think that it is a consistent bug with the amd stack, and is fine to be included to the spack package.

pelesh (dismissed):

eugeneswalker (dismissed):

alalazo (dismissed):

alalazo:
@spackbot fix style

alalazo:
@spackbot fix style

alalazo:
@spackbot fix style

alalazo:
@CameronRutherford Can you please take care of fixing style manually, and ping me back for a merge?

cameronrutherford:
@alalazo I ran `spack style --fix` and pushed those changes. Thanks for trying to help with the bot

alalazo (approved):

PR #31883, opened 2022-08-02
E4S: Convert specs to use Data and Vis SDK

kwryankrattiger:
Updates to E4S release to integrate with ECP Data and Vis SDK

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml
@@ -54,27 +54,28 @@ spack:
   specs:
   # CPU
   - adios
-  - adios2
+  #- adios2
   - alquimia
   - aml
   - amrex
   - arborx
   - archer
   - argobots
-  - ascent
+  #- ascent
   - axom
   - bolt
   - bricks
   - butterflypack
   - cabana
   - chai ~benchmarks ~tests
-  - conduit
-  - darshan-runtime
-  - darshan-util
+  #- conduit
+  #- darshan-runtime
+  #- darshan-util
   - datatransferkit
   - dyninst
   - exaworks
-  - faodel
+  - ecp-data-vis-sdk +adios2+ascent+cinema~cuda+darshan+faodel+hdf5+paraview+pnetcdf~rocm~sensei+sz+unifyfs+veloc+visit+vtkm+zfp
+  #- faodel
   - flecsi
   - flit
   - flux-core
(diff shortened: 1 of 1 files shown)


eugeneswalker:
I'm not sure we can replace all of the individual root specs commented out in this PR with the SDK metapackage. The reason being that we want to make sure E4S products continue to be installable individually with Spack.

What do you think about adding the data-viz-sdk meta-package without uncommenting everything else? This would ensure the products continue to be installable as they would concretize individually (i.e. without the additional constraints and logic embedded in the SDK metapackage) while still ensuring the data-viz SDK is included in E4S as you and other maintainers of it intend.

What do you think @wspear? Who else could we ask for comments on this?

kwryankrattiger:
> E4S products continue to be installable individually with Spack

The individual packages should still be individually installable, the SDK just forces options on that may not be on by default and that inter-op features are enabled.

eugeneswalker:
> > E4S products continue to be installable individually with Spack
> 
> The individual packages should still be individually installable, the SDK just forces options on that may not be on by default and that inter-op features are enabled.

If we remove the specs individually, and allow data viz sdk meta package to install them, then we are no longer providing the same coverage, as now data viz sdk is managing what variants are on. Obviously it is a benefit to use data viz sdk as we are benefiting from the well considered interoperability decisions you make in the meta package. But at the same time, if a user wants to take the E4S environment and only install a subset of the whole data viz sdk, we are no longer testing the build they would get doing that, because we are now testing only the builds that occur as a result of applying the variant/version decisions made by data-viz-sdk. Do you see what I'm saying? I'm not sure we can do this. We will need to solicit some feedback from others about the best way to proceed.

wspear:
Would it be feasible to make our individual product specs match the ones installed by the SDK closely enough that a 'reuse' style concretization would allow the SDK to pick up our individually specified specs as its dependents?

eugeneswalker:
Can you rebase this on the tip of spack@develop so that we can get the GitLab pipelines to run?

share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml
```diff
@@ -149,15 +150,16 @@ spack:
...
-  - vtk-m
+  #- veloc
+  #- vtk-m
   - wannier90
-  - zfp
+  #- zfp
 
   # CUDA
-  - adios2 +cuda cuda_arch=80
+  - ecp-data-vis-sdk +adios2+ascent+cinema+cuda+darshan+faodel+hdf5+paraview+pnetcdf~rocm~sensei+sz+unifyfs+veloc+visit+vtkm+zfp cuda_arch=80
```
kwryankrattiger, line 161: ~paraview

kwryankrattiger:
Closed in favor of #32223

PR #31887, opened 2022-08-03
VisIt: Add vtkm variant.

brugger1:
Added a vtkm variant to the VisIt package.

The vtkm support in VisIt is done through vtkh, so that support is also enabled in VisIt as part of the vtkm variant.

It includes a patch to VisIt to include the vtkh libraries in the install.
It includes a patch to vtkm that prevents throwing an exception with some vtkm functionality when used in VisIt.
The variant disables cuda in vtkm and vtkh since the cuda support doesn't work with VisIt with the current versions of vtkm and vtkh. It will in the next release.

I tested this on "crusher.olcf.ornl.gov", an early access system for the Frontier exascale system at Oak Ridge National Lab. I successfully displayed images in the GUI.


alalazo:
@cyrush Feel free to ping me for a merge when this is ready.

cyrush (approved):

cyrush:
@brugger1 @alalazo everything looks good

alalazo:
@spackbot run pipeline

PR #31905, opened 2022-08-03
Meaningful diagnostics on unimplemented phase

greenc-FNAL:
Produce meaningful diagnostics when a recipe invokes a `@run_before()`
or `@run_after()` directive for a phase not implemented in any package
base class, _e.g._:

```python
class PyBadPackage(PythonPackage):
    ...
    @run_before("build")
    def my_early_func(self)
        ...
```

would fail with:

```console
$ spack spec -It --reuse python@3.9.12
==> Error: bad directive @run_before() for functions [<function PyJltheme.ensure_readme_rst at 0x7fa46e16df80>] in package PyJltheme:
  no phase "build" in any base (<class 'spack.build_systems.python.PythonPackage'>,)
```

### DIFF
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -376,11 +376,13 @@ class PackageMeta(
                         else:
                             tty.die(
                                 """\
-bad directive @{directive}() for functions {funcs} in package {pkg}:""".format(
-                                    directive=check_name, funcs=funcs, pkg=name
+bad callback decorator @{decorator}() for functions {funcs} in package {pkg}:""".format(
+                                    decorator=check_name,
+                                    funcs=[f.__name__ for f in funcs],
+                                    pkg=name,
                                 ),
-                                'no phase "{phase}" in any base {bases}'.format(
-                                    phase=phase_name, bases=bases
+                                'no phase "{phase}" in {pkg} or any base class thereof'.format(
+                                    phase=phase_name, pkg=name
                                 ),
                             )
 
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -373,6 +373,14 @@ class PackageMeta(
                             phase = getattr(base, phase_attr, None)
                             if phase is not None:
                                 break
+                        else:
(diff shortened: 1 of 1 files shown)


lib/spack/spack/package_base.py
```diff
@@ -373,6 +373,16 @@ def _flush_callbacks(check_name):
...
                                 break
+                        else:
+                            tty.die(
+                                """\
+bad directive @{directive}() for functions {funcs} in package {pkg}:""".format(
+                                    directive=check_name, funcs=funcs, pkg=name
+                                ),
+                                'no phase "{phase}" in any base {bases}'.format(
+                                    phase=phase_name, bases=bases
+                                ),
```
becker33: I think we should avoid showing too much python internals as part of the error message. Something like this might be easier to interpret.
```suggestion
                            tty.die(
                                """\
bad directive @{directive}() for functions {funcs} in package {pkg}:""".format(
                                    directive=check_name, funcs=[f.__name__ for f in funcs], pkg=name
                                ),
                                'no phase "{phase}" in any {cls} or any base class thereof'.format(
                                    phase=phase_name, cls=cls
                                ),
```

becker33 (changes requested):

lib/spack/spack/package_base.py
greenc-FNAL: `cls` seems a little too low-level still: `<class 'spack.package_base.PackageMeta'>`: how about just `name` (`PyBadPackage`)?

greenc-FNAL:
@becker33 I understand this needs a test, but I have no experience writing a test that needs to read a package recipe that would poison everything if it were part of the mock area. Is there an example of this I can examine that you know of?

Essentially it needs to be a recipe with one or more defined functions with `@run_before()` or `@run_after()` decorators for phases that aren't defined for the current package and its base classes.

greenc-FNAL:
@becker33 Ping?

alecbcs:
ping @becker33

greenc-FNAL:
No longer meaningful after the advent of `Builder`

PR #31916, opened 2022-08-03
seq-gen: add seq-gen package to spack

downloadico:


var/spack/repos/builtin/packages/seq-gen/package.py
```diff
@@ -0,0 +1,30 @@
...
+    """Seq-Gen is a program that will simulate the evolution of nucleotide or
+    amino acid sequences along a phylogeny, using common models of the
+    substitution process. A range of models of molecular evolution are implemented
+    including the general reversible model. State frequencies and other parameters
+    of the model may be given and site-specific rate heterogeneity may also be
+    incorporated in a number of ways. Any number of trees may be read in and the
+    program will produce any number of data sets for each tree. Thus large sets of
+    replicate simulations can be easily created. It has been designed to be a
+    general purpose simulator that incorporates most of the commonly used (and
+    computationally tractable) models of molecular sequence evolution."""
```
alalazo: Can we have a shorter (ideally one-liner) short paragraph, followed by a longer description? The first paragraph in the docstring is used in contexts where a short description is needed.

alalazo (changes requested):
Only a minor request

alalazo (approved):

PR #31930, opened 2022-08-04
Windows rpath support

johnwparent:
This PR vendors initial pass Windows rpath analog support. Windows does not understand the concept of an RPATH, so Spack must implement similar functionality and perform such actions manually. This is accomplished via symlinking shared libraries, later to be replaced with a more sophisticated method. This PR is intended as a temporary stop gap support for the concept of RPATHs on Windows.

PR is currently in progress and pending review.


lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+    """Convenience function that returns the list of all shared libraries found
+    in the directory passed as argument.
+
+    Args:
+        root (str): directory where to look recursively for library files
+        recursive (bool):
+    Returns:
+        List of all libraries found in ``root`` and subdirectories.
+    """
+    return find_libraries('*', root=root, recursive=recursive)
```
scheibelp: This should explicitly set `shared=True`

lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+
+    return find_all_shared_libraries(root, recursive=recursive) +\
+        find_all_static_libraries(root, recursive=recursive)
+
+
+class WindowsRuntimePath:
+    """ Class representing Windows filesystem rpath analog
+
+    rpath is established for each library that a package's
+    executables or libraries require dynamic linking to.
```
scheibelp: I'm having trouble parsing this. In particular "rpath is established"

lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+            package (spack.package_base.PackageBase): Package requiring links
+            link_install_prefix (bool): Link against package's own install or stage root.
+                Packages that run their own executables during build and require rpaths to
+                the build directory during build time require this option. Default: install
+                root
+        """
+        self.pkg = package
+        self._addl_rpaths = []
+        self.link_install_prefix = link_install_prefix
+        self._internal_links
```
scheibelp: This is missing an assignment

lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        Set of directories where package binaries are located. Symlinks
+        to libraries/binaries on which this package depends
+        """
+        pkg_libs = set(self.pkg.libs.directories)\
+            if hasattr(self.pkg, 'libs') else set((self.pkg.prefix.lib, self.pkg.prefix.lib64))
+        return pkg_libs | set(self.internal_links)
+
+    @property.setter
+    def internal_links(self, *dest):
+        self._internal_links.extend(dest)
```
scheibelp: The setter should be idempotent.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        """
+        self.pkg = package
+        self._addl_rpaths = []
+        self.link_install_prefix = link_install_prefix
+        self._internal_links
+
+    @property
+    def link_dest(self):
+        """
+        Set of directories where package binaries are located. Symlinks
```
scheibelp: "where package binaries are located": since this doesn't return `bin/` (and therefore excludes executables), I think a term other than "binary" should be used.

lib/spack/spack/build_systems/intel.py
```diff
@@ -1383,3 +1383,6 @@ def modify_LLVMgold_rpath(self):
 
     # Check that self.prefix is there after installation
     run_after('install')(PackageBase.sanity_check_prefix)
+    # On Windows, concept of rpath does not exist, after install
+    # establish runtime linkage via Windows Runtime link object
+    run_after('install')(PackageBase.windows_establish_runtime_linkage)
```
scheibelp: This appears to be done for every build system, so it should be moved up to a consolidated location (e.g. `installer._real_install`). I would think that since you placed it in `PackageBase`, that would avoid the need to do it for each build system, but if not I think it's still worth finding a way to avoid doing it per-build-system.

lib/spack/spack/package_base.py
```diff
@@ -182,6 +182,41 @@ def copy(self):
...
+class WindowsRPathMeta(object):
+    """Collection of functionality surrounding Windows RPATH specific features
+
+    This is essentially meaningless for all other platforms
+    due to their use of RPATH. All methods within this class are no-ops on
+    non Windows. Packages can customize and manipulate this class as
+    they would a genuine RPATH, i.e. adding directories that contain
+    runtime binary dependencies"""
+
+    win_rpath = None
```
scheibelp: I think this should be a per-object member rather than a per-class member.

var/spack/repos/builtin/packages/clingo/package.py
```diff
@@ -65,6 +77,10 @@ def patch(self):
                         'clasp/CMakeLists.txt',
                         'clasp/libpotassco/CMakeLists.txt')
 
+    @property
+    def libs(self):
+        pass
```
scheibelp: (question) Is this intentionally returning `None` to avoid returning something problematic (i.e. based on the superclass definition)?

lib/spack/spack/package_base.py
```diff
@@ -182,6 +182,41 @@ def copy(self):
...
+        """ Denote a package component not included in the
+        binary install directories defined by a package that requires
+        linking to the packages dependencies or the package binaries themselves
+
+        A good example of this is a python extension module that requires dynamic
+        linking to it's dependent C++ binary, and similar dependency chains
+        """
+        self.win_rpath.interal_links = dest
+
+    def add_search_paths(self, *path):
```
scheibelp, line 197: I don't see any calls to this method. Is there a case where you think this will be needed? To be clear I think I see how packages might want to customize this, but a concrete example would be useful.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        recursive (bool):
+    Returns:
+        List of all libraries found in ``root`` and subdirectories.
+    """
+
+    return find_all_shared_libraries(root, recursive=recursive) +\
+        find_all_static_libraries(root, recursive=recursive)
+
+
+class WindowsRuntimePath:
```
scheibelp: (general) We haven't shed Python 2 support yet so we should inherit from `object` for now.

Also, could you rename this `WindowsSimulatedRpath`? Or the main thing I want to avoid is confusing this with `runpath` (a similar concept with differing semantics on Linux), so any name that excludes "run" should be fine.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        for link in itertools.product(self.link_dest, self.link_targets):
+            symlink(link[0], link[1])
+        # from build_environment.py:463
+            # The top-level package is always RPATHed. It hasn't been installed yet
+            # so the RPATHs are added unconditionally
+
+        # this naively symlinks from lib to bin and bin to lib etc
+        # in the future could be more nuanced and internal links handled
+        # by package attributes
+        for link in itertools.product(self.link_dest, repeat=2):
```
scheibelp: For the first linking "phase" above we do:

```
for link in itertools.product(self.link_dest, self.link_targets):
```

this produces a set of tuples `(x, y)` where `x` is a directory (i.e. where we want to place a symlink) and `y` is a library (i.e. the full path to a `.dll`). 

For this phase, the tuples are of the form `(directory, directory)` and I'm not clear what the intended result is: is the executable expected to search the symlinked directories?

lib/spack/llnl/util/filesystem.py
```diff
@@ -2002,6 +2020,140 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+
+    @property
+    def link_targets(self):
+        """
+        Set of libraries this package needs to link against during runtime
+        These packages will each be symlinked into the packages lib and binary dir
+        """
+
+        dependent_libs = []
+        for path in self.pkg.rpath:
```
scheibelp, line 2198: (question) Should `PackageBase.rpath` also include binary directories with library files on Windows?

scheibelp (commented):
Preliminary review. Feel free to ignore lower-level criticisms (only 2-3 of the comments are high-level).

lib/spack/spack/package_base.py
johnwparent: I haven't come across a specific use case as of yet, but thought it would be useful down the road. I'll add an example.

var/spack/repos/builtin/packages/clingo/package.py
johnwparent: This is part of the Clingo PR. This branch is based off that PR for testing purposes. I've removed that section of this branch's history locally and will update the upstream when appropriate, so this is probably best left for discussion on that PR.

lib/spack/spack/build_systems/intel.py
johnwparent: Sounds good, I agree.

lib/spack/llnl/util/filesystem.py
johnwparent: Not sure how that got dropped... or how I missed it. Apologies, amended.

lib/spack/llnl/util/filesystem.py
johnwparent: This is designed to iterate over libraries internal to the package, i.e. `Package.prefix.bin` <-> `Package.prefix.lib`, so in this phase the tuple is of form `(bindir, libdir)` and similar, so the next lines search the second directory in the tuple for full paths to `.dll` files (provided both directories are not the same) and then performs the symlinking.

This is not my ideal approach to this, and I believe (as denoted in the comment above this codeblock) that something along the lines of a package attribute would be appropriate here, but I wanted your insight before I stepped down that road.

lib/spack/llnl/util/filesystem.py
johnwparent: Thinking about this more, I think it should in fact return a package's bin directory, as most executables will likely require the ability to 'link' against their dependencies as well.

lib/spack/llnl/util/filesystem.py
johnwparent: My instinct is to say yes, and in my experience that is the case, but with that said we may want to open this topic to larger discussion.

lib/spack/spack/package_base.py
```diff
@@ -181,6 +184,39 @@ def copy(self):
...
+
+    def establish_internal_linkage(self, *dest):
+        """Denote a package component not included in the
+        binary install directories defined by a package that requires
+        linking to the packages dependencies or the package binaries themselves
+
+        A good example of this is a python extension module that requires dynamic
+        linking to it's dependent C++ binary, and similar dependency chains
+        """
+        self.win_rpath.interal_links = dest
```
scheibelp: My original thought on this was that if we can have a function that collects all libraries recursively, then we don't need a function like this.

(1) I'll add this marker for reference from another review comment

lib/spack/spack/package_base.py
```diff
@@ -181,6 +184,39 @@ def copy(self):
...
 
 
+class WindowsRPathMeta(object):
+    """Collection of functionality surrounding Windows RPATH specific features
+
+    This is essentially meaningless for all other platforms
+    due to their use of RPATH. All methods within this class are no-ops on
+    non Windows. Packages can customize and manipulate this class as
+    they would a genuine RPATH, i.e. adding directories that contain
+    runtime binary dependencies"""
```
scheibelp: I think the term "library" would be more precise than "binary". IIRC we spoke about libraries that may be placed in the `bin/` directory on windows, but I think it still makes sense to refer to them as libraries.

lib/spack/spack/build_systems/python.py
```diff
@@ -341,3 +341,7 @@ def remove_files_from_view(self, view, merge_map):
                 os.remove(dst)
 
         view.remove_files(to_remove)
+
+    # On Windows, concept of rpath does not exist, after install
+    # establish runtime linkage via Windows Runtime link object
+    run_after("install")(PackageBase.windows_establish_runtime_linkage)
```
scheibelp: I assume all packages would need to do this (my understanding is that `windows_establish_runtime_linkage` kicks off the generation of symblinks which would be needed for any package with dependencies) - why is it just being done for Python packages?

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,154 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        # from build_environment.py:463
+        # The top-level package is always RPATHed. It hasn't been installed yet
+        # so the RPATHs are added unconditionally
+
+        # for each binary install dir in self.pkg (i.e. pkg.prefix.bin, pkg.prefix.lib)
+        # install a symlink to each dependent library
+        for library, lib_dir in itertools.product(self.link_targets, self.link_dest):
+            if lib_dir not in library:
+                file_name = os.path.basename(library)
+                dest_dir = os.path.join(lib_dir, file_name)
```
scheibelp: Calling this `dest_dir` seems confusing since it's supposed to be a file (and not a directory)

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,154 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+    def internal_links(self):
+        """
+        linking that would need to be established within the package itself. Useful for links
+        against extension modules/build time executables/internal linkage
+        """
+        return self._internal_links
+
+    @internal_links.setter
+    def internal_links(self, *dest):
+        self._internal_links = self._internal_links | set(dest)
```
scheibelp: If possible, I think it's best to avoid non-idempotent setter behavior (e.g. create a function called `add_internal_links`).

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,154 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        # The top-level package is always RPATHed. It hasn't been installed yet
+        # so the RPATHs are added unconditionally
+
+        # for each binary install dir in self.pkg (i.e. pkg.prefix.bin, pkg.prefix.lib)
+        # install a symlink to each dependent library
+        for library, lib_dir in itertools.product(self.link_targets, self.link_dest):
+            if lib_dir not in library:
+                file_name = os.path.basename(library)
+                dest_dir = os.path.join(lib_dir, file_name)
+                if os.path.isdir(lib_dir):
```
scheibelp: When is `lib_dir` not a directory? You construct it in the `lib_dir` property and I only see things that would ever be files except possibly for `internal_links` - which could be filtered if needed.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,154 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+    def link_dest(self):
+        """
+        Set of directories where package binaries/libraries are located.
+        """
+        if hasattr(self.pkg, "libs") and self.pkg.libs:
+            pkg_libs = set(self.pkg.libs.directories)
+        else:
+            pkg_libs = set((self.pkg.prefix.lib, self.pkg.prefix.lib64))
+
+        return pkg_libs | set([self.pkg.prefix.bin]) | self.internal_links
```
scheibelp, line 2167: In relation to (1) there's a couple ways I think this could be updated to avoid the need for `establish_internal_linkage`:

* Have this function look for libraries in `bin/` etc.
* Or we could update the implementation of `_libs_default_handler` to do this (and expect that custom overrides of `.libs` in packages would search `bin/` etc.)

If the files that need to symlink libraries are not strictly libraries themselves (i.e. other libraries/exe's don't need them) then the first option is probably better.

scheibelp (changes requested):
Additional questions/requests.

lib/spack/spack/package_base.py
johnwparent: Oh this is actually vestigial for exactly the reasons you articulated. I hadn't thought about it. Thank for pointing it out.

lib/spack/spack/build_systems/python.py
johnwparent: This one I missed when refactoring away from this pattern as we discussed last week. Thanks for catching it.

lib/spack/llnl/util/filesystem.py
johnwparent: That's fair, will rename.

lib/spack/llnl/util/filesystem.py
johnwparent: Correct me if I misunderstood, but we do select the bin directory here for inclusion into the linking scheme.

lib/spack/llnl/util/filesystem.py
johnwparent: This should be `os.path.exists`

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,165 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        Incorporate additional paths into the rpath (sym)linking scheme.
+
+        Paths provided to this method are linked against by a package's libraries
+        and libraries found at these paths are linked against a package's binaries.
+        (i.e. /site-packages -> /bin and /bin -> /site-packages)
+
+        Specified paths should be outside of a package's lib, lib64, and bin
+        directories.
+        """
+        self._internal_links = self._internal_links | set(*dest)
```
scheibelp, line 2188: I have the same concern here as https://github.com/spack/spack/pull/31930#discussion_r954329663: if we can detect all the library files, then we wouldn't need to ever manually add them (and this function is not needed).

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,165 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+            if lib_dir not in library:
+                file_name = os.path.basename(library)
+                dest_file = os.path.join(lib_dir, file_name)
+                if os.path.exists(lib_dir):
+                    try:
+                        symlink(library, dest_file)
+                    # For py2 compatibility, we have to catch the specific Windows error code
+                    # associate with trying to create a file that already exists (winerror 183)
+                    except OSError as e:
+                        if e.winerror == 183:
```
scheibelp, line 2237: Will this check fail on Linux? Is there an analogous situation on Linux where it should also `pass`?

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,165 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        # The top-level package is always RPATHed. It hasn't been installed yet
+        # so the RPATHs are added unconditionally
+
+        # for each binary install dir in self.pkg (i.e. pkg.prefix.bin, pkg.prefix.lib)
+        # install a symlink to each dependent library
+        for library, lib_dir in itertools.product(self.link_targets, self.link_dest):
+            if lib_dir not in library:
+                file_name = os.path.basename(library)
+                dest_file = os.path.join(lib_dir, file_name)
+                if os.path.exists(lib_dir):
```
scheibelp, line 2231: This check seems unneeded: these files/directories were gathered by spidering existing paths. Do you anticipate a case where this would not be true?

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,165 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        Windows heuristics
+        """
+        # from build_environment.py:463
+        # The top-level package is always RPATHed. It hasn't been installed yet
+        # so the RPATHs are added unconditionally
+
+        # for each binary install dir in self.pkg (i.e. pkg.prefix.bin, pkg.prefix.lib)
+        # install a symlink to each dependent library
+        for library, lib_dir in itertools.product(self.link_targets, self.link_dest):
+            if lib_dir not in library:
```
scheibelp: Can you use `llnl.util.filesystem.path_contains_subdirectory` here?

lib/spack/spack/installer.py
```diff
@@ -1213,7 +1218,10 @@ def _install_task(self, task):
             spack.package_base.PackageBase._verbose = spack.build_environment.start_build_process(
                 pkg, build_process, install_args
             )
-
+            # On Windows, concept of rpath does not exist, after install
```
scheibelp: I think a more careful wording would be:

> Currently this is how RPATH-like behavior is achieved on Windows

since we've discussed modifying windows exe's/libraries to insert references to needed libraries, and that conceptually is much like an RPATH.

lib/spack/llnl/util/filesystem.py
scheibelp: > Correct me if I misunderstood, but we do select the bin directory here for inclusion into the linking scheme.

Yes but we don't recursively search it looking for link targets: if you have `bin/subdir/x.dll`, would we properly link to/from it?

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,165 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+    @property
+    def link_targets(self):
+        """
+        Set of libraries this package needs to link against during runtime
+        These packages will each be symlinked into the packages lib and binary dir
+        """
+
+        dependent_libs = []
+        for path in self.pkg.rpath:
+            dependent_libs.extend(list(find_all_shared_libraries(path, recursive=True)))
```
scheibelp, line 2199: `pkg.rpath` includes transitive dependencies: is it sufficient to just include direct link dependencies?

(3) (for reference)

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2096,165 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        for extra_path in self._addl_rpaths:
+            dependent_libs.extend(list(find_all_shared_libraries(extra_path, recursive=True)))
+        return set(dependent_libs)
+
+    def include_additional_runtime_paths(self, *paths):
```
scheibelp: Elsewhere you use the term "runtime library" and I think we should be careful: do you intend for there to be a distinction between "library" and "runtime library"? The only one I would draw is that "runtime libraries" includes the transitive libraries that are not directly linked.

The distinction could be meaningful depending on the answer to [3].

scheibelp (commented):
This follow up is mostly questions/responses (although depending on the answers, some changes may be useful)

lib/spack/llnl/util/filesystem.py
johnwparent: None of this code should ever be reachable from Unix based systems. If it were to be valuable from the Linux side, then yes, similar circumstances should also cause the Linux approach to `pass`.

lib/spack/llnl/util/filesystem.py
johnwparent: No, not at the moment. This is the scenario that I have been bringing up as a topic of concern. Trying to successfully balance digging into a package to find dependencies vs avoiding binaries/libraries that should not be linked against or overly verbose linking. I guess my thought is that this should follow similar rules as rpath/ld_library_path, i.e. not recursively dig into directories like `bin`.

lib/spack/llnl/util/filesystem.py
johnwparent: If we do chose to search the package recursively for binaries/libraries, we'd at the moment, only be searching `bin`, `lib`, and `lib64`, which leaves space for binaries/libs installed elsewhere to be overlooked. This method provides functionality to more selectively and precisely denote those locations. Alternatively, we could simply recurse from the package root, but that feels a bit dangerous.

lib/spack/llnl/util/filesystem.py
johnwparent: `/lib64/` is not present in a large number of package install prefixes. Further,  `lib_dir` here is derived from `self.link_dest` which is not generated by searching directories, and instead is derived from a predefined set of directories, or if the packages defines the appropriate attributes, paths specified by those attributes.

lib/spack/llnl/util/filesystem.py
johnwparent: Yes, thank you for calling this method to my attention! This is exactly what I am trying to do here.

lib/spack/spack/installer.py
johnwparent: That is true, and I'll update the phrasing.

lib/spack/llnl/util/filesystem.py
johnwparent: I did intend for there to be a difference w.r.t. to this method (other useages are an oversight/I had this in my mind while writing), where here runtime refers to exactly that, libraries that would not directly be linked but would still require it.

lib/spack/llnl/util/filesystem.py
johnwparent: I believe we still need to link to the transitive deps, however now that you ask, I'm not 100%, so I'll look into that this evening.

lib/spack/spack/test/llnl/util/file_list.py
```diff
@@ -52,6 +65,15 @@ def header_list():
     return h
 
 
+plat_static_ext = "lib" if is_windows else "a"
```
scheibelp, line 69: This looks good although I think a TODO comment should mention that this can be undone when `find_libraries` in `util.filesystem` is made Spec-aware (in a later PR).

var/spack/repos/builtin/packages/clingo/package.py
scheibelp: Would it make sense to only return `None` on Windows if this merges without that?

lib/spack/spack/test/conftest.py
```diff
@@ -803,7 +803,8 @@ def mock_store(tmpdir_factory, mock_repo_path, mock_configuration_scopes, _store
             with spack.store.use_store(str(store_path)) as store:
                 with spack.repo.use_repositories(mock_repo_path):
                     _populate(store.db)
-        store_path.copy(store_cache, mode=True, stat=True)
+        copy_tree(str(store_path), str(store_cache))
+        # store_path.copy(store_cache, mode=True, stat=True)
```
scheibelp: This commented-out line should be removed.

Normally I'd be concerned that it isn't preserving permissions the same way, but since all the tests pass, apparently that wasn't necessary.

lib/spack/llnl/util/filesystem.py
scheibelp: Ping on this question.

lib/spack/llnl/util/filesystem.py
scheibelp: > Alternatively, we could simply recurse from the package root, but that feels a bit dangerous.

It didn't seem dangerous since I assumed any libraries reachable from the base prefix are likely of interest, and libraries can be identified by their extensions (so there is no mistaking a non-library file as a library).

That being said, this doesn't have to be changed in this PR.

lib/spack/llnl/util/filesystem.py
scheibelp: This is related to https://github.com/spack/spack/pull/31930/files#r957640095. The resolution there was that we can keep this for now and tweak it to be more general in the future if need be.

lib/spack/llnl/util/filesystem.py
```diff
@@ -637,7 +637,11 @@ def copy_tree(src, dest, symlinks=True, ignore=None, _permissions=False):
                 if symlinks:
                     target = os.readlink(s)
                     if os.path.isabs(target):
-                        new_target = re.sub(abs_src, abs_dest, target)
+
+                        def escaped_path(path):
+                            return path.replace("\\", r"\\")
+
+                        new_target = re.sub(escaped_path(abs_src), escaped_path(abs_dest), target)
```
scheibelp, line 644: You should only have to escape the pattern (i.e. the first argument) here: it should not be required to escape the replacement string (although it also shouldn't cause problems other than if one printed out the path it would be in the regex-compatible format (which would work like a normal path but contain unnecessary `\`'s)

lib/spack/llnl/util/filesystem.py
```diff
@@ -637,7 +637,11 @@ def copy_tree(src, dest, symlinks=True, ignore=None, _permissions=False):
                 if symlinks:
                     target = os.readlink(s)
                     if os.path.isabs(target):
-                        new_target = re.sub(abs_src, abs_dest, target)
+
+                        def escaped_path(path):
+                            return path.replace("\\", r"\\")
```
scheibelp, line 642: Can this be `path.replace(r"\", r"\\")` - I think using the two different formats is slightly confusing here.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2100,163 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+def find_all_shared_libraries(root, recursive=False):
+    """Convenience function that returns the list of all shared libraries found
+    in the directory passed as argument.
+
+    Args:
+        root (str): directory where to look recursively for library files
+        recursive (bool):
+    Returns:
+        List of all libraries found in ``root`` and subdirectories.
+    """
```
scheibelp, line 2108: Can these (i.e. this, `find_all_static_libraries`, and `find_all_libraries`) just refer to the docstring of `find_libraries`? IMO this is too much documentation for convenience functions.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2100,163 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+class WindowsSimulatedRuntimePath(object):
+    """Class representing Windows filesystem rpath analog
+
+    One instance of this class is associated with a package (only on Windows)
+    For each lib/binary directory in an associated package, this class introduces
+    a symlink to any/all dependent libraries/binaries. This includes the packages
+    own bin/lib directories, meaning the libraries are linked to the bianry directory
+    and vis versa.
+
+
```
scheibelp: (minor) remove extra whitespace

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2100,163 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+    Returns:
+        List of all libraries found in ``root`` and subdirectories.
+    """
+
+    return find_all_shared_libraries(root, recursive=recursive) + find_all_static_libraries(
+        root, recursive=recursive
+    )
+
+
+class WindowsSimulatedRuntimePath(object):
```
scheibelp: In keeping with trying to avoid any association between this and the concept of runpath (in the same spirit as https://github.com/spack/spack/pull/31930#discussion_r939210169), can this be called `WindowsSimulatedRPath`?

lib/spack/llnl/util/filesystem.py
scheibelp: In that case, wherever you use the term "runtime" for this can you use "transitive" or just use the term `rpath` (e.g. this function could be called `include_additional_rpaths`. I'm trying to avoid introducing a new vocabulary term in this PR.

scheibelp (changes requested):
Additional requests, all minor (hopefully - let me know if you disagree).

lib/spack/llnl/util/filesystem.py
johnwparent: `r"\"` in Python is not a valid string as the `\` escapes the terminal `"`. Strings (even raw strings) cannot end in an odd number of back slashes.

var/spack/repos/builtin/packages/clingo/package.py
johnwparent: I think I'll be removing this entirely in favor of a new branch.

lib/spack/llnl/util/filesystem.py
johnwparent: My apologies for allowing this to languish. If the project is set up properly via CMake or MSBuild we shouldn't need to, however given how fragile (and inconsistent) Windows ports of many packages are, we should include transitive dependencies as a guarantee that packages will be able to find what they need. The DLL search heuristic seems to be the same for transitive as well as direct dependencies, so we'll need to have symbols in the transitive DLLs accessible through the same context as the direct deps.

lib/spack/spack/test/conftest.py
```diff
@@ -844,7 +844,8 @@ def mutable_database(database_mutable_config, _store_dir_and_cache):
     # Restore the initial state by copying the content of the cache back into
     # the store and making the database read-only
     store_path.remove(rec=1)
-    store_cache.copy(store_path, mode=True, stat=True)
+    copy_tree(str(store_cache), str(store_path))
+    # store_cache.copy(store_path, mode=True, stat=True)
```
scheibelp: This commented-out line should be removed

johnwparent: Thought I caught this, my bad. Will remove now.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2100,149 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        Windows heuristics
+        """
+        # from build_environment.py:463
+        # The top-level package is always RPATHed. It hasn't been installed yet
+        # so the RPATHs are added unconditionally
+
+        # for each binary install dir in self.pkg (i.e. pkg.prefix.bin, pkg.prefix.lib)
+        # install a symlink to each dependent library
+        for library, lib_dir in itertools.product(self.link_targets, self.link_dest):
+            if path_contains_subdirectory(library, lib_dir):
```
scheibelp: I'm guessing this should be `if not path_contains_subdirectory` (i.e. need a `not`)

johnwparent: This is correct. The issue has been amended.

lib/spack/llnl/util/filesystem.py
scheibelp: Update: it was pointed out I was wrong about this: the replacement string has its own syntax which attributes special meaning to `\`.

scheibelp:
@spackbot run pipeline

scheibelp:
@spackbot run pipeline

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2100,155 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+        """
+        Set of libraries this package needs to link against during runtime
+        These packages will each be symlinked into the packages lib and binary dir
+        """
+
+        dependent_libs = []
+        for path in self.pkg.rpath:
+            dependent_libs.extend(list(find_all_shared_libraries(path, recursive=True)))
+        for extra_path in self._addl_rpaths:
+            dependent_libs.extend(list(find_all_shared_libraries(extra_path, recursive=True)))
```
scheibelp, line 2201: [4] Ultimately I think it will be most idiomatic to iterate through dependencies (transitively) and ask for `.libs`: as it is, this essentially replicates that but excludes special cases (e.g. a specific package might know where to find libraries and `def libs` accordingly, but this would miss that). That can be refactored later.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2080,6 +2100,155 @@ def find_libraries(libraries, root, shared=True, recursive=False):
...
+
+    @property
+    def link_targets(self):
+        """
+        Set of libraries this package needs to link against during runtime
+        These packages will each be symlinked into the packages lib and binary dir
+        """
+
+        dependent_libs = []
+        for path in self.pkg.rpath:
```
scheibelp, line 2198: We spoke about this needing to be transitive at https://github.com/spack/spack/pull/31930/files#r957678049 - I should correct myself and say that `Package.rpath` only includes direct dependencies (not that this needs to change now given that this would likely change as part of [4]).

scheibelp (approved):
I have a couple of notes for future refactoring but I think this is in a good state to merge as is.

scheibelp:
@spackbot run pipeline

lib/spack/llnl/util/filesystem.py
```diff
@@ -2046,15 +2054,23 @@ def find_libraries(libraries, root, shared=True, recursive=False):
         message = message.format(find_libraries.__name__, type(libraries))
         raise TypeError(message)
 
+    if is_windows:
+        static = "lib"
+        shared = "dll"
+    else:
+        # Used on both Linux and macOS
+        static = "a"
+        shared = "so"
```
alalazo, line 2063: @mwkrentel @scheibelp @johnwparent That's the error. Here the `shared` argument is overridden, so we'll always hit the `True` condition in the conditional below.

scottwittenburg: Oh just found my way here looking into the `armadillo` build failure in the tutorial stack from today.  It can't find libraries in the relocated `superlu` binary.  I guess this is the source of the problem?

scottwittenburg: @zackgalbreath I think this :point_up: could the problem we've been looking into.

johnwparent: @alalazo Noticed this a bit ago, I have a solution for this if you haven't started.

https://github.com/spack/spack/pull/32653

scottwittenburg: No I didn't work on a solution, just getting to the point of figuring out the cause.

PR #31935, opened 2022-08-04
Add checksum for py-identify 2.5.3

blue42u:


adamjstewart (changes requested):
Python 3.7+ is required in this version

adamjstewart (approved):

PR #31936, opened 2022-08-04
Refactor `cmd/install.py` for better readability

alalazo:
Modifications:
- [x] Split long functions into multiple logical units
- [x] Renamed a few variables to better convey their properties (e.g. `specs` -> `concrete_specs`)
- [x] Removed dead code

### DIFF
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -332,30 +332,39 @@ def install_specs(cli_args, kwargs, specs):
         raise
 
 
-def install(parser, args, **kwargs):
-    # TODO: unify args.verbose?
-    tty.set_verbose(args.verbose or args.install_verbose)
-
-    if args.help_cdash:
-        parser = argparse.ArgumentParser(
-            formatter_class=argparse.RawDescriptionHelpFormatter,
-            epilog=textwrap.dedent(
-                """\
+def _print_cdash_help():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog=textwrap.dedent(
+            """\
 environment variables:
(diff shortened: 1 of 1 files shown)


lib/spack/spack/cmd/install.py
```diff
@@ -27,43 +27,28 @@
...
-        }
-    )
-
-    if hasattr(args, "setup"):
-        setups = set()
-        for arglist_s in args.setup:
-            for arg in [x.strip() for x in arglist_s.split(",")]:
-                setups.add(arg)
-        kwargs["setup"] = setups
-        tty.msg("Setup={0}".format(kwargs["setup"]))
```
alalazo, line 66: For the record: I completely removed this code during the refactor, since it doesn't seem to be called anywhere

alalazo:
@spackbot run pipeline

lib/spack/spack/cmd/install.py
```diff
@@ -27,43 +27,28 @@
...
+        "keep_stage": args.keep_stage,
+        "restage": not args.dont_restage,
+        "install_source": args.install_source,
+        "verbose": args.verbose or args.install_verbose,
+        "fake": args.fake,
+        "dirty": args.dirty,
+        "use_cache": args.use_cache,
+        "cache_only": args.cache_only,
+        "include_build_deps": args.include_build_deps,
+        "explicit": True,  # Always true for install command
```
haampie: FWIW, this comment is not 100% true anymore: `spack -e my_env install --no-add --only-concrete /hash` is `explicit` if and only if `/hash` is a root spec, but it's a sensible default...

alalazo: Should be fixed

alalazo:
Thanks for the offline comments and review @haampie !

haampie (approved):

PR #31914, opened 2022-08-03
Add documentation for git refs as versions

psakievich:
Extending #25981 to the current develop branch and updating for new features such as #31200  #30998 .  @becker33 feel free to rework this.

Closes #25981

PR #31937, opened 2022-08-04
Fix doc format for code blocks: GitRef versions

psakievich:
Rendering was not pretty using `console` from #31914


becker33 (approved):

adamjstewart:
Neither of these are shell scripts, should we just use "none"?

psakievich:
I was going off the formatting of the code blocks below. Shell looked nice since we added the comments with #'s. I don't have a strong preference though.

PR #31942, opened 2022-08-05
Enforcing ncurses with 'ABI=5' as dependency for ARM HPC Compiler 22.0.1 on Rocky Linux 8.6

fspiga:
Deploying ARM HPC Compiler 22.0.1 on Rocky Linux release 8.6 was successful but unable to use the compiler due to a missing library `libtinfo.so.5`. 

Solved by enforcing new ncurses with `abi=5` variant. It may be redundant for other OSes.

Suggestions how to improve this are welcome.

### DIFF
--- a/var/spack/repos/builtin/packages/arm/package.py
+++ b/var/spack/repos/builtin/packages/arm/package.py
@@ -78,7 +78,7 @@ class Arm(Package):
     conflicts("target=x86_64:", msg="Only available on Aarch64")
     conflicts("target=ppc64:", msg="Only available on Aarch64")
     conflicts("target=ppc64le:", msg="Only available on Aarch64")
-    depends_on('ncurses abi=5')
+    depends_on("ncurses abi=5")
     executables = [r"armclang", r"armclang\+\+", r"armflang"]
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/arm/package.py
```diff
@@ -149,3 +151,5 @@ def setup_run_environment(self, env):
         env.prepend_path("CPATH", join_path(arm_dir, "include"))
         env.prepend_path("MANPATH", join_path(arm_dir, "share", "man"))
         env.prepend_path("ARM_LICENSE_DIR", join_path(self.prefix, "licences"))
+        if "ncurses" in self.dependencies:
```
alalazo:
```suggestion
        if "ncurses" in self.spec:
```

alalazo (changes requested):

alalazo:
@spackbot fix style

OliverPerks (approved):
This looks like a very sensible addition to me. 
I have had a brief look and I dont think this would break support for any packages at the moment.

alalazo (approved):

PR #31945, opened 2022-08-05
LLVM: add @15.0.0-rc1

William-Mou:
Add llvm@15.0.0-rc1
It's an important version for the M1 mac.
Clang 15 (llvm@15.0.0) started to support `-march=native` for Apple M1.

ref: [why-does-march-native-not-work-on-apple-m1](https://stackoverflow.com/questions/65966969/why-does-march-native-not-work-on-apple-m1)

### DIFF
--- a/var/spack/repos/builtin/packages/llvm/package.py
+++ b/var/spack/repos/builtin/packages/llvm/package.py
@@ -36,7 +36,8 @@ class Llvm(CMakePackage, CudaPackage):
 
     # fmt: off
     version('main', branch='main')
-    version('14.0.6', sha256='98f15f842700bdb7220a166c8d2739a03a72e775b67031205078f39dd756a055')
+    version('15.0.0-rc1', sha256='b026a1b32ba0dc5612da36f14977c7e9fb910d545251a26dcfefca85d94139e4')
+    version('14.0.6', sha256='98f15f842700bdb7220a166c8d2739a03a72e775b67031205078f39dd756a055', preferred=True)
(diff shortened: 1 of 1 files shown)


trws:
Builds for me on an M1 mac, but this probably shouldn't be preferred since it's a pre-release version.

tldahlgren (dismissed):
Confirmed the sha256

tldahlgren:
Will defer merge decision to give maintainers a chance to comment.

trws:
@William-Mou, I'm good with adding this verison, but as it is it would become the default version of llvm built for `spack install llvm`, would you add `preferred=True` to the last stable version below this so we can avoid that?

William-Mou:
Hi @trws, Thank you for your reminder!
I've added this flag to the last stable version.

haampie:
Can we remove the rc version from Spack after LLVM 15 is released? Or will it stay around forever?

Notice that `spack install llvm@15.0.0-rc1` is possible too without this PR with the only downside that you don't get shasum verification on the sources.

William-Mou:
Hi @haampie,
I'm not sure if this is the question posed to me, but it seems to me that we should remove this version when llvm 15.0.0 is released.
In my view, `15.0.0-rc1` appears as an option in spack, which can enhance the confidence of users.

William-Mou:
Hi,
It has been a while since I responded.
Should I close this PR?

tldahlgren (approved):
Confirmed new sha256 and note @trws 's request for `14.0.6` to be preferred has be added.

tldahlgren:
Looks like CI pipeline has hung.

tldahlgren:
@spackbot run pipeline

PR #27610, opened 2021-11-23
Add in-place RPATH replacement

haampie:
Whenever the rpath string actually _grows_, it falls back to patchelf,
when it stays the same length or gets shorter, we update it in-place,
padded with null bytes.

This PR only deals with absolute -> absolute rpath replacement. We don't
use `_build_tarball(relative=True)` in our CI. If `relative` then it falls
back to the old replacement code.

With this PR, relocation time goes down significantly, likely because patchelf
does some odd things with mmap, causing lots of overhead. Example:

- `binutils`: 700MB installed, goes from `1.91s` to `0.57s`, or `3.4x` faster.
   Relocation time: 27% -> 10% of total install time
- `llvm`: 6.8GB installed, goes from `28.56s` to `5.38`, or `5.3x` faster.
   Relocation time: 44% -> 13% of total install time

The bottleneck is now decompression.

PR #31948, opened 2022-08-05
Experimental binding of shared ELF libraries

haampie:
Basically implements this idea: https://stoppels.ch/2022/08/04/stop-searching-for-shared-libraries.html (inspired by the macOS `install_name` tricks)

In short:

Adds another post install hook that loops over the install prefix, looking for shared libraries type of ELF files, and sets the soname to their own absolute paths.

The idea being, whenever somebody links against those libraries, the linker copies the soname (which is the absolute path to the library) as a "needed" library, so that at runtime the dynamic loader realizes the needed library is a path which should be loaded directly without searching.

As a result:

1. rpaths are not used for the fixed/static list of needed libraries in the dynamic section (only for _actually_ dynamically loaded libraries through `dlopen`), which largely solves the issue that Spack's rpaths are a heuristic (`<prefix>/lib` and `<prefix>/lib64` might not be where libraries really are...)
2. improved startup times (no library search required)

To enable, run 

```
spack config add config:shared_linking:bind:true
```


tgamblin:
@trws you probably want to take a look at this.
@fzakaria: FYI

trws:
Yup, @haampie and I talked about this, brought it up in the packaging taxonomy channel too. We need to do some testing to see what the effect is, and if it would let us get away with removing any of what we are doing with rewriting currently, but it could be a really nice way to deal with some problems, potentially even get musl working with shrinkwrap if we can touch the dependencies.

fzakaria:
Very cool — I remember us discussing this.
I think it makes a lot of sense for store-based systems where you know the name of the store address ahead of time and where you are writing too. I too am curious if it’s part of the SystemV spec and musl supports it.

With @trws I keep discussing how store-based systems should go deeper.
Why even offer a way for soname to be generated by the compiler in these systems. A patched compiler should always do the right thing :)

Why does the code do so much of the ELF checking itself but also relies on patchelf?
Surely seeing the soname and all the other elf stuff can be done then by relying on another tool if the hooks are not supposed to isolated. (I.e. could any tool help determine if it’s a shared library vs. executable?)
(FWIW though, I don’t write much Python and enjoyed reading this code)

haampie:
> Why does the code do so much of the ELF checking itself but also relies on patchelf?

I wanted to prevent weird situations where executables get sonames, patchelf happily does that. And I did not want to rely too much on conventions such as executables go in `bin/` and libraries go in `lib/`, since that's exactly the heuristic used for our rpaths, which can be wrong.

However, I've found that `libQt5Core.so.5` is an exception to the rule... it's a shared library with a soname that also has a PT_INTERP section, it is in fact an executable shared library... no clue how that is accomplished.

Maybe a better answer to the Q "is this a shared object" is: yes if it is `ET_DYN`, has a dynamic section, and either a soname or lacking a PT_INTERP. Then Qt's core lib qualifies as a shared lib too...

haampie:
- [x] Pushed a fix for `libQt5Core.so`.
- [x] Another issue is `openjdk`, it depends deeply on the glibc api, and the change in soname seems to break some invariant/assumption. Since we don't build openjdk from sources, there's nothing I can do except for excluding it.
- [x] ~Finally it seems `patchelf` is breaking libcublas in the amazon linux pipeline. Probably cause it's using an ancient broken version :(.~ confirmed this is just amazon linux 2 shipping a glibc too old to support cuda toolkit on aarch64.

haampie:
@tgamblin this seems to work just fine :tm:

The only build failure is now mfem on aarch64 amazon linux, but that failure is legit since cuda toolkit requires a newer glibc.

See https://github.com/spack/spack/pull/32206 and [attempt 1](https://gitlab.spack.io//spack/spack/-/pipelines/156355) and [attempt 2](https://gitlab.spack.io/spack/spack/-/pipelines/157162), 2 attempts because gitlab ci has tons of spurious kubernetes isues. Note that I tweaked CI to have a recent patchelf, unclear if ultimately it was necessary or not.

How to proceed from here? Enable this universally? Hide it behind a config flag? (I would prefer not, since the config value is not reflected in the hash -- different sonames under same hash doesn't sound right). Document it? Announce it?

bartoldeman:
I like this approach a lot (it could be ported to EasyBuild as well as an alternative to its rpath support), as it eliminates both LD_LIBRARY_PATH and the somewhat unreliable LD_RUN_PATH (alleviated by using a wrapper for `ld` as Nix does).

One thing I'm not sure about are the legal issues though, say you patch the soname of CUDA or MKL libraries, is that copyright or EULA infringement? Or only if you distribute said patched libraries?

haampie:
I doubt it. It's like changing a filename.

Worst case we could exclude some libraries from getting patched.

haampie:
Any news here @tgamblin / @trws / @becker33? At least a time line about when you would wanna review it?

It's just annoying this stuff never gets reviewed, which is exactly what happened in this very similar PR https://github.com/spack/spack/pull/27610 from Nov 23, 2021 which would drop the dependency on patchelf for the vast majority of cases :(.

trws:
Apologies @haampie, I'm going to review now.  It's been a crazy month 😞

lib/spack/spack/hooks/__init__.py
```diff
@@ -44,11 +41,18 @@ def __init__(self, hook_name):
     def _populate_hooks(cls):
         # Lazily populate the list of hooks
         cls._hooks = []
-        relative_names = list(llnl.util.lang.list_modules(spack.paths.hooks_path))
 
-        # We want this hook to be the last registered
-        relative_names.sort(key=lambda x: x == "write_install_manifest")
-        assert relative_names[-1] == "write_install_manifest"
+        # Since the order is important, we just fix it here, it's not like
+        # hooks get added every other day.
```
trws: I'm fine with having this be a static list, but perhaps have a test or audit that checks it's up to date?

lib/spack/spack/hooks/absolutify_elf_sonames.py
```diff
@@ -0,0 +1,291 @@
...
+        return not has_interpreter
+    except struct.error:
+        return False
+
+
+class SharedLibrariesVisitor(BaseDirectoryVisitor):
+    def __init__(self, exclude_list):
+
+        # List of file and directory names to be excluded
+        self.exclude_list = set(exclude_list)
```
trws: I don't see this having other things added, any reason it can't be a frozenset to make the `in` checks faster?

trws (commented):
Ok, I like this a lot.  I have some thoughts on factoring and other things, but the functionality assuming a wholly spack system looks good.  I'm wondering about how well this works out with externals and composition though.  I don't think it will break or anything, but this resolves the rpath issue only for things that have had their soname patched at link time right?

I'm wondering if we should combine this approach with what @fzakaria and I were up to on shrinkwrap to hoist the needed entries and paths for dependencies that don't have the appropriate soname transformation applied.  The two together seems to cover a whole lot of cases.

@becker33 did mention something this morning though about a race condition caused by having a library loaded both from needed and by its original un-modified soname in the same binary.  Do you have an example of that we could test with @becker33?

tgamblin:
@haampie: is this going to break on Alpine Linux or other `musl`-based distros?  Specifically, is it going to result in duplicate versions of librarie that are `dlopen`'d?  See this issue Julia still has with `musl`:  https://github.com/JuliaLang/julia/issues/40556

With `RPATH` I think you'll just see duplicated library loads, since `RPATH` has higher precedence than `LD_LIBRARY_PATH`, but with `RUNPATH`, if the user sets `LD_LIBRARY_PATH`, you could see two conflicting versions of the same library loaded.  That problem *sort of* exists already in the `RUNPATH` case since if you get a conflict you were probably loading an incompatible library to begin with (but at least there would not be two versions of it).

I am not sure what the best fallback is for the `musl` case.  I do think duplicated libs could be serious, though, if they have static storage and two modules in the same code expect to see the same state (but don't).

trws:
To add a bit of context to the julia musl issue, it shouldn't cause problems here (though we should always test).  The normal glibc thing is that it uses the soname as the cache key in the loading binary, so as long as the soname matches it's all good.  This is what musl does (pythonic pseudocode):

```python
cache_by_inode = {}
cache_by_soname = {}
if soname in cache_by_soname:
  return cache_by_soname[soname]
lib = find_lib(soname) # raises exception if not found
if '/' in soname:
  path = soname
  if lib.inode in cache_by_inode:
    cache_by_soname[lib.soname] = lib
    return lib
  cache_by_inode[lib.inode] = lib
  return lib
cache_by_soname[soname] = lib
return lib
```

The upshot is that it deduplicates libraries loaded by path *by inode* rather than by soname, but if the library is loaded by soname after by path and loads the same library then it is added to the soname cache for the rest of the run.  It's odd, but as long as we have *both* RPATH and this, it should always do the right thing because any attempt to load by soname will find the same inode anyway.

tgamblin:
> The upshot is that it deduplicates libraries loaded by path by inode rather than by soname, but if the library is loaded by soname after by path and loads the same library then it is added to the soname cache for the rest of the run.

Oh this is actually pretty awesome.  Ok I no longer have this concern.

haampie:
@bollig this pr is blocked because cuda toolkit for arm is not supported on amazon linux due to an old glibc. Can you please remove cuda stuff from the arm/amazonlinux pipelines?

haampie:
@tgamblin is amazon still alive?!

mboisson:
That's an interesting idea. A word of caution though, from Jeff Squyres recently in a thread about RPATH/RUNPATH issues with OpenMPI: 

The longer answer is that whenever you think you understand the shared library and run-time linkers, you inevitably find out that you don't.

trws:
The more I learn about them, the more I'm certain that's true.

tgamblin:
@mboisson:

> The longer answer is that whenever you think you understand the shared library and run-time linkers, you inevitably find out that you don't.

Is there something specific you see wrong here?

tgamblin:
Rebasing on develop to see if I can get CI to work.

var/spack/repos/builtin/packages/cuda/package.py
```diff
@@ -569,3 +569,8 @@ def libs(self):
             if "compat" not in parts and "stubs" not in parts:
                 filtered_libs.append(lib)
         return LibraryList(filtered_libs)
+
+    @property
+    def non_bindable_shared_objects(self):
+        # Avoid binding stub libraries by absolute path
+        return ["stubs"]
```
tgamblin: Not that it *really* matters but why not start off with an actual property?  Do you expect many of these to need a whole method?

```suggestion
    #: Avoid binding stub libraries by absolute path
    non_bindable_shared_objects = ["stubs"]
```

var/spack/repos/builtin/packages/nvhpc/package.py
```diff
@@ -401,3 +401,8 @@ def libs(self):
             libs.append("libnvf")
 
         return find_libraries(libs, root=prefix, recursive=True)
+
+    @property
+    def non_bindable_shared_objects(self):
+        # Avoid binding stub libraries by absolute path
+        return ["stubs"]
```
tgamblin: Could this be a regular property?

var/spack/repos/builtin/packages/openjdk/package.py
```diff
@@ -341,3 +341,10 @@ def setup_dependent_run_environment(self, env, dependent_spec):
             class_paths = find(dependent_spec.prefix, "*.jar")
             classpath = os.pathsep.join(class_paths)
             env.prepend_path("CLASSPATH", classpath)
+
+    @property
+    def non_bindable_shared_objects(self):
+        # Since we provide openjdk as a binary, we can't remove an obsolete glibc
+        # fix that prevents us from modifying the soname of libjvm.so. If we move
+        # to source builds this should be possible.
+        return ["libjvm.so"]
```
tgamblin: Could this be a regular property?

lib/spack/spack/hooks/__init__.py
```diff
@@ -44,11 +41,18 @@ def __init__(self, hook_name):
...
+        relative_names = [
+            "extensions",
+            "licensing",
+            "module_file_generation",
+            "permissions_setters",
+            "sbang",
+            "absolutify_elf_sonames",
+            "write_install_manifest",
+        ]
```
tgamblin: The thing about this block is that the docstring for the file is wrong now -- it says we load hooks from this directory. Also, if people were adding local hooks they'll break.

Could you instead add something to `llnl.util.lang`, like:

```python
def ensure_last(list, *elements):
    """Ensure that ``elements`` are placed at the end of the ``list``.

    Raises ``ValueError`` if any ``elements`` are not already in ``list``.
    """
    for elt in elements:
        list.append(list.pop(list.index(elt)))
```

Then, here:

```python
relative_names = list(llnl.util.lang.list_modules(spack.paths.hooks_path))
lang.ensure_last(
    relative_names,
    "absolutify_elf_sonames",
    "write_install_manifest",
)
```

That way new hooks still get executed as expected, but it's not as nasty as the sort/assert thing above.

lib/spack/spack/package_base.py
```diff
@@ -2794,6 +2794,17 @@ def rpath_args(self):
...
+    @property
+    def non_bindable_shared_objects(self):
+        """List of shared objects that should be replaced with a different library at
+        runtime. Typically includes stub libraries like libcuda.so. When linking
+        against a library listed here, the dependent will only record its soname
+        or filename, not its absolute path, so that the dynamic linker will search
+        for it. Note: accepts both file names and directory names, for example
+        ``["libcuda.so", "stubs"]`` will ensure libcuda.so and all libraries in the
+        stubs directory are not bound by path."""
+        return []
```
tgamblin: One worry for this PR is how many of these we're going to need to add -- e.g., is this needed for SYCL or for HIP as well?  What about OpenMPI?  SLURM?  Flux?

@trws are there any obvious cases that come to mind here off the top of your head?

lib/spack/spack/hooks/absolutify_elf_sonames.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+def get_byte_at(byte_array, idx):
+    if sys.version_info[0] < 3:
+        return ord(byte_array[idx])
+    return byte_array[idx]
+
+
+def is_probably_a_shared_library_elf_file(f):
+    # Read the 32/64 bit class independent part of the header and validate
```
tgamblin: Can most of what is in this file be moved to `spack.util.elf`, and can the hook be kept as a fairly minimal thing that just calls functions from that?  I think this code is going to end up being useful for binary relocation.

tgamblin (changes requested):
Ok, I left some requests for smaller changes, but I have two higher level concerns:

1. How many things are going to need `non_bindable_shared_objects` so that they do not break?  Worried about ROCm, SLURM, Flux, etc. -- are there other things with stubs?
2. I think this should really be a config option -- we already let you pick between `RPATH` and `RUNPATH`, and I suspect the people choosing `RUNPATH` are *not* going to want this feature most of the time -- it'll break their assumption that they can load a module to, e.g., replace an MPI.3. Suggest adding something like:
    ```yaml
    prebind_shared_library_paths = True
    ```
    in `config.yaml`, along with the existing `shared_linking = ["runpath" | "rpath"]` option.
    Or, we could expand the existing option, e.g.:
    ```yaml
    shared_linking:
        paths: "rpath"
        prebind: True
    ```

What do you think?

mboisson:
> @mboisson:
> 
> > The longer answer is that whenever you think you understand the shared library and run-time linkers, you inevitably find out that you don't.
> 
> Is there something specific you see wrong here?

I don't know enough about such things to know what I don't know. I'm just pointing out what some smarter people than me about such things are saying :) That makes me very hesitant to meddle with this sort of things, regardless of whether I think it looks like a good idea (I think it does).

haampie:
Two concerns with this PR:

- `patchelf` still has bugs for out of place patches, right now I'm running into an issue where a dynstr section generated by patchelf exceeds the bounds of a pt_load program header
- Someone mentioned they need `LD_PRELOAD`, which could be problematic when we change sonames (e.g. preload libf.so with soname libf.so, and you want to avoid that /path/to/another/libf.so gets loaded too when its soname is not exactly equal since its a path now) -- ~haven't tried this (yet)~, but I don't think it's necessarily a show stopper.

So, I tried `LD_PRELOAD`: for overriding symbols, it works fine. However, if `libf.so` is for example a C++ library that defines a global with dynamic initialization (constructor of a class), and it has a dependent library that uses its value to do dynamic initialization of _their_ globals, then their could be an issue. With identical sonames, glibc will simply "swap" the libraries. With different sonames, you don't swap libraries, you load them both, and in that case the globals are taken from the original library, not the LD_PRELOAD'ed one. This is due to initialization order. See https://github.com/haampie/loading-order/ for example.

haampie:
I'm not sure if `prebind` is the right term, since it's used interchangeably with `prelink`, which is also a tool that does something else, namely reducing relocations.

But yeah, it's *some* form of binding/fixing/pinning, to I just went with `bind:true/false`. We could also call it `pin: true/false`?

haampie:
@tgamblin this is ready for review again, I've also added documentation (https://spack--31948.org.readthedocs.build/en/31948/config_yaml.html#shared-linking-bind)

scheibelp:
> loaded libraries through dlopen), which largely solves the issue that Spack's rpaths are a heuristic (<prefix>/lib and <prefix>/lib64 might not be where libraries really are...)

Right now, `build_environment` uses `Package.libs.directories` to augment RPATHs, so is not a heuristic: it uses `find_libraries` and could be modified to be more general (the point being, that function should probably be returning the locations of any libraries that we'd want to rpath by default, and it should be finding them for linking); it also can be overridden by packages to say exactly where libraries are (that's generally if it should be returning a subset). 

Also I'm curious: I'd expect that our relocation logic would need to be updated, or is the string replacement it does "generic enough" to cover this also?

haampie:
>  it uses find_libraries and could be modified to be more general

It's still a heuristic; spack doesn't exactly know what to look for, only the build system knows what it needs, and only the linker knows what's linked.

> Also I'm curious: I'd expect that our relocation logic would need to be updated, or is the string replacement it does "generic enough" to cover this also?

It doesn't, it's covered by generic binary "text" replacement

scheibelp:
>> it uses find_libraries and could be modified to be more general

> It's still a heuristic; spack doesn't exactly know what to look for, only the build system knows what it needs, and only the linker knows what's linked.

I see that if the build system found all the needed libraries, then at that point the needed libraries are mentioned in the ELF data (assuming that there are not equivalently-named libraries with different directories in the dependencies) and there is no need for per-package logic to locate those libraries. That being said, sometimes we use `.libs` to provide hints to the build system (when it cannot locate specific dependency libraries on its own), and in that case, `.libs` would have to be knowledgeable.

Generally though this puts less pressure on `.libs` to be correct (i.e. whenever the build system is able to properly locate libraries within a dependency prefix).

haampie:
I've moved elf parsing to #33628, so this PR can focus just on changing sonames.

haampie:
Rebased now that #33628 is merged

etc/spack/defaults/config.yaml
```diff
@@ -187,10 +187,20 @@ config:
   package_lock_timeout: null
 
 
-  # Control whether Spack embeds RPATH or RUNPATH attributes in ELF binaries.
-  # Has no effect on macOS. DO NOT MIX these within the same install tree.
-  # See the Spack documentation for details.
-  shared_linking: 'rpath'
+  # Control how shared libraries are located at runtime on Linux. See the
+  # the Spack documentation for details.
+  shared_linking:
```
tgamblin, line 192: ensure that `shared_linking: <string>` still works.

lib/spack/spack/build_environment.py
```diff
@@ -317,7 +317,7 @@ def set_compiler_environment_variables(pkg, env):
     env.set("SPACK_LINKER_ARG", compiler.linker_arg)
 
     # Check whether we want to force RPATH or RUNPATH
-    if spack.config.get("config:shared_linking") == "rpath":
+    if spack.config.get("config:shared_linking:type") == "rpath":
```
tgamblin, line 320: This still needs to support the old format for backward compatibility.  If you get `config:shared_linking` and it's a string, assume it's the `type`.

We could warn and recommend updating to the new config format; ideally `spack config update` would allow the user to automatically fix it if we're going to warn about it.  I think it's simpler here to just continue accepting a string.  If you want to do the update thing, look at the various `update()` methods in `spack.schema.*` modules for how to handle it.

lib/spack/spack/hooks/__init__.py
tgamblin: Is this comment still relevant?  I think it's not since you now use `ensure_last`

lib/spack/spack/schema/config.py
```diff
@@ -20,7 +20,18 @@
         "type": "object",
         "default": {},
         "properties": {
-            "shared_linking": {"type": "string", "enum": ["rpath", "runpath"]},
+            "shared_linking": {
+                "anyOf": [
```
tgamblin, line 24: looks like you already support the string in the schema; it just isn't handled in the code?

lib/spack/spack/test/build_environment.py
```diff
@@ -465,7 +465,7 @@ def test_setting_dtags_based_on_config(config_setting, expected_flag, config, mo
     pkg = s.package
 
     env = EnvironmentModifications()
-    with spack.config.override("config:shared_linking", config_setting):
+    with spack.config.override("config:shared_linking", {"type": config_setting, "bind": False}):
```
tgamblin, line 468: test should handle both cases (string and type/bind dict)

tgamblin (changes requested):
LGTM -- minor backward compatibility requests.

Also, is it possible to get more coverage on this?

lib/spack/spack/schema/config.py
haampie: it is, `update` is always called after reading config

etc/spack/defaults/config.yaml
haampie: I've added a test for that

lib/spack/spack/build_environment.py
haampie: Was already the case

lib/spack/spack/test/build_environment.py
haampie: This is updating old syntax internally, there's a test elsewhere that ensures old syntax `override(config:shared_linking:runpath)` works too, so i won't change this here.

tgamblin (approved):

lib/spack/spack/schema/config.py
tgamblin: Ok awesome — thanks. I missed that you were already using update.

haampie:
CI is failing with the known `no binary for ...` issue. Might need `[Update branch]` :thinking:

PR #31953, opened 2022-08-05
Add netcdf-c 4.9.0 and netcdf-fortran 4.6.0

kgerheiser:
With v4.9.0 netcdf-c introduces zstandard compression option which is added as a variant.

@edwardhartnett @climbfuji FYI

### DIFF
--- a/var/spack/repos/builtin/packages/netcdf-c/package.py
+++ b/var/spack/repos/builtin/packages/netcdf-c/package.py
@@ -20,6 +20,7 @@ class NetcdfC(AutotoolsPackage):
     maintainers = ["skosukhin", "WardF"]
 
     version("main", branch="main")
+    version("4.9.0", sha256="9f4cb864f3ab54adb75409984c6202323d2fc66c003e5308f3cdf224ed41c0a6")
     version("4.8.1", sha256="bc018cc30d5da402622bf76462480664c6668b55eb16ba205a0dfb8647161dd0")
     version("4.8.0", sha256="aff58f02b1c3e91dc68f989746f652fe51ff39e6270764e484920cb8db5ad092")
     version("4.7.4", sha256="99930ad7b3c4c1a8e8831fb061cb02b2170fc8e5ccaeda733bd99c3b9d31666b")
@@ -65,7 +66,7 @@ class NetcdfC(AutotoolsPackage):
     patch("4.7.3-spectrum-mpi-pnetcdf-detect.patch", when="@4.7.3:4.7.4 +parallel-netcdf")
 
     # See https://github.com/Unidata/netcdf-c/pull/2293
-    patch("4.8.1-no-strict-aliasing-config.patch", when="@4.8.1:")
+    patch("4.8.1-no-strict-aliasing-config.patch", when="@4.8.1")
 
     variant("mpi", default=True, description="Enable parallel I/O for netcdf-4")
     variant("parallel-netcdf", default=False, description="Enable parallel I/O for classic files")
@@ -75,6 +76,7 @@ class NetcdfC(AutotoolsPackage):
     variant("dap", default=False, description="Enable DAP support")
     variant("jna", default=False, description="Enable JNA support")
     variant("fsync", default=False, description="Enable fsync support")
+    variant("zstd", default=False, description="Enable ZStandard compression", when="@4.9.0:")
 
     # It's unclear if cdmremote can be enabled if '--enable-netcdf-4' is passed
     # to the configure script. Since netcdf-4 support is mandatory we comment
@@ -122,6 +124,8 @@ class NetcdfC(AutotoolsPackage):
     # NetCDF 4.4.0 and prior have compatibility issues with HDF5 1.10 and later
     # https://github.com/Unidata/netcdf-c/issues/250
     depends_on("hdf5@:1.8", when="@:4.4.0")
+
+    depends_on("zstd", when="@4.9.0:")
 
     # The feature was introduced in version 4.1.2
     # and was removed in version 4.4.0
@@ -234,6 +238,16 @@ class NetcdfC(AutotoolsPackage):
             if "+external-xdr" in hdf4 and hdf4["rpc"].name != "libc":
                 libs.append(hdf4["rpc"].libs.link_flags)
 
+        if "+zstd" in self.spec:
+            zstd = self.spec["zstd"]
+            cppflags.append(zstd.headers.cpp_flags)
+            ldflags.append(zstd.libs.search_flags)
+            config_args.append("--with-plugin-dir={}".format(self.prefix.plugins))
+        elif "~zstd" in self.spec:
+            # Prevent linking to system zstd.
+            # There is no explicit option to disable zstd.
+            config_args.append("ac_cv_lib_zstd_ZSTD_compress=no")
+
         # Fortran support
         # In version 4.2+, NetCDF-C and NetCDF-Fortran have split.
         # Use the netcdf-fortran package to install Fortran support.
@@ -244,6 +258,13 @@ class NetcdfC(AutotoolsPackage):
         config_args.append("LIBS=" + " ".join(libs))
 
         return config_args
+
+    def setup_run_environment(self, env):
+        if "+zstd" in self.spec:
+            env.append_path("HDF5_PLUGIN_PATH", self.prefix.plugins)
+
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        self.setup_run_environment(env)
 
     def check(self):
         # h5_test fails when run in parallel
--- a/var/spack/repos/builtin/packages/netcdf-c/package.py
+++ b/var/spack/repos/builtin/packages/netcdf-c/package.py
@@ -125,7 +125,7 @@ class NetcdfC(AutotoolsPackage):
     # https://github.com/Unidata/netcdf-c/issues/250
     depends_on("hdf5@:1.8", when="@:4.4.0")
 
-    depends_on("zstd", when="@4.9.0:")
+    depends_on("zstd", when="+zstd")
 
     # The feature was introduced in version 4.1.2
     # and was removed in version 4.4.0
--- a/var/spack/repos/builtin/packages/netcdf-fortran/package.py
+++ b/var/spack/repos/builtin/packages/netcdf-fortran/package.py
@@ -21,6 +21,7 @@ class NetcdfFortran(AutotoolsPackage):
 
     maintainers = ["skosukhin", "WardF"]
 
+    version("4.6.0", sha256="198bff6534cc85a121adc9e12f1c4bc53406c403bda331775a1291509e7b2f23")
     version("4.5.4", sha256="0a19b26a2b6e29fab5d29d7d7e08c24e87712d09a5cafeea90e16e0a2ab86b81")
     version("4.5.3", sha256="123a5c6184336891e62cf2936b9f2d1c54e8dee299cfd9d2c1a1eb05dd668a74")
     version("4.5.2", sha256="b959937d7d9045184e9d2040a915d94a7f4d0185f4a9dceb8f08c94b0c3304aa")


var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -75,6 +76,7 @@ class NetcdfC(AutotoolsPackage):
     variant("dap", default=False, description="Enable DAP support")
     variant("jna", default=False, description="Enable JNA support")
     variant("fsync", default=False, description="Enable fsync support")
+    variant("zstd", default=False, description="Enable ZStandard compression")
```
haampie:
```suggestion
    variant("zstd", default=False, description="Enable ZStandard compression", when="@4.9.0:")
```

kgerheiser: Nice, I didn't think variant could do `when=`.

skosukhin: The corresponding conflict should be removed, I guess.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -123,13 +125,17 @@ class NetcdfC(AutotoolsPackage):
...
+
     # The feature was introduced in version 4.1.2
     # and was removed in version 4.4.0
     # conflicts('+cdmremote', when='@:4.1.1,4.4:')
 
     # The features were introduced in version 4.1.0
     conflicts("+parallel-netcdf", when="@:4.0")
     conflicts("+hdf4", when="@:4.0")
+    # zstd compression was introdcued in 4.9.0
+    conflicts("+zstd", when="@:4.8")
```
haampie:
```suggestion
```

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -76,7 +76,7 @@ class NetcdfC(AutotoolsPackage):
     variant("dap", default=False, description="Enable DAP support")
     variant("jna", default=False, description="Enable JNA support")
     variant("fsync", default=False, description="Enable fsync support")
-    variant("zstd", default=False, description="Enable ZStandard compression", when="@:4.8")
+    variant("zstd", default=False, description="Enable ZStandard compression", when="@:4.9.0")
```
haampie: the other way around with the `:` ;p it's from 4.9.0 onwards that this variant should be available

var/spack/repos/builtin/packages/netcdf-c/package.py
skosukhin: @kgerheiser please, keep the authorship of the suggested changes

var/spack/repos/builtin/packages/netcdf-c/package.py
skosukhin: @haampie you mean `@4.9.0:`, I guess

haampie:
Although support for zstd is optional, the configure script is written in such a way that it's used/enabled when found, correct? I'm getting
```
checking for ZSTD_compress in -lzstd... yes
checking for library containing ZSTD_compress... -lzstd
checking whether libzstd library is available... yes
```
and trying to disable it is futile
```
./configure --disable-zstd --without-zstd
configure: WARNING: unrecognized options: --disable-zstd, --without-zstd
```
In that case it might be better to just make it a fixed dependency, to avoid that it links against system libraries.

kgerheiser:
> Although support for zstd is optional, the configure script is written in such a way that it's used when found, correct? 

Yes

> In that case it might be better to just make it a fixed dependency, to avoid that it links against system libraries.

That's a good point, I think that makes sense.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -234,6 +240,12 @@ def configure_args(self):
             if "+external-xdr" in hdf4 and hdf4["rpc"].name != "libc":
                 libs.append(hdf4["rpc"].libs.link_flags)
 
+        if "+zstd" in self.spec:
+            zstd = self.spec["zstd"]
+            config_args.append("--with-plugin-dir={}".format(self.prefix.plugins))
+            cflags.append("-I" + zstd.prefix.include)
+            ldflags.append("-L" + zstd.prefix.lib)
```
skosukhin: 1. `-I` should go to `cppflags`
2. Is it possible to use `headers` and `libs`? For example:
    ```python
    cppflags.append(zstd.headers.cpp_flags)
    ldflags.append(zstd.libs.search_flags)
    libs.append(zstd.libs.link_flags)
    ```

kgerheiser: Yes, but zstd does not currently define those properties.

kgerheiser: Or does it automatically work?

skosukhin:
@haampie 
> In that case it might be better to just make it a fixed dependency, to avoid that it links against system libraries.

Do you mean not depending on the variant? If yes, I strongly object. I don't want to have `zstd` be built by default. We need a patch for the configure script then.

var/spack/repos/builtin/packages/netcdf-c/package.py
skosukhin: There are default implementations for the properties. If they work correctly, please, use them. Otherwise, leave a `TODO` comment as for other such cases.

haampie: You can drop `libs.append(zstd.libs.link_flags)` though, the build system tests for `-lzstd`.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -234,6 +237,12 @@ def configure_args(self):
             if "+external-xdr" in hdf4 and hdf4["rpc"].name != "libc":
                 libs.append(hdf4["rpc"].libs.link_flags)
 
+        if "zstd" in self.spec:
```
skosukhin:
```suggestion
        if "+zstd" in self.spec:
```

var/spack/repos/builtin/packages/netcdf-c/package.py
kgerheiser: The default implementations work

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -245,6 +254,10 @@ def configure_args(self):
 
         return config_args
 
+    def setup_run_environment(self, env):
+        if self.spec.satisfies("+zstd"):
+            env.append_path("HDF5_PLUGIN_PATH", self.prefix.plugins)
+
```
skosukhin, line 265: 1. Can `if self.spec.satisfies("+zstd")` be `if "+zstd" in self.spec`? They are the same but I would like to stick to either one or another.
2. @kgerheiser do you know whether I need to set `HDF5_PLUGIN_PATH` when the library is built with `+zstd` but I work with files that are NOT compressed with it?
3. @haampie do you think we should set the environment variable in `setup_dependent_build_environment` too? It's always confusing for me.

kgerheiser: 1. Yes
2. No, it's not required. If you open a non-zstd compressed file the filter will not be loaded. It's all done dynamically at runtime and unless you try to read/write a zstd compressed file those code paths will not be executed.
3. Was wondering this too, though it would be the `_run_` environment.  I experimented a bit and built NCO and HDF5_PLUGIN_PATH was correctly set using a `spack load nco`.

skosukhin: Thank you. Regarding `3.`, the variable might be required when a dependent package runs a build-time check of whether `zstd` is supported. Without the variable, the check might fail. There were discussions that `setup_run_environment` should be run also when building dependencies but I am not aware of the recent developments in Spack regarding this.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -123,6 +124,8 @@ class NetcdfC(AutotoolsPackage):
     # https://github.com/Unidata/netcdf-c/issues/250
     depends_on("hdf5@:1.8", when="@:4.4.0")
 
+    depends_on("zstd", when="@4.9.0:")
```
skosukhin: As I have already mentioned, I would like to have `zstd` as an optional dependency. It might be the case that we don't even have to patch the configure script for that and just simply set a cache variable on the command line like we do for [`curl`](https://github.com/spack/spack/blob/42de3ff968f8bc32221ced071ba3633bca670bb4/var/spack/repos/builtin/packages/netcdf-c/package.py#L187).

skosukhin: And before I forget. There is a caveat when declaring variants with `when`: if `+var1 in self.spec` is `False`, it does not always mean that `~var1 in self.spec` is `True` because versions that do not fall into `when` of `var1` have `False` in both cases. So, it might be the case that a clean solution for `zstd` would be something like:
```python
if "+zstd" in self.spec:
    # append to cppflags and ldflags
elif "~zstd" in self.spec:
    # prevent linking to a system installation of zstd
```

var/spack/repos/builtin/packages/netcdf-c/package.py
kgerheiser: Added. I could mabye see that being necessary, though I think there are better ways to check such as checking the NetCDF header for `HAVE_ZSTD`.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -245,6 +256,14 @@ def configure_args(self):
 
         return config_args
 
+    def setup_run_environment(self, env):
+        if "+zstd" in self.spec:
+            env.append_path("HDF5_PLUGIN_PATH", self.prefix.plugins)
+
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        if "+zstd" in self.spec:
+            env.append_path("HDF5_PLUGIN_PATH", self.prefix.plugins)
```
skosukhin:
```suggestion
        self.setup_run_environment(env)
```

var/spack/repos/builtin/packages/netcdf-c/package.py
kgerheiser: Done, setting the variable `ac_cv_lib_zstd_ZSTD_compress=no` worked.

haampie:
I reported the issue upstream, so that we can hopefully get rid of this `ac_cv_lib_zstd_ZSTD_compress=no` business

skosukhin:
LGTM

climbfuji:
@skosukhin @haampie Can we move this PR forward? Note that @kgerheiser is currently going through a job transition, I will try to help get this across the finish line if more work is needed. Thanks!

climbfuji:
@WardF @haampie @skosukhin Sorry to bug you, but can we review and merge this PR some time soon? Thanks.

climbfuji (approved):

eugeneswalker (approved):

haampie (approved):

alalazo (approved):

jedwards4b (approved):

climbfuji:
Hmm, this PR now has conflicts that must be resolved, but I do not have write access to @kgerheiser's fork/branch.

climbfuji:
@skosukhin Can you help resolving the conflicts, please, and pushing this PR across the finish line?

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -123,6 +125,8 @@ class NetcdfC(AutotoolsPackage):
     # https://github.com/Unidata/netcdf-c/issues/250
     depends_on("hdf5@:1.8", when="@:4.4.0")
```
edwardhartnett, line 126: We should be using hdf5-1.12.1.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -75,6 +76,7 @@ class NetcdfC(AutotoolsPackage):
     variant("dap", default=False, description="Enable DAP support")
     variant("jna", default=False, description="Enable JNA support")
     variant("fsync", default=False, description="Enable fsync support")
+    variant("zstd", default=False, description="Enable ZStandard compression", when="@4.9.0:")
```
edwardhartnett, line 79: I think it would be better to enable zstd by default. It should be supported everywhere.

var/spack/repos/builtin/packages/netcdf-c/package.py
climbfuji: This is just saying that if you use netcdf-c <= 4.4.0 you have to use hdf5 <= 1.8 because of the issue described here: https://github.com/Unidata/netcdf-c/issues/250

var/spack/repos/builtin/packages/netcdf-c/package.py
climbfuji: "should be" isn't necessarily "is" ... who knows if `zstd` breaks on some of the exotic systems we use like to old Cray machines ... just saying. The much more basic problem is that I cannot update this PR, since it is from @kgerheiser's fork/branch. We can issue follow-up PRs but the only person who can update this PR is Kyle.

edwardhartnett: We investigated this issue and zstandard is available and working on all platforms we need to support.

For those working on old platforms, they can configure zstd off with an option. But I want to hear that such a platform actually exists, rather than disable this important functionality by default. HPC users are exactly the kinds of users that need to use zstd, so it would be built in by default. And it is easy to do so.

The reason to have it on by default is that we are going to start using it to distribute data. Once we do, everyone who built netCDF without zstd will contact us and complain that they can't read the data. To minimize user disruption and user support, make zstd the default on HPC systems.

var/spack/repos/builtin/packages/netcdf-c/package.py
edwardhartnett: Ah, ok thanks!

var/spack/repos/builtin/packages/netcdf-c/package.py
climbfuji: Ok, that's fine with me. Now we only need to find a way to update the PR!

edwardhartnett:
@kgerheiser is there a way to transfer ownership of this PR?

climbfuji:
> @kgerheiser is there a way to transfer ownership of this PR?

What I'll do is check out Kyle's branch, add the zstd change on top of it, then create a follow-up PR. We can then merge this PR as-is and immediately afterwards the zstd PR.

climbfuji:
> > @kgerheiser is there a way to transfer ownership of this PR?
> 
> What I'll do is check out Kyle's branch, add the zstd change on top of it, then create a follow-up PR. We can then merge this PR as-is and immediately afterwards the zstd PR.

Actually, we can't do that because this PR has conflicts. I created PR https://github.com/spack/spack/pull/33514 which includes everything from this (Kyle's original) PR with the original commits (I don't want to claim that I made the code changes), and in addition resolves the conflicts and sets the `zstd` variant to `.true.`.

I would therefore like to close this PR in favor of https://github.com/spack/spack/pull/33514.

PR #29443, opened 2022-03-10
generate environment views in single pass

haampie:
environment views: single pass view generation

Reduces the number of stat calls to a bare minimum:
- Single pass over src prefixes
- Handle projection clashes in memory

Symlinked directories in the src prefixes are now conditionally
transformed into directories with symlinks in the dst dir. Notably
`intel-mkl`, `cuda` and `qt` has top-level symlinked directories that
previously resulted in empty directories in the view. We now avoid
cycles and possible exponential blowup by only expanding symlinks that:
- point to dirs deeper in the folder structure;
- are a fixed depth of 2.

PR #31958, opened 2022-08-05
copy view does not work because tries to copy directories

vsoch:
This fix ensures that we do a check to not read a directory as a binary file, and use shutil.copytree to copy a directory in full. The flux install to a view was broken (working on locally) and these two fixes resolved it (at least to finish the install / create the view).

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

### DIFF
--- a/lib/spack/spack/filesystem_view.py
+++ b/lib/spack/spack/filesystem_view.py
@@ -68,7 +68,7 @@ def view_copy(src, dst, view, spec=None):
     Use spec and view to generate relocations
     """
     if os.path.isdir(src):
-        shutil.copytree(src, dst)
+        shutil.copytree(src, dst, symlink=True)
     else:
         shutil.copy2(src, dst)
 
--- a/lib/spack/spack/filesystem_view.py
+++ b/lib/spack/spack/filesystem_view.py
@@ -67,7 +67,11 @@ def view_copy(src, dst, view, spec=None):
 
     Use spec and view to generate relocations
     """
-    shutil.copy2(src, dst)
+    if os.path.isdir(src):
+        shutil.copytree(src, dst)
+    else:
+        shutil.copy2(src, dst)
+
     if spec and not spec.external:
         # Not metadata, we have to relocate it
 
--- a/lib/spack/spack/relocate.py
+++ b/lib/spack/spack/relocate.py
@@ -775,6 +775,9 @@ def relocate_text(files, prefixes, concurrency=32):
 
     args = []
     for filename in files:
+        # Passing a directory to _replace_prefix_text will still throw error
+        if os.path.isdir(filename):
+            continue
         args.append((filename, compiled_prefixes))
 
     tp = multiprocessing.pool.ThreadPool(processes=concurrency)
--- a/lib/spack/spack/relocate.py
+++ b/lib/spack/spack/relocate.py
@@ -776,7 +776,7 @@ def relocate_text(files, prefixes, concurrency=32):
     args = []
     for filename in files:
         # Passing a directory to _replace_prefix_text will still throw error
-        if os.path.isdir(filename):
+        if os.path.isdir(filename) and os.path.islink(filename):
             continue
         args.append((filename, compiled_prefixes))
 


trws (dismissed):
LGTM

haampie:
I don't really use binary caches / relocation (for reason @vsoch will understand :laughing:) so I'm not all too familiar with the code, but from the docstring `relocate_text(files, prefixes, concurrency=32)`:

> Relocate text file from the original installation prefix to the new prefix. 

Makes me think that the real issue happens earlier? `relocate_text` should not receive dirs.

Similarly: `def view_copy(src, dst, view, spec=None)`:

> Copy a file from src to dst.

should not receive a dir in the first place

haampie (changes requested):
The issue as always is symlinked directories :tada:.

lib/spack/spack/relocate.py
```diff
@@ -775,6 +775,9 @@ def relocate_text(files, prefixes, concurrency=32):
 
     args = []
     for filename in files:
+        # Passing a directory to _replace_prefix_text will still throw error
+        if os.path.isdir(filename):
+            continue
```
haampie:
```suggestion
```

lib/spack/spack/filesystem_view.py
```diff
@@ -67,7 +67,11 @@ def view_copy(src, dst, view, spec=None):
 
     Use spec and view to generate relocations
     """
-    shutil.copy2(src, dst)
+    if os.path.isdir(src):
+        shutil.copytree(src, dst)
+    else:
+        shutil.copy2(src, dst)
+
```
haampie: `isdir` is true if and only if `src` is a symlink to a dir. You should be checking for `islink` instead and create a corresponding symlink in `dst`.

Unfortunately this is complicated by two things:
1) Python's sad filesystem API does not have a Python 2 compatible `shutil.copy2(src, dst, follow_symlinks=False)`.
2) Symlinks can be absolute paths too, and those need to be relocated as well.

There seems to be some function `spack.relocate.relocate_links` that could potentially be helpful.

lib/spack/spack/relocate.py
vsoch: How can you just remove this? That won't fix the bug.

vsoch:
If you don't like my solution could you please make a suggested change that can be used instead?

haampie:
As I said, I'm not very familiar with this code, so I just pointed out the issue and made some high level suggestions on how to fix it https://github.com/spack/spack/pull/31958#discussion_r939515918

vsoch:
Could we ping who wrote it? The issue is that your suggestion unfixes it. I’m not familiar enough either to suggest something different but I can confirm what I did here seemed to work.

haampie:
My suggestion does not "unfix" it.

Your pr may fix the particular issue you're facing, but it's incorrect in general in multiple ways.

Do you want to contribute the proper fix?

vsoch:
Yes obviously but I don’t  know what that is.

trws:
Ok, looks like maybe some friction here. @haampie I pinged you because I remembered you working on the view copy code, would you mind pointing out where that class and stuff ended up?  If we can re-use that I imagine it would take care of this, since I remember you dealing with almost exactly this problem before. As far as I can tell though, you're saying we should be checking islink rather than isdir and linking rather than copying through, is that what you meant? If so the change might be almost identical just special casing links rather than directories right?

haampie:
Yeah, the code that deals with merging prefixes uses a visitor where you're forced to think about symlinked dirs vs ordinary dirs. It sets up the directory structure and tries to be smart about changing symlinked dirs into normal dirs if that helps reducing merge conflicts.*

It also generates a list of files that must be linked into the view. This list of files may contain absolute and relative symlinks to dirs and files, or can be dangling.

Some link function (symlink/hardlink/copy) is applied to that list, and it should deal with the case of symlinks. For symlink/hardlink this is not an issue, since symlinking or hardlinking a symlink is fine. It seems however that for copy (with relocation) this was not implemented, even though there is already code for relocating symlinks elsewhere in Spack.

[\* actually, in the case of "copy" it would still be good to replace symlinked dirs with ordinary dirs to reduce merge conflicts, but the dir entries should then optimally be symlinks again, not copies, to reduce redundancy... but that's a separate issue.]

haampie:
@vsoch could it be that you somehow missed this comment earlier? https://github.com/spack/spack/pull/31958#discussion_r939515918

lib/spack/spack/filesystem_view.py
trws: How would shutil.copytree symlink=True fare for that case? I can't recall how copy2 differs.

vsoch:
Not in a completely conscious state - it was 5am here and I had only "gone to bed" at about 3:40am :)

> isdir is true if and only if src is a symlink to a dir. You should be checking for islink instead and create a corresponding symlink in dst.

Can you show me explicitly what you mean by this? Do you have code from https://github.com/spack/spack/pull/29443 that could help us? Unfortunately I'm not familiar with this view code either - my suggestion was to find the person that is, and combine that with what you worked on to get to the best solution. My changes were very prescriptive - "here is the exact lines for the problem - I'll add checks there so they don't happen" and I don't have the full picture for how this works. It looks like the functionality was added here: https://github.com/spack/spack/commit/3347ef2de4e08374750eb68f750800c1854d595f @becker33 can you help here to put the pieces together about what we should do? My needs are basically without the checks I added, the flux view creation crashes and fails - first trying to run the copy2 on a directory (fails) and then the multiprocess worker also fails running that function on a non file. So that's what I need to fix, and I'm agnostic how we get to that (and i don't know the best way to go about this). Also, see this comment from @haampie : 

> There seems to be some function spack.relocate.relocate_links that could potentially be helpful.

vsoch:
ping a ping ping ping ring a ding ding ding :bell:

vsoch:
Anyone around? We need to fix this.

lib/spack/spack/filesystem_view.py
```diff
@@ -67,7 +67,11 @@ def view_copy(src, dst, view, spec=None):
 
     Use spec and view to generate relocations
     """
-    shutil.copy2(src, dst)
+    if os.path.isdir(src):
+        shutil.copytree(src, dst, symlink=True)
```
trws, line 71: I think this needs to by `symlinks` rather than symlink? Oddly I don't see it failing, which makes me wonder, but that's how it's documented.

```suggestion
        shutil.copytree(src, dst, symlinks=True)
```

trws, line 71: More generally, I think @haampie is suggesting changing this to something like:

```suggestion
    if os.path.islink(src):
        os.symlink(src, dst)
```

If that works, it may take care of the issue without needing to traverse the directory.

lib/spack/spack/relocate.py
trws:
```suggestion
        # Passing a directory to _replace_prefix_text will still throw error
        if os.path.isdir(filename) and os.path.islink(filename):
            continue
```

vsoch: Thanks @trws !

lib/spack/spack/filesystem_view.py
haampie: It's not enough for absolute symlinks into the prefix dir, they have to be relocated

trws: If they're into the prefix dir, they will have been relocated by being crawled through the non-symlink directory anyway right? Ah, but the target will be wrong because the src is still in the prefix, is that what you mean?

haampie:
Closing this in favor of #32306

PR #31965, opened 2022-08-06
e4s oneapi ci: unify when possible

eugeneswalker:
Unify the E4S OneAPI stack when possible. 

FYI @wspear


eugeneswalker:
Looks like everything built OK! We should merge this!

alalazo:
For the record the `step_script` is taking 3 mins with 12 processors to concretize on develop, and less than 2 mins with 1 processor to concretize here.

alalazo (approved):

PR #31966, opened 2022-08-06
mercury: add version 2.2.0

soumagne:
add psm, psm2 and hwloc variants


carns (dismissed):
Looks great to me.  I appreciate the inclusion of the new psm and hwloc variants.

soumagne:
[data-vis-sdk-pr-build](https://gitlab.spack.io/spack/spack/-/pipelines/150248) failed in https://gitlab.spack.io/spack/spack/-/jobs/2955038#L1721
Because mercury 2.2.0 is now more strict on the libraries that it includes through pkg-config, I suspect unifyfs will have to be either updated to link against the mercury NA library by using the na.pc file or it should remove its use of the `NA_Set_log_level()` call, which should really not be needed.

alalazo (changes requested):
Pipelines are required to pass for the PR to be merged

carns:
The Unify team is aware of the build problem and looking at how to best handle it.  We can re-kick the CI action once we get an all clear from them.

bernhardkaindl:
@alalazo I test-built this PR, and triggered CI, could you approve? That the build didn't pass is fixed now

alalazo (approved):

PR #31967, opened 2022-08-06
py-macholib: add v1.16

sethrj:
macOS 10.15 and higher have new LC load commands. See #28344.


var/spack/repos/builtin/packages/py-macholib/package.py
```diff
@@ -11,6 +11,7 @@ class PyMacholib(PythonPackage):
 
     pypi = "macholib/macholib-1.11.tar.gz"
 
+    version("1.16", "001bf281279b986a66d7821790d734e61150d52f40c080899df8fefae056e9f7")
     version("1.11", "c4180ffc6f909bf8db6cd81cff4b6f601d575568f4d5dee148c830e9851eb9db")
 
     depends_on("py-setuptools", type=("build", "run"))
```
adamjstewart, line 17: altgraph 0.15+ is required now

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-macholib/package.py
sethrj: Looks like 0.15 is required even for 1.11. I'll update the dependency and check for any new changes.

adamjstewart (approved):

PR #31970, opened 2022-08-06
gxsview: gcc11 compilation rule for std::numeric_limits

cessenat:
When compiling with GNU gcc 11, typically with spack:

$ spack install gxsview %gcc@11.2.0 ^vtk@9.0.3

I get the following error
```
     183    In file included from ../core/fielddata/fieldcolordata.cpp:12:
  >> 184    ../core/fielddata/fieldcolordata.hpp:35:53: error: 'numeric_limits' is not a member of 'std'
     185       35 |     static constexpr double SMALL_FIELD_DATA = std::numeric_limits<double>::min()*10;
     186 | ^~~~~~~~~~~~~~

```

From what I read in
https://stackoverflow.com/questions/71296302/numeric-limits-is-not-a-member-of-std
it seems that putting
`#include <limits>`
in file `core/fielddata/fieldcolordata.cpp` corrects the issue.

This is the aim of the provided patch.

### DIFF
--- a/var/spack/repos/builtin/packages/gxsview/package.py
+++ b/var/spack/repos/builtin/packages/gxsview/package.py
@@ -29,7 +29,7 @@ class Gxsview(QMakePackage):
 
     patch("vtk9.patch", when="^vtk@9:")
     # gcc11 compilation rule for std::numeric_limits,
-    # avoid ‘numeric_limits’ is not a member of ‘std’
+    # avoid "numeric_limits" is not a member of "std"
     patch("gcc11.patch", when="@2021.07.01 %gcc@11:")
 
     build_directory = "gui"
--- a/var/spack/repos/builtin/packages/gxsview/package.py
+++ b/var/spack/repos/builtin/packages/gxsview/package.py
@@ -26,9 +26,10 @@ class Gxsview(QMakePackage):
     depends_on("qt@5.14.0:+opengl+gui")
     depends_on("vtk@8.0:+qt+opengl2")  # +mpi+python are optional
     conflicts("%gcc@:7.2.0", msg="Requires C++17 compiler support")  # need C++17 standard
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/gxsview/package.py
```diff
@@ -26,6 +26,7 @@ class Gxsview(QMakePackage):
     depends_on("qt@5.14.0:+opengl+gui")
     depends_on("vtk@8.0:+qt+opengl2")  # +mpi+python are optional
     conflicts("%gcc@:7.2.0", msg="Requires C++17 compiler support")  # need C++17 standard
+    patch("gcc11.patch", when="@2021.07.01 %gcc@11:")
```
alalazo: Can you add a brief comment describing the patch above the directive?

alalazo (changes requested):

var/spack/repos/builtin/packages/gxsview/package.py
cessenat: Yes, done.

var/spack/repos/builtin/packages/gxsview/package.py
```diff
@@ -28,6 +28,9 @@ class Gxsview(QMakePackage):
     conflicts("%gcc@:7.2.0", msg="Requires C++17 compiler support")  # need C++17 standard
 
     patch("vtk9.patch", when="^vtk@9:")
+    # gcc11 compilation rule for std::numeric_limits,
+    # avoid ‘numeric_limits’ is not a member of ‘std’
```
alalazo:
```suggestion
    # avoid "numeric_limits" is not a member of "std"
```

cessenat:
OK @alalazo since you made the improvement yourself I think there is nothing left to do.
Did I mis something or can you uncheck the requested change ?

alalazo (approved):

PR #31975, opened 2022-08-07
Adding package request workflow

vsoch:
This workflow will allow external package maintainers to automate opening pull requests with package updates. Basically, they can maintain their package in a remote repository (that also will get updates from upstream here) and when changes are found, an automation process generates them a link to click and open to create an issue with expected metadata. The workflow here parses newly opened issue, checks for the signature of the request, and then parses the metadata and in the workflow retrieves the updates from the remote and opens a pull request to spack. This is part of sciworks/spack-updater if you are curious to look - I have tested on my own spack repos and am now looking to add here to implement for flux maintainers for their packages.

Example:

Local (package.py) repository to update spack:

 1. We start with merging a successful build, and that opens [this issue](https://github.com/sciworks/spack-updater/issues/19) with a link for you to click.
 2. [This issue was opened](https://github.com/researchapps/spack/issues/2) by clicking the above.
 3. [It triggered this workflow](https://github.com/researchapps/spack/runs/7708080894?check_suite_focus=true) (using the workflow here)
 4. [Which opened this pull request](https://github.com/researchapps/spack/pull/4)

And although it's not relevant to adding anything here, we also do the opposite - updating spack from here. That was done with example m4, and it won't normally have that many commits (that was because I was testing via  PR). https://github.com/sciworks/spack-updater/pull/34/files

Thanks!

Signed-off-by: vsoch <vsoch@users.noreply.github.com>


trws:
@tgamblin, @becker33, if you get a chance to weigh in on this it would be appreciated.  It's the spack-side half of something @vsoch has been setting up for Flux, and potentially other projects later, to help us keep our spack packages tested (by maintaining them in our own repositories) and up to date in spack by making it easy to propagate the updates back.

vsoch:
@alalazo is this something you can review?

vsoch:
I'm going to stop trying to update with develop, which is a constantly moving target. This PR just cleanly adds a new workflow so changes to develop shouldn't alter it.

.github/workflows/update-package-request.yaml
```diff
@@ -0,0 +1,102 @@
...
+      env:
+        title: ${{ github.event.issue.title }}
+        number: ${{ github.event.issue.number }}
+      run: |  
+        printf "Issue title: ${title}\n"
+        printf "Issue number: ${number}\n"
+        echo "${{ github.event.issue.body }}" > new-issue.txt
+        cat new-issue.txt
+        # Script to parse the issue, sets envars needed for subsequent workflows
+        wget https://raw.githubusercontent.com/sciworks/spack-updater/main/parse_issue.py
```
becker33, line 26: Is there any reason we can't bring this script into the repo directly? I don't like having our CI pull a moving-target from github, that will be unnecessarily brittle.

becker33 (changes requested):
Is the opening of an issue a technically required intermediate step, or is that a workflow decision to create an issue to go along with the PR?

.github/workflows/update-package-request.yaml
tgamblin: I think the way to do this is to make the script an action and to pin it at a commit, as we do for all other external actions.

vsoch:
Update for all - @trws @tgamblin and I just had a discussion, going to try to tweak this a bit so it doesn't require an addition to spack. I'll update with a comment here (and close this) when we know if this is possible!

PR #31976, opened 2022-08-07
hypre: Homepage changed

cessenat:
It appears https://computing.llnl.gov/project/linear_solvers/software.php is not accessible while
https://computing.llnl.gov/projects/hypre-scalable-linear-solvers-multigrid-methods
is fine.


osborn9 (approved):
Thanks for the fix!

alalazo (approved):

PR #31980, opened 2022-08-08
crunch: add new package

adamjstewart:
Successfully builds on Ubuntu 18.04 with GCC 7.5.0. DDS shows up in the list of supported drivers in `gdalinfo --formats`.

### DIFF
--- a/var/spack/repos/builtin/packages/crunch/package.py
+++ b/var/spack/repos/builtin/packages/crunch/package.py
@@ -1,0 +1,21 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Crunch(CMakePackage):
+    """Advanced DXTc texture compression and transcoding library."""
+
+    homepage = "https://github.com/BinomialLLC/crunch"
+    # Fork with CMake build system
+    git = "https://github.com/rouault/crunch.git"
+
+    # No stable releases since 2012
+    version("master", branch="build_fixes")
+
+    depends_on("cmake@3.5:", type="build")
+
+    conflicts("platform=darwin")
--- a/var/spack/repos/builtin/packages/crunch/package.py
+++ b/var/spack/repos/builtin/packages/crunch/package.py
@@ -10,7 +10,10 @@ class Crunch(CMakePackage):
     """Advanced DXTc texture compression and transcoding library."""
 
     homepage = "https://github.com/BinomialLLC/crunch"
-    # Fork with CMake build system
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/crunch/package.py
```diff
@@ -0,0 +1,21 @@
...
+
+from spack.package import *
+
+
+class Crunch(CMakePackage):
+    """Advanced DXTc texture compression and transcoding library."""
+
+    homepage = "https://github.com/BinomialLLC/crunch"
+    # Fork with CMake build system
+    git = "https://github.com/rouault/crunch.git"
```
alalazo, line 17: I am a bit concerned adding a 0 star repository as git. Is this repo used by people? Can't we try to push for https://github.com/BinomialLLC/crunch/pull/33 being merged?

adamjstewart: Completely understand your concern. Let me add some background.

I have no interest in crunch, I only added it as a GDAL dependency. The [GDAL docs](https://gdal.org/build_hints.html#crnlib) explicitly point to this branch of this fork (presumably due to its ease of installation). The [original fork](https://github.com/BinomialLLC/crunch/graphs/contributors) seems poorly maintained at best. Most PRs and issues have no comments.

This fork was created by @rouault, who also maintains GDAL, so I think I actually trust it more than the original fork, especially for use with GDAL.

@BinomialLLC is there any chance of https://github.com/BinomialLLC/crunch/pull/33 being merged so we can use the original repo instead? Also, would you have any interest in maintaining this build recipe in Spack?

rouault: > so I think I actually trust it more than the original fork, especially for use with GDAL.

the ony point of my fork is that it collected a few patches from different PRs that I needed to get things working. It is in no way intended to be extended/maintained. It would be indeed less confusing if we could get rid of it if those changes were merged in upstream BinomialLLC/crunch

alalazo: Given that the upstream package seems not to be maintained, how many patches are we talking about? If they're just a few, can we pin a commit from upstream and cherry pick changes from the other repo as patches?

adamjstewart: We are talking about 5 patches: https://github.com/rouault/crunch/commits/build_fixes

We could use upstream and add these 5 commits as patches, but I'm not sure how much of an advantage that would be.

adamjstewart: @alalazo how do you want to progress here?

alalazo: Can you write briefly what you explained in https://github.com/spack/spack/pull/31980#discussion_r940481883 in a comment above the git repository? Then I'll be fine merging this.

alalazo (approved):

PR #24290, opened 2021-06-12
googletest: add v1.11 (default) and "live at head"

sethrj:
Add v1.11, and a "master" version for people who want to live at head like googletest now suggests.

EDIT: prefer v1.10 because 1.11  seems  to  break some  packages.

PR #31983, opened 2022-08-08
libjxl: add new package

adamjstewart:
Seeing some lib linking issues on macOS, will investigate later.

https://github.com/libjxl/libjxl/issues/1686


adamjstewart:
@rouault is it possible to build GDAL's JXL driver with external libtiff?

rouault:
> is it possible to build GDAL's JXL driver with external libtiff?

you must mean the GDAL's JXL TIFF codec (saying that, since in GDAL master, there's a JPEG-XL driver for .jxl images not embedded in TIFF file). To answer your question: no, the JXL TIFF codec in GDAL requires GDAL to be built with its internal libtiff

adamjstewart:
That's unfortunate, but hopefully support for that will be added someday. I think it's still worth finishing this PR since it sounds like GDAL has another way of using libjxl in master.

rouault:
> That's unfortunate, but hopefully support for that will be added someday.

will require the codec to be moded to upstream libtiff first. Probably once libjxl API has been frozen and that interest of JXL as a TIFF codec is obvious

> GDAL has another way of using libjxl in master.

The GDAL JPEG-XL standalone driver and JPEG-XL in TIFF are 2 different things /  serve different purposes

adamjstewart:
Fixed brotli linking issues by building with external brotli, should be good to merge now.

var/spack/repos/builtin/packages/googletest/package.py
```diff
@@ -14,12 +14,10 @@ class Googletest(CMakePackage):
     git = "https://github.com/google/googletest"
 
     version("main", branch="main")
+    version("1.12.1", sha256="81964fe578e9bd7c94dfdb09c8e4d6e6759e19967e397dbea48d1c10e45d0df2")
+    version("1.12.0", sha256="2a4f11dce6188b256f3650061525d0fe352069e5c162452818efbbf8d0b5fe1c")
     version("1.11.0", sha256="07b0896360f8e14414a8419e35515da0be085c5b4547c914ab8f4684ef0a3a8e")
-    version(
-        "1.10.0",
-        sha256="e4a7cd97c903818abe7ddb129db9c41cc9fd9e2ded654be57ced26d45c72e4c9",
-        preferred=True,
```
adamjstewart, line 21: @sethrj it looks like you added this in #24290 because it seemed like 1.11.0 broke some packages. Unfortunately, `@1.10.0+gmock` doesn't build for me (at least on macOS arm64), so I'm not sure if 1.10.0 is a good default either. I think it's better to default to the latest version and fix any packages that are broken in CI.

var/spack/repos/builtin/packages/lodepng/package.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```
adamjstewart, line 1: lodepng and sjpeg are deps of libjxl, but there doesn't seem to be a way to use external installations of these. So these packages don't necessarily have to be in this PR, let me know if you want me to move them to a separate PR to make this PR simpler to review.

alalazo (approved):

var/spack/repos/builtin/packages/googletest/package.py
sethrj: This was probably "solved" by using clingo instead of the old concretizer: I remember the newer versions causing E4S some issues, but either their code has been upgraded or conflicts have been added.

PR #31989, opened 2022-08-08
py-uproot: Add checksums for new versions

iarspider:


vvolkl (dismissed):

tldahlgren (dismissed):
Confirmed all new sha256.

tldahlgren:
@adamjstewart Did you want to comment on this package since it is python based?

var/spack/repos/builtin/packages/py-uproot/package.py
```diff
@@ -45,6 +53,7 @@ class PyUproot(PythonPackage):
     )
 
     depends_on("python@2.6:2,3.5:", type=("build", "run"))
+    depends_on("python@3.6:", type=("build", "run"), when="@4.2.0:")
     depends_on("py-setuptools", type="build")
     depends_on("py-setuptools@42:", type="build", when="@4.1.8:")
```
adamjstewart: I think setuptools should actually be build/run

adamjstewart (approved):

PR #31990, opened 2022-08-08
Swig externfind

cessenat:
Adds the external find feature,
Adds a tag as "build-tools"


sethrj (approved):
Confirmed that this works for my homebrew installed swig 4.0.2. Thanks!

PR #31995, opened 2022-08-08
New command, `spack change`, to change existing env specs

scheibelp:
If you have an environment like

```
$ cat spack.yaml
spack:
  specs: [openmpi@4.1.0+cuda]
```

this PR provides a new command `spack change` that you can use to adjust environment specs from the command line:

```
$ spack change openmpi~cuda
$ cat spack.yaml
spack:
  specs: [openmpi@4.1.0~cuda]
```

in other words, this allows you to tweak the details of environment specs from the command line.

Notes:

* This is only allowed for environments that do not define matrices
  * This is possible but not anticipated to be needed immediately
  * If this were done, it should probably only be done for "named"/not-anonymous specs (i.e. we can change `openmpi+cuda` but not spec like `+cuda` or `@4.0.1~cuda`)


haampie:
will anybody ever use this though?

scheibelp:
> will anybody ever use this though?

This was based on an internal request. There are a couple goals with this:

* Make it easier for tools running "on top of" Spack to manipulate data structures within it (without this PR, if you want to change an environment spec you have to manually open/edit the `spack.yaml`, which is fine for users interfacing with Spack directly but not for programs or scripts)
* Provide a more-streamlined interface for interacting with Spack (while I personally find it convenient to manually edit the `spack.yaml`, other users may not want to, and it admittedly requires familiarity with yaml and other concepts they may not care about in the moment that they want to change specs).

haampie:
Right, I think if I would use spack change I would immediately check the resulting yaml file, so there's no time saved

haampie:
Can it at least be put under `spack env` instead in the global set of commands? I'm still thinking it will rarely be used and `change` is such a generic command...

Also, I don't think this pr deals with changing say a provider. Or "changing" the flavor of a dependency not listed as root spec. :shrug:

scheibelp:
Close/reopen to restart docs build

lib/spack/spack/environment/environment.py
```diff
@@ -1024,6 +1024,47 @@ def add(self, user_spec, list_name=user_speclist_name):
 
         return bool(not existing)
 
+    def change_existing_spec(self, change_spec, list_name=user_speclist_name, match_spec=None):
+        """
+        Provide a single spec and assume you want to find another spec with the
+        same name and override it with any conflicting details in change_spec.
```
tgamblin: Needs arg docs, and docs above should be updated to account for `match_spec`

lib/spack/spack/environment/environment.py
```diff
@@ -1024,6 +1024,47 @@ def add(self, user_spec, list_name=user_speclist_name):
...
+
+        This is only allowed for environments that do not use matrices (i.e.
+        they have a single, explicitly-enumerated list of specs called
+        'specs').
```
tgamblin: I think this should work on any list that doesn't contain a matrix, right?  Or, could it not be made to work on any single spec in a list, even if the list has matrices?

lib/spack/spack/environment/environment.py
```diff
@@ -1024,6 +1024,47 @@ def add(self, user_spec, list_name=user_speclist_name):
...
+        """
+        if not (change_spec.name or (match_spec and match_spec.name)):
+            raise ValueError(
+                "Must specify a spec name to identify a single spec"
```
tgamblin, line 1049: does it have to match exactly one spec?

lib/spack/spack/environment/environment.py
```diff
@@ -1024,6 +1024,47 @@ def add(self, user_spec, list_name=user_speclist_name):
...
+                " specify a named list that is not a matrix"
+            )
+
+        matches = list(x for x in list_to_change if x.satisfies(match_spec))
+        if len(matches) == 0:
+            raise ValueError(
+                "There are no specs named {0} in {1}".format(match_spec.name, list_name)
+            )
+        elif len(matches) > 1:
+            raise ValueError("{0} matches multiple specs".format(str(match_spec)))
```
tgamblin, line 1067: is it not safe to allow modifying multiple specs?

lib/spack/spack/spec.py
```diff
@@ -2242,6 +2258,30 @@ def read_yaml_dep_specs(deps, hash_type=ht.dag_hash.name):
...
+        if change_spec.versions and not change_spec.versions == spack.version.ver(":"):
+            new_spec.versions = change_spec.versions
+        for variant, value in change_spec.variants.items():
+            if variant in package_cls.variants:
+                if variant in new_spec.variants:
+                    new_spec.variants.substitute(value)
+                else:
+                    new_spec.variants[variant] = value
+            else:
+                raise ValueError("{0} is not a variant of {1}".format(variant, new_spec.name))
```
tgamblin, line 2276: shouldn't flags also be handed here?

tgamblin (changes requested):
Some minor requests, and this needs docs.

I am still thinking about the name for the command, but it looks good and matches up at least semantically with `add`.

lib/spack/spack/environment/environment.py
scheibelp: Whoops this is out of date:

> I think this should work on any list that doesn't contain a matrix, right?

yes (and it does, and the tests check that).

> Or, could it not be made to work on any single spec in a list, even if the list has matrices?

possibly: the `SpecList` interface presents itself as a list so is opaque to single-element modifications. If it were refactored, this would be possible. IMO this PR is not the best place to do that.

lib/spack/spack/environment/environment.py
scheibelp: I consider it ambiguous (I assume a user would want to change a single spec, and therefore a request that changes multiple is unintentional - perhaps an option to allow that would be useful?).

lib/spack/spack/spec.py
scheibelp: I assume you mean compiler flags and yes I missed that.

lib/spack/spack/environment/environment.py
scheibelp: A `-a` option has now been added.

lib/spack/spack/spec.py
scheibelp: Those are now handled (and tested)

lib/spack/spack/environment/environment.py
scheibelp: I think this is resolved by https://github.com/spack/spack/pull/31995#discussion_r952303271

lib/spack/spack/environment/environment.py
scheibelp: The docstring is now updated

scheibelp:
@tgamblin I have addressed the review comments: I think this is ready for another look.

tgamblin (approved):

PR #32005, opened 2022-08-09
Fix performance regression with `spack mirror create --all`

### LINKED ISSUE #31985: `Spack mirror --all` now takes so long to run it's infeasible to do so.
### Steps to reproduce

1. Run `spack mirror --all`
2. Wait at least 3 days.
3. ...
4. Profit

### Error message

@alalazo Patch to fix the issues with `spack mirror --all` have added a concretization step for each package to be mirrored. Instead of taking less than a second to check if a package@version has already been mirrored, for the average package it takes about 20 seconds. 

I started a run of `spack mirror --all` yesterday to see how long it took. It used to take about 10 minutes to run, given that our mirror already had just about everything and only a few package changes needed syncing. It's now been running for over 20 hours, and is only at the letter F in the alphabet.

# spack debug report
* **Spack:** 0.19.0.dev0
* **Python:** 3.6.8
* **Concretizer:** clingo

### General information

alalazo:
fixes #31985 

This PR fixes the performance regression reported in #31985 and a few other issues found while refactoring the `spack mirror create` command (e.g. default directory not canonicalized, specs read from text file treated differently than specs from command line, allowing `-n` together with `--all`, etc.).

Modifications:
- [x] Do not require concretization for `spack mirror create --all`
- [x] Forbid using `--versions-per-spec` together with `--all`
- [x] Fixed a few issues when reading specs from input file (specs were not concretized, command would fail when trying to mirror dependencies)
- [x] Fix issue with default directory for `spack mirror create` not being canonicalized
- [x] Changed slightly the wording for reporting (it was mentioning "Successfully  created" even in presence of errors)
- [x] Fix issue with `colify` (was not called properly during error reporting)
- [x] Add more unit tests to poke `spack mirror create` 
- [x] Skip externals also when mirroring environments

### DIFF
--- a/lib/spack/llnl/util/lang.py
+++ b/lib/spack/llnl/util/lang.py
@@ -13,7 +13,7 @@ import re
 import sys
 import traceback
 from datetime import datetime, timedelta
-from typing import List, Tuple
+from typing import Any, Callable, Iterable, List, Tuple
 
 import six
 from six import string_types
@@ -977,7 +977,11 @@ def enum(**kwargs):
     return type("Enum", (object,), kwargs)
 
 
-def stable_partition(input_iterable, predicate_fn):
+def stable_partition(
+    input_iterable,  # type: Iterable
+    predicate_fn,  # type: Callable[[Any], bool]
+):
+    # type: (...) -> Tuple[List[Any], List[Any]]
     """Partition the input iterable according to a custom predicate.
 
     Args:
--- a/lib/spack/llnl/util/lang.py
+++ b/lib/spack/llnl/util/lang.py
@@ -981,8 +981,8 @@ def stable_partition(input_iterable, predicate_fn):
     """Partition the input iterable according to a custom predicate.
 
     Args:
-        input_iterable (list of Spec): input iterable to be partitioned.
-        predicate_fn (callable): predicate function accepting an iterable item
+        input_iterable: input iterable to be partitioned.
+        predicate_fn: predicate function accepting an iterable item
             as argument.
 
     Return:
--- a/lib/spack/llnl/util/lang.py
+++ b/lib/spack/llnl/util/lang.py
@@ -977,6 +977,27 @@ def enum(**kwargs):
     return type("Enum", (object,), kwargs)
 
 
+def stable_partition(input_iterable, predicate_fn):
+    """Partition the input iterable according to a custom predicate.
+
+    Args:
+        input_iterable (list of Spec): input iterable to be partitioned.
+        predicate_fn (callable): predicate function accepting an iterable item
+            as argument.
+
+    Return:
+        Tuple of the list of elements evaluating to True, and
+        list of elements evaluating to False.
+    """
+    true_items, false_items = [], []
+    for item in input_iterable:
+        if predicate_fn(item):
+            true_items.append(item)
+            continue
+        false_items.append(item)
+    return true_items, false_items
+
+
 class TypedMutableSequence(MutableSequence):
     """Base class that behaves like a list, just with a different type.
 
--- a/lib/spack/spack/build_systems/racket.py
+++ b/lib/spack/spack/build_systems/racket.py
@@ -40,13 +40,13 @@ class RacketPackage(PackageBase):
 
     pkgs = False
     subdirectory = None  # type: Optional[str]
-    name = None  # type: Optional[str]
+    racket_name = None  # type: Optional[str]
     parallel = True
 
     @lang.classproperty
     def homepage(cls):
         if cls.pkgs:
-            return "https://pkgs.racket-lang.org/package/{0}".format(cls.name)
+            return "https://pkgs.racket-lang.org/package/{0}".format(cls.racket_name)
 
     @property
     def build_directory(self):
@@ -66,7 +66,7 @@ class RacketPackage(PackageBase):
                 "-t",
                 "dir",
                 "-n",
-                self.name,
+                self.racket_name,
                 "--deps",
                 "fail",
                 "--ignore-implies",
@@ -86,5 +86,5 @@ class RacketPackage(PackageBase):
                     (
                         "Racket package {0} was already installed, uninstalling via "
                         "Spack may make someone unhappy!"
-                    ).format(self.name)
+                    ).format(self.racket_name)
                 )
--- a/lib/spack/spack/cmd/mirror.py
+++ b/lib/spack/spack/cmd/mirror.py
@@ -15,11 +15,11 @@ import spack.config
 import spack.environment as ev
 import spack.mirror
 import spack.repo
+import spack.spec
 import spack.util.path
 import spack.util.url as url_util
 import spack.util.web as web_util
 from spack.error import SpackError
-from spack.spec import Spec
 from spack.util.spack_yaml import syaml_dict
 
 description = "manage mirrors (source and binary)"
@@ -256,14 +256,7 @@ def concrete_specs_from_user(args):
     specs = concrete_specs_from_cli_or_file(args)
     if args.dependencies:
         specs = extend_with_dependencies(specs)
-
-    # Skip external specs, as they are already installed
-    external_specs = [s for s in specs if s.external]
-    specs = [s for s in specs if not s.external]
-
-    for spec in external_specs:
-        msg = "Skipping {0} as it is an external spec."
-        tty.msg(msg.format(spec.cshort_spec))
+    specs = filter_externals(specs)
 
     if versions_per_spec(args) == "all":
         mirror_specs = spack.mirror.get_all_versions(specs)
@@ -272,8 +265,20 @@ def concrete_specs_from_user(args):
             specs, num_versions=versions_per_spec(args)
         )
     mirror_specs.sort(key=lambda s: (s.name, s.version))
-    mirror_specs = filter_mirror_specs(args, mirror_specs)
+    mirror_specs, _ = spack.spec.partition_spec_list(
+        mirror_specs, predicate_fn=not_excluded_fn(args)
+    )
     return mirror_specs
+
+
+def filter_externals(specs):
+    specs, external_specs = spack.spec.partition_spec_list(
+        specs, predicate_fn=lambda x: not x.external
+    )
+    for spec in external_specs:
+        msg = "Skipping {0} as it is an external spec."
+        tty.msg(msg.format(spec.cshort_spec))
+    return specs
 
 
 def extend_with_dependencies(specs):
@@ -300,32 +305,40 @@ def concrete_specs_from_cli_or_file(args):
     return specs
 
 
-def filter_mirror_specs(args, mirror_specs):
+def not_excluded_fn(args):
+    """Return a predicate that evaluate to True if a spec was not explicitly
+    excluded by the user.
+    """
     exclude_specs = []
     if args.exclude_file:
         exclude_specs.extend(specs_from_text_file(args.exclude_file, concretize=False))
     if args.exclude_specs:
         exclude_specs.extend(spack.cmd.parse_specs(str(args.exclude_specs).split()))
-    if exclude_specs:
-        mirror_specs = list(
-            x for x in mirror_specs if not any(x.satisfies(y, strict=True) for y in exclude_specs)
-        )
-    return mirror_specs
+
+    def not_excluded(x):
+        return not any(x.satisfies(y, strict=True) for y in exclude_specs)
+
+    return not_excluded
 
 
 def concrete_specs_from_environment(args):
     env = ev.active_environment()
     assert env, "an active environment is required"
     mirror_specs = env.all_specs()
-    mirror_specs = filter_mirror_specs(args, mirror_specs)
+    mirror_specs = filter_externals(mirror_specs)
+    mirror_specs, _ = spack.spec.partition_spec_list(
+        mirror_specs, predicate_fn=not_excluded_fn(args)
+    )
     return mirror_specs
 
 
 def all_specs_with_all_versions(args):
-    specs = [Spec(n) for n in spack.repo.all_package_names()]
+    specs = [spack.spec.Spec(n) for n in spack.repo.all_package_names()]
     mirror_specs = spack.mirror.get_all_versions(specs)
     mirror_specs.sort(key=lambda s: (s.name, s.version))
-    mirror_specs = filter_mirror_specs(args, mirror_specs)
+    mirror_specs, _ = spack.spec.partition_spec_list(
+        mirror_specs, predicate_fn=not_excluded_fn(args)
+    )
     return mirror_specs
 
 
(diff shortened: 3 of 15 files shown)


alalazo:
@Paul-Ferrell Can you test this branch on your use case? It should fix all the issues you've seen so far. Note though that here I forbid using `--all` together with `--versions-per-spec`, so you have to remove the latter option. Let me know if you still see issues.

alalazo:
For the record, I'll stop here with refactoring, but there are many more improvements that can be done in future PRs related to mirrors:
- The classes representing a mirror are complex and scarcely related to each other. I was surprised for instance to see that a `Mirror` has no method to push or pull something, since that is done using a cache which is a totally separate object. Even mirror statistics are a separate object from both caches and mirrors. I think there's room here to have a more coherent design.
- Stages and fetchers are too intertwined and require a package object to be instantiated cleanly. Making them more loosely coupled can enable using only fetchers in mirrors, which would make it easier to fetch tarballs in parallel.

Paul-Ferrell:
Much better. I am getting this error though: 
```
lib/spack/spack/error.py:54 ==> [2022-08-09-06:33:20.936530] Error: Package 'base' not found.
  File "/storage/spack/spack/lib/spack/spack/repo.py", line 1306, in get_pkg_class
    module = importlib.import_module(fullname)
  File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'spack.pkg.builtin.base'

  File "/storage/spack/spack/lib/spack/spack/main.py", line 961, in main
    return _main(argv)
  File "/storage/spack/spack/lib/spack/spack/main.py", line 916, in _main
  File "/storage/spack/spack/lib/spack/spack/main.py", line 944, in finish_parse_and_run
  File "/storage/spack/spack/lib/spack/spack/main.py", line 608, in _invoke_command
  File "/storage/spack/spack/lib/spack/spack/cmd/mirror.py", line 492, in mirror
    action[args.mirror_command](args)
  File "/storage/spack/spack/lib/spack/spack/cmd/mirror.py", line 425, in mirror_create
    create_mirror_for_all_specs(args)
  File "/storage/spack/spack/lib/spack/spack/cmd/mirror.py", line 442, in create_mirror_for_all_specs
    pkg_cls = spack.repo.path.get_pkg_class(candidate.name)
  File "/storage/spack/spack/lib/spack/spack/repo.py", line 962, in get_pkg_class
    return self.repo_for_pkg(pkg_name).get_pkg_class(pkg_name)
  File "/storage/spack/spack/lib/spack/spack/repo.py", line 1308, in get_pkg_class
    raise UnknownPackageError(pkg_name)
spack.repo.UnknownPackageError: Package 'base' not found.
```

alalazo:
@Paul-Ferrell Can you give me more context (which command you were executing, configuration if relevant)? What I see is that `spack` fails at finding the `base` package in the builtin repository, which is correct. What I don't see is why it is looking for that.

EDIT: nevermind, I was able to reproduce.

alalazo:
@Paul-Ferrell Can you re-try? We had a few packages overriding an attribute that was meant for internal use.

Paul-Ferrell:
Ok, I think this one's a winner. After a couple minutes we're already up to the letter 'G' without any issues.

Paul-Ferrell:
I was checking on a different problem (more on that later) and ran into this when mirroring a single package:
I was running `spack --debug mirror create -n all gromacs`

```
  File "/home/pflarr/work_space/spack-mirror/bin/spack", line 100, in <module>
    sys.exit(spack.main.main())
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/main.py", line 961, in main
    return _main(argv)
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/main.py", line 916, in _main
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/main.py", line 944, in finish_parse_and_run
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/main.py", line 608, in _invoke_command
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/cmd/mirror.py", line 492, in mirror
    action[args.mirror_command](args)
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/cmd/mirror.py", line 432, in mirror_create
    create_mirror_for_individual_specs(args)
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/cmd/mirror.py", line 459, in create_mirror_for_individual_specs
    create_mirror_from_concrete_specs(mirror_specs, args)
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/cmd/mirror.py", line 365, in create_mirror_from_concrete_specs
    directory, mirror_specs, args.skip_unstable_versions
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/mirror.py", line 508, in create
    create_mirror_from_package_object(spec.package, mirror_cache, mirror_stats)
  File "/home/pflarr/work_space/spack-mirror/lib/spack/spack/spec.py", line 1521, in package
    assert self.concrete, "Spec.package can only be called on concrete specs"
AssertionError: Spec.package can only be called on concrete specs
```

Paul-Ferrell:
The other problem I ran into is throttling from gitlab.com. I think it will be fine over multiple runs, as the packages that got missed should just get picked up later.

alalazo:
The last commit should fix https://github.com/spack/spack/pull/32005#issuecomment-1209458406 It won't be as fast as it could, since it needs to concretize, but I'd rather try to speed that up in a following PR.

lib/spack/spack/spec.py
```diff
@@ -5364,6 +5364,26 @@ def save_dependency_specfiles(
             fd.write(dep_spec.to_json(hash=ht.dag_hash))
 
 
+def partition_spec_list(input_specs, predicate_fn):
```
haampie: Can you move this elsewhere? It's just a generic stable partition

lib/spack/spack/spec.py
```diff
@@ -5364,6 +5364,26 @@ def save_dependency_specfiles(
...
+    Return:
+        Tuple of the list of elements evaluating to True, and
+        list of elements evaluating to False.
+    """
+    true_items, false_items = [], []
+    for item in input_specs:
+        if predicate_fn(item):
+            true_items.append(item)
+            continue
+        false_items.append(item)
```
haampie: more fancy than necessary imho ;p 
```suggestion
        else:
           false_items.append(item)
```

lib/spack/spack/spec.py
alalazo: Moved to `llnl.util.lang` and renamed input args + reworked docstring

lib/spack/llnl/util/lang.py
```diff
@@ -977,6 +977,27 @@ def enum(**kwargs):
     return type("Enum", (object,), kwargs)
 
 
+def stable_partition(input_iterable, predicate_fn):
+    """Partition the input iterable according to a custom predicate.
+
+    Args:
+        input_iterable: input iterable to be partitioned.
+        predicate_fn: predicate function accepting an iterable item
+            as argument.
```
scheibelp, line 990: (minor) might be worth mentioning that we `bool()` the result of `predicate_fn` (i.e. we don't strictly require `True/False` and will e.g. interpret values like `None` as `False`).

lib/spack/spack/build_systems/racket.py
```diff
@@ -40,13 +40,13 @@ class RacketPackage(PackageBase):
 
     pkgs = False
     subdirectory = None  # type: Optional[str]
-    name = None  # type: Optional[str]
+    racket_name = None  # type: Optional[str]
```
scheibelp, line 43: (minor) Is it worth adding some package verification step that ensures that "reserved" members are not overridden? Not in this PR but later on (we could add a TODO here though).

lib/spack/spack/cmd/mirror.py
```diff
@@ -231,142 +234,225 @@ def mirror_list(args):
...
-        )
+    Args:
+        filename (str): name of the file containing the abstract specs.
+        concretize (bool): if True concretize the specs before returning
+            the list.
+    """
+    with open(filename, "r") as f:
+        specs_in_file = f.readlines()
+        specs_in_file = [s.strip() for s in specs_in_file]
+    return spack.cmd.parse_specs(" ".join(specs_in_file), concretize=concretize)
```
scheibelp, line 250: (question) is there a benefit to passing all of these to the same call?

lib/spack/spack/test/cmd/mirror.py
```diff
@@ -288,3 +295,115 @@ def test_mirror_destroy(
...
+            ),
+            (
+                {"specs": "hdf5", "file": "input.txt", "all": False},
+                "cannot specify specs with a file AND",
+            ),
+            ({"specs": None, "file": None, "all": False}, "no packages were specified"),
+            # Passed -n along with --all
+            (
+                {"specs": None, "file": None, "all": True, "versions_per_spec": 2},
+                "cannot specify '--versions_per-spec'",
```
scheibelp, line 327: Why is `-n -a` not allowed? I have found this useful in the past (e.g. it is convenient to make a mirror with the latest version of each package e.g. with `-a -n 1`)

lib/spack/spack/test/cmd/mirror.py
```diff
@@ -288,3 +295,115 @@ def test_mirror_destroy(
...
+    def test_error_conditions(self, cli_args, error_str):
+        args = MockMirrorArgs(**cli_args)
+        with pytest.raises(spack.error.SpackError, match=error_str):
+            spack.cmd.mirror.mirror_create(args)
+
+    @pytest.mark.parametrize(
+        "cli_args,expected_end",
+        [
+            ({"directory": None}, os.path.join("var", "spack", "cache")),
+            ({"directory": os.path.join("foo", "bar")}, os.path.join("foo", "bar")),
```
scheibelp, line 340: For a relative path, I think this produces `file://foo/bar`: is that valid?

Also I think this should use `mock_config` (since the fallback behavior depends on the config).

lib/spack/spack/cmd/mirror.py
```diff
@@ -231,142 +234,225 @@ def mirror_list(args):
...
+    )
+    for candidate in mirror_specs:
+        pkg_cls = spack.repo.path.get_pkg_class(candidate.name)
+        pkg_obj = pkg_cls(spack.spec.Spec(candidate))
+        mirror_stats.next_spec(pkg_obj.spec)
+        spack.mirror.create_mirror_from_package_object(pkg_obj, mirror_cache, mirror_stats)
+    process_mirror_stats(*mirror_stats.stats())
+
+
+def create_mirror_for_all_specs_inside_environment(args):
```
scheibelp: Would it be possible to resolve `args` into the actual arguments used by each function at this level?

For example I think the only thing this needs is the excluded specs.

(When I say "at this level" I mean for this function as well as `create_mirror_for_individual_specs`, `create_mirror_for_all_specs`, etc.)

scheibelp (changes requested):
I have a couple requests and a couple questions. The largest requests (which I think are small) are:

* Process `args` into actual named variables before calling methods like `create_mirror_for_individual_specs`
* Allow `-a -n 1`

(I have specific comments on these with more detail)

I don't expect either would be time-consuming to address, but if I didn't estimate that right, let me know.

lib/spack/spack/cmd/mirror.py
alalazo: Yes, it ensures that specs that are concretized are always treated in the same way, e.g. if they fail the concretization step they emit the same error with the same message. This, in the end, gives a more consistent behavior of the command.

lib/spack/spack/test/cmd/mirror.py
alalazo: Because I took the help message as the declared intent of what the function should do:
```console
  -a, --all             mirror all versions of all packages in Spack, or all packages in the current environment if there is an active environment (this requires significant time and space)
```
If we want to have something that collects only the latest version, can we add an argument like `--all-latest` to do that?

lib/spack/spack/test/cmd/mirror.py
alalazo: See 2d48799

lib/spack/spack/cmd/mirror.py
alalazo: > For example I think the only thing this needs is the excluded specs

It needs also the `directory` where to place the mirror and whether or not to skip unstable versions. 

> Would it be possible to resolve args into the actual arguments used by each function at this level?

Done, though I have to say I slightly prefer the previous version of the code.[^1]  

[^1]: I prefer the previous version since the main `mirror_create` function was just doing error checking and then forwarding to other function if `args` passed all checks - so a terser view of the overall behavior without further details expanded. It's a minor style preference though.

lib/spack/spack/build_systems/racket.py
alalazo: Sure, I'll try to setup a different PR to do that.

lib/spack/llnl/util/lang.py
alalazo: I don't think we should document that, nor we should complicate the implementation. So I added static checks with type hints to say what we expect.

lib/spack/spack/test/cmd/mirror.py
alalazo: Also, can we do this in a follow up PR?

alalazo:
@scheibelp Ready for another pass

lib/spack/spack/test/cmd/mirror.py
scheibelp: Yes. If I'm the only one who's ever used it, I don't mind if it takes a while to come back, it doesn't look hard to reintroduce (i.e. it doesn't look like any of the changes in this PR make it difficult to extract a limited number of versions for each package).

lib/spack/llnl/util/lang.py
scheibelp: > So I added static checks with type hints to say what we expect.

That seems fine to me, but I'm curious before this thread gets collapsed: do we use this approach anywhere else? I had to look at https://github.com/spack/spack/pull/32005/commits/ecf225f167ee203355e885e37044d454e7d93fa5 to figure out what you meant. Usually we specify constraints on arguments in the docstring (or enforce with assert/raise in the method body).

lib/spack/spack/cmd/mirror.py
scheibelp: Got it, my original concern was that it would concretize all the specs together (which would not be desirable for mirrors). AFAIK our API doesn't make any guarantees one way or the other (e.g. I see no unit tests enforcing this or docstrings mentioning this behavior explicitly).

lib/spack/spack/cmd/mirror.py
scheibelp: Thanks I appreciate this change (I often get confused when `args` from commands is passed transitively, since for me the method signature helps me understand at-a-glance what the function is "planning" to do, and when the parameters are grouped into a single opaque object I have to read the whole function and possibly function subcalls to understand it).

scheibelp (approved):
I have a very minor concern related to future proofing at https://github.com/spack/spack/pull/32005#discussion_r944010626 but this seems good

scheibelp:
Thanks!

lib/spack/llnl/util/lang.py
alalazo: I think this should be caught by `mypy` during style checks in CI. We have similar annotations in some class members here and there.

PR #32008, opened 2022-08-09
Add base class for dir visitor

haampie:
Standalone change from #31948, adds a simple interface for visiting a dir recursively, differentiating between normal files/dirs and symlinked files/dirs.

### DIFF
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -81,6 +81,8 @@ __all__ = [
     "unset_executable_mode",
     "working_dir",
     "keep_modification_time",
+    "BaseDirectoryVisitor",
+    "visit_directory_tree",
 ]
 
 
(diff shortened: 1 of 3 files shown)


lib/spack/llnl/util/filesystem.py
```diff
@@ -1133,21 +1135,31 @@ def lexists_islink_isdir(path):
     return True, is_link, is_dir
 
 
+class BaseDirectoryVisitor(object):
+    """Base class and interface for visit_directory_tree."""
```
alalazo: Just wanted to point out that we might use https://python.readthedocs.io/en/latest/library/abc.html#module-abc if it makes sense. Otherwise, this is fine coded as a normal base class.

lib/spack/llnl/util/filesystem.py
```diff
@@ -1133,21 +1135,31 @@ def lexists_islink_isdir(path):
...
+
+    def visit_file(self, root, rel_path, depth):
+        pass
```
alalazo: All these methods could benefit docstring to make the semantic of the function and input arguments explicit.

alalazo (dismissed):
LGTM. Left two minor comments.

lib/spack/llnl/util/filesystem.py
haampie: done

lib/spack/llnl/util/filesystem.py
haampie: I think it doesn't have to be abstract, cause the `return False` is sensible for symlinked dirs.

alalazo (approved):

PR #32014, opened 2022-08-09
gitlab ci: Make sure PR pipelines can push to buildcache

scottwittenburg:
On PR pipelines we need to override the buildcache destination to
point to the "spack-binaries-prs" bucket, otherwise, those pipelines
try to push to the default mirror in a bucket for which they don't
have write permission.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -193,6 +193,7 @@ e4s-mac-pr-generate:
     - cd share/spack/gitlab/cloud_pipelines/stacks/${SPACK_CI_STACK_NAME}
     - spack env activate --without-view .
     - spack ci generate --check-index-only
+      --buildcache-destination "${SPACK_BUILDCACHE_DESTINATION}"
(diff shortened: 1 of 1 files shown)


scottwittenburg:
@eugeneswalker I think this should address the problem you mentioned on slack, where PR pipelines for this stack failed to push binaries to the mirror and subsequently installed dependencies from source.

eugeneswalker (approved):

eugeneswalker:
> @eugeneswalker I think this should address the problem you mentioned on slack, where PR pipelines for this stack failed to push binaries to the mirror and subsequently installed dependencies from source.

Thank you!

eugeneswalker:
@spackbot run pipeline

eugeneswalker:
Do we need to rebase this on develop to get past the failure?
```
==> Error: Pipeline generation failed due to the presence of the following specs that are known to be broken in develop:
...
```

scottwittenburg:
@spackbot run pipeline

scottwittenburg:
> Do we need to rebase this on develop to get past the failure?

Not in this case.  Since that seemed a spurious failure which could not be reproduced by myself or by @trws, we decided to remove it from the list yesterday.

PR #26136, opened 2021-09-22
feature: new tags command

tldahlgren:
Fixes #22938 

This PR adds a `spack tags` command to output package tags or (available) packages with those tags.  It also ensures each package is listed in the tag cache ONLY ONCE per tag.

```
$ spack tags --help
usage: spack tags [-hia] [tag [tag ...]]

Show package tags and associated packages

positional arguments:
  tag              show packages with the specified tag

optional arguments:
  -h, --help       show this help message and exit
  -i, --installed  show information for installed packages only
  -a, --all        show packages for all available tags

Tags from known packages will be used if no tags are provided on the command
line. If tags are provided, packages with at least one will be reported.

You are not allowed to provide tags and use '--all' at the same time.


$ spack tags
30 available package tags:
    amr            finite-elements  hpc            radiuss
    build-tools    flow             hydrodynamics  simulations
    cfd            fluid            i/o            solver
    e4s            gibbs sampler    key4hep        spectral-elements
    ecp            gibbs-sampler    mcmc           storage
    ecp-apps       hep              navier-stokes  trace-tools
    ecp-proxy-app  heps             neutrino
    fem            high-order       proxy-app


$ spack tags -a
amr:
    mfem

build-tools:
    autoconf  cmake  gawk   libtool  pkg-config
    automake  flex   gmake  m4       pkgconf

[..snip..]

trace-tools:
    uftrace

$ spack tags -is
amr:
    No installed packages

build-tools:
    autoconf  automake  cmake  flex  libtool  m4  pkg-config  pkgconf

[..snip..]

trace-tools:
    No installed packages


$ spack tags ecp ecp-apps ecp-proxy-app
ecp:
    amr-wind          er       lammps     nekrs      qmcpack  shuffile
    axl               exawind  latte      nektools   rankstr  spath
    chombo            filo     nalu-wind  nwchem     redset   veloc
    ecp-data-vis-sdk  kvtree   nek5000    parsplice  rmgdft

ecp-apps:
    amr-wind  exawind  latte      nek5000  nektools  parsplice  rmgdft
    chombo    lammps   nalu-wind  nekrs    nwchem    qmcpack

ecp-proxy-app:
    amg                ember      macsio   miniqmc   nekbone     swfft
    candle-benchmarks  examinimd  miniamr  minitri   picsarlite  thornado-mini
    ecp-proxy-apps     laghos     minife   minivite  sw4lite     xsbench


$ spack tags -i radiuss ecp
ecp:
    No installed packages

radiuss:
[...]

PR #32016, opened 2022-08-09
spack list: add --tag flag

### LINKED ISSUE #31856: spack list: add --tag flag
### Summary

`spack find` has a `--tag` flag that lets you filter packages containing a certain tag. We should add the same flag to `spack list`.

### Rationale

A user may want to list _all_ packages belonging to a certain category, not just _installed_ packages.

### Description

We just need to add a `--tag` flag to `spack list`, should be straightforward to copy this setting from `spack find` and add tests.

### Additional information

We're planning to add an "ml" tag to all machine learning packages. We would like a way to list these packages so that we can ensure they all have common features (maintainers, post-install tests, etc.).

### General information

sparkyniner:
Hello!, I've modified list.py and added functionality for --tag but I coudn't get around adding tests.

Closes #31856


lib/spack/spack/cmd/list.py
```diff
@@ -72,6 +72,7 @@ def setup_parser(subparser):
         default=False,
         help="include virtual packages in list",
     )
+    arguments.add_common_arguments(subparser, ["long", "very_long", "tags"])
```
adamjstewart: long and very long aren't needed here, only tags.

lib/spack/spack/cmd/list.py
```diff
@@ -116,6 +117,11 @@ def match(p, f):
 
     return sorted(pkgs, key=lambda s: s.lower())
 
+     # If tags have been specified on the command line, filter by tags
+    if args.tags:
+        packages_with_tags = spack.repo.path.packages_with_tags(*args.tags)
+        results = [x for x in results if x.name in packages_with_tags]
```
adamjstewart: Since this is after the return statement, it is never executed

sparkyniner: Yupp! , made necessary changes

lib/spack/spack/cmd/list.py
```diff
@@ -113,9 +114,16 @@ def match(p, f):
                 return f.match(p)
 
         pkgs = [p for p in pkgs if any(match(p, f) for f in res)]
+        
+    # If tags have been specified on the command line, filter by tags
+    if args.tags:
+        packages_with_tags = spack.repo.path.packages_with_tags(*args.tags)
+        results = [x for x in results if x.name in packages_with_tags]
```
adamjstewart:
```suggestion
        pkgs = [p for p in pkgs if p.name in packages_with_tags]    
```
The results variable isn't defined or used anywhere in this function

lib/spack/spack/cmd/list.py
```diff
@@ -72,6 +72,7 @@ def setup_parser(subparser):
         default=False,
         help="include virtual packages in list",
     )
+    arguments.add_common_arguments(subparser, ["tags"])
```
adamjstewart, line 76: You'll also need to add the following import at the top:
```python
import spack.cmd.common.arguments as arguments
```

sparkyniner:
removed results variable and added import statement at top

lib/spack/spack/cmd/list.py
```diff
@@ -114,8 +117,15 @@ def match(p, f):
 
         pkgs = [p for p in pkgs if any(match(p, f) for f in res)]
 
+    # If tags have been specified on the command line, filter by tags
+    if args.tags:
+        packages_with_tags = spack.repo.path.packages_with_tags(*args.tags)
```
adamjstewart: This line was accidentally deleted:
```python
pkgs = [p for p in pkgs if p.name in packages_with_tags]  
```

adamjstewart: Also, looks like it should be `p`, not `p.name`

lib/spack/spack/cmd/list.py
```diff
@@ -114,8 +117,16 @@ def match(p, f):
 
         pkgs = [p for p in pkgs if any(match(p, f) for f in res)]
 
+    # If tags have been specified on the command line, filter by tags
+    pkgs = [p for p in pkgs if p.name in packages_with_tags]
```
adamjstewart: Needs to be in the if-statement, `packages_with_tags` hasn't been defined yet

lib/spack/spack/cmd/list.py
```diff
@@ -114,8 +117,17 @@ def match(p, f):
 
         pkgs = [p for p in pkgs if any(match(p, f) for f in res)]
 
+    # If tags have been specified on the command line, filter by tags
+    if args.tags:
+        pkgs = [p for p in pkgs if p in packages_with_tags] 
+        packages_with_tags = spack.repo.path.packages_with_tags(*args.tags)
```
adamjstewart: Line order needs to be switched, `packages_with_tags` isn't defined yet

sparkyniner:
Hello!, is there anything else I need to change or edit?

adamjstewart:
This will need to pass all unit tests. The style tests are going to fail, but this can be fixed using:

@spackbot fix style

Also, this will need tests before it can be merged.

sparkyniner:
Yupp got it.

adamjstewart:
Also, run `spack commands --update-completion` to update the tab completion scripts.

sparkyniner:
@spackbot spack commands --update-completion

sparkyniner:
@spackbot update completion

sparkyniner:
@spackbot commands

sparkyniner:
Where exactly should I run this command? In my local machine after compiling all the code?

adamjstewart:
You can run `spack commands` on your local machine. Everything is Python so there's nothing to compile.

sparkyniner:
Yup, done it. It wasn't pushing earlier for some reason but I succeeded on the second try.

sparkyniner:
Seems like all tests were successful except one

adamjstewart:
Yes, the failing coverage test means that your changes aren't being tested by any unit tests. You'll need to add tests in [`lib/spack/spack/test/cmd/list.py`](https://github.com/spack/spack/blob/develop/lib/spack/spack/test/cmd/list.py) to make sure that your feature actually works.

sparkyniner:
I'm not really sure if I've done this properly. I added the test tag function from find.py to list.py

lib/spack/spack/test/cmd/list.py
```diff
@@ -84,3 +84,12 @@ def test_list_update(tmpdir, mock_packages):
...
         assert f.read() == "empty\n"
+        
+def test_tag1(parser, specs):
+
+    args = parser.parse_args(["--tag", "tag1"])
+    spack.cmd.find.find(parser, args)
+
+    assert len(specs) == 2
+    assert "mpich" in [x.name for x in specs]
+    assert "mpich2" in [x.name for x in specs]
```
adamjstewart: This tests `spack find`, not `spack list`.
```suggestion
    output = list("--tag", "tag1")
    assert "mpich" in output
    assert "mpich2" in output
```

sparkyniner:
Committed your suggested changes!

adamjstewart:
@tldahlgren can you review this PR? I opened #31856 because I didn't know how to list packages with a certain tag. I just realized you created a separate `spack tags` command in #26136 with the same feature but removed the `--tag` flag from `spack list`. Is there any reason not to support the flag in both locations? I think most users (including myself) don't know that `spack tags` exists and first look at `spack list` to have this feature.

tldahlgren:
> @tldahlgren can you review this PR? I opened #31856 because I didn't know how to list packages with a certain tag. I just realized you created a separate `spack tags` command in #26136 with the same feature but removed the `--tag` flag from `spack list`. Is there any reason not to support the flag in both locations? 

Don't recall why I removed it off-hand.  I'm all for aliases to support "advanced"/experienced users and more intuitive (:wink:) for people ramping up on tools as long as the functional code is shared by the commands.

> I think most users (including myself) don't know that `spack tags` exists and first look at `spack list` to have this feature.

I'll take a look.

tldahlgren:
@spackbot fix style

tldahlgren:
@spackbot fix style

tldahlgren:
@sparkyniner Can you please create a PR-specific branch for your work?  Using `develop` makes it harder to pull and test your changes.

lib/spack/spack/cmd/list.py
```diff
@@ -114,6 +116,11 @@ def match(p, f):
 
         pkgs = [p for p in pkgs if any(match(p, f) for f in res)]
 
+    # If tags have been specified on the command line, filter by tags
+    if args.tags:
+        packages_with_tags = spack.repo.path.packages_with_tags(*args.tags)
+        pkgs = [p for p in pkgs if p in packages_with_tags]
+
```
tldahlgren: The name of this function is `filter_by_name`, which makes adding the additional filter by tags seem out of place.  I suggest moving this code to the `list` method *before* the `if args.update:` statement.

adamjstewart:
@tldahlgren it's still possible to do with a develop branch PR:
```console
$ git remote add sparkyniner git@github.com:sparkyniner/spack.git
$ git fetch sparkyniner
$ git checkout sparkyniner/develop
```
But yeah, creating a separate branch for each PR is recommended (but not required). @sparkyniner see https://spack.readthedocs.io/en/latest/contribution_guide.html#branching for instructions. Without branches, it's impossible to contribute more than one PR at the same time, and it becomes difficult to update your fork to match the upstream fork.

tldahlgren:
> @tldahlgren it's still possible to do with a develop branch PR:
> 
> ```
> $ git remote add sparkyniner git@github.com:sparkyniner/spack.git
> $ git fetch sparkyniner
> $ git checkout sparkyniner/develop
> ```
> 
> But yeah, creating a separate branch for each PR is recommended (but not required). @sparkyniner see https://spack.readthedocs.io/en/latest/contribution_guide.html#branching for instructions. Without branches, it's impossible to contribute more than one PR at the same time, and it becomes difficult to update your fork to match the upstream fork.

As a reviewer, it is fewer commands and no need to add a remote to your clone if the PR uses a non-`develop` branch.

lib/spack/spack/test/cmd/list.py
```diff
@@ -84,3 +84,9 @@ def test_list_update(tmpdir, mock_packages):
     assert update_file.exists()
     with update_file.open() as f:
         assert f.read() == "empty\n"
+
+
+def test_tag1(parser, specs):
```
tldahlgren: All you should need here is the mock packages fixture:

```suggestion
def test_tag1(mock_packages):
```

tldahlgren (changes requested):
Looks good and works as expected.  Just a couple of suggestions.

sparkyniner:
@tldahlgren, Added your suggestions!

alalazo:
> $ git remote add sparkyniner git@github.com:sparkyniner/spack.git
> $ git fetch sparkyniner
> $ git checkout sparkyniner/develop

@adamjstewart OT (and I strongly agree on not using `develop` in a fork), but to review a PR the simplest command is:
```console
$ git fetch upstream pull/<number>/merge:<your-local-name>
$ git checkout <your-local-name>
```
You can also use "head" instead of "merge" if you don't want the merge commit to be added automatically gy GitHub.

sparkyniner:
@adamjstewart is this ready for a PR?

lib/spack/spack/test/cmd/list.py
```diff
@@ -84,3 +84,15 @@ def test_list_update(tmpdir, mock_packages):
     assert update_file.exists()
     with update_file.open() as f:
         assert f.read() == "empty\n"
+<<<<<<< HEAD
+
+
+def test_tag1(mock_packages):
+=======
+>>>>>>> 5776319f7536add81ac14c940585ee410ca3257e
```
adamjstewart: Looks like a bad rebase, can you remove this diff stuff?

lib/spack/spack/test/cmd/list.py
```diff
@@ -84,3 +84,15 @@ def test_list_update(tmpdir, mock_packages):
...
+
+
+def test_tag1(parser, specs):
+    output = list("--tag", "tag1")
+    assert "mpich" in output
+    assert "mpich2" in output
```
adamjstewart, line 92: We should also test that other packages are not included:
```python
assert "openmpi" not in output
```

sparkyniner: Yupp looking into it asap

adamjstewart:
@spackbot fix style

adamjstewart:
@spackbot fix style

adamjstewart (dismissed):

sparkyniner:
@tldahlgren Hello!, added your requested changes, could you you please re-review?

lib/spack/spack/test/cmd/list.py
```diff
@@ -84,3 +84,14 @@ def test_list_update(tmpdir, mock_packages):
     assert update_file.exists()
     with update_file.open() as f:
         assert f.read() == "empty\n"
+
+
+def test_tag1(mock_packages):
+
+    args = parser.parse_args(["--tag", "tag1"])
+    spack.cmd.find.find(parser, args)
```
tldahlgren: This should be testing the `list` command, not the `find` command.

The style checks are also complaining that `parser`, `specs`, and `output` are undefined here. Plus this is more complicated than necessary.

What you could do is restore the tests that were removed in the `test/cmd/list.py` change at:  https://github.com/spack/spack/pull/26136/files#diff-37d3c92cdb69eb7346b5e92b66546344b7db5fc92deb3e9c2605ad2d5f0f14eb .

tldahlgren (changes requested):
See the issue and suggestion related to `test/cmd/list.py`.

lib/spack/spack/test/cmd/list.py
sparkyniner: Ok, so as I understand it, I'll restore the tests and replace the line "spack.cmd.find.find(parser, args)" with "spack.cmd.find.list(parser, args)" in the test_tag1 function?

adamjstewart: Just use `output = list(...)` like every other test in this file. You had it right previously but removed it in the last commit.

sparkyniner: Done, could you please have a look?

lib/spack/spack/cmd/list.py
```diff
@@ -306,6 +308,11 @@ def list(parser, args):
     # Filter the set appropriately
     sorted_packages = filter_by_name(pkgs, args)
 
+    # If tags have been specified on the command line, filter by tags
+    if args.tags:
+        packages_with_tags = spack.repo.path.packages_with_tags(*args.tags)
+        pkgs = [p for p in pkgs if p in packages_with_tags]
```
tldahlgren: You want to be working with the packages filtered by name at this point *and* make sure you are passing the filtered results to the next filter option.

```suggestion
        sorted_packages = [p for p in sorted_packages if p in packages_with_tags]
```

lib/spack/spack/test/cmd/list.py
```diff
@@ -35,6 +35,9 @@ def test_list_search_description(mock_packages):
     assert "depb" in output
 
 
+   
+
+
```
tldahlgren: Too many extra lines.  There should only be 2 between routines.

```suggestion
```

lib/spack/spack/test/cmd/list.py
tldahlgren: You appear to still have the code for testing `find`, not `list`.  The link I provided defined a single `test_list_tags` that call the feature three times and checks results (see below).  Your test would be equivalent to the first three statements in the example below.

```
def test_list_tags(mock_packages):
     output = list('--tag', 'tag1')
     assert 'mpich' in output
     assert 'mpich2' in output

     output = list('--tag', 'tag2')
     assert 'mpich\n' in output
     assert 'mpich2' not in output

     output = list('--tag', 'tag3')
     assert 'mpich\n' not in output
     assert 'mpich2' in output
```

So the first `output` assignment performs the programmatic equivalent of `spack list --tag tag1` and makes sure the two mock packages are listed.  The second performs the equivalent of `spack list --tag tag2` and the third of `spack list --tag tag3`.

tldahlgren (changes requested):
See comments.

sparkyniner:
Added recommended changes, hope this one is correct

sparkyniner:
@adamjstewart Hello! , are there any other changes I can make?

tldahlgren (dismissed):
LGTM

adamjstewart (dismissed):

sparkyniner:
@spackbot fix style

sparkyniner:
Hello!, any idea why the read the docs build failed?

adamjstewart:
That was a bug that has already been fixed in develop, merging develop should solve it.

adamjstewart (approved):

tldahlgren (approved):

tldahlgren:
Note: temperamental foreground test seems to have failed the `3.10, clingo` check.  Waiting for windows test to complete.

sparkyniner:
How do we fix the failing checks?

tldahlgren:
The test failure isn't with your changes.  Not sure if it will help but I clicked `Update branch` button to see if it resolves the issue.  That means you'll need to pull before making any more changes IF any are needed.

adamjstewart (approved):

PR #32024, opened 2022-08-10
Change python hash to fix broken build in pipeline.

alalazo:
Credit to @trws for getting the diagnosis on what was happening with the broken mesa build on `develop`
Screenshot from 2022-08-10 10-33-37

This is meant as a workaround, not a fix. Changing the hash of Python gives us a fair chance not to incur in a build broken in the same way.

### DIFF
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -336,18 +336,18 @@ class Python(Package):
 
         # Some variants enable multiple modules
         try:
-            python("-c", "import ssl", error=os.devnull)
-            python("-c", "import hashlib", error=os.devnull)
-            variants += "+ssl"
-        except ProcessError:
-            variants += "~ssl"
-
-        try:
             python("-c", "import xml.parsers.expat", error=os.devnull)
             python("-c", "import xml.etree.ElementTree", error=os.devnull)
             variants += "+pyexpat"
         except ProcessError:
             variants += "~pyexpat"
+
+        try:
+            python("-c", "import ssl", error=os.devnull)
+            python("-c", "import hashlib", error=os.devnull)
+            variants += "+ssl"
+        except ProcessError:
+            variants += "~ssl"
 
         # Some modules are version-dependent
         if Version(version_str) >= Version("3.3"):


alalazo:
Link to a broken build for completeness: https://spackpm.slack.com/archives/C02N33GM28H/p1659598130956839

adamjstewart (approved):

adamjstewart:
#32035 may also be sufficient depending on how quickly it gets merged

alalazo:
This one encountered a similar error with `mpfr`, I think I won't try to have silly changes also there :sweat:

trws:
FWIW, #31739 finally built after I added a version bump for sqlite to prod the downstream hashes.  I can't get into the build on the build the world PR #31752 to see what was failing there with gitlab down but if that's something innocuous we may be able to take that, or you could pull the sqlite bump over here.

adamjstewart:
Sounds like we should instead focus on fixing the underlying problem instead of trying to hack hashes. Would love to see a good write-up of what the problem actually is, it's crazy that the installation hash should affect the package.

haampie:
> what the problem actually is

For relocation purposes, Spack rewrites absolute paths to the install prefix in read-only data sections of binaries. Linkers optimize read-only data sections for space: if a string is contained in another you only need to store the longest. So, strings in binaries may be aliasing, and when Spack rewrites a path, it may inadvertently change other strings too.

Given that C-strings always end with a `\0` byte, the optimization only works if a string is the suffix of another string, and Spack install prefixes have the hash at the end. That means that if the binary contains the install prefix without trailing `/`, aliasing may happen with the hash.

Another way to fix it (with GCC on Linux at least) is `-fno-merge-constants` which ensures rodata sections aren't labeled as string sections so that binutils `ld` can't apply the string space optimization. FWIW gold does not apply the optimization by default, it's requires `-Wl,-O2` as a linker flag. But I don't think disabling `-fmerge-constants` is great, since it's an `-O1` optimization and it does more than dealing with constant strings.

IMHO it'd be better if Spack stops blindly modifying binaries and instead works harder to make them _actually_ relocatable. Or if it's really necessary, at least parse the binary so you know what sections you're modifying (which is what patchelf does).

trws:
FWIW I largely agree, but there are some packages, like flux-core, that are completely impossible to relocate correctly after install without this kind of thing. I work on that project, hate that it’s that way, but haven’t had the time to fix the direct embedding of paths. It’s not always feasible without substantial modifications. 

That said, we can do a *lot* better than just `file_contents.replace(old,new)`, which is literally what we currently do. @becker33 has a really nice looking PR in progress to precompute relocations and apply them directly to the file, while my changes are going to rework the replacement to be much safer, and add additional checks and workarounds to actually verify safety much better than we currently do.

In an ideal world we would actually do this with dwarf or other metadata to check every symbol involved for possible issues, and/or use that option by default (seriously think we should), but we have to do something in this vein until and unless we build a portable re-homing solution atriums bwrap or something.

PR #32026, opened 2022-08-10
pin tbb version for salmon & dyninst

rscohn2:
This PR advances the preferred intel-tbb version to the latest release and pins salmon and dyninst at an old tbb.

salmon and dyninst are not compatible with tbb after 2020.3. They are part of spack ci, which blocks me from making the latest tbb release the default. Pin the version for those packages.

I submitted a PR upstream to dyninst: https://github.com/dyninst/dyninst/pull/1292

salmon has already been fixed upstream, but there is no spack release. Simply adding the new version to the salmon spack package does not work because the build procedure for salmon has changed. There is no maintainer for salmon, so hoping @akthoma (original contributor) can do the work or find someone to do it.

### DIFF
--- a/var/spack/repos/builtin/packages/dyninst/package.py
+++ b/var/spack/repos/builtin/packages/dyninst/package.py
@@ -67,7 +67,10 @@ class Dyninst(CMakePackage):
     # libdwarf before that.
     depends_on("libdwarf", when="@:9")
 
+    # findtbb.cmake in the dynist repo does not work with recent tbb
+    # package layout. Need to use tbb provided config instead.
     conflicts("intel-tbb@2021.1:")
+    conflicts("intel-oneapi-tbb@2021.1:")
     depends_on("tbb@2018.6.0:", when="@10.0.0:")
 
     depends_on("cmake@3.4.0:", type="build", when="@10.1.0:")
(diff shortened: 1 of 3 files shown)


hainest (dismissed):
I'll update once our local CMake changes are done. Thanks!

var/spack/repos/builtin/packages/dyninst/package.py
```diff
@@ -67,7 +67,11 @@ class Dyninst(CMakePackage):
     # libdwarf before that.
     depends_on("libdwarf", when="@:9")
 
-    depends_on("tbb@2018.6:", when="@10.0.0:")
+    # findtbb.cmake in the dynist repo does not work with recent tbb
+    # package layout. Need to use tbb provided config instead.
+    conflicts("intel-tbb@2021.1:")
+    conflicts("intel-oneapi-tbb@2021.1:")
+    depends_on("tbb@2018.6.0:", when="@10.0.0:")
```
alalazo, line 74: Where is `tbb` versioned as a virtual dependency? Is this vestigial from having turned `tbb` into a virtual deps?

alalazo (approved):

var/spack/repos/builtin/packages/dyninst/package.py
rscohn2: It was like that before my change so I left it.

I guess this would work if intel-tbb and intel-oneapi-tbb did:
provides('tbb@2018.6.0', when='@2018.6.0')
but all I am doing is
provides('tbb')

PR #31331, opened 2022-06-28
llvm: Variant and dependency cleanup

chuckatkins:
This makes a few significant changes to llvm:

1. Expresses many of the variant-interaction conflicts as `when` clauses on the variants instead of as conflicts.  This creates an easier spec for the concertize to handle and simplifies the resulting spec by only showing variants that are valid to express.
2. Removed the `shared` variant as it wasn't really used
3. Removed unnecessary deps and make explicit the ones that are
4. Cleaned up patch conditions

PR #32006, opened 2022-08-09
llvm: fix LLDB_ENABLE_[N]CURSES

haampie:
@chuckatkins

PR #32028, opened 2022-08-10
llvm: fix ncurses/terminfo builds

aweits:
Set LLVM_ENABLE_TERMINFO with ncurses+terminfo as a function of the lldb variant

@chuckatkins @trws @haampie 

The changes in #31331 left me with llvm build errors like this:
```
/dev/shm/spack/opt/spack/linux-rhel7-skylake_avx512/gcc-9.3.0/binutils-2.38-ecdeglohh4j32jksrdoxqrczpw3tlmi5/bin/ld: lib/liblldbCo
re.a(IOHandlerCursesGUI.cpp.o): undefined reference to symbol 'acs_map@@NCURSES6_TINFO_5.0.19991023'
/dev/shm/spack/opt/spack/linux-rhel7-skylake_avx512/gcc-9.3.0/binutils-2.38-ecdeglohh4j32jksrdoxqrczpw3tlmi5/bin/ld: /dev/shm/spack/opt/spack/linux-rhel7-skylake_avx512/gcc-9.3.0/ncurses-6.2-aifdeiwtq4t2wesmu4ily65svcrezwr2/lib/libtinfo.so.6: error adding symbols: DSO missing from command line
```

### DIFF
--- a/var/spack/repos/builtin/packages/llvm/package.py
+++ b/var/spack/repos/builtin/packages/llvm/package.py
@@ -636,7 +636,7 @@ class Llvm(CMakePackage, CudaPackage):
             projects.append("lldb")
             cmake_args.append(define("LLDB_ENABLE_LIBEDIT", True))
             cmake_args.append(define("LLDB_ENABLE_CURSES", True))
-            if spec['ncurses'].satisfies('+termlib'):
+            if spec["ncurses"].satisfies("+termlib"):
                 cmake_args.append(define("LLVM_ENABLE_TERMINFO", True))
             else:
                 cmake_args.append(define("LLVM_ENABLE_TERMINFO", False))
--- a/var/spack/repos/builtin/packages/llvm/package.py
+++ b/var/spack/repos/builtin/packages/llvm/package.py
@@ -636,7 +636,7 @@ class Llvm(CMakePackage, CudaPackage):
             projects.append("lldb")
             cmake_args.append(define("LLDB_ENABLE_LIBEDIT", True))
             cmake_args.append(define("LLDB_ENABLE_CURSES", True))
-            if "^ncurses+termlib" in spec:
+            if spec['ncurses'].satisfies('+termlib'):
                 cmake_args.append(define("LLVM_ENABLE_TERMINFO", True))
             else:
                 cmake_args.append(define("LLVM_ENABLE_TERMINFO", False))
--- a/var/spack/repos/builtin/packages/llvm/package.py
+++ b/var/spack/repos/builtin/packages/llvm/package.py
@@ -572,7 +572,6 @@ class Llvm(CMakePackage, CudaPackage):
             define("LLVM_REQUIRES_RTTI", True),
             define("LLVM_ENABLE_RTTI", True),
             define("LLVM_ENABLE_EH", True),
-            define("LLVM_ENABLE_TERMINFO", False),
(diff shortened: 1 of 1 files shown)


haampie:
Can we coordinate the curses struggles a bit? Apparently #31870 was supposed to fix something, but there's also a typo #32006 and now this PR.

Note that `LLVM_ENABLE_TERMINFO` also influences `compiler-rt`.

aweits:
Issue replicates for me with both #31870 and #32006 in place.

haampie:
Okay, and what about the compiler-rt w/o lldb situation with this PR?

And testing for `^ncurses` can lead to false positives since it does not imply direct dependency.

aweits:
> Okay, and what about the compiler-rt w/o lldb situation with this PR?

working that now.

> And testing for `^ncurses` can lead to false positives since it does not imply direct dependency.

wasn't quite happy with that anyways, any suggestions?

eugeneswalker:
Thank you for submitting this. This resolves an issue we are facing.

aweits:
> > Okay, and what about the compiler-rt w/o lldb situation with this PR?
> 
> working that now.

+compiler-rt builds both with +lldb and ~lldb succeed with this PR (for me)

trws:
> > Okay, and what about the compiler-rt w/o lldb situation with this PR?
> 
> working that now.
> 
> > And testing for `^ncurses` can lead to false positives since it does not imply direct dependency.
> 
> wasn't quite happy with that anyways, any suggestions?

I think that check will work, since we can only have one in the dag anyway, but do we have a reason not to just require the term lib variant? At least when lldb is enabled it seems reasonable rather than dealing with the lack of it.

aweits:
> I think that check will work, since we can only have one in the dag anyway, but do we have a reason not to just require the term lib variant? At least when lldb is enabled it seems reasonable rather than dealing with the lack of it.

That was what was in place prior to #31331 - I guess I'd like to understand if there was a rationale for switching it out so as to avoid churn. @chuckatkins

eugeneswalker:
> And testing for `^ncurses` can lead to false positives since it does not imply direct dependency.

Is there a way to check if `ncurses` is a direct dependency? @becker33 @scheibelp ?

scheibelp:
>> And testing for ^ncurses can lead to false positives since it does not imply direct dependency.

> Is there a way to check if ncurses is a direct dependency?

Not directly no (nothing like `spec['ncurses'].is_direct('llvm')`) so you have to follow the other package constraints that relate to it:

```
    depends_on("ncurses", when="+lldb")
    ...
        if '+lldb' in spec and spec['ncurses'].satisfies('+termlib'):
            ...
```

eugeneswalker:
Perhaps we can incorporate the changes @scheibelp recommends above and get this merged? This PR in conjunction with https://github.com/spack/spack/pull/32028 fixes our LLVM issues on Ubuntu 22.04.

haampie:
Ok yeah, there wouldn't be false positives given that the `^ncurses+termlib` was tested for in a `+lldb` branch... but this is fine too.

haampie (approved):

eugeneswalker:
Final job of pipeline failed due to an internal server error. Failure was not related to this PR. I just restarted the failed job and am hoping this time it will be OK. 

UPDATE: final job completed OK, pipeline status should be reflected here shortly

PR #28205, opened 2022-01-02
Python: add ensurepip variant

adamjstewart:
Many users are surprised that `spack install python` does not install `pip` (#2065). Some packages actually require a version of Python built with `--ensurepip` (https://github.com/pypa/build/issues/266). This PR adds an `+ensurepip` variant to the Python package. 

I think the only thing we should think about is whether to default to `+ensurepip` or `~ensurepip`. I'm also a bit concerned with how this will interact with #27798. We may need to make the dependency on `py-pip` in `PythonPackage` conditional on the Python variant.

Note that the `ensurepip` module is available in Python even if `--without-ensurepip` is used. The module seems to work fine, but when you create a venv it doesn't contain `pip`. This causes tools like `build` to fail.

Closes #2065

PR #31912, opened 2022-08-03
Fix bootstrapping with Xcode command line tools

adamjstewart:
Apparently #31792 only fixed bootstrapping when using the full Xcode suite, this PR fixes bootstrapping when using only the Xcode command line tools.

Fixes #31889

PR #32035, opened 2022-08-10
Python: remove ensurepip variant

adamjstewart:
Partially reverts #28205
Closes #32024

`--with-ensurepip` is problematic because it causes the Python build to install a copy of setuptools and pip. It's unclear if this copy is downloaded with/without checksum during the build or if it's bundled in the tarball, but we have no way of controlling the version of setuptools/pip that gets installed. Also, all Python packages already have dependencies on pip (and many on setuptools), so Spack will install its own copies anyway. Since these versions may differ, this may result in conflicts when symlinking environment views. It's safer to require users to install pip themselves as a separate package.

@pradyunsg do you know how `--with-ensurepip` actually works? Curious how it decides which version of pip to install and where it comes from.

### DIFF
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -193,7 +193,6 @@ class Python(Package):
     variant("tkinter", default=False, description="Build tkinter module")
     variant("uuid", default=True, description="Build uuid module")
     variant("tix", default=False, description="Build Tix module")
-    variant("ensurepip", default=True, description="Build ensurepip module", when="@2.7.9:2,3.4:")
 
     if not is_windows:
         depends_on("pkgconfig@0.9.0:", type="build")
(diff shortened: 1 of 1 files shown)


adamjstewart:
@spackbot run pipeline

pradyunsg:
> @pradyunsg do you know how `--with-ensurepip` actually works? Curious how it decides which version of pip to install and where it comes from.

It's using a pip wheel bundled as part of CPython itself -- https://github.com/python/cpython/tree/3.10/Lib/ensurepip/_bundled.

alalazo (dismissed):

alalazo:
@spackbot run pipeline

alalazo:
The error was:
```
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
```
Hopefully it's transient, otherwise it might be worth checking if we had the same problem as discussed in #32024

adamjstewart:
Same error this time 😢

tgamblin:
It's not the error from https://github.com/spack/spack/pull/32024; it's the second issue described in https://github.com/spack/spack/pull/31912#issuecomment-1214250873 -- incorrectly setting `PYTHONHOME`

tgamblin (dismissed):

alalazo:
Rebased to get CI past the readthedocs error

alalazo (approved):

tgamblin:
@alalazo if this was passing everything but docs, we should just merge it and not do another pipeline.

tgamblin:
The error here is present on `develop` and has nothing to do with this PR.  It should be fixed by #32186.

PR #32036, opened 2022-08-10
openmpi: allow for ^ucx~thread_multiple

myan-rarcellino-nnl:
I currently cannot build `openmpi@3: fabrics=ucx ^ucx~thread_multiple`:

```console
$ spack spec openmpi@3: fabrics=ucx ^ucx~thread_multiple
==> Error: 'ucx' required multiple values for single-valued variant 'thread_multiple'
    Requested '~thread_multiple' and '+thread_multiple'
```

due to the openmpi recipe requiring `ucx +thread_multiple`:

```python
depends_on("ucx +thread_multiple", when="@3.0.0:")
```

Unfortunately this PR will change the current behavior of the openmpi recipe as `ucx` defaults to `~thread_multiple` and I could not figure out an easy way to maintain the existing behavior while allowing for `^ucx~thread_multiple`.

I would have thought an easy solution would be to propagate `openmpi+/~thread_multiple` to `ucx`, but `thread_multiple` is only applicable to `openmpi@1.5.4:2` and I am not familar enough with openmpi to understand why.

Alternative solutions would be appreciated.

### DIFF
--- a/var/spack/repos/builtin/packages/openmpi/package.py
+++ b/var/spack/repos/builtin/packages/openmpi/package.py
@@ -536,7 +536,6 @@ class Openmpi(AutotoolsPackage, CudaPackage):
     with when("fabrics=ucx"):
         depends_on("ucx")
         depends_on("ucx +thread_multiple", when="+thread_multiple")
-        depends_on("ucx +thread_multiple", when="@3.0.0:")
         depends_on("ucx@1.9.0:", when="@4.0.6:4.0")
         depends_on("ucx@1.9.0:", when="@4.1.1:4.1")
         depends_on("ucx@1.9.0:", when="@5.0.0:")


alalazo:
Waiting for @hppritcha comments on this one.

hppritcha:
Is there a mechanism for alerting spack users about this default change in behavior?  I would expect users of spack-based installs of newer Open MPI versions and using UCX PML within Open MPI would see a plunge in internode performance if they were using MPI_THREAD_MULTIPLE, as Open MPI would switch to using the TCP BTL in those cases for internode messages.

hppritcha (commented):
I'm not in favor of this as implemented owing to a change in default behavior (with potential significant bad performance implications) when building against UCX.  Note UCX and Open MPI are completely separate packages so it makes sense that Open MPI's configury doesn't include options to configure how UCX is built.

I'd be curious to know what the motivation for this PR is?  Was some big performance gain observed for a single threaded MPI application when using a UCX install that was not thread safe?

myan-rarcellino-nnl:
> I'm not in favor of this as implemented owing to a change in default behavior (with potential significant bad performance implications) when building against UCX. Note UCX and Open MPI are completely separate packages so it makes sense that Open MPI's configury doesn't include options to configure how UCX is built.
> 
> I'd be curious to know what the motivation for this PR is? Was some big performance gain observed for a single threaded MPI application when using a UCX install that was not thread safe?

A user that I support requested this change so I will have to get more detail from them on their use case.

That being said, is there a reason that building with `^ucx~thread_multiple` should be prohibited? I would think a user should be able to customize dependencies, say for debugging or prototyping purposes, unless it will actually lead to an error.

Would you be in favor of:

1. Making the `thread_multiple` variant available for all versions of openmpi, with proper documentation for the differences in behavior for `openmpi@:2` and `openmpi@3:`, and setting the default as True for `openmpi@3:`
2. Propagating `thread_multiple` to ucx for all versions of openmpi
3. Wherever `thread_multiple` is currently being used, wrap with if blocks to maintain the current behavior for versions `@1.5.4:2`

myan-rarcellino-nnl:
can confirm ^ucx~thread_multiple was not actually needed in the end. closing

PR #32038, opened 2022-08-10
add support for external-pio build in esmf 8.3.0+

jedwards4b:
Add support for esmf 8.3.0, add a feature to use an external parallelio build and to build the develop branch of esmf from github.   Add support for the sgi mpt mpi library.

### DIFF
--- a/var/spack/repos/builtin/packages/esmf/package.py
+++ b/var/spack/repos/builtin/packages/esmf/package.py
@@ -33,7 +33,7 @@ class Esmf(MakefilePackage):
     variant("external-lapack", default=False, description="Build with external LAPACK support")
     variant("netcdf", default=True, description="Build with NetCDF support")
     variant("pnetcdf", default=True, description="Build with pNetCDF support")
-    variant("xerces", default=True, description="Build with Xerces support")
+    variant("xerces", default=False, description="Build with Xerces support")
     variant(
         "parallelio",
         default=True,
@@ -275,11 +275,7 @@ class Esmf(MakefilePackage):
 
             # When defined, enables the use of Parallel-NetCDF.
             # ESMF_PNETCDF_LIBS will be set to "-lpnetcdf".
-            os.environ["ESMF_PNETCDF"] = "standard"
-
-            # FIXME: determine whether or not we need to set these.
-            # ESMF_PNETCDF_INCLUDE
-            # ESMF_PNETCDF_LIBPATH
+            os.environ["ESMF_PNETCDF"] = "pnetcdf-config"
 
         ##############
         # ParallelIO #
--- a/var/spack/repos/builtin/packages/esmf/package.py
+++ b/var/spack/repos/builtin/packages/esmf/package.py
@@ -18,7 +18,7 @@ class Esmf(MakefilePackage):
     homepage = "https://www.earthsystemcog.org/projects/esmf/"
     url = "https://github.com/esmf-org/esmf/archive/ESMF_8_0_1.tar.gz"
     git = "https://github.com/esmf-org/esmf/"
-    
+
     maintainers = ["climbfuji"]
 
     version("develop", branch="develop")
@@ -34,7 +34,12 @@ class Esmf(MakefilePackage):
     variant("netcdf", default=True, description="Build with NetCDF support")
     variant("pnetcdf", default=True, description="Build with pNetCDF support")
     variant("xerces", default=True, description="Build with Xerces support")
-    variant("external-pio", default=False, description="Build with external parallelio library", when="@8.3:")
+    variant(
+        "external-pio", 
+        default=False, 
+        description="Build with external parallelio library", 
+        when="@8.3:",
+    )
     variant("pio", default=True, description="Enable ParallelIO support")
     variant("debug", default=False, description="Make a debuggable version of the library")
 
--- a/var/spack/repos/builtin/packages/esmf/package.py
+++ b/var/spack/repos/builtin/packages/esmf/package.py
@@ -35,9 +35,9 @@ class Esmf(MakefilePackage):
     variant("pnetcdf", default=True, description="Build with pNetCDF support")
     variant("xerces", default=True, description="Build with Xerces support")
     variant(
-        "external-pio", 
-        default=False, 
-        description="Build with external parallelio library", 
+        "external-pio",
+        default=False,
+        description="Build with external parallelio library",
         when="@8.3:",
     )
     variant("pio", default=True, description="Enable ParallelIO support")
--- a/var/spack/repos/builtin/packages/esmf/package.py
+++ b/var/spack/repos/builtin/packages/esmf/package.py
@@ -35,12 +35,12 @@ class Esmf(MakefilePackage):
     variant("pnetcdf", default=True, description="Build with pNetCDF support")
     variant("xerces", default=True, description="Build with Xerces support")
     variant(
-        "external-pio",
-        default=False,
+        "parallelio",
+        default=True,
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/esmf/package.py
```diff
@@ -31,6 +34,12 @@ class Esmf(MakefilePackage):
     variant("netcdf", default=True, description="Build with NetCDF support")
     variant("pnetcdf", default=True, description="Build with pNetCDF support")
     variant("xerces", default=True, description="Build with Xerces support")
+    variant(
+        "external-pio",
```
alalazo: "External" has another meaning in Spack (something installed on the system, outside of Spack's control). Can you ~change the name to~ reuse the already existing `pio` variant and remove the "internal" build with a vendored library?

alalazo (changes requested):

var/spack/repos/builtin/packages/esmf/package.py
jedwards4b: I was following the pattern already found in the package, there is an "external-lapack" entry at line 33.  I'm not sure I follow the suggestion you are making - can you elaborate?   @climbfuji can you weigh in on the use of "external-" in the esmf package file?

alalazo: > can you elaborate?

If Spack doesn't manage the entire stack it cannot guarantee there won't be conflicts in the DAG, because the software has a dependency that is not modeled. Whenever possible we advise packagers to rely on another Spack package as a dependency, rather than use something vendored. Since this package has an option to choose between vendored and non-vendored for `parallelio`, it means that it can rely on Spack built dependencies. Therefore I think we should remove the option to build with the vendored one to avoid the problematic case.

jedwards4b: I agree but this is not my package, I am not in a position to make that decision.   @rsdunlapiv

jedwards4b: Also this option is only available in esmf > 8.3.0

rsdunlapiv: Yes, we should allow Spack to manage the ParallelIO dependency and the default for ESMF should be to have ParallelIO as a required dependency.  A variant should allow turning it off in which case ESMF_PIO should be OFF.  In Spack environments, I don't think we would want to support ESMF_PIO=internal given comments from @alalazo. 

We should consistently use "ParallelIO" (not PIO) if that is the name of the dependent Spack package and remove the "external-" prefix in the variant.

@theurich @uturuncoglu

uturuncoglu: I have also made couple of changes in the ESMF package and it would be nice to add them to this PR. Here is the comparison,

https://github.com/NOAA-EMC/spack/compare/jcsda_emc_spack_stack...uturuncoglu:spack:hotfix/esmf

I just change the default for xerces variant (it is now False) and change the pnetcdf to use pnetcdf-config (this will allow to appear pnetcdf related definitions in the esmf.mk file).

var/spack/repos/builtin/packages/esmf/package.py
```diff
@@ -17,9 +17,12 @@ class Esmf(MakefilePackage):
 
     homepage = "https://www.earthsystemcog.org/projects/esmf/"
     url = "https://github.com/esmf-org/esmf/archive/ESMF_8_0_1.tar.gz"
+    git = "https://github.com/esmf-org/esmf/"
 
     maintainers = ["climbfuji"]
 
+    version("develop", branch="develop")
```
uturuncoglu, line 24: I think this will not work since the patches are version dependent and at least the gcc patch is failing in my case because it is not needed for ESMF 8.4.0 versions and snapshots. So, we need to find a way to cover those cases.

var/spack/repos/builtin/packages/esmf/package.py
```diff
@@ -31,6 +34,12 @@ class Esmf(MakefilePackage):
     variant("netcdf", default=True, description="Build with NetCDF support")
     variant("pnetcdf", default=True, description="Build with pNetCDF support")
     variant("xerces", default=True, description="Build with Xerces support")
```
uturuncoglu: We could make xerces by default False since ESMf is moving to Info object.

uturuncoglu (changes requested):

var/spack/repos/builtin/packages/esmf/package.py
jedwards4b: @uturuncoglu I have added your changes to the PR.

var/spack/repos/builtin/packages/esmf/package.py
jedwards4b: Are you saying that the gcc patch from this line:
      patch("gcc.patch", when="@:7.0 %gcc@6:")   
is being applied in develop?   That should not happen.

var/spack/repos/builtin/packages/esmf/package.py
uturuncoglu: @jedwards4b thanks. that is great! we still need to find a way for patches in case of using develop. I was thinking that running `make info` before applying patch could be the solution to get used version or using same way that is used in here to return `ESMF_VERSION_STRING_GIT ` `https://github.com/esmf-org/esmf/blob/15670f87e16db8f6159356828c77fa786e27752d/makefile#L77`. BTW, have you tried to build develop branch using spark? I tried yesterday and it was failing due to this issue.

var/spack/repos/builtin/packages/esmf/package.py
jedwards4b: done

var/spack/repos/builtin/packages/esmf/package.py
uturuncoglu: Yes. It will apply to develop and it will fail. At least it was working like that in my case. It was failing because that part of line is already fixed in ESMF. But I think issue is more general in here, it seems that `when` argument in the patch definition is not working correctly with the branches (and maybe with tags which are automatically added - I have away to do it) due to the missing version information.

jedwards4b: As I understand it "develop" is considered > any named version.   But I will test with gcc and try to resolve.

uturuncoglu: Maybe this is not a fix for package.py and needs to be fixed in the spack in more general way. @alalazo Do you have any suggestion or comments about it?

climbfuji:
@jedwards4b @uturuncoglu I recommend looking at https://github.com/NOAA-EMC/spack/blob/jcsda_emc_spack_stack/var/spack/repos/builtin/packages/esmf/package.py - we already added the external PIO dependency for ESMF 8.3.0 in the spack-stack fork of the spack repo, but haven't been able to send the updates back. There is also an open issue in the spack repo that I created that describes the work. We should either update the official spack repo with our changes first, and then add your missing bits and pieces on top, or alternatively you can create a PR to update https://github.com/NOAA-EMC/spack/blob/jcsda_emc_spack_stack/var/spack/repos/builtin/packages/esmf/package.py and then we send the combined changes to the official spack repo?

climbfuji:
And apologies for the late reply, I was on leave entire last week.

jedwards4b:
@climbfuji I have opened https://github.com/NOAA-EMC/spack/pull/152

climbfuji:
Thanks, @jedwards4b ! I'll look at it right away.

jedwards4b:
@climbfuji closing this in anticipation of yours.

PR #32046, opened 2022-08-10
[WRF] Version update to 4.4

t-brown:
Updating the patches needed to compile WRF version 4.4.


alalazo:
@spackbot fix style

tldahlgren:
@spackbot fix style

tldahlgren (commented):
Confirmed sha256.

Is there any reason these patches can't be integrated upstream and referenced as patches here?

t-brown:
I'll ask, however upstream tends to always install `netcdf-c` and `netcdf-fortran` into the same `prefix` and hence do not need the extra `NETCDFF*` variables.

tldahlgren (approved):

MichaelLaufer:
> I'll ask, however upstream tends to always install `netcdf-c` and `netcdf-fortran` into the same `prefix` and hence do not need the extra `NETCDFF*` variables.

For what it is worth, there is a two year old PR (https://github.com/wrf-model/WRF/pull/1269) that fixes the netcdf-c and netcdf-fortran split problem upstream, but the WRF devs are still deliberating if to merge it.

PR #32047, opened 2022-08-10
Add riscv-gnu-toolchain package.

wanlinwang:
Add riscv-gnu-toolchain package, A cross-compilation tool for RISC-V.

### DIFF
--- a/var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
+++ b/var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
@@ -1,0 +1,88 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+import shutil
+import shlex
+from subprocess import PIPE, Popen
+
+from spack import *
+from spack.repo import GitExe
+
+class RiscvGnuToolchain(Package):
+    """A cross-compilation tool for RISC-V.
+    """
+
+    homepage = "https://spack-tutorial.readthedocs.io/"
+    git = "https://github.com/riscv-collab/riscv-gnu-toolchain.git"
+    
+    phases = ['configure', 'build']
+    build_targets = ""
+    if "+Newlib" in self.spec:
+        build_targets = ""
+    elif "+Linux" in self.spec:
+        build_targets = "linux"
+
+    maintainers = ['wanlinwang']
+
+    version('develop', branch='master', submodules=True)
+    version('2022.08.08', tag='2022.08.08', submodules=True)
+
+    # Dependencies:
+    depends_on('autoconf', type=('build', 'link'))
+    depends_on('automake', type=('build', 'link'))
+    depends_on('python', type=('build', 'link'))
+    depends_on('gawk', type=('build', 'link'))
+    depends_on('bison', type=('build', 'link'))
+    depends_on('flex@:2.6.1,2.6.4:', type=('build', 'link'))
+    depends_on('texinfo', type=('build', 'link'))
+    depends_on('patchutils', type=('build', 'link'))
+    depends_on('mpc', type=('build', 'link'))
+    depends_on('gmp', type=('build', 'link'))
+    depends_on('mpfr', type=('build', 'link'))
+    depends_on('gcc', type=('build', 'link'))
+    depends_on('zlib', type=('build', 'link'))
+    depends_on('expat', type=('build', 'link'))
+
+    variant("Newlib", default=True, description="To build the Newlib cross-compiler.")
+    variant("Linux", default=False, description="To build the Linux cross-compiler.")
+
+
+    def configure_args(self):
+        spec   = self.spec
+        prefix = self.prefix
+
+        config_args = [] 
+        return config_args
+
+
+    def configure(self, spec, prefix):
+        """Runs configure with the arguments specified in
+        :meth:`~spack.build_systems.autotools.AutotoolsPackage.configure_args`
+        and an appropriately set prefix.
+        """
+
+        with working_dir(self.stage.source_path, create=True):
+            options = getattr(self, 'configure_flag_args', [])
+            options += ['--prefix={0}'.format(prefix)]
+            options += self.configure_args()
+            configure(*options)
+
+
+    def build(self, spec, prefix):
+        """Makes the build targets specified by
+        :py:attr:``~.AutotoolsPackage.build_targets``
+        """
+
+        # configure and build in one step
+        with working_dir(self.stage.source_path):
+            # modify Makefile not to git init submodules.
+            # /bin/sed -i -r '/^# Rule for auto init submodules/,/git submodule update.*$/d' Makefile
+            cmd = "/bin/sed -i -r '/^# Rule for auto init submodules/,/git submodule update.*$/d' Makefile"
+            p = Popen(shlex.split(cmd))
+            p.communicate()
+
+            params = self.build_targets
+            make(*params)
--- a/var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
+++ b/var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
@@ -3,13 +3,11 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-import os
-import shutil
 import shlex
-from subprocess import PIPE, Popen
+from subprocess import Popen
 
 from spack import *
-from spack.repo import GitExe
+
 
 class RiscvGnuToolchain(Package):
     """A cross-compilation tool for RISC-V.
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,81 @@
...
+
+    maintainers = ["wanlinwang"]
+
+    version("develop", branch="master", submodules=True)
+    version("2022.08.08", tag="2022.08.08", submodules=True)
+
+    # Dependencies:
+    depends_on("pkg-config", type=("build", "link"))
+    depends_on("autoconf", type=("build", "link"))
+    depends_on("automake", type=("build", "link"))
```
becker33: These packages do not include libraries, so they definitely should not be link dependencies.

In general, I think you need to double-check the dependencies list for what dependency types are appropriate.

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,81 @@
...
+    depends_on("python", type=("build", "link"))
+    depends_on("gawk", type=("build", "link"))
+    depends_on("bison", type=("build", "link"))
+    depends_on("flex@:2.6.1,2.6.4:", type=("build", "link"))
+    depends_on("texinfo", type=("build", "link"))
+    depends_on("patchutils", type=("build", "link"))
+    depends_on("mpc", type=("build", "link"))
+    depends_on("gmp", type=("build", "link"))
+    depends_on("mpfr", type=("build", "link"))
+    depends_on("gcc", type=("build", "link"))
```
becker33: Is this using `gcc` as a compiler, or to link against libraries from gcc? If it's the former, this is not the right way of doing so.

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,81 @@
...
+    depends_on("zlib", type=("build", "link"))
+    depends_on("expat", type=("build", "link"))
+
+    conflicts("platform=windows", msg="Windows is not supported.")
+
+    variant("Newlib", default=True, description="To build the Newlib cross-compiler.")
+    variant("Linux", default=False, description="To build the Linux cross-compiler.")
```
becker33: Are these mutually exclusive options? I suspect these should be combined into a single variant (guessing based on the logic for using the options). Something like

```
variant('compiler', default='newlib', values=('newlib', 'linux'), description="Compiler back-end to build")
```
I'm guessing at the details since I'm not familiar with the package itself.

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,81 @@
...
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import shlex
+from subprocess import Popen
+
+from spack.package import *
+
+
+class RiscvGnuToolchain(Package):
```
becker33: This should inherit from `AutotoolsPackage`

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,81 @@
...
+    def configure(self, spec, prefix):
+        """Runs configure with the arguments specified in
+        :meth:`~spack.build_systems.autotools.AutotoolsPackage.configure_args`
+        and an appropriately set prefix.
+        """
+        with working_dir(self.stage.source_path, create=True):
+            options = getattr(self, "configure_flag_args", [])
+            options += ["--prefix={0}".format(prefix)]
+            options += self.configure_args()
+            configure(*options)
```
becker33: This should be unnecessary when you inherit from `AutotoolsPackage`

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,81 @@
...
+        :py:attr:``~.AutotoolsPackage.build_targets``
+        """
+        # configure and build in one step
+        with working_dir(self.stage.source_path):
+            # modify Makefile not to git init submodules.
+            cmd = "/bin/sed -i -r \
+            '/^# Rule for auto init submodules/,/git submodule update.*$/d' \
+            Makefile"
+            p = Popen(shlex.split(cmd))
+            p.communicate()
```
becker33, line 60:
```suggestion
            p.wait()
            p.communicate()
```

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,81 @@
...
+
+            params = ""
+            if "+Newlib" in self.spec:
+                params = ""
+            elif "+Linux" in self.spec:
+                params = "linux"
```
becker33:
```suggestion
            params = ""
            if compiler=linux in spec:
                params.append('linux')
```
This is assuming the variant suggestion I gave above, but even if you don't make that change, the params variable needs to be an iterable of strings, rather than a single string. The version you've written will send `'l'`, `'i'`, `'n'`, `'u'`, and `'x'` as separate arguments.

becker33 (changes requested):

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,70 @@
...
+
+    maintainers = ["wanlinwang"]
+
+    version("develop", branch="master", submodules=True)
+    version("2022.08.08", tag="2022.08.08", submodules=True)
+
+    # Dependencies:
+    depends_on("pkg-config", type=("build"))
+    depends_on("autoconf", type=("build"))
+    depends_on("automake", type=("build"))
```
becker33: If `configure` is distributed with the package, then these aren't necessary. If it's not, then you need an `autoreconf` phase added to the package (there's default logic for that).

For many packages, they use `depends_on('autoconf', when='@main:', type='build')` because they only need to autoreconf for dev versions.

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,70 @@
...
+
+    phases = ["configure", "build"]
+
+    maintainers = ["wanlinwang"]
+
+    version("develop", branch="master", submodules=True)
+    version("2022.08.08", tag="2022.08.08", submodules=True)
+
+    # Dependencies:
+    depends_on("pkg-config", type=("build"))
```
becker33: All of the build-only dependencies should be `type="build"` instead of `type=("build")`, since the latter is parentheses not a tuple. You could do `type=("build",)`, but there's no need to be so verbose.

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,70 @@
...
+    depends_on("mpfr", type=("build"))
+    depends_on("zlib", type=("build", "link"))
+    depends_on("expat", type=("build", "link"))
+    depends_on("bzip2", type=("build"))
+    depends_on("gmake@4.3:", type=("build"))
+
+    conflicts("platform=windows", msg="Windows is not supported.")
+
+    variant(
+        "compiler-type",
```
becker33: We require multi-word variant names to use `_` instead of `-`

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,70 @@
...
+            # modify Makefile not to git init submodules.
+            cmd = "/bin/sed -i -r \
+            '/^# Rule for auto init submodules/,/git submodule update.*$/d' \
+            Makefile"
+            p = Popen(shlex.split(cmd))
+            p.wait()
+            p.communicate()
+
+            params = []
+            if self.spec.variants["compiler-type"].value == "linux":
```
becker33:
```suggestion
            if self.spec.satisfies("compiler_type=linux"):
```

var/spack/repos/builtin/packages/riscv-gnu-toolchain/package.py
```diff
@@ -0,0 +1,70 @@
...
+from spack.package import *
+
+
+class RiscvGnuToolchain(AutotoolsPackage):
+    """A cross-compilation tool for RISC-V."""
+
+    homepage = "https://spack-tutorial.readthedocs.io/"
+    git = "https://github.com/riscv-collab/riscv-gnu-toolchain.git"
+
+    phases = ["configure", "build"]
```
becker33: I'm pretty sure this needs an `install` phase (which probably means you can leave it at the default for `AutotoolsPackage`). How are the built artifacts supposed to get from the build stage to the install prefix?

becker33 (changes requested):

becker33 (dismissed):

becker33 (approved):

PR #32032, opened 2022-08-10
perl-fth: adding version 0.527

cessenat:
initmak.pl compiles the Version.F90 file on mkversion.pl status.
This is needed for build of multiple exes in parallel.

PR #32049, opened 2022-08-11
keepassxc: as an option do not build documentation

cessenat:
This avoids the dependency to asciidoctor and thus to ruby.

### DIFF
--- a/var/spack/repos/builtin/packages/keepassxc/package.py
+++ b/var/spack/repos/builtin/packages/keepassxc/package.py
@@ -65,11 +65,8 @@ class Keepassxc(CMakePackage):
             "-DKEEPASSXC_BUILD_TYPE=Release",
             "-DCMAKE_INSTALL_DATADIR=%s" % join_path(spec.prefix, "share"),
         ]
-        if "+autotype" in spec:
-            args.append("-DWITH_XC_ALL=ON")
-        else:
-            args.append("-DWITH_XC_ALL=OFF")
-        args.append(self.define_from_variant("WITH_XC_DOCS", "doc"))
+        args.append(self.define_from_variant("WITH_XC_ALL", "autotype"))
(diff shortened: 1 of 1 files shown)


cessenat:
Note for reviewer : as a default, documentation is built - as is the case for linux distribution packages - but I'm not sure it is a good idea since most spack packages as a default do not build the documentation. What do you recommend ?

var/spack/repos/builtin/packages/keepassxc/package.py
```diff
@@ -68,6 +69,10 @@ def cmake_args(self):
             args.append("-DWITH_XC_ALL=ON")
         else:
             args.append("-DWITH_XC_ALL=OFF")
+        if "+docs" in spec:
+            args.append("-DWITH_XC_DOCS=ON")
+        else:
+            args.append("-DWITH_XC_DOCS=OFF")
```
alalazo:
```suggestion
        args.append(self.define_from_variant("WITH_XC_DOCS", "doc"))
```

alalazo: Can also be applied elsewhere in the recipe.

alalazo (changes requested):

alalazo:
@cessenat Can you also take care of:

> Can also be applied elsewhere in the recipe.

? Should be just the other variant above your modification.

cessenat:
Oh yes, thanks.
BTW the doc option I chose is "docs" as for cmake WITH_XC_DOCS.
Some package use doc other docs...

alalazo:
> Some package use doc other docs...

True, we don't have a de facto standard for that - so I guess it's up to you :wink:

alalazo (approved):

cessenat:
Hi @alalazo can you merge to develop ?
Same for PR #32032 ?
Thanks.

PR #32053, opened 2022-08-11
pwgen: new package, a password generator

cessenat:
I like this command line password generator ; using keepassxc is better but requires graphic tools.


var/spack/repos/builtin/packages/pwgen/package.py
```diff
@@ -0,0 +1,26 @@
...
+
+    version("2.08", sha256="dab03dd30ad5a58e578c5581241a6e87e184a18eb2c3b2e0fffa8a9cf105c97b")
+
+    depends_on("coreutils", type="build")
+
+    def build(self, spec, prefix):
+        make()
+
+    def install(self, spec, prefix):
+        make("install")
```
alalazo: I think this is the default behavior, so we can remove the methods:
```suggestion
```

alalazo (changes requested):

cessenat:
Changes done.

alalazo (approved):

PR #32058, opened 2022-08-11
update dftd4 package: adding a variant for compatibility layer for the 2.5.x API

hpcnpatel:
To use dftd4 in Vasp the compatibility layer for the 2.5.x API has to be enabled with -Dapi_v2=true (meson) or -DWITH_API_V2=ON (CMake). The Spack package file does not provide an option to pass this option. In my view implementing it as a variant is more helpful.


awvwgk:
Can you add the variant to Vasp as well which makes use of `dftd4 +api_v2`?

Also, you have a lot of unrelated commits in your PR, can you rebase your branch?

PR #32060, opened 2022-08-11
Update dftd4 & vasp package files

hpcnpatel:
To use dftd4 in Vasp the compatibility layer for the 2.5.x API has to be enabled with -Dapi_v2=true (meson) or -DWITH_API_V2=ON (CMake). The Spack package file does not provide an option to pass this option. In my view implementing it as a variant is more helpful.

### DIFF
--- a/.github/workflows/setup_git.ps1
+++ b/.github/workflows/setup_git.ps1
@@ -4,10 +4,6 @@ git config --global user.email "spack@example.com"
 git config --global user.name "Test User"
 git config --global core.longpaths true
 
-# See https://github.com/git/git/security/advisories/GHSA-3wp6-j8xr-qw85 (CVE-2022-39253)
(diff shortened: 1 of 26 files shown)


hpcnpatel:
As @awvwgk suggested, I have added a variant to Vasp as well which makes use of dftd4 +api_v2.

hpcnpatel:
I am not sure what should be the default (True or False) of dftd4 as a variant in the vasp package.py file.

awvwgk:
> I am not sure what should be the default (True or False) of dftd4 as a variant in the vasp package.py file.

Probably disabled by default, make sure to patch the Vasp Makefile in case the `+dftd4` variant is enabled (see instructions at https://dftd4.readthedocs.io/en/latest/recipe/vasp.html).

PR #30738, opened 2022-05-19
Allow packages with multiple build-systems (separate classes)

alalazo:
fixes #12736 (in this PR the `run_after` decorator has an optional `when=` argument)
closes #30411

This PR extends the DSL that can be used in packages to allow declaring that a package uses different build-systems under different conditions. The main design decision both in here and in #30411 is to require each and every spec to have a `build_system` single valued variant. This variant can be used in many contexts to manipulate (force, query, display etc.) the build-system associated with a concrete spec.

## Expected backward incompatibilities

-  ~Methods decorated with `@when` that expect the base case to be present in the build-system base class~
-  ~Methods using `super` to access build properties of the package, or expecting some of the methods that moved to be present in the `Package` class hierarchy~
- ~Packages defining a custom `phases` attribute~

## Pros

- This PR moves all the build behavior to separate classes that *have* a package. Doing so makes the overall design less coupled, and `Builder` objects easier to swap in and out.
- Since each builder is a new separate class, there are no name clash issues. The `buildsystem` variant will determine which `Builder` class is instantiated. 

## Cons

- ~This PR is fairly compatible with what we had before but not as compatible as #30411.~
- Moving forward packagers would be required to code multiple classes in `package.py`

## The `buildsystem` directive

To make it easy to declare the build-system, a new directive has been created. This directive is essentially a wrapper around `variant`:

```python
def buildsystem(*values, **kwargs):
     default = kwargs.get('default', None) or values[0]
     return variant(
         'buildsystem',
         values=tuple(values),
         description='Build-systems supported by the package',
         default=default,
         multi=False
     )
```

For packages with a single buildsystem, the directive is already used in base classes - so no modifications are needed for packages in that sense.

## Decorators to customize phases

There will be only a single `run_after` and a single `run_before` decorator, as it is the case currently. These decorators can be used within each specific builder class:
```python
class OpenJpeg(CMakePackage):
	...
	
class CMakeBuilder(BaseCMakeBuilder):
	def build(self, pkg, spec, prefix):
		pass
	
	@run_after('build')
	def foo(self):
		pass
```
There is no restriction on the choice of the names that can be used for buildsystems, e.g. we can have `buildsystem=cmake`. This is because we don't need to have separate decorators, since the partitioning of information is implemented by the class design.

## Dynamic adapter for backward compatibility

To obtain a reasonable degree of backward compatibility, if the `package.py` doesn't contain specializations of the base builder classes, an adapter class is constructed on the fly. This class tries to, in order:
[...]

PR #1186, opened 2016-07-07
do_install : allow for an arbitrary number of phases

alalazo:
**TLDR** : have a look at the `<name>/package.py` files to have an idea of what this PR does :smile: 

##### Description
This PR allows sub-classes of `PackageBase` (ex `Package`) to have an arbitrary number of phases.

It also add a convenient syntax to register arbitrary methods as `preconditions` or `sanity_checks`, meaning that they will be run **before** or **after** the phases they are bound to. 

In the end it should give more flexibility for further development of features like the ones discussed in #169 or #1305 and reduce the boiler-plate code needed in many packages.

A few packages have been converted to exemplify some of the benefits of this refactoring.

##### Modifications
- [x] `Package` has been renamed `PackageBase`, it has logic to be extended by subclasses that are tailored to specific build systems
- [x] decorator syntax to register preconditions and sanity checks to a given phase
- [x] added `Package` : simple subclass having only an `install` phase (backward-compatible)
- [x] added `EditableMakefile` : examples in `astyle/package.py`
- [x] added `AutotoolsPackage` : examples in `gmp/package.py` and others
- [x] added `CMakePackage` : examples in `openjpeg/package.py` and others
- [x] `log_output` has been turned into a "double" context manager : the external context spawns a daemon that wait for things to write to a file, the internal plug `stderr` and `stdout` to that file
- [x] all the changes that have been extracted in #1228 
- [ ] unit tests for the new feature
- [x] brief documentation of the new subclasses

PR #32068, opened 2022-08-11
Use property instead of methods in PythonPackage

alalazo:
This change helps maintaining the consistency in the signature of methods that are used during build. Currently we have:
1. `phases` which accept `(self, spec, prefix)` mainly because of historical reasons
2. Other methods which accept only `(self)`, like e.g. `cmake_args`
3. Properties

Keeping consistency helps in #30738 because we forward functions for packages coded with the current style.

Modifications:
- [x] Change `config_settings`, `install_options` and `global_options` to be property
- [x] Modify all the Python packages in the built-in repository

### DIFF
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -156,7 +156,6 @@ class PythonPackage(PackageBase):
         """
         return self.stage.source_path
 
-    @property
     def config_settings(self):
         """Configuration settings to be passed to the PEP 517 build backend.
 
@@ -171,7 +170,6 @@ class PythonPackage(PackageBase):
         """
         return {}
 
-    @property
     def install_options(self):
         """Extra arguments to be supplied to the setup.py install command.
 
@@ -184,7 +182,6 @@ class PythonPackage(PackageBase):
         """
         return []
 
-    @property
     def global_options(self):
         """Extra global options to be supplied to the setup.py call before the install
         or bdist_wheel command.
@@ -203,7 +200,7 @@ class PythonPackage(PackageBase):
 
         args = PythonPackage._std_args(self) + ["--prefix=" + prefix]
 
-        for key, value in self.config_settings.items():
+        for key, value in self.config_settings().items():
             if spec["py-pip"].version < Version("22.1"):
                 raise SpecError(
                     "'{}' package uses 'config_settings' which is only supported by "
@@ -211,9 +208,9 @@ class PythonPackage(PackageBase):
                     '    depends_on("py-pip@22.1:", type="build")'.format(spec.name)
                 )
             args.append("--config-settings={}={}".format(key, value))
-        for option in self.install_options:
+        for option in self.install_options():
             args.append("--install-option=" + option)
-        for option in self.global_options:
+        for option in self.global_options():
             args.append("--global-option=" + option)
 
         if self.stage.archive_file and self.stage.archive_file.endswith(".whl"):
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -156,7 +156,8 @@ class PythonPackage(PackageBase):
         """
         return self.stage.source_path
 
-    def config_settings(self, spec, prefix):
+    @property
+    def config_settings(self):
         """Configuration settings to be passed to the PEP 517 build backend.
 
         Requires pip 22.1+, which requires Python 3.7+.
@@ -170,7 +171,8 @@ class PythonPackage(PackageBase):
         """
         return {}
 
-    def install_options(self, spec, prefix):
+    @property
+    def install_options(self):
         """Extra arguments to be supplied to the setup.py install command.
 
         Args:
@@ -182,7 +184,8 @@ class PythonPackage(PackageBase):
         """
         return []
 
-    def global_options(self, spec, prefix):
+    @property
+    def global_options(self):
         """Extra global options to be supplied to the setup.py call before the install
         or bdist_wheel command.
 
@@ -200,7 +203,7 @@ class PythonPackage(PackageBase):
 
         args = PythonPackage._std_args(self) + ["--prefix=" + prefix]
 
-        for key, value in self.config_settings(spec, prefix).items():
+        for key, value in self.config_settings.items():
             if spec["py-pip"].version < Version("22.1"):
                 raise SpecError(
                     "'{}' package uses 'config_settings' which is only supported by "
@@ -208,9 +211,9 @@ class PythonPackage(PackageBase):
                     '    depends_on("py-pip@22.1:", type="build")'.format(spec.name)
                 )
             args.append("--config-settings={}={}".format(key, value))
-        for option in self.install_options(spec, prefix):
+        for option in self.install_options:
             args.append("--install-option=" + option)
-        for option in self.global_options(spec, prefix):
+        for option in self.global_options:
             args.append("--global-option=" + option)
 
         if self.stage.archive_file and self.stage.archive_file.endswith(".whl"):
--- a/var/spack/repos/builtin/packages/py-abipy/package.py
+++ b/var/spack/repos/builtin/packages/py-abipy/package.py
@@ -57,10 +57,8 @@ class PyAbipy(PythonPackage):
     depends_on("py-jupyter", type=("build", "run"), when="+ipython")
     depends_on("py-nbformat", type=("build", "run"), when="+ipython")
 
-    def install_options(self, spec, prefix):
-        args = []
-
-        if "+ipython" in spec:
-            args.append("--with-ipython")
-
-        return args
+    @property
+    def install_options(self):
+        if "+ipython" in self.spec:
+            return ["--with-ipython"]
+        return []
--- a/var/spack/repos/builtin/packages/py-abipy/package.py
+++ b/var/spack/repos/builtin/packages/py-abipy/package.py
@@ -57,7 +57,6 @@ class PyAbipy(PythonPackage):
     depends_on("py-jupyter", type=("build", "run"), when="+ipython")
     depends_on("py-nbformat", type=("build", "run"), when="+ipython")
 
-    @property
     def install_options(self):
         if "+ipython" in self.spec:
             return ["--with-ipython"]
--- a/var/spack/repos/builtin/packages/py-arcgis/package.py
+++ b/var/spack/repos/builtin/packages/py-arcgis/package.py
@@ -32,5 +32,4 @@ class PyArcgis(PythonPackage):
     depends_on("py-requests-toolbelt", type=("build", "run"))
     depends_on("py-requests-ntlm", type=("build", "run"))
 
-    def global_options(self, spec, prefix):
-        return ["--conda-install-mode"]
+    global_options = ["--conda-install-mode"]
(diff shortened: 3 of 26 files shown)


adamjstewart:
Spent a long time trying to figure out how to word this amicably, so here goes nothing...

### Current state of consistency

> This change helps maintaining the consistency in the signature of methods that are used during build.

I would disagree that there is any consistency to begin with. 

We used to only have a single `Package` class with a single `install(self, spec, prefix)` method. When you first proposed splitting this into multiple build system classes like `CMakePackage` and `AutotoolsPackage`, you proposed using `self` for all new methods and `self, spec, prefix` for phase methods for backwards compatibility. I proposed using `self, spec, prefix` for all methods. I don't believe we ever came to any agreement on which to standardize on.

If there is any consistency in our current method signatures, it is as follows:

1. Base classes written by @alalazo: `self, spec, prefix` for phases, `self` elsewhere
2. Base classes written by @adamjstewart: `self, spec, prefix` everywhere

Obviously this isn't ideal, method signatures shouldn't depend on who wrote the method, we should standardize on _something_.

### What to standardize on

If I understand correctly, your primary reason for wanting to standardize on `self` whenever possible is that `spec` and `prefix` are attributes of `self` so can be easily accessed as `self.spec` and `self.prefix` instead.

My primary reason for wanting to standardize on `self, spec, prefix` is as follows. For methods like `cmake_args` and `configure_args`, `self` (and `prefix`) are almost never used except when needing to access `spec`. In these methods, `spec` may be used dozens of times, and should be as easy as possible to access. We _could_ add `spec = self.spec` to the beginning of every one of these methods, but that seems kind of silly. Also, for users without OO programming experience, I think having `spec` and `prefix` as separate parameters is much more intuitive. It is extremely difficult to document all possible attributes of `self`.

Probably going to need @tgamblin to weigh in on `self` vs. `self, spec, prefix` since this design decision is above my pay grade 😛 

### Backwards compatibility

With that said, if we do continue to standardize on `self` instead of `self, spec, prefix`, we should think about how to solve the backwards compatibility issue. One suggestion would be to wrap `spec` and `prefix` with DeprecationWarnings such that `self.spec` raises no warnings but `spec` does. Would that be possible? We'll want to deprecate the `spec` and `prefix` parameters sooner rather than later so we have time to update all packages within Spack as well as external package repos.

### This PR

If we do decide to merge this PR, you'll need to update the `PythonPackage` docs as well. Also every other base class I've ever written, because they all do this.

alalazo:
> I would disagree that there is any consistency to begin with.

As far as I can tell only `scons` has methods which are not a phase and accept `(self, spec, prefix)`.[^1]

[^1]: Besides the ones in this PR.

alalazo:
> If I understand correctly, your primary reason for wanting to standardize on self whenever possible is that spec and prefix are attributes of self so can be easily accessed as self.spec and self.prefix instead.

My primary reason to standardize on `self` is to keep the code that deals with automatic backward compatibility in #30738 manageable. That PR allows naturally multiple build-systems for a single package by extracting the logic of the build into a separate class. To be able to reuse current recipes for a high percentage of the packages I wrote meta-code to generate a separate adapter class on the fly, which delegate certain calls to the Package. This class is created and instantiated at run-time, and knows which calls should be delegated because of lists like this:

https://github.com/spack/spack/blob/a7f5e2395b2ca4c6cbd50a8b3976d7e75b51e02f/lib/spack/spack/build_systems/autotools.py#L113-L139

and is based on the assumption that:
1. phases are called with `(self, spec, prefix)`
2. methods get only `self`
3. properties are accessed like an attribute

If we don't standardize on `self` I have to keep track of which call has which signature so to forward the correct arguments, which seems an overkill for 5 methods that, if I searched correctly, are used by 25-30 packages in the builtin repository.

alalazo:
> Base classes written by @adamjstewart: self, spec, prefix everywhere

IDE doesn't lie :stuck_out_tongue: 
Screenshot from 2022-08-11 21-33-20

adamjstewart:
> My primary reason to standardize on `self` is to keep the code that deals with automatic backward compatibility in https://github.com/spack/spack/pull/30738 manageable.

This is a reason to standardize, not a reason to standardize on self, correct?

adamjstewart:
Also, what's the reason to make `config_settings` a property instead of a self method?

alalazo:
> This is a reason to standardize, not a reason to standardize on self, correct?

It's a reason for me to standardize on `self` because all the most used base classes use `self`, like `AutotoolsPackage` or `CMakePackage`. If we have to be so invasive as to change most of the packages I would avoid the adapter code and write separate classes in each recipe. But that probably means that the PR will never get merged.

alalazo:
> Also, what's the reason to make config_settings a property instead of a self method?

No reason, that would be perfectly fine with me to make it a method accepting `self`.

adamjstewart:
I'll leave the `self` vs. `self, spec, prefix` decision up to @tgamblin, let's bikeshed the other parts.

I think we should probably use methods instead of properties for these specific flags. Functionally, they're equivalent to `cmake_args` or `configure_args` which are also methods.

More generally, I've been using methods for anything that often need to access the spec and properties for things that rarely need to access the spec. Is this a good design principle?

Can decorators like `@when()` be used in conjunction with properties? That could be another thing to consider, I'm using that in #32147.

alalazo:
@adamjstewart What would you like to be a method instead of a property? Only `config_settings`?

adamjstewart:
config_settings, install_options, and global_options. May also add a build_options if we need it.

alalazo:
> config_settings, install_options, and global_options. May also add a build_options if we need it.

Turned everything into methods. Let me know if this is fine with you.

adamjstewart:
Still need to update the docs.

Still think it's worth discussing `self` vs. `self, spec, prefix`.

alalazo:
> Still think it's worth discussing self vs. self, spec, prefix.

Do you want to turn every method into something that accepts `self, spec, prefix`? For instance `cmake_args(self, spec, prefix)`?

adamjstewart:
Yes, that would be my preference.

alalazo:
> Yes, that would be my preference

Why? That would introduce a lot of changes in packages and break compatibility with existing custom repository. What would be the benefit to counter that?

adamjstewart:
> That would introduce a lot of changes in packages and break compatibility with existing custom repository.

I could say the same thing about this PR. But we should standardize on something. We either standardize on `self` or `self, spec, prefix`. Both create backwards incompatibilities. With both we can deprecate the current behavior and move towards the preferred behavior slowly.

> What would be the benefit to counter that?

For arguments for `self` and for `self, spec, prefix`, see "What to standardize on": https://github.com/spack/spack/pull/32068#issuecomment-1212334617

alalazo:
> Both create backwards incompatibilities.

Well, here it's 70 lines. How many packages do we have using `CMakePackage` or `AutotoolsPackage`?

adamjstewart:
It's actually much easier to go from `self` to `self, spec, prefix` (one sed command) than it is to go from `self, spec, prefix` to `self` (requires manual inspection). But I don't think ease of transition should factor in to design decisions like this. Spack is still in beta release, if we're going to refactor things it should be sooner rather than later.

"I don't care much for backwards compatibility"
— Adam

tgamblin:
Just looking at this here are some stats for your consideration -- I am not sure what to take from them:

```console
(spackbook):packages> git grep -ho 'self\.\w*'|sort | uniq -c | sort -rn | head -14
   5635 self.spec
   1706 self.prefix
   1159 self.define_from_variant
   1125 self.define
    680 self.compiler
    451 self.stage
    280 self.enable_or_disable
    264 self.build_directory
    231 self.run_tests
    215 self.version
    203 self.run_test
    193 self.with_or_without
     65 self.test_suite
     49 self.projectdir
```

```console
(spackbook):packages> git grep -Pho 'self\.(spec|prefix)\b'|sort | uniq -c | sort -rn | head
   5635 self.spec
   1706 self.prefix
```

```console
(spackbook):packages> git grep -Phoe '(?<!\.)\b(spec|prefix)\b'|sort | uniq -c | sort -rn | head
   9656 spec
   6037 prefix
```

```console
(spackbook):packages> git grep 'spec = self.spec'|wc -l
    581
```

```console
(spackbook):packages> git grep 'prefix = self.prefix'|wc -l
     24
```

becker33:
I agree we should standardize, and I really don't care which -- I thought we were currently standardized on `self, spec, prefix` for phases and `self` for non-phase methods, but it sounds like I was incorrect about that.

adamjstewart:
I would like to go even further and standardize on the same args for both phases and non-phases. Phases almost never make use of `spec` and `prefix`, only non-phases like `cmake_args` and `configure_args`. So our current method signatures are almost backwards of what they should be.

alalazo:
I might have missed some point. We have lot of concerns on #30738 not causing backward incompatibilities in custom repositories coming from users. If I understand correctly, ideally we don't want them to touch their recipes and still be able to use the same repository they were using before, if that PR gets merged. 

The reason for the change in this PR is explained in https://github.com/spack/spack/pull/32068#issuecomment-1212398291 and while it's true it technically introduces possible backward incompatibilities, the impact on the current builtin repository is ~0,4% of packages - I expect it to be similar for custom repos too.

Here, if I get it correctly, we are considering the possibility to change all the signatures in all methods prior to #32068 getting merged. That would impact on most of the packages (70%?80%?) and require changes to user's custom repositories. At that point people would not be able to reuse their repository across Spack versions before and after the change.

If we have to discuss a change in signature I think we should do it in the context of #30738, not in the current repository. That, or drop/relax the constraint on backward compatibility in #30738.

> I thought we were currently standardized on self, spec, prefix for phases and self for non-phase methods, but it sounds like I was incorrect about that.

@becker33 That was my understanding too, since all the methods in build-systems but 4-5 follow this pattern. That includes also all the methods added by @adamjstewart in many classes. I reported QMake, but you can have a look at others too. Honestly, I thought that these methods and Scons were just typos that slipped through (so I was kind of surprised of the discussion that started here).

adamjstewart:
It seems like there are a lot of concerns here about backwards compatibility. Let me address those.
```python
try:
    args = self.cmake_args(spec, prefix)
except TypeError:
    args = self.cmake_args()
    tty.warn("`def cmake_args(self)` is deprecated, use `def cmake_args(self, spec, prefix)` instead")
```
That's it. We run `sed` once on the builtin repo, and anyone who maintains an external repo runs `sed` once, and that's it. We still maintain backwards compatibility while deprecating the historical usage in order to move towards a more consistent API. Or we go the other direction and deprecate `def cmake(self, spec, prefix)`. Either way, I think it makes sense to improve consistency, much as you are trying to do with this PR. My complaint with this PR isn't the consistency, it's what I perceive to be a step in the wrong direction.

> so I was kind of surprised of the discussion that started here

To be fair, the discussion actually started here: https://github.com/spack/spack/pull/1186#issuecomment-232719987. If you read through that discussion, you'll see that I raised the exact same concerns when build system-specific base classes were first introduced. Unfortunately, those concerns were never addressed, and the result is our current inconsistent API.

alalazo:
@adamjstewart Do you propose to change only self methods or also properties?

adamjstewart:
I think methods vs. properties is a separate question. So far, my strategy has been:

1. Methods for things that tend to be dynamic (depends on version, spec, etc.)
2. Properties for things that tend to be static (directories, etc.)

Not sure if this is a good strategy or not, but it seems to be working so far.

adamjstewart:
To clarify the `self` vs. `self, spec, prefix` discussion, here is my personal order of preference:

1. `self, spec, prefix` for all methods
2. `self` for all methods
3. `self, spec, prefix` for phases, `self` for non-phases
4. the current state of inconsistency

So I'm certainly not opposed to all `self`, although I personally prefer all `self, spec, prefix`. But if we're going to start changing things anyway, let's decide on a consistent design and stick with it.

PR #32077, opened 2022-08-11
pasta: updating to latest version

snehring:
Mostly an update, but also the old 'which python2.7' was causing issues where there was not a python2.7.


tldahlgren (approved):
LGTM.  Confirmed sha256.

PR #32078, opened 2022-08-11
py-setuptools: add v64 and v65

adamjstewart:
Successfully installs on macOS 12.5 (arm64) with Python 3.9.13 and Apple Clang 13.1.6.

### DIFF
--- a/var/spack/repos/builtin/packages/py-numpy/package.py
+++ b/var/spack/repos/builtin/packages/py-numpy/package.py
@@ -103,8 +103,7 @@ class PyNumpy(PythonPackage):
     depends_on("python@3.6:3.10", type=("build", "link", "run"), when="@1.19")
     depends_on("python@3.7:3.10", type=("build", "link", "run"), when="@1.20:1.21")
     depends_on("python@3.8:", type=("build", "link", "run"), when="@1.22:")
-    depends_on("py-setuptools", type=("build", "run"))
-    depends_on("py-setuptools@:59", when="@:1.22.1", type=("build", "run"))
+    depends_on("py-setuptools@:59", type=("build", "run"))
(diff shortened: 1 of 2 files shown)


tldahlgren (dismissed):
Confirmed sha256.

adamjstewart:
@spackbot run pipeline

adamjstewart:
GitLab failure seems legit. @rgommers we're seeing build errors with numpy and setuptools 64. Based on the deprecation warnings, we may want to pin to setuptools<60 for numpy until numpy.distutils is removed.

rgommers:
> GitLab failure seems legit. @rgommers we're seeing build errors with numpy and setuptools 64. Based on the deprecation warnings, we may want to pin to setuptools<60 for numpy until numpy.distutils is removed.

Yes please, that is already what we do in numpy itself. setuptools >= 60 is unsupported, so YMMV.

rgommers:
I see `64.0.1` and `64.0.2` are already out though, so maybe it's being resolved.

adamjstewart:
Two major version bumps in the span on 3 days, impressive! Seriously though, I'm surprised how unstable setuptools is, most libraries only change major versions in extreme circumstances once every few years.

adamjstewart:
@spackbot fix style

adamjstewart:
@spackbot fix style

rgommers:
> Seriously though, I'm surprised how unstable setuptools is, most libraries only change major versions in extreme circumstances once every few years.

For context, the setuptools devs have said that they are going to refactor distutils and expose a different (saner) API. Maybe that has finally started. I also saw for example that 65.0 removes complete `distutils.msvc*` modules. That is definitely breaking `numpy.distutils`, and we're not going to keep up with that.

becker33 (approved):

PR #32080, opened 2022-08-11
arpack-ng %oneapi: add -Wno-error=implicit-function-declaration

eugeneswalker:
Add `-Wno-error=implicit-function-declaration` to `arpack-ng %oneapi` via `flag_handler`

FYI @wspear

PR #32067, opened 2022-08-11
mpifileutils %oneapi: add -Wno-error=implicit-function-declaration

eugeneswalker:
Add `-Wno-error=implicit-function-declaration` to `mpifileutils %oneapi` via `flag_handler`

Silences errors like:
```
...
  >> 408     <truncated>/spack-src/src/common/mfu_io.c:1583:18: error: call to
             undeclared function 'lgetxattr'; ISO C99 and later do not support 
            implicit function declarations [-Wimplicit-function-declaration]
     409        ssize_t rc = lgetxattr(path, name, value, size);
     410                     ^
  >> 411     <truncated>/spack-src/src/common/mfu_io.c:1618:18: error: call to
             undeclared function 'getxattr'; ISO C99 and later do not support i
            mplicit function declarations [-Wimplicit-function-declaration]
     412        ssize_t rc = getxattr(path, name, value, size);
...
```

FYI @wspear @rscohn2

PR #30882, opened 2022-05-26
Control Werror by converting to Wno-error

trws:
Using `-Werror` is good practice for development and testing, but causes us a great deal of heartburn supporting multiple compiler versions, especially as newer compiler versions add warnings for released packages.  This PR adds support for suppressing `-Werror` through spack's compiler wrappers.  There are currently three modes for the `flags:keep_werror` setting:

* `none`: (default) cancel all `-Werror`, `-Werror=*` and `-Werror-*` flags by converting them to `-Wno-error[=]*` flags
* `specific`: preserve explicitly selected warnings as errors, such as `-Werror=format-truncation`, but reverse the blanket `-Werror`
* `all`: keeps all `-Werror` flags

These can be set globally in config.yaml, through the config command-line flags, or overridden by a particular package (some packages use Werror as a proxy for determining support for other compiler features).  We chose to use this approach because:

1. removing `-Werror` flags entirely broke *many* build systems, especially autoconf based ones, because of things like checking `-Werror=feature` and making the assumption that if that did not error other flags related to that feature would also work
2. attempting to preserve `-Werror` in some phases but not others caused similar issues
3. The per-package setting came about because some packages, even with all these protections, still use `-Werror` unsafely.  Currently there are roughly 3 such packages known.

PR #32081, opened 2022-08-11
superlu-dist %oneapi: add -Wno-error=implicit-function-declaration

eugeneswalker:
Add `-Wno-error=implicit-function-declaration` to `superlu-dist %oneapi` via `flag_handler`

Silence errors such as:
```
  >> 210    <TRUNCATED>/SRC/sec_structs.c:134:20: error: call to und
            eclared function 'getline'; ISO C99 and later do not support implic
            it function declarations [-Wimplicit-function-declaration]
     211        while ((read = getline(&line, &len, fp)) != -1)
```

FYI @wspear @balay @gchavez2 @liuyangzhuan @pghysels @xiaoye

### DIFF
--- a/var/spack/repos/builtin/packages/superlu-dist/package.py
+++ b/var/spack/repos/builtin/packages/superlu-dist/package.py
@@ -129,6 +129,8 @@ class SuperluDist(CMakePackage, CudaPackage, ROCmPackage):
             flags.append(self.compiler.cxx11_flag)
         if name == "cflags" and "%pgi" not in self.spec:
             flags.append("-std=c99")
+        if name == "cflags" and self.spec.satisfies("%oneapi"):
+            flags.append("-Wno-error=implicit-function-declaration")
         return (None, None, flags)
 
     examples_src_dir = "EXAMPLE"


liuyangzhuan (dismissed):
looks good.

sethrj:
Man this would also be fixed by https://github.com/spack/spack/pull/30882

eugeneswalker:
> Man this would also be fixed by #30882

I've been told that both LLVM and future OneAPI will have `-Wimplicit-function-declaration` by default. In that case, I think we'd see this error even in the absence of `-Werror`? What do you think?

Reference: https://reviews.llvm.org/rG7d644e1215b376ec5e915df9ea2eeb56e2d94626

sethrj:
Oh man, so this is a c99 issue, I thought from the `-W` that it was a warning-as-error 😬

alalazo:
So, after looking at the function prototype for [getline](https://en.cppreference.com/w/c/experimental/dynamic/getline), I'm not sure we want to "fix" it like that :grimacing:

eugeneswalker:
> So, after looking at the function prototype for [getline](https://en.cppreference.com/w/c/experimental/dynamic/getline), I'm not sure we want to "fix" it like that 😬

What do you suggest? I have other PRs pending that would add `-Wno-error=implicit-function-declaration` to flags:
* https://github.com/spack/spack/pull/32067
* https://github.com/spack/spack/pull/32080

and some others if you search for PRs by me. Talking with @becker33 and @rscohn2 on Spack Slack here: https://spackpm.slack.com/archives/C5VL7V81G/p1660318738321609 it seemed like this is the way to go to enable prior builds to be installed with newer compilers.

What do you think?

sethrj:
I agree with @becker33 based on the discussion. The "best" thing to do might be to patch by `#include <stdio.h>` but the most reliable fix will be to as you suggest with the flags.

sethrj (approved):

PR #32082, opened 2022-08-11
tophat: fixing build issue with newer automake versions

snehring:
At some point autoreconf started complaining about multiple uses of the AM_INIT_AUTOMAKE macro.

I opened https://github.com/DaehwanKimLab/tophat/pull/63 in hopes to merge this upstream, but the project doesn't seem terribly active at the moment.

### DIFF
--- a/var/spack/repos/builtin/packages/tophat/package.py
+++ b/var/spack/repos/builtin/packages/tophat/package.py
@@ -29,6 +29,10 @@ class Tophat(AutotoolsPackage):
 
     parallel = False
 
+    def patch(self):
+        # Newer versions of autoconf hate calling AM_INIT_AUTOMAKE twice
+        filter_file(r"^AM_INIT_AUTOMAKE$", "", "configure.ac")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/tophat/package.py
```diff
@@ -29,6 +29,10 @@ class Tophat(AutotoolsPackage):
 
     parallel = False
 
+    def patch(self):
+        # Newer versions of autoconf hate calling AM_INIT_AUTOMAKE twice
```
tldahlgren, line 33: Given this comment, should the `autoconf` version be checked before performing the file filter operation?

snehring: I _think_ calling it once is more correct regardless of version, but I'm not an autotools expert by far. I can hunt for the version where this changed though, if you'd like.

tldahlgren (approved):

PR #32083, opened 2022-08-11
libfabric: new versions

mpokorny:
Update to add new libfabric versions.

Also add a configure argument to prevent building against fabrics that are not explicitly enabled by the variant. A better solution would be to determine package dependencies required to build each of the fabric variants, and list those explicitly. 

Closes #31587


tldahlgren (dismissed):
Confirmed new sha256.  Deferring merge until maintainer has a chance to comment.

var/spack/repos/builtin/packages/libfabric/package.py
```diff
@@ -120,6 +120,7 @@ def configure_args(self):
         else:
             args.append("--with-kdreg=no")
 
+        args.append("--enable-only")
```
rajachan: Was there a particular reason why this was necessary? Libfabric's configure script is reasonable and only builds providers that have everything it requires in the build environment.

var/spack/repos/builtin/packages/libfabric/package.py
```diff
@@ -100,6 +107,14 @@ class Libfabric(AutotoolsPackage):
     depends_on("libtool", when="@develop", type="build")
 
     conflicts("@1.9.0", when="platform=darwin", msg="This distribution is missing critical files")
+    conflicts("fabrics=bgq", when="@:1.14.99")
+    conflicts("fabrics=net", when="@:1.14.99")
+    conflicts("fabrics=opx", when="@:1.14.99")
+    conflicts("fabrics=perf", when="@:1.14.99")
+    conflicts("fabrics=rstream", when="@:1.14.99")
```
rajachan: These two and BGQ have been in the code for a while now.

var/spack/repos/builtin/packages/libfabric/package.py
mpokorny: I needed to add this when building on my laptop, with only the default fabrics enabled. However, I just tried again, and was unable to reproduce the error. Perhaps I could remove this change.

var/spack/repos/builtin/packages/libfabric/package.py
mpokorny: I did not check for the exact version support, and only went back to v1.15.0. If those fabrics were left out because of build problems, then then they should be removed, otherwise one could determine the exact minimum version. I'm no expert, however -- all I want is a more recent version of libfabric.

var/spack/repos/builtin/packages/libfabric/package.py
rajachan: Yeah, let's drop this.

var/spack/repos/builtin/packages/libfabric/package.py
rajachan: They aren't broken but I don't know the semver->provider mapping off the top of my head either, would have to parse through the git log. Why don't we just get the new release in with this PR and bring in the additions separately?

mpokorny: OK, I'll fix up this PR to just add the new versions.

mpokorny:
@tldahlgren does the Spack team have a policy or preference on how to clean up PR branches? That is, can I force push or should I apply new commits?

mpokorny:
Reduced this PR to include only the new versions

rajachan (dismissed):

tldahlgren:
> @tldahlgren does the Spack team have a policy or preference on how to clean up PR branches? That is, can I force push or should I apply new commits?

You can do either as you see fit.  We generally automatically squash PRs that get merged into `develop`.

tldahlgren (dismissed):
Confirmed branch (and new versions).

mpokorny:
@rajachan I just tried rebuilding `libfabric` (no explicit variants) and encountered the error that caused me to add `--enable-only` to the `configure_args()` method.
```
623    libtool: compile:  /home/martin/spack/lib/spack/env/gcc/gcc -DHAVE_CONFIG_H -I. 
            -I./include -D_GNU_SOURCE -D__USE_XOPEN2K8 -DSYSCONFDIR=\"/home/martin/spack/opt
            /spack/linux-pop22-skylake/gcc-11.2.0/libfabric-1.15.1-bfoc34edy5ikfkuwyskjg7drw
            2uamghb/etc\" -DRDMADIR=\"@rdmadir@\" -DPROVDLDIR=\"/home/martin/spack/opt/spack
            /linux-pop22-skylake/gcc-11.2.0/libfabric-1.15.1-bfoc34edy5ikfkuwyskjg7drw2uamgh
            b/lib/libfabric\" -I./prov/sockets/include -I./prov/sockets -I./prov/opx/include
             -I./prov/opx/include -I./prov/opx/include/linux-i386 -I./prov/hook/include -I./
            prov/hook/perf/include -I./prov/hook/hook_debug/include -I./prov/hook/hook_hmem/
            include -I./prov/hook/dmabuf_peer_mem/include -Wall -O2 -DNDEBUG -fvisibility=hi
            dden -Wall -Wundef -Wpointer-arith -MT prov/opx/src/src_libfabric_la-fi_opx_hfi1
            .lo -MD -MP -MF prov/opx/src/.deps/src_libfabric_la-fi_opx_hfi1.Tpo -c prov/opx/
            src/fi_opx_hfi1.c  -fPIC -DPIC -o prov/opx/src/.libs/src_libfabric_la-fi_opx_hfi
            1.o
  >> 624    prov/opx/src/fi_opx_hfi1.c:34:10: fatal error: numa.h: No such file or directory
     625       34 | #include <numa.h>
     626          |          ^~~~~~~~
     627    compilation terminated.
  >> 628    make[1]: *** [Makefile:22008: prov/opx/src/src_libfabric_la-fi_opx_hfi1.lo] Erro
            r 1
```
Doesn't this look like code for the OPX fabric is being built?

rajachan:
Indeed, that looks like an OPX build issue, I will file an issue upstream and have that provider maintainer take a look.  Thanks!

tldahlgren:
@spackbot run pipeline

mpokorny:
@rajachan, given that this issue occurs for v1.15.1 and perhaps some earlier versions, but not v1.15.2rc2, do you have any suggestions about how to handle it in the Spack package? Version dependent patch? Version dependent addition of `--enable-only`? Something else?

mpokorny:
I just confirmed one possible solution: add `opx` to the list of supported fabrics and a conflict for `fabrics=opx` with older libfabric versions (for some value of "older"). This results in the addition of `--enable-opx=no` to the libfabric configure variables by the Spack package file when `opx` is not included in the `fabrics` variant, and then the build completes.

rajachan:
How about adding a `depends_on("numactl", when="fabrics=opx")`... and  also when="@:1.15.1" (with a comment calling it out as a workaround to handle the OPX bug)?

mpokorny:
I don't think that solution is sufficient; for example, with the default value of the `fabrics` variant, `opx` is not mentioned, the `depends_on` would pass, but OPX fabric code would be built anyway.

mpokorny:
I just occurred to me that the following part of `configure_args()` almost works as the `--enable-only` flag
```python
        for fabric in self.fabrics:
            if "fabrics=" + fabric in self.spec:
                args.append("--enable-{0}=yes".format(fabric))
            else:
                args.append("--enable-{0}=no".format(fabric))
```
with the only exception being fabrics in libfabric that are not listed in the package file `fabrics` tuple.

rajachan:
> I don't think that solution is sufficient; for example, with the default value of the fabrics variant, opx is not mentioned, the depends_on would pass, but OPX fabric code would be built anyway.

You are right, I missed that.

> I just confirmed one possible solution: add opx to the list of supported fabrics and a conflict for fabrics=opx with older libfabric versions (for some value of "older"). This results in the addition of --enable-opx=no to the libfabric configure variables by the Spack package file when opx is not included in the fabrics variant, and then the build completes.

This sounds reasonable then.

mpokorny:
Can you tell me the libfabric version in which OPX first appears? If not, I will try to find out myself, but might get it wrong.

tldahlgren (approved):
LGTM

PR #32093, opened 2022-08-12
hwloc: replace 'shared' variant with 'libs'

iarspider:


bgoglin:
I don't see any issue but I am not familiar enough with spack to review with some confidence.

var/spack/repos/builtin/packages/hwloc/package.py
```diff
@@ -74,7 +74,13 @@ class Hwloc(AutotoolsPackage):
         default=(sys.platform != "darwin"),
         description="Support analyzing devices on PCI bus",
     )
-    variant("shared", default=True, description="Build shared libraries")
+    variant(
+        "libs",
+        default="shared,static",
+        values=("shared", "static"),
+        multi=True,
+        description="Build shared libs, static libs or both",
```
tldahlgren, line 82: Is the ability to build both libraries supported in all versions listed in this package or only from some version on?

tldahlgren (approved):
LGTM.  Just have the question about multi-library support.

var/spack/repos/builtin/packages/hwloc/package.py
iarspider: As far as I can tell, for all versions.

haampie: It's more of an autotools thing, the pattern is quite common now in other packages. I'm fine with this change. Whenver a variant changes name, it's good to verify that it's not assumed to exist in things such as `setup_run_environment` etc. In this case not, so it's not a terribly breaking change to rename the variant.

haampie (approved):

PR #32100, opened 2022-08-12
Add new dependencies for rocm variant for py-torch recipe

renjithravindrankannath:
Removed redundant path variables and updated cmake module path for ROCm 5.2

### DIFF
--- a/var/spack/repos/builtin/packages/py-torch/package.py
+++ b/var/spack/repos/builtin/packages/py-torch/package.py
@@ -437,7 +437,6 @@ class PyTorch(PythonPackage, CudaPackage):
             env.set("HIPCUB_PATH", self.spec["hipcub"].prefix)
             env.set("ROCTHRUST_PATH", self.spec["rocthrust"].prefix)
             env.set("ROCTRACER_PATH", self.spec["roctracer-dev"].prefix)
-            env.set("USE_PTHREADPOOL", True)
             if self.spec.satisfies("^hip@5.2.0:"):
                 env.set("CMAKE_MODULE_PATH", self.spec["hip"].prefix.lib.cmake.hip)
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-torch/package.py
```diff
@@ -437,6 +437,10 @@ def enable_or_disable(variant, keyword="USE", var=None, newer=False):
             env.set("HIPCUB_PATH", self.spec["hipcub"].prefix)
             env.set("ROCTHRUST_PATH", self.spec["rocthrust"].prefix)
             env.set("ROCTRACER_PATH", self.spec["roctracer-dev"].prefix)
+            env.set("USE_NCCL", True)
```
adamjstewart: This depends on the value of the `+nccl` variant and is already set below.

var/spack/repos/builtin/packages/py-torch/package.py
```diff
@@ -437,6 +437,10 @@ def enable_or_disable(variant, keyword="USE", var=None, newer=False):
...
+            env.set("USE_PTHREADPOOL", True)
```
adamjstewart: This is already set below for PyTorch 1.6+

renjithravindrankannath: Removed USE_NCCL.

adamjstewart: The above comment was for NCCL, this comment is for PTHREADPOOL

var/spack/repos/builtin/packages/py-torch/package.py
renjithravindrankannath: Removed USE_NCCL.

var/spack/repos/builtin/packages/py-torch/package.py
renjithravindrankannath: Removed PTHREADPOOL as well from the recipe.

srekolam (commented):
LGTM

adamjstewart (approved):

PR #32037, opened 2022-08-10
ncurses: create symlink libcurses -> libncurses; add v6.3

eugeneswalker:
`ncurses` updates:

* add `@6.3`, including logic to accommodate the new expectation that `--with-pkg-config-libdir` specify an existing path
* create symlink for `libcurses.so` -> `libncurses.so` so that packages such as LLVM, which want to link `-lcurses`, will be able to pick up Spack-installed `ncurses`

FYI @haampie @aweits @trws

PR #32103, opened 2022-08-12
bridger: adding ncurses dep

snehring:
Not sure how I haven't hit this before.

Also actually depends on https://github.com/spack/spack/pull/32037 to build, otherwise it'll report that it can't find the curses library.


tldahlgren (approved):
LGTM

tldahlgren:
Oops. May have merged a bit prematurely based on the PR dependency.

snehring:
It's probably fine. I think most people probably have curses available, I just hit these things because my build env is really minimal.

PR #32109, opened 2022-08-12
quantum-espresso: adding some missing build deps

snehring:
the build requires git 2.13 or newer, and my paired down rhel7 env doesn't have that so it would be good to have that explicit. It also requires m4 or the checkouts fail.

### DIFF
--- a/var/spack/repos/builtin/packages/quantum-espresso/package.py
+++ b/var/spack/repos/builtin/packages/quantum-espresso/package.py
@@ -208,6 +208,8 @@ class QuantumEspresso(CMakePackage):
     depends_on("blas")
     depends_on("lapack")
     depends_on("fftw-api@3")
+    depends_on("git@2.13:", type="build")
+    depends_on("m4", type="build")
 
     # CONFLICTS SECTION
     # Omitted for now due to concretizer bug


ye-luo:
Git and m4 are not unique to CMake variant.

snehring:
Right, I'll fix that in a sec.

ye-luo (approved):
LGTM

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -208,6 +208,8 @@ class QuantumEspresso(CMakePackage):
     depends_on("blas")
     depends_on("lapack")
     depends_on("fftw-api@3")
+    depends_on("git@2.13:", type="build")
```
haampie, line 211: What is the lowerbound based on? I'm sure it's OK, but just checking we're not requiring more git builds than necessary. If it's some particular feature/flag, it's worth documenting that briefly

snehring: The CMakeLists.txt specifies that as the minimum.

ye-luo: I don't remember exactly why 2.13 was chosen. It is likely the oldest git that has been verified to work for the necessary submodule feature.

haampie (approved):

PR #32121, opened 2022-08-13
helics: update v3 options and add missing versions

nightlark:
I did test builds with a few different configurations. In particular, I'm not sure if I wrote the `when` version constraints correctly e.g. are `@3.0.0:` vs `@3:` equivalent?

Changes:
- Updates CMake options that were renamed/removed/added in 3.0.0+ releases
- Tries to better capture constraints for compilers/libraries/features that differ between versions
- Adds missing 3.1.0, 3.1.1, and 3.1.2 release versions


var/spack/repos/builtin/packages/helics/package.py
```diff
@@ -100,12 +127,21 @@ def cmake_args(self):
...
+            args.append(from_variant("HELICS_ENABLE_IPC_CORE", "ipc"))
+            args.append(from_variant("HELICS_ENABLE_INPROC_CORE", "inproc"))
+            args.append(from_variant("HELICS_ENABLE_MPI_CORE", "mpi"))
+        else:
+            args.append(from_variant("ENABLE_ZMQ_CORE", "zmq"))
+            args.append(from_variant("ENABLE_TCP_CORE", "tcp"))
+            args.append(from_variant("ENABLE_UDP_CORE", "udp"))
+            args.append(from_variant("ENABLE_IPC_CORE", "ipc"))
+            args.append(from_variant("ENABLE_INPROC_CORE", "inproc"))
+            args.append(from_variant("ENABLE_MPI_CORE", "mpi"))
```
tldahlgren: FWIW The code could be "tightened" with a simple change:

```suggestion
        # Options were renamed in v3
        pre = "HELICS_" if spec.satisfies("@3:") else ""
        args.append(from_variant("{0}ENABLE_ZMQ_CORE".format(pre), "zmq"))
        args.append(from_variant("{0}ENABLE_TCP_CORE".format(pre), "tcp"))
        args.append(from_variant("{0}ENABLE_UDP_CORE".format(pre), "udp"))
        args.append(from_variant("{0}ENABLE_IPC_CORE".format(pre), "ipc"))
        args.append(from_variant("{0}ENABLE_INPROC_CORE".format(pre), "inproc"))
        args.append(from_variant("{0}ENABLE_MPI_CORE".format(pre), "mpi"))
```

tldahlgren (commented):
Confirmed sha256.  Have a recommendation for tightening some of the code.

var/spack/repos/builtin/packages/helics/package.py
tldahlgren: @nightlark Let me know what you want to do with this suggestion and I'll approve and merge.

nightlark: This looks better, thanks!

nightlark:
@tldahlgren I applied the suggested changes. I'm not sure how to fix the readthedocs build error -- it doesn't look like it is related to any changes made?

After looking at the spack readthedoc build history, it seems like the exact same error occurs intermittently on about 1 out of every 8 spack readthedocs builds. For someone who has maintainer access to RTD, there should be a "Rebuild this build" link on the right side near the top of the page at https://readthedocs.org/projects/spack/builds/17809728/ just below where it has build time information and the "View raw" and "View docs" links at, which should be quicker/use less CI resources than adding a new commit that retriggers all of the CI builds.

tldahlgren (approved):
LGTM

tldahlgren:
> @tldahlgren I applied the suggested changes. I'm not sure how to fix the readthedocs build error -- it doesn't look like it is related to any changes made?
> 
> After looking at the spack readthedoc build history, it seems like the exact same error occurs intermittently on about 1 out of every 8 spack readthedocs builds. For someone who has maintainer access to RTD, there should be a "Rebuild this build" link on the right side near the top of the page at https://readthedocs.org/projects/spack/builds/17809728/ just below where it has build time information and the "View raw" and "View docs" links at, which should be quicker/use less CI resources than adding a new commit that retriggers all of the CI builds.

I don't have that ability on RTD.  If `maintainers` could update this PR, there would be a button here that would allow me to do that so usually recommend people perform the update.

PR #32129, opened 2022-08-14
pcsclite: new package, Middleware to access a smart card using SCard …

cessenat:
…API (PC/SC)
Home page https://pcsclite.apdu.fr/
When spack install keepassxc+autotype there is a dependency to PCSC which is provided on Ubuntu by libpcsclite-dev.
For the long story, there is also a need for other packages, that, still on Ubuntu, are provided by gengetopt and libusb-1.0-0-dev
According to the author, there is no need to build pcsc-lite on Darwin

### DIFF
--- a/var/spack/repos/builtin/packages/pcsclite/package.py
+++ b/var/spack/repos/builtin/packages/pcsclite/package.py
@@ -1,0 +1,41 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Pcsclite(AutotoolsPackage):
+    """PCSC lite project
+
+    Middleware to access a smart card using SCard API (PC/SC)."""
+
+    homepage = "https://pcsclite.apdu.fr"
+    url = "https://pcsclite.apdu.fr/files/pcsc-lite-1.9.8.tar.bz2"
+
+    maintainers = ["cessenat"]
+
+    version("1.9.8", sha256="502d80c557ecbee285eb99fe8703eeb667bcfe067577467b50efe3420d1b2289")
+
+    # no libudev/systemd package currently in spack
+    variant("libudev", default=False, description="Build with libudev")
+
+    depends_on("flex", type="build")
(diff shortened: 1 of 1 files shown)


cessenat:
Note for reviewer : there is no need for the bootstrap phase when using a released package.
I do not know how to set the phases accordingly, this is why I always call the bootstrap phase, dummy or not.
It is a bit heavy - but it works.

var/spack/repos/builtin/packages/pcsclite/package.py
```diff
@@ -0,0 +1,54 @@
...
+    depends_on("libusb")
+
+    depends_on("autoconf", type="build")
+    depends_on("autoconf-archive", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+    depends_on("m4", type="build")
+    depends_on("pkgconfig", type="build")
+
+    phases = ["bootstrap", "configure", "build", "install"]
```
alalazo: Can we stick with the `autoreconf` name, and override that?

alalazo (changes requested):

var/spack/repos/builtin/packages/pcsclite/package.py
cessenat: Well, no : without the specific phases for the 'master' release I get errors
`aclocal: error: couldn't open directory 'm4': No such file or directory`
So, this is the reason for this heavy workaround
[spack-build-out.txt](https://github.com/spack/spack/files/9382592/spack-build-out.txt)
[spack-build-env.txt](https://github.com/spack/spack/files/9382593/spack-build-env.txt)

alalazo: Sorry I wasn't clear enough. Why does it matter to call the phase `bootstrap`? You can simply:
```python
def autoreconf(self, spec, prefix):
    pass

@when("@master")
def autoreconf(self, spec, prefix):
    bootstrap = Executable("./bootstrap")
    bootstrap()
```
and remove both `phases` (which is not meant to be overridden) and `force_autoreconf` since you don't use the default autoreconf method.

cessenat: Yes you are right. Commited accordingly. Thanks.

alalazo (approved):

cessenat:
Thanks

PR #32132, opened 2022-08-14
config.yaml:  Adds a few more words about $tempdir.

### LINKED ISSUE #32131: config.yaml:  Improve description of $tempdir
### Summary

Adds a few words into `default/config.yaml` as to how `$tempdir` is generated.

### Rationale

I'm not a native speaker of python, so I lost an hour trying to figure out why most `$<lower-case-variables>` mapped to environment variables but `$tempdir` kept resolving to `/var`.  It's obvious after digging into the source.  Perhaps the additional comments will save someone else a bit of time.

### Description

Small documentation change to `config.yaml` comments.

### Additional information

0.19.0.dev0 (43b4b66cfe7e9ad898f997857ce25cbd6b4498ac)

### General information

rountree:
Adds a description of how $tempdir is generated.

Fixes #32131

### DIFF
--- a/etc/spack/defaults/config.yaml
+++ b/etc/spack/defaults/config.yaml
@@ -41,12 +41,20 @@ config:
   # Recommended options are given below.
   #
   # Builds can be faster in temporary directories on some (e.g., HPC) systems.
-  # Specifying `$tempdir` will ensure use of the default temporary directory
-  # (i.e., ``$TMP` or ``$TMPDIR``).
+  # The string `$tempdir$ will be substituted with the results of a call to
+  # tempfile.gettempdir().  The results will be, in order of decreasing 
+  # preference:
+  #   1) The contents of the $TMPDIR environment variable
+  #   2) The contents of the $TEMP environment variable
+  #   3) The contents of the $TMP environment variable
+  #   4) On Windows:  C:\TEMP, C:\TMP, \TEMP and TMP (in that order)
+  #      On all other platforms: /tmp, /var/tmp and /usr./tmp, in that order.
+  #   5) The current working directory
   #
   # Another option that prevents conflicts and potential permission issues is
   # to specify `$user_cache_path/stage`, which ensures each user builds in their
-  # home directory.
+  # home directory.  `$user_cache_path` maps to the environment variable
+  # $SPACK_USER_CACHE_PATH.
   #
   # A more traditional path uses the value of `$spack/var/spack/stage`, which
   # builds directly inside Spack's instance without staging them in a


etc/spack/defaults/config.yaml
```diff
@@ -41,12 +41,20 @@ config:
   # Recommended options are given below.
   #
   # Builds can be faster in temporary directories on some (e.g., HPC) systems.
-  # Specifying `$tempdir` will ensure use of the default temporary directory
-  # (i.e., ``$TMP` or ``$TMPDIR``).
+  # The string `$tempdir$ will be substituted with the results of a call to
```
tldahlgren, line 44:
```suggestion
  # The string `$tempdir` will be substituted with the results of a call to
```

tldahlgren, line 44:
```suggestion
  # The string `$tempdir` will be substituted with the results of a call to
```

etc/spack/defaults/config.yaml
```diff
@@ -41,12 +41,20 @@ config:
...
+  #   3) The contents of the $TMP environment variable
+  #   4) On Windows:  C:\TEMP, C:\TMP, \TEMP and TMP (in that order)
+  #      On all other platforms: /tmp, /var/tmp and /usr./tmp, in that order.
+  #   5) The current working directory
   #
   # Another option that prevents conflicts and potential permission issues is
   # to specify `$user_cache_path/stage`, which ensures each user builds in their
-  # home directory.
+  # home directory.  `$user_cache_path` maps to the environment variable
+  # $SPACK_USER_CACHE_PATH.
```
tldahlgren, line 57:
```suggestion
  # `$SPACK_USER_CACHE_PATH`.
```

etc/spack/defaults/config.yaml
```diff
@@ -41,12 +41,20 @@ config:
...
-  # (i.e., ``$TMP` or ``$TMPDIR``).
+  # The string `$tempdir$ will be substituted with the results of a call to
+  # tempfile.gettempdir().  The results will be, in order of decreasing 
+  # preference:
+  #   1) The contents of the $TMPDIR environment variable
+  #   2) The contents of the $TEMP environment variable
+  #   3) The contents of the $TMP environment variable
+  #   4) On Windows:  C:\TEMP, C:\TMP, \TEMP and TMP (in that order)
+  #      On all other platforms: /tmp, /var/tmp and /usr./tmp, in that order.
+  #   5) The current working directory
```
tldahlgren, line 52: I question whether we want to be elaborating on the ordering defined by the python implementation in our documentation.  

Perhaps a link to the relevant python docs (https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir) would be more suitable or a more general summary as before but with a little more elaboration?

etc/spack/defaults/config.yaml
tldahlgren: And if that variable isn't set, then Spack will use `"~%s.spack" % os.sep`.

tldahlgren (changes requested):

rountree:
@tldahlgren Good points.

I'm now wondering if a better solution would be to point out early on that `$lowercase` comes from the python execution environment and `$UPPERCASE` comes from the shell environment.  I'll poke around a bit to see if this is mentioned somewhere else.  Having a list of both would be useful.

PR #26782, opened 2021-10-17
add py-pytest to master version

joequant:
the master has pytest which needs to be added to dependents

PR #32134, opened 2022-08-14
gaudi: consistent test dependencies when +examples

wdconinc:
Gaudi requires testing to be enabled for examples to be built, so all
test dependencies are also there for `+examples`. This PR fixes a
missing pytest dependency when `+examples` is used but no testing is
enabled. The construct with the loop is to ensure the identical
dependencies are always used, even as version ranges may start to differ.

Testing with gaudi_add_tests was added in v35r0. Some nosetests and
QMtests were in the tree before, but not accessible it seems. The
effective dependency since 35.0 is not applied for pytest. This should now be identical to what was there before disentangling `optional`, at
https://github.com/spack/spack/blob/9d67d1e0346203a20b7ba44940061da855d405ca/var/spack/repos/builtin/packages/gaudi/package.py

### DIFF
--- a/var/spack/repos/builtin/packages/gaudi/package.py
+++ b/var/spack/repos/builtin/packages/gaudi/package.py
@@ -17,6 +17,7 @@ class Gaudi(CMakePackage):
     tags = ["hep"]
 
     version("master", branch="master")
+    version("36.7", sha256="8dca43185ba11e1b33f5535d2e384542d84500407b0d1f8cb920be00f05c9716")
     version("36.6", sha256="8fc7be0ce32f99cc6b0be4ebbb246f4bb5008ffbf0c012cb39c0aff813dce6af")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/gaudi/package.py
```diff
@@ -17,6 +17,7 @@ class Gaudi(CMakePackage):
     tags = ["hep"]
 
     version("master", branch="master")
+    version("36.7", sha256="8dca43185ba11e1b33f5535d2e384542d84500407b0d1f8cb920be00f05c9716")
```
wdconinc: Ah shoot, accidental include.
```suggestion
```

wdconinc:
This should obviate the need for https://github.com/spack/spack/pull/26782.

tldahlgren (approved):
LGTM.  Will defer merge decision until at least one maintainer has a chance to comment.

PR #32136, opened 2022-08-14
Local cache for buildcache files

blue42u:
There are cases (e.g. containers in CI) where the Spack install tree is frequently wiped. Buildcaches provide a convenient way to restore a Spack install from zero, but they require significant network I/O to redownload the buildcache files. Every. Single. Time.

While this can be reduced somewhat by caching the entire install prefix, this (a) affects concretization in non-trivial ways and (b) does not mesh well with [GitLab CI's `cache:` functionality](https://docs.gitlab.com/ee/ci/yaml/index.html#cache) (which [stores the cache as a zip file](https://docs.gitlab.com/ee/ci/caching/index.html#where-the-caches-are-stored)). Dodging these issues is possible but very awkward.

This patch adds a new local cache for some of these files (`.spack`). This cache is disabled by default but can be enabled by specifying a path in the `config:local_binary_cache` config entry (default `null`). Running `spack clean -B` (or `spack clean -a`) will wipe this cache.

**EDIT:** The `config:local_binary_cache` key can be a single directory where all cached files will be saved, or a list:
```yaml
local_binary_cache:
- prefixes: ['https://url/prefix', 's3://another/url/prefix']
  root: /path/to/cachedir
```
Any fetched binaries with the given prefix in their URL will be stored in the listed cache directory, first match wins. `prefixes` can be elided to match all URLs. This feature was added to better support mixed buildcaches, a temporary buildcache can be placed in a per-branch/PR GitLab cache (`key`) while a more permanent buildcache can use a shared GitLab cache (`key`).

**EDIT:** The key used in the cache is composed of a hash of the original URL plus the filename part of the path, e.g.:
```
cache-dir/817a9506ce_linux-almalinux8-x86_64_v3-gcc-8.5.0-pigz-2.7-dwvh5t2d3gbnjut6bkm74qzm6r6a6rg4.spack
```
The checksum from the `.json` or `.json.sig` file is used to verify the validity of the cached file and the file will be re-fetched if needed. The `.json` and `.json.sig` files are **not** cached, this was done to allow buildcaches to "refresh" their contents without requiring a manual cache refresh.

/cc @tgamblin @becker33

* * *

Performance reinstalling `llvm`, without this patch:
```console
$ time spack install --cache-only -f llvm.json
... preinstalled dependencies...
==> Installing llvm-12.0.1-odiouhgxol2wi55qkqtedbhcgqgoppwd
gpg: Signature made Fri 12 Aug 2022 07:33:15 AM UTC
gpg:                using RSA key D2C7EB3F2B05FA86590D293C04001B2E3DB0C723
gpg: Good signature from "Spack Project Official Binaries <maintainers@spack.io>" [ultimate]
==> Extracting llvm-12.0.1-odiouhgxol2wi55qkqtedbhcgqgoppwd from binary cache
... snip ...
real	5m43.580s
user	2m18.305s
sys	0m48.059s
```

With this patch:
```console
$ time spack install --cache-only -f llvm.json
... preinstalled dependencies...
==> Installing llvm-12.0.1-odiouhgxol2wi55qkqtedbhcgqgoppwd
==> Using cached archive: /tmp/binary_cache/s3___spack-binaries_develop-nmqjxwyuhckmvx47w7b7vvujnvwvsmox/linux-ubuntu18.04-x86_64-gcc-7.5.0-llvm-12.0.1-odiouhgxol2wi55qkqtedbhcgqgoppwd.spec.json.sig
gpg: Signature made Sun 14 Aug 2022 07:59:50 PM UTC
gpg:                using RSA key D2C7EB3F2B05FA86590D293C04001B2E3DB0C723
gpg: Good signature from "Spack Project Official Binaries <maintainers@spack.io>" [ultimate]
==> Using cached archive: /tmp/binary_cache/s3___spack-binaries_develop-nmqjxwyuhckmvx47w7b7vvujnvwvsmox/linux-ubuntu18.04-x86_64/gcc-7.5.0/llvm-12.0.1/linux-ubuntu18.04-x86_64-gcc-7.5.0-llvm-12.0.1-odiouhgxol2wi55qkqtedbhcgqgoppwd.spack
==> Extracting llvm-12.0.1-odiouhgxol2wi55qkqtedbhcgqgoppwd from binary cache
... snip ...
real	4m16.808s
user	2m36.772s
sys	0m43.384s
```

### DIFF
--- a/etc/spack/defaults/config.yaml
+++ b/etc/spack/defaults/config.yaml
@@ -216,6 +216,16 @@ config:
   # for updates, within a single Spack invocation. Defaults to 10 minutes.
   binary_index_ttl: 600
 
+  # Directory (or multiple directories) to cache files downloaded from buildcaches.
+  # Should be 'null' in most cases, unless you wipe your Spack install tree frequently (e.g. in CI).
+  local_binary_cache: null
+  # Examples:
+  # local_binary_cache: /path/to/cachedir
+  # local_binary_cache:
+  #   - prefixes: ['https://binaries.spack.io/']
+  #     root: /cachedir/for/public/buildcache
+  #   - root: /cachedir/for/everything/else
+
   flags:
     # Whether to keep -Werror flags active in package builds.
     keep_werror: 'none'
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -15,6 +15,7 @@ import tarfile
 import tempfile
 import time
 import traceback
+import urllib.parse
 import warnings
 from contextlib import closing
 from urllib.error import HTTPError, URLError
@@ -29,6 +30,7 @@ from llnl.util.filesystem import BaseDirectoryVisitor, mkdirp, visit_directory_t
 import spack.cmd
 import spack.config as config
 import spack.database as spack_db
+import spack.fetch_strategy
 import spack.hooks
 import spack.hooks.sbang
 import spack.mirror
@@ -550,6 +552,96 @@ def _binary_index():
 binary_index = llnl.util.lang.Singleton(_binary_index)
 
 
+class LocalBinaryCache(object):
+    """
+    The LocalBinaryCache caches files that would have been pulled from a (usually remote)
+    buildcache in a closer local directory. For cases where the Spack instance is being wiped
+    regularly (e.g. containers and/or CI) this can reduce the network I/O needed to
+    reinstall the packages.
+
+    By default this class does nothing and all requests go straight to the remote URL.
+    To change this set `config:binary_cache_root` to a value other than `null`.
+    """
+
+    def __init__(self):
+        self._file_caches = []
+        settings = config.get("config:local_binary_cache", None)
+        if settings is not None:
+            if isinstance(settings, str):
+                settings = [{"root": settings}]
+            for cache in settings:
+                prefixes = tuple(cache.get("prefixes", [""]))
+                self._file_caches.append((prefixes, file_cache.FileCache(cache["root"])))
+
+    def __bool__(self):
+        return len(self._file_caches) > 0
+
+    def destroy(self):
+        for _, c in self._file_caches:
+            c.destroy()
+
+    @staticmethod
+    def _url2key(url):
+        url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
+        parsed_url = urllib.parse.urlparse(url)
+        return "{0}_{1}".format(url_hash[:10], parsed_url.path.split("/")[-1])
+
+    def get(self, url, checksum, force=False):
+        """
+        Get the cached file for the given URL, fetching if needed.
+
+        Args:
+            url (str): Remote URL to fetch from.
+            checksum (str): Checksum for the target file.
+            force (bool): If True, cache even if ``url`` is a local path.
+
+        Returns tuple of:
+            cached_path (str): Path to the cached file
+            fetched (bool): ``True`` if the file had to be fetched, ``False`` otherwise
+
+            If the cache is disabled or if ``url`` is local (and ``force`` is not ``True``),
+            all fields above are None.
+        """
+        if len(self._file_caches) == 0 or (not force and url_util.local_file_path(url)):
+            return None, None
+
+        for prefixes, cache in self._file_caches:
+            if any(url[: len(prefix)] == prefix for prefix in prefixes):
+                break
+        else:
+            return None, None
+        checker = spack.util.crypto.Checker(checksum)
+        key = self.__class__._url2key(url)
+        cached_path = cache.cache_path(key)
+
+        # Check if the key is already in the cache with a valid checksum
+        if cache.init_entry(key):
+            with cache.read_transaction(key):
+                if checker.check(cached_path):
+                    tty.debug("Reusing cached result for url {0}".format(url))
+                    return cached_path, False
+
+        # Otherwise, attempt to fetch from the remote url and save to the cache
+        tty.debug("Saving url to local binary cache: {0}".format(url))
+        _, _, response = web_util.read_from_url(url)
+        with cache.write_transaction(key, binary=True) as (_, cache_f):
+            shutil.copyfileobj(response, cache_f)
+
+        # Validate the checksum. At this point failure is a hard error.
+        if not checker.check(cached_path):
+            raise NoChecksumException("File failed checksum verification.")
+
+        return cached_path, True
+
+
+def _local_binary_cache():
+    """Get the singleton local binary cache instance."""
+    return LocalBinaryCache()
+
+
+local_binary_cache = llnl.util.lang.Singleton(_local_binary_cache)
+
+
 class NoOverwriteException(spack.error.SpackError):
     """
     Raised when a file exists and must be overwritten.
@@ -845,18 +937,6 @@ def tarball_path_name(spec, ext):
     <tarball_directory_name>/<tarball_name>
     """
     return os.path.join(tarball_directory_name(spec), tarball_name(spec, ext))
-
-
-def checksum_tarball(file):
-    # calculate sha256 hash of tar file
-    block_size = 65536
-    hasher = hashlib.sha256()
-    with open(file, "rb") as tfile:
-        buf = tfile.read(block_size)
-        while len(buf) > 0:
-            hasher.update(buf)
-            buf = tfile.read(block_size)
-    return hasher.hexdigest()
 
 
 def select_signing_key(key=None):
@@ -1286,7 +1366,7 @@ def _build_tarball(
     shutil.rmtree(workdir)
 
     # get the sha256 checksum of the tarball
-    checksum = checksum_tarball(tarfile_path)
+    checksum = spack.util.crypto.checksum(hashlib.sha256, tarfile_path)
 
     # add sha256 checksum to spec.json
 
@@ -1423,21 +1503,34 @@ def try_verify(specfile_path):
     return True
 
 
-def try_fetch(url_to_fetch):
+def try_fetch(url_to_fetch, cache=True, checksum=None, **kwargs):
     """Utility function to try and fetch a file from a url, stage it
     locally, and return the path to the staged file.
 
     Args:
         url_to_fetch (str): Url pointing to remote resource to fetch
+        cache (bool): If False, don't cache via the LocalBinaryCache
+        checksum (str): Checksum for the fetched file, if available
 
     Returns:
         Path to locally staged resource or ``None`` if it could not be fetched.
     """
-    stage = Stage(url_to_fetch, keep=True)
+
+    fetcher = None
+    if cache and checksum is not None:
+        cached_path, fetched = local_binary_cache.get(url_to_fetch, checksum, **kwargs)
+        if cached_path is not None:
+            fetcher = spack.fetch_strategy.CacheURLFetchStrategy(cached_path, notify=not fetched)
+            checksum = None
+    fetcher = fetcher or spack.fetch_strategy.from_url_scheme(url_to_fetch, checksum=checksum)
+
+    stage = Stage(fetcher, keep=True)
     stage.create()
 
     try:
         stage.fetch()
+        if checksum is not None:
+            stage.check()
     except web_util.FetchError:
         stage.destroy()
         return None
@@ -1480,7 +1573,7 @@ def download_tarball(spec, unsigned=False, mirrors_for_spec=None):
        }
     """
     if not spack.mirror.MirrorCollection():
-        tty.die("Please add a spack mirror to allow " + "download of pre-compiled packages.")
+        tty.die("Please add a spack mirror to allow download of pre-compiled packages.")
 
     tarball = tarball_path_name(spec, ".spack")
     specfile_prefix = tarball_name(spec, ".spec")
@@ -1513,14 +1606,19 @@ def download_tarball(spec, unsigned=False, mirrors_for_spec=None):
 
     tried_to_verify_sigs = []
 
+    specfile_loaders = [
+        ("json.sig", Spec.extract_json_from_clearsig),
+        ("json", sjson.load),
+    ]
+
     # Assumes we care more about finding a spec file by preferred ext
     # than by mirrory priority.  This can be made less complicated as
     # we remove support for deprecated spec formats and buildcache layouts.
-    for ext in ["json.sig", "json"]:
+    for ext, loader in specfile_loaders:
         for mirror_to_try in mirrors_to_try:
             specfile_url = "{0}.{1}".format(mirror_to_try["specfile"], ext)
             spackfile_url = mirror_to_try["spackfile"]
-            local_specfile_stage = try_fetch(specfile_url)
+            local_specfile_stage = try_fetch(specfile_url, cache=False)
             if local_specfile_stage:
                 local_specfile_path = local_specfile_stage.save_filename
                 signature_verified = False
@@ -1533,6 +1631,18 @@ def download_tarball(spec, unsigned=False, mirrors_for_spec=None):
                     signature_verified = try_verify(local_specfile_path)
                     if not signature_verified:
                         tty.warn("Failed to verify: {0}".format(specfile_url))
+
+                # Extract the hash from the specfile to verify the spackfile
+                with open(local_specfile_path, "r") as specfile:
+                    specfile_data = loader(specfile.read())
+                if (
+                    "buildcache_layout_version" in specfile_data
+                    and int(specfile_data["buildcache_layout_version"]) >= 1
+                ):
+                    checksum = specfile_data["binary_cache_checksum"]["hash"]
+                else:
+                    # Old-form buildcache layout, don't verify
+                    checksum = None
 
                 if unsigned or signature_verified or not ext.endswith(".sig"):
                     # We will download the tarball in one of three cases:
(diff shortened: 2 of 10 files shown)


scheibelp:
(FYI: I have self-assigned this but won't have time to review it until next week (8/23) - if someone else is interested in the meantime, feel free to assign yourself)

becker33:
I think we need to ensure that signature files are fetched from cache *if and only if* the binary itself is fetched from cache. Otherwise we could "verify" a binary with thee wrong signature and run untrusted code. That will probably require more substantial changes.

blue42u:
@becker33 I've discovered since writing the OP that the `.json.sig` contains a hash for the `.spack` file, so even if the binary is not pulled from cache it is still verified with a cached signature file ([here](https://github.com/spack/spack/blob/develop/lib/spack/spack/binary_distribution.py#L1658-L1665)). I think that resolves the concern, but if not I can adjust this patch to only cache the binaries (`.spack`) and not the signatures.

What this patch doesn't do is re-fetch the binary if the cached version fails to verify. That should be possible without much extra work.

scheibelp:
Concerning the original PR description:

> While this can be reduced somewhat by caching the entire install prefix, this (a) affects concretization in non-trivial ways 

This part would be resolved by `spack concretize --fresh`, which ignores the current install state

> and (b) does not mesh well with [GitLab CI's cache: functionality](https://docs.gitlab.com/ee/ci/yaml/index.html#cache) (which [stores the cache as a zip file](https://docs.gitlab.com/ee/ci/caching/index.html#where-the-caches-are-stored)).

Apologies it's not immediately obvious to me why that's a problem: could you elaborate?

> This patch adds a new local cache for some of these files (.spack and .json/.sig). 

I think these files are only fetched if we anticipate having the associated package (i.e. we only download the .sig to verify a binary that we downloaded, and we only downloaded that because it matched the locally-concretized hash). If I got that right, I think the .sig/.json files are a small component of the overall bandwidth usage.

Caching the binaries themselves locally would save bandwidth but I think it would run into point (b) that I am also asking about.

I'm also curious: do you have performance numbers (or a notion of perceived time difference is also useful like "twice as fast")? That might neatly summarize something that I'm overlooking (but metrics aren't necessary if I get a better high-level understanding of the savings that we are trying to achieve here).

blue42u:
Thanks for the response @scheibelp. Sorry for the terse OP, I'll try to explain better.

> This part would be resolved by `spack concretize --fresh`, which ignores the current install state

Let me explain our use case a bit: we have a Spack environment we use in CI to supply build dependencies for HPCToolkit. Most but not all are in the Spack public buildcache or #32327 (our dependencies have shifted slightly due to ongoing refactors). We'd like to spend our CI-time on doing work for our research instead of compiling Spack packages, so we'd like to use the latest in the public buildcache as much as possible. Passing `--fresh` seems to go against that goal.

The issue with `--reuse` is, of course, that locally-installed packages often get preferred over updates from the buildcache. This patch provides a way to get the benefits of saving the install prefix (i.e. reduced network I/O from buildcache installs) without the change in behavior of `--reuse`.

> Apologies it's not immediately obvious to me why that's a problem: could you elaborate?

Sure: GitLab CI supports a [`cache:` job attribute](https://docs.gitlab.com/ee/ci/yaml/index.html#cache) that adds some code surrounding the job to load and/or save files to storage local to the CI runner (on the local filesystem by default). It's a convenient and easy way to cache files downloaded from the network across multiple CI jobs (reducing I/O time), and far more portable than exposing a custom directory inside the CI jobs (`cache:` will work with any runner's configuration).

So why not `cache: opt/spack`? The [GitLab CI Runner saves the cache as a zip file](https://docs.gitlab.com/ee/ci/caching/index.html#where-the-caches-are-stored), so doing this would erase all the file attributes and permissions packages use on Linux. In short, it's a recipe for disaster.

What we do now is `tar` the install prefix and then `cache:` that, but mixed with (a) this becomes a very awkward dance to get Spack to do what we want. This patch provides a far more convenient method with the same reduction in network I/O, just `spack config add config:build_cache_root:<path>` and then `cache: <path>`.

> I think these files are only fetched if we anticipate having the associated package (i.e. we only download the .sig to verify a binary that we downloaded, and we only downloaded that because it matched the locally-concretized hash). If I got that right, I think the .sig/.json files are a small component of the overall bandwidth usage.

I think you're right, there's not as much benefit to caching the `.sig`/`.json` files and as @becker33 mentioned it complicates the verification. I should have time to limit it to just `.spack` files over the weekend (and implement re-fetch on checksum failure).

> I'm also curious: do you have performance numbers (or a notion of perceived time difference is also useful like "twice as fast")? That might neatly summarize something that I'm overlooking (but metrics aren't necessary if I get a better high-level understanding of the savings that we are trying to achieve here).

I don't have any hard numbers, but I do have a back-of-the-envelope estimate for our use case. Assuming:
 - ~50 binary packages supplied by the buildcache (a CPU-only `hpctoolkit` spec is 48 packages),
 - ~50 MB (0.05 GB) per binary package (many are smaller but others are huge, e.g. `binutils` is ~260 MB),
 - ~20 CI jobs per pipeline that pull from the buildcache (we have 12 implemented so far and continue to expand),
 - ~1000 pipelines per month (overestimating ~25 commits per week times ~10 team members), and
 - $0.09 for 1 GB of download bandwidth on AWS (after first 100GB per month).

Then the total cost of the download bandwidth on AWS is \~$4500 a month. If we use the public buildcache for all but \~4 packages, that's still \~$360 a month on our bill.

In comparison, if the binaries are cached locally, we need \~10 CI jobs in a single pipeline to fill the caches and then no more downloads occur. If we refresh the cache weekly, the bandwidth is well within the first free 100GB per month, and our monthly AWS bill is a few cents.

So having a cache is *very* attractive to us. This patch makes it quite a bit easier to pull off.

blue42u:
Updated, now only `.spack` files will be cached and the checksum is verified before using a cached copy. To make it consistent, the checksum check is now in `download_tarball` instead of `extract_tarball` (except in the case of old-format buildcache files). I've also rearranged the code slightly to fit better with the current usage.

Demo:
```console
$ spack mirror add develop s3://spack-binaries/develop
$ spack config add config:binary_cache_root:/tmp/binary_cache
$ spack install --cache-only -f llvm.json
 # Measured 2030.2 MiB downloaded (via ip -s link)
$ spack uninstall -ay
$ spack install --cache-only -f llvm.json
 # Measured 0.781 MiB downloaded (via ip -s link)
```

scottwittenburg (commented):
@blue42u I'm guessing you host your own gitlab and runners on physical machines under your control?  In that case, I can see why a local cache would save you money (thanks for the "back-of-the-envelope", by the way).

While I don't see a practical way to use this in spack production pipelines, I have no objection to it in principle.  

However, I spot checked several of the gitlab pipelines that ran on this change and wasn't able to find any jobs that actually rebuilt anything.  While much of the code changes are theoretically checked with unit tests, I'd feel more confident it got exercised in a real pipeline. If we had "hey spackbot, rebuild everything" working already, I'd run it on this just to be sure, but we don't have that quite yet.

So I wonder if you're willing to make a WIP commit that just changes the mirror url in one of the smaller stacks to point to a non-existent location so we can see some job traces with this working?  For example, you could temporarily change the url on this [line](https://github.com/spack/spack/blob/5fcfce18fdb8c8574c049fbcb0792a513ec07fbb/share/spack/gitlab/cloud_pipelines/stacks/build_systems/spack.yaml#L32) to `s3://spack-binaries/develop/build_systems/does_not_exist`.

blue42u:
@scheibelp I followed your recommendation and got a successful rebuild of the `build_systems` environment, pipeline here: https://gitlab.spack.io/spack/spack/-/pipelines/163669. Does that ease your concern?

For your other questions:
> I'm guessing you host your own gitlab and runners on physical machines under your control? In that case, I can see why a local cache would save you money (thanks for the "back-of-the-envelope", by the way).

Indeed, we have very particular hardware requirements (mostly for GPUs) so our main CI workforce currently is all local machines. Having a local cache really does help.

> While I don't see a practical way to use this in spack production pipelines, I have no objection to it in principle.

Shameless self-promotion: https://github.com/spack/spack/pull/32300 should be able to help with this. It allows arbitrary job attributes to be configured, including the `cache:` needed to take advantage of this feature, like so:
```yaml
spack:
  specs: [...]

  config:
    binary_cache_root: $spack/opt/spack/binary_cache

  gitlab-ci:
    build-job-attributes:
      cache:
        key: binaries-$CI_JOB_IMAGE
        paths: [opt/spack/binary_cache/]
        when: always

    mappings:
    - ...
```

Without that PR, we've been using a separate Python script that adds the missing keys to the generated pipeline. Not exactly ideal and definitely not suitable for Spack production CI, but it works well enough for us.

haampie:
Maybe I'm a bit late. But why can't you just use a normal Spack install root / upstream? I don't see why you would even want to create tarballs if you're on a shared filesystem.

haampie:
Wouldn't another solution be to add a proxy / cache using [nginx](https://www.nginx.com/resources/wiki/start/topics/examples/reverseproxycachingexample/) or varnish for the remote mirror? That would save a lot of bandwidth. You can set a short ttl on the index.json, and all other `*.spack{,.sig}` files can be cached indefinitely. And then for writing to a local cache, use file:///my/local/cache?

I'm not a big fan of extending Spack with yet another type of buildcache.

blue42u:
Can you use nginx/varnish to proxy an S3 bucket? Does it pass through the authorization headers or do you need to configure those on the ngnix side? I'm not opposed to an alternative solution, but I didn't find any useful hits when I searched Google before so I figured there was some complicating factor preventing this from working.

> I'm not a big fan of extending Spack with yet another type of buildcache.

Unlike buildcaches this cache does not affect concretization or the install sequence, it's merely a network optimization. You can't "extend" a buildcache by adding files to this cache, and `*.json{,.sig}` **aren't** cached to ensure an up-to-date signature for the final binary (to allow the buildcache to update packages without a full wipe, the build-time configuration for a buildcache is non-trivial and wiping the buildcache multiple times when it's wrong isn't fun. It would be nice if Spack directly provided the "right" configs ([discussion](https://github.com/spack/spack/discussions/32749#discussioncomment-4111978)).)

In other words, it's the source cache, but for buildcaches. I personally don't consider that "another type of buildcache."

blue42u:
I've taken the opportunity to investigate @haampie's suggestion above. I was not able to find a viable solution that did not involve messing with the SSL Certificate Authorities certificates (*shivers*) or degrading the entire connection to HTTP. Neither of these are an acceptable compromise for our use case. Gory details in the spoiler:

<details>

### Prologue: Why we should avoid HTTP

Aside from the obvious concerns with sending an AWS key ID in cleartext across the internet, the more pressing issue is that I don't consider Spack's binary signing sufficiently secure for verifying binary code. Debian defines [a full time-limited chain of trust](https://wiki.debian.org/SecureApt) and they [*still* are vulnerable to certain attack vectors](https://wiki.debian.org/SecureApt/TufDerivedImprovements). Spack is a housecat in comparison, signing individual binary packages with no functionality to rotate/revoke keys. Even if it did, the amount of expertise needed to maintain such a buildcache is far past what an average user would be able to reliably accomplish.

The transport-level and server-host security is a lower bound on Spack's available security. Unencrypted HTTP provides pathetically low security, it should avoided at all costs.

### Reverse Proxy Approach
Reverse proxies like Nginx and Varnish stand in place of the proxied server (i.e. the client is mostly unaware that is it being proxied), so the client must be directed to use the proxy instead of the server (e.g. for AWS CLI used to test here, `--endpoint-url`). Unfortunately, the AWS `Authorization` slug used to secure a requester-pays or private bucket (like ours) [depends on the endpoint URL hostname (`Host` header)](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html). AWS will detect this case and redirect the client to the correct endpoint... which dodges the proxy:
```mermaid
sequenceDiagram
  participant C as Client
  participant P as rproxy.local
  participant S as AWS.com
  C->>+P: GET (rproxy.local)
  P->>S: GET (rproxy.local)
  S->>P: 301 Moved (to AWS.com)
  P->>-C: 301 Moved (to AWS.com)
  C->>S: GET (AWS.com)
  S->>C: 200 OK
```
Worse yet, this redirection preserves the scheme used in the first step. If `rproxy.local` is an HTTP server (far easier to set up), the later `GET` will cross the internet to AWS.com on HTTP, which we're trying to avoid.

The redirection could be prevented by making the proxy "AWS-aware" enough to rewrite the `Authorization` field, but this has the added downside storing strong AWS credentials in the proxy, not only in the client (Spack/Boto3)... as well as adding a lot of security-critical code already present in Boto3. Even if it's technically possible, it's unmaintainable for us.

### Forward Proxy Approach

I also looked into forward proxies like Squid.  These do not have the authorization problem since the client knows the real hostname of the server (i.e. the client is actively requesting to be proxied). The downside here is that HTTPS traffic gets shoveled through the proxy without decrypting it, meaning it can't be cached at all:
```mermaid
sequenceDiagram
  participant C as Client
  participant P as fproxy.local
  participant S as AWS.com
  C->>+P: CONNECT AWS.com
  C->>P: Encrypted (GET)
  P->>S: Encrypted (GET)
  S->>P: Encrypted (200 OK)
  P->>-C: Encrypted (200 OK)
```
This can be solved by configuring the proxy to ["bump" the SSL](https://wiki.squid-cache.org/Features/SslPeekAndSplice), but this requires fiddling with the CA certificates on the client side... which is a rabbit hole all of it's own. I did not have success implementing this approach.

The other option is to configure the client to use unencrypted HTTP, which the proxy can then rewrite as HTTPS before making the server connection. I got this to work, but if any bug in the client (Spack) accidentally ignores the `HTTP_PROXY`, that code will be transferring over insecure HTTP, which again, we're trying to avoid.

***

</details>

TL;DR: In our use case, caching in Spack is far easier and much more secure than any network proxy.

One concern @haampie alerted me to is that this PR adds a third block of caching / network code, with completely different implementation and caching behavior than the source cache or the BinaryIndexCache. Ideally these would instead all be more-or-less unified into a single code path with a single configurable "network cache" similar to what [pip](https://pip.pypa.io/en/stable/topics/caching/) and [npm](https://docs.npmjs.com/cli/v9/commands/npm-cache) do. Bonus points if this is [directly connected to the configured mirrors](https://github.com/spack/spack/discussions/32958).

I could work towards that in my spare time, but I'm apprehensive about spelunking in those dark corners as an external contributor. @tgamblin are there any plans for a refactor like this in the works? If not, is it possible to nudge this PR through as a stop-gap for our current problems?

haampie:
Source cache is content-addressable because we have shasums, similarly binary cache is content addressable because we have spec hashes. In principle we don't even have to care if we download spec.json.sig files from mirror 1 and tarballs from mirror 2, assuming hashes are universally unique -- it's called mirror for a reason. What would have to change is the (custom) handling of binary mirrors in binary cache stuff (it sort of reinvented it). Then if they follow the same logic, I think it wouldn't be so hard to use the same caching for sources and binaries.

What doesn't sit right with me in this PR is that it makes cache aware of mirror urls. Where you fetched something from shouldn't play any role in where to cache it. We just have "the" cache for all things that are content-addressable, and that keeps it very simple. Maybe we can distinguish between source and binary (cache) cache though.

blue42u:
> In principle we don't even have to care if we download spec.json.sig files from mirror 1 and tarballs from mirror 2, assuming hashes are universally unique -- it's called mirror for a reason.

I do not believe buildcaches satisfy this "universally unique" quality. There will only ever be one "correct" source tarball for `zlib@1.2.3`, but you can provide the same `zlib` spec in two different buildcaches and the `.spack` files will be different. The term "mirror" is a truly misnomer here.

More fundamentally, IMHO hashes should never be assumed to be universally unique, every cache entry needs to record the "request" that is being cached (e.g. package and version for sources and full spec for binaries). You wouldn't make a hashtable without a copy of the key after all. AFAICT the source cache is not currently implemented this way, nor does this PR help that situation (although URL keys are *slightly* better than nothing).

Unlikely as it is, the last thing I want is for `spack install zlib` to install LLVM. :face_with_spiral_eyes: 

> What would have to change is the (custom) handling of binary mirrors in binary cache stuff (it sort of reinvented it). Then if they follow the same logic, I think it wouldn't be so hard to use the same caching for sources and binaries.

I completely agree, since you made the point in our last discussion I am also unsatisfied with the separate cache re-implementation in this PR. I have a half-baked idea for what *could* work much better (expanding Mirrors to abstract network I/O w/ caching), but not the time to actually spelunk and implement such a refactor.

This PR at least gives us the features we need 3 months ago to run our CI, and with it our AWS monthly bill in practice is about a cup of Starbucks. So I'm keeping this barely alive until something better comes along.

> Where you fetched something from shouldn't play any role in where to cache it.

I do have one disagreement here, in CI it can make sense to distinguish between multiple mirrors. Like Spack's CI our CI also uses temporary per-PR buildcaches layered on top of a more permanent "`develop`" buildcache. In this arrangement, it makes a lot of sense (and is much more efficient!) to cache the temporary per-PR binaries separate from the shared permanent binaries, since a PR can't access another PR's buildcache they don't need the binaries cached from it.

It would be ideal if this cache was "mirror-centric." Without https://github.com/spack/spack/discussions/32958 this PR uses URL prefixes instead. Unsatisfying, but it works.

haampie:
> but you can provide the same zlib spec in two different buildcaches and the .spack files will be different.

True, forgot that the content hash of the tarball is in the (signed) spec.json file :| maybe we should make the tarballs content addressable and have mirrors provide the mapping of spec hash -> content hash.

Let me rephrase then: a pair (spec.json, tarball) can be fetched from any mirror, the mirror is irrelevant. We should make tarballs content-addressable, and then fetching would be trivially cache-able. Right now we always do a sequential fetch of spec.json followed by tarball, so instead we could get the content hash from the spec.json file and then fetch the tarball with cache.

PR #32141, opened 2022-08-15
Adding 0.11.9x versions of rose

mikerenfro:
Not sure if this was being maintained or not, as the last numbered Rose version in Spack (0.9.13.0) was released in December 2019.

Other notes or requests for feedback:

1. There's a Ghostscript dependency in current versions.
2. The 0.9.13.0 release doesn't build on a stock gcc-8.5.0 on Rocky. Not sure if there's a reason to keep it for other OSes, or how to flag it as incompatible with Rocky and similar OSes.
3. Current versions require C++14 support instead of C++11. Not sure how to make that conditional on Rose versions.


var/spack/repos/builtin/packages/rose/package.py
```diff
@@ -209,12 +215,16 @@ def configure_args(self):
 
         if "+cxx11" in spec:
             args.append("CXXFLAGS=-std=c++11")
+        if "+cxx14" in spec:
+            args.append("CXXFLAGS=-std=c++14")
```
alalazo, line 219: We need to use compiler specific flags for the recipe to be general, for instance `self.compiler.cxx14_flag`

alalazo (changes requested):

pinnown:
The third number in the version sequence indicates the weekly test so the 4 version are are not that different.

1. The Ghoscript requirement is an old one it was just not included in the list of dependencies. Older versions such as 9.25 should work as well.
2. Version 0.9.13.0 did not support gcc 8 on any system. 
3. A C++14 flag is unnecessary because all currently supported compilers default to C++14 or later.

alalazo:
@mikerenfro Are you planning on updating the PR according to comments?

alecbcs:
Closing due to a lack of response. @mikerenfro happy to reopen if you're still working on this.

PR #32142, opened 2022-08-15
py-scipy: Fortran compiler specify code is change to setup.cfg for Fujitsu compiler

tkameyama:
python package is build with pip In #27798.
And py-scipy%fj is used --install-option flag.
But new pip is ignored --install-option with message:

> Disabling all use of wheels due to the use of --build-option / --global-option / --install-option.

This pull request change to py-numpy's Fortran compiler detection order when fujitsu compiler, and successfull to build py-scipy%fj.

### DIFF
--- a/var/spack/repos/builtin/packages/py-numpy/package.py
+++ b/var/spack/repos/builtin/packages/py-numpy/package.py
@@ -150,14 +150,6 @@ class PyNumpy(PythonPackage):
         when="@1.22.0:1.22.3",
     )
 
-    # Change compiler search order for Fujitsu compiler
-    def patch(self):
-        f = join_path("numpy", "distutils", "fcompiler", "__init__.py")
-        if self.spec.satisfies("%fj"):
-            r = FileFilter(f)
-            r.filter(r", 'fujitsu'", "")
-            r.filter(r"'linux\.\*', \('", "'linux.*', ('fujitsu', '")
-
     # version 1.21.0 runs into an infinit loop during printing
     # (e.g. print(numpy.ones(1000)) when compiled with gcc 11
     conflicts("%gcc@11:", when="@1.21.0")
--- a/var/spack/repos/builtin/packages/py-numpy/package.py
+++ b/var/spack/repos/builtin/packages/py-numpy/package.py
@@ -150,6 +150,14 @@ class PyNumpy(PythonPackage):
         when="@1.22.0:1.22.3",
     )
 
+    # Change compiler search order for Fujitsu compiler
+    def patch(self):
+        f = join_path("numpy", "distutils", "fcompiler", "__init__.py")
+        if self.spec.satisfies("%fj"):
+            r = FileFilter(f)
+            r.filter(r", 'fujitsu'", "")
+            r.filter(r"'linux\.\*', \('", "'linux.*', ('fujitsu', '")
+
     # version 1.21.0 runs into an infinit loop during printing
     # (e.g. print(numpy.ones(1000)) when compiled with gcc 11
     conflicts("%gcc@11:", when="@1.21.0")
--- a/var/spack/repos/builtin/packages/py-scipy/package.py
+++ b/var/spack/repos/builtin/packages/py-scipy/package.py
@@ -77,6 +77,7 @@ class PyScipy(PythonPackage):
         "py-numpy@1.16.5:1.22+blas+lapack", when="@1.6.2:1.7", type=("build", "link", "run")
     )
     depends_on("py-numpy@1.17.3:1.24+blas+lapack", when="@1.8:", type=("build", "link", "run"))
+    depends_on("py-numpy %fj", when="%fj", type=("build", "link", "run"))
     depends_on("python@2.6:2.8,3.2:", when="@:0.17", type=("build", "link", "run"))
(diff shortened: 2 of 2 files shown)


adamjstewart:
At first glance, this seems fragile. Curious to see what @rgommers thinks and if there is a different way of passing in this flag that would work better, maybe `--config-settings`.

rgommers:
> But new pip is ignored --install-option with message:

Is it a new `pip` version or a new `setuptools` version? My first guess would be the latter, because of the major changes in setuptools 64. The release notes give an alternative: https://setuptools.pypa.io/en/latest/history.html#v64-0-0

adamjstewart:
I don't think it's the latter because my setuptools 64 PR hasn't been merged yet.

rgommers:
Ah okay, now I understand the problem after actually reading the linked PR. Spack changed from using `python setup.py xxx` to `pip install`. And using `def install_options` in a recipe with that is problematic - that will be the case for any recipe using it (it may or may not work, but it's the wrong thing to use). So yes, `--config-settings` should be the correct thing to use here.

adamjstewart:
@rgommers do you know off the top of your head approximately when numpy switched from a raw `setup.py` build to a PEP 517 `pyproject.toml` build? I think we want to use `--config-settings` for all versions after than and `--install-option` before that.

EDIT: scipy, not numpy

rgommers:
SciPy has a `pyproject.toml` since version 1.3.0 (https://github.com/scipy/scipy/pull/9924) and also before 1.1.0 but not in 1.1.x-1.2.x (see https://github.com/scipy/scipy/pull/8735). There is an interaction there with `pip` version though, so I'm not sure if it makes sense to worry about that <1.3.0 thing.

adamjstewart:
What's the correct flag to pass? `--config-settings=fcompiler=fujitsu`?

rgommers:
Yes that looks correct to me (didn't test though).

adamjstewart:
Hmm, does it matter if that `pyproject.toml` didn't have a `build-backend` key in the `[build-system]` section? What dictates whether or not PEP 517 is supported, and therefore `--config-settings` should be used?

rgommers:
I'm not sure without testing - I think it does, because even without specifying a build backend, the presence of `pyproject.toml` triggers build isolation. We only added a build backend key for scipy `1.9.0` and the PR for that isn't merged into Spack yet.

tkameyama:
Create setup.cfg instead of modifying numpy.
With this modification, we have confirmed that version 1.8.1 and version 1.2.3 can be built with the Fujitsu compiler.

adamjstewart:
@tkameyama what version of Spack are you using? We already create a `setup.cfg` for scipy, we call `self.spec['py-numpy'].package.setup_build_environment(env)` in `setup_build_environment`.

tkameyama:
> We already create a `setup.cfg` for scipy, we call `self.spec['py-numpy'].package.setup_build_environment(env)` in `setup_build_environment`.

We don't create setup.cfg.
We create site.cfg for scipy in  self.spec["py-numpy"].package.set_blas_lapack.

adamjstewart:
@rgommers what's the difference between the way that numpy and scipy use `site.cfg` vs. `setup.cfg`?

rgommers:
> @rgommers what's the difference between the way that numpy and scipy use `site.cfg` vs. `setup.cfg`?

`site.cfg` is a `numpy.distutils`-specific file that can be used to specify library search directories, which BLAS/LAPACK to use, and so on. It's defined in `numpy/distutils/system_info.py`. 

`setup.cfg` is a generic `distutils`/`setuptools` metadata file, which (I believe) can be used for things that normally come from `setup.py` but are known statically. It can also contain some dev tool config; the `numpy` one has `codespell` and `versioneer` sections. I personally never had a use for it.

adamjstewart:
Okay, so `site.cfg` to define dependency information and `setup.cfg` to define setuptools metadata. I wonder why the CLI flags from #32147 don't work but this PR does. I'm definitely not opposed to this solution, just want to understand it better. Maybe the `--config-settings` flag is just wrong. Actually, why do we need this at all? Numpy/scipy auto-detect the correct compiler for all other compilers, why does fujitsu need to be explicitly specified?

rgommers:
> Actually, why do we need this at all? Numpy/scipy auto-detect the correct compiler for all other compilers, why does fujitsu need to be explicitly specified?

Perhaps it's needed because users want a stack with two Fortran compilers, because not every program supports Fujitsu?

tkameyama:
> > Actually, why do we need this at all? Numpy/scipy auto-detect the correct compiler for all other compilers, why does fujitsu need to be explicitly specified?
> 
> Perhaps it's needed because users want a stack with two Fortran compilers, because not every program supports Fujitsu?

In My environment, gfortran and Fujitsu compiler is installed.
When config-fc option is missing, py-scipy is selected Gnu95FCompiler type.
Then Fortran programs is compiled by Fujitsu Compiler with Gnu95 options, and linked by gfortarn.
The py-scipy's .so files is linked without Fujitsu Fortran runtime libraries.

rgommers (approved):
Okay, this should be robust and not have effects outside of the scipy package being built. It's clean enough, I'm +1 for merging it as is.

adamjstewart (approved):

PR #32144, opened 2022-08-15
[WIP] [ycsb] added variants to enable/disable each backend individually

mdorier:
This PR modifies the YCSB package so that each backend can be enabled or disabled via a variant.

This prevents spack from building YCSB with all its backends by default (which currently doesn't work because some backends or backend dependencies fail are not up-to-date and fail to build with maven).

This PR also adds a master version.

### DIFF
--- a/var/spack/repos/builtin/packages/ycsb/package.py
+++ b/var/spack/repos/builtin/packages/ycsb/package.py
@@ -14,27 +14,105 @@ class Ycsb(MavenPackage):
     url = "https://github.com/brianfrankcooper/YCSB/archive/0.17.0.tar.gz"
     git = "https://github.com/brianfrankcooper/YCSB.git"
 
+    version('master', branch='master')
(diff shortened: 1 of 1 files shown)


mdorier:
@spackbot fix style

mdorier:
@spackbot fix style

mdorier:
@spackbot fix style

mdorier:
Note: I'm finding some problems with this PR, please don't merge yet.

mdorier:
I'm closing this PR:

The original motivation for this PR was to be able to select which backend to compile, because some of the backends seemed to not be maintained and to fail to build. This assumption was in fact not correct: some backends fail to compile because they require dependencies that aren't compatible with the latest maven. Downgrading to maven 3.6.3 makes the full build work correctly. So the right fix for this package is to force a maven version up to 3.6.3 but not greater. I will follow up with a new PR.

PR #32171, opened 2022-08-16
Add cxxstd flag to googletest and default mock to true

sethrj:


alalazo (approved):
Ahah, never refuse when somebody offers to be a maintainer! Thanks @sethrj

PR #32173, opened 2022-08-16
Require recent patchelf

haampie:
System `patchelf` is almost always full of bugs, so require a more recent version

Just leaving this here as a comment: why don't we use `spack external find` to detect system executables, install those so they're registered as external specs in the bootstrap store, and then we can simply query the bootstrap db instead of reinventing the wheel for externals.

### DIFF
--- a/lib/spack/spack/relocate.py
+++ b/lib/spack/spack/relocate.py
@@ -81,10 +81,8 @@ def _patchelf():
     if is_macos:
         return None
 
-    patchelf = executable.which("patchelf")
-    if patchelf is None:
-        with spack.bootstrap.ensure_bootstrap_configuration():
-            patchelf = spack.bootstrap.ensure_patchelf_in_path_or_raise()
+    with spack.bootstrap.ensure_bootstrap_configuration():
+        patchelf = spack.bootstrap.ensure_patchelf_in_path_or_raise()
 
     return patchelf.path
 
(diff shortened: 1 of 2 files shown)


lib/spack/spack/relocate.py
```diff
@@ -81,10 +81,8 @@ def _patchelf():
     if is_macos:
         return None
 
-    patchelf = executable.which("patchelf")
-    if patchelf is None:
-        with spack.bootstrap.ensure_bootstrap_configuration():
-            patchelf = spack.bootstrap.ensure_patchelf_in_path_or_raise()
+    with spack.bootstrap.ensure_bootstrap_configuration():
```
haampie, line 84: maybe this is expensive?

haampie:
@alalazo can we make patchelf download from the binary cache (with some logic avoiding recursion), patchelf doesn't need to be patchelf'ed itself. It seems that "no build from source" is too strict.

alalazo:
> Can we make patchelf download from the binary cache (with some logic avoiding recursion), patchelf doesn't need to be patchelf'ed itself. It seems that "no build from source" is too strict.

Yes, we can. On the Spack side we need either:
1. Logic avoiding recursion specific for `patchelf`, or
2. Logic to mark some buildcache as not needing RPATH relocation

Building `patchelf` always from sources is the "legacy" behavior we maintained so far.

haampie:
Closing in favor of #32262

PR #32179, opened 2022-08-16
CI: reduce the amount of tests run in the original concretizer

alalazo:
Modifications:
- [x] Skip all the tests marked as `maybeslow` when testing the original concretizer
- [x] Don't test Python 3.6 on the original concretizer


tgamblin (approved):

alalazo:
Removed Python 3.6, original concretizer from the required status check.

PR #32181, opened 2022-08-16
Kokkos: ROCm and CUDA are not compatible in Kokkos

kwryankrattiger:
This may be fixable in Kokkos, but for now it seems that there is not enough logic in the Kokkos build to support mixing HIP and CUDA devices in the same build of kokkos.

<details><summary>Error Message</summary><pre>
  >> 14    CMake Error at cmake/kokkos_test_cxx_std.cmake:132 (MESSAGE):
     15      Invalid compiler for CUDA.  The compiler must be nvcc_wrapper or Clang or
     16      use kokkos_launch_compiler, but compiler ID was HIPCC
</pre></details>

FYI: @eugeneswalker @wspear


kwryankrattiger:
rebase

alalazo:
@spackbot fix style

alalazo (approved):

keryell:
Perhaps you could suggest to use the SYCL back-end for this use-case?

PR #32183, opened 2022-08-16
Reorder workflow execution in GHA

trws:
This patchset refactors our GitHub actions into a single top-level ci workflow that invokes a series of reusable actions.  The main goal of this is to be able to easily control which tests run and in what order based on the success or failure of top-level prechecks.  Our previous workflows ran in three sets:

* nix tests: style and verification first, then linux and macos tests if successful
* windows tests: style and verification first, then linux and macos tests if successful
* bootstrap tests

As a result, the bootstrap tests ran even if the style failed, and style and verification had to run on two different platforms despite running identical checks.  I'm relatively sure that's because of the limitation on dependencies between steps in the jobs. Reusable workflows allow us to run the style, verification and now audit checks once, then depending on the results, and the files changed, run the appropriate nix, windows and bootstrap tests.  While it saves only a few minutes by itself, this makes it easier to refactor checks to subset tests without having to replicate tests or other workflow components in the future.

closes #25150

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -189,7 +189,7 @@ jobs:
                 not_found=0
                 old_path="$PATH"
                 export PATH="$ver_dir:$PATH"
-                ./bin/spack-tmpconfig ./.github/workflows/bootstrap-test.sh
+                ./bin/spack-tmpconfig -b ./.github/workflows/bootstrap-test.sh
                 export PATH="$old_path"
               fi
             fi
@@ -220,7 +220,7 @@ jobs:
                 not_found=0
                 old_path="$PATH"
                 export PATH="$ver_dir:$PATH"
-                ./bin/spack-tmpconfig ./.github/workflows/bootstrap-test.sh
+                ./bin/spack-tmpconfig -b ./.github/workflows/bootstrap-test.sh
                 export PATH="$old_path"
               fi
             fi
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -3,10 +3,7 @@ name: Bootstrapping
 on:
   # This Workflow can be triggered manually
   workflow_dispatch:
-  workflow_run:
-    workflows: [ "style" ]
-    types:
-      - completed
+  workflow_call:
   schedule:
     # nightly at 2:16 AM
     - cron: '16 2 * * *'
@@ -16,40 +13,9 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # Check which files have been updated by the PR
-  changes:
-    runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
-    # Set job outputs to values from filter step
-    outputs:
-      bootstrap: ${{ steps.filter.outputs.bootstrap }}
-    steps:
-    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # @v2
-      if: ${{ github.event_name == 'push' }}
-      with:
-        fetch-depth: 0
-    # For pull requests it's not necessary to checkout the code
-    - uses: dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721
-      id: filter
-      with:
-        # See https://github.com/dorny/paths-filter/issues/56 for the syntax used below
-          # Don't run if we only modified packages in the
-          # built-in repository or documentation
-        filters: |
-          bootstrap:
-          - 'var/spack/repos/builtin/**'
-          - '!var/spack/repos/builtin/packages/clingo-bootstrap/**'
-          - '!var/spack/repos/builtin/packages/clingo/**'
-          - '!var/spack/repos/builtin/packages/python/**'
-          - '!var/spack/repos/builtin/packages/re2c/**'
-          - '!lib/spack/docs/**'
-          - 'lib/spack/**'
-          - './share/spack/**'
   fedora-clingo-sources:
-    needs: [ changes ]
     runs-on: ubuntu-latest
     container: "fedora:latest"
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Install dependencies
         run: |
@@ -83,7 +49,6 @@ jobs:
   ubuntu-clingo-sources:
     runs-on: ubuntu-latest
     container: "ubuntu:latest"
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Install dependencies
         env:
@@ -120,7 +85,6 @@ jobs:
   ubuntu-clingo-binaries-and-patchelf:
     runs-on: ubuntu-latest
     container: "ubuntu:latest"
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Install dependencies
         env:
@@ -154,7 +118,6 @@ jobs:
   opensuse-clingo-sources:
     runs-on: ubuntu-latest
     container: "opensuse/leap:latest"
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Install dependencies
         run: |
@@ -183,7 +146,6 @@ jobs:
 
   macos-clingo-sources:
     runs-on: macos-latest
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Install dependencies
         run: |
@@ -205,7 +167,6 @@ jobs:
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
         macos-version: ['macos-11', 'macos-12']
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Install dependencies
         run: |
@@ -227,7 +188,6 @@ jobs:
     strategy:
       matrix:
         python-version: ['2.7', '3.6', '3.7', '3.8', '3.9', '3.10']
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Checkout
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
@@ -249,7 +209,6 @@ jobs:
   ubuntu-gnupg-binaries:
     runs-on: ubuntu-latest
     container: "ubuntu:latest"
-    if: ${{ needs.changes.outputs.bootstrap == 'true' &&  github.event.workflow_run.conclusion == 'success' && github.repository == 'spack/spack' }}
     steps:
       - name: Install dependencies
         env:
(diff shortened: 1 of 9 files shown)


.github/workflows/bootstrap.yml
```diff
@@ -3,19 +3,10 @@ name: Bootstrapping
...
-      # Don't run if we only modified packages in the
-      # built-in repository or documentation
-      - 'var/spack/repos/builtin/**'
-      - '!var/spack/repos/builtin/packages/clingo-bootstrap/**'
-      - '!var/spack/repos/builtin/packages/clingo/**'
-      - '!var/spack/repos/builtin/packages/python/**'
-      - '!var/spack/repos/builtin/packages/re2c/**'
-      - 'lib/spack/docs/**'
+  workflow_run:
+    workflows: [ "style" ]
```
vsoch: Should there be branches-ignore here too, or do we want to re-run on all merges?

vsoch (commented):
This is some seriously cool workflow-fu! (I've actually never tried the workflow run event and am curious to see how it works!) I think for workflows, the best review is seeing it work (or not) so I'm not sure I can provide much detailed review - the yaml looks good but I'm sure we might see a surprise or two when we run it!

.github/workflows/bootstrap.yml
trws: They don't compose, but because the style one only runs on the normal branch checks, and this only gets triggered when style finishes and succeeds, it should only trigger on those branches.  Worth testing though for sure.

alalazo: For some reason the workflows gated after style didn't run.

trws: I’ve been looking at that, apparently this kind of trigger only works once it exists on the develop or main branch. Will need to make up a test repo to the side to debug this. May be better to rework it as a reusable action and invoke from style instead.

vsoch:
I already gave some feedback, thanks!

alalazo:
Before merging this PR, since it will require to re-trigger unit tests on all the PRs that come after it[^1], can we also incorporate changes from https://github.com/spack/spack/pull/25150 and give more sensible names to some jobs?

[^1]: The names of "required checks" will change

trws:
That sounds excellent to me.  Would you like to target that over here, have me cherry pick it, or just manually propagate the changes? Last might be easiest as I'm considering decomposing this a bit more to fix the last of the reusable workflow problems.

alalazo:
@trws Whatever seems easier for you is fine with me. I'm looking forward to have `macos-unit-tests` instead of `build` :laughing:

.github/workflows/controller.yaml
```diff
@@ -0,0 +1,89 @@
...
+          filters: |
+            bootstrap:
+            - 'var/spack/repos/builtin/**'
+            - '!var/spack/repos/builtin/packages/clingo-bootstrap/**'
+            - '!var/spack/repos/builtin/packages/clingo/**'
+            - '!var/spack/repos/builtin/packages/python/**'
+            - '!var/spack/repos/builtin/packages/re2c/**'
+            - '!lib/spack/docs/**'
+            - 'lib/spack/**'
+            - './share/spack/**'
```
alalazo: Do I read correctly that `bootstrap` is set if we changed:
1. Anything under `share/spack`, or
2. Anything under `lib/spack` (but not docs), or
3. Any package (but not the packages involved in bootstrapping)

Shouldn't 3 be the opposite?

alalazo: I think we also need activate bootstrap on changes to the `bootstrap.yaml`.

.github/workflows/controller.yaml
```diff
@@ -0,0 +1,89 @@
...
+          echo Core changes: ${{ steps.filter.outputs.core }}
+          echo Event name: ${{ github.event_name }}
+          if [ "${{ steps.filter.outputs.core }}" == "true" ]
+          then
+            echo "::set-output name=with_coverage::true"
+          else
+            echo "::set-output name=with_coverage::false"
+          fi
+  bootstrap:
+    if: ${{ github.repository == 'spack/spack' && needs.changes.outputs.bootstrap == 'true' }}
```
alalazo: This workflow didn't run, I think because of the reasons above.

.github/workflows/controller.yaml
```diff
@@ -0,0 +1,89 @@
...
+    needs: [ style-and-validity, changes ]
+    uses: ./.github/workflows/bootstrap.yml
+  unit-tests:
+    if: ${{ github.repository == 'spack/spack' }}
+    needs: [ style-and-validity ]
+    uses: ./.github/workflows/unit_tests.yaml
+    with:
+      core: ${{ needs.changes.outputs.core }}
+      packages: ${{ needs.changes.outputs.packages }}
+      with_coverage: ${{ needs.changes.outputs.with_coverage }}
```
alalazo: I don't understand why unit tests were skipped. Do you have an explanation on that? I tried to find typos, but didn't see anything - and I found no logs on the reason why this job was skipped.

.github/workflows/unit_tests.yaml
```diff
@@ -1,107 +1,29 @@
...
-
   # Run unit tests with different configurations on linux
-  unittests:
-    needs: [ validate, style, changes ]
+  unit-tests-ubuntu:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['2.7', '3.6', '3.7', '3.8', '3.9', '3.10']
+        python-version: ['2.7', '3.6', '3.10']
```
alalazo: This will cut ~1h30 cpu-time for each CI workflow run on a core PR and use three less runners per PR. 

I have to note though that there is at least a counterexample to this strategy being the same as testing on the whole range. Using Python versions `[3.4 - 3.6]` we would catch bugs using 3.5 that we wouldn't using only the other two (because of dicts). I think the decision here is ultimately on @tgamblin since it's a trade-off between the time spent and the tests we perform.

Maybe, would it be easy to test on these versions for the PRs and test on every version on push to `develop`? If that is possible we might reap most of the time savings _and_ be notified rapidly if something breaks the compatibility with a specific Python version.

alalazo (changes requested):
I generally agree with the idea of the PR. It's not clear to me why unit-tests didn't run though.

alalazo:
@scottwittenburg @zackgalbreath Am I correct that before we merge this, since we change the name of a few workflow jobs, we also need to change something on the Gitlab side to allow pipelines to start when the new style checks completed? 

@trws If we go ahead we probably need to support both the old and the new name because of `releases` branches where I think this change won't be ported. Would you agree with that?

trws:
> I generally agree with the idea of the PR. It's not clear to me why unit-tests didn't run though.

Me either, I'm re-running with debug so we can see the values used. Had a bit of trouble with this before after renaming, so it may be spurious but I'd rather be sure.

.github/workflows/controller.yaml
trws: I agree, but it's worth noting these are the exact conditions we had before, except that we now also test on changes to core or share. I'll take another pass on these.

.github/workflows/controller.yaml
trws: Yes, that's correct. I'll add bootstrap.yaml so we can see that part trigger.

.github/workflows/controller.yaml
trws: My best guess right now is that the first run after update behaves differently than subsequent runs. I don't have an explanation for *why* that is, but let's see what the debug logins shows. That includes all the input and output values which helps a lot.

.github/workflows/unit_tests.yaml
trws: I could make it so develop would test the full range while PRs test the smaller set, or even restore this. Even without it the response time on PR or push should be rather better since we don't slam CI with everything up front. Still looking at how to collapse down the bootstrap test, it's not too bad but needs a bit of scripting.

.github/workflows/controller.yaml
trws: I reversed them, it was the reverse condition before.  Also added bootstrap.yaml.

trws:
Ok, I think what's going on is somehow the reusable workflows were getting the same value for github.workflow in the concurrency group setting, so they were cancelling each other.  This push should fix that.

alalazo (commented):
LGTM, modulo the decision on what to do for the Python versions used for the unit tests (can also be deferred to a following PR) and pipelines not starting

scottwittenburg:
> Am I correct that before we merge this, since we change the name of a few workflow jobs, we also need to change something on the Gitlab side

Yes, I think if the name is no longer `style`, we will need to update [here](https://github.com/spack/spack-infrastructure/blob/f9e4343ef053a99b9f6dfb6d42f10c23c529f11f/k8s/custom/gh-gl-sync/cron-jobs.yaml#L46).

alalazo:
@scottwittenburg Would it be an issue if the name changes in `develop` and stays the same in the `releases` branch?

scottwittenburg:
> Would it be an issue if the name changes in `develop` and stays the same in the `releases` branch?

Is this just going to be the case until the next release catches up with develop?  Just curious.  

We could easily invoke the sync script to match the name used in `develop`, since PRs are mostly based on that rather than on a release branch, and it seems we only gate the gitlab check on the style check for PRs, not protected branches.  @zackgalbreath please correct me if I'm wrong.  If we need more expressiveness than just a list of check names, all of which must run and succeed, before we push any PR branch, I'm sure we can come with something.

trws:
Ok, bootstrap test rework now in, uses the hostedtoolcache with my tmpconfig wrapper to test bootstrapping on all pythons in the same image. Cuts bootstrap builds on macOS and Ubuntu from 5 each to 1 each, and ~2 minutes per os/py version combo to ~3 minutes per OS. Not quite as fast, assuming max concurrency, but uses 12 less runners to get the bootstrap done almost as fast. Main cost in those is in pulling all of spack, so pulling it once rather than 5 times helps a whole lot.

vsoch (dismissed):
If it works, lgtm!

alalazo (dismissed):
Assuming we don't need to change any workflow or job names, and that we hook-up Gitlab, this LGTM.

.github/workflows/unit_tests.yaml
tgamblin: We should still run the full range somewhere and make sure we fix issues that come up.  There are some tests, particularly the ones for `spack.util.unparse`  and `spack.util.package_hash`, that are AST-sensitive -- those can vary in tricky ways with some of the small changes between versions.   The ones for spec hashing are also sensitive to dict ordering.

If that only happens on `develop` I don't think it's so bad.  I think it's the kind of thing that should happen on enqueue once we have merge queue, but we can watch commits to `develop` and make sure to fix them in the meantime.  Thoughts?

.github/workflows/controller.yaml
```diff
@@ -0,0 +1,90 @@
+name: MC
```
tgamblin: Can we just call this `ci`?  It's small and doesn't take too much room in the actions UI, and it avoids `master`, which we got rid of throughout Spack a while back.

`ci` is also intuitive to someone who doesn't really know how Spack's CI is structured -- they just know it's the top-level CI thing.

tgamblin (changes requested):

.github/workflows/controller.yaml
trws: Done.

trws:
@tgamblin, I agree on running those on develop, but am not 100% sure what the best way to actually *do that* is.  There are a couple of options, but for the moment leaving the full list there and pulling down the time with the parallelism PR seems a reasonable start, then we can prune after if things are still getting gummed up too badly.  Most recent push reinstates all the original versions.

alalazo:
As we are at it, do we want to change names also for the Windows unit tests? Otherwise this LGTM as it is.

trws:
Actually, you have a point, we have a bunch of duplicative naming right now, I can take one more pass on this before we do the gitlab update.

trws:
The names of the windows tests mostly made sense, but I did rename one to match the command it's testing and changed `windows-unit-tests` as the top label to `windows` which is qualified by the inner ones.  Also the linux and mac tests no longer say `unit-tests` twice for no reason.  Work for you @alalazo?

trws:
Actually... I'm also going to make audit a precheck.  If packages can't load, why run everything?

alalazo (dismissed):

alalazo:
@trws I think we lost the renaming done yesterday with the latest force-push?

trws:
Oof, you're absolutely right, I was trying to pull in the test updates that were killing the parallel PR and somehow lost that commit.  Thanks for the heads-up @alalazo.

trws:
Ok, should all be back.  The parallel PR has a commit that's listed twice, but attempting to move either removes the change while leaving the commit in place, which is really, really strange and seems broken, but I'm leaving it either way.

tgamblin (approved):

tgamblin:
Looks great!

scottwittenburg:
So, time to get this [PR](https://github.com/spack/spack-infrastructure/pull/312) into production then?

PR #32185, opened 2022-08-16
Geant4: update cmake defines and add support for nvhpc

sethrj:
Geant4 fails to build using the PGI-derived NVC++ (nvhpc) toolchain without these patches. The main two sets of errors are:
```
error: excessive recursion at instantiation of class "G4Number<191>"
```
due to a suboptimal implementation of "extensible enums" which can be circumvented using a magical `-Wc,--pending_instantiations` flag, and a couple instances of:
```
error: thread-local declaration follows non-thread-local declaration
```
which can probably be cleaned up in the next geant release (so for now I'm disabling threads with nvhpc).

[spack-build-out.txt](https://github.com/spack/spack/files/9353169/spack-build-out.txt)

### DIFF
--- a/var/spack/repos/builtin/packages/geant4/package.py
+++ b/var/spack/repos/builtin/packages/geant4/package.py
@@ -58,7 +58,7 @@ class Geant4(CMakePackage):
     variant("x11", default=False, description="Optional X11 support")
     variant("motif", default=False, description="Optional motif support")
     variant("qt", default=False, description="Enable Qt support")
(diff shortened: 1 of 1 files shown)


alalazo (approved):
@drbenmorgan I'm going ahead merging, since the changes look good from a packaging perspective, and this has been sitting in the queue for a week. Feel free to leave a comment here if you want some changes, I'm pretty sure @sethrj will submit a follow up PR for those, in case.

PR #32186, opened 2022-08-16
kokkos: fix conflict for deprecated `sycl` namespace

tgamblin:
We're seeing errors pop up with older versions of Kokkos and newer versions of `oneapi`, specifically:

    error: no member named 'ONEAPI' in namespace 'sycl'

This hapens because `sycl::ONEAPI` is `sycl::ext::oneapi` since oneapi `2022.0.0`. `kokkos@3.6:` uses the new namespace. A conflict was present for this, but it was too specific -- both to `dpcpp` and to the `2022.0.0` version.

This should fix the `oneapi` build on `develop`, or if it doesn't, the version(s) of Kokkos there will need to be updated (@eugeneswalker FYI).

- [x] Expand version ranges in `kokkos` conflict
- [x] Add conflict for `oneapi` in addition to `dpcpp`

### DIFF
--- a/var/spack/repos/builtin/packages/kokkos/package.py
+++ b/var/spack/repos/builtin/packages/kokkos/package.py
@@ -54,7 +54,8 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+openmptarget", when="@:3.5")
 
     # https://github.com/spack/spack/issues/29052
-    conflicts("@:3.5.00 +sycl", when="%dpcpp@2022.0.0")
+    conflicts("@:3.5 +sycl", when="%dpcpp@2022:")
+    conflicts("@:3.5 +sycl", when="%oneapi@2022:")
 
     tpls_variants = {
         "hpx": [False, "Whether to enable the HPX library"],


trws (approved):

eugeneswalker:
> This should fix the `oneapi` build on `develop`, or if it doesn't, the version(s) of Kokkos there will need to be updated (@eugeneswalker FYI).

How is `kokkos@3.5.00` even getting into develop? Our `e4s-oneapi` environment doesn't have kokkos version pinned, and when I concretize it manually (just now, using spack@develop), `kokkos@3.5.00` doesn't show up anywhere. 

Ref: https://github.com/spack/spack/blob/develop/share/spack/gitlab/cloud_pipelines/stacks/e4s-oneapi/spack.yaml

Update: Oh, actually, it is getting an even older `kokkos@3.4.01` and it must be because `concretizer: unify: when_possible` ... What could we do about this? Maybe we need to make `kokkos@3.6.xx` an explicit root spec? Although if we did that, we would have to be sure to update the version pinning on the explicit spec each time a new version of kokkos is available. There may be a spec in our environment that explicitly depends on an older version of kokkos.

tgamblin:
I suspect it would choose a newer version if some package wasn't mandating an older kokkos.  I'd look at the kokkos requirements on the specs you've got in the stack.

I think if you wait to see how it concretizes with this conflict, you'll either see it fail because the new conflict rules out your compiler choice, or you'll see two kokkos's in the stack and maybe you can see which package depends on the old one.

eugeneswalker (approved):

PR #32187, opened 2022-08-16
ci: ensure publishing to root happens serially

scottwittenburg:
Previously each pipeline copied the target buildcache to the root
as a final job in the child pipeline.  When multiple stacks contain
the same spec hashes, the associated binaries files are indistinguishable
by name, and race conditions could occur where the spec.json.sig from
one stack wins the copy race, while the .spack file from another
stack wins.  This results in binaries where the checksum stored in
the spec.json.sig does not match that of the .spack file, causing
installation failures.

Until we refactor gitlab ci to make addition of stacks easier, we
do not have a good way to generate code that should operate on all
stacks.  Until we have that, this workaround fixes the race condition
described above.

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -720,8 +720,6 @@ def generate_gitlab_ci_yaml(
     # Values: "spack_pull_request", "spack_protected_branch", or not set
     spack_pipeline_type = os.environ.get("SPACK_PIPELINE_TYPE", None)
 
-    spack_buildcache_copy = os.environ.get("SPACK_COPY_BUILDCACHE", None)
-
     if "mirrors" not in yaml_root or len(yaml_root["mirrors"].values()) < 1:
         tty.die("spack ci generate requires an env containing a mirror")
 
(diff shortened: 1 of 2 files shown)


scottwittenburg:
@eugeneswalker This is a quick approach to address the problems you have been seeing with dependencies failing to install because of the checksum failure.

@tgamblin We talked about this awhile ago on slack, but I didn't have any idea for a good solution that could be implemented quickly.  This is quick at least.

eugeneswalker (dismissed):

eugeneswalker:
> @eugeneswalker This is a quick approach to address the problems you have been seeing with dependencies failing to install because of the checksum failure.

Looking at this PR here as a way to get this other one merged: https://github.com/spack/spack/pull/32037

Would I just need to rebase the other one on this?

scottwittenburg:
> Would I just need to rebase the other one on this?

Yes, but it would only work once this is merged to `develop`, since the fix is to a job that only runs on protected branches.  

And I meant to try and simulate a protected pipeline run with this in a safe environment (because this change can't even be tested in a PR).  But I haven't had a chance to do that yet.  If it passes the checks, we could cross our fingers, merge it, and plan to iterate with another PR if it doesn't work.

eugeneswalker:
Now that https://github.com/spack/spack/commit/2d6087995caa2dc03c2468992e0b8ccec29b0e85 has come through, I expect rebasing on develop will fix the pipeline

zackgalbreath:
Replaced by #32407?

scottwittenburg:
> Replaced by #32407?

Yes indeed, thanks.

PR #32189, opened 2022-08-16
rocprofiler-dev: update `PROF_API_HEADER_PATH` path

lukebroskop:
change the path set for `PROF_API_HEADER_PATH` to use `include` instead of `inc`


tldahlgren (approved):
LGTM.  Will defer merge decision until at least one maintainer has a chance to comment.

srekolam (commented):
I think you may be referring to /opt/rocm-5.2.0/roctracer/include/ext/prof_protocol.h using spack externals 
This is what i remember - When it came to rocprofiler and roctracer recipes, there were some interdependencies between each other. This is planned to get resolved in 5.3.0 release. Also, hip is also referring to PROF_API_HEADER_PATH. and roctracer was dependent on HIP to build. So to overcome this , we introduced a Header only recipe , roctracer-dev-api .  But i see we are doing a mkdirp(prefix.roctracer.inc) which needs to change as well  to mkdirp(prefix.roctracer.include) and HIP needs to refer to this path as well i think.   This change alone will break the spack install rocprofiler-dev ( if you dont have packages.yaml file )
I am currently making additional changes for rocm-5.2.1 and can include changes to correct this inside the roctracer-dev-api as well as HIP.

srekolam:
@lukebroskop  i just added the changes to hip, rocprofiler, roctracer-dev-api in my commit  in #32195 . This PR changes are also included in. Let me know if you have any comments . i have verified  hip@4.5.0, hip@5.1.0 and hip@5.2.1

lukebroskop:
Thanks @srekolam! Since this change is now reflected in #32195 I will close this PR.

PR #32086, opened 2022-08-12
[do not merge] absolute sonames ci test

haampie:
Build the universe for #31948.

- hook: absolutify_elf_sonames
- poke gitlab/aws to build the world

PR #32196, opened 2022-08-16
bump patchelf on amazon linux

haampie:
@bollig can you make this persistent in the image, amazon linux ships patchelf
0.9 which produces broken binaries more often than not, it's a bit of a miracle
it works so far.

FWIW, bugs are surfacing in #32086, libcublas.so breaks after patchelf, likely
due to one of the many bugs related to large filesize, or any other bug that was
fixed in the past 6 years after the 0.9 release...


haampie:
Might soon be superseded by #32262.

haampie:
Closing since we bootstrap patchelf now

PR #32201, opened 2022-08-17
scorep: modify configure flags for Cray platforms

lukebroskop:
Remove --with-nocross-compiler-suite=cce and --with-mpi at configure time when targeting cray platforms.

From the scorep team:

>On Cray systems there is no need to/you should not provide
--with-nocross-compiler-suite
--with-mpi
--with-mpi=mpich
CFLAGS
CXXFLAGS
MPICC
MPICXX
MPIF77
MPIFC
>
>Instead, you load your desired PrgEnv, and we take care of using cc, CC,
and ftn for the compute node libraries and /usr/bin/g++ for the
login-node tools.


alalazo:
Can you rebase this PR so to trigger pipelines?

alalazo (approved):

PR #32202, opened 2022-08-17
Add HDF5 version 1.13.2.

lrknox:
Remove HDF5 versions 1.13.0 and 1.13.1.

### DIFF
--- a/var/spack/repos/builtin/packages/hdf5/package.py
+++ b/var/spack/repos/builtin/packages/hdf5/package.py
@@ -46,10 +46,7 @@ class Hdf5(CMakePackage):
     version("develop-1.8", branch="hdf5_1_8")
 
     # Odd versions are considered experimental releases
-    version(
-        "1.13.2",
-        sha256="01643fa5b37dba7be7c4db6bbf3c5d07adf5c1fa17dbfaaa632a279b1b2f06da"
-    )
+    version("1.13.2", sha256="01643fa5b37dba7be7c4db6bbf3c5d07adf5c1fa17dbfaaa632a279b1b2f06da")
 
     # Even versions are maintenance versions
     version(
--- a/var/spack/repos/builtin/packages/hdf5/package.py
+++ b/var/spack/repos/builtin/packages/hdf5/package.py
@@ -46,8 +46,10 @@ class Hdf5(CMakePackage):
     version("develop-1.8", branch="hdf5_1_8")
 
     # Odd versions are considered experimental releases
-    version("1.13.1", sha256="051655873105112f7aeccd5f59ab21f35f7f4907f06921ae61aaf1ef1c71fd53")
-    version("1.13.0", sha256="3049faf900f0c52e09ea4cddfb83af057615f2fc1cc80eb5202dd57b09820115")
+    version(
+        "1.13.2",
+        sha256="01643fa5b37dba7be7c4db6bbf3c5d07adf5c1fa17dbfaaa632a279b1b2f06da"
+    )
 
     # Even versions are maintenance versions
     version(


lrknox:
The HDF5 1.13.x versions are "experimental" releases from the develop branch to provide stable code updates for VOL developers to use in advance of the 1.14.0 release.  The latest 1.13.x release should be used.  We are removing the earlier versions.

var/spack/repos/builtin/packages/hdf5/package.py
```diff
@@ -46,8 +46,7 @@ class Hdf5(CMakePackage):
     version("develop-1.8", branch="hdf5_1_8")
 
     # Odd versions are considered experimental releases
-    version("1.13.1", sha256="051655873105112f7aeccd5f59ab21f35f7f4907f06921ae61aaf1ef1c71fd53")
-    version("1.13.0", sha256="3049faf900f0c52e09ea4cddfb83af057615f2fc1cc80eb5202dd57b09820115")
```
tldahlgren, line 50: Is there a reason to remove versus deprecate this -- actually both of these -- versions?

var/spack/repos/builtin/packages/hdf5/package.py
```diff
@@ -46,8 +46,7 @@ class Hdf5(CMakePackage):
...
+    version("1.13.2", sha256="01643fa5b37dba7be7c4db6bbf3c5d07adf5c1fa17dbfaaa632a279b1b2f06da")
```
tldahlgren, line 49: Confirmed sha256.

tldahlgren (commented):
Quick question about removing versions.  Normally we deprecate them but there is guidance at https://spack.readthedocs.io/en/latest/packaging_guide.html?highlight=deprecate%20#deprecating-old-versions.

hyoklee (approved):

hyoklee (changes requested):
@lrknox , please deprecate old versions instead of removing them per @tldahlgren 's suggestions.
If you really want to remove the old versions, first remove soruce packages from https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.13/ permanently so the users won't be able to use them.

derobins:
The reason to remove old versions of experimental releases is that they are unsupported by The HDF Group. They are essentially release candidates that have gone through a more formal release process. I'm not very familiar with spack's ideas about deprecation, but keeping them around is probably a mistake.

Perhaps a compromise would be to deprecate them until 1.14.0 is released and then remove them.

var/spack/repos/builtin/packages/hdf5/package.py
derobins: Yes. See my comments below.

tldahlgren:
> The reason to remove old versions of experimental releases is that they are unsupported by The HDF Group. They are essentially release candidates that have gone through a more formal release process. I'm not very familiar with spack's ideas about deprecation, but keeping them around is probably a mistake.
> 
> Perhaps a compromise would be to deprecate them until 1.14.0 is released and then remove them.

Fair enough.  From the comment, I inferred that `1.13.0` was not an experimental release.  Is that not the case?  If it is not experimental, then could it be deprecated?

derobins:
> > The reason to remove old versions of experimental releases is that they are unsupported by The HDF Group. They are essentially release candidates that have gone through a more formal release process. I'm not very familiar with spack's ideas about deprecation, but keeping them around is probably a mistake.
> > Perhaps a compromise would be to deprecate them until 1.14.0 is released and then remove them.
> 
> Fair enough. From the comment, I inferred that `1.13.0` was not an experimental release. Is that not the case? If it is not experimental, then could it be deprecated?

All 1.13.x releases are experimental releases, including 1.13.0. 

1.13.0 and 1.13.1 should be treated identically now that 1.13.2 is out, whether that is deprecation or deletion.

tldahlgren (approved):

tldahlgren:
> The HDF5 1.13.x versions are "experimental" releases from the develop branch to provide stable code updates for VOL developers to use in advance of the 1.14.0 release. The latest 1.13.x release should be used. We are removing the earlier versions.

🤦 Missed this earlier.

PR #31664, opened 2022-07-20
Backports v0.18.2

haampie:
- [x] #31626
- [x] #31646
- [x] #31677 
- [x] #31665 
- [x] #31860 
- [x] #31864 
- [x] #32175
- [x] #32426
- [ ] #31914 
- [ ] #32040 
- [ ] #32210 
- [ ] #33079

PR #32210, opened 2022-08-17
spec: fix typo in Spec._finalize_concretization

alalazo:
It seems in a loop to assign DAG hash to nodes we always call `self` instead of `spec`


tgamblin (approved):

haampie:
backports?

alalazo:
@haampie Added to the list in #31664

PR #32211, opened 2022-08-17
package_base: remove rpath_args and "deprecate" rpath properties

skosukhin:
I dared to remove and "deprecate" two properties of the `PackageBase` class:
1. Property `rpath_args` currently returns flags with hardcoded flag prefixes `-Wl,-rpath,`. Instead, it should be the compiler-specific properties `cc_rpath_arg`, `cxx_rpath_arg`, etc. We could try to extend the interface instead of removing it but that does not make much sense if I manage to convince you all that the property `rpath`, which is used in `rpath_args`, should be removed as well.
2. There are several problems with the `rpath` property, which I tried to describe in the comments:
    ```python
    # TODO: this method is deprecated and will be removed in the future versions of Spack:
    #  1. The method duplicates the logic of build_environment.set_wrapper_variables(), which
    #     sets the actual rpath flags to be used when building the package. Therefore, the
    #     output of this method is prone to not delivering what it promises. For example, the
    #     current implementation does not return the contents of SPACK_COMPILER_IMPLICIT_RPATHS
    #     and SPACK_COMPILER_EXTRA_RPATHS, which would be fair to assume that it does.
    #  2. Most of the packages should not need this method: the compiler wrappers are supposed
    #     to do the injection of the rpaths automatically, and if they do not, it is a bug in
    #     the core functionality of Spack, which should be fixed.
    #  3. A few packages that can make reasonable use of this method (e.g. gcc, which needs
    #     the rpath flags at the bootstrapping stage) should take over the responsibility and
    #     implement their own versions of the method that deliver what they actually need (e.g.
    #     gcc does not need its installation prefix on the list).
    ```
    So, I think that `rpath` should be removed too. I have added `TODO` comments to the packages that use the method to notify their maintainers about this change. I would like to ask the maintainers to check whether they actually need the method. It might be the case that the issues they tried to workaround with it have already been resolved and the respective code that uses `self.rpath` can be simply removed.
3. Regardless of whether `rpath` stays, it should filter the system paths out (relevant for external packages installed to, for example, `/usr`).

I also have two questions about `build_environment.set_wrapper_variables()`. Maybe somebody can answer them:
1. Should we really prefer `prefix.lib` over `prefix.lib64` [here](https://github.com/spack/spack/blob/8d8aa5c6cf294e65d96edc8ee276cc0a315a4693/lib/spack/spack/build_environment.py#L446) and [here](https://github.com/spack/spack/blob/8d8aa5c6cf294e65d96edc8ee276cc0a315a4693/lib/spack/spack/build_environment.py#L474), and not the other way around?
2. Given that we skip the system prefixes when iterating over the link dependencies
    https://github.com/spack/spack/blob/8d8aa5c6cf294e65d96edc8ee276cc0a315a4693/lib/spack/spack/build_environment.py#L438
    does `filter_system_paths` actually do anything?
    https://github.com/spack/spack/blob/8d8aa5c6cf294e65d96edc8ee276cc0a315a4693/lib/spack/spack/build_environment.py#L480

### DIFF
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -2751,13 +2751,6 @@ class PackageBase(six.with_metaclass(PackageMeta, PackageViewMixin, object)):
         rpaths.extend(d.prefix.lib64 for d in deps if os.path.isdir(d.prefix.lib64))
         return rpaths
 
-    @property
-    def rpath_args(self):
-        """
-        Get the rpath args as a string, with -Wl,-rpath, for each element
-        """
-        return " ".join("-Wl,-rpath,%s" % p for p in self.rpath)
-
     def _run_test_callbacks(self, method_names, callback_type="install"):
         """Tries to call all of the listed methods, returning immediately
         if the list is None."""
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -33,7 +33,7 @@ import six
 
 import llnl.util.filesystem as fsys
 import llnl.util.tty as tty
-from llnl.util.lang import classproperty, match_predicate, memoized, nullcontext
+from llnl.util.lang import classproperty, dedupe, match_predicate, memoized, nullcontext
 from llnl.util.link_tree import LinkTree
 
 import spack.compilers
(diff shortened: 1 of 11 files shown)


skosukhin:
Hmm, why didn't the bot ping maintainers of the packages that I have changed?

skosukhin:
The discussion in #32212 seems to be relevant.

skosukhin:
@alalazo I don't know why exactly the bot didn't ping you (and the other package maintainers) but it would be nice to know your opinion on this.

PR #32213, opened 2022-08-17
[ycsb] fixes build process and installation of YCSB

mdorier:
The existing YCSB package was failing to build with the most recent version of maven because this version does not allow pulling dependencies from http URLs (only https). The original author of this package had started to solve this problem by "manually" installing a dependency (mongodb-async-driver), but more dependencies would need to be installed in a similar manner for the package to properly install. Until the authors of YCSB fix their maven build files, the best way to solve this problem is to restrict the version of maven that the spack package requires. This, incidentally, removes the need for the mongodb-async-driver dependency since maven is now able to resolve it by itself.

Another problem with the previous version of this package is that its installation procedure simply copied the whole source tree into the installation prefix. This is not the right way to install YCSB: when doing so, YCSB thinks it is running from its source tree and consequently reinstall its dependencies!  The correct way of installing YCSB is to take the distribution archive that maven produced, and simply uncompress it at the installation prefix.

This PR solves both problems above, and also adds a master version.

Note: the `ycsb` executable that gets installed is a python script only compatible with python 2.x. I hesitated in adding a dependency to python 2 and a patch to replace the first line of the script so that it uses `python2` instead of `python`, but in the end I didn't do it because there is also a bash version of the executable (`ycsb.sh`) that does the exact same thing as the python script.

### DIFF
--- a/var/spack/repos/builtin/packages/ycsb/package.py
+++ b/var/spack/repos/builtin/packages/ycsb/package.py
@@ -56,9 +56,10 @@ class Ycsb(MavenPackage):
         # The build process builds a tar.gz archive in distribution/target
         # that can easily be installed by untaring it into the install prefix.
         from glob import glob
+
         distribution = "distribution/target/ycsb-*.tar.gz"
         with working_dir(self.build_directory):
             dist_file = glob(distribution)[0]
         tar = which("tar")
-        tar('xf', dist_file, '-C', prefix, '--strip-components=1')
+        tar("xf", dist_file, "-C", prefix, "--strip-components=1")
 
--- a/var/spack/repos/builtin/packages/ycsb/package.py
+++ b/var/spack/repos/builtin/packages/ycsb/package.py
@@ -14,27 +14,51 @@ class Ycsb(MavenPackage):
     url = "https://github.com/brianfrankcooper/YCSB/archive/0.17.0.tar.gz"
     git = "https://github.com/brianfrankcooper/YCSB.git"
 
+    version("master", branch="master")
     version("0.17.0", sha256="5dd1a3d4dd7ac336eadccc83b097c811e142cfe1b23fc278f247054a1892c0e0")
     version("0.16.0", sha256="4296fd5e90d7d6d7dfcbad90039ddf16e785706a07f99c1c8a06e6ee06440f71")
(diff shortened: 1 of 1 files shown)


adamjstewart:
I don't know anything about YCSB or Maven, I'm really only maintaining Python/geospatial packages.

mdorier:
Sorry, you were one of the people suggested by github to review, so I just clicked :-)

var/spack/repos/builtin/packages/ycsb/package.py
```diff
@@ -14,27 +14,51 @@ class Ycsb(MavenPackage):
...
+        # )
+        # path = join_path(self.spec["mongodb-async-driver"].prefix, jar_name)
+        # mvn(
+        #     "install:install-file",
+        #     "-Dfile={0}".format(path),
+        #     "-DgroupId=com.allanbank",
+        #     "-DartifactId=mongodb-async-driver",
+        #     "-Dversion=%s" % spec["mongodb-async-driver"].version.string,
+        #     "-Dpackaging=jar",
+        # )
```
tldahlgren, line 52: Is there a reason to keep this commented out code around?

tldahlgren (commented):
Confirmed branch exists.  We generally don't like to leave a lot of comments, especially commented out code, in packages.

tldahlgren:
@darmac I realize it's been a while since you added this package, but are you interested in reviewing this PR?

var/spack/repos/builtin/packages/ycsb/package.py
mdorier: I have explained why I kept this commented in the previous comment block.

This code corresponds to the way the `mongodb-async-driver` dependency was "manually" handled prior to my changes. This manual build was required because `mongodb-async-driver` is located at a URL that maven 3.8.4+ cannot pull from. My changes force the use of maven up to 3.6.3, so there is no problem with `mongodb-async-driver` anymore, but should someone want to update the package to support maven 3.8.4+ in the future, they would have to re-enable this manual build of `mongodb-async-driver`, and also manually build other dependencies that maven cannot handle by itself.

I could simply remove all the comments, and assume that build failures with maven 3.8.4+ should be up to the YCSB authors to fix. Just let me know.

tldahlgren (approved):

PR #32219, opened 2022-08-17
CCache: `redis` variant & macOS disable

ax3l:
On macOS, `ccache` is broken for years. The reason for that is that macOS seems to ship a system REDIS library that takes precedence on the final, linked binary when executed - which leads to crashes. (Could also be brew - either way, Spack does not RPath the installed binary it seems.)

I don't know how to fix linkage loader/rpaths stuff on macOS, so I add redis now as a variant and turn it off by default on macOS.

This fix is needed by half of my colleagues doing code development.

### DIFF
--- a/var/spack/repos/builtin/packages/ccache/package.py
+++ b/var/spack/repos/builtin/packages/ccache/package.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import re
-import sys
 
 from spack.package import *
 
@@ -44,10 +43,7 @@ class Ccache(CMakePackage):
     version("3.3", sha256="b220fce435fe3d86b8b90097e986a17f6c1f971e0841283dd816adb238c5fd6a")
     version("3.2.9", sha256="1e13961b83a3d215c4013469c149414a79312a22d3c7bf9f946abac9ee33e63f")
 
-    is_darwin = sys.platform == "darwin"
-
-    # macOS ships a redis lib by default, which at runtime crashes ccache
-    variant("redis", default=not is_darwin, description="Enable Redis secondary storage")
+    variant("redis", default=True, description="Enable Redis secondary storage")
 
     depends_on("cmake@3.15:", when="@4.7:", type="build")
     depends_on("cmake@3.10:", when="@4.4:", type="build")
--- a/var/spack/repos/builtin/packages/ccache/package.py
+++ b/var/spack/repos/builtin/packages/ccache/package.py
@@ -68,9 +68,8 @@ class Ccache(CMakePackage):
             self.define("ENABLE_DOCUMENTATION", False),
             self.define_from_variant("REDIS_STORAGE_BACKEND", "redis"),
             self.define("ZSTD_FROM_INTERNET", False),
+            self.define("HIREDIS_FROM_INTERNET", False),
         ]
-        if spec.satisfies("+redis"):
-            args.append(self.define("HIREDIS_FROM_INTERNET", False))
 
         return args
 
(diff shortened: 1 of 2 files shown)


haampie:
> On macOS, ccache is broken for years.

The ccache adding hiredis as a dep has its first birthday ... tomorrow ;p 

Can you instead make hiredis support rpaths?

```diff
diff --git a/var/spack/repos/builtin/packages/hiredis/package.py b/var/spack/repos/builtin/packages/hiredis/package.py
index 1e5bc12d03..ed0dfe4aea 100644
--- a/var/spack/repos/builtin/packages/hiredis/package.py
+++ b/var/spack/repos/builtin/packages/hiredis/package.py
@@ -19,3 +19,8 @@ class Hiredis(MakefilePackage):
 
     def install(self, spec, prefix):
         make("PREFIX={0}".format(prefix), "install")
+
+    @run_after("install")
+    def darwin_fix(self):
+        if self.spec.satisfies("platform=darwin"):
+            fix_darwin_install_name(self.prefix.lib)
```

ax3l:
Thanks!

Tried this now, but does not seem to do the trick:
```
$ otool -L $(which ccache)
/Users/axel/src/spack/opt/spack/darwin-monterey-m1/apple-clang-13.1.6/ccache-4.6.1-lo5t45y4y4bps7a6adkej2ai2ib4b6v5/bin/ccache:
	/Users/axel/src/spack/opt/spack/darwin-monterey-m1/apple-clang-13.1.6/zstd-1.5.2-r7rfcbo5v255mqhxwx3ohioih4obbif2/lib/libzstd.1.dylib (compatibility version 1.0.0, current version 1.5.2)
	/usr/local/lib/libhiredis.0.14.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

ax3l:
But it works if I add the same patch to `ccache`, too!

ax3l:
~~Ah no, spack reused the old `hiredis` package.all good~~ 

This works, thanks a lot!

haampie (approved):

var/spack/repos/builtin/packages/ccache/package.py
```diff
@@ -53,14 +55,24 @@ class Ccache(CMakePackage):
...
-        return [self.define("ENABLE_TESTING", False)]
+        spec = self.spec
+        args = [
+            self.define("ENABLE_TESTING", False),
+            self.define("ENABLE_DOCUMENTATION", False),
+            self.define_from_variant("REDIS_STORAGE_BACKEND", "redis"),
+            self.define("ZSTD_FROM_INTERNET", False),
+        ]
+        if spec.satisfies("+redis"):
+            args.append(self.define("HIREDIS_FROM_INTERNET", False))
```
haampie: Uh, this should remain unconditionally to avoid cmake from downloading it.

haampie (changes requested):
can you leave the HIREDIS_FROM_INTERNET bits?

var/spack/repos/builtin/packages/ccache/package.py
haampie: Ah, not really. But it can't hurt to set this unconditionally anyways.

ax3l: As of 4.6.1, this one is in an `if()` that would otherwise be an unused variable:
https://github.com/ccache/ccache/blob/v4.6.1/CMakeLists.txt#L103

This changed in master though:
https://github.com/ccache/ccache/blob/master/CMakeLists.txt#L115-L116

ax3l:
I can change it - but for the currently tagged releases, this can result in an unused variable warning in some cases (`~redis`).

haampie (approved):

PR #32220, opened 2022-08-17
pasta: missing setuptools dep

snehring:
pasta needs setuptools. I guess the python I was using last week to build this had that already, caught it today in a fresh python.

### DIFF
--- a/var/spack/repos/builtin/packages/pasta/package.py
+++ b/var/spack/repos/builtin/packages/pasta/package.py
@@ -3,10 +3,12 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from os import unlink
+
 from spack.package import *
 
 
-class Pasta(Package):
+class Pasta(PythonPackage):
     """PASTA (Practical Alignment using SATe and Transitivity)"""
 
     homepage = "https://github.com/smirarab/pasta"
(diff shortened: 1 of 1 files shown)


tldahlgren (dismissed):
LGTM.  Will give @adamjstewart a chance to comment before merging.

var/spack/repos/builtin/packages/pasta/package.py
```diff
@@ -19,6 +19,7 @@ class Pasta(Package):
     depends_on("python@3:", when="@1.9.0:")
     depends_on("python@:2", when="@:1.8.3")
     depends_on("py-dendropy")
+    depends_on("py-setuptools")
     depends_on("java")
```
adamjstewart: Guessing this should look something like:
```suggestion
    depends_on("python@3:", when="@1.9:", type=("build", "run"))
    depends_on("python@:2", when="@:1.8", type=("build", "run"))
    depends_on("py-dendropy@4:", type=("build", "run"))
    depends_on("py-setuptools", type="build")
    depends_on("java", type=("build", "run"))
```

var/spack/repos/builtin/packages/pasta/package.py
```diff
@@ -38,7 +39,5 @@ def install(self, spec, prefix):
         install_tree(".", prefix)
```
adamjstewart: Can you try removing the `install` method and using `PythonPackage`? Would like to change this for the same reasons as busco, we don't want to install all these READMEs, docs, tests, etc.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/pasta/package.py
snehring: I'll try, or at the very least make the same changes we had with busco.

var/spack/repos/builtin/packages/pasta/package.py
snehring: I'll try, or at the very least make the same changes we had with busco.

adamjstewart (approved):

tldahlgren (approved):

PR #32221, opened 2022-08-17
Update busco to 5.4.3

snehring:
Updates to the most recent busco release. Also adds in some deps that were necessary to run through the test datasets provided by the project. As part of that a few new packages are added.

I think the super calls that I changed in the busco package.py were typos introduced by #27798, they were causing errors.

### DIFF
--- a/var/spack/repos/builtin/packages/metaeuk/package.py
+++ b/var/spack/repos/builtin/packages/metaeuk/package.py
@@ -1,0 +1,20 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Metaeuk(CMakePackage):
+    """MetaEuk is a modular toolkit designed for large-scale gene discovery
+    and annotation in eukaryotic metagenomic contigs.
+    """
+
+    homepage = "https://metaeuk.soedinglab.org/"
+    url = "https://github.com/soedinglab/metaeuk/archive/refs/tags/6-a5d39d9.tar.gz"
+    maintainers = ["snehring"]
+
+    version("6-a5d39d9", sha256="be19c26f5bdb7dcdd7bc48172105afecf19e5a2e5555edb3ba0c4aa0e4aac126")
+
+    depends_on("cmake@3:", type="build")
--- a/var/spack/repos/builtin/packages/metaeuk/package.py
+++ b/var/spack/repos/builtin/packages/metaeuk/package.py
@@ -17,4 +17,4 @@ class Metaeuk(CMakePackage):
 
     version("6-a5d39d9", sha256="be19c26f5bdb7dcdd7bc48172105afecf19e5a2e5555edb3ba0c4aa0e4aac126")
 
-    depends_on("cmake@3:", type="build")
+    depends_on("cmake@2.8.12:", type="build")
--- a/var/spack/repos/builtin/packages/sepp/package.py
+++ b/var/spack/repos/builtin/packages/sepp/package.py
@@ -16,17 +16,18 @@ class Sepp(Package):
 
     version("4.5.1", sha256="51e052569ae89f586a1a94c804f09fe1b7910a3ffff7664e2005f18c7d3f717b")
 
-    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("python@3.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
+    depends_on("py-dendropy@4:", type=("build", "run"))
 
     depends_on("java@1.8:", type="run")
     depends_on("pasta@1:", type="run")
     depends_on("blast-plus@2.10.1:", type="run")
 
     def install(self, spec, prefix):
-        python = which("python")
         install_tree(".", prefix)
         with working_dir(prefix):
+            prefix_string = "--prefix=" + prefix
             python("setup.py", "config", "-c")
-            python("setup.py", "install")
+            python("setup.py", "install", prefix_string)
             python("setup.py", "upp", "-c")
--- a/var/spack/repos/builtin/packages/sepp/package.py
+++ b/var/spack/repos/builtin/packages/sepp/package.py
@@ -1,0 +1,32 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Sepp(Package):
+    """SEPP stands for SATe-enabled phylogenetic placement"""
+
+    homepage = "https://github.com/smirarab/sepp"
+
+    url = "https://github.com/smirarab/sepp/archive/refs/tags/4.5.1.tar.gz"
+    maintainers = ["snehring"]
+
+    version("4.5.1", sha256="51e052569ae89f586a1a94c804f09fe1b7910a3ffff7664e2005f18c7d3f717b")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+
+    depends_on("java@1.8:", type="run")
+    depends_on("pasta@1:", type="run")
+    depends_on("blast-plus@2.10.1:", type="run")
+
+    def install(self, spec, prefix):
+        python = which("python")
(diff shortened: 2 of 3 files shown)


var/spack/repos/builtin/packages/sepp/package.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Sepp(Package):
```
adamjstewart, line 9: Have you tried using the normal `PythonPackage` base class?

var/spack/repos/builtin/packages/metaeuk/package.py
```diff
@@ -0,0 +1,20 @@
...
+    and annotation in eukaryotic metagenomic contigs.
+    """
+
+    homepage = "https://metaeuk.soedinglab.org/"
+    url = "https://github.com/soedinglab/metaeuk/archive/refs/tags/6-a5d39d9.tar.gz"
+    maintainers = ["snehring"]
+
+    version("6-a5d39d9", sha256="be19c26f5bdb7dcdd7bc48172105afecf19e5a2e5555edb3ba0c4aa0e4aac126")
+
+    depends_on("cmake@3:", type="build")
```
adamjstewart: According to `CMakeLists.txt` it's 2.8.12+

var/spack/repos/builtin/packages/sepp/package.py
```diff
@@ -0,0 +1,32 @@
...
+    homepage = "https://github.com/smirarab/sepp"
+
+    url = "https://github.com/smirarab/sepp/archive/refs/tags/4.5.1.tar.gz"
+    maintainers = ["snehring"]
+
+    version("4.5.1", sha256="51e052569ae89f586a1a94c804f09fe1b7910a3ffff7664e2005f18c7d3f717b")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+
```
adamjstewart, line 22: MIssing build/run dep on py-dendropy 4+

var/spack/repos/builtin/packages/sepp/package.py
```diff
@@ -0,0 +1,32 @@
...
+    """SEPP stands for SATe-enabled phylogenetic placement"""
+
+    homepage = "https://github.com/smirarab/sepp"
+
+    url = "https://github.com/smirarab/sepp/archive/refs/tags/4.5.1.tar.gz"
+    maintainers = ["snehring"]
+
+    version("4.5.1", sha256="51e052569ae89f586a1a94c804f09fe1b7910a3ffff7664e2005f18c7d3f717b")
+
+    depends_on("python@3.7:", type=("build", "run"))
```
adamjstewart: Where do you see the 3.7+ restriction?

var/spack/repos/builtin/packages/sepp/package.py
snehring: I did. It does not install nicely with pip.

var/spack/repos/builtin/packages/metaeuk/package.py
snehring: Was just running off their readme, but I can change that.

adamjstewart: Nah, that's fine

var/spack/repos/builtin/packages/sepp/package.py
snehring: I absolutely read the python dep in https://github.com/smirarab/sepp/blob/master/README.HIPPI.md for 2.7 as 3.7.

var/spack/repos/builtin/packages/sepp/package.py
```diff
@@ -0,0 +1,32 @@
...
+    depends_on("py-setuptools", type="build")
+
+    depends_on("java@1.8:", type="run")
+    depends_on("pasta@1:", type="run")
+    depends_on("blast-plus@2.10.1:", type="run")
+
+    def install(self, spec, prefix):
+        python = which("python")
```
adamjstewart: Some installations only install `python3`, not `python`. There should actually already be a `python` executable available so you can remove this line.

var/spack/repos/builtin/packages/sepp/package.py
```diff
@@ -0,0 +1,32 @@
...
+        install_tree(".", prefix)
+        with working_dir(prefix):
+            python("setup.py", "config", "-c")
+            python("setup.py", "install")
+            python("setup.py", "upp", "-c")
```
adamjstewart: Instead of `-c` is it possible to use `--prefix=<prefix>`? I would like to avoid copying all the random text files from the root of the project if possible.

snehring: Without the -c it'll write a config file in ~/.sepp/ for the installing user. I can see if it still works as expected with busco though.

adamjstewart: Ah I thought `-c` simply meant install locally instead of to the Python installation prefix. Maybe combine `-c` and `--prefix` then?

snehring: I think it's picking up the prefix correctly as is, in that it is making a busco in the site-packages dir as expected for the package prefix. All -c seems to do is control where the 'config' lives, either in the user home or in the directory setup.py is run from. I can add it in explicitly though.

adamjstewart: It's picking up the prefix only because you manually copy the entire tarball to the prefix. What if we run `python setup.py -c --prefix=<prefix>` from the stage directory instead of the installation directory (remove the install_tree)?

snehring: Ugh ignore above about busco, other package obviously. The -c stuff holds true though, and omitting it it causes issues for people other than the installing user.

snehring: The config path situation for sepp is messy. Let me see if I can improve it any.

snehring: Not copying the entire tree doesn't seem to be an option. Not without changes to how the setup.py works. It hard codes a lot of paths based on where it's executed which then get passed into the config.

var/spack/repos/builtin/packages/sepp/package.py
```diff
@@ -0,0 +1,33 @@
...
+    depends_on("py-setuptools", type="build")
+    depends_on("py-dendropy@4:", type=("build", "run"))
+
+    depends_on("java@1.8:", type="run")
+    depends_on("pasta@1:", type="run")
+    depends_on("blast-plus@2.10.1:", type="run")
+
+    def install(self, spec, prefix):
+        install_tree(".", prefix)
+        with working_dir(prefix):
```
adamjstewart, line 29: Does this still work if you remove these lines? We don't really want to copy all of the READMEs to the root of the installation prefix, we really only want `prefix.bin` and `prefix.lib`.

var/spack/repos/builtin/packages/sepp/package.py
adamjstewart: That's unfortunate. Could we delete everything that isn't `prefix.bin` or `prefix.lib` so that there aren't a ton of README files in the installation prefix? Those will likely be ugly once you try to create a single symlinked view.

snehring: Yeah, I'll see if I can pare it down from the whole tree to a specific set of files.

snehring: Much cleaner result now.

var/spack/repos/builtin/packages/sepp/package.py
```diff
@@ -0,0 +1,61 @@
...
+            ]
+            for f in remove_files:
+                force_remove(f)
+            dirs = ["sepp", "tools", "build", "dist", "__pycache__", "sepp.egg-info"]
+            for d in dirs:
+                remove_linked_tree(d)
+
+    def setup_run_environment(self, env):
+        python_lib_path = self.spec["python"].package.platlib
+        env.set("PYTHONPATH", join_path(self.prefix, python_lib_path))
```
adamjstewart: If you add:
```python
extends("python")
```
above you can remove this `setup_run_environment` method.

snehring: Right, brainfart.

adamjstewart (approved):

PR #32224, opened 2022-08-17
HIP/HIP-rocclr: GLX is required by HIP

kwryankrattiger:
https://github.com/ROCm-Developer-Tools/ROCclr/issues/37

### DIFF
--- a/var/spack/repos/builtin/packages/hip-rocclr/package.py
+++ b/var/spack/repos/builtin/packages/hip-rocclr/package.py
@@ -42,6 +42,7 @@ class HipRocclr(CMakePackage):
     depends_on("cmake@3:", type="build")
     depends_on("gl@4.5:", type="link")
     depends_on("numactl", type="link")
+    depends_on("glx", type="build")
 
     for ver in [
         "5.1.0",
--- a/var/spack/repos/builtin/packages/hip-rocclr/package.py
+++ b/var/spack/repos/builtin/packages/hip-rocclr/package.py
@@ -42,7 +42,7 @@ class HipRocclr(CMakePackage):
     depends_on("cmake@3:", type="build")
     depends_on("gl@4.5:", type="link")
     depends_on("numactl", type="link")
-    depends_on("glx", type="build")
+    requires("^[virtuals=gl] glx")
 
     for ver in [
         "5.1.0",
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/hip/package.py
```diff
@@ -55,6 +55,8 @@ class Hip(CMakePackage):
 
     depends_on("cmake@3.16.8:", type="build")
     depends_on("perl@5.10:", type=("build", "run"))
+    depends_on("gl@4.5:")
+    depends_on("glx")
```
alalazo: Is `glx` the required `gl` provider?

kwryankrattiger: Correct, It needs headers from both the GLX and GL so they are both listed.

var/spack/repos/builtin/packages/hip-rocclr/package.py
```diff
@@ -42,6 +42,7 @@ class HipRocclr(CMakePackage):
     depends_on("cmake@3:", type="build")
     depends_on("gl@4.5:", type="link")
     depends_on("numactl", type="link")
+    depends_on("glx", type="build")
```
alalazo:
```suggestion
    requires("^[virtuals=gl] glx")
```

var/spack/repos/builtin/packages/hip/package.py
alalazo:
```suggestion
    requires("^[virtuals=gl] glx")
```

alalazo (commented):
The suggested changes should ensure that `glx` is used as the `gl` provider in a DAG. Otherwise, it might be considered as a "regular" package, and another provider may be used to satisfy the virtual dependency.

alalazo:
Is this still relevant?

kwryankrattiger:
I believe so, but it is flawed because we don't need GLX we need a header from GLX and that is it. @haampie has been looking at this as well and may have a better solution in mind.

haampie:
Hi @kwryankrattiger, this package-related pull request can be migrated from `spack/spack` to `spack/spack-packages` using the [**migrate-pkg-prs utility**](https://github.com/spack/migrate-package-prs).

---

PR #32227, opened 2022-08-17
e4s oneapi: compiler environment: prepend lib dir to LD_LIBRARY_PATH

eugeneswalker:
For our E4S OneAPI environment, prepend OneAPI's lib dir to LD_LIBRARY_PATH.
* Uncomment `gptune` as this build may be fixed by this change

Thank you @rscohn2 

FYI @wspear

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/e4s-oneapi/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/e4s-oneapi/spack.yaml
@@ -25,7 +25,9 @@ spack:
       operating_system: ubuntu20.04
       target: x86_64
       modules: [compiler]
-      environment: {}
+      environment:
+        prepend_path:
+          LD_LIBRARY_PATH: /opt/intel/oneapi/compiler/2022.1.0/linux/compiler/lib/intel64_lin
       extra_rpaths: []
   - compiler:
       spec: oneapi@2022.1.0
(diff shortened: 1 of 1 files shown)


share/spack/gitlab/cloud_pipelines/stacks/e4s-oneapi/spack.yaml
```diff
@@ -38,7 +40,9 @@ spack:
       operating_system: ubuntu20.04
       target: x86_64
       modules: [compiler]
-      environment: {}
+      environment:
+        prepend_path:
+          LD_LIBRARY_PATH: /opt/intel/oneapi/compiler/2022.1.0/linux/compiler/lib/intel64_lin
```
alalazo, line 45: Is this affecting the binaries that get generated in any way? For instance, can people relocate them is they end up in a public buildcache?

alalazo (dismissed):

share/spack/gitlab/cloud_pipelines/stacks/e4s-oneapi/spack.yaml
rscohn2: It should not affect the generated binaries. It is needed because the compiler needs to find its own shared libraries during compile-time. The generated binaries will use the runtime value of the LD_LIBRARY_PATH to find the libraries.

alalazo: That is fine, I was wondering if any of the library in there will get linked into an artifact and not RPATHed.

rscohn2 (approved):

PR #32230, opened 2022-08-18
root: new bugfix version 6.26.06

wdconinc:
Bugfix release, https://github.com/root-project/root/compare/v6-26-04...v6-26-06, no build system changes.

@vvolkl @drbenmorgan

### DIFF
--- a/var/spack/repos/builtin/packages/root/package.py
+++ b/var/spack/repos/builtin/packages/root/package.py
@@ -82,6 +82,18 @@ class Root(CMakePackage):
     # 6.16.00 fails to handle particular build option combinations, _cf_
     # https://github.com/root-project/ROOT/commit/e0ae0483985d90a71a6cabd10d3622dfd1c15611.
     patch("root7-webgui.patch", level=1, when="@6.16.00")
+    # 6.26.00:6.26.06 fails for recent libc versions when ROOT7 is enabled
+    patch(
+        "https://github.com/root-project/root/pull/11111.patch?full_index=1",
+        sha256="3115be912bd948979c9c2a3d89ffe6437fe17bd3b81396958c6cb6f51f64ae62",
+        when="@6.26:6.26.06 +root7",
+    )
+    # 6.26.00:6.26.06 fails for recent nlohmann-json single headers versions
+    patch(
+        "https://github.com/root-project/root/pull/11225.patch?full_index=1",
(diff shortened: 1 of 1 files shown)


HadrienG2 (dismissed):

vvolkl (dismissed):

alalazo (dismissed):

alalazo:
I'll let this one finish. so the builds in the pipelines are not lost, and then retrigger the pipeline (at least one is failing due to backend issues)

wdconinc:
Disabled auto-merge since testing with some more non-default variants uncovered two issues.

1. With `+root7` and `libc >= 2.35` this fails with reported issue https://github.com/root-project/root/issues/11130 (see original bug report), fixed by the following [patch](https://github.com/root-project/root/pull/11111) as merged into root master and backported to 6.26 (applied for all libc versions, and for any root version after https://github.com/root-project/root/commit/5db99109f8a09aa5b9af41ae68b0918928e64e75):
```python
    patch("https://github.com/root-project/root/pull/11111.patch?full_index=1",
        sha256="3115be912bd948979c9c2a3d89ffe6437fe17bd3b81396958c6cb6f51f64ae62",
        when="@6.25.02:6.26.06 +root7")
```

2. With `+root7` and `^nlohmann-json@3.11:` this fails (same issue report), fixed by the following [patch](https://github.com/root-project/root/pull/11205), not merged yet (applies cleanly for 6.26.00 onwards):
```python
    patch("https://github.com/root-project/root/pull/11205.patch?full_index=1",
        sha256="f894427be924aa9a68fc3da6425f50c9308f366d2f154a7469d504f30f39bcc1",
        when="@6.26:6.26.06 +root7 ^nlohmann-json@3.11:")
```

alalazo:
Ping me back when this is ready to merge

wdconinc:
I was hesitating between submitting the patches separately, but it makes sense to include them here since this is the update that will probably result in people recompiling root (especially with reuse). The two patches apply to 6.26.00 through this new bugfix version.
- The first one became an issue with a bugfix-level glibc 2.35 upgrade of the underlying ubuntu jammy system in July (as far as I can archeologize).
- The seond one only became an issue with the update of nlohmann-json to 3.11 in spack.

HadrienG2 (approved):

alalazo (approved):

PR #30768, opened 2022-05-20
autotools: extend patching of the libtool script

skosukhin:
This introduces the argument `start_at` for the `filter_file` function. The value of the argument is used as a marker to start applying the replacements. If a text line matches this marker filtering is started at the next line. All contents before the marker and the marker itself are copied verbatim. The default is to start filtering from the first line of the file.

This feature allows for more selective filtering of the libtool script. For example, currently, the empty PIC flag entries in a libtool script are filled with the flag for the C compiler (`cc_pic_flag`). The suggested modification fills the values with the language-specific flags (see NAG).

I have also extended the method `_do_patch_libtool` with fixes that currently live separately in several packages (e.g. `mpich`, `openmpi`, `netcdf-fortran`, `parallel-netcdf`, etc.). I plan to go over those packages later and remove the patches there but it looks like I have managed to implement the filtering in a way it can co-exist with similar patches that either are applied with Spack or come with the packages (e.g. `openmpi`, `pmix`).

PR #32232, opened 2022-08-18
autoconf-archive: fetch patch from github

skosukhin:
This removes a patch file from the repo and fetches it from github.

It also disables the libtool patching, which does not make much difference at the moment but will speed up the installation once #30768 is merged (note that, although this also cures the symptom https://github.com/spack/spack/pull/30768#issuecomment-1219135245, the real solution will be implemented in the aforementioned pull request).


tldahlgren (approved):
Confirmed patch exists and matches one being removed.

PR #32236, opened 2022-08-18
Remove unused argument in test fixture

alalazo:
The argument is very likely a typo, and was meant to be given to the fixture decorator. Since the value being passed is the default, let's just remove it.


vsoch (approved):
I don't see it used anywhere, so seems ok to me!

tldahlgren (approved):

PR #32239, opened 2022-08-18
ascent: fix oneapi build

rscohn2:
oneapi build is broken for 0.8.0. It is fixed in develop: https://github.com/Alpine-DAV/ascent/pull/981
The fix in develop depends on other commits so using a temporary workaround until it is released.


eugeneswalker (approved):
Thank you!

cyrush:
thanks @rscohn2 !

PR #32246, opened 2022-08-18
Refactor CI QA checks to separate GHA workflow

johnwparent:
Rather than running sets of QA CI checks on each platform (Unix/Windows), run the QA tests as their own workflow, and run platform specific workflows only on success of QA workflow.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -1,103 +1,24 @@
 name: linux tests
 
 on:
-  push:
-    branches:
-      - develop
-      - releases/**
-  pull_request:
-    branches:
-      - develop
-      - releases/**
+  workflow_run:
+    workflows: [style and validate]
+    types:
+      - completed
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
   cancel-in-progress: true
 
 jobs:
-  # Validate that the code can be run on all the Python versions
-  # supported by Spack
-  validate:
-    runs-on: ubuntu-latest
+  style-check:
+    runs-on: windows-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     steps:
-    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # @v2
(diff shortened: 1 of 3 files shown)


.github/workflows/windows_python.yml
```diff
@@ -1,14 +1,10 @@
...
-  push:
-    branches:
-      - develop
-      - releases/**
-  pull_request:
-    branches:
-      - develop
-      - releases/**
+  workflow_run:
+    workflows: [style and validate]
```
scheibelp, line 5: I think it would be better to use a name like `style-and-validate` or `style_and_validate` (even if the spaces are allowed in the name, I found it hard to parse this by eye).

.github/workflows/unit_tests.yaml
```diff
@@ -1,103 +1,24 @@
...
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
   cancel-in-progress: true
 
 jobs:
-  # Validate that the code can be run on all the Python versions
-  # supported by Spack
-  validate:
+  style-check:
```
scheibelp, line 14: I think this should be removed entirely (vs. right now it essentially appears to be a noop). In that case, there is no longer a need for a `needs:` section to many of these checks.

scheibelp (changes requested):
This looks good although I have a couple of cleanup requests.

.github/workflows/unit_tests.yaml
johnwparent: From what I understood from the GHA docs, this is a required workaround to achieve the workflow we want from the proposed setup of workflow files. In order to have one workflow depend on another in GHA, the `on` trigger should be set off by the run of a workflow (the style checks). If we then want the run of the subsequent workflow (the unit tests) to be dependent not only on the completion of the initial workflow, but the success of the contents, we need a 'job' in the dependent workflow to check for success of the completed job. I think this job could be cleaned up/clarified, but we cannot discard it entirely. 

That being said I may have missed something in my pass over the docs. I'll look again and update my comment if need be.

alalazo:
Duplicate of #32183

johnwparent:
Closing as duplicate was merged.

PR #32253, opened 2022-08-18
Fix relocation to avoid overwriting merged constant strings

trws:
This is the fix for issues observed in #31739, and causing failures to build the mesa package possibly since last September.  The observed symptom of the problem was that various floating point constants in `src/util/format_srgb.c` in mesa were producing syntax errors in exactly one configuration, of two, on ARM graviton nodes on AWS and absolutely nowhere else.  Specifically numbers that should have been `7.6052450e-01` were `7.6052450/-01` instead.  Thanks to some poking from @tgamblin, we determined that source file was generated by python, and tracked down the issue to python's `%` formatting:

```python-console
>>> i = 7.6052450e-01
>>> i
0.7605245
>>> "%.7Ef" % i
'7.6052450E-01f'
>>> "%.7ef" % i
'7.6052450/-01f'
```

Note that `E` formatting works correctly, but `e` does not.  The formatting logic comes from `pystrtod.c` in cpython, and uses [a pair of global string arrays](https://github.com/python/cpython/blob/main/Python/pystrtod.c#L966-L975) to differentiate between upper- and lower-case strings for floating point formatting.  The `lc_float_strings` table in `libpython.so` was corrupted:

```
(gdb) p lc_float_strings
$1 = {0x2e1528 "inf", 0x2f5920 "nan", 0x2f6f88 "/"}
```

More interesting was how, the original version that had it correct, when searching back from the start address of the "e" string showed that the linker re-used the "e" *at the end of the spack package prefix* as the string constant for e:

```
(gdb) p lc_float_strings
$1 = {0x2e1528 "inf", 0x2f5920 "nan", 0x2f6f88 "e"}
$6 = 0x2f6f50 "/gcc-7.3.1/python-3.9.13-g3acg4n4bzr2o5wuukebasdfkjym2x5e"
```

This happens whenever `-fmerge-constants` is active in gcc, clang etc. which it is for `-O1` and higher, so since approximately September of 2021 we have likely shipped uniquely broken binaries of python whenever the hash happened to end in "e".  The reason is that our relocation logic, originally designed to pad the new prefix with `os.sep` characters, was accidentally refactored to *suffix* with them instead.

After much discussion, we determined that ensuring splicing preserved a suffix, shortening when possible, and prefix padding with separators when not would be the best course of action.  This PR implements the hash suffix preservation for splicing with a length of seven, shortens the prefix and terminates with a null without overwriting the rest when the prefix is immediately followed by a null byte, and after that prefix pads whatever matches remain.

If we find that the prefix padding is a usability problem, we can shorten cases where the null terminator is further away as long as the intervening space doesn't contain another copy of the prefix, but there are risks, so we're starting with this.


lib/spack/spack/binary_distribution.py
```diff
@@ -1588,7 +1588,7 @@ def relocate_package(spec, allow_root):
     prefix_to_prefix_text[old_prefix] = new_prefix
     prefix_to_prefix_bin[old_prefix] = new_prefix
     prefix_to_prefix_text[old_layout_root] = new_layout_root
-    prefix_to_prefix_bin[old_layout_root] = new_layout_root
+    # prefix_to_prefix_bin[old_layout_root] = new_layout_root
```
haampie: I found this one odd too. Was it intended as a generic fallback? In that case it belongs below the loop over spec prefixes

trws: I'm honestly not sure, all the tests pass without it (though I'm still fixing up the rework for a single regex version).

trws:
Ok, this should be good to go pending comments.  I put the check for non-matching suffixes back in at a length of 7. This should not happen, but if someone puts something other than hash last in their projection and tries to relocate, they'll get an explanation rather than crazy failures.

trws:
This however is not great news, apparently bootstrap in CI relocates to *longer paths*? https://github.com/spack/spack/actions/runs/3374819566/jobs/5600910914#step:4:433

tgamblin:
@trws can you rebase this on `develop` and update the PR message with what we talked about in the meeting Tuesday (basically that you're not handling shortening when there are multiple instances of the prefix/hash)?

trws:
I'm merging rather than rebasing at the moment, can rebase to clean this up in a bit once we work out the  bootstrap situation.  Updating the description momentarily.

trws:
This should work with rewiring again now.  There's a case that we neglected to consider in rewiring, where the package being rewired needs to replace the paths *of other packages* that are not having their hashes replaced and thus we can't frankenhash.  For right now, I have the check failing these only if the prefix is followed immediately by a null byte, because that's the only case when that suffix could have been reused.  For all the cases we test in the repository, that works out fine, but I have no way to predict how often that will be a problem for rewiring in the wild.  We should probably consider solutions to that for the future.

alalazo:
@spackbot rebuild everything

lib/spack/spack/binary_distribution.py
haampie: I'll put this back in place, but after the prefix loop

lib/spack/spack/relocate.py
```diff
@@ -467,39 +480,66 @@ def _replace_prefix_bin(filename, byte_prefixes):
...
+            old = escaped_null.sub(b"", found_full)
+            terminated = old != found_full
+
+            new = byte_prefixes[old]
 
-        # Apply the replacements
-        for replacement, offset in replacements_at_offset:
-            f.seek(offset)
-            f.write(replacement)
+            if len(new) > len(old) or (terminated and new[-7:] != old[-7:]):
```
haampie: I'm confused, why require a common suffix? You get `len(old) - len(new) - 1` unused bytes after replacement + new null byte, so a new suffix is possible when that's >= 7. And if you require a common suffix, you can just as well pad with leading `/`, like in the non-null terminated case.

Right now if the the new length is 1 byte shorter and the suffix has 7 bytes in common, it's still broken. E.g. `/abc/1234567` -> `/ab/1234567` would turn `/abc/1234567\0` into `/ab/1234567\0\0`, which is incorrect.

trws: Darn it, you're right, the check isn't quite right anymore.

The suffix case is mainly to deal with same-length or shorter than safe width difference prefixes when terminated.  We can't do anything if the suffixes aren't the same.  It also impacts padding in similar cases. Probably it should be:
* if len(new) - len(old)  > 7 and terminated: ok, shorten
* if len(new) - len(old)  > 7 and not terminated: ok, pad
* if len(new) - len(old)  < 7 and terminated and matching: ok, pad
* if len(new) - len(old)  < 7 and terminated and not matching: 💥 
* if len(new) - len(old)  < 7 and not terminated and matching: ok, pad
* if len(new) - len(old)  < 7 and not terminated and not matching: may be ok to pad, depends on length of suffix

Need to add tests so we cover all of these.  Honestly kinda want to fuzz the 💩 out of this too.

haampie: I'm happy to push a few commits when you're on holidays ;p

haampie: How about

```python
if len(new) > len(old):
  raise NoInplaceUpdatePossible()
elif not null_terminated or old[-N:] == new[-N:]:
  replacement = b"/" * (len(old) - len(new)) + new
elif len(new) + N < old:  # needs < cause of \0 after new
  replacement = new + b"\0"
else:
  raise AliasingRiskIssues()
```

haampie: I added tons of tests now, and also introduced a lookahead of 7 bytes for a terminating null instead of direct nulls. That covers the case of replacing e.g. `/pkg-<hash>` in `/pkg-<hash>/\0` when the replacement has a different suffix (for some reason).

lib/spack/spack/relocate.py
```diff
@@ -463,43 +470,116 @@ def _replace_prefix_text(filename, compiled_prefixes):
...
+    files. See :func:`~spack.relocate.apply_binary_replacements` for details.
 
     Args:
         filename (str): target binary file
-        byte_prefixes (OrderedDict): OrderedDictionary where the keys are
-        binary strings of the old prefixes and the values are the new
-        binary prefixes
+        byte_prefixes (OrderedDict): ordered dictionary where the keys are
+            bytes representing the old prefixes and the values are the new
+        prefixes (all bytes utf-8 encoded)
```
trws:
```suggestion
        byte_prefixes (OrderedDict): ordered dictionary where the keys are
            bytes representing the old prefixes and the values are the new
            prefixes (all bytes utf-8 encoded)
```

lib/spack/spack/relocate.py
trws: Awesome! I'm floored @haampie, this looks fabulous.  I threw in one suggestion to fix an indentation bug but pending a defensive recheck on gitlab I think this is GTG.

trws:
@spackbot rebuild everything

haampie (approved):

haampie:
Rebuild everything worked except for the tau shasum which was fixed on develop 34724cae87623a36de0c34e6e6bba55e969c4a96

PR #32254, opened 2022-08-18
kassiopeia: new version 3.8.2

wdconinc:
There was no version 3.8.1, so here is the diff with 3.8.0: https://github.com/KATRIN-Experiment/Kassiopeia/compare/v3.8.0...v3.8.2

Build system changes:
- default `cxxstd` is now 17, or whatever `root` uses.

### DIFF
--- a/var/spack/repos/builtin/packages/kassiopeia/package.py
+++ b/var/spack/repos/builtin/packages/kassiopeia/package.py
@@ -63,7 +63,7 @@ class Kassiopeia(CMakePackage):
         if "+root" in self.spec:
             cxxstd = self.spec["root"].variants["cxxstd"].value
         else:
-            if "@:3.8.1" in self.spec:
+            if self.spec.satisfies("@:3.8.1"):
                 cxxstd = "14"
             else:
                 cxxstd = "17"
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/kassiopeia/package.py
```diff
@@ -62,7 +63,10 @@ def cmake_args(self):
         if "+root" in self.spec:
             cxxstd = self.spec["root"].variants["cxxstd"].value
         else:
-            cxxstd = "14"
+            if "@:3.8.1" in self.spec:
```
tldahlgren: We generally recommend using `self.spec.satisfies()` for range checks.

```suggestion
            if self.spec.satisfies("@:3.8.1"):
```

var/spack/repos/builtin/packages/kassiopeia/package.py
```diff
@@ -18,6 +18,7 @@ class Kassiopeia(CMakePackage):
     maintainers = ["wdconinc"]
 
     version("main", branch="main")
+    version("3.8.2", sha256="9da59697365540a8b0d66e9a63f57aff6d868f7c5b39dfec28bb11ec83f31527")
```
tldahlgren, line 21: Confirmed sha256.

tldahlgren (commented):
Looks good though recommend switching version range check on spec to use `satisfies()`.

tldahlgren (approved):
LGTM

PR #32257, opened 2022-08-19
Allow version to accept the '=' token without activating lexer switch

psakievich:
Closes #32184.  That PR seems to have been corrupted due to an unfortunate rebase.  

This PR will address the issue of `GitVerions` with assigned refs that inspired the creation of #32184 ~~but it will also make the parser behave in a more targeted manner when dealing with the `=` token.  This is done by giving higher precedent, targeted tokens for the cases where we want to glob several commands together (`ldlibs`, compiler flags and targets).~~ 

~~This should make the parser behave more cleanly when the `=` token is encountered since it is only the targeted tokens that will trigger the switch in behavior.~~

After back and forth it was decided that we should change the lexing specifically for versions to allow a version token to accept the entire string.  This adds additional complexity because there are two layers of lexing:

1. Shell lexing
2. Spack's SpecLexer

These two lexers treat spaces differently and so additional logic is required when there are spaces around the `@`. Another side effect is that different combinations of tokens can ultimately parse the same spec. A fun exercise is to go through the git history of this PR and see how I discover this over time.

The net result should be the same as #32437 minus the extra `@` in the token.  If this is acceptable then I will change #32437 to just be the fixes for submodules with GitRef versions and the other small bug fixes.

@becker33 and @tgamblin


psakievich:
@spackbot fix style

lib/spack/spack/spec.py
```diff
@@ -5304,16 +5306,28 @@ def variant(self, name=None):
...
-            start = self.token.value
-            if self.accept(EQ):
-                # This is for versions that are associated with a hash
-                # i.e. @[40 char hash]=version
-                start += self.token.value
-                self.expect(VAL)
-                start += self.token.value
+
+        def str_translate(value):
+            # return None for empty strings since we are dealing with data from split
```
psakievich:
```suggestion
            # return None for empty strings since we can end up with `'@'.strip('@')`
```

psakievich:
Pinging @becker33 and @tgamblin

psakievich:
@spackbot fix style

psakievich:
@spackbot re-run pipeline

becker33 (approved):

PR #32258, opened 2022-08-19
xpmem: fix configure with ~kernel-module

iarspider:


skosukhin:
Oops! Thank you!

haampie (approved):

haampie:
LGTM, but for the future let's be explicit:
```python
args.extend(self.enable_or_disable("kernel-module"))
```
defaults can change across versions.

skosukhin:
@haampie https://github.com/spack/spack/blob/8d99d331289b621bce17e964ac6d3b63cca302d8/var/spack/repos/builtin/packages/xpmem/package.py#L90

UPD. The problem is that the[ configure script](https://github.com/hpc/xpmem/blob/61c39efdea943ac863037d7e35b236145904e64d/m4/ac_path_kernel_source.m4#L49-L50) wants not `yes/no` as it should but `1/0`. I admit that we should use `enable_or_disable` but with a custom `actiavation_value`.

PR #32260, opened 2022-08-19
Allow default requirements in packages.yaml

### LINKED ISSUE #1371: Cannot set default compiler in packages.yaml
In the [Spack Configuration Documentation](http://software.llnl.gov/spack/configuration.html#concretization-preferences), it shows an example where you can set the default compiler for a package. However, this doesn't seem to work. Libpciaccess and OpenSSL cannot be built with the PGI compilers. I tried adding the following to my `~/.spack/packages.yaml`:

``` yaml
packages:
  libpciaccess:
    compiler: [gcc, intel, nag]
  openssl:
    compiler: [gcc, intel, nag]
```

The behavior I expect is that if I build `curl`, it will build everything with GCC. If I build `curl %intel`, it will build everything with Intel. And if I build `curl %pgi`, it will build curl with PGI but build OpenSSL with GCC. However, I get this behavior:

```
$ spack spec curl%pgi

  curl@7.49.1%pgi@16.5-0 arch=linux-centos6-x86_64
      ^openssl@1.0.2h%pgi@16.5-0 arch=linux-centos6-x86_64
          ^zlib@1.2.8%pgi@16.5-0 arch=linux-centos6-x86_64
```

Even when I set the default compiler to intel:

``` yaml
packages:
  libpciaccess:
    compiler: [intel]
  openssl:
    compiler: [intel]
```

I get:

```
$ spack spec curl

  curl@7.49.1%gcc@6.1.0 arch=linux-centos6-x86_64
      ^openssl@1.0.2h%gcc@6.1.0 arch=linux-centos6-x86_64
          ^zlib@1.2.8%gcc@6.1.0 arch=linux-centos6-x86_64
```

If I specifically try to build OpenSSL, then I get the right behavior:

```
$ spack spec openssl

  openssl@1.0.2h%intel@16.0.3 arch=linux-centos6-x86_64
      ^zlib@1.2.8%intel@16.0.3 arch=linux-centos6-x86_64
```

But this still doesn't help me solve the problem of trying to build OpenMPI and NetCDF with PGI.

alalazo:
fixes #1371 

This PR allow users to express default requirements in packages.yaml. These requirements are overridden if more specific requirements are present for a package.

Modifications:
- [x] Account for requirements under `all` as a fallback to package specific requirements
- [x] Updated docs
- [x] Added unit tests

### DIFF
--- a/lib/spack/docs/build_settings.rst
+++ b/lib/spack/docs/build_settings.rst
@@ -396,10 +396,33 @@ choose between a set of options using ``any_of`` or ``one_of``:
   ``mpich`` already includes a conflict, so this is redundant but
   still demonstrates the concept).
 
+You can also set default requirements for all packages under ``all``
+like this:
+
+.. code-block:: yaml
+
+   packages:
+     all:
+       require: '%clang'
+
+which means every spec will be required to use ``clang`` as a compiler. Note that
+in this case ``all`` represents a *default set of requirements*, so with a configuration
+like this:
+
+.. code-block:: yaml
+
+   packages:
+     all:
+       require: '%clang'
+     cmake:
+       require: '%gcc'
+
+Spack requires ``cmake`` to use ``gcc`` and all other nodes (including cmake dependencies)
+to use ``clang``. More in general, if there are specific package requirements, the default
+requirements under ``all`` are disregarded.
+
 Other notes about ``requires``:
 
-* You can only specify requirements for specific packages: you cannot
-  add ``requires`` under ``all``.
 * You cannot specify requirements for virtual packages (e.g. you can
   specify requirements for ``openmpi`` but not ``mpi``).
 * For ``any_of`` and ``one_of``, the order of specs indicates a
--- a/lib/spack/docs/build_settings.rst
+++ b/lib/spack/docs/build_settings.rst
@@ -405,9 +405,11 @@ like this:
      all:
        require: '%clang'
 
-which means every spec will be required to use ``clang`` as a compiler. Note that
-in this case ``all`` represents a *default set of requirements*, so with a configuration
-like this:
+which means every spec will be required to use ``clang`` as a compiler.
+
+Note that in this case ``all`` represents a *default set of requirements* -
+if there are specific package requirements, then the default requirements
+under ``all`` are disregarded. For example, with a configuration like this:
 
 .. code-block:: yaml
 
@@ -418,8 +420,7 @@ like this:
        require: '%gcc'
 
 Spack requires ``cmake`` to use ``gcc`` and all other nodes (including cmake dependencies)
-to use ``clang``. More in general, if there are specific package requirements, the default
-requirements under ``all`` are disregarded.
+to use ``clang``.
 
 Other notes about ``requires``:
 
--- a/lib/spack/spack/test/concretize_requirements.py
+++ b/lib/spack/spack/test/concretize_requirements.py
@@ -305,8 +305,6 @@ def test_default_requirements_with_all(spec_str, requirement_str, concretize_sco
     if spack.config.get("config:concretizer") == "original":
         pytest.skip("Original concretizer does not support configuration" " requirements")
 
-    # @2.3 is a deprecated versions. Ensure that any_of picks both constraints,
-    # since they are possible
     conf_str = """\
 packages:
   all:
@@ -319,3 +317,23 @@ packages:
     spec = Spec(spec_str).concretized()
     for s in spec.traverse():
         assert s.satisfies(requirement_str)
+
+
+def test_default_and_package_specific_requirements(concretize_scope, test_repo):
+    """Test that users can override a deprecated version with a requirement."""
+    if spack.config.get("config:concretizer") == "original":
+        pytest.skip("Original concretizer does not support configuration" " requirements")
+
+    conf_str = """\
+packages:
+  all:
+    require: "%gcc"
+  x:
+    require: "%clang"
+"""
+    update_packages_config(conf_str)
+
+    spec = Spec("x").concretized()
+    assert spec.satisfies("%clang")
+    for s in spec.traverse(root=False):
+        assert s.satisfies("%gcc")
--- a/lib/spack/spack/test/concretize_requirements.py
+++ b/lib/spack/spack/test/concretize_requirements.py
@@ -297,3 +297,25 @@ packages:
     s1 = Spec("y").concretized()
     assert s1.satisfies("@2.3")
     assert s1.satisfies("%gcc")
+
+
+@pytest.mark.parametrize("spec_str,requirement_str", [("x", "%gcc"), ("x", "%clang")])
+def test_default_requirements_with_all(spec_str, requirement_str, concretize_scope, test_repo):
+    """Test that users can override a deprecated version with a requirement."""
+    if spack.config.get("config:concretizer") == "original":
+        pytest.skip("Original concretizer does not support configuration" " requirements")
+
+    # @2.3 is a deprecated versions. Ensure that any_of picks both constraints,
+    # since they are possible
+    conf_str = """\
+packages:
+  all:
+    require: "{}"
+""".format(
+        requirement_str
+    )
+    update_packages_config(conf_str)
+
+    spec = Spec(spec_str).concretized()
+    for s in spec.traverse():
+        assert s.satisfies(requirement_str)
--- a/lib/spack/spack/test/concretize_requirements.py
+++ b/lib/spack/spack/test/concretize_requirements.py
@@ -301,7 +301,7 @@ packages:
 
 @pytest.mark.parametrize("spec_str,requirement_str", [("x", "%gcc"), ("x", "%clang")])
 def test_default_requirements_with_all(spec_str, requirement_str, concretize_scope, test_repo):
-    """Test that users can override a deprecated version with a requirement."""
+    """Test that default requirements are applied to all packages."""
     if spack.config.get("config:concretizer") == "original":
         pytest.skip("Original concretizer does not support configuration" " requirements")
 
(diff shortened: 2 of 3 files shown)


lib/spack/docs/build_settings.rst
```diff
@@ -396,10 +396,33 @@ choose between a set of options using ``any_of`` or ``one_of``:
...
+
+   packages:
+     all:
+       require: '%clang'
+     cmake:
+       require: '%gcc'
+
+Spack requires ``cmake`` to use ``gcc`` and all other nodes (including cmake dependencies)
+to use ``clang``. More in general, if there are specific package requirements, the default
+requirements under ``all`` are disregarded.
```
scheibelp: I might reorganize/reword 

```
(new sentence) Note that in this case ``all`` represents a 
*default set of requirements*, so with a configuration like this:

 .. code-block:: yaml

    packages:
      all:
        require: '%clang'
      cmake:
        require: '%gcc'

 Spack requires ``cmake`` to use ``gcc`` and all other nodes (including cmake dependencies)
 to use ``clang``. More in general, if there are specific package requirements, the default
 requirements under ``all`` are disregarded.
```

to 

```
(new paragraph)

Note that in this case ``all`` represents a *default set of requirements* - 
if there are specific package requirements, then the default requirements 
under ``all`` are disregarded. For example, with a configuration like this:

 .. code-block:: yaml

    packages:
      all:
        require: '%clang'
      cmake:
        require: '%gcc'

 Spack requires ``cmake`` to use ``gcc`` and all other nodes (including cmake dependencies)
 to use ``clang``.
```

lib/spack/spack/test/concretize_requirements.py
```diff
@@ -297,3 +297,25 @@ def test_requirements_are_higher_priority_than_deprecation(concretize_scope, tes
...
+
+
+@pytest.mark.parametrize("spec_str,requirement_str", [("x", "%gcc"), ("x", "%clang")])
+def test_default_requirements_with_all(spec_str, requirement_str, concretize_scope, test_repo):
+    """Test that users can override a deprecated version with a requirement."""
+    if spack.config.get("config:concretizer") == "original":
+        pytest.skip("Original concretizer does not support configuration" " requirements")
+
+    # @2.3 is a deprecated versions. Ensure that any_of picks both constraints,
+    # since they are possible
```
scheibelp: I don't think this comment applies here.

Also it might be worth testing the combination of `all:` and specific package constraints

scheibelp (changes requested):
Overall this looks great.

I have a couple requests about comments/docs and one about adding a test.

lib/spack/docs/build_settings.rst
alalazo: Done in 17da48d

lib/spack/spack/test/concretize_requirements.py
alalazo: Done in db7dab1

alalazo:
@scheibelp Should be ready for another review

lib/spack/spack/test/concretize_requirements.py
```diff
@@ -297,3 +297,43 @@ def test_requirements_are_higher_priority_than_deprecation(concretize_scope, tes
     s1 = Spec("y").concretized()
     assert s1.satisfies("@2.3")
     assert s1.satisfies("%gcc")
+
+
+@pytest.mark.parametrize("spec_str,requirement_str", [("x", "%gcc"), ("x", "%clang")])
+def test_default_requirements_with_all(spec_str, requirement_str, concretize_scope, test_repo):
+    """Test that users can override a deprecated version with a requirement."""
```
scheibelp: The docstring for this and the next test need to be updated, in this case:

`Test that default requirements are applied to all packages`

and for the next one

`Test that specific package requirements override default package requirements`

also, it might be worth enforcing with `test_default_and_package_specific_requirements` that all default requirements are ignored for example if you have

```
require:
  all:
    require: %gcc ~shared
  y:
    require: "2.4"
```

then `y` will concretize to `+shared` (since that is the default for `y`).

scheibelp (changes requested):
The docstrings for the new tests should be updated.

I also have a suggestion for how to extend `test_default_and_package_specific_requirements`, but that is not as important. Once those are in, consider this approved (or I'll keep an eye on it and approve it when I see the new changes).

lib/spack/spack/test/concretize_requirements.py
alalazo: Should be done in 199f491d5ef36439d1c528f996dc367dc9046e85

scheibelp (approved):

adamjstewart:
This isn't working as expected for me. For example, I have the following `packages.yaml`:
```yaml
packages:
  all:
    variants: ~mpi
```
I would like to make this a hard requirement such that `+mpi` is disabled _for all packages with an mpi variant_. I tried:
```yaml
packages:
  all:
    require: ~mpi
```
but I see:
```console
$ spack spec zlib
==> Error: variant "mpi" not found in package "zlib" [required from package "zlib"]
```

alalazo:
@adamjstewart There is a bug during the setup phase, but note that requirements under `all` are actually added everywhere - not everywhere where it exists. 

To achieve what you'd like, once the bug is fixed, you can exploit the ordering preference among requirement list and try require:
```yaml
packages:
  all:
    require: 
    # Assume you want to compile everything with GCC, clingo 
    # will try to satisfy `~mpi` first and '%gcc' later
    - any_of: ['~mpi', '%gcc']
```

adamjstewart:
Hmm, that's problematic. This means that `packages:all:require` is practically useless for anything other than compilers. Can I request this feature? The `any_of` option isn't useful if I want to enforce both `~mpi` AND `%gcc` on all packages.

alalazo:
> The any_of option isn't useful if I want to enforce both ~mpi AND %gcc on all packages.

`any_of` allows to match more than one option, so you should be able to do what you say here.

adamjstewart:
Okay, that may work, still feels a bit hacky though. Seems more like a workaround than a well-designed feature. Is there any reason that "require" can't change to "require if possible"? Or is that moving backwards to our old "suggest" style?

alalazo:
> Seems more like a workaround than a well-designed feature. 

Why do you say so? `require` has been designed to be as expressive as we could imagine, that's why it allows for lists of `any_of` and `one_of` groups as base conditions. Do you see anything that you can't do with combination of them, considering that `require` also works for virtual dependencies and for specific packages?[^1] 

[^1]: That doesn't mean that we cannot add syntactic sugar in the future for configuration, but wondering why do you say it's not well designed? 

> Is there any reason that "require" can't change to "require if possible"?

Right now we have clear semantics for `require`, and it works like it would on command line. When you add *if possible* you have first to give meaning to it, which in my opinion involves arbitrary decisions on our side that are best left to users.

If package A depends on B and `conflicts("b+mpi", when="+mpi")` then `+mpi` *is possible* for just one of the two. Whether we should prefer one solution or the other, or error out, depends on our definition of *is possible*. With the current design you can force whatever you want, resorting to `require` under specific packages when requirements under virtual or `all` are problematic.

adamjstewart:
> Why do you say so?

To me, a well designed feature is one that not only works, but is also intuitive. I never would have came up with `any_of: ['~mpi', '%gcc']` in a million years.

Also, it's not particularly robust. Let's say I use `any_of: ['~mpi', '%gcc']`. If I build `spack install hdf5 %gcc`, I get what I want. But if I also want to build `spack install hdf5 %clang`, wouldn't it build `hdf5~mpi%clang ^zlib%gcc`? I would like to avoid modifying any part of the spec other than a potential MPI variant.

I wonder if I could work around the above issue with `any_of: ['~mpi', '@:']`. Again, super unintuitive and ugly syntax.

alalazo:
You don't have only `all:` to use. Think for instance:
```yaml
packages:
  hdf5:
    require: "~mpi"
```
This seems pretty intuitive to me.

adamjstewart:
Yes, extremely intuitive, unfortunately there are 273 packages with an MPI variant.

Basically, the following worked great:
```yaml
packages:
  all:
    variants: ~mpi
```
but there seems to be no simple and intuitive alternative in the new `require:` scheme. I know we talked about deprecating and removing the old scheme, just want to make sure there will still be an easy way to get the same behavior.

alalazo:
> I know we talked about deprecating and removing the old scheme, just want to make sure there will still be an easy way to get the same behavior.

The only deprecation that is likely to happen going to November is that of https://github.com/spack/spack/pull/31261, which doesn't affect variant. *Likely* means that it seemed a good idea in June, and we decided to revisit the PR after we had requirements in place.

scheibelp:
@adamjstewart 

* `require_if_possible` is the same thing as a preference, with the single exception that currently preferences are prioritized lower than reuse; in that sense it would be easy to implement (and `spack spec --fresh` would get that too)
* You mention there are many packages with `+mpi`: generally my intuition for `requires` is that it is ideally used for packages you interact with "directly":
  * If you know that you need to build a package with `~mpi`, then you set requires for it specifically
  * If dependencies build with `+mpi`, then I don't immediately see why that would be a problem (if `x` depends on `y` and `x~mpi` needs `y~mpi`, then that should be encoded in the package.py for `x`)

Regarding https://github.com/spack/spack/pull/32260#issuecomment-1236191787, could you mention more about your use case? e.g. are you encountering some scenario where you set `requires: ~mpi` for a few packages and find that dependencies then build with `+mpi` and the result is incompatible?

adamjstewart:
MPI in particular has a long history of hard-coded variant forwarding and hacks, maybe a simpler example would be CUDA. I have two systems: a laptop without a GPU, and a cluster with a GPU. On the laptop, I would like to use:
```yaml
packages:
  all:
    require: "~cuda"
```
because there are no CUDA drivers installed and the build would fail. On the cluster, I would like to use:
```yaml
packages:
  all:
    require: "+cuda cuda_arch=80"
```
to force all CUDA-enabled packages to be built with CUDA support and explicitly tell them which CUDA arch to build for. Yes, I could gather a list of all packages in Spack with a `+cuda` variant and copy-n-paste this logic for all of those, but what's the point of `all:` if not to avoid this? Spack likely makes the correct decision 99% of the time, but the reason we added `require:` was because of that 1% of the time where it decides to ignore user preferences. I would like to avoid that 1% of the time here as well.

PR #32000, opened 2022-08-08
celeritas: update checksum for git describe difference

sethrj:
Using `$Format:%h$` and the export-subst gitattribute property mean that
the release checksum can change post release. I verified this by diffing fresh spack stage directories for the original (spack-cached) version and a freshly downloaded version.

```
$ diff -ru spack-stage-celeritas-0.1.0-hv7ewpjouekqws2y5iaql2cnp6tn76iz/ spack-stage-celeritas-0.1.0.wtf-ynyck3df2a7kkgkqrpwapgz3l2i62omz/
diff -ru spack-stage-celeritas-0.1.0-hv7ewpjouekqws2y5iaql2cnp6tn76iz/spack-src/cmake/CgvFindVersion.cmake spack-stage-celeritas-0.1.0.wtf-ynyck3df2a7kkgkqrpwapgz3l2i62omz/spack-src/cmake/CgvFindVersion.cmake
--- spack-stage-celeritas-0.1.0-hv7ewpjouekqws2y5iaql2cnp6tn76iz/spack-src/cmake/CgvFindVersion.cmake	2022-07-31 19:45:03.000000000 -0400
+++ spack-stage-celeritas-0.1.0.wtf-ynyck3df2a7kkgkqrpwapgz3l2i62omz/spack-src/cmake/CgvFindVersion.cmake	2022-07-31 19:45:03.000000000 -0400
@@ -75,7 +75,7 @@

   # Get a possible Git version generated using git-archive (see the
   # .gitattributes file)
-  set(_ARCHIVE_TAG "HEAD -> master, tag: v0.1.0")
+  set(_ARCHIVE_TAG "tag: v0.1.0")
   set(_ARCHIVE_HASH "04fe945d9")

   set(_TAG_REGEX "v([0-9.]+)(-dev[0-9.]+)?")
Binary files spack-stage-celeritas-0.1.0-hv7ewpjouekqws2y5iaql2cnp6tn76iz/v0.1.0.tar.gz and spack-stage-celeritas-0.1.0.wtf-ynyck3df2a7kkgkqrpwapgz3l2i62omz/v0.1.0.tar.gz differ
```

PR #32265, opened 2022-08-19
ForTrilinos: new versions 2.0.1, 2.1.0

sethrj:
I also had to update the checksum for the released 2.0.0: see #32000 for the explanation and solution to keep this from happening again.


tldahlgren (approved):
Confirmed new sha256.

PR #32268, opened 2022-08-19
Added entry to support rocky8

fspiga:
Added entry to support rocky8 OS explicitly (which is better!)

This committ superseed #31942.

### DIFF
--- a/var/spack/repos/builtin/packages/arm/package.py
+++ b/var/spack/repos/builtin/packages/arm/package.py
@@ -150,5 +150,3 @@ class Arm(Package):
         env.prepend_path("CPATH", join_path(arm_dir, "include"))
         env.prepend_path("MANPATH", join_path(arm_dir, "share", "man"))
         env.prepend_path("ARM_LICENSE_DIR", join_path(self.prefix, "licences"))
-        if "ncurses" in self.spec:
-            env.prepend_path("LD_LIBRARY_PATH", join_path(self.spec["ncurses"].prefix, "lib"))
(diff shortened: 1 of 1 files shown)


tldahlgren (commented):
Removing the `ncurses` variant means the statement that checks and updates the `LD_LIBRARY_PATH` at line 153-154 are no longer relevant.

tldahlgren (changes requested):
Highlighting need to make `ncurses` dependency and library path changes consistent.

OliverPerks (approved):
Ah ok, if this is sufficient then that is a much easier way of doing things.

tldahlgren:
@fspiga @OliverPerks Apologies.  I had variants on my mind at the time I associated the dependency with the library path change.  

How would `ncurses` appear as part of the spec for the package?  Is there an expectation it would be provided as an explicit dependency when needed?

fspiga:
By making sure the right package for the target OS is used (or the right proxy if Arm Ltd does not provide a native binary built), the `ncurses` dependency is no longer needed. My first fix worked as well but it was not generic enough and it could introduce an unnecessary dependency carried forward on packaged built with Arm compiler using spack.

The current default is set for RHEL7. Maybe too old (?) for many OS that are originated/forked from RHEL/CentOS. Maybe worth bumping to RHEL8? @OliverPerks

tldahlgren:
> By making sure the right package for the target OS is used (or the right proxy if Arm Ltd does not provide a native binary built), the `ncurses` dependency is no longer needed. My first fix worked as well but it was not generic enough and it could introduce an unnecessary dependency carried forward on packaged built with Arm compiler using spack.
> 
> The current default is set for RHEL7. Maybe too old (?) for many OS that are originated/forked from RHEL/CentOS. Maybe worth bumping to RHEL8? @OliverPerks

Just so I'm clear.  Does NOT explicitly adding `ncurses` to the `LD_LIBRARY_PATH` work properly?  If so, I'll approve and merge this PR.

alalazo:
The package is now called `armpl-gcc` and supports rocky8, so this was incorporated at some point.

PR #23661, opened 2021-05-15
performance: speed up existence checks in packages

tgamblin:
Fixes #19871.

Spack doesn't require users to manually index their repos; it reindexes the indexes automatically when things change. To determine when to do this, it has to `stat()` all package files in each repository to make sure that indexes up to date with packages. We currently index virtual providers, patches by sha256, and tags on packages.

When this was originally implemented, we ran the checker all the time, at startup, but that was slow (see #7587). But we didn't go far enough -- it still consults the checker and does all the stat operations just to see if a package exists (`Repo.exists()`).  That might've been a wash in 2018, but as the number of packages has grown, it's gotten slower -- checking 5k packages is expensive and users see this for small operations.  It's a win now to make `Repo.exists()` check files directly.

**Fix:**

This PR does a number of things to speed up `spack load`, `spack info`, and other commands:

- [x] Make `Repo.exists()` check files directly again with `os.path.exists()` (this is the big one)
- [x] Refactor `Spec.satisfies()` so that a checking for virtual packages only happens if needed
      (avoids some calls to exists())
- [x] Avoid calling `Repo.exists(spec)` in `Repo.get()`. `Repo.get()` will ultimately try to load
      a `package.py` file anyway; we can let the failure to load it indicate that the package doesn't
      exist, and avoid another call to exists().
- [x] Fix up some comments in spec parsing
- [x] Call `UnknownPackageError` more consistently in `repo.py`

I am not sure yet whether this last change will make some operations slower.  Comments would be appreciated if it does.

Issue #11160, opened 2019-04-10
Issues with recursive spec syntax in depends_on


alalazo:
The second test in #11158: 

https://github.com/spack/spack/blob/45dfbea3ec015ccc47f6272040898a8810cadf81/lib/spack/spack/test/concretize.py#L537-L542

didn't pass without the hack in:

https://github.com/spack/spack/blob/45dfbea3ec015ccc47f6272040898a8810cadf81/lib/spack/spack/concretize.py#L497-L502

used to split a recursive spec into its components.

### Steps to reproduce the issue
Checkout #11158, use `split_specs = abstract_specs` in the snippet above and run:
```console
$ spack test -k test_simultaneous_concretization_of_specs
```

### Additional information
It seems that recursive constraints, e.g the version of `dyninst` in:
```python
depends_on('callpath@0.9 ^dyninst@8.1.1')
```
won't be enforced unless the dependencies are splitted as:
```python
depends_on('callpath@0.9')
depends_on('dyninst@8.1.1')
```

PR #9612, opened 2018-10-23
Spack Environments (part 4): command-line, spack.yaml, and spack.lock

tgamblin:
This supersedes #8231, reworks the API, and adds a lot of features.

Spack environments are fill a number of needs:

1. Allow users to easily work with distinct sets of packages and configurations, e.g., for different projects or different types of deployments;
2. Allow users to reproduce a set of packages someone else built, either:
    a. Functionally, from the *abstract* specs of the prior installation (i.e., with an `spack.yaml` file)
    b. Exactly, from the *concrete* specs that were previously installed (i.e., with what `bundler`, `npm`, `cargo`, and `pipenv` call a "lockfile")
3. Allow users to interact with sets of specs easily on the command line;
4. Allow users to easily version dependency and configuration information, e.g., in a git repository; and
5. Provide an easier alternative to individually loading modules for packages with "virtual environments".

I've attempted to refactor the code so that one basic `Environment` concept can provide all of these. This builds on the prior work by @scheibelp and @citibeth, and it attempts to integrate some of what is described at #7944.

I think it is easiest to describe what's implemented here in terms of two workflows:

## Command-line environment usage

You can use enviroments to work with a subset of packages, like you'd normally work with spack:

```console
$ spack env create myenv         # create an environment
$ spack env list                 # what (named) environments are available?
==> 3 environments
    foo  my-env  my-project
$ spack env activate -p myenv    # "activate" it (-p optionally sets the prompt)
[myenv] $ spack find             # show packages that are in this environment
==> 0 installed packages.
[myenv] $ spack install libjpeg
[myenv] $ spack find             # now one's installed
==> 1 installed package.
-- darwin-highsierra-x86_64 / clang@9.0.0-apple -----------------
libjpeg@9c
[myenv] $ despacktivate          # deactivate (spack env deactivate also works :)
$
```

You can incrementally add things to environments:

```console
$ spack env create myenv
$ spack env activate -p myenv
[myenv] $ spack env add openmpi  # add some packages
[myenv] $ spack env add hdf5
[myenv] $ spack env add libelf
[myenv] $ spack env status       # what've I done so far?
==> In environment myenv
added:
---- openmpi
---- hdf5
---- libelf
[myenv] $ spack env install      # install the whole thing
...
[myenv] $ spack env status
$ spack env status
==> In environment my-env
concrete:
---- libelf
[+]  libelf@0.8.13%clang@9.0.0-apple arch=darwin-highsierra-x86_64
---- openmpi
[...]

PR #32273, opened 2022-08-19
Refactor MockPackage into MockRepositoryBuilder

alalazo:
This PR factors together code to generate a mock repository with fake packages, which was present in multiple places in Spack. It also construct indexes injecting a reference to the repository they refer to, removing any direct or indirect reference to global state in the class.

`Spec.virtual` now use the index, and computation done to compute the index use `Repository.is_virtual_safe`.

The approach taken is to provide an API to generate a mock repository in a temporary directory, instead of having fake objects mimicking a `RepoPath` and a `Package`. One advantage, besides the deduplication of the code, is that there will be no maintenance required to adapt the mocks when the API of `RepoPath` and `Package` change, since we'll be using the actual classes.

The decorator `spack.repo.use_repositories` now has an additional argument which permits to override (or not) existing repositories. It also pushes a corresponding configuration scope into the global config, so to keep it consistent with the changes.

Modifications:
- [x] `*Index` objects get the repository they refer to during construction
- [x] `Spec.virtual` uses the index
- [x] Remove `mock_package.py`
- [x] Add `MockRepositoryBuilder` in `spack.repo`
- [x] Add debug print for pushing and popping config scopes
- [x] Changed `spack.repo.use_repositories` so that it can override or not previous repos
- [x] `spack.repo.use_repositories` updates `spack.config.config` according to the modifications done
- [x] `spack -m` is now reflected in the configuration
- [x] `spack.concretize.concretize_specs_together` has been simplified by using `MockRepoBuilder`
- [x] Removed a peculiar behavior from `spack.config.Configuration` where `push` would always bubble-up a scope named `command_line` if it existed


lib/spack/spack/config.py
```diff
@@ -408,28 +408,22 @@ def __init__(self, *scopes):
     @_config_mutator
     def push_scope(self, scope):
         """Add a higher precedence scope to the Configuration."""
-        cmd_line_scope = None
-        if self.scopes:
-            highest_precedence_scope = list(self.scopes.values())[-1]
-            if highest_precedence_scope.name == "command_line":
-                # If the command-line scope is present, it should always
-                # be the scope of highest precedence
-                cmd_line_scope = self.pop_scope()
```
alalazo, line 418: @tgamblin @scheibelp This was added in 4b2f51d063111d0966404ff7f1eea627f2d286d8 which is part of #9612 There's no unit test failing on removing this behavior (in fact I have tests failing here due to this behavior + the refactor that keeps some of the context manager in sync with the configuration). 

Was there any reason why this was added? Asking because if there is, we need to add an additional test and restore the behavior _outside_ of this class, i.e. in the calling code:
```python
config.push_scope(...)
config.move_as_top_scope("command_line")
```

alalazo:
I have been timing a few solves on this branch:
radiuss_index_hist

and on `develop`:

radiuss_develop_hist

Using the index in `Spec.virtual` seem to have some effect in the setup phase. Not a game changer, but still an improvement.

alalazo:
@tgamblin @trws Would either of you have time to take a look at this refactoring? It started trying to inject repositories into the indexes that are used as caches at construction time, and ended up doing a few more things always related to repos. There's a comprehensive list in the description.

lib/spack/spack/config.py
scheibelp: I want the user to be able to use command-line scopes to have the "last-say": I think they will be frustrated if other configuration implicitly overrides config that they set with `-C`. If I failed to test this, then that is my mistake. IMO the behavior is still worthwhile (and worth testing).

share/spack/templates/mock-repository/package.pyt
```diff
@@ -0,0 +1,19 @@
...
+{% if dep_type and condition %}
+    depends_on("{{ dep_spec }}", type="{{ dep_type }}", when="{{ condition }}")
+{% elif dep_type %}
+    depends_on("{{ dep_spec }}", type="{{ dep_type }}")
+{% elif condition %}
+    depends_on("{{ dep_spec }}", when="{{ condition }}")
+{% else %}
+    depends_on("{{ dep_spec }}")
+{% endif %}
+{% endfor %}
```
scheibelp, line 19: Would the mechanism from https://github.com/spack/spack/pull/27987 also be a suitable replacement? i.e. that PR adds the ability to define packages exactly and in-line with the testing code:

```
_pkgy = (
     "y",
     """\
 class Y(Package):
     version('2.5')
     version('2.4')
     version('2.3', deprecated=True)

     variant('shared', default=True,
             description='Build shared libraries')
 """,
 )
```

I think this direct approach is useful because users know exactly what the package "should be" (e.g. without doing any mental substitution based on the template).

lib/spack/spack/config.py
alalazo: Sure, my only point is that shouldn't be part of the `Configuration.push` behavior but should be checked at the caller site where we assemble the configuration. Rationale for saying that:
1. Having an API where subsequent `push` and `pop` don't restore the initial state is very counter-intuitive.
2. `Configuration` should not bother with the names of the scopes, it's code that uses that class that could/should give meaning to them

lib/spack/spack/repo.py
```diff
@@ -983,20 +994,34 @@ def exists(self, pkg_name):
...
+        This function use the provider index. If calling from a code block that
+        is used to construct the provider index use the ``is_virtual_safe`` function.
+
+        Args:
+            pkg_name (str): name of the package we want to check
+        """
+        have_name = self._have_name(pkg_name)
+        return have_name and pkg_name in self.provider_index
+
+    def is_virtual_safe(self, pkg_name):
```
trws, line 1023: I get wanting to separate these, but doesn't this pessimize specs, especially merge_dependency, such that we can't ever use an index even if we have one?  This method shows up pretty heavily on performance traces because of so frequently calling get_pkg_class through spec, so I've been keeping an eye on it.

lib/spack/spack/spec.py
```diff
@@ -1532,16 +1532,7 @@ def package_class(self):
...
-
-        TODO: revisit this -- might need to use a separate namespace and
-        be more explicit about this.
-        Possible idea: just use conventin and make virtual deps all
-        caps, e.g., MPI vs mpi.
-        """
-        # This method can be called while regenerating the provider index
-        # So we turn off using the index to detect virtuals
-        return spack.repo.path.is_virtual(self.name, use_index=False)
+        return spack.repo.path.is_virtual(self.name)
```
alalazo, line 1551: @trws I didn't look at traces in this PR, but I would expect separating the function into two optimizes times. The timings I posted seem to show a consistent shave of 0,5 sec. on the setup phase across many specs (on my laptop with SSD). 

Note that before we were using `use_index=False` because we called `Spec.virtual` even when reconstructing repo caches. In this PR I took care of using `RepoPath.is_virtual_safe` in those places so I could enable using the cache when calling `Spec.virtual`.

alalazo: To be more explicit, the `use_index` parameter was removed in this PR and old `use_index=False` maps to `is_virtual_safe`, while `use_index=True` maps to `is_virtual`.

trws: Fair, and if we want to go further in spec we could always do something like:

```python
if currently_bootstrapping():
  is_virtual_safe()
else:
  is_virtual()
```

Works for me, thanks!

share/spack/templates/mock-repository/package.pyt
alalazo: I think I like the approach here more, since it is more compact and can be extended further. It also has a single place where these mock packages are collected, so we don't scatter the code around. 

That said, the main point of this change was to get rid of `MockPackageMultiRepo` etc. since that machinery requires maintenance every time we change properties of repositories. Here I remove the mock class and add the possibility to create small repositories to be used during tests on the fly.

scheibelp: > I think I like the approach here more, since it is more compact and can be extended further. 

I'm not sure what is more extensible than being able to write the package contents down to the character: a template is flexible but it will never be that flexible. Also, ultimately demands on the template will encourage for example replacing the hardcoded versions that you have there with a placeholder for versions that you insert: at that point, my concern is that you will have created an interface for creating packages which is analogous to (and redundant with) the package.py format.

> It also has a single place where these mock packages are collected, so we don't scatter the code around.

I'm not sure what you mean here: `test_save_dependency_spec_jsons_subset` for example creates its own repo as part of the test, that seems equally as disperse as the preceding approach.

Additionally, IMO consolidation of mock packages is problematic: I see in many cases that `builtin.mock` packages are used for multiple tests and thereby aggregate unrelated behaviors that makes them hard to understand (e.g. in the process of understanding a unit test, a user tracks down the associated `builtin.mock` package and reads it, and then must parse out the components of that package which happen to relate to the specific test).

alalazo:
@spackbot run pipeline

share/spack/templates/mock-repository/package.pyt
alalazo: > I'm not sure what is more extensible than being able to write the package contents down to the character: a template is flexible but it will never be that flexible.

This is not true, since writing a literal package can't possibly deal with dynamic content like:
https://github.com/spack/spack/blob/97bd867518c860c16acb85cf601fd4aa2c24e91c/lib/spack/spack/concretize.py#L753-L774

where the package to be generated depends on the function arguments. I can, of course, concatenate strings etc. but at that point I would have re-created a basic template engine. That said my point is:
- Before this PR we were mocking a `Package` without having a `package.py` available and that was brittle[^1]
- In #27987 we introduced a mechanism to use a literal `package.py` in tests. This is fine with me, and I don't want to change that in this PR.
- Here I introduce a class that can create a mock repository with mock `package.py` from API. That is flexible enough to be used in many places where we had similar code before (hence the saving of roughly 200 lines)

> Additionally, IMO consolidation of mock packages is problematic:

To be clear, I don't plan to consolidate `builtin.mock` but only the tests that are using `MockPackageMultiRepo` and related classes

[^1]: Changing a few signatures caused tests relying on the mock to fail, and solving that was not straightforward.

lib/spack/spack/concretize.py
```diff
@@ -752,37 +751,20 @@ def _concretize_specs_together_new(*abstract_specs, **kwargs):
...
-
-        return spack.repo.Repo(repo_path)
-
     abstract_specs = [spack.spec.Spec(s) for s in abstract_specs]
-    concretization_repository = make_concretization_repository(abstract_specs)
+    tmpdir = tempfile.mkdtemp()
+    builder = spack.repo.MockRepositoryBuilder(tmpdir)
+    # Split recursive specs, as it seems the concretizer has issue
+    # respecting conditions on dependents expressed like
+    # depends_on('foo ^bar@1.0'), see issue #11160
```
tgamblin, line 759: Since #11160 is closed, is this still needed?

lib/spack/spack/patch.py
```diff
@@ -329,17 +330,20 @@ class PatchCache(object):
...
-    def __init__(self, data=None):
+    def __init__(self, data=None, repository=None):
         if data is None:
             self.index = {}
         else:
             if "patches" not in data:
                 raise IndexError("invalid patch index; try `spack clean -m`")
             self.index = data["patches"]
 
+        assert repository is not None, "a 'repository=' argument is required"
```
tgamblin: why not just remove the default value?

lib/spack/spack/provider_index.py
```diff
@@ -146,14 +146,16 @@ def __init__(self, specs=None, restrict=False):
         if specs is None:
             specs = []
 
+        assert repository is not None, "a 'repository=' argument is required"
```
tgamblin: same here -- why not just omit the default value?

lib/spack/spack/tag.py
```diff
@@ -112,14 +113,15 @@ def merge(self, other):
...
-    def update_package(self, pkg_name):
+    def update_package(self, pkg_name, repository=None):
         """Updates a package in the tag index.
 
         Args:
             pkg_name (str): name of the package to be removed from the index
 
         """
-        pkg_cls = spack.repo.path.get_pkg_class(pkg_name)
+        repository = repository or spack.repo.path
```
tgamblin: Why can't this (and all other instance methods in the indexers) use `self.repository`?  It seems way cleaner to only pass that in the constructor than to require it on every method.

tgamblin (changes requested):
This looks much more readable and it's definitely an improvement.  I would really like to get to inline package classes, e.g.:

```python
class MyRepo(MockRepo):
    class A(Package):
        version("1.0", sha256="abcdefabcdefabcdef")
        depends_on("b")

    class B(Package):
        version("1.0", sha256="abcdefabcdefabcdef")

```

But I think moving to this first and trying that later makes sense.

I have a few change requests, and I'm not sure why the repository has to be passed to *all* the instance methods of the indexers and not just the constructor.  Otherwise this LGTM.

lib/spack/spack/concretize.py
alalazo: Yes, the old concretizer cannot handle things like:
```python
depends_on("foo ^bar")
```
so here we split into:
```python
depends_on("foo")
depends_on("bar")
```
The comments point to the original issue solved by the split.

lib/spack/spack/patch.py
alalazo: The idea was to avoid changing the signature from:
```python
def __init__(self, data=None):
```
to:
```python
def __init__(self, repository, data=None):
```
but I can do that if it is preferable.

lib/spack/spack/provider_index.py
alalazo: Same as before, but if preferred I can move `repository` to the front in the signature

lib/spack/spack/tag.py
alalazo: This is a leftover, thanks for catching it. Did you see any other on instance methods?

alalazo: Done in 46e0101

lib/spack/spack/provider_index.py
alalazo: Done in 9916991

lib/spack/spack/patch.py
alalazo: Done in 4a115051

alalazo:
@tgamblin This should be ready for a second review

lib/spack/spack/spec.py
tgamblin: I have a last question on this one: is this going to regress https://github.com/spack/spack/pull/23661?  The point of that PR was to avoid the massive stat we do on all packages for simple operations like `spack load`, where it is faster just to check each package individually.  I wonder if doing this *looks* fast here but also slows down operations like that.  What we want to avoid is the index check for simple operations where we don't *have* to reindex all virtuals.

@alalazo can you check on this and look at the issue that it fixes?

tgamblin (changes requested):
I have one final question about this PR before it goes in -- we really need to be careful about using the virtual index in the wrong places because it can cause unexpected performance problems.

alalazo:
@tgamblin I don't see differences with `develop` on a test run on Marconi, so I think this is safe to merge. In case I am wrong, and we get user issues about being slow on GPFS, the workaround is to change a single line and use:
```python
@property
def virtual(self):
    return spack.repo.path.is_virtual_safe(self.name)
```
in `spack.spec.Spec.virtual`, so it's very fast to act on this possible request.

---

The test performed on Marconi (CINECA). With this PR:
```console
[mculpo02@login02 spack]$ spack clean -a
==> Removing all temporary build stages
==> Removing cached downloads
==> Removing install failure marks
==> Removing cached information on repositories
==> Removing python cache files
$ time spack load hdf5

real	1m21,228s
user	0m56,768s
sys	0m6,822s
$ time spack unload hdf5

real	0m2,165s
user	0m1,344s
sys	0m0,825s
$ time spack load hdf5

real	0m2,498s
user	0m1,775s
sys	0m0,728s
$ time spack load hdf5

real	0m2,427s
user	0m1,577s
sys	0m0,858s
```
while on `develop`:
```console
$ git checkout develop
Switched to branch 'develop'
Your branch is up to date with 'upstream/develop'.
$ spack clean -a
==> Removing all temporary build stages
==> Removing cached downloads
==> Removing install failure marks
==> Removing cached information on repositories
==> Removing python cache files
$ time spack load hdf5

real	1m8,999s
user	0m57,426s
sys	0m6,980s
$ time spack unload hdf5

real	0m2,405s
user	0m1,524s
sys	0m0,883s
$ time spack load hdf5

real	0m2,740s
user	0m1,897s
sys	0m0,851s
$ time spack load hdf5

real	0m2,753s
user	0m1,866s
sys	0m0,896s
```

tgamblin (approved):

haampie:
This regressed overriding config from active environments:

```
spack env activate --temp
spack config add config:ccache:true
spack -c config:ccache:false config get config | grep ccache  # gives true instead of false
```

alalazo:
I guess it's:

> Removed a peculiar behavior from spack.config.Configuration where push would always bubble-up a scope named command_line if it existed

? I'll try to submit a fix quickly.

PR #32275, opened 2022-08-19
netcdf-cxx: Update the download URL

RemiLacroix-IDRIS:


tldahlgren (approved):
Confirmed sha256 stays the same with the new URL.

PR #32276, opened 2022-08-19
Doc: WarpX SENSEI Work-Around

ax3l:
Follow-up to #31542


kwryankrattiger (approved):
Thanks for finding this!

PR #32282, opened 2022-08-19
xgboost: GCC 8+ required

adamjstewart:
Tried building this with CUDA support in CI but it doesn't seem to work, at least with GCC 7: https://gitlab.spack.io/spack/spack/-/jobs/3061107


tldahlgren (approved):
LGTM

PR #32297, opened 2022-08-21
tensorflow-lite: new package

wdconinc:
TensorFlow Lite is TensorFlow's lightweight solution
for mobile and embedded devices. It enables low-latency
inference of on-device machine learning models with a
small binary size and fast performance supporting
hardware acceleration.

This package aims to support workflows on HPC and HTC
systems that use inference in highly size-optimized
contexts, such as in distributed event processing in
high energy physics data pipelines.

@adamjstewart Since this uses the same TensorFlow source archive, there may be better ways to approach this, maybe even as part of the `tensorflow` package. This just aims for a much smaller install size (~240 MB) for use in container that are pushed out to compute nodes. Interested in your thoughts.

### DIFF
--- a/var/spack/repos/builtin/packages/tensorflow-lite/package.py
+++ b/var/spack/repos/builtin/packages/tensorflow-lite/package.py
@@ -78,18 +78,18 @@ class TensorflowLite(CMakePackage):
         # Install library
         mkdirp(self.prefix.lib)
         with working_dir(self.build_directory):
-            for l in find(".", "libtensorflow-lite.*", recursive=False):
-                install(l, self.prefix.lib)
+            for lib in find(".", "libtensorflow-lite.*", recursive=False):
+                install(lib, self.prefix.lib)
 
         # Install headers for tensorflow itself
         mkdirp(self.prefix.include)
-        for h in find(join_path(self.stage.source_path, "tensorflow/lite"), "*.h", recursive=True):
-            relpath = os.path.relpath(h)
+        for header in find(join_path(self.stage.source_path, "tensorflow/lite"), "*.h", recursive=True):
+            relpath = os.path.relpath(header)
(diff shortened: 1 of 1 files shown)


adamjstewart (commented):
This looks great! I do agree with making it a separate package. As for the name, I think we should go with `py-tflite-runtime` to match the [PyPI package name](https://pypi.org/project/tflite-runtime). 

@aweits and @glennpj may be interested in this as well.

wdconinc:
> As for the name, I think we should go with `py-tflite-runtime`

I'm all for using standard naming, but I had to look into this first and I don't think that `tflite-runtime` (i.e. `import tflite_runtime`) is necessarily very similar (though `tflite` is definitely how people refer to TensorFlow Lite, so I think using that name is an entirely good suggestion). This package in particular installs a static library and associated C++ headers. Is `tflite-runtime` a python layer to use this same library? I'm not sure.

adamjstewart:
Hmm, I guess I'm not familiar with the C++ interface, only the Python interface. When I Google "install tensorflow-lite" the first hit gives me installation for the Python interface: 

* https://www.tensorflow.org/lite/guide/python

I also see instructions for building from source:

* https://www.tensorflow.org/lite/guide/build_cmake
* https://www.tensorflow.org/lite/guide/build_cmake_pip

How is this package intended to be used? If we want a single package for both C++ and Python then it's probably best to use the name on PyPI. We could also have separate packages for C++ and Python, although it depends on whether or not it's possible to build them separately.

wdconinc:
We're linking against this library and using it for inference in C++ based frameworks, using models that have been trained with a full TensorFlow installation and exported to the tflite file format.

bernhardkaindl:
@wdconinc Is this still a draft or soon to be ready for merge?

PR #31285, opened 2022-06-24
intel oneapi classic bootstrapping

becker33:
Closes #31278.

The `intel` compiler at versions > 20 is provided by the `intel-oneapi-compilers-classic` package (a thin wrapper around the `intel-oneapi-compilers` package), and the `oneapi` compiler is provided by the `intel-oneapi-compilers` package. 

Prior to this work, neither of these compilers could be bootstrapped by Spack as part of an install with `install_missing_compilers: True`.

Changes made to make these two packages bootstrappable:

1. The `intel-oneapi-compilers-classic` package includes a bin directory and symlinks to the compiler executables, not just logical pointers in Spack.
2. Spack can look for bootstrapped compilers in directories other than `$prefix/bin`, defined on a per-package basis
3. `intel-oneapi-compilers` specifies a non-default search directory for the compiler executables.
4. The `spack.compilers` module now can make more advanced associations between packages and compilers, not just simple name translations
5. Spack support for lmod hierarchies accounts for differences between package names and the associated compiler names for `intel-oneapi-compilers/oneapi`, `intel-oneapi-compilers-classic/intel@20:`, `llvm+clang/clang`, and `llvm-amdgpu/rocmcc`.

@rscohn2 do you see any issue with the particular choice of symlinks here?

@nicholas-sly @lee218llnl this is the PR for the work we've been discussing for intel compilers.

TODO:
- [x] full end-to-end testing
- [x] add unit tests

PR #32299, opened 2022-08-21
Fix intel-oneapi-compilers-classic bin path

OfirGan:


rscohn2:
Thanks! Apparently I did not do any testing for this.

rscohn2 (approved):

rscohn2:
@OfirGan: thanks for the change. I fixed the style issues in #32305, so closing this one. If you do:
```
spack style
```
It will run the checks locally so it is quicker to fix problems.

OfirGan:
there is another problem white the "MODULEPATH",
as for version 2022.1.0 the compiler version is "intel\2021.6.0" so when you load the module it doesn't automatically use the 
the correct path to the compiler modules.
i dont know how to fix it

rscohn2:
I think https://github.com/spack/spack/pull/31285 will fix it.

PR #32307, opened 2022-08-22
py-pynn: add version 0.10.0

AvidusMauch:


var/spack/repos/builtin/packages/py-pynn/package.py
```diff
@@ -15,20 +15,30 @@ class PyPynn(PythonPackage):
...
     version("0.8.1", sha256="ce94246284588414d1570c1d5d697805f781384e771816727c830b01ee30fe39")
     version("0.7.5", sha256="15f75f422f3b71c6129ecef23f29d8baeb3ed6502e7a321b8a2596c78ef7e03c")
 
     depends_on("python@2.6:2.8,3.3:", type=("build", "run"))
+    depends_on("python@3.7:", type=("build", "run"), when="@0.10.0:")
+
     # pip silently replaces distutils with setuptools
     depends_on("py-setuptools", type="build")
     depends_on("py-jinja2@2.7:", type=("build", "run"))
     depends_on("py-docutils@0.10:", type=("build", "run"))
```
adamjstewart: jinja2 and docutils are no longer needed in the latest release

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pynn/package.py
AvidusMauch: sorry, for my understanding, spack itself handles those two dependencies or what do you mean?

adamjstewart: I mean py-pynn 0.9.1 depends on these, but py-pynn 0.10.0 does not. So you should add `when="@:0.9"` to them.

AvidusMauch: they are at least still part of the requirements file https://github.com/NeuralEnsemble/PyNN/blob/0.10.0/requirements.txt
where do you get the information that they are not required anymore? I also don't see anything in the release notes

adamjstewart: The requirements file isn't used by pip/build/any package managers and is more or less meaningless. See https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#dependencies for a list of locations to check for dependencies. In this case, `setup.py` is the file that lists the dependencies of the project, all other files are ignored.

AvidusMauch: ok, thanks for the input and the review

adamjstewart (approved):

PR #28504, opened 2022-01-19
Switch to full hash everywhere

tgamblin:
fixes #21491.
fixes #23428.
fixes #30310.

For a long time, Spack has used a coarser hash to identify packages than it likely should. Packages are identified by `dag_hash()`, which includes only link and run dependencies. Build dependencies are stripped before hashing, and we have notincluded hashes of build artifacts or the `package.py` files used to build.  This means the DAG hash actually doesn't represent all the things Spack can build, and it reduces reproducibility.

We did this because, in the early days, users were (rightly) annoyed when a new version of CMake, autotools, or some other build dependency would necessitate a rebuild of their entire stack. Coarsening the hash avoided this issue and enabled a modicum of stability when only reusing packages by hash match.

Now that we have `--reuse`, we don't need to be so careful. Users can avoid unnecessary rebuilds much more easily, and we can add more provenance to the spec without worrying that frequent hash changes will cause too many rebuilds.

This PR makes the following changes:

- [x] Make `Spec.dag_hash()` include build, run, and link dependencies and the package hash (it is now equivalent to `full_hash()`).
- [x] Only use the old `build_hash()` and `full_hash()` methods where needed.
- [x] Simplify environment lockfiles to only use the single `dag_hash()`
    - [x] Document historical lockfile formats and ensure they can be read
- [x] Make the Spack database preserve build dependencies after installation
- [x] Fix the concretizer to ignore build dependencies of already-installed specs when doing reuse (as they're only relevant in their respective build environments).
- [x] Rework `spack ci` to *not* splice specs on update

There are also some bugfixes for problems discovered during the switch:

- [x] Don't add a `package_hash()` in `to_node_dict()` unless the spec is concrete (fixes breaks on abstract specs)
- [x] Don't add source ids to the package hash for packages without a known fetch strategy (many mock packages are like this).
- [x] Change how `Spec.patches` is memoized. Using `llnl.util.lang.memoized` on `Spec` objects causes specs to be stored in a `dict`, which means they need a hash.  But, `dag_hash()` now includes patch `sha256`'s via the package hash, which can lead to infinite recursion.

PR #28156, opened 2021-12-24
Make the package hash consistent

tgamblin:
This PR ensures that the package hash is consistent across Python versions, and adds a lot of tests to ensure that.  It also makes the package hash less sensitive, to avoid superfluous rebuilds in CI.

Now that we have #25310 (reusing installed packages), we want to make Spack hashes more fine-grained to preserve provenance.  To do that we want to switch to using full hashes everywhere.  Full hashes include all build dependencies, a hash of the canonicalized package file, and the hashes of artifacts and patches used in the build.

The existing full hash worked by editing the AST of a `package.py` file, removing multimethods (where statically possible b/c we know they won't be triggered), and stripping comments, docstrings, and directives.  Comments and docstrings don't affect the build, and directives are represented by other metadata on the spec.  Once all these modifications were done, we dumped the `repr()` of the AST and hashed that.

There were a few problems with the old package hash:
1. The AST changes pretty frequently from Python version to Python version, and so does its `repr()`.  So, Spack running under Python 2.7, 3.5, and 3.8 may calculate entirely different full hashes for the same `package.py` file.  This isn't as big a deal *right now*, but it needs to be fixed if we are going to move to full hashes.
2. Static multimethod reosolution (removing multi-methods if we know they won't be used) was not working in all cases in the previous version of the hash, and there weren't a lot of tests for it.  This resulted in the old package hash stripping *any* function with a decorator from the canonical representation, so we would lose some functions that affected builds.
3. The old docstring remover would only strip the first free-standing string in a module/class/method.  We want to strip all of them (a few packages had extra strings lying around).
4. The old package hash would not remove directive nested in complex logic at the package level (`if`, `for`, `with when()`, etc.), and that constitutes a lot of logic for some packages, even when all the loops are doing is setting up complex directives.  
5. The old package hash was considering changes to `maintainers` and `tags`, and we really don't want those to require a rebuild.  

This PR makes the following modifications:

- [x] To solve the AST issue, add `spack.util.unparse`, which regenerates Python source code from a parsed AST.  This is based on https://github.com/simonpercivall/astunparse, which is consistent across Python 2.7-3.8, with back ports from the new `ast.unparse` function in Python 3.9+.
    * The `unparse()` method implemented here is consistent for Python 2.7 and 3.5-3.10.
    * We added a `py_ver_consistent=True` argument to `unparse()`, which handles the following cases:
[...]

PR #7193, opened 2018-02-06
Features/package hash

mathstuf:
---
This is the `features/package-hash` branch rebased onto `develop` and split into smaller, more logical commits. Currently there are two test failures on the top of the branch, but the rest of the branch is passing the test suite.

Cc: @tgamblin @scheibelp 

See #7119

Issue #30720, opened 2022-05-18
Changes to base class does not change package hash


iarspider:
### Steps to reproduce

```
$ spack create -n cmsroot -t cmake https://github.com/cms-sw/root/
$ cat var/spack/repos/builtin/packages/cmsroot/package.py
from spack.pkg.builtin.root import Root as BuiltinRoot


class Cmsroot(BuiltinRoot):
    __doc__ = BuiltinRoot.__doc__

    git      = "https://github.com/cms-sw/root.git"

    version('1.2.3', '0123456789abcdef0123456789abcdef')

    def cmake_args(self):
        args = []
        replace_args = ('builtin_ftgl', 'builtin_gl2ps', 'builtin_glew')
        for arg in super().cmake_args():
            for rarg in replace_args:
                if arg.startswith('-D' + rarg):
                    args.append(self.define(rarg, True))
                    break
            else:
                args.append(arg)

        return args
$ spack spec --format '{hash}' cmsroot
idkrysrzi72lmyiah6ne2gyswrduzhy5
$ sed -i -e "s/return options/options.append(define('FOO', 'BAR'))\n        return options"
$ spack spec --format '{hash}' cmsroot
idkrysrzi72lmyiah6ne2gyswrduzhy5
```

### Error message

Hash should be different

### Information on your system

* **Spack:** 0.18.0.dev0 (c775c322ecb95828cc5b4ea6cabbb670423a2d5d)
* **Python:** 3.10.4
* **Platform:** linux-fedora36-skylake
* **Concretizer:** clingo

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

iarspider:
A bit of context: in CMS, we modify many recipes to match our setup (e.g. we drop documentation and static libraries to reduce the image size). In order not to copy and patch the upstream recipe every time it updates, we decided to use inheritance. However, it was discovered that if upstream package changes we may miss the changes in our build since the hash of the package will not change.

becker33:
@tgamblin how feasible is it to fix this for the release? I think we can do the static version of this (canonicalize all modules in class hierarchy between package and PackageBase) and not worry about whether those changes would be masked out by an overriding method the inheriting class.

PR #32312, opened 2022-08-22
specs: include source provenance in `spec.json` and package hash

tgamblin:
We've included a package hash in Spack since #7193 for CI, and we started using it on the spec in #28504. However, what goes into the package hash is opaque.

We want this information to be available *from the concrete spec* so that it can be retrieved as provenance *without* referring to a `package.py` file.  Why?

1. We want to be able to look up authoritative checksums, versions, etc. and pass them through scanners long *after* an installation takes place.
2. Right now you can look at the version on a `Spec`, then look up the checksum for a version from the `package.py`, but that's not reliable in the long term.  We may remove versions from `package.py` (in fact, we might want to as they become less relevant), and checksums for particular versions/releases may change over time.
3. Only the version is stored explicitly on the spec; that's not sufficient to know *exactly* what hashes a spec was built with.

Adding hashes explicitly gets around this and gets us closer to having all the information we need for a complete SBOM.

Here's what `spec.json` looked like before:

```json
{
  "spec": {
    "_meta": {
      "version": 3
    },
    "nodes": [
      {
        "name": "zlib",
        "version": "1.2.12",
        ...
        "patches": [
          "0d38234384870bfd34dfcb738a9083952656f0c766a0f5990b1893076b084b76"
        ],
        "package_hash": "pthf7iophdyonixxeed7gyqiksopxeklzzjbxtjrw7nzlkcqleba====",
        "hash": "ke4alug7ypoxp37jb6namwlxssmws4kp"
      }
    ]
  }
}
```

The `package_hash` there is a hash of the concatenation of:

* A canonical hash of the `package.py` recipe, as implemented in #28156;
* `sha256`'s of patches applied to the spec; and
* Archive `sha256` sums of archives or commits/revisions of repos used to build the spec.

There are some issues with this: patches are counted twice in this spec (in `patches` and in the `package_hash`), the hashes of sources used to build are conflated with the `package.py` hash, and we don't actually include resources anywhere.

With this PR, I've expanded the package hash out in the `spec.json` body. Here is the "same" spec with the new fields:

```json
{
  "spec": {
    "_meta": {
      "version": 3
    },
    "nodes": [
      {
        "name": "zlib",
        "version": "1.2.12",
        ...
        "package_hash": "6kkliqdv67ucuvfpfdwaacy5bz6s6en4",
        "sources": [
          {
            "type": "archive",
            "sha256": "91844808532e5ce316b3c010929493c0244f3d37593afd6de04f71821d5136d9"
          }
        ],
        "patches": [
          "0d38234384870bfd34dfcb738a9083952656f0c766a0f5990b1893076b084b76"
        ],
        "hash": "ts3gkpltbgzr5y6nrfy6rzwbjmkscein"
      }
    ]
  }
}
```

Now:

* Patches and archive hashes are no longer included in the `package_hash`;
* Artifacts used in the build go in `sources`, and we tell you their checksum in the `spec.json`;
* `sources` will include resources for packages that have it;
* Patches are the same as before -- but only represented once; and
* The `package_hash` is a base32-encoded `sha1`, like other hashes in Spack, and it only
  tells you that the `package.py` changed.

The behavior of the DAG hash (which includes the `package_hash`) is basically the same as before, except now resources are included, and we can see differences in archives and resources directly in the `spec.json`

Note that we do not need to bump the spec meta version on this, as past versions of Spack can still read the new specs; they just will not notice the new fields (which is fine, since we currently do not do anything with them).

Among other things, this will more easily allow us to convert Spack specs to SBOM and track relevant security information (like `sha256`'s of archives). For example, we could do continuous scanning of a Spack installation based on these hashes, and if the `sha256`'s become associated with CVE's, we'll know we're affected.  (@vsoch FYI)

- [x] Add a method, `spec_attrs()` to `FetchStrategy` that can be used to describe a fetcher for a `spec.json`.
- [x] Simplify the way package_hash() is handled in Spack. Previously, it was handled as a special-case spec hash in `hash_types.py`, but it really doesn't belong there. Now, it's handled as part of `Spec._finalize_concretization()` and `hash_types.py` is much simpler.
- [x] Change `PackageBase.content_hash()` to `PackageBase.artifact_hashes()`, and include more information about artifacts in it.
- [x] Update package hash tests and make them check for artifact and resource hashes.

### DIFF
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -49,6 +49,7 @@ import spack.oci.opener
 import spack.util.archive
 import spack.util.crypto as crypto
 import spack.util.git
+import spack.util.spack_yaml as syaml
 import spack.util.url as url_util
 import spack.util.web as web_util
 import spack.version
@@ -109,6 +110,20 @@ class FetchStrategy:
         self.cache_enabled = not kwargs.pop("no_cache", False)
 
         self.package = None
+
+    def spec_attrs(self):
+        """Create a dictionary of attributes that describe this fetch strategy for a Spec.
+
+        This is included in the serialized Spec format to store provenance (like hashes).
+        """
+        attrs = syaml.syaml_dict()
+        if self.url_attr:
+            attrs["type"] = "archive" if self.url_attr == "url" else self.url_attr
+        for attr in self.optional_attrs:
+            value = getattr(self, attr, None)
+            if value:
+                attrs[attr] = value
+        return attrs
 
     def set_package(self, package):
         self.package = package
@@ -253,6 +268,16 @@ class URLFetchStrategy(FetchStrategy):
         self._curl: Optional[Executable] = None
         self.extension: Optional[str] = kwargs.get("extension", None)
         self._effective_url: Optional[str] = None
+
+    def spec_attrs(self):
+        attrs = super(URLFetchStrategy, self).spec_attrs()
+        if self.digest:
+            try:
+                hash_type = spack.util.crypto.hash_algo_for_digest(self.digest)
+            except ValueError:
+                hash_type = "digest"
+            attrs[hash_type] = self.digest
+        return attrs
 
     @property
     def curl(self) -> Executable:
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -33,7 +33,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 from pathlib import PurePath
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 import llnl.url
 import llnl.util
@@ -111,10 +111,18 @@ class FetchStrategy:
 
         self.package = None
 
-    def spec_attrs(self):
-        """Create a dictionary of attributes that describe this fetch strategy for a Spec.
+    def source_provenance(self) -> Dict:
+        """Create a metadata dictionary that describes the artifacts fetched by this FetchStrategy.
 
-        This is included in the serialized Spec format to store provenance (like hashes).
+        The returned dictionary is added to the content used to determine the full hash
+        for a package. It should be serializable as JSON.
+
+        It should include data like sha256 hashes for archives, commits for source
+        repositories, and any information needed to describe exactly what artifacts went
+        into a build.
+
+        If a package has no soruce artifacts, it should return an empty dictionary.
+
         """
         attrs = syaml.syaml_dict()
         if self.url_attr:
@@ -167,17 +175,6 @@ class FetchStrategy:
             bool: True if can cache, False otherwise.
         """
 
-    def source_id(self):
-        """A unique ID for the source.
-
-        It is intended that a human could easily generate this themselves using
-        the information available to them in the Spack package.
-
-        The returned value is added to the content which determines the full
-        hash for a package using `str()`.
-        """
-        raise NotImplementedError
-
     def mirror_id(self):
         """This is a unique ID for a source that is intended to help identify
         reuse of resources across packages.
@@ -228,9 +225,9 @@ class BundleFetchStrategy(FetchStrategy):
         """Report False as there is no code to cache."""
         return False
 
-    def source_id(self):
-        """BundlePackages don't have a source id."""
-        return ""
+    def source_provenance(self) -> Dict:
+        """BundlePackages don't have a source of their own."""
+        return {}
 
     def mirror_id(self):
         """BundlePackages don't have a mirror id."""
@@ -269,8 +266,14 @@ class URLFetchStrategy(FetchStrategy):
         self.extension: Optional[str] = kwargs.get("extension", None)
         self._effective_url: Optional[str] = None
 
-    def spec_attrs(self):
-        attrs = super().spec_attrs()
+    @property
+    def curl(self) -> Executable:
+        if not self._curl:
+            self._curl = web_util.require_curl()
+        return self._curl
+
+    def source_provenance(self) -> Dict:
+        attrs = super().source_provenance()
         if self.digest:
             try:
                 hash_type = spack.util.crypto.hash_algo_for_digest(self.digest)
@@ -278,15 +281,6 @@ class URLFetchStrategy(FetchStrategy):
                 hash_type = "digest"
             attrs[hash_type] = self.digest
         return attrs
-
-    @property
-    def curl(self) -> Executable:
-        if not self._curl:
-            self._curl = web_util.require_curl()
-        return self._curl
-
-    def source_id(self):
-        return self.digest
 
     def mirror_id(self):
         if not self.digest:
@@ -759,16 +753,6 @@ class GitFetchStrategy(VCSFetchStrategy):
         self.get_full_repo = kwargs.get("get_full_repo", False)
         self.git_sparse_paths = kwargs.get("git_sparse_paths", None)
 
-    def spec_attrs(self):
-        attrs = super().spec_attrs()
-
-        # need to fully resolve submodule callbacks for node dicts
-        submodules = attrs.get("submodules", None)
-        if submodules and callable(submodules):
-            attrs["submodules"] = submodules(self.package)
-
-        return attrs
-
     @property
     def git_version(self):
         return GitFetchStrategy.version_from_git(self.git)
@@ -807,9 +791,15 @@ class GitFetchStrategy(VCSFetchStrategy):
     def cachable(self):
         return self.cache_enabled and bool(self.commit)
 
-    def source_id(self):
-        # TODO: tree-hash would secure download cache and mirrors, commit only secures checkouts.
-        return self.commit
+    def source_provenance(self) -> Dict:
+        attrs = super().source_provenance()
+
+        # need to fully resolve submodule callbacks for node dicts
+        submodules = attrs.get("submodules", None)
+        if submodules and callable(submodules):
+            attrs["submodules"] = submodules(self.package)
+
+        return attrs
 
     def mirror_id(self):
         if self.commit:
@@ -1119,17 +1109,6 @@ class CvsFetchStrategy(VCSFetchStrategy):
     def cachable(self):
         return self.cache_enabled and (bool(self.branch) or bool(self.date))
 
-    def source_id(self):
-        if not (self.branch or self.date):
-            # We need a branch or a date to make a checkout reproducible
-            return None
-        id = "id"
-        if self.branch:
-            id += "-branch=" + self.branch
-        if self.date:
-            id += "-date=" + self.date
-        return id
-
     def mirror_id(self):
         if not (self.branch or self.date):
             # We need a branch or a date to make a checkout reproducible
@@ -1231,9 +1210,6 @@ class SvnFetchStrategy(VCSFetchStrategy):
     @property
     def cachable(self):
         return self.cache_enabled and bool(self.revision)
-
-    def source_id(self):
-        return self.revision
 
     def mirror_id(self):
         if self.revision:
@@ -1341,9 +1317,6 @@ class HgFetchStrategy(VCSFetchStrategy):
     @property
     def cachable(self):
         return self.cache_enabled and bool(self.revision)
-
-    def source_id(self):
-        return self.revision
 
     def mirror_id(self):
         if self.revision:
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -270,7 +270,7 @@ class URLFetchStrategy(FetchStrategy):
         self._effective_url: Optional[str] = None
 
     def spec_attrs(self):
-        attrs = super(URLFetchStrategy, self).spec_attrs()
+        attrs = super().spec_attrs()
         if self.digest:
             try:
                 hash_type = spack.util.crypto.hash_algo_for_digest(self.digest)
@@ -758,6 +758,16 @@ class GitFetchStrategy(VCSFetchStrategy):
         self.submodules_delete = kwargs.get("submodules_delete", False)
         self.get_full_repo = kwargs.get("get_full_repo", False)
         self.git_sparse_paths = kwargs.get("git_sparse_paths", None)
+
+    def spec_attrs(self):
+        attrs = super().spec_attrs()
+
+        # need to fully resolve submodule callbacks for node dicts
+        submodules = attrs.get("submodules", None)
+        if submodules and callable(submodules):
+            attrs["submodules"] = submodules(self.package)
+
+        return attrs
 
     @property
     def git_version(self):
--- a/lib/spack/spack/hash_types.py
+++ b/lib/spack/spack/hash_types.py
@@ -5,7 +5,6 @@
 """Definitions that control how Spack creates Spec hashes."""
 
 import spack.deptypes as dt
-import spack.repo
 
 hashes = []
 
@@ -13,20 +12,17 @@ hashes = []
 class SpecHashDescriptor:
     """This class defines how hashes are generated on Spec objects.
 
-    Spec hashes in Spack are generated from a serialized (e.g., with
-    YAML) representation of the Spec graph.  The representation may only
-    include certain dependency types, and it may optionally include a
-    canonicalized hash of the package.py for each node in the graph.
+    Spec hashes in Spack are generated from a serialized JSON representation of the DAG.
+    The representation may only include certain dependency types, and it may optionally
+    include a canonicalized hash of the ``package.py`` for each node in the graph.
 
-    We currently use different hashes for different use cases."""
+    """
 
-    def __init__(self, depflag: dt.DepFlag, package_hash, name, override=None):
+    def __init__(self, depflag: dt.DepFlag, package_hash, name):
         self.depflag = depflag
         self.package_hash = package_hash
         self.name = name
         hashes.append(self)
-        # Allow spec hashes to have an alternate computation method
-        self.override = override
 
     @property
     def attr(self):
@@ -51,18 +47,6 @@ dag_hash = SpecHashDescriptor(depflag=dt.BUILD | dt.LINK | dt.RUN, package_hash=
 #: Hash descriptor used only to transfer a DAG, as is, across processes
 process_hash = SpecHashDescriptor(
     depflag=dt.BUILD | dt.LINK | dt.RUN | dt.TEST, package_hash=True, name="process_hash"
-)
-
-
-def _content_hash_override(spec):
-    pkg_cls = spack.repo.PATH.get_pkg_class(spec.name)
-    pkg = pkg_cls(spec)
-    return pkg.content_hash()
-
-
-#: Package hash used as part of dag hash
-package_hash = SpecHashDescriptor(
-    depflag=0, package_hash=True, name="package_hash", override=_content_hash_override
 )
 
 
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -1807,7 +1807,7 @@ class PackageBase(WindowsRPath, PackageViewMixin, RedistributionMixin, metaclass
                     tty.debug(message.format(s=self))
 
             for resource in self._get_needed_resources():
-                sources.append(resource.fetcher.spec_attrs())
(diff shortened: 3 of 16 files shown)


iarspider:
@tgamblin since you will be working on package hashes, can you also try fixing #30720 ?

iarspider:
Quick question: will this fix the following problem - if Spack has a recipe for package, whose sources are taken from Git (or any other version control, I think), and I have a local repository with the same package but with different git URL, Spack will blindly use sources from Spack mirror?

tgamblin:
@spackbot run pipeline

var/spack/repos/builtin.mock/packages/hash-test4/package.py
```diff
@@ -14,7 +14,7 @@ class HashTest4(Package):
     homepage = "http://www.hashtest4.org"
     url = "http://www.hashtest1.org/downloads/hashtest4-1.1.tar.bz2"
 
-    version("1.1", "a" * 32)
+    version("1.1", sha256="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
```
scheibelp: I think `"a" * 32` is clearer than the full string:

* We mandate the length of the hash
* Future readers may wonder why would someone go through the trouble of typing `"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"` when they could just type `"aaa"`
* Making it clear up front that the string is 32 characters is a good hint that there's a reason behind it

lib/spack/spack/hash_types.py
```diff
@@ -5,28 +5,24 @@
...
 
-    Spec hashes in Spack are generated from a serialized (e.g., with
-    YAML) representation of the Spec graph.  The representation may only
-    include certain dependency types, and it may optionally include a
-    canonicalized hash of the package.py for each node in the graph.
+    Spec hashes in Spack are generated from a serialized JSON representation of the DAG.
+    The representation may only include certain dependency types, and it may optionally
+    include a canonicalized hash of the ``package.py`` for each node in the graph.
 
-    We currently use different hashes for different use cases."""
```
scheibelp, line 21: (question) are you removing this because it's no longer true, or because you don't think it's the right place to document this?

(we've discussed using a single hash everywhere for a while, but I don't think this PR handles that, so what I'm asking is if it already happened and this PR is just tidying up the docstring)

lib/spack/spack/package_base.py
```diff
@@ -1610,53 +1609,48 @@ def content_hash(self, content=None):
...
-                # TODO? in cases where a digest or source_id isn't available,
-                # should this attempt to download the source and set one? This
-                # probably only happens for source repositories which are
-                # referenced by branch name rather than tag or commit ID.
+                # if this is a develop spec, say so
                 env = spack.environment.active_environment()
                 from_local_sources = env and env.is_develop(self.spec)
+
+                # don't bother setting a source id if none is available, but warn if
+                # it seems like there should be one.
```
scheibelp, line 1804: If `source_id` is no longer going to be used in this function, then this comment should not refer to one.

lib/spack/spack/fetch_strategy.py
```diff
@@ -124,6 +125,20 @@ def __init__(self, **kwargs):
...
+        This is included in the serialized Spec format to store provenance (like hashes).
+        """
+        attrs = syaml.syaml_dict()
+        if self.url_attr:
+            attrs["type"] = "archive" if self.url_attr == "url" else self.url_attr
+        for attr in self.optional_attrs:
+            value = getattr(self, attr, None)
+            if value:
+                attrs[attr] = value
+        return attrs
```
scheibelp, line 134: It seems odd to embed configuration-YAML-awareness into the fetchers, what is this doing that `source_id` wasn't?

If this function is in fact necessary, I think it could use a better name and docstring (which I can suggest pending an answer to the above question).

lib/spack/spack/package_base.py
```diff
@@ -1597,7 +1596,7 @@ def all_patches(cls):
 
         return patches
 
-    def content_hash(self, content=None):
+    def artifact_hashes(self, content=None):
         """Create a hash based on the artifacts and patches used to build this package.
```
scheibelp: I think the initial docstring should be updated:

* this is returning a list of hashes vs. a single hash combining the details described in the docstring.
* might want to explicitly state that resources are now included.

lib/spack/spack/spec.py
```diff
@@ -3030,12 +3044,16 @@ def _finalize_concretization(self):
...
-                spec._cached_hash(ht.package_hash, force=True)
+                # package hash assignment has to happen before we mark concrete, so that
+                # we know what was *already* concrete.
+                # can't use self.package here b/c not concrete yet
+                pkg_cls = spack.repo.path.get_pkg_class(spec.name)
+                pkg = pkg_cls(spec)
 
-                # keep this check here to ensure package hash is saved
-                assert getattr(spec, ht.package_hash.attr)
+                # TODO: make artifact hashes a static method
```
scheibelp: I don't think this could ever be static: the artifact hashes include patches and resources (i.e. `_get_needed_resources` accesses `Package.spec` which is only available for instantiations).

lib/spack/spack/spec.py
```diff
@@ -3030,12 +3044,16 @@ def _finalize_concretization(self):
...
+                artifact_hashes = pkg.artifact_hashes()
+                spec._package_hash = artifact_hashes.pop("package_hash")
```
scheibelp, line 2933: It seems like it would be cleaner to call `package_hash()` directly here vs. depending on `artifact_hashes()` and then modifying the resulting dictionary: it seems like that's the direction that `artifact_hashes` is going (i.e. the name suggests we could exclude the contents of the package.py).

lib/spack/spack/spec.py
```diff
@@ -1917,23 +1921,24 @@ def to_node_dict(self, hash=ht.dag_hash):
...
             d["concrete"] = False
 
-        if "patches" in self.variants:
-            variant = self.variants["patches"]
-            if hasattr(variant, "_patches_in_order_of_appearance"):
-                d["patches"] = variant._patches_in_order_of_appearance
+        # We use the attribute here instead of `self.package_hash()` because this should
+        # *always* be assigned at concretization time. We don't want to try to compute a
+        # package hash for concrete spec where a) the package might not exist, or b) the
+        # `dag_hash` didn't include the package hash when the spec was concretized.
```
scheibelp: (sanity check/question) this was already true for this function and you are documenting it here - did I get that right?

lib/spack/spack/package_base.py
```diff
@@ -1610,53 +1609,48 @@ def content_hash(self, content=None):
         determinable) portions of the hash will be included.
 
         """
-        # list of components to make up the hash
-        hash_content = []
+        hashes = syaml.syaml_dict()
 
         # source artifacts/repositories
         # TODO: resources
```
scheibelp: This comment is no longer needed

lib/spack/spack/spec.py
```diff
@@ -1917,23 +1921,24 @@ def to_node_dict(self, hash=ht.dag_hash):
...
-            if not isinstance(package_hash, str) and isinstance(package_hash, bytes):
-                package_hash = package_hash.decode("utf-8")
-            d["package_hash"] = package_hash
+                    # sources may be dictionaries (for archives/resources)
+                    def order(source):
+                        if isinstance(source, dict):
+                            return syaml.syaml_dict(sorted(source.items()))
+                        return source
+
+                    d[key] = [order(source) for source in source_list]
```
scheibelp, line 2227: From looking through this PR it looks like the only thing that would ever need to be ordered is a dictionary. But FYI I had to spend some time cross-referencing to confirm that. If another editor ever adds a list of things, then this has the potential to become unordered (i.e. inconsistent hash).

You could add a consideration for lists in the locally-defined `order`, but that might be defeated by gaps in the type checking. Ideally if `_artifact_hashes` values are guaranteed to be iterables of comparable elements, then you could just call `sorted` without constructing a new type around it (i.e. don't create an `syaml_dict`: admittedly this interferes with the representation as listed in the PR description, which relates to point [1] of my review).

lib/spack/spack/test/util/package_hash.py
```diff
@@ -137,18 +135,38 @@ def test_content_hash_cannot_get_details_from_ast(mock_packages, config):
...
     spec1 = Spec("hash-test1@1.3").concretized()
     spec2 = Spec("hash-test2@1.3").concretized()
-    compare_hash_sans_name(False, spec1, spec2)
+
+    assert package_hash_equal_sans_name(spec1, spec2)
+
+    # the sources for these two packages will not be the same b/c their archive hashes differ
+    assert spec1.to_node_dict()["sources"] != spec2.to_node_dict()["sources"]
+
+    assert spec1.dag_hash() != spec2.dag_hash()
```
scheibelp: The specs have different package names, so these will always be unequal (i.e., I don't think this assert is confirming anything of value).

scheibelp (changes requested):
Point for consideration:

1. This is trending toward listing out more information individually in the node dict: what is the point of that? I don’t perceive a benefit other than that it becomes more-readable; IMO there are already mechanisms for getting a better interface to this information, in particular reading it in as a Spec object.

(I only have one point to make here but I numbered it because I reference this in one of the PR comments)

DarylGrunau:
Just pinging here to see what the status/blockers are for this MR to be merged.  Having the capability to produce an SBOM is becoming highly desirable at LANL (perhaps there's another MR that does this?).

vsoch:
I had done https://github.com/spack/spack-sbom and a follow up PR here to add a command in 2021 (since closed) but you probably want the changes in first that @tgamblin thinks are important.

tgamblin:
@scheibelp: RE:

> This is trending toward listing out more information individually in the node dict:

Yes. 

> what is the point of that? I don’t perceive a benefit other than that it becomes more-readable; IMO there are already mechanisms for getting a better interface to this information, in particular reading it in as a Spec object.

There are not reliable interfaces for retrieving this from an old installation.  `package.py` files may drift; we need an authoritative place for this stuff. The Spec is where authoritative provenance goes.

I put some motivation in the description, but here it is for posterity:

> We want this information to be available *from the concrete spec* so that it can be retrieved as provenance *without* referring to a `package.py` file.  Why?
> 
> 1. We want to be able to look up authoritative checksums, versions, etc. and pass them through scanners long *after* an installation takes place.
> 2. Right now you can look at the version on a `Spec`, then look up the checksum for a version from the `package.py`, but that's not reliable in the long term.  We may remove versions from `package.py` (in fact, we might want to as they become less relevant), and checksums for particular versions/releases may change over time.
> 3. Only the version is stored explicitly on the spec; that's not sufficient to know *exactly* what hashes a spec was built with.
> 
> Adding hashes explicitly gets around this and gets us closer to having all the information we need for a complete SBOM.

greenc-FNAL:
Does the [launch](https://thenewstack.io/protobom-paving-the-path-for-sbom-adoption/) of [Protobom](https://openssf.org/projects/protobom/) affect this PR in any way? Alternatively, should there be read-in converters for `spec.yaml` and `spack.lock` files in [Protobom](http://github.com/bom-squad/protobom) itself?

vsoch:
oh man, protobom! I love protocol buffers so this is :pinched_fingers:

lib/spack/spack/hash_types.py
tgamblin: It's redundant with the text above it, but if you think it's necessary, we could keep it.

lib/spack/spack/package_base.py
tgamblin: Thanks! fixed.

lib/spack/spack/package_base.py
tgamblin: fixed this.

lib/spack/spack/package_base.py
tgamblin: fixed

var/spack/repos/builtin.mock/packages/hash-test4/package.py
tgamblin: ok all these are fixed

lib/spack/spack/spec.py
tgamblin: the latter -- I tried to make the comment better.

lib/spack/spack/spec.py
tgamblin: I'm not sure I understand this comment. If the item is a list, it doesn't need to be deterministically ordered. It should already be.  We sort dictionaries but list order is up to the creator.

lib/spack/spack/test/util/package_hash.py
tgamblin: ok I've removed this.

lib/spack/spack/spec.py
tgamblin: yes. the intent was to make it a *class* method *that takes a spec*. I clarified the comment.

tgamblin: actually I just went ahead and made it a class method.

tgamblin: Actually I just removed the comment b/c I don't think it's really needed. It is awkward to instantiate a package just for this but we can fix that later in a more broad way.

lib/spack/spack/spec.py
tgamblin: Did you read the docstring for this method?

lib/spack/spack/spec.py
scheibelp: Everything `artifact_hashes` calls (i.e. right now just `_get_needed_resources`) orders things, so this is fine for now. My worry is that this behavior is disconnected from the need: I had to dig a couple layers down to confirm this, vs. just seeing it enforced here where it really matters; that seems fragile to me. 

You may be arguing that we depend on this for other info we add in `to_node_dict`, but most of that is part of `_cmp_node` (which presumably tests order), and `_artifact_hashes` isn't.

lib/spack/spack/test/cmd/install.py
```diff
@@ -812,28 +821,92 @@ def test_install_no_add_in_env(tmpdir, mock_fetch, install_mockery, mutable_mock
...
+        )
+        print()
+        print(initial_libelf.variants == final_libelf.variants)
+        print(initial_libelf.variants)
+        print(final_libelf.variants)
+        print()
+        print(initial_libelf == final_libelf)
+        print(id(initial_libelf), id(final_libelf))
+        print(initial_libelf in initial_concrete_specs)
+        print(initial_libelf in post_install_concrete_specs)
```
scheibelp: This looks like leftover debugging code (or some of it at least seems too detailed - like printing out the `_cmp_iter` results with no accompanying context - and I thought `print` would dump to the terminal in the middle of the test vs. `tty.debug`).

lib/spack/spack/spec.py
scheibelp: My point was: however this was gotten in `pkg.artifact_hashes()`, get it directly here instead. I only see this called in one place in this PR, so I would have assumed calling

```
spack.util.package_hash(self)
```

directly (i.e. here, in this function) would make more sense than delegating it to `artifact_hashes` and then deconstructing it after the fact here.

lib/spack/spack/test/spec_yaml.py
```diff
@@ -96,6 +96,22 @@ def test_invalid_json_spec(invalid_json, error_message):
...
+        "git-url-top-level@3.4",
+        "git-url-top-level@3.3",
+        "git-url-top-level@3.2",
+        "git-url-top-level@3.1",
+        "git-url-top-level@3.0",
+        "git-url-top-level@2.3",
+        "git-url-top-level@2.2",
+        "git-url-top-level@2.1",
+        "git-url-top-level@2.0",
+        "git-url-top-level@2.3",
```
scheibelp: * `git-url-top-level@2.3` appears twice in this list
* AFAICT, 2.3 and 2.2 are the same thing, and likewise for 2.1 and 2.0
* The 2.x versions are URL-based, so IMO the leading comment is confusing: `all types of git versions`
* `check_x_round_trip` uses `Spec.eq_dag`: since none of the added functionality in this PR appears in `_cmp_node`, I'm not sure this is the test to use (i.e. I think it would continue to pass even if someone deleted one "side" of `to_node_dict`/`from_node_dict`)

lib/spack/spack/package_base.py
```diff
@@ -1667,66 +1666,75 @@ def all_patches(cls):
...
 
         # patch sha256's
         # Only include these if they've been assigned by the concretizer.
         # We check spec._patches_assigned instead of spec.concrete because
         # we have to call package_hash *before* marking specs concrete
         if self.spec._patches_assigned():
-            hash_content.extend(
-                ":".join((p.sha256, str(p.level))).encode("utf-8") for p in self.spec.patches
-            )
+            hashes["patches"] = [p.sha256 for p in self.spec.patches]
```
scheibelp: Applying a patch at a different level is a significant difference: why remove this?

(no change needed) Also, it appears you dropped all calls to `.encode(utf-8)`: that makes sense given the removal of computations using `hashlib` (handling with `spack.util.hash.*` instead, which deals with encoding).

lib/spack/spack/spec.py
```diff
@@ -1440,6 +1440,12 @@ def __init__(
         for h in ht.hashes:
             setattr(self, h.attr, None)
 
+        # dictionary of source artifact hashes, set at concretization time
```
scheibelp: I think this comment is copied from next attr (and does not accurately describe what `_package_hash` is).

scheibelp (commented):
> package.py files may drift; we need an authoritative place for this stuff. 

If they drift, the package_hash should catch that

> The Spec is where authoritative provenance goes.

While I definitely agree, I worry about storing the same provenance in redundant ways in the ~spec~ node dictionary representation, purely for readability for end users (for a storage format that, AFAIK, was not intended for API purposes other than to be a stable form of storage for our own reconstruction).

One thing I want to check: is this intended to precisely identify the commit that we end up using for source control repo branch tips (e.g. develop in a git repo)? I don't see that right now in `spec_attrs`, but if that isn't the aim, then I'm not sure why this implementation retrieves info from the fetcher, when all the provenance should be on the version declaration. This is related to https://github.com/spack/spack/pull/32312/files#r1041676449

lib/spack/spack/test/cmd/install.py
tgamblin: > I thought print would dump to the terminal in the middle of the test vs. tty.debug).

It won't -- on success this is all hidden. On failure it's shown. But it's too much anyway so I got rid of it.

lib/spack/spack/spec.py
tgamblin: fixed this one

lib/spack/spack/test/spec_yaml.py
tgamblin: I addressed all of this but the last one -- I am not sure what you mean with the last point. I think it's still important to make sure these can be round-tripped.

lib/spack/spack/package_base.py
tgamblin: good point -- added this back.

lib/spack/spack/fetch_strategy.py
tgamblin: I tried to document this and remembered that I did not remove `source_id`.  See what you think of the new `source_provenance` function.

lib/spack/spack/package_base.py
tgamblin: Ok, adding this back breaks spec reading, which is something I had wanted to avoid.  Apparently:

1. We encode `patches` as a variant.
2. We encode `patches` at the top level of the spec dictionary as a list of hashes, and it needs to stay there to be read properly by old versions of Spack.  I *could* bump the spec version, but was hoping not to.  Trying to decide whether we should.
3. We should preserve patch level, but I don't think that it can go in either of the other two places. The place to put it is *probably* in the sources list, along with other artifacts. I don't really like this b/c now `patches` will appear in 3 places on the spec, but maybe the others can be deprecated.  Maybe this is a good enough reason to bump the spec version, or at least something we should do when we next bump it.

Still deciding what to do here.

PR #32319, opened 2022-08-22
SDK: Option to disable fortran support.

### LINKED ISSUE #32283: ecp-data-vis-sdk: remove enforced `+fortran`
Hi,

the `ecp-data-vis-sdk` package enforces `+fortran` on a lot of packages, e.g., ADIOS2 and HDF5. This is not ideal for me, building a C++ stack where I like to avoid Fortran dependencies as much as possible to save compile time.

Can you please make the hard-coded `+fortran` forwards a variant? :)

kwryankrattiger:
This option was requested to support MacOS systems that do not have fortran installed. Close #32283

@ax3l

### DIFF
--- a/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
+++ b/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
@@ -87,6 +87,9 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
     variant("vtkm", default=False, description="Enable VTK-m")
     variant("zfp", default=False, description="Enable ZFP")
 
+    # Language Options
+    variant("fortran", default=True, sticky=True, description="Enable fortran language features.")
+
     ############################################################
     # Dependencies
     ############################################################
@@ -94,9 +97,9 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
     amdgpu_target_variants = ["amdgpu_target={0}".format(x) for x in ROCmPackage.amdgpu_targets]
 
     dav_sdk_depends_on(
-        "adios2+shared+mpi+fortran+python+blosc+sst+ssc+dataman",
+        "adios2+shared+mpi+python+blosc+sst+ssc+dataman",
         when="+adios2",
-        propagate=["hdf5", "sz", "zfp"],
+        propagate=["hdf5", "sz", "zfp", "fortran"],
     )
 
     dav_sdk_depends_on("darshan-runtime+mpi", when="+darshan", propagate=["hdf5"])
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
```diff
@@ -87,26 +87,29 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
...
     cuda_arch_variants = ["cuda_arch={0}".format(x) for x in CudaPackage.cuda_arch_values]
     amdgpu_target_variants = ["amdgpu_target={0}".format(x) for x in ROCmPackage.amdgpu_targets]
 
     dav_sdk_depends_on(
-        "adios2+shared+mpi+fortran+python+blosc+sst+ssc+dataman",
+        "adios2+shared+mpi+python+blosc+sst+ssc+dataman",
         when="+adios2",
-        propagate=["hdf5", "sz", "zfp"],
+        propagate=["hdf5", "sz", "zfp", "fortran"],
     )
```
tldahlgren, line 103: I'm not familiar with the dependencies and am a bit confused.  I interpreted the need to add `fortran` as a variant to mean the SDK could be built *with and without* `fortran`.  So I was thinking there would be two `day_sdk_depends_on` here:

```
    dav_sdk_depends_on(
         "adios2+shared+mpi+python+blosc+sst+ssc+dataman",
         when="+adios2",
         propagate=["hdf5", "sz", "zfp"],
     )
    dav_sdk_depends_on(
         "adios2+shared+mpi+fortran+python+blosc+sst+ssc+dataman",
         when="+adios2+fortran",
         propagate=["hdf5", "sz", "zfp", "fortran"],
     )
```

But perhaps I didn't read the issue closely enough.

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
```diff
@@ -87,26 +87,29 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
...
 
     dav_sdk_depends_on("darshan-runtime+mpi", when="+darshan", propagate=["hdf5"])
     dav_sdk_depends_on("darshan-util", when="+darshan")
 
     dav_sdk_depends_on("faodel+shared+mpi network=libfabric", when="+faodel", propagate=["hdf5"])
 
-    dav_sdk_depends_on("hdf5@1.12: +shared+mpi+fortran", when="+hdf5")
+    dav_sdk_depends_on("hdf5@1.12: +shared+mpi", when="+hdf5", propagate=["fortran"])
```
tldahlgren, line 110: Similarly, I was expecting something along the lines:

```suggestion
    dav_sdk_depends_on("hdf5@1.12: +shared+mpi", when="+hdf5")
    dav_sdk_depends_on("hdf5@1.12: +shared+mpi+fortran", when="+hdf5+fortran", propagate=["fortran"])
```

tldahlgren: And the same approach for line 112 and 164-165 below.

tldahlgren (commented):
I have some questions about the changes.

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
kwryankrattiger: So the purpose of the `dav_sdk_depends_on` wrapper is, in part, used to automatically add the positive and negative of the variant propagation without having to write out the boiler plate by hand. Adding `"fortran"` to the list of variants adds 

```python
depends_on("adios2 +fortran", when="+adios2+fortran")
depends_on("adios2 ~fortran", when="+adios2~fortran")
```

This was done because the nature of the SDK requires propagating a  lot of variants and it is far easier to read the intent quickly this way, rather than having to parse out all of the variations on every line.

tldahlgren: I should've looked more closely at `dav_sdk_depends_on`, especially lines 28-50.

tldahlgren (approved):
LGTM

tldahlgren:
Will wait another ~4 days to merge to give time in case @ax3l wants to comment.

ax3l (approved):
Love it, thanks a lot!

PR #32324, opened 2022-08-23
New package: py-pymdown-extensions

myan-rarcellino-nnl:
This package appears to use a metadata hook to define dependencies so you'll have to look at hatch_build.py to find the location of the requirements file


adamjstewart (approved):
Thanks, I never would have found those dependencies!

PR #32325, opened 2022-08-23
cpu-features: fix fetch failure

iarspider:
`master` branch was renamed to `main`


var/spack/repos/builtin/packages/cpu-features/package.py
```diff
@@ -12,7 +12,7 @@ class CpuFeatures(CMakePackage):
     homepage = "https://github.com/google/cpu_features"
     git = "https://github.com/google/cpu_features.git"
 
-    version("develop", branch="master")
+    version("develop", branch="main")
```
adamjstewart:
```suggestion
    version("main", branch="main")
    version("develop", branch="main", deprecated=True)
```
Let's deprecate the old name and rename it to the current name

adamjstewart (approved):

PR #32326, opened 2022-08-23
whizard: Fix passing of build options, update versions

graeme-a-stewart:
The dependency of whizard on `libtirpc` is now correctly passed down as an autotools option.

Update known versions of package with 3.0.2 and 3.0.3.

### DIFF
--- a/var/spack/repos/builtin/packages/whizard/package.py
+++ b/var/spack/repos/builtin/packages/whizard/package.py
@@ -102,7 +102,7 @@ class Whizard(AutotoolsPackage):
     def configure_args(self):
         spec = self.spec
         args = [
-            "TIRPC_CFLAGS=-I%s/include/tirpc" % spec["libtirpc"].prefix,
+            "TIRPC_CFLAGS=-I%s" % spec["libtirpc"].prefix.include.tirpc,
             "TIRPC_LIBS=-ltirpc",
             "--enable-hepmc=%s" % ("no" if "hepmc=off" in spec else "yes"),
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Fixed quotes in last commit

var/spack/repos/builtin/packages/whizard/package.py
```diff
@@ -100,6 +102,8 @@ def setup_build_environment(self, env):
     def configure_args(self):
         spec = self.spec
         args = [
+            "TIRPC_CFLAGS=-I%s/include/tirpc" % spec["libtirpc"].prefix,
```
tldahlgren: FYI. You could "simplify" with (assuming `tirpc` is also a directory):

```suggestion
            "TIRPC_CFLAGS=-I%s" % spec["libtirpc"].prefix.include.tirpc,
```

tldahlgren (dismissed):
Confirmed sha256.  Will defer merge decision until @vvolkl has a chance to comment.

vvolkl (dismissed):
Looks good to me - thanks!

var/spack/repos/builtin/packages/whizard/package.py
graeme-a-stewart: Thanks for the suggestion @tldahlgren - yes, that does work in this case, so I have done it in this more "spackish" way.

vvolkl (approved):

tldahlgren (approved):

PR #32328, opened 2022-08-23
openmpi: disable-per-user-config-files

haampie:
This feature requires `HOME` to be set, otherwise the compiler wrapper errors out.

```
$ env -i "PATH=$PATH" mpicc --version
 [nid003052:41640] Error: Unable to get the user home directory
 [nid003052:41640] Error: Unable to get the user home directory
 --------------------------------------------------------------------------
 It looks like opal_init failed for some reason;
 ```

Disabling it seems sensible.

(Could be a regression too by the way, only pops up when building `@main`).

### DIFF
--- a/var/spack/repos/builtin/packages/openmpi/package.py
+++ b/var/spack/repos/builtin/packages/openmpi/package.py
@@ -904,7 +904,11 @@ class Openmpi(AutotoolsPackage, CudaPackage):
 
     def configure_args(self):
         spec = self.spec
-        config_args = ["--enable-shared", "--disable-silent-rules"]
+        config_args = [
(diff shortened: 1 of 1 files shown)


haampie:
What do you think @hppritcha?

hppritcha:
I don't like this as a default.  I'm sure that are a number of OMPI users make use of this per-user config file capability.

haampie:
Should I file a bug instead? I find it surprising that `HOME` is required for `mpicc` to work.

hppritcha:
@haampie yes please open an issue against open mpi

haampie:
Closing this PR, thanks for the review!

PR #32330, opened 2022-08-23
gcc: add support for the D language (GDC)

skosukhin:
This turned out more complicated than I thought. The problem is that

> In order to build GDC, the D compiler, you need a working GDC compiler (GCC version 9.1 or later) and D runtime library, ‘libphobos’, as the D front end is written in D.
>
> Versions of GDC prior to 12 can be built with an ISO C++11 compiler, which can then be installed and used to bootstrap newer versions of the D front end.
>
> It is strongly recommended to use an older version of GDC to build GDC. More recent versions of GDC than the version built are not guaranteed to work and will often fail during the build with compilation errors relating to deprecations or removed features.

See https://gcc.gnu.org/install/prerequisites.html#GDC-prerequisite.

The suggested solution is to enable the detection of `gdc` in `determine_variants`, find `gcc` as a package in `os.path.dirname(self.compiler.cc)` and use its `spec.extra_attributes["compilers"]["d"]` as the path to the right `gdc`.

### DIFF
--- a/var/spack/repos/builtin/packages/gcc/package.py
+++ b/var/spack/repos/builtin/packages/gcc/package.py
@@ -980,14 +980,34 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage):
             [self.__class__], path_hints=([compiler_dir] if os.path.isdir(compiler_dir) else None)
         )
 
-        # We consider only packages that satisfy the compiler spec with an additional constraint:
-        required_spec = Spec("languages=c,c++,d").constrained(repr(self.compiler.spec))
+        # We consider only packages that satisfy the following constraint:
+        required_spec = Spec("languages=c,c++,d")
         candidate_specs = [
             p.spec
             for p in filter(
                 lambda p: p.spec.satisfies(required_spec), detected_packages.get(self.name, ())
             )
         ]
+
+        if candidate_specs:
+            # We now need to filter specs that match the compiler version:
+            compiler_spec = Spec(repr(self.compiler.spec))
+
+            # First, try to filter specs that satisfy the compiler spec:
+            new_candidate_specs = list(
+                filter(lambda s: s.satisfies(compiler_spec), candidate_specs)
+            )
+
+            # The compiler version might be more specific than what we can detect. For example, the
+            # user might have "gcc@10.2.1-sys" as the compiler spec in compilers.yaml. In that
+            # case, we end up with an empty list of candidates. To circumvent the problem, we try
+            # to filter specs that are satisfied by the compiler spec:
+            if not new_candidate_specs:
+                new_candidate_specs = list(
+                    filter(lambda s: compiler_spec.satisfies(s), candidate_specs)
+                )
+
+            candidate_specs = new_candidate_specs
 
         error_nl = "\n    "  # see SpackError.__str__()
 
--- a/var/spack/repos/builtin/packages/gcc/package.py
+++ b/var/spack/repos/builtin/packages/gcc/package.py
@@ -87,13 +87,17 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage):
     version("4.5.4", sha256="eef3f0456db8c3d992cbb51d5d32558190bc14f3bc19383dd93acc27acc6befc")
 
     # We specifically do not add 'all' variant here because:
-    # (i) Ada, Go, Jit, and Objective-C++ are not default languages.
+    # (i) Ada, D, Go, Jit, and Objective-C++ are not default languages.
     # In that respect, the name 'all' is rather misleading.
     # (ii) Languages other than c,c++,fortran are prone to configure bug in GCC
     # For example, 'java' appears to ignore custom location of zlib
     # (iii) meaning of 'all' changes with GCC version, i.e. 'java' is not part
     # of gcc7. Correctly specifying conflicts() and depends_on() in such a
     # case is a PITA.
+    #
+    # Also note that some languages get enabled by the configure scripts even if not listed in the
+    # arguments. For example, c++ is enabled when the bootstrapping is enabled and lto is enabled
+    # when the link time optimization support is enabled.
     variant(
         "languages",
         default="c,c++,fortran",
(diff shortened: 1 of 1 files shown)


skosukhin:
@alalazo do you have an idea why the clingo bootstrapping failed?

skosukhin:
@scottwittenburg maybe you could explain why the test fails with `Error: Package 'gcc' not found.` here?

var/spack/repos/builtin/packages/gcc/package.py
```diff
@@ -102,6 +106,7 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage):
             "brig",
             "c",
             "c++",
+            "d",
```
alalazo, line 109: Does:
```suggestion
            conditional("d", when="@9.1:"),
```
work too? If it does (as it should) you can delete a bunch of conflicts.

skosukhin: It looks like we could use `conditional` for several other `languages` too. Do you mind if we keep this part of the PR as is and I will come up with a follow-up PR that makes use of `conditional` whenever possible?

alalazo: No I don't.

alalazo:
> Do you have an idea why the clingo bootstrapping failed?

No idea, and it failed only on Python 2.7 Retrying to check if it was a transient failure or if it is consistent. 

EDIT: I could reproduce the failure locally, but it happens after bootstrapping.

alalazo:
Here's what's failing for me:
```console
$ python2.7 bin/spack -p --lines 20 spec mpileaks%gcc ^dyninst@10.0.0 ^elfutils@0.170
         29361327 function calls (28161400 primitive calls) in 20.847 seconds

   Ordered by: internal time
   List reduced from 11185 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
4595/4586    1.238    0.000   19.121    0.004 {imp.load_source}
101470/99795    1.207    0.000    9.055    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/spec.py:1172(__init__)
   148204    0.771    0.000    0.875    0.000 /usr/lib/python2.7/shlex.py:120(read_token)
    38389    0.665    0.000    3.067    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/spec.py:3760(_dup)
    62499    0.584    0.000    0.780    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/version.py:254(_generate_seperators_and_components)
   139170    0.550    0.000    0.550    0.000 {built-in method from_iterable}
   106878    0.463    0.000    1.271    0.000 /usr/lib/python2.7/re.py:321(scan)
373586/283914    0.441    0.000    1.853    0.000 /home/culpo/PycharmProjects/spack/lib/spack/llnl/util/lang.py:299(tuplify)
1159054/1051489    0.433    0.000    1.508    0.000 /home/culpo/PycharmProjects/spack/lib/spack/llnl/util/lang.py:301(<genexpr>)
    46411    0.372    0.000    0.975    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/spec.py:864(select)
106878/79904    0.365    0.000    1.645    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/parse.py:59(lex_word)
2078368/2078366    0.358    0.000    0.524    0.000 {isinstance}
   139859    0.331    0.000    0.363    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/variant.py:534(__init__)
45979/44831    0.292    0.000    1.457    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/spec.py:1614(traverse_edges)
    68294    0.289    0.000    0.336    0.000 /usr/lib/python2.7/shlex.py:23(__init__)
   573815    0.259    0.000    0.431    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/parse.py:117(accept)
266924/114700    0.252    0.000    0.466    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/directives.py:277(remove_directives)
    68294    0.235    0.000    1.643    0.000 /usr/lib/python2.7/shlex.py:274(split)
   228635    0.230    0.000    0.328    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/parse.py:53(token)
    40000    0.230    0.000   13.775    0.000 /home/culpo/PycharmProjects/spack/lib/spack/spack/directives.py:357(_depends_on)


==> Error: Package 'gcc' not found.

```

alalazo:
Oddly enough, it happens only on Python 2.7 and when a cache needs to be generated for the builtin repository

alalazo:
Back from lunch, added a couple of debug prints:
```console
[ ... ]
spack.pkg.builtin.gcc
invalid syntax in /home/culpo/PycharmProjects/spack/var/spack/repos/builtin/packages/gcc/package.py, line 782
[ ... ]
```

var/spack/repos/builtin/packages/gcc/package.py
```diff
@@ -720,6 +776,18 @@ def configure_args(self):
         options.append("--with-boot-ldflags=" + boot_ldflags)
         options.append("--with-build-config=spack")
 
+        if "languages=d" in spec:
+            # Phobos is the standard library for the D Programming Language. The documentation says
+            # that on some targets, 'libphobos' isn’t enabled by default, but compiles and works
```
alalazo: I love encodings:
```suggestion
            # that on some targets, 'libphobos' is not enabled by default, but compiles and works
```

skosukhin:
@alalazo I'm so sorry. Thanks!

alalazo:
No worries, these kind of errors are always hard to find, since we prevent the real cause:
```console
        try:
            module = importlib.import_module(fullname)
        except ImportError:
            raise UnknownPackageError(pkg_name)
```
from entering the log (so partially Spack's fault too).

alalazo (approved):
Thanks! Let's see how this goes!

PR #32331, opened 2022-08-23
minc-toolkit: add MINC_TOOLKIT environment variable at runtime

bcdarwin:
This variable is used by some programs both internal and external to the
toolkit itself to discover shared objects, data, etc.


tldahlgren (approved):
LGTM

PR #32332, opened 2022-08-23
add bug fix for geant4-10.7.2

bungernut:
Added a simple patch to prevent compilation error of geant4-10.7.2 associated with bug 2444 (https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2444). This was fixed by geant4 in 10.7.3 release.

### DIFF
--- a/var/spack/repos/builtin/packages/geant4/package.py
+++ b/var/spack/repos/builtin/packages/geant4/package.py
@@ -124,6 +124,8 @@ class Geant4(CMakePackage):
     depends_on("motif", when="+motif")
     depends_on("qt@5: +opengl", when="+qt")
 
+    # Fixes bug 2444 on 10.7.2: https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2444
+    patch("geant4-10.7.2_bug2444.patch", when="@10.7.2")
     # As released, 10.03.03 has issues with respect to using external
     # CLHEP.
     patch("CLHEP-10.03.03.patch", level=1, when="@10.3.3")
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/geant4/package.py
```diff
@@ -124,6 +124,8 @@ class Geant4(CMakePackage):
     depends_on("motif", when="+motif")
     depends_on("qt@5: +opengl", when="+qt")
 
+    # Fixes bug 2444 on 10.7.2: https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2444
```
tldahlgren, line 127: Since this is a known bug in the software, we generally prefer the patches be upstreamed and can be pulled using a URL.

tldahlgren: Any chance of getting that done?

bungernut: I cannot say for sure but I have reasons to believe that is not easy.
1) They have already patched it in 10.7.3 and recommend the upgrade (which changes a lot of other things that we don't necessarily want)
2) I've not seen this sort of thing from Grant4 team before. They tend to release only major version patches not individual bug fixes
3) I'm not even sure where they actively develop. They only import major versions to git from some private repo it seems.

tldahlgren: Okay.  Thanks for the response.

tldahlgren (approved):
LGTM.  Will defer merge decision until @drbenmorgan has a chance to comment.

wdconinc:
Don't want to mix myself into this unnecessarily (yet I do), but since there is a more recent upstream bugfix version that contains this fix and works with this compiler, can't we simply mark 10.7.2 as a conflict for `gcc@11:` and add in the `msg` that the build issue in 10.7.2 is fixed in 10.7.3? This patch is rather finely targeted to those who specifically want a bugfix release that's not the most recent one.

bungernut:
Let me say only that often a specific version of geant4 is requested, each version also changes the fundamental physics in the code. I know in my case we have not validated the physics changes in 10.7.3 or 11+ yet and will continue to need 10.7.2.

My patch works for me, I have been applying it to spack0.18, but I assumed the community might like to benefit.
________________________________

tldahlgren:
ping @drbenmorgan

alecbcs:
@drbenmorgan it looks like this PR is still waiting on your response. Is this something we should push on to merge or is this PR no longer relevant?

drbenmorgan:
I think given the age, let's close. If it's vital to the submitter, a clean PR would be easier.

PR #32335, opened 2022-08-23
New package: py-griffe and new build backend: py-pdm-pep517

myan-rarcellino-nnl:
griffe requires a build backend, pdm-pep517, which I have never heard of. I added both packages in this PR for more context, but I can create separate PRs if necessary. Do I need to add a few sentences to the build backend section of the PythonPackage docs?

### DIFF
--- a/lib/spack/docs/build_systems/pythonpackage.rst
+++ b/lib/spack/docs/build_systems/pythonpackage.rst
@@ -50,8 +50,9 @@ important to understand.
    include `setuptools <https://setuptools.pypa.io/>`__,
    `flit <https://flit.pypa.io/>`_,
    `poetry <https://python-poetry.org/>`_,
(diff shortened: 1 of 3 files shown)


tldahlgren (dismissed):
Confirmed sha256.  Will defer merge until @adamjstewart has a chance to comment.

adamjstewart:
> Do I need to add a few sentences to the build backend section of the PythonPackage docs?

Yes, if you don't mind, that would be really helpful. The template should be obvious, just look for all mentions of "flit" or "poetry" and add something about PDM in the same locations. PDM is one of the 4 options listed at https://packaging.python.org/en/latest/tutorials/packaging-projects/ so it must be somewhat common, or will be in the future.

Packages themselves look great.

adamjstewart (approved):

Issue #32336, opened 2022-08-23
Inherited packages break module-scoped variables


nicholas-sly:
### Steps to reproduce

Under an unknown set of conditions, running `spack install llvm+python` will generate this error.

### Error message

```
==> [2022-08-22-17:03:28.606870] llvm: Executing phase: 'cmake'
==> [2022-08-22-17:04:48.618765] llvm: Executing phase: 'build'
==> [2022-08-22-17:33:24.980516] llvm: Executing phase: 'install'
==> [2022-08-22-17:34:22.572029] Error: NameError: name 'python_platlib' is not defined

/usr/WS2/sly1/tce/spack/var/spack/repos/builtin/packages/llvm/package.py:760, in post_install:
        757                ninja()
        758                ninja("install")
        759        if "+python" in self.spec:
  >>    760            install_tree("llvm/bindings/python", python_platlib)
        761
        762            if "+clang" in self.spec:
        763                install_tree("clang/bindings/python", python_platlib)

See build log for details:
  /var/tmp/sly1/spack-stage-alastor/spack-stage-llvm-12.0.1-gxtmcgu3kn34e3d2n53wzbduw4i3ukg3/spack-build-out.txt

Traceback (most recent call last):
  File "/my/path/spack/lib/spack/spack/build_environment.py", line 1062, in _setup_pkg_and_run
    return_value = function(pkg, kwargs)
  File "/my/path/spack/lib/spack/spack/installer.py", line 2013, in build_process
    return installer.run()
  File "/my/path/spack/lib/spack/spack/installer.py", line 1875, in run
    self._real_install()
  File "/my/path/spack/lib/spack/spack/installer.py", line 1974, in _real_install
    phase(pkg.spec, pkg.prefix)
  File "/my/path/spack/lib/spack/spack/package_base.py", line 150, in phase_wrapper
    callback(instance)
  File "/my/path/spack/var/spack/repos/builtin/packages/llvm/package.py", line 761, in post_install
NameError: name 'python_platlib' is not defined
==> [2022-08-22-17:34:22.572342] Flagging llvm-12.0.1-gxtmcgu3kn34e3d2n53wzbduw4i3ukg3 as failed: NameError: name 'python_platlib' is not defined
```

### Information on your system

```
* **Spack:** 0.19.0.dev0 (c64b2a00395394cc767538eede9cdd3359b2705e)
* **Python:** 3.9.12
* **Platform:** linux-rhel8-ivybridge
* **Concretizer:** clingo
[...]

PR #32339, opened 2022-08-23
Update LLVM post-install steps.

nicholas-sly:
I have run into issues with the LLVM installation with Spack.  

One involves the attempt to install files from the bin directory of the build stage to the prefix/libexec/llvm directory of the installation when those files already exist.  A check was added to skip this error-causing code if it isn't necessary.

Another involves issues with the python and clang bindings section.  My builds have not needed the python binding installation and needed the clang bindings in a slightly different place.  I have been building with these changes in a separate package repo, but this led to its own issues where the `python_platlib` variable pulled from the python `setup_dependent_build` function wasn't available if the LLVM package existed in the separate package repo (noted in issue #32336 ).

@becker33


alalazo:

PR #32340, opened 2022-08-23
BuildEnvironment: accumulate module changes to poke to all relevant modules

### LINKED ISSUE #32336: Inherited packages break module-scoped variables

becker33:
Fixes #32336 

Currently, module changes from `setup_dependent_package` are applied only to the module of the package class, but not to any parent classes' modules between the package class module and `spack.package_base`.

In this PR, we create a custom class to accumulate module changes, and apply those changes to each class that requires it. This design allows us to code for a single module, while applying the changes to multiple modules as needed under the hood, without requiring the user to reason about package inheritance.

@tgamblin you'll probably want to look at this.


becker33:
@spackbot fix style

becker33:
@spackbot fix style

lib/spack/spack/build_environment.py
```diff
@@ -133,6 +133,24 @@ def should_set_parallel_jobs(jobserver_support=False):
...
+            super(ModuleChangeAccumulator, self).__setattr__(key, value)
+        else:
+            self.changes[key] = value
+
+    def __getattr__(self, key):
+        # Return what's in our dict, else delegate to superclass
+        if key in self.changes:
+            return self.changes[key]
+        else:
+            return super(ModuleChangeAccumulator, self).__getattribute__(key)
```
tgamblin: I think you can replace this with a `spack.util.pattern.Bunch` and use `__dict__` to get the attributes.

lib/spack/spack/build_environment.py
```diff
@@ -993,8 +1011,26 @@ def add_modifications_for_dep(dep):
...
+
+            # Execute changes as if on a single module
+            # copy dict to ensure prior changes are available
+            module_accumulator = ModuleChangeAccumulator()
+            module_accumulator.__dict__ = spec.package.module.__dict__
+            dpkg.setup_dependent_package(module_accumulator, spec)
+
+            for module in modules:
+                for key, value in module_accumulator.changes.items():
+                    module.__dict__[key] = value  # setattr(module, _) python 3.7+
```
tgamblin: Pretty sure this works:

```suggestion
            changes = spack.util.pattern.Bunch(spec.package.module.__dict__)
            dpkg.setup_dependent_package(changes, spec)

            for module in modules:
                module.__dict__.update(spec.package.module.__dict__)
```

tgamblin (changes requested):

lib/spack/spack/build_environment.py
becker33: Great idea, I had no idea we had that utility. Cleans this up a ton

becker33:
@spackbot fix style

tgamblin (approved):

haampie:
This was a breaking change :(

It's now impossible to read properties from the parent module.

How are build deps like `gmake` / `ninja` supposed to get the parent `module.make_jobs`?

lib/spack/spack/build_environment.py
```diff
@@ -993,8 +994,24 @@ def add_modifications_for_dep(dep):
...
+            for cls in inspect.getmro(type(spec.package)):
+                module = cls.module
+                if module == spack.package_base:
+                    break
+                modules.append(module)
+
+            # Execute changes as if on a single module
+            # copy dict to ensure prior changes are available
+            changes = spack.util.pattern.Bunch()
+            dpkg.setup_dependent_package(changes, spec)
```
alalazo, line 1010: I don't think many packages would be affected, but this potentially breaks previous semantics since we are not passing something that resembles the module. So if a package was accessing module properties, it won't find them in the `Bunch()`. Also, the `PackageBase` docstring is now misleading, since it says we are passing `package.module` as an argument.

EDIT: Sorry for the double posting, I should have refreshed the browser :laughing:

PR #32345, opened 2022-08-23
minc-toolkit: add runtime flex dependency

bcdarwin:
Some programs such as `minccalc` depend on libfl.so at runtime.


tldahlgren (approved):
LGTM

PR #32350, opened 2022-08-24
add eztrace 2.0 release

trahay:


var/spack/repos/builtin/packages/eztrace/package.py
```diff
@@ -15,7 +15,7 @@ class Eztrace(Package):
     git = "https://gitlab.com/eztrace/eztrace.git"
 
     version("master", branch="master")
-    version("2.0-rc1", sha256="2ad322df5aeb0c668ebadbfeb3b21c8841831b2c15944d86d1d217f738fee74e")
+    version("2.0", sha256="899788d67e651bbe54abf39099f8f28a843d6676")
```
tldahlgren: How did you calculate that?  I get the following from `spack checksum`:

```suggestion
    version("2.0", sha256="67bd296f059cdfab303c62f674af3e1e858213d6945bd79cb8ede4a035c0c2d6")
```

tldahlgren (changes requested):
The hash seems to be off.

tldahlgren:
Note the hash is being called out in the test failures:

```
E       AssertionError: assert [] == ['All packages must use sha2...ms. eztrace@2.0 uses sha1.']
E         Right contains more items, first extra item: 'All packages must use sha256 checksums. eztrace@2.0 uses sha1.'
E         Full diff:
E         - []
E         + ['All packages must use sha256 checksums. eztrace@2.0 uses sha1.']
```

trahay:
You're right ! I messed up with the hash. Sorry about that.

tldahlgren:
> You're right ! I messed up with the hash. Sorry about that.

No problem

tldahlgren (approved):

PR #32351, opened 2022-08-24
lammps: added build dependency to py-setuptools

prs513rosewood:
LAMMPS' setup.py uses setuptools as of lammps/lammps@2ed8e5cf022. For some reason, installing `lammps+python` did not report a failed build if setuptools was missing, causing an incomplete install.

Question: when built with python bindings, LAMMPS does not `extends("python")`, should it?


tldahlgren (approved):
LGTM

tldahlgren:
> Question: when built with python bindings, LAMMPS does not `extends("python")`, should it?

Depends.  There's a rule of thumb given at https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html?highlight=extends#extends-vs-depends-on.

I'm going to assume if you want to make the change it will be in a different PR.

PR #32355, opened 2022-08-24
Bumping HPX requirements for FleCSI V2.0 [DO-NOT-MERGE-YET (see comments)]

hkaiser:
The FleCSI v2.x HPX backend depends on fixes available with HPX v1.8.1 only.

### DIFF
--- a/var/spack/repos/builtin/packages/flecsi/package.py
+++ b/var/spack/repos/builtin/packages/flecsi/package.py
@@ -135,7 +135,7 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("legion+cuda", when="backend=legion +cuda @2.0:")
     depends_on("legion+rocm", when="backend=legion +rocm @2.0:")
     depends_on("hdf5@1.10.7:", when="backend=legion +hdf5 @2.0:")
-    depends_on("hpx@1.8.1: cxxstd=17 malloc=system", when="backend=hpx @2.0:")
+    depends_on("hpx@1.9.1: cxxstd=17 malloc=system", when="backend=hpx @2.0:")
     depends_on("mpi", when="@2.0:")
     depends_on("mpich@3.4.1:", when="@2.0: ^mpich")
     depends_on("openmpi@4.1.0:", when="@2.0: ^openmpi")


rspavel:
Is that true for all versions of `flecsi@2`? So 1.3.1 or whatever is there currently never worked?

Or is there a changing point? Currently the public flecsi versions are pretty coarse grain. But if we can avoid breaking compatibility with actual `flecsi@2.0` for a change that only applies to `flecsi@2.1.fourMonthsLater` that would be good to get into the habit of.

hkaiser:
> Is that true for all versions of `flecsi@2`? So 1.3.1 or whatever is there currently never worked?
> 
> Or is there a changing point? Currently the public flecsi versions are pretty coarse grain. But if we can avoid breaking compatibility with actual `flecsi@2.0` for a change that only applies to `flecsi@2.1.fourMonthsLater` that would be good to get into the habit of.

Anything before HPX 1.8 was never really tested, so the 1.3.1 in the script is probably an oversight from earlier versions of FleCSI.

tldahlgren (dismissed):
LGTM

tldahlgren:
Will defer merge to give at least one maintainer a chance to comment.

tldahlgren:
Looks like there were some node/pod issues.

@spackbot run pipeline

rbberger (changes requested):
Given that this is for the `2` branch and hasn't been merged into any flecsi/flecsi branch yet, this change should be merged at a later time. Only after a tagged `2` release that includes the HPX backend.

var/spack/repos/builtin/packages/flecsi/package.py
```diff
@@ -122,7 +122,7 @@ class Flecsi(CMakePackage, CudaPackage):
     depends_on("legion+hdf5", when="backend=legion +hdf5 @2.0:")
     depends_on("legion +kokkos +cuda", when="backend=legion +kokkos +cuda @2.0:")
     depends_on("hdf5@1.10.7:", when="backend=legion +hdf5 @2.0:")
-    depends_on("hpx@1.3.0 cxxstd=17 malloc=system", when="backend=hpx @2.0:")
+    depends_on("hpx@1.8.1 cxxstd=17 malloc=system", when="backend=hpx @2.0:")
```
rbberger: This should allow 1.8.1 and newer:
```suggestion
    depends_on("hpx@1.8.1: cxxstd=17 malloc=system", when="backend=hpx @2.0:")
```

rbberger (changes requested):

var/spack/repos/builtin/packages/flecsi/package.py
hkaiser: @rbberger thanks for this! I also think that we should make it depend on the soon to be released HPX V1.9.1 instead.

rbberger: Ok, but in any case please add the `:` to future proof it.

bernhardkaindl: @hkaiser @rbberger I tried to build 2.x with +hpx but despite having checked that hpx-1.8.1 / 1.9.1 are in the spack build of flecsi, the CMake variable FOUND_MPX was not set:
```py
  >> 61    CMake Error at cmake/hpx.cmake:30 (message):
     62      HPX is required for this configuration
```
Also it seems that flecsi-2.2.1 does not have the include files it likes to include in it's tree:
```c++
  >> spack-src/flecsi/run/backend.hh:24:10: fatal error: hpx/context.hh: No such file or directory
     100       24 | #include "hpx/context.hh"
     101          |          ^~~~~~~~~~~~~~~~
spack-src/flecsi/exec/backend.hh:58:10: fatal error: flecsi/exec/hpx/policy.hh: No such file or directory
     108       58 | #include "flecsi/exec/hpx/policy.hh"
     109          |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

hkaiser: I have updated the package to HPX V1.9.1 now.

rbberger:
@tldahlgren we can merge this. We'd rather have this already in the next Spack release.

alalazo (approved):

PR #32356, opened 2022-08-24
tixi: new variants (fortran,shared)

melven:
Since some tixi 3 versions, additional CMake flags are needed to build
tixi with shared libraries, respectively with Fortran support.


tldahlgren (approved):
LGTM

PR #32357, opened 2022-08-24
py-pyopengl: new package

melven:
New package file for pyopengl (simple pypi wheel).

### DIFF
--- a/var/spack/repos/builtin/packages/py-pyopengl/package.py
+++ b/var/spack/repos/builtin/packages/py-pyopengl/package.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Copyright 2013-2024 Lawrence Livermore National Security, LLC and other
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
--- a/var/spack/repos/builtin/packages/py-pyopengl/package.py
+++ b/var/spack/repos/builtin/packages/py-pyopengl/package.py
@@ -41,7 +41,7 @@ class PyPyopengl(PythonPackage):
         if "+glu" in self.spec:
             lib_dirs.extend(self.spec["glu"].libs.directories)
         if "+glut" in self.spec:
-            lib_dirs.append(self.spec["freeglut"].prefix.lib)
+            lib_dirs.extend(self.spec["freeglut"].libs.directories)
         libs = ":".join(lib_dirs)
         if sys.platform == "darwin":
             env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", libs)
--- a/var/spack/repos/builtin/packages/py-pyopengl/package.py
+++ b/var/spack/repos/builtin/packages/py-pyopengl/package.py
@@ -7,7 +7,7 @@ from spack.package import *
 
 
 class PyPyopengl(PythonPackage):
-    """yOpenGL is the most common cross platform Python binding to OpenGL and related APIs."""
+    """PyOpenGL is the most common cross platform Python binding to OpenGL and related APIs."""
 
     homepage = "http://pyopengl.sourceforge.net"
     url = "https://files.pythonhosted.org/packages/py3/p/pyopengl/PyOpenGL-3.1.6-py3-none-any.whl"
--- a/var/spack/repos/builtin/packages/py-pyopengl/package.py
+++ b/var/spack/repos/builtin/packages/py-pyopengl/package.py
@@ -1,0 +1,22 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyPyopengl(PythonPackage):
+    """yOpenGL is the most common cross platform Python binding to OpenGL and related APIs."""
+
+    homepage = "http://pyopengl.sourceforge.net"
+    url = "https://files.pythonhosted.org/packages/py3/p/pyopengl/PyOpenGL-3.1.6-py3-none-any.whl"
+    list_url = "https://pypi.org/simple/pyopengl/"
+
+    version(
+        "3.1.6",
+        sha256="a7139bc3e15d656feae1f7e3ef68c799941ed43fadc78177a23db7e946c20738",
+        expand=False,
+    )
+
+    depends_on("python@3.6:", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-pyopengl/package.py
+++ b/var/spack/repos/builtin/packages/py-pyopengl/package.py
@@ -24,11 +24,11 @@ class PyPyopengl(PythonPackage):
     depends_on("py-setuptools", type="build")
     # actually installing PyOpenGL itself just requires python
     # but tests (and possibly dependent packages) need OpenGL libraries
-    depends_on("gl", type=("test", "run"))
-    depends_on("glu", when="+glu", type=("test", "run"))
-    depends_on("freeglut+shared", when="+glut", type=("test", "run"))
+    depends_on("gl", type="link")
+    depends_on("glu", when="+glu", type="link")
+    depends_on("freeglut+shared", when="+glut", type="link")
 
-    def setup_build_environment(self, env):
+    def setup_run_environment(self, env):
         # PyOpenGL uses ctypes.cdll (or similar), which searches LD_LIBRARY_PATH
         lib_dirs = self.spec["gl"].libs.directories
         if "^glx" in self.spec:
@@ -47,11 +47,8 @@ class PyPyopengl(PythonPackage):
         else:
             env.prepend_path("LD_LIBRARY_PATH", libs)
 
-    def setup_run_environment(self, env):
-        self.setup_build_environment(env)
-
-    def setup_dependent_build_environment(self, env, dependent_spec):
-        self.setup_build_environment(env)
+    def setup_dependent_build_environment_(self, env, dependent_spec):
+        self.setup_run_environment(env)
 
     # only test import available module
     @property
--- a/var/spack/repos/builtin/packages/py-pyopengl/package.py
+++ b/var/spack/repos/builtin/packages/py-pyopengl/package.py
@@ -12,46 +12,30 @@ class PyPyopengl(PythonPackage):
     """PyOpenGL is the most common cross platform Python binding to OpenGL and related APIs."""
 
     homepage = "http://pyopengl.sourceforge.net"
-    url = "https://files.pythonhosted.org/packages/py3/p/pyopengl/PyOpenGL-3.1.6-py3-none-any.whl"
-    list_url = "https://pypi.org/simple/pyopengl/"
+    pypi = "pyopengl/PyOpenGL-3.1.6.tar.gz"
 
-    version(
-        "3.1.6",
-        sha256="a7139bc3e15d656feae1f7e3ef68c799941ed43fadc78177a23db7e946c20738",
-        expand=False,
-    )
+    version("3.1.6", sha256="8ea6c8773927eda7405bffc6f5bb93be81569a7b05c8cac50cd94e969dce5e27")
 
-    variant(
-        "gl",
-        default="glx" if sys.platform.startswith("linux") else "other",
-        values=("glx", "osmesa", "other"),
-        multi=False,
-        description="The OpenGL provider to use",
-    )
-    variant("glu", default=True)
-    variant("glut", default=True)
+    variant("glu", default=True, description="Enable OpenGL Utility (GLU) binding.")
+    variant("glut", default=True, description="Enable OpenGL Utility Toolkit (GLUT) binding.")
 
-    conflicts("osmesa", when="gl=glx")
-    conflicts("osmesa", when="gl=other")
-    conflicts("glx", when="gl=osmesa")
-    conflicts("glx", when="gl=other")
+    conflicts("osmesa", when="^glx")
+    conflicts("glx", when="^osmesa")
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyPyopengl(PythonPackage):
+    """yOpenGL is the most common cross platform Python binding to OpenGL and related APIs."""
```
adamjstewart:
```suggestion
    """PyOpenGL is the most common cross platform Python binding to OpenGL and related APIs."""
```

adamjstewart (commented):
Does this need a dependency on `gl`? Most of our Python packages build from source but I have been thinking about installing wheels instead, at least for pure-Python libraries like this oen.

bernhardkaindl:
@melven: As adamjsteward pointed out, this is missing dependencies. This fails on an Ubuntu 20.04 server:
```
spack install --test=root py-pyopengl
```
Even with this it still fails:
```
    extends("python@3.6:+tkinter")
    # depends_on("python@3.6:+tkinter", type=("build", "test", "run"))
    depends_on("tk", type=("build", "link", "test", "run"))
    depends_on("glx", type=("build", "link", "test", "run"))
    depends_on("osmesa", type=("build", "link", "test", "run"))
```
With:
```
Failure loading Togl package: can't find package Togl, on debian systems this is provided by `libtogl2`
OpenGL/Tk/__init__.py", line 119, in <module>
     31        _default_root.tk.call('package', 'require', 'Togl')
     32    _tkinter.TclError: can't find package Togl
pythn -c' 'import OpenGL.osmesa'
AttributeError: 'GLXPlatform' object has no attribute 'OSMesa'
_tkinter.TclError: can't find package Togl
```
And: `AttributeError: 'GLXPlatform' object has no attribute 'EGL'`

https://github.com/deepmind/dm_control/issues/129 suggests that PyOpenGL_accelerate needs to be installed as well. PyOpenGL itself would not be HW accelerated it seems.

Setting this to draft until the package is updated by the submitter.

melven:
> @melven: As adamjsteward pointed out, this is missing dependencies. This fails on an Ubuntu 20.04 server:
> 
> ```
> spack install --test=root py-pyopengl
> ```
> [...]
>

I played around with this and it seems that it only works when the required libraries are in the `LD_LIBRARY_PATH` variable.
So I added this to the `modules.yaml`:
```yaml
modules:
  prefix_inspections:
    # [...]
    lib:
      - LD_LIBRARY_PATH
    lib64:
      - LD_LIBRARY_PATH
```
And then used the (dirty) environment from the module:
```console
> module load py-pyopengl
> spack test run --dirty py-pyopengl
==> Spack test uiuahvfbydsexch2r5fdii2rmdzlpkul
==> Testing package py-pyopengl-3.1.6-kx43rfc
==> Error: TestFailure: 51 tests failed.
[...]
```

Without the libraries in the path, I obtain:
```console
> spack test run py-pyopengl
==> Spack test uiuahvfbydsexch2r5fdii2rmdzlpkul
==> Testing package py-pyopengl-3.1.6-kx43rfc
==> Error: TestFailure: 104 tests failed.
[...]
```

That there are still 51 errors is expected in my opinion because it tests all variants of OpenGL from different vendors and standards (EGL, GLES, WGL, ...).

The question here is if it would be helpful to add runtime dependencies and inject appropriate LD_LIBRARY_PATH variables.
Or just keep it as it is and it will detect and provide all available OpenGL variants at runtime. What do you think?

adamjstewart:
> The question here is if it would be helpful to add runtime dependencies and inject appropriate LD_LIBRARY_PATH variables.

That would be my suggestion. We don't normally set `LD_LIBRARY_PATH` because we use RPATH but make exceptions for Python libraries that don't support RPATH. See `py-shapely` for another example of this. Note that you'll want to use `DYLD_FALLBACK_LIBRARY_PATH` on macOS.

melven:
Works now in a fresh minimal Ubuntu 22.04 server. Also allows to choose between glx and osmesa like other opengl-packages.
I also added the desired modules for testing that should be available.

var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,88 @@
...
+import sys
+
+from spack.package import *
+
+
+class PyPyopengl(PythonPackage):
+    """PyOpenGL is the most common cross platform Python binding to OpenGL and related APIs."""
+
+    homepage = "http://pyopengl.sourceforge.net"
+    url = "https://files.pythonhosted.org/packages/py3/p/pyopengl/PyOpenGL-3.1.6-py3-none-any.whl"
```
adamjstewart: Should we build from source to match other packages?

var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,88 @@
...
+
+    variant(
+        "gl",
+        default="glx" if sys.platform.startswith("linux") else "other",
+        values=("glx", "osmesa", "other"),
+        multi=False,
+        description="The OpenGL provider to use",
+    )
+    variant("glu", default=True)
+    variant("glut", default=True)
```
adamjstewart: Need descriptions

var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,88 @@
...
+
+    conflicts("osmesa", when="gl=glx")
+    conflicts("osmesa", when="gl=other")
+    conflicts("glx", when="gl=osmesa")
+    conflicts("glx", when="gl=other")
+
+    depends_on("python@3.6:", type=("build", "run"))
```
adamjstewart: Spack only supports 3.7+ anyway, you can remove this line

var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,88 @@
...
+        if "gl=glx" in self.spec:
+            lib_dirs.extend(self.spec["glx"].libs.directories)
+            env.set("PYOPENGL_PLATFORM", "glx")
+        if "gl=osmesa" in self.spec:
+            lib_dirs.extend(self.spec["osmesa"].libs.directories)
+            env.set("PYOPENGL_PLATFORM", "osmesa")
+        if "+glu" in self.spec:
+            lib_dirs.extend(self.spec["glu"].libs.directories)
+        if "+glut" in self.spec:
+            lib_dirs.append(self.spec["freeglut"].prefix.lib)
```
adamjstewart:
```suggestion
            lib_dirs.append(self.spec["freeglut"].libs.directories[0])
```
If this doesn't find anything, you may need to add a libs property to the package

var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,88 @@
...
+    url = "https://files.pythonhosted.org/packages/py3/p/pyopengl/PyOpenGL-3.1.6-py3-none-any.whl"
+    list_url = "https://pypi.org/simple/pyopengl/"
+
+    version(
+        "3.1.6",
+        sha256="a7139bc3e15d656feae1f7e3ef68c799941ed43fadc78177a23db7e946c20738",
+        expand=False,
+    )
+
+    variant(
```
adamjstewart: I don't think a variant makes sense here. All of these are already possible `gl` providers, and the user can select one like so:
```console
> spack install py-pyopengl ^glx
> spack install py-pyopengl ^osmesa
...
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: This doesn't work for me. I get the following error even though the libraries are there:
```console
==> Error: NoLibrariesError: Unable to recursively locate freeglut libraries in /home/melven/spack/opt/spack/linux-ubuntu22.04-sandybridge/gcc-11.4.0/freeglut-3.2.2-l4y55y4dgi7jk3puwzfbi3uzyheqpbrj

/home/melven/spack/var/spack/repos/builtin/packages/py-pyopengl/package.py:60, in setup_build_environment:
         57        if "+glu" in self.spec:
         58            lib_dirs.extend(self.spec["glu"].libs.directories)
         59        if "+glut" in self.spec:
  >>     60            lib_dirs.append(self.spec["freeglut"].libs.directories)
         61        libs = ":".join(lib_dirs)
         62        if sys.platform == "darwin":
         63            env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", libs)


melven@bunti:~/spack$ spack edit py-pyopengl
melven@bunti:~/spack$ ls /home/melven/spack/opt/spack/linux-ubuntu22.04-sandybridge/gcc-11.4.0/freeglut-3.2.2-l4y55y4dgi7jk3puwzfbi3uzyheqpbrj
include  lib
melven@bunti:~/spack$ ls /home/melven/spack/opt/spack/linux-ubuntu22.04-sandybridge/gcc-11.4.0/freeglut-3.2.2-l4y55y4dgi7jk3puwzfbi3uzyheqpbrj/lib/
cmake  libglut.so  libglut.so.3  libglut.so.3.11.1  pkgconfig
```

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: I just copied it from the mesa-glu package. You cannot use both at once (that's what the `PYOPENGL_PLATFORM` environment variable is doing).

So I need a way to distinguish between osmesa and glx but I don't mind if that's through `"gl=osmesa"` or `^osmesa`...

adamjstewart: Yes, the latter is preferred over a variant.

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: Done.

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: Done.

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: Done.

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: Done.

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: Ok, that's a problem of freeglut which provides a library named libglut but does not set the `libs` property in the package.py.

I added that in  #39300. With this change, also `spec["freeglut"].libs.directories` works correctly.

melven: This MR depends on #39300 now!

melven:
Hi,
I updated the copyright notice and checked the package on an ubuntu 22.04 server.
Builds nicely in all different variants and a minimal glut example pops up a window with a red cube over ssh.

We have been using this for two years (?) now and it works nicely.

Please re-review if that's fine for you!

var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,72 @@
...
+    homepage = "http://pyopengl.sourceforge.net"
+    pypi = "pyopengl/PyOpenGL-3.1.6.tar.gz"
+
+    version("3.1.6", sha256="8ea6c8773927eda7405bffc6f5bb93be81569a7b05c8cac50cd94e969dce5e27")
+
+    variant("glu", default=True, description="Enable OpenGL Utility (GLU) binding.")
+    variant("glut", default=True, description="Enable OpenGL Utility Toolkit (GLUT) binding.")
+
+    conflicts("osmesa", when="^glx")
+    conflicts("glx", when="^osmesa")
```
adamjstewart: These _should_ be symmetric, so I don't think you actually need both lines.

var/spack/repos/builtin/packages/py-pyopengl/package.py
```diff
@@ -0,0 +1,72 @@
...
+
+    depends_on("py-setuptools", type="build")
+    # actually installing PyOpenGL itself just requires python
+    # but tests (and possibly dependent packages) need OpenGL libraries
+    depends_on("gl", type=("test", "run"))
+    depends_on("glu", when="+glu", type=("test", "run"))
+    depends_on("freeglut+shared", when="+glut", type=("test", "run"))
```
adamjstewart: Based on `setup_build_environment`, it looks like these should be `type="link"`, not test/run. Does the package every search `PATH` at build/run/test time, or does it only ever look in `LD_LIBRARY_PATH` to link to the library?

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: Done

var/spack/repos/builtin/packages/py-pyopengl/package.py
melven: `setup_build_environment` is actually not needed, it doesn't check the gl-libraries in the python setup/install script.

So I set the dependencies to "link" only and removed the `setup_build_environment`.

Does it make sense this way?

adamjstewart: Yep, makes more sense now.

adamjstewart (approved):
Looks good to me as long as it builds for you.

PR #32358, opened 2022-08-24
snap-berkeley: update to version 2.0.1

snehring:
I think most of the compiler conflicts in this were due to it hard coding g++ in the makefile, which I've just gone ahead and removed. I was able to test with clang without issue.

I'll leave the gcc conflict for the older versions, since I'm guessing it was causing build errors.


tldahlgren (approved):
Confirmed sha256.

PR #32360, opened 2022-08-24
Bug fix for esmf@8.3.0b09 with PIO

climbfuji:
ESMF 8.3.0 beta snapshot 9 (`esmf@8.3.0b09`) uses internal Parallel IO code, similar to releases 8.2.0 and earlier. The behavior changed for more recent beta snapshots of ESMF 8.3.0 and for the official 8.3.0 release.

This PR fixes a bug in the current ESMF package so that the Parallel IO logic in `esmf@8.3.0b09` is handled like in versions 8.2.0 and earlier. It also adds @jedwards4b to the list of maintainers of the package.


jedwards4b (approved):

tldahlgren (approved):
LGTM

PR #32361, opened 2022-08-24
Parallel testing

trws:
This PR leverages the tmpconfig script from #32183 and a number of fixes and test enhancements to enable parallel testing with pytest-xdist. On a 32-core zen workstation, a 16-way parallel run of the tests can finish in under four minutes, where a sequential run is more than 30 minutes.  The remaining question here is how to expose parallelism in the interface, and if we want to make xdist part of the bootstrap set or not.

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -11,7 +11,7 @@ on:
         type: string
 
 concurrency:
-  group: audit-${{inputs.python_version}}-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  group: audit-${{inputs.python_version}}-${{github.ref}}-${{github.run_number}}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -1,0 +1,44 @@
+name: audit
+
+on:
+  workflow_call:
+    inputs:
+      with_coverage:
+        required: true
+        type: string
+      python_version:
+        required: true
+        type: string
+
+concurrency:
+  group: audit-${{inputs.python_version}}-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  cancel-in-progress: true
+
+jobs:
+  # Run audits on all the packages in the built-in repository
+  package-audits:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # @v2
+    - uses: actions/setup-python@b55428b1882923874294fa556849718a1d7f2ca5 # @v2
+      with:
+        python-version: ${{inputs.python_version}}
+    - name: Install Python packages
+      run: |
+        pip install --upgrade pip six setuptools pytest codecov 'coverage[toml]<=6.2'
+    - name: Package audits (with coverage)
+      if: ${{ inputs.with_coverage == 'true' }}
+      run: |
+          . share/spack/setup-env.sh
+          coverage run $(which spack) audit packages
+          coverage combine
+          coverage xml
+    - name: Package audits (without coverage)
+      if: ${{ inputs.with_coverage == 'false' }}
+      run: |
+          . share/spack/setup-env.sh
+          $(which spack) audit packages
+    - uses: codecov/codecov-action@81cd2dc8148241f03f5839d295e000b8f761e378 # @v2.1.0
+      if: ${{ inputs.with_coverage == 'true' }}
+      with:
+        flags: unittests,linux,audits
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -11,7 +11,7 @@ on:
         type: string
 
 concurrency:
-  group: audit-${{inputs.python_version}}-${{github.ref}}-${{github.run_number}}
+  group: audit-${{inputs.python_version}}-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -9,7 +9,7 @@ on:
     - cron: '16 2 * * *'
 
 concurrency:
-  group: bootstrap-${{github.ref}}-${{github.run_number}}
+  group: bootstrap-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -9,7 +9,7 @@ on:
     - cron: '16 2 * * *'
 
 concurrency:
-  group: bootstrap-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  group: bootstrap-${{github.ref}}-${{github.run_number}}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -20,7 +20,7 @@ on:
     types: [published]
 
 concurrency:
-  group: build_containers--${{github.ref}}-${{github.run_number}}
+  group: build_containers--${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -20,7 +20,7 @@ on:
     types: [published]
 
 concurrency:
-  group: build_containers--${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
+  group: build_containers-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -20,7 +20,7 @@ on:
     types: [published]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  group: build_containers--${{github.ref}}-${{github.run_number}}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,6 +53,7 @@ jobs:
             - 'share/spack/**'
             - '.github/workflows/bootstrap.yml'
             core:
+            - '!lib/spack/docs/**'
             - './!(var/**)/**'
             packages:
             - 'var/**'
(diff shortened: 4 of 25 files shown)


lib/spack/spack/test/cmd_extensions.py
```diff
@@ -228,14 +230,17 @@ def test_missing_command():
...
     """Ensure that we are correctly validating configured extension paths
     for conformity with the rules: the basename should match
     ``spack-<name>``; <name> may have embedded hyphens but not begin with one.
     """
-    with spack.config.override("config:extensions", [extension_path]):
-        with pytest.raises(expected_exception):
-            spack.cmd.get_module("no-such-command")
+    # NOTE: if the directory is a valid extension directory name the "vacuous" test will
+    # fail because it resolves to current working directory
+    with working_dir(str(tmpdir)):
```
tgamblin: nitpick: this could be `with tmpdir.as_cwd()`

tgamblin (dismissed):

lib/spack/spack/test/cmd_extensions.py
trws: Ah yeah, that's a good point.  I'll fix this, then add what I think we need for the concurrency disambiguation.

.github/workflows/audit.yaml
```diff
@@ -11,7 +11,7 @@ on:
         type: string
 
 concurrency:
-  group: audit-${{inputs.python_version}}-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  group: audit-${{inputs.python_version}}-${{github.ref}}-${{github.run_number}}
```
alalazo: From the docs [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency):

> A concurrency group can be any string or expression. The expression can only use the [github context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context).

Do we have any guarantee when using `inputs`?

.github/workflows/build-containers.yml
```diff
@@ -20,7 +20,7 @@ on:
     types: [published]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  group: build_containers--${{github.ref}}-${{github.run_number}}
```
alalazo: The name before was designed to:
1. Avoid skipping commits on `develop`, i.e. if we merged multiple PRs in a short amount of time each commit would be tested
2. Cancel runs on PRs if a newer commit is being pushed while a workflow is running

It seems to me that always using `run_number` we are losing 2, or am I missing something?

.github/workflows/audit.yaml
trws: Thanks for that, I'll check, but the run_number should disambiguate it globally even if the python_version doesn't work.

.github/workflows/build-containers.yml
trws: Crud, you are probably right, I can put the conditional pull_request number back but keep the ref and prefixes.

trws:
I triggered a re-run of everything with debug logging so we can inspect what inputs and names and everything got computed for this.

trws:
@alalazo, I think this version will take care of the concurrency groups

@tgamblin, the issue with checksum seems to be not the SpackCommand or the logger, but the fact that `mock_fetch` only mocks the fetcher attribute of the PackageBase class, so it's ignored by Stage, and so the unit tests for checksum try to download releases from github.  When they time out or fail, the command errors out and fails the tests even when the result is correct.  This *seems* to work around the issue locally, but it's still pretty nasty, ideally we should figure out how to get that mock to work, or otherwise mock out fetching through stage for these tests.

.github/workflows/audit.yaml
```diff
@@ -0,0 +1,44 @@
...
+          coverage xml
+    - name: Package audits (without coverage)
+      if: ${{ inputs.with_coverage == 'false' }}
+      run: |
+          . share/spack/setup-env.sh
+          $(which spack) audit packages
+    - uses: codecov/codecov-action@81cd2dc8148241f03f5839d295e000b8f761e378 # @v2.1.0
+      if: ${{ inputs.with_coverage == 'true' }}
+      with:
+        flags: unittests,linux,audits
```
alalazo, line 44: We can possibly remove `unittests` in a following PR.

.github/workflows/build-containers.yml
```diff
@@ -20,7 +20,7 @@ on:
     types: [published]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  group: build_containers--${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
```
alalazo: (typo) Can be done in a following PR too:
```suggestion
  group: build_containers-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
```

.github/workflows/bootstrap.yml
```diff
@@ -9,7 +9,7 @@ on:
     - cron: '16 2 * * *'
 
 concurrency:
-  group: bootstrap-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_number }}
+  group: bootstrap-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}
```
alalazo, line 12: We need to write somewhere (or spread the knowledge) that `${{ github.ref }}` is needed to disambiguate in case `${{ github.event.pull_request_number }}` is the same as `${{ github.run_number }}`.

.github/workflows/ci.yaml
```diff
@@ -79,15 +92,21 @@ jobs:
...
-    if: ${{ github.repository == 'spack/spack' }}
+    if: ${{ github.repository == 'spack/spack' && needs.changes.outputs.core == 'true' }}
     needs: [ prechecks ]
     uses: ./.github/workflows/windows_python.yml
+  all:
+    needs: [ windows, unit-tests, bootstrap, audit-ancient-python ]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Success
+      run: "true"
```
alalazo, line 111: Are we planning on making this the only node required?

.github/workflows/unit_tests.yaml
```diff
@@ -25,11 +26,26 @@ jobs:
       matrix:
         python-version: ['2.7', '3.6', '3.7', '3.8', '3.9', '3.10']
         concretizer: ['clingo']
+        on_develop:
+        - ${{ github.ref == 'refs/heads/develop' }}
```
alalazo, line 30: Can be solved afterwards, but I think this means we'll treat releases in a different way than `develop`. Also, not clear to me where this is currently used.

lib/spack/llnl/util/tty/pty.py
```diff
@@ -228,8 +228,8 @@ def __init__(self, controller_function, minion_function):
         self.minion_function = minion_function
 
         # these can be optionally set to change defaults
-        self.controller_timeout = 1
-        self.sleep_time = 0
+        self.controller_timeout = 3
+        self.sleep_time = 0.1
```
alalazo, line 232: Why was this change necessary?

lib/spack/spack/main.py
```diff
@@ -623,15 +624,18 @@ class SpackCommand(object):
     their output.
     """
 
-    def __init__(self, command_name):
+    def __init__(self, command_name, subprocess=False):
         """Create a new SpackCommand that invokes ``command_name`` when called.
 
         Args:
             command_name (str): name of the command to invoke
+            subprocess (bool): whether to fork a subprocess or not
```
alalazo:
```suggestion
            subprocess (bool): whether to fork a subprocess or not. Currently not supported on 
                Windows, where it is always False.
```

alalazo (commented):
Overall LGTM, left a few comments that can be addressed in following PRs.

.github/workflows/audit.yaml
trws: Possibly, it’s still handy as a matcher.

.github/workflows/bootstrap.yml
trws: It would be good to document both that *and* the prefixes. They’re only necessary like this because of using reusable workflows forcing GitHub.workflow to the same value for all of them.

.github/workflows/ci.yaml
trws: Possibly. It’s there to give us that option but we need to experiment with how the requirements and check outputs appear.

.github/workflows/unit_tests.yaml
trws: Ah, we could add an || here to pick up the releases too. It’s used in unit_tests as part of the Ubuntu matrix so we can exclude 3.7-3.9 on PRs.

lib/spack/llnl/util/tty/pty.py
trws: The tty oriented tests, mainly foreground/background handling, were reliably failing under xdist on GitHub actions. That change relaxed the timing enough to let them succeed again, it was @tgamblin’s suggestion.

alalazo:
@trws I would merge this one asap, and take care of minor things in following PRs. What do you think? Do we need to coordinate with other PRs for pipelines?

trws:
I made sure the pipeline trigger names remain unchanged in this one. We can upgrade to better targets at any time after this merges. If you’re good, merge anytime.


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On September 7, 2022 at 8:05:39 AM PDT, Massimiliano Culpo ***@***.***> wrote:

@trws<https://urldefense.us/v3/__https://github.com/trws__;!!G2kpM7uM-TzIFchu!gfr5Mcvu_m_f8Yrqf-BD_HXIE-Egpt7lg6K1NTVpHUAXYx-9GMUFPt2rS8ITIUbCJQ$> I would merge this one asap, and take care of minor things in following PRs. What do you think? Do we need to coordinate with other PRs for pipelines?

alalazo:
@spackbot fix style

alalazo (approved):

.github/workflows/unit_tests.yaml
alalazo: Oh, got it. It's kind of confusing that we always include `original` with `on_develop: false`, but we can fix that later. I'm kind of eager to have this PR in `develop` and see unit tests run again there :slightly_smiling_face:

trws:
Ubuntu 2.7 got stuck, so that's re-running. Once that comes back green, if we're agreed it would be time to update the requirements list and merge this in.  @alalazo, @tgamblin

.github/workflows/unit_tests.yaml
trws: That's true, I should have thought to use the expression there.

alalazo:
Yeah, I'm keeping an eye on this. As soon as it gets :heavy_check_mark: I'll change requirements to `all` and merge it.

.github/workflows/unit_tests.yaml
alalazo: We'll fix it in another PR.

alalazo:
Thanks @trws this is much faster than before and seems to have fixed the issues on develop

PR #32364, opened 2022-08-25
new packages: r-cmdstanr and cmdstan

glennpj:


tldahlgren (approved):
Confirmed sha256 for all new versions/packages.  Also confirmed package not on bioconductor site.

PR #32369, opened 2022-08-25
Allow requirements for virtual packages

alalazo:
Extend the semantic of package requirements to allow using them also under a virtual package attribute in `packages.yaml`. These requirements are enforced whenever that virtual spec is present in the DAG.

Modifications:
- [x] Extend the semantic of requirements to virtual specs
- [x] Add unit tests to verify behavior
- [x] Add documentation

### DIFF
--- a/lib/spack/spack/solver/asp.py
+++ b/lib/spack/spack/solver/asp.py
@@ -946,7 +946,7 @@ class SpackSolverSetup(object):
             "require", []
         )
         rules = self._rules_from_requirements(pkg_name, requirements)
-        self.emit_facts_from_rules(rules, virtual=False)
+        self.emit_facts_from_requirement_rules(rules, virtual=False)
 
     def _rules_from_requirements(self, pkg_name, requirements):
         """Manipulate requirements from packages.yaml, and return a list of tuples
@@ -1171,11 +1171,10 @@ class SpackSolverSetup(object):
         for virtual_str in sorted(self.possible_virtuals):
             requirements = packages_yaml.get(virtual_str, {}).get("require", [])
             rules = self._rules_from_requirements(virtual_str, requirements)
-            self.emit_facts_from_rules(rules, virtual=True)
+            self.emit_facts_from_requirement_rules(rules, virtual=True)
 
-        self.gen.newline()
-
-    def emit_facts_from_rules(self, rules, virtual=False):
+    def emit_facts_from_requirement_rules(self, rules, virtual=False):
+        """Generate facts to enforce requirements from packages.yaml."""
         for requirement_grp_id, (pkg_name, policy, requirement_grp) in enumerate(rules):
             self.gen.fact(fn.requirement_group(pkg_name, requirement_grp_id))
             self.gen.fact(fn.requirement_policy(pkg_name, requirement_grp_id, policy))
@@ -1183,11 +1182,11 @@ class SpackSolverSetup(object):
                 spec = spack.spec.Spec(spec_str)
                 if not spec.name:
                     spec.name = pkg_name
-                required_spec = spec
+                when_spec = spec
                 if virtual:
-                    required_spec = spack.spec.Spec(pkg_name)
+                    when_spec = spack.spec.Spec(pkg_name)
                 member_id = self.condition(
-                    required_spec, imposed_spec=spec, name=pkg_name, node=True
+                    required_spec=when_spec, imposed_spec=spec, name=pkg_name, node=True
                 )
                 self.gen.fact(fn.requirement_group_member(member_id, pkg_name, requirement_grp_id))
                 self.gen.fact(fn.requirement_has_weight(member_id, requirement_weight))
--- a/lib/spack/spack/solver/asp.py
+++ b/lib/spack/spack/solver/asp.py
@@ -1186,7 +1186,7 @@ class SpackSolverSetup(object):
                 if virtual:
                     when_spec = spack.spec.Spec(pkg_name)
                 member_id = self.condition(
-                    required_spec=when_spec, imposed_spec=spec, name=pkg_name, node=True
+                    required_spec=when_spec, imposed_spec=spec, name=pkg_name, node=virtual
                 )
                 self.gen.fact(fn.requirement_group_member(member_id, pkg_name, requirement_grp_id))
                 self.gen.fact(fn.requirement_has_weight(member_id, requirement_weight))
--- a/lib/spack/spack/solver/asp.py
+++ b/lib/spack/spack/solver/asp.py
@@ -945,6 +945,13 @@ class SpackSolverSetup(object):
         requirements = config.get(pkg_name, {}).get("require", []) or config.get("all", {}).get(
             "require", []
         )
+        rules = self._rules_from_requirements(pkg_name, requirements)
+        self.emit_facts_from_rules(rules, virtual=False)
+
+    def _rules_from_requirements(self, pkg_name, requirements):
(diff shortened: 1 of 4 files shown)


lib/spack/spack/solver/asp.py
```diff
@@ -1161,6 +1160,38 @@ def provider_defaults(self):
...
+            for requirement_weight, spec_str in enumerate(requirement_grp):
+                spec = spack.spec.Spec(spec_str)
+                if not spec.name:
+                    spec.name = pkg_name
+                required_spec = spec
+                if virtual:
+                    required_spec = spack.spec.Spec(pkg_name)
+                member_id = self.condition(
+                    required_spec, imposed_spec=spec, name=pkg_name, node=True
+                )
```
scheibelp, line 1190: (question) This appears to be a slight change in behavior: originally for the non-virtual rules, we called condition without `node=True`. I assume it's needed for virtuals. The tests pass so it's not a problem, but would you be able to explain why (that's mainly for my benefit)?

lib/spack/spack/solver/asp.py
```diff
@@ -1161,6 +1160,38 @@ def provider_defaults(self):
...
+        packages_yaml = spack.config.config.get("packages")
+        assert self.possible_virtuals is not None, msg
+        for virtual_str in sorted(self.possible_virtuals):
+            requirements = packages_yaml.get(virtual_str, {}).get("require", [])
+            rules = self._rules_from_requirements(virtual_str, requirements)
+            self.emit_facts_from_rules(rules, virtual=True)
+
+        self.gen.newline()
+
+    def emit_facts_from_rules(self, rules, virtual=False):
```
scheibelp: I think `emit_facts_from_requirement_rules` would be more clear. A short docstring like 

```
Generate facts to enforce requirements from packages.yaml
```

would also be helpful

lib/spack/spack/solver/asp.py
```diff
@@ -1161,6 +1160,38 @@ def provider_defaults(self):
...
+            " maintainers"
+        )
+        packages_yaml = spack.config.config.get("packages")
+        assert self.possible_virtuals is not None, msg
+        for virtual_str in sorted(self.possible_virtuals):
+            requirements = packages_yaml.get(virtual_str, {}).get("require", [])
+            rules = self._rules_from_requirements(virtual_str, requirements)
+            self.emit_facts_from_rules(rules, virtual=True)
+
+        self.gen.newline()
```
scheibelp: (minor) I'm curious why this is done here but not in `package_requirement_rules`.

lib/spack/spack/solver/asp.py
```diff
@@ -1161,6 +1160,38 @@ def provider_defaults(self):
...
+            self.gen.fact(fn.requirement_policy(pkg_name, requirement_grp_id, policy))
+            for requirement_weight, spec_str in enumerate(requirement_grp):
+                spec = spack.spec.Spec(spec_str)
+                if not spec.name:
+                    spec.name = pkg_name
+                required_spec = spec
+                if virtual:
+                    required_spec = spack.spec.Spec(pkg_name)
+                member_id = self.condition(
+                    required_spec, imposed_spec=spec, name=pkg_name, node=True
```
scheibelp: (minor) Can you change the name `required_spec` to `when_spec` or `match_spec`? I realize that the method signature of `condition()` uses these names, but I had to go to the method to remind myself of the intended difference between them (perhaps this should be handled in a separate refactor PR, assuming you agree).

(the confusion I'm trying to avoid is that right now the configuration has a `requires` spec, which here is converted to `imposed_spec` while `require_spec` is used for something else)

lib/spack/spack/test/concretize_requirements.py
```diff
@@ -349,3 +349,51 @@ def test_default_and_package_specific_requirements(
...
+""".format(
+        mpi_requirement
+    )
+    update_packages_config(conf_str)
+
+    spec = Spec("callpath").concretized()
+    assert "mpi" in spec
+    assert mpi_requirement in spec
+
+
```
scheibelp, line 371: (question) If the users asks for `hdf5 ^opempi` but we have

```
packages:
  mpi:
    require:
      mpich
```

will that work? Is it worth testing that this fails if it should be prevented?

scheibelp (changes requested):
I have a few questions and a couple of minor comments about docs/renaming.

lib/spack/spack/solver/asp.py
alalazo: Just a refactoring mismatch, that newline was copied from somewhere before I unified similar code. I'll make the two functions match.

lib/spack/spack/solver/asp.py
alalazo: Up to now, we've been imposing constraint on the same spec from which we were inspecting requirements. For instance:
```yaml
packages:
  hdf5:
    require: "hdf5+mpi"
```
would have been translated into the following facts, before this change:
```
requirement_group("hdf5",0).
requirement_policy("hdf5",0,"one_of").
condition(1126,"None").
condition_requirement(1126,"node","hdf5").
condition_requirement(1126,"variant_value","hdf5","mpi","True").
imposed_constraint(1126,"variant_set","hdf5","mpi","True").
requirement_group_member(1126,"hdf5",0).
requirement_has_weight(1126,0).
```
Note that in the imposed constraint we don't have any:
```
imposed_constraint(1126,"node","hdf5").
```
which in this case would be redundant. For the case of a virtual spec:
```yaml
packages:
  mpi:
    require: mpich
```
that would result in the following facts:
```
requirement_group("mpi",0).
requirement_policy("mpi",0,"one_of").
condition(0,"None").
condition_requirement(0,"virtual_node","mpi").
requirement_group_member(0,"mpi",0).
requirement_has_weight(0,0).
```
where there is no imposed contraint at all. We thus need to emit a:
```
imposed_constraint(0,"node","mpich").
```
to be sure we apply the requirement.

lib/spack/spack/test/concretize_requirements.py
alalazo: Done in 52d6620

lib/spack/spack/solver/asp.py
alalazo: Done in 7f40293

lib/spack/spack/solver/asp.py
alalazo: Done in 7f40293

lib/spack/spack/solver/asp.py
alalazo: Done in 7f40293

alalazo:
@scheibelp Ready for a second review

lib/spack/spack/solver/asp.py
scheibelp: (EDIT: actual change is highlighted in comment underneath the code)

Would it make sense to change to

```
member_id = self.condition(
    required_spec=when_spec, imposed_spec=spec, name=pkg_name, node=virtual
)
# "node=virtual" instead of "node=True"
```

? In that case, we avoid generating the redundant constraint for non-virtuals (it seems to me that could be inferred automatically and that - at least in this case - the `node` argument is redundant but even if that's true, it's probably best to deal with that in another PR).

alalazo: For non-virtuals it's just a redundant fact.

scheibelp (approved):

scheibelp:
Thanks!

PR #32370, opened 2022-08-25
Add patch for missing template instantiation in pika 0.7.0 on macOS

msimberg:
Backports the patch in https://github.com/pika-org/pika/pull/405 to pika 0.7.0.

### DIFF
--- a/var/spack/repos/builtin/packages/pika/package.py
+++ b/var/spack/repos/builtin/packages/pika/package.py
@@ -104,6 +104,13 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
     patch("transform_mpi_includes.patch", when="@0.3.0 +mpi")
     patch("mimalloc_no_version_requirement.patch", when="@:0.5 malloc=mimalloc")
 
+    # Fix missing template instantiation on macOS
+    patch(
+        "https://github.com/pika-org/pika/commit/dd1dfb85781ec2e76fa37ce7311323e69fbe42a1.patch?full_index=1",
+        sha256="2944f746f5ae4385aba11b7c4a2f991abc108b08ea3dc394bf61c20fc7a2c4f2",
+        when="@0.7.0 platform=darwin",
+    )
+
     def cmake_args(self):
         spec, args = self.spec, []
 


var/spack/repos/builtin/packages/pika/package.py
```diff
@@ -104,6 +104,12 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
     patch("transform_mpi_includes.patch", when="@0.3.0 +mpi")
     patch("mimalloc_no_version_requirement.patch", when="@:0.5 malloc=mimalloc")
 
+    # Fix missing template instantiation on macOS
+    patch(
+        "https://github.com/pika-org/pika/commit/dd1dfb85781ec2e76fa37ce7311323e69fbe42a1.patch?full_index=1",
+        when="0.7 platform=darwin",
```
albestro: Probably pika does not plan to have version 0.7.1, but since it has not been released yet, I think this should apply just to `0.7.0`, otherwise `0.7` will match all `0.7.*`

msimberg: Yeah, fair point. I'll change that.

msimberg: Pushed.

I was also missing `@` from the version constraint...

var/spack/repos/builtin/packages/pika/package.py
```diff
@@ -104,6 +104,12 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
     patch("transform_mpi_includes.patch", when="@0.3.0 +mpi")
     patch("mimalloc_no_version_requirement.patch", when="@:0.5 malloc=mimalloc")
 
+    # Fix missing template instantiation on macOS
+    patch(
+        "https://github.com/pika-org/pika/commit/dd1dfb85781ec2e76fa37ce7311323e69fbe42a1.patch?full_index=1",
+        when="@0.7.0 platform=darwin",
```
albestro, line 111: I was testing it and I learnt that patches from URLs require also checksums.

You should add `sha256=2944f746f5ae4385aba11b7c4a2f991abc108b08ea3dc394bf61c20fc7a2c4f2` as parameter of the patch function.

albestro (dismissed):
See the required change about the checksum, for the rest it looks ok.

Thanks!

var/spack/repos/builtin/packages/pika/package.py
msimberg: Ah, I was also not aware. Thanks for checking! Does it look better now?

albestro: It looks good!

albestro (approved):

teonnik (approved):

tldahlgren (approved):
Confirmed patch exists.

tldahlgren:
There are some pod-related issues for a couple of stacks.

@spackbot run pipeline

PR #32378, opened 2022-08-25
charliecloud: add versions 0.28 and 0.29

rbberger:


tldahlgren (dismissed):
Confirmed both new sha256.

tldahlgren:
Will defer merge decision until at least one maintainer has a chance to comment.

var/spack/repos/builtin/packages/charliecloud/package.py
```diff
@@ -17,6 +17,8 @@ class Charliecloud(AutotoolsPackage):
     tags = ["e4s"]
 
     version("master", branch="master")
+    version("0.29", sha256="c89562e9dce4c10027434ad52eaca2140e2ba8667aa1ec9eadf789b4d7c1a6db")
```
j-ogas, line 20: One thing to note is that 0.29 introduces a git build cache that requires git 2.28.1 or greater.

We should add something like:

```
depends_on("git@2.28.1:", type="run", when="@0.29:")
```

j-ogas (changes requested):
See 0.29 git dependency. Aside from that, thanks!

var/spack/repos/builtin/packages/charliecloud/package.py
rbberger: done.

j-ogas (approved):
Looks good to me.

var/spack/repos/builtin/packages/charliecloud/package.py
reidpr: It doesn't need to be 2.28.1 exactly, and in fact if there's a choice, the newest available Git is likely the best one.

reidpr (commented):
It doesn't need to be Git 2.28.1 exactly, and in fact if there's a choice, the newest available Git is likely the best one.

j-ogas:
> It doesn't need to be Git 2.28.1 exactly, and in fact if there's a choice, the newest available Git is likely the best one.

I think the `git@2.28.1:` syntax, specifically the colon, sets 2.28.1 as the lower limit. That said, I'm not sure if it sets 2.28.1 as the default if not specified.

rbberger:
@reidpr `2.28.1:` means 2.28.1 **or newer**.

tldahlgren (approved):
LGTM

PR #32382, opened 2022-08-25
Improve error message for yaml config file

white238:
I had an ill-formed configuration file.  The error message was originally this:

```
==> Error: Error parsing yamlNone: mapping values are not allowed here
```

This changes it to tell you what file is wrong:

```
==> Error: Error parsing yaml: near /spack/etc/spack/defaults/packages.yaml, 39, 13: mapping values are not allowed here
```

### DIFF
--- a/lib/spack/spack/config.py
+++ b/lib/spack/spack/config.py
@@ -1020,9 +1020,13 @@ def read_config_file(filename, schema=None):
         raise ConfigFileError("Config file is empty or is not a valid YAML dict: %s" % filename)
 
     except MarkedYAMLError as e:
-        msg = "Error parsing yaml: %s" % (filename)
-        if e.context_mark:
-            msg += ": %s" % str(e.context_mark)
+        msg = "Error parsing yaml"
+        mark = e.context_mark if e.context_mark else e.problem_mark
+        if mark:
+            line, column = mark.line, mark.column
+            msg += ": near %s, %s, %s" % (mark.name, str(line), str(column))
+        else:
+            msg += ": %s" % (filename)
         msg += ": %s" % (e.problem)
         raise ConfigFileError(msg)
 
--- a/lib/spack/spack/config.py
+++ b/lib/spack/spack/config.py
@@ -1020,7 +1020,11 @@ def read_config_file(filename, schema=None):
         raise ConfigFileError("Config file is empty or is not a valid YAML dict: %s" % filename)
 
     except MarkedYAMLError as e:
-        raise ConfigFileError("Error parsing yaml%s: %s" % (str(e.context_mark), e.problem))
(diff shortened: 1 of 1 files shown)


tldahlgren (dismissed):
LGTM.  (Tested manually.)

lib/spack/spack/config.py
```diff
@@ -1020,7 +1020,11 @@ def read_config_file(filename, schema=None):
         raise ConfigFileError("Config file is empty or is not a valid YAML dict: %s" % filename)
 
     except MarkedYAMLError as e:
-        raise ConfigFileError("Error parsing yaml%s: %s" % (str(e.context_mark), e.problem))
+        msg = "Error parsing yaml: %s" % (filename)
+        if e.context_mark:
+            msg += ": %s" % str(e.context_mark)
+        msg += ": %s" % (e.problem)
```
tldahlgren: Now that I've experimented with this a little, I'm curious if you thought about providing the error string (since it appears to provide a bit more information) or checking `e.problem_mark`)?  I do notice the resulting line and column from the string may be misleading.

For example, the following for a problem with `packages.yaml` missing a space at line 438, column 15:
```suggestion
        msg = "Error parsing yaml"
        mark = e.context_mark if e.context_mark else e.problem_mark
        if mark:
            line, column = mark.line, mark.column
            msg += ": near %s, %s, %s" % (mark.name, str(line), str(column))
        else:
            msg += ": %s" % (filename)
        msg += ": %s" % (e.problem)
```
results in:
```
spack.config.ConfigFileError: Error parsing yaml: near $HOME/.spack/packages.yaml, 438, 13: mapping values are not allowed here
```

On the other hand, the following:
```suggestion
        msg = "Error parsing yaml: %s" % (str(e).replace("\n", ": "))
```
yields 
```
spack.config.ConfigFileError: Error parsing yaml: mapping values are not allowed here:   in "$HOME/.spack/packages.yaml", line 439, column 14
```
for the same issue.

white238: Those are a definite improvement. I did the bare minimum to make the message usable but any further hints would be greatly appreciated.

~~Do you know if `e.problem_mark` would ever be None like `e.context_mark` was for me?~~

Edit: Nevermind I see the if checking that... i can read i swear.

tldahlgren:
Doesn't look like that exception was covered before this PR.

Do you want me to merge this PR as-is?

lib/spack/spack/config.py
white238: I verified option 1 works for me and is an improvement.  It was off by 1 line in my case but would have helped me greatly in narrowing it down.

I'll commit your suggestion now.

Thanks @tldahlgren !

tldahlgren (approved):
LGTM.

PR #32390, opened 2022-08-26
udunits: Update download URL

RemiLacroix-IDRIS:
It seems like they are only providing the latest version of each branch which means that only version 2.2.28 is actually available. Not sure what I should do in this case?


tldahlgren:
> It seems like they are only providing the latest version of each branch which means that only version 2.2.28 is actually available. Not sure what I should do in this case?

It looks like there are a *lot* newer versions listed at https://artifacts.unidata.ucar.edu/#browse/search=keyword%3Dudunits.

As for what to do with the older versions, my understanding is they should be deprecated (https://spack.readthedocs.io/en/latest/packaging_guide.html#deprecating-old-versions, # 5) so IF any in the build cache can be used.

var/spack/repos/builtin/packages/udunits/package.py
```diff
@@ -10,7 +10,7 @@ class Udunits(AutotoolsPackage):
     """Automated units conversion"""
 
     homepage = "https://www.unidata.ucar.edu/software/udunits"
-    url = "https://artifacts.unidata.ucar.edu/repository/downloads-udunits/udunits-2.2.28.tar.gz"
+    url = "https://artifacts.unidata.ucar.edu/repository/downloads-udunits/2.2.28/udunits-2.2.28.tar.gz"
 
     version("2.2.28", sha256="590baec83161a3fd62c00efa66f6113cec8a7c461e3f61a5182167e0cc5d579e")
```
tldahlgren, line 18: Confirmed this sha256 still applies with the new URL and note all of the older appear to no longer be available.

RemiLacroix-IDRIS:
> It looks like there are a _lot_ newer versions listed at https://artifacts.unidata.ucar.edu/#browse/search=keyword%3Dudunits.

It feels like that's NetCDF versions, it might show there because it bundles udunits? The Git repository shows 2.2.28 as the latest one: https://github.com/Unidata/UDUNITS-2/commits/master.

> As for what to do with the older versions, my understanding is they should be deprecated (https://spack.readthedocs.io/en/latest/packaging_guide.html#deprecating-old-versions, # 5) so IF any in the build cache can be used.

All checksumed version are in the build cache. I have pushed a new commit, let me know if it's OK.

tldahlgren (dismissed):
LGTM

tldahlgren:
@spckbot fix style

tldahlgren:
@spackbot fix style

RemiLacroix-IDRIS:
@tldahlgren : Is there anything blocking this PR?

tldahlgren:
> @tldahlgren : Is there anything blocking this PR?

The required checks.  I'll update the branch to see if we can get that resolved.

tldahlgren (approved):

PR #32391, opened 2022-08-26
Windows: cmake package only depends on Ninja at build time

KiruyaMomochi:
Depends on Ninja at `run` will make `spack load` fails due to `module.make_jobs` does not exist. 

Question: should it has `link` dependency type too?


alalazo:
Already fixed in `develop`, thanks anyway!

PR #32393, opened 2022-08-26
openmpi: handle externals from system directories

skosukhin:
If `slurm` is an external with the prefix `/usr`, the compiler wrappers of the library append `-L/usr/lib[64] -Wl,-rpath -Wl,/usr/lib[64]`. This happens because the configure script is called with `--with-pmi=/usr`. I cannot come up with a good example of how this can break things, it's just my experience (I admit that I might have to change my mind on this, see below) shows that having linker flags for the system prefixes is generally a bad idea. Especially when we are not using Spack compiler wrapper, which de-prioritises linker flags for the system directories. It looks like the developers of the library, at least partially, share this point of view (see https://github.com/open-mpi/ompi/commit/b86e0f04bfca1d6c14177e00544f5e0a500ee02c).

This PR makes sure that prefixes to the system directories are not passed to the configure script. In particular, it is called with `--with-pmi=yes` (which is equal to just `--with-pmi`) instead of `--with-package=/usr`.

Also, @haampie had a short exchange on whether the system directories should be put to `RPATH`s, which you can find in #32212. My opinion is that, as long as Spack [assumes](https://github.com/spack/spack/blob/8d8aa5c6cf294e65d96edc8ee276cc0a315a4693/lib/spack/spack/build_environment.py#L438) that the linker and the dynamic linker find and use the correct libraries when they are installed to the system directories without any additional flags, we should try to be consistent and have these changes in.

### DIFF
--- a/var/spack/repos/builtin/packages/openmpi/package.py
+++ b/var/spack/repos/builtin/packages/openmpi/package.py
@@ -12,6 +12,7 @@ import sys
 import llnl.util.tty as tty
 
 from spack.package import *
+from spack.util.environment import is_system_path
 
 
 class Openmpi(AutotoolsPackage, CudaPackage):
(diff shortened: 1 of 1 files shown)


hppritcha (dismissed):

alalazo (dismissed):

alalazo:
@skosukhin Let me know if you want to revisit this one too.

hppritcha:
please rebase

skosukhin:
The source branch is rebased. Unfortunately, I don't have enough time now to double-check everything again (e.g. I don't know whether all options that accept a prefix also accept `yes` as a value).

alalazo:

PR #32394, opened 2022-08-26
libxml2: add new versions

skosukhin:
This adds the latest version of the library `2.10.1` and the latest `2.9.x` version, `2.9.14`.

It also adds a patch that prevents the library from raising `SIGFPE` in certain cases (see [here](https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/186)). I would not bother adding it if there were version `2.9.15`. However, the fix was added only starting `2.10.0` and I'm not sure whether `2.10.x` is fully compatible with the code that currently works with `2.9.x`.


tldahlgren (approved):
Confirmed new sha256 and patch URL exists.

tldahlgren:
Internal server errors are being reported in the pipeline.

@spackbot run pipeline

PR #32395, opened 2022-08-26
Initial commit for freesurfer package

robgics:
Add package.py

### DIFF
--- a/var/spack/repos/builtin/packages/freesurfer/package.py
+++ b/var/spack/repos/builtin/packages/freesurfer/package.py
@@ -1,0 +1,38 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Freesurfer(Package):
+    """Freesurfer is an open source software suite for processing and analyzing
+    (human) brain MRI images."""
+
+    homepage = "http://freesurfer.net/"
+
+    # A license is required, but is free to obtain.
+    license_required = True
+    license_files = ["./license.txt"]
+
+    def url_for_version(self, version):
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/freesurfer/package.py
```diff
@@ -0,0 +1,40 @@
...
+    homepage = "http://freesurfer.net/"
+
+    # A license is required, but is free to obtain.
+    license_required = True
+    license_files = ["./license.txt"]
+
+    maintainers = ["robgics"]
+
+    def url_for_version(self, version):
+        return "http://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/{0}/freesurfer-linux-centos7_x86_64-{1}.tar.gz".format(
```
tldahlgren: We prefer to use the `https` protocol.

```suggestion
        return "https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/{0}/freesurfer-linux-centos7_x86_64-{1}.tar.gz".format(
```

var/spack/repos/builtin/packages/freesurfer/package.py
```diff
@@ -0,0 +1,40 @@
...
+
+    homepage = "http://freesurfer.net/"
+
+    # A license is required, but is free to obtain.
+    license_required = True
+    license_files = ["./license.txt"]
+
+    maintainers = ["robgics"]
+
+    def url_for_version(self, version):
```
tldahlgren: Please move this method so it occurs after the last `depends_on` directive.

tldahlgren (changes requested):

var/spack/repos/builtin/packages/freesurfer/package.py
robgics: Changes made and committed

tldahlgren (approved):
LGTM.  (Did not download the software to confirm the three sha256 though.)

PR #32398, opened 2022-08-26
OpenBLAS 0.3.21: w/o Fortran

ax3l:
There is a new OpenBLAS release out that can be compiled w/o a Fortran compiler.

Refs.:
- https://github.com/xianyi/OpenBLAS/releases/tag/v0.3.21
- https://twitter.com/isuru_f/status/1562120002496565249

macOS XCode developers, rejoice. (not me, but my friends & colleagues.) Maybe at some point Spack becomes a package manager that can be used without using another package manager first (to get gfortran) :tada: (Spack has no user-friendly compiler bootstrap yet.)


ax3l:
@spackbot please not ;-)
Screenshot from 2022-08-26 11-32-13

adamjstewart:
Now if only scipy didn't also require Fortran...

ax3l:
I get scipy et al via pip once a Spack env is up unless I happen to match a binary cache on the platform :sweat: 
Our production dev env otherwise builds half a day+ on a developer laptop: https://warpx.readthedocs.io/en/latest/install/dependencies.html#spack-macos-linux


Just saw, current openblas `~fortran` install has these warnings but succeeds:
```
expr: syntax error: unexpected argument 'cc]'
expr: syntax error: unexpected argument 'cc]'
make -f Makefile.install install
```
Building `py-numpy` from it seems to work though.

ax3l:
Ready from my side for review approval :)

adamjstewart:
I'm only maintaining Python/geospatial packages, I'll let someone else review this.

tldahlgren (dismissed):
Confirmed sha256.

tldahlgren:
Looks like `perl` is failing for a couple of stacks.

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -309,10 +320,9 @@ def make_defs(self):
         # When mixing compilers make sure that
         # $SPACK_ROOT/lib/spack/env/<compiler> have symlinks with reasonable
         # names and hack them inside lib/spack/spack/compilers/<compiler>.py
-        make_defs = [
-            "CC={0}".format(spack_cc),
-            "FC={0}".format(spack_fc),
-        ]
+        make_defs = ["CC={0}".format(spack_cc)]
+        if "~fortran" not in self.spec:
```
haampie, line 324:
```suggestion
        if "+fortran" in self.spec:
```

i don't have issues not following double negations but I think it's more clear this way :p

haampie, line 324: Oh, this is a double negation because the variant is conditional? Maybe add a comment so that people don't change it into `+fortran in self.spec` later on?

ax3l: Oh, nothing special in mind. Let me simplify:
```suggestion
        if "+fortran" in self.spec:
```

ax3l:
Double negation simplified & rebased.

tldahlgren:
Looks like pipeline may have had some pod issues (at least one stack's `phist` build).  Performing branch update to force re-running checks.

tldahlgren (dismissed):

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -309,10 +320,9 @@ def make_defs(self):
         # When mixing compilers make sure that
         # $SPACK_ROOT/lib/spack/env/<compiler> have symlinks with reasonable
         # names and hack them inside lib/spack/spack/compilers/<compiler>.py
-        make_defs = [
-            "CC={0}".format(spack_cc),
-            "FC={0}".format(spack_fc),
-        ]
+        make_defs = ["CC={0}".format(spack_cc)]
+        if "+fortran" in self.spec:
```
haampie:
```suggestion
        if "~fortran" not in self.spec:
```

We should set `FC` for OpenBLAS < 0.3.21

haampie (changes requested):
Sorry for the late review, we do need that double negation for older versions of OpenBLAS that require `FC` to be set.

haampie (dismissed):

tldahlgren (dismissed):

tldahlgren:
Still having pipeline issues with `perl` on Mac and `phist` (https://gitlab.spack.io/spack/spack/-/jobs/3328481).

haampie:
Now looking into this phist issue. Seems like openblas updated reference lapack from 3.9.0 to 3.10.1 in openblas 0.3.21 with some changes to the headers.

```
/tmp/root/spack-stage/spack-stage-phist-1.10.0-7ytoen3ddba7zwdm4dtppn6bytwoscyk/spack-src/src/core/phist_sdFact_def.hpp:435:41: error: too few arguments to function 'void dgemqrt_(const char*, const char*, const int*, const int*, const int*, const int*, const double*, const int*, const double*, const int*, double*, const int*, double*, int*, size_t, size_t)
```

```console
$ spack cd -i openblas@0.3.20
$ cd include
$ gcc -E lapack.h | grep -A10 dgemqrt
void dgemqrt_(
    char const* side, char const* trans,
    int const* m, int const* n, int const* k, int const* nb,
    double const* V, int const* ldv,
    double const* T, int const* ldt,
    double* C, int const* ldc,
    double* work,
    int* info );
```

vs

```console
$ spack cd -i openblas@0.3.21
$ cd include
$ gcc -E lapack.h | grep -A12 dgemqrt
void dgemqrt_(
    char const* side, char const* trans,
    int const* m, int const* n, int const* k, int const* nb,
    double const* V, int const* ldv,
    double const* T, int const* ldt,
    double* C, int const* ldc,
    double* work,
    int* info

    , size_t, size_t

);
```

This is due to `#define LAPACK_FORTRAN_STRLEN_END` being defined unconditionally in `lapack.h`.

Simplest solution for now is maybe do just add a conflicts with phist?

haampie:
For reference: https://github.com/Reference-LAPACK/lapack/issues/604

Not sure how to handle this... I think phist isn't the only package that will break.

haampie:
Okay, so if this ever comes up, packages should move away from directly using the lapack function `xyz` from the header file, instead they should use the macro `LAPACK_xyz(...)`. For phist I'll just add a conflict.

haampie (approved):

PR #32403, opened 2022-08-26
add sra-tools and NCBI-VDT packages

downloadico:
Adding two packages because sra-tools depends on ncbi-vdt


downloadico:
I have an open issue with the sra-tools project because it doesn't build: https://github.com/ncbi/sra-tools/issues/708

tldahlgren (approved):
Confirmed the tags for the packages exist.

Issue #31941, opened 2022-08-05
binaries.spack.io: cannot install openblas from cache due to failed checksum verification


cameronrutherford:
### Steps to reproduce

https://github.com/LLNL/hiop/blob/develop/.github/workflows/spack_build.yml - this is the github workflow file that we are using.

This seems to be an intermittent error that has persisted across spack versions, and has been documented happening for several months. Here is a list of other failing pipeline examples:
- https://github.com/LLNL/hiop/runs/6866279062?check_suite_focus=true#step:4:96 - happening with zlib 07/13
- https://github.com/LLNL/hiop/runs/6838474890?check_suite_focus=true#step:4:91 - pkgconf 07/10
- https://github.com/LLNL/hiop/runs/6838376509?check_suite_focus=true#step:4:202 - bzip2 07/10

I have also seen this with both the E4S pipeline, and the spack binary buildcache. @eugeneswalker has been working with me and is currently unable to re-produce, but I am able to reproduce the issue in GitHub CI.

### Error message

Failure output without debugging mode:

```
==> Warning: Skipping build of hiop-develop-xaesidee7vv3efdzevuf6zbpaozjihaf since openblas-0.3.20-c55doyb7a445uicmmcoz4x3wiw4zskow failed
==> Installing autoconf-2.69-o7yngkilzghf3fgpxglxua2p4gdvpegk
==> Error: Failed to install openblas due to NoChecksumException: Package tarball failed checksum verification.
==> Fetching https://binaries.spack.io/develop/build_cache/linux-ubuntu18.04-x86_64-gcc-7.5.0-autoconf-2.69-o7yngkilzghf3fgpxglxua2p4gdvpegk.spec.json.sig
It cannot be installed.
==> Fetching https://binaries.spack.io/develop/build_cache/linux-ubuntu18.04-x86_64/gcc-7.5.0/autoconf-2.69/linux-ubuntu18.04-x86_64-gcc-7.5.0-autoconf-2.69-o7yngkilzghf3fgpxglxua2p4gdvpegk.spack
==> Error: Terminating after first install failure: Package tarball failed checksum verification.
==> Extracting autoconf-2.69-o7yngkilzghf3fgpxglxua2p4gdvpegk from binary cache
It cannot be installed.
[+] /opt/spack/opt/spack/linux-ubuntu18.04-x86_64/gcc-7.5.0/autoconf-2.69-o7yngkilzghf3fgpxglxua2p4gdvpegk
[...]

PR #32407, opened 2022-08-26
Fix cause of checksum failures in public binary mirror

scottwittenburg:
Protected pipeilnes are configured to sync all specs from stack-specific mirrors into a single root mirror for more convenient public use.  Until now, this was managed by a final job generated into each stack-specific child pipeline.  If multiple child pipelines built the same hash, this could result in the root mirror sometimes containing `spec.json.sig` and `.spack` files corresponding to the same hash, but coming from different child pipeline mirrors.  While two binary packages with the same hash are otherwise equivalent (and relocatable), the install path is reflected in the package and causes the two packages to have different checksums.

This change moves the copying of the buildcache to a root job that runs after all the child pipelines have finished, so that the operation can be coordinated across all child pipelines.  This lets us ensure the `.spec.json.sig` and `.spack` files for any spec in the root mirror always come from the same child pipeline mirror, and also it allows us to avoid copying of duplicates.

Also with this change, only the *new* binary packages (built in the current pipeline) are copied to the mirror.  To support this, `spack ci generate` writes out a json file containing the urls of buildcache files to be copied, and `spack buildcache sync` can now  consume json files from all child pipelines at once to determine the subset of specs to sync to a mirror.  This is enabled by providing the option `--manifest-glob` and a quoted file pattern to select the json files to read.  

<details>
<summary>JSON format</summary>

```json
{
    "cn7znah2eetd7ueztmxzf76plh6tftws": [
        {
            "src": "s3://spack-binaries/testing/develop/tutorial/build_cache/linux-ubuntu18.04-x86_64-gcc-7.5.0-zlib-1.2.8-cn7znah2eetd7ueztmxzf76plh6tftws.spec.json.sig",
            "dest": "s3://spack-binaries/testing/develop/build_cache/linux-ubuntu18.04-x86_64-gcc-7.5.0-zlib-1.2.8-cn7znah2eetd7ueztmxzf76plh6tftws.spec.json.sig"
        }, 
        {
            "src": "s3://spack-binaries/testing/develop/tutorial/build_cache/linux-ubuntu18.04-x86_64/gcc-7.5.0/zlib-1.2.8/linux-ubuntu18.04-x86_64-gcc-7.5.0-zlib-1.2.8-cn7znah2eetd7ueztmxzf76plh6tftws.spack",
            "dest": "s3://spack-binaries/testing/develop/tutorial/build_cache/linux-ubuntu18.04-x86_64/gcc-7.5.0/zlib-1.2.8/linux-ubuntu18.04-x86_64-gcc-7.5.0-zlib-1.2.8-cn7znah2eetd7ueztmxzf76plh6tftws.spack"
        }
    ],
    "6wpnwvhne2xnlsvsegcw2ap34ziapzqa": [
        {
            "src": "s3://spack-binaries/testing/develop/tutorial/build_cache/linux-ubuntu18.04-x86_64-gcc-7.5.0-libiconv-1.16-6wpnwvhne2xnlsvsegcw2ap34ziapzqa.spec.json.sig",
            "dest": "s3://spack-binaries/testing/develop/build_cache/linux-ubuntu18.04-x86_64-gcc-7.5.0-libiconv-1.16-6wpnwvhne2xnlsvsegcw2ap34ziapzqa.spec.json.sig"
        }, 
        {
            "src": "s3://spack-binaries/testing/develop/tutorial/build_cache/linux-ubuntu18.04-x86_64/gcc-7.5.0/libiconv-1.16/linux-ubuntu18.04-x86_64-gcc-7.5.0-libiconv-1.16-6wpnwvhne2xnlsvsegcw2ap34ziapzqa.spack",
            "dest": "s3://spack-binaries/testing/develop/tutorial/build_cache/linux-ubuntu18.04-x86_64/gcc-7.5.0/libiconv-1.16/linux-ubuntu18.04-x86_64-gcc-7.5.0-libiconv-1.16-6wpnwvhne2xnlsvsegcw2ap34ziapzqa.spack"
        }
    ],
    ...
}
```

</details>

The copying of buildcache files addressed here remains off by default, and enabled using the `SPACK_COPY_BUILDCACHE` environment variable set to the copy target url.

Though all the changes above only affected protected pipelines, this includes one small fix for PR pipelines as well.  Now PR binaries are created into stack-specific mirrors (the same as is already done for protected pipelines).  This avoids jobs in PR pipelines that rebuild many (or sometimes all) specs overwriting each others binaries when there are hash conflicts.

Also note that although this should prevent the creation of future mismatched binaries (fixes the cause of the problem), it doesn't address mismatched `.spec` and `.spack` files already present in the public binary mirror.  We will clean out and repopulate that mirror [manually](https://github.com/spack/spack-infrastructure/pull/310) once this PR is merged and after the termination of any remaining `develop` pipeline that might create more mismatches.

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -647,11 +647,9 @@ def generate_gitlab_ci_yaml(
 
     spack_buildcache_copy = os.environ.get("SPACK_COPY_BUILDCACHE", None)
     if spack_buildcache_copy:
-        buildcache_copies = {
-            "destination_url_prefix": spack_buildcache_copy,
-            "hashes": {},
-        }
+        buildcache_copies = {}
         buildcache_copy_src_prefix = remote_mirror_override or remote_mirror_url
+        buildcache_copy_dest_prefix = spack_buildcache_copy
 
     if "mirrors" not in yaml_root or len(yaml_root["mirrors"].values()) < 1:
         tty.die("spack ci generate requires an env containing a mirror")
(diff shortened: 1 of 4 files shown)


scottwittenburg:
fyi @kwryankrattiger

eugeneswalker (approved):
Thanks!

cameronrutherford:
Does this close #31941? I would love to re-enable my GitHub spack pipeline, but perhaps I have to wait for the index to.  be updated manually before things will work

scottwittenburg:
Almost, but not quite.  It should prevent creationj of future mismatches, but we still need to clear out all the existing ones.  We have a plan for that, and I'll respond back here when it's done.  Thanks for being patient for so long.

cameronrutherford:
No problem, thanks for fixing this!! If anything I'm encouraged to submit more spack bugs in the future

PR #32413, opened 2022-08-27
dealii:: fixed application of stdcxx variant

luca-heltai:
If the compiler does not set the c++17 standard by default (as is the case for my clang compiler), the variant `stdcxx=17` was simply ignored, because it was only used when `+optflags` and `cxx_flags_release` were being set.

With this PR, we set `CMAKE_CXX_FLAGS`, whenever the script sets `cxx_flags`. This fixes, among others, also the `stdcxx=17` issue I was observing.


tldahlgren (approved):
LGTM

PR #32414, opened 2022-08-27
coreutils: add the external find feature using md5sum

cessenat:
Adds the external find feature that works for GNU systems.
This uses md5sum.


alalazo (approved):

alalazo (approved):

cessenat:
Thanks @alalazo

PR #32425, opened 2022-08-29
snphylo: update to new version

snehring:
The old distribution point fell off the internet. New url appears to be the original author's github.

### DIFF
--- a/var/spack/repos/builtin/packages/snphylo/package.py
+++ b/var/spack/repos/builtin/packages/snphylo/package.py
@@ -10,10 +10,13 @@ class Snphylo(Package):
     """A pipeline to generate a phylogenetic tree from huge SNP data"""
 
     homepage = "http://chibba.pgml.uga.edu/snphylo/"
-    url = "http://chibba.pgml.uga.edu/snphylo/snphylo.tar.gz"
+    url = "https://github.com/thlee/SNPhylo/archive/refs/tags/20180901.tar.gz"
 
+    version("20180901", sha256="46273bcafc8f6cc2465fc524926af7dd122bcc7b7ce1b7c3f5c278111c3a9ddd")
     version(
-        "2016-02-04", sha256="d9e144021c83dbef97bebf743b92109ad0afcfe70f37c244059b43f11b8a50da"
+        "2016-02-04",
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/snphylo/package.py
```diff
@@ -10,10 +10,13 @@ class Snphylo(Package):
     """A pipeline to generate a phylogenetic tree from huge SNP data"""
 
     homepage = "http://chibba.pgml.uga.edu/snphylo/"
-    url = "http://chibba.pgml.uga.edu/snphylo/snphylo.tar.gz"
+    url = "https://github.com/thlee/SNPhylo/archive/refs/tags/20180901.tar.gz"
 
+    version("20180901", sha256="46273bcafc8f6cc2465fc524926af7dd122bcc7b7ce1b7c3f5c278111c3a9ddd")
```
tldahlgren, line 15: Confirmed new sha256 with new URL.

var/spack/repos/builtin/packages/snphylo/package.py
```diff
@@ -10,10 +10,13 @@ class Snphylo(Package):
...
     version(
-        "2016-02-04", sha256="d9e144021c83dbef97bebf743b92109ad0afcfe70f37c244059b43f11b8a50da"
+        "2016-02-04",
+        sha256="d9e144021c83dbef97bebf743b92109ad0afcfe70f37c244059b43f11b8a50da",
```
tldahlgren, line 18: This sha256 won't work with the new URL (in part because that version isn't present) so one option is to add the `url` argument to this command.  

However, the old URL site appears to be inaccessible.  Interestingly, I did not realize that `spack stage` will stage the latest version in this case and with the only feedback being that it could not find the package.  I'd hoped to at least see something about the version being deprecated.

tldahlgren (approved):

tldahlgren:
@snehring Do you know if the original site is only temporarily unavailable or if that will be permanent?

snehring:
I do not know for sure, but per https://github.com/thlee/SNPhylo/issues/44 it's probably been gone for over a year.

PR #24718, opened 2021-07-06
Format all of Spack with black

tgamblin:
This [blackens](https://github.com/psf/black) the `spack` repository, scientifically 😄.

We want:
1. To be able to fix most `flake8` errors automatically, so as not to burden people with code style.
2. To keep configuration simple and avoid having two code styles (one for packages and one for core).
3. To avoid forcing a tool on package contributors who aren't all familiar with the `python` ecosystem.  i.e., we do not want to hold up anyone's package PR on style -- we want this to be faster than what we're already doing, even *without* a tool installed.

### How many columns?

Black defaults to 88 columns, but since we want to have one width for both core and packages, that is too narrow.  You can see the effect of blackening packages and core for different column widths here:

181120716-f3895121-cb6b-4699-84e9-9207834279eb

Using 88 or fewer columns would add about 100k lines to the code base. We don't want that. Given the data and the spirit of black's 88-character limit, we will use a width of 99 characters because:

* 99 is one less than 100, a nice round number, and all lines will fit in a 100-character wide terminal (even when the text editor puts a \ at EOL).
* 99 is just past the knee the file size curve for packages, and it means that packages remain readable and not significantly longer than they are now.
* It doesn't seem to hurt core -- files in core might change length by a few percent but seem like they'll be mostly the same as before -- just a bit more roomy.

See [this comment](https://github.com/spack/spack/pull/24718#issuecomment-876933636) for more details.  

### How to fix style automatically?

Once this is merged, you can do a few different things to fix Spack's style automatically:

1. Run `spack style --fix` -- this will automatically bootstrap and run `isort`, `mypy`, `black`, and `flake8` in that order, and it will automatically fix `isort` and `black` errors.  `black` fixes *most* `flake8` errors, but `flake8` still does useful checks (e.g., finds unused variables), so we still run it.
2. Type `@spackbot fix style` in a PR comment.  Spackbot now knows how to run `spack style --fix`, in case it doesn't work in your environment for some reason.
3. Use `black` yourself. Note that you will need to use `black` at version 21 or older, as we still support Python 2.7.  Note that `black` *requires* Python 3, but versions 21 and before will check 2.7 code.  Versions 22 on drop support for 2.7.

### What else is different?

Because `black` formats them automatically, we no longer allow exceptions for long lines in Spack, *except* those that contain URLs.  The only way you can get a really long line now is if you have a really long string, and you will need to either:

1. Break up multi-line strings with parentheses, e.g.:
    ```python
    depends_on(
[...]

PR #32426, opened 2022-08-29
Fix spack locking on some NFS systems

### LINKED ISSUE #17407: Permission error when activating environment by a different user
We have a Spack installation on our NFSv4 cluster. When a user tries to activate an environment, sometimes it fails with a PermissionError as Spack tries acquire a read-only lock on the .spack-env/transaction_lock file.

1. Log in to a compute node under a different user than the owner of the .spack-env directory.

2. Try to activate an environment:
```console
$ spack env activate gromacs_patterson
==> Error: [Errno 13] Permission denied: '/opt/spack/var/spack/environments/gromacs_patterson/.spack-env'
```

3. See if the directory is already there
```console
$ ls /opt/spack/var/spack/environments/gromacs_patterson
spack.lock  spack.yaml  spack.yaml~
```

4. Try to activate once more: success!
```console
$ spack env activate gromacs_patterson
```

### Error Message

```console
$ spack -d env activate gromacs_patterson
==> [2020-07-07-16:05:28.170359, 27289] Imported env from built-in commands
==> [2020-07-07-16:05:28.172212, 27289] Imported env from built-in commands
==> [2020-07-07-16:05:28.172850, 27289] Reading config file /opt/spack/etc/spack/defaults/config.yaml
==> [2020-07-07-16:05:28.182003, 27289] Reading config file /opt/spack/etc/spack/config.yaml
==> [2020-07-07-16:05:28.182706, 27289] READ LOCK: /opt/spack/var/spack/environments/gromacs_patterson/.spack-env/transaction_lock[0:0] [Acquiring] 
  File "/opt/spack/bin/spack", line 64, in <module>
    sys.exit(spack.main.main())
  File "/opt/spack/lib/spack/spack/main.py", line 761, in main
  File "/opt/spack/lib/spack/spack/main.py", line 489, in _invoke_command
  File "/opt/spack/lib/spack/spack/cmd/env.py", line 428, in env
    action(args)
  File "/opt/spack/lib/spack/spack/cmd/env.py", line 112, in env_activate
    active_env = ev.get_env(namedtuple('args', ['env'])(env),
  File "/opt/spack/lib/spack/spack/environment.py", line 319, in get_env
    return read(env)
  File "/opt/spack/lib/spack/spack/environment.py", line 373, in read
    return Environment(root(name))
  File "/opt/spack/lib/spack/spack/environment.py", line 582, in __init__
    with lk.ReadTransaction(self.txlock):
  File "/opt/spack/lib/spack/llnl/util/lock.py", line 576, in __enter__
    if self._enter() and self._acquire_fn:
  File "/opt/spack/lib/spack/llnl/util/lock.py", line 603, in _enter
    return self._lock.acquire_read(self._timeout)
  File "/opt/spack/lib/spack/llnl/util/lock.py", line 304, in acquire_read
    wait_time, nattempts = self._lock(fcntl.LOCK_SH, timeout=timeout)
  File "/opt/spack/lib/spack/spack/util/lock.py", line 31, in _lock
    return super(Lock, self)._lock(op, timeout)
  File "/opt/spack/lib/spack/llnl/util/lock.py", line 154, in _lock
    parent = self._ensure_parent_directory()
  File "/opt/spack/lib/spack/llnl/util/lock.py", line 240, in _ensure_parent_directory
    os.makedirs(parent)
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/opt/spack/var/spack/environments/gromacs_patterson/.spack-env'
```

```console
$ spack debug report
* **Spack:** 0.15.0-38-be1009db6
* **Python:** 3.8.3
* **Platform:** linux-archrolling-skylake
```

### Additional information

This only happens on compute nodes where the Spack folder is shared using NFS4. I suspect the attribute caching mechanism of NFS to be the culprit. 

```python
>>> import os
>>> os.makedirs('/opt/spack/var/spack/environments/gromacs_patterson/.spack-env')
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/opt/spack/var/spack/environments/gromacs_patterson/.spack-env'
>>> os.path.isdir('/opt/spack/var/spack/environments/gromacs_patterson/.spack-env')
True
>>> os.makedirs('/opt/spack/var/spack/environments/gromacs_patterson/.spack-env')
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/opt/spack/var/spack/environments/gromacs_patterson/.spack-env'
```

As demonstrated above, if the user does not have write permission for the root directory of the environment, ```os.makedirs()``` fails with PermissionError before actually checking if the `.spack_env` subdirectory exists. However, if its existence is checked beforehand, `os.makedirs()` raises the FileExistsError, which is already handled in the `_ensure_parent_directory()` method of the `Lock` class in the `llnl.util.lock` module.

Although it goes against the Pythonic wisdom of "Better ask for forgiveness than permission", it might solve the problem to insert a 
```python
os.path.is_dir(parent)
```
check at the beginning of the above mentioned method.

The other, non-Spack solution would be to make the attribute caching of NFS4 more aggressive. However, this negatively affects the performance of the network file system.

Thank you in advance,

Andras

- [x] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [x] I have searched the issues of this repo and believe this is not a duplicate
- [x] I have run the failing commands in debug mode and reported the output

sethrj:
It turns out that on some filesystem/os combinations, calling `os.makedirs` can raise `PermissionError` (`errno.EPERM`) if the directory is extant and readable, but not writable. This call happened before trying to activate environments.

Fixes #17407

### DIFF
--- a/lib/spack/llnl/util/lock.py
+++ b/lib/spack/llnl/util/lock.py
@@ -393,9 +393,8 @@ class Lock(object):
                 # Directory already exists
                 if e.errno != errno.EEXIST:
                     tty.warn(
-                        "Got OSError when trying to create directory at '{!s}' that already exists: {!r}".format(
-                            parent, e
-                        )
+                        "Got OSError when trying to create directory "
+                        "at '{!s}' that already exists: {!r}".format(parent, e)
                     )
             else:
                 raise
--- a/lib/spack/llnl/util/lock.py
+++ b/lib/spack/llnl/util/lock.py
@@ -392,7 +392,11 @@ class Lock(object):
             elif os.path.isdir(parent):
                 # Directory already exists
                 if e.errno != errno.EEXIST:
-                    tty.warn("Got OSError when trying to create directory at '{!s}' that already exists: {!r}".format(parent, e))
+                    tty.warn(
+                        "Got OSError when trying to create directory at '{!s}' that already exists: {!r}".format(
+                            parent, e
+                        )
+                    )
             else:
                 raise
         return parent
--- a/lib/spack/llnl/util/lock.py
+++ b/lib/spack/llnl/util/lock.py
@@ -386,8 +386,14 @@ class Lock(object):
         try:
             os.makedirs(parent)
         except OSError as e:
-            # makedirs can fail when diretory already exists.
-            if not (e.errno == errno.EEXIST and os.path.isdir(parent) or e.errno == errno.EISDIR):
+            if e.errno == errno.EISDIR:
+                # Directory already exists
+                pass
+            elif os.path.isdir(parent):
+                # Directory already exists
+                if e.errno != errno.EEXIST:
+                    tty.warn("Got OSError when trying to create directory at '{!s}' that already exists: {!r}".format(parent, e))
+            else:
                 raise
         return parent
 
--- a/lib/spack/llnl/util/lock.py
+++ b/lib/spack/llnl/util/lock.py
@@ -386,17 +386,12 @@ class Lock(object):
         try:
             os.makedirs(parent)
         except OSError as e:
-            if e.errno == errno.EISDIR:
-                # Directory already exists
-                pass
-            elif os.path.isdir(parent):
-                # Directory already exists
-                if e.errno != errno.EEXIST:
-                    tty.warn(
-                        "Got OSError when trying to create directory "
-                        "at '{!s}' that already exists: {!r}".format(parent, e)
-                    )
-            else:
+            # os.makedirs can fail in a number of ways when the directory already exists.
+            # With EISDIR, we know it exists, and others like EEXIST, EACCES, and EROFS
+            # are fine if we ensure that the directory exists.
+            # Python 3 allows an exist_ok parameter and ignores any OSError as long as
+            # the directory exists.
+            if not (e.errno == errno.EISDIR or os.path.isdir(parent)):
                 raise
         return parent
 


lib/spack/llnl/util/lock.py
```diff
@@ -386,8 +386,17 @@ def _ensure_parent_directory(self):
...
+                # Directory already exists
+                pass
+            elif os.path.isdir(parent):
+                # Directory already exists
+                if e.errno != errno.EEXIST:
+                    tty.warn(
+                        "Got OSError when trying to create directory "
+                        "at '{!s}' that already exists: {!r}".format(parent, e)
+                    )
+            else:
```
tgamblin, line 399:
```suggestion
            # makedirs can fail several ways when diretory already exists.
            if not (
                # directory already exists and makedirs told us about it
                (e.errno == errno.EEXIST and os.path.isdir(parent))
                # directory already exists, and mode doesn't match default makedirs mode
                # (can happen when we try to take a read lock but don't have permission
                # on the parent directory)
                or (e.errno == errno.EACCESS and os.path.isdir(parent))
                # more specific way for makedirs to tell us the directory exists
                or e.errno == errno.EISDIR
            ):
```

tgamblin (changes requested):
I think catching `OSError` and just warning if the directory already exists is too broad.  We're trying to be really specific in `llnl.util.lock` and I think it's better to fail fast in the corner cases than to let something potentially dangerous slip through.

#17407 is giving you errno 13, which is `EACCES`.  If that is the same error you see with NFSv4, I think we should check for that `errno` specifically, and allow it if it happens that the directory exists.

This is ok since as you note, we *may* be trying to get a read-only lock.  I think we could avoid this issue altogether in Python 3 with the [`exist_ok` parameter](https://docs.python.org/3/library/os.html#os.makedirs) to `os.makedirs` but sadly we (for the time being) need to support 2.7.  It'd be interesting to know if `exist_ok` gets rid of the permissions error if you are trying this with Python 3.

tgamblin:
Ok actually I just looked at the cpython source, and it's broad.  So why don't we just go with that.  [Here is what they do](s](https://docs.python.org/3/library/os.html#os.makedirs)) in Python 3:

```python
    try:
        mkdir(name, mode)
    except OSError:
        # Cannot rely on checking for EEXIST, since the operating system
        # could give priority to other errors like EACCES or EROFS
        if not exist_ok or not path.isdir(name):
            raise
```

So apparently we care about `EROFS` and maybe others too.

lib/spack/llnl/util/lock.py
tgamblin:
```suggestion
            # os.makedirs can fail in a number of ways when the directory already exists.
            # With EISDIR, we know it exists, and others like EEXIST, EACCES, and EROFS
            # are fine if we ensure that the directory exists.
            # Python 3 allows an exist_ok parameter and ignores any OSError as long as
            # the directory exists.
            if not (e.errno == errno.EISDIR or os.path.isdir(parent)):
```

I don't think we need to warn if we're doing basically what cpython would do in Python 3.

tgamblin (changes requested):

tgamblin (approved):

sethrj:
Thanks @tgamblin !

tpdownes:
@tgamblin the way I read things, this change is not presently slated to make it into v0.18.2. Is it possible to make that happen? We have been hitting it regularly maintaining shared Spack installations for our HPC users.

https://github.com/GoogleCloudPlatform/hpc-toolkit/issues/539

haampie:
Me and @alalazo are usually the only people doing backports, but I've largely stopped doing that since #24718 because pretty much nothing applies cleanly.

This is a small fix, so we can probably do it.

PR #30594, opened 2022-05-10
visit: Overhaul to build in the DAV SDK

chuckatkins:

PR #32427, opened 2022-08-29
visit: actually set the dev env for plugins again

cessenat:
VisIt extendable feature is based on plugin development ; plugin dev is possible when cmake is run with
VISIT_INSTALL_THIRD_PARTY=ON
From https://www.visitusers.org/index.php?title=VisIt_CMake_Build_System:
 VISIT_INSTALL_THIRD_PARTY : Install VisIt's 3rd party I/O libs and includes to permit plugin development
There are further explanations here https://www.visitusers.org/index.php?title=Installing_using_CMake
This PR in a way reverses PR https://github.com/spack/spack/pull/30594 that removed the comment,
and also PR https://github.com/spack/spack/pull/31748 where VISIT_INSTALL_THIRD_PARTY was set to False.
Probably @chuckatkins will comment as author of these PRs.
My PR adds the 'plugins' variant that as a default is set to True ; do you agree this way @chuckatkins ?

### DIFF
--- a/var/spack/repos/builtin/packages/visit/package.py
+++ b/var/spack/repos/builtin/packages/visit/package.py
@@ -81,6 +81,7 @@ class Visit(CMakePackage):
     variant("vtkm", default=False, description="Enable VTK-m support")
     variant("conduit", default=True, description="Enable Conduit support")
     variant("mfem", default=True, description="Enable MFEM support")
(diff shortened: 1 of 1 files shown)


cyrush (approved):
@cessenat I think this is a good approach to restore what you need.

cessenat:
Thanks @alalazo and @cyrush

Issue #31961, opened 2022-08-05
Metapackages should register their sub-packages as explicitly installed


wspear:
### Summary

When a metapackage such as ecp-data-vis-sdk installs requested 'sub-packages' those packages are treated as typical dependents and not flagged as explicit installs. 

### Rationale

Metapackages essentially hide the packages they provide from any logic operating on explicitly installed packages. This is an issue for E4S in particular since we generate reports and run tests on explicitly installed packages. This is potentially an issue for general users who may want to locate a metapackage-provided package for their direct use and will have to jump through additional hoops to distinguish it from other installs of the same package that were implicitly installed as dependencies of other packages.

### Description

 I suggest adding an API feature to manually tag a dependency as explicit from within a spack package. When this is available (or if it already exists) I suggest applying it to the packages installed by ecp-data-vis-sdk and similar metapackages.

### Additional information

Observed as of:
0.19.0.dev0 (42de3ff968f8bc32221ced071ba3633bca670bb4)

@eugeneswalker 
@kwryankrattiger 
@tgamblin 

### General information

- [X] I have run `spack --version` and reported the version of Spack
- [X] I have searched the issues of this repo and believe this is not a duplicate

wspear:
This issue applies to other metapackages like: 
exaworks
@andre-merzky

alalazo:
> This is an issue for E4S in particular since we generate reports and run tests on explicitly installed packages.

In my opinion the solution in #32428 is backwards. For a use case that, if I understand correctly, happens outside of Spack the proposal is to add an argument to a directive that:
1. Can create issues with the interpretation of what `explicit` means (clashing values among edges)
2. Doesn't make sense most of the time for regular packages

Can't we solve this issue where the E4S reports are generated, and special case `BundlePackage`s there? Also, the argument you make for the potential issue on query is reversed here on uninstall. Right now, in principle, we can do:
```console
$ spack install <bundle>
...
$ spack uninstall <bundle>
$ spack gc  # Should remove everything installed by the bundle
```
If we mark the packages as explicit we have to chase them on uninstall, since `spack gc` won't act on explicit packages.

wspear:
What is the intention of 'explicit' if not marking the top level 'user requested' packages, to distinguish them from the slushpile of dependencies? Metapackages are there to curate sets of these user-facing packages. I figure E4S's issues just encapsulate the troubles of any other spack use case that uses metapackages and cares about explicit. 

[...]

PR #32428, opened 2022-08-29
Packaging: Support promoting packages to explicit

kwryankrattiger:
BundlePackages really just more advanced versions of yaml spec lists with `unify:True`, so they should be able to mark their dependencies as being installed explicitly since the parent package is really just a pass-through package wrapper. #31961

FYI: @eugeneswalker @wspear @tgamblin 

I made the default `None` because I didn't want to imply it with forcing packages to be demoted to implicitly, not explicitly, installed. So there is either `explicit=True` which means promote, `explicit=None` which means do nothing to the explicit nature of a package, and the `explicit=False` is currently ignored because it isn't needed but could easily be implemented.

Another possible parameter name could also be `promote` to be consistent with the phase name, then it would make sense as `True` or `False`, but I just thought of that as I was typing up this PR so it isn't in there.

The implementation was done in a phase because injecting this into the `PackageInstaller` was a nightmare and it probably won't be used outside of `BundlePackage`s. I could be convinced the other way...but I would prefer not to be :)

### DIFF
--- a/lib/spack/spack/build_systems/bundle.py
+++ b/lib/spack/spack/build_systems/bundle.py
@@ -6,6 +6,8 @@ import spack.builder
 import spack.directives
 import spack.package_base
 
+from typing import Tuple
+
 
 class BundlePackage(spack.package_base.PackageBase):
     """General purpose bundle, or no-code, package class."""
@@ -21,6 +23,16 @@ class BundlePackage(spack.package_base.PackageBase):
     has_code = False
 
     spack.directives.build_system("bundle")
+
+    phases: Tuple[str, ...] = ("promote")
+
+    def promote(self, spec, prefix):
+        # Update the explicit dependencies
+        for _, cond_dict in self.dependencies.items():
+            for cond, dep in cond_dict.items():
+                if dep.explicit and cond in spec:
+                    spack.store.db.update_explicit(spec[dep.spec.name], True)
+
 
 
 @spack.builder.builder("bundle")
--- a/lib/spack/spack/dependency.py
+++ b/lib/spack/spack/dependency.py
@@ -94,6 +94,7 @@ class Dependency:
         pkg: "spack.package_base.PackageBase",
         spec: "spack.spec.Spec",
         type: Optional[Tuple[str, ...]] = default_deptype,
+        explicit: None,
     ):
         """Create a new Dependency.
 
@@ -106,6 +107,7 @@ class Dependency:
 
         self.pkg = pkg
         self.spec = spec.copy()
+        self.explicit = explicit
 
         # This dict maps condition specs to lists of Patch objects, just
         # as the patches dict on packages does.
--- a/lib/spack/spack/directives.py
+++ b/lib/spack/spack/directives.py
@@ -356,7 +356,7 @@ def version(ver, checksum=None, **kwargs):
     return _execute_version
 
 
-def _depends_on(pkg, spec, when=None, type=default_deptype, patches=None):
+def _depends_on(pkg, spec, when=None, type=default_deptype, patches=None, explicit=None):
     when_spec = make_when_spec(when)
     if not when_spec:
         return
@@ -397,9 +397,11 @@ def _depends_on(pkg, spec, when=None, type=default_deptype, patches=None):
 
     # this is where we actually add the dependency to this package
     if when_spec not in conditions:
-        dependency = Dependency(pkg, dep_spec, type=type)
+        dependency = Dependency(pkg, dep_spec, type=type, explicit=explicit)
         conditions[when_spec] = dependency
     else:
+        if explicit:
+            dependency.explicit = True
         dependency = conditions[when_spec]
         dependency.spec.constrain(dep_spec, deps=False)
         dependency.type |= set(type)
@@ -444,7 +446,7 @@ def conflicts(conflict_spec, when=None, msg=None):
 
 
 @directive(("dependencies"))
-def depends_on(spec, when=None, type=default_deptype, patches=None):
+def depends_on(spec, when=None, type=default_deptype, patches=None, explicit=None):
     """Creates a dict of deps with specs defining when they apply.
 
     Args:
@@ -462,7 +464,7 @@ def depends_on(spec, when=None, type=default_deptype, patches=None):
     """
 
     def _execute_depends_on(pkg):
-        _depends_on(pkg, spec, when=when, type=type, patches=patches)
+        _depends_on(pkg, spec, when=when, type=type, patches=patches, explicit=explicit)
 
     return _execute_depends_on
 
--- a/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
+++ b/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
@@ -9,7 +9,7 @@ from spack.package import *
 # Wrapper around depends_on to propagate dependency variants
 def dav_sdk_depends_on(spec, when=None, propagate=None):
     # Do the basic depends_on
-    depends_on(spec, when=when)
+    depends_on(spec, when=when, explicit=True)
 
     # Strip spec string to just the base spec name
     # ie. A +c ~b -> A
--- a/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
+++ b/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
@@ -150,7 +150,7 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
     # releases 0.8 and ascent can build with conduit@0.8: and vtk-m@1.7:
     conflicts("ascent@develop", when="+ascent")
 
-    depends_on("py-cinemasci", when="+cinema")
+    dav_sdk_depends_on("py-cinemasci", when="+cinema")
 
     # ParaView needs @5.11: in order to use CUDA/ROCM, therefore it is the minimum
     # required version since GPU capability is desired for ECP


kwryankrattiger:
@spackbot fix style

tldahlgren:
Curious.  What about having the install command explicitly add the `BundlePackage` dependency packages to the spec list instead of modifying `depends_on` directive?

kwryankrattiger:
@tldahlgren The reason `depends_on` was modified is not all of the dependencies listed in a `BundlePacakge` should necessarily be promoted to `explicit`. The `ecp-data-vis-sdk` is a `BundlePackage`, but also a `CudaPackage` and `ROCmPackage`. The deps implied from those two packages should not be promoted to explicit since they aren't member packages or even strong dependencies, they are just included that way to ensure variant consistency with other packages that use those build systems.

As far as implementing this inside of `BundlePackage` instead of `PackageInstaller`, two reasons.

1. It was way easier to reason about how and where to write this correctly inside of `BundlePackage` and add a stage that is transparent to the user what is happening.
2. I only want this promotion to happen inside of `BundlePackage`s and checking `isinstance(pkg, BundlePackage)` sort of lowers the benefit of adding more complexity inside of `PackageInstaller`. One possible benefit is to complain if `explicit` is specified on package dependencies of packages that are not `BundlePackage`s.. Although..there could be a property in `PackageBase` that says if it is/is not compatible with promoting deps to explicit and just check that with `getattr` in the `_depends_on` call...I might add that.

tldahlgren:
FYI. There are currently 20 built-in `BundlePackage` packages.

kwryankrattiger:
> FYI. There are currently 20 built-in `BundlePackage` packages.

Wow, that is a lot more than I thought! Listing and grouping them here for future reference.

Packages that use `BundlePackage` to abstract SDKs (will probably all use the new `promote` phase):
- opencl-headers
- geant4-data
- ecp-data-vis-sdk
- ecp-proxy-apps
- fastmath
- damask
- xsdk
- ceed
- bigdft-suite
- amd-aocl
- exaworks

Packages that use `BundlePackage` as a package abstraction:
- osmesa
- opengl
- glx

Packages that use `BudnlePackage` to configure an external package:
- spectrum-mpi
- mpt
- essl
- lhapdfsets
- apple-libuuid
- libc

kwryankrattiger:
Of all of the SDK/Suite like packages, these two are also used as dependencies for other projects in the builtin repo

* opencl-headers
* geant4-data

In these cases, promoting them to `explicit` always doesn't really make sense since their parent spec may also be used as a dependency in another package. To fix that case, it would be necessary to check what the root spec was that concretized the bundle package. Unfortunately, it is not possible to query the database  for that info during the `promote` phase since packages don't seem to be added to the database until after then are installed. Fixing this case is a strong argument for moving this stuff to `PackageInstaller` :disappointed: 

That said, _most_ packages like this will always be their own root spec in a list of specs, so I don't see it as an immediate problem, but certainly something worth noting.

alalazo (changes requested):
I have a few design concerns:
1. "Promote" is not a phase of a build system (like e.g. `./configure` etc.), just an internal implementation detail
2. Exposing the `explicit` argument to a directive doesn't seem good. What is the meaning if a transitive dependency is declared `explicit` (e.g. if `hdf5` declares `depends_on("zlib", explicit=True)`? What is the meaning if different branches reach to the same node with different values for the `explicit` attribute?

If all this PR wants to do is to mark dependencies of `BundlePackages` explicit (which can be discussed on its own) we can try to use `run_after` decorators or add appropriate hooks that work _just_ for bundle packages.

alalazo:
Did you consider, as an alternative approach, to extend the `spack mark` command to be similat to `apt-mark`? We can add a command that marks dependencies of `BundlePackages` "explicit" and one that does the opposite.

tldahlgren:
> I have a few design concerns:
> 
> 1. "Promote" is not a phase of a build system (like e.g. `./configure` etc.), just an internal implementation detail
> 2. Exposing the `explicit` argument to a directive doesn't seem good. What is the meaning if a transitive dependency is declared `explicit` (e.g. if `hdf5` declares `depends_on("zlib", explicit=True)`? What is the meaning if different branches reach to the same node with different values for the `explicit` attribute?
> 
> If all this PR wants to do is to mark dependencies of `BundlePackages` explicit (which can be discussed on its own) we can try to use `run_after` decorators or add appropriate hooks that work _just_ for bundle packages.

👆

tldahlgren:
My take on this is that the goal is for packages explicitly listed in the `BundlePackage` to be treated -- I assumed during concretization and installation -- as if they were explicitly listed in a `spack.yaml` file but for that package.

alalazo:
@tldahlgren Sure. I've taken a look at what `apt` does and apparently meta-packages have their dependencies installed explicitly by default (but no package has a meta-package as a dependency afaiu). They then have:
```
$ apt-mark -h
apt 2.0.9 (amd64)
Usage: apt-mark [options] {auto|manual} pkg1 [pkg2 ...]

apt-mark is a simple command line interface for marking packages
as manually or automatically installed. It can also be used to
manipulate the dpkg(1) selection states of packages, and to list
all packages with or without a certain marking.

Most used commands:
  auto - Mark the given packages as automatically installed
  manual - Mark the given packages as manually installed
  minimize-manual - Mark all dependencies of meta packages as automatically installed.
  hold - Mark a package as held back
  unhold - Unset a package set as held back
  showauto - Print the list of automatically installed packages
  showmanual - Print the list of manually installed packages
  showhold - Print the list of packages on hold

See apt-mark(8) for more information about the available commands.
Configuration options and syntax is detailed in apt.conf(5).
Information about how to configure sources can be found in sources.list(5).
Package and version choices can be expressed via apt_preferences(5).
Security details are available in apt-secure(8).
```
To change how installed packages are marked. I think we might do something similar, i.e. keep the current convention we have around installing bundle packages and add subcommands to `spack mark` to mark/unmark direct dependencies of bundle packages as explicit.

tldahlgren:
> @tldahlgren Sure. I've taken a look at what `apt` does and apparently meta-packages have their dependencies installed explicitly by default (but no package has a meta-package as a dependency afaiu). They then have:
> 
> ```
> $ apt-mark -h
> apt 2.0.9 (amd64)
> Usage: apt-mark [options] {auto|manual} pkg1 [pkg2 ...]
> 
> apt-mark is a simple command line interface for marking packages
> as manually or automatically installed. It can also be used to
> manipulate the dpkg(1) selection states of packages, and to list
> all packages with or without a certain marking.
> 
> Most used commands:
>   auto - Mark the given packages as automatically installed
>   manual - Mark the given packages as manually installed
>   minimize-manual - Mark all dependencies of meta packages as automatically installed.
>   hold - Mark a package as held back
>   unhold - Unset a package set as held back
>   showauto - Print the list of automatically installed packages
>   showmanual - Print the list of manually installed packages
>   showhold - Print the list of packages on hold
> 
> See apt-mark(8) for more information about the available commands.
> Configuration options and syntax is detailed in apt.conf(5).
> Information about how to configure sources can be found in sources.list(5).
> Package and version choices can be expressed via apt_preferences(5).
> Security details are available in apt-secure(8).
> ```
> 
> To change how installed packages are marked. I think we might do something similar, i.e. keep the current convention we have around installing bundle packages and add subcommands to `spack mark` to mark/unmark direct dependencies of bundle packages as explicit.

I haven't looked at that yet.  I suspect the "trick" is that they don't want inherited dependencies from the build systems to be included but perhaps I'm missing something.

kwryankrattiger:
> I have a few design concerns:
> 
>     1. "Promote" is not a phase of a build system (like e.g. `./configure` etc.), just an internal implementation detail
> 

I agree, it is gross. I would argue that the term "build system" in spack is pretty loosely defined since cuda and rocm are neither "build systems" so much has their own sort of `BundlePackage`, but they are expressed that way. In the sense of build system in spack, the `BundlePackage` is really just a spec list with extra steps and `promote` is the step that makes the `BundlePackage` transparent.

> If all this PR wants to do is to mark dependencies of `BundlePackages` explicit (which can be discussed on its own) we can try to use `run_after` decorators or add appropriate hooks that work _just_ for bundle packages.

As far has I could tell, the `run_after` hooks didn't have access to the list of the dependencies in the package. This was actually my first idea but I couldn't figure out how to make it work which is why I used a stage instead. If you can help me get past that, I am thrilled with dropping the stage and using hooks instead!

>     2. Exposing the `explicit` argument to a directive doesn't seem good. What is the meaning if a transitive dependency is declared `explicit` (e.g. if `hdf5` declares `depends_on("zlib", explicit=True)`? What is the meaning if different branches reach to the same node with different values for the `explicit` attribute?
> 

So, I am not really sure on how else to do this. As you can see in the list of existing `BundlePackages`, there are three different types, SDK/Suite packages, spack level abstractions for weird stacks (GL), and configuration scripts for externals. Only the first type should be doing any promotion to explicit. However, there is no distinction in packages on where dependencies get added from, so Cuda and ROCm package dependencies that shouldn't be promoted are still in the `BundlePackage.dependencies` map. There needs to be a way in the package to explicitly say "this package _can_ be promoted". Additionally, since some of those SDKs are also included as dependencies it can't be as straight forward as the current implementation.

> Did you consider, as an alternative approach, to extend the `spack mark` command to be similat to `apt-mark`? We can add a command that marks dependencies of `BundlePackages` "explicit" and one that does the opposite.

I thought about this but it doesn't really make sense for this use case. In the cases of using the SDK, it is not attractive to tell people they have to run more commands when using the SDK.

As far as uninstalling the SDK after these are marked explicit, I don't think that is a big deal. Being able to install the SDK, and then uninstall individual packages from the SDK without having to uninstall everything is actually a feature IMO.

alalazo:
> I would argue that the term "build system" in spack is pretty loosely defined since cuda and rocm are neither "build systems" so much has their own sort of BundlePackage, but they are expressed that way. 

`CudaPackage` and `RocmPackage` are mixins, with metadata that can be reused across packages. It is confusing that they ended up being stored under `build_systems`, but they don't drive build in any way - they just provide metadata or convenience methods.

alalazo:
> There needs to be a way in the package to explicitly say "this package can be promoted".

Sure, but why not having a method or any other mechanism in `BundlePackage`? This seems more appropriate than adding an argument to the `depends_on` directive that makes sense only in 20 packages out of 6k. What is the current use case? In the issue you mentioned generating reports for E4S?

kwryankrattiger:
> Sure, but why not having a method or any other mechanism in BundlePackage? This seems more appropriate than adding an argument to the depends_on directive that makes sense only in 20 packages out of 6k.

That is a good point. I can re-write this that way. Do you think it would be better as a member or should I write a directive for promotion?

kwryankrattiger:
>  What is the current use case? In the issue you mentioned generating reports for E4S?

The current use case is and SDK is listed in the lists of specs for E4S, and E4S has a script that looks at all of the "explicit" installed packages. The problem is, we want to test the SDK packages as well, but because they are hidden by the SDK they are not tested. Just pulling out the packages that we want to make explicit to the spec level an turning on `unify:True` is not possible because the spec list has multiple variations of the same package and `unify:when_possible` is not sufficient to guarantee that the SDK installed packages are the top level and may also trigger undesired builds of SDK packages outside of the SDK because it is easier to resolve in the concretizer. So it seems that the only way to ensure the SDK installed packages are always preferred and tested is to promote them to explicit.

lib/spack/spack/package_base.py
```diff
@@ -2924,14 +2924,21 @@ class BundlePackage(PackageBase):
...
     #: Bundle packages do not have associated source or binary code.
     has_code = False
 
+    def promote(self, spec, prefix):
+        # Update the explicit dependencies
+        for _, cond_dict in self.dependencies.items():
+            for cond, dep in cond_dict.items():
+                if dep.explicit and cond in spec:
+                    spack.store.db.update_explicit(spec[dep.spec.name], True)
+
```
tgamblin: How about:

```suggestion
    # list of direct dependency names that should be marked explicit after install
    explicit = []
    
    @run_after("install")
    def mark_explicit(self):
        for dep in self.spec.dependencies():
            if dep.name in BundlePackage.explicit:
                spack.store.db.update_explicit(dep, True)
```

tgamblin (changes requested):

lib/spack/spack/package_base.py
kwryankrattiger: Is there a way to get_explicit from the database of a package in `run_after("install")` I don't think this is updated until after the install step of the pacakge is complete but I couldn't quite understand the order.

We don't want to promote the bundle deps to a higher level than their parent bundle package.

kwryankrattiger:
I don't have time to chase this. Closing, but if someone is still interested in it feel free to revive.

PR #32433, opened 2022-08-29
update esmf package for cce (cray) compiler

jedwards4b:
Update esmf package to build with cce compiler and add a patch for esmf < 8.4.0 and
cce > 13


climbfuji (approved):

alalazo (approved):
Maintainer approved

PR #32438, opened 2022-08-30
Checksum: Warn if version is deprecated

tldahlgren:
This PR adds a method to check if a version is deprecated *and* will print a warning if that is the case for `spack checksum`.

The current output for a deprecated version can be confusing/misleading, especially when a deprecated package's URL is no longer valid or available.  For example:

```
$ spack info --no-dependencies --no-variants snphylo
Package:   snphylo

Description:
    A pipeline to generate a phylogenetic tree from huge SNP data

Homepage: http://chibba.pgml.uga.edu/snphylo/

Preferred version:  
    20180901      https://github.com/thlee/SNPhylo/archive/refs/tags/20180901.tar.gz

Safe versions:  
    20180901      https://github.com/thlee/SNPhylo/archive/refs/tags/20180901.tar.gz

Deprecated versions:  
    2016-02-04    https://github.com/thlee/SNPhylo/archive/refs/tags/2016-02-04.tar.gz

$ spack checksum snphylo 2016-02-04
==> Error: Could not find any versions for snphylo
```

So this PR adds a warning and elaborates a *little* on the issue.

```
spack checksum snphylo 2016-02-04
==> Warning: Version 2016-02-04 is deprecated
==> Error: Could not find any remote versions for snphylo
```

TODO
- [x] Add unit test(s) I(F this has a chance to be merged)

### DIFF
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -100,6 +100,26 @@ _spack_configure_argsfile = "spack-configure-args.txt"
 is_windows = sys.platform == "win32"
 
 
+def deprecated_version(pkg, version):
+    """
+    Returns a list of the deprecated versions of the package.
+
+    Arguments:
+        pkg (Package): The package whose version is to be checked.
+        version (str or spack.version.VersionBase): The version being checked
+
+    Returns: (bool) True if version is deprecated, False otherwise
+    """
+    if not isinstance(version, VersionBase):
+        version = Version(version)
+
+    for k, v in pkg.versions.items():
+        if version == k and v.get("deprecated", False):
+            return True
+
+    return False
+
+
 def preferred_version(pkg):
     """
     Returns a sorted list of the preferred versions of the package.
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -101,14 +101,11 @@ is_windows = sys.platform == "win32"
 
 
 def deprecated_version(pkg, version):
-    """
-    Returns a list of the deprecated versions of the package.
+    """Return True if the version is deprecated, False otherwise.
 
     Arguments:
         pkg (Package): The package whose version is to be checked.
         version (str or spack.version.VersionBase): The version being checked
-
-    Returns: (bool) True if version is deprecated, False otherwise
     """
     if not isinstance(version, VersionBase):
         version = Version(version)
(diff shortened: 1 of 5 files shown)


tldahlgren:
Initial attempts to add the warning to `spack fetch` and `spack stage` didn't work as the commands go straight to fetching from the cached archive then reporting the package as not found.  Perhaps that will change when a `--fresh` option is added to those commands.

In the above example, the outputs for `spack stage` is:

```
$ spack stage snphylo 2016-02-04
==> Using cached archive: /usr/WS1/dahlgren/releases/spack/var/spack/cache/_source-cache/archive/46/46273bcafc8f6cc2465fc524926af7dd122bcc7b7ce1b7c3f5c278111c3a9ddd.tar.gz
==> Staged snphylo in /var/tmp/dahlgren/spack-stage/spack-stage-snphylo-20180901-zv43ygeisio4cussgerdsgiuqlpezqar
==> Error: Package '2016-02-04' not found.
```

tldahlgren:
Here's another case of macOS 3.8 unit test check's code coverage data collection warnings/problems.

tldahlgren:
Hopefully a rebase to develop gets past the macOS failure that was preventing code coverage from being updated.

lib/spack/spack/package_base.py
```diff
@@ -100,6 +100,26 @@
 is_windows = sys.platform == "win32"
 
 
+def deprecated_version(pkg, version):
+    """
+    Returns a list of the deprecated versions of the package.
```
alalazo:
```suggestion
    """Return True if the version is deprecated, False otherwise.
```

lib/spack/spack/package_base.py
```diff
@@ -100,6 +100,26 @@
...
+
+    Arguments:
+        pkg (Package): The package whose version is to be checked.
+        version (str or spack.version.VersionBase): The version being checked
+
+    Returns: (bool) True if version is deprecated, False otherwise
```
alalazo: (comment) We can probably omit this, since the first sentence already say that:
```suggestion
```

alalazo (changes requested):
Just a minot request, but LGTM

lib/spack/spack/package_base.py
tldahlgren: Thanks for catching that I didn't change the comment to match the final version of the method.

alalazo (approved):

PR #32443, opened 2022-08-30
sirius: new versions, libsci supports

simonpintarelli:
- add support for cray-libsci
- fix version dependencies (7.3.2 already depends on costa and eigen)

### DIFF
--- a/var/spack/repos/builtin/packages/sirius/package.py
+++ b/var/spack/repos/builtin/packages/sirius/package.py
@@ -21,6 +21,8 @@ class Sirius(CMakePackage, CudaPackage, ROCmPackage):
     version("develop", branch="develop")
     version("master", branch="master")
 
+    version("7.3.2", sha256="a256508de6b344345c295ad8642dbb260c4753cd87cc3dd192605c33542955d7")
+    version("7.3.1", sha256="8bf9848b8ebf0b43797fd359adf8c84f00822de4eb677e3049f22baa72735e98")
     version("7.3.0", sha256="69b5cf356adbe181be6c919032859c4e0160901ff42a885d7e7ea0f38cc772e2")
     version("7.2.7", sha256="929bf7f131a4847624858b9c4295532c24b0c06f6dcef5453c0dfc33fb78eb03")
     version("7.2.6", sha256="e751fd46cdc7c481ab23b0839d3f27fb00b75dc61dc22a650c92fe8e35336e3a")
@@ -213,6 +215,7 @@ class Sirius(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+boost_filesystem", when="~apps")
     conflicts("^libxc@5.0.0")  # known to produce incorrect results
     conflicts("+single_precision", when="@:7.2.4")
+    conflicts("+scalapack", when="^cray-libsci")
 
     # Propagate openmp to blas
     depends_on("openblas threads=openmp", when="+openmp ^openblas")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/sirius/package.py
```diff
@@ -159,6 +161,7 @@ class Sirius(CMakePackage, CudaPackage, ROCmPackage):
     variant(
         "profiler", default=True, description="Use internal profiler to measure execution time"
     )
+    variant("libsci", default=False, description="Use cray-libsci")
```
dev-zero: can't you do a "when libsci is the provider for lapack/scalapack, then..." instead of this explicit option?

tldahlgren (commented):
Confirmed new sha256.

var/spack/repos/builtin/packages/sirius/package.py
tldahlgren: @dev-zero Do you mean something along the lines:

```suggestion
    variant("libsci", default=True, when="+scalapack platform=cray", description="Use cray-libsci")
```

dev-zero: @tldahlgren I'd rather not introduce the `libsci` variant at all and later when it is being used in the current patchset, do this instead:

```python
    args = [ "-DUSE_CRAY_LIBSCI=%s" % "ON" if "^cray-libsci" in spec else "OFF" ] 
```

dev-zero: because with the current change one has to 1. make sure libsci is being used for blas/lapack (there is no explicit dep on it) and 2. enable the variant. Which seems odd.

tldahlgren:
@simonpintarelli @dev-zero Ping me when you're happy with this PR and I'll re-review.

var/spack/repos/builtin/packages/sirius/package.py
simonpintarelli: Thank you @dev-zero. You are right, using `sirius+scalapack ^cray-libsci` is the better option. I'll change it accordingly (need to test it first).

simonpintarelli:
Sorry for the delay. It should be correct now. `-DUSE_CRAY_LIBSCI` is passed to cmake if `^cray-libsci` is in the spec, scalapack can be turned on/off independently (this is how it is implemented in SIRIUS). If `~scalapack ^cray-libsci` is specified, BLAS/LAPACK functionality is taken from libsci, but parallel blas/lapack functionality remains disabled.
Ping @tldahlgren, @dev-zero

var/spack/repos/builtin/packages/sirius/package.py
```diff
@@ -299,6 +299,9 @@ def cmake_args(self):
                 ]
             )
 
+        if "^cray-libsci" in spec:
+            args.extend(["-DUSE_CRAY_LIBSCI=On"])
```
dev-zero:
```suggestion
            args.append(self.define("USE_CRAY_LIBSCI", "ON"))
```

untested, but for consistency with the rest?

dev-zero:
@simonpintarelli LGTM, thanks! small nitpicking detail, but no showstopper

var/spack/repos/builtin/packages/sirius/package.py
simonpintarelli: Thanks, I've changed it accordingly.

dev-zero:
@tldahlgren we're good, could you take a look again please?

dev-zero (dismissed):

alalazo (approved):

PR #32445, opened 2022-08-30
py-gpytorch: add v1.9.0; add new py-linear-operator

adamjstewart:
Successfully builds on macOS 12.5 (arm64) with Python 3.9.13 and Apple Clang 13.1.6.

https://github.com/cornellius-gp/gpytorch/releases/tag/v1.9.0

### DIFF
--- a/var/spack/repos/builtin/packages/py-gpytorch/package.py
+++ b/var/spack/repos/builtin/packages/py-gpytorch/package.py
@@ -17,6 +17,7 @@ class PyGpytorch(PythonPackage):
 
     maintainers = ["adamjstewart"]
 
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/py-gpytorch/package.py
```diff
@@ -17,6 +17,7 @@ class PyGpytorch(PythonPackage):
 
     maintainers = ["adamjstewart"]
 
+    version("1.9.0", sha256="a0608184c18a1f518d6a102473427abf00f5351421e12a934530953f6887b34b")
```
tldahlgren, line 20: Confirmed sha256.

tldahlgren (approved):
Confirmed sha256 for both packages.

PR #32446, opened 2022-08-30
pumi: fix path to smoketest input data

cwsmith:
This PR fixes the path to the input data needed to successfully run `spack test run pumi`.

https://github.com/SCOREC/core/issues/370

### DIFF
--- a/var/spack/repos/builtin/packages/pumi/package.py
+++ b/var/spack/repos/builtin/packages/pumi/package.py
@@ -134,13 +134,8 @@ class Pumi(CMakePackage):
             join_path(data_dir, "pipe.dmg"),
             join_path(data_dir, "pipe.smb"),
             join_path(self.prefix.bin, "pipe_2_.smb"),
-            "2"
+            "2",
         ]
         expected = "pipe_2_.smb written"
         description = "testing pumi mesh partitioning"
-        self.run_test(
-            mpiexec,
-            options,
-            expected,
-            purpose=description
-        )
+        self.run_test(mpiexec, options, expected, purpose=description)
--- a/var/spack/repos/builtin/packages/pumi/package.py
+++ b/var/spack/repos/builtin/packages/pumi/package.py
@@ -126,14 +126,21 @@ class Pumi(CMakePackage):
         self.run_test(exe, options, expected, purpose=description)
 
         mpiexec = self.spec["mpi"].prefix.bin.mpiexec
-        mpiopt = ["-n", "2"]
-        exe = ["split"]
-        options = ["../share/testdata/pipe.dmg", "../share/testdata/pipe.smb", "pipe_2_.smb", "2"]
-        expected = "mesh pipe_2_.smb written"
+        data_dir = self.prefix.share.testdata
+        options = [
+            "-n",
+            "2",
+            join_path(self.prefix.bin, "split"),
+            join_path(data_dir, "pipe.dmg"),
+            join_path(data_dir, "pipe.smb"),
+            join_path(self.prefix.bin, "pipe_2_.smb"),
+            "2"
+        ]
+        expected = "pipe_2_.smb written"
         description = "testing pumi mesh partitioning"
         self.run_test(
             mpiexec,
-            mpiopt + exe + options,
+            options,
             expected,
-            purpose=description,
+            purpose=description
         )
--- a/var/spack/repos/builtin/packages/pumi/package.py
+++ b/var/spack/repos/builtin/packages/pumi/package.py
@@ -115,7 +115,7 @@ class Pumi(CMakePackage):
         if self.spec.version <= Version("2.2.6"):
             return
         exe = "uniform"
-        options = ["../testdata/pipe.dmg", "../testdata/pipe.smb", "pipe_unif.smb"]
+        options = ['../share/testdata/pipe.dmg', '../share/testdata/pipe.smb', 'pipe_unif.smb']
         expected = "mesh pipe_unif.smb written"
         description = "testing pumi uniform mesh refinement"
         self.run_test(exe, options, expected, purpose=description, work_dir=self.prefix.bin)
@@ -123,7 +123,7 @@ class Pumi(CMakePackage):
         mpiexec = Executable(join_path(self.spec["mpi"].prefix.bin, "mpiexec")).command
         mpiopt = ["-n", "2"]
         exe = ["split"]
-        options = ["../testdata/pipe.dmg", "../testdata/pipe.smb", "pipe_2_.smb", "2"]
+        options = ['../share/testdata/pipe.dmg', '../share/testdata/pipe.smb', 'pipe_2_.smb', '2']
         expected = "mesh pipe_2_.smb written"
         description = "testing pumi mesh partitioning"
         self.run_test(
--- a/var/spack/repos/builtin/packages/pumi/package.py
+++ b/var/spack/repos/builtin/packages/pumi/package.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
+import warnings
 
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/pumi/package.py
```diff
@@ -109,15 +109,15 @@ def test(self):
         if self.spec.version <= Version("2.2.6"):
             return
         exe = "uniform"
-        options = ["../testdata/pipe.dmg", "../testdata/pipe.smb", "pipe_unif.smb"]
+        options = ["../share/testdata/pipe.dmg", "../share/testdata/pipe.smb", "pipe_unif.smb"]
         expected = "mesh pipe_unif.smb written"
         description = "testing pumi uniform mesh refinement"
         self.run_test(exe, options, expected, purpose=description, work_dir=self.prefix.bin)
```
tldahlgren: Stand-alone tests should not be run from the prefix due to issues with permissions in read-only installations. This can be avoided by either providing the full path to the files in the prefix or by copying the test data to the test stage and using the full path to executable(s).  For example,

```suggestion
        data_dir = self.prefix.share.testdata
        exe = self.prefix.bin.uniform
        options = [
            join_path(data_dir, "pipe.dmg"),
            join_path(data_dir, "pipe.smb"),
            join_path(self.prefix.bin, "pipe_unif.smb"),
        ]
        expected = "pipe_unif.smb written"
        description = "testing pumi uniform mesh refinement"
        self.run_test(exe, options, expected, purpose=description)
```

var/spack/repos/builtin/packages/pumi/package.py
```diff
@@ -109,15 +109,15 @@ def test(self):
...
 
         mpiexec = Executable(join_path(self.spec["mpi"].prefix.bin, "mpiexec")).command
```
tldahlgren: You don't need to pass an executable to `run_test` so you could simply set this to:

```suggestion
        mpiexec = self.spec["mpi"].prefix.bin.mpiexec
```

var/spack/repos/builtin/packages/pumi/package.py
```diff
@@ -109,15 +109,15 @@ def test(self):
...
         mpiopt = ["-n", "2"]
         exe = ["split"]
-        options = ["../testdata/pipe.dmg", "../testdata/pipe.smb", "pipe_2_.smb", "2"]
+        options = ["../share/testdata/pipe.dmg", "../share/testdata/pipe.smb", "pipe_2_.smb", "2"]
         expected = "mesh pipe_2_.smb written"
         description = "testing pumi mesh partitioning"
         self.run_test(
```
tldahlgren: As stated above, the test should not be executed in the installation directory so you want to remove the `work_dir` argument.

var/spack/repos/builtin/packages/pumi/package.py
```diff
@@ -109,15 +109,15 @@ def test(self):
...
+        options = ["../share/testdata/pipe.dmg", "../share/testdata/pipe.smb", "pipe_unif.smb"]
         expected = "mesh pipe_unif.smb written"
         description = "testing pumi uniform mesh refinement"
         self.run_test(exe, options, expected, purpose=description, work_dir=self.prefix.bin)
 
         mpiexec = Executable(join_path(self.spec["mpi"].prefix.bin, "mpiexec")).command
         mpiopt = ["-n", "2"]
         exe = ["split"]
-        options = ["../testdata/pipe.dmg", "../testdata/pipe.smb", "pipe_2_.smb", "2"]
+        options = ["../share/testdata/pipe.dmg", "../share/testdata/pipe.smb", "pipe_2_.smb", "2"]
```
tldahlgren: You can provide full paths here:

```suggestion
        options = [
            join_path(data_dir, "pipe.dmg"),
            join_path(data_dir, "pipe.smb"),
            "pipe_2_.smb",
            "2",
        ]
```

tldahlgren (commented):
Looks good though testing in the install prefix creates a problem for some use cases.

tldahlgren (changes requested):
See prior suggestions and comments.

cwsmith:
Hi @tldahlgren,

I made the suggested edits (outside github) and was able to run `spack test run pumi@2.2.7` without any obvious issues.

var/spack/repos/builtin/packages/pumi/package.py
```diff
@@ -108,22 +108,28 @@ def cmake_args(self):
...
         description = "testing pumi uniform mesh refinement"
-        self.run_test(exe, options, expected, purpose=description, work_dir=self.prefix.bin)
+        self.run_test(exe, options, expected, purpose=description)
 
-        mpiexec = Executable(join_path(self.spec["mpi"].prefix.bin, "mpiexec")).command
-        mpiopt = ["-n", "2"]
-        exe = ["split"]
-        options = ["../testdata/pipe.dmg", "../testdata/pipe.smb", "pipe_2_.smb", "2"]
-        expected = "mesh pipe_2_.smb written"
+        mpiexec = self.spec["mpi"].prefix.bin.mpiexec
```
wspear: Is there a case where the full path to the mpi executable is necessary? There are software environments where only the mpi compilers are provided in the mpi bin directory and the executors are located elsewhere. This test is breaking for us in that situation. Also, mpiexec is not always available. I recommend using something like this:

```
        mpiexe_list = ["mpirun", "mpiexec", "srun"]
        for mpiexe in mpiexe_list:
            if which(mpiexe) is not None:
                mpiexec = Executable(mpiexe).command
                break
```

wspear (changes requested):

var/spack/repos/builtin/packages/pumi/package.py
cwsmith: I'm not sure.  If the full path was not set, how would the spack test mechanism locate the mpi executor that matches the mpi install used for the build?  I haven't tried mixing an mpi executor from one mpi install with libraries/headers from another.

wspear: The correct mpi environment would have to be loaded for the test to run successfully so it would pick up the executor from the path. I think that is generally how spack smoke tests are implemented. This isn't to support mixing and matching mpi components but a single mpi with an atypical directory structure.

cwsmith: Sounds good.  Thank you.  I'll make the change.

cwsmith: Quick spack/python question.  Should we handle the case where `mpirun`, `mpiexec` or `srun` is not found?  If so, what is the clean way to report the error and exit?

wspear: That would be safer. Maybe skip the mpi test with a warning rather than a failure in that case? Ideally we should have an mpi executor method provided by spack so every mpi enabled test doesn't have to reimplement this...

tldahlgren: We have tests that pick the first of those executables it finds and runs with it.  If you want to demonstrate running with all available, you could do that.

(FYI. I'm refactoring stand-alone testing support so *maybe* I can sneak this in. 😉 )

alecbcs:
@cwsmith did you get a chance to take a look at the review feedback. Is this PR still something you'd like to pursue?

cwsmith:
@alecbcs Yes.  Sorry.  I'll get back to this after the holiday.

tldahlgren:
Do you want to resurrect this PR and resolve the conflicts?

cwsmith:
@tldahlgren Sure.  I'll rebase on develop.

tldahlgren:
> @tldahlgren Sure. I'll rebase on develop.

Great. Ping me here when you're done.

Sichao25:
I feel like the original path issue has already been resolved by #45181 .

alalazo:

PR #32448, opened 2022-08-30
hdf5-vol-log: Update package versions and update HDF5 dependency to version 1.13.2

lrknox:


var/spack/repos/builtin/packages/hdf5-vol-log/package.py
```diff
@@ -17,9 +17,11 @@ class Hdf5VolLog(AutotoolsPackage):
 
     version("master-1.1", branch="master")
 
-    version("1.1.0", commit="ca146fa7d320ec5c0b397669b330c78fceeabb57")
```
tldahlgren, line 20: Old versions are usually deprecated rather than deleted as they *may* exist in someone's cache.  See https://spack.readthedocs.io/en/latest/packaging_guide.html?highlight=deprecate#deprecating-old-versions.

var/spack/repos/builtin/packages/hdf5-vol-log/package.py
```diff
@@ -17,9 +17,11 @@ class Hdf5VolLog(AutotoolsPackage):
...
+    version("1.3.0", tag="logvol.1.3.0")
+    version("1.2.0", tag="logvol.1.2.0")
+    version("1.1.0", tag="logvol.1.1.0")
```
tldahlgren, line 22: Confirmed tags exist.

tldahlgren (commented):
See deprecation comment.

var/spack/repos/builtin/packages/hdf5-vol-log/package.py
lrknox: In this case, version 1.1.0 is still available but was updated from a commit to the tag logvol.1.1.0 tag in the vol-log-based source repo.  The hdf5 dependency minimum version was updated to the currently available 1.13.2 version from 1.13.0 which is no longer available.

hyoklee (approved):

tldahlgren (approved):

PR #32450, opened 2022-08-30
imagemagick: new variant ghostscript

wdconinc:
Ghostscript adds about 75 dependencies to an installation of just imagemagick (97 without, 172 with Ghostscript). This adds a variant (defaulting to true for backward compatibility) that allows users to turn off Ghostscript support.

### DIFF
--- a/var/spack/repos/builtin/packages/imagemagick/package.py
+++ b/var/spack/repos/builtin/packages/imagemagick/package.py
@@ -18,7 +18,7 @@ class Imagemagick(AutotoolsPackage):
     version("7.0.2-7", sha256="f2f18a97f861c1668befdaff0cc3aaafb2111847aab028a88b4c2cb017acfbaa")
     version("7.0.2-6", sha256="7d49ca8030f895c683cae69c52d8edfc4876de651f5b8bfdbea907e222480bd3")
 
-    variant("ghostscript", default=True, description="Compile with Ghostscript support")
+    variant("ghostscript", default=False, description="Compile with Ghostscript support")
 
     depends_on("jpeg")
     depends_on("pango")
--- a/var/spack/repos/builtin/packages/imagemagick/package.py
+++ b/var/spack/repos/builtin/packages/imagemagick/package.py
@@ -37,8 +37,6 @@ class Imagemagick(AutotoolsPackage):
         args = []
         spec = self.spec
         if spec.satisfies("+ghostscript"):
-            gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
-        if spec.satisfies("+ghostscript"):
             args.append("--with-gslib")
             gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
             args.append("--with-gs-font-dir={0}".format(gs_font_dir))
--- a/var/spack/repos/builtin/packages/imagemagick/package.py
+++ b/var/spack/repos/builtin/packages/imagemagick/package.py
@@ -18,6 +18,8 @@ class Imagemagick(AutotoolsPackage):
     version("7.0.2-7", sha256="f2f18a97f861c1668befdaff0cc3aaafb2111847aab028a88b4c2cb017acfbaa")
     version("7.0.2-6", sha256="7d49ca8030f895c683cae69c52d8edfc4876de651f5b8bfdbea907e222480bd3")
 
+    variant("ghostscript", default=True, description="Compile with Ghostscript support")
+
     depends_on("jpeg")
     depends_on("pango")
     depends_on("libtool", type="build")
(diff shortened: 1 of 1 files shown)


wdconinc:
> 172 with Ghostscript

Including both mesa and rust...

tldahlgren (dismissed):
LGTM

var/spack/repos/builtin/packages/imagemagick/package.py
```diff
@@ -26,15 +28,20 @@ class Imagemagick(AutotoolsPackage):
...
     depends_on("pkgconfig", type="build")
 
     def configure_args(self):
+        args = []
         spec = self.spec
-        gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
-        return ["--with-gs-font-dir={0}".format(gs_font_dir)]
+        if spec.satisfies("+ghostscript"):
+            gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
+            args.append("--with-gs-font-dir={0}".format(gs_font_dir))
```
tldahlgren, line 42: Curious.  Do you also need `--with-gslib` or is that "automatic" with `-with-gs-font-dir`?

wdconinc: It defaults to `with`, but you're right that it is better to be explicit.
```suggestion
        if spec.satisfies("+ghostscript"):
            args.append("--with-gslib")
            gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
            args.append("--with-gs-font-dir={0}".format(gs_font_dir))
```

var/spack/repos/builtin/packages/imagemagick/package.py
```diff
@@ -26,15 +28,23 @@ class Imagemagick(AutotoolsPackage):
...
     depends_on("libsm")
     depends_on("pkgconfig", type="build")
 
     def configure_args(self):
+        args = []
         spec = self.spec
-        gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
-        return ["--with-gs-font-dir={0}".format(gs_font_dir)]
+        if spec.satisfies("+ghostscript"):
+            gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
```
wdconinc:
```suggestion
```
Fix suggestion fail :-(

tldahlgren: BTW. If `font` is a subdirectory, you could just use `spec["ghostscript-fonts"].prefix.share.font`.

adamjstewart:
I would personally default to false to reduce build times...

wdconinc:
I agree, but we would need to get involvement from all dependent maintainers since they may be implicitly assuming that imagemagick always has ghostscript support. 

Let me tag them when I'm at a PC and can run that on a command line.

adamjstewart:
According to:
```console
$ spack maintainers $(spack dependents imagemagick)
```
it looks like  @glennpj @mtmiller  @pradyunsg  @siko1056 may have opinions since they maintain packages that depend on imagemagick.

Specific packages are:
```
meme  octave  py-matplotlib  py-wand  r-animation  r-magick  r-webshot
```

var/spack/repos/builtin/packages/imagemagick/package.py
```diff
@@ -26,15 +28,21 @@ class Imagemagick(AutotoolsPackage):
...
     depends_on("pkgconfig", type="build")
 
     def configure_args(self):
+        args = []
         spec = self.spec
-        gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share, "font")
-        return ["--with-gs-font-dir={0}".format(gs_font_dir)]
+        if spec.satisfies("+ghostscript"):
+            args.append("--with-gslib")
+            gs_font_dir = join_path(spec["ghostscript-fonts"].prefix.share.font)
```
tldahlgren: FYI. `join_path` is no longer needed in this case.

```suggestion
            gs_font_dir = spec["ghostscript-fonts"].prefix.share.font
```

adamjstewart:
Alternatively, if you don't want to wait for the above people to review this PR, you can try building those 7 packages and make sure they work with `imagemagick~ghostscript`.

glennpj:
I am okay with this change.

wdconinc:
I verified that all dependents build successfully without the ghostscript functionality.

I suspect the main impact will be that these dependent packages will not be able to save figures as `eps` (and maybe `pdf`) anymore unless they add `+ghostscript`. I tested with octave and it still writes to `pdf`, but not `eps`:
```
figure (1);
clf ();
surf (peaks);
print figure1.pdf
print figure1.eps
```

tldahlgren (dismissed):

cessenat:
> 

Concerning octave, which is of major interest for me, the default is that imagemagick support is not built.
To me, your PR is fair, and @adamjstewart suggestion to not enable ghostscript by default is fair.

pradyunsg:
I don't have any opinions on this. :)

adamjstewart (approved):

PR #32452, opened 2022-08-30
gdk-pixbuf: only build tests when requested

wdconinc:
The building of tests is optional [as of 2.42.9](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/commit/801eef111df624f4377baed9a90c94b6a2d4340c). This applies this option in the build.

The reason the option was added is to deal with test build failures in sandboxed environments and with certain glibc versions (caused by glib gresources, in fact). For example, with the latest version glibc and in the latest version of docker these tests [cannot be built](https://github.com/moby/moby/issues/43595).


tldahlgren (approved):
LGTM

PR #32461, opened 2022-08-31
NCCL should not be dependent on ROCm builds since RCCL is used instead.

rsanthanam-amd:
Allow the nccl flag to be specified even for ROCm builds so that NCCL kernels are included in the build.

In this case the NCCL kernels will use RCCL as the backend implementation.

### DIFF
--- a/var/spack/repos/builtin/packages/py-tensorflow/package.py
+++ b/var/spack/repos/builtin/packages/py-tensorflow/package.py
@@ -291,7 +291,7 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage):
         depends_on("rocblas")
         depends_on("rocfft")
         depends_on("hipfft")
-        depends_on("rccl")
+        depends_on("rccl", when="+nccl")
         depends_on("hipsparse")
         depends_on("hipcub")
         depends_on("rocsolver")
@@ -348,7 +348,7 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage):
     depends_on("cudnn@:6", when="@0.5:0.6 +cuda")
     depends_on("cudnn@:7", when="@0.7:2.2 +cuda")
     # depends_on('tensorrt', when='+tensorrt')
-    depends_on("nccl", when="+nccl")
+    depends_on("nccl", when="+nccl+cuda")
     depends_on("mpi", when="+mpi")
     # depends_on('android-ndk@10:18', when='+android')
     # depends_on('android-sdk', when='+android')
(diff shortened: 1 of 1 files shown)


adamjstewart:
Would it make sense to add a `+rccl` variant? Or should we use `depends_on("rccl", when="+nccl")`?

rsanthanam-amd:
depends_on("rccl", when="+nccl+rocm") maybe?

not sure if that is the right way to specify that rccl should only be used for rocm builds...

adamjstewart:
Yes, it's already in a `with when("+rocm"):` bracket so that should already be implied.

rsanthanam-amd:
> Would it make sense to add a `+rccl` variant? Or should we use `depends_on("rccl", when="+nccl")`?

Added this. PTAL.

Thanks!

var/spack/repos/builtin/packages/py-tensorflow/package.py
```diff
@@ -345,7 +345,8 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage):
     depends_on("cudnn@:6", when="@0.5:0.6 +cuda")
     depends_on("cudnn@:7", when="@0.7:2.2 +cuda")
     # depends_on('tensorrt', when='+tensorrt')
-    depends_on("nccl", when="+nccl")
+    depends_on("nccl", when="+nccl+cuda")
+    depends_on("rccl" when="+nccl")
```
adamjstewart: I think you'll want to modify the existing `depends_on("rccl")` line, not this one

rsanthanam-amd: Makes sense.  Fixed.  Sorry about that.

var/spack/repos/builtin/packages/py-tensorflow/package.py
```diff
@@ -289,7 +289,7 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage):
         depends_on("rocblas")
         depends_on("rocfft")
         depends_on("hipfft")
-        depends_on("rccl")
+        depends_on("rccl" when"+nccl")
```
adamjstewart:
```suggestion
        depends_on("rccl", when"+nccl")
```

rsanthanam-amd: Just noticed that.  fixed.

adamjstewart (dismissed):
LGTM, but I'll give @aweits a chance to review

rsanthanam-amd:
@adamjstewart @aweits 

gentle ping.

rsanthanam-amd:
@adamjstewart 

gentle ping

adamjstewart (approved):

PR #32463, opened 2022-08-31
tree: add 2.0.3

aphedges:
When I tried to compile `tree` in the past, it failed with `error: 'for' loop initial declarations are only allowed in C99 mode`. I needed to add `CFLAGS=-O3 -pedantic -Wall -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -std=c99` to use both the existing values from the Makefile and the new flag `-std=c99`. I felt this was too hacky to upstream. `tree` v2.0.3 now appends the environment-set `CFLAGS` to its own copy instead of overriding its own, so I feel much more comfortable upstreaming it.

### DIFF
--- a/var/spack/repos/builtin/packages/tree/package.py
+++ b/var/spack/repos/builtin/packages/tree/package.py
@@ -26,7 +26,7 @@ class Tree(Package):
 
     @when("@2:")
     def install(self, spec, prefix):
-        make("install", "PREFIX=%s" % prefix)
+        make("install", "PREFIX=%s" % prefix, "CFLAGS=-std=c99")
 
     @when("@:1")
     def install(self, spec, prefix):
--- a/var/spack/repos/builtin/packages/tree/package.py
+++ b/var/spack/repos/builtin/packages/tree/package.py
@@ -26,7 +26,13 @@ class Tree(Package):
 
     @when("@2:")
     def install(self, spec, prefix):
-        make("install", "PREFIX=%s" % prefix, "CFLAGS=-std=c99")
+        make(
+            "PREFIX=%s" % prefix,
+            "CC=%s" % spack_cc,
+            "CFLAGS=-O3 -pedantic -Wall -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -std=c99",
+            "LDFLAGS=-s",
+            "install",
+        )
 
     @when("@:1")
     def install(self, spec, prefix):
--- a/var/spack/repos/builtin/packages/tree/package.py
+++ b/var/spack/repos/builtin/packages/tree/package.py
@@ -19,6 +19,7 @@ class Tree(Package):
     homepage = "http://mama.indstate.edu/users/ice/tree/"
     url = "http://mama.indstate.edu/users/ice/tree/src/tree-1.7.0.tgz"
 
+    version("2.0.3", sha256="ba14e77b5f9dc7f8250c3f702ec5b6be2f93cd0fa87311bab3239676866a3b1d")
     version("2.0.2", sha256="7d693a1d88d3c4e70a73e03b8dbbdc12c2945d482647494f2f5bd83a479eeeaf")
     version("1.8.0", sha256="715d5d4b434321ce74706d0dd067505bb60c5ea83b5f0b3655dae40aa6f9b7c2")
     version("1.7.0", sha256="6957c20e82561ac4231638996e74f4cfa4e6faabc5a2f511f0b4e3940e8f7b12")


haampie:
Hm, they should be the ones setting a C standard :|.

haampie:
> v2.0.3 now appends the environment-set CFLAGS to its own copy instead of overriding its own

It doesn't?

```
gcc -std=c99 -c -o tree.o tree.c
gcc -std=c99 -c -o list.o list.c
gcc -std=c99 -c -o hash.o hash.c
gcc -std=c99 -c -o color.o color.c
gcc -std=c99 -c -o file.o file.c
gcc -std=c99 -c -o filter.o filter.c
gcc -std=c99 -c -o info.o info.c
gcc -std=c99 -c -o unix.o unix.c
gcc -std=c99 -c -o xml.o xml.c
gcc -std=c99 -c -o json.o json.c
gcc -std=c99 -c -o html.o html.c
gcc -std=c99 -c -o strverscmp.o strverscmp.c
gcc  -o tree tree.o list.o hash.o color.o file.o filter.o info.o unix.o xml.o json.o html.o strverscmp.o
```

var/spack/repos/builtin/packages/tree/package.py
```diff
@@ -19,13 +19,14 @@ class Tree(Package):
...
 
+    version("2.0.3", sha256="ba14e77b5f9dc7f8250c3f702ec5b6be2f93cd0fa87311bab3239676866a3b1d")
     version("2.0.2", sha256="7d693a1d88d3c4e70a73e03b8dbbdc12c2945d482647494f2f5bd83a479eeeaf")
     version("1.8.0", sha256="715d5d4b434321ce74706d0dd067505bb60c5ea83b5f0b3655dae40aa6f9b7c2")
     version("1.7.0", sha256="6957c20e82561ac4231638996e74f4cfa4e6faabc5a2f511f0b4e3940e8f7b12")
 
     @when("@2:")
     def install(self, spec, prefix):
-        make("install", "PREFIX=%s" % prefix)
+        make("install", "PREFIX=%s" % prefix, "CFLAGS=-std=c99")
```
haampie: Let's just stick to the hacky business for now...

```suggestion
        make(
            "PREFIX=%s" % prefix,
            "CC=%s" % spack_cc,
            "CFLAGS=-O3 -std=c99 -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64",
            "install",
        )
```

Or if you really don't like copying the defines, you can also inject `-O3 -std=c99` through the compiler wrapper:

```python
    def flag_handler(self, name, flags):
        if name == 'cflags':
            flags.append("-O3")
            flags.append("-std=c99")
        return (flags, None, None)
```

It would still be worth it passing `"CC=%s" % spack_cc,` to `make(...)` since they set `CC` in the makefile too.

aphedges:
> Hm, they should be the ones setting a C standard :|.

While I agree, there are some other problems with the `Makefile` as well. For example, it looks like the macOS build options are commented out and need to be manually changed.

> > v2.0.3 now appends the environment-set CFLAGS to its own copy instead of overriding its own
> 
> It doesn't?

The [changelog](http://mama.indstate.edu/users/ice/tree/changes.html) states the following:

- Use += when assigning CFLAGS and LDFLAGS in the Makefile allowing them to be modified by environment variables during make. (Ben Brown) Possibly assumes GNU make.

You are right, though. From the [Makefile](https://gitlab.com/OldManProgrammer/unix-tree/-/blob/8515e9df1e101630991de13404a6800be2ae517d/Makefile#L34-L37):

```make
# Linux defaults:
CFLAGS=-ggdb -pedantic -Wall -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64
#CFLAGS+=-O3 -pedantic -Wall -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64
#LDFLAGS+=-s
```

I guess the release notes are wrong. Looking at the commit itself (<https://gitlab.com/OldManProgrammer/unix-tree/-/commit/8515e9df1e101630991de13404a6800be2ae517d>) and the fact that the comments have switched, I suspect the author forgot to change the commenting used for testing.

I can improve the build changes here so we aren't building what appears to be the debug build.

haampie:
lmao, tree! There's no reason to worry about macOS, just setting `CFLAGS=-std=c99 -O3 -D_FILE_OFFSET_BITS=64` is portable enough, I don't believe `-D_LARGEFILE64_SOURCE` is really necessary.

var/spack/repos/builtin/packages/tree/package.py
aphedges: I'd prefer it the hacky way because the flags can be set properly, given the problems I noted in <https://github.com/spack/spack/pull/32463#issuecomment-1233406327>. I'll push an updated version shortly.

aphedges: I am also adding the commented-out `LDFLAGS`.

Does the order of the variables and `install` matter? I noticed you swapped them.

aphedges:
> lmao, tree! There's no reason to worry about macOS, just setting `CFLAGS=-std=c99 -O3 -D_FILE_OFFSET_BITS=64` is portable enough, I don't believe `-D_LARGEFILE64_SOURCE` is really necessary.

I don't know much about these compiler flags but according to <https://stackoverflow.com/a/14184121/2445901>, it's needed with `-std=c99`.

haampie:
As far as I can see they're only using `off_t` and some ifdefs for getting printf right, `-D_FILE_OFFSET_BITS=64` should be enough, also with `-std=c99`.

Note that the only thing `-D_LARGEFILE64_SOURCE` does is it exposes `fpos64_t`, `tmpfile64`, `fopen64`, `freopen64`, `fseeko64`, `ftello64`, `fgetpos64`, `fsetpos64` but those are not used in tree.

So, I'll get back to the PR tomorrow, if you're leaving the redundant define that's ok too.

haampie (approved):

PR #32467, opened 2022-08-31
Manifest parsing: skip invalid manifest files

scheibelp:
`spack external read-cray-manifest` is supposed to skip manifest files if they fail to parse (unless `--fail-on-error` is specified). I did not handle the case where the JSON is not valid (if it does not parse, or if it parses but does not conform to the schema). This widens the set of exceptions caught (the manifest parsing will still terminate if `--fail-on-error` is specified).


lib/spack/spack/cray_manifest.py
```diff
@@ -161,10 +163,18 @@ def entries_to_specs(entries):
...
+        decode_exception_type = ValueError
+
+    try:
+        with open(path, "r") as json_file:
+            json_data = json.load(json_file)
 
-    jsonschema.validate(json_data, manifest_schema)
+        jsonschema.validate(json_data, manifest_schema)
+    except (jsonschema.exceptions.ValidationError, decode_exception_type) as e:
+        raise ManifestValidationError(str(e))
```
tgamblin: use `six.raise_from()` to keep e around as the cause (at least in Python 3)

tgamblin (changes requested):

tgamblin:
Minor nitpick above.

Question: does this mean that there invalid json files in recent PE releases that we should be telling HPE about?

scheibelp:
> Question: does this mean that there are invalid json files in recent PE releases that we should be telling HPE about?

(EDIT: clarity) This was for an older release. If it occurs for newer schema versions I think it would be good to bring up with HPE.

Also: review is addressed (thanks!)

tgamblin (approved):

PR #32468, opened 2022-08-31
py-black: add v22.8.0

adamjstewart:
Successfully installs on macOS 12.5 (arm64) with Python 3.9.13 and Apple Clang 13.1.6.

https://github.com/psf/black/releases/tag/22.8.0


tldahlgren (approved):
Confirmed sha256.

PR #32469, opened 2022-09-01
Apply hip workaround for raja-framework

trws:
- add workaround for broken behavior in HIP
- propagate openmp requirement and workaround to camp

### DIFF
--- a/var/spack/repos/builtin/packages/camp/package.py
+++ b/var/spack/repos/builtin/packages/camp/package.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import glob
+
 from spack.package import *
 
 
@@ -27,6 +29,7 @@ class Camp(CMakePackage, CudaPackage, ROCmPackage):
 
     # TODO: figure out gtest dependency and then set this default True.
     variant("tests", default=False, description="Build tests")
+    variant("openmp", default=False, description="Build OpenMP support")
 
     depends_on("cub", when="+cuda")
 
@@ -39,6 +42,7 @@ class Camp(CMakePackage, CudaPackage, ROCmPackage):
 
         options.append("-DBLT_SOURCE_DIR={0}".format(spec["blt"].prefix))
 
+        options.append("-DENABLE_OPENMP=" + ("On" if "+openmp" in spec else "Off"))
         if "+cuda" in spec:
             options.extend(
                 ["-DENABLE_CUDA=ON", "-DCUDA_TOOLKIT_ROOT_DIR=%s" % (spec["cuda"].prefix)]
@@ -55,6 +59,12 @@ class Camp(CMakePackage, CudaPackage, ROCmPackage):
 
         if "+rocm" in spec:
             options.extend(["-DENABLE_HIP=ON", "-DHIP_ROOT_DIR={0}".format(spec["hip"].prefix)])
+            # there is only one dir like this, but the version component is unknown
+
+            options.append(
+                "-DHIP_CLANG_INCLUDE_PATH="
+                + glob.glob("{}/lib/clang/*/include".format(spec["llvm-amdgpu"].prefix))[0]
+            )
             archs = self.spec.variants["amdgpu_target"].value
             if archs != "none":
                 arch_str = ",".join(archs)
--- a/var/spack/repos/builtin/packages/camp/package.py
+++ b/var/spack/repos/builtin/packages/camp/package.py
@@ -29,6 +29,7 @@ class Camp(CMakePackage, CudaPackage, ROCmPackage):
     maintainers = ["trws"]
 
     version("main", branch="main", submodules="True")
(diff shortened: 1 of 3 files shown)


davidbeckingsale (dismissed):

davidbeckingsale:
Nice, thanks @trws!

trws:
@davidbeckingsale, I remember fixing this `-lopenmp` missing bug that's showing up in the gitlab builds for Umpire, but can't seem to remember how.  Do you happen to recognize it or remember what the workaround was? Maybe it's building a pre-fix version of umpire with a post-fix camp?

var/spack/repos/builtin/packages/raja/package.py
```diff
@@ -62,12 +63,14 @@ class Raja(CachedCMakePackage, CudaPackage, ROCmPackage):
     depends_on("camp@0.2.2:0.2.3", when="@0.14.0")
     depends_on("camp@0.1.0", when="@0.10.0:0.13.0")
     depends_on("camp@2022.03.0:", when="@2022.03.0:")
+    depends_on("camp+openmp", when="+openmp")
```
davidbeckingsale, line 68: @trws - Umpire needs this line too, I think that will fix the problem.

trws: Sadly, that did not take care of it.  I think we may need to get the updates in a release to get past this. =/

davidbeckingsale: :( - will constraining the camp version help?

trws: I can try it, I'll add the new version bump I put up for camp and see if it's enough, or if we need to bump the others.

trws: It looks like that might actually do it, I bumped camp to 2022.03.2 on the theory we might have to bump the other two to bump camp (the submodule is still wrong) but bumping it in spack seems to make gitlab happy: https://gitlab.spack.io/spack/spack/-/pipelines/164770

trws:
Ok, this is actually passing now.  Want to go forward with this and see if it gets us over the hump @davidbeckingsale?

davidbeckingsale:
> Ok, this is actually passing now. Want to go forward with this and see if it gets us over the hump @davidbeckingsale?

Yup that sounds good to me, thanks @trws !

cameronrutherford:
I have been working with this change in camp and it has broken the existing functioning ROCm installs that I have been working on. This ends up appending the string literal to the `cache.cmake` file used to configure RAJA and Umpire, and I get the following error:

```bash
2 errors found in build log:
  >> 5     CMake Error at login2-cray-sles15-zen3-clang@@14.0.0.cmake:42:
     6       Parse error.  Expected a command name, got unquoted argument with text
     7       "-DHIP_CLANG_INCLUDE_PATH=/opt/rocm-5.2.0/llvm/lib/clang/14.0.0/include".
     8
     9
  >> 10    CMake Error: Error processing file: /gpfs/alpine/proj-shared/csc359/spack-install/build-stage/rcruther/spack-stage-raja-0.1
           4.0-srsbc2bzv4ps3pbdja32wv3cfklbv4ug/spack-src/login2-cray-sles15-zen3-clang@@14.0.0.cmake
     11    -- Configuring incomplete, errors occurred!
```
I am determining a workaround, but it took me quite a while to figure out what I had broken when I updated the spack ref we are using...

var/spack/repos/builtin/packages/camp/package.py
```diff
@@ -3,9 +3,19 @@
...
+
 from spack.package import *
 
 
+def hip_repair_options(options, spec):
+    # there is only one dir like this, but the version component is unknown
+    options.append(
+        "-DHIP_CLANG_INCLUDE_PATH="
+        + glob.glob("{}/lib/clang/*/include".format(spec["llvm-amdgpu"].prefix))[0]
+    )
```
cameronrutherford, line 16: I think this needs to be changed slightly to resolve my issue:

```python
path = glob.glob("{}/lib/clang/*/include".format(spec["llvm-amdgpu"].prefix))[0]
options.append(cmake_cache_path("HIP_CLANG_INCLUDE_PATH", "{0}".format(path))

cameronrutherford: I have verified that with a camp already built, this patch fixes my problem. I will probably make a PR to add this change, but want to make sure I'm not breaking things elsewhere/in camp

trws: It's being caused by the fact that we have both CmakePackage and CachedCmakePackage for some reason.  There is a fix in the umpire package in #32504.

cameronrutherford: I see. Can/Will that fix also be applied to RAJA?

cameronrutherford: I see there is a RAJA PR linked to this as well, so I assume so. Thanks for the help!

PR #32472, opened 2022-09-01
Apply 6.26 COMPILE_DEFINITIONS patch unconditionally

graeme-a-stewart:
This bug isn't per-se tied to the root7 variant and should be applied always for these ROOT releases.

N.B. This patch is already applied in v6-26-00-patches (https://github.com/root-project/root/pull/11132) in the upstream ROOT repo, so any future stable release won't need it.


graeme-a-stewart:
I am quite confused by the test failures - they don't seem to have anything to do with my patch. Am I missing something?

wdconinc (dismissed):
Ah, yes, thank you!

vvolkl (dismissed):

graeme-a-stewart:
I rebased to re-trigger the CI, which worked, but unfortunately you have to reapprove @wdconinc @vvolkl

alalazo (approved):

PR #32405, opened 2022-08-26
Port package sanity unit tests to audits

alalazo:
Modifications:
- [x] Moved most unit test in `package_sanity.py` as checks in `audit.py`

PR #32473, opened 2022-09-01
Fix CI for package only PRs

alalazo:
This was broken by #32405 and not noticed because the bug happens in commits that modify only package recipes

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -321,7 +321,7 @@ jobs:
           rm lib/spack/docs/_spack_root
         else
           echo "ONLY PACKAGE RECIPES CHANGED [skipping coverage]"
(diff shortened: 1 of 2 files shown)


alalazo:
@trws FYI, since I went ahead merging this. We can definitely improve the workflow if we could say that unit-tests are required only for core changes.

alalazo:
Well, [this](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution) seems hacky, but interesting.

trws:
We're already subsetting based on what's changed actually, it would be really easy to make unit tests run only on core changes in fact because we already have a tag for it passed in there.

alalazo:
> it would be really easy to make unit tests run only on core changes in fact because we already have a tag for it passed in there.

Sure, what I thought was harder is avoid pending unit test if we activate status checks and we skip some jobs. It seems though if we skip something at the job level it is considered "Successful" - so should be fine.

PR #32475, opened 2022-09-01
Hyperqueue package

Nortamo:
New package for hyperqueue (https://github.com/It4innovations/hyperqueue). Scheduler for sub-node tasks for HPC systems with batch scheduling. 

Cargo takes care of all the dependencies, so the package does not specify them, but looking at other rust packages in spack this seems to be the current standard?


var/spack/repos/builtin/packages/hyperqueue/package.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
```
tldahlgren:
```suggestion
# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```

var/spack/repos/builtin/packages/hyperqueue/package.py
```diff
@@ -0,0 +1,31 @@
...
+    url = "https://github.com/It4innovations/hyperqueue/archive/refs/tags/v0.11.0.tar.gz"
+    git = "https://github.com/It4innovations/hyperqueue"
+
+    maintainers = ["Nortamo", "samiilvonen"]
+
+    version("main", branch="main")
+    version(
+        "0.12.0-rc1", sha256="0c7b5d567bb6cb8dd4e7bafdf784b0379cef74b3aecb958c7f20248f8fedfbc1"
+    )
+    version("0.11.0", sha256="07fa7eda3a8a5278e058a526fee92e1e524370813b362aaa1a5dfc49d1f3fc28")
```
tldahlgren: Do you want to flag this as the preferred version since the newer one is a release candidate?

var/spack/repos/builtin/packages/hyperqueue/package.py
```diff
@@ -0,0 +1,31 @@
...
+    version("0.10.0", sha256="53983b6382123ecd480884b9872a27296739db59e1bb9813b82225841f924790")
```
tldahlgren: Where did this come from?  It does not match what I'm seeing from `spack checksum`.

```suggestion
    version("0.10.0", sha256="dc022170bf45479cb03edfe55a2fde066e7bf69542f644a8e78db604e8c6d67f")
```

var/spack/repos/builtin/packages/hyperqueue/package.py
```diff
@@ -0,0 +1,31 @@
...
+    version("0.9.0", sha256="96e743ffac0512a278de9ca3277183536ee8b691a46ff200ec27e28108fef783")
```
tldahlgren: Same issue here:

```suggestion
    version("0.9.0", sha256="49be05b3a042382ee31dddea9a5de64d3e3ff747cb2c47fa170b407fa8fdd2e1")
```

tldahlgren (changes requested):
Question about preferred version plus two checksums appear to be incorrect.

var/spack/repos/builtin/packages/hyperqueue/package.py
Nortamo: Good idea.

var/spack/repos/builtin/packages/hyperqueue/package.py
Nortamo: Ah my bad those been left there from some earlier testing where I manually checksummed the source archive,
but accidentally picked the wrong asset (so `hyperqueue-0.10.0.tar.gz` instead of the `Source code (tar.gz)`  )

tldahlgren:
@kobzol Are you okay with being included as a maintainer on this package?

Kobzol:
Yes, I'm fine with it, thanks.

tldahlgren (approved):

PR #32397, opened 2022-08-26
llvm: fix 15.0.0rc builds on MacOS with command-line-tools

balay:
Ref: https://github.com/llvm/llvm-project/issues/57037

i.e use -DBUILTINS_CMAKE_ARGS=-DCOMPILER_RT_ENABLE_IOS=OFF. But this needs switching "compiler-rt" from "projects" to "runtimes".

Also fixing the warnings below fixes compile errors

CMake Warning at CMakeLists.txt:101 (message):
  Using LLVM_ENABLE_PROJECTS=libcxx is deprecated now, please use
  -DLLVM_ENABLE_RUNTIMES=libcxx or see the instructions at
  https://libcxx.llvm.org/BuildingLibcxx.html for building the runtimes.

CMake Warning at CMakeLists.txt:101 (message):
  Using LLVM_ENABLE_PROJECTS=libcxxabi is deprecated now, please use
  -DLLVM_ENABLE_RUNTIMES=libcxxabi or see the instructions at
  https://libcxx.llvm.org/BuildingLibcxx.html for building the runtimes.

CMake Warning at CMakeLists.txt:101 (message):
  Using LLVM_ENABLE_PROJECTS=libunwind is deprecated now, please use
  -DLLVM_ENABLE_RUNTIMES=libunwind or see the instructions at
  https://libcxx.llvm.org/BuildingLibcxx.html for building the runtimes.

/private/var/folders/nt/_m1t_x7j76q6sl3xt91tqgs00000gn/T/balay/spack-stage/spack-stage-llvm-15.0.0-rc2-h2t5bohzyy7exz2ub3m42pfycjcmbndk/spack-build-h2t5boh/include/c++/v1/cstdlib:135:9: error: no member named 'at_quick_exit' in the global namespace
using ::at_quick_exit _LIBCPP_USING_IF_EXISTS;
      ~~^

Also ref: https://github.com/spack/spack/pull/31945
cc: @William-Mou

PR #29530, opened 2022-03-15
Allow conditional possible values in variants

alalazo:
Allow declaring possible values for variants with an associated condition. If the variant takes one of those values, the condition is imposed as a further constraint.

The idea of this PR is to implement part of the mechanisms needed for modeling [packages with multiple build-systems]( https://github.com/spack/seps/pull/3). After this PR the build-system directive can be implemented as:
```python
variant(
    'build-system',
    default='cmake',
    values=(
        'autotools',
        conditional('cmake', when='@X.Y:')
    ), 
    description='...',
)
```

Modifications:
- [x] Allow conditional possible values in variants
- [x] Add a unit-test for the feature
- [x] Add documentation

PR #32476, opened 2022-09-01
llvm: optionally build compiler-rt, libcxx and libunwind as runtimes

skosukhin:
The main idea of this PR is to build `compiler-rt`, `libcxx` and `libunwind` as runtimes (not only when `@15:`, which was added in #32397 while I was working on this). This, for example, makes it possible to build `libcxx` even when `@13:%gcc@:10`.

I have also gone through several patches and updated constraints and comments for them. I also tried to remove patch files from Spack repo and switch to fetching them from the upstream repo, if possible (see https://github.com/spack/spack/pull/32476#issuecomment-1237151890). ~~There might be more patch files that can be removed from the repo in a similar way but I touched only those that I had to.~~

A short discussion that started with https://github.com/spack/spack/pull/29530#issuecomment-1228506094 might be worth mentioning here.


skosukhin:
The reason for [2d509b5](https://github.com/spack/spack/pull/32476/commits/2d509b51465d4132610eebea30984524096b7d10) is described in [#32492](https://github.com/spack/spack/pull/32492#issuecomment-1235582870).

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -88,7 +88,6 @@ class Llvm(CMakePackage, CudaPackage):
     variant(
         "flang",
         default=False,
-        when="@11: +clang",
```
trws, line 88: Why is this being removed?

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -97,31 +96,41 @@ class Llvm(CMakePackage, CudaPackage):
         default=False,
         description="Include debugging code in OpenMP runtime libraries",
     )
-    variant("lldb", default=True, when="+clang", description="Build the LLVM debugger")
```
trws, line 92: Why is this being removed?

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -97,31 +96,41 @@ class Llvm(CMakePackage, CudaPackage):
...
+    variant("lldb", default=True, description="Build the LLVM debugger")
     variant("lld", default=True, description="Build the LLVM linker")
     variant("mlir", default=False, when="@10:", description="Build with MLIR support")
     variant(
-        "internal_unwind",
-        default=True,
-        when="+clang",
-        description="Build the libcxxabi libunwind",
+        "libunwind",
```
trws, line 101: If you want to add an alias with this name, do it, and we can deprecate the old one, but too many things rely on the variants of this package to do it without a deprecation period at this point.

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -97,31 +96,41 @@ class Llvm(CMakePackage, CudaPackage):
...
+        values=("none", "project", conditional("runtime", when="+clang @6:")),
+        default="runtime",
```
trws, line 107: Is there ever a reason *not* to build it as a runtime in versions after it becomes safe to do so?

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -188,10 +197,11 @@ class Llvm(CMakePackage, CudaPackage):
         description="Build with OpenMP capable thread sanitizer",
     )
     variant(
-        "omp_as_runtime",
-        default=True,
-        when="+clang @12:",
-        description="Build OpenMP runtime via ENABLE_RUNTIME by just-built Clang",
+        "omp",
```
trws: Same variant issue as above, name needs to have a deprecation period.

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: https://github.com/spack/spack/pull/32476#issuecomment-1235585173

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -207,9 +217,7 @@ class Llvm(CMakePackage, CudaPackage):
         description="Add shared library symbol version",
         when="@13:",
     )
-    variant(
-        "z3", default=False, when="+clang @8:", description="Use Z3 for the clang static analyzer"
-    )
+    variant("z3", default=False, description="Use Z3 for the clang static analyzer")
```
trws, line 229: This is useful when `~clang` and/or `@:8`?

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -266,6 +280,12 @@ class Llvm(CMakePackage, CudaPackage):
     depends_on("gmp", when="@:3.6 +polly")
     depends_on("isl", when="@:3.6 +polly")
 
+    conflicts("+flang", when="@:10")
+    conflicts("+flang", when="~clang")
+    conflicts("+lldb", when="~clang")
+    conflicts("+z3", when="@:7")
+    conflicts("+z3", when="~clang")
```
trws: Ah, got it, I can resolve the other comments, but I would greatly prefer that these be grouped with the appropriate `variant()` directives so they're easier to take in as a whole.

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: This is also part of [2d509b5](https://github.com/spack/spack/pull/32476/commits/2d509b51465d4132610eebea30984524096b7d10) (`llvm: allow for explicit 'llvm~clang~flang~lldb~z3'`)

trws: Makes sense, moreso since I found the conflicts, a small refactor would be nice but I agree with the result. Thanks for working through that.

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -274,25 +294,35 @@ class Llvm(CMakePackage, CudaPackage):
...
-    conflicts("%gcc@7:", when="@:4+libcxx")
-    conflicts("%clang@6:", when="@:4+libcxx")
-    conflicts("%apple-clang@6:", when="@:4+libcxx")
-    conflicts("%gcc@7:", when="@:4+compiler-rt")
-    conflicts("%clang@6:", when="@:4+compiler-rt")
-    conflicts("%apple-clang@6:", when="@:4+compiler-rt")
+    # libcxx=project imposes compiler conflicts
+    # see https://libcxx.llvm.org/#platform-and-compiler-support for the latest release
+    # and https://github.com/llvm/www-releases for older releases
+    with when("libcxx=project"):
```
trws, line 312: @tgamblin we need a way to deal with useful things like this for the caching stuff, all the inputs are literal, but the directive calls receive variables.  Maybe something we could explore https://pypi.org/project/pystarlark/ for, make an invocation of pyskylark with a literal containing directives behave like invocations of those literals? Or, easier, make it so we can import directives from a flat yaml or json value or better dhall or cue document or something?  We need to be able to do this stuff, but not being able to cache llvm (given its ubiquity and complexity) would be really unfortunate. =/

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -305,45 +335,85 @@ class Llvm(CMakePackage, CudaPackage):
...
+            # see https://reviews.llvm.org/D47281
+            patch(
+                "https://github.com/llvm/llvm-project/commit/383fe5c8668f63ef21c646b43f48da9fa41aa100.patch?full_index=1",
+                sha256="66f01ac1769a6815aba09d6f4347ac1744f77f82ec9578a1158b24daca7a89e6",
+                when="@4:6.0.0",
+            )
+            # fix sanitizer-common build with glibc 2.31
+            # see https://reviews.llvm.org/D70662
+            patch("sanitizer-ipc_perm_mode.patch", when="@5:9")
+    del compiler_rt_as
```
trws: Is the result of this that we get the same patches in the same cases? It looks like it but I haven't checked through the llvm commit output to be completely sure.

var/spack/repos/builtin/packages/llvm/package.py
```diff
@@ -673,23 +761,20 @@ def cmake_args(self):
...
-            if self.spec.satisfies("@15.0.0:"):
-                runtimes.append("compiler-rt")
-            else:
-                projects.append("compiler-rt")
-        if "+libcxx" in spec:
-            if self.spec.satisfies("@15.0.0:"):
-                runtimes.extend(["libcxx", "libcxxabi"])
-            else:
-                projects.extend(["libcxx", "libcxxabi"])
+        if "compiler-rt=runtime" in spec:
```
trws, line 834: If we can standardize on this style, it might be nice to loop over keys in the spec with a given property, and add them to runtimes or projects based on their value.  Loving this refactor overall if we can make it fly. 👍

trws:
Also, if you'd like to be a maintainer, I think @haampie and I would welcome the help. There may be some significant changes coming down soon to try and unify the llvm building support across more parts of spack, and having another pair of eyes would be fantastic.

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: I am sorry. Is there a feature `alias` or should I simply add a new variant? Does it make much sense, given that the possible values of the variant change anyway from `True/False` to `"none"/"project"/"runtime"`? Regarding things that depend on this. Do you mean in the Spack repo? I think I have updated the corresponding packages that depend on `llvm`. Or do you mean that there are people who use `develop` in production and expect their configuration files to be compatible with it at all times? I personally dropped this idea 5 years ago.

I don't mind creating an alias, of course, I just want the variants to be consistent with the [upstream names](https://github.com/llvm/llvm-project/blob/f458d9f6f892c3765174ad8f0cc912d930329f43/runtimes/CMakeLists.txt#L18) to make it easier to understand what is what. I would also replace `omp` with ~~`libomp`~~ `openmp` but didn't want to touch `omp_debug` and `omp_tsan` too.

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: I agree. It's just for simple packages we usually have `variant`s, then `depends_on`s and then `conflicts`, which is not necessary in general, as far as I know, and makes the code significantly less readable in this case.

var/spack/repos/builtin/packages/llvm/package.py
trws: I mean in the spack repo and in other downstream package repos, we've had significant breakage and complaints from previous changes of variant names without a transition period, which is what I'm trying to avoid.  The problem is actually the people that don't live on develop, they want a release with the variant marked as deprecated before it is removed in the following one, basically.

We do not currently have an alias feature, though supposedly that's in the works, it would be a separate variant, probably with a default dependent on the value of the existing one for now and set up such that they can't disagree.  Annoying I know, but more kind to stable deployments.

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: ~~I guess I should give a summary of the patch modifications:~~

see https://github.com/spack/spack/pull/32476#issuecomment-1237151890

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: Ok. I will then go further and make it `openmp`, `openmp_tsan` and `openmp_debug`.

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: Could you, please, give an example of how to "loop over keys in the spec with a given property"? I thought of refactoring it but the problem was that the variant `omp` adds `openmp` and the variant `libcxx` adds `libcxx` and `libcxxabi`. Currently we'd have something like:
```python
for var, params in {
    "omp": ["openmp"],
    "libunwind": ["libunwind"],
    "libcxx": ["libcxx", "libcxxabi"],
    "compiler-rt": ["compiler-rt"],
}.items():
    val = spec.variants[var].value
    if val == "project":
        projects.extend(params)
    elif val == "runtime":
        runtimes.extend(params)
```
Does it really make sense?

skosukhin:
> Also, if you'd like to be a maintainer, I think @haampie and I would welcome the help.

I have already added myself to the list (see [42d01d8](https://github.com/spack/spack/pull/32476/commits/42d01d8ab3dfeb2035d16708b364ffc0f9b3e6f9)). Is there more to do for now regarding that?

var/spack/repos/builtin/packages/llvm/package.py
skosukhin: Good question. Do you mean that we could have `+libunwind` (or even `+internal_unwind`) and decide on how to build the component based on the version we build and whether `+clang` is set? I thought of it and have to admit that I don't have bulletproof arguments against that.

However, I generally try not to limit the users if possible: if a package can be built in a certain configuration and its developers do not say "don't do that", I don't want to make the decision for the users. Also, `llvm` is heavily used as just a set of libraries (without `clang`) and the users might want them to be built with the same compiler they use to compile their projects.

skosukhin: Two more things here:
1. For me, `+clang libunwind=runtime` and `~clang libunwind=project` read better then just `+clang` and `~clang` because I would have to read `package.py` to understand whether the library that I need is built with the compiler that I want to use.
2. If we are talking about some backwards compatibility, all packages that had `llvm+internal_unwind` (`llvm~internal_unwind`) as a dependency can simply go for `llvm libunwind=project` (`llvm libunwind=none`) and don't be afraid that suddenly `libunwind` is built as a runtime.

skosukhin:
Summary of the patch modifications:

1. `libomp-libflags-as-list.patch` - same contents fetched from upstream, the old condition was `@3.7:14`, the new one is `@3.7:14 openmp=project` (I could not reproduce the failure either with `openmp=project` or with `openmp=runtime` but the comment for the patch said `make libflags a list in openmp subproject when ~omp_as_runtime`, therefore I assume it is not needed when `openmp=runtime`)
2. `llvm14-hwloc-ompd.patch` - replaced with a "proper" fix from upstream, the old condition was `@14`, the new one is `@14 omp=runtime` (I have managed to reproduce the failure when `openmp=runtime` but `openmp=project` works without the patch).
3. `constexpr_longdouble.patch` and `constexpr_longdouble_9.0.patch` - same conditions (the old one was `+libcxx`, the new ones are `libcxx=project` and `libcxx=runtime`), same contents fetched from upstream.
4. `llvm5-sanitizer-ustat.patch` - same conditions (the old one was `+compiler-rt`, the new ones are `compiler-rt=project` and `compiler-rt=runtime`), same contents fetched from upstream.
5. `sanitizer-ipc_perm_mode.patch` - replaced with a "proper" fix from upstream but cannot be fetched since the names of the patched files have changed, the old conditions were `@5:7+compiler-rt%clang@11:` and `@5:9+compiler-rt%gcc@9:`, the new ones are `@5:9 compiler-rt=project` and `@5:9 compiler-rt=runtime` (the fixed issue has to do with the version of `glibc` and not the versions of the compiler).
6. `llvm_python_path.patch` - same contents fetched from upstream, the old condition was `@:11`, the new one is `@7:11` (the patched logic was added in `7.0.0`).
7. `xray_buffer_queue-cstddef.patch` - same contents, fixed comments why the patch is needed, no identical commit upstream (nothing to fetch), the old condition was `@4:5+compiler-rt%clang@7:`, the new one is `@4:5 compiler-rt=project` (`compiler-rt=runtime` is not possible when `@:5` and the issue is related to the implementation of the C++ standard library, not the compiler version).
8. `lldb_external_ncurses-10.patch` - same contents fetched from upstream, the old condition was `@10.0.0:11.0.1+lldb`, the new one is `@10:11+lldb` (other `11.x` versions need it too).
9. `llvm_py37.patch` - same contents fetched from upstream, the old condition was `@4:6 ^python@3.7:`, the new one is `@4:6+python+lldb ^python@3.7:` (the patch is not needed when either `~python` or `~lldb`).
10. `llvm5-lld-ELF-Symbols.patch` - same conditions, same contents fetched from the FreeBSD repo (where the patch initially comes from)
11. `llvm4-lld-ELF-Symbols.patch` - split into three parts: two parts are fetched from the FreeBSD repo (where the patch originally comes from) and applied under the same condition `@4+lld%clang@6:`, the third one is a duplicate of `llvm_gcc7.patch` (see below).
12. `llvm_gcc7.patch` - renamed to `llvm4-lldb-Utility-TaskPool.patch`, the old conditions were `@4.0.0:4.0.1+lldb %gcc@7.0:` and `@4+lld%clang@6:` (as part of `llvm4-lld-ELF-Symbols.patch`, see above), the new one is `@4+lldb` (the issue is related to the implementation of the C++ standard library, not the compiler version, and seems to be irrelevant for `+lld~lldb`.
13. `missing-includes.patch` - split into two parts: one part is fetched from upstream and applied under the new confition `@8:9.0.0` (was `@8`), the second part is a duplicate of `llvm-gcc11.patch` (see below).
14. `llvm-gcc11.patch` - same contents fetched from upstream, the old conditions were `@9:11%gcc@11:`, `@8` (as part of `missing-includes.patch`, see above) and `@8:11` (as yet another patch fetched from upstream, see [8bc9eea](https://github.com/spack/spack/pull/32476/commits/8bc9eeaf1c0080a5a47f58596724949eb83379e8)), the new one is `@8:11`.
15. `no_cyclades.patch` - same conditions, same contents fetched from upstream
16. `no_cyclades9.patch` - updated comments, cannot be fetched since the names and formatting of the patched files have changed, the old condition was `@6:9`, the new one is `@3.5.1:9` (the patch is applicable at least starting version `3.5.1`, the oldest we try to support).

There seem to be no patches that could be fetched from a reliable source anymore.

skosukhin:
@trws I have also renamed variants `omp`, `omp_tsan` and `omp_debug` to `openmp`, `openmp_tsan` and `openmp_debug`, respectively. I have added aliases `omp_debug`, `internal_unwind`, `omp_tsan` and `omp_as_runtime` (see [37021ef](https://github.com/spack/spack/pull/32476/commits/37021ef45fa53ea81fc69eb316a97c9e4d8229c1)). Two questions now:
1. What should we do with `libcxx` and `compiler-rt`? The names of the variants are the same but cannot be set with `+/~` anymore, e.g. `+libcxx~compiler-rt`. If we cannot solve this problem and make everything backward compatible, should we bother at all?
2. How do we warn the users that they use deprecated variants? There seem to be no parameter `deprecated` for the `variant` directive. I also thought of something like:
    ```python
        variant(
            "omp_as_runtime",
            values=(True, False, "deprecated"),
            default="deprecated",
            sticky=True, ...
    ```
    and then check whether any of the deprecated variants have value other than `"deprecated"` and issue a warning but that seems to be too much for Spack to handle.

**UPD:** The following does not seem to work either:
```python
variant(
    "libcxx",
    values=disjoint_sets(
        # Deprecated values:
        (True, False),
        # New values:
        ("none",),
        ("project", conditional("runtime", when="+clang @6:")),
    ).prohibit_empty_set().with_default("runtime"), ...
```

trws:
@haampie, how do you feel about this? I'd like to take it, but am still on the fence on how we should deal with variant breakage.  For packages in tree they're all fixed here at least.

skosukhin:
It's sad to see that quite some work goes to `/dev/null`.

naromero77:
Told Todd I would help look at some PRs. A lot of work was put into this PR -- the hold-up was variant breakage? Are we not allowed to break previous variants? I get the point, but that can become a big constraint.

trws:
My apologies @skosukhin, and thanks for the extra poke @naromero77, we've been pretty buried lately.  Now that 0.20 is out the door, would you be willing to rebase/merge this and we'll figure out how to get it through?  Maybe since we're right at the start of a release cycle we can manage the variant breakage a bit better.  I'd really like to get your input on this though @haampie, it seems like a really good fix overall but it's big enough I'd like more eyes on it.

var/spack/repos/builtin/packages/llvm/package.py
naromero77: I think that I follow the variant name change. Its just for consistently with upstream and I think this is a win for us.

naromero77 (dismissed):
These are great changes... @skosukhin You please rebase this so we can merge it in soon.

skosukhin:
@ye-luo I'm giving this MR another try and going through the recent changes to the `llvm` package. Could you, please, explain the reason for [this conflict](https://github.com/spack/spack/blob/34b8fe827eb05c68ab627e955def4e7fc724a4ba/var/spack/repos/builtin/packages/llvm/package.py#L290), which you introduced in #36675? If I comment it out, `llvm+libomptarget~lld` builds just fine. Or do I miss something?

ye-luo:
> @ye-luo I'm giving this MR another try and going through the recent changes to the `llvm` package. Could you, please, explain the reason for [this conflict](https://github.com/spack/spack/blob/34b8fe827eb05c68ab627e955def4e7fc724a4ba/var/spack/repos/builtin/packages/llvm/package.py#L290), which you introduced in #36675? If I comment it out, `llvm+libomptarget~lld` builds just fine. Or do I miss something?

lld is needed when using OpenMP offload (libomptarget) to AMD GPUs. You should see an error when compiling the user code not during the compilation of llvm.

naromero77 (approved):
@trws Do you have any additional comments about this PR?
@skosukhin Could you please test a couple of configuration with and with OpenMP target offload on your local desktop?

naromero77:
@spackbot help

naromero77:
@spackbot re-run pipeline

PR #32481, opened 2022-09-01
[CURL] New version.

t-brown:
Adding a new version of curl. This addresses issue [9081](https://github.com/curl/curl/issues/9081).


tldahlgren (approved):
Confirmed sha256.

PR #32480, opened 2022-09-01
Docs: Getting Started Dependencies

ax3l:
Finally document what one needs to install to use Spack on Linux and Mac :-)

With <3 for minimal container users (e.g., issues like #32479) and my colleagues with their fancy Macs.
Screenshot from 2022-09-01 12-03-42

PR #32482, opened 2022-09-01
Add: py-sphinx-design

ax3l:
Needed for #32480

### DIFF
--- a/var/spack/repos/builtin/packages/py-sphinx-design/package.py
+++ b/var/spack/repos/builtin/packages/py-sphinx-design/package.py
@@ -19,4 +19,3 @@ class PySphinxDesign(PythonPackage):
     depends_on("python@3.7:", type=("build", "run"))
     depends_on("py-flit-core@3.4:3", type=("build"))
     depends_on("py-sphinx@4:5", type=("build", "run"))
-
--- a/var/spack/repos/builtin/packages/py-sphinx-design/package.py
+++ b/var/spack/repos/builtin/packages/py-sphinx-design/package.py
@@ -1,0 +1,22 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PySphinxDesign(PythonPackage):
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-sphinx-design/package.py
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PySphinxDesign(PythonPackage):
+    """A sphinx extension for designing beautiful, screen-size responsive web components."""
+
+    homepage = "https://sphinx-design.readthedocs.io"
+    url = "https://github.com/executablebooks/sphinx-design/archive/refs/tags/v0.3.0.tar.gz"
```
adamjstewart: https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#pypi-vs-github

var/spack/repos/builtin/packages/py-sphinx-design/package.py
```diff
@@ -0,0 +1,23 @@
...
+    git = "https://github.com/executablebooks/sphinx-design.git"
```
ax3l:
```suggestion
    git = "https://github.com/executablebooks/sphinx-design.git"
    list_url = "https://pypi.org/simple/sphinx-design/"
```

adamjstewart: This isn't needed if you use the `pypi` attribute instead of `url`.

var/spack/repos/builtin/packages/py-sphinx-design/package.py
ax3l:
```suggestion
    pypi = "sphinx-design/sphinx_design-0.3.0.tar.gz"
```

adamjstewart: It uses tar.gz. Also, the checksums will likely change.

adamjstewart:
@spackbot fix style

ax3l:
@spackbot fix style

adamjstewart (approved):

adamjstewart:
> I've updated the branch with isort fixes.

@scottwittenburg this should say "black fixes", not "isort fixes".

PR #32487, opened 2022-09-01
Add (stand-alone) test failure issue template

tldahlgren:
Failures of stand-alone tests in Spack packages aren't package build failures or bugs in Spack core so those templates aren't that useful.

This PR adds a template for reporting package stand-alone test failures and includes the ability to reference CDash outputs.


tldahlgren:
"Interesting" that a unit test failed for a template-only change.

alalazo (approved):

PR #32509, opened 2022-09-03
root: add JSON conflict

sethrj:
See https://github.com/root-project/root/issues/11312 .  @pcanal reports this fixed in the current development version of ROOT.  I saw that a patch already exists for nljson@3.11, but it seems to be for a separate issue. I'm curious what configuration of `root@6.26.06 ^nlohmann-json@3.11` successfully builds...


sethrj:
Arg, I see now that @graeme-a-stewart already patched this in #32472 . :(

pcanal:
>  I saw that a patch already exists for nljson@3.11, but it seems to be for a separate issue

it is.  The 3.11 series of release also had some issues with the header file organization.

PR #32510, opened 2022-09-03
cuba: new package

wdconinc:
Cuba, https://feynarts.de/cuba/, is a library that offers a choice of four independent routines for multidimensional numerical integration: Vegas, Suave, Divonne, and Cuhre.

It is used as a dependency for several high energy and nuclear physics Monte Carlo event generators.

The Makefile build does not succeed when run in parallel.


wdconinc:
@vvolkl Do you have any mcegs that depend on Cuba? Mostly QCD community stuff.

vvolkl:
> @vvolkl Do you have any mcegs that depend on Cuba? Mostly QCD community stuff.

@wdconinc No we don't - but thanks for upstreaming it!

alalazo (approved):

PR #32516, opened 2022-09-05
bugfix: package hash should affect process, dag, and dunder hashes

tgamblin:
This fixes a bug where two installations that differ only by package hash will not show up in `spack find`.

The bug arose because `_cmp_node` on `Spec` didn't include the package hash in its yielded fields. So, any two `Spec` objects that were only different by package hash would appear to be equal and would overwrite each other when inserted into the same `dict`. Note that we could still *install* specs with different package hashes, and they would appear in the database, but we code that needed to put them into data structures that use `__hash__` would have issues.

This PR makes `Spec.__hash__` and `Spec.__eq__` include the `process_hash()`, and it makes `Spec._cmp_node` include the package hash. All of these *should* include all information in a spec so that we don't end up in a situation where we are blind to particular field differences.

Eventually, we should unify the `_cmp_*` methods with `to_node_dict` so there aren't two sources of truth, but this needs some thought, since the `_cmp_*` methods exist for speed. We should benchmark whether it's really worth having two types of hashing now that we use `json` instad of `yaml` for spec hashing.

- [x] Add `package_hash` to `Spec._cmp_node`
- [x] Add `package_hash` to `spack.solve.asp.spec_clauses` so that the `package_hash` will show up in `spack diff`.
- [x] Add `package_hash` to the `process_hash` (which doesn't affect abstract specs but will make concrete specs correct)
- [x] Make `_cmp_iter` report the dag_hash so that no two specs with different process hashes will be considered equal.

### DIFF
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -4056,6 +4056,9 @@ class Spec(object):
         yield self.compiler_flags
         yield self.architecture
 
+        # this is not present on older specs
+        yield getattr(self, "_package_hash", None)
+
     def eq_node(self, other):
         """Equality with another spec, not including dependencies."""
         return (other is not None) and lang.lazy_eq(self._cmp_node, other._cmp_node)
@@ -4064,6 +4067,16 @@ class Spec(object):
         """Lazily yield components of self for comparison."""
         for item in self._cmp_node():
             yield item
+
+        # This needs to be in _cmp_iter so that no specs with different process hashes
+        # are considered the same by `__hash__` or `__eq__`.
+        #
+        # TODO: We should eventually unify the `_cmp_*` methods with `to_node_dict` so
+        # TODO: there aren't two sources of truth, but this needs some thought, since
+        # TODO: they exist for speed.  We should benchmark whether it's really worth
+        # TODO: having two types of hashing now that we use `json` instead of `yaml` for
+        # TODO: spec hashing.
+        yield self.process_hash() if self.concrete else None
 
         def deps():
             for dep in sorted(itertools.chain.from_iterable(self._dependencies.values())):
--- a/lib/spack/spack/hash_types.py
+++ b/lib/spack/spack/hash_types.py
@@ -44,7 +44,7 @@ dag_hash = SpecHashDescriptor(deptype=("build", "link", "run"), package_hash=Tru
 
 #: Hash descriptor used only to transfer a DAG, as is, across processes
 process_hash = SpecHashDescriptor(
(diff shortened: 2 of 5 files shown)


lib/spack/spack/spec.py
```diff
@@ -4065,6 +4068,16 @@ def _cmp_iter(self):
         for item in self._cmp_node():
             yield item
 
+        # This needs to be in _cmp_iter so that no specs with different process hashes
+        # are considered the same by `__hash__` or `__eq__`.
+        #
+        # TODO: We should eventually unify the `_cmp_*` methods with `to_node_dict` so
+        # TODO: there aren't two sources of truth, but this needs some thought, since
+        # TODO: they exist for speed.  We should benchmark whether it's really worth
+        # TODO: having two types of hashing now that we use `json` instad of `yaml` for
```
tldahlgren:
```suggestion
        # TODO: having two types of hashing now that we use `json` instead of `yaml` for
```

tldahlgren (changes requested):
Minor requested change.  (Still want to test something out.)

lib/spack/spack/spec.py
```diff
@@ -4065,6 +4068,16 @@ def _cmp_iter(self):
         for item in self._cmp_node():
             yield item
 
+        # This needs to be in _cmp_iter so that no specs with different process hashes
+        # are considered the same by `__hash__` or `__eq__`.
+        #
+        # TODO: We should eventually unify the `_cmp_*` methods with `to_node_dict` so
+        # TODO: there aren't two sources of truth, but this needs some thought, since
+        # TODO: they exist for speed.  We should benchmark whether it's really worth
```
trws, line 4076: Yes, and whether caching the result of to_node_dict is better or worse than using the generator.

trws (dismissed):

tldahlgren:
> Minor requested change. (Still want to test something out.)

LGTM *except* for the suggested comment typo.

lib/spack/spack/spec.py
tgamblin: done

tgamblin:
@tldahlgren @trws ready for review again

tldahlgren (approved):
LGTM

haampie:
This PR makes `Spec(...) < Spec(...)` non-deterministic for abstract specs (over time).

Basically `<` compares the roots in a sensible way, and if equal, compares hashes. The latter is arbitrary by design, and also awkwardly depends on the contents of package.py, which may change over time.

PR #30627, opened 2022-05-11
Boost: add conditional variants

hainest:
Fixes #1486
Fixes #11467
Closes #31237
Fixes #30663

TODO
- [x] Remove `all_libs`.
    I _think_ this can be done with something like `[v.lib for v in self.variants if v.enabled]` and then call `self.libs()` in `install` (at `for lib in Boost.all_libs:` around line 692).
- [x] Check version ranges for the other variants
- [x] `boost@master` does not build on develop.
    I believe this is due to the `if not with_libs` check in `install`. I have not verified this yet. I will post build error messages later.
- [x] There are many b2 [properties/features](https://www.boost.org/doc/libs/1_79_0/tools/build/doc/html/index.html#bbv2.tutorial.properties) we don't support. See #30496

@trws @alalazo @eugeneswalker @wspear @eschnett @asarkar-parsys

PR #32517, opened 2022-09-05
Boost: Add constraints for library variants based on version available

aeropl3b:
Disable library variants for versions that do not support them.


hainest:
These conflicts are tricky because some libraries have versions where they are/were header-only. These conflicts are also ignored by the semi-maintained filtering step in [`install`](https://github.com/spack/spack/pull/30627/commits/3afdfb74acb47aaceb90dfc59a26a32af4a79aca), so those need to be removed before the version checks will work.

Both of these items are covered in #30627. I just need to find some cycles to finish that up. Help is welcome!

alalazo:
Closing, since there has been no activity in a while. Feel free to reopen if you want to continue work on this. Also, it seems this may be incorporated in #30627.

PR #32518, opened 2022-09-05
arm: Add version 22.0.2 and rename package from arm to acfl

annop-w:
- Add version 22.0.2
- Rename package from arm to acfl
- Add annop-w as maintainer
- Modify default url to point to version 22.0.2

### DIFF
--- a/var/spack/repos/builtin/packages/acfl/package.py
+++ b/var/spack/repos/builtin/packages/acfl/package.py
@@ -17,32 +17,32 @@ _os_map = {
 
 
 _versions = {
-    '22.0.2': {
-        'RHEL-7': (
-            'e4dec577ed2d33124a556ba05584fad45a9acf6e13dccadb37b521d1bad5a826',
-            'https://developer.arm.com/-/media/Files/downloads/hpc/arm-allinea-studio/22-0-2/arm-compiler-for-linux_22.0.2_RHEL-7_aarch64.tar'
+    "22.0.2": {
+        "RHEL-7": (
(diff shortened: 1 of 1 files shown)


alalazo:
@spackbot fix style

annop-w:
@alalazo Thank you for fixing the style. Is it now ready to be merged ?

alalazo:
I'm waiting for a review from @OliverPerks Can you explain why this has been renamed?[^1]

[^1]: Renaming packages has been causing issues to people moving across branches with git, so usually we try to avoid that if there isn't a compelling reason.

annop-w:
The main reason is we simply would like the package name to reflect the actual package installed, which is the Arm Compiler for Linux (ACfL) package. ACfL also installs Arm Performance Libraries (ArmPL), which we would like to expose later (in a separate commit). There are currently 2 versions of ArmPL: one compiled with Arm compiler and one compiled with GNU compiler. The version with GNU compiler can be installed via the armpl-gcc package in Spack.

I discussed with @OliverPerks some time ago and we seemed to agree with the above. Does this make sense ?

annop-w:
@alalazo Seems like Oliver is not available. Can somebody else review this ?

alalazo (approved):

alalazo:
@annop-w Merged, this has been sitting here long enough :wink:

annop-w:
@alalazo Thank you. I wanted this merged because we now have new releases for ACfL and ArmPL. So, two new PRs coming in soon 😉

alalazo:
@annop-w Feel free to ping me on those, thanks!

Issue #32408, opened 2022-08-27
Installation issue:  fides ~vtk-m fails b/c it still looks for vtk-m in the spec


cyrush:
### Steps to reproduce the issue

@caitlinross  @dpugmire

I don't have the reproducer wrapped up with a bow, but see below error. 
I expect any build with fides~vtk-m will cause this issue.

I can also give you a docker container if that helps. 

### Error message

See below giant error for specific details:

```
==> fides: Executing phase: 'cmake'
[...]

PR #32521, opened 2022-09-05
update fides recipe

cyrush:
Two fixes:

- Related CMake options were not dependent on spack variants. 
- I tried to make them dependent, but Fides did not build w/o VTK-m

Based on this, I removed the variants for VTK-m and Adios2  -- I don't think these are optional? (It would be great to get a confirmation)

- We require VTK-m 1.8, Fides was restricted to 1.7 or lower 
- I did a test build with VTK-m 1.8 and I was able to compile.

Based on this, I removed the`:1.7.0` version constraint on vtk-m


cyrush:
resolves https://github.com/spack/spack/issues/32408

caitlinross:
@cyrush yes, both ADIOS2 and VTK-m are required for Fides.  The changes look good to me.

cyrush:
@caitlinross thanks for confirming!

alalazo (approved):

Issue #32470, opened 2022-09-01
`spack unit-test`: recent test fails randomly


scheibelp:
### Steps to reproduce

Rerun `SPACK_PYTHON=python2 SPACK_TEST_SOLVER=clingo spack unit-test -k "installed_externals_are_reused"`

It will fail sometimes and succeed other times. This means that all PRs have a random chance of failing CI.

* The test only actually runs when the concretizer is clingo, and otherwise skips
* I'm not yet sure if this also applies to Python 3 but will check that shortly and update here
  * Update: I get this failure about 20% of the time with Python 2 but got no failures in ~30 attempts with Python 3, so I think it is specific to Python 2

### Error message

When the test fails:

```
>       assert external2.dag_hash() != external1.dag_hash()
E       AssertionError: assert 'g2vslm4o3q5wru32m7akp4523tgsyhlt' != 'g2vslm4o3q5wru32m7akp4523tgsyhlt'
E        +  where 'g2vslm4o3q5wru32m7akp4523tgsyhlt' = <bound method Spec.dag_hash of changing@1.0%gcc@4.5.0+fee+foo arch=test-debian6-core2>()
E        +    where <bound method Spec.dag_hash of changing@1.0%gcc@4.5.0+fee+foo arch=test-debian6-core2> = changing@1.0%gcc@4.5.0+fee+foo arch=test-debian6-core2.dag_hash
E        +  and   'g2vslm4o3q5wru32m7akp4523tgsyhlt' = <bound method Spec.dag_hash of changing@1.0%gcc@4.5.0+fee+foo arch=test-debian6-core2>()
E        +    where <bound method Spec.dag_hash of changing@1.0%gcc@4.5.0+fee+foo arch=test-debian6-core2> = changing@1.0%gcc@4.5.0+fee+foo arch=test-debian6-core2.dag_hash
```

### Information on your system

https://github.com/spack/spack/commit/0c6e3188ac7ca780783bb36c0763c069cc0c1043 adds a test called `test_installed_externals_are_reused`

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

scheibelp:
@alalazo @becker33 FYI as the writer/merger of the PR

alalazo:
Will investigate this, the only weird thing I notice is that:
```console
[...]

PR #32526, opened 2022-09-06
Mark a test xfail on Python 2.7

alalazo:
refers #32470


alalazo:
@scheibelp FYI, added this as a mitigation for the issue you opened since it seems a problem with the test fixture.

PR #32528, opened 2022-09-06
Allow target requirements in packages.yaml

### LINKED ISSUE #32471: Package requirements: target=... not respected.
### Steps to reproduce

This works/concretizes
```yaml
spack:
  specs: [ccache target=x86_64]
```

This does not work:

```yaml
spack:
  specs: [ccache]
  packages:
    all:
      require: target=x86_64
```

### Error message

```
$ spack -e . concretize 
==> Starting concretization
==> Error: Cannot satisfy requirement group for package 'ccache'
```

alalazo:
fixes #32471 

This PR solves the issue reported in #32471 specifically for targets and operating systems, by avoiding to add a default platform to anonymous specs. The issue still stands if users specify:
```yaml
packages:
  all:  
    require: "platform=linux"
```
but that should be a much rarer case. To solve it we need to modify the encoding in `concretize.lp`. Breaks a few tests for the original concretizer that were relying on this parsing bug.

Modifications:
- [x] Remove code in the parser that add `platform` even if a platform is not present
- [x] Add unit test for the specific case in #32471 
- [x] Skip tests that broke with the original concretizer

### DIFF
--- a/lib/spack/spack/test/architecture.py
+++ b/lib/spack/spack/test/architecture.py
@@ -199,6 +199,9 @@ def test_satisfy_strict_constraint_when_not_concrete(architecture_tuple, constra
 def test_concretize_target_ranges(root_target_range, dep_target_range, result):
     # use foobar=bar to make the problem simpler for the old concretizer
     # the new concretizer should not need that help
+    if spack.config.get("config:concretizer") == "original":
+        pytest.skip("Fixing the parser broke this test for the original concretizer.")
+
     spec_str = "a %%gcc@10 foobar=bar target=%s ^b target=%s" % (
         root_target_range,
         dep_target_range,
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -2,7 +2,6 @@
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
-
 """
 Spack allows very fine-grained control over how packages are installed and
 over how they are built and configured.  To make this easy, it has its own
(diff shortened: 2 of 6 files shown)


lib/spack/spack/test/architecture.py
```diff
@@ -199,6 +199,9 @@ def test_satisfy_strict_constraint_when_not_concrete(architecture_tuple, constra
 def test_concretize_target_ranges(root_target_range, dep_target_range, result):
     # use foobar=bar to make the problem simpler for the old concretizer
     # the new concretizer should not need that help
+    if spack.config.get("config:concretizer") == "original":
```
tldahlgren, line 202: Any reason not to parameterize (`pytest.mark.parametrize`) this (and tests below) to override the setting and test with `original` and `clingo`?

tldahlgren (commented):
Have a question at this point.

lib/spack/spack/test/architecture.py
alalazo: For `pytest.parametrize`, I can look into refactoring this test. but un this PR I just went with the minimal change I had to do to have CI pass.

> test with original and clingo

Maybe I wasn't super clear in the description. This fix actually breaks some behavior for the original concretizer. The behavior that is broken relies on the bug that is fixed here (i.e. if you parse a spec literal with just `target=XX` the spec also contains the platform). If we accept that we can break that behavior in the original concretizer, then the fix is removing code. If we want to fix the original concretizer for the case where it breaks (target ranges in some situations) we need to go back tweaking our greedy algorithm (which might take a while, at least for me).

tldahlgren: > Maybe I wasn't super clear in the description. This fix actually breaks some behavior for the original concretizer. The behavior that is broken relies on the bug that is fixed here (i.e. if you parse a spec literal with just `target=XX` the spec also contains the platform). If we accept that we can break that behavior in the original concretizer, then the fix is removing code. If we want to fix the original concretizer for the case where it breaks (target ranges in some situations) we need to go back tweaking our greedy algorithm (which might take a while, at least for me).

You were.  I thought it could be useful to have the tests run regardless of the concretizer settings so that both concretizers get exercised by the tests, even with the original getting skipped.

alalazo: > I thought it could be useful to have the tests run regardless of the concretizer settings so that both concretizers get exercised by the tests, even with the original getting skipped.

We run two jobs in CI, one tests with the original concretizer on the other with clingo. I agree it would be better if we can switch seamlessly and have one job - but that requires a big refactoring of code due to our use of globals.

haampie:
Should we merge this?

haampie (dismissed):

tgamblin (dismissed):

haampie (approved):

PR #32532, opened 2022-09-06
new package: routinator

aweits:


alalazo (approved):
Merging since the only failure is "No specs to rebuild"

PR #32535, opened 2022-09-06
Add new and legacy intel-mpi-benchmarks versions

### LINKED ISSUE #32534: intel-mpi-benchmarks is outdated and incomplete
### Summary

Currently, the available version of intel-mpi-benchmarks range from 2019.2 till 2019.6.
https://github.com/spack/spack/blob/v0.18.1/var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py#L22-L26

Intel has released much newer versions (latest being the 2021.3)
https://github.com/intel/mpi-benchmarks

Could you update the package.py including not only the latest versions but also all the former ones?

### Rationale

_No response_

### Description

_No response_

### Additional information

_No response_

### General information

- [x] I have run `spack --version` and reported the version of Spack

carsonwoods:
This is a very small PR to add new versions to the intel-mpi-benchmarks package at the request of #32534. This PR also adds legacy versions that are not currently included in the package file. 

Closes #32534.


adamjstewart:
You'll probably want to add a `url_for_version` function to simplify the changes to the URL string over releases.

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,31 +18,52 @@ class IntelMpiBenchmarks(MakefilePackage):
...
+            "mt",
+            "all",
+            conditional("p2p", when="@2019.2:"),
+        ),
         multi=False,
         description="Specify which benchmark to build",
     )
 
+    def url_for_version(self, version):
+        if version[0] <= 2019 and version[1] <= 1:
```
adamjstewart:
```suggestion
        if version <= Version("2019.1"):
```
This should be a bit more clear

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,31 +18,52 @@ class IntelMpiBenchmarks(MakefilePackage):
     """
 
     homepage = "https://software.intel.com/en-us/articles/intel-mpi-benchmarks"
-    url = "https://github.com/intel/mpi-benchmarks/archive/IMB-v2019.5.tar.gz"
```
adamjstewart, line 21: Can you keep the `url` here? `url_for_version` is used for downloads, but `url` is used to search for new versions. So things like `spack versions` and `spack checksum` only work if there is some `url` available. You can just use the newer URL format, people usually only care about adding new releases.

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
carsonwoods: Thanks, good idea.

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,31 +18,53 @@ class IntelMpiBenchmarks(MakefilePackage):
...
+    version("2019.1", sha256="fe0d065b9936b6943ea83cb3d00aede43b17565285c6b1791fee8e340853ef79")
+    version("2019.0", sha256="1c7d44aa7fd86ca84ac7cae1a69a8426243048d6294582337f1de7b4ffe68d37")
+    version("2018.1", sha256="718a4eb155f18cf15a736f6496332407b5837cf1f19831723d4cfe5266c43507")
+    version("2018.0", sha256="2e60a9894a686a95791be2227bc569bf81ca3875421b5307df7d83f885b1de88")
 
     depends_on("mpi")
 
     # https://github.com/intel/mpi-benchmarks/pull/19
-    patch("add_const.patch", when="@:2019.6")
+    patch("add_const.patch", when="@2019.0:2019.6")
```
adamjstewart: You could use `@2019` which is equivalent to `@2019.0:2019.999`

carsonwoods: Sorry that is actually an artifact of my debugging. I'm not actually sure the patch is good for that range. I'm actually building it right now to make sure the patch works for that version. I think its actually only good from 2019.1 onward, but I'm making sure of that right now.

carsonwoods: Ok, the first patch works for the entire 2019 range. The second patch doesn't. That should now be reflected in the code.

adamjstewart: Did you forget to push?

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,31 +18,53 @@ class IntelMpiBenchmarks(MakefilePackage):
...
+    version("2018.0", sha256="2e60a9894a686a95791be2227bc569bf81ca3875421b5307df7d83f885b1de88")
 
     depends_on("mpi")
 
     # https://github.com/intel/mpi-benchmarks/pull/19
-    patch("add_const.patch", when="@:2019.6")
+    patch("add_const.patch", when="@2019")
     # https://github.com/intel/mpi-benchmarks/pull/20
-    patch("reorder_benchmark_macros.patch", when="@:2019.6")
+    patch("reorder_benchmark_macros.patch", when="@2019.0:2019.6")
```
adamjstewart: You said this patch doesn't work for the entire 2019 range?

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
carsonwoods: Whoops yes sorry. I thought I had 😅. Please try again!

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
adamjstewart: Btw, if you want to test this:
```console
$ for ver in $(spack versions --safe intel-mpi-benchmarks); do spack patch intel-mpi-benchmarks@$ver; done
```

carsonwoods: I thought it might be causing a build failure, but it was something else unrelated to the patch on my system. I corrected my previous statement.
> Ok, the first patch works for the entire 2019 range. The second patch doesn't. That should now be reflected in the code.

Thank you for that command. I will run that now to be sure and will follow up here.

carsonwoods: Sorry for the confusion. Yes, that patch doesn't work on that version. It being included was an artifact of my testing. It wasn't intentionally included. I've fixed it now.

adamjstewart (dismissed):
Haven't tested, but looks good from a Spack perspective. @m4r1k can you test this and make sure it has everything you need?

carsonwoods:
@adamjstewart It looks like there is also a soft requirement in older versions of the benchmarks which requires intel-mpi instead of openmpi. This might just be a bug with openmpi, but building via intel-mpi fixes it. I'm testing using mpich right now to see if that would also work, but I might add that requirement (instead of the virtual dependency) for older versions.

carsonwoods:
Ok, the build issue was using OpenMPI, so I've added a conflict to prevent it from being selected on older versions. If @m4r1k agrees, it can be merged now.

adamjstewart (dismissed):

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,31 +18,63 @@ class IntelMpiBenchmarks(MakefilePackage):
...
+    depends_on("gmake", when="@2018")
 
-    depends_on("mpi")
+    conflicts(
+        "^openmpi",
+        when="@:2019.0",
+        msg="intel-mpi-benchmarks <= v2019.0 cannot be built with OpenMPI, "
+        "please specify a different MPI implementation",
+    )
+    conflicts("benchmark=p2p", when="@2018", msg="p2p benchmark available on <= v2019.0 only")
```
adamjstewart: Isn't this what the `conditional()` below is designed for? E.g., `@2018 benchmark=p2p` isn't even a valid input, it will never get to the concretization phase so it will never hit this conflict.

carsonwoods: Yes that is correct. I wasn't sure what was the *correct* way to represent this so I did both. I'll remove the conflicts if the `conditional()` is sufficient.

adamjstewart: It should be sufficient as far as I know. Try it. If it doesn't work, that's a bug.

carsonwoods: It does work, but I think there is a bug in the error reported. The message I wrote isn't displayed. Instead I get this concretization error:
```console
$ spack spec intel-mpi-benchmarks@2018.1 benchmark=p2p
==> Error: No version for 'intel-mpi-benchmarks' satisfies '@2018.1' and '@2021.3'
```

adamjstewart: @becker33 was working on improving those error messages, he may be interested in this.

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,31 +18,67 @@ class IntelMpiBenchmarks(MakefilePackage):
...
+            "mt",
+            "all",
+            conditional("p2p", when="@2019.2:"),
+        ),
         multi=False,
         description="Specify which benchmark to build",
     )
 
+    @run_before("build")
+    def set_build_dir(self):
```
adamjstewart: I think this was right before

carsonwoods: @adamjstewart I'm unable to get the build directory set just right. It seems to be ignoring whatever I try. I'm trying to make it so that runs the `make` command within a `src` subdirectory only when the package version satisfies `@2018`. Is there an idiomatic way to do this? 

I tried even temporarily getting rid of my version requirement and it didn't seem to make a difference. It still just tried to install as though nothing were set.

adamjstewart: Oh, that's because you override the `build` phase. The default runs:
```python
    make(self.build_targets)
```
So you could either modify your `build` and `install` phases to use `with working_dir` or replace them with changes to `build_targets` and `install_targets`

carsonwoods: Ok great. You can hold off on reviewing this PR for a while probably. I'm going to re-factor it to use `build_targets` but right now it isn't working. I will try and get it working by the end of the weekend or by Monday, but I'm not sure how much I'll be able to do tonight and it currently isn't worth reviewing. Thanks for the help.

carsonwoods:
@adamjstewart I found a bit more time to work on this tonight and I think its in a good spot now. I am able to install the old and new versions without any issues as far as I can tell. Unless @m4r1k finds more issues, I think it can be reviewed and merged if it looks good. 

Thanks for your repeated help and reviews on this one 😃.

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,74 +18,108 @@ class IntelMpiBenchmarks(MakefilePackage):
...
-            install("IMB-EXT", prefix.bin)
-            install("IMB-IO", prefix.bin)
-            install("IMB-MPI1", prefix.bin)
-            install("IMB-MT", prefix.bin)
-            install("IMB-NBC", prefix.bin)
-            install("IMB-P2P", prefix.bin)
-            install("IMB-RMA", prefix.bin)
+            build_directory = "."
+
+        with working_dir(build_directory):
```
adamjstewart: Use self.build_directory and you can drop the duplicate if statement.

carsonwoods: Ok cool, thank you for the suggestion. I've made the change.

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -18,74 +18,103 @@ class IntelMpiBenchmarks(MakefilePackage):
...
+    depends_on("intel-mpi", when="@2018")
+    depends_on("gmake", type="build", when="@2018")
 
-    depends_on("mpi")
+    conflicts(
+        "^openmpi",
+        when="@:2019.0",
+        msg="intel-mpi-benchmarks <= v2019.0 cannot be built with OpenMPI, "
+        "please specify a different MPI implementation",
+    )
```
adamjstewart, line 46: So if I'm understanding correctly, 2018 works with intel-mpi, 2019.0 works with all MPI except openmpi, and 2019.1+ works with all MPI? (obviously you haven't tested all possible MPI providers, but you know what I mean)

adamjstewart (approved):
Looks good although I haven't tested it. Will give @m4r1k a chance to test before I merge.

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
carsonwoods: Yes that seems to be the case. I suspect that some more elaborate modifications to the build process could enable other MPI implementations on older versions, however documentation for the package is unclear so I'm sticking with what works out of the box.

Issue #29695, opened 2022-03-23
Better control of how package and dependencies are installed


scottwittenburg:
### Summary

Currently `spack install` only has the following options for deciding when to use buildcache for package installation: `--cache-only` and `--no-cache` apply both to installation of the package and its dependencies.  It would be nice to have finer control over when buildcache is used during package installation, i.e. to allow specification of whether buildcache should be used for package and/or dependencies, and whether it should be considered an error if that cannot be accomplished.

### Rationale

Some use cases for this functionality include:

1. Fail gitlab pipelines when a job ends up having to build dependencies from source
2. Signal to gitlab pipeline that entire stack should be rebuilt, regardless of whether pkgs are up to date or not (to test core spack changes that don't change spec hashes).  Even when entire stack should be rebuilt from source, each gitlab job should still install its dependencies from the buildcache, as those should have been built in previous jobs of the pipeline.

### Description

In order to support this idea, options for the UI were discussed on slack awhile ago, but that discussion has since expired from the history.  This is an attempt to re-capture that discussion from notes, so the ideas are preserved until they can be implemented.

The slack discussion resulted in somewhat of a consensus around the name a new `spack install` option: `--use-buildcache`.  The values that option could accept were:

```
only|never|auto|package:only|never|auto,dependencies:only|never|auto
```

The default value would be simply `auto`, which would apply to both package and dependencies, and would mean allowing installation from buildcache but not insisting on it.  Supplying simply `only|never|auto` would similarly apply to both package and dependencies.   

The option values also allow for specifying different behavior for packages vs dependencies.  For example to tell spack to install the package from source only, but install dependencies from buildcache only, and fail if not able to do those things, we could say:

```
spack install --use-buildcache=package:never,dependencies:only
```

### Additional information

This isn't specific to any version of spack or any os/arch, but rather a new command line option available everywhere.

### General information

- [X] I have run `spack --version` and reported the version of Spack
- [X] I have searched the issues of this repo and believe this is not a duplicate

scottwittenburg:
@tgamblin @becker33 @haampie I think all of you were involved in this discussion at the time, please feel free to correct my memory where you think appropriate.

@alalazo This is how I would see us accomplishing what you described on slack a bit ago (re: forcing gitlab to rebuild everything from source, even when hashes didn't change).

tgamblin:
Fixed by #32537

PR #32537, opened 2022-09-06
Install: Add use-buildcache option to install

kwryankrattiger:
Allow differentiating between top level packages and dependencies when determining whether to install from the cache or not.

@scottwittenburg 

Implements this [feature request](https://github.com/spack/spack/issues/29695).

### DIFF
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -5,6 +5,7 @@
 
 import argparse
 import os
+import re
 import shutil
 import sys
 import textwrap
@@ -31,10 +32,64 @@ section = "build"
 level = "short"
 
 
+def parse_use_buildcache(args):
+    bc_parsed = {}
+    bc_keys = ["package", "dependencies", ""]
+    bc_values = ["only", "never", "auto"]
+    kv_list = re.findall("([a-z]+)?:?([a-z]+)", args.use_buildcache)
+
+    # Verify keys and values
+    bc_map = {k: v for k, v in kv_list if k in bc_keys and v in bc_values}
+    if not len(kv_list) == len(bc_map):
+        unrecognized_args = [(k, v) for k, v in kv_list if k not in bc_keys or v not in bc_values]
+        tty.error("Unrecognized arguments passed to use-buildcache: %s" % unrecognized_args)
+        tty.error(
+            "Expected: --use-buildcache "
+            ":{[auto|only|never],[[package:[auto|only|never]],[dependencies:[auto|only|never]]]"
+        )
+        exit(1)
+        return None
+
+    package_use_buildcache = "auto"
+    dependencies_use_buildcache = "auto"
+    if "" in bc_map:
+        package_use_buildcache = bc_map[""]
+        dependencies_use_buildcache = bc_map[""]
+    if "package" in bc_map:
+        package_use_buildcache = bc_map["package"]
+    if "dependencies" in bc_map:
+        dependencies_use_buildcache = bc_map["dependencies"]
+
+    if package_use_buildcache == "auto":
+        bc_parsed["package_use_cache"] = args.use_cache
+        bc_parsed["package_cache_only"] = args.cache_only
+    elif package_use_buildcache == "only":
+        bc_parsed["package_use_cache"] = True
+        bc_parsed["package_cache_only"] = True
+    elif package_use_buildcache == "never":
+        bc_parsed["package_use_cache"] = False
+        bc_parsed["package_cache_only"] = False
+
+    if dependencies_use_buildcache == "auto":
+        bc_parsed["dependencies_use_cache"] = args.use_cache
+        bc_parsed["dependencies_cache_only"] = args.cache_only
+    elif dependencies_use_buildcache == "only":
+        bc_parsed["dependencies_use_cache"] = True
+        bc_parsed["dependencies_cache_only"] = True
+    elif dependencies_use_buildcache == "never":
+        bc_parsed["dependencies_use_cache"] = False
+        bc_parsed["dependencies_cache_only"] = False
+
+    return bc_parsed
+
+
 def install_kwargs_from_args(args):
     """Translate command line arguments into a dictionary that will be passed
     to the package installer.
     """
+
+    cache_ops = parse_use_buildcache(args)
+
     return {
         "fail_fast": args.fail_fast,
         "keep_prefix": args.keep_prefix,
@@ -44,8 +99,10 @@ def install_kwargs_from_args(args):
         "verbose": args.verbose or args.install_verbose,
         "fake": args.fake,
         "dirty": args.dirty,
-        "use_cache": args.use_cache,
-        "cache_only": args.cache_only,
+        "package_use_cache": cache_ops["package_use_cache"],
+        "package_cache_only": cache_ops["package_cache_only"],
+        "dependencies_use_cache": cache_ops["dependencies_use_cache"],
+        "dependencies_cache_only": cache_ops["dependencies_cache_only"],
         "include_build_deps": args.include_build_deps,
         "explicit": True,  # Use true as a default for install command
         "stop_at": args.until,
@@ -122,6 +179,12 @@ the dependencies""",
         dest="cache_only",
         default=False,
         help="only install package from binary mirrors",
+    )
+    cache_group.add_argument(
+        "--use-buildcache",
+        dest="use_buildcache",
+        default="package:auto,dependencies:auto",
+        help="finer grain control over which packages to check mirrors for",
     )
 
     subparser.add_argument(
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -34,9 +34,9 @@ level = "short"
 
 def parse_use_buildcache(args):
     bc_parsed = {}
-    bc_keys = ["package", "dependencies", ""]
+    bc_keys = ["package:", "dependencies:", ""]
     bc_values = ["only", "never", "auto"]
-    kv_list = re.findall("([a-z]+)?:?([a-z]+)", args.use_buildcache)
+    kv_list = re.findall("([a-z]+:)?([a-z]+)", args)
 
     # Verify keys and values
     bc_map = {k: v for k, v in kv_list if k in bc_keys and v in bc_values}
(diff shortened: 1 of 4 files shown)


kwryankrattiger:
@spackbot fix style

scottwittenburg:
@kwryankrattiger Can you rebase this on the latest `develop` and force-push?  I'm hoping it will then get past the failing checks we're seeing currently.

scottwittenburg:
@spackbot fix style

lib/spack/spack/cmd/install.py
```diff
@@ -31,10 +32,64 @@
...
+        )
+        exit(1)
+        return None
+
+    package_use_buildcache = "auto"
+    dependencies_use_buildcache = "auto"
+    if "" in bc_map:
+        package_use_buildcache = bc_map[""]
+        dependencies_use_buildcache = bc_map[""]
+    if "package" in bc_map:
```
kwryankrattiger: This will permit passing `--use-buildcache auto,package:never` and use the auto for dependencies. It is of a shorthand because I for one never type dependencies correctly and it is a long keyword.

lib/spack/spack/cmd/install.py
```diff
@@ -31,10 +32,64 @@
...
+    bc_map = {k: v for k, v in kv_list if k in bc_keys and v in bc_values}
+    if not len(kv_list) == len(bc_map):
+        unrecognized_args = [(k, v) for k, v in kv_list if k not in bc_keys or v not in bc_values]
+        tty.error("Unrecognized arguments passed to use-buildcache: %s" % unrecognized_args)
+        tty.error(
+            "Expected: --use-buildcache "
+            ":{[auto|only|never],[[package:[auto|only|never]],[dependencies:[auto|only|never]]]"
+        )
+        exit(1)
+        return None
```
scottwittenburg: Given the `exit(1)` just above this, I wouldn't think this `return` statement could ever be executed.

lib/spack/spack/cmd/install.py
scottwittenburg: Is the idea here that you could also say `--use-buildcache auto,dependencies:never`?  I think it's weird if it only works the other way.

lib/spack/spack/cmd/install.py
```diff
@@ -31,10 +32,64 @@
 level = "short"
 
 
+def parse_use_buildcache(args):
+    bc_parsed = {}
+    bc_keys = ["package", "dependencies", ""]
+    bc_values = ["only", "never", "auto"]
+    kv_list = re.findall("([a-z]+)?:?([a-z]+)", args.use_buildcache)
```
scottwittenburg: I'm not sure this regex is doing what we want.  If I say:

```
spack install --no-check-signature --use-buildcache auto readline
```

then spack errors with:

```
==> Error: Unrecognized arguments passed to use-buildcache: [('aut', 'o')]
==> Error: Expected: --use-buildcache :{[auto|only|never],[[package:[auto|only|never]],[dependencies:[auto|only|never]]]
```

scottwittenburg (changes requested):
Thanks @kwryankrattiger,!  This is looking pretty good, but I left a comment or two to to be addressed, along with a question about how the shorthand is intended to work. 

Also, could you provide a brief description of the change?  Even just a pointer to the issue this addresses could help other reviewers know what they're looking at

lib/spack/spack/cmd/install.py
kwryankrattiger: ah, that isn't supposed to exit(1). I wasn't sure how to handle errors correctly with parsing.

lib/spack/spack/cmd/install.py
kwryankrattiger: Both ways work, also not listing a "default" and only specifying one works.

so...
`--use-buildcache package=never` is the same as `--use-buildcache auto,package=never` or `--use-buildcache dependencies=auto,package=never`

or you could do
`--use-buildcache dependencies=only` which is the same as `--use-buildcache auto,dependencies=only` or `--use-buildcache package=auto,dependencies=only`

and so on. The short is, by default the option is "auto", or it is the specified default (ie. specified in `--use-buildcache` without a prefix), or it is overridden by a prefixed option, ie. `package:only` or `dependencies:never` etc.

I left it open to allow some flexibility on how to specify the option based on the way the user is reasoning on what they want. Not sure if that is correct or not, trying to toe the line between overly restrictive and annoying to use.

lib/spack/spack/cmd/install.py
kwryankrattiger: I love regex! That is supposed to evaluate to `("", "auto")`, which it does for `never` and `only`...

kwryankrattiger: `([a-z]+:)?([a-z]+)` works, but requires a more complicated filtering

scottwittenburg:
I don't mind running the style fix for you, but you should at least pull the changes and incorporate them before you force push again :stuck_out_tongue_winking_eye:

scottwittenburg:
@spackbot fix style

lib/spack/spack/cmd/install.py
scottwittenburg: Normally, from spack commands, you can just return the value you want to be used as the exit code.  Probably worth verifying that value it makes it all the way back to the shell though.  Unless your test method already validates the exit code, in which case it will let you know.

lib/spack/spack/cmd/install.py
```diff
@@ -31,10 +32,61 @@
 level = "short"
 
 
+def parse_use_buildcache(args):
```
scottwittenburg: Can you add a docstring saying what `args` is?  In a lot of other places in spack commands, that variable name is used for the parsed args object, so I was sort of surprised it was just the value 0f the `--use-buildcache` option.  Also, the docstring should probably briefly describe the data structure returned to the caller.

lib/spack/spack/cmd/install.py
```diff
@@ -123,6 +177,12 @@ def setup_parser(subparser):
         default=False,
         help="only install package from binary mirrors",
     )
+    cache_group.add_argument(
+        "--use-buildcache",
+        dest="use_buildcache",
+        default="package:auto,dependencies:auto",
+        help="finer grain control over which packages to check mirrors for",
```
scottwittenburg: Is this the right place to describe a little more to the user about the expected format?  The `default` value does provide an example, but the values `only`, `never`, and `auto` are hidden from the argument parser in the `parse_use_buildcache()` method.  It's also probably worth mentioning (here, if this is the right place) the shorthand versions allowed.

scottwittenburg: Maybe also worth mentioning that `auto` means "fall back to the (maybe default) value of` either `--use-cache` or `--cache-only`.

lib/spack/spack/cmd/install.py
```diff
@@ -31,10 +32,61 @@
...
+    bc_parsed = {}
+    bc_keys = ["package:", "dependencies:", ""]
+    bc_values = ["only", "never", "auto"]
+    kv_list = re.findall("([a-z]+:)?([a-z]+)", args)
+
+    # Verify keys and values
+    bc_map = {k: v for k, v in kv_list if k in bc_keys and v in bc_values}
+    if not len(kv_list) == len(bc_map):
+        unrecognized_args = [(k, v) for k, v in kv_list if k not in bc_keys or v not in bc_values]
+        tty.error("Unrecognized arguments passed to use-buildcache: %s" % unrecognized_args)
```
scottwittenburg: I kind of don't like the way the list of string tuples gets printed, e.g.:

```
$ spack install --use-buildcache always readline
==> Error: Unrecognized arguments passed to use-buildcache: [('', 'always')]
```

What do you think of omitting that bit from this particular line?   Since you print out the expected values below, it seems possible to discern what the problem is.

scottwittenburg (changes requested):
This is getting closer, thanks @kwryankrattiger!  It's working for my particular use case (must install deps from buildcache, never install package from buildcache), but I have a couple  requests regarding the UI.

Also, it looks like some unit tests failed, which this may or may not have caused.  Can you check if those failures are reproducible on your end?

scottwittenburg:
@spackbot fix style

alalazo:
I have a high-level question. Failed pipeline jobs let you run the build in a docker image to inspect the failure. Within that Docker image there's an `install.sh` script which runs a single command to install the spec that failed. An example `install.sh` script is, for instance:
```bash
#!/bin/bash


# spack install command
"spack" "-d" "-v" "install" "--show-log-on-error" "--keep-stage" "--no-check-signature" "--cdash-upload-url" "https://cdash.spack.io/submit.php?project=Spack+Testing" "--cdash-build" "unifyfs@1.0%oneapi@2022.1.0 hash=36ftnpast6yvr3zawwnrcgxn5ao3t5zs arch=linux-ubuntu20.04-x86_64 (E4S OneAPI)" "--cdash-site" "Cloud Gitlab Infrastructure (uo-pegasus)" "--cdash-buildstamp" "20220919-0545-E4S OneAPI" "--no-add" "-f" "/builds/spack/spack/jobs_scratch_dir/reproduction/unifyfs.json"
```

I assume (let me know if this is a wrong assumption) that something similar to running `install.sh` happens in Gitlab. If the assumption is correct, why we discarded splitting a single `spack install` command into:
```console
$ spack install --cache-only --only=dependencies <spec>
$ spack install --no-cache --only=root <spec>
```
to ensure that dependencies are never installed from sources? Is there any technical issue that I overlooked proceeding that way?

scottwittenburg:
@alalazo We used to split it in two, but it complicated the ci module even more, as we had to make a copy of the `spack.yaml` before the first install, then use it to replace the version that spack clobbered before the next spack install.  

Also, the `install.sh` script you mentioned in your question, which ends up in the reproducer docker container, is the same one that gitlab runs in the pipeline.  Or did I misunderstand the question?

scottwittenburg:
By the way, in my pipelines testing "rebuild everything", this PR is working as expected.  @tgamblin was going to provide some insight on a better way, within the installer.py module, to distinguish dependencies from package.  But from my perspective, @kwryankrattiger figured that out in a way that seems reasonable to me.  So I'd be happy to accept this as-is, once the tests are passing.

alalazo:
> We used to split it in two, but it complicated the ci module even more, as we had to make a copy of the spack.yaml before the first install, then use it to replace the version that spack clobbered before the next spack install.

I don't think I get what you're describing above, which might be due to me being unfamiliar with `ci.py` and `cmd/ci.py`. Isn't this:

https://github.com/spack/spack/blob/0ddbb92ae3acad8b2b87198d127aa96ea95f48c0/lib/spack/spack/ci.py#L1985-L2005

the only part that eventually need to change? Based on the example above wouldn't it suffice to generate something similar to:
```bash
#!/bin/bash


# spack install command
"spack" "-d" "-v" "install" "--cache-only" "--only=dependencies" "--no-add" "-f" "/builds/spack/spack/jobs_scratch_dir/reproduction/unifyfs.json"
"spack" "-d" "-v" "install" "--show-log-on-error" "--no-cache" "--only=package" "--keep-stage" "--no-check-signature" "--cdash-upload-url" "https://cdash.spack.io/submit.php?project=Spack+Testing" "--cdash-build" "unifyfs@1.0%oneapi@2022.1.0 hash=36ftnpast6yvr3zawwnrcgxn5ao3t5zs arch=linux-ubuntu20.04-x86_64 (E4S OneAPI)" "--cdash-site" "Cloud Gitlab Infrastructure (uo-pegasus)" "--cdash-buildstamp" "20220919-0545-E4S OneAPI" "--no-add" "-f" "/builds/spack/spack/jobs_scratch_dir/reproduction/unifyfs.json"
```
?

scottwittenburg:
That might work, without having to make a backup copy of the environment and restore it between the installs, now that we have the `--no-add` option.  However, I would need to go back and make sure there wasn't some additional problem with the "two spack installs" approach.

scottwittenburg (approved):
I tested this with the other proposed changes we need for asking spackbot to rebuild everything, and it's working.

tgamblin (approved):

haampie (commented):
Seems like you're missing

```diff
diff --git a/lib/spack/spack/installer.py b/lib/spack/spack/installer.py
index 7ec31f0267..1ec21be45d 100644
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -2391,7 +2391,7 @@ def get_deptypes(self, pkg):
         """
         deptypes = ["link", "run"]
         include_build_deps = self.install_args.get("include_build_deps")
-        if not self.install_args.get("cache_only") or include_build_deps:
+        if not self.install_args.get("package_cache_only") or include_build_deps:
             deptypes.append("build")
         if self.run_tests(pkg):
             deptypes.append("test")
```
an old reference to `cache_only`, which now causes the installer not to prune redundant build deps.

PR #32539, opened 2022-09-06
Update/set CRTM version in jcsda-emc-bundle env packages

climbfuji:
Because there are different versions of CRTM for EMC and JCSDA, we need to explicitly set the version in the various jcsda-emc-bundle env packages.

This PR is a pre-requisite for and tested with **MISSING**


climbfuji:
Sorry, wrong target.

PR #32543, opened 2022-09-07
dtc:fix build error with clang or Fujitsu compiler

QianJianhua1:
An error occurred due to the '-werror' option when built with the Clang or Fujitsu compiler.
"error: -Wl,--disable-new-dtags: 'linker' input unused [-Werror,-Wunused-command-line-argument]"


alalazo:
@spackbot fix style

alalazo (approved):

PR #32547, opened 2022-09-07
CMakePackage: allow custom CMAKE_PREFIX_PATH entries

scheibelp:
Closes https://github.com/spack/spack/pull/31631

Users can update the CMAKE_PREFIX_PATH environment variable in `setup_dependent_build_environment`. However, CMakePackage passes a CMAKE_PREFIX_PATH option to `cmake` that does not account for these changes; the option overrides the environment variable.

https://github.com/spack/spack/pull/31631 reorganizes the code to make use of this, but reuses functions that are not necessarily idempotent.

This PR allows a package to define a `cmake_search_paths` attribute, which is then included in the CMAKE_PREFIX_PATH option to cmake.

Note: this PR also retains the existing modifications to the CMAKE_PREFIX_PATH environment variable in `build_environment.py` because not all CMake-dependent packages inherit from `CMakePackage`.


lib/spack/spack/build_environment.py
```diff
@@ -1030,8 +1030,15 @@ def get_cmake_prefix_path(pkg):
...
     ordered_build_link_deps = spack_built + externals
-    build_link_prefixes = filter_system_paths(x.prefix for x in ordered_build_link_deps)
-    return build_link_prefixes
+    cmake_prefix_path_entries = []
+    for spec in ordered_build_link_deps:
+        cmake_prefix_path_entries.append(spec.prefix)
+        try:
+            cmake_prefix_path_entries.extend(spec.package.cmake_search_paths)
+        except AttributeError:
+            pass
```
becker33: If you add a property to `PackageBase`

```
@property
def cmake_prefiix_paths(self):
    return [self.prefix]
```

then this code can be simplified to

```suggestion
    cmake_prefix_path_entries = []
    for spec in ordered_build_link_deps:
        cmake_prefix_path_entries.append(spec.package.cmake_prefix_paths)
```

var/spack/repos/builtin/packages/cuda/package.py
```diff
@@ -465,6 +465,13 @@ def setup_build_environment(self, env):
     def setup_dependent_build_environment(self, env, dependent_spec):
         env.set("CUDAHOSTCXX", dependent_spec.package.compiler.cxx)
 
+    @property
+    def cmake_search_paths(self):
+        if self.spec.satisfies("target=x86_64:"):
+            cub_path = self.prefix.targets + "/x86_64-linux/lib/cmake"
+            return [cub_path]
+        return []
```
becker33: If you make the changes suggested above, then this would need to be
```suggestion
        cmake_prefix_paths = [self.prefix]
        if self.spec.satisfies("target=x86_64:"):
            cub_path = self.prefix.targets + "/x86_64-linux/lib/cmake"
            cmake_prefix_paths.append(cub_path)
        return cmake_prefix_paths
```

becker33 (changes requested):
I think providing a default method on `PackageBase` will make this simpler.

var/spack/repos/builtin/packages/cuda/package.py
scheibelp: I was actually worried about taking this approach because I don't want to require that the user remembers to add `self.prefix`. I assume that (a) it never hurts to add this and (b) if it's generally needed that the user would be annoyed having to remember it every time.

If that doesn't seem like a problem to you, I can make this change.

becker33: I think it's easier for the user to understand "list the cmake prefix paths needed to findd the package" than "list the cmake prefix paths needed to run this package, except the one we've decided is obvious".

robertu94: @becker33 we already have a few magic defaults like this.  For example configure_args or cmake_args adds the `--prefix` or `-DCMAKE_INSTALL_PREFIX_PATH` option by default.  I think it better for this function to return a list of additional cmake_prefix_paths rather than the footgun of forgetting the default.

robertu94: To be fair, as far as I know for most packages, the default does the right thing.  I imagine there are others (also probably closed source packages developed by Nvidia), but we wouldn't be adding this functionality now (as opposed to years ago) if more packages need this to work properly so I don't want to bikeshed getting this fixed for the small number of packages that actually need this. 

tl;dr I'm fine either way

scheibelp:
@becker33 I have made the requested changes and added a test

becker33 (approved):

robertu94:
@scheibelp and @becker33 thank you both for helping this get merged!

PR #32549, opened 2022-09-07
e4s: use package requirements instead of preferences

alalazo:
Modifications:
- [x] Require `intel-tbb` as a `tbb` provider, `openblas` as a `blas` provider and `mpich` as an `mpi` provider


alalazo:
@eugeneswalker While trying to experiment with requirements, I made trilinos fail due to a conflict with variants and the version of CUDA used. Is this a good info, or was that expected and known?

For reference the error says:
```
Error: trilinos '+cuda +stokhos' conflicts with '^cuda@11.6:'
```

alalazo:
Outdated, and stale.

PR #24264, opened 2021-06-11
py-pyprecice: Fix checksum of 2.2.0.1 and 2.2.1.1

ajaust:
This PR fixes the sha256 checksums of `pyprecice` 2.2.0.1 and 2.2.1.1.

Adding @BenjaminRodenberg and @IshaanDesai for reference.

PR #14002, opened 2019-12-05
Add "same_version_as"-directive to fix py-protobuf dependencies

obreitwi:
### Problem:
Each version of `py-protobuf` depends on the exact same version of `protobuf`. Tracking this kind of dependency is only possible with a lot of boilerplate, which we aim to avoid.

### Solution:
* Add `same_version_as`-directive that automatically adds a one-to-one dependency of each version of the package to the exact same version of the other package.

* Use `same_version_as`-directive to keep versions of py-protobuf and protobuf in sync (i.e., `py-protobuf@2.6.1` depends on `protobuf@2.6.1`, `py-protobuf@.3.7.1` depends on `protobuf@3.7.1` and so forth).

* Add mock packages used in tests to ensure `same_version_as`-directive remains functional.

PR #32551, opened 2022-09-07
[py-pyprecice] add v2.5.0.0 and v2.5.0.1 and update checksum of v2.4.0.0

IshaanDesai:
This PR adds the two latest releases of the pyprecice package v2.5.0.0 and v2.5.0.1.


var/spack/repos/builtin/packages/py-pyprecice/package.py
```diff
@@ -33,6 +35,7 @@ class PyPyprecice(PythonPackage):
     version("2.0.0.1", sha256="96eafdf421ec61ad6fcf0ab1d3cf210831a815272984c470b2aea57d4d0c9e0e")
 
     depends_on("precice@develop", when="@develop")
+    depends_on("precice@2.5.0", when="@2.5.0.1:2.5.0")
```
adamjstewart: These `when` args look odd. As is, this says that `py-precice@2.5.0.0` does not depend on `precice` at all. I think a better way of organizing these would be something like:
```python
depends_on("precice@2.5.0", when="@2.5.0")
```
Here, `@2.5.0` implies `@2.5.0.0:2.5.0.999`. If you want to get really fancy, you could do this in a for loop:
```python
for ver in ["develop", "2.5.0", "2.4.0", ..., "2.0.0"]:
    depends_on("precice@" + ver, when="@" + ver)
```

var/spack/repos/builtin/packages/py-pyprecice/package.py
```diff
@@ -50,6 +53,7 @@ class PyPyprecice(PythonPackage):
     depends_on("py-cython@0.29:", type="build")
     depends_on("py-packaging", type="build")
     depends_on("py-pip@19.0.0:", type="build")
+    depends_on("py-pkgconfig", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-pkgconfig", type="build", when="@2.5:")
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pyprecice/package.py
IshaanDesai: Thanks for this, I made the change. However I am a bit confused about this now. If the `when` args were a problem for `2.5.0.0`, then how is it fine in the case of:
```python
depends_on("precice@2.4.0", when="@2.4.0.1:2.4.0")
```
is it because the version `2.4.0.1` does not exist and hence `2.4.0.0` is covered? I read up on the `when` args syntax in detail, but I guess I am still missing something.

fsimonis: @adamjstewart this still requires adding versions in two places.

An alternative would be a version wrapper:

```py
    def wrapper(v, **kwargs):
        version(v, **kwargs)
        depends_on("precice@"+v.rsplit(".", 1)[0], when="@"+v)

    # Always prefer final version of release candidate
    version("develop", branch="develop")

    wrapper('2.3.0.1', sha256='ed4e48729b662680beaa4ee2a9aff724a79e760534c6c58181be739988da2789')
```

It would be awesome if the following was possible:
```py
 for v in versions:
   depends_on("precice@"+v.rsplit(".", 1)[0], when="@"+v)
```

adamjstewart: > If the when args were a problem for 2.5.0.0, then how is it fine in the case of:

2.4.0.0 is also broken.

> It would be awesome if the following was possible:

#14002 was actually introduced for this very reason, although it got stuck.

var/spack/repos/builtin/packages/py-pyprecice/package.py
```diff
@@ -33,6 +35,7 @@ class PyPyprecice(PythonPackage):
     version("2.0.0.1", sha256="96eafdf421ec61ad6fcf0ab1d3cf210831a815272984c470b2aea57d4d0c9e0e")
 
     depends_on("precice@develop", when="@develop")
+    depends_on("precice@2.5.0", when="@2.5.0")
     depends_on("precice@2.4.0", when="@2.4.0.1:2.4.0")
```
adamjstewart: You'll also need to do this for all older versions too:
```suggestion
    depends_on("precice@2.4.0", when="@2.4.0")
```

IshaanDesai: Resolving this for all the past versions made me realize that doing a `for` loop as you previously mentioned is far more easier:

```python
    for ver in ["develop", "2.5.0", "2.4.0", "2.3.0", "2.2.1", "2.2.0", "2.1.1", "2.1.0", "2.0.2", "2.0.1", "2.0.0"]:
        depends_on("precice@" + ver, when="@" + ver)
```
I guess one could create that list in a more fancy way, but for now I would leave it at this.

adamjstewart:
@spackbot run pipeline

var/spack/repos/builtin/packages/py-pyprecice/package.py
```diff
@@ -19,7 +19,9 @@ class PyPyprecice(PythonPackage):
 
     # Always prefer final version of release candidate
     version("develop", branch="develop")
-    version("2.4.0.0", sha256="8877da85df97d66966892440c7e777b667d8e6a33747304ed3762f4c33b729ec")
+    version("2.5.0.1", sha256="e2602f828d4f907ea93e34f7d4adb8db086044a75a446592a4099423d56ed62c")
+    version("2.5.0.0", sha256="9f55a22594bb602cde8a5987217728569f16d9576ea53ed00497e9046a2e1794")
+    version("2.4.0.0", sha256="e80d16417b8ce1fdac80c988cb18ae1e16f785c5eb1035934d8b37ac18945242")
```
adamjstewart, line 24: Any idea what the reason for the checksum change is?

IshaanDesai: The change in the checksum was discussed before in https://github.com/spack/spack/pull/24264 and https://github.com/precice/python-bindings/issues/109. In short using the `versioneer` package to create the version is problematic for the `tar.gz` files as it stores references to branches. The solution which was decided from our side was to push an empty commit to master after releasing so that the version only points to a tag on GitHub and not to a branch. I simply forgot to push an empty commit after releasing `v2.4.0.0` and thus the checksum changed.

ajaust (approved):
To me this looks good.

adamjstewart (approved):

PR #32553, opened 2022-09-07
Spackage for mlperf's OpenCatalyst

rspavel:
Added initial spackage for the reference implementation for the MLPerf HPC OpenCatalyst DimeNet++ benchmark/proxy app.


var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+from spack.package import *
+
+
+class MlperfOpencatalyst(PythonPackage, CudaPackage):
```
adamjstewart: Personally I would name this `py-ocp-models` since that's the name used in https://github.com/mlcommons/hpc/blob/main/open_catalyst/setup.py

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
```diff
@@ -0,0 +1,40 @@
...
+
+    homepage = "https://opencatalystproject.org/"
+    git = "https://github.com/mlcommons/hpc.git"
+
+    version("main", branch="main")
+
+    tags = ["proxy-app"]
+
+    depends_on("python@3.8:", type=("build", "run"))
+    depends_on("py-ase@3.21:3.21.9999", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-ase@3.21", type=("build", "run"))
```
Same for other deps

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
```diff
@@ -0,0 +1,40 @@
...
+    depends_on("py-sphinx", type=("build", "run"))
+    depends_on("py-nbsphinx", type=("build", "run"))
+    depends_on("pandoc", type=("build", "run"))
+    depends_on("py-black", type=("build", "run"))
+    depends_on("py-torch+cuda", when="+cuda", type=("build", "run"))
+    depends_on("py-torch~cuda", when="~cuda", type=("build", "run"))
+    depends_on("py-torch-scatter+cuda", when="+cuda", type=("build", "run"))
+    depends_on("py-torch-scatter~cuda", when="~cuda", type=("build", "run"))
+    depends_on("py-torch-spline-conv+cuda", when="+cuda", type=("build", "run"))
+    depends_on("py-torch-spline-conv~cuda", when="~cuda", type=("build", "run"))
```
adamjstewart: I would remove `CudaPackage` since this package doesn't change how it's built if `+cuda` is requested. Dependencies should be controlled separately, either through `packages.yaml`:
```yaml
packages:
  all:
    variants: +cuda
```
or on the command line:
```console
$ spack install py-ocp-models ^py-torch+cuda
```

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
```diff
@@ -0,0 +1,40 @@
...
+
```
adamjstewart: Missing dependencies: https://github.com/mlcommons/hpc/blob/main/open_catalyst/env.common.yml#L22,L29

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
```diff
@@ -0,0 +1,40 @@
...
+    @property
+    def build_directory(self):
+        return "open_catalyst"
```
adamjstewart: This can be simplified:
```suggestion
    build_directory = "open_catalyst"
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
rspavel: Done

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
rspavel: Done

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
rspavel: That has bitten me countless times but following conventions: Done

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
rspavel: Good catch, done

var/spack/repos/builtin/packages/mlperf-opencatalyst/package.py
rspavel: Done

adamjstewart:
Did the last commit undo the changes in the previous commit?

rspavel:
@adamjstewart Good catch. Computers are hard (especially at the end of the FY).

var/spack/repos/builtin/packages/py-demjson/package.py
```diff
@@ -0,0 +1,27 @@
...
+    version("2.2.1", sha256="75055ebb6112a3c6bc3e009f528b209f9968773ebd3d3a1cd0d4b6c2006ee830")
+    version("2.2", sha256="2aaa78c4d6f08c1bbcb88babd18cb54bb3ba2446870088ca33581c69f603a3a2")
+    version("2.0.1", sha256="572c755df5aec7ac3fde5535795b8bf71f54e4681196c97a7a97d81916293504")
+    version("2.0", sha256="24f638daa0c28a9d44db2282d46ea3edfd4c7d11a656e38677b741620bf1483d")
+    version("1.6", sha256="1d989c310e33569ecc178b8182e53bde8f748bf5ea10cfbc0e331f8c313f6e29")
+    version("1.5", sha256="446f4a74ba95679fa2fbe31887beb0c9870b6b8b2706ed5798bbf91339e8c349")
+    version("1.4", sha256="e5858dc54a80290cecbc1d1514c6ce50dca44fbd15ccc195c4a8b6969b45a41f")
+
+    depends_on("py-setuptools", type="build")
+    depends_on("python@2.6:", type=("build", "run"), when="@:0.20")
```
adamjstewart: These versions don't exist in the package, I would remove this line

var/spack/repos/builtin/packages/py-ocp-models/package.py
```diff
@@ -0,0 +1,43 @@
...
+    """Reference implementation for the MLPerf HPC OpenCatalyst DimeNet++ benchmark"""
+
+    homepage = "https://opencatalystproject.org/"
+    git = "https://github.com/mlcommons/hpc.git"
+
+    version("main", branch="main")
+
+    tags = ["proxy-app"]
+
+    depends_on("python@3.8:", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("python@3.8", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-ocp-models/package.py
```diff
@@ -0,0 +1,43 @@
...
+    depends_on("py-torch-scatter", type=("build", "run"))
+    depends_on("py-torch-spline-conv", type=("build", "run"))
+    depends_on("py-demjson", type=("build", "run"))
+    depends_on("py-pillow", type=("build", "run"))
+    depends_on("py-torch-geometric", type=("build", "run"))
+    depends_on("py-wandb", type=("build", "run"))
+    depends_on("py-lmdb@1.1.1", type=("build", "run"))
+    depends_on("py-pytest@6.2.2", type=("build", "run"))
+    depends_on("py-submitit", type=("build", "run"))
+    depends_on("py-sphinx-rtd-theme", type=("build", "run"))
```
adamjstewart, line 43: Missing deps on torch-cluster and torch-sparse: https://github.com/mlcommons/hpc/blob/main/open_catalyst/env.cpu.yml

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-demjson/package.py
rspavel: Removed

var/spack/repos/builtin/packages/py-ocp-models/package.py
rspavel: Did spack specs and constraints change recently? I would think this would restrict it to only a specific version of python? But changed to match conventions

var/spack/repos/builtin/packages/py-ocp-models/package.py
rspavel: Good catch. Not entirely sure what conda merge was doing but just manually went through both yamls. If we aren't doing any gpu specialization at the ocp level then it simplifies things. Fixed

var/spack/repos/builtin/packages/py-ocp-models/package.py
adamjstewart: 3.8 means 3.8.*

This didn't used to work with the old concretizer, but it works fine with clingo now.

adamjstewart (approved):

var/spack/repos/builtin/packages/py-ocp-models/package.py
rspavel: Ah. Awesome. That will teach me for never reading the documentation

adamjstewart: It's poorly documented at best lol.

PR #32569, opened 2022-09-07
Skip test which fails randomly on Python 2.7

scheibelp:
Skip test which fails randomly when using Python 2.7 to run Spack.

See: https://github.com/spack/spack/issues/32470
This copies the approach used in https://github.com/spack/spack/pull/32526.

@alalazo based on the wording of the skip message, should this actually apply to all tests using the `repo_with_changing_recipe` fixture?

### DIFF
--- a/lib/spack/spack/test/concretize.py
+++ b/lib/spack/spack/test/concretize.py
@@ -1529,8 +1529,8 @@ class TestConcretize(object):
 
     @pytest.mark.regression("29201")
     @pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


alalazo:
> should this actually apply to all tests using the repo_with_changing_recipe fixture?

Not sure, I wasn't able to pin why it fails with that combination of Python version + `pytest`. One possible fix is to change the package in memory with monkeypatch - I just need to find time to fix it.

alalazo (approved):

PR #32572, opened 2022-09-08
Avoid `xxmr2d:out of memory` error for scalapack-2.2.0

DeeliN221:
In file `scalapack-2.2.0/REDIST/SRC/pgemraux.c`, change variable n type from `Int` to `long long`,  so that `xxmr2d:out of memory` can be avoid.

The created patch file was simply generated by:

$ diff `old_pgemraux.c` `new_pgemraux.c`

Further modification may be made to complete this fix.

### DIFF
--- a/var/spack/repos/builtin/packages/netlib-scalapack/mr2d_malloc-memory.patch
+++ b/var/spack/repos/builtin/packages/netlib-scalapack/mr2d_malloc-memory.patch
@@ -1,0 +1,4 @@
+108c108
+< mr2d_malloc(Int n)
+---
+> mr2d_malloc(long long n)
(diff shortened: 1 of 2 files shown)


alalazo:
@spackbot fix style

alalazo (changes requested):
There are flake 8 errors that our bot cannot fix. Could you take care of them?

DeeliN221:
> There are flake 8 errors that our bot cannot fix. Could you take care of them?

@alalazo Hi, I am not sure about the details, could you take a look and tell me what should I do to fix problems if there is any.

alalazo (commented):
There are failures in our build pipelines to be fixed as well, before this can be considered. For instance: https://gitlab.spack.io/spack/spack/-/pipelines/172901/failures

PR #32573, opened 2022-09-08
ci: avoid running coverage on package only PRs

alalazo:


tgamblin (dismissed):

alalazo:
Thanks for approving @tgamblin For reference, it seems required status checks that are skipped are considered as passed.

Screenshot from 2022-09-08 13-11-58

michaelkuhn (dismissed):

alalazo:
@spackbot run pipeline

trws (approved):
We have a few followup items for other PRs, but this is GTG.

PR #32574, opened 2022-09-08
py-cython: add 0.29.32

michaelkuhn:
This version is required by py-pandas@1.4.4. The missing version causes the concretizer to pick an old version that doesn't work.


adamjstewart:
> This version is required by py-pandas@1.4.4.

@alalazo why didn't the package audits catch this?

adamjstewart (approved):

PR #32576, opened 2022-09-08
Add whip package

msimberg:
Adds https://github.com/eth-cscs/whip as a package. No releases yet, hence only the main branch as a version.

### DIFF
--- a/var/spack/repos/builtin/packages/whip/package.py
+++ b/var/spack/repos/builtin/packages/whip/package.py
@@ -1,0 +1,30 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+from spack.package import *
+
+
+class Whip(CMakePackage, CudaPackage, ROCmPackage):
+    """whip is a small C++ abstraction layer for CUDA and HIP."""
+
+    homepage = "https://github.com/eth-cscs/whip/"
+    url = "https://github.com/eth-cscs/whip/archive/0.0.0.tar.gz"
+    git = "https://github.com/eth-cscs/whip.git"
+    maintainers = ["msimberg", "rasolca"]
+
+    version("main", branch="main")
+
+    depends_on("cmake@3.22:", type="build")
+
+    # Exactly one of +cuda and +rocm need to be set
+    conflicts("~cuda ~rocm")
+    conflicts("+cuda +rocm")
+
+    def cmake_args(self):
+        if self.spec.satisfies("+cuda"):
+            return [self.define("WHIP_BACKEND", "CUDA")]
+        else:
+            return [self.define("WHIP_BACKEND", "HIP")]


rasolca (dismissed):

rasolca (dismissed):

var/spack/repos/builtin/packages/whip/package.py
```diff
@@ -0,0 +1,27 @@
...
+    git = "https://github.com/eth-cscs/whip.git"
+    maintainers = ["msimberg", "rasolca"]
+
+    version("main", branch="main")
+
+    depends_on("cmake@3.22:", type="build")
+
+    variant("backend", values=("cuda", "hip"), default="cuda", description="The backend to wrap")
+    depends_on("cuda", when="backend=cuda")
+    depends_on("hip", when="backend=hip")
```
albestro: Would it be possible to have it in the more "spack-standard" way?

I mean, something along the lines of having `+cuda` / `+rocm` variants, with the logic for making them mutually exclusive, plus inheriting from `CUDAPackage` and `ROCmPackage` respectively?

msimberg: I had it like that initially and ended up at this exactly because I thought the spack standard way is a bit clunky in this case but I'm open to be convinced otherwise.

Pros of the current, non-standard way:
- The variant encodes directly the required behaviour, i.e. "exactly-one" (not zero, not both) without any custom conflicts
- `cuda_arch`/`amdgpu_target` are not required in whip itself since it doesn't build any device code

Pros of the standard way:
- It's more familiar

Let me know if I've missed some reasons to prefer one or the other.

rasolca: > cuda_arch/amdgpu_target are not required in whip

I think this might be a con and not a pro, as the target architecture can be easily propagated in the dependencies.

application +hip amdgpu_target=XXX --> whip +hip amdgpu_target=XXX --> hip amdgpu_target=XXX

This applies especially for the hip version as cuda libraries contains symbols for all archs anyway.

msimberg: I can understand neutral, but con? `amdgpu_target` isn't really propagated like that anyways, spack can happily end up concretizing `amdgpu_target` with different values unless one sets a global variant constraint in the `packages` config.

Another pro of the standard way though would be:
- packages can say `depends("whip +rocm", when="+rocm")` instead of `depends("whip backend=hip", when="+rocm")` which is not just more familiar but more regular.

msimberg: I pushed the standard way as well (comparison here: https://github.com/spack/spack/compare/750990ea8642a73a5211d9abdc7efb9756638ffc..5c6792d5ec93fc21eb8ce8c7f8d79d6168df6adc). I think I'd be happy with either way.

albestro: Yeah, I have nothing to add.

I think we can go with either, unless an authoritative spack maintainer suggests one way or the other 😉 is there any way to ask for a review from core maintainer? maybe we should ask in spack slack channel?

albestro (dismissed):

teonnik (dismissed):

var/spack/repos/builtin/packages/whip/package.py
```diff
@@ -0,0 +1,32 @@
...
+    conflicts("~cuda ~rocm")
+    conflicts("+cuda +rocm")
+
+    def cmake_args(self):
+        if self.spec.satisfies("+cuda"):
+            return [self.define("WHIP_BACKEND", "CUDA")]
+        elif self.spec.satisfies("+rocm"):
+            return [self.define("WHIP_BACKEND", "HIP")]
+        else:
+            raise InstallError("whip requires exactly one of +cuda and +rocm to be set")
```
alalazo: This cannot happen due to the conflicts above

alalazo (changes requested):

var/spack/repos/builtin/packages/whip/package.py
msimberg: I'm aware. However, I left it in there to make sure the conflicts aren't modified wrongly in the future. Perhaps too conservative? Happy to remove it though if you think it's unnecessary.

alalazo: Yeah, let's just remove it. Then I'm happy to merge this

msimberg: Removed and pushed.

alalazo (approved):

PR #32577, opened 2022-09-08
Add pika 0.8.0

msimberg:
- adds pika 0.8.0
- updates the cmake version requirement from 0.8.0 onwards
- constrains build with `hipcc` to only be applied for versions older than 0.8.0
- ~adds missing `hip` dependency for `+rocm` (it was always there implicitly because of hipblas/rocblas, but it should be an explicit dependency)~ (https://github.com/spack/spack/pull/32577#discussion_r966785597)
- constrains hip to at least 5.2.0 from pika 0.8.0 onwards

### DIFF
--- a/var/spack/repos/builtin/packages/pika/package.py
+++ b/var/spack/repos/builtin/packages/pika/package.py
@@ -17,6 +17,7 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
     git = "https://github.com/pika-org/pika.git"
     maintainers = ["msimberg", "albestro", "teonnik", "aurianer"]
 
+    version("0.8.0", sha256="058e82d7c8f95badabe52bbb4682d55aadf340d67ced1226c0673b4529adc182")
     version("0.7.0", sha256="e1bf978c88515f7af28ee47f98b795ffee521c15b39877ea4cfb405f31d507ed")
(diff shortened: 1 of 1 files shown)


aurianer (dismissed):
LGTM thanks!

var/spack/repos/builtin/packages/pika/package.py
```diff
@@ -84,6 +86,8 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("apex", when="+apex")
     depends_on("tracy-client", when="+tracy")
     depends_on("rocblas", when="+rocm")
+    depends_on("hip", when="+rocm")
```
albestro: This line is not needed since it is already enforced by the `ROCMPackage` from which pika inherits from.

https://github.com/spack/spack/blob/0d18c32bca8398e2daa736a940154caa235cbacc/lib/spack/spack/build_systems/rocm.py#L125

albestro (changes requested):

var/spack/repos/builtin/packages/pika/package.py
msimberg: Ah, that explains it better. Thanks @albestro for pointing this out! Removed.

albestro (approved):

teonnik (approved):

alalazo (approved):

PR #32579, opened 2022-09-08
new packages: py-arm-pyart and dependencies

glennpj:
- py-arm-pyart
- py-cylp
- rsl

### DIFF
--- a/var/spack/repos/builtin/packages/py-arm-pyart/StringIO.patch
+++ b/var/spack/repos/builtin/packages/py-arm-pyart/StringIO.patch
@@ -1,0 +1,30 @@
+diff -ru a/pyart/testing/data/make_small_mdv_ppi.py b/pyart/testing/data/make_small_mdv_ppi.py
+--- a/pyart/testing/data/make_small_mdv_ppi.py	2022-07-19 13:06:08.000000000 -0500
++++ b/pyart/testing/data/make_small_mdv_ppi.py	2022-09-10 14:26:11.021976219 -0500
+@@ -6,7 +6,10 @@
+ """
+ 
+ import struct
+-import StringIO
++try:
++    from StringIO import StringIO
++except ImportError:
++    from io import StringIO
+ import gzip
+ 
+ import pyart
+diff -ru a/pyart/testing/data/make_small_mdv_rhi.py b/pyart/testing/data/make_small_mdv_rhi.py
+--- a/pyart/testing/data/make_small_mdv_rhi.py	2022-07-19 13:06:08.000000000 -0500
++++ b/pyart/testing/data/make_small_mdv_rhi.py	2022-09-10 14:26:24.897985906 -0500
+@@ -6,7 +6,10 @@
+ """
+ 
+ import struct
+-import StringIO
++try:
++    from StringIO import StringIO
++except ImportError:
++    from io import StringIO
+ import gzip
+ 
+ import pyart
--- a/var/spack/repos/builtin/packages/py-arm-pyart/StringIO.patch
+++ b/var/spack/repos/builtin/packages/py-arm-pyart/StringIO.patch
@@ -1,6 +1,6 @@
 diff -ru a/pyart/testing/data/make_small_mdv_ppi.py b/pyart/testing/data/make_small_mdv_ppi.py
 --- a/pyart/testing/data/make_small_mdv_ppi.py	2022-07-19 13:06:08.000000000 -0500
-+++ b/pyart/testing/data/make_small_mdv_ppi.py	2022-09-10 14:26:11.021976219 -0500
++++ b/pyart/testing/data/make_small_mdv_ppi.py	2022-09-10 16:06:38.712279369 -0500
 @@ -6,7 +6,10 @@
  """
  
@@ -13,9 +13,18 @@ diff -ru a/pyart/testing/data/make_small_mdv_ppi.py b/pyart/testing/data/make_sm
  import gzip
  
  import pyart
+@@ -40,7 +43,7 @@
+ 
+ fdata_str = fdata.astype('uint16').byteswap().tostring()
+ uncompressed_data_size = len(fdata_str)
+-fileobj = StringIO.StringIO()
++fileobj = StringIO()
+ gzipfile = gzip.GzipFile(fileobj=fileobj, mode='w')
+ gzipfile.write(fdata_str)
+ gzipfile.close()
 diff -ru a/pyart/testing/data/make_small_mdv_rhi.py b/pyart/testing/data/make_small_mdv_rhi.py
 --- a/pyart/testing/data/make_small_mdv_rhi.py	2022-07-19 13:06:08.000000000 -0500
-+++ b/pyart/testing/data/make_small_mdv_rhi.py	2022-09-10 14:26:24.897985906 -0500
++++ b/pyart/testing/data/make_small_mdv_rhi.py	2022-09-10 16:07:12.775303046 -0500
 @@ -6,7 +6,10 @@
  """
  
@@ -28,3 +37,12 @@ diff -ru a/pyart/testing/data/make_small_mdv_rhi.py b/pyart/testing/data/make_sm
  import gzip
  
  import pyart
+@@ -39,7 +42,7 @@
+ 
+ fdata_str = fdata.astype('uint16').byteswap().tostring()
+ uncompressed_data_size = len(fdata_str)
+-fileobj = StringIO.StringIO()
++fileobj = StringIO()
+ gzipfile = gzip.GzipFile(fileobj=fileobj, mode='w')
+ gzipfile.write(fdata_str)
+ gzipfile.close()
--- a/var/spack/repos/builtin/packages/py-arm-pyart/package.py
+++ b/var/spack/repos/builtin/packages/py-arm-pyart/package.py
@@ -28,20 +28,16 @@ class PyArmPyart(PythonPackage):
     variant("hdf5", description="Support for HDF5 files", default=False)
     variant("rsl", description="Use RSL library", default=False)
     variant("wradlib", description="Calculate texture of differential phase field", default=False)
-    variant("xarray", description="Work with grid dataset", default=False)
 
     conflicts("~hdf5", when="+wradlib")
     conflicts("~gdal", when="+wradlib")
-    conflicts("~xarray", when="+wradlib")
 
     depends_on("python@3.6:3.10", type=("build", "run"))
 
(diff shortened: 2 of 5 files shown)


var/spack/repos/builtin/packages/py-arm-pyart/package.py
```diff
@@ -0,0 +1,68 @@
...
+    conflicts("~gdal", when="+wradlib")
+    conflicts("~xarray", when="+wradlib")
+
+    depends_on("python@3.6:3.10", type=("build", "run"))
+
+    depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools-scm@6.2:", type="build")
+    depends_on("py-cython", type="build")
+
+    depends_on("py-pytest", type="test")
```
adamjstewart: This can probably be removed, we don't run any tests other than imports

var/spack/repos/builtin/packages/py-arm-pyart/package.py
```diff
@@ -0,0 +1,68 @@
...
+    depends_on("py-fsspec", type=("build", "run"))
+    depends_on("py-s3fs", type=("build", "run"))
+
+    depends_on("py-cartopy", type="run", when="+cartopy")
+    depends_on("py-cylp", type="run", when="+cylp")
+    depends_on("gdal+python", type="run", when="+gdal")
+    depends_on("py-h5py", type="run", when="+hdf5")
+    depends_on("rsl", type=("build", "run"), when="+rsl")
+    depends_on("py-wradlib", type="run", when="+wradlib")
+    depends_on("py-xarray@0.21.1:", type="run", when="+xarray")
```
adamjstewart: xarray is always required

var/spack/repos/builtin/packages/py-cylp/package.py
```diff
@@ -0,0 +1,27 @@
...
+    solvers (CLP, CBC, and CGL). CyLP’s unique feature is that you can use it
+    to alter the solution process of the solvers from within Python."""
+
+    homepage = "https://github.com/coin-or/cylp"
+    pypi = "cylp/cylp-0.91.5.tar.gz"
+
+    version("0.91.5", sha256="d68ab1dde125be60abf45c8fd9edd24ab880c8144ad881718ddfa01ff6674c77")
+
+    depends_on("py-setuptools", type="build")
+    depends_on("py-cython@:2.9", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-cython@:2", type="build")
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-arm-pyart/package.py
glennpj: The documentation is not consistent on this. In the `INSTALL.rst` file, `xarray` is listed under both "Required Dependencies" and "Optional Dependencies". I think I am going to search the `imports` statements and check the dependencies that way.

adamjstewart: Regardless of what the docs say, it's listed under `install_requires`, so it isn't optional.

glennpj: Yea, with the docs and the commit history flipping that, I tried to interpret what was intended. Most of the `import` statements for xarray are wrapped in 
```
try:                                                                                                                                        
    import xarray as xr                                                                                                                     
    _XARRAY_AVAILABLE = True                                                                                                                
except ImportError:                                                                                                                         
    _XARRAY_AVAILABLE = False     
```
but there is one file that does a straight import. While looking at the `import` statements I discovered a couple of other issues. Please confirm if my thinking below is correct or not.

1. There is an import from `pylab`. This means that `pylab`, which is not in Spack, will need to be added as a dependency.
2. There is an import of `pandas`, which means that `py-pandas` will need to be added as a dependency.
3. There is an import of `pkg_resources`, which means that `py-setuptools` needs to also be a run dependency.
5. The following import
```
import StringIO
```
is not valid for Python-3 and needs to be 
```
from io import StringIO
```
Python-2 can use `import StringIO` but `from io import StringIO` also works. Given that Python-2 is mostly obsolete in Spack I am thinking of patching that to `from io import StringIO` and not differentiating between Python 2 and 3.  

I will also be filing issues with the project on those.

adamjstewart: Many packages have optional features that require optional dependencies. As long as you don't try to import or run that optional feature, then it doesn't really matter whether or not the imports are guarded. Of course, if importing `arm_pyart` imports every other file in the library, then it may be required.

In the case of `xarray`, since it's listed under `install_requires`, the wheel that gets installed will list it as a required dependency, and `pip check` will fail if it isn't installed. So `xarray` is definitely required.

For the other dependencies, I'll leave it to your judgement. I usually stick to `install_requires` and `extras_require` personally since I can't know all possible dependencies or how they are intended to be used.

I wouldn't worry too much about Python 2 support, I'm planning on dropping it.

glennpj: > Many packages have optional features that require optional dependencies. As long as you don't try to import or run that optional feature, then it doesn't really matter whether or not the imports are guarded.

Sorry if I was not clear, the imports for 
- `pylab`
- `pandas`

are non-guarded imports.

var/spack/repos/builtin/packages/py-arm-pyart/package.py
```diff
@@ -0,0 +1,69 @@
...
+    depends_on("py-cython", type="build")
+    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-scipy", type=("build", "run"))
+    depends_on("py-netcdf4", type=("build", "run"))
+    depends_on("py-matplotlib", type=("build", "run"))
+    depends_on("py-pooch", type=("build", "run"))
+    depends_on("py-cftime", type=("build", "run"))
+    depends_on("py-fsspec", type=("build", "run"))
+    depends_on("py-s3fs", type=("build", "run"))
+    depends_on("py-xarray@0.21.1:", type="run")
```
adamjstewart:
```suggestion
    depends_on("py-xarray@0.21.1:", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-arm-pyart/package.py
```diff
@@ -0,0 +1,69 @@
...
+    depends_on("rsl", type=("build", "run"), when="+rsl")
+    depends_on("py-wradlib", type="run", when="+wradlib")
+
+    def setup_build_environment(self, env):
+        if "+rsl" in self.spec:
+            env.set("RSL_PATH", self.spec["rsl"].prefix)
+        else:
+            # set RSL_PATH to empty to make sure RSL is not picked up from a
+            # non-spack install in /usr/local
+            env.set("RSL_PATH", "")
```
adamjstewart, line 71:
```suggestion
            env.unset("RSL_PATH")
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-arm-pyart/package.py
glennpj: The build system of `py-arm-pyart` will set `RSL_PATH` to a value in `/usr/local/...` if the variable does not exist. Unsetting the variable would potentially allow for a non-Spack build of `RSL` to get picked up. Having `RSL_PATH` set to an empty string is correct.

var/spack/repos/builtin/packages/py-arm-pyart/StringIO.patch
```diff
@@ -0,0 +1,30 @@
+diff -ru a/pyart/testing/data/make_small_mdv_ppi.py b/pyart/testing/data/make_small_mdv_ppi.py
+--- a/pyart/testing/data/make_small_mdv_ppi.py	2022-07-19 13:06:08.000000000 -0500
++++ b/pyart/testing/data/make_small_mdv_ppi.py	2022-09-10 14:26:11.021976219 -0500
+@@ -6,7 +6,10 @@
+ """
+ 
+ import struct
+-import StringIO
++try:
++    from StringIO import StringIO
```
adamjstewart, line 10: Shouldn't this be `import StringIO`? This seems to be different than the original. Again, I wouldn't worry too much about Python 2 support, we're removing it: https://github.com/spack/spack/discussions/31824

glennpj: There were other python files in there that already had the `try:` conditional in the form that I used. I felt it was best to be consistent with what else is in the `pyart` code.

adamjstewart: That's not how Python works. Later in this file, they run:
```python
fileobj = StringIO.StringIO()
```
This won't work anymore since you're using `StringIO.StringIO.StringIO()` now.

Anyway, based on the directory names, I'm guessing these files are only used to run the unit tests? I wouldn't bother patching those since we aren't testing anything.

glennpj: Those files are used for user testing, not the unit tests. The unit tests are in a different directory and not installed, although they also use `StringIO`. 

It would not be `StringIO.StringIO.StringIO()`, but `StringIO()`. Apparently, this file could only work with python-2. The original was set to import the `StringIO` module and then later use the `StringIO` function in that, so `StringIO.StringIO` in the assignment. I did not catch the later usage call. That will also need to be changed, because what I put in, based on other code, would require the call to be `StringIO()`. In any event, 
```
fileobj = StringIO.StringIO()
```
would not work in either case as that would need to be
```
fileobj = StringIO()
```
when importing with either
```
from io import StringIO
```
or
```
from StringIO import StringIO
```
I did not catch it but that assignment is setup such that it would only work for python-2, which can import `StringIO` as a module. For python-3, `StringIO` comes from the `io` module so `StringIO.StringIO` could never work.

Given all of that, I have to wonder how much the "testing" files get used. So, I can fix the assignment statement or just drop this patch altogether.

glennpj: I pushed a commit to fix the `fileobj` call.

adamjstewart (approved):

alalazo:
This PR introduced somehow a failure in the bootstrap tests. I'll investigate that, but we should probably require bootstrap to pass now that we have a unique workflow.

alalazo:
Another seemingly related failure: https://github.com/spack/spack/runs/8294123611?check_suite_focus=true

PR #32580, opened 2022-09-08
new package: FPLO

glennpj:


glennpj:
@spackbot run pipeline

var/spack/repos/builtin/packages/fplo/package.py
```diff
@@ -0,0 +1,144 @@
...
+
+    maintainers = ["glennpj"]
+
+    version("22.00-62", sha256="0d1d4e9c1e8e41900901e26c3cd08ee39dcfdeb3f2c4c8862055eaf704b6d69e")
+
+    patch("lapackconfig.patch")
+    patch("MMakefile.patch")
+    patch("ncurses.patch")
+    patch("qt-make.patch")
+    patch("fedit_py.patch")
```
alalazo: Can you add a brief comment above each `patch` directive to briefly state provenance, aim of the patch etc. ?

var/spack/repos/builtin/packages/fplo/package.py
```diff
@@ -0,0 +1,144 @@
...
+                mmakefile(*mmakefile_args)
+
+            # build main
+            make()
+            make("install")
+
+            # build XFBP
+            with working_dir(join_path("XFBP_rel", "XFBP")):
+                make()
+                make("install")
```
alalazo, line 122: Would it work to divide `build` and `install` and only pass `install` targets with `-C`?

alalazo (changes requested):

var/spack/repos/builtin/packages/fplo/package.py
glennpj: @alalazo Do you mean, as opposed to setting `with working_dir()` multiple times?

glennpj: The native build system for this is not setup to recurse into the various directories, and there is no top level `install` target. It essentially installs the components into the top level (source) directory. I then use Spack to move the desired components into `prefix` to separate it from the source tree.

alalazo: Ok, thanks for clarifying that.

var/spack/repos/builtin/packages/fplo/package.py
glennpj: Sure, I just made a commit for this.

alalazo (approved):

PR #32581, opened 2022-09-08
packages/phist: new version 1.10

jthies:


var/spack/repos/builtin/packages/phist/package.py
```diff
@@ -181,6 +185,11 @@ class Phist(CMakePackage):
     # The test_install compiles the examples and needs pkgconfig for it
     depends_on("pkgconfig", type="test")
 
+    # in 1.10 we removed some use of deprecated Trilinos interfaces
+    # (some functions in tpetra were renamed)
+    conflicts("^trilinos@13.4:", when="@:1.9")
+    conflicts("^trilinos@:13.4", when="@1.10:")
```
glennpj: I do not think the `conflicts` directives are correct here. The above is saying that `trilinos@13.4.x` does not work with any version of `phist`.

glennpj (changes requested):

tldahlgren (approved):
Confirmed sha256.

glennpj (approved):

PR #32585, opened 2022-09-09
ci: restore coverage computation

alalazo:
Modifications:
- [x] Use `pytest-cov` to integrate `coverage` with `pytest` and `pytest-xdist`
- [x] Pass `--cov-config=pyproject.toml` explicitly due to a long-standing bug


trws (approved):

alalazo:
@spackbot run pipeline

PR #32596, opened 2022-09-09
Support spackbot rebuilding all specs from source

scottwittenburg:
Support in spack's gitlab pipelines allowing maintainers to ask spackbot to rebuild everything from source.  This should help test core changes to spack that don't otherwise change the hashes of tested specs.

- [x] Allow overriding --prune-dag cli opt with env var
- [x] Use job variable to optionally prevent rebuild jobs early exit behavior
- [x] Supply new install argument to insist deps are always installed from binary, but package is only installed from source.

Steps to get `@spackbot rebuild everything`:

1. work out installation from concrete spec.json issue with #32537 and merge it
2. merge this PR
3. merge spackbot PR [#67](https://github.com/spack/spackbot/pull/67)
4. merge spack-intrastructure PR [#319](https://github.com/spack/spack-infrastructure/pull/319)

### DIFF
--- a/lib/spack/spack/cmd/ci.py
+++ b/lib/spack/spack/cmd/ci.py
@@ -284,7 +284,7 @@ def ci_rebuild(args):
     remote_mirror_override = get_env_var("SPACK_REMOTE_MIRROR_OVERRIDE")
     remote_mirror_url = get_env_var("SPACK_REMOTE_MIRROR_URL")
     spack_ci_stack_name = get_env_var("SPACK_CI_STACK_NAME")
-    check_updated = get_env_var("SPACK_REBUILD_CHECK_UP_TO_DATE")
+    rebuild_everything = get_env_var("SPACK_REBUILD_EVERYTHING")
 
     # Construct absolute paths relative to current $CI_PROJECT_DIR
     ci_project_dir = get_env_var("CI_PROJECT_DIR")
@@ -325,6 +325,8 @@ def ci_rebuild(args):
             spack_is_pr_pipeline, spack_is_develop_pipeline
         )
     )
+
+    full_rebuild = True if rebuild_everything and rebuild_everything.lower() == "true" else False
 
     # If no override url exists, then just push binary package to the
     # normal remote mirror url.
@@ -449,6 +451,8 @@ def ci_rebuild(args):
         fd.write(spack_info.encode("utf8"))
         fd.write(b"\n")
 
+    pipeline_mirrors = []
+
     # If we decided there should be a temporary storage mechanism, add that
     # mirror now so it's used when we check for a hash match already
     # built for this spec.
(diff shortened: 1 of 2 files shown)


alalazo:
My understanding from last meeting is that #32537 is not needed to have pipelines rebuild all specs. It's an orthogonal feature that ensures each job in the Gitlab pipeline will build from sources only the root node and not its dependencies. Am I correct saying that, as such, that PR can be merged independently (i.e. before or after "rebuild all") ?

scottwittenburg:
> My understanding from last meeting is that #32537 is not needed to have pipelines rebuild all specs

No, we do need that first, unless we change our approach to "spackbot rebuild everything".

alalazo:
> No, we do need that first, unless we change our approach to "spackbot rebuild everything".

What is in that PR that is needed for rebuilding everything?

scottwittenburg:
> > No, we do need that first, unless we change our approach to "spackbot rebuild everything".
> 
> What is in that PR that is needed for rebuilding everything?

A single `spack install` command that insists on rebuilding the package from source (even if a binary mirror has the spec with the same hash), while the dependencies are still installed from one of the binary mirrors.

scottwittenburg:
fyi @zackgalbreath

alalazo:
> A single spack install command 

That's the part I can't currently figure out from code. Why do we need a single command?

scottwittenburg:
fyi @kwryankrattiger

scottwittenburg:
@alalazo Should I retry the failed `ci / unit-tests / ubuntu (2.7, original, false)` and `ci / unit-tests / macos / (3.8)` checks?  I don't see any tests actually marked as failures in the output.  Maybe I'm just missing it though.

alalazo:
The test failing is:
```
lib/spack/spack/test/llnl/util/lock.py::test_write_lock_timeout_on_write_ranges_2[/run/user/1001]
```
which might be spurious. I'll trigger a re-run.

scottwittenburg:
Seems like the ubuntu python 2.7 and macos python 3.8 tests timed out this time.  I'll retry again.

scottwittenburg:
Retrying the ubunty python 2.7 check passed on the most recent retry.  However, the macos python 3.8 check is still failing, but I still don't see the error in there.

lib/spack/spack/cmd/ci.py
```diff
@@ -526,10 +547,8 @@ def ci_rebuild(args):
     if compiler_action != "FIND_ANY":
         install_args.append("--no-add")
 
-    # TODO: once we have the concrete spec registry, use the DAG hash
-    # to identify the spec to install, rather than the concrete spec
-    # json file.
-    install_args.extend(["-f", job_spec_json_path])
+    # Identify spec to install by hash
+    install_args.append("/{0}".format(job_spec.dag_hash()))
```
haampie, line 550: how does this work with "install missing compilers"? then `job_spec` is neither in the db nor in the environment so it fails? or not?

lib/spack/spack/cmd/ci.py
```diff
@@ -526,10 +547,8 @@ def ci_rebuild(args):
     if compiler_action != "FIND_ANY":
         install_args.append("--no-add")
```
haampie, line 547: also, what's the point here actually?

haampie: seems like the comment above is wrong?

lib/spack/spack/cmd/ci.py
scottwittenburg: Unfortunately compiler bootstrapping pipelines haven't worked ever since we started using the `spack.lock` to communicate what to build to downstream jobs.

lib/spack/spack/cmd/ci.py
scottwittenburg: I don't recall why we wouldn't want `--no-add` in all cases.  Since we never encounter this in practice any more, we just end up always adding this flag.  We use the flag because when the purpose of the current job is to install a dependency of one of the roots of the environment, we just want to install it, not add it as a root itself.

kwryankrattiger (dismissed):

scottwittenburg:
Retrying the unit-tests / macos (3.8) as it failed mysteriously again.

kwryankrattiger (approved):

scottwittenburg:
Python 2.7 timed out again, retrying.

PR #32599, opened 2022-09-10
py-pyproj: add v3.4.0

adamjstewart:
Successfully builds on macOS 12.5.1 (arm64) with Python 3.9.13 and Apple Clang 13.1.6.

https://github.com/pyproj4/pyproj/releases/tag/3.4.0


glennpj (approved):
Looks good to me. I like the way the dependencies are listed by source file.

PR #32609, opened 2022-09-12
compadre: new version 1.5.0

kuberry:
- url updates to indicate new home at https://github.com/sandialabs/compadre
- adds versions 1.4.1 and 1.5.0
- master branch is no longer the preferred version
- remove kokkos dependency which is redundant after requiring kokkos-kernels

### DIFF
--- a/var/spack/repos/builtin/packages/compadre/package.py
+++ b/var/spack/repos/builtin/packages/compadre/package.py
@@ -15,16 +15,17 @@ class Compadre(CMakePackage):
     constitute the rows of some globally sparse matrix.
     """
 
-    homepage = "https://github.com/SNLComputation/compadre"
-    git = "https://github.com/SNLComputation/compadre.git"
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/compadre/package.py
```diff
@@ -37,6 +38,12 @@ class Compadre(CMakePackage):
     variant("mpi", default=False, description="Enable MPI support")
     depends_on("mpi", when="+mpi")
 
+    patch(
```
alalazo, line 42: Can you add a brief comment above the directive explaining what the patch is for?

alalazo (changes requested):
One minor request, otherwise LGTM

var/spack/repos/builtin/packages/compadre/package.py
kuberry: Just added the comment on patch purpose. Thanks!

alalazo (approved):

PR #32610, opened 2022-09-12
CI: add coverage on Windows

alalazo:
Modifications:
- [x] Add and upload coverage data for unit-tests on Windows

### DIFF
--- a/.github/workflows/windows_python.yml
+++ b/.github/workflows/windows_python.yml
@@ -30,8 +30,9 @@ jobs:
     - name: Unit Test
       run: |
         echo F|xcopy .\spack\share\spack\qa\configuration\windows_config.yaml $env:USERPROFILE\.spack\windows\config.yaml
-        spack unit-test --verbose --cov --cov-config=pyproject.toml --ignore=lib/spack/spack/test/cmd
         cd spack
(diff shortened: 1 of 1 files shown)


alalazo:
@johnwparent @scheibelp Am I doing something clearly wrong here for Windows (I am trying to generate a coverage file and upload it to `codecov`)? I have no experience with that platform, and currently no access to it, so I am even wondering how can I check what is on the filesystem after the unit-tests finished.

trws:
Perhaps we still need the `coverage combine -a` and `coverage xml`?

alalazo:
> Perhaps we still need the coverage combine -a and coverage xml?

Maybe we need only `coverage.xml`? I can try to have a look at that, thanks for spotting that difference!

trws:
I'll try to find some time to poke at this, my windows box at home is back up so I can probably test it.

trws:
I hope you don't mind my popping in a commit, but since everything just worked locally, and I noticed the odd up a directory positioning of the shell on windows I thought this was worth a shot.

alalazo:
@trws Thanks for the fix, I think we're almost there. Now we just need to remove the recursive link we have in the docs before using `codecov` actions.

tgamblin (approved):

PR #28003, opened 2021-12-14
Deprecate Python 2 installations

adamjstewart:
NOTE: To clarify, this PR does not affect which versions of Python can be used to _run_ Spack. Spack will continue to support Python 2.7 and 3.5-3.6 for the foreseeable future. This PR only affects which versions of Python Spack can _install_.

### Contents

This PR deprecates versions of Python that have reached [end-of-life](https://endoflife.date/python). It also deprecates versions of other packages that require deprecated versions of Python. This is in the hopes that maintainers of those packages get pinged on this PR and offer feedback. This PR does not deprecate versions of packages that optionally require deprecated versions of Python (if I could deprecate a specific version/variant combo I would). This PR also does not deprecate versions of packages that rely on deprecated versions of package. I'm hoping that maintainers of those packages will notice this once it is merged. Once these versions are deprecated and a 0.18.0 release comes out, these versions/packages can be removed.

### Rationale

Python 2 has been unsupported for almost 2 years, and many of the Spack packages that rely on Python 2 either don't concretize, can't be downloaded, or haven't had a new release in over 10 years. When working on large PRs like #27798, it becomes exceedingly difficult to maintain support for Python 2 and test it properly. If any of these packages/versions are present in a DAG/environment, Spack will install an unsupported version of Python without warning. I don't believe this is safe behavior, and we should work towards deprecating/removing unsupported Python versions.

### Discussion

For anyone who has stumbled upon this PR or was automatically pinged, I would welcome feedback on this PR. This PR is not a hard requirement but a request. For anyone who absolutely needs Python 2 or 3.5-3.6 support, I'm willing to keep it, but I would ask that you add yourself as a maintainer for the Python package so that you can help review PRs like #27798. Also, for any of the packages where all versions have been deprecated, if you are aware of a newer release that adds Python 3 support, let me know and I can add it to the package. Otherwise, packages that only support Python 2 will be removed after the next release.

Closes #28022

PR #32615, opened 2022-09-12
Deprecate Python 2 support

adamjstewart:
This PR deprecates using Spack to install [EOL Python versions](https://endoflife.date/python), including Python 2.7, 3.1–3.6. It also deprecates running Spack with Python 2.7. Going forward, we expect Spack to have the following Python support timeline.

### Running Spack

* Spack 0.18 (spring 2022): Python 2.7, 3.5–3.10
* Spack 0.19 (fall 2022): Python 2.7 (deprecated), 3.6–3.11
* Spack 0.20 (spring 2023): Python 3.6–3.11

### Building with Spack

* Spack 0.18 (spring 2022): Python 2.7, 3.1–3.10
* Spack 0.19 (fall 2022): Python 2.7, 3.1–3.6 (deprecated), 3.7–3.11
* Spack 0.20 (spring 2023): Python 3.7–3.11

This is a reboot of #28003. See #31824 for a detailed discussion of the motivation for this PR. If you have concerns about this change, please comment on #31824.


adamjstewart:
@spackbot fix style

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -36,7 +36,7 @@ class Python(Package):
     list_url = "https://www.python.org/ftp/python/"
     list_depth = 1
 
-    maintainers = ["adamjstewart", "skosukhin", "scheibelp", "varioustoxins", "pradyunsg"]
+    maintainers = ["adamjstewart", "skosukhin", "scheibelp", "pradyunsg"]
```
adamjstewart, line 39: @varioustoxins if I remember correctly, your willingness to help maintain this build recipe was contingent on Spack having Python 2 support. I removed you since we're planning on removing Python 2 support, but let me know if you still want to be listed as a maintainer.

tldahlgren:
@tgamblin FYI. Since you mentioned a deprecation plan in a meeting earlier today.

adamjstewart:
@scottwittenburg how can I tell why e4s-pr-generate is failing?

scottwittenburg:
> @scottwittenburg how can I tell why e4s-pr-generate is failing?

Reading the trace of that job, it seems there's a known broken spec on develop:

 ```
==> Error: Pipeline generation failed due to the presence of the following specs that are known to be broken in develop:
* phist@1.10.0%gcc@11.2.0...
```

I think @zackgalbreath has been tracking down a fix for that spec to get develop pipelines passing again.

zackgalbreath:
@johnwparent deserves the credit here. [The fix has been merged](https://github.com/spack/spack/pull/32688) and develop pipelines are once again passing on GitLab CI.

adamjstewart:
@spackbot run pipeline

adamjstewart:
Yay, passing now!

tgamblin (approved):

tgamblin:
This PR deprecates using Spack to install [EOL Python versions](https://endoflife.date/python), including Python 2.7, 3.1–3.6. It also deprecates running Spack with Python 2.7. Going forward, we expect Spack to have the following Python support timeline.

### Running Spack

* Spack 0.18 (spring 2022): Python 2.7, 3.5–3.10
* Spack 0.19 (fall 2022): Python 2.7 (deprecated), 3.6–3.11
* Spack 0.20 (spring 2023): Python 3.6–3.11

### Building with Spack

* Spack 0.18 (spring 2022): Python 2.7, 3.1–3.10
* Spack 0.19 (fall 2022): Python 2.7, 3.1–3.6 (deprecated), 3.7–3.11
* Spack 0.20 (spring 2023): Python 3.7–3.11

This is a reboot of #28003. See #31824 for a detailed discussion of the motivation for this PR. If you have concerns about this change, please comment on #31824.

alalazo:
This is one of those cases where rebuilding all pipelines would have helped. At least this pipeline: https://gitlab.spack.io/spack/spack/-/pipelines/172567 is broken (we'll see the breakage on first rebuild) because it needs `python@3.6.X`. I noticed that just because of an unrelated PR: https://gitlab.spack.io/spack/spack/-/jobs/3370750

alalazo:
It seems the issue is due to the `Waf` (?) build-system. We have 3 packages using that, out of 6k. One of them already moved to CMake and was not updated. 

I'll be removing that build-system from the ones checked in pipelines. Would people agree on removing `Waf` entirely as a build-system (e.g. by updating or removing packages that rely on it)?

adamjstewart:
I don't see any reason to remove it entirely. We could test a different package if you're concerned about untested code.

alalazo:
> We could test a different package if you're concerned about untested code.

I'm slightly more concerned about maintaining a build-system for just 3 packages. Wondering if we can use another build-system for them. The three packages are:
- `ns3dev` (https://gitlab.com/nsnam/ns-3-dev already using CMake)
- `py-py2cairo` (https://www.cairographics.org/pycairo/ legacy resource for `py-pycairo` on Python 2)
- `tut` (https://github.com/mrzechonek/tut-framework/ unit-test framework for C++, I guess not widely used)

I think the first can move to `CMake`. The second will disappear when we remove Python 2 packages. For the third we are shipping a version from 2016, so I guess not many people are using it.

Not a high-priority thing, but if people agree, it wouldn't take a lot of time to remove the build-system and convert the packages.

adamjstewart:
* n3-3-dev: maybe @yee29 can comment, but I imagine we can update this to use CMake
* py-py2cairo: didn't even realize this was for Python 2, I'll mark it as such
* tut: requires Python 3.6 and older, so this will also be removed in the next release

So yeah, I'm fine with dropping `WafPackage` once ns-3-dev is converted and py-py2cairo and tut are removed. See https://waf.io/projects_using_waf.html for a list of other projects using Waf, not sure if we ever plan on adding any of these.

PR #32616, opened 2022-09-12
encode development requirements in pyproject.toml

trws:
Add a `project` block to the toml config along with development and CI dependencies and a minimal `build-system` block, doing basically nothing, so that spack can be bootstrapped to a full development environment with:

```shell
$ hatch shell
```

or for a minimal environment without hatch:

```shell
$ python3 -m venv venv
$ source venv/bin/activate
$ python3 -m pip install --upgrade pip
$ python3 -m pip install -e '.[dev]'
```

This means we can re-use the requirements list throughout the workflow yaml files and otherwise maintain this list in *one place* rather than several disparate ones.  We may be stuck with a couple more temporarily to continue supporting python2.7, but aside from that it's less places to get out of sync and a couple new bootstrap options.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,6 @@ include = [
 ]
 
 [tool.hatch.envs.default]
-detached = true
 features = [
   "dev",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,9 @@ dependencies=[
 ]
 dynamic = ["version"]
 
+[project.scripts]
+spack = "lib.spack.spack_installable.main:main"
+
 [tool.hatch.version]
 path = "lib/spack/spack/__init__.py"
 
@@ -33,6 +36,22 @@ ci = [
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
+include = [
+  "/bin",
+  "/etc",
+  "/lib",
+  "/share",
+  "/var",
+  "CITATION.cff",
+  "COPYRIGHT",
+  "LICENSE-APACHE",
+  "LICENSE-MIT",
+  "NOTICE",
+  "README.md",
+  "SECURITY.md",
+]
 
 [tool.hatch.envs.default]
 detached = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name="spack"
-version="0"
+version="0.19.0.dev0"
 description="The spack package manager"
 dependencies=[
   "six",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,54 @@
+[project]
+name="spack"
+version="0"
+description="The spack package manager"
+dependencies=[
+  "six",
+  "types-six",
+  "setuptools",
+  "clingo",
+]
+packages = []
+
+[project.optional-dependencies]
+dev = [
+  "pip>=21.3",
+  "pytest",
+  "pytest-xdist",
+  "setuptools",
+  "click==8.0.2",
+  'black==21.12b0',
+  "mypy",
+  "isort",
+  "flake8",
+  "vermin",
+]
+ci = [
+  "pytest-cov",
+  "codecov[toml]",
+]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.envs.default]
+detached = true
+features = [
+  "dev",
+]
+
+[tool.hatch.envs.default.scripts]
+spack = "./bin/spack"
+style = "./bin/spack style"
+test = "./bin/spack unit-test"
+
+[tool.hatch.envs.ci]
+features = [
+  "dev",
+  "ci",
+]
+
 [tool.black]
 line-length = 99
 target-version = ['py27', 'py35', 'py36', 'py37', 'py38', 'py39', 'py310']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,16 @@
 [project]
 name="spack"
-version="0.19.0.dev0"
 description="The spack package manager"
 dependencies=[
+  "clingo",
+  "setuptools",
   "six",
   "types-six",
-  "setuptools",
-  "clingo",
 ]
-packages = []
+dynamic = ["version"]
+
+[tool.hatch.version]
+path = "lib/spack/spack/__init__.py"
 
 [project.optional-dependencies]
 dev = [
--- a/bin/spack
+++ b/bin/spack
@@ -49,52 +49,8 @@ spack_prefix = os.path.dirname(os.path.dirname(spack_file))
 spack_lib_path = os.path.join(spack_prefix, "lib", "spack")
 sys.path.insert(0, spack_lib_path)
 
-# Add external libs
-spack_external_libs = os.path.join(spack_lib_path, "external")
-
-if sys.version_info[:2] <= (2, 7):
-    sys.path.insert(0, os.path.join(spack_external_libs, "py2"))
-
-sys.path.insert(0, spack_external_libs)
-
-# Here we delete ruamel.yaml in case it has been already imported from site
-# (see #9206 for a broader description of the issue).
-#
-# Briefly: ruamel.yaml produces a .pth file when installed with pip that
-# makes the site installed package the preferred one, even though sys.path
-# is modified to point to another version of ruamel.yaml.
-if "ruamel.yaml" in sys.modules:
-    del sys.modules["ruamel.yaml"]
-
-if "ruamel" in sys.modules:
-    del sys.modules["ruamel"]
-
-# The following code is here to avoid failures when updating
-# the develop version, due to spurious argparse.pyc files remaining
-# in the libs/spack/external directory, see:
-# https://github.com/spack/spack/pull/25376
-# TODO: Remove in v0.18.0 or later
-try:
-    import argparse
-except ImportError:
-    argparse_pyc = os.path.join(spack_external_libs, "argparse.pyc")
-    if not os.path.exists(argparse_pyc):
-        raise
-    try:
-        os.remove(argparse_pyc)
-        import argparse  # noqa: F401
-    except Exception:
-        msg = (
-            "The file\n\n\t{0}\n\nis corrupted and cannot be deleted by Spack. "
-            "Either delete it manually or ask some administrator to "
-            "delete it for you."
-        )
-        print(msg.format(argparse_pyc))
-        sys.exit(1)
-
-
-import spack.main  # noqa: E402
+from spack_installable.main import main  # noqa: E402
 
 # Once we've set up the system path, run the spack main method
 if __name__ == "__main__":
-    sys.exit(spack.main.main())
+    sys.exit(main())
--- a/lib/spack/spack/__init__.py
+++ b/lib/spack/spack/__init__.py
@@ -3,11 +3,18 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-#: (major, minor, micro, dev release) tuple
-spack_version_info = (0, 19, 0, "dev0")
-
 #: PEP440 canonical <major>.<minor>.<micro>.<devN> string
-spack_version = ".".join(str(s) for s in spack_version_info)
+__version__ = "0.19.0.dev0"
+spack_version = __version__
+
+def __try_int(v):
+    try:
+        return int(v)
+    except:
+        return v
+
+#: (major, minor, micro, dev release) tuple
+spack_version_info = tuple([__try_int(v) for v in __version__.split(".")])
+
 
 __all__ = ["spack_version_info", "spack_version"]
-__version__ = spack_version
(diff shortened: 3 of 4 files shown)


pyproject.toml
```diff
@@ -1,3 +1,54 @@
...
+  "vermin",
+]
+ci = [
+  "pytest-cov",
+  "codecov[toml]",
+]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
adamjstewart, line 38: Why hatchling?

pyproject.toml
```diff
@@ -1,3 +1,54 @@
+[project]
+name="spack"
+version="0"
```
adamjstewart:
```suggestion
version="0.19.0.dev0"
```

adamjstewart (commented):
Excited to some day be able to tell people to `pip install spack`.

pyproject.toml
trws: Honestly, because I couldn't figure out how to get setuptools to just let me do an editable install without erroring out or declaring explicit package parameters.  Poetry, which is my usual weapon of choice, seems to require a good bit of extra boilerplate.   This version made it "just work" so the dependencies will install with pip or hatch, or any pep-517/660 compliant tool presumably, out of the box.  The hatch env stuff I just find really handy, honestly, but that isn't tied to using hatchling like this, if you know a better way to get `pip install` with and without `-e` working to set this up I'm happy to explore alternatives.

pyproject.toml
trws: This is one part I wasn't sure about, it's the only place where this adds a *new* place to specify something that could get out of sync rather than removing one.  Probably still best to have it be right.

pyproject.toml
adamjstewart: What version of setuptools were you using? I know there was a recent refactor in v64 that added support for editable installs using `pyproject.toml`. I have no personal preference on build backend, just curious about the decision since I'm never sure which backend I should be using in my own projects.

pyproject.toml
adamjstewart: Looks like setuptools supports dynamically getting the version, not sure about hatchling: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

pyproject.toml
ofek: Maintainer here! Hatchling is also now used as the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/).

adamjstewart: @ofek do you know how the default was chosen? I haven't yet seen an official comparison between backends and PyPA refuses to admit that "default" == "recommended".

ofek: I think it was briefly mentioned in https://github.com/spack/spack/pull/31512, and discussed in https://github.com/pypa/packaging.python.org/pull/1031#issuecomment-1124977628

I'll quote Henry who expressed the common sentiment:

> Which backend should be the "only" backend? `tools.setuptools` is still experimental and there are a lot of older pre-PEP 621 copies of it around, and it's large and full of legacy cruft still, causing slower builds and harder error messages - it can be powerful, but it seems to be an overkill for newcomers and simple Python only projects. Flit is lite and fast, with simple error messages, but a crippled file include/exclude system that won't include `LICENSE.md` by default but will include `__pycache__` folders. Hatchling seems to be a nice balance, with reasonable defaults & nice error messages. For a nice newcomer experience, I think Hatchling is likely best.

@trws's experience is common and is by design. I've learned the best and worst parts of other backends in order to implement Hatching, which is why it Just Works ™️ while at the same time being simpler to configure and easier to extend.

trws:
@adamjstewart, this update makes the pip installed spack actually *work*.  It also reads the spack version from the canonical spack location, which unfortunately required a bit of reworking there because the pip mechanism for it uses a regex, and the original tuple isn't compatible with that.

Only the `spack` command is installed at the moment, but it's hooked in such that it will use the python managed by a venv or similar, and refactors the setup code so that both bin/spack and the pip generated script will use the same path setup logic.  It's kinda strange, but it seems to work.  I'm not sure I like using the externals like this, they really should just be pulled from packages for this probably, but this is more like how spack works when cloned so I went with it for now.

adamjstewart:
> This means we can re-use the requirements list throughout the workflow yaml files and otherwise maintain this list in _one place_ rather than several disparate ones.

Given that Spack's deps are still listed in all of the same places, this kinda feels like: https://xkcd.com/927/

trws:
> > This means we can re-use the requirements list throughout the workflow yaml files and otherwise maintain this list in _one place_ rather than several disparate ones.
> 
> Given that Spack's deps are still listed in all of the same places, this kinda feels like: https://xkcd.com/927/

Sadly this is true, we can replace the pip commands in the workflows to reuse this at least, but it doesn't necessarily get all of them (because I'm not sure a new enough pip can run on python2.7).

adamjstewart:
There's also `var/spack/repos/builtin/packages/spack/package.py` lol

trws:
🤦 

If someone were sufficiently motivated, we could write a shim `build-system` that actually uses spack to generate the dependency list for pip from that file, make wheels, etc.  I'm not sure if it's worth it, but we could.  Alternately, we could have that package reworked to have spack use pip to install spack... 🤔 

At least this way we can easily generate a local venv or similar with the same deps that are used in CI, and not have to re-list the stuff in mac, and linux, and windows and so-forth.  It's not everything, but it makes bootstrapping to style with pip less frustrating.

adamjstewart:
> Alternately, we could have that package reworked to have spack use pip to install spack

We would still need to list the deps in the package though.

trws:
Ok, I've actually gone through and replaced *all* the pip install commands in our workflows with the exact same string, installing using the deps listed in pyproject.toml.  With conditional dependencies that includes python2.7 *and* windows.  Assuming this works, pretty happy with how many of those just went away.

.github/workflows/audit.yaml
```diff
@@ -25,7 +25,8 @@ jobs:
         python-version: ${{inputs.python_version}}
     - name: Install Python packages
       run: |
-        pip install --upgrade pip six setuptools pytest codecov 'coverage[toml]<=6.2'
+        pip install --upgrade pip
+        pip install '.[dev,ci]' # not using -e because of python 2.7 support
```
adamjstewart: Is there any reason to use `-e` even for Python 3? It's all CI so why install in editable mode?

ofek: coverage output is readable by default without rewrite config

trws: Yeah, really it's to make sure that we don't accidentally run spack from the install location, for reasons like @ofek mentions.  We point to the in-tree version anyway, so it should be fine, but it's one less source of confusion.  If there were a way to only install the dependencies rather than the dependencies *and spack* that's what I would want to use, but it seems like that isn't a supported use-case in pip.

pyproject.toml
trws: @ofek, thanks for chiming in here, I hope you might be willing to answer a question for me.  I managed to get spack installing with hatchling, and largely am having a good experience, but there are two things I can't quite seem to figure out how to manage.  Our project structure is strange, because it's meant to be usable as a FHS prefix pretty much from the start, but the result is that when I use `include = [...]` and similar to install we end up with everything splatted at `site-packages/...`, and only the main spack script linked to a `bin` dir in the user's path.

* Is there a good way to say "install <dir> at <path>/<dir>"? For example, install `bin` and the others at `site-packages/spack/bin` etc?  I tried "sources" and it almost worked, the directories got produced, but all the files under them seemed to end up in the site-packages dir anyway.
* Is there a way to get symlinks or similar added to the bin directory for scripts that are not direct python entry points?  We have a couple of shell, powershell, and batch scripts to help users with environment setup and I haven't found a good way to place those yet.

ofek: - Try https://hatch.pypa.io/latest/config/build/#forced-inclusion
- No, but symlink files are resolved and included same as setuptools

trws: Ok, that works to install to a subdirectory of site-packages, so that's definitely better.  Is there any way to install to `<prefix>/<path>` though, as in outside of the `site-packages` directory?  I tried `"var" = "/var"` in force include for example but that places it in `site-packages/var`.

adamjstewart: I don't think there's a way to specify that you want to install to `<prefix>` instead of `<prefix>/lib/pythonX.Y/site-packages`. Some dependencies may also have things in `<prefix>/bin`, so you wouldn't want to map everything to a single directory anyway. @pradyunsg would be the one to ask though.

trws: Right, but there are some user-facing scripts and things that really ought to be there.  The other directories like `var` make for more of an interesting question.  If we want users to be able to install multiple spacks in different pythons, then they need to be in `site-packages` but if we want it to work like a normal FHS prefix then there are several things that are more appropriately installed at the root of the prefix.

ofek: Would `shared-data` work? https://hatch.pypa.io/latest/plugins/builder/wheel/#options

That's what Jupyter uses now https://blog.jupyter.org/packaging-for-jupyter-in-2022-c7be64c38926

adamjstewart: You can still get the best of both worlds. If things are set up correctly, `<root>/bin/spack*` will be installed to `<prefix>/bin` and `<root>/lib/spack/*` will be installed to `<prefix>/lib/pythonX.Y/site-packages`. Not sure exactly where to put other files, but I believe setuptools/etc. have options for that. This matches the FHS prefix.

Even with normal Python libs, you can't install multiple versions at the same time without having a separate environment for each one. So there's no reason Spack needs to do anything special to support that.

I think what you have right now is the bare minimum that will allow us to `pip install .` in CI. If we want to release on PyPI, we'll need to do quite a bit more. `etc`, `share`, and `var` will need to be installed somewhere, and `opt` may need more thought as well. And since these locations are different in our source code vs at run-time after an install, we'll have to think about how to support both, or think about changing the layout of our source code to match what would be installed.

trws: I agree this is nowhere near ready for PyPi, but you might be surprised by how well it actually works as an installed package.  Even with everything just installing to site-packages with a shim for the spack command, everything pretty much just works, because all the paths were relative to the repo before and the relative paths have stayed the same.

adamjstewart: In setuptools lingo, I think the Spack command is an entry point. Does hatchling also support entry points?

trws: It does, and that one works just fine with the new main shim I put in so we could move the path logic out of `bin/spack`.  It's the shell scripts etc. that don't work as entry points.  Seems like shared-data does work though, thanks @ofek!  With that we can actually install to the prefix for some of this stuff, but we might run into path problems, running a test or two. If the path resolution doesn't work this way I'm going to leave everything (but the scripts) in site-packages so things keep working and call it good enough for this PR.

adamjstewart: In the long run we may actually end up splitting core and packages into two separate libraries so don't worry too much about where var ends up.

ofek: > Does hatchling also support entry points?

https://hatch.pypa.io/latest/config/metadata/#cli

pradyunsg: shared-data seems fine, and is the right thing assuming that this isn't code and indeed is random data files. :)

trws: @pradyunsg it's a bit of a grey area IMO, it's things like package files, which are technically code but are most definitely not libraries or things meant to be run by users, configuration files, etc.

pradyunsg: I'm pretty sure that you can remove the bin/spack file, since it's being replaced by the script that will be generated for `project.scripts` declarations. FWIW, those are what the console_scripts entry points from setuptools map to -- I'm quite pleased that we have a friendlier user-facing name for the concept.

adamjstewart: I think it's a bit of a gray area. We would like to be able to run Spack just by cloning and running `./spack/bin/spack`, but we would also like to be able to install with pip. I think the former use case will actually be more common that the latter, so we can't just get rid of the script.

trws: @pradyunsg Yeah, it wouldn't be installed by pip, but we need it there for the "clone and go" use-case.

On installing, the shared-data option definitely works, but splitting our directories up breaks all the relative pathing stuff so things break, really badly.  I did manage to get install to produce a console_script that works, and install all the directories under `site-packages/spack_root` though rather than splatting them all over the place, and the resulting install can install packages and generally kinda just works.

pradyunsg: In thar case, I suggest removing it  from the list of files being included in the wheel (`python -m build && unzip -l dist/*.whl` shouldn't list any sort of `.scripts/spack` since that would be overwritten by a launcher script that pip generates instead.)

ofek:
CI flake?

trws:
So... I would really like to merge this, so it doesn't atrophy.  Is there a remaining blocker or concern here @adamjstewart, @tgamblin?  I'm rebasing to resolve the workflows right now.

trws:
Rebased again, from the pool. Any chance of a review? This is verified to work with hatch, pip, and with a bit of adaptor code nix as well.  Could certainly do with more helper commands and some thought on actually using it as an installed package, but everything is functional.

tgamblin (approved):

PR #32618, opened 2022-09-12
add ability to disable werror for rocksdb

robertu94:
The default `-Werror` in the package causes build failures on newer versions of GCC.

### DIFF
--- a/var/spack/repos/builtin/packages/rocksdb/package.py
+++ b/var/spack/repos/builtin/packages/rocksdb/package.py
@@ -33,6 +33,7 @@ class Rocksdb(MakefilePackage):
     variant("zlib", default=True, description="Enable zlib compression support")
     variant("zstd", default=False, description="Enable zstandard compression support")
     variant("tbb", default=False, description="Enable Intel TBB support")
+    variant("werror", default=False, description="Build with -Werror")
 
     depends_on("bzip2", when="+bz2")
     depends_on("gflags")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/rocksdb/package.py
```diff
@@ -93,6 +94,9 @@ def install(self, spec, prefix):
         env["CFLAGS"] = " ".join(cflags)
         env["PLATFORM_FLAGS"] = " ".join(ldflags)
 
+        if "+werror" not in self.spec:
```
tldahlgren: if this is tied to newer versions of GCC, would it be better to check that instead of making it a variant?  Or do you want to make this optional in general?

robertu94: I think this should be general. While I hit this with new GCC, being able to turn off warning to Error probably helps on other compilers which may generate warnings the devs haven't seen before.

tldahlgren: Suggest removing the double negative:

```suggestion
        if "~werror" in self.spec:
```

tldahlgren (commented):
Is it worth adding a `conflict()` for the newer `gcc` compilers if the new variant is not set properly?

robertu94:
> Is it worth adding a `conflict()` for the newer `gcc` compilers if the new variant is not set properly?

I don't think so.  I set the  werror variant by default to be False, but IMO in principle people should be able to include warnings as errors when ever they want them even if they lead to build failures.

tldahlgren (approved):
LGTM

PR #32619, opened 2022-09-12
Add libraqm package; enable in py-pillow

certik:
Now Pillow can render OTF fonts with ligatures and glyph substitutions correctly. Must be built with:
```
spack install py-pillow+freetype+raqm
```

### DIFF
--- a/var/spack/repos/builtin/packages/libraqm/package.py
+++ b/var/spack/repos/builtin/packages/libraqm/package.py
@@ -1,0 +1,19 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack import *
+
+
+class Libraqm(MesonPackage):
+    """Raqm is a small library that encapsulates the logic for complex text layout and provides a convenient API."""
+
+    homepage = "https://github.com/HOST-Oman/libraqm"
+    url      = "https://github.com/HOST-Oman/libraqm/releases/download/v0.9.0/raqm-0.9.0.tar.xz"
+    git      = "https://github.com/HOST-Oman/libraqm.git"
+
+    version('0.9.0', tag="v0.9.0")
+
+    depends_on("freetype")
+    depends_on("harfbuzz")
--- a/var/spack/repos/builtin/packages/libraqm/package.py
+++ b/var/spack/repos/builtin/packages/libraqm/package.py
@@ -10,10 +10,10 @@ class Libraqm(MesonPackage):
     """Raqm is a small library that encapsulates the logic for complex text layout and provides a convenient API."""
 
     homepage = "https://github.com/HOST-Oman/libraqm"
-    url      = "https://github.com/HOST-Oman/libraqm/releases/download/v0.9.0/raqm-0.9.0.tar.xz"
-    git      = "https://github.com/HOST-Oman/libraqm.git"
+    url = "https://github.com/HOST-Oman/libraqm/releases/download/v0.9.0/raqm-0.9.0.tar.xz"
+    git = "https://github.com/HOST-Oman/libraqm.git"
 
-    version('0.9.0', tag="v0.9.0")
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/libraqm/package.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Libraqm(MesonPackage):
+    """Raqm is a small library that encapsulates the logic for complex text layout and provides a convenient API."""
+
+    homepage = "https://github.com/HOST-Oman/libraqm"
+    url      = "https://github.com/HOST-Oman/libraqm/releases/download/v0.9.0/raqm-0.9.0.tar.xz"
+    git      = "https://github.com/HOST-Oman/libraqm.git"
+
+    version('0.9.0', tag="v0.9.0")
```
adamjstewart: Can we use stable tarball checksums instead of unstable git tags?

var/spack/repos/builtin/packages/py-pillow/package.py
```diff
@@ -25,6 +25,7 @@ class PyPillowBase(PythonPackage):
     variant("jpeg2000", default=False, description="JPEG 2000 functionality")
     variant("imagequant", when="@3.3:", default=False, description="Improved color quantization")
     variant("xcb", when="@7.1:", default=False, description="X11 screengrab support")
+    variant('raqm', default=False, description='RAQM support')
```
adamjstewart: Any idea what the first version of pillow supporting RAQM is?

var/spack/repos/builtin/packages/libraqm/package.py
certik: Sure, fixed in ff73e97770cfad348d2921d8d3b48b3499cfeac7.

var/spack/repos/builtin/packages/py-pillow/package.py
certik: Good question. Version 8.0.0 says `error in setup.cfg: command 'pil_build_ext' has no such option 'enable_raqm'`, the next version in Spack 8.4.0 works.

certik: Fixed in 53740ac22bdf7bc859ec8947f8bd0a594587ad31.

certik:
@adamjstewart this is ready for review.

var/spack/repos/builtin/packages/py-pillow/package.py
```diff
@@ -25,6 +25,7 @@ class PyPillowBase(PythonPackage):
     variant("jpeg2000", default=False, description="JPEG 2000 functionality")
     variant("imagequant", when="@3.3:", default=False, description="Improved color quantization")
     variant("xcb", when="@7.1:", default=False, description="X11 screengrab support")
+    variant("raqm", when="@8.2:", default=False, description="RAQM support")
```
adamjstewart, line 28: Tested this with in-between versions and it looks like 8.2.0 is the first version to support this flag.

var/spack/repos/builtin/packages/py-pillow/package.py
```diff
@@ -52,6 +53,10 @@ class PyPillowBase(PythonPackage):
     depends_on("openjpeg", when="+jpeg2000")
     depends_on("libimagequant", when="+imagequant")
     depends_on("libxcb", when="+xcb")
+    depends_on("libraqm", when="+raqm")
+
+    # Conflicting options
+    conflicts("+raqm", when="~freetype")
```
adamjstewart, line 59: setup.py will error if this combination is given

var/spack/repos/builtin/packages/py-pillow/package.py
```diff
@@ -70,7 +75,7 @@ def patch(self):
 
         def variant_to_cfg(variant):
             able = "enable" if "+" + variant in self.spec else "disable"
-            return "{0}-{1}=1\n".format(able, variant)
+            return "{0}_{1}=1\n".format(able, variant)
```
adamjstewart, line 78: setuptools 54.1+ issues a warning about using `enable-foo`, we should instead be using `enable_foo`. I didn't test ancient setuptools to make sure that this works.

adamjstewart (approved):
Builds for me, thanks for the PR!

certik:
Thanks for the fixes, testing it and a review!

PR #32625, opened 2022-09-12
unifyfs & gotcha: new releases

CamStan:
UnifyFS:
- Add 1.0 release
- Deprecate older, unsupported versions
- Set `fortran` variant to `true` by default
- Update `gotcha` and `mochi-margo` dependency versions for `unifyfs@1.0` and `unifyfs@develop`
- Add conflict of `unifyfs` with `libfabric` 1.13.* versions
- Update `configure_args` to use helper functions

GOTCHA: Hasn't been under active development for a couple years but the develop branch has some fixes UnifyFS uses. To avoid having UnifyFS v1.0 depend on a develop branch of a dependency, this creates a new release in the Gotcha Spack package based on the most recent commit of the develop branch. This can be
updated if/when GOTCHA development picks up again.

### DIFF
--- a/var/spack/repos/builtin/packages/gotcha/package.py
+++ b/var/spack/repos/builtin/packages/gotcha/package.py
@@ -17,6 +17,7 @@ class Gotcha(CMakePackage):
 
     version("develop", branch="develop")
     version("master", branch="master")
+    version("1.0.4", commit="08a5279cef051ed39bc92c5f77f55ca5b41a405e")
     version("1.0.3", tag="1.0.3")
     version("1.0.2", tag="1.0.2")
     version("0.0.2", tag="0.0.2")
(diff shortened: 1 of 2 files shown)


tldahlgren (approved):
Confirmed the `gotcha` commit and the new `unifyfs` sha256.  

Not sure about assigning a version to `gotcha`.  I've seen people put date stamps in this type of situation but I can see the ease of not doing that.

Changes to `configure_args` make it cleaner, more compact.

CamStan:
> Not sure about assigning a version to `gotcha`. I've seen people put date stamps in this type of situation but I can see the ease of not doing that.

I went back and forth on this as well. I wanted to make sure it ended up the preferred version, but now that I mention it, I could always use a date stamp and add `preferred=True`

tldahlgren:
> > Not sure about assigning a version to `gotcha`. I've seen people put date stamps in this type of situation but I can see the ease of not doing that.
> 
> I went back and forth on this as well. I wanted to make sure it ended up the preferred version, but now that I mention it, I could always use a date stamp and add `preferred=True`

Does this mean you want this PR merged as is or that you will make that change?

PR #32629, opened 2022-09-13
Update module_file_support.rst

waveman68:
Missing lead in "This may to inconsistent modulfiles if..."


tldahlgren (approved):
Great catch!

Issue #9149, opened 2018-08-31
External packages cannot access dependencies in spec


adamjstewart:
I'm trying to register `py-numpy` as an external package, but when I try to install it, it crashes saying that `python` is not in the spec.

### Steps to reproduce the issue

1. Register `py-numpy` as an external package
2. Install `py-numpy`

### Error Message

```console
$ spack --debug install py-numpy
==> py-numpy@1.14.2 : has external module in bwpy/1.2.4
==> py-numpy@1.14.2 : is actually installed in /opt/bwpy
==> py-numpy@1.14.2 : generating module file
Traceback (most recent call last):
  File "/u/sciteam/stewart1/spack/bin/spack", line 54, in <module>
    sys.exit(spack.main.main())
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/main.py", line 653, in main
    return _invoke_command(command, parser, args, unknown)
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/main.py", line 432, in _invoke_command
    return_val = command(parser, args)
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/cmd/install.py", line 273, in install
    install_spec(args, kwargs, spec)
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/cmd/install.py", line 154, in install_spec
    spec.package.do_install(**kwargs)
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/package.py", line 1380, in do_install
    return self._process_external_package(explicit)
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/package.py", line 1302, in _process_external_package
    spack.hooks.post_install(self.spec)
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/hooks/__init__.py", line 73, in __call__
    hook(*args, **kwargs)
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/hooks/module_file_generation.py", line 48, in <lambda>
    post_install = lambda spec: _for_each_enabled(spec, 'write')
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/hooks/module_file_generation.py", line 41, in _for_each_enabled
    getattr(generator, method_name)()
  File "/mnt/a/u/sciteam/stewart1/spack/lib/spack/spack/modules/common.py", line 698, in write
    context = self.context.to_dict()
[...]

PR #32632, opened 2022-09-13
[py-pip] fix setup_dependent_package in py-pip

mdorier:
This PR fixes a problem with the py-pip package: if python and py-pip are provided as external packages, the line that this PR changes will currently fail when installing packages that depend on py-pip.


adamjstewart:
Cross-linking to #9149 which explains why we have to make this change. External packages like `py-pip` cannot access their dependencies in their own spec.

adamjstewart (approved):

tldahlgren:
@spackbot run pipeline

PR #32638, opened 2022-09-13
json-glib: add pkgconfig dependency

adamjstewart:
Successfully builds on macOS 10.15.7 (x86_64) with Apple Clang 12.0.0.

I found this dependency to be required to properly locate gobject-introspection.


tldahlgren (approved):
LGTM

PR #32642, opened 2022-09-13
libcatalyst: use git instead of fixed urls

vicentebolea:
I had difficulties custom installing specific versions of Catalyst, here I change URL to git to be able to install commit at wish.

    The issue is that we are not not able to install (Fetch URL error) any
    version of catalyst other than the specified in the spack package.py.
    This very version is accessible only because it is cached by Spack. The
    real URL does not exist anymore, I believe the reason is that there used
    to be a tag in catalyst that does not exist anymore.

### DIFF
--- a/var/spack/repos/builtin/packages/libcatalyst/package.py
+++ b/var/spack/repos/builtin/packages/libcatalyst/package.py
@@ -11,14 +11,12 @@ class Libcatalyst(CMakePackage):
     scientific data producers) to analyze and visualize data in situ."""
 
     homepage = "https://gitlab.kitware.com/paraview/catalyst"
-    url = "https://gitlab.kitware.com/paraview/catalyst/-/archive/{0}/catalyst-{0}.tar.bz2"
+    git = "https://gitlab.kitware.com/paraview/catalyst.git"
 
     maintainers = ["mathstuf"]
 
     # master as of 2021-05-12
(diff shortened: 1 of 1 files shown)


mathstuf:
> I had difficulties

Such as?

vicentebolea:
@mathstuf 

The issue is that I am not able to install (Fetch URL error) any version of catalyst other than the specified in the spack package.py. This very version is accessible only because it is cached by Spack. The real URL does not exist anymore, I believe the reason is that there used to be a tag in catalyst that does not exist anymore.

mathstuf (dismissed):
It'd be nice if that information was provided in the commit messages. Otherwise it seems fine.

tldahlgren (approved):
Confirmed commit exists in the repository and that 'spack stage' of the version succeeds.

tldahlgren (commented):
Still testing the fetch.

tldahlgren:
@vicentebolea What version(s) are you trying to fetch with this change where removing the url helps?

I have not been able to convince myself this works as expected.

vicentebolea:
> I have not been able to convince myself this works as expected.

libcatalyst@3f7871c0a2e737cb9ed35fc1c2208456fcc00a0e

tldahlgren (approved):

PR #32643, opened 2022-09-13
Axom: bring in changes from axom repo

white238:
It has been a while since I pulled in the updates from the Axom repo.

This mainly adds HIP support and uses a Spack's BLT instead of the internal submodule.

This also removes the the need to pull submodules with the source.  The `data` submodule is particularly large.


white238:
@spackbot fix style

var/spack/repos/builtin/packages/axom/package.py
```diff
@@ -404,6 +506,8 @@ def initconfig_package_entries(self):
     def cmake_args(self):
         options = []
 
+        options.append("-DBLT_SOURCE_DIR:PATH={0}".format(spec["blt"].prefix))
```
tldahlgren:
```suggestion
        options.append("-DBLT_SOURCE_DIR:PATH={0}".format(self.spec["blt"].prefix))
```

tldahlgren (changes requested):
I have not reviewed the rest of this PR but the suggested change will at least help resolve the pipeline stack failures.

white238:
I have tested this in Axom's large set of specs across many LC machines.  It should be ready to go.

var/spack/repos/builtin/packages/axom/package.py
```diff
@@ -95,40 +92,56 @@ def submodules(self):
...
+    for val in ROCmPackage.amdgpu_targets:
+        depends_on("raja amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val))
+        depends_on("umpire amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val))
 
     depends_on("mfem", when="+mfem")
     depends_on("mfem~mpi", when="+mfem~mpi")
 
+    # Disable fortran, causing "cannot compile a simple Fortran program"
+    # with crayftn
+    depends_on("hypre~fortran", when="+mfem+rocm")
```
tldahlgren: Given the comment, should this be `when="+mfem+rocm platform=cray"`?

var/spack/repos/builtin/packages/axom/package.py
```diff
@@ -222,15 +249,73 @@ def initconfig_hardware_entries(self):
...
+            # Fixes for mpi for rocm until wrapper paths are fixed
+            # These flags are already part of the wrapped compilers on TOSS4 systems
+            hip_link_flags = ""
+            if "+fortran" in spec:
+                # Flags for crayftn
+                if self.is_fortran_compiler("crayftn"):
+                    # Fix for working around CMake adding implicit link directories
+                    # returned by the Cray crayftn compiler to link executables with
+                    # non-system default stdlib
+                    cray_exclude_path = "/opt/cray/pe/gcc/8.1.0/snos/lib64"
```
tldahlgren: 😱  Hard-coded paths here and below.

tldahlgren (dismissed):
Other than my question and aversion to hard-coded paths, this LGTM.

@scheibelp Did you want to review this package, especially regarding the cray-specific changes?

var/spack/repos/builtin/packages/axom/package.py
white238: I know but they are unfortunately required for our platforms and are guarded against non-existence.

Hopefully in the future I can make them less hardcoded.

white238: I looked it over and this is untested and possibly from an old system. I removed it.

var/spack/repos/builtin/packages/axom/package.py
white238: Just removed it. I think its from an old spec that we dont use/test anymore.

white238:
> Other than my question and aversion to hard-coded paths, this LGTM.
> 
> @scheibelp Did you want to review this package, especially regarding the cray-specific changes?

This has been addressed by just removing it

var/spack/repos/builtin/packages/axom/package.py
```diff
@@ -65,11 +62,11 @@ def submodules(self):
     # Variants
     # -----------------------------------------------------------------------
     variant("shared", default=True, description="Enable build of shared libraries")
-    variant("debug", default=False, description="Build debug instead of optimized version")
 
     variant("examples", default=True, description="Build examples")
     variant("tools", default=True, description="Build tools")
 
+    # Hard requirement after Axom 0.6.1
     variant("cpp14", default=True, description="Build with C++14 support")
```
becker33, line 70: We should make this variant only appear for `axom@:0.6.0`

`variant("cpp14", default=True, when="@:0.6.0", ...)`

doesn't need to be part of this PR

var/spack/repos/builtin/packages/axom/package.py
```diff
@@ -366,7 +441,8 @@ def initconfig_package_entries(self):
             entries.append("# Root directory for generated developer tools\n")
             entries.append(cmake_cache_path("DEVTOOLS_ROOT", devtools_root))
 
-            # Only turn on clangformat support if devtools is on
+        if "+devtools" in spec and "toss_4" not in self._get_sys_type(spec):
```
becker33, line 444: Short term hack to support a problem specific to LC systems

becker33 (approved):

PR #32250, opened 2022-08-18
ROCm: Enable OpenMP support always.

kwryankrattiger:
FYI: @eugeneswalker @wspear  This should fix the issues seen buliding `vtk-m +rocm`

#32197

PR #32645, opened 2022-09-14
rocfft test require rocm-openmp-extras instead llvm-amdgpu +opemnp

renjithravindrankannath:
test depends on llvm-amdgpu +openmp which is failing.
It can depend on rocm-openmp-extras with the latest changes.

### DIFF
--- a/var/spack/repos/builtin/packages/rocfft/package.py
+++ b/var/spack/repos/builtin/packages/rocfft/package.py
@@ -96,7 +96,7 @@ class Rocfft(CMakePackage):
     depends_on("googletest@1.10.0:", type="test")
     depends_on("fftw@3.3.8:", type="test")
     depends_on("boost@1.64.0: +program_options", type="test")
-    depends_on("llvm-amdgpu +openmp", type="test")
+    depends_on("rocm-openmp-extras", type="test")
 
     def check(self):
         exe = join_path(self.build_directory, "clients", "staging", "rocfft-test")


cgmb:
There are three packages that depend on `llvm-amdgpu +openmp`.
```
var/spack/repos/builtin/packages/rocblas/package.py:    depends_on('llvm-amdgpu +openmp', type='test')
var/spack/repos/builtin/packages/rocfft/package.py:    depends_on('llvm-amdgpu +openmp', type='test')
var/spack/repos/builtin/packages/rocm-tensile/package.py:        depends_on('llvm-amdgpu@' + ver,               when='@' + ver + '+openmp')
```

Can we not just fix `llvm-amdgpu +openmp`? I mean, the option exists... shouldn't it work?

cgmb:
I applied this patch on 31c24cd0d543f3bc06311fe148751fe88a11ecbb and installed it on a clean Ubuntu 20.04 docker image. When I tried `spack install --test root rocfft amdgpu_target=gfx1011`, it failed with an error:
```
/tmp/ubuntu/spack-stage/spack-stage-rocfft-5.2.0-jcvrmggxdttn3lrqvte7c2ndzey63l5n/spack-src/clients/tests/../fft_params.h:29:10: fatal err 
            or: 'omp.h' file not found
```
I assume that this built correctly for you, so perhaps rocfft found a copy of OpenMP in your system environment?

cgmb:
I think the cause of the failure is that openmp-extras from apt installs omp.h to

    /opt/rocm-5.2.0/llvm/lib/clang/14.0.0/include/omp.h

while spack installs omp.h to

    spack/opt/spack/linux-ubuntu20.04-zen2/gcc-9.4.0/rocm-openmp-extras-5.2.0-pjj4szpyeb47vvcdvzmlt5eeb5jdqnfl/include/omp.h

It seems that rocFFT assumes that omp.h will be in the clang include directory. I suppose I'll see if using `find_package(OpenMP)` can fix this problem

renjithravindrankannath:
Closing this PR since rocm-openmp-extras will be included in [32250](https://github.com/spack/spack/pull/32250)

PR #29889, opened 2022-04-05
Add checksum for py-awkward@1.5.1 and fix rapidjson dependecy

aandvalenzuela:
For the _rapidjson_ dependency, it turns out that the issue [#1924](https://github.com/Tencent/rapidjson/issues/1924) is still present in the latest version. Enforcing version < 1.1.0 skips the issue.

PR #29869, opened 2022-04-04
rapidjson: Add version 1.2.0-2022-03-09

haralmha:
Includes PR https://github.com/Tencent/rapidjson/pull/1940 which fixes https://github.com/Tencent/rapidjson/issues/1924 causing errors for newer compilers on centos

PR #32650, opened 2022-09-14
py-awkward: allow newer, fixed rapidjson versions

vvolkl:
Rapidjson was constrained due to a bug (see https://github.com/spack/spack/pull/29889) , but newer (although untagged) versions of rapidjson have since been added to spack (https://github.com/spack/spack/pull/29869). Tested the build of py-awkward with the latest, works fine.

@aandvalenzuela


adamjstewart (approved):
Reading through the discussions, it seems like the supported versions of `rapidjson` should have never been pinned in `py-awkward`. This isn't a `py-awkward` build issue, it's a `rapidjson` build issue. If `rapidjson` doesn't build with certain compilers, that should be marked in that package, not in `py-awkward`. Thanks for correcting this!

PR #32658, opened 2022-09-15
Adding intel-oneapi-itac package

stephenmsachs:
Tagging @rschon2 if he wants to be added as maintainer.

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
@@ -10,7 +10,8 @@ from spack.package import *
 
 @IntelOneApiPackage.update_description
 class IntelOneapiItac(IntelOneApiPackage):
-    """The Intel® Trace Analyzer and Collector profiles and analyzes MPI applications to help focus your optimization efforts.
+    """The Intel® Trace Analyzer and Collector profiles and analyzes MPI applications to help
+    focus your optimization efforts.
 
     Find temporal dependencies and bottlenecks in your code.
     Check the correctness of your application.
@@ -29,10 +30,8 @@ class IntelOneapiItac(IntelOneApiPackage):
     if platform.system() == "Linux":
         version(
             "2021.6.0",
-            url=
-            "https://registrationcenter-download.intel.com/akdlm/irc_nas/18694/l_itac_oneapi_p_2021.6.0.434_offline.sh",
-            sha256=
-            "1ecc2735da960041b051e377cadb9f6ab2f44e8aa44d0f642529a56a3cbba436",
+            url="https://registrationcenter-download.intel.com/akdlm/irc_nas/18694/l_itac_oneapi_p_2021.6.0.434_offline.sh",
+            sha256="1ecc2735da960041b051e377cadb9f6ab2f44e8aa44d0f642529a56a3cbba436",
             expand=False,
         )
 
--- a/var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
@@ -26,8 +26,8 @@ class IntelOneapiItac(IntelOneApiPackage):
     """
 
     homepage = "https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/trace-analyzer.html"
-    
-    maintainers = ['rscohn2']
+
+    maintainers = ["rscohn2"]
 
     if platform.system() == "Linux":
         version(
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
```diff
@@ -0,0 +1,40 @@
...
+    Analyze performance of subroutines or code blocks.
+    Learn about communication patterns, parameters, and performance data.
+    Identify communication hot spots.
+    Decrease time to solution and increase application efficiency.
+
+    """
+
+    homepage = "https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/trace-analyzer.html"
+
+    if platform.system() == "Linux":
```
tldahlgren, line 32: Curious.  Why not add the `version()` directive then a `conflicts()` if the platform is not `Linux`?

Could use something along the following (or in a loop):

```
    conflicts("platform=cray", msg="Cray is not supported")
    conflicts("platform=darwin", msg="macOS is not supported")
    conflicts("platform=windows", msg="Windows is not supported")
```

var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
```diff
@@ -0,0 +1,40 @@
...
+        version(
+            "2021.6.0",
+            url="https://registrationcenter-download.intel.com/akdlm/irc_nas/18694/l_itac_oneapi_p_2021.6.0.434_offline.sh",
+            sha256="1ecc2735da960041b051e377cadb9f6ab2f44e8aa44d0f642529a56a3cbba436",
```
tldahlgren, line 36: Confirmed sha256.

tldahlgren (changes requested):
It is more standard to use `conflicts()` instead of a conditional version.

var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
stephenmsachs: I held close to the related `intel-oneapi-*` packages. I think we should either change them all at once or none. And, thinking of it, adding support for linux seems more robust to me than removing support for all others. @rscohn2 what are your thoughts?

rscohn2: It would be good to move the logic here: https://github.com/spack/spack/blob/a0c7209dc1480b16531fcb992642986f4796bc28/lib/spack/spack/build_systems/oneapi.py#L22
Is it possible to do something like?
```
conflicts("platform!=linux", msg="Platform is not supported")
```
That would be a separate PR, and leave this PR consistent with all the others.

rscohn2:
@stephenmsachs: The CI failed, but it does not look related to your change so I rebased and restarted CI. I forgot that this would do a forced push. Sorry if that causes confusion.

rscohn2 (dismissed):
Thanks for the contribution.

var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
```diff
@@ -0,0 +1,40 @@
...
+    Visualize and understand parallel application behavior.
+    Evaluate profiling statistics and load balancing.
+    Analyze performance of subroutines or code blocks.
+    Learn about communication patterns, parameters, and performance data.
+    Identify communication hot spots.
+    Decrease time to solution and increase application efficiency.
+
+    """
+
+    homepage = "https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/trace-analyzer.html"
```
rscohn2, line 28:
```suggestion
    homepage = "https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/trace-analyzer.html"
    
    maintainers = ['rscohn2']
```

var/spack/repos/builtin/packages/intel-oneapi-itac/package.py
rscohn2:
```
conflicts("platform!=linux", msg="Platform is not supported")
```
doesn't work. Having:
```
    conflicts("platform=cray", msg="Cray is not supported")
    conflicts("platform=darwin", msg="macOS is not supported")
    conflicts("platform=windows", msg="Windows is not supported")
```

in the base class would be better than logic in every derived class as it is now. It would be even better if the logic was that linux was supported instead of everything else not supported. @alalazo do you have any suggestions?

rscohn2: @tldahlgren: Are you ok with merging this as-is?

I am willing to change it to something like:
```
    conflicts("platform=cray", msg="Cray is not supported")
    conflicts("platform=darwin", msg="macOS is not supported")
    conflicts("platform=windows", msg="Windows is not supported")
```

in a separate commit that changes all packages so it is in the base class. I am still hoping for a conflicts that says ``not linux`` instead of enumerating all platforms, but that is not a big deal.

tldahlgren (approved):

PR #32660, opened 2022-09-15
Add mirrors for octopus package

iamashwin99:
Octopus package has two sources for downloading, one from GitLab and one from their website,
This PR adds both sources to the package.


RemiLacroix-IDRIS:
@iamashwin99 : Are you sure it's the same files? As far as I can tell it's not.

iamashwin99:
Thanks for pointing it out.
I just talked with one of the octopus developers.
One of them is automatically generated by GitLab when creating a tag, and the other one is packaged by buildbot.
It's not clear why they have different sizes although they both seem to be able to compile. 
This does mean that we can't have the GitLab source mirror as the hashes don't match.
Sorry for the trouble, I'm closing this PR.

PR #32661, opened 2022-09-15
root: make X11 really optional on macOS

aphecetche:
This is an attempt to get Root compiled with OpenGL but without requiring X11 on macOS.

Note that this is my first Spack PR... Hopefully I've correctly read the [contribution guide]( https://spack.readthedocs.io/en/latest/contribution_guide.html) though. You'll let me know :wink: 

Note also that I have some other changes in the root recipe that I'd like to propose (in other PRs) to "activate" two more variants : arrow and monalisa. I assumed it would be better to stage the modifications, but I can of course propose a combined PR if preferred.

### DIFF
--- a/var/spack/repos/builtin/packages/apple-gl/package.py
+++ b/var/spack/repos/builtin/packages/apple-gl/package.py
@@ -3,7 +3,6 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-import sys
 
 from spack.package import *
 
@@ -19,7 +18,7 @@ class AppleGl(Package):
 
     version("4.1.0")
 
-    provides("gl@4.1", when=sys.platform == "darwin")
+    provides("gl@4.1")
 
     # Only supported on 'platform=darwin' and compiler=apple-clang
     conflicts("platform=linux")
--- a/var/spack/repos/builtin/packages/apple-gl/package.py
+++ b/var/spack/repos/builtin/packages/apple-gl/package.py
@@ -1,0 +1,58 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import sys
+
+from spack.package import *
+
+
+class AppleGl(Package):
+    """Shim package for the core OpenGL library from Apple"""
+
+    homepage = "https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_intro/opengl_intro.html"
+
+    maintainers = ["aphecetche"]
+
+    has_code = False
+
+    version("4.1.0")
+
+    provides("gl@4.1", when=sys.platform == "darwin")
+
+    # Only supported on 'platform=darwin' and compiler=apple-clang
+    conflicts("platform=linux")
+    conflicts("platform=cray")
+    conflicts("%gcc")
+    conflicts("%clang")
+
+    phases = []
+
+    sdk_base = (
+        "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/"
+        "Developer/SDKs/MacOSX"
+    )
+
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        # we try to setup a build environment with enough hints
+        # for the build system to pick up on the Apple framework version
+        # of OpenGL.
+        # - for a cmake build we actually needs nothing at all as
+        # find_package(OpenGL) will do the right thing
+        # - for the rest of the build systems we'll assume that
+        # setting the C_INCLUDE_PATH will be enough for the compilation phase
+        # and *** for the link phase.
+        env.prepend_path("C_INCLUDE_PATH", self.sdk_base)
+
+    @property
+    def headers(self):
+        return HeaderList(
+            "{}.sdk/System/Library/Frameworks/OpenGL.framework/Headers".format(self.sdk_base)
+        )
+
+    @property
+    def libs(self):
+        return LibraryList(
+            "{}.sdk/System/Library/Frameworks/OpenGL.framework".format(self.sdk_base)
+        )
--- a/var/spack/repos/builtin/packages/apple-glu/package.py
+++ b/var/spack/repos/builtin/packages/apple-glu/package.py
@@ -1,0 +1,58 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import sys
+
+from spack.package import *
+
+
+class AppleGlu(Package):
+    """Shim package for Apple implementation of OpenGL Utility Libray (GLU)"""
+
+    homepage = ""
+
+    maintainers = ["aphecetche"]
+
+    has_code = False
+
+    version("1.3.0")
+
+    provides("glu@1.3", when=sys.platform == "darwin")
+
+    # Only supported on 'platform=darwin' and compiler=apple-clang
+    conflicts("platform=linux")
+    conflicts("platform=cray")
+    conflicts("%gcc")
+    conflicts("%clang")
+
+    phases = []
+
+    sdk_base = (
+        "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/"
+        "Developer/SDKs/MacOSX"
+    )
+
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        # we try to setup a build environment with enough hints
+        # for the build system to pick up on the Apple framework version
+        # of OpenGL.
+        # - for a cmake build we actually needs nothing at all as
+        # find_package(OpenGL) will do the right thing
+        # - for the rest of the build systems we'll assume that
+        # setting the C_INCLUDE_PATH will be enough for the compilation phase
+        # and *** for the link phase.
+        env.prepend_path("C_INCLUDE_PATH", self.sdk_base)
+
+    @property
+    def headers(self):
+        return HeaderList(
+            "{}.sdk/System/Library/Frameworks/OpenGL.framework/Headers".format(self.sdk_base)
+        )
+
+    @property
+    def libs(self):
+        return LibraryList(
+            "{}.sdk/System/Library/Frameworks/OpenGL.framework".format(self.sdk_base)
+        )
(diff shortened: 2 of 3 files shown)


var/spack/repos/builtin/packages/apple-glu/package.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
```
tldahlgren:
```suggestion
# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```

var/spack/repos/builtin/packages/apple-opengl/package.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
```
tldahlgren:
```suggestion
# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```

tldahlgren (changes requested):
Only chiming in on the copyright ATM.

var/spack/repos/builtin/packages/apple-glu/package.py
aphecetche: indeed, those packages have been sitting in my private repo for a while :wink:
now fixed.

var/spack/repos/builtin/packages/apple-glu/package.py
```diff
@@ -0,0 +1,56 @@
...
+
+    homepage = ""
+
+    maintainers = ["aphecetche"]
+
+    has_code = False
+
+    version("4.0.0")
+
+    provides("glu")
```
HadrienG2: Should probably be...

```suggestion
    provides("glu", when=sys.platform == "darwin")
```

...as in other package ?

var/spack/repos/builtin/packages/apple-glu/package.py
```diff
@@ -0,0 +1,56 @@
...
+        # setting the C_INCLUDE_PATH will be enough for the compilation phase
+        # and *** for the link phase.
+        env.prepend_path(
+            "C_INCLUDE_PATH",
+            (
+                "/Applications/Xcode.app/Contents/Developer/",
+                "Platforms/MacOSX.platform/Developer/SDKs/MacOSX",
+            ),
+        )
+        # env.set("SPACK_APPLE_GLU","1")
```
HadrienG2: Commented code should go away unless it is meant to be uncommented.

var/spack/repos/builtin/packages/apple-opengl/package.py
```diff
@@ -0,0 +1,68 @@
...
+        # setting the C_INCLUDE_PATH will be enough for the compilation phase
+        # and *** for the link phase.
+        env.prepend_path(
+            "C_INCLUDE_PATH",
+            (
+                "/Applications/Xcode.app/Contents/Developer/",
+                "Platforms/MacOSX.platform/Developer/SDKs/MacOSX",
+            ),
+        )
+        # env.set("SPACK_APPLE_OPENGL","1")
```
HadrienG2: See above.

var/spack/repos/builtin/packages/apple-opengl/package.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    def headers(self):
+        return HeaderList(
+            (
+                "/Applications/Xcode.app/Contents/Developer/",
+                "Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/",
+                "Library/Frameworks/OpenGL.framework/Headers",
+            )
+        )
```
HadrienG2: Doesn't AppleGlu need a variation of that bit as well ?

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -211,11 +211,12 @@ class Root(CMakePackage):
...
-    depends_on("glew", when="+x+opengl")
-    depends_on("gl", when="+x+opengl")
-    depends_on("glu", when="+x+opengl")
-    depends_on("gl2ps", when="+x+opengl")
+    depends_on("ftgl@2.4.0:", when="+opengl")
+    depends_on("glew", when="+opengl")
+    depends_on("gl2ps", when="+opengl")
+    if sys.platform != "darwin":
+        depends_on("gl", when="+x+opengl")
+        depends_on("glu", when="+x+opengl")
```
HadrienG2: To pick up AppleGlu and AppleOpenGL, you probably need some sort of `depends_on("gl", when="+opengl")` even on Darwin + same for glu ?

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -428,8 +430,11 @@ def cmake_args(self):
 
         # Options related to ROOT's ability to download and build its own
         # dependencies. Per Spack convention, this should generally be avoided.
+
         options += [
-            define_from_variant("builtin_afterimage", "x"),
+            define_from_variant("builtin_afterimage", "x")
+            if "platform=darwin" not in self.spec
+            else define("builtin_afterimage", True),
```
HadrienG2: I think separately computing an `afterimage_enabled = "+x" in self.spec or "platform=darwin" in self.spec` boolean then doing a single `define("builtin_afterimage", afterimage_enabled)` option would be clearer.

HadrienG2 (changes requested):

var/spack/repos/builtin/packages/apple-glu/package.py
aphecetche: indeed.
I should also specify the version (glu@1.3) I guess. 
will change.

var/spack/repos/builtin/packages/apple-glu/package.py
aphecetche: ok

var/spack/repos/builtin/packages/apple-opengl/package.py
aphecetche: I only tested this within the `root` recipe, which uses CMake's `find_package(OpenGL)` and CMake was happy as is. Probably because the `glu.h` is in fact colocated with the `gl.h` one ?
Anyway I can certainly add it, yes.

var/spack/repos/builtin/packages/root/package.py
aphecetche: your comment makes sense. Will have to cross-check carefully why it worked (or so I believed...)  without.

var/spack/repos/builtin/packages/root/package.py
aphecetche: ok

aphecetche:
while browsing the other gl related packages (e.g. mesa and mesa-glu) I realise I should probably rename the `apple-opengl` package to `apple-gl` instead ?

var/spack/repos/builtin/packages/root/package.py
aphecetche: actually I think I'm mixing things up here. Most probably the `apple-g`l and `apple-glu` are actually not needed at all for the `root` recipe as CMake `find_package(OpenGL)` is correctly finding the Apple headers and libraries.
Those are needed in other packages I have in my stack.
Will confirm this and if that's the case then this PR will be much simpler :wink:

aphecetche:
actually I think I'm mixing things up here. Most probably the `apple-g`l and `apple-glu` are actually not needed at all for the `root` recipe as CMake `find_package(OpenGL)` is correctly finding the Apple headers and libraries.
Those are needed in other packages I have in my stack.
Will confirm this and if that's the case then this PR will be much simpler :wink:

aphecetche:
FYI I've implemented some changes and forced pushed.

I've changed `apple-opengl` to `apple-gl` and fixed the `C_INCLUDE_PATH` in both packages that was not a string but a tuple (my bad, was trying to pacify the black formatter which was complaining about too long lines and I added commas where there should not have...)

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -211,11 +211,15 @@ class Root(CMakePackage):
...
-    depends_on("gl2ps", when="+x+opengl")
+    depends_on("ftgl@2.4.0:", when="+opengl")
+    depends_on("glew", when="+opengl")
+    depends_on("gl2ps", when="+opengl")
+    if sys.platform != "darwin":
+        depends_on("gl", when="+x+opengl")
+        depends_on("glu", when="+x+opengl")
+    else:
+        depends_on("gl", when="+opengl")
+        depends_on("glu", when="+opengl")
```
HadrienG2: Given that `conflicts("+opengl", when="~x")` if `sys.platform != darwin`, I think this is equivalent to the simpler...

```suggestion
    depends_on("gl", when="+opengl")
    depends_on("glu", when="+opengl")
```

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -428,8 +433,11 @@ def cmake_args(self):
 
         # Options related to ROOT's ability to download and build its own
         # dependencies. Per Spack convention, this should generally be avoided.
+
+        afterimage_enabled = "+x" if "platform=darwin" not in self.spec else True
```
HadrienG2: I think that should be...

```suggestion
        afterimage_enabled = ("+x" in self.spec) if "platform=darwin" not in self.spec else True
```

HadrienG2 (changes requested):

HadrienG2 (dismissed):

var/spack/repos/builtin/packages/apple-gl/package.py
```diff
@@ -0,0 +1,58 @@
...
+
+    homepage = "https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_intro/opengl_intro.html"
+
+    maintainers = ["aphecetche"]
+
+    has_code = False
+
+    version("4.1.0")
+
+    provides("gl@4.1", when=sys.platform == "darwin")
```
tldahlgren: The `when` option is not needed here or in the `apple-glu`.

Conceptually, a package provides the interface/API (or not) typically for specific version(s) and or variant(s).  Then `conflicts` are used if it doesn't work or apply.

var/spack/repos/builtin/packages/apple-glu/package.py
```diff
@@ -0,0 +1,58 @@
...
+
+    homepage = ""
+
+    maintainers = ["aphecetche"]
+
+    has_code = False
+
+    version("1.3.0")
+
+    provides("glu@1.3", when=sys.platform == "darwin")
```
tldahlgren: As mentioned above, this is not needed as the `conflicts()` address the support issue.

tldahlgren (changes requested):

var/spack/repos/builtin/packages/apple-gl/package.py
aphecetche: fixed. thanks.

HadrienG2 (dismissed):

aphecetche:
@tldahlgren I believe I've addressed the comments, but I don't exactly know how to clear the "1 change requested" status ...

tldahlgren (approved):

PR #29175, opened 2022-02-24
py-ipython: add 8.0.1

manuelakuhn:
https://github.com/ipython/ipython/tree/8.0.1

PR #32664, opened 2022-09-15
Add skip_import to PythonPackage and use it in py-nilearn

manuelakuhn:
A while ago we discussed to add `skip_import` functionality to the `PythonPackage` (see #29175). Instead of explicitly defining `import_modules` for each version or variant, this enables to use the automatic detected modules and ignores certain ones.

For the `py-nilearn` example using `import_modules` would mean:

```
    @property
    def import_modules(self):
        modules = [
            'nilearn',
            'nilearn.connectome',
            'nilearn.datasets',
            'nilearn.decoding',
            'nilearn.decomposition',
            'nilearn.image',
            'nilearn.input_data',
            'nilearn.masking',
            'nilearn.mass_univariate',
            'nilearn.regions',
            'nilearn.signal',
            'nilearn.surface',
        ]

        if self.spec.satisfies('@0.6:'):
            modules.append('nilearn.externals')

        if self.spec.satisfies('@0.7:'):
            modules.append('nilearn.glm')

        if '+plotting' in self.spec:
            modules.append('nilearn.plotting')
            if self.spec.satisfies('@0.7:'):
                modules.append('nilearn.reporting')

        return modules
```

whereas with `skip_modules` this can be reduced to
```
    @property
    def skip_modules(self):
        modules = []

        if self.spec.satisfies("~plotting"):
            modules.append("nilearn.plotting")
            if self.spec.satisfies("@0.7:"):
                modules.append("nilearn.reporting")

        return modules
```
ping @adamjstewart

### DIFF
--- a/var/spack/repos/builtin/packages/py-nilearn/package.py
+++ b/var/spack/repos/builtin/packages/py-nilearn/package.py
@@ -28,13 +28,22 @@ class PyNilearn(PythonPackage):
 
     variant("plotting", default=False, description="Enable plotting functionalities")
 
+    depends_on("py-joblib@0.15:", when="@0.9.1:", type=("build", "run"))
+    depends_on("py-joblib@0.12:", when="@0.7:", type=("build", "run"))
+    depends_on("py-joblib@0.11:", when="@0.6:", type=("build", "run"))
+    depends_on("py-lxml", when="@0.9.1:", type=("build", "run"))
+    depends_on("py-nibabel@3:", when="@0.9.1:", type=("build", "run"))
+    depends_on("py-nibabel@2.5:", when="@0.8:", type=("build", "run"))
+    depends_on("py-nibabel@2.0.2:", type=("build", "run"))
+    depends_on("py-numpy@1.18:", when="@0.9.1:", type=("build", "run"))
     depends_on("py-numpy@1.16:", when="@0.8:", type=("build", "run"))
     depends_on("py-numpy@1.11:", when="@0.5:", type=("build", "run"))
     depends_on("py-numpy@1.6.1:", type=("build", "run"))
-    depends_on("py-scipy@1.2:", when="@0.8:", type=("build", "run"))
-    depends_on("py-scipy@0.19:", when="@0.6:", type=("build", "run"))
-    depends_on("py-scipy@0.17:", when="@0.5:", type=("build", "run"))
-    depends_on("py-scipy@0.14:", type=("build", "run"))
+    depends_on("py-pandas@1:", when="@0.9.1:", type=("build", "run"))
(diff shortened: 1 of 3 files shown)


var/spack/repos/builtin/packages/py-nilearn/package.py
```diff
@@ -53,29 +56,10 @@ class PyNilearn(PythonPackage):
     depends_on("py-matplotlib@1.1.1:3.3", when="@:0.6 +plotting", type=("build", "run"))
```
adamjstewart: Dependencies need to be updated:
```
install_requires =
        joblib>=0.15
        lxml
        nibabel>=3.0.0
        numpy>=1.18
        pandas>=1.0
        requests>=2
        scikit-learn>=0.22
        scipy>=1.5

[options.extras_require]
plotting =
        matplotlib>=3.0
```

var/spack/repos/builtin/packages/py-nilearn/package.py
```diff
@@ -53,29 +56,10 @@ class PyNilearn(PythonPackage):
...
-            modules.append("nilearn.externals")
-
-        if self.spec.satisfies("@0.7:"):
-            modules.append("nilearn.glm")
-
-        if "+plotting" in self.spec:
+        if self.spec.satisfies("~plotting"):
             modules.append("nilearn.plotting")
             if self.spec.satisfies("@0.7:"):
                 modules.append("nilearn.reporting")
```
adamjstewart, line 74: Is `nilearn.reporting` installed with nilearn 0.6? If not, you can remove this, as `import_modules` won't auto-detect something that doesn't exist, so there's nothing to skip.

If yes, does `nilearn.reporting` require `+plotting`? If not, this shouldn't be indented under the same if-statement.

adamjstewart (commented):
Actual feature and docs look great, just have a couple questions about nilearn.

var/spack/repos/builtin/packages/py-nilearn/package.py
manuelakuhn: Sorry, I seems that I only checked the dependencies for version 0.9.0 and missed the once for 0.9.2. This is fixed now.

var/spack/repos/builtin/packages/py-nilearn/package.py
manuelakuhn: Yes, `nilearn.reporting` is installed with `nilearn@0.6` but does not require a `matplotlib` dependency. Starting from `nilearn@0.7`, `nilearn.reporting` requires `matplotlib` to work.

adamjstewart (approved):

PR #32666, opened 2022-09-15
patch etsf-io tests

iamashwin99:
The package currently fails to install as the package fails at the `make check` phase.
The PR includes two patches to get the package working.
Both these patches are being applied to the etsf_io testing system and not the library itself.
Patches from meisam.tabriz@mpcdf.mpg.de and smoke tests from @fangohr 
I am only sending the PR with their permission.


tldahlgren (approved):
Confirmed the stand-alone tests pass after the package is installed.

PR #29761, opened 2022-03-28
Enable/Disable Propagation for Compiler Flags & Variants

RikkiButler20:
Fixes #29653.

Allow for inheritance to be enabled or disabled for both variants and compiler flags.

- [x] Add new (spec language) operators
- [x] Apply operators to variants and compiler flags
- [x] Resolve conflicts
- [x] Add tests
- [x] Add documentation

PR #32680, opened 2022-09-16
BabelStream package

jessj0nes:
Measure memory transfer rates to/from global device memory on GPUs. This benchmark is similar in spirit, and based on, the STREAM benchmark for CPUs.

New package with multiple implementations of BabelStream:

- OpenCL
- CUDA
- HIP
- OpenACC
- OpenMP
- C++ Parallel STL
- Kokkos
- RAJA
- SYCL and SYCL 2020
- TBB
- Thrust (via CUDA or HIP)


tldahlgren:
@rob64j @tom91136 Are you okay with being included as maintainers on this package?

jessj0nes:
> @rob64j @tom91136 Are you okay with being included as maintainers on this package?

- Yes

tom91136:
> @rob64j @tom91136 Are you okay with being included as maintainers on this package?

Yes

var/spack/repos/builtin/packages/babelstream/package.py
```diff
@@ -0,0 +1,189 @@
...
+
+class Babelstream(CMakePackage):
+    """Measure memory transfer rates to/from global device memory on GPUs.
+    This benchmark is similar in spirit, and based on, the STREAM benchmark for CPUs."""
+
+    homepage = "https://github.com/UoB-HPC/BabelStream"
+    git = "https://github.com/UoB-HPC/BabelStream.git"
+    url = "https://github.com/UoB-HPC/BabelStream/archive/refs/tags/v4.0.tar.gz"
+
+    maintainers = ["rob64j", "tomdeakin", "tom91136"]
```
alalazo: One username seems wrong:
```suggestion
    maintainers = ["robj0nes ", "tomdeakin", "tom91136"]
```

alalazo (changes requested):

var/spack/repos/builtin/packages/babelstream/package.py
```diff
@@ -0,0 +1,189 @@
...
+            "acc",
+            "raja",
+            "tbb",
+            "thrust",
+        ],
+        default="std-data",
+        description="Specify the model to be built",
+    )
+
+    variant("flags", values=str, default=" ", description="Additional CXX flags to be provided")
```
alalazo, line 55: This might need to be revisited after we merge #29761 Can you add a TODO on top of the variant?

var/spack/repos/builtin/packages/babelstream/package.py
```diff
@@ -0,0 +1,189 @@
...
+
+    variant(
+        "amd_arch", values=str, default="none", description="Target AMD GPU device being used."
+    )
+
+    variant(
+        "cuda_arch", values=str, default="none", description="The CUDA architecture being used."
+    )
```
alalazo, line 63: Can't we use `CudaPackage` and `RocmPackage` instead of duplicating the variants?

alalazo (changes requested):
A few other comments on directives

var/spack/repos/builtin/packages/babelstream/package.py
tomdeakin: Aren't these classes the whole package would need to inherit from? How do you select this based on the user choice of variant?

alalazo: Not sure I get what you're asking. Those two base classes are mixins, defining the `+cuda` variant etc. and a list of valid architecture for CUDA and ROCm (instead of accepting any string).

tomdeakin: Your Python is better than mine! Are you are suggesting we change `class Babelstream(CMakePackage)` to `class Babelstream(CMakePackage,CudaPackage,RocmPackage)`?

Does that not break the abstraction of the `model` variant here? I.e., we'd need to say `model=std-data` for that variant but spell it `+cuda` for the CUDA variant?

Further, how do we ensure that one-and-only-one variant is selected if we go via the mixins?

bernhardkaindl: @robj0nes can you follow up in any way? Thanks!

tomdeakin: I’m taking over progressing this. Is there an update to my question on the mixins?

alalazo: > Further, how do we ensure that one-and-only-one variant is selected if we go via the mixins?

You can use `conflicts` directives to disallow part of the search space. 

> Does that not break the abstraction of the model variant here? I.e., we'd need to say model=std-data for that variant but spell it +cuda for the CUDA variant?

I can't say much on the `model` variant, since this is package specific. It seems to me though that you can use conditional values and say the `cuda` value to `model` can be used only when `+cuda` is on:
```
        values=[
            "std-data",
            "std-indices",
            "std-ranges",
            "ocl",
            "omp",
            "hip",
            conditional("cuda", when="+cuda"),
            "kokkos",
            "sycl",
            "sycl2020",
            "acc",
            "raja",
            "tbb",
            "thrust",
        ],
```

tomdeakin: Thanks for the feedback.

The `model` option of the package as it is aims to provide identical functionality to the CUDA and ROCM mixins, but for *all* models. Spelling it `+cuda` but not `model=cuda` for CUDA, but `model=foo` for everything else is not sensible.

We're going to need to redesign the programming model selection to align with the mixin approach where that's supported, but provide the `+foo` notation where there is no mixin. I'm going to use the [Kokkos package](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/kokkos/package.py) as inspiration for implementing this as there is similarity in the mechanics of choosing a back end programming model between the two packages.

Watch this space for a rewrite!

alalazo: What is the status of this PR?

tomdeakin: We are working on an update which uses the mixins. We have something working on some systems and are testing on others. We will push the update soon for your review.

tomdeakin:
Closing in favour of #36164

PR #32688, opened 2022-09-16
Attempt at patching phist find mpi heuristic for updated phist version

johnwparent:
This is a tentative fix for the current Gitlab CI issues plaguing develop where phist fails due to GCC complaining about a type mismatch in some calls to MPI.

This solve is based on a previous issue with phist, where its bundled findMPI.cmake is conflicting with Spack derived MPI location determination.

### DIFF
--- a/var/spack/repos/builtin/packages/phist/package.py
+++ b/var/spack/repos/builtin/packages/phist/package.py
@@ -192,7 +192,7 @@ class Phist(CMakePackage):
     # the phist repo came with it's own FindMPI.cmake before, which may cause some other
     # MPI installation to be used than the one spack wants.
     def patch(self):
-        if self.spec.satisfies("@1.9.6"):
+        if self.spec.satisfies("@1.9.6:"):
             filter_file("USE mpi", "use mpi_f08", "src/kernels/builtin/crsmat_module.F90")
             # filter_file('use mpi', 'use mpi_f08', -> Needs more fixes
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/phist/package.py
```diff
@@ -192,7 +192,7 @@ class Phist(CMakePackage):
     # the phist repo came with it's own FindMPI.cmake before, which may cause some other
     # MPI installation to be used than the one spack wants.
     def patch(self):
-        if self.spec.satisfies("@1.9.6"):
+        if self.spec.satisfies("@1.9.6:"):
```
jthies, line 195: This patch undoes the switch from the unsafe 'mpi' to the better 'mpi_f08' module which is unfortunate. The actual fix should be to make it possible to query wether an mpi spec provides mpi_f08, and patch phist only if it doesn't.
Still I think this is OK for now.

jthies (approved):

var/spack/repos/builtin/packages/phist/package.py
johnwparent: This PR is intended as a stopgap to allow the CI to continue functioning without error, but thank you for pointing this out. I'll leave a TODO to indicate that this should not languish in its current state.

adamjstewart (approved):
Hit this same problem in CI. Let's merge for now and think about a better long term solution at our leisure.

PR #32689, opened 2022-09-16
Don't run bootstrap on package only PRs

alalazo:
Currently we always run bootstrap jobs in CI, due to a possibly wrong translation from GitHub Action's path filters to `dorny/path-filters` action. Here we limit bootstrap jobs to PRs that modify core or packages involved in bootstrap.

Modifications:
- [x] Change the file paths used to determine if the `bootstrap` is run

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -73,15 +73,15 @@ jobs:
       # setting environment variables from earlier steps: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
       #
   bootstrap:
-    if: ${{ github.repository == 'spack/spack' && needs.changes.outputs.bootstrap == 'true' }}
+    if: ${{ github.repository == 'spack/spack' && needs.changes.outputs.bootstrap == 'false' }}
     needs: [ prechecks, changes ]
     uses: ./.github/workflows/bootstrap.yml
   unit-tests:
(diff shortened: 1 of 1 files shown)


trws:
I like this change, but looking at it makes me wonder, should we expand the bootstrap testing to include the packages that are required by the bootstrap set?  OpenSSL for example can be kinda fragile, but it's required on most systems to build cmake to build clingo.

alalazo:
That would be good in principle, but to keep things kind of short we do:
```console
$ spack external find cmake bison
```
when trying bootstrapping from sources. The reasoning behind that was that `cmake` is built for sure in pipelines, so if it has to fail it will fail there.

trws:
Ok sweet, so when tests are passing I'm all good with this.

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

trws (approved):

tgamblin:
@alalazo @trws are you sure we shouldn't change this to include the `package.py` files for `clingo` and its dependencies?  They can still affect the build even if you `spack external find cmake bison`.

trws:
It does include clingo, and re2c for some reason even though we 
shouldn't need that for bootstrap if we're bootstrapping from release 
tarballs.  It just doesn't include *all* of the dependencies, which is 
mainly cmake.

tgamblin:
Ok never mind -- I paid attention to the `-` part but not the unchanged part.  Yep I think this is pretty good.

PR #32690, opened 2022-09-16
Support for building Pmix with Debian/Ubuntu external dependencies

asarkar-parsys:
Debian like distros use multiarch implementation spec
https://wiki.ubuntu.com/MultiarchSpec
Instead of being limited to /usr/lib64, architecture based lib directories are used. For instance, under ubuntu a library package on x86_64 installs binaries under /usr/lib/x86_64-linux-gnu. Building pmix with external dependencies like hwloc or libevent fail as with prefix set to /usr, that prefix works for headers and binaries but does not work for libraries. The default location for library /usr/lib64 does not hold installed binaries. Pmix build options --with-libevent and --with-libhwloc allow us to specify dependent library locations. This commit is an effort to highlight and resolve such an issue when a users want to use Debian like distro library packages and use spack to build pmix. There maybe other packages that might be impacted in a similar way.

### DIFF
--- a/var/spack/repos/builtin/packages/pmix/package.py
+++ b/var/spack/repos/builtin/packages/pmix/package.py
@@ -103,7 +103,7 @@ class Pmix(AutotoolsPackage):
         tgt_libpath = ""
         dir_list = spec[pkg_name].libs
         for entry in dir_list:
-            if path_match_str=="" or (path_match_str!="" and path_match_str in entry):
+            if path_match_str == "" or (path_match_str != "" and path_match_str in entry):
                 tgt_libpath = entry
                 break
         path_list = tgt_libpath.split(os.sep)
@@ -118,12 +118,17 @@ class Pmix(AutotoolsPackage):
         config_args.append("--with-libevent=" + spec["libevent"].prefix)
         config_args.append("--with-hwloc=" + spec["hwloc"].prefix)
 
+        # As of 09/22/22 pmix build does not detect the hwloc version
+        # for 32-bit architecture correctly. Since, we have only been
+        # able to test on 64-bit architecture, we are keeping this
+        # check for "64" in place. We will need to re-visit this when we
+        # have the fix in Pmix for 32-bit library version detection
         if "64" in platform.machine():
             if spec["libevent"].external_path:
-                dep_libpath = self.find_external_lib_path("libevent","64")
+                dep_libpath = self.find_external_lib_path("libevent", "64")
                 config_args.append("--with-libevent-libdir=" + dep_libpath)
             if spec["hwloc"].external_path:
-                dep_libpath = self.find_external_lib_path("hwloc","64")
+                dep_libpath = self.find_external_lib_path("hwloc", "64")
                 config_args.append("--with-hwloc-libdir=" + dep_libpath)
 
         config_args.extend(self.enable_or_disable("python-bindings", variant="python"))
--- a/var/spack/repos/builtin/packages/pmix/package.py
+++ b/var/spack/repos/builtin/packages/pmix/package.py
@@ -3,7 +3,11 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import os
+import re
+import platform
+import spack.platforms
 
+import llnl.util.filesystem as filesystem
 from spack.package import *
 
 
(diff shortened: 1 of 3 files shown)


rhc54 (commented):
Sorry for the delay. Is it true that ubuntu will _always_ have installed libevent and hwloc in those alternate locations? I ask because I believe sys admins have the ability to redirect the installation to any location they desire, and this is hardwired to a specific location. Will that be a problem?

asarkar-parsys:
> Sorry for the delay. Is it true that ubuntu will _always_ have installed libevent and hwloc in those alternate locations? I ask because I believe sys admins have the ability to redirect the installation to any location they desire, and this is hardwired to a specific location. Will that be a problem?

@rhc54 That's a good question and I had a similar query while working on the fix. As far as I know, Ubuntu follows File Hierarchy Standard and apt-get does not provide a mechanism to deviate from it. I believe redirections have to happen through symlinks from the standard directories. I am not an expert and so it will be good to know the answer to how would one change the default installation directories on Ubuntu. The following response to a similar question answers that
https://askubuntu.com/questions/165361/how-to-set-the-path-where-sudo-apt-get-installs

Though having a hard-coded location does seem restrictive. However, could this be handled better if the build method was cmake based in this case? @trws I think cmake find_package should be able to find these locations with "prefix=/usr". If so, that makes me question if solution should come from spack or the package's build methos itself?

trws:
Really find_package should be finding it in either location, if it's not it's probably use of find_library or similar without including the multiarch paths in the list of possibilities.  We shouldn't force it to look only in those places, because an admin could have manually built those and put them anywhere, but we could add it to the list or we could specify the path to those dependencies with the <name>_DIR parameters on the command line to work around it.

asarkar-parsys:
> Really find_package should be finding it in either location, if it's not it's probably use of find_library or similar without including the multiarch paths in the list of possibilities. We shouldn't force it to look only in those places, because an admin could have manually built those and put them anywhere, but we could add it to the list or we could specify the path to those dependencies with the _DIR parameters on the command line to work around it.

Thanks for the feedback @rhc54 @trws. pmix' build is not using cmake and the "auto-detect" it defaults to looking for libraries under /usr/lib64. I agree that we shouldn't force hardcoded locations and the new patch addresses that. Though it adds the @libs properties to both hwloc and libevent, which is being queried by pmix. I think it is ready for another quick round of review.

var/spack/repos/builtin/packages/pmix/package.py
```diff
@@ -101,6 +114,14 @@ def configure_args(self):
         config_args.append("--with-libevent=" + spec["libevent"].prefix)
         config_args.append("--with-hwloc=" + spec["hwloc"].prefix)
 
+        if "64" in platform.machine():
```
trws, line 126: I'm mostly happy with this PR, but is this check correct? Does it work on 32-bit?

asarkar-parsys: I tried installing the hwloc version 2.1 available with ubuntu20.04 (i386 version). Then I tried building pmix with spack without the 64-bit check. Pmix build failed in the configuration stage itself as it incorrectly deemed the version of hwloc < 1.5. So, I am not even sure if pmix build needs to be fixed before trying to add stuff to spack. Since, I've tested this only on a 64-bit platform and it works with 64-bit library binaries, that's why I have left this check.

I wonder if we can work on the 32-bit support when needed and have this 64-bit check in place at the moment. If that seems reasonable then I can leave a comment explaining how to use find_external_lib_path to accomplish 32-bit support.

rhc54:
> Pmix build failed in the configuration stage itself as it incorrectly deemed the version of hwloc < 1.5.

That seems rather odd - I don't have an immediate way to debug it, but will definitely look into it when time permits.

asarkar-parsys:
> > Pmix build failed in the configuration stage itself as it incorrectly deemed the version of hwloc < 1.5.
> 
> That seems rather odd - I don't have an immediate way to debug it, but will definitely look into it when time permits.

@rhc54 The platform is 64-bit and so spack is building pmix for 64-bit. I only wanted to see if my changes pick /usr/lib/i386-linux-gnu for building, which it did but instead of link failing, it failed in the configuration stage. Kindly, find the parts of the log attached below.
```
 
==> pmix: Executing phase: 'configure'
==> [2022-09-20-21:03:05.368447] '/tmp/root/spack-stage/spack-stage-pmix-4.1.2-zgufnbg36spljnxxianteqj5kxn4fbe5/spack-src/configure' '--prefix=/root/spack/opt/spack/linux-ubuntu20.04-sandybridge/gcc-9.4.0/pmix-4.1.2-zgufnbg36spljnxxianteqj5kxn4fbe5' '--enable-shared' '--enable-static' '--with-libevent=/usr' '--with-hwloc=/usr' '--with-libevent-libdir=/usr/lib/x86_64-linux-gnu' **'--with-hwloc-libdir=/usr/lib/i386-linux-gnu'** '--enable-pmi-backward-compatibility' '--disable-man-pages'

checking for header hwloc.h in /usr... no
checking for header hwloc.h in /usr/include... yes
checking library hwloc in /usr/lib/i386-linux-gnu... found
**checking if hwloc version is 1.5 or greater... no**
: error: Cannot continue
```

rhc54:
Okay, I won't put a lot of worry into it for now. I suspect the problem is that we do a "try_compile" of nothing more than the HWLOC version header to try and determine the HWLOC version, and that is failing due to the 32-bit environment.

asarkar-parsys:
@spackbot run pipeline

asarkar-parsys:
@spackbot run pipeline

asarkar-parsys:
@spackbot run pipeline

eugeneswalker:
@spackbot run pipeline

trws (approved):

PR #32692, opened 2022-09-16
Update OMB to v6.1

natshineman:
Update to v6.1 released 16SEP2022

### DIFF
--- a/var/spack/repos/builtin/packages/osu-micro-benchmarks/package.py
+++ b/var/spack/repos/builtin/packages/osu-micro-benchmarks/package.py
@@ -16,10 +16,11 @@ class OsuMicroBenchmarks(AutotoolsPackage, CudaPackage):
     and can be used for both traditional and GPU-enhanced nodes."""
 
     homepage = "https://mvapich.cse.ohio-state.edu/benchmarks/"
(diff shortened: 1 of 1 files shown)


alalazo (dismissed):

tldahlgren (approved):
Confirmed sha256.

var/spack/repos/builtin/packages/osu-micro-benchmarks/package.py
```diff
@@ -20,7 +20,7 @@ class OsuMicroBenchmarks(AutotoolsPackage, CudaPackage):
 
     maintainers = ["natshineman", "harisubramoni", "MatthewLieber"]
 
-    version("6.1", sha256="7f3be7526411449a765797ee1db7cee14a79a3d536e01063703457469fb9d0a3")
+    version("6.1", sha256="ecccedc868264f75db4d9529af79005419a2775113c7fae8f4e4a8434362e4a7")
```
alalazo, line 23: Did anything change in the package? Do we have to expect changes without version bumping for `osu-micro-benchmarks`?

natshineman: We identified an error in the CHANGELOG that indicated the wrong flag for a newly added feature. This change was purely documentation. 

In the future this should not be an issue and we do not expect further changes to the package after release.

PR #29039, opened 2022-02-17
py-pyqt5: add 5.15.6 and get sources from pypi

manuelakuhn:
General: at the moment `py-pyqt5` does not build without #28870 (also before this PR).

The build system has changed to `sip-install` for newer versions.

Installs with `spack install --test=root` without problems for all 4 versions.

Side note: The tarballs for versions <=5.13 are not available on the original sources anymore but only via the spack mirror.

PR #21554, opened 2021-02-09
Add version 5.15.2 for PyQt5

mwm126:

PR #32696, opened 2022-09-17
py-pyqt6: add new package

adamjstewart:
This PR includes the following changes:

- [x] Modernize SIPPackage and SIPBuilder
- [x] Get py-pyqt4 to build
- [x] Get py-pyqt5 to build
- [x] Get py-pyqt6 to build
- [x] Test py-matplotlib
- [x] Update SIPPackage documentation

Closes #29039 and #21554


Sinan81 (dismissed):
Looks good. Approved.

var/spack/repos/builtin/packages/qt-base/package.py
```diff
@@ -35,7 +35,7 @@ class QtBase(CMakePackage):
     variant("gui", default=True, description="Build the Qt GUI module and dependencies.")
     variant("shared", default=True, description="Build shared libraries.")
     variant("sql", default=True, description="Build with SQL support.")
-    variant("network", default=True, description="Build with SSL support.")
+    variant("network", default=False, description="Build with SSL support.")
```
adamjstewart: @wdconinc @sethrj this is a (temporary?) workaround for #34037

var/spack/repos/builtin/packages/qt/package.py
```diff
@@ -338,8 +338,6 @@ class Qt(Package):
             msg="Apple Silicon requires a very new version of qt",
         )
 
-    use_xcode = True
```
adamjstewart: @sethrj I found that the step where we copy Xcode and modify it to use Spack's compiler wrappers wasn't working for me. I get a prompt asking me to install Xcode command line tools even though they're already installed. If I comment this out, the Qt build succeeds, but Qt crashes at runtime because it can't find libharfbuzz (probably because of the lack of compiler wrappers). Ever seen this before? Any ideas on how we could fix this?

var/spack/repos/builtin/packages/py-sip/package.py
```diff
@@ -40,9 +40,9 @@ class PySip(PythonPackage):
 
     with when("@5:"):
         depends_on("python@3.5.1:", type=("build", "link", "run"))
-        depends_on("py-packaging", type="build")
-        depends_on("py-setuptools@30.3:", type="build")
-        depends_on("py-toml", type="build")
+        depends_on("py-packaging", type=("build", "run"))
+        depends_on("py-setuptools@30.3:", type=("build", "run"))
+        depends_on("py-toml", type=("build", "run"))
```
adamjstewart: Sip crashes at run-time because it can't find these required deps

var/spack/repos/builtin/packages/py-pyqt6/package.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyPyqt6(SIPPackage):
```
adamjstewart, line 9: Was planning on using `PythonPackage` but couldn't figure out how to pass `--confirm-license` to the build.

var/spack/repos/builtin/packages/py-pyqt5/package.py
```diff
@@ -12,45 +12,30 @@ class PyPyqt5(SIPPackage):
     Windows, OS X, Linux, iOS and Android. PyQt5 supports Qt v5."""
 
     homepage = "https://www.riverbankcomputing.com/software/pyqt/intro"
-    url = (
-        "https://www.riverbankcomputing.com/static/Downloads/PyQt5/5.13.0/PyQt5_gpl-5.13.0.tar.gz"
```
adamjstewart, line 16: This URL doesn't work and all older versions are unavailable, so I deprecated them

var/spack/repos/builtin/packages/py-pyqt4/package.py
```diff
@@ -13,39 +13,55 @@ class PyPyqt4(SIPPackage):
...
-            args.extend(["--qsci-api", "--qsci-api-destdir", self.prefix.share.qsci])
         return args
+
+    def configure(self, spec, prefix):
+        python("configure-ng.py", *self.configure_args())
+
+    @run_after("install")
+    def extend_path_setup(self):
+        # See github issue #14121 and PR #15297
+        module = self.pkg.spec["py-sip"].variants["module"].value
```
adamjstewart: This stuff was in the base class, but this variant only exists for ancient sip, so it was failing. I don't think this is needed for most modern sip packages.

lib/spack/spack/build_systems/sip.py
```diff
@@ -130,34 +130,17 @@ class SIPBuilder(BaseBuilder):
...
-                inspect.getmodule(self.pkg).python_platlib,
-            ]
-        )
 
-        self.pkg.python(configure, *args)
+        # https://www.riverbankcomputing.com/static/Docs/sip/command_line_tools.html
+        args = ["--verbose", "--target-dir", inspect.getmodule(self.pkg).python_platlib]
+        args.extend(self.configure_args())
+
+        sip_build = Executable(spec["py-sip"].prefix.bin.join("sip-build"))
```
adamjstewart, line 142: sip-build replaced configure.py a long time ago, pyqt4 should be the only one that needs the old syntax.

var/spack/repos/builtin/packages/qt/package.py
adamjstewart: @tgamblin @davydden @mathstuf I think all of you have touched the `use_xcode` logic in the past. Any idea what might have changed on newer versions of Xcode (or arm64)?

adamjstewart: Exact error message:
```
xcode-select: Failed to locate 'make', requesting installation of command line developer tools.
```
But this command is installed on the OS, so I'm not sure why it can't be found.

mathstuf: `/usr/bin/make` probaby uses `xcrun` to find an Xcode and run that. `use_xcode` tries to make a "fake" Xcode that spack can reliably inject its compiler wrapper into. Note that with Qt6, this might not be necessary anymore as it uses CMake.

adamjstewart: Yep, Qt6 doesn't need it. How do I get older Qt working?

sethrj: My colleagues and I have been seeing this problem with Qt5 since probably xcode 13... our solution has been to use homebrew 😢

mathstuf: An issue to that effect would be nice…not sure how it's supposed to be fixed without letting anyone know ;) . (While I did do the fake Xcode thing, I don't use Spack on macOS beyond testing that code, so…)

adamjstewart: Done: #34064

var/spack/repos/builtin/packages/py-pyqt5/package.py
Sinan81: I remember noticing this issue a while ago. If necessary, we might be able to use internet archive's wayback machine to get older tars.

var/spack/repos/builtin/packages/py-sip/package.py
Sinan81: makes sense.

var/spack/repos/builtin/packages/py-pyqt6/package.py
Sinan81: :thinking:

adamjstewart:
Update: this PR is somewhat blocked by my ability to test it. I'll be moving to an older laptop soon, so the Qt installation should hopefully work. I'm hoping to get back to this then. Once I can build multiple versions of Qt, it will be much easier to test whether or not the hacks I've removed here are needed or not.

Sinan81:
I'll see if I can do some testing myself.

Sinan81:
I've just tried installing qgis in this branch, where qscintilla installation fails:
```
 >> 334    configure.py: error: '.../py-sip-6.6.2-5moeuknve3jqp5e74emtmz6ie5olycl6/bin/sip' is not an executable

```
which is expected I guess.

Sinan81:
Also, tried installing py-pyqt6, which failed since qt-base-6.3.2 failed to build.

Sinan81:
> Also, tried installing py-pyqt6, which failed since qt-base-6.3.2 failed to build.

qt-base build might have failed since I was using gcc@7 as opposed to newer.

Sinan81:
created an issue for it: https://github.com/spack/spack/issues/34418

var/spack/repos/builtin/packages/py-pyqt5/package.py
mathstuf: This is why Spack should re-mirror a lot of these things. With content-addressed lookup, the URL can just be `archive.spack.tld/archives/$algo/$hash`.

adamjstewart: Spack does re-mirror all URLs. On develop:
```console
$ spack fetch py-pyqt5
==> Fetching https://mirror.spack.io/_source-cache/archive/54/54b7f456341b89eeb3930e786837762ea67f235e886512496c4152ebe106d4af.tar.gz
```
However, I don't think anyone really cares about installing these older versions, and newer versions are available on PyPI, where releases actually can't be removed, only yanked. So this new URL should be more reliable for folks not using Spack's mirror.

mathstuf: Releases can definitely be deleted from PyPI; I've had to do it for VTK multiple times because we hit storage quotas.

adamjstewart: Ah, maybe it's that you can't rerelease the same version

mathstuf: Yes, once a release is deleted on PyPI, it is gone forever.

adamjstewart:
For py-pyqt4, I'm seeing the same issue on this branch as on develop:
```
c++ -c -std=gnu++98 -pipe -O2 -arch x86_64 -Xarch_x86_64 -mmacosx-version-min=10.5 -Wall -W -DQT_NO_DEBUG -DQT_CORE_LIB -DQT_SHARED -I/Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/mkspecs/macx-g++ -I. -I/Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib/QtCore.framework/Versions/4/Headers -I/Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/include/QtCore -I/Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/include -I. -F/Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib -o qtdetail.o qtdetail.cpp
clang: warning: include path for libstdc++ headers not found; pass '-stdlib=libc++' on the command line to use the libc++ standard library instead [-Wstdlibcxx-not-found]
In file included from qtdetail.cpp:1:
In file included from /Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib/QtCore.framework/Versions/4/Headers/QCoreApplication:1:
In file included from /Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib/QtCore.framework/Versions/4/Headers/qcoreapplication.h:45:
In file included from /Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib/QtCore.framework/Headers/qobject.h:47:
In file included from /Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib/QtCore.framework/Headers/qobjectdefs.h:45:
In file included from /Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib/QtCore.framework/Headers/qnamespace.h:45:
/Users/Adam/spack/opt/spack/darwin-catalina-ivybridge/apple-clang-12.0.0/qt-4.8.7-lvspaubbgazl6kzhy5s4lcl6adwhmrvi/lib/QtCore.framework/Headers/qglobal.h:68:10: fatal error: 'algorithm' file not found
#include <algorithm>
         ^~~~~~~~~~~
1 error generated.
make: *** [qtdetail.o] Error 1
```
This feels like something specific to Apple Clang. Could someone test with GCC on Linux?

EDIT: I believe https://github.com/spack/spack/pull/32696/commits/0c3167bbd79c57784bfee176c9b6cedccba57b65 should fix this, we weren't patching one of the files and it was trying to use an ancient version of macOS.

var/spack/repos/builtin/packages/qt/qt4-mac.patch
```diff
@@ -215,9 +214,23 @@ diff -r -u qt-everywhere-opensource-src-4.8.7.orig/mkspecs/common/g++-base.conf
...
+--- a/mkspecs/common/g++-macx.conf	2015-05-07 09:14:42.000000000 -0500
++++ b/mkspecs/common/g++-macx.conf	2022-12-22 17:01:56.000000000 -0600
+@@ -16,8 +16,8 @@
+ 
+ QMAKE_LFLAGS_STATIC_LIB += -all_load
+ 
+-QMAKE_CFLAGS_X86_64 += -Xarch_x86_64 -mmacosx-version-min=10.5
+-QMAKE_CFLAGS_PPC_64 += -Xarch_ppc64 -mmacosx-version-min=10.5
++QMAKE_CFLAGS_X86_64 += -Xarch_x86_64 -mmacosx-version-min=$$QMAKE_MACOSX_DEPLOYMENT_TARGET
++QMAKE_CFLAGS_PPC_64 += -Xarch_ppc64 -mmacosx-version-min=$$QMAKE_MACOSX_DEPLOYMENT_TARGET
```
adamjstewart: This is the only thing in this file that changed. Previously we forgot to patch one of the config files. It just so happens that this config file is used when building pyqt4.

adamjstewart:
It's fun to rediscover that all of the hacks we were using are actually necessary

adamjstewart:
@Sinan81 I'm a bit stuck on qscintilla/qgis and could use your help.

I tried ripping out `qscintilla+python` and adding a separate `py-qscintilla` package but it didn't know where to find the sip files for PyQt5.

It seems like the sip build system has undergone huge changes since we first did this. None of the files/directories we used to use exist anymore. There is no `configure.py` anymore, you have to run `sip-build` instead. But this doesn't provide a way to specify where sip files live.

I'm inclined to deprecate all older versions and only support the latest versions, but I can't even get that working.

adamjstewart:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/qscintilla2.rb may provide some hints.

Sinan81:
I'll look into qscintilla this week. Thanks Adam!

Sinan81:
>There is no configure.py anymore, you have to run sip-build instead. But this doesn't provide a way to specify where sip files live.

following the example in homebrew formula, in qscintilla we should do something like:
```python
            ...
            with working_dir(join_path(self.stage.source_path, "Python")):
                mv = which('mv')
                mv('pyproject-qt6.toml', 'pyproject.toml')
                sip_inc_dir = join_path(self.spec[pyqtx].prefix, python_platlib, pyqtx, 'bindings' )
                tomlfile = FileFilter('pyproject.toml')
                tomlfile.filter('[tool.sip.metadata]','[tool.sip.project]\nsip-include-dirs = ["'+ sip_inc_dir + '"]\n[tool.sip.metadata]')
                pydir = join_path(python_platlib, pyqtx)
                mkdirp(os.path.join(self.prefix.share.sip, pyqtx))
                sip_install = Executable(spec["py-sip"].prefix.bin.join("sip-install"))
                sip_install(
                    #"configure.py",
                    #"--pyqt=" + pyqtx,
                    #"--sip=" + self.spec["py-sip"].prefix.bin.sip,
                    "--target-dir=" + self.spec.prefix,
                    "--qsci-include-dir=" + self.spec.prefix.include,
                    "--qsci-library-dir=" + self.spec.prefix.lib,
                    #"--qsci-sipdir=" + os.path.join(self.prefix.share.sip, pyqtx),
                    "--api-dir=" + self.prefix.share.qsci,
                    #"--destdir=" + pydir,
                    #"--pyqt-sipdir=" + os.path.join(self.spec[py_pyqtx].prefix.share.sip, pyqtx),
                    #"--sip-incdir="
                    #+ sip_inc_dir,
                    #"--stubsdir=" + pydir,
                )
                ...
```
I wasn't able to try it right away due to build system complications. I'll try building it manually first.
Update: manual build seems to be working.

adamjstewart:
@Sinan81 did you ever get another chance to hack on this? I'm moving back to a new laptop with new Xcode which means I'll no longer be able to install Qt or any Qt package, so I can no longer work on this. I might be able to extract more fixes from this PR and merge them separately, but I never managed to get newer versions of these packages to build. I'm probably going to close this PR soon.

Sinan81:
I'll take another look into this. Last time I checked, it seemed there was a complication with qmake build_system of Spack. I would suggest to keep this PR open a bit longer.

Sinan81:
`qscintilla+designer` builds just fine with the above mentioned build options. Next will sort out building it with `+python`

Sinan81:
made good progress, but there's an ABI mixup somewhere as shown below. will continue debugging.
```
==> [2023-03-13-23:46:52.444045] '/usr/bin/cp' 'pyproject-qt6.toml' 'pyproject.toml'
==> [2023-03-13-23:46:52.500307] '/mnt/local/sbulut/spack_develop/opt/spack/linux-centos7-haswell/gcc-8.5.0/py-sip-6.7.5-epai5j6skgsavetjp6v6mlgazhy4rifo/bin/sip-install' '--target-dir=/mnt/local/sbulut/spack_develop/opt/spack/linux-cento
s7-haswell/gcc-8.5.0/qscintilla-2.13.3-pedaevyiqyu7c7pao3yumhejf74cjr22' '--qsci-include-dir=/mnt/local/sbulut/spack_develop/opt/spack/linux-centos7-haswell/gcc-8.5.0/qscintilla-2.13.3-pedaevyiqyu7c7pao3yumhejf74cjr22/include' '--qsci-lib
rary-dir=/mnt/local/sbulut/spack_develop/opt/spack/linux-centos7-haswell/gcc-8.5.0/qscintilla-2.13.3-pedaevyiqyu7c7pao3yumhejf74cjr22/lib' '--api-dir=/mnt/local/sbulut/spack_develop/opt/spack/linux-centos7-haswell/gcc-8.5.0/qscintilla-2.1
3.3-pedaevyiqyu7c7pao3yumhejf74cjr22/share/qsci'
Querying qmake about your Qt installation...
Checking to see if the Qsci bindings can be built...
These bindings will be built: Qsci.
Generating the Qsci bindings...
sip-install: /mnt/local/sbulut/spack_develop/opt/spack/linux-centos7-haswell/gcc-8.5.0/py-pyqt6-6.4.0-rjwz4pmrpzsazoipkhzwbmikrrvu45or/lib/python3.8/site-packages/PyQt6/bindings/QtCore/QtCore.toml: 'QtCore' was built against ABI v13 but t
his module is being built against ABI v12
```

Sinan81:
speaking of 'mix-up's, currently qmake build_system depends on qt: should that be replaced with `qt-base` instead. As a result of this in `spack spec qscintilla` I see `qt@5` as dependency as well as `qt-base@6`

Sinan81:
when I use `qt-base` in qmake build system, spack install fails right away with error:
```
==> qscintilla: Executing phase: 'qmake'
==> Error: AttributeError: module 'spack.pkg.builtin.qscintilla' has no attribute 'qmake'

The 'qscintilla' package cannot find an attribute while trying to build from sources. This might be due to a change in Spack's package format to support multiple build-systems for a single package. You can fix this by updating the build recipe, and you can also report the issue as a bug. More information at https://spack.readthedocs.io/en/latest/packaging_guide.html#installation-procedure

/mnt/local/sbulut/spack_develop/lib/spack/spack/builder.py:144, in __forward:
  >>    144        def __forward(self, *args, **kwargs):
        145            return self.__getattr__(fn_name)(*args, **kwargs)

See build log for details:
  /cache/sbulut/spack-stage/spack-stage-qscintilla-2.13.3-xxaf2ennouvfk6px3wgcpxcsnyjzl2fs/spack-build-out.txt
```

wdconinc:
Qt6 (i.e. `qt-base`) doesn't require `qmake` anymore. In spack, `qt-base` uses CMake and doesn't install `qmake`.

adamjstewart:
`qt-base` does indeed install `qmake`. You just need to copy the following from `qt` to `qt-base`:
```python
        module.qmake = Executable(self.spec.prefix.bin.qmake)                         
```

wdconinc:
> `qt-base` does indeed install `qmake`

Right, I see now. Do you want to modify `qt-base` as part of this PR or should I file a separate PR for that change?

adamjstewart:
Let me make a separate PR, this one is likely going to take much longer.

adamjstewart:
@Sinan81 @manuelakuhn I believe this PR is once again ready for review.

Note that testing does not work at the moment: #38793

Recent versions of pyqt5 and pyqt6 no longer use the old `configure.py` build script and instead use `sip-build`. As such, I haven't figured out a way to build `+qsci_api`. This PR likely breaks packages that depend on pyqt5, but those packages are likely already broken on develop anyway.

This PR was basically done 6 months ago but was held up due to trying to get every dependent package to build correctly. I'm not going to make the same mistake this time. If anyone can prove that something that works on develop no longer works on this branch I can try to fix it, but I'm not going to try to get everything to build properly because it's a never ending endeavor.

var/spack/repos/builtin/packages/qt-base/package.py
```diff
@@ -89,6 +89,8 @@ class QtBase(QtPackage):
     url = QtPackage.get_url(__qualname__)
     list_url = QtPackage.get_list_url(__qualname__)
 
+    version("6.5.1", sha256="fdde60cdc5c899ab7165f1c3f7b93bc727c2484c348f367d155604f5d901bfb6")
```
adamjstewart: I found that QtCore did not build properly when using py-pyqt6@6.5 and qt-base@6.4. It's unclear if an exact version match is required or not. I couldn't find much documentation for this.

wdconinc: FYI, see also https://github.com/spack/spack/pull/36705/ which also adds the newer versions for other qt6 packages.

adamjstewart: Ah, let me review that

manuelakuhn:
Unfortunately, when trying to build some of the packages I encountered some issues:

These packages install and test fine: 
- `py-pyqt-builder`
- `py-sip@6.7.9` and `py-sip@4.19.25`
- `py-pyqt5-sip`
- `py-pyqt6-sip`
- `qscintilla` (but `qscintilla+python` does not unfortunately, see below)

As you already said for the py-pyqt packages the tests don't work, other than that `py-pyqt4@4.12.3` and `py-pyqt5@5.15.9` install fine, but unfortunately the other `py-pyqt5` packages fail to build:
- py-pyqt5@5.13.1
```
==> py-pyqt5: Executing phase: 'configure'
==> [2023-07-09-23:12:53.786375] '$spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/py-sip-6.7.9-jtdzyhij4ngzsvhe53dny4w4vdz77gdh/bin/sip-build' '--verbose' '--target-dir' '$spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/py-pyqt5-5.13.1-anxt3l3kuyz2jhli3w4pxrmtlbjp2s5m/lib/python3.10/site-packages' '--confirm-license' '--no-make' '--qmake' '$spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/qt-5.15.10-uum2qg6ix5nvo6nbenx2lpzcjoppxivv/bin/qmake'
sip-build: 'confirm_license' option has an invalid tools 'b'
```
- py-pyqt5@5.13.0 and 5.12.3
```
==> py-pyqt5: Executing phase: 'configure'
==> [2023-07-09-23:14:51.408828] '$spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/py-sip-6.7.9-jtdzyhij4ngzsvhe53dny4w4vdz77gdh/bin/sip-build' '--verbose' '--target-dir' '$spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/py-pyqt5-5.13.0-tjtl7jy4ftngpxypzwkerpbc3ua4vrrq/lib/python3.10/site-packages' '--confirm-license' '--no-make' '--qmake' '$spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/qt-5.15.10-uum2qg6ix5nvo6nbenx2lpzcjoppxivv/bin/qmake'
usage: sip-build [-h] [-V] [--quiet] [--verbose] [--build-dir DIR]
                 [--scripts-dir DIR] [--target-dir DIR] [--api-dir DIR]
                 [--no-compile] [--no-version-info] [--concatenate N]
                 [--debug] [--disabled-feature TAG] [--no-docstrings]
                 [--pep484-pyi] [--protected-is-public]
                 [--no-protected-is-public] [--tracing]
sip-build: error: unrecognized arguments: --confirm-license --no-make --qmake $spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/qt-5.15.10-uum2qg6ix5nvo6nbenx2lpzcjoppxivv/bin/qmake
```

`py-pyqt6` does not install for me because `qt-base` does not build with the error:
```
ERROR: The OpenGL functionality tests failed! You might need to modify the OpenGL package search path by setting the OpenGL_DIR CMake variable to the OpenGL library's installation directory.
```
(same happens when I build `qt-base` on its own)

A workaround is to install `py-pyqt6 ^qt-base+opengl`, then it installs fine (only the tests fail, same as `py-pyqt4` and `5`)

As for `ggis`, I could not test it because it depend on `qscintilla+python` which does not build with the error:
```
Usage: python configure.py [options]

configure.py: error: '$spack/opt/spack/linux-fedora37-x86_64/gcc-12.3.1/py-sip-6.7.9-jtdzyhij4ngzsvhe53dny4w4vdz77gdh/bin/sip' is not an executable
```

adamjstewart:
> unfortunately the other py-pyqt5 packages fail to build

Deprecated, don't care. These older versions can't even be downloaded anymore other than from cache.

> qt-base does not build

Maybe @wdconinc @sethrj can help with this one.

> qscintilla+python which does not build

Maybe try an older version of sip? Also don't care about this one.

wdconinc:
> > qt-base does not build

> Maybe @wdconinc @sethrj can help with this one.

Probably best to file a separate issue on that with more info (system etc). I don't have a macos to test things on, but this is fedora37 with gcc-12.3 and there's no issue reported yet for that.

sethrj (approved):

adamjstewart:
> qscintilla+python which does not build

Okay, I confirmed that this builds on develop and doesn't build on my branch so I'm going to reluctantly dig into this and see if I can get it working. FWIW, `qgis` doesn't build for me on develop either, so I'll leave that one to someone else.

Sinan81:
> FWIW, qgis doesn't build for me on develop either, so I'll leave that one to someone else.

Is this due pyqt stuff, or something else? I'll take a look at this today or so.

adamjstewart:
Not sure. I would wait until after this PR is merged because everything is going to change. I started working on qscintilla but that might need the qmake vdep PR if I want to test it on my laptop. Might just stick to Qt 5 so this PR isn't held up though.

Sinan81:
in develop branch py-pyqt5@5.13.1 fails to build due to a sip related issue. `qgis@3.28.3 ^py-pyqt5@5.13.0` builds about 70% giving a sip/python related error. I'll wait until after this PR is merged. (P.S: qgis@3.28.3 builds fine in spack@0.17.1 with py-pyqt5@5.13.1 and py-sip@4.19.21)

adamjstewart:
I'm probably not going to be able to get back to this for at least another week or two. If someone wants to hack on QScintilla (or just merge this and hack on that later) then be my guest.

Sinan81:
Made good progress on Qscintilla by combining a hint from linuxbrew recipe with RPM spec file for Fedora:
https://src.fedoraproject.org/rpms/qscintilla/blob/rawhide/f/qscintilla.spec
One weird issue I'm having with is that:
```
self.spec['py-pyqt6'].prefix
```
statement returns prefix for qscintilla package itself as opposed to py-pyqt6 package. Apart from this, the setup works fine, `qscintilla+python` builds with qt6 & py-pyqt6.
I'll create a branch for this soon and share here.

Sinan81:
@adamjstewart please see this branch that's based on your pyqt6 branch:
https://github.com/Sinan81/spack/tree/ajs_pyqt6_build_qscintilla
modified files are:
qscintilla package file and `lib/spack/spack/build_systems/qmake.py`

adamjstewart:
Thanks, I'll take a look once things settle down for me. Moving apartments and graduating and such.

adamjstewart:
@Sinan81 your latest commit fails with:
```
./SciAccessibility.h:31:10: fatal error: 'QAccessibleWidget' file not found
```

Sinan81:
>@Sinan81 your latest commit fails with:

I'll look into it this by end of today.. Could this be due to `self.spec['py-pyqt6...` issue I discussed above. Have you tried hard coding `sip_inc_dir`? Which version of qscintilla did  you try to install? I was using version 2.13.3 where I reported the successful build.

adamjstewart:
I just tried the default version. I'm busy for the rest of the week but I could maybe hack on this next week. Otherwise, maybe just merge this and fix qscintilla later.

Sinan81:
Yes. I think we can simply merge this. I am sure we can sort out qscintilla thing by end of next week at the latest.

Sinan81:
Please merge!

PR #32698, opened 2022-09-17
edm4hep: allow variant cxxstd 20, sync with podio

wdconinc:
Now that `podio` can support `cxxstd` variants 17 and 20, we can allow `edm4hep` to use `cxxstd=20` as well, but must ensure that `edm4hep` uses the same `cxxstd` variant as `podio`. Solution as in [`celeritas` ](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/celeritas/package.py#L23).


vvolkl (approved):

tldahlgren (approved):
LGTM.

PR #32701, opened 2022-09-17
arrow: add version 9 and more variants

aphecetche:


var/spack/repos/builtin/packages/arrow/package.py
```diff
@@ -15,6 +15,7 @@ class Arrow(CMakePackage, CudaPackage):
     homepage = "https://arrow.apache.org"
     url = "https://github.com/apache/arrow/archive/apache-arrow-0.9.0.tar.gz"
 
+    version("9.0.0", sha256="bb187b4b0af8dcc027fffed3700a7b891c9f76c9b63ad8925b4afb8257a2bb1b")
```
tldahlgren, line 18: Confirmed sha256.

var/spack/repos/builtin/packages/utf8proc/package.py
```diff
@@ -13,9 +13,17 @@ class Utf8proc(CMakePackage):
     homepage = "https://juliastrings.github.io/utf8proc/"
     url = "https://github.com/JuliaStrings/utf8proc/archive/v2.4.0.tar.gz"
 
+    version("2.7.0", sha256="4bb121e297293c0fd55f08f83afab6d35d48f0af4ecc07523ad8ec99aa2b12a1")
```
tldahlgren, line 16: Confirmed sha256

var/spack/repos/builtin/packages/xsimd/package.py
```diff
@@ -16,6 +16,7 @@ class Xsimd(CMakePackage):
     maintainers = ["ax3l"]
 
     version("develop", branch="master")
+    version("8.1.0", sha256="d52551360d37709675237d2a0418e28f70995b5b7cdad7c674626bcfbbf48328")
```
tldahlgren, line 19: Confirmed sha256.

var/spack/repos/builtin/packages/arrow/package.py
```diff
@@ -63,37 +96,49 @@ def patch(self):
             r"(include_directories\()SYSTEM ", r"\1", "cpp/cmake_modules/ThirdpartyToolchain.cmake"
         )
 
-    def cmake_args(self):
-        args = [
-            "-DARROW_USE_SSE=ON",
-            "-DARROW_BUILD_SHARED=ON",
-            "-DARROW_BUILD_STATIC=OFF",
-            "-DARROW_BUILD_TESTS=OFF",
```
tldahlgren, line 71: Did you intend to drop the `ARROW_BUILD_TESTS` definition?

var/spack/repos/builtin/packages/arrow/package.py
```diff
@@ -63,37 +96,49 @@ def patch(self):
...
+        args.append(self.define_from_variant("ARROW_TENSORFLOW", "tensorflow"))
+        args.append(self.define_from_variant("ARROW_WITH_BROTLI", "brotli"))
+        args.append(self.define_from_variant("ARROW_WITH_LZ4", "lz4"))
+        args.append(self.define_from_variant("ARROW_WITH_SNAPPY", "snappy"))
+        args.append(self.define_from_variant("ARROW_WITH_ZLIB", "zlib"))
+        args.append(self.define_from_variant("ARROW_WITH_ZSTD", "zstd"))
 
-        for dep in ("flatbuffers", "rapidjson", "snappy", "zlib", "zstd"):
-            args.append("-D{0}_HOME={1}".format(dep.upper(), self.spec[dep].prefix))
-        args.append("-DZLIB_LIBRARIES={0}".format(self.spec["zlib"].libs))
```
tldahlgren, line 98: Are these path arguments no longer needed?

tldahlgren (commented):
Confirmed all new sha256.  Have questions about some of the changes.

var/spack/repos/builtin/packages/arrow/package.py
aphecetche: honestly I had to x-check :wink: , but as far as I can tell `ARROW_BUILD_TESTS=OFF` is the default : https://github.com/apache/arrow/blob/40ec95646962cccdcd62032c80e8506d4c275bc6/cpp/cmake_modules/DefineOptions.cmake#L171

var/spack/repos/builtin/packages/arrow/package.py
aphecetche: in the stack I'm using arrow in, it works without, but indeed in that stack I don't use e.g. snappy or zstd. 
I will check...

aphecetche: ok, `spack install arrow +zstd+zlib+snappy` works but `spack install arrow@8 +zstd+zlib+snappy` complains about not finding Snappy indeed... 
Will put those lines back.

aphecetche: well, actually I'm not so sure `snappy` was working at all in `arrow@8` with the previous recipe. Seems snappy was a declared as dependency but in arrow's CMakeLists it was not requested, unless explicitly added with `ARROW_WITH_SNAPPY` ...

aphecetche: @tldahlgren instead of checking that the new variants I'm introducing works with previous versions, would it be acceptable if I only add them for "@9:" ?

And in the same spirit get the 

```
for dep in ("flatbuffers", "rapidjson", "snappy", "zlib", "zstd"):
            args.append("-D{0}_HOME={1}".format(dep.upper(), self.spec[dep].prefix))
        args.append("-DZLIB_LIBRARIES={0}".format(self.spec["zlib"].libs))
```

lines protected by a `with when("@:8"):` ?

tldahlgren: Yes, you definitely want to add version constraints to ensure constraints are only associated with the package when they are known to apply.

aphecetche: ok, I've added back the `args.append(...` lines for version < 9. 
I've kept the addition of the variants untouched though, as I believe their default values are "as before".

aphecetche:
@tldahlgren @ax3l is there something special that prevents this one from being merged ? thanks.

tldahlgren (approved):
LGTM

var/spack/repos/builtin/packages/arrow/package.py
```diff
@@ -28,32 +29,67 @@ class Arrow(CMakePackage, CudaPackage):
...
     depends_on("py-numpy", when="+python")
+    depends_on("python", when="+python")
     depends_on("rapidjson")
-    depends_on("snappy~shared")
-    depends_on("zlib+pic")
-    depends_on("zstd")
+    depends_on("re2+shared", when="+compute")
+    depends_on("re2+shared", when="+gandiva")
+    depends_on("snappy~shared", when="+snappy @9:")
+    depends_on("snappy~shared", when="@8:")
```
haampie, line 51: This doesn't look right... Would you be able to fix those conditions? I guess you meant `@:8`, but tbh, I cannot find how you got to 8 / 9, looks like it always had a corresponding CMake variable?

aphecetche: @haampie will try to have a look, can't say when though...

PR #32708, opened 2022-09-19
Fix benchmark recipe after branch name was changed

iarspider:


tldahlgren (approved):
Confirmed branch name change.

PR #32723, opened 2022-09-20
py-python-gitlab: add 3.9.0

manuelakuhn:
https://github.com/python-gitlab/python-gitlab/tree/v3.9.0

I was not sure about the version needed for `py-requests`. In [setup.py](https://github.com/python-gitlab/python-gitlab/blob/v3.9.0/setup.py#L31) they request version 2.25.0 but in [requirements.txt](https://github.com/python-gitlab/python-gitlab/blob/v3.9.0/requirements.txt#L1) they want version 2.28.1. I now used the one in `setup.py` because this is the one used in py-python-gitlab@2.10.1 (there different versions were already requested in `setup.py` and `requirements.txt` as well)


adamjstewart (approved):
`setup.py` is the right place to look, `requirements.txt` isn't actually used or checked by pip.

manuelakuhn:
@spackbot run pipeline

PR #32728, opened 2022-09-20
py-eccodes: optionally hardcode path to libeccodes

skosukhin:
It is rather annoying that `py-eccodes` cannot find its runtime dependency without additional environment variables. This introduces variant `hardcode-path`. When the variant is disabled (default), the package depends on `py-findlibs`, which finds `libeccodes` at the runtime with the help of environment variables `ECCODES_HOME` and `ECMWFLIBS_ECCODES` (instead of much more general `LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH`, which are used currently). If the variant `hardcode-path` is enabled, the dependency on `py-findlibs` is dropped and the path to `libeccodes` is hardcoded into `py-eccodes` with a patch.

This also adds the most recent version of the package.

### DIFF
--- a/var/spack/repos/builtin/packages/py-eccodes/hardcode-path.patch
+++ b/var/spack/repos/builtin/packages/py-eccodes/hardcode-path.patch
@@ -1,0 +1,18 @@
+--- a/gribapi/bindings.py
++++ b/gribapi/bindings.py
+@@ -25,14 +25,7 @@ __version__ = "1.5.0"
+ 
+ LOG = logging.getLogger(__name__)
+ 
+-try:
+-    import ecmwflibs as findlibs
+-except ImportError:
+-    import findlibs
+-
+-library_path = findlibs.find("eccodes")
+-if library_path is None:
+-    raise RuntimeError("Cannot find the ecCodes library")
++library_path = "@ECCODES_PATH@"
+ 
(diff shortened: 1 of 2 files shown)


adamjstewart:
Not sure how I feel about this approach. Patching the source code with something that can't be upstreamed is frowned upon, but so is relying on environment variables. I'm inclined to choose one or the other instead of having a variant. How do we handle this situation in other packages? I think I'm relying on env vars in the other geospatial libs I maintain, but I'm happy to change that.

skosukhin:
@adamjstewart do you think it's possible to come up with a patch for `setup.py` (which we could submit upstream) to make it less hacky? Something like a configure-time option?

adamjstewart:
I tried that once with shapely I think? Basically tried to convince them to use something similar to numpy where the path to the libs is hardcoded in a setup.cfg or site.cfg file that gets installed along with the package. They weren't interested. But eccodes developers may be interested. I would first explain the problem to them and ask what solutions they would be interested in before spending too much time hacking anything together.

bernhardkaindl:
@skosukhin Do you still want to keep this PR like you wrote in #32792?

skosukhin:
Some day I might come up with a PR upstream. So, I'm closing this.

PR #32736, opened 2022-09-21
Include exception info related to url retrieval in debug messages

scheibelp:
... which otherwise would be swallowed

I was speaking to a user who got a series of fetch errors without any elaboration when their CA configuration was not sufficient to retrieve an HTTPS link. I think some output for these exceptions would have been useful to identify that.


alalazo (approved):
LGTM. I am wondering if these options are better suited for debug `level=2`. Feel free to merge if you think they are not.

PR #32737, opened 2022-09-21
`spack diff` any specs you want

### LINKED ISSUE #31782: Ability to diff specs from files
### Summary

In spack v0.18.1 (and not only), there is `spack diff` however that is limited in its functionality. It would be very nice to be able to diff arbitrary specs from files

### Rationale

One could be running spack on two different machines and experiencing different behaviors. The administrator could save the spec from one machine into a file, bring it to the other machine and run `spack diff` on that file to compare it with the local one (which might be exported to a file too -- the mechanics are unimportant).

Similarly, one could more easily diff specs from separate environments on the same spack installation.

### Description of an alternative

Manually or semi-manually running various `diff` tools on the output of `spack spec` and/or on the content of the lock files for environments

### Additional information

@scheibelp this is the followup as it was discussed at today's concall 

### General information

scheibelp:
Fixes #31782

With this change, if a spec is concrete after parsing (e.g. spec.yaml or /hash-based), then it is not disambiguated (a process which requires (a) that the spec be installed and (b) that it be part of the currently active environment).

This means that you can:

* Diff specs from an environment regardless of whether they have been installed (more useful for projection/matrix-based envs)
* Diff specs read from `.yaml` files which may or may not be entirely different installations


lib/spack/spack/cmd/diff.py
```diff
@@ -198,10 +198,12 @@ def diff(parser, args):
...
-    specs = [
-        spack.cmd.disambiguate_spec(spec, env, first=args.load_first)
-        for spec in spack.cmd.parse_specs(args.specs)
-    ]
+    specs = []
+    for spec in spack.cmd.parse_specs(args.specs):
+        if spec.concrete:
+            specs.append(spec)
+        else:
+            specs.append(spack.cmd.disambiguiate_spec(spec, env, first=args.load_first))
```
alalazo:
```suggestion
            specs.append(spack.cmd.disambiguate_spec(spec, env, first=args.load_first))
```

scheibelp:
This is running into unrelated CI errors described by https://github.com/spack/spack/issues/32470

scheibelp:
Close/reopen to restart unit tests

becker33 (approved):

scheibelp:
@spackbot run pipeline

PR #32738, opened 2022-09-21
Ci generate fake install

kwryankrattiger:


scottwittenburg (commented):
Thanks @kwryankrattiger, I'll be interested to see some pipelines run with this!  If you can push this to a side repo on gitlab.spack.io so we can see it in action, that would be awesome.

Also, maybe a nitpick, but I feel referring to it as "fake packages" here is confusing, since we're still concretizing the real stacks/packages.  The difference is we're just doing fake installations.  So personally I prefer referring to it everywhere as "fake installs" rather than "fake packages".

alalazo:
Is this PR still relevant, or can we close it?

kwryankrattiger:
It is still relevant, it just hasn't been a priority.

The goal of this PR is to generate "real" pipelines using synthetic packages for testing the CI module features that are not necessarily testable in unit tests.

There as been a quite a bit of change since this was originally proposed, so that patch likely does not apply cleanly at all and some components will likely require re-write. I would like to leave it open though as it is my hope to come back and merge this when I have time to really test it in a staging environment.

PR #32745, opened 2022-09-21
add full_optimizations flags to force release configs for exago package

ryandanehy:
Updating exago's package.py to allow debuging and releasing config options.

PR #32750, opened 2022-09-21
Optimize flags and Release/Debug flags

ryandanehy:
Added optimize and release/debug flags for the exago package.

### DIFF
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -77,12 +77,12 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     # Profiling and optimizations
     depends_on("hiop+deepchecking build_type=RelWithDebInfo", when="+hiop+debug")
     depends_on("hiop~deepchecking  build_type=Release ", when="+hiop+full_optimizations")
-    
+
     # Control the package's build-type depending on the full_optimizations or debug flag
     for pkg in ["raja", "umpire", "magma", "camp"]:
         depends_on("{0} build_type=Release".format(pkg), when="+full_optimizations")
-        depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="+debug")	
-    
+        depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="+debug")
+
     # Force RelWithDebInfo when not using optimizations
     conflicts(
         "build_type=Release",
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -46,7 +46,11 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     conflicts("~hiop~ipopt", msg="ExaGO needs at least one solver enabled")
 
-    # Dependencides
+    # Profling and optimization flags
+    variant("full_optimizations", default=False, description="Enable/Disable optimizations and release type")
+    #variant("with_profiling", default="none", description="Specify profiling tools", multi=True, values=("none", "hpctoolkit", "tau"))
+
+    # Dependencies
     depends_on("pkgconfig", type="build")
     depends_on("mpi", when="+mpi")
     depends_on("blas")
@@ -61,6 +65,24 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("umpire+cuda~shared", when="+raja+cuda")
 
     depends_on("cmake@3.18:", type="build")
+
+    # Profiling and optimizations
+    depends_on("hiop+deepchecking~full_optimizations", when="+hiop~full_optimizations")
+    depends_on("hiop~deepchecking+full_optimizations", when="+hiop+full_optimizations")
+
+    # Force RelWithDebInfo when not using optimizations 
+    conflicts("build_type=Release", when="~full_optimizations", msg="Use RelWithDebInfo when not using optimizations")
+    conflicts("build_type=Debug", when="~full_optimizations", msg="Use RelWithDebInfo when not using optimizations")
+    conflicts("build_type=MinSizeRel", when="~full_optimizations", msg="Use RelWithDebInfo when not using optimizations")
+
+    # Force Release mode when using optimizations
+    conflicts("build_type=RelWithDebInfo", when="+full_optimizations", msg="Use Release when using optimizations")
+    conflicts("build_type=Debug", when="+full_optimizations", msg="Use Release when using optimizations")
+    conflicts("build_type=MinSizeRel", when="+full_optimizations", msg="Use Release when using optimizations")
+
+    #depends_on("hpctoolkit", when="with_profiling=hpctoolkit")
+    #depends_on("tau", when="with_profiling=tau")
+    #^ need to depend when both hpctoolkit and tau
 
     # HiOp dependency logic
     depends_on("hiop+raja", when="+hiop+raja")
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -46,14 +46,17 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     conflicts("~hiop~ipopt", msg="ExaGO needs at least one solver enabled")
 
-    # Profling and optimization flags
+    # optimization flag
     variant(
         "full_optimizations",
         default=False,
         description="Enable/Disable optimizations and release type",
     )
-    # variant("with_profiling", default="none",
-    # description="Specify profiling tools", multi=True, values=("none", "hpctoolkit", "tau"))
+    variant(
+        "debug",
+        default=False,
+        description="Enable RelWithDebInfo",
+    )
 
     # Dependencies
     depends_on("pkgconfig", type="build")
@@ -72,23 +75,23 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("cmake@3.18:", type="build")
 
     # Profiling and optimizations
-    depends_on("hiop+deepchecking~full_optimizations", when="+hiop~full_optimizations")
+    depends_on("hiop+deepchecking+debug", when="+hiop+debug")
     depends_on("hiop~deepchecking+full_optimizations", when="+hiop+full_optimizations")
 
     # Force RelWithDebInfo when not using optimizations
     conflicts(
         "build_type=Release",
-        when="~full_optimizations",
+        when="+debug",
         msg="Use RelWithDebInfo when not using optimizations",
     )
     conflicts(
         "build_type=Debug",
-        when="~full_optimizations",
+        when="+debug",
         msg="Use RelWithDebInfo when not using optimizations",
     )
     conflicts(
         "build_type=MinSizeRel",
-        when="~full_optimizations",
+        when="+debug",
         msg="Use RelWithDebInfo when not using optimizations",
     )
 
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -69,10 +69,29 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("hiop~deepchecking  build_type=Release ", when="+hiop build_type=Release ")
 
     # Control the package's build-type depending on the release or debug flag
-    for pkg in ["raja", "umpire", "magma", "camp", "ginkgo"]:
-        depends_on("{0} build_type=Release".format(pkg), when="+{0} build_type=Release".format(pkg))
-        depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="+{0} build_type=RelWithDebInfo".format(pkg))
+    for pkg in [
+        ("raja", "raja"),
+        ("umpire", "raja"),
+        ("magma", "hiop+cuda"),
+        ("magma", "hiop+rocm"),
+        ("camp", "raja"),
+    ]:
+        depends_on(
+            "{0} build_type=Release".format(pkg[0]), when="+{0} build_type=Release".format(pkg[1])
+        )
+        depends_on(
+            "{0} build_type=RelWithDebInfo".format(pkg[0]),
+            when="+{0} build_type=RelWithDebInfo".format(pkg[1]),
+        )
 
+    depends_on(
+        "{0} build_type=Release".format("hiop+ginkgo ^ginkgo"),
+        when="+{0} build_type=Release".format("hiop ^hiop+ginkgo"),
+    )
+    depends_on(
+        "{0} build_type=Debug".format("hiop+ginkgo ^ginkgo"),
+        when="+{0} build_type=RelWithDebInfo".format("hiop ^hiop+ginkgo"),
+    )
     # depends_on("hpctoolkit", when="with_profiling=hpctoolkit")
     # depends_on("tau", when="with_profiling=tau")
     # ^ need to depend when both hpctoolkit and tau
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -73,40 +73,6 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("{0} build_type=Release".format(pkg), when="build_type=Release")
         depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="build_type=RelWithDebInfo")
 
-    # Force RelWithDebInfo when not using release
-    conflicts(
-        "build_type=Release",
-        when="build_type=RelWithDebInfo",
-        msg="Use RelWithDebInfo when not using release build type",
-    )
-    conflicts(
-        "build_type=Debug",
-        when="build_type=RelWithDebInfo",
-        msg="Use RelWithDebInfo when not using release build type",
-    )
-    conflicts(
-        "build_type=MinSizeRel",
-        when="build_type=RelWithDebInfo",
-        msg="Use RelWithDebInfo when not using release build type",
-    )
-
-    # Force Release mode when using optimizations
-    conflicts(
-        "build_type=RelWithDebInfo",
-        when="build_type=Release",
-        msg="Use Release when using release build type",
-    )
-    conflicts(
-        "build_type=Debug",
-        when="build_type=Release",
-        msg="Use Release when using release build type",
-    )
-    conflicts(
-        "build_type=MinSizeRel",
-        when="build_type=Release",
-        msg="Use Release when using release build type",
-    )
-
     # depends_on("hpctoolkit", when="with_profiling=hpctoolkit")
     # depends_on("tau", when="with_profiling=tau")
     # ^ need to depend when both hpctoolkit and tau
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -70,8 +70,8 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     # Control the package's build-type depending on the release or debug flag
     for pkg in ["raja", "umpire", "magma", "camp", "ginkgo"]:
-        depends_on("{0} build_type=Release".format(pkg), when="build_type=Release")
-        depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="build_type=RelWithDebInfo")
+        depends_on("{0} build_type=Release".format(pkg), when="+{0} build_type=Release".format(pkg))
+        depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="+{0} build_type=RelWithDebInfo".format(pkg))
 
     # depends_on("hpctoolkit", when="with_profiling=hpctoolkit")
     # depends_on("tau", when="with_profiling=tau")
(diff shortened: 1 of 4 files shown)


var/spack/repos/builtin/packages/ginkgo/package.py
```diff
@@ -44,6 +44,11 @@ class Ginkgo(CMakePackage, CudaPackage, ROCmPackage):
         values=("Debug", "Release"),
     )
 
+    # make sure Debug mode is not used when optimizing
+    conflicts(
+        "build_type=Debug", when="+full_optimizations", msg="Use Release mode when optimizing"
+    )
+
```
tcojean: I have the same comments as the previous PR, see https://github.com/spack/spack/pull/32745#pullrequestreview-1115781215

And related comments, so that Ginkgo simply is used with the `build_type` system.

tcojean (changes requested):

var/spack/repos/builtin/packages/ginkgo/package.py
cameronrutherford: +1 - I don't think this PR needs to touch Ginkgo package

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -63,6 +63,52 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
...
+        "build_type=MinSizeRel",
+        when="+full_optimizations",
+        msg="Use Release mode when with optimizations",
+    )
+    conflicts(
+        "build_type=RelWithDebInfo",
+        when="+full_optimizations",
+        msg="Use Release mode when with optimizations",
+    )
+    # force no optimizations flag to use RelWithDebInfo mode
```
cameronrutherford: Perhaps it is too restrictive to force `RelWithDebInfo` when building `~full_optimizations`. Perhaps we can just add conflicts for `Release` and `MinSizeRel` here, while forcing `Release` mode with `+full_optmizations`

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -46,7 +46,16 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
     conflicts("~hiop~ipopt", msg="ExaGO needs at least one solver enabled")
 
-    # Dependencides
+    # Profling and optimization flags
+    variant(
+        "full_optimizations",
+        default=False,
+        description="Enable/Disable optimizations and release type",
+    )
+    # variant("with_profiling", default="none",
```
cameronrutherford: I'm happy to leave in these comments as an indication that we are working on adding this sort of support, even though it isn't productional yet.

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -62,6 +71,46 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("cmake@3.18:", type="build")
 
+    # Profiling and optimizations
+    depends_on("hiop+deepchecking~full_optimizations", when="+hiop~full_optimizations")
+    depends_on("hiop~deepchecking+full_optimizations", when="+hiop+full_optimizations")
+
+    # Force RelWithDebInfo when not using optimizations
```
cameronrutherford: Same comments as for HiOp in forcing a build type - not sure I have seen other packages experiment with this, so I can't really reference any best practices.

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -63,6 +63,52 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
...
+        msg="Use RelWithDebInfo mode when without optimizations",
+    )
+    conflicts(
+        "build_type=MinSizeRel",
+        when="~full_optimizations",
+        msg="Use RelWithDebInfo mode when without optimizations",
+    )
+
+    # force dependencies build types for optimizations
+    for pkg in ["raja", "umpire", "magma", "camp"]:
```
cameronrutherford: I think we can add ginkgo here as @tcojean suggests, but we might have to add a little logic to also propagate the `full_optimizations` flag to ginkgo.

cameronrutherford (changes requested):
Much cleaner than the previous PR. Thanks, @ryandanehy!

var/spack/repos/builtin/packages/hiop/package.py
tcojean: Yes adding Ginkgo here should be enough. Also, as I stated in https://github.com/spack/spack/pull/32745#pullrequestreview-1115781215 it is not necessary to use the `full_optimizations` flag of Ginkgo for HiOp, this specifically targets some optimizations for some preconditioners which you are not using, but increases the compile time significantly.

cameronrutherford: Ah, I didn't quite catch that. Thank you. Should be sufficient just to specify the build type then, and we can leave `full_optimizations` alone to speed up our Ginkgo compile time!

pelesh (changes requested):
I believe this PR is not using `full_optimization` option as it was intended (see comments by @tcojean). My understanding of the intent of this PR is to provide options for `Release` and `RelWithDebInfo` build types in order to simplify building HiOp for performance and CI testing, respectively. I don't think we need to use `full_optimization` for that.

pelesh (changes requested):
Remove `full_optimization` variant; otherwise it is ready to merge.

cameronrutherford:
> Remove `full_optimization` variant; otherwise it is ready to merge.

Are you referring to the `full_optimisations` variant in HiOp? I think we still want to keep that in ExaGO

pelesh:
> Are you referring to the `full_optimisations` variant in HiOp? I think we still want to keep that in ExaGO

Yes, I am referring to `full_optimisations` variant in HiOp. It does not change anything for HiOp but adds unnecessary complexity in the spack package and increases the compilation time significantly (per @tcojean):

> Also, as I stated in #32745 it is not necessary to use the `full_optimizations` flag of Ginkgo for HiOp, this specifically targets some optimizations for some preconditioners which you are not using, but increases the compile time significantly.

HiOp is using direct solvers from Ginkgo. It is unlikely HiOp will use Ginkgo preconditioners anytime soon.

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -19,7 +19,7 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
 
     homepage = "https://github.com/LLNL/hiop"
     git = "https://github.com/LLNL/hiop.git"
-    maintainers = ["ashermancinelli", "CameronRutherford", "pelesh"]
+    maintainers = ["ryandanehy", "CameronRutherford", "pelesh"]
```
ashermancinelli, line 22: It's been a good run y'all 😭

cameronrutherford: It has been! @ryandanehy can you also make the same change to the ExaGO package?

cameronrutherford:
> > Are you referring to the `full_optimisations` variant in HiOp? I think we still want to keep that in ExaGO
> 
> Yes, I am referring to `full_optimisations` variant in HiOp. It does not change anything for HiOp but adds unnecessary complexity in the spack package and increases the compilation time significantly (per @tcojean):
> 
> > Also, as I stated in #32745 it is not necessary to use the `full_optimizations` flag of Ginkgo for HiOp, this specifically targets some optimizations for some preconditioners which you are not using, but increases the compile time significantly.
> 
> HiOp is using direct solvers from Ginkgo. It is unlikely HiOp will use Ginkgo preconditioners anytime soon.

Ok I agree - I think we can keep the optimization and debug variants at the ExaGO level, and have it control the build types of any dependencies.

ryandanehy:
I made alot of the changes that were suggested, and I feel it is at a good place now. @pelesh @CameronRutherford @tcojean Curious if they were any opinions or notes on my new commits/updated pr.

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -46,7 +46,19 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     conflicts("~hiop~ipopt", msg="ExaGO needs at least one solver enabled")
 
-    # Dependencides
+    # optimization flag
+    variant(
+        "full_optimizations",
+        default=False,
+        description="Enable/Disable optimizations and release type",
```
cameronrutherford: [nitpick] I think this description could be updated to include that this affects dependencies.

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -46,7 +46,19 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
+    )
+    variant(
+        "debug",
+        default=False,
+        description="Enable RelWithDebInfo",
```
cameronrutherford: [nitpick] This also changes HiOp, and the description should reflect that.

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -62,6 +74,51 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("cmake@3.18:", type="build")
 
+    # Profiling and optimizations
+    depends_on("hiop+deepchecking build_type=RelWithDebInfo", when="+hiop+debug")
+    depends_on("hiop~deepchecking  build_type=Release ", when="+hiop+full_optimizations")
```
cameronrutherford: [nitpick] I think we should include modification of HiOp build type with other dependencies below, and only configure `deepchecking` in these lines

cameronrutherford (dismissed):
Left mostly nitpicks, but I think this is the final functionality we want

pelesh (changes requested):
References to `full_optimization` need to be removed. Otherwise looks good.

tcojean:
> I made alot of the changes that were suggested, and I feel it is at a good place now. @pelesh @CameronRutherford @tcojean Curious if they were any opinions or notes on my new commits/updated pr.

Thanks from my end it looks

cameronrutherford:
> References to `full_optimization` need to be removed. Otherwise looks good.

I disagree. We have used `full_optimizations` as a way of defining an optimised build of ExaGO that we can use for profiling. There would be no question about how to configure the dependencies when using this option.

This would also allow other developers/collaborators to also build optimised deployments when they desire, and contributes to the reproducibility of our software stack.

I also don't see any harm in keeping the variant, as building with `~full_optimizations` has no effect on the package.

pelesh:
> I disagree. We have used `full_optimizations` as a way of defining an optimised build of ExaGO that we can use for profiling. There would be no question about how to configure the dependencies when using this option.

I suggest we rename `full_optimization` variant to `release`, because it corresponds to a `Release` CMake build and that way we also avoid confusion with `full_optimization` option in Ginkgo. But this is just me nitpicking :).

var/spack/repos/builtin/packages/exago/package.py
ryandanehy: I am a little vague on what you mean? Are you suggesting that deep checking should be toggled based on the appearance of the debug or full optimization flag without any notion of hiop?
For example: depends_on("+deepchecking build_type=RelWithDebInfo", when="+debug")
Then include hiop with the other packages such as raja, magma, etc.

pelesh: fwiw, "deepchecking" is a HiOp specific flag, which turns on a bunch of assert statements within HiOp code. You want to have deepchecking enabled when running CI tests/debugging, but not in a production code. Setting the deepchecking flag will not affect any library other than HiOp.

cameronrutherford:
> I suggest we rename `full_optimization` variant to `release`, because it corresponds to a `Release` CMake build and that way we also avoid confusion with `full_optimization` option in Ginkgo. But this is just me nitpicking :).

That's completely fair. I think a `release` and `debug` are much better names

cameronrutherford (dismissed):

cameronrutherford:
I think this is good to be merged once tests are passing. Thanks @ryandanehy!! 

cc @alalazo or @haampie (since you have both been kind enough to merge pervious PRs) if you are able to get this merged that would be great.

pelesh (dismissed):

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -46,7 +46,19 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
+    variant(
+        "release",
+        default=False,
+        description="Enable release type for packages exago, hiop, raja, magma, camp",
+    )
+    variant(
+        "debug",
+        default=False,
+        description="Enable RelWithDebInfo for packages exago, hiop, raja, magma, camp",
+    )
```
alalazo: Hmm, I kind of disagree with this, since `+release` is a proxy for `build_type=Release` and `+debug` is a proxy for `build_type=RelWithDebInfo`. Both options are already covered.

alalazo (changes requested):

var/spack/repos/builtin/packages/exago/package.py
cameronrutherford: We use this variant to also control the build type of dependency packages - are you suggesting we should change the name, or is this something we should do at all?

I'm satisfied if we should just handle this in our `spack.yaml` environment, but I think we have good intentions with the variant.

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -62,6 +74,49 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("cmake@3.18:", type="build")
 
+    # Profiling
+    depends_on("hiop+deepchecking build_type=RelWithDebInfo", when="+hiop+debug")
```
alalazo: @CameronRutherford Why do you need to introduce an alias variant? Can't you do:
```suggestion
    depends_on("hiop+deepchecking build_type=RelWithDebInfo", when="+hiop build_type=RelWithDebInfo")
```

cameronrutherford: I see what you mean. I thought that controlling dependencies based on variants was a better practice, but I realise now we can control things through the build type alone.

@ryandanehy do you want to take a stab at making this change?

ryandanehy: @CameronRutherford if I understand correctly then this means the variants and hence the flags are not needed at all. So I just need to control things using the build type of exago and hiop. Long story short remove the variants and have the packages build types controlled by Exago build type.

cameronrutherford: Yes. You've got it

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -62,6 +62,55 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
+    conflicts(
+        "build_type=Debug",
+        when="build_type=Release",
+        msg="Use Release when using release build type",
+    )
+    conflicts(
+        "build_type=MinSizeRel",
+        when="build_type=Release",
+        msg="Use Release when using release build type",
+    )
```
cameronrutherford: These conflicts mean nothing anymore, and should be removed. 

I guess we were significantly over complicating things initially, but I like the solution we've arrived at.

cameronrutherford (changes requested):

ryandanehy (commented):
I think the suggested and needed changes have been made. Sorry for the delayed response.

alalazo (dismissed):

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -62,6 +62,21 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
     depends_on("cmake@3.18:", type="build")
 
+    # Profiling
+    depends_on(
+        "hiop+deepchecking build_type=RelWithDebInfo", when="+hiop build_type=RelWithDebInfo"
+    )
+    depends_on("hiop~deepchecking  build_type=Release ", when="+hiop build_type=Release ")
+
+    # Control the package's build-type depending on the release or debug flag
+    for pkg in ["raja", "umpire", "magma", "camp"]:
```
cameronrutherford:
```suggestion
    for pkg in ["raja", "umpire", "magma", "camp", "gingko"]:
```

alalazo:
Ping me back for a merge at the end of the day. I'll see it tomorrow. Thanks!

var/spack/repos/builtin/packages/exago/package.py
cameronrutherford: Just one minor change, and then I think this is ready to merge. Thanks @ryandanehy!!!

ryandanehy:
> 

@alalazo a quick reminder.

cameronrutherford:
I don't understand why certain CI pipelines are failing for ginkgo when specifying the build type - seems like this should be good to merge!

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -62,6 +62,21 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
     depends_on("cmake@3.18:", type="build")
 
+    # Profiling
+    depends_on(
+        "hiop+deepchecking build_type=RelWithDebInfo", when="+hiop build_type=RelWithDebInfo"
+    )
+    depends_on("hiop~deepchecking  build_type=Release ", when="+hiop build_type=Release ")
+
+    # Control the package's build-type depending on the release or debug flag
+    for pkg in ["raja", "umpire", "magma", "camp", "gingko"]:
```
alalazo:
```suggestion
    for pkg in ["raja", "umpire", "magma", "camp", "ginkgo"]:
```

alalazo (changes requested):
I'm always proud when audits spot these typos :laughing:

var/spack/repos/builtin/packages/exago/package.py
cameronrutherford: Impressive that there is a check to catch this! I don't know much about the spack audit pipeline, but seems very useful!

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -62,6 +62,21 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
+    # Profiling
+    depends_on(
+        "hiop+deepchecking build_type=RelWithDebInfo", when="+hiop build_type=RelWithDebInfo"
+    )
+    depends_on("hiop~deepchecking  build_type=Release ", when="+hiop build_type=Release ")
+
+    # Control the package's build-type depending on the release or debug flag
+    for pkg in ["raja", "umpire", "magma", "camp", "ginkgo"]:
+        depends_on("{0} build_type=Release".format(pkg), when="build_type=Release")
+        depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="build_type=RelWithDebInfo")
```
cameronrutherford:
```suggestion
        depends_on("{0} build_type=Release".format(pkg), when="+{0} build_type=Release".format(pkg))
        depends_on("{0} build_type=RelWithDebInfo".format(pkg), when="+{0} build_type=RelWithDebInfo".format(pkg))
```

I'm not as adept at reading audit logs as @alalazo, but I think this is the correct logic, and a upon re-reading I think I found the bug

cameronrutherford (changes requested):
Possibly the final bug?

var/spack/repos/builtin/packages/exago/package.py
cameronrutherford: Turns out there isn't actually a variant that maps directly to each package... @ryandanehy can you create a tuple list that pairs each package to the variant combination that controls it, and then modify this loop accordingly?

ryandanehy: I believe it just ginkgo that doesn't have a relwithDebug build type. @CameronRutherford so I think I can just do a depends on statements for ginkgo separately rather than have a tuple list.

cameronrutherford: All packages have the same set of build types, so I am not sure why ginkgo is different - perhaps we need to hard code for that package. Style is failing, but additionally we see errors in the audit:

```
1. exago: wrong variant in "depends_on" directive
    the variant 'umpire' does not exist
    in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/exago/package.py
2. exago: wrong variant in "depends_on" directive
    the variant 'magma' does not exist
    in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/exago/package.py
3. exago: wrong variant in "depends_on" directive
    the variant 'camp' does not exist
    in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/exago/package.py
4. exago: wrong variant in "depends_on" directive
    the variant 'ginkgo' does not exist
    in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/exago/package.py
5. exago: wrong variant used for a dependency in a 'depends_on' directive
    invalid values for variant "build_type" in package "ginkgo": ['RelWithDebInfo'] in package 'ginkgo'
    in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/exago/package.py
```

While it looks like you are right about ginkgo in particular not having that build type, there are additional errors that reference failures in looking for variants `camp`, `magma`, `umpire` and `ginkgo`. These packages are not controlled by their namesake variant, hence my suggestion.

cameronrutherford (approved):

cameronrutherford:
CI is passing - I think this is finally good to merge

alalazo (approved):

PR #32764, opened 2022-09-22
E4S: Convert specs to use Data and Vis SDK

kwryankrattiger:


kwryankrattiger:
This can no longer be used by spack CI. The bits fixing other packages will be ported to their own PRs.

@eugeneswalker I am not sure if there is another place to host and test concretizing and installing minimized e4s more than once every three months when images are built.

PR #32766, opened 2022-09-22
bootstrap: don't do relocation logic on binary patchelf

haampie:
Currently Spack implements the relocation logic for patchelf itself in a curious way:

1. `cp patchelf patchelf.bak`
2. `patchelf.bak --set-rpath ... patchelf`
3. `rm patchelf.bak`

The fact that this works at all means that `patchelf` is relocatable software
to start with.

That simply means we should skip relocation of patchelf during bootstrap, to
break the recursion where patchelf is required to bootstrap patchelf.

This PR informs the installer to not relocate patchelf.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -1600,7 +1600,9 @@ def _extract_inner_tarball(spec, filename, extract_to, unsigned, remote_checksum
     return tarfile_path
 
 
-def extract_tarball(spec, download_result, allow_root=False, unsigned=False, force=False):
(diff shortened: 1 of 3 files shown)


haampie:
Hopefully this also solves an issue on the amazing not-entirely-posix filesystem called Cray DVS.

For me, running patchelf.bak on patchelf would modify and append to the executable, and produce a funny binary with a string table that's not mapped to memory (which gdb points out). In the best case, this `patchelf` executable would error about issues with `mprotect`, in the worst case it would immediately segfault.

PR #32768, opened 2022-09-22
Skip all tests using problematic fixture on python 2.7

scheibelp:
This is an extension of https://github.com/spack/spack/pull/32569 but for all tests using the problematic fixture (2 additional tests, 4 in total).

From one perspective, this could be said to "fix" https://github.com/spack/spack/issues/32470 (although I don't know the cause - why the fixture doesn't work with Python 2.7).

I ran into this error again as part of testing https://github.com/spack/spack/pull/32737.


scheibelp:
@spackbot run pipeline

alalazo (approved):

PR #32770, opened 2022-09-22
[WIP] "spack find -c": show concretized-but-not-installed specs

scheibelp:
If you run `spack find -c`, it will show you all the concretized roots.

However, if your env is not installed, you currently cannot use `spack find <query_spec>` to query environment specs (i.e. the query spec is not used to pull uninstalled-but-concretized specs from the env). This is primarily useful if you are trying to understand how matrix specifications are expanded in combination with `unify` directives (so for example your environment tries to build several instances of the same package with different `mpi`s and compilers).

I'm calling this WIP because `spack find` output currently assumes everything returned by the query has been installed:

```
==> 16 installed packages
```

and also because passing `show_concretized` is handled awkwardly (I want to make sure the semantics are acceptable and then I'll refactor).

### DIFF
--- a/lib/spack/spack/cmd/common/arguments.py
+++ b/lib/spack/spack/cmd/common/arguments.py
@@ -88,7 +88,7 @@ class ConstraintAction(argparse.Action):
                 # This is fast for already-concrete specs
                 specs[s.dag_hash()] = s
 
-        if kwargs.get('show_concretized', False):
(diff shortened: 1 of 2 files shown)


eugeneswalker:
This will be great. Thank you!

scheibelp:
Superseded by https://github.com/spack/spack/pull/44713

PR #32771, opened 2022-09-22
py-torch: fix +rocm+nccl build

adamjstewart:
Same as #32461 but for PyTorch.

@kolamsrinivas @srekolam @rsanthanam-amd @renjithravindrankannath @nicholas-sly


srekolam:
adding @renjithravindrankannath  as reviewer.

var/spack/repos/builtin/packages/py-torch/package.py
```diff
@@ -177,7 +177,8 @@ class PyTorch(PythonPackage, CudaPackage):
     depends_on("cudnn@7:", when="@1.6:+cudnn")
     depends_on("magma+cuda", when="+magma+cuda")
     depends_on("magma+rocm", when="+magma+rocm")
-    depends_on("nccl", when="+nccl")
+    depends_on("nccl", when="+nccl+cuda")
+    depends_on("rccl", when="+nccl+rocm")
```
renjithravindrankannath: This can be removed and when="+nccl" can be added in line number 188

renjithravindrankannath:
Also we need to have +cuda at line 458 since we are not adding the spec for nccl with rocm.
Build fails with below error.
`==> Error: KeyError: 'No spec with name nccl
`
And
`/spack/var/spack/repos/builtin/packages/py-torch/package.py:458, in setup_build_environment:
        455
        456        enable_or_disable("nccl")
        457        if "+nccl" in self.spec:
  >>    458            env.set("NCCL_LIB_DIR", self.spec["nccl"].libs.directories[0])
        459            env.set("NCCL_INCLUDE_DIR", self.spec["nccl"].prefix.include)
        460
        461        # cmake/External/nnpack.cmake`
Below change resolved the failure.

`         enable_or_disable("nccl")
-        if "+nccl" in self.spec:
+        if "+nccl+cuda" in self.spec:
             env.set("NCCL_LIB_DIR", self.spec["nccl"].libs.directories[0])`

adamjstewart:
Yep, there's a few other places where this is broken (unconditional rccl_prefix). This is not super intuitive. Should we create a new `+rccl` variant to control RCCL and only use `+nccl` for NCCL? I can make the same change in TF too if we decide to do this.

renjithravindrankannath:
> Yep, there's a few other places where this is broken (unconditional rccl_prefix). This is not super intuitive. Should we create a new `+rccl` variant to control RCCL and only use `+nccl` for NCCL? I can make the same change in TF too if we decide to do this.

Currently py-torch code is still is using the flag USE_NCCL with rocm build for now 
So would recommend to keep nccl for now.

renjithravindrankannath:
LGTM.. @adamjstewart

renjithravindrankannath (dismissed):

adamjstewart:
Added a new conflict for `+tensorpipe+rocm`, apparently this isn't yet supported.

renjithravindrankannath (dismissed):

renjithravindrankannath (approved):

alalazo (approved):

PR #32773, opened 2022-09-23
Add aws ofi rccl

bvanessen:
AWS OFI RCCL is a plug-in which enables EC2 developers to use [libfabric](https://github.com/ofiwg/libfabric) as a network provider while running [AMD's RCCL](https://github.com/ROCmSoftwarePlatform/rccl) based applications.

### DIFF
--- a/var/spack/repos/builtin/packages/aluminum/package.py
+++ b/var/spack/repos/builtin/packages/aluminum/package.py
@@ -51,9 +51,10 @@ class Aluminum(CMakePackage, CudaPackage, ROCmPackage):
         " Put/Get and IPC RMA functionality",
     )
     variant("rccl", default=False, description="Builds with support for RCCL communication lib")
-    variant("libfabric",
-            default=False,
-            description="Builds with support for libfabric enhanced RCCL communication lib")
+    variant(
+        "libfabric",
+        default=False,
+        description="Builds with support for libfabric enhanced RCCL communication lib")
 
     depends_on("cmake@3.21.0:", type="build", when="@1.0.1:")
     depends_on("cmake@3.17.0:", type="build", when="@:1.0.0")
--- a/var/spack/repos/builtin/packages/aluminum/package.py
+++ b/var/spack/repos/builtin/packages/aluminum/package.py
@@ -54,7 +54,8 @@ class Aluminum(CMakePackage, CudaPackage, ROCmPackage):
     variant(
         "libfabric",
         default=False,
-        description="Builds with support for libfabric enhanced RCCL communication lib")
+        description="Builds with support for libfabric enhanced RCCL communication lib",
+    )
 
     depends_on("cmake@3.21.0:", type="build", when="@1.0.1:")
     depends_on("cmake@3.17.0:", type="build", when="@:1.0.0")
--- a/var/spack/repos/builtin/packages/aluminum/package.py
+++ b/var/spack/repos/builtin/packages/aluminum/package.py
@@ -52,9 +52,9 @@ class Aluminum(CMakePackage, CudaPackage, ROCmPackage):
     )
     variant("rccl", default=False, description="Builds with support for RCCL communication lib")
     variant(
-        "libfabric",
+        "ofi_rccl_plugin",
         default=False,
-        description="Builds with support for libfabric enhanced RCCL communication lib",
+        description="Builds with support for OFI libfabric enhanced RCCL communication lib",
     )
 
     depends_on("cmake@3.21.0:", type="build", when="@1.0.1:")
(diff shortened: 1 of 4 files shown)


bvanessen:
@spackbot fix style

var/spack/repos/builtin/packages/aluminum/package.py
```diff
@@ -50,7 +50,12 @@ class Aluminum(CMakePackage, CudaPackage, ROCmPackage):
         description="Builds with support for CUDA intra-node "
         " Put/Get and IPC RMA functionality",
     )
-    variant("rccl", default=False, description="Builds with support for NCCL communication lib")
+    variant("rccl", default=False, description="Builds with support for RCCL communication lib")
+    variant(
+        "libfabric",
+        default=False,
```
becker33, line 56:
```suggestion
        default=False,
        when='+rccl',
```
Then the new conflict is unnecessary

var/spack/repos/builtin/packages/aluminum/package.py
```diff
@@ -62,8 +67,13 @@ class Aluminum(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("cub", when="@:0.1,0.6.0: +cuda ^cuda@:10")
     depends_on("hipcub", when="@:0.1,0.6.0: +rocm")
 
+    depends_on("rccl", when="+rccl")
+    depends_on("aws-ofi-rccl", when="+libfabric platform=cray")
+
     conflicts("~cuda", when="+cuda_rma", msg="CUDA RMA support requires CUDA")
     conflicts("+cuda", when="+rocm", msg="CUDA and ROCm support are mutually exclusive")
+    conflicts("+nccl", when="+rccl", msg="NCCL and RCCL support are mutually exclusive")
+    conflicts("~rccl", when="+libfabric", msg="libfabric enhancements require RCCL support")
```
rajachan: Not familiar with aluminum, but I'm curious why this is the case. This in itself does not guarantee that libfabric's ROCm support is enabled (so p2p transfers with FI_HMEM_ROCR capability might not be supported).  From a middleware's perspective, you might be better off querying HMEM capabilities you need at fi_getinfo() to be sure. 

This is pointing out a gap we have in the libfabric package.py though. We need to add the CUDA/ROCR build variants that also pull in the respective runtimes as deps (like how you've done for the aws-ofi-rccl package).

var/spack/repos/builtin/packages/aluminum/package.py
```diff
@@ -62,8 +67,13 @@ class Aluminum(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("cub", when="@:0.1,0.6.0: +cuda ^cuda@:10")
     depends_on("hipcub", when="@:0.1,0.6.0: +rocm")
 
+    depends_on("rccl", when="+rccl")
```
becker33, line 70:
```suggestion
    depends_on("rccl", when="+rccl~libfabric")
```
I'm assuming `rccl` and `aws-ofi-rccl` are mutually exclusive?

If that's the case, it might be better to model this with a virtual, and rccl and aws-ofi-rccl both provide it.

rajachan (dismissed):
Looks fine, but the libfabric+HMEM piece needs some work to ensure the right capabilities are enabled.

var/spack/repos/builtin/packages/aluminum/package.py
rajachan: > I'm assuming rccl and aws-ofi-rccl are mutually exclusive?

`aws-ofi-rccl` depends on definitions in `rccl`... providing a dynamically loadable network plugin for the latter.

bvanessen: @becker33 No they are not mutually exclusive.  The `aws-ofi-rccl` package is a plugin to `rccl` on certain systems that can accelerate the `rccl` calls.  Note that not all systems would want it.

var/spack/repos/builtin/packages/aluminum/package.py
bvanessen: @becker33 We need to be careful because `libfabric` support is only valid on some `cray` systems, but not all.

var/spack/repos/builtin/packages/aluminum/package.py
bvanessen: @rajachan Hmmm, I am not too sure of what you are suggesting.  I guess I put this in here now because we don't have any software combination that would allow us to use the `rccl` `libfabric` plugin without already having `rccl` support. Perhaps this variant should be renamed because it is not tying `libfabric` support directly into `aluminum`, it is just adding the dependency on the `aws-ofi-rccl` plugin.

bvanessen:
@rajachan @becker33 Is there anything else that I should be doing with this PR?

var/spack/repos/builtin/packages/aluminum/package.py
rajachan: Yeah, the name is what threw me off. Looks good now.

rajachan (approved):
Looks good to me.

bvanessen:
@spackbot run pipeline

alalazo (approved):

PR #32775, opened 2022-09-23
pgplot: install rbg.txt and change PGPLOT_DIR

tukss:
The file `rbg.txt` is needed for many PGPLOT application, such as the MESA stellar evolution code. This change installs the file to the `PGPLOT_DIR`, where the library expects it.

`PGPLOT_DIR` was previously set to the prefix itself, which is an odd place to install `rgb.txt`. This commit changes it to `lib/pgplot5`, following the convention used by Debian.


eschnett (approved):
This looks good.

tldahlgren (approved):

Issue #29447, opened 2022-03-10
spack install --test root not working properly in environment


bvanessen:
### Steps to reproduce

When installing py-numpy in a environment with other root packages and the `--test root` flag is set, it fails to build because it cannot find `py-pytest` and it doesn't show up as a dependency in the concretization.

```
spack add lbann@gitlab-pip-ivybridge +lld +deterministic +vision +numpy +onednn +half +fft ^hydrogen@develop ^dihydrogen@develop ^aluminum@master %clang ^mvapich2 ^hwloc@1.11.13 ^libtool@2.4.2 ^python@3.9.10
spack develop --no-clone -p /usr/WS2/lbannusr/brian-lbann.git lbann@gitlab-pip-ivybridge
spack add py-numpy@1.16.0:
spack concretize --reuse --test root
spack install --reuse -j26 --test root
```

### Error message

```
==> [2022-03-09-13:26:12.477072] '/usr/WS2/lbannusr/spack_installed_packages/catalyst/opt/spack/linux-rhel7-ivybridge/clang-12.0.1/python-3.9.10-qassf33p2hkuptbbkb2yzlq4odqmngej/bin/python3.9' '-c' 'import numpy; numpy.test("full", verbose=2)'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/WS2/lbannusr/spack_installed_packages/catalyst/opt/spack/linux-rhel7-ivybridge/clang-12.0.1/py-numpy-1.22.2-wasb6dleomiv433qdtmlkn3msjhzzeqp/lib/python3.9/site-packages/numpy/_pytesttester.py", line 128, in __call__
    import pytest
ModuleNotFoundError: No module named 'pytest'
```
[spack_lock.txt](https://github.com/spack/spack/files/8226249/spack_lock.txt)
[spack_yaml.txt](https://github.com/spack/spack/files/8226250/spack_yaml.txt)
[spack-build-env.txt](https://github.com/spack/spack/files/8226251/spack-build-env.txt)
[spack-build-lbann-gitlab-pip-ivybridge.log](https://github.com/spack/spack/files/8226252/spack-build-lbann-gitlab-pip-ivybridge.log)
[spack-build-out.txt](https://github.com/spack/spack/files/8226253/spack-build-out.txt)

### Information on your system

This happened on pascal, catalyst, and lassen at LLNL LC.

```
* **Spack:** 0.17.1-1431-1c23f5abe0
* **Python:** 3.7.2
[...]

PR #32779, opened 2022-09-23
py-beautifulsoup4: add 4.11.1

manuelakuhn:
sources from pypi

testing with `--test=root` fails due to missing `py-pytest` but because I suspect that it is due to #29447, I did not add it as test dependency (once added the tests run fine).


tldahlgren (approved):
Confirmed sha256.

adamjstewart (approved):

PR #32789, opened 2022-09-23
py-coverage: add 6.4.4

manuelakuhn:
https://github.com/nedbat/coveragepy/tree/6.4.4

[This line](https://github.com/nedbat/coveragepy/blob/6.4.4/setup.py#L111) in the `setup.py` where they specify `python_full_version<="3.11.0a6"` was not clear to me, so I restricted the python version used when `+toml` is activated.

### DIFF
--- a/var/spack/repos/builtin/packages/py-coverage/package.py
+++ b/var/spack/repos/builtin/packages/py-coverage/package.py
@@ -26,10 +26,9 @@ class PyCoverage(PythonPackage):
     variant("toml", default=False, description="Enable pyproject.toml support")
 
     depends_on("python@3.7:", when="@6.3:", type=("build", "run"))
-    depends_on("python@3.7:3.11.0a6", when="@6.4.1: +toml", type=("build", "run"))
     depends_on("python@3.6:", when="@6:", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


tldahlgren (dismissed):
Confirmed sha256.

var/spack/repos/builtin/packages/py-coverage/package.py
```diff
@@ -25,6 +26,7 @@ class PyCoverage(PythonPackage):
     variant("toml", default=False, description="Enable pyproject.toml support")
 
     depends_on("python@3.7:", when="@6.3:", type=("build", "run"))
+    depends_on("python@3.7:3.11.0a6", when="@6.4.1: +toml", type=("build", "run"))
```
adamjstewart: Undo this change and use this instead:
```python
depends_on('py-tomli', when='@6: +toml ^python@:3.10', type=('build', 'run'))
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-coverage/package.py
manuelakuhn: Done

adamjstewart (approved):

PR #32792, opened 2022-09-23
py-eccodes: add version 1.5.0

skosukhin:
This is #32728 (which I would like to keep for now) minus the hack:

* add constraints on the runtime dependency `eccodes`;
* set less general environment variables that enable location of `libeccodes`;
* add version `1.5.0`.


tldahlgren (approved):
Confirmed sha256.

adamjstewart (approved):

PR #32793, opened 2022-09-23
py-cdo: add version 1.5.6, deprecate 1.3.2

skosukhin:
* add latest version `1.5.6`
* deprecate very old `1.3.2` (the only one we currently have)
* update dependencies
* setup runtime environment
* add maintainers (@Try2Code, please, confirm here that you agree to be on the list)

### DIFF
--- a/var/spack/repos/builtin/packages/py-cdo/package.py
+++ b/var/spack/repos/builtin/packages/py-cdo/package.py
@@ -12,10 +12,21 @@ class PyCdo(PythonPackage):
 
     pypi = "cdo/cdo-1.3.2.tar.gz"
 
-    version("1.3.2", sha256="9f78879d90d14134f2320565016d0d371b7dfe7ec71110fd313868ec1db34aee")
+    maintainers = ["Try2Code", "skosukhin"]
 
-    depends_on("cdo")
+    version("1.5.6", sha256="fec1a75382f01b3c9c368e8f143d98b12323e06975663f87d9b60c739ae1d335")
+    version(
+        "1.3.2",
+        sha256="9f78879d90d14134f2320565016d0d371b7dfe7ec71110fd313868ec1db34aee",
+        deprecated=True,
+    )
 
     depends_on("py-setuptools", type="build")
-    depends_on("py-scipy", type=("build", "run"))
-    depends_on("py-netcdf4", type=("build", "run"))
+    depends_on("cdo+netcdf", type="run")
+    depends_on("py-netcdf4", type="run")
+    depends_on("py-scipy", type="run", when="@:1.4")
+    depends_on("py-xarray", type="run", when="@1.3.4:")
+    depends_on("py-six", type="run", when="@1.3.3:")
+
+    def setup_run_environment(self, env):
+        env.set("CDO", self.spec["cdo"].prefix.bin.cdo)
--- a/var/spack/repos/builtin/packages/py-cdo/package.py
+++ b/var/spack/repos/builtin/packages/py-cdo/package.py
@@ -21,6 +21,8 @@ class PyCdo(PythonPackage):
         deprecated=True,
     )
 
+    depends_on("python@2.7:", type=("build", "run"))
+
     depends_on("py-setuptools", type="build")
     depends_on("cdo+netcdf", type="run")
     depends_on("py-netcdf4", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-cdo/package.py
+++ b/var/spack/repos/builtin/packages/py-cdo/package.py
@@ -23,10 +23,10 @@ class PyCdo(PythonPackage):
 
     depends_on("py-setuptools", type="build")
     depends_on("cdo+netcdf", type="run")
-    depends_on("py-netcdf4", type="run")
-    depends_on("py-scipy", type="run", when="@:1.4")
-    depends_on("py-xarray", type="run", when="@1.3.4:")
-    depends_on("py-six", type="run", when="@1.3.3:")
+    depends_on("py-netcdf4", type=("build", "run"))
+    depends_on("py-scipy", type=("build", "run"), when="@:1.4")
+    depends_on("py-xarray", type=("build", "run"), when="@1.3.4:")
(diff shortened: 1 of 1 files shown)


tldahlgren (dismissed):
Confirmed the new sha256.

Try2Code:
@skosukhin - happy to be on the maintainer site for this package, thx for the initiative to work on this. I was not even aware of this spack package tbh

skosukhin:
@tldahlgren regarding the change of the title, it doesn't add version `1.3.2`, it depreciates it

tldahlgren:
> @skosukhin - happy to be on the maintainer site for this package, thx for the initiative to work on this. I was not even aware of this spack package tbh

Thanks for the correction. Changed the title.

var/spack/repos/builtin/packages/py-cdo/package.py
```diff
@@ -12,10 +12,21 @@ class PyCdo(PythonPackage):
...
+    )
 
     depends_on("py-setuptools", type="build")
-    depends_on("py-scipy", type=("build", "run"))
-    depends_on("py-netcdf4", type=("build", "run"))
+    depends_on("cdo+netcdf", type="run")
+    depends_on("py-netcdf4", type="run")
+    depends_on("py-scipy", type="run", when="@:1.4")
+    depends_on("py-xarray", type="run", when="@1.3.4:")
+    depends_on("py-six", type="run", when="@1.3.3:")
```
adamjstewart: These should all be `type=('build', 'run')` so that the import tests work properly

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-cdo/package.py
skosukhin: Ok. I did that. For the future, what command fails if I don't do that? I tried `spack install --test=root py-cdo` and it worked without this change.

Try2Code:
Installing and using the package went just fine on my local machine
- CDO binary build from its dependency was properly build
- python tests ran fine with only the py-cdo package loaded before
- the environment  variable CDO was properly used if set
Please let me know, in case I missed something

var/spack/repos/builtin/packages/py-cdo/package.py
adamjstewart: That should trigger it, maybe another dependency is adding them to `PYTHONPATH`, but not a great behavior to rely on.

var/spack/repos/builtin/packages/py-cdo/package.py
```diff
@@ -12,10 +12,21 @@ class PyCdo(PythonPackage):
...
-    version("1.3.2", sha256="9f78879d90d14134f2320565016d0d371b7dfe7ec71110fd313868ec1db34aee")
+    maintainers = ["Try2Code", "skosukhin"]
 
-    depends_on("cdo")
+    version("1.5.6", sha256="fec1a75382f01b3c9c368e8f143d98b12323e06975663f87d9b60c739ae1d335")
+    version(
+        "1.3.2",
+        sha256="9f78879d90d14134f2320565016d0d371b7dfe7ec71110fd313868ec1db34aee",
+        deprecated=True,
+    )
```
adamjstewart, line 23:
```python
```

var/spack/repos/builtin/packages/py-cdo/package.py
```diff
@@ -12,10 +12,21 @@ class PyCdo(PythonPackage):
...
 
     depends_on("py-setuptools", type="build")
-    depends_on("py-scipy", type=("build", "run"))
+    depends_on("cdo+netcdf", type="run")
     depends_on("py-netcdf4", type=("build", "run"))
+    depends_on("py-scipy", type=("build", "run"), when="@:1.4")
+    depends_on("py-xarray", type=("build", "run"), when="@1.3.4:")
```
adamjstewart, line 30: Seems like netcdf4 and xarray are optional, not sure if we should add variants for these or always add them

var/spack/repos/builtin/packages/py-cdo/package.py
Try2Code: true, the code is python2 and python3. Are there any opinions about dropping python2 support?

var/spack/repos/builtin/packages/py-cdo/package.py
Try2Code: I recommend to always add them. From my experience most users have workflows, which involve netcdf4 files and XArray. Adding those dependencies just makes the package more user friendly IMO.

var/spack/repos/builtin/packages/py-cdo/package.py
skosukhin: @adamjstewart ok but why should they be `build` and not `test` then?

var/spack/repos/builtin/packages/py-cdo/package.py
skosukhin: We have discussed this with @Try2Code, the upstream developer, and agreed that the package loses significant amount of its functionality without `py-netcdf4` and `py-xarray`. We can, of course, add a variant for both now (if I am not mistaken, using one without another does not make much sense) but I think we can also do that later if there is a need to get rid of the dependencies in some cases. Given that the spackage has not been updated for a very long time, I assume that adding a variant later should not affect too many users.

var/spack/repos/builtin/packages/py-cdo/package.py
adamjstewart: Python 2 support will be dropped in Spack after the next release, but it's probably better to document this in the package anyway.

var/spack/repos/builtin/packages/py-cdo/package.py
adamjstewart: Hmm, I think you might be right, they should technically be `type=('test', 'run')`. But every other package uses `type=('build', 'run')` at the moment. Personally I would prefer to be consistent. Someday, I would like Spack to be smart enough to realize that post-install tests ARE run-time tests, so env vars should be modified accordingly. Maybe @tldahlgren can comment on this.

adamjstewart (approved):

PR #32795, opened 2022-09-23
e4s: use ubuntu 20.04 image and %gcc@11.1.0

eugeneswalker:
Downgrade runner image used for building E4S from Ubuntu 22.04 to Ubuntu 20.04
* `ecpe4s/ubuntu20.04-runner-x86_64:2022-10-01`
* `%gcc@11.1.0`

FYI @wspear @kwryankrattiger

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -267,11 +267,11 @@ e4s-mac-protected-build:
 
 e4s-pr-generate:
   extends: [ ".e4s", ".pr-generate"]
-  image: ecpe4s/ubuntu22.04-runner-x86_64:2022-07-01
+  image: ecpe4s/ubuntu20.04-runner-x86_64:2022-10-01
 
 e4s-protected-generate:
   extends: [ ".e4s", ".protected-generate"]
-  image: ecpe4s/ubuntu22.04-runner-x86_64:2022-07-01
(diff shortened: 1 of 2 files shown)


kwryankrattiger:
Do we want to downgrade to gcc@9.4.0 here? I think there is still an option to use gcc11, which is what is being used on facility deployments, so having that consistent is kind of nice.

Although, if facility deployments will also be downgraded to an older compiler I think that is fine.

eugeneswalker:
> Do we want to downgrade to gcc@9.4.0 here? I think there is still an option to use gcc11, which is what is being used on facility deployments, so having that consistent is kind of nice.

I would like to stick with GCC 11.2.0 because that is what is prevailing on the facilities we are most interested in. But the problem I see here is, in order to use `gcc@11.2.0` we would have to install it in this image using Spack. AFAIK, compiler paths are not relocated as part of the current binary relocation logic. So things like MPI that store the GCC location in the wrapper scripts, those paths would not be relocated when someone installs the generated binaries into an environment where GCC is in a different location. What do you think? Any thoughts on the best course here @scottwittenburg ?

kwryankrattiger:
I am using the e4s ubuntu20 2022-08-01 runner and it seems to have gcc 11.1 on it, I didn't have to build a compiler using spack.

eugeneswalker:
@spackbot run pipeline

kwryankrattiger (approved):

wspear (approved):

eugeneswalker:
@spackbot run pipeline

PR #32796, opened 2022-09-23
binary_distribution: parallelize fetches in buildcache update-index

scottwittenburg:
Parallelize fetching of spec files when updating a buildcache index to reduce time spent on it.  Additionally, if the mirror is on `s3` and the `aws` cli program is available, use that to fetch the spec files, and still use multiprocessing to read them.  Roughly, using the parallelized fetching approach results in about a 3x speedup, while using the `aws s3 sync` approach results in about a 16x speedup.

Below are some results comparing `develop` to the parallel spec fetching as well as to using `aws s3 sync`.  This was on my development machine,  a dual 8-core linux box running Ubuntu 18.04.

### On a mirror with 973 specs

If `aws s3 sync` is available:

```
    real	0m29.765s
    user	0m21.364s
    sys	0m2.356s
```

With just parallelized fetching:

```
    real	2m40.477s
    user	2m58.396s
    sys	1m8.703s
```

With spack `develop`:

```
    real	7m42.423s
    user	1m19.887s
    sys	0m6.379s
```

### On a mirror with 7209 specs

If `aws s3 sync` is available:

```
    real    3m42.034s
    user    2m56.416s
    sys	    0m19.803s
```

With just parallelized fetching:

```
    real    19m14.473s
    user    21m50.153s
    sys     8m32.921s
```

With spack `develop`:

```
    real    60m19.367s
    user    12m0.187s
    sys     0m43.325s
```

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -7,6 +7,7 @@ import codecs
 import collections
 import hashlib
 import json
+import multiprocessing.pool
 import os
 import shutil
 import sys
@@ -45,6 +46,7 @@ from spack.caches import misc_cache_location
 from spack.relocate import utf8_paths_to_single_binary_regex
 from spack.spec import Spec
 from spack.stage import Stage
+from spack.util.executable import which
 
 _build_cache_relative_path = "build_cache"
 _build_cache_keys_relative_path = "_pgp"
@@ -70,6 +72,10 @@ class FetchCacheError(Exception):
             err = errors[0]
             self.message = "{0}: {1}".format(err.__class__.__name__, str(err))
         super(FetchCacheError, self).__init__(self.message)
+
+
+class ListMirrorSpecsError(spack.error.SpackError):
+    """Raised when unable to retrieve list of specs from the mirror"""
 
 
 class BinaryCacheIndex(object):
@@ -881,37 +887,52 @@ def sign_specfile(key, force, specfile_path):
     spack.util.gpg.sign(key, specfile_path, signed_specfile_path, clearsign=True)
 
 
-def _fetch_spec_from_mirror(spec_url):
-    s = None
-    tty.debug("fetching {0}".format(spec_url))
-    _, _, spec_file = web_util.read_from_url(spec_url)
-    spec_file_contents = codecs.getreader("utf-8")(spec_file).read()
-    # Need full spec.json name or this gets confused with index.json.
-    if spec_url.endswith(".json.sig"):
-        specfile_json = Spec.extract_json_from_clearsig(spec_file_contents)
-        s = Spec.from_dict(specfile_json)
-    elif spec_url.endswith(".json"):
-        s = Spec.from_json(spec_file_contents)
-    elif spec_url.endswith(".yaml"):
-        s = Spec.from_yaml(spec_file_contents)
-    return s
+def _read_specs_and_push_index(file_list, read_method, cache_prefix, db, temp_dir, concurrency):
+    """Read all the specs listed in the provided list, using thread given thread parallelism,
+        generate the index, and push it to the mirror.
 
+    Args:
+        file_list (list(str)): List of urls or file paths pointing at spec files to read
+        read_method: A function taking a single argument, either a url or a file path,
+            and which reads the spec file at that location, and returns the spec.
+        cache_prefix (str): prefix of the build cache on s3 where index should be pushed.
+        db: A spack database used for adding specs and then writing the index.
+        temp_dir (str): Location to write index.json and hash for pushing
+        concurrency (int): Number of parallel processes to use when fetching
 
-def _read_specs_and_push_index(file_list, cache_prefix, db, db_root_dir):
-    for file_path in file_list:
-        try:
-            s = _fetch_spec_from_mirror(url_util.join(cache_prefix, file_path))
-        except (URLError, web_util.SpackWebError) as url_err:
-            tty.error("Error reading specfile: {0}".format(file_path))
-            tty.error(url_err)
+    Return:
+        None
+    """
 
-        if s:
-            db.add(s, None)
-            db.mark(s, "in_buildcache", True)
+    def _fetch_spec_from_mirror(spec_url):
+        spec_file_contents = read_method(spec_url)
+
+        if spec_file_contents:
+            # Need full spec.json name or this gets confused with index.json.
+            if spec_url.endswith(".json.sig"):
+                specfile_json = Spec.extract_json_from_clearsig(spec_file_contents)
+                return Spec.from_dict(specfile_json)
+            if spec_url.endswith(".json"):
+                return Spec.from_json(spec_file_contents)
+            if spec_url.endswith(".yaml"):
+                return Spec.from_yaml(spec_file_contents)
+
+    tp = multiprocessing.pool.ThreadPool(processes=concurrency)
+    try:
+        fetched_specs = tp.map(
+            llnl.util.lang.star(_fetch_spec_from_mirror), [(f,) for f in file_list]
+        )
+    finally:
+        tp.terminate()
+        tp.join()
+
+    for fetched_spec in fetched_specs:
+        db.add(fetched_spec, None)
+        db.mark(fetched_spec, "in_buildcache", True)
 
     # Now generate the index, compute its hash, and push the two files to
     # the mirror.
-    index_json_path = os.path.join(db_root_dir, "index.json")
+    index_json_path = os.path.join(temp_dir, "index.json")
     with open(index_json_path, "w") as f:
         db._write_to_file(f)
 
@@ -921,7 +942,7 @@ def _read_specs_and_push_index(file_list, cache_prefix, db, db_root_dir):
         index_hash = compute_hash(index_string)
 
     # Write the hash out to a local file
-    index_hash_path = os.path.join(db_root_dir, "index.json.hash")
+    index_hash_path = os.path.join(temp_dir, "index.json.hash")
     with open(index_hash_path, "w") as f:
         f.write(index_hash)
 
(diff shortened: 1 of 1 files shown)


eugeneswalker:
I just used this PR to update the index on `cache.e4s.io` where we have 92,000+ binaries going back over a year. It took ~7.3 hours with this PR, and believe it or not, that was a big improvement from what it was taking before! Thank you Scott!

scottwittenburg:
> It took ~7.3 hours with this PR, and believe it or not, that was a big improvement

That's still pretty disappointing.  Unless it was taking 21 hours previously.

But I guess it finished and produced the index?

eugeneswalker:
> But I guess it finished and produced the index?

It did finish OK, yes. First time I've had an up-to-date index for `cache.e4s.io` in longer than I can remember. It seemed to only be downloading three at a time -- at least that's the way it looked from the debug output. Just glancing at the code now, it looks like it should be downloading 32 at a time? Maybe it just appeared to only be pulling 3 at a time.

scottwittenburg:
Yes, the default concurrency is 32, so it should be attempting to do 32 at a time.  But you're saying when you ran with `spack -d buildcache update-index ...` the output made it look like it was only doing three concurrently?

eugeneswalker:
> ... you're saying when you ran with `spack -d buildcache update-index ...` the output made it look like it was only doing three concurrently?

It did appear that way, yes. I had debug turned on, and so it printed out `fetching ...` messages, and it seemed they were coming through in groups of three. I'm not sure how much stock to put in that as an indication that it wasn't actually fetching 32 at a time. Maybe it was an illusion.

scottwittenburg:
Yeah, hard to know from that.  I guess better would be to know how long it took when run with `develop` and compare it to the 7 hours you mentioned above.  Unfortunately not possible if the `develop` version doesn't finish.  For me so far, it looks like it takes about three times as long without this PR on my dual 8-core machine (so 32 hyper-threads).

scottwittenburg:
@spackbot help

scottwittenburg:
@tgamblin You asked what kind of performance gain we could get using `aws s3 sync` to fetch the specs for generating the binary index.  In the description you can see some results.  For a mirror with ~7K specs, `develop` took an hour, parallelized fetching took 20 min, and using `aws s3 sync` took under 4 min.

Some things to be aware of if we want to use `aws s3 sync`:

- You have to have `aws` cli program installed
- You need to have IAM credentails set (no big deal if you're already trying to write the index to an s3 bucket anyway)
- You need to have enough disk space to temporarily store all the specs you fetch (that was round 300MB for ~7K specs, so not necessarily a showstopper even for the 90K+ specs in cache.e4s.io) 

So if we want to install the `aws` cli on all the images we use in gitlab CI, we can get a lot of improvement in the time to generate a buildcache index.  For example, this [job](https://gitlab.spack.io/spack/spack/-/jobs/3690596) on a `develop` pipeline yesterday took 242 minutes to rebuild the e4s index, that could be down to ~15 minutes.

This PR currently tries to use `aws s3 sync` if it's an s3 mirror and the `aws` program is found.  However, if that attempt fails for any reason, it falls back to parallelized fetching.  

@eugeneswalker @zackgalbreath

kwryankrattiger (dismissed):

lib/spack/spack/binary_distribution.py
```diff
@@ -795,33 +797,39 @@ def sign_specfile(key, force, specfile_path):
...
+                s = Spec.from_json(spec_file_contents)
+            elif spec_url.endswith(".yaml"):
+                s = Spec.from_yaml(spec_file_contents)
 
         if s:
-            db.add(s, None)
-            db.mark(s, "in_buildcache", True)
+            fetched_specs.append(s)
+
+    tp = multiprocessing.pool.ThreadPool(processes=concurrency)
```
alalazo, line 920: Is there any reason to use a `ThreadPool` instead of a `Pool`?

lib/spack/spack/binary_distribution.py
```diff
@@ -795,33 +797,39 @@ def sign_specfile(key, force, specfile_path):
...
-        try:
-            s = _fetch_spec_from_mirror(url_util.join(cache_prefix, file_path))
-        except (URLError, web_util.SpackWebError) as url_err:
-            tty.error("Error reading specfile: {0}".format(file_path))
-            tty.error(url_err)
+def _read_specs_and_push_index(file_list, read_method, cache_prefix, db, db_root_dir, concurrency):
+    fetched_specs = []
+
+    def _fetch_spec_from_mirror(spec_url):
+        spec_file_contents = None
```
alalazo: The variable is reinitialized below
```suggestion
```

lib/spack/spack/binary_distribution.py
```diff
@@ -795,33 +797,39 @@ def sign_specfile(key, force, specfile_path):
...
-
-
-def _read_specs_and_push_index(file_list, cache_prefix, db, db_root_dir):
-    for file_path in file_list:
-        try:
-            s = _fetch_spec_from_mirror(url_util.join(cache_prefix, file_path))
-        except (URLError, web_util.SpackWebError) as url_err:
-            tty.error("Error reading specfile: {0}".format(file_path))
-            tty.error(url_err)
+def _read_specs_and_push_index(file_list, read_method, cache_prefix, db, db_root_dir, concurrency):
```
alalazo: This might benefit from a docstring. It's really not clear how `read_method` should look like, and we have to trace it back to the implementation at the calling site.

lib/spack/spack/binary_distribution.py
```diff
@@ -856,32 +864,87 @@ def _read_specs_and_push_index(file_list, cache_prefix, db, db_root_dir):
...
+            msg = "No packages at {0}: {1}".format(cache_prefix, inst)
+            tty.warn(msg)
+            return
+        except Exception as err:
+            # If we got some kind of S3 (access denied or other connection
+            # error), the first non boto-specific class in the exception
+            # hierarchy is Exception.  Just print a warning and return
+            msg = "Encountered problem listing packages at {0}: {1}".format(cache_prefix, err)
+            tty.warn(msg)
+            return
```
alalazo: Can we extract different functions here, and have code like:
```console
file_list, read_fn = spec_files_from_cache(cache_prefix)
```
That would make clear that the important result of the first 50 lines of computation is the list of `spec.json.sig` files and the "read function". Also, it would be good to have a single function that retrieves the list using `aws` cli, and another separate function implementing the fallback. 

In this way `spec_files_from_cache` would just be responsible to decide when to call the other two functions.

lib/spack/spack/binary_distribution.py
```diff
@@ -795,33 +797,39 @@ def sign_specfile(key, force, specfile_path):
...
+                s = Spec.from_yaml(spec_file_contents)
 
         if s:
-            db.add(s, None)
-            db.mark(s, "in_buildcache", True)
+            fetched_specs.append(s)
+
+    tp = multiprocessing.pool.ThreadPool(processes=concurrency)
+    try:
+        tp.map(llnl.util.lang.star(_fetch_spec_from_mirror), [(f,) for f in file_list])
```
alalazo: I think this call to map has either a race condition or is serialized by the GIL:
1.  `_fetch_spec_from_mirror` returns `None`
2. Therefore the result (we discard) from `tp.map` is a list of `None`s
3. The computations we do next are based on a global object (`fetched_specs`) that is bound to `_fetch_spec_from_mirror` as a closure

It's probably better to have `_fetch_spec_from_mirror` return a spec and turn this line into:
```suggestion
        fetched_specs = tp.map(llnl.util.lang.star(_fetch_spec_from_mirror), [(f,) for f in file_list])
```

lib/spack/spack/binary_distribution.py
```diff
@@ -795,33 +797,39 @@ def sign_specfile(key, force, specfile_path):
...
+
+        if spec_file_contents:
+            # Need full spec.json name or this gets confused with index.json.
+            if spec_url.endswith(".json.sig"):
+                specfile_json = Spec.extract_json_from_clearsig(spec_file_contents)
+                s = Spec.from_dict(specfile_json)
+            elif spec_url.endswith(".json"):
+                s = Spec.from_json(spec_file_contents)
+            elif spec_url.endswith(".yaml"):
+                s = Spec.from_yaml(spec_file_contents)
```
alalazo: If we use this with a process pool, it's probably better to return the single result, instead of relying on side effects over a global variable:
```suggestion
            if spec_url.endswith(".json.sig"):
                specfile_json = Spec.extract_json_from_clearsig(spec_file_contents)
                return Spec.from_dict(specfile_json)
            if spec_url.endswith(".json"):
                return Spec.from_json(spec_file_contents)
            if spec_url.endswith(".yaml"):
                return Spec.from_yaml(spec_file_contents)
```

lib/spack/spack/binary_distribution.py
```diff
@@ -895,13 +958,17 @@ def generate_package_index(cache_prefix):
...
+            file_list, read_method, cache_prefix, db, db_root_dir, concurrency
+        )
     except Exception as err:
         msg = "Encountered problem pushing package index to {0}: {1}".format(cache_prefix, err)
         tty.warn(msg)
         tty.debug("\n" + traceback.format_exc())
     finally:
         shutil.rmtree(tmpdir)
+        if tmpspecsdir:
+            shutil.rmtree(tmpspecsdir)
```
alalazo: This is needed only if we use `aws`. It would be better if we can scope it with that "reading strategy", not in a higher level function.

alalazo (changes requested):
We can possibly improve the use of `multiprocessing.pool`. I also have a question on `ThreadPool` vs. `Pool` and a few suggestions to improve structure and readability. Let me know what you think.

lib/spack/spack/binary_distribution.py
scottwittenburg: I've read that `ThreadPool` is better for IO bound tasks (e.g. like in this case), while `Pool` is better for CPU bound tasks.  For example, see [here](https://superfastpython.com/threadpool-vs-pool-in-python/).  But I'm open to discussion.

lib/spack/spack/binary_distribution.py
alalazo: @scottwittenburg :point_up: This is the most important comment

scottwittenburg: Yes, thanks, I'm attempting to address this now. :+1:

scottwittenburg:
Thanks for your review @alalazo, I like this better after your suggested changes.  I think indeed the GIL was serializing access to the closure variable, due to using `ThreadPool` instead of `Pool`, so maybe there wasn't actually a race condition.  But removing the possibility of it makes sense to me anyway.

I think I addressed all your suggestions, but let me know what you think when you get a chance.

lib/spack/spack/binary_distribution.py
```diff
@@ -881,33 +883,44 @@ def sign_specfile(key, force, specfile_path):
...
+    The argument ``read_method`` should be a function taking a single argument, either a
+    url or a file path, and which reads the spec file at that location, and returns the
+    spec.
+    """
 
-        if s:
-            db.add(s, None)
-            db.mark(s, "in_buildcache", True)
+    def _fetch_spec_from_mirror(spec_url):
+        tty.debug("reading {0}".format(spec_url))
```
alalazo: Let's remove the debug print from the worker function.
```suggestion
```

lib/spack/spack/binary_distribution.py
```diff
@@ -942,32 +955,125 @@ def _read_specs_and_push_index(file_list, cache_prefix, db, db_root_dir):
...
 
 
-def generate_package_index(cache_prefix):
-    """Create the build cache index page.
+def _specs_from_cache_aws_cli(cache_prefix):
+    """
+    Use aws cli to sync all the specs into a local temp directory, returning
+    the list of local file paths and a function that can read each one from
+    the file system.
+    """
```
alalazo: Can we use Google style for docstrings? That would be:
```suggestion
    """Use aws cli to sync all the specs into a local temporary directory.
    
    Args:
        cache_prefix (str): prefix of the build cache on s3
        
    Return:
        List of the local file paths and a function that can read each one from the file system.
    """
```

lib/spack/spack/binary_distribution.py
```diff
@@ -942,32 +955,125 @@ def _read_specs_and_push_index(file_list, cache_prefix, db, db_root_dir):
...
+        "*.spec.yaml",
+        cache_prefix,
+        tmpspecsdir,
+    ]
 
-    Creates (or replaces) the "index.json" page at the location given in
-    cache_prefix.  This page contains a link for each binary package (.yaml or
-    .json) under cache_prefix.
+    try:
+        tty.msg(
```
alalazo: Should we demote this and the next message to debug level?

lib/spack/spack/binary_distribution.py
```diff
@@ -942,32 +955,125 @@ def _read_specs_and_push_index(file_list, cache_prefix, db, db_root_dir):
...
+    read_fn = None
+    file_list = None
+
+    if aws:
+        file_list, read_fn = _specs_from_cache_aws_cli(cache_prefix)
+
+    if not read_fn or not file_list:
+        file_list, read_fn = _specs_from_cache_fallback(cache_prefix)
+
+    return file_list, read_fn
```
alalazo: I think here we might do even better, and have code like:
```python
callbacks = []
if cache_prefix.startswith("s3"):
    callbacks.append(_specs_from_cache_aws_cli)
    
callbacks.append(_specs_from_cache_fallback)

for specs_from_cache_fn in callbacks:
    file_list, read_fn = specs_from_cache_fn(cache_prefix)
    if file_list:
    	return file_list, read_fn

raise ....
```

alalazo (changes requested):
A few minor comments Once those are in, LGTM.

alalazo (approved):

alalazo:
I tried the fallback method locally, didn't try `s3`. Let's check this PR in action on a develop pipeline! Thanks @scottwittenburg

PR #32801, opened 2022-09-24
Environments: store spack version/commit in spack.lock

tldahlgren:
(Updated)  This PR adds tracking ~and using~ an environment's Spack version/commit to enhance reproducibility.

Changes include:
- adding the version/commit to `spack.lock`;
- refactoring `spack.main.get_version()`; and
- fixes a couple of environment lock file-related typos.

TODO:

~- [x] Update `test/data/legacy_env/v4.lock` to actually reflect version 4 of the lock file~
~- [ ] Add use of the version information~
- [x] Add more unit tests as needed


lib/spack/spack/environment/__init__.py
```diff
@@ -16,11 +16,12 @@
 The high-level format of a Spack lockfile hasn't changed much between versions, but the
 contents have.  Lockfiles are JSON-formatted and their top-level sections are:
 
-  1. ``_meta`` (object): this contains deatails about the file format, including:
+  1. ``_meta`` (object): this contains details about the file format, including:
       * ``file-type``: always ``"spack-lockfile"``
       * ``lockfile-version``: an integer representing the lockfile format version
       * ``specfile-version``: an integer representing the spec format version (since
         ``v0.17``)
+      * ``spack-commit``: the Spack git commit (since ``v0.19``)
```
tgamblin: This doesn't belong under `_meta` -- it's a top level field.  `_meta` is really about the format of *this* file.

tgamblin: This doesn't belong in `_meta`.  `_meta` is about the format of the lock file itself (it's really schema details so that we know if this version of spack can read the file).

The `spack-commit` is a new field in the file with "real" data.  It's complicated to talk about this here b/c `spack.lock` is a file full of metadata about the build.  `_meta` is metadata about the file.

From experience: I think we should have a top-level `spack` dictionary to make this more extensible.  Already I know we will want:

```json
"spack": {
    "type": "git",
    "commit": "457daf4be65478a592102177662eb9dad52b8767"
}
```

and:

```json
"spack": {
    "type": "release",
    "version": "0.18.1"
}
```

lib/spack/spack/environment/environment.py
```diff
@@ -100,7 +101,7 @@ def default_manifest_yaml():
 valid_environment_name_re = r"^\w[\w-]*$"
 
 #: version of the lockfile format. Must increase monotonically.
-lockfile_format_version = 4
+lockfile_format_version = 5
```
tgamblin: I am not sure we need to increment this.  I wonder if we should have major and minor lock file versions, because the addition of the `spack` section doesn't actually prevent prior versions of Spack from reading this file.

Unfortunately, changing this to `"4.1"` would prevent prior spack versions from reading the lockfile version... so we should either just make a clean break and go with, say, `"5.0"` or keep it at 4 and support the spack section optionally.

tgamblin (changes requested):

lib/spack/spack/environment/__init__.py
tldahlgren: Okay.  I didn't realize that.  Thanks!

tldahlgren: Great!  Thanks for the clarification.

lib/spack/spack/environment/environment.py
tldahlgren: I like the latter.

lib/spack/spack/environment/__init__.py
tldahlgren: As discussed off-line earlier, since it is easy to grab (and can be useful from a human reading the file perspective), will store the version even when storing the commit.

tldahlgren:
Moved changes to *use* the commit/version to #33648 per off-line discussion related to v0.19.0 release goals.

alalazo:
Thanks @tldahlgren ! This would allow us to do some more active cleaning of recipes, provided a way to reproduce an old environment by checking out the corresponding Spack version first.

tldahlgren:
@tgamblin Can you re-review this to confirm I resolved the requested change? 

Patch coverage is at 100% and project is up 2.11%.

tgamblin (approved):

PR #32802, opened 2022-09-24
cgal: depends on for older versions

steo85it:
It seems that some recent change to `cgal` dependencies is incompatible with older versions. This solves the issue for @4.12, at least.
Not addressed here, but it would also seem that cgal@4.12 won't compile with option `+headers_only`.

### DIFF
--- a/var/spack/repos/builtin/packages/cgal/package.py
+++ b/var/spack/repos/builtin/packages/cgal/package.py
@@ -52,8 +52,8 @@ class Cgal(CMakePackage):
     depends_on("cmake@2.8.11:", type="build")
 
     # Essential Third Party Libraries
-    depends_on("boost+exception+math+random+container", when='@5.0:')
-    depends_on('boost+thread+system', when='@:5.0')
+    depends_on("boost+exception+math+random+container", when="@5.0:")
+    depends_on("boost+thread+system", when="@:5.0")
     depends_on("gmp")
     depends_on("mpfr")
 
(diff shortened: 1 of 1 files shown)


tldahlgren:
@spackbot fix style

tldahlgren:
@nicolaslg As a recent contributor of a "significant" change, would you care to comment on this PR?

steo85it:
@tldahlgren Thank you for checking this out, I wasn't aware of all the available tools! This should be a minor change to fix backward compatibility, but great if someone else can comment on it.

tldahlgren:
> @tldahlgren Thank you for checking this out, I wasn't aware of all the available tools! This should be a minor change to fix backward compatibility, but great if someone else can comment on it.

That's why I asked a prior contributor if they wanted to comment.

tldahlgren (approved):
LGTM

tldahlgren:
Will still give @nicolaslg a few days to comment before merging.

nicolaslg:
I do not encounter any issue with installing cgal@4.12 on my setup (linux-ubuntu22.04-icelake gcc-11.2.0), but I think the proposed modification about the `boost` dependency seems more in line with the `CGAL` install instructions (prior to v5) :
https://doc.cgal.org/4.14.3/Manual/installation.html#secessential3rdpartysoftware

That said, I am not really familiar with building CGAL :-)

tldahlgren:
@steo85it It is easier (fewer steps) for people to review your PRs if they use a non-`develop` branch: `git checkout -b <feature-branch>; git push --set-upstream <origin-or-other-remote-name> <feature-branch>`.

steo85it:
Ah, good to know, thanks! Will do next time (sorry, I don't push PRs to large projects very often).

PR #32804, opened 2022-09-24
py-sphinx: add v5.2.0

adamjstewart:
Successfully installs on macOS 12.6 (arm64) with Python 3.9.13 and Apple Clang 14.0.0.

https://www.sphinx-doc.org/en/master/changes.html

### DIFF
--- a/var/spack/repos/builtin/packages/py-flit-core/package.py
+++ b/var/spack/repos/builtin/packages/py-flit-core/package.py
@@ -13,6 +13,7 @@ class PyFlitCore(PythonPackage):
     pypi = "flit-core/flit_core-3.3.0.tar.gz"
     maintainers = ["takluyver"]
 
+    version("3.7.1", sha256="14955af340c43035dbfa96b5ee47407e377ee337f69e70f73064940d27d0a44f")
     version("3.6.0", sha256="5892962ab8b8ea945835b3a288fe9dd69316f1903d5288c3f5cafdcdd04756ad")
     version("3.5.1", sha256="3083720351a6cb00e0634a1ec0e26eae7b273174c3c6c03d5b597a14203b282e")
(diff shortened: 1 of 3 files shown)


tldahlgren (dismissed):
LGTM. Confirmed all three new sha256.

tldahlgren:
Will defer merged until @takluyver has a chance to comment.

var/spack/repos/builtin/packages/py-flit-core/package.py
```diff
@@ -23,10 +24,13 @@ class PyFlitCore(PythonPackage):
     version("3.0.0", sha256="a465052057e2d6d957e6850e9915245adedfc4fd0dd5737d0791bf3132417c2d")
     version("2.3.0", sha256="a50bcd8bf5785e3a7d95434244f30ba693e794c5204ac1ee908fc07c4acdbf80")
 
-    # Dependencies listed in flit_core/build_thyself.py
+    # pyproject.toml
+    depends_on("python@3.7:", when="@3.7:", type=("build", "run"))
```
takluyver: `flit_core` should still work on Python 3.6 - we have CI that aims to test this. If you've found that it doesn't, please let me know.

I try to be fairly conservative with `flit_core`, while sometimes `flit` (the CLI) requires a newer version of Python.

takluyver:
I'm guessing I was tagged for the flit-core part, so I haven't looked at the other changes in any detail. Other than my one comment above, this LGTM.

var/spack/repos/builtin/packages/py-flit-core/package.py
adamjstewart: Good catch, not sure where I got that from

tldahlgren (approved):

adamjstewart:
@spackbot run pipeline

PR #32805, opened 2022-09-25
hdf5-vol-async: fix path needed for h5bench

jeanbez:
Fix the plugin path environment variable needed when using the VOL connector.


hyoklee (dismissed):

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -27,7 +27,7 @@ class Hdf5VolAsync(CMakePackage):
     depends_on("hdf5@1.13: +mpi +threadsafe")
 
     def setup_run_environment(self, env):
-        env.set("HDF5_PLUGIN_PATH", self.spec.prefix)
+        env.set("HDF5_PLUGIN_PATH", join_path(self.spec.prefix, "lib"))
```
tldahlgren: Looks good.

Did you know you could simply do the following instead?

```suggestion
        env.set("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
```

tldahlgren (dismissed):
LGTM.  Let me know if you want to shorten the chance as suggested or merge as-is.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
hyoklee: @jeanbez would you please update your PR following @tldahlgren 's suggestion?

jeanbez: @hyoklee sure!

tldahlgren (approved):

hyoklee (approved):

PR #32816, opened 2022-09-26
shell.py: fix undefined variable in csh deactivate

### LINKED ISSUE #27027: Environment activation/deactivation issue with [t]csh
As for your initial issue in https://github.com/spack/spack/pull/25409#issuecomment-953132990:

```
$ spack env activate a
SPACK_OLD_PROMPT: Undefined variable.
```

There seems to be a problem with our generated csh commands where we don't enclose them in ``` `` ``` and they therefore are evaluated, for example:

```
if ( $?SPACK_OLD_PROMPT ) set prompt="$SPACK_OLD_PROMPT" && unsetenv SPACK_OLD_PROMPT ;
```

Appears in `spack env activate --csh <name-of-env>`, but this fails in my env when `$SPACK_OLD_PROMPT` isn't defined because it tries to evaluate the variable inside the statement (even if the statement itself isn't evaluated). So the following appears to get desirable behavior:

```
if ( $?SPACK_OLD_PROMPT ) `set prompt="$SPACK_OLD_PROMPT" && unsetenv SPACK_OLD_PROMPT` ;
```

I likewise had to make a similar edit to `pathadd.csh` to get `setup-env.csh` to work:

```
diff --git a/share/spack/csh/pathadd.csh b/share/spack/csh/pathadd.csh
index ec394e8990..7f3424518a 100644
--- a/share/spack/csh/pathadd.csh
+++ b/share/spack/csh/pathadd.csh
@@ -22,7 +22,7 @@ endif
 set _pa_old_value = ""
 eval set _pa_set = '$?'$_pa_varname
 if ($_pa_set == 1) then
-    eval set _pa_old_value='$'$_pa_varname
+    eval `set _pa_old_value = '$'$_pa_varname`
 endif
 
 # Do the actual prepending here, if it is a dir and not already in the path
```

_Originally posted by @scheibelp in https://github.com/spack/spack/issues/25409#issuecomment-953332271_

alexhornburg-xlnx:
This commit fixes #27027.

The root cause of the issue is that the `SPACK_OLD_PROMPT` variable was evaluated in string interpolation regardless of whether the guard condition above evaluates to true or false. This commit uses the `eval` keyword to defer evaluation until the command is executed.

Tested as follows:
```
~/spack/ # source share/spack/setup-env.csh 
~/spack/ # spack env create myenv
~/spack/ # spack env activate myenv
~/spack/ # spack env deactivate
~/spack/ # spack env activate -p myenv
[myenv] ~/spack/ # spack env deactivate
~/spack/ #
```
Note that `SPACK_OLD_PROMPT: Undefined variable.` no longer appears.

Please also note that the suggested fix in #27027 of using back-ticks to enclose the two commands does not work for the case where the prompt flag (`-p`) is specified during environment activation.

### DIFF
--- a/lib/spack/spack/environment/shell.py
+++ b/lib/spack/spack/environment/shell.py
@@ -65,8 +65,8 @@ def deactivate_header(shell):
     if shell == "csh":
         cmds += "unsetenv SPACK_ENV;\n"
         cmds += "if ( $?SPACK_OLD_PROMPT ) "
-        cmds += 'set prompt="$SPACK_OLD_PROMPT" && '
-        cmds += "unsetenv SPACK_OLD_PROMPT;\n"
+        cmds += '    eval \'set prompt="$SPACK_OLD_PROMPT" &&'
+        cmds += "          unsetenv SPACK_OLD_PROMPT';\n"
         cmds += "unalias despacktivate;\n"
     elif shell == "fish":
         cmds += "set -e SPACK_ENV;\n"


alexhornburg-xlnx:
Hi, I rebased the PR on top of latest develop branch today. Before the rebase all checks were passing. Now the `codecov/project` check is no longer passing. When I look at the details I can see the modified lines of my patch are hit both before and after the patch. So I do not think this failure is related to the patch.

@tldahlgren @haampie since both of you have commented on #27027 could I please ask that one of you review this PR? Thanks

tldahlgren (approved):
This works for me.

PR #32819, opened 2022-09-26
amdblis: symlink libblis-mt to libblis

zzzoom:
`amdblis threads=openmp` installs `libblis-mt.so` instead of `libblis.so`.
CMake's FindBLAS currently looks for `libblis.so` only (https://gitlab.kitware.com/cmake/cmake/-/issues/23605) so I'm adding a symlink for compatibility.


tldahlgren (approved):
LGTM.  Deferring merge to allow @amd-toolchain-support to comment.

bernhardkaindl (approved):
@amd-toolchain-support didn't comment since Sept 26, merging after giving two months review time

PR #32558, opened 2022-09-07
superlu-dist: add 8.1.1, 8.1.0 and 8.0.0 versions

balos1:

PR #32825, opened 2022-09-26
Temporarily pin superlu-dist version to 7.2.0

tldahlgren:
This is a temporary stack fix to allow #32558 to pass pipeline checks.  There are 2 commits: one constrains the version for `trilinos` the other (newer) constrains for all packages since multiple packages depend on `superlu-dist`.


eugeneswalker:
Shouldn't these constraints be embedded in `trilinos` package.py? Wouldn't that allow https://github.com/spack/spack/pull/32558 to proceed while simultaneously sparing downstream users of the headache of having trilinos try to use a version of superlu-dist that it isn't compatible with? What do you think? Looks like @balos1 suggesting the same here:
* https://github.com/spack/spack/pull/32558#issuecomment-1258559352

CC @jwillenbring

tldahlgren:
> Shouldn't these constraints be embedded in `trilinos` package.py? Wouldn't that allow #32558 to proceed while simultaneously sparing downstream users of the headache of having trilinos try to use a version of superlu-dist that it isn't compatible with? What do you think? Looks like @balos1 suggesting the same here:
> 
> * [superlu-dist: add 8.1.0 and 8.0.0 versions #32558 (comment)](https://github.com/spack/spack/pull/32558#issuecomment-1258559352)
> 
> CC @jwillenbring

Yes, you are correct. Will close this.

PR #32828, opened 2022-09-26
geant4-vmc: unset MACOSX_DEPLOYMENT_TARGET

aphecetche:
I've noticed lately (possibly after the upgrade to Xcode 14 ?) that I need to unset this env variable otherwise I get failed builds (in a number of packages) with error : 
```
error: invalid version number in 'MACOSX_DEPLOYMENT_TARGET=12.0'
```

I'm not so sure the fix is a really good one, but pragmatically it works :wink: 
maybe @adamjstewart can comment further on the validity / limits of the fix ?

### DIFF
--- a/var/spack/repos/builtin/packages/geant4-vmc/package.py
+++ b/var/spack/repos/builtin/packages/geant4-vmc/package.py
@@ -36,3 +36,7 @@ class Geant4Vmc(CMakePackage):
     depends_on("cmake@3.3:", type="build")
     depends_on("geant4")
     depends_on("vmc")
+
+    def setup_build_environment(self, env):
+        if self.spec.satisfies("platform=darwin"):
+            env.unset("MACOSX_DEPLOYMENT_TARGET")


adamjstewart:
@sethrj any thoughts on this?

See #31811 to links to a bunch of issues/PRs relating to this env var and the motivation behind the hacks that we use for it.

adamjstewart:
> in a number of packages

Which other packages?

sethrj:
This is likely coming from ROOT's old version of Clang/llvm: probably a transient dependency through vmc? But there is code in the root package to remove said variable. We should debug why that is not working...

aphecetche:
> > in a number of packages
> 
> Which other packages?

```
> rg MACOSX_DEPLOYMENT_TARGET | grep unset
vmc/package.py:            env.unset("MACOSX_DEPLOYMENT_TARGET")
geant4-vmc/package.py:            env.unset("MACOSX_DEPLOYMENT_TARGET")
root/package.py:            env.unset("MACOSX_DEPLOYMENT_TARGET")
root/package.py:            env.unset("MACOSX_DEPLOYMENT_TARGET")
```

Note indeed the second `unset` in the root package is in `setup_dependent_build_environment` which should be picked by e.g. vmc and geant4-vmc (right ?) but are not as far as I can tell.

aphecetche:
I assume the `unset` in `setup_dependent_build_environment` of the `root` recipe is not effective as other dependencies of the `geant4-vmc` package do *not* make that `unset` ?
If that's the case then we have no other choice but repeating the `unset` in the `setup_build_environment` of `geant4-vmc`, right ? 
(or find a way to get `rootcling` to play nicely with the `MACOSX_DEPLOYMENT_TARGET=12.0` but I guess that's way less easy ;-) )

adamjstewart:
`setup_dependent_build_environment` is designed for this purpose, if it's not working then that's a bug. @alalazo did you design this feature, or was it someone else?

aphecetche:
so, what should we do with this one ?

adamjstewart (approved):
I'm fine with this hack, although we should look into why `setup_dependent_build_environment` no longer works. Maybe @becker33 or @cosmicexplorer would be interested in digging into this.

adamjstewart:
Perhaps the reason that `setup_dependent_build_environment` isn't working for this variable is because `setup_dependent_build_environment` is run before `setup_platform_environment`? It should also be noted that the `AutotoolsPackage` base class also sets this variable.

I say this because I used `setup_dependent_build_environment` in another package for a different variable and it seems to work fine.

PR #32832, opened 2022-09-26
Improve lexing error

white238:
With the following badly formed line in a package:

```
dep = "foo"
depends_on("{0}+debug".format(dep), when="+{0}+debug")
```

Notice the lack of the `.format(dep)` at the end of the when parameter.

The correct line should have been the following but it was hard to track down due to the amount of changes I made.

```
depends_on("{0}+debug".format(dep), when="+{0}+debug".format(dep))
```

You get the following error message when you run `spack spec ...`:

```
==> Error: Invalid character
```

This improves the error message to the following:

```
==> Error: Invalid character, '{', in '+{0}+debug' at index 1
```

### DIFF
--- a/lib/spack/spack/parse.py
+++ b/lib/spack/spack/parse.py
@@ -78,7 +78,9 @@ class Lexer(object):
                 break
 
         if remainder and not remainder_used:
-            raise LexError("Invalid character", word, word.index(remainder))
+            msg = "Invalid character, '{0}',".format(remainder[0])
+            msg += " in '{0}' at index {1}".format(word, word.index(remainder))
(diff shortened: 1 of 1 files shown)


tldahlgren (approved):
Love the extra detail!  (Manually tested with other commands.)

white238:
On the other hand, this could have just been talking about my moral character.

tldahlgren:
> On the other hand, this could have just been talking about my moral character.

😆 Perhaps I need to make sure I'm more neutral in my comments.  I usually am but I do prefer more detail in error messages.

PR #32833, opened 2022-09-26
zziplib: adding some missing build deps

snehring:
The zip and unzip utils are necessary for building part of this.


var/spack/repos/builtin/packages/zziplib/package.py
```diff
@@ -30,6 +30,8 @@ class Zziplib(AutotoolsPackage):
     # see zzip/CMakeLists.txt
     depends_on("coreutils", type="build", when="@0.13.71:")
     depends_on("pkgconfig", type="build", when="@0.13.71:")
+    depends_on("zip", type="build", when="@0.13.71:")
+    depends_on("unzip", type="build", when="@0.13.17:")
```
tldahlgren: Is this supposed to be the same version?

```suggestion
    depends_on("unzip", type="build", when="@0.13.71:")
```

snehring: Yes, typo'd.

tldahlgren (approved):
LGTM

PR #32835, opened 2022-09-27
py-numpy: add headers attribute

adamjstewart:
Copy-n-pasted this from the `PythonPackage` base class. For some reason, it doesn't work if it's in the base class, see https://github.com/spack/spack/pull/28527#issuecomment-1172046128

### DIFF
--- a/var/spack/repos/builtin/packages/py-numpy/package.py
+++ b/var/spack/repos/builtin/packages/py-numpy/package.py
@@ -6,6 +6,7 @@
 import platform
 import subprocess
 
+from spack.error import NoHeadersError
 from spack.package import *
 
 
@@ -372,6 +373,21 @@ class PyNumpy(PythonPackage):
             lapack = "lapack"
 
         env.set("NPY_LAPACK_ORDER", lapack)
+
+    @property
+    def headers(self):
+        """Discover header files in platlib."""
+
+        # Headers may be in either location
+        include = join_path(self.prefix, self.spec["python"].package.include)
+        platlib = join_path(self.prefix, self.spec["python"].package.platlib)
+        headers = find_all_headers(include) + find_all_headers(platlib)
+
+        if headers:
+            return headers
+
+        msg = "Unable to locate {} headers in {} or {}"
+        raise NoHeadersError(msg.format(self.spec.name, include, platlib))
 
     @run_after("install")
     @on_package_attributes(run_tests=True)
--- a/var/spack/repos/builtin/packages/py-torch/package.py
+++ b/var/spack/repos/builtin/packages/py-torch/package.py
@@ -6,6 +6,7 @@
 import os
 import sys
 
+from spack.error import NoHeadersError, NoLibrariesError
 from spack.operating_systems.mac_os import macos_version
 from spack.package import *
 


rgommers (commented):
`numpy` doesn't have any Python dependencies, so there should be no need to look for any headers in `platlib`. We should only need `Python.h` I think. Does this actually fix something that is not working right now? The linked comment seems to be about `py-torch`. I thought the other problem was in locating the numpy headers themselves. In that case, shouldn't this patch like in the package that needs those numpy headers?

adamjstewart:
The `headers` property of a package defines where to look for headers for that package. So with this PR, `spec['py-numpy'].headers` will tell you where the headers can be found. We _could_ define this in the package that needs to find the headers, but if there are more than one, we would need to copy-n-paste it every time. So the idea is to define it only once and let packages reuse it. It was _supposed_ to work by defining it in the base class, but it isn't working and I don't know why.

Btw, `py-petsc4py` is the package that has trouble finding these headers without adding them to `CPATH` @balay

rgommers:
Ah okay, that is helpful context info. In that case, recursively searching `platlib` seems watertight. The numpy headers should never be landing in `<prefix>/include`.

adamjstewart:
@spackbot run pipeline

balay:
> Btw, py-petsc4py is the package that has trouble finding these headers without adding them to CPATH @balay

Looks like petsc4py is querying numpy directly to get this info
```python
>>> import numpy
>>> numpy.get_include()
'/usr/lib64/python3.10/site-packages/numpy/core/include'
```
So assuming petsc4py will be unaffected by this change..

cosmicexplorer:
I'm seeing this pattern used in the `py-gpaw`, `vapor`, and `precice` packages to get a particular `numpy/core/include` subdirectory of `platlib`, which in `precice`'s case at least is used to populate a specific `-DNumPy_INCLUDE_DIR` CMake var: https://github.com/spack/spack/blob/977c89cee19b0c589f3ffba3f600892789acebf5/var/spack/repos/builtin/packages/precice/package.py#L153-L169

Outside of `py-matplotlib` and `py-torchvision`, I wasn't able to see any packages specifically requesting `py-numpy`'s `.headers` (which I guess makes sense if #28527 was just merged). The reason I mention this is: if multiple packages are trying to get at that *precise* subdirectory of numpy's includes, would it make sense to add that particular platlib subdirectory as a property of the `py-numpy` package as well here?

cosmicexplorer:
@adamjstewart: I don't fully understand what function that `inspect.getmodule(self).include` (added in #28527) is expected to invoke, and by inspecting the callstack, `inspect.getmodule(self).include` seemed to be calling `_headers_default_handler()` in `spec.py` (which I still do not understand whatsoever), which is why you saw both the default and the `build_systems/python.py` method being called in order.

However, I was regardless able to make `py-scipy` find the headers for `py-numpy` by taking the exact code you used in *this* PR, and instead moving it into `build_systems/python.py`'s `def headers(self)` method. The diff is here https://github.com/adamjstewart/spack/compare/packages/py-numpy...cosmicexplorer:spack:packages/py-numpy, but it's just one commit that does exactly that. Please let me know if this also works for you.

adamjstewart:
> would it make sense to add that particular platlib subdirectory as a property of the `py-numpy` package as well here?

That's the point of using `spec['py-numpy'].headers.directories[0]`, we should replace the hacks in those package with this if we can get it working.

> I don't fully understand what function that `inspect.getmodule(self).include` (added in https://github.com/spack/spack/pull/28527) is expected to invoke

It's supposed to access `spec['python'].package.include`, but I guess it's actually named `python_include`, not `include`. Let me try that. I'll try your patch as well.

adamjstewart:
Yep, that seems to work, nice find! I'll open a separate PR for this.

Issue #32823, opened 2022-09-26
Configurable named environment location


douglasjacobsen:
### Summary

It would be nice if there was a config option that let you set the location named environments can live in. Maybe this could even be a list of locations.

### Rationale

Sometimes it's useful to have environments live outside of spack's hard-coded environment location.

### General information

- [X] I have run `spack --version` and reported the version of Spack
- [X] I have searched the issues of this repo and believe this is not a duplicate

PR #32836, opened 2022-09-27
Allow users to specify root env dir

psakievich:
Environments managed by spack have some advantages over anonymous Environments but they are tucked away inside spack's directory tree. This PR gives users the ability to specify where the environments should live.

See #32823


lib/spack/spack/environment/environment.py
```diff
@@ -63,7 +63,10 @@
 
 
 #: path where environments are stored in the spack tree
-env_path = os.path.join(spack.paths.var_path, "environments")
+# circular dependency to use canonicalize_path but we want to ignore an active env anyways
+# env_path = spack.config.get(os.path.expandvars(os.path.expanduser("config:environments_root")))
+def env_path():
```
psakievich: Does anyone know of a way to do a module level property?  That seems like it would be a better replacement than this.

lib/spack/spack/environment/environment.py
```diff
@@ -63,7 +63,10 @@
...
+    return spack.util.path.canonicalize_path(spack.config.get("config:environments_root"), False)
```
becker33: The code default should be the same as the config default (i.e. `os.path.join(spack.paths.avar_path, "environments")`). We try to make the default config be only advisory anywhere we can.

I also think this line will be easier to read if we make the second argument explicit,

```suggestion
    return spack.util.path.canonicalize_path(
        spack.config.get("config:environments_root", os.path.join(spack.paths.var_path, "environments")),
        allow_env=False
    )
```

becker33 (changes requested):
This is a great idea

lib/spack/spack/environment/environment.py
```diff
@@ -77,6 +77,14 @@
 #: Name of the directory where environments store repos, logs, views
 env_subdir_name = ".spack-env"
 
+def env_root_path():
+    """Override default root path if the user specified it"""
+    config_path = spack.config.get("config:environments_root", default = None)
+    if config_path:
+        return config_path
+    else:
+        return env_path
```
becker33:
```suggestion
    return spack.config.get("config:environments_root", default=env_path)
```

psakievich:
Closes #32823

lib/spack/docs/environments.rst
```diff
@@ -72,7 +72,8 @@ Spack then creates the directory ``var/spack/environments/myenv``.
 
 .. note::
 
-   All named environments are stored in the ``var/spack/environments`` folder.
+   All named environments by default are stored in the ``var/spack/environments`` folder.
+   This location can be changed by setting the ``envionments_root`` variable in ``config.yaml``.
```
tldahlgren:
```suggestion
   This location can be changed by setting the ``environments_root`` variable in ``config.yaml``.
```

lib/spack/spack/test/config.py
```diff
@@ -1005,6 +1008,7 @@ def test_good_env_yaml(tmpdir):
     config:
         verify_ssl: False
         dirty: False
+        environtments_dir: ~/my/env/location
```
tldahlgren:
```suggestion
        environments_root: ~/my/env/location
```

share/spack/qa/setup.sh
```diff
@@ -11,7 +11,12 @@
 #
 
 QA_DIR="$(dirname ${BASH_SOURCE[0]})"
-export SPACK_ROOT=$(realpath "$QA_DIR/../../..")
+if realpath $QA_DIR >& /dev/null ; then
+    REALPATH=realpath
+else
+    REALPATH="readlink -f --"
+fi
+export SPACK_ROOT=$($REALPATH "$QA_DIR/../../..")
```
tldahlgren: Are the changes to this file required for the purposes of this PR?  Or is this more suited to a separate PR?

lib/spack/spack/environment/environment.py
```diff
@@ -62,7 +62,7 @@
 _active_environment = None
 
 
-#: path where environments are stored in the spack tree
+#: default path where environments are stored in the spack tree
```
tldahlgren, line 67: I wonder if the variable name needs to be changed, perhaps `default_env_path`, given it would no longer be the only option?

lib/spack/spack/util/path.py
```diff
@@ -235,7 +235,7 @@ def convert_to_platform_path(path):
     return format_os_path(path, mode=Path.platform_path)
 
 
-def substitute_config_variables(path):
+def substitute_config_variables(path, allow_env=True):
```
tldahlgren: When is (or would) `allow_env` ever be `False`?

tldahlgren (commented):
I suggest a few minor typo corrections and have a few questions.

@tgamblin @becker33 What would (or should) happen if a user already has environments defined in the default path then updates their configuration to point elsewhere?

Is it worth considering supporting a list of environment paths?

tldahlgren (changes requested):
Switching from comment to request changes due to the typos needing fixed.

share/spack/qa/setup.sh
psakievich: @tldahlgren this was pushed directly by @trws to try and debug the failing CI check which I haven't been able to understand.

lib/spack/spack/util/path.py
psakievich: In this PR it is set to `False` to allow path expansion/ canonicalization of `config:environments_root` to bypass the circular dependency.  See `environment.py:L84`

lib/spack/spack/environment/environment.py
psakievich: sure that makes sense to me

psakievich:
@tldahlgren I thought about having a list of directories.  This is a good question. It complicated things a bit which is why I didn't pursue it. 

One thing I'm not sure how to handle is where we add the new environments to.  Thinking further, if we treat this like a stack then we could just add environments to the directory at the top of the stack.  Another option could be to only allow two entries: the default directory and allow for a single option from the config.  I think I like the second option more. I'm not sure the utility of expanding beyond a single user location and the default location.

In the current form the already created environments in the default directory would become anonymous environments. Speaking for myself I typically only use one or the other (registered vs anonymous) in a given spack instance but that may not be reflective of the broader community.

psakievich:
Hmm another potential issue for this is if the environments have the same name in both locations

psakievich:
@becker33 @tgamblin can you weigh in on @tldahlgren's comments? I'm looking into multiple paths for environment roots. It's not trivial but doable if it's a must for this feature to go in.

lib/spack/spack/environment/environment.py
```diff
@@ -62,8 +62,13 @@
...
 
-#: path where environments are stored in the spack tree
-env_path = os.path.join(spack.paths.var_path, "environments")
+#: default path where environments are stored in the spack tree
+default_env_path = os.path.join(spack.paths.var_path, "environments")
+
+# ensure that this base directory exists since we require the root directory to be created
+# in our checks
+if not os.path.isdir(default_env_path):
+    os.mkdir(default_env_path)
```
becker33: Why do we need the directory to exist if Spack isn't configured to use it?

lib/spack/spack/util/path.py
```diff
@@ -270,9 +271,9 @@ def repl(match):
     return re.sub(r"(\$\w+\b|\$\{\w+\})", repl, path)
 
 
-def substitute_path_variables(path):
+def substitute_path_variables(path, allow_env=True):
```
becker33: What happens if you activate an environment that changes the location of the default environment list? I have a feeling Spack will explode if you do that, we should add some guardrails for it.

becker33 (changes requested):

lib/spack/spack/environment/environment.py
psakievich: Ugh it has been a while. IIRC this is what was causing the `shell` tests in CI to fail. Spack does a check in the root environment directory to see if any environments are installed and if the directory is missing it hangs/fails.  I opted to always create to see if it fixed the failing CI tests.  Not exactly certain how deep this rabbit hole goes though.

lib/spack/spack/util/path.py
psakievich: Agreed.

lib/spack/spack/test/config.py
```diff
@@ -1431,3 +1435,54 @@ def test_config_file_read_invalid_yaml(tmpdir, mutable_empty_config):
...
+        )
+    with tmpdir.as_cwd():
+        with pytest.raises(ev.SpackEnvironmentError):
+            # Why does this raise?
+            e = ev.Environment(tmpdir.strpath)
+            ev.activate(e)
+        with pytest.raises(ev.SpackEnvironmentError):
+            # When this does not?
+            env("create", "-d", tmpdir.strpath)
+            env("activate", "-d", tmpdir.strpath)
```
psakievich: 😵‍💫

share/spack/qa/setup.sh
psakievich: I reverted this commit.

lib/spack/spack/environment/environment.py
psakievich: Addressed this as requested. Now we create the default location if is the `env_root_path()`

lib/spack/spack/test/config.py
psakievich: So after digging into this it looks like the `SpackCommand('env')` is really doing shell calls so my test is not actually getting executed in this shell for pytest to catch.

psakievich:
Bump for @tldahlgren and @becker33 for another review.  @tldahlgren WRT to the existing environments, I chatted with @becker33 about this on slack and his take was if users want  pre-existing environments when they change the location of the root directory then they should move them themselves. 

I began prototyping what it would take to respect multiple locations for the registered environments, and it is a non-trivial change.  The registered environments code is built on the fundamental assumption of a single directory so I think it would be best to tackle that in a separate PR if it becomes a requested feature.

becker33:
> I began prototyping what it would take to respect multiple locations for the registered environments, and it is a non-trivial change. The registered environments code is built on the fundamental assumption of a single directory so I think it would be best to tackle that in a separate PR if it becomes a requested feature.

In the long run, I think @tgamblin has aims of having an environment database, so Spack can track all environments it knows about wherever they may be stored. Then there won't be such a difference between tracked vs untracked environments

psakievich:
@becker33 that is exactly what I was thinking we'd need to do it right. Feels like a win to be tracking along the same thought process as @tgamblin :tada:

lib/spack/spack/environment/environment.py
```diff
@@ -1016,7 +1048,9 @@ def env_file_config_scope(self):
 
     def config_scopes(self):
         """A list of all configuration scopes for this environment."""
-        return self.included_config_scopes() + [self.env_file_config_scope()]
+        return check_disallowed_env_config_mods(
+            self.included_config_scopes() + [self.env_file_config_scope()]
```
psakievich, line 1062: This is taking in more scopes than I thought when I use it. The `site` scope seems to be coming into play and not just the `spack.yaml` and `includes` scopes.  Is there a way for me to filter the scopes?  @becker33 any ideas on what to do here?

psakievich: NVM It was due to the `spack.config.override`.

lib/spack/spack/environment/environment.py
psakievich: Okay really addressed now.  The directory only gets created if it is used.

lib/spack/spack/test/config.py
```diff
@@ -1458,3 +1462,31 @@ def test_config_file_read_invalid_yaml(tmpdir, mutable_empty_config):
...
+
+def test_environment_errors_if_root_missing(mutable_config, tmpdir):
+    spack.config.set("config:environments_root", str(tmpdir.join("envs")))
+    env_dir = spack.config.get("config:environments_root")
+
+    assert tmpdir.strpath in env_dir
+    assert not os.path.isdir(env_dir)
+
+    with pytest.raises(ev.SpackEnvironmentError, match=env_dir):
+        env("create", "test")
```
becker33: This should all be in `lib/spack/spack/test/cmd/env.py`

lib/spack/spack/test/conftest.py
```diff
@@ -1529,11 +1529,11 @@ def get_rev():
 @pytest.fixture()
 def mutable_mock_env_path(tmpdir_factory):
```
becker33: This fixture now needs to guarantee that `environments_root` is overridden if set by the user. Using the `mutable_config` fixture is probably sufficient.

becker33 (dismissed):

lib/spack/spack/environment/environment.py
```diff
@@ -240,6 +260,8 @@ def read(name):
 
 def create(name, init_file=None, with_view=None, keep_relative=False):
     """Create a named environment in Spack."""
+    if not os.path.isdir(env_root_path()):
+        os.mkdir(env_root_path())
```
psakievich:
```suggestion
        llnl.util.filesystem.mkdirp(env_root_path())
```

lib/spack/spack/test/ci.py
```diff
@@ -421,6 +421,7 @@ def make_yaml_obj(use_artifact_buildcache, optimize, use_dependencies):
 def test_get_spec_filter_list(mutable_mock_env_path, config, mutable_mock_repo):
     """Test that given an active environment and list of touched pkgs,
     we get the right list of possibly-changed env specs"""
+    spack.config.set("config:environments_root", str(tmpdir.join("local_envs")))
```
becker33: This test doesn't include `tmpdir`, I'm surprised it would even run like this.

psakievich: It doesn't. This was a hacky attempt to get it to work based on what I did in a different test. For whatever reason the `mutable_mock_env` fixture's path is not correct for this test.

psakievich:
@spackbot fix style

psakievich:
@tldahlgren and @becker33 this is ready finally.  After #34758 some of the logic regarding environment paths and avoiding circular imports was cleaned up.  Most of the hang up on this was due to tests using the same `mutable_mock_env_path` in parallel.  I addressed this by making the fixture use a `function` scope and then cleaning up two more sets of tests.

bin/spack-tmpconfig
```diff
@@ -72,6 +72,7 @@ config:
     root: $TMP_DIR/install
   misc_cache: $$user_cache_path/cache
   source_cache: $$user_cache_path/source
+  environments_root: $TMP_DIR/envs
```
becker33, line 75: I'm having trouble tracking why this change is necessary.

becker33 (commented):
Besides the one inline comment, everything looks good to me. I agree with your assessment on slack that it's confusing that this is coming up for this PR and not previously. My best guess without fully diving in to debug it is that previously things "just happened" to work because environments would coincidentally wind up in the same place, even though they weren't being mocked in a way that was thread safe.

bin/spack-tmpconfig
psakievich: This was a suggestion from @trws to sandbox parallel execution. I don't fully understand it either.  @trws perhaps you could comment?

becker33 (dismissed):
I'm sorry this became such an unholy pain to get ready to merge

bin/spack-tmpconfig
trws: That script is used to separate the unit tests when run in parallel.  It ensures that each test runner gets its own separate spack environment so they don't tromp each other.  If the environments_root defaults to being under user_cache_path or something then it's unnecessary, but if it's in the spack_root by default or something like that then the script needs to put it somewhere else.

lib/spack/docs/environments.rst
```diff
@@ -72,7 +72,8 @@ Spack then creates the directory ``var/spack/environments/myenv``.
 
 .. note::
 
-   All named environments are stored in the ``var/spack/environments`` folder.
+   All named environments by default are stored in the ``var/spack/environments`` folder.
```
tldahlgren: Since you're changing this, do we want this to be `managed` (not `named`) environments?

(https://spack-tutorial.readthedocs.io/en/latest/tutorial_environments.html#managed-versus-independent-environments).

lib/spack/spack/environment/environment.py
```diff
@@ -252,9 +272,11 @@ def read(name):
 
 def create(name, init_file=None, with_view=None, keep_relative=False):
     """Create a named environment in Spack."""
+    if not os.path.isdir(env_root_path()):
+        fs.mkdirp(env_root_path())
     validate_env_name(name)
     if exists(name):
-        raise SpackEnvironmentError("'%s': environment already exists" % name)
+        raise SpackEnvironmentError("'%s': environment already exists at %s" % (name, root(name)))
```
tldahlgren, line 279: ❤️

lib/spack/spack/test/cmd/concretize.py
```diff
@@ -9,8 +9,7 @@
 import spack.environment as ev
 from spack.main import SpackCommand
 
-# everything here uses the mock_env_path
-pytestmark = pytest.mark.usefixtures("mutable_mock_env_path", "config", "mutable_mock_repo")
+pytestmark = pytest.mark.usefixtures("config", "mutable_mock_repo")
```
tldahlgren, line 12: Why was `mutable_mock_env_path` removed from here and added to all of the tests?

bin/spack-tmpconfig
becker33: Yeah @psakievich and I came to the conclusion that this probably isn't *technically* necessary since all of the environment tests use a mock environment root anyway, but that we should keep it to ensure future tests are separated for parallel execution.

tldahlgren (commented):
LGTM.  Just a couple of questions.

lib/spack/spack/environment/environment.py
tldahlgren: Love the extra detail to aid debugging.

lib/spack/spack/test/cmd/concretize.py
becker33: Because the tests fail in parallel otherwise -- I haven't followed up in the documentation for pytest, but it looks like function-scoped fixtures are not reinstantiated for each function properly when applied using this mechanism.

tldahlgren: GTK

tldahlgren (dismissed):
Would prefer the 'named' -> 'managed' comment change but won't require the change.

lib/spack/docs/environments.rst
psakievich:
```suggestion
   All managed environments by default are stored in the ``var/spack/environments`` folder.
```

psakievich: Done!

tldahlgren (dismissed):
LGTM

becker33 (approved):

tldahlgren (approved):
LGTM.  Thanks.

bin/spack-tmpconfig
trws: Sounds right to me.  The script is meant to be a safety net and a workaround for all the tests that haven't been converted yet. It's also oddly useful as a way to do one-off testing without messing up a current environment just on the command line, so having it there helps with that kind of use.

PR #32838, opened 2022-09-27
Add patch to link pthread library for llvm 15

msimberg:
This fixes exactly the same thing as https://github.com/spack/spack/pull/31661 but for llvm 15.

I have not yet submitted the patch upstream but will do so.


trws:
Does the old patch no longer work for 15 such that we need another one?

msimberg:
> Does the old patch no longer work for 15 such that we need another one?

Exactly. The files are almost the same but they've changed enough that the patch doesn't apply anymore (also commented here for llvm@main at the time: https://github.com/spack/spack/pull/31661#issuecomment-1213020372).

trws (approved):

PR #32839, opened 2022-09-27
Fix mpich build in Gitlab pipelines

alalazo:
The current hypothesis is that a system failure has been recorded as a broken spec on `develop`

### DIFF
--- a/var/spack/repos/builtin/packages/mpich/package.py
+++ b/var/spack/repos/builtin/packages/mpich/package.py
@@ -360,8 +360,8 @@ with '-Wl,-commons,use_dylibs' and without
         return results
 
     def setup_build_environment(self, env):
-        env.unset("F90")
         env.unset("F90FLAGS")
+        env.unset("F90")
 
         # https://bugzilla.redhat.com/show_bug.cgi?id=1795817
         if self.spec.satisfies("%gcc@10:"):


alalazo:
This pipeline failed https://gitlab.spack.io/spack/spack/-/pipelines/176936 due to system failures (see https://gitlab.spack.io/spack/spack/-/jobs/3386390). Retrying the pipeline.

alalazo:
@spackbot run pipeline

var/spack/repos/builtin/packages/mpich/package.py
```diff
@@ -360,8 +360,8 @@ def is_disabled(text):
         return results
 
     def setup_build_environment(self, env):
-        env.unset("F90")
         env.unset("F90FLAGS")
+        env.unset("F90")
```
tldahlgren, line 364: Why would the order of these statements matter?

tldahlgren: Or is this a PR just for testing?

alalazo: This was to check if the error in `develop` was a false positive, by changing the hash and triggering a rebuild in a PR. See https://spackpm.slack.com/archives/C02N33GM28H/p1664280616596179

PR #32853, opened 2022-09-27
tcl module template: automatically unload automatically loaded modules

skosukhin:
I think I could simply provide Spack with my own template but I decided to share this.

Currently, a tcl module `foo/1.2.3` that is configured to autoload module `bar/4.5.6` contains the following lines:
```tcl
if { [ module-info mode load ] && ![ is-loaded bar/4.5.6 ] } {
    module load bar/4.5.6
}
```
The annoying thing is that if `bar` is loaded automatically when loading `foo` it does not get unloaded automatically when unloading `foo`:
```console
$ module list
No Modulefiles Currently Loaded.
$ module load foo/1.2.3
Loading foo/1.2.3
  Loading requirement: bar/4.5.6
$ module unload foo/1.2.3
$ module list
Currently Loaded Modulefiles:
 1) bar/4.5.6
```

This PR changes the template so that `foo/1.2.3` now contains the following (the single quotes around `bar/4.5.6` are important):
```tcl
if { ![ is-loaded 'bar/4.5.6' ] } {
    module load bar/4.5.6
}
```
And we get the following behaviour:
```console
$ module list
No Modulefiles Currently Loaded.
$ module load foo/1.2.3
Loading foo/1.2.3
  Loading requirement: bar/4.5.6
$ module unload foo/1.2.3
Unloading foo/1.2.3
  Unloading useless requirement: bar/4.5.6
$ module list
No Modulefiles Currently Loaded.
```
Also note that if `bar` is loaded manually, it does not get unloaded automatically:
```console
$ module list
No Modulefiles Currently Loaded.
$ module load bar/4.5.6
$ module load foo/1.2.3
$ module unload foo/1.2.3
$ module list
Currently Loaded Modulefiles:
 1) bar/4.5.6
```

I don't understand the reason for having `[ module-info mode load ]` introduced in d163be6 but if it wasn't for it, the only change of this PR would be adding the single quotes.

### DIFF
--- a/lib/spack/spack/test/modules/tcl.py
+++ b/lib/spack/spack/test/modules/tcl.py
@@ -113,7 +113,7 @@ class TestTcl(object):
         assert len([x for x in content if x.startswith("prepend-path CMAKE_PREFIX_PATH")]) == 0
         assert len([x for x in content if 'setenv FOO "foo"' in x]) == 0
         assert len([x for x in content if "unsetenv BAR" in x]) == 0
-        assert len([x for x in content if "is-loaded foo/bar" in x]) == 1
+        assert len([x for x in content if "is-loaded 'foo/bar'" in x]) == 1
         assert len([x for x in content if "module load foo/bar" in x]) == 1
         assert len([x for x in content if "setenv LIBDWARF_ROOT" in x]) == 1
 
--- a/share/spack/templates/modules/modulefile.tcl
+++ b/share/spack/templates/modules/modulefile.tcl
@@ -23,7 +23,7 @@ proc ModulesHelp { } {
 
 {% block autoloads %}
 {% for module in autoload %}
-if {{ '{' }} [ module-info mode load ] && ![ is-loaded {{ module }} ] {{ '}' }} {{ '{' }}
+if {{ '{' }} ![ is-loaded '{{ module }}' ] {{ '}' }} {{ '{' }}
 {% if verbose %}
     puts stderr "Autoloading {{ module }}"
 {% endif %}


haampie (dismissed):
I think this makes sense. The reason `module-info mode load` was there is probably to work around a bug in an old version of `environment-modules`, I can't reproduce the problem.

haampie:
@skosukhin do you have time to fix the tests?

skosukhin:
@haampie it's green now

haampie (approved):

xdelaruelle:
@skosukhin @haampie Thanks for proposing such change. You mention that _the single quotes around bar/4.5.6 are important_, could you please elaborate on that?

Unlike shell script, single quotes are treated literally on Tcl. It is not considered as a character to enclose a string.

So the `![ is-loaded 'bar/4.5.6' ]` test is always true even if `bar/4.5.6` is loaded (as it looks for `'bar/4.5.6'` not `bar/4.5.6`).

But it is important to always run the `module load bar/4.5.6` command even if `bar/4.5.6` is already loaded. If you skip it, you skip the dependency recording, so `foo/1.2.3` does not say it depends on `bar/4.5.6`, thus unloading bar will not automatically unload foo if it is loaded.

I suggest to remove the `is-loaded` test. But I wonder why it was put there in the beginning (support for Modules v3? or Lmod?). The `module load` command checks if module is already loaded and does not reload it if so.

haampie:
> The module load command checks if module is already loaded and does not reload it if so.

Sounds sensible

> So the ![ is-loaded 'bar/4.5.6' ] test is always true

Oof...

To me it would make sense to remove this branch, cause I believe `environment-modules` can deal with autoloading itself for a long time.

`MODULES_AUTO_HANDLING=1` is the default since [I have to check]

haampie:
Would you want to create a PR for that @xdelaruelle?

At this point, I'm not even sure if there is a difference between `prereq` and `depends-on` and just calling `module load`.

xdelaruelle:
@haampie I will work on a PR for that, sure.

`prereq`, `depends-on` and `module load` are quite the same on Modules if `auto_handling` config is on.

Most of the work is to check if it works well with Modules 3.2 (for people still on an EL7 system and not installing Modules with Spack) and with Lmod (as documentation states that Tcl modulefiles can also be used with Lmod)

PR #32854, opened 2022-09-27
intel-oneapi-compilers-classic: extend setup_run_environment with the one from intel-oneapi-compilers

skosukhin:
Currently, we don't get the expected environment setup after loading the `intel-oneapi-compilers-classic` **TCL** module:
```console
$ module load intel-oneapi-compilers-classic
$ icc --version
bash: icc: command not found
$ man icc
No manual entry for icc
```
This happens because `tcl` modules are not configured to autoload dependencies by default (in contrast to `lmod` modules). On the one hand, this can be solved with a configuration file (plus an optional fixup from #32853). On the other hand, the whole purpose of `intel-oneapi-compilers-classic` is to provide the users with a module file that works. So, I am not so sure about this really. To add more value to it, we could filter out environment modifications done by `intel-oneapi-compilers` that are not relevant for `intel-oneapi-compilers-classic` (and vice versa).

If the main idea of this PR is not accepted, I can remove one line and keep only the refactoring.

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-compilers-classic/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-compilers-classic/package.py
@@ -38,6 +38,9 @@ class IntelOneapiCompilersClassic(Package):
     def setup_run_environment(self, env):
         """Adds environment variables to the generated module file."""
         oneapi_pkg = self.spec["intel-oneapi-compilers"].package
+
+        oneapi_pkg.setup_run_environment(env)
+
         bin_prefix = oneapi_pkg.component_prefix.linux.bin.intel64
         env.set("CC", bin_prefix.icc)
         env.set("CXX", bin_prefix.icpc)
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/intel-oneapi-compilers-classic/package.py
```diff
@@ -36,26 +36,13 @@ class IntelOneapiCompilersClassic(Package):
...
-            self.spec["intel-oneapi-compilers"].version,
-            "linux",
-            "bin",
-            "intel64",
-        )
-        env.set("CC", join_path(bin, "icc"))
-        env.set("CXX", join_path(bin, "icpc"))
-        env.set("F77", join_path(bin, "ifort"))
-        env.set("FC", join_path(bin, "ifort"))
+        bin_prefix = oneapi_pkg.component_prefix.linux.bin.intel64
```
rscohn2, line 44: Do you know how the ``.`` is implemented? I did not know that you could mix variables and literal strings this way.

rscohn2 (approved):
Thanks for the fix. I don't know much about modules, but I verified that your change fixes the problem.

var/spack/repos/builtin/packages/intel-oneapi-compilers-classic/package.py
skosukhin: https://github.com/spack/spack/blob/699f575976c2467cabbd79596181384d6d887517/lib/spack/spack/util/prefix.py#L40-L41

PR #32862, opened 2022-09-28
Fix non-parallel make under depfile jobserver

### LINKED ISSUE #32861: Under jobserver, spack doesn't respect non-parallel make
### Steps to reproduce

I did some massive spack env depfile, and I encountered a bug: long story story short, when a package ask for `make(parallel=False)`, spack runs `make`, which is fine until you are running under a jobserver (`spack env depfile`) then this `make` runs in parallel.

(I already have a fix, I'll create a pull request)

I first encountered the bug when installing `py-pyqt5` using `spack env depfile`

```sh
source share/spack/setup-env.sh
spack env create testdepfile
spack env activate testdepfile
spack add py-pyqt5 %gcc@11.2.0
spack concretize -f
spack depfile -o make
spack env depfile -o make
make -f make -j8
```

### Error message

```console
==> Installing py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a
==> No binary for py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a found: installing from source
==> Fetching ...
==> No patches needed for py-pyqt5
==> py-pyqt5: Executing phase: 'configure'
==> py-pyqt5: Executing phase: 'build'
==> py-pyqt5: Executing phase: 'install'
==> Error: ProcessError: Command exited with status 2:
    'make' 'install'

1 error found in build log:
     15505    /.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/qt-5.15.5-4gkuklmukjxbnd4vbn5fs2szfbeydwl3/bin/qmake -install qinstall /.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/sip/QtGui/qsyntaxhighlighter.sip /.../spack/linux-ubuntu22
              .04-icelake/gcc-11.2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/share/sip/PyQt5/QtGui/qsyntaxhighlighter.sip
     15506    Traceback (most recent call last):
     15507      File "/.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/mk_distinfo.py", line 112, in <module>
     15508        fn_f = open(fn, 'rb')
     15509    FileNotFoundError: [Errno 2] No such file or directory: '/.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/lib/python3.9/site-packages/PyQt5/QtGui.so'
     15510    /.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/qt-5.15.5-4gkuklmukjxbnd4vbn5fs2szfbeydwl3/bin/qmake -install qinstall /.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/sip/QtWinExtras/qwinthumbnailtoolbutton.sip /.../spack/lin
              ux-ubuntu22.04-icelake/gcc-11.2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/share/sip/PyQt5/QtWinExtras/qwinthumbnailtoolbutton.sip
  >> 15511    make[1]: *** [Makefile:4225: install_distinfo] Error 1
     15512    make[1]: *** Waiting for unfinished jobs....
     15513    /.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/qt-5.15.5-4gkuklmukjxbnd4vbn5fs2szfbeydwl3/bin/qmake -install qinstall /.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/sip/QtWidgets/qapplication.sip /.../spack/linux-ubuntu22.0
              4-icelake/gcc-11.2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/share/sip/PyQt5/QtWidgets/qapplication.sip
     15514    /.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/qt-5.15.5-4gkuklmukjxbnd4vbn5fs2szfbeydwl3/bin/qmake -install qinstall /.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/sip/QtXmlPatterns/qsimplexmlnodemodel.sip /.../spack/linux
              -ubuntu22.04-icelake/gcc-11.2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/share/sip/PyQt5/QtXmlPatterns/qsimplexmlnodemodel.sip
     15515    /.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/qt-5.15.5-4gkuklmukjxbnd4vbn5fs2szfbeydwl3/bin/qmake -install qinstall /.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/sip/QtGui/qtextcursor.sip /.../spack/linux-ubuntu22.04-ice
              lake/gcc-11.2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/share/sip/PyQt5/QtGui/qtextcursor.sip
     15516    /.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/qt-5.15.5-4gkuklmukjxbnd4vbn5fs2szfbeydwl3/bin/qmake -install qinstall /.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/QtNetwork.pyi /.../spack/linux-ubuntu22.04-icelake/gcc-11.
              2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/lib/python3.9/site-packages/PyQt5/QtNetwork.pyi
     15517    /.../spack/linux-ubuntu22.04-icelake/gcc-11.2.0/qt-5.15.5-4gkuklmukjxbnd4vbn5fs2szfbeydwl3/bin/qmake -install qinstall /.../spack-stage/spack-stage-py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/spack-src/sip/QtCore/qpycore_qpair.sip /.../spack/linux-ubuntu22.04-
              icelake/gcc-11.2.0/py-pyqt5-5.13.1-uwi2k5antktiavaoreiwkn6dgsqpdz4a/share/sip/PyQt5/QtCore/qpycore_qpair.sip
```

* **Spack:** 0.19.0.dev0 (43ceff4193ebb5ab12594f1ace3d39b61e95393c)
* **Python:** 3.10.6
* **Platform:** linux-ubuntu22.04-icelake
* **Concretizer:** clingo

### General information

Jordan474:
Fix #32861 

When a package asks for non-parallel make, we need to force `make -j1` because just doing `make` will run in parallel under jobserver (e.g. `spack env depfile`).

We now always add `-j1` when asked for a non-parallel execution (even if there is no jobserver).

And each `MakeExecutable` can now ask for jobserver support or not. For example: the default `ninja` does not support jobserver so spack applies the default `-j`, but `ninja@kitware` or `ninja-fortran` does, so spack doesn't add `-j`.

Tips: you can run `SPACK_INSTALL_FLAGS=-j1 make -f spack-env-depfile.make -j8` to avoid massive job-spawning because of build tools that don't support jobserver (ninja).


lib/spack/spack/build_environment.py
```diff
@@ -150,14 +150,19 @@ def __call__(self, *args, **kwargs):
         parallel = should_set_parallel_jobs(jobserver_support=True) and kwargs.pop(
             "parallel", self.jobs > 1
         )
+        if not parallel:
+            # Force -j1 to disable parallel even under jobserver
+            jobs = 1
```
haampie: this branch is now followed when job server support is detected in general, which is incorrect.

Jordan474: Yes, the first version did, but then the test failed, and I did a mistake in the new version. I'll mark this PR as draft until I find a proper fix.

Jordan474:
It took me a few tries. There is few parameters to think about:
- `spack install -j`
- `make(parallel=false)`
- `SPACK_NO_PARALLEL_MAKE`
- jobserver or not

I changed `should_set_parallel_jobs` because I think the semantic was confusing: when non-parallel but with a jobserver we should still "set parallel jobs" to 1. So I changed it a more explicit `jobserver_enabled()`. See d3a0ac1c0ab5e04cf0636362b0861e23d9e3018b.

I think explicit `jobserver_enabled()` handling in `package.py` is fine because you need to know if the tool behind actually support jobserver.

There may be a more generic refactor to find but would probably be too disruptive (`determine_number_of_jobs(support_jobserver=True) -> Optional[int]` ? what about `make_jobs`, `package.parallel` ?).

Jordan474:
> There may be a more generic refactor to find but would probably be too disruptive

Note on this possible refactor: maybe the default `make_jobs` etc. could be lowered for tools that don't explicitly support jobserver. One can already emulate this via `SPACK_INSTALL_FLAGS=-j4 make -f depfile.make -j16`, as a security to avoid worst case scenario of 16 * 16 tasks running because of non-jobserver-compatible builds.

haampie (approved):

var/spack/repos/builtin/packages/cmake/package.py
```diff
@@ -310,7 +310,7 @@ def bootstrap_args(self):
         if not sys.platform == "win32":
             args.append("--prefix={0}".format(self.prefix))
 
-            if spack.build_environment.should_set_parallel_jobs(jobserver_support=True):
+            if not spack.build_environment.jobserver_enabled():
```
haampie: This was actually the reason there was a `should_set_parallel_jobs` function: the idea was to have it return true if and only if `-j` should be passed explicitly.

With this change, cmake does not respect `SPACK_NO_PARALLEL_MAKE` anymore

haampie (changes requested):
Can you make the cmake build respect `SPACK_NO_PARALLEL_MAKE` again?

If it simplifies the code, I'm also happy with we always invoke `make -j1` when `parallel=False` instead of `make`, so you'd get something along the lines of:

```python
def get_effective_jobs(jobs, parallel=True):
    """Returns the effective number of jobs to be specified on the command line when invoking make

    Returns:
        int or None: an integer to be passed as a -j <n> flag, or None when no -j flag should be passed
    """
    if not parallel or env_flag(SPACK_NO_PARALLEL_MAKE):
        return 1
    elif jobserver_enabled():
        return None
    return jobs
```

var/spack/repos/builtin/packages/cmake/package.py
Jordan474: I consciously dropped the support of `SPACK_NO_PARALLEL_MAKE` here because of the `-j`/not-`-j` confusion, and it didn't seem supported before the introduction of `should_set_parallel_jobs`.

I'll work on reintroducing it.

Jordan474:
Updated:

Fix `SPACK_NO_PARALLEL_MAKE` regression: now behaves like before this MR

Always force `-j1` when non-parallel, even when there is no jobserver

Each `MakeExecutable` can now ask to disable jobserver support (this fixes a comment about ninja by @trws).  Jobserver support is disabled for `ninja`, but re-enabled for `ninja-fortran` Kitware's fork, and could be re-enabled for `ninja` if https://github.com/ninja-build/ninja/pull/1140 is merged.

(Updated description)

var/spack/repos/builtin/packages/ninja/package.py
```diff
@@ -79,4 +79,5 @@ def setup_dependent_package(self, module, dspec):
         module.ninja = MakeExecutable(
             which_string(name, path=[self.spec.prefix.bin], required=True),
             determine_number_of_jobs(parallel=self.parallel),
+            supports_jobserver=False,
```
haampie:
```suggestion
            supports_jobserver=self.spec.version == ver('kitware'),
```

var/spack/repos/builtin/packages/cmake/package.py
```diff
@@ -307,11 +307,20 @@ def bootstrap_args(self):
         args = []
         self.generator = make
 
+        if self.spec.satisfies("platform=windows"):
+            args.append("-GNinja")
+            self.generator = ninja
+
         if not sys.platform == "win32":
```
haampie, line 314:
```suggestion
        else:
```

haampie (changes requested):
Looks good, we have the kitware fork here, so if you incorporate my suggestion I think this is good to go.

If you're a `depfile` user yourself, use

```yaml
spack:
  specs: [...]
  packages:
    ninja:
      require: '@kitware'
```

haampie:
@spackbot fix style

haampie (dismissed):

trws (approved):

var/spack/repos/builtin/packages/cmake/package.py
```diff
@@ -307,11 +307,20 @@ def bootstrap_args(self):
...
             args.append("--prefix={0}".format(self.prefix))
 
-            if spack.build_environment.should_set_parallel_jobs(jobserver_support=True):
-                args.append("--parallel={0}".format(make_jobs))
+            jobs = get_effective_jobs(
```
trws: Missing import

trws (changes requested):
As soon as this passes I'm good, comment on the current breakage above.

var/spack/repos/builtin/packages/cmake/package.py
Jordan474: Thanks, fixed.

trws (approved):

haampie (approved):

Jordan474:
The gitlab pipeline [keeps failing](https://gitlab.spack.io/spack/spack/-/jobs/4170628#L174), but I don't understand how this could come from my changes ?

tgamblin:
The issues this is getting don't look like they're from your changes.  We're working on this!  Sorry for the false alarms!

haampie:
@spackbot run pipeline

haampie:
Can you just hit [merge] @tgamblin?

tgamblin:
Done

PR #32874, opened 2022-09-28
Optimize concurrent misc_cache provider index rebuild

Jordan474:
When concurrent misc_cache provider index rebuilds happen, try to rebuild it only once, so we don't exceed misc_cache lock timeout.

For example, when using `spack env depfile`, with no previous misc_cache, running `make -f depfile -j8` could run at most 8 concurrent `spack install` locking on misc_cache to rebuild the provider index. If one rebuild takes 30s, before this fix, the "worst" lock could wait up to 30s * 7, easily exceeding misc_cache lock timeout. Now, the "worst" lock would take 30s * 1 + ~1s * 6.

The indexer cache file is still re-read and re-written multiple times, but to optimize that, one would probably need to refactor `write_transaction`, separating the locking part from the read/write part.

### DIFF
--- a/lib/spack/spack/repo.py
+++ b/lib/spack/spack/repo.py
@@ -559,6 +559,9 @@ class FastPackageChecker(Mapping):
     def last_mtime(self):
         return max(sinfo.st_mtime for sinfo in self._packages_to_stats.values())
 
+    def modified_since(self, since):
+        return [name for name, sinfo in self._packages_to_stats.items() if sinfo.st_mtime > since]
+
     def __getitem__(self, item):
         return self._packages_to_stats[item]
 
@@ -739,8 +742,7 @@ class RepoIndex(object):
         # Compute which packages needs to be updated in the cache
         misc_cache = spack.caches.misc_cache
         index_mtime = misc_cache.mtime(cache_filename)
-
-        needs_update = [x for x, sinfo in self.checker.items() if sinfo.st_mtime > index_mtime]
+        needs_update = self.checker.modified_since(index_mtime)
 
         index_existed = misc_cache.init_entry(cache_filename)
         if index_existed and not needs_update:
@@ -752,6 +754,12 @@ class RepoIndex(object):
             # Otherwise update it and rewrite the cache file
             with misc_cache.write_transaction(cache_filename) as (old, new):
                 indexer.read(old) if old else indexer.create()
+
+                # Compute which packages needs to be updated **again** in case someone updated them
+                # while we waited for the lock
+                new_index_mtime = misc_cache.mtime(cache_filename)
+                if new_index_mtime != index_mtime:
+                    needs_update = self.checker.modified_since(new_index_mtime)
 
                 for pkg_name in needs_update:
                     namespaced_name = "%s.%s" % (self.namespace, pkg_name)


haampie:
~This sounds reasonable to me. Maybe split the duplicated code into a separate function?~

On second thought: it's better to unconditionally acquire the write lock, and do the mtime check once it's obtained, since Spack can come to a crawl on slow filesystems when tons of stat calls are made (1 for each package). With this change, we'd always repeat the mtime check; even if the second time is faster due to warm cache, it's still a slowdown for the most common use of Spack.

Jordan474:
> On second thought: it's better to unconditionally acquire the write lock

Do you mean removing the read lock entirely? If so, I could agree *if* one refactors `write_transaction` to separate the lock from the read/write: the current implementation always writes the indexer cache, so it would write it at (nearly) all spack commands (I've seen it take more than a few seconds). So, not sure if this is better, especially because most spack commands are not run concurrently.

My change doesn't do a second mass-stat on recipes, but only a single additional stat on misc_cache index, and also I think the `self.checker` caches recipes mtime anyway.

(There is a datarace with recipe modification, but it was there before: when a recipe is modified externally between the indexer update and the cache write. A fix could be to check recipe mtimes against a date manually saved before the `indexer.update` instead of the index mtime.)

haampie:
Hm ok. I didn't read it right the first time, thought you meant to check the mtime of the packages again.

How about just letting all processes create/update the index in-memory concurrently, and then let them fight over the write lock. So, some redundant work is done in parallel, followed by a few fast sequential writes (no slow updates in write_transaction). Functionally it's more or less equivalent to before.

```diff
diff --git a/lib/spack/spack/repo.py b/lib/spack/spack/repo.py
index 814f84049a..6c90c5b11b 100644
--- a/lib/spack/spack/repo.py
+++ b/lib/spack/spack/repo.py
@@ -743,20 +743,18 @@ def _build_index(self, name, indexer):
         needs_update = [x for x, sinfo in self.checker.items() if sinfo.st_mtime > index_mtime]
 
         index_existed = misc_cache.init_entry(cache_filename)
-        if index_existed and not needs_update:
-            # If the index exists and doesn't need an update, read it
+        if index_existed:
             with misc_cache.read_transaction(cache_filename) as f:
                 indexer.read(f)
-
         else:
-            # Otherwise update it and rewrite the cache file
-            with misc_cache.write_transaction(cache_filename) as (old, new):
-                indexer.read(old) if old else indexer.create()
+            indexer.create()
 
-                for pkg_name in needs_update:
-                    namespaced_name = "%s.%s" % (self.namespace, pkg_name)
-                    indexer.update(namespaced_name)
+        if needs_update:
+            for pkg_name in needs_update:
+                namespaced_name = "%s.%s" % (self.namespace, pkg_name)
+                indexer.update(namespaced_name)
 
+            with misc_cache.write_transaction(cache_filename) as (old, new):
                 indexer.write(new)
 
         return indexer.index
```

Jordan474:
I think your last proposed change would do more FS operations overall compared to my PR. Your change seems to make all jobs open recipes and lookup providers, in parallel, but still. My PR makes only the first concurrent job open recipes and lookup providers, the other jobs would see the cache mtime change, and most probably empty their package list to update.

haampie (approved):
Ok, I'm fine with this change as is

haampie:
@spackbot run pipeline

Issue #32863, opened 2022-09-28
Installation issue: Autotools-based builds failing with dpcpp


cmacmackin:
### Steps to reproduce the issue

There is a [known bug](https://community.intel.com/t5/Intel-oneAPI-Data-Parallel-C/dpcpp-and-GNU-Autotools/m-p/1296985) when using `dpcpp` within autotools-based builds. In this report I am using the `pcre` package (@drkennetz) as a representative and relatively fast-failing example.

```console
$ spack spec -I pcre%dpcpp
Input spec
--------------------------------
pcre%dpcpp

Concretized
--------------------------------
x3hbwbv  pcre@8.45%dpcpp@2022.1.0~jit+multibyte+utf arch=linux-linuxmint21-icelake
```

### Error message

<details><summary>Error message</summary><pre>
==> Installing pcre-8.45-x3hbwbvcfyekhxno7n3etjaluzgpzxz5
==> No binary for pcre-8.45-x3hbwbvcfyekhxno7n3etjaluzgpzxz5 found: installing from source
==> Using cached archive: /home/cmacmack/.spack/var/spack/cache/_source-cache/archive/4d/4dae6fdcd2bb0bb6c37b5f97c33c2be954da743985369cddac3546e3218bffb8.tar.bz2
==> No patches needed for pcre
==> pcre: Executing phase: 'autoreconf'
==> pcre: Executing phase: 'configure'
==> pcre: Executing phase: 'build'
==> Error: ProcessError: Command exited with status 2:
    'make' '-j8' 'V=1'

4 errors found in build log:
     514    libtool: link: ranlib .libs/libpcre.a
     515    libtool: link: ( cd ".libs" && rm -f "libpcre.la" && ln -s "../libpcre.la" "libpcre.la" )
     516    /bin/bash ./libtool  --tag=CC   --mode=link /home/cmacmack/.spack/lib/spack/env/oneapi/icx -fvisibility=h
            idden   -g -O2 -version-info 0:7:0  -o libpcreposix.la -rpath /home/cmacmack/.spack/opt/spack/linux-linux
            mint21-icelake/dpcpp-2022.1.0/pcre-8.45-x3hbwbvcfyekhxno7n3etjaluzgpzxz5/lib libpcreposix_la-pcreposix.lo
             libpcre.la
     517    /bin/bash ./libtool  --tag=CXX   --mode=link /home/cmacmack/.spack/lib/spack/env/oneapi/dpcpp -fvisibilit
[...]

PR #32876, opened 2022-09-28
autotools: Filter libtools when building with dpcpp

cmacmackin:
Due to a [known issue](https://community.intel.com/t5/Intel-oneAPI-Data-Parallel-C/dpcpp-and-GNU-Autotools/m-p/1296985) with dpcpp, autotool-based builds that try to use it will fail; see #32863. This pull request modifies the Autotools package class to filter spurious object files out of the libtool script, as is already done to address issues with other compilers.

### DIFF
--- a/lib/spack/spack/build_systems/autotools.py
+++ b/lib/spack/spack/build_systems/autotools.py
@@ -307,9 +307,7 @@ To resolve this problem, please try the following:
             fs.filter_file(
                 r"^(predep_objects=.*)/tmp/conftest-[0-9A-Fa-f]+\.o", r"\1", libtool_path
             )
-            fs.filter_file(
-                r"^(predep_objects=.*)/tmp/a-[0-9A-Fa-f]+\.o", r"\1", libtool_path
-            )
+            fs.filter_file(r"^(predep_objects=.*)/tmp/a-[0-9A-Fa-f]+\.o", r"\1", libtool_path)
 
     @property
     def configure_directory(self):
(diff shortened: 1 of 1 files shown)


alalazo:
@spackbot run pipeline

haampie:
I think the solution is alright, but is this still an issue with the latest dpcpp? If so, can you please ask for the status regarding

> We are working on your issue internally. We will get back to you soon.

Otherwise:

1. Please add a comment clearly explaining why this change was necessary, linking to the original issue.
2. Restrict the change to the lines starting with `predep_objects_...` to avoid false positives.

haampie (changes requested):
I think the solution is alright. Also, can you please ask for the status regarding

> We are working on your issue internally. We will get back to you soon.

on their forum / helpdesk? Otherwise we're stuck with these hacks forever...

Otherwise:

1. Please add a comment clearly explaining why this change was necessary, linking to the original issue.
2. Restrict the change to the lines starting with `predep_objects_...` to avoid false positives.

cmacmackin:
> I think the solution is alright. Also, can you please ask for the status regarding
> 
> > We are working on your issue internally. We will get back to you soon.
> 
> on their forum / helpdesk? Otherwise we're stuck with these hacks forever...

I wasn't the original person to create that thread, but I've [posted a reply](community.intel.com/t5/Intel-oneAPI-Data-Parallel-C/dpcpp-and-GNU-Autotools/m-p/1419079#M2568) asking about this.
 
> Otherwise:
> 
>     1. Please add a comment clearly explaining why this change was necessary, linking to the original issue.
> 
>     2. Restrict the change to the lines starting with `predep_objects_...` to avoid false positives.

I've just pushed a new commit with these changes.

haampie (approved):

PR #32887, opened 2022-09-28
Update `url` field in Trilinos recipe

kuberry:
Recipe had format:

```
https://github.com/trilinos/Trilinos/archive/trilinos-release-12-12-1.tar.gz
```

which at times (github redirect outage?) was inaccessible for versions 13+, whereas 

```
https://github.com/trilinos/Trilinos/archive/refs/tags/trilinos-release-12-12-1.tar.gz
```

continues to work when redirects are not functioning.


tldahlgren (approved):
Confirmed all existing sha256 have stayed the same.  FWIW. the sha256 for `12.18.1` is:

```
    version("12.18.1", sha256="7d2e1a1a3e2d2ce78bb248c63411fed5007c827b87cadee3d13c1a45c09a89cf")
```

PR #32890, opened 2022-09-29
libcroco does not respect gtk-doc configure flag, so removing variant

lee218llnl:
The libcroco configure script was not respecting --enable-gtk-doc=no and thus depends on gtk-doc regardless of that flag. Until this is fixed upsream, it seems like the +doc variant is the only option and thus, this PR removes the variant and depends on gtk-doc without condition.

### DIFF
--- a/var/spack/repos/builtin/packages/libcroco/package.py
+++ b/var/spack/repos/builtin/packages/libcroco/package.py
@@ -15,9 +15,11 @@ class Libcroco(AutotoolsPackage):
     version("0.6.13", sha256="767ec234ae7aa684695b3a735548224888132e063f92db585759b422570621d4")
     version("0.6.12", sha256="ddc4b5546c9fb4280a5017e2707fbd4839034ed1aba5b7d4372212f34f84f860")
 
-    # libcroco has a --enable-gtk-doc configure flag that appears to be ignored     # as of version 0.6.13. Until that flag is honored, the +doc variant is a
-    # no-op
-    #variant("doc", default=False, description="Build documentation with gtk-doc")
+    # libcroco has a --enable-gtk-doc configure flag that appears to be
+    # ignored as of version 0.6.13. Until that flag is honored, the +doc
+    # variant is a no-op
+    # variant("doc", default=False,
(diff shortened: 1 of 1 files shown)


lee218llnl:
@spackbot run pipeline

var/spack/repos/builtin/packages/libcroco/package.py
```diff
@@ -15,16 +15,18 @@ class Libcroco(AutotoolsPackage):
     version("0.6.13", sha256="767ec234ae7aa684695b3a735548224888132e063f92db585759b422570621d4")
     version("0.6.12", sha256="ddc4b5546c9fb4280a5017e2707fbd4839034ed1aba5b7d4372212f34f84f860")
 
-    variant("doc", default=False, description="Build documentation with gtk-doc")
+    # libcroco has a --enable-gtk-doc configure flag that appears to be
+    # ignored as of version 0.6.13. Until that flag is honored, the +doc
+    # variant is a no-op
+    # variant("doc", default=False,
```
tldahlgren, line 21: Is it worth making this a conditional variant that only applies to prior versions?

https://spack.readthedocs.io/en/latest/packaging_guide.html#conditional-variants

lee218llnl: I think it applies t all variants at this point. I tried to contact the libcroco developers to fix this, but they haven't responded. From what I can see in their public repo, this still isn't fixed. Perhaps if/when it is fixed, then we can make it a conditional variant on spack.

lee218llnl: @tldahlgren do you still think we should make this a conditional variant?

tldahlgren: No.

tldahlgren (approved):
LGTM

var/spack/repos/builtin/packages/libcroco/package.py
tldahlgren: Thanks for the "ping".

PR #32897, opened 2022-09-29
claw: enable building with oneapi and nag+gcc@10:

skosukhin:
1. Enable building with `%oneapi` (I am not aware of the version that would not need the `-fcommon` flag).
2. Enable building with `%nag` that is mixed with `gcc@10:`. To support this case, we have to detect the real version of `gcc`, which is implemented in `_get_real_compiler_spec`.


tldahlgren (approved):
LGTM

PR #25015, opened 2021-07-21
LAMMPS: Add version 20210702

RemiLacroix-IDRIS:
A lot of packages were renamed (https://github.com/lammps/lammps/issues/2525) so I tried to deal with that properly.

I still needed to duplicate a lot of `conflicts` and `depends_on` so it's far from perfect.

Also the `user-omp` package was renamed to `openmp` which is very unfortunate since the new name matches the name of an existing Spack variant used for something else. This means that this PR is currently broken and I'm not sure how to deal with this.

I could handle this manually (as I had to do for the `meam` package which existed in the past) but that means having a variant name that doesn't match the corresponding LAMMPS package name which would be an exception. I don't see any other solution though...

Any feedback would be appreciated.

PR #32898, opened 2022-09-29
[lammps] Add +intel and +user-intel options

stephenmsachs:
Going ahead adding this option individually since
https://github.com/spack/spack/pull/25015 needs some more time.


tldahlgren (approved):
LGTM

PR #32899, opened 2022-09-29
added a PoC for limiting reuse of specs (by target)

flxmr:
So, this does, what I want in https://github.com/spack/spack/issues/32888. 

essentially if I do `spack spec --limit-target-reuse=x86_64,zen2  tar %gcc@12.2.0 target=zen2` the resulting spec will only try to reuse two architectures, when concretizing. This enables me to build different "stacks" in a clean way within one spack installation

It has some problems however:

1) I can't overwrite all scopes (=settings from the config-files) from the commandline (how would I do that)
2) I probably should extract the targets of the specs I want to build to add to the list of valid targets? 
3) this logic applies at least to compiler too and it's handled there already, how do you do this there?
4) is there any other place where this should better be/what does it need to integrate (except for tests)?

### DIFF
--- a/etc/spack/defaults/concretizer.yaml
+++ b/etc/spack/defaults/concretizer.yaml
@@ -16,6 +16,8 @@ concretizer:
   # concretization. If `dependencies`, we'll only reuse dependencies but
   # give you a fresh concretization for your root specs.
   reuse: dependencies
+  reuse: true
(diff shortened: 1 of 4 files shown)


alalazo:
Implemented more generically in Spack v0.22

PR #32901, opened 2022-09-29
Set CMAKE_HIP_ARCHITECTURES with the value of amdgpu_target

aurianer:
Since we use `enable_language(HIP)` in pika, the architectures were not properly forwarded

### DIFF
--- a/var/spack/repos/builtin/packages/pika/package.py
+++ b/var/spack/repos/builtin/packages/pika/package.py
@@ -142,11 +142,11 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
         ]
 
         # HIP support requires compiling with hipcc for < 0.8.0
-        if "@:0.7 +rocm" in self.spec:
+        if self.spec.satisfies("@:0.7 +rocm"):
             args += [self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc)]
             if self.spec.satisfies("^cmake@3.21.0:3.21.2"):
                 args += [self.define("__skip_rocmclang", True)]
-        if "@0.8: +rocm" in self.spec:
+        if self.spec.satisfies("@0.8: +rocm"):
             rocm_archs = spec.variants["amdgpu_target"].value
             rocm_archs = ";".join(rocm_archs)
             args.append(self.define("CMAKE_HIP_ARCHITECTURES", rocm_archs))
--- a/var/spack/repos/builtin/packages/pika/package.py
+++ b/var/spack/repos/builtin/packages/pika/package.py
@@ -146,5 +146,9 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
             args += [self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc)]
             if self.spec.satisfies("^cmake@3.21.0:3.21.2"):
                 args += [self.define("__skip_rocmclang", True)]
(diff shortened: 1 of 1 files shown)


msimberg (dismissed):

msimberg (dismissed):

albestro:
Isn't this something that we should do in the `ROCmPackage` build system?

msimberg:
> Isn't this something that we should do in the `ROCmPackage` build system?

It's a good question. I think one problem is that `ROCmPackage` does not inherit from `CMakePackage` (or `AutotoolsPackage` etc.), so it can't automatically set `CMAKE_HIP_ARCHITECTURES` (though I don't know python very well so I may be missing something). I suppose one alternative would be a `cmake_flags` method like `hip_flags` (https://github.com/spack/spack/blob/679cd4b60dcd9e1fc471cc98a7f2002f06f194dd/lib/spack/spack/build_systems/rocm.py#L136-L139) that at least centralizes the definitions. However, that sounds like something that should be done separately from this. There are already a few other packages setting `CMAKE_HIP_ARCHITECTURES` like here. The same concern applies to `CMAKE_CUDA_ARCHITECTURES` by the way.

albestro:
It sounds reasonable, good point. I agree we can think about it in another PR and eventually “fix” later the packages. 😉

var/spack/repos/builtin/packages/pika/package.py
```diff
@@ -146,5 +146,9 @@ def cmake_args(self):
             args += [self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc)]
             if self.spec.satisfies("^cmake@3.21.0:3.21.2"):
                 args += [self.define("__skip_rocmclang", True)]
+        if "@0.8: +rocm" in self.spec:
```
tldahlgren: We generally recommend using `self.spec.satisfies()` when checking version ranges (https://spack.readthedocs.io/en/latest/packaging_guide.html#testing-spec-constraints), which this includes.

```suggestion
        if self.spec.satisfies("@0.8: +rocm"):
```

tldahlgren (dismissed):
Let me know if you want this PR merged as-is.

var/spack/repos/builtin/packages/pika/package.py
msimberg: Thanks for pointing that out! The `if "@:0.7 ..."` line earlier should presumably also be changed in that case?

To be honest I can't quite grok what the difference between the two is based on the documentation you linked. It says "The difference between the two functions is that `satisfies()` tests whether spec constraints overlap at all, while `in` tests whether a spec or any of its dependencies satisfy the provided spec". "spec constraints overlap at all" and "satisfy the provided spec" sounds similar-ish. The only meaningful distinction I see there is that the latter says "a spec or any of its dependencies", but I don't quite understand the relevance of it. Is there another place explaining the difference in more detail? More importantly, is there _any_ reason to prefer `in` (besides that it reads nicely)?

teonnik (dismissed):

msimberg (approved):

var/spack/repos/builtin/packages/pika/package.py
aurianer: I just updated the package to use `satisfies` but I'm still curious about the difference between the two :)

haampie (approved):

PR #32902, opened 2022-09-29
py-sphinxcontrib-bibtex: new version 2.5.0

sethrj:
2.4 seems to have issues with sphinx-rtd and sphinx 5.1:
```
AttributeError: 'Text' object has no attribute 'rawsource'
```


adamjstewart:
> 2.4 seems to have issues with sphinx-rtd and sphinx 5.1:

Can you change the depends_on or add conflicts or something to document this?

adamjstewart (changes requested):
Missing dep:
```
importlib_metadata>=3.6; python_version < '3.10'
```

adamjstewart:
Ping @sethrj

sethrj:
Sorry @adamjstewart , I set up this PR while trying to analyze some results on a deadline and couldn't address the feedback earlier.

> Can you change the depends_on or add conflicts or something to document this?

Would this look like `conflicts('py-sphinx-rtd-theme', when='@:2.4  ^py-sphinx@5:')`? i.e. it conflicts when in the same dependency tree as rtd while using sphinx  5? And, like with all conflicts, I can state the exact configuration that failed and can *guess* at the generalized version, but I can't prove that I'm getting the conflict exactly right.

adamjstewart:
Where exactly does the conflict arise? Is the issue that sphinxcontrib-bibtex 2.4 doesn't work for sphinx 5, or that sphinx-rtd-theme doesn't work for sphinx 5, or what? What's the full stack trace?

sethrj:
> Where exactly does the conflict arise? Is the issue that sphinxcontrib-bibtex 2.4 doesn't work for sphinx 5, or that sphinx-rtd-theme doesn't work for sphinx 5, or what? What's the full stack trace?

When using all three together, one of them emits a piece of code with a callback that another doesn't like. If you've written your own docutils/sphinx extensions before you know what I mean: it's nigh impossible to tell which package is at fault. I can try to reproduce, but I don't really think it's worth blocking the version update—it's unlikely anyone will ever run into that error again.

adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

PR #32903, opened 2022-09-29
py-ipython: add 8.5.0 and py-win-unicode-console: new package

manuelakuhn:
https://github.com/ipython/ipython/tree/8.5.0
https://github.com/Drekin/win-unicode-console/tree/0.5

Also fix old dependencies.

Version 3.1.0 and 2.3.1 do not build (also before this PR). What should we do with those?

### DIFF
--- a/var/spack/repos/builtin/packages/py-ipython/package.py
+++ b/var/spack/repos/builtin/packages/py-ipython/package.py
@@ -15,6 +15,11 @@ class PyIpython(PythonPackage):
     pypi = "ipython/ipython-7.18.1.tar.gz"
     git = "https://github.com/ipython/ipython"
 
+    # "IPython.kernel" was deprecated for py-ipython@:7 and fails to import, leave
+    # out of "import_modules" to ensure that import tests pass.
+    # for py-ipython@8: "IPython.kernel" was removed
+    skip_modules = ["IPython.kernel"]
+
     version("8.5.0", sha256="097bdf5cd87576fd066179c9f7f208004f7a6864ee1b20f37d346c0bcb099f84")
     version("8.0.1", sha256="ab564d4521ea8ceaac26c3a2c6e5ddbca15c8848fd5a5cc325f960da88d42974")
     version("7.31.1", sha256="b5548ec5329a4bcf054a5deed5099b0f9622eb9ea51aaa7104d215fece201d8c")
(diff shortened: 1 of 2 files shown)


adamjstewart:
> Version 3.1.0 and 2.3.1 do not build (also before this PR). What should we do with those?

Deprecate.

var/spack/repos/builtin/packages/py-ipython/package.py
```diff
@@ -35,43 +45,55 @@ class PyIpython(PythonPackage):
...
-    def import_modules(self):
-        modules = super(__class__, self).import_modules
+    def skip_modules(self):
 
         if self.spec.satisfies("@8:"):
-            return modules
+            return []
 
         # 'IPython.kernel' is deprecated and fails to import, leave out of
         # 'import_modules' to ensure that import tests pass.
```
adamjstewart: Trying to understand this logic. In IPython 7 and older, IPython.kernel was deprecated, so we skip that import. But in IPython 8 it is no longer deprecated and imports fine? Or it no longer exists so there's no reason to skip it? Maybe we can just always skip it if it's the latter.

manuelakuhn: `IPython.kernel` no longer exists in newer version, that's why the test pass now without skipping of modules. The idea behind the distinction between version was to modify `import_modules` as little as possible.

adamjstewart: Let's just always return `IPython.kernel`, not returning anything when it doesn't exist seems like an unnecessary step. Alternatively, we can only skip it after it becomes deprecated.

manuelakuhn: Done.

adamjstewart (dismissed):

var/spack/repos/builtin/packages/py-ipython/package.py
```diff
@@ -35,43 +45,50 @@ class PyIpython(PythonPackage):
...
+        "py-win-unicode-console@0.5:",
+        when="@5:7.9 ^python@:3.5 platform=windows",
+        type=("build", "run"),
+    )
 
-        return [i for i in modules if not any(map(i.startswith, ignored_imports))]
+    # "IPython.kernel" was deprecated for py-ipython@:7 and fails to import, leave
+    # out of "import_modules" to ensure that import tests pass.
+    # for py-ipython@8: "IPython.kernel" was removed
+    skip_modules = ["IPython.kernel"]
```
adamjstewart: Can we move this to the top? Maybe below `git`?

adamjstewart (approved):

PR #32906, opened 2022-09-29
Add aws-ofi-nccl library

bvanessen:
Added a package for the aws-ofi-nccl plug-in from to enable NCCL to use libfabric communication library as a network provider.

### DIFF
--- a/var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
+++ b/var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
@@ -19,8 +19,8 @@ class AwsOfiNccl(AutotoolsPackage):
 
     version("master", branch="master", default=True)
 
-    variant("enable-trace", default=False, description="Enable printing trace messages")
-    variant("disable-tests", default=False, description="Disable build of tests")
+    variant("trace", default=False, description="Enable printing trace messages")
+    variant("tests", default=False, description="Build tests")
 
     depends_on("libfabric")
     depends_on("cuda")
@@ -44,6 +44,8 @@ class AwsOfiNccl(AutotoolsPackage):
         spec = self.spec
         args = []
 
+        # Always set configure's external paths to use the Spack
+        # provided dependencies
         args.extend(
             [
                 "--with-libfabric={0}".format(spec["libfabric"].prefix),
@@ -53,10 +55,7 @@ class AwsOfiNccl(AutotoolsPackage):
             ]
         )
 
-        if "+enable-trace" in self.spec:
-            args.append("--enable-trace")
-
-        if "+disable-tests" in self.spec:
-            args.append("--disable-tests")
+        args.extend(self.enable_or_disable('trace'))
+        args.extend(self.enable_or_disable('tests'))
 
         return args
--- a/var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
+++ b/var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
@@ -32,7 +32,7 @@ class AwsOfiNccl(AutotoolsPackage):
 
     # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
     def setup_run_environment(self, env):
-        aws_ofi_nccl_home = self.spec["aws-ofi-nccl"].prefix
+        aws_ofi_nccl_home = self.spec.prefix
         env.append_path("LD_LIBRARY_PATH", aws_ofi_nccl_home.lib)
 
     # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
--- a/var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
+++ b/var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
@@ -1,0 +1,62 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
```diff
@@ -0,0 +1,62 @@
...
+    def configure_args(self):
+        spec = self.spec
+        args = []
+
+        args.extend(
+            [
+                "--with-libfabric={0}".format(spec["libfabric"].prefix),
+                "--with-cuda={0}".format(spec["cuda"].prefix),
+                "--with-nccl={0}".format(spec["nccl"].prefix),
+                "--with-mpi={0}".format(spec["mpi"].prefix),
```
tldahlgren, line 54: Have you considered using the `with_or_without` helper (see https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#activation-value)?

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
```diff
@@ -0,0 +1,62 @@
...
+            ]
+        )
+
+        if "+enable-trace" in self.spec:
+            args.append("--enable-trace")
+
+        if "+disable-tests" in self.spec:
+            args.append("--disable-tests")
```
tldahlgren: You could also use `enable_or_disable` for (simplified) versions of these variants (see https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#enable-or-disable).

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
```diff
@@ -0,0 +1,62 @@
...
+    libfabric as a network provider while running NVIDIA's NCCL based
+    applications."""
+
+    homepage = "https://github.com/aws/aws-ofi-nccl"
+    git = "https://github.com/aws/aws-ofi-nccl.git"
+    url = "https://github.com/aws/aws-ofi-nccl.git"
+
+    maintainers = ["bvanessen"]
+
+    version("master", branch="master", default=True)
```
tldahlgren, line 20: Confirmed this branch exists.

tldahlgren (commented):
LGTM though the code could be tightened by using some of the `AutotoolsPackage` helpers.  Let me know if you want the package merged as-is.

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
bvanessen: @tldahlgren I didn't know about those options, so thanks for the pointers, but I believe that we should always set these since they are not actually optional dependences and they should always use the spack versions.

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
bvanessen: @tldahlgren Done.

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
```diff
@@ -0,0 +1,61 @@
...
+    depends_on("nccl")
+    depends_on("mpi")
+    depends_on("autoconf", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+
+    # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
+    def setup_run_environment(self, env):
+        aws_ofi_nccl_home = self.spec["aws-ofi-nccl"].prefix
+        env.append_path("LD_LIBRARY_PATH", aws_ofi_nccl_home.lib)
```
haampie, line 36: Should this work on non-Linux too?

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
```diff
@@ -0,0 +1,61 @@
...
+    depends_on("cuda")
+    depends_on("nccl")
+    depends_on("mpi")
+    depends_on("autoconf", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+
+    # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
+    def setup_run_environment(self, env):
+        aws_ofi_nccl_home = self.spec["aws-ofi-nccl"].prefix
```
haampie:
```suggestion
        aws_ofi_nccl_home = self.spec.prefix
```

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
```diff
@@ -0,0 +1,61 @@
...
+        env.append_path("LD_LIBRARY_PATH", aws_ofi_nccl_home.lib)
+
+    # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
+    def setup_dependent_run_environment(self, env, dependent_spec):
+        aws_ofi_nccl_home = self.spec["aws-ofi-nccl"].prefix
```
haampie, line 40:
```suggestion
        aws_ofi_nccl_home = self.spec.prefix
```

var/spack/repos/builtin/packages/aws-ofi-nccl/package.py
bvanessen: @haampie I don't think so.  This is really for either AWS instances or for HPC cray instances with slingshot fabrics.  So I don't know that we need to support OSX or Windows.

tldahlgren (approved):
LGTM

PR #32907, opened 2022-09-29
Package updates for slate, lapackpp, and blaspp

G-Ragghianti:
These package updates correspond to SLATE release 2022-07-00.  The major change is the inclusion of HIP and CUDA support in lapackpp.

### DIFF
--- a/var/spack/repos/builtin/packages/lapackpp/package.py
+++ b/var/spack/repos/builtin/packages/lapackpp/package.py
@@ -58,6 +58,8 @@ class Lapackpp(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("blas")
     depends_on("lapack")
+    depends_on("rocblas", when="+rocm")
+    depends_on("rocsolver", when="+rocm")
 
     conflicts("+rocm", when="+cuda", msg="LAPACK++ can only support one GPU backend at a time")
 
--- a/var/spack/repos/builtin/packages/lapackpp/package.py
+++ b/var/spack/repos/builtin/packages/lapackpp/package.py
@@ -59,13 +59,23 @@ class Lapackpp(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("blas")
     depends_on("lapack")
 
+    conflicts("+rocm", when="+cuda", msg="LAPACK++ can only support one GPU backend at a time")
+
     def cmake_args(self):
         spec = self.spec
+
+        backend = "none"
+        if self.version >= Version("2022.07.00"):
+            if "+cuda" in spec:
+                backend = "cuda"
+            if "+rocm" in spec:
+                backend = "hip"
 
         args = [
             "-DBUILD_SHARED_LIBS=%s" % ("+shared" in spec),
             "-Dbuild_tests=%s" % self.run_tests,
             "-DLAPACK_LIBRARIES=%s" % spec["lapack"].libs.joined(";"),
+            "-Dgpu_backend=%s" % backend,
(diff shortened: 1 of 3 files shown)


var/spack/repos/builtin/packages/lapackpp/package.py
```diff
@@ -52,16 +56,34 @@ class Lapackpp(CMakePackage):
     for (lpp_ver, bpp_ver) in _versions:
         depends_on("blaspp@" + bpp_ver, when="@" + lpp_ver)
 
+    depends_on("blaspp ~cuda", when="~cuda")
+    depends_on("blaspp +cuda", when="+cuda")
+    depends_on("blaspp ~rocm", when="~rocm")
```
mgates3, line 61: Does `lapackpp~cuda` actually require `blaspp~cuda`,
and `lapackpp~rocm` require `blaspp~rocm`?
Normally that's what I would expect, that both are `+cuda`, or both `+rocm`, or both `~rocm ~cuda`, but I don't think it's required. It could be `lapackpp~rocm~cuda` and (`blaspp+rocm` or `blaspp+cuda`). But those are edge cases I wouldn't expect and wouldn't test, so may be simpler to disallow.

var/spack/repos/builtin/packages/slate/package.py
```diff
@@ -54,9 +57,14 @@ class Slate(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("blaspp ~cuda", when="~cuda")
     depends_on("blaspp +cuda", when="+cuda")
     depends_on("blaspp ~rocm", when="~rocm")
+    depends_on("lapackpp ~cuda", when="~cuda")
+    depends_on("lapackpp +cuda", when="+cuda")
+    depends_on("lapackpp ~rocm", when="~rocm")
```
mgates3, line 62: Ditto: `slate~cuda` doesn't necessarily imply `lapackpp~rocm` and `blaspp~rocm`. Again, an edge case that maybe isn't important.

mgates3 (approved):
Looks good. Just the one question about negative dependencies (`~cuda` requiring `~cuda`, `~rocm` requiring `~rocm`), which isn't a big deal.

tldahlgren (approved):
Confirmed all three new sha256.

var/spack/repos/builtin/packages/lapackpp/package.py
G-Ragghianti: I'd say that since we don't test those edge cases and they are unusual, that those cases should be considered unsupported.  It is more likely that someone would want cuda or rocm support turned on for all major components, so this makes it much easier for the user by chaining the gpu config down to the dependencies.

tldahlgren:
@G-Ragghianti @mgates3 Let me know when you're ready for this to be merged.

var/spack/repos/builtin/packages/slate/package.py
G-Ragghianti: If someone really needs to do that and we want to support it, it could be done, but it would complicate the spack package and would make it more difficult for the users who just want to turn on or off cuda nad rocm support for all components (slate, lapackpp, and blaspp).

G-Ragghianti: One addition: this just replicates the previous behavior such that slate depends on lapackpp in the same way that slate has previously depended on blaspp.

mgates3:
@tldahlgren Everything looks good. Thanks.

PR #32911, opened 2022-09-30
unifyfs: pin mercury version; add boost variant

CamStan:
Mercury has a new version (v2.2) releasing soon that UnifyFS does not build with and hasn't been tested with. This pins UnifyFS to the last version of Mercury used/tested (v2.1).

Add a variant to avoid building/using `boost`.

Append `-std=gnu99` to `cflags` if building with `gcc@4`. Needed for `mochi-margo` to compile.


tldahlgren (approved):
LGTM.

PR #32913, opened 2022-09-30
py-blis: fix environment settings

QianJianhua1:
Error occur when compiling with a compiler other than GCC, or on a platform other than x86_64. The reason is that two environment variables are not set.
-  BLIS_COMPILER
-  BLIS_ARCH

### DIFF
--- a/var/spack/repos/builtin/packages/py-blis/package.py
+++ b/var/spack/repos/builtin/packages/py-blis/package.py
@@ -18,3 +18,7 @@ class PyBlis(PythonPackage):
 
     depends_on("py-setuptools", type="build")
     depends_on("py-numpy@1.15:", type=("build", "run"))
+
+    def setup_build_environment(self, env):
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-blis/package.py
```diff
@@ -18,3 +18,7 @@ class PyBlis(PythonPackage):
 
     depends_on("py-setuptools", type="build")
     depends_on("py-numpy@1.15:", type=("build", "run"))
+
+    def setup_build_environment(self, env):
+        env.set("BLIS_COMPILER", self.compiler.cc)
```
adamjstewart: Is this the compiler used at build time or the compiler used at run time? If the former, you'll want to use:
```suggestion
        env.set("BLIS_COMPILER", spack_cc)
```
so that it uses Spack's compiler wrappers.

adamjstewart (approved):

PR #32914, opened 2022-09-30
armpl-gcc: Add version 22.1

annop-w:
- Add version 22.1
- Modify default url to point to version 22.1

### DIFF
--- a/var/spack/repos/builtin/packages/armpl-gcc/package.py
+++ b/var/spack/repos/builtin/packages/armpl-gcc/package.py
@@ -122,12 +122,16 @@ def get_os():
 def get_package_url(version):
     os = get_os()
     os_no_dash = get_os().replace("-", "")
+    os_lower = os.split(".")[0].lower()
     base_url = "https://developer.arm.com/-/media/Files/downloads/hpc/arm-performance-libraries/"
     armpl_version = version.split("_")[0]
     armpl_version_dashed = armpl_version.replace(".", "-")
     gcc_version = version.split("_")[1]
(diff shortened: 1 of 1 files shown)


annop-w:
@alalazo Could you please review this too ? Thank you.

haampie:
Why is the GCC version relevant here?

annop-w:
There are multiple versions of armpl compiled with different version of gcc. Therefore, we try to have users installed the version that is compatible with their gcc version.

var/spack/repos/builtin/packages/armpl-gcc/package.py
```diff
@@ -123,6 +154,12 @@ class ArmplGcc(Package):
     conflicts("target=ppc64:", msg="Only available on Aarch64")
     conflicts("target=ppc64le:", msg="Only available on Aarch64")
 
+    conflicts("%gcc@:11.0", when="@22.1_gcc-11.2")
```
haampie: Do you mean to say conflicts with GCC version `< 11.0`? Or `<= 11.0.x`?

annop-w: < 11.0.

haampie: In that case it's `@:10`, this upperbound _includes_ any `10.x` version.

annop-w: Done. I've also fixed a url issue.

haampie (approved):

PR #24025, opened 2021-05-31
add some more documentation for non-inclusive version ranges

cosmicexplorer:
### Problem

#26422 removes the `\.99+` suffix many packages append to version strings in order to support non-inclusive range bounds e.g. `1.2 <= v < 2`: previously we had recommended the use of `@1.2:1.2.999` for this, but we now recommend using `@1.2:1`. However, some packages still contain this `\.99+` workaround, and our documentation doesn't explicitly describe the new recommendation.

### Solution
- Fix some `\.99+` bounds that remained after #26422.
- Clarify in documentation how this works, and explicitly say that `\.99+` is deprecated.

### Result
We should be able to reserve `.99` components for actual versions named `.99` instead of using it as a workaround.

PR #32916, opened 2022-09-30
small refactorings for version.py

cosmicexplorer:
*I'm splitting this out from the work for #24025.*

`coerce_versions()` has some duplicated promotion logic that can be cut out without too much trickery. I think the modification here will be easier to maintain.


cosmicexplorer:
Someone whose regex opinions I respect has voiced their support (https://twitter.com/junyer/status/1575772555129987072?s=20&t=RYMYA1BBjkWzcVOHOIJXUQ) for the existing use of `[.]` to escape metacharacters, contrary to my initial impression, so I will probably revert the first commit here.

PR #32918, opened 2022-09-30
[rocksdb] Added rtti variant

mdorier:
This PR adds the `rtti` variant in rocksdb. This is needed when subclassing some rocksdb classes (e.g. Logger), and can be enabled via the `USE_RTTI` environment variable when calling `make`.


mdorier:
If someone can explain to me why style check isn't passing, because I don't get it...

haampie (approved):
Can you refactor the package next time to use `setup_build_environment` instead of `env[x] = y`?

PR #32919, opened 2022-09-30
codify version comparison protocols

cosmicexplorer:
### Problem

- The many "version-like" types in `version.py` each implement subsets of similarly-named methods, but these methods may have additional constraints (such as `strict=False`) which are only implemented for some but not all of the version-like types (`VersionBase`, `GitVersion`, `VersionStrComponent`, `VersionRange`, `VersionList`). 
- Compounding the confusion is that no type signatures are provided for these methods, along with the `@coerced` decorator which modifies the arguments before executing the actual function.
- Each version-like type's constructor accepts a wide array of inputs, but figuring out all of the things that e.g. `VersionList.__init__()` accepts is quite difficult.
- `eval(repr(x))` will often produce a list or string instead of another instance of `type(x)`, which breaks the `__repr__` contract.

### Solution

1. Split out shared comparison methods into a few abstract base classes, and add type signatures for those classes in a new file `version.pyi`.
    - Inconsistencies among similarly-named methods are fixed by ensuring each method conforms to the same contract.
    - Adding type signatures in a separate file avoids cluttering up the existing logic and git history in `version.py`.
1. Implement `__repr__` serialization uniformly with a `Serializable` base class, which introduces the `X.parse()` classmethod.
    - Tests are added for round-trip via `eval(repr(...))`.

### Result
I recognize that this is a highly opinionated change with unclear benefit. However, I believe it will make the work of #24025 significantly easier. As such, I will be leaving this PR as a draft until I can demonstrate its utility for implementing strict version inequalities.


cosmicexplorer:
After figuring out a much nicer way to do #24025 without modifying any logic in `version.py`, this can be closed.

PR #32928, opened 2022-09-30
casper: old domain fell off, adding github repo

snehring:
No apparently major changes. Dialect change doesn't seem to be needed any more.


tldahlgren (approved):
Confirmed the dated commit exists in the new repository.  Appreciate deprecation of the inaccessible `0.8.2` in case it is in some build cache.

I'm wondering if the `homepage` should be switch to the new GitHub repo but will leave that for another time.

PR #32934, opened 2022-09-30
Apply setup_dependent_package to superclass packages that are not shared

scheibelp:
If you have a package in the builtin repo like:

```
class Foo(Package):
    depends_on(cmake)

    def install():
        cmake(...)
```

and a custom package repository which extends that package like:

```
class Foo(builtin.Foo):
    ...
```

this will fail as of https://github.com/spack/spack/commit/b9e57006c98c5ea3836a33a55a41a516e52f3e4b#diff-a493ca2ca45c40f1a67e6ef60fe57654cb2077e3d20b50ceee9b85d0ec16c8e3 because `build_environment` only ensures that the subclass module has a `cmake` attribute defined.

This retrieves all packages in the inheritance DAG for the installed package, but excludes "shared" superclasses like `PackageBase`. This assumes that all packages in this inheritance hierarchy would have the same class name (e.g. in this case they would all be called `Foo`. I couldn't quickly think of a better criteria but other possible criteria is to only retrieve package superclasses if they are defined in a repo (this would cause problems if users created packages in their repo for the sole purpose of sharing amongst other packages).

### DIFF
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -775,6 +775,15 @@ def parent_class_modules(cls):
     return result
 
 
+def nonshared_parents(cls):
+    """Return list with cls and all superclasses that define a Package
+       class object with the same name (so excludes superclasses like
+       PackageBase).
+    """
+    all_parents = parent_class_modules(cls)
+    return list(x for x in all_parents if getattr(x, cls.__name__, None))
+
+
 def load_external_modules(pkg):
     """Traverse a package's spec DAG and load any external modules.
 
@@ -994,7 +1003,8 @@ def modifications_from_dependencies(
             if set_package_py_globals:
                 set_module_variables_for_package(dpkg)
             # Allow dependencies to modify the module
-            dpkg.setup_dependent_package(spec.package.module, spec)
+            for module in nonshared_parents(spec.package.__class__):
+                dpkg.setup_dependent_package(module, spec)
             if context == "build":
                 dpkg.setup_dependent_build_environment(env, spec)
             else:


lib/spack/spack/build_environment.py
```diff
@@ -775,6 +775,15 @@ def parent_class_modules(cls):
     return result
 
 
+def nonshared_parents(cls):
+    """Return list with cls and all superclasses that define a Package
+       class object with the same name (so excludes superclasses like
+       PackageBase).
+    """
+    all_parents = parent_class_modules(cls)
+    return list(x for x in all_parents if getattr(x, cls.__name__, None))
```
becker33, line 784: I don't think we want to exclude name changes -- what if someone has `my-hdf5` that shares everything except a URL with the builtin hdf5 -- they should be able to define a package `class MyHdf5(Hdf5):`.

What if instead we chose only classes that come from files named `package.py`? that should reasonably-effectively get only package classes, not base classes?

lib/spack/spack/build_environment.py
```diff
@@ -994,7 +1003,8 @@ def add_modifications_for_dep(dep):
             if set_package_py_globals:
                 set_module_variables_for_package(dpkg)
             # Allow dependencies to modify the module
-            dpkg.setup_dependent_package(spec.package.module, spec)
+            for module in nonshared_parents(spec.package.__class__):
+                dpkg.setup_dependent_package(module, spec)
```
becker33, line 1007: I worry that this will apply spec changes multiple times, which will be bad if they have side effects.

lib/spack/spack/build_environment.py
scheibelp: That would work IMO but would run into the same problem as my other suggestion:

> this would cause problems if users created packages in their repo for the sole purpose of sharing amongst other packages

lib/spack/spack/build_environment.py
scheibelp: Good point, although I don't know how to address that without changing the API (for all packages in builtin that define this, for example): we need to run `setup_dependent_package` on every superclass module; up to now, the `setup_dependent_package` has been free to modify the spec as it sees fit. If it only makes idempotent changes (e.g. `setattr`) then it would be fine (but we have not mandated that constraint before).

becker33:
There's also #32340 that attempts to solve the same issues.

scheibelp:
> There's also https://github.com/spack/spack/pull/32340 that attempts to solve the same issues

That looks great, like it avoids the issue you pointed out at https://github.com/spack/spack/pull/32934#discussion_r984974725

I assume you might want to use `nonshared_parents` (or your suggested tweak), but since `setup_package` occurs in the child process, then perhaps that is overdoing it as well.

In any case I'll close this.

PR #32937, opened 2022-10-01
octave: fix missing include <ctime>

manuelakuhn:
I could not install octave due to [this bug](https://savannah.gnu.org/bugs/?62750).

This PR adds the patch to fix it.

### DIFF
--- a/var/spack/repos/builtin/packages/octave/package.py
+++ b/var/spack/repos/builtin/packages/octave/package.py
@@ -107,6 +107,13 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
     depends_on("suite-sparse", when="+suitesparse")
     depends_on("zlib", when="+zlib")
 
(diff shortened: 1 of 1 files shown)


siko1056 (changes requested):
This problem is fixed with Octave 7.2.0

https://octave.org/news/release/2022/07/28/octave-7.2.0-released.html

Unfortunately, Octave 7.2.0 is not yet present in Spack.  I will add Octave 7.2.0 this weekend.

Do you agree @manuelakuhn to update to Octave 7.2.0 and avoid this patch?

I am sorry for the inconvenience.

manuelakuhn:
Sure, if you update octave that is also fine with me.

siko1056:
Thank you for your patience, see https://github.com/spack/spack/pull/32943 and I kindly close this PR.

PR #32940, opened 2022-10-01
py-tensorflow-estimator: add v2.10

adamjstewart:
https://github.com/tensorflow/tensorflow/blob/r2.10/RELEASE.md

Can't test because TF doesn't build, but hopefully this works.

### DIFF
--- a/var/spack/repos/builtin/packages/py-tensorflow-estimator/package.py
+++ b/var/spack/repos/builtin/packages/py-tensorflow-estimator/package.py
@@ -51,7 +51,7 @@ class PyTensorflowEstimator(Package):
         "2.0",
         "1.13",
     ]:
-        depends_on("py-tensorflow@", +ver, when="@" + ver, type=("build", "run"))
+        depends_on("py-tensorflow@" + ver, when="@" + ver, type=("build", "run"))
 
     depends_on("bazel@0.19.0:", type="build")
     depends_on("py-pip", type="build")
--- a/var/spack/repos/builtin/packages/py-tensorflow-estimator/package.py
+++ b/var/spack/repos/builtin/packages/py-tensorflow-estimator/package.py
@@ -17,6 +17,7 @@ class PyTensorflowEstimator(Package):
 
     maintainers = ["aweits"]
 
+    version("2.10", sha256="60df309377cf4e584ca20198f9639beb685d50616395f50770fc0999092d6d85")
     version("2.9.0", sha256="62d7b5a574d9c995542f6cb485ff1c18ad115afd9ec6d63437b2aab227c35ef6")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-tensorflow-estimator/package.py
```diff
@@ -31,23 +32,26 @@ class PyTensorflowEstimator(Package):
...
+        "2.8",
+        "2.7",
+        "2.6",
+        "2.5",
+        "2.4",
+        "2.3",
+        "2.2",
+        "2.1",
+        "2.0",
+        "1.13",
```
tldahlgren, line 52: This doesn't match the original requirement of `py-tensorflow@1.13.1`, does that matter?

var/spack/repos/builtin/packages/py-tensorflow-estimator/package.py
```diff
@@ -17,6 +17,7 @@ class PyTensorflowEstimator(Package):
 
     maintainers = ["aweits"]
 
+    version("2.10", sha256="60df309377cf4e584ca20198f9639beb685d50616395f50770fc0999092d6d85")
```
tldahlgren, line 20: Confirmed the sha256.

tldahlgren (commented):
Have a question about the new, much tighter dependency handling for one specific version.

var/spack/repos/builtin/packages/py-tensorflow-estimator/package.py
adamjstewart: Probably not, all TF estimator releases seem to correspond to a single TF release version. It's possible that older versions were more flexible as to which versions they claimed to support, but my experience with TF has been that versions need to match exactly, otherwise bad things happen. No one should be using such an old version of TF anyway.

tldahlgren (approved):
@aweits Did you want to comment on this PR before it is merged?

aweits (approved):
lgtm
```
==> Installing py-tensorflow-estimator-2.10-mniafgjp2k3gvbdfzkbwsfxcdgf7jt2q
==> No binary for py-tensorflow-estimator-2.10-mniafgjp2k3gvbdfzkbwsfxcdgf7jt2q found: installing from source
==> Fetching https://github.com/tensorflow/estimator/archive/v2.10.tar.gz
==> No patches needed for py-tensorflow-estimator
==> py-tensorflow-estimator: Executing phase: 'install'
==> py-tensorflow-estimator: Successfully installed py-tensorflow-estimator-2.10-mniafgjp2k3gvbdfzkbwsfxcdgf7jt2q
  Fetch: 1.34s.  Build: 9.12s.  Total: 10.47s.
[+] /dev/shm/spack/opt/spack/linux-rhel7-skylake_avx512/gcc-11.2.0/py-tensorflow-estimator-2.10-mniafgjp2k3gvbdfzkbwsfxcdgf7jt2q
```

PR #32905, opened 2022-09-29
Fixup errors introduced by Clingo Pr

johnwparent:
re2c depends on cmake on Windows

Winbison properly added to bootstrap package search list

PR #32942, opened 2022-10-01
Windows: fix bootstrap and package install failure

KiruyaMomochi:
After #32905 is merged, I will make this PR ready.

This PR fixes various bugs on Windows:

- Building `clingo@spack` fails on Windows if using VS2022 and maybe also older MSVC versions. It's because
  - `ssize_t` is not defined in MSVC.
  - `#undef _DEBUG` in pybind's `Python.h` makes it fails to compile on VS2022. https://github.com/pybind/pybind11/pull/3497 https://github.com/KiruyaMomochi/clingo/pull/1
- Python 3.8+ on Windows does not search dependent DLLs in PATH leading `clingo` fails to import. We nend to manually add it using `os.add_dll_directory()`.
  - DLL search path in 3.8+ is limited. https://github.com/python/cpython/issues/87339
  - Since https://github.com/spack/spack/pull/31930 is merged, maybe a prettier way to do it is to symlink `<clingo-prefix>/bin/clingo.dll` to `<clingo-prefix>/Lib/site-packages/`?
- The command returned by `which` should be quoted on Windows. Otherwise, it may work, but not always.
  1. Python use `CreateProcess` API to create subprocess on Windows[^1], which tries to search `C:\Program` first, and then `C:\Program Files` if that fails.[^2]
  2. If `C:\Program` doesn't exist, it *looks* like it works, but if someone has a file or directory at `C:\Program`, everything will break.   
    image
    image

[^1]: https://docs.python.org/3/library/subprocess.html#popen-constructor
[^2]: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa#parameters

### DIFF
--- a/lib/spack/spack/bootstrap.py
+++ b/lib/spack/spack/bootstrap.py
@@ -91,6 +91,14 @@ def _try_import_from_store(module, query_spec, query_info=None):
             os.path.join(candidate_spec.prefix, pkg.platlib),
         ]  # type: list[str]
         path_before = list(sys.path)
+
+        # Python 3.8+ on Windows does not search dependent DLLs in PATH,
+        # so we need to manually add it using os.add_dll_directory
+        # https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+        if sys.version_info[:2] >= (3, 8) and sys.platform == "win32":
+            if os.path.isdir(candidate_spec.prefix.bin):
+                os.add_dll_directory(candidate_spec.prefix.bin)  # novermin
+
         # NOTE: try module_paths first and last, last allows an existing version in path
         # to be picked up and used, possibly depending on something in the store, first
         # allows the bootstrap version to work when an incompatible version is in
--- a/var/spack/repos/builtin/packages/clingo/package.py
+++ b/var/spack/repos/builtin/packages/clingo/package.py
@@ -64,6 +64,8 @@ class Clingo(CMakePackage):
         depends_on("py-cffi", type=("build", "run"), when="@5.5.0: platform=cray")
 
     patch("python38.patch", when="@5.3:5.4.0")
+    patch("size-t.patch", when="platform=windows")
+    patch("vs2022.patch", when="%msvc@19.30:")
 
     def patch(self):
         # Doxygen is optional but can't be disabled with a -D, so patch
--- a/var/spack/repos/builtin/packages/clingo/package.py
+++ b/var/spack/repos/builtin/packages/clingo/package.py
@@ -64,7 +64,7 @@ class Clingo(CMakePackage):
         depends_on("py-cffi", type=("build", "run"), when="@5.5.0: platform=cray")
 
     patch("python38.patch", when="@5.3:5.4.0")
-    patch("size-t.patch", when="platform=windows")
+    patch("size-t.patch", when="%msvc")
     patch("vs2022.patch", when="%msvc@19.30:")
 
     def patch(self):
--- a/lib/spack/spack/util/executable.py
+++ b/lib/spack/spack/util/executable.py
@@ -10,6 +10,7 @@ import subprocess
 import sys
 
 from six import string_types, text_type
+from six.moves import shlex_quote
 
 import llnl.util.tty as tty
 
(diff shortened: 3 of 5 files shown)


lib/spack/spack/bootstrap.py
```diff
@@ -91,6 +91,14 @@ def _try_import_from_store(module, query_spec, query_info=None):
             os.path.join(candidate_spec.prefix, pkg.platlib),
         ]  # type: list[str]
         path_before = list(sys.path)
+
+        # Python 3.8+ on Windows does not search dependent DLLs in PATH,
+        # so we need to manually add it using os.add_dll_directory
+        # https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+        if sys.version_info[:2] >= (3, 8) and sys.platform == "win32":
```
johnwparent, line 98: What happens if this is removed?

var/spack/repos/builtin/packages/clingo/package.py
```diff
@@ -64,6 +64,8 @@ class Clingo(CMakePackage):
         depends_on("py-cffi", type=("build", "run"), when="@5.5.0: platform=cray")
 
     patch("python38.patch", when="@5.3:5.4.0")
+    patch("size-t.patch", when="platform=windows")
```
johnwparent: Is this compiler dependent?

lib/spack/spack/bootstrap.py
KiruyaMomochi: We will get `AttributeError: module 'os' has no attribute 'add_dll_directory'` on (Linux/macOS) or (Windows Python 3.7-)

johnwparent: Sorry, the comment was referring not to the if statement explicitly but rather the entire stanza.

var/spack/repos/builtin/packages/clingo/package.py
KiruyaMomochi: `ssize_t` is defined in `<unistd.h>`, so this should be compiler dependent.
Should I change it to something like `when="%msvc"`?

johnwparent: That would be preferable yes. Leaving the conditional so open ended may result in inscrutable errors down the line.

lib/spack/spack/bootstrap.py
KiruyaMomochi: After these lines are removed, Spack bootstrap will fail because it can't `import clingo` correctly. The debug log shows `Warning: spec <a long spec> did not provide module clingo`.

johnwparent: Is that from a git ref that includes the recent introduction of rpath support to Spack on Windows?

KiruyaMomochi: Yes.

Python 3.8+ searches clingo.dll under the directory of `clingo.cp310-win_amd64.pyd`, which is `<prefix>/Lib/site-packages`. However, this DLL is actually under `<prefix>/bin`.

The RPATH support only added DLLs under `<prefix>/bin` and `<prefix>/Lib`. Both of them are not searched by Python. Besides, clingo.dll is not symlinked, maybe because this is the build result, not a dependency.

Linux doesn't have this problem, because they have RPATH in the pyd file. If we don't want to solve it by `os.add_dll_directory`, we may need to symlink that `bin/clingo.dll` into `Lib/site-packages`, but I'm not sure if this symlink way is a proper way or not.

var/spack/repos/builtin/packages/clingo/package.py
KiruyaMomochi: Modified in the new commit `408f415`.

lib/spack/spack/bootstrap.py
johnwparent: #33400 resolves this issue using Spack's windows RPath support. RPath on Windows uses symlinking behind the scenes to achieve this, so the symlink methodology you suggested would have been correct.

scheibelp (approved):

scheibelp:
Apologies that it took so long to merge this. Thanks for the PR!

lib/spack/spack/bootstrap.py
KiruyaMomochi: Sorry but I still didn't figure out how to create a symlink from `prefix/bin/clingo.dll` to `prefix/Lib/site-packages`.
I have tried to add `self.win_rpath.add_internal_links(join_path(self.prefix.lib, "site-packages"))` into `patch` but it does not work.

PR #32950, opened 2022-10-03
Remove CMakePackage.define alias from most packages

alalazo:
The first 4 commits are extracted from #30738 The last one goes through all the packages that are not reworked in #30738 and removes the alias. This is mainly needed to ensure backward compatibility once #30738 is merged, while it changes no behavior before that, and as such it is safe to merge before.

Modifications:
- [x] Do not alias `CMakePackage.define` in recipes, but use `self.define` instead


haampie (approved):

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

PR #32953, opened 2022-10-03
Bug fix for `ca-certificates-mozilla/package.py` to enable `spack install --source`

### LINKED ISSUE #32948: `spack install --source` not working for `ca-certificates-mozilla`
### Steps to reproduce

I am getting an error for `ca-certificates-mozilla` when I am trying to install it with `spack install --source`. I don't care about the sources for this package specifically, rather I am kicking off an install for an entire environment with `spack install --source` and it fails almost immediately in `ca-certificates-mozilla` in the `install` phase. The error is that directory `shared` in the install prefix already exists. This is because of:
https://github.com/spack/spack/blob/5a0f4970df9e97afe22ce4d8f02cf5427da381ae/var/spack/repos/builtin/packages/ca-certificates-mozilla/package.py#L124
I think this needs a test if the directory already exists, which it apparently does when the sources get installed by `spack install --source`.

### Error message

Not needed, stated above and the error and solution are obvious.

```
$ spack debug report
* **Spack:** 0.19.0.dev0 (32598f6cb8a32bf004f8066e28c63b3477ae2e9a)
* **Python:** 3.8.10
* **Platform:** linux-ubuntu20.04-cascadelake
* **Concretizer:** clingo
$ uname -a
Linux c6i-32xlarge-dy-c6i-32xlarge-1 5.13.0-1031-aws #35~20.04.1-Ubuntu SMP Mon Jun 13 22:30:30 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

### General information

climbfuji:
Bug fix in `var/spack/repos/builtin/packages/ca-certificates-mozilla/package.py` to enable `spack install --source`.

Fixes https://github.com/spack/spack/issues/32948

### DIFF
--- a/var/spack/repos/builtin/packages/ca-certificates-mozilla/package.py
+++ b/var/spack/repos/builtin/packages/ca-certificates-mozilla/package.py
@@ -3,8 +3,6 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-import os
-
 from spack.package import *
 
 
@@ -124,6 +122,5 @@ class CaCertificatesMozilla(Package):
     def install(self, spec, prefix):
         share = join_path(prefix, "share")
         # https://github.com/spack/spack/issues/32948
-        if not os.path.isdir(share):
-            mkdir(share)
+        mkdirp(share)
         install("cacert-{0}.pem".format(spec.version), join_path(share, "cacert.pem"))
--- a/var/spack/repos/builtin/packages/ca-certificates-mozilla/package.py
+++ b/var/spack/repos/builtin/packages/ca-certificates-mozilla/package.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import os
+
 from spack.package import *
 
 
@@ -121,5 +123,7 @@ class CaCertificatesMozilla(Package):
     # Install the the pem file as share/cacert.pem
     def install(self, spec, prefix):
         share = join_path(prefix, "share")
-        mkdir(share)
+        # https://github.com/spack/spack/issues/32948
+        if not os.path.isdir(share):
+            mkdir(share)
         install("cacert-{0}.pem".format(spec.version), join_path(share, "cacert.pem"))


haampie:
Wouldn't it be better to put the sources in `<prefix>/.spack/src` or so? Then it's guaranteed not to clash.

var/spack/repos/builtin/packages/ca-certificates-mozilla/package.py
```diff
@@ -121,5 +123,7 @@ def setup_dependent_package(self, module, dep_spec):
     # Install the the pem file as share/cacert.pem
     def install(self, spec, prefix):
         share = join_path(prefix, "share")
-        mkdir(share)
+        # https://github.com/spack/spack/issues/32948
+        if not os.path.isdir(share):
+            mkdir(share)
```
cosmicexplorer: I believe `mkdirp()` from `llnl.util.filesystem` (implicitly imported from `spack.package`) does this check for you!

cosmicexplorer: And @haampie it might avoid having to decide whether to introduce the `<prefix>/.spack/src` convention if that's not used elsewhere yet.

cosmicexplorer: @climbfuji if you do choose to use `mkdirp()` here I would still retain the link to #32948, that's a helpful comment for later readers.

climbfuji: > Wouldn't it be better to put the sources in `<prefix>/.spack/src` or so? Then it's guaranteed not to clash.

Probably, but `spack install --source` doesn't take an argument, that's something for the spack developers to decide/change.

climbfuji: > @climbfuji if you do choose to use `mkdirp()` here I would still retain the link to #32948, that's a helpful comment for later readers.

Thanks, yes, I will do this later today.

haampie:
Ah, I did not realize Spack puts sources in `<prefix>/share/<pkg name>/src`. I suppose fixing `ca-certificates-mozilla` is fine then.

haampie (approved):

climbfuji:
I tried `mkdirp()` and it worked as expected.

cosmicexplorer (approved):
Thanks so much for diving in to fix this!!

haampie:
@spackbot run pipeline

PR #32956, opened 2022-10-03
julia: add latest 1.8.x and 1.6.x releases and update deps, remove deprecated versions

haampie:
- Add new julia 1.6.x and 1.8.x releases
- Remove deprecated versions and rc version
- Bump libuv-julia and set correct upper/lower bounds in julia
- Remove patch that only applied to rc of 1.8.0
- Update other patch


var/spack/repos/builtin/packages/libuv-julia/package.py
```diff
@@ -9,15 +9,13 @@ class LibuvJulia(AutotoolsPackage):
...
 
-    # julia's libuv fork doesn't tag releases, only has release branches, so we
-    # fix commits.
-    version(
-        "1.44.1",
-        sha256="f931e7825702cbb6d07486d92e5436990cf20f91e2b56d6f759822c0f832b13e",
-        url="https://github.com/JuliaLang/libuv/archive/1b2d16477fe1142adea952168d828a066e03ee4c.tar.gz",
-    )
+    # julia's libuv fork doesn't tag (all?) releases, so we fix commits.
+    version("1.44.2", commit="e6f0e4900e195c8352f821abe2b3cffc3089547b")
```
tldahlgren, line 16: This doesn't match what is tagged as the stable commit for this release (https://github.com/JuliaLang/libuv/releases/tag/v1.44.2).

```suggestion
    version("1.44.2", commit="0c1fa696aa502eb749c2c4735005f41ba00a27b8")
```

var/spack/repos/builtin/packages/libuv-julia/package.py
```diff
@@ -9,15 +9,13 @@ class LibuvJulia(AutotoolsPackage):
...
+    version("1.44.1", commit="1b2d16477fe1142adea952168d828a066e03ee4c")
```
tldahlgren, line 17: This also doesn't match the stable version per https://github.com/JuliaLang/libuv/commit/e8b7eb6908a847ffbe6ab2eec7428e43a0aa53a2:

```suggestion
    version("1.44.1", commit="e8b7eb6908a847ffbe6ab2eec7428e43a0aa53a2")
```

var/spack/repos/builtin/packages/libuv-julia/package.py
```diff
@@ -9,15 +9,13 @@ class LibuvJulia(AutotoolsPackage):
...
+    version("1.44.0", commit="3a63bf71de62c64097989254e4f03212e3bf5fc8")
```
tldahlgren: Same here (https://github.com/JuliaLang/libuv/commit/d2bff508457336d808ba7148b33088f6acbfe0a6):

```suggestion
    version("1.44.0", commit="d2bff508457336d808ba7148b33088f6acbfe0a6")
```

tldahlgren (commented):
Confirmed all new `julia` sha256.  Not sure about the `libuv-julia` commits as they don't appear to match those marked "stable" in the repository.

haampie:
@tldahlgren, Julia's libuv fork is messy: when upstream libuv creates a release, Julia creates a new branch in their fork from master and then merges the release into that, then adds a few cherry-picked commits, and then they may or may not tag it themselves... The 1.44.x versions you are referring to are the upstream versions, Julia typically only has branches, not tagged commits.

So, the source of truth I'm using now is:

https://github.com/JuliaLang/julia/blob/v1.8.2/deps/libuv.version
https://github.com/JuliaLang/julia/blob/v1.8.1/deps/libuv.version
https://github.com/JuliaLang/julia/blob/v1.8.0/deps/libuv.version

This contains the actual commit sha.

Edit: ah, so i just realized the versions are indeed off...

tldahlgren (approved):

PR #32960, opened 2022-10-03
lmod change family arch dir to arch dir

simonLeary42:
With lmod I get name conflicts for modules in different architecture because the directory structure is family based.
```
==> Error: Name clashes detected in module files:

file: /modules/spack-0.18.1/share/spack/lmod/linux-ubuntu20.04-x86_64/Core/zlib/1.2.12.lua
spec: zlib@1.2.12%gcc@9.4.0+optimize+pic+shared patches=0d38234 arch=linux-ubuntu20.04-x86_64
spec: zlib@1.2.12%gcc@9.4.0+optimize+pic+shared patches=0d38234 arch=linux-ubuntu20.04-x86_64_v3
spec: zlib@1.2.12%gcc@9.4.0+optimize+pic+shared patches=0d38234 arch=linux-ubuntu20.04-x86_64_v4
```
With this little change it makes directories for each generic architecture.

### DIFF
--- a/lib/spack/spack/modules/lmod.py
+++ b/lib/spack/spack/modules/lmod.py
@@ -239,7 +239,7 @@ class LmodFileLayout(BaseFileLayout):
         if arch_folder_conf:
             # include an arch specific folder between root and filename
             arch_folder = "-".join(
-                [str(self.spec.platform), str(self.spec.os), str(self.spec.target.family)]
+                [str(self.spec.platform), str(self.spec.os), str(self.spec.target)]
             )
             return os.path.join(self.dirname(), arch_folder)
         return self.dirname()


alalazo (changes requested):
This won't allow people to mix software built for different targets, while you can already do what you ask for by configuring different roots in the `modules.yaml`, see https://spack.readthedocs.io/en/latest/module_file_support.html#changing-the-modules-root

simonLeary42:
That would result in a structure like this:
```
$spack/share/spack/modules/x86_64/linux-ubuntu-20.04/x86_64
$spack/share/spack/modules/x86_64_v3/linux-ubuntu-20.04/x86_64
$spack/share/spack/modules/x86_64_v4/linux-ubuntu-20.04/x86_64
```
which is rather ugly.

Why would anyone want to mix targets in a single directory? Can they mix like that in TCL?

bernhardkaindl:
Closing due to lack of response on the review. A different way of doing this has been suggested.
Please reopen if needed still.

bernhardkaindl:
Closing due to lack of response on the review. A different way of doing this has been suggested.
Please reopen if needed still.

simonLeary42:
https://github.com/spack/spack/blob/85966a96b119b0ae5587e47a074f13a0a2819e39/lib/spack/spack/modules/common.py#L635
The TCL modules use the arch dir, so shouldn't the Lmod modules be consistent?

simonLeary42:
@alalazo can we revisit this?

simonLeary42:
One could mix modules from different targets by simply setting `modules:default:arch_folder=false`

PR #32961, opened 2022-10-03
install sources after package install() runs

cosmicexplorer:
### Problem

See the bugfix in #32953. `ca-certificates-mozilla` called `mkdir()` assuming that `<prefix>/share/` would not exist already, which fails if `spack install --source ca-certificates-mozilla` is used, since we currently install package sources into the prefix (at `<prefix>/share/pkg/src/`) *before* installing the package itself. As described in #32948, this means an attempt to install some *different* package with `--source` will fail if any of its transitive deps make the mistake of assuming an empty install prefix in their `install()` method.

### Solution
- Switch installing sources to run *after* the package's `install()` method.
    - `BuildProcessInstaller._install_source()` uses `fs.install_tree()`, which does not break if the target directory exists already.
    - This allows package `install()` methods to use either `mkdir()` or `mkdirp()` without worrying about the implementation detail of when spack installs sources.

### Result
#32948 shouldn't occur for any other packages.

### DIFF
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -34,8 +34,10 @@ import itertools
 import os
 import shutil
 import sys
+import tempfile
 import time
 from collections import defaultdict
+from contextlib import contextmanager
 
 import six
 
@@ -1872,14 +1874,15 @@ class BuildProcessInstaller(object):
             if self.fake:
                 _do_fake_install(self.pkg)
             else:
-                self._real_install()
-
                 # Run package's install() method before installing sources. This means package
                 # install scripts that don't expect <prefix>/share to exist yet won't fail. See
                 # https://github.com/spack/spack/pull/32953 for where installing sources first may
                 # cause breakages.
                 if self.install_source:
-                    self._install_source()
+                    with self._install_source():
+                        self._real_install()
+                else:
+                    self._real_install()
 
             # Stop the timer and save results
             self.timer.stop()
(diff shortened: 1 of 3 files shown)


haampie:
I think this is good, but we should also *warn* (not error) about clashes the other way around, when the source would overwrite something that's installed.

What we could do is reuse the view code that merges prefixes and checks for collisions, let me see if that's easy.

haampie (changes requested):
I think this is good, but we should also warn (not error) about clashes the other way around, when the source would overwrite something that's installed.

What we could do is reuse the view code that merges prefixes and checks for collisions, let me see if that's easy.

Edit: actually, it's not a good idea, it's installing from the stage dir which is not clean after the build. So, maybe it's easier to just re-stage in the `<prefix>/share/<name>/src` dir.

cosmicexplorer:
> it's installing from the stage dir which is not clean after the build

Ahhhhh, thank you for spotting this!

> So, maybe it's easier to just re-stage in the `<prefix>/share/<name>/src` dir.

By "re-stage" I was thinking of copying over the stage into a temporary directory before installing, then copying it over again. Is there another method you had in mind?

haampie:
Hm yeah, that's probably easiest. But (a) it should use the stage dir not mkdtemp for the temporary copy and (b) it still needs to deal gracefully with conflicts of source dirs/files mapping to installed dirs/files.

With that I'm wondering if this PR is worth it... It's rather unlikely we get file/dir clashes on install when sources are installed first, and also installing sources first avoids the copy (which can be expensive when a multi-GB nvhpc or cuda is part of the dependencies).

cosmicexplorer:
Yeah, I'm not a huge fan of copying the entire source dir, for perf reasons and because I think that might fill up someone's disk quota more than usual?

Especially given that we would still need to figure out a protocol for overlapping file paths as you described, I think this is probably not worth doing. I like the *idea* of being able to assume an empty install prefix, but that abstract benefit doesn't seem to outweigh the practical negatives.

In light of the above, I am closing this for now.

PR #32962, opened 2022-10-03
rdma-core: fix syntax for external discoverability

mpokorny:
Replace `libraries` package attribute with `executables`.

The package is not discoverable using `spack external find` in the original version.


tldahlgren (approved):
Great catch!

PR #28354, opened 2022-01-11
Don't set `LD_LIBRARY_PATH` on package load

haampie:
For me this has been mostly just annoying, cause if I do 
`spack env activate ...` and it sets LD_LIBRARY_PATH, system executables that don't
use rpath will now consider spack-built libs first, and are sometimes
incompatible.

And executables built by Spack already use rpath anyways, so they know
where there libs are.

The only case is maybe when libs are dlopen'ed by say Python, but even
then, I think the libs are generally located because of Python's own
rpaths set by Spack.

Closes #3955 #31731

PR #32964, opened 2022-10-03
py-pyfr: add v1.15.0, explicitly set env variables for dependencies

MichaelLaufer:
Explicit library paths to external libs added due to #28354 .
OpenMP v5.1 compiler support now required for Libxsmm (OpenMP) backend, might be possible to remove this restriction as PyFR compilation is JIT. Either way, an OpenMP virtual dependency would surely be helpful.

#### Also changed:
* `py-gimmik` version updated to v3.0
* `libxsmm` updated main branch name (needed by PyFR until v2 arrives)

@FreddieWitherden

### DIFF
--- a/var/spack/repos/builtin/packages/py-gimmik/package.py
+++ b/var/spack/repos/builtin/packages/py-gimmik/package.py
@@ -17,6 +17,7 @@ class PyGimmik(PythonPackage):
 
     maintainers = ["MichaelLaufer"]
 
+    version("3.0", sha256="45c2da7acff3201b7796ba731e4be7f3b4f39469ff1f1bc0ddf4f19c4a6af010")
     version("2.3", sha256="c019c85316bcf0d5e84de9b7d10127355dfe8037c0e37f1880a9819ce92b74e1")
     version("2.2", sha256="9144640f94aab92f9c5dfcaf16885a79428ab97337cf503a4b2dddeb870f3cf0")
 
--- a/var/spack/repos/builtin/packages/py-gimmik/package.py
+++ b/var/spack/repos/builtin/packages/py-gimmik/package.py
@@ -21,8 +21,8 @@ class PyGimmik(PythonPackage):
     version("2.3", sha256="c019c85316bcf0d5e84de9b7d10127355dfe8037c0e37f1880a9819ce92b74e1")
     version("2.2", sha256="9144640f94aab92f9c5dfcaf16885a79428ab97337cf503a4b2dddeb870f3cf0")
 
-    depends_on("python@3.8", when="@:2.3" type=("build", "run"))
-    depends_on("python@3.9:", when="@3.0:" type=("build", "run"))
+    depends_on("python@3.8", when="@:2.3", type=("build", "run"))
+    depends_on("python@3.9:", when="@3.0:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-numpy@1.7:", type=("build", "run"))
     depends_on("py-mako", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-gimmik/package.py
+++ b/var/spack/repos/builtin/packages/py-gimmik/package.py
@@ -21,7 +21,8 @@ class PyGimmik(PythonPackage):
     version("2.3", sha256="c019c85316bcf0d5e84de9b7d10127355dfe8037c0e37f1880a9819ce92b74e1")
     version("2.2", sha256="9144640f94aab92f9c5dfcaf16885a79428ab97337cf503a4b2dddeb870f3cf0")
 
-    depends_on("python@3.8:", type=("build", "run"))
+    depends_on("python@3.8", when="@:2.3" type=("build", "run"))
+    depends_on("python@3.9:", when="@3.0:" type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-numpy@1.7:", type=("build", "run"))
     depends_on("py-mako", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-pyfr/package.py
+++ b/var/spack/repos/builtin/packages/py-pyfr/package.py
@@ -7,7 +7,7 @@
 from spack.package import *
 
 
-class PyPyfr(PythonPackage):
+class PyPyfr(PythonPackage, CudaPackage, ROCmPackage):
     """PyFR is an open-source Python based framework for solving
     advection-diffusion type problems on streaming architectures
     using the Flux Reconstruction approach of Huynh."""
@@ -22,6 +22,10 @@ class PyPyfr(PythonPackage):
     version("master", branch="master")
 
     # pypi releases
+    version(
+        "1.15.0",
+        sha256="6a634b9d32447f45d3c24c9de0ed620a0a0a781be7cc5e57b1c1bf44a4650d8d",
+    )
     version(
         "1.14.0",
         sha256="ebf40ce0896cce9ac802e03fd9430b5be30ea837c31224531a6d5fd68f820766",
@@ -41,8 +45,8 @@ class PyPyfr(PythonPackage):
     # Required dependencies
     depends_on("python@3.9:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
-    depends_on("py-gimmik@2.2:2", type=("build", "run"))
-    depends_on("py-gimmik@2.3:2", when="@1.14.0:", type=("build", "run"))
+    depends_on("py-gimmik@2.3:2", when="@:1.14.0", type=("build", "run"))
+    depends_on("py-gimmik@3:", when="@:1.15.0", type=("build", "run"))
     depends_on("py-h5py@2.10:", type=("build", "run"))
     depends_on("py-mako@1.0.0:", type=("build", "run"))
     depends_on("py-mpi4py@3.1.0:", type=("build", "run"))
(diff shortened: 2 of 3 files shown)


tldahlgren (dismissed):
Confirmed new sha256 and the renamed branch (`libxsmm`) exists.

var/spack/repos/builtin/packages/py-pyfr/package.py
```diff
@@ -52,7 +56,48 @@ class PyPyfr(PythonPackage):
 
     # Optional dependencies
     depends_on("py-scipy", when="+scipy", type=("build", "run"))
+    depends_on("metis@5.0:", when="+metis", type=("run"))
     depends_on("scotch@6.0:", when="+scotch", type=("run"))
-    depends_on("cuda@8.0:", when="+cuda", type=("run"))
-    depends_on("rocblas@4.5.0:", when="+hip", type=("run"))
-    depends_on("libxsmm@1.18:+shared blas=0 +large_jit_buffer", when="+libxsmm", type=("run"))
+    depends_on("cuda@8.0: +allow-unsupported-compilers", when="+cuda", type=("run"))
```
FreddieWitherden: I think we require CUDA 11.4 greater for the UUID-query functions.

var/spack/repos/builtin/packages/py-pyfr/package.py
```diff
@@ -52,7 +56,48 @@ class PyPyfr(PythonPackage):
...
+    depends_on("rocblas@5.2.0:", when="+hip", type=("run"))
+    depends_on("libxsmm@1.18:+shared blas=0", when="+libxsmm", type=("run"))
+
+    # Conflicts for compilers not supporting OpenMP 5.1+ from v1.15.0:
+    conflicts("%gcc@:11", when="@1.15.0: +libxsmm", msg="OpenMP 5.1+ supported compiler required!")
+
+    # Explicitly set library path for dependencies
+    def setup_run_environment(self, env):
```
FreddieWitherden, line 70: If these libraries are installed somewhere known to the loader we may be able to get away without the exports.

MichaelLaufer: @FreddieWitherden These are installed in Spack installation paths and are no longer found without this from my testing.

var/spack/repos/builtin/packages/py-gimmik/package.py
```diff
@@ -17,6 +17,7 @@ class PyGimmik(PythonPackage):
 
     maintainers = ["MichaelLaufer"]
 
+    version("3.0", sha256="45c2da7acff3201b7796ba731e4be7f3b4f39469ff1f1bc0ddf4f19c4a6af010")
     version("2.3", sha256="c019c85316bcf0d5e84de9b7d10127355dfe8037c0e37f1880a9819ce92b74e1")
     version("2.2", sha256="9144640f94aab92f9c5dfcaf16885a79428ab97337cf503a4b2dddeb870f3cf0")
```
adamjstewart, line 23: python 3.9+ is required now

var/spack/repos/builtin/packages/py-pyfr/package.py
MichaelLaufer: Thanks! Missed that one.

var/spack/repos/builtin/packages/py-pyfr/package.py
```diff
@@ -41,8 +45,8 @@ class PyPyfr(PythonPackage):
     # Required dependencies
     depends_on("python@3.9:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
-    depends_on("py-gimmik@2.2:2", type=("build", "run"))
-    depends_on("py-gimmik@2.3:2", when="@1.14.0:", type=("build", "run"))
+    depends_on("py-gimmik@2.3:2", when="@:1.14.0", type=("build", "run"))
+    depends_on("py-gimmik@3:", when="@1.15.0:", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-gimmik@3", when="@1.15.0:", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-pyfr/package.py
```diff
@@ -52,7 +56,48 @@ class PyPyfr(PythonPackage):
...
+    # Conflicts for compilers not supporting OpenMP 5.1+ from v1.15.0:
+    conflicts("%gcc@:11", when="@1.15.0: +libxsmm", msg="OpenMP 5.1+ supported compiler required!")
+
+    # Explicitly set library path for dependencies
+    def setup_run_environment(self, env):
+        if "+metis" in self.spec:
+            lib_path = find_libraries(
+                    "libmetis", root=self.spec["metis"].prefix.lib, recursive=False
+                )
+            env.set("PYFR_METIS_LIBRARY_PATH", lib_path[0])
```
adamjstewart:
```suggestion
            env.set("PYFR_METIS_LIBRARY_PATH", self.spec["metis"].libs.directories[0])
```
Packages should always ask other packages where their libraries are installed. Same for all the below packages too.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pyfr/package.py
MichaelLaufer: done

var/spack/repos/builtin/packages/py-pyfr/package.py
MichaelLaufer: You mean not to assume that they are in a subdir "lib"? Thanks

var/spack/repos/builtin/packages/py-gimmik/package.py
MichaelLaufer: done

var/spack/repos/builtin/packages/py-pyfr/package.py
MichaelLaufer: done

FreddieWitherden:
Is anything special needed to get parallel HDF5?  This is an option PyFR will take advantage of if available but requires h5py and hdf5 to have support for MPI.

adamjstewart:
@FreddieWitherden this would look like:
```console
$ spack install py-pyfr ^hdf5+mpi
```

var/spack/repos/builtin/packages/py-pyfr/package.py
MichaelLaufer: done

MichaelLaufer:
@FreddieWitherden Regarding parallel HDF5, currently the spack concretizer will choose a parallel HDF5 configuration by default (based off the default `+mpi` variant in py-h5py. I can make this explicit in the spack package if you'd like.
My goal was to simplify this package as much as possible, in order to achieve a configuration that most closely matches the suggested one in the PyFR performance tuning guide, without having to specify any special variants.

FreddieWitherden:
PyFR also supports `PYFR_LIBRARY_PATH` for specifying search directories:
```python
dirs = ["hip", "rocblas", ...]
env.set("PYFR_LIBRARY_PATH",
        ":".join(self.spec[d].libs.directories[0] for d in dirs))
```

var/spack/repos/builtin/packages/py-pyfr/package.py
```diff
@@ -52,7 +56,49 @@ class PyPyfr(PythonPackage):
...
+
+    # Conflicts for compilers not supporting OpenMP 5.1+ from v1.15.0:
+    conflicts("%gcc@:11", when="@1.15.0: +libxsmm", msg="OpenMP 5.1+ supported compiler required!")
+
+    # Explicitly set library path for dependencies
+    def setup_run_environment(self, env):
+        if "+metis" in self.spec:
+            lib_path = join_path(
+                self.spec["metis"].libs.directories[0], "libmetis.{0}".format(dso_suffix)
+            )
```
adamjstewart: If you want the name of the library and not the directory, that's even easier:
```suggestion
            lib_path = self.spec["metis"].libs[0]
```
Same below

MichaelLaufer: @adamjstewart @FreddieWitherden 
I think this is a good solution:
https://github.com/spack/spack/blob/639f7ec42ad672629364990bdc20c36cb301cedb/var/spack/repos/builtin/packages/py-pyfr/package.py#L69-L76

var/spack/repos/builtin/packages/py-pyfr/package.py
```diff
@@ -52,7 +56,21 @@ class PyPyfr(PythonPackage):
...
+    conflicts("%gcc@:11", when="@1.15.0: +libxsmm", msg="OpenMP 5.1+ supported compiler required!")
+
+    # Explicitly add dependencies to PYFR_LIBRARY_PATH environment variable
+    def setup_run_environment(self, env):
+        deps = ["metis", "scotch", "libxsmm", "cuda", "hip", "rocblas"]
+        pyfr_library_path = []
+        for dep in deps:
+            if "+{}".format(dep) in self.spec:
+                pyfr_library_path.extend(self.spec[dep].libs.directories)
+        env.set("PYFR_LIBRARY_PATH", ":".join(pyfr_library_path))
```
adamjstewart, line 76: This may also work, and doesn't require you to explicitly list which packages/variants to check:
```suggestion
        pyfr_library_path = []
        for dep in self.spec.dependencies(deptype="link"):
            pyfr_library_path.extend(self.spec[dep.name].libs.directories)
        env.set("PYFR_LIBRARY_PATH", ":".join(pyfr_library_path))
```
This will recursively list all link dependencies, not sure if there's a way to restrict it to direct dependencies.

adamjstewart (approved):

FreddieWitherden (approved):

var/spack/repos/builtin/packages/py-pyfr/package.py
MichaelLaufer: Thanks @adamjstewart, I considered this but feel that verbose is better in this case. 
I think that it is ready to merge.

adamjstewart (approved):

PR #32969, opened 2022-10-03
acts: new versions

wdconinc:
In the **20.x release line**, these are the changes, https://github.com/acts-project/acts/compare/v20.0.0...v20.3.0
- [x] `option(ACTS_SETUP_ACTSVG "Build ActSVG display plugin" OFF)` introduced in v20.1.0
- [x] `option(ACTS_USE_SYSTEM_ACTSVG "Use the ActSVG system library" OFF)` introduced in v20.1.0
- [x] `option(ACTS_BUILD_PLUGIN_ACTSVG "Build SVG display plugin" OFF)` introduced in v20.1.0
- [x] `option(ACTS_USE_EXAMPLES_TBB "Use Threading Building Blocks library in examples" ON)` introduced in v20.1.0
- [ ] ~~`option(ACTS_EXATRKX_ENABLE_ONNX "Build the Onnx backend for the exatrkx plugin" OFF)` introduced in v20.2.0~~
  - missing dependencies `OnnxRuntime` and `cugraph`
- [ ] ~~`option(ACTS_EXATRKX_ENABLE_TORCH "Build the torchscript backend for the exatrkx plugin" ON)` introduced in v20.2.0~~
  - missing dependencies `TorchScatter`
- [x] `option(ACTS_USE_SYSTEM_LIBS "Use system libraries by default" OFF)` introduced in v20.3.0
- [x] `option(ACTS_USE_SYSTEM_ACTSDD4HEP "Use the ActsDD4hep glue library..." OFF)` removed in v20.3.0

In the **19.x release line**, these are the changes: https://github.com/acts-project/acts/compare/v19.7.0...v19.10.0
- [x] `option(ACTS_USE_EXAMPLES_TBB "Use Threading Building Blocks library in examples" ON)` introduced in v19.8.0
- [ ] ~~`option(ACTS_EXATRKX_ENABLE_ONNX "Build the Onnx backend for the exatrkx plugin" OFF)` introduced in v19.10.0~~
  - missing dependencies `OnnxRuntime` and `cugraph`
- [ ] ~~`option(ACTS_EXATRKX_ENABLE_TORCH "Build the torchscript backend for the exatrkx plugin" ON)` introduced in v19.10.0~~
  - missing dependencies `TorchScatter`
- [ ] ~~`option(ACTS_USE_SYSTEM_LIBS "Use system libraries by default" OFF)` introduced in v19.10.0~~
  - for simplicity sake in the package.py, choosing to ignore this for the 19.x backport series

### DIFF
--- a/var/spack/repos/builtin/packages/acts/package.py
+++ b/var/spack/repos/builtin/packages/acts/package.py
@@ -39,7 +39,11 @@ class Acts(CMakePackage, CudaPackage):
     # Supported Acts versions
     version("main", branch="main")
     version("master", branch="main", deprecated=True)  # For compatibility
+    version("20.2.0", commit="7750c1d24714314e8de716b92ebcd4a92cc4e303", submodules=True)
+    version("20.1.0", commit="be36226fb1be88d7be7c9b17a1c1f6e76ff0e006", submodules=True)
     version("20.0.0", commit="3740e6cdbfb1f75d8e481686acdfa5b16d3c41a3", submodules=True)
+    version("19.9.0", commit="b655e18929ae0ccb6926d8e217b1b3fc02978d35", submodules=True)
+    version("19.8.0", commit="7582072dbaa70802264f20b392de4313afd25667", submodules=True)
     version("19.7.0", commit="03cf7a3ae74b632b3f89416dc27cc993c9ae4628", submodules=True)
     version("19.6.0", commit="333082914e6a51b381abc1cf52856829e3eb7890", submodules=True)
     version("19.5.0", commit="bf9f0270eadd8e78d283557b7c9070b80dece4a7", submodules=True)
--- a/var/spack/repos/builtin/packages/acts/package.py
+++ b/var/spack/repos/builtin/packages/acts/package.py
@@ -366,7 +366,7 @@ class Acts(CMakePackage, CudaPackage):
             if cuda_arch != "none":
                 args.append("-DCUDA_FLAGS=-arch=sm_{0}".format(cuda_arch[0]))
 
-        if spec.satisfies("@19 +dd4hep"):
+        if spec.satisfies("@19:20.2 +dd4hep"):
             args.append("-DACTS_USE_SYSTEM_ACTSDD4HEP=ON")
 
         if spec.satisfies("@:16"):
--- a/var/spack/repos/builtin/packages/acts/package.py
+++ b/var/spack/repos/builtin/packages/acts/package.py
@@ -201,18 +201,6 @@ class Acts(CMakePackage, CudaPackage):
         description="Build the SyCL plugin",
         when="@1:",
     )
-    variant(
-        "tbb",
-        default=True,
-        description="Build the examples with Threading Building Blocks library",
-        when="@19.8:19 +examples",
-    )
-    variant(
-        "tbb",
-        default=True,
-        description="Build the examples with Threading Building Blocks library",
-        when="@20.1: +examples",
-    )
     variant("tgeo", default=False, description="Build the TGeo plugin", when="+identification")
 
     # Variants that only affect Acts examples for now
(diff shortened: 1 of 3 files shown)


HadrienG2:
The ACTS_SETUP_ACTSVG build option does not seem to do anything. I opened https://github.com/acts-project/acts/issues/1571 about this.

On the actsvg side, you may want to add explicit dependencies on boost when +examples and on googletest when either +examples or self.run_tests, instead on relying on the vendored versions.

> not implementing this support as it requires something like when=@19.8:19,20.1: which is AFAIK not possible

I think I remember it is possible to do this by duplicating the directive which the when= applies to (one with when=@19.8:19 and one with when=@20.1:). But I might be misremembering. Feel free to try it out and see if it works out.

Those little things aside, thanks for this contribution! Please ping me when it is ready for review as I'm not sure if Github will automatically do so.

var/spack/repos/builtin/packages/acts/package.py
```diff
@@ -353,6 +367,10 @@ def plugin_cmake_variant(plugin_name, spack_variant):
         if spec.satisfies("@18: +python"):
             args.append("-DACTS_USE_SYSTEM_PYBIND11=ON")
 
+        if spec.satisfies("@20.1: +svg"):
+            args.append("-DACTS_SETUP_ACTSVG=ON")
```
wdconinc: Removing `ACTS_SETUP_ACTSVG` (as removed in https://github.com/acts-project/acts/pull/1573)
```suggestion
```

var/spack/repos/builtin/packages/acts/package.py
```diff
@@ -228,10 +234,29 @@ class Acts(CMakePackage, CudaPackage):
...
+        default=True,
+        description="Build the examples with Threading Building Blocks library",
+        when="@19.8:19 +examples",
+    )
+    variant(
+        "tbb",
+        default=True,
+        description="Build the examples with Threading Building Blocks library",
+        when="@20.1: +examples",
+    )
```
michaelkuhn: Can't this be merged into one variant with `when="@19.8:19,20.1: +examples"`?

Edit: Sorry, just read the corresponding comment. Do we already have an issue for that? Sounds like a bug to me.

wdconinc: Indeed, this doesn't work. Even though the multiple variant when conditions are represented like that in `spack info`. E.g. `spack info acts` has the line
```console
Variants:
    Name [Default]                 When                              Allowed values          Description
    ===========================    ==============================    ====================    =========================================================

    examples [off]                 [@17:+fatras+identification+js    on, off                 Build the examples
                                   on+tgeo, @0.23:16+digitization
                                   +fatras+identification+json+tg
                                   eo]
```
for the two defined variants for `examples`.

Following this representation logically, it would probably need to be `when="@19.8:19 +examples, @20.1: +examples"`. But this fails. Even the simpler `when="@19.8:19, @20.1:"` fails with
```console
==> [2022-10-16-15:45:45.149817] SpecParseError: Invalid version specifier
==> [2022-10-16-15:45:45.149877] Error: Invalid version specifier
  Encountered when parsing spec:
    @19.8:19,@20.1:
             ^
  File "/home/wdconinc/git/spack/lib/spack/spack/spec.py", line 5164, in do_parse
    specs.append(self.spec(None))
  File "/home/wdconinc/git/spack/lib/spack/spack/spec.py", line 5269, in spec
    vlist = self.version_list()
  File "/home/wdconinc/git/spack/lib/spack/spack/spec.py", line 5364, in version_list
    vlist.append(self.version())
  File "/home/wdconinc/git/spack/lib/spack/spack/spec.py", line 5352, in version
    self.next_token_error("Invalid version specifier")
  File "/home/wdconinc/git/spack/lib/spack/spack/parse.py", line 129, in next_token_error
    raise ParseError(message, self.text[0], self.token.end)
spack.parse.ParseError: Invalid version specifier
```

I cannot find an issue about this. I think it's a pretty major change, though, since it adds an operator to the spec grammar and it needs to behave 'as expected'. Case in point: should it be `@19.8:19,20.1: +examples` (operator has high precedence and only works inside versions), or `@19.8:19,@20.1: +examples` (operator has lower precedence), or `@19.8:19 +examples, @20.1: +examples` (lowest precedence)?

wdconinc: Hmm, so it does seem to work when using the highest precedence comma operator... but only if the other variants are shared as they are in this case.

michaelkuhn (approved):
LGTM, but waiting for @HadrienG2 before merging

HadrienG2 (approved):

PR #32970, opened 2022-10-04
PythonPackage: fix libs/headers attributes

adamjstewart:
Fixes a bug that was introduced in #28527. See https://github.com/spack/spack/pull/28527#issuecomment-1172046128 for a detailed breakdown of the problem, and https://github.com/spack/spack/pull/32835#issuecomment-1266228744 for the proposed fix. Still curious why the previous code didn't work, and didn't error, but I'm happy with the solution.

Also fixes a few instances where we were manually searching for this directory instead of relying on this attribute: https://github.com/spack/spack/pull/32835#issuecomment-1266165295

Reverts #31446
Closes #32835

@cosmicexplorer @alalazo

### DIFF
--- a/var/spack/repos/builtin/packages/precice/package.py
+++ b/var/spack/repos/builtin/packages/precice/package.py
@@ -153,9 +153,7 @@ class Precice(CMakePackage):
         if "+python" in spec:
             python_library = spec["python"].libs[0]
             python_include = spec["python"].headers.directories[0]
-            numpy_include = join_path(
-                spec["py-numpy"].prefix, spec["python"].package.platlib, "numpy", "core", "include"
-            )
+            numpy_include = spec["py-numpy"].headers.directories[0]
             if xsdk_mode:
                 cmake_args.append("-DTPL_ENABLE_PYTHON:BOOL=ON")
             else:
--- a/var/spack/repos/builtin/packages/precice/package.py
+++ b/var/spack/repos/builtin/packages/precice/package.py
@@ -153,7 +153,9 @@ class Precice(CMakePackage):
         if "+python" in spec:
             python_library = spec["python"].libs[0]
             python_include = spec["python"].headers.directories[0]
-            numpy_include = spec["py-numpy"].headers.directories[0]
+            numpy_include = join_path(
+                spec["py-numpy"].prefix, spec["python"].package.platlib, "numpy", "core", "include"
+            )
             if xsdk_mode:
                 cmake_args.append("-DTPL_ENABLE_PYTHON:BOOL=ON")
             else:
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -243,8 +243,8 @@ class PythonPackage(PackageBase):
         """Discover header files in platlib."""
 
         # Headers may be in either location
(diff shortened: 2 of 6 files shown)


RemiLacroix-IDRIS (dismissed):
LGTM.

var/spack/repos/builtin/packages/precice/package.py
```diff
@@ -153,9 +153,7 @@ def variant_bool(feature, on="ON", off="OFF"):
         if "+python" in spec:
             python_library = spec["python"].libs[0]
             python_include = spec["python"].headers.directories[0]
-            numpy_include = join_path(
-                spec["py-numpy"].prefix, spec["python"].package.platlib, "numpy", "core", "include"
-            )
+            numpy_include = spec["py-numpy"].headers.directories[0]
```
cosmicexplorer: When I try this branch out with `spack python`, I get the wrong directory. Is there something I'm missing here? I assumed that the `HeaderList` could come out in arbitrary order (although it looks alphabetical below), which is why I was thinking that breaking out a specific method to expose `numpy/core/include` might make sense:
```bash
; spack python                    
Spack version 0.19.0.dev0
Python 3.10.7, Linux x86_64
>>> n = Spec('py-numpy').concretized()
>>> n.package.headers.directories
['/home/cosmicexplorer/tools/s1/opt/spack/linux-alpine3-zen3/gcc-12.1.1/py-numpy-1.23.3-n4htesvvw6nvg5js5ogmoph7l2csm5kr/lib/python3.9/site-packages/numpy/f2py/src', '/home/cosmicexplorer/tools/s1/opt/spack/linux-alpine3-zen3/gcc-12.1.1/py-numpy-1.23.3-n4htesvvw6nvg5js5ogmoph7l2csm5kr/lib/python3.9/site-packages/numpy/core/include', '/home/cosmicexplorer/tools/s1/opt/spack/linux-alpine3-zen3/gcc-12.1.1/py-numpy-1.23.3-n4htesvvw6nvg5js5ogmoph7l2csm5kr/lib/python3.9/site-packages/numpy/f2py/tests/src/module_data']
>>> n.package.headers.directories[0]
'/home/cosmicexplorer/tools/s1/opt/spack/linux-alpine3-zen3/gcc-12.1.1/py-numpy-1.23.3-n4htesvvw6nvg5js5ogmoph7l2csm5kr/lib/python3.9/site-packages/numpy/f2py/src'
>>> n.package.headers.directories[0].split('/')[-3:]
['numpy', 'f2py', 'src']
>>> n.package.headers.directories[1].split('/')[-3:]
['numpy', 'core', 'include']
```

cosmicexplorer (changes requested):

var/spack/repos/builtin/packages/precice/package.py
adamjstewart: Ah, I see. @alalazo has added options to the `libs` property (e.g., `hdf5:cxx`) for other packages. I wonder if we could do the same thing here. Maybe `numpy:core` and `numpy:f2py`.

adamjstewart: Decided to just undo this change and leave the numpy stuff to someone else.

alalazo (approved):

PR #32491, opened 2022-09-02
Match protobuf to py-protobuf version

bvanessen:
Fixed the py-protobuf recipe so that when cpp support is require so
that it uses the same major and minor version range for the protobuf
package.

PR #32977, opened 2022-10-04
Add checksum for py-protobuf 4.21.7, protobuf 21.7; remove protobuf and py-protobuf 2.x

iarspider:


var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -15,6 +15,7 @@ class Protobuf(Package):
     homepage = "https://developers.google.com/protocol-buffers"
     url = "https://github.com/protocolbuffers/protobuf/archive/v3.18.0.tar.gz"
 
+    version("21.7", sha256="75be42bd736f4df6d702a0e4e4d30de9ee40eac024c4b845d17ae4cc831fe4ae")
```
adamjstewart: @bvanessen it looks like you removed 21.1 in #32491, was there a reason for this?

var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -138,7 +139,10 @@ def install(self, spec, prefix):
         args = self.cmake_args()
         args.extend(std_cmake_args)
 
-        source_directory = join_path(self.stage.source_path, "cmake")
+        if self.spec.satisfies("@:3.21.7"):
```
adamjstewart:
```suggestion
        if self.spec.satisfies("@:3"):
```
In case they have more releases in the old branch (they seem to be doing this already)

var/spack/repos/builtin/packages/protobuf/package.py
bvanessen: Because there is a really weird thing that they are doing is having equivalent versions between 21.5 and 21.7 which are equivalent to 3.21.5 and 3.21.7.  I thought that we should be consistent.  So this should be 3.21.7.

adamjstewart: Any idea why they're doing this?

bvanessen: @adamjstewart Nope.  Maybe because the bumped the major version of the python interface changed to version number 4.x

adamjstewart:
@spackbot run pipeline

var/spack/repos/builtin/packages/protobuf/package.py
adamjstewart: @iarspider can you replace this with 3.21.7?

var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -15,6 +15,8 @@ class Protobuf(Package):
     homepage = "https://developers.google.com/protocol-buffers"
     url = "https://github.com/protocolbuffers/protobuf/archive/v3.18.0.tar.gz"
 
+    version("3.21.7", sha256="75be42bd736f4df6d702a0e4e4d30de9ee40eac024c4b845d17ae4cc831fe4ae",
+            url="https://github.com/protocolbuffers/protobuf/releases/download/v21.7/protobuf-all-21.7.tar.gz")
```
adamjstewart:
```suggestion
    version("3.21.7", sha256="ce2fbea3c78147a41b2a922485d283137845303e5e1b6cbd7ece94b96ade7031")
```

var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -138,7 +139,10 @@ def install(self, spec, prefix):
         args = self.cmake_args()
         args.extend(std_cmake_args)
 
-        source_directory = join_path(self.stage.source_path, "cmake")
+        if self.spec.satisfies("@:3"):
+            source_directory = join_path(self.stage.source_path, "cmake")
+        else:
+            source_directory = self.stage.source_path
```
adamjstewart: Is this change still necessary now that we're using 3.21.7?

iarspider: @adamjstewart yes, see the contents of `CMakeLists.txt` in `cmake/`:
> Calling of cmake with source directory set to "cmake" subdirectory of Protocol Buffers project is deprecated. Top-level directory of Protocol Buffers project should be used instead.

var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -138,7 +139,10 @@ def install(self, spec, prefix):
         args = self.cmake_args()
         args.extend(std_cmake_args)
 
-        source_directory = join_path(self.stage.source_path, "cmake")
+        if self.spec.satisfies("@:3"):
```
adamjstewart:
```suggestion
        if self.spec.satisfies("@:3.20"):
```
The root CMakeLists.txt didn't exist until 3.21. Honestly, we should just remove v2.5.0 and convert this to a `CMakePackage` at some point, but that doesn't have to happen in this PR if you don't want to.

var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -133,28 +118,25 @@ def cmake_args(self):
...
 
+    @property
+    def root_cmakelists_dir(self):
+        if self.spec.satisfies("@:3.20"):
+            return join_path(self.stage.source_path, "cmake")
+        else:
+            return self.stage.source_path
+
     @when("@3.0.2:")
     def install(self, spec, prefix):
```
adamjstewart: You should be able to delete this whole `install` method and use the builtin one.

var/spack/repos/builtin/packages/protobuf/package.py
iarspider: Done. I don't know how to properly combine `@when` and `@property`, please correct me if there is a more spack-ish way.

var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -97,12 +88,6 @@ class Protobuf(Package):
     # See https://github.com/protocolbuffers/protobuf/pull/7197
     patch("intel-v2.patch", when="@3.7:3.11.4 %intel")
 
-    patch(
-        "protoc2.5.0_aarch64.patch",
```
adamjstewart, line 101: We can delete this patch file too

var/spack/repos/builtin/packages/py-protobuf/package.py
```diff
@@ -19,6 +19,7 @@ class PyProtobuf(PythonPackage):
 
     variant("cpp", default=False, description="Enable the cpp implementation")
 
+    version("4.21.7", sha256="71d9dba03ed3432c878a801e2ea51e034b0ea01cf3a4344fb60166cb5f6c8757")
     version("4.21.5", sha256="eb1106e87e095628e96884a877a51cdb90087106ee693925ec0a300468a9be3a")
     version(
         "3.20.1",
```
adamjstewart, line 25: You'll probably also have to remove py-protobuf 2

adamjstewart (dismissed):
This looks good to me now, maybe @bvanessen can also review. Would anyone like to be the official maintainer of these packages?

iarspider:
@adamjstewart should I restore `py-protobuf@3.0.0a/b` or drop `py-tensorflow@0.x`?

adamjstewart:
Let's restore them for now but deprecate them. We should also probably deprecate TF 0, but we can think about that in a different PR.

adamjstewart (approved):

iarspider:
@spackbot run pipeline

iarspider:
@spackbot run pipeline

iarspider:
@spackbot run pipeline

PR #32985, opened 2022-10-04
superlu-dist_crayPE update

lukebroskop:
This will apply a patch needed for crayfortran for versions of superlu-dist from 7.1.1 and newer.


tldahlgren (approved):
LGTM

PR #32986, opened 2022-10-04
flecsi: upcoming release will no longer need lanl-cmake-modules

rbberger:


var/spack/repos/builtin/packages/flecsi/package.py
```diff
@@ -125,7 +125,6 @@ class Flecsi(CMakePackage, CudaPackage):
     depends_on("hpx@1.3.0 cxxstd=17 malloc=system", when="backend=hpx @2.0:")
     depends_on("mpich@3.4.1:", when="@2.0: ^mpich")
     depends_on("openmpi@4.1.0:", when="@2.0: ^openmpi")
-    depends_on("lanl-cmake-modules", when="@2.1.1:")
```
tldahlgren, line 128: Would it be more appropriate to retain this with the version range where they are still needed?

tldahlgren (commented):
Have a question.

var/spack/repos/builtin/packages/flecsi/package.py
rbberger: No, because it was never released so far. This dependency was added as a preparation for the next release after `2.1.0`. The next official release will be `2.2.0` and no longer require the extra package.

tldahlgren (approved):
LGTM

PR #32993, opened 2022-10-04
add version 4.2.1 of R

glennpj:


var/spack/repos/builtin/packages/r/package.py
```diff
@@ -79,6 +80,7 @@ class R(AutotoolsPackage):
     depends_on("xz")
     depends_on("which", type=("build", "run"))
     depends_on("zlib@1.2.5:")
+    depends_on("texinfo", type="build")
```
tldahlgren, line 83: Is this relevant to all provided versions or from a given version on?

tldahlgren (approved):
Confirmed the  sha256.  Have a quick question about the new dependency.

var/spack/repos/builtin/packages/r/package.py
glennpj: It is relevant to all versions. I am on the fence about whether it is actually necessary, to be honest. It depends on whether texinfo is installed on the system or not and whether we want to ensure that Spack builds that part of the documentation. There is no configuration option that would allow a variant to be made easily.

PR #32996, opened 2022-10-04
pbbam: switching to meson, adding version 2.1.0

snehring:
At some point pbbam switched fully to meson. The good news is the original version we had packaged also works with meson, so we've got the rare case of switching a build system not being super annoying!

Boost without any variants seems to provide what it requires.

I also winnowed the dependencies a bit.

### DIFF
--- a/var/spack/repos/builtin/packages/pbbam/package.py
+++ b/var/spack/repos/builtin/packages/pbbam/package.py
@@ -4,46 +4,43 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
-from spack.pkg.builtin.boost import Boost
 
 
-class Pbbam(CMakePackage):
+class Pbbam(MesonPackage):
     """The pbbam software package provides components to create, query,
     & edit PacBio BAM files and associated indices.
     These components include a core C++ library,
     bindings for additional languages, and command-line utilities."""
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/pbbam/package.py
```diff
@@ -4,46 +4,43 @@
...
 
     homepage = "https://github.com/PacificBiosciences/pbbam"
-    url = "https://github.com/PacificBiosciences/pbbam/archive/0.18.0.tar.gz"
-
+    url = "https://github.com/PacificBiosciences/pbbam/archive/refs/tags/0.18.0.tar.gz"
+    maintainers = ["snehring"]
+
+    version(
+        "2.1.0",
+        sha256="605944f09654d964ce12c31d67e6766dfb1513f730ef5d4b74829b2b84dd464f",
```
tldahlgren, line 21: Confirmed sha256.

tldahlgren (approved):
FWIW. I am unable to confirm the build for 0.18.0 due to `htslib` build failure.

var/spack/repos/builtin/packages/pbbam/package.py
snehring: I tested the 0.18.0 build on fc36 and gcc 12.2 and on rhel7 with gcc 11.2.0. So we _should_ be good.

PR #32999, opened 2022-10-04
canu: new version 2.2

snehring:
One of the utilities bundled with this failed to build on rhel9 (and newer fedoras) due to the removal of sys/sysctl.h in newer kernels. This is corrected in the latest release.

They've started using submodules for some of the vendored stuff, which I guess isn't picked up well by the default source archive that github generates, so they produce a .tar.xz as part of the release now with everything pulled in.

Also made the boost dep more specific.


tldahlgren (approved):
Confirmed sha256.

snehring:
@spackbot re-run pipeline

PR #33007, opened 2022-10-05
python: don't accidentally link to system db libs

haampie:
Avoid that Python looks for system `ndbm` / `berkeley-db`

Currently Spack builds

```
<prefix>/lib/python3.9/lib-dynload/_dbm.cpython-39-x86_64-linux-gnu.so
```

Which links to system `berkeley-db` on my system


adamjstewart (approved):
Makes sense to me.

haampie:
@spackbot run pipeline

PR #29093, opened 2022-02-20
cleanup spack.parse.Lexer

cosmicexplorer:
### Problem
The `SpecLexer` works fine, but is a little hard to extend, which we may wish to do when expanding the spec syntax as in #24025.

### Solution
- Rewrite the `Lexer` interface to allow navigation between multiple lexer modes by name.
- Also avoid the need to manually write out `lambda scanner, val: ...` in subclasses of `Lexer` for a more declarative interface.

### Result
The spec lexing code should be a little easier to follow. The path to extend the spec syntax as proposed in https://github.com/spack/spack/discussions/20256#discussioncomment-248931 should also be more clear now.

PR #33016, opened 2022-10-05
remove shlex.split() preprocessing from the spec lexer/parser

cosmicexplorer:
*This change is on top of #29093; see the diff against it at https://github.com/cosmicexplorer/spack/compare/arbitrary-lexer-modes...remove-shlex-lex-preprocessing?expand=1.*

### Problem

See discussion at https://github.com/spack/spack/pull/29093#issuecomment-1263714030. We currently call `shlex.split()` before lexing spec strings, and this is the main reason why calling `str()` on a spec instance often produces an unparseable string. We would like to be able to call e.g. `Spec(str(Spec('a b="c d"')))` and get the same spec. Unfortunately, on `develop` that produces the following:
```python
>>> Spec(str(Spec('a b="c d"')))
  File "<console>", line 1, in <module>
  File "/home/cosmicexplorer/tools/s1/lib/spack/spack/spec.py", line 1262, in __init__
    raise ValueError("More than one spec in string: " + spec_like)
ValueError: More than one spec in string: a b=c d
```

### Solution
- Build on the lexer refactoring from #29093 to define many new lexer modes to handle every part of the input string without relying on `shlex.split()`.
- Remove the call to `shlex.split()` in our spec parsing entry point, and move the work of interpreting quotes to the lexer.

### Result
```python
>>> Spec(str(Spec('a b="c d"')))
a b='c d'
```

As discussed in https://github.com/spack/spack/pull/29093#issuecomment-1267942948, this change induces a minor performance regression (this PR even more so than #29093). The regression seems very slight, so I'd like to ignore it, but we'd need to get consensus that this change is enough of a UX improvement to justify that.

It also may be possible to optimize this further; I have not focused on optimizing this approach at all yet.

### DIFF
--- a/lib/spack/spack/parse.py
+++ b/lib/spack/spack/parse.py
@@ -8,6 +8,7 @@ import re
 import shlex
 import sys
 from textwrap import dedent
+from typing import Callable, Dict, List, Tuple
 
 from six import string_types
 
@@ -39,19 +40,23 @@ class Token(object):
         return (self.type == other.type) and (self.value == other.value)
 
 
+_Lexicon = List['Tuple[str, Callable]']
+_Switches = Dict[str, List]
+
+
 class Lexer(object):
     """Base class for Lexers that keep track of line numbers."""
 
     __slots__ = "scanners", "mode", "switchbook"
 
     def __init__(self, lexicon_and_mode_switches):
-        self.scanners = []
-        self.switchbook = []
-        for lexicon, mode_switches_dict in lexicon_and_mode_switches:
-            self.scanners.append(re.Scanner(lexicon))
-            self.switchbook.append(mode_switches_dict)
-
-        self.mode = 0
+        # type: (List[Tuple[str, _Lexicon, _Switches]]) -> None
+        self.scanners = {}      # type: Dict[str, _Lexicon]
+        self.switchbook = {}    # type: Dict[str, _Switches]
+        self.mode = lexicon_and_mode_switches[0][0]
+        for mode_name, lexicon, mode_switches_dict in lexicon_and_mode_switches:
+            self.scanners[mode_name] = re.Scanner(lexicon)  # type: ignore[attr-defined]
+            self.switchbook[mode_name] = mode_switches_dict
 
     def token(self, type, value=''):
         cur_scanner = self.scanners[self.mode]
--- a/lib/spack/spack/parse.py
+++ b/lib/spack/spack/parse.py
@@ -6,7 +6,6 @@
 import abc
 import itertools
 import re
-import shlex
 import sys
 from textwrap import dedent
 from typing import Any, Dict, Iterable, Iterator, List, Tuple
@@ -16,7 +15,6 @@ import six
 import llnl.util.tty as tty
 
 import spack.error
-import spack.util.path as sp
 
 
 class Token(object):
@@ -61,7 +59,7 @@ class Lexer(object):
         self.scanners = {}  # type: Dict[str, re.Scanner] # type: ignore[name-defined]
         self.switchbook = {}  # type: Dict[str, Dict[str, List[Any]]]
         self.original_mode = lexicon_and_mode_switches[0][0]  # type: str
-        self.mode = self.original_mode                        # type: str
+        self.mode = self.original_mode  # type: str
 
         # Convert the static description of the token id into a callback that executes
         # self.token(...) when the pattern is recognized. This particular construction is required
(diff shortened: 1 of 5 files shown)


cosmicexplorer:
cc @psakievich @trws: would love your input on any further annoyances with spec parsing that I may have missed. I kinda punted on trying to interpret inner quotes, so any examples of that would be useful as well. My goal is to demonstrate that this approach can be extended to solve lingering UX issues with spec parsing.

cosmicexplorer:
@trws noted that we still join specs from the command line with `" ".join()` in `cmd/spec.py`, and proposed some inputs that we'd like to ensure don't get clobbered from the initial spec parsing:
> on the command line: `umpire%clang@9.0.0 cxxflags="-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0"`
> on the command line: `'umpire%clang@9.0.0 cxxflags="-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0"'`
> those two should be equivalent
> on the command line: `umpire%clang@9.0.0 "cxxflags=-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0"`
> that one too
> those should work the same if specified as a string inside spack as well
> and should be equivalent to `["umpire%clang@9.0.0", "cxxflags=-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0"]`

I'll make test cases out of these.

psakievich:
@cosmicexplorer I'm a little late at getting up to speed on all your recent work. Looks good so far from what I can tell.  I see you've absorbed the `COLON` token into the version regex.  IIRC the `COMMA` token may be eligible for the same treatment, unless a new use case has been added through these refactors.  I was playing around with this in #32257 and at one point I had removed both tokens ([see](https://github.com/spack/spack/pull/32257/commits/bcce1207263046b307abea82793e2cb87d32879f)). 

It seemed to work with a single string (passed the tests for `spec_syntax.py`, but definitely broke with `shlex` lexing.  Since you are removing that it could be back in play

cosmicexplorer:
@psakievich thank you! I had actually removed the `COMMA` earlier (and indeed, it is not used when parsing versions now), but brought it back to cover multiple-valued variants which may or may not be quoted (e.g. `a='b',c`).  That would be the new use case you mentioned:
> unless a new use case has been added through these refactors

There has definitely been an explosion of lexer modes in this PR (e.g. four modes for different types of single and double quoting), and it would probably improve readability and performance if we could slim those down somewhat, but I think I'm focusing on completing the functionality first (see Tom's spec inputs above).

psakievich:
That makes sense.  I'll try to give this a closer look later today and see if I can come up with any more suggestions.  Personally, I need a quiet room and a security blanket close at hand when dealing with Spack's parser. 😁

lib/spack/spack/parse.py
```diff
@@ -32,66 +36,108 @@ def __str__(self):
...
+
+
 class Lexer(object):
     """Base class for Lexers that keep track of line numbers."""
 
-    __slots__ = "scanner0", "scanner1", "mode", "mode_switches_01", "mode_switches_10"
+    __slots__ = "scanners", "original_mode", "mode", "switchbook"
+
+    def __init__(self, lexicon_and_mode_switches):
+        # type: (List[Tuple[str, _Lexicon, Dict[str, List[Any]]]]) -> None
```
psakievich, line 58: Is there any reason to not make these commented type definitions into actual types? From a code readability stand point trying to follow the construction of this object, and the implementation in spec.py would be simplified if this conglomeration of types were members inside a class.

We could also test each phase individually in spec.py if we instantiated the objects outside the Lexer.

lib/spack/spack/spec.py
```diff
@@ -5071,6 +5225,41 @@ def do_parse(self):
 
         try:
             while self.next:
+                if self.accept(SINGLE_QUOTE):
```
psakievich, line 5228: I've always found this section of the code really difficult to follow because of all the nested if/else's.

I'd suggest any new additions should be functions so we get a better description of what is going on through the name. Maybe something like:

```python
def starts_with_quotes(self, quote_type, specs):
   # same code for single and double

def process_initial_quotes(self, specs):
   # replace both new if statements with this function
   starts_with_quotes(SINGLE_QUOTE, specs)
   starts_with_quotes(DOUBLE_QUOTE, specs)
```

psakievich (commented):
I really like these changes, but I am finding myself getting a little lost in the design.  A couple of thoughts to improve readability and testability of the code here.

alalazo:
Superseded by #34151

PR #33019, opened 2022-10-05
Add checksum for py-urllib3 1.26.12

iarspider:


var/spack/repos/builtin/packages/py-urllib3/package.py
```diff
@@ -28,6 +29,7 @@ class PyUrllib3(PythonPackage):
 
     depends_on("python@2.7:2.8,3.4:", when="@:1.25", type=("build", "run"))
     depends_on("python@2.7:2.8,3.5:", when="@1.26.6", type=("build", "run"))
+    depends_on("python@2.7:2.8,3.6:", when="@1.26.12:", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("python@2.7:2.8,3.6:3", when="@1.26.12:", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-urllib3/package.py
```diff
@@ -28,6 +29,7 @@ class PyUrllib3(PythonPackage):
...
 
     depends_on("py-setuptools", type="build")
```
adamjstewart, line 35: `+secure` has a new dep on `py-urllib3-secure-extra`

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-urllib3/package.py
iarspider: I think we should even remove the `+secure` variant: https://github.com/urllib3/urllib3/issues/2680

adamjstewart: Looks like it's only needed by `py-azure-cli`. You could see if that's still true in the latest release. Or you could copy-n-paste the `+secure` deps from this package to `py-azure-cli`. Or leave it as is and update the dep.

adamjstewart (approved):

alalazo:
@spackbot run pipeline

PR #33021, opened 2022-10-05
Support for Windows development environment

johnwparent:
This establishes a base of support for a Windows development environment and introduces some fundamental Windows tools to Spack as packages and externally locatable components. This will support packages relying on low level Windows development components such as the Windows OpenGL driver, third party GL drivers, the driver development kit, and the Windows software development kit. This components will introduce new support or expand support for such packages as MSMPI, HDF5, Glew, Paraview, and VTK.


lib/spack/llnl/util/symlink.py
```diff
@@ -24,7 +24,7 @@ def symlink(real_path, link_path):
     On Windows, use junctions if os.symlink fails.
     """
     if not is_windows or _win32_can_symlink():
-        os.symlink(real_path, link_path)
+        os.symlink(real_path, link_path, target_is_directory=os.path.isdir(link_path))
```
scheibelp: If we create a link to `x` in location `y`, then I assume we care about whether `x` is a directory, I think right now this is `os.symlink(x, y, os.path.isdir(y))`

lib/spack/spack/detection/common.py
```diff
@@ -235,9 +269,44 @@ def compute_windows_program_path_for_package(pkg):
         return []
     # note windows paths are fine here as this method should only ever be invoked
     # to interact with Windows
-    program_files = "C:\\Program Files{}\\{}"
+    program_files = "{}:\\Program Files{}\\{}"
+    drive_letter = os.environ['WINDIR'][0]
```
scheibelp: (question) above, `drive_letter = os.environ['HOMEDRIVE']` - is this a different way of getting the same thing?

var/spack/repos/builtin/packages/win-sdk/package.py
```diff
@@ -0,0 +1,63 @@
...
+    version("10.0.16299")
+    version("10.0.15063")
+    version("10.0.14393")
+    version("10.0.10586")
+    version("10.0.26639")
+
+    # WinSDK versions depend on compatible compilers
+    # WDK versions do as well, but due to their one to one dep on the SDK
+    # we can ensure that requirment here
+    depends_on("%msvc:", when="@10.")
```
scheibelp: The `:` in `%msvc:` should not be necessary

var/spack/repos/builtin/packages/win-wdk/package.py
```diff
@@ -0,0 +1,73 @@
...
+    for plat in ["linux", "darwin", "cray"]:
+        conflicts("platform=%s" % plat)
+
+    @classmethod
+    def determine_version(cls, lib):
+        """
+        WDK is a set of drivers that we would like to
+        be discoverable externally by Spack.
+        The lib does not provide the WDK
+        version so we derive from the exe path
```
scheibelp: By "exe path" do you mean the full library path? I'm guessing this is examining the parent directories etc. If so, I think calling it the full library path would be clearer (given the definition of `.libraries`, I think we would be examining `ntoskernal.lib`)

lib/spack/spack/operating_systems/windows_os.py
```diff
@@ -16,8 +16,10 @@
 
 
 def windows_version():
-    """temporary workaround to return a Windows version as a Version object"""
-    return Version(platform.release())
+    """ Windows version as a Version object """
+    # include the build number as this provides important information
+    # for low lever packages and components like the SDK and WDK
```
scheibelp, line 20: It might be worth expanding on this slightly like by talking about how the build number is component z of x.y.z.

lib/spack/spack/fetch_strategy.py
```diff
@@ -408,7 +408,9 @@ def _fetch_curl(self, url):
                 partial_file,
             ]  # use a .part file
         else:
-            save_args = ["-O"]
+            # -J flag requires curl >= 7.21.2 ensures file will be named
+            # from redirect header if relevant
+            save_args = ["-O", "-J"]
```
scheibelp: * Is `rename_downloaded_executable` still needed in the `wdk` package with this?
* Currently `bootstrap.py` doesn't appear to check `curl` version, but changes which bump the minimum required version can be problematic. This might be reasonable but IMO we could still use a check to provide a clear error message when `curl` is insufficient (and in case users report incompatibilities, we may want to only conditionally do this on Windows in a follow-up PR)

var/spack/repos/builtin/packages/wgl/package.py
```diff
@@ -0,0 +1,51 @@
...
+    # version of SDK being used
+    depends_on("win-sdk")
+
+    # WGL has no meaning on other platforms, should not be able to spec
+    for plat in ["linux", "darwin", "cray"]:
+        conflicts("platform=%s" % plat)
+
+
+    def determine_version(self, exe):
+        """Allow for WGL to be externally detectable"""
```
scheibelp, line 65: I assume there would be a problem if this returns `None`: is there a way to get a version from the file? Also, this calls the file `exe` when in fact the class defines `libraries` (not strictly an error, but not idiomatic).

var/spack/repos/builtin/packages/win-wdk/package.py
```diff
@@ -0,0 +1,73 @@
...
+        installer = glob.glob(os.path.join(self.stage.source_path, "linkid=**"))
+        if len(installer) > 1:
+            raise RuntimeError("Fetch has failed, unable to determine installer path")
+        installer = installer[0]
+        os.rename(installer, os.path.join(self.stage.source_path, "wdksetup.exe"))
+
+    def install(self, spec, prefix):
+        install_args = ["/features", "+", "/quiet", "/installpath", self.prefix]
+        with working_dir(self.stage.source_path):
+            Executable("wdksetup.exe")(*install_args)
```
scheibelp: You mentioned that users can't install `wdk` if it's already on the system: could the error message be augmented with a suggestion to run `spack external find --not-buildable wdk`?

var/spack/repos/builtin/packages/win-wdk/package.py
```diff
@@ -0,0 +1,73 @@
...
+        os.environ["WDKContentRoot"] = self.prefix
+
+    @run_before("install")
+    def rename_downloaded_executable(self):
+        """WGL download is named by fetch based on name derived from Link redirection
+        This name is not properly formated so that Windows understands it as an executable
+        We rename so as to allow Windows to run the WGL installer"""
+        installer = glob.glob(os.path.join(self.stage.source_path, "linkid=**"))
+        if len(installer) > 1:
+            raise RuntimeError("Fetch has failed, unable to determine installer path")
```
scheibelp: IMO it's worth including the full list of paths so that users can examine the ambiguity with the error message (or maybe that should go into a `tty.debug` statement).

lib/spack/spack/detection/common.py
```diff
@@ -197,10 +200,41 @@ def update_configuration(detected_packages, scope=None, buildable=True):
...
 
+def find_windows_compiler_root_paths():
+    """Helper for Windows compiler installation root discovery
+
+    At the moment simply returns location of VS install paths from VSWhere
+    But should be extended to include more information as relevant"""
+    return list(winOs.WindowsOs.vs_install_paths)
+
+
+def find_windows_compiler_cmake_paths():
```
scheibelp: I think it would be useful to consolidate these `find_windows_..._paths` functions into an object to make the module more readable

scheibelp (changes requested):
I mostly have comments about readability, although there appears to be a couple issues (more detail in comments).

lib/spack/spack/detection/common.py
johnwparent: Yes, these were different iterations of development. I hadn't realized I was using a different env var good catch, I'll consolidate that. I'm also thinking python must have a way to detect current drive letter so perhaps this can be made more robust as well.

var/spack/repos/builtin/packages/win-sdk/package.py
johnwparent: Understood.

lib/spack/spack/fetch_strategy.py
johnwparent: It's actually the other way around. I thought I had removed this with the last rebase, it's no longer needed due to the rename (this approach never actually worked, will be reverting)

var/spack/repos/builtin/packages/win-wdk/package.py
johnwparent: Will do

lib/spack/spack/detection/common.py
johnwparent: Great idea. On it.

var/spack/repos/builtin/packages/win-wdk/package.py
johnwparent: I tested this, and the installer will report that scenario to the users prior to closure without Spack needing to do so. When spack runs `Executable("wdksetup.exe")(*install_args)` execution is handed over to the Windows installer and gui, which then detects other installations of the WDK and reports to that effect. Context is not returned to spack until a user acknowledges that scenario, so I don't believe reporting that would be necessary. Additionally, I don't know if the installer returns enough info to Spack via stdout that we would be able to differentiate that failure from another.

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,224 @@
...
+class WindowsRegistry(object):
+    """
+    Interface to provide access, querying, and searching to Windows registry entries.
+    This class represents a single key entrypoint into the Windows registry
+    and provides an interface to this key's values, its subkeys, and those subkey's values.
+    This class cannot be used to move freely about the registry, only subkeys/values of
+    the root key used to instantiate this class.
+    """
+
+    def __init__(self, key, root_string=winreg.HKEY_CLASSES_ROOT):
```
scheibelp: This default value is forcing a search for `winreg` and I think leading to https://readthedocs.org/projects/spack/builds/18633158/

var/spack/repos/builtin/packages/wgl/package.py
scheibelp: Apologies - my original comment was confusing: I meant that it would be problematic to *always* return `None`. Now that this has an implementation, it is preferable to raise an exception when the version cannot be determined (I assume that this implementation generally successfully extracts a version).

var/spack/repos/builtin/packages/win-wdk/package.py
scheibelp: Now that this PR includes `windows_registry` could this be updated? That doesn't have to happen in this PR, I just wanted to check if that made this possible (I assume that for this to make sense, it would also need to run `-unattended`, if that or an equivalent option is available for the install).

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,259 @@
...
+        used to construct class"""
+        if is_windows:
+            hkey_dict = {
+                "HKEY_CLASSES_ROOT": winreg.HKEY_CLASSES_ROOT,
+                "HKEY_CURRENT_USER": winreg.HKEY_CURRENT_USER,
+                "HKEY_USERS": winreg.HKEY_USERS,
+                "HKEY_LOCAL_MACHINE": winreg.HKEY_LOCAL_MACHINE,
+                "HKEY_CURRENT_CONFIG": winreg.HKEY_CURRENT_CONFIG,
+                "HKEY_PERFORMANCE_DATA": winreg.HKEY_PERFORMANCE_DATA
+            }
```
scheibelp: Is this dictionary needed if you do `getattr(winreg, self._hkey)`?

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,259 @@
...
+    the root key used to instantiate this class.
+    """
+
+    def __init__(self, key, root_key=HKEY_CONSTANT("HKEY_CURRENT_USER")):
+        """Constructs a Windows Registry entrypoint to key provided
+        root_string should be an already open root key or an hkey constant if provided
+
+        Args:
+            key (str): registry key to provide root for registry key for this clas
+            root_string: Already open registry key or HKEY constant to provide access into
```
scheibelp: The parameter is called `root_key` but you refer to it as `root_string` here (in multiple spots).

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,259 @@
...
+            os.path.join(self.path, sub_key),
+            winreg.OpenKeyEx(self.handle, sub_key, access=winreg.KEY_READ),
+        )
+
+    def get_value(self, val_name):
+        """Returns value associated with this key in RegistryValue object"""
+        return RegistryValue(val_name, winreg.QueryValueEx(self.handle, val_name)[0], self.handle)
+
+
+class WindowsRegistry(object):
```
scheibelp: I think it would be clearer to call this `WindowsRegistryQuery` or something like that (`WindowsRegistryEntrypoint`?): when the object is instantiated it is with a single key (so this object is in a sense a limited view into the registry, which makes it seem more like a query to me).

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,262 @@
...
+            return hkey_dict[self._hkey]
+        raise RuntimeError(
+            "HKEY Constants are for Windows registry use"
+            "and should not be reference by other platforms"
+        )
+
+
+# we alias all the HKEY constants to avoid needing to import winreg
+# where we use this registry interface as importing winreg involves the
+# messy import we see above.
```
scheibelp: By "messy import" do you mean 

```
if is_windows:
     import winreg
```

?

I think even with a conditional import, you can refer to `winreg.HKEY_LOCAL_MACHINE` for example in function bodies (as long as it's not a default value for a function parameter, but in that case just set `None` as the default and initialize it in the function body).

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,262 @@
...
+        """Composes all subjeys into a list for access"""
+        if self._keys:
+            return
+        sub_keys, _, _ = winreg.QueryInfoKey(self.handle)
+        for i in range(sub_keys):
+            sub_name = winreg.EnumKey(self.handle, i)
+            sub_handle = winreg.OpenKeyEx(self.handle, sub_name, access=winreg.KEY_READ)
+            self._keys.append(RegistryKey(os.path.join(self.name, sub_name), sub_handle))
+
+    def gather_value_info(self):
```
scheibelp: This function looks like it is used exclusively to support `.values`: for functions which only support other functions on the object, can you indicate that they are private by prefixing the name with `_`?

(likewise for `gather_subkey_info`)

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,262 @@
...
+            try:
+                self.get_value(val_name)
+            except FileNotFoundError as e:
+                if e.winerror == 2:
+                    return None
+                else:
+                    raise e
+        else:
+            key = self._traverse_subkeys(lambda x: val_name in x.values)
+            return key if not key else key.values[val_name]
```
scheibelp: (minor) I think it would be clearer to expand this sort of return out to something like:

```
if not key:
  return None
else:
  return key.values[val_name]
```

lib/spack/spack/detection/common.py
```diff
@@ -104,6 +107,19 @@ def _spec_is_valid(spec):
     return True
 
 
+def path_to_dict(search_paths):
+    """Return dictionary[path]:file from list of paths"""
```
scheibelp: I'm not familiar with this type syntax: what is it? Also I think a clearer description of the return type is `dictionary[fullpath]: basename`

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,262 @@
...
+            recursive (bool): optional argument, if True, subkey need not be a direct
+                                sub key of this registry entry, and this method will
+                                search all subkeys recursively.
+                                Default is True
+        Return:
+            the desired subkey as a RegistryKey object, or none
+        """
+        if not recursive:
+            try:
+                self.get_subkey(subkey_name)
```
scheibelp: This doesn't return anything (likewise for `find_value`)

Also, would it make sense to fold this try/except logic into `get_subkey` (and likewise for `get_value`)?

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,262 @@
...
+        return self._values
+
+    @property
+    def subkeys(self):
+        """Returns list of all subkeys of this key as RegistryKey objects"""
+        self.gather_subkey_info()
+        return self._keys
+
+    def gather_subkey_info(self):
+        """Composes all subjeys into a list for access"""
```
scheibelp: (minor) typo: "subjeys"

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,262 @@
...
+            raise RuntimeError(
+                "Cannot instantiate Windows Registry class on non Windows platforms"
+            )
+        self.key = key
+        self.root = root_key
+        try:
+            self.reg = RegistryKey(
+                os.path.join(str(self.root), self.key),
+                winreg.OpenKeyEx(self.root.hkey, self.key, access=winreg.KEY_READ),
+            )
```
scheibelp, line 183: Where possible, it is best to avoid relying on external services in initializers: could this be initialized with an already-constructed `RegistryKey`? That would make this easier to test (in isolation of the actual registry).

You can add a static function like:

```
class WindowsRegistry()
  @staticmethod
  from_string(key, root_key):
    return WindowsRegistry(RegistryKey(...))
```

scheibelp (changes requested):
Some preliminary comments. Some are more straightforward changes and others might merit discussion. Let me know if you agree.

lib/spack/spack/util/windows_registry.py
johnwparent: I think `WindowsRegistryView` would be best. The registry can only be observed through the context of a single key at a time (unless you're using regedit, but that's because the UI opens all keys at once) beginning at a given HKEY constant key. So this is a view of the windows registry rooted at the key we provide. `Entrypoint` would also work as that's pretty true to what's happening, although I like the read-only implication of `View`.

lib/spack/spack/util/windows_registry.py
johnwparent: I do.

You can do that, but I'd rather not have a lot of conditional imports if we can just use the class abstraction. If you feel strongly about removing that however, that's fine.

lib/spack/spack/util/windows_registry.py
johnwparent: Will do.

lib/spack/spack/util/windows_registry.py
johnwparent: Will do.

lib/spack/spack/util/windows_registry.py
johnwparent: .... I'll fix that.

I had intended this top level class to be more of the actual usable interface and the `RegistryKey` to be more of a raw interface for interacting with the winreg module directly, so I left `get_subkey` and `get_value` to fail in case an interface wanted to handle that case differently at some point.

lib/spack/spack/util/windows_registry.py
johnwparent: Nope!

lib/spack/spack/util/windows_registry.py
johnwparent: Refactor miss. Amended.

var/spack/repos/builtin/packages/win-wdk/package.py
johnwparent: Yes, we would be able to search for the registry key and report the already installed version. Windows installer APIs are not consistent across packages, each package creates their own. The WDK does not have the `-unattended` flag, instead the `-quiet` flag that I've already added handled both the normal interface and any error messages that may crop up. There are two to dos here, the first is the registry stuff, and the second is piping errors to the spack log file. I'll try to do both in this PR.

var/spack/repos/builtin/packages/wgl/package.py
johnwparent: I notice other package implementations of `determine_version` don't raise errors on failure, is there a particular reason we want to here? I'm perfectly fine with including it, I'd just like to understand.

lib/spack/spack/util/windows_registry.py
johnwparent: Fixed.

lib/spack/spack/detection/common.py
johnwparent: I don't know that it's official type syntax, I wanted to represent the actual contents, not type. Fixed as suggested.

var/spack/repos/builtin/packages/win-wdk/package.py
johnwparent: If you point the log at spack's log it blows things away, that will be saved for a later PR.

var/spack/repos/builtin/packages/wgl/package.py
scheibelp: Sorry: I was wrong about this - it's totally fine to return none (in that case, the object under consideration does not produce an associated spec).

lib/spack/spack/compilers/msvc.py
```diff
@@ -90,10 +93,26 @@ def __init__(self, *args, **kwargs):
...
+        Msvc.msvc_version
+        """
+        ver = self.msvc_version
         ver = "".join(ver.split(".")[:2])[:-1]
         return "MSVC" + ver
 
+    @property
+    def cc_version(self):
+        """Cl toolset version"""
+        return spack.compiler.get_compiler_version_output(self.cc)
```
scheibelp, line 114: By default the implementation for this is:

```
    @classmethod
    def default_version(cls, cc):
        """Override just this to override all compiler version functions."""
        output = get_compiler_version_output(
            cc, cls.version_argument, tuple(cls.ignore_version_errors)
        )
        return cls.extract_version_from_output(output)

    @classmethod
    @llnl.util.lang.memoized
    def extract_version_from_output(cls, output):
        """Extracts the version from compiler's output."""
        match = re.search(cls.version_regex, output)
        return match.group(1) if match else "unknown"
```

which is to say: it looks like your implementation is taking all of the output from `cl.exe` (invoked with no arguments) to be the version; I wanted to make sure that I understood that correctly. If I got that right, I think it would be more idiomatic to define a `version_regex = "(.*)"  # Capture all output as the version` class attribute.

I see this appears to be used primarily for 

```
"/p:PlatformToolset=%s" % self.compiler.cc_version
```

I'll note that if the output of `cl.exe` contains more than what would be considered a version description, we might want to store that information in another variable (e.g. `.platform_toolset`)

lib/spack/spack/compilers/msvc.py
```diff
@@ -103,11 +122,23 @@ def setup_custom_environment(self, pkg, env):
...
+        arch = spack.platforms.real_host().default.lower()
+        arch = arch.replace("-", "_")
+        # vcvars can target specific sdk versions, force it to pick up concretized sdk
+        # version, if needed by spec
+        sdk_ver = "" if "win-sdk" not in pkg.spec else pkg.spec["win-sdk"].version.string + ".0"
+        # provide vcvars with msvc version selected by concretization,
+        # not whatever it happens to pick up on the system (highest available version)
         out = subprocess.check_output(  # novermin
-            'cmd /u /c "{}" {} && set'.format(self.setvarsfile, "amd64"),
+            'cmd /u /c "{}" {} {} {}&& set'.format(
```
scheibelp: Should there be a space between the final `{}` and `set`?

lib/spack/spack/detection/common.py
```diff
@@ -158,11 +174,18 @@ def library_prefix(library_dir):
     assert os.path.isdir(library_dir)
 
     components = library_dir.split(os.sep)
-    if "lib64" in components:
-        idx = components.index("lib64")
+    # remove incidences where capital lib/bin dirs throw off prefix
```
scheibelp: (minor) I suggest replacing "remove incidences where capital lib/bin dirs throw off prefix computation" with "convert to lowercase to match lib, LIB, Lib, etc."

lib/spack/spack/detection/common.py
```diff
@@ -233,9 +363,34 @@ def compute_windows_program_path_for_package(pkg):
...
+    https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netframework-4.8"""
+    if not is_windows:
+        return []
+
+    # Current user directory
+    user = os.environ["USERPROFILE"]
+    app_data = "AppData"
+    app_data_locations = ["Local", "Roaming"]
+    user_appdata_install_stubs = [os.path.join(app_data, x) for x in app_data_locations]
+    user_appdata_install_stubs.append("")
```
scheibelp: (minor) I think it would be clearer to say

```
list(itertools.product( user_appdata_install_stubs, (pkg.name, pkg.name.capitalize()))
    + [pkg.name, pkg.name.capitalize()]
```

below and not append this empty string.

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,321 @@
...
+    @property
+    def handle(self):
+        if not self._handle:
+            self._handle = self._get_hkey(self.path)
+        return self._handle
+
+    @handle.setter
+    def handle(self, handle):
+        if not self._handle:
+            self._handle = handle
```
scheibelp: I don't see anything setting `.handle =` other than the superclass initializer - is this needed?

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,321 @@
...
+        return RegistryValue(val_name, winreg.QueryValueEx(self.handle, val_name)[0], self.handle)
+
+
+class _HKEY_CONSTANT(RegistryKey):
+    """Subclass of RegistryKey to represent the prebaked, always open registry HKEY constants"""
+
+    def __init__(self, hkey_constant):
+        hkey_name = hkey_constant
+        self._handle = None
+        # because we can't referene winreg yet, handle is none for now
```
scheibelp: (minor) might be worth mentioning that is because this class is instantiated at module import time (down in `HKEY`), since it is atypical for that to happen (it took me a minute to realize what you meant).

var/spack/repos/builtin/packages/perl/package.py
```diff
@@ -229,7 +229,7 @@ def do_stage(self, mirror_only=False):
     def nmake_arguments(self):
         args = []
         if self.spec.satisfies("%msvc"):
-            args.append("CCTYPE=%s" % self.compiler.msvc_version)
+            args.append("CCTYPE=%s" % self.compiler.short_msvc_version)
```
scheibelp, line 232: Note if you create a version object, it will automatically create separate indexable components based on `.`'s, so IMO it would be more idiomatic to set `msvc.version = Version(...)` and then you can say here `self.compiler.msvc_version[:2]` (or however many components that `perl` wants).

scheibelp (changes requested):
Additional requests - hopefully all minor. Let me know if you disagree.

lib/spack/spack/compilers/msvc.py
johnwparent: The version regex is correct for this scenario. Running just `cl.exe` with no arguments produces a block out output that happens to include the version string, which is successfully picked up by that regex. The other information is not really useful w.r.t. to what Spack cares about. For example running `cl.exe` for MSVC 2019 gives me:
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30145 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]
```

johnwparent: I supposed we could pick up plat arch from there but at this point Spack is well aware of the architecture it's running on.

lib/spack/spack/compilers/msvc.py
johnwparent: It should be `{} && set`. I'll add the space between `{}` and `&&`

lib/spack/spack/detection/common.py
johnwparent: Agreed. Done.

lib/spack/spack/detection/common.py
johnwparent: You're right that is cleaner. Thanks.

lib/spack/spack/util/windows_registry.py
johnwparent: We need the property interface to allow for the delayed loading of the handle, and I believe this setter is required by that initializer.

lib/spack/spack/util/windows_registry.py
johnwparent: Sounds good, I can see how that would be confusing if you're not writing this module. Also I'll fix the spelling in this docstring.

var/spack/repos/builtin/packages/perl/package.py
johnwparent: A common use case that a number of nmake packages require is to prefix the first three numbers in the version string with `MSVC` such as `MSVC142`. It's a common enough case I think the `Msvc` compiler class should provide the method. That being said I did not know the version object could do that, so I'll refactor the MSVC side to use that functionality. Super useful the version object can do that.

lib/spack/spack/compilers/msvc.py
scheibelp: > with no arguments produces a block out output that happens to include the version string, which is successfully picked up by that regex.

Are you saying it specifically extracts `19.29.30145`? If so, then I'm missing something and this is fine. What I assumed was that this was taking the entire output string and assigning it as the version.

johnwparent: Yes, that's exactly what I'm saying. It's just grabbing the version string from the rest of the output.

lib/spack/spack/build_environment.py
```diff
@@ -564,8 +564,13 @@ def _set_variables_for_single_module(pkg, module):
     # Don't use which for this; we want to find it in the current dir.
     m.configure = Executable("./configure")
 
+    m.meson = Executable("meson")
+    m.cmake = Executable("cmake")
```
scheibelp: Note that where this runs, the Spack-built cmake won't necessarily be part of the path. However, the CMake package itself has:

```
        """Called before cmake packages's install() methods."""

        module.cmake = Executable(self.spec.prefix.bin.cmake)
        module.ctest = Executable(self.spec.prefix.bin.ctest)
```

so this should already be handled (likewise for `meson` and `ctest`. `make` is a special case, so you might have gotten the impression that was how to add build tool references, but FYI that's best to avoid if possible (might make sense to still do this for `msbuild` though).

lib/spack/spack/detection/path.py
```diff
@@ -231,13 +235,14 @@ def by_executable(packages_to_check, path_hints=None):
         path_hints (list): list of paths to be searched. If None the list will be
             constructed based on the PATH environment variable.
     """
-    path_hints = [] if path_hints is None else path_hints
+    path_hints = spack.util.environment.get_path("PATH") if path_hints is None else path_hints
```
scheibelp, line 239: This is an overall qualitative change in behavior:

* On linux, if `path_hints` was none, this would carry through to `executables_in_path` call below, which would then be initialized from PATH
* On Windows `path_hints.extend(compute_windows_user_path_for_package(pkg))` that runs in the for loop below could lead to cases where the existence of any custom search path in a package would mean that `PATH` would never be searched by default

So this seems reasonable. We could preserve the old behavior by allowing users to explicitly *avoid* searching PATH by specifying their own path by tracking the per-package custom Windows search paths in a separate variable, but I don't think that's worthwhile.

My only concern is that when multiple nested function calls take a responsibility of initializing from `None`, it can be confusing which one is really "responsible". It might be worth removing the default from `executables_in_path` (cursory grep suggests that this might require updating a couple tests, but otherwise I think this is the only code that calls it).

var/spack/repos/builtin/packages/msmpi/package.py
```diff
@@ -3,40 +3,54 @@
...
+class GenericBuilder(GenericBuilder):
+    def is_64bit(self):
+        return platform.machine().endswith("64")
+
+    def build_command_line(self):
+        arch = "intel64" if self.is_64() else "ia32"
+        args = ["-noLogo"]
+        # The argument may say gfortran, but spack patches MSMPI to be compatible with IFortran
+        # however that variable name proved difficult to patch, so it remians the same
+        args.append("/p:GFORTRAN_BIN=%sbin\\%s" % (os.environ["IFORT_COMPILER21"], arch))
```
scheibelp: Is `IFORT_COMPILER21` guaranteed to be set? I don't see it set by this PR (or grepping through the current codebase).

lib/spack/spack/util/windows_registry.py
scheibelp: Now that the getter sets `_handle`, is this needed still?

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,324 @@
...
+    def invalid_reg_ref_error_handler(self):
+        try:
+            try:
+                yield
+            except FileNotFoundError as e:
+                if e.winerror == 2:
+                    tty.debug("Key %s at position %s does not exist" % (self.key, str(self.root)))
+                else:
+                    raise e
+        finally:
```
scheibelp: What is the purpose of this outer `finally`, would that swallow the exception on the line above? If so, why raise the exception? If not, then I'm not sure what the use of the nesting is.

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,324 @@
...
+        except FileNotFoundError as e:
+            if e.winerror == 2:
+                self._reg = -1
+                tty.debug("Key %s at position %s does not exist" % (self.key, str(self.root)))
+            else:
+                raise e
+
+    def _valid_reg_check(self):
+        if self.reg == -1:
+            tty.debug("Cannot perform operation for nonexistent key %s" % self.key)
```
scheibelp, line 193: Is it anticipated that users might ask for keys that would not be available? Or would this primarily result from mistyping a key? In the latter case, IMO we should just raise an exception. (if not, how would a user generally proceed without the key's value?)

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,324 @@
...
+        raise RuntimeError("Cannot invoke Windows registry methods on non Windows platform")
+    if type(root) == WindowsRegistryView or type(root) == RegistryKey:
+        return root.get_subkey(subkey)
+    return winreg.OpenKeyEx(root, subkey, access=winreg.KEY_READ)
+
+
+def get_value(root, value):
+    """
+    Returns registry value of name `value` associated with the key `root`
+    `root` must be an already open RegistryKey object or PyHKEY object
```
scheibelp: Can you also describe the shared interface between them (e.g. would users always just call `get_value()` on the result, or does `RegistryKey in fact implement all the same methods)?

scheibelp (changes requested):
Some additional questions and requests.

lib/spack/spack/util/windows_registry.py
johnwparent: Unsure, I think this may have been a rebasing mistake, or maybe I got turned around during implementation, either way, the nesting is unnecessary. Removed.

lib/spack/spack/util/windows_registry.py
johnwparent: The only way to check if a key exists or not is to query it and allow the query to fail if the key is missing. From that point it becomes difficult to distinguish between misspelling and invalid queries, so I've left it up to the user how they want to handle a missing key but just returning Null. Raising an exception is effectively the same thing, except instead of checking for null a user would need to do some exception handling. If you prefer the exception handling approach, I can remove this and the `invalid_reg_ref_handler` context manager.

lib/spack/spack/util/windows_registry.py
johnwparent: he `pyHKEY` object has effectively no API and can only be used as a method argument to other winreg methods. Technically it has a very limited API, but one that provides almost no utility to Spack. Do you think it's worth it to bother returning the raw `pyHEKY` or should I always wrap it in a `RegistryKey`. I initially thought it should be the way it currently is but after this comment I'm leaning the other direction.

lib/spack/spack/build_environment.py
johnwparent: This is a merge mistake. IIRC Cmake and the like used to be defined in the `build_environment` module, but they were refactored that way a few months ago. The MSMPI PR was created in June, and when I merged it into this branch, the addition of the MSBuild build tool probably carried these changes with it.

var/spack/repos/builtin/packages/msmpi/package.py
johnwparent: We specify OneAPI as an optional prereq for running Spack on Windows, and the MSVC compiler setup environment method will ensure this is present (from vcvars env setup script) so this will be present if the compiler has an associated fortran component (msvc compiler class just uses ifort). I'll add a check to ensure the compiler has an associated fortran. Edit: That env variable should also change w.r.t. to the version of the fotran compiler, so I'll add that as well.

lib/spack/spack/detection/path.py
johnwparent: Sounds good, change is made.

lib/spack/spack/util/windows_registry.py
johnwparent: Nope, removed.

lib/spack/spack/util/windows_registry.py
scheibelp: (EDIT) Ideally the method would return something with a consistent interface, so IMO wrapping in a RegistryKey would be preferable.

lib/spack/spack/util/windows_registry.py
scheibelp: Based on further conversation examples of missing keys are things like:

* If you query for a key related to an existing msmpi installation and it has not in fact been installed, then it would return nothing
* Same for WDK/SDK queries added by this PR to `detection/common.py`

However, most of the methods that use this shouldn't in fact be run (i.e. `get_subkeys` shouldn't be run on a failed query key) so IMO this should generally raise an exception (the initial query for the root key should be allowed to "miss" though without an exception).

lib/spack/spack/compilers/msvc.py
```diff
@@ -90,10 +94,25 @@ def __init__(self, *args, **kwargs):
...
+        """
+        This is the shorthand VCToolset version of form
+        MSVC<short-ver> *NOT* the full version, for that see
+        Msvc.msvc_version
+        """
+        ver = self.msvc_version[:2].joined.string[:3]
         return "MSVC" + ver
 
+    @property
+    def cl_version(self):
```
scheibelp, line 112: (minor) This was originally called `cc_version` (i.e. as implemented in this PR), which has a special meaning (i.e. other core parts of the code will use `cc_version` in some circumstances). Now that this is renamed, I don't think anything will by default call `.cl_version`, and I don't see anything calling it in this PR. Is this anticipated to be useful later?

lib/spack/spack/detection/path.py
```diff
@@ -44,36 +57,11 @@ def executables_in_path(path_hints=None):
         path_hints (list): list of paths to be searched. If None the list will be
             constructed based on the PATH environment variable.
     """
-    # If we're on a Windows box, run vswhere,
-    # steal the installationPath using windows_os.py logic,
-    # construct paths to CMake and Ninja, add to PATH
     path_hints = path_hints or spack.util.environment.get_path("PATH")
```
scheibelp: To expand on https://github.com/spack/spack/pull/33021#discussion_r1026074542, my original concern was that two different functions participating in the same call graph were taking the responsibility to initialize with a default of `PATH`. In general nested equivalent definitions are problematic. I'd settle for a TODO comment about this.

It occurs to me this might be confusing since callers would always need to provide a nonempty set of path_hints (if we don't default to PATH, not doing this would be an error). If we keep the hold behavior (default to PATH) then I think the default parameter value of None is good. If we stop defaulting to PATH, then we should `raise` when `bool(path_hints) == False`.

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,313 @@
...
+
+def open_key(root, subkey):
+    """Returns registry key handle derived from root and its subkey
+    root can be either a WindowsRegistry or RegistryKey object
+
+    Note: If called on non Windows platforms, will raise RuntimeError
+    """
+    if not is_windows:
+        raise RuntimeError("Cannot invoke Windows registry methods on non Windows platform")
+    return root.get_subkey(subkey)
```
scheibelp: IMO this and `get_value` below no longer add much: is the main thing that is raises an error? WindowsRegistryView already does that, although the RegistryKey initializer doesn't, so I could see that being useful here. If we move the `raise` into the initializer, I assume that would cover what these two functions intend.

lib/spack/spack/util/windows_registry.py
```diff
@@ -0,0 +1,313 @@
...
+                                search all subkeys recursively.
+                                Default is True
+        Return:
+            the desired subkey as a RegistryKey object, or none
+        """
+        if not recursive:
+            return self.get_subkey(subkey_name)
+
+        else:
+            return self._traverse_subkeys(lambda x: x.name == subkey_name)
```
scheibelp, line 266: I think this and `find_value` should avoid calling `_traverse_subkeys` if `not self._valid_reg_check()`, and `_traverse_subkeys` should fail if the root key does not exist (IMO allowing traversals starting from an invalid root is confusing). Would that cause problems?

var/spack/repos/builtin/packages/wgl/package.py
```diff
@@ -0,0 +1,86 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+import re
+
+from spack.package import *
+
+
+class Wgl(BundlePackage):
+    """External WGl and Windows OpenGL emulation representation in Spack"""
```
scheibelp, line 13: If this isn't meant to be installed, then I recommend using the same technique as for `win-sdk` (raise an error): if `wgl` is not present on the system, then - with this package in its current state - `spack install wgl` wouldn't actually install the `wgl` libraries.

Given the dependency version locking you do here, I assume the question would then be how to associate the right `wgl` with the correct `win-sdk` implementation. That is a more general issue for how we handle externals and for now we can assume that the `packages.yaml` will contain a single instance of `wgl` and `win-sdk` (or at least, if that isn't true then that's a problem in Spack core that we need to address and not specifically an issue with this PR).

scheibelp (changes requested):
Some additional requests: if any of them don't seem minor let me know.

lib/spack/spack/compilers/msvc.py
johnwparent: Yes, this is a "I will probably need this in the near future when porting a package" change, not an "I need it right now" change.

lib/spack/spack/util/windows_registry.py
johnwparent: I had intended to just consolidate the error handling for both into `_traverse_subkeys`, I just missed updating traverse subkeys to throw like the other methods do now.

lib/spack/spack/util/windows_registry.py
johnwparent: Agreed. They initially served a purpose but we refactored it away, removing.

var/spack/repos/builtin/packages/wgl/package.py
johnwparent: I was following the existing pattern for opengl style packages (WGL is essentially EGL), which lead me to believe bundle packages also could not be installed. I agree, and will refactor to have WGL throw during the install method. I don't believe this requires changes beyond that, but please correct me if that's not the case.

lib/spack/spack/detection/path.py
johnwparent: I intended the remove the `path_hints` defaulting to path from `exectuables_in_path` but must have dropped the change switching branches, would that resolve the issue here? Or are you referring to the default in this method?

scheibelp: > I intended the remove the path_hints defaulting to path from exectuables_in_path but must have dropped the change switching branches, would that resolve the issue here?

Yes (once you do that though the test in `test/cmd/external` will have to pass in the path when it calls `executables_in_path` or else it will fail.

scheibelp (dismissed):

scheibelp:
@spackbot run pipeline

scheibelp:
Seems like the build tests are still hitting some intermittent download issues. I'll retry the pipeline.

scheibelp:
@spackbot run pipeline

scheibelp (approved):

lib/spack/llnl/util/symlink.py
```diff
@@ -24,7 +24,7 @@ def symlink(real_path, link_path):
     On Windows, use junctions if os.symlink fails.
     """
     if not is_windows or _win32_can_symlink():
-        os.symlink(real_path, link_path)
+        os.symlink(real_path, link_path, target_is_directory=os.path.isdir(real_path))
```
haampie, line 27: Why was this added? ~It's a really bad idea, cause this way dangling symlinks cannot be created.~ Ok, I mistakenly thought isdir throws if the file doesn't exist. In any case, I submitted #34321.

In general I'd like to ask you to never introduce changes for Windows support that cause performance regressions for non-Windows. Just assume that every file system operation is the bottleneck in some part of Spack, whether it's views, post-install hooks, buildcaches, ...

scheibelp: > In general I'd like to ask you to never introduce changes for Windows support that cause performance regressions for non-Windows.

This request is impossible to satisfy, or even to break down into something realistic or manageable. It is equivalent to saying "don't introduce bugs" (I'm already trying to do that).

As for the specific issue, I personally think a mistake at this level is unavoidable. Even if your original complaint was true. We have this code, and then we have you noticing and making an adjustment that avoids the issue (presumably because you have a benchmark which detects that something started slowing down). If you're worried about dangling symlinks and the tests don't have it, then they should have it; if you think there should be a benchmark for tracking this type of slowdown and we don't have it, then we should have one.

haampie: I guess what I'm saying is: be mindful about filesystem operations.

lib/spack/spack/detection/common.py
```diff
@@ -195,10 +218,117 @@ def update_configuration(detected_packages, scope=None, buildable=True):
...
+    @staticmethod
+    def find_windows_kit_roots():
+        """Return Windows kit root, typically %programfiles%\\Windows Kits\\10|11\\"""
+        if not is_windows:
+            return []
+        program_files = os.environ["PROGRAMFILES(x86)"]
+        kit_base = os.path.join(
+            program_files, "Windows Kits", WindowsKitExternalPaths.plat_major_ver
+        )
+        return kit_base
```
alalazo, line 275: Is there any reason this function returns either an empty list or a string?

johnwparent: Nope, and that's definitely a mistake we missed. IMO we should actually always be returning a list from a glob over the base kits directory so we can support SDK 10 and SDK 11. 
I'll refactor, PR, and tag you.

alalazo: No worries, I stumbled on this while doing #39803

PR #33024, opened 2022-10-05
pfunit: Fix ~openmp for version 4

cferenba:
This fixes a bug in building pFUnit 4 with OpenMP disabled.  The syntax for turning off OpenMP changed between versions 3 and 4 but the `package.py` didn't reflect this.  With this change, the OpenMP is handled in the same way as MPI, which does the right thing for both 3 and 4.

@citibeth


cferenba:
FYI @clellsolomon @junghans

junghans (approved):

PR #33025, opened 2022-10-05
netcdf packages: filter compiler wrappers in the *-config files

skosukhin:
We should have done this long ago. However, it looks like nobody really cares about compiler-related entries in those files.

Whoever reviews this, please, take a look at the long-pending #31953. Thank you.

### DIFF
--- a/var/spack/repos/builtin/packages/netcdf-c/package.py
+++ b/var/spack/repos/builtin/packages/netcdf-c/package.py
@@ -247,6 +247,22 @@ class NetcdfC(AutotoolsPackage):
 
         return config_args
 
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        # Some packages, e.g. ncview, refuse to build if the compiler path returned by nc-config
+        # differs from the path to the compiler that the package should be built with. Therefore,
+        # we have to shadow nc-config from self.prefix.bin, which references the real compiler,
+        # with a backed up version, which references Spack compiler wrapper.
+        if os.path.exists(self._nc_config_backup_dir):
+            env.prepend_path("PATH", self._nc_config_backup_dir)
+
+    @run_after("install")
+    def backup_nc_config(self):
+        # We expect this to be run before filter_compiler_wrappers:
+        nc_config_file = self.prefix.bin.join("nc-config")
+        if os.path.exists(nc_config_file):
+            mkdirp(self._nc_config_backup_dir)
+            install(nc_config_file, self._nc_config_backup_dir)
+
     def check(self):
         # h5_test fails when run in parallel
         make("check", parallel=False)
@@ -255,3 +271,7 @@ class NetcdfC(AutotoolsPackage):
     def libs(self):
         shared = "+shared" in self.spec
         return find_libraries("libnetcdf", root=self.prefix, shared=shared, recursive=True)
+
+    @property
+    def _nc_config_backup_dir(self):
+        return join_path(self.metadata_dir, "spack-nc-config")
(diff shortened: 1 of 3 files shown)


skosukhin:
Oh, `ncview` now gives a wonderful configure-time message:
```
======================================================================
Configuration error: You specified that the "/scratch/local1/spack/lib/spack/env/gcc/gcc" C compiler should be
used to build ncview, but the netcdf library was compiled with the "/usr/bin/gcc"
compiler.  Here is the path where I found the netcdf library:
     -L/scratch/local1/mysw/packages/gcc-10.2.1-sys/netcdf-c-4.8.1/lib -lnetcdf
You must use the same compiler for ncview as was used to build the netcdf library!

There are two possible ways to fix this.

1) Download the netcdf library and compile it with the /scratch/local1/spack/lib/spack/env/gcc/gcc compiler,
install it, and try again to build ncview.  NOTE that if you do this,
you might have more than one version of the netcdf library on your system,
built with differnt compilers, which can get confusing.  If you do this,
you will have to specify the path to the (new) version of nc-config
that was compiled using the /scratch/local1/spack/lib/spack/env/gcc/gcc compiler by configuring ncview in a
way similar to this:

   ./configure --with-nc-config=/path/to/newly/compiled/nc-config

2) Configure ncview to use the /usr/bin/gcc compiler.
To do this, set environmental variable CC to /usr/bin/gcc
and run ./configure again
===================================================================
```
I will try to fix this one tomorrow but there might be more packages that do a similar thing.

alalazo (commented):
@skosukhin Feel free to ping me when done

skosukhin:
The most simple solution to the problem that I could come up with is to back up the unmodified `nc-config` and make it available in the dependent build environment via the `PATH` variable. This works for `ncview` but might not work for a package that makes assumptions on the relative path of `nc-config`:
1. A package might expect the installation prefix of `netcdf-c` to be the result of something like:
    ```bash
    netcdf_prefix="$(dirname $(which nc-config))/.."
    ```
    It doesn't look like a reasonable thing to do but you never know.
2. A package might be given an installation prefix to `netcdf-c` and expect `nc-config` in the `bin` subdirectory of the prefix. In this case, the package will work with the modified `nc-config`. As far as I remember, `cdo` will do that. Luckily, it does not care about what `nc-config --cc` returns.

An alternative solution that covers such cases could be to modify `nc-config` in a way that it returns Spack wrapper in the Spack build environment (e.g. `netcdf-c` sets something like `SPACK_NETCDF` for the dependent build environment, which would be the trigger for `nc-config`) and the real compiler otherwise. But that would require us to either duplicate the already quite complex logic of `filter_compiler_wrappers` or extend `filter_compiler_wrappers` itself. I don't really want to do that right now.

I also would like not to do the same thing for `netcdf-fortran` and `netcdf-cxx4` unless somebody gives me an example of when that is required (I'm not aware of packages that call `nf-config` or `ncxx4-config`). In that case, however, I would appreciate a piece of advice on how to avoid code duplication.

@alalazo

skosukhin:
I have searched for `nc-config`, `nf-config` and `ncxx4-config` in the package repository and decided to ping maintainers of packages that use the config scripts explicitly so they could tell whether they forsee any problems caused by this PR:

- @adamjstewart: `gmt`, `grass`
- @climbfuji, @jedwards4b: `esmf`
- @t-yamaura: `scale`
- @amd-toolchain-support: `roms`

Maybe we need a different solution after all.

jedwards4b (approved):

t-yamaura (approved):

climbfuji (approved):
I don't think this will cause a problem, but testing it to the full extent will take some time. Please go ahead with the PR, if we run into problems later on we'll let you know.

haampie:
@spackbot run pipeline

alalazo (approved):

alalazo:
@adamjstewart @amd-toolchain-support Waiting some time for you to report if this works with the packages mentioned above.

adamjstewart (approved):
Confirmed that `gmt` and `grass+netcdf` both build on this branch.

PR #33028, opened 2022-10-05
scotch: @7: uses cmake which doesn't add '-std=c99' for intel classic compiler

wrtobin:
In scotch, `common_integer.c` requires the `PRIu64` symbol to be defined, which is defined via including `<inttypes.h>` in `common.h` via checking the existence and value of the `__STDC_VERSION__` preprocessor macro. 

When using the intel classic compiler this macro is only defined when using the `-std=c99` flag: [intel docs](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/macros/iso-standard-predefined-macros.html).

This isn't handled by their internal config/build system, but adding the needed flag to the cmake command line invocation successfully allows it to build for me.


pghysels (approved):

haampie (changes requested):
Scotch should specify a C-standard, this has nothing to do with the Intel compiler.

Can you file an issue upstream that they should `set(CMAKE_C_STANDARD 99)`

And can you add this `self.define("CMAKE_C_STANDARD", "99")` unconditionally to the package.py?

bernhardkaindl:
@wrtobin: One of the spack maintainers asked you a question:
> Scotch should specify a C-standard, this has nothing to do with the Intel compiler.
> 
> Can you file an issue upstream that they should `set(CMAKE_C_STANDARD 99)`
> 
> And can you add this `self.define("CMAKE_C_STANDARD", "99")` unconditionally to the package.py?

PR #33033, opened 2022-10-05
trilinos and xyce: fix fortran library handling

kuberry:
Also in xyce:
  - add pymi_static_blas variant and logic
    handles blas and hdf5 conflicts for Xyce-PyMi
  - make +isorropia and +zoltan conditional on mpi

Accomplishes what PR #31546, but using the `flag_handler` rather than manipulating `CMAKE_SHARED_LINKER_FLAGS` or `Trilinos_EXTRA_LINK_FLAGS`.

### DIFF
--- a/var/spack/repos/builtin/packages/xyce/package.py
+++ b/var/spack/repos/builtin/packages/xyce/package.py
@@ -131,20 +131,16 @@ class Xyce(CMakePackage):
 
         options = []
 
-        trilinos = spec["trilinos"]
-        options.append("-DTrilinos_DIR:PATH={0}".format(trilinos.prefix))
-
-        build_type = spec.variants["build_type"].value
-        options.extend(["-DCMAKE_BUILD_TYPE:STRING={0}".format(build_type)])
-
         if "+mpi" in spec:
-            options.append("-DCMAKE_CXX_COMPILER:STRING={0}".format(spec["mpi"].mpicxx))
+            options.append(self.define("CMAKE_CXX_COMPILER", spec["mpi"].mpicxx))
         else:
-            options.append("-DCMAKE_CXX_COMPILER:STRING={0}".format(self.compiler.cxx))
+            options.append(self.define("CMAKE_CXX_COMPILER", self.compiler.cxx))
 
-        options.append(self.define_from_variant("Xyce_PLUGIN_SUPPORT", "plugin"))
         options.append(self.define_from_variant("BUILD_SHARED_LIBS", "shared"))
         options.append(self.define_from_variant("CMAKE_CXX_STANDARD", "cxxstd"))
+        options.append(self.define_from_variant("CMAKE_BUILD_TYPE", "build_type"))
+        options.append(self.define_from_variant("Xyce_PLUGIN_SUPPORT", "plugin"))
+        options.append(self.define("Trilinos_DIR", spec["trilinos"].prefix))
 
         if "+pymi" in spec:
             pybind11 = spec["py-pybind11"]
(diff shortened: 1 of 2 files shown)


kuberry:
@sethrj This moves the fortran flag manipulation into the `flag_handler` (for Trilinos and Xyce). It calls the `self.compiler.fc` rather than the `spack_fc`, which requires all of the Spack environment variables.

tldahlgren (dismissed):
LGTM

tldahlgren:
Will defer merge to give @keitat and or @sethrj a chance to comment.

tldahlgren:
> Will defer merge to give @keitat and or @sethrj a chance to comment.

*Unless* you want me to go ahead and merge the PR.

var/spack/repos/builtin/packages/xyce/package.py
```diff
@@ -59,17 +61,30 @@ class Xyce(CMakePackage):
...
     variant("pymi", default=False, description="Enable Python Model Interpreter for Xyce")
+    # Downstream dynamic library symbols from pip installed numpy and other
+    # pip installed python packages can cause conflicts. This is most often
+    # seen with blas symbols from numpy, and building blas static resolves
+    # this issue.
+    variant(
+        "pymi_static_blas",
+        default=True,
+        sticky=True,
+        description="Require static blas build for PyMi",
```
sethrj, line 73: Should have a conditional on +pymi

kuberry:
@tldahlgren Thank you. I'd prefer to wait for their input, especially since @sethrj and I had discussed this previously.

Do you know if there is a way to initiate the checks again since one failed from an SSL error (without closing/reopening preferably)?

alalazo:
@spackbot run pipeline

alalazo:
@spackbot help

alalazo:
@kuberry fyi :point_up:

var/spack/repos/builtin/packages/xyce/package.py
```diff
@@ -86,25 +101,41 @@ class Xyce(CMakePackage):
...
+        options = []
 
-        cxx_flags = [self.compiler.cxx_pic_flag]
-        try:
-            cxx_flags.append(self.compiler.cxx11_flag)
-        except ValueError:
-            pass
-        cxx_flags.append("-DXyce_INTRUSIVE_PCE -Wreorder -O3")
+        trilinos = spec["trilinos"]
+        options.append("-DTrilinos_DIR:PATH={0}".format(trilinos.prefix))
```
sethrj: Use self.define

var/spack/repos/builtin/packages/xyce/package.py
```diff
@@ -86,25 +101,41 @@ class Xyce(CMakePackage):
...
 
-        options = []
-        options.extend(
-            [
-                "-DTrilinos_DIR:PATH={0}".format(trilinos.prefix),
-                "-DCMAKE_CXX_FLAGS:STRING={0}".format(" ".join(cxx_flags)),
-            ]
-        )
+        build_type = spec.variants["build_type"].value
+        options.extend(["-DCMAKE_BUILD_TYPE:STRING={0}".format(build_type)])
```
sethrj: Ditto

var/spack/repos/builtin/packages/xyce/package.py
kuberry: I'd really like to keep this `pymi_static_blas` optional in the sense that I want users to be able to install with `+pymi` and not necessarily require that they use the static tpl constraints, which is why I default to `True` but am trying to avoid an explicit dependency between them.

sethrj: No, I mean this variant is meaningless when pymi isn't used right? So it should have when=

kuberry: Yes, you're right. As is, there is no logic where `+pymi_static_blas` does something without `+pymi`. I made the change. Thanks for the suggestion.

sethrj (approved):

PR #33052, opened 2022-10-06
Add a warning on Python 2.7 deprecation

alalazo:
We decided to remove support for running under Python 2.7 after the 0.19 release series. This PR adds a warning message to inform users.

### DIFF
--- a/lib/spack/spack/main.py
+++ b/lib/spack/spack/main.py
@@ -547,6 +547,11 @@ def setup_main_options(args):
     # Assign a custom function to show warnings
     warnings.showwarning = send_warning_to_tty
 
+    if sys.version_info[:2] == (2, 7):
+        msg = ("Running Spack using Python v2.7 is deprecated and support for it will be "
+               " removed after the Spack v0.19 release series.")
+        warnings.warn(msg)
+
     # Set up environment based on args.
     tty.set_verbose(args.verbose)
     tty.set_debug(args.debug)
--- a/lib/spack/spack/main.py
+++ b/lib/spack/spack/main.py
@@ -548,11 +548,10 @@ def setup_main_options(args):
     warnings.showwarning = send_warning_to_tty
 
     if sys.version_info[:2] == (2, 7):
-        msg = (
-            "Running Spack using Python v2.7 is deprecated and support for it will be "
-            " removed in Spack v0.20, use Python v3.6+ instead."
+        warnings.warn(
+            "Python 2.7 support is deprecated and will be removed in Spack v0.20.\n"
+            "    Please move to Python 3.6 or higher."
         )
-        warnings.warn(msg)
 
     # Set up environment based on args.
     tty.set_verbose(args.verbose)
--- a/lib/spack/spack/main.py
+++ b/lib/spack/spack/main.py
@@ -548,8 +548,10 @@ def setup_main_options(args):
     warnings.showwarning = send_warning_to_tty
 
     if sys.version_info[:2] == (2, 7):
-        msg = ("Running Spack using Python v2.7 is deprecated and support for it will be "
-               " removed after the Spack v0.19 release series.")
+        msg = (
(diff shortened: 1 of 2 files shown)


alalazo:
@spackbot fix style

lib/spack/spack/main.py
```diff
@@ -547,6 +547,13 @@ def setup_main_options(args):
     # Assign a custom function to show warnings
     warnings.showwarning = send_warning_to_tty
 
+    if sys.version_info[:2] == (2, 7):
+        msg = (
+            "Running Spack using Python v2.7 is deprecated and support for it will be "
+            " removed after the Spack v0.19 release series."
```
adamjstewart: I would say removed in 0.20 to be explicit.

alalazo:
```suggestion
            " removed in Spack v0.20."
```

lib/spack/spack/main.py
```diff
@@ -547,6 +547,13 @@ def setup_main_options(args):
     # Assign a custom function to show warnings
     warnings.showwarning = send_warning_to_tty
 
+    if sys.version_info[:2] == (2, 7):
+        msg = (
+            "Running Spack using Python v2.7 is deprecated and support for it will be "
+            " removed in Spack v0.20."
```
adamjstewart: Should we say ", use Python 3.6+ instead"?

alalazo:
```suggestion
            " removed in Spack v0.20, use Python v3.6+ instead."
```

adamjstewart:
Message looks good now. We also have version detection logic in `bin/spack`, should we move this to there? Also, that logic seems out-of-date, it says we still support Python 3.5.

alalazo:
@adamjstewart I would just add the message here, I'm about to submit another draft PR to remove Python 2.7 specific code.

adamjstewart:
I would hold off on that, core will change so much by the time we release 0.19. I'm also planning on using pyupgrade to update everything to modern Python syntax.

alalazo:
> I would hold off on that, core will change so much by the time we release 0.19. I'm also planning on using pyupgrade to update everything to modern Python syntax.

Too late (I read your comment after submitting #33063) The  PR is a draft, to check early what fails and get comments on parts that I missed on a first try. Don't know if that is clear, but I'm looking forward to removing Python 2.7 😅

lib/spack/spack/main.py
```diff
@@ -547,6 +547,13 @@ def setup_main_options(args):
     # Assign a custom function to show warnings
     warnings.showwarning = send_warning_to_tty
 
+    if sys.version_info[:2] == (2, 7):
+        msg = (
+            "Running Spack using Python v2.7 is deprecated and support for it will be "
+            " removed in Spack v0.20, use Python v3.6+ instead."
+        )
+        warnings.warn(msg)
```
tgamblin: This is really long -- I think it's better as:

```suggestion
        warnings.warn("Python 2.7 support is deprecated and will be removed in Spack v0.20.")
        warnings.warn("Please move to Python 3.6 or higher.")
```

or:

```suggestion
        warnings.warn(
            "Python 2.7 support is deprecated and will be removed in Spack v0.20.\n"
            "    Please move to Python 3.6 or higher."
        )
```

tgamblin (changes requested):

tgamblin (approved):

luweizheng:
Hi, I pulled the latest version and got the warning. I just want to know how to upgrade to python 3 using spack. Actually, there are multiple pythons on my system. Should I config in spack and let spack know the path of python 3?

adamjstewart:
There are a couple ways to tell Spack which python to use. On startup, Spack will search your PATH. It first searches for python3, then python, then python2. So as long as Python 3 is somewhere in your path, that's the one that will be used. Another way to force Spack to use a particular Python exe is to set the SPACK_PYTHON env var to the full path of the Python interpreter you want to use.

Issue #33036, opened 2022-10-05
Clingo can't use packages from non-primary namespaces


flxmr:
### Steps to reproduce

So, maybe I misunderstand some things, but the thing from here: https://github.com/spack/spack/issues/23298 is still bugging me.

Now it's even worse and when I have setup another repo in second place in `repos.yaml`:
```
$spack repo list
==> 2 package repositories.
builtin    /spack/var/spack/repos/builtin
sne        /extra-repo
```
I can't use select packages from this repo like
```
spack spec sne.package
```
this always concretizes to the builtin repo
```
Input spec                                                                                                                          
--------------------------------                                                                                                    
sne.py-torch-cluster                                                                                                                
                                                                                                                                    
Concretized                                                                                                                         
--------------------------------                                                                                                    
builtin.py-torch-cluster@1.5.8%gcc@12.2.1~cuda arch=linux-fedora36-x86_64_v3                                                        
    ^builtin.py-pip@22.1.2%gcc@12.2.1 arch=linux-fedora36-x86_64_v3             
......                                                    
```

I thought this should pick the one from the sne-namespace!


### Error message

_No response_

### Information on your system

* **Spack:** 0.19.0.dev0 (db2565cb5339e3c590cdb951588755a5e7a45f25)
* **Python:** 3.10.6
* **Platform:** linux-fedora36-zen3
* **Concretizer:** clingo


### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

flxmr:
So, this seems relatively hard. When I look into asp.py, it gets a list of possible specs using package_basepy/`PackageBase()`'s `possible_dependencies`-function. This only works with strings at some point. Which means, it will always default to the first package found! not good. I'll try creating a `override_full_name`-dict and see whether I can pass this through all the relevant functions collecting facts for the concretizer and then use it to rewrite namespaces in the specs in the end. Sounds similar to what I did with the old concretizer.

flxmr:
So, as this seems to be added to a milestone: How can I help? Namely: what should I test/support? The patch over here https://github.com/spack/spack/pull/33053/checks works for us now (also ported it forward once)

alalazo:
[...]

Issue #28747, opened 2022-02-03
Overriding packages in the builtin repository changed semantic from v0.16 to v0.17


npf:
### Steps to reproduce

I'm trying to use the overriding of packages as described in https://spack.readthedocs.io/en/latest/repositories.html#overriding-built-in-packages, with the following setup:
- the builtin repo contains a package with the most recent versions (e.g. zlib 1.2.11)
- a local repo contains the same package but only an old version (e.g. zlib 1.2.8)

localrepo is added to the repo list, thus it has priority over builtin. Since I want the latest version of the package, I run `spack install builtin.zlib@1.2.11`, but it does not behave as expected with v0.17.0 and v0.17.1, while it's ok with v0.16.3

### Error message

```console
$ spack install builtin.zlib@1.2.11
==> Warning: Missing a source id for zlib@1.2.11
==> Installing zlib-1.2.11-ysprwfsnfe2ybpqg4r3tkbkvyavgmjne
==> No binary for zlib-1.2.11-ysprwfsnfe2ybpqg4r3tkbkvyavgmjne found: installing from source
==> Warning: There is no checksum on file to fetch zlib@1.2.11 safely.
==>   Fetch anyway? [y/N] y
==> Fetching https://zlib.net/fossils/zlib-1.2.11.tar.gz
==> No patches needed for zlib
==> zlib: Executing phase: 'install'
==> zlib: Successfully installed zlib-1.2.11-ysprwfsnfe2ybpqg4r3tkbkvyavgmjne
  Fetch: 2.40s.  Build: 3.33s.  Total: 5.73s.
[+] /path/to/spack/opt/spack/linux-debian10-skylake/gcc-8.3.0/zlib-1.2.11-ysprwfsnfe2ybpqg4r3tkbkvyavgmjne
$ spack find -N
==> 2 installed packages
-- linux-debian10-skylake / gcc@8.3.0 ---------------------------
spack-localrepo.zlib@1.2.8  spack-localrepo.zlib@1.2.11
```
→ zlib@1.2.11 is not fetched from builtin

While it was ok with v0.16.3.
```console
$ spack install builtin.zlib@1.2.11
==> Installing zlib-1.2.11-35k2t3imxwoldp42uymipncstsisqeqt
==> No binary for zlib-1.2.11-35k2t3imxwoldp42uymipncstsisqeqt found: installing from source
[...]

Issue #11747, opened 2019-06-17
depends_on not respecting namespace


JaroslavHron:
specifying namespace in `depends_on` does not work correctly
(as expected from https://spack.readthedocs.io/en/v0.12.1/repositories.html?highlight=depends_on)

dependency in `package.py` definition:
```console
depends_on('localrepoA.petsc')
```

spack still selects `localrepoB.petsc` package from localrepoB if the repos.yaml has localrepoB first 
```
cat repos.yaml

repos:
- /usr/local/pkg/spack/localrepoB
- /usr/local/pkg/spack/localrepoA
```

tgamblin:
This is a bug in the `depends_on` implementation.  We'll work on a fix.  The issue is that in `directives.py`, we're not preserving `pkg.fullname` when we make the dependency specs.

haampie:
@alalazo can you check?

alalazo:
I'll be removing this from the point releases. The repository namespace is currently not part of the solve, so fixing this issue might require some discussion on the semantic we want to implement. For instance, it's not clear to me what should be the behavior when:
```python
depends_on('repo1.hdf5')
```
and
```python
depends_on('repo2.hdf5')
```
are explicitly requested by different packages in the same solve.

PR #33053, opened 2022-10-06
Make secondary namespaces accessible with clingo

flxmr:
My try at solving this: https://github.com/spack/spack/issues/33036

- spec'cing on the CLI: `spack  spec -N sne.py-torch-geometric+cuda ^py-torch-cluster ^py-torch cuda_arch=60 ^cuda+allow-unsupported-compilers` → this just uses the pytorch-geometric-package from the sne-secondary namespace
- unified concretization of envs with specs from secondary repos:
```
spack:
  # add package specs to the `specs` list
  specs:
  - py-torch cuda_arch=60
  - cuda+allow-unsupported-compilers
  - py-torch-geometric+cuda
  view: true
  concretizer:
    unify: true
```

Unclear to me are:
- should a patch for this also suppport `depends_on` with namespaces?
- it seems a lot of the problems around this arise around always using `spec.name` instead of `spec.fullname`. Might this be a more robust fix?
- in general because of the former this might lead to veeeery unintended side-effects.

So now that I've done this as PoC for the new concretizer, I would like to see whether this is integrateable. Imho this is a nice feature if you are a small installation and want to easily support fixes/extras/... for some packages in certain cases.

### DIFF
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -679,6 +679,7 @@ class PackageBase(WindowsRPath, PackageViewMixin, metaclass=PackageMeta):
         visited=None,
         missing=None,
         virtuals=None,
+        namespace_overrides=None,
     ):
         """Return dict of possible dependencies of this package.
 
@@ -693,6 +694,9 @@ class PackageBase(WindowsRPath, PackageViewMixin, metaclass=PackageMeta):
             missing (dict or None): dict to populate with packages and their
                 *missing* dependencies.
             virtuals (set): if provided, populate with virtuals seen so far.
+            namespace_overrides (dict or None): full names of explicitly
+                mentioned specs. Here a namespace-override should be possible,
+                i.e. the "full"-name should be used as mapped herein!
 
         Returns:
             (dict): dictionary mapping dependency names to *their*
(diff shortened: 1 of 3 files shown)


flxmr:
related:

- https://github.com/spack/spack/issues/28747 
- https://github.com/spack/spack/issues/11747

flxmr:
@alalazo I would really like to fix this! Can I somehow contribute on this issue (we've ran above patch now for 2 or 3 spack updates and backporting gets annoying!)?

alalazo:
To fix this properly we need to add the concept of "namespace" to the clingo logic program. I'll try to open an issue and reference all the repos related bugs we have, and put said issue in the v0.22 milestone (targeted next May). Would that be fine?

flxmr:
Yes, I figured as much when finding my "fix" (and kept as far away as possible from clingo) - actually taking this back into my braincontext my idea fails with unavailable variants (as I'm just a posteriori changing package sources). 

So if you have three namespaces with different variants:
```
builtin.sample: +rocm
A.sample: +cuda +hip
B.sample: +cuda +oneapi
C.sample: +cuda
```
It should pick `B.sample` if you ask for `spack spec sample +oneapi`.

So essentially we have to:
a) change the input to include the namespace: https://github.com/spack/spack/blob/a3d6714c8bbcc17bf1bce2d54fdc5de8535feea4/lib/spack/spack/solver/asp.py#L920
b) accordingly change what we load into the concretizer?
c) change the concretizer.lp
d) put the answers into a set of packages we then can install?

So I guess I can figure out a,b,d - as for c I don't have a clue yet (and shied away last time). Do you have a recommended primer to clingo somewhere?

PR #29501, opened 2022-03-14
tbl2asn: replace

NessieCanCode:
typo on url_for_version return

PR #33061, opened 2022-10-06
ncbi-toolkit: adding version 26_0_1

snehring:
Fetch: 0.08s.  Build: 57m 3.37s.  Total: 57m 3.45s. 

:|

Anyway, I missed the fact that new releases moved to github when I touched this last.

Per discussion in #29501 I've added berkeley-db as a dependency and verified that table2asn is built.

This recipe needs some attention, several scripts involved in the build and configure process hard code a /bin or /usr/bin path for executables, which is annoying with spack. Modern versions use cmake and make it easier to disable tests, which would likely improve build time. I don't have the time right now to give it the attention it deserves.

### DIFF
--- a/var/spack/repos/builtin/packages/ncbi-toolkit/package.py
+++ b/var/spack/repos/builtin/packages/ncbi-toolkit/package.py
@@ -14,6 +14,13 @@ class NcbiToolkit(AutotoolsPackage):
     homepage = "https://www.ncbi.nlm.nih.gov/IEB/ToolBox/CPP_DOC/"
     url = "ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools++/CURRENT/ncbi_cxx--22_0_0.tar.gz"
 
+    # Per https://ncbi.github.io/cxx-toolkit/pages/ch_getcode_svn#ch_getcode_svn.external
+    # New versions are released on github
+    version(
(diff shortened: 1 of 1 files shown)


snehring:
One day spackbot. One day.

var/spack/repos/builtin/packages/ncbi-toolkit/package.py
```diff
@@ -34,6 +41,8 @@ class NcbiToolkit(AutotoolsPackage):
 
     depends_on("boost@1.35.0:+test+log")
     depends_on("bzip2")
+    depends_on("cpio", type="build")
+    depends_on("diffutils", type="build")
```
tldahlgren, line 45: Are these for a specific version on (so could use a `when=` arg) or do they apply to all versions?

tldahlgren (approved):
LGTM and confirmed sha256 so fine as-is except I do have a quick question about other new dependencies.

var/spack/repos/builtin/packages/ncbi-toolkit/package.py
snehring: I believe they apply to all versions. The container I was building it in for rhel9 didn't have cmp or cpio in it, so I was hitting these.

PR #33064, opened 2022-10-06
Installer: Replace missed instance of cache_only

kwryankrattiger:
@scottwittenburg I missed this instance when doing the use-buildcache option

@alalazo @tgamblin FYI

### DIFF
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -2355,9 +2355,14 @@ class BuildRequest(object):
         Returns:
             tuple: required dependency type(s) for the package
         """
+        cache_only = self.install_args.get("package_cache_only")
+        # This is likely never true, but check it to be safe
+        if not pkg.name == self.pkg.name:
+            cache_only = self.install_args.get("dependencies_cache_only")
+
         deptypes = ["link", "run"]
         include_build_deps = self.install_args.get("include_build_deps")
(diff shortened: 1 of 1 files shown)


scottwittenburg (approved):

haampie:
Why was this not merged yet? I hit the same issue and submitted a PR, only found this now.

Closing this cause it's not adding a regression test. https://github.com/spack/spack/pull/33363 see here for a test.

lib/spack/spack/installer.py
```diff
@@ -2355,9 +2355,14 @@ def get_deptypes(self, pkg):
         Returns:
             tuple: required dependency type(s) for the package
         """
+        cache_only = self.install_args.get("package_cache_only")
+        # This is likely never true, but check it to be safe
```
haampie, line 2359: Not now, but in the future this might change. Please use the unique ID which is `pkg_id`, which depends on dag_hash.

kwryankrattiger: ah, noted. I don't know if it will ever be possible for root packages to have duplicates in a build tree though. Can you think of a case, or is this more a "best practices" kind of thing?

haampie: In the future we may have cross compilation, so then deptype=build targets the host, deptype=link targets the ... target, which causes duplication.

And even sooner we may relax "one spec per package" to "one spec per package per link-type connected subgraph"

PR #33066, opened 2022-10-06
mvapich2-gdr/mvapich2x: Nickc/update package config

MatthewLieber:
Updating max versions of dependencies, changing compiler flags based on gcc version, and adding none launcher build variant.


natshineman (dismissed):

var/spack/repos/builtin/packages/mvapich2-gdr/package.py
```diff
@@ -23,6 +23,7 @@ class Mvapich2Gdr(AutotoolsPackage):
 
     maintainers = ["ndcontini", "natshineman", "harisubramoni"]
 
+    version('2.3.7', sha256='7bf748ed3750aa607382fc96229e256d888824aed758ce364b1ed9429da4779e')
```
tldahlgren: Need to replace single with double quotes here to pass style/black checks.

var/spack/repos/builtin/packages/mvapich2-gdr/package.py
```diff
@@ -113,6 +114,18 @@ def process_manager_options(self):
...
+            opts.append("--with-pm=mpiexec")
+            if "pmi_version=simple" in spec:
+                opts.append("--with-pmi=simple")
+            if "pmi_version=pmi1" in spec:
+                opts.append("--with-pmi=pmi1")
+            if "pmi_version=pmi2" in spec:
+                opts.append("--with-pmi=pmi2")
+            if "pmi_version=pmix" in spec:
+                opts.append("--with-pmi=pmix")
+                opts.append("--with-pmix={0}".format(spec["pmix"].prefix))
```
tldahlgren: Have you seen the `with_or_without` helper available for `AutotoolsPackage`s?  https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#with-or-without

var/spack/repos/builtin/packages/mvapich2-gdr/package.py
tldahlgren: FYI. Cannot access the site to confirm the sha256 (without registering).

var/spack/repos/builtin/packages/mvapich2-gdr/package.py
natshineman: I am not familiar with this. @MatthewLieber please take a look and see if this would improve our packages.

var/spack/repos/builtin/packages/mvapich2-gdr/package.py
natshineman: Unfortunately even after going through that page the sha won't really be verifiable. MVAPICH2-GDR, along with MVAPICH2-X are closed source packages that are only available through our buildcache so the tarball is not actually posted anywhere publicly. We mostly update these for our internal use, but also so that users can see when we've added newer versions of the package. 

We are in the process of improving our Spack deployments, would there be a better way to handle this? Or should simply remove the versions altogether?

tldahlgren: > Unfortunately even after going through that page the sha won't really be verifiable. MVAPICH2-GDR, along with MVAPICH2-X are closed source packages that are only available through our buildcache so the tarball is not actually posted anywhere publicly. We mostly update these for our internal use, but also so that users can see when we've added newer versions of the package.
> 
> We are in the process of improving our Spack deployments, would there be a better way to handle this? Or should simply remove the versions altogether?

@tgamblin @alalazo Maybe you have opinions on conventions for packages like this?

var/spack/repos/builtin/packages/mvapich2-gdr/package.py
alalazo: @MatthewLieber

MatthewLieber: I do not believe the with_or_without worked well with our configure system but I have gone in and tryed to cut down on all of the appending of pmi options

natshineman (commented):
Looks good to me. I don't think this has any major effects since both MV2-GDR and MV2-X are closed source so we are the only ones who use these package configs. 

In the future would it be best to make the public versions of these packages indicate that users should download entries from the buildcache rather than try to build it themselves? We could keep these full package.py files for our internal builds.

natshineman (approved):

alalazo (approved):

PR #33076, opened 2022-10-07
py-cupy: 11.2

ax3l:
Add the latest version of `cupy`:
  https://github.com/cupy/cupy/tree/v11.2.0#installation

cc @qwertos @RemiLehe

- [x] testing


ax3l:
Auto-disabled sub modules spotted in install - maybe we can provide extra deps for those.

- cutensor: @bvanessen
```
  -------- Configuring Module: cutensor --------
  /tmp/tmp3lzspdo_/a.cpp:1:10: fatal error: cutensor.h: No such file or directory
      1 | #include <cutensor.h>
        |          ^~~~~~~~~~~~
  compilation terminated.
```
- cusparselt:
```
  -------- Configuring Module: cusparselt --------
  /tmp/tmpbjlo73o_/a.cpp:1:10: fatal error: cusparseLt.h: No such file or directory
      1 | #include <cusparseLt.h>
        |          ^~~~~~~~~~~~~~
  compilation terminated.
```

tldahlgren (dismissed):
Confirmed sha256.

var/spack/repos/builtin/packages/py-cupy/package.py
```diff
@@ -16,6 +16,7 @@ class PyCupy(PythonPackage):
     homepage = "https://cupy.dev/"
     pypi = "cupy/cupy-8.0.0.tar.gz"
 
+    version("11.2.0", sha256="c33361f117a347a63f6996ea97446d17f1c038f1a1f533e502464235076923e2")
     version("8.0.0", sha256="d1dcba5070dfa754445d010cdc952ff6b646d5f9bdcd7a63e8246e2472c3ddb8")
 
     depends_on("python@3.6:", type=("build", "run"))
```
adamjstewart: Need to update deps:
```python
setup_requires = [                                                                       
    'Cython>=0.29.22,<3',                                                                
    'fastrlock>=0.5',                                                                    
]                                                                                        
install_requires = [                                                                     
    'numpy>=1.20,<1.26',  # see #4773                                                    
    'fastrlock>=0.5',                                                                    
]              
    python_requires='>=3.7',                                                             
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-cupy/package.py
```diff
@@ -16,12 +16,15 @@ class PyCupy(PythonPackage):
     homepage = "https://cupy.dev/"
     pypi = "cupy/cupy-8.0.0.tar.gz"
 
+    version("11.2.0", sha256="c33361f117a347a63f6996ea97446d17f1c038f1a1f533e502464235076923e2")
     version("8.0.0", sha256="d1dcba5070dfa754445d010cdc952ff6b646d5f9bdcd7a63e8246e2472c3ddb8")
 
-    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("python@3.7:", type=("build", "run"))
```
adamjstewart, line 26: You should make the new dependencies conditional on version, so:
```python
depends_on("python@3.7:", when="@11:", type=("build", "run"))
```
or something like that. Same for other changed deps.

ax3l: My feeling is to ditch the 8.0 tagged version to be honest, because it's ancient.

Looking at cupy release numbers, they probably also have a 1:1 relation to CTK releases.

adamjstewart: In that case you should deprecate it, not remove it.

var/spack/repos/builtin/packages/py-cupy/package.py
```diff
@@ -16,12 +16,15 @@ class PyCupy(PythonPackage):
     homepage = "https://cupy.dev/"
     pypi = "cupy/cupy-8.0.0.tar.gz"
 
+    version("11.2.0", sha256="c33361f117a347a63f6996ea97446d17f1c038f1a1f533e502464235076923e2")
     version("8.0.0", sha256="d1dcba5070dfa754445d010cdc952ff6b646d5f9bdcd7a63e8246e2472c3ddb8")
```
ax3l:
```suggestion
    version("8.0.0", sha256="d1dcba5070dfa754445d010cdc952ff6b646d5f9bdcd7a63e8246e2472c3ddb8", deprecated=True)
```

adamjstewart:
@spackbot fix style

adamjstewart (dismissed):

adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

PR #33077, opened 2022-10-07
py-black: add v22.10.0

adamjstewart:
Successfully builds on macOS 12.6 (arm64) with Python 3.9.13 and Apple Clang 14.0.0.

https://github.com/psf/black/releases/tag/22.10.0

### DIFF
--- a/var/spack/repos/builtin/packages/py-black/package.py
+++ b/var/spack/repos/builtin/packages/py-black/package.py
@@ -17,6 +17,7 @@ class PyBlack(PythonPackage):
 
     maintainers = ["adamjstewart"]
 
+    version("22.10.0", sha256="f513588da599943e0cde4e32cc9879e825d58720d6557062d1098c5ad80080e1")
     version("22.8.0", sha256="792f7eb540ba9a17e8656538701d3eb1afcb134e3b45b71f20b25c77a8db7e6e")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-black/package.py
```diff
@@ -25,29 +26,30 @@ class PyBlack(PythonPackage):
     # This is the last v21 release, and it's needed to format for Python 2.7
     version("21.12b0", sha256="77b80f693a569e2e527958459634f18df9b0ba2625ba4e0c2d5da5be42e6f2b3")
 
-    variant("d", default=False, description="enable blackd HTTP server")
     variant("colorama", default=False, description="enable colorama support")
     variant("uvloop", default=False, description="enable uvloop support")
+    variant("d", default=False, description="enable blackd HTTP server")
```
tldahlgren, line 31: Curious with the motivation for the variant ordering.  Why move `d` here?

tldahlgren (approved):
Confirmed sha256.

var/spack/repos/builtin/packages/py-black/package.py
adamjstewart: I try to keep the variant and dependency order the same as documented in `pyproject.toml`/`setup.py`/`setup.cfg`. The developers moved things around, so I did too.

PR #33079, opened 2022-10-07
Proper terminal width with prompt decorator

vanderwb:
The Spack environment prompt decorator for Bourne shells (set by `spack env activate -p`) does not properly encapsulate color codes with square brackets. Thus, when a line of text spills to the next line in the terminal, the display is broken as the shell is confused about the width of the $PS1 value. This StackExchange post describes the problem more generally:

https://unix.stackexchange.com/questions/105958/terminal-prompt-not-wrapping-correctly

The proposed modification in this PR adds a keyword argument to the colorize method to add these square brackets around color sequences if the argument is true. The argument is only set when colorize is called by the shell decoration code in shell.py, and so other uses of colorize should be unaffected by the change.

### DIFF
--- a/lib/spack/llnl/util/tty/color.py
+++ b/lib/spack/llnl/util/tty/color.py
@@ -158,7 +158,7 @@ class match_to_ansi(object):
         """Returns a TTY escape sequence for a color"""
         if self.color:
             if self.enclose:
-                return "\[\033[%sm\]" % s
+                return r"\[\033[%sm\]" % s
             else:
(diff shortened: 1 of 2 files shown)


lrtfm:
Nice modification. There are three commits unrelated to this pr. You should remove those commits, I think.

haampie (approved):
Looks good to me. I've taken the liberty to rebase on develop to get rid of the unrelated commits

PR #33104, opened 2022-10-07
Add checksum for py-onnxmltools 1.11.0

iarspider:


tldahlgren (commented):
Confirmed sha256 for py-onnxmltools.

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
```diff
@@ -28,6 +28,8 @@ class PyOnnxRuntime(CMakePackage, PythonPackage):
     depends_on("ninja", type="build")
     depends_on("python", type=("build", "run"))
     depends_on("protobuf")
+    # https://github.com/microsoft/onnxruntime/pull/11639
+    depends_on("protobuf@:3.19", when="@:1.19.0")
```
adamjstewart: Did you mean:
```suggestion
    depends_on("protobuf@:3.19", when="@:1.11")
```
? It seems like 1.12.0 is the first release containing this patch.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
iarspider: Whoops... Yes

adamjstewart (approved):

Issue #28170, opened 2021-12-29
Padded install root with trailing os.path.sep doesn't relocate properly


vanderwb:
### Steps to reproduce

I've found that if I build packages in an environment with a padded install root that ends with a separator and then populate a build cache and attempt to install the cached package into a second - unpadded - environment, the RPATH and text-path relocation operations fail because this trailing character is not stripped from the old padded prefix, and so when the old prefix is replaced by the new prefix via a regex, a separating character is lost and an invalid path is produced.

Providing exact instructions to reproduce is difficult, since it will depend on the number of characters in your install root. For me, the following combination in the padded environment spack.yaml produces the issue:

```
    install_tree:
      root: /glade/scratch/vanderwb/spack/test/build/opt
      padded_length: 126
```

Meanwhile, for this particular install root, if I use a padded_length of 128 characters it works. See below for the implications.

I found at least two ways to fix the problem, but they basically boil down to rstrip'ing the os.path.sep character from the padded root, as is done for the user-provided root. I could provide a PR for how I've fixed it in our Spack clone, but I'm not sure where it would be best to put the fix. To me, the most logical place is in the parse_install_tree method, but in my testing this does not resolve the problem for already-created cached binary packages. It may be more robust to rstrip the `old_layout_root` variable in the relocate_package method.

### Error message

There is no specific error message, but when using the debug flag and installing a package (`spack -d install hdf+shared`), a message is provided that demonstrates the trailing-slash issue.

For the 128-character (working) example:
```
==> [2021-12-28-22:55:31.203264] Relocating package from
  /glade/scratch/vanderwb/spack/test/build/opt/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__ to /glade/scratch/vanderwb/spack/test/pub/opt.
```

And from the 126-character (failing) example:
```
==> [2021-12-28-23:30:11.173922] Relocating package from
  /glade/scratch/vanderwb/spack/test/build/opt/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/ to /glade/scratch/vanderwb/spack/test/pub/opt.
```

Finally, if we compare the RPATH's of a relocated library in the second environment, the missing slash problem is obvious.

128-character:
```
$ readelf -d opt/hdf/4.2.15/gcc/4.8.5/kx64zky/lib/libdf.so | grep RPATH | tr ':' '\n' | grep openssl
[...]

PR #33122, opened 2022-10-07
Fix for padding relocation (issue 28170)

vanderwb:
This is a proposed fix for #28170. The problem is that the trailing slash can be left at the end of the padded path, and when Spack subs that path during package relocation the slash is removed and an invalid path is produced in the resulting files/binaries. If we strip trailing slashes, as is done for non-padded paths, the problem goes away for any length padded prefix.

### DIFF
--- a/lib/spack/spack/store.py
+++ b/lib/spack/spack/store.py
@@ -121,6 +121,7 @@ def parse_install_tree(config_dict):
             msg = "Cannot pad %s to %s characters." % (root, padded_length)
             msg += " It is already %s characters long" % len(root)
             tty.warn(msg)
+        root = root.rstrip(os.path.sep)
     else:
         root = unpadded_root
 


bernhardkaindl (approved):
@becker33 LGTM:
```py
            msg = "Cannot pad %s to %s characters." % (root, padded_length)
            msg += " It is already %s characters long" % len(root)
            tty.warn(msg)
+       root = root.rstrip(os.path.sep)
    else:
        root = unpadded_root
```

bernhardkaindl:
@becker33:
@vanderwb: Looking at it closer, would it be better to replace the trailing slash with another padding char?

(Thanks to @haampie)

vanderwb:
Hi @bernhardkaindl - in either case, the padding will be off by one from what the user requests. I do see the logic of adding the character, since it seems like it will prevent failure scenarios in which one has padded to exactly what they need and gets surprised when it is one less. That said, it seems to me that the lengths are already off... or at least they are for the two examples I've checked. My request for 128-char length gives paths with 129 characters (even without the separator) and 126 gives 127. But perhaps there are cases where it will end up == to the request?

In any event, what specifically do you mean by replacing it with a padding character? Do you suggest a simple replacement of the separator by `_` or are you thinking to add another whole layer like `/__/` (on Linux)?

bernhardkaindl:
@vanderwb Thanks for the explanation, I'm not sure if I understand if fully yet, but I can answer your question:

Yes, I ment the former: To replace the trailing slash (/) with an underscore(_).

But I have to acknowledge that I do not yet understand the purpose of the longer padded paths beyond I think for relocating binaries to shorter paths. Is this the use case? Thanks, Bernhard

vanderwb:
@bernhardkaindl - yes, that is our use case, and the only one I've seen documented. I'm not sure if there are any others. To describe simply, our workflow for supporting our software stack on our HPC system has been:

1. Build packages in test environment with padded prefix
2. Push vetted package builds to binary cache
3. Installed (relocate) the binaries from that cache to user-visible public environment

I did some testing today and I found the `re.sub(r'/$', '_', root)` suggestion problematic. It does work just fine when relocating a binary from its cache, since this logic does not get used at that time. However, for any existing spack install that was using padding with the problematic specific number of characters, the install_tree:root is changed to one that ends with an extra _, and so the install cannot find any of its existing packages, breaking the software tree.

Regardless, after thinking on it further, I feel that the rstrip is still the right approach. The stripped separator is re-added when needed by path.join operations (notably, when rpath'ing the original padded root), so it is unaffected by the rstrip. The rstrip simply helps when the old padded root is replaced by a new unpadded root, which seems to be stored as `buildpath` in the binary distribution metadata. So for all intents and purposes, the amount of padding remains unchanged by this PR.

Happy to look at this further if you have more questions or disagree on any of the above points.

bernhardkaindl:
@vanderwb: Thanks, that's great information and explanation!

One of the other maintainers was anxious that I might have made a possible mistake by trusting that your PR is safe.
He thought that it needs more investigation if this is correct, but I think this should but any doubts to rest. Thanks!

> The rstrip simply helps when the old padded root is replaced by a new unpadded root, which seems to be stored as buildpath in the binary distribution metadata.

Yes, the location where the rstrip occurs was no clear from looking at the PR, but from this description, this PR is safe, and you tested it and as you say helps then when it comes to replace the padded root with the unpadded root (I think when relocating from the mirror to the final install location).

Thanks for the additional testing on your systems and the additional thinking afterwards and the explanation of it!

haampie:
This was a mistake... Reverting in #35077

PR #33124, opened 2022-10-07
gitlab ci: Print better information about broken specs

scottwittenburg:
When a pipeline generation job is automatically failed because it generated jobs for specs known to be broken on develop, print better information about the broken specs that were encountered.  Include at a minimum the hash and the url of the job whose failure caused it to be put on the broken specs list in the first place.

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import base64
+import codecs
 import copy
 import json
 import os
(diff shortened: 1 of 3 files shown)


scottwittenburg:
Same issue with the macos python 3.8 tests on this PR as I have been encountering on #32596 where I could never figure out what the actual failure was.  On that PR, the test passed after the third or fourth retry, will try the same thing here.

kwryankrattiger (approved):

lib/spack/spack/ci.py
```diff
@@ -1343,13 +1343,9 @@ def generate_gitlab_ci_yaml(
...
-        error_msg = (
-            "Pipeline generation failed due to the presence of the "
-            "following specs that are known to be broken in develop:\n"
-        )
-        for broken_spec in known_broken_specs_encountered:
-            error_msg += "* {0}\n".format(broken_spec)
-        tty.die(error_msg)
+        tty.error("This pipeline generated hashes known to be broken on develop:")
+        display_broken_spec_messages(broken_specs_url, known_broken_specs_encountered)
+        tty.die()
```
haampie, line 1348: This requires an arg

scottwittenburg: Yeah, sorry.  My test didn't notice it because it has to pass `allow_failure=True` to the SpackCommand, since the command is expected to fail in this case and the expected text was in the output.  I added a fix [here](https://github.com/spack/spack/pull/32596/commits/bc56abc68e355e3d885574cb03bd862733bc8ebb) since I was ready to merge that, but have had to retry the py27 and py38 tests more times on that PR.  If that requires any more retries, I'll make a PR just for this.

PR #33132, opened 2022-10-09
phist: new version 1.11 and patch to make previous versions compile w…

jthies:
…ith OpenBLAS


haampie (changes requested):
Regarding the actual issue being fixed, isn't LAPACKE the actual stable C interface? Anything else is just an attempt to capture the Fortran ABI in a C API (which apparently wasn't correct until a recent version of LAPACK).

Can you drop the conflict on `OpenBLAS` and `netlib-lapack`? Otherwise it looks fine to me.

https://github.com/spack/spack/blob/2a43571a68f9a976fad2aa46fdee69f6d99039aa/var/spack/repos/builtin/packages/phist/package.py#L193-L199

After this conflict is dropped, we can also test the patch in gitlab CI

haampie (approved):
Thanks a lot! Pipelines passing again

PR #33133, opened 2022-10-09
root: new variant `webgui` when `+root7` (default True)

wdconinc:
ROOT has a webgui which is available with the `+root7` variant. This is a fairly large part of a ROOT install (275MB out of 732MB on my system) which is not necessarily useful in all use cases (e.g. inside containers on network-restricted HPC/HTC compute nodes). This new variant adds the option to retain the ROOT7 functionality but not necessarily include the `webgui` aspects.

This variant is set to true by default for backwards compatibility reasons: before this change, the `webgui` feature was enabled whenever `root7` was enabled.

This is the first conditional variant for this package (conditional on `+root7`), which I want to point out in case it was a conscious decision to avoid conditional variants in the ROOT package.

Maintainers: @drbenmorgan @vvolkl @HadrienG2 chissg


HadrienG2 (approved):

alalazo (approved):
Maintainer approved

PR #33136, opened 2022-10-10
global: add `--with-ncurses` flag to configure

lrtfm:
Add this option to make global turely depend on `ncurses`.


bernhardkaindl (approved):
LGTM and builds

PR #33140, opened 2022-10-10
Add option to disable reordering of -I/-L/-isystem paths in spack compiler wrapper

iarspider:
Short description: add class property to prevent Spack from reordering -I/-L paths in compiler wrapper, treating all `-I/-L/-isystem`. Default behavior is to 

Reasoning: CMS offline software [cmssw](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_6_X/master/cmssw-tool-conf.spec) is an example of a large package with lots of dependencies. It is built using `scram` tool (makefile generator similar to `cmake`) using custom ordering of `-I` and `-L` paths to speed up compilation process (e.g. pushing ROOT's paths to the front of the command line, so that `gcc` does not need to check 100+ other paths for ROOT headers). 

@smuzaffar can provide more information on how CMS builds it's offline software.

### DIFF
--- a/lib/spack/env/cc
+++ b/lib/spack/env/cc
@@ -445,17 +445,25 @@ while [ $# -ne 0 ]; do
             arg="${1#-isystem}"
             isystem_was_used=true
             if [ -z "$arg" ]; then shift; arg="$1"; fi
-            if system_dir "$arg"; then
-                append isystem_system_include_dirs_list "$arg"
+            if empty "SPACK_REORDER_PATHS"; then
+                if system_dir "$arg"; then
+                    append isystem_system_include_dirs_list "$arg"
+                else
+                    append isystem_include_dirs_list "$arg"
+                fi
             else
-                append isystem_include_dirs_list "$arg"
+                append include_dirs_list "$arg"
             fi
             ;;
         -I*)
             arg="${1#-I}"
             if [ -z "$arg" ]; then shift; arg="$1"; fi
-            if system_dir "$arg"; then
-                append system_include_dirs_list "$arg"
+            if empty "SPACK_REORDER_PATHS"; then
+                if system_dir "$arg"; then
+                    append system_include_dirs_list "$arg"
+                else
+                    append include_dirs_list "$arg"
+                fi
             else
                 append include_dirs_list "$arg"
             fi
@@ -463,8 +471,12 @@ while [ $# -ne 0 ]; do
         -L*)
             arg="${1#-L}"
             if [ -z "$arg" ]; then shift; arg="$1"; fi
-            if system_dir "$arg"; then
-                append system_lib_dirs_list "$arg"
+            if empty "SPACK_REORDER_PATHS"; then
+                if system_dir "$arg"; then
+                    append system_lib_dirs_list "$arg"
+                else
+                    append lib_dirs_list "$arg"
+                fi
             else
                 append lib_dirs_list "$arg"
             fi
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -494,6 +494,8 @@ def set_wrapper_variables(pkg, env):
     env.set(SPACK_INCLUDE_DIRS, ":".join(include_dirs))
     env.set(SPACK_RPATH_DIRS, ":".join(rpath_dirs))
 
+    env.set("SPACK_REORDER_PATHS", pkg.reorder_paths)
+
 
 def determine_number_of_jobs(
     parallel=False, command_line=None, config_default=None, max_cpus=None
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -494,7 +494,8 @@ def set_wrapper_variables(pkg, env):
     env.set(SPACK_INCLUDE_DIRS, ":".join(include_dirs))
     env.set(SPACK_RPATH_DIRS, ":".join(rpath_dirs))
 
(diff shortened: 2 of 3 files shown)


haampie:
Not a fan... do you actually observe slowdowns?

Spack effectively does two things:

1. A _stable_ partition of `-I` and `-L` flags into `[non-system | system]`
2. (Heuristically) inject extra `-I`/`-L` flags from transitive dependencies's `<prefix>/{lib,lib64,include}` dirs if they exist

The reordering in (1) shouldn't really affect the number of filesystem operations except for system libraries and headers, otherwise it's the same linear search.

It's (2) that could cause slowdowns if the build system generates optimal `-L`/`-I` flags -- but I'm sure gcc/ld are smart enough to drop duplicates...

Your patch seems to address only (1) not (2).

smuzaffar:
@haampie , cmssw offline software has over 500 packages as its dependencies. Each binary ( shared lib, executable) product generated within cmssw dependend on fraction of these. For example https://github.com/cms-sw/cmssw/blob/master/Utilities/OpenSSL/BuildFile.xml only needs `openssl` dependency.  SCRAM (the build system for cmssw software) properly takes care of transitive dependencies and will only add `-I` for the dependencies needed for each individual product. Now if spack adds `-I` for all the dependencies of cmssw and reorders them then there is a good chance the compiler has to go through all `-I` to find `openssl` headers (which might be at the end due to `-isystem`). This will surely slow down the build.

In order to avoid compiler warnings from some of the external packages (e.g. clhep/root) , we (in cmssw) use `-isystem root/include -isystem clhep/include` . Now if spack reorders the include paths and move `-isystem` at the end then there is also a good chance that we might picking up these headers from system installed root/clhep packages.

I personally think that spack should not add/reorder `-I and -L` and should trust the build system to do the right thing.

haampie:
To me it sounds like the monolith package can better be split into components so that `depends_on` is scoped correctly. There's a thing called `BundlePackage` in Spack so that you get one package to refer to all the monolith components, and views help you create a single prefix. (But I guess the issue is you would have to create 500 spack packages...)

> I personally think that spack should not add/reorder `-I` and `-L` and should trust the build system to do the right thing.

I wish we could do that, but we can't, at least not everywhere. There's tons of build system that will detect `libfoo.so` by just running `cc test.c -lfoo`, and if we don't inject `-L<foo prefix>/lib` these end up using system dirs.

We could of course consider changing the compiler wrapper behavior for build systems that have a certain degree of strictness like scons/bazel/...

iarspider:
> There's tons of build system that will detect `libfoo.so` by just running `cc test.c -lfoo`, and if we don't inject `-L<foo prefix>/lib` these end up using system dirs.

> We could of course consider changing the compiler wrapper behavior for build systems that have a certain degree of strictness like scons/bazel/...

This is why the PR keeps the reordering on by default, but allows package maintainer to explicitly turn it off if they know what they are doing.

haampie:
> This is why the PR keeps the reordering on by default, but allows package maintainer to explicitly turn it off if they know what they are doing.

I think I missed the second commit initially. That makes it better. You're not solving the runtime issue though, rpaths are still injected by Spack, I would say those should be dropped then too, right?

smuzaffar:
> I wish we could do that, but we can't, at least not everywhere. There's tons of build system that will detect `libfoo.so` by just running `cc test.c -lfoo`, and if we don't inject `-L<foo prefix>/lib` these end up using system dirs.

In this case there should be a post build check to make sure that libs are properly linked from spack built external area and not from system. Currently we are building rpms for `cmssw` and its externals/dependencies and `rpm` dependency checking let us catch such issues.

haampie:
Sure, we can detect shared linking issues post-install, but then we still need to prevent it from happening, and the compiler wrapper is the easiest way. 

Alternatively we need to modify tons of packages, and maybe -L flags only need to be passed when particular combinations of configure flags are given, ... it's virtually impossible to guarantee reproducibility and working packages if we don't hide system dirs behind spack dirs.

Note that for static linking there is no easy post-install check, so the compiler wrapper remains the best solution there too.

smuzaffar:
@iarspider , looks like this is still not complete . We also need to avoid spack adding `-I` and `-L` for all the dependencies e.g.
- https://github.com/spack/spack/blob/e3170529b350513d3bb13d0b522035becec49c29/lib/spack/env/cc#L655-L666
- https://github.com/spack/spack/blob/e3170529b350513d3bb13d0b522035becec49c29/lib/spack/env/cc#L699-L707

should be ignored too. I would suggest to first test your changes to build cmssw and make sure spack does not reorder include and libdir paths and also it does not add the include and libdir for dependencies

For rpath/runpath to work, I would suggest to add each `-L` in to rpath e.g. setting `rp=$arg` at https://github.com/spack/spack/blob/e3170529b350513d3bb13d0b522035becec49c29/lib/spack/env/cc#L474 should be enough.

I would still suggest to first test these changes for cmssw before updating this PR

haampie:
> I would suggest to add each -L in to rpath

This leads to issues with stub libs, for example `-L <cuda prefix>/lib64/stubs -lcuda` should not translate in `-Wl,-rpath,<cuda prefix>/lib64/stubs`. I would say your build system should add the right rpaths too then.

smuzaffar:
> This leads to issues with stub libs, for example -L <cuda prefix>/lib64/stubs -lcuda should not translate in -Wl,-rpath,<cuda prefix>/lib64/stubs

ah right @haampie , thanks for pointing this out

> I would say your build system should add the right rpaths too then.

yes we are thinking the same.

PR #33146, opened 2022-10-10
Specify GCC prefix in LLVM-based compilers

haampie:
`spack.compiler.Compiler`: introduce prefix property

We currently don't really have something that gives the GCC install path, which
is used by many LLVM-based compilers (llvm, llvm-amdgpu, nvhpc, ...) to fix the
GCC toolchain once and for all.

This `prefix` property is dynamic in the sense that it queries the compiler
itself. This is necessary because it's not easy to deduce the install path from
the spack compiler `cc` property (might be a symlink, might be a filename like
`gcc` which works by having the compiler load a module that sets the PATH
variable, might be a generic compiler wrapper based on environment variables
like on cray...).

With this new property, we can clean up some recipes that have the logic
repeated for GCC.

Also makes `intel-oneapi-compilers` behave like `llvm` and `nvhpc`, it now
fixes the GCC toolchain based on what `%gcc` was used, instead of doing a
system search.

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
@@ -5,9 +5,9 @@
 
 import platform
 
+import spack.compilers
 from spack.build_environment import dso_suffix
 from spack.package import *
-import spack.compilers
 
 linux_versions = [
     {
@@ -119,7 +119,9 @@ class IntelOneapiCompilers(IntelOneApiPackage):
     # instead.
     for __compiler in spack.compilers.supported_compilers():
         if __compiler != "gcc":
-            conflicts("%{0}".format(__compiler), msg="intel-oneapi-compilers must be installed with %gcc")
+            conflicts(
+                "%{0}".format(__compiler), msg="intel-oneapi-compilers must be installed with %gcc"
+            )
 
     if platform.system() == "Linux":
         for v in linux_versions:
--- a/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
@@ -7,6 +7,7 @@ import platform
 
 from spack.build_environment import dso_suffix
 from spack.package import *
+import spack.compilers
 
 linux_versions = [
     {
@@ -113,6 +114,13 @@ class IntelOneapiCompilers(IntelOneApiPackage):
 
     depends_on("patchelf", type="build")
 
+    # TODO: effectively gcc is a direct dependency of intel-oneapi-compilers, but we
+    # cannot express that properly. For now, add conflicts for non-gcc compilers
+    # instead.
+    for __compiler in spack.compilers.supported_compilers():
+        if __compiler != "gcc":
+            conflicts("%{0}".format(__compiler), msg="intel-oneapi-compilers must be installed with %gcc")
+
     if platform.system() == "Linux":
         for v in linux_versions:
             version(v["version"], expand=False, **v["cpp"])
@@ -186,7 +194,9 @@ class IntelOneapiCompilers(IntelOneApiPackage):
         # TODO: it is unclear whether we should really use all elements of
         #  _ld_library_path because it looks like the only rpath that needs to be
         #  injected is self.component_prefix.linux.compiler.lib.intel64_lin.
-        flags = " ".join(["-Wl,-rpath,{0}".format(d) for d in self._ld_library_path()])
+        flags_list = ["-Wl,-rpath,{}".format(d) for d in self._ld_library_path()]
+        flags_list.append("--gcc-toolchain={}".format(self.compiler.prefix))
+        flags = " ".join(flags_list)
         for cmp in [
             "icx",
             "icpx",
--- a/lib/spack/spack/compiler.py
+++ b/lib/spack/spack/compiler.py
@@ -537,6 +537,14 @@ class Compiler(object):
             )
             return self.extract_version_from_output(output)
 
+    @property
+    def prefix(self):
+        """Query the compiler for its install prefix. This is the install
+        path as reported by the compiler. Note that paths for cc, cxx, etc
+        are not enough to find the install prefix of the compiler, since
+        the can be symlinks, wrappers, or filenames instead of absolute paths."""
+        raise NotImplementedError("prefix is not implemented for this compiler")
+
     #
     # Compiler classes have methods for querying the version of
     # specific compiler executables.  This is used when discovering compilers.
--- a/lib/spack/spack/compilers/gcc.py
+++ b/lib/spack/spack/compilers/gcc.py
@@ -6,8 +6,11 @@
 import os
 import re
 
+from llnl.util.filesystem import ancestor
+
 import spack.compiler
 import spack.compilers.apple_clang as apple_clang
+import spack.util.executable
 from spack.version import ver
 
 
@@ -196,3 +199,21 @@ class Gcc(spack.compiler.Compiler):
     @property
     def stdcxx_libs(self):
         return ("-lstdc++",)
+
+    @property
+    def prefix(self):
+        # GCC reports its install prefix when running ``-print-search-dirs``
+        # on the first line ``install: <prefix>``.
+        cc = spack.util.executable.Executable(self.cc)
+        with self.compiler_environment():
+            gcc_output = cc("-print-search-dirs", output=str, error=str)
+
+            for line in gcc_output.splitlines():
+                if line.startswith("install:"):
+                    gcc_prefix = line.split(":")[1].strip()
+                    # Go from <prefix>/lib/gcc/<triplet>/<version>/ to <prefix>
+                    return ancestor(gcc_prefix, 4)
+
+            raise RuntimeError(
+                "could not find install prefix of GCC from output:\n\t".format(gcc_output)
+            )
--- a/lib/spack/spack/compilers/gcc.py
+++ b/lib/spack/spack/compilers/gcc.py
@@ -215,5 +215,5 @@ class Gcc(spack.compiler.Compiler):
                     return ancestor(gcc_prefix, 4)
 
             raise RuntimeError(
-                "could not find install prefix of GCC from output:\n\t".format(gcc_output)
+                "could not find install prefix of GCC from output:\n\t{}".format(gcc_output)
             )
--- a/var/spack/repos/builtin/packages/llvm-amdgpu/package.py
+++ b/var/spack/repos/builtin/packages/llvm-amdgpu/package.py
@@ -234,18 +234,7 @@ class LlvmAmdgpu(CMakePackage):
 
         # Get the GCC prefix for LLVM.
         if self.compiler.name == "gcc":
-            compiler = Executable(self.compiler.cc)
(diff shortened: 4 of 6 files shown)


haampie:
Notice that we're still picking up the wrong `lib{gcc_s,stdc++}.so` at runtime for C++ sources compiled with intel-oneapi-compilers, due to missing rpaths to the compiler support library dir.

Also note that intel-oneapi-compilers's `libsyscl.so` depends on `libstdc++`, and is lacking rpaths too.

In principle we can just use `SPACK_COMPILER_IMPLICIT_RPATHS` in `def install` etc and run `patchelf --set-rpath ...`.

var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
```diff
@@ -113,6 +114,15 @@ class IntelOneapiCompilers(IntelOneApiPackage):
 
     depends_on("patchelf", type="build")
 
+    # TODO: effectively gcc is a direct dependency of intel-oneapi-compilers, but we
+    # cannot express that properly. For now, add conflicts for non-gcc compilers
+    # instead.
+    for __compiler in spack.compilers.supported_compilers():
+        if __compiler != "gcc":
```
trws, line 121: I see where you're coming from, this may be the best we can do for now, but is it that gcc is a direct dependency or that we need the system toolchain that's in use? The latter might let us propagate it transitively through other compilers as long as we can query them for the appropriate thing.

trws: More concretely, for clang-based compilers for example we could use the `Selected GCC installation` from clang -v, or the result of `dirname $(clang -print-libgcc-file-name)`.

haampie: It's also a direct dependency of the runtime libraries, although that's not covered by this PR:

```
$ readelf -d ./compiler/2022.2.0/linux/lib/libsycl.so | grep 'libgcc_s.so'
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
```

haampie: So, I was thinking to submit another PR that would `patchelf --add-rpath <gcc compiler support library dir> <intel oneapi compiler support library>`

trws: That makes sense, but all the places this is happening only add the GCC stuff if the compiler is gcc, so compiling with clang we still don't get a toolchain set right?

haampie: Yeah, in the PR I've added conflicts to only ever allow GCC as a "compiler" for intel-oneapi-compilers (just s.t. we can express the dependency on GCC support libs).

haampie: Regarding `llvm%clang`, we could introspect `clang` and have it report its chosen gcc toolchain, and use that to fix `GCC_INSTALL_PREFIX` once and for all too... but that would another PR.

trws (approved):

haampie:
@spackbot run pipeline

stephenmsachs:
This PR breaks `cmake%intel`. Reproducer:

```
git clone https://github.com/spack/spack.git
cd spack/
source share/spack/setup-env.sh 
spack install intel-oneapi-compilers
. "$(spack location -i intel-oneapi-compilers)/setvars.sh"
spack compiler add --scope site
spack install cmake%intel
```

Output:
```
==> Installing cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd
==> No binary for cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd found: installing from source
==> Warning: Intel's compilers may or may not optimize to the same degree for non-Intel microprocessors for optimizations that are not unique to Intel microprocessors
==> Fetching https://github.com/Kitware/CMake/releases/download/v3.24.2/cmake-3.24.2.tar.gz
==> No patches needed for cmake   
==> cmake: Executing phase: 'bootstrap'
==> Error: ProcessError: Command exited with status 11:
    './bootstrap' '--prefix=/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd' '--parallel=16' '--no-system-libs' '--no-qt-gui' '--' '-DCMAKE_BUILD_TYPE=Release' '-DCMake_TEST_INSTALL=OFF' '-DCMAKE_USE_OPENSSL=ON' '-DBUILD_CursesDialog=True' '-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON' '-DCMAKE_INSTALL_RPATH=/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd/lib;/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd/lib64;/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/ncurses-6.3-5sdhoxvcggmeysvawbmz2jmzd3z5u3fd/lib;/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/openssl-1.1.1q-lyvkhdkeemntujkxfzw2ux4iyjujce6c/lib;/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/zlib-1.2.12-uoldnmbrzup76pr4lfwfwh7a2d2kzk66/lib' '-DCMAKE_PREFIX_PATH=/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/openssl-1.1.1q-lyvkhdkeemntujkxfzw2ux4iyjujce6c;/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/zlib-1.2.12-uoldnmbrzup76pr4lfwfwh7a2d2kzk66;/tmp/spack/opt/spack/linux-amzn2-zen2/intel-2021.7.0/ncurses-6.3-5sdhoxvcggmeysvawbmz2jmzd3z5u3fd'

2 errors found in build log:
     839    -- Checking if compiler supports C++ make_unique
     840    -- Checking if compiler supports C++ make_unique - no
     841    -- Checking if compiler supports C++ unique_ptr
     842    -- Checking if compiler supports C++ unique_ptr - no
     843    -- Checking if compiler supports C++ filesystem
     844    -- Checking if compiler supports C++ filesystem - no
  >> 845    CMake Error at CMakeLists.txt:112 (message):
     846      The C++ compiler does not support C++11 (e.g.  std::unique_ptr).
     847
     848
     849    -- Configuring incomplete, errors occurred!
     850    See also "/tmp/ec2-user/spack-stage/spack-stage-cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd/spack-src/CMakeFiles/CMakeOutput.log".
     851    See also "/tmp/ec2-user/spack-stage/spack-stage-cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd/spack-src/CMakeFiles/CMakeError.log".
     852    ---------------------------------------------
  >> 853    Error when bootstrapping CMake:
     854    Problem while running initial CMake
     855    ---------------------------------------------
```

The cmake checks for `make_unique`, etc. also produce 'yes' before this PR.

rscohn2:
Does https://github.com/spack/spack/pull/33281#issuecomment-1277622839 fix it?

I can't reproduce the problem. I guess it depends on the system gcc available.

stephenmsachs:
> Does [#33281 (comment)](https://github.com/spack/spack/pull/33281#issuecomment-1277622839) fix it?

Unfortunately no.

> I can't reproduce the problem. I guess it depends on the system gcc available.

Yes, you are probably right. This is gcc-7.3.1

rscohn2:
I expected that your PR would undo the problem that this PR caused for icc and restore old behavior, but something else seems to be happening. I am installing old compilers so I can try to reproduce the problem. Can you upload the logs and env files for the cmake build that failed?

stephenmsachs:
I'll get to it right away. I am installing `intel-oneapi-compilers` on top of newer gcc right now to see whether it eliminates the cmake issue.

stephenmsachs:
I get the same error for :
```
-- linux-amzn2-zen2 / gcc@12.2.0 --------------------------------
intel-oneapi-compilers@2022.2.0
````
here are the logs:
[CMakeError.log](https://github.com/spack/spack/files/9786742/CMakeError.log)
[CMakeOutput.log](https://github.com/spack/spack/files/9786743/CMakeOutput.log)
[spack-build-env.txt](https://github.com/spack/spack/files/9786744/spack-build-env.txt)
[spack-build-out.txt](https://github.com/spack/spack/files/9786745/spack-build-out.txt)

**Comment by rscohn2** [comment]
I think I understand the problem:
```
/tmp/spack/lib/spack/env/intel/icpc   -std=gnu++17 -MD -MT CMakeFiles/cmTC_eb3ad.dir/cm_cxx_filesystem.cxx.o -MF CMakeFiles/cmTC_eb3ad.dir/cm_cxx_filesystem.cxx.o.d -o CMakeFiles/cmTC_eb3ad.dir/cm_cxx_filesystem.cxx.o -c /tmp/ec2-user/spack-stage/spack-stage-cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd/spack-src/Source/Checks/cm_cxx_filesystem.cxx
icpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.
icpc: command line warning #10148: option '--gcc-toolchain=/tmp/spack/opt/spack/linux-amzn2-zen/gcc-7.3.1/gcc-12.2.0-rzixe57jqsugv3xxll4nwglzkix4bst2' not supported
/tmp/ec2-user/spack-stage/spack-stage-cmake-3.24.2-kagbqz6qrgbdgwfwqoa45yloqn5fzrzd/spack-src/Source/Checks/cm_cxx_filesystem.cxx(2): catastrophic error: cannot open source file "filesystem"
  #include <filesystem>
```
It is trying to test for c++17 support and failing. The --gcc-toolchain would have helped, but it works for icpx and not icpc. It needs to use a different option when the compiler is icpc.

To work around it, I think you can `spack load gcc@12.2.0` so a more modern gcc is in your path. You can verify which compiler icpc is using this way:
```
rscohn1@anpfclxlin02:spack$ icpc ../test.cpp -v
icpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second ha\
lf of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this c\
ompiler. Use '-diag-disable=10441' to disable this message.
icpc version 2021.7.0 (gcc version 7.3.0 compatibility)
```

stephenmsachs:
You are right. I added this use case to https://github.com/spack/spack/pull/33281.

PR #33150, opened 2022-10-10
 add ANL mpi-serial package (used by parallelio)

jedwards4b:
The mpi-serial package is needed to build a single processor version of the parallelio library.
This single processor version is useful for testing and is used in some single processor applications.


tldahlgren (approved):
LGTM. Confirmed sha256.

PR #33152, opened 2022-10-10
Add checksum for py-vector 0.8.5

iarspider:


tldahlgren (commented):
Confirmed sha256.

adamjstewart (approved):
Wheel is already in base class so could be removed

PR #33155, opened 2022-10-10
Update py werkzeug

iarspider:


var/spack/repos/builtin/packages/py-werkzeug/package.py
```diff
@@ -9,9 +9,10 @@
 class PyWerkzeug(PythonPackage):
     """The Swiss Army knife of Python web development"""
 
-    homepage = "http://werkzeug.pocoo.org"
+    homepage = "https://werkzeug.palletsprojects.com/en/2.2.x/"
```
tldahlgren: Confirmed URL exists and that the old one redirects to it.

Thanks for updating to a URL with the `https` protocol.

tldahlgren (commented):
Confirmed new and, due to URL change, old sha256.

adamjstewart (dismissed):

var/spack/repos/builtin/packages/py-werkzeug/package.py
adamjstewart:
```suggestion
    homepage = "https://werkzeug.palletsprojects.com"
```
If you remove the version from the URL it will automatically update for the next release

adamjstewart (approved):

PR #33163, opened 2022-10-10
openblas confuses flang/flang-new, so do not set TIME with ~fortran

s-sajid-ali:
With `%clang-15.0.2` and `~fortran` (but with `flang-new` available as part of `clang/llvm-15.0.2`), I get the following error:
 ```
  >> 82500    flang-new: error: unknown argument '-Mrecursive'; did you mean '-frecursive'?
  >> 82501    flang-new: error: unknown argument: '-Kieee'
 ```
 This is because spack currently sets `"TIMER=INT_CPU_TIME"` which compiles a fortran file. This becomes an issue as `OpenBLAS` misclassifies `flang-new` to be `flang` and sets flags that are invalid for `flang-new`. To avoid these issues, this PR sets the `TIMER` definition only when `+fortran` is enabled.


haampie (dismissed):

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -379,8 +379,9 @@ def make_defs(self):
         if "+consistent_fpcsr" in self.spec:
             make_defs += ["CONSISTENT_FPCSR=1"]
 
-        # Flang/f18 does not provide ETIME as an intrinsic
-        if self.spec.satisfies("%clang"):
+        # Flang/f18 does not provide ETIME as an intrinsic.
+        # Do not set TIMER variable if fortran is disabled.
+        if self.spec.satisfies("%clang+fortran"):
```
haampie:
```suggestion
        if self.spec.satisfies("+fortran%clang"):
```

is a bit more clear imho

haampie (approved):

alalazo:
@spackbot run pipeline

PR #33164, opened 2022-10-10
Omega-H: Current constraint doesn't allow any cuda

kwryankrattiger:
From the issue referenced, it seems later and earlier versions of cuda work.


cwsmith (approved):
Good catch.  Thank you for the fix.

eugeneswalker (approved):

PR #33175, opened 2022-10-10
r-rjava: fix pcre dependency

glennpj:
fix pcre dependency for r-rjava

### DIFF
--- a/var/spack/repos/builtin/packages/r-rjava/package.py
+++ b/var/spack/repos/builtin/packages/r-rjava/package.py
@@ -28,7 +28,7 @@ class RRjava(RPackage):
     depends_on("bzip2")
     depends_on("icu4c")
     depends_on("libiconv")
-    depends_on("pcre2")
+    depends_on("pcre")
     depends_on("xz")
 
     def setup_build_environment(self, env):


var/spack/repos/builtin/packages/r-rjava/package.py
```diff
@@ -28,7 +28,7 @@ class RRjava(RPackage):
     depends_on("bzip2")
     depends_on("icu4c")
     depends_on("libiconv")
-    depends_on("pcre2")
+    depends_on("pcre")
```
tldahlgren, line 31: Just to confirm.  This dependency applies to *all* versions of `r-rjava`, or at least all of the ones listed in the package?

tldahlgren (commented):
LGTM. Quick question.

var/spack/repos/builtin/packages/r-rjava/package.py
glennpj: Your question prompted me to look at this more. I am observing oddities but I am not sure they are related to pcre2. After rebuilding this several times, I do see failures during linking, but not all of the time. I can not reproduce it consistently and I am now thinking this is not a pcre2 problem. It is possible for both pcre and pcre2 to be in the spec (glib uses pcre) so I wonder if that might be an issue at configure time if the order in the environment shifts between builds. The bottom line is this package should work with pcre2 and I now see that it builds properly with that most of the time. I will have to poke at this more but I am going to close this PR as I do not think it is the correct fix.

glennpj:
I am going to close this PR as the problem that I ran in to turns out to be intermittent and needs to be identified.

PR #33179, opened 2022-10-11
r-do-db: add bioc attribute

glennpj:
Add bioc attribute to r-do-db package


tldahlgren (approved):
LGTM (after checking the translation in `r.py` :grin:)

PR #33183, opened 2022-10-11
Enable tests for rocm packages - hsakmt-roct and rocm-smi-lib

srekolam:


var/spack/repos/builtin/packages/rocm-smi-lib/package.py
```diff
@@ -119,3 +130,32 @@ def post_install(self):
...
+        with working_dir(test_dir, create=True):
+            cmake_bin = join_path(self.spec["cmake"].prefix.bin, "cmake")
+            prefixes = ";".join(
+                [
+                    self.spec["rocm-smi-lib"].prefix,
+                ]
+            )
+            rocm_smi_lib_path = ";".join([self.spec["rocm-smi-lib"].prefix.lib])
+            cc_options = [
+                "-DCMAKE_PREFIX_PATH=" + prefixes,
```
afzpatel, line 156:
```suggestion
                "-DCMAKE_PREFIX_PATH=" + prefixes,
                "-DROCM_DIR=" + self.spec["rocm-smi-lib"].prefix,
```
I don't think it's necessary to set RSMI_LIB_DIR since it will already be set in the cmake file

var/spack/repos/builtin/packages/rocm-smi-lib/package.py
```diff
@@ -119,3 +130,32 @@ def post_install(self):
...
+                "-DRSMI_LIB_DIR=" + rocm_smi_lib_path,
```
afzpatel, line 157:
```suggestion
                "-DROCM_DIR=" + self.spec["rocm-smi-lib"].prefix,
```
I don't think it's necessary to set RSMI_LIB_DIR since it will already be set in the cmake file but we do need to set ROCM_DIR

var/spack/repos/builtin/packages/rocm-smi-lib/package.py
```diff
@@ -119,3 +130,32 @@ def post_install(self):
...
+                ".",
+            ]
+            self.run_test(cmake_bin, cc_options)
```
afzpatel, line 160:
```suggestion
            self.run_test(cmake_bin, cc_options)
            make()
            self.run_test("rsmitst64")
```

alalazo:
What is the status of this PR?

srekolam:
i am closing this PR as there is an updated PR that is submitted recently

PR #33185, opened 2022-10-11
Add checksum for py-keyring 23.9.1

iarspider:


var/spack/repos/builtin/packages/py-keyring/package.py
```diff
@@ -42,5 +43,6 @@ class PyKeyring(PythonPackage):
     depends_on("py-importlib-metadata@3.6:", when="@23:", type=("build", "run"))
```
adamjstewart: importlib-metadata isn't required in Python 3.10+

adamjstewart (changes requested):

adamjstewart (approved):

PR #33190, opened 2022-10-11
Add compatibility between RHEL8, Rocky8 and AlmaLinux8

iarspider:


iarspider:
Let's hope GitHub is better now...

alalazo (dismissed):
I need to double check this. This file was made to ensure compatibility on the same OS but different versions, as the comments at the top suggest. If the logic allows bidirectional compatibility, then I think:
```
os_compatible("rhel8", "almalinux8")
os_compatible("almalinux8", "rhel8")
```
would be better, unless it's only AlmaLinux binaries that can be reused on RHEL8 and not vice-versa.

iarspider:
@alalazo As far as I know, all EL8 clones (Alma, Rocky, CentOS Stream, RHEL) are [compatible](https://docs.google.com/spreadsheets/d/13M4Qf2va0-xJJr24adXYj0xk9aD20MlucgyR48Ng0lQ/edit#gid=705855033)

bernhardkaindl:
@iarspider: As alalazo suggested, the compatibility is likely both ways: Does alalazo's suggestion work as well?

iarspider:
@bernhardkaindl I was waiting for him to double-check that. Sure, I can add extra compatibility pairs.

bernhardkaindl (approved):
@alalazo: Bi-directional entries for `os_compatible` (rhel8, rocky8, almalinux8) are added now

bernhardkaindl:
@alalazo Please approve that your requested changes are incorporated,

We need this to fix the OpenBLAS spack compiler dilemma on RHEL7/8-derived distros.
https://github.com/xianyi/OpenBLAS/issues/3805#issuecomment-1319878852

alalazo:
Can you double check if this had any performance impact on the concretizer on those platforms? Or I can get to it later.

bernhardkaindl:
> Can you double check if this had any performance impact on the concretizer on those platforms? Or I can get to it later.

@alalazo I don't have one of those platforms, if you can check later, it would be great! (@iarspider: ping too)

alansill:
What sets up this `os_compatible` flag and how is it used? In the case of Red Hat Enterprise Linux workalikes, the equivalence would seem to me to be best derived from the value of the `PLATFORM_ID` in `/etc/os-release` -- for example, `PLATFORM_ID="platform:el8"` for the RHEL8/rocky8/alma8 family.

PR #33191, opened 2022-10-11
Relocate Text Dependencies

stephenmsachs:
When downloading from binary cache not only replace RPATHs to dependencies, but also text references to dependencies.

Example:
`autoconf@2.69` contains a text reference to the executable of its dependency `perl`:

```
$ grep perl-5 /shared/spack/opt/spack/linux-amzn2-x86_64_v3/gcc-7.3.1/autoconf-2.69-q3lo/bin/autoreconf
eval 'case $# in 0) exec /shared/spack/opt/spack/linux-amzn2-x86_64_v3/gcc-7.3.1/perl-5.34.1-yphg/bin/perl -S "$0";; *) exec /shared/spack/opt/spack/linux-amzn2-x86_64_v3/gcc-7.3.1/perl-5.34.1-yphg/bin/perl -S "$0" "$@";; esac'
```

These references need to be replace or any package using `autoreconf` will fail as it cannot find the installed `perl`.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -698,7 +698,7 @@ def write_buildinfo_file(spec, workdir, rel=False):
     prefix_to_hash = dict()
     prefix_to_hash[str(spec.package.prefix)] = spec.dag_hash()
     deps = spack.build_environment.get_rpath_deps(spec.package)
-    for d in deps + spec.dependencies(type="run"):
+    for d in deps + spec.dependencies(deptype="run"):
         prefix_to_hash[str(d.prefix)] = d.dag_hash()
 
     # Create buildinfo data and write it to disk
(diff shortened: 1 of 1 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -698,7 +698,7 @@ def write_buildinfo_file(spec, workdir, rel=False):
     prefix_to_hash = dict()
     prefix_to_hash[str(spec.package.prefix)] = spec.dag_hash()
     deps = spack.build_environment.get_rpath_deps(spec.package)
-    for d in deps:
+    for d in deps + spec.dependencies():
```
alalazo: I think this should be `spec.dependencies(type="run")` since build dependencies shouldn't be affected when installing from binaries. Also, it might be better to give this set of specs a name and extract a function.

alalazo (changes requested):
I just double checked, and since `perl` is already a `run` dependency of `autoconf` this is a real bug. The solution can probably be refined a little.

lib/spack/spack/binary_distribution.py
stephenmsachs: Is there any harm in including "build" dependencies? I can imagine that some package authors do not realize when a "build" dependency actually also is a "run" dependency because spack writes the full path to the dependency's executable into some run script of the dependent.

alalazo: It would be wrong from a modeling perspective. If we realize that a "build" dependency is also "run" we should modify the package, since the model affect contexts beyond just relocation.

stephenmsachs: OK, I'll change it.

alalazo (approved):

alalazo:
Merging as is since it fixes a bug, and it's very limited in scope. We probably need to go back and refactor this part of the code for better clarity (more meaningful names, gather similar code in the same place etc.).

PR #33192, opened 2022-10-11
Intel oneapi compilers: silence linker flag warnings for old versions

haampie:
oneapi: set -Wno-unused-command-line-argument

For older versions of intel-oneapi-compilers, running the compiler in
preprocessor / compilation mode would trigger warnings that
`-Wl,-rpath,...` flags were unused.

This in turn caused certain configure scripts to fail as they did not
expect output from the compiler (it's treated as an error). Notably
cmake's bootstrap phase failed to detect c++ features of the compiler.

As a workaround, add this flag to silence the warning, since I don't
think we can scope the flags to compile+link mode.

This PR builds on top of #33146

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
@@ -197,6 +197,16 @@ class IntelOneapiCompilers(IntelOneApiPackage):
         #  _ld_library_path because it looks like the only rpath that needs to be
         #  injected is self.component_prefix.linux.compiler.lib.intel64_lin.
         flags_list = ["-Wl,-rpath,{}".format(d) for d in self._ld_library_path()]
+
+        # Older versions trigger -Wunused-command-line-argument warnings whenever
+        # linker flags are passed in preprocessor (-E) or compilation mode (-c).
+        # The cfg flags are treated as command line flags apparently. Newer versions
+        # do not trigger these warnings. In some build systems these warnings can
+        # cause feature detection to fail, so we silence them with -Wno-unused-...
+        if self.spec.version < Version("2022.1.0"):
+            flags_list.append("-Wno-unused-command-line-argument")
+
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/llvm-amdgpu/package.py
```diff
@@ -234,18 +234,7 @@ def cmake_args(self):
...
-            gcc_prefix = ""
-            for line in gcc_output.splitlines():
-                if line.startswith("install:"):
-                    # Get path and strip any whitespace
-                    # (causes oddity with ancestor)
-                    gcc_prefix = line.split(":")[1].strip()
-                    gcc_prefix = ancestor(gcc_prefix, 4)
-                    break
-            args.append(self.define("GCC_INSTALL_PREFIX", gcc_prefix))
+            args.append(self.define("GCC_INSTALL_PREFIX", self.compiler.prefix))
```
rscohn2: I can't find any information about GCC_INSTALL_PREFIX. Is it a feature of llvm compilers, but not intel? Or is it something spack specific?

rscohn2 (dismissed):

var/spack/repos/builtin/packages/llvm-amdgpu/package.py
haampie: It's basically setting a compile time default for `--gcc-toolchain`, so we can only use it for open source `llvm` based compilers like clang itself and AMD's `llvm-amdgpu` fork. With closed-source `nvhpc` / `oneapi` compilers we have to rely on config files to accomplish roughly the same.

Without GCC_INSTALL_PREFIX / config files, llvm does a system scan for the toolchain, and will not be able to locate Spack's GCCs

haampie:
Rebased now that #33146 was merged

rscohn2 (approved):
Thanks for limiting this to older versions instead of unconditional.

PR #33200, opened 2022-10-11
eckit: update versions

dominichofer:
Updates versions and SHAs of eckit.

### DIFF
--- a/var/spack/repos/builtin/packages/eckit/package.py
+++ b/var/spack/repos/builtin/packages/eckit/package.py
@@ -16,12 +16,8 @@ class Eckit(CMakePackage):
     maintainers = ["skosukhin"]
 
     version("1.20.2", sha256="9c11ddaaf346e40d11312b81ca7f1b510017f26618f4c0f5c5c59c37623fbac8")
-    version("1.19.0", sha256="a5fef36b4058f2f0aac8daf5bcc9740565f68da7357ddd242de3a5eed4765cc7")
-    version("1.18.7", sha256="862b4d1a15020f33ec678791b1ce059efd50a73ef1baabc271a8d1700ead70bd")
-    version("2021.12.0", sha256="2ec9f837b416324e7c732aec7cdad9c14c8c97df3ba4a34eabc0935a2503b22e")
-    version("2021.08.0", sha256="42b03dfffecf084dd9628becdf5c70bea7d94f0691d2f9869a5616c638381fc6")
-    version("1.17.1", sha256="f5cea390ba38ce11c00e1dc12b6ac6d43b247edf82be07d1568724d62faab870")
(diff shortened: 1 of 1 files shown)


skosukhin:
This removes version `1.16.0` without a deprecation period. Is there a reason to even deprecate it? I'm also not sure about `YYYY.MM.X` versions. I have no idea what they mean. For example, [`2021.12.0`](https://github.com/ecmwf/eckit/releases/tag/2021.12.0) is annotated with `1.18.2` but there is a separate tag [`1.18.2`](https://github.com/ecmwf/eckit/releases/tag/1.18.2). Until we figure it out, I suggest we use only `1.XX.Y` versions.

alalazo (changes requested):
Can you take care of @skosukhin review?

bernhardkaindl:
@skosukhin, [alalazo](https://github.com/alalazo) wrote:
> Can you take care of @skosukhin review?

Would you approve (this needs to be merged before #33289) and notify alalazo to approve as well then?

bernhardkaindl:
@skosukhin ping?

skosukhin:
@bernhardkaindl my https://github.com/spack/spack/pull/33200#issuecomment-1274668677 is not addressed yet.

bernhardkaindl:
@PanicSheep: I looked up the linked PR which prompted you to add the new versions and only add the 0.16.3 and 1.20.2:

I'd propose that unless you'd need all those versions have a checksum: When asked, spack can build any of these versions even without a checksum.

Specially we should not add the 2021.x.x versions because they take precedence over newer versions such as 1.20.2.

@PanicSheep Because you didn't respond you [skosukhin](https://github.com/skosukhin) (the maintainer)'s change request so far I hope it's ok if we merge this soon even if you didn't respond, because none of these versions appear to be needed by the PR which prompted this PR.

bernhardkaindl:
[skosukhin](https://github.com/skosukhin) reviewed:
> This removes version `1.16.0` without a deprecation period. Is there a reason to even deprecate it? I'm also not sure about `YYYY.MM.X` versions. I have no idea what they mean. For example, [`2021.12.0`](https://github.com/ecmwf/eckit/releases/tag/2021.12.0) is annotated with `1.18.2` but there is a separate tag [`1.18.2`](https://github.com/ecmwf/eckit/releases/tag/1.18.2). Until we figure it out, I suggest we use only `1.XX.Y` versions.

Thanks, I've addressed your review and removed the 2021.x.x versions because they took precedence over 1.20.2 and I'd also skip the other intermediate versions, maybe except for 1.16.3 but otherwise as @PanicSheep noted in the linked PR that they appear compatible, I'd my commit just leaves 1.16.3 and 1.20.2 to be added.

skosukhin:
LGTM

bernhardkaindl:
@alalazo: skosukhin sent a "LGTM" now, can you select that your requested changes were adressed? Thanks!

alalazo (approved):

PR #33209, opened 2022-10-11
Add checksum for py-regex 2022.8.17

iarspider:


var/spack/repos/builtin/packages/py-regex/package.py
```diff
@@ -25,4 +27,5 @@ class PyRegex(PythonPackage):
         "2017.07.11", sha256="dbda8bdc31a1c85445f1a1b29d04abda46e5c690f8f933a9cc3a85a358969616"
     )
 
-    depends_on("py-setuptools", type="build", when="@:2018,2020:")
+    depends_on("py-setuptools", when="@:2018,2020:", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-setuptools", type="build")
```
No idea why the dep was the way it was, but all versions use setuptools.

adamjstewart (changes requested):

adamjstewart (approved):

PR #33231, opened 2022-10-12
py-xopen: version bump to 1.6.0

luke-dt:


var/spack/repos/builtin/packages/py-xopen/package.py
```diff
@@ -15,16 +15,21 @@ class PyXopen(PythonPackage):
...
 
+    depends_on("python@3.7:", type=("build", "run"), when="@1.5.0:")
+    depends_on("python@3.6:", type=("build", "run"), when="@1.1.0:")
     depends_on("python@3.5:", type=("build", "run"), when="@0.9.0:")
     depends_on("python@2.7,3.4:", type=("build", "run"), when="@0.5:0.8")
     depends_on("python@2.6:2,3.3:", type=("build", "run"), when="@0.1.1")
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools-scm@6.2:", type="build", when="@1.2.0:")
     depends_on("py-setuptools-scm", type="build")
     depends_on("py-bz2file", type=("build", "run"), when="@0.5: ^python@:2.8")
```
adamjstewart, line 36: Missing dep:
```
        isal>=1.0.0; platform.python_implementation == 'CPython' and (platform.machine == "x86_64" or platform.machine == "AMD64")
```

luke-dt: oops! How would I specify that platform and arch in the spec?

adamjstewart: That is a great question. I think it would be `when="target=x86_64:"`, but @alalazo may be able to confirm whether or not AMD64 is included in that.

luke-dt: Is there anything I need to specify for `platform.python_implementation == 'CPython'` ?

adamjstewart: Nah, Spack currently only support CPython. Someday I'm hoping to add PyPy, Jython, IronPython, Intel Python, etc., but we can fix those kinds of things when that happens. Currently, there's no way to have a virtual package that is also extendable.

luke-dt: Then is this truly a dependency if the condition is `platform.python_implementation == 'CPython' and (platform.machine == "x86_64" or platform.machine == "AMD64"` ? Like would this ever be relevant for a spack install? Sorry for basic questions, I'm still learning spack's inner workings.

adamjstewart: Yes, in Spack, `python_implementation` will always be `CPython` (it's the only version of Python we support right now) and `x86_64` is the most common platform.

adamjstewart (approved):

PR #33235, opened 2022-10-12
gptl: new version 8.1.1; use the correct mpi fortran compiler

wu-kan:


tldahlgren (dismissed):
LGTM. Confirmed sha256. 

Will give at least one of @edwardhartnett @kgerheiser @Hang-Lei-NOAA @jmrosinski a chance to comment before merging.

jmrosinski (changes requested):
I am at a bit of a disadvantage since I don't know much about python, or how to use spac, and these python scripts aren't part of the committed repo. So I'll leave it up to Ed Hartnett to decide how to proceed.

BUT, it looks to me like the "if +pmpi" test is incorrectly formulated. The GPTL configure script determines by itself whether MPI is to be enabled by running internal tests. If $CC can find basic MPI routines then MPI is enabled, otherwise disabled. For example the user enables MPI by running configure with something like "env CC=mpicc ./configure",  or disables MPI by running something like "env CC=gcc ./configure". The "pmpi" test ONLY enables auto-profiling of MPI calls. MPI can still be enabled (and statistics generated for regions inside MPI regions) even if "--disable-pmpi" (this is also the default) is specified to configure.

Also the bit about "FC=" and "mpifc" seems suspect to me, unless for example spac REQUIRES mpifc and not e.g. mpifort, mpif90, etc. for the MPI-enabled Fortran compiler.

edwardhartnett (approved):

tldahlgren (approved):

PR #33242, opened 2022-10-12
py-pythran: customize headers attribute

### LINKED ISSUE #33237: linux: environment variable values grow beyond `MAX_ARG_STRLEN`, making exec fail
Currently `py-scipy` fails to build in gitlab ci, due the following:

- `py-pythran` generates 256 include dirs, which are set as a colon-seperated string in the `SPACK_INCLUDE_DIRS` environment variable;
- due to install path padding each path is at least 512 bytes in size

This results in a string of at least 128KB, which is at least `MAX_ARG_STRLEN = 32 * PAGE_SIZE = 32 * 4KB = 128KB`.

And this makes exec syscalls fail, since they have a limit on the argument length (env variables are just arguments after argv):

https://github.com/torvalds/linux/blob/49da070062390094112b423ba443ea193527b2e4/fs/exec.c#L291-L294

https://github.com/torvalds/linux/blob/49da070062390094112b423ba443ea193527b2e4/fs/exec.c#L535-L541

resulting in `E2BIG` or `Argument list too long`, which is the build error we see.

alalazo:
Fixes `py-scipy` build in gitlab (see #33237 for details)

### DIFF
--- a/var/spack/repos/builtin/packages/py-pythran/package.py
+++ b/var/spack/repos/builtin/packages/py-pythran/package.py
@@ -5,6 +5,8 @@
 
 import sys
 
+import llnl.util.filesystem as fs
+
 from spack.package import *
 
 
@@ -59,7 +61,7 @@ class PyPythran(PythonPackage):
         # Pythran is mainly meant to be used as a compiler, so return no headers to
         # avoid issue https://github.com/spack/spack/issues/33237 This can be refined
         # later to allow using pythran also as a library.
-        return []
+        return fs.HeaderList([])
 
     def patch(self):
         # Compiler is used at run-time to determine name of OpenMP library to search for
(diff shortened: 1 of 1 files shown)


alalazo:
@spackbot fix style

haampie (approved):

haampie:
I've updated the description so that we don't close #33237, since this issue of long strings in env variables can pop up in general

haampie (dismissed):

adamjstewart (dismissed):

haampie (approved):

adamjstewart (approved):

haampie:
uh, similar issues in ci?

adamjstewart:
That `py-torch-geometric` test really doesn't want to finish. Do we wait for the 6 hr timeout or boot it?

alalazo:
We need to do something to speed-up relocation. 3hrs to relocate binaries is a bit too much

haampie:
@alalazo it's your lucky day: #33248 (assuming you meant that installing dependencies is slow)

PR #33245, opened 2022-10-12
env depfile: allow deps only install

haampie:
- Refactor `spack env depfile` to use a Jinja template, making it a bit
  easier to follow as a human being.
- Add a layer of indirection in the generated Makefile through an
  `<prefix>/.install-deps/<hash>` target, which allows one to specify
  different options when installing dependencies. For example, only
  verbose/debug mode on when installing some particular spec:
  ```console
  $ spack -e my_env env depfile -o Makefile --make-target-prefix example
  $ make -j16 example/.install-deps/{hash} SPACK_INSTALL_FLAGS=--please-never-build-from-sources
  $ make -j16 example/.install/{hash} SPACK="spack -d" SPACK_INSTALL_FLAGS=--verbose
  ```

This could be used to speed up `spack ci rebuild`:
- Parallel install of dependencies from buildcache
- Better readability of logs, e.g. reducing verbosity when installing
  dependencies, and splitting logs into deps.log and current_spec.log

Given that `spack ci rebuild` is already unix only and make is a required spack dep... I think it's worth doing this.

### DIFF
--- a/lib/spack/spack/tengine.py
+++ b/lib/spack/spack/tengine.py
@@ -11,6 +11,7 @@ import six
 import llnl.util.lang
 
 import spack.config
+import spack.extensions
 from spack.util.path import canonicalize_path
 
 
--- a/lib/spack/spack/cmd/env.py
+++ b/lib/spack/spack/cmd/env.py
@@ -24,6 +24,7 @@ import spack.config
 import spack.environment as ev
 import spack.environment.shell
 import spack.schema.env
+import spack.tengine
 import spack.util.string as string
 from spack.util.environment import EnvironmentModifications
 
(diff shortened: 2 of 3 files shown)


lib/spack/spack/tengine.py
```diff
@@ -11,6 +11,7 @@
 import llnl.util.lang
 
 import spack.config
+import spack.extensions
```
trws, line 14: Is this vestigial?

haampie: I noticed it was a missing import -- I can drop it from this PR to keep the squash commit on develop clean if you want

trws (approved):
Pending  whether that import is necessary, I'm happy.

lib/spack/spack/tengine.py
trws: Nah, if it's missing leave it.

Issue #33014, opened 2022-10-05
Embrace GNU Jobserver FIFO / named pipes for composable parallelism


haampie:
### Summary

GNU Make soon :tm: supports named pipes, in fact the alpha release `4.3.90` already supports it.

The advantage of named pipes is that we don't have to worry about file descriptors closing when make is invoked many sub-processes down the line, instead we only need to ensure that the `MAKEFLAGS` env variable is propagated to the relevant process. 

A small Python example goes like this: create a named pipe `os.mkfifo(path)`, write some bytes to it, and set `MAKEFLAGS=--jobserver-auth=fifo:{path}` before running a subprocess:

```python
#!/usr/bin/env spack-python

import argparse
import os
import tempfile

from spack.util.executable import Executable

# Args: jobserver.py --jobs n command [args...]
parser = argparse.ArgumentParser(description='FIFO jobserver example')
parser.add_argument('--jobs', type=int, default=1, help='Number of jobs')
parser.add_argument('command')
parser.add_argument('args', nargs=argparse.REMAINDER)
args = parser.parse_args()

num_bytes = args.jobs - 1

# Create a FIFO in temporary dir
tmpdir = tempfile.mkdtemp()
fifopath = os.path.join(tmpdir, "jobserver")

# Write #jobs bytes
os.mkfifo(fifopath)
writefd = os.open(fifopath, os.O_RDWR)
bytes_written = os.write(writefd, b'+' * num_bytes)
assert bytes_written == num_bytes

# Set up the right MAKEFLAGS for the executable
exe = Executable(args.command)
exe.add_default_env("MAKEFLAGS", "--jobserver-auth=fifo:{}".format(fifopath))
exe(*args.args)

# Verify that all tokens were written back.
readfd = os.open(fifopath, os.O_RDONLY | os.O_NONBLOCK)
read = os.read(readfd, args.jobs)
os.close(readfd)
os.close(writefd)

if len(read) == num_bytes:
    print("All tokens were returned! :)")
else:
    print('WARNING: expected {} tokens, got {}'.format(num_bytes, len(read)))
```

If I save that as `jobserver.py` and then put this `Makefile` with 32 independent targets that sleep 1 sec in the same directory:

```make
TARGETS := $(addprefix x-,$(shell seq 32))

all: $(TARGETS)

$(TARGETS):
	sleep 1 && echo $@
```

and run

```console
$ time ./jobserver.py --jobs 8 make
real	0m4.237s
```

[...]

PR #32262, opened 2022-08-19
bootstrap binary patchelf

haampie:
- Bootstrap binary patchelf
- Do not blindly use system patchelf
- Require recent patchelf

PR #33248, opened 2022-10-12
gitlab ci install binary deps faster

haampie:
This PR does a few things:

- Parallel install of binary dependencies from buildcache (with redundant build edges pruned in the DAG)
- Improve legibility:
  1. Reduce verbosity installing buildcache deps, so that users get to see the build log for the source build
  2. Show timers for binary installation, so we get more insight into why a job is slow.
  3. Use GNU Make's output sync feature to get orderly output when installing deps in parallel
  4. Remove a few `spack -d` flags, so we get to see the build logs before hitting the max log size.

Screenshot from 2022-10-20 11-42-10
...
Screenshot from 2022-10-20 11-44-14

This PR is built on top of these parts:

- #33305
- #33315 
- #33361 
- #33363
- #33367

### DIFF
--- a/lib/spack/spack/cmd/ci.py
+++ b/lib/spack/spack/cmd/ci.py
@@ -25,7 +25,8 @@ description = "manage continuous integration pipelines"
 section = "build"
 level = "long"
 
-CI_REBUILD_INSTALL_BASE_ARGS = ["spack", "--color=always", "--backtrace", "--verbose"]
+SPACK_COMMAND = "spack"
+MAKE_COMMAND = "make"
 INSTALL_FAIL_CODE = 1
 
 
@@ -509,41 +510,88 @@ def ci_rebuild(args):
     # No hash match anywhere means we need to rebuild spec
 
     # Start with spack arguments
-    install_args = [base_arg for base_arg in CI_REBUILD_INSTALL_BASE_ARGS]
+    spack_cmd = [SPACK_COMMAND, "--color=always", "--backtrace", "--verbose"]
 
     config = cfg.get("config")
     if not config["verify_ssl"]:
-        install_args.append("-k")
+        spack_cmd.append("-k")
 
-    install_args.extend(
-        [
-            "install",
-            "--keep-stage",
-            "--use-buildcache",
-            "dependencies:only,package:never",
-        ]
-    )
+    install_args = []
 
     can_verify = spack_ci.can_verify_binaries()
     verify_binaries = can_verify and spack_is_pr_pipeline is False
     if not verify_binaries:
         install_args.append("--no-check-signature")
 
+    cdash_args = []
     if cdash_handler:
         # Add additional arguments to `spack install` for CDash reporting.
-        install_args.extend(cdash_handler.args())
+        cdash_args.extend(cdash_handler.args())
 
-    # A compiler action of 'FIND_ANY' means we are building a bootstrap
-    # compiler or one of its deps.
-    # TODO: when compilers are dependencies, we should include --no-add
-    if compiler_action != "FIND_ANY":
-        install_args.append("--no-add")
+    slash_hash = "/{}".format(job_spec.dag_hash())
+    deps_install_args = install_args
+    root_install_args = install_args + [
+        "--no-add",
+        "--keep-stage",
+        "--only=package",
+        "--use-buildcache=package:never,dependencies:only",
+        slash_hash,
+    ]
 
-    # Identify spec to install by hash
-    install_args.append("/{0}".format(job_spec.dag_hash()))
+    # ["x", "y"] -> "'x' 'y'"
+    args_to_string = lambda args: " ".join("'{}'".format(arg) for arg in args)
+
+    commands = [
+        # apparently there's a race when spack bootstraps? do it up front once
+        [
+            SPACK_COMMAND,
+            "-e",
+            env.path,
+            "bootstrap",
+            "now",
+        ],
+        [
+            SPACK_COMMAND,
+            "-e",
+            env.path,
+            "config",
+            "add",
+            "config:db_lock_timeout:120",  # 2 minutes for processes to fight for a db lock
+        ],
+        [
+            SPACK_COMMAND,
+            "-e",
+            env.path,
+            "env",
+            "depfile",
+            "-o",
+            "Makefile",
+            "--use-buildcache=package:never,dependencies:only",
+            "--make-target-prefix",
+            "ci",
+            slash_hash,  # limit to spec we're building
+        ],
+        [
+            # --output-sync requires GNU make 4.x.
+            # Old make errors when you pass it a flag it doesn't recognize,
+            # but it doesn't error or warn when you set unrecognized flags in
+            # this variable.
+            "export",
+            "GNUMAKEFLAGS=--output-sync=recurse",
+        ],
+        [
+            MAKE_COMMAND,
+            "SPACK={}".format(args_to_string(spack_cmd)),
+            "SPACK_COLOR=always",
+            "SPACK_INSTALL_FLAGS={}".format(args_to_string(deps_install_args)),
+            "-j$(nproc)",
+            "ci/.install-deps/{}".format(job_spec.dag_hash()),
+        ],
+        spack_cmd + ["install"] + root_install_args,
+    ]
 
     tty.debug("Installing {0} from source".format(job_spec.name))
-    install_exit_code = spack_ci.process_command("install", install_args, repro_dir)
+    install_exit_code = spack_ci.process_command("install", commands, repro_dir)
 
     # Now do the post-install tasks
     tty.debug("spack install exited {0}".format(install_exit_code))
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -17,7 +17,7 @@ import tempfile
 import time
 import zipfile
 
-from six import iteritems
+from six import iteritems, string_types
 from six.moves.urllib.error import HTTPError, URLError
 from six.moves.urllib.parse import urlencode
 from six.moves.urllib.request import HTTPHandler, Request, build_opener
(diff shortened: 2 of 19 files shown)


haampie:
For reference:

- https://gitlab.spack.io/spack/spack/-/jobs/3618684 (Duration: 22 minutes 17 seconds, before this PR)
- https://gitlab.spack.io/spack/spack/-/jobs/3624534 (Duration: 8 minutes 11 seconds, after this PR)

or

- https://gitlab.spack.io/spack/spack/-/jobs/3618685 (Duration: 27 minutes 53 seconds, before this PR)
- https://gitlab.spack.io/spack/spack/-/jobs/3624526 (Duration: 8 minutes 53 seconds, after this PR)

haampie:
@trws funnily enough we're hitting a broken `gmake -Orecurse` in Ubuntu 20.04 based containers :D I have no clue what they're doing... maybe they have some patches that broke it or so. The flag effectively has no effect. Fixed in a later version of Ubuntu.

haampie:
Unfortunately the GNU make situation in our builders is rather sad:

- macOS system make is 3.81(too old for --output-sync)
- Amazon Linux 2 has gmake 3.82 (too old for --output-sync)
- Ubuntu 20.04 has a broken --output-sync gmake

Bootstrapping gmake is an option...

haampie:
@spackbot run pipeline

trws:
> Unfortunately the GNU make situation in our builders is rather sad:
> 
>     * macOS system make is 3.81(too old for --output-sync)
> 
>     * Amazon Linux 2 has gmake 3.82 (too old for --output-sync)
> 
>     * Ubuntu 20.04 has a broken --output-sync gmake
> 
> 
> Bootstrapping gmake is an option...

lets do that, or grab a statically linked one, or upgrade it from the package managers, or something.

haampie:
@spackbot run pipeline

(@zackgalbreath @scottwittenburg: seems like the above has triggered on the old HEAD of this PR -- I've force pushed and if you follow the link below, you get Merge 253ae0b0 into 0de1f989 but the force pushed commit was 0e8b9fa as you can see above).

haampie:
It's a one-liner to install julia's cross-compiled version, more people should use julia as a package manager

haampie:
@spackbot fix style

haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

haampie:
@scottwittenburg / @zackgalbreath it's apparently impossible to get a pipeline to run on the latest commit if it's in "Pending — waiting for base develop commit pipeline to succeed" mode? It continues to pick a commit from before a force push, even after pushing an ordinary commit in the meantime :/

haampie:
@scottwittenburg / @kwryankrattiger can you review? Notice that a more recent GNU Make is not a hard requirement, it's just that it improves legibility as parallel output appears nice and orderly. The stop-gap solution of using Julia's pre-built binaries can be improved by moving it into the relevant docker images, but I think it's OK to have this solution for now.

https://gitlab.spack.io/spack/spack/-/pipelines/195792 <- this was the last pipeline before I removed the commit 395d95f1d134b69b7035c3859be633e97daab655 that was meant to trigger Gitlab CI.

zackgalbreath:
> @scottwittenburg / @zackgalbreath it's apparently impossible to get a pipeline to run on the latest commit if it's in "Pending — waiting for base develop commit pipeline to succeed" mode? It continues to pick a commit from before a force push, even after pushing an ordinary commit in the meantime :/

FYI we just updated our sync script to remove this bit of functionality. Going forward, it should no longer block PRs with that "waiting for base develop commit pipeline to succeed" status.

lib/spack/spack/cmd/ci.py
```diff
@@ -515,42 +515,98 @@ def ci_rebuild(args):
...
+    deps_install_args = install_args
+    root_install_args = install_args + [
+        "--keep-stage",
+        "--only=package",
+        "--use-buildcache=package:never,dependencies:only",
+        slash_hash,
+    ]
+
+    commands = [
+        # apparently there's a race when spack bootstraps? do it up front once
```
scottwittenburg, line 545: Why did I think the runner images had everything and didn't need any bootstrapping? I suppose spack moves so fast that's not a practical goal?

scottwittenburg:
This seems like it could be a great improvement, at least until more improvements to the underlying problems can land in spack (e.g. parallel/faster fetching of binaries, faster binary relocation, etc).

One thing that occurred to me is that we make these specific cpu requests to assist kubernetes in scheduling rebuild jobs on pods, we try to put them in every mapping entry, e.g. [here](https://github.com/spack/spack/blob/develop/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml#L434), and make them match what we've seen them to require historically.     

I see this PR is basing make parallelism on `$nproc`, but these requests aren't limits, so I wonder if we need to worry about oversubscribing the instances we get run on?  

@zackgalbreath

lib/spack/spack/cmd/ci.py
haampie: It's part of https://github.com/spack/spack/pull/32262, some distros provide patchelf versions that are a decade old, and manipulating binaries is very bug-prone, so it makes sense to stick to a recent patchelf.

haampie:
As far as I know nproc is aware of cgroups, for example:

```
taskset -c 0-1 nproc # nproc is restricted to cores 0 and 1
2
taskset -c 0-1 nproc --all
16
```
but idk how things are with kubernetes, since they have these fractional cpus etc

scottwittenburg:
> but idk how things are with kubernetes, since they have these fractional cpus etc

IIUC the idea with Kubernetes is to tell it what kind of resources each job needs, and let it figure out how to pack jobs onto instances.  We can also give it resource limits, and if we do that for cpu, I think it throttles the processes to make sure they don't go over the limits we specify.  I'm not sure we want that though, maybe with cpu we just don't need to worry as much as with memory (where the risk is job death by OOM).

Opinions @zackgalbreath?

haampie:
If you haven't configured `config:build_jobs` the same oversubscription issue would be true for `spack install` itself, since Spack determines the number of jobs with `os.sched_getaffinity(0)` which is exactly what `nproc` does.

```
taskset -c 0-1 spack python -c 'spack.util.cpus.cpus_available()'
2
```

That said, Spack's relocation logic may not be using composable threading (I believe it fires up a potentially large thread pool?). I guess we'll find out if this is an issue by trying... -- would be nice if spack was jobserver aware, i made an issue about that: #33014

trws:
It's also aware of weak binding without cgroups, which is what taskset uses.  As long as only one spack install is allowed to actually *build* something at a time, this wont change the risks.  If there's a top-level `make -j 5` that runs multiple concurrent installs that run builds, rather than installing from the binary cache, then it's possible because of misbehaving build systems and linkers ignoring the jobserver, but I don't think we have that case here.

haampie:
@spackbot run pipeline

haampie:
@zackgalbreath / @scottwittenburg / @kwryankrattiger can you add a review?

haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

haampie:
Now seeing https://gitlab.spack.io/spack/spack/-/jobs/3862768 which is probably #31387, added #33478 and #33477 to improve at least the error... (I don't think it's a blocker since gitlab will retry and it happens at the start)

kwryankrattiger (dismissed):
Two gripes that don't need to be fixed, mostly just notes for me later when we get to the CI module refactor
* Downloading the make isn't my favorite, but I don't have a better solution, so end of gripe 😞 
* Windows CI, whenever ever it happens, will likely not work well with this. So it will eventually need to be an option instead of always on. Other options would be adding Ninja (@haampie and I talked about this briefly) or adding finer-grained parallel install support in spack itself.

Thanks for your work on this @haampie

haampie:
For windows we'd need to do some refactoring anyways, since we're now generating and executing a shell script. There's already room for `spack env depfile --generator ninja`, (currently it only supports `make` as a value). Downloading make is not great, but it can be moved in the base images.

haampie:
@spackbot run pipeline

haampie:
how do I make gitlab ci pass on this pr... Merge develop into here?

kwryankrattiger:
@spackbot run pipeline

haampie:
merging develop s.t. hopefully the openblas hash changes by in something more favorable...

haampie:
@zackgalbreath what can I do to make this PR pass? Apparently merging develop doesn't change the openblas hashes. Why is it failing in this PR?

haampie:
ping @scottwittenburg / @zackgalbreath / @kwryankrattiger, no clue how to proceed here.

scottwittenburg:
I'm not sure what you can do while there is a broken `openblas` on `develop`.  You might get @tgamblin to merge this in spite of the failure if you think it's enough of an emergency to get this in.

haampie:
It's not an emergency, but this is a ridiculous situation to be in... I don't need to fix openblas here, it's unrelated to my PR. Now my PR is blocked by it :|.

scottwittenburg:
> but this is a ridiculous situation to be in

I agree it's unfortunate, sorry.  But what policy do you disagree with specifically?  That we fail PRs that generate broken hashes?  Or just that it affects this particular PR?  Keep in mind that one policy is not to prune "untouched" specs from a stack if the PR touched the `spack.yaml` for that stack, and this PR touches every `spack.yaml`.  So maybe you disagree with how we don't prune "untouched" specs when the affected stack's `spack.yaml` was touched?  

@tgamblin in case you're interested in discussing whether any of these policies should be revisited.

haampie:
@zackgalbreath unmarked openblas cause it's a buildmachine related issue (apparently their makefile still checks the host arch even wtih DYNAMIC_ARCH=1, and the cpu type of the build node was something weird?)

haampie:
@spackbot run pipeline

alalazo (approved):
We should eventually bootstrap `make` and `tar` too.

PR #33251, opened 2022-10-12
pilercr: new package

luke-dt:


tldahlgren (commented):
Confirmed sha256.

var/spack/repos/builtin/packages/pilercr/package.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+from spack.package import *
+
+
+class Pilercr(Package):
```
adamjstewart:
```suggestion
class Pilercr(MakefilePackage):
```
The `edit` and `build` phases aren't even being run at the moment.

adamjstewart:
Not sure why @pradyunsg and I were asked to review this, it isn't a Python library.

tldahlgren:
> Not sure why @pradyunsg and I were asked to review this, it isn't a Python library.

Apologies.  That's what I get for reviewing aspects of multiple PRs at the same time.  Must've had the wrong PR tab in focus.

var/spack/repos/builtin/packages/pilercr/package.py
luke-dt: Apologies, I was testing on a couple different machines and think I just got mixed up with what fixed what. I was encountering a bug where some static libraries were not included on the version of Rocky Linux that is on our cluster. Editing the Makefile to use shared libraries fixed the error, but just realized that I didn't test this in spack on that machine. I instead tested it on a Fedora machine which does have the static libs installed.

I was just able to reproduce this error on Rocky Linux and will push a new commit when it's fixed.

luke-dt: @adamjstewart so when I build it as a `class Pilercr(Package)` it builds fine but `class Pilercr(MakefilePackage)` encounters the same error that I got earlier (regardless of whether that edit function is there):

```
4 errors found in build log:
     112    g++ -c -O3 -DNDEBUG=1  utils_win32.cpp -o utils_win32.o
     113    g++ -c -O3 -DNDEBUG=1  validaa.cpp -o validaa.o
     114    g++ -c -O3 -DNDEBUG=1  writearrays.cpp -o writearrays.o
     115    g++ -c -O3 -DNDEBUG=1  writegff.cpp -o writegff.o
     116    g++ -c -O3 -DNDEBUG=1  writepiles.cpp -o writepiles.o
     117    g++ -O3 -DNDEBUG=1 -o pilercr acceptaa.o accepthit.o alignarrays.o alignclust.o alignprofs.o aligntraps.o alnarray.o alngivenpath.o alpha.o assert.o bittb.o clustcons.o clustconsr.o cmp.o comp.o c
            ons.o distcalc.o distfunc.o dp.o estring.o expandpiles.o filterb.o filterb2.o findarrays.o findpiles.o fixbounds.o getarraydata.o getccs.o getconsaa.o getgdtree.o getparams.o getsubseq.o glbalign.
            o hitstoimages.o info.o kmerdist.o log.o logaa.o loghits.o main.o makecontigmap.o makegraph.o makeindex.o mem.o mergearrays.o mergefilterhits.o mkrootmsa.o msa.o multaln.o options.o outarray.o out
            detail.o outdetailaa.o outlocalaln.o params.o pilercr.o proffrommsa.o progaln.o progress.o pwpath.o quit.o readfilterhits.o readmfa.o revcomp.o savefhits.o savefilterhit.o saveseqs.o savetraps.o s
            coremx.o seq.o seqvect.o swsimple.o tbsw.o tree.o upgma.o usage.o utils.o utils_linux.o utils_other.o utils_unix.o utils_win32.o validaa.o writearrays.o writegff.o writepiles.o -lm -static
  >> 118    /usr/bin/ld: cannot find -lm
  >> 119    /usr/bin/ld: cannot find -lc
  >> 120    collect2: error: ld returned 1 exit status
  >> 121    make: *** [Makefile:24: pilercr] Error 1
```

Do you have any insights as to what's going on here?

adamjstewart: Nope. It builds fine with `Package` because it isn't actually building anything, it's just copying the pre-compiled binary to the installation prefix, but that won't work on many OSes/arches. Not sure why it can't find libm or libc, those usually come with the compiler. I would suggest overriding `GPP` and/or `CPP` so that Spack can control the compiler too.

luke-dt: I think the issue is that the `Makefile` is specifying the static verisons, which aren't included by default in all OSes (including Rocky Linux and Fedora 36 IME):

```
CFLAGS = -O3 -DNDEBUG=1
LDLIBS = -lm -static
# LDLIBS = -lm

OBJ = .o
EXE =

RM = rm -f
CP = cp

GPP = g++
LD = $(GPP) $(CFLAGS)
CPP = $(GPP) -c $(CFLAGS) 

all: pilercr

CPPSRC = $(sort $(wildcard *.cpp))
CPPOBJ  = $(subst .cpp,.o,$(CPPSRC))

$(CPPOBJ): %.o: %.cpp
        $(CPP) $< -o $@

pilercr: $(CPPOBJ)
        $(LD) -o pilercr $(CPPOBJ) $(LDLIBS)
        strip pilercr
```

Swapping out the comments lines 2/3 fixes the bug for me. How would I override `GPP` and `CPP` in the package?

adamjstewart: See https://spack.readthedocs.io/en/latest/build_systems/makefilepackage.html for examples of different ways you can override variables.

adamjstewart (approved):
LGTM but I don't know anything about this software

luke-dt:
> LGTM but I don't know anything about this software

Haha, I'm just impatient because this is a dep of something else I want to submit a PR for. You were helping me with overriding the Makefile variables earlier.

It actually builds now, and I have it working on my machine with some test data.

adamjstewart:
I'll let @tldahlgren make the final merge decision.

tldahlgren (approved):
LGTM.  Re-confirmed the sha256 and was able to successfully build the package.

PR #33252, opened 2022-10-12
py-alive-progress, py-about-time, py-graphme: new packages

luke-dt:


tldahlgren (dismissed):
LGTM.  Confirmed all 5 sha256.

var/spack/repos/builtin/packages/py-about-time/package.py
```diff
@@ -0,0 +1,21 @@
...
+    """A cool helper for tracking time and throughput of
+    code blocks, with beautiful human friendly renditions."""
+
+    homepage = "https://github.com/rsalmei/about-time"
+    pypi = "about-time/about-time-4.1.0.tar.gz"
+
+    version("4.1.0", sha256="963b1f3739b0c9732eb205031762b76f1291d89b5d0c8220a8d5b154e32ce650")
+    version("3.1.1", sha256="586b329450c9387d1ae8c42d2db4f5b4c57a54508d0f1b7bb00322ffd5ce9f9b")
+
+    depends_on("python@3.7:", type=("build", "run"), when="@4:")
```
adamjstewart:
```suggestion
    depends_on("python@3.7:3", type=("build", "run"), when="@4:")
```

var/spack/repos/builtin/packages/py-alive-progress/package.py
```diff
@@ -0,0 +1,25 @@
...
+
+    homepage = "https://github.com/rsalmei/alive-progress"
+    pypi = "alive-progress/alive-progress-2.4.1.tar.gz"
+
+    version("2.4.1", sha256="089757c8197f27ad972ba27e1060f6db92368d83c736884e159034fd74865323")
+    version("1.6.2", sha256="642e1ce98becf226c8c36bf24e10221085998c5465a357a66fb83b7dc618b43e")
+
+    depends_on("python@2.7:3.8", type=("build", "run"))
+    depends_on("python@3.6:", type=("build", "run"), when="@2:")
+    depends_on("python@3.7:", type=("build", "run"), when="@2.2:")
```
adamjstewart:
```suggestion
    depends_on("python@3.6:3", type=("build", "run"), when="@2:")
    depends_on("python@3.7:3", type=("build", "run"), when="@2.2:")
```

var/spack/repos/builtin/packages/py-grapheme/package.py
```diff
@@ -0,0 +1,22 @@
...
+    specifically, string manipulation and calculation functions for working
+    with grapheme cluster groups (graphemes) as defined by the
+    Unicode Standard Annex #29."""
+
+    homepage = "https://github.com/alvinlindstam/grapheme"
+    pypi = "grapheme/grapheme-0.6.0.tar.gz"
+
+    version("0.6.0", sha256="44c2b9f21bbe77cfb05835fec230bd435954275267fea1858013b102f8603cca")
+
+    depends_on("python@3.6:", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("python@3.6:", type=("build", "run"))
```
Where do you see 3.6+?

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-grapheme/package.py
luke-dt: oh, that's a typo, I think I meant 3.5. The setup.py lists 3.5, 3.6, 3.7, and 3.8

adamjstewart: I generally don't specify unless the packager uses `python_requires` since pip doesn't read that metadata, but up to you if you want to do 3.5+.

luke-dt: I'm fine with whatever is more correct for spack. The larger package that this is a dependency for requires 3.8 anyway so it doesn't affect me personally

adamjstewart: I would just leave it out then, Spack won't even support Python 3.6 and older after the next release anyway.

adamjstewart (approved):

PR #33254, opened 2022-10-12
Add some options to Trilinos

jrood-nrel:
These are some things we've been doing locally for Trilinos that seem to make sense to upstream. I'm open to discussion.

- Add ninja variant to Trilinos
- Add `stk_unit_tests` variant to Trilinos
- Add `stk_simd` variant to Trilinos
- Add better debug options for CUDA in Trilinos
- Add build target optimization to ROCm in Trilinos

### DIFF
--- a/var/spack/repos/builtin/packages/trilinos/package.py
+++ b/var/spack/repos/builtin/packages/trilinos/package.py
@@ -91,7 +91,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
         multi=False,
         description="global ordinal type for Tpetra",
     )
-    variant('ninja', default=False, description='Enable Ninja makefile generator')
+    variant('ninja', default=False, description="Enable Ninja makefile generator")
     variant("openmp", default=False, description="Enable OpenMP")
     variant("python", default=False, description="Build PyTrilinos wrappers")
     variant("shared", default=True, description="Enables the build of shared libraries")
@@ -140,8 +140,8 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     variant("rythmos", default=False, description="Compile with Rythmos")
     variant("sacado", default=True, description="Compile with Sacado")
     variant("stk", default=False, description="Compile with STK")
-    variant('stk_unit_tests', default=False, description='Enable STK unit tests')
-    variant('stk_simd', default=False, description='Enable SIMD in STK')
+    variant('stk_unit_tests', default=False, description="Enable STK unit tests")
+    variant('stk_simd', default=False, description="Enable SIMD in STK")
     variant("shards", default=False, description="Compile with Shards")
     variant("shylu", default=False, description="Compile with ShyLU")
     variant("stokhos", default=False, description="Compile with Stokhos")
@@ -385,7 +385,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("metis", when="+zoltan")
     depends_on("mpi", when="+mpi")
     depends_on("netcdf-c", when="+exodus")
-    depends_on("ninja", type="build", when='+ninja')
+    depends_on("ninja", type="build", when="+ninja")
     depends_on("parallel-netcdf", when="+exodus+mpi")
     depends_on("parmetis", when="+mpi +zoltan")
     depends_on("parmetis", when="+scorec")
@@ -455,10 +455,10 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
 
     @property
     def generator(self):
-          if '+ninja' in self.spec:
-              return "Ninja"
-          else:
-              return "Unix Makefiles"
+        if "+ninja" in self.spec:
+            return "Ninja"
+        else:
+            return "Unix Makefiles"
 
     def flag_handler(self, name, flags):
         spec = self.spec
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -514,6 +525,12 @@ def setup_dependent_package(self, module, dependent_spec):
     def setup_build_environment(self, env):
         spec = self.spec
         if "+cuda" in spec and "+wrapper" in spec:
+            # Add more debug options for CUDA
+            if (
+                spec.variants["build_type"].value == "RelWithDebInfo"
+                or spec.variants["build_type"].value == "Debug"
+            ):
```
tldahlgren: FWIW. Could simplify:

```suggestion
            if spec.variants["build_type"].value in ["RelWithDebInfo", "Debug"]:
```

tldahlgren (commented):
Just suggesting a simplification.  Deferring feedback to maintainers.

var/spack/repos/builtin/packages/trilinos/package.py
jrood-nrel: Excellent.

var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -91,6 +91,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
         multi=False,
         description="global ordinal type for Tpetra",
     )
+    variant("ninja", default=False, description="Enable Ninja makefile generator")
```
sethrj: Similiar variants have been proposed and rejected for several projects. Since it changes only the build system and shouldn't change the installed binaries, it shouldn't be a variant. Since ninja is better than make in most cases it might be good to have a global spack option to default the cmake generator to something different: I'll discuss this with the maintainers.

var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -139,6 +140,8 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     variant("rythmos", default=False, description="Compile with Rythmos")
     variant("sacado", default=True, description="Compile with Sacado")
     variant("stk", default=False, description="Compile with STK")
+    variant("stk_unit_tests", default=False, description="Enable STK unit tests")
```
sethrj: Unit tests don't change the installed binaries so shouldn't be a variant. You should instead add `if self.run_tests and "+stk" in spec`

var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -514,6 +525,9 @@ def setup_dependent_package(self, module, dependent_spec):
     def setup_build_environment(self, env):
         spec = self.spec
         if "+cuda" in spec and "+wrapper" in spec:
+            # Add more debug options for CUDA
+            if spec.variants["build_type"].value in ["RelWithDebInfo", "Debug"]:
+                env.append_flags("CXXFLAGS", "-Xcompiler -rdynamic -lineinfo")
```
sethrj: Under what circumstances is this needed? Does it have particular CUDA compatibility constraints? Shouldn't it be CUDA flags, not CXX flags? And did you know that you can set custom CXX flags through spack? I would shy away from flags that aren't required to get the build to pass.

var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -910,10 +929,16 @@ def define_tpl(trilinos_name, spack_name, have_dep):
                         define_kok_enable("HIP_RELOCATABLE_DEVICE_CODE", "rocm_rdc"),
                     ]
                 )
+                # The following is an optimization to compile only for the specified
+                # archs since all archs are done by default
+                amd_targets = self.spec.variants["amdgpu_target"].value
+                options.append("-DCMAKE_HIP_ARCHITECTURES=" + ";".join(str(x) for x in targets))
```
sethrj, line 919: `self.define` automatically formats lists.
```suggestion
                options.append(self.define("CMAKE_HIP_ARCHITECTURES", targets))
```

Also how are this and the AMDGPU/GPU defines (and the kokkos defines) related?

sethrj (changes requested):

var/spack/repos/builtin/packages/trilinos/package.py
jrood-nrel: Makes sense. I'll just remove this. Was curious how someone else felt about it here.

var/spack/repos/builtin/packages/trilinos/package.py
jrood-nrel: I think I will just remove this for now.

var/spack/repos/builtin/packages/trilinos/package.py
jrood-nrel: This basically just adds stuff that acts like `-g` but for CUDA when doing builds with debug symbols. I'm not familiar enough with how I would do this with CUDA flags since `nvcc_wrapper` handles this for us. I will just remove this for now. It's not strictly necessary, just convenient.

sethrj: Ah man, it's wrappers on top of wrappers on top of wrappers!

var/spack/repos/builtin/packages/trilinos/package.py
jrood-nrel: This is separate from Kokkos. These override stuff in `find_package(HIP)` and I've found it necessary to list them all. Talking with someone at AMD, it should just be one option in the future. However, this isn't strictly necessary, just builds faster.

PR #33275, opened 2022-10-12
packages.yaml: set url/git

scheibelp:
A user may want to set some attributes on a package without actually modifying the package (e.g. if they want to `git pull` updates to the package without conflicts). This PR adds a per-package configuration section called "set", which is a dictionary of attribute names to desired values. For example:

```
packages:
  openblas:
    set:
      submodules: true
```

in this case, the package will always retrieve git submodules. Right now the PR maintains a limited list of acceptable attributes that can be set.

### DIFF
--- a/lib/spack/docs/build_settings.rst
+++ b/lib/spack/docs/build_settings.rst
@@ -548,7 +548,8 @@ You can assign class-level attributes in the configuration:
       x: 1
 
 Attributes set this way will be accessible to any method executed
-in the package.py file (e.g. the ``install()`` method).
+in the package.py file (e.g. the ``install()`` method). Values for these
+attributes may be any value parseable by yaml.
 
 These can only be applied to specific packages, not "all" or
 virtual packages.
--- a/lib/spack/docs/build_settings.rst
+++ b/lib/spack/docs/build_settings.rst
@@ -531,3 +531,24 @@ directories inside the install prefix. This will ensure that even
 manually placed files within the install prefix are owned by the
 assigned group. If no group is assigned, Spack will allow the OS
 default behavior to go as expected.
+
+----------------------------
+Assigning Package Attributes
+----------------------------
+
+You can assign class-level attributes in the configuration:
+
+.. code-block:: yaml
+
+  packages:
+    mpileaks:
+      # Override existing attributes
+      url: http://www.somewhereelse.com/mpileaks-1.0.tar.gz
+      # ... or add new ones
+      x: 1
+
+Attributes set this way will be accessible to any method executed
+in the package.py file (e.g. the ``install()`` method).
+
+These can only be applied to specific packages, not "all" or
+virtual packages.
--- a/lib/spack/spack/repo.py
+++ b/lib/spack/spack/repo.py
@@ -1366,6 +1366,10 @@ class Repo(object):
         overidden_attrs = getattr(cls, 'overidden_attrs', {})
         for key, val in overidden_attrs.items():
             setattr(cls, key, val)
+        # Keep track of every class attribute that is overidden by the config:
+        # if the config changes between calls to this method, we make sure to
+        # restore the original config values (in case the new config no longer
+        # sets attributes that it used to)
         new_overidden_attrs = {}
         for key, val in new_cfg_settings.items():
             new_overidden_attrs[key] = getattr(cls, key)
--- a/lib/spack/spack/repo.py
+++ b/lib/spack/spack/repo.py
@@ -1378,6 +1378,8 @@ class Repo(object):
             setattr(cls, key, val)
         if new_overidden_attrs:
             setattr(cls, "overidden_attrs", dict(new_overidden_attrs))
+        else:
+            delattr(cls, "overidden_attrs")
 
         return cls
 
--- a/lib/spack/spack/repo.py
+++ b/lib/spack/spack/repo.py
@@ -1366,16 +1366,26 @@ class Repo(object):
             spack.config.get("packages").get(pkg_name, {}).get("package_attributes", {})
         )
         overidden_attrs = getattr(cls, "overidden_attrs", {})
+        attrs_exclusively_from_config = getattr(cls, "attrs_exclusively_from_config", [])
+        # Clear any prior changes to class attributes in case the config has
+        # since changed
         for key, val in overidden_attrs.items():
             setattr(cls, key, val)
+        for key in attrs_exclusively_from_config:
+            delattr(cls, key)
+
         # Keep track of every class attribute that is overidden by the config:
         # if the config changes between calls to this method, we make sure to
         # restore the original config values (in case the new config no longer
         # sets attributes that it used to)
         new_overidden_attrs = {}
+        new_attrs_exclusively_from_config = set()
         for key, val in new_cfg_settings.items():
             if hasattr(cls, key):
                 new_overidden_attrs[key] = getattr(cls, key)
+            else:
+                new_attrs_exclusively_from_config.add(key)
+
             if isinstance(val, list):
                 raise spack.config.ConfigError("Unsupported attribute value: list")
             elif isinstance(val, dict):
@@ -1388,6 +1398,10 @@ class Repo(object):
             setattr(cls, "overidden_attrs", dict(new_overidden_attrs))
         elif hasattr(cls, "overidden_attrs"):
             delattr(cls, "overidden_attrs")
+        if new_attrs_exclusively_from_config:
+            setattr(cls, "attrs_exclusively_from_config", new_attrs_exclusively_from_config)
+        elif hasattr(cls, "attrs_exclusively_from_config"):
+            delattr(cls, "attrs_exclusively_from_config")
 
         return cls
 
--- a/lib/spack/spack/repo.py
+++ b/lib/spack/spack/repo.py
@@ -1362,6 +1362,16 @@ class Repo(object):
         if not inspect.isclass(cls):
             tty.die("%s.%s is not a class" % (pkg_name, class_name))
 
+        new_cfg_settings = spack.config.get("packages").get(pkg_name, {}).get("package_attributes", {})
+        overidden_attrs = getattr(cls, 'overidden_attrs', {})
+        for key, val in overidden_attrs.items():
+            setattr(cls, key, val)
+        new_overidden_attrs = {}
+        for key, val in new_cfg_settings.items():
+            new_overidden_attrs[key] = getattr(cls, key)
+            setattr(cls, key, val)
+        setattr(cls, 'overidden_attrs', dict(new_overidden_attrs))
+
         return cls
 
     def __str__(self):
--- a/lib/spack/spack/repo.py
+++ b/lib/spack/spack/repo.py
@@ -1378,7 +1378,7 @@ class Repo(object):
             setattr(cls, key, val)
         if new_overidden_attrs:
             setattr(cls, "overidden_attrs", dict(new_overidden_attrs))
-        else:
+        elif hasattr(cls, "overidden_attrs"):
             delattr(cls, "overidden_attrs")
 
         return cls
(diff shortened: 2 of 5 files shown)


lib/spack/spack/schema/packages.py
```diff
@@ -82,6 +82,23 @@
                             },
                         },
                     },
+                    # If 'get_full_repo' is promoted to a Package-level
+                    # attribute, it could be useful to set it here
+                    "set": {
```
becker33: I would use `pkg_attributes` instead of `set` here -- I think it makes the config more clearly legible.

lib/spack/spack/schema/packages.py
```diff
@@ -82,6 +82,23 @@
...
+                        "properties": {
+                            "git": {
+                                "type": "string",
+                            },
+                            "url": {
+                                "type": "string",
+                            },
+                            "submodules": {
+                                "type": "boolean",
+                            },
```
becker33: I think there are other attributes that users will want to set here.

becker33 (changes requested):
Implementation LGTM

scheibelp:
Weirdly, removing the new test I added allows the entire suite of unit tests to pass:

* For each of the failures I was seeing, the test I added never even ran before hitting a failure (the failing test was `mirror.test_all_mirror`
* When I add my test locally, this test does not fail locally

lib/spack/spack/schema/packages.py
scheibelp: @becker33 I've updated this so you can enter arbitrary properties as strings and they are parsed as yaml values, which means that

```
x: "true"
```

is automatically applied such that `Package.x == True`. One side effect is that

```
x: "y"
```

is also converted to `Package.x == True` (since "y" is a truth value in yaml). We could allow users to fall back to something like:

```
x:
  value: "y"
  type: "string"
```

but that would remove the possibility to assign dictionary attributes using yaml structure like:

```
x:
  a: 1
  b: 2
```

although with the current logic you could do

```
x: "{a: 1, b: 2}"
```

which would automatically be loaded as a dictionary.

I'm curious if you have a preference.

biddisco:
Here's a use case that it would be nice to address with this PR.

Project A is under development and maintained locally, it uses Project B that is a generic spack dependency (we will use `openblas` as an example). `openblas` is maintained externally, but we have made some tweaks to it locally in a git clone of the public repo. These might be adjusted default buffer sizes or something  and you have N branches, with N different tweaks that you want to experiment with. On your local development machine, you manually switch branches and test various combinations of A and B to find good settings.

`rsync` is used to mirror repo A and repo B to a cluster where some heavy benchmarking is to take place, you have a script that builds project A using spack, then runs a set of tests and collect results, but you want your script to build proj A, N or more times with different options and on each build, use branch 1..N of your copy of `openblas`. You might invoke it with
```benchmark branchA branchB branchC branchD``` to collect 4 runs of data that you will compare timings of.

Because `openblas` is a package that spack knows about already, and has its git URL set to the `openblas` public repo (that you do not have write permission for and wouldn't want to clutter up with lots of useless branches anyway), when you build A, you want to change the git branch to ```branchA, branchB, branchC, branchD``` as you iterate over your benchmark list, which can be done already by changing the ```depends_on("openblas@branchA")``` entry in the package entry for your proj A (and the script can do this for you if necessary), but the script can't change the `openblas` git URL since that is hardcoded inside spack due to it being a package setup by spack that you have no real control over - and those branches simply don't exist in the public repo.
Being able to change the git URL of a dependency as well as the branch name (which seems so trivial it ought to be a very simple to implement), would make this workflow possible, without having to jump through hoops editing config files and environments for each build.
(passing the `openblas` git URL as a command line param when invoking spack would be fine - even via a config file generated on the fly - which looks possible with this PR, but I'm unsure of exactly what to set).

**Comment by scheibelp** [comment]
This PR would support your case if you did for example:

```
spack -C config1 install a ^openblas@gitbranch1
spack -C config2 install a ^openblas@gitbranch2
...
```

We have `spack config add` which could modify the configuration from the command line, but you cannot add values that contain colons (e.g. `spack -d config add packages:openblas:package_attributes:git:"https://somewhere.com/git/"` would fail). In the meantime, the script can start with a manually-generated config (or use `echo` if the git URL changes over time, but I got the impression that wasn't needed in your case).

> Being able to change the git URL of a dependency as well as the branch name (which seems so trivial it ought to be a very simple to implement), would make this workflow possible, without having to jump through hoops editing config files and environments for each build.

As mentioned above in this comment, based on your described use case you do not require a per-build config, but rather one config redirecting `openblas` (in your example) to use a different git URL (but consistent amongst your different builds). It may be possible that we could further compress from 3 commands down to 2 (or one) per build, but I'm not sure if we want to add a competing mechanism for doing this (more maintenance). To be clear, if you anticipate more work than 2-3 commands per build, I'm interested in how we could streamline that.

File: lib/spack/spack/repo.py [hunk 3]
```diff
@@ -1362,6 +1362,33 @@ def get_pkg_class(self, pkg_name):
         if not inspect.isclass(cls):
             tty.die("%s.%s is not a class" % (pkg_name, class_name))
 
+        new_cfg_settings = (
+            spack.config.get("packages").get(pkg_name, {}).get("package_attributes", {})
+        )
```
becker33, line 1367: Is there any attribute that could be reasonably set from `all` that we should account for here?

lib/spack/spack/repo.py
```diff
@@ -1362,6 +1362,33 @@ def get_pkg_class(self, pkg_name):
...
+        overidden_attrs = getattr(cls, "overidden_attrs", {})
```
becker33: Can you add a comment before this line making clear that this is to clear any previous changes to ensure they don't go stale? Also I think for performance it would be good if we could avoid doing this if the config hasn't changed.

lib/spack/spack/repo.py
```diff
@@ -1362,6 +1362,33 @@ def get_pkg_class(self, pkg_name):
...
+        for key, val in overidden_attrs.items():
+            setattr(cls, key, val)
+        # Keep track of every class attribute that is overidden by the config:
+        # if the config changes between calls to this method, we make sure to
+        # restore the original config values (in case the new config no longer
+        # sets attributes that it used to)
+        new_overidden_attrs = {}
+        for key, val in new_cfg_settings.items():
+            if hasattr(cls, key):
```
becker33, line 1385: If we want to avoid stale config, we also need to track attributes we've added that were not present before. Then we can know to delete them.

lib/spack/spack/repo.py
```diff
@@ -1362,6 +1362,33 @@ def get_pkg_class(self, pkg_name):
...
+                new_overidden_attrs[key] = getattr(cls, key)
+            if isinstance(val, list):
+                raise spack.config.ConfigError("Unsupported attribute value: list")
+            elif isinstance(val, dict):
+                raise spack.config.ConfigError("Unsupported attribute value: dict")
+            if key not in ["git", "url", "submodules"]:
```
becker33: Why special handling for these?

lib/spack/spack/repo.py
```diff
@@ -1362,6 +1362,33 @@ def get_pkg_class(self, pkg_name):
...
+                # Infer type
+                val = yaml.load(val)
```
becker33: Isn't the value already read in from yaml? This seems redundant.

lib/spack/spack/schema/packages.py
```diff
@@ -82,6 +82,35 @@
...
```
becker33: Is there a reason these need to be special-cased? They would seem to fit in the schema below

lib/spack/spack/test/concretize_preferences.py
```diff
@@ -172,6 +172,41 @@ def test_preferred_providers(self):
         spec = concretize("mpileaks")
         assert "zmpi" in spec
 
+    def test_config_set_pkg_property_url(self, mutable_mock_repo):
+        """Test preferred providers of virtual packages are
+        applied correctly
+        """
```
becker33: This appears to be out of date.

becker33 (changes requested):

lib/spack/spack/repo.py
scheibelp: This is so that users can write `x: true` instead of 

```
x:
  type: boolean
  value: true
```

lib/spack/spack/repo.py
scheibelp: This relates to YAML auto-type conversion in https://github.com/spack/spack/pull/33275/files#r1012368704: since the schema anticipates these, there is no reason for the type to be inferred. Admittedly the only cases I can think of that are problematic are things like:

```
pkg_attributes:
  # I have a URL called "true", I don't want it to be inferred as a boolean
  git: "true"
```

but my personal preference is never to leave to inference what is explicitly known.

lib/spack/spack/repo.py
scheibelp: At the moment no. I can look into how to prohibit this in the schema.

lib/spack/spack/repo.py
scheibelp: > for performance it would be good if we could avoid doing this if the config hasn't changed.

How do we know whether the config has changed (I don't think we have a means to track that)?

lib/spack/spack/schema/packages.py
scheibelp: If you don't agree with https://github.com/spack/spack/pull/33275/files#r1012373337, then they do not need to be special-cased.

lib/spack/spack/repo.py
scheibelp: To be clear, this ensures that `x: true` is converted such that `Package.x is True` (otherwise it would be `Package.x == "true"`)

lib/spack/spack/repo.py
scheibelp: This is resolved now

lib/spack/spack/repo.py
scheibelp: Actually it occurred to me this could be useful for:

* `submodules = True` - a user might want to set this universally
* If we promote `full_git_repo` from a per-version option to a package-level option (like `submodules`)

It seems more-constrained though: like we ought to perhaps white-list entries for `all`.

becker33: I think for now it's fine if we just document that you can't set it for `all` nor for a virtual

lib/spack/spack/repo.py
becker33: We could store a hash of the previous config, but that might be more bother than it's worth. We should test the performance with this PR and see if we need to do something about it

lib/spack/spack/repo.py
becker33: I don't see why we need double-yaml-parsing. If we just rely on yaml parsing the first time instead of forcing everything to be a string, then we can remove this entire block.

becker33: If you had a url "true" (weird and unlikely), you can always quote the url to turn it into a string in yaml.

lib/spack/docs/build_settings.rst
```diff
@@ -531,3 +531,24 @@ directories inside the install prefix. This will ensure that even
...
+
+  packages:
+    mpileaks:
+      # Override existing attributes
+      url: http://www.somewhereelse.com/mpileaks-1.0.tar.gz
+      # ... or add new ones
+      x: 1
+
+Attributes set this way will be accessible to any method executed
+in the package.py file (e.g. the ``install()`` method).
```
becker33:
```suggestion
in the package.py file (e.g. the ``install()`` method). Values for these
attributes may be any value parseable by yaml.
```

becker33 (dismissed):
LGTM, one request for the docs

becker33 (approved):

PR #33278, opened 2022-10-13
py-jupyterlab-pygments: install from wheel to avoid cyclic dependency

iarspider:
See also: https://github.com/jupyterlab/jupyterlab_pygments/issues/23 . Alternatively, we can install `py-jupyterlab-pygments` from wheel, since `py-jupyterlab` dependency is build-time only.


adamjstewart:
Good catch, and thanks for the issue link!

Can you add the issue link as a comment in the package? That will help us track whether or not the cycle is ever removed.

Does this build correctly? I'm a bit concerned about what will happen if jupyterlab isn't installed. It may be better to install from a wheel like you suggested. Then we can remove all build-only deps.

I'm told that there is work in progress to change Spack's concretizer such that the same dependency can appear in the DAG multiple times. This is important for cross-compilation, but may also help here. Since we're dealing with build-only dependencies, it should be possible to turn this cycle into a loop-the-loop, where different versions of build deps are used than run deps.

adamjstewart:
Can you keep the comment and link about cyclic deps so we know why we're using a wheel?

adamjstewart (dismissed):

iarspider:
@spackbot fix style

var/spack/repos/builtin/packages/py-jupyterlab-pygments/package.py
```diff
@@ -11,14 +11,15 @@ class PyJupyterlabPygments(PythonPackage):
     """Pygments theme using JupyterLab CSS variables."""
 
     homepage = "https://jupyter.org/"
-    pypi = "jupyterlab-pygments/jupyterlab_pygments-0.1.1.tar.gz"
+    url = "https://files.pythonhosted.org/packages/py2.py3/j/jupyterlab-pygments/jupyterlab_pygments-0.2.2-py2.py3-none-any.whl"
+    # We use wheels because in @0.2.2: there is a cyclic dependency between py-nbconvert and py-jupyter-server
```
adamjstewart: I think this line is too long

var/spack/repos/builtin/packages/py-jupyterlab-pygments/package.py
```diff
@@ -11,14 +11,19 @@ class PyJupyterlabPygments(PythonPackage):
     """Pygments theme using JupyterLab CSS variables."""
 
     homepage = "https://jupyter.org/"
-    pypi = "jupyterlab-pygments/jupyterlab_pygments-0.1.1.tar.gz"
+    url = (
+        "https://files.pythonhosted.org/packages/py2.py3/j/jupyterlab-pygments/"
+        "jupyterlab_pygments-0.2.2-py2.py3-none-any.whl"
+    )
```
adamjstewart: The URL can be on a single line, it's ignored by flake8

adamjstewart (approved):

iarspider:
@spackbot run pipeline

PR #33280, opened 2022-10-13
Add missing upperbound to docs/spack.yaml

haampie:
Was added to requirement.txt but not to the environment file of this other amazing package manager called Spack.


tldahlgren (approved):
LGTM

PR #33281, opened 2022-10-13
Classic Intel compilers do not support gcc-toolchain

stephenmsachs:
This fix removes `--gcc-toolchain=` from the ~.fcg` files for the classic Intel compilers. AFAIK this option is only supported for Clang based compilers.

This lead to an issue when installing cmake. Reproducer:
```
spack install cmake@3.24.2%intel@2021.7.0~doc+ncurses+ownlibs~qt
build_type=Release arch=linux-amzn2-skylake_avx512
```

Tagging maintainer @rscohn2


rscohn2:
@haampie @skosukhin:

I see more problems now.

icc accepts gcc-name, icx accepts -gcc-toolchain
icc does not accept -Wno-unused-command-line-argument, but does not need it because it does not warn against Wl,-rpath and -c

I will refactor this a little more and test against all compilers.

haampie:
Oof, thanks for doing this!

rscohn2 (approved):
Thanks for the fix.

PR #33292, opened 2022-10-13
meson: remove slash in path

adamjstewart:
Not sure if meson even supports windows, but this should make things better.

### DIFF
--- a/lib/spack/spack/build_systems/meson.py
+++ b/lib/spack/spack/build_systems/meson.py
@@ -75,7 +75,7 @@ class MesonPackage(PackageBase):
     @property
     def archive_files(self):
         """Files to archive for packages based on Meson"""
-        return [os.path.join(self.build_directory, "meson-logs/meson-log.txt")]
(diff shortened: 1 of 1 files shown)


eli-schwartz (approved):
Meson definitely does support windows! Both MSVC (and clang-cl) and mingw (gcc / clang).

This PR definitely helps in a supported configuration of Meson, so LGTM.

lib/spack/spack/build_systems/meson.py
```diff
@@ -75,7 +75,7 @@ class MesonPackage(PackageBase):
     @property
     def archive_files(self):
         """Files to archive for packages based on Meson"""
-        return [os.path.join(self.build_directory, "meson-logs/meson-log.txt")]
+        return [os.path.join(self.build_directory, "meson-logs", "meson-log.txt")]
```
tldahlgren, line 78: FYI. Could also just use `join_path()` instead of `os.path.join()`.

tldahlgren (approved):
LGTM

PR #33293, opened 2022-10-13
meson: update OneAPI compiler support patch

adamjstewart:
Previous patch was a hack that only worked for C/C++. This patch is an actual feature that also works for Fortran.

@rscohn2 @eli-schwartz

### DIFF
--- a/var/spack/repos/builtin/packages/meson/oneapi.patch
+++ b/var/spack/repos/builtin/packages/meson/oneapi.patch
@@ -1,20 +1,5 @@
-From 4e87a15b7abec2d1bc3922f3f6cf2a8dca91b2c0 Mon Sep 17 00:00:00 2001
-From: Robert Cohn <robert.s.cohn@intel.com>
-Date: Sun, 9 Oct 2022 09:14:52 -0400
-Subject: [PATCH] basic support for oneapi linux compilers
-
----
- docs/markdown/Reference-tables.md          |  1 +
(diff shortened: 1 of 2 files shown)


eli-schwartz (dismissed):

tldahlgren (dismissed):
LGTM

rgommers:
This looks like a good idea to me. You may want to incorporate at least the first one-line diff in https://github.com/mesonbuild/meson/pull/10909#issuecomment-1277779755. That avoids configure-time errors on SciPy.

rscohn2:
> You may want to incorporate at least the first one-line diff in [mesonbuild/meson#10909 (comment)](https://github.com/mesonbuild/meson/pull/10909#issuecomment-1277779755).

I added the changes to the backport: https://github.com/rscohn2/meson/tree/oneapi-backport

adamjstewart:
Thanks! Based on the conversation in https://github.com/mesonbuild/meson/pull/10909, it seems like there are still issues with `-fvisibility=hidden` that will prevent this from working for scipy. But it also seems like this is a OneAPI problem, not a meson/scipy problem. Should we merge this as is (because it's correct and complete) or should we wait to hear back about `-fvisibility`? We could theoretically remove scipy from the OneAPI pipeline just to allow my PR to be merged.

rscohn2:
You can merge this. The visibility issue is not related to meson support.

adamjstewart:
@spackbot run pipeline

michaelkuhn (approved):
LGTM

Issue #32894, opened 2022-09-29
Environments: matrix of OS/arch


adamjstewart:
### Summary

I would like to be able to test the same stack in CI across multiple platforms.

### Rationale

I'm working on creating a CI stack for ML packages: https://github.com/spack/spack/pull/31592

I would like to be able to test this stack on multiple operating systems:

1. Linux
2. macOS
3. Windows?
4. Cray?

and multiple architectures:

1. x86_64
2. aarch64
3. ppc64le

at the moment, this is impossible without copy-n-pasting the entire stack into a separate file and changing a single line. Code duplication is bad.

### Description

Something like the following would be a good syntax:
```yaml
spack:
  packages:
    platforms:
      matrix:
        - linux
        - macos
        - windows
    targets:
      matrix:
        - x86_64
        - aarch64
        - powerpc
```

### Additional information

@becker33 @alalazo @scheibelp 

### General information

- [X] I have run `spack --version` and reported the version of Spack
- [X] I have searched the issues of this repo and believe this is not a duplicate

bladwig1:
I am also trying to do the same thing. I would love this feature too.

adamjstewart:
This feature is still greatly needed to simplify our builds and avoid duplication. Let's keep it open until this or a suitable alternative is supported.

Issue #32893, opened 2022-09-29
Environments: matrix of optional variants


adamjstewart:
### Summary

I would like to be able to create a Spack Environment with all possible values of a set of variants. Not all packages in the environment will have these variants, and some will have these variants recursively. Basically, I want something like:
```yaml
spack:
  packages:
    all:
      matrix:
        - requires: +A~B
        - requires: ~A+B
        - requires: +A+B
        - requires: ~A~B
  specs:
    - C
    - D
    - E
```
Where `+A` and `+B` will be toggled one and off for `C`, `D`, and `E`, even if those packages don't have those variants, even if their dependencies also have those same variants.

### Rationale

I'm working on creating a CI stack for ML packages: https://github.com/spack/spack/pull/31592

I would like to be able to test at least 3 variations:

1. `~rocm+cuda cuda_arch=...`
2. `~cuda+rocm amdgpu_target=...`
3. `~cuda~rocm`

At the moment, it seems like the only way to do this is to copy-n-paste the entire stack 3 times and change a single line. With this approach, 90% of the packages will be built twice with an identical spec since it isn't possible to reuse installations across stacks.

### Description

I would like to be able to do something like:
```yaml
spack:
  packages:
    all:
      matrix:
        - ~rocm+cuda cuda_arch=...
        - ~cuda+rocm amdgpu_target=...
        - ~cuda~rocm
```
Alternatively, something like:
```yaml
spack:
  definitions:
    - packages:
      - py-torch
      - py-tensorflow
    - variants:
      - +cuda
      - +rocm
      - ~cuda~rocm
  specs:
    - matrix:
      - [$packages]
      - [$variants]
```
However, this doesn't work because not all packages in the list have those variants, and those variants won't be forwarded down the DAG. The only other alternative is to manually concretize every package in each DAG, get a list of all packages with those variants, and manually set the entire dep tree on each line myself. Nothing like a good ol':
```yaml
spack:
  specs:
    - py-horovod ~rocm+cuda cuda_arch=... ^py-tensorflow ~rocm+cuda cuda_arch=... ^py-torch ~rocm+cuda cuda_arch=... ...
    - py-horovod ~cuda+rocm amdgpu_target=... ^py-tensorflow ~cuda+rocm amdgpu_target=... ^py-torch ~cuda+rocm amdgpu_target=... ...
    - py-horovod ~cuda~rocm ^py-tensorflow ~cuda~rocm ^py-torch ~cuda~rocm ...
    ...
```

[...]

PR #33296, opened 2022-10-13
Add CI stacks for ML suite on ppc64le.

nicholas-sly:
Beginner stacks for ML suite on ppc64le runners.  Copied from existing suites and modified runner tags.  Need to verify what container to use for package signing on ppc64le runs.  Many packages are likely to fail and need some massaging on the package list to start with running CI jobs.


adamjstewart:
In terms of file and test names, in the near future we'll have the cross product of:

* CPU, CUDA, and ROCm
* x86_64, ppc64le, aarch64

In the further future, I also want to add the cross product of:

* Linux, macOS

Unlike GitHub Actions, renaming these in the future isn't a big deal, but someday we'll eventually want test names like:

* ml-linux-x86_64-cpu
* ml-linux-ppc64le-cuda
* ml-macos-aarch64-rocm

In the more distant future, I'm still hoping to combine all of these into a single file: #32893 #32894

adamjstewart:
Let's hold off on this for a bit, @becker33 gave me some great ideas for how we can do this all in a single pipeline file.

adamjstewart:
Nvm, that didn't work. We should carry on with this PR. Can you rename the files as in #34299?

nicholas-sly:
Closing this in lieu of #39174 .

PR #33297, opened 2022-10-13
Scotch: add 'link_error_lib' variant

MichaelLaufer:
Adds variant for v7.0.1 that links error handling library with the main `libscotch` and `libptscotch` libraries.

This allows for Spack to package Scotch in a way that is consistent with other package management systems who link the error library by default ([Debian](https://salsa.debian.org/science-team/scotch/-/blob/master/debian/patches/build-shared-library-soname.patch), [Macports](https://github.com/macports/macports-ports/blob/master/science/scotch/files/patch-libscotch-shared.diff), [Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/sci-libs/scotch/scotch-7.0.1-r1.ebuild))

This is required by PyFR, which assumes this type of packaging.

@pghysels @FreddieWitherden

### DIFF
--- a/var/spack/repos/builtin/packages/py-pyfr/package.py
+++ b/var/spack/repos/builtin/packages/py-pyfr/package.py
@@ -57,7 +57,7 @@ class PyPyfr(PythonPackage, CudaPackage, ROCmPackage):
     # Optional dependencies
     depends_on("py-scipy", when="+scipy", type=("build", "run"))
     depends_on("metis@5.0:", when="+metis", type=("run"))
-    depends_on("scotch@6.0:", when="+scotch", type=("run"))
+    depends_on("scotch@7.0.1: +link_error_lib", when="+scotch", type=("run"))
     depends_on("cuda@8.0: +allow-unsupported-compilers", when="@:1.14.0 +cuda", type=("run"))
     depends_on("cuda@11.4.0: +allow-unsupported-compilers", when="@1.15.0: +cuda", type=("run"))
     depends_on("rocblas@5.2.0:", when="+hip", type=("run"))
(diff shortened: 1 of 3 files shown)


pghysels (approved):
That looks reasonable to me, but why not just modify PyFR to link with both `scotch` and `scotcherr`?
Maybe send the patch to https://gitlab.inria.fr/scotch/scotch

MichaelLaufer:
@pghysels 
PyFR could be changed but it would be fairly invasive as it not as simple as linking an extra library as the PyFR implements custom wrappers for each external library that it uses. I'll defer to the core dev @FreddieWitherden for further comment.

In regards to pushing upstream, the Scotch devs purposefully do not link the two to allow for custom error handling. From [scotch_user7.0.pdf](https://gitlab.inria.fr/scotch/scotch/-/blob/master/doc/scotch_user7.0.pdf), sectioj 8.1.3
```
Therefore, the linking of applications that make use of the libScotch library with standard error handling is carried out by using the following options:
“-lscotch -lscotcherr -lm”. If you want to handle errors by yourself, you
should not link with library file libscotcherr.a, but rather provide a SCOTCH
errorPrint() routine. Please refer to section 8.19 for more information.
```

That being said, I think there is value in _optionally_ packaging Scotch like some of the other major package managers.

FreddieWitherden:
The error handling mechanism in SCOTCH is fundamentally broken.  It employs a global solution (a reliance on a function which must exist in the global namespace) to a local problem (error handling).  This makes it very difficult to compose together libraries who themselves use SCOTCH (whose error printing function should be used, what happens if they disagree?)  It also results in a broken shared library which can not be dlopen'ed without first injecting error handling symbols into the global namespace.

Most distributions accept that this mechanism is broken beyond repair and so just force the default error handling routine (which is the best of a bad set of options).

pghysels:
I approved this, but I can't merge. I don't know who else needs to approve this.
Please also consider adding yourself as a maintainer.

alalazo:
@spackbot run pipeline

tldahlgren:
@spackbot run pipeline

tldahlgren (approved):
LGTM

MichaelLaufer:
@tldahlgren Can this be merged? Thanks

PR #33300, opened 2022-10-13
ECP-SDK: enable hdf5-vfd-gds when +cuda

svenevs:
- hdf5-vfd-gds:
    - Add new version 1.0.2 compatible with hdf5@1.13.
    - CMake is a build dependency.
    - Set `HDF5_PLUGIN_PATH` in the runtime environment, this plugin
      is loaded dynamically.
- SDK:
    - The VFD GDS driver only has utility when CUDA is enabled.
    - Require hdf5-vfd-gds@1.0.2+ (1.0.1 and earlier do not compile).

--------

TODO:

- [X] Finish local testing:
    - [x] https://github.com/hpc-io/vfd-gds/issues/8 (new release 1.0.2!)
- [X] Test onsite.

### DIFF
--- a/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
+++ b/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
@@ -15,17 +15,22 @@ def dav_sdk_depends_on(spec, when=None, propagate=None):
     # ie. A +c ~b -> A
     spec = Spec(spec).name
 
-    if "+" in when and len(when.split()) == 1:
-        when_not = when.replace("+", "~")
-        # If the package is in the spec tree then it must
-        # be enabled in the SDK.
-        conflicts(when_not, "^" + spec)
+    # If the package is in the spec tree then it must be enabled in the SDK.
+    if "+" in when:
+        _when_variants = when.strip("+").split("+")
+        if any(tok in when for tok in ["~", "="]):
+            tty.error("Bad token in when clause, only positive boolean tokens allowed")
+
+        for variant in _when_variants:
+            conflicts("~" + variant, when="^" + spec)
 
     # Skip if there is nothing to propagate
     if not propagate:
         return
 
-    # Map the propagated variants to the dependency variant
+    # Map the propagated variants to the dependency variant.  Some packages may need
+    # overrides to propagate a dependency as something else, e.g., {"visit": "libsim"}.
+    # Most call-sites will just use a list.
     if not type(propagate) is dict:
         propagate = dict([(v, v) for v in propagate])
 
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
```diff
@@ -108,6 +108,7 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
     dav_sdk_depends_on("faodel+shared+mpi network=libfabric", when="+faodel", propagate=["hdf5"])
 
     dav_sdk_depends_on("hdf5@1.12: +shared+mpi", when="+hdf5", propagate=["fortran"])
+    dav_sdk_depends_on("hdf5-vfd-gds", when="+cuda+hdf5", propagate=cuda_arch_variants)
```
svenevs: @kwryankrattiger is this valid?  Other places have `propagate=["cuda"]` and `"hdf5"`, but it needs to get split because of how this package should be enabled.  I think it is right, but I'm not sure if we're getting the right `conflicts` statements.

hyoklee: I think line 110 can be removed.

hyoklee (dismissed):

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
kwryankrattiger: @hyoklee 
the HDF5 dependency needs to stay, otherwise HDF5 won't be pulled in without cuda which we don't want.

@svenevs 
the the when clause nees work, this will expand to

```python
conflicts("~cuda~hdf5", when="^hdf5-vfd-gds")
```

which I think is insufficient.

We would want it to expand to

```python
conflicts("~cuda", when="^hdf5-vfd-gds")
conflicts("~hdf5", when="^hdf5-vfd-gds")
```

This requires splitting the variant string somehow. This can either be done by magic under the hood, or by passing it as an array/tuple.

I prefer magic just to keep it consistent with how `depends_on` works, something like

```python
_when_variants = when.strip("+").split("+")
if any(tok in when for tok in ["~", "="]):
    tty.error("Bad token in when clause, only positive boolean tokens allowed")

for variant in _when_variants:
    conflicts("~"+ variant, when="^" + spec.name
```

The assumption we make is none of the top level depends will be negative, and boolean..I think that will remain a good assumption, but maybe it is worth checking that they are all true boolean variants

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
```diff
@@ -15,17 +15,22 @@ def dav_sdk_depends_on(spec, when=None, propagate=None):
...
+            conflicts("~" + variant, when="^" + spec)
 
     # Skip if there is nothing to propagate
     if not propagate:
         return
 
-    # Map the propagated variants to the dependency variant
+    # Map the propagated variants to the dependency variant.  Some packages may need
+    # overrides to propagate a dependency as something else, e.g., {"visit": "libsim"}.
+    # Most call-sites will just use a list.
```
svenevs, line 33: FYI before we forget about this comment I just want to sneak it in on one of these PRs

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
svenevs: TODO: hdf5-vfd-gds@{next_release_inbound}, once they release and pin we're looking good.

hyoklee (dismissed):
I'm not familiar with ecp-data-vis-dsk. The hdf5-vfd-gds  patch looks OK to me.

svenevs:
@hyoklee thanks for following along, I've got the 1.0.2 release in here and tested things work as we expect them to!

@kwryankrattiger this is ready for final review now.

kwryankrattiger (approved):

PR #33305, opened 2022-10-14
installer.py: show timers for binary install

haampie:
Print a message of the form

```
Fetch mm:ss.  Build: mm:ss.  Total: mm:ss
```

when installing from buildcache. Previously this only happened for
source builds.

This way we get better insights into where Gitlab CI wastes its time
(fetching from a slow bucket or is our relocation logic the bottleneck?)

--

Currently the format "fetch: build: total:" is static, maybe slightly better to actually dynamically show the phases? that would be `search` `fetch` `install` for binary installs.

### DIFF
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -319,7 +319,7 @@ def _install_from_cache(pkg, cache_only, explicit, unsigned=False):
     _print_timer(
         pre=_log_prefix(pkg.name),
         pkg_id=pkg_id,
-        fetch=timer.phases.get("fetch", 0),
+        fetch=timer.phases.get("search", 0) + timer.phases.get("fetch", 0),
         build=timer.phases.get("install", 0),
         total=timer.total,
     )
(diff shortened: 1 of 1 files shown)


alalazo:
@spackbot run pipeline

alalazo:
@spackbot help

alalazo:
Not sure why `spackbot` cannot be retriggered. The corresponding develop pipeline failed, but finished. This pipeline is still waiting though.

haampie:
@spackbot run pipeline

alalazo:
Python 2.7 failed on the infamous `test_foreground_background`. Restarted.

haampie:
btw @alalazo, what do you think about changing `fetch`/`build`/`total` into a dynamic list of actual phases followed by `total`? E.g. for cmake packages `cmake` `build` `install` etc... For binary installs it would be `search` `fetch` `install`.

(Let's first get this in though ;p)

alalazo:
That sounds good to me. I'd get this in first and refine later.

alalazo (approved):

PR #33307, opened 2022-10-14
Fix pika@0.9.0 sha

aurianer:


msimberg (approved):
Ouch, thanks.

I can confirm that 0.9.0 builds now.

PR #33310, opened 2022-10-14
py-jupyter-server: add 1.21.0

manuelakuhn:
https://github.com/jupyter-server/jupyter_server/tree/v1.21.0

I did not add the py-pywinpty dependency for windows because it is not in spack yet and it depends on the build system maturin which is also not in spack.


adamjstewart (approved):

bernhardkaindl (approved):
LGTM

PR #33314, opened 2022-10-14
gitlab ci: Do not force protected build jobs to run on aws runners

scottwittenburg:
To allow UO runners to build protected binaries for public deployment, we need to remove the constraint that all protected rebuild jobs run on aws runners.


scottwittenburg:
fyi @eugeneswalker @kotfic

eugeneswalker (approved):
Thanks!

PR #33315, opened 2022-10-14
depfile: reduce dag with `--use-buildcache` flag.

haampie:
When installing some/all specs from a buildcache, build edges are pruned
from those specs. This can result in a much smaller effective DAG. Until
now, `spack env depfile` would always generate a full DAG.

Ths PR adds the `spack env depfile --use-buildcache` flag that was
introduced for `spack install` before. This way, not only can we drop
build edges, but also we can automatically set the right buildcache
related flags on the specific specs that are gonna get installed.

On top of that, this PR introduces `spack env depfile [spec 1] [spec 2]...`
that reduces the dag to particular specs of interest in the environment.

This way we get parallel installs of binary deps of a particular spec
that's built from sources, without redundancy, which is useful for
Gitlab CI.

Example use:

```console
# install from buildcache w/o build deps
spack env depfile -o Makefile --use-buildcache only && make -j8

# drop build deps of deps
spack env depfile -o Makefile --use-buildcache dependencies:only && make -j8

# restrict environment install to pkg-a and pkg-b + deps
# build those two from sources
# dependencies are pulled from a buildcache, their build deps dropped
spack env depfile -o Makefile --use-buildcache package:never,dependencies:only pkg-a pkg-b && make -j8 
```

### DIFF
--- a/lib/spack/spack/cmd/env.py
+++ b/lib/spack/spack/cmd/env.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import argparse
 import os
 import shutil
 import sys
@@ -15,6 +16,8 @@ import llnl.util.tty as tty
 from llnl.util.tty.colify import colify
 from llnl.util.tty.color import colorize
 
+import spack.cmd
+import spack.cmd.common
 import spack.cmd.common.arguments
 import spack.cmd.common.arguments as arguments
 import spack.cmd.install
@@ -600,6 +603,15 @@ def env_depfile_setup_parser(subparser):
         help="disable POSIX jobserver support.",
     )
     subparser.add_argument(
+        "--use-buildcache",
+        dest="use_buildcache",
+        type=arguments.use_buildcache,
+        default="package:auto,dependencies:auto",
+        metavar="[{auto,only,never},][package:{auto,only,never},][dependencies:{auto,only,never}]",
+        help="When using `only`, redundant build dependencies are pruned from the DAG. "
+        "This flag is passed on to the generated spack install commands.",
+    )
+    subparser.add_argument(
         "-o",
         "--output",
         default=None,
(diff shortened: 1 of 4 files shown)


haampie:
@spackbot fix style

haampie:
@trws do you have opinions here?

So, `--shallow` now corresponds with dropping non-direct build deps. This is quite... specific. Only useful for CI type of situations where you wanna build one (or more) specific packages from sources, and require their dependencies to be installed from a binary cache.

But maybe `shallow` is not a great name, cause you might as well expect a DAG where _all_ build type deps are pruned.

`--prune-build-deps=all/package/dependencies` or so is also an option...

haampie:
@alalazo mentioned:

```
spack env depfile --[build-|install-]from-sources=all/root/none
```

If we do that, we can immediately add the right `--cache-only` type of flags too.

But actually it's kinda similar to the amazing `--use-buildcache=...` flag :thinking: then maybe that should be used for consistency.

lib/spack/spack/cmd/env.py
```diff
@@ -613,17 +620,103 @@ def env_depfile_setup_parser(subparser):
...
+
+class FullMakeTargetVisitor(MakeTargetVisitor, FullDagNeighborsMixin):
+    pass
+
+
+class ShallowMakeTargetVisitor(MakeTargetVisitor, MinimalDagNeighborsMixin):
+    pass
+
+
+def traverse_breadth_first(visitor, specs=[]):
```
scheibelp, line 711: `Spec.traverse(order='pre')` should match this in all the desired ways and might avoid some corner cases like:

```
1____
| \  \
2  3  4
|  |  |
|  5  6
|_/__/
|
7
```

i.e., we want something more-sophisticated than breadth first or we may end up visiting `7` before `[5, 6]` (if your implementation avoids this, we should probably call it something like "topological" rather than "breadth-first").

haampie: Hm, the point is I have multiple root specs, and need to handle the roots differently from the dependencies. Given that the second root may appear as a dependency of the first root, I don't want depth-first search. Breadth-first search initialized with all root specs guarantees that roots are found at depth 0 and that all connected nodes are visited once. I don't see the need for topological sort, for make targets all I need is the adjacency list.

See the bit

```python
    def neighbors(self, node):
        # At depth zero we return build / test deps too
        if node.depth == 0:
            return node.spec.dependencies(deptype="all")
        # Otherwise we just return link/run type deps
        return node.spec.dependencies(deptype=("link", "run"))
```

haampie:
This is ready for review, again. Includes tests now.

haampie:
@scheibelp I've moved the breadth-first traversal bits to #33406 so that this PR doesn't introduce its own "custom" traversal.

trws (approved):

PR #33321, opened 2022-10-14
Update command option for example

gsjaardema:
The `spack info <package>` command does not show the `Virtual Packages:` output unless the `--virtuals` command option is passed.  Before this changes, the information that the command is supposed to be illustrating is not shown in the example and is confusing.


tldahlgren (approved):
Great catch.  Thanks!

tldahlgren:
@spackbot run pipeline

PR #33322, opened 2022-10-14
libcatalyst: add 2.0.0-rc3 release

mathstuf:


var/spack/repos/builtin/packages/libcatalyst/package.py
```diff
@@ -12,9 +12,12 @@ class Libcatalyst(CMakePackage):
 
     homepage = "https://gitlab.kitware.com/paraview/catalyst"
     git = "https://gitlab.kitware.com/paraview/catalyst.git"
+    url = "https://gitlab.kitware.com/api/v4/projects/paraview%2Fcatalyst/packages/generic/catalyst/v2.0.0/catalyst-v2.0.0.tar.gz"
```
vicentebolea, line 15: This rule shows "not-found".

var/spack/repos/builtin/packages/libcatalyst/package.py
```diff
@@ -12,9 +12,12 @@ class Libcatalyst(CMakePackage):
...
 
     maintainers = ["mathstuf"]
 
+    version("2.0.0-rc3", sha256="8862bd0a4d0be2176b4272f9affda1ea4e5092087acbb99a2fe2621c33834e05")
```
vicentebolea, line 19: how about we specify this as a git revision rather than a url?

vicentebolea (changes requested):

var/spack/repos/builtin/packages/libcatalyst/package.py
mathstuf: 2.0.0 is in rc3, so there isn't a final. Does the replacement logic know how to deal with `-rc3` in template URLs?

var/spack/repos/builtin/packages/libcatalyst/package.py
mathstuf: For what purpose? Why is cloning better than a tarball?

tldahlgren (approved):
Confirmed sha256 with the new URL.

Typically people do not "prefer" release candidate versions (i.e., mark the last stable as preferred).  Is this something you want to do here?

var/spack/repos/builtin/packages/libcatalyst/package.py
vicentebolea: We can skip having both url|git property here

mathstuf:
It's fine to defer this until 2.0.0 is tagged as final too. I don't forsee any ABI changes that would cause problems though.

I can also mark the date stamp version as preferred in the meantime if that's better.

var/spack/repos/builtin/packages/libcatalyst/package.py
vicentebolea: Right, `s/2.0.0/2.0.0-rc3`pointed me to an existing url

vicentebolea: Yes the replacement logic would know how to handle this

tldahlgren: FWIW. I checked out the PR *then* performed `spack checksum` and it works for me.

tldahlgren:
> It's fine to defer this until 2.0.0 is tagged as final too. I don't forsee any ABI changes that would cause problems though.
> 
> I can also mark the date stamp version as preferred in the meantime if that's better.

I was just suggesting the flagging the date stamped version as preferred (https://spack.readthedocs.io/en/latest/packaging_guide.html#version-selection) only because I've had people complain about release candidates as (default) preferred versions.

It's really up to you.

var/spack/repos/builtin/packages/libcatalyst/package.py
tldahlgren: @vicentebolea Would that have made a difference for your check?

vicentebolea: @tldahlgren definitely, checking the url of a PR containing a url change with spack checksum would have validated this PR. I will use it next time for this situations, thanks for sharing your workflow :)

tldahlgren: Alternatively you can temporarily add the new/changed URL but I have a script to simplify the fetch/checkout process for me.

mathstuf:
IIRC, the datestamp is just a "random" snapshot of even lower stability guarantees than an rc (we certainly don't have a branch dedicated to whatever state the repo was in at that point).

var/spack/repos/builtin/packages/libcatalyst/package.py
vicentebolea: I have used the gh cli utility in this case, its a great help in situations like this

var/spack/repos/builtin/packages/libcatalyst/package.py
vicentebolea: Before the auto-merge we have this comment opened too. I am no expert on this but I wonder what would be the gain of adding two urls (git|http URL) for a package? with a git url we can also specify git tags.

mathstuf: I don't think asking people to contact `gitlab.kitware.com` for this package rather than being able to cache a tarball would work long-term for some HPC deployments…

vicentebolea: Fair enough

vicentebolea (approved):

mathstuf:
It looks like reviews are OK. What's the next step?

tldahlgren:
> It looks like reviews are OK. What's the next step?

We need to wait for at least all required checks to pass.  Looks like the pipeline check is still waiting on `base develop commit` before it starts the pipeline stacks for this PR.

tldahlgren:
> > It looks like reviews are OK. What's the next step?
> 
> We need to wait for at least all required checks to pass. Looks like the pipeline check is still waiting on `base develop commit` before it starts the pipeline stacks for this PR.

Hmm.  Looks like several of those stacks failed, with a couple complaining about broken specs.

tldahlgren:
I selected "Update Branch" to see if that helps the checks.

PR #25672, opened 2021-08-28
Switch http to https where latter exists

tldahlgren:
Scripts were used to extract, check for existence, and change the URLs from `http:` to `https:`.  This PR is the result.

Update: The second pass skipped homepage (only) changes to avoid duplication with recently merged PR(s). Note that the programmatic extraction and update processes did not distinguish between URLs in comments versus meaningful code.

PR #3113, opened 2017-02-10
Keep Curl Bootstrap Working With http://

citibeth:
In order to download from `https://` websites, Spack needs a working `curl`, with a new-enough OpenSSL.  That is not the case on some systems; in that case, Spack must install `curl` itself:
```
curl@7.52.1%clang@7.3.0-apple arch=darwin-elcapitan-x86_64 
    ^openssl@1.1.0d%clang@7.3.0-apple arch=darwin-elcapitan-x86_64 
        ^perl@5.24.1%clang@7.3.0-apple+cpanm arch=darwin-elcapitan-x86_64 
        ^zlib@1.2.10%clang@7.3.0-apple+pic+shared arch=darwin-elcapitan-x86_64 
```
The problem would appear if any of those packages requires SSL to download --- then, Spack would be unable to bootstrap a working `curl` if it didn't already have one.  This PR adds comments that discourage authors from changing these four packages to `https://`.

PR #33324, opened 2022-10-14
zlib, openssl: return to `http` URLs

mathstuf:
In #3113, `https` was removed to ensure that `curl` can be bootstrapped without SSL being present. This was lost in #25672 which aimed to use `https` where possible.


bernhardkaindl (approved):
LGTM, we still have the sha256 to ensure no unverified source gets built.

haampie:
```
> GET /source/openssl-1.1.1d.tar.gz HTTP/1.1
< HTTP/1.1 301 Moved Permanently
< Location: https://www.openssl.org/source/openssl-1.1.1d.tar.gz
```

It doesn't help :man_shrugging:

mathstuf:
Bleh. Not much to be done then :/ . Are there any known deployments with busted SSL pre-bootstrap?

haampie:
We can ask @tgamblin to make `http://mirror.spack.io` a thing, it's OK to use http when we the shasum method is strong enough.

PR #33327, opened 2022-10-14
vsearch: adding version 2.22.1

snehring:
The new deps aren't strictly required, but the functionality they add is almost never going to be undesired. So I just opted to make them deps without corresponding variants. Most likely they were already being pulled in from the system installs automatically for many folks.


tldahlgren (approved):
Confirmed sha256.

snehring:
@spackbot re-run pipeline

PR #33328, opened 2022-10-14
mpich: patch @3.4 to fix checking whether the datatype is contiguous

skosukhin:
1. Fix a typo in the argument passed to the configure script of `mpich`.
2. Add a patch to fix a bug that affects versions `3.4.x` of the package. I have to care about older versions of the package, in particular `3.4.3`, because more recent versions of `mpich` seem to have bugs (to be investigated and reported) that prevent us from using them.


alalazo:
LGTM, pending @raffenet and @yfguo reviews.

raffenet (dismissed):
@skosukhin is there another issue besides https://github.com/pmodels/mpich/issues/4358 for newer versions of MPICH? Sorry for the lack of response, and thanks for putting together this fix.

eugeneswalker (dismissed):

skosukhin:
@raffenet the only thing we currently know about MPICH 4.x versions is that they seem to have "some issues with the graph topology communicators in MPICH" (we discuss them [here](https://gitlab.dkrz.de/dkrz-sw/yaxt/-/issues/10)). I hope that my colleagues will come up with bug reports and minimal reproducers soon.

alalazo (approved):

raffenet (approved):

PR #33338, opened 2022-10-14
xz: add 5.2.7

michaelkuhn:


tldahlgren (dismissed):
Confirmed sha256.

michaelkuhn:
@spackbot run pipeline

haampie:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

alalazo:
@michaelkuhn There are legit issues with pipelines

Screenshot from 2022-10-28 10-48-14

michaelkuhn:
> @michaelkuhn There are legit issues with pipelines

Thanks for the hint, which log should I look out for? I can't seem to find that in the latest run. (The set of errors being different each run also doesn't help.)

Edit: Sorry, I didn't check your screenshot thoroughly enough, now I found it in the hpctoolkit build. Will take a look.

michaelkuhn:
OK, I was able to reproduce the problem locally. It seems that hpctoolkit wants to use the static library built with `-fPIC`, which somehow interferes with xz's new symbol versioning. The change I just pushed now disables symbol versioning for xz, which seems to work around the problem (and we shouldn't need symbol versioning anyway).

haampie:
Sorry, but we should _not_ disable symbol versioning, they use it correctly: `-fvisibilty=hidden`  + `-Wl,--version-script=...` to ensure that only their public API is exposed in the shared library, which reduces the risk of symbol clashes and also improves startup time since the dynamic loader has to deal with fewer symbols.

michaelkuhn:
I have found a way (and pushed a change) to force xz to use the old generic symbol versioning, which also seems to work (they introduced a new approach in 5.2.7, which breaks hpctoolkit).

haampie:
Wouldnt it be best if hpctoolkit just had an upperbound on xz and dropped that with the next release when they are compatible?

michaelkuhn:
xz's new approach triggers a problem with hpctoolkit's autotools tests (ld refuses to link with `version node not found for symbol`), so I'm not sure how easy it is to fix that.

haampie:
Ping @mwkrentel :D

mwkrentel:
@michaelkuhn I suggest you use @haampie's suggestion and restrict
hpctoolkit's dependency to <= 5.2.5.

```
depends_on("xz@:5.2.5 +pic", type="link")
```

You could add that to the hpctoolkit `package.py` in this PR and I
would be Ok with that.

Of course, I can't promise that will be the only failure.  The logs
seem to have a bunch of `json-c` errors that I don't understand since
json-c doesn't depend on xz.

If it becomes necessary to adjust the symbol versioning, I would make
that a variant with default native (unmodified).  But I don't yet know
if that will be necessary or not.  I'm curious if the symbol
versioning causes issues for other packages.

Also, I would also add 5.2.6 like the other PR did.

michaelkuhn:
I just pushed a new version that 1) also adds 5.2.6 and 5.2.1 and 2) limits hpctoolkit's xz dependency to `@:5.2.6`.

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
This has now also finally passed CI. :smiling_face_with_tear:

haampie (approved):

haampie:
congrats on getting the most difficult to merge pr merged!

mwkrentel:
@alalazo  @tldahlgren @haampie 
Now that the ci pipeline succeeded and the PR is merged, I can't find
the details of the tests.  But I remember a few things and I have some
remaining questions.

1. Why did the pipeline succeed now when it previously failed and
there were no changes to the patch?

2. Even when the pipeline succeeded and got a green check, there were
pieces of the piepline that failed and got a red x.  I think one was
`e4s-mac-pr-build` and it failed at `bzip2` with the mesage `SystemExit: 1`.

There wasn't an elaboration of why bzip2 actually failed, like a
partial transcript of the build output, just SystemExit: 1.

So, why did bzip2 fail, and was it really bzip2 that failed?

3. And if e4s-mac-pr-build failed, then why did ci/pipeline succeed?

Unfortunately, now that the PR is merged, I can't find the details.
Do they still exist somewhere or are they erased?
I downloaded the `artifacts`, but they don't have the logs that I want.

I'll see if I can find something similar in another PR.
Thanks.

michaelkuhn:
You can find the last CI run here: https://gitlab.spack.io/spack/spack/-/pipelines/206689 (click the green check mark next to the commit and scroll down)

I don't really know what's going on with the CI pipelines but they fail a lot and undeterministically.

haampie:
The mac pipeline is allowed to fail

mwkrentel:
> click the green check mark next to the commit and scroll down

Ok, there it is.  I would never have thought to click there.
Thanks!

I still don't get why bzip2 fails, maybe it always fails on MAC, I
don't know.

stephenmsachs:
I ran into an issue with this commit: I install an MPI application (openfoam), run it and then install `hpctoolkit +mpi`. It will then re-install MPI (in this case openmpi) and a lot of other dependencies as the original installation was based on `xz` later than 5.2.6. The MPIs then don't match and I cannot run openfoam with hpctoolkit.

If I understand correctly the fix used here is in the wrong place. @michaelkuhn which compiler threw the `version node not found for symbol` error in hpctoolkit? I ran into that error with `xz%intel` and added a fix in https://github.com/spack/spack/pull/33839. Long story short, I think we need to restrict `xz` package to those versions that will build and link rather than adding conflicts in dependent packages.

michaelkuhn:
> If I understand correctly the fix used here is in the wrong place. @michaelkuhn which compiler threw the `version node not found for symbol` error in hpctoolkit? I ran into that error with `xz%intel` and added a fix in #33839. Long story short, I think we need to restrict `xz` package to those versions that will build and link rather than adding conflicts in dependent packages.

I think the problem happened with GCC. It's not a problem with xz, though, but with hpctoolkit's checks not being compatible with newer versions of xz. Other packages build with newer versions of xz.

stephenmsachs:
You are right, I am seeing the error with GCC now too. Thanks for checking. They might have fixed this in `xz@5.2.10` https://github.com/xz-mirror/xz/commit/ef315163efdf82e327608ae9b8df1bfea42eb6b5 . I need to check that again.

mwkrentel:
Yes, currently xz 5.2.7 fails the configure test for hpctoolkit.  We
test that xz provides a static library with -fPIC.  Up to 5.2.6 works,
but 5.2.7 fails the test, I haven't tried 5.2.10 or later.

My apologies that I haven't gotten to this sooner (always too much to do).

@stephenmsachs As a workaround, you could provide MPI as an external.
That is, build openfoam + MPI as you do now.  Then take the spack MPI
(or system MPI) and specify that as a spack external to buiild hpctoolkit.

The xz that goes into MPI and xz for hpctoolkit don't have to be the
same version, but spack currently doesn't allow for this in a single dag.
Writing it as an external works around that issue.

Btw, for hpctoolkit, we build everything with GNU.  So fixing this for
Intel isn't going to solve the problem.

mwkrentel:
Good news!  I did some preliminary tests, and it looks like hpctoolkit
builds cleanly with both xz 5.2.10 and 5.4.1.

That gives an easy solution here, I'll change the conflict to just the
single rev 5.2.7.  Then I can take my time understanding exactly what
went wrong and how to avoid it in the future.

stephenmsachs:
Thanks for the hint with that workaround! Unfortunately I ran into another issue with OpenMPI. I'll open a separate issue on that.

PR #33344, opened 2022-10-14
Bugfix HIP and aluminum rocm build

bvanessen:
Fixed two bugs in the HIP package recipe:

- [x] `HIP_PATH` was being set to the actual spec, and not the spec prefix.
- [x] HIP is expected to be in `/opt/rocm-x.y.z/hip`
      but its libraries can exist at both `/opt/rocm-x.y.z/hip/lib` and
      `/opt/rocm-x.y.z/lib`.  This means that the external detection logic may
      find it in either and it turns out that some modules only expose one.
      of those two locations.  Logic is added to ensure that the internal
      `HIP_PATH` and associated `ROCM_PATH` are correctly set in both cases.

### DIFF
--- a/var/spack/repos/builtin/packages/hip/package.py
+++ b/var/spack/repos/builtin/packages/hip/package.py
@@ -289,7 +289,6 @@ class Hip(CMakePackage):
             # For external packages we only assume the `hip` prefix is known,
             # because spack does not set prefixes of dependencies of externals.
             hip_libs_at_top = "hip" not in self.spec.prefix
-            print(f'I thin k that hip libas at top is {hip_libs_at_top}')
             # We assume self.spec.prefix is  /opt/rocm-x.y.z for rocm-5.2.0 and newer
             # and /opt/rocm-x.y.z/hip for older versions
             if self.spec.satisfies("@5.2.0:"):
--- a/var/spack/repos/builtin/packages/hip/package.py
+++ b/var/spack/repos/builtin/packages/hip/package.py
@@ -288,12 +288,23 @@ class Hip(CMakePackage):
         if self.spec.external:
             # For external packages we only assume the `hip` prefix is known,
             # because spack does not set prefixes of dependencies of externals.
+            hip_libs_at_top = "hip" not in self.spec.prefix
+            print(f'I thin k that hip libas at top is {hip_libs_at_top}')
             # We assume self.spec.prefix is  /opt/rocm-x.y.z for rocm-5.2.0 and newer
             # and /opt/rocm-x.y.z/hip for older versions
             if self.spec.satisfies("@5.2.0:"):
                 rocm_prefix = Prefix(self.spec.prefix)
             else:
-                rocm_prefix = Prefix(os.path.dirname(self.spec.prefix))
+                # We assume self.spec.prefix is /opt/rocm-x.y.z/hip and rocm has a
+                # default installation with everything installed under
+                # /opt/rocm-x.y.z
+                # Note that since the key hip library can also exist at the top of the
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/hip/package.py
```diff
@@ -291,7 +291,13 @@ def get_paths(self):
             # We assume self.spec.prefix is /opt/rocm-x.y.z/hip and rocm has a
             # default installation with everything installed under
             # /opt/rocm-x.y.z
-            rocm_prefix = Prefix(os.path.dirname(self.spec.prefix))
+            # Note that since the key hip library can also exist at the top of the
+            # /opt/rocm-x.y.z/lib tree, it is possible that the package is detected
+            # without the correct prefix.  Work around it.
```
tgamblin: This is confusing me.  If you've registered `hip` as an external but it's in `/opt/rocm-x.y.z/` instead of `/opt/rocm-x.y.z/hip`, shouldn't we make Spack get the external definition right?

I *think* the better way to do this is to handle this in an override of `determine_spec_details()` so that external detection succeeds, but I don't know if that handles all the cases if `libamdhip64` can really exist in `/opt/rocm-x.y.z/lib` and `rocm-x.y.z/hip/lib`.  If it is in the former will it also exist in the latter if the latter exists?

var/spack/repos/builtin/packages/hip/package.py
```diff
@@ -300,7 +306,13 @@ def get_paths(self):
                 msg += "a workaround."
                 raise RuntimeError(msg)
 
+            if "hip" in self.spec.prefix:
```
tgamblin: Given that this is a pretty special heuristic I'd write it once, next to the complicated comment, and store the result in some descriptive variable, like `hip_libs_at_top` or something.

var/spack/repos/builtin/packages/hip/package.py
```diff
@@ -309,6 +321,7 @@ def get_paths(self):
             }
         else:
             paths = {
+                "hip-path": self.spec,
```
tgamblin: Is this one a typo?  I don't think we ever want this to be `self.spec`, right?  Should it be `self.spec.prefix`?

tgamblin (changes requested):

var/spack/repos/builtin/packages/hip/package.py
bvanessen: @tgamblin Yes this is the bug.

bvanessen: @tgamblin Actually that is a second spot that I missed.

var/spack/repos/builtin/packages/hip/package.py
bvanessen: @tgamblin I believe that it will exist in both of them, but part of the problem is that the modules which set this up are inconsistent in what they expose.  So if you use one it will expose the more complete path, and the other one will only expose it at the top.

var/spack/repos/builtin/packages/hip/package.py
bvanessen: Done

var/spack/repos/builtin/packages/hip/package.py
tgamblin: Can you say how you’re detecting this? Are you looking at variables from modules in a script?  Which ones?

bvanessen: @tgamblin I just have it looking for both paths and if it doesn't find it under a `/hip/` directory then it asserts that it is at the top of the `rocm` tree.

tgamblin (approved):

PR #33349, opened 2022-10-15
py-tensorflow: fix zlib

haampie:
tensorflow uses bazel, bazel vendors all deps and has urls pointing to a non-existing zlib url, since the new release of zlib of yesterday.


haampie:
@spackbot fix style

adamjstewart (approved):
Thanks for the fix! It's amazing that a project would prefer hardcoded URLs over simply using system installs of a package. Hopefully someday we can convince the devs to allow external deps instead of vendored deps.

I imagine all historical versions would also need a patch like this to build? I noticed a lot of similar failures when creating the ML pipeline, but I "solved" those by adding newer versions of packages like TF metadata so that the pipeline no longer tried to build old TF. I certainly won't require you to add patches for older versions, and I don't think any of us have the bandwidth to maintain ancient TF, but at least we know the fix if someone does need an old version.

PR #33361, opened 2022-10-17
database: don't warn adding missing build deps

haampie:
When installing an individual spec `spack --only=package --cache-only /xyz`
from a buildcache, Spack currently issues tons of warnings about
missing build deps (and their deps) in the database.

This PR disables these warnings, since it's fine to have a spec without
its build deps in the db (they are just "missing").


haampie:
@spackbot run pipeline

haampie:
@spackbot ping

kwryankrattiger (approved):
LGTM, thanks!

PR #33367, opened 2022-10-17
installer.py: traverse_dependencies has local deptype

haampie:
Currently `traverse_dependencies` fixes deptypes to traverse once and
for all in the recursion, but this is incorrect, since deptypes depend
on the node (e.g. if it's a dependency and cache-only, don't follow
build type edges, even if the parent is built from sources and needs
build deps.)

This was missed in #32537


kwryankrattiger (approved):
LGTM

PR #33379, opened 2022-10-17
Enable the spdlog recipe to use the bundled version of the fmt library

bvanessen:
that ships with its repo.

### DIFF
--- a/var/spack/repos/builtin/packages/spdlog/package.py
+++ b/var/spack/repos/builtin/packages/spdlog/package.py
@@ -45,17 +45,17 @@ class Spdlog(CMakePackage):
 
     variant("shared", default=True, description="Build shared libraries (v1.4.0+)")
     variant(
-        "bundled_fmt",
-        default=True,
+        "fmt_external",
+        default=False,
         description="Build using external fmt libraries instead of bundled one",
     )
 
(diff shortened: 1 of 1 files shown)


bvanessen:
@alalazo Do you have time to look at this one?

var/spack/repos/builtin/packages/spdlog/package.py
```diff
@@ -59,7 +64,7 @@ def cmake_args(self):
             args.extend(
                 [
                     self.define_from_variant("SPDLOG_BUILD_SHARED", "shared"),
-                    self.define("SPDLOG_FMT_EXTERNAL", "ON"),
+                    self.define("SPDLOG_FMT_EXTERNAL", ("~bundled_fmt" in self.spec)),
```
alalazo: Can we use a variant that has the same logic as this option? In this way we can use `define_from_variant`

alalazo (changes requested):
Only a minor request.

var/spack/repos/builtin/packages/spdlog/package.py
bvanessen: Done.

alalazo (approved):

alalazo (approved):

PR #33382, opened 2022-10-17
support pkgconfig for mkl

rscohn2:
Support pkgconfig by adding the mkl directories to PKG_CONFIG_PATH. Some of the pc files depend on pc files provided by the compiler, so try to add them as well.

@haampie 
@rgommers 
@adamjstewart

PR #33383, opened 2022-10-17
intel-oneapi-compilers: fix Python 2.7 compliance

alalazo:
Unit tests are currently failing on `develop` because of this comma too much. See #33281

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
@@ -131,7 +131,7 @@ class IntelOneapiCompilers(IntelOneApiPackage):
                 placement="fortran-installer",
                 when="@{0}".format(v["version"]),
                 expand=False,
-                **v["ftn"],
+                **v["ftn"]
             )
 
     @property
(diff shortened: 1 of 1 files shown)


adamjstewart:
Want to review #33382 when this is merged?

adamjstewart (approved):

adamjstewart:
Won't this break black? Why does Python 2.7 choke on this?

alalazo:
I think this explains it: https://github.com/psf/black/issues/1463

rscohn2 (approved):
Black is happy either way. Sorry for merging this. Can the python 2.7 compat test be turned on for package commits?

tgamblin:
I am not sure what we can do about this.  We pin `black` to 21.12b0 in our CI and in `spack style` because that's the last version of `black` that supports Python 2.7.  Unfortunately, this issue is present *in* that and all current versions of `black`, and will only be fixed in newer versions if it's fixed at all (and then only b/c it affects Python 3.5). 😖 

We *could* run package audits (which load all packages) for 2.7 in addition to a 3.x Python version until we stop supporting 2.7 in `develop`.

alalazo:
This has multiple failures in pipelines. I'm going to force merge, since it's unlikely that this change will break anything :slightly_smiling_face:

PR #33385, opened 2022-10-17
Testing CI on Windows

### LINKED ISSUE #33347: Unit tests on Windows are failing depite CI is passing
### Steps to reproduce

Reporting from a discussion among maintainers in Slack. Unit tests running on Windows are failing, despite CI on them passing. Examples of failing unit tests with :heavy_check_mark: CI:
- https://github.com/spack/spack/actions/runs/3246813188/jobs/5326063183
- https://github.com/spack/spack/actions/runs/3246813188/jobs/5326063112

The issue making unit tests fail possibly started in https://github.com/spack/spack/pull/30690, though we also need to fix CI scripts so that next time they report on the error immediately.

### Error message

Screenshot from 2022-10-14 18-39-56

NA (It happens on Github Actions)

### General information

johnwparent:
Windows CI is currently reporting only positives, many of which are false. Test to see under which conditions this occurs as unable to replicate with local pwsh env.

Note: This PR should be failing (at least the Windows tests), that means its working.

 Resolves #33347

### DIFF
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -2379,21 +2379,6 @@ def files_in(*search_paths):
     return files
 
 
-@system_path_filter
-def strip_extensions(*paths):
-    """Returns all paths passed as arguments with file extensions removed
-    If a file has no extension it is not modified. This method is predominately
-    for use
-
-    Args:
-        *paths: paths to have file extensions removed
-
-    Returns:
-        List of paths
-    """
-    return [os.path.splitext(x)[0] for x in paths]
-
-
 def is_readable_file(file_path):
     """Return True if the path passed as argument is readable"""
     return os.path.isfile(file_path) and os.access(file_path, os.R_OK)
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -99,7 +99,9 @@ def getuid():
 def rename(src, dst):
     # On Windows, os.rename will fail if the destination file already exists
     if is_windows:
-        if os.path.exists(dst):
+        # Windows path existence checks will sometimes fail on junctions/links/symlinks
+        # so check for that case
+        if os.path.exists(dst) or os.path.islink(dst):
             os.remove(dst)
     os.rename(src, dst)
 
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -2374,6 +2374,21 @@ def files_in(*search_paths):
     return files
 
 
+@system_path_filter
+def strip_extensions(*paths):
+    """Returns all paths passed as arguments with file extensions removed
+    If a file has no extension it is not modified. This method is predominately
+    for use
+
+    Args:
+        *paths: paths to have file extensions removed
+
+    Returns:
+        List of paths
+    """
+    return [os.path.splitext(x)[0] for x in paths]
+
+
 def is_readable_file(file_path):
     """Return True if the path passed as argument is readable"""
     return os.path.isfile(file_path) and os.access(file_path, os.R_OK)
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -288,7 +288,13 @@ def filter_file(regex, repl, *filenames, **kwargs):
         shutil.copy(filename, tmp_filename)
 
         try:
-            extra_kwargs = {"errors": "surrogateescape"}
+            # To avoid translating line endings (\n to \r\n and vis versa)
+            # we force os.open to ignore translations and use the line endings
+            # the file comes with
+            extra_kwargs = {
+                "errors": "surrogateescape",
+                "newline": ""
+                }
 
             # Open as a text file and filter until the end of the file is
             # reached or we found a marker in the line if it was specified
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -293,10 +293,7 @@ def filter_file(regex, repl, *filenames, **kwargs):
             # To avoid translating line endings (\n to \r\n and vis versa)
             # we force os.open to ignore translations and use the line endings
             # the file comes with
-            extra_kwargs = {
-                "errors": "surrogateescape",
-                "newline": ""
-                }
+            extra_kwargs = {"errors": "surrogateescape", "newline": ""}
 
             # Open as a text file and filter until the end of the file is
             # reached or we found a marker in the line if it was specified
--- a/lib/spack/spack/compiler.py
+++ b/lib/spack/spack/compiler.py
@@ -599,7 +599,7 @@ class Compiler(object):
         # Windows compilers generally have an extension of some sort
         # as do most files on Windows, handle that case here
         if is_windows:
-            ext = r"\.(exe|bat)+"
+            ext = r"\.(exe|bat)"
             cls_suf = [suf + ext for suf in cls.suffixes]
             ext_suf = [ext]
             suffixes = suffixes + cls.suffixes + cls_suf + ext_suf
--- a/lib/spack/spack/compiler.py
+++ b/lib/spack/spack/compiler.py
@@ -599,7 +599,7 @@ class Compiler(object):
         # Windows compilers generally have an extension of some sort
         # as do most files on Windows, handle that case here
         if is_windows:
-            ext = r"\.[a-zA-Z]+"
+            ext = r"\.(exe|bat)+"
             cls_suf = [suf + ext for suf in cls.suffixes]
             ext_suf = [ext]
             suffixes = suffixes + cls.suffixes + cls_suf + ext_suf
--- a/lib/spack/spack/compiler.py
+++ b/lib/spack/spack/compiler.py
@@ -9,6 +9,7 @@ import os
 import platform
 import re
 import shutil
+import sys
 import tempfile
 from typing import List, Optional, Sequence
 
@@ -26,6 +27,8 @@ from spack.util.environment import filter_system_paths
 from spack.util.path import system_path_filter
 
 __all__ = ["Compiler"]
+
+is_windows = sys.platform == "win32"
 
 
 @llnl.util.lang.memoized
@@ -592,7 +595,16 @@ class Compiler(object):
         # defined for the compiler
         compiler_names = getattr(cls, "{0}_names".format(language))
         prefixes = [""] + cls.prefixes
-        suffixes = [""] + cls.suffixes
+        suffixes = [""]
+        # Windows compilers generally have an extension of some sort
+        # as do most files on Windows, handle that case here
+        if is_windows:
+            ext = r"\.[a-zA-Z]+"
+            cls_suf = [suf+ext for suf in cls.suffixes]
+            ext_suf = [ext]
+            suffixes = suffixes + cls.suffixes + cls_suf + ext_suf
+        else:
+            suffixes = suffixes + cls.suffixes
         regexp_fmt = r"^({0}){1}({2})$"
         return [
             re.compile(regexp_fmt.format(prefix, re.escape(name), suffix))
--- a/lib/spack/spack/compiler.py
+++ b/lib/spack/spack/compiler.py
@@ -599,7 +599,7 @@ class Compiler(object):
         # Windows compilers generally have an extension of some sort
         # as do most files on Windows, handle that case here
         if is_windows:
-            ext = r"\.(exe|bat)"
+            ext = r"\.(?:exe|bat)"
             cls_suf = [suf + ext for suf in cls.suffixes]
             ext_suf = [ext]
             suffixes = suffixes + cls.suffixes + cls_suf + ext_suf
--- a/lib/spack/spack/compiler.py
+++ b/lib/spack/spack/compiler.py
@@ -600,7 +600,7 @@ class Compiler(object):
         # as do most files on Windows, handle that case here
         if is_windows:
             ext = r"\.[a-zA-Z]+"
-            cls_suf = [suf+ext for suf in cls.suffixes]
+            cls_suf = [suf + ext for suf in cls.suffixes]
             ext_suf = [ext]
             suffixes = suffixes + cls.suffixes + cls_suf + ext_suf
         else:
--- a/lib/spack/spack/compilers/__init__.py
+++ b/lib/spack/spack/compilers/__init__.py
@@ -722,9 +722,9 @@ def make_compiler_list(detected_versions):
         compiler_cls = spack.compilers.class_for_compiler_name(compiler_name)
         spec = spack.spec.CompilerSpec(compiler_cls.name, version)
         paths = [paths.get(x, None) for x in ("cc", "cxx", "f77", "fc")]
-        # in order for tests to function, this needs to refer to current host
-        # as determined by Spack, not archspec
-        target = archspec.cpu.TARGETS[str(spack.platforms.host().target("default_target"))]
+        # TODO: johnwparent - revist the following line as per discussion at:
+        # https://github.com/spack/spack/pull/33385/files#r1040036318
+        target = archspec.cpu.host()
         compiler = compiler_cls(spec, operating_system, str(target.family), paths)
         return [compiler]
 
--- a/lib/spack/spack/compilers/__init__.py
+++ b/lib/spack/spack/compilers/__init__.py
@@ -633,7 +633,7 @@ def arguments_to_detect_version_fn(operating_system, paths):
                         compiler_id = CompilerID(operating_system, compiler_name, None)
                         detect_version_args = DetectVersionArgs(
                             id=compiler_id,
-                            variation=NameVariation(*match.groups()),
+                            variation=NameVariation(*match.groups()[:2]),
                             language=language,
                             path=full_path,
                         )
--- a/lib/spack/spack/compilers/__init__.py
+++ b/lib/spack/spack/compilers/__init__.py
@@ -27,6 +27,7 @@ import spack.spec
 from spack.util.environment import get_path
 from spack.util.naming import mod_to_class
 
+win_exe_ext = ".exe"
 _path_instance_vars = ["cc", "cxx", "f77", "fc"]
 _flags_instance_vars = ["cflags", "cppflags", "cxxflags", "fflags"]
 _other_instance_vars = [
@@ -722,7 +723,9 @@ def make_compiler_list(detected_versions):
         compiler_cls = spack.compilers.class_for_compiler_name(compiler_name)
         spec = spack.spec.CompilerSpec(compiler_cls.name, version)
         paths = [paths.get(x, None) for x in ("cc", "cxx", "f77", "fc")]
-        target = archspec.cpu.host()
+        # in order for tests to function, this needs to refer to current host
+        # as determined by Spack, not archspec
+        target = archspec.cpu.TARGETS[str(spack.platforms.host().target("default_target"))]
         compiler = compiler_cls(spec, operating_system, str(target.family), paths)
         return [compiler]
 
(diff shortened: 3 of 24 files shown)


alalazo:
`unit-test-cmd` in CI is failing, but it's unclear to me why the script doesn't stop at the failing command rather than proceeding to the next one. Note that without at least `coverage xml` I think we'll lose codecov reporting.

johnwparent:
@alalazo I intend to return the xml related commands, this is more of a diagnostic at this point. ~~I believe is has something to do with the~~


~~defaults:
  run:
    shell:
     powershell Invoke-Expression -Command ".\share\spack\qa\windows_test_setup.ps1"; {0}~~

~~stanzas.~~

My initial intuition was incorrect. See comment below and comments on linked issue for more details on actual cause/solution.

johnwparent:
Update: this is actually not caused by the above lines (although they will be refactored because that script is actually redundant now), this is just how powershell behaves w.r.t. to script execution. So after the unit-tests fail, powershell executes the cov lines, and because those "succeed" then $LASTEXITCODE is set to 0 and the job passes. Couple solutions here. Will experiment to determine the most effective/cleanest workflow file.

alalazo:
> So after the unit-tests fail, powershell executes the cov lines, and because those "succeed" then $LASTEXITCODE is set to 0 and the job passes. Couple solutions here. Will experiment to determine the most effective/cleanest workflow file.

I left in https://github.com/spack/spack/issues/33347#issuecomment-1284544698 a link to the default configuration for powershell in GithubActions. It seems they also prepend `$ErrorActionPreference = 'stop'` to the script, which as far as I understand we are not doing?

lib/spack/llnl/util/filesystem.py
```diff
@@ -2367,6 +2372,21 @@ def files_in(*search_paths):
     return files
 
 
+@system_path_filter
+def strip_extensions(*paths):
+    """Returns all paths passed as arguments with file extensions removed
+    If a file has no extension it is not modified. This method is predominately
+    for use
```
scheibelp: Incomplete docstring.

Also, I don't see this function being used anywhere; is it necessary for this PR?

lib/spack/spack/compiler.py
```diff
@@ -592,7 +595,16 @@ def search_regexps(cls, language):
         # defined for the compiler
         compiler_names = getattr(cls, "{0}_names".format(language))
         prefixes = [""] + cls.prefixes
-        suffixes = [""] + cls.suffixes
+        suffixes = [""]
+        # Windows compilers generally have an extension of some sort
+        # as do most files on Windows, handle that case here
+        if is_windows:
+            ext = r"\.[a-zA-Z]+"
```
scheibelp: Would the extension generally be "exe'? I think it would be good to limit this if possible.

Presumably this is because you expect for example `gcc` to be called `gcc.exe` on Windows - did I get that right?

lib/spack/spack/compilers/__init__.py
```diff
@@ -722,7 +722,9 @@ def _default_make_compilers(cmp_id, paths):
         compiler_cls = spack.compilers.class_for_compiler_name(compiler_name)
         spec = spack.spec.CompilerSpec(compiler_cls.name, version)
         paths = [paths.get(x, None) for x in ("cc", "cxx", "f77", "fc")]
-        target = archspec.cpu.host()
+        # in order for tests to function, this needs to refer to current host
+        # as determined by Spack, not archspec
+        target = archspec.cpu.TARGETS[str(spack.platforms.host().target("default_target"))]
```
scheibelp: If this is only needed for tests, then we should probably add a context manager/fixture to overwrite this rather than changing the behavior everywhere (i.e. inside and outside of the tests). What goes wrong when this isn't set?

lib/spack/spack/test/concretize_preferences.py
```diff
@@ -270,7 +270,9 @@ def test_external_mpi(self):
         # ensure that once config is in place, external is used
         spec = Spec("mpi")
         spec.concretize()
-        assert spec["mpich"].external_path == os.sep + os.path.join("dummy", "path")
+        # This is a hardcoded value from the conf above so correct
+        # platform specific paths do not need to be used here
+        assert spec["mpich"].external_path == os.path.sep + os.path.join("dummy", "path")
```
scheibelp, line 273: This check is confusing WRT the comment that was added:

* `os.path.sep` == `os.sep` AFAIK (so this test hasn't changed)
* The comment says platform specific paths do not need to be accounted for, but it looks like that's what this is doing by using `os.path.join` (if it were fine to ignore, I assume it would be using something like `posixpath`).

lib/spack/spack/test/database.py
```diff
@@ -719,13 +719,13 @@ def test_external_entries_in_db(mutable_database):
     assert not rec.spec.external_modules
 
     rec = mutable_database.get_record("externaltool")
-    assert rec.spec.external_path == os.sep + os.path.join("path", "to", "external_tool")
+    assert rec.spec.external_path == os.path.sep + os.path.join("path", "to", "external_tool")
```
scheibelp, line 722: This change is fine but AFAIK `os.sep` == `os.path.sep` so this is just for clarity right?

var/spack/repos/builtin/packages/netlib-lapack/package.py
```diff
@@ -108,6 +108,7 @@ class NetlibLapack(CMakePackage):
     depends_on("netlib-xblas+fortran+plain_blas", when="+xblas")
     depends_on("python@2.7:", type="test")
 
+    build_system("cmake")
```
scheibelp: This shouldn't be required: packages should only have to define this if they have more than one possible build system.

lib/spack/spack/util/path.py
```diff
@@ -122,7 +122,7 @@ def path_to_os_path(*pths):
     """
     ret_pths = []
     for pth in pths:
-        if type(pth) is str and not is_path_url(pth):
+        if str in type(pth).__mro__ and not is_path_url(pth):
```
scheibelp: Would `isinstance` work as well? I think that would be more straightforward if so.

lib/spack/spack/test/builder.py
```diff
@@ -123,6 +124,9 @@ def test_old_style_compatibility_with_super(spec_str, method_name, expected):
     assert value == expected
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32", reason="log_output to file conflicts with pytest on Windows"
+)
```
scheibelp, line 130: Should this have a TODO comment to address (I'm fine skipping for now)

Also I want to make sure this is a log_output issue, since the checks appear to be on setting env variables and not dependent on reading output.

scheibelp (changes requested):

lib/spack/spack/compilers/__init__.py
johnwparent: It _should_ only be an error when Spack is overriding platform detection, however, this prevents errors from arising if Spack and archspec were to ever disagree about a current platform. This ensures internal consistency within Spack. That way, if there's an issue, it's immediately obvious from the platform detection, not a "no compiler for spec" issue, which can be hard to diagnose, and even harder to realize archspec is setting the compiler target to a different value. In short, it's not currently an issue outside of the tests, but I think this is important to a large benefit/future proofing.

lib/spack/spack/test/concretize_preferences.py
johnwparent: Oh I meant to remove the comment. That was from a previous (and incorrect) fix, that did fix this test, but did not fix the underlying problem.

lib/spack/spack/test/database.py
johnwparent: Yes.
Edit (for clarity): This is a clarity driven change, both because `os.pathsep` and `os.path.sep` and `os.sep` are all frequently confused and because it's a bit clearer and consistent with the rest of the useages of `os.path.<method>`

var/spack/repos/builtin/packages/netlib-lapack/package.py
johnwparent: I encountered a CI error without this during the extended build test. This change can be reverted as we're no longer testing that package, but regardless, not including this causes an error. If that's not correct behavior, then that's an issue. If you look at the CI run for the commit where I added this, you can see that the package install complained about not having a build system variant for package netlib-lapack

lib/spack/spack/util/path.py
johnwparent: I believe it would, I forgot that check handled sublcasses, I'll change that.

lib/spack/spack/test/builder.py
johnwparent: It does not depend on the log_output, but non the less that method is invoked over the course of this test as a context manager, which causes the stdout failure we discussed.

lib/spack/spack/compiler.py
johnwparent: I wanted to leave things open ended as a compiler may have a .bat file or something as it's driver script. I can't imagine it's common (and I've never seen it although I only used MSVC and LLVM-clang on Windows) but I didn't want to preclude detection because of a certain extension. It also allows for the tests to create .bat scripts much the same way they mock compilers with extensionless scripts on Unix. I could have constrained this to just .bat and .exe, but as I said, I like to err on the side of not accidently leaving things out. If you disagree I have no problem with that, and I'll happily change things.

lib/spack/llnl/util/filesystem.py
johnwparent: Oh, yeah, I forgot I took out the code that used this. No, nothing needs this method, I can remove it. It's sort of useful, but if you want it gone, it's gone.

lib/spack/spack/test/database.py
johnwparent: Honestly, I'm kinda shocked that `os.path.sep` and `os.pathsep` were allowed to be two different conceptual concepts by the python folks.

lib/spack/llnl/util/filesystem.py
scheibelp: I generally prefer not introducing utility methods if they are not used in the same PR (so my preference is to remove it)

lib/spack/spack/compiler.py
scheibelp: I think allowing *any* extension may be pulling in too many things (e.g. `.png`), but we could test for `(exe|bat)` for example to cover those). Technically the discovery logic accounts for failing to run something that isn't runnable, but I still think it's worth not trying to run things when it wouldn't make sense.

var/spack/repos/builtin/packages/netlib-lapack/package.py
scheibelp: Yes this does sound like a problem: ideally we could resolve it in a separate PR.

lib/spack/spack/test/builder.py
scheibelp: Oh I see, we never actually get to the assert. In general this must be working (e.g. for install tests that perform this in a subprocess) so I guess there's an issue with using `log_output` in the current process on Windows; if that sounds right to you, could you refine the `reason` accordingly?

lib/spack/spack/compilers/__init__.py
scheibelp: We do access this in the same way that you added elsewhere in the file. However, I worry this could exclude microarchitecture information from the compiler definitions. I have to think about whether that's actually a problem and if so how to resolve that alongside the consistency issue (I do agree there is an inconsistency here)

scheibelp: Additional notes:

* This was already generalizing by taking `target.family` below, so there is a generalization
* However, `default_target` is defined on a per-platform basis and can differ from `cpu.host()` outside of testing environments: on Cray, it is the backend target (whereas `cpu.host()` is always the frontend target); this might create a problem if you then index `archspec.cpu.TARGETS`
  * Sub-note: on the Cray system I just checked, these match, but the point is they don't have to (either for Cray, or for any other Platform defined in the future)
* Given the second point, it may be suitable not to index `archspec.cpu.TARGETS`, and to just set `target = str(spack.platforms.host().target("default_target"))`. These are different objects, but we refer to compilers using both `spack.target.Target` and `archspec.cpu.microarchitecture.Microarchitecture` so this seems OK to do

@alalazo as the person who wrote this line originally, does that seem workable to you to make this code change? Specifically replace

```
target = archspec.cpu.host()
```

with

```
target = spack.platforms.host().target("default_target")
```

? (this would also require discarding the reference to `family` below, so in fact this can be more specific)

Given the above notes I think it does.

lib/spack/spack/test/builder.py
johnwparent: Actually Markus uncovered an instance of this in the installer tests yesterday, so it does seem like it hits there too (albeit inconsistently). But I'll have the reason updated.

lib/spack/spack/compilers/__init__.py
alalazo: I don't think we should change the granularity of the target for compilers. As weird as it is, it always meant "the architecture family for which this compiler can emit code for". Two considerations:
1. If we want to refactor this part, it should probably be on an overall re-design that takes cross-compilation into account
2. If we change the granularity concretization logic is affected, and needs to change too

That said, what is the specific problem we are trying to solve? Reading the comment it seems to me this change is only needed to test the code?

johnwparent: Right, at the moment, Spack's mocked platform for testing sets an `x86_64` arch, while actually running the tests, archspec picks up the actual system arch, which for me at least is AMD64 (I've had a few others test the code to similar effect on Windows). This results it a number of tests failing with "no compiler for arch" errors. 
In general, I'm surprised this doesn't raise an issue on other platforms that the test suite runs on, so I must be missing something here. I was also a bit confused why there's a potential point for disparities between what Spack understands the platform to be, and what we're determining compilers can target. 

My intent was not to change granularity (maybe a side effect of my unfamiliarity with the archspec module at large) but to shore up a potential vector for internal inconsistency within Spack, and I just see the failing tests as a presentation of that issue, rather than the only place it will ever happen/the explicit cause.

alalazo: > In general, I'm surprised this doesn't raise an issue on other platforms that the test suite runs on, so I must be missing something here. 

It might be a Windows peculiarity to return `AMD64` instead of `x86_64`. For other platforms like macOS we have some basic translation going on in `archspec`. See e.g. https://github.com/archspec/archspec/blob/ed5340947574b1174f70b63edb2848da69e9993f/archspec/cpu/detect.py#L103-L106 Did we consider porting `archspec` to Windows? Would that be a good way forward if it doesn't require a lot of work?

Also, for tests we can `monkeypatch` the `archspec.cpu` module to our liking as a fast solution.

lib/spack/spack/compiler.py
```diff
@@ -592,7 +595,16 @@ def search_regexps(cls, language):
         # defined for the compiler
         compiler_names = getattr(cls, "{0}_names".format(language))
         prefixes = [""] + cls.prefixes
-        suffixes = [""] + cls.suffixes
+        suffixes = [""]
+        # Windows compilers generally have an extension of some sort
+        # as do most files on Windows, handle that case here
+        if is_windows:
+            ext = r"\.(exe|bat)+"
```
scheibelp: What is the purpose of `+` in this regex? My interpretation is that it would match `batbat` or `exebatexeexe` but that doesn't seem useful.

lib/spack/spack/compilers/__init__.py
```diff
@@ -633,7 +633,7 @@ def _default(search_paths):
                         compiler_id = CompilerID(operating_system, compiler_name, None)
                         detect_version_args = DetectVersionArgs(
                             id=compiler_id,
-                            variation=NameVariation(*match.groups()),
+                            variation=NameVariation(*match.groups()[:2]),
```
scheibelp: Why limit this?

share/spack/qa/windows_test_setup.ps1
```diff
@@ -1,11 +1,4 @@
...
-cmd /c "`"spack\bin\spack_cmd.bat`" print " |
-foreach {
-    if ($_ -match "=") {
-        $v = $_.split("=")
-        [Environment]::SetEnvironmentVariable($v[0], $v[1])
-    }
-}
+$ErrorActionPreference = "SilentlyContinue"
+Write-Output F|xcopy .\share\spack\qa\configuration\windows_config.yaml $env:USERPROFILE\.spack\windows\config.yaml
+(Get-Item '.\lib\spack\docs\_spack_root').Delete()
```
scheibelp, line 4: (question) I assume there is some issue with this being a symlink back into Spack is it just that it doesn't work on Windows? Or is the cyclic behavior causing problems? A comment might be useful.

scheibelp (changes requested):
I have a few questions/hopefully minor requests

lib/spack/spack/compiler.py
johnwparent: Cruft, It's effectively a noop unless someone truly wants to see what they can get away with in terms of extension names. I'll drop it out.

lib/spack/spack/compilers/__init__.py
johnwparent: Because before I added the regex pointed out in the comment above this one, the `ext = r"\.(exe|bat)"` there were only two groups being captured, and these were relevant to the construction of that named tuple. This regex adds a new group that we don't care about, so we drop it off. Not doing this causes any code that crosses that line to fail as the group unpacking provides too many positional arguments to the named tuple constructor.

scheibelp: I think I get it: I think this should either:

* add a comment here explaining this
* Use the non-capturing group syntax `(?:.....)` described at https://docs.python.org/3/library/re.html when looking for `bat/exe`

(I prefer the latter)

share/spack/qa/windows_test_setup.ps1
johnwparent: I don't think I actually added this line in this PR, but it doesn't seen to exist in develop. I truly can't remember adding it and I made this PR like a month ago, so it's still pretty fresh. When I check the blame locally it says I committed it back in January... IIRC this link causes havoc with the installer and a few other operations on Windows. The problem is the cyclic behavior, I don't think CMake handles it well at the moment. I'll add a comment.

lib/spack/spack/compilers/__init__.py
johnwparent: I also prefer the latter, forgot it was part of the re syntax. I'll amend.

scheibelp (dismissed):

lib/spack/spack/compilers/__init__.py
alalazo: > I was also a bit confused why there's a potential point for disparities between what Spack understands the platform to be, and what we're determining compilers can target.

A compiler that has support for `x86_64` architectures generally can optimize for every uarch in that family which is "as recent as its release" (roughly). We don't need a compiler for each uarch in the `x86_64` family, so imo the inconsistency is "just" in using the same name for two different concepts.

scheibelp (approved):

lib/spack/spack/test/conftest.py
```diff
@@ -259,6 +259,17 @@ def _verify_executables_noop(*args):
     return None
 
 
+def _host():
+    """Mock archspec host so there is no inconsistency on the Windows platform
+    This function cannot be local as it needs to be pickleable"""
+    return archspec.cpu.Microarchitecture("x86_64", [], "generic", [], {}, 0)
```
becker33, line 265: This line causes test failures on M1

@scheibelp @johnwparent you may be interested in this.

johnwparent: @scheibelp I think this means it's probably time to revisit the whole "Spack has multiple entry points for platform/arch detection and they conflict at test time" issue that was raised [here](https://github.com/spack/spack/pull/33385#discussion_r1040036318)

johnwparent: Thanks for the ping @becker33, feel free to tag my/reach out to iterate on a solution for your issue. Do you have a link to a CI run with this failure?

PR #33386, opened 2022-10-17
[WIP] Make sure external python extensions depend on python

scheibelp:
Since https://github.com/spack/spack/pull/32970, external python extensions will often fail since some of their methods now refer to the `python` dependency, and externals are stripped of all dependencies.

This proposes adding a step which associates python extensions with the external python. It is assumed there is only one instance of `python` in this case. There are a number of scenarios where this may go wrong, but it is a simple solution that is expected to apply in most cases.

### DIFF
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -2961,6 +2961,21 @@ class Spec(object):
         else:
             self._old_concretize(tests)
 
+        self.link_externals_extensions_to_extendees()
+
+    def link_externals_extensions_to_extendees(self):
+        import spack.build_systems.python
+
+        python_spec = None
+        for spec in self.traverse():
+            if spec.name == "python":
+                python_spec = spec
+                break
+
+        for spec in self.traverse():
+            if isinstance(spec, spack.build_systems.python.PythonPackage):
+                spec.add_dependency_edge(python_spec, ("build", "link", "run"))
+
     def _mark_root_concrete(self, value=True):
         """Mark just this spec (not dependencies) concrete."""
         if (not value) and self.concrete and self.installed:


scheibelp:
Note: this only works in cases where a root spec contains both `python` and the `py-...` extension: there are packages which e.g. just declare a dependency on `py-numpy`, and in that case if `py-numpy` is external this would not work. Options for updating this are:

* Users need to create their own manifest files to record this dependency (in other words, don't use external configuration in `packages.yaml` for this): presumably this would require some extra work to tell users how to do it, and it would slow down sharing of environments (since the config in `spack.yaml` would not be sufficient in cases where system Python extensions are used)
* Be more-careful to construct a `python` spec and connect it to `PythonPackage`s (in particular this also needs to extend to the other concretization methods (e.g. `concretize_together` in `concretize.py`)
* Restore the ability in `PythonPackage` to access `.include`/`.platlib` without depending on `python` (I have to read more to understand the initial reasons for #32970)

white238:
I was talking with @becker33 due to me running into this. We tried a few things, one being this PR.

This does not fix a case in axom where we ask for the prefix of an external python package dependency ("py-shroud").

https://github.com/spack/spack/blob/05fd39477e723b051f9b830d3c39ac0a74f9234d/var/spack/repos/builtin/packages/axom/package.py#L467-L469

I get the following debug output:

```
==> [2022-11-08-17:16:37.421886] Error: KeyError: 'No spec with name python in py-shroud@0.12.2%clang@10.0.1.1 ldlibs="-lgfortran"  build_system=python_pip arch=linux-rhel7-power9le'

/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/build_systems/python.py:241, in libs:
        238        """Discover libraries in platlib."""
        239
        240        # Remove py- prefix in package name
  >>    241        library = "lib" + self.spec.name[3:].replace("-", "?")
        242        root = self.prefix.join(self.spec["python"].package.platlib)
        243
        244        for shared in [True, False]:


  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/spec.py", line 3982, in __getitem__
    (x for x in order() if (not x.virtual) and x.package.provides(name)),
StopIteration

  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/build_environment.py", line 1068, in _setup_pkg_and_run
    pkg, dirty=kwargs.get("dirty", False), context=context
  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/build_environment.py", line 792, in setup_package
    set_wrapper_variables(pkg, env_mods)
  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/build_environment.py", line 487, in set_wrapper_variables
    update_compiler_args_for_dep(dspec)
  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/build_environment.py", line 467, in update_compiler_args_for_dep
    dep_link_dirs.extend(query.libs.directories)
  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/spec.py", line 1177, in __get__
    value = f()
  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/spec.py", line 1166, in <lambda>
    callbacks_chain.append(lambda: getattr(pkg, self.attribute_name))
  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/build_systems/python.py", line 242, in libs
    root = self.prefix.join(self.spec["python"].package.platlib)
  File "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_08_15_41_19/spack/lib/spack/spack/spec.py", line 3986, in __getitem__
    raise KeyError("No spec with name %s in %s" % (name, self))
KeyError: 'No spec with name python in py-shroud@0.12.2%clang@10.0.1.1 ldlibs="-lgfortran"  build_system=python_pip arch=linux-rhel7-power9le'

```

PR #33387, opened 2022-10-18
Feature: alter BuildTask priority when no uninstalled dependencies

tldahlgren:
At @haampie 's suggestion, this PR adds some randomization to the build task priority for packages to help reduce bottlenecks when running parallel package builds.  This PR changes the priority for packages with no uninstalled dependencies to a random number in [0,1) to allow multiple `spack install`s a better chance of interleaving the builds.

Preliminary, small-scale timing on an LLNL/LC system indicates there could be a small average elapsed time speedup.

TODO:

- [x] Run/update unit tests

### DIFF
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -2202,6 +2202,12 @@ class BuildTask(object):
 
         # Ensure key sequence-related properties are updated accordingly.
         self.attempts = 0
+
+        # Random number in [0, 1) to allow parallel installs a chance of more
+        # quickly finding the next package to install.
+        # (See the discussion in PR #33387.)
+        self.random = random.random()
+
         self._update()
 
     def __eq__(self, other):
@@ -2304,8 +2310,8 @@ class BuildTask(object):
 
     @property
     def key(self):
-        """The key is the tuple (# uninstalled dependencies, sequence)."""
-        return (self.priority, self.sequence)
+        """The key is the tuple (# uninstalled dependencies, randomizer, sequence)."""
+        return (self.priority, self.randomizer, self.sequence)
 
     def next_attempt(self, installed):
         """Create a new, updated task for the next installation attempt."""
@@ -2318,13 +2324,14 @@ class BuildTask(object):
     @property
     def priority(self):
         """The priority is based on the remaining uninstalled dependencies."""
-        priority = len(self.uninstalled_deps)
+        return len(self.uninstalled_deps)
 
-        # If there are no uninstalled dependencies then the priority should
-        # be a random number in [0, 1) to allow parallel installs a chance
-        # of more quickly finding the next package to install; otherwise,
-        # the priority will remain the number of uninstalled dependencies.
-        return random.random() if priority == 0 else priority
+    @property
+    def randomizer(self):
+        """The install-ready randomizer, which kicks in when the build task
+        is ready to be installed, and is intended to potentially help speed up
+        finding the next package to install."""
+        return self.attempts + self.random if install_priority(self.priority) else 0
 
 
 class BuildRequest(object):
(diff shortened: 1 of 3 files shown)


haampie:
Nice :upside_down_face: I didn't consider the re-queuing logic when I suggested the idea, hopefully that's not too involved, since right now it's nice and simple.

Edit: ah, you probably take `#retries` into account in the comparator when prio < 1

alalazo (dismissed):

lib/spack/spack/installer.py
```diff
@@ -2296,7 +2301,13 @@ def next_attempt(self, installed):
...
     def priority(self):
         """The priority is based on the remaining uninstalled dependencies."""
-        return len(self.uninstalled_deps)
+        priority = len(self.uninstalled_deps)
+
+        # If there are no uninstalled dependencies then the priority should
+        # be a random number in [0, 1) to allow parallel installs a chance
+        # of more quickly finding the next package to install; otherwise,
+        # the priority will remain the number of uninstalled dependencies.
+        return random.random() if priority == 0 else priority
```
haampie: Wouldn't it be better to generate the random number once ahead of time in the constructor or so, and order by `(#retries, self.random_number)`? Suppose you have 2 tasks and one is busy, then if you randomly select the busy task, it would make sense to always try the other task next, instead of potentially selecting the busy task by chance multiple times. Generating the random number ahead of time also ensures it's stable, since right now it just works because the key is retrieved once before pushing to the prio queue; if the priority queue dynamically accessed this property each time it had to compare priorities, it would break if it wasn't constant.

tldahlgren: Well, something like that.  The key throws a monkey wrench into the works.  So I tried a variant of this approach that I currently think will have the same affect.

lib/spack/spack/installer.py
```diff
@@ -262,6 +263,10 @@ def _hms(seconds):
     return " ".join(parts)
 
 
+def install_priority(priority):
```
haampie: This can also be reverted since no random number is added to priority anymore.

lib/spack/spack/installer.py
```diff
@@ -2311,8 +2322,8 @@ def cache_only(self):
 
     @property
     def key(self):
-        """The key is the tuple (# uninstalled dependencies, sequence)."""
-        return (self.priority, self.sequence)
+        """The key is the tuple (# uninstalled dependencies, randomizer, sequence)."""
+        return (self.priority, self.randomizer, self.sequence)
```
haampie:
```suggestion
        return (self.priority, self.attempts, self.random)
```

There's no need to special-case `self.priority == 0`, personally I find ^ easier to follow.

I don't think `self.sequence` is relevant anymore, it would tie-break two tasks deterministically in the unlikely case their random numbers are equal, there's no point in doing that deterministically. So, `_counter` and `self.sequence = next(_counter)` elsewhere can be dropped.

haampie (changes requested):
When keeping `self.priority = # uninstalled deps` as an integer, there's a couple of things to revert.

lib/spack/spack/installer.py
tldahlgren: For a reason that escapes me ATM, it was important in the original design that build tasks with the same priority retain the order in which they were added.  My changes here were intended to reflect that goal.

Perhaps @tgamblin @becker33 @alalazo @scheibelp remember why that was done.

haampie: That will continue to be the case if you order by `(priority, attemps, random)` since the random number is constant.

alalazo:
@tldahlgren Is this PR outdated by the recent work on the installer? Should we close, or rebase, it?

tldahlgren:
> @tldahlgren Is this PR outdated by the recent work on the installer? Should we close, or rebase, it?

I will revisit this.

tldahlgren:
Closing since dated, especially after the new parallel builds work.

PR #33388, opened 2022-10-18
Reusable --use-buildcache with better validation

haampie:
Use `type=arguments.use_buildcache` in argparse to get better errors, and also
show what the error was for invalid input.

Move the validation logic to `arguments` so it's reusable.

### DIFF
--- a/lib/spack/spack/cmd/common/arguments.py
+++ b/lib/spack/spack/cmd/common/arguments.py
@@ -6,6 +6,8 @@
 
 import argparse
 
+from llnl.util.lang import stable_partition
+
 import spack.cmd
 import spack.config
 import spack.dependency as dep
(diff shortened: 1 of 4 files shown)


lib/spack/spack/cmd/common/arguments.py
```diff
@@ -437,3 +439,46 @@ def add_s3_connection_args(subparser, add_help):
     subparser.add_argument(
         "--s3-endpoint-url", help="Endpoint URL to use to connect to this S3 mirror"
     )
+
+
+def use_buildcache(opt):
```
alalazo: Can we rename it to something like:
```suggestion
def use_buildcache(cli_args):
```
?

lib/spack/spack/cmd/common/arguments.py
```diff
@@ -437,3 +439,46 @@ def add_s3_connection_args(subparser, add_help):
...
+    """Argument type that accepts comma-separated subargs:
+    1. auto|only|never
+    2. package:auto|only|never
+    3. dependencies:auto|only|never"""
```
alalazo: Can we improve a bit the docstring? Something like (feel free to reword as you like though):
```suggestion
    """Translate buildcache related command line arguments into a pair of strings,
    representing whether the root or its dependencies can use buildcaches.
        
    Argument type that accepts comma-separated subargs:
    
        1. auto|only|never
        2. package:auto|only|never
        3. dependencies:auto|only|never
        
    Args:
        cli_args (str): command line argument to be translated
    
    Return:
        Tuple of strings ...
    """
```

lib/spack/spack/cmd/common/arguments.py
```diff
@@ -437,3 +439,46 @@ def add_s3_connection_args(subparser, add_help):
...
+    valid_keys = frozenset(["package", "dependencies"])
+    valid_values = frozenset(["only", "never", "auto"])
```
alalazo, line 461: (comment) Wondering if we should use enum-like class attributes instead of strings :thinking:

alalazo (changes requested):
Only minor comments, basically LGTM

alalazo (approved):

PR #33389, opened 2022-10-18
intel-oneapi-compilers: do not pass -Wno-unused-command-line-argument to icc + refactor

rscohn2:
-Wno-unused-command-line-argument was needed for older versions of icx when used with -Wl
Do not pass it to icc, which does not recognized it.

Some refactoring to make it easier to have flags specific to classic/oneapi.

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
@@ -182,6 +182,13 @@ class IntelOneapiCompilers(IntelOneApiPackage):
                 # should not be patched
                 patchelf(file, fail_on_error=False)
 
+    def write_config_file(self, flags, path, compilers):
+        for compiler in compilers:
+            p = path.join(compiler + ".cfg")
+            with open(p, "w") as f:
+                f.write(" ".join(flags))
+            set_install_permissions(p)
+
     @run_after("install")
     def extend_config_flags(self):
         # Extends compiler config files to inject additional compiler flags.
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
```diff
@@ -196,41 +203,28 @@ def extend_config_flags(self):
         # TODO: it is unclear whether we should really use all elements of
         #  _ld_library_path because it looks like the only rpath that needs to be
         #  injected is self.component_prefix.linux.compiler.lib.intel64_lin.
-        flags_list = ["-Wl,-rpath,{}".format(d) for d in self._ld_library_path()]
+        common_flags = ["-Wl,-rpath,{}".format(d) for d in self._ld_library_path()]
+
+        # Make sure that underlying clang gets the right GCC toolchain by default
+        llvm_flags = ["--gcc-toolchain={}".format(self.compiler.prefix)]
+        classic_flags = ["-gcc-name={}".format(self.compiler.cc)]
```
haampie, line 210: There's this annoying thing that:
1. `self.compiler.cc` can simply be say `gcc`, it's not guaranteed to be an absolute path -- and a module loaded with the compiler ensures PATH is set correctly.
2. Some environments such as Cray PE have a single compiler wrapper binary you're supposed to use for all compilers on the system, so that's what self.compiler.cc points to; environment variables make it dispatch to a particular compiler.

But I guess we can ignore that...

haampie (approved):

var/spack/repos/builtin/packages/intel-oneapi-compilers/package.py
rscohn2: The classic compilers execute gcc and query it for the paths so if `self.compiler.cc`` can be exec'ed then that will work. I am not sure about llvm.

haampie: Yup, so on a module-based system, users need to specify `modules: [gcc/x]` in Spack's `compilers.yaml` for the relevant intel compiler. (which can be confusing, but well...)

Regarding LLVM: we do what classic compilers do, but only once and for all during the installation.

haampie:
@spackbot run pipeline

PR #33391, opened 2022-10-18
Use `read_transaction` in installer code paths

haampie:
There are some bulk install status checks in environment.py and
installer.py. To reduce the number of locks required, wrap those in a
transaction.

Hopefully this reduces the risk of problems found in #31387.

In an environment with `pika`, Spack took 154 read locks on the database until
the first install job, with this PR that reduces to 9 read locks.

Initially I was thinking to do replace the install status checks in the
installer with a single bulk query to minimize time spent holding
a read lock, but the install status checks are scattered over the code
base, so I kept it simple.

(I don't immediately know of a slow filesystem to test this on ;p)


haampie:
@spackbot run pipeline

haampie:
@tldahlgren do you think this is a good idea?

tldahlgren:
> @tldahlgren do you think this is a good idea?

Sounds promising.  Let me look at it more closely and test it.

tldahlgren:
> > @tldahlgren do you think this is a good idea?
> 
> Sounds promising. Let me look at it more closely and test it.

FYI. Using a scaled down version of the example you provided in the references issue, I appear to consistently get the error (e.g., six `Error: Timed out waiting for write lock.`  errors when running with `-j8`) with the installations failing.

haampie:
Curiously, I can't reproduce it...

PR #33393, opened 2022-10-18
ECP-SDK: enable hdf5 VOL adapters

svenevs:
- When +hdf5, enable VOL adapters suitable for the SDK.
- Each VOL package must prepend to the HDF5_PLUGIN_PATH.
- hdf5: 1.13.3 will break existing VOL packages, constrain
  VOLs related to SDK and add note to keep 1.13.2 available.
- hdf5-vol-async:
    - Do not set HDF5_VOL_CONNECTOR, consumers must opt-in.
    - Enforce DAG constraints on MPI to require threaded version.
    - Depend on an explicit version of argbots to relax
      concretization issues in other spack environments.
- paraview: fix compiler flag usage for the 110 ABI (followup to #33617).
- [ ] Confirm builds.
    - [X] e4s container build
    - [X] vanilla build bare metal
    - [ ] e4s site build incl. site yamls
    - [ ] gitlab dav-sdk pipeline keeps failing in paraview, cannot reproduce locally
- [X] Confirm with HDF5 group about vol-async environment variables.

```console
root@79bbd49cf2a8:/# spack load ecp-data-vis-sdk                                                                                                        
root@79bbd49cf2a8:/# echo $HDF5_PLUGIN_PATH | tr : $'\n' | xargs ls                                                                                     
/build/src/opt/spack/linux-ubuntu20.04-zen2/gcc-11.1.0/hdf5-vol-async-1.3-qdnz5mqjb2puxdcwoy7muk63udmw7uwz/lib:
libasynchdf5.a  libh5async.so

/build/src/opt/spack/linux-ubuntu20.04-zen2/gcc-11.1.0/hdf5-vol-cache-v1.0-oevenlhfpnlhbnykhwudmpovgg4rihtu/lib:
libhdf5_vol_cache.so

/build/src/opt/spack/linux-ubuntu20.04-zen2/gcc-11.1.0/hdf5-vol-log-1.3.0-2dl7s3whgwz6hs2ftzrolzfjd54me4hs/lib:
libH5VL_log.a  libH5VL_log.so  libH5VL_log.so.0  libH5VL_log.so.0.0.0
```

### DIFF
--- a/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
+++ b/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
@@ -125,6 +125,9 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
         )
     conflicts("~cuda", when="^hdf5-vfd-gds@1.0.2:")
     conflicts("~hdf5", when="^hdf5-vfd-gds@1.0.2:")
+    dav_sdk_depends_on("hdf5-vol-async", when="+hdf5")
+    dav_sdk_depends_on("hdf5-vol-cache", when="+hdf5")
+    dav_sdk_depends_on("hdf5-vol-log", when="+hdf5")
 
     dav_sdk_depends_on("parallel-netcdf+shared", when="+pnetcdf", propagate=["fortran"])
 
--- a/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
+++ b/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
@@ -7,7 +7,9 @@ from spack.package import *
 
 
 class Hdf5VolAsync(CMakePackage):
-    """This package enables asynchronous IO in HDF5."""
+    """This package enables asynchronous IO in HDF5.  WARNING: Please refer to the
+    documentation.  This package will be available in HDF5_PLUGIN_PATH, but in order to
+    use it the consumer must set the HDF5_VOL_CONNECTOR environment variable."""
 
     homepage = "https://hdf5-vol-async.readthedocs.io"
     git = "https://github.com/hpc-io/vol-async.git"
@@ -23,13 +25,15 @@ class Hdf5VolAsync(CMakePackage):
     version("1.0", tag="v1.0")
 
     depends_on("mpi")
-    depends_on("argobots@main")
-    depends_on("hdf5@1.13: +mpi +threadsafe")
+    depends_on("argobots@1.1:")
+    depends_on("hdf5@1.13:1.13.2 +mpi +threadsafe", when="@:1.3")
+
+    # Enforce that MPI_THREAD_MULTIPLE is available.
+    depends_on("openmpi +thread_multiple", when="^openmpi")
+    depends_on("mvapich2 threads=multiple", when="^mvapich2")
 
     def setup_run_environment(self, env):
-        env.set("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
-        vol_connector = "async under_vol=0;under_info=[]"
-        env.set("HDF5_VOL_CONNECTOR", vol_connector)
+        env.prepend_path("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
         env.set("MPICH_MAX_THREAD_SAFETY", "multiple")
 
     def cmake_args(self):
--- a/var/spack/repos/builtin/packages/hdf5-vol-cache/package.py
+++ b/var/spack/repos/builtin/packages/hdf5-vol-cache/package.py
@@ -16,7 +16,11 @@ class Hdf5VolCache(CMakePackage):
     version("default", branch="develop")
     version("v1.0", tag="v1.0")
 
+    depends_on("hdf5@1.13:1.13.2 +mpi +threadsafe", when="@:v1.0")
     depends_on("hdf5-vol-async")
+
+    def setup_run_environment(self, env):
+        env.prepend_path("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
 
     def cmake_args(self):
         """Populate cmake arguments for HDF5 VOL."""
--- a/var/spack/repos/builtin/packages/hdf5-vol-log/package.py
+++ b/var/spack/repos/builtin/packages/hdf5-vol-log/package.py
@@ -21,11 +21,14 @@ class Hdf5VolLog(AutotoolsPackage):
     version("1.2.0", tag="logvol.1.2.0")
     version("1.1.0", tag="logvol.1.1.0")
 
-    depends_on("hdf5@1.13.2:")
+    depends_on("hdf5@1.13:1.13.2", when="@:1.3.0")
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool", type="build")
     depends_on("m4", type="build")
+
+    def setup_run_environment(self, env):
+        env.prepend_path("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
 
     def configure_args(self):
         args = []
--- a/var/spack/repos/builtin/packages/hdf5/package.py
+++ b/var/spack/repos/builtin/packages/hdf5/package.py
@@ -46,6 +46,11 @@ class Hdf5(CMakePackage):
     version("develop-1.8", branch="hdf5_1_8")
 
     # Odd versions are considered experimental releases
+    # NOTE: retain version 1.13.2 until:
+    # 1. hdf5-vol-async, hdf5-vol-cache, and hdf5-vol-log are ready to build off
+    #    hdf5@1.13.3+ (VOL struct underwent breaking changes).
+    # 2. New versions for each package are ready to be tagged and updated in spack.
+    # 4. The ecp-data-vis-sdk maintainers are ready to abandon 1.13.2.
     version("1.13.2", sha256="01643fa5b37dba7be7c4db6bbf3c5d07adf5c1fa17dbfaaa632a279b1b2f06da")
 
     # Even versions are maintenance versions
--- a/var/spack/repos/builtin/packages/paraview/package.py
+++ b/var/spack/repos/builtin/packages/paraview/package.py
@@ -299,7 +299,7 @@ class Paraview(CMakePackage, CudaPackage):
             elif self.spec.satisfies("@5.10: +hdf5"):
                 if self.spec["hdf5"].satisfies("@1.12:"):
                     flags.append("-DH5_USE_110_API")
-        return (flags, None, None)
+        return (None, None, flags)
 
     def setup_run_environment(self, env):
         # paraview 5.5 and later


var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
```diff
@@ -108,6 +108,9 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
     dav_sdk_depends_on("faodel+shared+mpi network=libfabric", when="+faodel", propagate=["hdf5"])
 
     dav_sdk_depends_on("hdf5@1.12: +shared+mpi", when="+hdf5", propagate=["fortran"])
+    dav_sdk_depends_on("hdf5-vol-async", when="+hdf5")
```
svenevs, line 128: @kwryankrattiger `hdf5-vol-async` depends on `mpi`, should I be doing something additionally special here for propagation?  We already depend explicitly on `hdf5 +shared+mpi`.

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
```diff
@@ -108,6 +108,9 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
...
+    dav_sdk_depends_on("hdf5-vol-cache", when="+hdf5")
```
svenevs, line 129: @kwryankrattiger `hdf5-vol-cache` depends on `hdf5-vol-async`, but since the package already depends on this just above it AFAICT we don't need to do anything special.  Correct?

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -27,10 +27,7 @@ class Hdf5VolAsync(CMakePackage):
     depends_on("hdf5@1.13: +mpi +threadsafe")
 
     def setup_run_environment(self, env):
-        env.set("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
-        vol_connector = "async under_vol=0;under_info=[]"
-        env.set("HDF5_VOL_CONNECTOR", vol_connector)
-        env.set("MPICH_MAX_THREAD_SAFETY", "multiple")
+        env.prepend_path("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
```
svenevs, line 36: @lrknox for each of these packages we're changing `env.set` to `env.prepend_path` for `HDF5_PLUGIN_PATH`.  Otherwise, whichever VOL package is loaded last will overwrite the variable (meaning only one VOL would ever be in this variable).

What is not clear is the role of `HDF5_VOL_CONNECTOR`, in theory that could be any value?  This is a slight point of tension, but the motivation for removing this (and `MPICH_MAX_THREAD_SAFETY`) is similar to `set` vs `prepend_path` -- my belief is that the connector needs to be controlled at the application level, this package setting this value makes it so that no other VOL would be used?

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -27,10 +27,7 @@ class Hdf5VolAsync(CMakePackage):
     depends_on("hdf5@1.13: +mpi +threadsafe")
 
     def setup_run_environment(self, env):
-        env.set("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
-        vol_connector = "async under_vol=0;under_info=[]"
```
jeanbez, line 31: Why are we removing this? When applications load the ASYNC-VOL (spack load hdf5-vol-async), they want to use the connector. So shouldn't spack handle that setting instead of having users set it?

jeanbez (changes requested):

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
jeanbez: Not any value, but it could be changed to combine multiple VOL connectors. We can safely assume that if a user loads a connector, they want to use it without much hassle, so those should be set. If they want to override the default behavior by combining multiple connectors, they can still do it.

jeanbez: Also, other packages already rely on and assume this behavior to hold. So changing this might cause things to break.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
svenevs: (marking resolved to keep discussion on the same thread above)

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
svenevs: Thanks @jeanbez!  I think the confusion is that we only have one example of `HDF5_VOL_CONNECTOR`.  Consider if both `hdf5-vol-async` and `hdf5-vol-cache` both needed to set that variable.

> but it could be changed to combine multiple VOL connectors.

So it would look like `HDF5_VOL_CONNECTOR=async under_vol=0;under_info=[]; additional_magic=11; etc=[]`.  It's not clear how to approach this problem correctly, other than that `env.set` is a problem.  For example, if I am a user and I have `HDF5_VOL_CONNECTOR=<my custom value>`, the current spack recipe just obliterates it and chooses its own value.  How to set that correctly in `spack` is not currently well defined (but we could possibly try and e.g., add it in if it already exists?  hard part is removing from environment).

Does the problem make more sense now?  The adios2 vol is the only other one not on this PR, it just prepends to `HDF5_PLUGIN_PATH` but daos is also planned (if you have any knowledge of these / they factor in here).

From my perspective, this package should not be allowed to force the VOL connector on people -- we want to make all the VOLs _available_ in the SDK.  If by installing one which sets `HDF5_VOL_CONNECTOR` that implicitly means it's always used, that isn't OK.  Is that what happens?  If it's not clear, I don't have a good understanding of the overall layout here with this variable so we're trying to figure it out :slightly_smiling_face:

> Also, other packages already rely on and assume this behavior to hold. So changing this might cause things to break.

Any interested parties we can @?  Silent breakages are never fun, but if the current behavior is incorrect (I believe this is the case), we should find the right way to fix them together.

Putting aside whether or not this change goes through, assume we remove it: what happens to the broken package?  Will it be a runtime error?  Or will it just silently not do something it was expected to do?

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
kwryankrattiger: you shouldn't need to do anything for that here.

var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
kwryankrattiger: correct, no additional actions are required.

Everything listed as a `dav_sdk_depends_on` is only saying "this is a package explicitly provided by the SDK".

If packages have co-dependent requirements with other SDK packages, those are handled at the level of the specific packages, not by the SDK. The SDK should not attempt to duplicate requirements better expressed at the package level.

The only time the SDK interferes with inter-dependencies is to propagate SDK variants or to handle higher order/decoupled dependency resolution. An example of decoupled dependency requirements is HDF5 and Ascent. Ascent requires conduit, but using a Conduit with HDF5 >= v1.8 requires `conduit ~hdf5_compat`. The concretizer struggles to make this happen on it's own, so it is specified at the SDK level.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
kwryankrattiger: One thing that may be useful for HDF5 Vol connectors to provide is a script that can be sources to configure the "default" vol connector. This way, users who which to use hdf5-vol-async could load the VOL Async module, and then source a `hdf5-vol-async-default.env`. Likewise for the other vol connectors.

Here is an example that does some possibly useful checks, there may also be more useful details or configurations that specific vol connects would want.

```sh
# hdf5-vol-async-default.env
echo "Loading HDF5 VOL Async..."

# Save current vol connector?
if [[ ! -z $HDF5_VOL_CONNECTOR ]]; then
  echo "Overriding current VOL Connector config: $HDF5_VOL_CONNECTOR"
  echo "Saving in HDF5_VOL_CONNECT_BAK"
  export HDF5_VOL_CONNECTOR_BAK=$HDF5_VOL_CONNECTOR
fi


# Validate that the target vol connector is in the path
IFS=:
for path in $HDF5_PLUGIN_PATH; do
  # Search for VOL Async plugin, if it isn't in the path then report an error to the user?
done

# Setup the default config
export HDF5_VOL_CONNECTOR="async under_vol=0;under_info=[]"
```

In theory, HDF5 could also provide a VOL preamble script that handles backing up `HDF5_VOL_CONNECTOR` and searching for the expected plugin library in the `HDF5_PLUGIN_PATH`.

kwryankrattiger (dismissed):
LGTM

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -6,6 +6,20 @@
...
 
+def require_mpi_thread_multiple(env=None):
+    if env is not None:
+        # Configure this for MPICH and the many MPICH forks
+        env.set("MPICH_MAX_THREAD_SAFETY", "multiple")
+    else:
+        # Don't allow building with MPIs that don't support
+        # MPI_THREAD_MULTIPLE
+        msg = "HDF5 VOL Async requires MPI with MPI_THREAD_MULTIPLE support"
+        conflicts("openmpi ~thread_multiple", msg=msg)
```
svenevs: I was playing around with the concretizer.  It seems you can actually get it to work with `spack spec -I ecp-data-vis-sdk +hdf5 ^openmpi~thread_multiple` because `thread_multiple` variant is not enabled for all versions.  I believe this still does exactly what we need it to.

houjun:
@derobins Any comment on setting the environment variables for VOL connectors in spack?

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -26,11 +40,11 @@ class Hdf5VolAsync(CMakePackage):
     depends_on("argobots@main")
     depends_on("hdf5@1.13: +mpi +threadsafe")
```
svenevs: we are anticipating 1.13.3 to land soon, but the VOLs will break -- we should look into pinning to 1.13.2 for 11/15 deadline

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
svenevs:
```
 PKG-DIRECTIVES: 2 issues found
1. hdf5-vol-async: wrong variant in "conflicts" directive
    the variant 'thread_multiple' does not exist
    in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
2. hdf5-vol-async: wrong variant in "conflicts" directive
    the variant 'threads' does not exist
    in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
Error: Process completed with exit code 1.
```

brtnfld:
I agree that Each VOL package should append to HDF5_PLUGIN_PATH.
As for HDF5_VOL_CONNECTOR, Spack should not be setting this environment variable. Setting this causes an HDF5 application, by default, to use the vol connector, which might not be what the user intended. The default for an HDF5 install should be the native connector unless specified otherwise. It is common practice to set HDF5_VOL_CONNECTOR when using a vol connector since this is how one enables/disables using the vol connectors. Enabling a vol connector should be explicit.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
svenevs: Marking as resolved per https://github.com/spack/spack/pull/33393#issuecomment-1291051774, the spack package will make all VOLs added here (async, cache, log) _available_ via `HDF5_PLUGIN_PATH` and we will remove `HDF5_VOL_CONNECTOR` getting set since this changes the application behavior without user consent.

derobins:
> I agree that Each VOL package should append to HDF5_PLUGIN_PATH. As for HDF5_VOL_CONNECTOR, Spack should not be setting this environment variable. Setting this causes an HDF5 application, by default, to use the vol connector, which might not be what the user intended. The default for an HDF5 install should be the native connector unless specified otherwise. It is common practice to set HDF5_VOL_CONNECTOR when using a vol connector since this is how one enables/disables using the vol connectors. Enabling a vol connector should be explicit.

Agree with Scot - append to HDF5_PLUGIN_PATH (order shouldn't be important) and don't set HDF5_VOL_CONNECTOR.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
derobins: 1.13.3 will be released on October 31 (plus or minus a day or two). The VOL struct will break in a few ways:

- The capabilities flag will be expanded to a 64-bit unsigned integer
- You will need to define flags for the VOL tests to run (we have an expanded set now)
- The dataset I/O calls now take arrays in order to support the multi-dataset feature

We'll have a migration guide that tells people how to update their connectors, but targeting 1.13.2 for your 11/15 deadline is probably the right decision.

svenevs: > targeting 1.13.2 for your 11/15 deadline is probably the right decision.

OK thanks for weighing in, right now the VOLs are constrained to need `1.13:1.13.2` and a `NOTE` in `hdf5` to not delete `1.13.2` until we're ready to update all three (the SDK cannot update say `hdf5-vol-async` which would need `>=1.13.3` until the other two are ready, keeping `1.13.2` available).

In the future the versions associated with `1.13` will no longer be possible to build, my understanding is this was the desired goal for 1.14.

kwryankrattiger (dismissed):

hyoklee (dismissed):

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
derobins: Yes. 1.13.x releases are essentially release candidates for 1.14.0. We just run them through the main release process like a regular release to make it easier for people to keep up with the next major release of HDF5 and give us early feedback.

derobins: You'll still be able to check out those branches and build them, but we will probably pull the binaries off of our website when 1.14.0 releases.

lrknox: Minor difference - you'll be able to check out the tags, not the branches;  tags differ from branch names only by hdf5- instead of hdf5_.  e.g.  tag hdf5-1_13_3 instead  of branch hdf5_1_13_3.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -24,12 +30,14 @@ class Hdf5VolAsync(CMakePackage):
 
     depends_on("mpi")
     depends_on("argobots@main")
```
kwryankrattiger: @lrknox can this be made to `argobots@1.1:`. This is causing me headache somewhere else.

lrknox: I tried argobots@1.1;  hdf5-vol-async built and tested with no problems on a machine here.  I think argobots@1.1 should be fine.  (1.1 is the preferred version for argobots, maybe depends_on("argobots") would also work?)

jeanbez (dismissed):

houjun (dismissed):

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
svenevs: Great, thanks!  It's in.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
svenevs: As long as `version("1.13.2", ...)` remains in `spack` until `ecp-data-vis-sdk` is ready to upgrade (see notes in `hdf5/package.py`) this should be OK.  We will likely be able to do that fairly quickly provided the VOLs here also update and release new versions.

kwryankrattiger (dismissed):

hyoklee (dismissed):

jeanbez (dismissed):

houjun (dismissed):

hyoklee:
@spackbot run pipeline

xdelaruelle:
Maybe this PR could be closed as #35195 has been merged.

PR #33397, opened 2022-10-18
Add Contributor Covenant badge

markcmiller86:
Here is a small update if you want it to include Contributor Covenant code of conduct badge.


markcmiller86:
Whoops...this badge is for version 2.1 of CC and Spack is using version 1.4. Lemme see if its an easy fix.

PR #33400, opened 2022-10-18
Fixup clingo dll lookup on Windows

johnwparent:
Current RPath on Windows support was lacking a proper package side interface to establish appropriate linkage to support Clingo (and other python extension libraries). This PR adds that support, provides an example and use case for RPath on Windows, and applies some much needed updates to RPath support.

Establishes Clingo as default concertizer on Windows

Partial fix to: #33347

### DIFF
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -2095,7 +2095,7 @@ def find_system_libraries(libraries, shared=True):
     return libraries_found
 
 
-def find_libraries(libraries, root, shared=True, recursive=False):
+def find_libraries(libraries, root, shared=True, recursive=False, runtime=True):
     """Returns an iterable of full paths to libraries found in a root dir.
 
     Accepts any glob characters accepted by fnmatch:
@@ -2116,6 +2116,10 @@ def find_libraries(libraries, root, shared=True, recursive=False):
             otherwise for static. Defaults to True.
         recursive (bool): if False search only root folder,
             if True descends top-down from the root. Defaults to False.
+        runtime (bool): Windows only option, no-op elsewhere. If true,
+            search for runtime shared libs (.DLL), otherwise, search
+            for .Lib files. If shared is false, this has no meaning.
+            Defaults to True.
 
     Returns:
         LibraryList: The libraries that have been found
@@ -2130,7 +2134,7 @@ def find_libraries(libraries, root, shared=True, recursive=False):
 
     if is_windows:
         static_ext = "lib"
-        shared_ext = "dll"
+        shared_ext = "dll" if runtime else "lib"
     else:
         # Used on both Linux and macOS
         static_ext = "a"
@@ -2174,13 +2178,13 @@ def find_libraries(libraries, root, shared=True, recursive=False):
     return LibraryList(found_libs)
 
 
-def find_all_shared_libraries(root, recursive=False):
+def find_all_shared_libraries(root, recursive=False, runtime=True):
     """Convenience function that returns the list of all shared libraries found
     in the directory passed as argument.
 
     See documentation for `llnl.util.filesystem.find_libraries` for more information
     """
-    return find_libraries("*", root=root, shared=True, recursive=recursive)
+    return find_libraries("*", root=root, shared=True, recursive=recursive, runtime=runtime)
 
 
 def find_all_static_libraries(root, recursive=False):
@@ -2189,7 +2193,7 @@ def find_all_static_libraries(root, recursive=False):
 
     See documentation for `llnl.util.filesystem.find_libraries` for more information
     """
-    return find_libraries("*", root=root, shared=False, recursive=recursive)
+    return find_libraries("*", root=root, shared=False, recursive=recursive, runtime=False)
 
 
 def find_all_libraries(root, recursive=False):
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -1083,7 +1083,11 @@ def temp_cwd():
         with working_dir(tmp_dir):
             yield tmp_dir
     finally:
-        shutil.rmtree(tmp_dir)
+        kwargs = {}
+        if is_windows:
+            kwargs["ignore_errors"] = False
+            kwargs["onerror"] = readonly_file_handler(ignore_errors=True)
+        shutil.rmtree(tmp_dir, **kwargs)
 
 
 @contextmanager
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -2233,12 +2233,7 @@ class WindowsSimulatedRPath(object):
         """
         Set of directories where package binaries/libraries are located.
         """
-        if hasattr(self.pkg, "libs") and self.pkg.libs:
-            pkg_libs = set(self.pkg.libs.directories)
-        else:
-            pkg_libs = set((self.pkg.prefix.lib, self.pkg.prefix.lib64, self.pkg.prefix.bin))
-
-        return pkg_libs | self._additional_library_dependents
+        return set(self.pkg.prefix.bin) | self._additional_library_dependents
 
     def add_library_dependent(self, *dest):
         """
@@ -2246,7 +2241,7 @@ class WindowsSimulatedRPath(object):
         common paths that need to link against other libraries
 
         Specified paths should fall outside of a package's common
-        link paths, i.e. the lib, lib64, and bin
+        link paths, i.e. the bin
         directories.
         """
         for pth in dest:
@@ -2261,7 +2256,6 @@ class WindowsSimulatedRPath(object):
         Set of libraries this package needs to link against during runtime
         These packages will each be symlinked into the packages lib and binary dir
         """
-
         dependent_libs = []
         for path in self.pkg.rpath:
             dependent_libs.extend(list(find_all_shared_libraries(path, recursive=True)))
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -2134,6 +2134,8 @@ def find_libraries(libraries, root, shared=True, recursive=False, runtime=True):
 
     if is_windows:
         static_ext = "lib"
+        # For linking (runtime=False) you need the .lib files regardless of
+        # whether you are doing a shared or static link
         shared_ext = "dll" if runtime else "lib"
     else:
         # Used on both Linux and macOS
@@ -2193,7 +2195,7 @@ def find_all_static_libraries(root, recursive=False):
 
     See documentation for `llnl.util.filesystem.find_libraries` for more information
     """
-    return find_libraries("*", root=root, shared=False, recursive=recursive, runtime=False)
+    return find_libraries("*", root=root, shared=False, recursive=recursive)
 
 
 def find_all_libraries(root, recursive=False):
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -2226,10 +2226,10 @@ class WindowsSimulatedRPath(object):
         self.pkg = package
         self._addl_rpaths = set()
         self.link_install_prefix = link_install_prefix
-        self._internal_links = set()
+        self._additional_library_dependents = set()
 
     @property
-    def link_dest(self):
+    def library_dependents(self):
         """
         Set of directories where package binaries/libraries are located.
         """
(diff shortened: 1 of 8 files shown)


lib/spack/llnl/util/filesystem.py
```diff
@@ -2237,29 +2237,30 @@ def link_dest(self):
...
-        return pkg_libs | set([self.pkg.prefix.bin]) | self.internal_links
-
-    @property
-    def internal_links(self):
-        """
-        linking that would need to be established within the package itself. Useful for links
-        against extension modules/build time executables/internal linkage
-        """
-        return self._internal_links
+        pkg_libs.add(self.pkg.prefix.bin)
```
scheibelp: Could this be part of the default (when no `.libs` is available)? If a package implements `.libs`, then that list should already include libraries that are in `bin/`.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2280,12 +2281,36 @@ def include_additional_link_paths(self, *paths):
...
-        that require manual inclusion in a runtime linking scheme
+        that require manual inclusion in a runtime linking scheme.
+        These links are unidirectional, and are only
+        intended to bring outside dependencies into this package
 
         Args:
             *paths (str): arbitrary number of paths to be added to runtime linking
         """
-        self._addl_rpaths = self._addl_rpaths | set(paths)
+        self._addl_rpaths.extend(paths)
```
scheibelp: Was something going wrong when these were sets? The set datastructure avoids storing duplicates (although lacks ordering).

var/spack/repos/builtin/packages/clingo/package.py
```diff
@@ -118,3 +120,6 @@ def cmake_args(self):
             args += ["-DCLINGO_BUILD_WITH_PYTHON=OFF"]
 
         return args
+
+    def win_setup_rpath(self):
+        self.add_internal_links(os.path.join(self.prefix.Lib, "site-packages"))
```
scheibelp: Should this only happen when building with `+python`?

You could also just make sure to only reference this directory if it exists. Also I think it would be clearer if this used `self.spec['python'].platlib` (the Spack `python` package.py includes logic for locating `site-packages` which can account for special cases).

lib/spack/spack/package_base.py
```diff
@@ -194,18 +194,29 @@ class WindowsRPathMeta(object):
...
+    def add_search_paths(self, *paths):
         """Add additional rpaths that are not implicitly included in the search
         scheme
         """
-        self.win_rpath.include_additional_link_paths(*path)
+        self.win_rpath.include_additional_link_paths(*paths)
+
+    def add_internal_links(self, *paths):
+        """Add additional rpaths internal to a package, i.e.
+        linking pkg.prefix.bin to pkg.prefix.lib.site-packages"""
```
scheibelp: (this is only important if you disagree with [1])

I'm not sure the example is clarifying: I think it's sufficient to say that `*paths` designates a set of directories that contain libraries, and it is assumed that libraries in the package will need to be able to find the libraries in these paths.

lib/spack/spack/package_base.py
```diff
@@ -194,18 +194,29 @@ class WindowsRPathMeta(object):
...
+        self.win_rpath.add_internal_links(*paths)
+
+    def win_setup_rpath(self):
+        """This method should be overridden by packages needing bespoke RPATH
+        support. No-op otherwise"""
+        pass
```
scheibelp: (1) 

I assume this would never be more in-depth than supplying a set of additional directories that should be RPATHed. I think the interface can be narrowed then so that packages may define a function that returns a list of directories (rather than being an open-ended function that can do whatever it wants).

Moreover, I think we can get away with removing `package_base.add_internal_links` and just do:

```
    def windows_establish_runtime_linkage(self):
         """Establish RPATH on Windows

         Performs symlinking to incorporate rpath dependencies to Windows runtime search paths
         """
         if is_windows:
             self.win_rpath.add_internal_links(self.win_setup_rpath())
             self.win_rpath.establish_link()
```

scheibelp (changes requested):
A few requests - let me know if any seem off track.

lib/spack/llnl/util/filesystem.py
johnwparent: It was something I tried during debug. I'll revert.

var/spack/repos/builtin/packages/clingo/package.py
johnwparent: I can set it to only occur when the spec specifies python, but I will note that without that variant set, this does become effectively a no-op. 

This extra rpath needs to point to the `Clingo` Package's `site-packages`, not `Python`'s, as that is where the clingo `.pyd` file (python extension module) that requires a link to `clingo.dll` lives, and where Spack attempts to load it from. Unless `self.spec["python"].platlib` returns the dependent packages `site-packages`?

lib/spack/spack/package_base.py
johnwparent: [1] works.

lib/spack/spack/package_base.py
johnwparent: I'm fine with these changes, however we may want to consider the case where a package's `win_setup_rpath` method may also want to invoke the `WindowsRPathMeta.add_search_paths` (which extends the directories that the bin/lib dirs link against) as opposed or in addition to the internal links. Ostensibly this new refactored version could call that method and return a set of directories for internal linking but that feels a bit convoluted.

var/spack/repos/builtin/packages/clingo/package.py
scheibelp: > I can set it to only occur when the spec specifies python, but I will note that without that variant set, this does become effectively a no-op.

That's fine: if the directory doesn't exist, then it should be.

> Unless self.spec["python"].platlib returns the dependent packages site-packages?

Yes - in more detail: `platlib` is relative. Packages which depend on `python` get a per-class setting of `python_platlib` (you can see an example of `llvm` using this).

lib/spack/spack/package_base.py
scheibelp: Do we need users to explicitly decide whether to invoke `add_internal_links` or `add_search_paths`? Can we have (from the user's perspective) a single function for this, and then examine the paths to decide whether they are internal or not?

johnwparent: Maybe internal isn't the best term (my fault for using it) to articulate what I was considering here. Essentially, I wanted the `add_search_paths` to behave as appending a path to the `-rpath` arg given to ld would, whereas the  `add_internal_links` was intended to be more along the lines of adding a binary/library to the manifest of things that need linking, and it would then be linked against everything in the RPath args. In practice I deviated from this slightly, `add_internal_links` does not link against its package deps derived rpath, and bin/lib dirs do not link against each other. This was done to avoid over linking and causing conflicts or issues. I can include this extra behavior very easily if we want.

To answer your question more directly, I think parsing the path to determine if it's inside of the package isn't really solving the issue here, as the `add_search_path` method could point to something inside the package prefix, but that isn't included in the `bin`, `lib`, `lib64` set of directories, or the rpath derived from dependent packages. Consider the case where we have directory structure 

|---- bin --- someexe.exe
|
|-----lib
| 
|-----other- | --- some-different-sub-project --- otherlib.dl

where someexe.exe requires a link to otherlib.dll. Parsing would determine the link is internal and then link the bin to the other directory, whereas we want the other way around (the otherlib.dll linked to the bin/lib). 

Alternatively, we could forgo the internal links method and have an "add to link manifest" method of some sort where we include the directory of the desired binary/library that also requires linking against. If we add the packages bin to the rpath property definition (for windows only), this would resolve our clingo problem, and perhaps clarify usage?

scheibelp: Ok so to make sure:

* Internal links is for registering directories/binaries/libraries that need to locate other libraries (could it in fact be any of these three or is it e.g. never directories?)
* `add_search_path`/`include_additional_link_paths` is for adding directories that the former would want to rpath to

if so, we should probably rename those here:

* `add_library_dependent`: for things that need to find libraries
* `add_library_dir` or `add_rpath`: for locations which include libs (`add_search_path` is also fine, this name is less confusing than `add_internal_links`)

Furthermore, instead of having a custom Package function that returns just the dependent paths (as I initially suggested in this specific thread), it can return two lists: the list of dependent paths, and the list of search paths. Would that cover it?

lib/spack/spack/package_base.py
```diff
@@ -194,18 +194,39 @@ class WindowsRPathMeta(object):
     they would a genuine RPATH, i.e. adding directories that contain
     runtime library dependencies"""
 
-    def add_search_paths(self, *path):
-        """Add additional rpaths that are not implicitly included in the search
-        scheme
+    def win_add_linked_library(self):
```
scheibelp: I think it would be clearer to call this `win_add_library_dependent`.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2236,30 +2236,27 @@ def link_dest(self):
...
-
-        Paths provided to this method are linked against by a package's libraries
-        and libraries found at these paths are linked against a package's binaries.
-        (i.e. /site-packages -> /bin and /bin -> /site-packages)
+        Add paths to set of common paths that need to link against other libraries
 
-        Specified paths should be outside of a package's lib, lib64, and bin
+        Specified paths should be outside of a package's common link paths, i.e. the lib, lib64, and bin
         directories.
         """
```
scheibelp, line 2256: Can `dest` include paths that aren't directories? I assume we could accept either and then extract a directory for paths that are not directories, but I think it's worth mentioning either way in the docstring.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2236,30 +2236,27 @@ def link_dest(self):
...
-        self._internal_links = self._internal_links | set(*dest)
+        self._internal_links = self._internal_links | set(dest)
```
scheibelp: If we're updating the names of the methods, I think the properties could also have clearer names:

* `self._internal_links`
* `.internal_link_targets`
* `link_dest`: ambiguity: is it "the places where symlinks should be pointing" or "the places where symlinks should be placed"? I think it's the latter from reading the code, but I mean to illustrate the ambiguity; I think being consistent with "dependent" as a term would be useful

Overall I think it would be good to constrain the terminology to "dependent" and "search_path"/"rpath"/"target" (any one of those 3 is fine, but don't use more than one)

lib/spack/llnl/util/filesystem.py
```diff
@@ -2236,30 +2236,27 @@ def link_dest(self):
...
+
+    @property
+    def internal_link_targets(self):
+        return set(
+            [
+                x
+                for internal_path in self.link_dest
+                for x in find_all_shared_libraries(internal_path, recursive=True)
+            ]
```
scheibelp: Can you explain what this list comprehension is doing or expand it out (actually expanding it into multiple statements is probably best)? I don't get the two consecutive `for` statements.

scheibelp (changes requested):
Additional requests related to naming and list comprehensions.

lib/spack/llnl/util/filesystem.py
johnwparent: I do agree that users should be able to specify libraries/binaries/etc, so I'll add the catch and update the docstring.

lib/spack/llnl/util/filesystem.py
johnwparent: The list comprehension is iterating over each directory in `self.link_dest`, searching each of those directories with `find_all_shared_libraries` and then storing each element from the list that call returns. It's the same as 
```
for directory in self.link_dest:
   for lib in find_all_shared_libraries(directory):
      name_of_set.add(lib)
```

Would you still like that expressed more verbosely?

scheibelp: Yes: I think most folks will find it more readable (I definitely will).

johnwparent: I ended up removing the function entirely. I've added the results of `add_library_dependents` to the `library_dependents` property, so now we have a more consolidated expression of the libraries that need linking vs the rpaths we're linking against. Also this way only requires a single loop in the `establish_link` phase.

lib/spack/spack/package_base.py
```diff
@@ -2780,6 +2801,8 @@ def fetch_remote_versions(self, concurrency=128):
     def rpath(self):
         """Get the rpath this package links with, as a list of paths."""
         rpaths = [self.prefix.lib, self.prefix.lib64]
+        if is_windows:
+            rpaths.append(self.prefix.bin)
         deps = self.spec.dependencies(deptype="link")
         rpaths.extend(d.prefix.lib for d in deps if os.path.isdir(d.prefix.lib))
```
scheibelp: On windows, I think we agreed `lib` and `lib64` should generally be excluded.

lib/spack/llnl/util/filesystem.py
```diff
@@ -2130,7 +2134,7 @@ def find_libraries(libraries, root, shared=True, recursive=False):
 
     if is_windows:
         static_ext = "lib"
-        shared_ext = "dll"
+        shared_ext = "dll" if runtime else "lib"
```
scheibelp, line 2143: I think it would be worth adding a comment here like:

```
# For linking (runtime=False) you need the .lib files regardless of
# whether you are doing a shared or static link
shared_ext = "dll" if runtime else "lib"
```

lib/spack/llnl/util/filesystem.py
```diff
@@ -2189,7 +2193,7 @@ def find_all_static_libraries(root, recursive=False):
 
     See documentation for `llnl.util.filesystem.find_libraries` for more information
     """
-    return find_libraries("*", root=root, shared=False, recursive=recursive)
+    return find_libraries("*", root=root, shared=False, recursive=recursive, runtime=False)
```
scheibelp: Given that the docstring of `find_libraries` mentions:

> If shared is false, this has no meaning.

this option does not have to be specified.

scheibelp (changes requested):
I have a few requests which hopefully improve clarity. Let me know if you disagree.

lib/spack/spack/package_base.py
johnwparent: Yes we did, I missed this one. I'll amend it now, thanks for catching that.

lib/spack/llnl/util/filesystem.py
johnwparent: Will do.

lib/spack/llnl/util/filesystem.py
johnwparent: Good point!

scheibelp (dismissed):

lib/spack/spack/package_base.py
```diff
@@ -194,18 +194,39 @@ class WindowsRPathMeta(object):
...
-        self.win_rpath.include_additional_link_paths(*path)
+        return []
 
     def windows_establish_runtime_linkage(self):
         """Establish RPATH on Windows
 
         Performs symlinking to incorporate rpath dependencies to Windows runtime search paths
         """
         if is_windows:
+            self.win_rpath.add_library_dependent(*self.win_add_linked_library())
```
scheibelp: I think the CI is failing because this should be `win_add_library_dependent`

johnwparent: Oh yes it should!

scheibelp:
Close/reopen to restart checks

scheibelp (approved):

PR #33403, opened 2022-10-19
gitlab: Do not use root_spec['pkg_name'] anymore

scottwittenburg:
Investigating why this `develop` pipeline [job](https://gitlab.spack.io/spack/spack/-/jobs/3741456) failed led to the discovery that some jobs in that `e4s-oneapi` pipeline were marked correctly in the pipeline generation job as needing to be rebuilt, but did not appear in the generated pipeline yaml file.  The only reason it got noticed is because gitlab pipelines just recently started using `--use-buildcache dependencies:only,package:never`, so instead of installing the spec from the missing job from source, the pipeline failed.

The cause of the scheduling failure was that a single root spec in the environment had two `openssl` dependencies with different hashes, and the pipeline generation code still used the concrete root spec indexed by the package name to identify the concrete spec to schedule/build.  It may have become unsafe to index a concrete spec with a package name when spack got separate concretization of build dependencies, but that's just a guess.

This PR changes pipeline generation and spec rebuilding so that they no longer use the `root_spec["<pkg_name>"]` approach to identify specs.

### DIFF
--- a/lib/spack/spack/cmd/ci.py
+++ b/lib/spack/spack/cmd/ci.py
@@ -277,8 +277,8 @@ def ci_rebuild(args):
     ci_pipeline_id = get_env_var("CI_PIPELINE_ID")
     ci_job_name = get_env_var("CI_JOB_NAME")
     signing_key = get_env_var("SPACK_SIGNING_KEY")
-    root_spec = get_env_var("SPACK_ROOT_SPEC")
     job_spec_pkg_name = get_env_var("SPACK_JOB_SPEC_PKG_NAME")
+    job_spec_dag_hash = get_env_var("SPACK_JOB_SPEC_DAG_HASH")
     compiler_action = get_env_var("SPACK_COMPILER_ACTION")
     spack_pipeline_type = get_env_var("SPACK_PIPELINE_TYPE")
     remote_mirror_override = get_env_var("SPACK_REMOTE_MIRROR_OVERRIDE")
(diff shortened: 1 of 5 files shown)


scottwittenburg:
@zackgalbreath I could reproduce the `e4s-oneapi` pipeline generation errors locally, and this change fixed it for me.

scottwittenburg:
Now that we can ask spackbot to rebuild everything, let's use it on this PR tomorrow once the sync script pushes the branch.

scottwittenburg:
fyi @kwryankrattiger

scottwittenburg:
Link to pipeline that passed because all specs got pruned out: https://gitlab.spack.io/spack/spack/-/pipelines/194458

scottwittenburg:
@spackbot rebuild everything

kwryankrattiger:
I am a big fan of all the delete code!

lib/spack/spack/cmd/ci.py
```diff
@@ -360,10 +359,10 @@ def ci_rebuild(args):
...
             tty.debug(mirror_msg)
 
-    # Whatever form of root_spec we got, use it to get a map giving us concrete
-    # specs for this job and all of its dependencies.
-    spec_map = spack_ci.get_concrete_specs(env, root_spec, job_spec_pkg_name, compiler_action)
-    job_spec = spec_map[job_spec_pkg_name]
+    # Get the concrete spec to be built by this job.  We're using all 32 chars
+    # of the hash, so the list will just have one entry, unless there are hash
+    # collisions.
+    job_spec = env.get_by_hash(job_spec_dag_hash)[0]
```
haampie: Can you turn this comment into a hard requirement please? You're obtaining `job_spec_dag_hash` from some environment variable `SPACK_JOB_SPEC_DAG_HASH` so it's only true by convention and maybe the assumption breaks after a refactor, so it feels brittle.

The database has `query_one` for example, which asserts that the number of matches in <= 1. Something like that should be added to `Environment`.

haampie (changes requested):

haampie (approved):

PR #33404, opened 2022-10-19
replaced package shortbred with py-shortbred

luke-dt:
Shortbred v0.9.5 is on pypi rather than github, and now supports python 3. Removed the old version since python 2 is depreciated


adamjstewart:
We should first deprecate the old package before removing it in a future release, see https://spack.readthedocs.io/en/latest/packaging_guide.html#deprecating-old-versions

adamjstewart:
@erimar77 you created this package, do you have any comments on the rename?

var/spack/repos/builtin/packages/py-shortbred/package.py
```diff
@@ -0,0 +1,25 @@
...
+class PyShortbred(PythonPackage):
+    """ShortBRED is a system for profiling protein families of interest at
+    very high specificity in shotgun meta'omic sequencing data."""
+
+    homepage = "https://huttenhower.sph.harvard.edu/shortbred/"
+    pypi = "shortbred/shortbred-0.9.5.tar.gz"
+
+    version("0.9.5", sha256="a6ac09b858f14e2c0b8622b122ec91e5d02d32c12429cad66626d7ef26df10d5")
+
+    depends_on("python@3:", type=("build", "run"))
```
adamjstewart: Where do you see that Python 3 is required? According to the README, it's still 2.7.9

luke-dt: Sigh, so this software is kind of a mess and basically abandoned. They don't advertise the existence of 0.9.5 other than a github tag and uploading it to PyPi (which they don't mention on their website or README). v0.9.5 isn't even listed as a release on Github.

If you look through the commit history, they basically just changed print statements to make it python 3 compatible, since the previous package spit out errors if you used python 3 with it. I'm not sure if they made other changes, so I just fixed it there in case (and the PyPi page mentions 3.4, but it's not in the setup.py as a requirement).

There are also some other problems in the software like dependency check errors that are the result of typos, so I don't even know how necessary all the version requirements are.

adamjstewart: `print(...)` should work for both Python 2.7 and 3, so if there's no other reason to change it, I would leave it as `python@2.7.9:` (which includes 3)

luke-dt: I'll see if it runs with python 2.7.9 on another machine. I never actually tested it with that version because my Macbook refuses to run Python 2.7

luke-dt: @adamjstewart is there any way to get spack to print more information about a dependency conflict? I keep getting `==> Error: No version for 'python' satisfies '@2.7.9' and '@3.9.14'`

adamjstewart: Are you using the new concretizer? If it doesn't work with the new concretizer, I'm guessing it isn't possible to build with Python 2 anyway.

luke-dt: I believe so:

```
* **Spack:** 0.19.0.dev0 (75dd122175c48bf1c378191f79f88e08b02890ee)
* **Python:** 3.6.8
* **Platform:** linux-rocky8-cascadelake
* **Concretizer:** clingo
```

adamjstewart: The problem is that `py-biopython` depends on `python@3.6:`, so it's impossible to test Python 2 anyway. My suggestion would be to leave the dep as is (2.7.9+)

luke-dt: oh, I think that's a mistake with the `py-biopython` package then. v1.65 was released in 2014 and supports python 2 and python 3:

https://github.com/biopython/biopython/tree/biopython-165

adamjstewart (dismissed):

adamjstewart (approved):

erimar77:
@adamjstewart no concerns here

PR #30806, opened 2022-05-24
Optimize reuse from buildcaches

alalazo:
On my laptop using e4s:
```console
$ spack mirror add E4S https://cache.e4s.io
$ spack buildcache keys -it
```
takes a very long time with reuse. This PR is an attempt at seeking small changes that might have some impact in this scenario.

Optimizations:
- [x] database: don't sort on return from query_local
- [x] don't build the hash-lookup dictionary twice for a solve with reuse
- [x] don't check if spec satisfies `dev_path=*` when retrieved from a buildcache

PR #28673, opened 2022-01-29
Allow for multiple dependencies/dependents from the same package

alalazo:
fixes #11983 
closes #16447

same as #21683 which I cannot reopen after a force push for some reason

This PR changes the internal representation of `Spec` to allow for multiple dependencies or dependents stemming from the same package. This change permits to represent cases which are frequent in cross compiled environments or to bootstrap compilers.

Modifications:
- [x] Substitute `DependencyMap` with `_EdgeMap`. The main differences are that the latter does not support direct item assignment and can be modified only through its API. It also provides a `select_by` method to query items.
- [x] Reworked a few public APIs of `Spec` to get list of dependencies or related edges.
- [x] Added unit tests to prevent regression on #11983 and prove the synthetic construction of specs with multiple deps from the same package. 

Since #22845 went in first, this PR reuses that format and thus it should not change hashes. What happens is that in the list of dependencies we may have the same package being present multiple times with different associated specs.

PR #21618, opened 2021-02-11
specs: use lazy lexicographic comparison instead of key_ordering

tgamblin:
We have been using the `@llnl.util.lang.key_ordering` decorator for specs and most of their components. This leverages the fact that in Python, tuple comparison is lexicographic. It allows you to implement a `_cmp_key` method on your class, and have `__eq__`, `__lt__`, etc. implemented automatically using that key. For example, you might use tuple keys to implement comparison, e.g.:

```python
class Widget:
    # author implements this
    def _cmp_key(self):
        return (
            self.a,
            self.b,
            (self.c, self.d),
            self.e
        )

    # operators are generated by @key_ordering
    def __eq__(self, other):
        return self._cmp_key() == other._cmp_key()

    def __lt__(self):
        return self._cmp_key() < other._cmp_key()

    # etc.
```

The issue there for simple comparators is that we have to build the tuples *and* we have to generate all the values in them up front. When implementing comparisons for large data structures, this can be costly.

This PR replaces `@key_ordering` with a new decorator, `@lazy_lexicographic_ordering`. Lazy lexicographic comparison maps the tuple comparison shown above to generator functions. Instead of comparing based on pre-constructed tuple keys, users of this decorator can compare using elements from a generator. So, you'd write:

```python
@lazy_lexicographic_ordering
class Widget:
    def _cmp_iter(self):
        yield a
        yield b
        def cd_fun():
            yield c
            yield d
        yield cd_fun
        yield e

    # operators are added by decorator (but are a bit more complex)
```

There are no tuples that have to be pre-constructed, and the generator does not have to complete. Instead of tuples, we simply make functions that lazily yield what would've been in the tuple. If a yielded value is a `callable`, the comparison functions will call it and recursively compare it. The comparator just walks the data structure like you'd expect it to.

The ``@lazy_lexicographic_ordering`` decorator handles the details of implementing comparison operators, and the ``Widget`` implementor only has to worry about writing ``_cmp_iter``, and making sure the elements in it are also comparable.

Using this PR shaves another 1.5 sec off the runtime of `spack buildcache list`, and it also speeds up Spec comparison by about 30%. The runtime improvement comes mostly from two things:

1. lazily stopping the comparison as soon as possible (e.g., many specs just have different names, which is the firs thing the lazy generators return)
2. avoiding the use of `hash()`  in `_cmp_iter()` (it was used in `_cmp_key()` before)

PR #33406, opened 2022-10-19
Consolidate DAG traversal in traverse.py, support DFS/BFS

haampie:
This PR introduces breadth-first traversal, and moves depth-first traversal
logic out of Spec's member functions, into `traverse.py`.

It introduces a high-level API with three main methods:

```python
spack.traverse.traverse_edges(specs, kwargs...)
spack.traverse.traverse_nodes(specs, kwags...)
spack.traverse.traverse_tree(specs, kwargs...)
```

with the usual `root`, `order`, `cover`, `direction`, `deptype`, `depth`, `key`,
`visited` kwargs for the first two.

What's new is that `order="breadth"` is added for breadth-first traversal.

The lower level API is not exported, but is certainly useful for advanced use
cases. The lower level API includes visitor classes for direction reversal and
edge pruning, which can be used to create more advanced traversal methods,
especially useful when the `deptype` is not constant but depends on the node
or depth. 

---

There's a couple nice use-cases for breadth-first traversal:

- Sometimes roots have to be handled differently (e.g. follow build edges of
  roots but not of deps). BFS ensures that root nodes are always discovered at
  depth 0, instead of at any depth > 1 as a dep of another root.
- When printing a tree, it would be nice to reduce indent levels so it fits in the 
  terminal, and ensure that e.g. `zlib` is not printed at indent level 10 as a 
  dependency of a build dep of a build dep -- rather if it's a direct dep of my
  package, I wanna see it at depth 1. This basically requires one breadth-first
  traversal to construct a tree, which can then be printed with depth-first traversal.
- In environments in general, it's sometimes inconvenient to have a double
  loop: first over the roots then over each root's deps, and maintain your own
  `visited` set outside. With BFS, you can simply init the queue with the
  environment root specs and it Just Works. [Example here](https://github.com/spack/spack/blob/3ec73046995d9504d6e135f564f1370cfe31ba34/lib/spack/spack/environment/environment.py#L1815-L1816)


alalazo:
@spackbot run pipeline

tgamblin:
@haampie and I talked about this and the one remaining nit is that BFS will produce a somewhat unintuitive display of dependencies -- i.e. you aren't guaranteed that the line above any node you look at is that node's parent.

@haampie is going to look into whether we can tweak the vis code slightly to handle this, since there are lot of advantages to BFS, but I think we still want clear connections bt/w nodes in the tree output. I think it'll be a *way* better tree if you're guaranteed to see each node's undiscovered children at each level *and* you're guaranteed that nodes shown at a given level of indentation *are* children of the first outdented thing above them.

haampie:
cover = nodes, depth vs breadth first in `tree(...)`:
Screenshot from 2022-10-20 00-47-45
Screenshot from 2022-10-20 00-47-35

cover = edges, depth vs breadth first in `tree(...)`:

Screenshot from 2022-10-20 00-49-18
Screenshot from 2022-10-20 00-49-10

tgamblin:
I like the new version better -- it presents things higher up when it can and the parent/child relationships are real ones.  LGTM!

You can see:
* `perl` pulled from depth 3 -> 2
* `xz` pulled from depth 4 -> 3
* `readline` pulled from depth 5 -> 3
* `m4` pulled from depth 3 -> 2
* `diffutils` pulled from depth 5 -> 3
* `libiconv` pulled from depth 6 -> 2
* etc.

It's more compact and easier to read IMO

tgamblin:
The other kind of cool thing about this is that you could add a depth argument/default to `spack spec`, and things left over after pruning would be the most important ones in a much more real sense than before, since now we're displaying at min depth of the node, not depth in a particular traversal.

alalazo:
Closed / reopen to trigger a new merge commit without changing history in the branch. Fix in #33429 should get us past the previous CI error.

haampie:
@tgamblin one key (pun intended) difference between current dfs logic and bfs is that the former defaults to unique on `id` whereas the latter uses `spec.dag_hash`. As far as I can see, there's no real issue with `dag_hash` by default, as it exists for non-concrete specs, and the dag_hash logic does not depend on `spec.traverse(...)` but has its own logic and recursion. The advantage of sticking to `dag_hash` is that you can load two root specs from the filesystem and print the dag where their common deps are not duplicated.

haampie:
@spackbot run pipeline

haampie:
Hm, can readthedocs be re-triggered? Otherwise code coverage looks good now.

haampie:
Not in this PR, but when feeding environment root specs to the BFS and outputting a tree, you can get these kinds of concretizer outputs (no duplication of python, and as shallow output as possible):

Screenshot from 2022-10-20 18-58-04

I think @adamjstewart would appreciate that ;p

adamjstewart:
I like this a lot. I always check the DAG to see what version of Python I'm building with but it's always buried somewhere in a deep tree even though it's a direct dependency.

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def traverse_breadth_first_tree_nodes(parent_id, edges, depth=0):
+    for edge in edges[parent_id]:
+        yield (depth, edge)
+        for item in traverse_breadth_first_tree_nodes(edge.spec.dag_hash(), edges, depth + 1):
+            yield item
+
+
+def traverse_breadth_first_tree(specs, cover="nodes", deptype="all"):
```
tgamblin: This and other functions that talk about trees should be renamed -- they don't just traverse trees; they traverse graphs, right?

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,328 @@
...
+    if cover == "edges":
+        edges, parents = breadth_first_to_tree_edges(specs, deptype)
+        generator = traverse_breadth_first_tree_edges(None, edges, parents)
+    elif cover == "nodes":
+        edges = breadth_first_to_tree_nodes(specs, deptype)
+        generator = traverse_breadth_first_tree_nodes(None, edges)
+    else:
+        raise ValueError("cover should be nodes or edges")
```
tgamblin: Does BFS support paths too?  I am actually not sure if that even makes sense.

lib/spack/spack/spec.py
```diff
@@ -4604,12 +4608,19 @@ def tree(self, **kwargs):
...
-        for d, dep_spec in self.traverse_edges(
-            order="pre", cover=cover, depth=True, deptype=deptypes
-        ):
+
+        if breadth_first and cover in ("nodes", "edges"):
+            generator = spack.traverse.traverse_breadth_first_tree(
+                [self], cover=cover, deptype=deptypes
+            )
+        else:
+            generator = self.traverse_edges(order="pre", cover=cover, depth=True, deptype=deptypes)
```
tgamblin: So, if we're generalizing traversal, should we not do both of these the same way?  I feel like we should put both of these into `traverse.py` and handle both with your more general design.

Of course:
1. I'm not sure all the places `traverse_edges()` is used don't make assumptions about DFS, but I think they likely don't.
2. I've optimized `traverse_edges` -- I wonder if the new traversal is faster or slower.  We probably don't care if it is just for `tree()`, but it might be worth comparing nodes/s for each of these. If this ends up getting used in `spack find` or for spec comparison, the performance will matter with large DB's.

Also should we use BFS in `build_environment` or in things like `spack env activate`?  In particular in setting up the build env, post-order BFS has the nice property that one root's settings will not be overridden by a lower child of some *other* root.  I know this has come up in discussion.

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,328 @@
...
+        self.depth = depth
+
+
+def sort_edges(edges):
+    edges.sort(key=lambda edge: edge.spec.name)
+    return edges
+
+
+class BaseVisitor(object):
+    def __init__(self, deptype="all"):
```
tgamblin, line 29: what if instead of hard-coding `dag_hash()` below, this took an id function (defaulting to `id`) that the subclasses could use?  I think it's worth parameterizing that.

tgamblin:
@haampie:
> As far as I can see, there's no real issue with `dag_hash` by default, as it exists for non-concrete specs, and the dag_hash logic does not depend on `spec.traverse(...)` but has its own logic and recursion. The advantage of sticking to `dag_hash` is that you can load two root specs from the filesystem and print the dag where their common deps are not duplicated.

Logically, I don't see a problem with using `dag_hash` by default.  It is probably the most sensible thing for reasons you mentioned above.  Could `tree()` just ask for it though, instead of making it the default?  I think it is better in the code, at least if both traversals have the same defaults.  See comment in review below about making the two traversals consistent.

There are a few other considerations to think about, most of which don't have anything to do with `tree()`... but they'll matter if we do more with BFS:

1. `id` is really there for the tests, and because it's fast.  The tests use it to make sure we're not building crazy graphs with duplicated nodes.  If you can guarantee that, then you don't have to care about `dag_hash` and you can do fast `id` comparisons.
2. Historically, we didn't use `dag_hash()` because it hashed YAML, and it was very expensive to generate the YAML. Now that we use `json`, it's way faster, but also much slower than asking for `id`, especially for abstract specs.  There are really two places where this stuff matters:
    a. Loading package files: parsing all the directives in packages involves putting lots of little tiny specs in dictionaries and comparing them. It is probably not good if we have to hash json for those comparisons, but see the TODO and comment at https://github.com/spack/spack/pull/32516/files#diff-9e28cdd6f9cd66e61101de6e9d22b1c33561959e4305c63147bbfbf7b22da936R4071-R4080.  It would be nice to consolidate the `dag_hash` and `__hash__` logic.
    b. Sorting large databases: If we compare specs like graphs, traversal needs to be fast.  See https://github.com/spack/spack/pull/21618.  Then again, @alalazo removed sorting on query here: https://github.com/spack/spack/pull/30806/files#diff-751ae3f6692a00d628a34b4729c3ca2f8a5bbead2b18d8c6c5314e4e7da2c193L1530-R1535, and I think that's likely the right thing to do -- just avoid major traversals when comparing concrete specs and compare by attributes the client code actually cares about instead of imposing a total order.
4. `dag_hash()` exists for abstract specs but it is not cached for them.  I wonder if that will affect performance (particularly in 2a above).
5. The two root specs thing seems nice, semantically.  Should we need it though?  I guess *right now* we probably do.  But it makes me want to make the spec implementation immutable, at least for concrete specs.  I think `Specs` with the same hash should probably also be the same object... eventually.  That may make things difficult if we ever have to deal with specs that get the same hash for bad reasons, though.

tgamblin (changes requested):
Some questions/change requests.  See also the comment above for some broader context.

lib/spack/spack/traverse.py
haampie: It can make sense to use `traverse_breadth_first_edges` to iterate over all edges in BFS order without pruning; it doesn't make sense to print a tree this way, since it's the same order that DFS would give, so it would be the same as DFS except that it's doing a redundant traversal in BFS order first.

lib/spack/spack/traverse.py
haampie: Hm, there's two cases, both for which `None` is the root of a tree.
- `cover=nodes`: the edge-induced subgraph is always a tree (since every node is discovered through 1 parent)
- `cover=edges`: the edges do not form a tree as there can be multiple parents per node, but you have to combine it with the `parents` map to interpret it as a tree (the parents map keeps track of how the node was discovered first).

I just called it `traverse_breadth_first_tree` since it produces edges of a tree in depth-first order

lib/spack/spack/spec.py
haampie: To me it makes sense if there's an artificial root node `None` with multiple children for DFS too, so you don't have to do `for s in specs: for edge in s.traverse_edges(...):` but just `for dep in traverse_edges(specs)`. `root=False` would drop all `None -> spec` edges, which I believe happens already anyways.

I do think there's some logic though that relies on pre/post order iteration to do/undo build environment related things in the right order. But yeah, if we extract a tree from a dag with bfs, we can still do pre/post order bfs on it.

lib/spack/spack/traverse.py
haampie: Probably this should also receive `order=pre/post` as a kwarg

haampie:
@tgamblin I've consolidated the depth/breadth first traversal in a single generator now:

```python
spack.traverse.traverse_edges([spec_a, spec_b], order="pre/post/breadth")
```

with the usual kwargs. The `key` function and `visited` set can now also be passed to be compatible with what we had. Note that it takes a list of specs; so `s.traverse_edges(...)` is implemented as `traverse_edges([self], ...)`.

I'm quite happy with the implementation, especially now that the cover/direction logic is extracted to visitors and reused in both traversal orders.

Two comments:

1. One big concern I have though, when moving from `key=lambda s: s.dag_hash()` to `key=id` to be compatible, one of my BFS tests failed, the reason being that I used something along the lines of:
   
   ```python
   # Get two roots:
   x = Spec("x").concretized()
   y = x["y"]
   
   traverse_edges([x, y], order="breadth", cover="nodes")
   ```
   
   And this traverses the edge from `x` -> `y` because `id(x) != id(y)`, since indexing a spec yields an object that *looks* like a Spec, but it's really `SpecBuildInterface`.
   
   So, I would *strongly* suggest to avoid `id` cause the bug is so subtle, instead we should default to `key= lambda s: s.dag_hash()`.

2. Not sure what the prettiest interface is for generating trees. `spack.traverse.traverse_edges([s], order="pre", cover="nodes")` "just happens" to generate edges in a favorable order. But `spack.traverse.traverse_edges([s], order="breadth", cover="nodes")` does not. Should I add say `spack.traverse.tree([s], ???)` that dispatches to `traverse_edges` for DFS and something more advanced for BFS? What's a nice API? For example it wouldn't make sense to be able to pass `spack.traverse.tree([s], order="post")` so the API should be different.

Edit: solved 2 by creating a small helper `traverse_tree` that returns an appropriate generator.

lib/spack/spack/traverse.py
haampie: I'm now doing this!

lib/spack/spack/cmd/env.py
```diff
@@ -29,6 +29,11 @@
 import spack.schema.env
 import spack.tengine
 import spack.util.string as string
+from spack.traverse import (
+    CoverNodesVisitor,
+    sort_edges,
+    traverse_breadth_first_with_visitor,
+)
```
tgamblin: Can this be converted to `import spack.traverse as traverse`?  We've started to prefer that form to avoid import issues as modules get more complicated.

lib/spack/spack/cmd/spec.py
```diff
@@ -65,6 +65,12 @@ def setup_parser(subparser):
         choices=["nodes", "edges", "paths"],
         help="how extensively to traverse the DAG (default: nodes)",
     )
+    subparser.add_argument(
+        "--breadth-first",
+        action="store_true",
+        default=False,
+        help="traverse the DAG in breadth-first order",
+    )
```
tgamblin: do we even need an argument for this?  I am not sure when I would prefer the old depth-first view, and I think we should make BFS default here.

If we do still need an argument, it'd be better to make it `--order={breadth,depth}`.

lib/spack/spack/spec.py
```diff
@@ -1611,144 +1607,12 @@ def installed_upstream(self):
...
-        # we don't want to run canonical_deptype every time through
-        # traverse, because it is somewhat expensive. This ensures we
-        # canonicalize only once.
-        canonical_deptype = kwargs.get("canonical_deptype", None)
-        if canonical_deptype is None:
-            deptype = dp.canonical_deptype(deptype)
-            kwargs["canonical_deptype"] = deptype
-        else:
-            deptype = canonical_deptype
+        """Shorthand for spack.traverse.traverse_nodes"""
```
tgamblin: Can you make this a :ref: so the docs link?

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,414 @@
...
+        self.edge = edge
+        self.depth = depth
+
+
+def sort_edges(edges):
+    edges.sort(key=lambda edge: edge.spec.name)
+    return edges
+
+
+class BaseVisitor(object):
```
tgamblin, line 25: Can you add a docstring here?

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,414 @@
...
+
+from collections import defaultdict
+
+import spack.spec
+
+# Export only the high-level API.
+__all__ = ["traverse_edges", "traverse_nodes", "traverse_tree"]
+
+
+class EdgeAndDepth(object):
```
tgamblin: docstring?  It wasn't clear to me the first time I read this what this was for, or that this is what most visitors are looking at in `accept()`.

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,414 @@
...
+            return []
+
+        self.visited.add(key)
+        return self.visitor.neighbors(node)
+
+
+def get_visitor_from_args(cover, direction, deptype, key=id, visited=None, visitor=None):
+    """
+    Create a visitor object from keyword arguments, which simplifies the API
+    a bit, as cover, direction, deptype kwargs have been around for long in Spack.
```
tgamblin: > which simplifies the API a bit, as cover, direction, deptype kwargs have been around for long in Spack.

I agree with this but it probably doesn't belong in the docstring

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,414 @@
...
+
+def root_specs(specs):
+    """Initialize a list of edges from an imaginary root node to the root specs."""
+    return [
+        EdgeAndDepth(edge=spack.spec.DependencySpec(parent=None, spec=s, deptypes=()), depth=0)
+        for s in specs
+    ]
+
+
+def traverse_depth_first_edges_generator(nodes, visitor, post_order=False, root=True, depth=False):
```
tgamblin, line 155: This is way clearer!

lib/spack/spack/traverse.py
```diff
@@ -0,0 +1,414 @@
...
+
+class EdgeAndDepth(object):
+    def __init__(self, edge, depth):
+        self.edge = edge
+        self.depth = depth
+
+
+def sort_edges(edges):
+    edges.sort(key=lambda edge: edge.spec.name)
+    return edges
```
tgamblin, line 22: Ok this part is worrying me.  I missed this when reviewing #33481 -- specifically this part: https://github.com/spack/spack/pull/33481/files#diff-9e28cdd6f9cd66e61101de6e9d22b1c33561959e4305c63147bbfbf7b22da936R1739.

In #28673, we made it so that edges are sorted when they're added to a `Spec`'s `EdgeMap` -- see https://github.com/spack/spack/pull/28673#discussion_r823132555.  They're sorted not just by name but by name and deptype.  We do that because at least in the model, we allow there to be *two* dependencies/dependents of the same name (e.g. a build dep and a run dep of the same name), or multiple versions of some dependent in the full DB (which has lots of roots).

So, there are two issues:

1. This will result in nondeterministic traversals when there are multiple dependencies of the same name, and that's used by hashing, so it'll make the hashes of such packages unstable.  I don't think that happens anywhere in the real world, but we need to support it for compilers-as-deps.
2. I'm worried that DFS (which is used for hashing) now does a sort every time it looks for neighbors, as I think that may impact performance -- I think just sorting on `add()` would be a lot faster.

Can you look at:
1. Make DFS just take the "natural" edge order like it used to (i.e. what `__iter__` does on `EdgeMap`) so that it takes into account name and deptypes.
2. See if BFS can avoid the sort in `neighbors()` too.  if not I don't care as much b/c it's not performance critical.  I think maybe we should pull the sort out of `neighbors()` and handle that ordering in the tree code.
3. We should ensure that BFS is deterministic -- it seems like you not only need to think about name and deptype but also about children of different parents (which could share names).  If the sort on name is stable I *think* the order you get right now is deterministic but I am not 100% sure.  Again I don't care as much b/c it's not used for hashing, but we do want consistent output for `spack spec`

tgamblin (changes requested):
Mostly minor change requests and one that requires more thought.

RE:

> 1. So, I would *strongly* suggest to avoid `id` cause the bug is so subtle, instead we should default to `key= lambda s: s.dag_hash()`.

Yeah I agree with this, but with the aforementioned caveat that `dag_hash()` is definitely the right default for concrete specs, but it's going to hurt for abstract specs.  Can you see how changing the default affects performance of these two tests?

```python
spack python -c 'import spack.store; [spec.traverse() for spec in spack.store.db.query()]'
spack python -c 'import spack.repo; [p for p in spack.repo.path.all_package_classes()]'
```

That will test traversal speed and comparison speed (which uses hashes) for concrete and abstract specs.

Can you also see how these compare with `develop` *before* #33481?  Probably better to do that comparison first, actually (as it'll tell you if the sort in `traverse()` is expensive)

> 2. Edit: solved 2 by creating a small helper `traverse_tree` that returns an appropriate generator.

This is nice.

lib/spack/spack/cmd/spec.py
haampie: I think it makes sense not to introduce yet another flag, but provide sensible defaults.

haampie:
Regarding the performance, here are some details regarding _concrete_ specs (I also checked they were doing the same thing). This is on a database with 485 specs.

```python
import spack.store
specs = spack.store.db.query()
%timeit [[x.name for x in s.traverse(...)] for s in specs]
```

### develop `order=pre`

```
cover=nodes
59.1 ms ± 261 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

cover=edges
66.6 ms ± 97.9 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

cover=edges, key=lambda s: s.dag_hash()
76.8 ms ± 94.2 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### PR `order=pre`

```
cover=nodes
44.6 ms ± 115 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

cover=edges
50.7 ms ± 60.3 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

cover=edges, key=lambda s: s.dag_hash()
62.1 ms ± 310 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### PR `order=breadth`

```
order=breadth, cover=nodes
38.8 ms ± 84.1 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

order=breadth, cover=edges
41.1 ms ± 333 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

order=breadth, cover=edges, key=lambda s: s.dagh_hash()
49.7 ms ± 259 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### PR `order=pre` *without* edge sort

```
cover=nodes
44.1 ms ± 689 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

covers=edges
48.9 ms ± 353 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Somewhat unexpectedly, breadth first iteration is faster, maybe function call overhead / recursion in Python is slow? And overall I think the improvements in this PR compared to develop are thanks to moving logic into visitors, so we simply have fewer branches during traversal.

Regarding edge sorting: apparently the overhead is very limited, probably because things are typically sorted already?

Finally wrt `key=lambda s: s.dag_hash()`, I think we can live with that, not a major penalty.

lib/spack/spack/traverse.py
haampie: I'm not entirely sure if I follow, since the PR you linked actually _introduced_ the following edge sorting code in `traverse_edges` itself:

```python
            # This code determines direction and yields the children/parents
            if direction == 'children':
                edges = self.edges_to_dependencies
                key_fn = lambda dspec: dspec.spec.name
                succ = lambda dspec: dspec.spec
            elif direction == 'parents':
                edges = self.edges_from_dependents
                key_fn = lambda dspec: dspec.parent.name
                succ = lambda dspec: dspec.parent
            else:
                raise ValueError('Invalid traversal direction: %s' % direction)

            for dspec in sorted(edges(), key=key_fn):
```

i.e. sort edges by edge target package name.

haampie:
Regarding abstract specs, yeah, there's a huge penalty, I guess it's an extra dag traversal per `dag_hash()` call?

<details>
<summary>Here's some code to generate a DAG with just abstract specs</summary>

```python
import random
import string
import spack.spec

def random_str():
    return ''.join(random.choice(string.ascii_lowercase) for _ in range(random.randint(3, 8)))

def random_variant():
    if random.random() < 0.4:
        return random.choice(('+', '~')) + random_str()
    return random_str() + '=' + random_str()

def random_spec_str():
    components = [random_str()]
    components.extend(random_variant() for _ in range(random.randint(0, 4)))
    return ' '.join(components)

def random_deptype():
    return random.sample(("link", "build", "run", "test"), random.randint(1, 4))

def generate_random_abstract_specs(num_nodes=500, edge_odds=0.1):
    specs = [spack.spec.Spec(random_spec_str()) for _ in range(num_nodes)]
    total_edges = 0
    
    for i in range(num_nodes):
        num_edges = int(random.random() * edge_odds * (num_nodes - i - 1))
        total_edges += num_edges
        for j in random.sample(range(i+1, num_nodes), num_edges):
            specs[i].add_dependency_edge(specs[j], deptype=random_deptype())
    
    return specs, total_edges
```

</details>

And then for example, generating a dag with a 100 nodes and 192 edges:

```python
In [1]: specs, edges = spack.spec.generate_random_abstract_specs(100, 0.1)

In [2]: %timeit [[x.name for x in s.traverse(order='breadth', cover='edges', key=lambda s: s.dag_hash())] for s in specs]
95 ms ± 809 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %timeit [[x.name for x in s.traverse(order='breadth', cover='edges')] for s in specs]
2.37 ms ± 7.34 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

haampie:
Finally, actually dumping a tree for each concrete spec in the database:

```python
In [12]: %timeit [s.tree(depth_first=True) for s in specs]
1.05 s ± 5.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [13]: %timeit [s.tree(depth_first=False) for s in specs]
1 s ± 9.27 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I also noticed there's an amazing O(n^2) string concat issue here, lots of `str += ...` in `tree`.

haampie:
Final comment before next round of reviews: I think it's good to leave the PR as is, with the original `key=id` default and edge sorting. I think changing the `key` may lead to unforeseen issues and edge sorting like this has been in place for a long time, and doesn't seem to be performing terribly bad (also: python promises stable sort, so if we just sort by pkg name, the edge type stays in original order anyways).

lib/spack/spack/traverse.py
tgamblin: Ok we discussed this in the meeting today, and we decided this does no harm *for now*.  We'll need to revisit this sort for determinism when we do compilers as deps not so long from now.

tgamblin:
> I also noticed there's an amazing O(n^2) string concat issue here, lots of str += ... in tree.

This seems maybe worth fixing in some later PR (though it's likely not bottlenecking anything at the moment)

tgamblin (approved):

PR #33410, opened 2022-10-19
Fix bad python.libs.ld_flags on macOS with external homebrew Python

### LINKED ISSUE #33409: Bad python.libs.ld_flags on macOS using external homebrew Python
### Steps to reproduce

When using homebrew on macOS, the spack Python package `self.config_vars["LDLIBRARY"]` variable points to `Python.framework/Versions/3.9/Python`, which is indeed a shared library but doesn't have the correct filename suffix. This confuses several packages (e.g. `met`) that use `python.libs.ld_flags` - see https://github.com/dtcenter/METplus/discussions/1762.

This can be fixed by pointing back to the symbolic link `Python.framework/Versions/3.9/lib/libpython3.9.dylib` (which points to `../Python`).

Tested to work with `met` on macOS and homebrew Python.

### Error message

The error message for `met` is
```
ld: cannot link directly with dylib/framework, your binary is not an allowed client of /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib/libPython.tbd for architecture x86_64
```
This is because `python.libs.ld_flags` is `-L/...../Python.framework/Versions/3.9 -lPython`, which makes the linker look for a non-existing `libPython.{a,dylib}` file. The only one it finds is `Python.tbd` in the default macOS system path.

```
* **Spack:** 0.19.0.dev0 (9c9b5b0d206b623a7fd188ba7388673adf87f11d)
* **Python:** 3.9.13
* **Platform:** darwin-monterey-x86_64
* **Concretizer:** clingo
```

### General information

climbfuji:
Fixes https://github.com/spack/spack/issues/33409. Tested on macOS using an external homebrew Python installing `met`.


adamjstewart:
I don't think `config_vars` is the right place to fix this, let's fix it in `libs` instead. It's also unclear to me whether or not this is a bug in python or a bug in met. If the path exists, and no other packages have complained about it not working, is it possible that met is the one that's broken?

climbfuji:
> I don't think `config_vars` is the right place to fix this, let's fix it in `libs` instead. It's also unclear to me whether or not this is a bug in python or a bug in met. If the path exists, and no other packages have complained about it not working, is it possible that met is the one that's broken?

I think the problem is not met - all met does is asking for `python.libs.ld_flags` which returns `-L/...../Python.framework/Versions/3.9 -lPython`. The linker then looks for `libPython.{a,so,dylib}` as it should.

I don't care if it gets fixed in config_vars or libs, as long as it gets fixed ;-)

climbfuji:
@adamjstewart I'll have an updated version shortly, just need to find time to test it.

climbfuji:
@adamjstewart I updated the PR, please let me know if this looks better (and works) for you.

adamjstewart:
I've asked @tgamblin to review this since he's also helped hack on this code recently for macOS.

climbfuji:
Sorry to ping @adamjstewart @scheibelp @tgamblin - can you take a look at the proposed bug fix for bad `python.libs.ld_flags` on macOS with external homebrew Python when you have a moment, please? Thanks ...

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1250,12 +1250,26 @@ def libs(self):
         # The values of LDLIBRARY and LIBRARY aren't reliable. Intel Python uses a
         # static binary but installs shared libraries, so sysconfig reports
         # libpythonX.Y.a but only libpythonX.Y.so exists. So we add our own paths, too.
-        shared_libs = [
-            self.config_vars["LDLIBRARY"],
+
+        # When using homebrew on macOS, self.config_vars["LDLIBRARY"] points to
+        # Python.framework/Versions/X.Y/Python, which is indeed a shared library
+        # but doesn't have the correct filename suffix. This confuses several
+        # packages (e.g. met) that use python.libs.ld_flags. Ignore this config.
```
scheibelp: I think this comment should be worded more directly based on the error listed in https://github.com/spack/spack/issues/33409: we want to avoid including a directory with a library that is not linkable on Mac OS. Also, `LDLIBRARY` yields `'Python.framework/Versions/3.9/Python'` for me on my Mac OS Python and it wasn't built with Homebrew, so I think the context can be simplified to "Python on Mac OS" vs. "Python with Homebrew on Mac OS".

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1250,12 +1250,26 @@ def libs(self):
...
+        if file_extension_shared == "":
+            shared_libs = []
+        else:
+            shared_libs = [self.config_vars["LDLIBRARY"]]
+        shared_libs += [
             "{}python{}.{}".format(lib_prefix, py_version, dso_suffix),
         ]
-        static_libs = [
-            self.config_vars["LIBRARY"],
+        # Similar for static libraries
```
scheibelp: This should be expanded to something like:

```
# Like LDLIBRARY for Python on Mac OS, LIBRARY may also refer to an un-linkable object
```

scheibelp (changes requested):
The logic looks good but I think the comments obfuscate the reasoning behind it.

var/spack/repos/builtin/packages/python/package.py
adamjstewart: It's prob best described as "Framework Python". Spack builds Python with the traditional Unix filesystem hierarchy, but Homebrew and other PMs can build framework Python instead.

climbfuji:
Thanks for your reviews. I updated the documentation/comments, does it look better now?

scheibelp (approved):

climbfuji:
@tgamblin @adamjstewart Gentle reminder to review this PR if you have time ... thanks!

scheibelp:
@spackbot run pipeline

scheibelp:
Thanks @climbfuji!

PR #33418, opened 2022-10-19
Update TotalView versions and website

petertea:
This commit updates the totalview/package.py script to contain the current website 'totalview.io' rather than the old RogueWave website.  It also contains only the 5 most recently available versions of TotalView, and expands the platforms covered to include macOs and the new CORAL-2 systems such as rzvernal and tioga at LLNL.  There is a change in the resource URL from the previous packages as I could never get the style

 url="totalview_{0}_linux_x86-64.tar".format(version),

to work for me.   Instead it would complain about
==> Error: ValueError: Invalid url format from url: totalview_<function version at 0x2aaab112b158>_linux_x86-64.tar
I have not been able to resolve that, but the method I changed it to  does seem to work.  

Let me know if you have any questions or issues with this pull request.   Spack style passed though it failed in issort as the options to issort were not recognized.   I ran the unit tests, but since I only changed the package for TotalView I don't think the warnings I saw applied, or were because I did not have some necessary packages installed.


var/spack/repos/builtin/packages/totalview/package.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
```
tldahlgren: We still need the ending copyright year to be current.

```suggestion
# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```

var/spack/repos/builtin/packages/totalview/package.py
```diff
@@ -19,67 +19,192 @@ class Totalview(Package):
     be used as the primary source and the architecture appropriate tarball will
     be downloaded as a resource."""
 
-    homepage = "https://www.roguewave.com/products-services/totalview"
-    maintainers = ["dshrader"]
+    homepage = "https://totalview.io"
+    maintainers = ["nicholas-sly", "dshrader", "petertea"]
```
tldahlgren: @nicholas-sly Are you okay with being added as a maintainer on this package?

tldahlgren (commented):
Chiming in on a couple of minor points.

Deferring merge decision to @dshrader and or @nicholas-sly .

var/spack/repos/builtin/packages/totalview/package.py
petertea: @tldahlgren 
Nice catch.  I'm probably still working with the original package from when I started looking at this last year.  Hence the old copyright and nicholas sly as one of the maintainers.   I also just found a problem with trying to install TotalView on a Mac with this version.  I get a message about there being no checksum.  Isn't that what the sha256 strings are?   And then the install fails if I chose to install anyway   I could remove the Mac lines for this pull request if that is acceptable.

var/spack/repos/builtin/packages/totalview/package.py
dshrader: Since @nicholas-sly was removed from the maintainer list a while ago and he was ok with it then, and he hasn't responded here, let's go ahead and leave him off. @petertea, would you please go ahead and remove him from your branch?

var/spack/repos/builtin/packages/totalview/package.py
```diff
@@ -19,67 +19,192 @@ class Totalview(Package):
...
     license_comment = "#"
-    license_files = ["licenses/license.dat"]
-    license_vars = ["TV_LICENSE_FILE"]
+    license_files = ["licenses/license.dat", "licenses/tv_license_file"]
+    license_vars = ["LM_LICENSE_FILE", "TV_LICENSE_FILE"]
 
     # As the install of Totalview is via multiple tarballs, the base install
     # will be the documentation.  The architecture-specific tarballs are added
     # as resources dependent on the specific architecture used.
-    version("2021.1.16", sha256="4c51c7b6ab6b6afa7635ba2e9fc3b0ef833806f775a0ad0da26b13d6320625dd")
```
dshrader, line 33: I'll need to double-check on removing this... it is specifically used in TCE, but I don't know if TCE will keep its custom spack package for totalview.

petertea: David,   It would be easy to put this back in place.   I was concerned on making this a very large package as TotalView releases 4 times a year and I thought I'd just keep the most recent ones in place.   But I have 2021.1.16, as well 2021.2.14 that could be added.  The original package had 2019.2.18 which was an interim build that is not publicly available.

dshrader: I've been trying to think about what would be the best way to manage this... the previous scheme was generic in the names of the archives to use, but the sha's definitely made each entry not generic. With so many archives per release, and each one needing a resource section, that is a lot to maintain. Getting rid of old versions may be problematic for some projects (e.g., TCE), but maybe those projects can manage it in some way.

var/spack/repos/builtin/packages/totalview/package.py
petertea: @dshrader I have just removed @nicholas-sly from the maintainer list

tldahlgren (dismissed):
Thanks for the requested changes.

tldahlgren:
Will give @dshrader a chance to comment before merging.

var/spack/repos/builtin/packages/totalview/package.py
```diff
@@ -19,67 +19,192 @@ class Totalview(Package):
...
     # Because the actual source tarball is architecture dependent, the main
     # download is the documentation tarball and the source is downloaded as a
     # resource once the target architecture is known.
+
+    # Version 2022.3.6
     resource(
-        name="x86_64",
-        url="totalview_{0}_linux_x86-64.tar".format(version),
+        name="crayx86",
+        url="file://{0}/totalview_2022.3.6_linux_x86-64.tar".format(os.getcwd()),
```
dshrader: I'm curious why the `{0}` is necessary. Does it work without it? Previously, just the name of the archive was needed, no path. I assume the spack mirror logic will take care of the path. Will it work without the `{0}` at all?

petertea: I'll admit to being curious about this as well.   The original package and your update has

url="totalview_{0}_linux_x86-64.tar".format(version),

and that just never worked for me.   I'm still not sure why.   I tried with just the bare file name

url="totalview_2022.3.6_linux_x86-64.tar",
and that comes back with an Invalid url format from url:  totalview_2022.3.6_linux_x86-64.tar
So i copied the format from

def url_for_version(self, version):
    return "file:://{0}/totalview.{1}-doc.tar".format(os.getcwd(), version)

I left out the version for the above reasons, and it all seemed to work so I left it in.   However, with your question finally parsed correctly by me, I tried

url="file://totalview_2022.3.6_linux_x86-64.tar",

And that DOES work.   I'd be glad to make that change to make it a bit less awkward.   I can test this out in my private branch and then push that if things go well.   Incidentally the same tar file is used for both platform=cray and platform=linux when the target is x86_64.   However spack does not like having the same name for both platforms, so I used our alternate name format for one, with dots and dashes instead of underscores, and that seems to work okay.

petertea: I've removed the {0} .format(os.getcwd()) from the resource urls and tested on various platforms.  It all seems to be working fine.  I just pushed those changes.

dshrader (approved):
This looks good to me at this point. @petertea, if you'd like to just keep a certain number of the last few releases going forward, I'm ok with that. We've put together a solution that we'll implement in TCE to track older versions, although it will require some testing. In the meantime, keeping this file straight-forward is probably better than keeping an exhaustive list of every single release of totalview for all time.

Thank you for the changes!
David

var/spack/repos/builtin/packages/totalview/package.py
bernhardkaindl: According to dshrader's comment
https://github.com/spack/spack/pull/33418#pullrequestreview-1189075145 all looks good now and he approved the PR. Setting this thread to resolved.

var/spack/repos/builtin/packages/totalview/package.py
bernhardkaindl: According to dshrader's comment
https://github.com/spack/spack/pull/33418#pullrequestreview-1189075145 all looks good now and he approved the PR. Setting this conversation to resolved.

bernhardkaindl (approved):
According to dshrader's comment
https://github.com/spack/spack/pull/33418#pullrequestreview-1189075145 all looks good now and he approved the PR. Therefore I approve and merge.

PR #33429, opened 2022-10-20
FIX CI after git update

alalazo:
We use local repositories in unit-tests to stress some git related functionality. The default value of `protocol.file.allow` after CVE-2022-39253 conflicts with our fixtures. Here we restore the previous default value.

Modifications:
- [x] Add `protocol.file.allow always` to our git configuration in CI

### DIFF
--- a/.github/workflows/setup_git.ps1
+++ b/.github/workflows/setup_git.ps1
@@ -6,6 +6,10 @@ git config --global user.email "spack@example.com"
 git config --global user.name "Test User"
 git config --global core.longpaths true
 
+# See https://github.com/git/git/security/advisories/GHSA-3wp6-j8xr-qw85 (CVE--2022-39253)
+# This is needed to let some fixture in our unit-test suite run
+git config --global protocol.file.allow always
+
 if ($(git branch --show-current) -ne "develop")
 {
     git branch develop origin/develop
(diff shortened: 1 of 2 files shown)


.github/workflows/setup_git.sh
```diff
@@ -2,6 +2,10 @@
 git config --global user.email "spack@example.com"
 git config --global user.name "Test User"
 
+# See https://github.com/git/git/security/advisories/GHSA-3wp6-j8xr-qw85 (CVE--2022-39253)
```
alalazo:
```suggestion
# See https://github.com/git/git/security/advisories/GHSA-3wp6-j8xr-qw85 (CVE-2022-39253)
```

.github/workflows/setup_git.ps1
```diff
@@ -6,6 +6,10 @@ git config --global user.email "spack@example.com"
 git config --global user.name "Test User"
 git config --global core.longpaths true
 
+# See https://github.com/git/git/security/advisories/GHSA-3wp6-j8xr-qw85 (CVE--2022-39253)
```
alalazo:
```suggestion
# See https://github.com/git/git/security/advisories/GHSA-3wp6-j8xr-qw85 (CVE-2022-39253)
```

alalazo:
For the record, this is the current state for the pipeline I am waiting for in `develop` to finish:
Screenshot from 2022-10-20 09-42-19

Since the changes here cannot affect at all pipelines, but they should solve issues with CI, I'm going to merge them.

Issue #32435, opened 2022-08-29
Installation issue: py-pybind11


wspear:
### Steps to reproduce the issue

Encountered while trying to install gptune but the standalone package has the same issue. Observed with both @2.9.1 and @2.8.1

```console
$ spack specspack spec -I py-pybind11
Input spec
--------------------------------
 -   py-pybind11

Concretized
--------------------------------
 -   py-pybind11@2.9.1%gcc@9.4.0~ipo build_type=RelWithDebInfo arch=linux-ubuntu20.04-westmere
[+]      ^cmake@3.23.2%gcc@9.4.0~doc+ncurses+ownlibs~qt build_type=Release arch=linux-ubuntu20.04-westmere
[+]          ^ncurses@6.2%gcc@9.4.0~symlinks+termlib abi=none arch=linux-ubuntu20.04-westmere
[+]              ^pkgconf@1.8.0%gcc@9.4.0 arch=linux-ubuntu20.04-westmere
[+]          ^openssl@1.1.1p%gcc@9.4.0~docs~shared certs=mozilla arch=linux-ubuntu20.04-westmere
[+]              ^ca-certificates-mozilla@2022-03-29%gcc@9.4.0 arch=linux-ubuntu20.04-westmere
[+]              ^perl@5.34.1%gcc@9.4.0+cpanm+shared+threads arch=linux-ubuntu20.04-westmere
[+]                  ^berkeley-db@18.1.40%gcc@9.4.0+cxx~docs+stl patches=b231fcc arch=linux-ubuntu20.04-westmere
[+]                  ^bzip2@1.0.8%gcc@9.4.0~debug~pic+shared arch=linux-ubuntu20.04-westmere
[+]                      ^diffutils@3.8%gcc@9.4.0 arch=linux-ubuntu20.04-westmere
[+]                          ^libiconv@1.16%gcc@9.4.0 libs=shared,static arch=linux-ubuntu20.04-westmere
[+]                  ^gdbm@1.19%gcc@9.4.0 arch=linux-ubuntu20.04-westmere
[+]                      ^readline@8.1.2%gcc@9.4.0 arch=linux-ubuntu20.04-westmere
[+]                  ^zlib@1.2.12%gcc@9.4.0+optimize+pic+shared patches=0d38234 arch=linux-ubuntu20.04-westmere
[+]      ^ninja@1.11.0%gcc@9.4.0 arch=linux-ubuntu20.04-westmere
[...]

PR #33430, opened 2022-10-20
New package: py-darshan

jeanbez:
Includes the py-darshan package that provides Python utilities to interact with Darshan log records of HPC applications

### DIFF
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -20,13 +20,19 @@ class PyDarshan(PythonPackage):
     depends_on("py-setuptools", type="build")
     depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
     depends_on("py-cffi", type=("build", "run"))
+    # NOTE: SciPy is an indirect dependency needed for interpolate usage in pandas
+    #       It will be fixed in the next release
     depends_on("py-scipy", type=("build", "run"))
     depends_on("py-numpy@1.21:", type=("build", "run"))
     depends_on("py-pandas", type=("build", "run"))
+    # NOTE: matplotlib should be pinned until next release, for details:
+    #       https://github.com/darshan-hpc/darshan/issues/742
     depends_on("py-matplotlib@3.4", type=("build", "run"))
     depends_on("py-seaborn", type=("build", "run"))
     depends_on("py-mako", type=("build", "run"))
     depends_on("py-pytest", type=("build", "run"))
+    # NOTE: lxml is test-only indirect dependency via pandas
+    #       It will become optional in the next release 
     depends_on("py-lxml", type=("test"))
 
     depends_on("darshan-util", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -30,7 +30,7 @@ class PyDarshan(PythonPackage):
     depends_on("py-matplotlib@3.4", type=("build", "run"))
     depends_on("py-seaborn", type=("build", "run"))
     depends_on("py-mako", type=("build", "run"))
-    depends_on("py-pytest", type=("build", "run"))
+    depends_on("py-pytest", type="test")
     # NOTE: lxml is test-only indirect dependency via pandas
     #       It will become optional in the next release 
     depends_on("py-lxml", type=("test"))
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -32,7 +32,7 @@ class PyDarshan(PythonPackage):
     depends_on("py-mako", type=("build", "run"))
     depends_on("py-pytest", type="test")
     # NOTE: lxml is test-only indirect dependency via pandas
-    #       It will become optional in the next release 
+    #       It will become optional in the next release
     depends_on("py-lxml", type=("test"))
 
     depends_on("darshan-util", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -34,6 +34,6 @@ class PyDarshan(PythonPackage):
     @run_after('install')
     @on_package_attributes(run_tests=True)
     def install_test(self):
-        with working_dir('./darshan/tests', create=True):
+        with working_dir('./darshan/tests'):
             pytest = which('pytest')
             pytest()
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -31,9 +31,9 @@ class PyDarshan(PythonPackage):
 
     depends_on("darshan-util", type=("build", "run"))
 
-    @run_after('install')
+    @run_after("install")
     @on_package_attributes(run_tests=True)
     def install_test(self):
-        with working_dir('./darshan/tests'):
-            pytest = which('pytest')
+        with working_dir("./darshan/tests"):
+            pytest = which("pytest")
             pytest()
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -1,0 +1,27 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyDarshan(PythonPackage):
+    """Python utilities to interact with Darshan log records of HPC applications."""
+
+    homepage = "https://www.mcs.anl.gov/research/projects/darshan"
+    pypi = "darshan/darshan-3.4.0.1.tar.gz"
+
+    maintainers = ["jeanbez", "shanedsnyder"]
+
+    version("3.4.0.1", sha256="0142fc7c0b12a9e5c22358aa26cca7083d28af42aeea7dfcc5698c56b6aee6b7")
+
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-pandas", type=("build", "run"))
+    depends_on("py-matplotlib", type=("build", "run"))
+    depends_on("py-seaborn", type=("build", "run"))
+    depends_on("py-mako", type=("build", "run"))
+
+    depends_on("darshan-util", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -17,8 +17,10 @@ class PyDarshan(PythonPackage):
     version("3.4.0.1", sha256="0142fc7c0b12a9e5c22358aa26cca7083d28af42aeea7dfcc5698c56b6aee6b7")
 
     depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+    depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
     depends_on("py-cffi", type=("build", "run"))
-    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-numpy@1.21:", type=("build", "run"))
     depends_on("py-pandas", type=("build", "run"))
     depends_on("py-matplotlib", type=("build", "run"))
     depends_on("py-seaborn", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -20,10 +20,20 @@ class PyDarshan(PythonPackage):
     depends_on("py-setuptools", type="build")
     depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
     depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-scipy", type=("build", "run"))
     depends_on("py-numpy@1.21:", type=("build", "run"))
     depends_on("py-pandas", type=("build", "run"))
-    depends_on("py-matplotlib", type=("build", "run"))
+    depends_on("py-matplotlib@3.4", type=("build", "run"))
     depends_on("py-seaborn", type=("build", "run"))
     depends_on("py-mako", type=("build", "run"))
+    depends_on("py-pytest", type=("build", "run"))
+    depends_on("py-lxml", type=("test"))
 
     depends_on("darshan-util", type=("build", "run"))
+
+    @run_after('install')
+    @on_package_attributes(run_tests=True)
+    def install_test(self):
+        with working_dir('./darshan/tests', create=True):
+            pytest = which('pytest')
+            pytest()


var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+    version("3.4.0.1", sha256="0142fc7c0b12a9e5c22358aa26cca7083d28af42aeea7dfcc5698c56b6aee6b7")
+
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-pandas", type=("build", "run"))
+    depends_on("py-matplotlib", type=("build", "run"))
+    depends_on("py-seaborn", type=("build", "run"))
+    depends_on("py-mako", type=("build", "run"))
```
adamjstewart, line 32: Missing deps:
```python
depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
```

adamjstewart (changes requested):
@carns may also be interested in this PR

var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+    depends_on("darshan-util", type=("build", "run"))
```
tylerjereddy, line 38: I'm a bit confused about what is going on here. There's no `shared` variant of `darshan-util` spackage at this time, so I'm not clear on how you'll be able to link the built C object into the Python library?

You may find some related challenges/complications discussed at previous attempts to provide a draft version of a spackage:

- shared variant was drafted in here when doing some rough work: https://github.com/tylerjereddy/spack/pull/3
- original issue: https://github.com/darshan-hpc/darshan/issues/597

Did you actually test the build from source in a clean `spack` environment?

tylerjereddy: Well @shanedsnyder seems to find that the new version of the C lib build system *does* produce the shared lib by default, so *maybe* you're ok here. Probably good to double check that `pytest --pyargs darshan` passes though, if you didn't already.

shanedsnyder: Yep, `darshan-util` should be installing shared libs if `libtool` determines that the compiler can produce shared libs at configure time.

I suppose some spack compiler variants could not support shared libs -- is that something that could be handled gracefully somehow (i.e., pydarshan package can not be built if there is no `libdarshan-util.so`)? Otherwise, I assume the build of py-darshan will ultimately just fail.

tylerjereddy: I checked that the tests don't pass though, that's always important to check.

- `spack install py-darshan % gcc@11.2.0 ^ python@3.10.4`
- `spack install py-pytest ^ python@3.10.4`
- `spack load py-pytest py-darshan`
- `pytest --pyargs darshan`

```
================================================================================================================================ test session starts =================================================================================================================================
platform linux -- Python 3.10.4, pytest-7.1.3, pluggy-1.0.0
rootdir: /home/tyler/github_projects/spack, configfile: pytest.ini
collected 378 items / 2 errors                                                                                                                                                                                                                                                       

======================================================================================================================================= ERRORS =======================================================================================================================================
_____________________________________________ ERROR collecting opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/examples/tutorial/hello.py _____________________________________________
opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/examples/tutorial/hello.py:5: in <module>
    r = darshan.DarshanReport("ior_hdf5_example.darshan")
opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/report.py:374: in __init__
    self.open(filename, read_all=read_all)
opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/report.py:426: in open
    raise RuntimeError("Failed to open file.")
E   RuntimeError: Failed to open file.
---------------------------------------------------------------------------------------------------------------------------------- Captured stderr -----------------------------------------------------------------------------------------------------------------------------------
Error: darshan_log_open failed to open darshan log file ior_hdf5_example.darshan: No such file or directory.
_____________________________________________ ERROR collecting opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/examples/tutorial/plot.py ______________________________________________
ImportError while importing test module '/home/tyler/github_projects/spack/opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/examples/tutorial/plot.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/python-3.10.4-syf6nm32j5ogompuqj5vdaxbxh3cve5p/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/examples/tutorial/plot.py:4: in <module>
    from darshan.experimental.plots.matplotlib import *
E   ModuleNotFoundError: No module named 'darshan.experimental.plots.matplotlib'
============================================================================================================================== short test summary info ===============================================================================================================================
ERROR opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/examples/tutorial/hello.py - RuntimeError: Failed to open file.
ERROR opt/spack/linux-ubuntu22.04-skylake_avx512/gcc-11.2.0/py-darshan-3.4.0.1-4dszt7wqbohm6yovl3pflwnizkrxqywh/lib/python3.10/site-packages/darshan/examples/tutorial/plot.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================================================================================================= 2 errors in 7.02s ==================================================================================================================================

```

var/spack/repos/builtin/packages/py-darshan/package.py
jeanbez: Included missing deps

var/spack/repos/builtin/packages/py-darshan/package.py
jeanbez: @tylerjereddy I did try running a clean install with an application that uses the library, and it works ok. I believe there might be something wrong with the tests, as it does not find the sample file. @shanedsnyder can you provide that file?

tylerjereddy: The files should be installed as part of the installation, could be a difference on `spack` build system details perhaps. Either way, that's the normal incantation of the test suite which passes in our CI.

tylerjereddy: perhaps stuff like `include_package_data` or `package_data` behavior has to be checked in the build logs when using `spack` & if there are any shims related to that stuff in standard `spack` installs

var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -0,0 +1,29 @@
...
+    homepage = "https://www.mcs.anl.gov/research/projects/darshan"
+    pypi = "darshan/darshan-3.4.0.1.tar.gz"
+
+    maintainers = ["jeanbez", "shanedsnyder"]
+
+    version("3.4.0.1", sha256="0142fc7c0b12a9e5c22358aa26cca7083d28af42aeea7dfcc5698c56b6aee6b7")
+
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+    depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
```
tylerjereddy, line 21: I think our test suite has a `pyproject.toml` requirement like this:

`"importlib_resources;python_version<'3.9'"`

Which is a bit different than the canonical usage of `importlib_resources`, so I think it may be required up to that version to get full functionality when testing at least.

var/spack/repos/builtin/packages/py-darshan/package.py
tylerjereddy: One other thing that might help--it looks to me like the `*.darshan` files are installed properly by `spack` so they should be availble for testing. The traceback above shows attempts to import files that aren't under test, so maybe Adam might know why they would get imported by `spack` but not in conventional usage outside of `spack`.

adamjstewart: Spack isn't importing anything (at least in this test), the files picked up by pytest are. Spack has no fancy shims for package_data, we just run `pip install .` with a few extra flags to prevent downloads.

tylerjereddy: The errors I report above are a result of running `pytest --pyargs darshan` in the `git` checkout of `spack` for this PR. The reason that happens is that `spack` has `pytest.ini` in its root with `python_files = *.py`, which overrides the largely-sensible default (https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-python_files) of:

> files matching test_*.py and *_test.py will be considered test modules.

I won't comment on that--I assume the `spack` team has good reason to treat every single `*.py` file as under test/import by `pytest`, as surprising at it may be under normal circumstances.

Okay, so moving on from that, go to another directory like `/tmp` and run `pytest --pyargs darshan` on latest version of this branch:

`31 failed, 287 passed, 60 skipped, 76 warnings in 67.12s (0:01:07)`

All of those failures are real and informative, and could be easily missed if you only just ran an example workflow instead of the full suite. There are three reasons for failure:

1. `py-scipy` is a real (runtime) dependency of the project because a subset of codepaths/analysis pathways use an interpolation routine in `pandas` that needs SciPy--it turns out we did remove the need for SciPy more recently on `main`, but not in a tagged version yet, so I'd say you pretty much need to add it here for now.
2. `py-lxml` is mandatory to pass the tests only, because `pandas` is used to verify the integrity of some HTML report generation tests and it uses `lxml` for that read-in; Shane convinced me to make this an optional testing dependency in the future, and it is now in `main` where the relevant tests will skip, but until the next tagged release I'd say it may be sensible to add a dependency in the spackage, or at least a scoped test-stage dep.
3. `py-matplotlib` needs to be pinned or version capped at the tag of `py-darshan` in use here because of: https://github.com/darshan-hpc/darshan/issues/742 ; that will be relaxed in the next tagged release

So, adjusting based on the feedback from running the test suite, a diff like this, which is just a rough draft, gets us much closer:

```diff
diff --git a/var/spack/repos/builtin/packages/py-darshan/package.py b/var/spack/repos/builtin/packages/py-darshan/package.py
index 8f4202164a..4452453403 100644
--- a/var/spack/repos/builtin/packages/py-darshan/package.py
+++ b/var/spack/repos/builtin/packages/py-darshan/package.py
@@ -21,9 +21,12 @@ class PyDarshan(PythonPackage):
     depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
     depends_on("py-cffi", type=("build", "run"))
     depends_on("py-numpy@1.21:", type=("build", "run"))
+    depends_on("py-scipy", type=("run"))
     depends_on("py-pandas", type=("build", "run"))
-    depends_on("py-matplotlib", type=("build", "run"))
+    depends_on("py-matplotlib@3.4", type=("build", "run"))
     depends_on("py-seaborn", type=("build", "run"))
     depends_on("py-mako", type=("build", "run"))
+    # should be test only really:
+    depends_on("py-lxml", type=("run"))
 
     depends_on("darshan-util", type=("build", "run"))
``` 

Now, all the tests pass:

`318 passed, 60 skipped, 17 warnings in 82.71s (0:01:22)`

As a next step--I did my tests with Python `3.10`; I'd suggest repeating my test procedure with Python below version `3.9` to check on the `importlib_resources` issue I mention as well (probably version `3.8`, since that would miss it in the current spackage I think).

If you want to be really thorough, though it is probably overkill for now, to avoid all the test skips you'd need to install the test data files from https://github.com/darshan-hpc/darshan-logs, but that seems like something for a follow-up perhaps.

jeanbez: @tylerjereddy and @shanedsnyder I made some changes, and it seems that all the tests are passing when installing `spack install --test=root py-darshan`

jeanbez: Oh, I missed your last comments, let me go over them.

jeanbez: @tylerjereddy yep, so the changes you suggested were those I made in the previous PR, which fixed the issues and passed all the tests (now using the correct tests). I also tried using Python 3.6 and 3.9, and both worked as expected with no test errors. With Python 3.10, I got the same error reported in #32435 while spack tried to install a dependency. @adamjstewart, do you have any updates on that issue? Other than that, we should be good to go.

adamjstewart: No updates from me.

tylerjereddy:
Checking with Python `3.8` for the reason noted in https://github.com/spack/spack/pull/33430#discussion_r1002208907 would be helpful. I tried it but the `py-pybind11` issue noted above blocked me.

jeanbez:
With 3.8 I get the same error, @adamjstewart is there any workaround we could use while that issue is not fixed?

adamjstewart:
You mean the error reported in https://github.com/spack/spack/issues/32435? @ax3l is the expert on pybind11.

shanedsnyder (dismissed):

shanedsnyder:
Looks good to me, I went ahead and approved. Looks like @jeanbez addressed other feedback from previous review, so I think we're okay here?

I know there's the outstanding issue of not being able to verify tests pass with all Python versions due to https://github.com/spack/spack/issues/32435, but I think that's a minor snag and am fine with merging as is.

var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -0,0 +1,39 @@
...
+    depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-scipy", type=("build", "run"))
```
adamjstewart, line 25: Scipy is not a dependency of this project. It isn't listed under `install_requires` and isn't imported anywhere.

var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -0,0 +1,39 @@
...
+    depends_on("py-numpy@1.21:", type=("build", "run"))
+    depends_on("py-pandas", type=("build", "run"))
+    depends_on("py-matplotlib@3.4", type=("build", "run"))
+    depends_on("py-seaborn", type=("build", "run"))
+    depends_on("py-mako", type=("build", "run"))
+    depends_on("py-pytest", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-pytest", type="test")
```

var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -0,0 +1,39 @@
...
+    depends_on("py-lxml", type=("test"))
```
adamjstewart, line 36: I don't actually see lxml used anywhere in this project

var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -0,0 +1,39 @@
...
+    version("3.4.0.1", sha256="0142fc7c0b12a9e5c22358aa26cca7083d28af42aeea7dfcc5698c56b6aee6b7")
+
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+    depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
+    depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-scipy", type=("build", "run"))
+    depends_on("py-numpy@1.21:", type=("build", "run"))
+    depends_on("py-pandas", type=("build", "run"))
+    depends_on("py-matplotlib@3.4", type=("build", "run"))
```
adamjstewart, line 30: Where do you see these version restrictions on numpy and matplotlib?

adamjstewart (changes requested):

tylerjereddy:
SciPy is an indirect dependency needed for `interpolate` usage in `pandas`, so the import is concealed. Will fail in some runtime cases without it until next release.

`lxml` is test-only dependency; indirect again via `pandas`. It will be optional in next release, but fail test in this tagged release.

`matplotlib` pin should be linked to an issue in discussion above, fixed in next release.. but really a test-time only problem..

adamjstewart:
Can you add comments explaining all of those things? Maybe a separate section that lists indirect deps.

jeanbez:
> Can you add comments explaining all of those things? Maybe a separate section that lists indirect deps.

@adamjstewart I've included the needed comments for each dependency so we can keep track.

adamjstewart (approved):

adamjstewart:
Thanks for the comments! We can remove these after the next release, just didn't want someone (aka me) to review the next PR and wonder where all these deps came from.

I think everyone has approved this PR at one time or another, so I'm assuming it's good to merge now.

PR #33439, opened 2022-10-20
[ELPA] add sha256 for elpa-2022.11.001.rc2.tar.gz

toxa81:
Latest elpa version requires CPP variable to be set. Current solution is to add the following lines.
```
       options += [
            "CPP=cpp"
        ]
```


tldahlgren (commented):
Confirmed sha256 of the new version.

tldahlgren:
@spackbot fix style

toxa81:
@haampie can we merge this?

var/spack/repos/builtin/packages/elpa/package.py
```diff
@@ -209,6 +212,8 @@ def configure_args(self):
                 "SCALAPACK_LDFLAGS={0}".format(spec["scalapack"].libs.joined()),
             ]
 
+        options += ["CPP=cpp"]
```
haampie: Why do you need this? As long as they don't set this variable it's better to use the [GNU Make default `$(CC) -E`](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html)

toxa81: Otherwise the build script will fail with the error kind of "CPP variable is not definced". Shall I try "CPP=$(CC) -E" ?

haampie: Okay, but this is not how to fix it, since we don't know if `cpp` is the preprocessor of the C compiler Spack uses.

Basically Elpa's release is broken.

Can you tell them to require `AC_PROG_CPP` if they need the preprocessor?

https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/C-Compiler.html

Then autotools figures it out correctly.

haampie: So for example:

```m4
AC_INIT([amhello], [1.0], [bug-automake@gnu.org])
AM_INIT_AUTOMAKE([-Wall -Werror foreign])
AC_PROG_CC
AC_PROG_CPP
AC_CONFIG_HEADERS([config.h])
AC_CONFIG_FILES([
 Makefile
 src/Makefile
])
AC_OUTPUT
```

will give

```
checking how to run the C preprocessor... gcc -E
```

and then the makefile contains `CPP=gcc -E`.

Spack shouldn't have to do this.

bernhardkaindl: Just for information for reviewers: Upstream has applied the fix for the upcoming 2022-11 release:
https://gitlab.mpcdf.mpg.de/elpa/elpa/-/commit/b8351486f74390d12dddf1046d2d60c3e88355df

@toxa81: When you update the package, also add the link to the the upstream gitlab repo in the package's recipe
and that upstream communication can be done using https://github.com/marekandreas/elpa/issues.

I ask because we could not find this information to communicate with them from the home page currently tagged in the recipe when processing #33639 for elpa recently. We thought they only offer email for contacting them from reading that page.

If they would update their home page to include this information prominently, it would alternatively also be great.

toxa81: @bernhardkaindl @haampie I updated the recipe. Now it compiles without extra CPP variables in the spack package.

haampie (approved):

haampie:
@spackbot run pipeline

PR #33443, opened 2022-10-20
gitlab: when_possible -> false

haampie:
`reuse` and `when_possible` concretization broke the invariant that
`spec[pkg_name]` has unique keys. This invariant is relied on in tons of
places, such as when setting up the build environment. Right now,
there is no guarantee that the intended spec is returned from the index
operator.

When using `when_possible` concretization, one may end up with two or
more `perl`s or `python`s among the transitive deps of a spec, because
reuse concretization prunes build-only deps that are later re-attached.

I think that a minimal reproducer could be created like this:

```
x
    ^ninja # reusable spec
        ^python # build type dep -- dropped during concretization
            ^openssl
    ^openssl # non-reusable for some reason
```

Then `x['openssl']` will likely traverse x -> ninja -> python -> openssl before
considering `x -> openssl` directly, due depth first traversal and edge ordering
by target package name.

Until the code base is fixed not to rely on this broken property of
`__getitem__`, we should disable reuse in CI.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/build_systems/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/build_systems/spack.yaml
@@ -3,7 +3,7 @@ spack:
 
   concretizer:
     reuse: false
-    unify: when_possible
+    unify: false
 
   config:
     install_tree:
--- a/share/spack/gitlab/cloud_pipelines/stacks/e4s-oneapi/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/e4s-oneapi/spack.yaml
@@ -3,7 +3,7 @@ spack:
 
   concretizer:
     reuse: false
-    unify: when_possible
+    unify: false
 
   config:
     concretizer: clingo


scottwittenburg (approved):

zackgalbreath:
the build of [`c-blosc/jvdbnnd`](https://gitlab.spack.io/spack/spack/-/jobs/3794435) actually succeeds now!

haampie:
Aargh, so this one failed https://gitlab.spack.io/spack/spack/-/jobs/3794431

```
Useless use of a variable in void context at /home/software/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeh/linux-ubuntu20.04-x86_64/oneapi-2022.1.0/automake-1.16.5-uzqwcqlszsuh5z5g27yrdekgsoyldisj/bin/automake line 8469.
```

I tried to reproduce locally, and indeed... automake has a spurious last line `$exit_code;` right after `exit $exit_code;` which triggers a warning (and it's running perl in warning = error mode).

Then I uninstalled, built automake from sources, spurious trailing line gone. Tried to install from binary cache a second time -- and the weird line is gone?! How's this even possible. Relocation bug?

zackgalbreath:
> How's this even possible. Relocation bug?

Dunno, but the job succeeded on an automatic retry, so hopefully we'll still get a passing pipeline soon. 🤷

haampie:
Just for my understanding: automake was built once, right?

haampie:
It seems unlikely the automake tarballs with binaries were modified between the runs, given how fast the restart was after the initial failure.

My bet is on a race when relocating files in parallel, given that the automake executable is a hardlink:

```
31336196 -rwxr-xr-x 2 harmen harmen 262415 Sep 14 13:19 automake
31336196 -rwxr-xr-x 2 harmen harmen 262415 Sep 14 13:19 automake-1.16
```

:hankey:

scottwittenburg:
When I ran the reproducer locally it succeeded, and it seems to have done everything the same up to the failure point the job you linked (version of autoake/autoconf, etc).  So at least initially a race conditions seems plausible.  I'm still looking at some weird stuff though:  

1. There's still two of the same set of specs in the concretized stack as there was before (openssl, ...)
2. Somehow it looks like we're back to scheduling the wrong one of each, i.e. the one actually used in the build was the one we didn't need rebuilding.

I'm looking into (2) now.

haampie:
I managed to reproduce that issue btw.

I don't see what (root) spec you're talking about @scottwittenburg?

scottwittenburg:
> I don't see what (root) spec you're talking about @scottwittenburg?

I was referring to the failed `yaksa` job you linked above, sorry for not being clear.

scottwittenburg:
> Somehow it looks like we're back to scheduling the wrong one of each, i.e. the one actually used in the build was the one we didn't need rebuilding.

Running it locally with some extra debug, the `yaksa` that needed to be rebuilt in that pipeline depended on the `automake` that didn't need to be rebuilt, while one of the `yaksa` that *did not* need to be rebuilt depended on the `automake` that did.  Which seems kind of weird, but doesn't actually appear to be incorrect scheduling of gitlab jobs/deps.

PR #33446, opened 2022-10-20
Add recent ReFrame versions

vkarak:


var/spack/repos/builtin/packages/reframe/package.py
```diff
@@ -24,6 +24,11 @@ class Reframe(Package):
     maintainers = ["victorusu", "vkarak"]
 
     version("master", branch="master")
+    version("4.0.0-dev.1", sha256="6db55c20b79764fc1f0e0a13de062850007425fa2c7f54a113b96adee50741ed")
+    version("4.0.0-dev.0", sha256="a96162a88a36ea0793836c492a39470010f6e63b8d9bd324c033614d27304fa6")
+    version("3.12.0", sha256="425cc546e24edd5b2dbfcdcb61dbbf723ca1a2a2977948e359e893514f5eb10f")
```
tldahlgren: Should this version be "preferred" over the two newer `dev` versions?

tldahlgren (commented):
Confirmed 5 new sha256.

var/spack/repos/builtin/packages/reframe/package.py
vkarak: Correct point! I'll add a `preferred=True`. for the 3.12.0.

alalazo:
@spackbot fix style

alalazo (approved):

PR #33449, opened 2022-10-20
py-checkm-genome and py-pysam: bumped version and updated deps

luke-dt:


var/spack/repos/builtin/packages/py-checkm-genome/package.py
```diff
@@ -13,18 +13,26 @@ class PyCheckmGenome(PythonPackage):
     homepage = "https://ecogenomics.github.io/CheckM"
     pypi = "checkm-genome/checkm-genome-1.0.11.tar.gz"
 
+    version("1.2.1", sha256="33907aa7bbf029f8345e33df80d5c89b7a719041f55ece4f7470cd061c8eff76")
     version("1.0.13", sha256="ffb7e4966c0fac07c7e6e7db6f6eb5b48587fa83987f8a68efbaff2afb7da82e")
     version("1.0.11", sha256="e475d9817d12fa771dbccc80f47758b742fc67c25261dc8ca0c0dc898c2a5190")
 
     # pip silently replaces distutils with setuptools
+
```
tldahlgren, line 21: I realize there is already precedence otherwise in this package but I find `depends_on("<spec>", when="<condition>", type="...")` more readable.

tldahlgren (commented):
Confirmed 2 new sha256.

var/spack/repos/builtin/packages/py-checkm-genome/package.py
tldahlgren: Looks like I forgot to note that I confirmed both new sha256.

adamjstewart: I usually put `when` before `type` just because that's the order they are defined in the directive definition. I also usually put deps in order from newest to oldest since that's the order we put versions. But both of these are preferences, not requirements. Wish we could teach black/flake8 to enforce this kind of ordering easily...

var/spack/repos/builtin/packages/py-checkm-genome/package.py
```diff
@@ -13,18 +13,26 @@ class PyCheckmGenome(PythonPackage):
...
+    depends_on("python@2.7.0:2.7", type=("build", "run"), when="@:1.0.18")
+    depends_on("python@3:", type=("build", "run"), when="@1.1.0:")
     depends_on("py-setuptools", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-setuptools", type=("build", "run"))
```
`pkg_resources` is imported at run-time.

adamjstewart (changes requested):

adamjstewart (approved):

PR #33450, opened 2022-10-20
axom@0.7.0: require cmake@3.21:

eugeneswalker:
I tried to build `axom@0.7.0 ^cmake@3.20.6` and it failed:
```
...
CMake Error at CMakeLists.txt:10 (cmake_minimum_required):
  CMake 3.21 or higher is required.  You are running version 3.20.6
```

This PR makes this constraint explicit.

Worth noting that both `RAJA` and `Umpire` constrain CMake to <3.21 when building `+rocm`
* RAJA: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/raja/package.py#L70-L71
* Umpire: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/umpire/package.py#L87

Perhaps those constraints can/should be relaxed? I was able to build both `raja+rocm` and `umpire+rocm` with `cmake@3.24.2` just now.

@white238 @davidbeckingsale @wspear

### DIFF
--- a/var/spack/repos/builtin/packages/axom/package.py
+++ b/var/spack/repos/builtin/packages/axom/package.py
@@ -91,9 +91,8 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     # Dependencies
     # -----------------------------------------------------------------------
     # Basics
-    depends_on("cmake@3.8.2:", type="build")
-    depends_on("cmake@3.16.8:", type="build", when="+rocm")
-    depends_on("cmake@3.21:", type="build", when="@0.7.0")
+    depends_on("cmake@3.14:", type="build")
+    depends_on("cmake@3.21:", type="build", when="+rocm")
 
     depends_on("blt", type="build")
     depends_on("blt@0.5.1:", type="build", when="@0.6.2:")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/axom/package.py
```diff
@@ -93,6 +93,7 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     # Basics
     depends_on("cmake@3.8.2:", type="build")
     depends_on("cmake@3.16.8:", type="build", when="+rocm")
+    depends_on("cmake@3.21:", type="build", when="@0.7.0")
```
white238: This is only true `when="+rocm"`:

https://github.com/LLNL/axom/blob/e605cfcb1fce1e1a58703dbc95f4eb9b95329f99/src/CMakeLists.txt#L9-L13

Otherwise its 3.14+

white238:
```suggestion
    depends_on("cmake@3.14:", type="build")
    depends_on("cmake@3.21:", type="build", when="+rocm")
```

white238: I would have to go back through releases to find a more accurate check but I am pretty sure this is okay for quite a few releases back.

eugeneswalker: Just committed the change you suggested in other post. Thank you!

white238 (approved):
Thanks @eugeneswalker

PR #33452, opened 2022-10-20
zfp: update for version 1.0.0

GarrettDMorrison:


var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
+
     # Conflicts
     conflicts("+c", when="@:0.5.3", msg="+c requires zfp 0.5.4 or later")
-    conflicts("+python", when="@:0.5.4", msg="+python requires zfp 0.5.5 or later")
-    conflicts("+fortran", when="@:0.5.4", msg="+fortran requires zfp 0.5.5 or later")
-    conflicts("+openmp", when="@:0.5.2", msg="+openmp requires zfp 0.5.3 or later")
-    conflicts("+cuda", when="@:0.5.3", msg="+cuda requires zfp 0.5.4 or later")
+    conflicts("+daz", when="@:0.5.5", msg="+daz required zfp 1.0.0 or later")
+    conflicts("+python", when="@:0.5.4", msg="+c requires zfp 0.5.4 or later")
+    conflicts("+fortran", when="@:0.5.4", msg="+c requires zfp 0.5.4 or later")
```
lindstro: "+c requires" -> "+fortran requires"

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
+    conflicts("+openmp", when="@:0.5.2", msg="+c requires zfp 0.5.4 or later")
```
lindstro: "+c requires" -> "+openmp requires"

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
+    conflicts("+cuda", when="@:0.5.3", msg="+c requires zfp 0.5.4 or later")
```
lindstro: "+c requires" -> "+cuda requires"

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
     conflicts("+fasthash", when="@:0.5.1", msg="+fasthash requires zfp 0.5.2 or later")
     conflicts("+profile", when="@:0.5.1", msg="+profile requires zfp 0.5.2 or later")
+    conflicts("round", when="@:0.5.5", msg="+round requires zfp 1.0.0 or later")
```
lindstro: "+round requires" -> "round requires"

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
+    conflicts("+tight_error", when="@:0.5.5", msg="+tight_error requires zfp 1.0.0 or later")
```
lindstro: I would probably prefer hyphens (`tight-error`) over underscores (`tight_error`).

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
-    depends_on("py-cython", type="build", when="+python")
-
+    # CMake options
     def cmake_args(self):
         spec = self.spec
 
-        # CMake options
         args = [
-            # TODO: self.define_from_variant('BUILD_UTILITIES', 'utilities'),
             self.define("BUILD_TESTING", self.run_tests),
```
lindstro, line 129: Out of curiosity, how will this interact with the new `BUILD_TESTING_FULL` on `develop`?

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
         default=False,
         description="Use a faster but more collision prone hash function",
     )
+
     variant("profile", default=False, description="Count cache misses")
 
+    variant(
+        "round",
+        default="ZFP_ROUND_NEVER",
+        values=("ZFP_ROUND_NEVER", "ZFP_ROUND_FIRST", "ZFP_ROUND_LAST"),
```
lindstro: If it's easy enough to implement, I would prefer `round=never` over `round=ZFP_ROUND_NEVER` for improved readability and user friendliness.

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
+    )
+
+    variant(
+        "strided", default=False, description="Enable strided access for progressive zfp streams"
+    )
+
+    variant(
+        "tight_error",
+        default=False,
+        description="EXPERIMENTAL: Enable rounding-based error reduction",
```
lindstro: This description is a bit misleading as this feature itself does not reduce error; in fact, it increases the error by reducing precision without exceeding the tolerance, which is possible due to the error reduction associated with rounding.  I would suggest something like "Use tighter error bound when rounding is enabled."

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
+    variant("twoway", default=False, description="Use two-way skew-associative cache")
+
     # Conflicts
     conflicts("+c", when="@:0.5.3", msg="+c requires zfp 0.5.4 or later")
-    conflicts("+python", when="@:0.5.4", msg="+python requires zfp 0.5.5 or later")
-    conflicts("+fortran", when="@:0.5.4", msg="+fortran requires zfp 0.5.5 or later")
-    conflicts("+openmp", when="@:0.5.2", msg="+openmp requires zfp 0.5.3 or later")
-    conflicts("+cuda", when="@:0.5.3", msg="+cuda requires zfp 0.5.4 or later")
+    conflicts("+daz", when="@:0.5.5", msg="+daz required zfp 1.0.0 or later")
+    conflicts("+python", when="@:0.5.4", msg="+c requires zfp 0.5.4 or later")
```
lindstro: "+c requires" -> "+python requires"

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,82 @@ class Zfp(CMakePackage, CudaPackage):
...
+
+    variant("twoway", default=False, description="Use two-way skew-associative cache")
+
     # Conflicts
     conflicts("+c", when="@:0.5.3", msg="+c requires zfp 0.5.4 or later")
-    conflicts("+python", when="@:0.5.4", msg="+python requires zfp 0.5.5 or later")
-    conflicts("+fortran", when="@:0.5.4", msg="+fortran requires zfp 0.5.5 or later")
-    conflicts("+openmp", when="@:0.5.2", msg="+openmp requires zfp 0.5.3 or later")
-    conflicts("+cuda", when="@:0.5.3", msg="+cuda requires zfp 0.5.4 or later")
+    conflicts("+daz", when="@:0.5.5", msg="+daz required zfp 1.0.0 or later")
```
lindstro: "required" -> "requires"

lindstro (changes requested):

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -18,16 +18,18 @@ class Zfp(CMakePackage, CudaPackage):
...
     homepage = "https://zfp.llnl.gov"
-    url = "https://github.com/LLNL/zfp/releases/download/0.5.5/zfp-0.5.5.tar.gz"
+    url = "https://github.com/LLNL/zfp/releases/download/1.0.0/zfp-1.0.0.tar.gz"
     git = "https://github.com/LLNL/zfp.git"
     maintainers = ["lindstro", "GarrettDMorrison"]
     tags = ["radiuss", "e4s"]
 
     # Versions
     version("develop", branch="develop")
+    version("1.0.0", sha256="0ea08ae3e50e3c92f8b8cf41ba5b6e2de8892bc4a4ca0c59b8945b6c2ab617c4")
```
tldahlgren, line 28: Confirmed sha256.

var/spack/repos/builtin/packages/zfp/package.py
tldahlgren: You'll also want one conflict for each of the possible values, e.g.

```suggestion
    conflicts("round=ZFP_ROUND_FIRST", when="@:0.5.5", msg="+round requires zfp 1.0.0 or later")
```

Otherwise, the conflict won't trigger.

var/spack/repos/builtin/packages/zfp/package.py
GarrettDMorrison: This should act identically. Its just that prior to 1.0.0 we stripped tests out so what Spack was running is the same set of tests that `BUILD_TESTING` runs now. `develop` would have run whats in `BUILD_TESTING_FULL` but that wasn't as far as I'm aware done for any reason. It happened automatically because of how tests were managed. We could add `BUILD_TESTING_FULL` as an option but as its intended for zfp development work I'd question the need.

var/spack/repos/builtin/packages/zfp/package.py
GarrettDMorrison: Thanks for catching this. It has been a little tricky figuring out some of the finer details of non-boolean variants and associated conflicts.

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -103,28 +103,32 @@ class Zfp(CMakePackage, CudaPackage):
...
+    conflicts("+python", when="@:0.5.4", msg="+python requires zfp 0.5.4 or later")
+    conflicts("+fortran", when="@:0.5.4", msg="+fortran requires zfp 0.5.4 or later")
+    conflicts("+openmp", when="@:0.5.2", msg="+openmp requires zfp 0.5.4 or later")
+    conflicts("+cuda", when="@:0.5.3", msg="+cuda requires zfp 0.5.4 or later")
     conflicts("+fasthash", when="@:0.5.1", msg="+fasthash requires zfp 0.5.2 or later")
     conflicts("+profile", when="@:0.5.1", msg="+profile requires zfp 0.5.2 or later")
-    conflicts("round", when="@:0.5.5", msg="+round requires zfp 1.0.0 or later")
-    conflicts("+tight_error", when="@:0.5.5", msg="+tight_error requires zfp 1.0.0 or later")
+
+    conflicts("round=never", when="@:0.5.5", msg="round requires zfp 1.0.0 or later")
```
lindstro: This is the default behavior for zfp 1.0.0 and all prior versions and should probably be accepted unconditionally.

@tldahlgren Does Spack have a policy in situations like this?

tldahlgren: If this value is acceptable for versions supporting the variant, then you don't want the package to have a conflict unless the goal is to highlight (i.e., trigger an error) when the variant is used.

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,53 +73,86 @@ class Zfp(CMakePackage, CudaPackage):
...
 
-    # Dependencies
-    depends_on("cmake@3.4.0:", type="build")
-    depends_on("cuda@7:", type=("build", "test", "run"), when="+cuda")
-    depends_on("python", type=("build", "test", "run"), when="+python")
-    depends_on("py-numpy", type=("build", "test", "run"), when="+python")
-    depends_on("py-cython", type="build", when="+python")
+    conflicts("round=never", when="@:0.5.5", msg="round requires zfp 1.0.0 or later")
+    conflicts("round=first", when="@:0.5.5", msg="round requires zfp 1.0.0 or later")
+    conflicts("round=last", when="@:0.5.5", msg="round requires zfp 1.0.0 or later")
```
tldahlgren: An alternative to using `conflicts` for the values -- which will trigger an error that shows the `msg` -- is to add a `when` option to the `variant` itself to indicate when the variant applies (e.g., `when="1.0.0:").

var/spack/repos/builtin/packages/zfp/package.py
lindstro: Just to be clear, earlier versions of the zfp spack recipe did not support this variant.  The behavior prior to zfp 1.0.0 is identical to the newly introduced variant `round=never`.  zfp 1.0.0 adds new features that correspond to `round=first` and `round=last`, with `round=never` remaining the default.

The question I had is if it's considered poor practice to now allow `round=never` with older versions, which prior to zfp 1.0.0 was really the one and only rounding option, but also something users did not even know about.  I don't feel strongly one way or the other, but I suspect this issue will recur for us in the future and would like to know if there's a recommended approach to handling conflicts for new variants with respect to old packages.

var/spack/repos/builtin/packages/zfp/package.py
lindstro: Ah, that seems like a better approach.  Thanks for the suggestion.

tldahlgren: 👍 I regret not mentioning it first.

GarrettDMorrison: I've added this change with the latest fix. Let me know if you see any other issues.

var/spack/repos/builtin/packages/strumpack/package.py
```diff
@@ -70,7 +70,8 @@ class Strumpack(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("butterflypack@1.2.0:", when="@4.0.0: +butterflypack+mpi")
     depends_on("butterflypack@2.1.0:", when="@6.3.0: +butterflypack+mpi")
     depends_on("cuda", when="@4.0.0: +cuda")
-    depends_on("zfp", when="+zfp")
+    depends_on("zfp@0.5.5", when="@:7.0.1 +zfp")
+    depends_on("zfp@1.0.0", when="@7.0.2: +zfp")
```
lindstro: Should STRUMPACK 7.0.2 and later perhaps just depend on zfp (not zfp@1.0.0)?  That way they don't have to update their recipe each time there's a new zfp release.

tldahlgren: It could be `zfp@1.0.0:` if you want to make sure it is at least that version.

lindstro: Right, though strumpack@7.0.2 has been updated to handle any version of zfp.  I wouldn't require zfp@1.0.0, e.g., if it's not available.

lindstro: Come to think of it, how does spack handle dependencies on "any" version?  If, say, zfp 0.5.1 is already installed, will it match that rather than go find the latest version of zfp?

bernhardkaindl: The process to to arrive on a fixed set of package versions in a build is the concretization step.
It can be adjusted and e.g. `spack install --reuse` tries to concretize by reusing already installed or built dependencies.

@lindstro: AFAICS, the last commit resolved this review comment.

var/spack/repos/builtin/packages/zfp/package.py
bernhardkaindl: @lindstro: I'm not aware of a recommend approach in this regard. Do you approve this PR now?

bernhardkaindl (dismissed):
Awaiting approval by @lindstro for a couple of days. ping me or others if there is no response.

When the review has passed, I suggest updating to the latest develop branch.
It will also trigger a re-run of the ci/gitlab-ci pipeline with the latest fixes which may help to pass it.

var/spack/repos/builtin/packages/zfp/package.py
robertu94: Isn't there the ability to add a when clause to a variant?  It seems like the answer here is to make the round variant when="@1.0.0:" and not bother with a conflicts clause

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,55 +73,86 @@ class Zfp(CMakePackage, CudaPackage):
         "use smaller for finer rate granularity. "
         "Use 8 for H5Z-ZFP filter.",
     )
+
     variant(
-        "strided", default=False, description="Enable strided access for progressive zfp streams"
+        "daz",
```
lindstro, line 78: We should add `when="@1.0.0:"` here, just like with the `round` variant.

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,55 +73,86 @@ class Zfp(CMakePackage, CudaPackage):
...
+    )
+
+    variant("twoway", default=False, description="Use two-way skew-associative cache")
+
     # Conflicts
     conflicts("+c", when="@:0.5.3", msg="+c requires zfp 0.5.4 or later")
-    conflicts("+python", when="@:0.5.4", msg="+python requires zfp 0.5.5 or later")
-    conflicts("+fortran", when="@:0.5.4", msg="+fortran requires zfp 0.5.5 or later")
-    conflicts("+openmp", when="@:0.5.2", msg="+openmp requires zfp 0.5.3 or later")
+    conflicts("+daz", when="@:0.5.5", msg="+daz requires zfp 1.0.0 or later")
```
lindstro: Let's remove this conflict and handle it in the variant with `when`.

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,55 +73,86 @@ class Zfp(CMakePackage, CudaPackage):
...
+    variant(
+        "tight-error",
+        default=False,
+        when="@1.0.0:",
+        description="EXPERIMENTAL: Use tighter error bound when rounding is enabled",
+    )
+
+    variant("twoway", default=False, description="Use two-way skew-associative cache")
+
     # Conflicts
```
lindstro, line 117: I would suggest we handle these "conflicts" via `when` in the variants, just like we're doing with the new `round` variant, etc.  The main conflict appears to be combinations of variants, like with `round` and `tight-error`.

lindstro (commented):
Please see comments on substituting conflicts with `when` in variants.

var/spack/repos/builtin/packages/zfp/package.py
lindstro: Yes, we have done so with some variants but not consistently with all of them.  I agree with your suggestion.

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -62,55 +73,80 @@ class Zfp(CMakePackage, CudaPackage):
...
+            self.define_from_variant("ZFP_WITH_DAZ", "daz"),
             self.define_from_variant("ZFP_WITH_CACHE_FAST_HASH", "fasthash"),
             self.define_from_variant("ZFP_WITH_CACHE_PROFILE", "profile"),
+            self.define_from_variant("ZFP_WITH_BIT_STREAM_STRIDED", "strided"),
+            self.define_from_variant("ZFP_WITH_TIGHT_ERROR", "tight-error"),
+            self.define_from_variant("ZFP_WITH_CACHE_TWOWAY", "twoway"),
         ]
 
+        if "round" in spec.variants:
+            args.append("ZFP_ROUNDING_MODE={0}".format(spec.variants["round"].value.upper()))
```
lindstro: I may be missing something obvious here, but don't we need to prepend `ZFP_ROUND_` to the value after conversion to upper case?  For example, Spack's `round=never` should translate to CMake's `ZFP_ROUNDING_MODE=ZFP_ROUND_NEVER`.

lindstro (dismissed):

GarrettDMorrison:
As @lindstro has approved the latest changes I am rebasing to sync with develop

lindstro (approved):

bernhardkaindl:
Merged develop again as a job of the GitLab CI/CD pipeline failed, apparently for an unrelated reason.

bernhardkaindl (approved):
Approving because @lindstro reviewed and approved the latest update.

tldahlgren:
Some container issues reported.

@spackbot run pipeline

tldahlgren (approved):

PR #33460, opened 2022-10-21
Relocation should take hardlinks into account

### LINKED ISSUE #33453: Binary relocation breaks hardlinked files
See https://github.com/spack/spack/pull/33443#issuecomment-1286198889 there was a spurious trailing line in `bin/automake`, with just

```
$exit_status;
```

right after what is normally the last line:

```perl
exit $exit_status;
```

leading to a `Useless use of private variable in void context` error when running `automake`. I could reproduce it once locally. But not a second time. @trws mentioned he had seen this in the past too

I diffed the file vs a local automake, and it's only different on this trailing line, and on some line mentioning the install prefix.

Maybe there's some issue with relocation logic and hardlinks (given that relocation runs in parallel?), since `automake` is a hardlink of `automake-<version>`. Or maybe it has to do with sbang, since it's a shell script?

---

Managed to reproduce it another time, and it's again the exact same:

```perl
exit $exit_code;
$exit_code;
```

To reproduce, I tried this about 5 times or so:

```console
$ git config --global --add safe.directory /builds/spack/spack/spac
$ source /builds/spack/spack/spack/share/spack/setup-env.sh
$ spack env activate /builds/spack/spack/jobs_scratch_dir/reproduction/

# repeat the following until failure
$ spack -c config:install_tree:root:/home/software/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeh/ uninstall --yes automake && spack clean -ds && /builds/spack/spack/jobs_scratch_dir/reproduction/install.sh
```

---

Retried it 3 more times, and got yet another failure:

```
syntax error at /home/software/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeh/linux-ubuntu20.04-x86_64/oneapi-2022.1.0/automake-1.16.5-uzqwcqlszsuh5z5g27yrdekgsoyldisj/bin/aclocal line 99, near "appear last in"
```

and there's again an odd repetition, now at line 99 of 1200+ (maybe relevant, the newline before the repetition is byte 4096):

```perl
# Remember the order into which we scanned the files.
# It's important to output the contents of aclocal.m4 in the opposite order.
# (Definitions in first files we have scanned should override those from
# later files.  So they must appear last in the output.)
ey must appear last in the output.)
my @file_order = ();
```

aclocal is also a hardlink, if that's relevant:

```
17451579 /home/software/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeh/linux-ubuntu20.04-x86_64/oneapi-2022.1.0/automake-1.16.5-uzqwcqlszsuh5z5g27yrdekgsoyldisj/bin/aclocal
17451579 /home/software/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeh/linux-ubuntu20.04-x86_64/oneapi-2022.1.0/automake-1.16.5-uzqwcqlszsuh5z5g27yrdekgsoyldisj/bin/aclocal-1.16
```

looks like a race then?

---

automated this with

```
#!/usr/bin/env spack-python

import spack.environment
import spack.store
from spack.util.executable import Executable

success = 0

for i in range(100):
    with spack.environment.Environment("/builds/spack/spack/jobs_scratch_dir/reproduction") as e:
        spec = e.get_by_hash("uzqwcqlszsuh5z5g27yrdekgsoyldisj")[0]
        spec.package.do_install(install_deps=False, unsigned=True, dependencies_cache_only=True)
        automake = Executable(spec.prefix.bin.automake)
        automake("--version", fail_on_error=False, stdout=str)
        if automake.returncode == 0:
            success += 1
        print("success rate", success, "/", i + 1)

        spec.package.do_uninstall()
```

after setting up `file://` based mirror, and it failed 10/100 times for me. When removing `bin/automake-1.16` from the list of files to relocate, keeping `bin/automake` there, it succeeded 100/100 times. So yes, hardlinks are the issue.

haampie:
Closes #33453

Currently `relocate_text` and `relocate_text_bin` are unsafe in the
sense that they run in parallel, and lead to races when modifying
different items pointing to the same inode.

This leads to the issue observed in #33453, where `automake`
is broken 10% of the time due to this race.

This PR:

1. Renames those functions to `unsafe_*` so people are aware
2. Adds logic to deal with hardlinks in currently available binary packages
3. Adds logic to deal with hardlinks when creating new binary tarballs,
   so the install side doesn't have to de-dupe hardlinks.
4. Adds a test for 3

The assumption is that all our relocation logic preserves inodes. That
is, we should never copy a file, modify it, and then move it back. I
quickly verified, and its seems like this is true for (binary) text
relocation, as well as rpath patching in patchelf (even when the file
grows) and mach-o fixes.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -642,8 +642,8 @@ def read_buildinfo_file(prefix):
     return buildinfo
 
 
-class BuildManifestVistor(BaseDirectoryVisitor):
-    """Visitior that collects a list of files and symlinks
+class BuildManifestVisitor(BaseDirectoryVisitor):
+    """Visitor that collects a list of files and symlinks
     that can be checked for need of relocation. It knows how
     to dedupe hardlinks and deal with symlinks to files and
     directories."""
@@ -708,7 +708,7 @@ def get_buildfile_manifest(spec):
     # Guard against filesystem footguns of hardlinks and symlinks by using
     # a visitor to retrieve a list of files and symlinks, so we don't have
     # to worry about hardlinks of symlinked dirs and what not.
-    visitor = BuildManifestVistor()
+    visitor = BuildManifestVisitor()
     root = spec.prefix
     visit_directory_tree(root, visitor)
 
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -23,7 +23,7 @@ from six.moves.urllib.error import HTTPError, URLError
 import llnl.util.filesystem as fsys
 import llnl.util.lang
 import llnl.util.tty as tty
-from llnl.util.filesystem import mkdirp
+from llnl.util.filesystem import BaseDirectoryVisitor, mkdirp, visit_directory_tree
 
 import spack.cmd
 import spack.config as config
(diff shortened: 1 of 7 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -642,6 +642,51 @@ def read_buildinfo_file(prefix):
     return buildinfo
 
 
+class BuildManifestVistor(BaseDirectoryVisitor):
```
zackgalbreath:
```suggestion
class BuildManifestVisitor(BaseDirectoryVisitor):
```

lib/spack/spack/binary_distribution.py
```diff
@@ -642,6 +642,51 @@ def read_buildinfo_file(prefix):
...
+    """Visitior that collects a list of files and symlinks
```
zackgalbreath:
```suggestion
    """Visitor that collects a list of files and symlinks
```

zackgalbreath (changes requested):
A couple typos, otherwise this LGTM! Thanks for investigating this issue & coming up with a fix so quickly.

lib/spack/spack/binary_distribution.py
haampie: it's french for visitor ;p

haampie:
We currently _assume_ in-place updates to files, this is a convention we could make a hard requirement by adding a pass over the install prefix before & after relocation -- but right now let's keep it a convention.

zackgalbreath (approved):

PR #33465, opened 2022-10-21
py-instrain: added required + optional dependency

luke-dt:


var/spack/repos/builtin/packages/py-instrain/package.py
```diff
@@ -35,3 +37,6 @@ class PyInstrain(PythonPackage):
     depends_on("py-psutil", type=("build", "run"))
     depends_on("py-lmfit", type=("build", "run"))
     depends_on("py-numba", type=("build", "run"))
```
adamjstewart, line 39: Can you add a comment with the following link:
https://instrain.readthedocs.io/en/latest/installation.html#dependencies

It's always hard to tell where non-Python deps come from since they aren't documented in `setup.py`

adamjstewart (approved):

PR #33466, opened 2022-10-21
updated python version requirements

luke-dt:
the py-biopython package was previously limited to python 3, which was preventing certain packages from building (see PR #33404). Added support for python 2 and updated python requirements for the different versions (according to the setup.py)

### DIFF
--- a/var/spack/repos/builtin/packages/py-biopython/package.py
+++ b/var/spack/repos/builtin/packages/py-biopython/package.py
@@ -10,11 +10,10 @@ class PyBiopython(PythonPackage):
     """A distributed collaborative effort to develop Python libraries and
     applications which address the needs of current and future work in
     bioinformatics.
-
     """
 
     homepage = "https://biopython.org/wiki/Main_Page"
-    url = "https://biopython.org/DIST/biopython-1.65.tar.gz"
+    pypi = "biopython/biopython-1.79.tar.gz"
 
     version("1.79", sha256="edb07eac99d3b8abd7ba56ff4bedec9263f76dfc3c3f450e7d2e2bcdecf8559b")
     version("1.78", sha256="1ee0a0b6c2376680fea6642d5080baa419fd73df104a62d58a8baf7a8bbe4564")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-biopython/package.py
```diff
@@ -10,18 +10,21 @@ class PyBiopython(PythonPackage):
...
 
     homepage = "https://biopython.org/wiki/Main_Page"
-    url = "https://biopython.org/DIST/biopython-1.65.tar.gz"
+    pypi = "biopython/biopython-1.79.tar.gz"
 
     version("1.79", sha256="edb07eac99d3b8abd7ba56ff4bedec9263f76dfc3c3f450e7d2e2bcdecf8559b")
     version("1.78", sha256="1ee0a0b6c2376680fea6642d5080baa419fd73df104a62d58a8baf7a8bbe4564")
     version("1.73", sha256="70c5cc27dc61c23d18bb33b6d38d70edc4b926033aea3b7434737c731c94a5e0")
     version("1.70", sha256="4a7c5298f03d1a45523f32bae1fffcff323ea9dce007fb1241af092f5ab2e45b")
     version("1.65", sha256="463cc81db84e9bfcdfb15629511c81ed556a6c0287e670dbfe80f03c65d2a88e")
```
adamjstewart:
```suggestion
    version("1.65", sha256="6d591523ba4d07a505978f6e1d7fac57e335d6d62fb5b0bcb8c40bdde5c8998e")
```

adamjstewart (changes requested):
Everything else looks good, thanks!

adamjstewart (approved):

adamjstewart:
Note that the next Spack release officially deprecates installing Python 3.6 and older, and the release after that will remove these versions entirely. Still doesn't mean the package recipes can't be correct, just warning you that you won't be able to install Python 2 in the near future.

luke-dt:
> Note that the next Spack release officially deprecates installing Python 3.6 and older, and the release after that will remove these versions entirely. Still doesn't mean the package recipes can't be correct, just warning you that you won't be able to install Python 2 in the near future.

Thanks for the heads up! We may then keep an older spack install for the handful of things things that need python 2 then.

adamjstewart:
Out of curiosity, how many packages still require Python 2? Are these libraries still maintained or is there anyone actively working on porting to Python 3? I realize that there are many libraries that are still used but no longer maintained, but I'm hoping that your suggestion of using older Spack should work for these.

P.S. See #31824 for discussion on this decision.

luke-dt:
I'd have to check, but I could probably count them on one hand. It's mostly old scripts that we have used in the past that work but aren't being maintained (like shortbred).

adamjstewart:
I've had a couple of packages where I was able to run `2to3` on them to get them working, really depends on how complicated the package is and whether it requires ctypes.

PR #33467, opened 2022-10-21
py-drep and ANIcalculator: new packages

luke-dt:


var/spack/repos/builtin/packages/anicalculator/package.py
```diff
@@ -0,0 +1,37 @@
...
+    manual_download = True
+
+    version("1", sha256="236596a9a204cbcad162fc66be3506b2530b1f48f4f84d9647ccec3ca7483a43")
+
+    depends_on("perl@5:")
+
+    conflicts("platform=darwin", msg="ANIcalculator requires Linux")
+
+    def url_for_version(self, version):
+        return "file://{0}/ANIcalculator_v{1}.tgz".format(os.getcwd(), version)
```
adamjstewart: You can prob do this with a `url`, I don't think `url_for_version` is necessary

var/spack/repos/builtin/packages/anicalculator/package.py
```diff
@@ -0,0 +1,37 @@
...
+    Alternatively, add this file to a mirror so that Spack can find it.
+    For instructions on how to set up a mirror, see
+    https://spack.readthedocs.io/en/latest/mirrors.html"""
+
+    homepage = "https://ani.jgi.doe.gov/html/download.php?"
+    manual_download = True
+
+    version("1", sha256="236596a9a204cbcad162fc66be3506b2530b1f48f4f84d9647ccec3ca7483a43")
+
+    depends_on("perl@5:")
```
adamjstewart: The default deptype is build/link. Since this package isn't compiled, I don't think it's being used. Should it be "run"?

luke-dt: yes, that's probably correct. I didn't realize that was the default dependency type. I'll test it with that change, I previously just had it print the help message but didn't put any data through it.

var/spack/repos/builtin/packages/anicalculator/package.py
luke-dt: Do you mean nixing the manual download? I wasn't sure if putting this on spack was a violation of the software license, so I figured the safest option was to have the user manually agree and download.

adamjstewart: I mean you could probably do something like:
```python
url = "file://{0}/ANIcalculator_v1.tgz".format(os.getcwd())
```
so you don't need to define any function.

var/spack/repos/builtin/packages/anicalculator/package.py
luke-dt: tested with real data and it works

adamjstewart (approved):

Issue #33461, opened 2022-10-21
Spack ignores `buildable=False` if available in binary cache


stephenmsachs:
### Steps to reproduce

It seems availaibility in binary cache is valued higher than the `buildable` flag by the concretizer.

## Reproducer
Use `packages.yaml` as
```yaml
packages:
  libfabric:
    externals:
    - spec: libfabric@1.16.0 fabrics=efa
      modules:
      - libfabric-aws/1.16.0
    buildable: false
``` 

The command will show that the extrnal defined libfabric version is used:
```console
spack spec openmpi fabrics=ofi | grep libfabric
```
Output:
```
    ^libfabric@1.16.0%gcc@7.3.1~debug~kdreg fabrics=efa,sockets,tcp,udp arch=linux-amzn2-skylake_avx512
```

Now add the binary mirror as:
```yaml
mirrors:
  binary_mirror: https://binaries.spack.io/develop
```

and the following line will produce an older version of libfabric which is available in the binary cache:
```console
spack buildcache keys --install --trust
spack spec openmpi fabrics=ofi | grep libfabric
```

Output:
```
 -       ^libfabric@1.14.1%gcc@7.3.1~debug~kdreg fabrics=efa,mrail,rxd,rxm,shm,sockets,tcp,udp,verbs arch=linux-amzn2-x86_64_v4
```

### Information on your system

* **Spack:** 0.19.0.dev0 (b1559cc831620ee2b2cf8e57fdecc5bb3bf8edfd)
* **Python:** 3.7.10
* **Platform:** linux-amzn2-skylake_avx512
* **Concretizer:** clingo

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

stephenmsachs:
After reading some of the code it seems to me that installed specs and specs available in the binary cache are treated the same way. If the `reuse` flag is set, the concretizer wil try to fulfill the request from installed and binary cache specs first before using uninstalled specs. 

I would argue this behavior is not optimal. I would assume binary cache objects to be treated with the same level of priority as installable specs. 

Removing line https://github.com/spack/spack/blob/7c3d93465c1d76e084d6b0ce4a44d1c0b4b1ae68/lib/spack/spack/solver/asp.py#L2354 shows the desired behavior.

alalazo:
[...]

PR #33468, opened 2022-10-21
Respect the `buildable` attribute for packages in binary cache

stephenmsachs:
When it comes to concretizing specs, packages in binary cache are treated in the same way as already installed packages. This leads to the fact that packages protected by `buildable:false` are still downloaded and installed from binary cache. In https://github.com/spack/spack/issues/33461 this even leads to preferring an older version of a package because it available in binary cache over a newer, locally installed one.

This patch filters out any packages that are locally protected by `buildable:false` before the list of packages available in binary cache gets sent as "reusable" to the concretizer.

### DIFF
--- a/lib/spack/spack/solver/asp.py
+++ b/lib/spack/spack/solver/asp.py
@@ -2335,7 +2335,8 @@ class Solver(object):
         return reusable
 
     def filter_buildable(self, index):
-        """Filter out buildcache entries of specs that match packages locally marked as buildable:false"""
+        """Filter out buildcache entries of specs that match packages locally marked as
+        buildable:false"""
         packages_yaml = spack.config.get("packages")
         filter = ()
         for pkg_name, data in packages_yaml.items():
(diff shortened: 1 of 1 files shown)


haampie:
I believe this was discussed before and then it was said that "technically it's not being built but downloaded", but maybe this was more of a joke. I agree that `buildable: false` is a signal to the concretizer that this package is always external (maybe force: true was a better name...)

alalazo:
> I believe this was discussed before and then it was said that "technically it's not being built but downloaded", but maybe this was more of a joke. 

I don't think it was. The config option was there to ensure Spack didn't go through long source builds iirc, and "reuse" was not even a concept we considered at the time. We probably need to give user more control of what gets reused, not change the meaning of what currently "reuse" does?

alalazo:
Superseded by https://github.com/spack/spack/pull/42782 and following PRs

PR #33470, opened 2022-10-21
VTK-h: honor global spack flags

white238:
Verified this by setting a noop flag and finding it both on the compile line and CMake's output.


cyrush (approved):

eugeneswalker (approved):

tldahlgren (approved):
LGTM

PR #33063, opened 2022-10-06
Remove support for running with Python 2.7

alalazo:
This PR removes support for running Spack with Python 2.7. Should be merged after the v0.19 release.

Modifications:
- [x] Remove all the tests in CI using or referring to Python 2.7
- [x] Remove all the branches in the code specific to Python 2.7
- [x] Remove vendored dependencies used only with Python 2.7

PR #33474, opened 2022-10-22
Don't install xdist in CI on Python 2.7

alalazo:
xdist + cov on Python 2.7 is causing the CI jobs to hang more frequently than what we were used to. While we wait for #33063 to be mergeable (i.e. while we wait for v0.19 to be tagged), we can just disable parallelism for Python 2.7

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -52,7 +52,12 @@ jobs:
               patchelf cmake bison libbison-dev kcov
     - name: Install Python packages
       run: |
-          pip install --upgrade pip six setuptools pytest codecov[toml] pytest-cov pytest-xdist
+          pip install --upgrade pip six setuptools pytest codecov[toml] pytest-cov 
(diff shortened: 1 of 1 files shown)


haampie (approved):

trws:
If xdist is off, how about turning off coverage for 2.7 as well? That will offset the performance degradation, and as you say it’s basically vestigial at this point anyway.

alalazo:
> If xdist is off, how about turning off coverage for 2.7 as well? 

Two thoughts:
1. The issue seems to be due to using xdist + cov. If we turn off coverage, we might want to turn xdist back on
2. Maybe we want to keep coverage in the v0.19 branch, to ensure backported commits still work for Python 2.7. If that is the case I'd just try to see if the current state of things is bearable for another 2-3 weeks.

trws:
What does coverage have to do with ensuring it works? If it doesn’t gum up the CI too bad it might be fine for a few weeks, but if it does dropping test coverage (but not tests) seems like a relatively cheap way to unblock it.

PR #33477, opened 2022-10-23
locks: improved errors

haampie:
Instead of showing

```
==> Error: Timed out waiting for a write lock.
```

show

```
==> Error: Timed out waiting for a write lock after 1.200ms and 4 attempts on file: /some/file
```

s.t. we actually get to see where acquiring a lock failed even when not
running in debug mode.

And use pretty time units everywhere, so we don't get 1.45e-9 seconds
but 1.450ns etc.

### DIFF
--- a/lib/spack/llnl/util/lock.py
+++ b/lib/spack/llnl/util/lock.py
@@ -12,6 +12,7 @@ from datetime import datetime
 from typing import Dict, Tuple  # novm
 
 import llnl.util.tty as tty
+from llnl.util.lang import pretty_seconds
 
 import spack.util.string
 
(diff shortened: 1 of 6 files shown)


haampie:
@cosmicexplorer friendly reminder to fix your `test_grouped_exception` test :grimacing:

haampie:
@spackbot run pipeline

lib/spack/llnl/util/lock.py
```diff
@@ -780,6 +782,12 @@ class LockLimitError(LockError):
 class LockTimeoutError(LockError):
     """Raised when an attempt to acquire a lock times out."""
 
+    def __init__(self, lock_type, path, time, attempts):
+        msg = "Timed out waiting for a {} lock after {} and {} attempts on file: {}".format(
+            lock_type, pretty_seconds(time), attempts, path
+        )
+        super(LockTimeoutError, self).__init__(msg)
```
tgamblin: Can you pass the attempts and filename as a second argument?  That'll make it look like this:

```console
==> Error: Timed out waiting for a write lock after 1.200ms.
    Made 4 attempts on file: /path/to/file
```

which means the path is less likely to wrap around (or to wrap around as much, anyway).

tgamblin (changes requested):
LGTM -- minor request

tgamblin (approved):

PR #33481, opened 2022-10-24
dfs traversal: simplify edges in reverse mode

haampie:
In the dfs code, flip edges so that `parent` always means `from` and
`spec` means `to` _in the direction of traversal_. This makes it slightly
easier to write generic/composable code. For example when using visitors
where one visitor reverses direction, and another only cares about
accepting particular edges or not depending on whether the `to` node
is seen before, it would be good if the second visitor didn't have to
know whether the order was changed or not.

As far as I can see, we never use `traverse_edges(direction="parents")`, we
only ever use `traverse(direction="parents")` which yields nodes not edges. So
should be fine to flip those edges.

This change should help unify the code of #33406, since we can then
reuse `ReverseVisitor` / `CoverNodesVisitor` / `CoverEdgesVisitor` in
DFS and BFS.

### DIFF
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -734,6 +734,9 @@ class DependencySpec(object):
     def canonical(self):
         return self.parent.dag_hash(), self.spec.dag_hash(), self.deptypes
 
+    def flip(self):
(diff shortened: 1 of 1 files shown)


tgamblin (approved):

tgamblin:
LGTM -- this simplifies the traversal cases quite a bit.

PR #33483, opened 2022-10-24
Remove recursive symbolic link in lib/spack/docs from git repository

alalazo:
The symbolic link creates issues with IDE indexers, since it is recursive. Create it lazily when building docs, since it doesn't need to be stored in the git repository.


lib/spack/docs/conf.py
```diff
@@ -32,6 +32,9 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
+link_name = os.path.abspath("_spack_root")
+if not os.path.exists(link_name):
+    os.symlink(os.path.abspath("../../.."), link_name, target_is_directory=True)
```
trws, line 37: Does this mean building the docs will put it back?

alalazo: Yes, I maintained that structure. I can have a look to see if that's really necessary. `git clean` will remove it though.

alalazo: It seems it's needed to be able to include literals that are two or three levels up. I can follow with a refactoring to move docs in the root directory, instead of `lib/spack/docs`, but I'd prefer to do that in another PR.

haampie (approved):
Probably we can do even better, but this looks like an improvement to me.

PR #33495, opened 2022-10-24
Lock, database and store don't need global configuration on construction

alalazo:
Modifications:
- [x] Lock objects can now be instantiated independently, without being tied to the global configuration. The same is true for database and store objects.
- [x] The database and store `__init__` methods have been simplified to take a single lock configuration object. Some common lock configurations[^1] have been named and are provided as globals.
- [x] The `use_store` context manager keeps the configuration consistent by pushing and popping an internal scope. It can also be tuned by passing extra data to set up e.g. upstreams or anything else that might be related to the store.
- [x] Unit tests have been added to ensure that database, locks and repos can be created without the need to access any  global variable.

[^1]: For instance `spack.database.NO_LOCK` or `spack.database.NO_TIMEOUT`

### DIFF
--- a/lib/spack/spack/database.py
+++ b/lib/spack/spack/database.py
@@ -253,8 +253,11 @@ class ForbiddenLockError(SpackError):
 
 
 class ForbiddenLock:
-    def __getattribute__(self, name):
+    def __getattr__(self, name):
         raise ForbiddenLockError("Cannot access attribute '{0}' of lock".format(name))
+
+    def __reduce__(self):
+        return ForbiddenLock, tuple()
 
 
 _QUERY_DOCSTRING = """
--- a/lib/spack/spack/database.py
+++ b/lib/spack/spack/database.py
@@ -18,7 +18,7 @@ as the authoritative database of packages in Spack.  This module
 provides a cache and a sanity checking mechanism for what is in the
 filesystem.
 """
-
+import collections
 import contextlib
 import datetime
 import os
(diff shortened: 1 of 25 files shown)


lib/spack/spack/database.py
```diff
@@ -116,7 +111,7 @@ def reader(version):
     return reader_cls[version]
 
 
-def _now():
+def _now() -> float:
```
haampie, line 114: not sure if this is helpful, but well. (Ok, maybe for mocking)

lib/spack/spack/database.py
```diff
@@ -308,73 +306,106 @@ def __getattribute__(self, name):
...
+    enable=True,
+    database_timeout=_DEFAULT_DB_LOCK_TIMEOUT,
+    package_timeout=_DEFAULT_PKG_LOCK_TIMEOUT,
+)
+
+
+def lock_configuration(configuration):
+    """Return a LockConfiguration from a spack.config.Configuration object."""
+    return LockConfiguration(
+        enable=configuration.get("config:locks", None),
```
haampie: `enable` should be bool, use `True`?

haampie:
Generally looks good to me, not sure though why mypy doesn't catch the type issue for `enable`.

lib/spack/spack/database.py
alalazo: Done

haampie (approved):

PR #33496, opened 2022-10-24
Relocation regex single pass

haampie:
Instead of looping the entire file for each prefix, create a giant
regex with all literal prefixes and do a single pass. Not only is a
single pass faster in IO, it's also likely that the regex does fewer
ops per byte in the file, given that most prefixes share a common
... prefix.

Note: this just applies to the @iarspider logic of scanning text
files before creating the tarball for a buildcache. Something
similar could potentially be done during install from the buildcache.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -42,7 +42,7 @@ import spack.util.spack_yaml as syaml
 import spack.util.url as url_util
 import spack.util.web as web_util
 from spack.caches import misc_cache_location
-from spack.relocate import utf8_path_to_binary_regex, utf8_paths_to_single_binary_regex
+from spack.relocate import utf8_paths_to_single_binary_regex
 from spack.spec import Spec
 from spack.stage import Stage
 
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -42,7 +42,7 @@ import spack.util.spack_yaml as syaml
 import spack.util.url as url_util
 import spack.util.web as web_util
 from spack.caches import misc_cache_location
-from spack.relocate import utf8_path_to_binary_regex
+from spack.relocate import utf8_path_to_binary_regex, utf8_paths_to_single_binary_regex
 from spack.spec import Spec
 from spack.stage import Stage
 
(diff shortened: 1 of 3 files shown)


iarspider (dismissed):

iarspider (approved):

iarspider:
Out of curiosity - did you make any actual time measurements? The file is read once, and is scanned "lazily" (once we have a hit, we return immediately).

haampie:
It's mostly useful for negative results, so for example if I take the 256KB `automake` script and drop the matching prefix it's 13x faster with 14 prefixes searched for:

```python
In [12]: %timeit file_matches_any_binary_regex("automake", compiled_prefixes)
126 ms ± 434 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [13]: %timeit file_matches("automake", compiled_prefix)
9.45 ms ± 12.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

For plain text files you should expect a speedup of # prefixes, since the lookbehind and almost all prefixes have the same initial characters, so if that matches nearly nothing for the first prefix, it won't match much for the remaining prefixes either; so effectively it would be almost O(n) work instead of O(# prefixes * n) where n is the bytes in the file.

And actually since we add `prefixes.append(str(spack.store.layout.root))` which is almost always a prefix of all other paths, the effective regex is roughly the work of scanning for a single prefix in case we do get matches. 

If I do include the first prefix verbatim somewhere at the top of the file:

```python
In [16]: %timeit file_matches_any_binary_regex("automake", compiled_prefixes)
63.2 µs ± 523 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [17]: %timeit file_matches("automake", compiled_prefix)
64.8 µs ± 172 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

But we care about the negative results, since they are slowest.

Another optimization would be to just test the shebang line before even reading the whole file, but well. It's more important to improve the regexes for replacement in binaries, since there the number of bytes can be orders of magnitude larger.

iarspider:
@tldahlgren I don't have merge access, so assigning me makes 0 sense.

alalazo (approved):

tldahlgren:
> @tldahlgren I don't have merge access, so assigning me makes 0 sense.

We often leverage the `Assignees` field to indicate someone has commented on the PR.  It makes it easier to tell from the list which PRs have more than just spackbot-generated comments.  

I can try to remember not to add you if that definition of `Assignee` bothers you.  I sometimes get on a roll when I'm trying to chip away at the PR list so may not always remember.

iarspider:
> > @tldahlgren I don't have merge access, so assigning me makes 0 sense.
> 
> We often leverage the `Assignees` field to indicate someone has commented on the PR. It makes it easier to tell from the list which PRs have more than just spackbot-generated comments.
> 
> I can try to remember not to add you if that definition of `Assignee` bothers you. I sometimes get on a roll when I'm trying to chip away at the PR list so may not always remember.

Ah, I see. No, it doesn't bother me, just don't want a PR to be stuck because someone expects me to merge it

Issue #4432, opened 2017-06-05
How should spack deal with packages that need to be installed as root?


krafczyk:
Some software packages such as `singularity` (currently working on adding it) need to be installed as root in order for them to work properly.

What should be the strategy spack takes for these packages?

Something which makes sense to me is to add a boolean member variable to the package class like `install_as_root`. This then be inspected on install to determine whether `sudo` should be invoked to install the package. Of course this scheme will only work with pre-defined build systems like `AutotoolsPackage` for `CMakePackage`, and ad-hoc build systems will be left out.

adamjstewart:
Are you talking about [this singularity](http://singularity.lbl.gov/install-linux)? If so, I don't think you have to install it with root privileges. You only have to use `sudo` if you want to install it to `/usr/local`; Spack will install it to its own installation prefix. I've never seen a software package that couldn't be installed without root. If there were, how would developers test it?

krafczyk:
I am referring to that singularity.

You can install it without sudo, and create images, however if you want to run images as your default user and not the root user, singularity needs to be installed with root.

samfux84:
@adamjstewart We recently installed Singularity on one node in our cluster for having a test environment to play with it. In order to be usable, one needs to run the "make install" step with sudo, because the software contains a setuid binary.

From the documentation about the singularity config file, http://singularity.lbl.gov/docs-config:

```
ALLOW SETUID (boolean, default=’yes’)

This parameter toggles the global ability to execute the SETUID (SUID) portion of the code if it exists. As mentioned earlier, if the SUID features are disabled, various Singularity features will not function (e.g. mounting of the Singularity image file format).

You can however disable SUID support iff (if and only if) you do not need to use the default Singularity image file format and if your kernel supports user namespaces and you choose to use user namespaces.

note: as of the time of this writing, the user namespace is rather buggy
```

alalazo:
> Some software packages such as singularity (currently working on adding it) need to be installed as root in order for them to work properly.

Another package of this kind is [likwid](https://github.com/RRZE-HPC/likwid)

krafczyk:
@adamjstewart: what @samfux84 said.

adamjstewart:
I see. This seems like it won't be very common, and isn't particularly specific to a single build system. I haven't tried it, but does the following code work?
```python
def install(self, spec, prefix):
    # Must be installed as root to work properly
    sudo = which('sudo')
    sudo('make', '-j{0}'.format(make_jobs), 'install')
```

[...]

PR #33503, opened 2022-10-25
likwid: add a permission fixing script à la singularity

matz-e:
Removes the install glue that changes ownership and add a script like
for singularity that will fix ownership and permission. Log instructions
to the build output.


TomTheBear:
Why? You replaced 3 lines of simple file filtering with around 50 of Python code with code generation of an additional script that will contain only 4 lines with `chmod` and `chown` in any case!

Have there been cases where the changing of ownership failed/caused problems? I have not seen any issues or have been notified of any.

It might be related to the installation of LIKWID as root with the `accessdaemon` backend to get all features. But you cannot select the backend anymore (`ACCESSMODE` variable), so all users end up getting the default `accessdaemon` backend. If they cannot execute the "change permissions script" with `sudo` in the end, they have a not usable LIKWID installation and cannot do anything about it except deleting it.

If you want to add logic to the package, add a variant to select the access mode:
- If `ACCESSMODE=perf_event`, no script generation needed, installation as user, maybe tell users that their scope of access is limited by the global `perf_event_paranoid` file (`BUILDFREQ=false` and `BUILDDAEMON=false`)
- If `ACCESSMODE=direct`, print a warning that you need root permissions when using LIKWID and no script generation needed (`BUILDFREQ=false` and `BUILDDAEMON=false`)
- if `ACCESSMODE=accessdaemon`, create your script and print the "run this script with sudo" instructions (`BUILDFREQ=true` and `BUILDDAEMON=true`)

This way, the users can select what they want depending on their permissions.

matz-e:
> Why? You replaced 3 lines of simple file filtering with around 50 of Python code with code generation of an additional script that will contain only 4 lines with `chmod` and `chown` in any case!

Both `singularity*` and `gaussian-*` do this. Maybe the example I lifted from `singularity` is a bit more verbose than needed, and it seems that maybe some of these things could be de-duplicated in the future.

> Have there been cases where the changing of ownership failed/caused problems? I have not seen any issues or have been notified of any.

We like using likwid on a few special nodes on our cluster. Due to "security", we cannot get access as `root`. Everything builds fine, but the _installation_ step fails at the `install -o root -g root <daemon_executable>` :shrug:

On my personal machine, I'm not going to do a `sudo spack install likwid` to get around this, so IMO whenever these daemons are installed, it's better to just fix the permissions with the right user afterwards.
 
> If you want to add logic to the package, add a variant to select the access mode:
> 
>     * If `ACCESSMODE=perf_event`, no script generation needed, installation as user, maybe tell users that their scope of access is limited by the global `perf_event_paranoid` file (`BUILDFREQ=false` and `BUILDDAEMON=false`)
> 
>     * If `ACCESSMODE=direct`, print a warning that you need root permissions when using LIKWID and no script generation needed (`BUILDFREQ=false` and `BUILDDAEMON=false`)
> 
>     * if `ACCESSMODE=accessdaemon`, create your script and print the "run this script with sudo" instructions (`BUILDFREQ=true` and `BUILDDAEMON=true`)
> 
> 
> This way, the users can select what they want depending on their permissions.

Would it hurt anything to just always set `BUILDFREQ` and `BUILDDAEMON`? Less complex logic that way manipulating the `Makefile`.

TomTheBear:
> > Why? You replaced 3 lines of simple file filtering with around 50 of Python code with code generation of an additional script that will contain only 4 lines with chmod and chown in any case!

> Both singularity* and gaussian-* do this. Maybe the example I lifted from singularity is a bit more verbose than needed, and it seems that maybe some of these things could be de-duplicated in the future.

This is not an answer to my why. Maybe "it's the workaround for https://github.com/spack/spack/issues/4432" is a valid one. My main issue is that it introduces a logic that might need maintenance in the future. If there will be an additional daemon or one of the existing daemons is renamed, the `package.py` needs changes as well. At the moment, the package relies on the installation mechanism of LIKWID.



> > Have there been cases where the changing of ownership failed/caused problems? I have not seen any issues or have been notified of any.

> We like using likwid on a few special nodes on our cluster. Due to "security", we cannot get access as root. Everything builds fine, but the installation step fails at the install -o root -g root <daemon_executable> shrug
> On my personal machine, I'm not going to do a sudo spack install likwid to get around this, so IMO whenever these daemons are installed, it's better to just fix the permissions with the right user afterwards.

This is new to me. With the old build settings (`ACCESSMODE=perf_event`, `BUILDFREQ=false`, `BUILDDAEMON=false`) no root privileges are required. The installation step with owner change should be skipped by `BUILDDAEMON=false`, same for `BUILDFREQ=false`.



> > If you want to add logic to the package, add a variant to select the access mode:
> >* If `ACCESSMODE=perf_event`, no script generation needed, installation as user, maybe tell users that their scope of access is limited by the global `perf_event_paranoid` file (`BUILDFREQ=false` and `BUILDDAEMON=false`)
> >* If `ACCESSMODE=direct`, print a warning that you need root permissions when using LIKWID and no script generation needed (`BUILDFREQ=false` and `BUILDDAEMON=false`)
> >* if `ACCESSMODE=accessdaemon`, create your script and print the "run this script with sudo" instructions (`BUILDFREQ=true` and `BUILDDAEMON=true`)
> >
> >    This way, the users can select what they want depending on their permissions.

> Would it hurt anything to just always set `BUILDFREQ` and `BUILDDAEMON`? Less complex logic that way manipulating the Makefile.

The settings should be adjusted by LIKWID's setup based on the `ACCESSMODE` but to be more clear what the installation should look like and avoid problems, setting the `BUILDFREQ` and `BUILDDAEMON` is better. And if we use the old style with `Makefile` manipulations, it's one if-elseif-else block with each 3 lines. Still less than 50 lines with code generation.

matz-e:
> This is not an answer to my why.

Sorry, that was a paragraph/quote lower.

> My main issue is that it introduces a logic that might need maintenance in the future. If there will be an additional daemon or one of the existing daemons is renamed, the `package.py` needs changes as well. At the moment, the package relies on the installation mechanism of LIKWID.

LIKWID's install mechanism require full root access when installing the software, I'm not going to do `sudo spack install` even on my personal machines. I can condense these 50 lines into way less.

> This is new to me. With the old build settings (`ACCESSMODE=perf_event`, `BUILDFREQ=false`, `BUILDDAEMON=false`) no root privileges are required. The installation step with owner change should be skipped by `BUILDDAEMON=false`, same for `BUILDFREQ=false`.

Part of my team would like the frequency option to be there, but as I mentioned, we don't get root access. Setting the permissions/ownership in hindsight is not an issue for our sysadmins, though.

> The settings should be adjusted by LIKWID's setup based on the `ACCESSMODE` but to be more clear what the installation should look like and avoid problems, setting the `BUILDFREQ` and `BUILDDAEMON` is better. And if we use the old style with `Makefile` manipulations, it's one if-elseif-else block with each 3 lines. Still less than 50 lines with code generation.

Fair enough. Although that still does not get around that the installation requires root privileges, so some code generation would still be needed.

TomTheBear:
OK OK, now I start to get the cause and reason for this. You could have framed it better in the beginning like:

> We use LIKWID in our team and require the feature of CPU frequency manipulation on some nodes. Since the current build settings with `ACCESSMODE=perf_event` do not allow this, we need a way to enable the installation with root privileges and `ACCESSMODE=accessdaemon`. Unfortunately, spack has not sorted out issue https://github.com/spack/spack/issues/4432 yet but the singularity package maintainers provided a suitable workaround in the comments. The idea is to postpone root-specific changes to the daemon files: a script is generated in the installation step and the user is asked to executed it with root privileges in a manual post-install step.

For me it looked like a not-really-required and over-engineered feature. Sorry for that.

A way to make less maintenance burden in the future would be:
- Add variant for `ACCESSMODE`
- Manipulate `ACCESSMODE`, `BUILDDAEMON` and `BUILDFREQ` in the `Makefile`
- Manipulate the `CHOWN` setting in the `Makefile` to use installing user (installation as user even with `ACCESSMODE=accessdaemon`)
- If `ACCESSMODE=accessdaemon`: Generate the script for all files in `$PREFIX/sbin` to change owner to root and set suid bit.
- If `ACCESSMODE=direct`: Generate the script for all files in `$PREFIX/bin/likwid-` to change owner to root

This way, when a daemon name changes or LIKWID adds one, it is picked up automatically. The `sbin` folder should contain only files installed by LIKWID, so not filtering step is required. And we still use the installation mechanism of LIKWID itself. You still have some package-specific info coded into it (all daemons in `sbin` require suid bit, all LIKWID tools start with `likwid-`) but is more flexible.

I see the need for the change (as a workaround or even a suitable solution to https://github.com/spack/spack/issues/4432) but I would be more happy if it is included into spack directly, so that any package can add post-install commands. This way, the output which scripts to execute on behalf of any package could be shown in the end as a summary. Having the script path somewhere is in the build logs might be overseen by the user.

matz-e:
Yes, sorry! I'm not the primary user of the software (and as such don't have much insight into the different modi operandi, I'm accustomed to different profiling tools) and was in a bit of a hurry to at least file something upstream while deploying our software.

I've just pushed the some updates for the variants, but I'll take your comments here into account still.

> If ACCESSMODE=direct: Generate the script for all files in $PREFIX/bin/likwid- to change owner to root

Just change ownership to root, no suid bit?

> I see the need for the change (as a workaround or even a suitable solution to https://github.com/spack/spack/issues/4432) but I would be more happy if it is included into spack directly, so that any package can add post-install commands. This way, the output which scripts to execute on behalf of any package could be shown in the end as a summary. Having the script path somewhere is in the build logs might be overseen by the user.

I totally agree with this. I wasn't aware of the `singularity` output for quite a while.

TomTheBear:
> 
> > If ACCESSMODE=direct: Generate the script for all files in $PREFIX/bin/likwid- to change owner to root
> 
> Just change ownership to root, no suid bit?

Well, yes and no. To work as a user, the suid bit is required, true. BUT these CLI tools are Lua scripts, so the Lua interpreter (`$PREFIX/bin/likwid-lua`) gets suid-root and therefore any user knowing the path to the interpreter can execute any command as root by putting it in a Lua script. I'm not sure whether it works when the scripts have the suid-root bit while the interpreter is owned by the user. I never tested such a scenario. The `direct` mode was always thought for single-user systems with root/sudo privileges for the user (e.g. my desktop system at home). I wouldn't have a problem if we do not provide this access mode variant in spack.

For the `accessdaemon` and it's frequency counterpart, I did an analysis with software security researchers. I incorporated all recommendations and fixed all issues. This is of course no guarantee but much less attack surface than the suid-root bit for the Lua interpreter.

bernhardkaindl (dismissed):
@matz-e @TomTheBear Has the conversation cleared this PR for merging?

Please ping me using @bernhardkaindl (or another reviewer/maintainer) to notify me/other people when it is read for merge, thanks! (maybe you can even merge yourself with this approve from me)

matz-e:
I hope I addressed all comments. I still have a local commit that removes the `direct` access mode, but if one uses Spack on the desktop (we do at our institute), this mode may be useful to retain.

TomTheBear:
If you have the installation with `ACCESSMODE=direct` working, you can add it. I just thought it is the access mode that is used only rarely.

I asked in the SC22 Spack BoF whether spack could add a feature to delay messages to the user until the end of an installation process so that the (in or case) generated script and description is printed after all installations are done and not somewhere in between.

matz-e:
I've removed the `direct` access mode, as I don't have the capacity to test it right now.

TomTheBear:
@matz-e : Looks good, Thanks.

@bernhardkaindl  : We are good to go from my side. We cannot merge it ourselves.

bernhardkaindl (approved):
Approved by @TomTheBear

PR #33505, opened 2022-10-25
Add support for Python 3.11

alalazo:
Modifications:
- [x] Test Spack on Python 3.11 in CI
- [x] Update getting started guide
- [x] Fix failures with Python 3.11

### DIFF
--- a/lib/spack/spack/main.py
+++ b/lib/spack/spack/main.py
@@ -343,17 +343,21 @@ class SpackArgumentParser(argparse.ArgumentParser):
                 self._remove_action(self._actions[-1])
             self.subparsers = self.add_subparsers(metavar="COMMAND", dest="command")
 
-        # each command module implements a parser() function, to which we
-        # pass its subparser for setup.
-        module = spack.cmd.get_module(cmd_name)
+        if cmd_name not in self.subparsers._name_parser_map:
+            # each command module implements a parser() function, to which we
+            # pass its subparser for setup.
+            module = spack.cmd.get_module(cmd_name)
 
-        # build a list of aliases
-        alias_list = [k for k, v in aliases.items() if v == cmd_name]
+            # build a list of aliases
+            alias_list = [k for k, v in aliases.items() if v == cmd_name]
 
-        subparser = self.subparsers.add_parser(
-            cmd_name, aliases=alias_list, help=module.description, description=module.description
-        )
-        module.setup_parser(subparser)
+            subparser = self.subparsers.add_parser(
+                cmd_name,
+                aliases=alias_list,
+                help=module.description,
+                description=module.description,
+            )
+            module.setup_parser(subparser)
 
         # return the callable function for the command
         return spack.cmd.get_command(cmd_name)
--- a/lib/spack/spack/test/util/unparse/unparse.py
+++ b/lib/spack/spack/test/util/unparse/unparse.py
@@ -178,6 +178,71 @@ async def f():
 """
 
 
+match_literal = """\
+match status:
+    case 400:
+        return "Bad request"
+    case 404 | 418:
+        return "Not found"
+    case _:
+        return "Something's wrong with the internet"
+"""
+
+match_with_noop = """\
+match status:
+    case 400:
+        return "Bad request"
+"""
+
+match_literal_and_variable = """\
+match point:
+    case (0, 0):
+        print("Origin")
+    case (0, y):
+        print(f"Y={y}")
+    case (x, 0):
+        print(f"X={x}")
+    case (x, y):
+        print(f"X={x}, Y={y}")
+    case _:
+        raise ValueError("Not a point")
+"""
+
+
+match_classes = """\
+class Point:
+    x: int
+    y: int
+
+def location(point):
+    match point:
+        case Point(x=0, y=0):
+            print("Origin is the point's location.")
+        case Point(x=0, y=y):
+            print(f"Y={y} and the point is on the y-axis.")
+        case Point(x=x, y=0):
+            print(f"X={x} and the point is on the x-axis.")
+        case Point():
+            print("The point is located somewhere else on the plane.")
+        case _:
+            print("Not a point")
+"""
+
+match_nested = """\
+match points:
+    case []:
+        print("No points in the list.")
+    case [Point(0, 0)]:
+        print("The origin is the only point in the list.")
+    case [Point(x, y)]:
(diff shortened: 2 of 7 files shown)


ax3l:
x-ref: #33507

adamjstewart (dismissed):

alalazo:
So here's a change  in behavior that we need to account for in Spack:
```console
$ python
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
>>> import argparse
>>> p = argparse.ArgumentParser()
>>> sp = p.add_subparsers()
>>> sp.add_parser("help")
ArgumentParser(prog=' help', usage=None, description=None, formatter_class=<class 'argparse.HelpFormatter'>, conflict_handler='error', add_help=True)
>>> sp.add_parser("help")
ArgumentParser(prog=' help', usage=None, description=None, formatter_class=<class 'argparse.HelpFormatter'>, conflict_handler='error', add_help=True)
```
and
```console
 python
Python 3.11.0 (main, Oct 31 2022, 20:42:01) [GCC 9.4.0] on linux
>>> import argparse
>>> p = argparse.ArgumentParser()
>>> sp = p.add_subparsers()
>>> sp.add_parser("help")
ArgumentParser(prog=' help', usage=None, description=None, formatter_class=<class 'argparse.HelpFormatter'>, conflict_handler='error', add_help=True)
>>> sp.add_parser("help")
  File "/home/culpo/opt/pythons/store/linux-ubuntu20.04-icelake/gcc-9.4.0/python-3.11.0-25wr4i3255kfgipfl2b22jcvckbrksxv/lib/python3.11/argparse.py", line 1185, in add_parser
    raise ArgumentError(self, _('conflicting subparser: %s') % name)
argparse.ArgumentError: argument {help}: conflicting subparser: help

```

alalazo:
So, looking at `argparse.py` it seems these lines have been added in `add_parser` from Python 3.11:
```python
if name in self._name_parser_map:
    raise ArgumentError(self, _('conflicting subparser: %s') % name)
```

alalazo:
@tgamblin The test currently failing is https://github.com/spack/spack/actions/runs/3388015149/jobs/5629609448 Looking at Python docs, it seems this was already an issue in Python 3.10, but our tests didn't catch it since we are testing core libraries and none of them was using the new language feature.

alalazo:
I think I have "fixes" for the unparser (I am porting what is done in Python 3.11) and adding tests. I'll do a run without the fix though to check why I can't reproduce the issue locally.

lib/spack/spack/main.py
```diff
@@ -343,17 +343,21 @@ def add_command(self, cmd_name):
...
-        module = spack.cmd.get_module(cmd_name)
-
-        # build a list of aliases
-        alias_list = [k for k, v in aliases.items() if v == cmd_name]
-
-        subparser = self.subparsers.add_parser(
-            cmd_name, aliases=alias_list, help=module.description, description=module.description
-        )
-        module.setup_parser(subparser)
+        if cmd_name not in self.subparsers._name_parser_map:
```
haampie, line 346: This is dangerous, cause we might not be forward compatible with Python 3.12+ whenever they feel like changing internals.

lib/spack/spack/test/util/unparse/unparse.py
```diff
@@ -512,3 +577,12 @@ def test_async_with():
 @pytest.mark.skipif(sys.version_info < (3, 5), reason="Not supported < 3.5")
 def test_async_with_as():
     check_ast_roundtrip(async_with_as)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="Not supported < 3.10")
```
haampie, line 582: 3.10?

alalazo: Yes, https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching We started failing in 3.11 because we have a test (without assertion, see #33722 ) that was testing the core libs - and core libraries started using pattern matching in 3.11

lib/spack/spack/main.py
alalazo: True, but I didn't see ani API call to get the same information. We should probably revisit the code here, but not a week before release?

lib/spack/spack/test/util/unparse/unparse.py
haampie: yeah, so skipif < (3, 11)?

lib/spack/spack/main.py
haampie: How about the second best option of doing the bookkeeping with our own `set` somewhere? Then at least it's more likely that people can use an old commit of Spack with a new Python in the future. Best solution would of course be to prevent the same command from being attempted to be registered twice, but well...

tgamblin: FWIW the internals of argparse (at least the field names) seldom change. We are already using internal fields above, so there’s more danger than just this. I don’t particularly *like* it, but this is all to support lazy command loading — which makes spack a lot snappier on a shared FS.

Where’s this getting called twice for the same command in spack? I could look at fixing that.

tgamblin (approved):
LGTM. Is the match unparsing just backported from ast.unparse? My only concern with it is that it’s deterministic across versions of python. I don’t see any problems with it at first glance.  OTOH we don’t have to worry about this until we only support Python >= 3.10, as packages pretty much have to be written in the lowest supported python (our mainline ones at least).

lib/spack/spack/test/util/unparse/unparse.py
tgamblin: This would be the quick fix but IMO it’s good to make sure unparsing is supported across Python versions for package hash. It’ll bite us one day if not.

alalazo:
> LGTM. Is the match unparsing just backported from ast.unparse?

Yes

lib/spack/spack/test/util/unparse/unparse.py
alalazo: @haampie This was an issue for unparsing also for 3.10 We just didn't realize it. It will be an issue for Spack when we support 3.10+

lib/spack/spack/main.py
alalazo: > Where’s this getting called twice for the same command in spack? I could look at fixing that.

As far as I can tell it's only `spack help` code path that results in adding the same definition of `help` twice.

PR #33506, opened 2022-10-25
Add directives for removing patches, conflicts and dependencies

iarspider:
As proposed here: https://github.com/spack/spack/discussions/33057, this is useful for changing recipes via inheritance. I have decided not to touch original directives, since their code is quite complicated


iarspider:
@spackbot fix style

alalazo:
Why not just extracting the base class somewhere, like in `fftw`?

iarspider:
In some cases (*) we don't want to push changes upstream, since we can't guarantee that every function of the affected package will work - "our tests pass + no one complains = it just works" approach. 

(*) Mostly for Python packages we remove version restrictions set by package authors, so some functionality not used by CMS _could be_ affected; also for many packages we remove `gettext` dependency since we don't need other languages and we need to save on image size

iarspider:
@spackbot fix style

iarspider:
> In some cases (*) we don't want to push changes upstream, since we can't guarantee that every function of the affected package will work - "our tests pass + no one complains = it just works" approach.
> 
> (*) Mostly for Python packages we remove version restrictions set by package authors, so some functionality not used by CMS _could be_ affected; also for many packages we remove `gettext` dependency since we don't need other languages and we need to save on image size

So, TLDR, a different use case: crude hacking of recipe vs special implementation of a common package.

alalazo:
Superseded by #35045

PR #33507, opened 2022-10-25
python: 3.11.0

haampie:
Haven't checked the patches etc in detail, but it builds on ubuntu 22.04.


adamjstewart (commented):
I think we're ready to make Python 3.10 the default. [Homebrew](https://formulae.brew.sh/formula/python@3.10) already defaults to Python 3.10, although [Anaconda](https://www.anaconda.com/products/distribution) hasn't yet updated.

Also, you'll likely need to modify existing patches for 3.11, unless there's some evidence that the patch has been incorporated upstream, or unless we can find a way to fix these issues without patches. You should also check all of the compiler- and OS-specific patches to make sure that those still apply.

ax3l:
Hey, all other package managers I use (conda-forge, brew, Ubuntu 22.04 etc.) ship 3.11 now - let's keep this going? :)

For the future, we could include Python RC candidates early, which is how other package managers migrate.

> Python 3.10 default

This is OT for this PR, please handle separately (e.g., #31623)

ax3l (dismissed):
good for me, I think one can add eventual patches for other OS & compilers later instead of stalling, especially since `3.9.15` is marked as default in Spack.

ax3l:
@spackbot re-run pipeline

(random libfabric issue unrelated to this PR)

ax3l:
X-ref: #33505

adamjstewart:
I'm fine with changing the Python version default in a separate PR.

I'm less fine with assuming that the current patches are no longer needed in the latest version or breaking installation with those random compilers. I would still like to see that in this PR before merging.

ax3l:
Ok 👍

haampie:
Maybe someone else can take over this PR. I don't want to get too involved in the mess that is Python. (Forgot to edit: I think I went through all the patches, just have my questions about this odd distutils situation)

All I see is: we ship 12 patches (?!) for an issue that's closed after 15 years upstream saying it's not going to be fixed: https://github.com/python/cpython/issues/42093... Then why ship patches?

> Please report the issue with setuptools. distutils is no longer under development. We recommend that all users use setuptools.

Why don't we set an upperbound on the Python version if that distutils i still an issue somewhere?

Regarding tkinter, it's been simplified in setup.py to:

```python
    def detect_tkinter(self):
        self.addext(Extension('_tkinter', ['_tkinter.c', 'tkappinit.c']))
```

Is dropping `self.detect_tkinter` from setup.py enough?

cray / fj patches still apply.

haampie:
@adamjstewart the gitlab situation has improved 😌 time to merge?

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -56,7 +61,6 @@ class Python(Package):
     version(
```
adamjstewart: Can you wrap this back onto a single line?

adamjstewart (commented):
For distutils stuff, most packages now use setuptools, but I believe numpy still uses distutils. Distutils is deprecated in 3.10 and removed in 3.12, so we don't need to patch it forever. I can't remember why we needed to patch it originally, but I imagine we have a lot of packages that use Python and C++.

For tkinter stuff, as long as the install procedure no longer searchers for tkinter when `~tkinter` is requested, that should be sufficient.

Do the patches for other compilers/OSes work for 3.11? You'll have to comment out the compiler/OS checks in the `when` clauses to make sure.

var/spack/repos/builtin/packages/python/package.py
haampie: Does that change the hash? If yes, then rather not :p

adamjstewart: It should not change the hash.

haampie: done

haampie:
Patches are OK as far as I can see

haampie:
@scottwittenburg this "broken on develop" stuff really does not work, it's extremely off-putting :( no, i'm not going to "fix" paraview in a PR that deals with a new Python version. Sigh.

Productive suggestion: can individual jobs be marked "allowed to fail"? In this case it would be helpful to get this pr in even though paraview is "broken"

haampie:
@spackbot run pipeline

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -415,19 +416,26 @@ class Python(Package):
     # C/C++ modules, consider installing a Spack-managed Python with
     # this patch instead. For more information, see:
     # https://github.com/spack/spack/pull/16856
+    #
+    # NOTE: distutils was deprecated in 3.10, so we don't add new patches
```
becker33: This probably requires an update to py-numpy and py-scipy, if not others.

becker33 (changes requested):

var/spack/repos/builtin/packages/python/package.py
adamjstewart: Yes, all versions of numpy still use distutils. We can't just ignore these patches.

haampie: py-numpy builds fine for me, py-scipy already has an upperbound of the form `depends_on(python@:3.10, when=@1.8:)` -- although it lacks it for `<=1.6.1` which cause spack to use that, but that's a mistake in the package and unrelated to this pr.

haampie: okay you win i added the patch back :|

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -415,19 +416,26 @@ class Python(Package):
...
+    # for 3.11. If packages still require distutils patches, restrict them
+    # to python@:3.10.
```
adamjstewart: You can remove this comment now. Comments aren't included in the hash.

adamjstewart (dismissed):
Tested to make sure that patches for all compilers/OSes still apply to 3.11.0.

var/spack/repos/builtin/packages/python/package.py
haampie:
```suggestion
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -420,14 +421,17 @@ class Python(Package):
     patch("python-2.7.17+-distutils-C++-fixup.patch", when="@2.7.17:2.7.18")
     patch("python-3.6.8-distutils-C++.patch", when="@3.6.8,3.7.2")
     patch("python-3.7.3-distutils-C++.patch", when="@3.7.3")
-    patch("python-3.7.4+-distutils-C++.patch", when="@3.7.4:")
+    patch("python-3.7.4+-distutils-C++.patch", when="@3.7.4:3.10")
     patch("python-3.7.4+-distutils-C++-testsuite.patch", when="@3.7.4:")
+    patch("python-3.11-distutils-C++.patch", when="@3.11.0:3.11")
     patch("cpython-windows-externals.patch", when="@:3.9.6 platform=windows")
     patch("tkinter.patch", when="@:2.8,3.3:3.7 platform=darwin")
+
```
haampie:
```suggestion
```

adamjstewart (approved):

becker33 (approved):

PR #33508, opened 2022-10-25
Gmsh: Fix CGNS support for version up to 4.7.1

RemiLacroix-IDRIS:
Gmsh started supporting the "scoping" option of CGNS in version 4.8.0.


tldahlgren (approved):
Addition and dependency for new variant LGTM.

PR #33509, opened 2022-10-25
Enable cooldown for failed index.json fetches

blue42u:
My previous PR (#32137) added a cooldown between fetches of a buildcache's `index.json` when requested, but only when the fetch was successful. In some cases (https://github.com/spack/spack/pull/32137#issuecomment-1289769151) this causes long repeated fetch failures.

Enable the cooldown for failed fetches as well, and propagate the success/failure status when a fetch is prevented by the cooldown.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -107,7 +107,8 @@ class BinaryCacheIndex(object):
         # cache (_mirrors_for_spec)
         self._specs_already_associated = set()
 
-        # mapping from mirror urls to the time.time() of the last index fetch.
+        # mapping from mirror urls to the time.time() of the last index fetch and a bool indicating
+        # whether the fetch succeeded or not.
         self._last_fetch_times = {}
 
         # _mirrors_for_spec is a dictionary mapping DAG hashes to lists of
@@ -346,21 +347,25 @@ class BinaryCacheIndex(object):
                     with_cooldown
                     and ttl > 0
                     and cached_mirror_url in self._last_fetch_times
-                    and now - self._last_fetch_times[cached_mirror_url] < ttl
+                    and now - self._last_fetch_times[cached_mirror_url][0] < ttl
                 ):
-                    # The fetch worked last time, so don't error
-                    all_methods_failed = False
+                    # We're in the cooldown period, don't try to fetch again
+                    # If the fetch succeeded last time, consider this update a success, otherwise
+                    # re-report the error here
+                    if self._last_fetch_times[cached_mirror_url][1]:
+                        all_methods_failed = False
                 else:
                     # May need to fetch the index and update the local caches
                     try:
                         needs_regen = self._fetch_and_cache_index(
                             cached_mirror_url, expect_hash=cached_index_hash
                         )
-                        self._last_fetch_times[cached_mirror_url] = now
+                        self._last_fetch_times[cached_mirror_url] = (now, True)
                         all_methods_failed = False
                     except FetchCacheError as fetch_error:
                         needs_regen = False
                         fetch_errors.extend(fetch_error.errors)
+                        self._last_fetch_times[cached_mirror_url] = (now, False)
                     # The need to regenerate implies a need to clear as well.
                     spec_cache_clear_needed |= needs_regen
                     spec_cache_regenerate_needed |= needs_regen
@@ -392,11 +397,12 @@ class BinaryCacheIndex(object):
                 # Need to fetch the index and update the local caches
                 try:
                     needs_regen = self._fetch_and_cache_index(mirror_url)
-                    self._last_fetch_times[mirror_url] = now
+                    self._last_fetch_times[mirror_url] = (now, True)
                     all_methods_failed = False
                 except FetchCacheError as fetch_error:
                     fetch_errors.extend(fetch_error.errors)
                     needs_regen = False
+                    self._last_fetch_times[mirror_url] = (now, False)
                 # Generally speaking, a new mirror wouldn't imply the need to
                 # clear the spec cache, so leave it as is.
                 if needs_regen:


lib/spack/spack/binary_distribution.py
```diff
@@ -346,21 +347,23 @@ def update(self, with_cooldown=False):
...
                     and ttl > 0
                     and cached_mirror_url in self._last_fetch_times
-                    and now - self._last_fetch_times[cached_mirror_url] < ttl
+                    and now - self._last_fetch_times[cached_mirror_url][0] < ttl
                 ):
-                    # The fetch worked last time, so don't error
-                    all_methods_failed = False
+                    if self._last_fetch_times[cached_mirror_url][1]:
+                        # The fetch worked last time, so don't error
+                        all_methods_failed = False
```
scheibelp, line 356: Can you add a comment here like:

```
if self._last_fetch_times[cached_mirror_url][1]:
    # The fetch worked last time, so don't error
    all_methods_failed = False
# else: we've already tried once within the cooldown period;
# re-report the fact that we failed
```

blue42u: How's this:
```python
# We're in the cooldown period, don't try to fetch again
# If the fetch succeeded last time, consider this update a success, otherwise
# re-report the error here
if self._last_fetch_times[cached_mirror_url][1]:
  all_methods_failed = False
```

lib/spack/spack/binary_distribution.py
```diff
@@ -346,21 +347,25 @@ def update(self, with_cooldown=False):
...
                 else:
                     # May need to fetch the index and update the local caches
                     try:
                         needs_regen = self._fetch_and_cache_index(
                             cached_mirror_url, expect_hash=cached_index_hash
                         )
-                        self._last_fetch_times[cached_mirror_url] = now
+                        self._last_fetch_times[cached_mirror_url] = (now, True)
                         all_methods_failed = False
```
scheibelp, line 364: @haampie I have a question that's only tangentially-related to this PR: it appears that if any `cached_mirror_url` succeeds, then `all_methods_failed` will be false for all the URLs, so this variable is specifically just tracking whether we could update at least one mirror index successfully. Is that what you wanted explicitly (from examining the `git blame`, it looks like you added that logic - let me know if I got that wrong)?

lib/spack/spack/binary_distribution.py
scheibelp: This works fine, thanks!

scheibelp (approved):

scheibelp:
@spackbot run pipeline

scottwittenburg:
While debugging some pipeline generation jobs from the cloud locally, I just pulled this commit in, but it didn't prevent the hundreds of those messages (example below).  I'll run with the debugger and see if I can figure out why not.

```
==> [2022-10-25-18:20:32.321047] Checking existence of https://mirror.spack.io/build_cache/index.json
==> [2022-10-25-18:20:32.466065] Failure reading URL: Download failed: HTTP Error 404: Not Found
```

scottwittenburg:
Oops, too late.

scheibelp:
Does merging this make it harder to debug? I can push a PR that reverts this. I figured it would be easier to figure out if it were in (normally I wouldn't handle it like that, but I wasn't sure how easy it was to debug the CI behavior with a PR).

scottwittenburg:
In that case, no worries.  You're right it will be easier not to need to pull this in separately.

lib/spack/spack/binary_distribution.py
haampie: I don't know what it has turned into, but the idea is to have an exception to propagate out of this if nothing worked. Before the pr from my side it was a sad debug message iirc, or no error at all?

scottwittenburg:
I did some debugging to get to the bottom of the repeated, failed attempts to fetch the index during pipeline generation.  The problem happens if the configuration contains any mirrors without an index.  The `_fetch_and_cache_index` function detects that condition and simply returns `False`, but given the new behavior, I'm guessing it should maybe also add an entry to `self._last_fetch_times`?

blue42u:
@scottwittenburg If `_fetch_and_cache_index` returns `False` it should already add an entry to `self._last_fetch_times`: https://github.com/spack/spack/blob/8be63786883111a5f4037ef73d2f7ffb22e8edf2/lib/spack/spack/binary_distribution.py#L360-L364

Can you check whether it's actually calling `_fetch_and_cache_index` after the first call (i.e. is [this conditional](https://github.com/spack/spack/blob/8be63786883111a5f4037ef73d2f7ffb22e8edf2/lib/spack/spack/binary_distribution.py#L346-L351) always false)? I'm wondering if the fetch failed in the first call, and after that it's constantly re-raising the `FetchCacheError` on every call to `update()` from `find_by_hash()`.

scottwittenburg:
This is the call [site](https://github.com/spack/spack/blob/8be63786883111a5f4037ef73d2f7ffb22e8edf2/lib/spack/spack/binary_distribution.py#L399) I'm hitting in local testing.  When that just returns `False` because there is no index, I guess it does put an entry into the `_last_fetch_times` (entering `True` for the success/failure indicator, is that ok?).  And that's the line that triggers the index fetch every time.  If `False` is returned [here](url), then the mirror url never gets put into `self._local_cache_index`, and hence it never gets into the loop with that conditional you linked above.

blue42u:
I see now. So in short, [this entire code block](https://github.com/spack/spack/blob/8be63786883111a5f4037ef73d2f7ffb22e8edf2/lib/spack/spack/binary_distribution.py#L397-L409) needs to be put inside a conditional [just like this](https://github.com/spack/spack/blob/8be63786883111a5f4037ef73d2f7ffb22e8edf2/lib/spack/spack/binary_distribution.py#L345-L357). That would enable the cooldown for "new" mirror urls, which then covers the case where a mirror has no index and is always considered "new."

> entering `True` for the success/failure indicator, is that ok?

That is how the logic worked before my PRs, a `_fetch_and_cache_index` call that returns is considered "successful" and prevents throwing a `FetchCacheError`. IMHO the logic doesn't make much sense, but I'm not the author here.

blue42u:
Implementation of the above idea over in https://github.com/spack/spack/pull/33781.

PR #33513, opened 2022-10-25
binary_distribution: compress level 9 -> 6

haampie:
Use the same compression level as `gzip` (6) instead of what Python uses
(9).

The LLVM tarball takes 4m instead of 12m to create, and goes from
2351091267 to 2365440599 bytes, so, less than 1%, not worth to
spend the majority of the time on...

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -1161,9 +1161,10 @@ def _build_tarball(
             tty.die(e)
 
     # create gzip compressed tarball of the install prefix
-    # compresslevel=6 corresponds to the default level of gzip, which is
-    # not reasonably fast and produces tarballs not much bigger than Python's
-    # default of level 9.
+    # On AMD Ryzen 3700X and an SSD disk, we have the following on compression speed:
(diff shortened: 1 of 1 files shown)


alalazo (changes requested):
I think this is worth a comment just above the context manager, reporting e.g. the time with LLVM. Apart from that I checked and we don't have other `tarfile.open` that need matching compression, so LGTM.

lib/spack/spack/binary_distribution.py
```diff
@@ -1161,7 +1161,10 @@ def _build_tarball(
             tty.die(e)
 
     # create gzip compressed tarball of the install prefix
-    with closing(tarfile.open(tarfile_path, "w:gz")) as tar:
+    # compresslevel=6 corresponds to the default level of gzip, which is
+    # not reasonably fast and produces tarballs not much bigger than Python's
+    # default of level 9.
```
alalazo:
```suggestion
    # much faster than Python's default of level 9, while producing binaries that are not
    # much bigger
```

alalazo (approved):

PR #33518, opened 2022-10-26
opencascade: new version 7.6.3

wdconinc:
New version of opencascade (well, a few new versions, but this is likely the last of the 7.6.x series before the 7.7.x series starts). Build tested on ubuntu 22.04 with gcc-12.

### DIFF
--- a/var/spack/repos/builtin/packages/opencascade/package.py
+++ b/var/spack/repos/builtin/packages/opencascade/package.py
@@ -21,7 +21,7 @@ class Opencascade(CMakePackage):
     version(
         "7.6.3",
         extension="tar.gz",
-        sha256="e7f989d52348c3b3acb7eb4ee001bb5c2eed5250cdcceaa6ae97edc294f2cabd",
+        sha256="baae5b3a7a38825396fc45ef9d170db406339f5eeec62e21b21036afeda31200",
     )
     version(
         "7.6.0",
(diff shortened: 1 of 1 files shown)


tldahlgren (commented):
Strange.  I'm not getting the commit for 7.6.3 when I run `spack checksum opencascade 7.6.3` using this branch.  Where is that coming from?

var/spack/repos/builtin/packages/opencascade/package.py
```diff
@@ -18,6 +18,11 @@ class Opencascade(CMakePackage):
 
     maintainers = ["wdconinc"]
 
+    version(
+        "7.6.3",
+        extension="tar.gz",
+        sha256="e7f989d52348c3b3acb7eb4ee001bb5c2eed5250cdcceaa6ae97edc294f2cabd",
+    )
```
tldahlgren: I get the following from `checksum`:

```suggestion
    version(
        "7.6.3",
        extension="tar.gz",
        sha256="baae5b3a7a38825396fc45ef9d170db406339f5eeec62e21b21036afeda31200",
    )
```

wdconinc: You're right. That was the checksum for 7.6.0...

wdconinc:
OT I've been trying to figure out how to make list_url do something useful with https://git.dev.opencascade.org/gitweb/?p=occt.git;a=tags to automate version finding... but that's out of scope for this PR.

wdconinc:
Verified that my successful test build of 7.6.3 was using the right version and right hash. I copied the wrong hash into the PR.

tldahlgren (approved):
Confirmed sha256.

tldahlgren:
Looks like pod/containers unready may have lead to issues during generation phase for one of the pipelines.

@spackbot run pipeline

PR #33523, opened 2022-10-26
gitlab ci: show build machine info

haampie:
We sometimes see build systems inspect the host cpu details (e.g. openblas), which sometimes fails. So, it would be helpful to have the host cpu info available in the logs.


alalazo:
I have encountered the OpenBLAS issue too, where the job was failing to detect the CPU. Restarting the job and getting a luckier rescheduling made the day, but this surely helps having more info on what's going on.

alalazo (approved):

PR #33526, opened 2022-10-26
Octopus: Refactor the ocotpus package to use  AutotoolsPackage

iamashwin99:
Octopus uses the Autotools build system, this PR refactors the spack package to use the `AutotoolsPackage` class instead of the generic `Package`.


fangohr (dismissed):
This looks good to me: spack supports autotools-based packages, Octopus is such a package and with this pull request we use spack's support for autotools. 

Disclaimer: @iamashwin99 works at my institute and we have discussed this proposed change before.

var/spack/repos/builtin/packages/octopus/package.py
```diff
@@ -90,7 +90,12 @@ class Octopus(Package, CudaPackage):
     # TODO: etsf-io, sparskit,
     # feast, libfm, pfft, isf, pnfft, poke
 
-    def install(self, spec, prefix):
+    def autoreconf(self, spec, prefix):
+
+        autoreconf("--install", "--verbose", "--force")
```
michaelkuhn: Is this really necessary? It seems the releases have a `configure` file, why regenerate it?

fangohr: Well spotted, and thank you for looking into this. 

The official octopus releases (distributed per tar ball) do indeed include the `configure` file. However, it is missing in the octopus git repository, and thus the `autoreconf` step needs to be executed at least for the `octopus@develop` spack target.

alalazo: @fangohr That's the default behavior of autoreconf:

 https://github.com/spack/spack/blob/9a51d42cecfdedd9e5776824a420efdf9877d95e/lib/spack/spack/build_systems/autotools.py#L539-L544

alalazo (changes requested):
Can we avoid running autoreconf every time?

var/spack/repos/builtin/packages/octopus/package.py
michaelkuhn: As @alalazo has already mentioned, it should be possible to simply remove the `autoreconf` method. Moreover, adding `when="@develop"` to the autoconf, automake, libtool and m4 dependencies avoids them for release versions.

fangohr: That's perfect - thank you for the suggestion. We'll get onto it (@iamashwin99) and report back.

alalazo (approved):

alalazo:
Failed two times on "No specs to rebuild".

PR #33529, opened 2022-10-26
slepc: fix for slepc+cuda ^petsc+kokkos+cuda ^kokkos+cuda+wrapper

balay:
kokkos wrappers modify mpicxx - breaking slepc build.

This slepc fix emulates the fix in petsc

Ideally kokkos should not force kokkos+wrapper on dependent pkgs using kokkos+cuda

cc: @masterleinad

### DIFF
--- a/var/spack/repos/builtin/packages/slepc/package.py
+++ b/var/spack/repos/builtin/packages/slepc/package.py
@@ -113,12 +113,20 @@ class Slepc(Package, CudaPackage, ROCmPackage):
         when="@3.13.0:+blopex",
     )
 
+    def revert_kokkos_nvcc_wrapper(self):
+        # revert changes by kokkos-nvcc-wrapper
+        if self.spec.satisfies("^kokkos+cuda+wrapper"):
+            env["MPICH_CXX"] = env["CXX"]
+            env["OMPI_CXX"] = env["CXX"]
+            env["MPICXX_CXX"] = env["CXX"]
+
     def install(self, spec, prefix):
         # set SLEPC_DIR for installation
         # Note that one should set the current (temporary) directory instead
         # its symlink in spack/stage/ !
         os.environ["SLEPC_DIR"] = os.getcwd()
 
+        self.revert_kokkos_nvcc_wrapper()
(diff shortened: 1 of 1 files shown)


joseeroman (approved):

eugeneswalker (approved):

masterleinad:
> Ideally kokkos should not force kokkos+wrapper on dependent pkgs using kokkos+cuda

In the end, every compilation unit that contains `Kokkos` headers needs to be compiled as device code (in case `Cuda` is enabled as a backend) and the easiest way to achieve this is to use `nvcc_wrapper` as the compiler (if `nvcc` should be used as device compiler). Packages using `CMake` and `target_link_libraries` would get this behavior automatically (with an arbitrary provided host compiler) using `CMake` magic.

balay:
If a kokkos dependent pkg needs `nvcc_wrapper` then it should be enabled in that pkg spec - and not forced by kokkos on all pkgs.

petsc, slepc don't need this - as they handle the kokkos issues in its build system. Enabling these wrappers break petsc, slepc builds. [hence these fixes  to revert the wrapper changes to MPI compilers]

balay:
https://gitlab.spack.io/spack/spack/-/jobs/3986646
```
$ spack --version
0.19.0.dev0 (19671b9721ac2894f89476b0fd82863841d7ad19)
$ cd share/spack/gitlab/cloud_pipelines/stacks/${SPACK_CI_STACK_NAME}
$ spack env activate --without-view .
$ spack ci generate --check-index-only --buildcache-destination "${SPACK_BUILDCACHE_DESTINATION}" --artifacts-root "${CI_PROJECT_DIR}/jobs_scratch_dir" --output-file "${CI_PROJECT_DIR}/jobs_scratch_dir/cloud-ci-pipeline.yml"
==> Starting concretization pool with 16 processes
Cleaning up project directory and file based variables 00:00
ERROR: Job failed: pod "runner-6ezzdmu1-project-2-concurrent-13vnkfd" status is "Failed"
```
I'm not sure why this failed.

@spackbot run pipeline

PR #33531, opened 2022-10-26
Add vasp variant to control shmem compile options

rfbgo:
Currently the vasp package always enables the use of shmem to reduce algorithm memory usage (see
https://www.vasp.at/wiki/index.php/Precompiler_options). This is great, but on some systems gives compile errors with the interoperability of C and Fortran. This PR makes that shmem flag optional, but retains the existing default on behavior


tldahlgren (approved):
LGTM

tldahlgren:
@alansill @amd-toolchain-support Did you want to review this PR before it is merged?

var/spack/repos/builtin/packages/vasp/package.py
```diff
@@ -46,6 +46,8 @@ class Vasp(MakefilePackage):
         "https://github.com/henniggroup/VASPsol",
     )
 
+    variant("shmem", default=True, description="Enable use_shmem build flag")
```
tldahlgren, line 49: Do you know if this option applies to all (at least listed) versions or from a certain version on?

rfbgo: I'm not sure, I've only worked with v6+ (and don't have access to v5) but adding this variant does not add any "risk" (and actually lessens it slightly, since if there is an incompatible version this will very useful).

Given the current package applies that flag to all versions, then some listed versions would not work as-is today if it is not compatible. Given the way the current package always applies that flag, I am comfortable assuming this new variant compatible with all listed versions, and the ability to turn it off can only be an improvement

prckent: This will work with v5.

rfbgo:
@tldahlgren anything else needed to get this merged?

PR #33540, opened 2022-10-26
Make --backtrace show non-SpackError backtraces

haampie:
Currently `--backtrace` only dumps backtraces for `SpackError`s, but that doesn't help with debugging the  LockTimeoutError (which is just an Exception) seen in CI.

### DIFF
--- a/lib/spack/spack/main.py
+++ b/lib/spack/spack/main.py
@@ -108,7 +108,7 @@ spack_working_dir = None
 spack_ld_library_path = os.environ.get("LD_LIBRARY_PATH", "")
 
 #: Whether to print backtraces on error
-show_backtrace = False
+SHOW_BACKTRACE = False
 
 
 def set_working_dir():
@@ -572,8 +572,8 @@ def setup_main_options(args):
 
     if args.debug or args.backtrace:
         spack.error.debug = True
-        global show_backtrace
-        show_backtrace = True
+        global SHOW_BACKTRACE
+        SHOW_BACKTRACE = True
 
     if args.debug:
         spack.util.debug.register_interrupt_handler()
@@ -1007,7 +1007,7 @@ def main(argv=None):
         e.die()  # gracefully die on any SpackErrors
 
     except KeyboardInterrupt:
-        if spack.config.get("config:debug") or show_backtrace:
+        if spack.config.get("config:debug") or SHOW_BACKTRACE:
             raise
         sys.stderr.write("\n")
         tty.error("Keyboard interrupt.")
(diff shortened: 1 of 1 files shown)


haampie:
@spackbot fix style

tldahlgren:
Would it not be sufficient to change line 578 from `if args.debug:` to `if args.debug or args.backtrace:`?  Or pull the setting of `config:debug` out of that if-statement to set if either argument is True.

haampie:
I checked `register_interrupt_handler`, it's dropping you in a Python shell on Ctrl+C which shouldn't happen; I don't really know what the environment tracing is to be honest, but it seems like I don't need that. I was hoping to just get backtraces on all errors with `--backtrace` s.t. Gitlab CI is not too noisy with debug messages, yet does give context when errors happen

tldahlgren:
> I checked `register_interrupt_handler`, it's dropping you in a Python shell on Ctrl+C which shouldn't happen; I don't really know what the environment tracing is to be honest, but it seems like I don't need that. I was hoping to just get backtraces on all errors with `--backtrace` s.t. Gitlab CI is not too noisy with debug messages, yet does give context when errors happen

Yeah, I was afraid of that when I looked up the line, which is why I modified my comment to include pulling out setting `config.debug` based on the args. 

Also, I'm not seeing any obvious place where `spack.error.debug` (line 574) is being used but I didn't do a lot of digging.

tldahlgren:
> > I checked `register_interrupt_handler`, it's dropping you in a Python shell on Ctrl+C which shouldn't happen; I don't really know what the environment tracing is to be honest, but it seems like I don't need that. I was hoping to just get backtraces on all errors with `--backtrace` s.t. Gitlab CI is not too noisy with debug messages, yet does give context when errors happen
> 
> Yeah, I was afraid of that when I looked up the line, which is why I modified my comment to include pulling out setting `config.debug` based on the args.
> 
> Also, I'm not seeing any obvious place where `spack.error.debug` (line 574) is being used but I didn't do a lot of digging.

My only point was that maybe you could skip the global in favor of leveraging `config:debug` as is already being done with `args.debug`.

haampie:
`spack.error.debug` is used to print stacktraces of `SpackError`

> My only point was that maybe you could skip the global in favor of leveraging config:debug as is already being done with args.debug.

It would require a dedicated config value, I think that's not really necessary unless someone asks for that specifically

haampie:
@spackbot help

haampie:
@spackbot rebuild everything

alalazo (approved):
I'll merge this for now, we can improve it later

PR #33541, opened 2022-10-26
new packages (py-auditwheel, py-medaka, py-parasail, py-progressbar33, py-pyspoa) and updates to others (py-scikit-build, py-ont-fast5-api)

luke-dt:


tldahlgren (commented):
Confirmed all 7 new sha256 (at one per package).

var/spack/repos/builtin/packages/py-auditwheel/package.py
```diff
@@ -0,0 +1,23 @@
...
+    homepage = "https://github.com/pypa/auditwheel"
+    pypi = "auditwheel/auditwheel-5.1.2.tar.gz"
+
+    version("5.1.2", sha256="3ee5830014931ea84af5cd065c637b6614efa03d9b88bd8fbfc924e7ed01d6ba")
+
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-setuptools@45:", type="build")
+    depends_on("py-setuptools-scm@6.2:", type="build")
+    depends_on("py-pyelftools", type=("build", "run"))
+    depends_on("py-importlib-metadata", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-pyelftools@0.24:", type=("build", "run"))
    depends_on("py-importlib-metadata", when="^python@:3.7", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-medaka/package.py
```diff
@@ -0,0 +1,41 @@
...
+    depends_on("py-grpcio", type=("build", "run"))
+    depends_on("py-h5py", type=("build", "run"))
+    depends_on("py-intervaltree", type=("build", "run"))
+    depends_on("py-tensorflow@2.7.0:2.7", type=("build", "run"))
+    depends_on("py-numpy", type=("build", "run"))
+    depends_on("minimap2", type=("build", "run"))
+    depends_on("py-ont-fast5-api", type=("build", "run"))
+    depends_on("py-parasail", type=("build", "run"))
+    depends_on("py-pysam@0.16.0.1:", type=("build", "run"))
+    depends_on("py-pyspoa@0.0.3:", type=("build", "run"))
```
adamjstewart: parasail and pyspoa aren't needed on aarch64: https://github.com/nanoporetech/medaka/blob/v1.7.2/setup.py#L66,L67. I think we want something like:
```python
depends_on("py-parasail", when="target=x86_64:", type=("build", "run"))
depends_on("py-parasail", when="target=ppc64le:", type=("build", "run"))
```
Not sure what other targets Spack supports, maybe @becker33 or @alalazo would know.

var/spack/repos/builtin/packages/py-parasail/package.py
```diff
@@ -0,0 +1,25 @@
...
+    """Python Bindings for the Parasail C Library. Parasail is a SIMD C (C99)
+    library containing implementations of the Smith-Waterman (local),
+    Needleman-Wunsch (global), and semi-global pairwise sequence alignment
+    algorithms."""
+
+    homepage = "https://github.com/jeffdaily/parasail-python"
+    pypi = "parasail/parasail-1.3.3.tar.gz"
+
+    version("1.3.3", sha256="06f05066d9cf624c0b043f51a1e9d2964154e1edd0f9843e0838f32073e576f8")
+
```
adamjstewart, line 20: Prob also want to add perl and m4: https://github.com/jeffdaily/parasail-python/blob/v1.3.3/setup.py#L164

var/spack/repos/builtin/packages/py-pyspoa/package.py
```diff
@@ -0,0 +1,23 @@
...
+    pypi = "pyspoa/pyspoa-0.0.8.tar.gz"
+
+    version("0.0.8", sha256="8299d18066b498a6ef294c5a33a99266ded06eeb022f67488d2caecba974b0a4")
+
+    depends_on("py-wheel@0.34.1:", type="build")
+    depends_on("py-setuptools", type="build")
+    depends_on("cmake@3.18.4", type="build")
+    depends_on("py-pybind11@2.4:", type=("build", "run"))
+    depends_on("py-auditwheel@3.1.0:", type=("build", "run"))
+    depends_on("py-scikit-build@0.11.1", type=("build", "run"))
```
adamjstewart: auditwheel and scikit-build aren't used by this project

var/spack/repos/builtin/packages/py-pyspoa/package.py
```diff
@@ -0,0 +1,23 @@
...
+
+class PyPyspoa(PythonPackage):
+    """Python bindings to spoa"""
+
+    homepage = "https://github.com/nanoporetech/pyspoa"
+    pypi = "pyspoa/pyspoa-0.0.8.tar.gz"
+
+    version("0.0.8", sha256="8299d18066b498a6ef294c5a33a99266ded06eeb022f67488d2caecba974b0a4")
+
+    depends_on("py-wheel@0.34.1:", type="build")
```
adamjstewart: Where did you find this wheel version restriction?

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pyspoa/package.py
luke-dt: https://github.com/nanoporetech/pyspoa/blob/v0.0.8/requirements.txt

var/spack/repos/builtin/packages/py-pyspoa/package.py
luke-dt: yeah, I was looking at the "Developer Quick Start" where it tells you to run `pip install -r requirements.txt`. Digging through the source code, it looks like auditwheel is [only used in this script](https://github.com/nanoporetech/pyspoa/blob/v0.0.8/.github/workflows/build-wheels.sh), and scikit-build isn't used at all. I'll try re-building with those removed. Odd that the devs included it.

adamjstewart: `requirements.txt` is used by devs to install the libraries needed to use the software but also the development tools needed to maintain it. You should never check `requirements.txt` unless `setup.py` explicitly uses it in some way.

var/spack/repos/builtin/packages/py-pyspoa/package.py
adamjstewart: This file isn't included in the source distribution, I wouldn't trust it.

luke-dt:
confirmed that py-medaka builds and runs with real data with the last commit

var/spack/repos/builtin/packages/py-medaka/package.py
adamjstewart: Same for pyspoa

adamjstewart (approved):

Issue #33547, opened 2022-10-27
Spack install race conditions


haampie:
There are still multiple races when firing up multiple `spack install` processes:

## sbang

```
==> Error: Failed to install zstd due to FileNotFoundError: [Errno 2] No such file or directory: '/home/software/spack/bin/.sbang.tmp'
```

## Provider cache hang

```
==> Error: Timed out waiting for a write lock after 120.202s.
    Made 281 attempts on file: /root/.spack/cache/providers/.builtin-index.json.lock
```

blue42u:
We're seeing these two races in HPCToolkit's CI on one of our CI runners (very underpowered machine):

## Binary index cache hang
```
==> Error: Failed to install bzip2 due to LockTimeoutError: Timed out waiting for a write lock after 120.200s.
    Made 281 attempts on file: /root/.spack/cache/indices/.783b91ff88_c3d0ebca2d.json.lock
```

## Binary index cache race
```
==> Error: Failed to install zstd due to FileNotFoundError: [Errno 2] No such file or directory: '/root/.spack/cache/indices/1f0f0ff7c4_8567a6c618.json'
```

haampie:
@blue42u are you still seeing this?

blue42u:
@haampie Our CI has avoided running Spack in parallel for the past year or so, so I can't confirm if the bug is still present. Enough has changed on our end that I don't think I can recreate the original scenario.

tgamblin:
Given @blue42u's response let's close this for now and reopen it if he sees issues again.

melven:
The following error still happens for me:

```
==> Error: llvm-14.0.6-ufgvxvrhl4zghtgdoj7up23mos4auffi: LockTimeoutError: Timed out waiting for a write lock after 60.073s.
    Made 161 attempts on file: /tmp/f_jkessx/spack-stage/.lock
```

Version: `develop-2025-02-16`
Setup: running 3 processes on the same node with 128 cores (with `nohup spack -e SomeEnv install -j 128`) with a local build cache and spack directory on NFSv3.
(I know it's using more threads than cores this way, but this helps with download delays, etc).

PR #33549, opened 2022-10-27
fix racy sbang

haampie:
refers #33547

Spack currently creates a temporary sbang that is moved "atomically" in place,
but this temporary causes races when multiple processes start installing sbang.

Let's just stick to an idempotent approach. Notice that we only re-install sbang
if Spack updates it (since we do file compare), and sbang was only touched
18 times in the past 6 years, whereas we hit the sbang tempfile issue
frequently with parallel install on a fresh spack instance in CI.

Also fixes a bug where permissions weren't updated if config changed but
the latest version of the sbang file was already installed.

### DIFF
--- a/lib/spack/spack/hooks/sbang.py
+++ b/lib/spack/spack/hooks/sbang.py
@@ -186,44 +186,38 @@ def install_sbang():
     ``sbang`` here ensures that users can access the script and that
     ``sbang`` itself is in a short path.
     """
-    # copy in a new version of sbang if it differs from what's in spack
     sbang_path = sbang_install_path()
-    if os.path.exists(sbang_path) and filecmp.cmp(spack.paths.sbang_script, sbang_path):
-        return
 
-    # make $install_tree/bin
+    all = spack.spec.Spec("all")
+    group_name = spack.package_prefs.get_package_group(all)
+    config_mode = spack.package_prefs.get_package_dir_permissions(all)
+    group_id = grp.getgrnam(group_name).gr_gid if group_name else None
+
+    # First setup the bin dir correctly.
     sbang_bin_dir = os.path.dirname(sbang_path)
-    fs.mkdirp(sbang_bin_dir)
+    if not os.path.isdir(sbang_bin_dir):
+        fs.mkdirp(sbang_bin_dir)
 
-    # get permissions for bin dir from configuration files
-    group_name = spack.package_prefs.get_package_group(spack.spec.Spec("all"))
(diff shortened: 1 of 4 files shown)


haampie:
@spackbot fix style

lib/spack/spack/hooks/sbang.py
```diff
@@ -186,42 +186,38 @@ def install_sbang():
...
+        already_installed = filecmp.cmp(spack.paths.sbang_script, sbang_path)
+    except (IOError, OSError):
+        already_installed = False
 
-    # set permissions on `sbang` (including group if set in configuration)
-    os.chmod(sbang_tmp_path, config_mode)
-    if group_name:
-        os.chown(sbang_tmp_path, os.stat(sbang_tmp_path).st_uid, grp.getgrnam(group_name).gr_gid)
+    if not already_installed:
+        shutil.copy(spack.paths.sbang_script, sbang_path)
```
tgamblin: This is still a race.  If two processes look at sbang and decide it's not there, one of these copies may end up appending or conflicting with the other.  We need *both* a unique tmpfile and the move into place.

I would tweak `llnl.util.filesystem.write_tmp_and_move` to be thread-safe (i.e. make it come up with a unique tmp name, which it doesn't currently do) and then use it.

Just use [`tempfile.NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) with `prefix`, `suffix`, and `dir`:

```
>>> f = tempfile.NamedTemporaryFile(prefix=".foo.bar", suffix=".tmp", dir=".")
>>> f.name
'/Users/gamblin2/.foo.barhnlfjiqn.tmp'
```

tgamblin (changes requested):

lib/spack/spack/hooks/sbang.py
tgamblin: FWIW I could see this race coming up in CI for teams that clone spack and do a parallel install right away.

haampie: Yeah, I think this is the best we can do. Confusingly, Linux has O_TMPFILE + linkat to accomplish pretty much what we need (invisible, auto-removed file, that can appear atomically after writing data into it and setting permissions), except that it's not able to rename, so, can't replace a file :(.

haampie:
@spack style fix

haampie:
@tgamblin can you have another look?

tgamblin (approved):

PR #33558, opened 2022-10-27
starpu: correct fxt dependency variant when simgrid

llnns:
- Correct starpu+fxt+simgrid build: force fxt+static variant


sthibaul:
Mmm, I believe static is only needed for simgrid+mpi+fxt, not for simgrid+fxt?

llnns:
Indeed

sthibaul (approved):

alalazo (approved):

PR #33563, opened 2022-10-27
Support ROCm backing in DiHydrogen

bvanessen:
Added support for building the DiHydrogen package and LBANN extensions
to DiHydrogen with ROCm libraries.

Fixed a bug on Cray systems where CMake didn't try hard enough to find an MPI-compatible compiler wrapper.  Make it look more.

Added support for the roctracer package when using ROCm libraries.

### DIFF
--- a/var/spack/repos/builtin/packages/dihydrogen/package.py
+++ b/var/spack/repos/builtin/packages/dihydrogen/package.py
@@ -157,9 +157,13 @@ class Dihydrogen(CMakePackage, CudaPackage, ROCmPackage):
             "-DH2_ENABLE_DISTCONV_LEGACY=%s" % ("+distconv" in spec),
             "-DH2_ENABLE_OPENMP=%s" % ("+openmp" in spec),
             "-DH2_ENABLE_FP16=%s" % ("+half" in spec),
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/dihydrogen/package.py
```diff
@@ -155,7 +157,7 @@ def cmake_args(self):
             "-DH2_ENABLE_DISTCONV_LEGACY=%s" % ("+distconv" in spec),
             "-DH2_ENABLE_OPENMP=%s" % ("+openmp" in spec),
             "-DH2_ENABLE_FP16=%s" % ("+half" in spec),
-            "-DH2_ENABLE_HIP_ROCM=%s" % ("+rocm" in spec),
```
alalazo, line 158: Was this plain wrong, or the option changed? If the option changed this new recipe will break the build of older versions.

bvanessen: @alalazo Good catch.  Let me fix that.

bvanessen: Fixed.

alalazo (approved):

PR #33564, opened 2022-10-27
Overhaul certain spack packages

johnwparent:
This PR ports adds Windows support to a number of packages and ports a number of packages to use the recent builder system merged in by #30738 

This PR also adds two new builders `NMakefiles` and `MSBuild`

You may also note that all of these packages are dependencies of Paraview/VTK. This is not a coincidence as this PR is part of a larger effort to support those two packages on Windows.

This PR ports:
Python - to multibuild system
Perl - to multibuild system
adios2 -  to Windows and multibuild system
c-blosc -  to Windows and multibuild system
double-conversion -  to  multibuild system
eigen -  to  multibuild system
expat - to Windows and multibuild system
freetype - to multibuild system
jsoncpp - to multibuild system
libjpeg-turbo - to multibuild system
libogg -  to Windows and multibuild system
libpng -  to multibuild system
libtheora -  to Windows and multibuild system
libxml2 - to Windows and multibuild system
lz4 -  to Windows and multibuild system
netcdf-cxx - to multibuild system
netcdf-c -  to Windows and multibuild system
pegtl -  to Windows and multibuild system
proj -  to Windows and multibuild system
protobuf -  to Windows and multibuild system
snappy -  to Windows and multibuild system
winlibiconv - created a new package to provide libiconv on Windows
xz -  to Windows and multibuild system
zfp -  to Windows and multibuild system
zstd -  to Windows and multibuild system

This pr depends on #31141 and #33021

### DIFF
--- a/var/spack/repos/builtin/packages/adios2/package.py
+++ b/var/spack/repos/builtin/packages/adios2/package.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import os
+import sys
 
 from spack.package import *
 
@@ -54,7 +55,11 @@ class Adios2(CMakePackage, CudaPackage):
     # change how we're supporting differnt library types in the package at anytime if
     # spack decides on a standardized way of doing it across packages
     variant("shared", default=True, when="+pic", description="Build shared libraries")
-    variant("pic", default=True, description="Build pic-enabled static libraries")
+    variant(
+        "pic",
+        default=not sys.platform == "win32",
+        description="Build pic-enabled static libraries",
+    )
 
     # Features
     variant("mpi", default=True, description="Enable MPI")
@@ -95,12 +100,18 @@ class Adios2(CMakePackage, CudaPackage):
     conflicts("%oneapi@:2022.1.0", when="+fortran")
 
     depends_on("cmake@3.12.0:", type="build")
-    depends_on("pkgconfig", type="build")
 
-    depends_on("libffi", when="+sst")  # optional in DILL
-    depends_on("libfabric@1.6.0:", when="+sst")  # optional in EVPath and SST
-    # depends_on('bison', when='+sst')     # optional in FFS, broken package
-    # depends_on('flex', when='+sst')      # optional in FFS, depends on BISON
+    for plat in ["linux", "darwin", "cray"]:
+        depends_on("pkgconfig", type="build", when="platform = %s" % plat)
+
+        # libffi and libfabric and not currently supported on Windows
+        # see Paraview's superbuild handling of libfabric at https://gitlab.kitware.com/paraview/paraview-superbuild/-/blob/master/projects/adios2.cmake#L3
+        depends_on("libffi", when="+sst platform=%s" % plat)  # optional in DILL
+        depends_on(
+            "libfabric@1.6.0:", when="+sst platform=%s" % plat
+        )  # optional in EVPath and SST
+        # depends_on('bison', when='+sst')     # optional in FFS, broken package
+        # depends_on('flex', when='+sst')      # optional in FFS, depends on BISON
 
     depends_on("mpi", when="+mpi")
     depends_on("libzmq", when="+dataman")
@@ -164,6 +175,45 @@ class Adios2(CMakePackage, CudaPackage):
         elif self.spec.satisfies("%fj +fortran"):
             env.set("FFLAGS", "-Ccpp")
 
+    @property
+    def libs(self):
+        spec = self.spec
+        libs_to_seek = set()
+
+        if "@2.6:" in spec:
+            libs_to_seek.add("libadios2_core")
+            libs_to_seek.add("libadios2_c")
+            libs_to_seek.add("libadios2_cxx11")
+            if "+fortran" in spec:
+                libs_to_seek.add("libadios2_fortran")
+
+            if "+mpi" in spec:
+                libs_to_seek.add("libadios2_core_mpi")
+                libs_to_seek.add("libadios2_c_mpi")
+                libs_to_seek.add("libadios2_cxx11_mpi")
+                if "+fortran" in spec:
+                    libs_to_seek.add("libadios2_fortran_mpi")
+
+            if "@2.7: +shared+hdf5" in spec and "@1.12:" in spec["hdf5"]:
+                libs_to_seek.add("libadios2_h5vol")
+
+        else:
+            libs_to_seek.add("libadios2")
+            if "+fortran" in spec:
+                libs_to_seek.add("libadios2_fortran")
+
+        return find_libraries(
+            list(libs_to_seek), root=self.spec.prefix, shared=("+shared" in spec), recursive=True
+        )
+
+    def setup_run_environment(self, env):
+        try:
+            all_libs = self.libs
+            idx = all_libs.basenames.index("libadios2_h5vol.so")
+            env.prepend_path("HDF5_PLUGIN_PATH", os.path.dirname(all_libs[idx]))
+        except ValueError:
+            pass
+
     def cmake_args(self):
         spec = self.spec
         from_variant = self.define_from_variant
@@ -214,42 +264,3 @@ class Adios2(CMakePackage, CudaPackage):
             args.append("-DPython_EXECUTABLE:FILEPATH=%s" % spec["python"].command.path)
 
         return args
-
-    @property
-    def libs(self):
-        spec = self.spec
-        libs_to_seek = set()
-
-        if "@2.6:" in spec:
-            libs_to_seek.add("libadios2_core")
-            libs_to_seek.add("libadios2_c")
-            libs_to_seek.add("libadios2_cxx11")
-            if "+fortran" in spec:
-                libs_to_seek.add("libadios2_fortran")
-
-            if "+mpi" in spec:
-                libs_to_seek.add("libadios2_core_mpi")
-                libs_to_seek.add("libadios2_c_mpi")
-                libs_to_seek.add("libadios2_cxx11_mpi")
-                if "+fortran" in spec:
-                    libs_to_seek.add("libadios2_fortran_mpi")
-
-            if "@2.7: +shared+hdf5" in spec and "@1.12:" in spec["hdf5"]:
-                libs_to_seek.add("libadios2_h5vol")
-
-        else:
-            libs_to_seek.add("libadios2")
-            if "+fortran" in spec:
-                libs_to_seek.add("libadios2_fortran")
-
-        return find_libraries(
-            list(libs_to_seek), root=self.spec.prefix, shared=("+shared" in spec), recursive=True
-        )
-
-    def setup_run_environment(self, env):
-        try:
-            all_libs = self.libs
-            idx = all_libs.basenames.index("libadios2_h5vol.so")
-            env.prepend_path("HDF5_PLUGIN_PATH", os.path.dirname(all_libs[idx]))
-        except ValueError:
-            pass
--- a/var/spack/repos/builtin/packages/expat/package.py
+++ b/var/spack/repos/builtin/packages/expat/package.py
@@ -8,7 +8,7 @@ import sys
 from spack.package import *
 
 
-class Expat(AutotoolsPackage):
+class Expat(CMakePackage):
     """Expat is an XML parser library written in C."""
 
     homepage = "https://libexpat.github.io/"
@@ -102,7 +102,7 @@ class Expat(AutotoolsPackage):
     # `~libbsd`.
     variant(
         "libbsd",
-        default=sys.platform != "darwin",
+        default=sys.platform != "darwin" and sys.platform != "win32",
         description="Use libbsd (for high quality randomness)",
     )
 
@@ -112,12 +112,19 @@ class Expat(AutotoolsPackage):
         url = "https://github.com/libexpat/libexpat/releases/download/R_{0}/expat-{1}.tar.bz2"
         return url.format(version.underscored, version.dotted)
 
-    def configure_args(self):
-        spec = self.spec
+    # def configure_args(self):
+    #     spec = self.spec
+    #     args = [
+    #         "--without-docbook",
+    #         "--enable-static",
+    #     ]
+    #     if "+libbsd" in spec and "@2.2.1:" in spec:
+    #         args.append("--with-libbsd")
+    #     return args
+
+    def cmake_args(self):
         args = [
-            "--without-docbook",
-            "--enable-static",
+            self.define("EXPAT_BUILD_DOCS", False),
+            self.define_from_variant("EXPAT__WITH_LIBBSD", "libbsd"),
         ]
-        if "+libbsd" in spec and "@2.2.1:" in spec:
-            args.append("--with-libbsd")
         return args
--- a/var/spack/repos/builtin/packages/freetype/package.py
+++ b/var/spack/repos/builtin/packages/freetype/package.py
@@ -3,10 +3,12 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
 
 
-class Freetype(AutotoolsPackage):
+class Freetype(CMakePackage):
     """FreeType is a freely available software library to render fonts.
     It is written in C, designed to be small, efficient, highly customizable,
     and portable while capable of producing high-quality output (glyph images)
@@ -30,7 +32,6 @@ class Freetype(AutotoolsPackage):
 
     depends_on("bzip2")
     depends_on("libpng")
-    depends_on("pkgconfig", type="build")
 
     conflicts(
         "%intel",
@@ -47,6 +48,13 @@ class Freetype(AutotoolsPackage):
         headers.directories = [self.prefix.include.freetype2]
         return headers
 
+
+class CMakeBuilder(CMakeBuilder):
+    def cmake_args(self):
+        return [self.define("FT_WITH_ZLIB", True), self.define("FT_WITH_BZIP2", True)]
+
+
+class AutoToolsBuilder(AutotoolsBuilder):
     def configure_args(self):
         args = [
             "--with-brotli=no",
--- a/var/spack/repos/builtin/packages/libiconv/package.py
+++ b/var/spack/repos/builtin/packages/libiconv/package.py
@@ -28,7 +28,8 @@ class Libiconv(AutotoolsPackage, GNUMirrorPackage):
     # We cannot set up a warning for gets(), since gets() is not part
     # of C11 any more and thus might not exist.
     patch("gets.patch", when="@1.14")
-    provides("iconv")
+
+    conflicts("platform=windows")
 
     conflicts("@1.14", when="%gcc@5:")
 
--- a/var/spack/repos/builtin/packages/libogg/package.py
+++ b/var/spack/repos/builtin/packages/libogg/package.py
@@ -44,19 +44,19 @@ class GenericBuilder(GenericBuilder):
 
     def build(self, spec, prefix):
         if spec.satisfies("%msvc"):
-            plat_tools = self.compiler.msvc_version
+            plat_tools = self.pkg.compiler.msvc_version
         else:
             raise RuntimeError("Package does not support non MSVC compilers on Windows")
         ms_build_args = ["libogg_static.vcxproj", "/p:PlatformToolset=v%s" % plat_tools]
         msbuild(*ms_build_args)
 
     def install(self, spec, prefix):
-        mkdirp(self.prefix.include.ogg)
-        mkdirp(self.prefix.lib)
-        mkdirp(self.prefix.share)
+        mkdirp(prefix.include.ogg)
+        mkdirp(prefix.lib)
+        mkdirp(prefix.share)
         install(
-            os.path.join(self.stage.source_path, "include", "ogg", "*.h"), self.prefix.include.ogg
+            os.path.join(self.pkg.stage.source_path, "include", "ogg", "*.h"), prefix.include.ogg
         )
         plat_prefix = "x64" if self.is_64bit() else "x86"
-        install(os.path.join(plat_prefix, "Debug", "*.lib"), self.prefix.lib)
-        install_tree(os.path.join(self.stage.source_path, "doc"), self.prefix.share)
+        install(os.path.join(plat_prefix, "Debug", "*.lib"), prefix.lib)
+        install_tree(os.path.join(self.pkg.stage.source_path, "doc"), prefix.share)
--- a/var/spack/repos/builtin/packages/libogg/package.py
+++ b/var/spack/repos/builtin/packages/libogg/package.py
@@ -3,10 +3,14 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import os
+import platform
+
+from spack.build_systems.generic import GenericBuilder
 from spack.package import *
 
 
-class Libogg(AutotoolsPackage):
+class Libogg(CMakePackage, AutotoolsPackage, Package):
     """Ogg is a multimedia container format, and the native file and stream
     format for the Xiph.org multimedia codecs."""
 
@@ -24,3 +28,35 @@ class Libogg(AutotoolsPackage):
         sha256="0f4d289aecb3d5f7329d51f1a72ab10c04c336b25481a40d6d841120721be485",
         when="@1.3.4 platform=darwin",
     )
+    build_system(
+        conditional("cmake", when="@1.3.4:"),
+        conditional("generic", when="@1.3.2 platform=windows"),
+        "autotools",
+        default="autotools",
+    )
+
+
+class GenericBuilder(GenericBuilder):
+    phases = ["build", "install"]
+
+    def is_64bit(self):
+        return platform.machine().endswith("64")
+
+    def build(self, spec, prefix):
+        if spec.satisfies("%msvc"):
+            plat_tools = self.compiler.msvc_version
+        else:
+            raise RuntimeError("Package does not support non MSVC compilers on Windows")
+        ms_build_args = ["libogg_static.vcxproj", "/p:PlatformToolset=v%s" % plat_tools]
+        msbuild(*ms_build_args)
+
+    def install(self, spec, prefix):
+        mkdirp(self.prefix.include.ogg)
+        mkdirp(self.prefix.lib)
+        mkdirp(self.prefix.share)
+        install(
+            os.path.join(self.stage.source_path, "include", "ogg", "*.h"), self.prefix.include.ogg
+        )
+        plat_prefix = "x64" if self.is_64bit() else "x86"
+        install(os.path.join(plat_prefix, "Debug", "*.lib"), self.prefix.lib)
+        install_tree(os.path.join(self.stage.source_path, "doc"), self.prefix.share)
--- a/var/spack/repos/builtin/packages/libpng/package.py
+++ b/var/spack/repos/builtin/packages/libpng/package.py
@@ -4,9 +4,9 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
+from spack.build_systems.cmake import CMakeBuilder
 
-
-class Libpng(AutotoolsPackage):
+class Libpng(CMakePackage):
     """libpng is the official PNG reference library."""
 
     homepage = "http://www.libpng.org/pub/png/libpng.html"
@@ -48,7 +48,9 @@ class Libpng(AutotoolsPackage):
         args += self.enable_or_disable("libs")
         return args
 
-    def check(self):
-        # Libpng has both 'check' and 'test' targets that are aliases.
-        # Only need to run the tests once.
-        make("check")
+class CMakeBuilder(CMakeBuilder):
+    def cmake_args(self):
+        return [
+            self.define("CMAKE_CXX_FLAGS", self.spec["zlib"].headers.include_flags),
+            self.define("ZLIB_ROOT", self.spec["zlib"].prefix),
+        ]
--- a/var/spack/repos/builtin/packages/libtheora/package.py
+++ b/var/spack/repos/builtin/packages/libtheora/package.py
@@ -98,7 +98,8 @@ class MSBuildBuilder(MSBuildBuilder):
         super().build(pkg, spec, prefix)
 
     def install(self, pkg, spec, prefix):
+        plat_arch = "x64" if self.is_64bit() else "x86"
         with working_dir(self.build_directory):
-            install_tree(os.path.join("Win32", "Release"), prefix.lib)
+            install_tree(os.path.join(plat_arch, "Release"), prefix.lib)
         with working_dir(pkg.stage.source_path):
             install_tree("include", prefix.include)
--- a/var/spack/repos/builtin/packages/libtheora/package.py
+++ b/var/spack/repos/builtin/packages/libtheora/package.py
@@ -3,14 +3,16 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import platform
 import sys
+import os
 
 from spack.build_systems.autotools import AutotoolsBuilder
-from spack.build_systems.scons import SConsBuilder
+from spack.build_systems.msbuild import MSBuildBuilder
 from spack.package import *
 
 
-class Libtheora(AutotoolsPackage, SConsPackage):
+class Libtheora(AutotoolsPackage, MSBuildPackage):
     """Theora Video Compression."""
 
     homepage = "https://www.theora.org"
@@ -30,7 +32,10 @@ class Libtheora(AutotoolsPackage, SConsPackage):
         depends_on("libtool", type="build")
         depends_on("m4", type="build")
 
-    build_system("scons", "autotools", default="autotools" if sys.platform != "win32" else "scons")
+    with when("platform=windows"):
+        variant("static", default=True, description="Enable static build, if false shared library is built")
+
+    build_system("msbuild", "autotools", default="autotools" if sys.platform != "win32" else "msbuild")
 
     patch("exit-prior-to-running-configure.patch", when="@1.1.1")
     patch("fix_encoding.patch", when="@1.1:")
@@ -39,7 +44,7 @@ class Libtheora(AutotoolsPackage, SConsPackage):
         sha256="8b1f256fa6bfb4ce1355c5be1104e8cfe695c8484d8ea19db06c006880a02298",
         when="^libpng@1.6:",
     )
-
+    patch("libtheora-inc-external-ogg.patch", when="platform=windows")
 
 class AutotoolsBuilder(AutotoolsBuilder):
     def configure_args(self):
@@ -55,5 +60,45 @@ class AutotoolsBuilder(AutotoolsBuilder):
             sh("./autogen.sh", "prefix={0}".format(prefix))
 
 
-class SConsBuilder(SConsBuilder):
-    pass
+class MSBuildBuilder(MSBuildBuilder):
+    def is_64bit(self):
+        return platform.machine().endswith("64")
+
+    def setup_build_environment(self, env):
+        spec = self.pkg.spec
+        env.set("SPACK_OGG_PREFIX", spec["libogg"].prefix)
+        devenv_path = os.path.join(self.pkg.compiler.vs_root, "Common7", "IDE")
+        env.prepend_path("PATH", devenv_path)
+
+    @property
+    def build_directory(self):
+        win_dir = os.path.join(super().build_directory, "win32")
+        vs_dir = "VS2008"
+        return os.path.join(win_dir, vs_dir)
+
+    @property
+    def toolchain_version(self):
+        return "v" + self.pkg.compiler.platform_toolset_ver
+
+    @property
+    def sln_file(self):
+        if self.pkg.spec.satisfies("+static"):
+            f =  "libtheora_static.sln"
+        else:
+            f = "libtheora_dynamic.sln"
+        return f
+
+    def msbuild_args(self):
+        return [self.define("Configuration", "Release"), self.sln_file]
+
+    def build(self, pkg, spec, prefix):
+        with working_dir(self.build_directory):
+            devenv = Executable("devenv")
+            devenv(self.sln_file, "/Upgrade")
+        super().build(pkg, spec, prefix)
+
+    def install(self, pkg, spec, prefix):
+        with working_dir(self.build_directory):
+            install_tree(os.path.join("Win32", "Release"), prefix.lib)
+        with working_dir(pkg.stage.source_path):
+            install_tree("include", prefix.include)
--- a/var/spack/repos/builtin/packages/libtheora/package.py
+++ b/var/spack/repos/builtin/packages/libtheora/package.py
@@ -3,10 +3,14 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import sys
+
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.scons import SConsBuilder
 from spack.package import *
 
 
-class Libtheora(AutotoolsPackage):
+class Libtheora(AutotoolsPackage, SConsPackage):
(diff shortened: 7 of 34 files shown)


var/spack/repos/builtin/packages/libtiff/package.py
```diff
@@ -6,7 +6,7 @@
 from spack.package import *
 
 
-class Libtiff(AutotoolsPackage):
+class Libtiff(CMakePackage):
```
adamjstewart: This should be changed to use Spack's new multi-build system support. See https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems for details.

adamjstewart (changes requested):
It would be helpful if each of these packages were changed in separate PRs since each change doesn't require other changes.

var/spack/repos/builtin/packages/libtiff/package.py
johnwparent: @adamjstewart I was unaware of this until about five minute ago. I've added a note to the top of this PR. I'll be refactoring as much of this PR as possible to use the multiple build systems. Feel free to hold of on review until that point, or review the in-situ version, whatever works.

adamjstewart: Yep, this was merged just yesterday. Should make Windows support significantly easier since packages can subclass AutotoolsPackage only on Linux/macOS and use a different build system on Windows. Hopefully we can refactor most packages like this.

johnwparent: Personally I'm very excited about this PR. As you can see a number of these ports ended up fairly messy. I'll be refactoring as much as possible as I go.

adamjstewart: Same, packages like Python can also be refactored as well.

johnwparent:
@alalazo I've used your PR to port/refactor a number of packages. If you have a second I'd appreciate a review just to be sure I'm using this new functionality as optimally as possible.

adamjstewart:
Can you split this PR into separate PRs for each package? Most of these changes are independent of each other, and it's really hard to properly review PRs that modify dozens of packages.

var/spack/repos/builtin/packages/libtiff/package.py
johnwparent: I've refactored Python and Perl  and included them here.

lib/spack/spack/build_systems/nmakefile.py
```diff
@@ -0,0 +1,126 @@
...
+    """Specialized class for packages built using a Makefiles."""
+
+    #: This attribute is used in UI queries that need to know the build
+    #: system base class
+    build_system_class = "NMakePackage"
+
+    build_system("nmake")
+    conflicts("platform=linux",  when="build_system=nmake")
+    conflicts("platform=darwin", when="build_system=nmake")
+    conflicts("platform=cray",   when="build_system=nmake")
```
alalazo: There's already a `nmake.py` file. Is this different? In case can you extend the other one?

lib/spack/spack/build_environment.py
```diff
@@ -558,6 +558,9 @@ def _set_variables_for_single_module(pkg, module):
 
     if sys.platform == "win32":
         m.nmake = Executable("nmake")
+        m.msbuild = Executable("msbuild")
+        # analog to configure for win32
+        m.cscript = Executable("cscript")
```
alalazo: We'll be trying to remove these global settings here, in favor of using `setup_dependent_package` in the corresponding builder. In this way we won't pick-up executables from the system accidentally (or at least it will be rarer).

alalazo (changes requested):
I have a comment on `nmake` and on adding more global executables. Also, can we start extracting single package changes from here? I have seen packages, for instance, that don't need modifications to core and can be reviewed on their own.

lib/spack/spack/build_environment.py
```diff
@@ -287,7 +287,7 @@ def clean_environment():
 def _add_werror_handling(keep_werror, env):
     keep_flags = set()
     # set of pairs
-    replace_flags = []  # type: List[Tuple[str,str]]
+    replace_flags = []
```
scheibelp: Was this type information incorrect?

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -3,10 +3,12 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
 
 
-class Freetype(AutotoolsPackage):
+class Freetype(CMakePackage):
```
scheibelp, line 11: Either:

* This should inherit both CMakePackage and AutotoolsPackage and define the build_system variant
* or (if CMake is now used for freetype on both Windows and Linux) the Autotools references should be removed

var/spack/repos/builtin/packages/libogg/package.py
```diff
@@ -3,10 +3,16 @@
...
+import platform
+
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.cmake import CMakeBuilder
+from spack.build_systems.generic import GenericBuilder
 from spack.package import *
 
 
-class Libogg(AutotoolsPackage):
+class Libogg(Package):
```
scheibelp: This should inherit CMakePackage, Package, and AutotoolsPackage (whenever it's possible for a package to need a build system, it should be included here).

var/spack/repos/builtin/packages/libogg/package.py
```diff
@@ -24,3 +30,43 @@ class Libogg(AutotoolsPackage):
...
+        default="autotools",
+    )
+
+
+class CMakeBuilder(CMakeBuilder):
+    pass
+
+
+class AutotoolBuilder(AutotoolsBuilder):
+    pass
```
scheibelp: If you aren't overriding any functions, you should not have to instantiate the class here (same for CMakeBuilder)

var/spack/repos/builtin/packages/c-blosc/package.py
```diff
@@ -43,6 +44,8 @@ class CBlosc(CMakePackage):
     def libs(self):
         return find_libraries("libblosc", root=self.prefix, recursive=True)
 
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: It's fine to do this, but was it necessary? IMO it's ideal to avoid using the builder classes when there's only one possible choice.

var/spack/repos/builtin/packages/double-conversion/package.py
```diff
@@ -29,5 +30,7 @@ class DoubleConversion(CMakePackage):
     version("1.1.4", sha256="24b5edce8c88f0f632c83e60e0bde11252656dc3b714ba195619c1798ff28834")
     version("1.1.3", sha256="f0d1b8621592a3cf010c04c3e1c0f08455fc0fc7ee22e1583e2a63dc6d3e3871")
 
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question here as for `c-blosc`)

var/spack/repos/builtin/packages/eigen/package.py
```diff
@@ -68,16 +68,18 @@ class Eigen(CMakePackage):
...
         env.prepend_path("CPATH", self.prefix.include.eigen3)
 
+    @property
+    def headers(self):
+        headers = find_all_headers(self.prefix.include)
+        headers.directories = [self.prefix.include.eigen3]
+        return headers
+
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question here as for `c-blosc`)

var/spack/repos/builtin/packages/expat/package.py
```diff
@@ -105,12 +106,11 @@ def url_for_version(self, version):
         url = "https://github.com/libexpat/libexpat/releases/download/R_{0}/expat-{1}.tar.bz2"
         return url.format(version.underscored, version.dotted)
 
-    def configure_args(self):
-        spec = self.spec
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question here as for `c-blosc`)

var/spack/repos/builtin/packages/jsoncpp/package.py
```diff
@@ -60,6 +61,8 @@ def patch(self):
             "src/lib_json/json_value.cpp",
         )
 
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question here as for `c-blosc`)

var/spack/repos/builtin/packages/libiconv/package.py
```diff
@@ -28,7 +29,8 @@ class Libiconv(AutotoolsPackage, GNUMirrorPackage):
     # We cannot set up a warning for gets(), since gets() is not part
     # of C11 any more and thus might not exist.
     patch("gets.patch", when="@1.14")
-    provides("iconv")
+    if sys.platform != "win32":
+        provides("iconv")
```
scheibelp: I think it would be clearer to add `conflicts(platform=windows)` since it looks like Spack shouldn't try to build this at all on Windows.

var/spack/repos/builtin/packages/libpng/package.py
```diff
@@ -3,10 +3,11 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
 
 
-class Libpng(AutotoolsPackage):
+class Libpng(CMakePackage):
```
scheibelp, line 9: The PR description mentions

> libpng - to multibuild system

but it looks like this just switches libpng to use CMake on both Windows and Linux (which presumably would work fine).

var/spack/repos/builtin/packages/libtheora/package.py
```diff
@@ -33,14 +31,25 @@ class Libtheora(AutotoolsPackage):
         when="^libpng@1.6:",
     )
 
+
+class AutotoolsBuilder(AutotoolsBuilder):
+    depends_on("autoconf", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+    depends_on("m4", type="build")
```
scheibelp: `depends_on` cannot be placed in the builder, it must appear in the package with something like `with when("build_system=autotools"):` (see `arpack-ng` and `libjpeg-turbo` for examples).

var/spack/repos/builtin/packages/libtheora/package.py
```diff
@@ -3,10 +3,12 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.scons import SConsBuilder
 from spack.package import *
 
 
-class Libtheora(AutotoolsPackage):
+class Libtheora(Package):
```
scheibelp: This should still inherit autotools. If this can build with scons in some cases (on Windows I assume) then it should also define a `build_system` variant. I thought `scons` worked on both Linux/Windows, so it would probably make sense to steer the default to be Autotools on linux and SCons on Windows.

var/spack/repos/builtin/packages/libtiff/package.py
```diff
@@ -128,3 +128,15 @@ def configure_args(self):
             args.extend(self.enable_or_disable(var))
 
         return args
+
+    def cmake_args(self):
```
scheibelp: Why does the AutotoolsBuilder have a cmake_args and cmake function defined?

var/spack/repos/builtin/packages/libxml2/package.py
```diff
@@ -17,6 +21,8 @@ class Libxml2(AutotoolsPackage):
     url = "https://download.gnome.org/sources/libxml2/2.9/libxml2-2.9.13.tar.xz"
     list_url = "https://gitlab.gnome.org/GNOME/libxml2/-/releases"
 
+    phases = ["configure", "build", "install"]
```
scheibelp: You shouldn't have to define `phases` here: I think it's clearer to keep that in the builders.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -86,11 +88,11 @@ class NetcdfC(AutotoolsPackage):
     #         description='Enable CDM Remote support')
 
     # The patch for 4.7.0 touches configure.ac. See force_autoreconf below.
-    depends_on("autoconf", type="build", when="@4.7.0,main")
-    depends_on("automake", type="build", when="@4.7.0,main")
-    depends_on("libtool", type="build", when="@4.7.0,main")
+    # depends_on("autoconf", type="build", when="@4.7.0,main")
```
scheibelp: This should be uncommented and likely inside of a `with when(build_system=autotools)` block.

var/spack/repos/builtin/packages/netcdf-cxx/package.py
```diff
@@ -26,8 +27,10 @@ def libs(self):
         shared = True
         return find_libraries("libnetcdf_c++", root=self.prefix, shared=shared, recursive=True)
 
-    def configure_args(self):
-        args = []
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: The PR description mentions

> netcdf-cxx - to multibuild system

but it looks like this just replaces an Autotools build with a CMake-based build. If so, that's fine but a separate CMakeBuilder class is not necessary.

var/spack/repos/builtin/packages/paraview/package.py
```diff
@@ -341,6 +342,8 @@ def setup_run_environment(self, env):
                     env.prepend_path("PYTHONPATH", join_path(pv_pydir, "_paraview.zip"))
                     env.prepend_path("PYTHONPATH", join_path(pv_pydir, "_vtk.zip"))
 
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question here as for `c-blosc`)

var/spack/repos/builtin/packages/pegtl/package.py
```diff
@@ -32,6 +32,14 @@ class Pegtl(CMakePackage):
...
     patch(patch_url, sha256=patch_checksum, level=0, when="@2.1.4:2.8.3")
 
+    @run_after("build")
+    @on_package_attributes(run_tests=True)
+    def check(self):
+        with working_dir(self.build_directory):
+            make("test", parallel=False)
+
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question here as for `c-blosc`)

var/spack/repos/builtin/packages/perl/package.py
```diff
@@ -515,3 +347,173 @@ def test(self):
         msg = "Hello, World!"
         options = ["-e", 'use warnings; use strict;\nprint("%s\n");' % msg]
         self.run_test(exe, options, msg, installed=True, purpose=reason)
+
+
+class RunAfter(object):
+    @run_after("install")
+    def symlink_windows(self):
+        if not is_windows:
+            return
```
scheibelp: I think it would be clearer to put this specific callback in the NmakeBuilder definition (since it only runs there).

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,7 +83,7 @@ class Proj(AutotoolsPackage):
     )
 
     # https://proj.org/install.html#build-requirements
-    depends_on("pkgconfig@0.9.0:", type="build", when="@6:")
+    # depends_on("pkgconfig@0.9.0:", type="build", when="@6:")
```
scheibelp: Remove commented-out dependency.

It looks like this now uses CMake everywhere (Linux and Windows) - that's fine if so just checking.

var/spack/repos/builtin/packages/protobuf/package.py
```diff
@@ -95,6 +96,8 @@ class Protobuf(CMakePackage):
         sha256="fa1abf042eddc1b3b43875dc018c651c90cd1c0c5299975a818a1610bee54ab8",
     )
 
+    build_system("cmake", "autotools", default="cmake")
```
scheibelp: This used to use CMake on Linux. Is there a reason to add this? If this is needed, the package should inherit AutotoolsPackage as well. If not, then the separate CMakeBuilder is not needed (i.e. like for `c-blosc`).

var/spack/repos/builtin/packages/snappy/package.py
```diff
@@ -23,6 +24,8 @@ class Snappy(CMakePackage):
 
     patch("link_gtest.patch", when="@:1.1.8")
 
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question as for `c-blosc`)

var/spack/repos/builtin/packages/vtk/package.py
```diff
@@ -178,6 +179,8 @@ def setup_build_environment(self, env):
                 env.append_flags("CFLAGS", "-DH5_USE_18_API")
                 env.append_flags("CXXFLAGS", "-DH5_USE_18_API")
 
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question as for `c-blosc`)

var/spack/repos/builtin/packages/zfp/package.py
```diff
@@ -121,7 +122,8 @@ class Zfp(CMakePackage, CudaPackage):
         msg="Using zfp with tight error requires a rounding mode other than never",
     )
 
-    # CMake options
+
+class CMakeBuilder(CMakeBuilder):
```
scheibelp: (same question as `c-blosc`)

var/spack/repos/builtin/packages/zstd/package.py
```diff
@@ -55,36 +58,20 @@ class Zstd(MakefilePackage):
...
-        # library targets
-        lib_args = ["-C", "lib"] + args + ["install-pc", "install-includes"]
-        if "libs=shared" in spec:
-            lib_args.append("install-shared")
-        if "libs=static" in spec:
-            lib_args.append("install-static")
 
-        # install the library
-        make(*lib_args)
+class CMakeBuilder(CMakeBuilder):
```
scheibelp, line 64: If this now uses CMake on all platforms, a separate CMakeBuilder is not needed (like `c-blosc`).

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1206,3 +903,350 @@ def test(self):
...
+                self.command("-c", "import ssl")
+                self.command("-c", "import hashlib")
+
+            # Ensure that sqlite3 module works
+            if "+sqlite3" in spec:
+                self.command("-c", "import sqlite3")
+
+            # Ensure that dbm module works
+            if "+dbm" in spec:
+                self.command("-c", "import dbm")
```
scheibelp: I don't see the Windows build handling `~dbm`. If it's not supported or we don't want to handle that for now, then I think `conflicts(~dbm, when=platform=windows)` would be appropriate.

scheibelp (changes requested):
Most of these comments are very minor: I think when a package doesn't define multiple build systems it doesn't need to add a BuildSystem definition (I refer to `c-blosc` often as an example). Let me know if I overlooked some reason that needs to happen.

lib/spack/spack/build_environment.py
johnwparent: No, it was causing a CI precheck failure.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: Autotools has been removed.

var/spack/repos/builtin/packages/libogg/package.py
johnwparent: Understood, amended.

var/spack/repos/builtin/packages/libogg/package.py
johnwparent: Ok, that was not clear. I'll remove.

lib/spack/spack/build_environment.py
adamjstewart: @tgamblin fixed this in a separate PR

var/spack/repos/builtin/packages/c-blosc/package.py
johnwparent: I was under the impression we wanted to convert all packages to using the builder class model. Is that not the case?

var/spack/repos/builtin/packages/double-conversion/package.py
johnwparent: All c-blosc comments can refer to that response and thread as the answer will likely be the same and I don't want to spam with the same comment over and over. If there is a unique case I will address it at that time.

var/spack/repos/builtin/packages/libiconv/package.py
johnwparent: Good idea. Will do.

var/spack/repos/builtin/packages/c-blosc/package.py
adamjstewart: Only packages that support multiple build systems, not packages that support a single build system

johnwparent: Ok, that's my misunderstanding then. Thankfully I did most of these ports atomically commit-wise so I'll just drop those commits. Thanks @adamjstewart

var/spack/repos/builtin/packages/double-conversion/package.py
johnwparent: Issue addressed above.

var/spack/repos/builtin/packages/libpng/package.py
johnwparent: Must have gotten my own wire cross when creating the lists of which changes were made to what packages, it's in my multi build list but clearly that's not the case.

var/spack/repos/builtin/packages/libtheora/package.py
johnwparent: Sounds good.

var/spack/repos/builtin/packages/perl/package.py
johnwparent: Ok, wasn't sure where callbacks could go.

var/spack/repos/builtin/packages/netcdf-cxx/package.py
johnwparent: Clearly my tracking of what was done where was severely disorganized. I'll fix the PR description.

var/spack/repos/builtin/packages/python/package.py
johnwparent: It never has, this PR does nothing to change that behavior. The python build script on Windows has an incredibly limited CLI so many of the variants will have no impact one way or another on Windows. I added the conflict just to represent the fact that whether or not you specify the variant, you're getting dbm on Windows.

var/spack/repos/builtin/packages/zlib/package.py
```diff
@@ -138,4 +138,4 @@ def installtree(dst, tree):
                 for file in tree[inst_dir]:
                     install(file, install_dst)
 
-        installtree(self.prefix, install_tree)
+        installtree(prefix, install_tree)
```
scheibelp, line 141: We have a function called `install_tree` but not one called `installtree` - has this been confirmed to work?

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,30 +83,16 @@ class Proj(AutotoolsPackage):
...
-                args.append("--with-curl=" + self.spec["curl"].prefix.bin.join("curl-config"))
-            else:
-                args.append("--without-curl")
-
-        return args
+    def cmake_args(self):
+        return [
+            self.define("PROJ_LIB", join_path(self.stage.source_path, "nad")),
+            self.define_from_variant("ENABLE_TIFF", "tiff"),
+        ]
```
scheibelp, line 95: Are there similar options for enabling `curl`? If not then it might make sense to remove the `+curl` variant (unless CMake detects this automatically)

lib/spack/spack/build_systems/msbuild.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def install(self, pkg, spec, prefix):
+        """Run "msbuild" on the install targets specified by the builder.
+        This is INSTALL by default"""
+        with fs.working_dir(self.build_directory):
+            inspect.getmodule(self.pkg).msbuild(
+                *self.msbuild_install_args(), self.define_targets(*self.install_targets)
+            )
+
+    # def setup_dependent_package(self, m, dep_spec):
```
scheibelp: This seems preferable to setting the variables in `build_environment`: i.e. even if we don't have an `msbuild`/`cscript` package, recording that this is the build system that needs them is worthwhile. Do any problems occur when you uncomment this?

(likewise for the section in `nmake.py`)

var/spack/repos/builtin/packages/adios2/package.py
```diff
@@ -161,6 +172,45 @@ def setup_build_environment(self, env):
         elif self.spec.satisfies("%fj +fortran"):
             env.set("FFLAGS", "-Ccpp")
 
+    @property
+    def libs(self):
```
scheibelp, line 179: It appears that this and `setup_environment` moved but were not modified at all - can you move them back to minimize the diff (also helps with `git blame`)?

var/spack/repos/builtin/packages/expat/package.py
```diff
@@ -112,12 +112,19 @@ def url_for_version(self, version):
...
-        spec = self.spec
+    # def configure_args(self):
+    #     spec = self.spec
+    #     args = [
+    #         "--without-docbook",
+    #         "--enable-static",
+    #     ]
+    #     if "+libbsd" in spec and "@2.2.1:" in spec:
+    #         args.append("--with-libbsd")
+    #     return args
```
scheibelp, line 123: Commented code should be removed. In this case it looks like not all the `configure` options have been ported to `cmake`. I think it might be worth splitting this package to another PR to resolve this issue.

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -47,14 +47,7 @@ def headers(self):
...
-            "--with-png=yes",
-            "--with-zlib=no",
-        ]
-        if self.spec.satisfies("@2.9.1:"):
-            args.append("--enable-freetype-config")
-        return args
+
+class CMakeBuilder(CMakeBuilder):
+    def cmake_args(self):
+        return [self.define("FT_WITH_ZLIB", True), self.define("FT_WITH_BZIP2", True)]
```
scheibelp, line 54: Is there a similar CMake option for `libpng`?

var/spack/repos/builtin/packages/libpng/package.py
```diff
@@ -26,17 +26,8 @@ class Libpng(AutotoolsPackage):
...
+        return [
+            self.define("CMAKE_CXX_FLAGS", self.spec["zlib"].headers.include_flags),
+            self.define("ZLIB_ROOT", self.spec["zlib"].prefix),
         ]
-        return args
-
-    def check(self):
-        # Libpng has both 'check' and 'test' targets that are aliases.
-        # Only need to run the tests once.
-        make("check")
```
scheibelp, line 54: Does CMake have a similar option for running tests/checks after the build?

var/spack/repos/builtin/packages/libtheora/package.py
```diff
@@ -17,13 +21,16 @@ class Libtheora(AutotoolsPackage):
...
     depends_on("doxygen", when="+doc", type="build")
     depends_on("libogg")
     depends_on("libpng")
+    with when("build_system=autotools"):
+        depends_on("autoconf", type="build")
+        depends_on("automake", type="build")
+        depends_on("libtool", type="build")
+        depends_on("m4", type="build")
+
+    build_system("scons", "autotools", default="autotools" if sys.platform == "win32" else "scons")
```
scheibelp: I think you mean `default="scons" if sys.platform == "win32" else "autotools"`

var/spack/repos/builtin/packages/lz4/package.py
```diff
@@ -44,6 +49,39 @@ def url_for_version(self, version):
...
 
+    def patch(self):
+        # Remove flags not recognized by the NVIDIA compiler
+        if self.spec.satisfies("%nvhpc@:20.11"):
+            filter_file("-fvisibility=hidden", "", "Makefile")
+            filter_file("-fvisibility=hidden", "", "lib/Makefile")
+            filter_file("-pedantic", "", "Makefile")
+
+    @run_after("install")
+    def darwin_fix(self):
```
scheibelp, line 60: I think this will need to move to a class that is inherited by the CMakeBuild and Makefile builder (like `libxml2`).

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -142,12 +145,62 @@ class NetcdfC(AutotoolsPackage):
 
     filter_compiler_wrappers("nc-config", relative_root="bin")
 
+    build_system("cmake", "autotools", default="cmake")
```
scheibelp, line 154: To minimize the effect of this PR on non-windows, IMO it would be worthwhile to have default build system match the prior chosen one where possible.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -142,12 +145,62 @@ class NetcdfC(AutotoolsPackage):
...
+        return base_cmake_args
+
+
+class AutotoolsBuilder(AutotoolsBuilder, BackupStep):
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        self.setup_run_environment(env)
+        # Some packages, e.g. ncview, refuse to build if the compiler path returned by nc-config
+        # differs from the path to the compiler that the package should be built with. Therefore,
+        # we have to shadow nc-config from self.prefix.bin, which references the real compiler,
+        # with a backed up version, which references Spack compiler wrapper.
```
scheibelp, line 202: This looks like it would apply for the CMake builder as well

var/spack/repos/builtin/packages/netcdf-cxx/package.py
```diff
@@ -26,8 +26,8 @@ def libs(self):
         shared = True
         return find_libraries("libnetcdf_c++", root=self.prefix, shared=shared, recursive=True)
 
-    def configure_args(self):
-        args = []
+    def cmake_args(self):
+        args = [self.define_from_variant("USE_NETCDF4", "netcdf4")]
         if "+netcdf4" in self.spec:
             # There is no clear way to set this via configure, so set the flag
             # explicitly
```
scheibelp, line 36: Does CMake also recognize `CPPFLAGS=` as an option?

var/spack/repos/builtin/packages/zstd/package.py
```diff
@@ -55,36 +57,18 @@ class Zstd(MakefilePackage):
...
 
-    def build(self, spec, prefix):
-        pass
-
-    def install(self, spec, prefix):
-        args = ["VERBOSE=1", "PREFIX=" + prefix]
-
-        # Tested %nvhpc@22.3. No support for -MP
-        if "%nvhpc" in self.spec:
-            args.append("DEPFLAGS=-MT $@ -MMD -MF")
```
scheibelp, line 66: It looks like this didn't get ported over to the CMake version.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1219,3 +800,464 @@ def test(self):
...
+                file_name.endswith(".exe")
+                or (file_name.endswith(".dll") and "python" in file_name)
+                or "vcruntime" in file_name
+            ):
+                copy(lib, prefix)
+            else:
+                copy(lib, prefix.DLLs)
+        static_libraries = glob.glob("%s\\*.lib")
+        for lib in static_libraries:
+            copy(lib, prefix.libs)
```
scheibelp, line 1264: Am I correct that this PR makes no changes in this package other than to move functionality into respective builder classes?

var/spack/repos/builtin/packages/perl/package.py
```diff
@@ -457,61 +367,170 @@ def filter_config_dot_pm(self):
...
+        if self.spec.satisfies("+threads"):
+            win_install_path += "-multi-thread"
+        else:
+            win_install_path += "-perlio"
+
+        for f in os.listdir(os.path.join(self.prefix.bin, win_install_path)):
+            lnk_path = os.path.join(self.prefix.bin, f)
+            src_path = os.path.join(win_install_path, f)
+            if not os.path.exists(lnk_path):
+                symlink(src_path, lnk_path)
```
scheibelp, line 542: Similar question here as for Python: does this PR make no other changes to this package other than to move functionality into respective builders?

scheibelp (changes requested):
Where possible, I think this PR might benefit from a smoother transition for build systems on non-Windows by preserving the old build system where possible. This would resolve some of the comments (i.e. in cases where I ask whether a given `configure` option has an anolog in CMake); beyond that though it might mitigate unforeseen issues with changing the build system.

lib/spack/spack/build_systems/msbuild.py
scheibelp: If build systems other than `msbuild` use these, then I can see the appeal of defining them in `build_environment` (similar to why `make` is defined there).

johnwparent: Only msbuild/nmake uses these, although there are currently generic classes using them, but those should be refactored. However, when I switched to using it this way, all nmake and msbuild `build`  builder phases fail because `nmake` or `msbuild` respectively are not defined, and it seems like this method is not invoked before the build stage. It seems as though other build tools either have packages to represent them, allowing for these tools to be defined when the package is loaded, or are defined in the `build_environment` module. According to the docs, `setup_dependent_packages` is only invoked before `install`, which makes it unsuitable for use in this case, as we need it during the `build` phase. If this is not the case, then something is broken and the docs need updating. If this is the case, can we implement something that does run before `build` or `configure`? Or maybe this is more configurable than I think and I'm not seeing it in the build system docs.

scheibelp: `setup_dependent_packages` (and all `*setup*` functions) should run before all phases and it is indeed a bug if they are not run before the `build` phase - can you mention the error message here?

johnwparent: The error:
```
  File "C:\spack_win\spack\lib\spack\spack\build_environment.py", line 1053, in _setup_pkg_and_run
  File "C:\spack_win\spack\lib\spack\spack\installer.py", line 2081, in build_process
    return installer.run()
  File "C:\spack_win\spack\lib\spack\spack\installer.py", line 1945, in run
    self._real_install()
  File "C:\spack_win\spack\lib\spack\spack\installer.py", line 2041, in _real_install
    phase_fn.execute()
  File "C:\spack_win\spack\lib\spack\spack\builder.py", line 420, in execute
    self.phase_fn(pkg, pkg.spec, pkg.prefix)
  File "C:\spack_win\spack\var\spack\repos\builtin\packages\winlibiconv\package.py", line 37, in build
NameError: name 'nmake' is not defined
```

When I include a line that, if the `setup_dependent_package` is invoked will exit with failure, nothing happens and execution continues as normal. Perhaps I'm doing something wrong because even if I remove the contents of the build method, the install method fails the same way, so either I've done something wrong, or this method is never being invoked. Considering I have a `raise` and an `exit(1)` in that stanza and nothing happens, I feel like it might never be invoked.

scheibelp: Whoops sorry it just occurred to me: this should go into `setup_package`, not `setup_dependent_package`: if you have A->B, then setup_dependent_package is run for B when building A, but not for B when building B. Does that make sense?

johnwparent: It does make sense, and I did think it was a bit of a bizarre phrasing for the function name. I just assumed it was for "packages dependent on the builder". That being said, switching to `setup_package` changes nothing, nor can I find a reference to that method being invoked anywhere except for a `build_environment` context. It's a method in `build_environment` that invokes some environmental information about the package, but doesn't seem to be invoked by anything that would call our packages `setup_package` method, at least according to a grep of the codebase. 

From what I can tell, we rely on `setup_dependent_package` for almost every build system but autotools/makefiles and maven, as they all have representative packages. Autotools pulls from `build_environment` as we know, and maven runs a `which` during the install phase. I'm happy to add stub packages for nmake and msbuild that don't install, but do allow for external detection but being required to do that feels like a bug with the way things are currently structured.

Maybe we should sync about this later today?

var/spack/repos/builtin/packages/python/package.py
johnwparent: In general, yes. I was under the impression that we wanted to move packages like this to multiple builders, and as its a dep of paraview, I thought it was warranted under the scope of this PR.

var/spack/repos/builtin/packages/perl/package.py
johnwparent: There are also changes w.r.t. to how Perl gets its MSVC toolchain version as I refactored the underlying OS class. But for the most part correct. I have a similar response to this as I do to the python comment.

var/spack/repos/builtin/packages/zstd/package.py
johnwparent: Discussed over slack. Will amend as we decided there.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: I didn't see one bizarrely enough, I thought that was weird as well. I'll fix as we discussed over slack.

var/spack/repos/builtin/packages/expat/package.py
johnwparent: Yes, I left configure sections that I was not able to replicate 1-1 in cmake's configure stage commented out for the most part to provide context for the missing args. Will amend as discussed on Slack.

var/spack/repos/builtin/packages/proj/package.py
johnwparent: I didn't see them. I think I'll refactor this one to retain the autotools build on non Windows platforms pending further investigation at a later date.

var/spack/repos/builtin/packages/zlib/package.py
johnwparent: It's installed on my local system from this pr.... But you're absolutely correct, that's not the name of the function defined in the filesystem module. Very confused about that, will have to investigate...

var/spack/repos/builtin/packages/libpng/package.py
johnwparent: It depends if the requisite target's have been defined/ctest was used. For this package in particular I don't know. I'll be addressing as discussed.

var/spack/repos/builtin/packages/libtheora/package.py
johnwparent: Yes, I reversed that, good catch,

var/spack/repos/builtin/packages/zlib/package.py
johnwparent: The `installtree` method was something I implemented some time ago when porting zlib for the first time. I'm not sure if its still needed, or why it was implemented in the first place as the individual commit message was lost to the Windows super PR. I'll try replacing it with the `filesystem.install_tree` method and see if that works.

lib/spack/spack/build_systems/msbuild.py
scheibelp: @alalazo it turns out that for now we need to keep defining `cscript`/`msbuild` in `build_environment`:

* We're able to customize CMake because packages which use it depend on the `cmake` Spack package, which does `setup_dependent_package`
* There is currently no Spack package which provides `msbuild`/`cscript`
* We don't have a way for build systems to indicate that they should modify the package module, i.e. there is no `setup_package` analog to `setup_dependent_package` (to avoid confusion, there *is* a method called `setup_package` but it is not used for this purpose)

For now I think the path forward is to:

* Define `msbuild`/`cscript` in `build_environment`
* Add a TODO message there
* In a follow-up PR, define a dependency package, or update the API to allow build systems to modify the module

Does that sound reasonable?

johnwparent:
Closed in favor of multiple smaller PRs

PR #33565, opened 2022-10-27
Add rocblas, cublas, and gpu-aware-mpi variants to hypre

jrood-nrel:
Also update some things with Umpire in hypre.

@psakievich @eugeneswalker 

@PaulMullowney maybe you could comment on needing to remove `--enable-device-memory-pool` when using Umpire?

### DIFF
--- a/var/spack/repos/builtin/packages/hypre/package.py
+++ b/var/spack/repos/builtin/packages/hypre/package.py
@@ -100,6 +100,7 @@ class Hypre(AutotoolsPackage, CudaPackage, ROCmPackage):
     depends_on("rocsparse", when="+rocm")
     depends_on("rocthrust", when="+rocm")
     depends_on("rocrand", when="+rocm")
+    depends_on("rocprim", when="+rocm")
     depends_on("umpire", when="+umpire")
     for sm_ in CudaPackage.cuda_arch_values:
         depends_on(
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/hypre/package.py
```diff
@@ -112,6 +115,8 @@ class Hypre(AutotoolsPackage, CudaPackage, ROCmPackage):
     conflicts("+cuda", when="+int64")
     conflicts("+rocm", when="+int64")
     conflicts("+rocm", when="@:2.20")
+    conflicts("+cublas", when="~cuda", msg="cuBLAS requires CUDA to be enabled")
+    conflicts("+rocblas", when="~rocm", msg="rocBLAS requires ROCm to be enabled")
```
psakievich, line 120: We should also probably add a conflict for `gpu-aware-mpi` here too.

PaulMullowney: What do you have in mind for the conflict? gpu-aware-mpi works on cuda and rocm, though you need the most recent architectures/devices on each.

psakievich: @PaulMullowney we can add those architectures specifics as requirements. Can you provide what the minimum devices that support this feature are?

var/spack/repos/builtin/packages/hypre/package.py
```diff
@@ -74,6 +74,9 @@ class Hypre(AutotoolsPackage, CudaPackage, ROCmPackage):
     variant("fortran", default=True, description="Enables fortran bindings")
     variant("gptune", default=False, description="Add the GPTune hookup code")
     variant("umpire", default=False, description="Enable Umpire support")
+    variant("gpu-aware-mpi", default=False, description="Use GPU-aware MPI")
+    variant("rocblas", default=False, description="Use rocBLAS")
```
osborn9, line 78: While rocblas is a valid configure option, hypre is *not* explicitly using rocBLAS so this variant may be misleading to users.

alalazo:
What is the status if this PR?

jrood-nrel:
I will just close it.

PR #33566, opened 2022-10-27
GDAL: multi-build system support

adamjstewart:
Tested the following configurations on macOS 12.6.1 (arm64) with Apple Clang 14.0.0:

* GDAL 3.5 (CMake)
* GDAL 3.5 (Autotools)
* GDAL 3.4 (Autotools)

All configurations build successfully and pass all build tests and post-install tests.

### DIFF
--- a/var/spack/repos/builtin/packages/gdal/package.py
+++ b/var/spack/repos/builtin/packages/gdal/package.py
@@ -228,7 +228,6 @@ class Gdal(CMakePackage, AutotoolsPackage, PythonExtension):
     )
 
     with when("build_system=cmake"):
-        generator = "Ninja"
         depends_on("cmake@3.9:", type="build")
         depends_on("ninja", type="build")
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/gdal/package.py
```diff
@@ -194,14 +216,26 @@ class Gdal(CMakePackage):
...
 
+    # Build system
+    build_system(
+        conditional("cmake", when="@3.5:"),
+        conditional("autotools", when="@:3.5"),
+        default="cmake",
+    )
+
+    with when("build_system=cmake"):
+        generator = "Ninja"
```
adamjstewart: @alalazo does the `generator` parameter belong in Package or Builder? At the moment it's in both in this PR.

alalazo: It should be on the builder.

adamjstewart: Confirmed that it works as expected.

alalazo (approved):

PR #33577, opened 2022-10-28
Added a dependecy to using Libfabric on Cray systems to take advantage

bvanessen:
of native messaging.

### DIFF
--- a/var/spack/repos/builtin/packages/hdf5/package.py
+++ b/var/spack/repos/builtin/packages/hdf5/package.py
@@ -196,6 +196,7 @@ class Hdf5(CMakePackage):
 
     depends_on("cmake@3.12:", type="build")
 
+    depends_on("libfabric", when="^cray-mpich")
     depends_on("mpi", when="+mpi")
     depends_on("java", type=("build", "run"), when="+java")
     depends_on("szip", when="+szip")


haampie:
libfabric is a dependency of cray-mpich itself (in some cases). Are you adding it because hdf5 needs it, or because the Cray PE is broken and can't locate its own dependencies?

bvanessen:
> libfabric is a dependency of cray-mpich itself (in some cases). Are you adding it because hdf5 needs it, or because the Cray PE is broken and can't locate its own dependencies?

@haampie I guess I don't really know the right answer here.  What I do know is that it is probably is a fight between how the Cray PE sets everything up and how Spack tries to take over and get away from depending on modules.  In the scenario that I am working on, the cray-mpich library is found as an external and `hdf5` isn't even able to complete configuration in `cmake`.  Additionally, I haven't seen this one directly yet, but my experience is that a standard manual build without spack will succeed.

bvanessen:
> > libfabric is a dependency of cray-mpich itself (in some cases). Are you adding it because hdf5 needs it, or because the Cray PE is broken and can't locate its own dependencies?
> 
> @haampie I guess I don't really know the right answer here. What I do know is that it is probably is a fight between how the Cray PE sets everything up and how Spack tries to take over and get away from depending on modules. In the scenario that I am working on, the cray-mpich library is found as an external and `hdf5` isn't even able to complete configuration in `cmake`. Additionally, I haven't seen this one directly yet, but my experience is that a standard manual build without spack will succeed.

Also, @haampie I am not seeing that `libfabric` is a dependency in `cray-mpich`  am I missing something?

haampie:
I mean that Cray's `libmpi.so` has `libfabric.so` among needed libs (try `readelf -d /path/to/libmpi.so`), not that Spack's cray-mpich package has a `depends_on("libfabric")`.

Last time I used a cray system, you could not use `module use cray-mpich` on its own, you had to also have `module load libfabric`, exactly because Cray's PE is :hankey: and doesn't know how to locate its own dependencies. The libfabric module would basically only set `LD_LIBRARY_PATH` to make libmpi.so find its deps.

Now, Spack may set up modules in the build env starting with a clean slate, so possibly libfabric isn't loaded then.

If that's (still) the issue, then this fix isn't right, since it's really a cray pe issue, not a spack issue. You would have to set `modules: [...]` to include both cray-mpich and libfabric for external cray-mpich.

bvanessen:
> I mean that Cray's `libmpi.so` has `libfabric.so` among needed libs (try `readelf -d /path/to/libmpi.so`), not that Spack's cray-mpich package has a `depends_on("libfabric")`.
> 
> Last time I used a cray system, you could not use `module use cray-mpich` on its own, you had to also have `module load libfabric`, exactly because Cray's PE is 💩 and doesn't know how to locate its own dependencies. The libfabric module would basically only set `LD_LIBRARY_PATH` to make libmpi.so find its deps.
> 
> Now, Spack may set up modules in the build env starting with a clean slate, so possibly libfabric isn't loaded then.
> 
> If that's (still) the issue, then this fix isn't right, since it's really a cray pe issue, not a spack issue. You would have to set `modules: [...]` to include both cray-mpich and libfabric for external cray-mpich.

@haampie So I agree that this PR is not going to be a "correct" fix, so I will mark it as don't merge, but the real issue is if I have a dependency on a spack external, I would expect to have it properly satisfy the actual dependencies, and right now it isn't.  So if `cray-mpich` can depend on `libfabrics` shouldn't that be reflected in the recipe, similarly to how `openmpi` is setup?

bvanessen:
> > I mean that Cray's `libmpi.so` has `libfabric.so` among needed libs (try `readelf -d /path/to/libmpi.so`), not that Spack's cray-mpich package has a `depends_on("libfabric")`.
> > Last time I used a cray system, you could not use `module use cray-mpich` on its own, you had to also have `module load libfabric`, exactly because Cray's PE is 💩 and doesn't know how to locate its own dependencies. The libfabric module would basically only set `LD_LIBRARY_PATH` to make libmpi.so find its deps.
> > Now, Spack may set up modules in the build env starting with a clean slate, so possibly libfabric isn't loaded then.
> > If that's (still) the issue, then this fix isn't right, since it's really a cray pe issue, not a spack issue. You would have to set `modules: [...]` to include both cray-mpich and libfabric for external cray-mpich.
> 
> @haampie So I agree that this PR is not going to be a "correct" fix, so I will mark it as don't merge, but the real issue is if I have a dependency on a spack external, I would expect to have it properly satisfy the actual dependencies, and right now it isn't. So if `cray-mpich` can depend on `libfabrics` shouldn't that be reflected in the recipe, similarly to how `openmpi` is setup?

So setting an external with the following definition does not solve the `cmake` problem.
```
    cray-mpich:
      buildable: false
      version:
      - 8.1.19
      externals:
      - spec: cray-mpich@8.1.19 arch=cray-sles15-zen3
        modules:
        - cray-mpich/8.1.19
        - libfabric/1.15.0
 ```

bvanessen:
> > > I mean that Cray's `libmpi.so` has `libfabric.so` among needed libs (try `readelf -d /path/to/libmpi.so`), not that Spack's cray-mpich package has a `depends_on("libfabric")`.
> > > Last time I used a cray system, you could not use `module use cray-mpich` on its own, you had to also have `module load libfabric`, exactly because Cray's PE is 💩 and doesn't know how to locate its own dependencies. The libfabric module would basically only set `LD_LIBRARY_PATH` to make libmpi.so find its deps.
> > > Now, Spack may set up modules in the build env starting with a clean slate, so possibly libfabric isn't loaded then.
> > > If that's (still) the issue, then this fix isn't right, since it's really a cray pe issue, not a spack issue. You would have to set `modules: [...]` to include both cray-mpich and libfabric for external cray-mpich.
> > 
> > 
> > @haampie So I agree that this PR is not going to be a "correct" fix, so I will mark it as don't merge, but the real issue is if I have a dependency on a spack external, I would expect to have it properly satisfy the actual dependencies, and right now it isn't. So if `cray-mpich` can depend on `libfabrics` shouldn't that be reflected in the recipe, similarly to how `openmpi` is setup?
> 
> So setting an external with the following definition does not solve the `cmake` problem.
> 
> ```
>     cray-mpich:
>       buildable: false
>       version:
>       - 8.1.19
>       externals:
>       - spec: cray-mpich@8.1.19 arch=cray-sles15-zen3
>         modules:
>         - cray-mpich/8.1.19
>         - libfabric/1.15.0
> ```

@tgamblin @becker33 @scheibelp any thoughts on how to support this?

PR #33583, opened 2022-10-28
gitlab ci: patched make 4.3.0

haampie:
Add the bugfix we also have in #33578


alalazo (approved):

alalazo (approved):
Tried that in a container. Seems to fix the issues with `libcap` and `tau` as intended:
Screenshot from 2022-10-28 14-37-15

PR #33589, opened 2022-10-28
Build_systems tests on windows

markus-ferrell:
Removes windows skips for non-autotools dependent tests. Autotools tests are skipped because they will not be supported on windows.

### DIFF
--- a/lib/spack/spack/build_systems/cmake.py
+++ b/lib/spack/spack/build_systems/cmake.py
@@ -248,7 +248,8 @@ class CMakeBuilder(BaseBuilder):
     @staticmethod
     def std_args(pkg, generator=None):
         """Computes the standard cmake arguments for a generic package"""
-        generator = generator or "Unix Makefiles"
+        default_generator = "Ninja" if sys.platform == "win32" else "Unix Makefiles"
+        generator = generator or default_generator
         valid_primary_generators = ["Unix Makefiles", "Ninja"]
         primary_generator = _extract_primary_generator(generator)
         if primary_generator not in valid_primary_generators:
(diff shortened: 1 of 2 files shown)


markus-ferrell:
@spackbot fix style

lib/spack/spack/build_systems/cmake.py
```diff
@@ -213,7 +213,7 @@ def std_cmake_args(self):
         return std_cmake_args
 
     @staticmethod
-    def std_args(pkg, generator=None):
+    def std_args(pkg, generator=generator):
```
johnwparent: What happens if we leave this?

markus-ferrell: The value set [above ](https://github.com/spack/spack/blob/74466b9d2f6a169a07c1439d44872d8fa9da227a/lib/spack/spack/build_systems/cmake.py#L182) gets overridden by the None default which leads to Windows getting the default "Unix Makefiles" generator rather than "Ninja"

johnwparent: If we're always going to be providing a generator then, I think it makes sense to remove that check for none, or to remove the optional argument and make both positional.

markus-ferrell: I think removing the null check makes more sense. Making them positional would lead changing quite a few other places to know about what generator to use.

markus-ferrell:
@spackbot fix style

lib/spack/spack/test/build_systems.py
```diff
@@ -45,6 +43,7 @@ def _func(dir_str):
     return _func
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="autotools not available on Windows")
```
markus-ferrell: Note: Change `autotools` to `make`

scheibelp (approved):

PR #33591, opened 2022-10-28
libtiff: CMake support, internal codecs variants

adamjstewart:
This PR adds several features to the libtiff package:

- [x] Can now be built with CMake, and therefore supports Windows
- [x] Adds v4.0.4 and v4.0.5
- [x] Adds variants to toggle internal codecs

I've tested this on macOS 12.6.1 with Apple Clang 14.0.0 for several versions:

- [x] 4.4
- [ ] 4.3 (doesn't support macOS)
- [x] 4.2
- [x] 4.1
- [x] 4.0
- [x] 3.9

@alalazo @johnwparent

### DIFF
--- a/var/spack/repos/builtin/packages/libtiff/package.py
+++ b/var/spack/repos/builtin/packages/libtiff/package.py
@@ -3,10 +3,34 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
 
+VARIANTS = [
+    # Internal codecs
+    "ccitt",
+    "packbits",
+    "lzw",
+    "thunder",
+    "next",
+    "logluv",
+    # External codecs
+    "zlib",
+    "libdeflate",
+    "pixarlog",
+    "jpeg",
+    "old-jpeg",
+    "jpeg12",
+    "jbig",
+    "lerc",
+    "lzma",
+    "zstd",
+    "webp",
+]
 
-class Libtiff(AutotoolsPackage):
+
+class Libtiff(CMakePackage, AutotoolsPackage):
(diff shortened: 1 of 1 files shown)


johnwparent:
Looks great from our end. Local testing gives the all clear. This is also one less thing for me to transition to the new multi-build system in the paraview port too and a great example to work from. Thanks @adamjstewart!

johnwparent (approved):

var/spack/repos/builtin/packages/libtiff/package.py
```diff
@@ -3,10 +3,34 @@
...
+    "jbig",
+    "lerc",
+    "lzma",
+    "zstd",
+    "webp",
+]
 
-class Libtiff(AutotoolsPackage):
+
+class Libtiff(CMakePackage, AutotoolsPackage):
```
alalazo, line 33: A tiny caveat:
```suggestion
class Libtiff(AutotoolsPackage, CMakePackage):
```
This would matter only for packages that have been installed _before_ #30738 and get reused after #30738. If CMake is first in the list they'll be guessed as CMake packages, inverting the two as Autotools packages.

adamjstewart: What would break without this change?

alalazo: In this specific case I think nothing, so I can go ahead merging this. In general if a build system has a `setup_dependent_build_environment` or `setup_dependent_package` they will not be taken into account properly in that specific reuse case (which is transitory and will "converge to non-existence" with time).

adamjstewart: Good to know. I'll try to watch out for that in the future. I think we can merge this as is.

alalazo: Also, to be clear, for every spec with a `build_system` variant (i.e. every spec installed after #30738) the change above would have no effect.

alalazo (approved):

PR #33598, opened 2022-10-28
exago: query and use MPI compilers from spack

balay:


balay:
@balos1 checking for the correct way to specify MPI to cmake..

cameronrutherford (dismissed):
I am still unclear if this is masking an underlying CMake configuration in ExaGO, or if this is the appropriate workaround for errors in xSDK pipelines.

I see no problem with merging this for the moment.

balay:
> I am still unclear if this is masking an underlying CMake configuration in ExaGO, or if this is the appropriate workaround for errors in xSDK pipelines.

In  a spack build - exago should not guess compilers [i.e not  pick up random stuff from PATH - that will break builds ].

The correct compilers should be queried from spack - and passed in as CMAKE arguments to exago. i.e this is primarily an issue with the current spack spec for exago. [and this PR fixes it]

There could be other underlying issues in exago cmake code - [if its ignoring - or unable to process these cmake compiler options..]

balos1:
You should not have to set the CMAKE_C_COMPILER and CMAKE_CXX_COMPILER to the MPI compilers if you are using the `MPI` targets in relatively recent version of CMake (they were a bit buggy when first introduced in cmake@3.12). I think this PR is fine though.

cameronrutherford:
@balos1 we are using `find_package(MPI)` within ExaGO CMake and this has never been required before.

For some reason PETSc and ExaGO were picking up different MPI versions in xSDK CI, and this seems to be the only workaround.

EDIT: we end up linking against `MPI::MPI_C` and `MPI::MPI_CXX` targets

eugeneswalker (dismissed):

eugeneswalker (approved):

balay:
I think this PR is ready for merge.

PR #33590, opened 2022-10-28
_replace_prefix_bin performance improvements

haampie:
- single pass over the binary data matching all prefixes
- collect offsets and replacement strings
- do in-place updates with `fseek` / `fwrite`, since typically our
  replacement touch O(few bytes) while the file is O(many megabytes)
- be nice: leave the file untouched if some string can't be
  replaced

Note: the order of ordered dict is preserved, since the regex will greedily match the left-hand side of `|` before backtracking and matching the rhs.

PR #33608, opened 2022-10-30
Speed up relocation and fix bug

haampie:
- `relocate_links` was skipped for macOS
- Remove the sequential filter of binaries (very slow)

LLVM before:

```
elf      47.34s
text      0.01s
filter   84.72s
text bin  0.003s
```

LLVM after:

```
elf      13.37s # not sure why this is so flaky...
text      0.01s
text bin  4.81s
```

All in all llvm installs in < 1 min, the bottleneck is decompressing the tarball now.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -1644,27 +1644,15 @@ def relocate_package(spec, allow_root):
                 old_prefix,
                 new_prefix,
             )
-            # Relocate links to the new install prefix
-            links = [link for link in buildinfo.get("relocate_links", [])]
-            relocate.relocate_links(links, old_layout_root, old_prefix, new_prefix)
+
+        # Relocate links to the new install prefix
+        links = [link for link in buildinfo.get("relocate_links", [])]
+        relocate.relocate_links(links, old_layout_root, old_prefix, new_prefix)
 
         # For all buildcaches
         # relocate the install prefixes in text files including dependencies
         relocate.unsafe_relocate_text(text_names, prefix_to_prefix_text)
 
-        paths_to_relocate = [old_prefix, old_layout_root]
-        paths_to_relocate.extend(prefix_to_hash.keys())
-        files_to_relocate = list(
-            filter(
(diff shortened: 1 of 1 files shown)


eugeneswalker (approved):

lib/spack/spack/binary_distribution.py
```diff
@@ -1644,27 +1644,15 @@ def is_backup_file(file):
...
-            filter(
-                lambda pathname: not relocate.file_is_relocatable(
-                    pathname, paths_to_relocate=paths_to_relocate
-                ),
-                map(
-                    lambda filename: os.path.join(workdir, filename),
-                    buildinfo["relocate_binaries"],
-                ),
-            )
-        )
```
alalazo, line 1667: Before we were filtering `buildinfo["relocate_binaries"]`, so the semantic is different (not saying it was correct before). Why are we doing a lot of extra processing in `relocate.file_is_relocatable`? Does that processing cover some edge cases or is it plain wrong or unneeded?

haampie: This can safely be dropped, currently it's horrible, and removing it makes Spack only better.

The old logic is as follows: take all strings minus the rpath from a binary (since `patchelf` can handle that). If among this list you find `old_prefix` or `old_layout_root`, it means this binary needs to be patched.

This is ridiculous because:

a. it's testing only *two of many* strings searched from the `prefix_to_prefix_bin` map, so it's simply incorrect;
b. it's sequential;
c. it adds a dependency on binutil's `strings` executable;
d. currently we only ever call binary text replacement *after* rpath substitution, so there's no need to special case rpaths.

Maybe in the past this filter was useful for performance reasons (calling `strings` is faster than scanning for strings in a binary using Python?), but clearly that's no longer true after #33590

alalazo (approved):

gartung:
@haampie thank you for fixing my bone-headed implementation.

gartung:
If I recall correctly I was trying to preserve backwards compatibility.

PR #33609, opened 2022-10-30
PETSc: make petsc depends on bison when ptscotch is on

lrtfm:


balay:
Hm - this extra dependency is a bug, looking at a different fix:

https://gitlab.com/petsc/petsc/-/issues/1276

lrtfm:
Thanks, close this one.

PR #33614, opened 2022-10-31
Fix pkgconfig dependencies

michaelkuhn:
Packages should depend on the virtual provider, pkgconfig, not on its implementations pkg-config or pkgconf.

### DIFF
--- a/var/spack/repos/builtin/packages/aml/package.py
+++ b/var/spack/repos/builtin/packages/aml/package.py
@@ -76,6 +76,7 @@ class Aml(AutotoolsPackage):
         depends_on("automake", type="build")
         depends_on("libtool", type="build")
         # Required to have pkg config macros in configure.
+        # Note: This does NOT work with pkg-config but requires pkgconf!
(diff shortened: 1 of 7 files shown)


robertu94:
This looks good to me. I should be a maintainer for LibPressio and Speer, but I forgot to add myself.

perarnau:
That's a no for me on aml. Building master requires pkgconf-specific m4 rules for autoconf.

michaelkuhn:
> That's a no for me on aml. Building master requires pkgconf-specific m4 rules for autoconf.

Thanks for the hint! I reverted the change and added a comment about the incompatibility.

shaomeng:
This looks good to me too. Thanks to everyone!

tldahlgren (approved):
LGTM.  I appreciate the switch to the virtual `pkgconfig` provided by both `pkgconf` and `pkg-config` when allowed.

wanlinwang (approved):
It looks also ok for me. Thank you！

PR #33617, opened 2022-10-31
ParaView: ParaView needs to set the HDF5 API

kwryankrattiger:
When building ParaView with a newer HDf5 than 1.10, it needs to select the 1.10 API using flags.

### DIFF
--- a/var/spack/repos/builtin/packages/paraview/package.py
+++ b/var/spack/repos/builtin/packages/paraview/package.py
@@ -311,6 +311,15 @@ class Paraview(CMakePackage, CudaPackage):
         if (name == "cflags" or name == "cxxflags") and self.spec.satisfies("%intel"):
             flags.append("-no-ipo")
             return (None, None, flags)
+
+        if name in ("cflags", "cxxflags"):
+            # Constrain the HDF5 API
+            if self.spec.satisfies("@:5.9 +hdf5"):
+                if self.spec["hdf5"].satisfies("@1.10:"):
+                    flags.append("-DH5_USE_18_API")
+            elif self.spec.satisfies("@5.10: +hdf5"):
+                if self.spec["hdf5"].satisfies("@1.12:"):
+                    flags.append("-DH5_USE_110_API")
         return (flags, None, None)
 
     def setup_run_environment(self, env):


var/spack/repos/builtin/packages/paraview/package.py
```diff
@@ -311,6 +311,15 @@ def flag_handler(self, name, flags):
         if (name == "cflags" or name == "cxxflags") and self.spec.satisfies("%intel"):
             flags.append("-no-ipo")
             return (None, None, flags)
+
+        if name in ("cflags", "cxxflags"):
+            if "+hdf5" in self.spec:
+                if "@:5.9" in self.spec:
```
tldahlgren:
```suggestion
                if self.spec.satisfies("@:5.9"):
```

var/spack/repos/builtin/packages/paraview/package.py
```diff
@@ -311,6 +311,15 @@ def flag_handler(self, name, flags):
...
+                    if "@1.10:" in self.spec["hdf5"]:
+                        flags.append("-DH5_USE_18_API")
+                if "@5.10:" in self.spec:
+                    if "@1.13:" in self.spec["hdf5"]:
```
tldahlgren: Suggest same use of `spec.satisfies()` for version range checks at 318, 320, and 321.

tldahlgren (commented):
Suggest switching to use of `satisfies()` when checking version ranges of specs.  See https://spack.readthedocs.io/en/latest/packaging_guide.html#testing-spec-constraints.

var/spack/repos/builtin/packages/paraview/package.py
kwryankrattiger: I generally avoid using `satisfies` in `package.py`s, the `<range/variant> in <spec>` format feels more "python"-y and is the prevalent method in the packages I work on. Although, if there is a stronger reason to use satisfies, I am happy to change my thinking.

kwryankrattiger: just noticed there was a link to docs 🤦🏽

mathstuf (dismissed):
Looks fine to me after how to spell the logic gets hammered out. Note that VTK should probably just do this on its own in the `VTK::hdf5` module logic…

tldahlgren (dismissed):
Thanks.  LGTM.

tldahlgren:
@spackbot fix style

tldahlgren (approved):

PR #33622, opened 2022-10-31
Rm unneeded call to add dll to PATH on win

johnwparent:


scheibelp (approved):

scheibelp:
@KiruyaMomochi note that since #33400 has been merged, the changes there to the `clingo` package mean that the call to `os.add_dll_directory` added by #32942 is no-longer necessary.

However, you may need to re-bootstrap clingo: the changes in #33400 are based on installing `clingo` in a particular way, and therefore wouldn't work for Spack installations on Windows that were bootstrapped before it was merged. To do that, you can `spack clean -b` (this clears out the packages installed in the bootstrap store, which will be reinstalled in a future Spack command).

PR #33628, opened 2022-10-31
add elf parsing utility function

haampie:
This PR introduces some basic ELF parsing logic, which is useful in various
places.

Example uses:

- Retrieve all ELF executables and shared libs in a prefix
- Get the offset of the rpath string in a file so it can be modified in-place
- Get better info about why a file has no rpath (e.g. not an elf file,
  malformed elf, no dynamic section, missing tag in dynamic array, ...), this
  is more fine-grained than in patchelf.

It's very basic, and it's really about parsing, since we don't want to reimplement
the advanced dynamic section growth logic that patchelf offers.

Current API is basically:

```python
import spack.util.elf as elf

try:
    with open("my_binary", "br") as f:
        parsed = elf.parse_elf(f)
except elf.ElfParsingError:
    # handle invalid elf file

# booleans
parsed.has_pt_interp
parsed.has_pt_dynamic
parsed.has_rpath
parsed.has_needed
parsed.has_soname
parsed.is_runpath

# byte strings
parsed.pt_interp_str
parsed.dt_soname_str
parsed.dt_rpath_str
parsed.dt_needed_strs  # list of byte strings

# more low-level offsets into the file, or the strtab associated with the dynamic section (useful for in-place updates)
```

Another advantage is that we get a ~850x speedup for retrieving rpaths.

```python
In [1]: spack.relocate._elf_rpaths_for("libclang.so.15.0.0") == spack.util.elf.get_rpaths("libclang.so.15.0.0")
Out[1]: True

In [2]: %timeit spack.relocate._elf_rpaths_for("libclang.so.15.0.0")
42.8 ms ± 448 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %timeit spack.util.elf.get_rpaths("libclang.so.15.0.0")
50.4 µs ± 435 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

### DIFF
--- a/lib/spack/spack/test/util/elf.py
+++ b/lib/spack/spack/test/util/elf.py
@@ -1,0 +1,130 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+import io
+
+import pytest
+
+import llnl.util.filesystem as fs
+
+import spack.platforms
+import spack.util.elf as elf
+import spack.util.executable
+
+
+# note that our elf parser is platform independent... but I guess creating an elf file
+# is slightly more difficult with system tools on non-linux.
+def skip_unless_linux(f):
+    return pytest.mark.skipif(
+        str(spack.platforms.real_host()) != "linux",
+        reason="implementation currently requires linux",
+    )(f)
+
+
+@pytest.mark.requires_executables("gcc")
+@skip_unless_linux
+def test_elf_get_rpaths(binary_with_rpaths):
+    # Compile an "Hello world!" executable and set RPATHs
+    long_rpaths = ["/very/long/prefix/x", "/very/long/prefix/y"]
+    executable = str(binary_with_rpaths(rpaths=long_rpaths))
+    assert elf.get_rpaths(executable) == long_rpaths
+
+
+@pytest.mark.requires_executables("gcc")
+@skip_unless_linux
+@pytest.mark.parametrize(
+    "linker_flag,is_runpath",
+    [
+        ("-Wl,--disable-new-dtags", False),
+        ("-Wl,--enable-new-dtags", True),
+    ],
+)
+def test_elf_parsing_shared_linking(linker_flag, is_runpath, tmpdir):
+    gcc = spack.util.executable.which("gcc")
+
+    with fs.working_dir(str(tmpdir)):
+        # Create a library to link to so we can force a dynamic section in an ELF file
+        with open("foo.c", "w") as f:
+            f.write("int foo(){return 0;}")
+        with open("bar.c", "w") as f:
+            f.write("int foo(); int _start(){return foo();}")
+
+        # Create library and executable linking to it.
+        gcc("-shared", "-o", "libfoo.so", "-Wl,-soname,libfoo.so.1", "-nostdlib", "foo.c")
+        gcc(
+            "-o",
+            "bar",
+            linker_flag,
+            "-Wl,-rpath,/first",
+            "-Wl,-rpath,/second",
+            "-Wl,--no-as-needed",
+            "-nostdlib",
+            "libfoo.so",
+            "bar.c",
+            "-o",
+            "bar",
+        )
+
+        with open("libfoo.so", "rb") as f:
+            foo_parsed = elf.parse_elf(f, interpreter=True, dynamic_section=True)
+
+        assert not foo_parsed.has_pt_interp
+        assert foo_parsed.has_pt_dynamic
+        assert not foo_parsed.has_rpath
+        assert not foo_parsed.has_needed
+        assert foo_parsed.has_soname
+        assert foo_parsed.dt_soname_str == b"libfoo.so.1"
+
+        with open("bar", "rb") as f:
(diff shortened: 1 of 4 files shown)


lib/spack/spack/test/util/elf.py
```diff
@@ -0,0 +1,120 @@
...
+import spack.util.executable
+
+
+# note that our elf parser is platform independent... but I guess creating an elf file
+# is slightly more difficult with system tools on non-linux.
+def skip_unless_linux(f):
+    return pytest.mark.skipif(
+        str(spack.platforms.real_host()) != "linux",
+        reason="implementation currently requires linux",
+    )(f)
```
trws, line 24: What, no BSD and Illumos support? 😛 Really I think this is fine, but it makes me think we might want to have a "platform using elf" rather than "is_linux" check that we can use for things like this.

lib/spack/spack/test/util/elf.py
```diff
@@ -0,0 +1,120 @@
...
+    with pytest.raises(elf.ElfParsingError, match="ELF header malformed"):
+        elf.parse_elf(io.BytesIO(b"\x7fELF\x02\x01" + b"\x00" * 50))
+
+    # 32-bit needs at least 52 bytes of header; this is only 46 bytes
+    with pytest.raises(elf.ElfParsingError, match="ELF header malformed"):
+        elf.parse_elf(io.BytesIO(b"\x7fELF\x01\x01" + b"\x00" * 40))
+
+    # Not a ET_DYN/ET_EXEC on a 32-bit LE ELF
+    with pytest.raises(elf.ElfParsingError, match="Not an ET_DYN or ET_EXEC"):
+        elf.parse_elf(io.BytesIO(b"\x7fELF\x01\x01" + (b"\x00" * 10) + b"\x09" + (b"\x00" * 35)))
```
trws, line 120: This is fabulous, I may recruit your help to create some variations on these to go with my relo PR, love how clean this is.

lib/spack/spack/util/elf.py
```diff
@@ -0,0 +1,455 @@
...
+    f.seek(elf.elf_hdr.e_phoff)
+
+    # Here we have to make a mapping from virtual address to offset in the file.
+    program_header_fmt = elf.byte_order + ("LLQQQQQQ" if elf.is_64_bit else "LLLLLLLL")
+    ProgramHeader = ProgramHeader64 if elf.is_64_bit else ProgramHeader32
+    program_header_fmt_size = calcsize(program_header_fmt)
+
+    for _ in range(elf.elf_hdr.e_phnum):
+        data = read_exactly(f, program_header_fmt_size, "Malformed program header")
+        ph = ProgramHeader._make(unpack(program_header_fmt, data))
```
trws: Not a big deal, but why read_exactly in a loop rather than something like the following so we can benefit from a larger contiguous read?

```python
full_data = read_exactly(f, phfs * elf.elf_hdr.e_phnum, "badness")
unpack_fron(phf, full_data, i * phfs)
```

lib/spack/spack/util/elf.py
```diff
@@ -0,0 +1,455 @@
...
+    elf_header_fmt = elf.byte_order + ("HHLQQQLHHHHHH" if elf.is_64_bit else "HHLLLLLHHHHHH")
+    hdr_size = calcsize(elf_header_fmt)
+    data = read_exactly(f, hdr_size, "ELF header malformed")
+    elf.elf_hdr = ElfHeader._make(unpack(elf_header_fmt, data))
+
+
+def _do_parse_elf(f, interpreter=True, dynamic_section=True):
+    # We currently do not deal with ELF files with an offset in the file; offsets are
+    # always absolute.
+    f.seek(0)
```
trws: This should be harmless but confuses me slightly, the default for seek is to seek based on offset 0 rather than from current anyway, is there a specific issue you're trying to work around here?

trws:
I like this, a lot, and the test changes are really welcome.  Had a couple of questions but I don't see anything that would make me want to keep this out.

lib/spack/spack/util/elf.py
haampie: My idea was this: someone passes an `f` not at offset 0; they already ran read/seek. Then the parser starts reading the ELF header at a nonzero offset, and does absolute fseek based on what offets it finds in the header, which is wrong. Instead here we just require that the header itself starts at 0.

lib/spack/spack/util/elf.py
haampie: Sounds good, Python also has `iter_unpack` (but we can only use that after dropping python 2)

lib/spack/spack/test/util/elf.py
haampie: Agreed, all you need is a compiler/linker that can output elf files... But for now linux is fine I guess

lib/spack/spack/util/elf.py
haampie: Alternatively, I can also make it defensively throw if `f.tell() != 0`, that's probably better in case somehow we're parsing elf files in a tarfile or something, and we expect it to start parsing at the current offset, not at offset 0 :p

trws: Got it, makes sense to me. 👍

lib/spack/spack/util/elf.py
trws: Right, that would be nice too but we're a bit stuck there for now.

trws (approved):

PR #33631, opened 2022-11-01
package/py-pykml: add new py3 compatible version

Sinan81:
add 0.2.0 providing py3 compatibility. otherwise package tries to import urllib2 which doesn't exist in python3.

### DIFF
--- a/var/spack/repos/builtin/packages/py-pykml/package.py
+++ b/var/spack/repos/builtin/packages/py-pykml/package.py
@@ -14,7 +14,9 @@ class PyPykml(PythonPackage):
 
     pypi = "pykml/pykml-0.1.3.tar.gz"
 
+    version("0.2.0", sha256="44a1892e7c2a649c8ae9f8e2899ff76cae79ec6749ffb64d11140b4e87d0f957")
     version("0.1.3", sha256="e1a133e582f0b4652a6b00bac970b446d90580664e5a634a670731c990ff9f05")
 
+    depends_on("python@3", type=('build', 'run'), when='@0.2.0:')
     depends_on("py-setuptools", type="build")
     depends_on("py-lxml@2.2.6:", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-pykml/package.py
```diff
@@ -14,7 +14,9 @@ class PyPykml(PythonPackage):
 
     pypi = "pykml/pykml-0.1.3.tar.gz"
 
+    version("0.2.0", sha256="44a1892e7c2a649c8ae9f8e2899ff76cae79ec6749ffb64d11140b4e87d0f957")
     version("0.1.3", sha256="e1a133e582f0b4652a6b00bac970b446d90580664e5a634a670731c990ff9f05")
 
+    depends_on("python@3", type=('build', 'run'), when='@0.2.0:')
```
adamjstewart: Where do you see that pykml 0.2.0 doesn't support Python 2?

Also, can you add something so that it's clear that 0.1 requires Python 2? Then I can remove that version once Python 2 is removed.

var/spack/repos/builtin/packages/py-pykml/package.py
```diff
@@ -14,7 +14,9 @@ class PyPykml(PythonPackage):
...
     depends_on("py-setuptools", type="build")
     depends_on("py-lxml@2.2.6:", type=("build", "run"))
```
adamjstewart, line 23: Dependency version needs to be updated

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pykml/package.py
Sinan81: sorry for the confusion. it's fixed now.

var/spack/repos/builtin/packages/py-pykml/package.py
Sinan81: fixed.

adamjstewart:
@spackbot fix style

adamjstewart (dismissed):

Sinan81:
Can this be merged now?

adamjstewart (approved):

PR #33636, opened 2022-11-01
Relocate links using prefix to prefix map

haampie:
Previously symlinks were not relocated when they pointed across packages.

Proper fix so that workarounds like #31843 / #31862 are no longer required.


eugeneswalker (approved):

gartung:
@haampie thank you for fixing my bone-headed implementation.

gartung:
If I recall correctly I was trying to preserve backwards compatibility.

PR #33638, opened 2022-11-01
Add tracy 0.9

msimberg:
Also adds a conflict between pika 0.9 and older and tracy 0.9 and newer. The constraint is only until pika 0.9 because pika 0.10.0 will again compile with all versions of tracy.


teonnik (approved):

tldahlgren (approved):
Confirmed new sha256, rest LGTM.

PR #33639, opened 2022-11-01
elpa: Fix build on ppc64le

loveshack:


var/spack/repos/builtin/packages/elpa/package.py
```diff
@@ -135,13 +135,15 @@ def configure_args(self):
         options += self.with_or_without("mpi")
 
         # TODO: --disable-sse-assembly, --enable-sparc64, --enable-neon-arch64
-        simd_features = ["vsx", "sse", "avx", "avx2", "avx512", "sve128", "sve256", "sve512"]
+        # Don't include vsx; as of 2022.05 it fails (reported upstream).
```
haampie, line 138: Can you put a link in place of `(reported upstram)`?

haampie (changes requested):

loveshack:
> Can you put a link in place of `(reported upstram)`?

No, I just sent mail, as the maintainers suggest.

var/spack/repos/builtin/packages/elpa/package.py
bernhardkaindl: @loveshack Altivec VSX intrinsics are from POWER7 which was big-endian only. I think that it is likely that the VSX implementation in ELPA was never tested on little-endian whith ppc64le.

Were you have to also test AltiVec VSX in ELPA on big-endian ppc64?

If not, guess the better fix would be to check for `feature == "vsx" && self.spec.target.family != "ppc64le"` before adding `--enable-vsx`. What do you think @loveshack?

loveshack:
> @bernhardkaindl commented on this pull request.
>
>
>
>> @@ -135,13 +135,15 @@ def configure_args(self):
>          options += self.with_or_without("mpi")
>  
>          # TODO: --disable-sse-assembly, --enable-sparc64, --enable-neon-arch64
> - simd_features = ["vsx", "sse", "avx", "avx2", "avx512", "sve128",
> "sve256", "sve512"]
> +        # Don't include vsx; as of 2022.05 it fails (reported upstream).
>
> @loveshack Altivec VSX intrinsics are from POWER7 which was big-endian
> only. I think that it is likely that the VSX implementation in ELPA
> was never tested on little-endian whith ppc64le.
>
> Were you have to also test AltiVec VSX in ELPA on big-endian ppc64?
>
> If not, guess the better fix would be to check for `feature == "vsx"
> && self.spec.target.family != "ppc64le"` before adding
> `--enable-vsx`. What do you think @loveshack?

The intrinsics are definitely supported by gcc on LE POWER9.  (See info
node "PowerPC AltiVec/VSX Built-in Functions".)  They're just called
with the wrong signature in the instances that stopped the build
initially.  In fact, there are lots more that look like typos, and the
code would never compile, so it obviously hasn't been tested.  The
maintainer doesn't have access to a POWER system to fix it, and I doubt
it's worthwhile, given the SSE intrinsics in reasonably recent gcc.

bernhardkaindl (approved):
Thanks loveshack for the explanation on the state of the VSX code. Approving.
@haampie Your question should now be fully adressed, I think you can approve now as well.

haampie (approved):

PR #33643, opened 2022-11-01
py-pytorch-lightning: add v1.8.0

adamjstewart:
Successfully builds on macOS 12.6.1 (arm64) with Python 3.9.13 and Apple Clang 14.0.0.

https://github.com/Lightning-AI/lightning/releases

### DIFF
--- a/var/spack/repos/builtin/packages/py-pytorch-lightning/package.py
+++ b/var/spack/repos/builtin/packages/py-pytorch-lightning/package.py
@@ -14,6 +14,7 @@ class PyPytorchLightning(PythonPackage):
 
     maintainers = ["adamjstewart"]
 
+    version("1.8.0", sha256="deff9bc7978ecebc8f45e881adef65dc8d9f4554e88c3b064f80587f32ab158d")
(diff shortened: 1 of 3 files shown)


var/spack/repos/builtin/packages/py-pytorch-lightning/package.py
```diff
@@ -34,21 +35,26 @@ class PyPytorchLightning(PythonPackage):
     version("1.3.8", sha256="60b0a3e464d394864dae4c8d251afa7aa453644a19bb7672f5ee400343cdf7b0")
     version("1.2.10", sha256="2d8365e30ded0c20e73ce6e5b6028478ae460b8fd33727df2275666df005a301")
 
+    # src/pytorch_lightning/__setup__.py
```
tldahlgren, line 38: I definitely appreciate the addition of comments like this and the one below.  I think it can help educate ppl on where to find dependencies.

tldahlgren (approved):
Confirmed all 5 new sha256.  Rest LGTM

PR #33649, opened 2022-11-01
orca: new version 5.0.3

snehring:
They changed from zstd to xz for the archives, so that simplifies the install process a bit for newer versions. Strictly speaking the openmpi version specified in the release is 4.1.1, but I was having build issues with openmpi@4.1.1 with +pmi from slurm. It'd probably work with anything in 4.1 though. Some of the binaries link against a few libraries included in the bundle, hence the environment modification.

### DIFF
--- a/var/spack/repos/builtin/packages/orca/package.py
+++ b/var/spack/repos/builtin/packages/orca/package.py
@@ -19,8 +19,14 @@ class Orca(Package):
 
     homepage = "https://cec.mpg.de"
     url = "file://{0}/orca_4_0_1_2_linux_x86-64_openmpi202.tar.zst".format(os.getcwd())
+    maintainers = ["snehring"]
     manual_download = True
 
+    version(
+        "5.0.3",
+        sha256="b8b9076d1711150a6d6cb3eb30b18e2782fa847c5a86d8404b9339faef105043",
+        url="file://{0}/orca_5_0_3_linux_x86-64_shared_openmpi411.tar.xz".format(os.getcwd()),
+    )
     version(
         "4.2.1",
         sha256="9bbb3bfdca8220b417ee898b27b2885508d8c82799adfa63dde9e72eab49a6b2",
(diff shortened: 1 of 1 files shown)


RemiLacroix-IDRIS:
@snehring : Just double checking, they put the libraries in the `bin` subdir?

snehring:
@RemiLacroix-IDRIS In the distributed archive they're in the root with everything else, but given we install everything in prefix.bin they end up there.

RemiLacroix-IDRIS:
@snehring Ok, got it!

RemiLacroix-IDRIS:
> I was having build issues with openmpi@4.1.1 with +pmi from slurm

I'm able to compil OpenMPI 4.1.1 with +pmi and schedulers=slurm. Not sure if you should try to get this fixed and use 4.1.1 or if it useless as 4.1.2 will indeed probably work anyway... I leave that to @tldahlgren.

snehring:
Does it still work if you do fabrics=ofi? The actual issue seems to be in common_ofi.c with pmix.

RemiLacroix-IDRIS:
> Does it still work if you do fabrics=ofi? The actual issue seems to be in common_ofi.c with pmix.

Hmm, I'm able to build with OFI (but it's not what I usually use).

snehring:
Specifically the spec that fails for me is `openmpi@4.1.1~atomics~cuda~cxx~cxx_exceptions~gpfs~internal-hwloc~java+legacylaunchers~lustre~memchecker+pmi+romio+rsh~singularity+static+vt+wrapper-rpath build_system=autotools fabrics=ofi schedulers=slurm`.

RemiLacroix-IDRIS:
I don't have the latest version of Spack installed so I don't have the `build_system` variant but apart from that it is basically what I used (with GCC 8.4.1 in my case).

alalazo (approved):

PR #33486, opened 2022-10-24
ParaView: add latest release v5.11.0-RC1

vicentebolea:
:tada: :tada: :tada:

ParaView 5.11.0-RC2 is out

This PR updates the ParaView Spack package to include this new release (and remove the RC1) while keeping its last final release ParaView v5.10.1 as the preferred version.

Let me know if we need to change anything else in the package.

:tada: :tada: :tada:

PR #33653, opened 2022-11-01
Paraview hip variant

mathstuf:


var/spack/repos/builtin/packages/paraview/package.py
```diff
@@ -213,6 +219,7 @@ class Paraview(CMakePackage, CudaPackage):
     depends_on("xz")
     depends_on("zlib")
     depends_on("libcatalyst", when="+libcatalyst")
+    depends_on("hip", when="+hip")
```
kwryankrattiger: I think this will require a minimum version, ref VTK-m

var/spack/repos/builtin/packages/paraview/package.py
```diff
@@ -27,6 +27,9 @@ class Paraview(CMakePackage, CudaPackage):
     tags = ["e4s"]
 
     version("master", branch="master", submodules=True)
+    version(
```
kwryankrattiger: Since the old RCs are not going to be released/have LTS, I think it is safe to replace the RC1 with RC2 here.

var/spack/repos/builtin/packages/paraview/package.py
```diff
@@ -73,6 +76,7 @@ class Paraview(CMakePackage, CudaPackage):
     variant("pagosa", default=False, description="Build the pagosa adaptor")
     variant("eyedomelighting", default=False, description="Enable Eye Dome Lighting feature")
     variant("adios2", default=False, description="Enable ADIOS2 support", when="@5.8:")
+    variant("hip", default=False, description="Enable HIP support", when="@5.11.0-RC2:")
```
kwryankrattiger: `when=5.11:`

var/spack/repos/builtin/packages/paraview/package.py
vicentebolea: This will need rebase shortly, this line will be updated shortly: https://github.com/spack/spack/pull/33486

var/spack/repos/builtin/packages/paraview/package.py
```diff
@@ -591,6 +598,9 @@ def nvariant_bool(feature):
         if "+advanced_debug" in spec:
             cmake_args.append("-DVTK_DEBUG_LEAKS:BOOL=ON")
 
+        if self.spec.satisfies("@5.11.0-RC2:"):
+            cmake_args.append("-DPARAVIEW_USE_HIP:BOOL=%s" % variant_bool("+hip"))
+
```
vicentebolea, line 595: We need to setup the following vars

``
    Kokkos_CXX_COMPILER:                                                                                                                                                            
    CMAKE_HIP_ARCHITECTURES:
```

File: var/spack/repos/builtin/packages/paraview/package.py (hunk 1 above)
Comment (mathstuf) [reply var/spack/repos/builtin/packages/paraview/package.py]: VTK/ParaView don't support external VTK-m, so how would it ever be out-of-sync?

File: var/spack/repos/builtin/packages/paraview/package.py (hunk 4 above)
Comment (mathstuf) [reply var/spack/repos/builtin/packages/paraview/package.py:595]: Kokkos is not discovered by default? And what do you suggest for the arches? Something like CUDA?

Comment (vicentebolea) [reply var/spack/repos/builtin/packages/paraview/package.py:595]: `Kokkos_CXX_COMPILER` needs to be specified in Kokkos when being transitively used through more than one cmake nested `projects()`. 
CMAKE_HIP_ARCHITECTURES is needed since the compiler default choice sometimes do not execute in target machines like in the ORNL supercomputers. We can delegate this by making this package a subclass of HipPackage (@kwryankrattiger has a mr attempting this)

Comment (mathstuf) [reply var/spack/repos/builtin/packages/paraview/package.py:595]: I assume you mean `ROCmPackage` based on other usages. I think I got this right…

Comment (vicentebolea) [reply var/spack/repos/builtin/packages/paraview/package.py:595]: For some reason we need `Kokkos_CXX_COMPILER` to be set to the hipcc in order to build this.

Comment (mathstuf) [reply var/spack/repos/builtin/packages/paraview/package.py:595]: I added a direct dep.

Comment (vicentebolea) [reply var/spack/repos/builtin/packages/paraview/package.py:595]: I think that it will not resolve the issue, the problem is at cmake level rather than at spack level. You can try to to perform the build of this MR, perhaps it works.

Comment (mathstuf) [reply var/spack/repos/builtin/packages/paraview/package.py:595]: Ok, I have a fix based on #33772, but it doesn't do all the fancy logic there. We really need the `hip` package to have an easy query to answer "where is your `clang`?" instead of duplicating everything in the method edited in the linked PR.

**Comment by mathstuf** [comment]
I've tested this on `crusher` (I used a different way to get `Kokkos_CXX_COMPILER`, but am testing this diff…if `spack` would actually get to the build bit).

**Comment by vicentebolea** [comment]
> I've tested this on `crusher` (I used a different way to get `Kokkos_CXX_COMPILER`, but am testing this diff…if `spack` would actually get to the build bit).

Thanks! I will try to replicate this in other HIP enable machines

File: var/spack/repos/builtin/packages/paraview/package.py [hunk 5]
```diff
@@ -111,6 +111,8 @@ class Paraview(CMakePackage, CudaPackage):
     conflicts("+python3", when="@:5.5")
     conflicts("+shared", when="+cuda")
     conflicts("+cuda", when="@5.8:5.10")
+    conflicts("+rocm", when="+cuda")
```
vicentebolea, line 109: why not hip as opposed to rocm. Note that CUDA is to CUDA TK what HIp is to ROCM.

vicentebolea: Nevermind, I can see that this variant is provided by the ROCmPackage

vicentebolea (commented):
We need to enable `use_vtkm` when `+rocm`

kwryankrattiger:
> We need to enable `use_vtkm` when `+rocm`

Really, it needs to conflict with `use_vtkm=off` since `default` and `on` should both be about to handle this correctly.

mathstuf:
Now conflicts with no-vtkm.

var/spack/repos/builtin/packages/paraview/package.py
kwryankrattiger: it requires a minimum HIP version, `depends_on("hip@3.7:", when="+rocm")`

kwryankrattiger: although, I would say make it a minimum of 5.2, I know we have had issues with earlier versions of HIP and 5.2 is the only version the ParaView has actually been tested against.

vicentebolea: > 4.5 < 5.1 is doable with -O0 optimization levels, anything beyond -O0 will hang template instantiations due to a known bug in hipcc

mathstuf: Minimum of 5.2.0 enforced.

vicentebolea:
@mathstuf 

I was able to build this PR with the following changes:


```
diff --git a/var/spack/repos/builtin/packages/paraview/package.py b/var/spack/repos/builtin/packages/paraview/package.py
index 0a2195260d..852960fcd7 100644
--- a/var/spack/repos/builtin/packages/paraview/package.py
+++ b/var/spack/repos/builtin/packages/paraview/package.py
@@ -204,7 +204,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("llvm-amdgpu", when="+rocm")
     for target in ROCmPackage.amdgpu_targets:
         depends_on(
-            "kokkos amdgpu_target={0}".format(target),
+            "kokkos +rocm amdgpu_target={0}".format(target),
             when="+rocm amdgpu_target={0}".format(target),
         )
 
@@ -591,7 +591,7 @@ def nvariant_bool(feature):
             if "+rocm" in spec:
                 cmake_args.append(
                     "-DKokkos_CXX_COMPILER=%s"
-                    % join_path(spec["llvm-amdgpu"].prefix.bin, "clang++")
+                    % join_path(spec["hip"].prefix.bin, "amdclang++")
                 )
 
         if "+catalyst" in spec:
```

I also had to rebase to the latest "Develop" as in the current stage certain dependencies would not build.

kwryankrattiger (approved):

kwryankrattiger:
This feature is now merged in #34790

PR #33657, opened 2022-11-02
demote warning for no source id to debug message

becker33:
We've talked for ages about making this change -- this warning doesn't mean much to users, and only signifies that we have slightly less reproducibility for the current build. It belongs as a debug message, not normal output.

### DIFF
--- a/lib/spack/spack/test/conftest.py
+++ b/lib/spack/spack/test/conftest.py
@@ -591,6 +591,16 @@ def linux_os():
     return LinuxOS(name=name, version=version)
 
 
+@pytest.fixture(scope="function")
+def debug_output():
+    saved = tty.debug_level()
+    tty.set_debug(2)
+
+    yield
+
+    tty.set_debug(saved)
+
+
 @pytest.fixture(autouse=is_windows, scope="session")
 def platform_config():
     spack.config.add_default_platform_scope(spack.platforms.real_host().name)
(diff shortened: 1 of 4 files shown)


becker33:
@spackbot fix style

lib/spack/spack/test/conftest.py
```diff
@@ -591,6 +591,16 @@ def linux_os():
     return LinuxOS(name=name, version=version)
 
 
+@pytest.fixture(scope="function")
+def debug_output():
+    saved = tty.debug_level()
+    tty.set_debug(2)
```
tldahlgren: Why debug level 2?  The change you made only needs level 1.

tldahlgren: I ask because there is a lot more output at level 2 that has to be parsed.

tldahlgren: FWIW.  There's already a fixture -- `ensure_debug` -- in `test_suite.py` that could be "promoted" to `conftest.py` that sets the level to 1.

tldahlgren (commented):
LG and works for me.  Just wondering about using debug level 2 versus 1.

becker33:
@spackbot fix style

becker33:
@tldahlgren thanks for the review, I hadn't known about that fixture. Refactored to use it.

alalazo (dismissed):

tldahlgren (dismissed):
LGTM

alalazo (approved):

tldahlgren (approved):

PR #33658, opened 2022-11-02
tests: fix group membership check in sbang tests.

tgamblin:
Fixes `develop`, which is broken by this test.

Fixes an issue on the RHEL8 UBI container where this test would fail because `gr_mem` was empty for every entry in the `grp` DB.

You have to check *both* the `pwd` database (which has primary groups) and `grp` (which has other groups) to do this correctly.

- [x] update `llnl.util.filesystem.group_ids()` to do this
- [x] use it in the `sbang` test

### DIFF
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -505,8 +505,15 @@ def group_ids(uid=None):
 
     if uid is None:
         uid = getuid()
-    user = pwd.getpwuid(uid).pw_name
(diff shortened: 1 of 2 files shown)


alalazo (approved):

tgamblin:
@trws can you rebase this on `develop`?

alalazo:
I'll force merge this hotfix, since it doesn't change anything related to pipelines[^1]

[^1]: CI is currently rebuilding `paraview`, then it has to rebuild the index 2 times and that will take a long amount of time before a needless pipelines starts over this PR.

PR #23854, opened 2021-05-21
refactor conduit package's spack post install testing logic [wip]

cyrush:

PR #33667, opened 2022-11-02
use spack build_directory var to determine build dir for tests

cyrush:


cyrush:
closing in lieu of existing PR #23854

PR #33599, opened 2022-10-28
flux-core: allow ncurses >= 6.2

eugeneswalker:
At least `flux-core@0.44.0` seems to build just fine w/ `^ncurses@6.3`

What do you think @grondo?

FYI @wspear

PR #33669, opened 2022-11-02
gitlab: Prune untouched specs less aggressively

scottwittenburg:
Untouched spec pruning was added to reduce the number of specs developers see getting rebuilt in their PR pipelines that they don't understand.  Because the state of the develop mirror lags quite far behind the tip of the develop branch, PRs often find they need to rebuild things untouched by their PR.

Untouched spec pruning was previously implemented by finding all specs in the environment with names of packages touched by the PR, traversing in both directions the DAGS of those specs, and adding all dependencies as well as dependents to a list of concrete specs that should not be considered for pruning.

We found that this heuristic results in too many pruned specs, and that dependents of touched specs must have all their dependencies added to the list of specs that should not be considered for pruning.

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -515,38 +515,36 @@ def compute_affected_packages(rev1="HEAD^", rev2="HEAD"):
     return spack.repo.get_all_package_diffs("ARC", rev1=rev1, rev2=rev2)
 
 
-def get_spec_filter_list(env, affected_pkgs, dependencies=True, dependents=True):
-    """Given a list of package names, and assuming an active and
-       concretized environment, return a set of concrete specs from
-       the environment corresponding to any of the affected pkgs (or
-       optionally to any of their dependencies/dependents).
+def get_spec_filter_list(env, affected_pkgs):
+    """Given a list of package names and an active/concretized
+       environment, return the set of all concrete specs from the
+       environment that could have been affected by changing the
+       list of packages.
 
     Arguments:
 
         env (spack.environment.Environment): Active concrete environment
         affected_pkgs (List[str]): Affected package names
-        dependencies (bool): Include dependencies of affected packages
-        dependents (bool): Include dependents of affected pacakges
 
     Returns:
 
-        A list of concrete specs from the active environment including
-        those associated with affected packages, and possible their
-        dependencies and dependents as well.
+        A set of concrete specs from the active environment including
+        those associated with affected packages, their dependencies and
+        dependents, as well as their dependents dependencies.
     """
     affected_specs = set()
     all_concrete_specs = env.all_specs()
     tty.debug("All concrete environment specs:")
     for s in all_concrete_specs:
         tty.debug("  {0}/{1}".format(s.name, s.dag_hash()[:7]))
-    for pkg in affected_pkgs:
-        env_matches = [s for s in all_concrete_specs if s.name == pkg]
-        for match in env_matches:
-            affected_specs.add(match)
-            if dependencies:
-                affected_specs.update(match.traverse(direction="children", root=False))
-            if dependents:
-                affected_specs.update(match.traverse(direction="parents", root=False))
+    env_matches = [s for s in all_concrete_specs if s.name in frozenset(affected_pkgs)]
+    visited = set()
+    dag_hash = lambda s: s.dag_hash()
+    for match in env_matches:
+        for parent in match.traverse(direction="parents", key=dag_hash):
+            affected_specs.update(
+                parent.traverse(direction="children", visited=visited, key=dag_hash)
+            )
     return affected_specs
 
 
@@ -625,7 +623,7 @@ def generate_gitlab_ci_yaml(
                 affected_specs = get_spec_filter_list(env, affected_pkgs)
                 tty.debug("all affected specs:")
                 for s in affected_specs:
-                    tty.debug("  {0}".format(s.name))
+                    tty.debug("  {0}/{1}".format(s.name, s.dag_hash()[:7]))
 
     # Allow overriding --prune-dag cli opt with environment variable
     prune_dag_override = os.environ.get("SPACK_PRUNE_UP_TO_DATE", None)
@@ -848,7 +846,11 @@ def generate_gitlab_ci_yaml(
 
                 if prune_untouched_packages:
                     if release_spec not in affected_specs:
-                        tty.debug("Pruning {0}, untouched by change.".format(release_spec.name))
+                        tty.debug(
+                            "Pruning {0}/{1}, untouched by change.".format(
+                                release_spec.name, release_spec.dag_hash()[:7]
+                            )
+                        )
                         spec_record["needs_rebuild"] = False
                         continue
 
--- a/lib/spack/spack/test/ci.py
+++ b/lib/spack/spack/test/ci.py
@@ -442,13 +442,16 @@ def test_get_spec_filter_list(mutable_mock_env_path, config, mutable_mock_repo):
     touched = ["libdwarf"]
 
     # traversing both directions from libdwarf in the graphs depicted
-    # above results in the following possibly affected env specs:
-    # mpileaks, callpath, dyninst, libdwarf, and libelf.  Unaffected
-    # specs are mpich, plus hypre and it's dependencies.
+    # above (and additionally including dependencies of dependents of
+    # libdwarf) results in the following possibly affected env specs:
+    # mpileaks, callpath, dyninst, libdwarf, libelf, and mpich.
+    # Unaffected specs are hypre and it's dependencies.
 
     affected_specs = ci.get_spec_filter_list(e1, touched)
     affected_pkg_names = set([s.name for s in affected_specs])
-    expected_affected_pkg_names = set(["mpileaks", "callpath", "dyninst", "libdwarf", "libelf"])
+    expected_affected_pkg_names = set(
+        ["mpileaks", "mpich", "callpath", "dyninst", "libdwarf", "libelf"]
+    )
 
     assert affected_pkg_names == expected_affected_pkg_names
 
--- a/lib/spack/spack/test/cmd/ci.py
+++ b/lib/spack/spack/test/cmd/ci.py
@@ -1837,8 +1837,8 @@ spack:
             yaml_contents = syaml.load(contents)
 
             for ci_key in yaml_contents.keys():
-                if "archive-files" in ci_key or "mpich" in ci_key:
-                    print("Error: archive-files and mpich should have been pruned")
+                if "archive-files" in ci_key:
+                    print("Error: archive-files should have been pruned")
                     assert False
 
 


scottwittenburg:
@eugeneswalker I tested this locally with your change from #33599 and it generated jobs for (among other specs) the things that were missing from the binary cache and causing failures like [here](https://gitlab.spack.io/spack/spack/-/jobs/4122956).

eugeneswalker:
> @eugeneswalker I tested this locally with your change from #33599 and it generated jobs for (among other specs) the things that were missing from the binary cache and causing failures like [here](https://gitlab.spack.io/spack/spack/-/jobs/4122956).

Sweet. I just cherry-picked your commit here onto this PR branch:
* https://github.com/spack/spack/pull/33599

Which is the one I've been having a lot of trouble getting to build. Let's see if the pipeline finishes now!

eugeneswalker:
Style check failing, not sure why...

scottwittenburg:
@spackbot fix style

lib/spack/spack/ci.py
```diff
@@ -546,7 +546,8 @@ def get_spec_filter_list(env, affected_pkgs, dependencies=True, dependents=True)
             if dependencies:
                 affected_specs.update(match.traverse(direction="children", root=False))
             if dependents:
-                affected_specs.update(match.traverse(direction="parents", root=False))
+                for dependent_spec in match.traverse(direction="parents", root=False):
+                    affected_specs.update(dependent_spec.traverse(direction="children", root=True))
```
haampie: why traverse down in every intermediate parent? note that with this change `dependents` implies `dependencies`, and you should either use a visited set to avoid traversing the same edge you came from going up, or find all the leaf nodes in the parents direction, and take all children recursively from them.

scottwittenburg: Because if a dependent of a touched specs turns out to be missing from the mirror (see the description), we'll schedule it for rebuilding.  If that dependent has any dependencies missing from the mirror too, we should schedule those as well, or we'll end up trying to build that dependent and we'll find the dependencies couldn't be installed from binary mirror when `--cache-only` was specified.

haampie: If you do the latter, you can use the new api `spack.traverse.traverse_nodes([root_a, root_b, ...])`

scottwittenburg: > either use a visited set to avoid traversing the same edge you came from going up

Even though `affected_specs` is a `set`?

haampie: Think of a dag that looks like `v1 <- v2 <- v3 <- ... <- vn` where `<-` points to the dependency. If you start with `v1` your traversal code will first iterate over `v1`, then `v2, v1`, then ... `v3, v2, v1`, ... `vn, ..., v1` which is quadratic and unnecessary. If you pass `traverse(visited=my_set)` it won't follow edges to nodes seen before.

haampie: Also, I would just drop the keyword args `dependencies=True` and `dependents=True`, they're never set to anything else so there's no point to make it configurable, especially now that with this change `dependents` doesn't mean `dependents` and in fact implies `dependencies`.

scottwittenburg: Oh, yes, now I get what you mean by "every intermediate parent".  Thanks for the suggestion.

haampie: You may need to be a bit careful, cause the visited kwarg to `traverse` takes elements of whatever type the `key` function produces (`id` by default) to distinguish unique specs. It's not a set of Spec instances.

scottwittenburg: But I should be fine if provide `key=lambda s: s.dag_hash()` to `traverse`, as you suggested last week on slack?

haampie: It will still be a set of strings (dag hashes), not specs:

```python
In [1]: cmake = spack.spec.Spec('cmake').concretized() # cmake

In [2]: child = cmake.dependencies()[0] # ncurses

In [3]: visited = set()

In [4]: [s.name for s in child.traverse(visited=visited, key=lambda s: s.dag_hash())]
Out[4]: ['ncurses', 'pkgconf']

In [5]: visited
Out[5]: {'e5ewulkrwwvdck3b3dxsepgiujr7iqbz', 'xzo47hkvc77pabwn4kyyiwkjo3yud7xd'}

# notice no ncurses / pkgconf here
In [6]: [s.name for s in cmake.traverse(visited=visited, key=lambda s: s.dag_hash())]
Out[6]: ['cmake', 'openssl', 'ca-certificates-mozilla', 'perl', 'berkeley-db', 'bzip2', 'diffutils', 'libiconv', 'gdbm', 'readline', 'zlib']

In [7]: visited
Out[7]: {'5fie3rqgdtbik6qopaa5plhgfinuuywz', 'atlyybqxtsfjnci7xosrv4cczm7pjeh4', 'e5ewulkrwwvdck3b3dxsepgiujr7iqbz', 'fdvtzu2nvuktthv2eygftlywmxxmlq5p', 'gtwazdob6rf6mlcojjoartzucqx4r35x', 'hlnjdqnhzbz4sdg4b2nx3goh44iwb2qs', 'i55tsxsepukei4klbtquplznt5oxzyvy', 'm6gfpiqxuhtlryro5qt36sbp2xie5o7s', 'q27oxeoxa22tkhck7pbv2f6azjydzkia', 'vc4kuphzwsemw23up4o2kwx5tja5gvmq', 'w2gmdr7nimmcm6l25gbf3fe7mhqdathb', 'xzo47hkvc77pabwn4kyyiwkjo3yud7xd', 'y4zrxrek36joqs5hrg3hgwoko3esi5mw'}
```

but it's better to key by `dag_hash` in case you somehow end up with two `Spec` instances describing the exact same spec with the same hash (due to a copy or loading them from disk twice or so)

scottwittenburg: I get it, thank you for providing the example.  I was confused at first because `In [4]` looks exactly the same as `In  [6]`, but I think it must have read `child.traverse(...)` when you evaluated it originally.

haampie: Yeah, i changed it a few times 😆

eugeneswalker:
I incorporated the changes here into:
* https://github.com/spack/spack/pull/33599

Seeing these failures:
* [(specs) flux-sched/pxjdqtn 0.25.0 oneapi@2022.1.0 linux-ubuntu20.04-x86_64 E4S OneAPI](https://gitlab.spack.io/spack/spack/-/jobs/4134724)
```
...
==> Installing flux-sched-0.25.0-pxjdqtnl3tlpc5b2hexupiwkdqixamkb
==> Error: TypeError: traverse_nodes() got an unexpected keyword argument 'deptypes'
/builds/spack/spack/var/spack/repos/builtin/packages/lua/package.py:118, in _setup_dependent_env_helper:
        116    def _setup_dependent_env_helper(self, env, dependent_spec):
        117        lua_paths = []
  >>    118        for d in dependent_spec.traverse(deptypes=("build", "run"), deptype_query="run"):
        119            if d.package.extends(self.spec):
        120                lua_paths.append(os.path.join(d.prefix, self.lua_lib_dir))
        121                lua_paths.append(os.path.join(d.prefix, self.lua_lib64_dir))
  File "/builds/spack/spack/lib/spack/spack/build_environment.py", line 1056, in _setup_pkg_and_run
    kwargs["env_modifications"] = setup_package(
  File "/builds/spack/spack/lib/spack/spack/build_environment.py", line 783, in setup_package
    env_mods.extend(modifications_from_dependencies(pkg.spec, context, custom_mods_only=False))
  File "/builds/spack/spack/lib/spack/spack/build_environment.py", line 1006, in modifications_from_dependencies
    add_modifications_for_dep(dspec)
  File "/builds/spack/spack/lib/spack/spack/build_environment.py", line 989, in add_modifications_for_dep
    builder.setup_dependent_build_environment(env, spec)
  File "/builds/spack/spack/lib/spack/spack/builder.py", line 191, in setup_dependent_build_environment
    return self.pkg_with_dispatcher.setup_dependent_build_environment(env, dependent_spec)
  File "/builds/spack/spack/var/spack/repos/builtin/packages/lua/package.py", line 142, in setup_dependent_build_environment
  File "/builds/spack/spack/var/spack/repos/builtin/packages/lua/package.py", line 119, in _setup_dependent_env_helper
    if d.package.extends(self.spec):
  File "/builds/spack/spack/lib/spack/spack/spec.py", line 1694, in traverse
    return traverse.traverse_nodes([self], **kwargs)
TypeError: traverse_nodes() got an unexpected keyword argument 'deptypes'
==> Error: flux-sched-0.25.0-pxjdqtnl3tlpc5b2hexupiwkdqixamkb: Package was not installed
==> Error: Installation request failed.  Refer to reported errors for failing package(s).
```

haampie:
Fixed on develop

lib/spack/spack/ci.py
```diff
@@ -542,11 +540,21 @@ def get_spec_filter_list(env, affected_pkgs, dependencies=True, dependents=True)
...
-                affected_specs.update(match.traverse(direction="children", root=False))
-            if dependents:
-                affected_specs.update(match.traverse(direction="parents", root=False))
+            visited = set()
+            affected_specs.update(
+                match.traverse(
+                    direction="children", root=True, visited=visited, key=lambda s: s.dag_hash()
+                )
+            )
+            for dependent_spec in match.traverse(direction="parents", root=False):
```
haampie:
```suggestion
            for dependent_spec in match.traverse(direction="parents", root=True):
```

Not only is this more clear, you're also simply missing the environment root.

haampie (changes requested):

lib/spack/spack/ci.py
```diff
@@ -515,24 +515,22 @@ def compute_affected_packages(rev1="HEAD^", rev2="HEAD"):
...
 
     Returns:
 
         A list of concrete specs from the active environment including
-        those associated with affected packages, and possible their
-        dependencies and dependents as well.
+        those associated with affected packages, their dependencies and
+        dependents, as well as their dependents dependencies.
     """
     affected_specs = set()
```
haampie, line 535: this can be a list now (actually that's what the docstring says too)

eugeneswalker:
> ... I just cherry-picked your commit here onto this PR branch:
> 
>     * [flux-core: allow ncurses >= 6.2 #33599](https://github.com/spack/spack/pull/33599)

My GitLab pipelines are passing now! Thanks! Just have these other Unit Test failures, and then we're in business!

lib/spack/spack/ci.py
scottwittenburg: In the traverse of the matched spec in the children direction, above, I included the root.  So I figured it was fine to skip it here.  But maybe I'm missing something else.

lib/spack/spack/ci.py
scottwittenburg: Or maybe it could be kept a `set` since we have to check membership of each spec we might otherwise schedule for rebuild?

haampie: I guess that could be done too, I didn't check what the call site looks like. It's just that set.update(...) is more expensive than list.extend(...), and idk how expensive the equality check is (does it have a fast path for concrete specs based on dag hash? or maybe a fast path for `id(x) == id(y)`?

lib/spack/spack/ci.py
haampie: Ah, sorry, I missed that second `root=True`. Can you still simplify to

```python
visited = set()
dag_hash = lambda s: s.dag_hash()
for match in env_matches:
    for parent in match.traverse(direction="parents", key=dag_hash):
        affected_specs.update(parent.traverse(direction="children", visited=visited, key=dag_hash)
```

it's much more succinct

lib/spack/spack/ci.py
```diff
@@ -542,11 +540,21 @@ def get_spec_filter_list(env, affected_pkgs, dependencies=True, dependents=True)
     for pkg in affected_pkgs:
         env_matches = [s for s in all_concrete_specs if s.name == pkg]
         for match in env_matches:
-            affected_specs.add(match)
-            if dependencies:
-                affected_specs.update(match.traverse(direction="children", root=False))
-            if dependents:
-                affected_specs.update(match.traverse(direction="parents", root=False))
+            visited = set()
```
haampie: this can go out of the loop, right?

lib/spack/spack/ci.py
haampie: and in fact you can drop the outer loop with `env_matches = [s for s in all_concrete_specs if s.name in affected_pkgs]`, where `affected_pkgs` could be also be replaced with `(frozen)set(...)`

lib/spack/spack/ci.py
scottwittenburg: At the call site it's checking `if <spec> not in <set>`, and if not, it prunes the job.

haampie (approved):

PR #33670, opened 2022-11-02
remove `deptype_query` remnants and fix incorrect `deptypes` kwarg 

haampie:
This is not a thing since 2017, it's just that it errors now because
`traverse` doesn't ignore unused `**kwargs` since #33406.

(Also, probably `traverse` is wrong here, `dependencies` was intended?
Why look for build deps of build deps of ... maybe @glennpj  knows.)

### DIFF
--- a/lib/spack/spack/build_systems/lua.py
+++ b/lib/spack/spack/build_systems/lua.py
@@ -76,7 +76,7 @@ class LuaBuilder(spack.builder.Builder):
     def generate_luarocks_config(self, pkg, spec, prefix):
         spec = self.pkg.spec
         table_entries = []
-        for d in spec.traverse(deptypes=("build", "run")):
+        for d in spec.traverse(deptype=("build", "run")):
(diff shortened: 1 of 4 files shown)


tgamblin (approved):

glennpj:
> (Also, probably `traverse` is wrong here, `dependencies` was intended?
> Why look for build deps of build deps of ... maybe @glennpj knows.)

I can only speak to R packages here. Part of the R package build process is to load the newly built package into an R session. That requires loading the direct dependencies, and the dependencies of those, and so on. Since each r- package is in a unique directory, it is necessary to construct the entire RLIBS path in order to be able to load all of the packages in the R session. From the perspective of R, there is no distinction between build and run dependencies. The convention in Spack has been to set r- package dependencies as `type=("build", "run")`.

haampie:
To me it sounds more natural if you'd take run-type deps recursively at ... runtime, and for the build env take direct build deps + their recursive run deps

PR #33621, opened 2022-10-31
E4S: Conservatively add ecp-data-vis-sdk

kwryankrattiger:
More conservative effort than #32764 but enough to get the SDK building in the e4s environment.

PR #33676, opened 2022-11-02
ECP-SDK: fixup +hdf5 +cuda contraints

svenevs:
Only enable the hdf5-vfd-gds package if it can compile.

- hdf5-vfd-gds needs cuda@11.7.1+ to be able to `find_library` for cuFile.
- Only enable hdf5-vfd-gds in the sdk if cuda@11.7.1+ is available. If an earlier version of cuda is being used, do not depend on the hdf5-vfd-gds package at all.

Identified by: #33621, follow-up to: #33300

FYI @hyoklee @brtnfld we confirmed that `hdf5-vfd-gds@1.0.2` can build with `cuda@11.7.1` but not with `cuda@11.7.0` so we're just stamping that into the package.


hyoklee (dismissed):

kwryankrattiger:
@spackbot run pipeline

kwryankrattiger (approved):

hyoklee:
Thank you, @kwryankrattiger !

PR #33677, opened 2022-11-02
use spack.traverse.traverse_nodes where useful

haampie:
We now have `for s in spack.traverse.traverse_nodes([s1, s2, ...])` as a replacement for `for x in [s1, s2, ...]; for s in x.traverse():`, with the small optimization that it automatically maintains a shared `visited` set, so we don't have to deal with duplicates at the call-site anymore. In some places de-duping was likely forgotten, in other places it was done but inefficiently -- this PR makes it clean and efficient.

### DIFF
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -34,6 +34,7 @@ import spack.spec
 import spack.stage
 import spack.store
 import spack.subprocess_context
+import spack.traverse
 import spack.user_environment as uenv
 import spack.util.cpus
 import spack.util.environment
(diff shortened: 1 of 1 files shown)


lib/spack/spack/environment/environment.py
```diff
@@ -1760,22 +1761,14 @@ def install_specs(self, specs=None, **install_args):
...
-                spec = self.specs_by_hash[h]
-            except KeyError:
-                tty.warn(
-                    "Environment %s appears to be corrupt: missing spec " '"%s"' % (self.name, h)
-                )
-                continue
-            all_specs.update(spec.traverse())
-
-        return sorted(all_specs)
+        roots = [self.specs_by_hash[h] for h in self.concretized_order]
```
haampie, line 1764: notice that we don't have this defensive validation of `KeyError` in tons of other places, so I dropped it here too.

lib/spack/spack/environment/environment.py
```diff
@@ -1760,22 +1761,14 @@ def install_specs(self, specs=None, **install_args):
...
+        specs = [s for s in spack.traverse.traverse_nodes(roots, lambda s: s.dag_hash())]
+        specs.sort()
+        return specs
 
     def all_hashes(self):
         """Return hashes of all specs."""
-        return list(set(s.dag_hash() for s in self.all_specs()))
+        return [s.dag_hash() for s in self.all_specs()]
```
haampie, line 1771: We now know for a fact there's no need to list(sort(...))

alalazo (approved):

PR #33678, opened 2022-11-02
e4s ci: more RAM for rocblas

zackgalbreath:


haampie:
You can also just try to cut down the number of jobs, the logs show it was using `nproc -a` jobs.

Actually is the k8s request a lowerbound? The jobs shows `make -j16` (is it capped by spack `config:build_jobs`?), the tensile script that rocblas invokes ran with 92 jobs or so, the k8s config was 11000 milli-CPUs?

zackgalbreath:
> Actually is the k8s request a lowerbound?

Pretty much, yeah. Here's a couple key quotes from the [relevant docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/):

"When you specify the resource request for containers in a Pod, the kube-scheduler uses this information to decide which node to place the Pod on."

"If the node where a Pod is running has enough of a resource available, it's possible (and allowed) for a container to use more resource than its request for that resource specifies."

> The jobs shows `make -j16` (is it capped by spack `config:build_jobs`?)

Yes, that is my understanding.

> the tensile script that rocblas invokes ran with 92 jobs or so, the k8s config was 11000 milli-CPUs?

FWIW this does seem to be a known issue. From the [rocblas docs](https://rocblas.readthedocs.io/en/rocm-5.3.1/Linux_Install_Guide.html#requirements
):

"As a general rule, 64GB of system memory is required for a full rocBLAS build"

haampie:
rocblas (or amd's llvm fork) is a joke. still, memory usage will go down if you reduce the number of processes in tensile.

zackgalbreath:
> You can also just try to cut down the number of jobs,

Good point. Let's see how that goes in #33692

PR #30669, opened 2022-05-14
errors: model error messages as an optimization problem

becker33:
Error messages for the clingo concretizer have proven challenging. The current messages are incredibly vague and often don't help users at all. Unsat cores in clingo are not guaranteed to be minimal, and lead to cores that are either not useful or need to be post-processed for hours to reach a minimal core.

Following up on an idea from a slack conversation with @kwryankrattiger on slack, this PR takes a new approach. We eliminate most integrity constraints and minima/maxima on choice rules in clingo, and instead force invalid states to imply certain error facts. The error facts can include context on the cause of the error (Package, Version, etc). These error facts are then heavily optimized against, to ensure that we do not include error facts in the solution when a solution with no error facts could be generated. When post-processing the clingo solution to construct specs, any error facts cause the program to raise an exception. This leads to much more legible error messages.

In terms of performance, "fresh" concretizations appear to suffer a ~25% performance penalty under this branch, while "reuse" concretizations see a substantial speedup (around 33%). Here is averaged performance data for `hdf5` and `trilinos` with develop and this branch, each with `--reuse` and `--fresh`.

<details><summary>hdf5</summary>

```
develop reuse
  setup 8.3892925
  load 0.027857500000000007
  ground 1.6178600000000003
  solve 24.494772499999993
 Total 34.5496125
develop fresh
  setup 7.7822024999999995
  load 0.028942500000000003
  ground 1.45309
  solve 1.963295
 Total 11.4762175
my_branch reuse
  setup 9.066266666666667
  load 0.028113333333333334
  ground 2.042183333333333
  solve 12.059683333333334
 Total 23.216083333333337
my_branch fresh
  setup 8.39289
  load 0.02755666666666667
  ground 1.877646666666667
  solve 3.5063833333333334
 Total 13.998306666666663
```
</details>

<details><summary>trilinos</summary>

```
develop reuse
  setup 8.980122
  load 0.027924
  ground 1.7392540000000003
  solve 20.526521999999996
 Total 31.320413999999992
develop fresh
  setup 8.400396
  load 0.028590000000000004
  ground 1.573622
  solve 2.142032
 Total 12.409289999999997
my_branch reuse
  setup 9.870964999999996
  load 0.02883
  ground 2.1984524999999993
  solve 10.775852500000001
 Total 22.9245325
my_branch fresh
  setup 9.109247499999999
  load 0.029365000000000002
  ground 1.9857200000000002
  solve 3.8974500000000005
 Total 15.222154999999997
```
</details>

[...]

PR #33687, opened 2022-11-03
propagation: don't count propagated variant values as non-default

tgamblin:
Builds on #29761 (fixes issue described in https://github.com/spack/spack/pull/29761#issuecomment-1300380955)

We try to avoid non-default variant values in the concretizer, but this doesn't make sense for variants forced to take some non-default value by variant propagation. Counting this as a penalty effectively biases the concretizer for small specs dependency graphs -- something we try very hard to avoid elsewhere because it can lead to very strange decisions.

Example: with the penalty, `spack spec hdf5` will choose the default `openmpi` as its `mpi` provider, but `spack spec hdf5 ~~shared` will choose `mpich` because it has to set fewer non-default variant values because `mpich`'s DAG is smaller. That's not a good reason to prefer a non-default virtual provider.

To fix this, if the user explicitly requests a non-default value to be propagated, there shouldn't be a penalty. Variant values set on the CLI already don't count as default; we just need to extend that to propagated values.

### DIFF
--- a/lib/spack/spack/solver/concretize.lp
+++ b/lib/spack/spack/solver/concretize.lp
@@ -673,6 +673,8 @@ variant_not_default(Package, Variant, Value)
     not variant_default_value(Package, Variant, Value),
     % variants set explicitly on the CLI don't count as non-default
     not variant_set(Package, Variant, Value),
+    % variant values forced by propagation don't count as non-default
+    not variant_propagate(Package, Variant, Value, _),
     % variants set on externals that we could use don't count as non-default
     % this makes spack prefer to use an external over rebuilding with the
     % default configuration


becker33 (approved):

becker33:
@spackbot run pipeline

alalazo:
Compared develop at commit ae99829af4393295953afa0f08394bdcd2b58b75 with commit 9c5c327a6feb47528b01029806477e84073a3a13 and timed on the `--fresh` concretization of packages in the `radiuss` stack:

[radiuss_propagation.csv](https://github.com/spack/spack/files/11080259/radiuss_propagation.csv)
[radiuss_develop.csv](https://github.com/spack/spack/files/11080260/radiuss_develop.csv)
[radiuss.txt](https://github.com/spack/spack/files/11080261/radiuss.txt)

The corresponding plot is:

totals

which shows a non-negligible impact only for `glvis`. The times compared to https://github.com/spack/spack/pull/30669#issuecomment-1485218161 increased in a noticeable way, in particular for the packages that take the most to concretize.

PR #33688, opened 2022-11-03
More jobs in Gitlab CI

haampie:
Updated description:

The PR applies only to Gitlab pipelines (restricted now to x86 / aarch64), where we do CPU claims such as "at least 11000 milli-cpus" (whatever that is) but often get a node with 64 / 128 / 192 cores; in that case Spack caps to -j16, which is... not particularly getting the most out of it. @zackgalbreath showed an example where the LLVM Gitlab job took 3.5 hours (the actual build took 2h), with the above using -j128 it took only 20 minutes (the actual build 10 minutes).

Notice that setting `config:build_jobs` really only sets an _upperbound_, the effective value is determined by the number of _available jobs_ (whatever is allowed to the process; not physical # cores / threads)

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug-aarch64/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug-aarch64/spack.yaml
@@ -246,6 +246,7 @@ spack:
     script:
       - uname -a || true
       - grep -E 'vendor|model name' /proc/cpuinfo 2>/dev/null | sort -u || head -n10 /proc/cpuinfo 2>/dev/null || true
+      - nproc
       - curl -Lfs 'https://github.com/JuliaBinaryWrappers/GNUMake_jll.jl/releases/download/GNUMake-v4.3.0+1/GNUMake.v4.3.0.aarch64-linux-gnu.tar.gz' -o gmake.tar.gz
       - printf '2322c175fb092b426f9eb6c24ee22d94ffa6759c3d0c260b74d81abd8120122b gmake.tar.gz' | sha256sum --check --strict --quiet
       - tar -xzf gmake.tar.gz -C /usr bin/make 2> /dev/null
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug-aarch64/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug-aarch64/spack.yaml
@@ -6,6 +6,7 @@ spack:
     unify: false
 
   config:
+    build_jobs: 256
     concretizer: clingo
     install_tree:
       root: /home/software/spack
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug-aarch64/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug-aarch64/spack.yaml
@@ -6,7 +6,7 @@ spack:
     unify: false
 
   config:
-    build_jobs: 256
+    build_jobs: 32
     concretizer: clingo
     install_tree:
       root: /home/software/spack
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
@@ -243,6 +243,7 @@ spack:
     script:
       - uname -a || true
       - grep -E 'vendor|model name' /proc/cpuinfo 2>/dev/null | sort -u || head -n10 /proc/cpuinfo 2>/dev/null || true
+      - nproc
       - curl -Lfs 'https://github.com/JuliaBinaryWrappers/GNUMake_jll.jl/releases/download/GNUMake-v4.3.0+1/GNUMake.v4.3.0.x86_64-linux-gnu.tar.gz' -o gmake.tar.gz
       - printf 'fef1f59e56d2d11e6d700ba22d3444b6e583c663d6883fd0a4f63ab8bd280f0f gmake.tar.gz' | sha256sum --check --strict --quiet
       - tar -xzf gmake.tar.gz -C /usr bin/make 2> /dev/null
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
@@ -6,6 +6,7 @@ spack:
     unify: false
 
   config:
+    build_jobs: 256
     concretizer: clingo
     install_tree:
       root: /home/software/spack
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
@@ -6,7 +6,7 @@ spack:
     unify: false
 
   config:
-    build_jobs: 256
+    build_jobs: 32
     concretizer: clingo
     install_tree:
       root: /home/software/spack
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-isc-aarch64/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-isc-aarch64/spack.yaml
@@ -6,6 +6,7 @@ spack:
     unify: false
 
   config:
+    build_jobs: 256
     concretizer: clingo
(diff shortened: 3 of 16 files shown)


haampie:
LLVM 14's build stage takes ~10min here, +10min overhead of setting up deps, tarballing and uploading.

tldahlgren:
FWIW. The number of CPUs should depend on the situation.  IME If you're using shared systems (e.g., supercomputers) without an exclusive hold on the node(s), etiquette dictates that you do not take more than half of the available CPUs.  Otherwise, the facilities folks may warn you or kill your jobs.

tldahlgren:
> FWIW. The number of CPUs should depend on the situation. IME If you're using shared systems (e.g., supercomputers) without an exclusive hold on the node(s), etiquette dictates that you do not take more than half of the available CPUs. Otherwise, the facilities folks may warn you or kill your jobs.

IIRC @tgamblin has opinions on this as well. 😀

haampie:
Ah, the description isn't great, mostly cause I wanted to test it in CI and if I was wrong close the PR. But basically the PR applies only to Gitlab pipelines (restricted now to x86 / aarch64), where we do CPU claims such as "at least 11000 milli-cpus" (whatever that is) but often get a node with 64 / 128 / 192 cores; in that case Spack caps to -j16, which is... not particularly getting the most out of it. @zackgalbreath showed an example where the LLVM Gitlab job took 3.5 hours (the actual build took 2h), with the above using -j128 it took only 20 minutes (the actual build 10 minutes)

tldahlgren:
> Ah, the description isn't great, mostly cause I wanted to test it in CI and if I was wrong close the PR. But basically the PR applies only to Gitlab pipelines (restricted now to x86 / aarch64), where we do CPU claims such as "at least 11000 milli-cpus" (whatever that is) but often get a node with 64 / 128 / 192 cores; in that case Spack caps to -j16, which is... not particularly getting the most out of it.

I think that *could* still be problematic for some bare metal CI runners.

haampie:
Yeah, maybe @zackgalbreath can review if I put the `config:build_jobs` in the right spots

eugeneswalker:
For the UO nodes, where we do have 64+ cores on almost every node we offer, we accept concurrent GitLab jobs on those nodes. For instance, we have `192 cores` on `uo-jupiter` -- we allow that node to accept 5 CI jobs from GitLab. So we could not have each of those jobs using `-j192`...

haampie:
Wouldn't it make sense to set the affinity for those jobs then instead of hoping users play nice? Then spack doesnt have to worry about exceptions

eugeneswalker:
> Wouldn't it make sense to set the affinity for those jobs then instead of hoping users play nice? Then spack doesnt have to worry about exceptions

I'd love to do this. How though?

haampie:
Is that bare metal? Kubernetes? Slurm? Non of the above? I think if it's a self-hosted runner and the jobs are spawned as child processes you could consider using slurm or (manually) taskset to set the cpu affinity of say 5 separate runners that all have concurrency 1, instead of 1 runner with concurrency 5. If they're spawned with k8s or so, I guess they have their own ways of capping resources

scottwittenburg:
I would also like to see this worked in, the speedup demonstrated building llvm is obviously compelling.  But I'd be worried about the same kind of cpu oversubscription in the cloud that Luke mentioned on UO hardware.  

If we could make the `build_job` config setting related to the resource requests we specify in each of the matches, that might be nice.

Or alternatively, @kotfic @zackgalbreath, is this where it could make sense to specify cpu resource limits (set to the same value as we currently set the resource requests)?  Would that provide the kind of resource capping @haampie alluded to above?

eugeneswalker:
> Is that bare metal? 

Yes, bare metal.

> I think if it's a self-hosted runner and the jobs are spawned as child processes you could consider using slurm or (manually) taskset to set the cpu affinity of say 5 separate runners that all have concurrency 1, instead of 1 runner with concurrency.

I'm not sure slurm will be feasible for our setup. I should look to see if `gitlab-runner` provides a way to use cgroups or something to constrain what resources are visible to the individual jobs. Do you have experience with `gitlab-runner` configuration? I wonder if someone in our community knows how to do this already that we could talk to about it.

haampie:
I used a custom executor in the past, but it used slurm for scheduling. For running on the same node, to me it sounds easiest to just start the gitlab runner itself with cpu affinity options, e.g.

```taskset -c 0-15 gitlab-runner config-a & taskset -c 16-31 gitlab-runner config-b & taskset -c ...```

But when running multiple runners with concurrency 1 on the same node, you have to be careful that they dont use the same path on the filesystem as workspace (so one config file per runner i think)

@scottwittenburg optimally we should be able to say no cap on build jobs (which was my intention with build_jobs:256), just use as many jobs as threads available to the ci process. I sort of expected that kubernetes would make nproc (ie available cpus) report a smaller number than nproc --all (installed cpus), but instead we dont seem to get any limits? Maybe the scheduler uses average load as an indicator of how many cpus are available or so?

tgamblin:
Not sure if this helps but it looks like we could use podman with the docker executor for standalone runners: https://docs.gitlab.com/runner/executors/docker.html#use-podman-to-run-docker-commands, and then I *believe* you can control cgroups via systemd.

trws:
We don't need systemd or cgroups for this, we could, but it would be a whole lot more work unless someone is already a grand oracle of all things systemd.

If using the docker executor of gitlab-runner is an option, it can use either the more classic cpuset_cpus to specify exactly which cores to run on, or just the cpus option of docker (issue from gitlab-runner [here](https://gitlab.com/gitlab-org/gitlab-runner/-/issues/2101) to constrain the job.  @tgamblin's suggestion would let us use podman the same way, which would be good for anything that might ever run on LC infra.

If we really don't want to use the docker executor and stick with shell, then taskset or numactl like @haampie mentioned are the right answer.  They don't require extra privs, they just works, and they're easy.  The trick is getting it to cover every job, because the gitlab-runner "local" runner doesn't actually, you know, do anything.  A custom executor would certainly do it, but we'd have to set that whole mess up, or we could do something cheeky and set up an ssh config on the system that does it, and use the ssh executor. 😛 

In the end, using the docker one would be the least work, and best supported.  In theory the kubernetes executor would work too, but then we would have to use kube for everything. 🤷

tgamblin:
I don't think kube is the right answer for UO or for other gitlab runners we may add.  It works ok for autoscaling in the cloud, but that's the extent of what I'd do with it -- esp. given that we already have node uarch request issue with kube/karpenter.

haampie:
To me it seems we can safely bump to at least -j32? That already helps.

trws:
On systems with resource limits, we should be able to do `-j $(hwloc-calc —restrict binding -N PU)`, if not all of them have resource limits could we set an env var for number of processes or something? That’s probably simplest if we can’t set the cpuset properly.

eugeneswalker:
> I used a custom executor in the past, but it used slurm for scheduling. For running on the same node, to me it sounds easiest to just start the gitlab runner itself with cpu affinity options, e.g.
> 
> `taskset -c 0-15 gitlab-runner config-a & taskset -c 16-31 gitlab-runner config-b & taskset -c ...`

I will try this very soon. I think this would be a good way to better utilize UO resources.

haampie:
@trws on the Spack side, can/should we improve the `cpus available` logic? Currently it's based on the `sched_getaffinity`, but dunno if there's more to consider to avoid oversubscription.

trws:
I think that actually honors the cpuset by default, so long as that’s true it’s probably fine. If not then yeah we’ll definitely want to fix it.

haampie:
Yeah, it does, just curious if there was more to it. I never looked into the source of `nproc`.

Edit, just read it: seems like `nproc` respects `OMP_NUM_THREADS`, which we won't anyways, otherwise nothing fancy.

haampie:
@eugeneswalker, are you happy with the current state of the PR? Still < 192/5 tasks, and even then, `build_jobs` is an upperbound, so if you would use `taskset -c 0-1 gitlab-runner ...` or so, spack would automatically use 2 tasks not 32.

eugeneswalker:
> @eugeneswalker, are you happy with the current state of the PR? Still < 192/5 tasks, and even then, `build_jobs` is an upperbound, so if you would use `taskset -c 0-1 gitlab-runner ...` or so, spack would automatically use 2 tasks not 32.

Yes, sounds good. Thank you.

eugeneswalker (approved):

haampie:
I'm just gonna merge, if it still leads to over-subscription in some cases, we can revert (partly).

PR #33689, opened 2022-11-03
add justbuild package

asartori86:


var/spack/repos/builtin/packages/justbuild/package.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class Justbuild(Package):
+    "just, a generic build tool"
+
+    git = "https://github.com/just-buildsystem/justbuild.git"
+
+    tags = ["build-tools"]
+
+    executables = ["^just$", "^just-mr$"]
```
tldahlgren: Are both of these always installed? Only one (that is always installed) is needed for https://spack.readthedocs.io/en/latest/packaging_guide.html#making-a-package-discoverable-with-spack-external-find.

tldahlgren: If your goal is to make sure the executables are installed, you could add sanity checks (see https://spack.readthedocs.io/en/latest/packaging_guide.html#adding-sanity-checks).

tldahlgren (commented):
Let me know if you want to make any changes to this package in this PR.

var/spack/repos/builtin/packages/justbuild/package.py
asartori86: thanks @tldahlgren 

I implemented your suggestions. Would you mind to take a second look at it?

both the executables are always installed

tldahlgren (approved):

PR #33692, opened 2022-11-03
Limit the number of parallel jobs launched by Tensile

zackgalbreath:


var/spack/repos/builtin/packages/rocblas/package.py
```diff
@@ -248,6 +248,7 @@ def cmake_args(self):
                 args.append(self.define("Tensile_LIBRARY_FORMAT", "msgpack"))
             if self.spec.satisfies("@:4.2.0"):
                 arch_define_name = "Tensile_ARCHITECTURE"
+            args.append(self.define("Tensile_CPU_THREADS", "16"))
```
haampie: Can you please do `min(16, make_jobs)` and add a comment why this is added, you can pass an int, no need to convert to string.

haampie (approved):

haampie:
Build issues are fixed on develop

PR #33695, opened 2022-11-03
Add zstd support for elfutils

cedricchevalier19:
Not defining `+zstd` implies `--without-zstd` flag to configure.

This avoids automatic library detection and thus make the build only depends on Spack installed dependencies. This was causing some problem on my Ubuntu 22.10 laptop with libzstd globally installed but not found later by `pkg_config`.

### DIFF
--- a/var/spack/repos/builtin/packages/elfutils/package.py
+++ b/var/spack/repos/builtin/packages/elfutils/package.py
@@ -46,6 +46,7 @@ class Elfutils(AutotoolsPackage, SourcewarePackage):
     # Libraries for reading compressed DWARF sections.
     variant("bzip2", default=False, description="Support bzip2 compressed sections.")
     variant("xz", default=False, description="Support xz (lzma) compressed sections.")
+    variant("zstd", default=False, description="Support zstd compressed sections.", when="@0.182:")
 
     # Native language support from libintl.
     variant("nls", default=True, description="Enable Native Language Support.")
@@ -69,6 +70,7 @@ class Elfutils(AutotoolsPackage, SourcewarePackage):
 
     depends_on("bzip2", type="link", when="+bzip2")
     depends_on("xz", type="link", when="+xz")
+    depends_on("zstd", type="link", when="+zstd")
     depends_on("zlib", type="link")
     depends_on("gettext", when="+nls")
     depends_on("m4", type="build")
@@ -113,6 +115,11 @@ class Elfutils(AutotoolsPackage, SourcewarePackage):
             args.append("--with-lzma=%s" % spec["xz"].prefix)
         else:
             args.append("--without-lzma")
+
+        if "+zstd" in spec:
+            args.append("--with-zstd=%s" % spec["zstd"].prefix)
+        else:
+            args.append("--without-zstd")
 
         # zlib is required
         args.append("--with-zlib=%s" % spec["zlib"].prefix)
--- a/var/spack/repos/builtin/packages/elfutils/package.py
+++ b/var/spack/repos/builtin/packages/elfutils/package.py
@@ -106,20 +106,11 @@ class Elfutils(AutotoolsPackage, SourcewarePackage):
         spec = self.spec
         args = []
 
-        if "+bzip2" in spec:
-            args.append("--with-bzlib=%s" % spec["bzip2"].prefix)
-        else:
-            args.append("--without-bzlib")
+        args.extend(self.with_or_without("bzip2", activation_value="prefix"))
+        args.extend(self.with_or_without("xz", activation_value="prefix"))
 
-        if "+xz" in spec:
-            args.append("--with-lzma=%s" % spec["xz"].prefix)
-        else:
-            args.append("--without-lzma")
-
-        if "+zstd" in spec:
-            args.append("--with-zstd=%s" % spec["zstd"].prefix)
-        else:
-            args.append("--without-zstd")
+        with when("zstd"):  # zstd variant is available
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/elfutils/package.py
```diff
@@ -114,6 +116,11 @@ def configure_args(self):
         else:
             args.append("--without-lzma")
 
+        if "+zstd" in spec:
```
tldahlgren: FYI.  I realize this addition is consistent with the style used for other args, but thought it useful to point to one of the AutotoolsPackage helpers for these types of arguments (see https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#activation-value).

tldahlgren (dismissed):
LGTM.

tldahlgren:
@mwkrentel Let me know when you want this PR to be merged.

mwkrentel:
I approve.  I'm Ok with merging as is.  At some point, I'll make a
pass and move everything to the new style.

I knew about enable/disable, but not with/without and activation_value.

mwkrentel (dismissed):

var/spack/repos/builtin/packages/elfutils/package.py
cedricchevalier19: I have updated this part, it makes the code a lot cleaner. Thanks for the information.

var/spack/repos/builtin/packages/elfutils/package.py
```diff
@@ -104,15 +106,11 @@ def configure_args(self):
...
-        else:
-            args.append("--without-bzlib")
+        args.extend(self.with_or_without("bzip2", activation_value="prefix"))
+        args.extend(self.with_or_without("xz", activation_value="prefix"))
 
-        if "+xz" in spec:
-            args.append("--with-lzma=%s" % spec["xz"].prefix)
-        else:
-            args.append("--without-lzma")
+        with when("zstd"):  # zstd variant is available
```
tldahlgren: Interesting use of `with when`.  This is only the second package I've seen  use it for arguments.  Usually it's used  to conditionally add variants, dependencies, and other directives.

I think you're missing the "+" for indicating you want to extend the args when the variant is enabled:

```suggestion
        with when("+zstd"):  # zstd variant is available
```

mwkrentel:
> args.extend(self.with_or_without("xz", activation_value="prefix"))

@cedricchevalier19 I don't see how this can possibly be correct.

The spack package is `xz`, but elfutils calls it `lzma` as in
`--with-lzma=PATH` and I don't see anything in this code that mentions
lzma.

The problem is that it will look like it built, but it won't actually
include xz/lama.  Try building with verbose turned on and I'll bet
there is something in the configure output that says `--with-xz` is an
unknown option.  A warning, but not a fatal error, so it builds, just
not how you think.

Same problem for bzip2/bzlib.

Why don't you go back to your original PR.  It's only a style issue,
but as a friend of mine likes to say, "10% of all sytle 'fixes' break
the build."

mwkrentel (changes requested):
I think the with/without options are broken.

var/spack/repos/builtin/packages/elfutils/package.py
cedricchevalier19: In fact I wanted to only call `with_or_without` when variant `zstd` existed. However, this case is already handled by `with_or_without` according to documentation.

cedricchevalier19:
@mwkrentel I've reverted as suggested. Sorry for the burden. I still use `with_or_without()` for `zstd` as it fits directly.

@tldahlgren I was not able to use the `variant` keyword for `with_or_without` together with the `activation_value`. I ended up with a `KeyError: 'No spec with name bzlib in ...'`

tldahlgren (approved):
LGTM.

tldahlgren:
> @mwkrentel I've reverted as suggested. Sorry for the burden. I still use `with_or_without()` for `zstd` as it fits directly.
> 
> @tldahlgren I was not able to use the `variant` keyword for `with_or_without` together with the `activation_value`. I ended up with a `KeyError: 'No spec with name bzlib in ...'`

Use the `variant` keyword?

tldahlgren:
Getting the following for one stack:

```
Waiting for pod pipeline/runner-vgquf2ek-project-2-concurrent-0ht6jt to be running, status is Pending
[14](https://gitlab.spack.io/spack/spack/-/jobs/4231045#L14)	ContainersNotInitialized: "containers with incomplete status: [init-permissions]"
[15](https://gitlab.spack.io/spack/spack/-/jobs/4231045#L15)	ContainersNotReady: "containers with unready status: [build helper]"
[16](https://gitlab.spack.io/spack/spack/-/jobs/4231045#L16)	ContainersNotReady: "containers with unready status: [build helper]"
```

@spackbot run pipeline

mwkrentel:
Well, it's still not technically correct.  The with/without for zstd
should be conditioned on version >= 0.182, or else you're adding the
config option `--without-zstd` to a version that doesn't know this option.

You could do this with `if` or `with`.  But whatever, it's only a warning,
I don't think it will actually break anything.  So let's say close enough
and move on.

haampie:
@mwkrentel see https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#conditional-variants, it's correct

mwkrentel:
Ah, Ok, then I approve.
Sorry, I wasn't giving spack enough credit.

mwkrentel (approved):

PR #26766, opened 2021-10-15
WIP: "spack uninstall": don't modify env

scheibelp:
Currently if you run `spack uninstall x` in an environment, and `x` refers to a user spec (e.g. something added with `spack add`), Spack will update the `spack.yaml` to remove that spec. This changes the behavior so that `spack uninstall` *just* uninstalls the spec: if you immediately run `spack install` in the environment (which in Spack is a request to install all specs registered with the environment), then - with this PR - it would reinstall the spec.

This is anticipated to be useful if a user is modifying the package.py for a package in their environment: they may want to uninstall and then reinstall it to get the latest state (at the moment, modifications to the package.py don't affect the hash that Spack uses to compute the install prefix and thereby to distinguish it from other installed specs).

`spack rm` can already remove specs, so there is still a command line mechanism available to edit the environment. 

This could interfere with scripts that depend on the current behavior of `spack uninstall`, although I consider that unlikely since this I don't think users would generally script uninstall actions.

If you have environments `e1` and `e2`, and specs [X, Y] such that [X, Y] is in `e1`, and [X] is in `e2`

* `spack uninstall y` in `e1`: uninstalls the spec (but with this PR it stays in e1)
* `spack uninstall x` in `e1`: complains because `x` is also in `e2`
*  (new) `spack uninstall --remove x` in `e1`: removes `x` from `e1` but does not actually uninstall the spec
* `spack uninstall -f x` in `e1`: uninstalls the spec (both e1 and e2 are now "missing" this spec)

If you have environments e1 and e2, and specs [P, Q, R] such that `P->R`, `Q->R`, and [P, R] are in e1 and [Q, R] are in e2

* `spack uninstall p` in `e1`: uninstalls the spec
* `spack uninstall r` in `e1`: will fail for two reasons: (1) P needs R in e1 and (2) e2 needs R
* `spack uninstall --dependents r` in `e1`: will fail because `e2` needs R
* (new) `spack uninstall --dependents --remove r` in `e1`: removes R from `e1` (but does not uninstall it) and uninstalls (and removes) P
* [C0] `spack uninstall -f --dependents r` in `e1`: will uninstall P, Q, and R (i.e. e2 will have dependent specs uninstalled as a side effect)
* (new) `spack uninstall -f --dependents --remove r` in `e1`: this uninstalls P, Q, and R, and removes [P, R] from `e1`
[...]

PR #33711, opened 2022-11-04
"spack uninstall": don't modify env

scheibelp:
Supersedes #26766

This is a redo of https://github.com/spack/spack/pull/26766. I found it problematic to merge (even after applying `spack style --fix` to the affected files, the merge listed a bunch of conflicts that seemed more effort than they were worth to resolve compared to "manual" merge).

All the desired behaviors from https://github.com/spack/spack/pull/26766#issue-1027660929 apply here. I need to do a more comprehensive examination of the underlying and related logic to know if more-extensive changes are required, but FWIW porting the statements over was straightforward.

### DIFF
--- a/lib/spack/spack/cmd/uninstall.py
+++ b/lib/spack/spack/cmd/uninstall.py
@@ -371,6 +371,13 @@ def get_uninstall_list(args, specs, env):
     remove_only = set()
     if args.remove and not args.force:
         remove_only.update(spec_to_other_envs)
+    if remove_only:
+        tty.debug(
+            "The following specs will be removed but not uninstalled because"
+            " they are also used by another environment: {speclist}".format(
+                speclist=", ".join(x.name for x in remove_only)
+            )
+        )
 
     # Compute the set of specs that should be removed from the current env.
     # This may overlap (some specs may be uninstalled and also removed from
--- a/lib/spack/spack/cmd/uninstall.py
+++ b/lib/spack/spack/cmd/uninstall.py
@@ -62,9 +62,12 @@ def setup_parser(subparser):
         help="remove regardless of whether other packages or environments " "depend on this one",
     )
     subparser.add_argument(
-         '--remove', action='store_true', dest='remove',
-         help="if in an environment, then the spec should also be removed from "
-         "the environment description")
+        "--remove",
+        action="store_true",
+        dest="remove",
+        help="if in an environment, then the spec should also be removed from "
+        "the environment description",
+    )
     arguments.add_common_arguments(
         subparser, ["recurse_dependents", "yes_to_all", "installed_specs"]
     )
@@ -275,22 +278,21 @@ def do_uninstall(env, specs, force):
 
 def get_uninstall_list(args, specs, env):
     """Returns uninstall_list and remove_list: these may overlap (some things
-       may be both uninstalled and removed from the current environment).
+    may be both uninstalled and removed from the current environment).
 
-       It is assumed we are in an environment if --remove is specified (this
-       method raises an exception otherwise).
+    It is assumed we are in an environment if --remove is specified (this
+    method raises an exception otherwise).
 
-       uninstall_list is topologically sorted: dependents come before
-       dependencies (so if a user uninstalls specs in the order provided,
-       the dependents will always be uninstalled first).
+    uninstall_list is topologically sorted: dependents come before
+    dependencies (so if a user uninstalls specs in the order provided,
+    the dependents will always be uninstalled first).
     """
     if args.remove and not env:
-         raise ValueError("Can only use --remove when in an environment")
+        raise ValueError("Can only use --remove when in an environment")
 
     # Gets the list of installed specs that match the ones given via cli
     # args.all takes care of the case where '-a' is given in the cli
-    base_uninstall_specs = set(
-        find_matching_specs(env, specs, args.all, args.force))
+    base_uninstall_specs = set(find_matching_specs(env, specs, args.all, args.force))
 
     active_dpts, outside_dpts = installed_dependents(base_uninstall_specs, env)
     # It will be useful to track the unified set of specs with dependents, as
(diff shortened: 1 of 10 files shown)


lib/spack/spack/cmd/uninstall.py
```diff
@@ -296,65 +344,85 @@ def get_uninstall_list(args, specs, env):
...
+    remove_only = set()
+    if args.remove and not args.force:
+        remove_only.update(spec_to_other_envs)
+    if remove_only:
+        tty.info(
+            "The following specs will be removed but not uninstalled because"
+            " they are also used by another environment: {speclist}".format(
+                speclist=", ".join(x.name for x in remove_only)
+            )
+        )
```
scheibelp, line 380: This is intended to resolve https://github.com/spack/spack/pull/26766#discussion_r1014522846 - I assume we don't want to record this as an error, since we want to support it.

All other comments from https://github.com/spack/spack/pull/26766#pullrequestreview-1169127076 are addressed without any caveats.

lib/spack/spack/test/cmd/uninstall.py
```diff
@@ -166,3 +169,175 @@ def _warn(*args, **kwargs):
...
+    e2 has dt-diamond-right -> dt-diamond-bottom
+    """
+
+    env = SpackCommand("env")
+    add = SpackCommand("add")
+    concretize = SpackCommand("concretize")
+    find = SpackCommand("find")
+
+    @pytest.fixture
+    def environment_setup(self, mutable_mock_env_path, config, mock_packages, mutable_database):
```
becker33: These should use the `install_mockery` fixture, then you won't need to specify fake for all the installs

lib/spack/spack/test/cmd/uninstall.py
```diff
@@ -166,3 +169,175 @@ def _warn(*args, **kwargs):
...
+    def test_uninstall_dependency_shared_between_envs_fail(self, environment_setup):
+        """If you "spack uninstall --dependents dt-diamond-bottom" from
+        e1 (without --remove or -f), then this should fail (this is needed by
+        e2).
+        """
+        e1 = spack.environment.read("e1")
+        with e1:
+            output = uninstall("-y", "--dependents", "dt-diamond-bottom", fail_on_error=False)
+            assert "There are still dependents." in output
+            assert "use `spack env remove`" in output
```
becker33, line 285: We should use `pytest.raises` for this, instead of `fail_on_error=False`. The former ensures that we actually raise an error, and don't just print it. There is a `match` parameter for the string to match in the error message.

lib/spack/spack/test/cmd/uninstall.py
```diff
@@ -166,3 +169,175 @@ def _warn(*args, **kwargs):
...
+            )
+            uninstall("-f", "-y", "--remove", "dt-diamond-bottom")
+            # dt-diamond-bottom was removed from the list of roots (note that
+            # it would still be installed since dt-diamond-left depends on it)
+            assert set(x.name for x in e1.roots()) == set(["dt-diamond-left"])
+            assert dtdiamondleft.package.installed
+
+        e2 = spack.environment.read("e2")
+        with e2:
+            assert set(root.name for (root, _) in e2.concretized_specs()) == set(
```
becker33, line 341: Can these be simplified using `set(root.name for root in e2.roots()) == set(...)`

lib/spack/spack/test/cmd/uninstall.py
```diff
@@ -166,3 +169,175 @@ def _warn(*args, **kwargs):
...
+                ["dt-diamond-right", "dt-diamond-bottom"]
+            )
+            (dtdiamondright,) = (
+                y for (x, y) in e2.concretized_specs() if y.name == "dt-diamond-right"
+            )
```
becker33:
```suggestion
            dtdiamondright = next(
                concrete for abstract, concrete in e2.concretized_specs() 
                if concrete.name == "dt-diamond-right"
            )
```
I think this is a bit easier to read this way, and would apply in a few other places

becker33 (changes requested):

lib/spack/spack/test/cmd/uninstall.py
scheibelp: Your suggestion is necessary as otherwise I will modify the real spack store, but `--fake` is still needed since the packages I'm using aren't legitimately install-able

lib/spack/spack/test/cmd/uninstall.py
scheibelp: I want to avoid a case where the roots are set but the set of concretized things is empty - leading to the next check being (incorrectly) vacuously true.

scheibelp: In other words, I want to ensure that (a) there are concretized specs and (b) that they are installed. Making this change would mean the test no longer guarantees (a)

lib/spack/spack/test/cmd/uninstall.py
scheibelp: When I expect there to be one element, I prefer `(x,) =`, and I consider it idiomatic Python, but also don't think it's worth an iteration of debate so I have made this change (everywhere applicable).

lib/spack/spack/test/cmd/uninstall.py
```diff
@@ -166,3 +169,192 @@ def _warn(*args, **kwargs):
...
+        """
+        e1 = spack.environment.read("e1")
+        with e1:
+            # FAILURE: this attempts to emulate tests in cmd/install.py but
+            # according to logic in main.py, none of the output printed by
+            # a command will be included in the exception
+            with pytest.raises(
+                spack.error.SpackError,
+                match=r".*Will not uninstall.*There are still dependents.*use `spack env remove`",
+            ):
```
scheibelp: @becker33 this is my current attempt at following https://github.com/spack/spack/pull/33711#discussion_r1015941410 , and something I'm doing is incorrect and I also am having trouble following the logic to see how this output would make it into the exception: I assumed it wouldn't based on reading `main.py:SpackCommand.__call__`, but as noted in this comment, there are tests which seem to be able to examine output.

I'll note one difference between this and `test/cmd/install.py:test_install_invalid_spec` is that the error is actually raised as an exception (via parse.py) which triggers:

```
            except BaseException as e:
                tty.debug(e)
                self.error = e
                if fail_on_error:
                    self._log_command_output(out)
                    raise
```

So that's my theory of why tests which  check for failures in commands typically run with `fail_on_error=False` (I originally copied this strategy from looking elsewhere, which is admittedly naive but on further inspection it would seem necessary pending a refactor of `SpackCommand`).

lib/spack/spack/test/cmd/uninstall.py
scheibelp: We agreed based on discussion relating to https://github.com/spack/spack/pull/33711#discussion_r1016006856 that this can stay for now (but some stuff should be refactored to make your request possible, since that seems like the right way to do things).

becker33 (approved):

PR #33717, opened 2022-11-04
MesonPackage: disable automatic download and install of dependencies

michaelkuhn:
Without this, Meson will use its Wraps to automatically download and install dependencies. We want to manage dependencies explicitly, therefore disable this functionality.


adamjstewart (approved):

adamjstewart:
We'll see how many packages this breaks...

Wish we could do this for rust too.

haampie:
Can you make a dummy change to the meson package itself to trigger CI to build all meson packages? (The CI situation should be slightly better now, except for random "system failures")

adamjstewart:
Could also run `spackbot rebuild everything` to rebuild all packages, not just meson.

eli-schwartz (approved):
Makes sense.

You can also use nofallback instead of nodownload, depending on whether you want to allow subproject dependencies if they were already available (such as shipped inside the release tarball). Probably not super important.

haampie:
@spackbot rebuild everything

haampie:
hm, the commit it's rebuilding everything on doesn't include 3346c0918bb939f3e82033ddde0a2378c602650d so it may actually fail, but let's see.

PR #33720, opened 2022-11-04
add umt package

aaroncblack:
Add package for the UMT code.

See https://github.com/LLNL/UMT for more information on this code.

### DIFF
--- a/bin/spack-tmpconfig
+++ b/bin/spack-tmpconfig
@@ -72,6 +72,7 @@ config:
     root: $TMP_DIR/install
   misc_cache: $$user_cache_path/cache
   source_cache: $$user_cache_path/source
+  environments_root: $TMP_DIR/envs
 EOF
 cat >"$SPACK_USER_CONFIG_PATH/bootstrap.yaml" <<EOF
 bootstrap:
--- a/etc/spack/defaults/config.yaml
+++ b/etc/spack/defaults/config.yaml
@@ -81,6 +81,10 @@ config:
   source_cache: $spack/var/spack/cache
 
 
+  ## Directory where spack managed environments are created and stored
+  # environments_root: $spack/var/spack/environments
+
+
   # Cache directory for miscellaneous files, like the package index.
   # This can be purged with `spack clean --misc-cache`
   misc_cache: $user_cache_path/cache
--- a/lib/spack/docs/environments.rst
+++ b/lib/spack/docs/environments.rst
@@ -58,9 +58,9 @@ Using Environments
 Here we follow a typical use case of creating, concretizing,
 installing and loading an environment.
 
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Creating a named Environment
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Creating a managed Environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 An environment is created by:
 
@@ -72,7 +72,8 @@ Spack then creates the directory ``var/spack/environments/myenv``.
 
 .. note::
 
-   All named environments are stored in the ``var/spack/environments`` folder.
+   All managed environments by default are stored in the ``var/spack/environments`` folder.
+   This location can be changed by setting the ``environments_root`` variable in ``config.yaml``.
 
 In the ``var/spack/environments/myenv`` directory, Spack creates the
 file ``spack.yaml`` and the hidden directory ``.spack-env``.
--- a/lib/spack/spack/cmd/env.py
+++ b/lib/spack/spack/cmd/env.py
@@ -165,7 +165,7 @@ def env_activate(args):
         short_name = os.path.basename(env_path)
         ev.Environment(env).write(regenerate=False)
 
-    # Named environment
+    # Managed environment
     elif ev.exists(env_name_or_dir) and not args.dir:
         env_path = ev.root(env_name_or_dir)
         short_name = env_name_or_dir
--- a/lib/spack/spack/cmd/location.py
+++ b/lib/spack/spack/cmd/location.py
@@ -95,7 +95,7 @@ def location(parser, args):
             spack.cmd.require_active_env("location -e")
             path = ev.active_environment().path
         else:
-            # Get named environment path
+            # Get path of requested environment
             if not ev.exists(args.location_env):
                 tty.die("no such environment: '%s'" % args.location_env)
             path = ev.root(args.location_env)
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -62,8 +62,8 @@ spack_env_var = "SPACK_ENV"
 _active_environment = None
 
 
-#: path where environments are stored in the spack tree
-env_path = os.path.join(spack.paths.var_path, "environments")
+#: default path where environments are stored in the spack tree
+default_env_path = os.path.join(spack.paths.var_path, "environments")
 
 
 #: Name of the input yaml file for an environment
@@ -76,6 +76,26 @@ lockfile_name = "spack.lock"
 
 #: Name of the directory where environments store repos, logs, views
 env_subdir_name = ".spack-env"
+
+
+def env_root_path():
+    """Override default root path if the user specified it"""
+    return spack.util.path.canonicalize_path(
+        spack.config.get("config:environments_root", default=default_env_path)
+    )
+
+
+def check_disallowed_env_config_mods(scopes):
+    for scope in scopes:
+        with spack.config.use_configuration(scope):
+            if spack.config.get("config:environments_root"):
+                raise SpackEnvironmentError(
+                    "Spack environments are prohibited from modifying 'config:environments_root' "
+                    "because it can make the definition of the environment ill-posed. Please "
+                    "remove from your environment and place it in a permanent scope such as "
+                    "defaults, system, site, etc."
+                )
+    return scopes
 
 
 def default_manifest_yaml():
@@ -214,7 +234,7 @@ def active_environment():
 
 def _root(name):
     """Non-validating version of root(), to be used internally."""
-    return os.path.join(env_path, name)
+    return os.path.join(env_root_path(), name)
 
 
 def root(name):
@@ -249,10 +269,12 @@ def read(name):
 
 
 def create(name, init_file=None, with_view=None, keep_relative=False):
-    """Create a named environment in Spack."""
+    """Create a managed environment in Spack."""
+    if not os.path.isdir(env_root_path()):
+        fs.mkdirp(env_root_path())
     validate_env_name(name)
     if exists(name):
-        raise SpackEnvironmentError("'%s': environment already exists" % name)
+        raise SpackEnvironmentError("'%s': environment already exists at %s" % (name, root(name)))
     return Environment(root(name), init_file, with_view, keep_relative)
 
 
@@ -266,10 +288,10 @@ def all_environment_names():
     """List the names of environments that currently exist."""
     # just return empty if the env path does not exist.  A read-only
     # operation like list should not try to create a directory.
-    if not os.path.exists(env_path):
+    if not os.path.exists(env_root_path()):
         return []
 
-    candidates = sorted(os.listdir(env_path))
+    candidates = sorted(os.listdir(env_root_path()))
     names = []
     for candidate in candidates:
         yaml_path = os.path.join(_root(candidate), manifest_name)
@@ -279,7 +301,7 @@ def all_environment_names():
 
 
 def all_environments():
-    """Generator for all named Environments."""
+    """Generator for all managed Environments."""
     for name in all_environment_names():
         yield read(name)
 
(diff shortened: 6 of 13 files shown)


var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+
+    #######################
+    # CMake
+    #######################
+    depends_on("cmake@3.13.3:", type="build")
+
+    #######################
+    # Dependencies
+    #######################
+    depends_on("mpi")
```
becker33:
```suggestion
    depends_on("mpi", when="+mpi")
```
I'm assuming you don't need it when MPI is off

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+        # if on llnl systems, we can use the SYS_TYPE
+        if "SYS_TYPE" in env:
+            sys_type = env["SYS_TYPE"]
+        host_config_path = "{0}-{1}-{2}-{3}.cmake".format(
+            socket.gethostname(), sys_type, spec.compiler, spec.dag_hash()
+        )
+        dest_dir = spec.prefix
+        host_config_path = os.path.abspath(join_path(dest_dir, host_config_path))
+
+        return host_config_path
```
becker33:
```suggestion
```
It doesn't look like this method is called anywhere

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # Override the parent class function until its fixed in spack.  (Greg Becker and Chris White are aware of bug)
+    def flag_handler(self, name, flags):
+        if name in ("cflags", "cxxflags", "cppflags", "fflags", "ldflags"):
+            return (None, None, None)  # handled in the cmake cache
+        return (flags, None, None)
```
becker33: Let's just fix this as part of this PR. I can push a fix.

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+        return "{0}-{1}-{2}@{3}.cmake".format(
+            hostname,
+            self._get_sys_type(self.spec),
+            self.spec.compiler.name,
+            self.spec.compiler.version,
+        )
+
+    def cmake_args(self):
+        options = []
+        return options
```
becker33:
```suggestion
```
This is effectively a no-op

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_RELEASE", "-qsmp=omp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_DEBUG", "-qsmp=omp"))
+
+                if "+openmp" in spec:
+                    entries.append(cmake_cache_string("TETON_OpenMP_Offload_Fortran_FLAGS", "-qoffload -qtgtarch=sm_70"))
+            else:
+                entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_RELEASE", "-fopenmp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_DEBUG", "-fopenmp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_RELEASE", "-fopenmp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_DEBUG", "-fopenmp"))
```
becker33:
```suggestion
            entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_RELEASE", self.compiler.openmp_flag))
            entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_DEBUG", self.compiler.openmp_flag))
            entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_RELEASE", self.compiler.openmp_flag))
                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_DEBUG", self.compiler.openmp_flag))
```

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if "+openmp" in spec:
+            if spec.satisfies("%xl"):
+                entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_RELEASE", "-qsmp=omp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_DEBUG", "-qsmp=omp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_RELEASE", "-qsmp=omp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_DEBUG", "-qsmp=omp"))
+
+                if "+openmp" in spec:
+                    entries.append(cmake_cache_string("TETON_OpenMP_Offload_Fortran_FLAGS", "-qoffload -qtgtarch=sm_70"))
```
becker33: This needs to be under the `openmp_offlag` variant, not the openmp variant, and it should have another variant for the NVIDIA architecture (standard in Spack is `cuda_arch`, and you can look at `CUDAPackage` to see how it can be conditional on another variant, although I see you may want a different name since it's fortran offload only).

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def initconfig_hardware_entries(self):
+        spec = self.spec
+        entries = super(Umt, self).initconfig_hardware_entries()
+
+        #######################
+        # Parallelism
+        #######################
+        if "+cuda" in spec:
+            entries.append(cmake_cache_option("ENABLE_CUDA", True))
```
becker33: There is no cuda variant on the package

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+        if spec.satisfies("%cce"):
+            entries.append(cmake_cache_option("OPENMP_HAS_USE_DEVICE_ADDR", True))
+            entries.append(cmake_cache_option("OPENMP_HAS_FORTRAN_INTERFACE", True))
+
+        return entries
+
+    def initconfig_mpi_entries(self):
+        entries = super(Umt, self).initconfig_mpi_entries()
+
+        return entries
```
becker33:
```suggestion
```
This is a no-op

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+            if "hypre" in spec:
+                entries.append(cmake_cache_path("HYPRE_ROOT", spec["hypre"].prefix))
+            if "metis" in spec:
+                entries.append(cmake_cache_path("METIS_ROOT", spec["metis"].prefix))
+            
+        if "+caliper" in spec:
+            entries.append(cmake_cache_option("ENABLE_CALIPER", True))
+            entries.append(cmake_cache_path("CALIPER_ROOT", spec["caliper"].prefix))
+            if "adiak" in spec:
+                entries.append(cmake_cache_path("ADIAK_ROOT", spec["adiak"].prefix))
```
becker33:
```suggestion
            entries.append(cmake_cache_path("ADIAK_ROOT", spec["adiak"].prefix))
```
The dependency guarantees it's there

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+
+        entries.append(cmake_cache_path("CONDUIT_ROOT", spec["conduit"].prefix))
+
+        if "+mfem" in spec:
+            entries.append(cmake_cache_option("ENABLE_MFEM", True))
+            entries.append(cmake_cache_path("MFEM_ROOT", spec["mfem"].prefix))
+            if "hypre" in spec:
+                entries.append(cmake_cache_path("HYPRE_ROOT", spec["hypre"].prefix))
+            if "metis" in spec:
+                entries.append(cmake_cache_path("METIS_ROOT", spec["metis"].prefix))
```
becker33: If the package needs to know about these, they should be exposed as direct dependencies

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+            entries.append(cmake_cache_path("CALIPER_ROOT", spec["caliper"].prefix))
+            if "adiak" in spec:
+                entries.append(cmake_cache_path("ADIAK_ROOT", spec["adiak"].prefix))
+
+        if "+umpire" in spec:
+            entries.append(cmake_cache_option("ENABLE_UMPIRE", True))
+            entries.append(cmake_cache_path("UMPIRE_ROOT", spec["umpire"].prefix))
+            if "camp" in spec:
+                entries.append(cmake_cache_option("ENABLE_CAMP", True))
+                entries.append(cmake_cache_path("CAMP_ROOT", spec["camp"].prefix))
```
becker33:
```suggestion
            entries.append(cmake_cache_option("ENABLE_CAMP", True))
            entries.append(cmake_cache_path("CAMP_ROOT", spec["camp"].prefix))
```
The dependency ensures it's present.

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+        if "+cuda" in spec:
+            entries.append(cmake_cache_option("ENABLE_CUDA", True))
+        if "+openmp" in spec:
+            entries.append(cmake_cache_option("ENABLE_OPENMP", True))
+        if "+openmp_offload" in spec:
+            entries.append(cmake_cache_option("ENABLE_OPENMP_OFFLOAD", True))
+
+        if spec.satisfies("%cce"):
+            entries.append(cmake_cache_option("OPENMP_HAS_USE_DEVICE_ADDR", True))
+            entries.append(cmake_cache_option("OPENMP_HAS_FORTRAN_INTERFACE", True))
```
becker33: This should probably be buried in the `if "+openmp" in spec` block above, and should probably be in the compilers section

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+        if name in ("cflags", "cxxflags", "cppflags", "fflags", "ldflags"):
+            return (None, None, None)  # handled in the cmake cache
+        return (flags, None, None)
+
+    def initconfig_compiler_entries(self):
+        spec = self.spec
+
+        if "umpire" in spec:
+            if "+rocm" in spec["umpire"]:
+                spec.compiler_flags["cxxflags"].append("-I{0}/include".format(spec["hip"].prefix))
```
becker33: We should make a `flags_to_build_system_args` method for `CachedCMakePackage` so that we don't have to do things like this, but that might be future work

var/spack/repos/builtin/packages/umt/package.py
```diff
@@ -0,0 +1,196 @@
...
+                    entries.append(cmake_cache_string("TETON_OpenMP_Offload_Fortran_FLAGS", "-qoffload -qtgtarch=sm_70"))
+            else:
+                entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_RELEASE", "-fopenmp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_Fortran_FLAGS_DEBUG", "-fopenmp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_RELEASE", "-fopenmp"))
+                entries.append(cmake_cache_string("TETON_OpenMP_CXX_FLAGS_DEBUG", "-fopenmp"))
+
+        if spec.satisfies("%cce"):
+            entries.append(cmake_cache_option("STRICT_FPP_MODE", True))
+            entries.append(cmake_cache_string("TETON_OpenMP_Offload_LINK_FLAGS", "-ldl"))
```
becker33: This should be under the `if "+openmp" in spec` block above

becker33 (changes requested):

var/spack/repos/builtin/packages/umt/package.py
aaroncblack: MPI is mandatory.

becker33: @aaroncblack then we should remove the mpi variant above

var/spack/repos/builtin/packages/umt/package.py
aaroncblack: This is a bit tricky for XL.  XL has two different openmp flags for OpenMP, and enabling OpenMP target offload kernels.
"-qsmp=omp" enables OpenMP, but not 'target' constructs.
"-qsmp=omp -qoffload" enables OpenMP and target constructs.

But I'll probably deprecate support for both of these and just allow the latter line or no line.  I need to double check if self.compiler.openmp_flags provides the full "-qsmp=omp -qoffload" on IBM platforms.

Or, I'll just put this logic into the UMT CMake instead of having the spack package deal with it.

var/spack/repos/builtin/packages/umt/package.py
aaroncblack: I'll add these.  UMT does not directly call anything in hypre or metis, but we do use mfem, and mfem uses them so I need to link these in for our executable.

alalazo:
@aaroncblack

PR #33721, opened 2022-11-05
Python: default to v3.10

adamjstewart:
Python 3.11 is out now. I think it's time to update the default version of Python to 3.10. For comparison:

* [Homebrew](https://formulae.brew.sh/formula/python@3.10) already defaults to 3.10
* [Anaconda](https://www.anaconda.com/) still defaults to 3.9

Let's see if CI likes this change.

Closes #31623 @haampie (unless you want to reopen your PR...)

### DIFF
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -44,7 +44,11 @@ class Python(Package):
     install_targets = ["install"]
     build_targets = []  # type: List[str]
 
-    version("3.10.8", sha256="f400c3fb394b8bef1292f6dc1292c5fadc3533039a5bc0c3e885f3e16738029a")
+    version(
+        "3.10.8",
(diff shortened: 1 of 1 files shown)


tldahlgren (approved):
Confirmed both ~new~ modified sha256 are still accurate.

adamjstewart:
There are no new sha256 lol

tldahlgren:
> There are no new sha256 lol

:facepalm:  You're right.  I should've paid more attention.  (I have a script that will check any version directives that show up in the diffs.)

adamjstewart:
We should add that to spackbot, or directly in package audits.

haampie:
@adamjstewart I closed the old pr because I added it to the new python 3.11 pr #33507. To me it doesn't matter much which pr goes in first, as long as we can get python 3.11 in before the 0.19 release ;)

adamjstewart:
Ah didn't notice that, thanks!

PR #33722, opened 2022-11-05
unparser: fix bug in unit test assertion

alalazo:


lib/spack/spack/test/util/unparse/unparse.py
```diff
@@ -178,16 +178,14 @@ async def f():
 """
 
 
-def assertASTEqual(ast1, ast2):
-    ast.dump(ast1) == ast.dump(ast2)
```
alalazo, line 182: @tgamblin The assertion was missing here

alalazo:
@tgamblin I'll go ahead and merge this myself, so to see in #33505 which Python source is causing issues. This is just for information since the unit test doesn't fail locally for me. Then I hope to have fixes for #33505 too.

PR #33716, opened 2022-11-04
py-transformers: add v4.24.0

adamjstewart:
Successfully builds on macOS 12.6.1 (arm64) with Python 3.9.13 and Apple Clang 14.0.0.

This is something we should test in CI. Transformers are incredibly popular right now, and this package previously wasn't installable.

Closes #32444

Issue #32444, opened 2022-08-30
py-transformers unable to be found


cameronrutherford:
### Steps to reproduce

This is quite a simple error to reproduce, and has consistently been an issue for me for a while now with various versions of spack.
```console
$ spack spec py-transformers
```
or, to see this error in an environment:
```console
$ spack env create -d ./test-env
$ spacktivate -p ./test-env
$ spack add py-transformers
$ spack concretize -f
```

### Error message

```console
$ spack --debug --stacktrace spec py-transformers
```
<details> <summary>Full error output</summary>

```console
lib/spack/spack/config.py:1008 ==> [2022-08-30-11:22:24.766764] Reading config file /Users/ruth521/Projects/spack/spack/etc/spack/defaults/config.yaml
lib/spack/spack/config.py:1008 ==> [2022-08-30-11:22:24.794562] Reading config file /Users/ruth521/Projects/spack/spack/etc/spack/defaults/concretizer.yaml
lib/spack/spack/config.py:1008 ==> [2022-08-30-11:22:24.802347] Reading config file /Users/ruth521/Projects/spack/test-env/spack.yaml
lib/spack/spack/environment/environment.py:173 ==> [2022-08-30-11:22:24.804814] Using environment '/Users/ruth521/Projects/spack/test-env'
lib/spack/spack/cmd/__init__.py:124 ==> [2022-08-30-11:22:24.813121] Imported spec from built-in commands
lib/spack/spack/cmd/__init__.py:124 ==> [2022-08-30-11:22:24.814080] Imported spec from built-in commands
lib/spack/spack/config.py:1008 ==> [2022-08-30-11:22:24.848930] Reading config file /Users/ruth521/Projects/spack/spack/etc/spack/defaults/bootstrap.yaml
lib/spack/spack/environment/environment.py:197 ==> [2022-08-30-11:22:24.857615] Deactivated environment '/Users/ruth521/Projects/spack/test-env'
lib/spack/spack/config.py:1008 ==> [2022-08-30-11:22:24.900861] Reading config file /Users/ruth521/Projects/spack/spack/etc/spack/defaults/config.yaml
lib/spack/spack/database.py:389 ==> [2022-08-30-11:22:24.919398] DATABASE LOCK TIMEOUT: 3s
lib/spack/spack/database.py:395 ==> [2022-08-30-11:22:24.919670] PACKAGE LOCK TIMEOUT: No timeout
[...]

PR #33723, opened 2022-11-05
Rust: don't apply constraints to nightly/beta versions

adamjstewart:
This is a follow-up to #33716. See https://github.com/spack/spack/issues/32444#issuecomment-1304278896 for a description of why this change is necessary.

TL;DR: `@nightly` doesn't behave like `@master` and is treated as older than all stable releases. By applying constraints like the Python 2 dependency only on older stable releases, it becomes possible to correctly concretize `@nightly`.

### DIFF
--- a/var/spack/repos/builtin/packages/py-tokenizers/package.py
+++ b/var/spack/repos/builtin/packages/py-tokenizers/package.py
@@ -21,7 +21,9 @@ class PyTokenizers(PythonPackage):
     depends_on("py-setuptools", type="build")
     depends_on("py-setuptools-rust", type="build")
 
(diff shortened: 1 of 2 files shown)


adamjstewart:
Alternative is to rename nightly and beta to something in the [list of infinity versions](https://github.com/spack/spack/blob/v0.18.1/lib/spack/spack/version.py#L62), but this isn't technically correct since nightly/beta could be between two releases and not actually newer.

tldahlgren (approved):
LGTM

adamjstewart:
spackbot has been bad at requesting reviews lately. @AndrewGaspar let me know if anything in this PR looks bad to you and I can revert.

PR #33733, opened 2022-11-07
Patch `fmt` for hipcc/dpcpp

### LINKED ISSUE #3005: dpcpp/IntelLLVM: Undefined template fmt::detail::dragonbox::float_info<long double>
With latest [dpcpp](https://www.intel.com/content/www/us/en/developer/tools/oneapi/data-parallel-c-plus-plus.html#gs.7mxi91) (Intel's LLVM/clang based SYCL compiler - which CMake classes as IntelLLVM), fmt 9.0.0 (and main branch) fails to compile with 

> `error: implicit instantiation of undefined template 'fmt::detail::dragonbox::float_info<long double>'`

Dockerfile to reproduce with current latest intel image (warning, this base image is a 20GB image/6GB download)
```dockerfile
FROM intel/oneapi-basekit:2022.2-devel-ubuntu20.04

WORKDIR /opt

RUN git clone https://github.com/fmtlib/fmt.git
RUN mkdir _build && \
    cd _build && \
    CXX=$(which dpcpp) cmake ../fmt
RUN cd _build && cmake --build .
```

Error output when running:

```
/opt/fmt/include/fmt/format.h:3098:38: error: implicit instantiation of undefined template 'fmt::detail::dragonbox::float_info<long double>'
    const auto f = basic_fp<typename info::carrier_uint>(converted_value);
                                     ^
/opt/fmt/include/fmt/format.h:3206:13: note: in instantiation of function template specialization 'fmt::detail::format_float<long double>' requested here
  int exp = format_float(convert_float(value), precision, fspecs, buffer);
            ^
/opt/fmt/include/fmt/format.h:3370:20: note: in instantiation of function template specialization 'fmt::detail::write<char, fmt::appender, long double, 0>' requested here
    return detail::write(out, value, specs, locale);
                   ^
/opt/fmt/include/fmt/core.h:1648:12: note: in instantiation of function template specialization 'fmt::detail::arg_formatter<char>::operator()<long double>' requested here
    return vis(arg.value_.long_double_value);
           ^
/opt/fmt/include/fmt/format.h:4108:26: note: in instantiation of function template specialization 'fmt::visit_format_arg<fmt::detail::arg_formatter<char> &, fmt::basic_format_context<fmt::appender, char>>' requested here
      context.advance_to(visit_format_arg(f, arg));
                         ^
/opt/fmt/include/fmt/format.h:4058:10: note: in instantiation of member function 'fmt::detail::vformat_to(buffer<char> &, basic_string_view<char>, basic_format_args<basic_format_context<detail::buffer_appender<type_identity_t<char>>, type_identity_t<char>>>, fmt::detail::locale_ref)::format_handler::on_format_specs' requested here
  struct format_handler : error_handler {
         ^
/opt/fmt/src/format.cc:34:23: note: in instantiation of function template specialization 'fmt::detail::vformat_to<char>' requested here
template FMT_API void vformat_to(buffer<char>&, string_view,
                      ^
/opt/fmt/include/fmt/format.h:1272:54: note: template is declared here
template <typename T, typename Enable = void> struct float_info;
```

As I understand it, dpcpp is "supposed" to act as a reasonably up-to-date version of clang, which depends on system standard library support (so might not have all C++20 headers e.g. `<bit>` is definitely missing, and there are also later errors with "error: no matching function for call to 'bit_cast'", but I don't know if this is caused by a cascade in the error checks so isn't the primary point of this issue).

There definitely seems to be some "long double" handling e.g. https://github.com/fmtlib/fmt/blob/81f1cc74a776581cdef8659d176049d3aeb743c6/include/fmt/core.h#L1648 so, perhaps this compiler has a slightly unusual separation of definitions?

msimberg:
Applies https://github.com/fmtlib/fmt/commit/0b0f7cfbfcebd021c910078003d413354bd843e2 to `fmt` to fix https://github.com/fmtlib/fmt/issues/3005.

The patch has already been merged to master so the patch is constrained to versions `9.0.0:9.1.0`. The patch isn't needed for older versions and hopefully won't be needed for newer versions.

The patch is not constrained by compiler because `fmt` can be built with e.g. `g++` and compiled in a dependency by `hipcc`.

### DIFF
--- a/var/spack/repos/builtin/packages/fmt/package.py
+++ b/var/spack/repos/builtin/packages/fmt/package.py
@@ -13,6 +13,7 @@ class Fmt(CMakePackage):
 
     homepage = "https://fmt.dev/"
     url = "https://github.com/fmtlib/fmt/releases/download/7.1.3/fmt-7.1.3.zip"
+    maintainers = ["msimberg"]
 
     version("9.1.0", sha256="cceb4cb9366e18a5742128cb3524ce5f50e88b476f1e54737a47ffdf4df4c996")
     version("9.0.0", sha256="fc96dd2d2fdf2bded630787adba892c23cb9e35c6fd3273c136b0c57d4651ad6")
--- a/var/spack/repos/builtin/packages/fmt/package.py
+++ b/var/spack/repos/builtin/packages/fmt/package.py
@@ -70,6 +70,13 @@ class Fmt(CMakePackage):
     # Only allow [[attributes]] on C++11 and higher
     patch("fmt-attributes-cpp11_4.1.0.patch", when="@4.1.0")
 
+    # Fix compilation with hipcc/dpcpp: https://github.com/fmtlib/fmt/issues/3005
+    patch(
+        "https://github.com/fmtlib/fmt/commit/0b0f7cfbfcebd021c910078003d413354bd843e2.patch?full_index=1",
+        sha256="08fb707bf8b4fc890d6eed29217ead666558cbae38f9249e22ddb82212f0eb4a",
+        when="@9.0.0:9.1.0",
+    )
+
     def cmake_args(self):
         spec = self.spec
         args = []


albestro (approved):
LGTM, I checked SHA too.

ChristianTackeGSI:
LGTM, didn't check anything though.

Good to have a new maintainer, welcome!

tldahlgren (approved):

PR #33746, opened 2022-11-07
gitlab: Add shared PR mirror to places pipelines look for binaries.

scottwittenburg:
While binaries built for PRs that get merged must still be rebuilt in develop pipelines, they can be used by other PRs that find they would otherwise need to rebuild them.  Now that spackbot is managing copying PR binaries from merged PRs into a shared location, keeping it pruned to a reasonable size, and making sure the indices are up to date, spack can use these mirrors as a potential source of binaries (for PR pipelines only).


lib/spack/spack/cmd/ci.py
```diff
@@ -472,6 +472,9 @@ def ci_rebuild(args):
             spack.mirror.add("mirror_override", remote_mirror_override, cfg.default_modify_scope())
         pipeline_mirrors.append(remote_mirror_override)
 
+    if spack_pipeline_type == "spack_pull_request":
+        pipeline_mirrors.append(spack_ci.SHARED_PR_MIRROR_URL)
```
kwryankrattiger: Should this add them per-stack?

scottwittenburg: Oh, yeah, thanks.  Will fix.

kwryankrattiger (approved):

scottwittenburg:
@spackbot run pipeline

PR #33749, opened 2022-11-08
Rework unit test to avoid tripping into concretization slowdown

alalazo:
This is just a workaround to avoid slow-down in CI, while the underlying issue gets investigated. The issue seems to be triggered by `os=...` constraints in abstract specs.

### DIFF
--- a/lib/spack/spack/test/concretize_preferences.py
+++ b/lib/spack/spack/test/concretize_preferences.py
@@ -8,8 +8,6 @@ import stat
 
 import pytest
 
-import archspec
-
 import spack.config
 import spack.package_prefs
 import spack.repo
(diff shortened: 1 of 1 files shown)


haampie (approved):

lib/spack/spack/test/concretize_preferences.py
```diff
@@ -105,28 +103,16 @@ def test_preferred_variants_from_wildcard(self):
         update_packages("multivalue-variant", "variants", "foo=bar")
         assert_variant_values("multivalue-variant foo=*", foo=("bar",))
 
-    def test_preferred_compilers(self):
+    @pytest.mark.parametrize(
+        "compiler_str,spec_str",
+        [("gcc@4.5.0", "mpileaks"), ("clang@12.0.0", "mpileaks"), ("gcc@4.5.0", "openmpi")],
```
becker33, line 108: @alalazo this will fail on `aarch64`

becker33: Can we use `gcc@10.2.1` instead?

alalazo: Sure, whatever passes without using `os=` doesn't trigger the slowdon.

PR #33754, opened 2022-11-08
New Package: y-cruncher

saqibkh:
y-cruncher is a program that can compute Pi and other constants to trillions of digits.


bernhardkaindl (approved):
Works! :1st_place_medal:

PR #33755, opened 2022-11-08
Speed-up unit tests

alalazo:
Results on my machine:

* **Spack:** 0.19.0.dev0 (fd5f8a666af292e2513ebe206bdb60f1387bf6bd)
* **Python:** 3.8.10
* **Concretizer:** clingo

| Test file  | Time on develop | Time in this PR | Speedup |
| ----------- | ---------------------- | -------------------- | ------------ |
|  lib/spack/spack/test/architecture.py  | 18.18s  | 3.32s | **5.48x** |
| lib/spack/spack/test/spec_semantics.py  | 87.96s  | 11.18s | **7.87x** |
| lib/spack/spack/test/ci.py  | 24.79s  | 3.83s | **6.49x** |
| lib/spack/spack/test/url_fetch.py  | 53.32s  | 12.99s | **4.10x** |
| lib/spack/spack/test/git_fetch.py  | 26.56s  | 11.89s | **2.33x** |
| lib/spack/spack/test/directory_layout.py  | 17.66s  | 3.67s | **4.81x** |
| lib/spack/spack/test/build_systems.py  | 18.85s  | 12.86s | **1.47x** |
| lib/spack/spack/test/packaging.py  | 4.66s  | 2.65s | 1.76x |
| lib/spack/spack/test/database.py  | 9.34s  | 8.85s | 1.06x |

Modifications:
- [x] Avoid concretization in a few tests in `architecture.py`
- [x] Use mock packages everywhere in `spec_semantics`, reduce re-try in a test
- [x] Cache recurring concretizations in other places

### DIFF
--- a/lib/spack/spack/test/conftest.py
+++ b/lib/spack/spack/test/conftest.py
@@ -1861,8 +1861,13 @@ def concretized_specs_cache():
 
 
 @pytest.fixture
-def mock_concretize(config, mock_packages, concretized_specs_cache):
-    """Return a concretized spec from immutable mock configuration."""
+def default_mock_concretization(config, mock_packages, concretized_specs_cache):
+    """Return the default mock concretization of a spec literal, obtained using the mock
+    repository and the mock configuration.
+
+    This fixture is unsafe to call in a test when either the default configuration or mock
+    repository are not used or have been modified.
+    """
 
     def _func(spec_str, tests=False):
         key = spec_str, tests
--- a/lib/spack/spack/test/conftest.py
+++ b/lib/spack/spack/test/conftest.py
@@ -1852,3 +1852,22 @@ def binary_with_rpaths(prefix_tmpdir):
         return executable
 
     return _factory
+
+
+@pytest.fixture(scope="session")
+def concretized_specs_cache():
+    """Cache for mock concrete specs"""
+    return {}
+
+
+@pytest.fixture
+def mock_concretize(config, mock_packages, concretized_specs_cache):
+    """Return a concretized spec from immutable mock configuration."""
+
+    def _func(spec_str, tests=False):
+        key = spec_str, tests
+        if key not in concretized_specs_cache:
+            concretized_specs_cache[key] = spack.spec.Spec(spec_str).concretized(tests=tests)
+        return concretized_specs_cache[key].copy()
+
+    return _func
--- a/lib/spack/spack/test/architecture.py
+++ b/lib/spack/spack/test/architecture.py
@@ -17,7 +17,7 @@ import spack.spec
 import spack.target
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def current_host_platform():
     """Return the platform of the current host as detected by the
     'platform' stdlib package.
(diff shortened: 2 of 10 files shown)


lib/spack/spack/test/conftest.py
```diff
@@ -1852,3 +1852,24 @@ def _factory(rpaths, message="Hello world!"):
...
+def mock_concretize(config, mock_packages, concretized_specs_cache):
+    """Return a concretized spec from immutable mock configuration."""
+
+    def _func(spec_str, tests=False):
+        key = spec_str, tests
+        if key not in concretized_specs_cache:
+            result = spack.spec.Spec(spec_str).concretized(tests=tests)
+            assert result.concrete, "{} is expected to be concrete".format(spec_str)
+            concretized_specs_cache[key] = result
+        return concretized_specs_cache[key]
```
haampie: Wouldn't you wanna copy?

alalazo: I was doing it in some tests only, but given it's fast enough we can do it here and avoid doing it at the calling site.

lib/spack/spack/test/conftest.py
```diff
@@ -1852,3 +1852,24 @@ def _factory(rpaths, message="Hello world!"):
...
+
+@pytest.fixture
+def mock_concretize(config, mock_packages, concretized_specs_cache):
+    """Return a concretized spec from immutable mock configuration."""
+
+    def _func(spec_str, tests=False):
+        key = spec_str, tests
+        if key not in concretized_specs_cache:
+            result = spack.spec.Spec(spec_str).concretized(tests=tests)
+            assert result.concrete, "{} is expected to be concrete".format(spec_str)
```
becker33: This seems overly careful here

trws:
Given the notes and how it looks like this works, should `mock_concretize` perhaps be named something like `concretize_and_cache`?  Or maybe `cache_concrete_mock` if you meant to be invoking that it's concretizing a mock package specifically?

(edit) I should have said, I ask because I originally misread it as not actually concretizing, but mocking out concretization.

becker33 (dismissed):

lib/spack/spack/test/conftest.py
alalazo: There was this assertion in some tests, so I ported it here. I can remove it if people agree on that.

alalazo:
@trws I have no real preferences on the name. I called it `mock_concretize` since the caveat with that fixture is that it implies using the `config` and `mock_packages` fixtures. This is an important bit, as the fixture might return wrong results if used in tests that need to modify the configuration or the `builtin.mock` repository.[^1] So, if there is agreement on a better name I'm fine with that.

[^1]: The full story is that the fixture can be used in those tests only _before_ applying the modifications.

lib/spack/spack/test/conftest.py
haampie: I would also drop it

alalazo: Removed

becker33:
Doesn't the `config` fixture already imply that we won't modify the config, as in that case we have to use the `mutable_config` fixture? Similarly, I think we have another fixture that @scheibelp wrote for tests that need to modify a repo, and we don't use `builtin.mock` for that.

I do think this would be a little clearer for readers with a name like `caching_test_concretizer`, but (A) I'm more sold on the idea of a clearer name, but I don't have a good idea for one, and (B) I don't think it is make-or-break for the PR.

alalazo:
> Doesn't the config fixture already imply that we won't modify the config, as in that case we have to use the mutable_config fixture? 

It implies we don't modify the existing scopes. But in some tests we add more scopes and then concretize. `mock_concretize` shouldn't be used in those cases.

alalazo:
What do people think about `fast_concretize`? I wouldn't use "cache" since it ties an implementation detail to the name of the function (and we might decide to change that detail at some point in the future).

trws:
It isn't a fast concretize though, it's a cached concretize.  I don't think that's an implementation detail, it's the intended behavior, and it matters for some things like testing something after the first concretization or second which we have tests for.  Either way, honestly I want this merged a whole lot more than I want to worry about the name.  The point about it being unusable after modifications are made to config or mock_packages does seem important, but I'm not sure how to incorporate that.  If you don't like cached, how about `default_concretization` or maybe append `_of_mock`? I kinda like that because it implies you get the default concretization regardless of what you might set without saying explicitly that it uses a cache, the extension would make it clear it uses a mock package repo by default too.

alalazo:
@trws `default_mock_concretization` seems good to me, I'll push that change.

haampie:
For me this reduces runtime by 13% (with 11 processes w/ xdist), so a solid improvement

haampie (approved):

haampie:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

PR #33757, opened 2022-11-08
cosma: Reorder BLAS list to prefer GPU versions

loveshack:
Previously, +cuda built for me with openblas, not nvblas

### DIFF
--- a/var/spack/repos/builtin/packages/cosma/package.py
+++ b/var/spack/repos/builtin/packages/cosma/package.py
@@ -57,13 +57,13 @@ class Cosma(CMakePackage):
 
     def cosma_blas_cmake_arg(self):
         query_to_cmake_arg = [
+            ("+cuda", "CUDA"),
+            ("+rocm", "ROCM"),
             ("^intel-mkl", "MKL"),
             ("^intel-oneapi-mkl", "MKL"),
             ("^cray-libsci", "CRAY_LIBSCI"),
             ("^netlib-lapack", "CUSTOM"),
             ("^openblas", "OPENBLAS"),
-            ("+cuda", "CUDA"),
-            ("+rocm", "ROCM"),
         ]
 
         if self.version >= Version("2.4.0"):


teonnik:
If I build `cosma +cuda` I get the CUDA backend, not the OpenBLAS. Perhaps you have a stray `^openblas` which takes precedence over `+cuda`? Can you specify the spec you are trying to build?

loveshack:
> If I build `cosma +cuda` I get the CUDA backend, not the
> OpenBLAS. Perhaps you have a stray `^openblas` which takes precedence
> over `+cuda`? Can you specify the spec you are trying to build?

Yes, like this to ensure it gets the right CPU BLAS:

  cosma+cuda+scalapack ^netlib-scalapack ^openblas threads=openmp

(Spack tends to try amdlibflame/amdblis and fail on ppc64le, and you
don't want BLIS on POWER9, even if it built.)

Doesn't +cuda mean you specifically want cublas?

haampie:
Yeah, depends_on("blas") does not include cublas

teonnik:
@loveshack

>   cosma+cuda+scalapack ^netlib-scalapack ^openblas threads=openmp

It doesn't make sense to specify more than one BLAS backend for COSMA. Only one backend is supported per build. 

We can either 1) decide to emit an error message or a warning if more than one BLAS backend is specified in spec or 2) decide on consistent behavior - e.g. select the first one in spec (in the case above this should be `+cuda`, however, if we swapped `+cuda`  and `^openblas` it  should be the latter - hence the suggested change in the PR doesn't achieve that yet).

> Doesn't +cuda mean you specifically want cublas?

Yes, more precisely - cublasXt.

haampie:
In that case you can do `depends_on("blas", "~cuda ~rocm")`, right? Depend on CPU blas if not GPU blas.

loveshack:
> It doesn't make sense to specify more than one BLAS backend for
> COSMA. Only one backend is supported per build.

Sure, but that's not what I was trying to do.

> We can either 1) decide to emit an error message or a warning if more
> than one BLAS backend is specified or 2) decide on consistent behavior
> - e.g. select the first one on the spec like (in the case above this
> should be `+cuda`, however, if we swapped `+cuda` and `^openblas` it
> should be the latter - hence the suggested change in the PR doesn't
> achieve that).

Experimentally, the change does what I want, and I don't see how it
would cause a problem.  I don't see how you'd expect not to use cublas
when you've specified +cuda, even if openblas was listed as a dependency
of scalapack that you're also building in the install command.  I don't
understand all the complexity of spack, but is doing that somehow wrong,
rather than having to build scalapack separately?

PR #33758, opened 2022-11-08
ffmpeg: adding version 5.1.2

snehring:
new version

avresample was removed in 5.0

### DIFF
--- a/var/spack/repos/builtin/packages/ffmpeg/package.py
+++ b/var/spack/repos/builtin/packages/ffmpeg/package.py
@@ -44,30 +44,32 @@ class Ffmpeg(AutotoolsPackage):
     # spack packages.
 
     # meta variants: These will toggle several settings
-    variant("X", default=False, description="X11 support")
+    variant("X", default=False, when="@2.4:", description="X11 support")
     variant("drawtext", default=False, description="drawtext filter")
 
     # options
     variant("bzlib", default=True, description="bzip2 support")
-    variant("libaom", default=False, description="AV1 video encoding/decoding")
+    variant("libaom", default=False, when="@4.0:", description="AV1 video encoding/decoding")
     variant("libmp3lame", default=False, description="MP3 encoding")
     variant("libopenjpeg", default=False, description="JPEG 2000 de/encoding")
     variant("libopus", default=False, description="Opus de/encoding")
-    variant("libsnappy", default=False, description="Snappy compression, needed for hap encoding")
+    variant(
+        "libsnappy",
+        default=False,
+        when="@2.7:",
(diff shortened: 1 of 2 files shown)


snehring:
We'll see if that satisfies the py-torchvision builds

var/spack/repos/builtin/packages/ffmpeg/package.py
```diff
@@ -15,6 +15,7 @@ class Ffmpeg(AutotoolsPackage):
 
     maintainers = ["xjrc"]
 
+    version("5.1.2", sha256="39a0bcc8d98549f16c570624678246a6ac736c066cebdb409f9502e915b22f2b")
```
tldahlgren, line 18: Confirmed sha256.

var/spack/repos/builtin/packages/ffmpeg/package.py
```diff
@@ -173,13 +175,14 @@ def configure_args(self):
...
             "libx264",
-            "avresample",
             "nonfree",
             "openssl",
             "shared",
             "version3",
         ]
 
+        if spec.satisfies("@:4.4"):
+            variant_opts.append("avresample")
```
tldahlgren: Spack supports conditional variants (https://spack.readthedocs.io/en/latest/packaging_guide.html#conditional-variants) that are likely more readable (IMHO) and can be used with the `enable_or_disable` helper (https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#enable-or-disable).

snehring: I agree, but I wasn't feeling up to the task today of reworking the variant logic on this when it was otherwise working. I do think it needs some attention though.

I also suspect the libaom and xml2 conflicts aren't relevant any more.

tldahlgren:
Deferring merge decision to give @xjrc a chance to comment.

snehring:
So, I am working on getting rid of those conflicts by making the variants conditional, but there's some weirdness in here. For instance there's a conflict for lzma, but lzma is still relevant per the 5.1.2 configure, likewise the libwebp variant conflicts when 2.1: despite that still being a valid option. I don't know what the intention is here.

snehring:
I tested this new version and the last version with all valid variants and it built successfully.

tldahlgren (approved):
Re-confirmed sha256, rest LGTM.  Am still hoping to see @xjrc review this PR.

bernhardkaindl (approved):
LGTM, @xjrc didn't review within 17 days,
but it looks good for tldahlgren, as well, so we are two.

PR #33768, opened 2022-11-08
Include concrete environments with `include_concrete`

### LINKED ISSUE #33313: Combine Environments Under One View
### Summary

Add the capability to combine any number of environments together under one view.

### Rationale

_No response_

### Description

Entry point
Check for conflicts in existing environments
Print new combined view
Entry point on the concretize command

Example: 
`spack env combine concretize --name current env1 env2 ... env6 eConf`
`spack -e current install`
generate: current/spack.yaml & current/spack.lock

### Additional information

* **Spack:** 0.19.0.dev0
* **Python:** 3.7.2
* **Concretizer:** clingo

### General information

RikkiButler20:
Resolves [#33313](https://github.com/spack/spack/issues/33313)

Add the ability to include any number of (potentially nested) concrete environments, e.g.:

```yaml
   spack:
     specs: []
     concretizer:
         unify: true
     include_concrete:
     - /path/to/environment1
     - /path/to/environment2
```

or, from the CLI:

```console
   $ spack env create myenv
   $ spack -e myenv add python
   $ spack -e myenv concretize
   $ spack env create --include-concrete myenv included_env
```

The contents of included concrete environments' spack.lock files are
included in the environment's lock file at creation time. Any changes
to included concrete environments are only reflected after the environment
is re-concretized from the re-concretized included environments.

- [x] Concretize included envs
- [x] Save concrete specs in memory by hash
- [x] Add included envs to combined env's lock file
- [x] Add test
- [x] Update documentation

### DIFF
--- a/lib/spack/docs/environments.rst
+++ b/lib/spack/docs/environments.rst
@@ -460,6 +460,125 @@ Sourcing that file in Bash will make the environment available to the
 user; and can be included in ``.bashrc`` files, etc.  The ``loads``
 file may also be copied out of the environment, renamed, etc.
 
+
+.. _environment_include_concrete:
+
+------------------------------
+Included Concrete Environments
+------------------------------
+
+Spack environments can create an environment based off of information in already
+established environments. You can think of it as a combination of existing
+environments. It will gather information from the existing environment's
+``spack.lock`` and use that during the creation of this included concrete
+environment. When an included concrete environment is created it will generate
+a ``spack.lock`` file for the newly created environment.
+
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Creating included environments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To create a combined concrete environment, you must have at least one existing
+concrete environment. You will use the command ``spack env create`` with the
+argument ``--include-concrete`` followed by the name or path of the environment
+you'd like to include. Here is an example of how to create a combined environment
+from the command line.
+
+.. code-block:: console
+
+   $ spack env create myenv
+   $ spack -e myenv add python
+   $ spack -e myenv concretize
+   $ spack env create --include-concrete myenv included_env
+
+
+You can also include an environment directly in the ``spack.yaml`` file. It
+involves adding the ``include_concrete`` heading in the yaml followed by the
+absolute path to the independent environments.
+
+.. code-block:: yaml
+
+   spack:
+     specs: []
+     concretizer:
+         unify: true
+     include_concrete:
+     - /absolute/path/to/environment1
+     - /absolute/path/to/environment2
+
+
+Once the ``spack.yaml`` has been updated you must concretize the environment to
+get the concrete specs from the included environments.
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Updating an included environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+If changes were made to the base environment and you want that reflected in the
+included environment you will need to reconcretize both the base environment and the
+included environment for the change to be implemented. For example:
+
+.. code-block:: console
+
+   $ spack env create myenv
+   $ spack -e myenv add python
+   $ spack -e myenv concretize
+   $ spack env create --include-concrete myenv included_env
+
+
+   $ spack -e myenv find
+   ==> In environment myenv
+   ==> Root specs
+   python
+
+   ==> 0 installed packages
+
+
+   $ spack -e included_env find
+   ==> In environment included_env
+   ==> No root specs
+   ==> Included specs
+   python
+
+   ==> 0 installed packages
+
+Here we see that ``included_env`` has access to the python package through
+the ``myenv`` environment. But if we were to add another spec to ``myenv``,
+``included_env`` will not be able to access the new information.
+
+.. code-block:: console
+
+   $ spack -e myenv add perl
+   $ spack -e myenv concretize
+   $ spack -e myenv find
+   ==> In environment myenv
+   ==> Root specs
+   perl  python
+
+   ==> 0 installed packages
+
+
+   $ spack -e included_env find
+   ==> In environment included_env
+   ==> No root specs
+   ==> Included specs
+   python
+
+   ==> 0 installed packages
+
+It isn't until you run the ``spack concretize`` command that the combined
+environment will get the updated information from the reconcretized base environmennt.
+
+.. code-block:: console
+
+   $ spack -e included_env concretize
+   $ spack -e included_env find
+   ==> In environment included_env
+   ==> No root specs
+   ==> Included specs
+   perl  python
+
+   ==> 0 installed packages
+
 .. _environment-configuration:
 
 ------------------------
@@ -811,6 +930,7 @@ For example, the following environment has three root packages:
 This allows for a much-needed reduction in redundancy between packages
 and constraints.
 
+
 ----------------
 Filesystem Views
 ----------------
@@ -1044,7 +1164,7 @@ other targets to depend on the environment installation.
 
 A typical workflow is as follows:
 
-.. code:: console
+.. code-block:: console
 
    spack env create -d .
    spack -e . add perl
@@ -1137,7 +1257,7 @@ its dependencies. This can be useful when certain flags should only apply to
 dependencies. Below we show a use case where a spec is installed with verbose
 output (``spack install --verbose``) while its dependencies are installed silently:
 
-.. code:: console
+.. code-block:: console
 
    $ spack env depfile -o Makefile
 
@@ -1159,7 +1279,7 @@ This can be accomplished through the generated ``[<prefix>/]SPACK_PACKAGE_IDS``
 variable. Assuming we have an active and concrete environment, we generate the
 associated ``Makefile`` with a prefix ``example``:
 
-.. code:: console
+.. code-block:: console
 
    $ spack env depfile -o env.mk --make-prefix example
 
--- a/lib/spack/spack/cmd/env.py
+++ b/lib/spack/spack/cmd/env.py
@@ -10,7 +10,7 @@ import shutil
 import sys
 import tempfile
 from pathlib import Path
-from typing import Optional
+from typing import List, Optional
 
 import llnl.string as string
 import llnl.util.filesystem as fs
@@ -87,6 +87,9 @@ def env_create_setup_parser(subparser):
         default=None,
         help="either a lockfile (must end with '.json' or '.lock') or a manifest file",
     )
+    subparser.add_argument(
+        "--include-concrete", action="append", help="name of old environment to copy specs from"
+    )
 
 
 def env_create(args):
@@ -104,12 +107,17 @@ def env_create(args):
         # the environment should not include a view.
         with_view = None
 
+    include_concrete = None
+    if hasattr(args, "include_concrete"):
+        include_concrete = args.include_concrete
+
     env = _env_create(
         args.env_name,
         init_file=args.envfile,
         dir=args.dir or os.path.sep in args.env_name or args.env_name in (".", ".."),
         with_view=with_view,
         keep_relative=args.keep_relative,
+        include_concrete=include_concrete,
     )
 
     # Generate views, only really useful for environments created from spack.lock files.
@@ -123,31 +131,43 @@ def _env_create(
     dir: bool = False,
     with_view: Optional[str] = None,
     keep_relative: bool = False,
+    include_concrete: Optional[List[str]] = None,
 ):
     """Create a new environment, with an optional yaml description.
 
     Arguments:
-        name_or_path: name of the environment to create, or path to it
-        init_file: optional initialization file -- can be a JSON lockfile (*.lock, *.json) or YAML
-            manifest file
-        dir: if True, create an environment in a directory instead of a managed environment
-        keep_relative: if True, develop paths are copied verbatim into the new environment file,
-            otherwise they may be made absolute if the new environment is in a different location
+        name_or_path (str): name of the environment to create, or path to it
+        init_file (str or file): optional initialization file -- can be
+            a JSON lockfile (*.lock, *.json) or YAML manifest file
+        dir (bool): if True, create an environment in a directory instead
+            of a named environment
+        keep_relative (bool): if True, develop paths are copied verbatim into
+            the new environment file, otherwise they may be made absolute if the
+            new environment is in a different location
+        include_concrete (list): list of the included concrete environments
     """
     if not dir:
         env = ev.create(
-            name_or_path, init_file=init_file, with_view=with_view, keep_relative=keep_relative
+            name_or_path,
+            init_file=init_file,
+            with_view=with_view,
+            keep_relative=keep_relative,
+            include_concrete=include_concrete,
         )
         tty.msg(
             colorize(
-                f"Created environment @c{{{cescape(env.name)}}} in: @c{{{cescape(env.path)}}}"
+                f"Created environment @c{{{cescape(name_or_path)}}} in: @c{{{cescape(env.path)}}}"
             )
         )
     else:
         env = ev.create_in_dir(
-            name_or_path, init_file=init_file, with_view=with_view, keep_relative=keep_relative
+            name_or_path,
+            init_file=init_file,
+            with_view=with_view,
+            keep_relative=keep_relative,
+            include_concrete=include_concrete,
         )
-        tty.msg(colorize(f"Created anonymous environment in: @c{{{cescape(env.path)}}}"))
+        tty.msg(colorize(f"Created independent environment in: @c{{{cescape(env.path)}}}"))
     tty.msg(f"Activate with: {colorize(f'@c{{spack env activate {cescape(name_or_path)}}}')}")
     return env
 
@@ -434,6 +454,12 @@ def env_remove_setup_parser(subparser):
     """remove an existing environment"""
     subparser.add_argument("rm_env", metavar="env", nargs="+", help="environment(s) to remove")
     arguments.add_common_arguments(subparser, ["yes_to_all"])
+    subparser.add_argument(
+        "-f",
+        "--force",
+        action="store_true",
+        help="remove the environment even if it is included in another environment",
+    )
 
 
 def env_remove(args):
@@ -443,13 +469,35 @@ def env_remove(args):
     and manifests embedded in repositories should be removed manually.
     """
     read_envs = []
+    valid_envs = []
     bad_envs = []
-    for env_name in args.rm_env:
+    invalid_envs = []
+
+    for env_name in ev.all_environment_names():
         try:
             env = ev.read(env_name)
-            read_envs.append(env)
+            valid_envs.append(env_name)
+
+            if env_name in args.rm_env:
+                read_envs.append(env)
         except (spack.config.ConfigFormatError, ev.SpackEnvironmentConfigError):
-            bad_envs.append(env_name)
+            invalid_envs.append(env_name)
+
+            if env_name in args.rm_env:
+                bad_envs.append(env_name)
+
+        # Check if env is linked to another before trying to remove
+        for name in valid_envs:
+            # don't check if environment is included to itself
+            if name == env_name:
+                continue
+            environ = ev.Environment(ev.root(name))
+            if ev.root(env_name) in environ.included_concrete_envs:
+                msg = f'Environment "{env_name}" is being used by environment "{name}"'
+                if args.force:
+                    tty.warn(msg)
+                else:
+                    tty.die(msg)
 
     if not args.yes_to_all:
         environments = string.plural(len(args.rm_env), "environment", show_n=False)
--- a/lib/spack/spack/environment/__init__.py
+++ b/lib/spack/spack/environment/__init__.py
@@ -34,6 +34,9 @@ contents have.  Lockfiles are JSON-formatted and their top-level sections are:
       * ``spec``: a string representation of the abstract spec that was concretized
 
   4. ``concrete_specs``: a dictionary containing the specs in the environment.
+  5. ``include_concrete`` (dictionary): an optional dictionary that includes the roots
+     and concrete specs from the included environments, keyed by the path to that
+     environment
 
 Compatibility
 -------------
@@ -50,8 +53,10 @@ upgrade Spack to use them.
      - ``v2``
      - ``v3``
      - ``v4``
+     - ``v5``
    * - ``v0.12:0.14``
      - ✅
+     -
      -
      -
      -
@@ -60,12 +65,21 @@ upgrade Spack to use them.
      - ✅
      -
      -
+     -
    * - ``v0.17``
      - ✅
      - ✅
      - ✅
      -
+     -
    * - ``v0.18:``
+     - ✅
+     - ✅
+     - ✅
+     - ✅
+     -
+   * - ``v0.22:``
+     - ✅
      - ✅
      - ✅
      - ✅
@@ -331,6 +345,118 @@ the commit or version.
                     }
                 ],
                 "hash": "<dag_hash 2>"
+            }
+        }
+    }
+
+
+Version 5
+---------
+
+Version 5 doesn't change the top-level lockfile format, but an optional dictionary is
+added. The dictionary has the ``root`` and ``concrete_specs`` of the included
+environments, which are keyed by the path to that environment. Since this is optional
+if the environment does not have any included environments ``include_concrete`` will
+not be a part of the lockfile.
+
+.. code-block:: json
+
+    {
+        "_meta": {
+            "file-type": "spack-lockfile",
+            "lockfile-version": 5,
+            "specfile-version": 3
+        },
+        "roots": [
+            {
+                "hash": "<dag_hash 1>",
+                "spec": "<abstract spec 1>"
+            },
+            {
+                "hash": "<dag_hash 2>",
+                "spec": "<abstract spec 2>"
+            }
+        ],
+        "concrete_specs": {
+            "<dag_hash 1>": {
+                "... <spec dict attributes> ...": { },
+                "dependencies": [
+                    {
+                        "name": "depname_1",
+                        "hash": "<dag_hash for depname_1>",
+                        "type": ["build", "link"]
+                    },
+                    {
+                        "name": "depname_2",
+                        "hash": "<dag_hash for depname_2>",
+                        "type": ["build", "link"]
+                    }
+                ],
+                "hash": "<dag_hash 1>",
+            },
+            "<daghash 2>": {
+                "... <spec dict attributes> ...": { },
+                "dependencies": [
+                    {
+                        "name": "depname_3",
+                        "hash": "<dag_hash for depname_3>",
+                        "type": ["build", "link"]
+                    },
+                    {
+                        "name": "depname_4",
+                        "hash": "<dag_hash for depname_4>",
+                        "type": ["build", "link"]
+                    }
+                ],
+                "hash": "<dag_hash 2>"
+            }
+        }
+        "include_concrete": {
+            "<path to environment>": {
+                "roots": [
+                    {
+                        "hash": "<dag_hash 1>",
+                        "spec": "<abstract spec 1>"
+                    },
+                    {
+                        "hash": "<dag_hash 2>",
+                        "spec": "<abstract spec 2>"
+                    }
+                ],
+                "concrete_specs": {
+                    "<dag_hash 1>": {
+                        "... <spec dict attributes> ...": { },
+                        "dependencies": [
+                            {
+                                "name": "depname_1",
+                                "hash": "<dag_hash for depname_1>",
+                                "type": ["build", "link"]
+                            },
+                            {
+                                "name": "depname_2",
+                                "hash": "<dag_hash for depname_2>",
+                                "type": ["build", "link"]
+                            }
+                        ],
+                        "hash": "<dag_hash 1>",
+                    },
+                    "<daghash 2>": {
+                        "... <spec dict attributes> ...": { },
+                        "dependencies": [
+                            {
+                                "name": "depname_3",
+                                "hash": "<dag_hash for depname_3>",
+                                "type": ["build", "link"]
+                            },
+                            {
+                                "name": "depname_4",
+                                "hash": "<dag_hash for depname_4>",
+                                "type": ["build", "link"]
+                            }
+                        ],
+                        "hash": "<dag_hash 2>"
+                    }
+                }
             }
         }
     }
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -16,7 +16,7 @@ import time
 import urllib.parse
 import urllib.request
 import warnings
-from typing import Dict, Iterable, List, Optional, Set, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Union
 
 import llnl.util.filesystem as fs
 import llnl.util.tty as tty
@@ -159,6 +159,8 @@ user_speclist_name = "specs"
 default_view_name = "default"
 # Default behavior to link all packages into views (vs. only root packages)
 default_view_link = "all"
+# The name for any included concrete specs
+included_concrete_name = "include_concrete"
 
 
 def installed_specs():
@@ -293,6 +295,7 @@ def create(
     init_file: Optional[Union[str, pathlib.Path]] = None,
     with_view: Optional[Union[str, pathlib.Path, bool]] = None,
     keep_relative: bool = False,
+    include_concrete: Optional[List[str]] = None,
 ) -> "Environment":
     """Create a managed environment in Spack and returns it.
 
@@ -309,10 +312,15 @@ def create(
             string, it specifies the path to the view
         keep_relative: if True, develop paths are copied verbatim into the new environment file,
             otherwise they are made absolute
+        include_concrete: list of concrete environment names/paths to be included
     """
     environment_dir = environment_dir_from_name(name, exists_ok=False)
     return create_in_dir(
-        environment_dir, init_file=init_file, with_view=with_view, keep_relative=keep_relative
+        environment_dir,
+        init_file=init_file,
+        with_view=with_view,
+        keep_relative=keep_relative,
+        include_concrete=include_concrete,
     )
 
 
@@ -321,6 +329,7 @@ def create_in_dir(
     init_file: Optional[Union[str, pathlib.Path]] = None,
     with_view: Optional[Union[str, pathlib.Path, bool]] = None,
     keep_relative: bool = False,
+    include_concrete: Optional[List[str]] = None,
 ) -> "Environment":
     """Create an environment in the directory passed as input and returns it.
 
@@ -334,6 +343,7 @@ def create_in_dir(
             string, it specifies the path to the view
         keep_relative: if True, develop paths are copied verbatim into the new environment file,
             otherwise they are made absolute
+        include_concrete: concrete environment names/paths to be included
     """
     initialize_environment_dir(root, envfile=init_file)
 
@@ -345,6 +355,12 @@ def create_in_dir(
 
         if with_view is not None:
             manifest.set_default_view(with_view)
+
+        if include_concrete is not None:
+            set_included_envs_to_env_paths(include_concrete)
+            validate_included_envs_exists(include_concrete)
+            validate_included_envs_concrete(include_concrete)
+            manifest.set_include_concrete(include_concrete)
 
         manifest.flush()
 
@@ -366,6 +382,14 @@ def create_in_dir(
                 _rewrite_relative_dev_paths_on_relocation(env, init_file_dir)
 
     return env
+
+    # Must be done after environment is initialized
+    if include_concrete:
+        manifest.set_include_concrete(include_concrete)
+
+    if not keep_relative and init_file is not None and str(init_file).endswith(manifest_name):
+        init_file = pathlib.Path(init_file)
+        manifest.absolutify_dev_paths(init_file.parent)
 
 
 def _rewrite_relative_dev_paths_on_relocation(env, init_file_dir):
@@ -417,6 +441,67 @@ def environment_dir_from_name(name: str, exists_ok: bool = True) -> str:
 def ensure_env_root_path_exists():
     if not os.path.isdir(env_root_path()):
         fs.mkdirp(env_root_path())
+
+
+def set_included_envs_to_env_paths(include_concrete: List[str]) -> None:
+    """If the included environment(s) is the environment name
+    it is replaced by the path to the environment
+
+    Args:
+        include_concrete: list of env name or path to env"""
+
+    for i, env_name in enumerate(include_concrete):
+        if is_env_dir(env_name):
+            include_concrete[i] = env_name
+        elif exists(env_name):
+            include_concrete[i] = root(env_name)
+
+
+def validate_included_envs_exists(include_concrete: List[str]) -> None:
+    """Checks that all of the included environments exist
+
+    Args:
+       include_concrete: list of already existing concrete environments to include
+
+    Raises:
+        SpackEnvironmentError: if any of the included environments do not exist
+    """
+
+    missing_envs = set()
+
+    for i, env_name in enumerate(include_concrete):
+        if not is_env_dir(env_name):
+            missing_envs.add(env_name)
+
+    if missing_envs:
+        msg = "The following environment(s) are missing: {0}".format(", ".join(missing_envs))
+        raise SpackEnvironmentError(msg)
+
+
+def validate_included_envs_concrete(include_concrete: List[str]) -> None:
+    """Checks that all of the included environments are concrete
+
+    Args:
+        include_concrete: list of already existing concrete environments to include
+
+    Raises:
+        SpackEnvironmentError: if any of the included environments are not concrete
+    """
+
+    non_concrete_envs = set()
+
+    for env_path in include_concrete:
+        if not os.path.exists(Environment(env_path).lock_path):
+            non_concrete_envs.add(Environment(env_path).name)
+
+    if non_concrete_envs:
+        msg = "The following environment(s) are not concrete: {0}\n" "Please run:".format(
+            ", ".join(non_concrete_envs)
+        )
+        for env in non_concrete_envs:
+            msg += f"\n\t`spack -e {env} concretize`"
+
+        raise SpackEnvironmentError(msg)
 
 
 def all_environment_names():
@@ -821,6 +906,18 @@ class Environment:
         self.specs_by_hash: Dict[str, Spec] = {}
         #: Repository for this environment (memoized)
         self._repo = None
+
+        #: Environment paths for concrete (lockfile) included environments
+        self.included_concrete_envs: List[str] = []
+        #: First-level included concretized spec data from/to the lockfile.
+        self.included_concrete_spec_data: Dict[str, Dict[str, List[str]]] = {}
+        #: User specs from included environments from the last concretization
+        self.included_concretized_user_specs: Dict[str, List[Spec]] = {}
+        #: Roots from included environments with the last concretization, in order
+        self.included_concretized_order: Dict[str, List[str]] = {}
+        #: Concretized specs by hash from the included environments
+        self.included_specs_by_hash: Dict[str, Dict[str, Spec]] = {}
+
         #: Previously active environment
         self._previous_active = None
         self._dev_specs = None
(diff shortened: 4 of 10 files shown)


RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
spack style --fix

RikkiButler20:
@spackbot spack style --fix

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

lib/spack/spack/environment/environment.py
```diff
@@ -270,14 +272,14 @@ def read(name):
     return Environment(root(name))
 
 
-def create(name, init_file=None, with_view=None, keep_relative=False):
+def create(name, init_file=None, with_view=None, keep_relative=False, include_concrete=None):
```
alalazo: @becker33 With respect to what we were discussing yesterday, in #36927 this function:

https://github.com/spack/spack/blob/bd86ad12b3353f436224e4c427d0c1017d500405/lib/spack/spack/environment/environment.py#L275-L297

delegates to another one:

https://github.com/spack/spack/blob/bd86ad12b3353f436224e4c427d0c1017d500405/lib/spack/spack/environment/environment.py#L2464-L2521

to initialize the directory for the environment. When merging this feature PR and that refactor, I think any modification to get a consistent `spack.yaml` should be done in the `initialize_environment_dir` function I linked above.

lib/spack/spack/environment/environment.py
```diff
@@ -876,6 +912,8 @@ def clear(self, re_read=False):
         self.concretized_user_specs = []  # user specs from last concretize
         self.concretized_order = []  # roots of last concretize, in order
         self.specs_by_hash = {}  # concretized specs by hash
+        self.included_concretized_user_specs = []  # user specs from last concretize's included env
+        self.included_concretized_order = []  # roots of the included envs, in order
```
alalazo: I think this makes the `Environment` class even less coherent than it is now, in the sense that it will have more attributes and will use just a few of them in most methods. This is usually an indicator that the class has too many responsibilities.

Also, these attributes are set outside of `__init__`. To be fair, they're not the first ones like that for `Environment`, but in #36927 I made sure to declare every attribute of an environment in the constructor.

alalazo (commented):
I haven't been looking closely at this PR before, so sorry if I miss context. Is there any example of how we would use this feature once implemented? What is the synchronization policy between this environment and the included one (if e.g. the included environment changes independently)?

RikkiButler20:
@spackbot fix style

lib/spack/spack/test/cmd/env.py
```diff
@@ -706,9 +753,8 @@ def test_with_config_bad_include():
             e.concretize()
 
     err = str(exc)
-    assert "missing include" in err
-    assert "/no/such/directory" in err
-    assert os.path.join("no", "such", "file.yaml") in err
+
+    assert "2 missing include" in err
```
becker33: I'm having trouble following why this one had to change

lib/spack/spack/test/cmd/env.py
```diff
@@ -1255,6 +1301,108 @@ def test_env_without_view_install(tmpdir, mock_stage, mock_fetch, install_mocker
...
+
+
+def test_env_include_concrete_envs_lockfile():
+    test1, test2, combined = setup_combined_multiple_env()
+
+    with open(combined.lock_path) as f:
+        lockfile_as_dict = combined._read_lockfile(f)
+
+    assert lockfile_as_dict["include"][test1.path]["roots"][0]["hash"] in test1.specs_by_hash
+    assert lockfile_as_dict["include"][test2.path]["roots"][0]["hash"] in test2.specs_by_hash
```
becker33: Could we make these checks a little stricter by comparing 

```suggestion
set(entry["hash"] for entry in lockfile_as_dict["include"][test1.path]["roots"]) == set(test1.specs_by_hash)
set(entry["hash"] for entry in lockfile_as_dict["include"][test2.path]["roots"]) == set(test2.specs_by_hash)

File: lib/spack/spack/test/cmd/env.py [hunk 5]
```diff
@@ -1255,6 +1301,108 @@ def test_env_without_view_install(tmpdir, mock_stage, mock_fetch, install_mocker
...
+
+    combined.concretize()
+    combined = ev.read("combined_env")
+    with open(combined.lock_path) as f:
+        lockfile_as_dict = combined._read_lockfile(f)
+
+    assert not lockfile_as_dict["roots"]
+    assert not lockfile_as_dict["concrete_specs"]
+
+
```
becker33, line 1865: Can you add another test for reconcretizing the included environment and making sure the combined environment 

1. Doesn't show the change before reconcretizing
2. Does show the change after reconcretizing

It looks like the code supports this, but we should have a test to make sure we don't regress on it later.

lib/spack/spack/test/cmd/find.py
```diff
@@ -347,6 +347,44 @@ def test_find_prefix_in_env(
...
+        env("create", "test2", "spack.yaml")
+
+    env("create", "--include-concrete", "test1", "--include-concrete", "test2", "combined_env")
+
+    with ev.read("combined_env"):
+        output = find()
+
+    assert "No root specs" in output
+    assert "mpileaks" in output
+    assert "libelf" in output
```
becker33: Can you have this test show/check the included roots specifically? I think this could be satisfied just be the installed specs section

lib/spack/spack/environment/environment.py
```diff
@@ -709,7 +713,7 @@ def __init__(self, path, init_file=None, with_view=None, keep_relative=False):
             with fs.open_if_filename(init_file) as f:
                 if hasattr(f, "name") and f.name.endswith(".lock"):
                     self._read_manifest(default_manifest_yaml())
-                    self._read_lockfile(f)
+                    self._read_lockfile(f)["_meta"]["lockfile-version"]
```
becker33: This looks vestigial -- should there be a second line here that checks the lockfile version? I think you should be able to just undo this one-line change.

lib/spack/spack/environment/environment.py
```diff
@@ -876,6 +924,10 @@ def clear(self, re_read=False):
         self.concretized_user_specs = []  # user specs from last concretize
         self.concretized_order = []  # roots of last concretize, in order
         self.specs_by_hash = {}  # concretized specs by hash
+        self.included_concretized_user_specs = {}  # user specs from last concretize's included env
+        # TODO: Rikki, included_specs_by_hash keyed by env path for debugging
```
becker33: Is this TODO still active?

lib/spack/spack/environment/environment.py
```diff
@@ -956,6 +1008,54 @@ def repo(self):
...
+            env.write()
+
+            with open(env.lock_path) as f:
+                lockfile_as_dict = env._read_lockfile(f)
+
+            # Lockfile_meta must match each env
+            if lockfile_meta is None:
+                lockfile_meta = lockfile_as_dict["_meta"]
+            elif lockfile_meta != lockfile_as_dict["_meta"]:
+                tty.die("All lockfile _meta values must match")
```
becker33: Can you write a more descriptive error here? The real issue is that one of the environments is using an outdated lockfile format, right? Then you could also prompt the user to reconcretize that environment if they want to go forward.

becker33 (changes requested):
This PR also needs to bump the lockfile version to 5, since it modifies the lockfile format. We can still read the old format, but we need the new version to show older Spack instances that they can't properly read the new ones, otherwise environments will be read incorrectly and miss the new information instead of throwing an error.

lib/spack/spack/environment/environment.py
```diff
@@ -956,6 +1008,54 @@ def repo(self):
             self._repo = make_repo_path(self.repos_path)
         return self._repo
 
+    def include_concrete_envs(self):
+        """Concretize included envs and save the included envs'
+        specs internally"""
+
+        lockfile_meta = None
+        self.included_concrete_specs = dict()
+        root_hash = set()
```
becker33: Can you rename this to `root_hash_seen`?

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
When I try reconcretizing the base env with a new spec it still shows up in the included env. And the test for this isn't failing.

RikkiButler20:
@spackbot fix style

lib/spack/docs/environments.rst
```diff
@@ -417,6 +417,101 @@ Sourcing that file in Bash will make the environment available to the
...
+------------------------------
+Included Concrete Environments
+------------------------------
+
+Spack environments can create an evironment based off of information in already
+established environments. You can think of it as a combination of existing
+environnments. It will gathering information from the existing environment's
+``spack.lock`` and use that during the creation of this included concrete
+environmenit. When an included concrete environment is create it will generate
+a ``spack.lock`` file.
```
tldahlgren: There are typos (e.g., `evironment`,   `environmenit`) and awkward wording (e.g., `will gathering`, `is create it`).  

Try to use more active voice.  It's also worth mentioning that the included environments need to be concrete and that the generated `spack.lock` file is for the new environment.

lib/spack/docs/environments.rst
```diff
@@ -417,6 +417,101 @@ Sourcing that file in Bash will make the environment available to the
...
+
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Creating included environment
```
tldahlgren:
```suggestion
Creating included environments
```

lib/spack/docs/environments.rst
```diff
@@ -417,6 +417,101 @@ Sourcing that file in Bash will make the environment available to the
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To create an included concrete environment, you must have at least one exsiting
```
tldahlgren: I think you mean something along the lines of creating a combined environment (versus included) requires at least one existing concrete environment.

lib/spack/docs/environments.rst
```diff
@@ -417,6 +417,101 @@ Sourcing that file in Bash will make the environment available to the
...
+concrete environment. You will use the command ``spack env create`` with the
+argument ``--include-concrete`` followed by the name or path of the environment
+you'd like to include.
+
+.. code:: console
```
tldahlgren: IMO It's worth introducing that this is an example and explaining what it is doing.

lib/spack/docs/environments.rst
```diff
@@ -417,6 +417,101 @@ Sourcing that file in Bash will make the environment available to the
...
+
+   $ spack env create myenv
+   $ spack -e myenv add python
+   $ spack -e myenv concretize
+   $ spack env create --include-concrete myenv included_env
+
+
+You can also create an included environment by adding the ``include_concrete``
+heading in the yaml.
```
tldahlgren: It would be clearer if you mention up front that this (i.e., creating the combined environment's `spack.yaml` file) involves adding the `include_concrete` element with a list of environment names (for managed environments) and or paths (for independent environments).

lib/spack/docs/environments.rst
```diff
@@ -417,6 +417,101 @@ Sourcing that file in Bash will make the environment available to the
...
+   ==> Root specs
+   perl  python
+
+   ==> 0 installed packages
+
+
+   $ spack -e included_env find
+   ==> In environment included_env
+   ==> No root specs
+   ==> Included specs
```
tldahlgren, line 563: I think you're missing the last line (see line 495) here.

tldahlgren: IMO It would also be good to explain what is happening in these examples.

lib/spack/docs/environments.rst
```diff
@@ -768,6 +863,66 @@ For example, the following environment has three root packages:
...
+
+   $ spack -e included_env concretize
+   $ spack -e included_env find
+   ==> In environment included_env
+   ==> No root specs
+   ==> Included specs
+   perl  python
+
+   ==> 0 installed packages
+
```
tldahlgren: This looks like a more complete example of the subsection starting at line 468.

lib/spack/spack/cmd/env.py
```diff
@@ -349,6 +372,9 @@ def env_remove_setup_parser(subparser):
     """remove an existing environment"""
     subparser.add_argument("rm_env", metavar="env", nargs="+", help="environment(s) to remove")
     arguments.add_common_arguments(subparser, ["yes_to_all"])
+    subparser.add_argument(
+        "-f", "--force", action="store_true", help="forcfully remove environment"
```
tldahlgren: "forcfully" -> "forcefully"

lib/spack/spack/cmd/env.py
```diff
@@ -362,6 +388,15 @@ def env_remove(args):
         env = ev.read(env_name)
         read_envs.append(env)
 
+    # Check if env is linked to another before trying to remove
+    for name in ev.all_environment_names():
+        eviron = ev.Environment(ev.root(name))
```
tldahlgren: Suggest using `environ` instead of `eviron` unless doing so would create a duplicate.

lib/spack/spack/cmd/env.py
```diff
@@ -362,6 +388,15 @@ def env_remove(args):
...
+        if ev.root(env_name) in eviron.include_concrete:
+            if args.force:
+                tty.warn('Environment "%s" is being used by environment "%s"' % (env_name, name))
```
tldahlgren: Suggest switching to the new f-strings wherever
you're adding code since that has come up as what we're supposed to use in recent meetings.

lib/spack/spack/environment/__init__.py
```diff
@@ -34,6 +34,10 @@
       * ``spec``: a string representation of the abstract spec that was concretized
 
   4. ``concrete_specs``: a dictionary containing the specs in the environment.
+  5. ``include_concrete`` (dictionary): an optional dictionary that includes the roots
+     and concrete specs from the included environments, keyed by the path to that
```
tldahlgren, line 38: Is it always the path or can it be the name if the included environment is managed by Spack?

lib/spack/spack/environment/__init__.py
```diff
@@ -50,26 +54,31 @@
...
      - ✅
      - ✅
      -
+     -
    * - ``v0.18:``
      - ✅
      - ✅
      - ✅
      - ✅
+     -
```
tldahlgren, line 87: You want to add another row here with the row "heading" of the release in which this PR will appear.  You could put ``v0.21:`` as a placeholder.  I believe that also means the version range at line 76 needs to end a ``v0.20``.

lib/spack/spack/environment/__init__.py
```diff
@@ -334,6 +343,118 @@
...
         }
     }
+
+
+Version 5
+---------
+
+Version 5 doesn't change the top-level lockfile format, but an optional dictionary is
+added. The dictionary has the ``root`` and ``concrete_specs`` of the included
+environments and they are keyed by the path to that environment. Since this is optional
```
tldahlgren: "and they are keyed" -> ", which are keyed"

Aso, is it only the path or are the names of managed environments supported?

lib/spack/spack/environment/__init__.py
```diff
@@ -334,6 +343,118 @@
...
+                        "name": "depname_4",
+                        "hash": "<dag_hash for depname_4>",
+                        "type": ["build", "link"]
+                    }
+                ],
+                "hash": "<dag_hash 2>"
+            }
+        }
+        "include_concrete": {
+            "<path to environment>": {
```
tldahlgren, line 415: Path or managed environment name?

lib/spack/spack/environment/environment.py
```diff
@@ -361,6 +379,60 @@ def ensure_env_root_path_exists():
...
+        if not os.path.exists(Environment(env_path).lock_path):
+            non_concrete_envs.add(Environment(env_path).name)
+            continue
+
+    if non_concrete_envs:
+        msg = "The following environment(s) are not concrete: {0}\n" "Please run:".format(
+            ", ".join(non_concrete_envs)
+        )
+        for env in non_concrete_envs:
+            msg += f"\n\t`spack -e {env} concretize`"
```
tldahlgren, line 502: 👏 Appreciating use of f-string here.

tldahlgren (changes requested):
I've made a quick pass at this point, mostly over documentation, code, and running unit tests.   Will review the code in more depth, hopefully tomorrow.

tldahlgren (changes requested):
I've made a quick pass at this point, mostly over documentation, code, and running unit tests.   Will review the code in more depth, hopefully tomorrow.

lib/spack/docs/environments.rst
RikkiButler20: Looks like I forgot to remove that when I was moving stuff around 😅

lib/spack/spack/environment/__init__.py
RikkiButler20: I've set it up so the path is always stored in the lock file

lib/spack/spack/environment/__init__.py
RikkiButler20: Path

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

lib/spack/docs/environments.rst
```diff
@@ -424,32 +424,34 @@ file may also be copied out of the environment, renamed, etc.
...
+environment. When an included concrete environment is created it will generate
+a ``spack.lock`` file for the newly created environment.
 
 
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Creating included environment
+Creating included environments
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-To create an included concrete environment, you must have at least one exsiting
+To create a combined concrete environment, you must have at least one exsiting
```
tldahlgren: "exsiting" -> "existing"

lib/spack/docs/environments.rst
```diff
@@ -424,32 +424,34 @@ file may also be copied out of the environment, renamed, etc.
...
 
    $ spack env create myenv
    $ spack -e myenv add python
    $ spack -e myenv concretize
    $ spack env create --include-concrete myenv included_env
 
 
-You can also create an included environment by adding the ``include_concrete``
-heading in the yaml.
+You can also create an included environment through the ``spack.yaml`` file. It
```
tldahlgren: Should this be "create a combined environment using the" or "You can also include an environment directly in the..."?

lib/spack/docs/environments.rst
```diff
@@ -472,7 +474,7 @@ If changes were made to the base environment and you want that reflected in the
 included environment you will need to reconcretize both the base environment and the
 included environment for the change to be implemented. For example:
 
-.. code:: console
+.. code-block:: console
```
tldahlgren, line 520: Great catch!

lib/spack/docs/environments.rst
```diff
@@ -496,6 +498,11 @@ included environment for the change to be implemented. For example:
 
    ==> 0 installed packages
 
+Here we see that included_env has access to the python packge through the myenv
```
tldahlgren: "packge" -> "package"

lib/spack/docs/environments.rst
```diff
@@ -496,6 +498,11 @@ included environment for the change to be implemented. For example:
...
+environment. But if we were to add another spec tp myenv, included_env will not
```
tldahlgren: "tp" -> "to"

lib/spack/spack/environment/__init__.py
```diff
@@ -350,7 +356,7 @@
 
 Version 5 doesn't change the top-level lockfile format, but an optional dictionary is
 added. The dictionary has the ``root`` and ``concrete_specs`` of the included
-environments and they are keyed by the path to that environment. Since this is optional
+environments and which are keyed by the path to that environment. Since this is optional
```
tldahlgren: Maybe "environments and which" -> "environments, which"

tldahlgren (changes requested):
Minor typos and wording corrections at this point while I test the feature.

tldahlgren:
~@becker33 @tgamblin Not sure we’ve discussed this use case (enough), but shouldn’t we be able to include another (previously concretized) environment to an already combined environment?~

~I tried and am getting `==> Error: All lockfile _meta values must match`.  This occurs even if I force re-concretization of the environment I'm trying to add post-creation.~

~So should this PR allow inclusion of any (compatible) lockfile version?  Or should `spack concretize --force` update the lockfile version? Or something else?~

Nvm.  The window I was using to re-concretize the environment I wanted to include had the wrong branch checked out.

tldahlgren:
The behavior I'm seeing manually testing this with included environments seems to work fine.  Testing the combined environments with changes to the local environment seems to behave as expected.

RikkiButler20:
@spackbot fix style

tldahlgren (dismissed):
The docs corrections look fine. Thanks.

lib/spack/spack/cmd/env.py
```diff
@@ -285,6 +285,12 @@ def env_create_setup_parser(subparser):
         default=None,
         help="either a lockfile (must end with '.json' or '.lock') or a manifest file.",
     )
+    subparser.add_argument(
+        "--include-concrete",
+        action="append",
+        help="path/name of environment to link to new environment",
+    )
+    # Check if name provided is a path (try except on root)
```
tldahlgren: Is this a `ToDo` item?  If not, you want to remove the comment.

lib/spack/spack/cmd/env.py
```diff
@@ -359,6 +381,9 @@ def env_remove_setup_parser(subparser):
     """remove an existing environment"""
     subparser.add_argument("rm_env", metavar="env", nargs="+", help="environment(s) to remove")
     arguments.add_common_arguments(subparser, ["yes_to_all"])
+    subparser.add_argument(
+        "-f", "--force", action="store_true", help="forcefully remove environment"
```
becker33:
```suggestion
        "-f", "--force", action="store_true", help="remove the environment even if it is included in another environment"
```

lib/spack/spack/test/cmd/env.py
```diff
@@ -447,6 +470,41 @@ def test_remove_command():
...
+    with test:
+        add("mpileaks")
+
+    test.concretize()
+    test.write()
+
+    env("create", "--include-concrete", "test", "combined_env")
+
+    output = env("remove", "-f", "-y", "test")
+    out = env("list")
```
becker33: Probably easier to read if we name these `rm_output` and `list_output` instead of just the short/long forms.

lib/spack/spack/test/cmd/env.py
```diff
@@ -1274,6 +1332,152 @@ def test_env_without_view_install(tmpdir, mock_stage, mock_fetch, install_mocker
     check_mpileaks_and_deps_in_view(view_dir)
 
 
+def setup_combined_multiple_env():
```
becker33: We should move the fixture to the top of the file, it's easier for code maintenance that way

lib/spack/spack/environment/environment.py
```diff
@@ -320,17 +328,26 @@ def create_in_dir(
             string, it specifies the path to the view
         keep_relative: if True, develop paths are copied verbatim into the new environment file,
             otherwise they are made absolute
+        include_concrete: concrete environment names/paths to be included
     """
-    initialize_environment_dir(manifest_dir, envfile=init_file)
 
-    if with_view is None and keep_relative:
+    if with_view is None and keep_relative and not include_concrete:
```
becker33: The ordering issues in this method are subtle. We should add comments explaining what's going on. As I understand it, the tricky things to make clear are:

1. The `ensure_included_*` methods have to be called before `initialize_environment_dir` so that we don't leave an empty directory behind if the checks fail
2. The other `include_concrete` conditional (calling `set_include_concrete`) has to be after `initialize_environment_dir`.
3. The conditional on this line is here to avoid paying a performance penalty for constructing the `EnvironmentManifestFile` object if we don't need to

I got pretty far into suggesting changes that would've violated those constraints before I had a handle on why they were ordered the way they are, so a couple comments to make it clearer in the future would help.

lib/spack/spack/environment/environment.py
```diff
@@ -363,6 +380,60 @@ def ensure_env_root_path_exists():
         fs.mkdirp(env_root_path())
 
 
+def ensure_included_envs_exist(include_concrete: List[str]) -> None:
```
becker33: These two methods should be renamed to use `validate_` instead of `ensure_`, since `ensure_` implies that they will fix the issue if they encounter one.

lib/spack/spack/environment/environment.py
```diff
@@ -363,6 +380,60 @@ def ensure_env_root_path_exists():
...
+    Raises:
+        SpackEnvironmentError: if any of the included environments are not concrete
+    """
+
+    non_concrete_envs = set()
+
+    for env_path in include_concrete:
+        if not os.path.exists(Environment(env_path).lock_path):
+            non_concrete_envs.add(Environment(env_path).name)
+            continue
```
becker33: This continue looks like it's vestigial from sometime when this method was doing more, I'm pretty sure it's not doing anything now.

lib/spack/spack/environment/environment.py
```diff
@@ -363,6 +380,60 @@ def ensure_env_root_path_exists():
...
+       include_concrete: list of already existing concrete environments to include
+
+    Raises:
+        SpackEnvironmentError: if any of the included environments do not exist
+    """
+
+    missing_envs = set()
+
+    for i in range(len(include_concrete)):
+        env_name = include_concrete[i]
```
becker33:
```suggestion
    for i, env_name in enumerate(include_concrete):
```

lib/spack/spack/environment/environment.py
```diff
@@ -363,6 +380,60 @@ def ensure_env_root_path_exists():
...
+
+        if is_env_dir(env_name):
+            include_concrete[i] = env_name
+        elif exists(env_name):
+            include_concrete[i] = root(env_name)
```
becker33: I think having this method mutate the list in place in addition to validation is dangerous and could lead to difficult issues to debug later. Can we move the list mutating to the `set_include_concrete` method?

lib/spack/spack/environment/environment.py
```diff
@@ -826,6 +906,21 @@ def _construct_state_from_manifest(self):
         else:
             self.views = {}
 
+        # Extract and process include_concrete
+        # Grabs include_concrete specs and put in memory
+        if not self.include_concrete:
```
becker33: Is there some reason we need the conditional here? I don't understand how `self.include_concrete` would already be set.

lib/spack/spack/environment/environment.py
```diff
@@ -933,6 +1059,65 @@ def repo(self):
...
+                continue
+
+            with open(env.lock_path) as f:
+                lockfile_as_dict = env._read_lockfile(f)
+
+            # Lockfile_meta must match each env and use at least format version 5
+            if lockfile_meta is None:
+                lockfile_meta = lockfile_as_dict["_meta"]
+            elif lockfile_meta != lockfile_as_dict["_meta"]:
+                raise SpackEnvironmentError("All lockfile _meta values must match")
```
becker33, line 1220: I think this check is too strict -- is the goal here to check that the set of keys are equal, or keys and values are equal?

lib/spack/spack/environment/environment.py
```diff
@@ -826,6 +906,21 @@ def _construct_state_from_manifest(self):
...
+
+        if self.include_concrete:
+            if os.path.exists(self.lock_path):
+                with open(self.lock_path) as f:
+                    data = self._read_lockfile(f)
+
+                if included_concrete_name in data:
+                    self.included_concrete_specs = data[included_concrete_name]
+            else:
+                self.include_concrete_envs()
```
becker33: What if the manifest include data is different than the data from the lockfile? I think that we need to check that the list of included environments is the same, and add a test for changing the included environments in an existing combined environment (both adding and removing).

lib/spack/spack/environment/environment.py
```diff
@@ -835,10 +930,37 @@ def _construct_state_from_manifest(self):
...
+            included_concretized_user_specs.extend(concretized_user_specs)
+
+        return self.concretized_user_specs + included_concretized_user_specs
+
+    def all_concretized_orders(self):
+        included_concretized_order = []
+        for concretized_orders in self.included_concretized_order.values():
+            included_concretized_order.extend(concretized_orders)
+
+        return self.concretized_order + included_concretized_order
```
becker33:
```suggestion
        concretized_order = self.concretized_order[:]
        for included_concretized_order in self.included_concretized_order.values():
            concretized_order.extend(included_concretized_order)

        return concretized_order
```
This version copies each included hash once instead of twice

lib/spack/spack/environment/environment.py
```diff
@@ -835,10 +930,37 @@ def _construct_state_from_manifest(self):
             if "path" not in entry:
                 self.dev_specs[name]["path"] = name
 
+    def all_concretized_user_specs(self):
+        included_concretized_user_specs = []
+        for concretized_user_specs in self.included_concretized_user_specs.values():
+            included_concretized_user_specs.extend(concretized_user_specs)
+
+        return self.concretized_user_specs + included_concretized_user_specs
```
becker33:
```suggestion
        concretized_user_specs = self.concretized_user_specs[:]
        for included_concretized_user_specs in self.included_concretized_user_specs.values():
            concretized_user_specs.extend(included_concretized_user_specs)

        return concretized_user_specs
```

lib/spack/spack/environment/environment.py
```diff
@@ -854,6 +976,10 @@ def clear(self, re_read=False):
         self.concretized_order = []  # roots of last concretize, in order
         self.specs_by_hash = {}  # concretized specs by hash
         self.invalidate_repository_cache()
+        self.included_concretized_user_specs = {}  # user specs from last concretize's included env
+        self.included_concretized_order = {}  # root specs of the included envs, keyed by env path
+        self.included_specs_by_hash = {}  # concretized specs by hash from the included envs
```
becker33: At line 825-832 we instantiate the same list of variables, but there is also `self.included_concrete_specs`. I'm having trouble tracking why that variable is substantively different and wouldn't need to be cleared here.

becker33 (changes requested):
Mostly minor comments

RikkiButler20:
@spackbot fix style

lib/spack/spack/environment/environment.py
RikkiButler20: This is for checking that all included environments have the same `file-type`, `lockfile-version`, and `specfile-version`

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

lib/spack/spack/environment/__init__.py
```diff
@@ -50,26 +53,37 @@
...
      - ✅
      -
+     -
    * - ``v0.18:``
      - ✅
      - ✅
      - ✅
      - ✅
+     -
+   * - ``v0.21:``
```
becker33: This should be `v0.22:` since it didn't merge before v0.21

lib/spack/spack/cmd/env.py
```diff
@@ -109,19 +112,32 @@ def env_create(args):
         # the environment should not include a view.
         with_view = None
 
+    include_concrete = None
+    if hasattr(args, "include_concrete"):
+        include_concrete = args.include_concrete
```
becker33, line 112: This should be able to be

```suggestion
    include_concrete = args.include_concrete
```

I'm assuming there's a reason that didn't work, but I'm curious why so we can hunt that down.

lib/spack/spack/test/cmd/env.py
```diff
@@ -1515,6 +1594,186 @@ def test_env_without_view_install(tmpdir, mock_stage, mock_fetch, install_mocker
...
+    test.concretize()
+    test.write()
+
+    env("create", "--include-concrete", test.path, "combined_env")
+
+    combined = ev.read("combined_env")
+    combined_yaml = combined.manifest["spack"]
+
+    assert "include_concrete" in combined_yaml
+    assert test.path in combined_yaml["include_concrete"]
```
becker33: These two tests are very similar, would it be easier to maintain if you used a parametrized boolean to see whether to pass a path or a name?

lib/spack/spack/environment/environment.py
```diff
@@ -1034,6 +1177,50 @@ def scope_name(self):
...
+            # Lockfile_meta must match each env and use at least format version 5
+            if lockfile_meta is None:
+                lockfile_meta = lockfile_as_dict["_meta"]
+            elif lockfile_meta != lockfile_as_dict["_meta"]:
+                raise SpackEnvironmentError("All lockfile _meta values must match")
+            elif lockfile_meta["lockfile-version"] < 5:
+                raise SpackEnvironmentError("The lockfile format must be at version 5 or higher")
+
+            # Copy unique root specs from env
+            self.included_concrete_specs[env_path] = {"roots": []}
```
becker33: It looks like this is only iterating over the "local" specs in the included environment, won't this cause a problem if we have a chain of includes?

becker33 (dismissed):
I think this still needs a test for the following:

```
$ spack env create test1
$ spack -e test1 add foo
$ spack -e test1 concretize
$ spack env create --include-concrete test1 test2
$ spack -e test2 add bar
$ spack -e test2 concretize
$ spack env create --include-concrete test2 test3
$ spack -e test3 concretize
# should include foo and bar
```

lib/spack/spack/cmd/env.py
RikkiButler20: Yea, when `spack activate` cmd is used with the `--create` flag it doesn't go through the def create method so include_concrete doesn't get initialized.

becker33: :+1: Can you add a comment explaining that?

RikkiButler20:
@spackbot fix style

lib/spack/spack/test/cmd/env.py
```diff
@@ -1515,6 +1594,197 @@ def test_env_without_view_install(tmpdir, mock_stage, mock_fetch, install_mocker
...
+
+    env("create", "--include-concrete", "test2", "test3")
+    test3 = ev.read("test3")
+
+    with open(test3.lock_path) as f:
+        lockfile_as_dict = test3._read_lockfile(f)
+
+    test3_include = lockfile_as_dict["include_concrete"][test2.path]["include_concrete"]
+
+    assert False
```
becker33: `assert test3_include and False` will fix the style issue while you can keep testing this.

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

RikkiButler20:
@spackbot fix style

tldahlgren:
Interesting that it has been 6 hours and prechecks are still in `Waiting for status to be reported` with only readthedocs completed.  Could it be conflicts related?

tldahlgren:
@spackbot help

tldahlgren:
TBD. Will closing and re-opening force pre-checks to complete despite conflicts?

lib/spack/spack/test/cmd/env.py
```diff
@@ -1636,6 +1715,275 @@ def test_env_without_view_install(tmpdir, mock_stage, mock_fetch, install_mocker
...
+    # Re-concretize test4 to reflect the re-concretization of included test2
+    # and ensure that its included specs are up-to-date
+    test2.concretize()
+    test2.write()
+    test4.concretize()
+
+    concrete_specs = [s for s, _ in test4.concretized_specs()]
+    assert Spec("zlib") not in concrete_specs
+
+    # TBD: Should duplicate (included) specs be allowed in concretized_specs()?
```
tldahlgren: IIRC. The answer to this is "no" so this comment needs to be removed before merge.

lib/spack/spack/environment/environment.py
```diff
@@ -942,6 +1053,31 @@ def _construct_state_from_manifest(self):
...
+
+        return concretized_user_specs
+
+    def all_concretized_orders(self) -> List[str]:
+        """Returns all of the concretized order of the environment and
+        its included environment(s)."""
+        concretized_order = self.concretized_order[:]
+        for included_concretized_order in self.included_concretized_order.values():
+            for included in included_concretized_order:
+                # TBD: Don't duplicate included spec(s)
```
tldahlgren: This comment should be removed if the decision and action are correct.

lib/spack/spack/environment/environment.py
```diff
@@ -942,6 +1053,31 @@ def _construct_state_from_manifest(self):
...
         self._process_view(spack.config.get("view", True))
+        self._process_concrete_includes()
+
+    def all_concretized_user_specs(self) -> List[Spec]:
+        """Returns all of the concretized user specs of the environment and
+        its included environment(s)."""
+        concretized_user_specs = self.concretized_user_specs[:]
+        for included_specs in self.included_concretized_user_specs.values():
+            for included in included_specs:
+                # TBD: Don't duplicate included spec(s)
```
tldahlgren: This comment should be removed if the decision and action are correct.

lib/spack/spack/test/cmd/env.py
tldahlgren: The unit test has been completed.

lib/spack/spack/environment/environment.py
tldahlgren: I thought the property name was confusing so renamed it to `included_concrete_spec_data` to reflect what it contains.

I also finished implementing  `set_included_concretized_user_specs` to include `add_specs` to populate the `included_concretized_order` and `included_concretized_user_specs` properties with included/nested concrete includes.

lib/spack/spack/environment/environment.py
tldahlgren: It was removed at some point.

lib/spack/spack/environment/environment.py
tldahlgren: This change was made at some point.

lib/spack/spack/environment/environment.py
tldahlgren: This change was made some time ago.

lib/spack/spack/environment/environment.py
tldahlgren: The method that now performs this mutation is being done **before** the list is passed to the validation methods.

lib/spack/spack/environment/environment.py
tldahlgren: The method no longer makes this check.  There is now a `_process_concrete_includes` method that extracts and loads included concrete spec data into memory that is called from `_construct_state_from_manifest`.  The new method is called as the last phase of construction.

lib/spack/spack/environment/environment.py
tldahlgren: @becker33 It would be helpful for you to revisit the method to determine if the requested changes were adequately addressed.

tldahlgren:
> I think this still needs a test for the following:
> 
> ```
> $ spack env create test1
> $ spack -e test1 add foo
> $ spack -e test1 concretize
> $ spack env create --include-concrete test1 test2
> $ spack -e test2 add bar
> $ spack -e test2 concretize
> $ spack env create --include-concrete test2 test3
> $ spack -e test3 concretize
> # should include foo and bar
> ```

There is a unit test -- `test_concretize_nested_included_concrete` -- that goes further to not only demonstrate the expected effects when two environments include the same environment concretized before and after a spec change (respectively) but to also exercise a case where those two environments are included in a fourth before and after reconcretization.

Goals of the test are to demonstrate:
1. each environment including a concrete environment has the concretized specs at creation time;
2. changes to the included environment are not reflected in the including environment until both are re-concretized; and
3. numbers 1 and 2 apply for nested included environments.

lib/spack/spack/environment/environment.py
tldahlgren: @alalazo Has this issue been addressed to your satisfaction?

tldahlgren:
> What is the synchronization policy between this environment and the included one (if e.g. the included environment changes independently)?

IMO `test_concrete_nested_included_concrete` illustrates the sequence as I understand it.

Creation of an environment that includes a concrete environment will have its concrete specs at creation time.  The only way the environment will reflect changes in an updated and re-concretized included environment is if the environment itself is re-concretized.  The same principle applies for nested included environments.  

For example, if `a` includes `b` and `b` includes `c`, `a` can only reflect changes to `c` if `b` is re-concretized to reflect those changes before `a` is re-concretized.

At least that's how the code is currently implemented.

lib/spack/spack/environment/environment.py
tldahlgren: I agree that there are too many properties/responsibilities but included equivalents were added to the constructor at some point since your comment.

tgamblin (dismissed):

tldahlgren:
@alalazo @becker33 Is it okay to resolve the remaining conversations so we can merge this PR?

tgamblin:
planning to after I deconflict it

tldahlgren:
> planning to after I deconflict it

Okay.  I was just starting to do that but won't if you've already started.

tldahlgren:
The CodeCov numbers currently reported in the checks do not look right at all.

Following `Details` -> `View this Pull Request on CodeCov` I see **+.01%** change (head **83.06%**, patch **90.70%**).

tgamblin (approved):

lib/spack/spack/environment/environment.py
alalazo: I still think this PR introduces a lot of complexity to `Environment`. The class now is ~1.6k lines long and with a hundred methods / properties. Most attributes are just used in a handful of methods.

lib/spack/spack/environment/environment.py
```diff
@@ -367,6 +383,14 @@ def create_in_dir(
 
     return env
 
+    # Must be done after environment is initialized
+    if include_concrete:
+        manifest.set_include_concrete(include_concrete)
+
+    if not keep_relative and init_file is not None and str(init_file).endswith(manifest_name):
+        init_file = pathlib.Path(init_file)
+        manifest.absolutify_dev_paths(init_file.parent)
```
alalazo, line 392: This code is unreachable, see:
```console
return env
```
above

alalazo (changes requested):

lib/spack/spack/environment/environment.py
alalazo: Not really. This interface seems to be redundant. Why did we add a new argument `include_concrete`? Couldn't the presence of an included environment be inferred from the `init_file` ? Does it make sense to pass an `init_file` with included environments and not add `include_concrete` or vice-versa?

alalazo:
> Is it okay to resolve the remaining conversations so we can merge this PR?

If we are in a hurry to merge a PR, in my opinion we should leave open discussions open and force merge. In that way point of discussions are easier to spot later. Otherwise the "Unresolved conversations" will just push us to close stuff only to the detriment of our workflows.

lib/spack/spack/environment/environment.py
tldahlgren: That looks like unresolved rebase cleanup from moving up and removing the duplicate create methods .  I'll track down where the unreachable code is from and fix it.

alalazo: @tgamblin Did you see :point_up: This is very likely a bug?

tldahlgren: I checked a couple of commits back for a couple of rebases and have not found where that was introduced yet.  Will proceed with the fix per Todd.

tldahlgren: Looks like it is likely rebase residual related to conflict resolution with #35273 .

tldahlgren: Removed in https://github.com/spack/spack/pull/44065.

PR #33772, opened 2022-11-08
Fixed external HIP library detection

bvanessen:
Fixed a bug where the external HIP library is found in a nested direc…tory, even on newer releases of ROCm.

@tgamblin This is a revision to #33344 where even on new versions of HIP, the external library can be found at either level of the hierarchy.

### DIFF
--- a/var/spack/repos/builtin/packages/hip/package.py
+++ b/var/spack/repos/builtin/packages/hip/package.py
@@ -291,8 +291,13 @@ class Hip(CMakePackage):
             hip_libs_at_top = os.path.basename(self.spec.prefix) != "hip"
             # We assume self.spec.prefix is  /opt/rocm-x.y.z for rocm-5.2.0 and newer
             # and /opt/rocm-x.y.z/hip for older versions
+            # However, depending on how an external is found it can be at either level
+            # of the installation path
             if self.spec.satisfies("@5.2.0:"):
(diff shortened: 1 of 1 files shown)


mathstuf (approved):
Picking this change fixed my `crusher` ROCm builds :) .

mathstuf:
Sidenote: it seems that ROCm-using packages need this to know where the `clang++` lives. What is the best way for consumers to copy/reuse this logic? Can a custom attribute be put into `spec["hip"]` when depended upon?

mathstuf:
My solution was to instead ask the `llvm-amdgpu` package where the `clang` lives.

tgamblin (approved):

tgamblin:
@mathstuf do you want to submit a PR for your approach?

mathstuf:
No idea if it's the right thing or not. I'd like some input from someone with more experience with HIP than "I made ParaView build with it once" :) .

kwryankrattiger:
The problem with external HIP is it can't inspect `spec["llvm-amdgpu"]` for the llvm path so it has to base it off of the `rocm_path`. The `ROCmPackage` could provide a helper function for getting clang and such, but that would require every HIP package to pass the correct variables to the build system which seems like a step backwards.

In theory, all of this could be moved to the `ROCmPackage`..but for the cases where HIP is used without that this would still be a problem.

PR #33773, opened 2022-11-09
depfile: improve tab completion

haampie:
This PR allows you to do:

```
spack env create -d .
spack -e . add python
spack -e . concretize
spack -e . env depfile -o Makefile

make in<tab>              # -> install
make install-<tab>        # -> install-deps/
make install-deps/py<tab> # -> install-deps/python-x.y.z-hash
make install/zl<tab>      # -> install/zlib-x.y.z-hash

make SP<tab>              # -> make SPACK
make SPACK_<tab>          # -> make SPACK_INSTALL_FLAGS=
```

This is thanks to a phony convenience targets (note that the actual targets are still put in `<env dir>/.spack-env/makedeps` independently of current working directory and location of the Makefile.

In the case of custom make target prefixes `--make-target-prefix hello` no such phony targets are created, since then tab completion already works from `make hello/<tab>` -- the point of custom targets is to avoid conflicts when the Makefile is included elsewhere, so we shouldn't those "shortcuts" to install and install-deps there.

Main changes of the PR:

- Drop the leading `.` in `.install` and `.install-deps` so they are tab-completed 
- Change `install` and `install-deps` targets format from `<hash>` into `<pkg name>-<version>-<hash>` so that targets, so targets are discoverable and easier to interpret
- Add phony convenience targets/shortcuts when there's no `--make-target-prefix <prefix>` given

[asciicast](https://asciinema.org/a/ximkYdruPKh7IDHgMyVM9Yacx)

### DIFF
--- a/lib/spack/spack/cmd/env.py
+++ b/lib/spack/spack/cmd/env.py
@@ -674,13 +674,14 @@ class MakeTargetVisitor(object):
         return ""
 
     def accept(self, node):
-        dag_hash = node.edge.spec.dag_hash()
+        fmt = "{name}-{version}-{hash}";
+        tgt = node.edge.spec.format(fmt)
         spec_str = node.edge.spec.format(
             "{name}{@version}{%compiler}{variants}{arch=architecture}"
         )
         buildcache_flag = self.build_cache_flag(node.depth)
-        prereqs = " ".join([self.target(dep.spec.dag_hash()) for dep in self.neighbors(node)])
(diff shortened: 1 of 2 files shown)


haampie:
@spackbot fix style

haampie:
Actually tab completion for SPACK_INSTALL_FLAGS is possible too by just setting it to empty variable, let me push that.

trws (approved):

PR #33777, opened 2022-11-09
ensure external PythonPackages have python deps

becker33:
Currently, external `PythonPackage`s cause install failures because the logic in `PythonPackage` assumes that it can ask for `spec["python"]`.  Because we chop off externals' dependencies, an external Python extension may not have a `python` dependency.

This PR resolves the issue by guaranteeing that a `python` node is present in one of two ways:
1. If there is already a `python` node in the DAG, we wire the external up to it.
2. If there is no existing `python` node, we wire up a synthetic external `python` node, and we assume that it has the same prefix as the external.

The assumption in (2) isn't always valid, but it's better than leaving the user with a non-working `PythonPackage`.

The logic here is specific to `python`, but other types of extensions could take advantage of it.  Packages need only define `update_external_dependencies(self)`, and this method will be called on externals after concretization.  This likely needs to be fleshed out in the future so that any added nodes are included in concretization, but for now we only bolt on dependencies post-concretization.

Closes #33386

### DIFF
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -218,6 +218,22 @@ class PythonPackage(PythonExtension):
             name = cls.pypi.split("/")[0]
             return "https://pypi.org/simple/" + name + "/"
 
+    def update_external_dependencies(self):
+        """
+        Ensure all external python packages have a python dependency
+        """
+        # TODO: Include this in the solve, rather than instantiating post-concretization
+        if "python" not in self.spec:
+            if "python" in self.spec.root:
+                python = self.spec.root["python"]
+            else:
+                python = Spec("python")
+                repo = spack.repo.path.repo_for_pkg(python)
+                python.namespace = repo.namespace
+                python._mark_concrete()
+                python.external_path = self.prefix
+            self.spec.add_dependency_edge(python, ("build", "link", "run"))
+
     @property
     def headers(self):
         """Discover header files in platlib."""
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -15,6 +15,7 @@ import llnl.util.tty as tty
 import spack.builder
 import spack.multimethod
 import spack.package_base
+import spack.spec
 from spack.directives import build_system, depends_on, extends
 from spack.error import NoHeadersError, NoLibrariesError, SpecError
 from spack.version import Version
@@ -227,7 +228,7 @@ class PythonPackage(PythonExtension):
             if "python" in self.spec.root:
                 python = self.spec.root["python"]
             else:
-                python = Spec("python")
+                python = spack.spec.Spec("python")
                 repo = spack.repo.path.repo_for_pkg(python)
                 python.namespace = repo.namespace
                 python._mark_concrete()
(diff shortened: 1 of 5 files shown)


tgamblin (approved):

haampie:
What a ~terrible~ courageous idea to create a hack like this and merge it just for a release :|

> This likely needs to be fleshed out in the future so that any added nodes are included in concretization, but for now we only bolt on dependencies post-concretization.

Why don't you do it in the future then.

tgamblin:
The reason this was merged is because mainline Spack broke previously working production python configurations (with externals).  This fixes those configurations without impacting much (anything?) else.

balay:
I'm getting the following errors:
```
[balay@pj01 spack]$ nice ./bin/spack install xsdk@0.8.0 %gcc@12.2.1 ^py-mpi4py@master ^exago~python |& tee spack-build.log
==> Warning: using "zlib@1.2.12" which is a deprecated version
==> Error: Class constructor failed for package 'builtin.py-libensemble'.

Caused by:
AssertionError:
  File "/home/balay/spack/lib/spack/spack/repo.py", line 1172, in get
    return package_class(spec)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/balay/spack/lib/spack/spack/package_base.py", line 691, in __init__
    if self.is_extension:
       ^^^^^^^^^^^^^^^^^
  File "/home/balay/spack/lib/spack/spack/package_base.py", line 1296, in is_extension
    return self.extendee_spec is not None
           ^^^^^^^^^^^^^^^^^^
  File "/home/balay/spack/lib/spack/spack/package_base.py", line 1265, in extendee_spec
    assert len(deps) == 1

```
Reverting this merge gets the above working. So how do I fix this issue?

balay:
And some of my builds don't show this issue. And I don't know what the trigger is here..
- error https://gitlab.com/xsdk-project/spack-xsdk/-/jobs/3299494118
- noerror https://gitlab.com/xsdk-project/spack-xsdk/-/jobs/3299494114

balay:
This issue comes up with `py-petsc4py` as well - and is likely related to having a `py-numpy` entry in `~/.spack/packages.yaml`. If I remove it -  the builds starts..
```console
[glci@pj01 spack]$ cat ~/.spack/packages.yaml
packages:
  python:
    externals:
    - spec: python@3.11.0
      prefix: /usr
  py-numpy:
    externals:
    - spec: py-numpy@1.22.0
      prefix: /usr
[glci@pj01 spack]$ ./bin/spack install py-petsc4py
==> Error: Class constructor failed for package 'builtin.py-petsc4py'.

Caused by:
AssertionError: 
  File "/home/glci/spack/lib/spack/spack/repo.py", line 1172, in get
    return package_class(spec)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/glci/spack/lib/spack/spack/package_base.py", line 691, in __init__
    if self.is_extension:
       ^^^^^^^^^^^^^^^^^
  File "/home/glci/spack/lib/spack/spack/package_base.py", line 1296, in is_extension
    return self.extendee_spec is not None
           ^^^^^^^^^^^^^^^^^^
  File "/home/glci/spack/lib/spack/spack/package_base.py", line 1265, in extendee_spec
    assert len(deps) == 1

[glci@pj01 spack]$ emacs ~/.spack/packages.yaml
[glci@pj01 spack]$ cat ~/.spack/packages.yaml
packages:
  python:
    externals:
    - spec: python@3.11.0
      prefix: /usr
[glci@pj01 spack]$ ./bin/spack install py-petsc4py
==> Installing libiconv-1.16-lwl5n7gciyzhhxf3lja6p726qxufyyza
...
```

adamjstewart:
Not opposed to this, but would have liked to know about it. I'm guessing this closes #9149.

PR #33781, opened 2022-11-09
Enable fetch cooldown in more cases

blue42u:
...And hopefully reduce the number of FetchCacheErrors that bubble up.

My previous PRs (#32137 and #33509) added a cooldown between attempts to fetch a buildcache's `index.json`, but only when the mirror had already been fetched once. The first of these two patches extends this to also include "new" mirrors that did not load an index into the cache, either due to failures or because there is no `index.json`. (Note that this does *not* include mirrors that were deleted and re-added, the `_last_fetch_times` entry is deleted in that case.)

The second patch adjusts the usage in `find_by_hash` to avoid throwing a FetchCacheError if the update fails. This function previously didn't throw exceptions, so keeping this from propagating reduces surprises in other locations.


blue42u:
@scottwittenburg Can you check if this fixes your cache fetch issue?

scottwittenburg:
> Can you check if this fixes your cache fetch issue?

Yes, thanks, it does.  With this change, each configured mirror that doesn't have an index seems to be checked just once.

blue42u:
Closing in favor of https://github.com/spack/spack/pull/34360

PR #33784, opened 2022-11-09
ucx: fix int overflow

haampie:
See https://github.com/openucx/ucx/issues/862 and the linked PR / commit on
master.


hppritcha (approved):

bernhardkaindl (approved):
LGTM, can only be an improvement to change the int -> ssize_t in these locations.

PR #33792, opened 2022-11-09
show external status as [e]

### LINKED ISSUE #24684: Add label in concretizer for external packages
I'd like ability to see external packages in the concretized output so its easy to know what packages are external and which are not

### Rationale

This is used to identify package preference and help pinpoint before doing `spack install` what are the issues.

### Description

In this concretized output, I have specified `m4` as an external. The symbol on left `-` is not clear if its an external or installed from source. The `[+]` is clear that package is installed.

```
[+]  yxi7z3o                  ^libtool@2.4.6%gcc@10.2.0 arch=cray-sles15-zen2
 -   wlj6ezo                      ^m4@1.4.18%gcc@10.2.0+sigsegv patches=3877ab548f88597ab2327a2230ee048d2d07ace1062efe81fc92e91b7f39cd00,fc9b61654a3ba1a8d6cd78ce087e7c96366c290bc8d2c299f09828d793b853c8 arch=cray-sles15-zen2
```
I would propose we have a `[E]` for external package. We can potentially extend this to other names during concretizer for labels to help see concretized output

This also relates to https://github.com/spack/spack/issues/18588 when trying to analyze output of `spack concretize` to make it easy to digest the information considering when we concretize a huge software stack. These changes to concretized output will be appreciated.

This was discussed in https://spackpm.slack.com/archives/C5W7NKZJT/p1625194987305100

becker33:
Fixes #24684

@lukebroskop this should help you with debugging what your externals are

### DIFF
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -174,12 +174,14 @@ CLEARSIGN_FILE_REGEX = re.compile(
 SPECFILE_FORMAT_VERSION = 3
 
 
+# InstallStatus is used to map install statuses to symbols for display
+# Options are artificially disjoint for dispay purposes
 class InstallStatus(enum.Enum):
-    installed = 0
-    upstream = 1
-    missing = 2
-    absent = 3
-    external = 4
+    installed = "@g{[+]}  "
+    upstream = "@g{[^]}  "
+    missing = "@g{[e]}  "
+    absent = "@K{ - }  "
+    external = "@r{[-]} "
 
 
 def colorize_spec(spec):
@@ -4437,7 +4439,10 @@ class Spec(object):
 
     def tree(self, **kwargs):
         """Prints out this spec and its dependencies, tree-formatted
-        with indentation."""
+        with indentation.
+
+        Status function may either output a boolean or an InstallStatus
+        """
         color = kwargs.pop("color", clr.get_color_when())
         depth = kwargs.pop("depth", False)
         hashes = kwargs.pop("hashes", False)
@@ -4469,20 +4474,12 @@ class Spec(object):
 
             if status_fn:
                 status = status_fn(node)
-                if status == InstallStatus.external:
-                    out += clr.colorize("@g{[e]}  ", color=color)
-                elif status == InstallStatus.upstream:
-                    out += clr.colorize("@g{[^]}  ", color=color)
-                elif status == InstallStatus.absent:
-                    out += clr.colorize("@K{ - }  ", color=color)
-                elif status == InstallStatus.installed:
-                    out += clr.colorize("@g{[+]}  ", color=color)
-                elif status == InstallStatus.missing:
-                    out += clr.colorize("@r{[-]}  ", color=color)
+                if status in list(InstallStatus):
+                    out += clr.colorize(status.value, color=color)
+                elif status:
+                    out += clr.colorize("@g{[+]} ", color=color)
                 else:
-                    raise ValueError(
-                        "Node %s status %s is not a valid install status" % (node, status)
-                    )
+                    out += clr.colorize("@r{ - } ", color=color)
 
             if hashes:
                 out += clr.colorize("@K{%s}  ", color=color) % node.dag_hash(hlen)
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -4479,7 +4479,7 @@ class Spec(object):
                 elif status:
                     out += clr.colorize("@g{[+]}  ", color=color)
                 else:
-                    out += clr.colorize("@r{ - }  ", color=color)
+                    out += clr.colorize("@r{[-]}  ", color=color)
 
             if hashes:
                 out += clr.colorize("@K{%s}  ", color=color) % node.dag_hash(hlen)
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -173,7 +173,7 @@ CLEARSIGN_FILE_REGEX = re.compile(
 SPECFILE_FORMAT_VERSION = 3
 
 
-class InstallStatus():
+class InstallStatus:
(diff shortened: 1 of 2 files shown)


lib/spack/spack/spec.py
```diff
@@ -4567,6 +4567,9 @@ def install_status(self):
         """Helper for tree to print DB install status."""
         if not self.concrete:
             return None
+
+        if self.external:
+            return "external"
```
haampie: I know Python doesn't have enums, but sure there are better things than a function returning the union of `str`, `bool` and `None` and giving meaning to those. Can you please use something decent? E.g.

```python
class InstallStatus:
    EXTERNAL = 0
    INSTALLED_UPSTREAM = 1
    INSTALLED_LOCALLY = 2
    NOT_INSTALLED = 3
```

Also, `get_record` calls `query_by_spec_hash(key)` which is also called later for upstreams, sounds like it should be a single call then.

haampie (changes requested):
See above

lib/spack/spack/spec.py
tldahlgren: Unfortunately, only since 3.4:  https://docs.python.org/3/library/enum.html

tldahlgren: There's also https://pypi.org/project/enum34/ that is supported in earlier versions (including 2.7).

haampie: So that's something we can use in 0.19 🙃

becker33:
@spackbot fix style

lib/spack/spack/spec.py
```diff
@@ -204,6 +204,14 @@
 specfile_format_version = 3
 
 
+class InstallStatus:
+    installed = 0
+    upstream = 1
+    missing = 2
+    absent = 3
+    external = 4
```
haampie: (minor) maybe we should leave a comment that this is purely for display purposes, since statuses are not mutually exclusive (you can have missing, installed, upstream and absent externals)

tldahlgren: Did you see the `purple` and `white` examples at https://docs.python.org/3/library/enum.html#enum.Flag.__contains__?

haampie: Yeah, flags are better than enums for that. But if the use case is really about showing a status symbol and we can only show one symbol, i guess the enum is OK.

lib/spack/spack/spec.py
```diff
@@ -4617,14 +4633,20 @@ def tree(self, **kwargs):
 
             if status_fn:
                 status = status_fn(node)
-                if node.installed_upstream:
+                if status == InstallStatus.external:
+                    out += clr.colorize("@g{[<]}  ", color=color)
```
shahzebsiddiqui: could we do `[e]` for external the `[<]` seems a bit misleading. Ideally we should have some output at bottom to highlight what all these notations means `[-]`, `[+]`, `[e]`

haampie:
This would fix #35005

haampie:
@becker33 wanna get this in?

lib/spack/spack/spec.py
alalazo: Can we use `Enum` consistently for new code, since we support running on 3.6+ ?

lib/spack/spack/spec.py
wyphan: I personally prefer `[x]` for eXternal

becker33: In the `spack find` command options we use `x` for eXplicit, I don't think it's a good idea to conflate the two.

wyphan:
Confirmed working.
```shell
wyp@basecamp:~/work/spack$ spack env activate test

wyp@basecamp:~/work/spack$ spack find
==> In environment test
==> Root specs
hwloc +cuda

==> 0 installed packages

wyp@basecamp:~/work/spack$ spack concretize
==> Concretized hwloc+cuda ^cuda@12.1.0
 -   fya5ohq  hwloc@2.9.1%gcc@10.3.0~cairo+cuda~gl~libudev+libxml2~netloc~nvml~oneapi-level-zero~opencl+pci~rocm build_system=autotools cuda_arch=86 libs=shared,static arch=linux-ubuntu20.04-zen2
[e]  dnomir5      ^cuda@12.1.0%gcc@10.3.0~allow-unsupported-compilers~dev build_system=generic arch=linux-ubuntu20.04-zen2
 -   yetrfbh      ^libpciaccess@0.17%gcc@10.3.0 build_system=autotools arch=linux-ubuntu20.04-zen2
[e]  xx52cfg          ^libtool@2.4.6%gcc@10.3.0 build_system=autotools arch=linux-ubuntu20.04-zen2
 -   lus3nht          ^util-macros@1.19.3%gcc@10.3.0 build_system=autotools arch=linux-ubuntu20.04-zen2
 -   d6tccfg      ^libxml2@2.10.3%gcc@10.3.0~python build_system=autotools arch=linux-ubuntu20.04-zen2
 -   s66yoef          ^libiconv@1.17%gcc@10.3.0 build_system=autotools libs=shared,static arch=linux-ubuntu20.04-zen2
 -   xkw556f          ^xz@5.4.1%gcc@10.3.0~pic build_system=autotools libs=shared,static arch=linux-ubuntu20.04-zen2
 -   kgfplvu          ^zlib@1.2.13%gcc@10.3.0+optimize+pic+shared build_system=makefile arch=linux-ubuntu20.04-zen2
[e]  7evsp57      ^ncurses@6.2.20200212%gcc@10.3.0+symlinks+termlib abi=6 build_system=autotools arch=linux-ubuntu20.04-zen2
 -   dcywop2      ^pkgconf@1.9.5%gcc@10.3.0 build_system=autotools arch=linux-ubuntu20.04-zen2
```

wyphan (approved):

haampie (approved):

PR #33793, opened 2022-11-09
Allow packages to detect presence of slingshot network

becker33:
This will replace the last use-case for `platform=cray` constraints that were not modeling the use of the module system. In a  very few cases, the platform was being used as a proxy for the network interconnect available.

@bvanessen

### DIFF
--- a/lib/spack/spack/platforms/cray.py
+++ b/lib/spack/spack/platforms/cray.py
@@ -36,6 +36,10 @@ _ex_craype_dir = "/opt/cray/pe/cpe"
 _xc_craype_dir = "/opt/cray/pe/cdt"
 
 
+def slingshot_network():
+    return os.path.exists("/lib64/libcxi.so")
+
+
 def _target_name_from_craype_target_name(name):
     return _craype_name_to_target_name.get(name, name)
 
--- a/var/spack/repos/builtin/packages/aluminum/package.py
+++ b/var/spack/repos/builtin/packages/aluminum/package.py
@@ -5,6 +5,7 @@
 
 import os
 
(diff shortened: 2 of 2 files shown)


bvanessen (approved):
Works for me.

bvanessen:
@tldahlgren I have already approved this PR.

tldahlgren:
> @tldahlgren I have already approved this PR.

Yes, and I added you as assignee to indicate that you had done so, which allows one to quickly see which PRs have been commented on (by a non-spackbot account) or approved from the main PR list page.

bvanessen:
> > @tldahlgren I have already approved this PR.
> 
> Yes, and I added you as assignee to indicate that you had done so, which allows one to quickly see which PRs have been commented on (by a non-spackbot account) or approved from the main PR list page.

@tldahlgren Ah, okay.

lib/spack/spack/platforms/cray.py
```diff
@@ -36,6 +36,10 @@
 _xc_craype_dir = "/opt/cray/pe/cdt"
 
 
+def slingshot_network():
+    return os.path.exists("/lib64/libcxi.so")
```
tldahlgren, line 40: Ah.  Hard-coded path.  Is this guaranteed to be the path across systems?

tldahlgren (approved):

lib/spack/spack/platforms/cray.py
tldahlgren: Prefer not having a hard-coded path BUT recognize others can be added if/when needed.

PR #33795, opened 2022-11-09
support mkl link options

rscohn2:
better match intel-mkl options by making provides scalapack unconditional and implementing thread= variants.

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-mkl/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-mkl/package.py
@@ -8,6 +8,7 @@ import platform
 
 from spack.package import *
 
+
 @IntelOneApiPackage.update_description
 class IntelOneapiMkl(IntelOneApiLibraryPackage):
     """Intel oneAPI Math Kernel Library (Intel oneMKL; formerly Intel Math
--- a/var/spack/repos/builtin/packages/intel-oneapi-mkl/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-mkl/package.py
@@ -8,7 +8,6 @@ import platform
 
 from spack.package import *
 
-
 @IntelOneApiPackage.update_description
 class IntelOneapiMkl(IntelOneApiLibraryPackage):
     """Intel oneAPI Math Kernel Library (Intel oneMKL; formerly Intel Math
@@ -87,13 +86,20 @@ class IntelOneapiMkl(IntelOneApiLibraryPackage):
     variant(
         "cluster", default=False, description="Build with cluster support: scalapack, blacs, etc"
     )
+    variant(
+        "threads",
+        default="none",
+        description="Multithreading support",
+        values=("openmp", "tbb", "none"),
+        multi=False,
+    )
 
     depends_on("intel-oneapi-tbb")
     # cluster libraries need mpi
     depends_on("mpi", when="+cluster")
 
     provides("fftw-api@3")
-    provides("scalapack", when="+cluster")
+    provides("scalapack")
     provides("mkl")
     provides("lapack")
     provides("blas")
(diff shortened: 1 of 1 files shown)


ye-luo (approved):

lpoirel:
Thanks @rscohn2!
Some libs listed in the link advisor and in the intel-mkl previous implementation do not appear here (pthread and omp for instance). Are they not necessary ?

rscohn2:
The libraries are needed needed, but usually the developer will add -fopenmp and the compiler will provide the necessary libraries. I wasn't confident that the logic for the intel-mkl package will always work. I wanted to understand the use cases better before I added the libraries here.

var/spack/repos/builtin/packages/intel-oneapi-mkl/package.py
```diff
@@ -87,13 +87,20 @@ class IntelOneapiMkl(IntelOneApiLibraryPackage):
...
+        multi=False,
+    )
 
     depends_on("intel-oneapi-tbb")
     # cluster libraries need mpi
     depends_on("mpi", when="+cluster")
 
     provides("fftw-api@3")
-    provides("scalapack", when="+cluster")
+    provides("scalapack")
```
balay, line 103: How would one avoid scalapack - as that's not always desired [even if build with +mpi]?

https://github.com/spack/spack/pull/34186

ye-luo: It is a question to spack virtual package design. MKL only provides it, using it or not is a decision in the app side not the MKL package. Probably the virtual package feature allows you to specify the actual choice. spack expert probably have an answer.

balay: Likely you might want to implement the way hdf5 provides the lib list for different attributes..
```
$ git grep 'spec\["hdf5:' |grep libs
var/spack/repos/builtin/packages/berkeleygw/package.py:            buildopts.append("HDF5LIB=%s" % spec["hdf5:hl,fortran"].libs.ld_flags)
var/spack/repos/builtin/packages/quantum-espresso/package.py:                        "--with-hdf5-libs={0}".format(spec["hdf5:hl,fortran"].libs.ld_flags),
```

becker33: @ye-luo in Spack we currently often use variants for external packages to describe how they are used (linked) rather than what's on disk. This is a hacky kludge because we don't currently have a concept of a "use-variant", which would be an edge attribute variant for how the dependent uses the dependency, rather than a node attribute variant for what is on disk.

I think something like what @balay suggested would be a good idea in the meantime, before we are able to implement use-variants.

rscohn2: @ye-luo : a possible problem is that someone wants to use mkl for blas and another package for scalapack. I don't know how spack resolves the issue when there are 2 providers.

The old intel-mkl unconditionally `provides` scalapack, but only adds the libraries into the library list if mpi is also there. Previously, intel intel-oneapi-mkl package made both available with the +cluster variant.

The explicit variant seemed easier to understand/control, but it sounded like there were some compatibility problems. I can do it either way based on what people want.

ye-luo: Making the use side requests libraries with specific attributes seems a reasonable solution.
The `+cluster` option is clearly bad variant. It doesn't provide sufficient resolution.
Users need to independently turn on/off each component of MKL including BLAS, LAPACK, SCALAPACK, FFT, ClusterFFT. Then the use side may use BLAS from MKL and FFT from fftw3.

lpoirel: In https://gitlab.com/spack/spack/-/blob/develop/CHANGELOG.md?plain=1#L1490, there is also something that could be a solution. 
The intel-mkl package does something like that, it is just not very portable between the different packages providing for instance lapack, and not used by dependent packages.
What would be nice, I think, is to make the libs attribute of a virtual dependency resolve to an attribute that gives only the libs related to this virtual package. For instance, if I have a package foo that depends on scalapack, and I install `foo ^intel-mkl`, I would want ` spec["scalapack"].libs` in `Foo.install` to automatically return the `IntelMkl.scalapack_libs` attribute instead of `IntelMkl.libs` attribute, if present.

rscohn2: If you are using shared libraries, then blas, lapack, fft have the same set of libraries. Cluster support does add libraries, and there are different optional libraries for the scalapack, cluster fft, and cluster sparse solver. There could be separate variants for each of these, but each unique variant requires a separate install (3.5G). It is harmless to add extra libraries to the link line that are not referenced. I don't see any real benefit for trying to customize the library list. Can you provide specific examples of problems you are having by lack of resolution on libraries?

balay: I think its good to have a separation at-least at the 'mpi' dependency level [even when mpi is enabled]

And there must be a reason why MKL has separate libraries for some of these features - instead of a single bundle

https://www.intel.com/content/www/us/en/developer/tools/oneapi/onemkl-link-line-advisor.html#gs.k7zb0g

"Select cluster library:"

Also from experience - had to deal with issues where users were attempting to eliminate duplicate libraries for the same functionality that inadvertently creep up into the link line.

rscohn2:
Closing
Added threads support in #37586 
Kept +cluster  variant

PR #28446, opened 2022-01-17
Add 'cxi' (slingshot) variant

roblatham00:
While it may be the case that only `/opt/cray/modulefiles/libfabric/1.13.0.0` on the early access 'crusher' machine has support for it, the 'cxi' provider is something a user could request.

Requires libfabric to be an external package using cray's libfabric... should  "install" fail with a message about external packaging if cxi variant requested?  (like spectrum-mpi and cray-mpich do?)

PR #33796, opened 2022-11-09
libfabric can now find external installation

bvanessen:
Added support for libfabric to find an external installation and identify variants supported.

### DIFF
--- a/var/spack/repos/builtin/packages/libfabric/package.py
+++ b/var/spack/repos/builtin/packages/libfabric/package.py
@@ -167,6 +167,7 @@ class Libfabric(AutotoolsPackage):
         fi_info = Executable(self.prefix.bin.fi_info)
         fi_info()
 
+
 # This code gets all the fabric names from the variants list
 # Idea taken from the AutotoolsPackage source.
 def get_options_from_variant(self, name):
--- a/var/spack/repos/builtin/packages/libfabric/package.py
+++ b/var/spack/repos/builtin/packages/libfabric/package.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import re
+import spack.platforms.cray
 
 from spack.package import *
 
@@ -48,7 +49,7 @@ class Libfabric(AutotoolsPackage):
     version("1.4.2", sha256="5d027d7e4e34cb62508803e51d6bd2f477932ad68948996429df2bfff37ca2a5")
 
     fabrics = (
-        "cxi",
+        conditional("cxi", when=spack.platforms.cray.slingshot_network()),
         "efa",
         "gni",
         "mlx",
--- a/var/spack/repos/builtin/packages/libfabric/package.py
+++ b/var/spack/repos/builtin/packages/libfabric/package.py
@@ -160,17 +160,21 @@ class Libfabric(AutotoolsPackage):
             args.append("--with-kdreg=no")
 
         for fabric in self.fabrics:
-            valid_fabric = fabric
+            valid_fabric = None
             if hasattr(fabric, "when"):
                 if fabric.when is True:
                     # Explicitly extract the True value for downstream use
                     valid_fabric = "{0}".format(fabric)
                 else:
                     continue
-            if "fabrics=" + valid_fabric in self.spec:
-                args.append("--enable-{0}=yes".format(valid_fabric))
             else:
-                args.append("--enable-{0}=no".format(valid_fabric))
+                valid_fabric = fabric
+
+            if valid_fabric is not None:
+                if "fabrics=" + valid_fabric in self.spec:
+                    args.append("--enable-{0}=yes".format(valid_fabric))
+                else:
+                    args.append("--enable-{0}=no".format(valid_fabric))
 
         return args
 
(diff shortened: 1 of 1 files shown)


becker33 (dismissed):

becker33:
@spackbot fix style

becker33 (dismissed):

var/spack/repos/builtin/packages/libfabric/package.py
```diff
@@ -44,6 +48,7 @@ class Libfabric(AutotoolsPackage):
     version("1.4.2", sha256="5d027d7e4e34cb62508803e51d6bd2f477932ad68948996429df2bfff37ca2a5")
 
     fabrics = (
+        "cxi",
```
rajachan: FWIW, I'm not in favor of adding this as a fabric option here as the cxi provider is still not available in upstream libfabric, where we pull the releases from. See discussion in https://github.com/spack/spack/pull/28446

I am not a spack expert, but this could be handled differently with a site-specific config and some combination of dynamically loading the external package and setting FI_PROVIDER=cxi at runtime??

bvanessen: @rajachan I understand that the `cxi` code is not available upstream, I believe that we have had spack packages have variants and features for non-publicly accessible code bases.  I think that `mvapich2-gdr` and `cray-libsci` are an examples of such packages.  The key thing that I think that we would want to recognize is that CXI is a valid variant for some systems.  @becker33 @tgamblin How should we proceed?

bvanessen:
@harisubramoni Here is the PR that I was mentioning.

var/spack/repos/builtin/packages/libfabric/package.py
rajachan: Libfabric is not in the same bucket as other wholly binary/closed-source packages (and I am definitely not suggesting spack should not pull in binary/closed-source packages at all, that would be a silly stance to take :) ).  Adding a native 'fabrics' option for a provider that is not in the release tarball we are fetching does not seem right to me. 

I know I am being pedantic, and I really want to get this package to be as useful to folks as possible, so if this is the best way to handle it with spack and your use case then let's merge it!

rajachan: cc: @shefty for visibility

harisubramoni:
Got it!

var/spack/repos/builtin/packages/libfabric/package.py
shefty: I'm unfamiliar with spack and the intent here.  But libfabric can make use of external provider libraries.  If there's some way to install a cxi package separately, and all this is doing is letting libfabric know that it might exists, this makes sense.  If this is assuming cxi is part of the libfabric package, then I agree adding it here may not make sense.  (Basically, I don't know. :)

rajachan: A 'fabric' here is the equivalent of toggling `--enable-<provider>` so this is a build time thing. I also think this should use the provider as if it was an external provider, discovering at runtime.

bvanessen: @rajachan and @becker33  Okay, I understand the issue here.  I have tried to mitigate it by making the CXI fabric only available when on a cray based system that has a slingshot fabric.  This still doesn't address the issue that the code isn't available for public builds, but should avoid impacting other instances of the package.  Thoughts?  Also it needs to be rebased on #33793 and #33939.

bvanessen:
@spackbot fix style

var/spack/repos/builtin/packages/libfabric/package.py
rajachan: Yes, this looks a lot better! 👍

rajachan (dismissed):

bvanessen:
Rebased on head of develop to include #33793

rajachan (dismissed):

bvanessen:
@spackbot run pipeline

bvanessen:
@spackbot run pipeline

bvanessen:
Rebased on develop now that #33939 merged

bvanessen:
@spackbot fix style

scottwittenburg:
@spackbot fix style

bvanessen:
@spackbot run pipeline

bvanessen:
@becker33 @haampie Can this PR get merged?

var/spack/repos/builtin/packages/libfabric/package.py
```diff
@@ -126,13 +160,42 @@ def configure_args(self):
             args.append("--with-kdreg=no")
 
         for fabric in self.fabrics:
```
becker33: What about `for fabric in [f if isinstance(f, str) else f[0].value for f in self.fabrics]:`

And then you can jump straight to the `if "fabrics=" + fabric in self.spec:` check with a lot less fuss.

becker33 (changes requested):

var/spack/repos/builtin/packages/libfabric/package.py
bvanessen: @becker33 Thanks for the suggestion.  I have updated it.

becker33 (approved):

PR #33800, opened 2022-11-09
external find now searches system libraries

bvanessen:
Added the system library paths to the external find by libraries command.  This will enable discovery of external packages in /lib, /lib64, /usr/lib, /usr/lib64, /usr/local/lib, /usr/local/lib64.


haampie:
`/usr/local/lib`, `/usr/local/lib64` these are not default directories. Also this doesn't help much on Debian systems. What makes much more sense is parsing `/etc/ld.so.conf` for glibc, which is easy enough (read the lines, drop the comments, follow (relative) includes, allow globbing). For musl libc there is an even simpler format, the file name escapes me but you can google.

bvanessen:
> `/usr/local/lib`, `/usr/local/lib64` these are not default directories. Also this doesn't help much on Debian systems. What makes much more sense is parsing `/etc/ld.so.cache`, which is easy enough (read the lines, follow (relative) includes, drop the comments, allow globbing).

@haampie I am using the set of paths that spack internally uses for the `find_system_libraries` function call.  It think that it is fair to debate if this is the best behavior or not, but I am trying to get consistency.  As for debian, I don't have access to a system to test against.

haampie:
I don't think it's good to consistenly use wrong paths

bvanessen:
> I don't think it's good to consistenly use wrong paths

@haampie I would agree that we shouldn't use wrong paths.  I would as @becker33 @tgamblin @scheibelp to chime in to see if they want to change this or be consistent.

haampie:
I'm happy to contribute an `ld.so.conf` parser.

FWIW:

```console
$ docker run --rm ubuntu find /usr/lib64 /usr/lib /usr/local/lib /usr/local/lib64 -type f -maxdepth 1 | wc -l
find: '/usr/local/lib64': No such file or directory
       1
```

vs

```console
$ docker run --rm ubuntu cat /etc/ld.so.conf
include /etc/ld.so.conf.d/*.conf

$ docker run --rm ubuntu ls /etc/ld.so.conf.d/
libc.conf
x86_64-linux-gnu.conf

$ docker run --rm ubuntu cat /etc/ld.so.conf.d/libc.conf                                                                                
# libc default configuration
/usr/local/lib

$ docker run --rm ubuntu cat /etc/ld.so.conf.d/x86_64-linux-gnu.conf
# Multiarch support
/usr/local/lib/x86_64-linux-gnu
/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu

$ docker run --rm ubuntu find /usr/local/lib/x86_64-linux-gnu /lib/x86_64-linux-gnu /usr/lib/x86_64-linux-gnu -type f -maxdepth 1 | wc -l
find: '/usr/local/lib/x86_64-linux-gnu': No such file or directory
     178
```

(Another thing to verify is whether the libraries found are actually host compatible -- you can pretty much use `spack.util.elf.parse_elf` and verify `e_ident[EI_CLASS]`, `e_ident[EI_DATA]`, and `e_machine`. This avoids issues on Debian if it's actually using multiarch stuff)

bvanessen:
> I'm happy to contribute an `ld.so.conf` parser.
> 
> FWIW:
> 
> ```
> $ docker run --rm ubuntu find /usr/lib64 /usr/lib /usr/local/lib /usr/local/lib64 -type f -maxdepth 1 | wc -l
> find: '/usr/local/lib64': No such file or directory
>        1
> ```
> 
> vs
> 
> ```
> $ docker run --rm ubuntu cat /etc/ld.so.conf
> include /etc/ld.so.conf.d/*.conf
> 
> $ docker run --rm ubuntu ls /etc/ld.so.conf.d/
> libc.conf
> x86_64-linux-gnu.conf
> 
> $ docker run --rm ubuntu cat /etc/ld.so.conf.d/libc.conf                                                                                
> # libc default configuration
> /usr/local/lib
> 
> $ docker run --rm ubuntu cat /etc/ld.so.conf.d/x86_64-linux-gnu.conf
> # Multiarch support
> /usr/local/lib/x86_64-linux-gnu
> /lib/x86_64-linux-gnu
> /usr/lib/x86_64-linux-gnu
> 
> $ docker run --rm ubuntu find /usr/local/lib/x86_64-linux-gnu /lib/x86_64-linux-gnu /usr/lib/x86_64-linux-gnu -type f -maxdepth 1 | wc -l
> find: '/usr/local/lib/x86_64-linux-gnu': No such file or directory
>      178
> ```

@haampie So as far as `/usr/local/lib64` and `/usr/local/lib` there is a check to make sure that any provided paths are valid.  As far as using the cache file, is that consistent across platforms and who maintains it?  If you were to add a parser for that and integrated into the default system search paths, how long would that take?  I don't know if that is simple and should be added to this PR or as a follow on PR.

haampie:
It can be done separately, shouldn't take more than a few hours.

Edit: you can rebase on #33812 and use `spack.util.ld_so_conf.host_dynamic_linker_search_paths()`.

It adds some overhead but not a lot:

```
In [2]: %timeit spack.util.ld_so_conf.host_dynamic_linker_search_paths()
126 µs ± 301 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

bvanessen:
> It can be done separately, shouldn't take more than a few hours.
> 
> Edit: you can rebase on #33812 and use `spack.util.ld_so_conf.host_dynamic_linker_search_paths()`.
> 
> It adds some overhead but not a lot:
> 
> ```
> In [2]: %timeit spack.util.ld_so_conf.host_dynamic_linker_search_paths()
> 126 µs ± 301 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
> ```

@haampie I am thinking that this should be added as a supplemental path exploration and not a as replacement for the explicit definition of some standard search paths.  Do you know how well this cache is maintained on all relevant systems?

haampie:
> Do you know how well this cache is maintained on all relevant systems?

It's not the cache, it's the input of the cache. This is pretty much used by all popular Linux distros, maybe except NixOS.

To me the only sensible thing is the following:

- `external find` for _executables_ follows the logic of [`execl`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/execl.html) (so e.g. on Linux: look in `PATH` or `confstr(_CS_PATH)` which is typically `/usr/bin:/bin`.)
- `external find` for _libraries_ follows the logic of the _dynamic_ linkers: [gnu+linux](https://man7.org/linux/man-pages/man8/ld.so.8.html), [musl linux](https://www.musl-libc.org/doc/1.0.0/manual.html), [freebsd](https://www.freebsd.org/cgi/man.cgi?query=ldconfig&sektion=8&apropos=0&manpath=FreeBSD+13.1-RELEASE+and+Ports), (Darwin does not have config files only variables). The logic of #33812 ensures that we consider the most sensible dynamic linker first, namely the one that's used by Python which Spack itself runs under.

So, I'd rather see this PR rebased on #33812, since right now it makes no sense for Ubuntu/Debian, which is are the most popular Linux distros. And it should use `LD_LIBRARY_PATH` and drop `LIBRARY_PATH` since the latter is not runtime.

bvanessen:
> > Do you know how well this cache is maintained on all relevant systems?
> 
> It's not the cache, it's the input of the cache. This is pretty much used by all popular Linux distros, maybe except NixOS.
> 
> To me the only sensible thing is the following:
> 
> * `external find` for _executables_ follows the logic of [`execl`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/execl.html) (so e.g. on Linux: look in `PATH` or `confstr(_CS_PATH)` which is typically `/usr/bin:/bin`.)
> * `external find` for _libraries_ follows the logic of the _dynamic_ linkers: [gnu+linux](https://man7.org/linux/man-pages/man8/ld.so.8.html), [musl linux](https://www.musl-libc.org/doc/1.0.0/manual.html), [freebsd](https://www.freebsd.org/cgi/man.cgi?query=ldconfig&sektion=8&apropos=0&manpath=FreeBSD+13.1-RELEASE+and+Ports), (Darwin does not have config files only variables). The logic of [Add dynamic linker config parsing #33812](https://github.com/spack/spack/pull/33812) ensures that we consider the most sensible dynamic linker first, namely the one that's used by Python which Spack itself runs under.
> 
> So, I'd rather see this PR rebased on #33812, since right now it makes no sense for Ubuntu/Debian, which is are the most popular Linux distros. And it should use `LD_LIBRARY_PATH` and drop `LIBRARY_PATH` since the latter is not runtime.

@haampie Done.

haampie (approved):

PR #33802, opened 2022-11-10
Remove dependency on platform=cray and switch it to %rocmcc.

bvanessen:


bvanessen:
@becker33 This cleans up another `platform=cray` dependency.

becker33 (changes requested):
I'm fairly convinced that this is not an attribute of the compiler. @bvanessen I might need to talk to you in more detail about what is being modeled here.

benson31:
It _is_ in fact an attribute of the compiler -- this compiler (whatever it _actually_ is) apparently includes MPI in its default include/link paths. This is typical behavior for Cray's compiler (`CC`), but this compiler doesn't superficially "look like" the Cray compiler in the sense that it has some silly name, and this is enough to throw off CMake's MPI detection. When CMake searches for the MPI library, it recognizes that the *C++ compiler* is able to build and link MPI programs implicitly, so it doesn't fill in any of the include paths/link paths/link libraries in the `MPI::MPI_CXX` target. This is fine for programs with C++ linkage, but not for programs linked by the HIP compiler, which does not have this feature. Setting this flag causes CMake to ignore the fact that the compiler can implicitly link MPI programs and to search for the required libraries anyway. This gives a complete `MPI::MPI_CXX` target, so the needed libraries get included when the "HIP linker" (HIP compiler's linker wrapper) does its thing.

becker33:
@benson31 what does the compiler config look like for this particular compiler?

benson31:
@bvanessen will have to answer that. I don't have this issue with my build method.

bvanessen:
> @bvanessen will have to answer that. I don't have this issue with my build method.

```
  compilers:
  - compiler:
      spec: rocmcc@5.3.0
      paths:
        cc: cc
        cxx: CC
        f77: ftn
        fc: ftn
      flags: {}
      operating_system: rhel8
      target: any
      modules:
      - PrgEnv-amd
      - amd/5.3.0
      environment: {}
      extra_rpaths: []
```

becker33:
Oh yeah you need to update your compiler description to match the new work.

You need to use the standalone compiler, not the cray wrapper that adds all sorts of things into your link line that you don't want.

Something like 

```
compilers:
  - compiler:
      spec: rocmcc@5.3.0
      paths:
        cc: /path/to/amdclang
        cxx: /path/to/amdclang++
        f77: /path/to/amdflang  # or /path/to/cce/crayftn
        fc: /path/to/amdflang  # or /path/to/cce/crayftn
      flags: {}
      operating_system: rhel8
      target: any
      modules: []
      environment: {}
      extra_rpaths: []

becker33:
@bvanessen can you confirm whether the modified compiler config works as expected?

bvanessen:
@becker33 I can try it again. I was sort of waiting for a few other PRs to merge and so I can update and test this.

alecbcs:
@bvanessen are you still waiting on any PRs before getting to this PR?

alalazo:
Outdated by #43796

Issue #31184, opened 2022-06-17
Installation issue: gcc


matthewcurry:
### Steps to reproduce the issue

```console
$ spack install gcc@10.2
...
```

### Error message

<details><summary>Error message</summary><pre>
cc1: error: switch '-mcpu=thunderx2t99' conflicts with '-march=ar
              mv8-a+lse' switch [-Werror]
</pre></details>

### Information on your system

* **Spack:** 0.19.0.dev0 (bf990bc8ec2d6c51f7f40bc6b0f7fb3bba897a96)
* **Python:** 3.6.8
* **Platform:** linux-rhel7-thunderx2
* **Concretizer:** clingo

### Additional information

This bug was introduced in commit 2113b625d17178f24189c0059a6f360c181b589f. Building GCC on ThunderX2 with gcc@4.8.5 selects different values for -mcpu and -march, and thus fails.
[spack-build-env.txt](https://github.com/spack/spack/files/8929482/spack-build-env.txt)
[spack-build-out.txt](https://github.com/spack/spack/files/8929484/spack-build-out.txt)

Maintainers: @alalazo @michaelkuhn

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have run `spack maintainers <name-of-the-package>` and **@mentioned** any maintainers
- [X] I have uploaded the build log and environment files
- [X] I have searched the issues of this repo and believe this is not a duplicate

matthewcurry:
FYI, the commit that introduced this issue was 2113b625d17178f24189c0059a6f360c181b589f

michaelkuhn:
@haampie Could you take a look?

alalazo:
@matthewcurry Thanks for the detailed report.

matthewcurry:
Is there something I can do to help find a resolution?

haampie:
@alalazo is this an issue with arch spec?

So, the package.py adds `-mcpu=thunderx2t99` to the later stage builds (thanks to arch spec), and then there's a test in some configure script:

```
[...]

PR #33813, opened 2022-11-10
gcc: drop target bootstrap flags for aarch64

haampie:
See https://github.com/spack/spack/issues/31184

GCC bootstrap logic adds `-mpcu` for libatomic (iirc), which conflicts
with the `-march` flag we provide.

### DIFF
--- a/var/spack/repos/builtin/packages/gcc/package.py
+++ b/var/spack/repos/builtin/packages/gcc/package.py
@@ -682,6 +682,7 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage):
         # result in compilation errors, when gcc@7 is built with gcc@11, and znver3
         # is taken as a the target, which gcc@7 doesn't support.
         # Note we're not adding this for aarch64 because of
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/gcc/package.py
```diff
@@ -681,7 +681,8 @@ def build_optimization_config(self):
         # concretization, so we'll stick to that. The other way around however can
         # result in compilation errors, when gcc@7 is built with gcc@11, and znver3
         # is taken as a the target, which gcc@7 doesn't support.
-        if "+bootstrap %gcc" in self.spec:
+        # Note we're not adding this for aarch64 because of
+        if "+bootstrap %gcc" in self.spec and self.spec.target.family != "aarch64":
```
haampie, line 686:
```suggestion
        # https://github.com/spack/spack/issues/31184
        if "+bootstrap %gcc" in self.spec and self.spec.target.family != "aarch64":
```

zackgalbreath (approved):

PR #33817, opened 2022-11-10
traverse: add topo

haampie:
@scheibelp since we talked about this. It's useful in the installer, as well as for removal (currently it screams about the db being in an inconsistent state), and as you suggested: environment modifications (although I'm still not convinced).

Basically gives you 

```python
spack.traverse.traverse_edges([s1, s2], order="topo", direction="parents/children", key=...)
```

and `spack.traverse.traverse_nodes`.

### DIFF
--- a/lib/spack/spack/test/traverse.py
+++ b/lib/spack/spack/test/traverse.py
@@ -25,64 +25,109 @@ def create_dag(nodes, edges):
 
 @pytest.fixture()
 def abstract_specs_dtuse():
-    nodes = [
-        "dtbuild1",
-        "dtbuild2",
-        "dtbuild3",
-        "dtlink1",
-        "dtlink2",
-        "dtlink3",
-        "dtlink4",
-        "dtlink5",
-        "dtrun1",
-        "dtrun2",
-        "dtrun3",
-        "dttop",
-        "dtuse",
-    ]
-    edges = [
-        ("dtbuild1", "dtbuild2", ("build")),
-        ("dtbuild1", "dtlink2", ("build", "link")),
-        ("dtbuild1", "dtrun2", ("run")),
-        ("dtlink1", "dtlink3", ("build", "link")),
-        ("dtlink3", "dtbuild2", ("build")),
-        ("dtlink3", "dtlink4", ("build", "link")),
-        ("dtrun1", "dtlink5", ("build", "link")),
-        ("dtrun1", "dtrun3", ("run")),
-        ("dtrun3", "dtbuild3", ("build")),
-        ("dttop", "dtbuild1", ("build",)),
-        ("dttop", "dtlink1", ("build", "link")),
-        ("dttop", "dtrun1", ("run")),
-        ("dtuse", "dttop", ("build", "link")),
-    ]
-    return create_dag(nodes, edges)
+    return create_dag(
+        nodes=[
+            "dtbuild1",
+            "dtbuild2",
+            "dtbuild3",
+            "dtlink1",
+            "dtlink2",
+            "dtlink3",
+            "dtlink4",
+            "dtlink5",
+            "dtrun1",
+            "dtrun2",
+            "dtrun3",
+            "dttop",
+            "dtuse",
+        ],
+        edges=[
+            ("dtbuild1", "dtbuild2", ("build")),
+            ("dtbuild1", "dtlink2", ("build", "link")),
+            ("dtbuild1", "dtrun2", ("run")),
+            ("dtlink1", "dtlink3", ("build", "link")),
+            ("dtlink3", "dtbuild2", ("build")),
+            ("dtlink3", "dtlink4", ("build", "link")),
+            ("dtrun1", "dtlink5", ("build", "link")),
+            ("dtrun1", "dtrun3", ("run")),
+            ("dtrun3", "dtbuild3", ("build")),
+            ("dttop", "dtbuild1", ("build",)),
+            ("dttop", "dtlink1", ("build", "link")),
+            ("dttop", "dtrun1", ("run")),
+            ("dtuse", "dttop", ("build", "link")),
+        ],
+    )
 
 
 @pytest.fixture()
 def abstract_specs_dt_diamond():
-    nodes = ["dt-diamond", "dt-diamond-left", "dt-diamond-right", "dt-diamond-bottom"]
-    edges = [
-        ("dt-diamond", "dt-diamond-left", ("build", "link")),
-        ("dt-diamond", "dt-diamond-right", ("build", "link")),
-        ("dt-diamond-right", "dt-diamond-bottom", ("build", "link", "run")),
-        ("dt-diamond-left", "dt-diamond-bottom", ("build")),
-    ]
-    return create_dag(nodes, edges)
+    return create_dag(
+        nodes=["dt-diamond", "dt-diamond-left", "dt-diamond-right", "dt-diamond-bottom"],
+        edges=[
+            ("dt-diamond", "dt-diamond-left", ("build", "link")),
+            ("dt-diamond", "dt-diamond-right", ("build", "link")),
+            ("dt-diamond-right", "dt-diamond-bottom", ("build", "link", "run")),
+            ("dt-diamond-left", "dt-diamond-bottom", ("build")),
+        ],
+    )
 
 
 @pytest.fixture()
 def abstract_specs_chain():
     # Chain a -> b -> c -> d with skip connections
     # from a -> c and a -> d.
-    nodes = ["chain-a", "chain-b", "chain-c", "chain-d"]
-    edges = [
-        ("chain-a", "chain-b", ("build", "link")),
-        ("chain-b", "chain-c", ("build", "link")),
-        ("chain-c", "chain-d", ("build", "link")),
-        ("chain-a", "chain-c", ("build", "link")),
-        ("chain-a", "chain-d", ("build", "link")),
-    ]
-    return create_dag(nodes, edges)
+    return create_dag(
+        nodes=["chain-a", "chain-b", "chain-c", "chain-d"],
+        edges=[
+            ("chain-a", "chain-b", ("build", "link")),
+            ("chain-b", "chain-c", ("build", "link")),
+            ("chain-c", "chain-d", ("build", "link")),
+            ("chain-a", "chain-c", ("build", "link")),
+            ("chain-a", "chain-d", ("build", "link")),
+        ],
+    )
+
+
+@pytest.mark.parametrize("direction", ("children", "parents"))
+@pytest.mark.parametrize("deptype", ("all", ("link", "build"), ("run", "link")))
+def test_all_orders_traverse_the_same_nodes(direction, deptype, abstract_specs_dtuse):
+    # Test whether all graph traversal methods visit the same set of vertices.
+    # When testing cover=nodes, the traversal methods may reach the same vertices
+    # through different edges, so we're using traverse_nodes here to only verify the
+    # vertices.
+    #
+    # NOTE: root=False currently means "yield nodes discovered at depth > 0",
+    # meaning that depth first search will yield dtlink5 as it is first found through
+    # dtuse, whereas breadth first search considers dtlink5 at depth 0 and does not
+    # yield it since it is a root. Therefore, we only use root=True.
+    # (The inconsistency cannot be resolved by making root=False mean "don't yield
+    # vertices without in-edges", since this is not how it's used; it's typically used
+    # as "skip the input specs".)
+    specs = [abstract_specs_dtuse["dtuse"], abstract_specs_dtuse["dtlink5"]]
+    kwargs = {"root": True, "direction": direction, "deptype": deptype, "cover": "nodes"}
+
+    def nodes(order):
+        s = traverse.traverse_nodes(specs, order=order, **kwargs)
+        return sorted(list(s))
+
+    assert nodes("pre") == nodes("post") == nodes("breadth") == nodes("topo")
+
+
+@pytest.mark.parametrize("direction", ("children", "parents"))
+@pytest.mark.parametrize("root", (True, False))
+@pytest.mark.parametrize("deptype", ("all", ("link", "build"), ("run", "link")))
+def test_all_orders_traverse_the_same_edges(direction, root, deptype, abstract_specs_dtuse):
+    # Test whether all graph traversal methods visit the same set of edges.
+    # All edges should be returned, including the artificial edges to the input
+    # specs when root=True.
+    specs = [abstract_specs_dtuse["dtuse"], abstract_specs_dtuse["dtlink5"]]
+    kwargs = {"root": root, "direction": direction, "deptype": deptype, "cover": "edges"}
+
+    def edges(order):
+        s = traverse.traverse_edges(specs, order=order, **kwargs)
+        return sorted(list(s))
+
+    assert edges("pre") == edges("post") == edges("breadth") == edges("topo")
 
 
 def test_breadth_first_traversal(abstract_specs_dtuse):
--- a/lib/spack/spack/test/traverse.py
+++ b/lib/spack/spack/test/traverse.py
@@ -284,3 +284,32 @@ chain-a
     ^chain-d
 """
     assert s.tree(depth_first=False, cover="edges", **args) == bfs_tree_edges
+
+
+def test_traverse_topo_order():
+    # Good old Wiki example
+    s = [Spec("s{}".format(i)) for i in range(12)]
+
+    for parent, child in (
+        (5, 11),
+        (7, 11),
+        (7, 8),
+        (3, 8),
+        (3, 10),
+        (11, 2),
+        (11, 9),
+        (11, 10),
+        (8, 9),
+    ):
+        s[parent].add_dependency_edge(s[child], "all")
+
+    # These are *actual* roots (all nodes with no incoming edges)
+    roots = [s[5], s[7], s[3]]
+    nodes = [x.spec for x in traverse.traverse_edges(roots, order="topo")]
+
+    # Ensure the invariant that all parents of node[i] are in node[0:i]
+    # Since we have actual roots, we can just traverse in reverse with a different
+    # traversal method from the one we're testing.
+    for i in range(len(nodes)):
+        parents = list(nodes[i].traverse(order="pre", direction="parents", root=False))
+        assert set(parents).issubset(set(nodes[:i]))
--- a/lib/spack/spack/test/traverse.py
+++ b/lib/spack/spack/test/traverse.py
@@ -289,7 +289,7 @@ chain-a
 @pytest.fixture()
 def abstract_specs_toposort():
     # Create a graph that both BFS and DFS would not traverse in topo order, assuming
-    # edges are traversed by target node name. Roots are {A, E} in forward order and
+    # edges are ordered by target node name. Roots are {A, E} in forward order and
     # {F, G} in backward order.
     # forward: DFS([A, E]) traverses [A, B, F, G, C, D, E] (not topo since C < B)
     # forward: BFS([A, E]) traverses [A, E, B, C, D, F, G] (not topo since C < B)
--- a/lib/spack/spack/test/traverse.py
+++ b/lib/spack/spack/test/traverse.py
@@ -319,15 +319,19 @@ def abstract_specs_toposort():
     )
 
 
-def test_traverse_topo_order(abstract_specs_toposort):
+def test_traverse_topo_nodes(abstract_specs_toposort):
     nodes = abstract_specs_toposort
 
     def test_topo(roots, direction="children"):
         # Ensure the invariant that all parents of specs[i] are in specs[0:i]
-        ordered = list(traverse.traverse_nodes(roots, order="topo", direction=direction))
+        ordered = list(
+            traverse.traverse_nodes(roots, order="topo", cover="nodes", direction=direction)
+        )
         reverse = "parents" if direction == "children" else "children"
         for i in range(len(ordered)):
-            parents = ordered[i].traverse(order="pre", direction=reverse, root=False)
+            parents = ordered[i].traverse(
+                order="pre", cover="nodes", direction=reverse, root=False
+            )
             assert set(list(parents)).issubset(set(ordered[:i]))
 
     # Traverse forward from roots A and E and a non-root D
@@ -335,3 +339,34 @@ def test_traverse_topo_order(abstract_specs_toposort):
 
     # Traverse reverse from leafs F and G and non-leaf D
     test_topo([nodes["F"], nodes["D"], nodes["G"]], direction="parents")
+
+
+def test_traverse_topo_edges(abstract_specs_toposort):
+    nodes = abstract_specs_toposort
+    roots = [nodes["E"], nodes["A"]]
+
+    # Collect pairs of (parent spec name, child spec name)
+    edges = [
+        (e.parent.name, e.spec.name)
+        for e in traverse.traverse_edges(roots, order="topo", cover="edges", root=False)
+    ]
+
+    # See figure above, we have 7 edges (excluding artifical ones to the root)
+    assert set(edges) == set(
+        [
+            ("A", "B"),
+            ("A", "C"),
+            ("B", "F"),
+            ("B", "G"),
+            ("C", "D"),
+            ("D", "B"),
+            ("E", "D"),
+        ]
+    )
+
+    # Verify that all in-edges precede all out-edges
+    for node in nodes.keys():
+        in_edge_indices = [i for (i, (parent, child)) in enumerate(edges) if node == child]
+        out_edge_indices = [i for (i, (parent, child)) in enumerate(edges) if node == parent]
+        if in_edge_indices and out_edge_indices:
+            assert max(in_edge_indices) < min(out_edge_indices)
--- a/lib/spack/spack/test/traverse.py
+++ b/lib/spack/spack/test/traverse.py
@@ -369,15 +369,13 @@ def test_traverse_topo_nodes(abstract_specs_toposort):
 
     def test_topo(roots, direction="children"):
         # Ensure the invariant that all parents of specs[i] are in specs[0:i]
-        ordered = list(
+        specs = list(
             traverse.traverse_nodes(roots, order="topo", cover="nodes", direction=direction)
         )
         reverse = "parents" if direction == "children" else "children"
-        for i in range(len(ordered)):
-            parents = ordered[i].traverse(
-                order="pre", cover="nodes", direction=reverse, root=False
-            )
-            assert set(list(parents)).issubset(set(ordered[:i]))
+        for i in range(len(specs)):
+            parents = specs[i].traverse(cover="nodes", direction=reverse, root=False)
+            assert set(list(parents)).issubset(set(specs[:i]))
 
     # Traverse forward from roots A and E and a non-root D
     test_topo([nodes["D"], nodes["E"], nodes["A"]], direction="children")
--- a/lib/spack/spack/test/traverse.py
+++ b/lib/spack/spack/test/traverse.py
@@ -303,13 +303,15 @@ def test_traverse_topo_order():
     ):
         s[parent].add_dependency_edge(s[child], "all")
 
-    # These are *actual* roots (all nodes with no incoming edges)
-    roots = [s[5], s[7], s[3]]
-    nodes = [x.spec for x in traverse.traverse_edges(roots, order="topo")]
+    def is_topo(ordered_specs):
+        # Ensure the invariant that all parents of specs[i] are in specs[0:i]
+        specs = [s for s in ordered_specs]
+        for i in range(len(specs)):
+            parents = specs[i].traverse(order="pre", direction="parents", root=False)
+            assert set(list(parents)).issubset(set(specs[:i]))
 
-    # Ensure the invariant that all parents of node[i] are in node[0:i]
-    # Since we have actual roots, we can just traverse in reverse with a different
-    # traversal method from the one we're testing.
-    for i in range(len(nodes)):
-        parents = list(nodes[i].traverse(order="pre", direction="parents", root=False))
-        assert set(parents).issubset(set(nodes[:i]))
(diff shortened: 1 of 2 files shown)


lib/spack/spack/test/traverse.py
```diff
@@ -209,3 +209,47 @@ def test_breadth_first_versus_depth_first_printing(config, mock_packages):
...
+
+def test_traverse_topo_order():
+    # All paths (5, 7 and 3 spawn the entire DAG)
+    # s05 -> s11 -> s02
+    #        s11 -> s09
+    #        s11 -> s10
+    # s07 -> s11
+    # s07 -> s08 -> s09
+    # s03 -> s08
+    # s03        -> s10
```
scheibelp: This should be represented as a more-traditional ascii graph (i.e. each node appears once). I assume that was not done here because it is painful. I wonder if `spack.graph.AsciiGraph` could be generalized for this purpose; in this particular case I don't think you would need to modify the function at all though, because you create the DAG using `Spec` objects (which is what `graph` wants).

lib/spack/spack/test/traverse.py
```diff
@@ -209,3 +209,47 @@ def test_breadth_first_versus_depth_first_printing(config, mock_packages):
...
+        (11, 9),
+        (11, 10),
+        (8, 9),
+    ):
+        s[parent].add_dependency_edge(s[child], "all")
+
+    def test_topo(roots, direction="children"):
+        # Ensure the invariant that all parents of specs[i] are in specs[0:i]
+        ordered_specs = traverse.traverse_nodes(roots, order="topo", direction=direction)
+        specs = [s for s in ordered_specs]
```
scheibelp: (minor) I'm assuming you're doing this to turn a generator into a re-traverse-able object: it's more idiomatic to do that like `specs = list(ordered_specs)` but IMO clearer to fold that into the prior line (to avoid more variable names).

lib/spack/spack/test/traverse.py
```diff
@@ -209,3 +209,47 @@ def test_breadth_first_versus_depth_first_printing(config, mock_packages):
...
+        reverse = "parents" if direction == "children" else "children"
+        for i in range(len(specs)):
+            parents = specs[i].traverse(order="pre", direction=reverse, root=False)
```
scheibelp: Why set `order`? In general, if a parameter's value shouldn't matter, it shouldn't be set.

lib/spack/spack/test/traverse.py
```diff
@@ -209,3 +209,47 @@ def test_breadth_first_versus_depth_first_printing(config, mock_packages):
...
+        (7, 11),
+        (7, 8),
+        (3, 8),
+        (3, 10),
+        (11, 2),
+        (11, 9),
+        (11, 10),
+        (8, 9),
+    ):
+        s[parent].add_dependency_edge(s[child], "all")
```
scheibelp: This looks similar to https://en.wikipedia.org/wiki/Topological_sorting, I'll note that IMO this doesn't sufficiently cover the types of cases we want to get right with topological ordering; in fact I don't think it contains any subgraphs that BFS wouldn't traverse in the same way (in particular if you started from a "pseudo-root" that attached to all the disconnected roots. An example a subgraph that ought to be in this test:

```
0
| \
1 |
| /
2 
```

i.e. there should be "shortcuts" to child nodes. Ideally this type of structure would be common (and repeated a few times at different scales) in the graph used for testing.

lib/spack/spack/traverse.py
```diff
@@ -110,6 +110,32 @@ def neighbors(self, node):
         return self.visitor.neighbors(node)
 
 
+class TopoVisitor(object):
```
scheibelp, line 113: It is unclear to me how this is achieving a topological order. I think the docstring should clarify that (it's probably worth talking about the general strategy here first though). If I had to guess, I'd say it's based on the depth-first-derived strategy discussed in https://en.wikipedia.org/wiki/Topological_sorting; it's unclear to me how composing visitors achieves that though.

lib/spack/spack/traverse.py
```diff
@@ -217,6 +243,25 @@ def traverse_breadth_first_with_visitor(specs, visitor):
             queue.append(EdgeAndDepth(edge, node.depth + 1))
 
 
+def traverse_depth_first_with_visitor(nodes, visitor):
+    # This is a somewhat non-standard implementation, but the reason to start with
+    # edges is that we don't have to deal with an artificial root node when doing DFS
+    # on multiple (root) specs.
```
scheibelp: I'm not sure what this comment applies to: the function starts by iterating over nodes. Also I think this should explain what the "standard implementation" is (or for example it could be worded like "instead of doing X, do Y because...").

lib/spack/spack/traverse.py
```diff
@@ -298,6 +353,7 @@ def traverse_edges(
         root (bool): Yield the root nodes themselves
         order (str): What order of traversal to use in the DAG. For depth-first
             search this can be ``pre`` or ``post``. For BFS this should be ``breadth``.
```
scheibelp, line 423: (question) what was the original purpose of being able to specify the difference between a pre-order and post-order DFS traversal? (I assume you don't have the original context, I want to resolve this though before merging the PR and I'll need to track other folks down).

lib/spack/spack/traverse.py
```diff
@@ -110,6 +110,32 @@ def neighbors(self, node):
...
+        return self.key(node.edge.spec) not in self.visited
+
+    def pre(self, node):
+        # You could add a temporary marker for cycle detection
+        # that's cleared in `post`, but we assume no cycles.
+        pass
+
+    def post(self, node):
+        self.visited.add(self.key(node.edge.spec))
+        self.topo_order.append(node.edge)
```
scheibelp: If I understand correctly, these are `EdgeAndDepth` instances. This isn't strictly an idiom established by this PR, but calling these `node`s is confusing to me, is there any way that could be changed? (if so, perhaps that ought to happen in a separate PR)

lib/spack/spack/traverse.py
```diff
@@ -275,6 +320,16 @@ def traverse_breadth_first_tree_nodes(parent_id, edges, key=id, depth=0):
             yield item
 
 
+# Topologic order
+def traverse_edges_topo(specs, direction="children", deptype="all", key=id):
```
scheibelp: This should have a docstring that explains `direction="children"` means that parents are ordered before children. I'm inclined to think the parameter should be renamed though: to something like `prioritize="parents"`.

lib/spack/spack/traverse.py
```diff
@@ -320,6 +376,11 @@ def traverse_edges(
         A generator that yields ``DependencySpec`` if depth is ``False``
         or a tuple of ``(depth, DependencySpec)`` if depth is ``True``.
     """
+
+    # TODO: cover=edges for order=topo? does visited make sense?
```
scheibelp: For now, I think it makes sense to prohibit `cover="edges"` for `order="topo"`: I don't know of an interpretation of topological ordering which applies for edges.

scheibelp (changes requested):
At a high-level:

* The implementation needs documentation (I'm not clear how composing visitors achieves a topological ordering)
* The test graph needs to be more complex
* Given that pre/post order relate to DFS, I want to check the expectations of callers of `traverse` in the code: did they actually mean "topo" when they were asking for pre/post? This isn't something you need to investigate yourself (although that would be helpful)

lib/spack/spack/test/traverse.py
haampie: It matters cause it shouldn't be `topo`

lib/spack/spack/traverse.py
haampie: In textbook dfs you pass a vertex not a list of edges

lib/spack/spack/test/traverse.py
haampie: Nice, it's a bit unfortunate that it works on a single Spec instead of actually accepting all nodes or at least multiple roots that induce the full dag.

lib/spack/spack/test/traverse.py
haampie: I see what you mean, but the example above would not be a failing test case for bfs either ;p

lib/spack/spack/traverse.py
haampie: Hm, I thought it was reasonably clear (traverse all possible paths from A -> B before pushing A), but I can add a comment. Edit: added the reference

lib/spack/spack/traverse.py
haampie: I don't agree with renaming it. Our vertices are specs, and the edges are directed from parent to child or dependent to dependency. Then this topo traversal code follows the definition of topo order. `direction="parents"` is there to flip the edges, and it's consistent with traverse_edges(...) which we've had forever.

lib/spack/spack/traverse.py
haampie: Typically for undoing actions, for example pushing (in pre) and popping (in post) items on a stack.

Edit: naively I thought we make environment variable changes in post order and then revert those changes in pre order, but `reverse(pre) != post`, so it would not always be correct.

lib/spack/spack/traverse.py
haampie: It's ~not hard~ may not be useful to define? For a set of edges `{u -> v}`, define the binary operator `(s -> t) < (u -> v)` as `t  < u` where `<` induces the topo order on the set of vertices.

lib/spack/spack/traverse.py
haampie: It's a remnant of refactoring I think. Going from a function that took a vertex and then looped over its neighboring vertices, to a function that took a list of neighbors :thinking:

haampie: Feel free to just submit a pr with these small things next time, then I can already merge it the next morning and rebase my work on it.

lib/spack/spack/test/traverse.py
```diff
@@ -284,3 +284,54 @@ def test_breadth_first_versus_depth_first_printing(abstract_specs_chain):
     ^chain-d
 """
     assert s.tree(depth_first=False, cover="edges", **args) == bfs_tree_edges
+
+
+@pytest.fixture()
+def abstract_specs_toposort():
+    # Create a graph that both BFS and DFS would not traverse in topo order, assuming
+    # edges are traversed by target node name. Roots are {A, E} in forward order and
```
haampie:
```suggestion
    # edges are ordered by target node name. Roots are {A, E} in forward order and
```

haampie:
@scheibelp my main Q is whether it fits the `traverse_edges` interface enough. I don't think we need `cover=edges` necessarily, but one way to think about it is: we yield all edges of the DAG, in such an order that all in-edges of a vertex are yielded before all out-edges.

Edit: I just pushed that.

lib/spack/spack/traverse.py
haampie: In the latest commits `cover=edges` + `order=topo` generates all edges of the DAG ordered s.t. all in-edges per vertex precede its out-edges.

lib/spack/spack/traverse.py
scheibelp: Not renaming is fine. In that case, the docstring should explain the meaning of `direction="children"`

lib/spack/spack/traverse.py
```diff
@@ -96,18 +96,80 @@ def __init__(self, visitor, key=id, visited=None):
...
+    vertices and edges).
+    """
+
+    def __init__(self, visitor, key=id, root=True, all_edges=False):
+        """
+        Arguments:
+            visitor: visitor that implements accept(), pre(), post() and neighbors()
+            key: uniqueness key for nodes
+            root (bool): Whether to include the root node.
+            all_edges (bool): when False, list ever node once in topo order.
```
scheibelp: Typo: "ever" -> "every"

lib/spack/spack/traverse.py
```diff
@@ -96,18 +96,80 @@ def __init__(self, visitor, key=id, visited=None):
...
+
+class TopoVisitor(object):
+    """Visitor that can be used in depth-first search to generate
+    a topological ordered list of vertices (we keep track of edges).
+
+    Algorithm based on "Section 22.4: Topological sort", Introduction to Algorithms
+    (2001, 2nd edition) by Cormen, Thomas H.; Leiserson, Charles E.; Rivest, Ronald L.;
+    Stein, Clifford.
+
+    The algorithm in plain English: realize that (1) we only append a vertex A before
```
scheibelp: Should "we only append a vertex A before" be "we only append a vertex A *after*"?

lib/spack/spack/test/traverse.py
scheibelp: I think you mean you don't want to rely on "topo" to retrieve the parents while testing the ordering of topo (`topo` should only affect the ordering: any traversal of a node - `pre`, `post`, `topo`, or `bfs` - should produce the same set of nodes).

If I got that right, is there a test that exists which confirms that all the ordering parameters produce the same set of parents? I think that would be useful (and in that case this test wouldn't have to specify the ordering).

lib/spack/spack/traverse.py
scheibelp: I'm generally reluctant to implement something we have no known use case for, but does not implementing this cause some other issue (i.e. a failure since the code using this makes assumptions about what is available in the interface)?

lib/spack/spack/traverse.py
```diff
@@ -96,18 +96,80 @@ def __init__(self, visitor, key=id, visited=None):
...
+    traversing all possible paths to any descendant B (that means A is always appended
+    to reverse_order after all its descendants) and (2) while following paths from
+    A -> B, we push B to reverse_order the first time it is reached, and guarantee its
+    ancestors are always appended and its descendants are not (on the first path
+    this is thanks to backtracking, on any other path we do not follow B as its
+    already seen.)
+
+    The neat thing about this algorithm is that it can operate on an immutable DAG,
+    it doesn't require marking or popping edges, and it works in linear time (in
+    vertices and edges).
```
scheibelp: > The neat thing about this algorithm is that it can operate on an immutable DAG, it doesn't require marking or popping edges, and it works in linear time (in vertices and edges).

This particular statement is hard to understand in this context: it is useful for comparing implementations of topological sort, but we only have one here. Also AFAIK the `visited` set effectively functions as a form of marking, so I think whether this statement applies is also debatable. Either way, I think the docstring is clearer without it.

lib/spack/spack/traverse.py
```diff
@@ -205,16 +275,39 @@ def traverse_breadth_first_with_visitor(specs, visitor):
...
+
+def traverse_depth_first_with_visitor(edges, visitor):
+    """Traverse a DAG in depth-first fashion using a visitor, starting from
+    a list of edges. Note that typically DFS would take a vertex not a list of edges,
+    but the API is like this so we don't have to create an artificial root node when
+    traversing from multiple roots in a DAG.
+
+    Arguments:
+        edges (list): List of EdgeAndDepth instances
+        visitor: class instance implementing accept(), pre() and neigbors()
```
scheibelp: And `post()`

IMO this nested composition of visitors in tandem with specialized user methods is more confusing than an "all-in-one" implementation of topological sorting: having to combine the behavior of this method with the implementation of `TopoVisitor` is confusing to me (on that note, I think `TopoVisitor` should specify that it produces a topological ordering when combined with this function, so that future readers know to examine both when trying to determine the semantics).

lib/spack/spack/traverse.py
scheibelp: I think the following description would be clearer, but I'm happy to iterate on what's here:

Perform a depth-first search; append a node only after visiting all of its children. This produces a list where the children of a node are always guaranteed to come before a node. This list can be reversed to produce a topological ordering.

Again, this depends on being used in tandem with `traverse_depth_first_with_visitor`, and to that end I think it would be clearer if this class were defined inside of the method body of `traverse_edges_topo`.

scheibelp (commented):
Additional comments

lib/spack/spack/test/traverse.py
haampie: Actually another reason why I did this is because `root=True/False` was not implemented, but it is now.

lib/spack/spack/traverse.py
haampie: I don't agree with the latter, it's not an exposed class in the module, but for advanced use cases you might want to override parts of it, like `neighbors`. So, not public API, but it can be useful nonetheless.

lib/spack/spack/traverse.py
haampie: Hm yeah, let me just drop it. It was in contrast to the Kahn algorithm where it's common delete edges, but even that can run in linear time if you keep a dict of `node id` -> `# in-edges`.

Btw, notice that if we would do Kahn instead of this DFS/backtracking algorithm, we would still need to do one pass with BFS/DFS to actually get the `# in-edges` count of the sub-DAG we're traversing over. So I think this DFS algorithm is typically more efficient.

lib/spack/spack/traverse.py
haampie: I get your point, but I disagree, because there are instances where the `traverse_nodes` interface is too limiting, for example because `deptype` is not constant but depends on depth (drop build edges at depth > 0) or the properties of the node. Also, the composability aspect is nice, and in principle it's possible to unit test the behavior of a visitor (for instance one that has a stateful `neighbors()` implementation).

lib/spack/spack/traverse.py
```diff
@@ -96,18 +96,75 @@ def __init__(self, visitor, key=id, visited=None):
...
 
         self.visited.add(key)
-        return self.visitor.neighbors(node)
+        return self.visitor.neighbors(item)
+
+
+class TopoVisitor(object):
+    """Visitor that can be used in :py:func:`depth-first traversal
+    <spack.traverse.traverse_depth_first_with_visitor>` to generate
+    a topologically ordered list of edges (in-edges before out-edges).
```
scheibelp: Note that by itself, the constraint "in-edges before out-edges" does not produce a topological order:

```
0
|\
1 |
|/
3
```

`[0, 3], [0, 1], [1, 3]` satisfies that but is not a valid topological ordering. The comment should probably first introduce the traditional concept of topological ordering and then extend it to edges.

This is partly my concern with generalizing the notion of topological ordering to edges: suddenly the documentation speaks in terms that are unfamiliar and need to be reasoned about, vs. concepts which are immediately accessible (e.g. via a Wikipedia article) - reorganizing the docstring would help with this though.

haampie: I agree. I'm not sure how useful it is to allow cover=edges at this point. You do have the property that `map(lambda s: s.parent, traverse_edges(order="topo))` is topo ordered, but yeah, that does not include all reachable vertices.

scheibelp: Sorry to be clear I trusted that this produces a topological order (or that one can be extracted from the list of edges), my concern is that it will be hard for future maintainers to understand it; the invariant stated in the docstring is insufficient to guarantee a topological ordering (the algorithm itself does more than that though), which I think is an example of how it's challenging to describe the semantics.

Are you inclined to undo the support for `cover=edges`, or would you prefer to improve the docstrings? Either way is fine with me.

haampie: I just updated the docstrings and kept `cover=edges`

lib/spack/spack/test/traverse.py
```diff
@@ -284,3 +329,87 @@ def test_breadth_first_versus_depth_first_printing(abstract_specs_chain):
...
+            assert set(list(parents)).issubset(set(specs[:i]))
+
+    # Traverse forward from roots A and E and a non-root D
+    test_topo([nodes["D"], nodes["E"], nodes["A"]], direction="children")
+
+    # Traverse reverse from leafs F and G and non-leaf D
+    test_topo([nodes["F"], nodes["D"], nodes["G"]], direction="parents")
+
+
+def test_traverse_topo_edges(abstract_specs_toposort):
```
scheibelp: Add comment explaining what this is testing (i.e. this is just testing the in-edges/out-edges ordering on a per-node basis, and the other test is responsible for verifying the traditional definition of a topological ordering).

lib/spack/spack/test/traverse.py
```diff
@@ -284,3 +329,87 @@ def test_breadth_first_versus_depth_first_printing(abstract_specs_chain):
...
 """
     assert s.tree(depth_first=False, cover="edges", **args) == bfs_tree_edges
+
+
+@pytest.fixture()
+def abstract_specs_toposort():
+    # Create a graph that both BFS and DFS would not traverse in topo order, assuming
+    # edges are ordered by target node name. Roots are {A, E} in forward order and
+    # {F, G} in backward order.
+    # forward: DFS([A, E]) traverses [A, B, F, G, C, D, E] (not topo since C < B)
```
scheibelp, line 339: I don't think DFS specifies whether C or B would be visited first from A, so it is more accurate to say that this is a valid DFS visitation order but not a valid topological order. I think this is important to mention because it clarifies that this is hard to test: I'm not sure if we can make a graph where DFS is *guaranteed* to produce an order that is not topological, so in that sense this test is a heuristic.

lib/spack/spack/test/traverse.py
```diff
@@ -284,3 +329,87 @@ def test_breadth_first_versus_depth_first_printing(abstract_specs_chain):
...
+        specs = list(
+            traverse.traverse_nodes(roots, order="topo", cover="nodes", direction=direction)
+        )
+        reverse = "parents" if direction == "children" else "children"
+        for i in range(len(specs)):
+            parents = specs[i].traverse(cover="nodes", direction=reverse, root=False)
+            assert set(list(parents)).issubset(set(specs[:i]))
+
+    # Traverse forward from roots A and E and a non-root D
+    test_topo([nodes["D"], nodes["E"], nodes["A"]], direction="children")
```
scheibelp, line 385: I think it's worth mentioning that if some root specs are children of others, the resulting traversal is effectively the same as if they had not been included [1]. This might be better to put in the docstring of `traverse_nodes` than here.

1. Except for depth: Reading the docstring of `traverse_nodes`, it mentions that the roots are considered depth 0: what does that mean in this example where D is a child of A (i.e., what is the depth of B)? That conceptual difficulty is not strictly related to any of the changes here, but I bring it up because I had to spend some time thinking about what it means for a topological ordering when some roots are children of others; I think it works fine here but introduces potential extra complexity for future maintainers.

lib/spack/spack/traverse.py
scheibelp: Thanks for checking into that! Any issues with pre/post elsewhere can be resolved separately and I consider this chain resolved for the purposes of this PR.

lib/spack/spack/traverse.py
```diff
@@ -320,17 +444,31 @@ def traverse_edges(
...
         or a tuple of ``(depth, DependencySpec)`` if depth is ``True``.
     """
-    root_edges = root_specs(specs)
+
+    if order == "topo":
+        if cover == "paths":
+            raise ValueError("cover=paths not supported for order=topo")
+        # For topo order it's somewhat unclear how to handle a pre-existing visited
+        # set. Exclude them? But what if excludes vertices have edges to non-excluded
+        # ones? Not following would violate topo order.
```
scheibelp: In this case I assume that all edges (both in and out) involving an excluded vertex would be discarded. I think the issue is not that we cannot define a reasonable interpretation, but that we have no use cases for topological traversals of edges.

Therefore, I think this comment should be rephrased:

```
# TODO: There is no known need for topological ordering of traversals (edge or node)
# with an initialized "visited" set. Revisit if needed.

File: lib/spack/spack/traverse.py [hunk 21]
```diff
@@ -275,6 +370,34 @@ def traverse_breadth_first_tree_nodes(parent_id, edges, key=id, depth=0):
             yield item
 
 
+# Topologic order
+def traverse_edges_topo(
+    specs, direction="children", deptype="all", key=id, root=True, all_edges=False
+):
+    """
+    Returns a list of edges where ordered in topo order, in the sense that all
```
scheibelp: (phrasing) 

I think

> Returns a list of edges where ordered in topo order...

should be 

> Returns a list of edges in topological order...

scheibelp (changes requested):
The remaining requests are all for documentation (although in one case I suggest that we may want to constrain the set of roots allowed for `traverse_nodes`).

lib/spack/spack/test/traverse.py
haampie: It is actually an invariant of `traverse_nodes(order="pre/post")` to order edges by target package name. The comment hints on that. I can reword it to make that more clear.

lib/spack/spack/test/traverse.py
haampie: They shouldn't be called roots if they have in-degree != 0 :man_shrugging: that's also how I'm phrasing it in the comment, but I can expand the comment. In `traverse_nodes` the positional argument is called `specs` not `roots`. It's just a bit unfortunate we have a kwarg named `root=True`, and it's always used to mean "skip the traversal starting point"; maybe `skip_start=True/False` was better. I guess `specs` is more like an ordered list of children of an artificial root (and that root induces the sub-dag of interest), so even then the input specs are not roots :p.

Notice though that in most use cases before we did something like this:

```python
visited = set()
for s in specs:
  for t in s.traverse(root=False, visited=visited):
    # will give you specs[1] if there's an edge specs[0] -> specs[1]
```

which is now equivalent to

```python
for t in traverse(specs, root=False):
    # will give you specs[1] if there's an edge specs[0] -> specs[1]
```

So, the "potential extra complexity" was already an issue before.

That complexity is gone when using `order=breadth`, since then you know for a fact that non-root input specs are discovered at depth 0.

haampie: So, I think I can clarify that elsewhere, but maybe in a separate PR. For now I've just updated the relevant comment in the test.

haampie: Aargh, I realize I *did* use the word `roots` in the test. Will fix that too.

lib/spack/spack/test/traverse.py
scheibelp: That is good to know: I think this test is much better than I originally thought. However, I don't see any mention of this invariant in the docstring for `traverse_nodes` or in `BaseVisitor.neighbors` (which appears to be the source of this invariant).

haampie: See also https://github.com/spack/spack/pull/33406#discussion_r1007822909 and the linked PR https://github.com/spack/spack/pull/28673#discussion_r823132555. If I remember correctly, the idea was that this was an invariant thanks to the Spec objects themselves (EdgeMap is always sorted), but in practice even the old traversal code uses an explicit sort, that I didn't remove when refactoring to visitors.

scheibelp: I agree that depending on the sorting properties of the edges is sufficient. I think it may be an issue though that we don't express that we use this property to guarantee a fixed, repeatable ordering of edges/nodes. This issue in documentation may predate this PR. It doesn't need to be fixed here.

lib/spack/spack/test/traverse.py
scheibelp: > So, the "potential extra complexity" was already an issue before.

I wouldn't consider that the same kind of complexity: I was worried about semantics within a single traversal. Multiple traversals might add other complications.

> That complexity is gone when using order=breadth, since then you know for a fact that non-root input specs are discovered at depth 0.

To clarify, my specific example was not about a non-root input spec, but rather a spec which is a child of two input specs, appearing at different depths WRT each. Nevertheless, I think I see how BFS produces a well-defined (but meaningless) value for the depth of B when `specs=[A, D]`. However, since we allow `specs=[A, D]` for topo,  it should also make sense for that, or it should be forbidden (the latter being my preference). 

Overall, I'm noticing a trend where allowing `specs` to contain nodes that are children of other nodes in `specs` is making it harder to reason about the semantics; I think forbidding this is appealing because I don't know of any use case that requires it (and supporting it seems to add complexity).

> They shouldn't be called roots if they have in-degree != 0

I'm fine with not calling them roots (although don't really see why it's an issue to call them that) although preferably something more descriptive than `specs`.

> In traverse_nodes the positional argument is called specs not roots.
> Aargh, I realize I did use the word roots in the test. Will fix that too.

And in the docstring for traverse_nodes itself:

`specs (list): List of root specs (considered to be depth 0)`

they may not be roots in a strict sense, but the term "root" captures what we intend to do with it in a single (and short) word.

lib/spack/spack/test/traverse.py
scheibelp: (question) would it also be required to sort `traverse.root_specs` to get a fixed ordering?

lib/spack/spack/test/traverse.py
haampie: I don't really get the point you're making. In environments you may want to pass a list of env roots to topo traverse, in uninstall you wanna pass a list of specs matching a db query, etc. You don't know if one is a descendant of the other, that requires traversal. Forbidding it is only annoying. What we care about is topo traversal of the subdag induced by the list of input specs.

lib/spack/spack/test/traverse.py
haampie: I don't want to sort them, since there are cases where there's a natural order such as with environment roots

lib/spack/spack/test/traverse.py
scheibelp: > in uninstall you wanna pass a list of specs matching a db query

That is an excellent point (i.e. some specs might be transitive children of others). It's kind of a headache thinking about the meaning of multi-root traversals for the different traversal methods, but given this example it is certainly worthwhile for topological ordering.

lib/spack/spack/test/traverse.py
scheibelp: This isn't strictly problematic since a user could sort this themselves before passing it in, but complicates the interface since its documentation should mention that the input specs should be pre-sorted to get a fixed ordering. I didn't anticipate a need for users to control traversal orders by sorting the specs themselves (is there a reason for example that users would want the environment spec listing order to influence the traversal order?).

scheibelp (approved):

PR #33829, opened 2022-11-10
warn about removal of deprecated format strings

becker33:
Removal scheduled for v0.20

Should have been scheduled long ago, but we kept configs in this format for too long.

### DIFF
--- a/etc/spack/defaults/config.yaml
+++ b/etc/spack/defaults/config.yaml
@@ -19,7 +19,7 @@ config:
   install_tree:
     root: $spack/opt/spack
     projections:
-      all: "${ARCHITECTURE}/${COMPILERNAME}-${COMPILERVER}/${PACKAGE}-${VERSION}-${HASH}"
+      all: "${architecture}/${compiler.name}-${compiler.version}/${name}-${version}-${hash}"
     # install_tree can include an optional padded length (int or boolean)
     # default is False (do not pad)
     # if padded_length is True, Spack will pad as close to the system max path
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -4384,7 +4384,8 @@ class Spec(object):
         TODO: allow, e.g., ``$6#`` to customize short hash length
         TODO: allow, e.g., ``$//`` for full hash.
         """
-
+        tty.warn("Using the old Spec.format method."
+                 " This method was deprecated in Spack v0.15 and will be removed in Spack v0.20")
         color = kwargs.get("color", False)
 
         # Dictionary of transformations for named tokens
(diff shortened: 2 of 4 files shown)


becker33:
@spackbot fix style

becker33:
@spackbot help

alalazo:
We should also backport this in v0.19.1 so we can claim the warning was in v0.19 (before removing the support in v0.20)

alalazo (approved):
I wonder if we can give a hint to users saying where they are using the old format, but not essential if it requires lot of boilerplate code.

alalazo:
@scottwittenburg @zackgalbreath Never seen this one before. The status check for pipelines says "Errored", but the pipeline succeeded. I'm not retriggering it in case you want to inspect better.

scottwittenburg:
I've never seen this before either.  Manually querying the gitlab api for the pipeline status just now, I see:

```
$ curl \
>   --header "PRIVATE-TOKEN: ${GITLAB_TOKEN}" \
>   https://gitlab.spack.io/api/v4/projects/2/pipelines?ref=pr33829_deprecations/deprecated-format-format
[{
    "id":214974,
    "iid":213380,
    "project_id":2,
    "sha":"faff64929cbb8a6241e9b258e0eac7967c2ce65e",
    "ref":"pr33829_deprecations/deprecated-format-format",
    "status":"success",
    "source":"push", 
    "created_at":"2022-11-11T17:51:58.164Z",
    "updated_at":"2022-11-13T20:35:08.921Z",
    "web_url":"https://gitlab.spack.io/spack/spack/-/pipelines/214974"
}]
```

So obviously gitlab thinks the pipeline was successful.  I'm speculating now that one of those spurious "no specs to rebuild" failures was retried 2 days after the pipeline was created (difference between the created and update timestamps) and succeeded at that point.  The only way the sync cron job could have missed it, is if it didn't run for a period of time around that `update_at` time, and then the next time it *did* run, the updated time was outside the normal pipeline query window.  

Any ideas @zackgalbreath?

Thanks for the heads up before re-running @alalazo, that helped to make sure new pipeline status didn't get created before I could take a look.  But I don't think there's anything more to be learned from logs at this point, so I'll just re-trigger.

scottwittenburg:
@spackbot run pipeline

becker33:
> I wonder if we can give a hint to users saying where they are using the old format, but not essential if it requires lot of boilerplate code.

I don't think most locations preserve the ruamel line-attributions through to where this is called, so I think it would be pretty plumbing-intensive to do

alalazo:
Sorry Greg, I'm going to revert this one. I didn't test drive it, and on `develop` we get:
Screenshot from 2022-11-22 12-29-55

PR #33832, opened 2022-11-11
jellyfish: add variants for python and ruby bindings

bernhardkaindl:
@teachers-uk-net:

Co-authored-by: teachers-uk-net <stuart.morrison@kcl.ac.uk>

Replace #31430 with this improved PR. I could not get the perl extends/variant working without AssertionErrors and even a currupted spack database, therefore I leave the perl bindings enabled by default like before.

The configure_args function had the issue when no variant was selected, args was not defined.

Fixed the conflict between the the extends for ruby and python with:
conflicts("+ruby", when="+python")

Added version 2.3.0 but 2.2.7 is still preferred to not cause anly problems.

Formatted using black which is now the format requested in spack.

### DIFF
--- a/var/spack/repos/builtin/packages/jellyfish/package.py
+++ b/var/spack/repos/builtin/packages/jellyfish/package.py
@@ -7,21 +7,41 @@ from spack.package import *
 
 
 class Jellyfish(AutotoolsPackage):
-    """JELLYFISH is a tool for fast, memory-efficient counting of k-mers in
-    DNA."""
+    """JELLYFISH is a tool for fast, memory-efficient counting of k-mers in DNA."""
 
     homepage = "https://www.cbcb.umd.edu/software/jellyfish/"
-    url = "https://github.com/gmarcais/Jellyfish/releases/download/v2.2.7/jellyfish-2.2.7.tar.gz"
-    list_url = "https://www.cbcb.umd.edu/software/jellyfish/"
+    url = "https://github.com/gmarcais/Jellyfish/releases/download/v2.3.0/jellyfish-2.3.0.tar.gz"
+    list_url = "https://github.com/gmarcais/Jellyfish/releases"
 
-    version("2.2.7", sha256="d80420b4924aa0119353a5b704f923863abc802e94efeb531593147c13e631a8")
+    version("2.3.0", sha256="e195b7cf7ba42a90e5e112c0ed27894cd7ac864476dc5fb45ab169f5b930ea5a")
+    version(
+        "2.2.7",
+        sha256="d80420b4924aa0119353a5b704f923863abc802e94efeb531593147c13e631a8",
+        preferred=True,
+    )
     version(
         "1.1.11",
         sha256="496645d96b08ba35db1f856d857a159798c73cbc1eccb852ef1b253d1678c8e2",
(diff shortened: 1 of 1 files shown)


teachers-uk-net:
Hi

Another pair of eyes is always welcome, I had let it go rather stale mainly to look at the difference between extends and depends_on (there are no assertion errors using depends_on).

I can see why there are assertion errors with extends although in the production version I am using I reverted the extends to depends_on although did not push it until I was more cognisant of the reason for the original suggestion to make these extends. It is on my todo list to look into it in more detail, however, would welcome any insight into those differences :smile: @alalazo @bernhardkaindl 

Kind regards

alalazo (dismissed):
If @teachers-uk-net is fine with merging this, it LGTM

bernhardkaindl:
@spackbot run-pipeline

bernhardkaindl:
@teachers-uk-net Spack dropped support for python EOL versions (python2):

CI no longer accepts python2 dependencies, see #33898 for more information.

For now I removed the dependency on python2 and leave it to the user to provide python2 by the system.

Otherwise the python bindings could be converted to python3, e.g. there is python2to3, if you need it.

var/spack/repos/builtin/packages/jellyfish/package.py
```diff
@@ -7,21 +7,41 @@
...
         url="https://www.cbcb.umd.edu/software/jellyfish/jellyfish-1.1.11.tar.gz",
     )
 
-    depends_on("perl", type=("build", "run"))
-    depends_on("python", type=("build", "run"))
+    depends_on("perl", when="@2.2.7:", type=("build", "run"))
+    # Spack dropped support for python2 and CI no longer accepts depends on python2:
+    # The bindings would have to be updated/converted to python3 or use external python.
+    # For now this only adds --enable-python-binding and you have to provide python2:
+    variant("python", default=False, description="Enable python bindings")
```
alalazo: I think we should remove the variant, if this package depends on Python 2 only, and update the recipe.  @adamjstewart Do you have opinions on this?

adamjstewart: Yes, I would drop this variant.

bernhardkaindl: Thanks, pushed dropping the variant for python2 bindings.

alalazo:
@spackbot run pipeline

alalazo (approved):

adamjstewart:
> Added version 2.3.0 but 2.2.7 is still preferred to not cause anly problems.

Why?

bernhardkaindl:
> > Added version 2.3.0 but 2.2.7 is still preferred to not cause anly problems.
> 
> Why?

Because can't say how compatible or how stable that `'.0'` release is.

@teachers-uk-net didn't add it himself, when he drafted the base for this PR, and he must have had reasons.

I simply don't have the knowledge on if that version would work for his users and he wasn't active lately, so I'm not going to break his package if have a choice to not break it.

adamjstewart:
Isn't that true for all software? Out of the close to 7K packages in Spack, less than 100 prefer a version that isn't the latest version. If you aren't aware of any issues with the latest release, we should prefer that for its bug fixes and new features.

bernhardkaindl:
@spackbot run pipeline

PR #33835, opened 2022-11-11
Openvdb package update v10.0.0

eloop:


bernhardkaindl (approved):
Builds, 8.0.1 didn't build for me, so definitely an improvement ;-)

bernhardkaindl:
@eloop: It takes a lot of Memory, CPU and Disk to build it. I saw that it builds with -g, I it would reduce all of them if using -g0 (or no -g) in the CXXFLAGS to disable generating debug infos.

eloop:
@bernhardkaindl I'm using the default build_type provided by the CMakePackage, you can e.g. do something like

`spack install -v openvdb build_type=MinSizeRel`

"spack info openvdb" will give you other possibilities.

PR #33836, opened 2022-11-11
Octopus : Separate serial and MPI dependencies

iamashwin99:
Octopus can be compiled with and without MPI, so this PR removes MPI as a dependency, introduces it as a variant, and chooses the appropriate compilers based on the variant chosen.
The PR also compiles fftw, libvdwxc with and without MPI.
Since the previous package had MPI as a dependency and hence only MPI variant was built, the new package keeps MPI variant as the default choice. 
To install MPI variant (default), one would just install it as before:
```bash
spack install octopus  #+others
```
or explicitly:
```bash
spack install octopus +mpi #+others
```
To install the serial variant one would try:
```bash
spack install octopus ~mpi 
```


fangohr (approved):
In the past, the octopus package could not be built without MPI. This PR addresses that.

The changes have been discussed and approved by members of the Octopus development team.

tldahlgren (approved):
LGTM

tldahlgren:
Going to `Update branch` to trigger re-testing.

iamashwin99:
are the failed builds due to errors in CI?

tldahlgren:
> are the failed builds due to errors in CI?

It looked like the pipeline didn't trigger properly after switching from a draft PR.   I don't recall if there was another reason as usually an `@spackbot run pipeline` should be sufficient.

PR #33839, opened 2022-11-11
[xz] icc does not support attribute __symver__

stephenmsachs:
xz have added attribute `__symver__` for compilers that identify as GCC>=10.0 via `__GNUC__`. Intels `icc` sets `__GNUC__` but currently does not support this attribute:
https://community.intel.com/t5/Intel-C-Compiler/symver-not-supported/m-p/1429028/emcs_t/S2h8ZW1haWx8dG9waWNfc3Vic2NyaXB0aW9ufExBQVRWMjIyUFFZTlZTfDE0MjkwMjh8U1VCU0NSSVBUSU9OU3xoSw#M40459


haampie:
Sigh, you're right :(. Thanks for the excellent report upstream. Is there an issue for `xz` too? I noticed that Linux solves this issue like this:

```c
/* Compiler specific macros. */
#ifdef __clang__
#include <linux/compiler-clang.h>
#elif defined(__INTEL_COMPILER)
#include <linux/compiler-intel.h>
#elif defined(__GNUC__)
/* The above compilers also define __GNUC__, so order is important here. */
#include <linux/compiler-gcc.h>
#else
#error "Unknown compiler"
#endif
```

But I think what really needs to happen is a configure check whether the attribute is supported so it doesn't require tons of ifdefs.

haampie (dismissed):

var/spack/repos/builtin/packages/xz/package.py
```diff
@@ -39,6 +39,8 @@ class Xz(AutotoolsPackage, SourceforgePackage):
         description="Build shared libs, static libs or both",
     )
 
+    conflicts("%intel", when="@5.2.7", msg="icc does not support attribute __symver__")
```
haampie, line 44: Maybe `5.2.7:`?

stephenmsachs:
> Is there an issue for `xz` too?

I couldn't find any public `xz` repo so I emailed the main developer and proposed checking for `__INTEL_COMPILER`.

var/spack/repos/builtin/packages/xz/package.py
stephenmsachs: Probably safer for now. I'll add a comment too so people can check this in the future.

haampie (dismissed):

haampie (approved):
Thanks for checking with xz devs :upside_down_face:

stephenmsachs:
@spackbot run pipeline

PR #33840, opened 2022-11-11
Turn traverse tests into actual unit tests

haampie:
They don't need concrete specs

9 passed in 2.06s -> 9 passed in 0.02s, so a 100x faster.

This is just a drop in the ocean of slow Spack tests, but since typically new tests are added by just copying and adapting old tests, let's set a good example.


haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

alalazo (approved):

PR #33844, opened 2022-11-11
ompss-2: new package

dave96:
Include the ompss-2 package.

OmpSs-2 is a data-flow parallel programming model developed by the Barcelona Supercomputing Center. More information is available on the following web page: https://pm.bsc.es/ompss-2

This package file is a bit peculiar since OmpSs-2 has three separate components with different build systems (two compilers and a runtime), and a dependence on Jemalloc built with a specific prefix. Maybe there is a better approach to write this package than the one I've taken.

### DIFF
--- a/var/spack/repos/builtin/packages/ompss-2/package.py
+++ b/var/spack/repos/builtin/packages/ompss-2/package.py
@@ -1,0 +1,124 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import glob
+import os
+
+from spack.package import *
+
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/ompss-2/package.py
```diff
@@ -0,0 +1,124 @@
...
+class Ompss2(Package):
+    """OmpSs-2 is a data-flow shared-memory programming model based on source annotations.
+    It is developed by the Barcelona Supercomputing Center as a successor to the StarSs and
+    OmpSs programming models."""
+
+    homepage = "https://pm.bsc.es/ompss-2"
+
+    maintainers = ["dave96"]
+
+    version("2021.11.1", sha256="9e0ee0c9f75cd558882465efc3d521c2fe93f1a6b50d4d9c8e614ab4eb3a9e6c")
```
tldahlgren, line 24: Took awhile but confirmed the sha256.

tldahlgren (approved):

PR #33845, opened 2022-11-11
libxcrypt: 4.4.30

haampie:
This is a (missing) dependency of Python and Perl.

It used to be in `glibc` until it wasn't. `libxcrypt` provides a drop-in replacement `libcrypt.so.1`, but also an ABI-incompatible version `libcrypt.so.2` which is the default everywhere except Linux. I think it's alright to default to this anywhere, since we don't really have issues with ABI-incompatibility by design :upside_down_face: except maybe when mixing externals, but then people can work around it by toggling the variant value.


haampie:
@spackbot rerun pipeline

haampie:
@spackbot rerun pipeline

haampie:
@adamjstewart can you review this? Then we can test it in the Python PR soon :tm:

adamjstewart (approved):

PR #28109, opened 2021-12-21
ELF: verify needed libraries are resolvable

haampie:
This commit adds a post-install hook that walks over the prefix and
verifies that each ELF executable/library that takes part in shared
linking can in fact resolve its direct depenendencies.

This is slightly simpler than running `ldd` in the sense that it is not
recursive (we only care about the package) and its verification logic is
very minimal (we require ET_DYN type elf files for libs, and they should
match the class / data / isa of the parent).

Also it is platform independent, so we don't have to worry about calling
the correct `ldd`.

This check does not *fail* the install, it just prints it as an error
message.

### `llvm` 

```
In [1]: import spack.hooks.check_dynamic_linking_elf

In [2]: spec = spack.store.db.query("llvm")[0]

In [3]: spack.hooks.check_dynamic_linking_elf.post_install(spec)

==> Error: Not all executables/libraries can resolve their dependencies:
bin/llvm-omp-device-info
        libomp.so => /home/harmen/spack/opt/spack/linux-ubuntu22.04-zen2/gcc-11.2.0/llvm-15.0.0-neuno6erto7nzlbttmzuepmgujk3c76t/bin/../lib/libomp.so
        libomptarget.so.15 => /home/harmen/spack/opt/spack/linux-ubuntu22.04-zen2/gcc-11.2.0/llvm-15.0.0-neuno6erto7nzlbttmzuepmgujk3c76t/bin/../lib/libomptarget.so.15
        libhwloc.so.15 => not found

lib/libomp.so
        libhwloc.so.15 => not found

lib/libompd.so
        libomp.so => not found
        libhwloc.so.15 => not found

lib/libomptarget.rtl.amdgpu.so.15
        libelf.so.1 => not found
        libz.so.1 => not found
        libtinfo.so.6 => not found

lib/libomptarget.rtl.cuda.so.15
        libelf.so.1 => not found
        libz.so.1 => not found
        libtinfo.so.6 => not found

lib/libomptarget.rtl.x86_64.so.15
        libffi.so.8 => not found
        libelf.so.1 => not found
        libz.so.1 => not found
        libtinfo.so.6 => not found

lib/libomptarget.so.15
        libz.so.1 => not found
        libtinfo.so.6 => not found

libexec/llvm/llvm-omp-device-info
        libomp.so => not found
        libomptarget.so.15 => not found
        libhwloc.so.15 => not found
```

### `python`

```
lib/python3.9/lib-dynload/_crypt.cpython-39-x86_64-linux-gnu.so
[...]

PR #33847, opened 2022-11-11
python: missing libxcrypt dep

haampie:
Detected using https://github.com/spack/spack/pull/28109

May need https://github.com/spack/spack/pull/28109

@adamjstewart see also https://peps.python.org/pep-0594/#operating-system-interface, should we just try to disable all of crypt/nis/spwd by default (maybe even without an option to enable it?)

In any case, it would be model Python's dependencies correctly. I'm hoping https://github.com/spack/spack/pull/28109 can be a strict requirement in Gitlab CI to avoid accidental system deps.


adamjstewart:
I think we should add a `crypt` variant, like we have for all other optional Python modules with external dependencies. Also, according to https://github.com/python/cpython/issues/92238, this dependency was only added in Python 3.9.

haampie:
This whole libcrypt.so.1 thing is the least to say painful to watch. See for example

https://github.com/pypa/manylinux/issues/305

Some distributions no longer have libcrypt.so.1 as it was dropped from glibc, so I think it's good to use our own.

Then `libcrypt.so` has incompatible APIs (and hence a bumped soname), with the flag `--disable-obsolete-api` you get `libcrypt.so.2` iiuc. Maybe the libxcrypt package should be updated too.

Updated libxcrypt. I think we can safely go for libcrypt.so.2; it works fine for Python, and we don't model glibc anyways.

Note that libxcrypt takes 2 seconds to build, so I think adding a variant is not really worth it.

adamjstewart:
I think I would still like a variant, just for consistency with other optional dependencies. We've gotten too many complaints in the past about Spack building a million dependencies. The default will be true, but people can always disable the variant if they really want to speed up build times. Note that you'll also need to add an if-statement to the post-install import tests. I can push a commit for this if you want.

haampie:
Feel free to push to this branch if it's easy 🙃

adamjstewart:
Tested this a bit. Even on Python 3.8, it's still linking to crypt. I think the issue I linked to was mistaken and Python 3.9 just improved pkgconfig detection. Left the dep in for all versions < 3.13 just to be safe.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -370,6 +370,9 @@ class Python(Package):
     variant("tkinter", default=False, description="Build tkinter module")
     variant("uuid", default=True, description="Build uuid module")
     variant("tix", default=False, description="Build Tix module")
+    variant("crypt", default=True, description="Build crypt module", when="@:3.12 platform=linux")
+    variant("crypt", default=True, description="Build crypt module", when="@:3.12 platform=darwin")
+    variant("crypt", default=True, description="Build crypt module", when="@:3.12 platform=cray")
```
adamjstewart, line 375: @becker33 excited to see cray being removed as a platform, but would love a `platform=unix:` option or a "not" operator in our spec syntax.

adamjstewart (approved):

haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -560,6 +564,14 @@ def determine_variants(cls, exes, version_str):
             except ProcessError:
                 variants += "~tix"
 
+        # Some modules are platform-dependent
+        if not self.spec.satisfies("platform=windows"):
```
alalazo, line 568: This breaks detection, since it's inside a class method (`self` doesn't exist here).

adamjstewart: Oops, will fix using `is_windows`

PR #33849, opened 2022-11-11
depfile: use --output-sync=recurse by default

haampie:
Support is tested for using `.FEATURES`.

I haven't checked yet how to improve this s.t. we don't override
`--output-sync` on the command line; I'd like to just set the default.

### DIFF
--- a/share/spack/templates/depfile/Makefile
+++ b/share/spack/templates/depfile/Makefile
@@ -1,5 +1,10 @@
 SPACK ?= spack
 
+# Enable --output-sync=recurse by default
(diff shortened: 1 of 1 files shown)


alalazo:
Why `--output-sync=recurse` and not `--output-sync=target`?

haampie:
The call chain is something like make calls spack calls make. With recursive, all output for a single install is grouped together. With target, the child make will print to screen, only the top-level target make output is grouped together

haampie:
I think I'll drop this for now and let the user decide.

PR #33851, opened 2022-11-11
Package/py pysam macos

luke-dt:
Fixed a bug that was preventing `py-pysam` from building on MacOS (`clang: error: invalid argument '-bundle' not allowed with '-dynamiclib'`), and cleaned up the package in general. Previously, the `package.py` included `bcftools`/`htslib`/`samtools` deps, but `pysam` actually bundles all those necessary libs for building in the tarball, and they are used by default. I removed those dependencies entirely since the package wasn't using them anyway. This [mirrors the changes in the bioconda recipe](https://github.com/bioconda/bioconda-recipes/pull/13093).

I also tested with external (spack-installed) `htslib` - which built fine - but thought that it would be easier to just use the included libs instead of making sure the spack deps match the bundled libs for every release. See https://pysam.readthedocs.io/en/latest/installation.html#external for more detail

### DIFF
--- a/var/spack/repos/builtin/packages/htslib/package.py
+++ b/var/spack/repos/builtin/packages/htslib/package.py
@@ -52,6 +52,10 @@ class Htslib(AutotoolsPackage):
     depends_on("automake", when="@1.2")
     depends_on("libtool", when="@1.2")
 
+    @property
+    def libs(self):
+        return find_libraries("*", root=self.prefix, recursive=True)
+
     # v1.2 uses the automagically assembled tarball from .../archive/...
     # everything else uses the tarballs uploaded to the release
     def url_for_version(self, version):
--- a/var/spack/repos/builtin/packages/htslib/package.py
+++ b/var/spack/repos/builtin/packages/htslib/package.py
@@ -54,7 +54,7 @@ class Htslib(AutotoolsPackage):
 
     @property
     def libs(self):
-        return find_libraries("*", root=self.prefix, recursive=True)
+        return find_libraries("libhts", root=self.prefix, recursive=True)
 
     # v1.2 uses the automagically assembled tarball from .../archive/...
     # everything else uses the tarballs uploaded to the release
--- a/var/spack/repos/builtin/packages/py-pysam/package.py
+++ b/var/spack/repos/builtin/packages/py-pysam/package.py
@@ -22,15 +22,20 @@ class PyPysam(PythonPackage):
     version("0.7.7", sha256="c9f3018482eec99ee199dda3fdef2aa7424dde6574672a4c0d209a10985755cc")
 
     depends_on("py-setuptools", type="build")
-    depends_on("py-cython@0.29.12:", when="@0.18:", type="build")
-    depends_on("py-cython@0.21:", when="@0.14:", type="build")
+    depends_on("py-cython@0.29.12:", type="build", when="@0.18:")
+    depends_on("py-cython@0.21:", type="build", when="@0.14:")
     depends_on("py-cython@0.17:", type="build")
     depends_on("curl")
     depends_on("bcftools")
+    depends_on("htslib@:1.6", when="@:0.13")
     depends_on("htslib")
     depends_on("samtools")
 
-    depends_on("htslib@:1.6", when="@:0.13")
-
     def setup_build_environment(self, env):
         env.set("LDFLAGS", self.spec["curl"].libs.search_flags)
+        # linking htslib, see:
+        # http://pysam.readthedocs.org/en/latest/installation.html#external
+        # https://github.com/pysam-developers/pysam/blob/v0.9.0/setup.py#L79
+        env.set("HTSLIB_MODE", "external")
+        env.set("HTSLIB_LIBRARY_DIR", self.spec["htslib"].prefix.lib)
+        env.set("HTSLIB_INCLUDE_DIR", self.spec["htslib"].prefix.include)
--- a/var/spack/repos/builtin/packages/py-pysam/package.py
+++ b/var/spack/repos/builtin/packages/py-pysam/package.py
@@ -39,5 +39,5 @@ class PyPysam(PythonPackage):
         # http://pysam.readthedocs.org/en/latest/installation.html#external
         # https://github.com/pysam-developers/pysam/blob/v0.9.0/setup.py#L79
         env.set("HTSLIB_MODE", "external")
-        env.set("HTSLIB_LIBRARY_DIR", self.spec["htslib"].prefix.lib)
-        env.set("HTSLIB_INCLUDE_DIR", self.spec["htslib"].prefix.include)
+        env.set("HTSLIB_LIBRARY_DIR", self.spec["htslib"].libs.directories[0])
+        env.set("HTSLIB_INCLUDE_DIR", self.spec["htslib"].headers.directories[0])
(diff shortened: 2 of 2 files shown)


adamjstewart:
I think we should prefer using Spack-installed dependencies instead of bundled dependencies. This allows us to inject compiler flags, add patches for specific OSes/compilers, and patch security vulnerabilities without having to worry about waiting for pysam to fix these things themselves. For example, our bcftools package contains a number of patches that likely aren't in pysam. This also improves reuse and prevents multiple versions of these packages from ending up in the DAG at the same time.

var/spack/repos/builtin/packages/py-pysam/package.py
```diff
@@ -26,11 +26,9 @@ class PyPysam(PythonPackage):
...
-    depends_on("htslib")
-    depends_on("samtools")
-
-    depends_on("htslib@:1.6", when="@:0.13")
+    depends_on("xz")
 
     def setup_build_environment(self, env):
         env.set("LDFLAGS", self.spec["curl"].libs.search_flags)
+        if self.spec.platform == "darwin":
+            env.remove_flags("LDSHARED", "-bundle")
```
adamjstewart, line 37: Where is this env var being set? Should we just unset it entirely if it's coming from outside of Spack?

luke-dt: I have no idea lol. The issue I was running into was that when building `pysam` on macOS, I would get the error `clang: error: invalid argument '-bundle' not allowed with '-dynamiclib'`. I dug into the source code, and it appears that [this modification wasn't happening in spack for some reason](https://github.com/pysam-developers/pysam/blob/v0.19.1/linker_tests/link_pre_489/cy_build.py#L15). It built fine for me outside of spack. This fix is just me manually adding that same modification.

Attached is the build log and env if you're curious.

[spack-build-env.txt](https://github.com/spack/spack/files/9993425/spack-build-env.txt)
[spack-build-out.txt](https://github.com/spack/spack/files/9993426/spack-build-out.txt)

luke-dt:
> 

so what should I do about `samtools` and `bcftools`? I can link external `htslib` so that `pysam` doesn't rebuild it, but as far as I could tell it was never using external `samtools` and `bcftools`, even when listed as a dep

adamjstewart:
Hmm, maybe open an issue and ask the devs if it's possible to use externals.

var/spack/repos/builtin/packages/py-pysam/package.py
adamjstewart: Prob worth reporting this as a bug to the devs and getting to the bottom of this. I try to avoid including hacks and patches unless the devs are aware of the situation and working on a fix.

luke-dt:
So I haven't got a response from the devs about using external libraries. I would guess that it's not a priority for them

https://github.com/pysam-developers/pysam/issues/1150

As it stands now, the `py-pysam` package ignores linked linked `bcftools` `htslib`, and `samtools`. I'm not sure it if makes sense to include them in the package if they aren't being used.

adamjstewart:
If it's currently not possible to use externally installed versions, then yes, we should remove the dependencies.

luke-dt:
I think htslib is the only one that is possible to use an external version. I got it to work one time, but I didn't know what you all would prefer: use all bundled libs, or bundled `bcftools` + `samtools` and external `htslib`

https://github.com/spack/spack/pull/33851/commits/3a1c19486daaeddf118ed0740072c748f12ebf6b

adamjstewart:
I would prefer the latter as it's a step in the right direction.

luke-dt:
OK, I'll make another commit to revert that change

var/spack/repos/builtin/packages/py-pysam/package.py
```diff
@@ -27,8 +27,17 @@ class PyPysam(PythonPackage):
...
         env.set("LDFLAGS", self.spec["curl"].libs.search_flags)
+        # this flag is supposed to be removed by cy_build.py, but for some reason isn't
         if self.spec.platform == "darwin":
             env.remove_flags("LDSHARED", "-bundle")
+        # linking htslib, see:
+        # http://pysam.readthedocs.org/en/latest/installation.html#external
+        # https://github.com/pysam-developers/pysam/blob/v0.9.0/setup.py#L79
+        env.set("HTSLIB_MODE", "external")
+        env.set("HTSLIB_LIBRARY_DIR", self.spec["htslib"].prefix.lib)
+        env.set("HTSLIB_INCLUDE_DIR", self.spec["htslib"].prefix.include)
```
adamjstewart:
```suggestion
        env.set("HTSLIB_LIBRARY_DIR", self.spec["htslib"].libs.directories[0])
        env.set("HTSLIB_INCLUDE_DIR", self.spec["htslib"].headers.directories[0])
```
On some systems lib64 is used instead of lib, this should support both.

luke-dt: I tried this and got a build error:

```
==> Installing py-pysam-0.19.1-g37uykesacdxvmjwik25ptdqsfd35v2o
==> No binary for py-pysam-0.19.1-g37uykesacdxvmjwik25ptdqsfd35v2o found: installing from source
==> Error: NoLibrariesError: Unable to recursively locate htslib libraries in /Users/luke/Documents/GitHub/dantaslab/spack/opt/spack/darwin-monterey-m1/apple-clang-14.0.0/htslib-1.16-su7mfvegt5xjb7o7lzulziwcigvpna64

/Users/luke/Documents/GitHub/dantaslab/spack/var/spack/repos/builtin/packages/py-pysam/package.py:42, in setup_build_environment:
         39        # http://pysam.readthedocs.org/en/latest/installation.html#external
         40        # https://github.com/pysam-developers/pysam/blob/v0.9.0/setup.py#L79
         41        env.set("HTSLIB_MODE", "external")
  >>     42        env.set("HTSLIB_LIBRARY_DIR", self.spec["htslib"].libs.directories[0])
         43        env.set("HTSLIB_INCLUDE_DIR", self.spec["htslib"].headers.directories[0])
```

adamjstewart: You'll need to override the `libs` property in the htslib package. What is the library called? See zlib for an example where this is done.

luke-dt: Do you mean add something like:

````
@property
def libs(self):
    return foo
````

to the `py-pysam` package or the `htslib` package?

Comment (adamjstewart) [reply var/spack/repos/builtin/packages/py-pysam/package.py]: To the htslib package

File: var/spack/repos/builtin/packages/htslib/package.py [hunk 3]
```diff
@@ -52,6 +52,10 @@ class Htslib(AutotoolsPackage):
     depends_on("automake", when="@1.2")
     depends_on("libtool", when="@1.2")
 
+    @property
+    def libs(self):
+        return find_libraries("*", root=self.prefix, recursive=True)
```
adamjstewart: You should include the exact name of the library. Otherwise, if it's installed in `/usr/lib`, you'll pick up thousands of libraries.

luke-dt: I think it's `lilbhts`? This is the output of `find_libraries("*", root=self.prefix, recursive=True)` :

```
SPACK_DIR/htslib-1.16-su7mfvegt5xjb7o7lzulziwcigvpna64/lib/libhts.1.16.dylib
SPACK_DIR/htslib-1.16-su7mfvegt5xjb7o7lzulziwcigvpna64/lib/libhts.dylib
SPACK_DIR/htslib-1.16-su7mfvegt5xjb7o7lzulziwcigvpna64/lib/libhts.3.dylib
SPACK_DIR/htslib-1.16-su7mfvegt5xjb7o7lzulziwcigvpna64/lib/libhts.1.16.dylib
SPACK_DIR/htslib-1.16-su7mfvegt5xjb7o7lzulziwcigvpna64/lib/libhts.dylib
SPACK_DIR/htslib-1.16-su7mfvegt5xjb7o7lzulziwcigvpna64/lib/libhts.3.dylib
```

adamjstewart (approved):

PR #33855, opened 2022-11-11
e4s ci: use runner image with awscli tool

eugeneswalker:
@alalazo @scottwittenburg

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -275,11 +275,11 @@ e4s-mac-protected-build:
 
 e4s-pr-generate:
   extends: [ ".e4s", ".pr-generate"]
-  image: ecpe4s/ubuntu20.04-runner-x86_64:2022-10-01
+  image: ecpe4s/ubuntu20.04-runner-x86_64:2022-11-11
 
 e4s-protected-generate:
   extends: [ ".e4s", ".protected-generate"]
-  image: ecpe4s/ubuntu20.04-runner-x86_64:2022-10-01
+  image: ecpe4s/ubuntu20.04-runner-x86_64:2022-11-11
 
 e4s-pr-build:
   extends: [ ".e4s", ".pr-build" ]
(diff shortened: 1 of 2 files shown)


scottwittenburg:
Several jobs in the e4s stack (I think 5 of them) failed the same way, any idea if these locking errors are new in spack recently, introduced by the change in image, or what?  It's something I haven't seen before:

```
==> Error: Failed to install bzip2 due to LockTimeoutError: Timed out waiting for a write lock after 120.146s.
    Made 281 attempts on file: /root/.spack/cache/indices/.74cb74a434_f65176471f.json.lock
```

Seems they all happened in installation of dependencies, and that lock file is one used by the `binary_distribution` module when it caches locally the indices it fetches from remote mirrors.  If it's something new since the installation of dependencies got parallelized a few weeks ago, I would think we'd have run into this before now, so maybe there's some other cause.

eugeneswalker:
I have seen these before, yes, but only in manual installations where I tried to say, install the entire E4S environment with 100+ root specs translating into >500 individual packages, in parallel. When I have tried parallel installs that way, there is always some kind of synchronization issue and the parallel install fails. I don't know why we are seeing it here.

Here are the changes I made to the standard runner image recipe:
* https://github.com/UO-OACISS/e4s/commit/249bf6a650dbd551601e9786fcf25f6c293c8dd0

I don't know enough about the how `awscli` is used to know if the code path that uses it could lead to these errors where the code path that does not use it, does not cause these errors...

scottwittenburg:
Since #33248, dependencies are now installed in parallel in gitlab ci, but if that were the cause, I would have thought we'd see this before now.  Maybe it's related to actually running more jobs in parallel, with #33688?  Needs some investigation I think.

The `aws` command is just used within the `spack buildcache update-index ...` command, so it seems unlikely it could be causing this locking issue when installing dependencies.  I've been wrong plenty of times though.

eugeneswalker:
Perhaps we're seeing it here where we didn't see it elsewhere because I'm modifying the `.gitlab-ci.yml` and that's triggering more jobs than the normal pipelines + everything you said about those recent PRs? The combination of it all. I'll dive in on the errors here a bit later in the week.

eugeneswalker:
Closing in favor of https://github.com/spack/spack/pull/34212

PR #33856, opened 2022-11-11
GDAL: add v3.6.0

adamjstewart:
Successfully builds on macOS 12.6.1 (arm64) with Apple Clang 14.0.0.

https://github.com/OSGeo/gdal/releases/tag/v3.6.0

@rouault I noticed that the recommended dependencies listed [here](https://gdal.org/build_hints.html#build-requirements) don't match the dependencies listed as recommended by CMake. For example, CMake lists brand new dependencies like Brunsli and QB3 as recommended even though they are only needed for a single driver, while things like cURL are listed as optional instead of recommended. Which should I trust when choosing defaults here?

### DIFF
--- a/var/spack/repos/builtin/packages/gdal/package.py
+++ b/var/spack/repos/builtin/packages/gdal/package.py
@@ -94,11 +94,11 @@ class Gdal(CMakePackage, AutotoolsPackage, PythonExtension):
         "basisu", default=False, when="@3.6:", description="Required for BASISU and KTX2 drivers"
     )
     variant("blosc", default=False, when="@3.4:", description="Required for Zarr driver")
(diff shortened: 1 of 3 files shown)


rouault:
> CMake lists brand new dependencies like Brunsli and QB3 as recommended

the commiter of those features was probably loving a bit too much his own niche ;-)

> Which should I trust when choosing defaults here?

I don't have an answer to this. Everybody has his own opinion of what GDAL should be. But if you want mine, curl is more recommended than Brunsli or QB3. I've pushed a PR adjusting this in https://github.com/OSGeo/gdal/pull/6672

tldahlgren (approved):
Confirmed new sha256 and branches for the new packages.

PR #33859, opened 2022-11-12
bump release version for tutorial command

becker33:
Use v0.19

We should probably add updating this to the release instructions, since we generally won't run a tutorial on an old version

### DIFF
--- a/lib/spack/spack/cmd/tutorial.py
+++ b/lib/spack/spack/cmd/tutorial.py
@@ -11,6 +11,7 @@ import shutil
 import llnl.util.tty as tty
 from llnl.util.filesystem import working_dir
 
+import spack
 import spack.cmd.common.arguments as arguments
 import spack.config
 import spack.paths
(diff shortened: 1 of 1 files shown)


lib/spack/spack/cmd/tutorial.py
```diff
@@ -24,7 +24,7 @@
 
 
 # tutorial configuration parameters
-tutorial_branch = "releases/v0.18"
+tutorial_branch = "releases/v0.19"
```
tgamblin: Should we just update this to:
```suggestion
tutorial_branch = "releases/v%s" % ".".join(spack.spack_version_info[:2])
```

tgamblin (changes requested):

lib/spack/spack/cmd/tutorial.py
becker33: That is a great idea

PR #33862, opened 2022-11-12
Add new custom markers to unit tests

alalazo:
This PR consolidate in one place common logic to skip tests. It also changes all the tests employing that logic accordingly. By using markers we can also list or run tests that e.g. don't run yet on Windows, or need clingo to pass:
```console
$ spack unit-test -m only_clingo 
```

The main modification is here: f0aafd7cf5f81a33fea9197530928055fc63d6a6 All the other changes are just substituting duplicated logic with the markers.

Modifications:
- [x] Add a `only_clingo` marker, to mark tests that run only with clingo
- [x] Add a `only_original` marker, to mark tests that are specific to the original concretizer
- [x] Add a `not_on_windows` marker, to mark tests that are not supported on windows
- [x] Modified test files to use the new markers

### DIFF
--- a/lib/spack/spack/test/architecture.py
+++ b/lib/spack/spack/test/architecture.py
@@ -199,14 +199,10 @@ def test_satisfy_strict_constraint_when_not_concrete(architecture_tuple, constra
     ],
 )
 @pytest.mark.usefixtures("mock_packages", "config")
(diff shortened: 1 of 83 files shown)


alalazo:
@spackbot fix style

haampie (approved):

PR #33864, opened 2022-11-12
add sz3 and mdz smoke test for inclusion in e4s

robertu94:
@ayzk  could you please update the release version for on master SZ3 to 3.1.7. After you do this, I'll add a version to the spack file.


robertu94:
Thanks @ayzk!  This is ready for review from @spack/e4s

var/spack/repos/builtin/packages/sz3/package.py
```diff
@@ -13,14 +13,17 @@ class Sz3(CMakePackage):
     git = "https://github.com/szcompressor/SZ3"
 
     maintainers = ["disheng222"]
+    tags = ["e4s"]
 
     version("master")
+    version("3.1.7", commit="c49fd17f2d908835c41000c1286c510046c0480e")
```
tldahlgren, line 19: Confirmed the commit.

var/spack/repos/builtin/packages/sz3/package.py
```diff
@@ -35,5 +38,15 @@ def cmake_args(self):
         return [
             "-DSZ3_USE_BUNDLED_ZSTD=OFF",
             "-DSZ3_DEBUG_TIMINGS=OFF",
+            self.define_from_variant("BUILD_MDZ", "mdz"),
             self.define_from_variant("BUILD_H5Z_FILTER", "hdf5"),
         ]
+
+    def test(self):
+        if self.version < Version("3.1.7"):
```
tldahlgren: Please output a message here about skipping the test and or only have tests for version 3.1.7:.

tldahlgren: Perhaps use `self.spec.satisfies("3.1.7:")` here?

tldahlgren (commented):
Confirmed the tests pass after installing the software.  Just have a couple of suggestions.

robertu94 (commented):
Fixes reqested by @tldahgren

var/spack/repos/builtin/packages/sz3/package.py
```diff
@@ -35,5 +38,16 @@ def cmake_args(self):
         return [
             "-DSZ3_USE_BUNDLED_ZSTD=OFF",
             "-DSZ3_DEBUG_TIMINGS=OFF",
+            self.define_from_variant("BUILD_MDZ", "mdz"),
             self.define_from_variant("BUILD_H5Z_FILTER", "hdf5"),
         ]
+
+    def test(self):
+        if self.spec.satisifes("@3.1.7:"):
```
tldahlgren: There is a typo here:  `satisifes` -> `satisfies`.

tldahlgren: Also, since the tests are supposed to be skipped for versions < `3.1.7` the call should be more like `self.spec.satisfies("@:3.1.6")`.

tldahlgren (changes requested):
There's a typo and the version range isn't quite right.

var/spack/repos/builtin/packages/sz3/package.py
tldahlgren: FWIW. I  tested three different installations of the package with the above fixes and the tests ran as expected.

robertu94: fixed

tldahlgren (approved):
LGTM

PR #33876, opened 2022-11-13
py-alphafold: update to 2.2.4, update dependencies

aweits:


adamjstewart:
Can you try enabling jax in CI? Just uncomment in `share/spack/gitlab/cloud_pipelines/stacks/ml-*/spack.yaml`

share/spack/gitlab/cloud_pipelines/stacks/ml-cpu/spack.yaml
```diff
@@ -29,8 +29,8 @@ spack:
 
     # JAX
     # https://github.com/google/jax/issues/12614
```
adamjstewart: If this works, you can remove this issue link, and we can close that issue.

aweits: removed, also dealt with the issue in tensorflow

var/spack/repos/builtin/packages/openmm/package.py
```diff
@@ -28,7 +29,9 @@ class Openmm(CMakePackage, CudaPackage):
 
     depends_on("python@2.7:", type=("build", "run"))
     depends_on("cmake@3.1:", type="build")
-    depends_on("doxygen", type="build")
+    # https://github.com/openmm/openmm/issues/3317
+    depends_on("doxygen@1.9.1", type="build", when="@:7.6.0")
```
adamjstewart:
```suggestion
    depends_on("doxygen@:1.9.1", type="build", when="@:7.6.0")
```

var/spack/repos/builtin/packages/py-alphafold/package.py
```diff
@@ -17,31 +17,38 @@ class PyAlphafold(PythonPackage, CudaPackage):
     url = "https://github.com/deepmind/alphafold/archive/refs/tags/v2.1.1.tar.gz"
     maintainers = ["aweits"]
 
+    version("2.2.4", sha256="8d756e16f6dc7897331d834aade8493820d0ff6a03bf60ce511bac4756c1b1e8")
     version("2.1.1", sha256="1adb6e213ba9ac321fc1acb1c563ba9b4fc054c1cebe1191bc0e2aaa671dadf7")
 
     conflicts("platform=darwin", msg="alphafold is only supported on Linux")
 
     depends_on("python@3.7:", type=("build", "run"))
```
adamjstewart, line 27: I'm sure I asked this question last time too, but why 3.7+? The `setup.py` lists 3.6 as supported.

var/spack/repos/builtin/packages/py-alphafold/package.py
```diff
@@ -17,31 +17,38 @@ class PyAlphafold(PythonPackage, CudaPackage):
...
     depends_on("py-biopython@1.79:", type=("build", "run"))
     depends_on("py-chex@0.0.7:", type=("build", "run"))
-    depends_on("py-dm-haiku@0.0.4:", type=("build", "run"))
+    depends_on("py-dm-haiku@0.0.4:", type=("build", "run"), when="@2.1.1")
+    depends_on("py-dm-haiku@0.0.7:", type=("build", "run"), when="@2.2.4")
     depends_on("py-dm-tree@0.1.6:", type=("build", "run"))
     depends_on("py-immutabledict@2.0.0:", type=("build", "run"))
-    depends_on("py-jax@0.2.14:", type=("build", "run"))
+    depends_on("py-jax@0.2.14:", type=("build", "run"), when="@2.1.1")
+    depends_on("py-jax@0.3.17:", type=("build", "run"), when="@2.2.4")
```
adamjstewart, line 39: Where do all of these versions come from? I don't see anything in `setup.py`. If there's some other source of truth out there, can you add a comment so I don't ask you this same question every time I review a PR for this package?

var/spack/repos/builtin/packages/py-alphafold/package.py
```diff
@@ -17,31 +17,38 @@ class PyAlphafold(PythonPackage, CudaPackage):
...
     depends_on("py-setuptools", type="build")
-    depends_on("py-absl-py@0.13.0:", type=("build", "run"))
+    depends_on("py-absl-py@0.13.0:", type=("build", "run"), when="@2.1.1")
+    depends_on("py-absl-py@1.0.0:", type=("build", "run"), when="@2.2.4")
     depends_on("py-biopython@1.79:", type=("build", "run"))
     depends_on("py-chex@0.0.7:", type=("build", "run"))
-    depends_on("py-dm-haiku@0.0.4:", type=("build", "run"))
+    depends_on("py-dm-haiku@0.0.4:", type=("build", "run"), when="@2.1.1")
+    depends_on("py-dm-haiku@0.0.7:", type=("build", "run"), when="@2.2.4")
     depends_on("py-dm-tree@0.1.6:", type=("build", "run"))
```
adamjstewart, line 35: Missing dep on `py-docker`?

var/spack/repos/builtin/packages/py-alphafold/package.py
```diff
@@ -17,31 +17,38 @@ class PyAlphafold(PythonPackage, CudaPackage):
...
+            "py-jax+cuda cuda_arch={0}".format(arch),
             type=("build", "run"),
             when="cuda_arch={0}".format(arch),
         )
     depends_on("py-ml-collections@0.1.0:", type=("build", "run"))
-    depends_on("py-numpy@1.19.5:", type=("build", "run"))
+    depends_on("py-numpy@1.19.5:", type=("build", "run"), when="@2.1.1")
+    depends_on("py-numpy@1.21.6:", type=("build", "run"), when="@2.2.4")
     depends_on("py-pandas@1.3.4:", type=("build", "run"))
+    depends_on("py-protobuf@3.20.1:", type=("build", "run"), when="@2.2.4")
```
adamjstewart: I don't see this in `setup.py`

var/spack/repos/builtin/packages/py-alphafold/package.py
```diff
@@ -17,31 +17,38 @@ class PyAlphafold(PythonPackage, CudaPackage):
...
     depends_on("py-scipy@1.7.0:", type=("build", "run"))
     depends_on("py-pdbfixer@1.7", type=("build", "run"))
```
adamjstewart, line 52: Same here

var/spack/repos/builtin/packages/py-dm-haiku/package.py
```diff
@@ -13,6 +13,7 @@ class PyDmHaiku(PythonPackage):
     homepage = "https://github.com/deepmind/dm-haiku"
     pypi = "dm-haiku/dm-haiku-0.0.5.tar.gz"
 
+    version("0.0.7", sha256="86c34af6952a305a4bbfda6b9925998577acc4aa2ad9333da3d6047f4f8ed7c1")
     version("0.0.5", sha256="e986237e1f840aa3bd26217ecad84b611bf1456e2139f0f79ea71f9c6222d231")
     depends_on("python@3.7:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
```
adamjstewart, line 19: Where does the jax dep come from? I see an optional jax/jaxlib option but no required jax dep.

var/spack/repos/builtin/packages/py-jax/package.py
```diff
@@ -21,22 +21,29 @@ class PyJax(PythonPackage, CudaPackage):
...
     depends_on("py-numpy@1.18:", type=("build", "run"))
     depends_on("py-absl-py", type=("build", "run"))
     depends_on("py-opt-einsum", type=("build", "run"))
-    depends_on("py-scipy@1.2.1:", type=("build", "run"))
+    depends_on("py-scipy@1.2.1:", type=("build", "run"), when="@0.2.25")
+    depends_on("py-scipy@1.5:", type=("build", "run"), when="@0.3.23")
     depends_on("py-typing-extensions", type=("build", "run"))
-    depends_on("py-jaxlib@0.1.69:", type=("build", "run"), when="~cuda")
-    depends_on("py-jaxlib@0.1.69:+cuda", type=("build", "run"), when="+cuda")
+    depends_on("py-etils+epath", type=("build", "run"))
```
adamjstewart: Is this needed for all versions or just the latest version?

var/spack/repos/builtin/packages/py-jax/package.py
```diff
@@ -21,22 +21,29 @@ class PyJax(PythonPackage, CudaPackage):
...
+    depends_on("py-jaxlib@0.3.15:", type=("build", "run"), when="@0.3.23~cuda")
+    depends_on("py-jaxlib@0.3.15:+cuda", type=("build", "run"), when="@0.3.23+cuda")
+    depends_on("py-jaxlib@0.1.69:", type=("build", "run"), when="@0.2.25~cuda")
+    depends_on("py-jaxlib@0.1.69:+cuda", type=("build", "run"), when="@0.2.25+cuda")
     for arch in CudaPackage.cuda_arch_values:
```
adamjstewart, line 44: Idk if we should even bother forwarding things like `+cuda` or `cuda_arch`. Spack now supports variant forwarding (`spack install py-jax ++cuda cuda_arch==...`) and people should really set these things in their `packages.yaml` instead.

var/spack/repos/builtin/packages/py-jaxlib/package.py
```diff
@@ -17,21 +17,42 @@ class PyJaxlib(PythonPackage, CudaPackage):
     url = "https://github.com/google/jax/archive/refs/tags/jaxlib-v0.1.74.tar.gz"
 
     tmp_path = ""
+    buildtmp = ""
 
+    version("0.3.22", sha256="680a6f5265ba26d5515617a95ae47244005366f879a5c321782fde60f34e6d0d")
     version("0.1.74", sha256="bbc78c7a4927012dcb1b7cd135c7521f782d7dad516a2401b56d3190f81afe35")
```
adamjstewart, line 22: Can you add a comment that the deps are listed in `jaxlib/setup.py`, not in `setup.py`? I look in the wrong place every time I review this package...

var/spack/repos/builtin/packages/py-tensorboard/package.py
```diff
@@ -84,7 +84,7 @@ class PyTensorboard(PythonPackage):
     depends_on("py-markdown@2.6.8:", type=("build", "run"))
     depends_on("py-numpy@1.12.0:", type=("build", "run"))
     depends_on("py-protobuf@3.6.0:3.19", type=("build", "run"), when="@:2.8")
-    depends_on("py-protobuf@3.9.2:3.19", type=("build", "run"), when="@2.9:")
+    depends_on("py-protobuf@3.9.2:3.20", type=("build", "run"), when="@2.9:")
```
adamjstewart: This is only true in TB 3.11, which hasn't been packaged yet. Before https://github.com/tensorflow/tensorboard/pull/5922 it was `<3.20`.

var/spack/repos/builtin/packages/py-tensorflow/package.py
```diff
@@ -299,8 +299,8 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage):
     depends_on("protobuf@:3.17")
     # https://github.com/protocolbuffers/protobuf/issues/10051
     # https://github.com/tensorflow/tensorflow/issues/56266
-    depends_on("py-protobuf@:3.19", type=("build", "run"))
-    depends_on("protobuf@:3.19", type=("build", "run"))
+    depends_on("py-protobuf@:3.20", type=("build", "run"))
+    depends_on("protobuf@:3.20", type=("build", "run"))
```
adamjstewart: According to https://github.com/tensorflow/tensorflow/issues/56266, 3.20 doesn't work?

adamjstewart (changes requested):
Protobuf changes look a little sus. Most other changes are prob right but I don't see where you're getting the dependency version ranges from so it's hard to review. Can you add comments to those packages whenever it isn't as simple as checking the root `setup.py` file for a list of deps?

var/spack/repos/builtin/packages/py-alphafold/package.py
aweits: I'd add it if we were actually building the docker image - I'm stopping at just spackify-ing the build

var/spack/repos/builtin/packages/py-dm-haiku/package.py
aweits: added comment

var/spack/repos/builtin/packages/py-jax/package.py
aweits: is there a better way to passthrough and make jax/jaxlib move together in this respect?

var/spack/repos/builtin/packages/py-jax/package.py
aweits: fixed

var/spack/repos/builtin/packages/openmm/package.py
```diff
@@ -28,7 +29,9 @@ class Openmm(CMakePackage, CudaPackage):
 
     depends_on("python@2.7:", type=("build", "run"))
     depends_on("cmake@3.1:", type="build")
```
adamjstewart, line 32: Latest version requires cmake 3.17+

var/spack/repos/builtin/packages/py-alphafold/package.py
```diff
@@ -17,31 +17,40 @@ class PyAlphafold(PythonPackage, CudaPackage):
     url = "https://github.com/deepmind/alphafold/archive/refs/tags/v2.1.1.tar.gz"
     maintainers = ["aweits"]
 
+    version("2.2.4", sha256="8d756e16f6dc7897331d834aade8493820d0ff6a03bf60ce511bac4756c1b1e8")
     version("2.1.1", sha256="1adb6e213ba9ac321fc1acb1c563ba9b4fc054c1cebe1191bc0e2aaa671dadf7")
 
     conflicts("platform=darwin", msg="alphafold is only supported on Linux")
 
+    # lots of hints on versions and patching taken from docker/Dockerfile
+    # and requirements.txt
```
adamjstewart, line 26: I'm not sure how much I trust these "hints". The docker file just so happens to use Python 3.7, but that doesn't suggest to me that that's the minimum version that works. Similarly with `requirements.txt`, they might just pin CI to those versions for stability reasons. I do the same thing in other projects, although it's the latest version, not the minimum version. If you want to use `requirements.txt` for minimums, I'm fine with it since you're maintaining this recipe, but it isn't what I would personally do.

var/spack/repos/builtin/packages/py-alphafold/package.py
adamjstewart: `setup.py` contains `docker`, so if I were to `pip install alphafold`, it would install docker. I don't think it's a good idea to diverge too much from the listed required deps. Things like `pip check` will fail if deps declared as required are missing.

var/spack/repos/builtin/packages/py-jaxlib/package.py
```diff
@@ -17,21 +17,43 @@ class PyJaxlib(PythonPackage, CudaPackage):
...
     depends_on("py-setuptools", type="build")
-    depends_on("py-numpy@1.18:", type=("build", "run"))
-    depends_on("py-scipy", type=("build", "run"))
+
+    depends_on("py-numpy@1.18:", type=("build", "run"), when="@0.1.74")
+    depends_on("py-numpy@1.20:", type=("build", "run"), when="@0.3.22")
+    depends_on("py-scipy@1.5:", type=("build", "run"))
     depends_on("py-absl-py", type=("build", "run"))
-    depends_on("py-flatbuffers@1.12:2", type=("build", "run"))
-    # Bazel 5 not yet supported: https://github.com/google/jax/issues/8440
```
adamjstewart, line 29: It'd be good to keep this comment since the upper bound isn't documented in older deps

var/spack/repos/builtin/packages/py-jaxlib/package.py
```diff
@@ -17,21 +17,43 @@ class PyJaxlib(PythonPackage, CudaPackage):
...
 
     def install(self, spec, prefix):
+        self.tmp_path = tempfile.mkdtemp(prefix="spack")
+        self.buildtmp = tempfile.mkdtemp(prefix="spack")
+        filter_file(
+            """f"--output_path={output_path}",""",
+            """f"--output_path={output_path}","""
+            """f"--sources_path=%s","""
+            """f"--nohome_rc'","""
+            """f"--nosystem_rc'",""" % self.tmp_path,
```
adamjstewart: No reason to use triple quotes here

var/spack/repos/builtin/packages/py-jaxlib/package.py
```diff
@@ -50,21 +72,3 @@ def install(self, spec, prefix):
             pip(*args)
         remove_linked_tree(self.tmp_path)
         remove_linked_tree(self.buildtmp)
-
-    def patch(self):
```
adamjstewart, line 54: Any reason to drop the patch method? I think it helps to logically separate patching from installation. Unless the multi-build system refactor broke this somehow.

adamjstewart (commented):
Mostly meta comments and non-required suggestions, I think this is basically good to go.

var/spack/repos/builtin/packages/py-jaxlib/package.py
aweits: the multi-build system has yielded some surprising behavior here - not sure I like this solution, but see all the references to "wrapped_package_object" perhaps @alalazo has some words of wisdom?

adamjstewart (dismissed):

aweits:
any objection to merging as-is?

adamjstewart:
We need to get CI passing first. Either rebase or keep pinging spackbot to rerun the pipeline

adamjstewart (approved):

PR #33884, opened 2022-11-14
Be strict on the markers used in unit tests

alalazo:
This needs a recent `pytest` version and doesn't work with our vendored dependency

Modifications:
- [x] Add `--strict-markers` to options used to run unit tests

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -144,6 +144,9 @@ jobs:
     - name: Run unit tests
       shell: runuser -u spack-test -- bash {0}
       run: |
+          /usr/libexec/platform-python -m venv platform-python
+          source platform-python/bin/activate
+          pip install -U pip setuptools six pytest 
           source share/spack/setup-env.sh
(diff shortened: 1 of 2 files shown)


haampie (dismissed):

alalazo:
This might be optimized further, as it doesn't strictly need concretization. For the time being let's just take it down from 20 secs to a couple of secs.

haampie (approved):

haampie (changes requested):
Needs to go after Python 2 is dropped and pytest is bumped

.github/workflows/unit_tests.yaml
```diff
@@ -144,6 +144,9 @@ jobs:
     - name: Run unit tests
       shell: runuser -u spack-test -- bash {0}
       run: |
+          /usr/libexec/platform-python -m venv platform-python
+          source platform-python/bin/activate
+          pip install -U pip setuptools six pytest
```
alalazo:
```suggestion
```

haampie (approved):

PR #33891, opened 2022-11-14
SingularityCE: Adding glib as a dependency 

usivakum:
As per the [SingularityCE Installation Guide](https://docs.sylabs.io/guides/main/admin-guide/installation.html#id1), `glib2` is a dependency, which was previously missing

### DIFF
--- a/var/spack/repos/builtin/packages/singularityce/package.py
+++ b/var/spack/repos/builtin/packages/singularityce/package.py
@@ -16,7 +16,8 @@ class SingularityBase(MakefilePackage):
     variant("network", default=True, description="install network plugins")
 
     depends_on("pkgconfig", type="build")
-    depends_on("glib")
+    depends_on("conmon", type=("build", "run"))
+    depends_on("squashfs", type=("build", "run"))
     depends_on("go@1.16:")
     depends_on("uuid")
     depends_on("libgpg-error")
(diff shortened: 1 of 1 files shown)


haampie:
Somehow it seems conmon is bundled as a binary, so adding it as a dependency is not helpful; it picks up system glib.

Maybe better is to add `--without-conmon` and add a build/run type dependency on the Spack version of it, so `depends_on("conmon", type=("build", "run"))`, since that one is fine:

```
harmen-desktop conmon-2.0.30-42gdgomn32fbeljwhh6ajys3ruf5wlgn $ ldd bin/conmon 
	linux-vdso.so.1 (0x00007ffd3138e000)
	libglib-2.0.so.0 => /home/harmen/spack/opt/spack/linux-ubuntu22.04-zen2/gcc-11.2.0/glib-2.74.1-nwkkuouyihqutidi7acjiwgzranzkvnn/lib/libglib-2.0.so.0 (0x00007f1829e3e000)
```

That might also allow you to drop the glib dep from `singularityce`, since it may only be necessary as a dependency for conmon?

Can you also add `squashfs` as a build/run-type dep?

```
checking: mksquashfs... yes (/usr/bin/mksquashfs)
```

usivakum:
@haampie Thanks for the pointers! I have added `conmon` as a dependency which resolves the `glib` issue. I have also added `squashfs` as a dependency as suggested

haampie (approved):

wdconinc:
This has caused a regression in apptainer, which does not like the `--without-conmon` flag in its [mconfig](https://github.com/apptainer/apptainer/blob/main/mconfig), since apptainer uses [singularityce]([url](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/apptainer/package.py)) as its base.

PR #33895, opened 2022-11-14
New Package: Prime95/Mprime

saqibkh:
Signed-off-by: saqibkh <saqibkhan@utexas.edu>

Prime95 is a CPU stress testing program. It tests your computer for
stability issues by stressing your CPU to its maximum limit. 
Prime95 runs indefinitely and only terminates a stress test when it
encounters an error and informs the user that the system may be unstable.

### DIFF
--- a/var/spack/repos/builtin/packages/prime95/package.py
+++ b/var/spack/repos/builtin/packages/prime95/package.py
@@ -1,0 +1,24 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Prime95(AutotoolsPackage):
+    """
+    Prime95, also distributed as the command-line utility mprime for FreeBSD
+    and Linux, is a freeware application. 
+
+    Prime95 is a CPU stress testing program. It tests your computer for
+    stability issues by stressing your CPU to its maximum limit. 
+    Prime95 runs indefinitely and only terminates a stress test when it
+    encounters an error and informs the user that the system may be unstable.
+    """
+
+    homepage = "https://www.mersenne.org/"
+    url = "https://www.mersenne.org/ftp_root/gimps/p95v308b17.linux64.tar.gz"
+    maintainers = ["saqibkh"]
+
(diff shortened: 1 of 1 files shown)


saqibkh:
I'm not sure if the CI check includes running the new package, but I have seen issues fetching the tar file on some systems.
"wget <tarfile>" works well on those systems, but I think there is an issue with python modules downloading the tarfile within spack.

Regardless, let me know if this is causing an issue while installing the package.

Thanks

var/spack/repos/builtin/packages/prime95/package.py
```diff
@@ -0,0 +1,24 @@
...
+    stability issues by stressing your CPU to its maximum limit.
+    Prime95 runs indefinitely and only terminates a stress test when it
+    encounters an error and informs the user that the system may be unstable.
+    """
+
+    homepage = "https://www.mersenne.org/"
+    url = "https://www.mersenne.org/ftp_root/gimps/p95v308b17.linux64.tar.gz"
+    maintainers = ["saqibkh"]
+
+    version("30.8b17", sha256="5180c3843d2b5a7c7de4aa5393c13171b0e0709e377c01ca44154608f498bec7")
```
tldahlgren: I cannot confirm the checksum (`spack checksum`) using either of the supported Spack url fetch methods:  `urllib` (default) or `curl`.

tldahlgren: And I get `All fetchers failed` when I run `spack fetch prime95` and `spack stage prime95`.

saqibkh: Thanks @tldahlgren 

The command "spack checksum" helped fix the version issue which was the root cause

var/spack/repos/builtin/packages/prime95/package.py
```diff
@@ -6,7 +6,7 @@
 from spack.package import *
 
 
-class Prime95(AutotoolsPackage):
+class Prime95(Package):
```
tldahlgren, line 9: Curious.  Why switch from `AutotoolsPackage`?

tldahlgren (approved):
Confirmed the 5 sha256.

var/spack/repos/builtin/packages/prime95/package.py
saqibkh: AutotoolsPackage forces an "autoreconf" stage that isn't valid here.
I could have overwritten that function, but felt it was simple to just switch to "Package" class instead.

Thanks

PR #33898, opened 2022-11-14
Python: drop EOL versions

adamjstewart:
This PR removes [end of life](https://endoflife.date/python) versions of Python from Spack. Specifically, this includes all versions of Python older than 3.7.

See https://github.com/spack/spack/discussions/31824 for rationale. Deprecated in #32615. And #28003.

For anyone using software that relies on Python 2, you have a few options:

* Upgrade the software to support Python 3. The `3to2` tool may get you most of the way there, although more complex libraries may need manual tweaking.
* Add Python 2 as an [external package](https://spack.readthedocs.io/en/latest/build_settings.html#external-packages). Many Python libraries do not support Python 2, but you may be able to add older versions that did once upon a time.
* Use Spack 0.19. Spack 0.19 is the last release to officially support Python 3.6 and older
* Create and maintain your own [custom repository](https://spack.readthedocs.io/en/latest/repositories.html). Basically, you would need a package for Python 2 and any other Python 2-specific libraries you need.


haampie:
@spackbot fix style

adamjstewart:
@spackbot maintainers

adamjstewart:
Too many packages for the bot, so here's a manual ping. First half:

@AndrewGaspar @BoboTiG @FrankD412 @G-071 @HadrienG2 @HaluskaR @Kerilk @KineticTheory @MakisH @RemiLacroix-IDRIS @Sinan81 @adamjstewart @agoodLANL @ajkotobi @amd-toolchain-support @apdavison @arjun-raj-kuppala @aweits @balay @barrysmith @bernhardkaindl @boegel @bryanherman @btovar @bvanessen @calebrob6 @cchambreau @charlesg3 @chenwany @chuckatkins @citibeth @cosmicexplorer @danlipsa @demartinofra @dev-zero @dhuth @dorton21 @doutriaux1 @drbenmorgan @dthain @emai-imcs @enrico-usai @eschnett @estebanpauli @estewart08 @francesco-giordano @fsimonis @glennpj @gmarciani @haampie

adamjstewart:
Second half:

@hanwen-pcluste @jcbowden @jeanbez @jedbrown @jke513 @joseeroman @jppelteret @kwryankrattiger @lee218llnl @leiterrl @likask @liuyangzhuan @luca-heltai @lukeseawalker @marcusboden @matthiasdiener @mhaseeb123 @migueldiascosta @murray55 @mwkrentel @nessiecancode @pcdslab @pfluegdk @pradyunsg @qobilidop @renjithravindrankannath @rgommers @robertu94 @robqiao @scemama @scheibelp @sethrj @shanedsnyder @skosukhin @smolins @snehring @srekolam @sshudler @svenevs @takluyver @tgaddair @trws @tzanio @v-dobrev @vicentebolea @vvolkl

Can y'all review this PR? It modifies (or in many cases removes) a package for which you are listed as a maintainer.

v-dobrev (dismissed):

migueldiascosta (dismissed):
regarding the changes to the berkeleygw package, lgtm (I'll separately look into adding back the dependencies using python 3)

pradyunsg (dismissed):
Yes please. :)

var/spack/repos/builtin/packages/petsc/package.py
```diff
@@ -230,8 +203,7 @@ def check_fortran_compiler(self):
     depends_on("rocprim", when="+rocm")
 
     # Build dependencies
-    depends_on("python@2.6:2.8", type="build", when="@:3.10")
-    depends_on("python@2.6:2.8,3.4:", type="build", when="@3.11:")
+    depends_on("python@2.6:2.8,3.4:", type="build")
```
jedbrown, line 206: Should this just say python3 is required?

jedbrown (dismissed):

HadrienG2 (dismissed):

srekolam (dismissed):
reviewed roctracer-dev and looks okay to me.

BoboTiG (dismissed):

joseeroman (dismissed):

marcusboden (dismissed):

apdavison (dismissed):

rgommers (dismissed):
Looks like a very good idea. I checked the packages I maintain: `py-numpy` looks good, `py-scipy` isn't touched in this PR. Not sure if that was intended; old versions can be dropped there as well.

jcbowden (dismissed):

btovar (dismissed):

drbenmorgan (dismissed):

haampie:
The giant number of :heavy_check_mark:s on this PR remind me of this scene https://youtu.be/rrZBpTZmGfI?t=239

matthiasdiener (dismissed):

MakisH (dismissed):

var/spack/repos/builtin/packages/easybuild/package.py
```diff
@@ -16,14 +16,12 @@ class Easybuild(PythonPackage):
     maintainers = ["boegel"]
 
     version("4.0.0", sha256="21bcc1048525ad6219667cc97a7421b5388068c670cabba356712e474896de40")
-    version("3.1.2", sha256="a966a08e41d3557d7fd3dc5322180c006d27281d9ecc1b1fce1b4518c32664d4")
 
-    depends_on("python@2.6:2.8", when="@:3", type=("build", "run"))
     depends_on("python@2.6:2.8,3.5:", when="@4:", type=("build", "run"))
```
boegel, line 20: @adamjstewart Should this be changed to `3.5:`? Likewise in `py-easybuild-*`

snehring (dismissed):

var/spack/repos/builtin/packages/petsc/package.py
adamjstewart: Technically, it supports all versions of Python that Spack supports, so we could just get rid of all version specifications. There are thousands of packages where this is true, so I haven't bothered to update all of them.

var/spack/repos/builtin/packages/easybuild/package.py
adamjstewart: Technically, it supports all versions of Python that Spack supports, so we could just get rid of all version specifications. There are thousands of packages where this is true, so I haven't bothered to update all of them.

adamjstewart:
Visit build failure in pipeline:
```
/usr/bin/ld: cannot find -lvtkGUISupportQtOpenGL
collect2: error: ld returned 1 exit status
```
@cyrush @chuckatkins @danlipsa @sethrj can you look into this and see if I broke anything in visit/vtk/qt?

EDIT: CI seems to pass on the latest commit, so maybe it's transient?

boegel (dismissed):

agoodLANL (dismissed):

var/spack/repos/builtin/packages/petsc/package.py
trws: That actually is an interesting point, would it be worth having a variable in spack for that version range? Especially for applying to other repo versions than current it might be a nice helper to have around.

var/spack/repos/builtin/packages/ganglia/package.py
```diff
@@ -1,35 +0,0 @@
-# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```
trws, line 1: It probably shouldn't have, but this package being in here surprised me.  If it's unused/unmaintained, leave it, but if there's a maintainer lurking about there have been patches for this for a while, like this one: https://github.com/ganglia/monitor-core/pull/313

var/spack/repos/builtin/packages/petsc/package.py
adamjstewart: > That actually is an interesting point, would it be worth having a variable in spack for that version range?

You mean the range of Python versions that Spack currently supports? Or the range of versions needed by this package?

> Especially for applying to other repo versions than current it might be a nice helper to have around.

You mean for external repos (not builtin), or for people using older core with newer builtin, or both? I have no idea how to support the former, and the latter isn't currently supported, although that's something we plan on doing before a 1.0 release.

trws (dismissed):
I spot checked everything I can recall having a hand in maintaining, and several others.  This looks *great*!

mhaseeb123 (approved):

var/spack/repos/builtin/packages/petsc/package.py
trws: I meant the range that spack supports, in terms of "supported by the python/package.py file in the spack repo in use".  It's a property of the python package we have, that could reasonably be used by another package, but it's easy enough to end up with a version of  python in spack that's outside that range that specifying it still seems useful.  Does that make more sense?

I mainly meant for external repos, or dealing with pythons built with older repo versions or brought in as externals such that they aren't accounted for in the python package.

var/spack/repos/builtin/packages/hicops/package.py
```diff
@@ -81,11 +81,9 @@ class Hicops(CMakePackage):
     depends_on("py-setuptools")
     depends_on("py-bottleneck")
     depends_on("py-pyparsing")
-    depends_on("py-subprocess32")
     depends_on("py-six")
     depends_on("py-setuptools-scm")
     depends_on("py-et-xmlfile")
-    depends_on("py-argparse")
```
mhaseeb123, line 88: Not sure if these packages (py-subprocess32 and py-argparse) depend on an EOL version of Python. The minimum version of python required for hicops package is 3.7 which should have be fine?

var/spack/repos/builtin/packages/petsc/package.py
balay: > I meant the range that spack supports, in terms of "supported by the python/package.py file in the spack repo in use

Welll - I'm usually able to `spack install pkg@ver` - even though this `ver` is not listed in pkg/package.py [assuming the tarball exists - the sha256 check gets skipped]

trws: Yes, which means that leaving a range unexpressed on the theory that "every version specified in spack works" leaves some room for badness.

var/spack/repos/builtin/packages/hicops/package.py
adamjstewart: Argparse is part of the standard library in Python 2.7 and 3.2+, the `py-argparse` package is merely a backport of the standard library and is only needed when using Python 2.6 or 3.0–3.1. Hicops should have never listed this dependency to begin with since it uses Python 3.7+. 

P.S. All of the dependency types listed in hicops are incorrect. These should likely be `type=('build', 'run')`, the default is `type=('build', 'link')`. This means that when you `spack load hicops`, none of these dependencies are present in `PYTHONPATH`. But we can fix this in a separate PR.

mwkrentel:
@adamjstewart 

1. Just to be clear, after this PR, the minimum python version will
become 3.7, is that right?

(Actually, if I'm reading the patch correctly, this just gets all the
packages ready for 3.7, the actual switch would come in a later patch.)

2. Will there be a lower version of python that could be used to
bootstrap to 3.7, the way 2.7 works now?

3. How long do you expect spack to allow 3.7?
Is this going to become 3.8 in 7 months?

I have to build on a lot of national labs machines that are (1) old,
and (2) I don't have root.  I expect that I can cope, but I do need to
know what to expect.

Thanks

var/spack/repos/builtin/packages/hicops/package.py
mhaseeb123: Thank you for clarifying @adamjstewart. We are planning to revamp the entire hicops package here once we merge the dev into production as there have been major changes in the dependencies! We can go ahead with these changes for now and can look at it later like you suggested,

var/spack/repos/builtin/packages/py-resultsfile/package.py
```diff
@@ -16,9 +16,7 @@ class PyResultsfile(PythonPackage):
     maintainers = ["scemama"]
 
     version("2.0", sha256="2a34208254e4bea155695690437f6a59bf5f7b0ddb421d6c1a2d377510f018f7")
-    version("1.0", sha256="e029054b2727131da9684fa2ec9fb8b6a3225dc7f648216a9390267b2d5d60c3")
 
-    depends_on("python@2.7:2.8", type=("build", "run"), when="@1.0:1")
-    depends_on("python@3:", type=("build", "run"), when="@2.0:")
+    depends_on("python@3:", type=("build", "run"))
     # pip silently replaces distutils with setuptools
     depends_on("py-setuptools", type="build")
```
scemama, line 22: OK!

var/spack/repos/builtin/packages/py-irpf90/package.py
```diff
@@ -1,23 +0,0 @@
...
-
-    homepage = "http://irpf90.ups-tlse.fr"
-    pypi = "irpf90/irpf90-1.7.7.tar.gz"
-
-    maintainers = ["scemama"]
-
-    version("1.7.7", sha256="c6b2eecb9180f1feaab9644bbed806637a4a30a0fad2c4775a985fcc01a99530")
-
-    depends_on("python@2.7.0:2.8", type=("build", "run"))
-    depends_on("py-setuptools", type="build")
```
scemama, line 23: I have done the 2->3 transition of IRPF90. I will make a PR with the update for that one.

mhaseeb123 (dismissed):

adamjstewart:
@mwkrentel I think the support timeline in #32615 should answer most of your questions, but I'll clarify this here. Note that there are actually two separate things you are asking about:

1. The versions of Python that can be used to run Spack
2. The versions of Python that Spack can install

These versions are not the same, so it's important to distinguish between 1 and 2.

> Just to be clear, after this PR, the minimum python version will become 3.7, is that right?

1. As of #33063, Python 3.6+ is the minimum version of Python that can be used to run Spack
2. As of this PR, Python 3.7+ is the minimum version of Python that Spack can install

> Will there be a lower version of python that could be used to bootstrap to 3.7, the way 2.7 works now?

Since 1 and 2 are different, you can currently use Python 3.6+ to run Spack and install Python 3.7+. If you don't have Python 3.6+ available, you can use Spack 0.19 and older to install it (or any other package manager).

> How long do you expect spack to allow 3.7? Is this going to become 3.8 in 7 months?

1. We don't have a specific timeline in mind for versions of Python that can be used to run Spack. This will likely lag behind 2 due to older OS support
2. For installing with Spack, I intend to follow https://endoflife.date/python. So you will no longer be able to install Python 3.7 with Spack develop after June 2023.

Let me know if you have any other questions or concerns!

var/spack/repos/builtin/packages/petsc/package.py
adamjstewart: Yep, so I'll leave these things in here for now until people either give up on using EOL Python or create their own repo for Python 2 support.

adamjstewart: I have no intention nor capability of supporting external repos, it's up to the maintainers of external repos to support builtin imo.

scemama (dismissed):

adamjstewart:
@spackbot fix style

var/spack/repos/builtin/packages/py-darshan/package.py
```diff
@@ -18,7 +18,6 @@ class PyDarshan(PythonPackage):
 
     depends_on("python@3.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
-    depends_on("py-importlib-resources", when="^python@3.6", type=("build", "run"))
```
shanedsnyder, line 21: There actually turns out to be an error in this version constraint on our end. importlib-resources is required for all versions less than 3.9, not just version 3.6. Could we just change this to:

`when="^python@:3.8"`

That would correct the erroneous version constraint and support the changes attempted here related to EOL.

shanedsnyder (changes requested):

adamjstewart:
> `py-scipy` isn't touched in this PR. Not sure if that was intended; old versions can be dropped there as well.

@rgommers which versions of scipy do not support Python 3.7+?

var/spack/repos/builtin/packages/py-darshan/package.py
adamjstewart: Done

shanedsnyder: Just to clarify, I think we want: `@:3.8` (not just 3.8)

adamjstewart: Also done

rgommers:
> @rgommers which versions of scipy do not support Python 3.7+?

SciPy 1.1.0 was the first release to support Python 3.7, so everything older than that can be removed.

shanedsnyder (dismissed):

doutriaux1 (dismissed):

var/spack/repos/builtin/packages/vtk/package.py
```diff
@@ -65,8 +65,6 @@ class Vtk(CMakePackage):
     with when("+python"):
         # Depend on any Python, add bounds below.
         extends("python@2.7:", type=("build", "run"))
```
kwryankrattiger, line 67: This could probably just be `extends("python")` now correct?

var/spack/repos/builtin/packages/sensei/package.py
```diff
@@ -55,7 +55,6 @@ class Sensei(CMakePackage):
     depends_on("paraview+mpi", when="+catalyst")
     depends_on("paraview+hdf5", when="+catalyst+hdf5")
     depends_on("paraview+python", when="@:2 +catalyst+python")
```
kwryankrattiger: Remove the version constraint here or change it to `@3:`

var/spack/repos/builtin/packages/vtk/package.py
adamjstewart: Technically, it supports all versions of Python that Spack supports, so we could just get rid of all version specifications. There are thousands of packages where this is true, so I haven't bothered to update all of them.

var/spack/repos/builtin/packages/sensei/package.py
adamjstewart: Done

bvanessen (dismissed):

adamjstewart:
@spackbot run pipeline

vicentebolea (dismissed):

kwryankrattiger (dismissed):

var/spack/repos/builtin/packages/qt/package.py
```diff
@@ -148,7 +63,7 @@ class Qt(Package):
     variant("sql", default=True, description="Build with SQL support.")
     variant("ssl", default=True, description="Build with OpenSSL support.")
     variant("tools", default=True, description="Build tools, including Qt Designer.")
-    variant("webkit", default=False, description="Build the Webkit extension")
+    variant("webkit", default=False, description="Build the Webkit extension", when="@:5.6,5.16:")
```
sethrj: I think a `conflicts` (indicating "this should work but doesn't") is more appropriate than a `when` (indicating "this feature is version-specific"). I think the variant should be restored and a `conflicts("+webkit", when="@5.7:5.15", msg="qtwebengine@5.7:5.15 are based on Google Chromium versions which depend on Py2")` added instead.

glennpj:
@adamjstewart I have made a couple of passes over this and it seems fine. That is a lot of clean-up! Are there any particular things that you want people to look for? If I get a chance I may try to build a few things from this PR as a check.

adamjstewart:
@glennpj nothing in particular I'm worried about, a simple spot check of all packages should suffice. If you can figure out why CI is failing that would be great, see https://github.com/spack/spack/pull/33898#issuecomment-1315540661.

adamjstewart:
@tgamblin note that once this PR is merged, you're going to see a whole lot of:
```
==> Error: Package 'py-pil' not found.
```
questions on Slack. Luckily people seem to be blindly following the instructions here, but it would be great if this bug was fixed.

adamjstewart:
Just rebuilt my personal Python stack which includes about 200 Python packages using this branch, no problems with any of those as far as I can tell.

sethrj (dismissed):
qt and py-h5sh changes look good

jppelteret (dismissed):
Changes to the deal.II package look fine, thanks. I see that you've removed these versions from the PETSc and SLEPc packages.

var/spack/repos/builtin/packages/gdb/package.py
```diff
@@ -37,7 +37,7 @@ class Gdb(AutotoolsPackage, GNUMirrorPackage):
     version("7.11.1", sha256="57e9e9aa3172ee16aa1e9c66fef08b4393b51872cc153e3f1ffdf18a57440586")
     version("7.10.1", sha256="ff14f8050e6484508c73cbfa63731e57901478490ca1672dc0b5e2b03f6af622")
 
-    variant("python", default=True, description="Compile with Python support")
+    variant("python", default=True, description="Compile with Python support", when="@8.2:")
```
robertu94, line 40: GDB release notes say that 9.1+ is required for Windows to build with Python 3

robertu94 (changes requested):
@adamjstewart one small changes for GDB

var/spack/repos/builtin/packages/gdb/package.py
adamjstewart: Is that only on Windows or on all systems? Can you share a link?

adamjstewart: At the moment, the GDB package builds with Autotools, so we don't yet support Windows for this package.

jeanbez (approved):

tgamblin (approved):

var/spack/repos/builtin/packages/gdb/package.py
robertu94: Just windows: https://sourceware.org/gdb/news/ | grep 9.1

adamjstewart: Well I'll let someone update this when they add Windows support for GDB.

tldahlgren:
@scheibelp As investigated during the Spack User meeting, this PR added a `conflicts` to `qt+webkit` that precludes building the only version of `rstudio` in the Spack package.

PR #33900, opened 2022-11-14
Fix incorrect timer

haampie:
Before this PR, the timer for a phase would effectively stop *before* it started.
Effectively all entries in `timers.json` were off by one.

Just moving the `timer.phase(...)` bit after the phase ends is not enough, since
there is no way to start the time as the phase starts; without that we would
include the time spent on fetching etc too. To solve this, we now simply have
a `timer.start("name")` and `timer.stop("name")` API, which is much easier
to follow than the current odd implementation where a `timer.phase("name")`
should be called _after_ completion.

I've also added a context manager so you can do:

```python
with timer.measure("phase"):
  # block
```

But I don't wanna use it cause it's only cluttering the stacktrace and nesting
code unnecessarily.

It would be nice if this was backported, but at the same time it has changed the
timer API.

### DIFF
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -321,9 +321,9 @@ def _install_from_cache(pkg, cache_only, explicit, unsigned=False):
     _print_timer(
         pre=_log_prefix(pkg.name),
         pkg_id=pkg_id,
-        fetch=timer.phases.get("search", 0) + timer.phases.get("fetch", 0),
-        build=timer.phases.get("install", 0),
-        total=timer.total,
+        fetch=timer.duration("search") + timer.duration("fetch"),
+        build=timer.duration("install"),
+        total=timer.duration(),
     )
     _print_installed_pkg(pkg.spec.prefix)
     spack.hooks.post_install(pkg.spec)
@@ -391,11 +391,11 @@ def _process_binary_cache_tarball(
         bool: ``True`` if the package was extracted from binary cache,
             else ``False``
     """
+    timer and timer.start("fetch")
     download_result = binary_distribution.download_tarball(
         binary_spec, unsigned, mirrors_for_spec=mirrors_for_spec
     )
-    if timer:
-        timer.phase("fetch")
+    timer and timer.stop("fetch")
     # see #10063 : install from source if tarball doesn't exist
     if download_result is None:
         tty.msg("{0} exists in binary cache but with different hash".format(pkg.name))
@@ -405,6 +405,7 @@ def _process_binary_cache_tarball(
     tty.msg("Extracting {0} from binary cache".format(pkg_id))
 
     # don't print long padded paths while extracting/relocating binaries
+    timer and timer.start("install")
     with spack.util.path.filter_padding():
         binary_distribution.extract_tarball(
             binary_spec, download_result, allow_root=False, unsigned=unsigned, force=False
@@ -412,8 +413,7 @@ def _process_binary_cache_tarball(
 
     pkg.installed_from_binary_cache = True
     spack.store.db.add(pkg.spec, spack.store.layout, explicit=explicit)
-    if timer:
-        timer.phase("install")
+    timer and timer.stop("install")
     return True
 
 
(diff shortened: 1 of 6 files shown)


psakievich (dismissed):

vsoch:
This depends on how we want the timer phases to be represented. The current way has the times as the start of each phase, and this change would move them to be the end of the phase (and the reader would need to derive that). I'm okay with either, but I'd say there should be clear labels / docs to indicate exactly what is happening.

haampie:
I think it's to be expected that `configure: 1m 3s` means we spend 1 minute and 3 seconds in the configure phase, etc.

So, maybe I should just refactor s.t. the API is `timer.start("phase")` / `timer.stop("phase")` or `with timer.phase(...): do_stuff`. This issue right now is there is no way to get the first or last interval correct.

lib/spack/spack/test/install.py
```diff
@@ -252,12 +252,8 @@ def test_install_times(install_mockery, mock_fetch, mutable_mock_repo):
 
     # The order should be maintained
     phases = [x["name"] for x in times["phases"]]
-    total = sum([x["seconds"] for x in times["phases"]])
-    for name in ["one", "two", "three", "install"]:
-        assert name in phases
-
-    # Give a generous difference threshold
-    assert abs(total - times["total"]["seconds"]) < 5
```
alalazo, line 260: :+1:

lib/spack/spack/installer.py
```diff
@@ -391,11 +391,11 @@ def _process_binary_cache_tarball(
         bool: ``True`` if the package was extracted from binary cache,
             else ``False``
     """
+    timer and timer.start("fetch")
```
alalazo: Not sure how idiomatic this is line is. It is effectively equal to:
```python
if timer:
    timer.start("fetch")
```
Would it be easy to code a `NullTimer` with the same interface but that performs no-op? If it is we can just:
```python
timer.start("fetch")
```
and set a NULL_TIMER object as the default value.

lib/spack/spack/util/timer.py
```diff
@@ -11,51 +11,103 @@
...
-        if self.end:
-            return self.end - self.start
-        return time.time() - self.start
+    def __init__(self, now=time.time):
+        """
+        Arguments:
+            now: function that gives the seconds since e.g. epoch
+        """
+        self._now = now
+        self._timers = OrderedDict()  # type: OrderedDict[str,Interval]
```
alalazo, line 68: Since we support Python 3.6+ now we don't strictly need an `OrderedDict`, though if we backport this change we would.

lib/spack/spack/util/timer.py
```diff
@@ -11,51 +11,103 @@
...
+
+        # _global is the overal timer since the instance was created
+        self._timers["_global"] = Interval(self._now(), end=None)
```
alalazo: Can we reference `"_global"` in a variable instead of using literal string in multiple places?

alalazo (changes requested):
Basically LGTM. Only request on `_global`. If we can do something for the `and` idiom it would be a plus, but not required.

alalazo:
There are also a few formatting errors caught by docs:
```
/home/docs/checkouts/readthedocs.org/user_builds/spack/checkouts/33900/lib/spack/docs/_spack_root/lib/spack/spack/util/timer.py:docstring of spack.util.timer.Timer.start:3: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/spack/checkouts/33900/lib/spack/docs/_spack_root/lib/spack/spack/util/timer.py:docstring of spack.util.timer:1: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/spack/checkouts/33900/lib/spack/docs/_spack_root/lib/spack/spack/util/timer.py:docstring of spack.util.timer.Timer.stop:4: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/spack/checkouts/33900/lib/spack/docs/_spack_root/lib/spack/spack/util/timer.py:docstring of spack.util.timer:2: WARNING: Block quote ends without a blank line; unexpected unindent.
looking for now-outdated files... none found
```

lib/spack/spack/installer.py
haampie: Let me do null timer

lib/spack/spack/util/timer.py
haampie: I still think we wanna fix this bug in 0.19.1, we can backport first and then improve ;p

alalazo (approved):

PR #33904, opened 2022-11-14
[glib] Spack cannot find sources without list_url

stephenmsachs:
Download times out after 0 seconds.


michaelkuhn:
What exactly does this fix for you? I get the same `spack versions glib` output with and without `list_url`.

stephenmsachs:
> What exactly does this fix for you?

I was running `spack install py-pymol` and ran into a timeout error when installing `glib`.

michaelkuhn:
That doesn't look like something that `list_url` would fix but rather a transitory problem with the server. To make `list_url` actually do something, I believe we also need `list_depth`:
```
list_url = "https://download.gnome.org/sources/glib/"
list_depth = 1
```

michaelkuhn (dismissed):

bernhardkaindl:
> I was running `spack install py-pymol` and ran into a timeout error when installing `glib`.

gnome.org is partly offline, access to https://developer.gnome.org/ was just restored, but https://wiki.gnome.org/Projects/GLib is still (or: again) broken. We just updated the download url in #33919, I force-pushed to change suggested by michaelkuhn now to list the versions, works for me.

michaelkuhn (approved):

bernhardkaindl (approved):

PR #33905, opened 2022-11-14
Stop using `six` in Spack

alalazo:
Since we dropped support for Python 2.7, there's no need so use `six` anymore. We still need to vendor it until we update our vendored dependencies.


alalazo:
This, on my local machine, also solves all the issues we had in unit test, reporting at the end:
```
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name=19 mode='r+' encoding='UTF-8'>
```
(but apparently it does not on GA)

haampie (approved):

haampie:
Relying on the tests being :heavy_check_mark:

lib/spack/llnl/util/filesystem.py
```diff
@@ -2121,7 +2119,7 @@ def find_libraries(libraries, root, shared=True, recursive=False, runtime=True):
     Returns:
         LibraryList: The libraries that have been found
     """
-    if isinstance(libraries, six.string_types):
```
trws, line 2124: I see this is merged, and I'm guessing this passed all the tests, but are we sure nobody is going to pass a path as bytes?

alalazo: We don't support Python 2.7 on `develop` anymore, so this shouldn't be a change in functional behavior, given that `six` docs says:

> Possible types for text data. This is [basestring()](https://docs.python.org/2/library/functions.html#basestring) in Python 2 and str in Python 3.

trws: Gotcha, thanks! 👍

PR #33914, opened 2022-11-15
FastJET: add extra variants

iarspider:
plugins, thread-safety, python interface

### DIFF
--- a/var/spack/repos/builtin/packages/fastjet/package.py
+++ b/var/spack/repos/builtin/packages/fastjet/package.py
@@ -75,7 +75,7 @@ class Fastjet(AutotoolsPackage):
     variant("auto-ptr", default=False, description="Use auto_ptr")  
     variant("atlas", default=False, description="Patch to make random generator thread_local")
 
-        variant(
+    variant(
 	"plugins",
         values=disjoint_sets(("all",), ("allcxx",), plugins_)
         .prohibit_empty_set()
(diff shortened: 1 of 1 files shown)


iarspider:
@spackbot fix style

var/spack/repos/builtin/packages/fastjet/package.py
```diff
@@ -56,15 +56,76 @@ class Fastjet(AutotoolsPackage):
...
     variant("auto-ptr", default=False, description="Use auto_ptr")
     variant("atlas", default=False, description="Patch to make random generator thread_local")
 
+    variant(
+        "plugins",
+        values=disjoint_sets(("all",), ("allcxx",), plugins_)
+        .prohibit_empty_set()
+        .with_default("all"),
+        description="List of plugins to build. allcxx is all minus PxCone.",
+    )
```
drbenmorgan, line 84: What's the use case for providing variants for each plugin rather than all of them always? It _feels_ simpler to always have all of them enabled and keep the variants simple, but I'm likely missing something.

iarspider: @drbenmorgan removing unused plugins to save disk space and network bandwidth

drbenmorgan: Thanks @iarspider, but what are we talking in terms of disk space for an "all" vs "none" build?

iarspider: @drbenmorgan 
`fastjet~atlas+python thread-safety=limited plugins=CMSIterativeCone`: 72,328 bytes
`fastjet~atlas+python thread-safety=limited plugins=allcxxplugins` (this is what CMS uses): 102,680 bytes 
`fastjet~atlas+python thread-safety=limited plugins=all`: 103,372 bytes

drbenmorgan: Thanks! Whilst 100MB overall is fairly large, and a 30MB saving is useful, I'm not sure it's enough to justify having multiple variants for the plugins. Doing the "everything" build feels like the better option both from end user (simplicity, expectations) and packager (simpler recipe, fewer potential installs from plugin variance).

What do you think @vvolkl?

PR #33916, opened 2022-11-15
Add v0.8 to Arbor

brenthuisman:


var/spack/repos/builtin/packages/arbor/package.py
```diff
@@ -53,20 +58,23 @@ class Arbor(CMakePackage, CudaPackage):
     )
 
     # https://docs.arbor-sim.org/en/latest/install/build_install.html#compilers
-    conflicts("%gcc@:8.3")
-    conflicts("%clang@:7")
+    conflicts("%gcc@:8")
```
schmitts, line 61: we restrict more now?

var/spack/repos/builtin/packages/arbor/package.py
```diff
@@ -53,20 +58,23 @@ class Arbor(CMakePackage, CudaPackage):
...
+    conflicts("%clang@:9")
     # Cray compiler v9.2 and later is Clang-based.
     conflicts("%cce@:9.1")
     conflicts("%intel")
 
-    depends_on("cmake@3.12:", type="build")
+    depends_on("cmake@3.19:", type="build")
```
schmitts, line 67: really necessary?

var/spack/repos/builtin/packages/arbor/package.py
brenthuisman: Yes, as of [this PR](https://github.com/arbor-sim/arbor/pull/2017) minimum versions for compilers were bumped. Is that a problem? I asked in the channel, and nobody objected (otherwise I would not have merged this). One reason to bump is to be able to support CUDA 11.

schmitts: Sorry, I was misreading conflicts vs. depends_on. It's fine.

var/spack/repos/builtin/packages/arbor/package.py
brenthuisman: I can't recall in which of the many CMake troubleshooting sessions this came up, but it was an oversight that it wasn't bumped last release, we had not been testing below 3.18 then either. If you want me to dig, I can find out what exactly we use. Does it present a problem for you? (If so, please file an issue.)

schmitts: Just out of curiosity. All fine.

tldahlgren (approved):
Confirmed new sha256, rest LGTM

PR #33918, opened 2022-11-15
Run Python 3.6 unit tests on ubuntu-20.04

alalazo:
GA runners have started rolling over [this change](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/) that makes `ubuntu-22.04` the image used for `ubuntu-latest`. Unfortunately `ubuntu-22.04` doesn't support Python 3.6 with `actions/setup-python`, so we need to start using `ubuntu-20.04` explicitly for that.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   # Run unit tests with different configurations on linux
   ubuntu:
-    runs-on: $${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/unit_tests.yaml
```diff
@@ -11,28 +11,38 @@ concurrency:
 jobs:
   # Run unit tests with different configurations on linux
   ubuntu:
-    runs-on: ubuntu-latest
+    runs-on: $${{ matrix.os }}
```
alalazo:
```suggestion
    runs-on: ${{ matrix.os }}
```

haampie (approved):

PR #33925, opened 2022-11-15
Update dask and related packages

charmoniumQ:
Also see: https://github.com/dask/distributed/pull/7309


var/spack/repos/builtin/packages/py-dask/package.py
```diff
@@ -14,6 +14,7 @@ class PyDask(PythonPackage):
 
     maintainers = ["skosukhin"]
 
+    version("2022.10.2", sha256="42cb43f601709575fa46ce09e74bea83fdd464187024f56954e09d9b428ceaab")
     version("2021.6.2", sha256="8588fcd1a42224b7cfcd2ebc8ad616734abb6b1a4517efd52d89c7dd66eb91f8")
     version("2021.4.1", sha256="195e4eeb154222ea7a1c368119b5f321ee4ec9d78531471fe0145a527f744aa8")
     version("2020.12.0", sha256="43e745afd4b464e6c0113131e430a16dce6ac42460b06e24d799093d098f7ab0")
```
adamjstewart, line 20: Need to update deps, check `setup.py` for a list

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -59,6 +60,7 @@ class PyDistributed(PythonPackage):
     depends_on("py-tblib@1.6.0:", type=("build", "run"), when="@2.11.0:")
```
adamjstewart, line 71: Missing dep on locket >= 1

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -59,6 +60,7 @@ class PyDistributed(PythonPackage):
...
     depends_on("py-toolz@0.7.4:", type=("build", "run"), when="@:2.12.0")
     depends_on("py-toolz@0.8.2:", type=("build", "run"), when="@2.13.0:")
+    depends_on("py-toolz@0.10.0:", type=("build", "run"), when="@2022.10.2:")
```
adamjstewart, line 76: toolz 0.8.2+ is still correct for this version, are you looking at master?

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -59,6 +60,7 @@ class PyDistributed(PythonPackage):
...
     depends_on("py-tornado@5:", type=("build", "run"), when="^python@:3.7")
     depends_on("py-tornado@6.0.3:", type=("build", "run"), when="^python@3.8:")
```
adamjstewart, line 78: Latest version has `torndao@6.0.3:6.1` without any restriction on Python version

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -59,6 +60,7 @@ class PyDistributed(PythonPackage):
...
     depends_on("py-zict@0.1.3:", type=("build", "run"))
```
adamjstewart, line 80: Missing dep on urllib3

var/spack/repos/builtin/packages/py-fsspec/package.py
```diff
@@ -12,6 +12,7 @@ class PyFsspec(PythonPackage):
     homepage = "https://github.com/intake/filesystem_spec"
     pypi = "fsspec/fsspec-0.4.4.tar.gz"
 
+    version("2022.11.0", sha256="259d5fd5c8e756ff2ea72f42e7613c32667dc2049a4ac3d84364a7ca034acb8b")
     version("2021.7.0", sha256="792ebd3b54de0b30f1ce73f0ba0a8bcc864724f2d9f248cb8d0ece47db0cbde8")
     version("2021.4.0", sha256="8b1a69884855d1a8c038574292e8b861894c3373282d9a469697a2b41d5289a6")
     version("0.9.0", sha256="3f7a62547e425b0b336a6ac2c2e6c6ac824648725bc8391af84bb510a63d1a56")
```
adamjstewart, line 18: Python 3.7+ required in this version

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-distributed/package.py
charmoniumQ: That is a bug. See https://github.com/dask/distributed/pull/7309

adamjstewart: Good catch, can you add a comment with a link to this PR?

charmoniumQ:
Comments resolved. For these four packages, I added all of the deps in the `setup.py`s of that package's pinned version.

var/spack/repos/builtin/packages/py-dask/package.py
```diff
@@ -56,6 +57,7 @@ class PyDask(PythonPackage):
     depends_on("py-fsspec@0.6.0:", type=("build", "run"), when="@2021.3.1:")
```
adamjstewart, line 84: Specific versions of pyyaml required now

adamjstewart: Python 3.8+ required now

var/spack/repos/builtin/packages/py-dask/package.py
```diff
@@ -152,6 +157,7 @@ class PyDask(PythonPackage):
         "py-distributed@2020.12.0:2021.8.0", type=("build", "run"), when="@2020.12.0: +distributed"
```
adamjstewart: These older versions should have upper bounds in `when`, otherwise they will also apply to the latest version and will conflict with the requirements of newer versions. You should find that it isn't actually possible to concretize `py-dask@2022.10.2+distributed` because of this.

var/spack/repos/builtin/packages/py-dask/package.py
```diff
@@ -152,6 +157,7 @@ class PyDask(PythonPackage):
...
     )
     depends_on("py-distributed@2021.6.2", type=("build", "run"), when="@2021.6.2 +distributed")
+    depends_on("py-distributed@2022.10.2", type=("build", "run"), when="@2022.10.2 +distributed")
 
     # Requirements for dask.diagnostics
     depends_on("py-bokeh@1.0.0:", type=("build", "run"), when="@2.0.0: +diagnostics")
```
adamjstewart, line 198: Bokeh version required changed, new jinja2 dep required for diagnostics

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -43,28 +44,39 @@ class PyDistributed(PythonPackage):
...
     depends_on("py-jinja2", type=("build", "run"), when="@2022.2.1:")
+    depends_on("py-locket@1:", type=("build", "run"), when="@2022.2.1:")
     depends_on("py-contextvars", type=("build", "run"), when="@2020: ^python@:3.6")
     depends_on("py-msgpack", type=("build", "run"), when="@:2.10.0")
     depends_on("py-msgpack@0.6.0:", type=("build", "run"), when="@2.11.0:")
     depends_on("py-packaging@20.0:", type=("build", "run"), when="@2022.2.1:")
     depends_on("py-psutil@5.0:", type=("build", "run"))
     depends_on("py-six", type=("build", "run"), when="@:1")
     depends_on("py-sortedcontainers@:1,2.0.2:", type=("build", "run"))
+    conflicts_with("py-sortedcontainers@2.0.0,2.0.1", when="@2022.10.2:")
```
adamjstewart: What is this? I've never heard of `conflicts_with`. Also, this is already handled by `depends_on`

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -43,28 +44,39 @@ class PyDistributed(PythonPackage):
...
     depends_on("py-tblib", type=("build", "run"), when="@:2.10.0")
+    depends_on("py-tblib@1.6.0:", type=("build", "run"), when="@2022.10.2:")
```
adamjstewart: Duplicate

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -43,28 +44,39 @@ class PyDistributed(PythonPackage):
...
+    depends_on("py-toolz@0.10.0:", type=("build", "run"), when="@2022.10.2:")
     depends_on("py-tornado@5:", type=("build", "run"), when="^python@:3.7")
     depends_on("py-tornado@6.0.3:", type=("build", "run"), when="^python@3.8:")
+    depends_on("py-tornado@6.0.3:6.1", type=("build", "run"), when="@2022.10.2:")
     depends_on("py-zict@0.1.3:", type=("build", "run"))
     depends_on("py-pyyaml", type=("build", "run"))
     depends_on("py-futures", when="@:1 ^python@2.7:2.8", type=("build", "run"))
     depends_on("py-singledispatch", when="@:1 ^python@2.7:2.8", type=("build", "run"))
+    depends_on("py-urllib3", type=("build", "run"), when="@2022.10.2:")
+    depends_on("py-zict@0.1.3:", type=("build", "run"), when="@2022.10.2:")
```
adamjstewart: Duplicate

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -43,28 +44,40 @@ class PyDistributed(PythonPackage):
...
     depends_on("py-jinja2", type=("build", "run"), when="@2022.2.1:")
+    depends_on("py-locket@1:", type=("build", "run"), when="@2022.2.1:")
     depends_on("py-contextvars", type=("build", "run"), when="@2020: ^python@:3.6")
     depends_on("py-msgpack", type=("build", "run"), when="@:2.10.0")
     depends_on("py-msgpack@0.6.0:", type=("build", "run"), when="@2.11.0:")
     depends_on("py-packaging@20.0:", type=("build", "run"), when="@2022.2.1:")
     depends_on("py-psutil@5.0:", type=("build", "run"))
     depends_on("py-six", type=("build", "run"), when="@:1")
     depends_on("py-sortedcontainers@:1,2.0.2:", type=("build", "run"))
+    conflicts("py-sortedcontainers@2.0.0,2.0.1", when="@2022.10.2:")
```
adamjstewart: This isn't correct (I believe it needs to be `^py-sortedcontainers`) and it's already handled by the `depends_on` line above it

var/spack/repos/builtin/packages/py-distributed/package.py
```diff
@@ -43,28 +44,40 @@ class PyDistributed(PythonPackage):
...
     depends_on("py-tornado@5:", type=("build", "run"), when="^python@:3.7")
     depends_on("py-tornado@6.0.3:", type=("build", "run"), when="^python@3.8:")
+    depends_on("py-tornado@6.0.3:6.1", type=("build", "run"), when="@2022.10.2:")
     depends_on("py-zict@0.1.3:", type=("build", "run"))
     depends_on("py-pyyaml", type=("build", "run"))
     depends_on("py-futures", when="@:1 ^python@2.7:2.8", type=("build", "run"))
     depends_on("py-singledispatch", when="@:1 ^python@2.7:2.8", type=("build", "run"))
+    depends_on("py-urllib3", type=("build", "run"), when="@2022.10.2:")
+    depends_on("py-zict@0.1.3:", type=("build", "run"), when="@2022.10.2:")
+    depends_on("py-pyyaml", type=("build", "run"), when="@2022.10.2:")
```
adamjstewart: Duplicate

var/spack/repos/builtin/packages/py-toolz/package.py
```diff
@@ -12,6 +12,9 @@ class PyToolz(PythonPackage):
     homepage = "https://github.com/pytoolz/toolz/"
     pypi = "toolz/toolz-0.9.0.tar.gz"
 
+    version("0.12.0", sha256="88c570861c440ee3f2f6037c4654613228ff40c93a6c25e0eba70d17282c6194")
     version("0.9.0", sha256="929f0a7ea7f61c178bd951bdae93920515d3fbdbafc8e6caf82d752b9b3b31c9")
 
     depends_on("py-setuptools", type="build")
+    depends_on("python", type=("build", "run"))
```
adamjstewart: This already comes in from the base class

var/spack/repos/builtin/packages/py-toolz/package.py
```diff
@@ -12,6 +12,9 @@ class PyToolz(PythonPackage):
...
+    depends_on("python@3.5:", type=("build", "run"), when="@0.11.0:")
```
adamjstewart, line 19: You can add this if you want, but it's technically not needed since Spack no longer supports Python 3.6 and older

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-dask/package.py
charmoniumQ: Do you mean they should have a trailing colon?

adamjstewart: They already have a trailing colon, but they shouldn't.

var/spack/repos/builtin/packages/py-locket/package.py
```diff
@@ -12,7 +12,10 @@ class PyLocket(PythonPackage):
     homepage = "https://github.com/mwilliamson/locket.py"
     pypi = "locket/locket-0.2.0.tar.gz"
 
+    version("1.0.0", sha256="5c0d4c052a8bbbf750e056a8e65ccd309086f4f0f18a2eac306a8dfa4112a632")
     version("0.2.0", sha256="1fee63c1153db602b50154684f5725564e63a0f6d09366a1cb13dffcec179fb4")
 
     # pip silently replaces distutils with setuptools
     depends_on("py-setuptools", type="build")
+    depends_on("python@2.7:", type="test", when="@1:")
+    conflicts("python@3:3.3", when="@1:")
```
adamjstewart: You could say:
```python
depends_on("python@2.7:2,3.4:", type=("build", "run"))
```
or you could just remove both of these lines since Spack no longer supports Python 3.6 and older anyway

var/spack/repos/builtin/packages/py-dask/package.py
charmoniumQ: If I remember, the `@a:b` is inclusive of `a` and `b`, so the `when` constraint should be `a:(b-1)` and `b:(c-1)`?

adamjstewart: Correct

var/spack/repos/builtin/packages/py-dask/package.py
```diff
@@ -149,13 +156,16 @@ class PyDask(PythonPackage):
     depends_on("py-distributed@1.22:2021.8.0", type=("build", "run"), when="@0.18.0: +distributed")
     depends_on("py-distributed@2.0:2021.8.0", type=("build", "run"), when="@2.0.0: +distributed")
```
adamjstewart: You'll also need to add upper bounds to all `when` args in older distributed deps, otherwise these constraints will also apply to the latest version.

charmoniumQ: How can I test that I did this correctly?

adamjstewart: You can run `spack spec py-dask@X.Y.Z+distributed` for various versions and make sure it concretizes correctly. I'll also spot check it when I'm back at my computer.

var/spack/repos/builtin/packages/py-dask/package.py
```diff
@@ -47,15 +48,18 @@ class PyDask(PythonPackage):
...
     depends_on("python@3.7:", type=("build", "run"), when="@2021.3.1:")
+    depends_on("python@3.8:", type=("build", "run"), when="@2022.10.2:")
 
     depends_on("py-setuptools", type="build")
 
     # Common requirements
     depends_on("py-pyyaml", type=("build", "run"), when="@2.17.1:")
+    depends_on("py-pyyaml@5.3.1:", type=("build", "run"), when="@2022.10.2:")
     depends_on("py-cloudpickle@1.1.1:", type=("build", "run"), when="@2021.3.1:")
     depends_on("py-fsspec@0.6.0:", type=("build", "run"), when="@2021.3.1:")
```
adamjstewart, line 84: Missing dep:
```python
    "packaging >= 20.0",
```

var/spack/repos/builtin/packages/py-dask/package.py
```diff
@@ -138,24 +145,37 @@ class PyDask(PythonPackage):
...
+    )
+    depends_on(
+        "py-distributed@2020.12.0:2021.8.0",
+        type=("build", "run"),
+        when="@2020.12.0:2021.6.1 +distributed",
+    )
+    depends_on(
+        "py-distributed@2021.6.2", type=("build", "run"), when="@2021.6.2:2022.10.1 +distributed"
+    )
+    depends_on("py-distributed@2022.10.2", type=("build", "run"), when="@2022.10.2: +distributed")
```
adamjstewart:
```suggestion
        "py-distributed@2021.6.2", type=("build", "run"), when="@2021.6.2 +distributed"
    )
    depends_on("py-distributed@2022.10.2", type=("build", "run"), when="@2022.10.2 +distributed")
```
These feel like matching version deps.

adamjstewart (changes requested):
Hopefully last 2 changes.

@skosukhin these packages are quickly becoming unmaintainable. How would you feel about deprecating older versions of dask and distributed and removing them someday?

skosukhin:
@adamjstewart please, feel free to deprecate versions that introduce too much overhead

charmoniumQ:
How to mark a version as deprecated? Or should it just be removed completely? What if it is already in someone's `spack.lock` file?

adamjstewart:
See https://spack.readthedocs.io/en/latest/packaging_guide.html#deprecating-old-versions. I'm personally fine with deprecating most versions of dask/distributed, I don't think any package other than py-devito requires those old versions. We can also drop some variants too. This could also be done in a separate PR since this one is almost ready to go (just need to add packaging).

charmoniumQ:
I was thinking of deprecating everything before 2020. Let me know if you wanted more or fewer.

adamjstewart:
I would also deprecate 2.16.0 and 2.10.0 so that the only undeprecated versions are after the versioning scheme change.

adamjstewart (approved):

PR #31810, opened 2022-07-29
py-scipy: add v1.9

adamjstewart:
https://github.com/scipy/scipy/releases/tag/v1.9.0

Successfully builds on macOS 12.4 (Apple M1 Pro) with Python 3.9.13 and Apple Clang 13.1.6.

Scipy 1.9.0 switches from setuptools to meson. **Pros**: build times are down from ~6 min to ~2.5 min! **Cons**: I don't yet see a way to define which BLAS/LAPACK library is used. It seems meson uses pkg-config and picks up whichever BLAS/LAPACK it finds first. It's unclear what the behavior is when multiple BLAS/LAPACK libraries are installed on the system. @rgommers do you know of any way to force scipy to use the same BLAS/LAPACK as numpy? Are there any plans to port numpy to use meson too?

Depends on #31809

PR #33929, opened 2022-11-15
Updating scipy to 1.9.3

rblake-llnl:
New scipy dropped on 2022-10-19.  

We need py-scipy@1.9.3 to compile with py-torch@1.13.0, because py-torch@1.13.0 requires py-pybind11@2.10 and that conflicts with the maximum supported version for py-scipy@:1.8


adamjstewart:
Duplicate of #31810, it's unfortunately far more complicated than this...

PR #33932, opened 2022-11-16
py-isal: use external libisal

snehring:
py-isal also requires the autotools packages and an assembler (I chose nasm) to build.

### DIFF
--- a/var/spack/repos/builtin/packages/py-isal/package.py
+++ b/var/spack/repos/builtin/packages/py-isal/package.py
@@ -19,3 +19,8 @@ class PyIsal(PythonPackage):
 
     depends_on("python@3.7:", type=("build", "run"))
     depends_on("py-setuptools@51:", type="build")
+    depends_on("py-wheel", type="build")
+    depends_on("automake", type="build")
+    depends_on("autoconf", type="build")
+    depends_on("libtool", type="build")
+    depends_on("nasm", type="build")
--- a/var/spack/repos/builtin/packages/py-isal/package.py
+++ b/var/spack/repos/builtin/packages/py-isal/package.py
@@ -19,8 +19,7 @@ class PyIsal(PythonPackage):
 
     depends_on("python@3.7:", type=("build", "run"))
     depends_on("py-setuptools@51:", type="build")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-isal/package.py
```diff
@@ -19,3 +19,8 @@ class PyIsal(PythonPackage):
 
     depends_on("python@3.7:", type=("build", "run"))
     depends_on("py-setuptools@51:", type="build")
+    depends_on("py-wheel", type="build")
```
adamjstewart: Technically wheel is already added by the `PythonPackage` base class so you don't need this line.

var/spack/repos/builtin/packages/py-isal/package.py
```diff
@@ -19,3 +19,8 @@ class PyIsal(PythonPackage):
...
+    depends_on("automake", type="build")
+    depends_on("autoconf", type="build")
+    depends_on("libtool", type="build")
+    depends_on("nasm", type="build")
```
adamjstewart: According to https://python-isal.readthedocs.io/en/stable/index.html#installation-with-pip it's actually possible to use an external installation of isal instead of building it in the Python package. Let's do that instead. So remove these deps and add:
```python
depends_on("libisal")

    # https://python-isal.readthedocs.io/en/stable/index.html#installation-with-pip
    env.set("PYTHON_ISAL_LINK_DYNAMIC", "true")

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-isal/package.py
snehring: That was left in while trying to sort out build errors, whoops.

var/spack/repos/builtin/packages/py-isal/package.py
snehring: Sounds good. I'll give it a test here.

snehring: Looks good.

adamjstewart (approved):

PR #33939, opened 2022-11-16
conditional variant values: allow boolean

becker33:
Most when specs in Spack allow either a spec or a boolean value that reflects system configuration.

@bvanessen pointed out on slack that this is not true for conditional variant values. This PR brings conditional variant values into line with the rest of Spack's `when`-clauses.

### DIFF
--- a/lib/spack/spack/solver/asp.py
+++ b/lib/spack/spack/solver/asp.py
@@ -1026,12 +1026,12 @@ class SpackSolverSetup(object):
                         trigger_id = self.condition(
                             condition_spec,
                             name=pkg.name,
-                            msg="invalid variant value {0}={1}".format(name, value)
+                            msg="invalid variant value {0}={1}".format(name, value),
                         )
                         constraint_id = self.condition(
                             spack.spec.Spec(),
                             name=pkg.name,
-                            msg="empty (total) conflict constraint"
+                            msg="empty (total) conflict constraint",
                         )
                         msg = "variant {0}={1} is conditionally disabled".format(name, value)
                         self.gen.fact(fn.conflict(pkg.name, trigger_id, constraint_id, msg))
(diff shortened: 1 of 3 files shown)


bvanessen (dismissed):

bvanessen:
This works for #33796

lib/spack/spack/solver/asp.py
```diff
@@ -1017,8 +1017,10 @@ def pkg_rules(self, pkg, tests):
                 values = [variant.default]
 
             for value in sorted(values):
-                self.gen.fact(fn.variant_possible_value(pkg.name, name, value))
-                if hasattr(value, "when"):
+                if getattr(value, "when", True) is not True:  # when=True means unconditional
+                    if value.when is False:
+                        continue
+
```
alalazo: Just a suggestion to improve readability, and avoid nested ifs:
```suggestion
                # when=True means unconditional
                value_condition = getattr(value, "when", True)

                if value_condition is False:
                    continue

                if isinstance(value_condition, str):  
```

alalazo (commented):
LGTM, I left just a suggestion to improve readability. Also, if it is easy to write a regression test we should probably add it.

becker33:
@spackbot fix style

bvanessen (approved):
Updates still work for me.

alalazo (approved):

adamjstewart:
Any idea how this could cause https://gitlab.spack.io/spack/spack/-/jobs/4817633? Pipelines are broken on develop. I'm guessing this PR in combination with another PR triggered the issue.

haampie:
@alalazo I think this should be 0.20 not 0.19?

alalazo:
This was a fix for conditional variant values, which are already in the v0.19 release. So maybe not for v0.19.1 but for v0.19.2?

haampie:
Ok, I interpreted it as new functionality, not as a bugfix

PR #33941, opened 2022-11-16
Herwig3: make njet, vbfnlo dependencies optional...

iarspider:
also drop openloops dependency when building on PowerPC. If there is a better way to drop `openloops` dependency only for PowerPC, I will gladly use it.


iarspider:
Building and installing FxFx separately is no longer necessary - at least since 7.2.1 (which is the oldest version here).

vvolkl (dismissed):

vvolkl (approved):

var/spack/repos/builtin/packages/herwig3/package.py
```diff
@@ -54,11 +62,16 @@ def configure_args(self):
...
-            "--with-vbfnlo=" + self.spec["vbfnlo"].prefix,
             "--with-evtgen=" + self.spec["evtgen"].prefix,
+            "--with-gosam=" + self.spec["py-gosam"].prefix,
         ]
 
+        if self.spec.satisfies("+njet"):
+            args.append("--with-njet=" + self.spec["njet"].prefix)
+
+        if self.spec.satisfies("+vbfnlo"):
+            args.append("--with-vbfnlo=" + self.spec["vbfnlo"].prefix)
```
tldahlgren, line 73: FYI.  You could take advantage of the with-or-without helper on these (see https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#activation-value).

iarspider: I had problems in the past with (hep) packages not handling `--without-X` correctly, so I'd prefer not to use it

tldahlgren (approved):

PR #33942, opened 2022-11-16
PackageBase should not define builder legacy attributes

### LINKED ISSUE #33928: Defaults in `package_base` unconditionally deactivate package checks for multiple build systems
### Steps to reproduce

* Attempt to install a package with `--test=root` for a package using any of the build systems:
  * autotools
  * cmake
  * makefile
  * meson
  * nmake
  * python
  * qmake
  * racket
  * scons
  * sip
  * waf
* **IF** the recipe/build system defines checks at build or install time **AND** the relevant `XXX_time_test_callbacks` attribute is mentioned in the `legacy_attributes` list of the build system **AND** the recipe **DOES NOT** override same in (specifically) the Package class...
* **THEN** those checks will not be executed. 
 

### Error message

**N/A**

* **Spack:** 0.19.0
* **Python:** 3.6.8
* **Platform:** linux-scientific7-skylake_avx512
* **Concretizer:** clingo

### General information

greenc-FNAL:
- Fixes #33928.

- Enhance audit function looking for "test" method in callback lists:

  - Accommodate MultiBuild.

  - Check both `build_time_test_callbacks` and
    `install_time_test_callbacks`.
  
- Audit of builder legacy attributes in package classes are
  addressed in https://github.com/spack/spack/pull/33943

### DIFF
--- a/lib/spack/spack/audit.py
+++ b/lib/spack/spack/audit.py
@@ -287,7 +287,7 @@ def _check_build_test_callbacks(pkgs, error_cls):
     errors = []
     for pkg_name in pkgs:
         pkg_cls = spack.repo.path.get_pkg_class(pkg_name)
-        test_callbacks = pkg_cls.build_time_test_callbacks
+        test_callbacks = getattr(pkg_cls, "build_time_test_callbacks", None)
 
         if test_callbacks and "test" in test_callbacks:
             msg = '{0} package contains "test" method in ' "build_time_test_callbacks"
--- a/lib/spack/spack/cmd/info.py
+++ b/lib/spack/spack/cmd/info.py
@@ -241,8 +241,8 @@ def print_tests(pkg):
     # So the presence of a callback in Spack does not necessarily correspond
     # to the actual presence of built-time tests for a package.
     for callbacks, phase in [
-        (pkg.build_time_test_callbacks, "Build"),
-        (pkg.install_time_test_callbacks, "Install"),
+        (getattr(pkg, "build_time_test_callbacks", None), "Build"),
+        (getattr(pkg, "install_time_test_callbacks", None), "Install"),
     ]:
         color.cprint("")
         color.cprint(section_title("Available {0} Phase Test Methods:".format(phase)))
--- a/lib/spack/spack/package_base.py
+++ b/lib/spack/spack/package_base.py
@@ -528,10 +528,6 @@ class PackageBase(WindowsRPath, PackageViewMixin, metaclass=PackageMeta):
     # These are default values for instance variables.
     #
 
-    #: A list or set of build time test functions to be called when tests
-    #: are executed or 'None' if there are no such test functions.
-    build_time_test_callbacks = None  # type: Optional[List[str]]
-
     #: By default, packages are not virtual
     #: Virtual packages override this attribute
     virtual = False
@@ -539,10 +535,6 @@ class PackageBase(WindowsRPath, PackageViewMixin, metaclass=PackageMeta):
     #: Most Spack packages are used to install source or binary code while
     #: those that do not can be used to install a set of other Spack packages.
     has_code = True
-
-    #: A list or set of install time test functions to be called when tests
-    #: are executed or 'None' if there are no such test functions.
-    install_time_test_callbacks = None  # type: Optional[List[str]]
 
     #: By default we build in parallel.  Subclasses can override this.
     parallel = True
--- a/lib/spack/spack/test/builder.py
+++ b/lib/spack/spack/test/builder.py
@@ -121,3 +121,17 @@ def test_old_style_compatibility_with_super(spec_str, method_name, expected):
     builder = spack.builder.create(s.package)
     value = getattr(builder, method_name)()
     assert value == expected
+
+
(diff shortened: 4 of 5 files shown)


lib/spack/spack/audit.py
```diff
@@ -282,18 +282,35 @@ def _search_duplicate_specs_in_externals(error_cls):
 
 
 @package_directives
-def _check_build_test_callbacks(pkgs, error_cls):
-    """Ensure stand-alone test method is not included in build-time callbacks"""
+def _check_test_callbacks(pkgs, error_cls):
+    """Ensure stand-alone test method is not included in test callbacks."""
```
alalazo: Not sure changing the semantics of the test is what we want. I'll have a look at the issue originating the PR soon.

alalazo (changes requested):

lib/spack/spack/audit.py
greenc-FNAL: 1. The audit check as it existed previously was faulty as it _assumed_ that the callbacks were defined in every package class and broke the audit with an exception at the first package for which that was not true. See the originating issue for why solving this problem by defaulting the callbacks in `PackageBase` causes worse problems than it solves.
2. There seemed to be no obvious reason (other than that the audit check predated the advent of `install_time_test_callbacks`) for the presence of the user `test()` method in `build_time_test_callbacks` is bad, but in `install_time_test_callbacks` is not, which is why I extended the audit to include both callbacks.
3. If it is only necessary to exclude the package's `test()` method from legacy  callback lists, but having a builder `test()` method in builder callback lists is OK, I would be happy to adjust this PR accordingly.
4. The auditing question is entirely secondary to the issue described in #33928, for which I believe the correct fix remains the excisions implemented in [ec2f858](https://github.com/spack/spack/pull/33942/commits/ec2f8589c872b81925612b581cd4f9bf1e25aaae)

alalazo: I agree with 4. I cherry picked your commit in #33992 where I also added a regression test.

For 1-3 my understanding is that _there are_ reasons for that. Install time tests are meant to be executed in the prefix, and as such it might make sense to reuse the standalone test method. Build time tests instead are executed in the stage directory, so the standalone test method should not be called there. @tldahlgren can correct me if my understanding is wrong (I am not using the standalone tests a lot, so may be missing details).

greenc-FNAL: @alalazo Thanks for this. I understand your explanation of why install-time tests may use the standalone `test()` method, but I'm missing why there's no need to audit the user builder's `build_time_test_callbacks` in addition to the legacy attribute?

greenc-FNAL: https://github.com/spack/spack/pull/33942/commits/b7f788951c1770700a8c8dbc069f7f4b47787ac3 removes the audit of `install_time_test_callbacks` while preserving the audit of builder attributes in addition to the legacy attribute. I'd also be happy to bring in https://github.com/spack/spack/pull/33992/commits/bd80adb15b9019c1bce74658e3e2eea9dd16ba27 if you agree— I initially didn't understand exactly how one would test the regression efficiently, and was intending to discuss same as part of the PR.

Going forward, please know that I am always happy to revise PRs based on reviewer or other comments—even those suggesting minor or cosmetic changes—and take on board knowledge of why things are/aren't/should be/shouldn't be done. As a core developer, I'm sure you have more pressing demands on your time and expertise than re-implementing fixes to minor bugs.

alalazo: > As a core developer, I'm sure you have more pressing demands on your time and expertise than re-implementing fixes to minor bugs.

Apologies @chissg For this bug I cherry-picked the commit and submitted a new PR since the bug has a high impact on the builtin repository, so I could have a fast turnaround. In that sense it's not minor. 

Also this PR was missing a regression test and is also dealing with audits, which are a separate issue. If you want to remove any change to audit and cherry-pick https://github.com/spack/spack/commit/bd80adb15b9019c1bce74658e3e2eea9dd16ba27 here I am absolutely fine with that - the objective is to fix Spack, not to get _my_ PR in.

alalazo: > why there's no need to audit the user builder's build_time_test_callbacks in addition to the legacy attribute?

As far as I understand, the standalone tests are based on the `PackageBase` hierarchy and they do not need builders (they assume something is already installed). Therefore any `test` method in the `Builder` will not interfere with `spack test`.

greenc-FNAL: Fair enough, thank you for the explanation.

alalazo:
There's also https://github.com/spack/spack/pull/33992/commits/44bba6618bdaa0bc1bc1dd28229f881d68f49e22 needed to pass audits

greenc-FNAL:
@alalazo I have excised audit-related commits from this PR and cherry-picked your regression test and `audit.py` from #33992 as you suggested.

Given that you have explained the differences between my suggested audit fixes and the one you implemented, I would suggest that there is no need for a split PR.

Thanks again for the explanations and your understanding.

alalazo (dismissed):

alalazo (approved):

PR #33943, opened 2022-11-16
Add audit check for builder legacy attributes (with test)

greenc-FNAL:


greenc-FNAL:
As expected, audits flag _lots_ of packages for legacy attribute issues. Should I retrofit them all as part of this PR, or just have maintainers migrate them as PRs come in for affected recipes?

lib/spack/spack/audit.py
```diff
@@ -297,6 +297,30 @@ def _check_build_test_callbacks(pkgs, error_cls):
     return errors
 
 
+@package_directives
+def _check_legacy_attributes(pkgs, error_cls):
+    """Ensure legacy attributes are not defined in package classes."""
```
alalazo, line 302: This is wrong, we marked legacy attributes those that _can_ be used in package classes, if people code the package as a single class. I'll have a look at your issue going to the end of the week.

alalazo (changes requested):

lib/spack/spack/audit.py
greenc-FNAL: Fair enough. Withdrawing PR.

PR #33945, opened 2022-11-16
Clean unit-test workflow file

alalazo:
Delete statements related to Python 2.7, and avoid installing patchelf since now we can bootstrap it.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -49,20 +49,11 @@ jobs:
           # Needed for unit tests
           sudo apt-get -y install \
               coreutils cvs gfortran graphviz gnupg2 mercurial ninja-build \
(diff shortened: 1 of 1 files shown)


trws:
I'm delighted to see the python 2.7 lines go, but while I realize we *can* bootstrap patchelf, how much slower is it?

alalazo:
Next to zero, since we bootstrap from binaries. For instance, in this PR I have seen:
- https://github.com/spack/spack/actions/runs/3481419015/jobs/5822555432 (42 secs. with patchelf)

in other unrelated runs:
- https://github.com/spack/spack/actions/runs/3482320445/jobs/5824575905 (44 secs. without patchelf)

trws (approved):
Awesome, works for me. I asked because the python deps for the prechecks were ~3-6 minutes, which for those was really painful, but order <1 for this I'm happy.

Issue #33696, opened 2022-11-03
Installation issue: ncurses


ax3l:
### Steps to reproduce the issue

```console
$ spack spec -I ncurses
Input spec
--------------------------------
 -   ncurses

Concretized
--------------------------------
[+]  ncurses@6.3%gcc@10.3.0~symlinks+termlib abi=none arch=linux-ubuntu20.04-skylake
[+]      ^pkgconf@1.8.0%gcc@10.3.0 arch=linux-ubuntu20.04-skylake
```
```
$ spack install ncurses
```

### Error message

<details><summary>Error message</summary><pre>
==> Error: Terminating after first install failure: AttributeError: 'str' object has no attribute 'propagate'

The 'ncurses' package cannot find an attribute while trying to build from sources. This might be due to a change in Spack's package format to support multiple build-systems for a single package. You can fix this by updating the build recipe, and you can also report the issue as a bug. More information at https://spack.readthedocs.io/en/latest/packaging_guide.html#installation-procedure
</pre></details>

### Information on your system

* **Spack:** 0.19.0.dev0 (68a5fe84a7f814af7f1c54f9701f33c5f4b2cd17)
* **Python:** 3.8.10
* **Platform:** linux-ubuntu20.04 on Azure Pipelines
* **Concretizer:** clingo

### Additional information

non-interactive execution

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have run `spack maintainers <name-of-the-package>` and **@mentioned** any maintainers
- [X] I have uploaded the build log and environment files
- [X] I have searched the issues of this repo and believe this is not a duplicate

bernhardkaindl:
I cannot reproduce with ncurses, but reproduce a similar error with `spack -d install -v berkeleygw:
* **Spack:** 0.19.0.dev0 (0f0f5332b66562f182e9b24ea9f0fb650b7f3332)
* **Spack:** 0.20.0.dev0 (2f057d729da61e8c12828df44377f0a93fed820f)
* **Python:** 3.8.10
* **Platform:** linux-ubuntu20.04-zen2
* **Platform:** linux-ubuntu20.04-neoverse_n1
* **Concretizer:** clingo
After the install phase:
```
[...]

Issue #33, opened 2015-04-21
scipy fails to build on RHEL6


certik:
```
==> Trying to fetch from https://pypi.python.org/packages/source/s/scipy/scipy-0.15.0.tar.gz
######################################################################## 100.0%
==> Staging archive: /home/certik/repos/spack/var/spack/stage/py-scipy@0.15.0%gcc@4.9.2=unknown_arch-ad579252/scipy-0.15.0.tar.gz
==> Created stage in /home/certik/repos/spack/var/spack/stage/py-scipy@0.15.0%gcc@4.9.2=unknown_arch-ad579252.
==> No patches needed for py-scipy.
==> Building py-scipy.
/home/certik/repos/spack/opt/unknown_arch/gcc@4.9.2/python@2.7.8-3cadb17a/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
/home/certik/repos/spack/opt/unknown_arch/gcc@4.9.2/python@2.7.8-3cadb17a/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'setup_requires'
  warnings.warn(msg)
/home/certik/repos/spack/opt/unknown_arch/gcc@4.9.2/python@2.7.8-3cadb17a/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'test_suite'
  warnings.warn(msg)
lapack_opt_info:
openblas_lapack_info:
  libraries openblas not found in ['/home/certik/repos/spack/opt/unknown_arch/gcc@4.9.2/python@2.7.8-3cadb17a/lib', '/usr/local/lib64', '/usr/local/lib', '/usr/lib64', '/usr/lib', '/usr/lib/']
  NOT AVAILABLE

lapack_mkl_info:
mkl_info:
  libraries mkl,vml,guide not found in ['/home/certik/repos/spack/opt/unknown_arch/gcc@4.9.2/python@2.7.8-3cadb17a/lib', '/usr/local/lib64', '/usr/local/lib', '/usr/lib64', '/usr/lib', '/usr/lib/']
  NOT AVAILABLE

  NOT AVAILABLE

atlas_threads_info:
Setting PTATLAS=ATLAS
  libraries ptf77blas,ptcblas,atlas not found in /home/certik/repos/spack/opt/unknown_arch/gcc@4.9.2/python@2.7.8-3cadb17a/lib
[...]

PR #33948, opened 2022-11-16
fix berkelygw without-mpi variant

glaweh:
closes #34061

The package.py assumed "+mpi" in many places, without checking for the variant. This problem went undetected, as a hard dependency on scalapack pulled an mpi implementation into the dependency chain (this is also fixed).


migueldiascosta:
thanks @glaweh, you're absolutely right

better let https://github.com/spack/spack/pull/33898 go in first, and maybe we can then also address the python deps (https://github.com/spack/spack/pull/33898#pullrequestreview-1180065850)

bernhardkaindl (dismissed):
@migueldiascosta: I think dropping python EOL versions in #33898 may need more time

AFAICS, this PR has nothing to do with python and it is usually better to not lump many different things into one PR.

Also, I do not want to burden @glaweh with having to wait an monitor the python EOL PR and then burden him to add this as well. It's more focussed and light-weight to merge this now, get this fix out and then handle python in a separate PR - This PR already has many lines of diff.

migueldiascosta:
@bernhardkaindl ok. but are you guys able to actually install berkeleygw with the current `develop` branch? I get  `Error: AttributeError: 'str' object has no attribute 'propagate'` at the end of the `install` step

bernhardkaindl:
> @bernhardkaindl ok. but are you guys able to actually install berkeleygw with the current `develop` branch? I get `Error: AttributeError: 'str' object has no attribute 'propagate'` at the end of the `install` step

Thanks for testing! Yes, I did, and I reproduced this as well. 

But I found that this PR is not the cause of it:

Thru the github search, I found that this error was reported with `ncurses` 16 days ago,
in issue #33696.

I also found that I reproduced the same error also without this PR.

I bisected where the Error occurred first: It appears since https://github.com/spack/spack/pull/29761 (Enable/Disable Propagation for Compiler Flags & Variants) was merged.

I came up with a simple check and fallback for when this attribute is missing and found that this fixes the AttributeError: #34019

@migueldiascosta Because you can review #34019, would you review (maybe test) and approve it?

Otherwise, would you add a comment to #34019 that this fixes the AttributeError? Thanks!

Since the `AttributeError: 'str' object has no attribute 'propagate'` with #33 is an independent issue, I think we can apply this update as unrelated to the AttributeError.

bernhardkaindl:
@glaweh @migueldiascosta `spack install berkeleygw~scalapack` results in an error, scalapack is needed for MPI:
```py
../Common/primme_interface.f90:3:4: error: #error ScaLAPACK is required for MPI builds.
```
We need to add this to make them to be selected and deselected together as a unit by the solver:
```py
+    # scalapack and mpi variants need each other, make the solver (de)select the other too:
     conflicts(
         "+scalapack", when="~mpi", msg="scalapack is a parallel library and needs MPI support"
     )
+    conflicts(
+        "~scalapack", when="+mpi", msg="ScaLAPACK is required for MPI builds of berkelygw"
+    )


```

bernhardkaindl:
I could not push (deny updating a hidden ref) to @glaweh's branch, possibly due not giving write permission when opening the PR. **Opened #34061 to review and merge the needed update to develop and the conflicts for the solver.**

bernhardkaindl:
migueldiascosta was right, there is now another conflict with develop due to #33898 being merged.

I've pushed the rebase to the new post-python2 develop branch in #34061 (and alalazo requested a variant change)

bernhardkaindl:
@migueldiascosta and @glaweh You may want to review #34061, after review from alalazo, the new updated title is:

**berkelygw: Use mpi variant to select parallel or serial support**

alalazo:
Oh, I didn't realize #34061 was a re-pushed PR. Maybe @glaweh wants to bring @bernhardkaindl changes into this PR ?

alalazo (approved):

alalazo:
Thanks @glaweh @bernhardkaindl !

PR #33949, opened 2022-11-16
hiredis: updated package definition to use CMake

lpottier:
Hiredis was using Make prior to 1.0.0 and then moved to CMake. I have updated the package to use Cmake and added variants for SSL.


haampie:
Can you please use

https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems

There is not need to deprecate it over a build system change, and since `ccache` uses hiredis and should be easy to bootstrap, it would be a shame if it required building cmake and its dependencies first.

haampie (changes requested):
Can you please use separate classes for the 2 build systems?

https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems

Deprecation shouldn't happen over build system changes, just over (sever) bugs that won't be fixed.

var/spack/repos/builtin/packages/hiredis/package.py
```diff
@@ -6,16 +6,53 @@
...
+        conditional("makefile", when="@:0.14.1"),
+        default="cmake",
+    )
+
+    variant("ssl", default=False, description="Builds with SSL enabled")
+    variant("test", default=False, description="Builds test suite")
+    variant("test_ssl", default=False, description="Builds  test suite for SSL")
+    variant("test_async", default=False, description="Builds test suite for async primitives")
+
+    depends_on("cmake@3.18:", type="build", when="@1.0.0:")
```
haampie:
```suggestion
    depends_on("cmake@3.18:", type="build", when="build_system=cmake")
```

var/spack/repos/builtin/packages/hiredis/package.py
```diff
@@ -6,16 +6,53 @@
...
+    depends_on("openssl@1.1:", type=("build", "link"), when="+ssl")
+    depends_on("openssl@1.1:", type=("build", "link"), when="+test_ssl")
 
     def install(self, spec, prefix):
         make("PREFIX={0}".format(prefix), "install")
```
haampie, line 54: Can you move the old Makefile related methods into

```python
class MakefileBuilder(spack.build_systems.makefile.MakefileBuilder):
    def install(...):
```

and

```python
class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
    def cmake_args(...):
```

haampie (changes requested):

var/spack/repos/builtin/packages/hiredis/package.py
lpottier: Yes, thanks for the tips. I just did it and I also removed the deprecated flags!

var/spack/repos/builtin/packages/hiredis/package.py
```diff
@@ -6,21 +6,16 @@
...
+    git = "https://github.com/redis/hiredis.git"
 
-    maintainers = ["rblake-llnl"]
-
-    build_system(
-        conditional("cmake", when="@1:"),
-        conditional("makefile", when="@0"),
-        default="cmake",
-    )
+    maintainers = ["lpottier"]
```
rblake-llnl: You can be sole maintainer, or you can add me to help share the load.  I think you're more familiar with the package.

var/spack/repos/builtin/packages/hiredis/package.py
```diff
@@ -29,17 +24,53 @@ class Hiredis(CMakePackage, MakefilePackage):
     version("0.13.3", sha256="717e6fc8dc2819bef522deaca516de9e51b9dfa68fe393b7db5c3b6079196f78")
     version("0.13.2", sha256="b0cf73ebe039fe25ecaaa881acdda8bdc393ed997e049b04fc20865835953694")
 
+    build_system(
```
rblake-llnl, line 27: I think `1:` and `:0` work better, right?  That way if 0.14.2 comes out it can be covered.

var/spack/repos/builtin/packages/hiredis/package.py
```diff
@@ -29,17 +24,53 @@ class Hiredis(CMakePackage, MakefilePackage):
...
     def install(self, pkg, spec, prefix):
         make("PREFIX={0}".format(prefix), "install")
+        if "+test" in self.spec or "+test_async" in self.spec or "+test_ssl" in self.spec:
+            make("PREFIX={0}".format(prefix), "test")
 
-    run_after("install", when="platform=darwin")(darwin_fix)
+    @run_after("install")
+    def darwin_fix(self):
+        if self.spec.satisfies("platform=darwin"):
+            fix_darwin_install_name(self.prefix.lib)
```
rblake-llnl, line 61: Does cmake also need the darwin fix?  If not, let's use your version.

rblake-llnl (commented):
LGTM modulo some minor things.  @lpottier actually both had the same idea at the same time.

var/spack/repos/builtin/packages/hiredis/package.py
haampie: I believe CMake does rpaths better

var/spack/repos/builtin/packages/hiredis/package.py
lpottier: I am good with sharing the load! I will add you.

var/spack/repos/builtin/packages/hiredis/package.py
lpottier: Yes, good point! I have switched to `1:` and `:0`.

bernhardkaindl:
@lpottier Are your updates to apply the requested changes finished?

> Can you please use separate classes for the 2 build systems?
> https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems
> Deprecation shouldn't happen over build system changes, just over (sever) bugs that won't be fixed.

On the 1st glance, I't looks it's not split into separate classes yet.
If you want another review, request it from haampie using the circular icon in his entry in the reviewers list on the top-right edge.

lpottier:
> @lpottier Are your updates to apply the requested changes finished?
> 
> > Can you please use separate classes for the 2 build systems?
> > https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems
> > Deprecation shouldn't happen over build system changes, just over (sever) bugs that won't be fixed.
> 
> On the 1st glance, I't looks it's not split into separate classes yet. If you want another review, request it from haampie using the circular icon in his entry in the reviewers list on the top-right edge.

@bernhardkaindl I did split it into two classes (following [Spack multiple-build-systems](https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems)) . Unless, I am missing something here?

Otherwise, my updates are finished (unless the splitting into two classes is not correct per Spack guidelines)

haampie (approved):

bernhardkaindl:
I only had a first glance, so thanks for confirming that your changes are finished.

PR #33951, opened 2022-11-16
Use "vendoring" to maintain Spack's pure Python dependencies

alalazo:
depends on #34029

In this PR I am using [vendoring](https://github.com/pradyunsg/vendoring) to manage our vendored dependencies. The important modifications are in the first commit. The other two are just deleting the old dependencies and adding back the new ones.

To regenerate the tree of dependencies, from the root of our repo:
```console
$ vendoring sync -v
Working in .
Load configuration... 
  Will attempt to load pyproject.toml.
  Read configuration file.
  Parsed configuration file.
  Validated configuration.
  Done!
Clean existing libraries... 
  Done!
Add vendored libraries... 
  Running pip install -t lib/spack/external/_vendoring -r lib/spack/external/vendor.txt --no-compile --no-deps
    Collecting distro==1.8.0
      Using cached distro-1.8.0-py3-none-any.whl (20 kB)
    Collecting jsonschema==3.0.2
      Using cached jsonschema-3.0.2-py2.py3-none-any.whl (54 kB)
[ ... ]
```

The list of our vendored dependencies is maintained in the `lib/spack/external/vendor.txt` file. Licenses etc. are copied automatically by the `vendoring` tool in the folder created for the package.


alalazo:
Currently I am not vendoring `pytest` and erroring out with a silly error message. Before marking this PR as "Ready for review" I'll either add it back or find a way to bootstrap it as we do for the tools used in `spack style`.

EDIT: #34029 adds capabilities to bootstrap pytest

alalazo:
@spackbot fix style

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

lib/spack/docs/developer_guide.rst
```diff
@@ -232,7 +232,7 @@ Spack Subcommands
 Unit tests
 ^^^^^^^^^^
 
-:mod:`spack.test`
+``spack.test``
```
tgamblin, line 235: why not a module reference?

pyproject.toml
```diff
@@ -192,3 +190,43 @@ ignore_errors = true
...
+  # interpreter and OS specific msgpack libs
+  "msgpack/*.so",
+  # unneeded parts of setuptools
+  "easy_install.py",
+  "setuptools",
+  "pkg_resources/_vendor/",
+  "pkg_resources/extern/",
+  # trim vendored pygments styles and lexers
+  "pygments/styles/[!_]*.py",
+  '^pygments/lexers/(?!python|__init__|_mapping).*\.py$',
```
tgamblin, line 217: Random comment: I have thought a few times that it would be nice to vendor pygments so we could colorize YAML and JSON output.

tgamblin (approved):
This LGTM as-is.  A couple comments but honestly this is great.  Thanks to @pradyunsg for `vendoring`.

My only worry is this, from [here](https://github.com/pradyunsg/vendoring):

> This tool has no stability promises -- it has only one intended user: pip. There may be unannounced changes to this codebase at any time, as long as the intended user (i.e. the pip project) is prepared for those changes.

But we deal with more rapidly changing dependencies than this anyway... and if `vendoring` disappears it won't break `spack`... the externals are already copied in and committed here.

pradyunsg:
You can have my word that I’ll bump the major version when I make a backwards incompatible change. Pin with a ~=1.2 or whatever variant you prefer. :)

pyproject.toml
pradyunsg: The only reason pip trims this is for distribution sizes.

PR #33953, opened 2022-11-16
ci: restore ability to reproduce gitlab job failures

scottwittenburg:
Now that the pipeline-generated `install.sh` script contains the absolute path to the directory containing the spack environment concretized during pipeline generation, we need to copy the `spack.lock` and `spack.yaml` artifacts when setting up the reproducer, so they can be found in the same absolute path.

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -1777,9 +1777,9 @@ def reproduce_ci_job(url, work_dir):
     download_and_extract_artifacts(url, work_dir)
 
     lock_file = fs.find(work_dir, "spack.lock")[0]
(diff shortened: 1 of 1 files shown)


kwryankrattiger (approved):
Tested locally on a job I was working on, fixed for me too!

haampie:
Thanks. I knew this was an issue but didn't get to it.

I was contemplating to also add `-e <path>` to the direct `spack install` step in `install.sh`, then reproducing is just two steps (source setup-env.sh and run install.sh) instead of three (spack env activate ...)

kwryankrattiger (approved):

PR #33955, opened 2022-11-16
Python: fix bug detection, trying to access self

alalazo:
Typo introduced in #33847

We cannot reference `self` in a classmethod.


adamjstewart (approved):
I was just about to submit a PR for this, thanks!

Any idea why flake8 doesn't catch this?

alalazo:
Not really. I bet it would have been caught by pylint, but we are not ready for that yet.

alalazo:
The error in the pipeline is unrelated. This only affects detecting Python. Force merging instead of waiting for another pipelines run.

PR #33956, opened 2022-11-16
Python: do not set PYTHONHOME during build

alalazo:
Setting `PYTHONHOME` is rarely needed (since each interpreter has various ways of setting it automatically) and very often it is difficult to get right manually. 

For instance, the change done to set `PYTHONHOME` to `sysconfig["base"]` broke bootstrapping Spack's dev dependencies for me when working inside a virtual environment in Linux. This PR is to just try removing it and check if anything breaks in pipelines.


adamjstewart (approved):
Can't remember why we added this in the first place, but if removing it causes issues for people, we can always revert this PR.

alalazo:
@spackbot run pipeline

PR #33958, opened 2022-11-16
Fix minor issues with Waf build system MultiBuild

greenc-FNAL:
[ Spotted while investigating #33928 ]

* Remove unnecessary duplication in `legacy_methods`

* Fix erroneous duplication of `build_time_test_callbacks` in
  `legacy_attributes`: one of the duplicates should be
  `install_time_test_callbacks`

* Remove now-unnecessary directives around `install_test()` in
  `py-py2cairo` (requires fix to #33928)

### DIFF
--- a/lib/spack/spack/build_systems/waf.py
+++ b/lib/spack/spack/build_systems/waf.py
@@ -73,8 +73,8 @@ class WafBuilder(BaseBuilder):
     #: Names associated with package attributes in the old build-system format
     legacy_attributes = (
         "build_time_test_callbacks",
-        "build_time_test_callbacks",
         "build_directory",
+        "install_time_test_callbacks",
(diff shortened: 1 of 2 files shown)


lib/spack/spack/build_systems/waf.py
```diff
@@ -72,9 +72,9 @@ class WafBuilder(BaseBuilder):
 
     #: Names associated with package attributes in the old build-system format
     legacy_attributes = (
-        "build_time_test_callbacks",
         "build_time_test_callbacks",
         "build_directory",
+        "install_time_test_callbacks",
```
alalazo, line 77: Thanks for spotting this typo!

alalazo (approved):

PR #33959, opened 2022-11-16
py-cutadapt: adding version 4.1

snehring:
Also updating a dep to address some build issues I've started to see.

### DIFF
--- a/var/spack/repos/builtin/packages/py-cutadapt/package.py
+++ b/var/spack/repos/builtin/packages/py-cutadapt/package.py
@@ -22,16 +22,13 @@ class PyCutadapt(PythonPackage):
     version("1.13", sha256="aa9f2c1f33dc081fe94f42b1250e4382b8fb42cabbf6e70a76ff079f211d5fc0")
 
     # universal deps
-    depends_on("py-setuptools", type=("build", "run"))
-    depends_on("py-setuptools-scm", type="build", when="@2.0:")
-    # 4.1 deps
+    depends_on("py-setuptools@43:", type=("build", "run"))
+    depends_on("py-setuptools-scm@6.2:", type="build", when="@2.0:")
     depends_on("python@3.7:", type=("build", "run"), when="@4.1:")
+    # 4.1 deps
     depends_on("py-dnaio@0.7.1:", type=("build", "run"), when="@4.1:")
     depends_on("py-xopen@1.1:", type=("build", "run"), when="@4.1:")
     # older version deps
-    depends_on("python@2.6:", type=("build", "run"), when="@1.13")
-    depends_on("python@3.4:", type=("build", "run"), when="@2.0:2.5")
-    depends_on("python@3.5:", type=("build", "run"), when="@2.6:2.10")
     depends_on("py-xopen@0.1.1:", type=("build", "run"), when="@1.13")
     depends_on("py-xopen@0.5.0:", type=("build", "run"), when="@2.0:2.3")
     depends_on("py-xopen@0.7.3:", type=("build", "run"), when="@2.4")
--- a/var/spack/repos/builtin/packages/py-cutadapt/package.py
+++ b/var/spack/repos/builtin/packages/py-cutadapt/package.py
@@ -15,23 +15,30 @@ class PyCutadapt(PythonPackage):
     pypi = "cutadapt/cutadapt-1.13.tar.gz"
     git = "https://github.com/marcelm/cutadapt.git"
 
+    version("4.1", sha256="be745ff24adfb4a3eaf715dfad0e2ccdfad7792ef00c1122adf4fbf3aed9227b")
     version("2.10", sha256="936b88374b5b393a954852a0fe317a85b798dd4faf5ec52cf3ef4f3c062c242a")
     version("2.9", sha256="cad8875b461ca09cea498b4f0e78b0d3dcd7ea84d27d51dac4ed45080bf1499e")
     version("2.5", sha256="ced79e49b93e922e579d0bb9d21298dcb2d7b7b1ea721feed484277e08b1660b")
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/py-cutadapt/package.py
```diff
@@ -15,23 +15,30 @@ class PyCutadapt(PythonPackage):
...
     version("2.9", sha256="cad8875b461ca09cea498b4f0e78b0d3dcd7ea84d27d51dac4ed45080bf1499e")
     version("2.5", sha256="ced79e49b93e922e579d0bb9d21298dcb2d7b7b1ea721feed484277e08b1660b")
     version("1.13", sha256="aa9f2c1f33dc081fe94f42b1250e4382b8fb42cabbf6e70a76ff079f211d5fc0")
 
-    depends_on("python@2.6:", type=("build", "run"), when="@1.13")
-    depends_on("python@3.4:", type=("build", "run"), when="@2.0:2.5")
-    depends_on("python@3.5:", type=("build", "run"), when="@2.6:")
+    # universal deps
     depends_on("py-setuptools", type=("build", "run"))
     depends_on("py-setuptools-scm", type="build", when="@2.0:")
```
adamjstewart: Need to specify required versions in latest version:
```toml
requires = ["setuptools >= 43", "wheel", "setuptools_scm >= 6.2", "Cython >= 0.29.20"]
```

var/spack/repos/builtin/packages/py-cutadapt/package.py
```diff
@@ -15,23 +15,30 @@ class PyCutadapt(PythonPackage):
...
+    # 4.1 deps
+    depends_on("python@3.7:", type=("build", "run"), when="@4.1:")
+    depends_on("py-dnaio@0.7.1:", type=("build", "run"), when="@4.1:")
+    depends_on("py-xopen@1.1:", type=("build", "run"), when="@4.1:")
+    # older version deps
+    depends_on("python@2.6:", type=("build", "run"), when="@1.13")
+    depends_on("python@3.4:", type=("build", "run"), when="@2.0:2.5")
+    depends_on("python@3.5:", type=("build", "run"), when="@2.6:2.10")
```
adamjstewart: Spack no longer supports Python 3.6 and older, so you don't technically have to specify these at all, but you can leave them in if you want.

var/spack/repos/builtin/packages/py-dnaio/package.py
```diff
@@ -13,11 +13,17 @@ class PyDnaio(PythonPackage):
     pypi = "dnaio/dnaio-0.3.tar.gz"
     git = "https://github.com/marcelm/dnaio.git"
 
+    version("0.9.1", sha256="a1a14181995b27197b7e2b8897994a3107c649b9fc2dfe263caff3c455b0d0c4")
     version("0.4.2", sha256="fa55a45bfd5d9272409b714158fb3a7de5dceac1034a0af84502c7f503ee84f8")
     version("0.3", sha256="47e4449affad0981978fe986684fc0d9c39736f05a157f6cf80e54dae0a92638")
 
-    depends_on("python@3.4:", type=("build", "run"))
     depends_on("py-setuptools", type=("build", "run"))
     depends_on("py-setuptools-scm", when="@0.4:", type="build")
```
adamjstewart: Same here:
```toml
requires = ["setuptools >= 52", "setuptools_scm >= 6.2", "Cython >= 0.29.20"]
```

var/spack/repos/builtin/packages/py-dnaio/package.py
```diff
@@ -13,11 +13,17 @@ class PyDnaio(PythonPackage):
...
+    # 0.9.1 deps
+    depends_on("python@3.7:", type=("build", "run"), when="@0.9.1:")
+    depends_on("py-cython@0.29.20:", type=("build", "run"), when="@0.9.1:")
+    depends_on("py-xopen@1.4.0:", type=("build", "run"), when="@0.9.1:")
+    # older deps
+    depends_on("python@3.4:", type=("build", "run"))
```
adamjstewart: Same here, not required

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-cutadapt/package.py
snehring: I was actually looking for your feed back for this, I didn't see either of these packages in the big eol pr. Would you like me to just remove the older versions entirely?

var/spack/repos/builtin/packages/py-cutadapt/package.py
snehring: The cython dep doesn't seem to be strictly necessary here (unlike dnaio) would you like me to include it?

var/spack/repos/builtin/packages/py-cutadapt/package.py
snehring: I've not tested, but looking at the setup.py and the like for the old versions it sounds like 3.7 would probably work without issue for everything except possibly 1.13

var/spack/repos/builtin/packages/py-cutadapt/package.py
adamjstewart: I would still add it. If we didn't manually pass in flags to disable it, pip would install cython since the pyproject.toml says to. You can reach out to the devs and ask them if this is really necessary to include.

var/spack/repos/builtin/packages/py-cutadapt/package.py
adamjstewart: Most packages are forwards compatible, so there's no reason to remove the old versions. But there's also no reason to say that certain versions support Python 2.6+ and others support Python 3.4+ when Spack only supports Python 3.7+ anyway.

adamjstewart (approved):

PR #33960, opened 2022-11-17
Environments: Add support for including definitions files

tldahlgren:
This PR adds support for including separate definitions from `spack.yaml`.

Supporting the inclusion of files with definitions enables user to make curated/standardized collections of packages that can re-used by others.

### DIFF
--- a/lib/spack/spack/schema/__init__.py
+++ b/lib/spack/spack/schema/__init__.py
@@ -62,3 +62,25 @@ def _make_validator():
 
 
 Validator = llnl.util.lang.Singleton(_make_validator)
+
+spec_list_schema = {
+    "type": "array",
+    "default": [],
+    "items": {
+        "anyOf": [
+            {
+                "type": "object",
+                "additionalProperties": False,
+                "properties": {
+                    "matrix": {
+                        "type": "array",
+                        "items": {"type": "array", "items": {"type": "string"}},
+                    },
+                    "exclude": {"type": "array", "items": {"type": "string"}},
+                },
+            },
+            {"type": "string"},
+            {"type": "null"},
+        ]
+    },
+}
--- a/lib/spack/spack/schema/env.py
+++ b/lib/spack/spack/schema/env.py
@@ -12,33 +12,10 @@ from llnl.util.lang import union_dicts
 
 import spack.schema.gitlab_ci  # DEPRECATED
 import spack.schema.merged
-import spack.schema.packages
 import spack.schema.projections
 
 #: Top level key in a manifest file
 TOP_LEVEL_KEY = "spack"
-
-spec_list_schema = {
-    "type": "array",
-    "default": [],
-    "items": {
-        "anyOf": [
-            {
-                "type": "object",
-                "additionalProperties": False,
-                "properties": {
-                    "matrix": {
-                        "type": "array",
-                        "items": {"type": "array", "items": {"type": "string"}},
-                    },
-                    "exclude": {"type": "array", "items": {"type": "string"}},
-                },
-            },
-            {"type": "string"},
-            {"type": "null"},
-        ]
-    },
-}
 
 projections_scheme = spack.schema.projections.properties["projections"]
 
@@ -75,16 +52,7 @@ schema = {
                             }
                         },
                     },
-                    "definitions": {
-                        "type": "array",
-                        "default": [],
-                        "items": {
-                            "type": "object",
-                            "properties": {"when": {"type": "string"}},
-                            "patternProperties": {r"^(?!when$)\w*": spec_list_schema},
-                        },
-                    },
-                    "specs": spec_list_schema,
+                    "specs": spack.schema.spec_list_schema,
                     "view": {
                         "anyOf": [
                             {"type": "boolean"},
--- a/lib/spack/spack/test/cmd/env.py
+++ b/lib/spack/spack/test/cmd/env.py
@@ -805,15 +805,18 @@ spack:
     )
 
 
-def test_env_with_included_config_file(tmp_path, environment_from_manifest, packages_file):
+def test_env_with_included_config_file(mutable_mock_env_path, packages_file):
     """Test inclusion of a relative packages configuration file added to an
     existing environment.
     """
+    env_root = mutable_mock_env_path
+    fs.mkdirp(env_root)
     include_filename = "included-config.yaml"
-    included_path = os.path.join(tmp_path, include_filename)
+    included_path = env_root / include_filename
     shutil.move(packages_file.strpath, included_path)
 
-    e = environment_from_manifest(
+    spack_yaml = env_root / ev.manifest_name
+    spack_yaml.write_text(
         f"""\
 spack:
   include:
@@ -823,6 +826,7 @@ spack:
 """
     )
 
+    e = ev.Environment(env_root)
     with e:
         e.concretize()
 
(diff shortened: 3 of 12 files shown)


lib/spack/spack/schema/env.py
```diff
@@ -81,16 +53,8 @@
...
-                        "type": "array",
-                        "default": [],
-                        "items": {
-                            "type": "object",
-                            "properties": {"when": {"type": "string"}},
-                            "patternProperties": {r"^(?!when$)\w*": spec_list_schema},
-                        },
-                    },
-                    "specs": spec_list_schema,
+                    "definitions": spack.schema.definitions_schema,
```
becker33: This should be part of `spack.schema.merged.properties` which is used above.

lib/spack/spack/schema/config.py
```diff
@@ -83,6 +83,7 @@
             "url_fetch_method": {"type": "string", "enum": ["urllib", "curl"]},
             "additional_external_search_paths": {"type": "array", "items": {"type": "string"}},
             "binary_index_ttl": {"type": "integer", "minimum": 0},
+            "definitions": spack.schema.definitions_schema,
```
becker33: This does not belong under config -- it's a separate section in its own right.

lib/spack/spack/schema/__init__.py
```diff
@@ -64,3 +65,40 @@ def _deprecated_properties(validator, deprecated, instance, schema):
...
+
+definitions_schema = {
+    "type": "array",
+    "default": [],
+    "items": {
+        "type": "object",
+        "properties": {"when": {"type": "string"}},
+        "patternProperties": {r"^(?!when$)\w*": spec_list_schema},
+    },
+}  # type: Dict[str, Any]
```
becker33: This belongs in the definitions module

becker33 (dismissed):

tldahlgren:
Ping @becker33

tldahlgren:
Ping @becker33

lib/spack/spack/test/cmd/env.py
```diff
@@ -852,68 +856,67 @@ def test_env_with_included_config_missing_file(tmpdir, mutable_empty_config):
...
+    env_root = mutable_mock_env_path
+    config_scope_path = env_root / "config"
 
     # Copy the packages.yaml file to the environment configuration
     # directory, so it is picked up during concretization. (Using
     # copy instead of rename in case the fixture scope changes.)
     fs.mkdirp(config_scope_path)
     include_filename = os.path.basename(packages_file.strpath)
-    included_path = os.path.join(config_scope_path, include_filename)
+    included_path = config_scope_path / include_filename
```
alecbcs, line 879: I find the `path / path` syntax a little confusing compared to `os.path.join`, since truediv isn't specified for strings. If this is becoming the preferred way of doing things though, I don't have any real objection.

lib/spack/spack/test/cmd/env.py
```diff
@@ -852,68 +856,67 @@ def test_env_with_included_config_missing_file(tmpdir, mutable_empty_config):
...
     config_real_path = substitute_path_variables(config_var_path)
-    fs.mkdirp(os.path.dirname(config_real_path))
-    shutil.move(packages_file.strpath, config_real_path)
+    shutil.move(included_file, config_real_path)
     assert os.path.exists(config_real_path)
 
+    e = ev.Environment(env_path)
     with e:
-        e.concretize()
+        concretize()
```
alecbcs: Is the following,
```
e = ev.Environment(env_path)
with e:
    concretize()
```

more efficient than just calling,

```
e = ev.Environment(env_path)
e.concretize()
```

my understanding of `with` is that it let's you define exit functions which is why we use it to implicitly close a file after writing to it.

lib/spack/spack/test/cmd/env.py
```diff
@@ -924,29 +927,50 @@ def test_env_config_precedence(environment_from_manifest):
...
+  include:
+  - {os.path.join(".", included_file)}
+  specs:
+  - mpileaks
+"""
+    )
+
+    e = ev.Environment(tmp_path)
     with e:
         e.concretize()
```
alecbcs, line 950: We should probably standardize and either not use `with` here or not call it as `e.concretize()` as was fixed above.

lib/spack/spack/test/cmd/env.py
```diff
@@ -966,12 +990,16 @@ def test_included_config_precedence(environment_from_manifest):
 """
         )
 
+    e = ev.Environment(tmp_path)
     with e:
         e.concretize()
```
alecbcs, line 999: Same thing here. Just need to standardize on a single way of calling `concretize()`.

lib/spack/spack/test/env.py
```diff
@@ -690,3 +691,64 @@ def test_removing_spec_from_manifest_with_exact_duplicates(
...
+
+  specs:
+  - matrix:
+    - [$core_specs]
+    - [$my_compilers]
+"""
+    )
+
+    e = ev.Environment(env_path)
+    ev.activate(e)
```
alecbcs: If we decide `with` is more efficient/better practice we be using it here too.

alecbcs (commented):
@tldahlgren looks great! Just a few questions around standardization and syntax conventions.

lib/spack/spack/test/cmd/env.py
tldahlgren: This is a great question to bring up in the core meeting as I've seen it in more PRs "lately".

lib/spack/spack/test/cmd/env.py
tldahlgren: The `with` context manager is more concise than a `try-catch` with guarantees that the exit is called.  In this case, we are guaranteed the environment is deactivated (and any previous re-activated so hopefully there aren't any).  Unit tests can have weird behavior if an environment from another unit test is unexpectedly (still) active.

lib/spack/spack/test/cmd/env.py
tldahlgren: Fair point since I am modifying the test case.

lib/spack/spack/test/cmd/env.py
tldahlgren: Again, since I'm modifying the test case.

lib/spack/spack/test/cmd/env.py
tldahlgren: You can see it's starting to be used more with the transition @alalazo suggested to push for more use of `pathlib` (see https://docs.python.org/3/library/pathlib.html#basic-use).

lib/spack/spack/test/cmd/env.py
tldahlgren: You actually pointed out another issue with this case that I'll resolve.

tldahlgren: That should be:

```
    e = ev.Environment(env_path)
    with e:
        e.concretize()
```

tldahlgren: Don't think we really want to go through the concretize command since there are other tests that already do that.

lib/spack/spack/test/cmd/env.py
tldahlgren: Actually, you need to call it with `e.concretize()` here if you want to use the method in the environment instance; otherwise, you're using the `concretize()` command.  There are other tests that exercise the latter so, from efficiency standpoint, this is better.

lib/spack/spack/test/cmd/env.py
tldahlgren: Same here about not going through the overhead of the concretize command.

lib/spack/spack/test/env.py
tldahlgren: Ooops.  Needed to *force* push to see the changes since also rebased.

tldahlgren: Done.

tldahlgren:
@becker33 @alecbcs All but project coverage passes and I'm not sure why it is reduced given I'm hitting 100% of diffs for patch coverage.

alecbcs (approved):
Looks good to me. I'm not sure I completely understand the use of `with e:` but that can be addressed in a follow up PR if needed.

lib/spack/spack/test/env.py
```diff
@@ -716,3 +717,64 @@ def test_variant_propagation_with_unify_false(tmp_path, mock_packages):
...
+    env_path.mkdir()
+    filename = "missing-def.yaml"
+    defs_file = env_path / filename
+    defs_file.write_text("definitions:\n- my_compilers: ['%gcc']\n")
+
+    spack_yaml = env_path / ev.manifest_name
+    spack_yaml.write_text(
+        f"""spack:
+  include:
+  - file://{defs_file}
```
tgamblin, line 768: The only examples here are using file URLs -- that isn't required, though, right?  Would be good to have a test for a plain path.

tgamblin (approved):

PR #33968, opened 2022-11-17
openblas: Fix build on ARM Neoverse with gcc@:9 (no sve2+bf16)

bernhardkaindl:
Fixes this build failure when building openblas@0.3.21 on a system with an ARM Neoverse CPU from `gcc@:9`:
```yaml
cc1: error: invalid feature modifier 'sve2' in '-march=armv8.5-a+sve+sve2+bf16'
```
- Apply the upstream fix: Do not use `sve2+bf16` when using `gcc@:9` with this version (requires `gcc@10:` or `clang`)

- Also improve the `InstallError` message when using `+fortran` but no `FC` was added.

### DIFF
--- a/var/spack/repos/builtin/packages/openblas/package.py
+++ b/var/spack/repos/builtin/packages/openblas/package.py
@@ -168,6 +168,19 @@ class Openblas(MakefilePackage):
     # See <https://github.com/xianyi/OpenBLAS/issues/3760>
     patch("linktest.patch", when="@0.3.20")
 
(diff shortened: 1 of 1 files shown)


alalazo:
The failure is spurious:
```console
==> Error: Failed to install py-requests due to timeout: The read operation timed out
==> Error: py-requests-2.28.1-x7xypofdxosgvukg5dtwhemat7eclu3u: Package was not installed
==> Error: Installation request failed.  Refer to reported errors for failing package(s).
```
restarting the pipeline.

@spackbot run pipeline

alalazo (approved):

PR #33971, opened 2022-11-17
Update libjpeg-turbo using new multibuildsystem approach

iarspider:


iarspider:
The lack of documentation about this new approach is sad.

bernhardkaindl:
> The lack of documentation about this new approach is sad.

I also have no knowledge about it!

@iarspider: How could I select one or the other build system?

Could you send in your knowledge so it can be added to `lib/spack/docs/build_systems`?

iarspider:
@bernhardkaindl I can try doing it in a separate PR.

iarspider:
@spackbot run pipeline

bernhardkaindl:
> @bernhardkaindl I can try doing it in a separate PR.

Yes, this would be nice of you! As far as I can see now, the diff of this PR actually looks like a perfect example how to work with it! Thanks!

@iarspider: I'm looking forward for your documentation PR!

bernhardkaindl:
I merged develop into this PR because (at least AFAICS) gitlab-ci failed (at least in part) because fetching of `glib` failed and we just updated to download URL for fetching `glib` in develop, so there is hope this might help. (gnome.org is apparently having web site troubles)

bernhardkaindl (approved):
I've built the current version using cmake and an old version with autoconf, LGTM.

alalazo:
> The lack of documentation about this new approach is sad.

Can you be more precise on what is missing?

- https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems
- https://spack.readthedocs.io/en/latest/packaging_guide.html#package-class-architecture
- https://spack-tutorial.readthedocs.io/en/latest/tutorial_packaging.html#multiple-build-systems

iarspider:
@alalazo I must've overlooked the second code block (restricting build systems to version ranges) under https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems - this is what I was looking for ! Sorry!

I think Spack would benefit from a packaging cheat-sheet of sorts: a single page with just "How do I..." recipes (of which half would be "what is the signature of method X") - what do you think?

PR #33973, opened 2022-11-17
spack external find: library detection improvements

haampie:
- Ensure that we use the correct env variable per platform
- Dedupe dir paths (e.g. /usr/lib == /lib)
- On ELF platforms ensure that shared libraries are only added when they are host-compatible (checked against current Python interpreter's own EI_CLASS, EI_DATA and e_machine).

The third point is relevant for multi-arch distros, such as Ubuntu, where sometimes i686 directories have higher priority than x86_64. For executables we typically inspect `exe --version` which also ensures host-compatibility, so to me it makes sense if for libraries we verify "is the lib compatible with the Python interpreter we're currently running".

This doesn't deal with static libraries, nor with Darwin.

### DIFF
--- a/lib/spack/spack/detection/path.py
+++ b/lib/spack/spack/detection/path.py
@@ -15,9 +15,12 @@ import warnings
 from typing import Dict, List, Optional, Set, Tuple
 
 import llnl.util.filesystem
+import llnl.util.lang
 import llnl.util.tty
 
+import spack.util.elf as elf_utils
 import spack.util.environment
+import spack.util.environment as environment
 import spack.util.ld_so_conf
 
 from .common import (
@@ -57,6 +60,11 @@ def common_windows_package_paths(pkg_cls=None) -> List[str]:
     return paths
 
 
+def file_identifier(path):
+    s = os.stat(path)
+    return (s.st_dev, s.st_ino)
+
+
 def executables_in_path(path_hints: List[str]) -> Dict[str, str]:
     """Get the paths of all executables available from the current PATH.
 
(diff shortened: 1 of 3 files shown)


alalazo (approved):

lib/spack/spack/detection/path.py
```diff
@@ -94,17 +130,45 @@ def libraries_in_ld_and_system_library_path(
...
+            search_paths.extend(environment.get_path("LD_LIBRARY_PATH"))
+
+        # Dynamic linker paths
+        search_paths.extend(spack.util.ld_so_conf.host_dynamic_linker_search_paths())
+
+        # Drop redundant paths
+        search_paths = list(filter(os.path.isdir, search_paths))
+
+    # Make use we don't doubly list /usr/lib and /lib etc
+    search_paths = list(llnl.util.lang.dedupe(search_paths, key=file_identifier))
```
alalazo, line 155: Since the file identifier contains the inode, would this dedupe on e.g. Ubuntu:
```console
$ stat /lib /usr/lib
  File: /lib -> usr/lib
  Size: 7         	Blocks: 0          IO Block: 4096   symbolic link
Device: 10303h/66307d	Inode: 14          Links: 1
Access: (0777/lrwxrwxrwx)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2023-11-01 12:05:01.533555114 +0100
Modify: 2021-08-20 16:17:21.544002711 +0200
Change: 2021-08-20 16:17:21.544002711 +0200
 Birth: -
  File: /usr/lib
  Size: 12288     	Blocks: 24         IO Block: 4096   directory
Device: 10303h/66307d	Inode: 25165829    Links: 154
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2023-11-02 09:20:46.064505013 +0100
Modify: 2023-10-28 13:07:40.883302892 +0200
Change: 2023-10-28 13:07:40.883302892 +0200
 Birth: -
```
? Or am I reading this wrong?

haampie: It's been a while since I wrote that, but I guess the idea is that `os.stat` follows the symlink, it's not `lstat`

haampie: Hm, do we actually filter existing paths?

haampie: Ah yes, filter(os.path.isdir) above should do it...

alalazo: Gotcha, `os.stat` maps to `stat -L`, while `os.lstat` maps to `stat` (after reading docs). Weird on the Python side, but correct here.

PR #33982, opened 2022-11-18
Updating hiredis with new software versions

rblake-llnl:
In the process, I've updated the build system and added myself as a reviewer.

I'd love a check on this commit-- this is my first time implementing the new build system.

### DIFF
--- a/var/spack/repos/builtin/packages/hiredis/package.py
+++ b/var/spack/repos/builtin/packages/hiredis/package.py
@@ -30,21 +30,16 @@ class Hiredis(CMakePackage, MakefilePackage):
     version("0.13.2", sha256="b0cf73ebe039fe25ecaaa881acdda8bdc393ed997e049b04fc20865835953694")
 
 
+def darwin_fix(builder):
+    fix_darwin_install_name(builder.pkg.prefix.lib)
+
+
 class MakefileBuilder(spack.build_systems.makefile.MakefileBuilder):
     def install(self, pkg, spec, prefix):
         make("PREFIX={0}".format(prefix), "install")
 
-    @run_after("install")
-    def darwin_fix(self):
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/hiredis/package.py
```diff
@@ -6,21 +6,45 @@
...
     def darwin_fix(self):
         if self.spec.satisfies("platform=darwin"):
-            fix_darwin_install_name(self.prefix.lib)
+            fix_darwin_install_name(self.pkg.prefix.lib)
+
+
+class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
+    def cmake_args(self):
+        return []
+
```
alalazo:
```suggestion
```

var/spack/repos/builtin/packages/hiredis/package.py
```diff
@@ -6,21 +6,45 @@
...
+    @run_after("install")
+    def darwin_fix(self):
+        if self.spec.satisfies("platform=darwin"):
+            fix_darwin_install_name(self.pkg.prefix.lib)
```
alalazo: Here and above you can:
```suggestion
    @run_after("install", when="platform=darwin")
    def darwin_fix(self):
        fix_darwin_install_name(self.pkg.prefix.lib)
```
Also, in principle you can define a:
```python
def darwin_fix(builder):
    fix_darwin_install_name(builder.pkg.prefix.lib)
```
function directly at the module scope, and call it in two places with:
```python
run_after("install", when="platform=darwin")(darwin_fix)
```
Not essential though, so I'll leave it up to you which one you consider better

alalazo (changes requested):
A couple of small fixes, but essentially LGTM.

alalazo (approved):

PR #33983, opened 2022-11-18
Updating faiss with new versions

rblake-llnl:
Updating build system to cmake, set up python bindings.

Also adding myself as a reviewer, along with @bhatiaharsh .

### DIFF
--- a/var/spack/repos/builtin/packages/faiss/package.py
+++ b/var/spack/repos/builtin/packages/faiss/package.py
@@ -8,7 +8,7 @@ import os
 from spack.package import *
 
 
-class Faiss(AutotoolsPackage, CudaPackage):
+class Faiss(AutotoolsPackage, CMakePackage, CudaPackage):
     """Faiss is a library for efficient similarity search and clustering of
      dense vectors.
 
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/faiss/package.py
```diff
@@ -56,12 +63,65 @@ class Faiss(AutotoolsPackage, CudaPackage):
...
+
+        if "+cuda" in spec:
+            key = "CMAKE_CUDA_ARCHITECTURES"
+            args.append(self.define_from_variant(key, "cuda_arch"))
+            # args.append(self.define_from_variant(
+            # 'CMAKE_CUDA_STANDARD', 'cudastd'))
+        return args
+
+    def install(self, pkg, spec, prefix):
+        spack.build_systems.cmake.CMakeBuilder.install(self, pkg, spec, prefix)
```
alalazo: Doesn't this:
```suggestion
        super().install(pkg, spec, prefix)
```
work as well?

var/spack/repos/builtin/packages/faiss/package.py
```diff
@@ -56,12 +63,65 @@ class Faiss(AutotoolsPackage, CudaPackage):
...
+
+class PythonPipBuilder(spack.build_systems.python.PythonPipBuilder):
+    def __init__(self, pkg, build_dirname):
+        spack.build_systems.python.PythonPipBuilder.__init__(self, pkg)
+        self.build_dirname = build_dirname
+
+    @property
+    def build_directory(self):
+        return os.path.join(self.pkg.stage.path, self.build_dirname, "faiss", "python")
+        # return self.pkg.stage.source_path
```
alalazo:
```suggestion
```

var/spack/repos/builtin/packages/faiss/package.py
```diff
@@ -56,12 +63,65 @@ class Faiss(AutotoolsPackage, CudaPackage):
...
+    # a header is missing in a test file
+    #   https://github.com/facebookresearch/faiss/issues/2300
+    patch("fixes-in-v1.7.2.patch", when="@1.7.2")
+
+    def setup_run_environment(self, env):
+        if "+python" in self.spec:
+            env.prepend_path("PYTHONPATH", python_platlib)
+
+
+class PythonPipBuilder(spack.build_systems.python.PythonPipBuilder):
```
alalazo: This one is used as part of another object, which is fine. Maybe it's better scoped as a local class defined in the `CMakeBuilder.install` method?

alalazo (changes requested):
There seem to be build systems related imports missing. Besides the minor comments in the review LGTM.

alalazo (approved):
LGTM, thanks! Let me know if you have any feedback on the multiple build systems interface :slightly_smiling_face:

rblake-llnl:
I had to read the documentation several times to get it, and there was some strangeness with needing to specifically reconcretize when I changed the build systems around. In general it works quite well though!  Way better than if I had to use both methods simultaneously.

haampie:
@rblake-llnl is there no way to make cmake install the python part?

CMake distinguishes between build and install rpaths, if you do an ad-hoc install with pip like this, you install shared libraries with build rpaths instead of install rpaths.

Issue #33985, opened 2022-11-18
Better concretizer error message: Cannot satisfy requirement group for package


adamjstewart:
### Summary

Okay, this one was my own dumb fault. I added a `packages.yaml` along the lines of:
```yaml
packages:
  foo:
    require: "@1.2.3"
```
then later forgot about this and tried to install a newer version:
```console
$ spack install foo@2.3.4
...
==> Error: Cannot satisfy requirement group for package 'foo'
```
I'm wondering if we can somehow improve this message to clarify that:

1. I, the user, expressly forbid this in `packages.yaml`
2. Specifically, `~/.spack/packages.yaml`
3. And if I want to allow this, I need to delete this specific section of `packages.yaml`

### General information

- [X] I have run `spack --version` and reported the version of Spack
- [X] I have searched the issues of this repo and believe this is not a duplicate

PR #33988, opened 2022-11-18
Improve error message for requirements

alalazo:
refers #33985

### DIFF
--- a/lib/spack/spack/solver/concretize.lp
+++ b/lib/spack/spack/solver/concretize.lp
@@ -563,7 +563,7 @@ requirement_weight(Package, W) :-
   requirement_policy(Package, X, "any_of"),
   requirement_group_satisfied(Package, X).
 
-error(2, "Cannot satisfy requirement group for package '{0}'", Package) :-
+error(2, "Cannot satisfy the requirements in packages.yaml for the '{0}' package. You may want to delete them to proceed with concretization. To check where the requirements are defined run 'spack config blame packages'", Package) :-
   activate_requirement_rules(Package),
   requirement_group(Package, X),
   not requirement_group_satisfied(Package, X).


alalazo:
@adamjstewart Not sure you consider this sufficient, but it's a low hanging fruit to improve the error message.

adamjstewart:
Much better than it was before!

alalazo:
@adamjstewart Then if you approve I'll merge it as soon as tests pass.

haampie (approved):
I don't like error messages created in the Spack core referring to spack commands, since spack commands are just one of the many ways this error can be triggered. This also causes weird error messages in buildcache commands for example, where some errors refer to `spack buildcache` args when using `spack install` or the other way around; they take different flags. So preferably this error message is created a few layers up. But at least something is better, so hitting approve.

alalazo:
> So preferably this error message is created a few layers up. But at least something is better, so hitting approve.

Hmm, I don't get your comment. This is a concretization error, that could happen _only_ because of something in `packages.yaml`. The error message says how to use Spack to check your `packages.yaml` and doesn't refer to any other argument :thinking:

alalazo:
@spackbot run pipeline

haampie:
No, `packages.yaml` is a way to construct a `ConfigScope`, you can get those instances without packages.yaml files. I'm just saying the concretizer doesn't take a `packages.yaml` as input, that happens many layers up.

The point is that guessing the path that leads to an error where the error happens is always wrong, it should just bubble up as an exception until it reaches a layer where it's clear "you have this offending config file on the filesystem".

adamjstewart:
@haampie I get where you're coming from. I just tried to install a package but had a typo in the name, and Spack told me I needed to run `spack clean -m`, which obviously isn't going to help. If we could make these more fine-grained that would be better.

PR #33990, opened 2022-11-18
Experiment: On M1 Mac, we need to run codesign after patching binaries

bernhardkaindl:
Tries to fix gitlab-ci by using `codesign` after `modify_macho_object` to sign relocated macho binaries.

RFC - Untested:

I have no access to an Apple M1 running Monterey, to be checked by gitlab-ci pipeline `e4s-mac-pr-build`

If this does not fix the gitlab-ci pipeline `e4s-mac-pr-build`, we'll have to disable the buildcache for M1 MacOS.


bernhardkaindl:
@spackbot run-pipeline

bernhardkaindl:
codesign --verify fails now, we need the output from it, need to modify/replace the `which()` to see the output.

PR #33992, opened 2022-11-18
Remove leftover `*_test_callbacks` attributes from PackageBase

### LINKED ISSUE #33928: Defaults in `package_base` unconditionally deactivate package checks for multiple build systems

alalazo:
fixes #33928 

The two attributes:
- `PackageBase.build_time_test_callbacks`
- `PackageBase.install_time_test_callbacks`

have been overlooked in #30738, where they should have been removed. Leaving them in place causes issues with the run-time adapter, that stops at `PackageBase` rather than falling back to the base class of the builder.

Modifications:
- [x] Add a regression test for the bug in #33928
- [x] Remove the attributes from `PackageBase` and fix the minor failures due to that

### DIFF
--- a/lib/spack/spack/audit.py
+++ b/lib/spack/spack/audit.py
@@ -287,7 +287,7 @@ def _check_build_test_callbacks(pkgs, error_cls):
     errors = []
     for pkg_name in pkgs:
         pkg_cls = spack.repo.path.get_pkg_class(pkg_name)
-        test_callbacks = pkg_cls.build_time_test_callbacks
+        test_callbacks = getattr(pkg_cls, "build_time_test_callbacks", None)
 
         if test_callbacks and "test" in test_callbacks:
             msg = '{0} package contains "test" method in ' "build_time_test_callbacks"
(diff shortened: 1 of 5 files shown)


alalazo:
@chissg This should fix the issue We can iterate on the audit in your PR.

haampie (approved):

PR #33994, opened 2022-11-18
gcc: Ensure matching assembler/binutils on RHEL8

bernhardkaindl:
`gcc@10:` Newer binutils than RHEL8's are required to for guaranteed operaton.

Therefore, on RHEL8-ish, conflict with `~binutils you need to add +binutils to be safe.

See this discussion with the OpenBLAS devs for reference:
https://github.com/xianyi/OpenBLAS/issues/3805#issuecomment-1319878852

Needs #33190 to makje rhel8, rock8 and almalinux8 equivalent.


tldahlgren (approved):
LGTM.  Will give maintainers time to comment.

sethrj (approved):

PR #33995, opened 2022-11-18
pflotran: add "rxn" variant

balay:
potentially work-around for build failure with cray compilers [due to incomplete F2003 support]

cc: @ghammond86


tldahlgren (approved):
LGTM

PR #34001, opened 2022-11-18
spack uninstall: reduce verbosity

haampie:
The following is terribly verbose in Spack 0.19.0 when you have named
environments:
   
```console
spack uninstall --all # or %gcc
```
   
as we print the environments the spec is in _for each matching spec in the
database_. Also, we print a separate section of dependents per matching
spec. (Notice not per input spec on the command line, but the possibly
hundreds of matching ones in the database.)
   
Instead, this PR takes the more pragmatic approach of showing error messages
as follows:

Screenshot from 2022-11-22 16-33-30

### DIFF
--- a/lib/spack/spack/cmd/uninstall.py
+++ b/lib/spack/spack/cmd/uninstall.py
@@ -235,6 +235,7 @@ def get_uninstall_list(args, specs, env):
             tty.info("The following environments still reference these specs:")
             colify([e.name for e in other_dependent_envs.keys()], indent=4)
             msgs.append("use `spack env remove` to remove environments")
+        msgs.append("use `spack uninstall --force` to override")
         print()
         tty.die("There are still dependents.", *msgs)
 
--- a/lib/spack/spack/cmd/uninstall.py
+++ b/lib/spack/spack/cmd/uninstall.py
@@ -6,6 +6,7 @@
 from __future__ import print_function
 
 import sys
+from typing import Dict, List, Optional
 
 from llnl.util import tty
 from llnl.util.tty.colify import colify
@@ -16,6 +17,7 @@ import spack.environment as ev
 import spack.error
 import spack.package_base
 import spack.repo
+import spack.spec
 import spack.store
 import spack.traverse as traverse
 from spack.database import InstallStatuses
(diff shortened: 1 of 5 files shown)


haampie:
@scheibelp thoughts?

lib/spack/spack/cmd/uninstall.py
```diff
@@ -344,23 +257,9 @@ def get_uninstall_list(args, specs, env):
...
-
-    all_uninstall_specs -= remove_only
-    # Inefficient topological sort: uninstall dependents before dependencies
-    all_uninstall_specs = sorted(
-        all_uninstall_specs, key=lambda x: sum(1 for i in x.traverse()), reverse=True
-    )
+    remove_specs = all_specs_in_env(env, all_uninstall_specs) if args.remove else []
 
-    return list(all_uninstall_specs), list(remove_specs)
+    return list(frozenset(all_uninstall_specs) - frozenset(remove_only)), remove_specs
```
alalazo: Why are we using a frozenset instead of a set?

lib/spack/spack/cmd/uninstall.py
```diff
@@ -78,7 +78,7 @@ def setup_parser(subparser):
     )
 
 
-def find_matching_specs(env, specs, allow_multiple_matches=False, force=False, origin=None):
+def find_matching_specs(env, specs, allow_multiple_matches=False, origin=None):
```
alalazo: Can we add type hints to the functions in this file?

lib/spack/spack/cmd/uninstall.py
```diff
@@ -128,91 +128,50 @@ def find_matching_specs(env, specs, allow_multiple_matches=False, force=False, o
...
+
+    specs = traverse.traverse_nodes(
+        specs,
+        root=False,
+        order="breadth",
+        cover="nodes",
+        deptype=("link", "run"),
+        direction="parents",
+        key=lambda s: s.dag_hash(),
+    )
```
alalazo, line 155: :+1:

lib/spack/spack/cmd/uninstall.py
```diff
@@ -128,91 +128,50 @@ def find_matching_specs(env, specs, allow_multiple_matches=False, force=False, o
...
-    spec_inactive_envs = {}
-    for spec, de_list in spec_envs.items():
-        inactive = [de for de in de_list if not de.active]
-        if inactive:
-            spec_inactive_envs[spec] = inactive
 
-    return spec_inactive_envs
+def all_specs_in_env(env, specs):
+    """Given a list of specs, return those that are in the env"""
+    hashes = frozenset(env.all_hashes())
```
alalazo: Same question on frozenset

alalazo (changes requested):
A couple of questions and a single minor request (type-hints). The output is really much improved when environments are present.

lib/spack/spack/cmd/uninstall.py
haampie: I think when I wrote this I mistakenly thought frozenset had better performance

alalazo (approved):

alalazo:
@spackbot run pipeline

PR #34002, opened 2022-11-18
py-mypy: add new versions; add new py-types packages

adamjstewart:
Successfully builds on macOS 12.6.1 (arm64) with Python 3.10.8 and Apple Clang 14.0.0.


tldahlgren (approved):
Confirmed the 5 new sha256 for py-mypy and each sha256 for the 3 new packages.

PR #32522, opened 2022-09-06
LAMMPS: Add new versions

rbberger:
* adds missing versions to LAMMPS spackage
* update official website URL
* add myself as a maintainer (I am one of the LAMMPS core devs and want to start to be more actively involved in supporting this spackage)
* deprecate versions before stable release on March 3, 2020 (I would deprecate more, but I want to do it gradually, this is a suggestion for an initial batch)

PR #31181, opened 2022-06-17
[lammps] Adding new CMake package names

stephenmsachs:
This is a workaround that includes new and old package names. 

I currently do not see any ideal solution. You could look at every release which package names are the valid ones, but then the spack user would need to change their `variant` specifiers depending on Lammps version.

Another solution would be to have Spack translate package names. Then the question remains whether the old or new ones should be used as `variant` specifiers in Spack.

The full list of changes (with release dates) can be found here:
https://github.com/lammps/lammps/commits/patch_7Jan2022/cmake/presets/all_on.cmake

PR #34011, opened 2022-11-18
lammps: Refactor variants, add new versions

rbberger:
Given that both #25015 and #31181 seem to be stale PRs, I've started my own. Maybe we can get this one through the finish line.

This is based on #32522 but does a whole set of additional changes to the variants and the available versions.

* Make use of conditional variants and remove redundant conflicts
* Add rules for how to add LAMMPS releases
  - new stable versions should be marked as preferred=True
  - a stable version that has updates and any of its outdated update releases should be marked deprecated=True
  - patch releases older than a stable release should be marked deprecated=True
* Update `url_for_version` to download `stable_*.tar.gz` instead of `patch_*.tar.gz`, which changes the checksums for some stable releases (I only updated them down to March 3, 2020 and left off `preferred=True` in older stable releases, that way they continue downloading the `patch_*.tar.gz` tarballs)
* Deprecate all patch releases older than a stable release
* Deprecate all releases older than the March 3, 2020 stable release
* Add December 22, 2022 patch release, adds new ml-pod variant
* Change defaults of variants. E.g. remove optional dependencies such as FFMPEG, PNG and JPG support. Default selected packages represent the same as the `basic` CMake preset, which usually is enough to let the unit tests pass.
* Add missing February 10, 2021 and July 27, 2021 patch releases
* Add missing package variants


bernhardkaindl:
@rbberger Still a work in progress? (This PR is marked as draft: When ready, click the "Ready for review" button!)

rbberger:
@bernhardkaindl yes, I still got some more changes incoming that need testing and will change defaults and dependencies. Overall clean up. Once the other PR gets merged I'll rebase this one and continue here. Unless ppl want to directly go to this one.

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -855,25 +598,32 @@ def cmake_args(self):
...
-            if cuda_arch != "none":
-                args.append("-DGPU_ARCH=sm_{0}".format(cuda_arch[0]))
-            args.append(self.define_from_variant("CUDA_MPS_SUPPORT", "cuda_mps"))
-        elif spec.satisfies("+opencl"):
-            args.append("-DPKG_GPU=ON")
-            args.append("-DGPU_API=opencl")
-        else:
-            args.append("-DPKG_GPU=OFF")
+        if "~kokkos" in spec:
+            # LAMMPS can be build with the GPU package OR the KOKKOS package
```
RemiLacroix-IDRIS, line 637: It feels like you are silently ignoring `+ cuda` when `+kokkos`, wouldn't it be better to add an explicit conflict?

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -204,146 +309,154 @@ def url_for_version(self, version):
             update = ""
         else:
             update = "_update{0}".format(split_ver[1])
-        return "https://github.com/lammps/lammps/archive/patch_{0}{1}.tar.gz".format(
-            vdate.strftime("%d%b%Y").lstrip("0"), update
+
+        is_stable = (
+            version in self.versions
+            and "preferred" in self.versions[version]
+            and self.versions[version]["preferred"]
```
RemiLacroix-IDRIS, line 327: Looking at this I understand that all stable versions should be marked `preferred` and that patch releases should never be marked preferred, is this correct?

If it is, I'm not sure to understand the comment:
> only the latest update of any given stable release should be marked preferred=True

var/spack/repos/builtin/packages/lammps/package.py
rbberger: I didn't want to add a conflict yet. I still have much more GPU related changes incoming and don't want to be too restrictive.

var/spack/repos/builtin/packages/lammps/package.py
rbberger: Stable release sometimes have update releases. If a stable has an update release, that usually means there was a serious enough bug that the stable version prior to the update shouldn't be used anymore. E.g. if there was a `.1` update release before and there is now a `.2` update release, only the `.2` should be preferred.

rbberger:
@bernhardkaindl I think this is ready for review. There is already plenty of things in this. I still have a few open issues that I want to address (GPU/Kokkos package with ROCm support), but probably better in a new PR.

var/spack/repos/builtin/packages/lammps/package.py
rbberger: Just to clarify, what I would want in the end is `+gpu+opencl`, `+gpu+cuda`, `+gpu+rocm` OR `+kokkos+cuda`, `+kokkos+rocm`. So if I would add a conflict, it would be between `+gpu` and `+kokkos`, even though it will probably even compile and there might be cases were we don't want that conflict. It just is a source of troubles if you try to mix them in the same binary. But I'm thinking of moving the GPU related things into a new PR.

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -25,47 +25,152 @@ class Lammps(CMakePackage, CudaPackage):
...
+        deprecated=True,
+    )
+    version(
+        "20211027",
+        sha256="c06f682fcf9d5921ca90c857a104e90fba0fe65decaac9732745e4da49281938",
+        deprecated=True,
+    )
+    version(
+        "20210929.3",
+        sha256="e4c274f0dc5fdedc43f2b365156653d1105197a116ff2bafe893523cdb22532e",
```
tldahlgren, line 89: I'm getting a different sha256 here:

```suggestion
        sha256="d7df30f7ce0bd8b7c21949cf5bdf1877858815379cddb8a264da7a8cbf699639",
```

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -25,47 +25,152 @@ class Lammps(CMakePackage, CudaPackage):
...
     )
     version(
-        "20210929.1", sha256="3b792e20864bf88b855332486996f2c540deabb4e3507e48fa4ee96ad79615ec"
+        "20200505",
+        sha256="c49d77fd602d28ebd8cf10f7359b9fc4d14668c72039028ed7792453d416de73",
+        deprecated=True,
+    )
+    version(
+        "20200303",
+        sha256="a1a2e3e763ef5baecea258732518d75775639db26e60af1634ab385ed89224d1",
```
tldahlgren, line 182: Also inconsistent with `spack checksum` result:

```suggestion
        sha256="9aa56dfb8673a06e6c88588505ec1dfc01dd94f9d60e719ed0c605e48cc06c58",
```

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -25,47 +25,152 @@ class Lammps(CMakePackage, CudaPackage):
 
     maintainers = ["rbberger"]
 
+    # rules for new versions and deprecation
+    # * new stable versions should be marked as preferred=True
+    # * a stable version that has updates and any of its outdated update releases should be
+    #   marked deprecated=True
+    # * only the latest update of any given stable release should be marked preferred=True
```
tldahlgren: Maybe move this to between lines 29 and 30?

https://spack.readthedocs.io/en/latest/packaging_guide.html#version-selection

tldahlgren (commented):
Initial review comments.

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -25,47 +25,152 @@ class Lammps(CMakePackage, CudaPackage):
...
     version(
-        "20220623.2", sha256="fdb5474135b17005030cd78ac85762fecf0944bdd27a81fbbe79b2b4522ccae7"
+        "20220623.2",
+        sha256="8a560213e83919623525c4a7c4b5f0eda35cdf3b0c0e6548fd891379e04ca9e6",
+        preferred=True,
     )
     version(
-        "20220623.1", sha256="1d4fb06a4621b271e8b54f31fef241d0569717a69c1b7646a2b76fe5ce47c966"
+        "20220623.1",
+        sha256="58e3b2b984f8935bb0db5631e143be2826c45ffd48844f7c394f36624a3e17a2",
```
tldahlgren, line 42: On a second check, after fetching the PR, `spack checksum` indicates the sha256 should not be change, i.e.,:

```suggestion
        sha256="1d4fb06a4621b271e8b54f31fef241d0569717a69c1b7646a2b76fe5ce47c966",
```

var/spack/repos/builtin/packages/lammps/package.py
rbberger: I realized that last "rule" created a problem with the sha256 for updates. Since I'm relying on the fact of `preferred=True` means to download the `stable_` tarballs, the checksum gets messed up if we then remove it later on (because that causes a `patch_` tarball to be downloaded).. So, for consistency, all stable versions (starting from 2020) should have `preferred=True`, independent of whether they are deprecated or not.

var/spack/repos/builtin/packages/lammps/package.py
rbberger: @RemiLacroix-IDRIS turns out you were right, there was a logical error in that last rule, since it conflicted with the others.

var/spack/repos/builtin/packages/lammps/package.py
rbberger: I've fixed my logical mistake, added some necessary `preferred=True`, and incorporated the changes from #35133. Could you please recheck? I went through all stable releases (from 2020 upward) and ran `spack checksum` on them. I would be surprised if there is still some mismatch (unless there are some proxy troubles on my end). Can you verify that they pull the `stable_` tarballs as intended?

tldahlgren (dismissed):
(Re-)Confirmed branch and all 57 version sha256.

tldahlgren:
@RemiLacroix-IDRIS Let me know if you're okay with this being merged.

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -204,148 +310,157 @@ def url_for_version(self, version):
...
+        "electrode": {"when": "@20220504:"},
+        "granular": {},
+        "intel": {},
+        "kim": {},
+        "kspace": {"default": True},
+        "kokkos": {},
+        "latte": {},
+        "manybody": {"default": True},
+        "mc": {},
+        "meam": {"when": "@20210728:"},
```
RemiLacroix-IDRIS: Meam existed before, in my tentative PR I had the following conflicts:
```python
    conflicts(
        '+meam', when='@20190104:20210527',
        msg='+meam was removed from version @20190104 to version @20210527, ' +                                                          
            'use +user-meamc instead')
    conflicts(
        '+user-meamc', when='@:20181212,20210702:',
        msg='+user-meamc only supported after version @20181212 and before ' +                                                           
            'version @20210702, use +meam instead')
```

RemiLacroix-IDRIS (changes requested):
Looks good to me, just meam was added, then removed and finally added back.

var/spack/repos/builtin/packages/lammps/package.py
rbberger: @RemiLacroix-IDRIS Went over all packages and corrected their version ranges. Wrote myself a little tool that went over the git history:
https://gist.github.com/rbberger/fdaa38ff08e5961c4741624a4719cdb6

rbberger:
@tldahlgren @RemiLacroix-IDRIS from my end this is ready.

tldahlgren (approved):
Confirmed all 25 of the affected versions' sha256.

tldahlgren:
Will give @RemiLacroix-IDRIS time to weigh in.

RemiLacroix-IDRIS (approved):
LGTM.

PR #34020, opened 2022-11-20
gitlab-ci: disable buildmirror for M1 Macs for now: relocation breaks codesigning

bernhardkaindl:
Relocation of the extracted binary cache breaks code signatures on M1 Macs because it changes the binaries.

codesign could re-sign the binaries, but it's not done yet(see #33990),
so disable use of the buildcache mirror in gitlab-ci for now.


haampie:
This doesn't make sense

PR #34021, opened 2022-11-20
krb5: Add new versions 1.19.4 and 1.20.1

bernhardkaindl:
Added the latest maintenance-versions of the current and previous minor versions.


michaelkuhn (approved):
Works for me.

PR #34022, opened 2022-11-20
bash: Update 5.1 to current patchlevel 5.1.16

bernhardkaindl:
Updated bash-5.1 to the current patchlevel 5.1.16


adamjstewart (approved):
Thanks, I've been meaning to update this for a long time. Do you also want to add 5.2 and patches?

adamjstewart:
> Do you also want to add 5.2 and patches?

Done in #34301

PR #34024, opened 2022-11-20
petsc, slepc: enable parallel builds

balay:


joseeroman (dismissed):

var/spack/repos/builtin/packages/petsc/package.py
```diff
@@ -206,6 +206,7 @@ def check_fortran_compiler(self):
     patch("disable-DEPRECATED_ENUM.diff", when="@3.14.1 +cuda")
     patch("revert-3.18.0-ver-format-for-dealii.patch", when="@3.18.0")
 
+    depends_on("gmake", type="build")
```
haampie: We typically don't add a dep on `gmake`, we assume it's on the system.

balay: I was trying to switch gmake in ~/.spack/packages.yaml (to a different version) - and noticed it wasn't getting picked up by the build - hence assumed - this change is preferred [than picking up whatever is in default PATH.]

Or shouldn't spack be automatically using this [gmake spec] over  default path - without the above code?

balay: reverted the `gmake` change for now

balay: Actually we do require gmake for petsc/slepc builds. [so bsd make in PATH won't be sufficient.]

balay: Perhaps we can defer this issue - how best should spack handle default `make` used by pkgs - and merge in the current change [enable parallel build]

Not sure who in spack team to tag. cc: @tgamblin

bernhardkaindl: @balay: Spack implicitly requires GNU make for many packages.

Without it, I'm sure that the dependencies of petsc, slepc won't be building.

I think you can be certain, and implicitly depend on that the `make` command provied in spack is really GNU `make`.

bernhardkaindl (approved):
Test-built on a 48-core machine, build was fast.

PR #34026, opened 2022-11-20
py-numpy: add v1.23.5

adamjstewart:
https://github.com/numpy/numpy/releases/tag/v1.23.5


rgommers (approved):
The hash matches, and all green - this should be painless:)

michaelkuhn (approved):

PR #29026, opened 2022-02-17
Environments: Add support for include URLs

tldahlgren:
Fixes #28840 

FYI. @adrienbernede 

This PR adds support in environments for external configuration files obtained from a URL with a preference for grabbing raw text from `GitHub` and `gitlab` for efficient downloads of the relevant files.

Spack currently assumes these files reside on the local file system but projects and facilities put a lot of effort into defining standard, or base, configuration files for their systems. Spack should make it easier to use configuration files fetched from the web.  

Some have system-specific configurations, more inclusive environment files, or both.  Examples can be found at:
- https://github.com/LLNL/radiuss-spack-configs
- https://github.com/spack/spack-configs

You can provide the URL for a folder, which will result in all `.yaml` files directly under it to be cached (under the environment), or a specific (`.yaml`) file.  Consequently, this PR works for managed/named and independent environments.

See https://spack.readthedocs.io/en/latest/environments.html#included-configurations for use.

STAGING/CACHING
This PR will *not* re-stage or overwrite configuration files with the same name if they have already been cached. This will ensure the configuration is unchanged *and*, should someone intentionally modify their cached file(s), any local changes are not replaced.

TBD
- What changes are needed to support sharing environments from open to closed networks?  (Note the environment's configuration subdirectory could be archived with the `spack.*` files and copied over.)
- What would be the syntax for including a repository URL and identifying the relevant configuration file(s)? 

TODO:
- [x] Switch to automatically caching the file(s) associated with a configuration-related URL under the environment root
- [x] Replace URL-based tests with approach used for for other fetch-related unit tests (per Todd)
- [x] Add more unit tests as needed
- [x] Remove commit 495b6d301be5efcb8d42666b1518e942088a5559 once #31449 is merged

PR #34029, opened 2022-11-21
Bootstrap most of Spack dependencies using environments

alalazo:
The main objective of this PR is to rework the bootstrapping procedure to use Spack environments as much as possible. 

While doing so, the `spack.bootstrap` module has been reorganized into a Python package. For reviewers, the new code is under `spack.bootstrap.environment` (most of the other changes are just adding docs or linting).

In this PR a distinction is made among "core" Spack dependencies (clingo, GnuPG, patchelf) and other dependencies. For a number of reasons, explained in the `spack.bootstrap.core` module docstring, "core" dependencies are bootstrapped with the current ad-hoc method. All the other dependencies are instead bootstrapped using a Spack environment that lives in a directory specific to the interpreter and the architecture being used. An example of the directory structure is:
```console
$ tree $(spack bootstrap root) -L 3
/home/culpo/.spack/bootstrap
├── config
│   └── linux
│       ├── bootstrap.yaml
│       ├── compilers.yaml
│       └── config.yaml
├── environments
│   └── python3.8-x86_64-57f95
│       ├── Makefile
│       ├── spack.lock
│       ├── spack.yaml
│       └── view -> /home/culpo/.spack/bootstrap/environments/python3.8-x86_64-57f95/._view/dgsyi3525yl3vzeakomxqrekchzd7l6s
└── store
    ├── bin
    │   └── sbang
    ├── linux-centos7-x86_64
    │   └── gcc-10.2.1
    └── linux-ubuntu20.04-x86_64
        └── gcc-9.4.0
```
The bootstrapping output has been reduced and now looks like:
Screenshot from 2022-11-21 07-00-59

Modifications:
- [x] Bootstrap most of Spack dependencies using a Spack environment
- [x] Add `pytest` to the list of bootstrapped dependencies
- [x] Reduced verbosity of bootstrapping output
- [x] Added a `--dev` argument to `spack bootstrap now`
- [x] Reorganized `spack.bootstrap` into a package and linted it


alalazo:
A few notes in addition to the description:
1. This PR makes it possible to set the bootstrap store and the user store in the same place. I still think it's a very bad idea to mix installations needed by Spack for its own working with installations needed by the user, but if people want to experiment with that it is now possible (@psakievich might be interested given the discussions during community meetings)
2. After this PR is in, we can delete `pytest` from the Python packages we vendor (see #33951)
3. The default environment modifications in https://github.com/spack/spack/blob/bdd665ed78d9bf756173d473d9e03b0f6f0ee5d4/lib/spack/spack/bootstrap/environment.py#L105-L119 are manual, since it seems environments by default add both views and "store" paths. In the case of PYTHONPATH this can result in a large number of directories to be searched, which clutters the environment and may slow-down Python.
4. To bootstrap on `rhel8` with `platform-python` and no Python headers we require both https://github.com/spack/spack/pull/34029/commits/d5142d1cb651705113c02b8b48b03ddb27b950e7 (install 2 Python extensions from wheels, otherwise we'd fail on headers being missing) and  https://github.com/spack/spack/pull/34029/commits/2ca96804873152cca4c12212ad634419b6fa1a4f (fall back on the Python command to be `/usr/libexec/platform-python` if a series of unfortunate circumstances occur :slightly_smiling_face: ). Installing from wheels is needed for 2 `mypy` dependencies, otherwise so far all the other Python packages we depend on are pure Python code.

psakievich:
@alalazo I agree with you that there are a lot of dangers in mixing the stores. I would primarily like to reuse the bootstrapped dependencies from `spack style` and soon to be `spack unit-test`.  I'm not totally sure of what else is involved beyond those and the "core" dependencies you've outlined in the PR description. I don't have the time to really dig into a PR this big right now so I'll ask a couple of questions instead.

1) Are the "core" dependencies (clingo, patchelf, GnuPG) included in the store mixing?  I think it would be best to keep those in a separate store/install_tree, and keep them off limits.  From reading the comments and doc string I believe that is correct?
2) How is the "store mixing" done?  Will `spack uninstall --all` blow away the bootstrap store, or will it be treated like an upstream with some level of additional protection?  I guess if it's just some python packages that get blown away then I can live with that.  If it is something more fundamental, then that is less ideal.

alalazo:
@psakievich 

> I'm not totally sure of what else is involved beyond those and the "core" dependencies you've outlined in the PR description.

https://github.com/spack/spack/blob/d5142d1cb651705113c02b8b48b03ddb27b950e7/lib/spack/spack/bootstrap/core.py#L5-L23

> Are the "core" dependencies (clingo, patchelf, GnuPG) included in the store mixing? [ ... ]

To be clear:
1. There is no mixing by default, things remain exactly as on `develop` with that respect. I just removed a check that was preventing people from setting the "bootstrapping store" and `config:install_tree:root` to the same place.
2. The mixing is done as you like, if you want to experiment with that. You can use the bootstrap store as an upstream, or you can set both stores to the same place.[^1]

[^1]: I wouldn't recommend any of the two, but they are possible

lib/spack/spack/bootstrap/environment.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+        make = spack.util.executable.which("make")
+        kwargs = {}
+        if not tty.is_debug():
+            kwargs = {"output": os.devnull, "error": os.devnull}
+        make(
+            "-C",
+            str(self.environment_root()),
+            "-j",
+            str(spack.build_environment.determine_number_of_jobs(parallel=True)),
```
haampie, line 140: One issue here is that it will regenerate the view for every package being installed

alalazo: This is in any case much faster than a serial installation, so I guess we can optimize later to generate a depfile that can install without generating the view?[^1]

[^1]: It's also ugly that we have to spawn another process to generate the Makefile, but given the difference in performance I think it's preferable to `env.install_all()`.

lib/spack/spack/bootstrap/config.py
```diff
@@ -0,0 +1,168 @@
...
+
+@contextlib.contextmanager
+def _ensure_bootstrap_configuration():
+    bootstrap_store_path = store_path()
+    user_configuration = _read_and_sanitize_configuration()
+    with spack.environment.no_active_environment():
+        with spack.platforms.prevent_cray_detection():
+            with spack.platforms.use_platform(spack.platforms.real_host()):
+                with spack.repo.use_repositories(spack.paths.packages_path):
+                    with spack.store.use_store(bootstrap_store_path):
```
trws: Thoughts on maybe collapsing this to a single `with` statement with multiple values? The semantics are the same, but avoiding the extra four levels of nesting might be nice.

var/spack/repos/builtin/packages/py-black/package.py
```diff
@@ -52,6 +52,9 @@ class PyBlack(PythonPackage):
     depends_on("py-ipython@7.8:", when="+jupyter", type=("build", "run"))
     depends_on("py-tokenize-rt@3.2:", when="+jupyter", type=("build", "run"))
 
+    # Needed because this package is used to bootstrap Spack (Spack supports Python 3.6+)
+    depends_on("py-dataclasses@0.6:", when="^python@:3.6", type=("build", "run"))
```
trws, line 56: Are we supporting anything this would match for anymore? I may be misreading but this only applies below 3.6 right?

trws: Nevermind, I misread it, sorry.

trws:
All my testing shows this working well, mac, linux, arm, x86, I have yet to find a problem with bootstrapping.  The one thing that struck me as a bit odd though, I'm getting output describing what's going to be built *after* it's built rather than before but I'm not seeing obviously why that is:

```
==> clingo-bootstrap: Successfully installed clingo-bootstrap-spack-ls7pgf6b6ytnlgtgim4cpdsfvm4hrmqo
  Stage: 0.33s.  Cmake: 2.03s.  Build: 28.68s.  Install: 0.55s.  Total: 31.73s
[+] /Users/scogland1/.spack/bootstrap/store/darwin-monterey-aarch64/apple-clang-14.0.0/clingo-bootstrap-spack-ls7pgf6b6ytnlgtgim4cpdsfvm4hrmqo
==> [BOOTSTRAPPING] Spack has missing dependencies, creating a bootstrapping environment
==> [BOOTSTRAPPING] Installing dependencies (py-isort@4.3.5:, py-mypy@0.900:, py-black, py-flake8, py-pytest)
```

alalazo:
Took me a while to find, but the unit tests failing [here](https://github.com/spack/spack/actions/runs/3581628451/jobs/6025916090) are because some environment modification leak, fixed in b8f0964, of tests running before them. The unit tests have been modified in #30882

> The one thing that struck me as a bit odd though, I'm getting output describing what's going to be built after it's built rather than before but I'm not seeing obviously why that is: [ ... ]

That's because `clingo`, ` patchelf` and `GnuPG` are treated differently. We first bootstrap them with an ad-hoc procedure, then use the environment for everything else (currently py-isort@4.3.5:, py-mypy@0.900:, py-black, py-flake8, py-pytest) - so what you see is the bootstrapping of "core" software before the environment.

trws:
Got it, doing some more testing, but I'm reminded of a related bootstrap question.  Do we need to keep using a development version of clingo? As it is, we're forcing an increase in build dependencies to do it, is there no release new enough to get rid of the re2c and bison dependencies?  Especially with an external cmake and python available, the size of the build set is substantially larger because of it:

```
# spack version
clingo-bootstrap@spack%apple-clang@14.0.0~docs~ipo+python~static_libstdcpp build_system=cmake build_type=Release arch=darwin-monterey-m1
    ^bison@3.8.2%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
        ^diffutils@3.8%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
            ^libiconv@1.16%apple-clang@14.0.0 build_system=autotools libs=shared,static arch=darwin-monterey-m1
        ^gnuconfig@2022-09-17%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1
        ^m4@1.4.19%apple-clang@14.0.0+sigsegv build_system=autotools patches=9dc5fbd,bfdffa7 arch=darwin-monterey-m1
            ^libsigsegv@2.13%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
        ^perl@5.36.0%apple-clang@14.0.0+cpanm+shared+threads build_system=generic arch=darwin-monterey-m1
            ^berkeley-db@18.1.40%apple-clang@14.0.0+cxx~docs+stl build_system=autotools patches=26090f4,b231fcc arch=darwin-monterey-m1
            ^bzip2@1.0.8%apple-clang@14.0.0~debug~pic+shared build_system=generic arch=darwin-monterey-m1
            ^gdbm@1.23%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
                ^readline@8.1.2%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
                    ^ncurses@6.3%apple-clang@14.0.0~symlinks+termlib abi=none build_system=autotools arch=darwin-monterey-m1
                        ^pkgconf@1.8.0%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
            ^zlib@1.2.13%apple-clang@14.0.0+optimize+pic+shared build_system=makefile arch=darwin-monterey-m1
    ^cmake@3.24.2%apple-clang@14.0.0~doc+ncurses+ownlibs~qt build_system=generic build_type=Release arch=darwin-monterey-m1
    ^python@3.9.6%apple-clang@14.0.0+bz2+crypt+ctypes+dbm~debug+libxml2+lzma+nis~optimizations+pic+pyexpat~pythoncmd+readline+shared+sqlite3+ssl+tix+tkinter+uuid+zlib build_system=generic patches=0d98e93,f2fd060 arch=darwin-monterey-m1
    ^re2c@2.2%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1

# release version
clingo-bootstrap@5.5.2%apple-clang@14.0.0~docs~ipo+python~static_libstdcpp build_system=cmake build_type=Release arch=darwin-monterey-m1
    ^cmake@3.24.2%apple-clang@14.0.0~doc+ncurses+ownlibs~qt build_system=generic build_type=Release arch=darwin-monterey-m1
    ^py-cffi@1.15.0%apple-clang@14.0.0 build_system=python_pip arch=darwin-monterey-m1
        ^libffi@3.4.2%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
            ^gnuconfig@2022-09-17%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1
        ^pkgconf@1.8.0%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
        ^py-pip@22.2.2%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1
        ^py-pycparser@2.21%apple-clang@14.0.0 build_system=python_pip arch=darwin-monterey-m1
        ^py-setuptools@65.5.0%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1
        ^py-wheel@0.37.1%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1
    ^python@3.9.6%apple-clang@14.0.0+bz2+crypt+ctypes+dbm~debug+libxml2+lzma+nis~optimizations+pic+pyexpat~pythoncmd+readline+shared+sqlite3+ssl+tix+tkinter+uuid+zlib build_system=generic patches=0d98e93,f2fd060 arch=darwin-monterey-m1

**Comment by alalazo** [comment]
> Do we need to keep using a development version of clingo?

We are using our binaries most of the time, which are installing a single node. So the explosion in dependencies is there only when building from sources (this didn't change in this PR). That said, that particular version was a requirement up until v0.19.0 (we pinned to the last commit of clingo supporting Python 2.7). We can look into updating clingo for v0.20.0.

If you can't use binaries on `m1`, do you know if you fall in one of these cases https://github.com/spack/spack/issues/33987 ?

**Comment by trws** [comment]
I'm intentionally bootstrapping from source to test everything.  We can certainly look into updating clingo separately, just avoiding that darned perl build would be a bonus.  Bit of a surprise on an ubuntu container too actually, apparently we don't bootstrap curl but try to use it?  We also don't check for ca-certificates existing, so we get curl missing and try `spack -k`, but only when it hits patchelf (gnuconfig installs without either). 

```
==> Installing patchelf-0.16.1-cljjfq4ynfu2txmln6mbcn4kxv2jfvj7
==> No binary for patchelf-0.16.1-cljjfq4ynfu2txmln6mbcn4kxv2jfvj7 found: installing from source
==> Error: spack requires 'curl'. Make sure it is in your path.
==> Error: spack requires 'curl'. Make sure it is in your path.
==> Error: spack requires 'curl'. Make sure it is in your path.
==> Error: spack requires 'curl'. Make sure it is in your path.
==> Warning: Spack was unable to fetch url list due to a certificate verification problem. You can try running spack -k, which will not check SSL certificates. Use this at your own risk.
```

alalazo:
> Bit of a surprise on an ubuntu container too actually, apparently we don't bootstrap curl but try to use it? 

`curl` is not managed by the bootstrapping mechanism currently. I thought we relied on Python `urllib` by default, but something might have changed in #29026. Need to check that PR more closely.

trws:
Based on what I'm seeing it definitely tries urllib, but as a fallback from curl, and it doesn't try to grab ca-certificates if they're unavailable.  Normally we can reasonably expect those, but it might be nice to have a way to grab them if they're missing.

alalazo:
> Based on what I'm seeing it definitely tries urllib, but as a fallback from curl, and it doesn't try to grab ca-certificates if they're unavailable. Normally we can reasonably expect those, but it might be nice to have a way to grab them if they're missing.

I think it's a bug introduced in #29026. Pinged you in a discussion there too :slightly_smiling_face:

tldahlgren:
> > Based on what I'm seeing it definitely tries urllib, but as a fallback from curl, and it doesn't try to grab ca-certificates if they're unavailable. Normally we can reasonably expect those, but it might be nice to have a way to grab them if they're missing.
> 
> I think it's a bug introduced in #29026. Pinged you in a discussion there too 🙂

It is.  If no-one else is already fixing it I will.

alalazo:
@spackbot run pipeline

alalazo:
@trws Is there any further request for changes?

trws:
I don't see any more problems with this, but now that core bootstrap is split out it looks like it should be a great deal easier to solve the bootstrap locking problem.  In fact I think you have solved it for everything but the core set because of the way the environment install works.  Would you be willing to take a stab at adding locking to the bootstrap.core end, perhaps shortly after this merges?

trws (approved):

alalazo:
> Would you be willing to take a stab at adding locking to the bootstrap.core end, perhaps shortly after this merges?

Sure!

alalazo:
@spackbot run pipeline

alalazo:
Merging, since the spec only that is reported broken is also broken in develop, see https://gitlab.spack.io/spack/spack/-/jobs/4817633

alalazo:
I'll work on fixing that spec next. For the record the pipeline finished here https://gitlab.spack.io/spack/spack/-/pipelines/233092 and so far never reported back to the PR status:
Screenshot from 2022-12-06 11-51-05

tgamblin:
I am not 100% sure it's this PR, but I am seeing a lot of random bootstrap failures in GHA and I think it may be due to this PR.

PR #34032, opened 2022-11-21
update easi package

Thomas-Ulrich:
Add package impalajit
 Update easi package
- add version1.2.0
- allow using impalajit and lua together
- now use default impalajit repository (uphoffc/ImpalaJIT has been merged into the main repository).


ravil-mobile (dismissed):
~~Do we really need to delete these? As far as I remember it compiles default (legacy Impala jit)~~

ravil-mobile (dismissed):

var/spack/repos/builtin/packages/impalajit/package.py
```diff
@@ -0,0 +1,29 @@
...
+
+class Impalajit(CMakePackage):
+    """A lightweight JIT compiler for flexible data access in simulation applications."""
+
+    homepage = "https://github.com/Manuel1605/ImpalaJIT/blob/master/README.md"
+    version("develop", git="https://github.com/Manuel1605/ImpalaJIT.git", branch="master")
+    maintainers = ["Thomas-Ulrich", "ravil-mobile"]
+    variant("static", default=True, description="compile as a static lib")
+    depends_on("cmake", type="build")
+    depends_on("pkg-config", type="build")
```
alalazo: We already have `impalajit-llvm`, which seems to build from another fork. This seems to be the upstream repository: https://github.com/manuel-fasching/ImpalaJIT 

Can you explain the relationship among these and why do we need to have two packages from fork repos (and not have the upstream)?

alalazo (changes requested):

var/spack/repos/builtin/packages/impalajit/package.py
Thomas-Ulrich: Hi,
impalajit is restricted to x86 architectures, while impalajit-llvm is supposed to work on any architecture supported by llvm. For example, @ravil-mobile used impalajit-live on uncommon architectures likes PowerPC. Downsizes are that impalajit-llvm is pretty annoying to install (extra dependencies), not well tested, and has no advantage over impalajit on x86 architectures.
impalajit-llvm lies on a fork as its developer (@ravil-mobile ) worked independently of the developer of impalajit.
I think it make sense to have 2 packages.
I could add some text from above in the documentation of impalajit-llvm?

Thomas-Ulrich: done

alalazo: Since these two packages are forks of the same upstream, can't we make them a single package and differentiate by the version?

Thomas-Ulrich: ok, I can do that.

Thomas-Ulrich: Done

alalazo (approved):

alalazo:
Thanks!

PR #34036, opened 2022-11-21
libelf: fix build on macOS arm64

adamjstewart:
Successfully builds and passes all tests on macOS 12.6.1 (arm64) with Apple Clang 14.0.0.

### DIFF
--- a/var/spack/repos/builtin/packages/libelf/package.py
+++ b/var/spack/repos/builtin/packages/libelf/package.py
@@ -29,8 +29,23 @@ class Libelf(AutotoolsPackage):
 
     provides("elf@0")
 
+    # configure: error: neither int nor long is 32-bit
+    depends_on("automake", when="platform=darwin target=aarch64:", type="build")
+    depends_on("autoconf", when="platform=darwin target=aarch64:", type="build")
+    depends_on("libtool", when="platform=darwin target=aarch64:", type="build")
+    depends_on("m4", when="platform=darwin target=aarch64:", type="build")
+
+    @property
+    def force_autoreconf(self):
+        return self.spec.satisfies("platform=darwin target=aarch64:")
+
     def configure_args(self):
-        args = ["--enable-shared", "--disable-dependency-tracking", "--disable-debug"]
+        args = ["--enable-shared", "--disable-debug"]
+
+        # config.sub: invalid option -apple-darwin21.6.0
+        if self.spec.satisfies("platform=darwin target=aarch64:"):
+            args.append("--build=aarch64-apple-darwin")
+
         return args
 
     def install(self, spec, prefix):


var/spack/repos/builtin/packages/libelf/package.py
```diff
@@ -29,8 +29,23 @@ class Libelf(AutotoolsPackage):
...
 
+    # configure: error: neither int nor long is 32-bit
+    depends_on("automake", when="platform=darwin target=aarch64:", type="build")
+    depends_on("autoconf", when="platform=darwin target=aarch64:", type="build")
+    depends_on("libtool", when="platform=darwin target=aarch64:", type="build")
+    depends_on("m4", when="platform=darwin target=aarch64:", type="build")
+
+    @property
+    def force_autoreconf(self):
+        return self.spec.satisfies("platform=darwin target=aarch64:")
```
adamjstewart, line 40: Homebrew and Conan use the same approach

var/spack/repos/builtin/packages/libelf/package.py
```diff
@@ -29,8 +29,23 @@ class Libelf(AutotoolsPackage):
...
+
     def configure_args(self):
-        args = ["--enable-shared", "--disable-dependency-tracking", "--disable-debug"]
+        args = ["--enable-shared", "--disable-debug"]
```
adamjstewart, line 43: The `--disable-dependency-tracking` flag is not recognized by configure

var/spack/repos/builtin/packages/libelf/package.py
```diff
@@ -29,8 +29,23 @@ class Libelf(AutotoolsPackage):
...
+
+        # config.sub: invalid option -apple-darwin21.6.0
+        if self.spec.satisfies("platform=darwin target=aarch64:"):
+            args.append("--build=aarch64-apple-darwin")
```
adamjstewart, line 47: We use the same approach in the `libffi` package. Not sure if there's a better way to do this or any idea why it doesn't recognize the arch automatically.

var/spack/repos/builtin/packages/libelf/package.py
bernhardkaindl: Yep, no problem

var/spack/repos/builtin/packages/libelf/package.py
bernhardkaindl: Yep, this option should never be needed, it's only a shortcut.

var/spack/repos/builtin/packages/libelf/package.py
bernhardkaindl: Maybe the update of gnuconfig I just merged now could help, but setting it fixed should make it more stable. In any case, this is the most safe change.

bernhardkaindl (approved):
Still builds on Linux, checked the review discussion, looks good.

adamjstewart:
@spackbot run pipeline

PR #34039, opened 2022-11-21
zlib: fix shared libraries when '%nvhpc'

skosukhin:
@haampie you will probably like this


haampie:
sigh :(

(personally I think zlib should be replaced with zstd everywhere)

alalazo:
@spackbot run pipeline

alalazo (approved):

haampie:
Ugh, this really feels like it's not the way :|

alalazo:
It's a fix for `%nvhpc` only, happy to merge anything better.

haampie:
I don't agree with the soname change (it's not necessarily related to the zlib version number), and I think it's better to spend the energy fixing this upstream instead of in Spack. Also Mark Adler is open for that: https://github.com/madler/zlib/blob/04f42ceca40f73e2978b50e93806c2a18c1281fc/configure#L360

PR #29287, opened 2022-03-02
snakemake: Updated to new version 7.0.2

marcusboden:
Added new snakemake versions and a dependency

PR #34041, opened 2022-11-21
Add new version of snakemake

charmoniumQ:
Closes #29287


bernhardkaindl (dismissed):
Builds for me, but maybe wait for review from `adamjstewart` a bit (he reviews python packages)

charmoniumQ (commented):
Updated comments

var/spack/repos/builtin/packages/py-throttler/package.py
```diff
@@ -0,0 +1,21 @@
...
+    homepage = "https://github.com/uburuntu/throttler"
+    pypi = "throttler/throttler-1.2.1.tar.gz"
+
+    maintainers = ["charmoniumQ"]
+
+    version("1.2.1", sha256="8b23d3485a96d98484024a850c1887ccc685bead17e86c8a9e4b0335e7d74778")
+
+    depends_on("py-setuptools", type="build")
+
+    patch("setup.patch")
```
adamjstewart: Why is this patch necessary?

charmoniumQ: The `setup.py` references files that are in the Git repository but not the PyPI archived tarball, so the vendored `setup.py` will fail.

adamjstewart: Ah, I was looking at 1.2.2 (released yesterday) which fixes this. Do you want to just add 1.2.2 and drop 1.2.1 and the patch?

var/spack/repos/builtin/packages/snakemake/package.py
```diff
@@ -13,6 +13,7 @@ class Snakemake(PythonPackage):
     pypi = "snakemake/snakemake-6.12.3.tar.gz"
     maintainers = ["marcusboden"]
 
+    version("7.18.2", sha256="23f52b9a0c86da3b974a3cfc097fa82b41c49dab05543c0d18377c854852f771")
     version("6.15.1", sha256="a219601d57037f565ead9963e6bd8d04d3bdd985d172371e54197dcbdba79865")
     version("6.13.1", sha256="22f57dcd8b1ca8a30aaa45c5d2c0f56d381d4731abd0988f24f9de46b7d9827c")
     version("6.12.3", sha256="af86af9a540da3dceb05dad1040f1d3d733e6a695f8b3f8c30f8cf3bc6570a88")
```
adamjstewart, line 19: setuptools 42+ required now

var/spack/repos/builtin/packages/snakemake/package.py
```diff
@@ -53,3 +54,27 @@ class Snakemake(PythonPackage):
...
+    with when("+s3"):
+        depends_on("py-boto3")
+        depends_on("py-botocore")
+
+    variant("http", default=False, description="Enable snakemake.remote.HTTP")
+    with when("+http"):
+        depends_on("py-requests")
+
+    with when("@7:"):
+        depends_on("python@3.7:")
```
adamjstewart:
```suggestion
        depends_on("python@3.7:", type=("build", "run"))
```
Or just remove this line since Spack only supports 3.7+

var/spack/repos/builtin/packages/snakemake/package.py
adamjstewart: filelock no longer needed

adamjstewart: ratelimiter no longer needed

var/spack/repos/builtin/packages/snakemake/package.py
```diff
@@ -53,3 +54,27 @@ class Snakemake(PythonPackage):
...
+        depends_on("py-throttler", type=("build", "run"))
+        depends_on("py-yte@1:1", type=("build", "run"))
```
adamjstewart:
```suggestion
        depends_on("py-yte@1", type=("build", "run"))
```

var/spack/repos/builtin/packages/snakemake/package.py
adamjstewart: jinja2@3 required now

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-throttler/package.py
charmoniumQ: Sounds good

var/spack/repos/builtin/packages/snakemake/package.py
```diff
@@ -53,3 +59,18 @@ class Snakemake(PythonPackage):
...
+    # needed to make certain parts of Snakemake work.
+    variant("ftp", default=False, description="Enable snakemake.remote.FTP")
+    depends_on("py-ftputil", when="+ftp")
+
+    variant("s3", default=False, description="Enable snakemake.remote.S3")
+    depends_on("py-boto3", when="+s3")
+    depends_on("py-botocore", when="+s3")
+
+    variant("http", default=False, description="Enable snakemake.remote.HTTP")
+    depends_on("py-requests", when="+http")
```
adamjstewart: These should probably be `type="run"` or `type=("build", "run")` so that they are available when you load the package.

adamjstewart (changes requested):

adamjstewart (dismissed):

adamjstewart:
@spackbot fix style

adamjstewart (approved):

PR #34045, opened 2022-11-21
libjpeg-turbo: build_system settings should be comprehensive

greenc-FNAL:
The covered version range for `build_system` should be comprehensive
rather than simply "known versions." Otherwise, externals declared in
`packages.yaml` with non-checksummed versions will cause perplexing
concretization errors.


greenc-FNAL:
Branch issues in PR creation.

PR #34051, opened 2022-11-22
Dyninst: add checksums for all supported versions

hainest:
This will confict with #34050, so I'll fix it once that has been merged.

@mwkrentel


alalazo:
#34050 has been merged!

hainest:
@alalazo Thanks! I've rebased here, so it's ready for review when you have time.

bernhardkaindl (approved):
Since it's only adding checksums, it's safe to approve and merge.

PR #34055, opened 2022-11-22
Use correct method in cpu-features recipe

iarspider:
Sorry!

### DIFF
--- a/var/spack/repos/builtin/packages/cpu-features/package.py
+++ b/var/spack/repos/builtin/packages/cpu-features/package.py
@@ -23,5 +23,5 @@ class CpuFeatures(CMakePackage):
 
     def cmake_args(self):
         args = ["-DBUILD_TESTING:BOOL=OFF"]
-        args += self.enable_or_disable("BUILD_SHARED_LIBS", variant="shared")
+        args.append(self.define_from_variant("BUILD_SHARED_LIBS", variant="shared"))
         return args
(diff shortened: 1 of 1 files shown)


bernhardkaindl (changes requested):
Thanks for the fixup, would you also fix this warning for 0.20? (the next version!)
```py
==> Warning: Using the old Spec.format method. This method was deprecated in Spack v0.15 and will be removed in Spack v0.20
```
Also we should fix this (update, I pushed a fix for this):
```py
Warning: Fetching from mirror without a checksum!
  This package is normally checked out from a version control system, but it has been archived on a spack mirror.  This means we cannot know a checksum for the tarball in advance. Be sure that your connection to this mirror is secure!
```
It's easy to fix, just use the path to the tag's archive e.g. like in `eckit` which is trivial.

bernhardkaindl:
@iarspider: Added a commit: [Use the release tarball for v0.7.0 to fix spack warning](https://github.com/spack/spack/pull/34055/commits/391eae01572b8f606a9a7b2541322d9229052c90)

Do you know the source of this warning? 
```py
Warning: Using the old Spec.format method. This method was deprecated in Spack v0.15 and will be removed in Spack v0.20
```

bernhardkaindl (approved):

bernhardkaindl:
Ok, this appears to be common to `CMakePackage`, has to be adressed there:
```py
Warning: Using the old Spec.format method. This method was deprecated in Spack v0.15 and will be removed in v0.20
```
Ready to go.

iarspider:
No idea

haampie:
develop was temporarily broken, see #34056

PR #34057, opened 2022-11-22
make/ninja: correct jobs

haampie:
`gmake` / `ninja` use their own `parallel` attribute instead of the dependent's
to determine the number of jobs, when used as a build dep.


haampie:
Can't use `module.make_jobs` because of the breaking change in #32340 :(

trws (approved):

haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

PR #34058, opened 2022-11-22
spack uninstall: follow run/link edges on --dependents

haampie:
`spack gc` removes build deps of explicitly installed specs, but somehow
if you take one of the specs that `spack gc` would remove, and feed it
to `spack uninstall /<hash>` by hash, it complains about all the
dependents that still rely on it.

This PR resolves the inconsistency by only following run/link type deps in
spack uninstall when looking for dependent specs.

That way you can finally do `spack uninstall cmake` without having to
remove all packages built with cmake.

Note: this PR is just a few lines of changes, but requires a bunch of tests to
be updated.

### DIFF
--- a/lib/spack/spack/cmd/uninstall.py
+++ b/lib/spack/spack/cmd/uninstall.py
@@ -133,7 +133,7 @@ def find_matching_specs(env, specs, allow_multiple_matches=False, force=False, o
     return specs_from_cli
 
 
-def installed_dependents(specs, env):
+def installed_runtime_dependents(specs, env):
     """Map each spec to a list of its installed dependents.
 
     Args:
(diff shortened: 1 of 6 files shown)


adamjstewart (approved):
Solves the issue I reported on Slack. Not confident enough in this part of Spack to pull the trigger and merge. Would be great to get reviews from others.

alalazo (approved):

alalazo:
Second thought: should we also add an option to restore the old behavior on demand? I.e. after this PR can I say something like:
```console
$ spack uninstall -aR cmake@X.Y
```
and uninstall everything that was build with `cmake@X.Y`?

adamjstewart:
I wouldn't describe that as the old behavior, I would describe that as a relatively recent bug that was introduced. Why would you want to do that?

alalazo:
> Why would you want to do that?

Clean an environment or a database. If I have a pletora of duplicates that differ because of build deps, is there a way to say "remove everything that was built using `foo@X.Y`"?

EDIT: to be clear, I agree this should be the default. I wonder though if we should add a flag like `spack uninstall --follow-build-deps ...` that maintains the current behavior upon user request.

adamjstewart:
> If I have a plethora of duplicates that differ because of build deps

I miss the old days where this sentence made no sense...

haampie:
I would say let's add a flag to follow build deps when people complain about it ;p

PR #34061, opened 2022-11-22
berkelygw: Use mpi variant to select parallel or serial support

bernhardkaindl:
This includes the changes to make ~mpi really use a non-mpi configuration by @glaweh from #339448 (which is approved by @migueldiascosta)

### DIFF
--- a/var/spack/repos/builtin/packages/berkeleygw/package.py
+++ b/var/spack/repos/builtin/packages/berkeleygw/package.py
@@ -35,31 +35,28 @@ class Berkeleygw(MakefilePackage):
         expand=False,
     )
 
-    variant("mpi", default=True, description="Builds with MPI support")
+    # For parallel computing support, enable +mpi. It uses MPI and ScaLAPACK
+    # which are inter-dependent in the berkeleygw code(they need each other):
+    # https://github.com/spack/spack/pull/33948#issuecomment-1323805817
+    variant("mpi", default=True, description="Build with MPI and ScaLAPACK support")
     variant("elpa", default=True, description="Build with ELPA support")
     variant("openmp", default=True, description="Build with OpenMP support")
-    variant("scalapack", default=True, description="Build with ScaLAPACK support")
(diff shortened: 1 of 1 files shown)


bernhardkaindl (commented):
Reviewed, looks good, @migueldiascosta, for you good too?

var/spack/repos/builtin/packages/berkeleygw/package.py
```diff
@@ -63,9 +62,11 @@ class Berkeleygw(MakefilePackage):
 
     depends_on("perl", type="test")
 
+    # scalapack and mpi variants need each other, make the solver (de)select the other too:
     conflicts(
         "+scalapack", when="~mpi", msg="scalapack is a parallel library and needs MPI support"
     )
+    conflicts("~scalapack", when="+mpi", msg="ScaLAPACK is required for MPI builds of berkelygw")
```
alalazo: Can you use a conditional variant? It works better when scoring possible solutions with clingo:
```python
variant("scalapack", default=True, when="+mpi", description="Build with ScaLAPACK support")
```

alalazo (changes requested):

var/spack/repos/builtin/packages/berkeleygw/package.py
bernhardkaindl: The `scalapack` and `mpi` variants need each other. I'm guessing it's better to just make it one variant then,
but it would mean that done of them would have to go.

alalazo: If `~scalapack+mpi` is not possible (sorry, I didn't look carefully enough!) then I'd remove `scalapack`

bernhardkaindl: I've applied your requested change! *(removed the `scalapack` variant)*

bernhardkaindl:
It was merged into #33948, closing

PR #34065, opened 2022-11-22
Prefix should depend on PrgEnv in Cray pkgs

vanderwb:
Currently, one cannot create a single external spec for `cray-mpich` and `cray-libsci` versions because Spack needs some way to disambiguate between Cray programming environments. Here is the [documented approach](https://spack.readthedocs.io/en/latest/getting_started.html#spack-on-cray):

```
  mpich:
    externals:
    - spec: "mpich@7.3.1%gcc@5.2.0 arch=cray_xc-haswell-CNL10"
      modules:
      - cray-mpich
    - spec: "mpich@7.3.1%intel@16.0.0.109 arch=cray_xc-haswell-CNL10"
      modules:
      - cray-mpich
```

However, all one needs to do on modern Cray systems is set the `PE_ENV` variable for Spack to obtain the correct prefix, and we can map this to a PrgEnv via an already-set dictionary. This PR adds logic to those two packages to find the correct prefix based on compiler (or provide a helpful error if the compiler is not supported), simplifying the external representation to:

```
  mpich:
    externals:
    - spec: "mpich@7.3.1 arch=cray_xc-haswell-CNL10"
      modules:
      - cray-mpich
```

Finally, I've also mapped OneAPI to the "INTEL" programming environment so that modern Intel releases are supported. (*as per HPE folks*)

### DIFF
--- a/var/spack/repos/builtin/packages/cray-libsci/package.py
+++ b/var/spack/repos/builtin/packages/cray-libsci/package.py
@@ -2,6 +2,9 @@
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+
 from spack.package import *
 from spack.util.module_cmd import get_path_args_from_module_line, module
 
@@ -37,6 +40,7 @@ class CrayLibsci(Package):
         "gcc": "GNU",
         "cce": "CRAY",
         "intel": "INTEL",
+        "oneapi": "INTEL",
         "clang": "ALLINEA",
         "aocc": "AOCC",
         "nvhpc": "NVIDIA",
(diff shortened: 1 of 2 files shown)


becker33:
@vanderwb have you played at all with the path-based support for cray-mpich in the newer releases of the cray PE? I'm asking because we're trying to move away from using the PrgEnv modules at all, for a lot of the same reasons that you highlight here.

vanderwb:
@becker33 Thanks for the reply. Do you mean defining the external using a prefix instead of a module or something else? I haven't seen it documented (at least in the readthedocs), but I'd be happy to try it out if there are instructions somewhere.

vanderwb:
@becker33 - can we discuss the intention here a bit further? When you mention using prefixes, are you suggesting something like this:

```
    cray-mpich:
      externals:
      - spec: cray-mpich@8.1.21 %gcc
        prefix: /opt/cray/pe/mpich/8.1.21/ofi/gnu/9.1
      - spec: cray-mpich@8.1.21 %cce
        prefix: /opt/cray/pe/mpich/8.1.21/ofi/cray/10.0
      - spec: cray-mpich@8.1.21 %nvhpc
        prefix: /opt/cray/pe/mpich/8.1.21/ofi/nvidia/20.7
      - spec: cray-mpich@8.1.21 %oneapi
        prefix: /opt/cray/pe/mpich/8.1.21/ofi/intel/19.0
      - spec: cray-mpich@8.1.21 %intel
        prefix: /opt/cray/pe/mpich/8.1.21/ofi/intel/19.0
```

The above is what we are doing now (plus loading the libfabric module for each one) instead of my PR, which attempted to reduce things down to:
```
      - spec: cray-mpich@8.1.21
        modules:
        - cray-mpich/8.1.21
        - craype-network-ofi
        - libfabric/1.15.0.0
```

The first version is quite verbose and laborious to maintain, but if there is good reason for it, then so be it. But I'd like to understand whether I am doing it the way you intend.

What I would really like is some logic in the `cray-mpich` package to eliminate the need to load the network and/or libfabric modules, as this makes using our Spack install as an upstream dependent on the downstream user's module state (i.e. are the required modules available in the environment...). `libfabric` is a dependency of this package, but since it has to be specified as an external, dependencies are irrelevant. Given that, I don't really know what the intention is to make this work without module loads. The only workaround I see is to append Cray's libfabric to LD_LIBRARY_PATH for each compiler, but this is also laborious and probably user-error prone. Am I missing something here?

vanderwb:
Figure I should just close this, since by now it's very out of sync with develop.

PR #34066, opened 2022-11-22
Pastix: new version 6.2.2

fpruvost:


ramet (dismissed):

alalazo (commented):
Can we keep previous versions? It's unusual to delete the latest version of a package on an update in Spack.

var/spack/repos/builtin/packages/pastix/package.py
```diff
@@ -12,12 +12,12 @@ class Pastix(CMakePackage, CudaPackage):
     based on direct methods"""
 
     homepage = "https://gitlab.inria.fr/solverstack/pastix/blob/master/README.md"
-    url = "https://gitlab.inria.fr/solverstack/pastix/uploads/baa57033d98378e0f3affbf45900fb6e/pastix-6.2.1.tar.gz"
+    url = "https://gitlab.inria.fr/solverstack/pastix/uploads/d579e932b415103974ced5d216481332/pastix-6.2.2.tar.gz"
```
mfaverge: To answer the comment from @alalazo, if we manage multiple versions, the url should be constructed with:

"https://files.inria.fr/pastix/releases/v6/pastix-" + version + ".tar.gz" instead of using the gitlab one.

alalazo: That would be perfect, but you can also stick a specific `url=...` argument in the `version` directive.

ramet:
I don't really understand how to handle previous releases. This way, you keep hash for the previous release, but the url is fixed.
It seems to be the same for others spack packages, but, in homebrew packaging for instance, there is a way to build the url depending on the version specified...

fpruvost:
Hi @ramet it is handled this way by spack, I guess it analyses the url and deduce the pattern to use. I have tested the current package it works with `spack install pastix@6.2.1`

ramet (approved):

mfaverge (approved):

bernhardkaindl (approved):
Reviewed by alalazo and after changes approved by @ramet and @mfaverge, and builds.

PR #34074, opened 2022-11-23
perl: Detect ~threads when Thread/Queue.pm cannot be used

bernhardkaindl:
Motivation:

1. `perl-Thread-Queue` may not always be pre-installed on new RedHat-based hosts by default.
2. This, in combination with `spack external find perl` breaks the build of `automake` (which uses `Threads::Queue`)

One part of the fix is #34076, The other part of the bug is that `'spack external find perl'` only checks for `-Dusethreads`, but not if `'perl -e "require Thread::Queue"'` really works.

Because Red Hat perl is always built with `-Dusethreads`, it is currently always detected as: `perl+threads`, even when `'perl -e "require Thread::Queue"'` is not possible (when the rpm `perl-Thread-Queue` is not installed)

The fix is straight forward:

Fix perl's `determine_variants()` to call `'perl -e "require Thread::Queue"'` when `-Dusethreads` was found and only add `+threads` when it succeeds.


bernhardkaindl:
@spackbot run pipeline

bernhardkaindl:
@spackbot run pipeline

sethrj (dismissed):
LGTM

sethrj:
Not sure what's up with the ci though. @spackbot run pipeline

bernhardkaindl:
@becker33 Thanks for your review in Chat! Your suggestion (with your attribution) is now pushed!

He saw that `if perl(...)` is checking whether the perl executable wrote anything to stderr
and was little worried we’ll get false negatives if perl emits a warning when it’s run.
Therefore he suggested the new code for checking perl.returncode:
```py
perl("-e", "require Thread::Queue", error=str, ignore_errors=2)
if perl.returncode == 2:
    return "~threads"
elif perl.returncode == 0:
    return "+threads"
else:
    raise ...
```
I inserted and tested the updated code (for the true and false case) as well as tested the formatting of the string for the exception if the return code is not as expected.

It is ready for review & merge.

sethrj:
I still don't know what's wrong with the CI. Since perl is a dependency of basically *everything* I think each build has a pretty good chance of failing because of the number of packages.

Why would this be causing build errors in hiprand/rocrand? https://gitlab.spack.io/spack/spack/-/jobs/5006434

becker33 (approved):

becker33:
@sethrj yeah I'm pretty convinced that the errors on this are spurious, not sure where they came from.

alalazo:
@spackbot run pipeline

haampie:
Hi @bernhardkaindl, this package-related pull request can be migrated from `spack/spack` to `spack/spack-packages` using the [**migrate-pkg-prs utility**](https://github.com/spack/migrate-package-prs).

---

PR #34076, opened 2022-11-23
automake: depend_on "perl+threads" (it requires Thread::Queue)

bernhardkaindl:
Motivation:

1. `perl-Thread-Queue` may not always be pre-installed on new RedHat-based hosts by default.
2. This, in combination with `spack external find perl` breaks the build of `automake` (which uses `Threads::Queue`)

So lets fix this.

The fix is pretty simple, tested and also obvious to review:
```diff
-    depends_on("perl", type=("build", "run"))
+    depends_on("perl+threads", type=("build", "run"))
```

Full Background story:

To completely and really fix this, #34074 is also needed (see it for details on that)


alalazo:
@spackbot run pipeline

bernhardkaindl (commented):
@alalazo Ready for review & merge

becker33 (approved):

PR #34086, opened 2022-11-23
armpl-gcc: add post-installation check by running examples

annop-w:
Check installation by running examples and compare results
with expected ones. The makefile is included as part of the
package.

### DIFF
--- a/var/spack/repos/builtin/packages/armpl-gcc/package.py
+++ b/var/spack/repos/builtin/packages/armpl-gcc/package.py
@@ -201,23 +201,25 @@ class ArmplGcc(Package):
         exe("--accept", "--force", "--install-to", prefix)
 
     @property
+    def lib_suffix(self):
+        suffix = ""
+        suffix += "_ilp64" if self.spec.satisfies("+ilp64") else ""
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/armpl-gcc/package.py
```diff
@@ -200,24 +200,33 @@ def install(self, spec, prefix):
...
+    def is_shared(self):
+        return self.spec.satisfies("+shared")
+
+    @property
+    def is_ilp64(self):
+        return self.spec.satisfies("+ilp64")
+
+    @property
+    def is_omp(self):
+        return self.spec.satisfies("threads=openmp")
```
alalazo: Instead of adding these functions, that are just shorthands to `self.spec.satisfies` with a specific argument, did you consider adding a `libname` and a `suffix` property to the package?

alalazo (changes requested):

var/spack/repos/builtin/packages/armpl-gcc/package.py
annop-w: Thank you for the idea. Done.

alalazo (approved):

PR #34087, opened 2022-11-23
Revert "gitlab: Add shared PR mirror to places pipelines look for binaries."

haampie:
Reverts spack/spack#33746

This does not work:

```
Computed hash (9ad8f218508333d9256c6927a0d3c932b4e0f10b7d84d0530ad709a94f952865) did not match remote (None), indicating error in index transmission
```

Which causes every single `spack install` in CI to fetch the mirror index over and over.

Can the code quality regarding mirrors / binary caches _please_ be improved. It's an absolute disaster. Takes lots of time to troubleshoot. Error handling in particular is really bad. Please stop writing code that requires me to run everything with `spack -d` and add even more print statements.

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -49,7 +49,6 @@ JOB_RETRY_CONDITIONS = [
 
 TEMP_STORAGE_MIRROR_NAME = "ci_temporary_mirror"
 SPACK_RESERVED_TAGS = ["public", "protected", "notary"]
-SHARED_PR_MIRROR_URL = "s3://spack-binaries-prs/shared_pr_mirror"
 
 spack_gpg = spack.main.SpackCommand("gpg")
 spack_compiler = spack.main.SpackCommand("compiler")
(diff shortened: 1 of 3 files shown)


haampie:
Also: why are you not using https://en.wikipedia.org/wiki/HTTP_ETag. It's also supported by S3.

The race conditions with two files `index.json` and `index.json.hash` couldn't be more obvious :weary:.

Also, _why_ are we even fetching a full index in `spack install` on a concrete environment? All the specs are concrete, you know where to grab the relevant files. It's not like fetching an index helps in any way? Between the moment you've fetched the index and make the request the file may have been deleted. The whole point of Spack's model is that the tarballs are "content-addressable" (well, addressable by hash), so we should just fetch by pre-computed URL. There's no point to download an index, unless you do reuse concretization?

haampie:
@spackbot run pipeline

tgamblin (approved):

scottwittenburg:
> The race conditions with two files index.json and index.json.hash couldn't be more obvious

Sorry, it's not obvious to me yet.  Is it because now dependencies are installed in parallel?  I don't understand why reverting this PR helps, i.e. why did just adding a new mirror to check cause a problem we didn't already face with the other mirrors we check?

> Also, why are we even fetching a full index in spack install on a concrete environment?

I don't think we ever did that until #32137.

haampie:
The race is that we index.json / index.json.hash are not updated atomically, or failure to upload index.json.hash does not revert the change to index.json. Also, downloading it is not atomic (index.json.hash and index.json are fetched at different times, not together). I don't know how the issue was caused that let me to revert this PR, but basically index.json and index.json.hash were out of sync, causing a new download every time; the error for that is ignored/hidden. I've also opened https://github.com/spack/spack/pull/34152 about that.

Thanks for the pointer, I commented on https://github.com/spack/spack/pull/32137#issuecomment-1331101914.

I've opened an issue regarding races/cache invalidation here: https://github.com/spack/spack/issues/34155

PR #34088, opened 2022-11-23
[openfoam] Apply patch adding missing include file

stephenmsachs:
Needs to be added in order to compile `openfoam@2012_220610 %intel`


bernhardkaindl (approved):
PR is safe: Adds `#include <array>` to a single file of a single version.

olesenm:
@stephenmsachs - that's unfortunate. I thought that it was already patched in that version, but I see that it was applied after that.

PR #34093, opened 2022-11-23
spack install: fix permissions of intermediary module directories

alexhornburg-xlnx:
As part of installing a spack package, a module file is also created. For TCL modules this is typically under: $spack/share/spack/modules/$arch/ This path may not exist before the package has been installed if it is the first time a package is installed for this $arch. In this case spack helpfully creates the necessary directory. While the module file itself correctly inherits the permissions of the package, the intermediary directories (.../modules/$arch/...) are created using the OS default permissions. On Linux this means using the users current  umask.

This is an issue in use-cases where the spack instance is shared among many users and machines _and_ the umask of these users is restricted such as disallowing writes for the group (g-w). This means that user A may install a first package on system X and user B will receive a permission denied error when trying to install a second package on system X since they do not have permission to add a file to the intermediary directory.

To workaround this users must currently be aware of whether it is the first time they are installing any package on a given machine and manually ensure that the permissions of those intermediary folders are suitable for other users.

The package install directory already solves the same issue by inheriting the permissions of the parent directory and applying it to the intermediary directories. This patch follows the same approach.

### DIFF
--- a/lib/spack/spack/modules/common.py
+++ b/lib/spack/spack/modules/common.py
@@ -863,8 +863,10 @@ class BaseModuleFileWriter(object):
         # If the directory where the module should reside does not exist
         # create it
         module_dir = os.path.dirname(self.layout.filename)
+        parent_dir = os.path.dirname(self.layout.dirname())
+        perms = os.stat(parent_dir).st_mode
         if not os.path.exists(module_dir):
-            llnl.util.filesystem.mkdirp(module_dir)
+            llnl.util.filesystem.mkdirp(module_dir, mode=perms)
 
         # Get the template for the module
         template_name = self._get_template()


alexhornburg-xlnx:
@becker33 I noticed your name in commits for the affected code and in commits solving the same issue for `create_install_directory()`. Would you be able to review this PR or maybe point to someone who would?

Regarding the test failures, I do not think they are related to this patch. I cannot reproduce the style failure locally and I did not modify any lines which make use of `Optional`. Also, the CI style checks passed when I first posted this PR last week. It started failing once I rebased on top of latest commits in develop.
The CI failure I saw last week was with MacOS and I believe that other PRs which also touch core spack files failed in the same manner.

fgava90:
Same thing is happening for environments directories.

The environment folder is created with pathlib.Path().mkdir() which by default uses mode=0o777, anything inside inherits the same permissions.
Files are symlinked/hardlinked/copied from the install_tree which have the correct permissions, but this doesn't stop users from adding stuff into environments (a user running pip install will add stuff into a python environment for example).
It would be great if environments would also respect the packages:all:permissions config, or if they could be given a custom config

PR #34056, opened 2022-11-22
Revert "warn about removal of deprecated format strings"

alalazo:
Reverts spack/spack#33829

Some update was missed, which causes:
Screenshot from 2022-11-22 12-29-55

PR #34101, opened 2022-11-23
warn about removal of deprecated format strings

becker33:
Fixes bug in [#33829](https://github.com/spack/spack/pull/33829) which was reverted

@alalazo

### DIFF
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -4420,7 +4420,7 @@ class Spec(object):
         TODO: allow, e.g., ``$6#`` to customize short hash length
         TODO: allow, e.g., ``$//`` for full hash.
         """
-        tty.warn(
+        warnings.warn(
             "Using the old Spec.format method."
             " This method was deprecated in Spack v0.15 and will be removed in Spack v0.20"
         )
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -4418,8 +4418,10 @@ class Spec(object):
         TODO: allow, e.g., ``$6#`` to customize short hash length
         TODO: allow, e.g., ``$//`` for full hash.
         """
-        tty.warn("Using the old Spec.format method."
-                 " This method was deprecated in Spack v0.15 and will be removed in Spack v0.20")
+        tty.warn(
+            "Using the old Spec.format method."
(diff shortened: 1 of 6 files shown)


becker33:
@spackbot fix style

alalazo:
Can we add a unit test to check that canonicalization of the default config brings in all the components we expect for the install tree (and any other path that might be significant)? This is to avoid https://github.com/spack/spack/pull/34056#issuecomment-1323533176 in the future.

becker33:
Added a test for the default install tree

lib/spack/spack/spec.py
```diff
@@ -4418,7 +4420,10 @@ def old_format(self, format_string="$_$@$%@+$+$=", **kwargs):
         TODO: allow, e.g., ``$6#`` to customize short hash length
         TODO: allow, e.g., ``$//`` for full hash.
         """
-
+        tty.warn(
```
alalazo: This:
```suggestion
        warnings.warn(
```
should help a little with reducing the clutter if the user has a deprecated format (I think it emits a warning for each caller to this function, but need to double check). The net result is this:
Screenshot from 2022-11-24 17-41-47
instead of this:
Screenshot from 2022-11-24 17-42-41

if you temporarily revert the changes to the default `config.yaml` to check the warnings.

lib/spack/spack/spec.py
```diff
@@ -4418,7 +4420,10 @@ def old_format(self, format_string="$_$@$%@+$+$=", **kwargs):
...
+            "Using the old Spec.format method."
+            " This method was deprecated in Spack v0.15 and will be removed in Spack v0.20"
```
alalazo, line 4425: I would also like to suggest an easy way to point the users to where the offending format was read into Spack, but I can't come up with anything. The best I could do myself is to run with:
```console
python -W error bin/spack -d install zlib
```
after I modified the code to use `warnings`. This get us a stacktrace, but is certainly not very "ergonomic".

Maybe we can advise users to check their configurations and point somewhere in the docs where we show what to use instead?

alalazo: Link to the docs (took some time to find): https://spack.readthedocs.io/en/latest/spack.html#spack.spec.Spec.format

alalazo (approved):

PR #34105, opened 2022-11-23
 `spack python`: use `Executable` instead of `InteractiveConsole`

tgamblin:
Running a `spack-python` script like this:

```python

import spack
import multiprocessing

def echo(args):
    print(args)

if __name__ == "__main__":
    pool = multiprocessing.Pool(2)
    pool.map(echo, range(10))
```

will fail in `develop` with an error like this:

```console
_pickle.PicklingError: Can't pickle <function echo at 0x104865820>: attribute lookup echo on __main__ failed
```

Python expects to be able to look up the method `echo` in `sys.path["__main__"]` in
subprocesses spawned by `multiprocessing`, but because we use `InteractiveConsole` to
run `spack python`, the executed file isn't considered to be the `__main__` module, and
lookups in subprocesses fail. We tried to fake this by setting `__name__` to `__main__`
in the `spack python` command, but that doesn't fix the fact that no `__main__` module
exists.

Another annoyance with `InteractiveConsole` is that `__file__` is not defined in the
main script scope, so you can't use it in your scripts.

We can fix this by using `spack.util.executable.Executable` to launch `spack python`
instead of using `InteractiveConsole`. This makes `spack python` behave more like
normal python. After this change:

```console
> ./test.py
0
1
2
3
4
5
6
7
8
9
```

- [x] Use `Executable` to launch non-interactive `spack python` invocations
- [x] Only use `InteractiveConsole` for interactive `spack python`

### DIFF
--- a/lib/spack/spack/__init__.py
+++ b/lib/spack/spack/__init__.py
@@ -39,6 +39,7 @@ class LazyRepoImporter:
 
     def find_spec(self, fullname, python_path, target=None):
         if fullname.startswith(ROOT_PYTHON_NAMESPACE):
+            sys.meta_path.remove(self)  # remove so we don't do this twice
             import spack.repo  # noqa: F401
         return None
 
--- a/lib/spack/spack/__init__.py
+++ b/lib/spack/spack/__init__.py
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 #: PEP440 canonical <major>.<minor>.<micro>.<devN> string
+
+import sys
+
 __version__ = "0.22.0.dev0"
 spack_version = __version__
 
@@ -17,6 +20,35 @@ def __try_int(v):
 
 #: (major, minor, micro, dev release) tuple
 spack_version_info = tuple([__try_int(v) for v in __version__.split(".")])
+
+
+#: Package modules are imported as spack.pkg.<repo-namespace>.<pkg-name>
+#: This is a special namespace because it's generated dynamically from package repos.
+ROOT_PYTHON_NAMESPACE = "spack.pkg"
+
+
+# To import `spack.pkg`, `spack.repo` *must* be imported first. This class lazily
+# imports `spack.repo` (in case it hasn't been imported already) when `spack.pkg` is
+# imported, allowing `import spack.pkg...` to be the first thing you import.
+class LazyRepoImporter:
+    """Fake loader for ``sys.meta_path``.
+
+    Ensures ``spack.repo`` is always imported before ``spack.pkg``.
+
+    """
+
+    def find_spec(self, fullname, python_path, target=None):
+        if fullname.startswith(ROOT_PYTHON_NAMESPACE):
+            import spack.repo  # noqa: F401
+        return None
+
+    def compute_loader(self, fullname):
+        return None
+
+
+# ensure our LazyRepoImporter is the first thing on the PATH. We could just import
+# `spack.repo` here, but this allows us to only load it when needed.
+sys.meta_path.insert(0, LazyRepoImporter())
 
 
 __all__ = ["spack_version_info", "spack_version"]
--- a/lib/spack/spack/cmd/__init__.py
+++ b/lib/spack/spack/cmd/__init__.py
@@ -17,7 +17,6 @@ from llnl.util.tty.colify import colify
 from llnl.util.tty.color import colorize
 
 import spack.config
-import spack.environment as ev
 import spack.error
 import spack.extensions
 import spack.parser
@@ -184,6 +183,8 @@ def matching_spec_from_env(spec):
     If no matching spec is found in the environment (or if no environment is
     active), this will return the given spec but concretized.
     """
+    import spack.environment as ev
+
     env = ev.active_environment()
     if env:
         return env.matching_spec(spec) or spec.concretized()
@@ -526,6 +527,8 @@ def require_active_env(cmd_name):
     Returns:
         (spack.environment.Environment): the active environment
     """
+    import spack.environment as ev
+
     env = ev.active_environment()
 
     if env:
@@ -555,6 +558,7 @@ def find_environment(args):
     Returns:
         (spack.environment.Environment): a found environment, or ``None``
     """
+    import spack.environment as ev
 
     # treat env as a name
     env = args.env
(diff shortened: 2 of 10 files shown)


trws (dismissed):

lib/spack/spack/__init__.py
```diff
@@ -19,4 +22,34 @@ def __try_int(v):
...
+    """Fake loader for ``sys.meta_path``.
+
+    Ensures ``spack.repo`` is always imported before ``spack.pkg``.
+
+    """
+
+    def find_spec(self, fullname, python_path, target=None):
+        if fullname.startswith(ROOT_PYTHON_NAMESPACE):
+            sys.meta_path.remove(self)  # remove so we don't do this twice
+            import spack.repo  # noqa: F401
```
alalazo, line 43: Instead of removing the meta-path finder, can we add a boolean state to it and record it was triggered?

lib/spack/spack/cmd/__init__.py
```diff
@@ -240,6 +238,8 @@ def matching_spec_from_env(spec):
     If no matching spec is found in the environment (or if no environment is
     active), this will return the given spec but concretized.
     """
+    import spack.environment as ev
```
alalazo, line 185: This function can probably be moved somewhere under `spack.environment`, and then be imported where needed.

lib/spack/spack/cmd/__init__.py
```diff
@@ -503,6 +503,8 @@ def format_list(specs):
 def filter_loaded_specs(specs):
     """Filter a list of specs returning only those that are
     currently loaded."""
+    import spack.user_environment as uenv
```
alalazo, line 450: This one instead can likely be moved to `spack.user_environment`

lib/spack/spack/cmd/__init__.py
```diff
@@ -584,6 +586,8 @@ def require_active_env(cmd_name):
     Returns:
         (spack.environment.Environment): the active environment
     """
+    import spack.environment as ev
```
alalazo, line 531: This one is a bit trickier because of the need to know the originating command :thinking:

lib/spack/spack/cmd/__init__.py
```diff
@@ -613,6 +617,7 @@ def find_environment(args):
     Returns:
         (spack.environment.Environment): a found environment, or ``None``
     """
+    import spack.environment as ev
```
alalazo, line 562: This function has implicit assumptions on the `Argparse` namespace passed in as argument (it must have an `env` attribute and optionally an `end_dir` attribute).

lib/spack/spack/main.py
```diff
@@ -587,6 +586,8 @@ def setup_main_options(args):
         SHOW_BACKTRACE = True
 
     if args.debug:
+        import spack.util.debug
```
alalazo: We need this:
```diff
diff --git a/lib/spack/spack/main.py b/lib/spack/spack/main.py
index 654967d34a..fb014e3527 100644
--- a/lib/spack/spack/main.py
+++ b/lib/spack/spack/main.py
@@ -569,6 +569,8 @@ def send_warning_to_tty(message, *args):
 
 def setup_main_options(args):
     """Configure spack globals based on the basic options."""
+    import spack.util.debug
+
     # Assign a custom function to show warnings
     warnings.showwarning = send_warning_to_tty
 
@@ -586,8 +588,6 @@ def setup_main_options(args):
         SHOW_BACKTRACE = True
 
     if args.debug:
-        import spack.util.debug
-
         spack.util.debug.register_interrupt_handler()
         spack.config.set("config:debug", True, scope="command_line")
         spack.util.environment.TRACING_ENABLED = True

```
to avoid a "reference before assignment" at line 584.

lib/spack/spack/repo.py
```diff
@@ -58,7 +56,8 @@ def python_package_for_repo(namespace):
     Args:
         namespace (str): repo namespace
     """
-    return "{0}.{1}".format(ROOT_PYTHON_NAMESPACE, namespace)
+    #    return "{0}.{1}".format(spack.ROOT_PYTHON_NAMESPACE, namespace)
```
alalazo, line 61:
```suggestion
```

lib/spack/spack_installable/main.py
```diff
@@ -7,19 +7,24 @@
...
 def main(argv=None):
     # Find spack's location and its prefix.
     this_file = os.path.realpath(os.path.expanduser(__file__))
     spack_prefix = dn(dn(dn(dn(this_file))))
 
-    # Allow spack libs to be imported in our scripts
-    spack_lib_path = os.path.join(spack_prefix, "lib", "spack")
-    sys.path.insert(0, spack_lib_path)
+    # Add all the sys paths that allow spack libs to be imported
+    sys.path[:0] = get_spack_sys_paths(spack_prefix)
```
alalazo, line 26: :+1: This is definitely clearer

lib/spack/spack/cmd/python.py
```diff
@@ -6,52 +6,54 @@
...
 import sys
 
+import spack_installable.main as sim
+
 import llnl.util.tty as tty
 
 import spack
+import spack.paths
+import spack.util.environment as sue
+import spack.util.executable as exe
```
alalazo, line 18: Maybe we can do something like:
```suggestion
from spack.util import environment, executable
```
?

lib/spack/spack/cmd/python.py
```diff
@@ -66,103 +68,76 @@ def python(parser, args, unknown_args):
...
+    env = os.environ.copy()
+    mods.apply_modifications(env)
+
+    print_spack_version_if_interactive(args)
+
+    # note: this used to use code.InteractiveConsole, but exec'ing python and not using
+    # interactive mode ensures that things like the __main__ module will be defined and
+    # that multiprocessing will work.
+    python_args = construct_python_args(args, unknown_args)
+    python = exe.Executable(sys.executable)
```
alalazo, line 140: :+1:

alalazo (changes requested):
Basically LGTM, once tests are passing. I left a few minor comments and a diff for an import typo that I am pretty sure causes issues when running in debug mode.

alalazo:
There's some circular import error to be fixed (I wish Python was like Go in this respect :slightly_smiling_face: )

tgamblin:
Not sure the timeline on this because it’s *many* such errors (see history). I was just rebasing but haven’t had a chance to dive into it. I am curious whether #40262 will help.

tgamblin:
Superseded by #41789

PR #34106, opened 2022-11-23
Drop parallel install in CI again

haampie:
The downside is that multiple processes fetch all the `index.json` files, compared to 1 spack install process.

Also now that relocation is faster, I wonder what the difference really is.


haampie:
@spackbot rebuild everything

haampie:
Closing since ci runs OK again

PR #34107, opened 2022-11-23
PMIx: enable python bindings

### LINKED ISSUE #17083: Enable Python bindings in PMIx
We are seeing an uptick in the number of people wanting to use PMIx in Python libraries and applications. I'm not sure of your distribution policies, but it might be worth considering enabling the Python bindings in the PMIx package.

Building the bindings requires Cython (in the standard locations for the Python being used) plus addition of the `--enable-python-bindings` configure option. It will generate an "egg" that is installed in the prefix location - this would have to be added to the rpm or whatever method you choose for distributing it.

s-sajid-ali:
Fixes https://github.com/spack/spack/issues/17083

### DIFF
--- a/var/spack/repos/builtin/packages/pmix/package.py
+++ b/var/spack/repos/builtin/packages/pmix/package.py
@@ -69,10 +69,7 @@ class Pmix(AutotoolsPackage):
         description="allow a PMIx server to request services from " "a system-level REST server",
     )
 
-    variant('python',
-            default=False,
-            when="@4.1.2:",
-            description="Enable python bindigs")
+    variant("python", default=False, when="@4.1.2:", description="Enable python bindigs")
 
     variant("docs", default=False, description="Build manpages")
 
@@ -88,8 +85,8 @@ class Pmix(AutotoolsPackage):
     depends_on("hwloc@1.11:1,2:", when="@3:")
     depends_on("curl", when="+restful")
     depends_on("jansson@2.11:", when="+restful")
-    depends_on('python', when="+python")
-    depends_on('py-cython', when="+python")
+    depends_on("python", when="+python")
+    depends_on("py-cython", when="+python")
 
     def autoreconf(self, spec, prefix):
         """Only needed when building from git checkout"""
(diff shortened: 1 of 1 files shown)


rhc54 (dismissed):
I'm no spack expert, but it looks correct to me. Note that PMIx does require that it be Python 3.4 or above - don't know if that should be in your rules?

s-sajid-ali:
>Note that PMIx does require that it be Python 3.4 or above - don't know if that should be in your rules?

Python versions that have reached EOL have been dropped from spack recently : https://github.com/spack/spack/pull/33898. 

I'm guessing this means that spack is guaranteed to install at least python3.7, which should be enough for this package. I'll wait for someone with more knowledge of the aforementioned PR to confirm this.

rhc54:
> > Note that PMIx does require that it be Python 3.4 or above - don't know if that should be in your rules?
> 
> Python versions that have reached EOL have been dropped from spack recently : #33898.
> 
> I'm guessing this means that spack is guaranteed to install at least python3.7, which should be enough for this package. I'll wait for someone with more knowledge of the aforementioned PR to confirm this.

Sounds good! Thanks

rhc54:
@alalazo It appears that the earlier problem I mentioned was simply an issue with the configure code in the tarball. @s-sajid-ali resolved it by upgrading to a newer version. Could you please take a look and see what you think of the integration?

alalazo:
@spackbot fix style

rhc54:
Looks great to me! Thanks to both of you for doing this!

rhc54:
I'm afraid I don't really know much about Spack's internals - the failed CI seems to be something about Spack infrastructure and not the PMIx library. I didn't see anything obvious that I could help address, I'm afraid.

alalazo (approved):

vsoch:
This broke flux-pmix, because now pmix -> python brings in a need for a fortran compiler, which spack doesn't find. So unless you write your own spack.yaml with the path, this fails. This means that current develop `spack install flux-pmix` does not work. See https://github.com/spack/spack/issues/34256

rhc54:
Not sure I understand, though perhaps someone more knowledgeable about spack would see the connection - why would python force the need for a fortran compiler?

PR #34109, opened 2022-11-23
py-cartopy: older versions don't support newer matplotlib

adamjstewart:
https://github.com/SciTools/cartopy/issues/2086


bernhardkaindl (approved):
Tested by building 0.16, 0.17, 0.18 and 0.19

PR #34112, opened 2022-11-23
Add packages related to the LWA software stack

downloadico:
Adding a few packages related to the LWA software stack

### DIFF
--- a/var/spack/repos/builtin/packages/psrcat/package.py
+++ b/var/spack/repos/builtin/packages/psrcat/package.py
@@ -8,7 +8,7 @@ from spack.package import *
 
 class Psrcat(MakefilePackage):
     """ATNF Pulsar Catalogue
-    A catalogue of pulsars provided by the Australian Telescope National 
+    A catalogue of pulsars provided by the Australian Telescope National
     Facility"""
 
     homepage = "https://www.atnf.csiro.au/research/pulsar/psrcat/"
--- a/var/spack/repos/builtin/packages/psrcat/package.py
+++ b/var/spack/repos/builtin/packages/psrcat/package.py
@@ -1,0 +1,24 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Psrcat(MakefilePackage):
+    """A proper description of your package."""
+
+    homepage = "https://www.atnf.csiro.au/research/pulsar/psrcat/"
+    url = "https://www.atnf.csiro.au/research/pulsar/psrcat/downloads/psrcat_pkg.v1.68.tar.gz"
+
+    version("1.68", sha256="fbe4710c9122e4f93dbca54cf42cc2906f948f76885b241d1da2f8caecfbc657")
+
+    def build(self, spec, prefix):
+        makeit = which("./makeit")
+        makeit()
+
+    def install(self, spec, prefix):
+        bindir = prefix + "/bin/"
+        mkdir(bindir)
+        copy("psrcat", bindir)
--- a/var/spack/repos/builtin/packages/psrcat/package.py
+++ b/var/spack/repos/builtin/packages/psrcat/package.py
@@ -19,6 +19,5 @@ class Psrcat(MakefilePackage):
         makeit()
 
     def install(self, spec, prefix):
-        bindir = prefix + "/bin/"
-        mkdir(bindir)
-        copy("psrcat", bindir)
+        mkdirp(prefix.bin)
+        install("psrcat", bindir)
--- a/var/spack/repos/builtin/packages/psrcat/package.py
+++ b/var/spack/repos/builtin/packages/psrcat/package.py
@@ -7,7 +7,9 @@ from spack.package import *
 
 
 class Psrcat(MakefilePackage):
-    """A proper description of your package."""
+    """ATNF Pulsar Catalogue
+    A catalogue of pulsars provided by the Australian Telescope National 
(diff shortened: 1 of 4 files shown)


var/spack/repos/builtin/packages/psrcat/package.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Psrcat(MakefilePackage):
+    """A proper description of your package."""
```
becker33: Looks like you missed a fixup here

var/spack/repos/builtin/packages/psrcat/package.py
```diff
@@ -0,0 +1,24 @@
...
+    version("1.68", sha256="fbe4710c9122e4f93dbca54cf42cc2906f948f76885b241d1da2f8caecfbc657")
+
+    def build(self, spec, prefix):
+        makeit = which("./makeit")
+        makeit()
+
+    def install(self, spec, prefix):
+        bindir = prefix + "/bin/"
+        mkdir(bindir)
+        copy("psrcat", bindir)
```
becker33:
```suggestion
        mkdirp(prefix.bin)
        install("psrcat", bindir)
```
Prefix has convenience attributes for subdirectories, and the `install` method manages installation permissions in addition to copying the files.

var/spack/repos/builtin/packages/psrchive/package.py
```diff
@@ -0,0 +1,82 @@
...
+    git = "https://git.code.sf.net/p/psrchive/code.git"
+
+    version(
+        "2022-05-14", sha256="4d25609837cba1be244fa8adc8f105afe31972f2650bc0b90438862cf35395e1"
+    )
+
+    # version specified in
+    # https://github.com/lwa-project/pulsar/blob/master/SoftwareStack.md
+    # as of Nov 23 2022
+    version("10-17-2020", commit="ca12b4a279f3d4adcca223508116d9d270df8cc6")
```
becker33: Can you unify the date format across these? I think YYYY-MM-DD is preferable unless there's a reason to do otherwise, so the sorting works properly

var/spack/repos/builtin/packages/psrchive/package.py
```diff
@@ -0,0 +1,82 @@
...
+    depends_on("libtool", type="build")
+    depends_on("psrcat")
+    depends_on("epsic")
+    depends_on("tempo")
+    depends_on("fftw@3:")
+    depends_on("python")
+    depends_on("pgplot")
+
+    depends_on("mpi", when="+mpi")
+    depends_on("intel-mkl", when="+mkl")
```
becker33:
```suggestion
    depends_on("mkl", when="+mkl")
```

There are multiple providers for `mkl` among the different intel packages/bundles, this way it will allow Spack to use any of them to satisfy the dependency.

var/spack/repos/builtin/packages/psrchive/package.py
```diff
@@ -0,0 +1,82 @@
...
+    depends_on("qt", when="+qt")
+
+    def configure_args(self):
+        spec = self.spec
+        args = ["--enable-shared"]
+        args.append("--with-python_prefix={0}".format(spec["python"].prefix))
+        args.append("--with-epsic-dir={}".format(spec["epsic"].prefix))
+        args.append("--with-epsic-include-dir={}".format(spec["epsic"].prefix.include))
+        args.append("--with-epsic-lib-dir={}".format(spec["epsic"].prefix.lib))
+        args.append("--with-psrcat={0}".format(spec["psrcat"].prefix + "bin/psrcat"))
```
becker33:
```suggestion
        args.append("--with-psrcat={0}".format(spec["psrcat"].prefix.bin.psrcat))
```

becker33 (changes requested):
For the LWA-specific versions, do you know whether these are commonly used commits of the packages or just something specific to LWA?

var/spack/repos/builtin/packages/psrcat/package.py
downloadico: D'OH! :)

downloadico:
I don't have any idea if the commits specified in these packages are used by anyone outside of the LWA stack.
:(

var/spack/repos/builtin/packages/psrchive/package.py
```diff
@@ -0,0 +1,82 @@
...
+            args.append("--with-armadillo-dir={0}".format(spec["armadillo"].prefix))
+        if spec.satisfies("+cfitsio"):
+            args.append("--with-cfitsio-dir={0}".format(spec["cfitsio"].prefix))
+        if spec.satisfies("+eigen"):
+            args.append("--with-eigen-dir={0}".format(spec["eigen"].prefix))
+        if spec.satisfies("+xml"):
+            args.append("--with-xml-prefix={0}".format(spec["libxml2"].prefix))
+        if spec.satisfies("+qt"):
+            args.append("--with-qt-dir={0}".format(spec["qt"].prefix))
+        return args
```
becker33: For all of these dependencies, do you know whether `configure` will try to find them in the `PATH` if you don't specify anything? If so, we may need to specify `--without-*` for each of them when the variant is turned off.

becker33 (commented):
Mostly LGTM now

var/spack/repos/builtin/packages/psrchive/package.py
downloadico: I think explicitly setting them to "off" is a better idea in general.

becker33 (approved):

PR #34114, opened 2022-11-23
bcache: support external gettext when `libintl` is in glibc

bernhardkaindl:
Hi!

I've used `spack external find --all` on Rocky Linux 8 (RHEL 8) and built loads of packages with it.

glibc-based systems like Ubuntu, RHEL, etc don't provide `libintl` as a separate library,
because the `libintl` **functions used by gettext are already provided by the system's GNU libc**.

As a result, on these systems, **the system-provided `gettext` packages do not install a separate `libintl`**!

A few packages failed to build with this setup because their recipes assume that there is always a `libintl` with `gettext`.

`bcache` was one of these:

This is easy to fix by asking the gettext recipe if it was able to find a separate libintl library and not using it if not.

Tested with `spack external find gettext` in these environments to make sure it works in both situations (with libintl and without):
* **musl**-based Alpine Linux (not a glibc-based system hence no libintl in libc) chroot
* **glibc**-based Linux using external gettext

### DIFF
--- a/var/spack/repos/builtin/packages/bcache/package.py
+++ b/var/spack/repos/builtin/packages/bcache/package.py
@@ -25,7 +25,9 @@ class Bcache(MakefilePackage):
     depends_on("pkgconfig", type="build")
 
     def setup_build_environment(self, env):
-        env.append_flags("LDFLAGS", "-lintl")
+        # Add -lintl if provided by gettext, otherwise libintl is provided by the system's glibc:
+        if "libintl" in self.spec["gettext"].libs.joined():
+            env.append_flags("LDFLAGS", "-lintl")
 
     patch(
         "func_crc64.patch",
--- a/var/spack/repos/builtin/packages/bcache/package.py
+++ b/var/spack/repos/builtin/packages/bcache/package.py
@@ -26,7 +26,7 @@ class Bcache(MakefilePackage):
 
     def setup_build_environment(self, env):
         # Add -lintl if provided by gettext, otherwise libintl is provided by the system's glibc:
-        if "libintl" in self.spec["gettext"].libs.joined():
+        if any("libintl" in filename for filename in self.libs):
             env.append_flags("LDFLAGS", "-lintl")
 
     patch(


var/spack/repos/builtin/packages/bcache/package.py
```diff
@@ -25,7 +25,9 @@ class Bcache(MakefilePackage):
     depends_on("pkgconfig", type="build")
 
     def setup_build_environment(self, env):
-        env.append_flags("LDFLAGS", "-lintl")
+        # Add -lintl if provided by gettext, otherwise libintl is provided by the system's glibc:
+        if "libintl" in self.spec["gettext"].libs.joined():
```
sethrj: I think `self.spec["gettext"] is self`? Also since `.libs` is essentially a list of strings, it's better to do
```suggestion
        if any("libintl" in filename for filename in self.libs):
```
and hope that the user hasn't installed into a directory named `/my/libintl-doesnt-work/`

sethrj (approved):

bernhardkaindl:
@sethrj: Oops, I was too trustful of your suggestion, it caused:
> The 'bcache' package cannot find an attribute while trying to build from sources. This might be due to a change in Spack's package format to support multiple build-systems for a single package. You can fix this by updating the build recipe, and you can also report the issue as a bug. More information at https://spack.readthedocs.io/en/latest/packaging_guide.html#installation-procedure
```py

/data/spack/spack/lib/spack/spack/builder.py:132, in __getattr__:
  >>    132        def __getattr__(self, item):
```
The suggested check in `self.libs` (which does not exist) is the cause of this error from `__getattr__`.

### Checking for the `libs` property from `gettext` if it contains a `libintl` is what is needed.

That is what how I tested and pushed it initially for this PR as well.

Here is the implementation of it from `var/spack/repos/builtin/packages/gettext/package.py`:
```py
   @property
    def libs(self):
        return find_libraries(
            ["libasprintf", "libgettextlib", "libgettextpo", "libgettextsrc", "libintl"],
            root=self.prefix,
            recursive=True,
```
Test: Output line from `spack install -v bcache ^gettext@0.21.1` (at gettext with libintl)
`spack/lib/spack/env/gcc/gcc -O2 -Wall -g -std=gnu99  -lintl  ...` vs. the same without `-lintl` when linking against the glibc-system-provided gettext API - passed. The fix is:
```py
-        if any("libintl" in filename for filename in self.libs):
+        if any("libintl." in filename.split("/")[-1] for filename in self.spec["gettext"].libs):
```
This checks that a `libintl` library (**checks the file name, not the path to it**) is provided by gettext before adding `-lintl`.

var/spack/repos/builtin/packages/bcache/package.py
sethrj: Whoops I just saw that this is inside the bcache package not gettext package... my bad.

sethrj:
Related:  https://github.com/spack/spack/pull/21472 ?  https://github.com/spack/spack/issues/34269 ?

Issue #34085, opened 2022-11-23
provider index / cache race conditions


haampie:
### Steps to reproduce

```python
Traceback (most recent call last):
  File "/builds/spack/spack/bin/spack", line 54, in <module>
    sys.exit(main())
  File "/builds/spack/spack/lib/spack/spack_installable/main.py", line 36, in main
    sys.exit(spack.main.main(argv))
  File "/builds/spack/spack/lib/spack/spack/main.py", line 1000, in main
    return _main(argv)
  File "/builds/spack/spack/lib/spack/spack/main.py", line 955, in _main
    return finish_parse_and_run(parser, cmd_name, env_format_error)
  File "/builds/spack/spack/lib/spack/spack/main.py", line 983, in finish_parse_and_run
    return _invoke_command(command, parser, args, unknown)
  File "/builds/spack/spack/lib/spack/spack/main.py", line 632, in _invoke_command
    return_val = command(parser, args)
  File "/builds/spack/spack/lib/spack/spack/cmd/install.py", line 549, in install
    install_specs(zip(abstract_specs, concrete_specs), install_kwargs, args)
  File "/builds/spack/spack/lib/spack/spack/cmd/install.py", line 268, in install_specs
    install_specs_inside_environment(specs, install_kwargs, cli_args)
  File "/builds/spack/spack/lib/spack/spack/cmd/install.py", line 340, in install_specs_inside_environment
    env.install_specs(specs_to_install, **install_kwargs)
  File "/builds/spack/spack/lib/spack/spack/environment/environment.py", line 1735, in install_specs
    builder.install()
  File "/builds/spack/spack/lib/spack/spack/installer.py", line 1587, in install
    self._init_queue()
  File "/builds/spack/spack/lib/spack/spack/installer.py", line 1536, in _init_queue
    self._add_tasks(request, all_dependencies)
  File "/builds/spack/spack/lib/spack/spack/installer.py", line 1221, in _add_tasks
    spack.store.db.clear_failure(request.spec, force=True)
  File "/builds/spack/spack/lib/spack/spack/database.py", line 484, in clear_failure
    lock = self._prefix_failures.pop(spec.prefix, None)
  File "/builds/spack/spack/lib/spack/spack/spec.py", line 1738, in prefix
    self.prefix = spack.store.layout.path_for_spec(self)
  File "/builds/spack/spack/lib/spack/spack/directory_layout.py", line 336, in path_for_spec
    path = self.relative_path_for_spec(spec)
[...]

PR #34118, opened 2022-11-24
OpenFileTracker: close by inode, add deleter

haampie:
I'm having a hard time reproducing the exact issue in #34085, but in any case,
to me it looks safer to close file handlers by fstat instead of stat'ing their
path.


haampie:
@spackbot run pipeilne

haampie:
@spackbot run pipeline

haampie:
Easier to just go with #34122 instead of two separate prs.

PR #34122, opened 2022-11-24
Track locks by (dev, ino); close file handlers between tests

haampie:
- Don't remove by path, but by dev+ino.
- Make sure we close file descriptors before going to the next test.
- Make sure open files in the file tracker are actually open if they aren't removed at the end of a test
- Use (st_dev, st_ino, pid) as keys instead of just (st_ino, pid)

In a perfect world this cache should of course be empty after the test has
finished, but we have a few cases where locks are not released, and where
failing to acquire a lock leaves an inode -> fh cache entry.

### DIFF
--- a/lib/spack/llnl/util/lock.py
+++ b/lib/spack/llnl/util/lock.py
@@ -9,7 +9,7 @@ import socket
 import sys
 import time
 from datetime import datetime
-from typing import Dict, Tuple  # novm
+from typing import Dict, Tuple
 
 import llnl.util.tty as tty
 from llnl.util.lang import pretty_seconds
@@ -81,7 +81,7 @@ class OpenFileTracker(object):
 
     def __init__(self):
         """Create a new ``OpenFileTracker``."""
-        self._descriptors = {}  # type: Dict[Tuple[int, int], OpenFile]
+        self._descriptors: Dict[Tuple[int, int], OpenFile] = {}
 
     def get_fh(self, path):
         """Get a filehandle for a lockfile.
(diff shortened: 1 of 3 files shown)


haampie:
@spackbot run pipeline

lib/spack/llnl/util/lock.py
```diff
@@ -9,7 +9,7 @@
 import sys
 import time
 from datetime import datetime
-from typing import Dict, Tuple  # novm
+from typing import Dict, Tuple
```
haampie:
```suggestion
```

lib/spack/llnl/util/lock.py
```diff
@@ -81,7 +81,7 @@ class OpenFileTracker(object):
 
     def __init__(self):
         """Create a new ``OpenFileTracker``."""
-        self._descriptors = {}  # type: Dict[Tuple[int, int], OpenFile]
+        self._descriptors: Dict[Tuple[int, int], OpenFile] = {}
```
haampie:
```suggestion
        self._descriptors = {}
```

haampie:
@spackbot run pipeline

lib/spack/llnl/util/lock.py
```diff
@@ -129,32 +128,32 @@ def get_fh(self, path):
...
             del self._descriptors[key]
             open_file.fh.close()
 
+    def release_by_fh(self, fh):
+        self.release_by_stat(os.fstat(fh.fileno()))
+
+    def purge(self):
+        for key in list(self._descriptors.keys()):
+            self._descriptors[key].fh.close()
+            del self._descriptors[key]
```
alalazo, line 155: Not sure this is worth a comment that the list preserves the keys when the dict is modified in the loop.

alalazo (approved):

PR #34125, opened 2022-11-25
berkeleygw: add back python dependencies and tweak testsuite

migueldiascosta:


migueldiascosta:
@spackbot fix style

migueldiascosta:
@glaweh @bernhardkaindl care to test this?

bernhardkaindl (commented):
@migueldiascosta
with just `spack install --test=root berkeleygw`, got this on an 48-thread AMD Rome CPU - no error message before exit with exit code 2 from `make check`, also there were no progress messages from `make check`:
```
spack.util.executable.ProcessError: Command exited with status 2:
    'make' '-j1' 'check'
```
On an ARM64, I got a different error which I'll re-run.

migueldiascosta:
@bernhardkaindl the log of the tests should be in the spack-stage folder, I guess

from past experience with different CPU architectures, this might help, can you try?

```diff
@@ -110,6 +110,18 @@
             "testsuite/run_testsuite.sh",
         )
 
+        # slightly raise tolerance of some tests
+        si_epm_tests = ['Si', 'Si_cplx_spin']
+        if self.version >= Version("3.0"):
+            si_epm_tests.append('Si_hdf5')
+        for test in si_epm_tests:
+            filter_file("Precision : 6e-15", "Precision : 7e-15",
+                        join_path('testsuite', 'Si-EPM', test + '.test'))
+        for test_file in ['Si_subspace', 'Si_subspace_cplx', 'Si_subspace_cplx_spin']:
+            filter_file("Precision : 6e-15", "Precision : 7e-15",
+                        join_path('testsuite', 'Si-EPM_subspace', test_file + '.test'))
+        filter_file("Precision : 8e-15", "Precision : 9e-15", "testsuite/GaAs-EPM/GaAs.test")
+
     def build(self, spec, prefix):
 
         buildopts = []
```

bernhardkaindl:
@migueldiascosta `testsuite/run_testsuite.sh` deletes all log files by default and it's help text is wrong, -c is the default:
```py
     -c            delete all .log files and work directories after the run
...
# Some default settings.
cleanup="yes"
...
#       c) cleanup="yes";;
...
    [ "$cleanup" == "yes" ] && rm -f *.log
```

I had to run it manually to even get some output on stdout (the test execution from spack did never produce output), and the tests take a long time. By default they use `mpiexec -n 4` and with 4 CPU threads per job, that uses 16 cores. With that they never completed on my systems maybe due to some timeout. I then ran `make check BGW_TEST_MPI_NPROCS=24` directly from the shell, but of course that overcomitted the AMD Rome with 24 CPUs/48 HW-thready by a factor of 2 (load average was at ~112), but it "only" took a few hours (don't know, didn't time it), and the tests completed with this message:
```py
    testfile                                          # failed testcases
    --------------------------------------------------------------------
    Si-EPM/Si.test                                    1
    Si-EPM/Si_cplx.test                               1
    Si-EPM/Si_cplx_spin.test                          1
    Si2-SAPO/Si2-SAPO.test                            1
```
Of couse, due to `cleanup=yes` in `testsuite/run_testsuite.sh` all log files (which existed while the script was executing tests) were deleted when the script exited. Your change requires less precision for these tests could help, but I am not sure when I can take the system hostage again to run these tests again.
The tests also left 1.4 GB of data in about two dozen temporary directories named /tmp/BGW.xxxxxx (x=any char).

I think it does not make much sense to run these tests on a normal `spack --test=root install` but maybe deliver the full testsuite to let users run these tests after berkeleygw is installed to let them verify their installation. Maybe a small subset of these tests which complete quickly could be run from spack.

migueldiascosta:
@bernhardkaindl strange, they take 20m on my laptop, with 2 mpi procs and 2 threads (edit: it's important to limit the number of threads, otherwise it can indeed be very slow)

let me have a look at how to keep the output when using spack (besides the log files, the output is sent to stdout, so spack should keep it somewhere...) (edit: at least during the test, the output is in `install-time-test-log.txt` in the `spack-stage-berkeleygw-...` folder)

migueldiascosta:
hm, `install-time-test-log.txt` from the `check_build` step gets overwritten by the `installcheck` step, even if I add

```python
def installcheck(self):
    pass
```

how can we prevent that?

(the issue of the files in `/tmp` can be easily solved, by e.g. setting the environment variable `TEMPDIRPATH` to e.g. `join_path(self.build_directory, 'testsuite', 'tmp')`, which will then get deleted if all goes well)

migueldiascosta:
@spackbot fix style

bernhardkaindl:
On the `installcheck`: It should do nothing at all if the `Makefile` has no `installcheck` target:
```py
lib/spack/spack/build_systems/makefile.py
    def installcheck(self):
        """Searches the Makefile for an ``installcheck`` target
        and runs it if found.
        """
        with fs.working_dir(self.build_directory):
            self.pkg._if_make_target_execute("installcheck")
```
So it should not make any difference. You can remove `def installcheck(self): pass` later when pushing a new commit.

BTW: I think this
```py
    @run_after("build")
    def check_build(self):
            make("check")
```
could be replaced by:
```python
    def check(self):
            make("check")
```
and `"check"` has a default implementation:
```py
    def check(self):
        """Run "make" on the ``test`` and ``check`` targets, if found."""
        with fs.working_dir(self.build_directory):
            self.pkg._if_make_target_execute("test")
            self.pkg._if_make_target_execute("check")
```
It is the default of `build_time_test_callbacks` which is called this way:
```py
def execute_build_time_tests(builder):
    """Execute the build-time tests prescribed by builder.

    Args:
        builder (Builder): builder prescribing the test callbacks. The name of the callbacks is
            stored as a list of strings in the ``build_time_test_callbacks`` attribute.
    """
    builder.pkg.run_test_callbacks(builder, builder.build_time_test_callbacks, "build")
```
That means it only gets called when `run_tests==True`.

So besides trying to run `make test` before `make check` it would be ok if you just remove `check_build` entirely and use the default (remove the code to call `make check` entirely, and `make check` should be called when `install --test=...` is used)

Then `"make check"` is called the way it is designed to be run in `spack` by default.

On the log files, I don't have any further analysis so far, except that it would be great if you could patch 
also use `filter_file` on `testsuite/run_testsuite.sh` to repalce `cleanup="yes"` with `cleanup="no"`, to keep the log files for the event that the build or `make check` fails (otherwise spack deletes the stage dir by default).

migueldiascosta:
@bernhardkaindl indeed, there's no need to define `check` at all (facepalm)

keeping the tests can be done by setting the environment variable `SAVETESTDIRS=yes`, done

but if everything succeeds, `install-time-test-log.txt`, which during the build did have the output of the testsuite, gets overwritten with `(...) RUN-TESTS: install-time tests [installcheck] (...)` and that's what ends up in the `.spack` folder of the installation folder - it would be nice to keep the testsuite output (even when everything passes, it can have warnings)

migueldiascosta:
ah, ok, the individual log files are now in `share/BerkeleyGW/testsuite`...

but then we need to do something else with that `tmp` folder with all the large BGW.xxxxxx folders, let me see

bernhardkaindl (approved):

bernhardkaindl (approved):
I think this is a good step forward.

migueldiascosta:
Great, thanks for all the help @bernhardkaindl!

migueldiascosta:
(I still think Spack shouldn't be overriding the output of `make check` with the output of `make installcheck` though...)

PR #34128, opened 2022-11-25
Octopus: branch for Octopus development is now "main"

fangohr:
Historically, development of the Octopus code was done on the "develop" branch on https://gitlab.com/octopus-code/octopus but now development takes place on "main" (since Q3 2022).

The suggestion in this PR to solve the issue is to keep the spack label `octopus@develop` as this better indicates this is the development branch on git than `octopus@main`, but of course to use the `main` branch (there is no choice here - the `develop` branch is not touched anymore). Sticking to `octopus@develop` as the version label also keeps backwards compatibility.


RemiLacroix-IDRIS (commented):
Seems OK to me.

michaelkuhn (approved):

PR #34130, opened 2022-11-25
Fix recipe for py-onnx-runtime

iarspider:


var/spack/repos/builtin/packages/py-onnx-runtime/package.py
```diff
@@ -126,4 +127,9 @@ def cmake_args(self):
 
     @run_after("install")
     def install_python(self):
-        PythonPackage.install(self, self.spec, self.prefix)
+        platlib = self.spec["python"].package.platlib
+        with working_dir(self.build_directory):
+            py = which("python")
```
adamjstewart: This won't work for `python~pythoncmd` where only the `python3` executable exists. Spack already has a `python` function injected into the module, so you can use `python(...)` without using `which(...)` first.

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
```diff
@@ -126,4 +127,9 @@ def cmake_args(self):
...
+            py("setup.py", "build")
```
adamjstewart: Direct invocation of `setup.py` is deprecated, see https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html and #27798. We should instead use a build tool like pip.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
```diff
@@ -126,4 +132,11 @@ def cmake_args(self):
...
     def install_python(self):
-        PythonPackage.install(self, self.spec, self.prefix)
+        """Install everything from build directory."""
+        # Truncated version of PythonPipBuilder.install
+
+        args = PythonPipBuilder.std_args(self) + ["--prefix=" + prefix, "."]
+
+        pip = inspect.getmodule(self).pip
+        with fs.working_dir(self.build_directory):
+            pip(*args)
```
adamjstewart:
```suggestion
        args = std_pip_args + ["--prefix=" + prefix, "."]
        with working_dir(self.build_directory):
            pip(*args)
```
This should be sufficient. This is also what we do in `py-tensorflow` and many other packages like this.

iarspider:
@spackbot fix style

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
```diff
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import inspect
```
adamjstewart: You can remove this import

adamjstewart (approved):

Issue #34012, opened 2022-11-19
spack mirroring hungs


zeronewb:
### Steps to reproduce
Clean room env for CI ( without keys, github fingerprints, etc )
spack mirror create -d ~/mirror --all
Wait for a day or two ( nic and wan speed 10g+ ).

Mirroring is slow due to nested loops and direct invocations of curl binary; please, consider switching to PycURL  + asyncio with pools, or to PyPy, if you don't like external libs.

### Information on your system

* **Spack:** 0.20.0.dev0 (839a14c0ba11f88ec4cd944adad628430a8076c5)
* **Python:** 3.9.2
* **Platform:** linux-debian11-zen3
* **Concretizer:** clingo

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

zeronewb:
Related 
https://github.com/spack/spack/issues/13604
https://github.com/spack/spack/pull/14616
Please, consider switching to at least https://github.com/libgit2/pygit2, instead of spawning a subshell with git running;
and, please, provide options for creating a mirror directly from s3, with help of AWS SDK.

```lib/spack/spack/util/web.py:392 ==> [2022-11-20-06:52:06.507986] Checking existence of https://mirror.spack.io/sollve/polly-1.0a2.tar.gz
lib/spack/spack/util/web.py:441 ==> [2022-11-20-06:52:07.109886] Failure reading URL: Download failed: HTTP Error 404: Not Found
lib/spack/spack/fetch_strategy.py:339 ==> [2022-11-20-06:52:07.110075] URL does not exist: https://mirror.spack.io/sollve/polly-1.0a2.tar.gz
lib/spack/spack/stage.py:524 ==> [2022-11-20-06:52:07.110211] FailedDownloadError: Failed to fetch file from URL: https://mirror.spack.io/sollve/polly-1.0a2.tar.gz
lib/spack/spack/fetch_strategy.py:843 ==> [2022-11-20-06:52:07.110343] Cloning git repository: https://github.com/SOLLVE/polly.git at commit 96168ae6fb436e95cd756950855a57b895070047
lib/spack/spack/util/executable.py:195 ==> [2022-11-20-06:52:07.110651] '/usr/bin/git' '--version'
[...]

PR #34133, opened 2022-11-25
Disable git terminal promt via env variable GIT_TERMINAL_PROMPT.

zeronewb:
Stupid fix for mirroring process.
Partially solves https://github.com/spack/spack/issues/34012


robertu94:
What happens if you have a private external repo with packages that require a password protected ssh-key or a git username+password to access. My team uses packages like this for our pre-release software where we don’t want to distribute outside our team. While upstream spack generally r equires things to be accessible, this may or may not work well for others with private packages.

lib/spack/spack/config.py
```diff
@@ -95,6 +95,7 @@
         "debug": False,
         "connect_timeout": 10,
         "verify_ssl": True,
+        "git_terminal_prompt_enable": False,
```
tgamblin, line 98: I don't think we need a config option for this -- we can just hard-disable it in the git fetch strategy, unless I'm missing something.

tgamblin (changes requested):

lib/spack/spack/fetch_strategy.py
```diff
@@ -786,7 +786,9 @@ def git(self):
             # with git as well.
             if not spack.config.get("config:verify_ssl"):
                 self._git.add_default_env("GIT_SSL_NO_VERIFY", "true")
-
+            # stupid fix for mirroring
+            if not spack.config.get("config:git_terminal_prompt_enable"):
+                self._git.add_default_env("GIT_TERMINAL_PROMPT", "0")
```
tgamblin, line 783: Instead of having a config option for this, could you instead do:

```suggestion
            if not sys.stdout.isatty():
                self._git.add_default_env("GIT_TERMINAL_PROMPT", "0")
```

That will preserve the prompt for interactive sessions, but will disable it in CI, which is where I think you want to avoid the prompts -- yes?

@robertu94: would that address your concern?

tgamblin (changes requested):

lib/spack/spack/fetch_strategy.py
tgamblin: If we do this my suggestion is that we remove the config option.

robertu94: yes.

PR #34134, opened 2022-11-25
Add tree-sitter support for Emacs 29+

berquist:
It's currently on `master` and will be present for the upcoming 29.1 release, hence the error message (https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg01443.html).


bernhardkaindl (approved):
Built with +treesitter, starts up.

PR #34136, opened 2022-11-25
concrete specs are not virtual

haampie:
Checking this ahead of time avoids hitting the repo cache.

Should make bootstrapping of gnupg in CI fast again, since it regressed
in de8c827.

Edit: hm, since `de8c827` we almost always trigger provider cache indexing :(
This PR just delays it (at least it doesn't feel like a download is broken).


haampie:
Clsoing in favor of #34139

PR #34138, opened 2022-11-26
Fix dependency specification for CuSolver variant in HiOp.

### LINKED ISSUE #571: Spack does not configure cuSOLVER LU dependency
The Spack package for HiOp variant `cusolver` sets a CMake option that is not present in HiOp anymore:
```python
self.define_from_variant("HIOP_USE_CUSOLVER", "cusolver"),
```
The correct flag is `HIOP_USE_CUSOLVER_LU`. It is used to set up KLU+cuSOLVER refactorization linear solver. Installing HiOp with Spack and `+cusolver` option does not build cuSOLVER LU module and its tests. No error message is returned. This will likely cause issues in packages like ExaGO who are using Spack to build HiOp.

To fix this, one first needs to set
```python
self.define_from_variant("HIOP_USE_CUSOLVER_LU", "cusolver"),
```
and then set Spack dependencies
```python
depends_on("cuda", when="+cusolver")
depends_on("suite-sparse", when="+cusolver")
```
I also suggest renaming this option from `cusolver` -> `cusolver_lu` to be consistent with HiOp configuration option.

This variant should not be available for HiOp versions < 0.7.1.

CC @nychiang @cnpetra @kswirydo

cameronrutherford:
Closes https://github.com/LLNL/hiop/issues/571.

Not sure if a conflict is the correct way to require `cuda` with `cusolver_lu`.

cc @pelesh

### DIFF
--- a/var/spack/repos/builtin/packages/hiop/package.py
+++ b/var/spack/repos/builtin/packages/hiop/package.py
@@ -63,7 +63,12 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
         description="Ultra safety checks - " "used for increased robustness and self-diagnostics",
     )
     variant("ginkgo", default=False, description="Enable/disable ginkgo solver")
-    variant("cusolver_lu", default=False, when="@0.7.1:", description="Enable/disable cuSovler LU refactorization")
+    variant(
+        "cusolver_lu",
+        default=False,
+        when="@0.7.1:",
+        description="Enable/disable cuSovler LU refactorization",
+    )
     depends_on("lapack")
     depends_on("blas")
     depends_on("cmake@3.18:", type="build")
@@ -97,7 +102,6 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("cuda@11:", when="@develop:+cuda")
     conflicts("+cusolver_lu", when="~cuda", msg="Cusolver LU refactorization requires CUDA")
-    conflicts("+cusolver_lu", when="~suite-sparse", msg="Cusolver LU refactorization requires KLU from SuiteSparse library")
     depends_on("raja", when="+raja")
     depends_on("umpire", when="+raja")
     depends_on("raja+openmp", when="+raja~cuda~rocm")
--- a/var/spack/repos/builtin/packages/hiop/package.py
+++ b/var/spack/repos/builtin/packages/hiop/package.py
@@ -63,8 +63,7 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
         description="Ultra safety checks - " "used for increased robustness and self-diagnostics",
     )
     variant("ginkgo", default=False, description="Enable/disable ginkgo solver")
-    variant("cusolver_lu", default=False, when="@0.7.1:", description="Enable/disable cuSovler")
-
+    variant("cusolver_lu", default=False, when="@0.7.1:", description="Enable/disable cuSovler LU refactorization")
     depends_on("lapack")
     depends_on("blas")
     depends_on("cmake@3.18:", type="build")
--- a/var/spack/repos/builtin/packages/hiop/package.py
+++ b/var/spack/repos/builtin/packages/hiop/package.py
@@ -66,7 +66,7 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
     variant(
         "cusolver_lu",
         default=False,
-        when="@0.7.1:",
+        when="+cuda @0.7.1:",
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -63,7 +63,7 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
         description="Ultra safety checks - " "used for increased robustness and self-diagnostics",
     )
     variant("ginkgo", default=False, description="Enable/disable ginkgo solver")
-    variant("cusolver", default=False, description="Enable/disable cuSovler")
+    variant("cusolver_lu", default=False, when="@0.7.1:", description="Enable/disable cuSovler")
```
pelesh: `cusolver_lu` variante enables very specific cuSolver feature, so it migth be a good idea to mention that specifically here.

```suggestion
    variant("cusolver_lu", default=False, when="@0.7.1:", description="Enable/disable cuSovler LU refactorization")

```

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -97,6 +97,7 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("magma@{0}:".format(magma_v), when="@{0}:+rocm".format(hiop_v))
 
     depends_on("cuda@11:", when="@develop:+cuda")
+    conflicts("+cusolver_lu", when="~cuda", msg="Cusolver requires CUDA")
```
pelesh: cuSolver LU refactorization needs KLU to get initial L and U factors and their sparsity patterns. 
```suggestion
    conflicts("+cusolver_lu", when="~cuda", msg="Cusolver LU refactorization requires CUDA")
    conflicts("+cusolver_lu", when="~suite-sparse", msg="Cusolver LU refactorization requires KLU from SuiteSparse library")

```

pelesh (changes requested):
cuSolver LU refactorization needs a host LU solver to generate initial L and U factors. Currently, only KLU from SuiteSparse library can be used. Therefore dependency on SuiteSparse needs to be enforced in `cusolver_lu`.

pelesh:
> Not sure if a conflict is the correct way to require `cuda` with `cusolver_lu`.

Only minor adjustments needed ;).

pelesh:
I believe my earlier suggestion may have been misleading. Package `suite-sparse` is dependency for HiOp variants `kron` and `cusolver_lu`, but I am not sure `suite-sparse` should be a HiOp variant for itself. I am still learning Spack best practices so I will only suggest what needs to be done but let you decide on how 😄.

pelesh (approved):
I think this is ready to merge.

jaelynlitz (approved):

jaelynlitz (dismissed):
Looks good!

pelesh (dismissed):
Lokks good, I suggest to merge!

cameronrutherford:
@haampie or @alalazo would so you kindly merge? You both have been very helpful in the past :)

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -63,8 +63,12 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
         description="Ultra safety checks - " "used for increased robustness and self-diagnostics",
     )
     variant("ginkgo", default=False, description="Enable/disable ginkgo solver")
-    variant("cusolver", default=False, description="Enable/disable cuSovler")
-
+    variant(
+        "cusolver_lu",
+        default=False,
+        when="@0.7.1:",
```
alalazo: Would this:
```suggestion
        when="+cuda @0.7.1:",
```
and removing the conflict work too?

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -97,7 +101,7 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("magma@{0}:".format(magma_v), when="@{0}:+rocm".format(hiop_v))
 
     depends_on("cuda@11:", when="@develop:+cuda")
-
+    conflicts("+cusolver_lu", when="~cuda", msg="Cusolver LU refactorization requires CUDA")
```
alalazo:
```suggestion
```

var/spack/repos/builtin/packages/hiop/package.py
pelesh: I think so. Thanks!

pelesh (approved):

alalazo (approved):

PR #34140, opened 2022-11-26
aragorn: updated package version and URL

luke-dt:
aragorn is hosted on a new URL now, I updated that and added the new version.


bernhardkaindl (approved):
Built and tested `spack checksum`

PR #34144, opened 2022-11-26
Revert "`url_exists` related improvements (#34095)"

balay:
This reverts commit d06fd26c9ac8dd525fc129096188e2ea9fd2d0d7.

```
$ ./bin/spack fetch blaspp
==> Error: All fetchers failed for spack-stage-blaspp-2022.07.00-zrcutz3ecr5dikupw4paiqnzo5ht5i7r
```

### DIFF
--- a/lib/spack/spack/test/web.py
+++ b/lib/spack/spack/test/web.py
@@ -6,7 +6,6 @@ import collections
 import os
 import posixpath
 import sys
-from urllib.request import Request
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


haampie:
For posterity: the problem is that Bitbucket's API forwards download requests to an S3 bucket using a temporary URL. This URL includes a signature for the request, which embeds the HTTP verb. That means only GET requests are allowed, and HEAD requests would fail verification, leading to 403 erros. The same is observed when using `curl -LI ...`

haampie (approved):

PR #34146, opened 2022-11-27
ftk: adding version 0.0.7.1

hguo:
adding version 0.0.7.1 to ftk

### DIFF
--- a/var/spack/repos/builtin/packages/ftk/package.py
+++ b/var/spack/repos/builtin/packages/ftk/package.py
@@ -12,7 +12,7 @@ class Ftk(CMakePackage):
 
     # Add a proper url for your package's homepage here.
     homepage = "https://github.com/hguo/ftk"
-    url = "https://github.com/hguo/ftk/archive/0.0.5.tar.gz"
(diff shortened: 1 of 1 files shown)


bernhardkaindl (commented):
@hguo I see that the gmp variant is disabled with this PR as well.

Just for completeness, what's the reason for that?

Normally, it would be great to have these other changes (also like the new metis variant, and 

depends_on("metis"), when "+mpi")

explained in the commit text+message as well, but for now please just reply in the PR to no trigger a rebuild of the entire CI pipeline yet.

Another look shows that +mpi is essentiall +metis+mpi without  adding -DFTK_USE_METIS - why not enable it then fully as well?

You could force the concretizer/user to enable metis when +mpi with:

conflicts("~metis", when="+mpi")

hguo:
I removed the GMP variant because the code does not necessarily rely on GMP, and the use of GMP may produce inconsistent results with the default fixed-precision arithmetic.

The MPI variant without metis is still functional but not ideal in performance, but enabling metis with +mpi is a good idea.  Would you advise adding the conflicts statement in this PR?  Thank you.

bernhardkaindl (approved):
Builds in default and `+gmp+metis` variants

PR #34149, opened 2022-11-27
Add py-python-lsp-server and dependencies

alecbcs:
Add `python-lsp-server` and its required dependencies (`python-lsp-jsonrpc` and `docstring-to-markdown`). 

Python-lsp-server is a language server using jedi to provide ide-like autocompletion in text editors like emacs and vim for python.

### DIFF
--- a/var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
+++ b/var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
@@ -16,6 +16,6 @@ class PyPythonLspJsonrpc(PythonPackage):
 
     version("1.0.0", sha256="7bec170733db628d3506ea3a5288ff76aa33c70215ed223abdb0d95e957660bd")
 
-    depends_on("python@3.7:", type=("build", "run"))
-    depends_on("py-setuptools@61.2.0:", type="build")
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
     depends_on("py-ujson@3.0.0:", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
+++ b/var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
@@ -1,0 +1,21 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyPythonLspJsonrpc(PythonPackage):
+    """A Python 3.7+ server implementation of the JSON RPC 2.0 protocol."""
+
+    homepage = "https://github.com/python-lsp/python-lsp-jsonrpc"
+    pypi = "python-lsp-jsonrpc/python-lsp-jsonrpc-1.0.0.tar.gz"
+
+    maintainers = ["alecbcs"]
+
+    version("1.0.0", sha256="7bec170733db628d3506ea3a5288ff76aa33c70215ed223abdb0d95e957660bd")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools@61.2.0:", type="build")
+    depends_on("py-ujson@3.0.0:", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
+++ b/var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
@@ -16,6 +16,5 @@ class PyPythonLspJsonrpc(PythonPackage):
 
     version("1.0.0", sha256="7bec170733db628d3506ea3a5288ff76aa33c70215ed223abdb0d95e957660bd")
 
-    depends_on("python@3.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-ujson@3.0.0:", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-python-lsp-server/package.py
+++ b/var/spack/repos/builtin/packages/py-python-lsp-server/package.py
@@ -17,8 +17,8 @@ class PyPythonLspServer(PythonPackage):
     version("1.6.0", sha256="d75cdff9027c4212e5b9e861e9a0219219c8e2c69508d9f24949951dabd0dc1b")
 
     depends_on("python@3.7:", type=("build", "run"))
-    depends_on("py-setuptools@61.2.0:", type="build")
-    depends_on("py-setuptools-scm@3.4.3:+toml", type="build")
+    depends_on("py-setuptools@61.2.0:", type=("build", "run"))
+    depends_on("py-setuptools-scm@3.4.3:+toml", type=("build", "run"))
     depends_on("py-docstring-to-markdown", type=("build", "run"))
     depends_on("py-jedi@0.17.2:0.18", type=("build", "run"))
     depends_on("py-pluggy@1.0.0:", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-python-lsp-server/package.py
+++ b/var/spack/repos/builtin/packages/py-python-lsp-server/package.py
@@ -18,8 +18,9 @@ class PyPythonLspServer(PythonPackage):
 
     depends_on("python@3.7:", type=("build", "run"))
     depends_on("py-setuptools@61.2.0:", type="build")
-    depends_on("py-jedi@0.17.2:0.18", type=("build", "run"))
-    depends_on("py-python-lsp-jsonrpc@1.0.0:", type=("build", "run"))
-    depends_on("py-pluggy@1.0.0:", type=("build", "run"))
+    depends_on("py-setuptools-scm@3.4.3:+toml", type="build")
     depends_on("py-docstring-to-markdown", type=("build", "run"))
+    depends_on("py-jedi@0.17.2:0.18", type=("build", "run"))
+    depends_on("py-pluggy@1.0.0:", type=("build", "run"))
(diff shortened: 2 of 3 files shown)


var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
```diff
@@ -0,0 +1,21 @@
...
+
+    homepage = "https://github.com/python-lsp/python-lsp-jsonrpc"
+    pypi = "python-lsp-jsonrpc/python-lsp-jsonrpc-1.0.0.tar.gz"
+
+    maintainers = ["alecbcs"]
+
+    version("1.0.0", sha256="7bec170733db628d3506ea3a5288ff76aa33c70215ed223abdb0d95e957660bd")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools@61.2.0:", type="build")
```
adamjstewart: Where do these version constraints come from?

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
```diff
@@ -0,0 +1,25 @@
...
+
+    homepage = "https://github.com/python-lsp/python-lsp-server"
+    pypi = "python-lsp-server/python-lsp-server-1.6.0.tar.gz"
+
+    maintainers = ["alecbcs"]
+
+    version("1.6.0", sha256="d75cdff9027c4212e5b9e861e9a0219219c8e2c69508d9f24949951dabd0dc1b")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools@61.2.0:", type="build")
```
adamjstewart: Missing build dep: `setuptools_scm[toml]>=3.4.3`

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
```diff
@@ -0,0 +1,25 @@
...
+    depends_on("py-jedi@0.17.2:0.19.0", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-jedi@0.17.2:0.18", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
```diff
@@ -0,0 +1,25 @@
...
+    depends_on("py-python-lsp-jsonrpc@1.0.0:", type=("build", "run"))
+    depends_on("py-pluggy@1.0.0:", type=("build", "run"))
+    depends_on("py-docstring-to-markdown", type=("build", "run"))
+    depends_on("py-ujson@3.0.0:", type=("build", "run"))
```
adamjstewart, line 26: Missing build/run dep on setuptools

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
alecbcs: Haven't packaged too many python applications before outside of pypi, so I was looking at the project's [pyproject.toml](https://github.com/python-lsp/python-lsp-server/blob/develop/pyproject.toml) file. That contains the version constraints under the build-system.requires property. Is this not the right place/are those constraints created as part of a template?

alecbcs: Oh and the general `python@3.7:` I got from the [description](https://github.com/python-lsp/python-lsp-server#readme) of the project.
> A Python 3.7+ implementation of the [Language Server Protocol](https://github.com/Microsoft/language-server-protocol).

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
alecbcs: Oh good catch! I'll make sure to update and add this dep! By the way what is `setuptools_scm[toml]`? I've seen this as part of a couple of packages but haven't ever used it myself.

adamjstewart: This would be `py-setuptools-scm@3.4.3:+toml`. Pip's extras (brackets) are equivalent to Spack's variants (plus/minus).

var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
adamjstewart: That file is present on develop but not in 1.0.0, that's why I'm confused.

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
alecbcs: Oh very cool! Thanks for the explanation.

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
alecbcs: Not quite sure if I'm following. Do you mean that `py-setuptools` should be both a build and runtime dependency instead of just a build dep?

var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
alecbcs: Ah! Wow you're right I'd assumed since the project seems fairly stable/inactive that the pyproject.toml would have been there at the 1.0.0 release but you're right that's not the case! I'll relax the constraints and see if it continues to build correctly.

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
adamjstewart: Correct

alecbcs: Gotcha! Should we change this in the default python template? I believe that it listed `py-setuptools` as a build only dependency and I defaulted to that when I uncommented the code and added the version constraint.

adamjstewart: Setuptools is rarely needed at run-time, but it is for this package.

alecbcs: Ah okay interesting! For future reference how were you able to tell that setup tools was also a runtime dependency from looking at the package?

adamjstewart: There's a few ways you can be clued off to this. Things listed in `requires` under the `[build-system]` section of `pyproject.toml` or imported in `setup.py` are at a bare minimum a build dependency. If you also see them in `dependencies` under `[project]` in `pyproject.toml`, or in `install_requires` in `setup.py` or `setup.cfg`, then they are also needed at run-time. Oftentimes, developers will assume that setuptools is always installed and won't list it explicitly. If you recursively grep the source directory and see imports of `setuptools`, or more often `pkg_resources`, outside of `setup.py`, then it's also needed at run-time. Also, for older versions of Python, any usage of entrypoints also implies usage of setuptools at run-time. This is easier to see post-installation as there will be files in `prefix.bin` that didn't exist in the source dir but import setuptools/pkg_resources.

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
```diff
@@ -0,0 +1,26 @@
...
+    homepage = "https://github.com/python-lsp/python-lsp-server"
+    pypi = "python-lsp-server/python-lsp-server-1.6.0.tar.gz"
+
+    maintainers = ["alecbcs"]
+
+    version("1.6.0", sha256="d75cdff9027c4212e5b9e861e9a0219219c8e2c69508d9f24949951dabd0dc1b")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools@61.2.0:", type=("build", "run"))
+    depends_on("py-setuptools-scm@3.4.3:+toml", type=("build", "run"))
```
adamjstewart: I don't think setuptools-scm is needed at run-time, only build-time

var/spack/repos/builtin/packages/py-python-lsp-jsonrpc/package.py
```diff
@@ -0,0 +1,21 @@
...
+    """A Python 3.7+ server implementation of the JSON RPC 2.0 protocol."""
+
+    homepage = "https://github.com/python-lsp/python-lsp-jsonrpc"
+    pypi = "python-lsp-jsonrpc/python-lsp-jsonrpc-1.0.0.tar.gz"
+
+    maintainers = ["alecbcs"]
+
+    version("1.0.0", sha256="7bec170733db628d3506ea3a5288ff76aa33c70215ed223abdb0d95e957660bd")
+
+    depends_on("python@3.6:", type=("build", "run"))
```
adamjstewart: Spack only supports Python 3.7+, so you don't actually need to list this dep, up to you

alecbcs: Oh I hadn't realized we only support back to 3.7 and onwards, good catch!

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
alecbcs: Yep! Updated to just be a build time dependency.

adamjstewart (approved):

var/spack/repos/builtin/packages/py-python-lsp-server/package.py
alecbcs: Wow thanks for the detailed explanation! This is great! I hadn't thought of using grep to check for imports of setup tools throughout the project, I'll definitely have to try this out on future packages.

adamjstewart: Feel free to add this to the docs if you think it would help: https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#setuptools

PR #34151, opened 2022-11-28
parser: refactor with coarser token granularity

### LINKED ISSUE #13752: Spec parsing broken for key=value pairs
If a spec contains more than one `key=value` pair or the `key=value` pair is not at the end of the spec, parsing breaks:

```console
$ spack spec "py-matplotlib backend=agg target=x86_64 +image ^python@3.7.5"
Input spec
py-matplotlib backend=agg target=x86_64 +image ^python@3.7.5

Concretized
==> Error: invalid values for variant "backend" in package "py-matplotlib": ['agg target=x86_64 +image ^python@3.7.5']
```

This is due to the fact that the regexp for matching the value part only checks for a non-whitespace character at the start and then matches everything (one-line PR incoming).

alalazo:
fixes #13752

## Motivation

Our parser grew to be quite complex, with a 2-state lexer and logic in the parser that has up to 5 levels of nested conditionals:

https://github.com/spack/spack/blob/066ec316047cb6dfc92a3fef9980718ade47d4aa/lib/spack/spack/spec.py#L5085-L5108
    
In the future, to turn compilers into proper dependencies, we'll have to increase the complexity further as we foresee the need to add:
1. Edge attributes
2. Spec nesting

to the spec syntax (see https://github.com/spack/seps/pull/5 for an initial discussion of those changes).  The main attempt here is thus to _simplify the existing code_ before we start extending it later. We try to do that by adopting a different token granularity, and by using more complex regexes for tokenization. This allow us to a have a "flatter" encoding for the parser (i.e. with less nested conditionals):

https://github.com/alalazo/spack/blob/ee99ded647691a355a31b548da667ffa3ec7318c/lib/spack/spack/parser.py#L270-L357

and a trivial lexer:

https://github.com/alalazo/spack/blob/ee99ded647691a355a31b548da667ffa3ec7318c/lib/spack/spack/parser.py#L134-L158

## Differences with the previous parser

~There are currently 2 known differences with the previous parser, which have been added on purpose:~

- ~No spaces allowed after a sigil (e.g. `foo @ 1.2.3` is invalid while `foo @1.2.3` is valid)~
- ~`/<hash> @1.2.3` can be parsed as a concrete spec followed by an anonymous spec (before was invalid)~

~We can recover the previous behavior on both ones but, especially for the second one, it seems the current behavior in the PR is more consistent.~

The parser is currently 100% backward compatible.

## Error handling

Being based on more complex regexes, we can possibly improve error handling by adding regexes for common issues and hint users on that. I'll leave that for a following PR, but there's a stub for this approach here:

https://github.com/alalazo/spack/blob/ee99ded647691a355a31b548da667ffa3ec7318c/lib/spack/spack/parser.py#L155-L158

## Performance

To be sure we don't add any performance penalty with this new encoding, I measured:
```console
$ spack python -m timeit -s "import spack.spec" -c "spack.spec.Spec(<spec>)"
```
for different specs on my machine:

* **Spack:** 0.20.0.dev0 (c9db4e50ba045f5697816187accaf2451cb1aae7)
* **Python:** 3.8.10
* **Concretizer:** clingo

results are:

| Spec          | develop       | this PR |
| ------------- | ------------- | ------- |
| `trilinos`  |  28.9 usec | 13.1 usec |
| `trilinos @1.2.10:1.4.20,2.0.1`  | 131 usec  | 120 usec |
| `trilinos %gcc`  | 44.9 usec  | 20.9 usec |
| `trilinos +foo`  | 44.1 usec  | 21.3 usec |
| `trilinos foo=bar`  | 59.5 usec  | 25.6 usec |
| `trilinos foo=bar ^ mpich foo=baz`  | 120 usec  | 82.1 usec |

so this new parser seems to be consistently faster than the previous one.

## Modifications

In this PR we just substituted the Spec parser, which means:
- [x] Deleted in `spec.py` the `SpecParser` and `SpecLexer` classes. deleted `spack/parse.py`
- [x] Added a new parser in `spack/parser.py`
- [x] Hooked the new parser in all the places the previous one was used
- [x] Adapted unit tests in `test/spec_syntax.py`

## Possible future improvements

Random thoughts while working on the PR:
- Currently we transform hashes and files into specs during parsing. I think we might want to introduce an additional step and parse special objects like a `FileSpec` etc. in-between parsing and concretization.


lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+VALID_QUOTED_VALUE = r"[\"']+([a-zA-Z_0-9\-+\*.,:=\~\/\\\s]+)[\"']+"
+
+VALID_VERSION = r"([a-zA-Z_0-9-.][a-zA-Z_0-9-.]*)"
+VALID_VERSION_RANGE = rf"({VALID_VERSION}:{VALID_VERSION}|:{VALID_VERSION}|{VALID_VERSION}:|:)"
+VALID_VERSION_LIST = (
+    rf"({VALID_VERSION_RANGE}|{VALID_VERSION})([,]({VALID_VERSION_RANGE}|{VALID_VERSION}))*"
+)
+
+
+class TokenKind(Enum):
```
tgamblin: Can this be `TokenType`?

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+]
+
+ERROR_HANDLING_REGEXES = TOKEN_REGEXES + [
+    rf"(?P<{TokenKind.UNEXPECTED}>(.[\s]*))",
+]
+
+#: Maps a string representation to the corresponding token kind
+STR_TO_TOKEN = {str(x): x for x in TokenKind}
+
+MASTER_REGEX = re.compile("|".join(TOKEN_REGEXES))
```
tgamblin: Avoid "master"... how about just `ALL_TOKENS`?  Or `ANY_TOKEN`?

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+            end of the input text.
+    """
+    scanner = MASTER_REGEX.scanner(text)  # type: ignore[attr-defined]
+    match: Optional[re.Match] = None
+    for match in iter(scanner.match, None):
+        yield Token(
+            STR_TO_TOKEN[match.lastgroup],  # type: ignore[attr-defined]
+            match.group(),  # type: ignore[attr-defined]
+            match.start(),  # type: ignore[attr-defined]
+            match.end(),  # type: ignore[attr-defined]
```
tgamblin: Do we need all these ignores?  Is there a cleaner way to do it?

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+"""Parser for spec literals"""
```
tgamblin: I think this module should be called `spec_parser`, or maybe `spec` should be a package and this should be `spec.parser`.  Can you bring over the relevant parts of the module docstring from `spec.py` and clean them up to reflect the new parse details that have been added since that comment came into existence?  Especially the BNF for the parser - it would be useful to have that here.

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+    # Compilers
+    COMPILER_AND_VERSION = auto()
+    COMPILER = auto()
+    # Package name
+    UNQUALIFIED_PACKAGE_NAME = auto()
+    FULLY_QUALIFIED_PACKAGE_NAME = auto()
+    # White spaces
+    WS = auto()
+    # Unexpected character
+    UNEXPECTED = auto()
```
tgamblin: The enum is great -- I think it would be even cleaner if you did something [like this](https://stackoverflow.com/a/19300424/9122) so that each enum value can have its regex associated with it right here, e.g. here is a simplified version of what I mean:

```python
#!/usr/bin/env python3

import enum
import re

FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
VERSION = r"([a-zA-Z_0-9-.][a-zA-Z_0-9-.]*)"
VERSION_RANGE = rf"({VERSION}:{VERSION}|:{VERSION}|{VERSION}:|:)"
VERSION_LIST = rf"({VERSION_RANGE}|{VERSION})([,]({VERSION_RANGE}|{VERSION}))*"


class TokenType(enum.Enum):
    def __new__(cls, *args, **kwds):
        value = len(cls.__members__) + 1
        obj = object.__new__(cls)
        obj._value_ = value
        return obj

    def __init__(self, regex):
        self.regex = regex

    FILENAME = rf"({FILENAME})"
    VERSION = rf"@({VERSION_LIST})"

    UNEXPECTED = None


ALL_TOKENS = re.compile(
    "|".join(
        rf"(?P<{t.name}>({t.regex}))" for t in TokenType.__members__.values() if t.regex
    )
)

print(TokenType.FILENAME, id(TokenType.FILENAME))

print(ALL_TOKENS)
```

I think this puts most of the information you want to read in the `Enum` class *and* eliminates the need for maintainers to read regexes with named groups in them (which makes things harder to read).

The `__members__` dict preserves declaration order so you can still use the order to parse things correctly, and for tokens like `UNEXPECTED` that aren't associated with a particular parse, you can just indicate that with `None`.

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+
+import spack.error
+import spack.spec
+import spack.variant
+import spack.version
+
+#: Valid name for specs and variants. Here we are not using
+#: the previous "w[\w.-]*" since that would match most
+#: characters that can be part of a word in any language
+VALID_NAME_WITHOUT_DOTS = r"[a-zA-Z_0-9][a-zA-Z_0-9\-]*"
```
tgamblin: I think these are more readable within already complex regexes if they don't start with `VALID_`.  If we have to namespace them we could put `_RE` at the end to indicate they're regexes but I don't see why it's needed.

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+    rf"(?P<{TokenKind.COMPILER}>%({VALID_NAME}))",
+    # Filename
+    rf"(?P<{TokenKind.FILENAME}>({VALID_FILENAME}))",
+    # Spec name regexes
+    rf"(?P<{TokenKind.FULLY_QUALIFIED_PACKAGE_NAME}>{VALID_NAME_WITH_DOTS})",
+    rf"(?P<{TokenKind.UNQUALIFIED_PACKAGE_NAME}>{VALID_NAME_WITHOUT_DOTS})",
+    # DAG hash
+    rf"(?P<{TokenKind.DAG_HASH}>/({VALID_HASH}))",
+    # White spaces (the lowest priority)
+    rf"(?P<{TokenKind.WS}>\s+)",
```
tgamblin: See above -- I think most of this can be simplified

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+
+    Raises:
+        SpecTokenizationError: if we can't tokenize anymore, but didn't reach the
+            end of the input text.
+    """
+    scanner = MASTER_REGEX.scanner(text)  # type: ignore[attr-defined]
+    match: Optional[re.Match] = None
+    for match in iter(scanner.match, None):
+        yield Token(
+            STR_TO_TOKEN[match.lastgroup],  # type: ignore[attr-defined]
```
tgamblin: You can avoid `STR_TO_TOKEN` entirely:

```suggestion
            TokenKind.__members__[match.lastgroup],  # type: ignore[attr-defined]
```

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,458 @@
...
+    """Parse a single spec node from a stream of tokens"""
+
+    __slots__ = "ctx", "has_compiler", "has_version"
+
+    def __init__(self, ctx):
+        self.ctx = ctx
+        self.has_compiler = False
+        self.has_version = False
+
+    def parse(self, spec_buffer: spack.spec.Spec) -> spack.spec.Spec:
```
tgamblin: Can we call this something other than "buffer"?  A buffer always means raw bytes or chars to me -- and this thing is an object under construction.  Honestly just `spec` seems clear to me (esp. since the caller won't usually pass it).

tgamblin (changes requested):
This is looking great and it's clearly way faster.  Some simplifying suggestions below.

RE:
> We try to do that by adopting a different token granularity, and by using more complex regexes for tokenization. This allow us to a have a "flatter" encoding for the parser (i.e. with less nested conditionals):

I don't think the performance win here is actually the nesting, though it's related.  It's that you're compiling the regexes and delegating much more parsing responsibility to the regex engine.  So where the prior parser would loop several times in Python for each token, this parser is leveraging fast regex parsing to grab a bigger chunk on the C side and get it all unpacked in one iteration on the Python side. Makes a lot of sense and I think the regexes are very clear.

RE: differences:

> * No spaces allowed after a sigil (e.g. foo @ 1.2.3 is invalid while foo @1.2.3 is valid)

I don't see what the point is of enforcing this, and I don't know many language parsers that don't have optional whitespace between operators.  I think this is going to break private package repositories and scripts built on top of Spack, for not a lot of benefit.  So can you just add `\s*` to the relevant regexes?  I think with the simplifications I suggest below it'll still be readable.

> * /<hash> @1.2.3 can be parsed as a concrete spec followed by an anonymous spec (before was invalid)

The reason I implemented it this way initially is for the CLI use case:

```
(spackbook):spack> spack uninstall python@3.8
==> Error: python@3.8 matches multiple packages:

    -- darwin-bigsur-skylake / apple-clang@12.0.5 -------------------
    wx7u4y4 python@3.8.11  x24x4ec python@3.8.11

==> Error: You can either:
    a) use a more specific spec, or
    b) specify the spec by its hash (e.g. `spack uninstall /hash`), or
    c) use `spack uninstall --all` to uninstall ALL matching specs.

# Now you can hit up and just add the hash
> spack uninstall python/wx7u4y4 @3.8
```

Though now that I've written that I realize they could just do:

```
> spack uninstall python @3.8 /wx7u4y4
```

However, it still doesn't really sit well with me that these would be different:

```
python @3.8 /wx7u4y4  # one spec
python /wx7u4y4 @3.8  # two specs
```

I think that is pretty confusing.  There is no other place in the Spec grammar where anything but the *initial* absence of an identifier creates an anonymous spec, and the only way you could initiate a new spec on the CLI was to type a name or a `^` for a dependency constraint.  Now, hashes are special and always terminate the current spec parse, while no other sigil does that.

I think it's better without that inconsistency.

lib/spack/spack/parser.py
alalazo: I think I'll go with `JOINED_REGEX`, which hopefully reminds of the `"|".join`

tgamblin: I think it's better to keep it semantic -- ALL_TOKENS tells you what it does but JOINED_REGEX doesn't.

lib/spack/spack/parser.py
alalazo: Not that I know of. I'll try to double check, but here the issue is that `scanner` is not "public" and I need to explain `mypy` that we'll never reach these lines with a `match` set to `None`.

lib/spack/spack/parser.py
alalazo: I'll go with `initial_spec` here. Let me know if that is fine.

lib/spack/spack/parser.py
alalazo: Done in 3049bd2

lib/spack/spack/parser.py
alalazo: Done in d5f0dd3b8f9fe8da132a0e954c815bd9c2dcd348

lib/spack/spack/parser.py
alalazo: Done in 75cc896e1ea4d90452bc59dcf2e382282823aa3e

lib/spack/spack/parser.py
alalazo: Done in c3e8f0734633503222f1df94c9513e55becd8c18

lib/spack/spack/parser.py
alalazo: Done in 0b94c97

lib/spack/spack/parser.py
alalazo: Done in 0b94c97

lib/spack/spack/parser.py
alalazo: Here I did everything except:

> [ ... ] for tokens like UNEXPECTED that aren't associated with a particular parse, you can just indicate that with None.

The idea of that token is that it can be extended later, i.e. we have a valid set of regexes and we might add common error patterns afterwards - in following PRs. When we match an error pattern we can emit a more meaningful message, like "The `@` sigil cannot be repeated multiple times" and point to the actual parsing error, since each token has a `start` and an `end` position in the text stream.

alalazo:
> I don't see what the point is of enforcing this, and I don't know many language parsers that don't have optional whitespace between operators. 

@tgamblin I tried adding this back, but it will complicate the parsing due to version ranges:
```
@ 1.2.3               :                  develop  # This is a version range 1.2.3:develop
@ 1.2.3               :                  develop=foo  # This is a version range 1.2.3: followed by a key-value pair
```
so I would be for keeping the parser as is in this PR, since it's true it may break private packages - but in a way that is fixable and backward compatible. Also I would say that `@` (or `+` or `~` etc.) is not an "operator" per se, since it doesn't combine two homogeneous factors[^1]. To me it's the entire `@1.2.3:develop` to be a "version symbol".

Example where spaces could be allowed, but are not 
```python
>>> def foo      :
  File "<stdin>", line 1
    def foo      :
                 ^
SyntaxError: invalid syntax
```

[^1]: Not sure I am expressing myself in a formally correct way, but I mean that we don't have rules like `expr = expr @ term`. Rather we have an optional "version symbol" that might appear in each spec node.

alalazo:
> There is no other place in the Spec grammar where anything but the initial absence of an identifier creates an anonymous spec, and the only way you could initiate a new spec on the CLI was to type a name or a ^ for a dependency constraint. Now, hashes are special and always terminate the current spec parse, while no other sigil does that. I think it's better without that inconsistency.

Restored the old behavior for hashes in 624cb43

lib/spack/spack/parser.py
alalazo: Moved EBNF grammar to `spack/parser.py` in d40379a

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,512 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+"""Parser for spec literals
+
+Here is the EBNF grammar for a spec::
```
tgamblin, line 7: probably this grammar should correspond to the one you have outlined below.... with separate symbols for propagated/non-propagated variables

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,512 @@
...
+import spack.error
+import spack.spec
+import spack.variant
+import spack.version
+
+#: Valid name for specs and variants. Here we are not using
+#: the previous "w[\w.-]*" since that would match most
+#: characters that can be part of a word in any language
+PACKAGE_NAME_WITHOUT_DOTS = r"[a-zA-Z_0-9][a-zA-Z_0-9\-]*"
+PACKAGE_NAME_WITH_DOTS = rf"{PACKAGE_NAME_WITHOUT_DOTS}(\.{PACKAGE_NAME_WITHOUT_DOTS})+"
```
tgamblin: I'd call these `IDENTIFIER` and `DOTTED_IDENTIFIER` since they're used for more than just packages.

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,512 @@
...
+GIT_VERSION = rf"(((git.)?({PACKAGE_NAME_WITH_DOTS}))|({PACKAGE_NAME_WITHOUT_DOTS}))"
```
tgamblin: I believe the only scenario where we accept a git version *without* the `git.` prefix is when it is a hash, which is exactly 40 hex digits.  That should be easy to parse (`[A-Fa-f0-9]{40}`) right?  I don't think we started accepting short hashes did we?

lib/spack/spack/parser.py
```diff
@@ -0,0 +1,512 @@
...
+HASH = r"[a-zA-Z_0-9]+"
+
+#: A filename starts either with a "." or a "/" or a "{name}/"
+FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
+
+VALUE = r"([a-zA-Z_0-9\-+\*.,:=\~\/\\]+)"
+QUOTED_VALUE = r"[\"']+([a-zA-Z_0-9\-+\*.,:=\~\/\\\s]+)[\"']+"
+
+VERSION = r"([a-zA-Z0-9_][a-zA-Z_0-9\-\.]*)"
+VERSION_RANGE = rf"({VERSION}:{VERSION}|:{VERSION}|{VERSION}:|:)"
```
tgamblin: I believe you can avoid the ambiguity with KVP's with:

```suggestion
VERSION_RANGE = rf"({VERSION}\s*:\s*{VERSION}(?!\s*=)|:\s*{VERSION}(?!\s*=)|{VERSION}\s*:|:)"
```
which should force this to only match up to the first colon if the second version is followed by `=` (and also to ignore other internal whitespace)

tgamblin (changes requested):

lib/spack/spack/parser.py
alalazo: Done in b638e86 As we discussed offline the key was to add `\b` to the `VERSION` regex Modifications to packages are reverted in 5952536 Now the PR should be 100% backward compatible.

lib/spack/spack/parser.py
alalazo: Done in c2f1ad1

lib/spack/spack/parser.py
alalazo: Done in a303c8a

tgamblin (approved):

tgamblin:
@alalazo thanks for getting this to 100% compatibility!  It looks great to me.

PR #34152, opened 2022-11-28
binary distribution: warn about issues

haampie:
- binary_distribution: fix warning message
- binary_distribution: at least warn about fetch issues of binary cache

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -504,9 +504,9 @@ class BinaryCacheIndex(object):
 
         if fetched_hash is not None and locally_computed_hash != fetched_hash:
             msg = (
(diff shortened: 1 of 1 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -504,9 +509,9 @@ def _fetch_and_cache_index(self, mirror_url, expect_hash=None):
 
         if fetched_hash is not None and locally_computed_hash != fetched_hash:
             msg = (
-                "Computed hash ({0}) did not match remote ({1}), "
+                "Computed index hash [{0}] did not match remote [{1}, url:{2}] "
```
scottwittenburg, line 512: Agreed it will be good to know the mirror where the hash doesn't match.

scottwittenburg (approved):

PR #34153, opened 2022-11-28
Revert "Revert "gitlab: Add shared PR mirror to places pipelines look for binaries.""

haampie:
Long story short: we need the mirror, but we have to fix the issue where `index.json.hash` is incorrect:

1. Could be due to a race (both pushing and pulling is not atomic...)
2. Could be due to successful upload of index.json but unsuccessful upload of index.json.hash

To be fair, I don't get the point of index.json.hash:

1. It leads to races
2. For http or s3 protocols: Etag is the way to handle cache invalidation
3. For file protocols: you can just compute the content hash instead
4. Failure to validate index.json against index.json.hash leads to downloading the index every time spack install is called
5. For some reason sha256 is used, which is cryptographically secure, but clearly this is not about improved security: the index.json.hash file lives on the same host, anybody who can tamper with index.json can also change the hash.
 
Therefore I would say it's better to drop index.json.hash entirely.

Reverts spack/spack#34087

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -49,6 +49,7 @@ JOB_RETRY_CONDITIONS = [
 
 TEMP_STORAGE_MIRROR_NAME = "ci_temporary_mirror"
 SPACK_RESERVED_TAGS = ["public", "protected", "notary"]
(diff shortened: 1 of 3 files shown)


alalazo (approved):

haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

PR #34157, opened 2022-11-28
Deduplicate code to propagate module changes across MRO

alalazo:
Follow up from #34059 

Modifications:
- [x] Use the same object to setup `make_jobs` etc.

### DIFF
--- a/lib/spack/spack/test/build_environment.py
+++ b/lib/spack/spack/test/build_environment.py
@@ -436,20 +436,13 @@ dt-diamond-left:
     assert not (set(os.path.normpath(x) for x in link_dirs[:-2]) & external_lib_paths)
 
 
-def test_parallel_false_is_not_propagating(config, mock_packages):
-    class AttributeHolder(object):
-        pass
-
-    # Package A has parallel = False and depends on B which instead
-    # can be built in parallel
-    s = spack.spec.Spec("a foobar=bar")
-    s.concretize()
-
+def test_parallel_false_is_not_propagating(default_mock_concretization):
+    # Package A has parallel = False and depends on B which instead can be built in parallel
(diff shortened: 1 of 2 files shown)


haampie (approved):

haampie:
Better!

lib/spack/spack/test/build_environment.py
```diff
@@ -436,20 +436,13 @@ def _trust_me_its_a_dir(path):
...
-
+def test_parallel_false_is_not_propagating(default_mock_concretization):
+    # Package A has parallel = False and depends on B which instead can be built in parallel
+    s = default_mock_concretization("a foobar=bar")
     for spec in s.traverse():
-        expected_jobs = spack.config.get("config:build_jobs") if s.package.parallel else 1
-        m = AttributeHolder()
-        spack.build_environment._set_variables_for_single_module(s.package, m)
-        assert m.make_jobs == expected_jobs
+        expected_jobs = spack.build_environment.determine_number_of_jobs(spec.package.parallel)
```
haampie: This test should be written differently though, cause it's testing whether a property on the parent package is NOT set on the child (for whatever reason...), but now it takes the the package properties  as input while verifying that

haampie (changes requested):

lib/spack/spack/test/build_environment.py
alalazo: I think the test was wrong before. If you look, we were iterating over nodes, but never using the nodes - we were always checking the result on the root.

haampie: how about not iterating at all, it's really just two packages a and b

haampie (approved):

PR #34160, opened 2022-11-28
root: new versions 6.26.08 and 6.26.10

wdconinc:
Bugfixes only, https://root.cern/doc/v626/release-notes.html#bugs-and-issues-fixed-in-this-release-4.

### DIFF
--- a/var/spack/repos/builtin/packages/root/package.py
+++ b/var/spack/repos/builtin/packages/root/package.py
@@ -31,6 +31,8 @@ class Root(CMakePackage):
     # Development version (when more recent than production).
 
     # Production version
+    version("6.26.10", sha256="8e56bec397104017aa54f9eb554de7a1a134474fe0b3bb0f43a70fc4fabd625f")
+    version("6.26.08", sha256="4dda043e7918b40743ad0299ddd8d526b7078f0a3822fd06066df948af47940e")
     version("6.26.06", sha256="b1f73c976a580a5c56c8c8a0152582a1dfc560b4dd80e1b7545237b65e6c89cb")
     version("6.26.04", sha256="a271cf82782d6ed2c87ea5eef6681803f2e69e17b3036df9d863636e9358421e")
     version("6.26.02", sha256="7ba96772271a726079506c5bf629c3ceb21bf0682567ed6145be30606d7cd9bb")


tldahlgren (approved):
Confirmed both new sha256.

tldahlgren:
Will defer merge decision to give one or more maintainer a chance to comment.

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -31,6 +31,8 @@ class Root(CMakePackage):
     # Development version (when more recent than production).
 
     # Production version
+    version("6.26.10", sha256="8e56bec397104017aa54f9eb554de7a1a134474fe0b3bb0f43a70fc4fabd625f")
+    version("6.26.08", sha256="4dda043e7918b40743ad0299ddd8d526b7078f0a3822fd06066df948af47940e")
```
HadrienG2, line 35: If I understand the release notes correctly, [you need 6.26.08+ to build with nlohmann_json 3.11+](https://github.com/root-project/root/issues/11312). The dependency on nlohmann_json should probably be updated accordingly.

HadrienG2: Similarly, clang 16 will have issues with older root : https://github.com/root-project/root/issues/11128 .

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -31,6 +31,8 @@ class Root(CMakePackage):
     # Development version (when more recent than production).
 
     # Production version
+    version("6.26.10", sha256="8e56bec397104017aa54f9eb554de7a1a134474fe0b3bb0f43a70fc4fabd625f")
```
HadrienG2, line 34: Need at least this release for python 3.11+ : https://github.com/root-project/root/issues/11713 .

HadrienG2: Similar for macOS 13 (though I'm not sure if we can set OS version requirements or conflicts in Spack) : https://github.com/root-project/root/issues/11714

HadrienG2 (changes requested):
While bugfixes may seem harmless, they do introduce support for newer dependency versions, and thus implicitly state that older versions are incompatible with older dependency versions.

This is something we care about in Spack, because people may want to build older ROOT releases on newer systems and it's better if we pick the right dependency versions by default without further intervention.

wdconinc:
I'm going to close this PR and let someone else add the necessary version restrictions. I don't have the time until 2023 to dig through the ROOT PRs to address and test these version incompatibilities.

var/spack/repos/builtin/packages/root/package.py
wdconinc: I am wondering, though, whether from a community contribution workflow perspective it doesn't make sense to model these issues with upper limits on dependencies for older version. I.e. we could use `depends_on("python@:3.10")` until someone can verify that 3.11 works with a certain version. At that point they can add `depends_on("python@:3.11")` and modify `depends_on("python@:3.10", when=":6.26.06")`.

Even though I currently don't have the time to figure out the dependencies (and test it, that's what takes most of the time), this still leaves the current package with the same issue: the most recent ROOT version is not working with the most recent python version (not the preferred version).

PR #34162, opened 2022-11-28
py-bidskit: new package and dcm2niix: add 1.0.20220720

manuelakuhn:
bidskit sources from pypi (no tag on github): https://files.pythonhosted.org/packages/10/22/0c5fc8932a78c4a975afd44d04653ad530ccfa8f7b7cae6d09d608c5a5f8/bidskit-2022.10.13.tar.gz

https://github.com/rordenlab/dcm2niix/tree/v1.0.20220720

### DIFF
--- a/var/spack/repos/builtin/packages/dcm2niix/package.py
+++ b/var/spack/repos/builtin/packages/dcm2niix/package.py
@@ -11,8 +11,12 @@ class Dcm2niix(CMakePackage):
     """DICOM to NIfTI converter"""
 
     homepage = "https://github.com/rordenlab/dcm2niix"
-    url = "https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20210317.tar.gz"
+    url = "https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20220720.tar.gz"
+    list_url = "https://github.com/rordenlab/dcm2niix/tags"
 
+    version(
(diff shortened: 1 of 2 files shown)


manuelakuhn:
ping @adamjstewart

var/spack/repos/builtin/packages/dcm2niix/package.py
```diff
@@ -11,8 +11,12 @@ class Dcm2niix(CMakePackage):
     """DICOM to NIfTI converter"""
 
     homepage = "https://github.com/rordenlab/dcm2niix"
-    url = "https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20210317.tar.gz"
+    url = "https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20220720.tar.gz"
+    list_url = "https://github.com/rordenlab/dcm2niix/tags"
```
adamjstewart: @tldahlgren did something change in `spack checksum` recently that broke support for GitHub URLs? This used to work automatically.

I would remove this change here and open an issue to track this, we should be able to fix this for all GitHub URLs simultaneously.

manuelakuhn: I was not aware that this should work automatically. I have opened an issue and removed `list_url` here.
Thanks.

adamjstewart (approved):

PR #34166, opened 2022-11-28
rudimentary jobserver support

haampie:
When Spack is run under `make -j <N>` it should optimally not blindly
spin up `min(32, nproc)` threads, but respect the global bound `N`.

This commit adds basic support: detect a GNU jobserver, and wrap the
threadpool operations with acquiring and releasing jobs from the
jobserver. It's basic in the sense that we try to take as many jobs as
available, or, when we can't because another spack process took all,
we continue single-threaded.

It's still not terribly efficient since we don't dynamically retrieve more
jobs when they become available; it's static & ahead of time.


haampie:
Dropping this, since `make --load-average` is the alternative

PR #2196, opened 2016-11-01
Bugfix in concretize_compiler

js947:
The loop at `lib/spack/spack/concretize.py:389` was not checking all of the elements in the list.

PR #34167, opened 2022-11-28
cmake fails with icpc warning

stephenmsachs:
The Intel compiler isn't able to deal with noinline member functions of template classses defined in headers.  As such it outputs
   warning #2196: routine is both "inline" and "noinline"
cmake bootstrap will fail due to the word 'warning'.

See spack/var/spack/repos/builtin/packages/protobuf/intel-v2.patch for reference.


var/spack/repos/builtin/packages/cmake/package.py
```diff
@@ -312,6 +312,13 @@ def bootstrap_args(self):
         args = []
         self.generator = make
 
+        # The Intel compiler isn't able to deal with noinline member functions of
+        # template classses defined in headers.  As such it outputs
+        #   warning #2196: routine is both "inline" and "noinline"
+        # cmake bootstrap will fail due to th word 'warning'.
+        if spec.satisfies("%intel@:2021.6.0"):
+            args.append("CXXFLAGS=-diag-disable=2196")
```
alalazo:
```suggestion
        # The Intel compiler isn't able to deal with noinline member functions of
        # template classes defined in headers.  As such it outputs
        #   warning #2196: routine is both "inline" and "noinline"
        # cmake bootstrap will fail due to the word 'warning'.
        if spec.satisfies("%intel@:2021.6.0"):
            args.append("CXXFLAGS=-diag-disable=2196")
```

Pinging @rscohn2 to confirm that's the best way to solve this

rscohn2: It seems ok. What about icc 2021.7.0, 2021.7.1, & 2021.8.0? They don't have the problem?

stephenmsachs: No, they don't issue this warning anymore:
```
$~: compiler=/shared/spack/opt/spack/linux-amzn2-x86_64_v3/gcc-12.2.0/intel-oneapi-compilers-2022.2.0-uqvb2553zy5toeapvoopacndd27x6p5m/compiler/2022.2.0/linux/bin/intel64/icpc
$~: $compiler unique.c 
icpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.
```

stephenmsachs:
Seems like I broke something by merging develop. I'll open a new PR.

PR #34168, opened 2022-11-28
py-bidscoin, py-multiecho: add new packages

manuelakuhn:
https://github.com/Donders-Institute/bidscoin/tree/3.7.4
https://github.com/Donders-Institute/multiecho/tree/0.28

### DIFF
--- a/var/spack/repos/builtin/packages/py-multiecho/package.py
+++ b/var/spack/repos/builtin/packages/py-multiecho/package.py
@@ -1,0 +1,22 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyMultiecho(PythonPackage):
+    """Combine multi-echoes from a multi-echo fMRI acquisition."""
+
+    homepage = "https://github.com/Donders-Institute/multiecho"
+    pypi = "multiecho/multiecho-0.28.tar.gz"
+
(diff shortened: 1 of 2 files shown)


manuelakuhn:
ping @adamjstewart

tldahlgren (commented):
Confirmed sha256 for both packages

var/spack/repos/builtin/packages/py-multiecho/package.py
```diff
@@ -0,0 +1,22 @@
...
+
+class PyMultiecho(PythonPackage):
+    """Combine multi-echoes from a multi-echo fMRI acquisition."""
+
+    homepage = "https://github.com/Donders-Institute/multiecho"
+    pypi = "multiecho/multiecho-0.28.tar.gz"
+
+    version("0.28", sha256="d0459bd03398547116d8e989b2d2b7922af0ae7ae77e233794dd7253a2abced3")
+
+    depends_on("python@3.6:3.9", type=("build", "run"))
```
adamjstewart, line 17: I don't see any `python_requires` so the supported versions of Python should be unrestricted, I would remove this line.

manuelakuhn: The python restriction I got from here:
https://github.com/Donders-Institute/multiecho/blob/0.28/pyproject.toml#L11

adamjstewart: Hmm, I'm not actually sure if setuptools reads that part of the file, but let's keep it for now.

adamjstewart (approved):

manuelakuhn:
@spackbot run pipeline

PR #34169, opened 2022-11-28
omegah: new scorec version, fix cuda flags

cwsmith:
This PR adds the newest SCOREC version of Omega_h and fixes an issue passing the CUDA architecture flags to CMake.


tldahlgren (approved):
Confirmed the new version commit (via `spack fetch` due to inclusion of `git`).  Interesting check of the version.

PR #34172, opened 2022-11-28
acfl: add post-installation check by running examples

annop-w:
Check installation by running examples and compare results
with expected ones. The makefile is included as part of the
package.

### DIFF
--- a/var/spack/repos/builtin/packages/acfl/package.py
+++ b/var/spack/repos/builtin/packages/acfl/package.py
@@ -276,3 +276,12 @@ class Acfl(Package):
         env.prepend_path("PATH", join_path(arm_dir, "bin"))
         env.prepend_path("CPATH", join_path(arm_dir, "include"))
         env.prepend_path("MANPATH", join_path(arm_dir, "share", "man"))
+
+    @run_after("install")
+    def check_install(self):
+        armpl_dir = get_armpl_prefix(self.spec)
+        armpl_example_dir = join_path(armpl_dir, "examples")
+        # run example makefile
+        make("-C", armpl_example_dir, "CC=" + self.cc, "F90=" + self.fortran)
+        # clean up
+        make("-C", armpl_example_dir, "clean")


var/spack/repos/builtin/packages/acfl/package.py
```diff
@@ -131,6 +131,8 @@ class Acfl(Package):
     # Build Versions: establish OS for URL
     acfl_os = get_os()
 
+    test_requires_compiler = True
+
```
tldahlgren: This property is only used for stand-alone, post-installation tests that are run days to weeks after the installation (see https://spack.readthedocs.io/en/latest/packaging_guide.html#enabling-test-compilation).  

What you are adding here is an always on, build-time check (see https://spack.readthedocs.io/en/latest/packaging_guide.html#adding-installation-phase-tests).

var/spack/repos/builtin/packages/acfl/package.py
```diff
@@ -276,3 +278,12 @@ def setup_run_environment(self, env):
         env.prepend_path("PATH", join_path(arm_dir, "bin"))
         env.prepend_path("CPATH", join_path(arm_dir, "include"))
         env.prepend_path("MANPATH", join_path(arm_dir, "share", "man"))
+
+    @run_after("install")
+    def check_install(self):
```
tldahlgren, line 281: If you want this to only run when passing `--test` option (see https://spack.readthedocs.io/en/latest/packaging_guide.html#adding-installation-phase-test)), you'll want to add the package attributes directive before the method definition.  There is an example after the table at the link above.

```suggestion
    @on_package_attributes(run_tests=True)
    def check_install(self):
```

annop-w: I had tried adding `@on_package_attributes(run_tests=True)`, but it does not work when passing `--test` option. So, for now I would like this to always run post-installation.

var/spack/repos/builtin/packages/acfl/package.py
annop-w: This was left here by mistake from when I was experimenting. I am going to remove this line.

var/spack/repos/builtin/packages/acfl/package.py
tldahlgren: Did you pass it as `--test=root`?

annop-w: Yes. I observed the same behavior with some other packages as well. Perhaps I can try to reproduce it again and raise an issue.

tldahlgren: Up to you.  If you want the tests to run after every install and, ideally, they run quickly, then we could leave this as-is.

You could also consider switching the checks to be stand-alone, post-install checks as described https://spack.readthedocs.io/en/latest/packaging_guide.html#stand-alone-or-smoke-tests.  Then they would only be performed -- against the installed software -- when someone runs e.g. `spack test run acfl`.

annop-w: These tests run rather quickly. So, I would like to leave it as-is. I will find time to try to reproduce the problem I saw with `@on_package_attributes(run_tests=True)` and raise an issue, if I can reproduce. 

I had considered making the stand-alone check as well (hence the leftover line above), but that is perhaps for another commit. For now, I will just push a new commit to remove the line above.

var/spack/repos/builtin/packages/acfl/package.py
tldahlgren: Ping me when this is removed.

annop-w: Done.

tldahlgren (approved):
I was not able to install and test on my local machine but LGTM.

Issue #29168, opened 2022-02-24
Installation issue: binutils


wspear:
### Steps to reproduce the issue

```console
$ spack spec -I binutils+libiberty
Input spec
--------------------------------
 -   binutils+libiberty

Concretized
--------------------------------
 -   binutils@2.38%gcc@9.3.0~gas~gold~headers~interwork~ld+libiberty~lto+nls+plugins libs=shared,static arch=linux-ubuntu20.04-westmere
 -       ^diffutils@3.7%gcc@9.3.0 arch=linux-ubuntu20.04-westmere
[+]      ^gettext@0.21%gcc@9.3.0+bzip2+curses+git~libunistring+libxml2+tar+xz arch=linux-ubuntu20.04-westmere
[+]          ^bzip2@1.0.8%gcc@9.3.0~debug~pic+shared arch=linux-ubuntu20.04-westmere
[+]          ^libiconv@1.16%gcc@9.3.0 libs=shared,static arch=linux-ubuntu20.04-westmere
[+]          ^libxml2@2.9.12%gcc@9.3.0~python arch=linux-ubuntu20.04-westmere
[+]              ^pkgconf@1.8.0%gcc@9.3.0 arch=linux-ubuntu20.04-westmere
[+]              ^xz@5.2.4%gcc@9.3.0~pic libs=shared,static arch=linux-ubuntu20.04-westmere
[+]              ^zlib@1.2.11%gcc@9.3.0+optimize+pic+shared arch=linux-ubuntu20.04-westmere
[+]          ^ncurses@6.2%gcc@9.3.0~symlinks+termlib abi=none arch=linux-ubuntu20.04-westmere
[+]          ^tar@1.34%gcc@9.3.0 arch=linux-ubuntu20.04-westmere
 -       ^texinfo@6.7%gcc@9.3.0 patches=12f6edb,1732115 arch=linux-ubuntu20.04-westmere
```
This works with binutils@2.37 and without +libiberty

### Error message

<details><summary>Error message</summary><pre>
==> binutils: Executing phase: 'install'
==> Error: ProcessError: Command exited with status 2:
    'make' '-j16' 'install'

9 errors found in build log:
     2983    checking for O_CLOEXEC... /bin/bash ./libtool  --tag=CC   --mode=compile /home/wspear/bin/SPACK/spack/lib/spack/env/gcc/gcc -DHAVE_CONFIG_H -I.  -I. -I. -I../bfd -I./../include -I./../bfd
[...]

PR #26994, opened 2021-10-27
m4: (@1.4.19) Fix not running texinfo for m4.info

jrood-nrel:
m4 had recently not been building for me on Summit due to:
```
     2619    /gpfs/alpine/yadda/yadda/stage/spack-stage-m4-1.4.19-65qjzp2gyiunr7mmf233e3giax6phee4/spack-src/build-aux/missing: line 81: makeinfo: command not found
  >> 2620    WARNING: 'makeinfo' is missing on your system.
     2621             You should only need it if you modified a '.texi' file, or
     2622             any other file indirectly affecting the aspect of the manual.
     2623             You might want to install the Texinfo package:
     2624             <https://www.gnu.org/software/texinfo/>
     2625             The spurious makeinfo call might also be the consequence of
     2626             using a buggy 'make' (AIX, DU, IRIX), in which case you might
```
This change solves it for me. Not sure if there is any better fix or if I'm missing something else.
@eugeneswalker

PR #34173, opened 2022-11-28
Add texinfo dependency for binutils through version 2.38

jrood-nrel:
Personally I think it should just be `depends_on("texinfo", type="build")` to be safe. However, I still see:
```
  >> 3066    WARNING: 'makeinfo' is missing on your system.
     3067             You should only need it if you modified a '.texi' file, or
     3068             any other file indirectly affecting the aspect of the manual.
     3069             You might want to install the Texinfo package:
     3070             <http://www.gnu.org/software/texinfo/>
     3071             The spurious makeinfo call might also be the consequence of
     3072             using a buggy 'make' (AIX, DU, IRIX), in which case you might
     3073             want to install GNU make:
     3074             <http://www.gnu.org/software/make/>
```
in 2.38 so I think `texinfo` should be required through 2.38. The build also fails on my machine without `texinfo` installed, and this change allows the build to succeed.

### DIFF
--- a/var/spack/repos/builtin/packages/binutils/package.py
+++ b/var/spack/repos/builtin/packages/binutils/package.py
@@ -81,12 +81,9 @@ class Binutils(AutotoolsPackage, GNUMirrorPackage):
     depends_on("m4", type="build", when="@:2.29 +gold")
     depends_on("bison", type="build", when="@:2.29 +gold")
 
-    # 2.38 with +gas needs makeinfo due to a bug, see:
-    # https://sourceware.org/bugzilla/show_bug.cgi?id=28909
-    depends_on("texinfo", type="build", when="@2.38 +gas")
-    # 2.34 needs makeinfo due to a bug, see:
+    # 2.34:2.38 needs makeinfo due to a bug, see:
     # https://sourceware.org/bugzilla/show_bug.cgi?id=25491
-    depends_on("texinfo", type="build", when="@2.34")
+    depends_on("texinfo", type="build", when="@2.34:2.38")
 
     conflicts("+gold", when="platform=darwin", msg="Binutils cannot build linkers on macOS")
 


mwkrentel:
> Personally I think it should just be depends_on("texinfo", type="build") to be safe.

IMHO, I disagree.  The intent of binutils has always been that
makeinfo is only needed if you're modifying the .texi files.  That is,
not needed in a one-time build like spack.  But in 2.34, somebody made
a boo-boo and broke this rule.

From looking at the bug report in

https://sourceware.org/bugzilla/show_bug.cgi?id=25491

it says it is fixed on 2020-05-25 which would put it in 2.35, released
on 2020-07-24.

I guess it somehow still breaks on your system?
Can you post the error message and/or build log?
Also, what platform, etc are you using?

What you have posted above is not an error.  That's a harmless warning
that says you won't be able to modify the .texi files and rebuild.
But the spack build doesn't do that.

eugeneswalker:
This has been encountered by a few people over the past year. I'm not sure what was the ultimate resolution.

Here is one example: (CC @wspear)
* https://github.com/spack/spack/issues/29168

I know I have hit this before when building in minimal container environments.

jrood-nrel:
I will reproduce my error to give the actual error. In the past I have done this for m4 https://github.com/spack/spack/pull/26994 . Why do we need to avoid `texinfo` as a dependency since we already need it for 2.34? I've encountered this issue so many times. I think I would rather have Spack be robust than optimize the DAG.

jrood-nrel:
```
perl ./../etc/texi2pod.pl -I "./doc" -I doc -I "./../libiberty" -I "./../bfd/doc" -I ../bfd/doc --no-split -Dman < ./doc/as.texi > doc/as.pod
restore=: && backupdir=".am$$" && \
rm -rf $backupdir && mkdir $backupdir && \
if (/tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src/missing makeinfo --split-size=5000000 --split-size=5000000 --version) >/dev/null 2>&1; then \
  for f in doc/as.info doc/as.info-[0-9] doc/as.info-[0-9][0-9] doc/as.i[0-9] doc/as.i[0-9][0-9]; do \
    if test -f $f; then mv $f $backupdir; restore=mv; else :; fi; \
  done; \
else :; fi && \
if /tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src/missing makeinfo --split-size=5000000 --split-size=5000000 -I "./doc" -I doc -I "./../libiberty" -I "./../bfd/doc" -I ../bfd/doc --no-split  -I doc -I ./doc \
 -o doc/as.info `test -f 'doc/as.texi' || echo './'`doc/as.texi; \
then \
  rc=0; \
else \
  rc=$?; \
  $restore $backupdir/* `echo "./doc/as.info" | sed 's|[^/]*$||'`; \
fi; \
rm -rf $backupdir; exit $rc
/tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src/missing: line 81: makeinfo: command not found
WARNING: 'makeinfo' is missing on your system.
         You should only need it if you modified a '.texi' file, or
         any other file indirectly affecting the aspect of the manual.
         You might want to install the Texinfo package:
         <http://www.gnu.org/software/texinfo/>
         The spurious makeinfo call might also be the consequence of
         using a buggy 'make' (AIX, DU, IRIX), in which case you might
         want to install GNU make:
         <http://www.gnu.org/software/make/>
(pod2man --center="GNU Development Tools" --release="binutils-2.38" --section=1 doc/as.pod | \
        sed -e '/^.if n .na/d' > doc/as.1.T$$ && \
        mv -f doc/as.1.T$$ doc/as.1) || \
        (rm -f doc/as.1.T$$ && exit 1)
make[5]: *** [Makefile:1466: doc/as.info] Error 127
make[5]: *** Waiting for unfinished jobs....
rm -f doc/as.pod
make[5]: Leaving directory '/tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src/gas'
make[4]: *** [Makefile:1648: all-recursive] Error 1
make[4]: Leaving directory '/tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src/gas'
make[3]: *** [Makefile:994: all] Error 2
make[3]: Leaving directory '/tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src/gas'
make[2]: *** [Makefile:5428: all-gas] Error 2
make[2]: Leaving directory '/tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src'
make[1]: *** [Makefile:1000: all] Error 2
make[1]: Leaving directory '/tmp/jrood/spack-stage/spack-stage-binutils-2.38-cv4leyeromzgivv2qmtiwbo7t6idktgw/spack-src'
```

jrood-nrel:
`binutils+gold` will succeed on this machine, but regular `binutils` will fail without the `texinfo` dependency.

mwkrentel:
I believe that makeinfo is getting a bad rap here.  I know that I've
built binutils on machines that don't have texinfo installed, so
that's not the real problem.

True, adding texinfo covers up the problem, but I'd rather figure out
the root cause of the problem and fix that.

Looking at the older PR #26994 for m4, it's starting to look like the
real problem is the time stamps on some files for some advanced
(remote) file systems.

The way binutils (and several other gnu packages) handle their docs is
they distribute the pre-built info files.  That is, they run makeinfo
at the factory and distribute the result in the tarball.  But they're
dependent on time stamps in the unpacked file system to be perfect.
And since make runs off time stamps, I suspect that's what's really
going on here.

I'll bet this problem only every happens on remotely mounted file
systems, never on a local disk.  Anyone have a counter example to this
claim?

Perhaps this could become a standard thing.  For old gnu packages that
treat info files this way, maybe we could whack the time stamps on
some key files that would fix make's detection of what does and does
not need to be rebuilt.

Either that or create a fake makeinfo that does nothing but is enough
not to cause make to fall over with "makeinfo: not found".  
It doesn't need to do anything since the info files are already built.

@alalazo @haampie What do you think?

bernhardkaindl (approved):
@mwkrentel: You are wrong (and sound arrogant too to me). I just got the `makeinfo` missing error on Ubuntu 20.04 running in a regular Linux arm64 VM, no remote filesystem at all on that machine, single host. 
```py
     2606    checking for makeinfo... /tmp/spack-stage/spack-stage-binutils-2.38-.../spack-src/missing makeinfo --split-size=5000000
     2607    /tmp/spack-stage/spack-stage-binutils-2.38-.../spack-src/missing:
              81: makeinfo: not found
  >> 2608    WARNING: 'makeinfo' is missing on your system.
     2609             You should only need it if you modified a '.texi' file, or
     2610             any other file indirectly affecting the aspect of the manual.
...
df -T /tmp
Filesystem     Type 1K-blocks     Used Available Use% Mounted on
/dev/sda1      ext4  47158992 35202496  11940112  75% /
```

mwkrentel:
> You are wrong (and sound arrogant too to me). I just got the makeinfo
> missing error on Ubuntu 20.04 running in a regular Linux arm64 VM, no
> remote filesystem at all on that machine, single host.

@bernhardkaindl  But did the build for binutils actually fail?
It's not a counter-example unless the build fails.

As I said before, what you've quoted is not an error, it's a harmless
warning that says makeinfo doesn't exist AND you don't need it.
This is common among various old gnu projects that use texinfo.

Sorry, I don't mean to be arrogant.  I know I've successfully built
binutils on systems without makeinfo, so that's not the problem.
I'm simply trying to find what the root cause really is.

IMO, this patch is the wrong fix and something along the lines of
#26994 is the right fix.

jrood-nrel:
My error occurs on a local SSD. I posted my full error above as well. Does anybody have a suggestion on how else to solve that specific error? `texinfo` has solved _all_ of my errors regarding missing `makeinfo` in the past.

psakievich:
I'll weigh in here.  It seems like to me that manually scrubbing the files is a perfectly valid solution, or just adding the dependency is also a valid solution.  Adding the dependency seems robust but it may be overkill. 

So the questions in my mind to resolve this are:
1) Is the intent of spack's scope to include making up for legacy packaging procedures?
2) It is not the minimal working solution to add `texinfo` as a dependency to `binutils`, but is the added robustness worth the overhead of the additional dependency?

bkmgit:
@jrood-nrel @bernhardkaindl @alalazo The problem is that the gas build copies a texinfo file and updates its timestamp. 

https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gas/doc/local.mk;h=f611a50913ca85b4218b11a66a724c23a7eff8c5;hb=HEAD#l44

contains

`$(AM_V_GEN)cp $(srcdir)/%D%/$(CONFIG).texi %D%/asconfig.texi`

this  can be changed with a patch to

`$(AM_V_GEN)cp -p $(srcdir)/%D%/$(CONFIG).texi %D%/asconfig.texi`

Though other solutions are possible and maybe better. Trying to get this resolved upstream

PR #34176, opened 2022-11-29
exago@develop %gcc@11.2.0: CXXFLAGS=-Wno-error=non-pod-varargs

eugeneswalker:
Building `exago@develop %gcc@11.2.0` without this fix I see these errors:
```
...
  >> 188    /tmp/eugeneswalker/spack-stage/spack-stage-exago-develop-6mgccbqrryevycnpjix2irtzi4mwuecw/spack-src/src/utils/utils.cpp:122:29: error: cannot pass
             non-trivial object of type 'const std::string' (aka 'const basic_string<char>') to variadic function; expected type from format string was '
            char *' [-Wnon-pod-varargs]
     189      sprintf(sbuf, "%s%s", opt.default_value,
```

@CameronRutherford @pelesh @ryandanehy

### DIFF
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -140,6 +140,10 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     flag_handler = build_system_flags
 
+    def setup_build_environment(self, env):
+        if self.spec.satisfies("@develop %gcc@11.2.0:"):
+            env.set("CXXFLAGS", "-Wno-error=non-pod-varargs")
+
     def cmake_args(self):
         args = []
         spec = self.spec


pelesh:
@eugeneswalker, I am guessing here, but I believe adding `.c_str()` in ExaGO code would fix this issue. The line 122 probably should look like:
```c++
  sprintf(sbuf, "%s%s", opt.default_value.c_str(),
          (opt.possible_values.size() ? "|" : ""));
```
I think it is better if we fix this issue in ExaGO code.

eugeneswalker:
> @eugeneswalker, I am guessing here, but I believe adding `.c_str()` in ExaGO code would fix this issue. The line 122 probably should look like:
> 
> ```c++
>   sprintf(sbuf, "%s%s", opt.default_value.c_str(),
>           (opt.possible_values.size() ? "|" : ""));
> ```

Yes, I just tried a new build, and with your recommended change there is no need for `-Wno-error=non-pod-varargs`.

Thank you!

cameronrutherford:
This reflects our lack of configuration for logging in exago at the moment. Will be transitioning our dependency of spdlog to spack instead of submodules, and currently logging is disabled by default for cmake.

Patch for exago package.py would be to add `exago+logging` that `depends_on(spdlog)` when `+logging` and enabled CMake variable `EXAGO_ENABLE_LOGGING` would fix this for tests, and then we can tag this feature in the next release.

pelesh:
I added the [issue](https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/issues/397) to ExaGO's GitLab so we can track it there.

pelesh:
I am closing this PR because we have a fix in ExaGO in preparation. It will be merged to `develop` branch in a day or two. Feel free to reopen, if you think the workaround proposed here is still needed.

eugeneswalker:
Sounds good, thanks!

PR #34177, opened 2022-11-29
Cray manifest: automatically convert 'cray' platform to 'linux'

scheibelp:
Spack now recognizes Cray systems as being Linux (i.e. `spack arch` on Cray systems will list "linux" as the platform). For manifests which continue to list the platform as Cray, this PR automatically converts them so they are considered compatible with other packages built on the system by Spack.

### DIFF
--- a/lib/spack/spack/cray_manifest.py
+++ b/lib/spack/spack/cray_manifest.py
@@ -61,9 +61,17 @@ def compiler_from_entry(entry):
 def spec_from_entry(entry):
     arch_str = ""
     if "arch" in entry:
+        local_platform = spack.spec.ArchSpec.default_arch().platform
+        spec_platform = entry["arch"]["platform"]
+        # Note that Cray systems are now treated as Linux. Specs
+        # in the manifest which specify "cray" as the platform
+        # should be registered in the DB as "linux"
+        if local_platform == "linux" and spec_platform.lower() == "cray":
+            spec_platform = "linux"
+
         arch_format = "arch={platform}-{os}-{target}"
         arch_str = arch_format.format(
-            platform=entry["arch"]["platform"],
+            platform=spec_platform,
             os=entry["arch"]["platform_os"],
             target=entry["arch"]["target"]["name"],
         )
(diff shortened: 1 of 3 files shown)


lib/spack/spack/cray_manifest.py
```diff
@@ -61,9 +61,17 @@ def compiler_from_entry(entry):
 def spec_from_entry(entry):
     arch_str = ""
     if "arch" in entry:
+        local_platform = spack.spec.ArchSpec.default_arch().platform
```
becker33:
```suggestion
        local_platform = spack.platforms.host()
```

lib/spack/spack/cray_manifest.py
```diff
@@ -61,9 +61,17 @@ def compiler_from_entry(entry):
...
+        spec_platform = entry["arch"]["platform"]
+        # Note that Cray systems are now treated as Linux. Specs
+        # in the manifest which specify "cray" as the platform
+        # should be registered in the DB as "linux"
+        if local_platform == "linux" and spec_platform.lower() == "cray":
```
becker33:
```suggestion
        if local_platform.name == "linux" and spec_platform.lower() == "cray":
```

becker33 (changes requested):

becker33 (dismissed):

becker33:
@scheibelp have you tested this out locally? I know this isn't something we can add unit tests for yet.

scheibelp:
> have you tested this out locally? I know this isn't something we can add unit tests for yet.

I have. I could in fact update the tests to check this:

* I can create a manifest with the entries
* I can monkeypatch `platform.host()` to return a linux platform

would that seem like a reasonable way to do it?

becker33:
@scheibelp yeah that seems reasonable. I think it's better if you could only generate the in-memory entry associated with the cray manifest, that way we don't suffer a performance penalty in the test for doing IO.

scheibelp:
@becker33 a test has now been added

becker33 (approved):

PR #34179, opened 2022-11-29
gnuplot: make readline optional

iarspider:


iarspider:
@spackbot fix style

var/spack/repos/builtin/packages/gnuplot/package.py
```diff
@@ -51,6 +51,7 @@ class Gnuplot(AutotoolsPackage):
         description="Enable PBM (Portable Bit Map) and other older bitmap terminals",
     )
     variant("qt", default=False, description="Build with QT")
+    variant("readline", default=True, description="Build with readline")
 
     # required dependencies
     depends_on("readline")
```
tldahlgren: Should this only depend on `readline` when `+readline`?

tldahlgren: See https://spack.readthedocs.io/en/latest/packaging_guide.html#conditional-dependencies.

iarspider: Thanks for the catch!

tldahlgren (approved):
LGTM

PR #34183, opened 2022-11-29
fix libxcrypt package failure on centos8 .

srekolam:
I have observed the below libxcrypt package failure on the centos-8 docker container.
So, i tried to fix it by creating a new perl package for perl-file-open. 

checking whether all ucontext.h functions are available... yes
Can't locate open.pm in @INC (you may need to install the open module) (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5) at ./build-aux/scripts/expand-selected-hashes line 20.
BEGIN failed--compilation aborted at ./build-aux/scripts/expand-selected-hashes line 20.
configure: error: bad value 'all' for --enable-hashes
[libxcrypt-build-error.txt](https://github.com/spack/spack/files/10114340/libxcrypt-build-error.txt)


var/spack/repos/builtin/packages/libxcrypt/package.py
```diff
@@ -31,6 +31,7 @@ def url_for_version(self, version):
     variant("obsolete_api", default=False, when="@4.4.30:")
 
     patch("truncating-conversion.patch", when="@4.4.30")
+    depends_on("perl-file-open", type=("build", "run"), when="@4.3.31:")
```
haampie:
```suggestion
    depends_on("perl", type="build", when="@4.3.30:")
```

`open.pm` is part of perl itself

haampie (changes requested):
Thanks, can you make it just depend on Perl?

var/spack/repos/builtin/packages/libxcrypt/package.py
srekolam: Thanks.. Made the changes.

haampie (dismissed):

haampie:
@spackbot run pipeline

hseara:
To mention that the same behavior appears in Almalinux 9.1. And that Adding Perl ad a build dependency eliminates the problem, as suggested here.

bernhardkaindl (changes requested):
Hi! I've run into this error (with other packages) myself as well on Rocky 8 (another RHEL8 derivative):

I had set up spack with `spack external find perl` on a fresh install of Rocky 8 which (as can be deduced like @srekolam and @hseara systems as well) didn't have the RPM package `perl-open` preinstalled and therefore builds failed when they tried to `use open`.

It is therefore not a proper fix to just use `depends_on("perl", type="build")` when `spack external find perl` is in play as well: It would not be sufficient then because external perl also provides `"perl"`.

While it would be possible to define a variant `"perl+open"` and `depend_on` that, it's more light-weight to fix it like @srekolam already did in his initial commit. (By adding `perl-file-open` as a tiny addon-package, ensuring that the perl module `open` is present.)

Because the change requested by @haampie does not fix `spack external find perl` on such hosts where perl-open is not installed, I request a change which does take this into account.

The solution I'd prefer would be the initial commit of @srekolam, because implementing a `perl+open` variant in the perl recipe is more heavvy-weight in terms of having to add another variant to perl and having to rebuild perl for this if `perl+open` is not detected.

In #34074 (@maintainers: Please review), I implemented a fix where it it needed to check instead if `perl -e "require Thread::Queue"` is provided by the external perl. In this case, the check would likely be: `perl -e "use open IN"`

#34074 shows that it can be done, but I am not sure if the effort to implement a variant `perl+open` is worth it (It's more complicated to add than a tiny new package).

To have `perl+open` as a normal variant of `perl` of course it would have to be added as variant which can be enabled and disabled.

@srekolam 1st commit is light-weight, so I'd be ok with that unless @haampie rejects it and demands `perl+open` as a new variant.

bernhardkaindl:
I confirm that `depends_on("perl", type=("build", "run"), when="@4.3.31:")` does not fix the build of libxcrypt on RHEL8-based systems and newer a perl defined by `spack external find perl` is selected in the spec and the RPM `perl-open` is not installed (or more precisely: not installat at build time) because `perl-file-open` is then missing.

Because the pipeline failed and needs a re-run anyways, I'll try to push this PR rebased to current develop,
without the last commit which reverts adding `perl-file-open`, so `perl-file-open` is added and used for `libxcrypt`
until we have an even better solution which is able to detect if `perl-file-open` is available in the perl used at build time.

For the case would not be wanted by someone reviewing this, look at #34343 which adds perl as build dependency but misses the dependence on perl-file-open. That would be a step into the right direction, but not complete.

bernhardkaindl (dismissed):
While testing, I discovered that `perl-file-open` is not used by the build of `libxcrypt`, but `perl-open` is:
https://metacpan.org/pod/open - perl pragma to set default PerlIO layers for input and output

I thus wrote a new recipe to provide perl-open and use it as dependency.

It's only needed when `spack external find perl` is used on distros based on RHEL8+ when the RPM `perl-open` is not installed. I propose to to use this until there is some better solution to have perl-open on such hosts.

Another solution would be to add https://fastapi.metacpan.org/source/RJBS/perl-5.36.0/lib/open.pm as an additional source file to the packages which use perl-open (there are few) and use:
```py
        env.prepend_path("PERL5LIB", self.stage.path)
```

var/spack/repos/builtin/packages/libxcrypt/package.py
```diff
@@ -32,6 +32,8 @@ def url_for_version(self, version):
 
     patch("truncating-conversion.patch", when="@4.4.30")
 
+    depends_on("perl-open", type=("build", "run"), when="@4.4.30:")
```
michaelkuhn: I don't think this needs to be a run dependency, build should be fine.

michaelkuhn (commented):
I don't like the `perl-open` workaround too much, to be honest. It's more or less the same problem of external compilers not providing everything we need. Maybe the Perl package should check for the module to be available and only be allowed as an external package if it is available? We could also point users to RHEL's `perl-open` package to upgrade their local installation.

haampie:

PR #34185, opened 2022-11-29
ROOT: new versions with associated dependency constraints

HadrienG2:
@wdconinc Here's my take on the dependency constraints, so that the info on your closed PR don't get lost :)

I couldn't spell out the macOS conflict because I don't have access to a Mac so I can't check how macOS fills uname information.

Feel free to test/approve this once you have the time.

### DIFF
--- a/var/spack/repos/builtin/packages/root/package.py
+++ b/var/spack/repos/builtin/packages/root/package.py
@@ -233,8 +233,8 @@ class Root(CMakePackage):
 
     # Python
     depends_on("python@2.7:", when="+python", type=("build", "run"))
-    depends_on("python@2.7:3.6", when=":6.13 +python", type=("build", "run"))
     depends_on("python@2.7:3.10", when=":6.26.09 +python", type=("build", "run"))
+    depends_on("python@2.7:3.6", when=":6.13 +python", type=("build", "run"))
     depends_on("py-numpy", type=("build", "run"), when="+tmva")
     # This numpy dependency was not intended and will hopefully
     # be fixed in 6.20.06.
--- a/var/spack/repos/builtin/packages/root/package.py
+++ b/var/spack/repos/builtin/packages/root/package.py
@@ -234,7 +234,6 @@ class Root(CMakePackage):
     # Python
     depends_on("python@2.7:", when="+python", type=("build", "run"))
     depends_on("python@2.7:3.10", when="@:6.26.09 +python", type=("build", "run"))
-    depends_on("python@2.7:3.6", when="@:6.13 +python", type=("build", "run"))
     depends_on("py-numpy", type=("build", "run"), when="+tmva")
     # This numpy dependency was not intended and will hopefully
     # be fixed in 6.20.06.
@@ -319,6 +318,9 @@ class Root(CMakePackage):
     # See https://github.com/root-project/root/issues/11714
     if sys.platform == "darwin" and macos_version() >= Version("13"):
         conflicts("@:6.26.09", msg="macOS 13 support was added in 6.26.10")
+
+    # ROOT <6.14 is incompatible with Python >=3.7, which is the minimum supported by spack
+    conflicts("+python", when="@:6.13", msg="Spack wants python >=3.7, too new for ROOT <6.14")
 
     @classmethod
     def filter_detected_exes(cls, prefix, exes_in_prefix):
--- a/var/spack/repos/builtin/packages/root/package.py
+++ b/var/spack/repos/builtin/packages/root/package.py
@@ -292,7 +292,9 @@ class Root(CMakePackage):
 
     # Python incompatibilities
     conflicts("^python@3.7:", when="@:6.13 +python", msg="Python 3.7+ support was added in 6.14")
-    conflicts("^python@3.11:", when="@:6.26.09 +python", msg="Python 3.11+ support was added in 6.26.10")
+    conflicts(
+        "^python@3.11:", when="@:6.26.09 +python", msg="Python 3.11+ support was added in 6.26.10"
+    )
 
     # See https://github.com/root-project/root/issues/9297
     conflicts("target=ppc64le:", when="@:6.24")
@@ -316,7 +318,11 @@ class Root(CMakePackage):
     conflicts("+vmc", when="@6.26:", msg="VMC was removed in ROOT v6.26.00.")
 
     # nlohmann_json incompatibilities
-    conflicts("^nlohmann-json@3.11:", when="@:6.26.07", msg="nlohmann_json 3.11+ support was added in 6.26.08")
+    conflicts(
+        "^nlohmann-json@3.11:",
+        when="@:6.26.07",
+        msg="nlohmann_json 3.11+ support was added in 6.26.08"
+    )
 
     # Clang incompatibilities
     conflicts("%clang@16:", when="@:6.26.07", msg="clang 16+ support was added in 6.26.08")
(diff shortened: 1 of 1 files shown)


wdconinc:
As commented [here](https://github.com/spack/spack/pull/34160#discussion_r1034979936) (simultaneously to this PR), doesn't it make more sense to have this modeled as closed-ended dependencies rather than open-ended dependencies with conflicts? That would avoid 'automatically' allowing for example new major python versions when they come out even for older ROOT versions. It would 'pin' the dependencies to a known major version until a certain ROOT version that has consciously added support for it.

HadrienG2:
I'm not sure about full close-ended dependencies because it introduces significant extra maintenance effort where most dependency updates are actually harmless. But I pushed a version that avoids conflicts where possible in favor of version dependency constraints (as far as I know, it's not currently possible to avoid conflicts for OS and compiler constraints).

HadrienG2:
...and using https://stackoverflow.com/questions/1777344/how-to-detect-mac-os-version-using-python as a template, I added a conflict for macOS 13+ on older ROOT, though again I cannot test it due to lack of suitable hardware.

vvolkl:
Spack already knows the macOS version, it's probably easier to just call `spack.operating_systems.mac_os.macos_version` than deal with `platform`

Also, I'm currently trying to get macOS builds to work for key4hep, I can do some testing on our build machines.

HadrienG2:
That looks interesting, but in typical Python API documentation fashion, [the docs](https://spack.readthedocs.io/en/latest/spack.operating_systems.html#spack.operating_systems.mac_os.macos_version) claim that this function returns a "version object" without clarifying how such an object is actually constructed or used...

HadrienG2:
...oh, well, let's just copy-paste an incantation from another package that uses that API and hope that it's correct :shrug:

wdconinc:
Tested that this builds correctly on my system for 6.26.10, with the following variants:
```
4fmb3qq root@6.26.10~aqua~arrow+davix~dcache~emacs+examples~fftw~fits~fortran+gdml+gminimal~graphviz+gsl+http~ipo~jemalloc+math~memstat+minuit~mlp~mysql+opengl~oracle~postgres~pythia6~pythia8+python~qt4~r+roofit+root7+rpath~shadow~spectrum~sqlite+ssl~table+tbb+threads~tmva+unuran+vc+vdt+veccore~vmc~webgui+x+xml+xrootd build_system=cmake build_type=RelWithDebInfo cxxstd=17 patches=22af347,89294c4
```
(`89294c4` is a local patch; system is gcc@12; relevant dependency: nlohmann-json@3.11.2)

HadrienG2:
Then let's undraft this.

tldahlgren:
Looks like some pod-related issues with one of the pipeline stacks:

```
Waiting for pod pipeline/runner-k3q2udsw-project-2-concurrent-349nrm to be running, status is Pending
	ContainersNotInitialized: "containers with incomplete status: [init-permissions]"
	ContainersNotReady: "containers with unready status: [build helper]"
	ContainersNotReady: "containers with unready status: [build helper]"
Running on runner-k3q2udsw-project-2-concurrent-349nrm via x86-v2-pub-gitlab-runner-7684c89569-5zzvv...
```

@spackbot run pipeline

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -31,6 +32,8 @@ class Root(CMakePackage):
     # Development version (when more recent than production).
 
     # Production version
+    version("6.26.10", sha256="8e56bec397104017aa54f9eb554de7a1a134474fe0b3bb0f43a70fc4fabd625f")
+    version("6.26.08", sha256="4dda043e7918b40743ad0299ddd8d526b7078f0a3822fd06066df948af47940e")
```
tldahlgren, line 36: Confirmed these sha256 are correct.

tldahlgren:
Same stack failed again (see https://gitlab.spack.io/spack/spack/-/jobs/4803378).

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -310,6 +312,16 @@ class Root(CMakePackage):
     # Feature removed in 6.26.00:
     conflicts("+vmc", when="@6.26:", msg="VMC was removed in ROOT v6.26.00.")
 
+    # See https://github.com/root-project/root/issues/11128
+    conflicts("%clang@16:", when="@:6.26.07", msg="clang 16+ support was added in 6.26.08")
+
+    # See https://github.com/root-project/root/issues/11714
+    if sys.platform == "darwin" and macos_version() >= Version("13"):
+        conflicts("@:6.26.09", msg="macOS 13 support was added in 6.26.10")
```
sethrj, line 317: Ideally this would be
```python
    conflicts("os=ventura:", when="@:6.26.09", msg="macOS 13 support was added in 6.26.10")
```
but the range `:` doesn't work with OS. If you're OK with ignoring future macOS for now, you could do
```suggestion
    conflicts("os=ventura", when="@:6.26.09", msg="macOS 13 support was added in 6.26.10")
```

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -229,6 +233,7 @@ class Root(CMakePackage):
 
     # Python
     depends_on("python@2.7:", when="+python", type=("build", "run"))
+    depends_on("python@2.7:3.10", when="@:6.26.09 +python", type=("build", "run"))
```
sethrj, line 239: Should this be a `conflict` indicating an issue with newer pythons on older ROOT? I'm not sure it makes a difference to the concretizer these days, but semantically it might be more appropriate.

var/spack/repos/builtin/packages/root/package.py
HadrienG2: I would prefer to keep semantically correct if uglier code until os has improved enough that we can use it as you suggest.

var/spack/repos/builtin/packages/root/package.py
HadrienG2: Historically, one difference between this syntax and conflicts has been that in the conflict case, the spack concretizer would not try older dependency versions and just give up stating that there is a conflict. If you are telling me that this is not an issue anymore with newer spack versions, I can turn those back into conflicts.

sethrj: The new concretizer is not "greedy" like the old one, so I think it should work. The real expert is @alalazo who could perhaps share some insight into best practices here?

sethrj (dismissed):
I successfully installed 6.26.10 on macOS 13.

var/spack/repos/builtin/packages/root/package.py
alalazo: In general it's better to add "positive" directives like done here. Conflicts add penalties to our minimization and they might trigger more easily weird situations where, to avoid the penalty, Spack turns off some other default.

alalazo:
There's a conflict to be solved here

sethrj:
The conflict was my bad (#34244) so I've fixed it.

sethrj (approved):

sethrj:
@tldahlgren would you mind checking the box for a second reviewer? Since I added a commit it won't accept my review to qualify for merge.

tldahlgren (approved):
LGTM

sethrj:
@spackbot run ci

sethrj:
@spackbot run pipeline

PR #34186, opened 2022-11-29
Fix forced dependencies in petsc

heerener:
This commit fixes multiple virtual dependencies all being added to petsc, even if we don't want them.

Example: we're building with intel-mkl which provides scalapack, so we're automatically building with scalapack which we don't want.

### DIFF
--- a/var/spack/repos/builtin/packages/petsc/package.py
+++ b/var/spack/repos/builtin/packages/petsc/package.py
@@ -492,7 +492,9 @@ class Petsc(Package, CudaPackage, ROCmPackage):
                 useinc = True
                 uselib = True
 
-            library_requested = spacklibname.split(":")[0] in direct_dependencies
+            dependency_name = spacklibname.split(":")[0]
+            dependency_names = [d.name for d in spec.dependencies()]
+            library_requested = dependency_name in direct_dependencies and dependency_name in dependency_names
             options.append(
                 "--with-{library}={value}".format(
                     library=petsclibname, value=("1" if library_requested else "0")


balay:
Tagging @scheibelp as he understands `spec.dependencies()` feature

balay:
> Example: we're building with intel-mkl which provides scalapack, so we're automatically building with scalapack which we don't want.

So I can't reproduce this issue. I tried:
 ```
./bin/spack install petsc%oneapi@2022.2.0~hdf5~superlu-dist ^intel-oneapi-mpi ^intel-oneapi-mkl
```
And I see in the build:
```
--with-blaslapack-lib="/home/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-wjdqks5qqnraqs76ta74pokzni6c4d4k/mkl/2022.2.1/lib/intel64/libmkl_intel_lp64.so /home/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-wjdqks5qqnraqs76ta74pokzni6c4d4k/mkl/2022.2.1/lib/intel64/libmkl_sequential.so /home/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-wjdqks5qqnraqs76ta74pokzni6c4d4k/mkl/2022.2.1/lib/intel64/libmkl_core.so /lib64/libm.so"
...
--with-scalapack=0
```

i.e scalapack wasn't automatically picked up from MKL.

File: var/spack/repos/builtin/packages/petsc/package.py [hunk 1]
```diff
@@ -492,7 +492,9 @@ def configure_options(self):
                 useinc = True
                 uselib = True
 
-            library_requested = spacklibname.split(":")[0] in direct_dependencies
+            dependency_name = spacklibname.split(":")[0]
+            dependency_names = [d.name for d in spec.dependencies()]
+            library_requested = dependency_name in direct_dependencies and dependency_name in dependency_names
```
scheibelp, line 497: This effectively "undoes" the logic above:

```
        direct_dependencies = []
        for dep in spec.dependencies():
            direct_dependencies.append(dep.name)
            direct_dependencies.extend(set(vspec.name for vspec in dep.package.virtuals_provided))
```

It would be equivalent to replace that with

```
        direct_dependencies = []
        for dep in spec.dependencies():
            direct_dependencies.append(dep.name)
```

I think the issue for this user is that they are building `petsc~scalapack` with `^intel-mkl`, and don't want to build `--with-scalapack`. So I have two questions:

* @heerener is this causing an error? Or is the problem that you would rather not be building with this support?
* @balay I'll note that `intel-oneapi-mkl` only provides `scalapack` when built with `+cluster` (which is off by default), from reading through the `petsc` code I think it would need to also compare with the variants to decide whether to enable `--with-scalapack` (for example). Does that seem like a reasonable desire?

balay: > I'll note that intel-oneapi-mkl only provides scalapack when built with +cluster (which is off by default)

Ok - I see it now

`./bin/spack install petsc%oneapi@2022.2.0~hdf5~superlu-dist ^intel-oneapi-mpi ^intel-oneapi-mkl+cluster`

```
--with-blaslapack-l\
ib="/home/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2\
022.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/intel64/libmkl_scalapack_lp64.so /ho\
me/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1\
-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/intel64/libmkl_cdft_core.so /home/balay/spa\
ck.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5o\
aiyxcz3o25azys735tye3/mkl/2022.2.1/lib/intel64/libmkl_intel_lp64.so /home/balay/spack.orig/opt\
/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5oaiyxcz3o25a\
zys735tye3/mkl/2022.2.1/lib/intel64/libmkl_sequential.so /home/balay/spack.orig/opt/spack/linu\
x-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/\
mkl/2022.2.1/lib/intel64/libmkl_core.so /home/balay/spack.orig/opt/spack/linux-fedora37-skylak\
e/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/\
intel64/libmkl_blacs_intelmpi_lp64.so /lib64/libm.so"
--with-scalapack=1 --with-scalapack-lib="/home/balay/spack.orig/opt/spack/linux-fedora37-sk\
ylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/\
lib/intel64/libmkl_scalapack_lp64.so /home/balay/spack.orig/opt/spack/linux-fedora37-skylake/o\
neapi-2022.2.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/int\
el64/libmkl_cdft_core.so /home/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2\
.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/intel64/libmkl_\
intel_lp64.so /home/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-on\
eapi-mkl-2022.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/intel64/libmkl_sequential.\
so /home/balay/spack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-20\
22.2.1-vkmzx63sg5oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/intel64/libmkl_core.so /home/balay/sp\
ack.orig/opt/spack/linux-fedora37-skylake/oneapi-2022.2.0/intel-oneapi-mkl-2022.2.1-vkmzx63sg5\
oaiyxcz3o25azys735tye3/mkl/2022.2.1/lib/intel64/libmkl_blacs_intelmpi_lp64.so /lib64/libm.so"
```
So even if this is fixed  to use   `--with-scalapack=0`, `--with-blas-lapack-lib` gets and links with `libmkl_scalapack_lp64.so` due to `+cluster` - I'm not sure how that can be avoided.

> I think it would need to also compare with the variants to decide whether to enable --with-scalapack (for example).

Looks like that requires reworking that entire section that deals with external pkgs. [as some of them are variants - others are not]. i.e update
```
# tuple format (spacklibname, petsclibname, useinc, uselib)                                                                                                                                                
```
to
```
# tuple format (variantname, spacklibname, petsclibname, useinc, uselib)
                                                                                                                                                
```

heerener: > This effectively "undoes" the logic above:

Whoops

I'm not entirely familiar with everything that's happening here yet, but it seems like we're adamant about building without fortran support, which makes scalapack unhappy. We do want to use intel-mkl where possible.

Output heavily snipped for readability, but what I'm seeing without my patch is this:

```
==> Concretized petsc
-   dhqzawn  petsc@3.18.1%gcc@11.2.0~X~batch~cgns~complex~cuda~debug+double~exodusii~fftw~fortran~giflib+hdf5~hpddm~hwloc+hypre~int64~jpeg~knl~kokkos~libpng~libyaml~memkind+metis~mkl-pardiso~mmg~moab~mpfr+mpi~mumps~openmp~p4est~parmmg~ptscotch~random123~rocm~saws+shared~strumpack~suite-sparse~tetgen~trilinos~valgrind build_system=generic clanguage=C arch=linux-rhel7-skylake
 [...]
 [+]  4vxhjid      ^intel-mkl@2020.4.304%gcc@11.2.0~ilp64+shared build_system=generic threads=none arch=linux-rhel7-skylake
 [...]
 ==> [2022-11-30-09:11:43.651046] [...] '--with-scalapack=1' '--with-scalapack-lib=/[...]/libmkl_scalapack_lp64.so [...]
 ```
 
In my config, I have:
 
```
packages:
  petsc:
    variants: ~fortran~scalapack
  intel-mkl:
    externals:
      - spec: intel-mkl@2020.4.304 build_system=generic threads=none ~ilp64+shared
    version: [2020.4.304]
    buildable: false
  all:
    providers:
      scalapack:: [intel-mkl]
 ```
 
And my environment:
```
spack:
  specs:
    - petsc
```

balay:
> but it seems like we're adamant about building without fortran support

Any particular reason why enabling fortran this causes grief?

>  We do want to use intel-mkl where possible.

Looks like `inte-oneapi-mkl` is now the preferred package and it avoids pulling in scalapack [as it defaults to ~cluster] 

But yeah - setting `--with-scalapack=1` when `petsc~scalapack` is used is not good.

var/spack/repos/builtin/packages/petsc/package.py
scheibelp: @balay regarding

> Looks like that requires reworking that entire section that deals with external pkgs. [as some of them are variants - others are not]. i.e update

It's possible that you could keep the format and say:

```
if spacklibname in self.variants and self.spec.variants[spacklibname].value:
    ...
```

This assumes that for every package you enable, that the spack package name matches the variant name. If that is not true, then I think you'd have to do as you said.

heerener:
> > but it seems like we're adamant about building without fortran support
> 
> Any particular reason why enabling fortran this causes grief?

I'm guessing historical reasons and attempts to reduce build complexity.

> > We do want to use intel-mkl where possible.
> 
> Looks like `inte-oneapi-mkl` is now the preferred package and it avoids pulling in scalapack [as it defaults to ~cluster]

That actually helps a lot - as far as I'm concerned this solves the issue for us.

> But yeah - setting `--with-scalapack=1` when `petsc~scalapack` is used is not good.

I'll leave the decision on what to do with this PR up to you, but from the discussion above it seems like my fix was somewhat clunky and this might be better off being spun off into a different issue for that bigger rework you suggested,

balay:
> I'm guessing historical reasons and attempts to reduce build complexity.

In some sense `+fortran`  is the more stable option  [assuming its compatible with c/c++ compilers]  so might result in reduced build complexity.

And  this build failure (that triggered this PR) is due to `~fortran` [as the dependency on scalapack has this fortran complexity - need fortran for netlib-scalapack - but not scalapack from mkl. However petsc configure assumes fortran is always required for any scalapack)

There are 2 modes for disabling fortran in petsc.

- --with-fc=0 - do not configure a fortran compiler (i.e no petsc fortran interface, and can't easily use external fortran pkgs)
-  --with-fortran-bindings=0 - do configure a fortran compiler (and use it to manage external fortran pkgs) but do not build petsc fortran bindings

Right now `~fortran` defaults the the first one.  Perhaps also add `~fortran-bindings`? [then somehow manage the complexity of fortran dependencies between pkgs] - this is getting more complex...

haampie:
Hi @heerener, this package-related pull request can be migrated from `spack/spack` to `spack/spack-packages` using the [**migrate-pkg-prs utility**](https://github.com/spack/migrate-package-prs).

---

PR #34187, opened 2022-11-29
evtgen: add v02.02.00

vvolkl:
The  release tarballs are unfortunately a bit inconsistent, but the latest version exports its CMake targets! :tada: 

Also, the patch function errored for me when building on macOS with apple-clang, seems like this  is gcc specific.


iarspider (approved):
I don't have access to mac, so can't test if patch is needed for all darwin builds, or only for clang. Otherwise, LGTM

tldahlgren (approved):
Confirmed sha256, rest LGTM too

PR #34188, opened 2022-11-29
(Binary) text relocation cleanup

haampie:
(REALLY the last PR in this series)

Also drop threaded text replacement, since it does not help
significantly (relocation got faster, bottleneck is decompression). It
would be better to parallelize over packages, instead of over files per
package.

Small improvement with separate classes for text replacement is that we
now compile the regex in the constructor; previously it was compiled per
binary to be relocated.

### DIFF
--- a/lib/spack/spack/relocate_text.py
+++ b/lib/spack/spack/relocate_text.py
@@ -1,3 +1,8 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
 import re
 from collections import OrderedDict
 
--- a/lib/spack/spack/relocate_text.py
+++ b/lib/spack/spack/relocate_text.py
@@ -65,7 +65,7 @@ class TextFilePrefixReplacer(PrefixReplacer):
 
         data = f.read()
         new_data = re.sub(self.regex, replacement, data)
-        if data == new_data:
+        if id(data) == id(new_data):
             return
         f.seek(0)
         f.write(new_data)
--- a/lib/spack/spack/relocate_text.py
+++ b/lib/spack/spack/relocate_text.py
@@ -1,0 +1,215 @@
+import re
+from collections import OrderedDict
+
+import spack.error
+
+
+def encode_path(p):
+    return p if isinstance(p, bytes) else p.encode("utf-8")
+
+
+def prefix_to_prefix_as_bytes(prefix_to_prefix):
+    return OrderedDict((encode_path(k), encode_path(v)) for (k, v) in prefix_to_prefix.items())
+
+
+def utf8_path_to_binary_regex(prefix):
+    """Create a (binary) regex that matches the input path in utf8"""
+    prefix_bytes = re.escape(prefix).encode("utf-8")
+    return re.compile(b"(?<![\\w\\-_/])([\\w\\-_]*?)%s([\\w\\-_/]*)" % prefix_bytes)
+
+
+def byte_strings_to_single_binary_regex(prefixes):
+    all_prefixes = b"|".join(re.escape(p) for p in prefixes)
+    return re.compile(b"(?<![\\w\\-_/])([\\w\\-_]*?)(%s)([\\w\\-_/]*)" % all_prefixes)
+
+
+def utf8_paths_to_single_binary_regex(prefixes):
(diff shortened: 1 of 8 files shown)


lib/spack/spack/relocate_text.py
```diff
@@ -0,0 +1,240 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
```
alalazo, line 8: Can we add more docstrings to this module (top-level and in classes / functions)?

alalazo (commented):
I wonder if we can structure `spack.relocate` as a single package, containing the code of both the current `spack.relocate` and `spack.relocate_text`. It should ideally export a subset of the API, so other packages would just import `spack.relocate`.

Also, would it be worth, in this or another PR, to separate platform specific implementations like in https://github.com/spack/spack/pull/34779 ? Still wondering if this would simplify our maintenance or not.

lib/spack/spack/relocate_text.py
```diff
@@ -0,0 +1,288 @@
...
+Prefix = Union[str, bytes]
+
+
+def encode_path(p: Prefix) -> bytes:
+    return p if isinstance(p, bytes) else p.encode("utf-8")
+
+
+def _prefix_to_prefix_as_bytes(prefix_to_prefix) -> Dict[bytes, bytes]:
+    return OrderedDict((encode_path(k), encode_path(v)) for (k, v) in prefix_to_prefix.items())
```
alalazo, line 23: General comment (not a request). Since we support Python 3.6+, I don't think we need to use `OrderedDict` anymore

lib/spack/spack/relocate_text.py
```diff
@@ -0,0 +1,288 @@
...
+                /first/sub is matched and replaced before /first.
+        """
+        self.prefix_to_prefix = filter_identity_mappings(prefix_to_prefix)
+
+    @property
+    def is_noop(self) -> bool:
+        """Returns true when the prefix to prefix map
+        is mapping everything to the same location (identity)
+        or there are no prefixes to replace."""
+        return not bool(self.prefix_to_prefix)
```
alalazo, line 76: Another very minor comment. I don't think it's necessary to cast to bool here. Using:
```suggestion
        return not self.prefix_to_prefix
```
should have the same effect.

lib/spack/spack/relocate_text.py
```diff
@@ -0,0 +1,288 @@
...
+    def apply_to_filename(self, filename):
+        if self.is_noop:
+            return
+        with open(filename, "rb+") as f:
+            self.apply_to_file(f)
+
+    def apply_to_file(self, f):
+        if self.is_noop:
+            return
+        self._apply_to_file(f)
```
alalazo, line 93: In this base class we could add a stub:
```python
def _apply_to_file(self, f):
    raise NotImplementedError("must be implemented by derived classes")
```
to help IDEs and tools.

lib/spack/spack/relocate_text.py
```diff
@@ -0,0 +1,288 @@
...
+        Arguments:
+            binary_prefixes (list): List of byte strings of prefixes to match
+            suffix_safety_size (int): Sizeof the lookahed for null-terminated string.
+
+        Returns: compiled regex
+        """
+        return re.compile(
+            b"("
+            + b"|".join(re.escape(p) for p in binary_prefixes)
+            + b")([^\0]{0,%d}\0)?" % suffix_safety_size
```
alalazo, line 159: Fwiw, I didn't double check all the regexes - just a couple at random. They're all moved from previous locations, so they should be fine.

lib/spack/spack/relocate_text.py
```diff
@@ -0,0 +1,288 @@
...
+        If there *is* a null terminator we can do the same as long as the replacement
+        has a sufficiently long common suffix with the original prefix.
+        As a last resort when the replacement does not have a long enough common suffix,
+        we can try to shorten the string, but this only works if the new length is
+        sufficiently short (typically the case when going from large padding -> normal path)
+        If the replacement string is longer, or all of the above fails, we error out.
+
+        Arguments:
+            f: file opened in rb+ mode
+        """
```
alalazo, line 197: :+1:

alalazo (approved):

lib/spack/spack/relocate_text.py
haampie: True, but I thought the intent was clearer like this

Issue #34193, opened 2022-11-29
Installation issue: krb5~shared


charmoniumQ:
### Steps to reproduce the issue

Note that `krb5+shared` works fine.

```console
$ spack spec -I <spec>
Input spec
--------------------------------
 -   krb5~shared

Concretized
--------------------------------
 -   krb5@1.20.1%gcc@11.3.0~shared build_system=autotools arch=linux-ubuntu22.04-skylake
[+]      ^bison@3.8.2%gcc@11.3.0 arch=linux-ubuntu22.04-skylake
[+]          ^m4@1.4.19%gcc@11.3.0+sigsegv patches=9dc5fbd,bfdffa7 arch=linux-ubuntu22.04-skylake
[+]              ^libsigsegv@2.13%gcc@11.3.0 arch=linux-ubuntu22.04-skylake
[+]          ^perl@5.34.1%gcc@11.3.0+cpanm+shared+threads arch=linux-ubuntu22.04-skylake
[+]              ^berkeley-db@18.1.40%gcc@11.3.0+cxx~docs+stl patches=b231fcc arch=linux-ubuntu22.04-skylake
[+]              ^gdbm@1.19%gcc@11.3.0 arch=linux-ubuntu22.04-skylake
[+]                  ^readline@8.1%gcc@11.3.0 arch=linux-ubuntu22.04-skylake
[+]              ^zlib@1.2.12%gcc@11.3.0+optimize+pic+shared patches=0d38234 arch=linux-ubuntu22.04-skylake
[+]      ^diffutils@3.8%gcc@11.3.0 arch=linux-ubuntu22.04-skylake
[+]          ^libiconv@1.16%gcc@11.3.0 libs=shared,static arch=linux-ubuntu22.04-skylake
[+]      ^gettext@0.21.1%gcc@11.3.0+bzip2+curses+git~libunistring+libxml2+tar+xz build_system=autotools arch=linux-ubuntu22.04-skylake
[+]          ^bzip2@1.0.8%gcc@11.3.0~debug~pic+shared arch=linux-ubuntu22.04-skylake
[+]          ^libxml2@2.10.3%gcc@11.3.0~python build_system=autotools arch=linux-ubuntu22.04-skylake
[+]              ^pkgconf@1.8.0%gcc@11.3.0 arch=linux-ubuntu22.04-skylake
[+]          ^ncurses@6.2%gcc@11.3.0~symlinks+termlib abi=none arch=linux-ubuntu22.04-skylake
[...]

PR #34195, opened 2022-11-29
Add new package: micromamba

charmoniumQ:
Micromamba is a C++ rewrite of the conda package manager that aims to be faster and more hermetic. Miniconda is written in Python, so it's installation process runs a script that downloads its own Python and Python packages (neither of which are managed by Spack). Micromamba is a C++ standalone rewrite of Miniconda that does not need to install Python or Python packages.


charmoniumQ:
Currently blocked by https://github.com/spack/spack/issues/34193, although I am looking for workarounds.

charmoniumQ:
Getting this error:

```
cd /tmp/sam/spack-stage/spack-stage-micromamba-1.0.0-hsyfletrgwe3rpkmoisedxw2zts5r7it/spack-build-hsyflet/libmamba && /home/sam/.local/share/spack/lib/spack/env/gcc/g++ -DLIBMAMBA_EXPORTS -DSPDLOG_FMT_EXTERNAL -Dlibmamba_EXPORTS -I/tmp/sam/spack-stage/spack-stage-micromamba-1.0.0-hsyfletrgwe3rpkmoisedxw2zts5r7it/spack-src/libmamba/include -I/tmp/sam/spack-stage/spack-stage-micromamba-1.0.0-hsyfletrgwe3rpkmoisedxw2zts5r7it/spack-src/libmamba/src -isystem /home/sam/.local/share/spack/opt/spack/linux-ubuntu22.04-skylake/gcc-11.3.0/yaml-cpp-0.7.0-zm5yle4qzcg77uulpxar7fapzmmf2ns6/include -isystem /home/sam/.local/share/spack/opt/spack/linux-ubuntu22.04-skylake/gcc-11.3.0/libreproc-14.2.4-lbugs4waibjhtzoaqh5rxsvn5lqutyjw/include -isystem /home/sam/.local/share/spack/opt/spack/linux-ubuntu22.04-skylake/gcc-11.3.0/tl-expected-master-a2awf2nwoihkbpron67iobxi76gzdgme/include -isystem /home/sam/.local/share/spack/opt/spack/linux-ubuntu22.04-skylake/gcc-11.3.0/tl-expected-master-a2awf2nwoihkbpron67iobxi76gzdgme/share -isystem /home/sam/.local/share/spack/opt/spack/linux-ubuntu22.04-skylake/gcc-11.3.0/fmt-9.1.0-xki4etfiiw7rmyi64afr6tjxcso3iwvp/include -isystem /home/sam/.local/share/spack/opt/spack/linux-ubuntu22.04-skylake/gcc-11.3.0/spdlog-1.10.0-k7cacussi2trtrd3h3v7aqal2iwow5oh/include -Wall -O2 -g -DNDEBUG -fPIC -std=gnu++17 -MD -MT libmamba/CMakeFiles/libmamba.dir/src/core/channel.cpp.o -MF CMakeFiles/libmamba.dir/src/core/channel.cpp.o.d -o CMakeFiles/libmamba.dir/src/core/channel.cpp.o -c /tmp/sam/spack-stage/spack-stage-micromamba-1.0.0-hsyfletrgwe3rpkmoisedxw2zts5r7it/spack-src/libmamba/src/core/channel.cpp
/tmp/sam/spack-stage/spack-stage-micromamba-1.0.0-hsyfletrgwe3rpkmoisedxw2zts5r7it/spack-src/libmamba/src/core/singletons.cpp:9:10: fatal error: curl/urlapi.h: No such file or directory
    9 | #include <curl/urlapi.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [libmamba/CMakeFiles/libmamba.dir/build.make:141: libmamba/CMakeFiles/libmamba.dir/src/core/singletons.cpp.o] Error 1
```

However, `curl` is included as a build-dependency, and it is `find_package`d by CMake ([this line](https://github.com/mamba-org/mamba/blob/micromamba-1.0.0/libmamba/CMakeLists.txt#L425),

var/spack/repos/builtin/packages/libreproc/package.py
```diff
@@ -0,0 +1,30 @@
...
+
+    maintainers = ["charmoniumQ"]
+
+    version("14.2.4", sha256="55c780f7faa5c8cabd83ebbb84b68e5e0e09732de70a129f6b3c801e905415dd")
+
+    variant("cxx", default=False, description="Build reproc C++ bindings")
+    variant("shared", default=True, description="Build shared libraries, otherwise static")
+
+    depends_on("cmake@3.14:", type="build")
+    depends_on("zlib+shared", type=("link", "run"), when="+shared")
```
adamjstewart: "run" doesn't make sense for zlib since there is no `prefix.bin` directory

var/spack/repos/builtin/packages/libsolv/package.py
```diff
@@ -0,0 +1,31 @@
...
+
+    homepage = "https://en.opensuse.org/OpenSUSE:Libzypp_satsolver"
+    url = "https://github.com/opensuse/libsolv/archive/0.7.22.tar.gz"
+
+    maintainers = ["charmoniumQ"]
+
+    version("0.7.22", sha256="968aef452b5493751fa0168cd58745a77c755e202a43fe8d549d791eb16034d5")
+
+    variant("shared", default=True, description="Build shared libraries, otherwise build static")
+    variant("conda", default=False, description="Include solv/conda.h, otherwise not")
```
adamjstewart: You don't need to define both the true and false cases, I think the opposite will be obvious

var/spack/repos/builtin/packages/tl-expected/package.py
```diff
@@ -0,0 +1,22 @@
...
+    url = "https://github.com/TartanLlama/expected/archive/1.0.0.tar.gz"
+
+    maintainers = ["charmoniumQ"]
+
+    version("1.0.0", sha256="8f5124085a124113e75e3890b4e923e3a4de5b26a973b891b3deb40e19c03cee")
+    # Note that the 1.0.0 has this issue:
+    # https://github.com/TartanLlama/expected/issues/114
+    # But no new patch version has been released,
+    # so I will use the latest commit at the time of writing:
+    version("b74fecd", sha256="c87d54bfbfd9deb11dd58e5ce2960cc9981dcaea675bf1a05e21a52f4321216b")
```
adamjstewart: Commit hashes aren't good version names because they don't sort properly. Instead, I would recommend a version like `2022-11-24` (the date of the commit)

adamjstewart: Also, how does this fetch properly? Don't you need to specify the `git` attribute and download from git?

var/spack/repos/builtin/packages/micromamba/package.py
```diff
@@ -0,0 +1,153 @@
...
+            self.define("BUILD_MICROMAMBA", True),
+            self.define("BUILD_STATIC", linkage == "static"),
+            self.define("BUILD_STATIC_DEPS", linkage == "full_static"),
+            self.define("BUILD_SHARED", linkage == "dynamic"),
+            self.define("MICROMAMBA_LINKAGE", linkage.upper()),
+        ]
+
+    @run_after("install")
+    @on_package_attributes(run_tests=True)
+    def check_install(self):
```
adamjstewart, line 154: @tldahlgren what's the easiest way to support this test both at build time and post-install?

var/spack/repos/builtin/packages/micromamba/package.py
```diff
@@ -0,0 +1,153 @@
...
+        # See https://github.com/mamba-org/mamba/blob/micromamba-1.0.0/libmamba/CMakeLists.txt#L423
+        depends_on("libsolv+conda", type=("link", "run"))
+        depends_on("curl libs=shared", type=("link", "run"))
+        depends_on("libarchive crypto=mbedtls xar=libxml2", type=("link", "run"))
+        depends_on("openssl", type=("link", "run"))
+        depends_on("yaml-cpp", type=("link", "run"))
+        depends_on("libreproc+cxx+shared", type=("link", "run"))
+        depends_on("tl-expected@b74fecd", type=("link", "run"))
+        depends_on("fmt", type=("link", "run"))
+        depends_on("spdlog", type=("link", "run"))
```
adamjstewart: Run probably not needed

var/spack/repos/builtin/packages/micromamba/package.py
```diff
@@ -0,0 +1,153 @@
...
+        # See linkage=dynamic for what that entails.
+        depends_on("libsolv+conda", type=("link", "run"))
+        depends_on("curl libs=shared", type=("link", "run"))
+        depends_on("libarchive crypto=mbedtls xar=libxml2", type=("link", "run"))
+        depends_on("openssl", type=("link", "run"))
+        depends_on("yaml-cpp", type=("link", "run"))
+        depends_on("libreproc+cxx", type=("link", "run"))
+        depends_on("tl-expected@b74fecd", type=("link", "run"))
+        depends_on("fmt", type=("link", "run"))
+        depends_on("spdlog", type=("link", "run"))
```
adamjstewart: run probably not needed

var/spack/repos/builtin/packages/tl-expected/package.py
charmoniumQ: It uses the <https://github.com/TartanLlama/expected/archive/b74fecd.tar.gz> which GitHub resolves properly.

Is it preferred to talk to GitHub repos over HTTP (archive tarball) or Git?

charmoniumQ: ... which is why the version is named after the commit hash.

var/spack/repos/builtin/packages/micromamba/package.py
```diff
@@ -0,0 +1,153 @@
...
+    patch("fix-threads.patch")
+
+    def cmake_args(self):
+        # See https://mamba.readthedocs.io/en/latest/developer_zone/build_locally.html#build-micromamba
+        if "linkage=dynamic" in self.spec:
+            linkage = "dynamic"
+        elif "linkage=static" in self.spec:
+            linkage = "static"
+        elif "linkage=full_static" in self.spec:
+            warnings.warn("Full static does not work yet!")
```
adamjstewart: `tty.warn` is better than `warnings` since that's how all Spack error/warning/debug/info messages are printed. But I wouldn't do this here, I would just add a conflict or even remove the option until we get it working.

var/spack/repos/builtin/packages/tl-expected/package.py
adamjstewart: Gotcha. We generally prefer HTTP over Git because it can be checksummed, but git hashes also work for this purpose. I think most package specify the commit and download via git instead of using HTTP for a random commit. This also makes the version more meaningful.

var/spack/repos/builtin/packages/tl-expected/package.py
```diff
@@ -0,0 +1,23 @@
...
+    git = "https://github.com/TartanLlama/expected.git"
+
+    maintainers = ["charmoniumQ"]
+
+    version("1.0.0", sha256="8f5124085a124113e75e3890b4e923e3a4de5b26a973b891b3deb40e19c03cee")
+    # Note that the 1.0.0 has this issue:
+    # https://github.com/TartanLlama/expected/issues/114
+    # But no new patch version has been released,
+    # so I will use the latest commit at the time of writing:
+    version("2022-11-24", commit="b74fecd4448a1a5549402d17ddc51e39faa5020c")
```
adamjstewart, line 22: since 2022 > 1, this version will be considered newer and will install by default. Is this what you want? Maybe we should put it above 1.0.0 just to clarify this.

adamjstewart (dismissed):
LGTM, not sure if anyone else would be interested in reviewing this.

adamjstewart (approved):

PR #34199, opened 2022-11-29
Environments: refresh cached URL-included config on re-concretization

tldahlgren:
@tuxfan 

This PR makes the following changes regarding included environment configuration files cached from a URL:

- changes the permissions on the cached file to read-only (user, group) on unix to discourage local modifications;
- allows `spack concretize -f` to re-fetch (or refresh) the cached file(s) resulting in any local modifications being lost; and
- changes skipping re-stage warnings to debug messages.

### DIFF
--- a/lib/spack/spack/config.py
+++ b/lib/spack/spack/config.py
@@ -34,6 +34,7 @@ import copy
 import functools
 import os
 import re
+import stat
 import sys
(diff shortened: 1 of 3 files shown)


tldahlgren:
@tgamblin Here's a PR for the approach to the issue that we discussed the other day.

PR #34202, opened 2022-11-30
external python extensions: more robust external python finding

becker33:
This fixes an error found by @scheibelp with the implementation in #33777, in which python externals were being constructed that were not sufficiently fleshed out for the install logic to succeed.

This should result in usable python external dependencies for external python extensions.

@tgamblin you asked me to put you on this as well when I mentioned it.

### DIFF
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -2,6 +2,7 @@
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
+import archspec
 import inspect
 import os
 import re
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -2,12 +2,13 @@
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
-import archspec
 import inspect
 import os
 import re
 import shutil
 from typing import Optional
+
+import archspec
 
 import llnl.util.filesystem as fs
 import llnl.util.lang as lang
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -240,11 +240,40 @@ class PythonPackage(PythonExtension):
                 if not python.concrete:
                     repo = spack.repo.path.repo_for_pkg(python)
                     python.namespace = repo.namespace
-                    python._mark_concrete()
+
+                    # Ensure architecture information is present
+                    if not python.architecture:
+                        host_platform = spack.platforms.host()
+                        host_os = host_platform.operating_system("default_os")
+                        host_target = host_platform.target("default_target")
+                        python.architecture = spack.spec.ArchSpec(
+                            str(host_platform), str(host_os), str(host_target)
+                        )
+                    else:
+                        if not python.architecture.platform:
+                            python.architecture.platform = spack.platforms.host()
+                        if not python.architecture.os:
+                            python.architecture.os = "default_os"
+                        if not python.architecture.target:
+                            python.architecture.target = archspec.cpu.host().family
+
+                    # Ensure compiler information is present
+                    if not python.compiler:
+                        python.compiler = self.spec.compiler
+
                     python.external_path = self.prefix
+                    python._mark_concrete()
             self.spec.add_dependency_edge(python, ("build", "link", "run"))
 
     def get_external_python_for_prefix(self):
+        """
+        For an external package that extends python, find the most likely spec for the python
+        it depends on.
+
+        First search: an "installed" external that shares a prefix with this package
+        Second search: a configured external that shares a prefix with this package
+        Third search: search this prefix for a python package
+        """
         python_externals_installed = [
             s for s in spack.store.db.query("python") if s.prefix == self.prefix
         ]
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -298,7 +298,9 @@ class PythonPackage(PythonExtension):
         )
 
         python_externals_detected = [
-            d.spec for d in python_externals_detection.get("python", []) if d.prefix == self.spec.external_path
+            d.spec
+            for d in python_externals_detection.get("python", [])
+            if d.prefix == self.spec.external_path
         ]
         if python_externals_detected:
             return python_externals_detected[0]
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -13,9 +13,12 @@ import llnl.util.lang as lang
 import llnl.util.tty as tty
 
 import spack.builder
+import spack.config
+import spack.detection
 import spack.multimethod
 import spack.package_base
 import spack.spec
+import spack.store
 from spack.directives import build_system, depends_on, extends
 from spack.error import NoHeadersError, NoLibrariesError, SpecError
 from spack.version import Version
@@ -233,12 +236,40 @@ class PythonPackage(PythonExtension):
             if "python" in self.spec.root:
                 python = self.spec.root["python"]
             else:
-                python = spack.spec.Spec("python")
-                repo = spack.repo.path.repo_for_pkg(python)
-                python.namespace = repo.namespace
-                python._mark_concrete()
-                python.external_path = self.prefix
+                python = self.get_external_python_for_prefix()
+                if not python.concrete:
+                    repo = spack.repo.path.repo_for_pkg(python)
+                    python.namespace = repo.namespace
+                    python._mark_concrete()
+                    python.external_path = self.prefix
             self.spec.add_dependency_edge(python, ("build", "link", "run"))
+
+    def get_external_python_for_prefix(self):
+        python_externals_installed = [
+            s for s in spack.store.db.query("python") if s.prefix == self.prefix
+        ]
+        if python_externals_installed:
+            return python_externals_installed[0]
+
+        python_external_config = spack.config.get("packages:python:externals", [])
+        python_externals_configured = [
+            spack.spec.Spec(item["spec"]) for item in python_external_config
+            if item["prefix"] == self.prefix
+        ]
+        if python_externals_configured:
+            return python_externals_configured[0]
+
+        python_externals_detection = spack.detection.by_executable(
+            [spack.repo.path.get_pkg_class("python")], path_hints=[self.prefix]
+        )
+
+        python_externals_detected = [
+            d.spec for d in python_externals_detection.get("python", []) if d.prefix == self.prefix
+        ]
+        if python_externals_detected:
+            return python_externals_detected[0]
+
+        raise StopIteration("No external python could be detected for %s to depend on" % self.spec)
 
     @property
     def headers(self):
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -277,6 +277,9 @@ class PythonPackage(PythonExtension):
         First search: an "installed" external that shares a prefix with this package
         Second search: a configured external that shares a prefix with this package
         Third search: search this prefix for a python package
+
+        Returns:
+          spack.spec.Spec: The external Spec for python most likely to be compatible with self.spec
         """
         python_externals_installed = [
             s for s in spack.store.db.query("python") if s.prefix == self.spec.external_path
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -224,7 +224,7 @@ class PythonPackage(PythonExtension):
             name = cls.pypi.split("/")[0]
             return "https://pypi.org/simple/" + name + "/"
 
-    def update_external_dependencies(self):
+    def update_external_dependencies(self, extendee_spec=None):
         """
         Ensure all external python packages have a python dependency
 
@@ -235,7 +235,9 @@ class PythonPackage(PythonExtension):
         """
         # TODO: Include this in the solve, rather than instantiating post-concretization
         if "python" not in self.spec:
-            if "python" in self.spec.root:
+            if extendee_spec:
+                python = extendee_spec
+            elif "python" in self.spec.root:
                 python = self.spec.root["python"]
             else:
                 python = self.get_external_python_for_prefix()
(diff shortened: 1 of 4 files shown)


becker33:
@spackbot fix style

lib/spack/spack/build_systems/python.py
```diff
@@ -233,13 +236,42 @@ def update_external_dependencies(self):
...
+            return python_externals_installed[0]
+
+        python_external_config = spack.config.get("packages:python:externals", [])
+        python_externals_configured = [
+            spack.spec.Spec(item["spec"])
+            for item in python_external_config
+            if item["prefix"] == self.prefix
+        ]
+        if python_externals_configured:
+            return python_externals_configured[0]
```
scheibelp, line 297: In my env I have an external python which looks like:

```
externals:
- spec: python@3.7.2
```

The result is that the spec `python@3.7.2` is added to the DAG, and it does not have details added during concretization, possibly because it is a child of an external.

After installing that and doing another `concretize -f`, this spec is picked up by the db check (on line 249 here) and later there is a failure because of insufficient spec details.

```
File ".../spack/lib/spack/spack/solver/asp.py", line 625, in solve
  setup.setup(self, specs, reuse=reuse)
File ".../spack/lib/spack/spack/solver/asp.py", line 1989, in setup
  self._facts_from_concrete_spec(reusable_spec, possible)
File ".../spack/lib/spack/spack/solver/asp.py", line 1910, in _facts_from_concrete_spec
  self.possible_oses.add(dep.os)
File ".../spack/lib/spack/spack/spec.py", line 4699, in os
  return self.architecture.os
```

i.e. the spec is missing `arch=...`.

A couple notes:

* we could assign spec details ourselves but this would result in multiple instances of the Python spec in the same dag
  * I'll furthermore note that right now `traverse.py` maintains `visited` using `id` as the key by default, so even if the specs are `==` according to `Spec.__eq__`, `Package.extendee_spec` can fail its internal assertion
* Earlier I suggested concretizing the python spec but (a) it runs into the same potential issue as the prior point and (b) you mentioned that running a concretization can be slow because of the setup time

I'm becoming more convinced that this type of coordination needs to occur at the group-of-specs level, i.e. that this loop in `asp.py`:

```
        for root in self._specs.values():
             for dep in root.traverse():
                 if dep.external:
                     dep.package.update_external_dependencies()
```

should not do the patching individually: it should search all of them for a python, and then attach it to any that need it, so they all have the same one.

lib/spack/spack/build_systems/python.py
```diff
@@ -233,13 +236,42 @@ def update_external_dependencies(self):
...
-                python.external_path = self.prefix
+                python = self.get_external_python_for_prefix()
+                if not python.concrete:
+                    repo = spack.repo.path.repo_for_pkg(python)
+                    python.namespace = repo.namespace
+                    python._mark_concrete()
+                    python.external_path = self.prefix
             self.spec.add_dependency_edge(python, ("build", "link", "run"))
 
+    def get_external_python_for_prefix(self):
```
adamjstewart, line 272: Needs a docstring. Type hints would also be useful. I have no idea what this function does without documentation, and it's in a file I maintain.

lib/spack/spack/build_systems/python.py
becker33: We can definitely find the platform/os for the python spec easily, but the target is actually quite hard -- we don't have any way to know how conservatively it was compiled

Doing this at the multiple-package level may be fine, but doesn't solve the problem in the case of a single spec which doesn't have any other dependencies on python. We need a more general solution than that.

scheibelp: There are two problems then:

* Consistency: solved by patching all the specs together rather than individually
* Choosing the right target: if we have to make a Python spec up, why not just use whatever `spack arch` would return? Or we could fall back to that if the user didn't set target on the external Python (or the external Python extension)

becker33:
@spackbot fix style

lib/spack/spack/build_systems/python.py
becker33: The basic idea is to find a python spec that should be associated with an external python depedendent spec. I've added a docstring, let me know if you think the docstring should be more descriptive.

adamjstewart: Thanks! Can you also add a `Returns:` section and add type hints?

scheibelp:
I think your edits have addressed point 2 of https://github.com/spack/spack/pull/34202#discussion_r1037724622 but not point 1: if you have specs like

```
A
| \
B |
| /
C
```

Where B is an external python extension and a spec list like:

```
- A
- B
```

Then in the current logic, A will get the concretized external Python, and B will get an invented Python (in which case, the DAG for A will contain two different instances of the Python spec). That creates a consistency issue, which I think can only be addressed "above" this point (in `asp.build_specs`).

becker33:
@spackbot fix style

lib/spack/spack/test/concretize.py
```diff
@@ -1953,3 +1956,62 @@ def test_external_python_extensions_have_dependency(self):
...
+        }
+        spack.config.set("packages", external_conf)
+
+        spec = Spec("py-extension1").concretized()
+
+        assert "python" in spec["py-extension1"]
+        assert spec["python"].prefix == prefix
+        assert spec["python"] == python_spec
+
+    def test_external_python_extension_find_unified_python(self, monkeypatch):
```
scheibelp: This test doesn't use monkeypatching

lib/spack/spack/solver/asp.py
```diff
@@ -2245,6 +2245,12 @@ def external_spec_selected(self, pkg, idx):
         )
         self._specs[pkg].extra_attributes = spec_info.get("extra_attributes", {})
 
+        # If this is an extension, update the dependencies to include the extendee
+        package = self._specs[pkg].package_class(self._specs[pkg])
+        extendee_spec = package.extendee_spec
+        if extendee_spec:
+            package.update_external_dependencies(self._specs.get(extendee_spec.name, None))
```
scheibelp, line 2252: Sanity check: nothing is buried transitively in `_specs` right?

lib/spack/spack/solver/asp.py
```diff
@@ -2313,15 +2319,17 @@ def deprecated(self, pkg, version):
...
-            return (-3, 0)
+            return (-4, 0)
         elif name == "node_compiler":
-            return (-2, 0)
+            return (-3, 0)
         elif name == "node_flag":
-            return (-1, 0)
+            return (-2, 0)
+        elif name == "external_spec_selected":
+            return (0, 0)  # note out of order so this goes last
```
scheibelp, line 2339: This comment is confusing to me.

Is there an overall explanation on the importance of this sorting (e.g. `build_specs` has a comment but doesn't expand other than to say they need to be considered "in the right order") ? I think it might be worth expanding the docstring to explain some of the ordering rules, e.g. :

```
# Make sure to build external specs last, because they may need to pull details from non-external specs
```

Also, do we need to make sure to build extendee specs before non-extendee specs (in other words, does this sorting need to be finer-grained)?

lib/spack/spack/test/concretize.py
```diff
@@ -1953,3 +1956,62 @@ def test_external_python_extensions_have_dependency(self):
...
+        }
+        spack.config.set("packages", external_conf)
+
+        spec = Spec("py-extension1").concretized()
+
+        assert "python" in spec["py-extension1"]
+        assert spec["python"].prefix == "/fake"
+        # The spec is not equal to spack.spec.Spec("python@configured") because it gets a
+        # namespace and an external prefix before marking concrete
+        assert spec["python"].satisfies("python@configured")
```
scheibelp: At least one other scenario that might be worth testing:

* A->Bextension->python, A->python, there is an external python but A doesn't use it (because it demands a different version). Make sure Bextension uses the same Python as A.

I need to think more about whether more use cases would be helpful for ensuring coverage of this

scheibelp (commented):
I have a couple minor requests and a couple questions. I need to think more about the unit tests and whether they cover all the behavior we want to check.

lib/spack/spack/solver/asp.py
becker33: Correct, every node in the solve becomes an entry in `_specs`.

lib/spack/spack/solver/asp.py
becker33: We will assemble everything about each spec before considering whether the spec is external, so we don't need to do anything else to make sure the other specs are assembled. I can certainly expand on the explaining comment.

lib/spack/spack/test/concretize.py
becker33: If an extension E is in the same solve as python P, E will use P. Regardless of what else we can say about the structure.

My inclination is that we don't need to create every structure we can think of, because understanding the code makes clear that all these structures will be handled identically (which is definitely a change from earlier iterations of this PR). If you feel strongly that we need to add more tests or that I'm wrong about this, let me know.

becker33:
@adamjstewart @scheibelp I hope these comment clarifications help, let me know if there's more detail you think belongs in the source code.

lib/spack/spack/test/concretize.py
scheibelp: > understanding the code makes clear that all these structures will be handled identically

I agree this is likely the case for the current implementation, but I think more tests are still worthwhile: this logic feels more critical and therefore more effort should be made to make it robust to underlying changes (in particular since at least one iteration of this PR would not have behaved identically).

The tests also enforce policy and make intended behavior clear to readers (for example: I don't think any of the docstrings discuss the importance of performing this external-unifying operation on a per-concretization-group basis).

lib/spack/spack/solver/asp.py
scheibelp: The docstring has improved but I suggest replacing

> note out of order so this goes last

with

> note out of order

Also:

* I'm actually not clear why `external_spec_selected` can't go into the catch-all category: what features of nodes need to be set in order for `update_external_dependencies` to set things up properly? It would seem to me that the specs don't even need to be fully fleshed out as long as the connections between the specs are established (which is what I assume the `node`/`hash` steps take care of).
* (not specifically changed in this PR but part of me understanding the significance of the sorting) is there a reason that `node_flag` is explicitly part of this sorting? It seems like it would fit into the "everything else" category (the spec needs to exist to assign flags, but I think this is addressed by using `node` first)

lib/spack/spack/test/concretize.py
scheibelp: Other test that is worthwhile: this PR makes an effort to do it's best with `when_possible`, so should there be a test where a concretization ends up having to split the input steps into more than one group.

scheibelp (commented):
Additional questions/comments (e.g. trying to understand more about the sorting requirements for the spec builder).

lib/spack/spack/solver/asp.py
becker33: `depends_on` is in the catch-all category. We could equally well ensure `depends_on` goes first, but then we would need to ensure we never use `satisfies` in the code for `update_external_dependencies`, so I think making `external_spec_selected` go last is more robust.

I tried to figure out why `node_flag` was sorted separately while I was working on this, and I couldn't figure it out either.

quellyn:
Tested this PR just now on LLNL's RZvernal system (x86_64, Cray CCE 15.0.0) and got the following when I try to concretize:
```console
[quellyn@rzvernal11:2.3.0]$ spack concretize -fU |& tee -a concretize.CCE
==> Warning: using "git@2.31.1" which is a deprecated version
==> Error: update_external_dependencies() takes 1 positional argument but 2 were given
```

I'm attaching the debug output of the above as well:
[concretize-debug-CC.txt](https://github.com/spack/spack/files/10363691/concretize-debug-CC.txt)

becker33:
@quellyn thanks for trying it out, you caught a bug that made this fail for any non-python externals. It should be fixed now if you want to give it another go.

quellyn:
> @quellyn thanks for trying it out, you caught a bug that made this fail for any non-python externals. It should be fixed now if you want to give it another go.

This looks like it's working for me now! Any idea when you might be ready to merge this MR ?

becker33:
cycling closed to bump CI

becker33:
@spackbot fix style

lib/spack/spack/build_systems/python.py
```diff
@@ -230,16 +235,81 @@ def update_external_dependencies(self):
         """
         # TODO: Include this in the solve, rather than instantiating post-concretization
         if "python" not in self.spec:
-            if "python" in self.spec.root:
+            if extendee_spec:
+                python = extendee_spec
+            elif "python" in self.spec.root:
                 python = self.spec.root["python"]
```
scheibelp, line 241: Apologies if I'm mistaken but I don't think this can be hit:

* `extendee_spec` is set if there is any instance of `python` in the concretization group (as determined in `external_spec_selected`)
* If `python` is in `self.spec.root`, the previous statement would always be true (i.e. the existence of a `python` attached to the root would mean there was a `python` that is part of the concretization group)

Furthermore, I added a debug breakpoint here and never hit it with `spack unit-test -k concretize.py` or `spack unit-test` (although I will note that the coverage results indicate that line was visited somehow).

If this check is not required, is it still required to handle `external_spec_selected` after dependency management in `sort_fn`?

scheibelp (commented):
I think there may be a redundant check: see comment

lib/spack/spack/build_systems/python.py
becker33: @scheibelp 

It is hit by the old concretizer.

The ordering in `sort_fn` is still required, because the code in `external_spec_selected` calls `extendee_spec`, and determining whether a spec extends another requires knowing the dependency information.

scheibelp (approved):

lib/spack/spack/test/concretize.py
```diff
@@ -1953,3 +1956,106 @@ def test_external_python_extensions_have_dependency(self):
...
         assert spec["python"] == spec["py-extension1"]["python"]
+
+    target = spack.platforms.test.Test.default
+
+    @pytest.mark.parametrize(
+        "python_spec",
+        [
+            "python@configured",
+            "python@configured platform=test",
+            "python@configured os=debian",
```
alalazo, line 1967: @scheibelp @becker33 Is it a typo or is it wanted to use `os=debian` instead of `os=debian6`? Found this while debugging a couple of unit test failures I still have in the branch I'm working on.

alalazo:
develop (fce95e2efb76da3c6ee549f2a9b4e257fb42ae50, blue) vs. PR (08101639cd3af48d101410a03f4791ca13dde2ea, orange) 

[radiuss_develop.csv](https://github.com/spack/spack/files/11087696/radiuss_develop.csv)
[radiuss_34202.csv](https://github.com/spack/spack/files/11087697/radiuss_34202.csv)
[radiuss.txt](https://github.com/spack/spack/files/11087698/radiuss.txt)

No effect on concretization time for this benchmark:

totals

PR #34203, opened 2022-11-30
py-twine: add 4.0.1, py-readme-renderer: add 37.3

manuelakuhn:
https://github.com/pypa/twine/tree/4.0.1
https://github.com/pypa/readme_renderer/tree/37.3

### DIFF
--- a/var/spack/repos/builtin/packages/py-twine/package.py
+++ b/var/spack/repos/builtin/packages/py-twine/package.py
@@ -11,12 +11,26 @@ class PyTwine(PythonPackage):
 
     homepage = "https://twine.readthedocs.io/"
     pypi = "twine/twine-2.0.0.tar.gz"
+    git = "https://github.com/pypa/twine.git"
 
+    version("4.0.1", sha256="96b1cf12f7ae611a4a40b6ae8e9570215daff0611828f5fe1f37a16255ab24a0")
     version("2.0.0", sha256="9fe7091715c7576df166df8ef6654e61bada39571783f2fd415bdcba867c6993")
 
+    depends_on("python@3.7:", when="@4:", type=("build", "run"))
     depends_on("python@3.6:", type=("build", "run"))
-    depends_on("py-pkginfo@1.4.2:", type=("build", "run"))
-    depends_on("py-readme-renderer@21.0:", type=("build", "run"))
-    depends_on("py-requests-toolbelt@0.8.0:0.8,0.9.1:", type=("build", "run"))
+    depends_on("py-setuptools@45:", when="@3.4.1:", type="build")
     depends_on("py-setuptools@0.7.0:", type=("build", "run"))
-    depends_on("py-tqdm@4.14:", type=("build", "run"))
+    depends_on("py-setuptools-scm+toml@6:", when="@3.4.2:", type="build")
+
+    depends_on("py-pkginfo@1.8.1:", when="@3.7:", type=("build", "run"))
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/py-twine/package.py
```diff
@@ -11,12 +11,26 @@ class PyTwine(PythonPackage):
...
     homepage = "https://twine.readthedocs.io/"
     pypi = "twine/twine-2.0.0.tar.gz"
+    git = "https://github.com/pypa/twine.git"
 
+    version("4.0.1", sha256="96b1cf12f7ae611a4a40b6ae8e9570215daff0611828f5fe1f37a16255ab24a0")
     version("2.0.0", sha256="9fe7091715c7576df166df8ef6654e61bada39571783f2fd415bdcba867c6993")
 
+    depends_on("python@3.7:", when="@4:", type=("build", "run"))
     depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-setuptools@45:", when="@3.4.1:", type="build")
```
adamjstewart, line 21:
```suggestion
    depends_on("py-setuptools@45:", when="@3.4.1:", type=("build", "run"))
```

manuelakuhn: Just for my understanding: why is `py-setuptools` needed here as a run dependency? I don't see it in install_requires.

adamjstewart: Hmm, maybe I'm mistaken. At one time, for older versions of Python or maybe setuptools, setuptools was needed at run-time for packages that use entrypoints. The auto-generated files in `prefix.bin` contained imports of setuptools or pkg_resources. I would suggest trying to install older twine with older Python and older setuptools and see if you can reproduce this. Maybe @pradyunsg would know if or when this changed.

Note that the line below this implies that setuptools is always a build/run dep, so you should either make it conditional on version or always build-only if you want the latest version to be build-only.

manuelakuhn: I checked 
- `py-twine@4.0.1 ^py-setuptools@65.5.0 ^python@3.10.8`
- `py-twine@2.0.0 ^py-setuptools@49.6.0 ^python@3.7`
- `py-twine@2.0.0 ^py-setuptools@46.1.3 ^python@3.7`
- `py-twine@2.0.0 ^py-setuptools@44.1.1 ^python@3.7`
- `py-twine@2.0.0 ^py-setuptools@43.0.0 ^python@3.7`

and could not find an import of setuptools or pkg_resources in the bin directories.

With older `py-setuptools` I could not test because `py-twine` does not concretize with them for me (Error: No version for 'py-setuptools' satisfies '@40.8.0' and '@65.5.0').
I also checked the packages I have currently installed on my test system (only ~500): 
`grep -r "import.*pkg_resource" opt/spack/linux-fedora37-skylake/*/*/bin` as well as `grep -r "import.*setuptools" opt/spack/linux-fedora37-skylake/*/*/bin`, except of a hit in `py-cython`, did not give any further result. So the need for setuptools as a run dependency seems to have changed.

> Note that the line below this implies that setuptools is always a build/run dep, so you should either make it conditional on version or always build-only if you want the latest version to be build-only.

Thanks, I missed that. With the above, I think we can change `py-setuptools` to be a build-only dependency. The above combinations I tested and `twine --help` worked as well without `py-setuptools` as run dependency.

adamjstewart (approved):

PR #34204, opened 2022-11-30
Clean: support removing cached environment includes with downloads

tldahlgren:
@tuxfan 

This PR adds support for removing the current/active environment's cached/downloaded include files fetched from a URL. 

```
$ spack clean -d
==> Removing cached downloads
==> Removing cached, included config files for '<env-name>' environment
```

If there are configuration file-related issues preventing the activation of an environment, a hint is provided explaining the steps to take to remove that environment's configuration files.  For example:

```
$ spacktivate my-env
==> Warning: config:module_roots has been replaced by modules:[module set]:roots and is ignored
==> Error: [Errno 13] Permission denied: '/spack'

$ spack clean -d
==> Removing cached downloads
==> Warning: An active environment is required to remove cached, included config files.  If you cannot activate (e.g., unsupported schema, permissions), try 'spack -e <environment-name> clean -d'.

$ spack -e my-env clean -d
...
==> Warning: config:module_roots has been replaced by modules:[module set]:roots and is ignored
==> Removing cached downloads
==> Removing cached, included config files for 'my-env' environment
```


tldahlgren:
Not sure why pipeline is stuck in pending.

@spackbot run pipeline

tldahlgren:
@spackbot run pipeline

scheibelp:
Note `spack clean -d` currently (outside of this PR) removes downloaded sources (i.e. the tarball with the source used to build a package): users may not want to do this as part of refreshing their included files (if I understand, this is required to get the latest instance of a remote config, and otherwise the user would be working with a stale copy if the remote has been updated).

I think a separate option would be useful vs. adding this to `clean -d`.

lib/spack/spack/cmd/clean.py
```diff
@@ -88,6 +88,26 @@ def remove_python_cache():
...
 
 
+def remove_downloads():
+    """Remove all (relevant) cached download files.
+
+    This removes files in the fetch cache and, if relevant, the active
+    environment's included configuration files cache."""
+    tty.msg("Removing cached downloads")
+    spack.caches.fetch_cache.destroy()
+    what = "cached, included config files"
```
scheibelp, line 98: (minor - redundant if removal of remote configs is made into a separate option) I think the comma implies we are removing two separate things (i.e. "removing cached, included config files") but my understanding is this is one type of thing: we are removing the included config files that have been cached from remote locations.

lib/spack/spack/cmd/clean.py
```diff
@@ -88,6 +88,26 @@ def remove_python_cache():
...
+    env = spack.environment.active_environment()
+    if env:
+        tty.msg("Removing {0} for '{1}' environment".format(what, env.name))
+        spack.environment.clear_included_configs()
```
scheibelp, line 102: Usually the clean options are global (at least without further specification) e.g. "remove *all* downloaded tarballs" if you run `spack clean -d`. For directory-based environments, I think you're doing the best that can be done (since they are not tracked by Spack). However, for Spack-managed environments I think it could make sense to have this option clear cached include files for all of them.

If - for simplicity of implementation - this should work on just the current active environment, then IMO the `else:` part of this statement should raise an error (there is a `require_active_environment` or similarly-named function available for this purpose).

lib/spack/spack/environment/environment.py
```diff
@@ -907,6 +907,10 @@ def repo(self):
             self._repo = make_repo_path(self.repos_path)
         return self._repo
 
+    def do_clean(self):
```
scheibelp, line 910: IMO this should have a more-specific name.

My recommendation is to entirely replace this with `clear_included_configs` (make it an instance method): as a (minor) side effect, that would remove the redundancy of retrieving the active environment here and in the `clean` command, but more importantly enables doing this for inactive environments as well.

lib/spack/spack/test/cmd/clean.py
```diff
@@ -109,3 +110,23 @@ def _check_files(directory):
...
+    env = spack.environment.Environment(cache_dir)
+    spack.environment.activate(env)
+
+    # Assumes the test automatically uses spack.test.conftest.MockCache
+    # with destroy() as a noop.
+    spack.cmd.clean.remove_downloads()
+
+    # Ensure the cached files no longer exist
+    for f in cache_files:
+        fs.touch(fs.join_path(cache_dir, f))
```
scheibelp, line 132: I'm assuming the idea is that this would fail if the file already exists, but`touch` does not in fact fail if the file already exists. I think a check like `assert not os.path.exists(...)` is more straightforward.

lib/spack/spack/test/cmd/clean.py
```diff
@@ -109,3 +110,23 @@ def _check_files(directory):
 
     for d in [source_dir, var_dir]:
         _check_files(d)
+
+
+def test_remove_env_includes(tmpdir):
+    cache_dir = fs.join_path(tmpdir, "var", "spack", "environment", "included")
+    cache_files = ["config.yaml", "packages.yaml"]
```
scheibelp, line 117: (question) If I understand, these files are not fetched from some remote location (or at least, it isn't clear to me that is the case since the environment specification doesn't mention it): if they were the only copy, then what happens? Or is the idea that if config files were included from a remote, that this is where they would be stored? If so, then this is fine, but a comment would be useful to indicate how remote configs are being simulated.

lib/spack/spack/test/cmd/clean.py
```diff
@@ -109,3 +110,23 @@ def _check_files(directory):
...
+    fs.mkdirp(cache_dir)
+    for f in cache_files:
+        fs.touch(fs.join_path(cache_dir, f))
+
+    env = spack.environment.Environment(cache_dir)
+    spack.environment.activate(env)
+
+    # Assumes the test automatically uses spack.test.conftest.MockCache
+    # with destroy() as a noop.
+    spack.cmd.clean.remove_downloads()
```
scheibelp, line 128: I think it would be worthwhile to replace this with `clean('-the-option')` (this makes sure that the option is handled as expected)

scheibelp (changes requested):

PR #34209, opened 2022-11-30
XRootD: add checksum + patch for 5.5.1

iarspider:


tldahlgren (dismissed):
Confirmed the new sha256.  (FWIW Was unable to fetch/download the package due to a time out.)

var/spack/repos/builtin/packages/xrootd/package.py
```diff
@@ -15,6 +15,7 @@ class Xrootd(CMakePackage):
     url = "http://xrootd.org/download/v5.3.1/xrootd-5.3.1.tar.gz"
     list_url = "https://xrootd.slac.stanford.edu/dload.html"
```
tldahlgren: We generally prefer that fetch (of a version) use the `https` protocol for downloads.  You could change the `url` ~and remove the `list_url`~ as follows and all of the checksums are still valid:

```suggestion
    url = "https://xrootd.slac.stanford.edu/download/v5.5.1/xrootd-5.5.1.tar.gz"
```

iarspider:
@tldahlgren out of curiosity - why did you remove `list_url`?

tldahlgren:
> @tldahlgren out of curiosity - why did you remove `list_url`?

Apologies.  I should've left it in for 'spack checksum rooted' to list the package versions.

tldahlgren (approved):

PR #34217, opened 2022-11-30
Add conflicts between gcc@12.2.0 and rocblas@5.2.1:5.2.3

aurianer:
Fails with error:
```
include/c++/12.2.0/bits/shared_ptr_base.h:196:22: error: use of undeclared identifier 'noinline'; did you mean 'inline'?
     1456          __attribute__((__noinline__))
     1457                         ^
include/hip/amd_detail/host_defines.h:50:37: note: expanded from macro '__noinline__'
     1459    #define __noinline__ __attribute__((noinline))
     1460                                        ^
```

### DIFF
--- a/var/spack/repos/builtin/packages/rocblas/package.py
+++ b/var/spack/repos/builtin/packages/rocblas/package.py
@@ -106,6 +106,8 @@ class Rocblas(CMakePackage):
     conflicts("amdgpu_target=gfx1011", when="@:4.2.1")
     conflicts("amdgpu_target=gfx1012", when="@:4.2.1")
     conflicts("amdgpu_target=gfx1030", when="@:4.2.1")
+    # https://reviews.llvm.org/D124866
+    conflicts("%gcc@12", when="@5.2.1:5.2.3")
 
     depends_on("cmake@3.16.8:", type="build", when="@4.2.0:")
     depends_on("cmake@3.8:", type="build", when="@3.9.0:")
--- a/var/spack/repos/builtin/packages/rocblas/package.py
+++ b/var/spack/repos/builtin/packages/rocblas/package.py
@@ -107,6 +107,8 @@ class Rocblas(CMakePackage):
     conflicts("amdgpu_target=gfx1012", when="@:4.2.1")
     conflicts("amdgpu_target=gfx1030", when="@:4.2.1")
     # https://reviews.llvm.org/D124866
+    # https://github.com/ROCm-Developer-Tools/HIP/issues/2678
+    # https://github.com/ROCm-Developer-Tools/hipamd/blob/rocm-5.2.x/include/hip/amd_detail/host_defines.h#L50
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/rocblas/package.py
```diff
@@ -106,6 +106,7 @@ class Rocblas(CMakePackage):
     conflicts("amdgpu_target=gfx1011", when="@:4.2.1")
     conflicts("amdgpu_target=gfx1012", when="@:4.2.1")
     conflicts("amdgpu_target=gfx1030", when="@:4.2.1")
+    conflicts("%gcc@12.2.0", when="@5.2.1:5.2.3")
```
msimberg:
```suggestion
    conflicts("%gcc@12", when="@5.2.1:5.2.3")
```
as I'm quite sure this is going to fail with any minor version of GCC 12.

Could you also add a link e.g. to https://reviews.llvm.org/D124866 to explain the conflict?

For context for others, this _should_ be fixed in the upcoming HIP 5.3.0 (based on llvm 15) but hasn't been tested. That's why the upper bound is the restrictive 5.2.3. If it turns out to be needed with 5.3.0 as well we can always change the bound later.

msimberg (changes requested):

var/spack/repos/builtin/packages/rocblas/package.py
msimberg: To add more context, compiling libstdc++ version 12's `<memory>` header with `hipcc` is what fails to compile. In that sense this could be a blanket conflict between GCC 12 and HIP <= 5.2. However, it seems like rocBLAS doesn't include `<memory>` in every version (at least in the files that are compiled by `hipcc`) so there's also a lower bound on the version in the conflict.

Other projects that attempt to compile `<memory>` with `hipcc` and GCC 12 are likely to hit the same issue though.

cgmb (commented):
If the incompatibility is between HIP and libstdc++, why are we putting the conflict directive in rocBLAS? I suppose it's rare for it to matter because few libraries use the broken `__noinline__`? EDIT: I see this was already [addressed in another comment](https://github.com/spack/spack/pull/34217#discussion_r1036151110).

msimberg:
> If the incompatibility is between HIP and libstdc++, why are we putting the conflict directive in rocBLAS? I suppose it's rare for it to matter because few libraries use the broken `__noinline__`?

Only because it happens to be triggered while building `rocblas` but if you see https://github.com/spack/spack/pull/34217#discussion_r1036151110 we could also just make it a conflict for base HIP itself (in which case it should be for all versions less than 5.3). In theory one can use GCC 12 with the problematic versions as long as one doesn't include `<memory>` in the wrong place... I'm fine with either approach.

cgmb (approved):
I have no strong opinions about whether this should be in HIP or rocBLAS. So, I'm fine with this proposal.

msimberg (approved):

PR #34225, opened 2022-11-30
Bugfix: Fetch should not force use of curl to check url existence

tldahlgren:
This PR does not force `fetch` to use `curl` to check URL existence.

Bug was introduced in #29026 as discussed in https://github.com/spack/spack/pull/29026#discussion_r1036427946.

### DIFF
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -335,7 +335,7 @@ class URLFetchStrategy(FetchStrategy):
         url = None
         errors = []
         for url in self.candidate_urls:
-            if not web_util.url_exists(url, self.curl):
+            if not web_util.url_exists(url):
                 tty.debug("URL does not exist: " + url)
                 continue
 
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -106,12 +106,12 @@ class FetchStrategy(object):
     #: The URL attribute must be specified either at the package class
     #: level, or as a keyword argument to ``version()``.  It is used to
     #: distinguish fetchers for different versions in the package DSL.
-    url_attr = None  # type: Optional[str]
+    url_attr: Optional[str] = None
 
     #: Optional attributes can be used to distinguish fetchers when :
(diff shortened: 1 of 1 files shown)


tldahlgren:
@alalazo @trws What fun!  `flake8` doesn't like the `typing` imports:

```
==> Error: spack style found errors
[30](https://github.com/spack/spack/actions/runs/3587158973/jobs/6037191189#step:6:31)
lib/spack/spack/fetch_strategy.py:33: [F401] 'typing.List' imported but unused
[31](https://github.com/spack/spack/actions/runs/3587158973/jobs/6037191189#step:6:32)
lib/spack/spack/fetch_strategy.py:33: [F401] 'typing.Optional' imported but unused
```

trws:
Thankfully now that we don't have to deal with ancient, pre-typing, python those are a pretty easy fix.  Make the type-hint a type-hint rather than a comment and it should pass.

tldahlgren:
@alalazo @trws Looks like project coverage dropped due to "indirect coverage changes".

alalazo:
I don't think coverage is essential here. What we may have dropped are just the lines in `self.curl`, or am I missing anything?

tldahlgren:
> I don't think coverage is essential here. What we may have dropped are just the lines in `self.curl`, or am I missing anything?

I don't think you missed anything.  The value for that arg defaults to `None` is not provided so I just dropped it.

I suppose I could add a test of that method -- preferably mocking `which` -- to see if that helps with coverage.  Although I didn't look closely at all of the indirect coverage changes (once I was able to see them), I didn't quite follow how dropping that method call mattered.

alalazo (approved):
2 for `mypy` and one for the bug! (Thanks Tammy!)

tldahlgren:
> 2 for `mypy` and one for the bug! (Thanks Tammy!)

I grep'd for places where curl is used and, AFAICT, the rest were gated by the configuration parameter.  So yes, a one line fix as you'd pointed out.

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

tldahlgren:
@alalazo Thank you!

PR #34230, opened 2022-12-01
redis: newer version and added TLS support

lpottier:

### DIFF
--- a/var/spack/repos/builtin/packages/redis/package.py
+++ b/var/spack/repos/builtin/packages/redis/package.py
@@ -18,8 +18,22 @@ class Redis(MakefilePackage):
     """
 
     homepage = "https://redis.io"
-    url = "https://download.redis.io/releases/redis-6.2.5.tar.gz"
+    urls = [
+        "https://download.redis.io/releases/redis-7.0.5.tar.gz",
+        "https://github.com/redis/redis/archive/refs/tags/7.0.5.tar.gz",
+    ]
+    git = "https://github.com/redis/redis.git"
 
+    maintainers = ["lpottier"]
+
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/redis/package.py
```diff
@@ -36,6 +50,25 @@ class Redis(MakefilePackage):
     version("4.0.12", sha256="6447259d2eed426a949c9c13f8fdb2d91fb66d9dc915dd50db13b87f46d93162")
     version("4.0.11", sha256="fc53e73ae7586bcdacb4b63875d1ff04f68c5474c1ddeda78f00e5ae2eed1bbb")
 
+    variant("tls", default=False, description="Builds with TLS support")
```
alalazo:
```suggestion
    variant("tls", default=False, when="@6:", description="Builds with TLS support")
```

var/spack/repos/builtin/packages/redis/package.py
```diff
@@ -36,6 +50,25 @@ class Redis(MakefilePackage):
...
+    depends_on("openssl@1.1:", type=("build", "link"), when="+tls")
+    conflicts("@:6", when="+tls", msg="TLS support requires Redis >= 6.0.")
```
alalazo:
```suggestion
```
A note that this includes all the versions in the 6 major release. According to the comment I think you wanted `@:5`.

alalazo (changes requested):
A minor request to use a conditional variant in place of a variant and a conflict.

lpottier:
@alalazo Thanks for suggestions!

alalazo (approved):

PR #34234, opened 2022-12-01
Add recipe for singularity-hpc, py-spython

marcodelapierre:
Hey @vsoch do you want to provide feedback this one? :-)


vsoch:
This is fantastic! I think @marcodelapierre and I could be maintainers  too?

I'll give it a test build in a few minutes!

vsoch:
okay testing now - sorry fighting other spack fires :fire:

var/spack/repos/builtin/packages/singularity-hpc/package.py
```diff
@@ -0,0 +1,42 @@
...
+
+    variant(
+        "runtime",
+        default="none",
+        description="Container runtime installed by Spack for this package",
+        values=("none", "singularityce", "singularity"),
+        multi=False,
+    )
+
+    depends_on("singularityce@3:", when="runtime=singularityce", type="run")
```
vsoch: Build works!

We probably want to have podman be an option here too https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages/podman. And agree on the default choice!

And then what about the module software? Should we have a similar variant with lmod/environment modules but default to none? I suspect most installing this will have their own module software, but perhaps it should be an option to install here too.

marcodelapierre:
@vsoch I have added you as maintainer :-)

var/spack/repos/builtin/packages/singularity-hpc/package.py
marcodelapierre: cool, I hadn't noticed that `podman` had made it into Spack. Variant added!

marcodelapierre: added `modules` variant, too

marcodelapierre:
ok @vsoch I think I have addressed all your points for now

marcodelapierre:
thanks for your feedback by the way!

marcodelapierre:
@vsoch could you act as reviewer for this PR?
or shall I ping @alalazo ?

thanks!

marcodelapierre:
hi @vsoch, let me know if anything else is required here.

tldahlgren (dismissed):
Confirmed the sha256 for the two packages.

tldahlgren:
Adding @adamjstewart in case he wants to comment on the new python package.

var/spack/repos/builtin/packages/py-spython/package.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    depends_on("singularityce@3.5.2:", when="runtime=singularityce", type="run")
+    depends_on("singularity@3.5.2:", when="runtime=singularity", type="run")
+
+    depends_on("python", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+
+    depends_on("py-requests", type=("build", "run"))
+    depends_on("py-semver", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-semver@2.8.1:", type=("build", "run"))
```
This can be found in `spython/version.py`

var/spack/repos/builtin/packages/py-spython/package.py
```diff
@@ -0,0 +1,33 @@
...
+        multi=False,
+    )
+
+    depends_on("singularityce@3.5.2:", when="runtime=singularityce", type="run")
+    depends_on("singularity@3.5.2:", when="runtime=singularity", type="run")
+
+    depends_on("python", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+
+    depends_on("py-requests", type=("build", "run"))
```
adamjstewart: This package does not appear to use the requests library

var/spack/repos/builtin/packages/py-spython/package.py
```diff
@@ -0,0 +1,33 @@
...
+        default="none",
+        description="Container runtime installed by Spack for this package",
+        values=("none", "singularityce", "singularity"),
+        multi=False,
+    )
+
+    depends_on("singularityce@3.5.2:", when="runtime=singularityce", type="run")
+    depends_on("singularity@3.5.2:", when="runtime=singularity", type="run")
+
+    depends_on("python", type=("build", "run"))
```
adamjstewart: This is already added by the `PythonPackage` base class

var/spack/repos/builtin/packages/py-spython/package.py
```diff
@@ -0,0 +1,33 @@
...
+    variant(
+        "runtime",
+        default="none",
+        description="Container runtime installed by Spack for this package",
+        values=("none", "singularityce", "singularity"),
+        multi=False,
+    )
+
+    depends_on("singularityce@3.5.2:", when="runtime=singularityce", type="run")
+    depends_on("singularity@3.5.2:", when="runtime=singularity", type="run")
```
adamjstewart, line 27: Should we make a virtual provider for singularity so we don't need this complexity?

var/spack/repos/builtin/packages/py-spython/package.py
marcodelapierre: I got it from the `requirements.txt` file of the package: https://github.com/singularityhub/singularity-cli/blob/master/requirements.txt

I suggest leaving it, to be consistent with that specification by the package authors.

Oh well, @vsoch is the author! What do you think?

var/spack/repos/builtin/packages/py-spython/package.py
marcodelapierre: fixed, see new commit.

var/spack/repos/builtin/packages/py-spython/package.py
adamjstewart: The `requirements.txt` is incorrect, if you recursively grep the code, it isn't used anywhere. Also, `requirements.txt` isn't used by the build system in any way.

var/spack/repos/builtin/packages/py-spython/package.py
marcodelapierre: Good point. Seems out of scope for this PR though; I strongly suggest leaving it to a separate one. 

Opening a github issue about it: #34386

var/spack/repos/builtin/packages/py-spython/package.py
marcodelapierre: edited, see new commit

var/spack/repos/builtin/packages/py-spython/package.py
adamjstewart: Not necessarily out of scope, could even be fewer lines of code than all of the multi-value variant stuff. Just a suggestion, not a requirement.

var/spack/repos/builtin/packages/py-spython/package.py
marcodelapierre: thanks for the insight Adam.

just edited in a new commit.

@vsoch I am opening an issue in the spython repo to remind you about updating the requirements file.

var/spack/repos/builtin/packages/py-spython/package.py
marcodelapierre: The reason why I prefer to keep this aspect distinct is that I don't have the required knowledge of these two projects, to be able to guarantee that their APIs/user interfaces are, and crucially will be in the future, compatible with each other.  
Singularity is stale, as it has forked into SingularityCE and Apptainer, and I am concerned that future developments of SingularityCE may make it not compatible with the original Singularity.

adamjstewart (approved):

vsoch:
Thanks @adamjstewart @tldahlgren ! And we just removed the semver requirement so I can do a quick PR to update the recipe.

PR #34235, opened 2022-12-01
Fix month in citation file

adamjstewart:
According to https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsreferencemonth, month should be an integer, not a string.


tgamblin:
This breaks the bibtex output — see https://github.com/citation-file-format/ruby-cff/issues/79#issuecomment-937330391

markcmiller86:
I see this was left unmerged due to a bibtex issue.

Unaware of this, I submitted a PR with a larger change to the file that I've also validated with `convercff` in #40363

PR #34236, opened 2022-12-01
Stand-alone testing: Make recipe support and processing spack-/pytest-like

### LINKED ISSUE #35519: spack test: Spec.package can only be called on concrete specs
### Steps to reproduce

Unclear exactly which packages are required to reproduce this. I have a massive environment I'm testing:
```yaml
spack:
  specs:
  - gdal
  - opencv
  - py-black
  - py-cartopy
  - py-cmocean
  - py-einops
  - py-fiona
  - py-flake8
  - py-geocube
  - py-geopandas
  - py-h5py
  - py-ipywidgets
  - py-isort
  - py-jax
  - py-kornia
  - py-laspy
  - py-matplotlib
  - py-mypy
  - py-nbmake
  - py-nbsphinx
  - py-netcdf4
  - py-numpy
  - py-omegaconf
  - py-pandas
  - py-pillow-simd
  - py-pycocotools
  - py-pydocstyle
  - py-pygeos
  - py-pyproj
  - py-pyro-ppl
  - py-pytest
  - py-pytest-cov
  - py-pytorch-lightning
  - py-pytorch-sphinx-theme
  - py-pyupgrade
  - py-radiant-mlhub
  - py-rarfile
  - py-rasterio
  - py-rtree
  - py-setuptools
  - py-scikit-image
  - py-scikit-learn
  - py-scipy
  - py-segmentation-models-pytorch
  - py-shapely
  - py-sphinx
  - py-tensorboard
  - py-timm
  - py-torch
  - py-torchmetrics
  - py-torchvision
  - py-xarray
  - py-zipfile-deflate64
  view: true
  concretizer:
    unify: true
```
I run `spack install` and `spack test run`.

### Error message

```console
$ spack -d test run
...
==> [2023-02-16-11:55:18.673419] Installing /projects/dali/spack/var/spack/repos/builtin/packages/openjdk/package.py to /home/adamjs5/.spack/test/32ccsglunnecbl3oi2fztym6ti6alinu/repo/openjdk
==> [2023-02-16-11:55:18.677207] Installing /projects/dali/spack/var/spack/repos/builtin/packages/cuda/package.py to /home/adamjs5/.spack/test/32ccsglunnecbl3oi2fztym6ti6alinu/repo/cuda
  File "/projects/dali/spack/bin/spack", line 54, in <module>
    sys.exit(main())
  File "/projects/dali/spack/lib/spack/spack_installable/main.py", line 37, in main
    sys.exit(spack.main.main(argv))
  File "/projects/dali/spack/lib/spack/spack/main.py", line 1004, in main
    return _main(argv)
  File "/projects/dali/spack/lib/spack/spack/main.py", line 959, in _main
  File "/projects/dali/spack/lib/spack/spack/main.py", line 987, in finish_parse_and_run
  File "/projects/dali/spack/lib/spack/spack/main.py", line 636, in _invoke_command
  File "/projects/dali/spack/lib/spack/spack/cmd/test.py", line 476, in test
    globals()["test_%s" % args.test_command](args)
  File "/projects/dali/spack/lib/spack/spack/cmd/test.py", line 226, in test_run
    externals=args.externals,
  File "/projects/dali/spack/lib/spack/spack/install_test.py", line 169, in __call__
    self.write_reproducibility_data()
  File "/projects/dali/spack/lib/spack/spack/install_test.py", line 393, in write_reproducibility_data
    spack.repo.path.dump_provenance(vspec, repo_cache_path)
  File "/projects/dali/spack/lib/spack/spack/repo.py", line 304, in converter
    return function(self, spec_like, *args, **kwargs)
  File "/projects/dali/spack/lib/spack/spack/repo.py", line 859, in dump_provenance
    return self.repo_for_pkg(spec).dump_provenance(spec, path)
  File "/projects/dali/spack/lib/spack/spack/repo.py", line 304, in converter
    return function(self, spec_like, *args, **kwargs)
  File "/projects/dali/spack/lib/spack/spack/repo.py", line 1068, in dump_provenance
    for patch in itertools.chain.from_iterable(spec.package.patches.values()):
  File "/projects/dali/spack/lib/spack/spack/spec.py", line 1615, in package
    assert self.concrete, "Spec.package can only be called on concrete specs"
AssertionError: Spec.package can only be called on concrete specs
```
I tried `spack clean -m` in case the cache contained an old patch that no longer existed but it didn't help.

* **Spack:** 0.20.0.dev0
* **Python:** 3.6.8
* **Platform:** linux-rhel7-zen3
* **Concretizer:** clingo

@tldahlgren 

### General information

tldahlgren:
fixes #35519

This PR is a refactor of Spack's stand-alone test process to be more spack- and pytest-like.  It is more spack-like in that test parts are no longer "hidden" in a package's `run_test()` method and pytest-like in that any package method whose name starts `test_` (i.e., a "test" method) is a test part.  We also support the ability to embed test parts in a test method when that makes sense.

Test methods are now implicit test parts.  The docstring is the purpose for the test part.  The name of the method is the name of the test part.  The working directory is the active spec's test stage directory.  You can embed test parts using the `test_part` context manager.

This PR adds other functionality (in part because it was originally tested here).

PR highlights:
- adds support for multiple `test_*` stand-alone package test methods, each of which is an implicit `test_part` for execution and reporting purposes;
- deprecates package use of `run_test()`;
- exposes some functionality from `run_test()` as optional helper methods;
- adds a `SkipTest` exception that can be used to flag stand-alone tests as being skipped;
- updates the packaging guide section on stand-alone tests to provide more examples (per feedback from multiple people);
- restores the ability to run tests "inherited" from provided virtual packages (was also in #35380);
- prints the test log path (like we currently do for build log paths) (was also in #35315);
- times and reports the post-install process (since it can include post-install tests) (was also in #35315); and
- corrects context-related error message to distinguish test recipes from build recipes.

### DIFF
--- a/lib/spack/docs/packaging_guide.rst
+++ b/lib/spack/docs/packaging_guide.rst
@@ -5074,9 +5074,23 @@ installed example.
            example = which(self.prefix.bin.example)
            example()
 
-Spack outputs ``test_always_fails: use assert to always fail`` before running
-``test_always_fails``. Similarly, it logs ``test_example: run installed example`` 
-before running ``test_example``.
+Output showing the identification of each test part after runnig the tests
+is illustrated below.
+
+.. code-block:: console
+
+   $ spack test run --alias mypackage mypackage@1.0
+   ==> Spack test mypackage
+   ...
+   $ spack test results -l mypackage
+   ==> Results for test suite 'mypackage':
+   ...
+   ==> [2023-03-10-16:03:56.625204] test: test_always_fails: use assert to always fail
+   ...
+   FAILED
+   ==> [2023-03-10-16:03:56.625439] test: test_example: run installed example
+   ...
+   PASSED
 
 
 .. note::
@@ -5139,11 +5153,28 @@ embedded test parts.
                     exe = which(join_path(self.prefix.bin, example))
                     exe()
 
-Spack will output ``test_example: run installed examples`` before executing
-``test_example``, then ``test_example_ex1: run installed ex1`` before the first
-test part and ``test_example_ex2: run installed ex2`` before the second test
-part. The test ``test_example_ex2`` will run whether or not ``test_example_ex1``
-succeeds.
+In this case, there will be an implicit test part for ``test_example``
+and separate sub-parts for ``ex1`` and ``ex2``. The second sub-part
+will be executed regardless of whether the first passes. The test
+log for a run where the first executable fails and the second passes
+is illustrated below.
+
+.. code-block:: console
+
+   $ spack test run --alias mypackage mypackage@1.0
+   ==> Spack test mypackage
+   ...
+   $ spack test results -l mypackage
+   ==> Results for test suite 'mypackage':
+   ...
+   ==> [2023-03-10-16:03:56.625204] test: test_example: run installed examples
+   ==> [2023-03-10-16:03:56.625439] test: test_example_ex1: run installed ex1
+   ...
+   FAILED
+   ==> [2023-03-10-16:03:56.625555] test: test_example_ex2: run installed ex2
+   ...
+   PASSED
+   ...
 
 .. warning::
 
@@ -5262,16 +5293,17 @@ and using ``foo.c`` in a test method is illustrated below.
            src_dir = join_path(
                self.test_suite.current_test_cache_dir, "examples"
            )
-           cc = which(os.environ["CC"])
-           cc(
-               "-L{0}".format(self.prefix.lib),
-               "-I{0}".format(self.prefix.include),
-               "{0}.c".format(exe),
-               "-o",
-               exe
-           )
-           foo = which(exe)
-           foo()
+           with working_dir(src_dir):
+               cc = which(os.environ["CC"])
+               cc(
+                   "-L{0}".format(self.prefix.lib),
+                   "-I{0}".format(self.prefix.include),
+                   "{0}.c".format(exe),
+                   "-o",
+                   exe
+               )
+               foo = which(exe)
+               foo()
 
 In this case, the method copies the associated files from the build
 stage, **after** the software is installed, to the package's test
--- a/lib/spack/docs/packaging_guide.rst
+++ b/lib/spack/docs/packaging_guide.rst
@@ -26,7 +26,7 @@ package author can understand.  Packages allow the packager to
 encapsulate the build logic for different versions, compilers,
 options, platforms, and dependency combinations in one place.
 Essentially, a package translates a spec into build logic. It
-also allow the packager to write spec-specific tests of the
+also allows the packager to write spec-specific tests of the
 installed software.
 
 Packages in Spack are written in pure Python, so you can do anything
@@ -4973,82 +4973,92 @@ be left in the build stage directory as illustrated below:
 Stand-alone tests
 ^^^^^^^^^^^^^^^^^
 
-While build-time tests are integrated with the installation process,
-stand-alone tests are independent, allowing them to run days and weeks
-after the software is installed. Tests are defined in the package using
-methods whose names begin ``test_``, allowing for multiple independent
-checks. The methods themselves can contain embedded test parts. Files
-from the installation process can be saved off for use by the tests
-(e.g., compilation). Extra files, such as source, data, and expected
-outputs, can also be stored with the package. Both sets of files are
-automatically copied to the spec's test stage directory prior to execution
-of the test methods.
+While build-time tests are integrated with the build process, stand-alone
+tests are expected to run days, weeks, even months after the software is
+installed. The goal is to provide a mechanism for gaining confidence that
+packages work as installed **and** *continue* to work as the underlying
+software evolves. Packages can add and inherit stand-alone tests. The
+`spack test`` command is used to manage stand-alone testing.
 
 .. note::
 
-    Execution speed is important because these tests are intended to
-    quickly assess whether the installed spec works on the system.
-    Consequently, they should run relatively quickly -- as in on the
-    order of at most a few minutes while ideally executing all, or at
-    least key, aspects of the installed software.
+    Execution speed is important since these tests are intended to quickly
+    assess whether installed specs work on the system. Consequently, they
+    should run relatively quickly -- as in on the order of at most a few
+    minutes -- while ideally executing all, or at least key aspects of the
+    installed software.
 
-    Failing stand-alone tests are an indicator that there is an issue
-    with the installation and, therefore, no reason to proceed with more
-    resource-intensive tests.
+.. note::
 
-    Passing stand-alone tests can indicate that more thorough testing,
-    such as extensive unit or regression tests, or tests that run at
-    scale are in order.
+    Failing stand-alone tests indicate problems with the installation and,
+    therefore, there is no reason to proceed with more resource-intensive
+    tests until those have been investigated.
+
+    Passing stand-alone tests indicate that more thorough testing, such
+    as running extensive unit or regression tests, or tests that run at
+    scale can proceed without wasting resources on a problematic installation.
+
+Tests are defined in the package using methods with names beginning ``test_``.
+This allows Spack to support multiple independent checks, or parts. Files
+needed for testing, such as source, data, and expected outputs, may be saved
+from the build and or stored with the package in the repository. Regardless
+of origin, these files are automatically copied to the spec's test stage
+directory prior to execution of the test method(s). Spack also provides some
+helper functions to facilitate processing.
+
+.. _configure-test-stage:
 
 """"""""""""""""""""""""""""""""""""
 Configuring the test stage directory
 """"""""""""""""""""""""""""""""""""
 
-Stand-alone tests rely on a test stage directory for building, running,
-and tracking results in the same way Spack relies on a stage directory
-during builds. The default test stage, ``~/.spack/test``, is defined in
-:ref:`etc/spack/defaults/config.yaml <config-yaml>`. You can configure
-the location in the high-level ``config`` by adding or changing the
-``test_stage`` path in the appropriate ``config.yaml`` file such that:
+Stand-alone tests utilize a test stage directory for building, running,
+and tracking results in the same way Spack uses a build stage directory.
+The default test stage root directory, ``~/.spack/test``, is defined in
+:ref:`etc/spack/defaults/config.yaml <config-yaml>`. This location is
+customizable by adding or changing the ``test_stage`` path in the high-level
+``config`` of the appropriate ``config.yaml`` file such that:
 
 .. code-block:: yaml
 
    config:
      test_stage: /path/to/test/stage
 
-The package can access this path **during test processing** using
-`self.test_suite.stage`.
+Packages can use the ``self.test_suite.stage`` property to access this setting.
+Other package properties that provide access to spec-specific subdirectories
+and files are described in :ref:`accessing staged files <accessing-files>`.
 
 .. note::
 
-   The test stage path is established for the entire suite. That
-   means it is the root directory for **all specs** being tested
-   with the same `spack test run` command and each spec gets its
-   own subdirectory.
+   The test stage path is the root directory for the **entire suite**.
+   In other words, it is the root directory for **all specs** being
+   tested by the ``spack test run`` command. Each spec gets its own
+   stage subdirectory. Use ``self.test_suite.test_dir_for_spec(self.spec)``
+   to access the spec-specific test stage directory.
 
-   Use ``self.test_suite.test_dir_for_spec(self.spec)`` to access
-   the stage directory for the spec currently being tested. Refer to
-   :ref:`accessing staged files <accessing-files>` for more information.
 
+.. _adding-standalone-tests:
 
 """"""""""""""""""""""""
 Adding stand-alone tests
 """"""""""""""""""""""""
 
-Tests are defined in the package using methods with names beginning
-``test_``, which means multiple independent tests can be implemented
-to check the installation. The tests can be customized to the source
-build since Spack tracks options, compilers, and dependencies. Standard
-python ``assert`` statements and other error reporting mechanisms can
-be used in these methods. Spack will report such errors as test failures.
+Test recipes are defined in the package using methods with names beginning
+``test_``. This allows for the implementation of multiple independent tests.
+Each method has access to the information Spack tracks on the package, such
+as options, compilers, and dependencies, supporting the customization of tests
+to the build. Standard python ``assert`` statements and other error reporting
+mechanisms are available. Such exceptions are automatically caught and reported
+as test failures.
 
-Each test method is an implicit test part whose name is the name of
-the method and whose purpose is the method's docstring. The purpose is 
-provides context that can be helpful during debugging. Spack will output
-the test name and purpose prior to running the test part.
-
-For example, the ``MyPackage`` package below provides a skeleton for
-the test method.
+Each test method is an implicit test part named by the method and whose
+purpose is the method's docstring. Providing a purpose gives context for
+aiding debugging. A test method may contain embedded test parts. Spack
+outputs the test name and purpose prior to running each test method and
+any embedded test parts. For example, ``MyPackage`` below provides two basic
+examples of installation tests: ``test_always_fails`` and ``test_example``.
+As the name indicates, the first always fails. The second simply runs the
+installed example.
 
 .. code-block:: python
 
@@ -5064,9 +5074,19 @@ the test method.
            example = which(self.prefix.bin.example)
            example()
 
-You can implement multiple test parts within a stand-alone test method
-using the ``test_part`` context manager. Each test part is treated as
-an independent check with subsequent test executed even after a failure.
+Spack outputs ``test_always_fails: use assert to always fail`` before running
+``test_always_fails``. Similarly, it logs ``test_example: run installed example`` 
+before running ``test_example``.
+
+
+.. note::
+
+   If ``MyPackage`` were a recipe for a library, the tests should build
+   an example or test program that is then executed.
+
+A test method can include test parts using the ``test_part`` context manager.
+Each part is treated as an independent check to allow subsequent test parts
+to execute even after a test part fails.
 
 .. _test-part:
 
@@ -5074,40 +5094,23 @@ The signature for ``test_part`` is:
 
 .. code-block:: python
 
-   @contextlib.contextmanager
    def test_part(pkg, test_name, purpose, work_dir=".", verbose=False):
 
 where each argument has the following meaning:
 
 * ``pkg`` is an instance of the package for the spec under test.
 
-* ``test_name`` is the name of the test part and needs to start with ``test_``
+* ``test_name`` is the name of the test part, which must start with ``test_``.
 
-  The name of every ``test_`` method is automatically the name of
-  its implicit test part. 
-  
-  It is recommended that embedded test parts extend the name of the test
-  method. For example, if there are examples called ``ex10`` and ``ex51``
-  run within the ``test_example`` method, then the test parts would be
-  called ``test_example_ex10`` and ``test_example_ex51``, respectively.
+* ``purpose`` is a brief description used as a heading for the test part.
 
-* ``purpose`` is the heading describing the test part.
+  Output from the test is written to a test log file allowing the test name
+  and purpose to be searched for test part confirmation and debugging.
 
-  The docstring of every ``test_`` method is automatically the name of
-  its implicit test part.
+* ``work_dir`` is the path to the directory in which the test will run.
 
-  Output from the test is written to a test log file so this argument
-  serves as a searchable heading in text logs to highlight the start
-  of the test part. 
-  
-  Having a description can be helpful when debugging failing tests.
-
-* ``work_dir`` is the path to the directory from which the executable
-  will run.
-
-  The default of ``None`` corresponds to the current directory (``"."``),
-  which is the spec's test stage (i.e.,
-  ``self.test_suite.test_dir_for_spec(self.spec)``).
+  The default of ``None``, or ``"."``, corresponds to the the spec's test
+  stage (i.e., ``self.test_suite.test_dir_for_spec(self.spec)``.
 
 .. admonition:: Tests should **not** run under the installation directory.
 
@@ -5117,7 +5120,8 @@ where each argument has the following meaning:
    read-only file systems or directories.
 
 Suppose ``MyPackage`` actually installs two examples we want to use for tests.
-These checks could be embedded test parts implemented below.
+These checks can be implemented as separate checks or, as illustrated below,
+embedded test parts.
 
 .. code-block:: python
 
@@ -5135,13 +5139,24 @@ These checks could be embedded test parts implemented below.
                     exe = which(join_path(self.prefix.bin, example))
                     exe()
 
+Spack will output ``test_example: run installed examples`` before executing
+``test_example``, then ``test_example_ex1: run installed ex1`` before the first
+test part and ``test_example_ex2: run installed ex2`` before the second test
+part. The test ``test_example_ex2`` will run whether or not ``test_example_ex1``
+succeeds.
+
+.. warning::
+
+   Test results reporting requires that each test method and embedded
+   test part for a package have a unique name.
+
 Stand-alone tests run in an environment that provides access to information
 Spack has on how the software was built, such as build options, dependencies,
 and compilers. Build options and dependencies are accessed with the normal
 spec checks. Examples of checking :ref:`variant settings <variants>` and
-:ref:`spec constraints <testing-specs>` can be found at those links.
-Accessing compilers used by the build in stand-alone tests requires another
-step described :ref:`below <test-compilation>`.
+:ref:`spec constraints <testing-specs>` can be found at the provided links.
+Accessing compilers in stand-alone tests that are used by the build requires
+setting a package property as described :ref:`below <test-compilation>`.
 
 
 .. _test-compilation:
@@ -5150,20 +5165,21 @@ step described :ref:`below <test-compilation>`.
 Enabling test compilation
 """""""""""""""""""""""""
 
-If you want to build tests before running them, then you'll want to tell
+If you want to build and run binaries in tests, then you'll need to tell
 Spack to load the package's compiler configuration. This is accomplished
 by setting the package's ``test_requires_compiler`` property to ``True``.
-Doing so allows you to access the compiler through the canonical environment
-variables (e.g., ``CC``, ``CXX``, ``FC``, ``F77``). If appropriate, it
-also gives access to build dependencies like ``cmake`` through
-``self.spec["cmake"].prefix.bin.cmake``.
+
+Setting the property to ``True`` ensures access to the compiler through
+canonical environment variables (e.g., ``CC``, ``CXX``, ``FC``, ``F77``).
+It also gives access to build dependencies like ``cmake`` through their
+``spec objects`` (e.g., ``self.spec["cmake"].prefix.bin.cmake``).
 
 .. note::
 
-   We recommend adding the property at the top of the package with the
-   other attributes, such as ``homepage`` and ``url``.
+   The ``test_requires_compiler`` property should be added at the top of
+   the package near other attributes, such as the ``homepage`` and ``url``.
 
-Below is an example using this feature to compile an example.
+Below illustrates using this feature to compile an example.
 
 .. code-block:: python
 
@@ -5192,15 +5208,20 @@ Below is an example using this feature to compile an example.
 .. _cache_extra_test_sources:
 
 """""""""""""""""""""""
-Adding build-time files
+Saving build-time files
 """""""""""""""""""""""
 
 .. note::
 
-    We highly recommend re-using build-time tests and input files
-    for testing installed software. These files are easier to keep
-    synchronized since they reside within the software's repository
-    than maintaining custom install test files with the Spack package.
+    We highly recommend re-using build-time test sources and pared down
+    input files for testing installed software. These files are easier
+    to keep synchronized with software capabilities since they reside
+    within the software's repository. 
+    
+    If that is not possible, you can add test-related files to the package
+    repository (see :ref:`adding custom files <cache_custom_files>`). It
+    will be important to maintain them so they work across listed or supported
+    versions of the package.
 
 You can use the ``cache_extra_test_sources`` method to copy directories
 and or files from the source build stage directory to the package's
@@ -5212,15 +5233,17 @@ The signature for ``cache_extra_test_sources`` is:
 
    def cache_extra_test_sources(self, srcs):
 
-where ``srcs`` is a string *or* a list of strings corresponding to
-the paths for the files and or subdirectories, relative to the staged
-source, that are to be copied to the corresponding relative test path
-under the prefix. All of the contents within each subdirectory will
-also be copied.
+where ``srcs`` is a string *or* a list of strings corresponding to the
+paths of subdirectories and or files needed for stand-alone testing. 
+The paths must be relative to the staged source directory. Contents of
+subdirectories and files are copied to a special test cache subdirectory
+of the installation prefix. They are automatically copied to the appropriate
+relative paths under the test stage directory prior to executing stand-alone
+tests.
 
 For example, a package method for copying everything in the ``tests``
 subdirectory plus the ``foo.c`` and ``bar.c`` files from ``examples``
-can be implemented as shown below.
+and using ``foo.c`` in a test method is illustrated below.
 
 .. code-block:: python
 
@@ -5228,7 +5251,7 @@ can be implemented as shown below.
        ...
 
        @run_after("install")
-       def copy_test_sources(self):
+       def copy_test_files(self):
            srcs = ["tests",
                    join_path("examples", "foo.c"),
                    join_path("examples", "bar.c")]
@@ -5239,29 +5262,36 @@ can be implemented as shown below.
            src_dir = join_path(
                self.test_suite.current_test_cache_dir, "examples"
            )
-           ...
-           # TODO: Compile foo using source in src_dir
+           cc = which(os.environ["CC"])
+           cc(
+               "-L{0}".format(self.prefix.lib),
+               "-I{0}".format(self.prefix.include),
+               "{0}.c".format(exe),
+               "-o",
+               exe
+           )
            foo = which(exe)
            foo()
 
 In this case, the method copies the associated files from the build
-stage **after** the software is installed to the package's metadata
-directory. 
+stage, **after** the software is installed, to the package's test
+cache directory. Then ``test_foo`` builds ``foo`` using ``foo.c``
+before running the program.
 
 .. note::
 
-   The method name ``copy_test_sources`` here is for illustration
-   purposes. You are free to use a name that is more suited to your
-   package.
+   The method name ``copy_test_files`` here is for illustration purposes.
+   You are free to use a name that is more suited to your package.
 
-   The key to copying the files at build time for stand-alone testing
-   is use of the ``run_after`` directive, which ensures the associated
-   files are copied **after** the provided build stage.
+   The key to copying files for stand-alone testing at build time is use
+   of the ``run_after`` directive, which ensures the associated files are
+   copied **after** the provided build stage where the files **and**
+   installation prefix are available.
 
-These paths are **automatically copied** to the test stage directory
-during stand-alone testing. The package's stand-alone test method can access
-them using the ``self.test_suite.current_test_cache_dir`` property.
-In our example, the method can use the following paths to reference
+These paths are **automatically copied** from cache to the test stage
+directory prior to the execution of any stand-alone tests. Tests access
+the files using the ``self.test_suite.current_test_cache_dir`` property.
+In our example above, test methods can use the following paths to reference
 the copy of each entry listed in ``srcs``, respectively:
 
 * ``join_path(self.test_suite.current_test_cache_dir, "tests")``
@@ -5276,22 +5306,21 @@ the copy of each entry listed in ``srcs``, respectively:
 .. note::
 
     While source and input files are generally recommended, binaries
-    **may** also be cached by the build process for install testing.
-    Only you, as the package writer or maintainer, know whether these
-    would be appropriate for ensuring the installed software continues
-    to work as the underlying system evolves.
+    **may** also be cached by the build process. Only you, as the package
+    writer or maintainer, know whether these files would be appropriate
+    for testing the installed software weeks to months later.
 
 .. note::
 
-    If one or more of the copied files needs to be modified to the
-    installation, it is recommended that those changes are made
-    **once** in the ``copy_test_sources`` method after the call to
-    ``self.cache_extra_test_sources()``.  This will reduce the 
-    amount of unnecessary work in the test method **and** avoid
-    problems testing shared instances and facility deployments.
+    If one or more of the copied files needs to be modified to reference
+    the installed software, it is recommended that those changes be made
+    to the cached files **once** in the ``copy_test_sources`` method and
+    ***after** the call to ``self.cache_extra_test_sources()``. This will
+    reduce the amount of unnecessary work in the test method **and** avoid
+    problems testing in shared instances and facility deployments.
 
-    The ``filter_file`` function can be quite useful for such
-    changes. See :ref:`file manipulation <file-manipulation>`.
+    The ``filter_file`` function can be quite useful for such changes.
+    See :ref:`file manipulation <file-manipulation>`.
 
 
 .. _cache_custom_files:
@@ -5300,19 +5329,24 @@ the copy of each entry listed in ``srcs``, respectively:
 Adding custom files
 """""""""""""""""""
 
-Some tests may require additional files not available from the (source) build.
-Examples include:
+In some cases it can be useful to have files that can be used to build or
+check the results of tests.  Examples include:
 
 - test source files
 - test input files
 - test build scripts
-- expected test output
+- expected test outputs
 
-These extra files should be added to the ``test`` subdirectory of the
-package in the Spack repository.
+While obtaining such files from the software repository is preferred (see
+:ref:`adding build-time files  <cache_extra_test_sources>`), there are
+circumstances where that is not feasible (e.g., the software is not being
+actively maintained). When test files can't be obtained from the repository
+or as a supplement to files that can, Spack supports the inclusion of
+additional files under the ``test`` subdirectory of the package in the
+Spack repository.
 
-Spack will **automatically copy** the contents of that directory to the
-test staging directory prior to running stand-alone tests, which can
+Spack **automatically copies** the contents of that directory to the
+test staging directory prior to running stand-alone tests. Test methods
 access those files using the ``self.test_suite.current_test_data_dir``
 property as shown below.
 
@@ -5329,10 +5363,9 @@ property as shown below.
            data_dir = self.test_suite.current_test_data_dir
            exe = "custom-example"
            src = datadir.join("{0}.cpp".format(exe))
-
            ...
-           # TODO: Compile foo using source in src_dir
-
+           # TODO: Build custom-example using src and exe
+           ...
            custom_example = which(exe)
            custom_example()
 
@@ -5345,7 +5378,8 @@ Reading expected output from a file
 
 The helper function ``get_escaped_text_output`` is available for packages
 to retrieve and properly format the text from a file that contains the
-expected output from running an executable.
+expected output from running an executable that may contain special 
+characters.
 
 The signature for ``get_escaped_text_output`` is:
 
@@ -5356,10 +5390,13 @@ The signature for ``get_escaped_text_output`` is:
 where ``filename`` is the path to the file containing the expected output.
 
 The ``filename`` for a :ref:`custom file <cache_custom_files>` can be
-accessed and used as illustrated by a simplified version of an ``sqlite``
-package check:
+accessed by tests using the ``self.test_suite.current_test_data_dir``
+property. The example below illustrates how to read a file that was
+added to the package's ``test`` subdirectory.
 
 .. code-block:: python
+
+   import re
 
    class Sqlite(AutotoolsPackage):
        ...
@@ -5369,54 +5406,66 @@ package check:
            test_data_dir = self.test_suite.current_test_data_dir
            db_filename = test_data_dir.join("packages.db")
            ..
-
            expected = get_escaped_text_output(test_data_dir.join("dump.out"))
            sqlite3 = which(self.prefix.bin.sqlite3)
            out = sqlite3(
                db_filename, ".dump", output=str.split, error=str.split
            )
-           check_outputs(expected, out)
+           for exp in expected:
+               assert re.search(exp, out), "Expected '{0}' in output".format(exp)
 
-
-.. note::
-
-    Expected outputs do not have to be stored with the Spack package.
-    Maintaining them with the source is actually preferable.
-
-Suppose a package's source has ``examples/foo.c`` and ``examples/foo.out``
-files that are copied for stand-alone test purposes using
-:ref:`cache_extra_test_sources <cache_extra_test_sources>` and the
-stand-alone test method builds the executable ``examples/foo``. The package
-can retrieve the expected output from ``examples/foo.out`` using:
+If the file was instead copied from the ``tests`` subdirectory of the staged
+source code, the path would be obtained as shown below.
 
 .. code-block:: python
 
-   class MyFooPackage(Package):
-       ...
+       def test_example(self):
+           """check example table dump"""
+           test_cache_dir = self.test_suite.current_test_cache_dir
+           db_filename = test_cache_dir.join("packages.db")
 
-       def test_foo_example(self):
-           ..
-           filename = join_path(
-               self.test_suite.current_test_cache_dir, "examples", "foo.out"
-           )
-           expected = get_escaped_text_output(filename)
-           ..
-
-Alternatively, suppose ``MyFooPackage`` **installs** tests in ``share/tests``
-and their outputs in ``share/tests/outputs``. The expected output for
-``foo``, assuming it is still called ``foo.out``, can be retrieved as
+Alternatively, if the file was copied to the ``share/tests`` subdirectory
+as part of the installation process, the test could access the path as 
 follows:
 
 .. code-block:: python
 
-   class MyFooPackage(Package):
-       ...
+       def test_example(self):
+           """check example table dump"""
+           db_filename = join_path(self.prefix.share.tests, "packages.db")
 
-       def test_foo(self):
-           ..
-           filename = join_path(self.prefix.share.tests.outputs, "foo.out")
-           expected = get_escaped_text_output(filename)
-           ..
+
+.. _check_outputs:
+
+""""""""""""""""""""""""""""""""""""
+Comparing expected to actual outputs
+""""""""""""""""""""""""""""""""""""
+
+The helper function ``check_outputs`` is available for packages to ensure
+the expected outputs from running an executable are contained within the
+actual outputs.
+
+The signature for ``check_outputs`` is:
+
+.. code-block:: python
+
+   def check_outputs(expected, actual):
+
+where each argument has the expected type and meaning:
+
+* ``expected`` is a string or list of strings containing the expected (raw)
+  output.
+
+* ``actual`` is a string containing the actual output from executing the command
+
+Invoking the method is the equivalent of:
+
+.. code-block:: python
+
+   for check in expected:
+       if not re.search(check, actual):
+           raise RuntimeError("Expected '{0}' in output '{1}'".format(check, actual))
+
 
 .. _accessing-files:
 
@@ -5426,11 +5475,14 @@ Accessing package- and test-related files
 
 You may need to access files from one or more locations when writing
 stand-alone tests. This can happen if the software's repository does not
-include test source files or includes files but has no way to build the
-executables using the installed headers and libraries. In these cases,
-you may need to reference the files relative to one or more root
-directory. The properties containing package- (or spec-) and test-related
-directory paths are provided in the table below.
+include test source files or includes them but has no way to build the
+executables using the installed headers and libraries. In these cases
+you may need to reference the files relative to one or more root directory.
+The table below lists relevant path properties and provides additional
+examples of their use.
+:ref:`Reading expected output <expected_test_output_from_file>` provides
+examples of accessing files saved from the software repository, package
+repository, and installation.
 
 .. list-table:: Directory-to-property mapping
    :header-rows: 1
@@ -5448,7 +5500,7 @@ directory paths are provided in the table below.
      - ``self.test_suite.stage``
      - ``join_path(self.test_suite.stage, "results.txt")``
    * - Spec's Test Stage
-     - ``self.test_suite.test_dir_for_spec``
+     - ``self.test_suite.test_dir_for_spec(<spec>)``
      - ``self.test_suite.test_dir_for_spec(self.spec)``
    * - Current Spec's Build-time Files
      - ``self.test_suite.current_test_cache_dir``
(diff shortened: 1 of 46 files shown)


adamjstewart:
I like this idea a lot. Let me know if you need any opinions on design decisions. I would love `spack test` to be no different than `spack unit-test` but with a different test discovery directory. If we could use pytest under the hood, we could then add a lot of cool features automatically.

tldahlgren:
Weird python and pytest failures in unit test checks and readthedocs.

tldahlgren:
> Weird python and pytest failures in rechecks and readthedocs.

And I'm getting a LOT of failures and errors locally from my rebase to develop.  So I'll try again.

tldahlgren:
Very interesting post-commit (https://github.com/spack/spack/pull/34236/commits/eeeb696ef2f3c93ad54506a23422ab335b2c0aa9) results that at least initially seem to involve sections of code that, at first blush, were not touched by this PR.  I also saw some issues reported earlier that I thought I'd resolved a couple of days ago.

lib/spack/docs/packaging_guide.rst
```diff
@@ -4822,197 +4828,409 @@ installed executable. The check is implemented as follows:
...
+                   purpose="run installed {0}".format(example),
+                ):
+                    exe = which(join_path(self.prefix.bin, example))
+                    exe()
+
+Spack will output ``test_example: run installed examples`` before executing
+``test_example``, then ``test_example_ex1: run installed ex1`` before the first
+test part and ``test_example_ex2: run installed ex2`` before the second test
+part. The test ``test_example_ex2`` will run whether or not ``test_example_ex1``
+succeeds.
```
alalazo: (nitpick) This doesn't render well, maybe we can use some code blocks?

lib/spack/docs/packaging_guide.rst
```diff
@@ -4822,197 +4828,409 @@ installed executable. The check is implemented as follows:
...
            srcs = ["tests",
                    join_path("examples", "foo.c"),
                    join_path("examples", "bar.c")]
            self.cache_extra_test_sources(srcs)
 
+       def test_foo(self):
+           exe = "foo"
+           src_dir = join_path(
+               self.test_suite.current_test_cache_dir, "examples"
+           )
```
alalazo, line 5295: Not clear to me where `src_dir` is used in this example :thinking:

lib/spack/spack/audit.py
```diff
@@ -289,9 +289,13 @@ def _check_build_test_callbacks(pkgs, error_cls):
         pkg_cls = spack.repo.path.get_pkg_class(pkg_name)
         test_callbacks = getattr(pkg_cls, "build_time_test_callbacks", None)
 
-        if test_callbacks and "test" in test_callbacks:
-            msg = '{0} package contains "test" method in ' "build_time_test_callbacks"
-            instr = 'Remove "test" from: [{0}]'.format(", ".join(test_callbacks))
+        # TODO: change "test" and "test*" to "test_" and "test_*" once remove "test"
```
alalazo: Is this a TODO for this PR, of for a next version of Spack (e.g. after a deprecation period)?

lib/spack/spack/build_environment.py
```diff
@@ -1075,19 +1076,18 @@ def _setup_pkg_and_run(
...
-                "The '{}' package cannot find an attribute while trying to build "
-                "from sources. This might be due to a change in Spack's package format "
+                "The '{}' package cannot find an attribute while trying to {}. "
+                "This might be due to a change in Spack's package format "
                 "to support multiple build-systems for a single package. You can fix this "
-                "by updating the build recipe, and you can also report the issue as a bug. "
+                "by updating the {} recipe, and you can also report the issue as a bug. "
                 "More information at https://spack.readthedocs.io/en/latest/packaging_guide.html#installation-procedure"
-            ).format(pkg.name)
+            ).format(pkg.name, process, context)
```
alalazo, line 1090: Still need to read through the PR, but I'd be curious why this might be related to multiple build systems too. Aren't standalone tests relying solely on attributes of the package (and not of the builder)?

lib/spack/spack/install_test.py
```diff
@@ -3,32 +3,73 @@
...
 
 import spack.error
 import spack.paths
 import spack.util.spack_json as sjson
 from spack.spec import Spec
 from spack.util.prefix import Prefix
+from spack.util.string import plural
 
+#: Stand-alone test failure info type
+TEST_FAILURE_TYPE = Tuple[BaseException, str]
```
alalazo: Can we adhere to `pylint` style, i.e. use camel case for types:
```python
TestFailureType = Tuple[BaseException, str]
```
and capital letters for global variables:
```python
TEST_SUITE_FILENAME = "test.suite.lock"
```
? If we use single standard it would help reading code (since the name would convey information on the associated variable).

lib/spack/spack/install_test.py
```diff
@@ -3,32 +3,73 @@
...
+
+
+ListOrStringType = Union[str, List[str]]
+LogType = Union["tty.log.nixlog", "tty.log.winlog"]
+SpecType = List[Spec]
+
+
+@enum.unique
+class TestStatus(enum.Enum):
+    """Names of different stand-alone test states."""
```
alalazo: Do we use `unique` because we rely on the actual value somewhere?

lib/spack/spack/install_test.py
```diff
@@ -3,32 +3,73 @@
...
+#: Name of the test suite results (summary) file
 results_filename = "results.txt"
 
+#: Name of the Spack install phase-time test log file
+spack_install_test_log = "install-time-test-log.txt"
+
+
+ListOrStringType = Union[str, List[str]]
+LogType = Union["tty.log.nixlog", "tty.log.winlog"]
+SpecType = List[Spec]
```
alalazo: Hmm, given the name `SpecType` I would expect `type(Spec)` :thinking: Why not using `List[Spec]` directly?

lib/spack/spack/install_test.py
```diff
@@ -52,6 +93,638 @@ def get_test_stage_dir():
     )
 
 
+def cache_extra_test_sources(pkg: "spack.package_base.PackageBase", srcs: ListOrStringType):
+    """Copy relative source paths to the corresponding install test subdir
```
alalazo, line 99: Is it worth to check upfront that the paths are relative? Or is this is this function usually called in a context where it wouldn't add much value?

lib/spack/spack/install_test.py
```diff
@@ -52,6 +93,638 @@ def get_test_stage_dir():
...
+        expected: expected raw output string(s)
+        actual: actual output string
+
+    Raises:
+        RuntimeError: the expected output is not found in the actual output
+    """
+    expected = expected if isinstance(expected, (list, set)) else [expected]
+    for check in expected:
+        if not re.search(check, actual):
+            raise RuntimeError("Expected '{0}' in output '{1}'".format(check, actual))
```
alalazo: Would it be better to collect all the failures and report them afterwards?

lib/spack/spack/install_test.py
```diff
@@ -52,6 +93,638 @@ def get_test_stage_dir():
...
+        files = ": {0}".format(", ".join(paths)) if num_paths else ""
+        raise SkipTest(
+            "Expected {0} of {1} under {2} but {3} found{4}".format(
+                plural(expected, "copy", "copies"),
+                filename,
+                root,
+                plural(num_paths, "copy", "copies"),
+                files,
+            )
+        )
```
alalazo: Are we using this exception to signal that a test is skipped? In that case wouldn't it be better to not derive from `SpackError`? I'm thinking of code written like:
```python
try:
   # <some call including the above>
except SpackError as e:
   raise AnotherError("msg") from e

```

alalazo (commented):
A few early comments / questions after starting to get acquainted with this PR. So far I read the docs, which are imo excellent, and I started looking at `install_test.py`.

lib/spack/spack/audit.py
tldahlgren: For next iteration after a deprecation period.

lib/spack/spack/install_test.py
tldahlgren: Good point.  I don't think so but will check.

lib/spack/spack/install_test.py
tldahlgren: Yes, to show a test is skipped.  Excellent point on what exception type to derive from.  Thanks!

lib/spack/spack/install_test.py
tldahlgren: IIRC we weren't doing that before but I like that way of thinking.

lib/spack/spack/install_test.py
tldahlgren: Great catch.  

Looks like whatever I was using that for at one point isn't currently there anymore (i.e., this is the only use ATM).  I'll look into whether the drop is deliberate or if it was a conflict resolution issue.

lib/spack/spack/install_test.py
tldahlgren: This is called by users so that's an excellent point about checking the precondition.

lib/spack/docs/packaging_guide.rst
tldahlgren: Great catch!  Thanks.

lib/spack/spack/build_environment.py
tldahlgren: This has to do with the re-use of `_setup_pkg_and_run` for both the `build` and `test` processes.  

I noticed I would get `build from sources` and `updating the build recipe` when I encountered a problem during `spack test run`.  After this fix, the error reported the correct (`test`) context.

lib/spack/spack/install_test.py
tldahlgren: Sure.  I initially re-used the case I'd seen someone else use elsewhere (e.g., `package_base.py`'s use for `FLAG_HANDLER_[RETURN_]TYPE`).

lib/spack/spack/install_test.py
tldahlgren: Looks like I made a point of using the name.

lib/spack/spack/install_test.py
tldahlgren: Nope

lib/spack/spack/install_test.py
tldahlgren: Done.

lib/spack/spack/install_test.py
tldahlgren: Done.

lib/spack/spack/install_test.py
tldahlgren: Done.

lib/spack/docs/packaging_guide.rst
tldahlgren: Great catch.  Thanks. Fixed.

lib/spack/docs/packaging_guide.rst
alalazo: Looking great now!

alalazo:
I have just test-driven the PR on a simple workflow. What I tried:
1. Install `mpich`
2. Run tests on it
3. Check logs and status in general

I didn't try out complex operations, like uploading a CDash file, but the local logging etc. worked fine for when checking `mpich` tests. I was able to spot a bug in `libxml2` recipe that was nicely reported:

<details>

<summary> `libxml2` test logs </summary>

```console
$ spack test results --logs --failed
==> Failing results for test suite 'mpich':
==> test specs:
==>   libxml2-2.10.3-oqirxwn FAILED
==> Testing package libxml2-2.10.3-oqirxwn
==> [2023-04-25-20:44:14.954822] Warning: libxml2: the 'test' method is deprecated. Convert stand-alone test(s) to methods with names starting 'test_'.
==> [2023-04-25-20:44:14.955155] test: test: Perform smoke tests on the installed package
==> [2023-04-25-20:44:14.955219] test: Performing simple import test
FAILED: test: 'Libxml2' object has no attribute 'import_module_test'
  File "/home/culpo/PycharmProjects/spack/bin/spack", line 54, in <module>
    sys.exit(main())
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack_installable/main.py", line 37, in main
    sys.exit(spack.main.main(argv))
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/main.py", line 1004, in main
    return _main(argv)
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/main.py", line 959, in _main
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/main.py", line 987, in finish_parse_and_run
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/main.py", line 636, in _invoke_command
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/cmd/test.py", line 454, in test
    globals()["test_%s" % args.test_command](args)
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/cmd/test.py", line 209, in test_run
    test_suite(
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/install_test.py", line 880, in __call__
    spec.package.do_test(dirty=dirty, externals=externals)
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/package_base.py", line 1882, in do_test
    self.tester.stand_alone_tests(kwargs)
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/install_test.py", line 353, in stand_alone_tests
    spack.build_environment.start_build_process(self.pkg, test_process, kwargs)
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/build_environment.py", line 1185, in start_build_process
    p.start()
  File "/usr/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/usr/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/usr/lib/python3.8/multiprocessing/context.py", line 277, in _Popen
    return Popen(process_obj)
  File "/usr/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/usr/lib/python3.8/multiprocessing/popen_fork.py", line 75, in _launch
    code = process_obj._bootstrap(parent_sentinel=child_r)
  File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/build_environment.py", line 1051, in _setup_pkg_and_run
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/install_test.py", line 713, in test_process
    test_parts_process(pkg, test_specs, verbose)
  File "/home/culpo/PycharmProjects/spack/lib/spack/spack/install_test.py", line 670, in test_parts_process
    test_fn(pkg)
  File "/usr/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
==> [2023-04-25-20:44:14.982613] Completed testing
==> [2023-04-25-20:44:14.982701] 
======================= SUMMARY: libxml2-2.10.3-oqirxwn ========================
Libxml2::test .. FAILED
============================= 1 failed of 1 parts ==============================
==> [2023-04-25-20:44:14.982729] 

See test results at:
  /home/culpo/.spack/test/ioo6rxvkrtcwkhmbjnjkwjgujbepohha/libxml2-2.10.3-oqirxwn-test-out.txt
Test failure: TestFailure: 1 test failed.


'Libxml2' object has no attribute 'import_module_test'

/home/culpo/PycharmProjects/spack/var/spack/repos/builtin/packages/libxml2/package.py:108, in test:
        105        """Perform smoke tests on the installed package"""
        106        # Start with what we already have post-install
        107        tty.msg("test: Performing simple import test")
  >>    108        self.import_module_test()
        109
        110        data_dir = self.test_suite.current_test_data_dir
        111

```

</details>

At a high level, the only (very minor) comment is that we might want to highlight with different colors the test result summary obtained with:
```console
$ spack test results mpich
```
So that a user can spot immediately which test failed, if any. I have to say that the command help was quite useful, so I the lack of color only costed an additional command:
```console
$ spack test results --failed mpich
==> Failing results for test suite 'mpich':
==> test specs:
==>   libxml2-2.10.3-oqirxwn FAILED
================= 30 no_tests, 7 passed, 1 failed of 38 specs ==================
```
Which is :+1: @tldahlgren I'll try to dive into the code tomorrow and leave a more detailed review, but from a high-level perspective this seems fine.

lib/spack/spack/install_test.py
```diff
@@ -52,6 +91,649 @@ def get_test_stage_dir():
...
+    package_dir = spack.package_base.package_directory(pkg_cls)
+    data_source = Prefix(package_dir).test
+    data_dir = pkg.test_suite.current_test_data_dir
+    if os.path.isdir(data_source) and not os.path.exists(data_dir):
+        # We assume data dir is used read-only
+        # maybe enforce this later
+        shutil.copytree(data_source, data_dir)
+
+
+TestPackageType = Union["spack.package_base.PackageBase", Type]
```
alalazo: Reading [here](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-type-of-class-objects), I think this patch needs to be applied:
```diff
diff --git a/lib/spack/spack/install_test.py b/lib/spack/spack/install_test.py
index 42909e1cb4..a0d5e18d5f 100644
--- a/lib/spack/spack/install_test.py
+++ b/lib/spack/spack/install_test.py
@@ -13,7 +13,7 @@
 import shutil
 import sys
 from collections import Counter
-from typing import Callable, List, Optional, Tuple, Type, Union
+from typing import Callable, List, Optional, Tuple, Type, TypeVar, Union
 
 import llnl.util.filesystem as fs
 import llnl.util.tty as tty
@@ -545,7 +545,8 @@ def copy_test_files(pkg: "spack.package_base.PackageBase", test_spec: spack.spec
         shutil.copytree(data_source, data_dir)
 
 
-TestPackageType = Union["spack.package_base.PackageBase", Type]
+U = TypeVar("U", bound="spack.package_base.PackageBase")
+TestPackageType = Type[U]
 
 
 def test_function_names(pkg: TestPackageType, add_virtuals: bool = False) -> List[str]:
```

lib/spack/spack/install_test.py
```diff
@@ -3,32 +3,71 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import base64
+import contextlib
```
alalazo, line 6: I think we can change most of the strings here to f-strings. It should be pretty fast to do with `pyupgrade`:
```console
$ pip install pyupgrade
$ pyupgrade --py36-plus lib/spack/spack/install_test.py 
Rewriting lib/spack/spack/install_test.py
```

lib/spack/spack/cmd/test.py
```diff
@@ -358,18 +358,17 @@ def _report_suite_results(test_suite, args, constraints):
...
-                    untested += 1
-                elif status == "SKIPPED":
-                    skipped += 1
-
-                if args.failed and status != "FAILED":
+                # Backward-compatibility:  NO-TESTS => NO_TESTS
+                status = "NO_TESTS" if status == "NO-TESTS" else status
+
+                status = spack.install_test.TestStatus[status]
+                counts[status] += 1
```
alalazo, line 369: (very minor, not required for this PR) We can probably rework a bit the logic to just collect the statuses, and let the counter do its work. Leaving the comment, since I was not sure why we are using a `Counter` and not a `defaultdict` as we update the counts manually.

lib/spack/spack/audit.py
alalazo: For these TODOs (including others in the PR), can we add something to make them greppable in groups? I was thinking something trivial like:
```suggestion
        # TODO (stand-alone test like pytest): change "test" and "test*" to "test_" and "test_*" once remove "test"
```
or any other unique identifier for this set of changes. The rationale is that, when we want to take care of them after v0.20, they can be spotted easily in the code.

lib/spack/spack/install_test.py
```diff
@@ -52,6 +91,649 @@ def get_test_stage_dir():
...
+
+    Args:
+        builder: builder for package being tested
+        phase_name: the name of the build-time phase (e.g., ``build``, ``install``)
+        method_names: phase-specific callback method names
+    """
+    if not builder.pkg.run_tests or not method_names:
+        return
+
+    builder.pkg.tester.phase_tests(builder, phase_name, method_names)
```
alalazo: This function seems to be used only in `spack.builder._checks`. I think it might make sense to inline it. Would you agree?

lib/spack/spack/installer.py
```diff
@@ -535,6 +548,25 @@ def install_msg(name, pid):
     return pre + colorize("@*{Installing} @*g{%s}" % name)
 
 
+def archive_install_logs(pkg, phase_log_dir):
+    """
+    Copy install logs to their destination directory(ies)
+    Args:
+        pkg (spack.package_base.PackageBase): the package that was built and installed
+        phase_log_dir (str): path to the archive directory
```
alalazo, line 557: (comment, not a request of any kind) Since we support Python 3.6, I think it makes sense to start using more often `pathlib.Path` objects. Maybe it's worth discussing this briefly in one of the core dev meetings. Anyhow, this is just a general comment - not anything specific to this PR.

lib/spack/spack/package_base.py
```diff
@@ -785,7 +777,7 @@ def possible_dependencies(
     @classproperty
     def package_dir(cls):
         """Directory where the package.py file lives."""
-        return os.path.abspath(os.path.dirname(cls.module.__file__))
+        return package_directory(cls)
```
alalazo: Why was this extracted as a function? Wasn't it possible to use `pkg_cls.package_dir` in `spack.install_test`?

lib/spack/spack/package_base.py
```diff
@@ -1148,30 +1140,36 @@ def configure_args_path(self):
...
-        return os.path.join(self.metadata_dir, _spack_times_log)
+        return os.path.join(self.metadata_dir, spack_times_log)
 
     @property
     def install_configure_args_path(self):
         """Return the configure args file path on successful installation."""
         return os.path.join(self.metadata_dir, _spack_configure_argsfile)
 
+    # TODO: Update tests and all packages that use this as a package method
+    # TODO: to the routine made available in package so can remove this.
```
alalazo: Is it worth to add a warning in this wrapper, so that people are encouraged to update? In the warning we can give instructions on how to update and say that the property will not be supported anymore in v0.21 Also, we can mark the TODO with the same unique id used for the comment above.

lib/spack/spack/package_base.py
```diff
@@ -1842,6 +1840,8 @@ def do_install(self, **kwargs):
         builder = PackageInstaller([(self, kwargs)])
         builder.install()
 
+    # TODO: Update tests and all packages that use this as a package method to
+    # TODO: use the routine from package so can remove this.
```
alalazo: Same as above. With a warning we can inform users that this will be removed and remove it on `develop` when working towards v0.21

lib/spack/spack/package_base.py
```diff
@@ -1909,23 +1871,30 @@ def do_test(self, dirty=False, externals=False):
...
+            "dirty": dirty,
+            "fake": False,
+            "context": "test",
+            "externals": externals,
+            "verbose": tty.is_verbose(),
+        }
 
+        self.tester.stand_alone_tests(kwargs)
+
+    # Deprecated: Remove this once eliminate test() methods from packages
```
alalazo: Another candidate for a warning

lib/spack/spack/package_base.py
```diff
@@ -1947,6 +1916,7 @@ def run_test(
                 in the install prefix bin directory or the provided work_dir
             work_dir (str or None): path to the smoke test directory
         """
+        tty.warn("This method is deprecated.  Use 'test_part' instead.")
```
alalazo: Is:
```suggestion
        warnings.warn("This method is deprecated.  Use 'test_part' instead.")
```
behaving the same? If so it's preferred, since we can change in a unique place what to do with the warning (right now it gets forwarded by default to `tty.warn`)

lib/spack/spack/package_base.py
```diff
@@ -2007,16 +1977,17 @@ def run_test(
                 if spack.config.get("config:fail_fast", False):
                     raise TestFailure([(exc, m)])
                 else:
-                    self.test_failures.append((exc, m))
+                    self.tester.add_failure(exc, m)
                 return False
 
+    # Deprecated: remove when remove test() methods
```
alalazo: Another candidate for a `# TODO (uniqueid)`

lib/spack/spack/repo.py
```diff
@@ -1063,17 +1063,22 @@ def dump_provenance(self, spec, path):
                 "Repository %s does not contain package %s." % (self.namespace, spec.fullname)
             )
 
-        # Install patch files needed by the package.
+        # Install patch files needed by the (concrete) package.
```
alalazo, line 1072: Hmm, I'm wondering if this should operate on package classes, rather than package objects. Where was this function failing in the context of this PR?

lib/spack/spack/repo.py
```diff
@@ -1063,17 +1063,22 @@ def dump_provenance(self, spec, path):
...
+                    else:
+                        tty.warn("Patch file did not exist: %s" % patch.path)
 
         # Install the package.py file itself.
-        fs.install(self.filename_for_package_name(spec.name), path)
+        try:
+            fs.install(self.filename_for_package_name(spec.name), path)
+        except IOError as e:
+            # package-less virtuals won't have a package.py to copy
+            tty.debug("Could not copy package.py for {0}: {1}".format(spec.name, str(e)))
```
alalazo: Would it be easy to take care of special cases at the beginning, e.g. if the package is a virtual without a `package.py` return early (and avoid a try-except later)?

lib/spack/spack/reporters/cdash.py
```diff
@@ -331,6 +331,10 @@ def test_report(self, directory_name, specs):
         self.finalize_report()
 
     def test_skipped_report(self, directory_name, spec, reason=None):
+        """Explicitly report spec as being skipped (e.g., CI).
+
+        Examples the installation failed or the package is known to have
+        broken tests."""
```
alalazo: (minor) If possible, can we add a brief description of the arguments?

lib/spack/spack/reporters/extract.py
```diff
@@ -36,6 +42,8 @@ def elapsed(current, previous):
     return diff.total_seconds()
 
 
+# TODO: Should remove with deprecated run_test(), etc. since don't have an XFAIL
+# TODO: mechanism with the new test_part() approach.
```
alalazo: Another TODO to mark

lib/spack/spack/reporters/extract.py
```diff
@@ -58,8 +66,8 @@ def new_part():
     }
 
 
+# TODO: Remove this when remove deprecated test_part(), etc.
```
alalazo: Another TODO to mark

lib/spack/spack/reporters/extract.py
```diff
@@ -73,6 +81,8 @@ def process_part_end(part, curr_time, last_time):
 
         stat = part["status"]
         if stat in completed:
+            # TODO: remove the expected failure mapping along with removal of
+            # TODO: deprecated run_test(), etc.
```
alalazo: Another TODO to mark

lib/spack/spack/reporters/extract.py
```diff
@@ -96,16 +106,20 @@ def status(line):
...
 
 
 def extract_test_parts(default_name, outputs):
     parts = []
     part = {}
-    testdesc = ""
     last_time = None
     curr_time = None
+
+    # TODO: Remove these once remove deprecated run_test() processing
```
alalazo: Another TODO to mark

lib/spack/spack/reporters/extract.py
```diff
@@ -10,15 +10,21 @@
...
 
 log_regexp = re.compile(r"^==> \[([0-9:.\-]*)(?:, [0-9]*)?\] (.*)")
 returns_regexp = re.compile(r"\[([0-9 ,]*)\]")
 
 skip_msgs = ["Testing package", "Results for", "Detected the following"]
 skip_regexps = [re.compile(r"{0}".format(msg)) for msg in skip_msgs]
 
-status_values = ["FAILED", "PASSED", "NO-TESTS"]
+# Status values need to match those in spack.install_test.TestStatus
+status_values = ["FAILED", "PASSED", "SKIPPED", "NO_TESTS"]
```
alalazo: Can we import `spack.install_test.TestStatus` and below use:
```python
status_regexps = [re.compile(r"^({0})".format(str(stat))) for stat in TestStatus]
```
?

lib/spack/spack/test/cmd/env.py
```diff
@@ -705,10 +705,7 @@ def test_with_config_bad_include():
         with e:
             e.concretize()
 
-    err = str(exc)
-    assert "missing include" in err
-    assert "/no/such/directory" in err
-    assert os.path.join("no", "such", "file.yaml") in err
+    assert "2 missing include path" in str(exc)
```
alalazo: Can we use a `match=` to `pytest.raises`?

lib/spack/spack/test/cmd/install.py
```diff
@@ -1177,3 +1177,9 @@ def test_report_filename_for_cdash(install_mockery_mutable_config, mock_fetch):
     specs = spack.cmd.install.concrete_specs_from_cli(args, {})
     filename = spack.cmd.install.report_filename(args, specs)
     assert filename != "https://blahblah/submit.php?project=debugging"
+
+
+@pytest.mark.disable_clean_stage_check
```
alalazo: (not a request, just a question) Do you know why this fixture is necessary? Why don't we leave the test with a clean stage?

lib/spack/spack/test/cmd/test.py
```diff
@@ -62,6 +63,7 @@ def test_test_dup_alias(
     # Run the tests with the alias once
     out = spack_test("run", "--alias", "libdwarf", "libdwarf")
     assert "Spack test libdwarf" in out
+    assert "1 failed" not in out
```
alalazo: (not necessarily for this PR) I wonder if we have a more structural way to check the result of a command operation here, that is not checking the screen output.

lib/spack/spack/test/installer.py
```diff
@@ -1331,3 +1331,29 @@ def test_single_external_implicit_install(install_mockery, explicit_args, is_exp
     s.external_path = "/usr"
     create_installer([(s, explicit_args)]).install()
     assert spack.store.db.get_record(pkg).explicit == is_explicit
+
+
+def test_print_install_test_log(tmpdir, install_mockery, mock_packages, ensure_debug, capfd):
```
alalazo: Two questions:
1. Can you add a brief docstring saying what we are testing here?
2. Is it possible to parametrize this test or split it into 4 smaller tests? Or are the assertions dependent on the previous output?

lib/spack/spack/test/package_class.py
```diff
@@ -192,3 +193,135 @@ def test_cache_extra_sources(install_mockery, spec, sources, extras, expect):
 
     # Perform a little cleanup
     shutil.rmtree(os.path.dirname(source_path))
+
+
+def test_cache_extra_sources_fails(install_mockery):
```
alalazo, line 201: Same request here. For new tests, can we add a brief docstring to explain the purpose of the test?[^1]

[^1]: I'm asking that since I did some refactoring lately, and sometimes trying to guess what a very old test was meant to verify is hard. So I started taking the habit of having a one or two lines docstring explaining the purpose of new tests I added.

lib/spack/spack/test/package_class.py
```diff
@@ -192,3 +193,135 @@ def test_cache_extra_sources(install_mockery, spec, sources, extras, expect):
...
+    s = spack.spec.Spec("a")
+    pkg = BaseTestPackage(s)
+    pkg.test_requires_compiler = True
+    pkg.do_test()
+    error = capfd.readouterr()[1]
+    assert "Skipping tests for package" in error
+    assert "test requires missing compiler" in error
+
+
+# TODO: Remove when remove deprecated run_test(), etc.
```
alalazo: Another TODO to mark

lib/spack/spack/test/repo.py
```diff
@@ -150,3 +150,17 @@ def test_repo_path_handles_package_removal(tmpdir, mock_packages):
...
+    vspec = spack.spec.Spec("something")
+    mutable_mock_repo.dump_provenance(vspec, tmpdir)
+    captured = capsys.readouterr()[1]
+    assert "Could not copy package" in captured
+
+    # Now with a virtual with a package
+    vspec = spack.spec.Spec("externalvirtual")
+    mutable_mock_repo.dump_provenance(vspec, tmpdir)
+    captured = capsys.readouterr()[1]
+    assert "package.py" in os.listdir(tmpdir), "Expected the virtual's package to be copied"
```
alalazo: Can we add a docstring and maybe split / parametrize the test?

lib/spack/spack/test/test_suite.py
```diff
@@ -2,13 +2,17 @@
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
+import collections
```
alalazo, line 5: Same general remark about docstrings in tests and splitting tests where possible

lib/spack/spack/test/test_suite.py
```diff
@@ -182,3 +189,196 @@ def add_suite(package):
     with pytest.raises(spack.install_test.TestSuiteNameError) as exc_info:
         spack.install_test.get_test_suite(name)
     assert "many suites named" in str(exc_info)
+
+
+@pytest.mark.parametrize(
+    "virtuals,expected",
+    [(False, ["Mpich.test_mpich"]), (True, ["Mpi.test_hello", "Mpich.test_mpich"])],
+)
```
alalazo, line 196: :+1:

alalazo (changes requested):
I went through the details of the code. The PR substantially LGTM. 

I left a lot of minor comments, but that is more due to the size of the PR than anything else. In the most part they should be trivially fixable, and there's nothing touching the design of the testing mechanism.

There are a few recurring comments in the review:
1. Mark all the TODOs in this PR with a unique identifier, so they can be easily distinguished from other TODOs later
2. Add warnings with instructions on how to upgrade from using deprecated functions. State in which Spack version the deprecated functions will be removed (Spack v0.21?)
3. Add very brief docstrings in newly added tests, to state the purpose of the test. 
4. Try splitting / parametrizing tests that are repeating the same operation n times (with different data) into n tests

Of course feel free to ping me if you want to discuss anything :slightly_smiling_face:

lib/spack/spack/install_test.py
tldahlgren: Hmm.  The function can take either an instance or the class so wouldn't `TestPackageType = Union[U, Type[U]]` be more accurate?

lib/spack/spack/install_test.py
tldahlgren: Very cool!  Noticed it changed one of the exceptions so had to manually fix it in the docstring and test.

lib/spack/spack/cmd/test.py
tldahlgren: I am using `Counter` to 1) track occurrences of the test status values *and* 2) for the `total()` (see the `parts()` and `write_test_summary()` methods).

lib/spack/spack/install_test.py
tldahlgren: Done.

lib/spack/spack/audit.py
tldahlgren: How about "TODO (post-34236)"?  I made that change and noticed another place where I forgot one.

alalazo: Sure, anything uniquely identifiable is fine with me.

lib/spack/spack/install_test.py
alalazo: Correct. This is done in 34f33972c520ed34d25deb462080619bbd9a4560 so I'm "resolving" this conversation

lib/spack/spack/cmd/test.py
alalazo: Sure, the point I'm making is that:
1. For tracking occurences, we're doing it manually (so there's no benefit in using `Counter` which does it in `__init__`)
2. For `Counter.total()` we have the implementation that extract the same value from a generic dict, since the method has been added in Python 3.10

But, as I said, this is very minor, so I'm going to collapse this item.

lib/spack/spack/package_base.py
tldahlgren: IDR

lib/spack/spack/package_base.py
tldahlgren: That method is being inspected to check for having an implementation (or being "empty").

lib/spack/spack/repo.py
tldahlgren: When the `TestSuite` is called, the first method is `write_reproducibility_data`, which iterates over the associated specs.  IIRC The specs included a package-less virtual though I don't recall which one off the top of my head (and it isn't in my notes unfortunately).

lib/spack/spack/package_base.py
tldahlgren: Added warning.  

The test outputs from existing tests can be really filled with warnings now.  Not only do you get the one about `test` being present, each `run_test` gets its own warning.  For some packages (e.g., `OpenMPI`) this looks "messy".  Needed to change CDash test extraction (and package base) to better handle these deprecated tests.

lib/spack/spack/test/cmd/env.py
tldahlgren: Yes, now that there's only the one check.

lib/spack/spack/test/cmd/install.py
tldahlgren: Test package lists a method that does not exist so get `TestFailure: 1 test failed.` due to `object has no attribute 'undefined-install-test'` so test leaving the test log in place (hence the assert `See test log for details`).

lib/spack/spack/package_base.py
tldahlgren: Appears to be, though not for some of the unit testing.  Adjusted tests accordingly.

lib/spack/spack/test/cmd/test.py
tldahlgren: Have to think about it.

lib/spack/spack/test/cmd/install.py
tldahlgren: Removed the unit test since I can piggy back the check off of an existing one.

tldahlgren:
@adamjstewart Running `py-six` using any of the `verbose` option placements with this PR work for me.  See below:

```
$ spack -v install --test=root py-six
...
==> Testing package py-six-1.16.0-rsnbm5d
==> [2023-05-08-17:33:42.074994] Running install-time tests
==> [2023-05-08-17:33:42.090933] RUN-TESTS: install-time tests [test]
==> [2023-05-08-17:33:42.098625] Detected the following modules: ['six']
==> [2023-05-08-17:33:42.098936] Warning: The 'run_test' method is deprecated. Use 'test_part' instead for py-six to process python3.10.
==> [2023-05-08-17:33:42.099459] test: test_python3.10_c_importsix: checking import of six
==> [2023-05-08-17:33:42.100213] Expecting return code in [0]
==> [2023-05-08-17:33:42.100656] '/usr/WS1/dahlgren/spack/opt/spack/linux-rhel8-broadwell/gcc-10.3.1/python-3.10.10-uypaufv2bj3q56bfc2qpr3pfujo6odba/bin/python3.10' '-c' 'import six'
PASSED: PySix::test_python3.10_c_importsix
==> [2023-05-08-17:33:42.175526] Completed testing
==> py-six: Successfully installed py-six-1.16.0-rsnbm5dermzruuggmkxlakmxws37gvwg
  Stage: 0.05s.  Install: 3.28s.  Post-install: 0.66s.  Total: 5.47s
  
$ spack uninstall -y /rsnbm5d    # to install from source again
$ spack install -v --test=root py-six
...
==> Testing package py-six-1.16.0-rsnbm5d
==> [2023-05-08-17:43:30.622153] Running install-time tests
==> [2023-05-08-17:43:30.633689] RUN-TESTS: install-time tests [test]
==> [2023-05-08-17:43:30.641657] Detected the following modules: ['six']
==> [2023-05-08-17:43:30.641934] Warning: The 'run_test' method is deprecated. Use 'test_part' instead for py-six to process python3.10.
==> [2023-05-08-17:43:30.642522] test: test_python3.10_c_importsix: checking import of six
==> [2023-05-08-17:43:30.643282] Expecting return code in [0]
==> [2023-05-08-17:43:30.643653] '/usr/WS1/dahlgren/spack/opt/spack/linux-rhel8-broadwell/gcc-10.3.1/python-3.10.10-uypaufv2bj3q56bfc2qpr3pfujo6odba/bin/python3.10' '-c' 'import six'
PASSED: PySix::test_python3.10_c_importsix
==> [2023-05-08-17:43:30.717449] Completed testing
==> py-six: Successfully installed py-six-1.16.0-rsnbm5dermzruuggmkxlakmxws37gvwg
  Stage: 0.02s.  Install: 3.96s.  Post-install: 1.08s.  Total: 9.70s
[+] /usr/WS1/dahlgren/spack/opt/spack/linux-rhel8-broadwell/gcc-10.3.1/py-six-1.16.0-rsnbm5dermzruuggmkxlakmxws37gvwg
22.211u 2.935s 0:46.15 54.4%	0+0k 712+2200io 0pf+0w

$ spack uninstall -y /rsnbm5d    # to install from source again
$ spack -v install -v --test=root py-six
...
==> Testing package py-six-1.16.0-rsnbm5d
==> [2023-05-08-17:48:07.520216] Running install-time tests
==> [2023-05-08-17:48:07.531475] RUN-TESTS: install-time tests [test]
==> [2023-05-08-17:48:07.539563] Detected the following modules: ['six']
==> [2023-05-08-17:48:07.539937] Warning: The 'run_test' method is deprecated. Use 'test_part' instead for py-six to process python3.10.
==> [2023-05-08-17:48:07.540501] test: test_python3.10_c_importsix: checking import of six
==> [2023-05-08-17:48:07.541284] Expecting return code in [0]
==> [2023-05-08-17:48:07.541673] '/usr/WS1/dahlgren/spack/opt/spack/linux-rhel8-broadwell/gcc-10.3.1/python-3.10.10-uypaufv2bj3q56bfc2qpr3pfujo6odba/bin/python3.10' '-c' 'import six'
PASSED: PySix::test_python3.10_c_importsix
==> [2023-05-08-17:48:07.610124] Completed testing
==> py-six: Successfully installed py-six-1.16.0-rsnbm5dermzruuggmkxlakmxws37gvwg
  Stage: 0.02s.  Install: 3.36s.  Post-install: 0.75s.  Total: 5.87s
[+] /usr/WS1/dahlgren/spack/opt/spack/linux-rhel8-broadwell/gcc-10.3.1/py-six-1.16.0-rsnbm5dermzruuggmkxlakmxws37gvwg
21.214u 2.319s 0:36.65 64.1%	0+0k 712+2200io 0pf+0w
```

lib/spack/spack/install_test.py
```diff
@@ -3,34 +3,76 @@
...
 
+#: Name of the Spack install phase-time test log file
+spack_install_test_log = "install-time-test-log.txt"
+
+
+ListOrStringType = Union[str, List[str]]
+LogType = Union["tty.log.nixlog", "tty.log.winlog"]
+
+Pb = TypeVar("Pb", bound="spack.package_base.PackageBase")
+TestPackageType = Union[Pb, Type[Pb]]
```
alalazo: (Refactor) I would call this:
```suggestion
PackageObjectOrPackageClass = Union[Pb, Type[Pb]]
```

lib/spack/spack/install_test.py
```diff
@@ -52,6 +94,640 @@ def get_test_stage_dir():
...
+            the staged source path that are to be copied to the corresponding
+            location(s) under the install testing directory.
+
+    Raises:
+        OSError: if any of the source paths do not exist under the build stage
+        ValueError: if any of the source paths are absolute
+    """
+    paths = [srcs] if isinstance(srcs, str) else srcs
+    if any(os.path.isabs(path) for path in paths):
+        raise ValueError("Only relative paths can be used to copy test sources")
```
alalazo: (Refactor) I'm a bit worried about raising built-in exception like `OSError` We might turn this into a `SpackError` of some kind in the future, so that Spack owns its own errors.

lib/spack/spack/install_test.py
```diff
@@ -52,6 +94,640 @@ def get_test_stage_dir():
...
+    for path in paths:
+        src_path = os.path.join(pkg.stage.source_path, path)
+        dest_path = os.path.join(install_test_root(pkg), path)
+        if os.path.isdir(src_path):
+            fs.install_tree(src_path, dest_path)
+        elif os.path.exists(src_path):
+            fs.mkdirp(os.path.dirname(dest_path))
+            fs.copy(src_path, dest_path)
+        else:
+            raise OSError(f"Source path ('{path}') for the copy does not exist")
```
alalazo: (Refactor) This loop is a good candidate for collecting errors in advance and reporting all of them at the end (if any).

lib/spack/spack/package_base.py
alalazo: > For some packages (e.g., OpenMPI) this looks "messy". Needed to change CDash test extraction (and package base) to better handle these deprecated tests.

(Question) Do you think we have time to update those packages in `builtin`, after this PR is merged?

lib/spack/spack/install_test.py
```diff
@@ -323,7 +323,8 @@ def phase_tests(
 
             for name in method_names:
                 try:
-                    fn = getattr(builder, name)
+                    # Prefer the method in the package over the builder's
+                    fn = getattr(builder.pkg, name, getattr(builder, name))
```
alalazo: (Refactor) This might be worth extracting into a small function, and explaining why we prefer the method in the package over the builder's in a docstring.

lib/spack/spack/install_test.py
```diff
@@ -294,7 +294,20 @@ def status(self, name: Optional[str], status: "TestStatus"):
...
+        # The containing test part can be PASSED while sub-parts (assumed
+        # to start with the same name) may not have PASSED. This extra
+        # check is used to ensure the containing test part is not claiming
+        # to have passed when at least one subpart failed.
+        if status == TestStatus.PASSED:
+            for name, substatus in self.test_parts.items():
+                if name != part_name and name.startswith(part_name):
+                    if substatus == TestStatus.FAILED:
+                        status = substatus
+                        break
```
alalazo: (Question) Not sure I understand what is happening here... In which case a sub-part fails but the enclosing part passes?

lib/spack/spack/package_base.py
```diff
@@ -1922,7 +1924,31 @@ def run_test(
...
+            if match:
+                # The test title has all the expected parts
+                return purpose
+
+            match = re.search(r"test: (.*)", purpose)
+            if match:
+                reason = match.group(1)
+                return f"test: {test_name}: {reason}"
+
+            return f"test: {test_name}: {purpose}"
```
alalazo: (Just a comment) The `run_test` function is really "complex". Looking forward to removing that, per TODO above :sweat_smile:

lib/spack/spack/reporters/extract.py
```diff
@@ -129,6 +131,18 @@ def extract_test_parts(default_name, outputs):
...
+        match = re.search(r"Test failure: (.*)", line)
+        if match:
+            stat = "failed"
+            part = new_part()
+            part["command"] = "none"
+            part["completed"] = completed[stat]
+            part["elapsed"] = 0.0
+            part["name"] = default_name
+            part["loglines"].append(match.group(1))
+            part["status"] = stat
```
alalazo: (Refactor) When we remove the deprecated methods, we might as well give some more structure to the code, for instance check if a "test part" is a concept that is worth to have its own class.

lib/spack/spack/test/ci.py
```diff
@@ -594,9 +594,10 @@ def test_ci_skipped_report(tmpdir, mock_packages, config):
     reason = "Testing skip"
     handler.report_skipped(spec, tmpdir.strpath, reason=reason)
 
-    report = fs.join_path(tmpdir, "{0}_Testing.xml".format(pkg))
-    expected = "Skipped {0} package".format(pkg)
-    with open(report, "r") as f:
+    reports = [name for name in tmpdir.listdir() if str(name).endswith("Testing.xml")]
+    assert len(reports) == 1
+    expected = f"Skipped {pkg} package"
+    with open(reports[0], "r") as f:
```
alalazo, line 600: (Question) Not sure I get why this had changed since the last review... :thinking: Can you elaborate a bit more  on this test?

lib/spack/spack/reporters/extract.py
```diff
@@ -178,9 +185,13 @@ def extract_test_parts(default_name, outputs):
                 # (e.g., output=str.split, error=str.split) will not have
                 # a command printed to the test log.
                 elif msg.startswith("'") and msg.endswith("'"):
-                    if part["command"]:
-                        part["command"] += "; " + msg.replace("'", "")
+                    if part:
+                        if part["command"]:
+                            part["command"] += "; " + msg.replace("'", "")
+                        else:
+                            part["command"] = msg.replace("'", "")
```
alalazo, line 193: (Comment + Refactor) We should probably give more structure to this module in a later PR, after v0.20. Here we have 3 nested conditionals, which are performing low level operations[^1], delegating to the comment above them an explanation of what is done at a high-level and why. 

I'd prefer to have, in general, classes and methods with names that evoke what is being performed at a high-level. I think we can do that when we'll remove the deprecated methods (so clean-up + refactor).

[^1]: Is a variable false-ish? Is a string starting and ending with `'`?

lib/spack/spack/test/reporters.py
```diff
@@ -136,9 +136,9 @@ def test_reporters_extract_empty_and_xfail():
 
     parts = spack.reporters.extract.extract_test_parts("fake", outputs)
 
-    assert len(parts) == 2
+    assert len(parts) == 1
```
alalazo, line 139: (Question) Why was this test expecting 2 parts before and only 1 now? Apologies, but that is not immediately clear to me.

lib/spack/spack/package_base.py
```diff
@@ -1897,7 +1897,8 @@ def _test_deprecated_warning(self):
     # TODO (post-34236): run_test, etc.
     def test(self):
         # Defer tests to virtual and concrete packages
-        warnings.warn(self._test_deprecated_warning)
+        # Ensure passing warn() a string for python 3.6/rhel8,ubuntu,etc.
+        warnings.warn(str(self._test_deprecated_warning))
```
alalazo: (Bug) Is this a typo? I don't understand why we want to cast to `str`. Isn't the failure due to the fact that either:
1. `_test_deprecated_warning` was not declared as a property
2. `_test_deprecated_warning` is not getting called

?

lib/spack/spack/test/builder.py
```diff
@@ -151,6 +151,9 @@ def test_monkey_patching_test_log_file():
     assert builder.pkg_with_dispatcher.tester.test_log_file == "/some/file"
 
 
+# Windows context manager's __exit__ fails with ValueError ("I/O operation
+# on closed file").
+@pytest.mark.skipif(sys.platform == "win32", reason="Does not run on windows")
```
alalazo, line 156: (Question) Do we have any idea why this is happening?

lib/spack/spack/install_test.py
```diff
@@ -52,6 +94,640 @@ def get_test_stage_dir():
...
+            lines = [ln.strip() for ln in source if not skip(ln)]
+            if not lines:
+                continue
+
+            tests.append((clss.__name__, test_fn))  # type: ignore[union-attr]
+
+    return tests
+
+
+def test_parts_process(pkg: Pb, test_specs: List[spack.spec.Spec], verbose: bool = False):
```
alalazo: (Refactor) If we could avoid having `test_*` functions in code outside of unit tests it would be great :sweat_smile: 
```suggestion
def process_test_parts(pkg: Pb, test_specs: List[spack.spec.Spec], verbose: bool = False):
```

lib/spack/spack/package_base.py
```diff
@@ -1909,23 +1877,39 @@ def do_test(self, dirty=False, externals=False):
...
+            "context": "test",
+            "externals": externals,
+            "verbose": tty.is_verbose(),
+        }
+
+        self.tester.stand_alone_tests(kwargs)
 
+    # TODO (post-34236): Remove this deprecated method when eliminate test,
+    # TODO (post-34236): run_test, etc.
+    def _test_deprecated_warning(self):
```
alalazo, line 1893: (Bug) This is what I meant a couple of comments above:
```suggestion
    @property
    def _test_deprecated_warning(self):
```

alalazo (changes requested):
**TL;DR** The only required change is the one marked as `(Bug)`. I also have a few questions for aspects of the PR I'd like to understand before merging. Anything marked as `(Refactor)` you can skip over. I left the comment there for future reference, but I don't think it's essential to merge this PR.

---

I see that there is some new code. In general I confirm what I said before:
1. Documentation is excellent
2. Use from the command line works as expected, at least on small tests that could be done on a laptop[^1]

[^1]: For the record, I didn't try any CDash upload with tests reports or similar, more complex, operations.

On the code, I think the structure and naming can be improved but, to be fair, this isn't an issue introduced by this PR. Some thoughts:
1. We can take advantage of the future removal of the deprecated features to refactor code in `spack.reporters.extract` and related modules
2. We can, in general, remove a few comments in favor of methods or functions with a name that conveys the message of the comments
3. Unit tests for "spack test" related features are sometimes brittle, in the sense that they assert if some literal strings are in the output. When the literal string changes it's hard for me to understand if that was accidental or if there was a functional change in some part of the code. That might also be improved in a future refactor.

lib/spack/spack/install_test.py
tldahlgren: How about `PackageObjectOrClass`?

tldahlgren: It's a little tighter and IMO conveys the same types.

lib/spack/spack/install_test.py
tldahlgren: IIRC I used `OSError` because that is what would have been returned by the `fs` call if the path didn't exist.  I was just trying to avoid the extra calls when I could check in advance.

lib/spack/spack/install_test.py
tldahlgren: Changed to the name I mentioned.

lib/spack/spack/install_test.py
tldahlgren: Changed to capture and report the errors for the method as `InstallError` since that's when the sources are cached.

lib/spack/spack/install_test.py
tldahlgren: All ATM without this.  

This is more a case needing refactoring as I haven't yet spent time investigating how best to handle it.  I know we have precedence with at least the (prefix) locks and config scope.

lib/spack/spack/reporters/extract.py
tldahlgren: Excellent point!

lib/spack/spack/test/ci.py
tldahlgren: Without the change, someone running `spack test run --log-format=cdash` for package that has multiple installations would end up with one (the latest) `cdash` report file (`<package>_Testing.xml`).  

I changed `report_test_data` (in `cdash.py`) to reformat the name to `<package>_<package-id/hash>_Testing.xml`) so I would get an xml file for each package. Would've preferred `<package>_<version>_<short-hash>_Testing.xml` but that information isn't currently readily/generally available.

Bottom line: Changed as a result of re-running testing at the command line and being irritated that I hadn't addressed the report file naming issue yet.

lib/spack/spack/reporters/extract.py
tldahlgren: I wholeheartedly agree.  Will be easier once the deprecated code is removed AND have more use cases.  (Given the milestone, I haven't revisited the outputs for the other PRs that are waiting for this one to be merged.)

lib/spack/spack/test/reporters.py
tldahlgren: Detected a bug.  Thanks for calling my attention to it as I found a missing 'continue' as well.

lib/spack/spack/package_base.py
tldahlgren: 🤦

lib/spack/spack/install_test.py
tldahlgren: Fair enough.  I debated that myself though it is in `install_test` not `package_base`.

Changed.

lib/spack/spack/package_base.py
tldahlgren: Yes, I got that from your description.

lib/spack/spack/test/builder.py
tldahlgren: No.  I don't have a Windows box to test and wasn't getting more from the PR check than what I added to the comment.

alalazo (approved):
I think we are ready to merge this one. We can keep a close eye on issues that may arise in the forthcoming days, in case we need some minor fixup to follow. For the record I just tested `openmpi` with both the current setup and with #35802 and it works correctly.

tldahlgren:
> I think we are ready to merge this one. We can keep a close eye on issues that may arise in the forthcoming days, in case we need some minor fixup to follow. For the record I just tested `openmpi` with both the current setup and with #35802 and it works correctly.

Great!  I'm inclined to start working on the follow-on test PRs (i.e., confirming they work and making adjustments as needed).

PR #34237, opened 2022-12-01
kealib: update to 1.5.0

gillins:
Note: this version is not ABI compatible with 1.4.x so users will have to rebuild anything linked against it. It is source compatible.


neilflood:
@gillins 
Yup, that looks fine to me. I double-checked the sha256, too, all good.

tldahlgren (approved):
Confirmed the sha256 for the new version.

PR #34238, opened 2022-12-01
[quantum-espresso] Parallel make fails for 6.{6,7}

stephenmsachs:
I run into a race condition in `make` with Intel compiler on icelake when building QE 6.6 and 6.7.


danielecesarini (dismissed):

bellenlau (dismissed):

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -582,7 +582,7 @@ def install(self, pkg, spec, prefix):
                 filter_file(zlib_libs, format(spec["zlib"].libs.ld_flags), make_inc)
 
         # QE 6.6 and later has parallel builds fixed
-        if spec.satisfies("@:6.5"):
+        if spec.satisfies("@:6.7"):
```
ye-luo, line 585: Please update the note above.

stephenmsachs: Thanks for catching that.

tldahlgren (approved):
LGTM

PR #34240, opened 2022-12-01
vecgeom: add new 1.2.1 version

drbenmorgan:
Adds the just released patch version 1.2.1, which will also be the required minimum for the upcoming Geant4 minor release.

@sethrj, as we noted elsewhere, this, and earlier versions (at least 1.1.18-1.1.20) do not build with GCC 11.3 and CUDA >=11.6 (GCC 11.**2** and compatible CUDAs works though!). Should we use a `conflict` to express this?

### DIFF
--- a/var/spack/repos/builtin/packages/vecgeom/package.py
+++ b/var/spack/repos/builtin/packages/vecgeom/package.py
@@ -60,7 +60,11 @@ class Vecgeom(CMakePackage, CudaPackage):
     depends_on("veccore@0.4.2", when="@:1.0")
 
     conflicts("+cuda", when="@:1.1.5")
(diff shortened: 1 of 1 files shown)


sethrj:
Yes @drbenmorgan please add a conflict to document that 😊

sethrj:
@spackbot help

sethrj:
@spackbot fix style

drbenmorgan:
Oh, that's new to me - very nice!

sethrj (approved):

sethrj:
One thing to put here for the record is that the `__builtin_ia32_serialize` failure is due to VecCore I think, but since VecCore is header-only we have to express in VecGeom that the failure happens when CUDA is enabled.

tldahlgren (approved):
Confirmed sha256, conflict LGTM too

PR #34242, opened 2022-12-01
meep: add new versions and additional variants incl. dependencies

muffgaga:


var/spack/repos/builtin/packages/meep/package.py
```diff
@@ -106,6 +112,11 @@ def configure_args(self):
         if "+single" in spec:
             config_args.append("--enable-single")
 
+        if "+openmp" in spec:
+            config_args.append("--with-openmp")
+        else:
+            config_args.append("--without-openmp")
+
```
tldahlgren: You can simplify this with one of the AutotoolsPackage helpers (see https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#with-or-without).

tldahlgren (commented):
Confirmed all four sha256.

var/spack/repos/builtin/packages/meep/package.py
muffgaga: Fixed && I also added a commit on top which simplifies two other trivial with/without options.

tldahlgren (approved):
LGTM

PR #34249, opened 2022-12-01
doxygen: add build-tools tag

sethrj:
This allows it to be included automatically as an external. No one links against doxygen so this should be ok.

### DIFF
--- a/var/spack/repos/builtin/packages/doxygen/package.py
+++ b/var/spack/repos/builtin/packages/doxygen/package.py
@@ -39,6 +39,8 @@ class Doxygen(CMakePackage):
 
     variant("mscgen", default=False, description="Build with support for code graphs from mscgen.")
 
+    tags = ["build-tools"]
+
     executables = ["doxygen"]
 
     @classmethod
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/doxygen/package.py
```diff
@@ -39,8 +39,12 @@ class Doxygen(CMakePackage):
 
     variant("mscgen", default=False, description="Build with support for code graphs from mscgen.")
 
+    tags = ["build-tools"]
```
tldahlgren, line 42: Nit: Is it a build or documentation tool?

I can understand why someone would want to flag the package as a build tool for query purposes so I'll approve as is.

tldahlgren (approved):

var/spack/repos/builtin/packages/doxygen/package.py
sethrj: There is no `doc-tools` tag, and usually documentation tools *are* build tools since they're used as executables to generate documentation at build time.

tldahlgren: I noticed there are no `doc`-related tags.  You could start one if you want 😉 .

Fair point about how it fits in workflows.

sethrj:
Thanks for the review @tldahlgren !

PR #34255, opened 2022-12-02
Add toggle for autoloading upstream modules

vanderwb:
Currently, if you tell Spack to autoload modules, `depends_on` statements will be added for both local instance and upstream instance dependencies. This behavior could be fine/desired if you add both module trees to the MODULEPATH, but some users may want to load dependencies built in the local instance but ignore ones built in the upstream (since they may not exist!).

This PR adds a *modules.yaml* toggle called `exclude_upstream_autoloads`, similar to `exclude_implicits`, that if **true** allows the user to ignore upstream packages when generating `depends_on` statements.

### DIFF
--- a/lib/spack/spack/modules/common.py
+++ b/lib/spack/spack/modules/common.py
@@ -519,6 +519,14 @@ class BaseConfiguration(object):
         installed_implicitly = not spec._installed_explicitly()
         excluded_as_implicit = exclude_implicits and installed_implicitly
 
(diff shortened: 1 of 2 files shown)


vanderwb:
There appears to be an import that is unused, causing a failing style test. I did not introduce this line of code but could clean it up or comment it out in this PR if the concept is judged sound.

vanderwb:
Will revisit if needed.

PR #34261, opened 2022-12-02
[texinfo] @7.0: needs c-11 syntax

stephenmsachs:
gnulib/lib/malloca.c uses single value `static_assert()` only available in c-11 syntax. `gcc` seems to be fine, but `icc` needs extra flag.


alalazo (approved):

alalazo:
@spackbot run pipeline

PR #34265, opened 2022-12-02
LLVM: replace libelf dependency with elf

iarspider:
I didn't test this extensively, but in CMS LLVM builds just fine with elfutils.

### DIFF
--- a/var/spack/repos/builtin/packages/llvm/package.py
+++ b/var/spack/repos/builtin/packages/llvm/package.py
@@ -598,9 +598,7 @@ class Llvm(CMakePackage, CudaPackage):
                     [
                         define("LIBOMPTARGET_NVPTX_ENABLE_BCLIB", True),
                         # work around bad libelf detection in libomptarget
-                        define(
-                            "LIBOMPTARGET_DEP_LIBELF_INCLUDE_DIR", spec["elf"].prefix.include
-                        ),
+                        define("LIBOMPTARGET_DEP_LIBELF_INCLUDE_DIR", spec["elf"].prefix.include),
                     ]
                 )
         else:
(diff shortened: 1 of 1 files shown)


iarspider:
@spackbot fix style

iarspider:
@spackbot run pipeline

bernhardkaindl:
@iarspider It should work of coruse because the libelf from elfutils is compatible with libelf, but what should be the benefit?

I only see the drawback of adding more dependencies (the dependencies of elfutils) into the build for (I think) no gain.

For your PRs in general but for this one specially: Can you please add the motivation why merging your PR(s) would be beneficial and worth the incurred drawbacks? Thanks!

bernhardkaindl (commented):
@iarspider It should work of course because the libelf from elfutils is compatible with libelf, but what should be the benefit?

I only see the drawback of adding more dependencies (the dependencies of elfutils) into the build for (I think) no gain.

For your PRs in general but for this one specially: Can you please add the motivation why merging your PR(s) would be beneficial and worth the incurred drawbacks? Thanks!

iarspider:
@bernhardkaindl `libelf` and `elfutils` both provide virtual package `elf`. Also, `libelf` is not maintained anymore, and it's recommended to use `elfutils` instead. This PR replaces hard dependency on `libelf` with a virtual dependency on `elf`. I think, this is kind of recommended way of writing package recipes: unless package X can only be used (built or ran) with specific provider of virtual package Y, put dependency on Y.

iarspider:
One additional benefit of using `elf` vs `libelf`: if `elfutils` is already in the environment, it will be picked up instead of `libelf` - one less dependency.

bernhardkaindl (approved):
Thanks, looks good.

iarspider:
Unless there is a package that can only be built with `libelf`, we can probably deprecate `libelf` and replace all dependencies on `libelf` or `elf` with `elfutils` (since this will be the only provider of `elf`), but this is outside of this PR's scope.

PR #34272, opened 2022-12-02
CI boilerplate reduction

kwryankrattiger:
Reduce boilerplate in the CI yaml by allowing composition based on pipeline targets OS/arch

CC: @blue42u


blue42u:
@kwryankrattiger Thanks for working on this! One high-level question, based on [`ci_process_includes`](https://github.com/spack/spack/pull/34272/files#diff-49f4a785e6a8295a5cd5af6a072cde881c00aecefe40fadf40abb33d5c35f6f5R622) it seems like the `included` configurations are merged and then `mappings` are applied later. If so, this has the [quirky nuance I alluded to before](https://github.com/spack/spack/pull/32300#issuecomment-1319605192) and was hoping to avoid. Given this configuration sketch:
```yaml
---  # common/ci.yaml
ci:
  variables:
    SIZE: "small"
  mappings:
  - match: [llvm]
    runner-attributes:
      variables:
        SIZE: "medium"

---  # big-env/spack.yaml
spack:
  spec:
  - llvm targets=all
  - paraview
  ci:
    include: common
    variables:
      SIZE: "large"  # All the packages in this environment are large
```
In the generated pipeline, a `paraview` build job will be `SIZE:"large"`, but an `llvm` build job will be `SIZE:"medium"`. That doesn't make intuitive sense to me, if I set the `SIZE` in `spack.yaml` I would expect it to override the variable set in the `common` configuration, no matter "where" it was set.

If this PR doesn't already, is there anything preventing the `included` configurations from being applied in order, instead of merged?

kwryankrattiger:
@blue42u So the `ci_process_includes` function is no longer used in favor of adding a "ci.yaml" and "cdash.yaml" and "view.yaml" to spack. That said, you concern persists even with spacks default merging strategy.

To address the quirky behavior, I think that has more to do with how the "scope" is defined. At the top most level, you don't specify the value the must be applied to each job, you specify the "default" value that is populated to all jobs, including the service jobs and signing job.

The next scope that is applied is the job type scope. For the build jobs the `build-job-attributes` section is applied.

The last scope that is applied for build jobs is the mapping scope which will override and append to everything defined in the previous scopes.

This is all relatively consistent with how things were done before, but now configurations are allowed to span multiple files to move boilerplate out of the specific stack files.

To achieve the desired result of forcing all build jobs to set a variable you could either change the custom configuration scopes passed to the generate step to exclude the default mappings (subtle) or specify a mapping to all discovered packages that sets the `variables:SIZE:"large"` (less subtle).

---

One of the things that I am less thrilled about with the default spack merging is that `script` is going to merge using the prepend strategy which was the major difference between the `spack.config.merge_yaml` and `spack.ci._merge_attributes` routines. For `before_script` and `after_script` we probably want to use an `append` strategy and for `script` we always want to override.

I think maybe I am just reading the config stuff backwards and "source" prepending means the the lower precedence configs are processed last and therefore everything after is appended to it. I will run some tests once I get this working again.

But the non-override behavior for script is an issue. @tgamblin @scottwittenburg if either of you have ideas or comments on how this should work.

kwryankrattiger:
One the override behavior, I think I found the solution, `script:: []` should allow overriding values it seems, not sure if there is more needed for that, but that is what the docs seem to imply.

blue42u:
@kwryankrattiger It worries me that you're trying to add key-specific merging behavior here, in https://github.com/spack/spack/pull/32300 I was avoiding that completely (to decouple as much as possible the Spack configuration format from GitLab CI's own schema). There are also weird edge cases I do not believe the Spack config-merge can or should handle, e.g. `cache:` can be a `list` or a `dict` (interpreted as a 1-element `list`).

Regarding "scope," while I understand the explanation and know that the proposed workaround works, I still find it very unintuitive and confusing. It introduces cross-dependencies where one config needs to know where values are set in "previous" configs to even work. Which means a setup using "mixin" configs is more-or-less unmaintainable (or at the very least, *really* confusing).

kwryankrattiger:
>  I do not believe the Spack config-merge can or should handle, e.g. cache: can be a list or a dict (interpreted as a 1-element list).

I did check this, and it is handled by checking if one of them is a list, converting the dict value to a list, and then applying the merge. This should be okay.

>  where one config needs to know where values are set in "previous" configs to even work.

I don't think there is really a way around this. The way spack configs work in general is read -> merge -> get processed by spack command. 

Even if each yaml was read and processed in order in the CI module, which is what we decided not to do by moving to the spack config merge instead of the `ci_process_includes`, what is the correct logic for overriding values at different scopes across different configuration precedence? I feel like the rules would be even more convoluted in that workflow than they are using the default spack workflow.

Having to know which configs are being used, the order of precedence they are specified in, and what values they set I think is an okay requirement since that is already a requirement when using configuration scopes in spack. The biggest difference is the CI module has scopes to make it easier to say "do this for everything" once at a high level rather than having to always specify configs at the lower level. Things like `SIZE` that get specified at a mapping level in general will just always need to be specified at a mapping level.

kwryankrattiger:
> It worries me that you're trying to add key-specific merging behavior here

The most recent changes are trying not to do this because your concerns were echoed by others. The scripts are the only ones that I am a little concerned about the default behavior being applied, but they will probably be fine.

blue42u:
Crazy half-baked idea, what if it was a list of "modifications" applied to the generated pipeline in reverse order:
```yaml
--- # common/ci.yaml
ci:
  pipeline-gen:  # TODO: Better name for this key
  - match_behavior: first
    submapping:
    - match: [llvm]
      build-job-attributes:
        +variables: {SIZE: "medium"}
  - build-job-attributes:
      +variables: {SIZE: "small"}

--- # big-env/spack.yaml
spack:
  include: common/ci.yaml
  spec:
  - llvm targets=all
  - paraview
  ci:
    pipeline-gen:
    - +variables: {SIZE: "large"}
```

After the Spack config-merge, the above two should become the unified config:
```yaml
spack:
  spec:
  - llvm targets=all
  - paraview
  ci:
    pipeline-gen:
    - +variables: {SIZE: "large"}
    - match_behavior: first
      submapping:
      - match: [llvm]
        build-job-attributes:
          +variables: {SIZE: "medium"}
    - build-job-attributes:
        +variables: {SIZE: "small"}
```

The entries below `pipeline-gen:` can then be "applied" to the pipeline in reverse order using a custom merge (e.g. the `+` merge prefix from #32300), and the final end result correctly prefers `spack.yaml` over `common/ci.yaml`.

Still a bit confusing from the reversed order (it would need to be explained as something like "value from first matching key" and "`+` prefix indicates use next matching sub-key"). IMHO it's still less confusing (and more maintainable) than 3 different "scopes" at which changes could be made.

kwryankrattiger:
> Crazy half-baked idea, what if it was a list of "modifications" applied to the generated pipeline in reverse order: 

From what I can tell, this is exactly like what is already there, but less verbose. The biggest change this is making I think is the contents of the mappings list can either be just a runner attributes dict, or a submapping/matches specification. I do like `submapping` over `matches`.

I like the idea of having a better shorthand for saying "everything gets this mapping if it gets here" that is not the somewhat odd `match: ["@:"]` hack to catch all of the things.

kwryankrattiger:
actually...I think I am seeing the more subtle part of this proposal. There would be no top level runner attributes, except for maybe a subset which we can chose like `image`.

I think the way this would look then would be 

```yaml
--- # common/ci.yaml
ci:
  pipeline-gen:  # TODO: Better name for this key
  - match_behavior: first
    submapping:
    - match: [llvm]
      build-job-attributes:
        +variables: {SIZE: "medium"}
  # build-job-attrbutes apply all block
  -  +variables: {SIZE: "small"}

--- # big-env/spack.yaml
spack:
  include: common/ci.yaml
  spec:
  - llvm targets=all
  - paraview
  ci:
    pipeline-gen:
    # stack override apply all block
    - +variables: {SIZE: "large"}
```

blue42u:
I wanted to keep the `build-job-attributes:` key to apply a modification only to the build jobs (and not `rebuild-index`, `no-specs-to-rebuild`, etc.). It is also needed to properly distinguish between `submapping:` the Spack key and `submapping:` the potential GitLab CI key.

But otherwise, exactly!

kwryankrattiger:
Mappings always only apply to the build jobs, so maybe it would make sense to nest the mappings under that section? If we go this route, the idea of having a `build-job-attributes` section kind of becomes redundant. It would make sense to me to just unify all of the job attribute mapping to the `_find_matching_config` routine and call it done. Then I could remove the extra step of merging the `build-job-attributes` section into the top level configs when initializing build job configs.

blue42u:
There's also `signing-job-attributes:`, and from #32300 `noop-job-attributes:`, `reindex-job-attributes:` and `pipeline-attributes:`. I would like to give them different attributes so having separate keys is important in `pipeline-gen:` entries.

The problem with `submapping` under `build-job-attribute:` is that it conflicts with the potential GitLab key `submapping:`, while if you remove `build-job-attributes:` from the `match:` clause you conflict with the potential GitLab key `match:`. It's redundant but necessary.

Basically, for #32300 the parts of the config that are GitLab CI attributes should be considered (practically) a opaque blob of data (except in `_merge_attributes`):
```yaml
--- # common/ci.yaml
ci:
  pipeline-gen:  # TODO: Better name for this key
  - match_behavior: first
    submapping:
    - match: [llvm]
      build-job-attributes: GITLAB BITS
  - build-job-attributes: GITLAB BITS

--- # big-env/spack.yaml
spack:
  include: common/ci.yaml
  spec:
  - llvm targets=all
  - paraview
  ci:
    pipeline-gen:
    - build-job-attributes: GITLAB BITS
```
So, I'm pretty sure all the `build-job-attributes:` keys are needed to make this work properly.

(It looks like there was a missing `build-job-attributes:` in my first example, sorry. This one is correct.)

kwryankrattiger:
Okay, I think I am seeing more where you are going with this, your idea is taking this quite a bit further than I originally was thinking, but I think that is a good thing. It certainly leans more into your work with handling then merge semantics, which I think is good.

I like this idea of the `pipeline-gen` that pushes all of the configuration blocks into a list, rather than trying to track map levels and mixed scope level between configs. In the end, it makes it more consistent when trying to reason about what you actually get in the end and validating that configs are being scoped in the correct order.

This will make the schema a tad more complicated, because there will be three types of items allowed now instead of one, `global`, `<job>-attributes`, and `submapping` items. But, in the ci module code, this could be modeled by looping the `pipeline-gen` list and calling handlers based on the type of scope encountered which would likely simplify parts of the CI module quite a bit since applying configs to jobs will always be done the same, it is only the job selection that would be different.

I would like to get more feedback from @scottwittenburg on this, but I think this is a good direction.

adamjstewart (dismissed):
I like this a lot! This resolves most of my complaints about adding new pipelines. Some additional ideas (alternatives?) worth considering, either in this PR or more likely a follow-up PR:

### Platform based on directory name

I wonder if we could condition arch on pipeline location. For example, move:
```
share/spack/gitlab/cloud_pipelines/stacks/ml-linux-x86_64-cpu/spack.yaml
```
to:
```
share/spack/gitlab/cloud_pipelines/linux/x86_64/ml-cpu/spack.yaml
```
and have the arch automatically detected based on the file location.

### Matrix of platforms

For a lot of (most?) pipelines, we want to build a set of packages on most (all?) platforms. For example, for the ML stack, I'm planning on testing the following cross product:

* platform: linux, darwin
* target: x86_64, aarch64, ppc64le
* variants: +cuda, +rocm, ~cuda~rocm

This requires me to duplicate the list of packages I want to build in 18 files. If I could instead specify a matrix in `spack.yaml` and exclude known invalid combinations (no such thing as darwin ppc64le, tensorflow doesn't build on aarch64) that would make it much easier to maintain. As is, if I want to add a new package to the list, I would have to update 18 files.

See #32894 and #32893 for details on this.

P.S. Not trying to hold up this wonderful PR by suggesting more work, just wanted to put these ideas out here so people have them in mind while working on this refactor. I'm guessing these will end up being follow-up PRs in the end.

Will review again after this is rebased, the ML pipelines have changed quite a bit since this PR branch was first created and I want to make sure nothing is lost in the refactor.

blue42u:
Thanks @adamjstewart for pushing this PR forward!

I'd like to request one change to be made before this is merged. Currently this PR changes the `*script:` fields representation to strings containing YAML code, I would much prefer if they were instead recursive lists of strings as is supported by the GitLab CI format. This allows sharing script-fragments between parts of the configuration by using YAML anchors, which I use frequently in my `spack ci` configuration, e.g.:
```yaml
  ci:
    pipeline-gen:
    - reindex-job:
        script:
          - &init_spack
            - git clone ...
            - source ...
          - spack buildcache update-index ...
    - build-job:
        tags: [foo]
        script:
          - *init_spack
          - spack ci rebuild ...
```
I already raised this request with @kwryankrattiger over Slack DMs, we settled on an idea to reduce surprises when using this format: `*script:` keys default to override (`*script::`) unless explicitly indicated as merge (`*script+:`), all other keys merge by default (unless explicitly indicated as override). Since "merge" for lists is confusingly prepend (instead of append) and `*script:` keys are uniquely sensitive to order, this reduces surprises by making "merge" opt-in.

kwryankrattiger:
@blue42u 
I think lists of lists will solve the issue I had just as well as strings, and since it is already compatible with GitLab I think that is a great solution. I will add that today and rebase on master...hopefully that goes well.

@adamjstewart 
TL;DR, I think both ideas are good but I want to keep this change mostly to changes related to YAML structure changes. The restructuring in this PR right now is just a subset of the changes it makes possible going forward.

> Platform based on directory name

I think this could make sense, then the `.gitlabci.yml` script for ml pipelines could be reused for each stack. In general, it could change the setup to just be set the platform/arch/pipeline name environment variables. Since this is more of a spack CI setup chawnge vs a yaml one I think this could come after this PR.

> Matrix of platforms

I think we get this for free using the custom config scopes. The ML pipeline should be able to maintain a common `spack.yaml` and then add that to the CLI configuration scope list in the setup pipelines. I think this is also something the can be added after this since it is not changing any of the core YAML. If it isn't already there, I think it should be relatively easy to add @tgamblin ?

lib/spack/spack/config.py
```diff
@@ -1044,6 +1053,34 @@ def _override(string):
...
+    """Test if a spack YAML string is an override.
+
+    See ``spack_yaml`` for details.  Keys in Spack YAML can end in `+:`,
+    and if they do, their values append lower-precedence
+    configs.
+
+    str, str : concatenate strings.
+    [str], str : append string to list. (maybe)
+
+    """
```
scottwittenburg, line 1074: This docstring looks identical to the one above, probably copy-pasted and needs update?

lib/spack/spack/config.py
```diff
@@ -360,6 +363,12 @@ def _process_dict_keyname_overrides(data):
             if sk.endswith(":"):
                 key = syaml.syaml_str(sk[:-1])
                 key.override = True
+            elif sk.endswith("+"):
+                key = syaml.syaml_str(sk[:-1])
+                key.prepend = True
+            elif sk.endswith("-"):
+                key = syaml.syaml_str(sk[:-1])
+                key.append = True
```
scottwittenburg, line 370: I think these new config modifiers should be explained in the PR description.  Then the content could maybe be used to update the config section of the docs.

lib/spack/spack/ci.py
```diff
@@ -595,12 +728,36 @@ def generate_gitlab_ci_yaml(
...
     yaml_root = ev.config_dict(env.yaml)
 
-    if "gitlab-ci" not in yaml_root:
-        tty.die('Environment yaml does not have "gitlab-ci" section')
-
-    gitlab_ci = yaml_root["gitlab-ci"]
-
-    cdash_handler = CDashHandler(yaml_root.get("cdash")) if "cdash" in yaml_root else None
+    # Add config scopes to environment
+    if True:
```
scottwittenburg: Why is this in an if block if the condition is just `True`?

lib/spack/spack/ci.py
```diff
@@ -595,12 +728,36 @@ def generate_gitlab_ci_yaml(
...
+            and os.path.exists(s.path)
+        ]
+        include_scopes = []
+        for scope in cli_scopes:
+            if scope not in include_scopes and scope not in env_includes:
+                include_scopes.insert(0, scope)
+        env_includes.extend(include_scopes)
+        yaml_root["include"] = env_includes
+        with open(os.path.join(env.path, "spack.yaml"), "w") as fd:
+            fd.write(syaml.dump_config(env.yaml, default_flow_style=False))
```
scottwittenburg: Is this overwriting the active environment with the merged config?  That seems weird to me, but maybe I just don't get it yet.  If this is the intent, I think it warrants a comment.

lib/spack/spack/test/cmd/ci.py
```diff
@@ -229,6 +234,13 @@ def test_ci_generate_with_env(
 
             assert "rebuild-index" in yaml_contents
             rebuild_job = yaml_contents["rebuild-index"]
+            with capsys.disabled():
+                print("Summary...")
```
scottwittenburg: These `print` statements should be removed I think.

lib/spack/spack/ci.py
```diff
@@ -546,6 +486,235 @@ def get_spec_filter_list(env, affected_pkgs):
...
+                    "script:": [
+                        "cd {env_dir}",
+                        "spack env activate --without-view .",
+                        "spack ci rebuild",
+                    ],
+                }
+            },
+        ]
+
+        # Job overrides
```
scottwittenburg, line 612: So are these the bits of different jobs that the user has no control over?  I.e. we're forcing it here, regardless of what they put?

scottwittenburg: I guess that's what we did before, but now, it's easier to see what's going on :+1:

lib/spack/spack/ci.py
```diff
@@ -1367,7 +1517,7 @@ def generate_gitlab_ci_yaml(
             sys.exit(1)
 
     with open(output_file, "w") as outf:
-        outf.write(syaml.dump_config(sorted_output, default_flow_style=True))
+        outf.write(syaml.dump_config(sorted_output, default_flow_style=False))
```
scottwittenburg: Can the flow style of the output be an option provided on the command line?  

Just spitballing the option here, but maybe something like `--output-style readable|condensed`?  I feel like most of the time, this is consumed by gitlab and never seen by a human, and adding the extra size for readability might not make sense in those situations.  The option could be `readable` by default, and we could pass `condensed` from gitlab ci or anywhere else we're concerned about yaml size.

scottwittenburg (commented):
After talking with you about this many times over the past few weeks, I'm starting to have a pretty good sense of how this will change how we define pipelines, and I think it's going to be great!

For anyone who didn't talk to you about it many times, I think this PR needs a description summarizing the changes and giving some examples.

I've added some initial questions and observations, but I'll make another pass when you're ready.

lib/spack/spack/ci.py
```diff
@@ -833,6 +1026,13 @@ def generate_gitlab_ci_yaml(
         else:
             broken_spec_urls = web_util.list_url(broken_specs_url)
 
+    # Make sure all reserved tags are removed for all jobs
```
scottwittenburg: I don't think this comment makes sense here.

lib/spack/spack/ci.py
```diff
@@ -546,6 +486,235 @@ def get_spec_filter_list(env, affected_pkgs):
...
+            has_submapping = "submapping" in section
+            section = cfg.InternalConfigScope._process_dict_keyname_overrides(section)
+
+            if name:
+                remove_job_name = self.__job_name(name, suffix="-remove")
+                merge_job_name = self.__job_name(name)
+                do_remove = remove_job_name in section
+                do_merge = merge_job_name in section
+
+                def _apply_section(attrs, section):
```
scottwittenburg: Why define this function within the loop?  Could you move it outside to be defined only once if you pass in `do_merge` and `do_remove` as parameters?

lib/spack/spack/ci.py
kwryankrattiger: The point of this is to add the includes to the environment to make it easier for someone using the reproducer. I think it is probably sufficient to only write this out to the copied spack.yaml as we discussed earlier.

lib/spack/spack/ci.py
kwryankrattiger: I thought about making it an option..but I don't know of a good use case to not do this operation so maybe I will leave that as always on until it creates an issue.

lib/spack/spack/ci.py
kwryankrattiger: I am using it more like a lambda with a capture list, I thought it was a little more readable to just say "apply the section to the job attributes" without the extra args. If this isn't idiomatic python I can change it though, I didn't think it cost that much to define a function in a loop.

lib/spack/docs/pipelines.rst
```diff
@@ -354,113 +351,27 @@ arguments you can pass to ``spack ci reproduce-build`` in order to reproduce
...
-progresses, this build group may have jobs added or removed. The url, project,
-and site are used to specify the CDash instance to which build results should
-be reported.
-
-Take a look at the
-`schema <https://github.com/spack/spack/blob/develop/lib/spack/spack/schema/gitlab_ci.py>`_
-for the gitlab-ci section of the spack environment file, to see precisely what
-syntax is allowed there.
+.. code-block:: bash
+  cd ${concrete_environment_dir}
```
scottwittenburg, line 379: Adding a newline here fixes the doc build error for me locally.  I did find some mention of this buried in the `Directives` section of the sphinx rst docs [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives).

lib/spack/docs/pipelines.rst
```diff
@@ -354,113 +351,27 @@ arguments you can pass to ``spack ci reproduce-build`` in order to reproduce
...
+  spack env activate --without-view .
+  spack ci rebuild
 
 .. _rebuild_index:
 
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Note about rebuilding buildcache index
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^
```
scottwittenburg: You have to extend these to the end of the text they surround, or else they end up rendered like this:

broken_section

lib/spack/docs/pipelines.rst
```diff
@@ -475,21 +386,33 @@ not correctly reflect the mirror's contents at the end of a pipeline.
...
 spack generates a job to update the buildcache index of the target mirror
 at the end of each pipeline by default.  You can disable this behavior by
-adding ``rebuild-index: False`` inside the ``gitlab-ci`` section of your
-spack environment.  Spack will assign the job any runner attributes found
-on the ``service-job-attributes``, if you have provided that in your
-``spack.yaml``.
+adding ``rebuild-index: False`` inside the ``ci`` section of your
+spack environment.
+
+Reindex jobs do not allow modifying the ``script`` attribute since it is automatically generate using the target mirror listed in the ``mirrors::mirror`` configuration.
```
scottwittenburg:
```suggestion
Reindex jobs do not allow modifying the ``script`` attribute since it is automatically generated using the target mirror listed in the ``mirrors::mirror`` configuration.
```

lib/spack/docs/pipelines.rst
```diff
@@ -499,47 +422,78 @@ Following is an example of this section added to a ``spack.yaml``:
...
+             name: 'some.image.registry/custom-image:latest'
+             entrypoint: ['/bin/bash']
+           script::
+             - echo "Custom message in a custom script"
 
 The example above illustrates how you can provide the attributes used to run
 the NO-OP job in the case of an empty pipeline.  The only field for the NO-OP
 job that might be generated for you is ``script``, but that will only happen
-if you do not provide one yourself.
+if you do not provide one yourself. Notice in this example the ``script`` uses the ``::`` notation to perscribe override behavior. Without this, the ``echo`` command would have been prepended to the automatically generated script rather than replacing it.
```
scottwittenburg:
```suggestion
if you do not provide one yourself. Notice in this example the ``script`` uses the ``::`` notation to prescribe override behavior. Without this, the ``echo`` command would have been prepended to the automatically generated script rather than replacing it.
```

lib/spack/docs/pipelines.rst
```diff
@@ -499,47 +422,78 @@ Following is an example of this section added to a ``spack.yaml``:
...
+      build-group: Release Testing
+      url: https://cdash.spack.io
+      project: Spack
+      site: Spack AWS Gitlab Instance
+
+The ``ci`` config section is used to configure how the pipeline workload should be
+generated, mainly how the jobs for building specs should be assigned to the
+configured runners on your instance. The main section for configuring pipelines
+is ``pipeline-gen``, which is a list of job attribute sections that are merged,
+using the same rules as Spack configs (TODO:ref:`spack_configs`), from the bottom up.
```
scottwittenburg: What is the TODO about here?  Just making a link to another section of the docs, or something more?

lib/spack/docs/pipelines.rst
```diff
@@ -499,47 +422,78 @@ Following is an example of this section added to a ``spack.yaml``:
...
+The order sections are applied is to be consistent with how spack orders scope precendense when merging lists.
```
scottwittenburg:
```suggestion
The order sections are applied is to be consistent with how spack orders scope precedence when merging lists.
```

lib/spack/docs/pipelines.rst
```diff
@@ -475,21 +386,33 @@ not correctly reflect the mirror's contents at the end of a pipeline.
...
+adding ``rebuild-index: False`` inside the ``ci`` section of your
+spack environment.
+
+Reindex jobs do not allow modifying the ``script`` attribute since it is automatically generate using the target mirror listed in the ``mirrors::mirror`` configuration.
+
+^^^
+Signing (signing)
+^^^
+
+This job is run after all of the rebuild jobs are completed and is intended to be used to sign the package binaries built by a protected CI run. Signing jobs are generated only if a signing job ``script`` is specified and the spack CI job type is protected. Note, if an ``any-job`` section contains a script, this will not implicitly create a ``signing`` job, a signing job may only exist if it is explicitly specified in the configuration with a ``script`` attribute. Specifying a signing job without a script does not create a signing job and the job configuration attributes will be ignored. Signing jobs are always assigned the runner tags ``aws``, ``protected``, and ``notary``.
```
scottwittenburg: Everywhere else in this document and others, it seems newlines have been used to control line length.  We should probably do the same with all your additions here.

lib/spack/docs/pipelines.rst
```diff
@@ -499,47 +422,78 @@ Following is an example of this section added to a ``spack.yaml``:
...
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+^^^
+Job Attribute Sections
+^^^
+
+Each type of job may have attributes added or removed via sections in the ``pipeline-gen`` list. Job type specific attributes may be specified using the keys ``<type>-job`` to add attributes to all jobs of type ``<type>`` or ``<type>-job-remove`` to remove attributes of type ``<type>``. Each section may only contain one type of job attribute specification, ie. , ``build-job`` and ``noop-job`` may not coexist but ``build-job`` and ``build-job-remove`` may. Note, ``*-remove`` specifications are applied before the additive attribute specification.
+
+All of the attributes specified are forwarded to the generated CI jobs, however special treatment is applied to the attributes ``tags``, ``image``, ``variables``, ``script``, ``before_script``, and ``after_script`` as they are components recognized explicitly by the Spack CI generator. For the ``tags`` attribute, Spack will remove reserved tags (:ref:`reserved_tags`) from all jobs specified in the config. In some cases, such as for ``signing`` jobs, reserved tags will be added back based on the type of CI that is being run.
```
scottwittenburg: The next two paragraphs following this one probably need updated to remove mention of `runner-attributes`, which I don't think is used anymore, is it?

lib/spack/docs/pipelines.rst
```diff
@@ -565,46 +519,27 @@ environment directory is located within your ``--artifacts_root`` (or if not
 provided, within your ``$CI_PROJECT_DIR``), activates that environment for
 you, and invokes ``spack ci rebuild``.
 
-.. _staging_algorithm:
+Sections that specify scripts (``script``, ``before_script``, ``after_script``) are all read as lists of commands or lists of lists of commands. It is recommended to write scripts as lists of lists if scripts will be composes via merging. The default behavior of merging lists will remove duplicate commands and potentially apply unwanted reordering, where as merging lists of lists will preserve the local ordering and never removes duplicate commands. When writting commands to the CI target script, all lists are expanded and flattened into a single list.
```
scottwittenburg:
```suggestion
Sections that specify scripts (``script``, ``before_script``, ``after_script``) are all read as lists of commands or lists of lists of commands. It is recommended to write scripts as lists of lists if scripts will be composed via merging. The default behavior of merging lists will remove duplicate commands and potentially apply unwanted reordering, whereas merging lists of lists will preserve the local ordering and never removes duplicate commands. When writing commands to the CI target script, all lists are expanded and flattened into a single list.
```

lib/spack/docs/pipelines.rst
```diff
@@ -565,46 +519,27 @@ environment directory is located within your ``--artifacts_root`` (or if not
...
-Gitlab CI rules about jobs/stages.  In the staging process the goal is to maximize
-the number of jobs in any stage of the pipeline, while ensuring that the jobs in
-any stage only depend on jobs in previous stages (since those jobs are guaranteed
-to have completed already).  As a runner is determined for a job, the information
-in the ``runner-attributes`` is used to populate various parts of the job
-description that will be used by Gitlab CI. Once all the jobs have been assigned
-a runner, the ``.gitlab-ci.yml`` is written to disk.
+A special case of attribute specification is the ``submapping`` section which may be used to apply job attributes to build jobs based on the package spec associated with the rebuild job. Submapping is specified as a list of spec ``match`` lists associated with ``build-job``/``build-job-remove`` sections. There are two option for ``match_behavior``, either ``first`` or ``merge`` may be specified. In either case, the ``submapping`` list is processed from the bottom up, and then each ``match`` lists is searched for a string that satisfies the check ``spec.satisfies({match_item})`` for each concrete spec.
```
scottwittenburg:
```suggestion
A special case of attribute specification is the ``submapping`` section which may be used to apply job attributes to build jobs based on the package spec associated with the rebuild job. Submapping is specified as a list of spec ``match`` lists associated with ``build-job``/``build-job-remove`` sections. There are two options for ``match_behavior``, either ``first`` or ``merge`` may be specified. In either case, the ``submapping`` list is processed from the bottom up, and then each ``match`` list is searched for a string that satisfies the check ``spec.satisfies({match_item})`` for each concrete spec.
```

lib/spack/docs/configuration.rst
```diff
@@ -239,6 +239,11 @@ lower-precedence settings. Completely ignoring higher-level configuration
 options is supported with the ``::`` notation for keys (see
 :ref:`config-overrides` below).
 
+There are also special notations for string concatenation and precendense override.
```
scottwittenburg, line 245:
```suggestion
There are also special notations for string concatenation and precedence override.
```

lib/spack/docs/configuration.rst
```diff
@@ -279,6 +284,47 @@ command:
        - ~/.spack/stage
 
 
+.. _config-prepend-append:
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+String Concatenation
+^^^^^^^^^^^^^^^^^^^^^^^^^^
```
scottwittenburg: While it works to extend the section specifiers past the end of the text in between, I don't see it happening anywhere else, so maybe best to adhere to the convention of making the lengths match.

lib/spack/docs/configuration.rst
```diff
@@ -279,6 +284,47 @@ command:
...
+
+Above, the user ``config.yaml`` *completely* overrides specific settings in the
+default ``config.yaml``. Sometimes, it is useful to add addition a suffix/prefix
```
scottwittenburg:
```suggestion
default ``config.yaml``. Sometimes, it is useful to add a suffix/prefix
```

lib/spack/docs/pipelines.rst
```diff
@@ -9,27 +9,32 @@
...
-`here <https://about.gitlab.com/product/continuous-integration/>`_) and configure
-at least one `runner <https://docs.gitlab.com/runner/>`_.  Then the basic steps
-for setting up a build pipeline are as follows:
+To get started with automated build pipelines a Gitlab instance with version ``>= 12.9``
+(more about Gitlab CI `here <https://about.gitlab.com/product/continuous-integration/>`_)
+with at least one `runner <https://docs.gitlab.com/runner/>`_ configured is required. This
+can be done quickly by setting up a local Gitlab instance.
 
-#. Create a repository on your gitlab instance
+It is possible to set up pipelines on gitlab.com the builds there are limited to
```
scottwittenburg:
```suggestion
It is possible to set up pipelines on gitlab.com, but the builds there are limited to
```

lib/spack/docs/pipelines.rst
```diff
@@ -9,27 +9,32 @@
...
+60 minutes and generic hardware.  It is possible to
+`hook up <https://about.gitlab.com/blog/2018/04/24/getting-started-gitlab-ci-gcp>`_
+Gitlab to Google Kubernetes Engine (`GKE <https://cloud.google.com/kubernetes-engine/>`_)
+or Amazon Elastic Kubernetes Service (`EKS <https://aws.amazon.com/eks>`_), though those
+topics are outside the scope of this document.
+
+After setting up a Gitlab instance for running CI, the basic steps for setting up a build pipeline are as follows:
+
+#. Create a repository in the Gitlab instance CI and a runner enabled.
```
scottwittenburg:
```suggestion
#. Create a repository in the Gitlab instance CI and enable a runner.
```

lib/spack/docs/pipelines.rst
```diff
@@ -132,24 +130,21 @@ And here's the spack environment built by the pipeline represented as a
...
+           before_script:
+             - git clone ${SPACK_REPO}
+             - pushd spack && git checkout ${SPACK_CHECKOUT_VERSION} && popd
+             - . "./spack/share/spack/setup-env.sh"
+       - build-job:
+           tags: [docker]
+           image:
+             name: ghcr.io/scottwittenburg/ecpe4s-ubuntu18.04-runner-x86_64:2020-09-01
+             entrypoint: [""]
+
```
scottwittenburg, line 147: Does it make sense below here, to mention why you don't see any `script` section, and that one from a config scope will be inserted for you by default?

lib/spack/docs/pipelines.rst
```diff
@@ -354,113 +351,27 @@ arguments you can pass to ``spack ci reproduce-build`` in order to reproduce
...
 ------------------------------------
 
-Here's an example of a spack environment file that has been enhanced with
-sections describing a build pipeline:
+^^^
+Rebuild (build)
+^^^
 
-.. code-block:: yaml
+Rebuild jobs, denoted as ``build-job``'s in the ``pipeline-gen`` list, are jobs associated with a concrete spec that have been marked for rebuild. By default a simple script for doing rebuild is generated, but may be modified as needed.
```
scottwittenburg:
```suggestion
Rebuild jobs, denoted as ``build-job``'s in the ``pipeline-gen`` list, are jobs associated with concrete specs that have been marked for rebuild. By default a simple script for doing rebuild is generated, but may be modified as needed.
```

lib/spack/docs/pipelines.rst
```diff
@@ -475,21 +386,33 @@ not correctly reflect the mirror's contents at the end of a pipeline.
...
+Signing (signing)
+^^^
+
+This job is run after all of the rebuild jobs are completed and is intended to be used to sign the package binaries built by a protected CI run. Signing jobs are generated only if a signing job ``script`` is specified and the spack CI job type is protected. Note, if an ``any-job`` section contains a script, this will not implicitly create a ``signing`` job, a signing job may only exist if it is explicitly specified in the configuration with a ``script`` attribute. Specifying a signing job without a script does not create a signing job and the job configuration attributes will be ignored. Signing jobs are always assigned the runner tags ``aws``, ``protected``, and ``notary``.
+
+^^^
+Cleanup (cleanup)
+^^^
+
+When using ``temporary-storage-url-prefix`` the cleanup job will destroy the mirror created for the parent Gitlab pipeline job. Cleanup jobs do not allow modifying the script, but do expect that the spack command is in the path and require a ``before_script`` to be specified that sources the ``setup-env.sh`` script.
```
scottwittenburg:
```suggestion
When using ``temporary-storage-url-prefix`` the cleanup job will destroy the mirror created for the associated Gitlab pipeline. Cleanup jobs do not allow modifying the script, but do expect that the spack command is in the path and require a ``before_script`` to be specified that sources the ``setup-env.sh`` script.
```

lib/spack/spack/ci.py
```diff
@@ -546,6 +486,245 @@ def get_spec_filter_list(env, affected_pkgs):
...
+        self.ir = {
+            "jobs": {},
+            "temporary-storage-url-prefix": self.ci_config.get(
+                "temporary-storage-url-prefix", None
+            ),
+            "enable-artifacts-buildcache": self.ci_config.get(
+                "enable-artifacts-buildcache", False
+            ),
+            "bootstrap": self.ci_config.get(
+                "enable-artifacts-buildcache", []
```
scottwittenburg:
```suggestion
                "bootstrap", []
```

scottwittenburg: I'm not sure, but I don't think you want to mix up `bootstrap` and `enable-artifacts-buildcache` here.

lib/spack/spack/ci.py
```diff
@@ -546,6 +486,245 @@ def get_spec_filter_list(env, affected_pkgs):
...
+            ),  # This is deprecated and should be removed
+            "rebuild-index": self.ci_config.get("enable-artifacts-buildcache", True),
```
scottwittenburg: Same, why `enable-artifacts-buildcache` here?

lib/spack/spack/ci.py
```diff
@@ -546,6 +486,245 @@ def get_spec_filter_list(env, affected_pkgs):
...
+            "broken-specs-url": self.ci_config.get("enable-artifacts-buildcache", None),
```
scottwittenburg: ditto `enable-artifacts-buildcache`

lib/spack/spack/ci.py
```diff
@@ -546,6 +486,245 @@ def get_spec_filter_list(env, affected_pkgs):
...
+            "broken-tests-packages": self.ci_config.get("enable-artifacts-buildcache", []),
```
scottwittenburg: ditto `enable-artifacts-buildcache`

lib/spack/spack/ci.py
scottwittenburg: I thought it was also weird to shadow the closure variable `section` with the function parameter of the same name, but whatever.  I'm not an idiomatic python person, I just thought this was weird and suspicious when I was debugging the extra `:` getting written out in the final gitlab yaml.  But it turned out to have nothing to do with the issue, so I guess it's fine the way it is.

lib/spack/spack/ci.py
scottwittenburg: I think my comment here is still valid, isn't it?

lib/spack/spack/config.py
scottwittenburg: Looks like you did this already, thanks!

share/spack/gitlab/cloud_pipelines/configs/config.yaml
```diff
@@ -0,0 +1,8 @@
+config:
+  build_jobs: 32
+  concretizer: clingo
+  install_tree:
+    #root: /home/software/spack
```
scottwittenburg: Is this just commented out for testing?

share/spack/gitlab/cloud_pipelines/configs/linux/x86_64_v1
```diff
@@ -0,0 +1 @@
+x86_64
```
scottwittenburg: Is there such a thing as `x86_64_v1`?  I thought it was just `x86_64`.

share/spack/gitlab/cloud_pipelines/configs/linux/x86_64_v2
```diff
@@ -0,0 +1 @@
+x86_64
```
scottwittenburg: Is this missing the `_v2`?

share/spack/gitlab/cloud_pipelines/configs/linux/x86_64_v3
```diff
@@ -0,0 +1 @@
+x86_64
```
scottwittenburg: Is this missing the `_v3`?

share/spack/gitlab/cloud_pipelines/configs/linux/x86_64_v4
```diff
@@ -0,0 +1 @@
+x86_64
```
scottwittenburg: Is this missing the `_v4`?

share/spack/gitlab/cloud_pipelines/stacks/e4s-mac/spack.yaml
```diff
@@ -39,39 +39,15 @@ spack:
...
-      - lambda
-
-  cdash:
-    build-group: E4S Mac
-    url: https://cdash.spack.io
-    project: Spack Testing
-    site: Cloud Gitlab Infrastructure
+  ci:
+    pipeline-gen:
+    - cleanup-job:
```
scottwittenburg, line 31: Is this cleanup-job configured because the e4s-mac stack is using the temporary storage feature?  If so, I think both could just be removed.

scottwittenburg (changes requested):
Wow, it's great to see all that yaml getting whacked out!  This pass I focused mainly on the docs, which is looking good, thank you.  I'll make another pass when you're ready.  Thanks again @kwryankrattiger  :tada: :balloon:

lib/spack/spack/ci.py
kwryankrattiger: yup, this got moved into `generate_ir`

lib/spack/spack/ci.py
kwryankrattiger: ah, yeah, shadowing is not really good...I will at least change the name here to make it clear it is the passed parameter not the loop parameter.

share/spack/gitlab/cloud_pipelines/configs/linux/x86_64_v1
kwryankrattiger: I think so. I was mostly just copy-pasting to cover my bases. I came up with a new strategy though since the auto detection wasn't working reliably.

share/spack/gitlab/cloud_pipelines/stacks/e4s-mac/spack.yaml
kwryankrattiger: I put it in to keep it consistent, but it looks like all of these macos jobs are disabled at the .gitlab-ci.yml level. I will leave removing it to @eugeneswalker since this is an E4S stack.

lib/spack/docs/pipelines.rst
kwryankrattiger: yeah, TODO, figure out how to reference the other section in the docs.

lib/spack/docs/pipelines.rst
kwryankrattiger: Since enabling CI and Runners are configs in each gitlab repo, I am trying to convey that the repo they create must have those components enabled. Would it be better as `with CI and a runner enabled`?

scottwittenburg: Yeah, that sounds good to me.

share/spack/gitlab/cloud_pipelines/configs/ci.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - printf "fef1f59e56d2d11e6d700ba22d3444b6e583c663d6883fd0a4f63ab8bd280f0f gmake.tar.gz" | sha256sum --check --strict --quiet
+        - tar -xzf gmake.tar.gz -C /usr bin/make 2> /dev/null
+        - . "./share/spack/setup-env.sh"
+        - spack --version
+        - spack arch
+        - cd ${SPACK_CONCRETE_ENV_DIR}
+        - spack env activate --without-view .
+        - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}:'morepadding/{architecture}/{compiler.name}-{compiler.version}/{name}-{version}-{hash}'"
+        - mkdir -p ${SPACK_ARTIFACTS_ROOT}/user_data
+        - if [[ -r /mnt/key/intermediate_ci_signing_key.gpg ]]; then spack gpg trust /mnt/key/intermediate_ci_signing_key.gpg; fi
```
scottwittenburg, line 20: Since you opened this PR, some things changed in the stacks on `develop`, and it won't be easy to know what might be missing, since things got moved all around.  One thing I know is missing for sure is support for UO doing protected builds and doing signing, see e.g. [here](https://github.com/spack/spack/blob/develop/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml#L253-L256).  Can you add that functionality here and maybe take a quick look to see if you can spot any other differences between the script there and here?

lib/spack/docs/pipelines.rst
```diff
@@ -680,6 +675,86 @@ environment/stack file, and in that case no bootstrapping will be done (only the
...
+^^^^^^^^^^^^^
+Reserved Tags
+^^^^^^^^^^^^^
+
+Spack has a subset of tags (``public``, ``protected``, and ``notary``) that it reserves
+for classifying runners that may require special permissions or access. The tags
+``public`` and ``protected`` are used to distinguish between runners that use public
+permissions and runners with protected permissions. The ``notary`` tag is a special tag
+that is used to indicate runners that have access to the highly protected information
+used for signing binaries using a the ``signing`` job.
```
scottwittenburg:
```suggestion
used for signing binaries using the ``signing`` job.
```

lib/spack/spack/ci.py
```diff
@@ -546,6 +486,245 @@ def get_spec_filter_list(env, affected_pkgs):
...
+
+                if name == "build":
+                    # Apply attributes to all build jobs
+                    for _, job in jobs.items():
+                        if job["spec"]:
+                            _apply_section(job["attributes"], section)
+                elif name == "any":
+                    # Apply section attributes too all jobs
+                    for _, job in jobs.items():
+                        _apply_section(job["attributes"], section)
```
scottwittenburg, line 663: I was expecting a case for `noop-job` here, what am I missing?

lib/spack/spack/ci.py
```diff
@@ -839,6 +1041,12 @@ def generate_gitlab_ci_yaml(
         else:
             broken_spec_urls = web_util.list_url(broken_specs_url)
 
+    spack_ci = SpackCI(ci_config, phases, staged_phases)
+    spack_ci_ir = spack_ci.generate_ir()
+
+    with open("spack_ci_ir.txt", "w") as f:
```
scottwittenburg: Are you just writing out the ir for debugging purposes?  Should it be optional or always produced?

lib/spack/spack/schema/cdash.py
```diff
@@ -15,7 +15,8 @@
     "cdash": {
         "type": "object",
         "additionalProperties": False,
-        "required": ["build-group", "url", "project", "site"],
+        # "required": ["build-group", "url", "project", "site"],
```
scottwittenburg, line 18: Why are `url`, `project` and `site` no longer required?  And if they aren't required, we probably don't want this comment in here anymore.

share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
```diff
@@ -7,6 +7,27 @@ default:
 # Job templates
 ########################################
 
+# CI Platform-Arch
+.linux_x86_64:
```
scottwittenburg: Should this be `.darwin_x86_64`?

scottwittenburg:
Whew, this is a big change!  It's looking pretty good to me at this point, but I just added a few more one-off comments for you to consider.

When do you think it will be ready for a "rebuild everything"?

kwryankrattiger:
> When do you think it will be ready for a "rebuild everything"?

As soon as I get a normal passing CI I am going to run the rebuild-everything

lib/spack/spack/ci.py
kwryankrattiger: noop is handled via the fall through else

lib/spack/spack/schema/cdash.py
kwryankrattiger: This is an issue that arose because part of cdash is specified in one config scope, and the rest of the config is in the stack `spack.yaml` They are all ultimately required.

kwryankrattiger:
@spackbot run pipeline

lib/spack/spack/ci.py
```diff
@@ -757,6 +942,27 @@ def generate_gitlab_ci_yaml(
     shutil.copyfile(env.manifest_path, os.path.join(concrete_env_dir, "spack.yaml"))
     shutil.copyfile(env.lock_path, os.path.join(concrete_env_dir, "spack.lock"))
 
+    with open(env.manifest_path, "r") as env_fd:
```
kwryankrattiger, line 894: @alalazo @tgamblin do you guys know of a better way to do this? The goal here is to add the spack config paths to the concrete environment yaml `spack.include` section so the extra config scopes only need to be specified once in the root `.gitlab-ci.yaml` rather than having to add them every time we call spack for CI.

One issue I would like to resolve is that the includes are kind of fragile. One issue that we found was the reproducer puts SPACK_ROOT in a different path than CI. This is fixable in the reproducer, but it would be nice to be able to point those includes at something like `$spack/share/spack/gitlab/cloud_pipelines/...` similarly to how the `config` section handles paths.

Another issue is there is no programmatic or composition method available to add includes to an environment. I could make one by moving includes into a "section" schema but I am not sure if that is be best course since that could in theory create recursive includes where includes include scopes of the includes that included the include scope.

kwryankrattiger:
@spackbot rebuild everything

kwryankrattiger:
@spackbot rebuild everything

lib/spack/spack/environment/environment.py
```diff
@@ -773,7 +773,7 @@ def _read_manifest(self, f, raw_yaml=None):
         )
         self.spec_lists[user_speclist_name] = user_specs
 
-        enable_view = config_dict(self.yaml).get("view")
+        enable_view = spack.config.get("view")
```
kwryankrattiger: @tgamblin @alalazo I changed this here to match moving the `view` section to a config section. In the write manifest section it looks at the env view and writes it out. I think that is fine..but maybe this needs to be re-thought now?

kwryankrattiger:
So in the spirit of removing hurdles, the view config is unchanged. It seems that view configuration is deeply embedded in environment and requires more in-depth investigation to handle correctly.

kwryankrattiger:
@spackbot rebuild everything

kwryankrattiger:
@spackbot rebuild everything

kwryankrattiger:
@spackbot rebuild everything

kwryankrattiger:
Itemizing all of the failures in "rebuild everything".
@scottwittenburg 

Hash issue. May a GitHub fix gone stale? @haampie I think you have seen this before [here](https://github.com/spack/spack/pull/35161)
* https://gitlab.spack.io/spack/spack/-/jobs/5446323 libxc
* https://gitlab.spack.io/spack/spack/-/jobs/5443298 libxc
* https://gitlab.spack.io/spack/spack/-/jobs/5447261 wrf
* https://gitlab.spack.io/spack/spack/-/jobs/5446343 wrf
* https://gitlab.spack.io/spack/spack/-/jobs/5447484 wrf

Other
* https://gitlab.spack.io/spack/spack/-/jobs/5447338 - Network issue
* https://gitlab.spack.io/spack/spack/-/jobs/5447682 - paraview Log too long
* https://gitlab.spack.io/spack/spack/-/jobs/5447581 - py-tensorlfow log too long
* https://gitlab.spack.io/spack/spack/-/jobs/5447695 - py-tensorlfow log too long
* https://gitlab.spack.io/spack/spack/-/jobs/5442935 - pdt%oneapi, cannot reproduce locally

Didn't run rebuild script for some reason, no artifacts
* https://gitlab.spack.io/spack/spack/-/jobs/5447304
* https://gitlab.spack.io/spack/spack/-/jobs/5447591
* https://gitlab.spack.io/spack/spack/-/jobs/5447591

Lost runner (@zackgalbreath )
* https://gitlab.spack.io/spack/spack/-/jobs/5447465
* https://gitlab.spack.io/spack/spack/-/jobs/5447466
* https://gitlab.spack.io/spack/spack/-/jobs/5447419
* https://gitlab.spack.io/spack/spack/-/jobs/5446316
* https://gitlab.spack.io/spack/spack/-/jobs/5447716
* https://gitlab.spack.io/spack/spack/-/jobs/5443251

Deprecated zlib packages
* https://gitlab.spack.io/spack/spack/-/jobs/5443182
* https://gitlab.spack.io/spack/spack/-/jobs/5443187
* https://gitlab.spack.io/spack/spack/-/jobs/5443208

kwryankrattiger:
@spackbot run pipeline

kwryankrattiger:
@scottwittenburg quick, review this, CI passed!

haampie:
Sorry for reviewing this only so late, but regarding `+:` and `-:`, I would sort of like the opposite... so that it mimics

```python
>>> a = [1, 2, 3]
>>> a += [4]
>>> a
[1, 2, 3, 4]
```

So, `key+:` would be append.

scottwittenburg (approved):

PR #34276, opened 2022-12-02
Osu/mvapich 3.0 new package

MatthewLieber:
Creating a new package called mvapich and adding support for the MVAPICH 3a


var/spack/repos/builtin/packages/mvapich/package.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+class Mvapich(AutotoolsPackage):
+    """Mvapich is a High-Performance MPI Library for clusters with diverse
+    networks (InfiniBand, Omni-Path, Ethernet/iWARP, and RoCE) and computing
+    platforms (x86 (Intel and AMD), ARM and OpenPOWER)"""
+
+    homepage = "https://mvapich.cse.ohio-state.edu/userguide/userguide_spack/"
+    url = "https://mvapich.cse.ohio-state.edu/download/mvapich/mv2/mvapich2-3.0a.tar.gz"
+    list_url = "https://mvapich.cse.ohio-state.edu/downloads/"
```
alalazo, line 20: We already have an `mvapich2` package. Can you explain why this one downloads from:
```
https://mvapich.cse.ohio-state.edu/download/mvapich/mv2/mvapich2-3.0a.tar.gz
```
but creates a new package?

alalazo (changes requested):

var/spack/repos/builtin/packages/mvapich/package.py
natshineman: We are in the process of a pretty substantial overhaul to the MVAPICH Project. As part of that we are dropping the 2 from the name and moving to just MVAPICH, thus the new package. The beta version (which should be coming in the next month) should reflect that name change as well. 

We have also rewritten some pretty substantial sections of the code that have resulted in new configuration options that are not compatible with the older versions of the MVAPICH2 2.3.x series. However, while MVAPICH 3.0 is in the pre-release stage, we would like to continue supporting the older versions for their stability and performance on certain systems.

var/spack/repos/builtin/packages/mvapich/package.py
```diff
@@ -0,0 +1,300 @@
...
+    maintainers = ["natshineman", "harisubramoni", "ndcontini"]
+
+    executables = ["^mpiname$", "^mpichversion$"]
+
+    # Prefer the latest stable release
+    version("3.0a", "71f6593bfbfe9a9f6f5c750904461f007bf74bec479544e4da375b7d4a56b2ac")
+
+    provides("mpi")
+    provides("mpi@:3.1", when="@2.3:")
+    provides("mpi@:3.0", when="@2.1:")
```
alalazo: Can we simplify these directives? My understanding is that this new package will contain only MVAPICH >= 3.0 right?

var/spack/repos/builtin/packages/mvapich/package.py
```diff
@@ -0,0 +1,300 @@
...
+        "netmod",
+        description="Select the netmod to be enabled for this build."
+        "For IB/RoCE systems, use the ucx netmod, for interconnects supported "
+        "by libfabrics, use the ofi netmod. For more info, visit the "
+        "homepage url.",
+        default="ofi",
+        values=(
+            "ofi",
+            "ucx",
+            #       "ucr",
```
alalazo:
```suggestion
```

var/spack/repos/builtin/packages/mvapich/package.py
```diff
@@ -0,0 +1,300 @@
...
+    )
+
+    depends_on("findutils", type="build")
+    depends_on("bison", type="build")
+    depends_on("pkgconfig", type="build")
+    depends_on("zlib")
+    depends_on("libpciaccess", when=(sys.platform != "darwin"))
+    depends_on("libxml2")
+    depends_on("cuda", when="+cuda")
+    # depends_on("rdma-core", when="netmod=ucr")
```
alalazo:
```suggestion
```

var/spack/repos/builtin/packages/mvapich/package.py
```diff
@@ -0,0 +1,300 @@
...
+
+    @classmethod
+    def determine_version(cls, exe):
+        if exe.endswith("mpichversion"):
+            output = Executable(exe)(output=str, error=str)
+            match = re.search(r"^MVAPICH2 Version:\s*(\S+)", output)
+        elif exe.endswith("mpiname"):
+            output = Executable(exe)("-a", output=str, error=str)
+            match = re.search(r"^MVAPICH2 (\S+)", output)
+        return match.group(1) if match else None
```
alalazo, line 117: This I think needs to change, or the same executable will be discovered by both `mvapich2` and `mvapich`

var/spack/repos/builtin/packages/mvapich/package.py
```diff
@@ -0,0 +1,300 @@
...
+        opts = []
+        # From here on I can suppose that only one variant has been selected
+        if "netmod=ofi" in self.spec:
+            opts = ["--with-device=ch4:ofi"]
+        elif "netmod=ucx" in self.spec:
+            opts = ["--with-device=ch4:ucx"]
+        # elif "netmod=ucr" in self.spec:
+        #    opts = ["--with-device=ch4:ucr"]
+        # elif "netmod=mrail" in self.spec:
+        #    opts = ["--with-device=ch3:mrail", "--with-rdma=gen2", "--disable-mcast"]
```
alalazo:
```suggestion
```

alalazo (changes requested):
Thanks for the explanation. A few initial requests.

var/spack/repos/builtin/packages/mvapich/package.py
MatthewLieber: For the 3.0a version we have not yet changed this part.  We will change this for the next release.  For now we would like to keep this section in.

natshineman (approved):
Looks good to me, this should be ready to go.

MatthewLieber:
@alalazo Is there anything else you would like us to do on this pull request or can we get it merged in

alalazo (approved):

PR #34278, opened 2022-12-02
wannier90: gfortran-10 support and libs property

glennpj:


glennpj:
@spackbot run pipeline

alalazo (approved):

PR #34285, opened 2022-12-02
Support ASCI control

johnwparent:
Allow Spack to set Windows console associated with Spack process to handle ASCI control characters. This allows for colorization on Windows and removed the significant output cruft that appears in STDOUT when this functionality is not enabled. Behavior is enabled as soon as the tty module is loaded.

### DIFF
--- a/lib/spack/llnl/util/tty/color.py
+++ b/lib/spack/llnl/util/tty/color.py
@@ -107,6 +107,64 @@ color_when_values = {"always": True, "auto": None, "never": False}
 _force_color = None
 
 
+def try_enable_terminal_color_on_windows():
+    """Turns coloring in Windows terminal by enabling VTP in Windows consoles (CMD/PWSH/CONHOST)
+    Method based on the link below
+    https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#example-of-enabling-virtual-terminal-processing
+
+    Note: No-op on non windows platforms
+    """
+    if sys.platform == "win32":
+        import ctypes
+        import msvcrt
+        from ctypes import wintypes
+
+        try:
+            ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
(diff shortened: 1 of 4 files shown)


lib/spack/llnl/util/tty/color.py
```diff
@@ -107,6 +107,56 @@ def __init__(self, message):
...
+        dw_new_mode_request = ENABLE_VIRTUAL_TERMINAL_PROCESSING | DISABLE_NEWLINE_AUTO_RETURN
+        dw_new_mode = dw_new_mode_request | dw_orig_mode.value
+        kernel32.SetConsoleMode(con_handle, wintypes.DWORD(dw_new_mode))
+
+
+# On Windows console handling for ASCI/VT100 sequences is not
+# on by default. Turn on before we try to write to console
+# with color
+if sys.platform == "win32":
+    _enable_windows_vtp()
```
scheibelp: Would there be cases where this fails (and I assume one of the control commands in this function would raise an exception)? This could catch it, set `tty.color._force_color = False`, and then proceed (possibly also printing a debug message with the exception).

johnwparent: You're right, ostensibly any of these commands could fail, and I have no idea how they'll behave on Windows 11, so I'll add the error handling.

johnwparent: Addressed.

johnwparent:
@scheibelp I believe I've addressed all the changes requested during out meeting yesterday.

johnwparent:
@scheibelp RE: our last conversation about this pr, looks like I was wrong and we do have to enable this color functionality each time spack is executed, rather than just once in the startup script unfortunately. I've refactored to reflect that, and folded in some changes I made in the course of testing the discussed functionality.
This includes removing an external find we were running when loading the script for some reason.

scheibelp (approved):

PR #34286, opened 2022-12-02
perl: ensure use of spack compiler wrapper

glennpj:
Currently, the building of perl and perl extensions does not use the Spack compiler wrapper. This PR fixes that by 
1. specifying the `cc` to be the Spack compiler wrapper.
2. ensuring that the configuration files use the Spack compiler wrapper for building extensions in Spack.

Extensions built outside of Spack will use the actual compiler.

### DIFF
--- a/var/spack/repos/builtin/packages/perl/package.py
+++ b/var/spack/repos/builtin/packages/perl/package.py
@@ -290,6 +290,12 @@ class Perl(Package):  # Perl doesn't use Autotools, it should subclass Package
         else:
             make()
 
+    # The non-filtered config files will need to be used for dependents. This
+    # will have to be inserted in the @INC search space.
+    @property
+    def spack_config(self):
+        return join_path(prefix, "lib", self.version, "spack")
+
     @run_after("build")
     @on_package_attributes(run_tests=True)
     def build_test(self):
@@ -347,7 +353,7 @@ class Perl(Package):  # Perl doesn't use Autotools, it should subclass Package
         """Set PATH and PERL5LIB to include the extension and
         any other perl extensions it depends on,
         assuming they were installed with INSTALL_BASE defined."""
-        perl_lib_dirs = []
+        perl_lib_dirs = [self.spack_config]
         for d in dependent_spec.traverse(deptype=deptype):
             if d.package.extends(self.spec):
                 perl_lib_dirs.append(d.prefix.lib.perl5)
@@ -422,6 +428,10 @@ class Perl(Package):  # Perl doesn't use Autotools, it should subclass Package
             "-MModule::Loaded", "-MConfig", "-e", "print is_loaded(Config)", output=str
         )
 
+        # Make a copy of config_dot_pm
+        mkdir(self.spack_config)
+        copy(config_dot_pm, self.spack_config)
+
         with self.make_briefly_writable(config_dot_pm):
             match = "cc *=>.*"
             substitute = "cc => '{cc}',".format(cc=self.compiler.cc)
@@ -430,6 +440,8 @@ class Perl(Package):  # Perl doesn't use Autotools, it should subclass Package
         # And the path Config_heavy.pl
         d = os.path.dirname(config_dot_pm)
         config_heavy = join_path(d, "Config_heavy.pl")
+        # copy config_heavy
+        copy(config_heavy, self.spack_config)
 
         with self.make_briefly_writable(config_heavy):
             match = "^cc=.*"
--- a/var/spack/repos/builtin/packages/perl/package.py
+++ b/var/spack/repos/builtin/packages/perl/package.py
@@ -349,25 +349,25 @@ class Perl(Package):  # Perl doesn't use Autotools, it should subclass Package
                 maker()
                 maker("install")
 
-    def _setup_dependent_env(self, env, dependent_spec, deptype):
+    def setup_dependent_build_environment(self, env, dependent_spec):
         """Set PATH and PERL5LIB to include the extension and
         any other perl extensions it depends on,
         assuming they were installed with INSTALL_BASE defined."""
         perl_lib_dirs = [self.spack_config]
-        for d in dependent_spec.traverse(deptype=deptype):
+        for d in dependent_spec.traverse(deptype=("build", "run", "test")):
             if d.package.extends(self.spec):
                 perl_lib_dirs.append(d.prefix.lib.perl5)
         if perl_lib_dirs:
             perl_lib_path = ":".join(perl_lib_dirs)
-            env.prepend_path("PERL5LIB", perl_lib_path)
+            env.set("PERL5LIB", perl_lib_path)
         if sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


glennpj:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

tgamblin:
See also #34312, where I had to override a perl config.  One question I have here is whether this ends up changing the C compiler that you get if you query perl's configuration. That could affect building perl extensions *outside* spack.

In #34312, XS is asking Perl what to build with and it gets clang. I think outside of Spack, perl builds probably *should* use the non-wrapped compiler, as the wrappers won't even work.

glennpj:
@tgamblin Actually, the potential problem is the other way around. There is code in the perl recipe to set the non-wrapped compiler for use outside of Spack. The question is will building the extensions in Spack use the wrapper. I had to deal with that exact situation for R packages. I will mark this PR as "draft" and look into how perl extensions determine the build environment in Spack.

glennpj:
For #34312, since perl is not currently using the Spack compiler wrapper it would always return the non-wrapped compiler. However, I do not yet know what it would return if perl itself does use the wrapper.

tgamblin:
@glennpj in 34312, I was able to fix this by setting `PERL_EXT_CC`.  What I am not sure of is whether `PERL_EXT_CC` is a common convention or whether it's specific to the build system for `texinfo`.  If it is a common thing for perl extension builds, then maybe we could just have the `perl` package set it to the right thing in `setup_dependent_build_environment`

bernhardkaindl (dismissed):
Addressing @tgamblin's questions:

> See also https://github.com/spack/spack/pull/34312, where I had to override a perl config. One question I have here is whether this ends up changing the C compiler that you get if you query perl's configuration. That could affect building perl extensions outside spack.

No, this PR would not change anything for perl extensions because as @glennpj described, because after install, 
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/perl/package.py#L421intentionally modifies Config.pm to set `cc=self.compiler.cc` so that `$Config{cc}` will by set to self.compiler.cc, not the wrapper script (which would not work without of spack's build-env unfortunately), here is the docstring of it:
> Run after install so that Config.pm records the compiler that Spack built the package with.  If this isn't done, $Config{cc} will be set to Spack's cc wrapper script

But, `perl` `Configure` seems to prefer `gcc`, and the patches to build with the NVIDIA compiler even have to patch it:
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/perl/package.py#L141
```py
    # Enable builds with the NVIDIA compiler
    # The Configure script assumes some gcc specific behavior, and use
    # the mini Perl environment to bootstrap installation.
    patch("nvhpc-5.30.patch", when="@5.30.0:5.30 %nvhpc")
    patch("nvhpc-5.32.patch", when="@5.32.0:5.32 %nvhpc")
    patch("nvhpc-5.34.patch", when="@5.34.0:5.34 %nvhpc")
    conflicts(
        "@5.32.0:",
        when="%nvhpc@:20.11",
        msg="The NVIDIA compilers are incompatible with version 5.32 and later",
    )
```

@tgamblin wrote in the previous comment here:
> @glennpj in 34312, I was able to fix this by setting `PERL_EXT_CC`. What I am not sure of is whether `PERL_EXT_CC` is a common convention or whether it's specific to the build system for `texinfo`. If it is a common thing for perl extension builds, then maybe we could just have the `perl` package set it to the right thing in `setup_dependent_build_environment`

Answer: `PERL_EXT_CC` is specific to texinfo. Origin thread: https://lists.gnu.org/archive/html/bug-texinfo/2016-09/msg00035.html
> What comes to mind is to use another variable PERL_EXT_CC which would
be set in the same way as CC, assuming that is a valid trick to use
with libtool. configure.ac could check if PERL_EXT_CC is defined, and
if it is, use its value as the value of CC instead of using the
compiler that Perl says it needs.

As far as I can tell so far: Adding `"-Dcc=" + spack_cc,` should work and should not change anything for perl extensions.

glennpj:
The situation winds up being similar to R in that extensions need to be built with the original perl config files, not the filtered ones that are created post install. I uploaded a commit to copy the config files prior to filtering. Those config files are then used when building spack perl packages. Packages built outside of spack will use the non-wrapped compiler.

Building texinfo prior to this PR:
```console
$ spack install --fresh texinfo
...
checking for perl... /opt/packages/gpjohnsn/spack/opt/spack/linux-centos7-cascadelake/gcc-9.5.0/perl-5.36.0-4lb74yda7s34yosdorscx6qovne2p5x6/bin/perl
checking Perl configuration value cc... /opt/packages/gpjohnsn/spack/opt/spack/linux-centos7-haswell/gcc-4.8.5/gcc-9.5.0-6iebqrr7dxexwih2v65zc5wfhlxnqxwk/bin/gcc
...
```
Building texinfo after this PR:
```console
$ spack install --fresh texinfo
...
checking for perl... /opt/packages/gpjohnsn/spack/opt/spack/linux-centos7-cascadelake/gcc-9.5.0/perl-5.36.0-h5643n5do6bpmbrdtopzlnguo65ojava/bin/perl
checking Perl configuration value cc... /opt/packages/gpjohnsn/spack/lib/spack/env/gcc/gcc
...
```
@tgamblin Can you see if this PR fixes the texinfo build in lieu of #34312?

glennpj:
@spackbot run pipeline

adamjstewart:
#34312 was merged, is this PR still needed or were they trying to solve the same problem?

tgamblin:
@adamjstewart these are different problems.  #34312 is about `texinfo` failing to build perl extensions that require `gettext` because it wasn't including the spack compiler wrappers (or depending on `gettext`).

This PR is about the custody battle between `gettext` and `glibc` over `libintl`.  `glibc` distros include `libintl` in `libc` and `libintl.h` is on the system.  other distros/macOS/etc. do not.  `gettext` dependents need to know how to deal with that.  Things like this are the reason that nobody, absolutely nobody, can have nice things.

glennpj:
> #34312 was merged, is this PR still needed or were they trying to solve the same problem?

This PR is about having perl use the Spack compiler wrapper, as the title states, so is independent of #34312, except that it would make setting `PERL_EXT_CC` unnecessary. 

So, yes, this PR is still needed.

glennpj:
> This PR is about the custody battle between `gettext` and `glibc` over `libintl`.

No, that is not what this PR is about. This seems to have gotten conflated with #34312 more than it should have.

bernhardkaindl (dismissed):
@tldahlgren Hi Tamara, am short on time for a while, cannot take part in further review, but looks good to me (see my earlier review comment)

tldahlgren:
> @tldahlgren Hi Tamara, am short on time for a while, cannot take part in further review, but looks good to me (see my earlier review comment)

I will try to remember when I'm looking at PRs.  Thanks for the notice and comment.

glennpj:
@spackbot run pipeline

glennpj:
@spackbot run pipeline

PR #34288, opened 2022-12-02
libxau: add libs property

glennpj:


glennpj:
@spackbot run pipeline

var/spack/repos/builtin/packages/libxau/package.py
```diff
@@ -19,3 +19,7 @@ class Libxau(AutotoolsPackage, XorgPackage):
     depends_on("xproto")
     depends_on("pkgconfig", type="build")
     depends_on("util-macros", type="build")
+
+    @property
+    def libs(self):
+        return find_libraries("libXau", self.prefix, shared=True, recursive=True)
```
alalazo, line 25: Can we restrict the search to some subdirs (more worried about this becoming an external that anything else)?

glennpj: Will just using `self.prefix.lib` be okay?

alalazo: We can try that. `/usr/lib` is definitely better than `/usr` :grimacing:

glennpj: Is it possible that some systems wind up installing the library in `prefix.lib64` rather than `prefix.lib`?

bernhardkaindl: Yes, I think SUSE may use /usr/lib64 (not sure, don't have one at hand to check, but it was used many years ago),
Debian uses /usr/lib/x86_64-linux-gnu/libXau.so.6, so that should be ok.

glennpj: @alalazo Is what @bernhardkaindl proposes in #34390 a better solution?

alalazo: Are there systems where we might end up finding the library in both places? Does it make sense to search both and return the result? Also, I think this kind of search - when we got it correctly - will be useful to many other packages so we can probably extract at some point a reusable function out of that.

EDIT: Missed to write one thought. A possible issue I see with @bernhardkaindl approach is if we have a system where both `lib64` and `lib` exist, and the library we search if in the sub-folder we skip. That's why I am wondering if it makes sense to search both of them.

glennpj: I think searching both is probably safer. For externals though, depending on how the OS defines the layout it could be 

1. /lib
2. /usr/lib
3. /lib64
4. /usr/lib64

Of course, RHEL systems link the respective pairs.
```
ls -l /lib /lib64
lrwxrwxrwx. 1 root root 7 Dec 29  2020 /lib -> usr/lib
lrwxrwxrwx. 1 root root 9 Dec 29  2020 /lib64 -> usr/lib64
```
For RHEL, /usr/lib is 32-bit and /usr/lib64 is 64-bit. I am not sure if that is a universal pattern though.

Also, it is not clear to me how the `prefix.lib` syntax in a package is translated to finding libraries for externals. I would think those would default to the "standard" system library paths.

glennpj:
@alalazo In searching the Spack tree, all of the other cases where the `libs` property is set are done as I originally did it, searching from `prefix`. I understand that may not be the best way going forward but it is not clear to me how to best find the libraries without starting the search at the parent directory level. Could you please provide more guidance on how this should look?

alalazo:
@glennpj Fair point. I'll merge this as is, and try to think of something to solve the issue in a more general way.

alalazo (approved):

PR #34290, opened 2022-12-02
julia: add ncurses dependency

glennpj:
Julia links to libtinfo.


haampie:
It's not a library used by Julia, it's a dep of llvm. Can you check what llvm-config gives for linking against llvm?

glennpj:
> It's not a library used by Julia, it's a dep of llvm. Can you check what llvm-config gives for linking against llvm?

Indeed. Unfortunately, the build environment that I was using is now gone but I must have had something bad in my llvm installation.

PR #34295, opened 2022-12-03
py-mne: add 1.2.2, 1.2.3 and dependency packages

manuelakuhn:
https://github.com/mne-tools/mne-python/tree/v1.2.2
I did not update the `full` variant because it is not installable via pip anyway and the dependencies get kind of out of hand.

https://github.com/h5io/h5io/tree/h5io-0.1.7
https://gitlab.com/obob/pymatreader/-/tree/v0.0.30

### DIFF
--- a/var/spack/repos/builtin/packages/py-mne/package.py
+++ b/var/spack/repos/builtin/packages/py-mne/package.py
@@ -13,6 +13,7 @@ class PyMne(PythonPackage):
     pypi = "mne/mne-0.23.4.tar.gz"
     git = "https://github.com/mne-tools/mne-python.git"
 
+    version("1.2.3", sha256="b300dcee69ffb878cdbc5c02490e877df385c1b9482622e3aa1da06a604a6e37")
     version("1.2.2", sha256="d40743d6ca7ae3919a557166fd5fc4c00a9719e40c07346baad57964e15f02bb")
     version("0.23.4", sha256="ecace5caacf10961ebb74cc5e0ead4d4dbc55fed006eab1e644da144092354e9")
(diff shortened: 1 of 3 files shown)


manuelakuhn:
Right when I opened this PR the `py-mne` people published version 1.2.3 , so I added this as well :-)
https://github.com/mne-tools/mne-python/tree/v1.2.3

var/spack/repos/builtin/packages/py-mne/package.py
```diff
@@ -11,20 +11,39 @@ class PyMne(PythonPackage):
...
+    depends_on("py-matplotlib@3.1:", when="@1:", type=("build", "run"))  # *
+    depends_on("py-tqdm", when="@1:", type=("build", "run"))
+    depends_on("py-pooch@1.5:", when="@1:", type=("build", "run"))
+    depends_on("py-decorator", when="@1:", type=("build", "run"))
+    depends_on("py-packaging", when="@1:", type=("build", "run"))
+    depends_on("py-jinja2", when="@1:", type=("build", "run"))
+
+    with when("+hdf5"):
+        depends_on("py-h5io", when="@1:", type=("build", "run"))
+        depends_on("py-pymatreader", when="@1:", type=("build", "run"))
```
adamjstewart:
```suggestion
        depends_on("py-h5io", type=("build", "run"))
        depends_on("py-pymatreader", type=("build", "run"))
```
`when` isn't needed since `@:0 +hdf5` isn't possible

manuelakuhn: You're right. I have removed it. Thanks.

adamjstewart (approved):

Issue #34269, opened 2022-12-02
Installation issue: texinfo


omsai:
### Steps to reproduce the issue

```console
$ spack spec -I texinfo
Input spec
--------------------------------
 -   texinfo

Concretized
--------------------------------
 -   texinfo@7.0%apple-clang@13.1.6 build_system=autotools arch=darwin-ventura-skylake
[+]      ^perl@5.36.0%apple-clang@13.1.6+cpanm+shared+threads build_system=generic arch=darwin-ventura-skylake
[+]          ^berkeley-db@18.1.40%apple-clang@13.1.6+cxx~docs+stl build_system=autotools patches=26090f4,b231fcc arch=darwin-ventura-skylake
[+]          ^bzip2@1.0.8%apple-clang@13.1.6~debug~pic+shared build_system=generic arch=darwin-ventura-skylake
[+]              ^diffutils@3.8%apple-clang@13.1.6 build_system=autotools arch=darwin-ventura-skylake
[+]                  ^libiconv@1.16%apple-clang@13.1.6 build_system=autotools libs=shared,static arch=darwin-ventura-skylake
[+]          ^gdbm@1.23%apple-clang@13.1.6 build_system=autotools arch=darwin-ventura-skylake
[+]              ^readline@8.1.2%apple-clang@13.1.6 build_system=autotools arch=darwin-ventura-skylake
[+]                  ^ncurses@6.3%apple-clang@13.1.6~symlinks+termlib abi=none build_system=autotools arch=darwin-ventura-skylake
[+]                      ^pkgconf@1.8.0%apple-clang@13.1.6 build_system=autotools arch=darwin-ventura-skylake
[+]          ^zlib@1.2.13%apple-clang@13.1.6+optimize+pic+shared build_system=makefile arch=darwin-ventura-skylake
```

### Error message

<details><summary>Error message</summary>

```
==> Installing texinfo-7.0-z3gotxjpfmxmpxn3zpcvafamlapuu5sh
==> No binary for texinfo-7.0-z3gotxjpfmxmpxn3zpcvafamlapuu5sh found: installing from source
==> Fetching https://ftpmirror.gnu.org/texinfo/texinfo-7.0.tar.gz
==> No patches needed for texinfo
==> texinfo: Executing phase: 'autoreconf'
==> texinfo: Executing phase: 'configure'
==> texinfo: Executing phase: 'build'
==> Error: ProcessError: Command exited with status 2:
    'make' '-j16' 'V=1'

[...]

PR #34296, opened 2022-12-03
texinfo: add nls variant with CFLAGS fix (#34269)

omsai:
texinfo defaults to `--enable-nls` but spack provides no dependency to satisfy it.  Looking at other packages that use the `nls` variant in spack, GNU `binutils` and  GNU `gmake` also have this variant and `default=True`, so I use that default behavior here.

Yes, many systems have an nls provider like `gettext` installed on the system as part of the build environment.  However macOS does not, and there are some suggestions on stackoverflow for installing and force linking with homebrew.  But given that our spack documentation now suggests using spack instead of homebrew, texinfo should not rely on a homebrew library and should provide the dependency itself like the `binutils` and `gmake` packages.

Also version bumped to the new `texinfo@7.0.1` bug fix released just on November 30, 2022.

### DIFF
--- a/var/spack/repos/builtin/packages/texinfo/package.py
+++ b/var/spack/repos/builtin/packages/texinfo/package.py
@@ -23,6 +23,7 @@ class Texinfo(AutotoolsPackage, GNUMirrorPackage):
 
     tags = ["build-tools"]
 
+    version("7.0.1", sha256="53c01dd34c9bf116866d03aabd38c5cd9ee95ac10043d61aa33dec0d27825d8d")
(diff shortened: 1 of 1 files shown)


omsai:
Hmmm, I think passing `--with-libintl-prefix=` to `configure` to set CPPFLAGS, etc would be less heavy handed than appending to CFLAGS in `setup_build_environment` but I'm still troubleshooting why the former approach isn't addressing the original error.

omsai:
Close in lieu of #34312 which corrects for the deeper issue of the texinfo XS perl module not using spack's provided perl.

PR #34298, opened 2022-12-03
[WIP] ML CI stack: matrix of targets

adamjstewart:
Instead of having separate pipelines for every OS/target, this PR attempts to create a matrix of OS/targets on which we want to build our ML stack. If this works, the long-term plan is to add the cross product of:

* all ML packages
* on macOS and Linux
* on x86_64, aarch64, and ppc64le
* with the following variants:
  * +cuda ~rocm cuda_arch=...
  * +rocm ~cuda admgpu_target=...
  * ~cuda ~rocm

I'm going to start with a small subset of this for testing and expand once we get it working.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/ml-cpu/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/ml-cpu/spack.yaml
@@ -20,7 +20,7 @@ spack:
       variants: ~cuda~rocm
 
   definitions:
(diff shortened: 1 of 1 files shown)


adamjstewart:
This doesn't seem to work. The pipeline generation step runs on x86_64 and doesn't like it when I try to concretize specs for aarch64. Guess we'll have to keep all pipelines in separate files for now 😢

PR #34299, opened 2022-12-04
ML CI: Linux x86_64

adamjstewart:
This PR simply renames our current ML pipelines by adding `linux-x86_64`. I'm planning on adding many more platforms and targets, so this is necessary to differentiate.


adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

tgamblin (approved):

PR #34305, opened 2022-12-04
typing: move from comment type hints to Python 3.6 hints

tgamblin:
We've stopped supporting Python 2, and contributors are noticing that our CI no longer allows Python 2.7 comment type hints. They end up having to adapt them, but this adds extra unrelated work to PRs.

- [x] Move to 3.6 type hints across the entire code base

### DIFF
--- a/lib/spack/spack/test/llnl/util/tty/log.py
+++ b/lib/spack/spack/test/llnl/util/tty/log.py
@@ -11,10 +11,8 @@ import os
 import signal
 import sys
 import time
-from typing import TYPE_CHECKING, Optional  # novm
-
-if TYPE_CHECKING:
-    from types import ModuleType  # novm
+from types import ModuleType
+from typing import Optional
 
 import pytest
 
@@ -24,7 +22,7 @@ import llnl.util.tty.pty as pty
 
 from spack.util.executable import which
 
-termios = None  # type: Optional[ModuleType]
+termios: Optional[ModuleType] = None
 try:
     import termios as term_mod
 
--- a/lib/spack/spack/bootstrap.py
+++ b/lib/spack/spack/bootstrap.py
@@ -16,6 +16,7 @@ import re
 import sys
 import sysconfig
 import uuid
+from typing import List
 
 import archspec.cpu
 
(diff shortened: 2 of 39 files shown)


adamjstewart (dismissed):

tgamblin:
Sigh.  TIL about `typing.Match`.  Match objects have different types in 3.6 vs. 3.7+:

```console
Python 3.6.15 (default, Dec 21 2021, 08:03:21)
[GCC 10.2.1 20210110] on linux
>>> import re
>>> re.match("foo", "foo")
<_sre.SRE_Match object; span=(0, 3), match='foo'>
```

```console
Python 3.7.15 (default, Nov 15 2022, 10:33:03)
[GCC 10.2.1 20210110] on linux
>>> import re
>>> re.match("foo", "foo")
<re.Match object; span=(0, 3), match='foo'>
```

So you have to use `typing.Match[str]` in type hints.

adamjstewart (dismissed):

adamjstewart:
Are we running the style tests on multiple versions of Python? I think we only need to test on the latest version.

adamjstewart (dismissed):

lib/spack/spack/test/llnl/util/tty/log.py
```diff
@@ -11,10 +11,7 @@
 import signal
 import sys
 import time
-from typing import TYPE_CHECKING, Optional  # novm
-
-if TYPE_CHECKING:
-    from types import ModuleType  # novm
+from typing import ModuleType, Optional  # novm
```
adamjstewart: I think `ModuleType` comes from `types`, not `typing`

tgamblin:
@adamjstewart: 

> Are we running the style tests on multiple versions of Python? I think we only need to test on the latest version.

When you use the real type hints, things can fail in older versions even if we do not run the style checks.  e.g.:

```python
import re
from typing import List, Tuple

    def __init__(self, all_unquoted_flag_pairs: List[Tuple[re.Match, str]]):
        ...
```

`re.Match` is used in code that needs to work at class definition time for every version we support, but it does not exist in 3.6.  The code above should be:

```python
import re
from typing import List, Match, Tuple

    def __init__(self, all_unquoted_flag_pairs: List[Tuple[Match[str], str]]):
        ...
```

adamjstewart (approved):

adamjstewart:
Gotcha. There's also:
```python
from __future__ import annotations
```
but I've been avoiding that because of https://docs.python.org/3/library/__future__.html#id2.

tgamblin:
Sadly looks like we cannot have that until 3.7

PR #34306, opened 2022-12-04
[WIP] ML CI: macOS aarch64

adamjstewart:
Copied our existing CI from Linux to macOS.

I copied the design from the existing e4s-mac pipeline.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -886,6 +886,7 @@ ml-linux-x86_64-rocm-protected-build:
 .ml-darwin-aarch64-cpu:
   variables:
     SPACK_CI_STACK_NAME: ml-darwin-aarch64-cpu
+  allow_failure: True
 
 .ml-darwin-aarch64-cpu-generate:
   extends: .ml-darwin-aarch64-cpu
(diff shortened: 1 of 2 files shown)


adamjstewart:
<a href="https://imgflip.com/i/73557a"></a>

scottwittenburg:
The output from the generate job for this pipeline indicates that none of your specs found any matching mappings, which is why there was nothing to rebuild.  Somehow all the specs got generated for linux, ubuntu, and skylake.  Maybe you need a compiler/arch dimension in your matrix like exists in the `e4s-mac` stack?  I.e.:

```
  - arch:
    - '%apple-clang@13.1.6 target=m1'
 ```

As it stands, all the specs seemed to have been generated like this one:

```
py-torch-scatter@2.0.5%gcc@7.5.0~cuda build_system=python_pip arch=linux-ubuntu18.04-skylake_avx512
```

And since the only mapping is for `os=monterey`, nothing got generated.

adamjstewart:
It was still trying to generate the specs on Linux x86_64, hoping the last commit solves the issue based on the e4s-mac stack.

adamjstewart:
Last commit has a bug but I don't know what it is...

eugeneswalker:
Given that macOS relocation doesn't work, and that you correctly noted that `e4s-mac` stack fails every time as a result, and that you requested that that stack be disabled until the issue is resolved, why would we add another mac stack that will fail every time for the same reason? I'm confused.

adamjstewart:
I'm not planning on merging this until we get relocation working again. This was just to experiment and see how far I can get things.

adamjstewart:
Would it be possible to get nodes for older versions of macOS back when relocation did work?

adamjstewart:
This is outdated after the CI refactor. I'll create a new PR when I get a chance.

PR #34119, opened 2022-11-24
remove # novm comments

haampie:

PR #34308, opened 2022-12-05
vermin: remove all novm comments from code

tgamblin:
All of our vermin `#novm` annotations are for optional features introduced in early Python 3 versions. We no longer need any of them, as we only support Python 3.6+. If we start optionally using features from newer Pythons than 3.6, we will need more vermin annotations.

### DIFF
--- a/lib/spack/spack/build_systems/autotools.py
+++ b/lib/spack/spack/build_systems/autotools.py
@@ -7,7 +7,7 @@ import os
 import os.path
 import stat
 import subprocess
-from typing import List  # novm
+from typing import List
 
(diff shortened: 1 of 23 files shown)


adamjstewart:
Duplicate of #34119

tgamblin:
I added @haampie as a co-author.  This one isn't exactly a duplicate since it actually works (it's based on #34305).

alalazo (approved):

adamjstewart (approved):

alalazo:
@spackbot run pipeline

PR #34313, opened 2022-12-05
OpenCV: checksum for 4.5.5, make contrib optional

iarspider:


iarspider:
@spackbot fix style

adamjstewart (dismissed):
LGTM, but will give @glennpj a chance to review

glennpj:
I do not understand what this PR is attempting to accomplish with the `contrib` variant. What does the `contrib` variant do? I initially thought it was meant to hide the packages in the contrib directory but that is not what it does.
@iarspider Could you explain what the intention is?

iarspider:
> I do not understand what this PR is attempting to accomplish with the `contrib` variant. What does the `contrib` variant do? I initially thought it was meant to hide the packages in the contrib directory but that is not what it does. @iarspider Could you explain what the intention is?

`~contrib` blocks downloading and installing OpenCV contribs (unstable experimental modules).

glennpj:
> `~contrib` blocks downloading and installing OpenCV contribs (unstable experimental modules).

Okay; that is what I thought your goal was. However, just blocking the download of the resource does not accomplish that. In fact, it breaks the opencv package recipe. Each of those contrib packages is a variant in spack. If any of those are enabled in combination with `~contrib` the build will break. You will have to establish a dependency relationship with the `contrib` variant for each of the variants for the respective contrib packages.

iarspider:
> > `~contrib` blocks downloading and installing OpenCV contribs (unstable experimental modules).
> 
> Okay; that is what I thought your goal was. However, just blocking the download of the resource does not accomplish that. In fact, it breaks the opencv package recipe. Each of those contrib packages is a variant in spack. If any of those are enabled in combination with `~contrib` the build will break. You will have to establish a dependency relationship with the `contrib` variant for each of the variants for the respective contrib packages.

Added, thanks!

var/spack/repos/builtin/packages/opencv/package.py
```diff
@@ -394,6 +396,7 @@ class Opencv(CMakePackage, CudaPackage):
             default=False,
             description="Include opencv_{0} contrib module".format(mod),
         )
+        conflicts("+{0}".format(mod), when="~contrib")
```
glennpj: Since `contrib` is like a super variant, I think using a `when` makes more sense.
```
for mod in contrib_modules:
        variant(
            mod,
            when="+contrib",
            default=False,
            description="Include opencv_{0} contrib module".format(mod),
        )

File: var/spack/repos/builtin/packages/opencv/package.py [hunk 2]
```diff
@@ -782,6 +785,7 @@ class Opencv(CMakePackage, CudaPackage):
         description="Enable -ffast-math (not recommended for GCC 4.6.x)",
     )
     variant("nonfree", default=False, description="Enable non-free algorithms")
+    variant("contrib", default=True, description="Install extra OpenCV modules")
```
glennpj: Perhaps the default should be `False`? Also, change the description to:
"Enable OpenCV contrib modules"
as "extra" is not as specific.

glennpj (changes requested):

var/spack/repos/builtin/packages/opencv/package.py
iarspider: Good catch, thanks!

var/spack/repos/builtin/packages/opencv/package.py
iarspider: As a personal rule, whenever I introduce a variant to an existing package, I try to preserve the default behavior. Currently, the `contrib` modules are always installed, so I keep it that way.

var/spack/repos/builtin/packages/opencv/package.py
```diff
@@ -921,13 +925,15 @@ def determine_variants(cls, libs, version_str):
...
-                "OPENCV_EXTRA_MODULES_PATH",
-                join_path(self.stage.source_path, "opencv_contrib/modules"),
-            ),
             self.define("BUILD_opencv_core", "on"),
         ]
 
+        if self.spec.satisfies("+contrib"):
+            args += self.define(
+                "OPENCV_EXTRA_MODULES_PATH",
+                join_path(self.stage.source_path, "opencv_contrib/modules"),
```
adamjstewart:
```suggestion
                join_path(self.stage.source_path, "opencv_contrib", "modules"),
```

adamjstewart (approved):
LGTM as long as CI passes. I would also be fine with disabling this variant by default, I'm more interested in faster build times than backwards compatibility.

glennpj (approved):
Looks good to me.

Thanks!

iarspider:
@spackbot run pipeline

PR #34320, opened 2022-12-05
py-nbclassic: add 0.4.8 and new package py-notebook-shim

manuelakuhn:
https://github.com/jupyter/nbclassic/tree/v0.4.8
https://github.com/jupyter/notebook_shim/tree/v0.2.2


var/spack/repos/builtin/packages/py-nbclassic/package.py
```diff
@@ -12,13 +12,19 @@ class PyNbclassic(PythonPackage):
...
     depends_on("py-jupyter-packaging11", when="@0.3.3:", type="build")
     # depends_on('py-jupyter-packaging@0.9:1', when='@0.3.3:', type='build')
-    depends_on("py-jupyter-server@1.8:1", type=("build", "run"))
-    depends_on("py-notebook@:6", type=("build", "run"))
+    depends_on("py-babel", when="@0.4:", type="build")
+    depends_on("py-jupyter-server@1.17:", when="@0.4:", type=("build", "run"))
+    depends_on("py-jupyter-server@1.8:", when="@0.3.7", type=("build", "run"))
+    depends_on("py-jupyter-server@1.8:1", when="@:0.3.6", type=("build", "run"))
+    depends_on("py-notebook-shim@0.1:", when="@0.3.6:", type=("build", "run"))
+    depends_on("py-notebook@:6", when="@:0.3.7", type=("build", "run"))
```
adamjstewart, line 48: I think we're missing a lot of dependencies:
```python
    install_requires = [                                                                 
        'jinja2',                                                                        
        'tornado>=6.1',                                                                  
        # pyzmq>=17 is not technically necessary,                                        
        # but hopefully avoids incompatibilities with Tornado 5. April 2018              
        'pyzmq>=17',                                                                     
        'argon2-cffi',                                                                   
        'traitlets>=4.2.1',                                                              
        'jupyter_core>=4.6.1',                                                           
        'jupyter_client>=6.1.1',                                                         
        'ipython_genutils',                                                              
        'jupyter_server>=1.8',                                                           
        'nbformat',                                                                      
        'notebook_shim>=0.1.0',                                                          
        'nbconvert>=5',                                                                  
        'nest-asyncio>=1.5',                                                             
        'ipykernel', # bless IPython kernel for now                                      
        'Send2Trash>=1.8.0',                                                             
        'terminado>=0.8.3',                                                              
        'prometheus_client'                                                              
    ],                                  
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-nbclassic/package.py
manuelakuhn: Oh, I only saw the `install_require` in [`setup.cfg`](https://github.com/jupyter/nbclassic/blob/v0.4.8/setup.cfg#L29) and did not check if there was an additional, different one in `setup.py`. 
I will work on them.

But just as a question for understanding. Do you happen do know which one is used if `install_requires` is configured twice as it is here?

adamjstewart: Hmm, that is actually a good question. It's complicated by the fact that the build backend is `jupyter_packaging`, not `setuptools`. Maybe @pradyunsg would know the answer?

manuelakuhn: The dependencies are added now.

adamjstewart: In develop, `setup.cfg` has been updated with all the deps of `setup.py`, so I think `setup.py` is right.

adamjstewart (approved):

PR #34321, opened 2022-12-05
Avoid stat call in `llnl.util.symlink` on non-windows

haampie:
Avoid the redundant stat call on non-windows introduced in #33021

When creating environment views etc we really don't want to double the stat calls.


haampie:
Ah wait, sorry, I mistakenly thought `os.path.isdir` throws.

scheibelp (approved):

haampie:
@spackbot run pipeline

PR #34250, opened 2022-12-01
SWIG: new versions 4.1.x

sethrj:

PR #34323, opened 2022-12-05
swig: add versions 4.1.1 and 4.1.0

lorddavidiii:


sethrj:
Half-duplicate of #34250, and I just added 4.1.1 to that PR (and am adding the new fortran equivalents)

lorddavidiii:
ah sorry, didn't check if there is already a PR, so closing this in flavor of https://github.com/spack/spack/pull/34250

bernhardkaindl (changes requested):
On one of my hosts without pcre development installed, `spack install swig@4.1.1` fails with:
```
     34    checking whether to enable PCRE2 support... yes
     35    checking whether to use local PCRE2... no
     36    checking for a sed that does not truncate output... /usr/bin/sed
     37    checking for pcre2-config... no
  >> 38    configure: error: in `/tmp/bkaindl/spack-stage/spack-stage-swig-4.1.1-nz5qi6wmarcc6v36k4roi75zg5ik5gvt/spack-src/spack-build':
  >> 39    configure: error:
     40            Cannot find pcre2-config script from PCRE2 (Perl Compatible Regular Expressions)
     41            library package. This dependency is needed for configure to complete,
```
Please test-build without pcre development installed on the host and add a dependency on pcre, if pcre support is requested (could make it conditional on a pcre variant too)

sethrj:
@bernhardkaindl PCRE is already a dependency of SWIG... do you have pcre set up as an external?

bernhardkaindl:
No, I've found the cause: swig@4.1: needs pcre2, not pcre.

The depends_on needs to use pcre2 when 4.1:

sethrj:
Aha, quite right. Thanks @bernhardkaindl ! (See https://github.com/swig/swig/pull/2138 )

PR #34324, opened 2022-12-05
Use `urllib` handler for `s3://` and `gs://`, improve `url_exists` through HEAD requests

haampie:
- The idea of `urllib` is to add handlers for custom protocols [through `<protocol>_open(req)`](https://docs.python.org/3/library/urllib.request.html#basehandler-objects), so let's do that for `s3://` and `gs://`.

- Use `head_object` for `Request("s3://...", method="HEAD")` requests

- Use `HEAD` requests in `url_exists` to avoid expensive downloads from s3.

- Create a "custom" `urlopen` that instantiates an opener once and only once.

- Keep using the standard `HTTPRedirectHandler` for `HEAD` requests; in a previous iteration of this PR, I redirected HEAD requests as HEAD requests which led to issues when downloading from BitBucket.

Before:

```python
In [1]: import spack.util.web

In [2]: %timeit spack.util.web.url_exists("s3://spack-binaries/develop/aws-ahug-aarch64/build_cache/index.json")
526 ms ± 11 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit spack.util.web.url_exists("https://binaries.spack.io/develop/aws-ahug-aarch64/build_cache/index.json")
558 ms ± 9.17 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

After:

```python
In [1]: import spack.util.web

In [2]: %timeit spack.util.web.url_exists("s3://spack-binaries/develop/aws-ahug-aarch64/build_cache/index.json")
104 ms ± 2.89 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit spack.util.web.url_exists("https://binaries.spack.io/develop/aws-ahug-aarch64/build_cache/index.json")
146 ms ± 3.57 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/lib/spack/spack/util/web.py
+++ b/lib/spack/spack/util/web.py
@@ -344,12 +344,6 @@ def url_exists(url, curl=None):
     Simple Storage Service (`s3`) URLs; otherwise, the configured fetch
     method defined by `config:url_fetch_method` is used.
 
-    If the method is `curl`, it also uses the following configuration option:
-
-        * config:verify_ssl (str): Perform SSL verification
-
-    Otherwise, `urllib` will be used.
-
     Arguments:
         url (str): URL whose existence is being checked
         curl (spack.util.executable.Executable or None): (optional) curl
@@ -358,14 +352,11 @@ def url_exists(url, curl=None):
     Returns (bool): True if it exists; False otherwise.
     """
     tty.debug("Checking existence of {0}".format(url))
-    url = url_util.parse(url)
+    url_result = url_util.parse(url)
 
-    # Otherwise, use the configured fetch method
-    if spack.config.get("config:url_fetch_method") == "curl" and url.scheme in (
-        "http",
-        "https",
-        "ftp",
-    ):
+    # Use curl if configured to do so
+    use_curl = spack.config.get("config:url_fetch_method", "urllib") == "curl" and url_result.scheme not in ("gs", "s3")
+    if use_curl:
         curl_exe = _curl(curl)
         if not curl_exe:
             return False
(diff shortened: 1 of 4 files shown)


lib/spack/spack/util/web.py
```diff
@@ -35,6 +37,28 @@
...
+
+def _urlopen():
+    s3 = spack.s3_handler.UrllibS3Handler()
+    gcs = spack.gcs_handler.GCSHandler()
+
+    # One opener with HTTPS ssl enabled
+    with_ssl = build_opener(s3, gcs, HTTPSHandler(context=ssl.create_default_context()))
+
+    # One opener with HTTPS ssl disabled
+    without_ssl = build_opener(s3, gcs, HTTPSHandler(context=ssl._create_unverified_context()))
```
scottwittenburg, line 49: This is *really* nice, compared to how it was done in those blocks you removed!

scottwittenburg (approved):
This looks great, thanks @haampie!  One thing I didn't understand from the description.  You mentioned "Keep using standard HTTPRedirectHandler for HEAD requests", can you point me to where that is happening?  I didn't find it in here or in the code right off the bat.

scottwittenburg:
One thing about these web/s3 PRs is they don't change any hashes and never rebuild anything.  Additionally, I don't think the unit tests may do a very good job of testing the s3 bits.

haampie:
Yeah, some more tests around here would help (also it would force fewer globals...)

I honestly think Python has a bug in urllib (HEAD on redirect becomes GET), for which I also submitted a PR to cpython, but didn't get a response yet, and it causes some issues with bitbucket redirects to an aws bucket, so in this pr I'm not touching it

PR #34325, opened 2022-12-05
Stop checking for {s3://path}/index.html

haampie:
This is very strange and unusual.


alalazo (approved):
https://gist.github.com/zulhfreelancer/24f73015c5437281f3b98c3cb34ea225

PR #34330, opened 2022-12-05
feat(Hermes)!: add yaml-cpp dependency

hyoklee:
The 0.9.0-beta requires yaml-cpp for parsing the configuration file format in YAML.

P.S. I'm using https://www.conventionalcommits.org/en/v1.0.0/#specification for this commit message.


tldahlgren (approved):
LGTM.

Issue #34198, opened 2022-11-29
Installation issue: autoconf@2.71 fails with "config:install_tree:padded_length: 128"


benkirk:
### Steps to reproduce the issue

```console
$ spack spec -I autoconf@2.71
Input spec
--------------------------------
 -   autoconf@2.71

Concretized
--------------------------------
 -   autoconf@2.71%gcc@4.8.5 build_system=autotools arch=linux-centos7-haswell
[+]      ^m4@1.4.19%gcc@4.8.5+sigsegv build_system=autotools patches=9dc5fbd,bfdffa7 arch=linux-centos7-haswell
[+]          ^diffutils@3.8%gcc@4.8.5 build_system=autotools arch=linux-centos7-haswell
[+]              ^libiconv@1.16%gcc@4.8.5 build_system=autotools libs=shared,static arch=linux-centos7-haswell
[+]          ^libsigsegv@2.13%gcc@4.8.5 build_system=autotools arch=linux-centos7-haswell
[+]      ^perl@5.36.0%gcc@4.8.5+cpanm+shared+threads build_system=generic arch=linux-centos7-haswell
[+]          ^berkeley-db@18.1.40%gcc@4.8.5+cxx~docs+stl build_system=autotools patches=26090f4,b231fcc arch=linux-centos7-haswell
[+]          ^bzip2@1.0.8%gcc@4.8.5~debug~pic+shared build_system=generic arch=linux-centos7-haswell
[+]          ^gdbm@1.23%gcc@4.8.5 build_system=autotools arch=linux-centos7-haswell
[+]              ^readline@8.1.2%gcc@4.8.5 build_system=autotools arch=linux-centos7-haswell
[+]                  ^ncurses@6.3%gcc@4.8.5~symlinks+termlib abi=none build_system=autotools arch=linux-centos7-haswell
[+]                      ^pkgconf@1.8.0%gcc@4.8.5 build_system=autotools arch=linux-centos7-haswell
[+]          ^zlib@1.2.13%gcc@4.8.5+optimize+pic+shared build_system=makefile arch=linux-centos7-haswell

$ spack install autoconf@2.71
```

### Error message

<summary>/bin/sh: bin/autom4te: (padded path): bad interpreter: No such file or directory</summary>

<details><pre>
[...]

PR #34331, opened 2022-12-05
autoconf: fix for long perl shebangs with autoconf@2.70+

benkirk:
Fix for Issue  #34198 

In `autoconf` versions >=2.70, the  Perl shebang to be patched is `"^#! @PERL@"` instead of `"^#! @PERL@ -w"`


haampie:
Can't you just drop the `-w` so you don't need to branch on version?

benkirk:
@haampie I'm unclear on that.  The original `package.py`reverts the source to its pristine condition inside the `fliter_sbang` method.  I assume that's important, so then the branching is important to get it back to original form.

If that is not important, the `-w` could be dropped.

haampie (dismissed):

haampie (approved):

PR #34332, opened 2022-12-05
Prevent SameFileError when rpathing on Windows

johnwparent:
This behavior is only present when a user does not have developer mode enabled

### DIFF
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -2278,10 +2278,14 @@ class WindowsSimulatedRPath(object):
         """
         self._addl_rpaths = self._addl_rpaths | set(paths)
 
-    def _link(self, path, dest):
+    def _link(self, path, dest_dir):
+        """Perform link step of simulated rpathing, installing simlinks of file in path to the dest_dir
+        location. This method deliberately prevents the case where a path points to a file inside the dest_dir.
+        This is because it is both meaningless from an rpath perspective, and will cause an error when Developer
+        mode is not enabled"""
         file_name = os.path.basename(path)
-        dest_file = os.path.join(dest, file_name)
-        if os.path.exists(dest) and not dest_file == path:
+        dest_file = os.path.join(dest_dir, file_name)
+        if os.path.exists(dest_dir) and not dest_file == path:
             try:
                 symlink(path, dest_file)
             # For py2 compatibility, we have to catch the specific Windows error code
@@ -2295,7 +2299,7 @@ class WindowsSimulatedRPath(object):
                         "Linking library %s to %s failed, " % (path, dest_file) + "already linked."
                         if already_linked
                         else "library with name %s already exists at location %s."
-                        % (file_name, dest)
+                        % (file_name, dest_dir)
                     )
                     pass
                 else:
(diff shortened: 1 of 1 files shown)


johnwparent:
For users who have developer mode enabled and do not want to change that, this behavior and the fix introduced by this PR can be tested by manually causing the `win32_can_symlink` check in `llnl.util.symlink` to fail. This causes Spack on Windows to resort to fallback behavior, creating Windows junctions, which raises the `SameFileError`

lib/spack/llnl/util/filesystem.py
```diff
@@ -2281,7 +2281,7 @@ def add_rpath(self, *paths):
     def _link(self, path, dest):
```
scheibelp: A few requests:

* rename the `dest` parameter to `dest_dir`
* add a docstring mentioning that this skips cases where `path == dest_file` (to avoid errors when the user does not have developer mode enabled on Windows)
* (since this function is private to the class this is not essential but) it might be worth adding an explicit check that `dest_dir` is in fact a directory

scheibelp (approved):

lib/spack/llnl/util/filesystem.py
```diff
@@ -2278,10 +2278,17 @@ def add_rpath(self, *paths):
...
 
-    def _link(self, path, dest):
+    def _link(self, path, dest_dir):
+        """Perform link step of simulated rpathing, installing
+        simlinks of file in path to the dest_dir
+        location. This method deliberately prevents
+        the case where a path points to a file inside the dest_dir.
+        This is because it is both meaningless from an rpath
+        perspective, and will cause an error when Developer
+        mode is not enabled"""
```
alalazo, line 2288: Reading the code it seems the function now just returns if we try to link on the same location. I'm not clear though on what the docstring is saying... Can you help me clarifying how the docstring relates to the change?

lib/spack/llnl/util/filesystem.py
```diff
@@ -2278,10 +2278,17 @@ def add_rpath(self, *paths):
...
         file_name = os.path.basename(path)
-        dest_file = os.path.join(dest, file_name)
-        if os.path.exists(dest):
+        dest_file = os.path.join(dest_dir, file_name)
+        if os.path.exists(dest_dir) and not dest_file == path:
```
alalazo, line 2291: Below here I see `# for py2 compatibility`. We are not supporting Python 2 anymore, so I guess below we can just have:
```
symlink(path, dest_file)
```
? Or is there something I am missing?

lib/spack/llnl/util/filesystem.py
haampie: the docstring is written from a particular use case, it doesn't belong there

lib/spack/llnl/util/filesystem.py
johnwparent: Thanks for reminding me, we'll still need the try catch block, but now we can directly catch the Python class error as the exception rather than catching all OSErrors and checking the underlying Windows error.

lib/spack/llnl/util/filesystem.py
scheibelp: This is a private method used only once inside of a class with a specific use case. Normally I'd agree about where this type of check should occur, but IMO private methods have more leeway. If you disagree, would you want to add a TODO?

PR #34334, opened 2022-12-05
psrcat: fixed typo/undefined variable problem

downloadico:
replaced the reference to the undefined "bindir" variable with prefix.bin


tldahlgren (approved):
Great catch!

PR #34335, opened 2022-12-05
add perl dep to libxcrypt

luke-dt:
added a missing depencency that was preventing me from building `libxcrpyt` on my machine

https://github.com/besser82/libxcrypt#build-requirements-and-instructions


michaelkuhn:
Sorry, I totally missed this PR when I opened my own (#34343). I think a build dependency should be enough.

haampie:
currently everything is blocked due to broken hdf5-vol-async

luke-dt:
@michaelkuhn looks like we opened our PRs within hours of each other haha. You're right, a build dependency is probably all you need. I'll just close mine since yours already has that change.

haampie:
There's like 5 prs to fix this issue, but thanks to CI struggles none are merged :|

PR #34338, opened 2022-12-05
Add back missing dependencies after #33898

alalazo:
Cherry-picked from #34029

Modifications:
- [x] Add back a few packages that are needed by Spack to bootstrap dev dependencies when running under Python 3.6

### DIFF
--- a/var/spack/repos/builtin/packages/py-dataclasses/package.py
+++ b/var/spack/repos/builtin/packages/py-dataclasses/package.py
@@ -15,5 +15,5 @@ class PyDataclasses(PythonPackage):
     version("0.8", sha256="8479067f342acf957dc82ec415d355ab5edb7e7646b90dc6e2fd1d96ad084c97")
     version("0.7", sha256="494a6dcae3b8bcf80848eea2ef64c0cc5cd307ffc263e17cdf42f3e5420808e6")
 
-    depends_on("python@3.6.00:3.6", type=("build", "run"))
+    depends_on("python@3.6", type=("build", "run"))
     depends_on("py-setuptools", type="build")
--- a/var/spack/repos/builtin/packages/py-dataclasses/package.py
+++ b/var/spack/repos/builtin/packages/py-dataclasses/package.py
@@ -1,0 +1,19 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyDataclasses(PythonPackage):
+    """A backport of the dataclasses module for Python 3.6"""
+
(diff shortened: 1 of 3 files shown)


alalazo:
@spackbot fix style

tgamblin:
@alalazo can we ensure that bootstrapping of style tools gets tested for the *oldest* supported python release?  I guess in this case under rhel8 specifically?

alalazo:
If it's fine to wait, I'm already doing that `rhel8` check in https://github.com/spack/spack/pull/34029

tgamblin:
Works for me

var/spack/repos/builtin/packages/py-dataclasses/package.py
```diff
@@ -0,0 +1,19 @@
...
+class PyDataclasses(PythonPackage):
+    """A backport of the dataclasses module for Python 3.6"""
+
+    homepage = "https://github.com/ericvsmith/dataclasses"
+    pypi = "dataclasses/dataclasses-0.7.tar.gz"
+
+    version("0.8", sha256="8479067f342acf957dc82ec415d355ab5edb7e7646b90dc6e2fd1d96ad084c97")
+    version("0.7", sha256="494a6dcae3b8bcf80848eea2ef64c0cc5cd307ffc263e17cdf42f3e5420808e6")
+
+    depends_on("python@3.6.00:3.6", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("python@3.6", type=("build", "run"))
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -107,6 +107,13 @@ class Python(Package):
     version("3.7.1", sha256="36c1b81ac29d0f8341f727ef40864d99d8206897be96be73dc34d4739c9c9f06")
     version("3.7.0", sha256="85bb9feb6863e04fb1700b018d9d42d1caac178559ffa453d7e6a436e259fd0d")
 
+    # Python 3.6.15 has been added back only to allow bootstrapping Spack on Python 3.6
```
adamjstewart, line 110: Is this necessary? I would prefer to support external-only Python 3.6.

alalazo: Easiest way to not have audits to complain.

tgamblin: Is it just audits?  Could we add `#noaudit` for lines in the packages that trigger this?

alalazo: Yes, I was thinking about that. That means changes in the audit command too. I think the question is if we should do it here or we can do it in a separate PR.

tgamblin: Separate is fine w/me if it's fine with @adamjstewart

alalazo:
Closed, since #34029 has been merged. @adamjstewart I'll work on `#noaudit ` trying to find a way to also solve the macOS issue with CUDA.

PR #34340, opened 2022-12-05
mercurial/py-pybind11: print_string no longer exists

nicholas-sly:
Mercurial currently fails with python missing the `print_string` method.  This change fixes it by specifying use of `print()` making it python3-specific, but that's okay now.


becker33 (dismissed):

adamjstewart:
`py-pybind11` also uses this, can you fix that too?

adamjstewart (approved):

becker33 (approved):

PR #34343, opened 2022-12-06
libxcrypt: add perl build dependency

michaelkuhn:
Without it, the build fails with:
```
Can't locate open.pm in @INC (you may need to install the open module) (@INC contains: ...) at ./build-aux/scripts/expand-selected-hashes line 20.
BEGIN failed--compilation aborted at ./build-aux/scripts/expand-selected-hashes line 20.
configure: error: bad value 'all' for --enable-hashes
```


alalazo (approved):

bernhardkaindl:
Also have a look at #34183:

It appears that `perl-file-open` is only needed for `perl@4.3.30:` and also take a look at my comment/requested change for it:

https://github.com/spack/spack/pull/34183#pullrequestreview-1205325463

bernhardkaindl (changes requested):
It's actually `perl-open` that is missing and `depends_on("perl")` would be fooled by `spack external find perl` when the external perl does not provide `perl-open`, which is the case on all RHEL8-clones and newer.

BTW, `perl-open` is only needed for `perl@4.4.30:`. Please see #34183, specifically the comment https://github.com/spack/spack/pull/34183#pullrequestreview-1209301036

haampie:

PR #34346, opened 2022-12-06
htop: adding version 3.2.1

dpzmick:


var/spack/repos/builtin/packages/htop/package.py
```diff
@@ -10,9 +10,10 @@ class Htop(AutotoolsPackage):
     """htop is an interactive text-mode process viewer for Unix systems."""
 
     homepage = "https://github.com/htop-dev/htop/"
-    url = "https://github.com/htop-dev/htop/archive/refs/tags/3.1.1.tar.gz"
+    url = "https://github.com/htop-dev/htop/archive/refs/tags/3.2.1.tar.gz"
```
sethrj, line 13: It's not necessary to update the URL: it's just used to detect the formula for all versions.

sethrj (approved):
Checksum is good.

var/spack/repos/builtin/packages/htop/package.py
dpzmick: awesome, thanks for the note!

PR #34347, opened 2022-12-06
remove legcay yaml from buildcache fetch

haampie:
This was deprecated in 0.19.0, so it can be removed.

Reduces the number of 404s by a third on miss


alalazo (approved):
:100:

PR #31347, opened 2022-06-28
Remove redundent spec fetches for binary installation

scottwittenburg:
installer: remove redundant fetch of spec files from mirror
    
Since binary_distribution::download_tarball() is going to iterate
over the mirrors fetching spec files anyway (as it will likely need
to verify the signature on a clear-signed spec), there should be no
need to first get_mirrors_for_spec() before calling download_tarball().
The result should be that for each spec we install from binary, we
fetch the associated spec file once instead of twice.

PR #34350, opened 2022-12-06
binary cache: do not create separate `spec.json.sig` files

haampie:
There's no reason to generate a separate `spec.json.sig` file, instead we
can peek into the `spec.json` file and see if it is clearsigned json.

This PR also changes the iteration order: outer loop iterates over
extension types, inner loop over mirrors, since in time we'd expect
*.json to always be correct, whether signed or unsigned, so on a
(binary) cache hit, at most 1 request per mirror for the specfile is
necessary.

In the future we can stop using `spec.json.sig` files in general to avoid
404s. (In the near future we could also just add a config option to stop
looking for these `spec.json.sig` files.)

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -6,6 +6,7 @@
 import codecs
 import collections
 import hashlib
+import io
 import json
 import multiprocessing.pool
 import os
@@ -876,6 +877,8 @@ def select_signing_key(key=None):
 
 
 def sign_specfile(key, force, specfile_path):
+    """Create a clearsigned <specfile_path>.sig file in the same directory
+    and return its path."""
     signed_specfile_path = "%s.sig" % specfile_path
     if os.path.exists(signed_specfile_path):
         if force:
@@ -885,6 +888,43 @@ def sign_specfile(key, force, specfile_path):
 
     key = select_signing_key(key)
     spack.util.gpg.sign(key, specfile_path, signed_specfile_path, clearsign=True)
+    return signed_specfile_path
+
+
+def _load_clearsigned_json(stream):
+    # Skip the PGP header
+    stream.readline()
+    stream.readline()
+    json = stream.read()
+    footer_index = json.rfind("-----BEGIN PGP SIGNATURE-----")
+    if footer_index == -1 or not json[footer_index - 1].isspace():
+        raise ValueError("Could not find PGP signature in clearsigned json file.")
+    return sjson.load(json[:footer_index])
+
+
+def _load_possibly_clearsigned_json(stream):
+    if _is_clearsigned_stream(stream):
+        return _load_clearsigned_json(stream)
+    return sjson.load(stream)
+
+
+def _is_clearsigned_stream(stream):
+    curr = stream.tell()
+    header = stream.read(34)
+    stream.seek(curr)
+    return header == "-----BEGIN PGP SIGNED MESSAGE-----"
+
+
+def is_clearsigned_file(path):
+    with open(path, "r") as f:
+        return _is_clearsigned_stream(f)
+
+
+def load_possibly_clearsigned_json(s):
+    """Deserialize JSON from a string or stream s, removing any clearsign
+    header/footer."""
+    s = io.StringIO(s) if isinstance(s, str) else s
+    return _load_possibly_clearsigned_json(s)
 
 
 def _read_specs_and_push_index(file_list, read_method, cache_prefix, db, temp_dir, concurrency):
@@ -909,11 +949,7 @@ def _read_specs_and_push_index(file_list, read_method, cache_prefix, db, temp_di
 
         if spec_file_contents:
             # Need full spec.json name or this gets confused with index.json.
-            if spec_url.endswith(".json.sig"):
-                specfile_json = Spec.extract_json_from_clearsig(spec_file_contents)
-                return Spec.from_dict(specfile_json)
-            if spec_url.endswith(".json"):
-                return Spec.from_json(spec_file_contents)
+            return Spec.from_dict(load_possibly_clearsigned_json(spec_file_contents))
 
     tp = multiprocessing.pool.ThreadPool(processes=concurrency)
     try:
@@ -1221,23 +1257,17 @@ def _build_tarball(
     spec_file = spack.store.layout.spec_file_path(spec)
     specfile_name = tarball_name(spec, ".spec.json")
     specfile_path = os.path.realpath(os.path.join(cache_prefix, specfile_name))
-    signed_specfile_path = "{0}.sig".format(specfile_path)
 
     remote_specfile_path = url_util.join(
         outdir, os.path.relpath(specfile_path, os.path.realpath(tmpdir))
     )
-    remote_signed_specfile_path = "{0}.sig".format(remote_specfile_path)
 
     # If force and exists, overwrite. Otherwise raise exception on collision.
-    if force:
-        if web_util.url_exists(remote_specfile_path):
+    if web_util.url_exists(remote_specfile_path):
+        if force:
             web_util.remove_url(remote_specfile_path)
-        if web_util.url_exists(remote_signed_specfile_path):
-            web_util.remove_url(remote_signed_specfile_path)
-    elif web_util.url_exists(remote_specfile_path) or web_util.url_exists(
-        remote_signed_specfile_path
-    ):
-        raise NoOverwriteException(url_util.format(remote_specfile_path))
+        else:
+            raise NoOverwriteException(url_util.format(remote_specfile_path))
 
     # make a copy of the install directory to work with
     workdir = os.path.join(tmpdir, os.path.basename(spec.prefix))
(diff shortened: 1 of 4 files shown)


haampie:
Ping @blue42u :upside_down_face:

lib/spack/spack/binary_distribution.py
```diff
@@ -1503,68 +1529,52 @@ def download_tarball(spec, unsigned=False, mirrors_for_spec=None):
...
+            # If it is a clearsign file, we must verify it (unless disabled)
+            should_verify = not unsigned and is_clearsigned_file(specfile_path)
+            if should_verify and not try_verify(specfile_path):
+                verification_failure = True
+                tty.warn("Failed to verify: {0}".format(specfile_url))
+                specfile_stage.destroy()
+                continue
+
+            # In case the spec.json is not clearsigned, it means it's a legacy
+            # format, where the signature is in the tarball with binaries. Verification
```
scottwittenburg:
```suggestion
            # format, where either the signature is in the tarball with binaries, or 
            # the package is unsigned. Verification
```

Isn't it true?  Or maybe I'm missing something.  Look at the code path taken below in this case, I think it's handled correctly, so this is maybe just a comment nitpick.

lib/spack/spack/binary_distribution.py
```diff
@@ -2014,58 +2020,35 @@ def install_single_spec(spec, allow_root=False, unsigned=False, force=False):
...
-        )
+            utf8_stream = codecs.getreader("utf-8")(fs)
+            fetched_spec = Spec.from_dict(load_possibly_clearsigned_json(utf8_stream))
+
+            # read the spec from the build cache file. All specs in build caches
+            # are concrete (as they are built) so we need to mark this spec
+            # concrete on read-in.
+            fetched_spec._mark_concrete()
+
+            return [
```
scottwittenburg: You no longer want this to return information about all the mirrors where the spec may live, but just return when we find the first?  

This may be fine, currently the relationship between knowing where binaries may live (the results of this method) and where we download them from (`download_tarball()`) is sort of tangled and unsatisfactory.  Maybe it's not useful here to know all the mirrors where the spec lives.  

If that's the case, I guess we no longer need the `found_specs` variable, and can just return `[]` at the end if we never found what we were looking for.

lib/spack/spack/binary_distribution.py
```diff
@@ -876,6 +877,8 @@ def select_signing_key(key=None):
 
 
 def sign_specfile(key, force, specfile_path):
+    """Create a clearsigned <specfile_path>.sig file in the same directory
+    and return its path."""
```
scottwittenburg, line 881: Based on the PR description, I was thinking we would stop writing `.json.sig` files now (instead clearsigning the `.json` files), but continue to support reading them.  So there's something I'm not getting about this part of the description:

> There's no reason to generate a separate spec.json.sig file, instead we
> can peek into the spec.json file and see if it is clearsigned json.

scottwittenburg (commented):
Thanks @haampie this seems like a good idea :+1:  I just have a couple questions, but otherwise like the change.

lib/spack/spack/binary_distribution.py
haampie: I think it's best to have a (temporary) signed file. `gpg` can't update a file in-place, so the alternative would be `mv spec.json.sig spec.json`, but at that point I thought we could just as well upload from `<local path>/spec.json.sig` to `<remote url>/spec.json`.

lib/spack/spack/binary_distribution.py
haampie: Ah sorry, I didn't mean to change the behavior of other functions and forgot that I changed it here. Should I just revert things here? At the end of the day what I care about is that `spack install` in CI jobs is reasonably fast; this function is not called during spack install, right?

lib/spack/spack/binary_distribution.py
haampie: Yeah, you're right 👍

lib/spack/spack/binary_distribution.py
haampie: Ah uh, so we do a request to check for existence and then later another request for validation. It would be nice if all this was just one request

haampie: Would it be OK if I write it more in a tell-dont-ask fashion?

We have:

```python
matches = binary_distribution.get_mirrors_for_spec(pkg.spec)
return _process_binary_cache_tarball(
    pkg, pkg.spec, explicit, unsigned, mirrors_for_spec=matches, timer=timer
)
```

Is there any compelling reason to collect `matches` instead of running `_process_binary_cache_tarball` on all mirrors? It doubles the number of requests for the matches, right?

haampie: I'm very slow to understand the logic, but do I get this right: the original idea of `get_mirrors_for_spec` is to do an _offline_ search (local version of remote database), so that remote mirrors are ordered in such a way that it's likely to get a direct hit?

It seems however that with Jonathon's changes this search is no longer offline, since a full copy of the index gets downloaded if there's no local hits. And previously it was also not fully offline, but then instead only the spec searched for was located directly.

If this is the original idea, I'm happy with that, but it would be great if querying the local/offline cache caused a remote query on failure.

haampie:
@scottwittenburg how about this: I'll restore the `try_direct_fetch` bits in this PR so it's just about going from spec.json.sig to spec.json.

In a separate PR I'll make `get_mirrors_for_spec` offline. 

Offline in the sense that it should only query the local (possibly outdated) info we have about the remote mirror.

As I understand it, that's used as an optimization to improve the order in which mirrors are considered. That's great. What's not great: if we locally have no data about the remote mirror, it's strictly more work to update the local cache than it is to directly fetch in default order. So, obviously this optimization only works given local data, and if there's no local data, the best you can do is fall back to going over the mirror in default order.

lib/spack/spack/binary_distribution.py
scottwittenburg: > Is there any compelling reason to collect matches instead of running _process_binary_cache_tarball on all mirrors? It doubles the number of requests for the matches, right?

Yes, this was my assessment when looking into how the installer uses these methods a few months ago.  So I wrote #31347 to do pretty much what I think you just suggested.  But then I wasn't sure how/whether it would work with Jonathan's change to always fetch the index.

scottwittenburg: > the original idea of get_mirrors_for_spec is to do an offline search (local version of remote database), so that remote mirrors are ordered in such a way that it's likely to get a direct hit?

Yes, that was part of the idea.  Also, during gitlab ci pipeline generation, it gave us the option (`--check-index-only`) to fetch a few indices and then do offline queries for each concrete spec in the environment (to check if it needed to be rebuilt or not).  The alternative was to directly querying each mirror for the concrete specs in the environment, which takes a lot longer.

scottwittenburg: > If this is the original idea, I'm happy with that, but it would be great if querying the local/offline cache caused a remote query on failure.

Did you mean it would be great if querying the local/offline cache *did not cause* a remote query on failure?  In other words, you wish we would not fetch the index on local query miss, and not try to fetch the spec directly in that case either?

lib/spack/spack/binary_distribution.py
scottwittenburg: Got it, I missed that you were uploading from `.json.sig` locally to `.json` remotely.  Thanks for clarifying.

blue42u:
>  ... if we locally have no data about the remote mirror, it's strictly more work to update the local cache than it is to directly fetch in default order.

This statement is false when there are multiple buildcaches configured and all of them are small (< a few MB), which is exactly our use case in HPCToolkit's CI. But it is true that the worst-case scenario is worse if you update the local cache compared to using a default order, so not opposed.

lib/spack/spack/binary_distribution.py
haampie: > you wish we would not fetch the index on local query miss,

Yes! I'm doing that in #34359 now.

haampie:
This should be ready for review. Only concern is that it's not forward compatible (old spack will fail to deal with clearsigned spec.json file without sig extension), but as far as I know, Spack is never forward compatible.

scottwittenburg:
It occurs to me that if we don't want `.spec.json.sig` files anymore, we need to change some things about how we do signing with the reputational key as well.

@kotfic Can you chime in on whether you think this will be problematic?  Looking around, I think the notary job script needs changed [here](https://github.com/spack/spack-infrastructure/blob/53219550a71a3fbb0e9fa359b42fe1bddc111b8f/images/notary/sign.sh#L71) so we no longer write back `.sig` files.  Is there any problem getting GPG to sign "in place"?  I.e. just replace the current `.spec.json` with the re-signed version?

I'll make a PR to fix all the stacks signing jobs where they refer to the extension, e.g. [here](https://github.com/spack/spack/blob/f5bff1674563156cef6e56e374da989460f37b9a/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml#L460).  Another case where it will be great to have #34272.

I think due to the way this is implemented, it *could* be merged before we fix how the notary job signs things with the reputational key, as spack can still handle `.spec.json.sig` for awhile longer.  But I think it means more requests will get made as we look first for `.spec.json` after this.  However, it might make sense to coordinate these changes to go in around the same time.

haampie:
Closing this in favor of #34490 to get better compatibility backward/forward compat

PR #34352, opened 2022-12-06
py-torch: fix build with gcc@12:

michaelkuhn:
Applies all the way back to 1.0.0.


adamjstewart (approved):

bernhardkaindl (approved):
Builds on x86-linux with `%gcc@12.2.0`

PR #34353, opened 2022-12-06
[WIP] ML CI stack: matrix of targets

adamjstewart:
Instead of having separate pipelines for every OS/target, this PR attempts to create a matrix of OS/targets on which we want to build our ML stack. If this works, the long-term plan is to add the cross product of:

* all ML packages
* on macOS and Linux
* on x86_64, aarch64, and ppc64le
* with the following variants:
  * +cuda ~rocm cuda_arch=...
  * +rocm ~cuda admgpu_target=...
  * ~cuda ~rocm

I'm going to start with a small subset of this for testing and expand once we get it working.

Reboot of #34298


scottwittenburg:
My intuition on this one is that it will be difficult if it's even possible at this point, and it might be better to wait to combine things like this until we have #34272.  That will allow way more control over application of runner mappings and require a lot less yaml.  But @kwryankrattiger can chime in if he agrees or not.

adamjstewart:
Thanks! That PR looks really interesting. I'm not sure if it will solve my problem since I'll still need to list every package we want to test in dozens of files, but it will at least simplify the file and remove all this cruft I don't understand. I'll keep an eye on that PR and try to push forward with my alternative PRs in the meantime.

scottwittenburg:
> I'll still need to list every package we want to test in dozens of files

Actually, maybe some features added by @blue42u could help here.  The ability to choose the matching behavior `merge` in your mappings (the default is `first`) allows you to apply generic attributes with some mappings, and then add more specific attributes in subsequent mappings.  If you use that feature, then I understand that another feature he added, `removeAttributes` is useful to remove things in later mappings.  For reference, I think those features were both added in #32290 which has some examples.

adamjstewart:
Going to close this for now, but hoping to revisit this someday.

PR #34137, opened 2022-11-25
pip/wheel/setuptools: extend PythonExtension

adamjstewart:
This allows us to run post-install import tests and other niceties of PythonPackage while still avoiding circular deps.

PR #34355, opened 2022-12-06
Add back depends_on directives needed to bootstrap on Python 3.6

alalazo:
Reverts spack/spack#34137

@adamjstewart Trying to check if this broke bootstrapping when merged to develop

### DIFF
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -41,7 +41,7 @@ class PythonExtension(spack.package_base.PackageBase):
         in the source tarball directory. If the module names are incorrectly
         detected, this property can be overridden by the package.
 
-        Return:
+        Returns:
             list: list of strings of module names
         """
         modules = []
--- a/lib/spack/spack/build_systems/python.py
+++ b/lib/spack/spack/build_systems/python.py
@@ -41,7 +41,7 @@ class PythonExtension(spack.package_base.PackageBase):
         in the source tarball directory. If the module names are incorrectly
         detected, this property can be overridden by the package.
 
-        Returns:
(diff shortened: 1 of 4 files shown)


adamjstewart:
@tgamblin when is that merge queue coming? 😅

alalazo:
Arrgh, I need to push also some silly change to core to run what I want to check :laughing:

alalazo:
@adamjstewart Ok, this is working as expected. Wondering what is the best way to check that bootstrapping on Python 3.6 continues working when there's a change in a Python related package. :thinking:

alalazo:
Also, do you want me to bring back in the `PythonExtension` modification? I think it's just dropping the `depends_on` that broke bootstrapping.

lib/spack/spack/build_systems/python.py
```diff
@@ -41,7 +41,7 @@ def import_modules(self):
         in the source tarball directory. If the module names are incorrectly
         detected, this property can be overridden by the package.
 
-        Returns:
+        Return:
```
adamjstewart: Returns is correct

adamjstewart:
Yes, let's keep PythonExtension.

For testing, can we test bootstrapping for both the latest and oldest versions of Python?

lib/spack/spack/build_systems/python.py
alalazo:
```suggestion
        Returns:
```

alalazo: Yes, I just needed something to trigger unit tests

alalazo:
@adamjstewart If this is fine with you I'll merge this hotfix for develop.

PR #31391, opened 2022-06-30
Bug/fix credentials s3 buildcache update

josephsnyder:
A set of fixes for the S3 connection information found during the usage of a mirror and the buildcache command.  See https://spackpm.slack.com/archives/C01UZDL35C4/p1655420904901919 for initial issue and further discussion.

PR #34359, opened 2022-12-06
Don't check remote mirrors to determine fetch order from mirrors

haampie:
When installing binary tarballs, Spack has to download from its
binary mirrors.

Sometimes Spack has cache available for these mirrors.

That cache helps to order mirrors to increase the likelihood of
getting a direct hit.

However, currently, when Spack can't find a spec in any local cache
of mirrors, it's very dumb:

1. A while ago it used to query each mirror to see if it had a spec,
   and use that information to order the mirror again, only to go
   about and do exactly a part of what it just did: fetch the spec
   from that mirror :confused: 
2. Recently, it was changed to download a _full_ index.json, which
   can be multiple dozens of MBs of data and may take a minute to
   process thanks to the blazing fast performance you get with
   Python.

In a typical use case of concretizing with reuse, the full index.json
is already available, and it likely that the local cache gives a perfect
mirror ordering on install. (There's typically no need to update any
caches).

However, in the use case of Gitlab CI, the build jobs _don't have cache_,
and it would be smart to just do direct fetches instead of all the
redundant work of (1) and/or (2).

Also, direct fetches from mirrors will soon :tm: be fast enough to
prefer these direct fetches over the excruciating slowness of
index.json files.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -266,10 +266,7 @@ class BinaryCacheIndex(object):
                 None, just assumes all configured mirrors.
         """
         if find_hash not in self._mirrors_for_spec:
-            # Not found in the cached index, pull the latest from the server.
-            self.update(with_cooldown=True)
-        if find_hash not in self._mirrors_for_spec:
-            return None
+            return []
         results = self._mirrors_for_spec[find_hash]
         if not mirrors_to_check:
             return results
@@ -2079,8 +2076,8 @@ def get_mirrors_for_spec(spec=None, mirrors_to_check=None, index_only=False):
         spec (spack.spec.Spec): The spec to look for in binary mirrors
         mirrors_to_check (dict): Optionally override the configured mirrors
             with the mirrors in this dictionary.
-        index_only (bool): Do not attempt direct fetching of ``spec.json``
-            files from remote mirrors, only consider the indices.
+        index_only (bool): When ``index_only`` is set to ``True``, only the local
+            cache is checked, no requests are made.
 
     Return:
         A list of objects, each containing a ``mirror_url`` and ``spec`` key
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -48,6 +48,7 @@ import spack.binary_distribution as binary_distribution
 import spack.compilers
 import spack.error
 import spack.hooks
+import spack.mirror
 import spack.package_base
 import spack.package_prefs as prefs
 import spack.repo
@@ -419,18 +420,24 @@ def _try_install_from_binary_cache(pkg, explicit, unsigned=False, timer=timer.NU
             otherwise, ``False``
         timer (Timer):
     """
+    # Early exit if no mirrors are configured.
+    if not spack.mirror.MirrorCollection():
+        return False
+
     pkg_id = package_id(pkg)
     tty.debug("Searching for binary cache of {0}".format(pkg_id))
 
     timer.start("search")
-    matches = binary_distribution.get_mirrors_for_spec(pkg.spec)
+    matches = binary_distribution.get_mirrors_for_spec(pkg.spec, index_only=True)
     timer.stop("search")
 
-    if not matches:
-        return False
-
     return _process_binary_cache_tarball(
-        pkg, pkg.spec, explicit, unsigned, mirrors_for_spec=matches, timer=timer
+        pkg,
+        pkg.spec,
+        explicit,
+        unsigned,
+        mirrors_for_spec=matches,
+        timer=timer,
     )
 
 


lib/spack/spack/binary_distribution.py
```diff
@@ -266,10 +267,7 @@ def find_by_hash(self, find_hash, mirrors_to_check=None):
                 None, just assumes all configured mirrors.
         """
         if find_hash not in self._mirrors_for_spec:
-            # Not found in the cached index, pull the latest from the server.
-            self.update(with_cooldown=True)
```
scottwittenburg, line 270: I'm happier without this, but I think this is where @blue42u and I may disagree.

lib/spack/spack/installer.py
```diff
@@ -419,18 +421,29 @@ def _try_install_from_binary_cache(pkg, explicit, unsigned=False, timer=timer.NU
...
     pkg_id = package_id(pkg)
     tty.debug("Searching for binary cache of {0}".format(pkg_id))
 
     timer.start("search")
-    matches = binary_distribution.get_mirrors_for_spec(pkg.spec)
+    # Note: this is an optimization to order mirrors _before_ fetching, therefore it's
+    # offline only.
+    prioritized_fetch_urls = [
+        hit["mirror_url"]
+        for hit in binary_distribution.get_mirrors_for_spec(pkg.spec, index_only=True)
```
scottwittenburg: Trying to think this bit through.  Seems we'll continue to check this out of a sense of optimism, but under the hood make no requests in case of a miss.  Is there any chance we'll get a hit without having arranged to call `bindist.update()` in the current process first?  At one point I thought we should just remove this (see #31347), but I guess if it's cheap, it doesn't hurt anything.

scottwittenburg (dismissed):
This makes sense to me, I just had a couple questions.

One more question too: Are you planning to rebase this on #34350 after that is merged, or does this supercede that PR?  It seems to have some of the same changes.

haampie:
> Are you planning to rebase this on https://github.com/spack/spack/pull/34350 after that is merged, or does this supercede that PR? It seems to have some of the same changes.

Yeah, it was on top of it, cause I expected #34350 to be merged soon. But that seems not so likely now, given the issues mentioned there.

So, I've rebased on develop and dropped unrelated changes now.

lib/spack/spack/installer.py
haampie: It can still be useful in the "normal" use case of `spack concretize` + `spack install`, since concretization with reuse will fetch the index.json, so a few seconds later during install the cache can be expected to be up to date and will likely lead to direct binary cache hits.

scottwittenburg (approved):

lib/spack/spack/binary_distribution.py
blue42u: I am uncomfortable without this fallback. It confuses me that `spack clean -m` has an impact on the install speed for concrete environments/specs from then on, and even more confusing is that `spack spec` (a "read-only" operation) can *also* have an impact.

Thanks to @haampie for making PRs to lessen the impact, but I'm fairly certain this will give me headaches when tuning our CI.

I'd be happier with this change if Spack exposed an `apt update && apt install` pattern, where the buildcache outright wouldn't be used unless you ran `spack buildcache update` or some such. But that breaks how `--reuse` works with buildcaches currently (always up-to-date for maximum reuse).

So... :disappointed: :shrug:

haampie: Can you check if by now it's even a significant overhead? This pr rebased on develop 

> but that breaks how reuse works

I don't think it does, it's a different code path, and it ensures the local cache is up to date

blue42u: I was going to take a minute today to run this PR in our CI, but it fell over immediately with an `Access Denied` error from `spack ci generate`. Haven't investigated yet, but @haampie does https://github.com/spack/spack/pull/34324 have something to do with this? Last I saw this failure was from before https://github.com/spack/spack/pull/31391 was merged.

I'll try again once things have settled a bit, it's hard to replicate our exact CI environment and there are other tasks (and Spack PRs!) that need my attention.

> I don't think it does, it's a different code path, and it ensures the local cache is up to date

My point is that folks now assume `spack spec --reuse zlib` will always give the latest `zlib` in their enabled buildcaches. If the `index.json` is only fetched by explicit request (e.g. `spack buildcache fetch`), that behavior will be wildly different.

haampie: > My point is that folks now assume spack spec --reuse zlib will always give the latest zlib in their enabled buildcaches. If the index.json is only fetched by explicit request (e.g. spack buildcache fetch), that behavior will be wildly different.

That behavior is not going to change. `spack spec` requires an up-to-date index.json, so it will fetch it. `spack install` will not, but given that people concretize with reuse and then fetch, the cache is typically up to date.

#34324 was reverted (again) :weary: does `head_object` give you access denied or so? I think https://github.com/spack/spack/commit/7e054cb7fc55c35d50480f04552536eed48f5e21 is more likely to be the issue.

haampie:
I just took the liberty to merge this now, since we're seeing CI spend a lot of time on fetching dependencies.

PR #34360, opened 2022-12-06
BinaryIndexCache: Work in parallel

blue42u:
The `BinaryIndexCache` did not behave properly in parallel or when the mirrors config change. A single persistent "contents" JSON is loaded in memory and persisted when mirrors are updated, however due to how the lock was managed multiple parallel Spack invocations may fetch the `index.json` in parallel or remove each other's cache entries. While these data races are innocuous, they can be very expensive on time and network.

This PR redesigns the cache structure and simplifies the code, losing about 270 lines overall. Instead of a single "contents" JSON that is held in memory, all operations now maintain the hash (`index.json.hash`) and data (`index.json`) as separate cache keys. The cache keys used are based on a hash of the mirror's URL, as opposed to alternative approaches:
 - Mirror names are not necessarily consistent across parallel or subsequent Spack invocations, one Spack may refer to the same mirror URL via a different name or vice versa.
 - Content hashes change over time, requiring the removal of out-of-date cache keys during normal operations. Changing keys significantly complicates the parallelism for little real gain.

The error handling has been reworked, network failure to fetch indices from all configured mirrors is **no longer an error**. Users that use the default https://mirror.spack.io mirror will see no difference in behavior, this mirror does not have an index and so the `FetchCacheError` is never raised. These errors can be added back as warnings once a mirror configuration for non-buildcache mirrors can be implemented.

Incorporates https://github.com/spack/spack/pull/33781. /cc @scottwittenburg @haampie

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -15,6 +15,7 @@ import tarfile
 import tempfile
 import time
 import traceback
+import typing
 import warnings
 from contextlib import closing
 from urllib.error import HTTPError, URLError
@@ -52,28 +53,6 @@ _build_cache_relative_path = "build_cache"
 _build_cache_keys_relative_path = "_pgp"
 
 
-class FetchCacheError(Exception):
-    """Error thrown when fetching the cache failed, usually a composite error list."""
-
-    def __init__(self, errors):
-        if not isinstance(errors, list):
-            raise TypeError("Expected a list of errors")
-        self.errors = errors
-        if len(errors) > 1:
-            msg = "        Error {0}: {1}: {2}"
-            self.message = "Multiple errors during fetching:\n"
-            self.message += "\n".join(
-                (
-                    msg.format(i + 1, err.__class__.__name__, str(err))
-                    for (i, err) in enumerate(errors)
-                )
-            )
-        else:
-            err = errors[0]
-            self.message = "{0}: {1}".format(err.__class__.__name__, str(err))
-        super(FetchCacheError, self).__init__(self.message)
-
-
 class ListMirrorSpecsError(spack.error.SpackError):
     """Raised when unable to retrieve list of specs from the mirror"""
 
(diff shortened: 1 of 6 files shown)


haampie:
I appreciate that this PR removes code, but it fails to reduce complexity. It's a lot of machinery to be able to answer the query `[concrete spec] in [remote]`. I don't think you should be extending the Database class for this. If you want to do trivial queries without verifying the integrity and everything of the database, then don't instantiate a Database instance.

lib/spack/spack/binary_distribution.py
```diff
@@ -100,436 +79,334 @@ class BinaryCacheIndex(object):
...
+        # mapping from mirror urls to either:
+        #    The set[str] of DAG hashes available in that mirror, or
+        #    The dict[str, Spec] of concrete specs available indexed by DAG hash.
+        self._mirror_specs = {}
 
-    def _init_local_index_cache(self):
-        if not self._index_file_cache:
-            self._index_file_cache = file_cache.FileCache(self._index_cache_root)
+    @classmethod
+    def _cache_keys_for(cls, mirror_url: str, discriminator: int = 0):
```
haampie: What's wrong with the mirror name, it's unique.

blue42u: Multiple Spack environments may refer to the same mirror URL via different names, or to different URLs with the same name. They aren't unique enough, so this code works by URL. (It did before too, that trait hasn't changed.)

blue42u:
> I don't think you should be extending the Database class for this. If you want to do trivial queries without verifying the integrity and everything of the database, then don't instantiate a Database instance.

For clarification, the only reason the `Database` class is even involved here is because the `index.json` is the same format (as in, same code paths) as the Database's JSON. IMHO they should be separate code paths that share some common functions to de/serialize a list of Specs to JSON, but I'm not changing that in this PR.

blue42u:
Updated to separate this code from `Database`, now the `index.json` creation and parsing is all handled within `binary_distribution.py`. The format itself is unchanged, an extra test is added to ensure the `Database` code-path used prior is able to read the new implementation's output.

haampie:
@blue42u as a general comment, if you want your changes to land, consider making smaller PRs. This is just too many changes at once.

lib/spack/spack/binary_distribution.py
```diff
@@ -100,438 +80,368 @@ class BinaryCacheIndex(object):
...
+                with self._index_file_cache.write_transaction(data_key) as (_, data_f):
+                    data_f.write(data)
+
+                # Update the metadata to match our new reality
+                mdata = metadata.setdefault(mirror_url, {})
+                mdata["key"] = data_key
+                mdata["hash"] = compute_hash(data)
+                sjson.dump(metadata, newmeta_f)
+        except (URLError, web_util.SpackWebError) as e:
+            tty.warn("Ignoring error while fetching binary cache indices", e)
```
haampie: This is worse error handling than it used to be, I'm afraid I wouldn't accept it like this. Earlier the idea was to collect errors and show them only when they were an actual issue.

lib/spack/spack/binary_distribution.py
```diff
@@ -926,41 +834,60 @@ def _fetch_spec_from_mirror(spec_url):
         tp.terminate()
         tp.join()
 
-    for fetched_spec in fetched_specs:
-        db.add(fetched_spec, None)
-        db.mark(fetched_spec, "in_buildcache", True)
+    # Recursively lower the Specs into spec-dict (+ record) form
```
haampie: Why are you reinventing the database format here? Better to leave it as it was so we don't have to do double maintenance.

lib/spack/spack/binary_distribution.py
```diff
@@ -100,438 +80,368 @@ class BinaryCacheIndex(object):
...
+                # We will catch the exception this raises outside the write_transaction
+                _, _, fs = web_util.read_from_url(index_fetch_url)
+                data = codecs.getreader("utf-8")(fs).read()
+
+                # Decide on the cache key for the data blob, if one hasn't been chosen
+                if mirror_url in metadata:
+                    data_key = metadata[mirror_url]["key"]
+                else:
+                    data_key = newdata_key
+                    if len(metadata) > 0:
```
haampie: I would really like to keep this simple. It's OK to just use content hash as the key. To me as a rule of thumb when there's more than a few nested levels of control flow it's just a sign that it should be simplified.

haampie:
I'd suggest to split this up into multiple PRs so they can be reviewed (and possibly reverted) more easily.

For this PR it would be good to stick to that idea of obtaining the mapping of spec -> mirror list faster, and get rid of all unrelated changes like changes to error handling and cache keys etc.

blue42u:
I'll boil this PR down to the original goal: fixing BinaryIndexCache to work in parallel with other Spack invocations, that requires a refactor of the BinaryIndexCache class. The other features will almost certainly require that refactor to land first due to merge conflicts.

blue42u:
Code reduced and some parts redesigned and simplified based on comments. OP updated with new description of the patch.

lib/spack/spack/binary_distribution.py
blue42u: Note that *any* error here is not an actual issue, since we do not require that the buildcache have an `index.json` (direct fetches are allowed). The new implementation absorbs all errors silently, which makes no user-facing difference if a non-buildcache mirror (e.g. https://mirror.spack.io) was enabled.

lib/spack/spack/binary_distribution.py
blue42u: The intention was to separate the `index.json` code paths here from the `Database` implementation, without changing the format itself (for compatibility with older Spack). This would allow the `index.json` format to evolve separately from the far more complete `Database`.

Delaying this change for a later PR.

lib/spack/spack/binary_distribution.py
blue42u: Content hashes make this complex, the new implementation uses URL hashes which are far simpler. Quoting part of the updated OP:
> The cache keys used are based on a hash of the mirror's URL, as opposed to alternative approaches:
> - Mirror names are not necessarily consistent across parallel or subsequent Spack invocations, one Spack may refer to the same mirror URL via a different name or vice versa.
> - Content hashes change over time, requiring the removal of out-of-date cache keys during normal operations. Changing keys significantly complicates the parallelism for little real gain.

haampie:
Thanks for making the PR a bit smaller. It's still touching a few unrelated things imo, so again, if you want to get things merged fast, try to do one change per PR.

This PR cannot be merged until error handling is restored. In some cases (e.g. bootstrapping) an index is _required_, and errors should be as specific as possible (could not verify ssl, https://xyz/index.json 404'd, ...). Anything that would lead to generic errors is a step back.

> While these data races are innocuous

There was no race? Or am I missing something?

> Content hashes change over time, requiring the removal of out-of-date cache keys during normal operations. Changing keys significantly complicates the parallelism for little real gain.

I agree it's easier not to deal with removal and leave that do `spack clean -m`.

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
-        if not web_util.url_exists(index_fetch_url):
-            # A binary mirror is not required to have an index, so avoid
-            # raising FetchCacheError in that case.
-            return False
+        hash_url = url_util.join(mirror_url, _build_cache_relative_path, filename + ".hash")
+        data_url = url_util.join(mirror_url, _build_cache_relative_path, filename)
 
-        old_cache_key = None
-        fetched_hash = None
+        hash_key, data_key = self._cache_keys_for(mirror_url, filename)
```
haampie, line 259: Can you just drop `filename`, it's static.

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
 
-        errors = []
+        # Fast path: we already have the file and it is up-to-date
+        previous_hash = None
+        if self._file_cache.init_entry(hash_key):
+            with self._file_cache.read_transaction(hash_key) as hash_f:
+                try:
+                    _, _, fs = web_util.read_from_url(hash_url)
+                    expected_hash = codecs.getreader("utf-8")(fs).read()
+                except (URLError, web_util.SpackWebError):
```
haampie:
```suggestion
                except web_util.SpackWebError:
```

We catch the former and throw the latter in web_util.

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
+                    pass  # No expected hash upstream, ignore
```
haampie, line 269: keep track of errors please

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
+                # Stash the final updated file in the cache at the appropriate key
+                with self._file_cache.write_transaction(data_key) as (_, data_f):
+                    data_f.write(data)
+
+                # Update the hash to match our new reality
+                newhash_f.write(compute_hash(data))
+        except (URLError, web_util.SpackWebError):
+            # We had an error while attempting to fetch the data. These indices
+            # do not need to exist for proper operation, so ignore the error.
+            return None
```
haampie, line 301: same

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
+                        return False
+
+                # At this point: we need to fetch the file. No two ways about it.
+                # Do the fetch now, so that if it fails we don't leave any cruft behind.
+                # We will catch the exception this raises outside the write_transaction
+                _, _, fs = web_util.read_from_url(data_url)
+                data = codecs.getreader("utf-8")(fs).read()
+
+                # Stash the final updated file in the cache at the appropriate key
+                with self._file_cache.write_transaction(data_key) as (_, data_f):
```
haampie, line 293: If you only write when you have a write transaction on the metadata file, you can drop this lock altogether.

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
+                    data_f.write(data)
+
+                # Update the hash to match our new reality
+                newhash_f.write(compute_hash(data))
```
haampie, line 297: We probably wanna stick to a serialized version of `{"index_hash": compute_hash(data)}`, since it's extensible. We sure want to be able to store extra fields in the future (e.g. etags)

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
+            fetch_url = m.fetch_url
 
-    def _fetch_and_cache_index(self, mirror_url, expect_hash=None):
-        """Fetch a buildcache index file from a remote mirror and cache it.
+            # If we're in the cooldown period, don't re-fetch and assume the
+            # current state is good enough.
+            if (
+                with_cooldown
+                and ttl > 0
+                and fetch_url in self._last_fetch_times
```
haampie, line 222: Now that you're changing the format anyways, can you please get rid of this process local `_last_fetch_times` and make it part of the metadata file? Simplest is to just use `os.path.getmtime` if available. Right now this optimization is not very useful.

Could also be follow-up, but now that you've changed the metadata format, it's easy. I never understood a ttl of 10 minutes that applies only to a single running process :confused:

lib/spack/spack/binary_distribution.py
```diff
@@ -97,441 +76,230 @@ class BinaryCacheIndex(object):
...
-                # We expected a hash, we fetched a hash, and they were not the
-                # same.  If we end up fetching an index successfully and
-                # replacing our entry for this mirror, we should clean up the
-                # existing cache file
-                if mirror_url in self._local_index_cache:
-                    existing_entry = self._local_index_cache[mirror_url]
-                    old_cache_key = existing_entry["index_path"]
-
-        tty.debug("Fetching index from {0}".format(index_fetch_url))
+            with self._file_cache.write_transaction(hash_key) as (oldhash_f, newhash_f):
```
haampie, line 277: If you do the (slow) fetch in a write transaction, you have to think about lock timeouts. Did you test this? Right now it's 2 min, maybe that's ok.

haampie (changes requested):
See above

lib/spack/spack/binary_distribution.py
blue42u: I'll do this in a follow-up PR. I'll persist the fetch's `time.time()` in the metadata, that should be consistent (enough) across machines to implement the cooldown across a shared-filesystem cluster or a CI pipeline (in caches).

lib/spack/spack/binary_distribution.py
blue42u: `_load_specs_for` holds only a read lock on the data file, thus to prevent races the same write lock has to be held while writing.

I guess I could hold a read lock on the metadata in `_load_specs_for`, but IMHO this way is simpler and easier to reason about. `_load_specs_for` only needs a data file lock since it *only* cares about the data file. `_update_for()` is the only function that reads the metadata and thus is the only one using the metadata file lock.

blue42u:
Closing this in favor of #34686

PR #34362, opened 2022-12-06
py-numba: add 0.56.4

manuelakuhn:
https://github.com/numba/numba/tree/0.56.4

### DIFF
--- a/var/spack/repos/builtin/packages/py-numba/package.py
+++ b/var/spack/repos/builtin/packages/py-numba/package.py
@@ -11,7 +11,9 @@ class PyNumba(PythonPackage):
 
     homepage = "https://numba.pydata.org/"
     pypi = "numba/numba-0.35.0.tar.gz"
+    git = "https://github.com/numba/numba.git"
 
+    version("0.56.4", sha256="32d9fef412c81483d7efe0ceb6cf4d3310fde8b624a9cecca00f790573ac96ee")
     version("0.56.0", sha256="87a647dd4b8fce389869ff71f117732de9a519fe07663d9a02d75724eb8e244d")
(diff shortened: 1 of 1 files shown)


adamjstewart (approved):

tldahlgren (approved):
Confirmed sha256.

var/spack/repos/builtin/packages/py-numba/package.py
```diff
@@ -21,29 +23,30 @@ class PyNumba(PythonPackage):
...
+    depends_on("python@3.6:3.8", when="@0.52", type=("build", "run"))
     # set upper bound for python the same as newer release
-    depends_on("python@3.6:3.8", type=("build", "run"), when="@0.48:0.51")
-    depends_on("python@3.3:3.7", type=("build", "run"), when="@0.40.1:0.47")
-    depends_on("py-numpy@1.18:1.22", type=("build", "run"), when="@0.55.2:")
-    depends_on("py-numpy@1.18:1.21", type=("build", "run"), when="@0.55.0:0.55.1")
-    depends_on("py-numpy@1.17:1.20", type=("build", "run"), when="@0.54")
+    depends_on("python@3.6:3.8", when="@0.48:0.51", type=("build", "run"))
+    depends_on("python@3.3:3.7", when="@0.40.1:0.47", type=("build", "run"))
+    depends_on("py-numpy@1.18:1.23", when="@0.56.1:", type=("build", "run"))
```
dylex, line 33: It seems like this line is not doing anything, since the following like has a stricter dependency range for a wider set of versions. Should py-numpy@1.23 be allowed with py-numba@0.56.4?

adamjstewart: Ah yes, the line after this should be changed to `when="@0.55.2:0.56.0"`. Can you submit a PR to fix this?

PR #34363, opened 2022-12-07
error messages: wip condition chaining

becker33:
This is work to create chains of causation for error messages.

This is currently very WIP.

The basic idea is that there are `condition` facts in the solver, and if we track how one condition causes attributes which cause another condition, we can track every fact in the solve back to a source, and this could provide traceback-like error messages for ASP.


becker33:
Closing in favor of #34500, as the performance of this method was roughly 50x slower than develop, and #34500 shows performance roughly 1.1x slower than develop.

PR #34368, opened 2022-12-07
directives: depends_on should not admit anonymous specs

tgamblin:
Writing a long dependency like:

```python
     depends_on(
         "llvm"
         "targets=amdgpu,bpf,nvptx,webassembly"
         "version_suffix=jl +link_llvm_dylib ~internal_unwind"
     )
```

when it should be formatted like this:

```python
     depends_on(
         "llvm"
         " targets=amdgpu,bpf,nvptx,webassembly"
         " version_suffix=jl +link_llvm_dylib ~internal_unwind"
     )
```

can cause really subtle errors. Specifically, you'll get something like this in the package sanity tests:

```
    AttributeError: 'NoneType' object has no attribute 'rpartition'
```

because Spack happily constructs a class that has a dependency with name `None`.

We can catch this earlier by banning anonymous dependency specs directly in `depends_on()`.  This causes the package itself to fail to parse, and emits a much better error message:

```
==> Error: Invalid dependency specification in package 'julia':
    llvmtargets=amdgpu,bpf,nvptx,webassemblyversion_suffix=jl +link_llvm_dylib ~internal_unwind
```

- [x] Add a check in `_depends_on`
- [x] Add a test

### DIFF
--- a/lib/spack/spack/directives.py
+++ b/lib/spack/spack/directives.py
@@ -361,6 +361,8 @@ def _depends_on(pkg, spec, when=None, type=default_deptype, patches=None):
         return
 
     dep_spec = spack.spec.Spec(spec)
+    if not dep_spec.name:
+        raise DependencyError("Invalid dependency specification in package '%s':" % pkg.name, spec)
     if pkg.name == dep_spec.name:
         raise CircularReferenceError("Package '%s' cannot depend on itself." % pkg.name)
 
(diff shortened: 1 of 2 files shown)


lib/spack/spack/directives.py
```diff
@@ -361,6 +361,8 @@ def _depends_on(pkg, spec, when=None, type=default_deptype, patches=None):
         return
 
     dep_spec = spack.spec.Spec(spec)
+    if not dep_spec.name:
+        raise DependencyError("Invalid dependency specification in package '%s':" % pkg.name, spec)
```
alalazo, line 365: OT (not strictly related to the review). When do we need to add a new exception type vs. reuse another one? 

Right now we have a lot of exceptions, which sometimes are used only once or twice. I have always been wondering if we should add a new exception for each new error, or we can use a more generic one (in this case simply `DirectiveError`).

I think a benefit of having just a few is that we might collect them in a common place so people can look at a single file to know which errors can be raised.

alalazo (approved):
This LGTM. I restarted the tests on macOS since the failure seemed spurious.

PR #34372, opened 2022-12-07
s3: cache client instance

haampie:
Use a cache of the form

```
(mirror name, fetch/push) -> Client(...)
```

for S3 connections, to avoid the overhead of creating it on every request.

With this PR:

```python
In [1]: import spack.util.web

In [2]: %timeit spack.util.web.url_exists("s3://spack-binaries/does/not/exist")
105 ms ± 7.54 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Before:

```python
In [1]: import spack.util.web

In [2]: %timeit spack.util.web.url_exists("s3://spack-binaries/does/not/exist")
2.2 s ± 238 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```


lib/spack/spack/test/web.py
```diff
@@ -246,14 +247,24 @@ def get_object(self, Bucket=None, Key=None):
...
 def test_gather_s3_information(monkeypatch, capfd):
-    mock_connection_data = {
-        "access_token": "AAAAAAA",
-        "profile": "SPacKDeV",
-        "access_pair": ("SPA", "CK"),
-        "endpoint_url": "https://127.0.0.1:8888",
-    }
+    mirror = spack.mirror.Mirror.from_dict(
+        {
+            "fetch": {
```
scottwittenburg, line 252: Something I'm confused about is why you changed from an arbitrary dictionary of connection data (`mock_connection_data`) to this thing that is actually a `Mirror`.  For one thing, my understanding was that the `endpoint_url` was optional (and only used/needed for S3 buckets that don't live in AWS S3, but rather `minio` or similar).  If this were a real S3 mirror config, would it have another url?  I.e. the `s3:\\` url?

Also, I'm surprised it's valid to specify all three auth methods simultaneously, I would have thought you'd want to specify only one of `access_token`, `profile`, or `access_pair`.  Maybe specifying all three just helps with testing though?

lib/spack/spack/s3_handler.py
```diff
@@ -44,7 +44,7 @@ def __getattr__(self, key):
 
 def _s3_open(url):
     parsed = url_util.parse(url)
-    s3 = s3_util.create_s3_session(parsed, connection=s3_util.get_mirror_connection(parsed))
```
scottwittenburg, line 47: Was it a bug that this didn't pass `url_type="fetch"` previously?

lib/spack/spack/util/web.py
```diff
@@ -377,9 +375,7 @@ def url_exists(url, curl=None):
     # Check if Amazon Simple Storage Service (S3) .. urllib-based fetch
     if url_result.scheme == "s3":
         # Check for URL-specific connection information
-        s3 = s3_util.create_s3_session(
-            url_result, connection=s3_util.get_mirror_connection(url_result)
```
scottwittenburg, line 381: Is this another case where `url_type="fetch"` should have been passed before?

lib/spack/spack/util/s3.py
```diff
@@ -4,83 +4,115 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import os
 import urllib.parse
+from typing import Any, Dict, Tuple
 
 import spack
+import spack.config
 import spack.util.url as url_util
 
+#: Map (mirror name, method) tuples to s3 client instances.
```
scottwittenburg, line 13: Based on our conversation in slack, I was expecting a mapping from `(url, method)` to s3 client instance, rather than `(name, method)` to client instance.

lib/spack/spack/util/s3.py
```diff
@@ -4,83 +4,115 @@
...
+
+    def get_mirror_url(mirror):
+        return mirror.fetch_url if method == "fetch" else mirror.push_url
+
+    # Get all configured mirrors that could match.
+    all_mirrors = MirrorCollection()
+    mirrors = [
+        (name, mirror)
+        for name, mirror in all_mirrors.items()
+        if url_str.startswith(get_mirror_url(mirror))
```
scottwittenburg, line 43: Maybe I get it.  This is where you've converted from the provide `url` to a mirror name, which you've possibly done in support of the future goal of using mirror names (rather than mirror urls) everywhere?

scottwittenburg (approved):
Thanks @haampie!  Caching s3 clients is a great idea, and this looks to be a fairly straightforward (let's put that in air quotes though) adaptation of what we had before to support that.  I wasn't really familiar with this bit of the code before, so I added some questions to check my understanding.  Feel free to ignore them unless you have time to answer.

Otherwise, LGTM.  I'm not sure how well the unit tests mock `s3` read/write, so maybe I should keep an eye on the pipelines after this is merged :stuck_out_tongue_winking_eye:

lib/spack/spack/test/web.py
haampie: Many different urls correspond to one (or no) mirror, so I thought it was an easier key to use

lib/spack/spack/s3_handler.py
haampie: I think so yes

lib/spack/spack/util/s3.py
haampie: Actually the old code was already like this, but maybe a comment got lost

haampie:
@spackbot run pipeline

PR #34378, opened 2022-12-07
Add option skip_on_error when creating build caches

climbfuji:
This PR adds a new option `-s`/`--skip-on-error` to the `spack buildcache create` command to skip packages that cannot be relocated.

This allows users to create a build cache for all other packages in an environment with one command, simply skipping the ones that cannot be relocated. Maybe that flag should be called `skip-on-relocation-error` for clarity.


haampie:
To be honest I think it would be better to get rid of the pedantic `--allow-root` flag and make that the default. The name is obscure and you need it almost always.

climbfuji:
> To be honest I think it would be better to get rid of the pedantic `--allow-root` flag and make that the default. The name is obscure and you need it almost always.

It is not entirely clear to me what happens if the root path remains in those binaries, what is the effect? I definitely want to be absolutely certain that a package can be relocated before I add it to the build cache.

haampie:
When you install from a binary cache, Spack pretty much does find & replace on binaries & text files. Absolute paths in binaries are typically in string tables like `<string 1>\0<string 2>\0<string 3>`, with the start offsets referenced in the program, meaning that we can only keep strings the same length or shorten them.

So, what can go wrong when using a buildcache is when the packages were tarballed from a short prefix, and are installed to a long prefix. That's currently an error on install.

climbfuji:
> When you install from a binary cache, Spack pretty much does find & replace on binaries & text files. Absolute paths in binaries are typically in string tables like `<string 1>\0<string 2>\0<string 3>`, with the start offsets referenced in the program, meaning that we can only keep strings the same length or shorten them.
> 
> So, what can go wrong when using a buildcache is when the packages were tarballed from a short prefix, and are installed to a long prefix. That's currently an error on install.

But isn't there the `padded_length` config entry that can be set to avoid such problems?

haampie:
Yeah, my issue with `--allow-root` is that it's pedantic when you're already doing the right thing and build with long paths thanks to that padding config.

IMHO better would be for Spack to warn or error if the prefix you're pushing from is suspiciously short, instead of erroring when an absolute path is detected in the binary.

climbfuji:
Would would be the drawback of making `padded_length: true`  the default?

And then raise an error on install if you really, really run into a problem with the path being too short?

haampie:
The drawback is that people who don't use build caches end up with very awkward paths

```
$ which tar
/home/user/spack/opt/spack/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/.../bin/tar
```

I think the good thing about having --allow-root disabled by default is that it ensures
all buildcache entries are usable.

The bad thing is that you pretty much always need --allow-root because when
pushing multiple binaries, the odds are very high that at least one binary contains
its own install path or so (some binaries store the ./configure args, which may
contain the install dir, etc).

*If* you use padded length for building some packages, and you wanna push
packages of ^ that type, you *must* use --allow-root. But then errors may still
happen, cause maybe you're accidentally pushing a dependency that was still
installed in a shorter path or so.

Lastly, it's super annoying if you use padding but forget --allow-root in some
CI script or so, and it works just fine a couple times by chance because there
were no absolute paths embedded in binaries, until it builds something new
that happens to have an absolute path embedded in the binary, and Spack
errors and your build is gone.

So, with all that in mind, I think a friendlier approach for Spack is to:

1. Check _ahead of time_ what the prefix length of each spec is, instead
   of checking it halfway the process. This is a very quick & cheap check.
2. Error only if the prefix length is below a certain threshold

While (1) may be overly strict, with (2) it's at least less strict than Spack is
right now if you're nice and using padding during the build.

climbfuji:
> The drawback is that people who don't use build caches end up with very awkward paths
> 
> ```
> $ which tar
> /home/user/spack/opt/spack/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/__spack_placeholder__/.../bin/tar
> ```
> 
> I think the good thing about having --allow-root disabled by default is that it ensures all buildcache entries are usable.
> 
> The bad thing is that you pretty much always need --allow-root because when pushing multiple binaries, the odds are very high that at least one binary contains its own install path or so (some binaries store the ./configure args, which may contain the install dir, etc).
> 
> _If_ you use padded length for building some packages, and you wanna push packages of ^ that type, you _must_ use --allow-root. But then errors may still happen, cause maybe you're accidentally pushing a dependency that was still installed in a shorter path or so.
> 
> Lastly, it's super annoying if you use padding but forget --allow-root in some CI script or so, and it works just fine a couple times by chance because there were no absolute paths embedded in binaries, until it builds something new that happens to have an absolute path embedded in the binary, and Spack errors and your build is gone.
> 
> So, with all that in mind, I think a friendlier approach for Spack is to:
> 
> 1. Check _ahead of time_ what the prefix length of each spec is, instead
>    of checking it halfway the process. This is a very quick & cheap check.
> 2. Error only if the prefix length is below a certain threshold
> 
> While (1) may be overly strict, with (2) it's at least less strict than Spack is right now if you're nice and using padding during the build.

So, the short-term solution for me to keep working with build caches is to simply add `padded_length: True` and `--allow-root` when I am creating build caches. None of this is needed when I am installing from build caches, correct, hence it won't affect users of the build caches I created.

If that is all correct, then I will close this PR, use what I just suggested above and see what the spack developers decide to do.

PR #34379, opened 2022-12-07
streamstat: new package

cyrush:

PR #34380, opened 2022-12-07
add talass pkg

cyrush:


cyrush:
hmm, didn't realize there was a version here already, sorry about that.

cyrush:
@xuanhuang1  to support babelflow usage in spack proper, we will need streamstat and talass packages with releases. 

I started a PR for `streamstat`: https://github.com/spack/spack/pull/34379/

But talass already has a packged that has an older version. 

Can you create releases for these projects and we can update the spack packages?

PR #34381, opened 2022-12-07
Bump libxcrypt minimum automake to v1.14:

bvanessen:
According to the projects documentation it requires a version of automake 1.14 or newer.


bernhardkaindl (approved):
Correct, with automake@1.13, build aborts with:
```py
Makefile.am:7: error: require Automake 1.14, but have 1.13.4
```
Thanks!

PR #34382, opened 2022-12-07
mfem +rocm: depends_on rocsparse

eugeneswalker:
`mfem +rocm` needs explicit dependency on `rocsparse` in order to resolve this error when trying to install `mfem+rocm` against an externally provided ROCm stack:
```
==> Installing mfem-4.5.0-qc4qy7hqqdf36k6qtmforan6uqwk5je5
==> No binary for mfem-4.5.0-qc4qy7hqqdf36k6qtmforan6uqwk5je5 found: installing from source
==> Fetching https://bit.ly/mfem-4-5
==> No patches needed for mfem
==> mfem: Executing phase: 'configure'
==> Error: KeyError: 'No spec with name rocsparse in mfem@4.5.0 ...
```

@acfisher @goxberry @markcmiller86 @tzanio @v-dobrev 

CC @wspear

### DIFF
--- a/var/spack/repos/builtin/packages/mfem/package.py
+++ b/var/spack/repos/builtin/packages/mfem/package.py
@@ -238,6 +238,7 @@ class Mfem(Package, CudaPackage, ROCmPackage):
     conflicts("^mpich@4:", when="@:4.3+mpi")
 
     depends_on("mpi", when="+mpi")
+    depends_on("rocsparse", when="@4.4.0:+rocm")
     depends_on("hipsparse", when="@4.4.0:+rocm")
     depends_on("hypre@2.10.0:2.13", when="@:3.3+mpi")
     depends_on("hypre@:2.20.0", when="@3.4:4.2+mpi")


v-dobrev:
The dependency of `mfem` on `rocsparse` is indirect -- it is a dependency through `hipsparse` when using non-external `hipsparse`, i.e. Spack-built `hipsparse`. When using external `hipsparse`, the `mfem` package assumes that `hip` is also configured as external package and does not query the `rocsparse` package. It seems that the configuration that produces the error above breaks this assumption -- in other words: it uses external `hipsparse` and non-external (i.e. Spack-built) `hip` which looks like an invalid configuration to me.

If my conclusion above is not correct, can you please provide some more information on how to reproduce the error?

eugeneswalker:
> The dependency of `mfem` on `rocsparse` is indirect -- it is a dependency through `hipsparse` when using non-external `hipsparse`, i.e. Spack-built `hipsparse`. When using external `hipsparse`, the `mfem` package assumes that `hip` is also configured as external package and does not query the `rocsparse` package. It seems that the configuration that produces the error above breaks this assumption -- in other words: it uses external `hipsparse` and non-external (i.e. Spack-built) `hip` which looks like an invalid configuration to me.
> 
> If my conclusion above is not correct, can you please provide some more information on how to reproduce the error?

I'm attaching a Spack environment here that sets up the Crusher-provided ROCm 5.2.0 stack as externals, and contains a single spec `mfem +rocm` to be built with PrgEnv-gnu's GCC 11.2.0. If you simply try installing this environment you should be able to reproduce the error.

Spack environment: [mfem-spack.yaml.txt](https://github.com/spack/spack/files/10187319/mfem-spack.yaml.txt)

EDIT: To reduce the time required to reproduce the error, you could set up the E4S 22.11 PrgEnv-gnu deployment on Crusher as an Upstream Spack instance, in which case all the dependencies will be provided already, and you can jump right to the `mfem+rocm` install part. See https://spack.readthedocs.io/en/latest/chain.html
```
upstreams:
  e4s-22.11-prgenv-gnu: /gpfs/alpine/csc439/world-shared/E4S/ParaTools/22.11/PrgEnv-gnu/spack/opt/spack
```

v-dobrev:
I'll look into this. Just to confirm: are you using the `develop` branch of Spack or some release version?

Just as another data point: as part of the recent xSDK v0.8.0 release, we tested both on Crusher and Tioga with HIP and there were no issues building MFEM, in particular. The setups that we used are posted here:
```
https://github.com/xsdk-project/installxSDK/tree/r-0.8.0/platformFiles/crusher/gcc
https://github.com/xsdk-project/installxSDK/tree/r-0.8.0/platformFiles/tioga/gcc
```

eugeneswalker:
> I'll look into this. Just to confirm: are you using the `develop` branch of Spack or some release version?

For this most recent build on Crusher, I am using `e4s-22.11` branch of Spack, which is spack@develop frozen ~mid November.

> 
> Just as another data point: as part of the recent xSDK v0.8.0 release, we tested both on Crusher and Tioga with HIP and there were no issues building MFEM, in particular. The setups that we used are posted here:
> 
> ```
> https://github.com/xsdk-project/installxSDK/tree/r-0.8.0/platformFiles/crusher/gcc
> https://github.com/xsdk-project/installxSDK/tree/r-0.8.0/platformFiles/tioga/gcc
> ```

Interesting. Looking at the packages.yaml in platformFiles/crusher/gcc I don't see anything that stands out as explaining why it worked for you and not for me. I think you will see the problem reproduced if you use my spack.yaml. When you look at my spack.yaml, do you see anything that stands out as not a correct way to set things up?

v-dobrev:
It looks like the branch you are using does not have the latest fixes from `develop`. In particular, it has this section: https://github.com/spack/spack/blob/29ec128583da787c182c44e2e6388da83d9c2bfe/var/spack/repos/builtin/packages/mfem/package.py#L866-L877

which is now replaced by this section: https://github.com/spack/spack/blob/391ad8cec4209085a313d90ec183a1678e3242ca/var/spack/repos/builtin/packages/mfem/package.py#L873-L902

which handles external `hip`/`hipsparse` correctly.

eugeneswalker:
> It looks like the branch you are using does not have the latest fixes from `develop`. In particular, it has this section:
> ...
> which handles external `hip`/`hipsparse` correctly.

Excellent! I will make sure we cherry-pick this fix onto our E4S 22.11 branch, and use it for deployments going forward. Thank you!

PR #34388, opened 2022-12-08
new command: `spack pkg grep` to search package files

tgamblin:
It's very common for us to tell users to grep through the existing Spack packages to find examples of what they want, and it's also very common for package developers to do it. Now, searching packages is even easier.

`spack pkg grep` runs grep on all `package.py` files in repos known to Spack. It has no special options other than the search string; all options passed to it are forwarded along to `grep`.

```console
> spack pkg grep --help
usage: spack pkg grep [--help] ...

positional arguments:
  grep_args  arguments for grep

options:
  --help     show this help message and exit
```

## Examples:

```console
> spack pkg grep CMakePackage | head -3
/Users/gamblin2/src/spack/var/spack/repos/builtin/packages/3dtk/package.py:class _3dtk(CMakePackage):
/Users/gamblin2/src/spack/var/spack/repos/builtin/packages/abseil-cpp/package.py:class AbseilCpp(CMakePackage):
/Users/gamblin2/src/spack/var/spack/repos/builtin/packages/accfft/package.py:class Accfft(CMakePackage, CudaPackage):
```

```console
> spack pkg grep -Eho '(\S*)\(PythonPackage\)' | head -3
AwsParallelcluster(PythonPackage)
Awscli(PythonPackage)
Bueno(PythonPackage)
```

## Return value

This retains the return value semantics of `grep`:
* 0  for found,
* 1  for not found
* >1 for error

## Choosing a `grep`

You can set the ``SPACK_GREP`` environment variable to choose the ``grep`` executable this command should use.

### DIFF
--- a/lib/spack/spack/cmd/pkg.py
+++ b/lib/spack/spack/cmd/pkg.py
@@ -5,6 +5,9 @@
 
 from __future__ import print_function
 
+import argparse
+import itertools
+import os
 import sys
 
 import llnl.util.tty as tty
@@ -14,6 +17,7 @@ import spack.cmd
 import spack.cmd.common.arguments as arguments
 import spack.paths
 import spack.repo
+import spack.util.executable as exe
 import spack.util.package_hash as ph
 
 description = "query packages associated with particular git revisions"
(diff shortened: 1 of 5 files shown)


adamjstewart:
Interesting! This is obviously something I do daily. I currently have a `csv` alias that Cd's to Spack's Var directory, then I run grep from there. This is slightly quicker.

This is up there with `spack edit` and `spack unit-test` in the sense that it's a tiny shortcut on top of a normal shell command. I don't think it's too ridiculous to add, doesn't have a ton of overhead to maintain, and non-power users can just ignore it. So it's fine by me.

alalazo (approved):
Agree completely with @adamjstewart review

alalazo (changes requested):
...but tests must pass first! :laughing:

trws:
Looks good if tests pass.  Only thing I'd request, if we're doing this maybe allow the user to specify the name of their grep command? Mainly to deal with things like ggpep as default name for gnu grep on mac, wanting egrep or pt or whatever for different flags etc.

adamjstewart:
For @trws's suggestion, maybe we could rely on a `SPACK_GREP` env var like we do for `spack edit`?

trws:
I like that idea, it's similar to how FZF and others normally handle it so it would be convenient to add to existing environment setups that pick a grep for other tools.

tgamblin:
@trws added

adamjstewart (dismissed):
Works for me!

alalazo (approved):
@trws @adamjstewart If you want to have a final look this LGTM

adamjstewart (approved):

PR #34390, opened 2022-12-08
pixman: libs property: Use lib64 if it exists, else lib

bernhardkaindl:
@glennpj: After seeing alalzo's comment in #34288, I want to make sure to restrict the libs search
to .lib64 (if it exists, that should work on SUSE distros) or .lib.

@glennpj  Do you have a testcase for `spec["pixman"].libs` that you can test this with?


glennpj:
> @glennpj Do you have a testcase for `spec["pixman"].libs` that you can test this with?

I think it was building `cairo` when I noticed it. Beware of picking up the system library and masking it.

alalazo:

PR #34392, opened 2022-12-08
intel-parallel-studio: It is only available for x86_64

bernhardkaindl:
When trying to build other packages there were attempts to build intel-parallel-studio on arm64 and ppc64, but they failed as it is only provided for x86_64 by Intel. Copied from e.g. povray and others:
```py
var/spack/repos/builtin/packages/povray/package.py:    conflicts("+mkl", when="target=aarch64:", msg="Intel MKL only runs on x86")
var/spack/repos/builtin/packages/povray/package.py:    conflicts("+mkl", when="target=ppc64:", msg="Intel MKL only runs on x86")
var/spack/repos/builtin/packages/povray/package.py:    conflicts("+mkl", when="target=ppc64le:", msg="Intel MKL only runs on x86")
```

### DIFF
--- a/var/spack/repos/builtin/packages/intel-parallel-studio/package.py
+++ b/var/spack/repos/builtin/packages/intel-parallel-studio/package.py
@@ -547,6 +547,10 @@ class IntelParallelStudio(IntelPackage):
     provides("mpi", when="+mpi")
     provides("tbb", when="+tbb")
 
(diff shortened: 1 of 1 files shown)


alalazo (approved):

rscohn2 (approved):
Thanks!

PR #34396, opened 2022-12-08
py-reportlab: add 3.6.12

manuelakuhn:
sources from pypi

### DIFF
--- a/var/spack/repos/builtin/packages/py-reportlab/package.py
+++ b/var/spack/repos/builtin/packages/py-reportlab/package.py
@@ -10,8 +10,10 @@ class PyReportlab(PythonPackage):
     """The ReportLab Toolkit. An Open Source Python library for generating
     PDFs and graphics."""
 
+    homepage = "https://www.reportlab.com"
     pypi = "reportlab/reportlab-3.4.0.tar.gz"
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-reportlab/package.py
```diff
@@ -20,7 +22,19 @@ class PyReportlab(PythonPackage):
...
+    depends_on("pil@9:", when="@3.6.10:", type=("build", "run"))
     depends_on("pil@2.4.0:", type=("build", "run"))
     depends_on("freetype")
+
+    def patch(self):
+        filter_file(
+            "[FREETYPE_PATHS]",
+            "[FREETYPE_PATHS]\n"
+            + "lib={}\n".format(self.spec["freetype"].prefix.lib)
+            + "inc={}\n".format(self.spec["freetype"].prefix.include),
```
adamjstewart:
```suggestion
            + "lib={}\n".format(self.spec["freetype"].libs.directories[0])
            + "inc={}\n".format(self.spec["freetype"].headers.directories[0]),
```
This will work even if libs are in lib64 instead.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-reportlab/package.py
manuelakuhn: Thanks.

adamjstewart (approved):

PR #34397, opened 2022-12-08
Add pika-algorithms package and pika 0.11.0

msimberg:
Adds a new version of pika and a new package pika-algorithms. I'm adding them together because pika-algorithms is split out from pika, and requires pika 0.11.0.

This also adds some minor fixes and changes to the pika package:
- adds fmt as a dependency from 0.11.0 onwards
- adds 20 as a valid option for `cxxstd` in the fmt package 
- fixes the `when` clause for the `+p2300` variant

@albestro, @aurianer, @teonnik I've added you "by default" as maintainers as well since you're listed on the pika package. Let me know if you'd prefer not to be there.

### DIFF
--- a/var/spack/repos/builtin/packages/pika-algorithms/package.py
+++ b/var/spack/repos/builtin/packages/pika-algorithms/package.py
@@ -1,0 +1,53 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+from spack.package import *
+
+
+class PikaAlgorithms(CMakePackage):
+    """C++ parallel algorithms built on pika."""
+
+    homepage = "https://github.com/pika-org/pika-algorithms/"
+    url = "https://github.com/pika-org/pika-algorithms/archive/0.0.0.tar.gz"
+    git = "https://github.com/pika-org/pika-algorithms.git"
+    maintainers = ["msimberg", "albestro", "teonnik", "aurianer"]
+
+    version("0.1.0", sha256="64da008897dfa7373155595c46d2ce6b97a8a3cb5bea33ae7f2d1ff359f0d9b6")
+    version("main", branch="main")
+
+    generator = "Ninja"
+
+    map_cxxstd = lambda cxxstd: "2a" if cxxstd == "20" else cxxstd
+    cxxstds = ("17", "20")
+    variant(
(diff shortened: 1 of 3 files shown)


tldahlgren (dismissed):
Confirmed sha256 and branch of pika-algorithms.  Rest LGTM.

tldahlgren:
Looks like there's a style fix needed:

```
==> Running flake8 checks
[31](https://github.com/spack/spack/actions/runs/3647760453/jobs/6160375456#step:6:32)
==> Error: spack style found errors
[32](https://github.com/spack/spack/actions/runs/3647760453/jobs/6160375456#step:6:33)
var/spack/repos/builtin/packages/pika-algorithms/package.py:7: [F401] 'sys' imported but unused
```

teonnik (dismissed):

var/spack/repos/builtin/packages/pika-algorithms/package.py
```diff
@@ -0,0 +1,53 @@
...
+    depends_on("git", type="build")
+    depends_on("ninja", type="build")
+    depends_on("cmake@3.22:", type="build")
+
+    conflicts("%gcc@:8")
+    conflicts("%clang@:8")
+
+    # Other dependencies
+    depends_on("fmt@0.9:")
+    depends_on("pika@0.11:")
```
albestro: I read from the project page that `fmt` and `boost` are required for tests and examples.


> The following are required for tests and examples:
> - [header-only Boost](https://boost.org/) 1.71.0 or greater
> - [fmt](https://fmt.dev/latest/index.html)

Probably you are not going to enable testing for the package at this stage but, does that mean that you can drop the `fmt` dependency (instead of adding also `boost`)?

albestro (dismissed):
LGTM, see just minor comment about `pika-algorithms` dependencies.

var/spack/repos/builtin/packages/pika-algorithms/package.py
msimberg: Good catch. Unfortunately I have to go the latter way (adding `boost`) since I've made the mistake of requiring the two even with tests and examples disabled. It's not such a big deal since they're anyway transitive dependencies required by pika, but we should fix that for the next release.

msimberg: Also for reference: https://github.com/pika-org/pika-algorithms/issues/9.

albestro (approved):

teonnik:
LGTM!

I'd be happy to be a maintainer for this package.

teonnik (approved):

PR #34400, opened 2022-12-08
Add checksum for py-onnxmltools 1.11.1

iarspider:


iarspider:
@spackbot fix style

tldahlgren (commented):
Confirmed sha256.

var/spack/repos/builtin/packages/py-onnxmltools/package.py
```diff
@@ -12,6 +12,13 @@ class PyOnnxmltools(PythonPackage):
     homepage = "https://github.com/onnx/onnxmltools"
     pypi = "onnxmltools/onnxmltools-1.10.0.tar.gz"
 
+    # Source tarball not available on PyPI
```
adamjstewart, line 15: If this is the new normal we should update the `url` above and add a `url_for_version` so we don't have to specify the URL for all versions. Could also always use GitHub as the official source, even for older versions.

iarspider: I'd wait for next release: if they will not put source tarball on PyPI, then we can change the URL.

adamjstewart (approved):

iarspider:
@spackbot run pipeline

PR #34404, opened 2022-12-08
OpenCV:Generate .pc file for pkg-config build tool

koparasy:
Add a variant option to generate a ".pc" file for the OpenCV installation. This is required for some python-pip packages.

### DIFF
--- a/var/spack/repos/builtin/packages/opencv/package.py
+++ b/var/spack/repos/builtin/packages/opencv/package.py
@@ -775,6 +775,7 @@ class Opencv(CMakePackage, CudaPackage):
 
     # Other (variants)
     variant("shared", default=True, description="Enables the build of shared libraries")
+    variant("pkgconfig", default=False, description="Generate .pc file for pkg-config build tool")
     variant("powerpc", default=False, description="Enable PowerPC for GCC")
     variant(
         "fast-math",
@@ -972,6 +973,7 @@ class Opencv(CMakePackage, CudaPackage):
             [
                 self.define("ENABLE_CONFIG_VERIFICATION", True),
                 self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+                self.define_from_variant("OPENCV_GENERATE_PKGCONFIG", "pkgconfig"),
                 self.define("ENABLE_PRECOMPILED_HEADERS", False),
                 self.define_from_variant("WITH_LAPACK", "lapack"),
                 self.define_from_variant("ENABLE_POWERPC", "powerpc"),


adamjstewart:
Is there any reason someone wouldn't want a pkg-config file? Can we just always install it with no variant?

koparasy:
My only concern is that this is a deprecated option in OpenCV.  ---> [OpenCV-Cmake](https://github.com/opencv/opencv/blob/4.x/CMakeLists.txt#L529)

tldahlgren (approved):
LGTM. Confirmed the option.

adamjstewart:
Any idea why it's deprecated? Is that because it's now the default and no flag is needed? Or are they trying to move away from this? Which packages require this? What will they do when this flag is removed?

koparasy:
I had to enable this option for this python module : https://github.com/libffcv/ffcv . I am not sure if any other packages need this.

tldahlgren:
@adamjstewart Did you want any changes to this PR or are you okay with it being merged?

adamjstewart:
Only 1 of my 5 questions above has been answered, still waiting for more information before making a decision.

koparasy:
1) Any idea why it's deprecated? 

I don't know. I am not a power user of opencv, I just needed the ffcv package. 

2) Is that because it's now the default and no flag is needed? 

No, the option does not default. Thus, the ".pc" file is missing if not enabled.

3) Are they trying to move away from this? 

I am not sure, I am not actively involved in OpenCV. 

4) Which packages require this?  I only encountered it with ffcv. 

I found online also the same issue with GoCV. I guess any package that uses pkg-config and requires OpenCV4 may run to this issue. I have not tested this.  

5) What will they do when this flag is removed? 

I don't know.

I PRed this as an optional flag, to avoid someone having the same issue as I did. Few other packages may need this option. Thus, spack should expose it.

adamjstewart:
Thanks! I'm not really a heavy OpenCV user either, I just use it for ML. I did a bit of digging and the best reference I could find was https://github.com/opencv/opencv/issues/13154. It may be deprecated, but it doesn't seem like there are immediate plans to remove it, it just isn't supported very well. The authors suggest using cmake instead, but obviously not everyone uses cmake, so it's unfortunate.

I'm leaning towards adding this. Any thoughts from @bvanessen @glennpj?

glennpj:
I will note that I am not an opencv user myself, I provide the package as part of the software stack that I provide for researchers. With that as cover, I do not see an issue with adding a pkgconfig file. It seems that applications that can use it will, and applications that do not can just ignore it.

haampie:
Hi @koparasy, this package-related pull request can be migrated from `spack/spack` to `spack/spack-packages` using the [**migrate-pkg-prs utility**](https://github.com/spack/migrate-package-prs).

---

PR #34405, opened 2022-12-08
Support windows paths in a spec

johnwparent:
Current spec parsing precludes users on Windows from specifying a file path in a spec, this PR amends this by adding support for the Windows drive designator and file path separators  to the regex expressions that the spec syntax parses uses for file paths.

Obviously the spec is an complex area of Spack, so it may be that this change is not the best vector of approach or these changes introduce ambiguity or other problems into the spec, so this PR may not be the way we want to go, and in that case it can just serve as a starting point for a discussion.


alalazo (changes requested):
Can we add a unit test in `lib/spack/spack/test/spec_syntax.py` to avoid future regressions?

johnwparent:
> Can we add a unit test in `lib/spack/spack/test/spec_syntax.py` to avoid future regressions?

@alalazo There is one already, that's how I discovered this issue. According to the blame it was recently either added or moved from some other test suite where it was presumably skipped on Windows.

In general, is this change acceptable from a spec parsing standpoint?

alalazo:
> According to the blame it was recently either added or moved from some other test suite where it was presumably skipped on Windows.

Yes, we reworked the parser a few days ago. Can you re-enable the test on Windows?

> In general, is this change acceptable from a spec parsing standpoint?

If it passes tests it's acceptable for me. That regex needs to be able to parse all the file names we care about. Maybe we should specialize it by platform, so that on linux we can't match Win filenames and vice-versa.

johnwparent:
> If it passes tests it's acceptable for me. That regex needs to be able to parse all the file names we care about. Maybe we should specialize it by platform, so that on linux we can't match Win filenames and vice-versa.

I think that's a good idea, I'll implement and reenable the tests.

lib/spack/spack/test/spec_syntax.py
```diff
@@ -819,15 +822,26 @@ def test_redundant_spec(query_str, text_fmt, database):
...
+        ("libfoo ^..{0}..{0}libdwarf.yaml".format(os.path.sep), spack.spec.NoSuchSpecFileError),
+        ("libfoo ^.{0}libdwarf.yaml".format(os.path.sep), spack.spec.NoSuchSpecFileError),
+        (
+            "{0}bogus{1}path{1}libdwarf.yamlfoobar".format(abs_path_prefix, os.path.sep),
+            spack.spec.SpecFilenameError,
+        ),
+        (
+            "libdwarf^{0}bogus{1}path{1}libelf.yamlfoobar ^{0}path{1}to{1}bogus.yaml".format(
+                abs_path_prefix, os.path.sep
+            ),
```
alalazo: Instead of formatting input data to tests depending on the platform, I'd rather have the paths hard-coded as before[^1], but skip linux tests on Windows and vice-versa. 

[^1]: I think it is clearer to read

lib/spack/spack/parser.py
```diff
@@ -68,6 +69,7 @@
 import spack.variant
 import spack.version
 
+IS_WINDOWS = sys.platform == "win32"
```
alalazo, line 72: (Not for this PR, just a general comment) I wonder if we should have a single module where to add these constants. Right now I see ~40 matches for this same definition.

lib/spack/spack/parser.py
```diff
@@ -80,8 +82,13 @@
...
 
-#: A filename starts either with a "." or a "/" or a "{name}/"
-FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
+#: A filename starts either with a "." or a "/" or a "{name}/,
+# or on Windows, a drive letter followed by a colon and "\"
+# or "." or {name}\
+if not IS_WINDOWS:
+    FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
+else:
+    FILENAME = r"(\.|[a-zA-Z0-9-_]*\\|[a-zA-Z]:\\)([a-zA-Z0-9-_\.\\]*)(\.json|\.yaml)"
```
alalazo: I think we should also test these regexes as a standalone, to check that the desired behavior is met. We can check both expected matches and expected "rejections" and skip the test that is not dealing with the current platform. Also, let me know if you need help with anything related our testing modules.

alalazo (changes requested):

lib/spack/spack/parser.py
johnwparent: This is something I've been poking at here and there. It should be ready soon, just trying to find somewhere to put the work that makes logical sense and doesn't cause endless import cycles.

lib/spack/spack/test/spec_syntax.py
johnwparent: Sounds good, will do.

lib/spack/spack/parser.py
johnwparent: Sounds good I'll set that up. Thanks for the offer, I'll reach out if anything unexpected happens.

johnwparent: This test has been setup so that it runs essentially the same test on unix/windows and skips the platform it's not on at test time. There's a good amount of code duplication caused by this setup, so this is fine for now, but I'm working on setting up some class or utility that will allow us to create cross platform tests with less code duplication. Pytest offers some pretty need features when working with pytest meta-classes so I think that will help similar cases going forward.

lib/spack/spack/test/spec_syntax.py
```diff
@@ -818,26 +822,111 @@ def test_redundant_spec(query_str, text_fmt, database):
         ("x platform=test platform=test", spack.spec.DuplicateArchitectureError),
         ("x os=fe platform=test target=fe os=fe", spack.spec.DuplicateArchitectureError),
         ("x target=be platform=test os=be os=fe", spack.spec.DuplicateArchitectureError),
+    ],
+)
+def test_error_conditions(text, exc_cls):
+    with pytest.raises(exc_cls):
+        SpecParser(text).next_spec()
```
alalazo, line 842: We don't need to split the test, since we can mark individual input, see [here](https://docs.pytest.org/en/7.1.x/example/markers.html#marking-individual-tests-when-using-parametrize)

lib/spack/spack/test/spec_syntax.py
```diff
@@ -818,26 +822,111 @@ def test_redundant_spec(query_str, text_fmt, database):
...
+        "relative\\windows\\path.yaml",
+        pytest.param("/absolute/path/to/file.yaml", marks=pytest.mark.xfail(raises=AssertionError)),
+        pytest.param("relative/path/to/file.yaml", marks=pytest.mark.xfail(raises=AssertionError)),
+        pytest.param("./dot/rel/to/file.yaml", marks=pytest.mark.xfail(raises=AssertionError))
+    ]
+)
+def test_specfile_parsing_windows(filename):
+    match = re.match(FILENAME, filename)
+    assert match
+    assert match.end() == len(filename)
```
alalazo: Same here, we can have a single test and parametrize on input.

lib/spack/spack/test/spec_syntax.py
johnwparent: I saw that and thought it would make the tests a bit difficult to read/understand at a glance, but if it's alright with you I'll go ahead

johnwparent: Oh you can make `xfail` conditional, that cleans things up a lot!

johnwparent:
@spackbot run pipeline

johnwparent:
@alalazo I believe this is ready for another pass when you get the chance. Thanks for your review!

alalazo:
@johnwparent Apologies for the delay in reviewing this. I think we can do better in terms of reducing duplicated code. Let me know what you think of this diff:
```diff
diff --git a/lib/spack/spack/parser.py b/lib/spack/spack/parser.py
index 8e4ecb3b5a..452a08e7f6 100644
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -85,10 +85,12 @@
 #: A filename starts either with a "." or a "/" or a "{name}/,
 # or on Windows, a drive letter followed by a colon and "\"
 # or "." or {name}\
+WINDOWS_FILENAME = r"(\.|[a-zA-Z0-9-_]*\\|[a-zA-Z]:\\)([a-zA-Z0-9-_\.\\]*)(\.json|\.yaml)"
+UNIX_FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
 if not IS_WINDOWS:
-    FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
+    FILENAME = UNIX_FILENAME
 else:
-    FILENAME = r"(\.|[a-zA-Z0-9-_]*\\|[a-zA-Z]:\\)([a-zA-Z0-9-_\.\\]*)(\.json|\.yaml)"
+    FILENAME = WINDOWS_FILENAME
 
 VALUE = r"([a-zA-Z_0-9\-+\*.,:=\~\/\\]+)"
 QUOTED_VALUE = r"[\"']+([a-zA-Z_0-9\-+\*.,:=\~\/\\\s]+)[\"']+"
diff --git a/lib/spack/spack/test/spec_syntax.py b/lib/spack/spack/test/spec_syntax.py
index 8db1738de4..7fe674cda2 100644
--- a/lib/spack/spack/test/spec_syntax.py
+++ b/lib/spack/spack/test/spec_syntax.py
@@ -12,11 +12,24 @@
 import spack.platforms.test
 import spack.spec
 import spack.variant
-from spack.parser import FILENAME, SpecParser, SpecTokenizationError, Token, TokenType
+from spack.parser import (
+    UNIX_FILENAME,
+    WINDOWS_FILENAME,
+    SpecParser,
+    SpecTokenizationError,
+    Token,
+    TokenType,
+)
 
-is_windows = sys.platform == "win32"
+FAIL_ON_WINDOWS = pytest.mark.xfail(
+    sys.platform == "win32",
+    raises=(SpecTokenizationError, spack.spec.NoSuchHashError),
+    reason="Unix style path on Windows",
+)
 
-abs_path_prefix = "c:\\" if is_windows else "/"
+FAIL_ON_UNIX = pytest.mark.xfail(
+    sys.platform != "win32", raises=SpecTokenizationError, reason="Windows style path on Unix"
+)
 
 
 def simple_package_name(name):
@@ -834,116 +847,54 @@ def test_error_conditions(text, exc_cls):
     [
         # Specfile related errors
         pytest.param(
-            "/bogus/path/libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
-            ),
-        ),
-        pytest.param(
-            "../../libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
-            ),
-        ),
-        pytest.param(
-            "./libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
-            ),
+            "/bogus/path/libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS
         ),
+        pytest.param("../../libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS),
+        pytest.param("./libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS),
         pytest.param(
             "libfoo ^/bogus/path/libdwarf.yaml",
             spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
-            ),
+            marks=FAIL_ON_WINDOWS,
         ),
         pytest.param(
-            "libfoo ^../../libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
-            ),
+            "libfoo ^../../libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS
         ),
         pytest.param(
-            "libfoo ^./libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
-            ),
+            "libfoo ^./libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS
         ),
         pytest.param(
-            "/bogus/path/libdwarf.yamlfoobar",
-            spack.spec.SpecFilenameError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
-            ),
+            "/bogus/path/libdwarf.yamlfoobar", spack.spec.SpecFilenameError, marks=FAIL_ON_WINDOWS
         ),
         pytest.param(
             "libdwarf^/bogus/path/libelf.yamlfoobar ^/path/to/bogus.yaml",
             spack.spec.SpecFilenameError,
-            marks=pytest.mark.xfail(
-                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
-            ),
-        ),
-        pytest.param(
-            "c:\\bogus\\path\\libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
-        ),
-        pytest.param(
-            "..\\..\\libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
+            marks=FAIL_ON_WINDOWS,
         ),
         pytest.param(
-            ".\\libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
+            "c:\\bogus\\path\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX
         ),
+        pytest.param("..\\..\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX),
+        pytest.param(".\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX),
         pytest.param(
             "libfoo ^c:\\bogus\\path\\libdwarf.yaml",
             spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
+            marks=FAIL_ON_UNIX,
         ),
         pytest.param(
-            "libfoo ^..\\..\\libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
+            "libfoo ^..\\..\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX
         ),
         pytest.param(
-            "libfoo ^.\\libdwarf.yaml",
-            spack.spec.NoSuchSpecFileError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
+            "libfoo ^.\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX
         ),
         pytest.param(
             "c:\\bogus\\path\\libdwarf.yamlfoobar",
             spack.spec.SpecFilenameError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
+            marks=FAIL_ON_UNIX,
         ),
         pytest.param(
             "libdwarf^c:\\bogus\\path\\libelf.yamlfoobar ^c:\\path\\to\\bogus.yaml",
             spack.spec.SpecFilenameError,
-            marks=pytest.mark.xfail(
-                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
-            ),
+            marks=FAIL_ON_UNIX,
         ),
     ],
 )
@@ -953,48 +904,18 @@ def test_specfile_error_conditions_windows(text, exc_cls):
 
 
 @pytest.mark.parametrize(
-    "filename",
+    "filename,regex",
     [
-        pytest.param(
-            "c:\\abs\\windows\\path.yaml",
-            marks=pytest.mark.xfail(
-                not is_windows, raises=AssertionError, reason="Windows style path on unix"
-            ),
-        ),
-        pytest.param(
-            ".\\relative\\dot\\win\\path.yaml",
-            marks=pytest.mark.xfail(
-                not is_windows, raises=AssertionError, reason="Windows style path on unix"
-            ),
-        ),
-        pytest.param(
-            "relative\\windows\\path.yaml",
-            marks=pytest.mark.xfail(
-                not is_windows, raises=AssertionError, reason="Windows style path on unix"
-            ),
-        ),
-        pytest.param(
-            "/absolute/path/to/file.yaml",
-            marks=pytest.mark.xfail(
-                is_windows, raises=AssertionError, reason="Unix style path on Windows"
-            ),
-        ),
-        pytest.param(
-            "relative/path/to/file.yaml",
-            marks=pytest.mark.xfail(
-                is_windows, raises=AssertionError, reason="Unix style path on Windows"
-            ),
-        ),
-        pytest.param(
-            "./dot/rel/to/file.yaml",
-            marks=pytest.mark.xfail(
-                is_windows, raises=AssertionError, reason="Unix style path on Windows"
-            ),
-        ),
+        ("c:\\abs\\windows\\path.yaml", WINDOWS_FILENAME),
+        (".\\relative\\dot\\win\\path.yaml", WINDOWS_FILENAME),
+        ("relative\\windows\\path.yaml", WINDOWS_FILENAME),
+        ("/absolute/path/to/file.yaml", UNIX_FILENAME),
+        ("relative/path/to/file.yaml", UNIX_FILENAME),
+        ("./dot/rel/to/file.yaml", UNIX_FILENAME),
     ],
 )
-def test_specfile_parsing(filename):
-    match = re.match(FILENAME, filename)
+def test_specfile_parsing(filename, regex):
+    match = re.match(regex, filename)
     assert match
     assert match.end() == len(filename)
 
```

alalazo (changes requested):
This LGTM, modulo reducing code duplication

johnwparent:
> @johnwparent Apologies for the delay in reviewing this. I think we can do better in terms of reducing duplicated code. Let me know what you think of this diff:
> 
> ```diff
> diff --git a/lib/spack/spack/parser.py b/lib/spack/spack/parser.py
> index 8e4ecb3b5a..452a08e7f6 100644
> --- a/lib/spack/spack/parser.py
> +++ b/lib/spack/spack/parser.py
> @@ -85,10 +85,12 @@
>  #: A filename starts either with a "." or a "/" or a "{name}/,
>  # or on Windows, a drive letter followed by a colon and "\"
>  # or "." or {name}\
> +WINDOWS_FILENAME = r"(\.|[a-zA-Z0-9-_]*\\|[a-zA-Z]:\\)([a-zA-Z0-9-_\.\\]*)(\.json|\.yaml)"
> +UNIX_FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
>  if not IS_WINDOWS:
> -    FILENAME = r"(\.|\/|[a-zA-Z0-9-_]*\/)([a-zA-Z0-9-_\.\/]*)(\.json|\.yaml)"
> +    FILENAME = UNIX_FILENAME
>  else:
> -    FILENAME = r"(\.|[a-zA-Z0-9-_]*\\|[a-zA-Z]:\\)([a-zA-Z0-9-_\.\\]*)(\.json|\.yaml)"
> +    FILENAME = WINDOWS_FILENAME
>  
>  VALUE = r"([a-zA-Z_0-9\-+\*.,:=\~\/\\]+)"
>  QUOTED_VALUE = r"[\"']+([a-zA-Z_0-9\-+\*.,:=\~\/\\\s]+)[\"']+"
> diff --git a/lib/spack/spack/test/spec_syntax.py b/lib/spack/spack/test/spec_syntax.py
> index 8db1738de4..7fe674cda2 100644
> --- a/lib/spack/spack/test/spec_syntax.py
> +++ b/lib/spack/spack/test/spec_syntax.py
> @@ -12,11 +12,24 @@
>  import spack.platforms.test
>  import spack.spec
>  import spack.variant
> -from spack.parser import FILENAME, SpecParser, SpecTokenizationError, Token, TokenType
> +from spack.parser import (
> +    UNIX_FILENAME,
> +    WINDOWS_FILENAME,
> +    SpecParser,
> +    SpecTokenizationError,
> +    Token,
> +    TokenType,
> +)
>  
> -is_windows = sys.platform == "win32"
> +FAIL_ON_WINDOWS = pytest.mark.xfail(
> +    sys.platform == "win32",
> +    raises=(SpecTokenizationError, spack.spec.NoSuchHashError),
> +    reason="Unix style path on Windows",
> +)
>  
> -abs_path_prefix = "c:\\" if is_windows else "/"
> +FAIL_ON_UNIX = pytest.mark.xfail(
> +    sys.platform != "win32", raises=SpecTokenizationError, reason="Windows style path on Unix"
> +)
>  
>  
>  def simple_package_name(name):
> @@ -834,116 +847,54 @@ def test_error_conditions(text, exc_cls):
>      [
>          # Specfile related errors
>          pytest.param(
> -            "/bogus/path/libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
> -            ),
> -        ),
> -        pytest.param(
> -            "../../libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
> -            ),
> -        ),
> -        pytest.param(
> -            "./libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
> -            ),
> +            "/bogus/path/libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS
>          ),
> +        pytest.param("../../libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS),
> +        pytest.param("./libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS),
>          pytest.param(
>              "libfoo ^/bogus/path/libdwarf.yaml",
>              spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
> -            ),
> +            marks=FAIL_ON_WINDOWS,
>          ),
>          pytest.param(
> -            "libfoo ^../../libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
> -            ),
> +            "libfoo ^../../libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS
>          ),
>          pytest.param(
> -            "libfoo ^./libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=SpecTokenizationError, reason="Unix style path on Windows"
> -            ),
> +            "libfoo ^./libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_WINDOWS
>          ),
>          pytest.param(
> -            "/bogus/path/libdwarf.yamlfoobar",
> -            spack.spec.SpecFilenameError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
> -            ),
> +            "/bogus/path/libdwarf.yamlfoobar", spack.spec.SpecFilenameError, marks=FAIL_ON_WINDOWS
>          ),
>          pytest.param(
>              "libdwarf^/bogus/path/libelf.yamlfoobar ^/path/to/bogus.yaml",
>              spack.spec.SpecFilenameError,
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=spack.spec.NoSuchHashError, reason="Unix style path on Windows"
> -            ),
> -        ),
> -        pytest.param(
> -            "c:\\bogus\\path\\libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> -        ),
> -        pytest.param(
> -            "..\\..\\libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> +            marks=FAIL_ON_WINDOWS,
>          ),
>          pytest.param(
> -            ".\\libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> +            "c:\\bogus\\path\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX
>          ),
> +        pytest.param("..\\..\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX),
> +        pytest.param(".\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX),
>          pytest.param(
>              "libfoo ^c:\\bogus\\path\\libdwarf.yaml",
>              spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> +            marks=FAIL_ON_UNIX,
>          ),
>          pytest.param(
> -            "libfoo ^..\\..\\libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> +            "libfoo ^..\\..\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX
>          ),
>          pytest.param(
> -            "libfoo ^.\\libdwarf.yaml",
> -            spack.spec.NoSuchSpecFileError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> +            "libfoo ^.\\libdwarf.yaml", spack.spec.NoSuchSpecFileError, marks=FAIL_ON_UNIX
>          ),
>          pytest.param(
>              "c:\\bogus\\path\\libdwarf.yamlfoobar",
>              spack.spec.SpecFilenameError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> +            marks=FAIL_ON_UNIX,
>          ),
>          pytest.param(
>              "libdwarf^c:\\bogus\\path\\libelf.yamlfoobar ^c:\\path\\to\\bogus.yaml",
>              spack.spec.SpecFilenameError,
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=SpecTokenizationError, reason="Windows style path on unix"
> -            ),
> +            marks=FAIL_ON_UNIX,
>          ),
>      ],
>  )
> @@ -953,48 +904,18 @@ def test_specfile_error_conditions_windows(text, exc_cls):
>  
>  
>  @pytest.mark.parametrize(
> -    "filename",
> +    "filename,regex",
>      [
> -        pytest.param(
> -            "c:\\abs\\windows\\path.yaml",
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=AssertionError, reason="Windows style path on unix"
> -            ),
> -        ),
> -        pytest.param(
> -            ".\\relative\\dot\\win\\path.yaml",
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=AssertionError, reason="Windows style path on unix"
> -            ),
> -        ),
> -        pytest.param(
> -            "relative\\windows\\path.yaml",
> -            marks=pytest.mark.xfail(
> -                not is_windows, raises=AssertionError, reason="Windows style path on unix"
> -            ),
> -        ),
> -        pytest.param(
> -            "/absolute/path/to/file.yaml",
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=AssertionError, reason="Unix style path on Windows"
> -            ),
> -        ),
> -        pytest.param(
> -            "relative/path/to/file.yaml",
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=AssertionError, reason="Unix style path on Windows"
> -            ),
> -        ),
> -        pytest.param(
> -            "./dot/rel/to/file.yaml",
> -            marks=pytest.mark.xfail(
> -                is_windows, raises=AssertionError, reason="Unix style path on Windows"
> -            ),
> -        ),
> +        ("c:\\abs\\windows\\path.yaml", WINDOWS_FILENAME),
> +        (".\\relative\\dot\\win\\path.yaml", WINDOWS_FILENAME),
> +        ("relative\\windows\\path.yaml", WINDOWS_FILENAME),
> +        ("/absolute/path/to/file.yaml", UNIX_FILENAME),
> +        ("relative/path/to/file.yaml", UNIX_FILENAME),
> +        ("./dot/rel/to/file.yaml", UNIX_FILENAME),
>      ],
>  )
> -def test_specfile_parsing(filename):
> -    match = re.match(FILENAME, filename)
> +def test_specfile_parsing(filename, regex):
> +    match = re.match(regex, filename)
>      assert match
>      assert match.end() == len(filename)
>  
> ```

This diff looks great, thanks for the suggestions! I'll apply and push.

alalazo:
@spackbot run pipeline

alalazo (approved):

PR #34408, opened 2022-12-08
py-scikit-learn: add v1.2.0

adamjstewart:
https://scikit-learn.org/stable/whats_new/v1.2.html

### DIFF
--- a/var/spack/repos/builtin/packages/py-scikit-learn/package.py
+++ b/var/spack/repos/builtin/packages/py-scikit-learn/package.py
@@ -16,6 +16,7 @@ class PyScikitLearn(PythonPackage):
     maintainers = ["adamjstewart"]
 
     version("master", branch="master")
+    version("1.2.0", sha256="680b65b3caee469541385d2ca5b03ff70408f6c618c583948312f0d2125df680")
     version("1.1.3", sha256="bef51978a51ec19977700fe7b86aecea49c825884f3811756b74a3b152bb4e35")
     version("1.1.2", sha256="7c22d1305b16f08d57751a4ea36071e2215efb4c09cb79183faa4e8e82a3dbf8")
     version("1.1.1", sha256="3e77b71e8e644f86c8b5be7f1c285ef597de4c384961389ee3e9ca36c445b256")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-scikit-learn/package.py
```diff
@@ -51,43 +52,38 @@ class PyScikitLearn(PythonPackage):
     variant("openmp", default=True, description="Build with OpenMP support")
 
     # setup.py
-    depends_on("python@2.6:2.8,3.3:", type=("build", "run"))
-    depends_on("python@2.7:2.8,3.4:", when="@0.20:", type=("build", "run"))
-    depends_on("python@3.5:", when="@0.21:", type=("build", "run"))
-    depends_on("python@3.6:", when="@0.23:", type=("build", "run"))
-    depends_on("python@3.7:", when="@1.0:", type=("build", "run"))
     depends_on("python@3.8:", when="@1.1:", type=("build", "run"))
```
tldahlgren, line 55: Should this be a general requirement (to reflect earlier versions of the package that are still included above)?

tldahlgren (approved):
Confirmed the new sha256.

var/spack/repos/builtin/packages/py-scikit-learn/package.py
adamjstewart: I'm on the fence. Spack no longer supports Python 3.6 and older, so as far as I'm concerned, all older versions of py-scikit-learn support all older versions of Python (that Spack supports). That's why I removed the historical record. In the past, we kept this in in case someone used an external version of a dependency that was older. I'm not sure whether I want to be short and simple or long and complete here.

tldahlgren: FWIW. I'm on a machine at LLNL whose default is still `2.7.16`.  I know Spack  doesn't support older so just providing a data point.

adamjstewart: Oh yeah, it isn't hard to find Python 2, but if someone tries to use it and finds that it doesn't work, I'm no longer supporting it.

PR #34414, opened 2022-12-08
Mochi-Margo: needs to link with pthread for examples

kwryankrattiger:
Missing link flag prevents builds on crusher

CC: @eugeneswalker

### DIFF
--- a/var/spack/repos/builtin/packages/mochi-margo/package.py
+++ b/var/spack/repos/builtin/packages/mochi-margo/package.py
@@ -64,3 +64,10 @@ class MochiMargo(AutotoolsPackage):
     def autoreconf(self, spec, prefix):
         sh = which("sh")
         sh("./prepare.sh")
+
+    def flag_handler(self, name, flags):
+        if name == "ldflags":
+            flags.append("-pthread")
+            return (None, flags, None)
+
+        return (flags, None, None)


eugeneswalker:
This resolves the issue I face trying to build `mochi-margo %clang@14.0.0-rocm5.2.0` on OLCF Crusher using PrgEnv-amd.

I didn't have any trouble building `mochi-margo` with PrgEnv-gnu and PrgEnv-cray, perhaps one of the maintainers can speculate as to why mochi-margo doesn't get libpthread linked when the build is done with PrgEnv-amd?

Thanks!

mdorier:
This is weird, margo is looking for pthread here: https://github.com/mochi-hpc/mochi-margo/blob/main/configure.ac#L60
I'm not sure adding `-pthread` systematically and via spack is the right solution. I would rather suggest you open an issue on the Margo repository and we can look into adding the missing flag in margo's configure.ac directly. I think what may be happening is that the line I mentioned in configure.ac will add `-lpthread`, which is fine for most compilers, but the more correct way of adding pthread support is `-pthread`.

EDIT: I checked on my Linux machine with gcc and it adds both `-pthread` and `-lpthread`, so next we should check on Crusher what it add (or rather, doesn't add). Maybe the solution is just to manually add `-pthread` on line 61 and 62.

kwryankrattiger:
I am happy to close this in favor of a patch to mochi-margo. This is blocking time sensitive work so it was just faster to slap it in spack like this right now.

carns:
It's also possible that the ax_pthread.m4 macros need to be updated in the mochi-margo repo.  I think that file came from the gnu m4 collection (https://www.gnu.org/software/autoconf-archive/ax_pthread.html).  Maybe a newer version would produce a more correct result for this particular toolchain combination?

At any rate, the purpose of that autoconf m4 macro is to isolate the detection of platform-specific pthread settings in one place.  It would be nice if we could keep fixes there if needed.

eugeneswalker:
> It's also possible that the ax_pthread.m4 macros need to be updated in the mochi-margo repo. I think that file came from the gnu m4 collection (https://www.gnu.org/software/autoconf-archive/ax_pthread.html). Maybe a newer version would produce a more correct result for this particular toolchain combination?

How could I try a newer version of the ax_pthread.m4 macros? I'm not familiar with this but am willing to try it out on `Crusher + PrgEnv-amd` and report back. Just need a little guidance if you know how I could test this out.

I tried staging `mochi-margo` source and searching for:
```
$spack-src> grep -Inr AX_PTHREAD ./*
# no result

#spack-src> find . -type f -name ax_pthread.m4
# no result
```

mdorier:
I don't think this file is the problem (looks like it's a pretty recent version, from 2019); it's on the m4 folder: https://github.com/mochi-hpc/mochi-margo/blob/main/m4/ax_pthread.m4 but my best guess is that the compiler wrapper trips it somehow and prevents it from adding the flag.

eugeneswalker:
> I don't think this file is the problem (looks like it's a pretty recent version, from 2019); it's on the m4 folder: https://github.com/mochi-hpc/mochi-margo/blob/main/m4/ax_pthread.m4 but my best guess is that the compiler wrapper trips it somehow and prevents it from adding the flag.

I see `ax_pthread.m4` is not present in `mochi-margo@0.9.6` which is the version we're having issues with, which we need as a dependency for `unifyfs`. Building `v0.10.0` works, all else equal, and `ax_pthread.m4` is present there. Perhaps we could patch in the `ax_pthread.m4` file for version 0.9.6 and it would work? Would this be difficult to do? We could do it on our end if we just had some guidance on how to make ax_pthread.m4 work in 0.9.6.

carns:
This is the complete commit upstream in mochi-margo that includes the corresponding autoconf and pkgconfig changes to use the gnu m4 macro:

https://github.com/mochi-hpc/mochi-margo/commit/95f8904f70d6a535618b10b02be2ff7549723e76

My preference would be for the build to use the most recent version of mochi-margo if possible, but these changes are not that invasive if this specific change must be backported for your use case.

eugeneswalker:
> This is the complete commit upstream in mochi-margo that includes the corresponding autoconf and pkgconfig changes to use the gnu m4 macro:
> 
> [mochi-hpc/mochi-margo@95f8904](https://github.com/mochi-hpc/mochi-margo/commit/95f8904f70d6a535618b10b02be2ff7549723e76)
> 
> My preference would be for the build to use the most recent version of mochi-margo if possible, but these changes are not that invasive if this specific change must be backported for your use case.

I am checking with `unifyfs` package maintainer @CamStan to see if we can't use a newer version of `mochi-margo`. In the meantime, patching in the commit you identified resolved the error for me. Thank you!

mdorier:
Closing this PR as the problem is resolved in Margo.

PR #34424, opened 2022-12-09
Don't build tests for `fmt`

msimberg:
Building the tests take a decent amount of time and is not necessary, so this turns the tests off.

### DIFF
--- a/var/spack/repos/builtin/packages/fmt/package.py
+++ b/var/spack/repos/builtin/packages/fmt/package.py
@@ -106,4 +106,7 @@ class Fmt(CMakePackage):
         # and call to build "doc" target
         args.append("-DFMT_DOC=OFF")
 
+        # Don't build tests
+        args.append(self.define("FMT_TEST", self.run_tests))
+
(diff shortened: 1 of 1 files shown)


alecbcs:
@msimberg do these tests not verify that `fmt` is compiled and executes in an installation properly? Just wondering if we'll loose anything by turning off the tests here.

msimberg:
> @msimberg do these tests not verify that `fmt` is compiled and executes in an installation properly? Just wondering if we'll loose anything by turning off the tests here.

That's a good question. I'm not really familiar with the testing capabilities of spack, but by default it for sure doesn't _run_ tests (the tests were built, but that's why I opened this PR...). Squinting at other packages I made `FMT_TEST` now use `self.run_tests` instead which looks like it should be correct, but I couldn't get it to run the tests with `spack install --test=root fmt`. Do you know more about the testing features in spack?

alalazo (approved):
LGTM. Build-time tests here are run only if requested by the user

PR #34430, opened 2022-12-09
py-simplejson: add 3.18.0

manuelakuhn:
https://github.com/simplejson/simplejson/tree/v3.18.0

### DIFF
--- a/var/spack/repos/builtin/packages/py-simplejson/package.py
+++ b/var/spack/repos/builtin/packages/py-simplejson/package.py
@@ -13,6 +13,7 @@ class PySimplejson(PythonPackage):
     homepage = "https://github.com/simplejson/simplejson"
     pypi = "simplejson/simplejson-3.10.0.tar.gz"
 
+    version("3.18.0", sha256="58a429d2c2fa80834115b923ff689622de8f214cf0dc4afa9f59e824b444ab31")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-simplejson/package.py
```diff
@@ -27,5 +28,5 @@ class PySimplejson(PythonPackage):
     version("3.8.0", sha256="217e4797da3a9a4a9fbe6722e0db98070b8443a88212d7acdbd241a7668141d9")
     version("3.3.0", sha256="7a8a6bd82e111976aeb06138316ab10847adf612925072eaff8512228bcf9a1f")
 
-    depends_on("python@2.5:2.8,3.3:", type=("build", "run"), when="@3.16.0:")
+    depends_on("python@2.5:2,3.3:", when="@3.16.0:", type=("build", "run"))
```
adamjstewart, line 31: You could honestly just delete this line, Spack only supports Python 3.7+ anyway

adamjstewart (approved):

var/spack/repos/builtin/packages/py-simplejson/package.py
manuelakuhn: I'm always reluctant to just delete working specs (just in case) :-)

PR #34431, opened 2022-12-09
package/libproxy: fix py3 install

Sinan81:
in centos7, libproxy tries to install python under system directories. Hence, why this fix is needed.


Sinan81:
pinging @adamjstewart since he added python variant recently. This is not urgent.

Sinan81:
@spackbot fix style

var/spack/repos/builtin/packages/libproxy/package.py
```diff
@@ -26,11 +26,18 @@ class Libproxy(CMakePackage):
     depends_on("perl", type=("build", "run"), when="+perl")
     depends_on("python", type=("build", "run"), when="+python")
```
adamjstewart: You can remove this line, it's implied by `extends`

var/spack/repos/builtin/packages/libproxy/package.py
```diff
@@ -26,11 +26,18 @@ class Libproxy(CMakePackage):
...
             self.define_from_variant("WITH_PERL", "perl"),
             self.define_from_variant("WITH_PYTHON3", "python"),
             self.define("WITH_DOTNET", False),
             self.define("WITH_PYTHON2", False),
             self.define("WITH_VALA", False),
         ]
+        if "+python" in self.spec:
+            pynver = "python{0}".format(self.spec["python"].version.up_to(2))
+            pysite = join_path(self.prefix.lib, pynver, "site-packages")
+            args.append(self.define("PYTHON3_SITEPKG_DIR", pysite))
```
adamjstewart:
```suggestion
            args.append(self.define("PYTHON3_SITEPKG_DIR", python_platlib))
```
The exact directory changes depending on OS, Python version, external vs. internal, time of day, mood of your computer, etc, etc. Let me know if this doesn't work, sometimes it's finicky and thinks `python_platlib` isn't defined.

adamjstewart (changes requested):

Sinan81:
Why not disable python variant by default?

adamjstewart:
Up to you, this package has no official maintainer listed. I translate this to "no one care about the package or its defaults, so we can change whatever we want"

adamjstewart (dismissed):
LGTM, just confirm that `python_platlib` is working as expected and I'll merge.

adamjstewart (dismissed):

Sinan81:
I confirm that python_platlib worked **during build.** when i try to import libproxy though, I am getting runtime errors due to python referring to the system installation of libproxy. Not sure if I'm importing correctly. Let me find an example.

adamjstewart:
Might need to set LD_LIBRARY_PATH, Python is a nightmare

Sinan81:
it works after adding setup_run_env.

Sinan81:
also noticed version 0.4.16 throws runtime errors with latest python. Hence added a conflict statement.

Sinan81:
Does Spack CI automatically checks if after the build, python packages import just fine?

var/spack/repos/builtin/packages/libproxy/package.py
```diff
@@ -20,17 +20,27 @@ class Libproxy(CMakePackage):
...
-    variant("python", default=True, description="Enable Python bindings", when="@0.4.16:")
+    variant("python", default=False, description="Enable Python bindings", when="@0.4.16:")
 
     depends_on("zlib")
     depends_on("perl", type=("build", "run"), when="+perl")
-    depends_on("python", type=("build", "run"), when="+python")
+
+    extends("python", when="+python")
+
+    conflicts("python@3.9:", when="@:0.4.16")
```
adamjstewart:
```suggestion
    extends("python@:3.8", when="+python")
```
I believe this works as well

var/spack/repos/builtin/packages/libproxy/package.py
```diff
@@ -20,17 +20,27 @@ class Libproxy(CMakePackage):
...
             self.define("WITH_PYTHON2", False),
             self.define("WITH_VALA", False),
         ]
+        if "+python" in self.spec:
+            args.append(self.define("PYTHON3_SITEPKG_DIR", python_platlib))
+        return args
+
+    def setup_run_environment(self, env):
+        if "+python" in self.spec:
+            env.prepend_path("LD_LIBRARY_PATH", self.prefix.lib)
```
adamjstewart: It's unfortunate that rpath doesn't work for Python, but we do the same thing in other packages too. Actually, can you do this for macOS too? Something like:
```python
    env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", self.prefix.lib)                                    
else:                                                                                       
    env.prepend_path("LD_LIBRARY_PATH", self.prefix.lib) 
```
See `py-cartopy` for another example of this. Any idea if the path is ever in `lib64` instead of `lib`?

adamjstewart:
> Does Spack CI automatically checks if after the build, python packages import just fine?

1. Spack CI only tests packages in our CI stack (E4S, ML, etc.), it doesn't test all packages
2. I don't believe we build with `--test=root`, but I would like to add that someday, at least for our Python stacks

tgamblin:
The main E4S stack builds with tests enabled. We could expand the number of stacks that do this.

adamjstewart:
`--test=root` should be possible for the ML stack, although I wouldn't be surprised if some bugs pop out. `--test=all` would break with a lot of circular deps. At the very least, we should run `spack test` after `spack install`. For Python packages, the main thing to be concerned about is import tests, which are included in both.

Sinan81:
In regards to an earlier question, there's no `lib64` directory but there's libexec:
```$ tree
.
├── bin
│   └── proxy
├── include
│   └── proxy.h
├── lib
│   ├── libproxy
│   │   └── 0.4.17
│   │       └── modules
│   │           ├── config_gnome3.so
│   │           ├── config_kde.so
│   │           └── config_pacrunner.so
│   ├── libproxy.so -> libproxy.so.1
│   ├── libproxy.so.1 -> libproxy.so.1.0.0
│   ├── libproxy.so.1.0.0
│   └── pkgconfig
│       └── libproxy-1.0.pc
├── libexec
│   └── pxgsettings
└── share
    └── cmake
        └── Modules
            └── Findlibproxy.cmake

11 directories, 11 files
```

adamjstewart:
There's no lib64 directory on CentOS 7, but that doesn't mean there's no lib64 directory on other OSes...

var/spack/repos/builtin/packages/libproxy/package.py
```diff
@@ -20,17 +20,29 @@ class Libproxy(CMakePackage):
...
             self.define("WITH_PYTHON2", False),
             self.define("WITH_VALA", False),
         ]
+        if "+python" in self.spec:
+            args.append(self.define("PYTHON3_SITEPKG_DIR", python_platlib))
+        return args
+
+    def setup_run_environment(self, env):
+        if "+python" in self.spec:
+            libs = self.prefix.lib
```
adamjstewart: Would probably be better to use `self.spec["libproxy"].libs.directories[0]`. You may have to implement a `libs` method if the libraries aren't named `libproxy.so`.

adamjstewart (approved):

PR #34432, opened 2022-12-09
Fix openblas build with intel compiler

glennpj:
This PR patches the f_check script to detect the ifort compiler and ensure that F_COMPILER is iset to INTEL. This problem was introduced with openblas-0.3.21. Without this patch, the value of F_COMPILER falls back to G77 and icc rather than ifort is used for the linking stage. That results in the openblas library missing libifcore, which in turn means many Fotran programs can not be compiled with ifort.


glennpj:
Note that I will be posting this patch upstream as well.

glennpj:
This was posted upstream as well.
https://github.com/xianyi/OpenBLAS/pull/3858

rscohn2 (approved):
Thanks!

PR #34433, opened 2022-12-09
Add `--fresh` to docs to actually upgrade spack environments

alecbcs:
From my experience, if I don't add the `--fresh` option when running concretize it will instead decide to reuse all of my existing installed packages and thus will never upgrade my installed environment.

### DIFF
--- a/lib/spack/docs/replace_conda_homebrew.rst
+++ b/lib/spack/docs/replace_conda_homebrew.rst
@@ -184,7 +184,7 @@ simply run the following commands:
 .. code-block:: console
 
    $ spack env activate myenv
-   $ spack concretize --force
+   $ spack concretize --force --fresh
    $ spack install
 
 The ``--force`` flag tells Spack to overwrite its previous concretization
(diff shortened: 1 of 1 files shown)


alecbcs:
Going to add @haampie to this conversation incase he's got any ideas about how we could make the process of upgrading spack environments a bit easier.

lib/spack/docs/replace_conda_homebrew.rst
```diff
@@ -184,7 +184,7 @@ simply run the following commands:
 .. code-block:: console
 
    $ spack env activate myenv
-   $ spack concretize --force
+   $ spack concretize --force --fresh
```
alalazo: Can you add a couple of lines below on what `--fresh` does (similarly to what written for `--force`)?

alalazo (changes requested):
Thanks! Only a very minor request.

lib/spack/docs/replace_conda_homebrew.rst
alecbcs: Good idea! I wasn't sure if `--fresh` was sufficiently covered elsewhere in the docs but I like think it's worth having a little block about it here too.

alecbcs:
I realized that we're not telling folks how to prevent their environments from ending up with multiple versions of the same package so I also just added another little section describing how you can use `spack mark -i --all` and `spack gc` to more closely mimmic the behavior of Homebrew.

alalazo (approved):

PR #34434, opened 2022-12-10
OpenBLAS: change make target to all

joequant:
This causes the Makefile to serialize the make targets by calling the target all.  If the make runs multiple targets they will run in parallel causing compile errors.


adamjstewart:
I wonder if this is related to OpenBLAS being broken in CI on develop.

haampie:
It's unrelated

haampie:
This is supserseded by #35395, thanks for opening it though!

PR #34436, opened 2022-12-10
valgrind: add v3.20.0, mark macOS conflict

adamjstewart:
Was quite surprised to find that valgrind simply does not support macOS anymore. Couldn't test the new version, hopefully it works.


michaelkuhn (approved):
Works for me (tested with `~mpi`)

PR #34442, opened 2022-12-10
Drop Python 3.6 again

adamjstewart:
Possible hack to avoid audits complaining

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -1050,7 +1050,7 @@ def generate_package_index(cache_prefix, concurrency=32):
     try:
         file_list, read_fn = _spec_files_from_cache(cache_prefix)
     except ListMirrorSpecsError as err:
-        tty.error("Unabled to generate package index, {0}".format(err))
+        tty.error("Unable to generate package index, {0}".format(err))
         return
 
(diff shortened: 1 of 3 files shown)


alalazo (commented):
Fine with me, but we probably need to mark a few packages as "core" files, so that we run unit tests on them. Here we see audit passing, but we don't see `rhel8` platform python (since that is not triggered by package only changes so far).

adamjstewart:
Can this be merged?

alalazo:
We still need to run unit tests on it. Right now this is considered a "package only" PR and thus unit tests are skipped

adamjstewart:
What's the easiest way to do that?

adamjstewart:
Fixed a typo in core, hopefully that triggers core tests.

alalazo:
@spackbot run pipeline

alalazo (dismissed):

alalazo:
The two issues with pipelines seems unrelated (`cabana` and `src` broken on `develop`)

adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

adamjstewart:
The tests actually passed! 😮

adamjstewart:
Ping @alalazo

alalazo (approved):

PR #34443, opened 2022-12-10
openblas: fix bound :7.3 to :7.3.0

haampie:
This patch:

https://gcc.gnu.org/legacy-ml/gcc-patches/2018-01/msg01962.html

is actually in Amazon Linux GCC 7.3.1, which we use in CI.

So we should not hold OpenBLAS back to something ancient because of it.

Old versions of OpenBLAS fail to detect the host arch of some of the AVX512 cpus of build nodes, causing build failures.

Of course we should try to set ARCH properly in OpenBLAS to avoid that it looks up the build arch, but that's quite some work.


adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

PR #34444, opened 2022-12-10
geant4-data: depends_on g4emlow@7.9.1 when @10.6

wdconinc:
Per https://geant4.web.cern.ch/node/1837 the correct dependency for 10.6 is on `g4emlow@7.9.1`, not on both `g4emlow@7.9` and `g4emlow@7.9.1`.

This is a minor cosmetic fix. The concretization for 10.6 works just fine here. But this removes the duplicate entry.

### DIFF
--- a/var/spack/repos/builtin/packages/geant4-data/package.py
+++ b/var/spack/repos/builtin/packages/geant4-data/package.py
@@ -86,7 +86,6 @@ class Geant4Data(BundlePackage):
         ],
         "10.6.0:10.6": [
             "g4ndl@4.6",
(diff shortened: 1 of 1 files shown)


drbenmorgan (approved):
If we want to be super strict, 7.9 was used for 10.6.0, with 7.9.1 for 10.6.1 onwards, but I don't think is critical for functionality.

drbenmorgan:
@spackbot help

drbenmorgan:
@spackbot re-run pipeline

alalazo (approved):

PR #27931, opened 2021-12-12
geant4: new version 11.0.0 (preferred remains 10.7.3)

wdconinc:
This adds support for new major geant4 version 11.0.0, [release notes](https://geant4-data.web.cern.ch/ReleaseNotes/ReleaseNotes.11.0.html).

Keeping preferred version at 10.7.3 (until 11.1.0?) since breakage is expected in downstream versions due to several interface changes. 

Dependency and requirement changes:
- `clhep@2.4.5.1:` (new version added)
- `vecgeom@1.1.18:` (new version added)
- `cmake@3.16:`
- `cxxstd=17:` (added conflicts with 11 and 14)
- updated `geant4-data` sets G4EMLOW-8.0, G4PARTICLEXS-4.0 (new versions added)

Test: successful build of `geant4@11.0.0+ipo~motif+opengl+python+qt+threads+vecgeom~x11 build_type=RelWithDebInfo cxxstd=17` on `%gcc@11.2.0 arch=linux-ubuntu21.10-skylake`

Maintainer tag: @drbenmorgan (EIC will require the geant4@11 features in the very near future)

TODO:
- [x] change now-deprecated GEANT4_BUILD_CXXSTD into standard CMAKE_CXX_STANDARD with allowed values 17, 20, 23,
- [x] support new GEANT4_USE_TBB,
- [x] support new GEANT4_USE_VTK visualization for `vtk@8.2:`
- [x] (decided not to do this) support new GEANT4_BUILD_SANITIZER (do we care?),
- [x] (decided not to do this) support new GEANT4_USE_Pythia8 (seems only used in one example)
- [x] add the optional G4TENDL `geant4-data` set for ParticleHP physics list proton-nucleus scattering as part of this PR (with variant in `geant4-data` that defaults to False since at 870 MB =~ sum of all other data sets),
- [x] (decided not to do this)if optional G4TENDL is set by variant, then apply the same to G4RealSurface which is also optional,

PR #34445, opened 2022-12-10
geant4: rm preference for 10.7.3 now that 11.1.0 is out

wdconinc:
Back when we added `geant4@11.0.0` to spack we kept the preferred version pinned on the last release of the 10.x series to allow people time to upgrade their dependents. [The thoughts back then](https://github.com/spack/spack/pull/27931) were that we would change this with 11.1.0. Since that is out now, we can remove the preference for 10.7.3. Dependents should (and by now hopefully have) add an explicit dependency on the 10.x series if that is the most recent version that works for them.


drbenmorgan:
I think this is o.k., as you say, users can, and probably should, specify an exact version if they require this.

drbenmorgan (approved):

drbenmorgan:
@spackbot re-run pipeline

alalazo (approved):

alalazo:
@spackbot run pipeline

PR #34446, opened 2022-12-10
Revert "Use `urllib` handler for `s3://` and `gs://`, improve `url_exists` through HEAD requests"

tgamblin:
Reverts spack/spack#34324

Attempting to fix windows CI -- fetch errors started appearing in `develop` after this PR.


haampie:
Windows support has bugs. I'd rather make Windows CI allowed to fail again until someone manages to fix file urls. Right now it's just broken all over the place thanks to use of `"file://" + path`.

PR #34447, opened 2022-12-10
hdf5: "hdf5@1.13:" depends on "cmake@3.18:"

bernhardkaindl:
Fixes build error: `CMake 3.18 or higher is required. You are running ...` from:
```py
spack install hdf5@1.13 # (or also: spack install hdf-vol-async@1.4) 
```

### DIFF
--- a/var/spack/repos/builtin/packages/hdf5/package.py
+++ b/var/spack/repos/builtin/packages/hdf5/package.py
@@ -193,6 +193,7 @@ class Hdf5(CMakePackage):
     )
 
     depends_on("cmake@3.12:", type="build")
+    depends_on("cmake@3.18:", type="build", when="@1.13:")
 
     depends_on("msmpi", when="+mpi platform=windows")
     depends_on("mpi", when="+mpi")


alalazo (commented):
LGTM, but curious where you got the message, since it seems in the code they require `3.12` as a minimum version [here](https://github.com/hpc-io/vol-async/blob/develop/CMakeLists.txt)

bernhardkaindl:
> LGTM, but curious where you got the message, since it seems in the code they require `3.12` as a minimum version [here](https://github.com/hpc-io/vol-async/blob/develop/CMakeLists.txt)

Thanks for the check! The correction is:

Instead of `hdf-vol-async@1.4`,

it is `hdf5@1.13:` (depended on by `hdf-vol-async@1.4)` which needs `cmake@3.18`!

References:
> https://github.com/HDFGroup/hdf5/blob/develop/CMakeLists.txt#L1
> https://github.com/HDFGroup/hdf5/blob/develop/release_docs/INSTALL_CMake.txt#L278
```py 
   1. We suggest you obtain the latest CMake for windows from the Kitware
      web site. The HDF5 1.13."X" product requires a minimum CMake version 3.18.
```      
----
Fix: Updated to apply to `var/spack/repos/builtin/packages/hdf5/package.py` instead:

hdf5:
```py
 depends_on("cmake@3.12:", type="build")
+depends_on("cmake@3.18:", type="build", when="@1.13:")
```

As all versions of `hdf5` already `depends_on` on `cmake@3.12`, `hdf5-vol-async` needs no change.

sethrj (approved):

PR #34448, opened 2022-12-10
cernlib: depends_on libxaw libxt

wdconinc:
Based on the following lines in the top level `CMakeLists.txt` (I can't deep link since gitlab.cern.ch not public), `cernlib` needs an explicit dependency on `libxaw` and `libxt`:
```cmake
find_package(X11  REQUIRED)
message(STATUS "CERNLIB: X11_Xt_LIB=${X11_Xt_LIB} X11_Xaw_LIB=${X11_Xaw_LIB} X11_LIBRARIES=${X11_LIBRARIES}")
```

### DIFF
--- a/var/spack/repos/builtin/packages/cernlib/package.py
+++ b/var/spack/repos/builtin/packages/cernlib/package.py
@@ -22,6 +22,8 @@ class Cernlib(CMakePackage):
 
     depends_on("motif")
     depends_on("libx11")
+    depends_on("libxaw")
+    depends_on("libxt")
 
     def cmake_args(self):
         args = ["-DCERNLIB_BUILD_SHARED:BOOL=ON"]


wdconinc:
Just to be clear, this failed to build for me (gcc-12, ubuntu22.10) without `libxaw`, then I checked the `CMakeLists.txt` and added both of them. In this case, it appears that `libxt` is transitively included via the `motif` dependency. Still, it makes sense to add the explicit dependency.

(Still have a `cernlib` build failure due to missing `libcrypt`, so another PR likely coming too.)

wdconinc:
@vvolkl Do you use this `cernlib` package?

vvolkl:
> @vvolkl Do you use this `cernlib` package?

No. We could add the spec to our builds to test it anyway though.

alalazo (dismissed):

alalazo (approved):

PR #34449, opened 2022-12-10
Restore usage of  file URL handling method on Win

johnwparent:
#34324 Introduces a regression that precludes the usage of filepath URLs on Windows. This PR resolves this regression.

the `urlopen` method cannot be directly invoked on Windows as filepath urls do not place nicely with that module. Perhaps the `read_from_url` method could use some optimization, but it is imperative that it be used.

### DIFF
--- a/lib/spack/spack/util/web.py
+++ b/lib/spack/spack/util/web.py
@@ -371,14 +371,13 @@ def url_exists(url, curl=None):
         _ = curl_exe(*curl_args, fail_on_error=False, output=os.devnull)
         return curl_exe.returncode == 0
 
-    # Otherwise use urllib.
+    # If we get here, then the only other fetch method option is urllib.
+    # So try to "read" from the URL and assume that *any* non-throwing
+    #  response contains the resource represented by the URL.
     try:
-        urlopen(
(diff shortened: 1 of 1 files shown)


haampie:
> as filepath urls do not place nicely with that module

that would be surprising to me, since cpython has a [FileHandler](https://github.com/python/cpython/blob/024ac542d738f56b36bdeb3517a10e93da5acab9/Lib/urllib/request.py#L1474-L1523) which is [always added]( https://github.com/python/cpython/blob/024ac542d738f56b36bdeb3517a10e93da5acab9/Lib/urllib/request.py#L575-L578)

Although not entirely unsurprising because it's Python. Lol at the comment in their standard library:

> not entirely sure what the rules are here

I guess file urls work differently on Windows. Notice that it's calling https://github.com/python/cpython/blob/024ac542d738f56b36bdeb3517a10e93da5acab9/Lib/urllib/request.py#L1503 which on Windows is https://github.com/python/cpython/blob/024ac542d738f56b36bdeb3517a10e93da5acab9/Lib/nturl2path.py#L8-L43

tgamblin:
Decided to go with a full revert with #34446 while we figure out how to fix this.

PR #34451, opened 2022-12-11
Add RPC lib with NCL+HDF4

vanderwb:
If you attempt to build `ncl` with `+hdf4`, the build can fail if: `hdf` was built with `+external-xdr` and the RPC library used is not `libc`. This PR adds logic to the NCL build recipe to add the RPC library (likely `-ltirpc`) to the Site.local config file.

### DIFF
--- a/var/spack/repos/builtin/packages/ncl/package.py
+++ b/var/spack/repos/builtin/packages/ncl/package.py
@@ -17,8 +17,10 @@ class Ncl(Package):
     Numerous analysis functions are built-in."""
 
     homepage = "https://www.ncl.ucar.edu"
-
+    git = "https://github.com/NCAR/ncl.git"
     url = "https://github.com/NCAR/ncl/archive/6.4.0.tar.gz"
+
+    maintainers = ["vanderwb"]
 
     version("6.6.2", sha256="cad4ee47fbb744269146e64298f9efa206bc03e7b86671e9729d8986bb4bc30e")
     version("6.5.0", sha256="133446f3302eddf237db56bf349e1ebf228240a7320699acc339a3d7ee414591")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/ncl/package.py
```diff
@@ -293,6 +293,17 @@ def prepare_install_config(self):
         with open(config_answers_filename, "r") as f:
             config_script(input=f)
 
+        # Modify Site.local to add RPC library if HDF4 is used w/external XDR
+        if "+hdf4" in self.spec:
+            hdf4 = self.spec["hdf"]
+
+            if "+external-xdr" in hdf4 and hdf4["rpc"].name != "libc":
```
alalazo: Can we unify into:
```
if self.spec.satisfies("^hdf4+external-xdr ^libc"):
```
?

alalazo (changes requested):

var/spack/repos/builtin/packages/ncl/package.py
vanderwb: Hi @alalazo - that will not work because we need to check for the absence of libc, not its presence. I could change it to:

```
if self.spec.satisfies("^hdf4+external-xdr") and not self.spec.satisfies("^libc"):
```

but I'd still rather check for it in the hdf(4) spec because I worry that libc could creep into the NCL spec via some other dependency.

If that sounds reasonable, I'll push to my branch a small syntax change to use `satisfies` and volunteer to maintain the package.

alalazo: Ok, then what about:
```
if self.spec.satisfies("^hdf4+external-xdr") and not self.spec["hdf4"].satisfies("^libc"):
```
? Great if you want to maintain the package!

vanderwb: @alalazo - sure, no problem. I've made the suggested change and pushed to my fork.

alalazo (approved):
Thanks!

PR #34452, opened 2022-12-11
Use file paths/urls correctly

haampie:
The main issue that's fixed is that Spack passes paths (as strings) to
functions that require urls. That wasn't an issue on unix, since there
you can simply concatenate `file://` and `path` and all is good, but on
Windows that gives invalid file urls.

Instead of all sorts of ad-hoc `if windows: fix_broken_file_url` this PR
adds two helper functions around Python's own path2url and reverse.
`spack.util.url.path_to_file_url` / `spack.util.url.file_url_string_to_path`

This allows me to remove the custom `spack.util.url.parse` function.
That function is generally broken, as it has its own interpretation of
the `file://` protocol: it tries to be smart about relative paths, but the
truth is that this is simply not supported by design, so any attempt
to make `file://x/y` mean the relative local path `x/y` is plain wrong.
This URI has a meaning: `hostname=x` and `path=/y`.

Also the variable interpolation in `spack.util.url.parse` has similar
issues, where Spack just tries to be too smart, disagreeing with
the specification, and not particularly following the principle of
least surprise. Just let the call-site interpolate variables if necessary.

Also fixes what I think is a bug, where some `spack buildcache` commands
used `-d` as a flag to mean `--mirror-url` requiring a URL, and others
`--directory`, requiring a path. (For @scottwittenburg)

Adding @tldahlgren as a reviewer for the environment include URL
stuff; fyi it also fixes a bug where a non-existing local path would
result in a FetchError, which is weird. See the updated test.

### DIFF
--- a/lib/spack/spack/cmd/buildcache.py
+++ b/lib/spack/spack/cmd/buildcache.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import sys
 import tempfile
+import urllib.parse
 
 import llnl.util.tty as tty
 
@@ -45,7 +46,7 @@ def setup_parser(subparser):
         "-r",
         "--rel",
         action="store_true",
-        help="make all rpaths relative" + " before creating tarballs.",
+        help="make all rpaths relative before creating tarballs.",
     )
     create.add_argument(
         "-f", "--force", action="store_true", help="overwrite tarball if it exists."
@@ -54,13 +55,13 @@ def setup_parser(subparser):
         "-u",
         "--unsigned",
         action="store_true",
-        help="create unsigned buildcache" + " tarballs for testing",
+        help="create unsigned buildcache tarballs for testing",
     )
     create.add_argument(
         "-a",
         "--allow-root",
         action="store_true",
-        help="allow install root string in binary files " + "after RPATH substitution",
+        help="allow install root string in binary files after RPATH substitution",
     )
     create.add_argument(
         "-k", "--key", metavar="key", type=str, default=None, help="Key for signing."
@@ -71,31 +72,31 @@ def setup_parser(subparser):
         "--directory",
         metavar="directory",
         type=str,
-        help="local directory where " + "buildcaches will be written.",
+        help="local directory where buildcaches will be written.",
     )
     output.add_argument(
         "-m",
         "--mirror-name",
         metavar="mirror-name",
         type=str,
-        help="name of the mirror where " + "buildcaches will be written.",
+        help="name of the mirror where buildcaches will be written.",
     )
     output.add_argument(
         "--mirror-url",
         metavar="mirror-url",
         type=str,
-        help="URL of the mirror where " + "buildcaches will be written.",
+        help="URL of the mirror where buildcaches will be written.",
     )
     create.add_argument(
         "--rebuild-index",
         action="store_true",
         default=False,
-        help="Regenerate buildcache index " + "after building package(s)",
+        help="Regenerate buildcache index after building package(s)",
     )
     create.add_argument(
         "--spec-file",
         default=None,
-        help=("Create buildcache entry for spec from json or " + "yaml file"),
+        help="Create buildcache entry for spec from json or yaml file",
     )
     create.add_argument(
         "--only",
@@ -124,19 +125,19 @@ def setup_parser(subparser):
         "-a",
         "--allow-root",
         action="store_true",
-        help="allow install root string in binary files " + "after RPATH substitution",
+        help="allow install root string in binary files after RPATH substitution",
     )
     install.add_argument(
         "-u",
         "--unsigned",
         action="store_true",
-        help="install unsigned buildcache" + " tarballs for testing",
+        help="install unsigned buildcache tarballs for testing",
     )
     install.add_argument(
         "-o",
         "--otherarch",
         action="store_true",
-        help="install specs from other architectures" + " instead of default platform and OS",
+        help="install specs from other architectures instead of default platform and OS",
     )
 
     arguments.add_common_arguments(install, ["specs"])
@@ -155,7 +156,7 @@ def setup_parser(subparser):
         "-a",
         "--allarch",
         action="store_true",
-        help="list specs for all available architectures" + " instead of default platform and OS",
+        help="list specs for all available architectures instead of default platform and OS",
     )
     arguments.add_common_arguments(listcache, ["specs"])
     listcache.set_defaults(func=list_fn)
@@ -204,7 +205,7 @@ def setup_parser(subparser):
     check.add_argument(
         "--spec-file",
         default=None,
-        help=("Check single spec from json or yaml file instead of release " + "specs file"),
+        help=("Check single spec from json or yaml file instead of release specs file"),
     )
 
     check.set_defaults(func=check_fn)
@@ -217,7 +218,7 @@ def setup_parser(subparser):
     download.add_argument(
         "--spec-file",
         default=None,
-        help=("Download built tarball for spec (from json or yaml file) " + "from mirror"),
+        help=("Download built tarball for spec (from json or yaml file) from mirror"),
     )
     download.add_argument(
         "-p", "--path", default=None, help="Path to directory where tarball should be downloaded"
@@ -234,7 +235,7 @@ def setup_parser(subparser):
     getbuildcachename.add_argument(
         "--spec-file",
         default=None,
-        help=("Path to spec json or yaml file for which buildcache name is " + "desired"),
+        help=("Path to spec json or yaml file for which buildcache name is desired"),
     )
     getbuildcachename.set_defaults(func=get_buildcache_name_fn)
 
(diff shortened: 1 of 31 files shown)


haampie:
@spackbot fix style

haampie:
@johnwparent can we get rid of the "custom" urlparse? To me it's the wrong approach.

lib/spack/spack/util/url.py
```diff
@@ -365,20 +289,3 @@ def parse_git_url(url):
             raise ValueError("bad port in git url: %s" % url)
 
     return (scheme, user, hostname, port, path)
-
-
-def is_url_format(url):
-    return re.search(r"^(file://|http://|https://|ftp://|s3://|gs://|ssh://|git://|/)", url)
-
-
-def require_url_format(url):
```
johnwparent, line 374: Are we still enforcing this transitively through `urllib`? I.e. `urllib` will throw if given anything but an expected url and the error message is clear, or does this refactor mean this is no longer necessary and we're just converting any potential non url strings into urls? I think it's the latter but it's not immediately obvious.

lib/spack/spack/test/cmd/env.py
```diff
@@ -707,9 +707,9 @@ def test_with_config_bad_include():
             e.concretize()
 
     err = str(exc)
-    assert "not retrieve configuration" in err
-    assert os.path.join("no", "such", "directory") in err
-
+    assert "missing include" in err
+    assert "/no/such/directory" in err
+    assert os.path.join("no", "such", "file.yaml") in err
```
johnwparent, line 712: This test isn't currently run on Windows but thanks for fixing it anyway!

johnwparent:
Overall this looks fine. I left one comment, which is fairly minor and does not need to hold up the progress of this branch. Testing locally it looks like the `urllib` utilities provides at least the same functionality (likely better) than our custom url parsing and we should be moving away from our own url parsing anyway, so I think it's perfectly reasonable to remove the custom urlparse.

lib/spack/spack/util/url.py
haampie: urllib won't throw during parsing. My thinking was that the parser can't be a validator, cause what scheme is supported depends on the call site. For buildcache etc we still have some validation that at least errors when a url is passed to `--directory` and vice versa, although I find it a bit pedantic to be honest ;p 

It's easier to validate after parsing, e.g. `assert url.scheme in supported_schemes`. Are there any obvious places where that should be added? For fetching at some point you'll just get a URLError such as `URLError: <urlopen error unknown url type: unknown>`.

lib/spack/spack/cmd/buildcache.py
```diff
@@ -692,11 +713,16 @@ def update_index(mirror_url, update_keys=False):
 
 def update_index_fn(args):
     """Update a buildcache index."""
-    outdir = "file://."
+    if args.directory:
+        push_url = spack.mirror.push_url_from_directory(args.directory)
+
+    if args.mirror_name:
```
tldahlgren: I know these are mutually exclusive, but why not `elif` here and at line 722?

haampie: It was copied from elsewhere, but sounds good to me

haampie:
@johnwparent can we merge this rather sooner than later? Then I can move on with #34498

Also @johnwparent, how about hostnames in file urls? E.g. `file://hostname/x` has `netloc=hostname` and `path=/x`. I don't think it should be part of this PR, since Spack erroneously interpreted that file url as a relative path `hostname/x` in before, so it was incorrect anyways.

johnwparent:
> @johnwparent can we merge this rather sooner than later? Then I can move on with #34498

I think this is ok to merge now.

> Also @johnwparent, how about hostnames in file urls? E.g. `file://hostname/x` has `netloc=hostname` and `path=/x`. I don't think it should be part of this PR, since Spack erroneously interpreted that file url as a relative path `hostname/x` in before, so it was incorrect anyways.

I'm inclined to agree with you here. I've started a secondary effort to retool our approach to urls/paths for better cross platform support and I can handle this there. This is blocking your other works so LGTM.

haampie:
Actually it's triggering a problem in CI where (probably) a `urlparse(...)` object is fed to `urlparse(...)`. Apparently that's not tested by unit tests

johnwparent:
> Actually it's triggering a problem in CI where (probably) a `urlparse(...)` object is fed to `urlparse(...)`. Apparently that's not tested by unit tests

Huh, I didn't notice a point where that could happen, surprising we don't test for something like that. 

If you don't want to add that test here, and just fix the issue, I'll add a to do to my effort to update that test suite.

haampie:
Apparently CI did not run on the right commit :(

1f627df did [run here](https://gitlab.spack.io/spack/spack/-/pipelines/238355), which actually ran on this [merge commit](https://gitlab.spack.io/spack/spack/-/commit/60e7ed08dc11ecba297f1b93acc3c3d67c2a54e9), which has e0ab93226264d3218545daefc876da17142ee413 as a parent instead of 1f627df :|

I tried to rebase on 94bac8d6dd9140820dbde9d1dda4ea6c5736ca43 to see if it syncs better. Ping @zackgalbreath / @scottwittenburg

scottwittenburg:
> which has [e0ab932](https://github.com/spack/spack/commit/e0ab93226264d3218545daefc876da17142ee413) as a parent instead of [1f627df](https://github.com/spack/spack/commit/1f627dfe40880231848588c2eab6c3efddeed320)

Hrm, yeah, there's definitely something wrong there.  The commit message has the right parents (including the PR HEAD sha we wanted to test), but it doesn't match the actual parents of the commit.

johnwparent:
> > which has [e0ab932](https://github.com/spack/spack/commit/e0ab93226264d3218545daefc876da17142ee413) as a parent instead of [1f627df](https://github.com/spack/spack/commit/1f627dfe40880231848588c2eab6c3efddeed320)
> 
> Hrm, yeah, there's definitely something wrong there. The commit message has the right parents (including the PR HEAD sha we wanted to test), but it doesn't match the actual parents of the commit.

I think this may somehow be a GH issue, I've seen RTD pulling the wrong commit a few times as well, and have heard complaints from other users as well today.

scottwittenburg:
So the sync script does have two sources of information that need to jive on the GH end.  The branch we create starts, basically, with:

```
git fetch github refs/pull/<pull_number>/head:<tmp_branch>
```

Then later we merge `<tmp_branch>` into a new branch based on the latest tested `develop`.  So that's the source of the *actual* parents of the merge commit we make.  On the other hand, we get one of the parents of the commit *message*  we push from a GH api endpoint that provides the pull request head sha.

If those two sources of information don't agree, then we're in trouble.  But I'm looking to see if there's a way we can be immune to a mismatch there, assuming that's what happened.

haampie:
> I think this may somehow be a GH issue, I've seen RTD pulling the wrong commit a few times as well, and have heard complaints from other users as well today.

Yeah, it's likely a Github issue.

https://readthedocs.org/projects/spack/builds/18906514/

Notice: **Failed to checkout revision: 1f627dfe40880231848588c2eab6c3efddeed320**

But then it should also really be a Gitlab CI failure instead of running on the wrong/old commit.

haampie:
Should we run a spackbot-rebuild-everything on this?

haampie:
@spackbot rebuild everything

haampie:
Seems like everything is on fire today :(

lib/spack/spack/environment/environment.py
```diff
@@ -926,46 +929,54 @@ def included_config_scopes(self):
...
 
-            # strip file URL prefix, if needed, to avoid unnecessary remote
-            # config processing for local files
-            config_path = config_path.replace("file://", "")
+            include_url = urllib.parse.urlparse(config_path)
 
-            if not os.path.exists(config_path):
+            # Transform file:// URLs to direct includes.
+            if include_url.scheme == "file":
+                config_path = urllib.request.url2pathname(include_url.path)
```
tldahlgren, line 936: What is an example of a case where `config_path` != `include_url.path` here?

haampie:
@spackbot run pipeline

lib/spack/spack/environment/environment.py
```diff
@@ -926,46 +929,54 @@ def included_config_scopes(self):
...
-                    staged_configs = (
-                        os.listdir(self.config_stage_dir)
-                        if os.path.exists(self.config_stage_dir)
-                        else []
+                staged_configs = (
+                    os.listdir(self.config_stage_dir)
+                    if os.path.exists(self.config_stage_dir)
+                    else []
+                )
+                remote_path = urllib.request.url2pathname(include_url.path)
```
tldahlgren, line 946: Same question here about `remote_path` != `include_url.path`.

tldahlgren: Also, IMO the variable name is misleading.  Is this not a local path?

tldahlgren (commented):
At this point looked at the environment change as requested offline.  Have some questions.

lib/spack/spack/environment/environment.py
tldahlgren: Is this needed for `windows`?

lib/spack/spack/test/cmd/env.py
tldahlgren: Great catch btw!

lib/spack/spack/environment/environment.py
tldahlgren: BUT, I can adjust when I resolve conflicts with one of my other environment-related PRs.

tldahlgren: BUT I can adjust the name after addressing merge conflicts in one of my other environment-related PRs.

haampie:
So, gitlab rebuild a few things during rebuild everything, but the vast majority failed with unrelated issues. I think it's good enough.

Now I'm praying that it won't take 10 attempts to make gitlab ci pass on nothing to rebuild.

@spackbot run pipeline

lib/spack/spack/environment/environment.py
haampie: Windows because `file:///C:/x` would give a path `/C:/x` which url2pathname transforms into `C:\\x`.

But also on Linux it decodes the url, e.g. `file:///hello%20world` -> `/hello world`.

lib/spack/spack/environment/environment.py
haampie: Yeah, that sounds like a better name

haampie:
@spackbot run pipeline

:|

haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

haampie:
Finally it's in a merge-able state again! Pinging @johnwparent one more time

johnwparent (approved):
Looks good from the Windows side of things!

PR #34460, opened 2022-12-12
py-submitit: add 1.4.5

matz-e:


adamjstewart (approved):
pkg_resources is still imported in one file, unclear if setuptools is still needed at run-time or not, but I'm fine to sticking with documented dependencies until we discover otherwise

matz-e:
> pkg_resources is still imported in one file, unclear if setuptools is still needed at run-time or not, but I'm fine to sticking with documented dependencies until we discover otherwise

IIRC, there is still a `pkg_resources` from `distutils` in the Python distribution? And yes, I'd expect this to break at some point, too.

adamjstewart:
For some reason I though pkg_resources was only included in setuptools, not distutils. But yeah, distutils will be gone entirely in Python 3.12.

adamjstewart (approved):

Issue #34192, opened 2022-11-29
apptainer: conmon config flag added in singularityce not available 


wdconinc:
#33891 has caused a regression in apptainer, which does not like the `--without-conmon` flag in its [mconfig](https://github.com/apptainer/apptainer/blob/main/mconfig), since apptainer uses [singularityce]([url](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/apptainer/package.py)) as its base.

_Originally posted by @wdconinc in https://github.com/spack/spack/issues/33891#issuecomment-1331101581_

wdconinc:
I don't know what the right approach is here: `if singularity_name == 'apptainer'` before `confstring += " --without-conmon"` seems hacky and won't handle well if/when apptainer adds conmon support...

alalazo:
The correct approach is either to have a common shared base class, or (simpler) to create an overrideable method in which all the SingularityCE options are put. That method can be then overridden by Apptainer.

wdconinc:
This was fixed in https://github.com/spack/spack/pull/34329 by overriding `edit`, but I only noticed that after I opened https://github.com/spack/spack/pull/34474.

PR #34474, opened 2022-12-12
singularity, apptainer: --without-conmon into @property config_options

wdconinc:
Per https://github.com/spack/spack/issues/34192, apptainer does not support `--without-conmon`, so we introduce a base class `config_options` property that can be overridden in the `apptainer` package.


alalazo (dismissed):

alalazo:
@wdconinc Is this ready for review?

wdconinc:
This is ready for review. It is possible that apptainer and singularityce may diverge too much to continue doing this... but for now it's good.

wdconinc:
@spackbot rerun pipeline

alalazo (approved):

PR #34475, opened 2022-12-12
py-shapely: add v2.0.0

adamjstewart:
Successfully builds on macOS 12.6.1 with Python 3.10.8 and Apple Clang 14.0.0.

### DIFF
--- a/var/spack/repos/builtin/packages/py-numpy/package.py
+++ b/var/spack/repos/builtin/packages/py-numpy/package.py
@@ -93,7 +93,9 @@ class PyNumpy(PythonPackage):
     depends_on("python@3.6:3.10", type=("build", "link", "run"), when="@1.19")
     depends_on("python@3.7:3.10", type=("build", "link", "run"), when="@1.20:1.21")
     depends_on("python@3.8:", type=("build", "link", "run"), when="@1.22:")
-    depends_on("py-setuptools@:59", type=("build", "run"))
+    # https://github.com/spack/spack/pull/32078
+    depends_on("py-setuptools@:63", type=("build", "run"))
+    depends_on("py-setuptools@:59", when="@:1.22.1", type=("build", "run"))
     # Check pyproject.toml for updates to the required cython version
     depends_on("py-cython@0.29.13:2", when="@1.18.0:", type="build")
     depends_on("py-cython@0.29.14:2", when="@1.18.1:", type="build")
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/py-numpy/package.py
```diff
@@ -93,7 +93,9 @@ class PyNumpy(PythonPackage):
     depends_on("python@3.6:3.10", type=("build", "link", "run"), when="@1.19")
     depends_on("python@3.7:3.10", type=("build", "link", "run"), when="@1.20:1.21")
     depends_on("python@3.8:", type=("build", "link", "run"), when="@1.22:")
-    depends_on("py-setuptools@:59", type=("build", "run"))
+    # https://github.com/spack/spack/pull/32078
+    depends_on("py-setuptools@:63", type=("build", "run"))
+    depends_on("py-setuptools@:59", when="@:1.22.1", type=("build", "run"))
```
adamjstewart, line 98: Setuptools 59 is fine, 60–63 raise deprecation warnings, and 64 fails entirely. Unfortunately, Spack cannot yet resolve build deps separately, and each DAG must contain a single version of each package. Since shapely now requires newer setuptools, I've relaxed this dep to support this. Once Spack can resolve deps separately, we can go back to pinning this.

var/spack/repos/builtin/packages/py-shapely/package.py
```diff
@@ -33,21 +33,32 @@ class PyShapely(PythonPackage):
...
+    depends_on("geos@3.3:", when="@:1")
     depends_on("py-pytest", type="test")
     depends_on("py-pytest-cov", type="test")
 
-    # https://github.com/Toblerity/Shapely/pull/891
+    # https://github.com/shapely/shapely/pull/891
     patch(
-        "https://github.com/Toblerity/Shapely/commit/98f6b36710bbe05b4ab59231cb0e08b06fe8b69c.patch?full_index=1",
+        "https://github.com/shapely/shapely/commit/98f6b36710bbe05b4ab59231cb0e08b06fe8b69c.patch?full_index=1",
         sha256="8583cdc97648277fa4faea8bd88d49e43390e87f697b966bd2b4290fba945ba0",
```
adamjstewart, line 50: Confirmed that checksum did not change

var/spack/repos/builtin/packages/py-shapely/package.py
```diff
@@ -78,4 +89,9 @@ def setup_dependent_build_environment(self, env, dependent_spec):
     @run_after("install")
     @on_package_attributes(run_tests=True)
     def test_install(self):
-        python("-m", "pytest")
+        # https://shapely.readthedocs.io/en/latest/installation.html#testing-shapely
```
adamjstewart, line 92: Latest version only supports testing of installed code, old versions only supported testing of source code

tldahlgren (commented):
Confirmed all of the py-shapely checksums are still valid.

tldahlgren (approved):

PR #31408, opened 2022-07-01
node-js: fix build on macOS 12

adamjstewart:
On macOS 10, `which` and `whereis` had the same output format:
```console
$ which libtool
/usr/bin/libtool
$ whereis libtool
/usr/bin/libtool
```
On macOS 12, the output format of `whereis` changed:
```console
$ which libtool
/usr/bin/libtool
$ whereis libtool
libtool: /usr/bin/libtool
```
This PR should correctly parse both formats. Tested on macOS 12.4.

PR #34478, opened 2022-12-12
Node.js: new versions, newer Python support, macOS fixes

adamjstewart:
All added versions successfully build on macOS 12.6.1 with Python 3.10.8 and Apple Clang 14.0.0.

@meyersbs discovered that our default version didn't support Python 3.10, so I fixed this.

### DIFF
--- a/var/spack/repos/builtin/packages/node-js/package.py
+++ b/var/spack/repos/builtin/packages/node-js/package.py
@@ -10,8 +10,7 @@ from spack.package import *
 
 
 class NodeJs(Package):
-    """Node.js is a JavaScript runtime built on Chrome's V8 JavaScript
-    engine."""
+    """Node.js is an open-source, cross-platform JavaScript runtime environment."""
 
     homepage = "https://nodejs.org/"
     url = "https://nodejs.org/dist/v13.5.0/node-v13.5.0.tar.gz"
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/node-js/package.py
```diff
@@ -57,9 +60,15 @@ class NodeJs(Package):
 
     # https://github.com/nodejs/node/blob/master/BUILDING.md#unix-and-macos
     depends_on("gmake@3.81:", type="build")
+    depends_on("python@3.6:3.11", when="@19.1:", type="build")
+    depends_on("python@3.6:3.10", when="@16.11:19.0", type="build")
+    depends_on("python@3.6:3.9", when="@16.0:16.10", type="build")
+    depends_on("python@2.7,3.5:3.8", when="@15", type="build")
+    depends_on("python@2.7,3.6:3.10", when="@14.18.2:14", type="build")
+    depends_on("python@2.7,3.5:3.8", when="@13.1:14.18.1", type="build")
+    depends_on("python@2.7,3.5:3.7", when="@12:13.0", type="build")
```
adamjstewart, line 69: This is actually relatively easy to confirm/update. Simply checkout the git repo and run the following bash script:
```bash
for tag in $(git tag -l | sort -Vr)
do
    echo
    echo $tag
    echo
    git checkout $tag BUILDING.md && grep Python BUILDING.md && git restore --staged BUILDING.md && git restore BUILDING.md
done
```
Older versions stored this info in README.md instead of BUILDING.md, but we don't have any versions that old.

var/spack/repos/builtin/packages/node-js/package.py
```diff
@@ -79,17 +88,30 @@ def setup_build_environment(self, env):
...
+            #
+            # /usr/bin/libtool
             process_pipe = subprocess.Popen(["which", "libtool"], stdout=subprocess.PIPE)
             result_which = process_pipe.communicate()[0].strip()
+
+            # Possible output formats:
+            #
+            # /usr/bin/libtool
+            # libtool: /usr/bin/libtool
+            # libtool: /usr/bin/libtool /Applications/Xcode.app/.../share/man/man1/libtool.1
```
adamjstewart, line 104: They keep changing the output format on me. This is a follow-up to #31408

var/spack/repos/builtin/packages/node-js/package.py
```diff
@@ -79,17 +88,30 @@ def setup_build_environment(self, env):
...
+            result_whereis_list = process_pipe.communicate()[0].strip().split()
+            if len(result_whereis_list) == 1:
+                result_whereis = result_whereis_list[0]
+            else:
+                result_whereis = result_whereis_list[1]
+
             assert result_which == result_whereis, (
-                "On OSX the system libtool must be used. Please"
-                "(temporarily) remove \n %s or its link to libtool from"
-                "path"
```
adamjstewart, line 92: Was missing spaces between lines

tldahlgren (commented):
Confirmed all new sha256.

alalazo (approved):

PR #34481, opened 2022-12-12
scons: fix signature for `install_args`

alalazo:
Hopefully the last issue falling out from #30738. @amd-toolchain-support This should fix the build of `amdlibm`.

### DIFF
--- a/lib/spack/spack/build_systems/scons.py
+++ b/lib/spack/spack/build_systems/scons.py
@@ -46,10 +46,10 @@ class SConsBuilder(BaseBuilder):
     phases = ("build", "install")
 
     #: Names associated with package methods in the old build-system format
-    legacy_methods = ("install_args", "build_test")
+    legacy_methods = ("build_test",)
 
     #: Same as legacy_methods, but the signature is different
-    legacy_long_methods = ("build_args",)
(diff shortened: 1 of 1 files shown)


tldahlgren (commented):
Should `MavenPackage` (e.g., `hadoop-xrootd`) and `fpm` be changed for `install_args` and `build_args` to pass the same arguments for consistency?

alalazo:
No for both. The point is exactly that - I did a breaking change to enforce consistency of the `*_args` methods in Scons and in a few other build systems early when writing #30738 (the vast majority accepts just `self`). Later we decided to be 100% backward compatible, including these inconsistencies, but I overlooked reverting some lines.

alalazo:
For reference, the method that have a different signature from the others of the same kind are the ones listed in the `legacy_long_methods` attributes in build system classes.

tldahlgren:
Okay.  Thanks for the clarification.

adamjstewart (approved):
This reminds me, we should get back to our discussion of whether functions should be `self`, `self, spec, prefix`, `self, pkg, spec, prefix`, or some combination of these.

alalazo:
@spackbot run pipeline

PR #34484, opened 2022-12-12
bugfix rmda-core external find library

bvanessen:
Added a more robust check for an external version of the library.
Included a guard to identify when the library gives no discernible version information and then to substitute with "unknown_ver" identifier.


bvanessen:
@rblake-llnl Here is a PR that makes the find command more robust.

becker33 (approved):

PR #34486, opened 2022-12-13
add_new_package: py-file-magic

Sinan81:
Note that this is a different package than `py-filemagic`

### DIFF
--- a/var/spack/repos/builtin/packages/py-file-magic/package.py
+++ b/var/spack/repos/builtin/packages/py-file-magic/package.py
@@ -1,0 +1,18 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyFileMagic(PythonPackage):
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-file-magic/package.py
```diff
@@ -0,0 +1,18 @@
...
+
+class PyFileMagic(PythonPackage):
+    """This library is a Python ctypes interface to libmagic"""
+
+    homepage = "https://pypi.org/project/file-magic/"
+    pypi     = "file-magic/file-magic-0.4.1.tar.gz"
+
+    version('0.4.1', sha256='a91d1483117f7ed48cd0238ad9be36b04824d57e9c38ea7523113989e81b9c53')
+
+    depends_on('py-setuptools', type='build')
```
adamjstewart:
```suggestion
    depends_on('py-setuptools@61:', type='build')
```

adamjstewart (changes requested):

Sinan81:
@spackbot fix style

adamjstewart (approved):

alalazo:
_One year and a half later_ Do you know where the sources are located for this package?

alalazo:
Ok, these seems to come from https://github.com/file/file directly.

PR #34440, opened 2022-12-10
package_base: Fix Python package tests

michaelkuhn:
Running `spack install --test=root py-nipype` results in the following error:
```
'super' object has no attribute 'test_log_file'
```
Replacing `test_log_file` with `test_install_log_path` (as seems to be intended) still causes the following error:
```
'NoneType' object has no attribute 'append'
```
This can be fixed by making sure that `test_failures` is not `None`.

PR #34487, opened 2022-12-13
Bugfix: support package.test as install-time callback

tldahlgren:
Addresses issue reported in #34440 

Is this sufficient to handle tests that require the package build environment?

TBD:
- Add unit test  (Update: I'm struggling to add a unit test to exercise the test method of the package (and ideally parent).)


tldahlgren:
@alalazo I'm having problems adding a unit test that can go through the builder to run the package's `test` method (ideally to include that of the parent class as well so similar to the `PythonPackage` import tests plus the package's tests.  I must be missing something.  Advice would be appreciated.

PR #34490, opened 2022-12-13
Allow spec.json files to be clearsigned, use transparent compression for *.json

haampie:
This PR does two things:

1. Reduce the number of requests to a mirror for metadata
2. Make Spack accept gzipped metadata (i.e. transmit fewer bytes)

It does *not* make any changes to how we create binary caches.
The idea is to allow this PR to be backported. 

---

Currently we have separate `spec.json` and `spec.json.sig` files. The
former is a pure spec, the latter is clearsigned spec. This is unfortunate,
cause metadata lookups now requires at most two requests per remote.

Further, update-index logic fetches all spec.json files, which can take
quite some bandwidth. We can drop ~85% of the size by compressing
those files. To avoid yet another lookup (spec.json.sig.gz etc) we can
use transparent decompression. So a file is decompressed when the
magic bytes signal it's gzip, otherwise it's treated as text.

This PR allows `spec.json` (and `index.json`) files to be any combination
of the cartesian product `[gzip compressed, uncompressed]` x 
`[clearsigned, unsigned]`.

The idea is to backport this to 0.19.1, since for 0.19.1 it is not a
behavioral change, but rather a bugfix for improved Spack 0.21 forward
compatibility.

Next steps after this PR:

- Backport to 0.19.1
- Start uploading spec.json.sig files both as spec.json.sig and
  spec.json. This ensure they work with any Spack version.
- Compress before pushing (if the filesize is reduced)
- Deprecate spec.json.sig in 0.20.
- After Spack 0.20 branches off, stop uploading *.sig, and stop checking
  the remote for *.sig files.
- Potentially we can also add a config option to disable checking for
  *.sig files on develop earlier.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -5,6 +5,7 @@
 
 import codecs
 import collections
+import gzip
 import hashlib
 import io
 import json
@@ -614,6 +615,28 @@ def read_buildinfo_file(prefix):
     return buildinfo
 
 
+def transparently_decompress_bytes(binary_stream):
+    """Wrap stream in a decompress wrapper if gzip compressed"""
+    # Get magic bytes
+    if isinstance(binary_stream, io.BytesIO):
+        # Not peekable... Alternatively io.BufferedReader(io.BytesIO(...))
+        # but to add yet another wrapper just to read two bytes that are
+        # already in memory... sigh.
+        magic = binary_stream.read(2)
+        binary_stream.seek(0)
+    else:
+        magic = binary_stream.peek(2)
+
+    # Verify magic
+    if magic.startswith(b"\x1f\x8b"):
+        return gzip.GzipFile(fileobj=binary_stream)
+    return binary_stream
+
+
+def transparently_decompress_bytes_to_string(binary_stream, encoding="utf-8"):
+    return codecs.getreader(encoding)(transparently_decompress_bytes(binary_stream))
+
+
 class BuildManifestVisitor(BaseDirectoryVisitor):
     """Visitor that collects a list of files and symlinks
     that can be checked for need of relocation. It knows how
@@ -966,8 +989,8 @@ def _specs_from_cache_aws_cli(cache_prefix):
     aws = which("aws")
 
     def file_read_method(file_path):
-        with open(file_path) as fd:
-            return fd.read()
+        with open(file_path, "rb") as f:
+            return transparently_decompress_bytes_to_string(f).read()
 
     tmpspecsdir = tempfile.mkdtemp()
     sync_command_args = [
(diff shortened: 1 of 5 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -1467,61 +1523,47 @@ def download_tarball(spec, unsigned=False, mirrors_for_spec=None):
...
-                    if tarball_stage:
-                        return {
-                            "tarball_stage": tarball_stage,
-                            "specfile_stage": local_specfile_stage,
-                            "signature_verified": signature_verified,
-                        }
+    # Only look for json.sig files as a last resort, since we don't generate them
+    # anymore.
+    verification_failure = False
+    for ext in ["json", "json.sig"]:
```
scottwittenburg: Since all the currently uploaded specs are `.json.sig`, is switching the order we look for them here going to force two trips for each spec/mirror until we start doing the next steps in the description (specifically, where we start uploading all spec files as both `.json` and `.json.sig`)?

Just checking if I understand.

lib/spack/spack/binary_distribution.py
```diff
@@ -613,6 +615,28 @@ def read_buildinfo_file(prefix):
...
+        # Not peekable... Alternatively io.BufferedReader(io.BytesIO(...))
+        # but to add yet another wrapper just to read two bytes that are
+        # already in memory... sigh.
+        magic = binary_stream.read(2)
+        binary_stream.seek(0)
+    else:
+        magic = binary_stream.peek(2)
+
+    # Verify magic
+    if magic.startswith(b"\x1f\x8b"):
```
scottwittenburg, line 631: Good thing you double-checked the solution given by ChatGPT :grin:

scottwittenburg (dismissed):
Looks great, thanks @haampie :+1: 

I just had one question, but feel free to merge regardless.

lib/spack/spack/binary_distribution.py
haampie: Thanks for pointing that out, not sure why I changed the order 😬 

Edit: Yup, it's probably because I extracted this from a commit where I did both download & uploading cleartext as *.json. I will revert this then!

lib/spack/spack/binary_distribution.py
```diff
@@ -1467,61 +1523,47 @@ def download_tarball(spec, unsigned=False, mirrors_for_spec=None):
...
+            specfile_stage = try_fetch(specfile_url)
+            if not specfile_stage:
+                continue
+
+            specfile_path = specfile_stage.save_filename
 
-                local_specfile_stage.destroy()
+            # If it is a clearsign file, we must verify it (unless disabled)
+            should_verify = not unsigned and is_clearsigned_file(specfile_path)
+            if should_verify and not try_verify(specfile_path):
```
haampie, line 1555: Hm, another issue is that this is not yet working, cause it's first clearsigned and then gzipped, so verification requires decompression. Should the order be reverted? Sign the compressed json?

haampie: Went with compress(sign(json)) so that we can swap the compression scheme without the need to re-sign the files.

scottwittenburg (approved):
Looks good to my eyeballs :laughing: thanks @haampie!  I notice the past several gitlab pipelines haven't built anything (as expected), but since this isn't actually changing what we push (still pushing `.spec.json.sig`, not compressing them yet), this isn't likely to break anything when run in the wild.  But it's good preparation for the next steps, thanks again :+1:

PR #34494, opened 2022-12-13
gromacs: enable linking with acfl FFT

annop-w:


marvinbernhardt (commented):
why the list joined with ';'? I think it should be ':' for CMake.

marvinbernhardt:
> why the list joined with ';'? I think it should be ':' for CMake.

Ok, I see [in the gromacs manual](https://manual.gromacs.org/documentation/current/install-guide/index.html#using-mkl) that ';' is probably correct.

marvinbernhardt (approved):

tldahlgren (approved):
LGTM

PR #34496, opened 2022-12-13
qmcpack: enable linking with armpl-gcc and acfl for BLAS and FFT

annop-w:


prckent:
Thank you.

var/spack/repos/builtin/packages/qmcpack/package.py
```diff
@@ -270,6 +270,16 @@ def cmake_args(self):
...
+        elif "^armpl-gcc" in spec:
+            args.append("-DBLA_VENDOR=Arm")
+            args.append("-DBLAS_LIBRARIES={0}".format(spec["armpl-gcc"].libs.joined(";")))
+            args.append("-DFFTW_LIBRARIES={0}".format(spec["armpl-gcc"].libs.joined(";")))
+            args.append("-DFFTW_INCLUDE_DIR={0}".format(spec["armpl-gcc"].headers.directories[0]))
+        elif "^acfl" in spec:
+            args.append("-DBLA_VENDOR=Arm")
+            args.append("-DBLAS_LIBRARIES={0}".format(spec["acfl"].libs.joined(";")))
+            args.append("-DFFTW_LIBRARIES={0}".format(spec["acfl"].libs.joined(";")))
+            args.append("-DFFTW_INCLUDE_DIR={0}".format(spec["acfl"].headers.directories[0]))
```
ye-luo, line 278: It is not good to have FFT and BLAS control mixed.
Without your change, doesn't ARMPL gets picked up by CMake?
By the way, for BLAS/LAPACK, Arm_mp is preferred.

annop-w: No, it does not. It is also more preferable to set them explicitly because ArmPL is shipped with all variants (i.e. lp64, ilp64, threaded) together. CMake might pick up lp64 non-threaded version by default although the user requests for ilp64_mp version for example. The default variant is threads=none for both acfl and armpl-gcc. It might be a good idea to change that default to threads=openmp, but that is for a separate commmit.

williamfgc:
@annop-w FYI, I built QMCPACK using `Arm_mp` on OLCF's [Wombat Ampere nodes](https://www.olcf.ornl.gov/olcf-resources/compute-systems/wombat/) and built and run. The `Arm_ilp64*` variants listed in [CMake doc](https://cmake.org/cmake/help/v3.25/module/FindBLAS.html#blas-lapack-vendors) crashed, the Arm folks were able to reproduce. As @ye-luo mentioned `Arm_mp` is preferred, I can test on Wombat using this PR. Thanks, BTW!

var/spack/repos/builtin/packages/qmcpack/package.py
annop-w: I also see no reason to separate BLAS and FFT libraries settings in this case since one cannot do, e.g.,

spack install qmcpack ^armpl-gcc ^fftw

since both armpl-gcc and fftw provides the virtual fftw-api.

ye-luo: `CMake might pick up lp64 non-threaded version by default although the user requests for ilp64_mp version for example.`
That sounds like a bug in CMake. `Arm_ilp64` `Arm_ilp64_mp` `Arm` `Arm_mp`, if we request one, I would expect it either found the library or error out.

annop-w: I meant, when BLA_VENDOR is not set. CMake picks up the correct version of the libraries, if BLA_VENDOR is set.

In fact, in this commit setting BLA_VENDOR to Arm is redundant. I'd better remove it as it might be confusing. I'll push a new commit. Thanks.

ye-luo: It remains unclear to me what spack virtual packages exactly provide (include path, link path...). headerless fortrain libraries may behave one way and C library can be quite different. QMCPACK may allow mixed FFTW if the libraries link interface may distinguish one from another.

ye-luo:
> @annop-w FYI, I built QMCPACK using `Arm_mp` on OLCF's [Wombat Ampere nodes](https://www.olcf.ornl.gov/olcf-resources/compute-systems/wombat/) and built and run. The `Arm_ilp64*` variants listed in [CMake doc](https://cmake.org/cmake/help/v3.25/module/FindBLAS.html#blas-lapack-vendors) crashed, the Arm folks were able to reproduce. As @ye-luo mentioned `Arm_mp` is preferred, I can test on Wombat using this PR. Thanks, BTW!

Do we have OpenMP control exposed in spack? When OpenMP is on, we should turn on `Arm_mp` othersise `Arm` only.
But that is a separate topic from this PR. For now, `Arm` should be fine.

var/spack/repos/builtin/packages/qmcpack/package.py
annop-w: Sure. QMCPACK probably allows mixing fftw and blas vendors.

annop-w:
> > @annop-w FYI, I built QMCPACK using `Arm_mp` on OLCF's [Wombat Ampere nodes](https://www.olcf.ornl.gov/olcf-resources/compute-systems/wombat/) and built and run. The `Arm_ilp64*` variants listed in [CMake doc](https://cmake.org/cmake/help/v3.25/module/FindBLAS.html#blas-lapack-vendors) crashed, the Arm folks were able to reproduce. As @ye-luo mentioned `Arm_mp` is preferred, I can test on Wombat using this PR. Thanks, BTW!
> 
> Do we have OpenMP control exposed in spack? When OpenMP is on, we should turn on `Arm_mp` othersise `Arm` only. But that is a separate topic from this PR. For now, `Arm` should be fine.

You can do 

spack install qmcpack ^armpl-gcc threads=openmp

As I mentioned the line to set BLA_VENDOR in this patch is redundant.

tldahlgren:
Added commenters as assignees so we can see that the PR is getting non-spackbot "feedback" from the main list.

williamfgc:
> You can do spack install qmcpack ^armpl-gcc threads=openmp

A first attempt on Wombat results in an error. I'll check with system admins if it's an issue with the installed gcc modules.

```
$ ./bin/spack install qmcpack ^armpl-gcc threads=openmp
==> Installing armpl-gcc-22.1_gcc-11.2-zeeogzzwaikegbiraoshbje7dze4ubpw
==> No binary for armpl-gcc-22.1_gcc-11.2-zeeogzzwaikegbiraoshbje7dze4ubpw found: installing from source
==> Using cached archive: /autofs/nccs-svm1_envoy_od/wfg/spack-qmcpack/spack/var/spack/cache/_source-cache/archive/24/24f9f4496e41c2314d4ace25b6e3d63127bd586ff7bdd8a732471cbc65a8023e.tar
==> No patches needed for armpl-gcc
==> armpl-gcc: Executing phase: 'install'
==> Error: SpackError: Only compatible with GCC.


/autofs/nccs-svm1_envoy_od/wfg/spack-qmcpack/spack/var/spack/repos/builtin/packages/armpl-gcc/package.py:197, in install:
        195    def install(self, spec, prefix):
        196        if self.compiler.name != "gcc":
  >>    197            raise spack.error.SpackError(("Only compatible with GCC.\n"))
        198
        199        armpl_version = "{}".format(spec.version.up_to(3)).split("_")[0]
        200        exe = Executable("./arm-performance-libraries_{0}_{1}.sh".format(armpl_version, get_os()))

See build log for details:
  /tmp/wfg/spack-stage/spack-stage-armpl-gcc-22.1_gcc-11.2-zeeogzzwaikegbiraoshbje7dze4ubpw/spack-build-out.txt
```

annop-w:
> > You can do spack install qmcpack ^armpl-gcc threads=openmp
> 
> A first attempt on Wombat results in an error. I'll check with system admins if it's an issue with the installed gcc modules.
> 
> ```
> $ ./bin/spack install qmcpack ^armpl-gcc threads=openmp
> ==> Installing armpl-gcc-22.1_gcc-11.2-zeeogzzwaikegbiraoshbje7dze4ubpw
> ==> No binary for armpl-gcc-22.1_gcc-11.2-zeeogzzwaikegbiraoshbje7dze4ubpw found: installing from source
> ==> Using cached archive: /autofs/nccs-svm1_envoy_od/wfg/spack-qmcpack/spack/var/spack/cache/_source-cache/archive/24/24f9f4496e41c2314d4ace25b6e3d63127bd586ff7bdd8a732471cbc65a8023e.tar
> ==> No patches needed for armpl-gcc
> ==> armpl-gcc: Executing phase: 'install'
> ==> Error: SpackError: Only compatible with GCC.
> 
> 
> /autofs/nccs-svm1_envoy_od/wfg/spack-qmcpack/spack/var/spack/repos/builtin/packages/armpl-gcc/package.py:197, in install:
>         195    def install(self, spec, prefix):
>         196        if self.compiler.name != "gcc":
>   >>    197            raise spack.error.SpackError(("Only compatible with GCC.\n"))
>         198
>         199        armpl_version = "{}".format(spec.version.up_to(3)).split("_")[0]
>         200        exe = Executable("./arm-performance-libraries_{0}_{1}.sh".format(armpl_version, get_os()))
> 
> See build log for details:
>   /tmp/wfg/spack-stage/spack-stage-armpl-gcc-22.1_gcc-11.2-zeeogzzwaikegbiraoshbje7dze4ubpw/spack-build-out.txt
> ```

Looks to me like Spack couldn't find your gcc. Perhaps try to module load your gcc and then `spack compiler find`.
After that check with `spack compilers` and your gcc should show up.

williamfgc:
> Looks to me like Spack couldn't find your gcc. 

```
$ ./bin/spack compilers
==> Available compilers
-- arm centos8-aarch64 ------------------------------------------
arm@22.0.2.1776

-- gcc centos8-aarch64 ------------------------------------------
gcc@11.1.0  gcc@8.3.1
```

I modified it a bit to get a better error message:
```
==> Error: SpackError: Only compatible with GCC. Found arm


/autofs/nccs-svm1_envoy_od/wfg/spack-qmcpack/spack/var/spack/repos/builtin/packages/armpl-gcc/package.py:197, in install:
        195    def install(self, spec, prefix):
        196        if self.compiler.name != "gcc":
  >>    197            raise spack.error.SpackError(("Only compatible with GCC. Found " + self.compiler.name + "\n"))
```

It looks like `arm` is being picked up by `spack.compiler.name` and it's not supported by default. Not sure if this is a bug in spack. Thoughts?

annop-w:
> > Looks to me like Spack couldn't find your gcc.
> 
> ```
> $ ./bin/spack compilers
> ==> Available compilers
> -- arm centos8-aarch64 ------------------------------------------
> arm@22.0.2.1776
> 
> -- gcc centos8-aarch64 ------------------------------------------
> gcc@11.1.0  gcc@8.3.1
> ```
> 
> I modified it a bit to get a better error message:
> 
> ```
> ==> Error: SpackError: Only compatible with GCC. Found arm
> 
> 
> /autofs/nccs-svm1_envoy_od/wfg/spack-qmcpack/spack/var/spack/repos/builtin/packages/armpl-gcc/package.py:197, in install:
>         195    def install(self, spec, prefix):
>         196        if self.compiler.name != "gcc":
>   >>    197            raise spack.error.SpackError(("Only compatible with GCC. Found " + self.compiler.name + "\n"))
> ```
> 
> It looks like `arm` is being picked up by `spack.compiler.name` and it's not supported by default. Not sure if this is a bug in spack. Thoughts?

I'm not entirely sure how spack picks the default compiler. You could try to be specific, e.g. `spack install qmcpack ^armpl-gcc %gcc@11.1.0` or `spack install qmcpack ^acfl %arm`.

annop-w:
@williamfgc @ye-luo Do you have any more suggestion for this PR ? or is this ready for merging ? Thanks.

var/spack/repos/builtin/packages/qmcpack/package.py
ye-luo: My initial request remains.
1) please do not mix BLAS and FFTW in the current giant if block
2) Control BLAS using BLA_VENDOR

annop-w: 1. Sure, I can move BLAS settings to a separate if-block. 
2. There is no need to set BLA_VENDOR. The correct variant will be fetched by `spec["armpl-gcc"].libs`. To use openmp enabled version the user should type `spack install qmcpack ^armpl-gcc threads=openmp` for example.

annop-w:
@ye-luo Could you please take a look at this once more ? Thank you.

annop-w:
@tldahlgren I have addressed all the comments. Do you think this is ready to go in ? Thank you.

ye-luo (approved):

var/spack/repos/builtin/packages/qmcpack/package.py
tldahlgren: > I also see no reason to separate BLAS and FFT libraries settings in this case since one cannot do, e.g.,
> 
> spack install qmcpack ^armpl-gcc ^fftw
> 
> since both armpl-gcc and fftw provides the virtual fftw-api.

Is this worth adding a `conflicts()` e.g., `conflicts("^armpl-gcc", when="^fftw")`?

tldahlgren (commented):
LGTM.  Just a minor question related to the resolved discussion (i.e., "since both armpl-gcc and fftw provides the virtual fftw-api"):

Is it worth adding a conflict e.g., `conflicts("^armpl-gcc", when="^fftw")`?

annop-w:
I think there is no need to add that conflict because spack won't allow `^armpl-gcc ^fftw` anyway.

tldahlgren (approved):

PR #34497, opened 2022-12-13
py-sqlalchemy: add 1.4.45

manuelakuhn:
https://github.com/sqlalchemy/sqlalchemy/tree/rel_1_4_45

I first wanted to replace `import_modules` with `skip_modules` but since pytest is now a internal dependency for tests anyway this can be removed completely.

`spack install --test=root` only runs with the fix in #34440

### DIFF
--- a/var/spack/repos/builtin/packages/py-sqlalchemy/package.py
+++ b/var/spack/repos/builtin/packages/py-sqlalchemy/package.py
@@ -11,32 +11,9 @@ class PySqlalchemy(PythonPackage):
 
     homepage = "http://www.sqlalchemy.org/"
     pypi = "SQLAlchemy/SQLAlchemy-1.3.9.tar.gz"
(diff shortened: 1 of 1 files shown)


tldahlgren (commented):
Confirmed the sha256.

Are the `import_modules` automatically picked up by `PythonPackage` for stand-alone/install testing or do they no longer apply?

adamjstewart (approved):
Another WIP PR also updates this package, can't remember which. Don't really care which gets merged first.

adamjstewart:
> Are the `import_modules` automatically picked up by `PythonPackage` for stand-alone/install testing or do they no longer apply?

They are, but sometimes they include things that users aren't expected to import (usually tests) or files with optional deps that aren't always installed. You can either override the auto-detected list using `import_modules`, or more recently, you can skip modules you don't want to import using `skip_modules`.

PR #34498, opened 2022-12-13
Revert "Revert "Use `urllib` handler for `s3://` and `gs://`, improve `url_exists` through HEAD requests (#34324)""

haampie:
revert-revert-final-FINAL-2.docx

This reverts commit 8035eeb36d5068fcbae613e51dd13cb1ae9f4888.

It also now removes a dubious initial `HEAD` request in `read_from_url`, which used to prevent a follow-up `GET` request, but was broken 4 years ago in fce1c4133f104c1a09b09909f12bed81a83bb46a. The idea there was to prevent a GET request to potentially large files; this is typically not an issue since attachments are often downloaded only once you start reading from the stream. So, let's just remove.

### DIFF
--- a/lib/spack/spack/util/web.py
+++ b/lib/spack/spack/util/web.py
@@ -18,7 +18,7 @@ import traceback
 import urllib.parse
 from html.parser import HTMLParser
 from urllib.error import URLError
-from urllib.request import Request, urlopen
+from urllib.request import HTTPSHandler, Request, build_opener
 
 import llnl.util.lang
 import llnl.util.tty as tty
@@ -27,6 +27,8 @@ from llnl.util.filesystem import mkdirp, rename, working_dir
 import spack
 import spack.config
 import spack.error
+import spack.gcs_handler
+import spack.s3_handler
 import spack.url
 import spack.util.crypto
 import spack.util.gcs as gcs_util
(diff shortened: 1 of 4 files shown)


haampie:
@spackbot run pipeline

lib/spack/spack/util/web.py
```diff
@@ -60,86 +84,33 @@ def handle_starttag(self, tag, attrs):
...
-
-    content_type = None
-    is_web_url = url_scheme in ("http", "https")
-    if accept_content_type and is_web_url:
-        # Make a HEAD request first to check the content type.  This lets
-        # us ignore tarballs and gigantic files.
-        # It would be nice to do this with the HTTP Accept header to avoid
-        # one round-trip.  However, most servers seem to ignore the header
-        # if you ask for a tarball with Accept: text/html.
-        req.get_method = lambda: "HEAD"
```
scottwittenburg, line 115: In the commit you said broke stuff, is it the check `if accept_content_type:` that was the problem?  I think that was just to avoid always checking for `text/html`, but I'm not following what the problem was.  At any rate, if you say there's no need for this anymore anyway, I'm fine with this particular change.

scottwittenburg (approved):
Other than the bit I commented on (the bit that's not just reverting a revert), this all looks as good to me today as it did last week :laughing:  Thanks @haampie!

lib/spack/spack/util/web.py
haampie: How it's supposed to work: in certain cases (spack checksum), it should only send GET requests if the content-type is something specific, which is obtained through a cheap, initial HEAD request. It was broken cause the GET request was no longer conditional. In the general case it's not necessary to send a HEAD request first, and I think this is how this function is used almost always -- this was not broken.

PR #34499, opened 2022-12-13
py-tifffile: add 2022.10.10

manuelakuhn:
https://github.com/cgohlke/tifffile/tree/v2022.10.10


tldahlgren (commented):
Confirmed the sha256.

adamjstewart (approved):

PR #34500, opened 2022-12-13
error messages: wip condition chaining

becker33:
This is work to create chains of causation for error messages.

The current implementation is only completed for two of the many errors presented by the concretizer. The rest will need to be filled out over time, but this demonstrates the capability.

The basic idea is to associate conditions in the solver with one another in causal relationships, and to associate errors with the proximate causes of their facts in the condition graph. Then we can construct causal trees to explain errors, which will hopefully present users with useful information to avoid the error or report issues.

Example:

```
$ time spack solve --timers hdf5 ^cmake@3.0.1
==> Error: Version '3.0.1' of cmake does not satisfy '@3.12:'
required because hdf5 ^cmake@3.0.1 is provided as input spec 
required because hdf5 depends on cmake@3.12: 
  required because hdf5 ^cmake@3.0.1 is provided as input spec 

real	0m7.541s
user	0m6.581s
sys	0m0.270s
```

In addition to fleshing out the causes of other errors, I would like to find a way to associate different components of the error messages with different causes. In this example it's pretty easy to infer which part is which, but I'm not confident that will always be the case.

### DIFF
--- a/lib/spack/spack/solver/concretize.lp
+++ b/lib/spack/spack/solver/concretize.lp
@@ -425,25 +425,15 @@ external(Package) :- attr("external_spec_selected", Package, _).
 
 % determine if an external spec has been selected
 attr("external_spec_selected", Package, LocalIndex) :-
-    external_conditions_hold(Package, LocalIndex),
+    attr("external_conditions_hold", Package, LocalIndex),
     attr("node", Package),
     not attr("hash", Package, _).
-
-external_conditions_hold(Package, LocalIndex) :-
-    possible_external(ID, Package, LocalIndex), condition_holds(ID).
 
 % it cannot happen that a spec is external, but none of the external specs
 % conditions hold.
 error(100, "Attempted to use external for '{0}' which does not satisfy any configured external spec", Package)
  :- external(Package),
-    not external_conditions_hold(Package, _).
-
-#defined possible_external/3.
-#defined external_spec_index/3.
-#defined external_spec_condition/3.
-#defined external_spec_condition/4.
-#defined external_spec_condition/5.
-#defined external_spec_condition/6.
+    not attr("external_conditions_hold", Package, _).
 
 %-----------------------------------------------------------------------------
 % Config required semantics
@@ -653,7 +643,7 @@ variant_default_not_used(Package, Variant, Value)
 external_with_variant_set(Package, Variant, Value)
  :- attr("variant_value", Package, Variant, Value),
     condition_requirement(ID, "variant_value", Package, Variant, Value),
-    possible_external(ID, Package, _),
+    imposed_constraint(ID, "external_conditions_hold", Package, _),
     external(Package),
     attr("node", Package).
 
--- a/lib/spack/spack/solver/concretize.lp
+++ b/lib/spack/spack/solver/concretize.lp
@@ -182,9 +182,8 @@ condition_holds(ID) :-
   attr(Name, A1, A2, A3)     : condition_requirement(ID, Name, A1, A2, A3);
   attr(Name, A1, A2, A3, A4) : condition_requirement(ID, Name, A1, A2, A3, A4).
 
-% condition_holds(ID) implies all imposed_constraints, unless do_not_impose(ID)
-% is derived. This allows imposed constraints to be canceled in special cases.
-impose(ID) :- condition_holds(ID), not do_not_impose(ID).
+% condition_holds(ID) implies all imposed_constraints.
+impose(ID) :- condition_holds(ID).
 
 % conditions that hold impose constraints on other specs
 attr(Name, A1)             :- impose(ID), imposed_constraint(ID, Name, A1).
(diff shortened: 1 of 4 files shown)


becker33:
While I haven't formally tested the performance, I'm seeing concretization times on my machine around 5.73s for `spack solve hdf5`, and times roughly 5.46s for the same command on develop, so I'm optimistic that the performance will be acceptable overall.

The one concern is that as I add causal details for more error messages, performance may degrade further.

becker33:
@spackbot fix style

becker33:
Performance update: As of the most recent commit I am seeing performance degradation of ~12% compared to develop. It does appear that each message this is applied to affects performance.

lib/spack/spack/solver/concretize.lp
```diff
@@ -196,6 +218,28 @@ attr(Name, A1, A2, A3, A4) :- impose(ID), imposed_constraint(ID, Name, A1, A2, A
...
+condition_cause(Effect, Cause) :-
+  condition_holds(Effect), condition_holds(Cause),
+  attr(Name, A1, A2, A3),
+  condition_requirement(Effect, Name, A1, A2, A3),
+  imposed_constraint(Cause, Name, A1, A2, A3).
+condition_cause(Effect, Cause) :-
+  condition_holds(Effect), condition_holds(Cause),
+  attr(Name, A1, A2, A3, A4),
+  condition_requirement(Effect, Name, A1, A2, A3, A4),
+  imposed_constraint(Cause, Name, A1, A2, A3, A4).
```
alalazo: Is this mapping complete? As far as I understand it maps imposed constraints (the cause) to condition requirements (the effect) which are exactly the same. So my understanding is that it would catch:
```
% cause
imposed_constraint(Cause,"node_version_satisfies","amdblis","3.0.1:").
% effect
condition_requirement(Effect,"node_version_satisfies","amdblis","3.0.1:").
```
but miss:
```
% cause
imposed_constraint(Cause,"node_version_satisfies","amdblis","3.0.1:").
% effect
condition_requirement(Effect,"node_version_satisfies","amdblis","2.0.0:").
```
? Is that correct, or am I missing some steps?

alalazo (commented):
Just started the review. An initial thought I have is that, if this has effects on performance, we might add it to a separate `*.lp` file and activate that part upon user request. Other than that I have a comment so far. I'll finish this by today.

alalazo:
I did a timing on these specs:

- [radiuss.txt](https://github.com/spack/spack/files/10238583/radiuss.txt)

The results are:

- [radiuss.develop.csv](https://github.com/spack/spack/files/10238584/radiuss.develop.csv) (on develop)
- [radiuss.error.csv](https://github.com/spack/spack/files/10238587/radiuss.error.csv) (on this PR)

Histogram for `develop`:

radiuss_develop

and for this PR:

radiuss_error

It seems that for large specs like `lbann` both grounding and solving times are doubled.

@becker33 Can you reproduce these timings, qualitatively? I used:

* **Spack:** 0.20.0.dev0 (9fb099a9b7addc1669c0853b7023ab449860456d)
* **Python:** 3.8.10
* **Concretizer:** clingo

alalazo (commented):
At a high-level I really like the idea. I think though it might be incomplete in some circumstances when reconstructing the chain of causation (see https://github.com/spack/spack/pull/34500#pullrequestreview-1219265029) and it seems to slow-down the solve in a non-negligible way. 

Would it be easy / possible to structure this as an extra part of a logic program that may be appended to the problem on demand, and that comes with a different error handler?

becker33:
There are 4 components to this PR as I see it, and they have different levels of separability.

1. The logic to handle errors that include causes in asp.py. I don't think this requires any work to separate, because it will transparently do nothing if there are no causes.
2. The new error messages that include causal information. These can easily be stored in a separate file
3. The new logic to associate causal chains of conditions. I think this can easily be stored in a separate file as well.
4. Changes to the logic program to make the causal chains closer to completeness. I don't think these can be easily separated, but I also doubt they're the cause of the slowdowns. Definitely worth testing those on their own for performance.

becker33:
The causal chains are definitely incomplete. The case you listed is one, but a simpler case is that it can track through chains of real dependencies, but will lose the causal chain through a virtual dependency.

If we continue down this route, we will want to make the causal chaining more robust in the long term. It may or may not be feasible to implement the missing causations as ad-hoc additional rules alongside the generic rules for causation. For example, I believe this rule will close the gap on virtual dependencies.

```
condition_cause(Effect, Cause)
  :- condition_holds(Effect), condition_holds(Cause),
     condition_requirement(Effect, "node", Provider),
     imposed_condition(Cause, "virtual_node", Virtual),
     provider(Virtual, Provider).
```

Writing a similar rule for the version example you gave will be more challenging.

becker33:
@alalazo updated this so that the slower, more accurate errors are only triggered in debug mode.

We may want to make that a separate flag, instead of piggybacking on debug, but this seemed like a place to start.

I haven't done anything as extensive as your testing, but it looks to be performance negligible in the base case, performance hit when you want the error messages now.

becker33:
@spackbot fix style

becker33:
Closing in deference to #40173

PR #34501, opened 2022-12-13
bacio: Patch VERSION file for v2.4.1

alexrichert:
## Description
In version 2.4.1, bacio's VERSION file, which is used to generate the cmake configuration files, contains 2.4.0. This can cause bacio to not be findable through cmake.

## Summary of changes
This PR adds a patch using filter_file to update the version when installing bacio@2.4.1. There do not seem to be similar issues for other bacio releases.


alexrichert:
Going to reissue PR with other account.

PR #34503, opened 2022-12-13
Add variant to libpng to allow static build

alexrichert:
This PR adds a variant "shared" to libpng to give the option of building the static lib only.


alexrichert:
@dmentock @rice31 @scheibelp @ax3l @alalazo Any volunteers to help maintain libpng? I'm happy to add myself as well but ideally would like to find maintainers with past experience with the package.

alexrichert:
Going to reissue PR with other account.

PR #34505, opened 2022-12-13
gitlab ci: more resources for slow builds

zackgalbreath:
The goal of this change is to speed up our GitLab CI pipelines.

I arrived at this list by querying the GitLab database for packages that averaged more than 20 minutes per build since November 1. After that, I performed some subsequent queries to figure out what stacks have built these packages in the past.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml
@@ -285,6 +285,7 @@ spack:
 
       - match:
           - cuda
+          - dealii
           - dray
           - dyninst
           - ginkgo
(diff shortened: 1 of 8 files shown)


share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml
```diff
@@ -294,12 +295,22 @@ spack:
           - magma
           - mfem
           - mpich
+          - nvhpc
```
haampie, line 298: nvhpc doesn't build anything. it downloads, decompresses, extracts, moves, tarballs, compresses and uploads, everything is single theaded.

scottwittenburg:
@zackgalbreath General comment: no need to put any abstract specs in the match list after "@:", as that matches everything and is the catch-all when match behavior is `first`.

share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml
scottwittenburg: @haampie I don't know if I get your point here.  Are you saying don't run `nvhpc` in this resource group?

haampie: It doesn't run faster if you throw more cores at it. Only compression might go faster then, but we don't have threaded compression right now

scottwittenburg (approved):

PR #34506, opened 2022-12-13
nwchem: restricting current versions to python@3.9 at latest

snehring:
Python 3.10 removed the old parser, this looks for headers that are part of that. It looks like this will be fixed in the next release.

### DIFF
--- a/var/spack/repos/builtin/packages/nwchem/package.py
+++ b/var/spack/repos/builtin/packages/nwchem/package.py
@@ -42,7 +42,7 @@ class Nwchem(Package):
     depends_on("mpi")
     depends_on("scalapack")
     depends_on("fftw-api")
-    depends_on("python@3:", type=("build", "link", "run"))
+    depends_on("python@3:3.9", type=("build", "link", "run"), when="@:7.0.2")
 
     def install(self, spec, prefix):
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
LGTM

var/spack/repos/builtin/packages/nwchem/package.py
```diff
@@ -42,7 +42,7 @@ class Nwchem(Package):
     depends_on("mpi")
     depends_on("scalapack")
     depends_on("fftw-api")
```
bernhardkaindl, line 44: When trying to build it without fftw installed in the host OS, I get missing lib for `-lfftw3`, so I propose:
```suggestion
    depends_on("fftw")
    depends_on("fftw-api")
```

snehring: I didn't run into this yesterday, and I'm building in a very sparse container. What've you got setup for providers for fftw-api?

bernhardkaindl: @snehring: Somehow `depends_on("fftw-api")` didn't add `fftw` into the spec in that install.

It's unrelated to this PR and a completely different issue.

PR #34513, opened 2022-12-13
odc: add v1.4.5

eugeneswalker:
@skosukhin

### DIFF
--- a/var/spack/repos/builtin/packages/odc/package.py
+++ b/var/spack/repos/builtin/packages/odc/package.py
@@ -14,6 +14,7 @@ class Odc(CMakePackage):
 
     maintainers = ["skosukhin"]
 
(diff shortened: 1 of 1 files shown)


skosukhin:
@eugeneswalker are you sure that the new version does not need more recent versions of `eckit`, `cmake`, etc.?

eugeneswalker:
> @eugeneswalker are you sure that the new version does not need more recent versions of `eckit`, `cmake`, etc.?

I am not sure what constraints might be unique to `odc@1.4.5` as compared to `@1.3.0` - no. I know that I was able to build the newer version with a fairly default concretization (i.e. using newest versions of eckit and cmake) but not detailed specification of the constraints on the dependencies. I reviewed the README here: https://github.com/ecmwf/odc and did not see detailed version constraints. Do you know?

skosukhin (approved):

wspear (approved):

PR #34479, opened 2022-12-12
Update paramiko

charmoniumQ:
2.12.0 includes an important bugfixing commit over 2.7.1: https://github.com/paramiko/paramiko/commit/81064206bf3cec2ca4372257ff138481e1227b91

PR #34466, opened 2022-12-12
Add nf-core-tools, and three remainder deps

marcodelapierre:
Key changes:
* Added nf-core-tools recipe, and its large python dependency tree (https://nf-co.re/tools/ ; https://pypi.org/project/nf-core/)
* Renamed tower-agent and tower-cli to nf-tower-agent and nf-tower-cli
* Added new package versions for nf-tower-cli and nextflow

Note I have tested the target new tool, `nf-core-tools`, for some functionalities, against a standard pip installation.

PR #34516, opened 2022-12-14
Edits to 8x existing recipes, mostly new versions, plus two dependency fixes

marcodelapierre:
Second chunk of updates from #34466.

These small changes are needed, or were figured out, in preparation for the addition of the nf-core-tools package (see PR above).

@adamjstewart 

Closes #34479

### DIFF
--- a/var/spack/repos/builtin/packages/py-click-repl/package.py
+++ b/var/spack/repos/builtin/packages/py-click-repl/package.py
@@ -11,7 +11,7 @@ class PyClickRepl(PythonPackage):
     homepage = "https://github.com/click-contrib/click-repl"
     pypi = "click-repl/click-repl-0.1.6.tar.gz"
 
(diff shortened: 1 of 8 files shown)


var/spack/repos/builtin/packages/py-click-repl/package.py
```diff
@@ -11,7 +11,7 @@ class PyClickRepl(PythonPackage):
     homepage = "https://github.com/click-contrib/click-repl"
     pypi = "click-repl/click-repl-0.1.6.tar.gz"
 
-    version("0.2.0", sha256="b0cac32a625c24cd1414cc323e314a79278e2310e41596a6e27997e1c9f99e72")
+    version("0.2.0", sha256="cd12f68d745bf6151210790540b4cb064c7b13e571bc64b6957d98d120dacfd8")
```
adamjstewart, line 14: Looks like this old checksum must have come from the GitHub tarball instead of the PyPI tarball, good catch

adamjstewart (approved):
For future reference, this should technically be 8 different PRs since none of the changes in one file require the changes in another file

PR #34521, opened 2022-12-14
py-archspec fails to build

finkandreas:
Commit https://github.com/spack/spack/commit/30c9ff50dd048573e3cd7a63a8ab9c05c0bee339 broke the package `py-archspec` because `PythonPackage.build_directory` was removed.
Potentially this commit broke also other packages, but at this point only `py-archspec` failed to build in my environment.

Maybe the fix is not in line with development guidelines, so please review and adapt to correct guidelines.


bernhardkaindl (approved):
Good, this is what other packages (430 use self.stage.source_path) already do.

PR #34522, opened 2022-12-14
add fixing prb with netlib-scalpack linking

likask:
Fix fortran MPI linking. Patch is straightforward, adds ${MPI_Fortran_LIBRARIES} to target_link_libraries.

I tested in in aarch64, monterey. However error fix should be working for other systems.


likask:
This was the wrong base.

PR #34523, opened 2022-12-14
mirrors: distinguish between source/binary mirror; simplify schema

haampie:
This PR allows the following formats for `mirrors.yaml` in a way that's not breaking:

```yaml
# simplest form
mirrors:
  name: <url>
```

```yaml
# additional props for both fetching and pushing
mirrors:
  name:
    url: s3://xyz
    access_pair: [username, password]
```

```yaml
# extra props just for pushing
mirrors:
  name:
    url: http://xyz
    push:
      url: s3://xyz
      access_pair: [x, y]
```

```yaml
# specify credentials just for pushing
mirrors:
  name:
    url: s3://bucket
    push:
      access_pair: [username, password]
```

The properties under `push` and `fetch` override the top-level entries.

On top of that, two new properties are added to indicate the mirror type (e.g.
mirror.spack.io is a source mirror, not a binary cache)

```yaml
mirrors:
  spack-public:
    source: true
    binary: false
    url: https://mirror.spack.io
```

This is used to avoid redundant requests to `mirror.spack.io` during install.

It also adds a command that can be used to conveniently modify
the mirror config, in a way that's more intuitive than `spack config add mirrors:...`

```sh
# set the top-level url / endpoint attributes in mirrors:<name>:...
$ spack mirror set <name> --url <url> --s3-endpoint <endpoint> ...

# set the push url and endpoint attribute in mirrors:<name>:push:...
$ spack mirror set <name> --push --url <url> --s3-endpoint <endpoint> ...

# set the fetch url and endpoint attribute in mirrors:<name>:fetch:...
$ spack mirror set <name> --fetch --url <url> --s3-endpoint <endpoint> ...

# toggle a mirror as a source / binary mirror
$ spack mirror set <name> --binary --no-source
```

this ensures that config remains minimal, so for example given

```yaml
mirrors:
  name: https://example.com
```

if you run `spack mirror set name --push --s3-endpoint https://helloworld` you get

```yaml
mirrors:
  name:
    url: https://example.com
    push:
      endpoint_url: https://helloworld
```


etc/spack/defaults/mirrors.yaml
```diff
@@ -1,2 +1,4 @@
 mirrors:
-  spack-public: https://mirror.spack.io
+  spack-public:
+    binary: false
```
alalazo, line 3: :+1:

lib/spack/spack/binary_distribution.py
```diff
@@ -2296,7 +2295,7 @@ def download_buildcache_entry(file_descriptions, mirror_url=None):
         mirror_root = os.path.join(mirror_url, _build_cache_relative_path)
         return _download_buildcache_entry(mirror_root, file_descriptions)
 
-    for mirror in spack.mirror.MirrorCollection().values():
+    for mirror in spack.mirror.MirrorCollection(binary=True).values():
```
alalazo, line 2298: (minor, comment) Given the amount of times we do:
```python
spack.mirror.MirrorCollection(binary=True).values()
```
I wonder if we should have a convenience wrapper:
```python
def binary_mirrors():
   return MirrorCollection(binary=True).values()
```
and use that in this module.

lib/spack/spack/schema/mirrors.py
```diff
@@ -6,29 +6,55 @@
 """Schema for mirrors.yaml configuration file.
 
 .. literalinclude:: _spack_root/lib/spack/spack/schema/mirrors.py
+   :lines: 12-69
 """
 
+#: Common properties for connection specification
+connection = {
+    "url": {"type": "string"},
+    # todo: replace this with named keys "username" / "password" or "id" / "secret"
```
alalazo, line 15: How do we plan to deprecate this?

lib/spack/spack/mirror.py
```diff
@@ -71,36 +63,24 @@ class Mirror:
...
-        return sjson.dump(self.to_dict(), stream)
-
-    def to_yaml(self, stream=None):
-        return syaml.dump(self.to_dict(), stream)
-
     @staticmethod
     def from_yaml(stream, name=None):
-        data = syaml.load(stream)
-        return Mirror.from_dict(data, name)
+        return Mirror(syaml.load(stream), name)
```
alalazo, line 72: This was not introduced in this PR, but I think we ought to be consistent when treating JSON and YAML - so either catch and re-raise in both or don't catch in either.

lib/spack/spack/mirror.py
```diff
@@ -111,165 +91,203 @@ def from_url(url: str):
...
-                ")",
-            )
-        )
+    def display(self, max_len=0):
+        fetch, push = self.fetch_url, self.push_url
+        # don't print the same URL twice
+        url = fetch if fetch == push else f"fetch: {fetch} push: {push}"
+        source = "s" if self.source else " "
+        binary = "b" if self.binary else " "
+        print(f"{self.name: <{max_len}} [{source}{binary}] {url}")
```
alalazo, line 122: :+1:

lib/spack/spack/mirror.py
```diff
@@ -111,165 +91,203 @@ def from_url(url: str):
...
+        if top_level:
+            keys += ["binary", "source"]
+        changed = False
+        for key in keys:
+            if key in new_data and current_data.get(key) != new_data[key]:
+                current_data[key] = new_data[key]
+                changed = True
+        return changed
+
+    def update(self, data: dict, direction: Optional[str] = None) -> bool:
```
alalazo, line 157: Kind of strange that `mypy` doesn't complain on having `dict` without more info on keys and values :thinking:

lib/spack/spack/mirror.py
```diff
@@ -111,165 +91,203 @@ def from_url(url: str):
...
+            bool: True if the mirror was updated, False otherwise."""
+
+        # Modify the top-level entry when no direction is given.
+        if not data:
+            return False
+
+        # If we only update a URL, there's typically no need to expand things to a dict.
+        set_url = data["url"] if len(data) == 1 and "url" in data else None
+
+        if direction is None:
```
alalazo, line 177: (comment) I wonder if we could get a simpler implementation by doing a trivial update on a normalized internal structure and then "minimizing" it :thinking:

lib/spack/spack/mirror.py
```diff
@@ -111,165 +91,203 @@ def from_url(url: str):
...
+        return self._get_value("endpoint_url", direction)
 
 
 class MirrorCollection(collections.abc.Mapping):
     """A mapping of mirror names to mirrors."""
 
-    def __init__(self, mirrors=None, scope=None):
-        self._mirrors = collections.OrderedDict(
-            (name, Mirror.from_dict(mirror, name))
+    def __init__(self, mirrors=None, scope=None, binary=any, source=any):
```
alalazo: Using the built-in `any` is a bad habit we had in other places. Can we have a separate placeholder object?

lib/spack/spack/schema/mirrors.py
```diff
@@ -6,29 +6,55 @@
...
+    "access_pair": {
+        "type": "array",
+        "items": {"type": ["string", "null"], "minItems": 2, "maxItems": 2},
+    },
+    "access_token": {"type": ["string", "null"]},
+    "profile": {"type": ["string", "null"]},
+    "endpoint_url": {"type": ["string", "null"]},
```
alalazo, line 22: Where do we use `endpoint_url`?

lib/spack/spack/stage.py
```diff
@@ -449,22 +449,11 @@ def fetch(self, mirror_only=False, err_msg=None):
...
-                    ):
-                        mirror_urls[mirror_url] = {
-                            "access_token": mirror.get_access_token("fetch"),
-                            "access_pair": mirror.get_access_pair("fetch"),
-                            "access_profile": mirror.get_profile("fetch"),
-                            "endpoint_url": mirror.get_endpoint_url("fetch"),
-                        }
+            mirror_urls = [
+                url_util.join(mirror.fetch_url, rel_path)
+                for mirror in spack.mirror.MirrorCollection(source=True).values()
```
alalazo, line 454: Similarly to the comment above, did we consider having a `source_mirrors()` convenience function?

lib/spack/spack/test/mirror.py
```diff
@@ -291,3 +296,70 @@ def test_get_all_versions(specs, expected_specs):
     output_list = [str(x) for x in output_list]
     # Compare sets since order is not important
     assert set(output_list) == set(expected_specs)
+
+
+def test_update_1():
+    # No change
+    m = spack.mirror.Mirror("https://example.com")
+    assert not m.update({"url": "https://example.com"})
+    assert m.to_dict() == "https://example.com"
```
alalazo, line 305: (comment) These four tests can be parametrized

alalazo (commented):
Only minor comments so far, I'll test drive the PR and report outcomes soon.

lib/spack/spack/schema/mirrors.py
haampie: lib/spack/spack/util/s3.py

lib/spack/spack/mirror.py
haampie: That's what I did initially, but has the downside of potentially reformatting the yaml config file

lib/spack/spack/mirror.py
haampie: ~None~ ~Went for enum~ actually reverted that too, cause it was very verbose. Now it's back to None/True/False tristate.

lib/spack/spack/test/mirror.py
haampie: but they would be much harder to read to me

lib/spack/spack/binary_distribution.py
haampie: We mix MirrorCollection(...) and MirrorCollection(...).values(), and these objects get passed on conditionally to other functions, etc, so I'm not very comfortable changing this right now. 

`binary_mirrors().values()` doesn't look like a big improvement either, and then there is a bunch of places where MirrorCollection(mirrors=..., binary=True) is called and this function cannot be used

lib/spack/spack/mirror.py
haampie: Pretty sure this is not used anywhere other than tests, didn't intend to really touch it

lib/spack/spack/stage.py
haampie: I believe it's only referred to in one place

lib/spack/spack/schema/mirrors.py
haampie: deprecatedProperties can be used, right?

lib/spack/spack/binary_distribution.py
alalazo: Ok, we can check this later and submit a PR if it makes sense.

alalazo:
A few comments after test-driving the PR.

----

The arguments of `spack mirror add` and `spack mirror set` are not consistent, since `spack mirror add` allows to set S3 options but not other options. Can we deprecate setting S3 options there?

----

I think this might be slightly confusing, since it requires to reason with negative logic:
```console
$ spack mirror add v0.20.1 https://binaries.spack.io/v0.20.1/e4s
$ spack config get mirrors
mirrors:
  v0.20.1: https://binaries.spack.io/v0.20.1/e4s
  spack-public:
    binary: false
    url: https://mirror.spack.io
$ spack mirror list
v0.20.1      [sb] https://binaries.spack.io/v0.20.1/e4s
spack-public [s ] https://mirror.spack.io
```
So far so good, since I just registered a new mirror. If I want to say the new mirror is a binary mirror it seems the key is to say `--no-source` rather than `--binary`:
```console
$ spack mirror set --binary v0.20.1
$ spack mirror list
v0.20.1      [sb] https://binaries.spack.io/v0.20.1/e4s
spack-public [s ] https://mirror.spack.io
$ spack config get mirrors
mirrors:
  v0.20.1:
    url: https://binaries.spack.io/v0.20.1/e4s
    binary: true
  spack-public:
    binary: false
    url: https://mirror.spack.io
$ spack mirror rm v0.20.1 
==> Removed mirror v0.20.1.
$ spack mirror add v0.20.1 https://binaries.spack.io/v0.20.1/e4s
$ spack mirror set --no-source v0.20.1
$ spack mirror list
v0.20.1      [ b] https://binaries.spack.io/v0.20.1/e4s
spack-public [s ] https://mirror.spack.io
```
Can we make it such that an explicit positive implies the counterpart if implicitly negative and vice-versa? Basically I'd like:
```
$ spack mirror set --binary v0.20.1
$ spack mirror set --no-source v0.20.1
```
to be equivalent above

haampie:
Solved it with `--type`, which is a bit shorter anyways.

So, `spack mirror add/set` both have optional `--type`, and

```
spack mirror set --type binary mirror # binary only
spack mirror set --type source mirror # source only
spack mirror set --type binary --type source mirror # both
```

alalazo (approved):

PR #34524, opened 2022-12-14
libcint package update

bruneval:
libcint: variant "coulomb_erf" was broken and new variant "pypzpx" created


bernhardkaindl (approved):
Built on Linux-arm64 and Linux-amd64 with and without the new variant

PR #34526, opened 2022-12-14
podio, edm4hep: Add latest versions

tmadlener:
add the latest available versions and make sure that the runtime environment is setup correctly to make everything work.

### DIFF
--- a/var/spack/repos/builtin/packages/podio/package.py
+++ b/var/spack/repos/builtin/packages/podio/package.py
@@ -122,7 +122,7 @@ class Podio(CMakePackage):
             # dynamicaly load the python bindings library
             env.prepend_path("LD_LIBRARY_PATH", self.spec["sio"].libs.directories[0])
 
-        if self.satisfies("@0.16.1:"):
+        if self.spec.satisfies("@0.16.1:"):
             # Frame header needs to be available for python bindings
             env.prepend_path("ROOT_INCLUDE_PATH", self.prefix.include)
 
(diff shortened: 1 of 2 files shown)


vvolkl (dismissed):

bernhardkaindl (approved):
Approved by vvolkl

var/spack/repos/builtin/packages/podio/package.py
```diff
@@ -121,6 +122,10 @@ def setup_run_environment(self, env):
             # dynamicaly load the python bindings library
             env.prepend_path("LD_LIBRARY_PATH", self.spec["sio"].libs.directories[0])
 
+        if self.satisfies("@0.16.1:"):
+            # Frame header needs to be available for python bindings
+            env.prepend_path("ROOT_INCLUDE_PATH", self.prefix.include)
```
bernhardkaindl: This needs to be fixed because only self.spec has .satisfies:
```suggestion
        if self.spec.satisfies("@0.16.1:"):
            # Frame header needs to be available for python bindings
            env.prepend_path("ROOT_INCLUDE_PATH", self.prefix.include)
```

bernhardkaindl (changes requested):

bernhardkaindl (approved):
Built it and looks good so far.

alalazo (approved):

var/spack/repos/builtin/packages/podio/package.py
tmadlener: thanks :)

PR #34528, opened 2022-12-14
reframe sanity checks outdated

finkandreas:
Since version 3.11.0 there is no directory `cscs-checks`. I removed it from the sanity checks. A more complete solution could be to have the sanity check still available for all version <3.11.0, but the value of it is probably very limited, besides polluting `package.py`.

### DIFF
--- a/var/spack/repos/builtin/packages/reframe/package.py
+++ b/var/spack/repos/builtin/packages/reframe/package.py
@@ -113,7 +113,6 @@ class Reframe(Package):
         "reframe",
         "tutorials",
         "unittests",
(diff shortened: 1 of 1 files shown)


vkarak:
Imho, we can safely remove that from all versions, but perhaps we can add `hpctestlib` instead from 3.11.0 onward (I need to check when we introduced that). @victorusu what do you think?

tldahlgren:
Added @vharak to assignees since has commented on this PR.

vkarak:
Replaced by #34584.

PR #34529, opened 2022-12-14
arpack-ng: Add variant for ICB option

sebastiangrimberg:
This has been in ARPACK-NG since v3.6.0.

### DIFF
--- a/var/spack/repos/builtin/packages/arpack-ng/package.py
+++ b/var/spack/repos/builtin/packages/arpack-ng/package.py
@@ -57,7 +57,7 @@ class ArpackNg(CMakePackage, AutotoolsPackage):
 
     variant("shared", default=True, description="Enables the build of shared libraries")
     variant("mpi", default=True, description="Activates MPI support")
-    variant("icb", default=False, description="Activates iso_c_binding support")
+    variant("icb", default=False, when="@3.6:", description="Activates iso_c_binding support")
 
     # The function pdlamch10 does not set the return variable.
     # This is fixed upstream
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/arpack-ng/package.py
```diff
@@ -57,6 +57,7 @@ class ArpackNg(CMakePackage, AutotoolsPackage):
 
     variant("shared", default=True, description="Enables the build of shared libraries")
     variant("mpi", default=True, description="Activates MPI support")
+    variant("icb", default=False, description="Activates iso_c_binding support")
```
alalazo:
```suggestion
    variant("icb", default=False, when="@3.6:", description="Activates iso_c_binding support")
```

var/spack/repos/builtin/packages/arpack-ng/package.py
```diff
@@ -85,6 +86,8 @@ class ArpackNg(CMakePackage, AutotoolsPackage):
         depends_on("libtool@2.4.2:", type="build")
         depends_on("pkgconfig", type="build")
 
+    conflicts("+icb", when="@:3.5.0")
+
```
alalazo:
```suggestion
```

alalazo (changes requested):
Can you check if using a conditional variant works too? It would be preferable, when possible, for clingo.

var/spack/repos/builtin/packages/arpack-ng/package.py
sebastiangrimberg: Good catch, resolved in [9bcb31d](https://github.com/spack/spack/pull/34529/commits/9bcb31dfc61b8a203b4fe28121689f2360248566).

alalazo (approved):

PR #34533, opened 2022-12-14
Add G'MIC package with only the "cli" target available

downloadico:


var/spack/repos/builtin/packages/gmic/package.py
```diff
@@ -0,0 +1,41 @@
...
+    depends_on("libxcb")
+    depends_on("libpng")
+    depends_on("openexr")
+    depends_on("opencv")
+    depends_on("zlib")
+    depends_on("zstd")
+    depends_on("libx11")
+
+    def build(self, spec, prefix):
+        make("cli")
```
alalazo, line 38:
```suggestion
    build_targets = ["cli"]
```

alalazo (approved):

var/spack/repos/builtin/packages/gmic/package.py
alalazo: For reference, I think it could have been written also this way :point_up:

PR #34541, opened 2022-12-15
elfutils: support external gettext with libintl from glibc

bernhardkaindl:
@sethrj Thank you very much for reviewing my gettext/libintl check for `bcache` in #34114 and #34383!
---
This adds the same check to `elfutils` to support `spack external find gettext` on glibc systems for building elfutils:
---
glibc's libc.so itself provides an integrated libintl, which is why on glibc systems, the gettext packages provide no libintl.so library.

To support `spack external find gettext` on glibc systems, the recipe should check if the gettext spec says that libintl is provided before adding -lintl.

The fix to add this check has been successfully reviewed and merged in the bcache recipe, so use it for adding -lintl in elfutils likewise.

### See #34114 and #34383 for the details on arriving on this already used check.

### DIFF
--- a/var/spack/repos/builtin/packages/elfutils/package.py
+++ b/var/spack/repos/builtin/packages/elfutils/package.py
@@ -124,7 +124,9 @@ class Elfutils(AutotoolsPackage, SourcewarePackage):
 
         if "+nls" in spec:
             # configure doesn't use LIBS correctly
-            args.append("LDFLAGS=-Wl,--no-as-needed -L%s -lintl" % spec["gettext"].prefix.lib)
+            # Add -lintl if provided by gettext, otherwise libintl is provided by system's glibc:
+            if any("libintl." in filename.split("/")[-1] for filename in spec["gettext"].libs):
+                args.append("LDFLAGS=-Wl,--no-as-needed -L%s -lintl" % spec["gettext"].prefix.lib)
         else:
             args.append("--disable-nls")
 
--- a/var/spack/repos/builtin/packages/glib/package.py
+++ b/var/spack/repos/builtin/packages/glib/package.py
@@ -339,7 +339,7 @@ class Glib(Package):
         # the gettext library directory. The patch below explitly adds the
         # appropriate -L path.
         spec = self.spec
-        if spec.satisfies("@2.0:2"):
+        if spec.satisfies("@2.0:2") and spec["gettext"].libs.directories:
             pattern = "Libs:"
             repl = "Libs: -L{0} -Wl,-rpath={0} ".format(spec["gettext"].libs.directories[0])
             myfile = join_path(self.spec["glib"].libs.directories[0], "pkgconfig", "glib-2.0.pc")


mwkrentel:
@bernhardkaindl I'm not sure I understand.
Are you saying there are cases where gettext does not install libintl ?

tldahlgren:
Added @mwkrentel to assignees since he has commented on this PR.

bernhardkaindl:
> @bernhardkaindl I'm not sure I understand. Are you saying there are cases where gettext does not install libintl ?

Spack's gettext recipe installs a libintl as shared library.

gettext may also be already installed with the host OS.

In the case of glibc-based Linux distributions, they don't deliver a `libintl` at all because `glibc`'s `libc` already provides the API provided by `libintl`. Example (and no, the gettext debian or rpm packages themselfes also do not provide libintl as a file):

```py
apt search libintl 2>/dev/null|grep libintl
libintl-perl/jammy,jammy 1.26-3build2 all
libintl-xs-perl/jammy 1.26-3build2 amd64
```

Likewiese on RPM-based Linux distributions.

When a user once used `spack external find gettext` or `spack external find --all`  on such normal Linux system, spack will find the `gettext` exectuable, but the `libs` property will not find a `libintl` library.

Thus, recipes which want to add `-lintl` to the linker flags should do so only after checking if a separate `libintl` has been found.
We've added this for other packages alread, this PR does this for elfutils.

bernhardkaindl:
@sethrj Hi Seth! This PR is like the PR for bcache: Fix it to only add -lintl when libintl was found in the library directories.

var/spack/repos/builtin/packages/elfutils/package.py
```diff
@@ -124,7 +124,9 @@ def configure_args(self):
 
         if "+nls" in spec:
             # configure doesn't use LIBS correctly
-            args.append("LDFLAGS=-Wl,--no-as-needed -L%s -lintl" % spec["gettext"].prefix.lib)
+            # Add -lintl if provided by gettext, otherwise libintl is provided by system's glibc:
+            if any("libintl." in filename.split("/")[-1] for filename in spec["gettext"].libs):
+                args.append("LDFLAGS=-Wl,--no-as-needed -L%s -lintl" % spec["gettext"].prefix.lib)
```
sethrj, line 129: `-Wl,--no-as-needed` might still be required, don't you think? (It's a little strange that it's only under `+nls` and nothing else, so maybe it's not needed at all, or only needed for ubuntu or other distributions that default to `-Wl,--as-needed`.) You should also take the opportunity to move this addition to the [flag_handler](https://spack.readthedocs.io/en/latest/packaging_guide.html#compiler-flags) method.

mwkrentel:
I see, so depending on where gettext comes from (built from spack or
in /usr), you may need to put `-lintl` on the link line or not.

The thing I don't like about this solution is that it puts the burden
of figuring out the right link line on all the packages that use
gettext instead of gettext.  Of course, it would be better if it could
be done once in gettext, but maybe this is a difficult case.

@becker33 @tgamblin 
Does spack have the machinery to say in the gettext package, "if I
build gettext from spack, then add `-L<libdir> -lintl`, otherwise add
nothing (since it comes from glibc)."

See #34587

As for `--no-as-needed`, I'm guessing that was used as the original
workaround for sometimes needing `-lintl` and sometimes not.

bernhardkaindl:
Actually, only recipes whose sources are assuming glibc with integrated libintl API and need to add `-lintl` to build on Alpine, MacOS or Windows need to care.

They need to be careful to add -lintl only when the gettext recipe says that it is available to not break the build.

So far, I only encountered bcache, elfutils, binutils and libxpm.

The rest can be found using `grep lintl var/`, but they are the exception.

I'd say that these few should be fixed to be careful to use it only when it's available and not make changes which affect other packages.

> Does spack have the machinery to say in the gettext package, "if I
build gettext from spack, then add -L<libdir> -lintl, otherwise add
nothing (since it comes from glibc)."

Good idea! Maybe this is a good suggestion for an implementation:

Spack recipes have `def setup_dependent_package(self, module, dependent_spec):`, in which we can add attributes to the gettext `spec` which the dependent recipes can then query, e.g. it could define `self.spec.provides_libintl` and/or could define `self.spec.libintl_link` to `-L<libdir> -lintl`, if libintl is provided or empty if not. Recipes which want to add libintl can then use `self.spec["gettext"].libintl_link`.

mwkrentel:
We discussed this at the Spack Wednesday meeting today.  We agreed 
that `libintl_link` was a good idea.  I agree, my only issue is that I
don't really know how to write the code.

So, we suggested that @bernhardkaindl   should write another PR for
gettext to create `libintl_link`, supervised by @scheibelp.
Is that OK?

scheibelp:
> So, we suggested that @bernhardkaindl should write another PR for gettext to create libintl_link, supervised by @scheibelp.

Or you can change this one. In any case: if you have to time to update this (or open a new PR) can you let me know when that is ready (@ me here) or if not, can you let me know and I can make time for it?

mwkrentel:
@tgamblin @scheibelp 
Could we bump up the priority of the better fix from gettext and put
that on someone's to do list?

That is, have gettext compute something that is either `-L/path -lintl`
if there is a separate libintl or else empty if the libintl functions
are available from libc.

There are several other packages that would benefit from this,
basically anything that has a nls variant (native lang support),
eg, #35287

bernhardkaindl:
Obsolete after the merge of #35450.

PR #34543, opened 2022-12-15
libxpm: support external gettext with libintl from glibc

bernhardkaindl:
### This adds the same check to elfutils to support spack external find gettext on glibc systems for building elfutils:

glibc's libc.so provides an integrated libintl, which is why on glibc systems, the gettext packages provide no libintl.so library.

To support `spack external find gettext` on glibc systems, the recipe must check if `spec["gettext"].libs` contains libintl if adding -lintl.

The fix to add this check has been successfully reviewed and merged in the bcache recipe, so use it for adding -lintl in elfutils likewise.

See #34114 and #34383 for the details on arriving on this already used check.

#### PS: Since #9149 is fixed by #33777, `self.spec[dependency]` can be used in any case.


bernhardkaindl:
Obsolete after the merge of #35450.

PR #34544, opened 2022-12-15
cpuinfo: new versions, shared libs

adamjstewart:
Successfully builds on macOS 12.6.1 with Apple Clang 14.0.0.

### DIFF
--- a/var/spack/repos/builtin/packages/cpuinfo/package.py
+++ b/var/spack/repos/builtin/packages/cpuinfo/package.py
@@ -26,36 +26,12 @@ class Cpuinfo(CMakePackage):
     depends_on("cmake@3.5:", type="build")
     depends_on("ninja", type="build")
 
-    # No option to use pre-installed deps
-    # https://github.com/pytorch/cpuinfo/issues/5
-    resource(
-        name="googlebenchmark",
-        url="https://github.com/google/benchmark/archive/v1.6.1.zip",
-        sha256="367e963b8620080aff8c831e24751852cffd1f74ea40f25d9cc1b667a9dd5e45",
-        destination="deps",
-        placement="googlebenchmark",
-    )
-    resource(
-        name="googletest",
-        url="https://github.com/google/googletest/archive/release-1.11.0.zip",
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/cpuinfo/package.py
```diff
@@ -10,11 +10,12 @@ class Cpuinfo(CMakePackage):
     """cpuinfo is a library to detect essential
     for performance optimization information about host CPU."""
 
-    homepage = "https://github.com/Maratyszcza/cpuinfo/"
-    git = "https://github.com/Maratyszcza/cpuinfo.git"
+    homepage = "https://github.com/pytorch/cpuinfo"
+    git = "https://github.com/pytorch/cpuinfo.git"
```
adamjstewart, line 14: https://github.com/Maratyszcza/cpuinfo is just a fork of https://github.com/pytorch/cpuinfo. PyTorch used to use the fork, and now uses upstream. Commits seems to exist in both repos.

var/spack/repos/builtin/packages/cpuinfo/package.py
```diff
@@ -25,33 +26,12 @@ class Cpuinfo(CMakePackage):
...
-            self.define(
-                "GOOGLETEST_SOURCE_DIR", join_path(self.stage.source_path, "deps", "googletest")
-            ),
-            self.define(
-                "GOOGLEBENCHMARK_SOURCE_DIR",
-                join_path(self.stage.source_path, "deps", "googlebenchmark"),
-            ),
-            self.define("CPUINFO_BUILD_UNIT_TESTS", self.run_tests),
-            self.define("CPUINFO_BUILD_MOCK_TESTS", self.run_tests),
-            self.define("CPUINFO_BUILD_BENCHMARKS", self.run_tests),
```
adamjstewart, line 56: Decided to remove these. They aren't technically tests since they change what gets installed, and the things that get installed are broken anyway.

alalazo (approved):

PR #34546, opened 2022-12-15
Package libxcrypt: Added dependency on perl

G-Ragghianti:
libxcrypt needs to run a perl script at build time which requires a perl pragma "use open".  If this is not provided by the OS, then libxcrypt fails to install.  This causes more important things to fail such as python which directly depends on libxcrypt.  This failure occurs in OSes like RHEL9 which do not by default install the "perl-open" package.

### DIFF
--- a/var/spack/repos/builtin/packages/libxcrypt/package.py
+++ b/var/spack/repos/builtin/packages/libxcrypt/package.py
@@ -33,7 +33,7 @@ class Libxcrypt(AutotoolsPackage):
     patch("truncating-conversion.patch", when="@4.4.30")
 
     depends_on("perl@5.15:", type="build")
-    
+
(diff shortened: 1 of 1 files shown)


alalazo (dismissed):

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

G-Ragghianti:
@alazo I see that you reran the pipeline to see if the checks would all pass, but there are still some jobs in the ci pipelines that are failing.  I don't see how my change could have affected these.

alalazo:
@G-Ragghianti I'm trying to rebase to check if the errors persist. Will get back to you as soon as pipelines finish.

haampie:

PR #34547, opened 2022-12-15
Propagate exceptions from Spack python console

### LINKED ISSUE #34489: spack-python does not exit failure on exceptions
### Steps to reproduce

Spack-python could be powerful tool to easily write scripts, but it ignores exceptions, which makes writing reliable scripts impossible.

For example: 

```console
$ cat myscript
#!/usr/bin/env spack-python
import badimport

$ chmod +x myscript

$ source share/spack/setup-env.sh
$ ./myscript; echo $?
  File "/usr/lib/python3.10/code.py", line 90, in runcode
    exec(code, self.locals)
  File "./myscript", line 2, in <module>
    import badimport
ModuleNotFoundError: No module named 'badimport'
0
```

Its exit status is 0 instead of non-zero.

It seems `spack-python` actually runs an interactive console, I'm not sure this is appropriate for this use-case (described in `bin/spack-python` file) ?

### Error message

Spack-python should exit with a non-zero status on errors.

* **Spack:** 0.20.0.dev0 (f5fd7e5109b24bfb5e5e5321091d083fa32cae17)
* **Python:** 3.10.6
* **Platform:** linux-ubuntu22.04-skylake
* **Concretizer:** clingo

### General information

alalazo:
fixes #34489

Customize sys.excepthook to raise SystemExit when any unhandled exception reaches the hook.

### DIFF
--- a/lib/spack/spack/cmd/python.py
+++ b/lib/spack/spack/cmd/python.py
@@ -127,8 +127,10 @@ def python_interpreter(args):
                 console.runsource(startup.read(), startup_file, "exec")
 
     if args.python_command:
+        propagate_exceptions_from(console)
         console.runsource(args.python_command)
     elif args.python_args:
+        propagate_exceptions_from(console)
         sys.argv = args.python_args
         with open(args.python_args[0]) as file:
             console.runsource(file.read(), args.python_args[0], "exec")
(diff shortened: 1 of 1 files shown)


alalazo:
@Jordan474 Can you check this works as expected for you?

haampie (approved):

Jordan474:
This works as expected; Thank You.

PR #34552, opened 2022-12-15
mirrors: auto-update paths to urls

### LINKED ISSUE #34532: spack mirror command not working
### Steps to reproduce

I was trying to build survey on snow at LANL using a mirror with the latest 12-14-2022 spack sources, but I get this message.   I was able to use the same mirror with an older 11-21-2022 version of the spack sources.   So, thought I would report this.
```
sn-fe1:/<4>survey/spack_12142022/spack [841] $ spack install --no-cache --keep-stage [survey@1.0.6.release-openmpi-2.1.2](mailto:survey@1.0.6.release-openmpi-2.1.2) %gcc@8.3.0 +mpi ^openmpi@2.1.2
==> Error: unknown url type: 'usr/projects/hpctools/survey/spack_12142022/survey-mirror-12-14-2022/build_cache/index.json'
```

### Error message

```
sn-fe1:/<4>survey/spack_12142022/spack [841] $ spack install --no-cache --keep-stage [survey@1.0.6.release-openmpi-2.1.2](mailto:survey@1.0.6.release-openmpi-2.1.2) %gcc@8.3.0 +mpi ^openmpi@2.1.2
==> Error: unknown url type: 'usr/projects/hpctools/survey/spack_12142022/survey-mirror-12-14-2022/build_cache/index.json'
```

```
$ spack debug report
* **Spack:** 0.20.0.dev0 (2387c116add454d6b7cf5400e368fb95bd3de415)
* **Python:** 3.6.8
* **Concretizer:** clingo
```

### General information

haampie:
Fixes #34532

- Fix an issue where paths were used in urls in bootstrap mirrors
- Fix an issue where file paths in `mirrors.yaml` config instead of urls made Spack open it as a URL
- Add `spack config update` to auto-update file paths to urls in mirrors.yaml
- Add flags `spack mirror add (--directory x | --url x) name` while remaining backwards compatible with now deprecated `spack mirror add <name> <url>`.

The reason to split up --directory / --url in the input is that we can throw reasonable errors in case people do:

```
spack mirror add --url example.com example # missing scheme, error!
spack mirror add example example.com # would be interpreted as a directory without error :|
```
and it helps disambiguate Windows:
```
spack mirror add --directory 'c:\\x\y\z' example # just a path, fine! 
spack mirror add example 'c:\\x\y\z' # directory or url with scheme `c`?
```

### DIFF
--- a/lib/spack/spack/cmd/mirror.py
+++ b/lib/spack/spack/cmd/mirror.py
@@ -96,7 +96,13 @@ def setup_parser(subparser):
     # Add
     add_parser = sp.add_parser("add", help=mirror_add.__doc__)
     add_parser.add_argument("name", help="mnemonic name for mirror", metavar="mirror")
-    add_parser.add_argument("url", help="url of mirror directory from 'spack mirror create'")
+
+    dir_or_url = add_parser.add_mutually_exclusive_group(required=True)
+    dir_or_url.add_argument("--directory", type=str, help="local path to a directory")
+    dir_or_url.add_argument("--url", type=str, dest="new_url_flag", help="remote mirror URL")
+    # old interface: does not distinguish between urls and paths
+    dir_or_url.add_argument("url", nargs="?", help="deprecated: use --url or --path instead")
+
     add_parser.add_argument(
         "--scope",
         choices=scopes,
@@ -148,17 +154,21 @@ def promote_deprecated_path_to_url(path_or_url):
     # but the commands really expect URLs.
     if not url_util.is_path_instead_of_url(path_or_url):
         return path_or_url
-
-    url = url_util.path_to_file_url(path_or_url)
-
-    tty.warn("Converted path to URL: {}.".format(url))
-
-    return path_or_url
+    return url_util.path_to_file_url(path_or_url)
 
 
 def mirror_add(args):
     """Add a mirror to Spack."""
-    url = url_util.format(promote_deprecated_path_to_url(args.url))
+    if args.url:
+        tty.warn(
+            "Providing the URL as a positional argument is deprecated, use "
+            "`spack mirror add --url ... name` instead"
+        )
+        url = promote_deprecated_path_to_url(args.url)
+    elif args.directory:
+        url = url_util.path_to_file_url(args.directory)
+    else:  # elif args.new_url_flag:
+        url = args.new_url_flag
     spack.mirror.add(args.name, url, args.scope, args)
 
 
--- a/lib/spack/spack/cmd/mirror.py
+++ b/lib/spack/spack/cmd/mirror.py
@@ -144,7 +144,7 @@ def setup_parser(subparser):
 
 
 def promote_deprecated_path_to_url(path_or_url):
-    # Spack tutorial still mention spac mirror add <name> <path>
+    # Spack tutorial still mentions spack mirror add <name> <path>
     # but the commands really expect URLs.
     if not url_util.is_path_instead_of_url(path_or_url):
         return path_or_url
--- a/lib/spack/spack/cmd/mirror.py
+++ b/lib/spack/spack/cmd/mirror.py
@@ -143,9 +143,22 @@ def setup_parser(subparser):
     )
 
 
+def promote_deprecated_path_to_url(path_or_url):
+    # Spack tutorial still mention spac mirror add <name> <path>
+    # but the commands really expect URLs.
+    if not url_util.is_path_instead_of_url(path_or_url):
+        return path_or_url
+
+    url = url_util.path_to_file_url(path_or_url)
+
+    tty.warn("Converted path to URL: {}.".format(url))
+
+    return path_or_url
+
+
 def mirror_add(args):
     """Add a mirror to Spack."""
-    url = url_util.format(args.url)
+    url = url_util.format(promote_deprecated_path_to_url(args.url))
     spack.mirror.add(args.name, url, args.scope, args)
 
 
@@ -156,7 +169,7 @@ def mirror_remove(args):
 
 def mirror_set_url(args):
     """Change the URL of a mirror."""
-    url = url_util.format(args.url)
+    url = url_util.format(promote_deprecated_path_to_url(args.url))
     mirrors = spack.config.get("mirrors", scope=args.scope)
     if not mirrors:
         mirrors = syaml_dict()
--- a/lib/spack/spack/test/cmd/mirror.py
+++ b/lib/spack/spack/test/cmd/mirror.py
@@ -26,25 +26,6 @@ uninstall = SpackCommand("uninstall")
 pytestmark = pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
 
 
-@pytest.fixture
-def tmp_scope():
-    """Creates a temporary configuration scope"""
-
-    base_name = "internal-testing-scope"
-    current_overrides = set(
-        x.name for x in spack.config.config.matching_scopes(r"^{0}".format(base_name))
-    )
-
-    num_overrides = 0
-    scope_name = base_name
-    while scope_name in current_overrides:
-        scope_name = "{0}{1}".format(base_name, num_overrides)
-        num_overrides += 1
-
-    with spack.config.override(spack.config.InternalConfigScope(scope_name)):
-        yield scope_name
-
-
 @pytest.mark.disable_clean_stage_check
 @pytest.mark.regression("8083")
 def test_regression_8083(tmpdir, capfd, mock_packages, mock_fetch, config):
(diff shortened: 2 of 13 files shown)


lib/spack/spack/test/cmd/mirror.py
```diff
@@ -154,48 +135,44 @@ def test_exclude_file(mock_packages, tmpdir, config):
     assert not any(spec.satisfies(y) for spec in mirror_specs for y in expected_exclude)
 
 
-def test_mirror_crud(tmp_scope, capsys):
+def test_mirror_crud(mutable_config, capsys):
```
tgamblin, line 138: This is a good change.

lib/spack/spack/cmd/mirror.py
```diff
@@ -143,9 +143,22 @@ def setup_parser(subparser):
...
 
+def promote_deprecated_path_to_url(path_or_url):
+    # Spack tutorial still mentions spack mirror add <name> <path>
+    # but the commands really expect URLs.
+    if not url_util.is_path_instead_of_url(path_or_url):
+        return path_or_url
+
+    url = url_util.path_to_file_url(path_or_url)
+
+    tty.warn("Converted path to URL: {}.".format(url))
```
tgamblin: I don't think this should be a warning. If you're using a filesystem mirror, you should be able to supply a path, right?  Is the issue that on windows we're unsure of whether we're doing this correctly?  `is_path_instead_of_url` looked pretty good to me... are there situations where it'll still do something unexpected and we need to warn?

If not, I think this can either be a debug message, or we can remove it.

tgamblin (changes requested):
2 questions:

1. See question below on the warning.
2. Does this preclude ever using relative paths to a mirror in `spack.yaml`?  If so then I think we should figure out how to support that.  Doesn't have to be this PR, but jut "promoting to URLs" loses functionality in that the URL has to absolute, right?

lib/spack/spack/test/cmd/mirror.py
haampie: I forgot to motivate the change I think, but the Real Reason is that argparse validates --scope values with a static list, so the first times `spack mirror` is executed it's fixed and custom scopes are no longer allowed. The fact that it worked so far is just by chance, randomizing the test order makes it break most of the time.

haampie:
> Does this preclude ever using relative paths to a mirror in spack.yaml

~I think so yes.~ Actually not:

```yaml
spack:
  specs:
  - zlib
  mirrors::
    typo: binaries.spack.io/develop
    url: https://binaries.spack.io/develop
    path: ./x
```

```yaml
$ spack -e . config get mirrors
mirrors:
  typo: file:///tmp/tmp.X0vFGNzE4P/binaries.spack.io/develop
  url: https://binaries.spack.io/develop
  path: file:///tmp/tmp.X0vFGNzE4P/x
```

lib/spack/spack/cmd/mirror.py
haampie: Hm, not sure to be honest. We should just improve the API. If we promote without warning you get weird things like:

`spack mirror add x binaries.spack.io/develop` => `file://$PWD/binaries.spack.io/develop` where it's interpreted as a relative path

If we expect a URL we can just error if the scheme is unsupported, or when it's lacking, or decide to default to https or w/e.

lib/spack/spack/cmd/mirror.py
```diff
@@ -143,9 +143,22 @@ def setup_parser(subparser):
...
+
+    return path_or_url
```
haampie: lol, this was  bug.

lib/spack/spack/cmd/mirror.py
haampie: I also don't like `spack mirror add [--directory ... | --url ...] [name]` to be honest :sweat:

haampie: The documentation is quite explicit that it requires a URL, so let's just stick to the warning before improving the API.

```
$ spack mirror add --help
usage: spack mirror add ... mirror url

positional arguments:
  mirror                mnemonic name for mirror
  url                   url of mirror directory from 'spack mirror create'
```

haampie: At the same time our tutorial has:

```
spack mirror add mymirror ~/mirror
```

all over the place :weary: feels awkward to replace with `"file://$HOME/mirror"`

haampie:
Okay. How about this:

```
spack mirror add (--url ... | --directory ...) name [url]
```

The `[url]` positional arg is for backwards compatibility and miraculously you can tell argparse to make 2 flags and 1 positional arg mutually exclusive.

This allows you to do `spack mirror add --directory ~/my_cache my_cache`, `spack mirror add --directory example.com name`, `spack mirror add --url https://binaries.spack.io name` etc.

At the end of the day only URLs are stored in config, and input is validated in a sensible way.

You can still store paths in `mirrors.yaml`, and they are silently promoted to file urls. Also `spack config update mirrors` will persist them as file urls.

tgamblin:
Ok sorry to be a pain but I'm still not convinced we want to complicate the API.  Consider git:

```console
> git remote add foo foo

> git push foo
fatal: 'foo' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

> git init foo
Initialized empty Git repository in /Users/gamblin2/src/spack/foo/.git/

> git push foo
Enumerating objects: 409651, done.
Counting objects: 100% (409651/409651), done.
Delta compression using up to 10 threads
Compressing objects: 100% (137018/137018), done.
Writing objects: 100% (409651/409651), 66.46 MiB | 34.06 MiB/s, done.
Total 409651 (delta 202057), reused 363668 (delta 163445), pack-reused 0
remote: Resolving deltas: 100% (202057/202057), done.
remote: Checking connectivity: 409651, done.
To foo
 * [new branch]            develop -> develop
```

And if you do not provide a valid scheme, it's just considered a file:

```console
> git remote add foo example.com
> git push foo
fatal: 'example.com' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

vs.:

```console
> git remote add foo https://example.com
> git push foo
fatal: repository 'https://example.com/' not found
```

Why do we need to make users use a more cumbersome API here when we can tell this stuff apart?

If the issue is that windows makes things more ambiguous, can we not decide to *only* consider `C:\` etc. paths on Windows?  I just see the extra args here as putting the disambiguation task on the user when it really should be Spack's job.

The other thing I want to avoid is this (sorry if I was less clear earlier about relative paths):

> You can still store paths in mirrors.yaml, and they are silently promoted to file urls. 

If we silently promote as in https://github.com/spack/spack/pull/34552#issuecomment-1372475395, we end up promoting what was a relative path to an absolute path (the last one that ends up pointing at `file:///tmp...`).  The issue I see with that is for environments: you could previously have a valid environment-relative path to a mirror in your env, and someone could potentially clone that and push to a local directory *within* the cloned repository directory.  With the new scheme the path is made into something that won't work in that context -- the relative path is conveyed to an abspath that is only valid on the system of the original caller of `spack mirror add`.

I think `git remote add` is a pretty good API (or, at least, it's well understood) -- keeping this API similar I think will be good for your eventual proposed `push` API too.  e.g., this:

```console
> spack mirror push https://github.com/blah /abc123
```

is more natural than:

```console
> spack mirror push --url https://github.com/blah /abc123
```

I look at that second one and *know* I'm going to forget `--url` every time -- especially since my brain is so used to `git remote`.

haampie:
Sigh, so, the suggestion is to do this:

```
spack mirror create [name / path / url] [specs...]
spack buildcache create [name / path / url] [specs...]
spack buildcache update-index [name / path / url]
spack buildcache sync [src name / path / url] [dst name / path / url]
```

And then to retrieve the url, check if it is a valid name (e.g. [a-zA-Z_-]), if so, look it up, if not found, then it's a relative path, if it's not a valid name, it's either a url or a path, so parse as url, if it has a scheme, use it as a url, otherwise use it as a path. 

Actually we can just follow git exactly, I can check if they have comments on how they resolve the remote url.

---

Regarding relative paths, I think that's a separate issue though. Cause in Spack's config model we first flatten the scopes and then interpret the values, so if you wanna support relative paths, it should probably be through `$env/../mirror`, not `../mirror`, since it's unclear what it's relative to (the original file where it's defined, the environment scope, the current working directory -- there's also config includes from a different dir). This is different from environment-local settings like where the view is.

Also note the following, relative paths for repos get expanded:

```
$ spack repo create ./here
$ spack env create -d .
$ spack -e . repo add ./here # same with spack -e . repo add '$env/here/'
$ cat spack.yaml 
spack:
  ...
  repos:
  - /tmp/tmp.qgz4YTXJjG/here
```

tgamblin:
> Actually we can just follow git exactly, I can check if they have comments on how they resolve the remote url.

I like it -- I think it makes the most sense since this is the paradigm we're going for.

> Regarding relative paths, I think that's a separate issue though.

See here -- the markers we put in parsed YAML will tell you what specific YAML files strings came from -- that is already handled in `canonicalize_path()`, which is intended to be called when reading data from configs:

https://github.com/spack/spack/blob/3a0db729c7fc0def2ac58c7c487cd43d5578ba78/lib/spack/spack/util/path.py#L362-L377

`$env` is for *forcing* the thing to be env-relative regardless of what include/config scope it came from, but the default interpretation is "relative to the file it came from", which is intended to make it easy to refer to files in git repos (as you'd expect with, say, source code).

> Cause in Spack's config model we first flatten the scopes and then interpret the values

Yep, we also jump through some small hoops to preserve the top-most strings in `merge_yaml()` so that this file provenance stays intact.  See: https://github.com/spack/spack/blob/3a0db729c7fc0def2ac58c7c487cd43d5578ba78/lib/spack/spack/config.py#L1147-L1157

PR #34554, opened 2022-12-15
binary_distribution: show extract / relocate time separately

haampie:
Replace `Install: 34s` with `Extract: 30s.  Relocate 1s.  Database: 3s.` so that we know where
we ~waste~ spend time

Result from https://gitlab.spack.io/spack/spack/-/jobs/4994086, the totals in seconds:

| Search | Fetch  | Extract | Relocate | Database | Total  |
|--------|--------|---------|----------|----------|--------|
| 0.0    | 324.55 | 265.49  | 49.32    | 210.73   | 851.86 |

x

Notice that this is CPU time, not wall time


alalazo:
This would indeed be useful.

PR #34573, opened 2022-12-16
Remove dep on jupyter meta-package

CreRecombinase:
Right now r-irkernel requires py-jupyter, which is a metapackage for everything in the jupyter ecosystem.  That includes (among other things) the desktop client, qtconsole, which depends on qt. Besides being a huge (and unnecessary) dependency on its own, qt by default transitively pulls in llvm (qt->glx->mesa->llvm), which adds a lot of time to a fresh install. 

I'm pretty sure all that's really needed here at runtime is the jupyter client.


glennpj (approved):
This sounds reasonable to me.

PR #34575, opened 2022-12-16
bacio version fix

AlexanderRichert-NOAA:
This PR patches an incorrect version in bacio's VERSION file for bacio@2.4.1. For that release, VERSION contains 2.4.**0**, which gets propagated into the cmake config files.

### DIFF
--- a/var/spack/repos/builtin/packages/bacio/package.py
+++ b/var/spack/repos/builtin/packages/bacio/package.py
@@ -41,4 +41,5 @@ class Bacio(CMakePackage):
         return args
 
     def patch(self):
-        filter_file(".*", "2.4.1", "VERSION", when="bacio@2.4.1")
+        if self.spec.satisifes("@2.4.1"):
+            filter_file(".*", "2.4.1", "VERSION")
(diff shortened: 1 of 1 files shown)


Hang-Lei-NOAA (dismissed):

alalazo (approved):

var/spack/repos/builtin/packages/bacio/package.py
```diff
@@ -39,3 +39,6 @@ def cmake_args(self):
         args = [self.define_from_variant("CMAKE_POSITION_INDEPENDENT_CODE", "pic")]
 
         return args
+
+    def patch(self):
+        filter_file(".*", "2.4.1", "VERSION", when="bacio@2.4.1")
```
alalazo: After looking at the definition of `filter_file`, I think you need to (there is no `when=` argument):
```suggestion
        if self.spec.satisifes("@2.4.1"):
            filter_file(".*", "2.4.1", "VERSION")
```

alalazo (changes requested):

alalazo (approved):

PR #34578, opened 2022-12-16
libpng static

AlexanderRichert-NOAA:
This PR enables static-only build of libpng. The package does not currently have a maintainer, so I have added myself. This PR is needed to support NOAA/NWS R&D and operational needs.

### DIFF
--- a/var/spack/repos/builtin/packages/libpng/package.py
+++ b/var/spack/repos/builtin/packages/libpng/package.py
@@ -26,6 +26,8 @@ class Libpng(AutotoolsPackage):
 
     depends_on("zlib@1.0.4:")  # 1.2.5 or later recommended
 
+    variant("shared", default=True)
+
     def configure_args(self):
         args = [
             # not honored, see
@@ -34,6 +36,8 @@ class Libpng(AutotoolsPackage):
             f"CPPFLAGS={self.spec['zlib'].headers.include_flags}",
             f"LDFLAGS={self.spec['zlib'].libs.search_flags}",
         ]
+        if "~shared" in self.spec:
+            args += ["--enable-static", "--disable-shared"]
         return args
 
     def check(self):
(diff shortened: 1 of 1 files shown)


tldahlgren (dismissed):
LGTM.  Will defer merge decision to @scheibelp since he's on the reviewer list.

var/spack/repos/builtin/packages/libpng/package.py
```diff
@@ -34,6 +38,8 @@ def configure_args(self):
             f"CPPFLAGS={self.spec['zlib'].headers.include_flags}",
             f"LDFLAGS={self.spec['zlib'].libs.search_flags}",
         ]
+        if "~shared" in self.spec:
+            args += ["--enable-static", "--disable-shared"]
```
alalazo: If both shared and static can be built at the same time, can we add a `libs` variant like e.g. in the `lzo` package:
```python
 variant(
        "libs",
        default="shared,static",
        values=("shared", "static"),
        multi=True,
        description="Build shared libs, static libs or both",
    )
```
?

alalazo (changes requested):

var/spack/repos/builtin/packages/libpng/package.py
AlexanderRichert-NOAA: Yeah I'm good with that, I'll add it and test it out.

var/spack/repos/builtin/packages/libpng/package.py
```diff
@@ -38,8 +44,16 @@ def configure_args(self):
...
+
+        if self.spec.satisfies("libs=shared"):
+            args += ["--enable-shared"]
+        else:
+            args += ["--disable-shared"]
+        if self.spec.satisfies("libs=static"):
+            args += ["--enable-static"]
+        else:
+            args += ["--disable-static"]
+
```
alalazo:
```suggestion
        args += self.enable_or_disable("libs")
```

alalazo (changes requested):
I think the recipe can be simplified further.

alalazo (dismissed):

haampie (approved):

PR #34583, opened 2022-12-16
dashan-runtime,darshan-util,py-darshan: update package versions for darshan-3.4.2

shanedsnyder:


tldahlgren (dismissed):
Confirmed the three sha256.

jeanbez (dismissed):

tldahlgren:
Added @adamjstewart for the python package.

var/spack/repos/builtin/packages/darshan-runtime/package.py
```diff
@@ -25,10 +25,14 @@ class DarshanRuntime(AutotoolsPackage):
     test_requires_compiler = True
 
     version("main", branch="main", submodules=True)
+    version(
+        "3.4.2",
+        sha256="b095c3b7c059a8eba4beb03ec092b60708780a3cae3fc830424f6f9ada811c6b",
+        preferred=True,
```
adamjstewart: I would just remove preferred, we already default to the latest stable release. You can even put a lot of these versions on a single line if you do

adamjstewart (commented):
Python stuff looks fine, my review is an optional suggestion

var/spack/repos/builtin/packages/darshan-runtime/package.py
shanedsnyder: Good to know, I pushed in a small change for this.

adamjstewart (approved):

PR #34587, opened 2022-12-17
binutils: Fix support for using the libintl of glibc

bernhardkaindl:
This reuses same check to support `spack external find gettext` on glibc systems like in #34114 

gettext may be installed by spack or when `spack external find gettext` was used from the host OS.

Glibc-based Linux distributions don't deliver a libintl at all because glibc's libc already provides the API provided by libintl.

Whe using external gettext, but the libs property will not find a libintl library on glibc systems therefore.

Thus, recipes which want to add -lintl to the linker flags should do so only after checking if a separate libintl has been found by gettext's libs property.

The added check has already be reviewed and improved in #34114, and spec = self.spec is used as a simple helper and to no exceed 99 chars on one line.

### DIFF
--- a/var/spack/repos/builtin/packages/binutils/package.py
+++ b/var/spack/repos/builtin/packages/binutils/package.py
@@ -186,8 +186,11 @@ class AutotoolsBuilder(spack.build_systems.autotools.AutotoolsBuilder):
         install(join_path(self.build_directory, "bfd", "*.h"), extradir)
 
     def setup_build_environment(self, env):
-        if self.spec.satisfies("%cce"):
+        spec = self.spec
+        if spec.satisfies("%cce"):
             env.append_flags("LDFLAGS", "-Wl,-z,muldefs")
 
-        if "+nls" in self.spec:
-            env.append_flags("LDFLAGS", "-lintl")
+        if "+nls" in spec:
+            # Add -lintl if provided by gettext. Otherwise, libintl is provided by system's glibc:
+            if any("libintl." in filename.split("/")[-1] for filename in spec["gettext"].libs):
+                env.append_flags("LDFLAGS", "-lintl")


var/spack/repos/builtin/packages/binutils/package.py
```diff
@@ -186,8 +186,11 @@ def install_headers(self):
...
-        if self.spec.satisfies("%cce"):
+        spec = self.spec
+        if spec.satisfies("%cce"):
             env.append_flags("LDFLAGS", "-Wl,-z,muldefs")
 
-        if "+nls" in self.spec:
-            env.append_flags("LDFLAGS", "-lintl")
+        if "+nls" in spec:
+            # Add -lintl if provided by gettext. Otherwise, libintl is provided by system's glibc:
+            if any("libintl." in filename.split("/")[-1] for filename in spec["gettext"].libs):
```
sethrj, line 195: This is used in several packages now in your pull requests: I think you should add a property `provides_libintl ` to the gettext package, and replace this (and the same call in your others PR) with `if spec["gettext"].provides_libintl`

It might also be good to add a `libintl` variant to `gettext` with autodetection capability in the external case, and use that variant instead.

bernhardkaindl: Thanks for both suggestions! If understand correctly:
* If we can reliably detect "gettext+libintl" (I think we'd have to search for the library file), a `depends_on("gettext+libintl")` would force spack to build it's own internal gettext package on all glibc-Linux systems. The external gettext from the glibc operating system then would not be used.
* `if spec["gettext"].provides_libintl` will work with with both: It will be happy with a `spack external find gettext` on any Linux distro (glibc or not).

With `spack external find gettext`, `if spec["gettext"].provides_libintl`'
* can reuse the system-`gettext` package on any glibc-based Linux and a spack,
* can reuse the `gettext` on Alpine Linux (musl) if wanted
* and only needs to build `gettext` for MacOS and Windows.

I'd to do the latter, as it does not have a need to compile gettext for almost all Linux users if they have their gettext rpm/deb/etc installed and ran `spack external find gettext` or `spack external find --all` once.

sethrj: Instead of `depends_on("gettext+libintl")` I was thinking
```python
if "gettext+nls" in spec:
    # libintl is provided by gettext
```
would be more canonical than having a special property.

sethrj: Also I think `gettext` is almost always built even for linux users. Since it's not tagged, and it potentially has several system library dependencies, it's safer for spack to build its own. You'd have to manually `spack external find gettext` to set up the external, and even then it's not set up to detect variants...

bernhardkaindl: AFAICS, no spack recipe uses any of the variants of the gettext recipe, so they could be removed or not?

There exists `spack external find --all`, which makes sense when wanting to build some packages reuse as much as possible from the system, and it includes gettext.

If you want that all spack users always must install spack's gettext, then the easiest solution would be to remove the feature to use an external getext:

```py
--- a/var/spack/repos/builtin/packages/gettext/package.py
+++ b/var/spack/repos/builtin/packages/gettext/package.py
@@ -16,3 +16,2 @@
 
     maintainers = ["michaelkuhn"]
 
-    executables = [r"^gettext$"]
-
```

What many recipes mean whan they add `depends_on("gettext") is actually:

"I would like to link with libintl", think calling the gettext tool should acutally be not so common.

`depends_on("gettext+libintl")` would be most specific to what packages with add "-lintl" actually want, or just the lookup of the `libs` property as in this PR.

@michaelkuhn, since you are the maintainer of gettext, what do you think?

bernhardkaindl: @mwkrentel wrote:
> Does spack have the machinery to say in the gettext package, "if I
build gettext from spack, then add -L<libdir> -lintl, otherwise add
nothing (since it comes from glibc)."

Good idea! Maybe this is a good suggestion for an implementation:

Spack recipes have `def setup_dependent_package(self, module, dependent_spec):`, in which we can add attributes to the gettext `spec` which the dependent recipes can then query, e.g. it could define `self.spec.provides_libintl` and/or could define `self.spec.libintl_link` to `-L<libdir> -lintl`, if libintl is provided or empty if not. Recipes which want to add libintl can then use `self.spec["gettext"].libintl_link`.

michaelkuhn: I agree that it would probably be best to solve this in the `gettext` package so we don't have to carry the same (or similar) workaround in multiple packages. I don't have a strong opinion on whether it should be a variant or a property, though.

alalazo:

PR #34589, opened 2022-12-17
VTK: fix 9.1–9.2 builds

adamjstewart:
This PR includes the following changes:

- [x] Add missing nlohmann-json dependency
- [x] Fix missing verdict dependency, use internal copy since no spack package
- [x] Change default version to latest version (supports default Python version)
- [x] Fix cmake configuration error in 9.1.0

Unfortunately, it still doesn't build. Here is 9.2.2:
```
Undefined symbols for architecture arm64:
  "Ioss::Init::Initializer::Initializer()", referenced from:
      vtkIOSSReader::vtkInternals::vtkInternals(vtkIOSSReader*) in vtkIOSSReader.cxx.o
      vtkIOSSWriter::vtkIOSSWriter() in vtkIOSSWriter.cxx.o
  "Ioss::Init::Initializer::~Initializer()", referenced from:
      vtkIOSSReader::vtkInternals::vtkInternals(vtkIOSSReader*) in vtkIOSSReader.cxx.o
      vtkIOSSReader::vtkInternals::~vtkInternals() in vtkIOSSReader.cxx.o
      std::__1::unique_ptr<vtkIOSSWriter::vtkInternals, std::__1::default_delete<vtkIOSSWriter::vtkInternals> >::~unique_ptr() in vtkIOSSWriter.cxx.o
      vtkIOSSWriter::~vtkIOSSWriter() in vtkIOSSWriter.cxx.o
      vtkIOSSWriter::vtkIOSSWriter() (.cold.1) in vtkIOSSWriter.cxx.o
  "Iotr::Factory::create(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)", referenced from:
      vtkIOSSUtilities::GetConnectivity(Ioss::GroupingEntity*, int&, vtkIOSSUtilities::Cache*) in vtkIOSSUtilities.cxx.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/libvtkIOIOSS-9.2.9.2.2.dylib] Error 1
make[1]: *** [IO/IOSS/CMakeFiles/IOIOSS.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
9.1.0 also has problems (note that I'm building `~mpi`):
```
/var/folders/j1/68dlgpr91vlgs26vty2c8xk80000gn/T/ajstewart/spack-stage/spack-stage-vtk-9.1.0-t2rhk4gjtvvw7em3dvbpfhb2qnmrwevh/spack-src/IO/IOSS/vtkIOSSReader.cxx:1304:38: error: use of undeclared identifier 'MPI_COMM_WORLD'
      dbasename, Ioss::READ_RESTART, MPI_COMM_WORLD, properties));
                                     ^
```
Any suggestions would be useful.

### DIFF
--- a/var/spack/repos/builtin/packages/vtk/package.py
+++ b/var/spack/repos/builtin/packages/vtk/package.py
@@ -23,11 +23,7 @@ class Vtk(CMakePackage):
 
     version("9.2.2", sha256="1c5b0a2be71fac96ff4831af69e350f7a0ea3168981f790c000709dcf9121075")
     version("9.1.0", sha256="8fed42f4f8f1eb8083107b68eaa9ad71da07110161a3116ad807f43e5ca5ce96")
-    version(
-        "9.0.3",
-        sha256="bc3eb9625b2b8dbfecb6052a2ab091fc91405de4333b0ec68f3323815154ed8a",
-        preferred=True,
-    )
+    version("9.0.3", sha256="bc3eb9625b2b8dbfecb6052a2ab091fc91405de4333b0ec68f3323815154ed8a")
     version("9.0.1", sha256="1b39a5e191c282861e7af4101eaa8585969a2de05f5646c9199a161213a622c7")
     version("9.0.0", sha256="15def4e6f84d72f82386617fe595ec124dda3cbd13ea19a0dcd91583197d8715")
     version("8.2.0", sha256="34c3dc775261be5e45a8049155f7228b6bd668106c72a3c435d95730d17d57bb")
@@ -149,7 +145,6 @@ class Vtk(CMakePackage):
     depends_on("seacas@2021-05-12:+mpi", when="@9.1: +mpi")
     depends_on("seacas@2021-05-12:~mpi", when="@9.1: ~mpi")
     depends_on("nlohmann-json", when="@9.2:")
-    # depends_on("verdict", when="@9.2:")
 
     # For finding Fujitsu-MPI wrapper commands
     patch("find_fujitsu_mpi.patch", when="@:8.2.0%fj")
@@ -228,6 +223,8 @@ class Vtk(CMakePackage):
                         "-DVTK_MODULE_USE_EXTERNAL_VTK_fmt:BOOL=OFF",
                     ]
                 )
+            if spec.satisfies("@9.2:"):
+                cmake_args.append("-DVTK_MODULE_USE_EXTERNAL_VTK_verdict:BOOL=OFF")
 
         # Some variable names have changed
         if spec.satisfies("@8.2.0"):
--- a/var/spack/repos/builtin/packages/vtk/package.py
+++ b/var/spack/repos/builtin/packages/vtk/package.py
@@ -129,7 +129,7 @@ class Vtk(CMakePackage):
     depends_on("lz4")
     depends_on("netcdf-c~mpi", when="~mpi")
     depends_on("netcdf-c+mpi", when="+mpi")
-    depends_on("netcdf-cxx")
+    depends_on("netcdf-cxx4")
     depends_on("libpng")
     depends_on("libtiff")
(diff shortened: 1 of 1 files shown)


mathstuf:
We have definitely pulled snapshots of Ioss (from Seacas) at various points. I now try to stick to tags (with picked patches). Is Spack's VTK trying to use an external Ioss? If so, I'd just use the internal one. Otherwise, I think this might be related to `-fvisibility=hidden`; is Spack injecting this flag? Ioss only recently gained support for properly exporting symbols.

I suspect we also lack sufficient coverage of non-MPI builds in CI.

adamjstewart:
Yes, we're trying to link to external seacas.

adamjstewart:
I can't figure out how to use OpenGL on macOS so I can no longer test this locally, but if we can get things working in CI then that should be enough to merge this. I'll try pinning seacas to the same versions in third-party, or remove the dep entirely, until we can get things working.

adamjstewart:
It doesn't even seem like the latest version of VTK is being built in CI...

alalazo (approved):
@danlipsa @kwryankrattiger Can you double check this one too?

var/spack/repos/builtin/packages/vtk/package.py
```diff
@@ -15,19 +15,15 @@ class Vtk(CMakePackage):
...
+    list_url = "https://www.vtk.org/download/"
 
     maintainers = ["chuckatkins", "danlipsa"]
 
     version("9.2.2", sha256="1c5b0a2be71fac96ff4831af69e350f7a0ea3168981f790c000709dcf9121075")
     version("9.1.0", sha256="8fed42f4f8f1eb8083107b68eaa9ad71da07110161a3116ad807f43e5ca5ce96")
-    version(
-        "9.0.3",
-        sha256="bc3eb9625b2b8dbfecb6052a2ab091fc91405de4333b0ec68f3323815154ed8a",
-        preferred=True,
```
kwryankrattiger, line 29: This should still be preferred over the 9.0.0 and 9.0.1 releases, I think that is why it is here. Does the new concretizer prioritize highest possible version now? The old one didn't which is why this went in originally.

kwryankrattiger:
> It doesn't even seem like the latest version of VTK is being built in CI...

Correct, the only package that uses VTK in CI right now is VisIt which is still stuck on 8.

var/spack/repos/builtin/packages/vtk/package.py
adamjstewart: The concretizer always has and always will prioritize the most recent stable release.

kwryankrattiger: That...is not entirely true in practice. Building out the SDK package I have seen the concretizer struggle a lot with selecting a newer version when possible and have had to jump through hoops to make it happen reliably.

That said, worst case I suppose I can come back later and add preferred if needed with a comment explaining why and an issue number for reference so it isn't just speculation as to why it was needed. I didn't make this change, so there may not have been an issue at all and I am speculating.

adamjstewart: If you notice that again, ping @alalazo and he can probably explain what's going on. I'm guessing it's trying to avoid rebuilding something that is already installed or trying to minimize dependencies by selecting an older version.

kwryankrattiger (approved):
Thanks! This has been on my list of things to do for awhile!

adamjstewart:
Did anyone confirm that it builds? It wasn't working for me the last time I tried it but I'm no longer able to test it due to OpenGL issues on macOS.

kwryankrattiger:
+2 Validated

PR #34594, opened 2022-12-17
Add Go v1.19.5, v1.18.10, and deprecate unsupported versions

alecbcs:
Add latest Go v1.19.4 and security fix release v1.18.10. Deprecate all older versions which are insecure and unsupported.


alalazo (commented):
Can we avoid adding new versions which are already deprecated?

alecbcs:
@alalazo is there a reason why we'd want to limit the number of deprecated versions? Providing known checksums for past versions seems like a more secure option to inform users of possible man in the middle attacks. In addition, marking the versions as deprecated helps folks know that installing `go@1.11.2` isn't a wise idea even if they're completely new to the Go ecosystem.

alecbcs:
@alalazo ping

alalazo:
> is there a reason why we'd want to limit the number of deprecated versions?

Mainly to reduce the clutter. If we don't provide a version Spack won't download that without some user intervention - so I think it will be safe by default. 

>  In addition, marking the versions as deprecated helps folks know that installing go@1.11.2 isn't a wise idea even if they're completely new to the Go ecosystem.

Sure, but if 1.11.2 is deprecated, I assume they could infer that prior patches are deprecated too. If we had to do this for every package, recipes would be cluttered very soon. Also, a lot of versions increase the burden on clingo (they extend the solution space, and while the solver itself might be smart to avoid low score solutions, I expect the work of the grounder to be increased by that). 

@tgamblin Do you have opinions on this?

alecbcs:
> is there a reason why we'd want to limit the number of deprecated versions?
>> Mainly to reduce the clutter. If we don't provide a version Spack won't download that without some user intervention - so I think it will be safe by default.

Okay, this makes sense. My argument is that won't package recipes trend towards increased complexity over time anyway unless we begin deleting package versions? (i.e. say that we've got an active maintainer for a package like go who adds each and every minor version to Spack over time. Unless we begin deleting versions from the recipe wouldn't the `package.py` file end up in the same state?)

And so perhaps this is more of a Spack philosophy question I just don't know the answer to. In the Nix ecosystem maintainers actively prune their recipes, quickly removing deprecated versions. In Spack however, we seem to treat recipes similar append only logs in order to maximize reproducibility and so I've always thought of Spack as a sort of version maximalist where it wants to store all the versions of everything to ensure the reproducibility of old environments. (And I could totally be wrong in my understanding here. Maybe this thinking is unsustainable for the concretizer/grounder.)

alalazo:
> My argument is that won't package recipes trend towards increased complexity over time anyway unless we begin deleting package versions?

Yes, that's why I am strongly in favor of cleaning up recipes from one major release to the next. I think @adamjstewart agrees with that. See for instance the cleanup he has done over the Python ecosystem. We definitely don't want, however, to keep only the most recent version of everything like  some other package managers do.

adamjstewart:
I'm the one who added the deprecate=True flag. The only purpose of this flag imo is to remove older versions but give users warning ahead of time. Honestly, I think we should go through all packages and remove all deprecated versions after each new minor release. So I agree that it doesn't make much sense to add a new version and deprecate it at the same time.

alecbcs:
Okay this sounds like a reasonable stance on cleaning up old package versions. Is there any worry that this'll harm environment reproducibility or is the expectation that users will go back and use the version of Spack they'd previously used for an experiment?

tgamblin:
I am on the fence about this, but I'll go with @adamjstewart and @alalazo's suggestion here.  There are packages where we likely DO want to maintain reproducibility, even if there are security issues -- otherwise people are not going to be able to go back and reproduce a prior version of something.

That said, we don't (and can't) test everything, so keeping around versions that we don't test somehow, or that nobody is maintaining or checking the recipe with, doesn't make a whole lot of sense to me either.

I think the right approach is likely to maintain packages the way that @adamjstewart and @alalazo suggest, but to keep the old recipes with the package installations/environments (we actually do this right now) and move towards a model where we reproduce from *exactly* the `package.py` files that were used to install.  That way you can always try to reproduce something that you *know* built successfully, but you don't run the risk that the recipe is out of sync with the old version.  We don't currently do this and I think it would require a little work and a more stable package API to really be useful, so I would still err on the side of what they are suggesting.  I am sensitive to keeping things manageable for package maintainers.

I don't know enough about Go and its versioning scheme, to say what old versions should be kept around, but I would at least keep the old versions *if* something else in the repo depends on them (I don't think anything does).

alalazo (approved):

alalazo:
Thanks @alecbcs !

PR #34597, opened 2022-12-17
xsbench: update version and add cuda support

pranav-sivaraman:
XSBench has new implementations as of version 19. For now I have added the CUDA version. Will continue working on updating this package and more variants


var/spack/repos/builtin/packages/xsbench/package.py
```diff
@@ -18,29 +18,41 @@ class Xsbench(MakefilePackage):
...
     version("13", sha256="b503ea468d3720a0369304924477b758b3d128c8074776233fa5d567b7ffcaa2")
 
     variant("mpi", default=True, description="Build with MPI support")
     variant("openmp", default=True, description="Build with OpenMP support")
+    variant("cuda", default=False, when="@19:", description="Build with CUDA support")
 
     depends_on("mpi", when="+mpi")
 
+    conflicts("cuda_arch=none", when="+cuda", msg="Must select a CUDA architecture")
+    conflicts("+cuda", when="+openmp", msg="OpenMP must be disabled to support CUDA")
```
alalazo, line 34: Can we make the `openmp` variant conditional on `~cuda`? Would that make sense?

pranav-sivaraman: I think that makes sense. XSBench is a mini-app so the goal would be to have all the different implementations installed and you can easily benchmark them. So all the variants are independent of each other and I just kept the default to OpenMP.

haampie (approved):

PR #34604, opened 2022-12-19
hpctoolkit: replace filter_file with upstream patch

mwkrentel:
Replace the filter_file for older configure with rocm 5.3 with an upstream patch.  Further, the patch is no longer needed for develop or later releases.


renjithravindrankannath (approved):
Looks good to me.

alalazo (approved):

PR #34605, opened 2022-12-19
gaudi: add tbb patch

joequant:
Add patch to gaudi that allows compilation with latest tbb


haampie:
If this patch makes it incompatible with older versions, please update the version bounds for intel-tbb.

wdconinc:
In gaudi v36, the TBB support was improved to include this patch among other things (https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1193). In gaudi v37, the support for newer TBB was improved even further. In https://github.com/spack/spack/pull/40802 this was reflected in the variants for the spack package of gaudi.

I will therefore close this PR as not applicable anymore. Thanks for the patch, and please keep the bug fixes coming as they are very much appreciated!

PR #34609, opened 2022-12-19
archspec: add support for zen4

alalazo:
Also add:
- [x] Upper bound for Xeon Phi compiler support
- [x] Better detection for a64fx


alalazo:
@spackbot run pipeline

haampie (approved):

PR #34612, opened 2022-12-19
Shrink the buildcache `index.json`

blue42u:
Based on #34360 and https://github.com/spack/spack/pull/34627. CC @haampie @scottwittenburg 

The `index.json` is huge (dozens of MB) and causes performance issues when it is downloaded. Some trivial changes can be made to improve this situation. This PR changes the `index.json` and `spec.json` code to output in minified form (i.e. unnecessary whitespace removed), and adds a new `index.json.gz`, a `gzip`-compressed version of the `index.json`, which is fetched first whenever available.

Based on the current [`index.json` from the `develop` buildcache](https://binaries.spack.io/develop/build_cache/index.json), these changes should collectively reduce the total network I/O by 20x:
| Changes | Size (MB) | Ratio |
|:--- | ---:| ---:|
| None | 184.73 | 1x |
| Minified | 88.43 | 2.09x |
| Minified + `gzip` | 9.102 | 20.3x |

Note that this also doubles the number of failing GET requests when a buildcache does not have an index. This should not be a major issue (`O(# mirrors)`) but could also be resolved in a later version.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -29,7 +29,7 @@ from llnl.util.filesystem import BaseDirectoryVisitor, mkdirp, visit_directory_t
 
 import spack.cmd
 import spack.config as config
-import spack.database as spack_db
+import spack.hash_types as ht
 import spack.hooks
 import spack.hooks.sbang
 import spack.mirror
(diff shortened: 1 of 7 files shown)


haampie:
Try `zstd` 🙃 for me its gets to 5.7MB without minification in less time.

Fwiw, I think the "normal" way to handle compression is through `Accept-Encoding: gzip` headers, that way you don't need to have a separate url for the compressed version.

(relevant: https://news.ycombinator.com/item?id=34059051)

blue42u:
| Changes | Size (MB) | Ratio |
|:--- | ---:| ---:|
| Minified + `zstd` | 5.74 | 32.2x |

@haampie Can you point me towards some current code that uses `zstd`? It's not coming up in my `grep`.

haampie:
We're not using it, but I wanna bootstrap `tar` + `zstd` to make that the default. Until then we're stuck with slow gzip

haampie:
Can you rebase on develop instead of #34360?

blue42u:
I can't rebase on `develop` since this code relies on the disentanglement from `Database` done by https://github.com/spack/spack/pull/34627. That is based on https://github.com/spack/spack/pull/34360 to reduce merge conflicts.

PR #34613, opened 2022-12-19
Boost: add version 1.81.0

hainest:


alalazo:
@spackbot run pipeline

alalazo (dismissed):

hainest:
@alalazo The change here didn't break the pipeline. I need to try to fix that PCH issue with %oneapi so we don't have to keep dealing with it on all of the Boost changes.

alalazo:
Can we work around that PCH issue for oneapi, until a proper fix is ready?

alalazo:
@spackbot run pipeline

hainest:
@alalazo The Dyninst build failure in ci/gitlab-ci is a known issue and has been fixed in dyninst/master. We can merge here.

alalazo:
Ok, can you constrain the version of boost that `dyninst` can use in this PR, so that pipelines pass? After that this is mergeable.

hainest:
>Ok, can you constrain the version of boost that dyninst can use in this PR, so that pipelines pass? After that this is mergeable.

It's not a Boost issue, it's an STL one. Newer versions of gcc and clang have cleaned up the transitive includes of their STL headers. There are a couple of places in Dyninst that were affected by this.

alalazo:
> It's not a Boost issue, it's an STL one. Newer versions of gcc and clang have cleaned up the transitive includes of their STL headers. There are a couple of places in Dyninst that were affected by this.

Do we know why this built in the first place then?[^1] I'd like to double check why dyninst was :heavy_check_mark: at some point, before force merging something that will break CI on `develop`.

[^1]: I didn't look at the logs in the previous comment, and thought it was a known incompatibility between recent boost and dyninst

hainest:
@alalazo The pipeline stall here will be fixed by #35630

alalazo (approved):

alalazo:
@spackbot run pipeline

hainest:
@alalazo Boost 1.82.0 is coming out next week. This is working now, so we can merge. Thanks for the help!

alalazo:
Thanks @hainest !

PR #34614, opened 2022-12-20
Create Windows package defaults

johnwparent:
Additionally update MSMPI external detection logic as we're now defaulting to that implementation to provide MPI.

### DIFF
--- a/var/spack/repos/builtin/packages/msmpi/package.py
+++ b/var/spack/repos/builtin/packages/msmpi/package.py
@@ -18,7 +18,7 @@ class Msmpi(Package):
     url = "https://github.com/microsoft/Microsoft-MPI/archive/refs/tags/v10.1.1.tar.gz"
     git = "https://github.com/microsoft/Microsoft-MPI.git"
 
-    executable = ["mpiexec.exe"]
+    executable = ["mpiexec"]
 
     version("10.1.1", sha256="63c7da941fc4ffb05a0f97bd54a67968c71f63389a0d162d3182eabba1beab3d")
     version("10.0.0", sha256="cfb53cf53c3cf0d4935ab58be13f013a0f7ccb1189109a5b8eea0fcfdcaef8c1")
@@ -31,9 +31,9 @@ class Msmpi(Package):
 
     @classmethod
     def determine_version(cls, exe):
-        output = Executable(exe)()
-        ver_str = re.search("[Version ([0-9.]+)]", output)
-        return Version(ver_str.group(0)) if ver_str else None
+        output = Executable(exe)(output=str, error=str)
+        ver_str = re.search(r"\[Version ([0-9.]+)\]", output)
+        return Version(ver_str.group(1)) if ver_str else None
 
 
 class GenericBuilder(GenericBuilder):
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/msmpi/package.py
```diff
@@ -31,9 +31,9 @@ class Msmpi(Package):
 
     @classmethod
     def determine_version(cls, exe):
-        output = Executable(exe)()
-        ver_str = re.search("[Version ([0-9.]+)]", output)
-        return Version(ver_str.group(0)) if ver_str else None
+        output = Executable(exe)(output=str, error=str)
+        ver_str = re.search(r"\[Version ([0-9.]+)\]", output)
+        return Version(ver_str.group(1)) if ver_str else None
```
scheibelp, line 41: I don't *think* this would run into the same problem as https://github.com/spack/spack/pull/34574, but if we don't anticipate MSMPI installations on non-windows systems, then we may want to either

* Refine the regular expression to be less generic
* Always return `None` on non-Windows for this package

scheibelp (commented):
I have a minor concern about the removal of `".exe"` (more info in comment)

var/spack/repos/builtin/packages/msmpi/package.py
johnwparent: I'm fine with returning none.

When searching for the actual binary in the external detection run, we append an additional "exe" extension onto whatever is in that executables list, so with the "exe" provided we end up searching "mpiexec.exe.exe". I'd rather keep the regex generic to support not specifying an extension for cross platform packages, and just force Windows specific to behave as such. FWIW I think this version regex does this for us as other `mpiexec` don't return the version in that format with no argument, but I'm also always a fan of being more explicit about it, I'll update.

scheibelp: > I'd rather keep the regex generic to support not specifying an extension for cross platform packages, 

To be clear, I was not referring to the regex used to identify the executable, but rather the regex used to extract the version. For example, right now something like `"Version 4.5"` would match, and it seems like other MPI implementations could easily have this string. If `msmpi` outputs something like `"Microsoft MPI Version 4.5"` then we could add that static string to the regex used in this function.

If `msmpi` could be installed on Linux, then I think refining the regex in this function is preferable. If it can't, then returning `None` on non-Windows would be fine.

johnwparent: I've updated to return none. MSMPI is intended to be a Windows implementation of mpich, so even if it could be installed, mpich would likely be preferable, to the point where I think returning none is fine.

I misunderstood, thanks for clarifying. That regex actually does need updating, I intended for there to be two literal brackets there that I forgot to include when I refactored to have that regex actually capture what we want from it. I'll add the msmpi string as well.

scheibelp (approved):

scheibelp:
@spackbot run pipeline

johnwparent:
@spackbot run pipeline

johnwparent:
@spackbot run pipeline

PR #34616, opened 2022-12-20
starpu: Add conflict for ~blocking +simgrid

lpoirel:
see https://gitlab.inria.fr/starpu/starpu/-/commit/1f5a911d43ba8e49f7229ab16f5dafdfdd596bd2

### DIFF
--- a/var/spack/repos/builtin/packages/starpu/package.py
+++ b/var/spack/repos/builtin/packages/starpu/package.py
@@ -94,6 +94,7 @@ class Starpu(AutotoolsPackage):
     )
 
     conflicts("+papi", when="+simgrid")
+    conflicts("~blocking", when="+simgrid")
 
     def autoreconf(self, spec, prefix):
         if not os.path.isfile("./configure"):


sthibaul:
I would have rather seen something like a depends_on, so that enabling blocking is automatic, rather than having the user have to specify it?

lpoirel:
> I would have rather seen something like a depends_on, so that enabling blocking is automatic, rather than having the user have to specify it?

Just adding `+simgrid` to starpu is enough to automatically infer `+blocking` from the conflict, thanks to the new concretizer.

sthibaul (approved):

sthibaul:
> > I would have rather seen something like a depends_on, so that enabling blocking is automatic, rather than having the user have to specify it?
> 
> Just adding `+simgrid` to starpu is enough to automatically infer `+blocking` from the conflict, thanks to the new concretizer.

Ah, ok, good :)

nfurmento (approved):

alalazo (approved):

PR #34618, opened 2022-12-20
py-scikit-image: add 0.19.3

manuelakuhn:
https://github.com/scikit-image/scikit-image/tree/v0.19.3


var/spack/repos/builtin/packages/py-scikit-image/package.py
```diff
@@ -29,40 +31,52 @@ class PyScikitImage(PythonPackage):
     depends_on("python@3.6:", when="@0.16:", type=("build", "link", "run"))
     depends_on("python@2.7:", when="@0.13:", type=("build", "link", "run"))
     depends_on("python@2.6:", type=("build", "link", "run"))
+    depends_on("py-setuptools@59.4:", when="@0.19.1:", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-setuptools@:59.4", when="@0.19.1:", type="build")
```

adamjstewart (changes requested):

adamjstewart (approved):

PR #33801, opened 2022-11-09
Trilinos Dependencies for Windows

jpopelar:
Includes Windows packaging for a number of Trilinos dependencies, including hdf5, boost, openblas, py-setuptools

PR #34622, opened 2022-12-20
Trilinos Dependencies For Windows v2

jpopelar:
For porting of packages that Trilinos requires to build. A far cleaner and more recent version of #33801. Will include boost and py-setuptools to start with; will need to consult with others about makefile packages and whether it is fine to port them over to CMake (esp. with openblas)


var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -476,7 +476,15 @@ def determine_bootstrap_options(self, spec, with_libs, options):
         with open("user-config.jam", "w") as f:
             # Boost may end up using gcc even though clang+gfortran is set in
             # compilers.yaml. Make sure this does not happen:
-            f.write("using {0} : : {1} ;\n".format(boost_toolset_id, spack_cxx))
+            if not spec.satisfies("%intel") and not spec.satisfies("platform=windows"):
```
hainest: Was this happening with just the classic intel compiler or with oneapi, as well?

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -495,6 +503,8 @@ def determine_bootstrap_options(self, spec, with_libs, options):
             if "+python" in spec:
                 f.write(self.bjam_python_line(spec))
 
+        options.append("-show-libraries")
```
hainest: Is this some debug logging to-be-removed?

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -665,12 +676,15 @@ def install(self, spec, prefix):
         # to make Boost find the user-config.jam
         env["BOOST_BUILD_PATH"] = self.stage.source_path
 
-        bootstrap = Executable("./bootstrap.sh")
-
-        bootstrap_options = ["--prefix=%s" % prefix]
+        bootstrap_options = []
```
hainest: The `"--prefix=%s"` wasn't propagated to the `else` clause below. Was that intentional?

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -682,21 +696,33 @@ def install(self, spec, prefix):
 
         # b2 used to be called bjam, before 1.47 (sigh)
         b2name = "./b2" if spec.satisfies("@1.47:") else "./bjam"
+        if self.spec.satisfies("platform=windows"):
+            b2name = "b2.exe" if spec.satisfies("@1.47:") else "bjam.exe"
 
         b2 = Executable(b2name)
         jobs = make_jobs
+        path_to_config = ""
```
hainest: Could we merge this with line 717 so we only do the check once?

var/spack/repos/builtin/packages/pcre/package.py
```diff
@@ -51,3 +51,25 @@ def configure_args(self):
...
+            args.append("-DPCRE_SUPPORT_UTF:BOOL=ON")
+            args.append("-DPCRE_SUPPORT_UNICODE_PROPERTIES:BOOL=ON")
+
+        return args
+
+    def cmake(self, spec, prefix):
+        if self.spec.satisfies("platform=windows"):
+            super(Pcre, self).cmake(spec, prefix)
+        else:
+            configure("--prefix=" + prefix, *self.configure_args())
```
scheibelp: Could you refactor this to be a multi-build-system package? i.e. inherit `CMakePackage` and `AutotoolsPackage` etc.

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -476,7 +476,15 @@ def determine_bootstrap_options(self, spec, with_libs, options):
...
             # compilers.yaml. Make sure this does not happen:
-            f.write("using {0} : : {1} ;\n".format(boost_toolset_id, spack_cxx))
+            if not spec.satisfies("%intel") and not spec.satisfies("platform=windows"):
+                # using intel-linux : : spack_cxx in user-config.jam leads to
+                # error: at project-config.jam:12
+                # error: duplicate initialization of intel-linux with the following parameters:  # noqa
+                # error: version = <unspecified>
+                # error: previous initialization at ./user-config.jam:1
+                # We also use a hack right now to get around the fact we don't have
+                # a wrapper in spack/env for cl.exe
```
scheibelp:
```
# We also use a hack right now to get around the fact we don't have
# a wrapper in spack/env for cl.exe
```

I think it would be clearer to rephrase this as:

```
There is no spack wrapper for cl.exe, so skip this on Windows
```

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -476,7 +476,15 @@ def determine_bootstrap_options(self, spec, with_libs, options):
         with open("user-config.jam", "w") as f:
             # Boost may end up using gcc even though clang+gfortran is set in
             # compilers.yaml. Make sure this does not happen:
-            f.write("using {0} : : {1} ;\n".format(boost_toolset_id, spack_cxx))
+            if not spec.satisfies("%intel") and not spec.satisfies("platform=windows"):
+                # using intel-linux : : spack_cxx in user-config.jam leads to
+                # error: at project-config.jam:12
+                # error: duplicate initialization of intel-linux with the following parameters:  # noqa
+                # error: version = <unspecified>
+                # error: previous initialization at ./user-config.jam:1
```
scheibelp: This appears to undo https://github.com/spack/spack/commit/38592b9e4f974254c18ce82e8b7ccca65a07539e - have you confirmed there was an issue with that change on Linux?

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -562,13 +572,14 @@ def determine_b2_options(self, spec, options):
...
 
-        if not spec.satisfies("@:1.75 %intel"):
+        if not (spec.satisfies("@:1.75 %intel") or spec.satisfies("platform=windows")):
             # When building any version >= 1.76, the toolset must be specified.
             # Earlier versions could not specify Intel as the toolset
             # as that was considered to be redundant/conflicting with
             # --with-toolset in bootstrap.
             # (although it is not currently known if 1.76 is the earliest
             # version that requires specifying the toolset for Intel)
+            # And again, with no cl.exe/ifx wrapper this will fail on Windows
```
scheibelp: Avoid using the term "again": it makes sense within the context of the PR, but once the changes get merged into the file, it provides no real context.

var/spack/repos/builtin/packages/boost/package.py
scheibelp: (I might be saying the same thing as the above user but wanted to make sure to ask) With this change, I don't see `prefix` being specified anywhere: how is boost installing to the appropriate directory?

scheibelp (changes requested):
Initial requests

jpopelar:
@scheibelp PTAL when convenient. Will fix style issues momentarily and also email you a text doc with some lingering trilinos issues (mostly related to the shared and mpi variants)

var/spack/repos/builtin/packages/boost/package.py
scheibelp: This is still true, but will be untrue pending https://github.com/spack/spack/pull/38495 (which is not merged yet and probably won't be merged until after this, but I note it here for reference).

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -6,18 +6,20 @@
...
-class Openblas(MakefilePackage):
+class Openblas(CMakePackage, MakefilePackage):
     """OpenBLAS: An optimized BLAS library"""
 
     homepage = "https://www.openblas.net"
     url = "https://github.com/xianyi/OpenBLAS/releases/download/v0.2.19/OpenBLAS-0.2.19.tar.gz"
     git = "https://github.com/xianyi/OpenBLAS.git"
 
-    libraries = ["libopenblas"]
+    libraries = ["libopenblas", "openblas"]
```
scheibelp, line 24: (no change needed) Note this attribute is just for finding external instances. We might want to automatically strip `lib` on Windows in `detection/path.py` (doesn't have to happen in this PR)

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -209,6 +214,9 @@ class Openblas(MakefilePackage):
     )
 
     depends_on("perl", type="build")
+    depends_on("cmake", when="platform=windows")
```
scheibelp: Note: `makefile.py` already declares a conflict on Windows when `build_system=makefile`, so this should be unnecessary (also, this doesn't technically force using the Cmake build system, but that's a more-abstract reason).

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -246,6 +254,35 @@ def check_compilers(self):
...
+        suffix = spec.variants["symbol_suffix"].value
+        if suffix != "none":
+            name += suffix
+
+        return find_libraries(name, spec.prefix, shared=search_shared, recursive=True)
+
+
+class SetupEnvironment(object):
+    def setup_build_environment(self, env):
+        pass
```
scheibelp: This class appears unnecessary.

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -246,6 +254,35 @@ def check_compilers(self):
...
+        # one of the source files implementing functions declared in these
+        # headers.
+        return find_headers(["cblas", "lapacke"], self.prefix.include)
+
+    @property
+    def libs(self):
+        spec = self.spec
+
+        # Look for openblas{symbol_suffix}
+        name = "libopenblas"
```
scheibelp: Note: if you had to specify `libraries = [libopenblas, openblas]` above, I'd be surprised if you didn't have to do it here.

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -496,3 +511,28 @@ def check_install(self):
 
         output = compile_c_and_execute(source_file, [include_flags], link_flags.split())
         compare_output_file(output, blessed_file)
+
+
+class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder, SetupEnvironment):
+    def cmake_args(self):
+        cmake_defs = [self.define("TARGET", "GENERIC")]
```
scheibelp, line 515: Because you set CMake as the default build system everywhere (including Linux), we'd be particularly interested in supporting all the variants and build functionality we supported with Makefiles, which includes the variants:

* shared
* fortran
* locking
* ilp64
* consistent_fpcsr
* bignuma

I’m guessing you’d also want to promote _microarch_target_args from a Makefile-specific method (although `GENERIC` is the sensible choice on Windows for now, we'd want to be specific when building with CMake on  other systems), unless you see CMake handling this itself.

We could keep the CMake build stripped-down, but I'd say in that case it should only be available on Windows.

var/spack/repos/builtin/packages/superlu/package.py
```diff
@@ -43,7 +43,10 @@ class Superlu(CMakePackage, Package):
     )
 
     build_system(
-        conditional("cmake", when="@5:"), conditional("generic", when="@:4"), default="cmake"
+        conditional("cmake", when="@5:"),
+        conditional("autotools", when="@:4"),
```
scheibelp: This should be `conditional(generic, when=@:4)` (this inherits `Package` rather than `AutotoolsPackage`).

var/spack/repos/builtin/packages/openblas/ifort-msvc.patch
```diff
@@ -0,0 +1,13 @@
+diff -ruN spack-src/cmake/fc.cmake spack-src-new/cmake/fc.cmake
+--- spack-src/cmake/fc.cmake	2023-04-01 14:18:01.000000000 -0600
++++ spack-src-new/cmake/fc.cmake	2023-06-06 09:34:12.921982500 -0600
+@@ -89,6 +89,9 @@
+ 
+ if (${F_COMPILER} STREQUAL "INTEL")
+   set(CCOMMON_OPT "${CCOMMON_OPT} -DF_INTERFACE_INTEL")
++  if (MSVC)
++    set(FCOMMON_OPT "${FCOMMON_OPT} -names:uppercase -assume:underscore")
++  endif ()
```
scheibelp, line 10: (question) Do you know if there are any pending PRs in Boost itself for this?

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -694,21 +709,30 @@ def install(self, spec, prefix):
...
-        ]
+        b2_options = ["-j", "%s" % jobs]
+        if not self.spec.satisfies("platform=windows"):
+            b2_options.append(path_to_config)
 
         threading_opts = self.determine_b2_options(spec, b2_options)
 
+        # Windows just wants a b2 call with no args
+        if self.spec.satisfies("platform=windows"):
+            b2_options.clear()
```
scheibelp: Don't clear this at the end, do something like:

```
if spec.satisfies(platform=windows)
  b2_options = []
else:
  all the logic to set b2 options
```

I'll need to read this carefully to know if there's any options in `b2_options` that are not specified in `bootstrap_options`, and if so my concern would be that we aren't customizing the build as we would want.

var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -362,6 +363,18 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+stokhos", when="%xl")
     conflicts("+stokhos", when="%xl_r")
 
+    # Current Windows support, only have serial static builds
+    conflicts(
+        "+shared",
+        when="platform=windows",
+        msg="Only serial static builds are supported on Windows currently.",
```
scheibelp: Customize each message to the variant, e.g.:

```
Only static builds are supported on Windows currently
```

somewhat repetitious, but that allows us to address each of these in potentially separate PRs without the message becoming wrong.

scheibelp (changes requested):
Additional requests

var/spack/repos/builtin/packages/openblas/ifort-msvc.patch
jpopelar: Not that I am aware of, no, but I can verify

var/spack/repos/builtin/packages/boost/package.py
jpopelar: For sure. Either way, cleaned up for posterity in next commit

jpopelar:
@scheibelp Looks like prechecks are passing. PTAL whenever convenient and we'll get this reviewed and merged asap

var/spack/repos/builtin/packages/boost/package.py
scheibelp: Bump on this question.

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -507,6 +509,8 @@ def determine_bootstrap_options(self, spec, with_libs, options):
             if "+python" in spec:
                 f.write(self.bjam_python_line(spec))
 
+        options.append("--prefix=%s" % prefix)
```
scheibelp: Originally you removed the `--prefix=` from install so it was set nowhere (i.e. https://github.com/spack/spack/pull/34622#discussion_r1063894135). Now you restored that and set it here, so it's set twice.

var/spack/repos/builtin/packages/boost/package.py
scheibelp: Bump: you can move 

```
            path_to_config = "--user-config=%s" % os.path.join(
                self.stage.source_path, "user-config.jam"
            )
```

like

```
        if not is_windows:
            path_to_config = "--user-config=%s" % os.path.join(
                self.stage.source_path, "user-config.jam"
            )
            b2_options = ["-j", "%s" % jobs]
            b2_options.append(path_to_config)
```

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -694,20 +701,27 @@ def install(self, spec, prefix):
...
-            "--user-config=%s" % os.path.join(self.stage.source_path, "user-config.jam"),
-        ]
+        # Windows just wants a b2 call with no args
+        b2_options = []
+        if not self.spec.satisfies("platform=windows"):
+            b2_options = ["-j", "%s" % jobs]
+            b2_options.append(path_to_config)
 
-        threading_opts = self.determine_b2_options(spec, b2_options)
+            threading_opts = self.determine_b2_options(spec, b2_options)
```
scheibelp: How does this install succeed on windows, as below we have

```
        for threading_opt in threading_opts:
            b2("install", "threading=%s" % threading_opt, *b2_options)
```

(including on Windows) but the `threading_opts` variable is not defined on Windows.

var/spack/repos/builtin/packages/openblas/ifort-msvc.patch
scheibelp: Checking in on this

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -88,6 +90,9 @@ class Openblas(MakefilePackage):
     provides("lapack@3.9.1:", when="@0.3.15:")
     provides("lapack@3.7.0", when="@0.2.20")
 
+    # https://github.com/xianyi/OpenBLAS/pull/2519/files
```
scheibelp, line 96: It's great that you linked this, but does this mean that only certain versions need to apply this patch (e.g. it looks like that PR was merged in 2020, so perhaps all versions after that don't need this patch)?

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -246,6 +253,30 @@ def check_compilers(self):
                 + " has no Fortran compiler added in spack. Add it or use openblas~fortran!"
             )
 
+    @property
+    def headers(self):
+        # As in netlib-lapack, the only public headers for cblas and lapacke in
```
scheibelp: Drop the "as in netlib-lapack" (this won't provide much context to people).

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -246,6 +253,30 @@ def check_compilers(self):
...
+        # one of the source files implementing functions declared in these
+        # headers.
+        return find_headers(["cblas", "lapacke"], self.prefix.include)
+
+    @property
+    def libs(self):
+        spec = self.spec
+
+        # Look for openblas{symbol_suffix}
+        name = "libopenblas, openblas"
```
scheibelp: Note that the first argument of `find_libraries` allows a string, but only if that string represents a single library. If you want to look for multiple, you need to break it into a list of strings.

Also, you'd need to individually append the symbol suffix anyway. That being said, I don't see any `symbol_suffix` variant in `boost` right now and I don't see this diff adding one. Without the `symbol_suffix` logic, this matches the behavior of `Spec._libs_default_handler`.

var/spack/repos/builtin/packages/openblas/package.py
scheibelp: You may have added a few of these options since, but I still don't see handling of `ilp64`, `consistent_fpcsr`, `locking`, or `bignuma` (so I'd still be inclined not to make `CMake` the default on systems other than Windows).

var/spack/repos/builtin/packages/superlu/package.py
```diff
@@ -43,7 +43,10 @@ class Superlu(CMakePackage, Package):
     )
 
     build_system(
-        conditional("cmake", when="@5:"), conditional("generic", when="@:4"), default="cmake"
+        conditional("cmake", when="@5:"),
+        conditional("generic", when="@:4"),
+        conditional("cmake", when="platform=windows"),
```
scheibelp: Do you know if cmake builds were available for versions before 5? I would guess they weren't in which case this should be replaced with a separate `conflicts(platform=windows, when=@:4)`.

var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -375,7 +388,10 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("cgns", when="+exodus")
     depends_on("cmake@3.23:", type="build", when="@14.0.0:")
     depends_on("hdf5+hl", when="+hdf5")
-    depends_on("hypre~internal-superlu~int64", when="+hypre")
+    # Go through CMake for Hypre on Windows
```
scheibelp: Can you remove this comment (it's not wrong, but I don't think it's necessary)?

scheibelp (changes requested):

var/spack/repos/builtin/packages/openblas/package.py
jpopelar: Decent point! I can try building after my next commit here and let you know

var/spack/repos/builtin/packages/openblas/ifort-msvc.patch
jpopelar: Just checked, the only recent PR that even mentions MSVC is closed and only touches something in the command prompt. Hence I believe this patch is still required

var/spack/repos/builtin/packages/boost/package.py
jpopelar: I'm honestly not completely sure; regardless I've insulated it better now

var/spack/repos/builtin/packages/boost/package.py
jpopelar: I had this on oneAPI as well I believe, though I'd need to review what happened

var/spack/repos/builtin/packages/superlu/package.py
```diff
@@ -42,9 +42,7 @@ class Superlu(CMakePackage, Package):
         url="https://crd-legacy.lbl.gov/~xiaoye/SuperLU/superlu_4.2.tar.gz",
     )
 
-    build_system(
-        conditional("cmake", when="@5:"), conditional("generic", when="@:4"), default="cmake"
-    )
```
scheibelp, line 47: This `build_system` declaration is still required. If you restore it as it was, in combination with the conflict you added just below, the desirable result should be achieved in all circumstances.

Also I realized that replacing the `conflict` with:

```
require(build_system=cmake, when="platform=windows")
```

would be slightly more robust: for now since the support ranges do not overlap they are equivalent, but if they ever did for some reason, I assume we would want to force building with CMake on Windows.

var/spack/repos/builtin/packages/boost/package.py
scheibelp: You no longer need to set a default value for `path_to_config` here.

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -246,6 +253,30 @@ def check_compilers(self):
...
+        spec = self.spec
+
+        # Look for openblas{symbol_suffix}
+        name = ["libopenblas", "openblas.lib"]
+        search_shared = bool(spec.variants["shared"].value)
+        suffix = spec.variants["symbol_suffix"].value
+        if suffix != "none":
+            name += suffix
+
+        return find_libraries(name, spec.prefix, shared=search_shared, recursive=True)
```
scheibelp, line 279: I'm already seeing an existing implementation of libs/headers in openblas, and I think it would cover what you need. Note that other implementation does not search for "openblas.lib", but `find_libraries` actually expects to do all the suffix appending itself, so in fact we should not be passing "openblas.lib" to it.

var/spack/repos/builtin/packages/boost/package.py
scheibelp: I'll also note that `determine_b2_options` determines more than just threading options. Are all of the options handled by that function not relevant on Windows? I see for example the following variants handled:

* `+debug`
* `+icu`

And other important-seeming things not determined by variants (e.g. c++ standard).

My guess would be that you'd want to do this everywhere, but keep the below if (where on Windows you don't invoke `b2` with `threading=`).

scheibelp (changes requested):

var/spack/repos/builtin/packages/superlu/package.py
jpopelar: Are we sure that syntax is correct? I currently have

```
    build_system(
        conditional("cmake", when="@5:"), conditional("generic", when="@:4"), default="cmake"
    )

    requires(build_system="cmake", when="platform=windows")
```

CI then tells me that I can't use build_system in a requires() directive

var/spack/repos/builtin/packages/boost/package.py
scheibelp: I'm looking closer at https://github.com/spack/spack/commit/38592b9e4f974254c18ce82e8b7ccca65a07539e and I think this is also on account of a merge conflict: I think you can make the if statement just:

```
if not spec.satisfies("platform=windows"):
```

var/spack/repos/builtin/packages/superlu/package.py
```diff
@@ -46,6 +46,8 @@ class Superlu(CMakePackage, Package):
         conditional("cmake", when="@5:"), conditional("generic", when="@:4"), default="cmake"
     )
 
+    requires("cmake", when="platform=windows")
```
scheibelp: This needs to be `requires("build_system=cmake", when=...)`

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -246,6 +253,30 @@ def check_compilers(self):
...
+        # one of the source files implementing functions declared in these
+        # headers.
+        return find_headers(["cblas", "lapacke"], self.prefix.include)
+
+    @property
+    def libs(self):
+        spec = self.spec
+
+        # Look for openblas{symbol_suffix}
+        name = ["libopenblas", "openblas.lib"]
```
scheibelp: This should be:

```
name = ["libopenblas", "openblas"]
```

(`find_libraries` just wants the name, and will append `.lib` itself).

var/spack/repos/builtin/packages/boost/package.py
```diff
@@ -581,7 +583,7 @@ def determine_b2_options(self, spec, options):
 
         options.extend(["link=%s" % ",".join(link_types), "--layout=%s" % layout])
 
-        if not spec.satisfies("@:1.75 %intel"):
+        if not (spec.satisfies("@:1.75 %intel") or spec.satisfies("platform=windows")):
```
scheibelp: I'm assuming the idea is you don't want to set `toolset=` if we're on windows, in that case, this `if` would need to be:

`if not (spec.satisfies("@:1.75 %intel") and not spec.satisfies("platform=windows")):`

scheibelp (changes requested):
I have a few minor requests. Once those are resolved, can you remove the draft status? The CI will start building some of these packages.

var/spack/repos/builtin/packages/superlu/package.py
jpopelar: So I tried that earlier but it broke prechecks. Apparently you can't have `build_system` stuff inside a requires block. I figured this would be the closest to what we want? Is there another way to do this?

var/spack/repos/builtin/packages/boost/package.py
jpopelar: Sorry, I deMorgan's lawed that one. Your method is definitely clearer

var/spack/repos/builtin/packages/superlu/package.py
jpopelar: @scheibelp Just making sure you saw this

jpopelar: @scheibelp Bump. I'll rebase and merge as soon as I get an answer on this

var/spack/repos/builtin/packages/boost/package.py
scheibelp: I misread what you originally had, and I should have been OK with it. Thanks for making the change (either formulation is fine).

jpopelar:
@scheibelp PTAL when convenient, I think this is good to go after I rebase

scheibelp (dismissed):

scheibelp (approved):

var/spack/repos/builtin/packages/openblas/package.py
```diff
@@ -242,13 +249,37 @@ def check_compilers(self):
...
+    @property
+    def libs(self):
+        spec = self.spec
+
+        # Look for openblas{symbol_suffix}
+        name = ["libopenblas", "openblas"]
+        search_shared = bool(spec.variants["shared"].value)
+        suffix = spec.variants["symbol_suffix"].value
+        if suffix != "none":
+            name += suffix
```
johnwparent, line 277: How does this work? Concatenating a string and a list like this results in the string being decomposed into chars and each char being added to the list, so at best this is a no op polluting the lib list?
@scheibelp

johnwparent: Sorry to comment on a dead PR, I just went to look at this for the PATH issue when considering how to handle virtual deps and this jumped out at me.
Happy to submit a patch if I I'm not wrong about this. I've got one locally.

scheibelp: > How does this work?

It almost definitely doesn't. A patch would be great.

johnwparent: Ok, one will be up shortly!

PR #34623, opened 2022-12-20
FileCache: Delete the new cache file on exception

blue42u:
The code in `FileCache` for `write_transaction` attempts to delete the temporary file when an exception occurs under the context by calling `shutil.rmtree`. However, [`rmtree` only operates on directories](https://docs.python.org/3/library/shutil.html#shutil.rmtree) while the rest of `FileCache` uses normal files. This causes an empty file to be written to the cache key when unneeded.

Use `os.remove` instead which [operates on normal files](https://docs.python.org/3/library/os.html#os.remove).

### DIFF
--- a/lib/spack/spack/test/util/file_cache.py
+++ b/lib/spack/spack/test/util/file_cache.py
@@ -33,6 +33,7 @@ def test_write_and_read_cache_file(file_cache):
         assert text == "foobar\n"
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Locks not supported on Windows")
 def test_failed_write_and_read_cache_file(file_cache):
     """Test failing to write then attempting to read a cached file."""
     with pytest.raises(RuntimeError, match=r"^foobar$"):
--- a/lib/spack/spack/test/util/file_cache.py
+++ b/lib/spack/spack/test/util/file_cache.py
@@ -5,7 +5,6 @@
 
 """Test Spack's FileCache."""
 import os
-import re
 import sys
 
 import pytest
--- a/lib/spack/spack/test/util/file_cache.py
+++ b/lib/spack/spack/test/util/file_cache.py
@@ -43,9 +43,7 @@ def test_failed_write_and_read_cache_file(file_cache):
             raise RuntimeError("foobar")
 
     # Cache dir should have exactly one (lock) file
-    files = os.listdir(file_cache.root)
-    assert len(files) <= 1
-    assert not files or re.fullmatch(r"^\..*\.lock$", files[0])
+    assert os.listdir(file_cache.root) == [".test.yaml.lock"]
 
     # File does not exist
     assert not file_cache.init_entry("test.yaml")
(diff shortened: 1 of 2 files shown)


lib/spack/spack/test/util/file_cache.py
```diff
@@ -32,6 +33,28 @@ def test_write_and_read_cache_file(file_cache):
         assert text == "foobar\n"
 
 
+def test_failed_write_and_read_cache_file(file_cache):
+    """Test failing to write then attempting to read a cached file."""
+    with pytest.raises(RuntimeError, match=r"^foobar$"):
```
haampie, line 38:
```suggestion
    with pytest.raises(RuntimeError, match="foobar"):
```

lib/spack/spack/test/util/file_cache.py
```diff
@@ -32,6 +33,28 @@ def test_write_and_read_cache_file(file_cache):
...
+        with file_cache.write_transaction("test.yaml") as (old, new):
+            assert old is None
+            assert new is not None
+            raise RuntimeError("foobar")
+
+    # Cache dir should have exactly one (lock) file
+    files = os.listdir(file_cache.root)
+    assert len(files) <= 1
+    assert not files or re.fullmatch(r"^\..*\.lock$", files[0])
```
haampie:
```suggestion
    # Cache dir should have exactly one (lock) file
    assert os.listdir(file_cache.root) == [".test.yaml.lock"]
```

lib/spack/spack/util/file_cache.py
```diff
@@ -145,7 +145,7 @@ def __exit__(cm, type, value, traceback):
 
                 if value:
                     # remove tmp on exception & raise it
```
haampie:
```suggestion
```

haampie: I don't think this comment was relevant either

haampie (changes requested):

lib/spack/spack/test/util/file_cache.py
```diff
@@ -32,6 +33,28 @@ def test_write_and_read_cache_file(file_cache):
         assert text == "foobar\n"
 
 
+def test_failed_write_and_read_cache_file(file_cache):
```
haampie, line 36:
```suggestion
@pytest.mark.skipif(sys.platform == "win32", reason="Locks not supported on Windows")
def test_failed_write_and_read_cache_file(file_cache):
```

haampie (changes requested):

lib/spack/spack/test/util/file_cache.py
haampie: It would be sad if Windows would influence the correctness on non-Windows.

haampie:
Thanks for splitting this out, since this can be backported.

lib/spack/spack/test/util/file_cache.py
```diff
@@ -5,6 +5,7 @@
 
 """Test Spack's FileCache."""
 import os
+import re
```
haampie:
```suggestion
```

haampie:
@spackbot run pipeline

haampie (approved):

PR #34626, opened 2022-12-20
py-cartopy: older versions don't support Python 3.10

adamjstewart:
Discovered this when the concretizer decided to install an ancient version of cartopy for some reason. When building with Python 3.10, I see:
```
error: no member named 'use_tracing' in '_ts'
```
Either older versions are pre-cythonized or simply aren't compatible with Python 3.10.

Also replaced `import_modules` with `skip_modules` to simplify the code a bit.


adamjstewart:
@spackbot run pipeline

alalazo (approved):

PR #34633, opened 2022-12-20
py-pyproj: add new versions

adamjstewart:
Needed 3.3.1 to get a newer version (that doesn't use distutils) and an older version (that doesn't require newer setuptools).


adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipelines

alalazo (approved):

PR #34637, opened 2022-12-21
spack graph: rework to use Jinja templates and builders

alalazo:
`spack graph` has been reworked to use:
- Jinja templates
- builder objects to construct the template context 
 
when DOT graphs are requested. This allowed to add a new colored output for DOT graphs that highlights both the dependency types and the nodes that are needed at runtime for a given spec. Here is an example showing the graph for `lbann`:

lbann

In red are the nodes that are needed to use `lbann` at runtime, in `blue` the node that are only needed to build `lbann` from sources. The edges are also drawn differently according to their dependency type.

Modifications:
- [x] Restructure `graph_dot` to use Jinja2 templates and builder objects
- [x] Extract a separate function for static graph in DOT format
- [x] Removed `spack.graph` implementation of topological sort in favor of the one in `spack.traverse`
- [x] Add a `--color` option to `spack graph` to output DOT graphs like the one shown above
- [x] Polished code (according to `pylint` default warnings) and added type hints

### DIFF
--- a/lib/spack/docs/conf.py
+++ b/lib/spack/docs/conf.py
@@ -200,12 +200,14 @@ nitpick_ignore = [
     ("py:class", "_frozen_importlib_external.SourceFileLoader"),
     ("py:class", "clingo.Control"),
     ("py:class", "six.moves.urllib.parse.ParseResult"),
+    ("py:class", "TextIO"),
     # Spack classes that are private and we don't want to expose
     ("py:class", "spack.provider_index._IndexBase"),
     ("py:class", "spack.repo._PrependFileLoader"),
     ("py:class", "spack.build_systems._checks.BaseBuilder"),
     # Spack classes that intersphinx is unable to resolve
     ("py:class", "spack.version.VersionBase"),
+    ("py:class", "spack.spec.DependencySpec"),
 ]
 
 # The reST default role (used for this markup: `text`) to use for all documents.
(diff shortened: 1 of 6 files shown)


alalazo:
@tgamblin @becker33 @haampie I think this PR will be handy to discuss graph changes for _compilers as dependencies_ Here the `py-shapely %oneapi` example drawn with the options added here:

shapely

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

haampie:
Should we make build deps red, cause they can be gc'd ;p

alalazo:
> Should we make build deps red

You mean the node `fillcolor`, right? I'd be fine with that

alalazo:
@haampie Done

haampie (approved):

haampie:
@spackbot run pipeline

haampie:
Maybe one style issue: currently the edge colors are blue going to build deps, maybe keep them the same color?

Also, maybe we can just do `style="dashed"` for build deps type edges

alalazo:
> Also, maybe we can just do style="dashed" for build deps type edges

I found a way to have multiple colors alongside each other for e.g. build/run deps. I can check if I can do the same for dashed build type edges. I'd merge this as is, since it's :heavy_check_mark: and welcome any style adjustments in following PRs.

PR #34641, opened 2022-12-21
etags for index.json invalidation, test coverage

haampie:
Implement an alternative strategy to do index.json invalidation.

The current approach of pairs of index.json / index.json.hash is
problematic because it leads to races.

The standard solution for cache invalidation is etags, which are
supported by both http and s3 protocols, which allows one to do
conditional fetches.

This PR implements that for the http/https schemes. It should also work
for s3 schemes, but that requires other prs to be merged.

Also it improves unit tests for index.json fetches.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -9,12 +9,16 @@ import hashlib
 import json
 import multiprocessing.pool
 import os
+import re
 import shutil
 import sys
 import tarfile
 import tempfile
 import time
 import traceback
+import urllib.error
+import urllib.parse
+import urllib.request
 import warnings
 from contextlib import closing
 from urllib.error import HTTPError, URLError
@@ -342,7 +346,6 @@ class BinaryCacheIndex(object):
 
         for cached_mirror_url in self._local_index_cache:
             cache_entry = self._local_index_cache[cached_mirror_url]
-            cached_index_hash = cache_entry["index_hash"]
             cached_index_path = cache_entry["index_path"]
             if cached_mirror_url in configured_mirror_urls:
                 # Only do a fetch if the last fetch was longer than TTL ago
@@ -361,13 +364,14 @@ class BinaryCacheIndex(object):
                     # May need to fetch the index and update the local caches
                     try:
                         needs_regen = self._fetch_and_cache_index(
-                            cached_mirror_url, expect_hash=cached_index_hash
+                            cached_mirror_url,
+                            cache_entry=cache_entry,
                         )
                         self._last_fetch_times[cached_mirror_url] = (now, True)
                         all_methods_failed = False
-                    except FetchCacheError as fetch_error:
+                    except FetchIndexError as e:
                         needs_regen = False
-                        fetch_errors.extend(fetch_error.errors)
+                        fetch_errors.append(e)
                         self._last_fetch_times[cached_mirror_url] = (now, False)
                     # The need to regenerate implies a need to clear as well.
                     spec_cache_clear_needed |= needs_regen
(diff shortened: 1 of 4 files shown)


haampie:
@blue42u the stuff you're overhauling has really bad test coverage, so I'm inclined to get this in first.

lib/spack/spack/binary_distribution.py
```diff
@@ -2413,3 +2369,118 @@ def __call__(self, spec, **kwargs):
...
+            return FetchCacheError("Remote index {} is invalid".format(url_index), e)
+
+        computed_hash = compute_hash(result)
+
+        if computed_hash != remote_hash:
+            raise FetchIndexError(
+                "Remote hash {} of {} does not match computed hash {}".format(
+                    remote_hash, url_index, computed_hash
+                )
+            )
```
blue42u: This can be caused by race conditions, we should not error on the reading end if the computed vs. remote hashes are different. The `computed_hash` should be persisted in the cache.

lib/spack/spack/binary_distribution.py
```diff
@@ -2413,3 +2369,118 @@ def __call__(self, spec, **kwargs):
...
+
+        # Validate the hash
+        remote_hash = response.read(64)
+        if not re.match(rb"[a-f\d]{64}$", remote_hash):
+            raise FetchIndexError("Invalid hash format in {}".format(url_index_hash))
+        remote_hash = remote_hash.decode("utf-8")
+
+        # No need to update further
+        if self.local_hash and remote_hash == self.local_hash:
+            return FetchIndexResult(etag=None, hash=None, data=None, fresh=True)
```
blue42u: If you do not have a previous fetch's hash, skip fetching the `index.json.hash` here since you always will need to fetch the `index.json` anyway. (Saves an unneeded request.)

lib/spack/spack/binary_distribution.py
```diff
@@ -623,7 +577,9 @@ class UnsignedPackageException(spack.error.SpackError):
 
 
 def compute_hash(data):
-    return hashlib.sha256(data.encode("utf-8")).hexdigest()
+    if not isinstance(data, bytes):
```
blue42u:
```suggestion
    if isinstance(data, str):
```

blue42u (changes requested):
I'm more or less fine with landing this first. See behavioral concerns below.

lib/spack/spack/binary_distribution.py
haampie: Yeah, I agree that keeping `index.json.hash` as a cache-identifier only is the way. But on develop it is also used to verify the integrity of index.json.

Edit: actually failing to fetch index.json.hash was not an error, but mismatch was :thinking: so i'm not even keeping original behavior.

haampie: Ok, I think I'll change that behavior to make `index.json.hash` a cache-identifier only.

lib/spack/spack/binary_distribution.py
haampie: Yup, that implies index.json.hash is only used as a cache identifier. I just pushed a commit that does that

blue42u (approved):

haampie:
It would be nice if Gitlab CI would work here (and everywhere else too)

alalazo:
@spackbot run pipeline

haampie:
@spackbot rebuild everything

haampie:
@spackbot run pipeline

alalazo (approved):
@haampie and @blue42u agreed on this, so merging! Thanks!

PR #34643, opened 2022-12-21
index.json.hash, no fatal error if key cannot be fetched

haampie:
Not being able to fetch `index.json.hash` shouldn't prevent fetching `index.json`.

Also set the Spack `User-Agent`


haampie:
@spackbot run pipeline

alalazo (approved):

PR #34648, opened 2022-12-21
Bump memory requests for e4s/ml generate jobs

mvandenburgh:
Bumps memory requests from 8G => 16G for some stacks. These particular stacks have consistently shown memory usage higher than 8G, which has led to pods being `OOMKilled`.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -289,10 +289,14 @@ protected-publish:
 e4s-pr-generate:
   extends: [ ".e4s", ".pr-generate"]
   image: ecpe4s/ubuntu20.04-runner-x86_64:2023-01-01
+  variables:
+    KUBERNETES_MEMORY_REQUEST: 16G
 
 e4s-protected-generate:
   extends: [ ".e4s", ".protected-generate"]
   image: ecpe4s/ubuntu20.04-runner-x86_64:2023-01-01
+  variables:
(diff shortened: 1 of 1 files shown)


zackgalbreath:
This looks right to me, but let's wait and see when the pipelines run. We should be able to see these new memory requests in the relevant job logs.

mvandenburgh:
@spackbot run pipeline

zackgalbreath (changes requested):
@mvandenburgh it looks like that worked! see here:
https://gitlab.spack.io/spack/spack/-/jobs/5037528
> "MemoryRequest" overwritten with "16G"

Would you please make the same changes for the relevant `-protected-generate` jobs too?

zackgalbreath (dismissed):

zackgalbreath:
@spackbot run pipeline

scottwittenburg:
Be aware that @haampie is doing something similar in #34784 but there he is getting the number of concretization processes to match with the cpu request.  He also did some profiling and found that about 1GB of memory per core is required for concretization (I'm not sure if that guideline is across the board or just for E4S), so the memory request is set accordingly in that PR as well.

zackgalbreath (dismissed):

zackgalbreath:
@spackbot run pipeline

zackgalbreath (approved):

scottwittenburg:
Since we have verified some of these recent jobs failing with "pod status is failed" are getting evicted due to memory pressure, let's merge this while we figure out if #34784 is a better approach or it it suffers from similar problem.

PR #34654, opened 2022-12-22
doxygen: use tarballs instead of git checkouts and add 1.9.5

lorddavidiii:
Downloading a source tarball is faster and there is no reason why the whole repo should be cloned I think. This also adds version 1.9.5 and changes the homepage url (I use that one, which is on there github about section, but it is in the end also only a redirection to https://www.doxygen.nl/).


sethrj (approved):
URL fetcher is good and checked the latest and a few other versions. Thanks!

PR #34659, opened 2022-12-22
Add MSBuild and NMake Builders

johnwparent:
Add NMake and MSBuild buildsystems to Spack packages and builders
This is functionality duplicated from #33564 in an effort to break down that large PR.


johnwparent:
@scheibelp As discussed I'm spinning off changes that can stand on their own from #33564 so we can maybe start getting the easier changes through.

lib/spack/spack/build_systems/nmake.py
```diff
@@ -30,73 +30,117 @@ class NMakePackage(spack.package_base.PackageBase):
...
 class NMakeBuilder(BaseBuilder):
     """The NMake builder encodes the most common way of building software with
-    NMake on Windows. It has three phases that can be overridden, if need be:
+    Mircosoft's MSBuild tool. It has three phases that can be overridden, if need be:
 
-            1. :py:meth:`~.NMakeBuilder.edit`
-            2. :py:meth:`~.NMakeBuilder.build`
-            3. :py:meth:`~.NMakeBuilder.install`
+            1. :py:meth:`~.NMakeBuilder.build`
+            2. :py:meth:`~.NMakeBuilder.install`
```
scheibelp, line 36: This mentions 3 phases but lists 2

lib/spack/spack/build_systems/nmake.py
```diff
@@ -30,73 +30,117 @@ class NMakePackage(spack.package_base.PackageBase):
 @spack.builder.builder("nmake")
 class NMakeBuilder(BaseBuilder):
     """The NMake builder encodes the most common way of building software with
-    NMake on Windows. It has three phases that can be overridden, if need be:
+    Mircosoft's MSBuild tool. It has three phases that can be overridden, if need be:
```
scheibelp: (question) One thing that confuses me: we have an `MSBuild` class as well (i.e. the other file added in this PR): is that a distinct concept from Microsoft's MSBuild tool?

lib/spack/spack/build_systems/nmake.py
```diff
@@ -30,73 +30,117 @@ class NMakePackage(spack.package_base.PackageBase):
...
+
+    def nmake_install_args(self):
+        """Define arguments appropriate only for install phase to NMake.
+        This is an empty list by default.
+        Individual packages should override to specify NMake args to command line"""
+        return []
 
     def build(self, pkg, spec, prefix):
-        """Run "make" on the build targets specified by the builder."""
+        """Run "msbuild" on the build targets specified by the builder."""
```
scheibelp: Should this way "run nmake" rather than "run msbuild"?

lib/spack/spack/build_systems/nmake.py
```diff
@@ -30,73 +30,117 @@ class NMakePackage(spack.package_base.PackageBase):
...
+            opts.append("/f {}".format(self.nmakefile_name))
         with fs.working_dir(self.build_directory):
-            inspect.getmodule(self.pkg).nmake(*self.build_targets)
+            inspect.getmodule(self.pkg).nmake(
+                *opts, *self.build_targets, ignore_quotes=self.ignore_quotes
+            )
 
     def install(self, pkg, spec, prefix):
-        """Run "make" on the install targets specified by the builder."""
+        """Run "msbuild" on the install targets specified by the builder.
```
scheibelp: Should this way "run nmake" rather than "run msbuild"?

lib/spack/spack/build_systems/nmake.py
```diff
@@ -30,73 +30,117 @@ class NMakePackage(spack.package_base.PackageBase):
...
-    build_time_test_callbacks = ["check"]
+    install_targets: List[str] = ["INSTALL"]
 
-    #: Callback names for install-time test
-    install_time_test_callbacks = ["installcheck"]
+    @property
+    def ignore_quotes(self):
+        """Returns boolean representing whether arguments to nmake
+        should have quotes ignored. Default False"""
+        return False
```
scheibelp, line 75: (question) what is the purpose of ignoring quotes?

Generally we encourage packages to specify options without quotes, and ignore_quotes is ~true~ (`False` by default) for `Executable`. In other words I suspect this shouldn't be exposed as a public method on this class (and may not need to exist as a method at all on the class).

scheibelp (changes requested):
I have a couple questions and a couple requests related to documentation

lib/spack/spack/build_systems/nmake.py
johnwparent: There are a number of cases where packages need custom quoting behavior to perform correctly, particularly in the case where there may be a path with a space specified on the command line. In those instances it's necessary to allow packages to specify the way they handle this, and I thought it would be cleaner, if that's the case, for a package to specify the arguments to nmake as they want and to then handle the quoting with that method rather than overriding the entire install/build methods to set that one flag. Perl is a good example of this.

scheibelp: > particularly in the case where there may be a path with a space specified on the command line.

Are there other cases? It might be preferable to register these arguments as paths and have the `Executable` object automatically except such arguments from examination on Windows. Basically: if there's a policy we can extract into the common-use objects I want to avoid users having to consider the same problem repeatedly.

johnwparent: Other cases are mostly related to handling issues with ill formed nmake or msbuild files that don't process CLI arguments very well and rely on there being quotes in the argument passed in for various reasons (I recall at least one package like this).  

The issue with abstracting this away from user handling is the use cases are fairly niche, inconsistent, and unpredictable. That being said, I don't really like handling this as a property either, maybe a class variable?

scheibelp: I think this should

* keep the builder property (no need to make it a class variable)
* rephrase the docstring for `ignore_quotes` to explain that this controls whether we warn about the presence of quotes, like `Executable` (by itself, I think the meaning of "ignoring quotes" is confusing: e.g. someone might take that to mean that we discard them)
* also mention in the docstring that builders for some packages may want to set `ignore_quotes=True` when the builder in fact needs to quote some arguments

Overall then build systems would set ignore_quotes=True (to mean that we shouldn't complain about the presence of quotes) if:

* Paths contain spaces (specifically paths passed as arguments to the builder exe)
* The build script parsing expects quotes

johnwparent: Sounds good, will amend now.

johnwparent: Amended and pushed.

lib/spack/spack/build_systems/msbuild.py
```diff
@@ -0,0 +1,118 @@
...
+    def msbuild_install_args(self):
+        """Define install arguments to MSBuild outside of the INSTALL target. This is the same
+        as `msbuild_args` by default."""
+        return self.msbuild_args()
+
+    def build(self, pkg, spec, prefix):
+        """Run "msbuild" on the build targets specified by the builder."""
+        with fs.working_dir(self.build_directory):
+            inspect.getmodule(self.pkg).msbuild(
+                *self.msbuild_args(), self.define_targets(*self.build_targets)
```
scheibelp: Would it also make sense to include `std_msbuild_args` here? Doing that would be analogous to how `cmake` treats `std_cmake_args` in its `cmake` phase.

lib/spack/spack/build_systems/nmake.py
```diff
@@ -30,73 +30,119 @@ class NMakePackage(spack.package_base.PackageBase):
 @spack.builder.builder("nmake")
 class NMakeBuilder(BaseBuilder):
     """The NMake builder encodes the most common way of building software with
-    NMake on Windows. It has three phases that can be overridden, if need be:
+    Mircosoft's MSBuild tool. It has two phases that can be overridden, if need be:
```
scheibelp: Both this and `MSBuilder` claim to encode the most common way of building software "with Microsoft's MSBuild tool": I presume this part was copy/pasted and could use an adjustment.

scheibelp (changes requested):
I have one question and a request to update the docstring of the builder classes

scheibelp (dismissed):

scheibelp (approved):

scheibelp:
@spackbot run pipeline

PR #34664, opened 2022-12-22
Add CUDA 12.0

wyphan:
SHA256 sums from @fspiga


wyphan:
@ax3l @Rombur does `lib/spack/spack/build_systems/cuda.py` need to be updated too?

wyphan:
@spackbot fix style

alalazo (dismissed):

haampie (dismissed):

haampie:
I've added the conflicts (clang 15+ still not supported, gcc supported up to 12)

haampie (dismissed):

haampie:
@ulrikeyang @osborn9 @balay i've added an upperbound on cuda for hypre

balay:
@haampie - many pkgs fail to build  with cuda-12 [from what I tried  - petsc, hypre, kokkos, magma, amgx]

Perhaps its beter to set 11.8 as default?

wyphan:
@balay I think Spack automatically picks the newest version, as no version has been marked as the preferred version in the `cuda` recipe.

balay:
> @balay I think Spack automatically picks the newest version, as no version has been marked as the preferred version in the cuda recipe.

Yeah - and I was suggesting - perhaps  version '11.8' should now be marked as 'preferred'.

But I don't know how one would do this in cuda/package.py - where the version()  is indirectly called here.. [and preferred=True is usually set via version()]

balay (approved):

haampie:
If those packages are failing they should set an upperbound. Setting a preferred=... either delays the problem or we end up with new versions of those packages with support for newer cuda, and are simply missing upperbounds on older versions cause they were never built. Both are equally bad.

PR #34668, opened 2022-12-22
Change regex in bacio patch to avoid python re bug

AlexanderRichert-NOAA:
Apparently in sufficiently recent versions of Python (I've found it in 3.8 and 3.10), there's what I can only assume is a bug where for the re module, ".*" matches twice for a given string. So `re.sub(".*","replacement","pattern")` yields "replacementreplacement". For the bacio patch, this causes the contents of VERSION to be set to "2.4.12.4.1". This can be avoided by using ".+" instead.


alalazo (approved):
This is not a bug in Python, but a bugfix, see https://stackoverflow.com/a/54714497/771663

PR #34672, opened 2022-12-23
faiss: fixed error when importing faiss python package

lpottier:
Missing directory in `LD_LIBRARY_PATH` resulting into the following error when doing `import faiss`:
```python
ImportError: libfaiss_python_callbacks.so: cannot open shared object file: No such file or directory
```

Signed-off-by: Loïc Pottier <48072795+lpottier@users.noreply.github.com>

### DIFF
--- a/var/spack/repos/builtin/packages/faiss/package.py
+++ b/var/spack/repos/builtin/packages/faiss/package.py
@@ -75,7 +75,7 @@ class Faiss(AutotoolsPackage, CMakePackage, CudaPackage):
     def setup_run_environment(self, env):
         if "+python" in self.spec:
             env.prepend_path("PYTHONPATH", python_platlib)
-            env.append_path("LD_LIBRARY_PATH", os.path.join(python_platlib, "faiss/"))
+            env.append_path("LD_LIBRARY_PATH", os.path.join(python_platlib, "faiss"))
 
 
 class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
--- a/var/spack/repos/builtin/packages/faiss/package.py
+++ b/var/spack/repos/builtin/packages/faiss/package.py
@@ -22,7 +22,7 @@ class Faiss(AutotoolsPackage, CMakePackage, CudaPackage):
     homepage = "https://github.com/facebookresearch/faiss"
     url = "https://github.com/facebookresearch/faiss/archive/v1.6.3.tar.gz"
 
-    maintainers = ["bhatiaharsh", "rblake-llnl"]
+    maintainers = ["bhatiaharsh", "rblake-llnl", "lpottier"]
 
     build_system(
         conditional("cmake", when="@1.7:"), conditional("autotools", when="@:1.6"), default="cmake"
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/faiss/package.py
```diff
@@ -75,6 +75,7 @@ class Faiss(AutotoolsPackage, CMakePackage, CudaPackage):
     def setup_run_environment(self, env):
         if "+python" in self.spec:
             env.prepend_path("PYTHONPATH", python_platlib)
+            env.append_path("LD_LIBRARY_PATH", os.path.join(python_platlib, "faiss/"))
```
haampie:
```suggestion
            env.append_path("LD_LIBRARY_PATH", os.path.join(python_platlib, "faiss"))
```

Notice that the relevant variable on darwin is `DYLD_FALLBACK_LIBRARY_PATH`

haampie (changes requested):
Hm, with this interaction between cmake and pip we end up with bad rpaths pointing to the stage dir (this is due to `CMAKE_INSTALL_RPATH_USE_LINK_PATH`):

```
 0x000000000000000f (RPATH)              Library rpath: [/tmp/harmen/spack-stage/spack-stage-faiss-1.7.2-a7ouyzvscepi75hjspfdvstcrnnwoh7v/spack-build-a7ouyzv/faiss:/tmp/harmen/spack-stage/spack-stage-faiss-1.7.2-a7ouyzvscepi75hjspfdvstcrnnwoh7v/spack-build-a7ouyzv/faiss/python:/home/harmen/spack/opt/spack/linux-ubuntu22.10-zen2/gcc-12.2.0/faiss-1.7.2-a7ouyzvscepi75hjspfdvstcrnnwoh7v/lib:/home/harmen/spack/opt/spack/linux-ubuntu22.10-zen2/gcc-12.2.0/faiss-1.7.2-a7ouyzvscepi75hjspfdvstcrnnwoh7v/lib64:/home/harmen/spack/opt/spack/linux-ubuntu22.10-zen2/gcc-12.2.0/openblas-0.3.21-sluybbuclqjdcmbf73xola4lddhjncrz/lib]
```

Is faiss building the python extension in a standard way?

Can you drop the use-link-path option in this PR? Should be `self.define("CMAKE_INSTALL_RPATH_USE_LINK_PATH", False)` in `cmake_args` of `faiss`, overriding Spack's defaults.

The whole problem is gone if you add an `$ORIGIN` rpath to `_swigfaiss.so`, but I must admit I don't immediately know how to do that in CMake s.t. it works both on linux ($ORIGIN) and macos (@loader_path).

haampie:
I did some research on macos (but didn't try to install faiss because my macbook air is too old and slow)

Shared libraries are created with install_name set to `@rpath/lib<name>.dylib` [since CMake 3.0](https://cmake.org/cmake/help/latest/policy/CMP0042.html#policy:CMP0042), so the linker always copies `@rpath/lib<name>.dylib` as a load command. That means that adding an rpath as `@loader_path` on macOS is enough:

```cmake
if(APPLE)
  set_target_properties(your_target PROPERTIES INSTALL_RPATH "@loader_path;${CMAKE_INSTALL_RPATH}")
elseif(UNIX)
  set_target_properties(your_target PROPERTIES INSTALL_RPATH "$ORIGIN;${CMAKE_INSTALL_RPATH}")
endif()
```

The `;` is there to inform cmake this is a list of strings; cmake will then concatenated them properly with `:` on linux (if you use `:` you get a trailing `:` if the variable is empty/not set)

---

Edit: ok. the problem is cmake's install isn't used to install the python part :( that explains why the rpaths refer to the build dir.

lpottier:
Hi @haampie, thanks for your review! 
I will do some research tonight about how we could improve this recipe without messing up too much `LD_LIBRARY_PATH`. Right now, it works on macOS and Linux.

haampie:
Well, it turns out to be a giant mess :p 

The build system is missing support for properly installing the Python part. And right now libraries are installed by Python instead of through CMake, and that causes issues since CMake patches binaries (their rpaths) upon install.

I'm not sure if there is a better solution. Making an issue upstream would be good.

haampie (approved):

lpottier:
Okay got it, thanks @haampie . I have opened an issue upstream: https://github.com/spack/spack/issues/34893.

PR #10944, opened 2019-03-20
Fix QT4 build for mac

sethrj:
This PR:

- [x] adds support for the "latest" QT 4,
- [x] fixes an error in the `package` script that manifests on mac+clang builds,
- [x] fixes the build for High Sierra,
- [x] allows more components of QT to be disabled to improve chance of success
- [x] adds "+static" variant so it can be (in theory) built statically

Fixes #12139

PR #34682, opened 2022-12-23
Qt: patch additional config file

adamjstewart:
Qt 4 includes config files that inject flags like `-mmacosx-version-min=10.5` into the build on macOS. These flags cause the compiler to behave in a backwards-compatible way, but that means that newer things like C++11 won't work. #10944 introduced a patch file to fix this.

However, this patch file was missing `mkspecs/common/g++-macx.conf`. This config file is used when building `py-pyqt4`. The additional section added to our patch file allows me to successfully build py-pyqt4 now.

Separated from #32696 to simplify that PR.

### DIFF
--- a/var/spack/repos/builtin/packages/qt/qt4-mac.patch
+++ b/var/spack/repos/builtin/packages/qt/qt4-mac.patch
@@ -1,8 +1,7 @@
-diff --git a/src/gui/painting/qpaintengine_mac.cpp b/src/gui/painting/qpaintengine_mac.cpp
-index 4aa0668..63b646d 100644
---- a/src/gui/painting/qpaintengine_mac.cpp
-+++ b/src/gui/painting/qpaintengine_mac.cpp
-@@ -340,13 +340,7 @@ CGColorSpaceRef QCoreGraphicsPaintEngine::macDisplayColorSpace(const QWidget *wi
+diff -Naur a/src/gui/painting/qpaintengine_mac.cpp b/src/gui/painting/qpaintengine_mac.cpp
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/qt/qt4-mac.patch
```diff
@@ -215,9 +214,23 @@ diff -r -u qt-everywhere-opensource-src-4.8.7.orig/mkspecs/common/g++-base.conf
...
+ 
+ QMAKE_LFLAGS_STATIC_LIB += -all_load
+ 
+-QMAKE_CFLAGS_X86_64 += -Xarch_x86_64 -mmacosx-version-min=10.5
+-QMAKE_CFLAGS_PPC_64 += -Xarch_ppc64 -mmacosx-version-min=10.5
++QMAKE_CFLAGS_X86_64 += -Xarch_x86_64 -mmacosx-version-min=$$QMAKE_MACOSX_DEPLOYMENT_TARGET
++QMAKE_CFLAGS_PPC_64 += -Xarch_ppc64 -mmacosx-version-min=$$QMAKE_MACOSX_DEPLOYMENT_TARGET
+ 
+ QMAKE_CXXFLAGS_X86_64         = $$QMAKE_CFLAGS_X86_64
+ QMAKE_CXXFLAGS_PPC_64         = $$QMAKE_CFLAGS_PPC_64
```
adamjstewart, line 230: This is the only real change in this PR. If you want, I can remove all the extraneous header changes too.

sethrj (approved):
Removing the spurious diff would've been good but it's fine.

PR #34687, opened 2022-12-23
style: Run mypy last

blue42u:
Run `mypy` last during `spack style` so the error line numbers are useful when `--fix` is passed.


haampie (approved):

haampie:
@spackbot run pipeline

Issue #34037, opened 2022-11-21
Installation issue: qt-base on macOS arm64


adamjstewart:
### Steps to reproduce the issue

```console
$ spack spec -I qt-base
Input spec
--------------------------------
 -   qt-base

Concretized
--------------------------------
 -   qt-base@6.3.2%apple-clang@14.0.0+accessibility~dbus+framework~gtk+gui~ipo+network~opengl+shared+sql+widgets build_system=cmake build_type=RelWithDebInfo arch=darwin-monterey-m1
[+]      ^at-spi2-core@2.40.1%apple-clang@14.0.0~strip build_system=meson buildtype=debugoptimized default_library=shared arch=darwin-monterey-m1
[+]          ^dbus@1.12.8%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1
[+]              ^docbook-xml@4.4%apple-clang@14.0.0 build_system=generic arch=darwin-monterey-m1
[+]              ^docbook-xsl@1.79.2%apple-clang@14.0.0 build_system=generic patches=a92c397 arch=darwin-monterey-m1
[+]              ^libsm@1.2.3%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
[+]                  ^libice@1.0.9%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
[+]          ^fixesproto@5.0%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
[+]              ^util-macros@1.19.3%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
[+]          ^gettext@0.21.1%apple-clang@14.0.0+bzip2+curses+git~libunistring+libxml2+tar+xz build_system=autotools arch=darwin-monterey-m1
[+]              ^tar@1.34%apple-clang@14.0.0 build_system=autotools zip=pigz arch=darwin-monterey-m1
[+]                  ^pigz@2.7%apple-clang@14.0.0 build_system=makefile arch=darwin-monterey-m1
[+]          ^glib@2.74.1%apple-clang@14.0.0~libmount build_system=generic tracing=none arch=darwin-monterey-m1
[+]              ^libelf@0.8.13%apple-clang@14.0.0 build_system=autotools arch=darwin-monterey-m1
[+]          ^gobject-introspection@1.72.0%apple-clang@14.0.0~strip build_system=meson buildtype=debugoptimized default_library=shared arch=darwin-monterey-m1
[...]

PR #34688, opened 2022-12-23
qt-base: ~network by default

adamjstewart:
`qt-base+network` doesn't build reliably: #34037. Until it does, we should disable it by default.

Extracted out of #32696 to make that PR easier to review.


wdconinc (approved):
This change sounds good to me.

sethrj (approved):

PR #34692, opened 2022-12-24
py-pytest-datadir: Init at 1.4.1

HaoZeke:
As it says on the tin.

### DIFF
--- a/var/spack/repos/builtin/packages/py-pytest-datadir/package.py
+++ b/var/spack/repos/builtin/packages/py-pytest-datadir/package.py
@@ -17,4 +17,5 @@ class PyPytestDatadir(PythonPackage):
 
     depends_on("python@3.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools-scm", type="build")
     depends_on("py-pytest@5.0:", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-pytest-datadir/package.py
```diff
@@ -0,0 +1,20 @@
...
+    """Pytest plugin for manipulating test data directories and files."""
+
+    homepage = "https://github.com/gabrielcnr/pytest-datadir"
+    pypi = "pytest-datadir/pytest-datadir-1.4.1.tar.gz"
+    maintainers = ["HaoZeke"]
+
+    version("1.4.1", sha256="9f7a3c4def6ac4cac3cc8181139ab53bd2667231052bd40cb07081748d4420f0")
+
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
```
adamjstewart, line 19: Missing build dep on py-setuptools-scm

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pytest-datadir/package.py
HaoZeke: Excellent catch, added.

adamjstewart (approved):

PR #24639, opened 2021-06-30
Adding ability to compare git references to spack install

vsoch:
This will allow a user to (from anywhere a Spec is parsed including both name and version) refer to a git commit in lieu of 
a package version, and be able to make comparisons with releases in the history based on commits (or with other commits). We do this by way of:

 - Adding a property, is_commit, to a version, meaning I can always check if a version is a commit and then change some action.
 - Adding an attribute to the Version object which can lookup commits from a git repo and find the last known version before that commit, and the distance
 - Construct new Version comparators, which are tuples. For normal versions, they are unchanged. For commits with a previous version x.y.z, d commits away, the comparator is (x, y, z, '', d). For commits with no previous version, the comparator is ('', d) where d is the distance from the first commit in the repo.
 - Metadata on git commits is cached in the misc_cache, for quick lookup later.
 - Git repos are cached as bare repos in `~/.spack/git_repos`
 - In both caches, git repo urls are turned into file paths within the cache

If a commit cannot be found in the cached git repo, we fetch from the repo. If a commit is found in the cached metadata, we do not recompare to newly downloaded tags (assuming repo structure does not change). The cached metadata may be thrown out by using the `spack clean -m` option if you know the repo structure has changed in a way that invalidates existing entries. Future work will include automatic updates.

# Finding previous versions
Spack will search the repo for any tags that match the string of a version given by the `version` directive. Spack will also search for any tags that match `v + string` for any version string. Beyond that, Spack will search for tags that match a SEMVER regex (i.e., tags of the form x.y.z) and interpret those tags as valid versions as well. Future work will increase the breadth of tags understood by Spack

For each tag, Spack queries git to determine whether the tag is an ancestor of the commit in question or not. Spack then sorts the tags that are ancestors of the commit by commit-distance in the repo, and takes the nearest ancestor. The version represented by that tag is listed as the previous version for the commit.

Not all commits will find a previous version, depending on the package workflow. Future work may enable more tangential relationships between commits and versions to be discovered, but many commits in real world git repos require human knowledge to associate with a most recent previous version. Future work will also allow packages to specify commit/tag/version relationships manually for such situations.

# Version comparisons.
[...]

PR #34696, opened 2022-12-26
Windows Install tests

markus-ferrell:
This PR changes the install tests to to make them windows compatible

### DIFF
--- a/lib/spack/spack/cmd/unit_test.py
+++ b/lib/spack/spack/cmd/unit_test.py
@@ -209,12 +209,11 @@ def unit_test(parser, args, unknown_args):
     # mock configuration used by unit tests
     # Note: skip on windows here because for the moment,
     # clingo is wholly unsupported from bootstrap
-    if sys.platform != "win32":
-        with spack.bootstrap.ensure_bootstrap_configuration():
-            spack.bootstrap.ensure_core_dependencies()
-            if pytest is None:
-                spack.bootstrap.ensure_environment_dependencies()
-                import pytest
+    with spack.bootstrap.ensure_bootstrap_configuration():
+        spack.bootstrap.ensure_core_dependencies()
+        if pytest is None:
+            spack.bootstrap.ensure_environment_dependencies()
+            import pytest
 
     if args.pytest_help:
         # make the pytest.main help output more accurate
--- a/lib/spack/spack/directory_layout.py
+++ b/lib/spack/spack/directory_layout.py
@@ -10,8 +10,8 @@ import posixpath
 import re
 import shutil
 import sys
-from pathlib import Path
 from contextlib import contextmanager
+from pathlib import Path
 
 import llnl.util.filesystem as fs
 import llnl.util.tty as tty
--- a/lib/spack/spack/directory_layout.py
+++ b/lib/spack/spack/directory_layout.py
@@ -10,6 +10,7 @@ import posixpath
 import re
 import shutil
 import sys
+import pathlib
 from contextlib import contextmanager
 
 import llnl.util.filesystem as fs
@@ -334,7 +335,7 @@ class DirectoryLayout:
 
         path = self.relative_path_for_spec(spec)
         assert not path.startswith(self.root)
-        return os.path.join(self.root, path)
+        return str(pathlib.Path(self.root) / pathlib.Path(path))
 
     def remove_install_directory(self, spec, deprecated=False):
         """Removes a prefix and any empty parent directories from the root.
--- a/lib/spack/spack/directory_layout.py
+++ b/lib/spack/spack/directory_layout.py
@@ -10,7 +10,7 @@ import posixpath
 import re
 import shutil
 import sys
-import pathlib
+from pathlib import Path
 from contextlib import contextmanager
 
 import llnl.util.filesystem as fs
@@ -105,7 +105,7 @@ class DirectoryLayout:
 
         projection = spack.projections.get_projection(self.projections, spec)
         path = spec.format(projection)
-        return path
+        return str(Path(path))
 
     def write_spec(self, spec, path):
         """Write a spec out to a file."""
@@ -335,7 +335,7 @@ class DirectoryLayout:
 
         path = self.relative_path_for_spec(spec)
         assert not path.startswith(self.root)
-        return str(pathlib.Path(self.root) / pathlib.Path(path))
+        return os.path.join(self.root, path)
 
     def remove_install_directory(self, spec, deprecated=False):
         """Removes a prefix and any empty parent directories from the root.
--- a/lib/spack/spack/test/cmd/install.py
+++ b/lib/spack/spack/test/cmd/install.py
@@ -1101,7 +1101,7 @@ def test_install_empty_env(
 
 @pytest.mark.skipif(
     sys.platform == "win32",
-    reason="Windows log_output logs phase header out of order",
+    reason="Windows logger I/O operation on closed file when install fails",
 )
 @pytest.mark.disable_clean_stage_check
 @pytest.mark.parametrize(
@@ -1206,7 +1206,7 @@ def test_install_use_buildcache(
 
 @pytest.mark.skipif(
     sys.platform == "win32",
-    reason="Windows log_output logs phase header out of order",
+    reason="Windows logger I/O operation on closed file when install fails",
 )
 @pytest.mark.regression("34006")
 @pytest.mark.disable_clean_stage_check
--- a/lib/spack/spack/test/cmd/install.py
+++ b/lib/spack/spack/test/cmd/install.py
@@ -548,8 +548,7 @@ def test_cdash_report_concretization_error(
 
 
 @pytest.mark.skipif(
-    sys.platform == "win32",
-    reason="Windows log_output logs phase header out of order",
+    sys.platform == "win32", reason="Windows log_output logs phase header out of order"
 )
 @pytest.mark.disable_clean_stage_check
 def test_cdash_upload_build_error(tmpdir, mock_fetch, install_mockery, capfd):
@@ -1100,8 +1099,7 @@ def test_install_empty_env(
 
 
 @pytest.mark.skipif(
-    sys.platform == "win32",
-    reason="Windows log_output logs phase header out of order",
+    sys.platform == "win32", reason="Windows log_output logs phase header out of order"
 )
 @pytest.mark.disable_clean_stage_check
 @pytest.mark.parametrize(
@@ -1205,8 +1203,7 @@ def test_install_use_buildcache(
 
 
 @pytest.mark.skipif(
-    sys.platform == "win32",
-    reason="Windows log_output logs phase header out of order",
+    sys.platform == "win32", reason="Windows log_output logs phase header out of order"
 )
 @pytest.mark.regression("34006")
 @pytest.mark.disable_clean_stage_check
--- a/lib/spack/spack/test/cmd/install.py
+++ b/lib/spack/spack/test/cmd/install.py
@@ -549,7 +549,7 @@ def test_cdash_report_concretization_error(
 
 @pytest.mark.skipif(
     sys.platform == "win32",
-    reason="log_ouput cannot currently be used outside of subprocess on Windows",
+    reason="log_output cannot currently be used outside of subprocess on Windows",
 )
 @pytest.mark.disable_clean_stage_check
 def test_cdash_upload_build_error(tmpdir, mock_fetch, install_mockery, capfd):
--- a/lib/spack/spack/test/cmd/install.py
+++ b/lib/spack/spack/test/cmd/install.py
@@ -919,7 +919,7 @@ def test_cdash_auth_token(tmpdir, mock_fetch, install_mockery, capfd):
             assert "Using CDash auth token from environment" in out
 
 
-@pytest.mark.skipif(sys.platform == "win32", reason="Logger race condition in windows")
+@pytest.mark.skipif(sys.platform == "win32", reason="Windows log_output logs phase header out of order")
 @pytest.mark.disable_clean_stage_check
 def test_cdash_configure_warning(tmpdir, mock_fetch, install_mockery, capfd):
     # capfd interferes with Spack's capturing of e.g., Build.xml output
--- a/lib/spack/spack/test/cmd/install.py
+++ b/lib/spack/spack/test/cmd/install.py
@@ -7,6 +7,7 @@ import builtins
 import filecmp
 import itertools
 import os
+import pathlib
 import re
 import sys
 import time
@@ -37,8 +38,6 @@ mirror = SpackCommand("mirror")
 uninstall = SpackCommand("uninstall")
 buildcache = SpackCommand("buildcache")
 find = SpackCommand("find")
-
-pytestmark = pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
 
 
 @pytest.fixture()
@@ -204,7 +203,7 @@ def test_show_log_on_error(
     assert isinstance(install.error, spack.build_environment.ChildError)
     assert install.error.pkg.name == "build-error"
 
-    assert "==> Installing build-error" in out
+    assert "Installing build-error" in out
     assert "See build log for details:" in out
 
 
@@ -263,9 +262,9 @@ def test_install_commit(mock_git_version_info, install_mockery, mock_packages, m
 
     """
     repo_path, filename, commits = mock_git_version_info
-    monkeypatch.setattr(
-        spack.package_base.PackageBase, "git", "file://%s" % repo_path, raising=False
-    )
+    file_url = pathlib.Path(repo_path).as_uri()
+
+    monkeypatch.setattr(spack.package_base.PackageBase, "git", file_url, raising=False)
 
     # Use the earliest commit in the respository
     spec = Spec(f"git-test-commit@{commits[-1]}").concretized()
@@ -548,6 +547,10 @@ def test_cdash_report_concretization_error(
             assert any(x in content for x in expected_messages)
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="log_ouput cannot currently be used outside of subprocess on Windows",
+)
 @pytest.mark.disable_clean_stage_check
 def test_cdash_upload_build_error(tmpdir, mock_fetch, install_mockery, capfd):
     # capfd interferes with Spack's capturing
@@ -747,6 +750,10 @@ def test_install_deps_then_package(tmpdir, mock_fetch, install_mockery):
     assert os.path.exists(root.prefix)
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="Environment views not supported on windows. Revisit after #34701",
+)
 @pytest.mark.regression("12002")
 def test_install_only_dependencies_in_env(
     tmpdir, mock_fetch, install_mockery, mutable_mock_env_path
@@ -896,7 +903,7 @@ def test_install_help_does_not_show_cdash_options(capsys):
         assert "CDash URL" not in captured.out
 
 
-def test_install_help_cdash(capsys):
+def test_install_help_cdash():
     """Make sure `spack install --help-cdash` describes CDash arguments"""
     install_cmd = SpackCommand("install")
     out = install_cmd("--help-cdash")
@@ -913,6 +920,7 @@ def test_cdash_auth_token(tmpdir, mock_fetch, install_mockery, capfd):
             assert "Using CDash auth token from environment" in out
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Logger race condition in windows")
 @pytest.mark.disable_clean_stage_check
 def test_cdash_configure_warning(tmpdir, mock_fetch, install_mockery, capfd):
     # capfd interferes with Spack's capturing of e.g., Build.xml output
@@ -938,6 +946,10 @@ def test_cdash_configure_warning(tmpdir, mock_fetch, install_mockery, capfd):
             assert "foo: No such file or directory" in content
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="ArchSpec gives test platform debian rather than windows",
+)
 def test_compiler_bootstrap(
     install_mockery_mutable_config,
     mock_packages,
@@ -954,6 +966,7 @@ def test_compiler_bootstrap(
     install("a%gcc@=12.0")
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Binary mirrors not supported on windows")
 def test_compiler_bootstrap_from_binary_mirror(
     install_mockery_mutable_config,
     mock_packages,
@@ -994,6 +1007,10 @@ def test_compiler_bootstrap_from_binary_mirror(
     install("--no-cache", "--only", "package", "b%gcc@10.2.0")
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="ArchSpec gives test platform debian rather than windows",
+)
 @pytest.mark.regression("16221")
 def test_compiler_bootstrap_already_installed(
     install_mockery_mutable_config,
@@ -1037,6 +1054,10 @@ def test_install_fails_no_args_suggests_env_activation(tmpdir):
     assert "using the `spack.yaml` in this directory" in output
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="Environment views not supported on windows. Revisit after #34701",
+)
 def test_install_env_with_tests_all(
     tmpdir, mock_packages, mock_fetch, install_mockery, mutable_mock_env_path
 ):
@@ -1048,6 +1069,10 @@ def test_install_env_with_tests_all(
         assert os.path.exists(test_dep.prefix)
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="Environment views not supported on windows. Revisit after #34701",
+)
 def test_install_env_with_tests_root(
     tmpdir, mock_packages, mock_fetch, install_mockery, mutable_mock_env_path
 ):
@@ -1059,6 +1084,10 @@ def test_install_env_with_tests_root(
         assert not os.path.exists(test_dep.prefix)
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="Environment views not supported on windows. Revisit after #34701",
+)
 def test_install_empty_env(
     tmpdir, mock_packages, mock_fetch, install_mockery, mutable_mock_env_path
 ):
@@ -1072,6 +1101,10 @@ def test_install_empty_env(
     assert "no specs to install" in out
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="log_ouput cannot currently be used outside of subprocess on Windows",
+)
 @pytest.mark.disable_clean_stage_check
 @pytest.mark.parametrize(
     "name,method",
@@ -1095,6 +1128,7 @@ def test_installation_fail_tests(install_mockery, mock_fetch, name, method):
     assert "See test log for details" in output
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Buildcache not supported on windows")
 def test_install_use_buildcache(
     capsys,
     mock_packages,
@@ -1172,6 +1206,10 @@ def test_install_use_buildcache(
             install_use_buildcache(opt)
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="log_ouput cannot currently be used outside of subprocess on Windows",
+)
 @pytest.mark.regression("34006")
 @pytest.mark.disable_clean_stage_check
 def test_padded_install_runtests_root(install_mockery_mutable_config, mock_fetch):
--- a/lib/spack/spack/test/cmd/install.py
+++ b/lib/spack/spack/test/cmd/install.py
@@ -947,8 +947,7 @@ def test_cdash_configure_warning(tmpdir, mock_fetch, install_mockery, capfd):
 
 
 @pytest.mark.skipif(
-    sys.platform == "win32",
-    reason="ArchSpec gives test platform debian rather than windows",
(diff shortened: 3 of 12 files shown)


markus-ferrell:
@spackbot fix style

markus-ferrell:
@spackbot fix style

markus-ferrell:
@spackbot fix style

lib/spack/spack/installer.py
```diff
@@ -2034,6 +2034,9 @@ def _real_install(self):
 
                         # Catch any errors to report to logging
                         self.timer.start(phase_fn.name)
+                        import time
+
+                        time.sleep(1)
```
markus-ferrell: This is to prevent a race condition in the logger for windows. This should be removed and replaced with a logger refactor.

lib/spack/llnl/util/filesystem.py
```diff
@@ -85,6 +85,13 @@
 ]
 
 
+def remove_file_mount(path):
```
markus-ferrell: I don't like the name of this function since it does more than it sounds like. The function changes C:\\.....\\Temp\\...\\repo.git into file\\Temp\\...\\repo.git

markus-ferrell: It removes everything before Temp due to long path issues with git clone

lib/spack/spack/installer.py
johnwparent: What's the error here again? I know we went through this a little while ago, but I'd like to record that conversation here. Also can we sleep for less time?

lib/spack/spack/test/cmd/install.py
```diff
@@ -1010,13 +1024,18 @@ def test_compiler_bootstrap_from_binary_mirror(
...
+    reason="ArchSpec gives test platform debian rather than linux for windows",
+)
 @pytest.mark.regression("16221")
 def test_compiler_bootstrap_already_installed(
     install_mockery_mutable_config,
     mock_packages,
     mock_fetch,
     mock_archive,
     mutable_config,
+    archspec_host_is_spack_test_host,
```
johnwparent: If we're skipping we probably don't need the fixture.

var/spack/repos/builtin.mock/packages/build-warnings/package.py
```diff
@@ -15,16 +17,33 @@ class BuildWarnings(Package):
...
-echo 'checking for suffix of executables...'
-echo 'foo.c:89: warning: some weird warning!'
-exit 1
-"""
-            )
-        configure()
+        if sys.platform == "win32":
+            with open("configure.bat", "w") as f:
+                f.write(
+                    """
```
johnwparent, line 23: Do we need to turn echo off here? I'm surprised that the parsing works with the duplicated output.

lib/spack/spack/version.py
```diff
@@ -1284,7 +1284,7 @@ def repository_uri(self):
             return os.path.join(*components)
         except ValueError:
             # If it's not a git url, it's a local path
-            return os.path.abspath(self.pkg.git)
+            return llnl.util.filesystem.remove_file_mount(os.path.abspath(self.pkg.git))
```
johnwparent: `repository_uri` should return a valid uri, and IIRC file uri have to be absolute paths. Additionally are you intentionally dropping the `://` component of the file url in that method?

lib/spack/spack/util/url.py
```diff
@@ -258,56 +258,3 @@ def _join(base_url, path, *extra, **kwargs):
...
-git_re = (
-    r"^(?:([a-z]+)://)?"  # 1. optional scheme
-    r"(?:([^@]+)@)?"  # 2. optional user
-    r"([^:/~]+)?"  # 3. optional hostname
-    r"(?(1)(?::([^:/]+))?|:)"  # 4. :<optional port> if scheme else :
-    r"(.*[^/])/?$"  # 5. path
-)
-
-
-def parse_git_url(url):
```
markus-ferrell, line 255: This function is very similar to urllib.parse.urlparse(). Using urlparse instead lowers maintenance efforts here especially as we would have to maintain a windows specific regex.

>>> import urllib
>>>p = file:///C:/path/to/repo.git
>>> urllib.parse.urlparse(p)
ParseResult(scheme='file', netloc='', path='/C:/path/to/repo.git', params='', query='', fragment='')

>>> g = "https://host.com:443/path/to/repo.git"
>>> urllib.parse.urlparse(g)
ParseResult(scheme='https', netloc='host.com:443', path='/path/to/repo.git', params='', query='', fragment='')

johnwparent:
@spackbot fix style

lib/spack/spack/installer.py
markus-ferrell: The error was when running `test_cdash_upload_build_error`. This test is skipped in this PR but this change was more here for the discussion.

 The installer would output the line [Executing phase install](https://github.com/spack/spack/blob/42a452d54c8a25f9f415fef8cf9e3a5c64b7a46a/lib/spack/spack/installer.py#L2031) after the actual install output. This makes cdash's build output reader fail because the build output is out of order.

lib/spack/spack/version.py
markus-ferrell: Repository_uri never returned a valid uri. It had a bit of a confusing name and should likely be renamed. It would return all of the pieces of the repository uri concatenated together to form a path.

var/spack/repos/builtin.mock/packages/build-warnings/package.py
markus-ferrell: I think we probably should just to be as clear as possible. It has echo off now. The tests don't seem to mind either way though.

johnwparent: Huh... well, either way thanks for making that change!

lib/spack/spack/build_environment.py
```diff
@@ -601,7 +601,10 @@ def set_module_variables_for_package(pkg):
 
     # Find the configure script in the archive path
     # Don't use which for this; we want to find it in the current dir.
-    m.configure = Executable("./configure")
+    if sys.platform == "win32":
+        m.configure = Executable("configure.bat")
```
johnwparent: I don't love this (I assume test only change) bleeding into the larger codebase. Is there any use case for this outside of the tests? We're in general trying to move away from defining these tools globally here, so ideally if it is necessary outside of the tests we'd find another place for it. Otherwise, can we make this a monkeypatch operation from a session scoped pytest fixture?

lib/spack/spack/cmd/unit_test.py
```diff
@@ -211,12 +211,11 @@ def unit_test(parser, args, unknown_args):
     # mock configuration used by unit tests
     # Note: skip on windows here because for the moment,
     # clingo is wholly unsupported from bootstrap
-    if sys.platform != "win32":
-        with spack.bootstrap.ensure_bootstrap_configuration():
-            spack.bootstrap.ensure_core_dependencies()
-            if pytest is None:
-                spack.bootstrap.ensure_environment_dependencies()
-                import pytest
+    with spack.bootstrap.ensure_bootstrap_configuration():
```
johnwparent, line 212: Great change, thanks for making it!

lib/spack/spack/test/cmd/install.py
```diff
@@ -741,6 +746,7 @@ def test_install_deps_then_package(tmpdir, mock_fetch, install_mockery):
     assert os.path.exists(root.prefix)
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
```
johnwparent: I know a lot of our tests are skipped for the "does not run on Windows" reason, but if we're adding new skips, can we add why? If it's a lot of work to determine why, then this is fine (which is why there are so many of the vague reason string), but ideally we'd be able to offer more of an explanation so we can know when we can turn the tests back on.

lib/spack/spack/test/cmd/install.py
```diff
@@ -906,6 +913,7 @@ def test_cdash_auth_token(tmpdir, mock_fetch, install_mockery, capfd):
             assert "Using CDash auth token from environment" in out
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
```
johnwparent: Same comment as above. In general, apply these comments to all skips where there is a vague reason string.

lib/spack/spack/test/cmd/install.py
```diff
@@ -931,6 +939,10 @@ def test_cdash_configure_warning(tmpdir, mock_fetch, install_mockery, capfd):
             assert "foo: No such file or directory" in content
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="ArchSpec gives test platform debian rather than linux for windows",
```
johnwparent: I think the test platform is supposed to select debian (at least it does on my system). I'm not saying we shouldn't skip here, but I think you may have the reasoning backwards. Whatever is looking for `platform=linux` is in error AFAICT.

lib/spack/spack/util/url.py
johnwparent: I'm on board with this change (IMO we should have as little custom url parsing as possible) but we'll need buyin from @scheibelp here as he has a better understanding of the large potential implications of this change.

lib/spack/spack/version.py
```diff
@@ -1250,10 +1253,7 @@ def __init__(self, pkg_name):
     def cache_key(self):
         if not self._cache_key:
             key_base = "git_metadata"
-            if not self.repository_uri.startswith("/"):
-                key_base += "/"
-            self._cache_key = key_base + self.repository_uri
-
+            self._cache_key = Path(key_base).joinpath(self.repository_uri)
```
johnwparent: When using pathlib be sure that none of the `pathlib.Path` objects bleed out from their local useage as much of Spack/the `os` module do not handle those objects well. It looks like you might be returning a `Path` object here.

johnwparent: If the path you're constructing does not exist on the filesystem you may also want to consider using `PurePath`

lib/spack/spack/version.py
```diff
@@ -1284,18 +1284,15 @@ def fetcher(self):
...
-        """
-        try:
-            components = [
-                str(c).lstrip("/") for c in spack.util.url.parse_git_url(self.pkg.git) if c
-            ]
-            return os.path.join(*components)
-        except ValueError:
-            # If it's not a git url, it's a local path
-            return os.path.abspath(self.pkg.git)
+            return Path("file").joinpath(Path(path).relative_to(Path.home()))
```
johnwparent: The `as_uri` method may be a little more wieldy here. Wouldn't this create something like `file\\<path>` relative to the users home directory? As opposed to `file:///<path>` It's also not immediately obvious why this is relative to `%USERPROFILE%`, can you clarify in a comment? Also just clarify in general as some local testing suggests this might be fragile.

lib/spack/spack/version.py
```diff
@@ -1330,9 +1327,10 @@ def lookup_ref(self, ref):
         known version prior to the commit, as well as the distance from that version
         to the commit in the git repo. Those values are used to compare Version objects.
         """
-        dest = os.path.join(spack.paths.user_repos_cache_path, self.repository_uri)
-        if dest.endswith(".git"):
-            dest = dest[:-4]
+        dest = Path(spack.paths.user_repos_cache_path).joinpath(self.repository_uri)
+
+        dest = dest.with_suffix("")
```
johnwparent: Can we use the idiomatic `/` to join paths?

johnwparent: Whoops commented this on the wrong line, see line two above.

johnwparent: W.R.T. to this line, is removing the file extension _always_ what we want to do here? It looks like the previous behavior was only to conditionally remove the ext when it was `.git`.

lib/spack/spack/version.py
```diff
@@ -1330,9 +1327,10 @@ def lookup_ref(self, ref):
...
+        dest = dest.as_posix()
```
johnwparent: This also appears to be a deviation in behavior. Not necessarily a problem, but I'd like us to be sure this is fine.

lib/spack/spack/version.py
markus-ferrell: This is bc the `repository_uri` function as it was written before is poorly named. It used to return the pieces of a file uri concatenated together to form a path. An example would be `file/AppData/Local/Temp/pytest-of-markus.ferrell/pytest-634/test_install_commit0/git_repo` instead of the real file uri of `file:///C:/Users/markus.ferrell/AppData/Local/Temp/pytest-of-markus.ferrell/pytest-634/test_install_commit0/git_repo`. They would use that path segment to concatenate onto another path to find the real file uri.

The relative to home was to try and resolve long path issues by removing part of the repeated path. I agree that it's too fragile though. It seems to be causing failures for other platforms too.

markus-ferrell: This likely needs a larger refactor to make the naming a bit more intuitive

markus-ferrell:
@spackbot fix style

markus-ferrell:
@spackbot fix style

lib/spack/spack/build_environment.py
johnwparent: Ping.

lib/spack/spack/installer.py
```diff
@@ -1555,7 +1555,7 @@ def _install_action(self, task):
             return InstallAction.INSTALL
 
         # Ensure install_tree projections have not changed.
-        assert task.pkg.prefix == rec.path
+        assert os.path.samefile(task.pkg.prefix, rec.path)
```
johnwparent: This is changing a path comparison to be about lower level filesystem details about each file rather than just path equivalency. I assume that this is OK, and at the very least the paths are still being compared so we're fine in that regard, but I just want to double check this is the right move as opposed to using something like `pathlib`.

lib/spack/spack/test/cmd/install.py
```diff
@@ -260,9 +259,9 @@ def test_install_commit(mock_git_version_info, install_mockery, mock_packages, m
 
     """
     repo_path, filename, commits = mock_git_version_info
-    monkeypatch.setattr(
-        spack.package_base.PackageBase, "git", "file://%s" % repo_path, raising=False
-    )
+    file_url = pathlib.Path(repo_path).as_uri()
+
+    monkeypatch.setattr(spack.package_base.PackageBase, "git", file_url, raising=False)
```
johnwparent, line 267: Nice!

lib/spack/spack/test/cmd/install.py
```diff
@@ -741,6 +744,7 @@ def test_install_deps_then_package(tmpdir, mock_fetch, install_mockery):
     assert os.path.exists(root.prefix)
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Environments not supported on windows")
```
johnwparent: Same for other lines like this... envs _are_ supported on Windows post Dan's PR landing, at least AFAIK, however views are NOT currently supported. If this is failing because of views, can we update the reason string, and if it's failing just on envs inherently, can you send me the error message?

lib/spack/spack/test/cmd/install.py
johnwparent: Ping

johnwparent (commented):
Some minor comments and pinging on some old ones. In general looks great, loving the pathlib changes.

lib/spack/spack/test/cmd/install.py
markus-ferrell: Good catch, most of them are failing rn because of the views. As far as I can tell, the symlink changes should make these pass. It's a bunch of the same error which I had resolved for the Environment tests PR.

johnwparent: Ok great, lets maybe even call out that PR in the reason message? It'll be helpful to grep and re-enable things

lib/spack/spack/build_environment.py
markus-ferrell: Changed up the test package that depended on this do an `Execute` call rather than a `configure()` so I could take that line of code out.

lib/spack/spack/version.py
markus-ferrell: We now return the posix path so we don't leak the Pathlib object

lib/spack/spack/build_environment.py
johnwparent: Sounds good! Thanks

lib/spack/spack/version.py
markus-ferrell: I switched it back to removing only on `.git`. My assumption was that everything coming in would have a `.git` extention but I don't think that's actually the case.

lib/spack/spack/test/cmd/install.py
markus-ferrell: I'm pretty sure I was mistaken here when writing the comment. Archspec should be returning windows and it is returning the default debian which is the problem. I amended the comment.

markus-ferrell:
@spackbot fix style

lib/spack/spack/test/cmd/install.py
markus-ferrell: Makes sense. I added the PR No. in the message

lib/spack/spack/installer.py
markus-ferrell: I dug into this a bit more than I did the first time. It looks like os.path.samefile changes the path strings around if they're a bit invalid. That's why the simple path comparison didn't work but samefile did. I added the bugfix to the underlying path problem and reverted this back to the path comparison they had before.

markus-ferrell:
@spackbot fix style

lib/spack/spack/test/cmd/install.py
```diff
@@ -1063,6 +1090,10 @@ def test_install_empty_env(
     assert "no specs to install" in out
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="log_ouput cannot currently be used outside of subprocess on Windows",
```
scheibelp: (minor typo) log_ouput -> log_output

(this message is repeated in a few places)

Also I don't fully follow: 

* this test doesn't create a log_output object itself
* any test using `SpackCommand` on Windows will end up creating a log_output object. There are at least a few `test/cmd` modules running on windows (e.g. `list`, although at least one of those tests has special considerations for output on Windows).

lib/spack/spack/util/url.py
scheibelp: IMO if you reinstate the test, and it passes with the stock `urllib.parse.urlparse()`, then it would make sense to remove this (and the test - I don't think either is needed if the provided function works, but I'm curious if the test *does* pass). I see this was introduced in https://github.com/spack/spack/pull/24639 but didn't see any specific comments about why it might have been added at the time (I think `urllib.parse.urlparse` existed at the time of that PR as well).

lib/spack/spack/directory_layout.py
```diff
@@ -334,7 +335,7 @@ def path_for_spec(self, spec):
 
         path = self.relative_path_for_spec(spec)
         assert not path.startswith(self.root)
-        return os.path.join(self.root, path)
+        return str(pathlib.Path(self.root) / pathlib.Path(path))
```
scheibelp: (question) what does this improve over os.path.join? It's not clear to me in this case.

lib/spack/spack/version.py
```diff
@@ -1226,14 +1225,13 @@ def fetcher(self):
...
-                str(c).lstrip("/") for c in spack.util.url.parse_git_url(self.pkg.git) if c
-            ]
-            return os.path.join(*components)
-        except ValueError:
-            # If it's not a git url, it's a local path
-            return os.path.abspath(self.pkg.git)
+        (scheme, netloc, path, params, query, fragment) = urllib.parse.urlparse(self.pkg.git)
+        path = path.lstrip("/")
+        if scheme == "file":
+            return Path("file") / Path(path).name
```
scheibelp: I'm not clear why `Path(...).name` is used, since that removes all the path components except for the last:

```
>>> import pathlib
>>> pathlib.Path('x/y/z').name
'z'
```

That seems like different behavior than before (and also wrong)

scheibelp (changes requested):
I have a couple questions and a couple requests.

lib/spack/spack/version.py
markus-ferrell: The previous implementation returned the full path to the local git repo. It is later used to concatenate onto a path. This works for linux because the absolute paths start with `/`. This would result in something later of `/my_new_location/file/local/git/repo`. Because windows uses different volume mounts, you end up getting something along the lines of `C:\\my_new_location\\file\\C:local\\git\\repo`. If we simply strip the file mount and use the rest of the path to match linux, we run into windows path length issues when calling git commands on the new path. I had spend a good amount of time on this section and hadn't found a better solution than using `.name` to remove the file mount and enough of the path to make the git commands work. It does change functionality from before though. @scheibelp Do you have any thoughts on how to resolve this problem?

lib/spack/spack/util/url.py
markus-ferrell: The stock `urllib.parse.urlparse()` does not pass those tests. There's a bit of massaging of data after that would need to happen to make those tests pass. I'm not sure that we need that exact parsing structure to fill out the `repository_uri` function. This is the only place where it's used.

markus-ferrell: For reference, this is what you'd have to do to make it pass almost all of the test cases:
```python
def parse_git_url(url):
    parsed = urllib.parse.urlparse(url)

    scheme = parsed.scheme if parsed.scheme else None
    username, hostname, port = parsed.username, parsed.hostname, parsed.port

    path = parsed.path.rstrip("/")
    if path.startswith("/~"):
        path = path.lstrip("/")

    if "@" in path:
        username, path = path.split("@")
    if ":" in path:
        hostname, path = path.split(":")

    if not (scheme or username or hostname or port) and path:
        raise ValueError("Url was actually a path")

    if scheme == "file" and hostname == "~":
        hostname = None
        path = "~" + path

    return (scheme, username, hostname, port, path)
```

markus-ferrell: @scheibelp ping for awareness. When I try to view this comment, it only shows up in the Files Changed tab and not the Conversations tab. I'm not sure what's going wrong with github.

lib/spack/spack/directory_layout.py
markus-ferrell: The value of `path` is not using the same path separator as `self.root`. os.path.join will append the path with windows appropriate separator but `path` uses the linux separator. We would get something like `C:\Users\...\test-debian6-core2/gcc-10.2.1/...`. We really only need the pathlib cast to switch the separators on the `path` variable and could rewrite it to `os.path.join(self.root, pathlib.Path(path))` if that's preferable.

scheibelp: This makes sense, although I think it would also be worth updating `relative_path_for_spec` to return a string-based path with appropriate separators (and furthermore we might want to update `DirectoryLayout.__init__` to convert `self.root`).

lib/spack/spack/version.py
scheibelp: > It is later used to concatenate onto a path. This works for linux because the absolute paths start with /

and also I assume because we have `lstrip(/)` (which works for Linux but not Windows)

> It is later used to concatenate onto a path. 

(this logic has since moved and is one of the reasons for conflicts, but I think the principle is the same in the new location)

I think by "later" you mean in `lookup_ref`.

If we ignore everything but the final component, we run the risk of caching two repositories with the same identifier. (e.g. if two different users fork a repo). I think it would be safer to hash (i.e. `spack.util.hash.b32_hash`) the full pathname and take the first several characters (to fall within the path length limits on Windows).

lib/spack/spack/util/url.py
scheibelp: Given that this is only called in `git_ref_lookup`, now that I've read through that, I'm less concerned with matching the old behavior exactly: I think we can resolve any issues here by discussing that function (which I have a separate comment about).

lib/spack/spack/version.py
markus-ferrell: Notes: 

7 char of hash on the path that would be constructed. Use hash as folder name for later clone. Make windows only for hashing.

markus-ferrell:
@spackbot fix style

lib/spack/spack/directory_layout.py
markus-ferrell: I moved the pathlib call into `relative_path_for_spec` which let me move the `path_for_spec` function back to no changes.

lib/spack/spack/version.py
markus-ferrell: I added the hashing for windows only and on linux, it will use the full path

lib/spack/spack/test/cmd/install.py
markus-ferrell: touched up these typos

lib/spack/spack/version/git_ref_lookup.py
```diff
@@ -93,14 +94,16 @@ def fetcher(self):
...
-            return os.path.abspath(self.pkg.git)
+        (scheme, netloc, path, params, query, fragment) = urllib.parse.urlparse(self.pkg.git)
+        path = path.lstrip("/")
+        if scheme == "file":
+            unique_dir = Path(path)
+            if sys.platform == "win32":
+                unique_dir = spack.util.hash.b32_hash(Path(path).as_posix())[-7:]
+            return Path("file") / unique_dir
+
+        (hostname, port) = netloc.split(":")
```
scheibelp: If netloc does not contain a port, this will fail.

Also, 

```
>>> parse.urlparse("git@github.com:spack/spack.git")
ParseResult(scheme='', netloc='', path='git@github.com:spack/spack.git', params='', query='', fragment='')
```

the original implementation did some extra work to remove `@` and `:` from this (overall, urlparse doesn't appear to break up this git-based URL into components).

Also, long URLs could trigger the same path-length issues we'd worry about with files. So we'd probably want to hash everywhere.

It occurs to me we could avoid all this parsing confusion by hashing everywhere (vs. we discussed only doing so on Windows). So like:

```
spack.util.hash.b32_hash(self.pkg.git)
```

And another option is to keep the logic (i.e. keep `url.parse_git_url`) and then hash on Windows like:

```
       try:
            components = [
                str(c).lstrip("/") for c in spack.util.url.parse_git_url(self.pkg.git) if c
            ]
            path = os.path.join(*components)
       except ValueError:
            # If it's not a git url, it's a local path
            path = os.path.abspath(self.pkg.git)
       if win32:
            return b32_hash(path)
       else:
            return path
```

markus-ferrell: urllib does have trouble with git-based urls. It also has trouble with amazon s3 urls. `parse_git_url` would cover those better but it seems like it would add a lot of complexity for a little more human readability. I like the idea of using hashing for everything. That makes the code much less complex so long as it passes ci.

lib/spack/spack/test/cmd/install.py
markus-ferrell: I changed the message to be more specific to this issue. 

We talked about this on slack but I'll reiterate it here to create a text history. The specific issue for these tests is that [this log context](https://github.com/spack/spack/blob/f3c080e54667115cf8b531184912a5bce49f802b/lib/spack/spack/installer.py#L2450) produces [this output](https://github.com/spack/spack/blob/f3c080e54667115cf8b531184912a5bce49f802b/lib/spack/spack/installer.py#L2455) after [this output](https://github.com/spack/spack/blob/f3c080e54667115cf8b531184912a5bce49f802b/lib/spack/spack/installer.py#L2461). The tests depend on the order and this currently do not output in the correct order on windows.

markus-ferrell:
@spackbot fix style

lib/spack/spack/test/cmd/install.py
```diff
@@ -913,6 +919,7 @@ def test_cdash_auth_token(tmpdir, mock_fetch, install_mockery, capfd):
             assert "Using CDash auth token from environment" in out
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Logger race condition in windows")
```
scheibelp: Is this a unique error compared to the other notes here about calling `log_output` in `install()`? I ask because the error message differs

markus-ferrell: This is the same reason as the others. I can switch this one to match the others.

lib/spack/spack/version/git_ref_lookup.py
```diff
@@ -131,7 +123,8 @@ def lookup_ref(self, ref) -> Tuple[Optional[str], int]:
         known version prior to the commit, as well as the distance from that version
         to the commit in the git repo. Those values are used to compare Version objects.
         """
-        dest = os.path.join(spack.paths.user_repos_cache_path, self.repository_uri)
+        pathlib_dest = Path(spack.paths.user_repos_cache_path) / self.repository_uri
+        dest = pathlib_dest.as_posix()
         if dest.endswith(".git"):
             dest = dest[:-4]
```
scheibelp: Now that the `repository_uri` is always a hash, this `if` will never trigger, so should be removed.

Also, will instantiating this `as_posix` cause problems with `mkdirp` on Windows (a few lines down)?

scheibelp (changes requested):

lib/spack/spack/version/git_ref_lookup.py
markus-ferrell: it shouldn't cause problems being a posix path but I'll switch it to a string cast to be consistent.

markus-ferrell:
@spackbot fix style

scheibelp (approved):

haampie:
@scheibelp / @markus-ferrell I'm uncomfortable with this path normalization you've introduced in `lib/spack/spack/directory_layout.py`.

Prefixes of installations are registered in the database. Now there's a change in how those paths are computed, also on Linux/macOS/BSD, since `str(Path(...))` normalizes `x/./y` to `x/y`.

This may lead to very hard to discover issues with relocation, when the prefix is recomputed (for some reason).

Can you leave the old behavior on non-Windows? Or fix it in another way?

johnwparent:
@haampie FYI Markus is no longer on the project, so feel free to ping me in his place for this and all future issues related to his PRs.

I will look into alternative solutions, but if that looks like it'll be a long refactor, I'll just fall back on to the old behavior for this operation on *nix.

PR #34697, opened 2022-12-26
py-numpy: add v1.24.1

adamjstewart:
https://github.com/numpy/numpy/releases/tag/v1.24.1

### DIFF
--- a/var/spack/repos/builtin/packages/py-numpy/package.py
+++ b/var/spack/repos/builtin/packages/py-numpy/package.py
@@ -23,6 +23,7 @@ class PyNumpy(PythonPackage):
     maintainers = ["adamjstewart", "rgommers"]
 
     version("main", branch="main")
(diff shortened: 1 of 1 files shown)


adamjstewart:
@rgommers I saw https://github.com/numpy/numpy/pull/22820, that doesn't allow us to support setuptools 64+ does it?

rgommers:
> @rgommers I saw [numpy/numpy#22820](https://github.com/numpy/numpy/pull/22820), that doesn't allow us to support setuptools 64+ does it?

Maybe - but probably not. NumPy itself won't build with the most recent `setuptools` versions, there are known issues on some platforms. That PR does resolve a problem though to make it possible to try. I know Debian was stuck - probably still stuck, but when you're Linux-only with a fixed set of package versions, maybe it works (no guarantees, but CI may give the answer). Spack is in a much more difficult situation here than Linux distros,given it has (almost) every version of every package plus multi-OS support.

rgommers (approved):

alalazo (approved):

PR #34698, opened 2022-12-26
qt: new versions 6.4.0, 6.4.1, 6.4.2

wdconinc:
- New libpsl vendored dependency in qt-base.
- New embree and tinyexr dependency in qt-quick3d.

We need to figure out a better way to deal with these vendored dependencies in src/3rdparty. Removing them was a way to make sure they are not used unintentionally. Many of these dependencies cannot be overridden with a QT_FEATURE_system_* flag and are included directly in cpp files. Many change versions from release to release, so even if they use system (ie spack managed) versions we need to support this in the depends_on lines.

What we can rely on?
- src/3rdparty is where vendored stuff is stored
- not much else...

Possible ways to deal with this:
- Change vendor_deps_to_keep to dict with versions, eg
```
vendor_deps_to_keep = {
    "xatlas": "@6:",
    "embree": "@6.4:",
    "tinyexr": "@6.4:",
}
```
- Similarly introduce system_deps_to_use:
```
system_deps_to_use = {
    "assimp@5.2:": "@6:",
}
```
and derive depends_on and QT_FEATURE_system_* from this dict.


wdconinc:
The main issue with removing all src/3rdparty deps except those in `vendor_deps_to_keep` is that this is pretty fragile. It may work on a particular test configuration but is likely to fail to build on eg mac or with different variants of only those configurations require a certain 3rdparty dep.

On the other hand, we don't want to use a vendored dependency when we think we are using the system version. So maybe we need to invert this logic and remove only those src/3rdparty deps that we explicitly ask to use the system version of.

And finally, we might want to have a summary at the end of the job that is a grep for all src/3rdparty compilation in the logs...

var/spack/repos/builtin/packages/qt-base/package.py
```diff
@@ -4,24 +4,99 @@
...
+        return _url.format(qualname.lower())
+
+    @staticmethod
+    def get_list_url(qualname):
+        _list_url = "https://github.com/qt/{}/tags"
+        return _list_url.format(qualname.lower())
+
+    generator = "Ninja"
 
     maintainers = ["wdconinc", "sethrj"]
```
wdconinc:
```suggestion
    maintainers("wdconinc", "sethrj")
```

wdconinc:
@spackbot fix style

wdconinc:
Since all these qt-* packages share a lot of the build approach, I've moved a lot into a `QtPackage` base layer, and every component (including qt-base) inherits from this. That allows us to specify a lot of boilerplate only once:
- maintainers
- generator
- cmake, python, etc dependencies
- removing of vendor deps
- QT_ADDITIONAL_PACKAGES_PREFIX_PATH
- installing config.summary

AFAIK, we cannot access the versions of the current package in its definition. But the loop to keep the dependencies version synced is now rewritten by at least pulling the versions from qt-base. I don't see how to pull this in the QtPackage base layer.

@sethrj I think this is ready for another round of review.

wdconinc:
I am using `__qualname__` which is only available in python 3.3+. I figured spack audit was going to complain about this... Are we ok with python 3.3 as requirement?

sethrj (approved):
Nice refactor!

PR #34699, opened 2022-12-26
libpsl: new versions through 0.21.2

wdconinc:
This adds the final bugfix versions through the 0.21.2 just released.

With 0.21.1 the tag name pattern was changed, hence url_for_version.

No major changes to build system (meson is also supported upstream as of 0.21.0, but keeping it out of the spack package for now).


wdconinc:
@spackbot fix style

alalazo (approved):

PR #34700, opened 2022-12-26
Consolidate how Spack uses `git`

tgamblin:
Local `git` tests will fail with `fatal: transport 'file' not allowed` when using git 2.38.1 or higher, due to a fix for `CVE-2022-39253`.

This was fixed in CI in #33429, but that doesn't help the issue for anyone's local environment. Instead of fixing this with git config in CI, we should ensure that the tests run anywhere.

- [x] Introduce `spack.util.git`.
- [x] Use `spack.util.git.get_git()` to get a git executable, instead of `which("git")` everywhere.
- [x] Make all `git` tests use a `git` fixture that goes through `spack.util.git.get_git()`.
- [x] Add `-c protocol.file.allow=always` to all `git` invocations under `pytest`.
- [x] Revert changes from #33429, which are no longer needed.

### DIFF
--- a/lib/spack/spack/test/conftest.py
+++ b/lib/spack/spack/test/conftest.py
@@ -68,9 +68,11 @@ def ensure_configuration_fixture_run_before(request):
 
 
 @pytest.fixture(scope="session")
-@pytest.mark.skipif(not spack.util.git.git(), reason="requires git to be installed")
 def git():
     """Fixture for tests that use git."""
+    if not spack.util.git.git():
+        pytest.skip("requires git to be installed")
+
     return spack.util.git.git(required=True)
 
 
--- a/lib/spack/spack/test/conftest.py
+++ b/lib/spack/spack/test/conftest.py
@@ -46,6 +46,7 @@ import spack.store
 import spack.subprocess_context
 import spack.test.cray_manifest
 import spack.util.executable
+import spack.util.git
 import spack.util.gpg
 import spack.util.spack_yaml as syaml
 import spack.util.url as url_util
@@ -66,12 +67,18 @@ def ensure_configuration_fixture_run_before(request):
         request.getfixturevalue("mutable_config")
 
 
+@pytest.fixture(scope="session")
+@pytest.mark.skipif(not spack.util.git.get_git(), reason="requires git to be installed")
+def git():
+    """Fixture for tests that use git."""
+    return spack.util.git.get_git(required=True)
+
+
 #
 # Return list of shas for latest two git commits in local spack repo
 #
 @pytest.fixture(scope="session")
-def last_two_git_commits():
-    git = spack.util.executable.which("git", required=True)
+def last_two_git_commits(git):
     spack_git_path = spack.paths.prefix
     with working_dir(spack_git_path):
         git_log_out = git("log", "-n", "2", output=str, error=os.devnull)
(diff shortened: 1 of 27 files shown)


lib/spack/spack/util/git.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@llnl.util.lang.memoized
+def get_git(required=False):
+    """Get a git executable.
+
+    Arguments:
+        required: if ``True``, fail if ``git`` is not found. By default return ``None``.
+    """
+    git = spack.util.executable.which("git", required=required)
```
alalazo: (Minor) Why not calling this just  `git`? Also, can we add a `Raises:` part in the docstring?

lib/spack/spack/util/git.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # If we're running under pytest, add this to ignore the fix for CVE-2022-39253 in
+    # git 2.38.1+. Do this in one place; we need git to do this in all parts of Spack.
+    if "pytest" in sys.modules:
+        git.add_default_arg("-c")
+        git.add_default_arg("protocol.file.allow=always")
```
alalazo: Should we work around a CVE without letting users know? Where would this be relevant besides CI?

lib/spack/spack/test/conftest.py
```diff
@@ -66,12 +67,18 @@ def ensure_configuration_fixture_run_before(request):
         request.getfixturevalue("mutable_config")
 
 
+@pytest.fixture(scope="session")
+@pytest.mark.skipif(not spack.util.git.get_git(), reason="requires git to be installed")
```
alalazo: Looking [at the docs](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-skip-ref), it seems `skip` and `skipif` are always applied to test functions, not to fixtures. Does the `skipif` above propagate to tests using the fixture?

alalazo (commented):
This basically LGTM. I just left a few minor comments:
1. The rename `spack.util.git.get_git()` -> `spack.util.git.git()` is absolutely minor. I think we are pretty inconsistent at the moment, so I'd be fine with either if you prefer the current state.
2. I need to double check if `skipif` can be applied on a fixture (I can't find clear docs on it, all the examples seem to suggest we need to use the imperative method during the setup phase)

Other than that it's :100: to unify code!

lib/spack/spack/util/git.py
tgamblin: We know the CVE doesn't affect us because it's not a multi-user environment and we created the directories ourselves.

lib/spack/spack/test/conftest.py
tgamblin: yeah good catch -- apparently you have to use `pytest.skip()`

tgamblin:
@alalazo all should be fixed

lib/spack/spack/util/git.py
tgamblin: I tried to add Raises and had to add an exception since Sphinx can't seem to resolve the name (fully qualified or otherwise).

tgamblin:
@alalazo I can't seem to get past this sphinx issue for `Raises:`:

```
/Users/gamblin2/src/spack/lib/spack/spack/util/git.py:docstring of spack.util.git.git:1: WARNING: py:exc reference target not found: CommandNotFoundError
```

It doesn't seem to matter if it's fully qualified or if I add it to `nitpick_ignore` in `conf.py`.

But everything else is done.

alalazo (approved):
Sphinx tests seems to be passing on readthedocs. I think this one just needs to pass Gitlab, once it is available again.

PR #34702, opened 2022-12-26
cernlib: depends_on freetype, libnsl, libxcrypt, openssl; and patch

wdconinc:
In addition to #34448, cernlib depends on these additional packages. This also applies a patch to the current release in which crypto is specified where libcrypt (in libxcrypt) is actually needed. Because the upstream git repository is behind a CERN login, we cannot patch by gitlab URL link.

- [x] This PR will require merging #34448 first, and likely trivial conflict resolution. Hence, this is marked as draft.


wdconinc:
@spackbot fix style

alalazo (approved):

var/spack/repos/builtin/packages/cernlib/package.py
```diff
@@ -20,10 +20,19 @@ class Cernlib(CMakePackage):
...
 
+    depends_on("freetype")
     depends_on("motif")
+    depends_on("libnsl")
     depends_on("libx11")
     depends_on("libxaw")
     depends_on("libxt")
+    depends_on("libxcrypt")
+
+    depends_on("openssl", when="os=linux")
```
haampie, line 31: platform=linux?

PR #34703, opened 2022-12-26
bugfix: compiler bootstrapping tests

tgamblin:
This compiler bootstrapping test breaks on (my?) M1 mac beacuse it can't find the bootstrap compiler after it's added to configuration. Removing the archspec host fixture seems to fix it.

@becker33 not sure if this is the right fix. Can you take a look?

### DIFF
--- a/lib/spack/spack/test/installer.py
+++ b/lib/spack/spack/test/installer.py
@@ -488,7 +488,7 @@ def test_update_tasks_for_compiler_packages_as_compiler(mock_packages, config, m
 
 
 def test_bootstrapping_compilers_with_different_names_from_spec(
-    install_mockery, mutable_config, mock_fetch, archspec_host_is_spack_test_host
+    install_mockery, mutable_config, mock_fetch
 ):
     with spack.config.override("config:install_missing_compilers", True):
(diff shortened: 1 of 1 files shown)


becker33:
@tgamblin the fixture was added by @johnwparent in https://github.com/spack/spack/pull/33385, I assume it was added for a reason. I suspect this will break the tests on windows, but I don't immediately see why.

tgamblin:
@johnwparent can you say why this is needed for windows?

johnwparent:
> @johnwparent can you say why this is needed for windows?

@tgamblin @becker33 this was required due to Spack's mocked testing 'platform' and archspec detecting the actual platform introducing inconsistency into Spack, which resulted in every concretization/attempt to install anything resulting in "no compiler for spec" errors.

More discussion involving myself, @scheibelp and @alalazo can be found [here](https://github.com/spack/spack/pull/33385#discussion_r1040036318)

This solution was by no means intended to be permanent, but at the moment, the Windows test will not pass without it. We could of course just skip the test on Windows as we do with many others and punt this to a later date, as this issue only occurs under testing conditions.

alalazo:
Outdated by #46237

PR #34707, opened 2022-12-28
docs: avoid errors by using type hints instead of doc types

tgamblin:
There are a number of places in our docstrings where we write "list of X" as the type, even though napoleon doesn't actually support this. It ends up causing warnings when generating docs.

Now that we require Python 3, we don't have to rely on type hints in docs -- we can just use Python type hints and omit the types of args and return values from docstrings.

We should probably do this for all types in docstrings eventually, but this PR focuses on the ones that generate warnings during doc builds.

Some `mypy` annoyances we should consider in the future:
1. Adding some of these type annotations gets you:
    ```
    note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
    ```
   because they are in unannotated functions (like constructors where we don't really need any annotations).
   You can silence these with `disable_error_code = "annotation-unchecked"` in `pyproject.toml`
2. Right now we support running `mypy` in Python `3.6`.  That means we have to support `mypy` `.971`, which does not support `disable_error_code = "annotation-unchecked"`, so I just filter `[annotation-unchecked]` lines out in `spack style`.
3. I would rather just turn on `check_untyped_defs` and get more `mypy` coverage everywhere, but that will require about 1,000 fixes.  We should probably do that eventually.
4. We could also consider only running `mypy` on newer python versions.  This is not easy to do while supporting `3.6`, because you have to use `if TYPE_CHECKING` for a lot of things to ensure that 3.6 still parses correctly.  If we only supported `3.7` and above we could use [`from __future__ import annotations`](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations-import-pep-563), but we have to support 3.6 for now. Sigh.

- [x] Convert a number of docstring types to Python type hints
- [x] Get rid of "list of" wherever it appears

### DIFF
--- a/lib/spack/spack/cmd/style.py
+++ b/lib/spack/spack/cmd/style.py
@@ -48,6 +48,13 @@ tool_names = [
 #: tools we run in spack style
 tools = {}
 
+#: warnings to ignore in mypy
+mypy_ignores = [
+    # same as `disable_error_code = "annotation-unchecked"` in pyproject.toml, which
+    # doesn't exist in mypy 0.971 for Python 3.6
+    "[annotation-unchecked]",
+]
+
 
 def is_package(f):
     """Whether flake8 should consider a file as a core file or a package.
@@ -210,6 +217,10 @@ def rewrite_and_print_output(
 
     for line in output.split("\n"):
         if not line:
+            continue
+        if any(ignore in line for ignore in mypy_ignores):
+            # some mypy annotations can't be disabled in older mypys (e.g. .971, which
+            # is the only mypy that supports python 3.6), so we filter them here.
             continue
         if not args.root_relative and re_obj:
             line = re_obj.sub(translate, line)
--- a/lib/spack/spack/config.py
+++ b/lib/spack/spack/config.py
@@ -36,7 +36,7 @@ import os
 import re
 import sys
 from contextlib import contextmanager
-from typing import List
+from typing import Dict, List, Optional
 
 import ruamel.yaml as yaml
 from ruamel.yaml.error import MarkedYAMLError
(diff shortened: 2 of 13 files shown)


var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1050,7 +1050,8 @@ def include(self):
         return os.path.join("include", "python{}".format(self.version.up_to(2)))
 
     def setup_run_environment(self, env):
-        env.prepend_path("CPATH", os.pathsep.join(self.spec["python"].headers.directories))
+        # this call prevents us from using most python packages without dev headers
+        pass  # env.prepend_path("CPATH", os.pathsep.join(self.spec["python"].headers.directories))
```
adamjstewart: I would just delete the whole thing

tgamblin: this was a mistake -- moving this to another PR

lib/spack/spack/cmd/style.py
```diff
@@ -48,6 +48,13 @@ def grouper(iterable, n, fillvalue=None):
 #: tools we run in spack style
 tools = {}
 
+#: warnings to ignore in mypy
+mypy_ignores = [
+    # same as `disable_error_code = "annotation-unchecked"` in pyproject.toml, which
+    # doesn't exist in mypy 0.971 for Python 3.6
```
adamjstewart, line 54: Mentioned this in chat, but we prob don't need to support running mypy with Python 3.6.

lib/spack/spack/config.py
```diff
@@ -391,63 +391,66 @@ class Configuration(object):
     This class makes it easy to add a new scope on top of an existing one.
     """
 
-    def __init__(self, *scopes):
+    # convert to typing.OrderedDict when we drop 3.6, or OrderedDict when we reach 3.9
+    scopes: Dict[str, ConfigScope]
+
+    def __init__(self, *scopes: ConfigScope):
```
adamjstewart, line 397: Should this be `List[ConfigScope]`?

lib/spack/spack/config.py
```diff
@@ -391,63 +391,66 @@ class Configuration(object):
...
         """Add a higher precedence scope to the Configuration."""
         tty.debug("[CONFIGURATION: PUSH SCOPE]: {}".format(str(scope)), level=2)
         self.scopes[scope.name] = scope
 
     @_config_mutator
-    def pop_scope(self):
+    def pop_scope(self) -> ConfigScope:
         """Remove the highest precedence scope and return it."""
-        name, scope = self.scopes.popitem(last=True)
+        name, scope = self.scopes.popitem(last=True)  # type: ignore[call-arg]
```
adamjstewart, line 419: Ignores always scare me. What's the reason to ignore here?

tgamblin: This is because you can't subscript `OrderedDict` in 3.6, so you have to ignore the `OrderedDict` specific call.  I'll fix this in another PR and only run mypy with newer python.

lib/spack/spack/cmd/style.py
tgamblin: Going to fix this and other tools in another PR

lib/spack/spack/config.py
tgamblin: No: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

adamjstewart: Oh right

adamjstewart (approved):

PR #34709, opened 2022-12-28
gmake: 4.4, remove alpha release

haampie:
With fifo-based jobserver, as well as --load-average :tada: 

@finkandreas you should use this version of gnu make too, it solves the issue of perl not returning job tokens (I never looked into that issue, but named pipes are definitely more stable across child processes), and on top of that you can do `--load-average N` to avoid over-subscription on shared systems, or when build systems take more jobs than they should


alalazo (approved):

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

haampie:
@spackbot run pipeline

PR #34710, opened 2022-12-28
New version for openCARP packages, v12.0

MarieHouillon:
Add a new version for the packages of the openCARP ecosystem, opencarp, meshtool and py-carputils.

Also sets the default cmake build type to "Release".


var/spack/repos/builtin/packages/meshtool/package.py
```diff
@@ -16,6 +16,7 @@ class Meshtool(MakefilePackage):
 
     version("master", branch="master", preferred=True)
     # Version to use with openCARP releases
+    version("oc12.0", commit="867431d")
```
haampie, line 19: same commit as oc11.0, typo or intentional?

haampie (commented):
See above

var/spack/repos/builtin/packages/meshtool/package.py
MarieHouillon: This is intentional

tldahlgren (commented):
Confirmed all of the commits exist.

alalazo (approved):

PR #34712, opened 2022-12-28
Simplify creation of test and install reports

alalazo:
The code in Spack to generate install and test reports currently suffers from unneeded complexity. For instance, we have classes in Spack core packages, like `spack.reporters.CDash`, that need an `argparse.Namespace` to be initialized and have "hard-coded" string literals on which they branch to change their behavior:
```python
if do_fn.__name__ == "do_test" and skip_externals:
    package["result"] = "skipped"
else:
    package["result"] = "success"
package["stdout"] = fetch_log(pkg, do_fn, self.dir)
package["installed_from_binary_cache"] = pkg.installed_from_binary_cache
if do_fn.__name__ == "_install_task" and installed_already:
    return
```
This PR attempt to polish the major issues encountered in both `spack.report` and `spack.reporters`. In details:

- [x] `spack.reporters` is now a package that contains both the base class `Reporter` and all the derived classes (`JUnit` and `CDash`)
- [x] Classes derived from `spack.reporters.Reporter` don't take an `argparse.Namespace` anymore as argument to `__init__`. The rationale is that code for commands should be built upon Spack core classes, not vice-versa.
- [x] An `argparse.Action` has been coded to create the correct `Reporter` object based on command line arguments
- [x] The context managers to generate reports from either `spack install` or from `spack test` have been greatly simplified, and have been made less "dynamic" in nature. In particular, the `collect_info` class has been deleted in favor of two more specific context managers.
- [x] The `InfoCollector` class has been turned into a simple hierarchy, so to avoid conditional statements within methods that assume a knowledge of the context in which the method is called.

### DIFF
--- a/lib/spack/spack/reporters/cdash.py
+++ b/lib/spack/spack/reporters/cdash.py
@@ -47,7 +47,8 @@ CDASH_PHASES.add("update")
 
 
 CDashConfiguration = collections.namedtuple(
-    "CDashConfiguration", ["upload_url", "packages", "build", "site", "buildstamp", "track"]
+    "CDashConfiguration",
+    ["upload_url", "packages", "build", "site", "buildstamp", "track", "ctest_parsing"],
 )
 
 
@@ -105,6 +106,7 @@ class CDash(Reporter):
             self.revision = git("rev-parse", "HEAD", output=str).strip()
         self.generator = "spack-{0}".format(spack.main.get_version())
         self.multiple_packages = False
+        self.ctest_parsing = configuration.ctest_parsing
 
     def report_build_name(self, pkg_name):
         return (
@@ -396,7 +398,7 @@ class CDash(Reporter):
                     directory_name,
                     package,
                     duration,
-                    input_data["ctest-parsing"],
+                    self.ctest_parsing,
                 )
 
         self.finalize_report()
--- a/lib/spack/spack/reporters/cdash.py
+++ b/lib/spack/spack/reporters/cdash.py
@@ -233,13 +233,13 @@ class CDash(Reporter):
                 f.write(t.render(report_data))
             self.upload(phase_report)
 
-    def build_report(self, directory_name, input_data):
+    def build_report(self, directory_name, specs):
         # Do an initial scan to determine if we are generating reports for more
         # than one package. When we're only reporting on a single package we
         # do not explicitly include the package's name in the CDash build name.
         self.multipe_packages = False
         num_packages = 0
-        for spec in input_data["specs"]:
+        for spec in specs:
             # Do not generate reports for packages that were installed
             # from the binary cache.
             spec["packages"] = [
@@ -257,7 +257,7 @@ class CDash(Reporter):
                 break
 
         # Generate reports for each package in each spec.
-        for spec in input_data["specs"]:
+        for spec in specs:
             duration = 0
             if "time" in spec:
                 duration = int(spec["time"])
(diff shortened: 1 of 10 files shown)


lib/spack/spack/reporters/cdash.py
```diff
@@ -239,13 +233,13 @@ def clean_log_event(event):
                 f.write(t.render(report_data))
             self.upload(phase_report)
 
-    def build_report(self, directory_name, input_data):
+    def build_report(self, directory_name, specs):
```
alalazo, line 236: This is another part of the API that bugged me. The base class was accepting a `filename` (it started with `JUnit` which produces a single file). To accomodate CDash the string became a `directory_name`. Should we just call this "location" in the base class and document that the meaning of it varies according to the specific implementation?

lib/spack/spack/reporters/cdash.py
```diff
@@ -42,8 +42,14 @@
...
 # Initialize data structures common to each phase's report.
-cdash_phases = set(map_phases_to_cdash.values())
-cdash_phases.add("update")
+CDASH_PHASES = set(MAP_PHASES_TO_CDASH.values())
+CDASH_PHASES.add("update")
+
+
+CDashConfiguration = collections.namedtuple(
+    "CDashConfiguration",
+    ["upload_url", "packages", "build", "site", "buildstamp", "track", "ctest_parsing"],
```
alalazo, line 51: Here I maintained the same semantic for `ctest_parsing` as before[^1], but I see no place where we have such an argument defined. If this is a leftover, and I didn't overlook some definition, I can remove a few lines more.

[^1]: In fact the entire PR is just restructuring code, without changing observed behavior (or at least that is the aim).

alalazo (commented):
Another structural change I'd do, maybe in a following PR, is to:
1. Aggregate all `Reporter`s in the same file
2. Move the code in `spack.report` within the `spack.reporters` package

I didn't do it here since I think it would prevent the UI to display correct diffs for the code.

tgamblin:
@spackbot run pipeline

tgamblin:
@alalazo not sure why the pipeline is failing here.

alalazo:
@tgamblin Given that it takes > 6hrs, I think @zackgalbreath mentioned helper pods that might be killed and leave runners hanging? For reference, here's the pipeline failing:
- https://gitlab.spack.io/spack/spack/-/pipelines/246215

I'll try ~a second restart~ fixing the conflict and pushing again.

alalazo:
@tgamblin So, this time the concretization jobs failing before passed, but there's one that got scheduled on the wrong node:
- https://gitlab.spack.io/spack/spack/-/jobs/5060366

I'll let the other jobs finish and restart again.

alalazo:
@spackbot run pipeline

alalazo:
This is still failing with the same scheduling error, but on different pipelines. Given that this PR is still to be reviewed and that the union of "pipelines passed over re-runs" cover all the possible pipelines (so the errors seem spurious), I'll give up trying to get a :heavy_check_mark: for now

lib/spack/spack/reporters/cdash.py
tgamblin: Maybe when @zackgalbreath gets back he can clarify this but I think it's good for now.

lib/spack/spack/reporters/cdash.py
tgamblin: I think it's either that, or change JUnit to generate the single file in a directory.  I kind of like the latter better, as it means that every reporter gets its own scratch directory to work in.  But if JUnit only needs one file maybe it's overkill.

tgamblin (approved):
Everything looks much more explicit here and I think the logic is clearer than before.

PR #33405, opened 2022-10-19
rocm packages: Update the version for 5.3.0 release

srekolam:
This PR contains the version updates for some of the rocm recipes for rocm-5.3.0 release. This PR is dependent on PR #33320 .

Issue #5269, opened 2017-09-01
Push towards documented consensus on shared packages


becker33:
As discussed in the telcon this week: https://github.com/LLNL/spack/wiki/Telcon%3A-2017-08-31

Also as discussed on the telcon, I'm going to use "common variants" to refer to variants that happen to be present on many packages, and "global variants" or "universal variants" to refer to a variant that is technically ensured to be present for all packages (or some major subset).

We need to document our community standards for common variants. There are reasons that we don't use universal variants for shared/static libraries, but community standards make Spack more usable. Unless/until we put those community standards into the documentation and start rejecting pull requests that don't adhere to those standards, we can't get the full benefits of standardization.

I have two concrete proposals for what the standard for shared/static variants should be. These will also interact with the (already standardized) pic variant. The goal is to hash out any corner cases of my proposals, choose one of them (or theoretically some third proposal), and document and implement it. We will then request changes to any incoming packages that fail to adhere to the standard.

Standard 1. This is what we (mostly) currently do.
All packages that can build either static or shared libraries have a variant `shared`. This variant defaults to `True`. When `True`, we build static and shared libraries. When false, we build only static libraries.

Pros: Already in use for most packages
Cons: Does not allow us to only build shared. Requires us to depend on `~shared` or `+pic` versions of all dependencies when building `~static`. Developers think of building "static only" more than "no shared"

Standard 2. Use multi-valued variants
All packages that can provide either static or shared libraries have a variant `libraries`, with possible values `static` and `shared`, non-exclusively. We provide whichever libraries are requested in this variant.

Pros: Allows any combination of shared and static libraries.
Cons: Requires changes to many packages. Requires manual variant forwarding (until #391 is reintroduced) for the `libraries` variant.

I'm leaning slightly towards option 2. I'd like input on possible corner cases for each of them and how they may improve or disrupt existing workflows.

@citibeth @markcmiller86 @tgamblin @alalazo @adamjstewart I know you have all had previous feelings on this topic.

davydden:
If we allow building both shared and static simultaneously, can we get in trouble when within the same DAG there are packages which use both?

If not, I like multivalued variant approach.

scheibelp:
I think we were also talking about the possibility of a single-valued variant with the possible values `shared`, `static`, and `static_and_shared`. The idea would be that `static` implies building only the static libraries.

citibeth:
OK, here goes how I've thought it through...

[...]

PR #34714, opened 2022-12-28
googletest: set PIC flag for static build

renjithravindrankannath:
Rocm-validation-suite failing on Centos because CMAKE_POSITION_INDEPENDENT_CODE was not added in google test.
Adding the argument CMAKE_POSITION_INDEPENDENT_CODE in google test.

### DIFF
--- a/var/spack/repos/builtin/packages/googletest/package.py
+++ b/var/spack/repos/builtin/packages/googletest/package.py
@@ -50,6 +50,7 @@ class Googletest(CMakePackage):
             # New style (contains both Google Mock and Google Test)
             args.append(self.define("BUILD_GTEST", True))
             args.append(self.define_from_variant("BUILD_GMOCK", "gmock"))
+            args.append(self.define("CMAKE_POSITION_INDEPENDENT_CODE", True))
 
         return args
 
--- a/var/spack/repos/builtin/packages/hipfft/package.py
+++ b/var/spack/repos/builtin/packages/hipfft/package.py
@@ -114,5 +114,7 @@ class Hipfft(CMakePackage):
         elif self.spec.satisfies("@5.2.0:"):
             args.append(self.define("CMAKE_MODULE_PATH", self.spec["hip"].prefix.lib.cmake.hip))
             args.append(self.define("BUILD_FILE_REORG_BACKWARD_COMPATIBILITY", True))
+        if self.spec.satisfies("@5.3.0:"):
+            args.append(self.define("CMAKE_INSTALL_LIBDIR", "lib"))
 
         return args
--- a/var/spack/repos/builtin/packages/hipfft/package.py
+++ b/var/spack/repos/builtin/packages/hipfft/package.py
@@ -114,7 +114,5 @@ class Hipfft(CMakePackage):
         elif self.spec.satisfies("@5.2.0:"):
             args.append(self.define("CMAKE_MODULE_PATH", self.spec["hip"].prefix.lib.cmake.hip))
             args.append(self.define("BUILD_FILE_REORG_BACKWARD_COMPATIBILITY", True))
-        if self.spec.satisfies("@5.3.0:"):
-            args.append(self.define("CMAKE_INSTALL_LIBDIR", "lib"))
 
         return args


sethrj:
@renjithravindrankannath Please split these into two PRs since they're unrelated changes to different packages. You can just remove the hipfft changes from this one and create a new PR for that.

var/spack/repos/builtin/packages/googletest/package.py
```diff
@@ -50,6 +50,7 @@ def cmake_args(self):
             # New style (contains both Google Mock and Google Test)
             args.append(self.define("BUILD_GTEST", True))
             args.append(self.define_from_variant("BUILD_GMOCK", "gmock"))
+            args.append(self.define("CMAKE_POSITION_INDEPENDENT_CODE", True))
```
sethrj, line 53: This should only matter if `~shared` because shared libraries always use PIE/PIC. Can you provide more details on the use case with static libraries that require PIC?

renjithravindrankannath: Hi @sethrj ,
rocm-validation-suite@5.3.0 spec ([rocm-validation-suite-5.3.0-spec.txt](https://github.com/spack/spack/files/10317044/rocm-validation-suite-5.3.0-spec.txt)) includes `^googletest@1.12.1%gcc@8.3.1+gmock~ipo+pthreads~shared build_system=cmake build_type=RelWithDebInfo cxxstd=11 arch=linux-centos8-skylake_avx512`

renjithravindrankannath:
> @renjithravindrankannath Please split these into two PRs since they're unrelated changes to different packages. You can just remove the hipfft changes from this one and create a new PR for that.

Removed the hipfft change from this PR.

var/spack/repos/builtin/packages/googletest/package.py
sethrj: @renjithravindrankannath Do you know why it explicitly chooses static but happens to get built into another shared library (which is the only situation where the PIC flag should matter)? And again, this change should only apply when `~shared`.

cgmb:
I'm not sure if we should be changing this setting on individual libraries on a case-by-case basis. The eventual result will be that you'll never know whether you're going to get PIC or not when you spack install a library with `~shared`, because we'll have enabled it for some packages but not others.

One of the Spack core developers might be able to provide some insight into the best way of handling this, but one possibility would be to add a conditional variant to CMakePackage for `pic`, much like [the `ipo` variant](https://github.com/spack/spack/blob/v0.19.0/lib/spack/spack/build_systems/cmake.py#L64-L71). Then, packages that actually need a dependency with `~shared +pic` can explicitly state their requirement.

> hipfft library path was lib64 in Centos. Setting this to lib to make uniform across the platform.

I believe that is addressed in #33405.

cgmb:
> one possibility would be to add a conditional variant to CMakePackage for `pic`, much like [the `ipo` variant](https://github.com/spack/spack/blob/v0.19.0/lib/spack/spack/build_systems/cmake.py#L64-L71). Then, packages that actually need a dependency with `~shared +pic` can explicitly state their requirement.

A smaller step might just be doing this on Googletest directly, rather than on CMakePackage.

sethrj:
Yeah I agree this shouldn't be always on, and there is lots of inconsistency between packages. We could for the time being add a `variant("+pic", when="~shared")` flag. I still though don't understand why `rocm-validation-suite` wants a static version of googletest but seems to have lots of other shared libraries. This seems more like a concretization bug than a feature needed for googletest.

*EDIT*: well that was easy to find: in the validation suite spec it requires
```
    depends_on("googletest~shared", when="@4.5.0:")
```
but why? And why allow other libraries in the spec to be `+shared`?

cgmb:
> _EDIT_: well that was easy to find: in the validation suite spec it requires
> 
> ```
>     depends_on("googletest~shared", when="@4.5.0:")
> ```
> 
> but why? And why allow other libraries in the spec to be `+shared`?

That's a good point. After changing the package definition

```patch
diff --git a/var/spack/repos/builtin/packages/rocm-validation-suite/package.py b/var/spack/repos/builtin/packages/rocm-validation-suite/package.py
index e387696de5..5e478b5ef1 100644
--- a/var/spack/repos/builtin/packages/rocm-validation-suite/package.py
+++ b/var/spack/repos/builtin/packages/rocm-validation-suite/package.py
@@ -118,8 +118,8 @@ class RocmValidationSuite(CMakePackage):
 
     depends_on("cmake@3.5:", type="build")
     depends_on("zlib", type="link")
-    depends_on("yaml-cpp~shared")
-    depends_on("googletest~shared", when="@4.5.0:")
+    depends_on("yaml-cpp")
+    depends_on("googletest", when="@4.5.0:")
     depends_on("doxygen", type="build", when="@4.5.0:")
 
     def setup_build_environment(self, build_env):
```

it built fine¹ (and seemed to run correctly too, though I have never used rvs before).

¹ Well, I did have to `apt install libpci-dev`. I don't see a Spack package for that library, but one should probably be created.

sethrj:
OK @renjithravindrankannath , since the PIC stuff is more contentious (see https://github.com/spack/spack/issues/5269), can you close this in favor of a PR for fixing `rocm-validation-suite`?

renjithravindrankannath:
> OK @renjithravindrankannath , since the PIC stuff is more contentious (see #5269), can you close this in favor of a PR for fixing `rocm-validation-suite`?

@sethrj , Thanks for your review. I verified 5.3.0 with googletest+shared in RVS and it works. 
Currently I am verify the same with earlier versions. I will close this PR once it is complete and update RVS recipe accordingly.

renjithravindrankannath:
Verified RVS with googletest+shared on 5.1.0. Closing the PR since RVS works without ~shared build.

renjithravindrankannath:
Created PR #34802 to set the dependency to googletest shared build

PR #34717, opened 2022-12-29
Setting lib install directory for hipfft

renjithravindrankannath:
Setting CMAKE_INSTALL_LIBDIR to lib since hipfft installing it to lib64 in CentOS


haampie:
This is thanks to GNUInstallDirs. What issue is this causing?

renjithravindrankannath:
This change is merged as part of #33405

PR #34216, opened 2022-11-30
spdlog: Conflict with external fmt if ~fmt_external

blue42u:
Fixes #34213

`spdlog ~fmt_external` bundles a version of `fmt` (`@8.1.1` or so in this case) which becomes part of the ABI for `libspdlog.so`. If a dependency also uses `fmt` (e.g. default `@9.1.0`) the build will fail due to the version mismatch on the `fmt` library.

This PR adds a conflict so `spdlog ~fmt_external` never allows another (separate) `fmt` to exist in the spec.

PR #34731, opened 2022-12-30
spdlog: new version 1.11.0 depends_on fmt@9: when +fmt_external

wdconinc:
This adds a new version of spdlog,
https://github.com/gabime/spdlog/releases/tag/v1.11.0

While the release notes are ambiguous, I think that this PR, https://github.com/gabime/spdlog/issues/2485, indicates that spdlog from that point on uses features of fmt@9:.


wdconinc:
See also #34216, @blue42u and @trws.

Maybe defaulting to the vendored fmt in #33379 was the less optimal default?

haampie (approved):

PR #34732, opened 2022-12-30
style: fix spurious `mypy` errors from `numpy`

tgamblin:
Spack imports `pytest`, which *can* import `numpy`. Recent versions of `numpy` require Python 3.8 or higher, and they use 3.8 type annotations in their type stubs (`.pyi` files). At the same time, we tell `mypy` to target Python 3.7, as we still support older versions of Python.

What all this means is that if you run `mypy` on `spack`, `mypy` will follow all the static import statements, and it ends up giving you this error when it finds numpy stuff that is newer than the target Python version:

```
==> Running mypy checks
src/spack/var/spack/environments/default/.spack-env/._view/4g7jd4ibkg4gopv4rosq3kn2vsxrxm2f/lib/python3.11/site-packages/numpy/__ini\
t__.pyi:638: error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)
  mypy found errors
```

We can fix this by telling `mypy` to skip all imports of `numpy` in `pyproject.toml`:

```toml
   [[tool.mypy.overrides]]
   module = 'numpy'
   follow_imports = 'skip'
   follow_imports_for_stubs = true
```

- [x] don't follow imports from `numpy` in `mypy`
- [x] get rid of old rule not to follow `jinja2` imports, as we now require Python 3

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,11 +141,14 @@ ignore_missing_imports = true
   ignore_errors = true
   ignore_missing_imports = true
 
-  # jinja has syntax in it that requires python3 and causes a parse error
-  # skip importing it
+  # pytest (which we depend on) optionally imports numpy, which requires Python 3.8 in
+  # recent versions. mypy still imports its .pyi file, which has positional-only
+  # arguments, which don't work in 3.7, which causes mypy to bail out early if you have
+  # numpy installed.
   [[tool.mypy.overrides]]
-  module = 'jinja2'
+  module = 'numpy'
   follow_imports = 'skip'
+  follow_imports_for_stubs = true
 
 [tool.pyright]
 useLibraryCodeForTypes = true


adamjstewart:
This was a bug in mypy that affects Python 3.10.7+ and was fixed in 0.981, see https://github.com/python/mypy/issues/13627 for details. We correctly note this in our `py-mypy` package. Personally I wouldn't bother adding a workaround for this and instead instruct users to use newer mypy versions.

tgamblin:
@adamjstewart: this isn't that bug. I'm using `0.991`.  See https://github.com/python/mypy/issues/14373.

adamjstewart:
Wow, that's annoying. The alternative would be to not set `python_version` for mypy so that it uses the current Python version by default. I'm not sure what setting `python_version` gets us.

tgamblin:
Setting `python_version` ensures that mypy is targeting the versions of python we support.  If you don't set it, it may allow things that only newer pythons support, so we lose version compatibility.

This is basically the same reason we set the python_version for `black` as well, just that the issues would be introduced through PRs by devs using newer pythons, and not through auto-formatting.

adamjstewart:
I'm still not convinced that we should set `python_version`. Black is a different story as we want to prevent forced formatting changes. We use vermin to ensure compatibility with older Python versions, what does mypy add that vermin doesn't have? I'm guessing that setting `python_version` will also disallow things like `"int | float | None"` (which vermin wouldn't catch?). This PR may fix the issue with numpy, but as we continue to allow more dependencies to be bootstrapped, what happens when those dependencies also start including features from newer Python versions?

tgamblin:
> I'm guessing that setting python_version will also disallow things like "int | float | None" (which vermin wouldn't catch?).

Yes.

> This PR may fix the issue with numpy, but as we continue to allow more dependencies to be bootstrapped, what happens when those dependencies also start including features from newer Python versions?

I don't think the mypy people have an answer to this, so we have to keep adding exceptions.  Right now, I believe most projects either don't bother shipping type info in `.pyi` files or they try to support back to 3.7 to avoid the issue. `numpy` dumped `3.7` and only supports `3.8` so it makes sense for them to have `.pyi` file that will break 3.7.

We should NOT have use the same type annotation version as our dependencies, or limit dependency versions based on their type info versions.  At the moment, though, I believe the only way to ensure *we* don't use newer type annotations than we should is to use `python_version`.

I don't really want to debate what to set `python_version` to here -- we could have that discussion in some other PR, and we could even bump it to 3.8 and make `spack style` require 3.8 for `mypy` checks.  I think this is ultimately what we should do (pick our own minimum type version).  At the moment, though, we don't do that and this will break anyone running `spack style` who also has numpy installed.

tgamblin:
If you look at other projects that do this, they set it to the lowest supported python version: https://github.com/pypa/build/commit/c2ed447f9fc24daef21f0214d871324a85ed6777

So probably we should set this to 3.6...

tgamblin:
@adamjstewart see also here: https://github.com/python/mypy/issues/9972#issuecomment-1368281725

adamjstewart:
> I don't really want to debate what to set `python_version` to here

Changing `python_version` makes the changes in this PR unnecessary. I'd rather make a final decision on this here instead of adding and removing these changes in a later PR.

I'm still of the mindset that we should not bother supporting `spack style` for older Python versions. So my vote would be to either increase `python_version` to at least "3.8" or drop it altogether and let it default to the version we use in CI. But I'll let you make the final decision.

If you really really want to support running `spack style` with older Python versions, then this PR is the correct fix.

tgamblin:
> If you really really want to support running spack style with older Python versions, then this PR is the correct fix.

I'd rather do this for now.

adamjstewart (dismissed):
Grudgingly approving...

tgamblin:
@adamjstewart I updated slightly --  I think you may like the latest formulation better -- at least with this one we only have to know that `numpy` is bad, not everything that could import it.  Can you reapprove?

adamjstewart (approved):

PR #22228, opened 2021-03-10
updated deps to get gtkplus to build

lee218llnl:

PR #34733, opened 2022-12-30
Make docbook-xml work across all entity versions

glennpj:
Packages that use docbook-xml may specify a specific entity version. When this is specified as a version constraint in the package recipe it will cause problems when using `unify = True` in a Spack environment, as there could be multiple versions of docbook-xml in the spec. In practice, any entity version should work with any other version and everything should work with the latest version. This PR maps all Spack docbook-xml entity versions to the docbook-xml version in the spec. Ideally, the version in the spec would be the latest version. With this PR, even if a package specifies an older entity version, it will map to the entity version (latest) in the spec. This means that there can be one docbook-xml version in a Spack environment spec and packages requesting older entity versions will still work.

To help facilitate this, docbook-xml version constraints for packages that have them have been removed. Those packages are dbus and gtk-doc.

### DIFF
--- a/var/spack/repos/builtin/packages/dbus/package.py
+++ b/var/spack/repos/builtin/packages/dbus/package.py
@@ -28,11 +28,12 @@ class Dbus(Package):
     version("1.8.2", sha256="5689f7411165adc953f37974e276a3028db94447c76e8dd92efe910c6d3bae08")
 
     depends_on("pkgconfig", type="build")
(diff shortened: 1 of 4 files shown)


lee218llnl:
my main interest in this is that gtkplus builds OK per https://github.com/spack/spack/pull/22228. My testing indicates that gtkplus still builds OK with this PR, so it has my approval

lee218llnl (approved):

PR #34735, opened 2022-12-31
libcroco: Get the `doc` variant working

glennpj:


var/spack/repos/builtin/packages/libcroco/package.py
```diff
@@ -15,18 +17,39 @@ class Libcroco(AutotoolsPackage):
...
+                    "--disable-gtk-doc",
+                    "--disable-gtk-doc-html",
+                    "--disable-gtk-doc-pdf",
+                ]
+            )
+
         # macOS ld does not support this flag
         # https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/libcroco.rb
-        return ["--disable-Bsymbolic"]
+        if sys.platform == "darwin":
```
alalazo: Why not using:
```suggestion
        if self.spec.satisfies("platform=darwin"):
```
? I don't think we'll ever cross compile across platforms, but it seems more correct to reference the spec rather than the current interpreter.

glennpj: Ah, some patterns are just ingrained. I will remove the `import sys` as well.

alalazo (approved):

PR #34736, opened 2022-12-31
nmap: update to depend on python2 for ndiff and zenmap; add 7.93

dpzmick:
This fixes the build on my linux machine. I'm still having an issue with macOS though; that's still under investigation


dpzmick:
Since this failed CI due to python2 not existing, how should the python2 dep be expressed?

PR #34737, opened 2022-12-31
new package: py-kb-python + dependencies

glennpj:
- py-loompy
- py-ngs-tools
- py-numpy-groupies

### DIFF
--- a/var/spack/repos/builtin/packages/py-kb-python/package.py
+++ b/var/spack/repos/builtin/packages/py-kb-python/package.py
@@ -20,7 +20,7 @@ class PyKbPython(PythonPackage):
 
     depends_on("py-anndata@0.6.22.post1:", type=("build", "run"))
     depends_on("py-h5py@2.10.0:", type=("build", "run"))
-    depends_on("py-jinja2@2.10.1:", type=("build", "run"))
+    depends_on("py-jinja2@2.10.2:", type=("build", "run"))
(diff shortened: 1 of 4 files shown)


var/spack/repos/builtin/packages/py-kb-python/package.py
```diff
@@ -0,0 +1,34 @@
...
+
+    version("0.27.3", sha256="dc98f6ceb4402d666b7e0d19be17c63d33e8b710a35cdc33de7c0f457122f43f")
+
+    depends_on("python@3.6:", type=("build", "run"))
+
+    depends_on("py-setuptools", type="build")
+
+    depends_on("py-anndata@0.6.22.post1:", type=("build", "run"))
+    depends_on("py-h5py@2.10.0:", type=("build", "run"))
+    depends_on("py-jinja2@2.10.1:", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-jinja2@2.10.2:", type=("build", "run"))
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-kb-python/package.py
glennpj: Good catch. I read the Jinja2>2.10.1 dependency and "filled-in" the ">=".

adamjstewart (approved):

PR #34757, opened 2022-12-31
cp2k: add new version and tweak settings

glennpj:
The @2022: versions use different settings than before when using CUDA.

### DIFF
--- a/var/spack/repos/builtin/packages/cp2k/package.py
+++ b/var/spack/repos/builtin/packages/cp2k/package.py
@@ -23,6 +23,7 @@ class Cp2k(MakefilePackage, CudaPackage):
 
     maintainers = ["dev-zero"]
 
+    version("2022.2", sha256="1a473dea512fe264bb45419f83de432d441f90404f829d89cbc3a03f723b8354")
     version("2022.1", sha256="2c34f1a7972973c62d471cd35856f444f11ab22f2ff930f6ead20f3454fd228b")
     version("9.1", sha256="fedb4c684a98ad857cd49b69a3ae51a73f85a9c36e9cb63e3b02320c74454ce6")
     version("8.2", sha256="2e24768720efed1a5a4a58e83e2aca502cd8b95544c21695eb0de71ed652f20a")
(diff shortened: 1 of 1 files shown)


alalazo (dismissed):
Waiting for @dev-zero to :heavy_check_mark: before merging

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -577,6 +587,10 @@ def edit(self, spec, prefix):
                     "-DOFFLOAD_TARGET=cuda",
                 ]
                 libs += ["-lcublas"]
+
+                if spec.satisfies("+cuda_fft"):
+                    cppflags += ["-D__PW_CUDA"]
+                    libs += ["-lcufft"]
```
dev-zero, line 593: you disabled the availability of the `cuda_fft` variant above I think

alalazo: From the perspective of Spack directives, this condition branch is entered only for version `9.*` when `+cuda_fft`. Not sure if this makes sense for `cp2k` though.

dev-zero: @alalazo exactly what I meant. In this PR there is also a `when="@:9",  # req in CP2K v2022+` added for the `cuda_fft` variant.

glennpj: I do not think `cuda_fft` is disabled above as it is enabled by default in `@2022:`. That is why I set the condition on the variant. However, looking at the cp2k docs again, I see that it _can_ be disabled in the 2022 version. If I am understanding correctly, versions up to 9 would use `D__PW_CUDA` to turn it on, and versions 2022 and above would use `D__NO_OFFLOAD_PW` to turn it off. So, I should redo the logic to reflect that. Does that sound correct @dev-zero?

glennpj: I added a commit to enable the `cuda_fft` variant for `@2022:` and adjusted the `if...` logic.

dev-zero: I think that looks right, yes.
Thanks a lot!

dev-zero (approved):

alalazo:
@spackbot run pipeline

alalazo (approved):

PR #34760, opened 2023-01-01
mesa: new versions

wdconinc:
No major changes like in the transition to 22.0.

Note: I've kept the preferred version within the same 22.1 series but moved to the latest bugfix release.


wdconinc:
Nvm, unknown option `gallium-xmvc`.

wdconinc:
> Nvm, unknown option `gallium-xmvc`.

Removed as of 22.3.0, per https://gitlab.freedesktop.org/mesa/mesa/-/commit/8cc766d8f7eac26b7c029a2fac1bdfdba4776c29. Should be easy to adapt in the `package.py`. Will probably take a stab today.

alalazo (approved):

PR #34759, opened 2023-01-01
config: modularize path replacements

tgamblin:
This is a WIP attempt to fix an error in tests that I see on M1. Specifically, I see the circular import error below in `test_autotools_gnuconfig_replacement_no_gnuconfig` on `develop`.

I don't understand why this fix breaks everything, but simply adding lambdas to the substitutions as in #34758 works.  So, we probably want to merge #34758, but I am really curious why this one blows up.

```console
> spack -d unit-test -k test_autotools_gnuconfig_replacement_no_gnuconfig

...

==> [2022-12-31-15:44:21.771459] Error: AttributeError:

The 'autotools-config-replacement' package cannot find an attribute while trying to build from sources. This might be due to a change in Spack's package format to support multiple build-systems for a single package. You can fix this by updating the build recipe, and you can also report the issue as a bug. More information at https://spack.readthedocs.io/en/latest/packaging_guide.html#installation-procedure

/Users/gamblin2/src/spack/lib/spack/spack/package_base.py:1332, in prefix:
       1330    @property
       1331    def prefix(self):
  >>   1332        """Get the prefix into which this package should be installed."""
       1333        return self.spec.prefix

Traceback (most recent call last):
  File "/Users/gamblin2/src/spack/lib/spack/spack/build_environment.py", line 1030, in _setup_pkg_and_run
    kwargs["env_modifications"] = setup_package(
                                  ^^^^^^^^^^^^^^
  File "/Users/gamblin2/src/spack/lib/spack/spack/build_environment.py", line 757, in setup_package
    set_module_variables_for_package(pkg)
  File "/Users/gamblin2/src/spack/lib/spack/spack/build_environment.py", line 596, in set_module_variables_for_package
    m.std_cmake_args = spack.build_systems.cmake.CMakeBuilder.std_args(pkg)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gamblin2/src/spack/lib/spack/spack/build_systems/cmake.py", line 241, in std_args
    define("CMAKE_INSTALL_PREFIX", pkg.prefix),
                                   ^^^^^^^^^^
  File "/Users/gamblin2/src/spack/lib/spack/spack/package_base.py", line 1333, in prefix
    return self.spec.prefix
           ^^^^^^^^^^^^^^^^
  File "/Users/gamblin2/src/spack/lib/spack/spack/spec.py", line 1710, in prefix
[...]

PR #34767, opened 2023-01-02
refactor: `llnl` packages should not import from `spack`

tgamblin:
Commits here should be kept intact, as each is a different refactor.

Code in `llnl.util` is designed to be free-standing and not rely on Spack. Over time, parts of `llnl.util` have started importing things from `spack`, which causes us to have to jump through hoops to avoid circular imports. For example, `llnl.util.filesystem` was importing `spack.util.executable`, but `llnl.util.filesystem` is used by `spack.paths` and lots of other things that are likely to lead to bootstrapping issues in `spack`.  We couldn't import `spack.util.environment` in `spack.util.path` without nesting it in a function, for similar reasons.

This PR completely separates  `llnl` from `spack`, and it moves some low-level stuff from `spack` into `llnl`.  I think it also gets us closer to factoring out some of the utility layers in Spack (e.g., environment modifications and executables) into their own packages that could live on PyPI, but this doesn't go that far.

- [x] Move `spack.util.string` to `llnl.util.string`. None of this relies on Spack, and it was being used in `llnl`.
- [x] Move windows path logic out of `spack.util.path` and into `llnl.util.path`. This was being imported from `spack` to `llnl` for windows support.
- [x] Move most of `spack.util.environment` -> `llnl.util.envmod`.  This is all of the code needed for `EnvironmentModifications`, which allow us to bundle many environment changes up and apply them generically for different shells. There were other routines in `spack.util.environment` that are used for things like dumping the build environment after installation, but these are more Spack-specific and not related to the `envmod` code.
- [x] Move `spack.util.executable` -> `llnl.util.executable`.  `spack.util.executable` has traditionally been more coupled to Spack, because the build environment notices when its `ProcessError` is raised and does some special handling to show build log context.  There wasn't much more than that, though, so I've decoupled the two and now `llnl.util` packages can import it without risking circular import issues with the `spack` package.

Moving `spack.util.executable` to `llnl.util.executable` revealed a bunch of packages that were using parts of `spack` without importing them (and getting away with it b/c so much was getting indirectly pulled into packages).  These have been fixed.


var/spack/repos/builtin/packages/perl/package.py
```diff
@@ -181,7 +181,7 @@ def patch(self):
 
     @classmethod
     def determine_version(cls, exe):
-        perl = spack.util.executable.Executable(exe)
+        perl = llnl.util.executable.Executable(exe)
```
adamjstewart: `Executable` is already imported right? Shouldn't we just use:
```suggestion
        perl = Executable(exe)
```

lib/spack/spack/package_test.py
```diff
@@ -5,7 +5,7 @@
 
 import os
 
-from spack.util.executable import Executable, which
+from llnl.util.executable import Executable, which
```
adamjstewart, line 8: Same here, these are already imported

adamjstewart (commented):
I guess I'm confused why we still separate llnl.util from spack at all. Ostensibly it's because we may want to move these out into external libraries some day like we did with archspec, but we haven't done that nor shown any interest in doing that in the last decade. Many of those things can be replaced with existing public libraries or builtin modules like pathlib, logging, and colorama. If there's anything I think is worth publishing as a separate library it would be the executable logic, but there's likely another library out there that does something similar already. Personally I don't see a lot of value in moving things around compared to trying to replace these things with existing libraries. If the only reason is to avoid circular imports, it feels like there has to be a better way of doing this, or at least detecting and preventing circular imports in CI.

var/spack/repos/builtin/packages/perl/package.py
tgamblin: fixed

lib/spack/spack/package_test.py
tgamblin: This one is in core.

HadrienG2 (approved):

haampie:
Is there a reason to do this right now? It's breaking API, so merging this means that people may need to update their recipes & current develop packages would be incompatible with 0.19.x in case people backport security updates to packages while otherwise keeping a stable version of spack.

tgamblin:
We can save it for later. I *thought* this would help with #34759, but it doesn’t.

I do think it’s ultimately a better way to factor the code,  but it will surface latent import issues in packages, and it also seems to confuse isort somehow (`isort` seems to want to delete `import spack` even when it is used).

tgamblin:
Making this draft for the time being.

alalazo:
> I do think it’s ultimately a better way to factor the code, but it will surface latent import issues in packages, and it also seems to confuse isort somehow (isort seems to want to delete import spack even when it is used).

For what is worth, I agree with this. In my opinion we should have a "layered" structure where `spack.cmd` is at the top, then `spack` core packages and `llnl` is at the bottom. I also agree with @adamjstewart that it would be worth trying to get rid of some low-level code if we have a good 3rd party substitute for it. 

If we want to move code to an external package, can we do it incrementally? If we find a good "theme" (and a good name) for a new Python package, I think both `Executable` and our locks are possible first candidates to be extracted.

adamjstewart:
I'm not opposed to having a layered structure, but I would suggest `spack.cmd`, then `spack` core, then `spack.util`. There is currently stuff in `spack.util` that imports `spack` core, but we could move stuff around if needed. It just feels weird that LLNL is hardcoded in the code now that we've moved from https://github.com/llnl/spack to https://github.com/spack/spack.

tgamblin:
> It just feels weird that LLNL is hardcoded in the code

Yeah definitely. This isn’t about moving things into the `llnl` namespace specifically. FWIW that’s always just been part of `spack`. It’s just the broadest name I could envision for the generic layer of `spack` in 2013 😆 .

We can rename it if you want, or we could move everything in it to externals.

I think it’s useful for whatever it is to live outside the `spack` namespace, though. If it’s under `spack`, we can’t import it without pulling in part of `spack`, which makes it harder to separate (cf. all the imports I had to add to packages here). Id like to get `mypy` checking for undefined symbols in packages to avoid this (it can analyze through * imports) but I haven’t had time to set that up.

adamjstewart:
> Id like to get mypy checking for undefined symbols in packages to avoid this (**it can analyze through * imports**) but I haven’t had time to set that up.

Tell me more about this...

vvolkl (approved):

trws:
> > Id like to get mypy checking for undefined symbols in packages to avoid this (**it can analyze through * imports**) but I haven’t had time to set that up.
> 
> Tell me more about this...

We actually get some of this already, but currently it only applies to functions with typed signatures because of our configuration.  If we turn that flag on, we get a *lot* better coverage and we could turn off some of the much less useful, and much louder, parts of flake8's pylint checks in favor of mypy. Unfortunately, doing it right this instant would dump a ton of errors on us.

alalazo:
Moving to the next release, due to the number of conflicts

tgamblin:
Superseded by #40262

PR #34768, opened 2023-01-02
fix: python tix detection blocks reuse

haampie:
Now that the `tix` variant is conditional, it should also be detected
conditionally, otherwise the spec violates the when condition and
is not considered during concretization

Reported by @giordano

### DIFF
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -298,11 +298,12 @@ class Python(Package):
             variants += "~pyexpat"
 
         # Some variant names do not match module names
-        try:
-            python("-c", "import tkinter.tix", error=os.devnull)
-            variants += "+tix"
-        except ProcessError:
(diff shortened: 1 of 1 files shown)


adamjstewart (dismissed):
Good catch!

haampie:
@spackbot run pipeline

haampie:
CI fix is in #34775

adamjstewart (dismissed):

haampie:
@adamjstewart have you seen 

```
WARNING: Download from https://storage.googleapis.com/mirror.tensorflow.org/github.com/tensorflow/runtime/archive/be42c742dfda390ea69f506d89a5dbf9049421d8.tar.gz failed: class java.io.FileNotFoundException GET returned 404 Not Found
```

in the past?

Ah uh, it's really `gcc: internal compiler error: Killed (program cc1plus)` nevermind.

adamjstewart (approved):

PR #34772, opened 2023-01-03
move default values for rocm/cuda arch into packags:all:variants

haampie:
- `amdgpu_target` is sticky
- cabana needs kokkos
- `amdgpu_target` was not set for kokkos in that case, so it is none
- that's a conflict for kokkos
- the concretizer works around it by using cabana@master (which doesn't have a dep on kokkos)

This PR sets a uniform default for amdgpu_target in package preferences so that cabana@master is no longer concretized to

Ultimately this is an issue with the cabana package, but that can be addressed separately.


haampie:
@spackbot run pipeline

haampie:
@junghans @streeve @sslattery can you please fix the cabana package s.t. the master version doesn't have missing deps?

alalazo (approved):

alalazo:
@eugeneswalker FYI Let me know if we have to apply any change with respect to what is done here in following PRs. Currently `develop` is broken, so I'm going to merge this one as is.

PR #34779, opened 2023-01-03
Move OS specific code into its own folder

alalazo:
Move all the `llnl.*` code that is either Windows or Unix specific into separate folders. The API for both platforms is the same and the correct code is imported based on "sys.platform". 

Modifications:
- [x] Move platform specific code into separate folders
- [x] Import in `llnl.util.filesystem` and `llnl.util.symlink` the correct implementation based on the current `sys.platform`


alalazo:
This is an attempt to avoid having `is_windows` appearing into multiple places within a package. If the approach seems better than the current state it can be extended to other parts in `spack.*` code. Any comment on the layout used in this PR is welcome.

lib/spack/llnl/util/filesystem.py
```diff
@@ -84,26 +74,49 @@
 ]
 
 
-def getuid():
-    if is_windows:
-        import ctypes
+if sys.platform == "win32":
```
alalazo, line 77: The assumption here is that we'll import different implementations of the same functions in both conditional branches.

alalazo:
@johnwparent @BetsyMcPhail @tgamblin @scheibelp fyi let me know if this approach is worth polishing / applying to other parts of the code.

lib/spack/llnl/util/_windows/__init__.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
```
tldahlgren, line 1:
```suggestion
# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```

lib/spack/llnl/util/_windows/symlink.py
```diff
@@ -0,0 +1,104 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
```
tldahlgren, line 1:
```suggestion
# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```

tldahlgren (commented):
Real quick pass I noticed there are a couple of copyright end years off.

lib/spack/llnl/util/_windows/symlink.py
alalazo: Great catch, I wonder where I copied those from and more importantly why CI was passing :thinking:

scheibelp:
My general preference is that if a method can abstract Windows-specific details, then having `is_windows` checks inside of it is good and can avoid code duplication. I should add specifics here but wanted to voice that general concern.

alalazo:
Since this was just a proof of concept, and is now outdated, close it.

Issue #14448, opened 2020-01-09
Semantics for Variant Merging


citibeth:
@tgamblin @scheibelp @alalazo 

This is in response to #10466, but I though it could use its own Issue:

> I would ask for unit tests, but before doing this I think we should agree on how something like:
> 
> ```yaml
> packages:
>   all:
>     variants: ['~shared']
>   opencv:
>     variants: ['+shared']
> ```
> 
> is supposed to behave, which as far as I understand is the motivation behind your fixme?

The Problem
=========

I think the root problem here stems from conflicting/vague views of what on/off variants are supposed to be/mean.  There are two ways they are currently used in Spack recipes:

1. The most common is as a feature to be turned on/off (I'll call this a "feature on/off variant").  This usage is widespread; for example, in `python/package.py`:
   ```
   variant('tk', default=False, description='Provide support for Tkinter')
   ```
   Feature variants are often used to avoid bringing in dependencies for features that are not needed.  This is true for `python+tk`, as one can see later in the recipe:
   ```
   depends_on("tk", when="+tk")
   ```

2. Sometimes on/off variants are also currently used to select between mutually exclusive alternatives (I'll call this an "alternative on/off variant").  For example, the package `symengine` apparently needs a multi-precision library, and it can use either `boost` or `gmp` for that purpose.  It does so as follows:
   ```
   variant('boostmp',      default=False,
            description='Compile with Boost multi-precision integer library')

    depends_on('boost',    when='+boostmp')
    depends_on('gmp',      when='~boostmp')
    depends_on('mpc',      when='+mpc~boostmp')
    depends_on('mpfr',     when='+mpfr~boostmp')
   ````

IMHO, usage (1) is correct and usage (2) is incorrect.  Spack didn't previously have multi-valued variants, but now it does.  Usage (2) should be recoded to use them; for example, in this way:

```
   variant('mplib',      default='gmp',
            values=('boost', 'gmp'), multi=False,
            description='Multi-precision integer library to use')

    depends_on('boost',    when='mplib=boost')
    depends_on('gmp',      when='mplib=gmp')
    depends_on('mpc',      when='+mpc mplib=gmp')
    depends_on('mpfr',     when='+mpfr mplib=gmp')
```

Ramifications
==========

The mixture of feature on/off and alternative on/off variants causes us problems.

Conflicting Concretizations
----------------------------

[...]

PR #34781, opened 2023-01-03
intersect "all" and "pkg" variant prefs

haampie:
The following

```yaml
packages:
  all:
    variants: +x +y
  pkg:
    variants: +y +z
```

should really mean `+x +y +z` as a preference for `pkg`. Currently it means `+y +z`, which is confusing.

This PR takes the intersection `all.constrain(pkg)` for variant preferences.

The case of

```yaml
packages:
  all:
    variants: +x
  pkg:
    variants: ~x
```

errors about ambiguity and informs the user through an error message that

```yaml
packages:
  all:
    variants: +x
  pkg:
    require: ~x
```

resolves that ambiguity.

### DIFF
--- a/lib/spack/spack/package_prefs.py
+++ b/lib/spack/spack/package_prefs.py
@@ -5,8 +5,10 @@
 import stat
 import warnings
 
+import spack.config
 import spack.error
 import spack.repo
+import spack.spec
 from spack.config import ConfigError
 from spack.util.path import canonicalize_path
 from spack.version import VersionList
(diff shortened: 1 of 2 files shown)


haampie:
On second thought, I don't really like this anymore either :(

The reason being, with  `all:variants:cuda_arch=60` and `pkg:variants:cuda_arch=70`, the logic of this PR would turn the effective default into `pkg:variants:cuda_arch=60,70` instead of erroring like in `~mpi/+mpi`. And for `libs=shared` / `libs=static` we would get `libs=shared,static`. It's not wrong, it's just somewhat unexpected too.

sethrj:
May close https://github.com/spack/spack/issues/14448

alalazo:
Just my 2 cents, but the current way of doing things make sense, if you consider `all` as a set of default values for other entries. Here:
```yaml
packages:
  all:
    variants: +x +y
  pkg:
    variants: +y +z
```
`pkg` will have defaults for `+y +z`, all the other packages `+x +y` - since they lack a `variants` attribute and thus fall back to the default value under `all`.

haampie:
Okay, I'll close for now.

sethrj:
Actually I think the arch 60/70 is the right approach. Suppose that the site's main system has 60 but they're testing out a new build with 70 hardware. Now the build supports both??

PR #34783, opened 2023-01-03
jube: Add versions 2.5.0 and 2.5.1

picsel2:


adamjstewart (changes requested):
It seems like pyyaml is always required, I wouldn't bother adding a variant

picsel2:
The tutorial says that it is optional: https://apps.fz-juelich.de/jsc/jube/jube2/docu/tutorial.html

The package did not depend on pyyaml before and I tried to be conservative with "new" dependencies.

adamjstewart:
The tutorial may say that it's optional but [the build system says it's required](https://github.com/FZJ-JSC/JUBE/blob/REL-2.5.1/setup.py#L31) and always has been. If it isn't installed, our import tests will fail, and some build backends may fail during installation.

picsel2:
Indeed. Thanks for the clarification!

I've changed it to always depend on pyyaml.

adamjstewart (approved):

PR #34784, opened 2023-01-03
concretize with configurable # processes

haampie:
Peak memory [can be 1.2GB](https://gist.github.com/haampie/6893cb08538aa3f94db5b63355c233a0) for some specs in the e4s stack, so either we have to bump the memory request, or reduce the number of processes.

They were rather arbitrary anyways, and not realistic (we were using 16 processes, so ~16GB peak, the request was 4 cpus and 8GB memory)

### DIFF
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -1399,7 +1399,9 @@ class Environment(object):
 
         # Solve the environment in parallel on Linux
         start = time.time()
-        max_processes = min(len(arguments), 16)  # Number of specs  # Cap on 16 cores
+
+        # Number of processes, capped to config:build_jobs & # specs.
+        max_processes = min(len(arguments), spack.config.get("config:build_jobs", 16))
 
         # TODO: revisit this print as soon as darwin is parallel too
         msg = "Starting concretization"
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -40,7 +40,7 @@ default:
     - spack --version
     - cd share/spack/gitlab/cloud_pipelines/stacks/${SPACK_CI_STACK_NAME}
     - spack env activate --without-view .
-    - spack config add config:build_jobs:12 # concretize <= 12 processes
+    - spack config add "config:build_jobs:${SPACK_CONCRETIZATION_PROCS}"
     - spack ci generate --check-index-only
       --buildcache-destination "${SPACK_BUILDCACHE_DESTINATION}"
       --artifacts-root "${CI_PROJECT_DIR}/jobs_scratch_dir"
@@ -52,8 +52,11 @@ default:
       - "${CI_PROJECT_DIR}/jobs_scratch_dir"
   tags: ["spack", "aws", "public", "medium", "x86_64"]
   variables:
-    KUBERNETES_CPU_REQUEST: 12000m
-    KUBERNETES_MEMORY_REQUEST: 12G
+    # This should be overridden for environments with unify:false
+    # Observed peak RSS is ~1GB / process.
+    KUBERNETES_CPU_REQUEST: 1000m
+    KUBERNETES_MEMORY_REQUEST: 1G
+    SPACK_CONCRETIZATION_PROCS: 1
   interruptible: true
   retry:
     max: 2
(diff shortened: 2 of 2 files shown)


lib/spack/spack/environment/environment.py
```diff
@@ -1399,7 +1399,9 @@ def _concretize_separately(self, tests=False):
 
         # Solve the environment in parallel on Linux
         start = time.time()
-        max_processes = min(len(arguments), 16)  # Number of specs  # Cap on 16 cores
+
+        # Number of processes, capped to config:build_jobs & # specs.
```
scottwittenburg, line 1403:
```suggestion
        # Number of processes, capped to config:build_jobs or 16.
```

scottwittenburg (commented):
Thanks @haampie I think it's great to be able to specify the number of concretization jobs this way.  I think most of the time (of at least a lot of the time) we didn't hit OOM on these jobs, but enjoyed the faster concretization.  Should we consider using 16 cores (and requesting 16GB of RAM) instead of 8, to keep that speed-up?

lib/spack/spack/environment/environment.py
haampie: It's actually bounded by `# specs` too (that is `len(arguments)`). I think the literal 16 doesn't really need a comment ;p

haampie:
> ERROR: Preparation failed: couldn't prepare overwrites: invalid build requests specified: the resource "CPURequest" requested "16000m" is higher than limit allowed "12"

So, apparently we do have a limit ;) confusingly this is a limit set in the Gitlab runner config for the cpu request variable itself, it's not the kubernetes cpu limit.

Edit: this config may only apply to the e4s environment; I haven't checked the others really.

lib/spack/spack/environment/environment.py
scottwittenburg: Ah, got it, thanks.

scottwittenburg:
> confusingly this is a limit set in the Gitlab runner config for the cpu request variable itself

Taking a look, that may be coming from [here](https://github.com/spack/spack-infrastructure/blob/a2b8760a021ab8dfb2d0108182f8576fe8be4a0c/k8s/runners/public/x86_64/v3/release.yaml#L40), in which case we can change it.  But we can see how it goes with only 12 at first, now that the memory should be in line with number of processes.

scottwittenburg (dismissed):

haampie:
@scottwittenburg can you explain why https://gitlab.spack.io/spack/spack/-/jobs/5092361 fails?

At this point it's neither CPU (1 single-threaded process) nor memory (8GB free)...

https://gitlab.spack.io/spack/spack/-/jobs/5092361 and https://gitlab.spack.io/spack/spack/-/jobs/5092360 were maybe put on the same node? Is that node simply broken

scottwittenburg:
> @scottwittenburg can you explain why https://gitlab.spack.io/spack/spack/-/jobs/5092361 fails?

When I've seen that message in the past, I've wondered if it was spot instance termination, but I think @zackgalbreath has seen evidence that some of those *are* OOM kills.  But I'm currently trying to find some concrete evidence of why those jobs might have failed.  

However, I'm a bit suspicious that those were both jobs that used the new default requests of 1 cpu and 2GB mem.  Can you explain why you say those default requests should only be overridden when `unify: false`?

haampie:
Ah, I didn't realize we could be using those types of nodes. Does that leave a trace somewhere?

Clingo runs single-threaded since 2d45a9d6176eda9375515b77699a17cfd5221111 (actually there was no observed speedup generally, and performance was typically strictly worse for > 2 threads)

The only parallelism comes from `unify: false`, since that makes root specs independent (and therefore trivially parallel) 

For `unify: true` and `unify: when_possible` the environment is treated as a whole and there is no parallelism.

scottwittenburg:
> Ah, I didn't realize we could be using those types of nodes. Does that leave a trace somewhere?

I've not been able to find such a trace yet.  We recently switched from cluster autoscaler to karpenter, and with that change it seems we lost the very limited ability we had to get this kind of information.  We're currently working on a way to have more information about spot evictions.

Thanks for clarifying the parallelism points, that makes sense.  Based on that explanation, I guess it seems less likely to be a memory problem.  Still, we're missing a smoking gun.

I notice those failed jobs weren't retried, even though they're set to retry the max number of times (twice) in the events of `runner_system_failure` and `stuck_or_timeout_failure`.  The full list of reasons can be found [here](https://docs.gitlab.com/ee/ci/yaml/#retry).  I don't know how gitlab classifies that particular error unfortunately, or we could just add that as a reason for retry.  

One option is just to `always` retry pipeline generation jobs twice, like we do for build jobs, though if we do, we know we waste some cycles retrying jobs that are sure to fail.  We could do this for a little while though, while we try to come up with ways to: a) distinguish spot evictions from other types of jobs, and/or b) have better detection of OOM kills.

haampie:
lol, i think the general rule of thumb is that gitlab has a multiple years old open issue for every feature that'd be useful to have: https://gitlab.com/gitlab-org/gitlab/-/issues/262674

PR #34789, opened 2023-01-04
gitlab ci: fix tags for pipeline generation jobs

scottwittenburg:
Gitlab does not merge lists when a job extends two other definitions that include the same list (e.g. tags).  Also, it merges dictionaries as long as the keys are distinct, but just takes the last mentioned value when there are key collisions.

This change makes sure that when different tags are needed by a pipeline, the ones we want are actually provided.  It also changes the example stack to better follow this pattern so we do not lead developers astray in the future.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -116,7 +116,13 @@ protected-publish:
 # First add a new spack.yml defining the pipeline to run in
 # share/spack/gitlab/cloud_pipelines/stacks/my-super-cool-stack/spack.yaml
 #
-# Then add the following entries at the bottom of this file:
+# Then add the following entries at the bottom of this file.
+#
+# Note that when extending other jobs, gitlab does not merge lists (nor
+# does it merge dictionary values in the case of key conflicts).  So lists
+# and duplicated dictionary keys are simply overridden.  For this reason,
+# you should inlclude your custom definitions at the end of the of the
+# extends list.
 #
 ########################################
 # My Super Cool Pipeline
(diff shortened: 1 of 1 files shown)


scottwittenburg:
@zackgalbreath I think this will fix the problem of the missing `x86_64_v4` tags on pipeline generation jobs.

haampie:
Ah, I also noticed this in https://github.com/spack/spack/pull/34784 before I found this PR :upside_down_face: variables are still merged, right?

zackgalbreath (dismissed):

scottwittenburg:
> Ah, I also noticed this in #34784 before I found this PR upside_down_face variables are still merged, right?

Variables are merged when keys don't conflict.  If there are conflicting keys, the behavior is to override with the last one (last in the `extends` list, or the one in the job itself).

scottwittenburg:
pipeline generation jobs failed with: `image pull failed: pull QPS exceeded`

scottwittenburg:
@spackbot run pipeline

scottwittenburg:
@zackgalbreath I checked all the stacks that are applying `x86_64_v4` tags and they all specify only `x86_64_v3` (we're not building binaries for the 1%, as somebody famously said), so I was thinking it would be good to change that tagging in this PR.  Because don't `v4` instances cost more?

The problem I see is that karpenter is not currently provisioning `v3` instances.  So if I change it in this PR, what will karpenter do with the `x86_64_v3` tags when my pipeline runs?  On the other hand, if we turn those helm releases back on to test this PR, wont' that break pipelines for everyone else's PRs and `develop`?

scottwittenburg:
@zackgalbreath As we discussed a bit ago, let's just merge this as-is, then turn back on the v2/v3 provisioning capability, then I'll make another PR to switch the v4 tags to v3.

I mean, let's just merge this after retrying the macos unit-tests however many times it takes.  And I need another approval, sorry.

zackgalbreath (approved):

PR #34798, opened 2023-01-04
numactl: remove bash

haampie:
It doesn't require bash on any registered version, and the file is
executable and has a /bin/sh shebang.


tldahlgren (approved):
LGTM

PR #34675, opened 2022-12-23
Update py-meson-python (0.11.0, 0.12.0) and meson (0.64.1, 1.0.0)

rgommers:
Also add myself as maintainer for `py-meson-python`.

PR #34799, opened 2023-01-04
py-scipy: add v1.10.0

adamjstewart:

https://github.com/scipy/scipy/releases/tag/v1.10.0


var/spack/repos/builtin/packages/py-meson-python/package.py
```diff
@@ -14,17 +14,21 @@ class PyMesonPython(PythonPackage):
 
     maintainers = ["eli-schwartz", "adamjstewart"]
 
+    version("0.12.0", sha256="8cb159a8093a2e73cfa897f8092ec93b74e3842f94dff7fde381c6fe0e0b064d")
+    version("0.11.0", sha256="110258837c2ffe762f5f855c7ea5385f1edd44074e93a0f317ffefc7aab42b09")
```
rgommers: I'm a little confused why this is showing up here. It's a subset of what was merged in gh-34675

adamjstewart: Oh, my branch is out of date, let me rebase

var/spack/repos/builtin/packages/py-scipy/package.py
```diff
@@ -55,33 +56,38 @@ class PyScipy(PythonPackage):
     # https://github.com/mesonbuild/meson-python/pull/167
     depends_on("py-build", when="@1.9:", type="build")
 
-    depends_on("py-meson-python@0.9:", when="@1.9.2:", type="build")
+    depends_on("py-meson-python@0.11", when="@1.10:", type="build")
```
rgommers, line 59: Note that 0.12 is available now and I've just bumped the constraint in scipy's 1.10.x branch to allow using it. So you may want to add `:0.12` here already

rgommers (dismissed):
LGTM (assuming Gitlab CI comes back green). Just one minor comment, feel free to ignore if that's easier.

var/spack/repos/builtin/packages/py-scipy/package.py
adamjstewart: I'd like to keep the deps as close as possible to what is documented in pyproject.toml. We're currently disabling dep checks for pip but when we move to build/installer I think it will complain if the versions don't match. So once 1.10.1 is released I'll update this.

var/spack/repos/builtin/packages/py-scipy/package.py
```diff
@@ -111,9 +111,17 @@ class PyScipy(PythonPackage):
...
+
+    # https://docs.scipy.org/doc/scipy/dev/toolchain.html#other-libraries
+    depends_on("lapack@3.7.1:", when="@1.9:")
     depends_on("lapack@3.4.1:", when="@1.2:")
 
+    # https://docs.scipy.org/doc/scipy/dev/toolchain.html#compilers
+    conflicts("%gcc@:7", when="@1.10:")
+    conflicts("%gcc@:4.7", when="@:1.9")
+    conflicts("%apple-clang@:9", when="@1.10:")
+    conflicts("%msvc@:18", when="@1.10:")
```
rgommers: Maybe `mscv@:19.10`? The `meson.build` in 1.10.x and main contains:
```meson
elif cc.get_id() == 'msvc'
  if not cc.version().version_compare('>=19.20')
    error('SciPy requires at least vc142 (default with Visual Studio 2019) ' + \
          'when building with MSVC')
  endif
endif
```

Good catch on all these!

rgommers (approved):

haampie (approved):

PR #34801, opened 2023-01-04
Add new versions of Reframe

carsonwoods:
This is a very small PR that adds the now-released versions of Reframe 4.0.0 and 4.0.1. It also deselects 3.12.0 as the preferred version as there is now no pre-release version that is newer than the latest released version. 

If someone thinks 3.12.0 should still be the preferred version, I can change that back before this is merged.


vkarak (approved):
Thanks for submitting!

haampie (approved):

PR #34804, opened 2023-01-04
Remove **kwargs from function signatures in llnl.util.filesystem

alalazo:
Since we dropped support for Python 2.7, we can embrace using keyword only arguments for many functions in Spack that use `**kwargs` in the function signature. Here I changed all the functions of that kind in `llnl.util.filesystem` to have an explicit list of arguments. 

There were a couple of bugs lurking in the code related to typo-like errors when retrieving from kwargs. Those have been fixed as well.

Modifications:
- [x] Modified functions in `llnl.util.filesystem` using `**kwargs` in function signatures to have explicit parameters
- [x] Added type hints to those functions
- [x] Fixed typo-like bugs found during the refactoring

### DIFF
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -224,7 +224,7 @@ def filter_file(
     ignore_absent: bool = False,
     start_at: Optional[str] = None,
     stop_at: Optional[str] = None,
-):
+) -> None:
     r"""Like sed, but uses python regular expressions.
 
     Filters every line of each file through regex and replaces the file
@@ -344,8 +344,26 @@ class FileFilter(object):
     def __init__(self, *filenames):
         self.filenames = filenames
 
-    def filter(self, regex, repl, **kwargs):
-        return filter_file(regex, repl, *self.filenames, **kwargs)
+    def filter(
+        self,
+        regex: str,
+        repl: Union[str, Callable[[re.Match], str]],
+        string: bool = False,
+        backup: bool = False,
+        ignore_absent: bool = False,
+        start_at: Optional[str] = None,
+        stop_at: Optional[str] = None,
+    ) -> None:
+        return filter_file(
+            regex,
+            repl,
+            *self.filenames,
+            string=string,
+            backup=backup,
+            ignore_absent=ignore_absent,
+            start_at=start_at,
+            stop_at=stop_at,
+        )
 
 
 def change_sed_delimiter(old_delim, new_delim, *filenames):
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -671,7 +671,13 @@ def resolve_link_target_relative_to_the_link(link):
 
 
 @system_path_filter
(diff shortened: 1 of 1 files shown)


lib/spack/llnl/util/filesystem.py
```diff
@@ -710,7 +735,7 @@ def copy_tree(src, dest, symlinks=True, ignore=None, _permissions=False):
             abs_src,
             abs_dest,
             order="pre",
-            follow_symlinks=not symlinks,
+            follow_links=not symlinks,
```
alalazo, line 738: `follow_symlinks` does not exist in `traverse_tree`

lib/spack/llnl/util/filesystem.py
```diff
@@ -1150,16 +1197,11 @@ def traverse_tree(source_root, dest_root, rel_path="", **kwargs):
             ``src`` that do not exit in ``dest``. Default is True
         follow_links (bool): Whether to descend into symlinks in ``src``
     """
-    follow_nonexisting = kwargs.get("follow_nonexisting", True)
-    follow_links = kwargs.get("follow_link", False)
```
alalazo, line 1154: We document `follow_links` but we get `follow_link` (missing `s`) from the dictionary

lib/spack/llnl/util/filesystem.py
```diff
@@ -884,7 +898,10 @@ def mkdirp(*paths, **kwargs):
                         os.chmod(intermediate_path, intermediate_mode)
                     if intermediate_group is not None:
                         chgrp_if_not_world_writable(intermediate_path, intermediate_group)
-                        os.chmod(intermediate_path, intermediate_mode)  # reset sticky bit after
+                        if intermediate_mode is not None:
```
alalazo, line 899: Without this check `mypy` was (correctly) warning about possibly calling `os.chmod` with `intermediate_mode` set to `None`

alalazo (commented):
Leaving, for reviewers convenience, a couple of notes on the bug I found.

lib/spack/llnl/util/filesystem.py
alalazo: @becker33 fyi, so you can double check if this the correct fix here

lib/spack/llnl/util/filesystem.py
```diff
@@ -893,6 +908,27 @@ def mkdirp(*paths, **kwargs):
...
+
+    Args:
+        path: directory to be created
+    """
+    # detect missing intermediate folders
+    intermediate_folders = []
+    last_parent = ""
+    intermediate_path = os.path.dirname(path)
+    while intermediate_path:
+        if os.path.exists(intermediate_path):
```
haampie: This can just be `lexists`, if the deepest thing is a dangling symlink, makedirs will error anyways with ENOENT (and we will handle it the same above under `not os.path.isdir` in the error handling, whether it's a file or a dangling symlink)

alalazo: Done

haampie (approved):

PR #34805, opened 2023-01-04
Setting LIBDIR to lib in rocm-smi-lib across all platform

renjithravindrankannath:
Setting LIBDIR to lib in rocm-smi-lib across all platform


var/spack/repos/builtin/packages/rocm-smi-lib/package.py
```diff
@@ -116,7 +116,11 @@ class RocmSmiLib(CMakePackage):
     patch("disable_pdf_generation_with_doxygen_and_latex.patch", when="@4.5.2:")
 
     def cmake_args(self):
-        return [self.define_from_variant("BUILD_SHARED_LIBS", "shared")]
+        args = [
+            self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+            self.define("CMAKE_INSTALL_LIBDIR", "lib"),
```
tldahlgren: Do you want `"lib"` or `self.prefix.lib`?

tldahlgren (commented):
Quick question about the lib setting.

var/spack/repos/builtin/packages/rocm-smi-lib/package.py
renjithravindrankannath: @tldahlgren , Update as recommended above.
Thank you..

tldahlgren (approved):

tldahlgren:
Will defer merge decision to give @srekolam a chance to comment.

var/spack/repos/builtin/packages/rocm-smi-lib/package.py
haampie: You do want `"lib"`, see [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html). 

>  It should typically be a path relative to the installation prefix so that it can be converted to an absolute path in a relocatable way

But to be honest: why do you want this? The reason you end up with `lib64` is because you use `GNUInstallDirs` in that project. So it's expected that the libs to into `lib64` on CentOS.

Spack typically doesn't care about lib vs lib64, so there's no need to change it?

renjithravindrankannath: Hi @haampie, 
This change was indented to match with the RPM packaging distros of ROCm releases.
Library path as per the RMP distro is `/opt/rocm-<ver>/rocm_smi/lib/` in CentOS.
I added this change to align with this.

srekolam (approved):

srekolam:
This looks okay to me

tldahlgren:
@spackbot run pipeline

renjithravindrankannath:
Thank you @alalazo.

PR #34807, opened 2023-01-04
spack.dependency: add type hints

alalazo:
Extracted from a larger branch that modifies edges in DAGs to keep track of virtual dependencies. This helps avoiding type-matching issues while we refactor the code.

Modifications:
- [x] Add type hints to code in `spack.dependency`

### DIFF
--- a/lib/spack/spack/dependency.py
+++ b/lib/spack/spack/dependency.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 """Data structures that represent Spack's dependency relationships."""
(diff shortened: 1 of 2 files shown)


lib/spack/spack/dependency.py
```diff
@@ -85,13 +89,18 @@ class Dependency(object):
 
     """
 
-    def __init__(self, pkg, spec, type=default_deptype):
+    def __init__(
+        self,
+        pkg: "spack.package_base.PackageBase",
```
haampie, line 94: sigh, python :laughing:

haampie (approved):

PR #34821, opened 2023-01-05
Add virtual information on DAG edges

### LINKED ISSUE #34886: Annotate virtual dependencies on DAG edges
### Summary

As a user I want virtual dependencies to be annotated on DAG edges, so I can distinguish between software that depends on a specific virtual provider (without depending on the virtual) and software that depends on the virtual.

### Rationale

This is information that we currently lose after concretization, and that should be retained. It is needed to be able to cherry-pick virtual dependencies.

### Description

In memory and JSON representation of a spec should be extended to incorporate information on virtual dependencies:
1. For the in memory representation, we need to extend the `DependencySpec` class
2. For the JSON representation we need to add some new piece of information under the `"dependencies"` section of a node

A possible encoding for the JSON specfile could be an extra, optional, attribute under the `"dependencies"` section of a node:
```json
    "dependencies": [
          {
            "name": "cmake",
            "hash": "7qhwgdbqwfpo6ataty7gqbkveciw3udn",
            "type": [
              "build"
            ]
          },
          {
            "name": "pkgconf",
            "hash": "i4avrindvhcamhurzbfdaggbj2zgsrrh",
            "type": [
              "run"
            ],
            "virtuals": [
              "pkgconfig"
            ]
          },
          {
            "name": "zlib",
            "hash": "nizxi5u5bbrzhzwfy2qb7hatlhuswlrz",
            "type": [
              "build",
              "link"
            ]
          }
        ],
```

### Additional information

1. If we bump specfile version correctly, we can reconstruct virtual edges in memory from old specfiles when we read them. 
2. We could also update the specfile, being careful about not changing an existing hash.
3. Selecting a virtual dependency from the DAG can also be encoded in terms of edge attributes.
4. The specfile representation can also be used to implement "use variants" as sketched in https://github.com/spack/seps/pull/2.

### General information

alalazo:
depends on #34807 
closes #35258
fixes #34886

:warning: **Bumps format of specfile, and thus of lockfile and database** :warning: 

Currently virtual dependencies are used only during concretization to select a provider. No annotation is kept on edges to reflect that a leaf node is providing a virtual dependency. This PR adds virtual information on edges, so we can keep track of which spec is providing what virtual both in memory and in the JSON / YAML representation.

An example entry in JSON format is:
```json
        "dependencies": [
          {
            "name": "cmake",
            "hash": "iqpoju67adln3yqzvqzccrtkjmpc666m",
            "parameters": {
              "deptypes": [
                "build"
              ],
              "virtuals": []
            }
          },
          {
            "name": "openmpi",
            "hash": "p23v372ij7bomrarjrhcv4uw6ieahjt3",
            "parameters": {
              "deptypes": [
                "build",
                "link"
              ],
              "virtuals": [
                "mpi"
              ]
            }
          },
          {
            "name": "pkgconf",
            "hash": "i4avrindvhcamhurzbfdaggbj2zgsrrh",
            "parameters": {
              "deptypes": [
                "run"
              ],
              "virtuals": [
                "pkgconfig"
              ]
            }
          },
          {
            "name": "zlib",
            "hash": "nizxi5u5bbrzhzwfy2qb7hatlhuswlrz",
            "parameters": {
              "deptypes": [
                "build",
                "link"
              ],
              "virtuals": []
            }
          }
        ],
```

Modifications:
- [x] Add a `virtuals` attribute to `DependencySpec`
- [x] Modify `Spec.to_node_dict` to output information on virtuals in the `"dependencies" section
- [x] Add unit tests (round-trip specs / check annotation)
- [x] Modified `DependencySpec` to hold an arbitrary number of attributes (98b98069df9357c9b6c97e8d8b98dc48fbba7d9a)
- [x] Separated loading functions from dict according to specfile version, add unit-tests to ensure we can read old specfiles (f19daa800998a24f9ff32770d85fe4864e65fe08)
- [x] Reconstruct virtuals from old specfile formats (0b2db308fe19cfb74d705678ea17518cee71550e)


alalazo:
@tgamblin Do we want to reconstruct virtuals on edges when reading "old" specs in JSON format? I think that is possible in principle, since specs have a unique provider.

haampie:
We could also consider printing this info in `spack spec`:

```
pkg
    ^openblas (provides lapack, blas)
```

lib/spack/spack/cray_manifest.py
```diff
@@ -162,7 +162,8 @@ def entries_to_specs(entries):
                     continue
                 parent_spec = spec_dict[entry["hash"]]
                 dep_spec = spec_dict[dep_hash]
-                parent_spec._add_dependency(dep_spec, deptypes=deptypes)
+                # TODO: Check this with Peter
+                parent_spec._add_dependency(dep_spec, deptypes=deptypes, virtuals=())
```
alalazo: @scheibelp Is :point_up: correct when reading from a Cray manifest? Or are there cases where we read DAGs that should include virtual dependency annotations on edges?

scheibelp: Virtual information would not be included in the manifest currently. If `virtuals=()`, does that mean that the dependency provides no virtuals (i.e. `spec["mpi"]` would fail for a dag that contains `openmpi`)? Or does it default to providing all the virtuals that it could provide?

alalazo: It means that the dependency provides no virtuals on the edge being added. Currently I didn't change any type of computation internal to Spack, so `spec["mpi"]` still works without looking at edge attributes - but I'm pretty sure that will change in the future.

When reading from previous formats I have added a function that tries to reconstruct the virtual attributes on edges:

https://github.com/spack/spack/blob/5a41fb2aeccb69a38de18a74df3b86312168d399/lib/spack/spack/spec.py#L4808-L4817

Should we do something similar for Cray too (or reuse the same function if possible)?

scheibelp: > It means that the dependency provides no virtuals on the edge being added.

I think we should probably default to inferring that all possible virtuals are provided, but perhaps the next bit will provide clarity on why I think that (and if I'm wrong).

> Should we do something similar for Cray too (or reuse the same function if possible)?

Cray (specifically specs read from manifests on Cray machines) is not different from reading/using specs that were installed prior to this merge. I am confused by the notion that Cray-specific logic is needed:

* If a new install directly depends on a spec read from the manifest, I assume this PR could assign virtual information to the edges
* However, DAGs read from the manifest do not encode the virtual information on the edges, so Specs using these cannot select a subset of virtual functionality (i.e., if `X -> foo` and `foo` provides `blas` and `lapack`, any package that uses `X` must use `foo` for both `blas` and `lapack`)
  * I don't anticipate this will be a problem, as the manifest is primarily for describing things like `cray-libsci` that are direct providers

Does that resolve the issue? Or am I missing something?

alalazo:
For reference, moving from the current `develop` to this PR is fine and any write operation will trigger an upgrade of:
- The database
- The lockfiles
- The spec.json files

where appropriate. 

When moving from this PR back to develop, we'll see the following errors. 

## Develop (e70755f69212a49177d4c4ac0e365c989b90c7df)

Bad error message for databases, understandable error message for environments.

For databases:
```console
$ spack find
==> Error: Version("7")
```
For environments:
```console
$ spack concretize
==> Error: Spack 0.20.0.dev0 cannot read environment lockfiles using the v5 format
```

## releases/v0.19 (9e8e72592d05456b9898350b84f1e4cb8c090dd1)

Understandable error message for databases, bad error message for environments.

For databases:
```console
$ spack find
==> Error: Expected database version 6 but found version 7.
`spack reindex` may fix this, or you may need a newer Spack version.
```
For environments:
```console
$ spack concretize
==> Error: 'type'
```

So not sure how to fix these properly. I'll post some tests with buildcaches tomorrow.

tgamblin:
moving this to 0.21, as it's a large change and we don't *need* it for 0.20.  I think we need to get the safeguards in develop, 0.20, 0.19.3, then get this in soon after 0.20

lib/spack/spack/cray_manifest.py
becker33: @alalazo it seems like this would need the virtual inference that you use for reading old spec formats here.

lib/spack/spack/environment/environment.py
```diff
@@ -1571,7 +1572,7 @@ def _concretize_separately(self, tests=False):
                         if test_dependency in current_spec[node.name]:
                             continue
                         current_spec[node.name].add_dependency_edge(
-                            test_dependency.copy(), deptypes="test"
+                            test_dependency.copy(), deptypes="test", virtuals=()
```
becker33: This seems like it would lose information if the package has a test dependency on a virtual? Am I missing something?

lib/spack/spack/installer.py
```diff
@@ -231,7 +231,9 @@ def _packages_needed_to_bootstrap_compiler(compiler, architecture, pkgs):
     dep.concretize()
     # mark compiler as depended-on by the packages that use it
     for pkg in pkgs:
-        dep._dependents.add(spack.spec.DependencySpec(pkg.spec, dep, deptypes=("build",)))
+        dep._dependents.add(
+            spack.spec.DependencySpec(pkg.spec, dep, deptypes=("build",), virtuals=())
+        )
```
becker33, line 236: I'm going to be so happy when we get to compilers as dependencies and we can rip this kludge out...

lib/spack/spack/test/concretize.py
```diff
@@ -2172,3 +2172,11 @@ def test_concretization_with_compilers_supporting_target_any(self):
...
             s = spack.spec.Spec("a").concretized()
         assert s.satisfies("%gcc@12.1.0")
+
+    def test_virtuals_are_annotated_on_edges(self, default_mock_concretization):
+        spec = default_mock_concretization("mpileaks ^mpich")
+
+        edges = spec.edges_to_dependencies(name="mpich")
+        assert len(edges) == 1 and edges[0].virtuals == ("mpi",)
+        edges = spec.edges_to_dependencies(name="callpath")
+        assert len(edges) == 1 and edges[0].virtuals == ()
```
becker33: Can we make this test the case in which the dependency is found entirely by spack as well? Something like

```suggestion
    @pytest.mark.parametrize("spec", ["mpileaks", "mpileaks ^mpich"])
    def test_virtuals_are_annotated_on_edges(self, spec, default_mock_concretization):
        spec = default_mock_concretization(spec)

        mpiname = spec["mpi"].name
        edges = spec.edges_to_dependencies(name=mpiname)
        assert len(edges) == 1 and edges[0].virtuals == ("mpi",)
        edges = spec.edges_to_dependencies(name="callpath")
        assert len(edges) == 1 and edges[0].virtuals == ()
```

lib/spack/spack/spec.py
```diff
@@ -714,47 +714,81 @@ class DependencySpec:
...
+        """Return a copy of this edge"""
+        return DependencySpec(
+            self.parent, self.spec, deptypes=self.deptypes, virtuals=self.virtuals
+        )
 
     def _cmp_iter(self):
         yield self.parent.name if self.parent else None
         yield self.spec.name if self.spec else None
         yield self.deptypes
+        yield self.virtuals
```
becker33, line 777: Wouldn't it be more forward thinking to yield a dict of `self.parameters`?

lib/spack/spack/spec.py
```diff
@@ -1620,10 +1663,10 @@ def add_dependency_edge(self, dependency_spec: "Spec", *, deptypes: dp.Dependenc
                 # If we are here, it means the edge object was previously added to
                 # both the parent and the child. When we update this object they'll
                 # both see the deptype modification.
-                edge.add_type(deptypes)
+                edge.update_deptypes(deptypes=deptypes)
```
becker33, line 1682: I think we should be able to update the virtual information here as well, is there a reason that's not implemented as part of this?

becker33 (changes requested):
We should probably note in the docs that this extension to the syntax doesn't work well with the old concretizer -- I only know of one user still using it and I think that will be zero pretty soon, but still worth noting. It looks like the syntax will work, but the virtual splitting looks like it will not. Is that accurate?

lib/spack/spack/spec.py
alalazo: We can extend that when we'll work on `use_variants` and we'll introduce parameters other than these two. Using a dict requires us to make the object hashable, since:
```python
>>> {} < {}
TypeError: '<' not supported between instances of 'dict' and 'dict'
```

lib/spack/spack/cray_manifest.py
alalazo: Done in 9c31dea6ea91ec2493c39f3552715028794256b2 It would be good if someone can suggest how to add a test that checks virtuals are reconstructed properly[^1]

[^1]: I stepped in with a debugger and I believe it works, but it would be good to test it I guess.

lib/spack/spack/test/concretize.py
alalazo: Done in 29a8457ce561f956aa24067cb61bf080749f9050

lib/spack/spack/environment/environment.py
alalazo: Done in d594511be3649b904dce68a18c859c9e0ee4226a

lib/spack/spack/spec.py
alalazo: Done in 427bb95c56ce9aa8ad0af54ba25e40108795f860

alalazo:
> We should probably note in the docs that this extension to the syntax doesn't work well with the old concretizer -- I only know of one user still using it and I think that will be zero pretty soon, but still worth noting. It looks like the syntax will work, but the virtual splitting looks like it will not. Is that accurate?

This PR doesn't introduce any change in the spec syntax. It's just a preliminary change in the internal representation. https://github.com/spack/spack/pull/35322 is coded on top of this PR and introduces that change. Are you referring to that PR?

alalazo:
Do we need to take any action on the `develop` buildcache when we merge this? In my opinion it may be worth wiping the `develop` buildcache out and rebuild it from scratch after this PR is merged (so that we'll move every spec to the new format), but let me know if you disagree. 

Also, people using the develop buildcache with a commit preceding the eventual merge, will be unable to continue doing so. I think that is fine, since we also have `develop` snapshots, but this is worth noting explicitly.

becker33:
> This PR doesn't introduce any change in the spec syntax. It's just a preliminary change in the internal representation. https://github.com/spack/spack/pull/35322 is coded on top of this PR and introduces that change. Are you referring to that PR?

Yes, I was conflating with that PR.

becker33 (commented):
LGTM now

becker33 (approved):

PR #32710, opened 2022-09-19
e4s: add power stack

eugeneswalker:
In collaboration with @scottwittenburg -- we have no power nodes in the protected infrastructure set -- going to try and find a way to make E4S Power builds happen on PRs while skipping the post-merge build and cache step that requires protected runner infrastructure.

FYI @wspear

PR #34828, opened 2023-01-05
protected runner jobs at uo: trust e4s-uo signing key

eugeneswalker:
Excluding these changes from the `aarch64` stacks as we do not plan to provide `aarch64` runners for the protected pool.

@kotfic @scottwittenburg

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
@@ -254,6 +254,9 @@ spack:
       - spack env activate --without-view .
       - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}:'morepadding/{architecture}/{compiler.name}-{compiler.version}/{name}-{version}-{hash}'"
       - mkdir -p ${SPACK_ARTIFACTS_ROOT}/user_data
+      # AWS runners mount E4S public key (verification), UO runners mount public/private (signing/verification)
+      - if [[ -r /mnt/key/e4s.gpg ]]; then spack gpg trust /mnt/key/e4s.gpg; fi
+      # UO runners mount intermediate ci public key (verification), AWS runners mount public/private (signing/verification)
       - if [[ -r /mnt/key/intermediate_ci_signing_key.gpg ]]; then spack gpg trust /mnt/key/intermediate_ci_signing_key.gpg; fi
       - if [[ -r /mnt/key/spack_public_key.gpg ]]; then spack gpg trust /mnt/key/spack_public_key.gpg; fi
       - spack --color=always --backtrace ci rebuild > >(tee ${SPACK_ARTIFACTS_ROOT}/user_data/pipeline_out.txt) 2> >(tee ${SPACK_ARTIFACTS_ROOT}/user_data/pipeline_err.txt >&2)
--- a/share/spack/gitlab/cloud_pipelines/stacks/aws-isc/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/aws-isc/spack.yaml
@@ -175,6 +175,9 @@ spack:
       - spack env activate --without-view .
       - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}:'morepadding/{architecture}/{compiler.name}-{compiler.version}/{name}-{version}-{hash}'"
       - mkdir -p ${SPACK_ARTIFACTS_ROOT}/user_data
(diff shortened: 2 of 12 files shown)


scottwittenburg:
The last time I was thinking about this, I thought we were only having UO signing to support the power builds, so #32710 had a bit to support it, but only for that stack.

Based on this PR and the conversation I saw in slack today though, it seems like the plan is to have UO do builds for all stacks?

eugeneswalker:
> ... it seems like the plan is to have UO do builds for all stacks?

Well, the thing is, we need at least a couple of our X86_64 machines in the protected runner pool in order to start running GPU tests using our GPUs. We would like to leverage the existing `spack test` integration in CI to try testing some of the packages like `kokkos +cuda` on the GPU-enabled runners we have.

Maybe we could flesh this out more in one of the CI meetings if there are some aspects we need to consider prior to merge? Bottom line I think we need to have at least a couple of our X86_64 GPU instances in the protected runner pool, as well as our power runners to support (a) `spack test` of GPU-requiring packages as part of CI and (b) E4S power pipeline. What do you think?

share/spack/gitlab/cloud_pipelines/stacks/aws-ahug/spack.yaml
```diff
@@ -254,6 +254,7 @@ spack:
       - spack env activate --without-view .
       - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}:'morepadding/{architecture}/{compiler.name}-{compiler.version}/{name}-{version}-{hash}'"
       - mkdir -p ${SPACK_ARTIFACTS_ROOT}/user_data
+      - if [[ -r /mnt/key/e4s-uo-spack-01.priv ]]; then spack gpg trust /mnt/key/e4s-uo-spack-01.priv; fi
```
scottwittenburg: Adding this new mounted key location, my opinion is the name should be more generic (and not include `.priv`), since aws runners will be mounting the UO public key instead.  Also, I think as @kotfic mentioned in the slack conversation, this line as well as the next one need a comment like:

```
# AWS runners mount UO public key (verification), UO runners mount public/private (signing/verification)
```

... or similar.

Eventually we could rename the `intermediate_ci_signing_key` key to be more generic as well, but for now just a comment like the above (but the other way around) might be enough.

scottwittenburg (changes requested):
I think we can move ahead with letting UO pick up any protected jobs (instead of restricting it to just power builds).  But before we merge this, we'll need to coordinate a PR on spack/spack-infrastructture to provision the UO public key to aws protected runners, and make sure we add that secret wherever it belongs.

eugeneswalker:
@scottwittenburg I updated this PR to address your comments. Can you review? Is this OK?

scottwittenburg (dismissed):
Looks good, thank you!  I think this could be merged now, although we can't make the change to actually let UO runners pick up protected jobs until we get the UO public key mounted onto our runners.

eugeneswalker:
> Looks good, thank you! I think this could be merged now, although we can't make the change to actually let UO runners pick up protected jobs until we get the UO public key mounted onto our runners.

Sounds great. Thanks for the quick review.

I'm thinking it might be best to keep this one open until the other work is done, in case we realize we could tweak this to be better. I'm not saying that will happen, but by leaving this open we keep our options open. And since merging now won't have an effect until the other work is done, there is no cost to leaving it open. What do you think?

scottwittenburg:
> I'm thinking it might be best to keep this one open until the other work is done

Yeah, that makes sense.

kotfic:
K8S infrastructure has been updated to include the e4s.gpg public key which means dependencies built on UO infrastructure should now verify in AWS. I believe once this is rebased on `develop` it should be good to go.

eugeneswalker:
Just re-based this on develop. Hoping we get a green CI pipeline here and we can proceed with merge.

eugeneswalker:
Looks like this is all green and ready to go. @scottwittenburg What do you think?

scottwittenburg:
Just waiting for some protected build jobs to run so I can validate the e4s key mounted on the image, then I'll approve again and merge.  Thanks for the reminder!

scottwittenburg:
I tested trusting the key that gets mounted in the protected runners now (thanks @kotfic), this is what I see:

```
gpg: key 954DB486ABE5E7AA: public key "E4S - University of Oregon - Spack 01" imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg: inserting ownertrust of 6
```

So the key is properly formatted and I'll merge this now.

scottwittenburg (approved):
Thanks @eugeneswalker and @kotfic!

PR #24968, opened 2021-07-20
shellcheck: new package

ajkotobi:

PR #34831, opened 2023-01-06
shellcheck: add new package

aphedges:
From its README:

> [Shellcheck](https://www.shellcheck.net) is a GPLv3 tool that gives warnings and suggestions for bash/sh shell scripts.

This PR is based off someone else's previous attempt in #24968 and existing code in [`cuda/package.py`](https://github.com/spack/spack/blob/6fdb5dfa0a300f352c5eac51884e36083ab00393/var/spack/repos/builtin/packages/cuda/package.py). It's written in Haskell, so downloading the binary seemed like the best way to implement this. I only bothered to include the current version. I was only able to test the two x86 binaries, so I can't confirm the ARM binaries will work properly.

### DIFF
--- a/var/spack/repos/builtin/packages/shellcheck/package.py
+++ b/var/spack/repos/builtin/packages/shellcheck/package.py
@@ -1,0 +1,42 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import platform
+
+from spack.package import *
+
+_versions = {
+    "0.9.0": {
+        "darwin-x86_64": "7d3730694707605d6e60cec4efcb79a0632d61babc035aa16cda1b897536acf5",
+        "linux-aarch64": "179c579ef3481317d130adebede74a34dbbc2df961a70916dd4039ebf0735fae",
+        "linux-armv6hf": "03deed9ded9dd66434ccf9649815bcde7d275d6c9f6dcf665b83391673512c75",
+        "linux-x86_64": "700324c6dd0ebea0117591c6cc9d7350d9c7c5c287acbad7630fa17b1d4d9e2f",
+    },
+}
+
+
+class Shellcheck(Package):
+    """ShellCheck is a GPLv3 tool that gives warnings and suggestions for bash/sh shell scripts."""
+
+    homepage = "https://www.shellcheck.net"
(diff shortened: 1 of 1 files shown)


aphedges:
The CI failure appears to be completely unrelated to this PR. I have rebased on `develop` and force-pushed to trigger checks again.

aphedges:
The checks fortunately pass now.

tldahlgren (approved):
Confirmed the sha256 for my platform.  Rest LGTM.

haampie:
@tldahlgren are we OK with binary packages like these? The project is written in Haskell, the Spack package just installs the pre-built binaries

tldahlgren:
> @tldahlgren are we OK with binary packages like these? The project is written in Haskell, the Spack package just installs the pre-built binaries

That's an excellent point!  What is our policy on this?  Should this PR be reverted?

aphedges:
From someone else's attempt to add `shellcheck` (#24968), it seemed to be acceptable.

It might make sense to build from source some day, but there doesn't appear to be a Haskell toolchain in Spack. At the very least, I couldn't find any mention of Haskell in the documentation, and GHC did not appear to be a package.

aphedges:
Having an official policy about this would be useful, though. I can't find one in the [Packaging Guide](https://spack.readthedocs.io/en/latest/packaging_guide.html).

aphedges:
I found some precedent: Pandoc, written in Haskell, is also installed as a binary: [`pandoc/package.py`](https://github.com/spack/spack/blob/8d83af732a5c9ec42965f7fd94f22a96e5d9823b/var/spack/repos/builtin/packages/pandoc/package.py)

It specifically points to #1408 "for a discussion of the challenges with Haskell".

aphedges:
@tldahlgren, would you like me to add a similar note in the file I created, or would you like to look into the problem first?

tldahlgren:
> @tldahlgren, would you like me to add a similar note in the file I created, or would you like to look into the problem first?

I like that idea.  I'm still awaiting feedback from others.

How about also adding that it installs a binary in the package docstring?  That way `spack info` will let ppl know if they check the package that way.

aphedges:
@tldahlgren, I have created a PR with the suggested changes: #34877.

tldahlgren:
Based on feedback, we'll keep these changes but add the policy to the Packaging Guide (https://github.com/spack/spack/pull/34878).

tldahlgren:
Thanks for your contribution and to you and @haampie for calling attention to the need to have a policy.

aphedges:
@tldahlgren, thanks for creating a policy and for grandfathering this package in! I don't have the knowledge and definitely don't have the time to create a Haskell toolchain.

PR #34832, opened 2023-01-06
Refactor codebase to prefer Pathlib over os.path

johnwparent:
With the introduction of Windows support, better internal path handling is required for best function. Initial stop gap solutions have been effective, but notably fragile and frequently introduced bugs/other headaches (re: the many url handling issues). Additionally continually developing bespoke solutions to problems when they arise as has been the procedure is intractable moving forward as development progresses and users increase (the Path class as it stands now is a good example of this).
The `os.path` module provides a large number of utilities that _can_ mitigate this issue, however, they still require developers to be cognizant of and take great care around Windows vs Posix path handling, platform context, and the Windows drive designation. 

The `Pathlib` module provides much broader, simpler, and effective support for the issue of cross platform path handling, and greatly simplifies the impression Windows specific changes make on the code base (far fewer `is_windows` expressions and related logical statements). 

This PR effects the preliminary steps for such a refactor by removing most current usages of `os.path`  and replacing them with their respective operations in the `Pathlib` module and amending the surrounding handling to conform with the `Pathlib` methods.


johnwparent:
@scheibelp This isn't quite done, but I'm at the point where I'm fixing tests/small issues now, so I wanted to give you a large amount of time to review given the scope and the fact that changes will be very minimal going forward.

PR #34839, opened 2023-01-06
Fix the version number for the fmt dep of pika package

aurianer:


msimberg (approved):
Excellent catch, thanks!

PR #34847, opened 2023-01-06
Charliecloud: add internal SquashFUSE support

benkirk:
Charliecloud's internal SquashFUSE support is enabled with this PR as a variant.  

This requires squashfuse@0.1.105; which is also added.

### DIFF
--- a/var/spack/repos/builtin/packages/charliecloud/package.py
+++ b/var/spack/repos/builtin/packages/charliecloud/package.py
@@ -74,6 +74,7 @@ class Charliecloud(AutotoolsPackage):
         sha256="15ce63353afe1fc6bcc10979496a54fcd5628f997cb13c827c9fc7afb795bdc5",
     )
     variant("docs", default=False, description="Build man pages and html docs")
+    variant("squash", default=True, description="Support internal SquashFUSE mounting")
 
     # Autoconf.
     depends_on("m4", type="build")
@@ -98,6 +99,10 @@ class Charliecloud(AutotoolsPackage):
     # Require pip and wheel for git checkout builds (master).
     depends_on("py-pip@21.1.2:", type="build", when="@master")
     depends_on("py-wheel", type="build", when="@master")
+
+    # squashfs & squashfuse required to build with internal SquashFUSE mounting
+    depends_on("squashfs", when="+squash")
+    depends_on("squashfuse@0.1.105:", when="+squash")
 
     # See https://github.com/spack/spack/pull/16049.
     conflicts("platform=darwin", msg="This package does not build on macOS")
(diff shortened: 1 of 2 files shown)


reidpr (approved):
LGTM, thanks for doing this @benkirk! Very exciting to have this important feature in Spack.

IMO the squash variant should be the default, but I wouldn't want discussion of that to hold up this PR.

benkirk:
Great @reidpr; `+squash` is the default in the PR, happy that's agreeable to you!

haampie:
I'd kinda prefer to stick to the tarball with configure scripts for squashfuse, also those tarballs have a stable checksum, whereas the github download may not. So if you need the 0.1.105 version right now, please make an exception for its url: https://spack.readthedocs.io/en/latest/packaging_guide.html#version-urls

Also please bump the thread here: https://github.com/vasi/squashfuse/issues/74 the 0.1.105 release has issues in general :|

tldahlgren (approved):
Confirmed the sha256 for the 6 new versions of `squashfuse`.

tldahlgren:
Deferring merge decision to @haampie

alecbcs:
@benkirk would it be possible to add SquashFUSE support while still continuing to use the manually created tarball artifacts?

var/spack/repos/builtin/packages/charliecloud/package.py
```diff
@@ -99,6 +100,10 @@ class Charliecloud(AutotoolsPackage):
     depends_on("py-pip@21.1.2:", type="build", when="@master")
     depends_on("py-wheel", type="build", when="@master")
 
+    # squashfs & squashfuse required to build with internal SquashFUSE mounting
+    depends_on("squashfs", when="+squash")
```
j-ogas, line 104:
```suggestion
    depends_on("squashfs", type="build", when="+squash")
```

j-ogas (changes requested):
Very nice, thank you. I believe the `configure_args` method needs something like the following to build charliecloud with squashfuse automatic mount-and-unmount feature.

```python
    def configure_args(self):
[...]
        if "+squashfuse" in self.spec:
            squashfuse_prefix = "{0}".format(self.spec["squashfuse"].prefix)
            args.append("--with-libsquashfuse={0}".format(squashfuse_prefix))
```

var/spack/repos/builtin/packages/charliecloud/package.py
```diff
@@ -99,6 +100,10 @@ class Charliecloud(AutotoolsPackage):
...
+    depends_on("squashfuse@0.1.105:", when="+squash")
```
j-ogas, line 105:
```suggestion
    depends_on("squashfuse@0.1.105:", type="build", when="+squash")
```

j-ogas:
Also, it is worth noting that the Charliecloud auto-mounting feature will also need to link to the host `libfuse3` library at build time. It isn't clear to me how to handle such a thing in Spack. A shot in the dark would be something like:

```python
depends_on("libfuse3.so", type="link", when="+squashfuse")
```

benkirk:
Closing in deference to #41276.

PR #34857, opened 2023-01-09
Add package file for xtb and py-xtb

awvwgk:
Add packages for xtb and py-xtb, also fixes simple-dftd3, dftd4 and mctc-lib packages to account for latest MesonPackage update.

### DIFF
--- a/var/spack/repos/builtin/packages/dftd4/package.py
+++ b/var/spack/repos/builtin/packages/dftd4/package.py
@@ -44,8 +44,16 @@ class Dftd4(MesonPackage):
             lapack = "auto"
 
         return [
-            "--wrap-mode=nodownload",
             "-Dlapack={0}".format(lapack),
             "-Dopenmp={0}".format(str("+openmp" in self.spec).lower()),
             "-Dpython={0}".format(str("+python" in self.spec).lower()),
         ]
+
+    def setup_run_environment(self, env):
+        if "+python" in self.spec:
+            pydir = join_path(
+                self.prefix.lib,
+                "python" + str(self.spec["python"].version.up_to(2)),
+                "site-packages",
+            )
+            env.prepend_path("PYTHONPATH", pydir)
--- a/var/spack/repos/builtin/packages/dftd4/package.py
+++ b/var/spack/repos/builtin/packages/dftd4/package.py
@@ -10,12 +10,13 @@ class Dftd4(MesonPackage):
     """Generally Applicable Atomic-Charge Dependent London Dispersion Correction"""
 
     homepage = "https://www.chemie.uni-bonn.de/pctc/mulliken-center/software/dftd4"
-    url = "https://github.com/dftd4/dftd4/releases/download/v3.3.0/dftd4-3.3.0-source.tar.xz"
+    url = "https://github.com/dftd4/dftd4/releases/download/v3.5.0/dftd4-3.5.0-source.tar.xz"
     git = "https://github.com/dftd4/dftd4.git"
 
     maintainers = ["awvwgk"]
 
     version("main", branch="main")
+    version("3.5.0", "d2bab992b5ef999fd13fec8eb1da9e9e8d94b8727a2e624d176086197a00a46f")
     version("3.4.0", "24fcb225cdd5c292ac26f7d3204ee3c4024174adb5272eeda9ae7bc57113ec8d")
     version("3.3.0", "408720b8545532d5240dd743c05d57b140af983192dad6d965b0d79393d0a9ef")
     version("3.2.0", "cef505e091469aa9b8f008ee1756545bb87b02760bb2c7ca54854e20ba8c590a")
@@ -26,13 +27,14 @@ class Dftd4(MesonPackage):
     variant("python", default=False, description="Build Python extension module")
 
     depends_on("blas")
-    depends_on("cmake", type="build")
     depends_on("lapack")
     depends_on("mctc-lib")
     depends_on("meson@0.57.1:", type="build")  # mesonbuild/meson#8377
     depends_on("pkgconfig", type="build")
     depends_on("py-cffi", when="+python")
     depends_on("python@3.6:", when="+python")
+
+    extends("python", when="+python")
 
     def meson_args(self):
         lapack = self.spec["lapack"].libs.names[0]
@@ -48,12 +50,3 @@ class Dftd4(MesonPackage):
             "-Dopenmp={0}".format(str("+openmp" in self.spec).lower()),
             "-Dpython={0}".format(str("+python" in self.spec).lower()),
         ]
-
-    def setup_run_environment(self, env):
-        if "+python" in self.spec:
-            pydir = join_path(
-                self.prefix.lib,
-                "python" + str(self.spec["python"].version.up_to(2)),
-                "site-packages",
-            )
-            env.prepend_path("PYTHONPATH", pydir)
--- a/var/spack/repos/builtin/packages/mctc-lib/package.py
+++ b/var/spack/repos/builtin/packages/mctc-lib/package.py
@@ -27,6 +27,5 @@ class MctcLib(MesonPackage):
 
     def meson_args(self):
         return [
-            "--wrap-mode=nodownload",
             "-Djson={0}".format("enabled" if "+json" in self.spec else "disabled"),
         ]
--- a/var/spack/repos/builtin/packages/py-xtb/package.py
+++ b/var/spack/repos/builtin/packages/py-xtb/package.py
@@ -10,18 +10,19 @@ class PyXtb(PythonPackage):
     """Python API for the extended tight binding program package"""
 
     homepage = "https://xtb-python.readthedocs.org"
-    url = "https://github.com/grimme-lab/xtb-python/releases/download/v22.1/xtb-22.1.tar.gz"
+    pypi = "xtb/xtb-22.1.tar.gz"
+    git = "https://github.com/grimme-lab/xtb-python.git"
 
     maintainers = [
         *PythonPackage.maintainers,
         "awvwgk",
     ]
 
-    version("22.1", sha256="7a59e7b783fc6e8b7328f55211de681e535a83991b07c4bab73494063f5e9018")
+    version("22.1", "7a59e7b783fc6e8b7328f55211de681e535a83991b07c4bab73494063f5e9018")
 
     depends_on("pkgconfig", type="build")
     depends_on("python@3.7:", type=("build", "run"))
     depends_on("py-cffi", type=("build", "run"))
-    depends_on("py-meson-python@0.11.0:", type="build")
-    depends_on("py-numpy", type="run")
+    depends_on("py-meson-python", type="build")
+    depends_on("py-numpy", type=("build", "run"))
     depends_on("xtb", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-xtb/package.py
+++ b/var/spack/repos/builtin/packages/py-xtb/package.py
@@ -12,7 +12,10 @@ class PyXtb(PythonPackage):
     homepage = "https://xtb-python.readthedocs.org"
     url = "https://github.com/grimme-lab/xtb-python/releases/download/v22.1/xtb-22.1.tar.gz"
 
-    maintainers = ["awvwgk"]
+    maintainers = [
+        *PythonPackage.maintainers,
+        "awvwgk",
+    ]
 
     version("22.1", sha256="7a59e7b783fc6e8b7328f55211de681e535a83991b07c4bab73494063f5e9018")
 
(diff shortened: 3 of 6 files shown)


tldahlgren (approved):
Confirmed the sha256 for the two new packages.  Rest LGTM.

tldahlgren:
@adamjstewart Did you want to comment on the python aspects (e.g., `py-xtb`) of this PR?

var/spack/repos/builtin/packages/py-xtb/package.py
```diff
@@ -0,0 +1,24 @@
...
+from spack.package import *
+
+
+class PyXtb(PythonPackage):
+    """Python API for the extended tight binding program package"""
+
+    homepage = "https://xtb-python.readthedocs.org"
+    url = "https://github.com/grimme-lab/xtb-python/releases/download/v22.1/xtb-22.1.tar.gz"
+
+    maintainers = ["awvwgk"]
```
tldahlgren: FYI. This will overwrite maintainers from the base package.  I'll let Adam comment on whether he likes this idea, but one option is to change this to `maintainers = PythonPackage.maintainers + ["awvwgk"]` so he can also be notified when changes are made.

tldahlgren (dismissed):

var/spack/repos/builtin/packages/py-xtb/package.py
awvwgk: Thanks for the hint, I wasn't aware of this. It would be useful to have a function in `spack.package` to deal with this. Maybe something like this:

```diff
-maintainers = [
-    *PythonPackage.maintainers,
-    "user1",
-    "user2",
-]
+maintainer("user1")
+maintainer("user2")
```

---

Suggested in https://github.com/spack/spack/issues/34879.

awvwgk:
```suggestion
    maintainers = [
        *PythonPackage.maintainers,
        "awvwgk",
    ]
```

var/spack/repos/builtin/packages/dftd4/package.py
```diff
@@ -44,8 +44,16 @@ def meson_args(self):
             lapack = "auto"
 
         return [
-            "--wrap-mode=nodownload",
             "-Dlapack={0}".format(lapack),
             "-Dopenmp={0}".format(str("+openmp" in self.spec).lower()),
             "-Dpython={0}".format(str("+python" in self.spec).lower()),
         ]
+
+    def setup_run_environment(self, env):
```
adamjstewart: You can remove this and replace:
```python
depends_on("python@3.6:", when="+python")
```
with:
```python
```
and it will have the same effect.

var/spack/repos/builtin/packages/mctc-lib/package.py
```diff
@@ -27,6 +27,5 @@ class MctcLib(MesonPackage):
 
     def meson_args(self):
         return [
-            "--wrap-mode=nodownload",
```
adamjstewart, line 30: @eli-schwartz what does this flag do and should we be using it?

var/spack/repos/builtin/packages/py-xtb/package.py
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyXtb(PythonPackage):
+    """Python API for the extended tight binding program package"""
+
+    homepage = "https://xtb-python.readthedocs.org"
+    url = "https://github.com/grimme-lab/xtb-python/releases/download/v22.1/xtb-22.1.tar.gz"
```
adamjstewart: https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#pypi-vs-github

var/spack/repos/builtin/packages/py-xtb/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+    maintainers = [
+        *PythonPackage.maintainers,
+        "awvwgk",
+    ]
```
adamjstewart, line 19: I don't particularly care about the default list of maintainers being overridden. I agree it would be nice to have a way to add maintainers instead of overwriting. If you want to submit a proposal for this I would be happy to review. `maintainers += ...` doesn't work does it?

var/spack/repos/builtin/packages/py-xtb/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+    version("22.1", sha256="7a59e7b783fc6e8b7328f55211de681e535a83991b07c4bab73494063f5e9018")
+
+    depends_on("pkgconfig", type="build")
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-meson-python@0.11.0:", type="build")
+    depends_on("py-numpy", type="run")
```
adamjstewart:
```suggestion
    depends_on("py-numpy", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-xtb/package.py
```diff
@@ -0,0 +1,27 @@
...
+        *PythonPackage.maintainers,
+        "awvwgk",
+    ]
+
+    version("22.1", sha256="7a59e7b783fc6e8b7328f55211de681e535a83991b07c4bab73494063f5e9018")
+
+    depends_on("pkgconfig", type="build")
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-meson-python@0.11.0:", type="build")
```
adamjstewart: Where does 0.11+ come from?

var/spack/repos/builtin/packages/simple-dftd3/package.py
```diff
@@ -34,7 +34,15 @@ class SimpleDftd3(MesonPackage):
     def meson_args(self):
 
         return [
-            "--wrap-mode=nodownload",
             "-Dopenmp={0}".format(str("+openmp" in self.spec).lower()),
             "-Dpython={0}".format(str("+python" in self.spec).lower()),
         ]
+
+    def setup_run_environment(self, env):
```
adamjstewart: Same here, replace `depends_on` with `extends` and remove this function

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-xtb/package.py
tldahlgren: No, `maintainers += ..` results in `NameError: name 'maintainers' is not defined`.

The above also fails (on a simple test) with a `SyntaxError` on the asterisk.  

I've been recommending the following syntax:

`maintainers = PythonPackage.maintainers + ...`

eli-schwartz: The "correct" way to do this is inside an `__init__()` function, probably. Class inheritance doesn't define this variable in advance, but looking it up via self.maintainers works.

Probably wisest to specify the behavior via a new dedicated `extra_maintainers` or something, yeah...

var/spack/repos/builtin/packages/mctc-lib/package.py
awvwgk: I only removed it because MesonPackage now applies it by default, so I don't have to enforce it on per package basis anymore.

eli-schwartz: This is a pretty useful flag, but you might want to set it in the default meson args from MesonPackage instead.

Basically, it's common for projects to vendor code for... various reasons. System integrators, such as linux distros, homebrew, spack, etc. rarely like this, and prefer using dependency packages instead.

Meson tries to solve this by making it *really* easy to use vendored code. So easy, in fact, that no one would dream of doing it in a manner other than the officially approved one. :D And that manner is, doing standard system dependency lookup plus an ini file defining an url to vendor it with, and what dependencies it contains. Then, Meson can guarantee via a global policy how to handle that.

- by default, it's:
  - searched as a system dependency (spack depends_on, in this case)
  - if not found, it downloads and vendors a copy
- if `--wrap-mode=nodownload` is used, then it will never download a copy, and prefer to error out if it must, but if one is already downloaded, it's eligible for use
- if `--wrap-mode=nofallback` is used, vendoring will NEVER be done, missing system dependencies are always errors
- if `--wrap-mode=forcefallback` is used, vendoring is forced, and system dependencies are totally ignored when vendoring is possible


Either nodownload or nofallback are reasonable defaults for spack to use.

eli-schwartz: Ah yes, just saw the new reply :D indeed, it's the default since 258edf7dac07710ea6c50316dc53e5154d9cb3b7 / https://github.com/spack/spack/pull/33717#pullrequestreview-1169543266 (which makes sense).

var/spack/repos/builtin/packages/py-xtb/package.py
awvwgk: Numpy is actually not needed at build time.

var/spack/repos/builtin/packages/py-xtb/package.py
awvwgk: I have followed the development of meson-python  and it has been robustly usable for most of my projects since the 0.11 release. I can try to figure out if older versions will build successfully as well.

var/spack/repos/builtin/packages/py-xtb/package.py
awvwgk: This seems specific to the git-archive generated tarballs, here the sdist is already used from the GH release.

var/spack/repos/builtin/packages/py-xtb/package.py
tldahlgren: This is invalid syntax (at the asterisk).

var/spack/repos/builtin/packages/py-xtb/package.py
adamjstewart: I would still prefer to use PyPI. We have much better support for finding new versions on PyPI (try `spack versions py-xtb` before and after) and PyPI tends to be more stable. See https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#id2 for the correct syntax.

var/spack/repos/builtin/packages/py-xtb/package.py
adamjstewart: It is if you run `spack install --test=root py-xtb`. If anyone can figure out how to add run-time deps to the path during a post-install test let me know, but for now it needs to be both.

var/spack/repos/builtin/packages/py-xtb/package.py
adamjstewart: I would remove the constraint entirely and stick to what the developers require (no version constraint) unless you find evidence otherwise.

var/spack/repos/builtin/packages/py-xtb/package.py
eli-schwartz: Hmm, other packaging systems would have something like `type="test"` and this would be evaluated as part of the build dependencies but only if the install (or package creation) command specified to run tests.

adamjstewart: Spack also has a `type="test"`: https://spack.readthedocs.io/en/latest/packaging_guide.html#dependency-types

We _could_ convert every `type=("build", "run")` to `type=("test", "run")` but it feels kinda silly. We could also drop the ability to test Python imports during installation and only allow them with `spack test` but I don't love that solution either. I wish Spack would just know how to handle this.

awvwgk:
Thanks for the reviews, I hope I could address all comments.

var/spack/repos/builtin/packages/py-xtb/package.py
adamjstewart: Also, some build tools will check `PYTHONPATH` at build-time to make sure the dependencies are actually installed. So they do need to be build deps regardless.

var/spack/repos/builtin/packages/simple-dftd3/package.py
```diff
@@ -12,15 +12,24 @@ class SimpleDftd3(MesonPackage):
...
     version("0.6.0", "c057361565f570cb128489c70131487f71b6891a40e5292dfe37041596810dfe")
     version("0.5.1", "0411fdaebe438f652a970cb232ae3199c4cc840366ed05fda4c38e634632040d")
 
+    def url_for_version(self, version):
+        return (
+            "https://github.com/dftd3/simple-dftd3/releases/download/v{version}/"
+            + "s-dftd3-{version}-source.tar.xz"
+            if version < Version("0.7")
+            else "s-dftd3-{version}.tar.xz"
+        ).format(version=str(version))
```
adamjstewart: This is a bit hard to read. Might be more legible as:
```suggestion
        url = "https://github.com/dftd3/simple-dftd3/releases/download/v{0}/s-dftd3-{0}{1}.tar.xz"
        source = ""
        if version >= Version("0.7"):
            source = "-source"
        return url.format(version, source)
```

awvwgk: Fixed this upstream by naming the artifacts consistently

adamjstewart (approved):

PR #34858, opened 2023-01-09
s3_open: return peekable stream

haampie:
botocore has been using a wrapper around `urllib3.request.HTTPResponse`
since 2014 (https://github.com/boto/botocore/commit/969f0c22cb2727123bc62832ceac93dc4a23005c)
which does two things: allow setting a timeout on the socket (which we
don't use) and erroring when content-length: ... does not match the
actual content length.

The downside is that this wrapper class doesn't derive from any of the
Python IO classes, so we used to create yet another wrapper to make it
more compatible with the type of stream urlopen returns for its builtin
responses for http/https etc... And this wrapper was incomplete as it
did not implement `peek()`. So instead the proposal here is to replace
the two awkward wrappers with BufferedReader, which ensures the stream
object is peekable and derives from BufferedIOBase.

BUT, there's yet another caveat, namely that the underlying stream
is urllib3.request.HTTPResponse instead of http.client.HTTPResponse,
and urllib3 has this weird property of returns closed() == True upon
reading the EOF, which is non-standard. They fixed that weirdness
trying to be backward compatible in [urllib3 v1.25.4](https://github.com/urllib3/urllib3/commit/f0d9ebc41e51c4c4c9990b1eed02d297fd1b20d8). The first
botocore to support that is `botocore v1.12.154` and the
first botocore to *require* it is v1.19. That means that
1. `py-boto3@1.9` and above support `urllib3@v1.25.4`
2. `py-boto3@1.16` and above _require_ `urllib3@1.25.4:`

In our CI we use `py-boto3` version 1.20.35 or higher.

Ubuntu 18.04 provides (shock) an apt package `python3-boto3`
which uses urllib3 version 1.22 :(. Alternatively through `python3-pip3`
you can install boto3 1.23 and urllib 1.26. Ubuntu 20.04's system boto3
is more up-to-date (urllib3 1.25.8).

Notice that `http.client.HTTPResponse` (what urllib returns for e.g.
http/https) does implement `peek` separately from deriving from
BufferedIOBase. I also checked `ftp`, it gives a BufferedReader
object too.

### DIFF
--- a/lib/spack/spack/s3_handler.py
+++ b/lib/spack/spack/s3_handler.py
@@ -7,39 +7,9 @@ import urllib.error
 import urllib.parse
 import urllib.request
 import urllib.response
-from io import BufferedReader, BytesIO, IOBase
+from io import BufferedReader, BytesIO
 
 import spack.util.s3 as s3_util
-
-
-# NOTE(opadron): Workaround issue in boto where its StreamingBody
-# implementation is missing several APIs expected from IOBase.  These missing
-# APIs prevent the streams returned by boto from being passed as-are along to
-# urllib.
-#
-# https://github.com/boto/botocore/issues/879
-# https://github.com/python/cpython/pull/3249
-class WrapStream(BufferedReader):
-    def __init__(self, raw):
-        # In botocore >=1.23.47, StreamingBody inherits from IOBase, so we
-        # only add missing attributes in older versions.
-        # https://github.com/boto/botocore/commit/a624815eabac50442ed7404f3c4f2664cd0aa784
-        if not isinstance(raw, IOBase):
-            raw.readable = lambda: True
-            raw.writable = lambda: False
-            raw.seekable = lambda: False
-            raw.closed = False
-            raw.flush = lambda: None
-        super(WrapStream, self).__init__(raw)
-
-    def detach(self):
-        self.raw = None
-
-    def read(self, *args, **kwargs):
-        return self.raw.read(*args, **kwargs)
-
-    def __getattr__(self, key):
-        return getattr(self.raw, key)
 
 
 def _s3_open(url, method="GET"):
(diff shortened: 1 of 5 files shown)


lib/spack/spack/s3_handler.py
```diff
@@ -60,8 +30,24 @@ def _s3_open(url, method="GET"):
...
+            # urllib3.response.HTTPResponse object. Unlike
+            # http.client.HTTPResponse it does *not* support peek() out of the
+            # box, so we wrap it in BufferedReader. There's again a caveat, namely
+            # https://github.com/urllib3/urllib3/commit/f0d9ebc41e51c4c4c9990b1eed02d297fd1b20d8
+            # is required for this wrapper to work. Thefore we end up needing
+            # needing urllib3 v1.25.4, which is required from botocore >= 1.19,
+            # or py-boto3 >= 1.16, and a *possible dependency* from botocore >= 1.12,
+            # or py-boto3 >= 1.9.
+            stream = obj["Body"]._raw_stream
+            stream.auto_close = False
```
scottwittenburg, line 50: Just curious what is the purpose of `auto_close = False` here?  Is it in case `seek()` takes us to the end of the stream, so the response isn't closed?

lib/spack/spack/s3_handler.py
```diff
@@ -60,8 +30,24 @@ def _s3_open(url, method="GET"):
...
+            # 2. Throw when content-length is incorrect.
+            # Right now the benefits of using this class and having and adapter
+            # for old versions etc does not outweigh the benefits.  Instead, we
+            # can use raw_stream directly, which is a
+            # urllib3.response.HTTPResponse object. Unlike
+            # http.client.HTTPResponse it does *not* support peek() out of the
+            # box, so we wrap it in BufferedReader. There's again a caveat, namely
+            # https://github.com/urllib3/urllib3/commit/f0d9ebc41e51c4c4c9990b1eed02d297fd1b20d8
+            # is required for this wrapper to work. Thefore we end up needing
+            # needing urllib3 v1.25.4, which is required from botocore >= 1.19,
```
scottwittenburg:
```suggestion
            # urllib3 v1.25.4, which is required from botocore >= 1.19,
```

Nit: `needing` is duplicated

scottwittenburg (dismissed):
Makes sense to me @haampie, I just had one tiny nitpick (nbd) and a question, but feel free to merge.  Thanks for the thorough research, good PR description and useful code comments.

lib/spack/spack/s3_handler.py
haampie: See https://github.com/urllib3/urllib3/commit/f0d9ebc41e51c4c4c9990b1eed02d297fd1b20d8 for context

alecbcs:
@scottwittenburg ping! This looks generally good to me to implement a lower bound on boto-core.

haampie:
b055eb4eb69c6bca8567c087c20697cfc362b456 should be reverted before this gets merged

haampie:
@scottwittenburg b055eb4eb69c6bca8567c087c20697cfc362b456 still rebuilds the world. I've stopped it since it was enough to just download a spec.json(.sig) file once over s3://

PR #34860, opened 2023-01-09
spack buildcache create: push all deps / cleanup

haampie:
`spack buildcache create` now follows all edges instead of (link, run) deps.

This is the better default for many situations, such as pushing everything
that built in CI / docker build etc so that subsequent builds are faster. If
e.g. cmake is not pushed, it will still become needed as a build dep at a
later stage when a new version of the root spec is installed, so it would
be weird *not* to push build deps. 

This PR also changes the traverse code into linear time instead of
quadratic (as always :weary:), and changes the criterion of selecting
a spec for packaging to `not external and installed`.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -40,6 +40,7 @@ import spack.platforms
 import spack.relocate as relocate
 import spack.repo
 import spack.store
+import spack.traverse as traverse
 import spack.util.file_cache as file_cache
 import spack.util.gpg
 import spack.util.spack_json as sjson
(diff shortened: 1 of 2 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -1298,57 +1299,44 @@ def _build_tarball(
...
+    elif dependencies:
+        nodes = traverse.traverse_nodes(specs, root=root, deptype="all")
+    else:
+        nodes = set(specs)
 
-        for node in nodes:
-            if not skip_node(node):
-                expanded_set.add(node)
+    # Limit to installed non-externals.
+    packagable = lambda n: not n.external and n.installed
```
alalazo:
```suggestion
    packageable = lambda n: not n.external and n.installed
```

lib/spack/spack/binary_distribution.py
```diff
@@ -1298,57 +1299,44 @@ def _build_tarball(
...
 
-    return expanded_set
+    # Mass install check
+    with spack.store.db.read_transaction():
+        return list(filter(packagable, nodes))
```
alalazo:
```suggestion
        return list(filter(packageable, nodes))
```

alalazo (commented):
LGTM, but I think there's a typo

alalazo (approved):

PR #34861, opened 2023-01-09
spack buildcache create -> push; --update-index

haampie:
spack buildcache create is a misnomer cause it's the only way to push to
an existing buildcache (and it in fact calls `binary_distribution.push`).

Also we have `spack buildcache update-index` but for create the flag is
`--rebuild-index`, which is confusing (and also... why "rebuild"
something if the command is "create" in the first place, that implies it
wasn't there to begin with).

So, after this PR, you can use either

```
spack buildcache create --rebuild-index
```

or

```
spack buildcache push --update-index
```

Also I've added `spack buildcache rebuild-index` as an alias for
`spack buildcache update-index` so there's fewer surprises.

### DIFF
--- a/lib/spack/docs/binary_caches.rst
+++ b/lib/spack/docs/binary_caches.rst
@@ -36,7 +36,7 @@ Build caches are created via:
 
 .. code-block:: console
 
-    $ spack buildcache create <path/url/mirror name> <spec>
+    $ spack buildcache push <path/url/mirror name> <spec>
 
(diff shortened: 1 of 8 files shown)


haampie:
I don't think it's worth deprecating the old API (yet?)

alalazo (changes requested):
Apparently another:
```
$ spack commands --update-completion
```
is needed. Do we have a test checking that the alias works too?

alalazo (approved):

PR #34867, opened 2023-01-10
add dummy perl+open variant, use it in external find, require it in libxcrypt

haampie:
So, this is PR #inf in the series of attempts to fix libxcrypt depends on perl.

The problem previously was that if you do `yum install autoconf` or so on
rhel8, you get a partial perl install, lacking `open.pm`, and if you relied on
`spack external find perl` the build of libxcrypt would fail, which is required
for pretty much anything.

This is now solved by adding a dummy `perl +open` variant, which defaults to
`True`, and can be detected in external find, and is required by `libxcrypt`.


haampie:
@spackbot rerun pipeline

alalazo (approved):

PR #34870, opened 2023-01-10
make boost discoverable by CMake

frankwillmore:
Add Boost_INCLUDE_DIR to expose boost to dependent CMakePackages.

### DIFF
--- a/var/spack/repos/builtin/packages/boost/package.py
+++ b/var/spack/repos/builtin/packages/boost/package.py
@@ -740,3 +740,4 @@ class Boost(Package):
                 env.append_flags("CXXFLAGS", "-DBOOST_USE_UCONTEXT")
             elif context_impl == "winfib":
                 env.append_flags("CXXFLAGS", "-DBOOST_USE_WINFIB")
+        env.set("Boost_INCLUDE_DIR", self.prefix.inc)
--- a/var/spack/repos/builtin/packages/boost/package.py
+++ b/var/spack/repos/builtin/packages/boost/package.py
@@ -740,4 +740,4 @@ class Boost(Package):
                 env.append_flags("CXXFLAGS", "-DBOOST_USE_UCONTEXT")
             elif context_impl == "winfib":
                 env.append_flags("CXXFLAGS", "-DBOOST_USE_WINFIB")
-        env.set("Boost_INCLUDE_DIR", self.prefix.inc)
+        env.set("BOOST_ROOT", self.prefix)


hainest:
I'm likely missing something obvious, but I'm a bit confused as to how the dependent package can't find Boost without this. Does the dependent package not `depends_on("boost...`?

The CMake guidelines generally suggest using the `Package_ROOT`-family of values for influencing dependent CMake projects so that it automatically works with `find_package` and can be overridden by the consumer. I'd recommend using BOOST_ROOT instead of Boost_INCLUDE_DIR.

frankwillmore:
> I'm likely missing something obvious, but I'm a bit confused as to how the dependent package can't find Boost without this. Does the dependent package not `depends_on("boost...`?
> 

This tells spack about the dependency. It doesn't tell the *build system* of the dependent package where to find boost. 
> The CMake guidelines generally suggest using the `Package_ROOT`-family of values for influencing dependent CMake projects so that it automatically works with `find_package` and can be overridden by the consumer. I'd recommend using BOOST_ROOT instead of Boost_INCLUDE_DIR.

CMake will look for packages in a number  of ways. In my use case, CMake complained about not finding Boost_INCLUDE_DIR. I'm agnostic as to how CMake finds it, but it needs *something* here.

hainest:
> > I'm likely missing something obvious, but I'm a bit confused as to how the dependent package can't find Boost without this. Does the dependent package not `depends_on("boost...`?
> 
> This tells spack about the dependency. It doesn't tell the _build system_ of the dependent package where to find boost.

That's the responsibility of the author of the dependent package. Spack doesn't have a uniform way of doing it, so you have to override `configure_args`, `cmake_args`, or the other build system's `_args` and pass the flags manually. What is the dependent package that isn't doing this?

frankwillmore:
The dependent package is rstudio. 

The intent here isn't to pass flags, but to set a variable in the dependent build environment that makes it easily discoverable by CMake.

Agree that there is more than one way to pass this information to the dependent package, and your suggestion of setting BOOST_ROOT is the standard CMake way, so I've updated the PR.

hainest:
@tldahlgren There are currently 257 packages that depend on Boost, 177 of which are CMake packages. Of the 257, 175 do not appear to explicitly set a path to it, and 145 of these are CMake packages. I didn't visually inspect all of them, but they don't use the pattern `<var>.spec["boost"].prefix`. Setting BOOST_ROOT in the dependent build environments for these 145 packages _should_ cause them to correctly find the spack-provided Boost, assuming they are using somewhat sane CMake procedures. However, this would be an action-at-a-distance change. Even if it's more correct, I'm still hesitant to do it. I'd like to get something working for Frank, but I don't want to break the world. Thoughts?

@frankwillmore I promise I'm not intentionally being a pain! I just want to make sure we don't break too many builds. This package hasn't had the best track record in terms of maintenance, so I'm trying to be more vigilant about it. Thanks for your patience!

haampie:
The boost situation is rather cumbersome.

Spack *always* gives `-DCMAKE_PREFIX_PATH=<prefix>` so cmake can use it as a hint in `find_package(...) `

That should really be enough for `find_package(boost)` to work, _as long as Boost provides a cmake config file_, which is only true for version 1.70 onwards.

We already [jump through hoops](https://github.com/spack/spack/blob/7659912dc40bc42df3eb108471974260b3e3b6e8/var/spack/repos/builtin/packages/boost/package.py#L722-L733) to make old Spack-boost (without cmake config file) detectable when system-boost is newer (with cmake config file).

I would be surprised if this still was not enough, so if you can add just a few more details about the rstudio / cmake / boost versions, that'd be great.

Maybe it's related to https://github.com/rstudio/rstudio/blob/7e0900b4542547cd4e76a8ffb1405cc008ee16ec/src/cpp/CMakeLists.txt#LL269C15-L269C15?

frankwillmore:
Looks like this goes deeper than I expected. The rstudio package actually hardcodes CMake options to force it to use system BOOST and system SOCi, even though it was requiring spack to provide them. I was trying to do one better. 

Rstudio also requires BOOST to be built with a half dozen or so variants, which it doesn't expose at all in its spack package. It just fails the build and I would then re-build boost with each of those variants. 

I was ultimately able to find a suitable rpm build of rstudio (the target of my effort) and by adding a few links to the compute node image able to relocate it. 

Thanks all for looking at it.

alalazo:

PR #34877, opened 2023-01-11
shellcheck: add comments about installing binary

aphedges:
ShellCheck is installed with a downloaded binary instead of being compiled from source, and there should be comments to point out this unorthodox approach.

These changes were suggested in <https://github.com/spack/spack/pull/34831#issuecomment-1378057857> after the PR was merged.


tldahlgren (approved):
Thanks.

PR #34873, opened 2023-01-10
Fix issue when generating compiler hierarchy

alalazo:
fixes #34728

Modifications:
- [x] Fix bug affecting generation of Lmod modules

PR #34880, opened 2023-01-11
CompilerSpec: add type hints

alalazo:
Pulled out from #34873 , since we decided to revert the change in `CompilerSpec` constructor.

Modifications:
- [x] Added type hints to `CompilerSpec` methods
- [x] Removed `CompilerSpec._add_versions`, `Spec._add_versions` and `Spec._set_compiler` (dead code)

### DIFF
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -55,7 +55,7 @@ import itertools
 import os
 import re
 import warnings
-from typing import Tuple
+from typing import Tuple, Union, overload
 
 import ruamel.yaml as yaml
 
@@ -559,6 +559,14 @@ class CompilerSpec(object):
 
     __slots__ = "name", "versions"
 
+    @overload
+    def __init__(self, name: str, version: str):
+        ...
+
+    @overload
+    def __init__(self, compiler_spec_str: str):
+        ...
+
     def __init__(self, *args):
         nargs = len(args)
         if nargs == 1:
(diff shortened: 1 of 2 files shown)


lib/spack/spack/spec.py
```diff
@@ -572,35 +581,26 @@ def __init__(self, *args):
...
-                "A spec cannot contain multiple version signifiers." " Use a version list instead."
-            )
-        self.versions = vn.VersionList()
-        for version in version_list:
-            self.versions.add(version)
+            raise TypeError(f"__init__ takes 1 or 2 arguments. ({nargs:d} given)")
 
-    def _autospec(self, compiler_spec_like):
+    @staticmethod
+    def _maybe_literal(compiler_spec_like: Union[str, "CompilerSpec"]) -> "CompilerSpec":
```
tgamblin, line 602: Why do we need this if we can just add another overload to the constructor that admits a `CompilerSpec` arg?

lib/spack/spack/spec.py
```diff
@@ -572,35 +581,26 @@ def __init__(self, *args):
...
+        """Convert a possible string literal to a CompilerSpec object"""
         if isinstance(compiler_spec_like, CompilerSpec):
             return compiler_spec_like
-        return CompilerSpec(compiler_spec_like)
+        spec = spack.parser.parse_one_or_raise(f"%{compiler_spec_like}")
+        return spec.compiler
 
-    def satisfies(self, other, strict=False):
-        other = self._autospec(other)
+    def satisfies(self, other: Union[str, "CompilerSpec"], strict: bool = False) -> bool:
```
tgamblin: I think `CompilerSpecLike = Union[str, "CompilerSpec"]` would be useful somewhere -- then we wouldn't have to write this out.

tgamblin (changes requested):

lib/spack/spack/spec.py
alalazo: Done

lib/spack/spack/spec.py
alalazo: This is keeping the same behavior we had before. Using the constructor creates a copy, in this way we avoid it. Not sure it matters for performance, but I thought that was the reason we added `_autospec`

alalazo:
@spackbot fix style

alalazo:
Superseded by #43114

Issue #34371, opened 2022-12-07
Various binary cache improvements


haampie:
Over the last week(s) there's been quite some discussion about binary caches.

This issue is meant to give an overview of discussions, and previous & current problems, and a suggested way forward.

---

The main problems that triggered these discussions:

1. Requests to `s3://` URLs are very slow, let's say it takes 1-3 seconds per request. Compare this to ~150ms for the equivalent `https://` URL for our spack-binaries bucket.
2. Multiple requests per mirror were issued to locate a spec: `spec.yaml`, `spec.json`, `spec.json.sig`
3. There are typically multiple mirrors configured, like 3-5.

This leads to a significant overhead to fetch binaries in CI, @blue42u reported a lot of time wasted (let's say about 30 minutes) just trying to fetch binaries from mirrors.

---

We already had a small optimization to reduce the number of requests:

4. We check the _local_ or _offline_ cache of the mirror for a spec, and then prioritize mirrors with matching specs, resulting in direct hits (typically).

However, the optimization has a bug: if the spec cannot be located in any local cache (which is either because none of the remotes have the spec at all, or because we don't have a local cache for the mirror), Spack would do a _partial_ update of the cache. Partial in the sense that it would query _each_ mirror if it had the spec by directly fetching the relevant spec.json files. So, in this case, the optimization is doing strictly damage: all mirrors are queried before starting a download; without the "optimization", Spack would simply stop at the first mirror where it can download from.

---

5. @blue42u changed the optimization in point 4. to go from a _partial_ update to a _full_ update of the cache if there was no local cache hit. The idea being, for the next spec to install, the cache can finally be exploited, and the optimization works: mirrors are ordered to get direct hits.

However, 5. only makes sense for terribly slow mirrors, since there is a high startup cost of fetching `index.json` with say a 100K specs. Slow mirrors are not the norm (e.g. file:// mirrors or mirrors on the local network with low latency), so this PR makes the Spack experience only worse for other Spack users. For fast mirrors, we'd really like to do direct fetches (and also use that fully offline mirror order optimization).

In fact, we never had any issues with the https://mirror.spack.io URLs for sources, it would be absurd if Spack would first download an index of _all sources available_ on mirror.spack.io so that it could use that when installing packages from sources.

---

What has not really been looked into is why these s3:// requests are so slow in the first place, and it turns out it's because of various trivial issues:

6. Each s3:// 404 would cause Spack to try and download `<failing url>/index.html`, this was fixed in #34325 
[...]

PR #34891, opened 2023-01-11
Refer to mirrors by name, path, or url

haampie:
With this change we get the invariant that `mirror.fetch_url` and
`mirror.push_url` return valid URLs, even when the backing config
file is actually using (relative) paths with potentially `$spack` and
`$env` like variables.

Secondly it avoids expanding mirror path / URLs too early,
so if I say `spack mirror add name ./path`, it stays `./path` in my
config. When it's retrieved through MirrorCollection() we
exand it to say `file://<env dir>/path` if `./path` was set in an
environment scope.

Thirdly, the interface is simplified for the relevant buildcache
commands, so it's more like `git push`:

```
spack buildcache create [mirror] [specs...]
```

`mirror` is either a mirror name, a path, or a URL.

Resolving the relevant mirror goes as follows:
    
- If it contains either / or \ it is used as an anonymous mirror with
   path or url.
- Otherwise, it's interpreted as a named mirror, which must exist.

This helps to guard against typos, e.g. typing `my-mirror` when there
is no such named mirror now errors with:

```
$ spack -e . buildcache create my-mirror
==> Error: no mirror named "my-mirror". Did you mean ./my-mirror?
```

instead of creating a directory in the current working directory. I
think this is reasonable, as the alternative (requiring that a local dir
exists) feels a bit pendantic in the general case -- spack is happy to
create the build cache dir when needed, saving a `mkdir`.

The old (now deprecated) format will still be available in Spack 0.20,
but is scheduled to be removed in 0.21:

```
spack buildcache create (--directory | --mirror-url | --mirror-name) [specs...]
```

This PR also touches `tmp_scope` in tests, because it didn't really
work for me, since spack fixes the possible --scope values once and
for all across tests, so tests failed when run out of order.

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -1482,7 +1482,7 @@ def _push_mirror_contents(env, specfile_path, sign_binaries, mirror_url):
     tty.debug("Creating buildcache ({0})".format("unsigned" if unsigned else "signed"))
     hashes = env.all_hashes() if env else None
     matches = spack.store.specfile_matches(specfile_path, hashes=hashes)
-    push_url = spack.mirror.push_url_from_mirror_url(mirror_url)
+    push_url = spack.mirror.Mirror.from_url(mirror_url).push_url
     spec_kwargs = {"include_root": True, "include_dependencies": False}
     kwargs = {"force": True, "allow_root": True, "unsigned": unsigned}
     bindist.push(matches, push_url, spec_kwargs, **kwargs)
--- a/lib/spack/spack/cmd/buildcache.py
+++ b/lib/spack/spack/cmd/buildcache.py
@@ -8,7 +8,6 @@ import os
 import shutil
 import sys
 import tempfile
-import urllib.parse
 
 import llnl.util.tty as tty
 
@@ -66,27 +65,37 @@ def setup_parser(subparser):
     create.add_argument(
         "-k", "--key", metavar="key", type=str, default=None, help="Key for signing."
     )
-    output = create.add_mutually_exclusive_group(required=True)
+    output = create.add_mutually_exclusive_group(required=False)
+    # TODO: remove from Spack 0.21
     output.add_argument(
         "-d",
         "--directory",
         metavar="directory",
-        type=str,
-        help="local directory where buildcaches will be written.",
+        dest="mirror_flag",
+        type=arguments.mirror_directory,
+        help="local directory where buildcaches will be written. (deprecated)",
     )
+    # TODO: remove from Spack 0.21
     output.add_argument(
         "-m",
         "--mirror-name",
         metavar="mirror-name",
-        type=str,
-        help="name of the mirror where buildcaches will be written.",
+        dest="mirror_flag",
+        type=arguments.mirror_name,
+        help="name of the mirror where buildcaches will be written. (deprecated)",
     )
+    # TODO: remove from Spack 0.21
     output.add_argument(
         "--mirror-url",
         metavar="mirror-url",
-        type=str,
-        help="URL of the mirror where buildcaches will be written.",
+        dest="mirror_flag",
+        type=arguments.mirror_url,
+        help="URL of the mirror where buildcaches will be written. (deprecated)",
     )
+    # Unfortunately we cannot add this to the mutually exclusive group above,
+    # because we have further positional arguments.
+    # TODO: require from Spack 0.21
+    create.add_argument("mirror", type=str, help="Mirror name, path, or URL.", nargs="?")
     create.add_argument(
         "--rebuild-index",
         action="store_true",
@@ -179,7 +188,7 @@ def setup_parser(subparser):
         "-m",
         "--mirror-url",
         default=None,
-        help="Override any configured mirrors with this mirror url",
+        help="Override any configured mirrors with this mirror URL",
     )
 
     check.add_argument(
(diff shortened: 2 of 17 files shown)


lib/spack/spack/cmd/buildcache.py
```diff
@@ -265,56 +274,106 @@ def setup_parser(subparser):
         default=None,
         help="A quoted glob pattern identifying copy manifest files",
     )
-    source = sync.add_mutually_exclusive_group(required=False)
```
haampie, line 268: @scottwittenburg do you recall why this was optional?

lib/spack/spack/cmd/ci.py
```diff
@@ -241,8 +241,9 @@ def ci_reindex(args):
     ci_mirrors = yaml_root["mirrors"]
     mirror_urls = [url for url in ci_mirrors.values()]
     remote_mirror_url = mirror_urls[0]
+    mirror = spack.mirror.Mirror(remote_mirror_url)
```
haampie, line 244: @scottwittenburg why weren't you just using spack.config limited to the environment scope + spack.mirror.MirrorCollection? This is wrong right now because the schema has changed.

lib/spack/spack/cmd/mirror.py
```diff
@@ -145,8 +145,7 @@ def setup_parser(subparser):
 
 def mirror_add(args):
     """Add a mirror to Spack."""
-    url = url_util.format(args.url)
-    spack.mirror.add(args.name, url, args.scope, args)
+    spack.mirror.add(args.name, args.url, args.scope, args)
```
haampie, line 148: @tgamblin so, I'm just adding mirror urls/paths verbatim now, what do you think? That way you can do `spack -e . mirror add my_mirror '$env/../mirror'` properly. Expanding/canonicalizing/urlifying is done when `mirror_instance.fetch_url` is retrieved whenever mirrors are actually used.

share/spack/spack-completion.bash
```diff
@@ -500,7 +500,7 @@ _spack_buildcache_create() {
     then
         SPACK_COMPREPLY="-h --help -r --rel -f --force -u --unsigned -a --allow-root -k --key -d --directory -m --mirror-name --mirror-url --rebuild-index --spec-file --only"
     else
-        _all_packages
+        _mirrors
```
haampie, line 503: This is a bit sad, but nothing I can do about it :(

lib/spack/spack/cmd/buildcache.py
scottwittenburg: The purpose of it is to communicate the specs that needed rebuliding in the pipeline and should be copied up to the root of the mirror (from the stack-specific mirrors).  While we always do that in our protected pipelines, it's not a feature I would expect gets used by others who might use the `spack ci` commands.  So I made it optional.

haampie: I still don't get it :p it has to be synced from mirror/buildcache A to B, so you need to tell Spack what A and B are? Right now you can also not specify A and B?

scottwittenburg: The manifest files are json files that say exactly what subset of specs to copy and include `src` and `dest` urls for each spec.  In the case of spack pipelines, the manifest files are produced by the pipeline generation job, and they're passed (via artifacts) to `protected-publish` job.  Here's an [example](https://gitlab.spack.io/spack/spack/-/jobs/5128518/artifacts/browse/jobs_scratch_dir/specs_to_copy/).

lib/spack/spack/cmd/ci.py
scottwittenburg: I don't think I understand the question.  But in case it helps: for pipelines, the first mirror in the `spack.yaml` is the special one we use to decide what needs to be rebuilt by comparing the hashes in the environment with binary packages in that mirror.

haampie: Mirrors may have a different fetch / push url. That means in config it can look like this

```yaml
mirrors:
  name: <url>
```

or like this

```yaml
mirrors:
  name:
    fetch:
      url: http://example.com
    push:
      url: s3://bucket/x
```

So, it's wrong for multiple reasons: 

1. it cannot work with the second type of config
2. if you do `spack.config.get(...)` the yaml may go through transformations that you're missing when you directly look at the yaml.

Currently it feels very brittle. Wouldn't it be much better to just refer to mirrors by name? How much work is that to fix?

scottwittenburg: IIUC, you're suggesting the ci module might have some documented naming convention for the special mirrors it uses, and then use those names instead of just bare urls or accessing `yaml_root["mirrors"][0]?  

I think that makes sense, and I don't think it's *that* much work.  But there are a few such mirrors used throughout the `ci` and `cmd/ci` modules.  Do you want me to make a PR to do that?

The special mirrors I can think of off the top of my head include 1) the main stack mirror you pointed to above, indexed directly from the `spack.yaml`, 2) the buildcache override mirror url we specify to customize where binaries should be pushed for pipelines in the same stack, but on different branches, 3) some older and maybe shouldn't be supported any more mirrors like the `artifacts_buildcache_mirror` for when we used to pass buildcaches from one stage to the next via artifacts.  Maybe there are some others I'm forgetting, but those first two are the main ones.

haampie: Ok, I think it's a bit out of scope for this PR (also because I'm not very familiar with the magic mirrors). If you can do that, that'd be great 🙃 also that gives us hopefully an opportunity to switch to https:// for fetching and s3:// for pushing in gitlab ci, in my experience it has much lower latency.

lib/spack/spack/cmd/buildcache.py
haampie: Okay, well, whatever. I made it optional again.

lib/spack/spack/cmd/ci.py
haampie: Ok @scottwittenburg, I'll mark this as "resolved", looking forward to a PR that improves magic mirrors

lib/spack/spack/cmd/mirror.py
tgamblin: I like this because it goes along with respecting the user said -- either on the CLI or in a YAML file.  I guess the only only question I have is what happens if the user says:

```
spack mirror add ./foo
```

When the scope to add ./foo in is an environment but the user is not in the env directory?  I am not 100% sure whether in that case we should be absolutifying the path.  I wonder how much this comes up.

It looks like git does what you do and stores the path verbatim.  Here's an example in a directory that is not alongside my `spack` installation in `~/src/spack`:

```console
(spackbook):2022> git init ../foobar
Initialized empty Git repository in /Users/gamblin2/pubs/presentations/foobar/.git/
(spackbook):2022> git -C ~/src/spack remote add blah ../foobar
(spackbook):2022> git -C ~/src/spack remote -v|grep foobar
blah	../foobar (fetch)
blah	../foobar (push)
```

```console
(spackbook):2022> git -C ~/src/spack pull blah
fatal: '../foobar' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

```console
(spackbook):2022> mv ../foobar ~/src
(spackbook):2022> git -C ~/src/spack pull blah
You asked to pull from the remote 'blah', but did not specify
a branch. Because this is not the default configured remote
for your current branch, you must specify a branch on the command line.
```

So I think what you're doing is fine -- I know I've never actually noticed this issue for `git` so I doubt I would notice it for `spack`.  I suspect people using relative paths will be trying to do for some special environment.

lib/spack/spack/ci.py
```diff
@@ -1482,7 +1482,7 @@ def _push_mirror_contents(env, specfile_path, sign_binaries, mirror_url):
     tty.debug("Creating buildcache ({0})".format("unsigned" if unsigned else "signed"))
     hashes = env.all_hashes() if env else None
     matches = spack.store.specfile_matches(specfile_path, hashes=hashes)
-    push_url = spack.mirror.push_url_from_mirror_url(mirror_url)
+    push_url = spack.mirror.Mirror.from_url(mirror_url).push_url
```
tgamblin, line 1485: This is better

lib/spack/spack/mirror.py
```diff
@@ -663,31 +698,12 @@ def create_mirror_from_package_object(pkg_obj, mirror_cache, mirror_stats):
...
-
-def push_url_from_mirror_name(mirror_name):
-    """Given a mirror name, return the URL on which to push binary packages."""
-    mirror = spack.mirror.MirrorCollection().lookup(mirror_name)
-    if mirror.name == "<unnamed>":
+def require_mirror_name(mirror_name):
+    """Find a mirror by name and raise if it does not exist"""
+    mirror = spack.mirror.MirrorCollection().get(mirror_name)
+    if not mirror:
         raise ValueError('no mirror named "{0}"'.format(mirror_name))
```
tgamblin, line 705: Should this raise a `MirrorError` subclass?  I know it didn't before but I wonder for consistency.  Or is `ValueError` better since it's really a programmer or user error and not an issue with mirrors per se.  Anyway it's minor and shouldn't hold up the merge.

lib/spack/spack/test/ci.py
```diff
@@ -23,22 +23,6 @@
...
-    num_overrides = 0
-    scope_name = base_name
-    while scope_name in current_overrides:
-        scope_name = "{0}{1}".format(base_name, num_overrides)
-        num_overrides += 1
-
-    with cfg.override(cfg.InternalConfigScope(scope_name)):
-        yield scope_name
-
-
```
tgamblin, line 41: thank goodness

lib/spack/spack/test/cmd/gpg.py
```diff
@@ -25,25 +25,6 @@
...
-    num_overrides = 0
-    scope_name = base_name
-    while scope_name in current_overrides:
-        scope_name = "{0}{1}".format(base_name, num_overrides)
-        num_overrides += 1
-
-    with spack.config.override(spack.config.InternalConfigScope(scope_name)):
-        yield scope_name
-
-
```
tgamblin, line 46: oof there were 3 of these

lib/spack/spack/test/cmd/mirror.py
```diff
@@ -26,25 +26,6 @@
...
-    num_overrides = 0
-    scope_name = base_name
-    while scope_name in current_overrides:
-        scope_name = "{0}{1}".format(base_name, num_overrides)
-        num_overrides += 1
-
-    with spack.config.override(spack.config.InternalConfigScope(scope_name)):
-        yield scope_name
-
-
```
tgamblin, line 47: sigh

tgamblin:
@blue42u FYI

haampie:
Would be good to get this merged :soon: , since in total 4 people have reported issues in #34532 / #34371

blue42u:
I like the direction this is going, thanks!

Would this be easy to extend to `spack ci (generate|rebuild)` as well, á la the proposal in https://github.com/spack/spack/discussions/32958?

haampie:
Can we consider merging this @tgamblin?

lib/spack/spack/cmd/buildcache.py
```diff
@@ -383,10 +436,30 @@ def _concrete_spec_from_args(args):
...
+        tty.warn(
+            "Using flags to specify mirrors is deprecated and will be removed in "
+            "Spack 0.21, use positional arguments instead."
+        )
+
+    # TODO: remove this in 0.21. If we have mirror_flag, the first
+    # spec is in the positional mirror arg due to argparse limitations.
+    specs = args.specs
+    if args.mirror_flag and args.mirror:
+        specs.insert(0, args.mirror)
```
tgamblin, line 456: I tried this locally with just the mirror arg and it doesn't seem to work:

```console
> spack mirror create ./cache zlib
==> Concretizing input specs
==> Error: unexpected tokens in the spec string
./cache zlib
^
```

```console
> spack mirror create cache zlib
==> Concretizing input specs
==> Error: Package 'cache' not found.
```

```console
> spack mirror create cache
==> Concretizing input specs
==> Error: Package 'cache' not found.
```

I believe that's because it's trying to concretize the specs in the argparse action... but the first spec ends up being the directory name.  Right?  Or am I doing something wrong?

Do you need to put in some special handling to move over the argument *before* the specs get concretized?  I can see how this is annoying.

tgamblin (changes requested):

lib/spack/spack/cmd/buildcache.py
haampie: Heh, I did not touch `spack mirror create`, only `spack buildcache create`. I suppose it should have the same signature as `spack buildcache create` then.

tgamblin: I guess I shouldn’t review at 1am 🤦🏻

tgamblin: And yeah they should probably be symmetric!

haampie: Unfortunately a positional argument is breaking :sob: 

`spack mirror create [specs...]` without `--directory` mirrors sources into a default directory.

I don't want this PR to be breaking, so I'm afraid `spack mirror create` will be the exception :(

tgamblin (approved):

PR #34892, opened 2023-01-11
rabbitmq-c: added RabbitMQ C interface

lpottier:


bernhardkaindl (changes requested):
spack install `rabbitmq-c~static` fails with:
```py
  >> 34    CMake Error at CMakeLists.txt:293 (message):
     35      Tests can only be built against static libraries (set BUILD_STATIC_LIBS=ON)
```
I suggest removing the variant `static` unless it is absolutely required to prevent/require building a static lib (otherwise fix it)

bernhardkaindl (approved):
Builds now with +static and with ~static.

PR #34896, opened 2023-01-12
Add fpart package

drkrynstrng:
Adding an fpart package, which provides the fpart and fpsync tools.

### DIFF
--- a/var/spack/repos/builtin/packages/fpart/package.py
+++ b/var/spack/repos/builtin/packages/fpart/package.py
@@ -15,6 +15,8 @@ class Fpart(AutotoolsPackage):
     url = "https://github.com/martymac/fpart/archive/refs/tags/fpart-1.5.1.tar.gz"
     git = "https://github.com/martymac/fpart.git"
 
+    maintainers = ["drkrynstrng"]
+
     version("master", branch="master")
     version("1.5.1", sha256="c353a28f48e4c08f597304cb4ebb88b382f66b7fabfc8d0328ccbb0ceae9220c")
 
--- a/var/spack/repos/builtin/packages/fpart/package.py
+++ b/var/spack/repos/builtin/packages/fpart/package.py
@@ -1,0 +1,31 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Fpart(AutotoolsPackage):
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/fpart/package.py
```diff
@@ -0,0 +1,33 @@
...
+    to launch several synchronization jobs in parallel."""
+
+    homepage = "https://www.fpart.org"
+    url = "https://github.com/martymac/fpart/archive/refs/tags/fpart-1.5.1.tar.gz"
+    git = "https://github.com/martymac/fpart.git"
+
+    maintainers = ["drkrynstrng"]
+
+    version("master", branch="master")
+    version("1.5.1", sha256="c353a28f48e4c08f597304cb4ebb88b382f66b7fabfc8d0328ccbb0ceae9220c")
```
tldahlgren, line 21: Confirmed the branch exists and the sha256 is correct.

var/spack/repos/builtin/packages/fpart/package.py
```diff
@@ -0,0 +1,33 @@
...
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+    depends_on("m4", type="build")
+    # fpsync has the following run dependencies
+    depends_on("rsync", type="run")
+    depends_on("tar", type="run")
+    depends_on("cpio", type="run")
+
+    def autoreconf(self, spec, prefix):
+        autoreconf("--install", "--verbose", "--force")
```
tldahlgren: Curious.  Why is the default implementation, which already uses these options, not acceptable?  (See third paragraph, after the figure at https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#build-system-dependencies.)

tldahlgren: If the goal is to force `autoconf` you can set a property (see https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#force-autoreconf).

drkrynstrng: Oops, yeah, the default implementation works.

var/spack/repos/builtin/packages/fpart/package.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
```
tldahlgren: We haven't switched to enforcing 2023 yet.  Can you change this to 2022 so I can merge it?

drkrynstrng: Yes, switched.

tldahlgren (approved):

PR #34900, opened 2023-01-12
flex%oneapi@2023.0.0: -Wno-error=implicit-function-declaration

eugeneswalker:
Needed to resolve the following error. Not sure why the existing `flag_handler` method is not sufficient to silence this. If anyone knows, please share!

```
...
==> flex: Executing phase: 'autoreconf'
==> flex: Executing phase: 'configure'
==> flex: Executing phase: 'build'
==> Error: ProcessError: Command exited with status 2:
    'make' '-j16' 'V=1'
...
  >> 336    scan.l:1036:3: error: use of undeclared identifier 'infilename'
     337                    infilename = xstrdup(file);
     338                    ^
     339    scan.l:1036:16: warning: call to undeclared function 'xstrdup'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
     340                    infilename = xstrdup(file);
     341                                 ^
  >> 342    scan.l:1037:17: error: use of undeclared identifier 'infilename'
     343                    yyin = fopen( infilename, "r" );
     344                                  ^
     345    scan.l:1040:4: warning: call to undeclared function 'lerr'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
     346                            lerr( _( "can't open %s" ), file );
     347                            ^
     348    scan.l:1040:10: warning: call to undeclared function '_'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
     349                            lerr( _( "can't open %s" ), file );
     350                                  ^
  >> 351    scan.l:1046:3: error: use of undeclared identifier 'infilename'
     352                    infilename = xstrdup("<stdin>");
     353                    ^
     354    scan.l:1046:16: warning: call to undeclared function 'xstrdup'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
     355                    infilename = xstrdup("<stdin>");
     356                                 ^
  >> 357    scan.l:1049:2: error: use of undeclared identifier 'linenum'
     358            linenum = 1;
```


wspear (approved):

bernhardkaindl (approved):
@eugeneswalker: Regarding `"ISO C99 and later do not support implicit function declarations"`:

Recently, the CLang developers attempted to take a draconic stance on that and enabled it by default on one of the versions of CLang, but they were forced to revert his change to allow implicit function declarations at least until the next major version of CLang:
https://reviews.llvm.org/D122983

In slack, for a brief amount of time this LLVM build was also the latest build and as far as I remember we had the same issue even with the spack llvm+clang build, but spack has since added a newer version of clang which reverted this.

I suggest that the OneAPI toolchain should be updated likewise to the latest LLVM/CLang which n longer takes this draconic stance and also fixes other bugs.

In general, until the OneAPI developers notice and correct the problem, I think it would be better to add a detection of the broken Clang version and add `-Wno-error=implicit-function-declaration` on the Spack compiler flags level (just for the broken version, I suggest)

eugeneswalker:
@spackbot run pipeline

eugeneswalker:
@spackbot run pipeline

kuberry (commented):
Introducing `def setup_build_environment` breaks an x86_64 build on the mac using `apple-clang@14.0`. It doesn't matter if the contents of that function are simply `pass`.

Can we get this reverted and look into ways to handle this in the `flag_handler`?

eugeneswalker:
> Introducing `def setup_build_environment` breaks an x86_64 build on the mac using `apple-clang@14.0`. It doesn't matter if the contents of that function are simply `pass`.
> 
> Can we get this reverted and look into ways to handle this in the `flag_handler`?

Yes, let us do that. I tried flag_handler initially but it did not have the desired effect.

kuberry:
Ok, I think we could choose to inject that into `envf` rather than `injf` (https://spack.readthedocs.io/en/latest/packaging_guide.html#compiler-flags) with something like:

```py
    def flag_handler(self, name, flags):
        spec = self.spec
        eflags = []
        if name == "cflags":
            if spec.satisfies("%oneapi"):
                eflags.append("-Wno-error=implicit-function-declaration")
        return (None, eflags, None)
```

I can't test this right now, but just a suggestion for how it could work.

Issue #34897, opened 2023-01-12
libtool is not filtered with Fujitsu compilers are used


ggouaillardet:
### Steps to reproduce

$ spack install --verbose testfilter %fj
[+] /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/fj-4.8.1/gnuconfig-2021-08-14-4m7rbpbv5rxc3rmornr7kh5556c2ajan
==> Installing testfilter-0.3-auug2b6oxhttnilfjv3ylqp6h7vdhkyp
==> No binary for testfilter-0.3-auug2b6oxhttnilfjv3ylqp6h7vdhkyp found: installing from source
==> Fetching https://github.com/ggouaillardet/testfilter/archive/refs/tags/v0.3.tar.gz
==> No patches needed for testfilter
==> testfilter: Executing phase: 'autoreconf'
==> testfilter: Executing phase: 'configure'
==> [2023-01-12-13:17:07.518990] FILTER FILE: /tmp/r00018/spack-stage/spack-stage-testfilter-0.3-auug2b6oxhttnilfjv3ylqp6h7vdhkyp/spack-src/configure [replacing "^(\s*if test x-L = )("\$p" \|\|\s*)$"]
==> [2023-01-12-13:17:07.544766] FILTER FILE: /tmp/r00018/spack-stage/spack-stage-testfilter-0.3-auug2b6oxhttnilfjv3ylqp6h7vdhkyp/spack-src/configure [replacing "^(\s*test x-R = )("\$p")(; then\s*)$"]
==> [2023-01-12-13:17:07.567225] FILTER FILE: /tmp/r00018/spack-stage/spack-stage-testfilter-0.3-auug2b6oxhttnilfjv3ylqp6h7vdhkyp/spack-src/configure [replacing "^(\s*test \$p = "-R")(; then\s*)$"]
==> [2023-01-12-13:17:07.592455] '/tmp/r00018/spack-stage/spack-stage-testfilter-0.3-auug2b6oxhttnilfjv3ylqp6h7vdhkyp/spack-src/configure' '--prefix=/vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/fj-4.8.1/testfilter-0.3-auug2b6oxhttnilfjv3ylqp6h7vdhkyp'
[...]

PR #34916, opened 2023-01-13
fix libtool filter for Fujitsu compilers

ggouaillardet:
Since legit regular expressions are matched, do not pass string=True to the filter

Refs #34897

### DIFF
--- a/lib/spack/spack/build_systems/autotools.py
+++ b/lib/spack/spack/build_systems/autotools.py
@@ -427,15 +427,15 @@ To resolve this problem, please try the following:
             x.filter(regex="-nostdlib", repl="", string=True)
             rehead = r"/\S*/"
             for o in [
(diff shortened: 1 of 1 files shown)


lib/spack/spack/build_systems/autotools.py
```diff
@@ -435,7 +435,7 @@ def _do_patch_libtool(self):
                 "crtbeginS.o",
                 "crtendS.o",
```
skosukhin: This and above is now passed as a regular expression, therefore I would escape the dots.
```suggestion
                r"crtendS\.o",
```

skosukhin (approved):
LGTM

alalazo:
@spackbot run pipeline

alalazo (approved):

PR #34924, opened 2023-01-13
Add CUDA/ROCm arch/target propagation to dependencies in pika package

aurianer:


var/spack/repos/builtin/packages/pika/package.py
```diff
@@ -105,6 +105,20 @@ class Pika(CMakePackage, CudaPackage, ROCmPackage):
...
+            depends_on("whip amdgpu_target={0}".format(val),
+                when="@0.9: amdgpu_target={0}".format(val))
+            depends_on("rocsolver amdgpu_target={0}".format(val),
+                when="@0.5: amdgpu_target={0}".format(val))
+            depends_on("rocblas amdgpu_target={0}".format(val),
+                when="amdgpu_target={0}".format(val))
+
+    with when("+cuda"):
+        for val in CudaPackage.cuda_arch_values:
+            depends_on("whip cuda_arch={0}".format(val),
```
msimberg: This actually also needs the version constraint on `whip`.

aurianer: Thanks for spotting that!

msimberg (approved):

PR #34926, opened 2023-01-13
clingo-bootstrap: pgo, lto, allocator optimizations

haampie:
This enables 3 optimizations for clingo:

1. Profile Guided Optimization "trained" on Spack's solver
2. A more efficient allocator on Linux (mimalloc)
3. Link-time optimization

With mimalloc the idea is to locally override `operator new` / `delete`
in `libclingo.so`, without affecting or being affected by other C++ libraries
that Python happens to `dlopen` before/after clingo is imported.
Currently the easiest solution seems to put mimalloc's operator new
in one of clingo's source files, statically link libstdc++, and hide all
symbols except for the clingo public api.

PGO is implemented in the build steps and was tested with
`%gcc`, `%clang`, and `%apple-clang`.

```
$ spack solve --timers hdf5
    setup          4.231s
    load           0.016s
    ground         2.039s
    solve          3.155s
    total          9.464s

$ spack install clingo-bootstrap +optimized  # installs clingo-bootstrap/6wg272k

$ spack load /6wg272k

$ spack solve --timers hdf5
    setup          4.179s
    load           0.014s
    ground         1.435s
    solve          2.279s
    total          7.930s
```

Runtime reduces by 15% overall, time spent in clingo drops about 30%.

I tried to concretize the build_systems stack using the e4s docker image, and got this:

Before

```
real	0m20.279s
user	0m19.999s
sys	0m0.280s
```

After

```
real	0m16.724s
user	0m16.397s
sys	0m0.327s
```

that is -17.5%

### DIFF
--- a/var/spack/repos/builtin/packages/clingo-bootstrap/package.py
+++ b/var/spack/repos/builtin/packages/clingo-bootstrap/package.py
@@ -90,8 +90,12 @@ class ClingoBootstrap(Clingo):
 
     @run_before("cmake", when="+optimized")
     def pgo_train(self):
-        if self.spec.compiler.name in ("clang", "apple-clang"):
+        if self.spec.compiler.name == "clang":
             llvm_profdata = which("llvm-profdata", required=True)
+        elif self.spec.compiler.name == "apple-clang":
+            llvm_profdata = Executable(
+                Executable("xcrun")("-find", "llvm-profdata", output=str).strip()
+            )
 
         # First configure with PGO flags, and do build apps.
         reports = os.path.abspath("reports")
@@ -116,6 +120,8 @@ class ClingoBootstrap(Clingo):
         python_runtime_env = EnvironmentModifications()
         for s in self.spec.traverse(deptype=("run", "link"), order="post"):
             python_runtime_env.extend(spack.user_environment.environment_modifications_for_spec(s))
+        python_runtime_env.unset("SPACK_ENV")
+        python_runtime_env.unset("SPACK_PYTHON")
         self.spec["python"].command(
             spack.paths.spack_script, "solve", "--fresh", "hdf5", extra_env=python_runtime_env
         )
--- a/var/spack/repos/builtin/packages/clingo-bootstrap/package.py
+++ b/var/spack/repos/builtin/packages/clingo-bootstrap/package.py
@@ -2,8 +2,15 @@
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
+import glob
+import os
+
+import spack.compilers
+import spack.paths
+import spack.user_environment
 from spack.package import *
 from spack.pkg.builtin.clingo import Clingo
+from spack.util.environment import EnvironmentModifications
 
 
 class ClingoBootstrap(Clingo):
@@ -13,24 +20,51 @@ class ClingoBootstrap(Clingo):
 
     variant("build_type", default="Release", values=("Release",), description="CMake build type")
 
-    variant("static_libstdcpp", default=False, description="Require a static version of libstdc++")
+    variant(
+        "static_libstdcpp",
+        default=False,
+        when="platform=linux",
+        description="Require a static version of libstdc++",
+    )
+
+    variant(
+        "optimized",
+        default=False,
+        description="Enable a series of Spack-specific optimizations (PGO, LTO, mimalloc)",
+    )
+
+    # Enable LTO
+    conflicts("~ipo", when="+optimized")
+
+    with when("+optimized platform=linux"):
+        # Statically linked. Don't use ~override so we don't duplicate malloc/free, they
+        # get resolved to Python's libc's malloc in our case anyway.
+        depends_on("mimalloc +ipo libs=static ~override", type="build")
+        conflicts("~static_libstdcpp", msg="Custom allocator requires static libstdc++")
+        # Override new/delete with mimalloc.
+        patch("mimalloc.patch", when="@5.5.0:")
+        patch("mimalloc-pre-5.5.0.patch", when="@:5.4")
+        # ensure we hide libstdc++ with custom operator new/delete symbols
+        patch("version-script.patch")
 
     # CMake at version 3.16.0 or higher has the possibility to force the
     # Python interpreter, which is crucial to build against external Python
     # in environment where more than one interpreter is in the same prefix
     depends_on("cmake@3.16.0:", type="build")
 
-    # On Linux we bootstrap with GCC
-    for compiler_spec in [c for c in spack.compilers.supported_compilers() if c != "gcc"]:
+    # On Linux we bootstrap with GCC or clang
+    for compiler_spec in [
+        c for c in spack.compilers.supported_compilers() if c not in ("gcc", "clang")
+    ]:
         conflicts(
             "%{0}".format(compiler_spec),
             when="platform=linux",
-            msg="GCC is required to bootstrap clingo on Linux",
+            msg="GCC or clang are required to bootstrap clingo on Linux",
         )
         conflicts(
             "%{0}".format(compiler_spec),
             when="platform=cray",
-            msg="GCC is required to bootstrap clingo on Cray",
+            msg="GCC or clang are required to bootstrap clingo on Cray",
         )
     conflicts("%gcc@:5", msg="C++14 support is required to bootstrap clingo")
 
@@ -51,28 +85,64 @@ class ClingoBootstrap(Clingo):
 
     def cmake_args(self):
         args = super().cmake_args()
-        args.extend(
-            [
-                # Avoid building the clingo executable
-                self.define("CLINGO_BUILD_APPS", "OFF")
-            ]
-        )
+        args.append(self.define("CLINGO_BUILD_APPS", False))
         return args
 
-    def setup_build_environment(self, env):
-        opts = None
-        if "%apple-clang platform=darwin" in self.spec:
-            opts = "-mmacosx-version-min=10.13"
-        elif "%gcc" in self.spec:
-            if "+static_libstdcpp" in self.spec:
-                # This is either linux or cray
-                opts = "-static-libstdc++ -static-libgcc -Wl,--exclude-libs,ALL"
-        elif "platform=windows" in self.spec:
-            pass
-        else:
-            msg = 'unexpected compiler for spec "{0}"'.format(self.spec)
-            raise RuntimeError(msg)
+    @run_before("cmake", when="+optimized")
+    def pgo_train(self):
+        if self.spec.compiler.name in ("clang", "apple-clang"):
+            llvm_profdata = which("llvm-profdata", required=True)
 
-        if opts:
-            env.set("CXXFLAGS", opts)
-            env.set("LDFLAGS", opts)
+        # First configure with PGO flags, and do build apps.
+        reports = os.path.abspath("reports")
+        sources = os.path.abspath(self.root_cmakelists_dir)
+        cmake_options = self.std_cmake_args + self.cmake_args() + [sources]
+
+        # Set PGO training flags.
+        generate_mods = EnvironmentModifications()
+        generate_mods.append_flags("CFLAGS", "-fprofile-generate={}".format(reports))
+        generate_mods.append_flags("CXXFLAGS", "-fprofile-generate={}".format(reports))
+        generate_mods.append_flags("LDFLAGS", "-fprofile-generate={} --verbose".format(reports))
+
+        with working_dir(self.build_directory, create=True):
+            cmake(*cmake_options, sources, extra_env=generate_mods)
+            make()
+            make("install")
+
+        # Clean the reports dir.
+        rmtree(reports, ignore_errors=True)
+
+        # Run spack solve --fresh hdf5 with instrumented clingo.
+        python_runtime_env = EnvironmentModifications()
+        for s in self.spec.traverse(deptype=("run", "link"), order="post"):
+            python_runtime_env.extend(spack.user_environment.environment_modifications_for_spec(s))
+        self.spec["python"].command(
+            spack.paths.spack_script, "solve", "--fresh", "hdf5", extra_env=python_runtime_env
+        )
+
+        # Clean the build dir.
+        rmtree(self.build_directory, ignore_errors=True)
+
+        if self.spec.compiler.name in ("clang", "apple-clang"):
+            # merge reports
+            use_report = join_path(reports, "merged.prof")
+            raw_files = glob.glob(join_path(reports, "*.profraw"))
+            llvm_profdata("merge", "--output={}".format(use_report), *raw_files)
+            use_flag = "-fprofile-instr-use={}".format(use_report)
+        else:
+            use_flag = "-fprofile-use={}".format(reports)
+
+        # Set PGO use flags for next cmake phase.
+        use_mods = EnvironmentModifications()
+        use_mods.append_flags("CFLAGS", use_flag)
+        use_mods.append_flags("CXXFLAGS", use_flag)
+        use_mods.append_flags("LDFLAGS", use_flag)
+        cmake.add_default_envmod(use_mods)
+
+    def setup_build_environment(self, env):
+        if "%apple-clang" in self.spec:
+            env.append_flags("CFLAGS", "-mmacosx-version-min=10.13")
+            env.append_flags("CXXFLAGS", "-mmacosx-version-min=10.13")
+            env.append_flags("LDFLAGS", "-mmacosx-version-min=10.13")
+        elif self.spec.compiler.name in ("gcc", "clang") and "+static_libstdcpp" in self.spec:
+            env.append_flags("LDFLAGS", "-static-libstdc++ -static-libgcc -Wl,--exclude-libs,ALL")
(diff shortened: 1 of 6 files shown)


var/spack/repos/builtin/packages/clingo-bootstrap/package.py
```diff
@@ -49,15 +69,45 @@ class ClingoBootstrap(Clingo):
...
         return self.define("CLINGO_BUILD_PY_SHARED", "OFF")
 
-    def cmake_args(self):
-        args = super(ClingoBootstrap, self).cmake_args()
-        args.extend(
-            [
-                # Avoid building the clingo executable
-                self.define("CLINGO_BUILD_APPS", "OFF"),
-            ]
+    @when("+optimized")
```
haampie: This seems to have no effect?

haampie: Didn't you fix this @becker33 ?

alalazo: Try:
```
@run_before("cmake", when="+optimized")
```
(the change was part of multiple build-system).

alecbcs:
@spackbot run pipeline

haampie:
Some more timings.

Summary

1. Python 3.11 with pgo/lto helps a lot
2. `clingo` regressed significantly in https://github.com/potassco/clingo/commit/f92009c4933e88a2f47cd251d55e19e25747c859
3. No significant difference between clang and gcc here. 

**system python 3.10.7, clingo-bootstrap@spack official binaries**

    setup          4.113s
    load           0.018s
    ground         1.848s
    solve          2.523s
    total          8.698s

**clingo-bootstrap@spack ~optimizations %gcc ^python@3.11 +optimizations**

    setup          3.381s
    load           0.018s
    ground         1.802s
    solve          2.595s
    total          7.963s

**clingo-bootstrap@spack +optimizations %clang ^python@3.11 +optimizations** :medal_sports: 

    setup          3.389s
    load           0.018s
    ground         1.247s
    solve          1.792s
    total          6.615s

**clingo-bootstrap@spack +optimizations %gcc ^python@3.11 +optimizations** :medal_sports: 

    setup          3.372s
    load           0.015s
    ground         1.234s
    solve          1.872s
    total          6.657s

**clingo-bootstrap@5.6.2 +optimizations %clang ^python@3.11 +optimizations**

    setup          3.965s
    load           0.015s
    ground         1.434s
    solve          1.798s
    total          7.393s

**clingo-bootstrap@5.6.2 +optimizations %gcc ^python@3.11 +optimizations**

    setup          3.951s
    load           0.015s
    ground         1.567s
    solve          1.870s
    total          7.582s

haampie:
@trws fyi, mimalloc works "by chance" if you dynamically link libstdc++.so: operator new/delete symbols must be located in libclingo first, so if we would use another Python extension that links to libstdc++ and it gets imported earlier than clingo, the operator new/delete symbols are no longer overridden.

For bootstrapping though, we statically link libstdc++, so all is good. (Well, not 100% sure if we control the order on the link line `-lmimalloc -lstdc++`, but right now that is the case, fortunately)

haampie:
I've updated this locally with changes for macOS, but without mimalloc.

haampie:
Might be worth bisecting Spack's commits. The 20 - 30% reduction in solve time still exists today, but hdf5 solve times have doubled between 68ce49a57b5d8727fa0503d97cf3def3462c44e1 (at the moment I opened the PR) and f9457fa80ba51d14610df8b57a6b5a514d6bb6ff now (develop at the time of writing).

Edit: bisected to 28c4809a8fea6f51217bca105c8ad19a62a2c5ae... before that the solve time is 2.1s, after 3.4s :thinking: 

Edit 2: right, `spack solve --timers --fresh freetype` goes from *2.4ms* solve time to *2.8s*, so more than a 1000x slower, due to cmake as a possible build dep. Some improvements in #39360

haampie:
More speed comparisons:


```
 Row │ pkg            before_total  after_total  speedup_total  before_clingo  after_clingo  speedup_clingo 
─────┼──────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ aluminum             17.425       13.748          1.267         13.152         9.986           1.317
   2 │ ascent               37.138       30.863          1.203         26.208        20.501           1.278
   3 │ axom                 35.695       29.129          1.225         25.882        20.067           1.29
   4 │ blt                  14.406       11.585          1.244         10.779         8.08            1.334
   5 │ caliper              18.247       15.092          1.209         14.392        11.6             1.241
   6 │ care                 15.591       12.941          1.205         11.225         8.851           1.268
   7 │ chai                 15.409       12.514          1.231         10.972         8.679           1.264
   8 │ conduit              20.319       17.541          1.158         16.197        13.644           1.187
   9 │ dihydrogen           19.39        15.956          1.215         14.975        11.887           1.26
  10 │ flux-core            16.974       13.848          1.226         13.293        10.391           1.279
  11 │ flux-sched           17.703       14.135          1.252         13.903        10.775           1.29
  12 │ flux-security        15.273       12.622          1.21          11.647         9.358           1.245
  13 │ glvis                35.697       28.373          1.258         26.642        19.988           1.333
  14 │ hydrogen             17.196       14.357          1.198         13.096        10.515           1.245
  15 │ hypre                23.511       19.515          1.205         18.851        15.065           1.251
  16 │ lbann                34.321       27.585          1.244         24.901        19.089           1.304
  17 │ lvarray              19.693       16.001          1.231         14.283        11.1             1.287
  18 │ mfem                 33.937       28.696          1.183         25.004        20.389           1.226
  19 │ py-hatchet           25.908       20.835          1.243         21.587        16.757           1.288
  20 │ py-maestrowf         18.127       14.746          1.229         13.837        11.203           1.235
  21 │ py-merlin            19.904       15.984          1.245         15.839        12.297           1.288
  22 │ py-shroud            15.383       12.473          1.233         11.822         9.132           1.295
  23 │ raja                 15.219       12.199          1.248         11.281         8.522           1.324
  24 │ samrai               19.468       15.218          1.279         15.711        11.508           1.365
  25 │ scr                  19.318       14.962          1.291         15.359        11.412           1.346
  26 │ sundials             30.22        25.351          1.192         22.17         17.965           1.234
  27 │ umpire               14.815       11.796          1.256         10.712         8.298           1.291
  28 │ visit                34.759       28.839          1.205         25.371        19.684           1.289
  29 │ xbraid               18.273       14.179          1.289         14.377        10.9             1.319
  30 │ zfp                  13.953       11.151          1.251         10.249         7.844           1.307
```

seems pretty consistent.

On average, clingo runs 1.28x faster, and including Python overhead it's on average 1.23x faster.

haampie:
Regarding mimalloc: I think our best bet is to (a) statically link `libstdc++.so`, (b) use the mimalloc object file (or equivalently include `mimalloc-new-delete.h` in one source file), and (c) ensure all symbols are local except for `clingo_*` etc.

With shared libstdc++ other override tricks may not work because they depend on the load order.

Alternatively, using a custom allocator in clingo is a major refactor, not great...

trws:
FWIW, truly static libc++ is much easier than truly static libstdc++. Not sure if we expose it, but there's literally a flag on the llvm cmake to build a truly static libc++, makes it a lot nicer to deal with.

haampie:
Also FWIW, the issue being solved is mainly how to override `operator new/delete` _locally_. Some approaches:

1. `libclingo.so` with mimalloc `operator new` and shared libstdc++ works generally (libstdc++ resolves internal operator new also in libclingo.so), but (a) if it works it overrides `operator new` globally, (b) it depends on load order, which ultimately comes down to python import order.
2. `libclingo.so` with mimalloc `operator new` and static libstdc++ also seems to work (no duplicate symbol issues), except that mimalloc static lib contains symbols with global visibility (maybe I'm missing a linker flag to avoid that), so other libraries that need libstdc++ may bind operator new to the on in libclingo.so
3. `libclingo.so` with mimalloc `operator new` and static libstdc++ and a version/linker script that hides all non-clingo symbols; should be robust

trws:
Option 3, 100% no question, whichever C++ stdlib it is.  That's the approach we use for a lot of tool libs, and the approach used by nvidia for libcudart.  It's reliable, and keeps everything segregated.  The only trick will be if we find out that something does unclean things with the system allocator between libclingo and outside.  Here's hoping not.

It's worth noting that there are also multiple memory pools in libstdc++, the string one has bitten me in the past when trying to do things like this.  Not sure if libc++ has the same problem or not.

haampie:
Nice, so, yeah, that's what the PR currently does :upside_down_face: I don't think libclingo resources are managed outside of libclingo, or at least I don't expect it, since the only interface is with python 

I could also add your upstream hash function improvements as a patch for the latest clingo version.

haampie:
I've upstreamed the version script change and [that got merged](https://github.com/potassco/clingo/pull/449)

rkaminsk:
@haampie not sure exactly how spack works but if you are processing a lot of symbols, there might also be some overhead simply because a lot of functions a called. Standard python does not inline functions. I have sometimes seen huge performance improvements switching to pypy. At the expense of ugly code, one could probably also do the inlining manually.

haampie:
I believe that's the reason we're stuck at commit https://github.com/potassco/clingo/commit/2a025667090d71b2c9dce60fe924feb6bde8f667, one commit before cffi. With cffi it's about a 10% performance regression due to many C calls, if I remember correctly.

rkaminsk:
> I believe that's the reason we're stuck at commit [potassco/clingo@2a02566](https://github.com/potassco/clingo/commit/2a025667090d71b2c9dce60fe924feb6bde8f667), one commit before cffi. With cffi it's about a 10% performance regression due to many C calls, if I remember correctly.

There is still a bit of optimization potential here by inlining `_c_call` and `_c_call2` from the `_internal` module in critical places. Also checking the result before calling `_handle_error` should give a small boost. There is a penalty as compared to the previous implementation. Classes implemented via the C-API have less overhead as compared to classes declared in Python. However, the maintenance burden of an extension based on CPython is just too high. Maybe Cython would be a faster alternative. At least for the standard Python implementation...

trws:
At least for us, most of the overhead last I checked was adding clauses one at a time. If that's still the case, the issue wasn't so much the call overhead as it was allocating in python, populating in python, then allocating and translating into C++. If we could bulk-add somehow or directly generate the c++ objects as we go rather than building a python object that's then converted that might be a bigger overall win. Even going back to generating it as a big string and passing it off to clingo might do it TBH.

rkaminsk:
> At least for us, most of the overhead last I checked was adding clauses one at a time. If that's still the case, the issue wasn't so much the call overhead as it was allocating in python, populating in python, then allocating and translating into C++. If we could bulk-add somehow or directly generate the c++ objects as we go rather than building a python object that's then converted that might be a bigger overall win. Even going back to generating it as a big string and passing it off to clingo might do it TBH.

I see. Some other possible options:
1. You could write the facts to a temporary file and then use `Control.load`. (This is even something that could be sped up on the clingo side by adding a specialized fact reader.)
2. Another option would be to avoid the Python representation of the Symbol and use cffi directly. For symbols it should not be that complicated.
3. Writing a custom C function for this part would also be an option. It has the highest maintenance/porting overhead, however.

alalazo:
Point 1 above shouldn't be hard to implement. We always generate facts going through the same object, so we can cache all the facts we want to generate and dump them to a file, instead of adding them one by one with this function:

https://github.com/spack/spack/blob/c882214273793c52f2865e068aa0499c610a2f37/lib/spack/spack/solver/asp.py#L695-L712

haampie:
Sounds good, we can do that in another PR.

In the meantime, can this get merged? :upside_down_face:

alalazo (commented):
The PR LGTM. I'm going to benchmark it on a few platforms against the `radiuss` packages, and post the results.

alalazo:
Benchmark on linux:

* **Spack:** 0.21.0.dev0 (0e27f0561182e21c838606074ff3eeba0543d6ce)
* **Python:** 3.8.10
* **Concretizer:** clingo


[radiuss-pr.csv](https://github.com/spack/spack/files/12405537/radiuss-pr.csv)
[radiuss-develop.csv](https://github.com/spack/spack/files/12405539/radiuss-develop.csv)
[radiuss.txt](https://github.com/spack/spack/files/12405540/radiuss.txt)

radiuss

<details>

<summary>Table results</summary>

| Spec solved   |   Total time [develop, secs.] |   Total time [PR, secs.] |   Speedup [%] |
|---------------|-------------------------------|--------------------------|---------------|
| aluminum      |                      11.6029  |                  9.73918 |       16.0623 |
| ascent        |                      30.0139  |                 26.0173  |       13.3161 |
| axom          |                      29.4293  |                 24.5493  |       16.5822 |
| blt           |                       2.80539 |                  2.07582 |       26.0058 |
| caliper       |                      12.6565  |                 10.1348  |       19.9242 |
| care          |                      11.8909  |                  9.96472 |       16.1991 |
| chai          |                      11.3995  |                  9.4142  |       17.4157 |
| conduit       |                      12.4496  |                 10.1501  |       18.4703 |
| dihydrogen    |                      12.7949  |                 10.9246  |       14.6176 |
| flux-core     |                      10.8858  |                  9.03948 |       16.9611 |
| flux-sched    |                      11.0894  |                  9.19394 |       17.0923 |
| flux-security |                       2.89337 |                  2.20923 |       23.6449 |
| glvis         |                      26.8223  |                 23.1809  |       13.5762 |
| hiptt         |                       9.60946 |                  8.04974 |       16.2311 |
| hydrogen      |                      12.0701  |                 10.0883  |       16.4193 |
| hypre         |                      14.0524  |                 12.1672  |       13.4158 |
| lbann         |                      30.2002  |                 26.1254  |       13.4925 |
| lvarray       |                      16.93    |                 13.9645  |       17.516  |
| mfem          |                      26.6137  |                 22.9639  |       13.7141 |
| py-hatchet    |                      13.6504  |                 10.9445  |       19.8231 |
| py-maestrowf  |                       3.84452 |                  3.12878 |       18.6172 |
| py-merlin     |                      12.0912  |                  9.87339 |       18.3426 |
| py-shroud     |                       2.96705 |                  2.18109 |       26.4897 |
| raja          |                      10.9228  |                  8.77776 |       19.6382 |
| samrai        |                      10.835   |                  9.05608 |       16.4185 |
| scr           |                      11.5734  |                  9.53503 |       17.6128 |
| sundials      |                      23.8507  |                 20.5012  |       14.0436 |
| umpire        |                      10.3822  |                  8.94438 |       13.8493 |
| visit         |                      28.1617  |                 24.0164  |       14.7195 |
| xbraid        |                      11.0041  |                  9.1531  |       16.8211 |
| zfp           |                       9.9177  |                  8.23858 |       16.9305 |

</details>

alalazo:
Benchmark on darwin:

* **Spack:** 0.21.0.dev0 (0e27f0561182e21c838606074ff3eeba0543d6ce)
* **Python:** 3.9.6
* **Platform:** darwin-ventura-m1
* **Concretizer:** clingo

[radiuss.txt](https://github.com/spack/spack/files/12408513/radiuss.txt)
[radiuss-pr.csv](https://github.com/spack/spack/files/12408514/radiuss-pr.csv)
[radiuss-develop.csv](https://github.com/spack/spack/files/12408515/radiuss-develop.csv)

radiuss

<details>

<summary>Table details</summary>

| Spec solved   |   Total time [develop, secs.] |   Total time [PR, secs.] |   Speedup [%] |
|---------------|-------------------------------|--------------------------|---------------|
| aluminum      |                       4.92675 |                  4.64714 |       5.67531 |
| ascent        |                      14.8728  |                 13.7688  |       7.42297 |
| axom          |                      12.7773  |                 12.0498  |       5.69321 |
| blt           |                       1.15733 |                  1.09977 |       4.97316 |
| caliper       |                       4.52976 |                  4.28288 |       5.45007 |
| care          |                       5.31877 |                  5.12422 |       3.65771 |
| chai          |                       4.72034 |                  4.55208 |       3.56452 |
| conduit       |                       4.98953 |                  4.82712 |       3.25506 |
| dihydrogen    |                       5.43769 |                  5.18232 |       4.6963  |
| flux-security |                       1.19515 |                  1.17156 |       1.97383 |
| glvis         |                      13.1791  |                 12.0497  |       8.57012 |
| hiptt         |                       4.29786 |                  3.90945 |       9.03726 |
| hydrogen      |                       5.38992 |                  4.84296 |      10.1479  |
| hypre         |                       6.00553 |                  5.58805 |       6.95168 |
| lbann         |                      13.4174  |                 12.4325  |       7.34086 |
| lvarray       |                       7.53549 |                  6.95636 |       7.68542 |
| mfem          |                      12.1974  |                 11.5786  |       5.07297 |
| py-hatchet    |                       5.27827 |                  4.95733 |       6.08039 |
| py-maestrowf  |                       1.6974  |                  1.53959 |       9.29704 |
| py-merlin     |                       4.95965 |                  4.64712 |       6.3015  |
| py-shroud     |                       1.16995 |                  1.13006 |       3.4097  |
| raja          |                       4.43408 |                  4.15822 |       6.22138 |
| samrai        |                       4.55573 |                  4.27484 |       6.16559 |
| scr           |                       4.64586 |                  4.36847 |       5.97052 |
| sundials      |                      10.318   |                  9.78748 |       5.14168 |
| umpire        |                       4.34871 |                  4.09094 |       5.92751 |
| visit         |                      15.5421  |                 14.2304  |       8.43981 |
| xbraid        |                       4.43679 |                  4.1789  |       5.81235 |
| zfp           |                       4.11844 |                  3.94556 |       4.19767 |

</details>

alalazo (approved):
Speed-up is much better on x86_64 than Apple M1, but in both cases :+1:

PR #34930, opened 2023-01-13
Port basic Paraview deps to Windows

johnwparent:
Port:

- libtiff
- freetype
- proj
- expat
- libpng
- adios2
- lz4
- nasm
- zlib
- zstd
To Windows support and to the new multi builder system as relevant.

PR #34932, opened 2023-01-13
Windows: port lower level pkgs

### LINKED ISSUE #34938: Paraview Support on Windows
### Summary

johnwparent:
Supporting package ports/amendments for the port of Paraview to Windows. 

Includes changes to:
- wgl
- msmpi
- hdf5
- apple-gl

Also sets wgl as default opengl provider for Windows

### DIFF
--- a/etc/spack/defaults/windows/packages.yaml
+++ b/etc/spack/defaults/windows/packages.yaml
@@ -19,3 +19,4 @@ packages:
     - msvc
     providers:
       mpi: [msmpi]
+      gl: [wgl]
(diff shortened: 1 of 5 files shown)


bernhardkaindl (changes requested):
Like Adam pointed out in #34930, this too does not fulfill the "one package per PR" policy:
> These changes are all independent and belong in separate PRs for each package. Spack has a strict "one package per PR" policy unless the changes to one package depend on the changes to another package.

The OpenGL package changes have no obvious connection to the MPI changes, please split them into new PRs.

PR #34933, opened 2023-01-13
Python: multi-build system port

### LINKED ISSUE #34938: Paraview Support on Windows

johnwparent:
Python - to multi builder system and other relevant changes to support paraview

### DIFF
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -234,7 +234,6 @@ class Python(AutotoolsPackage, Package):
 
     executables = [r"^python[\d.]*[mw]?$"]
 
-
     @classmethod
     def determine_version(cls, exe):
         # Newer versions of Python support `--version`,
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -1074,6 +1074,10 @@ class AutotoolsBuilder(AutotoolsBuilder, RunAfter, BuildEnvironment):
 
         return config_args
 
+    def install(self, pkg, spec, prefix):
+        with working_dir(self.build_directory):
+            make(*self.install_targets, parllel=False)
+
 
 class GenericBuilder(GenericBuilder, RunAfter, BuildEnvironment):
     phases = ("build", "install")
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -876,7 +876,7 @@ class RunAfter:
             return
         kwargs = {"ignore_absent": True, "backup": False, "string": True}
 
-        filenames = [self.pkg.get_sysconfigdata_name(), self.config_vars["makefile_filename"]]
+        filenames = [self.pkg.get_sysconfigdata_name(), self.pkg.config_vars["makefile_filename"]]
 
         filter_file(spack_cc, self.pkg.compiler.cc, *filenames, **kwargs)
         if spack_cxx and self.pkg.compiler.cxx:
@@ -901,7 +901,7 @@ class RunAfter:
         # https://devguide.python.org/gdb/
         src = os.path.join("Tools", "gdb", "libpython.py")
         if os.path.exists(src):
-            install(src, self.command.path + "-gdb.py")
+            install(src, self.pkg.command.path + "-gdb.py")
 
     @run_after("install")
     @on_package_attributes(run_tests=True)
@@ -913,66 +913,66 @@ class RunAfter:
         with working_dir("spack-test", create=True):
             # Ensure that readline module works
             if "+readline" in spec:
-                self.command("-c", "import readline")
+                self.pkg.command("-c", "import readline")
 
             # Ensure that ssl module works
             if "+ssl" in spec:
-                self.command("-c", "import ssl")
-                self.command("-c", "import hashlib")
+                self.pkg.command("-c", "import ssl")
+                self.pkg.command("-c", "import hashlib")
 
             # Ensure that sqlite3 module works
             if "+sqlite3" in spec:
-                self.command("-c", "import sqlite3")
+                self.pkg.command("-c", "import sqlite3")
 
             # Ensure that dbm module works
             if "+dbm" in spec:
-                self.command("-c", "import dbm")
+                self.pkg.command("-c", "import dbm")
 
             # Ensure that nis module works
             if "+nis" in spec:
-                self.command("-c", "import nis")
+                self.pkg.command("-c", "import nis")
 
             # Ensure that zlib module works
             if "+zlib" in spec:
-                self.command("-c", "import zlib")
+                self.pkg.command("-c", "import zlib")
 
             # Ensure that bz2 module works
             if "+bz2" in spec:
-                self.command("-c", "import bz2")
+                self.pkg.command("-c", "import bz2")
 
             # Ensure that lzma module works
             if "+lzma" in spec:
-                self.command("-c", "import lzma")
+                self.pkg.command("-c", "import lzma")
 
             # Ensure that pyexpat module works
             if "+pyexpat" in spec:
-                self.command("-c", "import xml.parsers.expat")
-                self.command("-c", "import xml.etree.ElementTree")
+                self.pkg.command("-c", "import xml.parsers.expat")
+                self.pkg.command("-c", "import xml.etree.ElementTree")
 
             # Ensure that ctypes module works
             if "+ctypes" in spec:
-                self.command("-c", "import ctypes")
+                self.pkg.command("-c", "import ctypes")
 
             # Ensure that tkinter module works
             # https://wiki.python.org/moin/TkInter
             if "+tkinter" in spec:
                 # Only works if ForwardX11Trusted is enabled, i.e. `ssh -Y`
                 if "DISPLAY" in env:
-                    self.command("-c", "import tkinter; tkinter._test()")
+                    self.pkg.command("-c", "import tkinter; tkinter._test()")
                 else:
-                    self.command("-c", "import tkinter")
+                    self.pkg.command("-c", "import tkinter")
 
             # Ensure that uuid module works
             if "+uuid" in spec:
-                self.command("-c", "import uuid")
+                self.pkg.command("-c", "import uuid")
 
             # Ensure that tix module works
             if "+tix" in spec:
-                self.command("-c", "import tkinter.tix")
+                self.pkg.command("-c", "import tkinter.tix")
 
             # Ensure that crypt module works
             if "+crypt" in spec:
-                self.command("-c", "import crypt")
+                self.pkg.command("-c", "import crypt")
 
 
 class AutotoolsBuilder(autotools.AutotoolsBuilder, RunAfter, BuildEnvironment):
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -908,7 +908,6 @@ class RunAfter(object):
                 os.path.join(prefix.bin, "python-config"),
             )
 
-
     @run_after("install")
     def install_python_gdb(self):
         # https://devguide.python.org/gdb/
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -343,6 +343,11 @@ class Python(AutotoolsPackage, Package):
         """Returns the Python command, which may vary depending
         on the version of Python and how it was installed.
 
+        In general, Python 3 only comes with a ``python3`` command. However, some
+        package managers will symlink ``python`` to ``python3``, while others
+        may contain ``python3.11``, ``python3.10``, and ``python3.9`` in the
+        same directory.
+
         Returns:
             Executable: the Python command
         """
@@ -365,6 +370,12 @@ class Python(AutotoolsPackage, Package):
                 return Executable(path)
 
         else:
+            # Give a last try at rhel8 platform python
+            if self.spec.external and self.prefix == "/usr" and self.spec.satisfies("os=rhel8"):
+                path = os.path.join(self.prefix, "libexec", "platform-python")
+                if os.path.exists(path):
+                    return Executable(path)
+
             msg = "Unable to locate {0} command in {1}"
             raise RuntimeError(msg.format(self.name, self.prefix.bin))
 
@@ -1060,7 +1071,7 @@ class AutotoolsBuilder(autotools.AutotoolsBuilder, RunAfter, BuildEnvironment):
     def install(self, pkg, spec, prefix):
         with working_dir(self.build_directory):
             # See https://github.com/python/cpython/issues/102007
-            make(*self.install_targets, parllel=False)
+            make(*self.install_targets, parallel=False)
 
 
 class GenericBuilder(generic.GenericBuilder, RunAfter, BuildEnvironment):
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -231,9 +231,6 @@ class Python(AutotoolsPackage, Package):
     # Used to cache various attributes that are expensive to compute
     _config_vars: Dict[str, Dict[str, str]] = {}
 
-    # An in-source build with --enable-optimizations fails for python@3.X
-    build_directory = "spack-build"
-
     executables = [r"^python[\d.]*[mw]?$"]
 
     build_system(conditional("generic", when="platform=windows"), "autotools", default="autotools")
@@ -992,6 +989,9 @@ class RunAfter(object):
 
 
 class AutotoolsBuilder(AutotoolsBuilder, RunAfter, BuildEnvironment):
+    # An in-source build with --enable-optimizations fails for python@3.X
+    build_directory = "spack-build"
+
     def configure_args(self):
         spec = self.spec
         config_args = []
@@ -1035,7 +1035,7 @@ class AutotoolsBuilder(AutotoolsBuilder, RunAfter, BuildEnvironment):
         config_args.append("--without-ensurepip")
 
         if "+pic" in spec:
-            cflags.append(self.compiler.cc_pic_flag)
+            cflags.append(self.pkg.compiler.cc_pic_flag)
 
         if "+ssl" in spec:
             config_args.append("--with-openssl={0}".format(spec["openssl"].prefix))
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -1074,7 +1074,7 @@ class AutotoolsBuilder(RunAfter, BuildEnvironment, autotools.AutotoolsBuilder):
             make(*self.install_targets, parallel=False)
 
 
-class GenericBuilder(RunAfter, BuildEnvironment, generic.GenericBuilder,):
+class GenericBuilder(RunAfter, BuildEnvironment, generic.GenericBuilder):
     phases = ("build", "install")
 
     @property
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -17,8 +17,7 @@ from llnl.util.filesystem import is_nonsymlink_exe_with_shebang, path_contains_s
 from llnl.util.lang import dedupe
 
 from spack.build_environment import dso_suffix, stat_suffix
-from spack.build_systems.autotools import AutotoolsBuilder
-from spack.build_systems.generic import GenericBuilder
+from spack.build_systems import autotools, generic
 from spack.package import *
 from spack.util.environment import is_system_path
 from spack.util.prefix import Prefix
@@ -105,6 +104,8 @@ class Python(AutotoolsPackage, Package):
     version("3.7.2", sha256="f09d83c773b9cc72421abba2c317e4e6e05d919f9bcf34468e192b6a6c8e328d")
     version("3.7.1", sha256="36c1b81ac29d0f8341f727ef40864d99d8206897be96be73dc34d4739c9c9f06")
     version("3.7.0", sha256="85bb9feb6863e04fb1700b018d9d42d1caac178559ffa453d7e6a436e259fd0d")
+
+    build_system(conditional("generic", when="platform=windows"), "autotools", default="autotools")
 
     extendable = True
 
@@ -233,7 +234,6 @@ class Python(AutotoolsPackage, Package):
 
     executables = [r"^python[\d.]*[mw]?$"]
 
-    build_system(conditional("generic", when="platform=windows"), "autotools", default="autotools")
 
     @classmethod
     def determine_version(cls, exe):
@@ -340,20 +340,10 @@ class Python(AutotoolsPackage, Package):
         # allow flags to be passed through compiler wrapper
         return (flags, None, None)
 
-    # ========================================================================
-    # Set up environment to make install easy for python extensions.
-    # ========================================================================
-
     @property
     def command(self):
         """Returns the Python command, which may vary depending
         on the version of Python and how it was installed.
-
-        In general, Python 2 comes with ``python`` and ``python2`` commands,
-        while Python 3 only comes with a ``python3`` command. However, some
-        package managers will symlink ``python`` to ``python3``, while others
-        may contain ``python3.6``, ``python3.5``, and ``python3.4`` in the
-        same directory.
 
         Returns:
             Executable: the Python command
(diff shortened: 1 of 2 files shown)


adamjstewart:
@spackbot fix style

johnwparent:
@adamjstewart Thanks for running that, I knew there was a command to that effect, I just couldn't recall the exact phrasing. Is there a doc with all Spackbot's functionality (presumably the spackbot repo, if that's public)?

adamjstewart:
There are a few places to look:

* Repo: https://github.com/spack/spackbot
* Docs: https://spack.github.io/spackbot/

or simply run one of the following:

@spackbot help
@spackbot commands

johnwparent:
Super cool, thanks!

adamjstewart:
@alalazo is there documentation or source code somewhere where I can find a list of which methods/attributes belong in the builder and which belong in the package? Our Python package uses just about every possible feature and I'm not sure where some of them belong.

With bd860216aaa0a47814014a326b218b547001f52b I'm now able to build this on macOS successfully so this PR looks good to me. Let's hope all the CI tests pass.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -21,14 +20,16 @@
 from llnl.util.lang import dedupe
 
 from spack.build_environment import dso_suffix, stat_suffix
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.generic import GenericBuilder
```
alalazo: I'd avoid this, for clarity. Can we do:
```python
from spack.build_system import autotools, generic
```
and below use e.g. `autotools.AutoolsBuilder` for the base class? That would make the name unambiguous and reduce the amount of typing wrt fully qualified namespaces.

alalazo:
@adamjstewart I'll give a more thorough review to this one during this week. Let me know if this is a blocker for anything and I'll try to prioritize it.

var/spack/repos/builtin/packages/python/package.py
adamjstewart: I don't really see it as ambiguous and this is the syntax I've used in the packages I've written. I've never needed to access a property of the super class, and if I did, I'd probably use super().

alalazo: You rebind a name that was imported from another module. I find that ambiguous,  since the same name is set two times in the same global scope. The change proposed above reduce the typing to a minimum and removes the ambiguity, so I'd go for that.

johnwparent:
@alalazo @scheibelp suggested I ping you on this once I had rebased on develop. That's been done, and baring any related failures, this is ready for review.

var/spack/repos/builtin/packages/python/package.py
alalazo:
```suggestion
from spack.build_systems import autotools, generic
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1201,3 +751,435 @@ def test(self):
...
+            # Ensure that tix module works
+            if "+tix" in spec:
+                self.command("-c", "import tkinter.tix")
+
+            # Ensure that crypt module works
+            if "+crypt" in spec:
+                self.command("-c", "import crypt")
+
+
+class AutotoolsBuilder(AutotoolsBuilder, RunAfter, BuildEnvironment):
```
alalazo:
```suggestion
class AutotoolsBuilder(autotools.AutotoolsBuilder, RunAfter, BuildEnvironment):
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1201,3 +751,435 @@ def test(self):
...
+            config_args.append("CFLAGS={0}".format(" ".join(cflags)))
+
+        return config_args
+
+    def install(self, pkg, spec, prefix):
+        with working_dir(self.build_directory):
+            make(*self.install_targets, parllel=False)
+
+
+class GenericBuilder(GenericBuilder, RunAfter, BuildEnvironment):
```
alalazo:
```suggestion
class GenericBuilder(generic.GenericBuilder, RunAfter, BuildEnvironment):
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -222,14 +217,15 @@ class Python(Package):
...
+
     # Used to cache various attributes that are expensive to compute
     _config_vars: Dict[str, Dict[str, str]] = {}
 
-    # An in-source build with --enable-optimizations fails for python@3.X
-    build_directory = "spack-build"
-
     executables = [r"^python[\d.]*[mw]?$"]
 
+    build_system(conditional("generic", when="platform=windows"), "autotools", default="autotools")
```
alalazo: Nitpick: can we move this to the top, just after versions?

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -359,415 +331,78 @@ def flag_handler(self, name, flags):
...
-        """
-
-        arch_map = {"AMD64": "x64", "x86": "Win32", "IA64": "Win32", "EM64T": "Win32"}
-        arch = platform.machine()
-        if arch in arch_map:
-            arch = arch_map[arch]
-        return arch
+    # ========================================================================
+    # Set up environment to make install easy for python extensions.
+    # ========================================================================
```
alalazo: I'm always a bit skeptical when we divide methods / classes in a file according to a comment, since it's very likely to lose meaning in time[^1]

[^1]: Speaking for myself, for instance, I rarely read an entire file to check if somebody put some section-like comment in it

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -359,415 +331,78 @@ def flag_handler(self, name, flags):
...
+        on the version of Python and how it was installed.
 
-    def win_installer(self, prefix):
-        """
-        Python on Windows does not export an install target
-        so we must handcraft one here. This structure
-        directly mimics the install tree of the Python
-        Installer on Windows.
+        In general, Python 2 comes with ``python`` and ``python2`` commands,
+        while Python 3 only comes with a ``python3`` command. However, some
```
alalazo: Is this part relevant / true ? For what is worth, virtual environment provide both a `python` and a `python3` command for `python@3:`.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -359,415 +331,78 @@ def flag_handler(self, name, flags):
...
-                [
-                    "--with-tcltk-includes=-I{0} -I{1}".format(
-                        spec["tcl"].prefix.include, spec["tk"].prefix.include
-                    ),
-                    "--with-tcltk-libs={0} {1}".format(
-                        spec["tcl"].libs.ld_flags, spec["tk"].libs.ld_flags
-                    ),
-                ]
-            )
+        .. code-block:: console
```
alalazo, line 389: The code block below:
```
# https://docs.python.org/3.8/library/sqlite3.html#f1
if spec.satisfies("+sqlite3 ^sqlite+dynamic_extensions"):
    config_args.append("--enable-loadable-sqlite-extensions")
```
is completely unrelated. Should have been:
```
python -m sysconfig
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1201,3 +751,435 @@ def test(self):
         self.run_test(
             exe, options=options, expected=[self.spec.prefix], installed=True, purpose=reason
         )
+
+
+class BuildEnvironment(object):
```
alalazo: Nitpick (we are not supporting Python 2.7 anymore):
```suggestion
class BuildEnvironment:
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1201,3 +751,435 @@ def test(self):
...
+    def setup_build_environment(self, env):
```
alalazo, line 761: I assume these functions have just been moved, so didn't go through them in details. Let me know if there is any change I should have a look to.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1201,3 +751,435 @@ def test(self):
...
+
+            # There is logic in the sysconfig module that is sensitive to the
+            # fact that LDSHARED is set in the environment, therefore we export
+            # the variable only if the new value is different from what we got
+            # from the sysconfigdata file:
+            if config_link != new_link and sys.platform != "win32":
+                env.set(link_var, new_link)
+
+
+class RunAfter(object):
```
alalazo:
```suggestion
class RunAfter:
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1201,3 +751,435 @@ def test(self):
...
+        if self.spec.satisfies("~ssl"):
+            args.append("--no-ssl")
+        if self.spec.satisfies("~tkinter"):
+            args.append("--no-tkinter")
+        return args
+
+    def build(self, pkg, spec, prefix):
+        """Makes the build targets specified by
+        :py:attr:``~.AutotoolsPackage.build_targets``
+        """
```
alalazo: This comment doesn't seem correct (refers to `AutotoolsPackage`):
```suggestion
```

alalazo (changes requested):
At line 143 we can use:
```python
with when("build_system=autotools"):
    depends_on("pkgconfig@0.9.0:", type="build")
    depends_on("gettext +libxml2", when="+libxml2")
    depends_on("gettext ~libxml2", when="~libxml2")
```
instead of relying on `sys.platform`. Also, can you double check if comments and docstrings throughout the file still make sense after moving methods?

alalazo:
Tried to install Python on linux, got
Screenshot from 2023-03-17 14-38-26

johnwparent:
> Tried to install Python on linux, got Screenshot from 2023-03-17 14-38-26

Thanks for running that, found a few more places where I missed referencing the package as well, will push those fixes with the rest of the changes. I'll spin up a linux machine so that doesn't happen again.

var/spack/repos/builtin/packages/python/package.py
johnwparent: Yeah, I'm not sure what happened with the comments... The merge conflicts were fairly messy.

var/spack/repos/builtin/packages/python/package.py
johnwparent: Right! I ported this before we could use new style classes, always super happy to see this suggestion!

var/spack/repos/builtin/packages/python/package.py
johnwparent: Just a straightforward copy/paste, this is several months old but IIRC I didn't touch any of this.

var/spack/repos/builtin/packages/python/package.py
johnwparent: Right! I ported this prior to the py2 drop and haven't revisited in a few months. Great!

var/spack/repos/builtin/packages/python/package.py
johnwparent: Merge conflict resolution issue. That block looks like it came from `configure_args`... I'll verify that that method contains what it's supposed to as well.

var/spack/repos/builtin/packages/python/package.py
johnwparent: I have no idea about the original context from the author, I did not write that comment, only copied it.
On Windows (on my system at least) there is only a `python` (no `python3`) regardless of venv, except for the Windows store nonsense. The linux machine I just spun up is the same as yours...
Are we even interested in continuing to consider python2?

var/spack/repos/builtin/packages/python/package.py
johnwparent: Sounds good!

var/spack/repos/builtin/packages/python/package.py
alalazo: Then let's just remove this part of the docstring

johnwparent: Sounds good to me!

var/spack/repos/builtin/packages/python/package.py
johnwparent: Just for my own edification, should I be placing this after the versions as a practice? I.e. in my several other package port PRs.

var/spack/repos/builtin/packages/python/package.py
johnwparent: This has lost meaning, it's basically describing the `BuildEnvironment` class, which IMO is pretty self evident as to its function. I'll remove from this part of the file.

var/spack/repos/builtin/packages/python/package.py
johnwparent: Confirmed that the `configure_args` have this stanza.

var/spack/repos/builtin/packages/python/package.py
alalazo: We have no strong rules on that, but I usually do it like this. The rationale is that I declare the build systems before eventually using:
```python
with when("build_system=xxx"):
```
and the build systems themselves are rarely conditional on other parts of the spec that are declared in `package.py`.

johnwparent:
Amended and pushed. Testing myself now on my linux machine.

var/spack/repos/builtin/packages/python/package.py
johnwparent: Makes sense to me, I'll start doing this from this point, thanks for taking the time to explain!

johnwparent:
@alalazo All comments have been addressed and it builds on the linux machine I provisioned (as well as still building on Windows).

adamjstewart:
Once @alalazo is happy with this I'll give one more detailed review from my side. Very busy at the moment so I'll wait for @alalazo to finish his review.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -360,710 +331,470 @@ def flag_handler(self, name, flags):
...
+    def home(self):
+        """Most of the time, ``PYTHONHOME`` is simply
+        ``spec['python'].prefix``. However, if the user is using an
+        externally installed python, it may be symlinked. For example,
+        Homebrew installs python in ``/usr/local/Cellar/python/2.7.12_2``
+        and symlinks it to ``/usr/local``. Users may not know the actual
+        installation directory and add ``/usr/local`` to their
+        ``packages.yaml`` unknowingly. Query the python executable to
+        determine exactly where it is installed.
+        """
```
alalazo: Not a request, since this was there also before this PR, but it is unclear to me what this one returns. Maybe:
```
"""Returns the real installation prefix of the current Python.

Most of the time, this would be the same as ``spec["python"].prefix`` ...
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -360,710 +331,470 @@ def flag_handler(self, name, flags):
...
+    # https://docs.python.org/3/library/sysconfig.html#installation-paths
+    # https://discuss.python.org/t/understanding-site-packages-directories/12959
+    # https://github.com/pypa/pip/blob/22.1/src/pip/_internal/locations/__init__.py
+    # https://github.com/pypa/installer/pull/103
 
-        with working_dir("spack-test", create=True):
-            # Ensure that readline module works
-            if "+readline" in spec:
-                self.command("-c", "import readline")
+    # NOTE: XCode Python's sysconfing module was incorrectly patched, and hard-codes
```
alalazo:
```suggestion
    # NOTE: XCode Python's sysconfig module was incorrectly patched, and hard-codes
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -360,710 +331,470 @@ def flag_handler(self, name, flags):
...
+        if path.startswith(prefix):
+            return path.replace(prefix, "")
+        return os.path.join("include", "python{}".format(self.version.up_to(2)))
 
-        In general, Python 3 only comes with a ``python3`` command. However, some
-        package managers will symlink ``python`` to ``python3``, while others
-        may contain ``python3.11``, ``python3.10``, and ``python3.9`` in the
-        same directory.
+    def setup_run_environment(self, env):
+        env.prepend_path("CPATH", os.pathsep.join(self.spec["python"].headers.directories))
```
alalazo: Why was it necessary to set `CPATH` in the run environment? I don't see this on `develop`.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1122,82 +853,317 @@ def setup_dependent_build_environment(self, env, dependent_spec):
...
 
-        # test hello world
-        msg = "hello world!"
-        reason = "test: running {0}".format(msg)
-        options = ["-c", 'print("{0}")'.format(msg)]
-        self.run_test(exe, options=options, expected=[msg], installed=True, purpose=reason)
+        if "+debug" in spec:
+            config_args.append("--with-pydebug")
+        else:
+            config_args.append("--without-pydebug")
```
alalazo: Here you can simplify a lot of boilerplate (beware of typos, I wrote this in the UI :slightly_smiling_face: ):
```suggestion
        config_args.extend(self.with_or_without("pydebug", variant="debug"))
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1122,82 +853,317 @@ def setup_dependent_build_environment(self, env, dependent_spec):
...
-        # checks import works and executable comes from the spec prefix
-        reason = "test: checking import and executable"
-        options = ["-c", "import sys; print(sys.executable)"]
-        self.run_test(
-            exe, options=options, expected=[self.spec.prefix], installed=True, purpose=reason
-        )
+        if "+shared" in spec:
+            config_args.append("--enable-shared")
+        else:
+            config_args.append("--disable-shared")
```
alalazo: Here you can:
```suggestion
        config_args.extend(self.enable_or_disable("shared"))
```
and likewise for the other options below, where it makes sense.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1122,82 +853,317 @@ def setup_dependent_build_environment(self, env, dependent_spec):
...
+            cflags.append("-fp-model=strict")
+
+        if cflags:
+            config_args.append("CFLAGS={0}".format(" ".join(cflags)))
+
+        return config_args
+
+    def install(self, pkg, spec, prefix):
+        with working_dir(self.build_directory):
+            make(*self.install_targets, parllel=False)
```
alalazo:
```suggestion
            # See https://github.com/python/cpython/issues/102007
            make(*self.install_targets, parallel=False)
```

alalazo (changes requested):
I have a few minor requests and some comments. @adamjstewart I think you can have a look too, so that @johnwparent can fix the remaining issues in one go :slightly_smiling_face:

var/spack/repos/builtin/packages/python/package.py
johnwparent: Considering `CPATH` isn't a consideration on Windows, I'm almost 100% sure I did not make this change (and I can't remember making it either), it's likely rebase/maintenance cruft since this change is so old. Removed.

johnwparent:
Waited a few weeks for additional review, but with none forthcoming, all outstanding suggestions have been addressed and the changes pushed. Please feel free to re-review @alalazo @scheibelp

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -369,845 +339,826 @@ def flag_handler(self, name, flags):
...
-        architecture support map
+    def command(self):
+        """Returns the Python command, which may vary depending
+        on the version of Python and how it was installed.
 
-        Note: This function really only makes
-        sense to use on Windows, could be overridden to
-        cross compile however.
+        Returns:
+            Executable: the Python command
```
alalazo, line 352: (nitpick) let's just remove these two lines, as they only repeat the first sentence in the docstring
```suggestion
```

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -369,845 +339,826 @@ def flag_handler(self, name, flags):
...
+        if path.startswith(prefix):
+            return path.replace(prefix, "")
+        return os.path.join("include", "python{}".format(self.version.up_to(2)))
 
-        else:
-            # Give a last try at rhel8 platform python
-            if self.spec.external and self.prefix == "/usr" and self.spec.satisfies("os=rhel8"):
-                path = os.path.join(self.prefix, "libexec", "platform-python")
-                if os.path.exists(path):
-                    return Executable(path)
```
alalazo, line 746: Can you please double check this file, comparing it to develop? This for instance was deleted in the current PR.

var/spack/repos/builtin/packages/python/package.py
alalazo: :point_up: This issue wasn't fixed

alalazo (changes requested):
CI is failing, since apparently some code got lost during a rebase. Can you please double check the methods that moved and ping when CI is passing?

var/spack/repos/builtin/packages/python/package.py
johnwparent: From my UI it appears as though it has been

image

alalazo: Sorry, the most important thing was the typo. That seems not fixed.

johnwparent: Honestly didn't even notice it. Amended.

var/spack/repos/builtin/packages/python/package.py
adamjstewart: I would vote to keep this. If we ever adopt pydocstyle it actually requires all methods to document their args and returns

alalazo: Ok, that's nitpick. Closing it.

johnwparent:
@alalazo I went over the changes by hand and I believe you caught the only significant rebase artifact remaining (I caught a couple dropped docstrings, but nothing else functional). CI seems to agree as the GH runs and the Gitlab python builds are passing.

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1132,82 +870,306 @@ def setup_dependent_build_environment(self, env, dependent_spec):
...
+            # Ensure that tix module works
+            if "+tix" in spec:
+                self.pkg.command("-c", "import tkinter.tix")
+
+            # Ensure that crypt module works
+            if "+crypt" in spec:
+                self.pkg.command("-c", "import crypt")
+
+
+class AutotoolsBuilder(autotools.AutotoolsBuilder, RunAfter, BuildEnvironment):
```
adamjstewart: According to #36825, the builder base class should actually come last (@alalazo what's the reason for this, and is it possible to make it order invariant?)

var/spack/repos/builtin/packages/python/package.py
```diff
@@ -1132,82 +870,306 @@ def setup_dependent_build_environment(self, env, dependent_spec):
...
+        doc_dir = os.path.join(proj_root, "Doc")
+        copy_tree(doc_dir, prefix.Doc)
+        tools_dir = os.path.join(proj_root, "Tools")
+        copy_tree(tools_dir, prefix.Tools)
+        lib_dir = os.path.join(proj_root, "Lib")
+        copy_tree(lib_dir, prefix.Lib)
+        pyconfig = os.path.join(proj_root, "PC", "pyconfig.h")
+        copy(pyconfig, prefix.include)
+        shared_libraries = []
+        shared_libraries.extend(glob.glob("%s\\*.exe" % build_root))
```
adamjstewart:
```suggestion
        shared_libraries.extend(glob.glob(r"%s\*.exe" % build_root))
```
If you make it a raw string you don't have to escape the directory path

var/spack/repos/builtin/packages/python/package.py
johnwparent: Will amend.

var/spack/repos/builtin/packages/python/package.py
johnwparent: Addressed.

adamjstewart (approved):
Didn't go over things with a fine-toothed comb, but it looks good to me at first glance. As long as the tests pass and @alalazo is happy then I'm happy.

Note that the default values returned by `config_vars` are designed for Unix and may not be appropriate on Windows. These default values are only used when we're unable to run `python -m sysconfig` for some reason (e.g., cross-compilation), so it may not really matter that much.

johnwparent:
> Note that the default values returned by `config_vars` are designed for Unix and may not be appropriate on Windows. These default values are only used when we're unable to run `python -m sysconfig` for some reason (e.g., cross-compilation), so it may not really matter that much.

We don't support cross compilation on Windows at the moment, and I can't think of another scenario where we couldn't run that module, however FWIW on Windows we don't care about the vast majority of those config settings, and those that we do/need to use, either work as is or have handlers specifically for Windows (see `find_library` for a good example of this).

adamjstewart:
Is this still a WIP? We recently merged #41211 to add a gmake dep to Python but that's only needed because it doesn't subclass AutotoolsPackage like it should.

alalazo:
@johnwparent  I guess this one needs to be either closed, or migrated to the spack-packages repo?

PR #34935, opened 2023-01-13
Port netcdf-c to Windows

### LINKED ISSUE #34938: Paraview Support on Windows

johnwparent:
Ports of Netcdf-c to be compatible with Windows - end goal of supporting Paraview on Windows.


skosukhin:
I guess it makes sense to reference #33564 here.

zackgalbreath:
@spackbot run pipeline

scheibelp:
@spackbot run pipeline

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -142,12 +153,63 @@ class NetcdfC(AutotoolsPackage):
...
+            self.define("ENABLE_PARALLEL_TESTS", False),
+        ]
+        if "+parallel-netcdf" in self.spec:
+            base_cmake_args.append(self.define("ENABLE_PNETCDF", True))
+        return base_cmake_args
+
+
+class AutotoolsBuilder(AutotoolsBuilder, BackupStep):
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        self.setup_run_environment(env)
```
scheibelp: I don't think this will work because `setup_run_environment` belongs to the package rather than the builder. You could retrieve it with `self.pkg.setup_run_environment()`, but it might make sense for this to be applied for both CMake and Autotools builds.

var/spack/repos/builtin/packages/netcdf-c/netcdfc-hdf5-link-mpi.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/liblib/CMakeLists.txt b/liblib/CMakeLists.txt
+index aa3a842d..691902c2 100644
+--- a/liblib/CMakeLists.txt
++++ b/liblib/CMakeLists.txt
+@@ -50,6 +50,7 @@ ADD_LIBRARY(netcdf nc_initialize.c ${LARGS} )
+ 
+ IF(MPI_C_INCLUDE_PATH)
+     target_include_directories(netcdf PUBLIC ${MPI_C_INCLUDE_PATH})
++    target_link_libraries(netcdf MPI::MPI_C)
```
scheibelp: These patches appear to be fundamental - do you know if there are any PRs for `netcdf-c` adding this functionality?

scheibelp (changes requested):
I have one question and one request

var/spack/repos/builtin/packages/netcdf-c/package.py
johnwparent: Ok, I'll refactor that as suggested, thanks.

var/spack/repos/builtin/packages/netcdf-c/netcdfc-hdf5-link-mpi.patch
johnwparent: Sorry, I must have missed this comment last week. Yes, there are, I made a PR implementing the relevant changes in netcdf around when I made this PR. FWIW the include change was not included in my changes as it seems as if there is some debate as to that point between netcdfc and hdf5.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -62,6 +64,14 @@ class NetcdfC(AutotoolsPackage):
         when="@4.7.2",
     )
 
+    patch("4.8.1-win-hdf5-with-zlib.patch", when="@4.8.1: platform=windows")
+
+    patch("netcdfc-win-inc-mpi.patch", when="platform=windows")
```
scheibelp: I think this and the next two patches should be combined: are the upstream changes these patches cover being managed as one PR? If not, then combining them could potentially lead to conflicts; I assume they'd all make it into the same release though.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -142,12 +153,65 @@ class NetcdfC(AutotoolsPackage):
...
+            self.define("BUILD_UTILITIES", True),
+            self.define("ENABLE_NETCDF_4", True),
+            self.define_from_variant("ENABLE_DAP", "dap"),
+            self.define("CMAKE_INSTALL_PREFIX", self.prefix),
+            self.define_from_variant("ENABLE_HDF4", "hdf4"),
+            self.define("ENABLE_PARALLEL_TESTS", False),
+        ]
+        if "+parallel-netcdf" in self.spec:
+            base_cmake_args.append(self.define("ENABLE_PNETCDF", True))
+        return base_cmake_args
```
scheibelp, line 212: There appears to be at least one missing option: 

* `ENABLE_FSYNC`
* `ENABLE_V2_API`


a couple that don't appear to be available in CMake (but are in `configure`):

* `--enable-dynamic-loading`
* `--enable-largefile`

scheibelp (changes requested):
I have a couple requests

var/spack/repos/builtin/packages/netcdf-c/package.py
johnwparent: This patch in particular should not be folded into the other changes as it was not included in my upstream PR for reasons mentioned in the other comment (I've found several threads comprised of debate as to whether this is hdf5's issue or netcdfc's). That just leaves the two below, which I'll try to combine into one patch.

johnwparent: Oh sorry, this comment was pinned under the wrong line in the UI. Disregard, my comment, all three should be in one patch.

var/spack/repos/builtin/packages/netcdf-c/package.py
johnwparent: `ENABLE_V2_API` is on by default in the CMake build so we don't need to touch it to mirror the autotools behavior. 

I'll amend FSYNC. I also found the other two in CMake (at least the version I checked out) so I'll amend those as appropriate (per version presumably).

johnwparent: Actually I'll just enable other other two for all versions of netcdfc and cmake will ignore those it does not support.

var/spack/repos/builtin/packages/netcdf-c/package.py
johnwparent: All patches have been condensed.

scheibelp (dismissed):

var/spack/repos/builtin/packages/netcdf-cxx/package.py
```diff
@@ -29,8 +29,8 @@ def libs(self):
         shared = True
         return find_libraries("libnetcdf_c++", root=self.prefix, shared=shared, recursive=True)
 
-    def configure_args(self):
-        args = []
+    def cmake_args(self):
+        args = [self.define_from_variant("USE_NETCDF4", "netcdf4")]
```
scheibelp: The error at: https://gitlab.spack.io/spack/spack/-/jobs/5522408 looks legitimate.

```
>> 9    CMake Error: The source directory "/tmp/root/spack-stage/spack-stage-netcdf-cxx-4.2-yax57ewy54t2cr2bvkkfgltizxtkcz5h/spack-build-yax57ew/CPPFLAGS=-DUSE_NETCDF4" does not exist.
```

It's not clear to me why this change is causing the error but when I examine the surrounding logic I can say at least that we should no longer have to set `CPPFLAGS` (in fact all of the below search logic may be unnecessary).

johnwparent: I took a closer look at the CMake system for NetCDF-CXX and you're absolutely right, we don't need any of the find logic. I've taken it all out and pushed. Hopefully Gitlab CI will work now.

scheibelp:
Interestingly, the CI appears to fail legitimately here too, but for `paraview` (after building `netcdf-c` and `netcdf-cxx`):

```
__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path/linux-amzn2-x86_64_v3/gcc-7.3.1/libxml2-2.10.3-g6w7zcj4juhpgptfzeoauubewsgcsvmn/lib/libxml2.so  -lpthread && :
/usr/bin/ld: cannot find -lhdf5_hl-shared
/usr/bin/ld: cannot find -lhdf5-shared
collect2: error: ld returned 1 exit status
```

Paraview depends on `netcdf-c` (but not `netcdf-cxx`) so this would have something to do with locating `hdf5` as part of the `netcdf-c` build: it's possible that the `netcdf-c` build needs to be updated to locate `hdf5`, and this in turn helps the `paraview` build. Alternatively, it may be that `paraview` needs to be updated to locate `hdf5` itself, and whatever prior logic existed in `netcdf-c` was simply obscuring a mistake in the `paraview` build.

var/spack/repos/builtin/packages/netcdf-cxx/package.py
```diff
@@ -29,17 +29,6 @@ def libs(self):
...
-        args = []
-        if "+netcdf4" in self.spec:
-            # There is no clear way to set this via configure, so set the flag
-            # explicitly
-            args.append("CPPFLAGS=-DUSE_NETCDF4")
-        # Add these to LDFLAGS explicitly, so the linker doesn't accidentally
-        # use system versions
-        ldflags = [
-            self.spec["netcdf-c"].libs.search_flags,
-            self.spec["hdf5"].libs.search_flags,
```
johnwparent: @scheibelp is this maybe the reason paraview is now failing to find hdf5?

scheibelp: I thought paraview uses `netcdf-c` (not `netcdf-cxx`) so I don't think changes in `netcdf-cxx` would affect it

johnwparent: Ah good point, I think I'm mixing up VTK and Paraview dependencies in my head. VTK is the one that uses netcdf-cxx.

johnwparent:
@scheibelp I have determined the origin of the error (Netcdfc's cmake is the culprit) and am in the process of creating yet another patch and upstream PR to netcdfc to fix this.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -89,11 +95,12 @@ class NetcdfC(AutotoolsPackage):
...
 
     # The patch for 4.7.0 touches configure.ac. See force_autoreconf below.
-    depends_on("autoconf", type="build", when="@4.7.0,main")
-    depends_on("automake", type="build", when="@4.7.0,main")
-    depends_on("libtool", type="build", when="@4.7.0,main")
+    with when("build_system=autotools"):
+        depends_on("autoconf", type="build", when="@4.7.0,main")
+        depends_on("automake", type="build", when="@4.7.0,main")
+        depends_on("libtool", type="build", when="@4.7.0,main")
+        depends_on("m4", type="build")
```
skosukhin: Are you sure, we don't need `m4` when building with `cmake` (e.g. [here](https://github.com/Unidata/netcdf-c/blob/43abd699e19db24e27e1e800086ff8142f3b07ad/CMakeLists.txt#L2016))?

johnwparent: netcdf-c master (and any subsequent releases post `4.9.0`) will likely need this enabled yes, however for the time being netcdf-c builds perfectly fine without it as the files m4 is looking to generate are checked into source. I'll add the dep back for linux platforms, but as Spack does not currently support installing m4 for Windows, we won't add it as a dep for Windows.

scheibelp (approved):

var/spack/repos/builtin/packages/netcdf-c/4.8.1-win-hdf5-with-zlib.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index ba66a6d4..217aa712 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -856,7 +856,7 @@ IF(USE_HDF5)
+   # This needs to be near the beginning since we
+   # need to know whether to add "-lz" to the symbol
+   # tests below.
+-  CHECK_C_SOURCE_COMPILES("#include <H5public.h>
++  CHECK_C_SOURCE_COMPILES("#include <H5pubconf.h>
```
skosukhin, line 10: @johnwparent This patch reverts https://github.com/Unidata/netcdf-c/pull/1731 and `H5public.h` does include `H5pubconf.h`. So, why is it needed on Windows?

I am asking because I am refactoring the `netcdf-c` package now and figuring out whether we need all the patches we apply.

var/spack/repos/builtin/packages/netcdf-c/package.py
```diff
@@ -145,13 +153,72 @@ class NetcdfC(AutotoolsPackage):
...
+    def _nc_config_backup_dir(self):
+        return join_path(self.pkg.metadata_dir, "spack-nc-config")
+
+    @run_after("install")
+    def backup_nc_config(self):
+        # We expect this to be run before filter_compiler_wrappers:
+        nc_config_file = self.prefix.bin.join("nc-config")
+        if os.path.exists(nc_config_file):
+            mkdirp(self._nc_config_backup_dir)
+            install(nc_config_file, self._nc_config_backup_dir)
```
alalazo, line 192: If we want to decorate with `run_after` a method from a class that is not a builder, we need to set explicitly the correct metaclass:
```python
class BackupStep(metaclass=spack.builder.PhaseCallbacksMeta):
    ...
```

johnwparent: Is this or any of the other several multi build gotchas documented anywhere that I seem to be missing?
Edit: This is clearly a python language thing rather than a spack thing, but could still use a more obvious callout.

PR #34936, opened 2023-01-13
More Windows Paraview support ports

### LINKED ISSUE #34938: Paraview Support on Windows

johnwparent:
More ports supporting Paraview on Windows
Ports include packages:
- xz
- libtheora
- libogg

Changes to msvc compiler class supporting these ports.

### DIFF
--- a/lib/spack/spack/compilers/msvc.py
+++ b/lib/spack/spack/compilers/msvc.py
@@ -112,7 +112,8 @@ class Msvc(Compiler):
     def platform_toolset_ver(self):
         """
         This is the platform toolset version of current MSVC compiler
-        of form v<toolset-ver>, i.e. v142. This is different from the VC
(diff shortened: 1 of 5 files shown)


bernhardkaindl (changes requested):
Like Adam pointed out in #34930, this too does not fulfill the "one package per PR" policy:
> These changes are all independent and belong in separate PRs for each package. Spack has a strict "one package per PR" policy unless the changes to one package depend on the changes to another package.

bernhardkaindl:
@johnwparent Even though these packages are somewhat smaller, one could still be liable to violating the rule when merging this, so better also split this into 3 separate PRs. Also see https://github.com/spack/spack/pull/34932 which I closed as well (there is no fixing these PRs - open new, separate PRs for each package)

PR #34944, opened 2023-01-14
rclone: Add v1.61.1 and deprecate old versions due to CVE-2020-28924

alecbcs:
Add Rclone v1.61.1 which includes multiple bug fixes.

**Changelog**
- Show only significant parts of version number in version introduced label.
- Fix unescaped HTML.
- lib/http: Shutdown all servers on exit to remove unix socket.
- Don't serve via http if serving via --stdio.
- Fix running duplicate Serve call.

Full changelog can be found [here](https://rclone.org/changelog/#v1-61-1-2022-12-23).

### DIFF
--- a/var/spack/repos/builtin/packages/rclone/package.py
+++ b/var/spack/repos/builtin/packages/rclone/package.py
@@ -15,6 +15,7 @@ class Rclone(Package):
 
     maintainers = ["alecbcs"]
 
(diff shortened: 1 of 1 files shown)


alecbcs:
@spackbot run pipeline

alalazo (approved):

PR #34354, opened 2022-12-06
New package: py-fastfold

meyersbs:
This was not fun to package: [GitHub](https://github.com/hpcaitech/FastFold/tree/0.2.0). It doesn't exist in PyPI and the preferred installation method is via conda environment. As a result, the `setup.py` is incomplete.

Notes:
- I manually tested imports for dependencies listed in the conda `environment.yml` until I got everything working.
- The conda environment lists `PyYAML`, `tqdm`, `typing-extensions`, and `pyarrow`, but those aren't used anywhere in the source code, so I left them out of the spack package.
- I also had to pull a few dependencies from the README, and add some dependencies that aren't even listed in the conda environment because of import errors.

Import test:
```
Python 3.8.7 (default, Dec 2 2022, 14:27:46) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import fastfold; from fastfold.config import * ; from fastfold.workflow import * ; from fastfold.utils import * ; from fastfold.relax import * ; from fastfold.model import * ; from fastfold.distributed import * ; from fastfold.data import * ; from fastfold.common import * ; from fastfold.data.tools import * ; from fastfold.model.fastnn import * ; from fastfold.model.hub import * ; from fastfold.model.nn import * ; from fastfold.model.nn.embedders import * ; from fastfold.model.nn.msa import * ; from fastfold.model.nn.dropout import * ; from fastfold.model.hub.alphafold import * ; from fastfold.model.fastnn.embedders import * ; from fastfold.model.fastnn.evoformer import * ; from fastfold.utils.geometry import * ; from fastfold.utils.geometry.rigid_matrix_vector import * ; from fastfold.workflow.factory import * ; from fastfold.workflow.template import *; from fastfold.workflow.factory.hhsearch import *; from fastfold.data.data_modules import *; from fastfold.relax.amber_minimize import *
[12/02/22 15:42:52] WARNING  colossalai - root - WARNING: Triton is not available, fallback to old kernel.                                                                                                                                                    
                    WARNING  colossalai - root - WARNING: Triton is not available, fallback to old kernel.                                                                                                                                                    
>>> 
```

Issue #18276, opened 2020-08-25
Installation issue: npm


RikkiButler20:
[2020-08-25T17_15_36_969Z-debug.log](https://github.com/spack/spack/files/5126581/2020-08-25T17_15_36_969Z-debug.log)
Unable to install npm on its own or as a dependency.

### Steps to reproduce the issue

```console
$ spack install npm
...
==> Error: ProcessError: Command exited with status 2:
    'make' '-j12'

1 error found in build log:
     188    npm ERR!
     189    npm ERR! Failed at the npm-cli-docs@0.1.0 build:static script.
     190    npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
     191    
     192    npm ERR! A complete log of this run can be found in:
     193    npm ERR!     /Users/butler59/.npm/_logs/2020-08-25T21_08_57_981Z-debug.log
  >> 194    make: *** [htmldocs] Error 1
```

### Information on your system

* **Spack:** 0.15.4-698-8b5fcbfaf
* **Python:** 3.7.8
* **Platform:** darwin-catalina-skylake

### Additional information

The log files show

```console
Error: self signed certificate in certificate chain
```

adamjstewart:
That's odd. I'm on an almost identical system and I was able to install `npm`:

* **Spack:** 0.15.4-680-afa907f549
* **Python:** 3.7.8
* **Platform:** darwin-catalina-ivybridge

There's no maintainer for the package, so your best bet is to contact the `npm` developers.

becker33:
@adamjstewart then maybe this is an LLNL firewall problem, I had the same issue installing npm.

I'll see what happens if I drop off vpn.

becker33:
@RikkiButler20 try building without logging into your LLNL vpn. It seems that the vpn is the issue.

RikkiButler20:
It worked when I wasn't logged into my VPN.

adamjstewart:
It sounds like the problem here is that our `npm` package requires internet access at build-time and downloads additional dependencies. We should figure out how to disable this or add resources for the expected dependencies. I imagine this problem also affects users on air-gapped networks.

becker33:
I haven't been able to determine whether it's trying to download something or whether it's trying to get up-to-date data for something.

PR #34947, opened 2023-01-14
npm: Add latest version, update build

tgamblin:
The `npm` package had gotten a bit long in the tooth and only suported the last version for which running `configure` / `make` / `make install` actually worked.

- [x] Update the package to support npm@9, in which `npm install .` works properly and installation is easier.

- [x] Update the package so that `npm@6:8` also install successfullly. The incantation that is *supposed* to work on these versions is `node bin/npm-cli.js install $(node bin/npm-cli.js pack . | tail -1)`, but depending on the version one of `npm install` or `npm pack` will fail when run straight from the install directory. So now we just manually copies things over.

This seems to make the `npm` install much more reliable for all of `npm@6:9` (at least for me).

udpates the `npm` build to support versions 6-9 and fixes the install for all of them on macos.

### DIFF
--- a/var/spack/repos/builtin/packages/npm/package.py
+++ b/var/spack/repos/builtin/packages/npm/package.py
@@ -16,18 +16,43 @@ class Npm(Package):
     # base https://www.npmjs.com/
 
     git = "https://github.com/npm/cli.git"
-    url = "https://registry.npmjs.org/npm/-/npm-9.30.0.tgz"
+    url = "https://registry.npmjs.org/npm/-/npm-9.3.1.tgz"
 
-    version("9.30.0", commit="4d275927e069606c8047b678ad09aafbe18b1673")
-    version("8.19.3", commit="ef4299658330470b56966b48394824895b2d6d04")
-    version("7.24.2", commit="04eb43f2b2a387987b61a7318908cf18f03d97e0")
+    version("9.3.1", sha256="41caa26a340b0562bc5429d28792049c980fe3e872b42b82cad94e8f70e37f40")
+    version("8.19.3", sha256="634bf4e0dc87be771ebf48a058629960e979a209c20a51ebdbc4897ca6a25260")
+    version("7.24.2", sha256="5b9eeea011f8bc3b76e55cc33339e87213800677f37e0756ad13ef0e9eaccd64")
+    version("6.14.18", sha256="c9b15f277e2a0b1b57e05bad04504296a27024555d56c2aa967f862e957ad2ed")
 
-    version("6.14.9", sha256="1e0e880ce0d5adf0120fb3f92fc8e5ea5bac73681d37282615d074ff670f7703")
-    version("6.14.8", sha256="fe8e873cb606c06f67f666b4725eb9122c8927f677c8c0baf1477f0ff81f5a2c")
-    version("6.13.7", sha256="6adf71c198d61a5790cf0e057f4ab72c6ef6c345d72bed8bb7212cb9db969494")
-    version("6.13.4", sha256="a063290bd5fa06a8753de14169b7b243750432f42d01213fbd699e6b85916de7")
-    version("3.10.9", sha256="fb0871b1aebf4b74717a72289fade356aedca83ee54e7386e38cb51874501dd6")
-    version("3.10.5", sha256="ff019769e186152098841c1fa6325e5a79f7903a45f13bd0046a4dc8e63f845f")
+    version(
+        "6.14.9",
+        sha256="1e0e880ce0d5adf0120fb3f92fc8e5ea5bac73681d37282615d074ff670f7703",
+        deprecated=True,
+    )
+    version(
+        "6.14.8",
+        sha256="fe8e873cb606c06f67f666b4725eb9122c8927f677c8c0baf1477f0ff81f5a2c",
+        deprecated=True,
+    )
+    version(
+        "6.13.7",
+        sha256="6adf71c198d61a5790cf0e057f4ab72c6ef6c345d72bed8bb7212cb9db969494",
+        deprecated=True,
+    )
+    version(
+        "6.13.4",
+        sha256="a063290bd5fa06a8753de14169b7b243750432f42d01213fbd699e6b85916de7",
+        deprecated=True,
+    )
+    version(
+        "3.10.9",
+        sha256="fb0871b1aebf4b74717a72289fade356aedca83ee54e7386e38cb51874501dd6",
+        deprecated=True,
+    )
+    version(
+        "3.10.5",
+        sha256="ff019769e186152098841c1fa6325e5a79f7903a45f13bd0046a4dc8e63f845f",
+        deprecated=True,
+    )
 
     depends_on("node-js", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/npm/package.py
```diff
@@ -9,13 +9,18 @@
...
-# NOTE: not actually an Autotools package
 class Npm(Package):
     """npm: A package manager for javascript."""
 
     homepage = "https://github.com/npm/cli"
     # base https://www.npmjs.com/
-    url = "https://registry.npmjs.org/npm/-/npm-6.13.4.tgz"
+
+    git = "https://github.com/npm/cli.git"
+    url = "https://registry.npmjs.org/npm/-/npm-9.30.0.tgz"
```
adamjstewart: This URL doesn't exist, which I guess is the reason you're using git instead. I would just leave it with the old version number in the string.

var/spack/repos/builtin/packages/npm/package.py
```diff
@@ -9,13 +9,18 @@
...
+
+    version("9.30.0", commit="4d275927e069606c8047b678ad09aafbe18b1673")
+    version("8.19.3", commit="ef4299658330470b56966b48394824895b2d6d04")
+    version("7.24.2", commit="04eb43f2b2a387987b61a7318908cf18f03d97e0")
 
     version("6.14.9", sha256="1e0e880ce0d5adf0120fb3f92fc8e5ea5bac73681d37282615d074ff670f7703")
```
adamjstewart: Thoughts on deprecating and eventually removing older versions? There are currently only 2 packages in Spack that depend on npm, suggesting it isn't very popular in the HPC community. These older versions significantly increase the complexity of the build recipe. I'm not sure if anyone actually needs npm 3 or 6 anymore, or even 7 and 8...

var/spack/repos/builtin/packages/npm/package.py
```diff
@@ -35,30 +40,60 @@ class Npm(Package):
...
         install_tree("node-gyp/package", "node_modules/node-gyp")
         filter_file(r'"node-gyp": "\^5\..*"', '"node-gyp": "^6.0.1"', "package.json")
         install_tree("env-paths/package", "node_modules/env-paths")
 
-    def configure(self, spec, prefix):
-        configure("--prefix={0}".format(prefix))
+    @when("@:8")
+    def install(self, spec, prefix):
+        # `npm install .` doesn't work properly out of the box on npm up to 8, so we do
+        # what it would do manually. The only thing we seem to miss is docs) In
```
adamjstewart:
```suggestion
        # what it would do manually. The only thing we seem to miss is docs. In
```

adamjstewart (commented):
Is there any documentation that lists which versions of npm are compatible with which versions of node-js?

@meyersbs can you test out this PR and see if this solves the npm build issues you reported in #34354?

@RikkiButler20 same question for #18276

@cosmicexplorer you're listed as the maintainer for node-js, would you like to help maintain npm too? Or @tgamblin now that he's an npm installation expert 😛

var/spack/repos/builtin/packages/npm/package.py
meyersbs: Maybe throw a warning when people try to use "deprecated" versions, but keep them in the recipe so reproducibility isn't impacted?

meyersbs:
> @meyersbs can you test out this PR and see if this solves the npm build issues you reported in #34354?

@adamjstewart Yes, this PR seems to have fixed my build issue with npm:

```
$ spack install py-fastfold ^python@3.10:
...
==> Installing npm-9.30.0-r6ddw2jkmfjceovbancmbadkndjj7ohx
==> npm exists in binary cache but with different hash
==> No binary for npm-9.30.0-r6ddw2jkmfjceovbancmbadkndjj7ohx found: installing from source
==> No patches needed for npm
==> npm: Executing phase: 'install'
==> npm: Successfully installed npm-9.30.0-r6ddw2jkmfjceovbancmbadkndjj7ohx
  Stage: 16.19s.  Install: 3.94s.  Total: 20.47s
[+] /dev/shm/bsmits_spack/opt/spack/linux-rhel7-skylake_avx512/gcc-9.3.0/npm-9.30.0-r6ddw2jkmfjceovbancmbadkndjj7ohx
...
==> Installing py-fastfold-0.2.0-sivc25ixrafpwj33awttdzfakzvw2ql6
...
[+] /dev/shm/bsmits_spack/opt/spack/linux-rhel7-skylake_avx512/gcc-9.3.0/py-fastfold-0.2.0-sivc25ixrafpwj33awttdzfakzvw2ql6
```

However, the concretizer picked py-numpy@1.24.1, which is throwing an import error:

```
$ python
>>> from fastfold.data.data_modules import *
...
AttributeError: module 'numpy' has no attribute 'int'. Did you mean: 'inf'?
```

When I was building py-fastfold, it grabbed py-numpy@1.22.3, which works -- I'll investigate and submit a new PR updating py-fastfold.

var/spack/repos/builtin/packages/npm/package.py
adamjstewart: Marking them as deprecated will throw a warning automatically. But the only point of deprecating them is to someday remove them.

meyersbs:
> new PR updating py-fastfold.

https://github.com/spack/spack/pull/34962

var/spack/repos/builtin/packages/npm/package.py
tgamblin: I just had the version number wrong (it's actually 9.3.0) and 9.3.1 came out since I wrote this.  I updated that and switched everything to sha256, and kept the git url.

var/spack/repos/builtin/packages/npm/package.py
tgamblin: fixed

tgamblin:
Ok this is ready for another review.

meyersbs:
LGTM.

```
==> Installing py-fastfold-0.2.0-ri4wz2nhxtcd4pe74bnm34lgnonyey57
==> py-fastfold exists in binary cache but with different hash
==> No binary for py-fastfold-0.2.0-ri4wz2nhxtcd4pe74bnm34lgnonyey57 found: installing from source
==> Using cached archive: /dev/shm/bsmits_spack/var/spack/cache/_source-cache/archive/67/6760dbae9809b8b26219c9477489d34325807be504098901d0375fbdc3103f88.tar.gz
==> No patches needed for py-fastfold
==> py-fastfold: Executing phase: 'install'
==> py-fastfold: Successfully installed py-fastfold-0.2.0-ri4wz2nhxtcd4pe74bnm34lgnonyey57
  Stage: 0.06s.  Install: 2m 41.88s.  Total: 2m 42.73s
[+] /dev/shm/bsmits_spack/opt/spack/linux-rhel7-skylake_avx512/gcc-9.3.0/py-fastfold-0.2.0-ri4wz2nhxtcd4pe74bnm34lgnonyey57
```

adamjstewart:
node-js build took two hours, but this finally built on my system, thanks!

adamjstewart (approved):

PR #34951, opened 2023-01-15
python: 3.11.1, use -flto=thin when clang

haampie:
--with-lto is quite slow, --with-lto=thin is pretty fast.

### DIFF
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -478,7 +478,7 @@ class Python(Package):
         if "+optimizations" in spec:
             config_args.append("--enable-optimizations")
             # Prefer thin LTO for faster compilation times.
-            if "@3.11.0: %clang@3.9:" in spec or "@3.11.0: %apple-clang@8:":
+            if "@3.11.0: %clang@3.9:" in spec or "@3.11.0: %apple-clang@8:" in spec:
                 config_args.append("--with-lto=thin")
             else:
                 config_args.append("--with-lto")
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -477,7 +477,8 @@ class Python(Package):
 
         if "+optimizations" in spec:
             config_args.append("--enable-optimizations")
-            if "@3.11.0: %clang" in spec:
+            # Prefer thin LTO for faster compilation times.
+            if "@3.11.0: %clang@3.9:" in spec or "@3.11.0: %apple-clang@8:":
                 config_args.append("--with-lto=thin")
             else:
                 config_args.append("--with-lto")
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -44,6 +44,7 @@ class Python(Package):
     install_targets = ["install"]
     build_targets: List[str] = []
 
+    version("3.11.1", sha256="baed518e26b337d4d8105679caf68c5c32630d702614fc174e98cb95c46bdfa4")
     version("3.11.0", sha256="64424e96e2457abbac899b90f9530985b51eef2905951febd935f0e73414caeb")
     version(
         "3.10.8",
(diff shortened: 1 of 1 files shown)


adamjstewart:
I barely know what these options do and don't know if I can review this properly. Is there some documentation to explain why we should do this only for Python 3.11+ and only for Clang?

haampie:
It's just lto but with more possibilities for parallelism during linking. It's the preferred way of using lto with clang.

adamjstewart:
> It's the preferred way of using lto with clang.

Link to docs or discussion or something? Not that I don't trust you, I just want to give a useful review.

haampie:
See https://github.com/python/cpython/issues/88506#issuecomment-1093918111 / https://wiki.gentoo.org/wiki/Clang#Link-time_optimizations_with_Clang / https://clang.llvm.org/docs/ThinLTO.html. CMake defaults to -flto=thin for clang since it introduced lto support in v3.9. It's the best default cause it makes LTO builds run in comparable time as without. And --with-optmizations benefits from faster builds since it has to build Python twice (so, it's at least twice as slow, and with full LTO it's more in the direction of 4x as slow...)

I can make it conditional on clang 3.9 and xcode 8 if you want, since that's where llvm introduced it.

haampie:
Semi-related, it'd be nice to make `+optimizations` the default, since it improves Python significantly, and otherwise Spack's Python is typically much slower than system Python

adamjstewart:
I'd be fine with that, just have to make sure it builds with most compilers.

haampie:
Regarding `+optimizations`, it's just that GCC builds are terribly slow because (a) their LTO is single-threaded and (b) their instrumented binaries may have more overhead.

So, GCC 12, total build time for me: `9m 56.96s` of which `1 min 50 sec` is running unit tests for to collect profiling data

Compare to clang with thin LTO: `2m 56.51s` of which `41.3 sec` is running unit tests. (however, the logs are saying that 8 of 44 tests failed due to "env changed")

10 minute builds are maybe a bit too much for a default...

bernhardkaindl:
> Regarding `+optimizations`, it's just that GCC builds are terribly slow because (a) their LTO is single-threaded and (b) their instrumented binaries may have more overhead.
> 
> So, GCC 12, total build time for me: `9m 56.96s` of which `1 min 50 sec` is running unit tests for to collect profiling data

@haampie: AFAIK, It's only `--with-lto` which adds this so much higher CPU usage to the build:
```py 
        if "+optimizations" in spec:
            config_args.append("--enable-optimizations")
            config_args.append("--with-lto")
            config_args.append("--with-computed-gotos")
```            
Since the mission of spack is to provide performance for HCP, building python without `--enable-optimizations` by default should be fixed. IMHO, we should change `"--enable-optimizations"` and `"--with-computed-gotos"` to be always enabled without any variant. `+optimizations"` also looks like a misnomer to me, it should be "+lto" because LTO is a special kind of optimization, and the "lto" variant is already implemented by `gdb`, `binutils`, and `timemory`.

bernhardkaindl (commented):
Looks good to me, but want to add at the same time:

> Regarding `+optimizations`, it's just that GCC builds are terribly slow because (a) their LTO is single-threaded and (b) their instrumented binaries may have more overhead.
> 
> So, GCC 12, total build time for me: `9m 56.96s` of which `1 min 50 sec` is running unit tests for to collect profiling data

@haampie: AFAIK, It's only `--with-lto` which adds this so much higher CPU usage to the build:
```py 
        if "+optimizations" in spec:
            config_args.append("--enable-optimizations")
            config_args.append("--with-lto")
            config_args.append("--with-computed-gotos")
```            

Since the mission of spack is to provide performance for HPC, building python without `--enable-optimizations` by default should be fixed. IMHO, we should change `"--enable-optimizations"` and `"--with-computed-gotos"` to be always enabled without any variant. `+optimizations"` also looks like a misnomer to me, it should be "+lto" because LTO is a special kind of optimization, and the "lto" variant is already implemented by `gdb`, `binutils`, and `timemory`.

alalazo:
> 10 minute builds are maybe a bit too much for a default...

That was the original reason for having `False` as a default. On the other hand I think we should maybe discuss (Wed. meetings?) if our defaults should privilege:
1. Performance
2. Speed of building from sources
3. A mix of the two 

and be mostly consistent among packages.

adamjstewart:
If we had public binaries this would be a much easier decision.

haampie:
Lets' leave it for another PR and just get the latest patch release + thin lto for clang in

alalazo (approved):

adamjstewart (approved):

PR #34952, opened 2023-01-15
pfunit: Fix max_array_rank variant

mathomp4:
As I learn more about spack, I found that the pfunit package had a bug:
```python
     variant(
     |   "max_array_rank",
     |   values=int,
     |   default=5,
     |   description="Max number of Fortran dimensions of array asserts",
     )
```
That `values=int` is doing wonders to `spack info pfunit`.

The solution is remove that line. Then `spack info pfunit` is happy.


var/spack/repos/builtin/packages/pfunit/package.py
```diff
@@ -55,7 +55,6 @@ class Pfunit(CMakePackage):
 
     variant(
         "max_array_rank",
-        values=int,
```
alalazo, line 58: Without adding a set of allowed values in place of this, any string is a valid value for the variant. So I think this PR introduces a new bug. The long string in `spack info` comes from the fact that, if a function is used as a validator, the description of it is its docstring. 

That said, if you can turn this into a variant with a finite set of allowed values, it would be great.

alalazo (changes requested):

var/spack/repos/builtin/packages/pfunit/package.py
mathomp4: @alalazo While @tclune would know better than me, I think the maximum array rank in Fortran is 15, so I'll make a values tuple from that. Please see latest commit.

mathomp4: Note also, I have no idea how this scales. I've never done anything other than 5!

alalazo: > Note also, I have no idea how this scales. I've never done anything other than 5!

If it is fine for maintainers of the package to remove the variant completely and use only `5` in the recipe that would be fine with me too. Let me know which option you prefer. Like it is now it's also fine with me.

tclune (dismissed):

alalazo:
@spackbot run pipeline

var/spack/repos/builtin/packages/pfunit/package.py
mathomp4: @alalazo I think it's good. At least the option is there. Maybe someday I'll do a scaling study and find out 15 would take a year or something and pull it back. But I tried to put in a warning to users about it!

alalazo (dismissed):

alalazo:
@spackbot run pipeline

alalazo:
Pipelines are failing to build `pfunit`: https://gitlab.spack.io/spack/spack/-/jobs/5204801

tclune:
I have only taken a quick look at the error messages, but my guess is that there are explicit unit tests that require a minimum rank.   So the build with smaller `max_rank` make work, but not the tests.    The fix would be to propagate `max_rank` into the test layer and deactivate tests that have too high rank.   Or, with a bit more time, parameterize tests so that something is exercised at each rank.   (What, you don't trust the compiler?)

I don't really have time to work on that for the moment though.

If this failed with max_rank=5, then my guess above is not correct.

mathomp4:
> Pipelines are failing to build `pfunit`: https://gitlab.spack.io/spack/spack/-/jobs/5204801

@alalazo Does your CI build *every* version or just select ones? I ask because it is trying to build pFUnit 3.3.3:

==> Installing pfunit-3.3.3-b6b7v2qhg2lc5mfuuyt7x25m6xyte6oj

which is pretty old at this point. Or does it build the latest of each major? I'm just trying to figure out why it decided to do 3.3.3.

alalazo:
@mathomp4 As far as I can see the dependency gets in there because of `flecsi`, which has:
```python
depends_on("pfunit@3.0:3", when="@:1.9")
```

mathomp4:
> @mathomp4 As far as I can see the dependency gets in there because of `flecsi`, which has:
> 
> ```python
> depends_on("pfunit@3.0:3", when="@:1.9")
> ```

I think I might see the issue. I added a dependency to `fargparse` but that only came in at v4.0.0. But without a `when` 3.3.3 was grabbing it and I think then it was just all wrong.

Hopefully this fixes it...

alalazo (approved):

mathomp4:
Just for my own benefit in the future. If you compare the failed CI:

https://gitlab.spack.io/spack/spack/-/jobs/5204801

to the newly successful:

https://gitlab.spack.io/spack/spack/-/jobs/5207818

You'll see that not depending on `fargparse` seemed to do the trick. You don't see spack building the dependencies of `fargparse` which I think was the actual cause (probably `gftl` and `gftl-shared` where interfaces have changes since pFUnit 3.3.3 days).

PR #34580, opened 2022-12-16
Bypass compiler check against netcdf-c compiler in ncview, fix finding udunits on certain systems

climbfuji:
## Description

nciew (both versions that are configured in spack) has the annoying "feature" to check if the compiler used for ncview is the same as the one for netcdf-c. While this is in principle a good idea, we don't need it in spack (spack should take care of this), and it doesn't work with package relocation or on Cray systems that use the cray compiler wrappers.

This PR removes the consistency test from configure. Tested successfully on my Mac, on NOAA's RDHPC Gaea system (Cray), and on NASA's Discover system (required the `udunits` updates).

PR #34956, opened 2023-01-16
intel-oneapi-mpi: fix setup_dependent_build_environment

skosukhin:
The MPI compiler wrappers prefer `I_MPI_*` over `MPICH_*`, which means that if the former are set in the environment, the MPI compiler wrappers do not use the Spack compiler wrapper. This PR should fix that.

### DIFF
--- a/var/spack/repos/builtin/packages/intel-oneapi-mpi/package.py
+++ b/var/spack/repos/builtin/packages/intel-oneapi-mpi/package.py
@@ -114,11 +114,11 @@ class IntelOneapiMpi(IntelOneApiLibraryPackage):
             self.spec.mpifc = join_path(self.component_prefix.bin, "mpiifort")
 
     def setup_dependent_build_environment(self, env, dependent_spec):
-        env.set("MPICH_CC", spack_cc)
-        env.set("MPICH_CXX", spack_cxx)
-        env.set("MPICH_F77", spack_f77)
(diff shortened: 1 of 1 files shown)


skosukhin:
@climbfuji could you, please, check if this makes the workaround that you mentioned in https://github.com/spack/spack/pull/34580#issuecomment-1382326458 unnecessary? Thank you.

alalazo (approved):
Waiting for @rscohn2 and @climbfuji confirmation before merging

rscohn2 (approved):
Thanks

climbfuji:
@skosukhin Is it sufficient to only set `I_MPI_*` or should we also set `MPICH_*` just because some weird package may be looking for this explicitly?

skosukhin:
In my opinion, we set the variables for the MPI compiler wrappers of `intel-oneapi-mpi` to use the Spack compiler wrapper. No more, no less. Other packages are not supposed to look into these variables. And if they do, it's their problem, which they should work around.

climbfuji:
> In my opinion, we set the variables for the MPI compiler wrappers of `intel-oneapi-mpi` to use the Spack compiler wrapper. No more, no less. Other packages are not supposed to look into these variables. And if they do, it's their problem, which they should work around.

Agreed, make sense.

climbfuji (approved):

PR #34958, opened 2023-01-16
environments: don't replace relative view path with absolute path on concretize/install

haampie:
Currently if you have a `spack.yaml` that specifies a view by relative
path like this:

```yaml
spack:
  view: ./view
```

Spack expands it to an absolute path on `spack -e . install` and
persists that to disk like this:

```yaml
spack:
  view: /the/path/where/the/env/is/view
```

This is rather annoying when you share a `spack.yaml` file inside a git
repo, cause you want to use relative paths to make it relocatable, but
you constantly have to undo the changes made to `spack.yaml` by Spack
when committing something else.

So, as an alternative:

1. Treat `spack.yaml` view paths in read-only mode
2. Work with a path property that is canonicalized internally
3. Turn relative paths on the command line into absolute paths before
   storing to spack.yaml. This way you can do `spack env create --dir`
   `./env --with-view ./view` and both `./env` and `./view` are resolved
   to the current working dir, as expected (not `./env/view`). This
   corresponds to the old behavior of `spack env create`.

### DIFF
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -384,7 +384,7 @@ class ViewDescriptor(object):
         link_type="symlink",
     ):
         self.base = base_path
-        self._root = root
+        self.raw_root = root
         self.root = spack.util.path.canonicalize_path(root, default_wd=base_path)
         self.projections = projections
         self.select = select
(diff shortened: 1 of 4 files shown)


lib/spack/spack/environment/environment.py
```diff
@@ -384,7 +384,8 @@ def __init__(
         link_type="symlink",
     ):
         self.base = base_path
-        self.root = spack.util.path.canonicalize_path(root)
+        self._root = root
```
becker33: I would change this variable name to `self.raw_root` so that we don't have to have a reference from another class to an "private" (underscore) variable in this one.

becker33 (changes requested):
This looks good to me overall, tiny style request.

becker33 (approved):

PR #34964, opened 2023-01-16
gcc: add patch for noexcept declarations

glennpj:
There are some declarations that need to be declared noexcept for valarray. This affects gcc 9.5:11.2.


bernhardkaindl (approved):
LGTM

alalazo (approved):

PR #34965, opened 2023-01-16
sra-tools: add version 3.0.3 and fix build

glennpj:


bernhardkaindl (approved):
Yep, indeed. This fixes the build of sra-tools.

PR #34966, opened 2023-01-16
hisat2: add version 2.2.1 and sra variant

glennpj:
- add python dependency
- add perl dependency
- set python and perl interpreters in scripts


bernhardkaindl (approved):
LGTM

PR #34967, opened 2023-01-16
itk: add version 5.3.0

glennpj:


bernhardkaindl (changes requested):
Fails for me on Ubuntu 20.04 with the default gcc-8.4.0 with this error:
```py
hdf5-1.12.2-v533pkcfav45zvedpxl4cmdc6qn2amau/include/H5public.h:68:10: fatal error: mpi.h: No such file or directory
     1612     #include <mpi.h>
     1613              ^~~~~~~
```
I guess the recipe misses `depends_on("mpi")` (for `etc/spack/defaults/packages.yaml:      mpi: [openmpi, mpich])` or so.

bernhardkaindl (approved):
Builds with this change for me.

PR #34969, opened 2023-01-16
Add packages for scine modules

awvwgk:
- lbfgspp (dependency for scine-utilities)
- scine-core
- scine-utilities
- scine-readuct
- scine-sparrow (deprecating sparrow)
- scine-database
- scine-molassembler
- scine-xtb
- py-scine-chemoton
- py-scine-puffin

Mostly adapted from conda-forge feedstocks.


var/spack/repos/builtin/packages/scine-core/package.py
```diff
@@ -0,0 +1,42 @@
...
+    depends_on("googletest")
+
+    def patch(self):
+        os.rmdir("dev")
+        os.rename("_dev", "dev")
+
+    def cmake_args(self):
+        return [
+            self.define("SCINE_BUILD_TESTS", self.run_tests),
+            self.define("SCINE_MARCH", ""),
```
ltalirz, line 38: I'm not familiar with best practices for spack recipes but shouldn't this be set to the architecture that you are currently building on?

If the architecture is not forwarded here, I suspect it is possible that more "standard" ways of specifying the build architecture may be ignored by the build (?).

tldahlgren:
Added @adamjstewart since there are a couple of Python packages in this PR.

var/spack/repos/builtin/packages/scine-sparrow/package.py
```diff
@@ -8,7 +8,7 @@
 from spack.package import *
 
 
-class Sparrow(CMakePackage):
+class ScineSparrow(CMakePackage):
```
tldahlgren, line 11: @frobnitzem What are your thoughts on the name change?

tldahlgren (commented):
Confirmed the twelve version sha256 across the ten packages.

var/spack/repos/builtin/packages/scine-core/package.py
awvwgk: See https://spack.readthedocs.io/en/latest/basic_usage.html#support-for-specific-microarchitectures, spack has a way to inject the correct flags for the requested microarch. Having scine set them as well will interfere with this mechanism, therefore it is disabled here.

var/spack/repos/builtin/packages/py-scine-chemoton/package.py
```diff
@@ -0,0 +1,24 @@
...
+
+class PyScineChemoton(PythonPackage):
+    """Software driving the automated exploration of chemical reaction networks"""
+
+    homepage = "https://scine.ethz.ch/download/chemoton"
+    pypi = "scine_chemoton/scine_chemoton-2.2.0.tar.gz"
+
+    version("2.2.0", "cda4f9de1e1c00ecc5e0b2d9c17a5edb56b468d454022e3f4045ec116ba2ec45")
+
+    depends_on("python@3.6:", type=("build", "run"))
```
adamjstewart, line 19: Missing a dependency on py-networkx

var/spack/repos/builtin/packages/py-scine-chemoton/package.py
```diff
@@ -0,0 +1,24 @@
...
+    depends_on("py-numpy", type="run")
+    depends_on("py-scine-database", type="run")
+    depends_on("py-scine-molassembler", type="run")
+    depends_on("py-scine-utilities", type="run")
+    depends_on("py-scipy", type="run")
+    depends_on("py-setproctitle", type="run")
```
adamjstewart: These should all be `type=("build", "run")`. Some build tools check to make sure the dependencies exist in `PYTHONPATH` before building. Also, our import tests try to import all files during install and will require these to be available.

var/spack/repos/builtin/packages/py-scine-puffin/package.py
```diff
@@ -0,0 +1,29 @@
...
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-gitpython", type="run")
+    depends_on("py-numpy", type="run")
+    depends_on("py-psutil", type="run")
+    depends_on("py-python-daemon", type="run")
+    depends_on("py-pyyaml", type="run")
+    depends_on("py-scine-database", type="run")
+    depends_on("py-scine-molassembler", type="run")
+    depends_on("py-scine-readuct", type="run")
+    depends_on("py-scine-utilities", type="run")
```
adamjstewart: These deps aren't listed in requirements.txt

var/spack/repos/builtin/packages/py-scine-puffin/package.py
```diff
@@ -0,0 +1,29 @@
...
+    depends_on("py-scipy", type="run")
+    depends_on("py-setproctitle", type="run")
```
adamjstewart: Should all be type=build/run (except setuptools)

var/spack/repos/builtin/packages/scine-core/package.py
```diff
@@ -0,0 +1,42 @@
...
+            self.define("BOOST_ROOT", self.spec["boost"].prefix),
+            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].prefix.lib),
+            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].prefix.include),
```
adamjstewart:
```suggestion
            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].libs.directories[0]),
            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].headers.directories[0]),
```
On some systems, libs are in `lib64`, not `lib`. This will check both and return the correct location.

var/spack/repos/builtin/packages/scine-molassembler/package.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    variant("python", default=False, description="Build Python extension module")
+
+    depends_on("boost+system+filesystem+program_options cxxstd=17 @1.65.0:")
+    depends_on("eigen@3:")
+    depends_on("googletest")
+    # depends_on("nauty")
+    depends_on("nlohmann-json", type="build")
+    depends_on("python@3.7:", when="+python")
```
adamjstewart: What build type? build/run? build/link/run?

var/spack/repos/builtin/packages/scine-molassembler/package.py
```diff
@@ -0,0 +1,75 @@
...
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-pybind11", when="+python")
+    depends_on("py-scine-utilities", when="+python")
```
adamjstewart: Same here, build/run?

var/spack/repos/builtin/packages/scine-molassembler/package.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def cmake_args(self):
+        args = [
+            self.define("BUILD_SHARED_LIBS", True),
+            self.define("SCINE_BUILD_TESTS", self.run_tests),
+            self.define("SCINE_BUILD_PYTHON_BINDINGS", "+python" in self.spec),
+            self.define("SCINE_MARCH", ""),
+            self.define("BOOST_ROOT", self.spec["boost"].prefix),
+            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].prefix.lib),
+            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].prefix.include),
```
adamjstewart:
```suggestion
            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].libs.directories[0]),
            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].headers.directories[0]),
```

var/spack/repos/builtin/packages/scine-readuct/package.py
```diff
@@ -0,0 +1,69 @@
...
+        sha256="089ca500fc191e04a968ea166d2fe80178b227bc2e6d3926523fa2eee5f6492d",
+        placement="_dev",
+    )
+
+    variant("python", default=False, description="Build Python extension module")
+
+    depends_on("boost+system+filesystem+program_options cxxstd=17 @1.65.0:")
+    depends_on("eigen@3:")
+    depends_on("googletest")
+    depends_on("python@3.7:", when="+python")
```
adamjstewart: build/run?

var/spack/repos/builtin/packages/scine-readuct/package.py
```diff
@@ -0,0 +1,69 @@
...
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-pybind11", when="+python")
+    depends_on("py-scine-utilities", when="+python")
```
adamjstewart: build/run?

var/spack/repos/builtin/packages/scine-molassembler/package.py
```diff
@@ -0,0 +1,75 @@
...
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-pybind11", when="+python")
+    depends_on("py-scine-utilities", when="+python")
+    # depends_on("ringdecomposerlib")
+    depends_on("scine-core")
+    depends_on("scine-utilities")
+
+    extends("python", when="+python")
+
+    provides("py-scine-molassembler", when="+python")
```
adamjstewart: Provides is meant for virtual dependencies, I don't think it can have the same name as a real package. Is there any reason to have both `py-scine-molassembler` and `scine-molassember+python`?

var/spack/repos/builtin/packages/scine-readuct/package.py
```diff
@@ -0,0 +1,69 @@
...
+    depends_on("scine-core")
+    depends_on("scine-utilities")
+    depends_on("yaml-cpp")
+
+    extends("python", when="+python")
+
+    provides("py-scine-readuct", when="+python")
```
adamjstewart: Prob can't have this

var/spack/repos/builtin/packages/scine-database/package.py
```diff
@@ -0,0 +1,57 @@
...
+    depends_on("eigen@3:")
+    depends_on("googletest")
+    depends_on("mongo-cxx-driver@3.2.1:")
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-pybind11", when="+python")
+    depends_on("scine-utilities@5:")
+
+    extends("python", when="+python")
+
+    provides("py-scine-database", when="+python")
```
adamjstewart: Prob can't have this

var/spack/repos/builtin/packages/scine-sparrow/package.py
```diff
@@ -24,53 +24,68 @@ class Sparrow(CMakePackage):
...
+    variant("python", default=False, description="Build Python extension module")
+
     depends_on("boost+filesystem+program_options cxxstd=17 @1.65.0:")
-    depends_on("py-pybind11@2.6.2")
-    depends_on("py-pip", type="build")
-    depends_on("yaml-cpp")
     depends_on("cereal")
+    depends_on("eigen@3.3.2:")
     depends_on("googletest")
+    depends_on("python@3.7:", when="+python")
```
adamjstewart: build/run?

var/spack/repos/builtin/packages/scine-sparrow/package.py
```diff
@@ -24,53 +24,68 @@ class Sparrow(CMakePackage):
...
+    depends_on("py-pip", type="build", when="+python")
+    depends_on("py-scine-utilities", when="+python")
```
adamjstewart: build/run?

var/spack/repos/builtin/packages/scine-sparrow/package.py
```diff
@@ -24,53 +24,68 @@ class Sparrow(CMakePackage):
...
+    depends_on("scine-core")
+    depends_on("scine-utilities")
+    depends_on("yaml-cpp")
+
+    extends("python", when="+python")
+
+    provides("py-scine-sparrow", when="+python")
```
adamjstewart: Prob can't have this

var/spack/repos/builtin/packages/scine-sparrow/package.py
```diff
@@ -24,53 +24,68 @@ class Sparrow(CMakePackage):
...
 
     def cmake_args(self):
         args = [
-            self.define("SCINE_BUILD_PYTHON_BINDINGS", True),
             self.define("SCINE_BUILD_TESTS", self.run_tests),
+            self.define("SCINE_BUILD_PYTHON_BINDINGS", "+python" in self.spec),
+            self.define("SCINE_MARCH", ""),
+            self.define("BOOST_ROOT", self.spec["boost"].prefix),
+            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].prefix.lib),
+            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].prefix.include),
```
adamjstewart:
```suggestion
            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].libs.directories[0]),
            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].headers.directories[0]),
```

var/spack/repos/builtin/packages/scine-utilities/package.py
```diff
@@ -0,0 +1,79 @@
...
+        placement="_dev",
+    )
+
+    variant("python", default=False, description="Build Python extension module")
+
+    depends_on("boost+system+filesystem+program_options cxxstd=17 @1.65.0:")
+    depends_on("eigen@3:")
+    depends_on("googletest")
+    depends_on("lbfgspp", type="build")
+    depends_on("python@3.7:", when="+python")
```
adamjstewart: type?

var/spack/repos/builtin/packages/scine-utilities/package.py
```diff
@@ -0,0 +1,79 @@
...
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-pybind11", when="+python")
+    depends_on("scine-core")
+    depends_on("yaml-cpp")
+
+    extends("python", when="+python")
+
+    provides("py-scine-utilities", when="+python")
```
adamjstewart: Can't have

var/spack/repos/builtin/packages/scine-utilities/package.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+    def cmake_args(self):
+        args = [
+            self.define("SCINE_BUILD_TESTS", self.run_tests),
+            self.define("SCINE_BUILD_PYTHON_BINDINGS", "+python" in self.spec),
+            self.define("SCINE_MARCH", ""),
+            self.define("BOOST_ROOT", self.spec["boost"].prefix),
+            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].prefix.lib),
+            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].prefix.include),
```
adamjstewart: libs/headers

var/spack/repos/builtin/packages/scine-xtb/package.py
```diff
@@ -0,0 +1,59 @@
...
+        sha256="089ca500fc191e04a968ea166d2fe80178b227bc2e6d3926523fa2eee5f6492d",
+        placement="_dev",
+    )
+
+    variant("python", default=False, description="Build Python extension module")
+
+    depends_on("boost+system+filesystem+program_options cxxstd=17 @1.65.0:")
+    depends_on("eigen@3:")
+    depends_on("pkgconfig", type="build")
+    depends_on("python@3.7:", when="+python")
```
adamjstewart: type?

var/spack/repos/builtin/packages/scine-xtb/package.py
```diff
@@ -0,0 +1,59 @@
...
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-scine-utilities", when="+python")
```
adamjstewart: type?

var/spack/repos/builtin/packages/scine-xtb/package.py
```diff
@@ -0,0 +1,59 @@
...
+    depends_on("scine-core")
+    depends_on("scine-utilities")
+    depends_on("xtb")
+
+    extends("python", when="+python")
+
+    provides("py-scine-xtb", when="+python")
```
adamjstewart: Can't have

var/spack/repos/builtin/packages/scine-xtb/package.py
```diff
@@ -0,0 +1,59 @@
...
+        os.rename("_dev", "dev")
+
+    def cmake_args(self):
+        args = [
+            self.define("SCINE_BUILD_TESTS", self.run_tests),
+            self.define("SCINE_BUILD_PYTHON_BINDINGS", "+python" in self.spec),
+            self.define("SCINE_MARCH", ""),
+            self.define("BOOST_ROOT", self.spec["boost"].prefix),
+            self.define("BOOST_LIBRARY_DIR", self.spec["boost"].prefix.lib),
+            self.define("BOOST_INCLUDE_DIR", self.spec["boost"].prefix.include),
```
adamjstewart: libs/headers

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-scine-puffin/package.py
awvwgk: These dependencies are optional in puffin because it can download the source and build them if not available. It seems preferable to just promote them from optional to explicit to have spack rather than puffin manage their installation.

var/spack/repos/builtin/packages/scine-molassembler/package.py
awvwgk: I think those packages can use ``scine-module+python`` as well. I thought it could be more convenient for users to install ``py-scine-sparrow`` and ``py-scine-chemoton`` rather than having to remember that Sparrow is a C++ project and needs to be installed as ``scine-sparrow+python`` to get the Python bindings.

var/spack/repos/builtin/packages/scine-core/package.py
ltalirz: Ok!

After looking into the cmake files, it appears indeed that this option is mostly used to prevent mixing of scine packages compiled for different architectures (which should not really be an issue with spack).
In some cases, the option is used to set compile flags explicitly, which indeed is better handled by spack.

In one case it is used to forward this information to a dependency it is building (in the serenity wrapper https://github.com/qcscine/serenity_wrapper/blob/a2214474601ab4e75a9da35a1deb28ff6efe1874/CMakeLists.txt#L27, which is indeed not included here).
Also there, the correct solution is to have spack build the dependency rather than cmake.

All good!

Edit: for some reason I am not able to resolve this conversation

var/spack/repos/builtin/packages/scine-sparrow/package.py
frobnitzem: This makes good sense, since it reflects the group of packages sparrow is a part of.   @awvwgk has done the hard work of putting all the dependencies into spack recipes, but it will take some time to review them.

adamjstewart: We may want to keep the old package and deprecate all versions, in addition to the new package. This will prevent surprises for users expecting the old name. See https://spack.readthedocs.io/en/latest/packaging_guide.html#deprecating-old-versions

var/spack/repos/builtin/packages/scine-database/package.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    variant("python", default=False, description="Build Python extension module")
+
+    depends_on("eigen@3:")
+    depends_on("googletest")
+    depends_on("mongo-cxx-driver@3.2.1:")
+    depends_on("python@3.6:", when="+python", type=("build", "run"))
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-pybind11", when="+python", type=("build", "run"))
```
adamjstewart: Usually pybind11 is build or build/link only, it isn't needed at run-time.

var/spack/repos/builtin/packages/scine-database/package.py
```diff
@@ -0,0 +1,56 @@
...
+    depends_on("scine-utilities@5:")
```
adamjstewart, line 36: +python when +python

var/spack/repos/builtin/packages/scine-utilities/package.py
```diff
@@ -0,0 +1,77 @@
...
+
+    variant("python", default=False, description="Build Python extension module")
+
+    depends_on("boost+system+filesystem+program_options cxxstd=17 @1.65.0:")
+    depends_on("eigen@3:")
+    depends_on("googletest")
+    depends_on("lbfgspp", type="build")
+    depends_on("python@3.7:", when="+python", type=("build", "run"))
+    depends_on("py-pip", when="+python", type="build")
+    depends_on("py-pybind11@2.6.2:", when="+python", type=("build", "run"))
```
adamjstewart: Missing deps on numpy/scipy

var/spack/repos/builtin/packages/py-scine-chemoton/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class PyScineChemoton(PythonPackage):
+    """Software driving the automated exploration of chemical reaction networks"""
+
+    homepage = "https://scine.ethz.ch/download/chemoton"
+    pypi = "scine_chemoton/scine_chemoton-2.2.0.tar.gz"
+    git = "https://github.com/qcscine/puffin.git"
+
+    version("develop", branch="master")
```
adamjstewart: I would just call the version master to match the github

awvwgk: It is not very consistent in spack, I found several packages which call the latest version "develop" regardless of branch name, but also a lot of packages use "main" or "master" depending on the actual branch name. I'm fine with either way, but found the former strategy slightly better because the git version doesn't change the name if upstream renames from "master" to "main".

adamjstewart: That's for historical reasons. Once upon a time (several years ago) develop was the only version that meant "newer than all stable releases". Other versions like main or master were considered to be older than all stable releases, and patches wouldn't apply correctly. This was fixed a while back but there's still a lot of packages that haven't been updated. See https://github.com/spack/spack/blob/v0.19.0/lib/spack/spack/version.py#L64 for the full list of versions that are accepted.

var/spack/repos/builtin/packages/scine-sparrow/package.py
awvwgk: Done, there is now a (deprecated) `sparrow@3.0.0` package which depends on `scine-sparrow+python@3.0.0`.

var/spack/repos/builtin/packages/py-scine-chemoton/package.py
awvwgk: Thanks for the clarification, changed accordingly.

var/spack/repos/builtin/packages/sparrow/package.py
```diff
@@ -3,83 +3,23 @@
...
-    """Sparrow: fast semiempirical quantum chemical calculations.
-
-    When publishing results obtained with Sparrow, please cite
-    the corresponding release as archived on Zenodo
-    (DOI 10.5281/zenodo.3244105; please use the DOI of the respective
-    release).
-
-    In addition, we kindly request you to cite the following article
-    when using Sparrow:
+class Sparrow(Package):
```
adamjstewart: The old package should be identical but with `deprecated=True` added to the version. That way users have at least 1 release cycle to switch to the new name without anything breaking.

awvwgk:
Thanks a lot for the reviews. Still learning how to package for spack, so this was very helpful for me.

var/spack/repos/builtin/packages/py-scine-chemoton/package.py
adamjstewart: It still says develop

adamjstewart (approved):
Looks good from my end

awvwgk:
Is this ready to be merged?

PR #34970, opened 2023-01-16
py-macs2: add version 2.2.7.1 and support python@3.10:

glennpj:
The tarball from PyPi includes the Cythonized C files. The tarball from github does not. Remove the Cythonized C files from the source so that they are rebuilt with the Spack Python/Cython combination. This is necessary for python-3.10 but make sense for other combinations as well.

### DIFF
--- a/var/spack/repos/builtin/packages/py-macs2/package.py
+++ b/var/spack/repos/builtin/packages/py-macs2/package.py
@@ -17,34 +17,25 @@ class PyMacs2(PythonPackage):
 
     version("2.2.7.1", sha256="ad2ca69bdd02a8942a68aae23133289b5c16ba382bcbe20c39fabf3948929de5")
     version("2.2.4", sha256="b131aadc8f5fd94bec35308b821e1f7585def788d2e7c756fc8cac402ffee25b")
-    version("2.1.4", sha256="e4966d001914320829ab859c7bc8e92c6410aa7bdbddfd00b7625e9a0fb15c97")
-    version("2.1.3.3", sha256="00959e523f45ed92b8429f55944eca6984623ac008d7cdb488c3ffe59c21984a")
-    version(
-        "2.1.1.20160309", sha256="2008ba838f83f34f8e0fddefe2a3a0159f4a740707c68058f815b31ddad53d26"
-    )
 
     # patch to correctly identify python-3.10 as greater than required version
     patch(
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-macs2/package.py
```diff
@@ -13,13 +15,38 @@ class PyMacs2(PythonPackage):
...
+    version("2.1.3.3", sha256="00959e523f45ed92b8429f55944eca6984623ac008d7cdb488c3ffe59c21984a")
+    version(
+        "2.1.1.20160309", sha256="2008ba838f83f34f8e0fddefe2a3a0159f4a740707c68058f815b31ddad53d26"
+    )
+
+    # patch to correctly identify python-3.10 as greater than required version
+    patch(
+        "https://github.com/macs3-project/MACS/pull/497.patch?full_index=1",
+        sha256="eaff891b9b3c6a910bd5d454dcc6e21288c8d1ad4d6d6f77e370bc8f90921cbd",
+        when="^python@3.10:",
```
adamjstewart: Does this patch work for all versions? I would at least add `@:2.2.7.1` since it won't be needed in the future (if there ever is a new release someday).

var/spack/repos/builtin/packages/py-macs2/package.py
```diff
@@ -13,13 +15,38 @@ class PyMacs2(PythonPackage):
...
+    )
+
+    depends_on("python@3.6:", when="@2.2.7.1:", type=("build", "run"))
+    depends_on("python@3.5:", when="@2.2:", type=("build", "run"))
+    depends_on("python@2.7:2.8", when="@:2.1", type=("build", "run"))
```
adamjstewart: Spack no longer supports Python 2, you can remove all 2.1 versions

var/spack/repos/builtin/packages/py-macs2/package.py
glennpj: The 2.1 versions would not need it because they are python-2 only. I missed version 2.2.4 in there. Technically, it would need the patch, but I see that it does not apply cleanly to that version.

glennpj: Ah, version 2.2.4 does not build with python-3.10 anyway so a version constraint is in order.

adamjstewart (approved):

PR #34971, opened 2023-01-16
py-xxhash: fix xxhash dependency

glennpj:
The dependency should be xxhash@0.8.0:.


adamjstewart:
Is this documented anywhere?

glennpj:
> Is this documented anywhere?

I verified that this works with xxhash-0.8.1 but I could only find, "needs xxhash" in `setup.py`. The bundled version is 0.8.0, from 2.5 years ago, which would have been the latest version at the time. That gets bumped to 0.8.1 for py-xxhash-3 and I am not sure why I did not pull in the update as well. I should add that. I suppose the Spack version constraint could also be `xxhash@0.8`. My guess is the constraint was set based on the bundled version at the time, but that seems limiting, although perhaps less so if the version 3 was added.

adamjstewart (approved):

PR #34973, opened 2023-01-16
stringtie: add version 2.2.1 and fix build

glennpj:
- stringtie uses internal samtools so remove dependency on samtools
- add dependency for zlib
- use make release target to get correct flags


bernhardkaindl (approved):
Builds on Linux/x86 and command shows help text. LGTM

PR #34978, opened 2023-01-17
[armpl-gcc] Make pkg-config files available

stephenmsachs:
ARMpl pkgconfig files are located in a non-default location and do not have the .pc extension. Changing those both helps pkgconfig pick them up correctly, e.g. in the meson build of `py-scipy`.

### DIFF
--- a/var/spack/repos/builtin/packages/armpl-gcc/package.py
+++ b/var/spack/repos/builtin/packages/armpl-gcc/package.py
@@ -264,10 +264,10 @@ class ArmplGcc(Package):
 
     @run_after("install")
     def make_pkgconfig_files(self):
-        # ARMpl pkcfonfig files do not have .pc extension and are thus not found by pkg-config
+        # ArmPL pkgconfig files do not have .pc extension and are thus not found by pkg-config
         armpl_dir = get_armpl_prefix(self.spec)
         for f in find(join_path(armpl_dir, "pkgconfig"), "*"):
-            copy(f, f + ".pc")
+            symlink(f, f + ".pc")
 
     def setup_dependent_build_environment(self, env, dependent_spec):
         # pkgconfig directory is not in standard ("lib", "lib64", "share") location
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/armpl-gcc/package.py
```diff
@@ -262,3 +261,15 @@ def check_install(self):
         make("-C", armpl_example_dir, "ARMPL_DIR=" + armpl_dir)
         # clean up
         make("-C", armpl_example_dir, "ARMPL_DIR=" + armpl_dir, "clean")
+
+    @run_after("install")
+    def make_pkgconfig_files(self):
+        # ARMpl pkcfonfig files do not have .pc extension and are thus not found by pkg-config
```
annop-w: nit: # ArmPL pkgconfig

var/spack/repos/builtin/packages/armpl-gcc/package.py
```diff
@@ -262,3 +261,15 @@ def check_install(self):
...
+        armpl_dir = get_armpl_prefix(self.spec)
+        for f in find(join_path(armpl_dir, "pkgconfig"), "*"):
+            copy(f, f + ".pc")
```
annop-w: Is it better to create symlink instead ?

annop-w (changes requested):

var/spack/repos/builtin/packages/armpl-gcc/package.py
stephenmsachs: thanks.

var/spack/repos/builtin/packages/armpl-gcc/package.py
stephenmsachs: symlinks are probably clearer.

annop-w (approved):

annop-w (approved):

tldahlgren (approved):
LGTM

PR #34984, opened 2023-01-17
py-h5py %oneapi@2023.0.0: let warnings be warnings

eugeneswalker:
For `py-h5py %oneapi@2023.0.0` set the following:
* `-Wno-error=incompatible-function-pointer-types`
* `-Wno-error=incompatible-pointer-types-discards-qualifiers`

@bryanherman @takluyver @wspear


wspear (approved):

tgamblin:
@eugeneswalker sorry to resurrect an old PR but I'm curious if #30882 handles this or not.  Are the flags you're setting here erroring:
1. because the build uses `-Werror` or more specific stuff like `-Werror=incompatible-function-pointer-types`, or
2. because oneapi 2023 and above are more strict by default than prior versions and just make these things errors where they weren't errors before?

I'm guessing (2), but curious because I wonder if there is more that we need to do in #30882.  It seems like if these are warnings by default they shouldn't fail the build.

@trws FYI

PR #34986, opened 2023-01-17
environments: `develop` paths were not getting expanded

albestro:
This PR aims at fixing wrong expansion of develop paths in environment configuration.

I opted for `substitute_path_variables` instead of `canonicalize_path`, because I just want to expand variables/users without making any decision in case the given path is relative, leaving up to the environment logic what has to be done in that case.

### The problem

Given an environment

```
# /path-to-original-env/spack.yaml
spack:
  specs: [dla-future]
  develop:
    dla-future:
      spec: dla-future@develop
      path: ~/path-to-my-pkg
```

when cloning it

```bash
$ spack env create new_env original_env/spack.yaml
```

the `new_env` develop path resulted int

```
spack:
  specs: [dla-future]
  develop:
    dla-future:
      spec: dla-future@develop
      path: /path-to-original-env/~/workspace/repos/dla-future.master
```

### DIFF
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -379,8 +379,8 @@ def _rewrite_relative_dev_paths_on_relocation(env, init_file_dir):
         if not dev_specs:
             return
         for name, entry in dev_specs.items():
-            dev_path = spack.util.path.canonicalize_path(entry["path"])
-            expanded_path = os.path.normpath(os.path.join(init_file_dir, dev_path))
+            dev_path = substitute_path_variables(entry["path"])
+            expanded_path = spack.util.path.canonicalize_path(dev_path, default_wd=init_file_dir)
 
             # Skip if the expanded path is the same (e.g. when absolute)
             if dev_path == expanded_path:
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -379,8 +379,8 @@ def _rewrite_relative_dev_paths_on_relocation(env, init_file_dir):
         if not dev_specs:
             return
         for name, entry in dev_specs.items():
-            dev_path = entry["path"]
-            expanded_path = os.path.normpath(os.path.join(init_file_dir, entry["path"]))
+            dev_path = spack.util.path.canonicalize_path(entry["path"])
+            expanded_path = os.path.normpath(os.path.join(init_file_dir, dev_path))
 
             # Skip if the expanded path is the same (e.g. when absolute)
             if dev_path == expanded_path:
--- a/lib/spack/spack/test/cmd/dev_build.py
+++ b/lib/spack/spack/test/cmd/dev_build.py
@@ -234,7 +234,7 @@ def test_dev_build_env_with_vars(tmpdir, install_mockery, mutable_mock_env_path,
     with envdir.as_cwd():
         with open("spack.yaml", "w") as f:
             f.write(
-                f"""\
+                """\
 spack:
   specs:
   - dev-build-test-install@0.0.0
--- a/lib/spack/spack/test/cmd/dev_build.py
+++ b/lib/spack/spack/test/cmd/dev_build.py
@@ -225,9 +225,8 @@ def test_dev_build_env_with_vars(tmpdir, install_mockery, mutable_mock_env_path,
     # store the build path in an environment variable that will be used in the environment
     monkeypatch.setenv("CUSTOM_BUILD_PATH", build_dir)
 
-    with build_dir.as_cwd():
-        with open(spec.package.filename, "w") as f:
-            f.write(spec.package.original_string)
+    with build_dir.as_cwd(), open(spec.package.filename, "w") as f:
+        f.write(spec.package.original_string)
 
     # setup environment
     envdir = tmpdir.mkdir("env")
--- a/lib/spack/spack/test/cmd/dev_build.py
+++ b/lib/spack/spack/test/cmd/dev_build.py
@@ -215,6 +215,45 @@ spack:
         assert f.read() == spec.package.replacement_string
 
 
+def test_dev_build_env_with_vars(tmpdir, install_mockery, mutable_mock_env_path, monkeypatch):
+    """Test Spack does dev builds for packages in develop section of env (path with variables)."""
+    # setup dev-build-test-install package for dev build
+    build_dir = tmpdir.mkdir("build")
+    spec = spack.spec.Spec("dev-build-test-install@0.0.0 dev_path=%s" % build_dir)
+    spec.concretize()
+
+    # store the build path in an environment variable that will be used in the environment
+    monkeypatch.setenv("CUSTOM_BUILD_PATH", build_dir)
+
+    with build_dir.as_cwd():
+        with open(spec.package.filename, "w") as f:
+            f.write(spec.package.original_string)
+
+    # setup environment
+    envdir = tmpdir.mkdir("env")
+    with envdir.as_cwd():
+        with open("spack.yaml", "w") as f:
+            f.write(
+                f"""\
+spack:
+  specs:
+  - dev-build-test-install@0.0.0
+
+  develop:
+    dev-build-test-install:
+      spec: dev-build-test-install@0.0.0
+      path: $CUSTOM_BUILD_PATH
+"""
+            )
+        env("create", "test", "./spack.yaml")
+        with ev.read("test"):
+            install()
+
+    assert spec.package.filename in os.listdir(spec.prefix)
+    with open(os.path.join(spec.prefix, spec.package.filename), "r") as f:
+        assert f.read() == spec.package.replacement_string
+
+
 def test_dev_build_env_version_mismatch(tmpdir, install_mockery, mutable_mock_env_path):
     """Test Spack constraints concretization by develop specs."""
     # setup dev-build-test-install package for dev build
--- a/lib/spack/spack/test/cmd/dev_build.py
+++ b/lib/spack/spack/test/cmd/dev_build.py
@@ -219,7 +219,7 @@ def test_dev_build_env_with_vars(tmpdir, install_mockery, mutable_mock_env_path,
     """Test Spack does dev builds for packages in develop section of env (path with variables)."""
     # setup dev-build-test-install package for dev build
     build_dir = tmpdir.mkdir("build")
-    spec = spack.spec.Spec("dev-build-test-install@0.0.0 dev_path=%s" % build_dir)
+    spec = spack.spec.Spec(f"dev-build-test-install@0.0.0 dev_path={build_dir}")
     spec.concretize()
 
     # store the build path in an environment variable that will be used in the environment


lib/spack/spack/environment/environment.py
```diff
@@ -696,8 +696,8 @@ def _rewrite_relative_paths_on_relocation(self, init_file_dir):
             return
 
         for name, entry in self.dev_specs.items():
-            dev_path = entry["path"]
-            expanded_path = os.path.normpath(os.path.join(init_file_dir, entry["path"]))
+            dev_path = substitute_path_variables(entry["path"])
```
haampie: I think it's an overall improvement, although not sure how to deal with `$env`. Most sensible is when it refers to `original_env/`. (Or ... not?)

haampie: Ok, it's fine. `$env` always refers to the currently active env, otherwise `./x` can be used.

albestro: That's a good point. 

Note that `$env` is expanded just when there is an active environment, so it is not enough cloning from its `spack.yaml`.

As you are implicitly saying, among possible expandable variables, `$env` is the only one that still have a "relative" semantic.

Documentation says
> $env: name of the currently active environment

So I can have a `another_env` activated, then I can `spack env create new_one original_env/spack.yaml`, and the `$env` variable in `original_env/spack.yaml` will be expanded to `another_env` and not `original_env`.

Probably using `$env` is the easiest way to loose control of spack environments 😄 but it might be me missing some important usage of it.

albestro: I didn't realise you commented again while I was writing my comment 😅

haampie: Actually what should be used is `canonicalize_path` which makes the result absolute. But I just checked and `canonicalize_path` has a bug if the yaml file was opened by relative path itself... then the output is relative. Will look tomorrow.

albestro: As I said above, I don’t use `canonicalize_path` because it always make paths absolute, and at that point the management of relative paths when copying environments does not have any reason to exist anymore.

Probably you have a broader view since you are working on this relative vs absolute paths in #34958.

Look forward to your comments tomorrow 😉

haampie: If you read the docstring of `canonicalize_path` that is what we *should* use, but somehow it doesn't live up to its promise in returning an absolute path if you do `spack env create --dir ./new ./old/spack.yaml`... it returns `./old/<dev path>`. That is yet another bug :tada:

haampie:
Not sure if this is the *best* fix, but at least it uses `canonicalize_path`:

```diff
diff --git a/lib/spack/spack/cmd/env.py b/lib/spack/spack/cmd/env.py
index e4c62758c4..1c1603ed26 100644
--- a/lib/spack/spack/cmd/env.py
+++ b/lib/spack/spack/cmd/env.py
@@ -300,7 +300,7 @@ def env_create(args):
         # the environment should not include a view.
         with_view = None
     if args.envfile:
-        with open(args.envfile) as f:
+        with open(os.path.abspath(args.envfile)) as f:
             _env_create(
                 args.create_env, f, args.dir, with_view=with_view, keep_relative=args.keep_relative
             )
diff --git a/lib/spack/spack/environment/environment.py b/lib/spack/spack/environment/environment.py
index 4ef9e13dfa..8159d1f2fd 100644
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -644,7 +644,7 @@ def __init__(self, path, init_file=None, with_view=None, keep_relative=False):
                 directory.
         """
         self.path = os.path.abspath(path)
-        self.init_file = init_file
+        self.init_file = os.path.abspath(init_file) if isinstance(init_file, str) else init_file
         self.with_view = with_view
         self.keep_relative = keep_relative
 
@@ -655,11 +655,11 @@ def __init__(self, path, init_file=None, with_view=None, keep_relative=False):
         self.unify = None
         self.clear()
 
-        if init_file:
+        if self.init_file:
             # If we are creating the environment from an init file, we don't
             # need to lock, because there are no Spack operations that alter
             # the init file.
-            with fs.open_if_filename(init_file) as f:
+            with fs.open_if_filename(self.init_file) as f:
                 if hasattr(f, "name") and f.name.endswith(".lock"):
                     self._read_manifest(default_manifest_yaml())
                     self._read_lockfile(f)
@@ -670,8 +670,7 @@ def __init__(self, path, init_file=None, with_view=None, keep_relative=False):
                 # Rewrite relative develop paths when initializing a new
                 # environment in a different location from the spack.yaml file.
                 if not keep_relative and hasattr(f, "name") and f.name.endswith(".yaml"):
-                    init_file_dir = os.path.abspath(os.path.dirname(f.name))
-                    self._rewrite_relative_paths_on_relocation(init_file_dir)
+                    self._rewrite_relative_paths_on_relocation()
         else:
             with lk.ReadTransaction(self.txlock):
                 self._read()
@@ -688,24 +687,12 @@ def __init__(self, path, init_file=None, with_view=None, keep_relative=False):
     def __reduce__(self):
         return _create_environment, (self.path, self.init_file, self.with_view, self.keep_relative)
 
-    def _rewrite_relative_paths_on_relocation(self, init_file_dir):
+    def _rewrite_relative_paths_on_relocation(self):
         """When initializing the environment from a manifest file and we plan
         to store the environment in a different directory, we have to rewrite
         relative paths to absolute ones."""
-        if init_file_dir == self.path:
-            return
-
-        for name, entry in self.dev_specs.items():
-            dev_path = entry["path"]
-            expanded_path = os.path.normpath(os.path.join(init_file_dir, entry["path"]))
-
-            # Skip if the expanded path is the same (e.g. when absolute)
-            if dev_path == expanded_path:
-                continue
-
-            tty.debug("Expanding develop path for {0} to {1}".format(name, expanded_path))
-
-            self.dev_specs[name]["path"] = expanded_path
+        for entry in self.dev_specs.values():
+            entry["path"] = spack.util.path.canonicalize_path(entry["path"])
 
     def _re_read(self):
         """Reinitialize the environment object if it has been written (this

```

albestro:
I rebased this PR and adapted as suggested by @haampie using `canonicalize_path`.

Let me report results of my quick checks. The `canonicalize_path` @haampie was talking about seems fixed. Indeed, this is how previously was working (look at first line `dev_path`)

```console
➜  tmp spack env create -d ./b ./a/spack.yaml
entry['path']='./tmp-zlib' dev_path='a/tmp-zlib' expanded_path='/Users/ialberto/tmp/a/a/tmp-zlib'
entry['path']='~/tmp-tar' dev_path='/Users/ialberto/tmp-tar' expanded_path='/Users/ialberto/tmp-tar'
==> Created environment in /Users/ialberto/tmp/b
==> You can activate this environment with:
==>   spack env activate /Users/ialberto/tmp/b
```

and now it works correctly by actually returning an absolute path.

```console
➜  tmp spack env create -d ./b ./a/spack.yaml
entry['path']='./tmp-zlib' dev_path='/Users/ialberto/tmp/b/tmp-zlib' expanded_path='/Users/ialberto/tmp/a/tmp-zlib'
entry['path']='~/tmp-tar' dev_path='/Users/ialberto/tmp-tar' expanded_path='/Users/ialberto/tmp/a/~/tmp-tar'
==> Created environment in /Users/ialberto/tmp/b
==> You can activate this environment with:
==>   spack env activate /Users/ialberto/tmp/b
```

Moreover, it can be seen that also the `~` symbol is correctly expanded.

Just for completeness: given this environment

```yaml
# old/spack.yaml
spack:
  specs: []
  view: true
  concretizer:
    unify: true
  develop:
    zlib:
      spec: zlib@1
      path: ./tmp-zlib
    tar:
      spec: tar@1
      path: ~/tmp-tar
```

if we would perform `spack env create -d new old/spack.yaml` without this PR we would get

```yaml
# new/spack.yaml (without this PR)
spack:
  specs: []
  view: true
  concretizer:
    unify: true
  develop:
    zlib:
      spec: zlib@1
      path: /Users/ialberto/tmp/a/tmp-zlib
    tar:
      spec: tar@1
      path: /Users/ialberto/tmp/a/~/tmp-tar
```

which is clearly wrong, since `~` does not get expanded resulting in an invalid path, while with this PR we get

```yaml
# new/spack.yaml (with this PR)
spack:
  specs: []
  view: true
  concretizer:
    unify: true
  develop:
    zlib:
      spec: zlib@1
      path: ./tmp-zlib
    tar:
      spec: tar@1
      path: ~/tmp-tar
```

Not sure it is the expected behavior, but at least is a step forward since at least it generates a valid path. Let me know if you think a different behavior would be expected (e.g. you might want to translate relative paths to absolute paths pointing to the old environment).

albestro:
Looking at `lib/spack/spack/test/cmd/dev_build.py::test_dev_build_env` output is seems that expected behavior was slightly different.

This PR now produces

```yaml
# spack env create -d ./b ./a/spack.yaml
spack:
  specs: []
  view: true
  concretizer:
    unify: true
  develop:
    zlib:
      spec: zlib@1
      path: /Users/ialberto/tmp/a/tmp-zlib
    tar:
      spec: tar@1
      path: ~/tmp-tar
```

note: I'm struggling a bit realising what are the relevant tests to check (I'm not getting all-greens on `develop`, but apparently it is a platform related problem, since I'm on `darwin-aarch64`) so I just pushed and let's see if the CI helps me (in the meanwhile I will try to setup a different machine for spack development).

lib/spack/spack/test/cmd/dev_build.py
```diff
@@ -215,6 +215,45 @@ def test_dev_build_env(tmpdir, install_mockery, mutable_mock_env_path):
         assert f.read() == spec.package.replacement_string
 
 
+def test_dev_build_env_with_vars(tmpdir, install_mockery, mutable_mock_env_path, monkeypatch):
+    """Test Spack does dev builds for packages in develop section of env (path with variables)."""
+    # setup dev-build-test-install package for dev build
+    build_dir = tmpdir.mkdir("build")
+    spec = spack.spec.Spec("dev-build-test-install@0.0.0 dev_path=%s" % build_dir)
```
haampie, line 222:
```suggestion
    spec = spack.spec.Spec(f"dev-build-test-install@0.0.0 dev_path={build_dir}")
```

haampie: Can you grab it from the concrete env instead of concretizing twice? (Although TIL you can set `dev_path` before concretizing)

Something like `env.matching_spec(name)` or `env.concrete_roots()[0]`.

lib/spack/spack/test/cmd/dev_build.py
```diff
@@ -215,6 +215,45 @@ def test_dev_build_env(tmpdir, install_mockery, mutable_mock_env_path):
...
+    # setup dev-build-test-install package for dev build
+    build_dir = tmpdir.mkdir("build")
+    spec = spack.spec.Spec(f"dev-build-test-install@0.0.0 dev_path={build_dir}")
+    spec.concretize()
+
+    # store the build path in an environment variable that will be used in the environment
+    monkeypatch.setenv("CUSTOM_BUILD_PATH", build_dir)
+
+    with build_dir.as_cwd():
+        with open(spec.package.filename, "w") as f:
```
haampie: `with A, B:` is a bit shorter

lib/spack/spack/test/cmd/dev_build.py
```diff
@@ -215,6 +215,45 @@ def test_dev_build_env(tmpdir, install_mockery, mutable_mock_env_path):
...
+            f.write(spec.package.original_string)
+
+    # setup environment
+    envdir = tmpdir.mkdir("env")
+    with envdir.as_cwd():
+        with open("spack.yaml", "w") as f:
```
haampie, line 234: same

albestro: These nested `with` statements have the following structure

```python
with env.as_cwd():
  with open() as f:
    f.write()
  env("create", ...) # from f
  with ev.read(...): # created by previous line
    install()
```

switching to

```python
with env, open() as f, ev.read():
  f.write()
  env()
  install()
```

is equivalent to

```python
with env:
  with open() as f:
    with ev.read(): # environment does not exist yet
      f.write()
      env("create") # is going to read the file `f` while it is still open for writing
      install()
```

lib/spack/spack/test/cmd/dev_build.py
albestro: Not sure I get what you mean, probably I'm missing something.

The environment at this point is yet to be created, and this is just a concretization of the package in the wild so that it can be later used for comparison.

I can try reversing the logic of the test, to see if we can create the environment first and then get the package information from within the environment. To me it looks like the test semantic changes a bit. Let me know what you think and in case I can have a look (FIY I just copy-pasted the test above this).

haampie (approved):

lib/spack/spack/test/cmd/dev_build.py
haampie: I guess it's fine like this, but in principle it could also just concretize the env and check the `dev_path` variant, and maybe stage dir.

PR #34987, opened 2023-01-17
hypre: add option for sycl variant

eugeneswalker:
Starting this to add a `hypre +sycl` variant.

Question for maintainers:
* What is minimum version of `hypre` that should accept `+sycl` ?
* Should we make a conflict or error of some sort if someone tries to, say, build `hypre +sycl %gcc` ?

What do you think?

@balay @osborn9 @ulrikeyang @wspear

### DIFF
--- a/var/spack/repos/builtin/packages/hypre/package.py
+++ b/var/spack/repos/builtin/packages/hypre/package.py
@@ -276,6 +276,12 @@ class Hypre(AutotoolsPackage, CudaPackage, ROCmPackage):
 
         if "+sycl" in spec:
             configure_args.append("--with-scyl")
+            sycl_compatible_compilers = ["dpcpp", "icpx"]
+            if not (os.path.basename(self.compiler.cxx) in sycl_compatible_compilers):
+                raise InstallError(
+                    "Hypre's SYCL GPU Backend requires DPC++ (dpcpp)"
+                    + " or the oneAPI CXX (icpx) compiler."
+                )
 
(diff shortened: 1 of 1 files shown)


osborn9:
> Question for maintainers:
> 
>     * What is minimum version of `hypre` that should accept `+sycl` ?

2.24.0 is when sycl support was introduced. 

> 
>     * Should we make a conflict or error of some sort if someone tries to, say, build `hypre +sycl %gcc` ?
> 
Yes. Maybe mimic AMReX: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/amrex/package.py#L269? 
An issue is with default sycl options in hypre, we're assuming the location of Intel oneMKL libs are in ${MKLROOT}/lib/intel64 so we'll have problems with using spack installed `intel-oneapi-mkl`. We had a similar issue with ROCm so could do something similar as a workaround.

eugeneswalker:
I just pushed some changes in response to your post @osborn9 . Does this OK for a first pass? And we can resolve the issue with other mkl like `intel-oneapi-mkl` as needed in a future PR?

osborn9:
> I just pushed some changes in response to your post @osborn9 . Does this OK for a first pass? And we can resolve the issue with other mkl like `intel-oneapi-mkl` as needed in a future PR?

Yes, this looks good. Thanks!

osborn9 (approved):

wspear (approved):

PR #34988, opened 2023-01-17
Add further pkgconfig patching for HDF5

AlexanderRichert-NOAA:
This PR adds further corrections to HDF5 pkgconfig files, originally developed while working on #34577.


haampie:
Can you check https://github.com/HDFGroup/hdf5/issues/2259 and the referenced commits there? If we can avoid spack-specific post-install patching that'd be great.

PR #34990, opened 2023-01-17
updated samtools to 1.16.1 and htslib to 1.16

downloadico:
Updated the versions of these two packages.
N.B. There is no version 1.16.1 of htslib (as of 2023-01-17).


glennpj:
Can you add bcftools to this set as well? It needs to be kept in sync with samtools and htslib.

downloadico:
will re-open

PR #34991, opened 2023-01-17
Implements `--best-arch` option for spack load

ljyanesm:
Implements an option to load a spack spec that best matches the current platform's architecture.

When a package has been installed for more than a single architecture, a simple `spack load ...` will fail.
```
$ spack load libelf
==> Error: libelf matches multiple packages.
  Matching packages:
    bcuy2dh libelf@0.8.13%gcc@12.2.0 arch=linux-amzn2-skylake_avx512
    lkukxhw libelf@0.8.13%gcc@12.2.0 arch=linux-amzn2-zen2
  Use a more specific spec (e.g., prepend '/' to the hash).
```

Using the `--best-arch` parameter will instead select the spec that best matches the platform's architecture, in a way that is also compatible in situations where the output of $(spack arch) is not available in the specs installed on the site. This may happen if Spack base compiler is not as recent as the architecture on the platform.

E.g:

```
$ spack load libelf
==> Error: libelf matches multiple packages.
  Matching packages:
    lea4cp6 libelf@0.8.13%gcc@7.3.1 arch=linux-amzn2-skylake_avx512
    v5covop libelf@0.8.13%gcc@7.3.1 arch=linux-amzn2-zen
  Use a more specific spec (e.g., prepend '/' to the hash).
$ spack load libelf arch=$(spack arch)
==> Error: Spec 'libelf arch=linux-amzn2-zen2' matches no installed packages.
```

When using `--best-arch` the `arch=linux-amzn2-zen` is correctly selected

```
$ spack load --best-arch libelf
```

### DIFF
--- a/lib/spack/spack/cmd/__init__.py
+++ b/lib/spack/spack/cmd/__init__.py
@@ -262,9 +262,7 @@ def disambiguate_spec(spec, env, local=False, installed=True, best=False):
     return disambiguate_spec_from_hashes(spec, hashes, local, installed, best)
 
 
-def disambiguate_spec_from_hashes(
-    spec, hashes, local=False, installed=True, best=False
-):
+def disambiguate_spec_from_hashes(spec, hashes, local=False, installed=True, best=False):
     """Given a spec and a list of hashes, get concrete spec the spec refers to.
 
     Arguments:
--- a/lib/spack/spack/cmd/__init__.py
+++ b/lib/spack/spack/cmd/__init__.py
@@ -262,7 +262,7 @@ def disambiguate_spec(spec, env, local=False, installed=True, first=False, best_
 
 
 def disambiguate_spec_from_hashes(
-        spec, hashes, local=False, installed=True, first=False, best_arch=False
+    spec, hashes, local=False, installed=True, first=False, best_arch=False
 ):
     """Given a spec and a list of hashes, get concrete spec the spec refers to.
 
--- a/lib/spack/spack/cmd/__init__.py
+++ b/lib/spack/spack/cmd/__init__.py
@@ -245,7 +245,7 @@ def matching_spec_from_env(spec):
         return spec.concretized()
 
 
-def disambiguate_spec(spec, env, local=False, installed=True, first=False):
+def disambiguate_spec(spec, env, local=False, installed=True, first=False, best_arch=False):
     """Given a spec, figure out which installed package it refers to.
 
     Arguments:
@@ -258,10 +258,10 @@ def disambiguate_spec(spec, env, local=False, installed=True, first=False):
             See ``spack.database.Database._query`` for details.
     """
     hashes = env.all_hashes() if env else None
-    return disambiguate_spec_from_hashes(spec, hashes, local, installed, first)
+    return disambiguate_spec_from_hashes(spec, hashes, local, installed, first, best_arch)
 
 
-def disambiguate_spec_from_hashes(spec, hashes, local=False, installed=True, first=False):
+def disambiguate_spec_from_hashes(spec, hashes, local=False, installed=True, first=False, best_arch=False):
     """Given a spec and a list of hashes, get concrete spec the spec refers to.
 
     Arguments:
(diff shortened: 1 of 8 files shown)


jack-morrison:
nit: Is "best" used elsewhere in Spack for this sort of operation? Is naming like `--current-arch` more accurate?

CuiJinku:
Bravo! 👍

The feature is a godsend for me. I have had a similar situation and plan to make a script to solve the problem.

I have applied the change in my local branch of `spack` before the merge completes.

ljyanesm:
> nit: Is "best" used elsewhere in Spack for this sort of operation? Is naming like `--current-arch` more accurate?

I wondered the same, but this PR will load a `spec` that is not necessarily exactly the `current` arch, but compatible. So, thought it would be misleading.

haampie:
There's also --first I think. So, maybe a combination of `--host --first` also sounds like a good contender.

ljyanesm:
I am happy to rename it as `--host` combined with `--first` and implement the cli changes if that may get it merged.

haampie:
Or `--native`, since `--host` sounds more like compile-time jargon

lib/spack/spack/test/cmd/load.py
```diff
@@ -125,6 +125,20 @@ def test_load_first(install_mockery, mock_fetch, mock_archive, mock_packages):
     load("--sh", "--first", "libelf")
 
 
+def test_load_best(install_mockery, mock_fetch, mock_archive, mock_packages):
+    """Test with and without the --best option"""
+    install("libelf@0.8.12 arch=x86_64")
+    install("libelf@0.8.12")
```
haampie, line 120: The test here concretizes to the host arch from a potentially different family, so I think it needs to mock `archspec.cpu.host`

lib/spack/spack/cmd/load.py
```diff
@@ -57,10 +57,18 @@ def setup_parser(subparser):
         "--first",
         action="store_true",
         default=False,
-        dest="load_first",
+        dest="load_best",
         help="load the first match if multiple packages match the spec",
     )
 
+    subparser.add_argument(
+        "--best",
```
haampie, line 66: I believe
```suggestion
        "--best", "--first"
```
is supported, then you don't have to repeat yourself

jack-morrison:
@spackbot fix style

ljyanesm:
@spackbot fix style

lib/spack/spack/test/cmd/load.py
ljyanesm: @haampie, could you point me to an example for how this is done? I've had some issues with trying to test this and would like to get the PR merged if possible

lib/spack/spack/cmd/__init__.py
```diff
@@ -279,7 +284,26 @@ def disambiguate_spec_from_hashes(spec, hashes, local=False, installed=True, fir
...
+
+        # these are harder but can be done w/package prefs:
+        # TODO: pick version with most preferred variants
+        # TODO: pick version with most preferred virtual provider (e.g. MPI)
+
+        # sort key that orders specs by prefs above
+        def sort_key(spec):
+            return (ver(spec.version), target(spec.target), compiler(spec.compiler), spec)
+
+        matching_specs.sort(key=sort_key)
```
tgamblin, line 306: @adamjstewart, @haampie: see here -- it's not particularly hard to order candidates by target or version -- we already do this in the concretizer (`PackagePrefs` is what it uses).  I think this is a good enough start to merge and make default -- thoughts?

adamjstewart:
Is there any time when we _wouldn't_ want to load a package for the native arch? Personally I would say no flag and just make this the default behavior to match `module load`.

tgamblin:
The code above will load your most preferred version/ compiler for the best arch you have (you may not have native eg if you installed from a portable buildcache).

ljyanesm:
bump

ljyanesm:
Hi @alalazo, @tgamblin 

I was wondering if there was anything I could do to help with PR make it over the line or if there are some fundamental issues with what's proposed, for example, if maybe spack is not really intended as the tool to use for loading packages and maybe I should be doing this with some alternative tool? I would understand if that was the case, otherwise, if there's something missing for this PR to get through, please do let me know and I'll make some time for it.

Many thanks for the review and changes already done to the PR!

robertu94:
I think think this has been long running feature request.  I know several people at Clemson who would use this if it was merged.  For them, they have a cluster with different generations of CPUs.  It would be helpful to say, load the version that is best for the current compute node.

ljyanesm:
@robertu94,

Yes, that was the original intention of the PR. I wonder if there are alternatives to this now in Spack or potentially other ways of using Spack which prevent the need for this PR.

PR #34992, opened 2023-01-17
spack mirror create takes local paths only

### LINKED ISSUE #34989: Spack mirror mishandles paths
### Steps to reproduce

```
spack mirror create -d /some/absolute/path --all
```
What's happening here is that the Spack mirror command hands the -d/--directory argument off to `spack.util.url.path_to_file_url`, which converts the path into a `file:///some/absolute/path` URI. Then that URI gets passed to `spack.mirror.cache_and_stats`, which uses `os` lib functions to check the existence of that directory and create it if necessary. 

The problem is, `os` filesystem operations don't understand URI's (it interprets it as the relative path `file:/some/absolute/path`). Spack happily creates this directory and puts the mirror there. 

### Error message

None

* **Spack:** 0.20.0.dev0
* **Python:** 3.6.8
* **Concretizer:** clingo

### General information

haampie:
Fixes #34989 

spack mirror create cannot work with urls, so it shouldn't promote local
paths to file urls.

There's already a test covering this `test_mirror_from_env`, not sure why it
did not fail before.

### DIFF
--- a/lib/spack/spack/cmd/mirror.py
+++ b/lib/spack/spack/cmd/mirror.py
@@ -9,6 +9,7 @@ import llnl.util.lang as lang
 import llnl.util.tty as tty
 import llnl.util.tty.colify as colify
 
+import spack.caches
 import spack.cmd
 import spack.cmd.common.arguments as arguments
 import spack.concretize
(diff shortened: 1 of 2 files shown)


haampie:
@spackbot run pipeline

alalazo (approved):

PR #34997, opened 2023-01-18
acfl: update hash for 22.1 on RHEL-7

annop-w:


var/spack/repos/builtin/packages/acfl/package.py
```diff
@@ -19,7 +19,7 @@
 _versions = {
     "22.1": {
         "RHEL-7": (
-            "189119a72b2cf2840dc85d4fab74435c018e145d03dd3098f23364bd235ffb7b",
+            "bfbfef9099bf0e90480d48b3a1a741d583fc939284f869958e9c09e177098c73",
```
alalazo, line 22: Do you know why the hash changed? Are there notes somewhere?

annop-w: I've just checked with our product manager. The hash didn't change and the updated one here is correct. It could be my mistake when auto-generating these hashes in the first place.

alalazo (approved):

PR #34999, opened 2023-01-18
py-flask-sqlalchemy: add v2.5.1

charmoniumQ:
`py-parsl+monitoring` depends on `py-sqlalchemy@1.3` and `py-flask-sqlalchemy`, which depends on `py-sqlalchemy@1.4.18:`, which is a contradiction.

Sqlalchemy 1.3 -> 1.4 is a difficult upgrade, because it is meant as a transitional version to 2.x, which deprecates a lot of functions. The tip of development of Parsl _is_ compatible, but it hasn't been released yet. The simplest solution is to add in the most recent version of `py-flask-sqlalchemy` that is compatible with `sqlalchemy@1.3`.


var/spack/repos/builtin/packages/py-parsl/package.py
```diff
@@ -38,7 +38,7 @@ class PyParsl(PythonPackage):
     depends_on("py-psutil@5.5.1:", type=("build", "run"))
 
     with when("+monitoring"):
-        depends_on("py-sqlalchemy@1.3", type=("build", "run"))
+        depends_on("py-sqlalchemy@1.3:", type=("build", "run"))
```
hategan: The parsl setup.py for 1.1.0 (https://github.com/Parsl/parsl/blob/1.1.0/setup.py) contains the following: `sqlalchemy>=1.3.0,!=1.3.4,<1.4`. Should the version constraint be, instead, "py-sqlalchemy@1.3.0:1.3" to restrict it to <1.4?

charmoniumQ: Whoops, I think `@1.3` will already do that. I was looking at master instead of the `1.1.0` tag of parsl.

hategan: I'm not entirely sure. I didn't know exactly how the spack version specification works, but according to https://spack.readthedocs.io/en/latest/packaging_guide.html#dependencies, 1.3 is not quite right (see what follows the sentence "What if a package can only be built with Python 2.7?")

charmoniumQ: It turns out that documentation is out of date. https://github.com/spack/spack/pull/35002

bernhardkaindl (approved):
Build results on Ubuntu 22.04.4 LTS:
```py
bin/spack find --variants py-flask-sqlalchemy@2.5.1 | sed 's/~[a-z]*//g;s/ [a-z0-9_]*=[a-zA-Z0-9]*//g'  # Short variants: Hide disabled and value variants
-- linux-ubuntu22.04-skylake / clang@14.0.0 ---------------------
py-flask-sqlalchemy@2.5.1
```

Generated by https://github.com/spack/build-quality-tools/blob/main/build_pr_changes.py
```py
~/build-quality-tools/build_pr_changes.py -q queue.txt -umay
```

var/spack/repos/builtin/packages/py-parsl/package.py
bernhardkaindl: @charmoniumQ! Thanks for the PRs and the updates of them! Agreed!

PR #35002, opened 2023-01-18
Fix documentation relating to partial version specification

charmoniumQ:
This fixes an out-of-date documentation page (according to @adamjstewart).

One can confirm that the previous documentation is inaccurate by writing:

```
# spack/var/spack/repos/builtin/packages/sams-test-package/package.py
from spack.package import *
class SamsTestPackage(Package):
    version("1")
    depends_on("zlib@1.2", type=("build"))
    conflict("^zlib@1.2.13")
```
and execute
```
spack env create test-partial-versions
spack env activate test-partial-versions
spack add sams-test-package
spack concretize
# Notice that we get zlib@1.2.11, meaning that the depends_on was really 1.2.x for any x.
```

Changing this to `zlib@:1.2` or `zlib@1.2:` gets `1.2.13`, meaning that the "x" is interpreted to make the range as loose as possible (smallest possible x if on the left of a colon, largest if on the right).


lib/spack/docs/packaging_guide.rst
```diff
@@ -2315,17 +2315,27 @@ relationships between packages.
...
-Although some packages require a specific version for their dependencies,
-most can be built with a range of versions. For example, if you are
-writing a package for a legacy Python module that only works with Python
-2.4 through 2.6, this would look like:
+Although some packages require a specific version for their dependencies, most
+can be built with a range of versions. The simplest way to construct a range is
+to use a partial version number. If all of the packages are versioned with three
+numbers, ``x.y.z``, then ``x.y`` can match any version beginning with ``x.y.``
+(the third version number is a free variable). This is different from previous
+versions of Spack where ``x.y`` would match only the latest ``x.y.z`` and none
```
adamjstewart: In previous versions of Spack, `x.y` meant `x.y`. `spack install python@3.9` would try to download https://www.python.org/ftp/python/3.9/Python-3.9.tgz which doesn't exist.

charmoniumQ: Do you mean this was the behavior before the old behavior, 3.9.x where x can only be the latest?

adamjstewart: > 3.9.x where x can only be the latest?

I have never heard of this behavior before.

lib/spack/docs/packaging_guide.rst
```diff
@@ -2315,17 +2315,27 @@ relationships between packages.
...
+example, if you are writing a package for a legacy Python module that only works
+with Python 2.4.x through 2.6.x, this would look like:
 
 .. code-block:: python
 
    depends_on('python@2.4:2.6')
 
 Version ranges in Spack are *inclusive*, so ``2.4:2.6`` means any version
-greater than or equal to ``2.4`` and up to and including any ``2.6.x``. If
+greater than or equal to ``2.4.0`` and lesser than or equal to ``2.6.x``. If
```
adamjstewart: In Spack, alphabetic characters sort as older than numeric characters (with the exception of [infinity versions](https://github.com/spack/spack/blob/v0.19.0/lib/spack/spack/version.py#L64)). So technically, `2.4:2.6` means `2.4.a:2.6.develop`, where `2.4.a < 2.4.0`. Not sure how technical we want to get here. I would prefer "greater than or equal to `2.4`" over `2.4.0` since `2.4.0` is technically incorrect.

charmoniumQ: Good point.

adamjstewart (dismissed):
LGTM, will give @haampie a chance to review

haampie (changes requested):
I'm not a big fan of starting a section on version ranges by introducing a new concept of "partial version number", which is not a thing in Spack internals.

It's better to just introduce `@x:y` first, and then note that in some cases you can write it succinctly as `@x` (instead of @x.0:x)

lib/spack/docs/packaging_guide.rst
```diff
@@ -2325,8 +2325,13 @@ writing a package for a legacy Python module that only works with Python
    depends_on('python@2.4:2.6')
 
 Version ranges in Spack are *inclusive*, so ``2.4:2.6`` means any version
-greater than or equal to ``2.4`` and up to and including any ``2.6.x``. If
-you want to specify that a package works with any version of Python 3 (or
+greater than or equal to ``2.4`` and up to and including any ``2.6.x``.
+
+Note that while Python is versioned with three digis, like ``2.4.0``, we can
```
adamjstewart:
```suggestion
Note that while Python is versioned with three digits, like ``2.4.0``, we can
```

lib/spack/docs/packaging_guide.rst
```diff
@@ -2343,21 +2348,23 @@ requires Python 2, you can similarly leave out the lower bound:
...
-exactly at ``2.7``, which is a non-existent version. The correct way to
-specify this would be:
+since we don't care about the latter digits. You can think of this as a
+short-hand for ``@3:3``. Among equivalent specs, usually the most succinct is
+preferred.
 
-.. code-block:: python
+Note that if a version part is alphabetic, it is considered to precede
+numeric versions, and it is compared to other alphabetic sequences using
+dictionary order, unless it is a special name, so:
```
adamjstewart: We should document which "special names" are accepted. In `lib/spack/spack/version.py`, can you convert `#` to `#:` before `infinity_version`? Then you can link to it here:
```suggestion
dictionary order, unless it is a :py:data:`special name <spack.version.infinity_versions>`, so:
```

charmoniumQ: What is # and #: ?

adamjstewart: Comment and docstring comment.

adamjstewart (dismissed):
LGTM. There are other intricacies of Spack versions, but I'm not sure if they are common enough to be worth mentioning.

### Patching packages with changing number of digits

Some packages (e.g., mpich) have versions like:

* 1.0
* 1.0.1
* 1.0.2

If you use something like:
```python
patch("foo.patch", when="@1.0")
```
it will patch 1.0, 1.0.1, and 1.0.2. If you only want to patch 1.0 exactly, you have to use the non-intuitive `@1.0:1.0.a`. This is not restricted to patching, but comes up most frequently with patching.

### a/b/rc version sorting

Unlike other package managers, Spack does not give any special meaning to dev/a/b/rc as version specifiers, everything is sorted lexicographically. For example, [in Python](https://peps.python.org/pep-0440/):

* 1.0.0.dev0 < 1.0.0a < 1.0.0b < 1.0.0rc < 1.0.0 < 1.0.0.post0

whereas in Spack:

* 1.0.0 < 1.0.0a < 1.0.0b < 1.0.0.dev0 < 1.0.0.post0 < 1.0.0rc

This is important when you realize that Spack installs the most recent numeric version by default, so if you have both 1.0.0 and 1.0.0rc, Spack will install the latter. Also important when a package says that it depends on 1.0.0a+, since `@1.0.0a:` would not include 1.0.0.

Honestly, I would love to change this. The reason things are the way they are is because of packages like openssl and libjpeg that use a/b/c/d versioning. But those are far less common than packages that use a/b/rc versioning.

lib/spack/docs/packaging_guide.rst
```diff
@@ -2343,21 +2348,24 @@ requires Python 2, you can similarly leave out the lower bound:
 Notice that we didn't use ``@:3``. Version ranges are *inclusive*, so
 ``@:3`` means "up to and including any 3.x version".
 
-What if a package can only be built with Python 2.7? You might be
-inclined to use:
+If we require Python 3 exactly, we can simply write:
```
haampie:
```suggestion
If we require Python 3 exclusively, we can simply write:
```

lib/spack/docs/packaging_guide.rst
```diff
@@ -2343,21 +2348,24 @@ requires Python 2, you can similarly leave out the lower bound:
...
-However, this would be wrong. Spack assumes that all version constraints
-are exact, so it would try to install Python not at ``2.7.18``, but
-exactly at ``2.7``, which is a non-existent version. The correct way to
-specify this would be:
+since we don't care about the latter digits. You can think of this as a
+short-hand for ``@3:3``. Among equivalent specs, usually the most succinct is
+preferred.
 
-.. code-block:: python
+Note that if a version part is alphabetic, it is considered to precede
```
haampie: This is already explained in https://spack.readthedocs.io/en/latest/packaging_guide.html#version-comparison, I don't think it should be part of ranges

haampie (changes requested):

alecbcs:
@charmoniumQ did you get a chance to take a look at the review feedback. Is this PR still something you'd like to pursue?

charmoniumQ:
I may have time to work on this over my winter break. Sorry I've been slow to respond.

adamjstewart:
> Patching packages with changing number of digits

Update: Spack now has a `@=` syntax that can match an exact version of a package, so `@=1.0` won't match 1.0.1 or 1.0.2, only 1.0.

charmoniumQ (commented):
Updated

adamjstewart (approved):

bernhardkaindl (approved):

bernhardkaindl (approved):
@haampie: Your change request has been resolved, can you approve now?

Thanks!

haampie:
I would close this PR. I don't think it adds much. The bit about `@3` is just repetition of the next paragraph.

haampie:
Ok, I'll reopen, but would suggest to keep it a bit shorter. There's a lot of repetition, so then just merge the examples.

Maybe the docs were improved in the meantime, hence the duplication? If so, then would suggest again to close.

haampie:
Closing in favor of #47402

PR #35010, opened 2023-01-18
fms: add 32bit variant, add protections

mathomp4:
The GEOS Earth System Model currently requires the 32-bit variant of FMS. This PR adds that variant so we can pursue moving to use Spack.

### DIFF
--- a/var/spack/repos/builtin/packages/fms/package.py
+++ b/var/spack/repos/builtin/packages/fms/package.py
@@ -48,15 +48,18 @@ class Fms(CMakePackage):
     )
 
     variant(
-        "32bit",
-        default=True,
-        description="Build a version of the library with default 32 bit reals",
+        "precision",
+        values=('32', '64'),
+        description="Build a version of the library with default 32 or 64 bit reals or both",
+        default='64',
+        multi=True,
     )
-    variant(
-        "64bit",
-        default=False,
-        description="Build a version of the library with default 64 bit reals",
+    conflicts(
+        "precision=32,64", 
+        when="@:2022.03",
+        msg="FMS versions prior to 2022.04 do not support both 32 and 64 bit precision",
     )
+
     variant("gfs_phys", default=True, description="Use GFS Physics")
     variant("openmp", default=True, description="Use OpenMP")
     variant("quad_precision", default=True, description="quad precision reals")
@@ -84,27 +87,15 @@ class Fms(CMakePackage):
     depends_on("libyaml", when="+yaml")
 
     def cmake_args(self):
-        # If we are older than 2022.04, we allow only one of 32bit or 64bit
-        if self.spec.satisfies("@:2022.03"):
-            if "+32bit" in self.spec and "+64bit" in self.spec:
-                raise InstallError(
-                    "FMS 2022.03 and older can only build as "
-                    "either 32bit or 64bit, not both. Please choose one."
-                )
-
-        # No matter the version, a user must choose at least one of 32bit or 64bit
-        if "+32bit" not in self.spec and "+64bit" not in self.spec:
-            raise InstallError("FMS requires at least one variant of 32bit or 64bit")
-
         args = [
-            self.define_from_variant("32BIT"),
-            self.define_from_variant("64BIT"),
             self.define_from_variant("GFS_PHYS"),
             self.define_from_variant("OPENMP"),
             self.define_from_variant("ENABLE_QUAD_PRECISION", "quad_precision"),
             self.define_from_variant("WITH_YAML", "yaml"),
             self.define_from_variant("CONSTANTS"),
             self.define_from_variant("FPIC"),
+            "-D32BIT:BOOL={}".format("ON" if "precision=32" in self.spec else "OFF"),
+            "-D64BIT:BOOL={}".format("ON" if "precision=64" in self.spec else "OFF"),
         ]
 
         args.append(self.define("CMAKE_C_COMPILER", self.spec["mpi"].mpicc))
--- a/var/spack/repos/builtin/packages/fms/package.py
+++ b/var/spack/repos/builtin/packages/fms/package.py
@@ -46,7 +46,12 @@ class Fms(CMakePackage):
     version(
         "2019.01.03", sha256="60a5181e883e141f2fdd4a30c535a788d609bcbbbca4af7e1ec73f66f4e58dc0"
     )
-
+    
+    variant(
+        "32bit",
+        default=True,
+        description="Build a version of the library with default 32 bit reals",
+    )
     variant(
         "64bit",
         default=True,
@@ -80,6 +85,7 @@ class Fms(CMakePackage):
 
     def cmake_args(self):
         args = [
+            self.define_from_variant("32BIT"),
             self.define_from_variant("64BIT"),
             self.define_from_variant("GFS_PHYS"),
             self.define_from_variant("OPENMP"),
--- a/var/spack/repos/builtin/packages/fms/package.py
+++ b/var/spack/repos/builtin/packages/fms/package.py
@@ -54,7 +54,7 @@ class Fms(CMakePackage):
     )
     variant(
         "64bit",
-        default=True,
+        default=False,
         description="Build a version of the library with default 64 bit reals",
     )
     variant("gfs_phys", default=True, description="Use GFS Physics")
@@ -84,6 +84,19 @@ class Fms(CMakePackage):
     depends_on("libyaml", when="+yaml")
 
     def cmake_args(self):
+        # If we are older than 2022.04, we allow only one of 32bit or 64bit
+        if self.spec.satisfies("@:2022.03"):
+            if "+32bit" in self.spec and "+64bit" in self.spec:
+                raise InstallError(
+                    "FMS 2022.03 and older can only build as either 32bit or 64bit, not both.  Please choose one."
+                )
+
+        # No matter the version, a user must choose at least one of 32bit or 64bit
+        if "+32bit" not in self.spec and "+64bit" not in self.spec:
+            raise InstallError(
+                "FMS requires at least one variant of 32bit or 64bit"
+            )
+
         args = [
             self.define_from_variant("32BIT"),
             self.define_from_variant("64BIT"),
--- a/var/spack/repos/builtin/packages/fms/package.py
+++ b/var/spack/repos/builtin/packages/fms/package.py
@@ -88,7 +88,8 @@ class Fms(CMakePackage):
         if self.spec.satisfies("@:2022.03"):
             if "+32bit" in self.spec and "+64bit" in self.spec:
                 raise InstallError(
-                    "FMS 2022.03 and older can only build as either 32bit or 64bit, not both.  Please choose one."
+                    "FMS 2022.03 and older can only build as "
+                    "either 32bit or 64bit, not both. Please choose one."
                 )
 
         # No matter the version, a user must choose at least one of 32bit or 64bit
--- a/var/spack/repos/builtin/packages/fms/package.py
+++ b/var/spack/repos/builtin/packages/fms/package.py
@@ -46,13 +46,13 @@ class Fms(CMakePackage):
 
     variant(
         "precision",
-        values=('32', '64'),
+        values=("32", "64"),
         description="Build a version of the library with default 32 or 64 bit reals or both",
-        default='64',
+        default="64",
         multi=True,
     )
     conflicts(
-        "precision=32,64", 
+        "precision=32,64",
         when="@:2022.03",
         msg="FMS versions prior to 2022.04 do not support both 32 and 64 bit precision",
     )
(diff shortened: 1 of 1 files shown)


climbfuji (commented):
Shouldn't setting variant `64bit` to false do the same as setting `32bit` to true?

mathomp4:
> Shouldn't setting variant `64bit` to false do the same as setting `32bit` to true?

@climbfuji Not really. In GEOS we build FMS (currently in model) as *BOTH* 32bit and 64bit. (Because we use FV3 as 32bit but MOM6 is 64bit). 

And FMS [acknowledges this as doable](https://github.com/NOAA-GFDL/FMS/blob/main/CMAKE_INSTRUCTIONS.md#user-configurable-options) (emphasis mine):

> ### User configurable options:
> By default, FMS is built without `OpenMP` and in `single precision (r4)`
> 
> The 64BIT and 32BIT precision options will build _distinct libraries_ when enabled with the given default
> real size, libfms_r4 or libfms_r8.

If anything, I suppose, `32bit` should be default `True` and `64bit` default `False` if we were to echo what the defaults in FMS itself are...

climbfuji:
> > Shouldn't setting variant `64bit` to false do the same as setting `32bit` to true?
> 
> @climbfuji Not really. In GEOS we build FMS (currently in model) as _BOTH_ 32bit and 64bit. (Because we use FV3 as 32bit but MOM6 is 64bit).
> 
> And FMS [acknowledges this as doable](https://github.com/NOAA-GFDL/FMS/blob/main/CMAKE_INSTRUCTIONS.md#user-configurable-options) (emphasis mine):
> 
> > ### User configurable options:
> > By default, FMS is built without `OpenMP` and in `single precision (r4)`
> > The 64BIT and 32BIT precision options will build _distinct libraries_ when enabled with the given default
> > real size, libfms_r4 or libfms_r8.
> 
> If anything, I suppose, `32bit` should be default `True` and `64bit` default `False` if we were to echo what the defaults in FMS itself are...

That's right, FMS 2022.04 can be built in both 32bit and 64bit in parallel. I think that's not the case for previous versions. Thus, for anything older than 2022.04 only one of them must be true, the other false. For 2022.04, one or both of them must be true. Makes the logic a little more complicated, but it's doable with spack.

I don't care which precision is true by default, we always set +64bit explicitly anyway.

mathomp4:
Oh nuts. It needs to be default `True` or I change the behavior! Per FMS:

> The following build options are available:
> ```
> -DOPENMP   "Build FMS with OpenMP support"        DEFAULT: OFF
> -D32BIT    "Build 32-bit (r4) FMS library"        DEFAULT: ON
> -D64BIT    "Build 64-bit (r8) FMS library"        DEFAULT: OFF
> -DFPIC     "Build with position independent code" DEFAULT: OFF
> 
> -DCONSTANTS             "Build with <X> constants parameter definitions"     DEFAULT:GFDL  OPTIONS:GFS|GEOS|GFDL
> -DINTERNAL_FILE_NML     "Enable compiler definition -DINTERNAL_FILE_NML"     DEFAULT: ON
> -DENABLE_QUAD_PRECISION "Enable compiler definition -DENABLE_QUAD_PRECISION" DEFAULT: ON
> -DGFS_PHYS              "Enable compiler definition -DGFS_PHYS"              DEFAULT:OFF
> -DLARGEFILE             "Enable compiler definition -Duse_LARGEFILE"         DEFAULT:OFF
> -DWITH_YAML             "Enable compiler definition -Duse_yaml"              DEFAULT:OFF
> ```
> 

So this would break things for you, I think. You don't set anything with `32BIT` and I've now turned it off!

I'll make it `True` by default.

ETA: I suppose this now gives users the ability to *not* build the 32bit variant if they don't want to!

mathomp4:
Maybe we should wait for @rem1776 to chime in as to what GFDL would prefer here. Like @climbfuji when I get GEOS to work with Spack, we'll explicitly do:
```
fms@2022.04 ~gfs_phys +fpic ~quad_precision +32bit +64bit +yaml constants=GEOS
```
because we need our constants like you need yours!

climbfuji (commented):
This now enables both 64bit and 32bit, but this only works for fms@2022.04 and later. I think we need to accommodate for earlier versions and allow only one of the two in that case.

climbfuji:
> This now enables both 64bit and 32bit, but this only works for fms@2022.04 and later. I think we need to accommodate for earlier versions and allow only one of the two in that case.

Maybe the best solution is to make 64bit false by default (regardless of the fms version), and exit with an error if both are enabled for fms@2022.03 and earlier, and exit with an error if none is enabled for any version of fms.

mathomp4:
> > This now enables both 64bit and 32bit, but this only works for fms@2022.04 and later. I think we need to accommodate for earlier versions and allow only one of the two in that case.
> 
> Maybe the best solution is to make 64bit false by default (regardless of the fms version), and exit with an error if both are enabled for fms@2022.03 and earlier, and exit with an error if none is enabled for any version of fms.

Ahh. That is true. Let me see if I can code that up...

mathomp4:
> Maybe the best solution is to make 64bit false by default (regardless of the fms version), and exit with an error if both are enabled for fms@2022.03 and earlier, and exit with an error if none is enabled for any version of fms.

@climbfuji I think this is how this works with my latest update.

climbfuji (dismissed):
The logic makes sense to me, approving based on that. I am seeing style check errors that need to be addressed, though.

var/spack/repos/builtin/packages/fms/package.py
```diff
@@ -48,8 +48,13 @@ class Fms(CMakePackage):
...
 
     variant(
-        "64bit",
+        "32bit",
         default=True,
+        description="Build a version of the library with default 32 bit reals",
+    )
+    variant(
+        "64bit",
+        default=False,
```
alalazo: Can we use a `precision` multi-valued variant, with allowed values equals to `32` and `64`, and have:
```python
conflicts("precision=32,64", when="@:2022.03")
```
?

alalazo (changes requested):

mathomp4:
> The logic makes sense to me, approving based on that. I am seeing style check errors that need to be addressed, though.

I've fixed the style error I hope.

var/spack/repos/builtin/packages/fms/package.py
mathomp4: @alalazo Actually didn't even know `conflict()` was a thing! I'm learning.

@climbfuji What do you think? I'm fine as I'm not using spack operationally yet. But this would mean you'd need to change around some variants for FMS if so. I can try prototyping it out.

climbfuji: I think @alalazo's idea is great, it makes `precision` similar to the `constants` variant, and it makes handling conflicts easier. No problem to change our configs when we pull this in later.

rem1776:
> This now enables both 64bit and 32bit, but this only works for fms@2022.04 and later. I think we need to accommodate for earlier versions and allow only one of the two in that case.

@mathomp4  @climbfuji The mixed precision support goes farther back than that, it was just expanded on to more areas of the code with that release but should work with the other versions available on spack. 

The CMake build wasn't added by our team so it yields different resulting libraries ie. the distinct 32 and 64 versions. Not really sure why 32 bit is the default, but its always been that for the cmake build. We don't usually build with cmake, instead we use autotools or mkmf(gfdl-developed build tool) with default 64 and use options for 32 bit overloads if needed.

 As far as these edits go, I think the check for 2022.03 should really be for the CONSTANTS options instead since it was only added in then. It should be able to handle compiling with both defaults in older versions. The other changes look good to me though.

climbfuji:
> > This now enables both 64bit and 32bit, but this only works for fms@2022.04 and later. I think we need to accommodate for earlier versions and allow only one of the two in that case.
> 
> @mathomp4 @climbfuji The mixed precision support goes farther back than that, it was just expanded on to more areas of the code with that release but should work with the other versions available on spack.
> 
> The CMake build wasn't added by our team so it yields different resulting libraries ie. the distinct 32 and 64 versions. Not really sure why 32 bit is the default, but its always been that for the cmake build. We don't usually build with cmake, instead we use autotools or mkmf(gfdl-developed build tool) with default 64 and use options for 32 bit overloads if needed.
> 
> As far as these edits go, I think the check for 2022.03 should really be for the CONSTANTS options instead since it was only added in then. It should be able to handle compiling with both defaults in older versions. The other changes look good to me though.

Thanks @rem1776 for these insights. Do you since when building in both precisions was supported? All versions listed in the spack package?

rem1776:
> Thanks @rem1776 for these insights. Do you since when building in both precisions was supported? All versions listed in the spack package?

It's been in the CMake build since 2020.03. The only older version (2019.01.03) should really be removed from spack (maybe I'll do that after this PR) since the cmake build files weren't added yet.

mathomp4:
> > Thanks @rem1776 for these insights. Do you since when building in both precisions was supported? All versions listed in the spack package?
> 
> It's been in the CMake build since 2020.03. The only older version (2019.01.03) should really be removed from spack (maybe I'll do that after this PR) since the cmake build files weren't added yet.

@rem1776 Well, I can do that for you in this PR if you like.

Also, my guess is the separate 32 and 64 bit libraries in CMake might have come about because of GEOS. Maybe Rahul added it? We needed that because our FV3 is built as 32-bit, but MOM is 64-bit so we had to build both precisions so interfaces would match, etc. At the moment we do that with FMS inside of GEOS as as sub repo, but I'm hoping to go away from that soon (put FMS in Baselibs → Move to spack).

Now, the hope had always been that FMS could overload the functions that made this necessary and I think maybe the new-ish "mixed-mode" might do that? But, I only see a `mixed-mode` build option in the autotools build not the CMake build and all my testing with FMS 2022.04 has been building with CMake.

rem1776:
> > > Thanks @rem1776 for these insights. Do you since when building in both precisions was supported? All versions listed in the spack package?
> > 
> > 
> > It's been in the CMake build since 2020.03. The only older version (2019.01.03) should really be removed from spack (maybe I'll do that after this PR) since the cmake build files weren't added yet.
> 
> @rem1776 Well, I can do that for you in this PR if you like.
> 
> Also, my guess is the separate 32 and 64 bit libraries in CMake might have come about because of GEOS. Maybe Rahul added it? We needed that because our FV3 is built as 32-bit, but MOM is 64-bit so we had to build both precisions so interfaces would match, etc. At the moment we do that with FMS inside of GEOS as as sub repo, but I'm hoping to go away from that soon (put FMS in Baselibs → Move to spack).
> 
> Now, the hope had always been that FMS could overload the functions that made this necessary and I think maybe the new-ish "mixed-mode" might do that? But, I only see a `mixed-mode` build option in the autotools build not the CMake build and all my testing with FMS 2022.04 has been building with CMake.

@mathomp4 Yeah that would be great, thanks!

Yeah that sounds about right for the CMake options. The mixed mode we have in our autotools build might be able to handle that, it's meant to overload the available mixed precision routines rather than creating the two libraries. It just sets the default to 64 and sets CPP flags OVERLOAD_R4 and OVERLOAD_R8 so I imagine that could be added to CMake as well. 

We are actually working on expanding the mixed precision overload support in FMS to basically all of the modules, aiming for like late this year.

var/spack/repos/builtin/packages/fms/package.py
mathomp4: I think I have this now. Please take a look @alalazo

mathomp4:
All right. I think I have something now. 

For @alalazo I'm using a multi-valued variant, which seems to conflict successfully:
```
❯ spack spec -IL fms@2022.03 precision=32,64
==> Error: FMS versions prior to 2022.04 do not support both 32 and 64 bit precision
```

For @rem1776, I removed 2019.01.03 since, as you said, there was no CMake build possible in that version so it was uninstallable.

mathomp4:
Note: I'm building 32 as default, since that's what the [CMake in FMS defaults to](https://github.com/NOAA-GFDL/FMS/blob/7b6e867518369a319db3bb92a42b19bc5a2d28d8/CMakeLists.txt#L57-L61). I suppose we could default to both as well (with conditions). But I figure we can echo FMS for now.

var/spack/repos/builtin/packages/fms/package.py
```diff
@@ -80,13 +85,14 @@ class Fms(CMakePackage):
...
         args = [
-            self.define_from_variant("64BIT"),
             self.define_from_variant("GFS_PHYS"),
             self.define_from_variant("OPENMP"),
             self.define_from_variant("ENABLE_QUAD_PRECISION", "quad_precision"),
             self.define_from_variant("WITH_YAML", "yaml"),
             self.define_from_variant("CONSTANTS"),
             self.define_from_variant("FPIC"),
+            "-D32BIT:BOOL={}".format("ON" if "precision=32" in self.spec else "OFF"),
+            "-D64BIT:BOOL={}".format("ON" if "precision=64" in self.spec else "OFF"),
```
alalazo: A minor suggestion (beware of typos, since I wrote on the UI only):
```suggestion
            self.define("32BIT", "precision=32" in self.spec),
            self.define("64BIT", "precision=64" in self.spec),
```

mathomp4: @alalazo I'll try this out tonight (my home laptop is where I'm doing the testing). I tried something similar last night, but it didn't seem to work, but your code does look better than what I tried!

mathomp4: @alalazo Seems to work. I committed it (forgot I could use GitHub to do so).

alalazo (approved):
Thanks!

PR #35019, opened 2023-01-19
buildcache create: make "file exists" less verbose

haampie:
Currently we print 3 lines of "file exist" warning per tarball, this is
a bit excessive. Instead, it can be a simple single-line tty.warn
message.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -510,9 +510,9 @@ class NoOverwriteException(spack.error.SpackError):
     """
 
     def __init__(self, file_path):
-        err_msg = "\n%s\nexists\n" % file_path
-        err_msg += "Use -f option to overwrite."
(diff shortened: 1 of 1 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -1356,7 +1356,7 @@ def push(specs, push_url, specs_kwargs=None, **kwargs):
         try:
             _build_tarball(node, push_url, **kwargs)
         except NoOverwriteException as e:
-            warnings.warn(str(e))
```
alalazo: Is it this change to reduce the number of calls? I'd be curious to know why in that case.

haampie: I just thought we did not really use warnings; reverted it, I only care about the 3 lines. Actually this warning could also just be an info message

alalazo (approved):

PR #35024, opened 2023-01-19
feature: compiler as deps

haampie:
This currently only contains a commit that guards against mixing the `develop` database & user cache, so it's a bit more convenient.

```
$ cd spack
$ git fetch [upstream] features/compilers-as-deps
$ git worktree add ../spack-compilers-as-deps features/compilers-as-deps
```


zackgalbreath:
@spackbot run pipeline

PR #35027, opened 2023-01-19
Identify Windows support with tags

johnwparent:
Document which packages are supported on Windows via package 'tags' rather than a constantly out of date list in the docs.

I've tried to tag as many packages as I can recall that @jpopelar and myself have ported, however I likely missed a few so if anyone has more packages that should be tagged with `"windows"` feel free to commit to this branch or request the tag.

### DIFF
--- a/lib/spack/docs/getting_started.rst
+++ b/lib/spack/docs/getting_started.rst
@@ -1700,29 +1700,15 @@ If in the previous step, you did not have CMake or Ninja installed, running the 
 Windows Compatible Packages
 """""""""""""""""""""""""""
 
-Many Spack packages are not currently compatible with Windows, due to Unix
-dependencies or incompatible build tools like autoconf. Here are several
-packages known to work on Windows:
-
-* abseil-cpp
-* bzip2
-* clingo
-* cpuinfo
-* cmake
-* hdf5
-* glm
-* nasm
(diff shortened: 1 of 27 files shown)


adamjstewart:
Won't most CMake and Python packages automatically support Windows?

adamjstewart (dismissed):

johnwparent:
> Won't most CMake and Python packages automatically support Windows?

At the moment all Python Packages fail for me locally due to some bizarre pip issues I haven't had a chance to chase down, but once that's resolved, it really depends on the package.

In general though, yes, most likely Python and CMake packages will probably at least _in theory_ support Windows. However this is not universally true, as some packages have requirements that do not/cannot exist on Windows, and even if it is the case for a given package that it supports Windows, the Spack recipe may require tweaking or other arguments to work on Windows. For example Paraview, while a CMake package, without changes tries to enable X11 support on Windows, which is obviously problematic. Additionally Paraview has a large number of deps that don't work out of the box, so we can't say we support all CMake/Python packages out of the box/by default.

johnwparent:
How is there already a conflict I only created this branch an hour ago. Resolving now.

scheibelp (approved):

scheibelp:
@spackbot run pipeline

PR #35038, opened 2023-01-19
New bootstrap container for Rocky Linux

anderbubble:
Adds support for Rocky Linux for `spack containerize`.

I think this is all that's needed, but I haven't tested the resultant image yet; marking draft until then.

I'd also like to get this working on Rocky Linux 9; but it wasn't quite as drop-in as 8, so I'm starting with this to make sure the process is at least correct, and to answer any questions that might come up. Let me know if there's anything I've missed.

# to do

* Confirm that the resultant rockylinux8 image is working
* Update documentation


anderbubble:
I don't think I caused the failing centos7 and leap15 builds here, but let me know if I need to look more closely at those.

anderbubble:
Already done elsewhere now.

PR #35039, opened 2023-01-19
opencascade: new version 7.7.0

wdconinc:
No major build system interface changes.

- Now requires cmake 3.1, but that's hardly a requirement worth mentioning.
- New BUILD_CPP_STANDARD, but not added as variant until a need arises.


alalazo:
@spackbot fix style

alalazo (approved):

PR #35040, opened 2023-01-19
Allocate more memory for generate jobs in all stacks

zackgalbreath:
Our hope is that this will eliminate the "timeout after 6 hours" errors that we've seen in the generate jobs of our GitLab CI pipelines.

### DIFF
--- a/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
+++ b/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml
@@ -54,6 +54,7 @@ default:
     KUBERNETES_CPU_REQUEST: 4000m
     KUBERNETES_MEMORY_REQUEST: 16G
   interruptible: true
+  timeout: 60 minutes
   retry:
     max: 2
     when:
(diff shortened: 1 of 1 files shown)


scottwittenburg (dismissed):
Thanks @zackgalbreath!  I think it makes sense to bump the memory request for generate jobs across the board to prevent problems due to memory pressure.  Once jobs are no longer dying, we can attempt to collect metrics on how much memory is actually getting used by these jobs, and if warranted, reduce the requests where it makes sense.

mvandenburgh (dismissed):

haampie:
This is only relevant for `unify: false` jobs

scottwittenburg:
> This is only relevant for `unify: false` jobs

But in your PR where you treat `unify: when_possible` jobs differently (wrt resource requests), the last pipeline failed on those two jobs.  I know at the time you said it couldn't be memory or cpu anymore, and normally I trust anything you say, but until we're 100% sure of that, I think it makes sense to bump the resource requests on all of them.

scottwittenburg (approved):

PR #35042, opened 2023-01-19
Allow buildcache specs to be referenced by hash

nhanford:
## What this PR does

Currently, specs on buildcache mirrors must be referenced by their full description. This PR allows buildcache specs to be referenced by their hashes, rather than their full description.

## How it works

Hash resolution has been moved from `SpecParser` into `Spec`, and now includes the ability to execute a `BinaryCacheQuery` after checking the local store, but before concluding that the hash doesn't exist.

### Side-effects of Proposed Changes

Failures will take longer when nonexistent hashes are parsed, as mirrors will now be scanned.

## Other Changes

- `BinaryCacheIndex.update` has been modified to fail appropriately only when mirrors have been configured.
- Tests of hash failures have been updated to use `mutable_empty_config` so they don't needlessly search mirrors.
- Documentation has been clarified for `BinaryCacheQuery`, and more documentation has been added to the hash resolution functions added to `Spec`.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -419,7 +419,7 @@ class BinaryCacheIndex(object):
 
         self._write_local_index_cache()
 
-        if all_methods_failed:
+        if configured_mirror_urls and all_methods_failed:
             raise FetchCacheError(fetch_errors)
         if fetch_errors:
             tty.warn(
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -2408,6 +2408,10 @@ class BinaryCacheQuery(object):
         self.possible_specs = specs
 
     def __call__(self, spec, **kwargs):
+        """
+        Args:
+            spec (str): The spec being searched for in its string representation or hash.
+        """
         matches = []
         if spec.startswith("/"):
             # Matching a DAG hash
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -519,6 +519,11 @@ def parse_one_or_raise(
             message += color.colorize(f"@*r{{{underline}}}")
         raise ValueError(message)
 
+    if result is None:
+        message = "a single spec was requested, but none was parsed:"
+        message += f"\n{text}"
+        raise ValueError(message)
+
     return result
 
 
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -309,7 +309,7 @@ class SpecNodeParser:
         self.has_version = False
         self.has_hash = False
 
-    def parse(self, initial_spec: Optional[spack.spec.Spec] = None) -> spack.spec.Spec:
+    def parse(self, initial_spec: Optional[spack.spec.Spec] = None) -> Optional[spack.spec.Spec]:
         """Parse a single spec node from a stream of tokens
 
         Args:
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -399,6 +399,9 @@ class SpecNodeParser:
                 if not matches:
                     matches = spack.store.db.get_by_hash(dag_hash)
                 if not matches:
+                    query = spack.binary_distribution.BinaryCacheQuery(True)
+                    matches = query(self.ctx.current_token.value)
+                if not matches:
                     raise spack.spec.NoSuchHashError(dag_hash)
 
                 if len(matches) != 1:
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -258,7 +258,7 @@ class SpecParser:
         return list(filter(lambda x: x.kind != TokenType.WS, tokenize(self.literal_str)))
 
     def next_spec(
-            self, initial_spec: Optional[spack.spec.Spec] = None
+        self, initial_spec: Optional[spack.spec.Spec] = None
     ) -> Optional[spack.spec.Spec]:
         """Return the next spec parsed from text.
 
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -271,6 +271,8 @@ class SpecParser:
         root_spec = SpecNodeParser(self.ctx).parse(initial_spec)
         while True:
             if self.ctx.accept(TokenType.DEPENDENCY):
+                if root_spec.abstract_hash:
+                    raise spack.spec.RedundantSpecError(root_spec, self.ctx.current_token)
                 dependency = SpecNodeParser(self.ctx).parse(spack.spec.Spec())
 
                 if dependency == spack.spec.Spec():
@@ -390,15 +392,23 @@ class SpecNodeParser:
                 value = value.strip("'\" ")
                 initial_spec._add_flag(name, value, propagate=True)
             elif self.ctx.accept(TokenType.DAG_HASH):
-                if initial_spec.abstract_hash:
-                    msg = (
-                        f"Parsed multiple hashes /{initial_spec.abstract_hash} and "
-                        "/{self.current_token}. If you were attempting to specify a file path,"
-                        " check that it was formatted properly"
-                    )
-                    raise spack.spec.RedundantSpecError(msg, self.ctx.current_token.value)
+                # if initial_spec.abstract_hash:
+                #     msg = (
+                #         f"Parsed multiple hashes /{initial_spec.abstract_hash} and "
+                #         "/{self.current_token}. If you were attempting to specify a file path,"
+                #         " check that it was formatted properly"
+                #     )
+                #     raise spack.spec.RedundantSpecError(msg, self.ctx.current_token.value)
                 initial_spec.abstract_hash = self.ctx.current_token.value[1:]
-                self.has_hash = True
+                if self.ctx.next_token and self.ctx.next_token.kind not in [
+                    TokenType.UNQUALIFIED_PACKAGE_NAME,
+                    TokenType.FULLY_QUALIFIED_PACKAGE_NAME,
+                    TokenType.DAG_HASH,
+                    TokenType.FILENAME,
+                    TokenType.DEPENDENCY,
+                ]:
+                    raise spack.spec.RedundantSpecError(initial_spec, self.ctx.next_token)
+                break
             else:
                 break
 
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -315,7 +315,6 @@ class SpecNodeParser:
         Return
             The object passed as argument
         """
-        import spack.environment  # Needed to retrieve by hash
 
         # If we start with a package name we have a named spec, we cannot
         # accept another package name afterwards in a node
@@ -391,31 +390,15 @@ class SpecNodeParser:
                 value = value.strip("'\" ")
                 initial_spec._add_flag(name, value, propagate=True)
             elif not self.has_hash and self.ctx.accept(TokenType.DAG_HASH):
-                dag_hash = self.ctx.current_token.value[1:]
-                matches = []
-                active_env = spack.environment.active_environment()
-                if active_env:
-                    matches = active_env.get_by_hash(dag_hash)
-                if not matches:
-                    matches = spack.store.db.get_by_hash(dag_hash)
-                if not matches:
-                    query = spack.binary_distribution.BinaryCacheQuery(True)
-                    matches = query(self.ctx.current_token.value)
-                if not matches:
-                    raise spack.spec.NoSuchHashError(dag_hash)
-
-                if len(matches) != 1:
-                    raise spack.spec.AmbiguousHashError(
-                        f"Multiple packages specify hash beginning '{dag_hash}'.", *matches
-                    )
-                spec_by_hash = matches[0]
-                if not spec_by_hash.satisfies(initial_spec):
-                    raise spack.spec.InvalidHashError(initial_spec, spec_by_hash.dag_hash())
-                initial_spec._dup(spec_by_hash)
-
+                initial_spec.abstract_hash = self.ctx.current_token.value[1:]
+                print("found abstract hash: setting {}".format(self.ctx.current_token.value[1:]))
                 self.has_hash = True
             else:
                 break
+        if self.has_hash:
+            print("final parsed abstract hash is {}".format(initial_spec.abstract_hash))
+            print("final parsed spec is {}".format(initial_spec))
+            print("type of parsed spec is {}".format(type(initial_spec)))
 
         return initial_spec
 
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -391,14 +391,11 @@ class SpecNodeParser:
                 initial_spec._add_flag(name, value, propagate=True)
             elif not self.has_hash and self.ctx.accept(TokenType.DAG_HASH):
                 initial_spec.abstract_hash = self.ctx.current_token.value[1:]
-                print("found abstract hash: setting {}".format(self.ctx.current_token.value[1:]))
                 self.has_hash = True
             else:
                 break
         if self.has_hash:
             print("final parsed abstract hash is {}".format(initial_spec.abstract_hash))
-            print("final parsed spec is {}".format(initial_spec))
-            print("type of parsed spec is {}".format(type(initial_spec)))
 
         return initial_spec
 
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -389,7 +389,12 @@ class SpecNodeParser:
                 name = name.strip("'\" ")
                 value = value.strip("'\" ")
                 initial_spec._add_flag(name, value, propagate=True)
-            elif not self.has_hash and self.ctx.accept(TokenType.DAG_HASH):
+            elif self.ctx.accept(TokenType.DAG_HASH):
+                if initial_spec.abstract_hash:
+                    msg = (f"Parsed multiple hashes /{initial_spec.abstract_hash}and "
+                           "/{self.current_token}. If you were attempting to specify a file path,"
+                           " check that it was formatted properly")
+                    raise spack.spec.RedundantSpecError(msg, self.ctx.current_token.value)
                 initial_spec.abstract_hash = self.ctx.current_token.value[1:]
                 self.has_hash = True
             else:
--- a/lib/spack/spack/parser.py
+++ b/lib/spack/spack/parser.py
@@ -241,6 +241,9 @@ class TokenContext:
             return True
         return False
 
+    def expect(self, *kinds: TokenType):
+        return self.next_token and self.next_token.kind in kinds
+
 
 class SpecParser:
     """Parse text into specs"""
@@ -270,7 +273,7 @@ class SpecParser:
             The spec that was parsed
         """
         if not self.ctx.next_token:
-            return None
+            return initial_spec
 
         initial_spec = initial_spec or spack.spec.Spec()
         root_spec = SpecNodeParser(self.ctx).parse(initial_spec)
@@ -278,9 +281,9 @@ class SpecParser:
             if self.ctx.accept(TokenType.DEPENDENCY):
                 if root_spec.abstract_hash:
                     raise spack.spec.RedundantSpecError(root_spec, self.ctx.current_token)
-                dependency = SpecNodeParser(self.ctx).parse(spack.spec.Spec())
+                dependency = SpecNodeParser(self.ctx).parse()
 
-                if not dependency.abstract_hash and dependency == spack.spec.Spec():
+                if dependency is None:
                     msg = (
                         "this dependency sigil needs to be followed by a package name "
                         "or a node attribute (version, variant, etc.)"
@@ -313,7 +316,7 @@ class SpecNodeParser:
         self.has_version = False
         self.has_hash = False
 
-    def parse(self, initial_spec: spack.spec.Spec) -> spack.spec.Spec:
+    def parse(self, initial_spec: Optional[spack.spec.Spec] = None) -> Optional[spack.spec.Spec]:
         """Parse a single spec node from a stream of tokens
 
         Args:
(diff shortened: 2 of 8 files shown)


haampie (dismissed):
Can we please not make the parser do HTTP requests :confused:

The current implementation is already wrong; it shouldn't look in the active env / database.

If you really need this, can you make the parser return a non-concrete Spec with the hash set (maybe in a separate property, like `query_hash` or `user_hash` or w/e), or some wrapper/subclass. Yes, that's more work, but the PR is really the wrong direction.

That would also be useful for a lot of other things, like being able to do `spack add /xyz` without it getting expanded to a 1000 characters of spec description inside `spack.yaml`.

When these things get passed to concretize or find (in env), then they can be replaced by a concrete spec.

nhanford:
> Can we please not make the parser do HTTP requests 😕

My exact hesitation when Greg suggested putting this here ;).

> The current implementation is already wrong; it shouldn't look in the active env / database.
> 
> If you really need this, can you make the parser return a non-concrete Spec with the hash set (maybe in a separate property, like `query_hash` or `user_hash` or w/e), or some wrapper/subclass. Yes, that's more work, but the PR is really the wrong direction.
> 
> That would also be useful for a lot of other things, like being able to do `spack add /xyz` without it getting expanded to a 1000 characters of spec description inside `spack.yaml`.
> 
> When these things get passed to concretize or find (in env), then they can be replaced by a concrete spec.

Yeah that sounds pretty reasonable. If you don't mind, we can perhaps discuss when/where hash validation should occur on Slack when it's not so late in your time zone :).

becker33:
> If you really need this, can you make the parser return a non-concrete Spec with the hash set (maybe in a separate property, like query_hash or user_hash or w/e), or some wrapper/subclass. Yes, that's more work, but the PR is really the wrong direction.

I think this will cause problems. We rely pretty heavily on the invariant that a spec with a hash is concrete and vice-versa.

> Can we please not make the parser do HTTP requests 😕

I think we can avoid that without needing to have abstract specs with hashes. We could query only the local index (not check for updates from the mirrors). That would make a certain sort of sense, since anyone using a hash for a remote spec has already done something (`spack buildcache list`) that would put the spec in the local cache.

alalazo:
Ideally, I think the parser shouldn't be aware of the state of our installations, or of the hashes / files we have available. I'd really like if we had placeholders for hash and filename specifications that delay the lookup until somebody asks for concretization.

haampie:
> We rely pretty heavily on the invariant that a spec with a hash is concrete and vice-versa.

Do we really? Most commands that care about concrete specs run concretize when they get abstract specs.

lib/spack/spack/test/spec_syntax.py
```diff
@@ -631,42 +631,60 @@ def test_spec_by_hash_tokens(text, tokens):
 
 
 @pytest.mark.db
-def test_spec_by_hash(database):
+def test_spec_by_hash(database, monkeypatch, mutable_empty_config):
```
johnwparent, line 634: See https://github.com/spack/spack/pull/33385#discussion_r1040036318 for more context. This will get things passing on Windows. Concretizing `a` vs `b` doesn't matter w.r.t. this failure, so if `a` was important, you can restore that after adding this suggestion.
```suggestion
def test_spec_by_hash(database, monkeypatch, mutable_empty_config, archspec_host_is_spack_test_host):
```

johnwparent:
Ignore my previous suggestion, after a quick conversation with @scheibelp, I think the ideal approach is to remove `mutable_empty_config` entirely so the method sig is:
`def test_spec_by_hash(database, monkeypatch):`
This is slightly more optimal because:
 * Any test that concretizes things needs our mocked config (which includes fake compiler configs) 
 * The database fixture provides that functionality
 * The mutable_empty_config fixture overrides that and requires compilers just happen to be available during the CI run
 * This test does not change the config, so does not need mutable_config or mutable_empty_config
 
This should resolve things.

scheibelp:
(FYI: I have assigned myself to do a full review of this, which I can do on Monday 5/1)

lib/spack/spack/binary_distribution.py
```diff
@@ -421,7 +421,7 @@ def update(self, with_cooldown=False):
 
         self._write_local_index_cache()
 
-        if all_methods_failed:
+        if configured_mirror_urls and all_methods_failed:
```
scheibelp, line 422: (question) I'm confused about why this is now only an error when mirror URLs have been configured: is the idea that the cache may contain references to mirrors that have since been removed? If so, I think it would be ideal to clean that up somewhere before here.

lib/spack/spack/parser.py
```diff
@@ -399,26 +400,24 @@ def parse(self, initial_spec: spack.spec.Spec) -> spack.spec.Spec:
...
-
-                self.has_hash = True
+            elif self.ctx.accept(TokenType.DAG_HASH):
+                # if initial_spec.abstract_hash:
+                #     msg = (
+                #         f"Parsed multiple hashes /{initial_spec.abstract_hash} and "
+                #         "/{self.current_token}. If you were attempting to specify a file path,"
+                #         " check that it was formatted properly"
+                #     )
+                #     raise spack.spec.RedundantSpecError(msg, self.ctx.current_token.value)
```
scheibelp: (question) does this problem still exist? Have you seen it or are you worried about it as a potential issue.

scheibelp (commented):
This is preliminary: right now I just want to resolve some questions.

My primary question is: why can't the hashes be resolved at spec-parsing time (thereby avoiding the new `.abstract_hash` Spec property)?

I'm confused at least in part because `BinaryCacheQuery` at the moment has:

```
        if spec.startswith("/"):
            # Matching a DAG hash
            query_hash = spec.replace("/", "")
            for candidate_spec in self.possible_specs:
                if candidate_spec.dag_hash().startswith(query_hash):
                    matches.append(candidate_spec)
```

which indicates to me that it is at least somewhat prepared to resolve hashes into specs from the binary cache.

becker33:
@scheibelp see the earlier comments in this chain to see why the hash resolution had to be moved away from parsing time. The short version is that parsing shouldn't be hung up on failing https requests.

lib/spack/spack/binary_distribution.py
becker33: It's not a failure to fail to fetch from a mirror if there is no mirror. If there's no mirror, then we don't expect to find anything in mirrors.

haampie:
I quickly tried this PR, and I don't think it's entirely in a state where it's useful enough. Lookups still happen too early? For example, parsing multiple specs from command line args triggers lookups:

```
  File "/spack/lib/spack/spack/cmd/__init__.py", line 219, in parse_specs
    specs = spack.parser.parse(sargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/spack/lib/spack/spack/parser.py", line 473, in parse
    return SpecParser(text).all_specs()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/spack/lib/spack/spack/parser.py", line 297, in all_specs
    return list(iter(self.next_spec, spack.spec.Spec()))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/spack/lib/spack/llnl/util/lang.py", line 430, in eq
    return (other is not None) and lazy_eq(self._cmp_iter, other._cmp_iter)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/spack/lib/spack/llnl/util/lang.py", line 308, in lazy_eq
    for left, right in itertools.zip_longest(liter, riter, fillvalue=done):
  File "/spack/lib/spack/spack/spec.py", line 4100, in _cmp_iter
    cmp_spec = self.lookup_hash() or self
               ^^^^^^^^^^^^^^^^^^
  File "/spack/lib/spack/spack/spec.py", line 1869, in lookup_hash
    new = spec._lookup_hash()
          ^^^^^^^^^^^^^^^^^^^
```

1. To me it would make sense if `x.satisfies(y)` for `x` concrete and `y` abstract with partial hash is true iff x satisfies y for all other props and `x.dag_hash().startswith(y)`. That should be enough for `find` in databases*, and I don't think it should be necessary to replace the abstract spec with a concrete one in satisfies/intersects etc 🤔 (could also improve performance a bit, given that you don't have to compare a whole concrete tree). I guess the current implementation is sort of a compromise between parse-time-lookup and full abstract/concrete spec separation, but I'm worried there are just too many code paths that cause a lookup where it's not desired.

2. Another thing: `Spec.__str__` should print the (partial) hash.

3. `spack -e x add /abc` is not working for unknown `/abc` because it causes a lookup; but also doesn't work for known `/abc` because it refuses to add "anonymous" specs. Maybe better then to add a `Spec.anonymous` prop which returns `not self.name and not self.abstract_hash`.

\* just as an example, if you run `db.query_local(Spec("openmpi/xyz"))` in the current state of this PR, what happens is a double lookup: (1) the first `satisifies` call will search for `/xyz` in _any_ database, local / upstream, buildcache, even the active environment, and once found (2) it will match that concrete spec against the local database. Instead, it would be correct to ... query the local database on `openmpi/xyz`. It might be that `db` in the example is not even a configured database, so the first lookup would be nonsense.

lib/spack/spack/traverse.py
```diff
@@ -18,7 +18,7 @@
 
 
 def sort_edges(edges):
-    edges.sort(key=lambda edge: edge.spec.name)
+    edges.sort(key=lambda edge: edge.spec.name or edge.spec.abstract_hash)
```
haampie: You probably meant `(name or  '', abstract_hash or '')`, it's weird to order as `(bar, /baz, foo)`

alalazo:
Moved to v0.21, but feel free to undo if I'm missing context and this seems ready to go for v0.20

becker33:
@haampie I think that if you defer the lookup further, you'll get into some really nasty corner cases.

e.g.

```yaml
spack:
  specs:
  - matrix:
    - [hdf5, openmpi, zlib/abcdef]
    - ['%gcc@10.3.1']
    exclude:
    - ~shared
```

Now obviously this is a toy example, since no one would actually want to do that. But if you don't do a hash lookup on `zlib/abcdef` before checking whether it satisfies `~shared`, you can't know whether to produce 2 or 3 roots for that environment.

Maybe a better example (but harder to read)

```yaml
spack:
  specs:
  - matrix:
    - [trilinos ^hdf5+shared, trilinos ^hdf5~shared, trilinos ^hdf5/defabc]
    - [^openmpi, ^mpich, ^mvapich2]
    exclude:
    - ^hdf5~shared ^mvapich2
```

still somewhat of a toy problem, but now we're talking about something that a user could plausibly want.

becker33:
@spackbot fix style

haampie:
@spackbot fix style

lib/spack/spack/parser.py
tgamblin: what should we do about this commented code?

lib/spack/spack/parser.py
```diff
@@ -267,13 +269,18 @@ def next_spec(self, initial_spec: Optional[spack.spec.Spec] = None) -> spack.spe
...
         """
+        if not self.ctx.next_token:
+            return None
+
         initial_spec = initial_spec or spack.spec.Spec()
         root_spec = SpecNodeParser(self.ctx).parse(initial_spec)
         while True:
             if self.ctx.accept(TokenType.DEPENDENCY):
+                if root_spec.abstract_hash:
+                    raise spack.spec.RedundantSpecError(root_spec, self.ctx.current_token)
```
tgamblin: is this strictly needed or just to avoid the call to `satisfies()` in the parser (which would trigger a lookup)?

I am not sure we should rule out constraining abstract hashes, since it was previously a valid thing to do.  If it triggers a lookup, IMO that's on the user.  They can specify the hash *without* the additional constraint.  I think it can be useful to add the constraints if you want to verify, e.g., that your hash does meet certain criteria.  Note that we still allow `name/hash`, which is also a constraint.

Another thought here: TECHNICALLY speaking, this isn't necessarily a redundancy issue.  e.g., if you have two `zlib` installations with the same hash prefix, and they're at different versions, `/hash` may be ambiguous while `/hash @1.2` may not.

lib/spack/spack/parser.py
```diff
@@ -390,27 +396,24 @@ def parse(self, initial_spec: spack.spec.Spec) -> spack.spec.Spec:
...
+                #     )
+                #     raise spack.spec.RedundantSpecError(msg, self.ctx.current_token.value)
+                initial_spec.abstract_hash = self.ctx.current_token.value[1:]
+                if self.ctx.next_token and self.ctx.next_token.kind not in [
+                    TokenType.UNQUALIFIED_PACKAGE_NAME,
+                    TokenType.FULLY_QUALIFIED_PACKAGE_NAME,
+                    TokenType.DAG_HASH,
+                    TokenType.FILENAME,
+                    TokenType.DEPENDENCY,
+                ]:
```
tgamblin: I would move this into a `TokenContxt.expect(*tokens: TokenType) -> bool` method, and call it, e.g.:

```suggestion
                if not self.ctx.expect(
                    TokenType.UNQUALIFIED_PACKAGE_NAME,
                    TokenType.FULLY_QUALIFIED_PACKAGE_NAME,
                    TokenType.DAG_HASH,
                    TokenType.FILENAME,
                    TokenType.DEPENDENCY,
                ):
```

lib/spack/spack/traverse.py
tgamblin: I agree w/this

tgamblin (changes requested):
I think @haampie's 2 and 3 above are addressed, and 1 is addressed to the extent that it's possible.  Specifically, we made it so that `all_specs()` doesn't trigger comparisons (and therefore lookups) anymore.  @becker33: see if you agree with my simplification of your parser change there.

@haampie: I talked with @becker33 about the suggestion to reduce any `.satisfies(partial_prefix)` to `.startswith`, and @becker33 is right that doing that is going to cause `satisfies()` to be wrong sometimes.  If the partial prefix is actually ambiguous, you *can't* know whether a spec satisfies it or not, and returning `True` on prefix match may end up being wrong.  I think we have to do a lookup there.  He's got some more examples above to illustrate this, but this is the fundamental issue.

Aside from that, I have some pretty minor remaining questions -- @becker33 @nhanford please address them and I think this is good to go.

haampie:
>  If the partial prefix is actually ambiguous, you can't know whether a spec satisfies it or not, and returning True on prefix match may end up being wrong.

Hm, I disagree. It's no different from having an abstract spec `x = pkg +variant`, doing a lookup in the database to find a matching concrete spec `y = pkg +variant ...` (that satisfies `y`). Then you also have if some z satisfies x, it doesn't mean z satisfies y.

lib/spack/spack/parser.py
becker33: This is actually in to maintain previous behavior. `@1.2/hash` is allowed, but `/hash@1.2` is not.

alalazo:
> If the partial prefix is actually ambiguous, you can't know whether a spec satisfies it or not, and returning True on prefix match may end up being wrong. 

I would expect a one character abstract hash to be satisfied by any other abstract hash spec and any other concrete spec whose hash starts with that character. So, for instance:
```
abstract \ab satisfies abstract \a
abstract \ac satisfies abstract \a

concrete \abcde satisfies abstract \a
concrete \azzzz satisfies abstract \a
```
Where would this semantic break?

lib/spack/spack/parser.py
becker33: This is to avoid adding dependencies to a spec we already know by hash. While I agree that technically you could constrain an otherwise ambiguous hash by `/hash ^zlib`, I think that's something we can improve on later.

tgamblin:
@spackbot fix style

tgamblin:
Ok on further discussion, I think I agree with @haampie and @alalazo here but I also think it would be better to do that as a follow-on optimization to this.  Anyone disagree?  There is a corner case I still haven't thought through around `excludes:` in environments.

@becker33 has addressed 2 and 3; we can reduce the lookups further with the optimization.

alalazo:
> There is a corner case I still haven't thought through around excludes: in environments.

I think we can treat this as something orthogonal to the abstract hashes, and part of the matrix expansion mechanism. Basically, we should document that any abstract hash in a matrix is expanded during the matrix evaluation and must match a single known concrete spec.

tgamblin (dismissed):

lib/spack/spack/spec.py
```diff
@@ -1825,6 +1829,74 @@ def process_hash_bit_prefix(self, bits):
...
+        """Given a spec with an abstract hash, attempt to populate all properties and dependencies
+        by looking up the hash in the environment, store, or finally, binary caches.
+        This is destructive."""
+
+        if not any(node for node in self.traverse(order="post") if node.abstract_hash):
+            return
+
+        spec_by_hash = self.lookup_hash()
+
+        # if spec_by_hash != self or not self.eq_dag(spec_by_hash):
```
tgamblin: @becker33 found another piece of commented code here -- should we get rid of it?

lib/spack/spack/spec.py
```diff
@@ -1825,6 +1829,74 @@ def process_hash_bit_prefix(self, bits):
...
+        # Get dependencies that need to be replaced
+        for node in self.traverse(root=False):
+            if node.abstract_hash:
+                new = node._lookup_hash()
+                spec._add_dependency(new, deptypes=())
+
+        # reattach nodes that were not otherwise satisfied by new dependencies
+        for node in self.traverse(root=False):
+            if not any(n._satisfies(node) for n in spec.traverse()):
+                spec._add_dependency(node.copy(), deptypes=())
```
tgamblin, line 1883: I'm not 100% sure this is the right semantic, because it can return a contradiction.  Suppose you have:

```
foo ^/abc ^zlib@1.2
```

and the only lookup of `/abc` has `zlib@1.3` somewhere in the DAG.  This logic may give you a spec with a contradiction in it.  It *could* concretize if `/abc` is a build dependency or if it's `zlib` only reachable through build dependencies, since we allow multiple versions of build dependencies in a DAG. If `/abc` and its `zlib` are link dependencies, though, we know it'll fail to concretize.

On the other hand, we currently do this:

```console
> spack spec tcl ^zlib@1.2 ^zlib@1.3
==> Error: Cannot depend on incompatible specs 'zlib@1.2' and 'zlib@1.3'
```

IMO this is a useful check for users, since it's most likely a mistake, and if (as we've said) our intent is to make `^` mean "unified spec", i.e., a transitive link or run dependency of the root, not some dependency of a build dep, then it's reporting a real error early. I guess `tcl ^zlib@1.2 ^%zlib@1.3` would be ok in our future model.

The inconsistency, I guess, is that we detect this earlier at parse time than we do in `lookup_hash`.  Technically if you had a choice between two `/abc`'s, we *could* be smart enough to choose the one with the satisfiable `zlib` configuration here.  We do that already for constraints *on the hash spec*; it's just for transitive abstract hashes that we aren't considering the whole spec.  I guess this isn't so bad.  It's a corner case where we might complain about an abstract hash being ambiguous when we don't need to, but I think I could come up with cases where you'd *have* to solve to disambiguate, and I wouldn't expect a lookup to imply a solve.

Ok, I think I convinced myself that this is fine.  I just wanted to document this and get others' opinions.

lib/spack/spack/spec.py
```diff
@@ -2781,8 +2855,13 @@ def ensure_no_deprecated(root):
...
 
-        if not self.name:
-            raise spack.error.SpecError("Spec has no name; cannot concretize an anonymous spec")
+        self.replace_hash()
+
+        for node in self.traverse():
+            if not node.name:
+                raise spack.error.SpecError(
+                    f"Spec {node} has no name; cannot concretize an anonymous spec"
+                )
```
tgamblin, line 2863: I guess we don't currently model this, but I don't think it's unreasonable to concretize, say, `hdf5` with *some* dependency built with gcc, e.g.: `hdf5 ^%gcc`.  `develop` fails much less gracefully on this, though:

```console
> spack spec zlib ^%gcc
==> Error: Invalid module name:
```

vs. this branch:

```console
> spack spec zlib ^%gcc
==> Error: Spec %gcc has no name; cannot concretize an anonymous spec
```

So I guess this is an improvement and we can "fix" it later if needed.

tgamblin (changes requested):
@becker33 one last actual change request and two comments (really for you and @alalazo to look at)

tgamblin:
I pushed a commit that gets rid of the commented code, as I think it's correct to do that.

`git diff develop... | grep '^+\s*#'` doesn't return anything so I think we're done with commented code for real this time.

@alalazo @becker33 I'm ok with this PR so if you agree with the changes above and have no further objections, feel free to merge.  Again, we can add the hash optimization after.

tgamblin (approved):

lib/spack/spack/spec.py
```diff
@@ -1825,6 +1829,73 @@ def process_hash_bit_prefix(self, bits):
         """Get the first <bits> bits of the DAG hash as an integer type."""
         return spack.util.hash.base32_prefix_bits(self.process_hash(), bits)
 
+    def _lookup_hash(self):
+        """Lookup just one spec with an abstract hash, returning a spec from the the environment,
+        store, or finally, binary caches."""
+        import spack.environment
```
alalazo, line 1835: This needs to be refactored at some point, after v0.20, to try disentangling concepts. The design issue I see is that a spec needs to know about all the global objects where specs can be stored to perform a lookup.

alalazo (commented):
I think we can improve the current code, but I agree merging this and have following PRs after v0.20 I'll leave to @becker33 to give a final look

lib/spack/spack/spec.py
tgamblin: Yep agree with this. I think having some central place for managing the world of known Specs would be useful.

becker33: I like this idea for improvement, but I'm having trouble thinking where else this information would belong, it's not really a natural fit anywhere. It could live in `spack.spec` but outside the `Spec` class, but that doesn't feel satisfying really.

PR #35043, opened 2023-01-19
update Samtools, bcftools, and htslib versions

downloadico:


var/spack/repos/builtin/packages/bcftools/package.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```
kwryankrattiger, line 1: Fixup the copyright headers

kwryankrattiger:
LGTM after the copyright headers are reverted to the 2023.

var/spack/repos/builtin/packages/bcftools/package.py
downloadico: OOPS!  We *did* increment the year recently, didn't we?! ;)

tgamblin:
We can merge this and fix that later.

var/spack/repos/builtin/packages/samtools/package.py
```diff
@@ -41,6 +41,7 @@ class Samtools(Package):
     depends_on("python", type="run")
 
     # htslib became standalone @1.3.1, must use corresponding version
+    depends_on("htslib@1.16", when="@1.16.1")
```
kwryankrattiger, line 44: @downloadico It looks like you added the dependency here, but not the new version :)

PR #35044, opened 2023-01-19
py-sphobjinv: add v2.3.1

adamjstewart:
Successfully installs on macOS 10.15.7 with Python 3.10.8 and Apple Clang 12.0.0.

### DIFF
--- a/var/spack/repos/builtin/packages/py-sphobjinv/package.py
+++ b/var/spack/repos/builtin/packages/py-sphobjinv/package.py
@@ -10,13 +10,18 @@ class PySphobjinv(PythonPackage):
     """Sphinx objects.inv Inspection/Manipulation Tool."""
 
     homepage = "https://github.com/bskinn/sphobjinv"
-    url = "https://github.com/bskinn/sphobjinv/archive/refs/tags/v2.1.tar.gz"
+    pypi = "sphobjinv/sphobjinv-2.3.1.tar.gz"
 
-    version("2.1", sha256="e41950a578dfd5acae24f12c7fe12b8d5e44f9162487aaa27189ca2e5c45d30c")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-sphobjinv/package.py
```diff
@@ -10,13 +10,18 @@ class PySphobjinv(PythonPackage):
...
+        deprecated=True,
+    )
 
-    depends_on("python@3.6:", type=("build", "run"))
     depends_on("py-setuptools", type=("build", "run"))
     depends_on("py-attrs@19.2:", type=("build", "run"))
     depends_on("py-certifi", type=("build", "run"))
-    depends_on("py-fuzzywuzzy@0.8:", type=("build", "run"))
     depends_on("py-jsonschema@3.0:", type=("build", "run"))
+    depends_on("py-fuzzywuzzy@0.8:", when="@2.1", type=("build", "run"))
```
tldahlgren, line 27: 🤣 Love the package name!

tldahlgren (approved):
Confirmed the sha256 for the new and updated versions.

var/spack/repos/builtin/packages/py-sphobjinv/package.py
adamjstewart: It's actually super useful, does fuzzy matching of strings. Been thinking about using it in Spack so Spack can suggest alternate names when you misspell a package.

bskinn: Be warned, `fuzzywuzzy` is GPL, if it matters. I had to rip it out of `sphobjinv` in v2.2 because I wanted to keep the latter MIT.

I have a lot I want to do around speedups and custom search functions for `sphobjinv suggest` -- I'm planning on looking at, e.g., [`rapidfuzz`](https://pypi.org/project/rapidfuzz/) when the time comes.

PR #35045, opened 2023-01-19
Add a new `drop` directive for extending packages

tgamblin:
Closes #33506; related to #33057.

When you inherit from a package in Spack, you inherit all the metadata from its directives, including `version`, `provides`, `depends_on`, `conflicts`, etc.

In some cases, you may not want this metadata. This PR adds a new `drop` directive that allows you to throw out things inherited from the base class.  For example:

```python
from spack.pkg.builtin.mpich import Mpich

class MyMpich(Mpich):
    drop("versions")    # don't inherit any versions from builtin Mpich
    version("5.0", "08721a102fefcea2ae4add8c9cc548df77e9224f5385ad0872a9150fdd26a415")
    version("6.0", "9cc39dd33dd4227bb82301d285437588d705290846d22ab6b8791c7e631ce385")
```

Without the `drop("versions")` directive, this package would have versions `5.0`, `6.0`, *and* anything inherited from `Mpich`. With it, this package has only versions `5.0` and `6.0`.

You can `drop` any of: `conflicts`, `dependencies`, `extendees`, `patches`, `provided`, `resources`, `variants`, or `versions`.

- [x] add new `drop` directive
- [x] Two packages were modifying their `versions` dictionary in their constructors to achieve this, but this causes `spack url stats` to fail with a concurrent modification exception as it iterates over all packages. Fixed these to use `drop` instead.
- [x] Update documentation
- [x] Add tests

### DIFF
--- a/lib/spack/docs/packaging_guide.rst
+++ b/lib/spack/docs/packaging_guide.rst
@@ -2410,7 +2410,7 @@ package, and a `canonical hash <https://github.com/spack/spack/pull/28156>`_ of
 the ``package.py`` recipes). ``test`` dependencies do not affect the package
 hash, as they are only used to construct a test environment *after* building and
 installing a given package installation. Older versions of Spack did not include
-build dependencies in the hash, but this has been 
+build dependencies in the hash, but this has been
 `fixed <https://github.com/spack/spack/pull/28504>`_ as of |Spack v0.18|_.
 
 .. |Spack v0.18| replace:: Spack ``v0.18``
@@ -3603,6 +3603,70 @@ for instance:
 In the example above ``Cp2k`` inherits all the conflicts and variants that ``CudaPackage`` defines.
 
 .. _install-environment:
+
+--------------------------------
+Package Inheritance
+--------------------------------
+
+Spack packages are Python classes, and you can use inheritance with them just as you can
+with any Python class. This is common when you have your own package :ref:`repository
+<repositories>` with packages that extend Spack's ``builtin`` packages.
+
+You can extend a ``builtin`` package like this:
+
+.. code-block:: python
+
+   from spack.pkg.builtin.mpich import Mpich
+
+   class MyPackage(Mpich):
+       version("1.0", "0209444070d9c8af9b62c94095a217e3bc6843692d1e3fdc1ff5371e03aac47c")
+       version("2.0", "5dda192154047d6296ba14a4ab2d869c6926fd7f44dce8ce94f63aae2e359c5b")
+
+Every repository registered with Spack ends up in a submodule of ``spack.pkg`` with a
+name corresponding to its :ref:`namespace <namespaces>`. So, if you have a different
+repository with namespace ``myrepo`` you want to import packages from , you might write:
+
+.. code-block:: python
+
+   from spack.pkg.myrepo.my_package import MyPackage
+
+   class NewPackage(MyPackage):
+       version("3.0", "08721a102fefcea2ae4add8c9cc548df77e9224f5385ad0872a9150fdd26a415")
+       version("4.0", "9cc39dd33dd4227bb82301d285437588d705290846d22ab6b8791c7e631ce385")
+
+^^^^^^^^^^^^^^^^^^^^^^^^
+``disinherit``
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+When you inherit from a package in Spack, you inherit all the metadata from its
+directives, including ``version``, ``provides``, ``depends_on``, ``conflicts``, etc. For
+example, ``NewPackage`` above will have four versions: ``1.0`` and ``2.0`` inherited
+from ``MyPackage``, as well as, ``3.0``, and ``4.0`` defined in ``NewPackage``.
+
+If you do not want your package to define all the same things as its base class, you can
+use the ``disinherit`` directive to start fresh in your subclass:
+
+.. code-block:: python
+
+   from spack.pkg.myrepo.my_package import MyPackage
+
+   class NewerPackage(MyPackage):
+       disinherit("versions")    # don't inherit any versions from MyPackage
+       version("5.0", "08721a102fefcea2ae4add8c9cc548df77e9224f5385ad0872a9150fdd26a415")
+       version("6.0", "9cc39dd33dd4227bb82301d285437588d705290846d22ab6b8791c7e631ce385")
+
+Now, ``NewerPackage`` will have **only** versions ``5.0`` and ``6.0``, and will not
+inherit ``1.0`` or ``2.0`` from ``MyPackage``. You can ``disinherit`` many different
+properties from base packages. The full list of options is:
+
+* ``conflicts``
+* ``dependencies``
+* ``extendees``
+* ``patches``
+* ``provided``
+* ``resources``
+* ``variants``
+* ``versions``
 
 -----------------------
 The build environment
--- a/lib/spack/spack/directives.py
+++ b/lib/spack/spack/directives.py
@@ -53,6 +53,7 @@ __all__ = [
     "version",
     "conflicts",
     "depends_on",
+    "disinherit",
     "extends",
     "provides",
     "patch",
@@ -235,12 +236,12 @@ class DirectiveMeta(type):
         if isinstance(dicts, str):
             dicts = (dicts,)
 
-        if not isinstance(dicts, collections.abc.Sequence):
-            message = "dicts arg must be list, tuple, or string. Found {0}"
-            raise TypeError(message.format(type(dicts)))
+        if dicts is not None:
+            if not isinstance(dicts, collections.abc.Sequence):
+                raise TypeError(f"dicts arg must be list, tuple, or string. Found {type(dicts)}")
 
-        # Add the dictionary names if not already there
-        DirectiveMeta._directive_dict_names |= set(dicts)
+            # Add the dictionary names if not already there
+            DirectiveMeta._directive_dict_names |= set(dicts)
 
         # This decorator just returns the directive functions
         def _decorator(decorated_function):
@@ -765,6 +766,35 @@ def build_system(*values, **kwargs):
         default=default,
         multi=False,
     )
+
+
+@directive()
+def disinherit(dict_name: str):
+    """Clear all values in a dict inherited from base packages.
+
+    You can use this to, e.g., remove all inherited versions from a base package:
+
+        disinherit("versions")  # this package doesn't share any versions w/parents
+        version("2.0", ...)     # new versions specific to this package
+        version("3.0", ...)
+
+    The dictionary name is checked, so you can't call this on somethign that's not a
+    valid directive dictonary.
+
+    Arguments:
+        dict_name: name of directive dictionary to clear.
+
+    """
+    if dict_name not in DirectiveMeta._directive_dict_names:
+        names = ", ".join(DirectiveMeta._directive_dict_names)
+        raise DirectiveError(f"Can't disinherit '{dict_name}'. Options are: {names}")
+
+    def _execute_disinherit(pkg):
+        dictionary = getattr(pkg, dict_name, None)
+        if dictionary:
+            dictionary.clear()
+
+    return _execute_disinherit
 
 
 class DirectiveError(spack.error.SpackError):
--- a/lib/spack/spack/test/directives.py
+++ b/lib/spack/spack/test/directives.py
@@ -68,3 +68,21 @@ def test_error_on_anonymous_dependency(config, mock_packages):
     pkg = spack.repo.path.get_pkg_class("a")
     with pytest.raises(spack.directives.DependencyError):
         spack.directives._depends_on(pkg, "@4.5")
+
+
+def test_disinherited_version(config, mock_packages):
+    pkg = spack.repo.path.get_pkg_class("disinherit")
+
+    # these would be inherited from A if disinherit failed
+    assert "1.0" not in pkg.versions
+    assert "2.0" not in pkg.versions
+
+    # these are defined in dininherit
+    assert "3.0" not in pkg.versions
+    assert "4.0" not in pkg.versions
+
+
+def test_disinherit(config, mock_packages):
+    with pytest.raises(spack.directives.DirectiveError):
+        # This is not a valid thing to disinherit
+        spack.directives.disinherit("foobarbaz")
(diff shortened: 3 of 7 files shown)


tgamblin:
Would like to get some thoughts on this before it gets merged, so don't just merge it after one approval :)

w8jcik:
The change looks great. Thank you for the effort that you put into adding it.

Coming to implementation, disinherit could accept predicate what to remove, but it would be also unnecessary complication. Current solution covers my use case and looks simple.

So again, it looks good to me :)

lib/spack/spack/directives.py
```diff
@@ -767,6 +768,35 @@ def build_system(*values, **kwargs):
     )
 
 
+@directive()
+def disinherit(dict_name: str):
```
alalazo, line 772: Can we call this "reset" or "clear"? "Disinherit" seems a bit too tied to the implementation for a package directive

alalazo (commented):
I think this works very well for simple packages. 

The current implementation though require  users to duplicate code, if they want to get rid of e.g. a single variant. This is fine for small packages, less so if we want to remove a single variant, or conflict, from a package like `trilinos`.

Another consideration is that methods inherited from base classes might refer to entities that got pruned, and produce some inconsistent behavior if people don't carefully check the recipe.

lib/spack/spack/directives.py
tgamblin: I really like `disinherit` because the user *has* to know that the versions are inherited for this to be useful (i.e., you wouldn't use it to reset version declarations you made in the same class).  I don't see how this can't be tied to that bit of implementation.

alalazo: In my mind "disinherit" means we do something against OO inheritance.

tgamblin:
> The current implementation though require users to duplicate code, if they want to get rid of e.g. a single variant.

I was thinking of adding:

```python
disinherit("versions", except=["1.0", "2.0", ...])
```

and, e.g.:

```python
# keeps all provides("mpi...") (by matching the spec)
disinherit("provided", except="mpi")
```

and:

```python
# get rid of any conflicts with hdf5 > 10
disinherit("conflicts", except="hdf5@:10")
```

Of the things you could disinherit:
* Versions have obvious identifiers (the version) and also have `satisfies()`, e.g. `disinherit("versions", except="10:")`
* Variants and resources have names
* Conflicts, dependencies, extendees, and provided can be matched with spec semantics
* Patches I don't really know what to do with -- you could specify the sha256, the URL, or the path but they all seem cumbersome.  I would lean towards just disallowing `except=` for patches unless we really need this.

If that sounds good it wouldn't be hard to throw something together.

becker33:
This looks substantially like a duplicate of https://github.com/spack/spack/pull/33506

> Patches I don't really know what to do with -- you could specify the sha256, the URL, or the path but they all seem cumbersome. I would lean towards just disallowing except= for patches unless we really need this.

We definitely need to be able to disinherit patches, but I don't off the top of my head have a use-case for `except`

tgamblin:
@becker33: Thanks -- I guess I missed #33506.  Sorry @iarspider -- I didn't mean to step on your toes.

Personally I like this better b/c it's a single directive, not a bunch of them, and I think we need to be able to drop all of something and also drop specific things with `except=`.  I kind of like `drop()` from #33506 instead of `disinherit()` -- would that alleviate your concern @alalazo?

I would be curious what @iarspider thinks of this -- with the changes mentioned in https://github.com/spack/spack/pull/35045#issuecomment-1398742374 would this work for you?

I think I can bring over the filename/url naming for patches from #33506 and adapt @iarspider's test too.

lib/spack/spack/directives.py
tgamblin: How about `drop()`?  See the top-level comment below.

On OO: We only use inheritance in Spack packages for sharing (some) metadata and for passing along the build mechanism.  It is inheritance, but the directives and packager-implemented methods are not the polymorphic part.

For example, we don't have functions that take an `Mpich` and expect to be able to do the same things with any subclass of `Mpich`. The main client code for packages is the solver and the installer... and they only require that the same metadata dictionaries are present and that the phases/do_install API works.  So I don't think we're violating polymorphism here, or at least we never really had it to begin with.

The other clients for packages are other packages that depend on them... but the only polymorphism you'd expect there is if a package in one repository has the same name as a package in a lower-precedence repo.  In that situation the client package *might* ask for one of the versions in the base package.  But if it can't find it that's not awful... the resolver just says the version doesn't exist... and that's something that happens in the normal course of maintaining things.

alalazo: > How about drop()? See the top-level comment below.

:+1:  "drop" sounds better to me.

alalazo: > So I don't think we're violating polymorphism here, or at least we never really had it to begin with.

Sure, I didn't mean to say that (and if it was the case, it wouldn't be solved by just renaming :slightly_smiling_face: ). What I wanted to say is that it "sounds like it has to do" with undoing something related to OO, when in fact it isn't related to that.

iarspider:
@tgamblin I would prefer the opposite syntax - `disinherit(..., only=...)`, since I rarely need to drop more than one of something. Or maybe both, while obv not allowing things like `disinherit(..., only=..., except=...)` :)

tgamblin:
@iarspider: We were thinking we'd go with your `drop()` over `disinherit` -- is that ok?

I don't mind doing both `drop(..., only="")` and `drop(..., except="")`... I suspect someone will ask for both eventually.

iarspider:
Sure. I don't care about the actual name.

becker33:
@tgamblin here is the use case for this that we discussed on our call earlier:

On RHEL systems, RHEL provides special `gcc@x.y.1` versions that include substantial work to make them binary compatible with the system gcc.

We want to be able to install e.g. `gcc@103.1` from a tarball of binaries to take advantage of this work. This requires being able to drop all dependencies (because they are irrelevant) and patches (they won't apply) for that one particular version (eventually probably a list of versions, to include `gcc@11.3.1` as well).

The ideal interface for this use case would probably be something like:

```
with when("@10.3.1"):
    drop("patches")
    drop("dependencies")
```

scheibelp:
The use case that originally drew my interest to this PR does not require `when`: if there is a package in the `builtin` repository like:

```
class Foo(Package):
    depends_on(cmake@:upper.version)
    ...
    # a bunch of useful logic
```

* I want to inherit `Foo`, but my DAG (i.e. some higher dependent) needs a CMake version after `upper.version`
* I experiment personally with installing `Foo` using that later CMake and find that it works

I can submit a PR to loosen the constraints in `Foo` but:

* To use that I pull the latest state of `builtin` (along with possibly other incompatible changes to other packages)
* While waiting I need to maintain this change as a patch to the `Foo` package (e.g. a `.patch` file, or a separate branch of Spack with those changes applied)

So it would be nice to inherit `Foo` but discard the constraints that aren't necessary:

```
class Foo(builtin.Foo):
    drop("dependencies")
```

ideally, we could drop only specific dependencies and keep a subset of them, but for this use case it shouldn't be a major issue to rewrite the `depends_on` constraints that I want to keep.

alalazo:
I have a couple of use cases where `drop` would be useful. One is build system specific:
- https://github.com/spack/spack/pull/35552#pullrequestreview-1304036540

In that case I am using `conflicts` to hide some weird implementation details when we have to drop one of the two generators. A better implementation would be using `drop` to drop the `generator` variant, and any metadata dependent on the presence of the value that isn't used.

The other one is a from a user (can't find the PR right now). That user wanted to override for a package the list of supported `cuda_arch`, but needs to get rid to all the metadata that reference them (which  is inherited from the base class).

danielsjensen1:
> I think this works very well for simple packages.
> 
> The current implementation though require users to duplicate code, if they want to get rid of e.g. a single variant. This is fine for small packages, less so if we want to remove a single variant, or conflict, from a package like `trilinos`.
> 
> Another consideration is that methods inherited from base classes might refer to entities that got pruned, and produce some inconsistent behavior if people don't carefully check the recipe.

I put in an implementation that allows for the removal of individual directives in addition to entire categories at https://github.com/spack/spack/pull/48947 as that has primarily been my use case.  I often need to remove just one conflict when I'm inheriting a package.  I didn't see this issue earlier so I was dumb and wrote a completely separate implementation.

lib/spack/docs/packaging_guide.rst
```diff
@@ -3604,6 +3604,70 @@ In the example above ``Cp2k`` inherits all the conflicts and variants that ``Cud
...
+
+Spack packages are Python classes, and you can use inheritance with them just as you can
+with any Python class. This is common when you have your own package :ref:`repository
+<repositories>` with packages that extend Spack's ``builtin`` packages.
+
+You can extend a ``builtin`` package like this:
+
+.. code-block:: python
+
+   from spack.pkg.builtin.mpich import Mpich
```
becker33, line 3619:
```suggestion
   from spack_repo.builtin.packages.mpich.package import Mpich
```

lib/spack/docs/packaging_guide.rst
```diff
@@ -3604,6 +3604,70 @@ In the example above ``Cp2k`` inherits all the conflicts and variants that ``Cud
...
+       version("1.0", "0209444070d9c8af9b62c94095a217e3bc6843692d1e3fdc1ff5371e03aac47c")
+       version("2.0", "5dda192154047d6296ba14a4ab2d869c6926fd7f44dce8ce94f63aae2e359c5b")
+
+Every repository registered with Spack ends up in a submodule of ``spack.pkg`` with a
+name corresponding to its :ref:`namespace <namespaces>`. So, if you have a different
+repository with namespace ``myrepo`` you want to import packages from , you might write:
+
+.. code-block:: python
+
+   from spack.pkg.myrepo.my_package import MyPackage
```
becker33, line 3631:
```suggestion
   from spack_repo.myrepo.packages.my_package.package import MyPackage
```

lib/spack/docs/packaging_guide.rst
```diff
@@ -3604,6 +3604,70 @@ In the example above ``Cp2k`` inherits all the conflicts and variants that ``Cud
...
+
+.. code-block:: python
+
+   from spack.pkg.builtin.mpich import Mpich
+
+   class MyPackage(Mpich):
+       version("1.0", "0209444070d9c8af9b62c94095a217e3bc6843692d1e3fdc1ff5371e03aac47c")
+       version("2.0", "5dda192154047d6296ba14a4ab2d869c6926fd7f44dce8ce94f63aae2e359c5b")
+
+Every repository registered with Spack ends up in a submodule of ``spack.pkg`` with a
```
becker33, line 3625:
```suggestion
Every repository registered with Spack ends up in a submodule of ``spack_repo`` with a
```

lib/spack/docs/packaging_guide.rst
```diff
@@ -3604,6 +3604,70 @@ In the example above ``Cp2k`` inherits all the conflicts and variants that ``Cud
...
+properties from base packages. The full list of options is:
+
+* ``conflicts``
+* ``dependencies``
+* ``extendees``
+* ``patches``
+* ``provided``
+* ``resources``
+* ``variants``
+* ``versions``
```
becker33, line 3669: 1. This is out of date in the directive name.
2. This talks about the composition of the versions as if no versions come from `mpich`.
3. We should discuss an example where you only want to remove one version -- in that case, you have to remove them all and reimplement the ones you want to remove

lib/spack/spack/directives.py
becker33, line 772: We agreed to change this to `drop`

lib/spack/spack/test/directives.py
```diff
@@ -68,3 +68,21 @@ def test_error_on_anonymous_dependency(config, mock_packages):
...
+    # these would be inherited from A if disinherit failed
+    assert "1.0" not in pkg.versions
+    assert "2.0" not in pkg.versions
+
+    # these are defined in dininherit
+    assert "3.0" not in pkg.versions
+    assert "4.0" not in pkg.versions
+
+
+def test_disinherit(config, mock_packages):
```
becker33, line 85:
```suggestion
def test_disinherit_invalid(config, mock_packages):
```

var/spack/repos/builtin.mock/packages/disinherit/package.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+from spack.pkg.builtin.mock.a import A
```
becker33, line 7:
```suggestion
from spack_repo.builtin_mock.packages.a.package import A
```

becker33 (changes requested):
There are two levels of additions to this that I think are necessary, but I don't know that they need to be part of this PR.

1. The use case to tighten a when constraint:

If we have a builtin package `foo` with some dependencies

```
class Foo(Package):
    depends_on("bar", when="+extra_capabilities")
    ... # 5000 lines of additional dependencies
```
and a user discovers that `bar` is not actually necessary for `+extra_capabilities` after some version cut-off X.Y.Z. Then the user will want to drop the specific dependency on "bar" without having to recreate the rest of the dependencies, so that they can recreate the constrained `bar` dependencies.

```
from spack_repo.builtin.packages.foo.package import Foo as BuiltinFoo
class Foo(BuiltinFoo):
    drop("dependencies", "bar")
    depends_on("bar", when="@:X.Y +extra_capabilities")
```

2. The further case that we've discussed more, regarding `when` clauses on the drop directive

I'm fully convinced that (2) should wait for a follow-on PR (if we determine it's necessary. But (1) is much more essential to the use case in my mind. It seems we would want to be able to drop:

1. A particular version by its version string
2. A conflict by its first parameter (arbitrarily, this one is less clear how it maps to user intuition than the others
3. A dependency by either its name or by a matching spec (not sure which is better)
4. An extension by the same as a dependency
5. A provider by the same as a dependency (more likely name in this case)
6. A patch by its name/url
7. A variant by its name
8. A resource by its name/url
9. (no need for dropping a specific build_system, since it's implemented as a variant)

All of these seem reasonably easy to match on if we can settle on a clear interface and documentation of what the additional argument does in terms of constraining the `drop`.

The obvious next step of conditions/exceptions to the constraint is clearly a future work issue.

tgamblin:
This is likely to be superseded by #48947 for v1.2. Closing in anticipation of that; we can reopen if it becomes interesting again.

PR #35046, opened 2023-01-19
libfuse: Check libfuse3.so/.a when running `spack external find`

bernhardkaindl:
Hello @haampie, I hope this is OK for you!

Many of my hosts don't have `libfuse3-dev` / `fuse-devel` installed, so I ran into this while actually reviewing another PR:

Having had used `spack external find --all` long before, it had added `libfuse` from the system to `packages.yaml`.

But this fuse had no dev-libraries and no headers (these are in the `libfuse3-dev` / `fuse-devel` packages, which weren't installed)

Thus, I want to propose a determine_version(cls, lib), derived from other library packages, which check the provided libs, and:

```diff
+    maintainer = ["bernhardkaindl"]
```
Here is the new `determine_version(cls, lib)`, works correctly even on Debian/Ubuntu with their libs in `lib/{aarch,x86-}64-linux-gnu` dirs:
```py
    @classmethod
    def determine_version(cls, lib):
        # When checking the .so lib, we can get the version from the realpath's extension
        if lib.endswith(".so"):
            match = re.search(r"lib\S*\.so\.(\d+)\.(\d+)\.(\d+)", os.path.realpath(lib))
            if match:
                return f"{match.group(1)}.{match.group(2)}.{match.group(3)}"
        # Otherwise or if no match, we have to get the version from the matching pkgconfig file
        if lib.endswith(".so") or lib.endswith(".a"):
            pkgconfig = f"{os.path.dirname(lib)}/pkgconfig/{os.path.basename(lib)[3:][:-2]}.pc"
            try:
                return re.search(r"Version: (\S+)", open(pkgconfig).read()).group(1)
            except FileNotFoundError:
                return None
        # Linking needs .so/.a (and headers for compiling sources would then also be missing)
        return None
```
Since this is pretty generic, a next step once this method is approved would be to turn this into a callable library function.

### DIFF
--- a/var/spack/repos/builtin/packages/libfuse/package.py
+++ b/var/spack/repos/builtin/packages/libfuse/package.py
@@ -14,6 +14,7 @@ class Libfuse(MesonPackage):
     Userspace) interface"""
 
     homepage = "https://github.com/libfuse/libfuse"
+    maintainer = ["bernhardkaindl"]
     url = "https://github.com/libfuse/libfuse/releases/download/fuse-2.9.9/fuse-2.9.9.tar.gz"
 
     version("3.11.0", sha256="25a00226d2d449c15b2f08467d6d5ebbb2a428260c4ab773721c32adbc6da072")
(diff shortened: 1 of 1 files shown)


haampie:
Hi @bernhardkaindl, this package-related pull request can be migrated from `spack/spack` to `spack/spack-packages` using the [**migrate-pkg-prs utility**](https://github.com/spack/migrate-package-prs).

---

PR #35050, opened 2023-01-20
gmake: bootstrap without make

haampie:
Would be nice to have a gmake that bootstraps in a few seconds

Also fixes the build of `gmake +guile`


alalazo (approved):

haampie:
@spackbot run pipeline

PR #35052, opened 2023-01-20
Restore the openspeedshop and openspeedshop-utils updating the cbtf components that support openspeedshop.

jgalarowicz:
Restore the openspeedshop and openspeedshop-utils updating the cbtf components that support openspeedshop.
They were removed because we hadn't updated to python3 and I didn't have the maintainer statement in the package.py files.


adamjstewart:
I don't know anything about any of these packages so it would be better for someone else to review them. I'm only maintaining Python/ML/Geo packages in Spack.

alalazo (approved):
Let's restore these packages, with an added maintainer. Since they are quite self-contained and don't impact other packages much, we can improve them later with smaller, more focused, PRs.

PR #35057, opened 2023-01-20
Bugfix: package requirements with git commits

scheibelp:
Specs which define 'new' versions in the require: section need to generate associated facts to indicate that those versions are valid.

(Update: since addressed, although pending further discussion) ~There's a couple conceptual issues at the moment (noted as TODOs in the code).~


scheibelp:
@spackbot help

scheibelp:
@spackbot fix style

lib/spack/spack/test/concretize_requirements.py
```diff
@@ -124,6 +124,62 @@ def test_requirement_isnt_optional(concretize_scope, test_repo):
...
+""".format(
+        commits[0], commits[1]
+    )
+    update_packages_config(conf_str)
+
+    s1 = Spec("v").concretized()
+    assert s1.satisfies("@2.2")
+
+    # TODO: this fails now and it shouldn't
+    # s2 = Spec("v@{0}".format(commits[1])).concretized()
```
scheibelp: Specifically the concretization fails with:

```
>       raise UnsatisfiableSpecError(msg)
E       spack.solver.asp.UnsatisfiableSpecError: No version for 'v' satisfies '@917660c3732b09cc2d1dfffcaa1cbc94b4f872a7=2.2' and '@f9f34e773aef46fef1d90a60694ae3690f6e40c7'
```

The following works:

```
Spec("v@2.3").concretized()
v@commit1=2.3%gcc@10.2.1 build_system=generic arch=test-debian6-core2
```

(where `commit1` is the full commit hash: `commits[1]`)

scheibelp: Also, saying

```
Spec("v@commit1=2.3").concretized()
```

works.

lib/spack/spack/solver/asp.py
```diff
@@ -2070,6 +2070,48 @@ def setup(self, driver, specs, reuse=None):
...
+
+        config = spack.config.get("packages")
+        for pkg_name, d in config.items():
+            if pkg_name == "all":
+                continue
+            if "require" in d:
+                req_version_specs.extend(specs_from_requires(pkg_name, d["require"]))
+
+        # TODO: if I use version_provenance.packages_yaml, the associated test fail
+        self.add_concrete_versions_from_specs(req_version_specs, version_provenance.spec)
```
becker33: I think it would be best to factor this section out into a method that we call here, simply to keep the top-level method easier to read and keep the low-level concerns in targeted methods.

lib/spack/spack/test/concretize_requirements.py
```diff
@@ -124,6 +124,62 @@ def test_requirement_isnt_optional(concretize_scope, test_repo):
...
+    monkeypatch.setattr(
+        spack.package_base.PackageBase, "git", "file://%s" % repo_path, raising=False
+    )
+
+    a_commit_hash = commits[0]
+    conf_str = """\
+packages:
+  v:
+    require: "@{0}=2.2"
+    version: ["2.2"]
```
becker33: We should add documentation about this being necessary

becker33 (dismissed):
mostly LGTM

lib/spack/spack/test/concretize_requirements.py
scheibelp: Added in https://github.com/spack/spack/pull/35057/commits/8886e3d6f2b01db494013c2abea7b082e84fc425

lib/spack/spack/solver/asp.py
scheibelp: This is now done

lib/spack/spack/solver/asp.py
```diff
@@ -878,6 +878,7 @@ def key_fn(version):
             if (
                 isinstance(v.version, spack.version.GitVersion)
                 and v.version.user_supplied_reference
+                and v.origin != version_provenance.packages_yaml
```
scheibelp: Without this, I cannot set `version_provenance.packages_yaml` below like:

```
self.add_concrete_versions_from_specs(req_version_specs, version_provenance.packages_yaml)
```

While I think this change is technically correct, I wonder if ~the uses of `hash=numbered-version` in `packages.yaml`~ the automatic treatment of all `hash=numbered-version` (including from packages.yaml) as `user_supplied_reference` is confusing (it is to me).

More detail: if this block is entered then the check will fail, because the `hash=number` version has the same provenance as the `number` version introduced by the preference (so the check, which is looking for a `number` version defined in the preferences *but not equivalent*, ends up excluding everything that would satisfy it).

lib/spack/spack/solver/asp.py
```diff
@@ -1873,6 +1874,12 @@ def define_version_constraints(self):
             # generate facts for each package constraint and the version
             # that satisfies it
             for v in allowed_versions:
+                # TODO: with the below 3 lines, the latter check in
+                # test_requirement_adds_multiple_new_versions passes, but
+                # this doesn't seem like the right thing to do...
+                # if isinstance(v, spack.version.GitVersion) and v.user_supplied_reference:
+                #     x, _ = str(v).split("=")
+                #     self.gen.fact(fn.version_satisfies(pkg_name, x, v))
```
scheibelp: Right now the ASP facts generated for the spec `v@commits[1]` look like:

```
condition_requirement(0,"node_version_satisfies","v","commits[1]=2.2").
imposed_constraint(0,"node_version_satisfies","v","commits[1]=2.2").

%============================================================================
% Version Constraints
%============================================================================
version_satisfies("v","commits[1]","commits[1]").
version_satisfies("v","commits[1]=2.2","commits[1]=2.2").
version_satisfies("v","commits[1]=2.3","commits[1]=2.3").
```

Which looks strange to me. I wrote this in an attempt to generate:

```
version_satisfies("v","commits[1]","commits[1]=2.3").
```

It feels like these `hash=numbered-version` strings are being treated like atomic pieces of data though, which seems odd to me.

scheibelp:
I've addressed the direct review comments but haven't fully resolved my original questions. I've iterated on those at:

* https://github.com/spack/spack/pull/35057#discussion_r1109406779
* https://github.com/spack/spack/pull/35057#discussion_r1109410864

scheibelp:
I also added https://github.com/spack/spack/pull/35057/commits/fd58f2b6584b943d29062c25459dfc0202f10c4a to demonstrate that without these changes you also cannot do `require: @hash` (you can test the before/after with `git checkout develop -- lib/spack/spack/solver/asp.py` - although I'm trying to think of a better way to confirm the before/after of this in general).

lib/spack/spack/solver/asp.py
scheibelp: Partial update: I've broken down the versions as they are entered into the solver, such that you don't have atoms like "hash=number". The resulting challenge is restoring that information to the specs after building them from the solution. For now I've resolved that when `hash=number` is given on the command line, but if it exists in the `requires:` section, that would involve extra effort.

scheibelp:
@becker33 

I have now broken up the atomic hash=number versions in the solver. A consequence is that this gitversion info is no longer being propagated to the final spec by way of the solver/SpecBuilder (e.g. in some cases you could end up with `package@2.2` instead of `package@hash=2.2`), so I added logic in the SpecBuilder to reintroduce that (not sure if there's a better place for it: I can't recall if there are other places we reattach data which is discarded as part of the solve).

Those changes did not address the issues with assigning `version_provenance.packages_yaml` to the versions extracted from the requirements, so as discussed, I updated the `__eq__` method of `DeclaredVersion` (and added a comment on the considerations which determine equality in the context of `DeclaredVersion` vs. `Version`).

lib/spack/docs/build_settings.rst
```diff
@@ -398,6 +398,23 @@ under ``all`` are disregarded. For example, with a configuration like this:
...
+version requirement, then you need to define an equivalence to a numbered
+version as part of the requirement spec and also mention it as a version
+preference for the package:
+
+.. code-block:: yaml
+
+   packages:
+     openmpi:
+       require: "@full-commit-hash=4.1.5"
+       version: ["4.1.5"]
```
tgamblin: Maybe I'm misunderstanding something, but we don't currently require you to specify all this when you just do `spack install foo@abcd1234abcd1234abcd1234abcd1234`.  Why should requirements require more from the user?

lib/spack/spack/solver/asp.py
```diff
@@ -116,8 +116,32 @@ def getter(node):
     version_origin_fields,
 )(**{name: i for i, name in enumerate(version_origin_fields)})
 
-#: Named tuple to contain information on declared versions
-DeclaredVersion = collections.namedtuple("DeclaredVersion", ["version", "idx", "origin"])
+
+class DeclaredVersion(object):
+    def __init__(self, version, idx, origin):
+        if not isinstance(version, spack.version.VersionBase):
+            raise ValueError("Unexpected type for declared version: {0}".format(type(version)))
```
tgamblin, line 122: Use type annotations for this, since we check with `mypy` now.

```python
def __init__(self, version: VersionBase, idx: int, origin: str):
```

I'm on the fence on whether we also want to keep the dynamic check.  I could go either way -- I think it's useful if external code ever calls this, and also for routines where we don't currently check types.

lib/spack/spack/solver/asp.py
```diff
@@ -116,8 +116,32 @@ def getter(node):
...
+        self.idx = idx
+        self.origin = origin
+
+    def __eq__(self, other):
+        if not isinstance(other, DeclaredVersion):
+            return False
+        n_gitversions = sum(
+            1 for x in [self.version, other.version] if isinstance(x, spack.version.GitVersion)
+        )
+        if n_gitversions == 1:
```
tgamblin: I find this more intuitive and readable:

```suggestion
        is_git = lambda v: isinstance(v, spack.version.GitVersion)
        if is_git(self.version) != is_git(other.version):
```

lib/spack/spack/solver/asp.py
```diff
@@ -2070,6 +2111,9 @@ def setup(self, driver, specs, reuse=None):
         self.add_concrete_versions_from_specs(specs, version_provenance.spec)
         self.add_concrete_versions_from_specs(dev_specs, version_provenance.dev_spec)
```
tgamblin, line 2126: I'm reading through this and I still don't 100% understand why the transformations are necessary.

Is it not sufficient to add another invocation of `add_concrete_versions_from_specs` here, and to pass in a list of the specs from requirements?  I don't see why the other parts of this PR are necessary.

tgamblin (changes requested):

lib/spack/docs/build_settings.rst
scheibelp: Note: we *do* require the user to specify a `version:` if 

* you ask for `spack install foo@ abcd1234abcd1234abcd1234abcd1234=1.0` (i.e. a `=numbered-version` at the end)
* and `1.0` isn't mentioned as a `version()` in the `package.py`.

that could be improved, but I carried on that particular constraint here.

lib/spack/spack/solver/asp.py
scheibelp: Initially, that's what I had, and you can for example see an experimental commit I added which reverts the other stuff: https://github.com/spack/spack/pull/35057/commits/a99e5b273b11dd6f286000f8d27d8d47318cace7. If you check that out and run the tests, then the test `test_requirement_adds_multiple_new_versions` will fail at

```
s2 = Spec("v@{0}".format(commits[1])).concretized()
```

with 

```
    def handle_error(self, msg, *args):
        """Handle an error state derived by the solver."""
        msg = msg.format(*args)
    
        # For variant formatting, we sometimes have to construct specs
        # to format values properly. Find/replace all occurances of
        # Spec(...) with the string representation of the spec mentioned
        specs_to_construct = re.findall(r"Spec\(([^)]*)\)", msg)
        for spec_str in specs_to_construct:
            msg = msg.replace("Spec(%s)" % spec_str, str(spack.spec.Spec(spec_str)))
    
        # TODO: this raises early -- we should handle multiple errors if there are any.
>       raise UnsatisfiableSpecError(msg)
E       spack.solver.asp.UnsatisfiableSpecError: Cannot satisfy the requirements in packages.yaml for the 'v' package. You may want to delete them to proceed with concretization. To check where the requirements are defined run 'spack config blame packages'
```

why is that you might ask? It is because adding versions with `add_concrete_versions_from_specs` does not define the version equivalence between `@hash=number` and the `@hash`. Something needs to do that, or asking for `v@hash` will fail.

Although package requirements are added as imposed specs, they do not get added as version constraints in `define_version_constraints` because

```
(Pdb) x = Spec("v@{0}".format(commits[1]))
(Pdb) y = Spec("v@{0}=2.2".format(commits[1]))
(Pdb) z = Spec("v@2.2")
(Pdb) x.satisfies(y)
False
(Pdb) y.satisfies(x)
False
(Pdb) z.satisfies(y)
False
(Pdb) y.satisfies(z)
True
```

i.e. a hash would never be considered to satisfy a hash=version in the concretizer.

https://github.com/spack/spack/pull/35681 relates to this, although I would need to investigate it to check whether it's considered the case that it would allow `v@hash` to satisfy `v@hash=version`.

This problem would also "go away" if we forbid users from asking for `spec@hash` on the command line (`spec@version` would be OK). However,

* This doesn't seem strictly correct (how can the git commit by itself not satisfy?)
* The error message is quite vague, so anyone responding to someone's bug report would have to be familiar with this idiosyncrasy or risk spending an unpleasant and lengthy span of time to learn this

alalazo: > https://github.com/spack/spack/pull/35681 relates to this, although I would need to investigate it to check whether it's considered the case that it would allow v@hash to satisfy v@hash=version

In that PR I didn't fix the semantic of this case, or at least I didn't yet. To be consistent with the subset semantic for `satisfies` I think it should work the other way around:

1. `v@hash=version` $\subseteq$ `v@hash`
2. `v@hash=version` $\subseteq$ `v@version`
3. `v@hash` $\cap$ `v@version` = `v@hash=version`

where $\subseteq$ is `satisfies` and $\cap$ is `intersects`.

tgamblin: @alalazo: can you clear up (3) here?  Is that supposed to be `constrain()`?   `intersects()` returns `bool`.

alalazo: ~Yes, it's supposed to be `constrain`.~ To be crystal clear `intersects` is returning `False` if:

`v@hash` $\cap$ `v@version` = $\emptyset$

and `True` otherwise, so for these two:
```
Spec("v@version").intersects("v@hash") is True
Spec("v@version").constrain("v@hash") == Spec("v@hash=version")
```
Still to be implemented though. If people agree I can do it as a followup of #35681

lib/spack/docs/build_settings.rst
tgamblin: Ok I would argue that this is one of those places where we should leave things better than we found them, rather than carrying on the nastiness.  Really, for *any* user-specified spot (prefs, CLI, etc.) we should be adding the user's version to the set of possible solutions.  That's the spirit of the two calls to `add_concrete_versions_from_specs()`.

lib/spack/spack/solver/asp.py
tgamblin: @scheibelp thanks for explaining that in detail -- the example helps a lot.  Looking at this, I think:
1. We want to allow the user to specify `@commit` (with no version) either on the CLI or in any YAML file (they're going to do that)
2. We want them to be able to specify `@commit=ver` in any of those places, as well.

Basically we shouldn't restrict what they can write -- if it parses as a version, they should be able to write it without prickliness, which is bad UX.

I also think the root cause of the issue is what you showed -- `satisfies()` isn't doing the right thing, and we should fix the issue there.

Looking at what @alalazo wrote, I *think* his semantics are the right ones.  I would tweak this PR so that:

1. `v@hash=version` $\subseteq$ `v@hash`
2. `v@hash` $\subsetneq$ `v@hash=version`
3. `v@hash=version` $\subseteq$ `v@version`
4. `v@version` $\subsetneq$ `v@hash=version`
5. `v@version` $\cap$ `v@hash` = `v@hash=version`
6. `v@v1` $\cap$ `v@hash=v2` = `v@hash=(` `v1` $\cap$ `v2` `)`

etc.

I think we have a minor issue with this one right now:

2. `v@hash` $\subsetneq$ `v@hash=version`

in that we can currently have concrete specs that have a hash but not a version.  So if you do `spack find @commit=version`, and `foo@commit` is installed, we won't find it.  IMO the solution there is to eventually ensure that concretization assigns a version to every commit, and we never get an installed package with just a commit.  But we can change that part later.  Here, we just want to get the support for commit requirements working.

@alalazo I don't think this breaks anything -- it strictly widens the set of things that can be satisfied, right?  I don't think I'm missing something that needs to be preserved with commit `satisfies()`, right?

scheibelp: @tgamblin to make sure, are you saying that,

> `v@hash` $\subsetneq$ `v@hash=version`

*should* be the case, and by that do you mean that `v@hash` should not satisfy `v@hash=version`? If so, then `test_requirement_adds_multiple_new_versions` will continue to fail (at least, so long as we depend on `define_version_constraints` to address this), because it would never add a `.version_satisfies` fact to indicate that `v@hash` would be sufficient.

> IMO the solution there is to eventually ensure that concretization assigns a version to every commit,

how does that work? Should we make up a version? e.g. if the package declares:

```
version(2.0, commit=c2)
version(1.0, commit=c1)
```

and you pick some commit x such that c1 < x < c2, should it generate a number like "1.0.revision_number"? That works ok for a single branch but I'm not sure if that would be guaranteed to work for multiple branches.

scheibelp: >  If so, then `test_requirement_adds_multiple_new_versions` will continue to fail

Actually, I'm wrong about this (another piece of logic was interfering with this process and I'll be pushing commits shortly to highly that), although I am actually not sure how this works entirely at the moment.

scheibelp: I am confused because if it is not the case that `v@hash` satisfies `v@hash=version`, then we have for instance (i.e. as of the current latest commit in this branch) the following `spack solve --show asp` output in `test_requirement_adds_multiple_new_versions` for the spec `v@commits[1]`:


```
version_satisfies("v","ef73aeddeee888a7d5e2d9cc646ec933661f3920","ef73aeddeee888a7d5e2d9cc646ec933661f3920").
version_satisfies("v","ef73aeddeee888a7d5e2d9cc646ec933661f3920","ef73aeddeee888a7d5e2d9cc646ec933661f3920=2.3").
version_satisfies("v","ef73aeddeee888a7d5e2d9cc646ec933661f3920=2.3","ef73aeddeee888a7d5e2d9cc646ec933661f3920=2.3").
version_satisfies("v","fca011772147dbd7c7e1014d88c3baf059a28aa6=2.2","fca011772147dbd7c7e1014d88c3baf059a28aa6=2.2").
```

In particular, we do not have 

```
version_satisfies("v","ef73aeddeee888a7d5e2d9cc646ec933661f3920=2.3","ef73aeddeee888a7d5e2d9cc646ec933661f3920").
```

i.e. that choosing the hash satisfies the constraint that the version is "hash=number", in that case I don't know how the concretizer is producing a solution. I thought it might have something to do with the declaration of `version_equivalent` in `asp.py`, but if I comment that out, `test_requirement_adds_multiple_new_versions` still passes.

scheibelp: Oh mystery solved: the resulting spec in that test is

`v@11455396ff5e38a39364436654242f417f114414=2.3%gcc@10.2.1 build_system=generic arch=test-debian6-core2`

i.e. the input spec is `v@hash`, and the output is `v@hash=number`, so in that sense the concretizer "overrides" the user's request with the information from the package requirements.

lib/spack/spack/solver/asp.py
scheibelp: Now added (but dynamic check retained).

lib/spack/spack/version.py
```diff
@@ -603,7 +606,18 @@ def satisfies(self, other):
...
+                return False
+            elif self.user_supplied_reference and other.user_supplied_reference:
+                return self.ref_version == other.ref_version
+            elif other.user_supplied_reference:
+                return False
+            else:
+                # In this case, neither supplies a version equivalence with "="
+                # and the commit strings are equal
+                return True
+        elif other.is_ref:
```
scheibelp, line 646: @alalazo @tgamblin I think this implements the desired semantics and I added `test_git_user_supplied_reference_satisfaction` to confirm that (I didn't see a test covering these in `test/versions.py`). I don't think this will conflict with  https://github.com/spack/spack/pull/35681 (that doesn't change `satisfies`, and  furthermore `satisfies` can be more strict than `intersects`).

lib/spack/docs/build_settings.rst
scheibelp: This is now addressed and the associated tests are updated (plus a preexisting test that was expecting a failure has been updated).

tgamblin (approved):

tgamblin:
Pipeline here is only failing because a power runner could not be found.  Everything else is passing, and `version_satisfies()` is *clearly* implementing `satisfies()`, so I am happy with this. Thanks @scheibelp!

lib/spack/spack/solver/asp.py
```diff
@@ -115,8 +115,30 @@ def getter(node):
...
+
+class DeclaredVersion(object):
+    def __init__(self, version: spack.version.VersionBase, idx, origin):
+        if not isinstance(version, spack.version.VersionBase):
+            raise ValueError("Unexpected type for declared version: {0}".format(type(version)))
+        self.version = version
+        self.idx = idx
+        self.origin = origin
+
+    def __eq__(self, other):
```
haampie, line 127: Do we really need all this complexity? Is this still necessary when `@y` gets parsed as `VersionRange(y, y)` and `@=y` as `Version(y)`. imho this looks very messy

lib/spack/spack/solver/asp.py
```diff
@@ -115,8 +115,30 @@ def getter(node):
...
+        if not isinstance(other, DeclaredVersion):
+            return False
+        is_git = lambda v: isinstance(v, spack.version.GitVersion)
+        if is_git(self.version) != is_git(other.version):
+            # GitVersion(hash=x) and Version(y) compare equal if x == y, but
```
haampie, line 132: Why do they compare equal in the first place if they're different things.

alalazo:
I think this broke `develop` somehow, as it was tested sometimes ago :sweat:  https://github.com/spack/spack/actions/runs/4498754677/jobs/7915852590

PR #35058, opened 2023-01-21
Add package `py-nvidia-dali`

thomas-bouvier:
Hello, I'm not sure about some of the decisions I've made in this PR, especially:

- The name `py-nvidia-dali`
- The condition on the architecture `platform.machine()`
- The `depends_on` instructions

Tested with `cuda@11.8`.

https://github.com/NVIDIA/DALI

### DIFF
--- a/var/spack/repos/builtin/packages/py-nvidia-dali/package.py
+++ b/var/spack/repos/builtin/packages/py-nvidia-dali/package.py
@@ -49,5 +49,8 @@ class PyNvidiaDali(PythonPackage):
             url="https://developer.download.nvidia.com/compute/redist/nvidia-dali-cuda110/nvidia_dali_cuda110-1.22.0-6988993-py3-none-manylinux2014_aarch64.whl",
         )
 
+    depends_on("python@3.6:3.10", type=("build", "run"))
     depends_on("cuda@12", when="@1.22.0-cuda120", type=("build", "run"))
     depends_on("cuda@11", when="@1.22.0-cuda110", type=("build", "run"))
+    depends_on("py-astunparse@1.6.0:", type=("build", "run"))
+    depends_on("py-gast@0.2.1:0.4.0", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-nvidia-dali/package.py
+++ b/var/spack/repos/builtin/packages/py-nvidia-dali/package.py
@@ -1,0 +1,52 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import platform
+
+from spack.package import *
+
+
+class PyNvidiaDali(PythonPackage):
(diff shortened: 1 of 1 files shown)


tldahlgren (commented):
Confirmed the `x86_64` sha256.

tldahlgren:

var/spack/repos/builtin/packages/py-nvidia-dali/package.py
```diff
@@ -0,0 +1,52 @@
...
+    """A GPU-accelerated library containing highly optimized building blocks and
+    an execution engine for data processing to accelerate deep learning training
+    and inference applications. """
+
+    homepage = "https://developer.nvidia.com/dali"
+    url = "https://developer.download.nvidia.com/compute/redist/"
+
+    maintainers = ["thomas-bouvier"]
+
+    arch = platform.machine()
```
adamjstewart, line 22: We should also add a condition so that these are only applied on linux, not macos/windows

var/spack/repos/builtin/packages/py-nvidia-dali/package.py
```diff
@@ -0,0 +1,52 @@
...
+        version(
+            "1.22.0-cuda110",
+            sha256="0da47629fec01abf418fda0eeb393998820e40f6fae6b4c7d3e625aa4cdba6bd",
+            expand=False,
+            preferred=True,
+            url="https://developer.download.nvidia.com/compute/redist/nvidia-dali-cuda110/nvidia_dali_cuda110-1.22.0-6988993-py3-none-manylinux2014_aarch64.whl",
+        )
+
+    depends_on("cuda@12", when="@1.22.0-cuda120", type=("build", "run"))
+    depends_on("cuda@11", when="@1.22.0-cuda110", type=("build", "run"))
```
adamjstewart, line 54: Should we use `cuda@12.0` and `@11.0` or are these compatible with all minor releases?

thomas-bouvier: Yes, all minor releases are compatible. From the [docs](https://docs.nvidia.com/deeplearning/dali/main-user-guide/docs/installation.html#nvidia-dali) "CUDA 11.0 and CUDA 12.0 build uses CUDA toolkit enhanced compatibility. It is built with the latest CUDA 11.x/12.x respectively toolkit while it can run on the latest, stable CUDA 11.0 and CUDA 12.0 capable drivers (450.80 or later and 525.60 or later respectively). Using the latest driver may enable additional functionality."

var/spack/repos/builtin/packages/py-nvidia-dali/package.py
```diff
@@ -0,0 +1,53 @@
...
```
adamjstewart, line 54: Missing dependencies:
```
Requires-Python: >=3.6, <3.11
Requires-Dist: astunparse (>=1.6.0)
Requires-Dist: gast (<=0.4.0,>=0.2.1)
```

adamjstewart (changes requested):

adamjstewart (approved):

PR #35061, opened 2023-01-21
binutils: 2.30, 2.39, 2.40

haampie:


mwkrentel:
@haampie You could also add 2.39 and 2.40.

alalazo (dismissed):

alalazo (approved):

mwkrentel:
@haampie Sorry, I didn't realize 2.39 or 2.40 would cause problems like this.  :-(
So far, I don't see that the reason it's failing should be any different from earlier versions.

alalazo:
Superseded by #35128

PR #35065, opened 2023-01-21
`arrow`: don't use `when` as a keyword argument to filter_file()

thomas-bouvier:
Building arrow is broken because of changes introduced in https://github.com/spack/spack/commit/9d00e7d15d5da43cd098f695f15d0ee4d1aeaedc


thomas-bouvier:
@spackbot fix style

alalazo (approved):

PR #35066, opened 2023-01-21
axom: need to specify +rocm as part of dependency constraints

eugeneswalker:
Not being explicit about `+rocm` on the dependency yields an error during concretization:
* Using `spack@develop` (f593309b4e1c4590b757d7190dc14450241efd51 from `Sat Jan 21 10:17:53 2023 -0500`)
```
$> spack spec axom +rocm amdgpu_target=gfx90a
==> Error: Cannot set variant 'amdgpu_target' for package 'raja' because the variant condition cannot be satisfied for the given spec
```

Although this same error doesn't happen with `axom +cuda` I thought it might be good to be explicit there too?

@white238 @wspear

### DIFF
--- a/var/spack/repos/builtin/packages/axom/package.py
+++ b/var/spack/repos/builtin/packages/axom/package.py
@@ -117,7 +117,6 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
         depends_on("umpire@6.0.0", when="@0.6.0")
         depends_on("umpire@5:5.0.1", when="@:0.5.0")
         depends_on("umpire +openmp", when="+openmp")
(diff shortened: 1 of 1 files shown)


white238 (dismissed):
Thanks @eugeneswalker .

Could you remove line 120 and 128? They should be redundant at this point.

eugeneswalker:
> Could you remove line 120 and 128? They should be redundant at this point.

Done! Thanks!

white238 (dismissed):

eugeneswalker:
Had to fix style issue. Sorry!

white238 (approved):

PR #35067, opened 2023-01-21
freeimage: fails to compile with c++17, use c++14

wdconinc:
Only `opencascade` when a (non-default) variant depends on `freeimage`, which seems to have gone unmaintained. There are c++17 standard violations [[1]]( https://en.cppreference.com/w/cpp/language/except_spec) in the code, so we can at most expect c++14. Since some compilers default to c++17 (gcc-12) we need to be explicit.

### DIFF
--- a/var/spack/repos/builtin/packages/freeimage/install_fixes.patch
+++ b/var/spack/repos/builtin/packages/freeimage/install_fixes.patch
@@ -4,10 +4,10 @@
 
  # General configuration variables:
  DESTDIR ?= /
--INCDIR ?= $(DESTDIR)/include
--INSTALLDIR ?= $(DESTDIR)/lib
-+INCDIR ?= $(DESTDIR)/usr/include
-+INSTALLDIR ?= $(DESTDIR)/usr/lib
+-INCDIR ?= $(DESTDIR)/usr/include
+-INSTALLDIR ?= $(DESTDIR)/usr/lib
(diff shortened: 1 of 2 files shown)


wdconinc:
@matz-e maybe you are still using this since you wrote the original package.

wdconinc:
Questions to reviewers:
- a way to require c++14 only when the compiler actually defaults to a newer standard?
- is `-std` a sufficiently standard (no pun intended) flag to stick in CXXFLAGS?

matz-e:
> @matz-e maybe you are still using this since you wrote the original package.

Thanks for the ping, we stopped using this package luckily. I actually got our users to switch to other libraries, since it looked like upstream abandoned `freeimage`.

tldahlgren:
> Questions to reviewers:
> 
> * a way to require c++14 only when the compiler actually defaults to a newer standard?
> * is `-std` a sufficiently standard (no pun intended) flag to stick in CXXFLAGS?

Probably not the best person to answer this but there are a few packages that explicitly set the standard for flags (i.e., `axom`, `gridlab-d`, `raja`, and `strumpack`).  Using `conflicts()` based on compiler versions is another option.

tldahlgren (approved):

tldahlgren:
Let me know if you want to make changes or keep as-is.

wdconinc:
I'll just keep this as is. This package likely does not have as large of an install base.

PR #35077, opened 2023-01-23
Revert "Fix for padding relocation (issue 28170)"

haampie:
Reverts spack/spack#33122

The linked PR was not properly reviewed. The result is that the effective prefix is one byte shorter than required by the config setting, so that relocation may fail when moving to a location of exactly `padded_length` bytes.

The latter is a very typical use-case, for example when `config:install_tree:padded_length` is globally configured to say `128`, but the `config:install_tree:root` has variable length in different environments.

### DIFF
--- a/lib/spack/spack/store.py
+++ b/lib/spack/spack/store.py
@@ -119,7 +119,6 @@ def parse_install_tree(config_dict):
             msg = "Cannot pad %s to %s characters." % (root, padded_length)
             msg += " It is already %s characters long" % len(root)
             tty.warn(msg)
-        root = root.rstrip(os.path.sep)
     else:
         root = unpadded_root
 


vanderwb:
@haampie - I'm sorry to hear that it caused issues. Can you describe what the failure looks like?

Just reverting will still be reintroducing broken behavior that makes relocation functionally dependent on path length. I believe you had suggested replace the slash with another character. This will cause us problems with our current deployment but it's better than just reverting - we can reinstall our stacks.

It sounds like there needs to be a bit more testing around the relocation mechanism in any case.

haampie:
Someone else reported the following issue, where relocation to a new path with the same padded length is causing failures. For example:

```
01234567890123456789012345
/x/__spack_placeholder__//
/xyz/__spack_placeholder_/
```

here the padded length is 24, but packages in /x have effectively used length 23 because the trailing / was stripped. Creating a buildcache from /x and installing to /xyz then fails.

I'm working on a better fix to also deal with multibyte characters and wide encodings like utf-16 on Windows

haampie:
The better fix is in #35116, but right now it's a breaking change

haampie:
@vanderwb can you actually still reproduce that issue you were solving in 33122? I could not

vanderwb:
@haampie - are you wanting me to test your newer fixes against my case or are you asking whether I can still reproduce it in the older commit?

haampie:
Sorry, I meant: can you reproduce the issue when the original fix is reverted, so on commit 7eb3f83feaf75505a044d4652676cca86e927d17

vanderwb:
Thanks for the clarification. I will test this ASAP, likely tomorrow.

vanderwb:
@haampie - I'm happy to report that my prior test case that resulted in a broken RPATH no longer does so. I'm glad things are resolved in a way that doesn't break anything else too. Thanks for following up on this.

(this was testing `develop`, since I could not check out the commit you linked in my up-to-date clone)

haampie:
Yeah, develop contains your fix... Can you try `develop` with `git revert c721aab0067e127cd9db6a4eb919747814549b4f`?

vanderwb:
Apologies. I misunderstood. In any case, with that commit reverted I still cannot replicate. Is that what you expected? (i.e. is there some other change already in place to resolve it?)

haampie:
Yeah, I found the bug report curious, I couldn't reproduce the issue myself, so i'm happy you can't either at this point.

There's been quite some changes to the buildcache/relocation code, and generally we don't use patchelf anymore for modifying rpaths. If we do `patchelf` it may be on a more recent version than at the time the issue was reported.

So, I'm leaning towards merging this and backporting it to 0.19.

We still need to make up our minds about the weird edge case, though (#35116 doesn't fix it yet)

alalazo (approved):

haampie:
Ah, this didn't end up in the 0.19 release, removed the milestone.

PR #35078, opened 2023-01-23
append cflags, cxxflags and fflags instead of prepend

haampie:
cflags/cxxflags/fflags are pretty much useless in Spack, with all
options of (1) setting the relevant env variable, (2) putting the
variable on the make command invocation, (3) injecting through the
compiler wrapper, because:

1. Env variables may be ignored by the build system (e.g. `CFLAGS = ...`
   not `CFLAGS ?= ...` in Makefiles).
2. Putting them on the command line is not composable, and may in fact
   override required defaults for badly written Makefiles; it's not
   uncommon to see Makefiles like these:
   ```Makefile
   CFLAGS = -std=c99 -D_FILE_OFFSET_BITS=64
   ```
3. The compiler wrapper *prepends* flags, which typically means they
   have no effect:
   ```
   # Our -O3 -g0 flags are overridden by -O2 -g
   gcc -O3 -g0 ... -O2 -g
   ```

The latter looks like a bug to me, so this PR ensures we *append* the
compiler flags.

It does *not* touch CPPFLAGS / LDFLAGS, since they may contain -I and
-L flags which should take precedence, so should be prepended. Maybe
Spack should just do a pass over CPPFLAGS/LDFLAGS in Python and move
-I and -L into other variables so that we can also append the other
preprocessor/linker flags... cause it'd be nice to ensure that ldflags
of the form `-fuse-ld=...` come last too.

### DIFF
--- a/lib/spack/env/cc
+++ b/lib/spack/env/cc
@@ -626,39 +626,10 @@ elif [ "$SPACK_ADD_DEBUG_FLAGS" = "custom" ]; then
     extend flags_list SPACK_DEBUG_FLAGS
 fi
 
-# Fortran flags come before CPPFLAGS
-case "$mode" in
-    cc|ccld)
-        case $lang_flags in
-            F)
-                extend flags_list SPACK_FFLAGS
-                ;;
-        esac
-        ;;
-esac
-
-# C preprocessor flags come before any C/CXX flags
+# C preprocessor flags
 case "$mode" in
     cpp|as|cc|ccld)
         extend flags_list SPACK_CPPFLAGS
-        ;;
-esac
-
-
-# Add C and C++ flags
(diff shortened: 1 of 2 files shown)


haampie:
It may lead to issues:

https://git.gnupg.org/cgi-bin/gitweb.cgi?p=libgcrypt.git;a=blob;f=random/jitterentropy-base.c;h=cf7630d982af58d6cd758dc65e71c9cb04669570;hb=HEAD#l58

This `#error`s with `-O1/2/3/` set. Great.

haampie:
Ok, that issue was fixed by setting the flags through env variables, so the general concept still works

haampie:
Superseded by another PR.

PR #35081, opened 2023-01-23
CPMD: revised configuration; libxc support; fft variants

g-mathias:
- rewritten `package.py` by @RemiLacroix-IDRIS  
- separate `configure/spack`
- `MakefilePackage` -> `AutotoolsPackage` 
- libxc support
- different fft variants supported
- `+openmp+mpi` as default variants
-  support for extra configuration flags (via CPP)
- build `manual.pdf`
- maintainer added
- website updated
- `develop` version added


var/spack/repos/builtin/packages/cpmd/package.py
```diff
@@ -3,73 +3,209 @@
...
+    extraflags = [
+        "ALPHALINUX",
+        "ALTIX",
+        "BG",
+        "CHECKS_IEEE_ARITHMETIC",
+        "CUDA",
+        "EREG",
+        "ES",
+        "FFT_HASNT_THREADED_COPIES",
+        "FFT_HAS_LOW_LEVEL_TIMERS",
```
haampie, line 49: I'm confused about this giant array. Isn't the build system supposed to set these?

g-mathias: Hi Harmen,
the list has been extracted with sed&awk and maps all preprocessor flags that can be used within cpmd. I created it for the sake of completeness. Most flags are probably outdated since they relate to compiler/library features, which have not been available on some older HPC systems (build +10y ago) that might even not be operational anymore. Some of the compiler wisdom is contained in the files of the "configure" directory, but has not been included in the the package.py file (yet).
One goal is to enable automatized builds with spack and check if the flags are still valid or needed and supply build recipes for current systems in the package.

Cheers,
Gerald

nboelte:
This package.py is very comprehensive and helped us a lot with getting CPMD installed with more recent compilers. It would be nice if it could be merged since it took me a bit to find it here!

The only issue we ran into is that 'svnversion' is required to compile so a depends_on('subversion') might be good, and that the prefix.doc in the install function ends up being a file instead of a directory.

nboelte:
The current compiler wisdom on our system looks like this now:

```
        if spec.satisfies("%gcc@10:"):
            cp.filter(r"FFLAGS='([^']*)'", "FFLAGS='\\1 -fallow-argument-mismatch'")

        if spec.satisfies("%gcc@14:") or spec.satisfies("%aocc"):
            cp.filter(r"CFLAGS='([^']*)'", "CFLAGS='\\1 -Wno-error=int-conversion'")
```

alalazo:
@nboelte If this PR is still of interest, it needs to be ported to the https://github.com/spack/spack-packages repository

g-mathias:
@nboelte Hi Niklas, I will withdraw this MR since it points to the frozen CPMD-code repo. If I manage in August I will try to set up a new one ready for  spack 1.0 and using the [open-cpmd](https://github.com/OpenCPMD) repo.

PR #35082, opened 2023-01-23
Add package `py-continuum`

thomas-bouvier:


thomas-bouvier:
@spackbot fix style

thomas-bouvier:
@spackbot run pipeline

tldahlgren (commented):
Confirmed sha256.

tldahlgren:
Added @adamjstewart since this is a python package.

var/spack/repos/builtin/packages/py-continuum/package.py
```diff
@@ -0,0 +1,36 @@
...
+    depends_on("py-torchvision@0.4.0:", type=("build", "run"))
+    depends_on("py-numpy@1.17.2:", type=("build", "run"))
+    depends_on("pil@6.2.1:", type=("build", "run"))
+    depends_on("py-matplotlib@3.1.0:", type=("build", "run"))
+    depends_on("py-scipy@1.3.3:", type=("build", "run"))
+    depends_on("py-scikit-image@0.15.0:", type=("build", "run"))
+    depends_on("py-scikit-learn@0.24.1:", type=("build", "run"))
+    depends_on("py-pandas@1.1.5:", type=("build", "run"))
+    depends_on("py-pytest@5.0.1:", type=("build", "run"))
+    depends_on("py-pytest-mock@3.6.1:", type=("build", "run"))
```
adamjstewart, line 37: Missing dep:
```python
depends_on("py-prospector@1.2:+with_mypy", type=("build", "run"))
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-continuum/package.py
thomas-bouvier: `prospector` is not called from any Python code in continuum (it's only called as a Makefile target). I opened a PR which has been merged here https://github.com/Continvvm/continuum/pull/280. Do you think it's worth adding this dep in Spack?

adamjstewart: The only problem with leaving it out is that `pip check` will complain that it's missing. Not really a big deal. Could also add your PR as a patch to the file. I'll leave the final decision to you, I'm fine with whatever since I don't personally use this package.

var/spack/repos/builtin/packages/py-continuum/package.py
```diff
@@ -16,6 +16,8 @@ class PyContinuum(PythonPackage):
 
     version("1.2.7", sha256="baadcdbe2f5b3c05254307c73434a012f2a3d663e1be9469a03d4b82559e98e1")
 
+    patch("move_prospector_dev_requires.patch")
```
adamjstewart: Can you add a comment with a link to the PR? Also, probably want to add:
```suggestion
    patch("move_prospector_dev_requires.patch", when="@:1.2.7")
```
I'm guessing this won't be needed for the next release.

adamjstewart (approved):

PR #35083, opened 2023-01-23
Add a `maintainers` directive

### LINKED ISSUE #34879: Add possibility to inherit maintainers from base package
### Summary

Add a function to `spack.package` to declare maintainers which constructs the `maintainers` attribute of the package.

### Rationale

Adding a Python package with `maintainers = [...]` will overwrite the maintainers declared in `PythonPackage`, to circumvent this the contributor and/or reviewer have to be aware of this issue and change the maintainers declaration to

````python
maintainers = [
    *PythonPackage.maintainers,
    "user1",
    "user2",
]
````

### Description

A solution could be a `maintainer` function to allow declaration of maintainers with

````python
maintainer("user1")
maintainer("user2")
````

The function will construct the `maintainers` attribute by appending the user handles instead of overwriting. Overwriting can still be done by directly writing to ``maintainers``.

### Additional information

See https://github.com/spack/spack/pull/34857#discussion_r1066698621

---

```
❯ spack --version
0.20.0.dev0 (2001be99b36bb18864886c9ae00f41dd55e9c800)
```

### General information

### PR DESCRIPTION
**Pull request by alalazo**
fixes #34879

This commit adds a new "maintainers" directive, which by default extend the list of maintainers for a given package. The directive is backward compatible with the current practice of having a `maintainers` list declared at the class level. In cases where we need to restart the list after e.g. inheriting from a base package, the reset=True keyword can be used.

### DIFF
--- a/lib/spack/docs/packaging_guide.rst
+++ b/lib/spack/docs/packaging_guide.rst
@@ -268,7 +268,7 @@ generates a boilerplate template for your package, and opens up the new
 
        # FIXME: Add a list of GitHub accounts to
        # notify when the package is updated.
-       # maintainers = ["github_user1", "github_user2"]
+       # maintainers("github_user1", "github_user2")
 
        version("6.2.1", sha256="eae9326beb4158c386e39a356818031bd28f3124cf915f8c5b1dc4c7a36b4d7c")
 
@@ -319,14 +319,8 @@ The rest of the tasks you need to do are as follows:
 
 #. Add a comma-separated list of maintainers.
 
-   The ``maintainers`` field is a list of GitHub accounts of people
-   who want to be notified any time the package is modified. When a
-   pull request is submitted that updates the package, these people
-   will be requested to review the PR. This is useful for developers
-   who maintain a Spack package for their own software, as well as
-   users who rely on a piece of software and want to ensure that the
-   package doesn't break. It also gives users a list of people to
-   contact for help when someone reports a build error with the package.
+   Add a list of Github accounts of people who want to be notified
+   any time the package is modified. See :ref:`package_maintainers`.
 
 #. Add ``depends_on()`` calls for the package's dependencies.
 
@@ -497,27 +491,30 @@ some examples:
 In general, you won't have to remember this naming convention because
 :ref:`cmd-spack-create` and :ref:`cmd-spack-edit` handle the details for you.
 
+.. _package_maintainers:
+
 -----------
 Maintainers
 -----------
 
-In Spack, packages can optionally have one or more maintainers. The ``maintainers``
-directive can be used to add them:
+Each package in Spack may have one or more maintainers, i.e. one or more
+GitHub accounts of people who want to be notified any time the package is
+modified.
+
+When a pull request is submitted that updates the package, these people will
+be requested to review the PR. This is useful for developers who maintain a
+Spack package for their own software, as well as users who rely on a piece of
+software and want to ensure that the package doesn't break. It also gives users
+a list of people to contact for help when someone reports a build error with
+the package.
+
+To add maintainers to a package, simply declare them with the ``maintainers`` directive:
 
 .. code-block:: python
 
    maintainers("user1", "user2")
 
-The list of maintainers is inherited from parent packages by default, but users can pass
-the ``reset=True`` argument if they want to restart this list from scratch:
-
-.. code-block:: python
-
-   class CustomFoo(Foo):
-      maintainers("user1", reset=True)
-
-In the snippet above, maintainers declared in the `Foo` package will not be part of the list
-for the ``CustomFoo`` package.
+The list of maintainers is additive, and includes all the accounts eventually declared in base classes.
 
 -----------------
 Trusted Downloads
--- a/lib/spack/docs/packaging_guide.rst
+++ b/lib/spack/docs/packaging_guide.rst
@@ -501,13 +501,12 @@ In general, you won't have to remember this naming convention because
 Maintainers
 -----------
 
-In Spack, packages can optionally have one or more maintainers. The ``maintainer``
+In Spack, packages can optionally have one or more maintainers. The ``maintainers``
 directive can be used to add them:
 
 .. code-block:: python
 
-   maintainer("user1")
-   maintainer("user2")
+   maintainers("user1", "user2")
 
 The list of maintainers is inherited from parent packages by default, but users can pass
 the ``reset=True`` argument if they want to restart this list from scratch:
@@ -515,7 +514,7 @@ the ``reset=True`` argument if they want to restart this list from scratch:
 .. code-block:: python
 
    class CustomFoo(Foo):
-      maintainer("user1", reset=True)
+      maintainers("user1", reset=True)
 
 In the snippet above, maintainers declared in the `Foo` package will not be part of the list
 for the ``CustomFoo`` package.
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by adamjstewart** [comment]
Can we call it maintainers (plural) and have it accept a list of *args? I know this probably introduces a name clash, but we can handle things by type if we need to. This would also be relatively easy to search and replace if we want to deprecate the old behavior.

**Comment by alalazo** [comment]
I didn't use `maintainers` exactly because of the name clash. I also thought that if we want to add specific attributes later on (e.g. a description of which part of the package is maintained by whom in large packages), this will already give a declaration separated by users.

**Comment by alalazo** [comment]
Also, the "old" way of storing maintainers is the implementation behind the directive.

**Comment by adamjstewart** [comment]
But we can always change the backend. For packages like `aws-parallelcluster` with 8+ maintainers, not supporting lists is a lot of extra code to type.

**Comment by alalazo** [comment]
@spackbot run pipeline

File: lib/spack/docs/packaging_guide.rst [hunk 1]
```diff
@@ -497,6 +497,28 @@ some examples:
 In general, you won't have to remember this naming convention because
```
adamjstewart, line 491: There are a few other places this needs to be updated:
```console
$ grep -R maintainers *
contribution_guide.rst:dependencies. Currently, the maintainers manage these branches by
developer_guide.rst:project maintainers, as the tasks described here require maintainer
features.rst:       # maintainers = ["github_user1", "github_user2"]
packaging_guide.rst:       # maintainers = ["github_user1", "github_user2"]
packaging_guide.rst:#. Add a comma-separated list of maintainers.
packaging_guide.rst:   The ``maintainers`` field is a list of GitHub accounts of people
packaging_guide.rst:asked to help maintain this version of the package if the current maintainers
packaging_guide.rst:developers and maintainers provide additional stand-alone tests
```
I believe the SC tutorial also needs to be updated, but that can be done later. We can still support the comma-separated list, but I would be in favor of not documenting it and only documenting the directive. The directive is clearly superior.

lib/spack/docs/packaging_guide.rst
```diff
@@ -497,6 +497,28 @@ some examples:
...
 :ref:`cmd-spack-create` and :ref:`cmd-spack-edit` handle the details for you.
 
+-----------
+Maintainers
+-----------
+
+In Spack, packages can optionally have one or more maintainers. The ``maintainers``
```
adamjstewart: Should probably clarify what a maintainer is and is not. We do this already in the packaging guide but maybe we should copy the text here too.

lib/spack/docs/packaging_guide.rst
```diff
@@ -497,6 +497,28 @@ some examples:
...
+directive can be used to add them:
+
+.. code-block:: python
+
+   maintainers("user1", "user2")
+
+The list of maintainers is inherited from parent packages by default, but users can pass
```
adamjstewart:
```suggestion
The list of maintainers is inherited from base class by default, but users can pass
```
I think this would be less confusing and is true 99% of the time

lib/spack/docs/packaging_guide.rst
```diff
@@ -497,6 +497,28 @@ some examples:
...
+
+   class CustomFoo(Foo):
+      maintainers("user1", reset=True)
+
+In the snippet above, maintainers declared in the `Foo` package will not be part of the list
+for the ``CustomFoo`` package.
+
 -----------------
 Trusted Downloads
 -----------------
```
adamjstewart, line 521: Can we also update `spack create` to use the new template?

adamjstewart (commented):
Looks great! Tested with a few packages and works as advertised!

tgamblin (changes requested):
> In cases where we need to restart the list after e.g. inheriting from a base package, the `reset=True` keyword can be used.

Can we make this more in line with #35045?  Once that is in we'll have a general `drop()` directive -- and `reset=True` will be redundant.  So basically you'd do:

```python
    drop("maintainers")
    maintainers("tgamblin", "adamjstewart", ...)
```

I think for long lists this is actually clearer than sticking `reset=True` at the end.

alalazo:
@tgamblin :+1: on the combination of the two directives.

One note: `maintainers` is a list, not a dictionary. I am not sure we can hook that into the directive meta class without breaking backward compatibility with the current practice of setting a class level variable. In the worst case we can treat "maintainers" differently within "drop", even though we have enough special cases in the source code already.

lib/spack/docs/packaging_guide.rst
alalazo: There were 2 other mentions of the list. They've been updated now.

var/spack/repos/builtin.mock/packages/py-extension1/package.py
```diff
@@ -13,8 +13,7 @@ class PyExtension1(PythonPackage):
     homepage = "http://www.example.com"
     url = "http://www.example.com/extension1-1.0.tar.gz"
 
-    # Override settings in base class
-    maintainers = []
+    maintainers = ["user1", "user2"]
```
alalazo, line 16: When "drop" is ready, we just need to substitute this with:
```suggestion
    drop("maintainers")
    maintainers("user1", "user2")
```

lib/spack/docs/packaging_guide.rst
alalazo: Moved the description of what a maintainer is here, and reference that description in "Create a New Package"

lib/spack/docs/packaging_guide.rst
alalazo: Done

lib/spack/docs/packaging_guide.rst
alalazo: Done

tgamblin (commented):
LGTM!

tgamblin (approved):

lib/spack/docs/packaging_guide.rst
```diff
@@ -497,6 +491,31 @@ some examples:
...
+a list of people to contact for help when someone reports a build error with
+the package.
+
+To add maintainers to a package, simply declare them with the ``maintainers`` directive:
+
+.. code-block:: python
+
+   maintainers("user1", "user2")
+
+The list of maintainers is additive, and includes all the accounts eventually declared in base classes.
```
adamjstewart, line 517:
```suggestion
The list of maintainers is additive, and includes all the accounts declared in base classes.
```
Or previously

adamjstewart (commented):
Any objections to replacing all current usages of `maintainers = [...]` in packages with the new syntax? Can be done in a follow-up PR if you want to keep these PR lightweight. I at least want to do this for all Python packages, but don't see any reason not to do it for all packages.

lib/spack/docs/packaging_guide.rst
alalazo: > Any objections to replacing all current usages of maintainers = [...] in packages with the new syntax?

No objections, but I'd do it in a follow-up PR.

alalazo:
@spackbot run pipeline

lib/spack/docs/packaging_guide.rst
adamjstewart: Should we add this doc fix though?

alalazo:
@spackbot run pipeline

alalazo:
Restarted a second time since the errors seem to be in the backend:

Screenshot from 2023-01-27 07-23-27

alalazo:
@adamjstewart CI is passing, so I'll merge this PR and follow with one including:
- doc fixup
- package changes

PR #35089, opened 2023-01-23
py-planetary-computer: add new package

adamjstewart:


var/spack/repos/builtin/packages/py-planetary-computer/package.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```
tldahlgren, line 1: Unless I missed something, shouldn't this end `2023` now?

var/spack/repos/builtin/packages/py-pystac-client/package.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```
tldahlgren, line 1: Same about year

tldahlgren (commented):
Confirmed the two sha256.  Do have a question about the copyright year since that change went in 4 days ago.

var/spack/repos/builtin/packages/py-planetary-computer/package.py
adamjstewart: I stopped updating Spack since Spack stopped running my pipelines for 24 hrs after the PR was submitted if I update first. Is this something that needs to be changed manually? If the tests don't fail, does it matter what the copyright year says?

tldahlgren (approved):

PR #35094, opened 2023-01-23
Extract functions to read spec files from different format

alalazo:
Extracted from #34821

This commit makes explicit the format version of the spec file we are reading from.

Before there were different functions capable of reading some part of the spec file at multiple format versions. The decision was implicit, since checks were based on the structure of the JSON without ever checking a format version number.

The refactor makes also explicit which spec file format is used by which database and lockfile format, since the information is stored in global mappings.

To ensure we don't change the hash of old specs, JSON representations of specs have been added as data. A unit tests checks that we read the correct hash in, and that the hash stays the same when we re-serialize the spec using the most recent format version.


lib/spack/spack/spec.py
```diff
@@ -2063,137 +2068,13 @@ def to_json(self, stream=None, hash=ht.dag_hash):
...
-                    spec.compiler_flags[name] = []
-                    for val in values:
-                        spec.compiler_flags.add_flag(name, val, False)
-                else:
-                    spec.variants[name] = vt.MultiValuedVariant.from_node_dict(name, values)
-        elif "variants" in node:
-            for name, value in node["variants"].items():
-                spec.variants[name] = vt.MultiValuedVariant.from_node_dict(name, value)
-            for name in FlagMap.valid_compiler_flags():
-                spec.compiler_flags[name] = []
```
alalazo, line 2112: For reviewers: while looking at this, I noted that what we call specfile V1 is indeed a couple of different, ancient, formats. I removed these 5 lines after checking that they were preceding Spack v0.13 (the first version of Spack supporting a JSON serialization for concrete specs). I can put it back if need be.

lib/spack/spack/test/spec_yaml.py
```diff
@@ -507,3 +508,33 @@ def test_legacy_yaml(tmpdir, install_mockery, mock_packages):
...
+    "specfile,expected_hash,reader_cls",
+    [
+        # First version supporting JSON format for specs
+        ("specfiles/hdf5.v013.json", "vglgw4reavn65vx5d4dlqn6rjywnq76d", spack.spec.SpecfileV1),
+        # Introduces full hash in the format, still has 3 hashes
+        ("specfiles/hdf5.v016.json", "stp45yvzte43xdauknaj3auxlxb4xvzs", spack.spec.SpecfileV1),
+        # Introduces "build_specs", see https://github.com/spack/spack/pull/22845
+        ("specfiles/hdf5.v017.json", "xqh5iyjjtrp2jw632cchacn3l7vqzf3m", spack.spec.SpecfileV2),
+        # Use "full hash" everywhere, see https://github.com/spack/spack/pull/28504
+        ("specfiles/hdf5.v019.json", "iulacrbz7o5v5sbj7njbkyank3juh6d3", spack.spec.SpecfileV3),
```
alalazo: The huge numbers of lines changed is mostly due to the data added for this test. The test reads different spec serializations, and check that if they are round-tripped to JSON using the most recent format the hash doesn't change.

haampie: I've pushed a commit so the data it minified and gzipped

alalazo: :+1: as we agreed in DM

lib/spack/spack/environment/environment.py
```diff
@@ -1435,7 +1444,7 @@ def _concretize_separately(self, tests=False):
                         if test_dependency in current_spec[node.name]:
                             continue
                         current_spec[node.name].add_dependency_edge(
-                            test_dependency.copy(), deptype="test"
+                            test_dependency.copy(), deptypes="test"
```
haampie, line 1447: :thinking: was this never an issue?

lib/spack/spack/spec.py
```diff
@@ -1550,34 +1550,39 @@ def _add_dependency(self, spec, deptypes):
                 "Cannot depend on incompatible specs '%s' and '%s'" % (dspec.spec, spec)
             )
 
-    def add_dependency_edge(self, dependency_spec, deptype):
```
alalazo, line 1553: No, I just changed this inconsistency

lib/spack/spack/environment/environment.py
```diff
@@ -1974,10 +1983,11 @@ def _read_lockfile_dict(self, d):
 
         # Track specs by their DAG hash, allows handling DAG hash collisions
         first_seen = {}
-
+        current_lockfile_format = d["_meta"]["lockfile-version"]
+        reader = READER_CLS[current_lockfile_format]
```
haampie: Should this give a nice error instead of KeyError? "This version of Spack does not support environments version ..."

alalazo: Done in 0d752d8

haampie (approved):

PR #35095, opened 2023-01-23
Windows: Port XZ package to support Windows

johnwparent:
Port of XZ to Windows and some related changes to MSVC compiler class. Part of #34938 

Directly depends on #35098 so that PR should go in first.

### DIFF
--- a/var/spack/repos/builtin/packages/xz/package.py
+++ b/var/spack/repos/builtin/packages/xz/package.py
@@ -97,10 +97,6 @@ class MSBuildBuilder(MSBuildBuilder):
         newest_compiler = max(compiler_dirs, key=get_file_string_number)
         return os.path.join(win_dir, newest_compiler)
 
-    @property
-    def toolchain_version(self):
-        return "v" + self.pkg.compiler.platform_toolset_ver
-
     def is_64bit(self):
         return platform.machine().endswith("64")
 
--- a/var/spack/repos/builtin/packages/xz/package.py
+++ b/var/spack/repos/builtin/packages/xz/package.py
@@ -7,7 +7,6 @@ import glob
 import os
 import platform
 import re
-import sys
 
 from spack.build_systems.autotools import AutotoolsBuilder
 from spack.build_systems.msbuild import MSBuildBuilder
@@ -133,7 +132,7 @@ class MSBuildBuilder(MSBuildBuilder):
                 libs_to_install = glob.glob(
                     os.path.join(self.build_directory, "**", lib), recursive=True
                 )
-                for l in libs_to_install:
-                    install(l, prefix.lib)
+                for library in libs_to_install:
+                    install(library, prefix.lib)
         with working_dir(pkg.stage.source_path):
             install_tree(os.path.join("src", "liblzma", "api"), prefix.include)
--- a/var/spack/repos/builtin/packages/xz/package.py
+++ b/var/spack/repos/builtin/packages/xz/package.py
@@ -5,7 +5,6 @@
 
 import glob
 import os
-import platform
 import re
 
 from spack.build_systems.autotools import AutotoolsBuilder
--- a/var/spack/repos/builtin/packages/xz/package.py
+++ b/var/spack/repos/builtin/packages/xz/package.py
@@ -90,6 +90,7 @@ class MSBuildBuilder(MSBuildBuilder):
         def get_file_string_number(f):
             s = re.findall(r"\d+$", f)
             return (int(s[0]) if s else -1, f)
+
         win_dir = os.path.join(super().build_directory, "windows")
         compiler_dirs = []
         with working_dir(win_dir):
@@ -109,9 +110,9 @@ class MSBuildBuilder(MSBuildBuilder):
     def msbuild_args(self):
         plat = "x64" if self.is_64bit() else "x86"
         if self.pkg.spec.satisfies("libs=shared,static"):
-            f =  "xz_win.sln"
+            f = "xz_win.sln"
         elif self.pkg.spec.satisfies("libs=shared"):
-            f =  "liblzma_dll.vcxproj"
+            f = "liblzma_dll.vcxproj"
         else:
             f = "liblzma.vcxproj"
         return [self.define("Configuration", "Release"), self.define("Platform", plat), f]
@@ -129,7 +130,9 @@ class MSBuildBuilder(MSBuildBuilder):
             else:
                 libs_to_find.append("*.lib")
             for lib in libs_to_find:
-                libs_to_install = glob.glob(os.path.join(self.build_directory, "**", lib), recursive=True)
+                libs_to_install = glob.glob(
+                    os.path.join(self.build_directory, "**", lib), recursive=True
+                )
                 for l in libs_to_install:
                     install(l, prefix.lib)
         with working_dir(pkg.stage.source_path):
--- a/var/spack/repos/builtin/packages/xz/package.py
+++ b/var/spack/repos/builtin/packages/xz/package.py
@@ -98,7 +98,7 @@ class MSBuildBuilder(MSBuildBuilder):
         return os.path.join(win_dir, newest_compiler)
 
     def is_64bit(self):
-        return platform.machine().endswith("64")
+        return "64" in self.pkg.spec.target.family
 
     def msbuild_args(self):
         plat = "x64" if self.is_64bit() else "x86"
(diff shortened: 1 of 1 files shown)


johnwparent:
@spackbot run pipeline

var/spack/repos/builtin/packages/xz/package.py
```diff
@@ -61,7 +68,69 @@ def determine_version(cls, exe):
         match = re.search(r"xz \(XZ Utils\) (\S+)", output)
         return match.group(1) if match else None
 
+
+class RunAfter(object):
```
scheibelp: Nothing inherits this class. FWIW it would never be used on Windows, so you could move this method to the AutotoolsBuilder vs. using inheritance.

johnwparent: Moved to autotools builder, thanks!

var/spack/repos/builtin/packages/xz/package.py
```diff
@@ -61,7 +68,67 @@ def determine_version(cls, exe):
...
+        with working_dir(win_dir):
+            for obj in os.scandir():
+                if obj.is_dir():
+                    compiler_dirs.append(obj.name)
+        newest_compiler = max(compiler_dirs, key=get_file_string_number)
+        return os.path.join(win_dir, newest_compiler)
+
+    @property
+    def toolchain_version(self):
+        return "v" + self.pkg.compiler.platform_toolset_ver
```
scheibelp: I don't see where this is used: is this method required for this builder to succeed?

johnwparent: Yes, if you look at the MSBuild builder, it defines a `toolchain_version` property that this builder class overrides. This argument is provided to MSBuild itself by the builder and is required to ensure consistent handling of the toolchains loaded by spack vs those autodetected by the system. Without it, this build will fail inconsistently based on whatever toolchain the msbuild system happens to target during a given run.

scheibelp: It looks like by default

```
    @property
    def toolchain_version(self):
        return self.compiler.msvc_version

    @property
    def msvc_version(self):
        return Version(re.search(Msvc.version_regex, self.cc).group(1))
```

Which appears to be unchanging. Or does the value of `.cc` change? I'm not clear on:

* How this is a consistency error
* Why this isn't automatically-detectable and fixable as part of the MSBuild class itself

johnwparent: Sorry, for whatever reason I _badly_ misread your comment and thought you were questioning why the property was needed in the first place. 

This has nothing to do with consistency and everything to do with how we organized these PRs. Looking back in my notes this should be removed and the actual MSbuild builder patched to perform this operation now that #35098 has landed.
 
Essentially this property was defined so I could actually test the individual PRs after I split them up.

scheibelp: > Sorry, for whatever reason I badly misread your comment and thought you were questioning why the property was needed in the first place.

To be clear I think your original reading of https://github.com/spack/spack/pull/35095#discussion_r1096461931 was correct: I indeed missed that `toolchain_version` is used in the `MSBuild` class. That being said, my response in https://github.com/spack/spack/pull/35095#discussion_r1097945726 raises an additional concern about clarity; it sounds like you agree with that part but if that leaves anything unclear let me know.

johnwparent:
@spackbot run pipeline

var/spack/repos/builtin/packages/xz/package.py
```diff
@@ -63,7 +70,63 @@ def determine_version(cls, exe):
...
+        compiler_dirs = []
+        with working_dir(win_dir):
+            for obj in os.scandir():
+                if obj.is_dir():
+                    compiler_dirs.append(obj.name)
+        newest_compiler = max(compiler_dirs, key=get_file_string_number)
+        return os.path.join(win_dir, newest_compiler)
+
+    def is_64bit(self):
+        return platform.machine().endswith("64")
```
scheibelp: Can this work with the `spec`'s target rather than the host target?

var/spack/repos/builtin/packages/xz/package.py
```diff
@@ -63,7 +70,63 @@ def determine_version(cls, exe):
...
+            f = "liblzma_dll.vcxproj"
+        else:
+            f = "liblzma.vcxproj"
+        return [self.define("Configuration", "Release"), self.define("Platform", plat), f]
+
+    def install(self, pkg, spec, prefix):
+        with working_dir(self.build_directory):
+            # Ensure we have libs directory
+            if not os.path.isdir(prefix.lib):
+                mkdirp(prefix.lib)
```
scheibelp: Note you don't have to do an `if`: `mkdirp` doesn't fail if the directory already exists.

var/spack/repos/builtin/packages/xz/package.py
```diff
@@ -63,7 +70,63 @@ def determine_version(cls, exe):
...
+            libs_to_find = []
+            if self.pkg.spec.satisfies("libs=shared,static"):
+                libs_to_find.extend(["*.dll", "*.lib"])
+            elif self.pkg.spec.satisfies("libs=shared"):
+                libs_to_find.append("*.dll")
+            else:
+                libs_to_find.append("*.lib")
```
scheibelp, line 120: My recollection is that `.dll` files are not used for linking, so we'll want to copy over all `.lib` files regardless of the value of the `libs` variant. I suppose if `libs=static`, we might not want to copy the `.dll` files. If they are built though, I'm inclined to copy them over. It might make sense to not provide the `libs` variant on Windows (since it looks like it doesn't affect what is built).

scheibelp (changes requested):
A few questions/requests

var/spack/repos/builtin/packages/xz/package.py
johnwparent: If `libs=static` there should be no .dll files and if there are we shouldn't be copying them.  Good catch on the not copying libs for shared. Will make those changes. The variant does impact what is built... where are you seeing it doesn't change what is built, perhaps I'm overlooking something.

var/spack/repos/builtin/packages/xz/package.py
johnwparent: Change has been made locally. I've gotta say I think that's the longest accessor chain I've ever written.

var/spack/repos/builtin/packages/xz/package.py
scheibelp: > The variant does impact what is built... where are you seeing it doesn't change what is built, perhaps I'm overlooking something.

Sorry yes I missed that, so the variant should stay.

johnwparent:
@spackbot fix style

var/spack/repos/builtin/packages/xz/package.py
```diff
@@ -63,7 +70,60 @@ def determine_version(cls, exe):
...
+        else:
+            f = "liblzma.vcxproj"
+        return [self.define("Configuration", "Release"), self.define("Platform", plat), f]
+
+    def install(self, pkg, spec, prefix):
+        with working_dir(self.build_directory):
+            # Ensure we have libs directory
+            mkdirp(prefix.lib)
+            libs_to_find = []
+            if "shared" in self.pkg.spec:
```
scheibelp: I think this has to be `if libs=shared in self.pkg.spec`

johnwparent: Would `shared` appear elsewhere in the spec? If so you're correct, otherwise I think its fine.

scheibelp: This isn't an issue of potential ambiguity, this check will always be false even for `xz libs=shared`, because variant queries are either of the form:

`if "+varname" in spec`

or

`if "varname=..." in spec`

other forms are not valid variant queries (you can try this in `spack-python`). We should probably be raising an error with this query, but for some ambiguous queries it would be impossible to determine whether the query was truly an error. In particular, if you have a dag like `x -> y` you can ask `if 'y' in specx`.

johnwparent: Oh I see, I always assume those checks were performed by simple string parsing, I never realized it was actually a standardized query, thanks for taking the time to explain that!

I've been formatting them correctly up to this point purely by assuming it was a style convention (which would have necessitated this change anyway)

scheibelp (approved):

johnwparent:
@spackbot run pipeline

PR #35098, opened 2023-01-23
MSVC compiler: add platform toolset versioning

johnwparent:
Add the platform toolset version associated with active MSVC compiler as a property of the MSVC compiler class.

This is necessary to support #34938 in general and supports features in #35095 and #35099 directly. This should be merged prior to the aforementioned PRs as they both rely on behavior from this PR. All changes from these PRs were originally made in #34936 as one unit.

### DIFF
--- a/lib/spack/spack/compilers/msvc.py
+++ b/lib/spack/spack/compilers/msvc.py
@@ -113,8 +113,9 @@ class Msvc(Compiler):
     def platform_toolset_ver(self):
         """
         This is the platform toolset version of current MSVC compiler
-        of form v<toolset-ver>, i.e. v142. This is different from the VC
-        toolset version as established by `short_msvc_version`
+        i.e. 142.
+        This is different from the VC toolset version as established
+        by `short_msvc_version`
         """
         return self.msvc_version[:2].joined.string[:3]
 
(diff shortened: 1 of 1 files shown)


lib/spack/spack/compilers/msvc.py
```diff
@@ -103,11 +103,21 @@ def short_msvc_version(self):
...
         return "MSVC" + ver
 
+    @property
+    def platform_toolset_ver(self):
+        """
+        This is the platform toolset version of current MSVC compiler
+        of form v<toolset-ver>, i.e. v142. This is different from the VC
+        toolset version as established by `short_msvc_version`
+        """
+        return self.msvc_version[:2].joined.string[:3]
```
scheibelp, line 120: If this should be returning for example "v142", then I'll note that `[:3]` takes the first 3 characters. If it in fact should be "142" then this looks right (although I think the docstring should be clarified in that case).

johnwparent: It should only be 142, and packages are left to prepend the 'v' character if necessary. I'll update the docstring.

scheibelp (approved):

PR #35099, opened 2023-01-23
Windows: port theora & ogg

johnwparent:
Ports for libtheora and libogg (and related changes to the MSVC compiler class to support that end) to the multi build system and their own Windows supported builders. 

This PR supports #34938 and depends on #35098 before it can go in.

Note: libtheora depends on libogg and changes to libtheora require an installable libogg on Windows to test so these package changes should be grouped together and this PR should not be closed on those grounds.

### DIFF
--- a/lib/spack/spack/compilers/msvc.py
+++ b/lib/spack/spack/compilers/msvc.py
@@ -124,12 +124,15 @@ class Msvc(Compiler):
         """Cl toolset version"""
         return Version(
             re.search(
-                Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, "")
+                Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, version_arg=None)
             ).group(1)
         )
 
     @property
     def vs_root(self):
+        # The MSVC install root is located at a fix level above the compiler
+        # and is referenceable idiomatically via the pattern below
+        # this should be consistent accross versions
         return os.path.abspath(os.path.join(self.cc, "../../../../../../../.."))
 
     def setup_custom_environment(self, pkg, env):
--- a/lib/spack/spack/compilers/msvc.py
+++ b/lib/spack/spack/compilers/msvc.py
@@ -124,7 +124,8 @@ class Msvc(Compiler):
         """Cl toolset version"""
         return Version(
             re.search(
-                Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, version_arg=None)
+                Msvc.version_regex,
+                spack.compiler.get_compiler_version_output(self.cc, version_arg=None),
             ).group(1)
         )
 
--- a/lib/spack/spack/compilers/msvc.py
+++ b/lib/spack/spack/compilers/msvc.py
@@ -122,7 +122,11 @@ class Msvc(Compiler):
     @property
     def cl_version(self):
         """Cl toolset version"""
-        return Version(re.search(Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, "")).group(1))
+        return Version(
+            re.search(
+                Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, "")
+            ).group(1)
+        )
 
     @property
     def vs_root(self):
--- a/lib/spack/spack/compilers/msvc.py
+++ b/lib/spack/spack/compilers/msvc.py
@@ -122,7 +122,11 @@ class Msvc(Compiler):
     @property
     def cl_version(self):
         """Cl toolset version"""
-        return spack.compiler.get_compiler_version_output(self.cc)
+        return Version(re.search(Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, "")).group(1))
+
+    @property
+    def vs_root(self):
+        return os.path.abspath(os.path.join(self.cc, "../../../../../../../.."))
 
     def setup_custom_environment(self, pkg, env):
         """Set environment variables for MSVC using the
--- a/var/spack/repos/builtin/packages/libogg/package.py
+++ b/var/spack/repos/builtin/packages/libogg/package.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import os
-import platform
 
 from spack.build_systems.generic import GenericBuilder
 from spack.package import *
--- a/var/spack/repos/builtin/packages/libogg/package.py
+++ b/var/spack/repos/builtin/packages/libogg/package.py
@@ -40,7 +40,7 @@ class GenericBuilder(GenericBuilder):
     phases = ["build", "install"]
 
     def is_64bit(self):
-        return platform.machine().endswith("64")
+        return "64" in self.pkg.spec.target.family
 
     def build(self, spec, prefix):
         if spec.satisfies("%msvc"):
--- a/var/spack/repos/builtin/packages/libogg/package.py
+++ b/var/spack/repos/builtin/packages/libogg/package.py
@@ -3,10 +3,14 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import os
+import platform
+
+from spack.build_systems.generic import GenericBuilder
 from spack.package import *
 
 
-class Libogg(AutotoolsPackage):
+class Libogg(CMakePackage, AutotoolsPackage, Package):
     """Ogg is a multimedia container format, and the native file and stream
     format for the Xiph.org multimedia codecs."""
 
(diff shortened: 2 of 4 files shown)


johnwparent:
@spackbot run pipeline

var/spack/repos/builtin/packages/libtheora/package.py
```diff
@@ -40,7 +67,64 @@ def autoreconf(self, spec, prefix):
...
+        env.prepend_path("PATH", devenv_path)
+
+    @property
+    def build_directory(self):
+        win_dir = os.path.join(super().build_directory, "win32")
+        vs_dir = "VS2008"
+        return os.path.join(win_dir, vs_dir)
+
+    @property
+    def toolchain_version(self):
```
scheibelp: I have a similar question here as for https://github.com/spack/spack/pull/35095#discussion_r1096461931 - I don't see this method being used anywhere in this PR.

johnwparent: Response is the same as it is for #35095, this method is required by the builder.

johnwparent:
@spackbot run pipeline

lib/spack/spack/compilers/msvc.py
```diff
@@ -122,7 +122,15 @@ def platform_toolset_ver(self):
...
-        return spack.compiler.get_compiler_version_output(self.cc)
+        return Version(
+            re.search(
+                Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, "")
+            ).group(1)
+        )
+
+    @property
+    def vs_root(self):
+        return os.path.abspath(os.path.join(self.cc, "../../../../../../../.."))
```
scheibelp, line 137: This looks fragile, is there some way to query the compiler for the root? If not, can a docstring mention the structure?

lib/spack/spack/compilers/msvc.py
```diff
@@ -122,7 +122,15 @@ def platform_toolset_ver(self):
     @property
     def cl_version(self):
         """Cl toolset version"""
-        return spack.compiler.get_compiler_version_output(self.cc)
+        return Version(
+            re.search(
+                Msvc.version_regex, spack.compiler.get_compiler_version_output(self.cc, "")
```
scheibelp: IMO slightly clearer if you do:

```
- get_compiler_version_output(self.cc, "")
+ get_compiler_version_output(self.cc, version_arg=None)
```

var/spack/repos/builtin/packages/libogg/package.py
```diff
@@ -24,3 +28,35 @@ class Libogg(AutotoolsPackage):
...
+    phases = ["build", "install"]
+
+    def is_64bit(self):
+        return platform.machine().endswith("64")
+
+    def build(self, spec, prefix):
+        if spec.satisfies("%msvc"):
+            plat_tools = self.pkg.compiler.msvc_version
+        else:
+            raise RuntimeError("Package does not support non MSVC compilers on Windows")
```
scheibelp, line 48: (no change needed) Ideally this would be encoded as `conflict`s. Given that it's onerous to do it that way, this can persist for now.

var/spack/repos/builtin/packages/libtheora/package.py
```diff
@@ -40,7 +67,60 @@ def autoreconf(self, spec, prefix):
...
+            os.path.join(prefix.lib, "theora.lib"),
+        )
+        # The encoder and decoder libs are baked into the theora lib on Windows, so we spoof them
+        # here so libtheora is detectable by CMake and pkg-config and linkable by
+        # consuming projects
+        with working_dir(prefix.lib):
+            copy("theora.lib", "theoradec.lib")
+            copy("theora.lib", "theoraenc.lib")
+        with working_dir(pkg.stage.source_path):
+            install_tree("include", prefix.include)
```
scheibelp, line 128: (question) is a `.dll` installed for `~static`? It wasn't clear to me looking through this function (i.e. I don't see a search for `.dll`).

scheibelp (changes requested):
I have a couple requests and a question

lib/spack/spack/compilers/msvc.py
johnwparent: We rely on almost exactly the same logic in WindowsOS to find the VCVARS script. One of the nice things about MSVC is its relatively consistent structure. so while this looks incredibly fragile. it should be solid, and if it does break, our VCVARS detection is also broken so nothing will compile anyway.

johnwparent: I'll add a docstring.

lib/spack/spack/compilers/msvc.py
johnwparent: Sounds good!

var/spack/repos/builtin/packages/libogg/package.py
johnwparent: That was my thinking as well.

var/spack/repos/builtin/packages/libtheora/package.py
johnwparent: Very likely yes. This was an error on my part as I was too focused on what Paraview needs. I'll add the dlls to the install line.

johnwparent:
@scheibelp it looks like this gitlab failure might actually be real but I'm unable to replicate locally on my linux VM and can't really see how I might have impacted the autotools build with this PR. Any thoughts?

var/spack/repos/builtin/packages/libtheora/package.py
```diff
@@ -40,7 +67,63 @@ def autoreconf(self, spec, prefix):
...
             sh("./autogen.sh", "prefix={0}".format(prefix))
 
-    def configure_args(self):
-        args = []
-        args += self.enable_or_disable("doc")
-        return args
+
+class MSBuildBuilder(MSBuildBuilder):
+    def is_64bit(self):
+        return "64" in self.pkg.spec.target.family
```
johnwparent, line 72: @scheibelp I've refactored this the same way it was done for XZ. That seem right to you as well?

var/spack/repos/builtin/packages/libtheora/package.py
johnwparent: This has been done.

var/spack/repos/builtin/packages/libtheora/package.py
scheibelp: Yes

scheibelp:
(EDIT: resolved)

> @scheibelp it looks like this gitlab failure might actually be real but I'm unable to replicate locally on my linux VM and can't really see how I might have impacted the autotools build with this PR. Any thoughts?

I think you mean https://gitlab.spack.io/spack/spack/-/jobs/5510672?  To be clear you can build `libogg` and `libtheora` on your VM? Can you build `libogg build_system=cmake` as well as `libogg build_system=autotools`?

I agree that error message isn't much to go off of, I just see:

```
==> Error: Failed to install libogg due to SystemExit: 1
```

and

* I can't find any `spack-build-out.txt` amongst the artifacts (i.e. in the "Job Artifacts" section of https://gitlab.spack.io/spack/spack/-/jobs/5510672)
* I can't find a specific link for `libogg` (the prerequisite for `libtheora`) in https://gitlab.spack.io/spack/spack/-/pipelines/273940 (it doesn't seem to show stages before `libtheora`)

@scottwittenburg do you know where I could locate those?

var/spack/repos/builtin/packages/libogg/package.py
```diff
@@ -24,3 +28,35 @@ class Libogg(AutotoolsPackage):
...
+        "autotools",
+        default="autotools",
+    )
+
+
+class GenericBuilder(GenericBuilder):
+    phases = ["build", "install"]
+
+    def is_64bit(self):
+        return platform.machine().endswith("64")
```
scheibelp: Like https://github.com/spack/spack/pull/35095, can this analyze the spec to determine 64-bit support?

Other than this the PR looks good.

johnwparent: Oh huh I thought I already went back and did this to all my prs, guess I missed it here, good catch!

johnwparent: I'll also make a new PR converting perl to this style

johnwparent:
@spackbot run pipeline

johnwparent:
Gitlab pipeline previously succeeded and no changes that impact the linux builds have been introduced since. Rerunning for that reason, and because current failures seem unrelated to theora/ogg.

johnwparent:
Gitlab pipeline is failing due to a job exceeding execution time constraints, same job has failed multiple times for this reason, @scheibelp any thoughts?

scheibelp:
I'm not sure:

* `visit` depends on `libogg` and `libtheora` (via `vtk`)
* The changes here should produce exactly the same build instructions on Linux 
* This is failing on a Power/Linux system, but other builds of `visit` are succeeding in this pipeline

I should ask the folks who manage the CI about it.

johnwparent:
@scheibelp How do you want to proceed on this one? I ran the power PC visit build via QEMU and nothing failed for me (although it did take forever, possibly just because QEMU).

scheibelp (approved):

PR #35101, opened 2023-01-23
Windows: Port lz4

johnwparent:
lz4 - add CMake build system

lz4 build in serial on Windows - otherwise there is a transient build error caused by Ninja over parallelizing the build and dependent targets being built out of order due to the structure of the lz4 CMake system.

lz4 - multi buildsystem refactor

Part of #34938

### DIFF
--- a/var/spack/repos/builtin/packages/lz4/package.py
+++ b/var/spack/repos/builtin/packages/lz4/package.py
@@ -3,12 +3,15 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import os
 import sys
 
+from spack.build_systems.cmake import CMakeBuilder
+from spack.build_systems.makefile import MakefileBuilder
 from spack.package import *
 
 
-class Lz4(MakefilePackage):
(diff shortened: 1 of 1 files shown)


scheibelp (approved):

var/spack/repos/builtin/packages/lz4/package.py
```diff
@@ -44,6 +49,39 @@ def url_for_version(self, version):
...
+        # Remove flags not recognized by the NVIDIA compiler
+        if self.spec.satisfies("%nvhpc@:20.11"):
+            filter_file("-fvisibility=hidden", "", "Makefile")
+            filter_file("-fvisibility=hidden", "", "lib/Makefile")
+            filter_file("-pedantic", "", "Makefile")
+
+    @run_after("install")
+    def darwin_fix(self):
+        if sys.platform == "darwin":
+            fix_darwin_install_name(self.prefix.lib)
```
alalazo, line 62: Once you add builders, these methods should be on the builders (otherwise they are not taken into consideration). I'll check how easy it is to add an audit for stuff like this.

PR #35102, opened 2023-01-23
Windows: Port adios2

johnwparent:
Small changes to Adios2 package to support #34938

### DIFF
--- a/var/spack/repos/builtin/packages/adios2/package.py
+++ b/var/spack/repos/builtin/packages/adios2/package.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import os
+import sys
 
 from spack.package import *
 
@@ -54,7 +55,11 @@ class Adios2(CMakePackage, CudaPackage):
     # change how we're supporting differnt library types in the package at anytime if
     # spack decides on a standardized way of doing it across packages
     variant("shared", default=True, when="+pic", description="Build shared libraries")
-    variant("pic", default=True, description="Build pic-enabled static libraries")
+    variant(
+        "pic",
+        default=not sys.platform == "win32",
+        description="Build pic-enabled static libraries",
+    )
 
     # Features
     variant("mpi", default=True, description="Enable MPI")
@@ -95,12 +100,18 @@ class Adios2(CMakePackage, CudaPackage):
     conflicts("%oneapi@:2022.1.0", when="+fortran")
 
     depends_on("cmake@3.12.0:", type="build")
-    depends_on("pkgconfig", type="build")
 
-    depends_on("libffi", when="+sst")  # optional in DILL
-    depends_on("libfabric@1.6.0:", when="+sst")  # optional in EVPath and SST
-    # depends_on('bison', when='+sst')     # optional in FFS, broken package
-    # depends_on('flex', when='+sst')      # optional in FFS, depends on BISON
+    for plat in ["linux", "darwin", "cray"]:
+        depends_on("pkgconfig", type="build", when="platform = %s" % plat)
+
+        # libffi and libfabric and not currently supported on Windows
+        # see Paraview's superbuild handling of libfabric at https://gitlab.kitware.com/paraview/paraview-superbuild/-/blob/master/projects/adios2.cmake#L3
+        depends_on("libffi", when="+sst platform=%s" % plat)  # optional in DILL
+        depends_on(
+            "libfabric@1.6.0:", when="+sst platform=%s" % plat
+        )  # optional in EVPath and SST
+        # depends_on('bison', when='+sst')     # optional in FFS, broken package
+        # depends_on('flex', when='+sst')      # optional in FFS, depends on BISON
 
     depends_on("mpi", when="+mpi")
     depends_on("libzmq", when="+dataman")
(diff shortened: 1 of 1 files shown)


williamfgc:
LGTM, glad to see Windows support!

williamfgc (dismissed):

var/spack/repos/builtin/packages/adios2/package.py
```diff
@@ -95,12 +100,18 @@ class Adios2(CMakePackage, CudaPackage):
...
 
-    depends_on("libffi", when="+sst")  # optional in DILL
-    depends_on("libfabric@1.6.0:", when="+sst")  # optional in EVPath and SST
-    # depends_on('bison', when='+sst')     # optional in FFS, broken package
-    # depends_on('flex', when='+sst')      # optional in FFS, depends on BISON
+    for plat in ["linux", "darwin", "cray"]:
+        depends_on("pkgconfig", type="build", when="platform = %s" % plat)
+
+        # libffi and libfabric and not currently supported on Windows
+        # see Paraview's superbuild handling of libfabric at https://gitlab.kitware.com/paraview/paraview-superbuild/-/blob/master/projects/adios2.cmake#L3
```
vicentebolea: break long line in two

var/spack/repos/builtin/packages/adios2/package.py
```diff
@@ -95,12 +100,18 @@ class Adios2(CMakePackage, CudaPackage):
     conflicts("%oneapi@:2022.1.0", when="+fortran")
 
     depends_on("cmake@3.12.0:", type="build")
-    depends_on("pkgconfig", type="build")
 
-    depends_on("libffi", when="+sst")  # optional in DILL
-    depends_on("libfabric@1.6.0:", when="+sst")  # optional in EVPath and SST
-    # depends_on('bison', when='+sst')     # optional in FFS, broken package
-    # depends_on('flex', when='+sst')      # optional in FFS, depends on BISON
+    for plat in ["linux", "darwin", "cray"]:
```
vicentebolea: can we use `platform` instead of `plat`, it makes it more clear at the expense of 4 more chars.

var/spack/repos/builtin/packages/adios2/package.py
```diff
@@ -164,6 +175,45 @@ def setup_build_environment(self, env):
         elif self.spec.satisfies("%fj +fortran"):
             env.set("FFLAGS", "-Ccpp")
 
+    @property
```
vicentebolea: Why moving this two functions blocks? If considered important for the consistency of the file, please do it in a independent commit

var/spack/repos/builtin/packages/adios2/package.py
```diff
@@ -54,7 +55,11 @@ class Adios2(CMakePackage, CudaPackage):
     # change how we're supporting differnt library types in the package at anytime if
     # spack decides on a standardized way of doing it across packages
     variant("shared", default=True, when="+pic", description="Build shared libraries")
-    variant("pic", default=True, description="Build pic-enabled static libraries")
+    variant(
+        "pic",
+        default=not sys.platform == "win32",
```
vicentebolea: I wonder if a conflict() will be better than this. I am not 100% sure it is the case.

vicentebolea (changes requested):
@johnwparent Many thanks for the interest and effort to make the ADIOS2 spack package compatible with Windows and submitting this, it looks great. Just a couple of suggestions.

var/spack/repos/builtin/packages/adios2/package.py
johnwparent: This is an artifact of the significant amount of history manipulation this file has been exposed to, it can/will be removed.

var/spack/repos/builtin/packages/adios2/package.py
johnwparent: You're right, this isn't the best way we could be handling this, so I'm going to move this to the same stanza that loops over the platforms as the concept doesn't exist on Windows.

var/spack/repos/builtin/packages/adios2/package.py
johnwparent: `platform` is also the name of a python module and the platform designation in the spec so there's avenue for confusion and actual errors there as well. I feel like `plat` vs `platform` is a fairly minor distinction and FWIW we use `plat` in almost every case where this sentiment is expressed. I have no problem with changing the name here, but if we're doing this for clarity sake, we should consider all factors and consistency.

var/spack/repos/builtin/packages/adios2/package.py
johnwparent: Good point!

alalazo:
@spackbot fix style

var/spack/repos/builtin/packages/adios2/package.py
vicentebolea: Fair point with regard that it can also add confusion due to existing symbols being named as platform as well, although there is no naming overwrite.

I think there is still a net good of being a bit more verbose in this file since spack packages are often read by Spack users who can be very unfamiliar with both spack and ADIOS2, but for instance they might needs figure out which variant does what and or something else.

Alternatively, as we do in the C++/C world, how about underscoring the name of the variable such as `platform_`, it might look ugly in python but it will satisfy both requirements: clarity and unique name.

vicentebolea (approved):
Many thanks for addressing the issues.

alalazo (approved):
Approved by maintainer, changes in directives LGTM

PR #35105, opened 2023-01-23
Convert libpng to build with CMake

johnwparent:
Refactor libpng to be built with CMake instead of autotools and in doing so, support Windows.

Part of #34938

### DIFF
--- a/var/spack/repos/builtin/packages/libpng/package.py
+++ b/var/spack/repos/builtin/packages/libpng/package.py
@@ -3,8 +3,9 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack.package import *
 from spack.build_systems.cmake import CMakeBuilder
+from spack.package import *
+
 
 class Libpng(CMakePackage):
     """libpng is the official PNG reference library."""
@@ -43,5 +44,5 @@ class CMakeBuilder(CMakeBuilder):
             self.define("CMAKE_CXX_FLAGS", self.spec["zlib"].headers.include_flags),
             self.define("ZLIB_ROOT", self.spec["zlib"].prefix),
             self.define("PNG_SHARED", "shared" in self.pkg.variants["libs"]),
-            self.define("PNG_STATIC", "static" in self.pkg.variants["libs"])
+            self.define("PNG_STATIC", "static" in self.pkg.variants["libs"]),
         ]
--- a/var/spack/repos/builtin/packages/libpng/package.py
+++ b/var/spack/repos/builtin/packages/libpng/package.py
@@ -43,6 +43,6 @@ class CMakeBuilder(CMakeBuilder):
         return [
             self.define("CMAKE_CXX_FLAGS", self.spec["zlib"].headers.include_flags),
             self.define("ZLIB_ROOT", self.spec["zlib"].prefix),
-            self.define("PNG_SHARED", "shared" in self.pkg.variants["libs"]),
-            self.define("PNG_STATIC", "static" in self.pkg.variants["libs"]),
+            self.define("PNG_SHARED", "shared" in self.spec.variants["libs"].value),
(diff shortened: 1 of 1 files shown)


AlexanderRichert-NOAA (changes requested):
This PR breaks existing functionality. "configure_args" should be removed and the functionality in it should be updated for cmake.

johnwparent:
> This PR breaks existing functionality. "configure_args" should be removed and the functionality in it should be updated for cmake.

Thanks for pointing that out. The proper refactor must have been dropped when I cannibalized this from a few different PRs. I've added the actual cmake args back from it's original state. Feel free to re-review.

AlexanderRichert-NOAA:
The default values for PNG_SHARED and PNG_STATIC in CMakeLists.txt are both ON. In the current version, the use of "enable_or_disable" means that when a given option isn't specified, it gets the "--no-..." configure option, so that logic should get reproduced here.

johnwparent:
> The default values for PNG_SHARED and PNG_STATIC in CMakeLists.txt are both ON. In the current version, the use of "enable_or_disable" means that when a given option isn't specified, it gets the "--no-..." configure option, so that logic should get reproduced here.

Thanks, I didn't realize that's how that helper worked. I'll refactor now.

johnwparent:
> The default values for PNG_SHARED and PNG_STATIC in CMakeLists.txt are both ON. In the current version, the use of "enable_or_disable" means that when a given option isn't specified, it gets the "--no-..." configure option, so that logic should get reproduced here.

Perhaps I'm misunderstanding.

The way the package is right now in the PR if the spec is `libpng libs=shared`, `PNG_SHARED` will be True, while `PNG_STATIC` will be false because the string `"static"` will not be present in the spec. When the spec is `libpng libs=static`, `PNG_SHARED` will be false and `PNG_STATIC` will be true for the same reason as before. The only place where I can see a potential disconnect is the default behavior. If the spec is just `libpng` then the libs default is `shared,static`, and with the behavior I add in this PR,  this means both `PNG_SHARED` and `PNG_STATIC` will be true, should this not be the case?

AlexanderRichert-NOAA:
I would suggest testing it out. From what I can tell, what's going on is that PNG_SHARED and PNG_STATIC are both default ON in CMakeLists.txt, therefore if eitheir of them is not explicitly set to OFF in package.py, then it will be ON. Does that make sense?

johnwparent:
`PNG_STATIC` and `PNG_SHARED` are on by default in the CMake interface for libpng. This is correct. However Spack is _never_ not overridding the defaults. We are _always_ providing CMake some definition for these options on the command line. As stated above, the way it is currently configured, if _neither_ `libs=shared` or `libs=static` are specified by the calling spec, the default case is triggered, where both are True. If one or the other is specified, then either `"static" in self.spec.variants["libs"].value` or `"shared" in self.spec.variants["libs"]` will evaluate to false. This sets the cmake command line to `PNG_STATIC` or `PNG_SHARED` to `OFF` depending on which was specified. 

Testing simply confirms what I have stated above.

AlexanderRichert-NOAA:
Okay this looks good. I may have tested incorrectly earlier in which case apologies for the confusion.

AlexanderRichert-NOAA (approved):

scheibelp (approved):

climbfuji:
@scheibelp We've run into issues building `libpng` on macOS arch64 after this PR: https://github.com/NOAA-EMC/spack-stack/issues/572

Have you encountered that? Does the spack CI system cover macOS arch64 ?

climbfuji:
@scheibelp See here for a solution: https://github.com/NOAA-EMC/spack-stack/issues/572#issuecomment-1542620688

We'll get it fixed in our fork appropriately and then push the commit for a PR to spack develop.

climbfuji:
PR is here: https://github.com/spack/spack/pull/37613

PR #35107, opened 2023-01-23
Freetype: to CMake

johnwparent:
Convert Freetype to use CMake to support Windows

part of #34938

### DIFF
--- a/var/spack/repos/builtin/packages/freetype/package.py
+++ b/var/spack/repos/builtin/packages/freetype/package.py
@@ -30,6 +30,8 @@ class Freetype(AutotoolsPackage, CMakePackage):
     version("2.7", sha256="7b657d5f872b0ab56461f3bd310bd1c5ec64619bd15f0d8e08282d494d9cfea4")
     version("2.5.3", sha256="41217f800d3f40d78ef4eb99d6a35fd85235b64f81bc56e4812d7672fca7b806")
 
+    build_system("cmake", "autotools", default="autotools")
+
     depends_on("bzip2")
     depends_on("libpng")
     for plat in ["linux", "darwin", "cray"]:
--- a/var/spack/repos/builtin/packages/freetype/package.py
+++ b/var/spack/repos/builtin/packages/freetype/package.py
@@ -3,7 +3,6 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack.build_systems.autotools import AutotoolsBuilder
 from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
 
@@ -51,18 +50,10 @@ class Freetype(CMakePackage):
 
 class CMakeBuilder(CMakeBuilder):
     def cmake_args(self):
-        return [self.define("FT_WITH_ZLIB", True), self.define("FT_WITH_BZIP2", True)]
-
-
-class AutoToolsBuilder(AutotoolsBuilder):
-    def configure_args(self):
-        args = [
-            "--with-brotli=no",
-            "--with-bzip2=yes",
-            "--with-harfbuzz=no",
-            "--with-png=yes",
-            "--with-zlib=no",
+        return [
+            self.define("FT_DISABLE_ZLIB", True),
+            self.define("FT_DISABLE_BROTLI", True),
+            self.define("FT_DISABLE_HARFBUZZ", True),
+            self.define("FT_DISABLE_PNG", True),
+            self.define("FT_WITH_BZIP2", True),
         ]
-        if self.spec.satisfies("@2.9.1:"):
-            args.append("--enable-freetype-config")
-        return args
--- a/var/spack/repos/builtin/packages/freetype/package.py
+++ b/var/spack/repos/builtin/packages/freetype/package.py
@@ -3,11 +3,12 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack.build_systems.autotools import AutotoolsBuilder
 from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
 
 
-class Freetype(CMakePackage):
+class Freetype(AutotoolsPackage, CMakePackage):
     """FreeType is a freely available software library to render fonts.
     It is written in C, designed to be small, efficient, highly customizable,
     and portable while capable of producing high-quality output (glyph images)
@@ -31,6 +32,8 @@ class Freetype(CMakePackage):
 
     depends_on("bzip2")
     depends_on("libpng")
+    for plat in ["linux", "darwin", "cray"]:
+        depends_on("pkgconfig", type="build", when="platform=%s" % plat)
 
     conflicts(
         "%intel",
@@ -48,12 +51,26 @@ class Freetype(CMakePackage):
         return headers
 
 
+class AutotoolsBuilder(AutotoolsBuilder):
+    def configure_args(self):
+        args = [
+            "--with-brotli=no",
+            "--with-bzip2=yes",
+            "--with-harfbuzz=no",
+            "--with-png=yes",
+            "--with-zlib=no",
+        ]
+        if self.spec.satisfies("@2.9.1:"):
+            args.append("--enable-freetype-config")
+        return args
+
+
 class CMakeBuilder(CMakeBuilder):
     def cmake_args(self):
         return [
             self.define("FT_DISABLE_ZLIB", True),
             self.define("FT_DISABLE_BROTLI", True),
             self.define("FT_DISABLE_HARFBUZZ", True),
-            self.define("FT_DISABLE_PNG", True),
-            self.define("FT_WITH_BZIP2", True),
+            self.define("FT_REQUIRE_PNG", True),
+            self.define("FT_REQUIRE_BZIP2", True),
         ]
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -3,10 +3,12 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack.build_systems.autotools import AutotoolsBuilder
+from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
 
 
-class Freetype(AutotoolsPackage):
+class Freetype(CMakePackage):
```
michaelkuhn: Shouldn't this include `AutotoolsPackage`? Even when adding it, `autotools` does not show up as a valid choice for `build_system`, though:
```
==> Error: invalid values for variant "build_system" in package "freetype": ['autotools']
```

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -47,6 +48,13 @@ def headers(self):
         headers.directories = [self.prefix.include.freetype2]
         return headers
 
+
+class CMakeBuilder(CMakeBuilder):
+    def cmake_args(self):
+        return [self.define("FT_WITH_ZLIB", True), self.define("FT_WITH_BZIP2", True)]
+
+
+class AutoToolsBuilder(AutotoolsBuilder):
```
michaelkuhn: Not sure if it makes a difference: `Autotools` instead of `AutoTools`?

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -47,6 +48,13 @@ def headers(self):
         headers.directories = [self.prefix.include.freetype2]
         return headers
 
+
+class CMakeBuilder(CMakeBuilder):
+    def cmake_args(self):
+        return [self.define("FT_WITH_ZLIB", True), self.define("FT_WITH_BZIP2", True)]
```
michaelkuhn: 1. Why is `FT_WITH_ZLIB` true? It's explicitly disabled for Autotools. We also do not have a `zlib` dependency.
2. What about the other options? This should probably enable/disable the same features as the Autotools variant.

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -30,7 +32,6 @@ class Freetype(AutotoolsPackage):
 
     depends_on("bzip2")
     depends_on("libpng")
-    depends_on("pkgconfig", type="build")
```
michaelkuhn, line 33: Why is this not required anymore?

michaelkuhn (changes requested):

johnwparent:
@michaelkuhn This was taken from another PR. Looks like I didn't cherry-pick enough history, let me grab the rest of it, I think you'll find the port is much better.

My apologies for this mistake.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: This will not be required as we are not longer using autotools.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: This was a mistake I made during a (very old) previous iteration. It was _supposed_ to be `FT_DISABLE_ZLIB` set to ON. This is fixed on the current state.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: This has been removed all together in the latest revisions.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: CMake is the buildsystem for this package now. We no longer want to inherit autotools at that point.

michaelkuhn: Does it hurt to keep Autotools? It's still Freetype's default build system.

var/spack/repos/builtin/packages/freetype/package.py
michaelkuhn: CMake still seems to use pkg-config, though: https://github.com/freetype/freetype/blob/master/CMakeLists.txt#L251

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -47,14 +47,13 @@ def headers(self):
...
-            "--with-png=yes",
-            "--with-zlib=no",
+
+class CMakeBuilder(CMakeBuilder):
+    def cmake_args(self):
+        return [
+            self.define("FT_DISABLE_ZLIB", True),
+            self.define("FT_DISABLE_BROTLI", True),
+            self.define("FT_DISABLE_HARFBUZZ", True),
+            self.define("FT_DISABLE_PNG", True),
```
michaelkuhn: Why do we disable PNG? We used to enable it and still have a `libpng` dependency.

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -47,14 +47,13 @@ def headers(self):
...
+            self.define("FT_WITH_BZIP2", True),
```
michaelkuhn: Are you sure this is correct?
```
CMake Warning:
  Manually-specified variables were not used by the project:

    BUILD_TESTING
    FT_WITH_BZIP2
```

var/spack/repos/builtin/packages/freetype/package.py
```diff
@@ -47,14 +47,13 @@ def headers(self):
...
         ]
-        if self.spec.satisfies("@2.9.1:"):
-            args.append("--enable-freetype-config")
```
michaelkuhn, line 59: It seems the CMake version is missing `freetype-config`. We should keep the same behavior.

michaelkuhn (changes requested):

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: That was me reading the wrong line. Good catch.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: ~~I took it from freetype's cmake file, let me double check.~~ My mistake, you're right this is not the correct variable.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: I don't personally see utility in maintaining two build systems when one should suffice and provides improved cross platform support. While Spack does support a package having multiple builders, I think avoiding it where possible is ideal.

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: It's certainly not a hard requirement as I am able to build perfectly fine without it on both Unix and Windows, and as evidenced by the lack of `REQUIRED` in the find package invocation. Many find package modules may use PkgConfig to perform some package detection, but it's not required, and the only direct use I see is the `pkg_check_modules` which comes from the CMake module so it'll be available anyway. I'll make the requirement optional and off on Windows and on everywhere else, I think that's probably best, as you're right, other platforms may get some utility out of it (although I'm not sure how much)

var/spack/repos/builtin/packages/freetype/package.py
johnwparent: Perhaps you can elaborate here as I'm not particularly familiar with Freetype's build system but would that not be the flag for generating the config files that CMake generates by default?  Otherwise I cannot find a reference to the config files in the CMake interface. If this is indeed not a true paralell, you can disregard my comment regarding dropping autotools altogether, as you're right we don't want to drop functionality.

var/spack/repos/builtin/packages/freetype/package.py
michaelkuhn: Perfect, thanks!

var/spack/repos/builtin/packages/freetype/package.py
michaelkuhn: This flag causes a `freetype-config` binary to be generated that can be used similarly to `pkg-config freetype`. I think some of Freetype's dependents might need it.

johnwparent: After digging into things a bit, CMake properly exports this project and produces a freetype config that in general should be sufficient for dependent projects, but I found a few that truly required the autotools style config that can be treated like a binary so I'm adding back the autotools build. When/If those packages are ever ported to be solely cmake projects, this can be just a cmake package, but until then, autotools will need to be an option on unix.

johnwparent: So feel free to disregard my "removing autotools entirely" comment and related thoughts, as they are wrong.

johnwparent:
@spackbot run pipeline

michaelkuhn (approved):
Thanks for your work on this!

johnwparent:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

PR #35108, opened 2023-01-23
Proj: to CMake

johnwparent:
Convert Proj to CMake builder to support Windows

Part of #34938

### DIFF
--- a/var/spack/repos/builtin/packages/proj/package.py
+++ b/var/spack/repos/builtin/packages/proj/package.py
@@ -109,6 +109,9 @@ class Setup:
     def setup_dependent_build_environment(self, env, dependent_spec):
         self.pkg.setup_run_environment(env)
 
+    def setup_build_environment(self, env):
+        env.set("PROJ_LIB", join_path(self.pkg.stage.source_path, "nad") )
+
 
 class CMakeBuilder(cmake.CMakeBuilder, Setup):
     def cmake_args(self):
@@ -123,7 +126,7 @@ class CMakeBuilder(cmake.CMakeBuilder, Setup):
 
 class AutotoolsBuilder(autotools.AutotoolsBuilder, Setup):
     def configure_args(self):
-        args = ["PROJ_LIB={0}".format(join_path(self.stage.source_path, "nad"))]
+        args = ["PROJ_LIB={0}".format(join_path(self.pkg.stage.source_path, "nad"))]
 
         if self.spec.satisfies("@6:"):
             args.append("--with-external-gtest")
@@ -135,7 +138,7 @@ class AutotoolsBuilder(autotools.AutotoolsBuilder, Setup):
                 args.append("--disable-tiff")
 
             if "+curl" in self.spec:
-                args.append("--with-curl=" + self.spec["curl"].prefix.bin.join("curl-config"))
+                args.append("--with-curl=" + self.pkg.spec["curl"].prefix.bin.join("curl-config"))
             else:
                 args.append("--without-curl")
 
--- a/var/spack/repos/builtin/packages/proj/package.py
+++ b/var/spack/repos/builtin/packages/proj/package.py
@@ -4,9 +4,10 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
+from spack.build_systems import autotools, cmake
 
 
-class Proj(CMakePackage):
+class Proj(CMakePackage, AutotoolsPackage):
     """PROJ is a generic coordinate transformation software, that transforms
     geospatial coordinates from one coordinate reference system (CRS) to
     another. This includes cartographic projections as well as geodetic
@@ -87,12 +88,14 @@ class Proj(CMakePackage):
     depends_on("sqlite@3.11:", when="@6:")
     depends_on("libtiff@4.0:", when="@7:+tiff")
     depends_on("curl@7.29.0:", when="@7:+curl")
+    depends_on("pkgconfig@0.9.0:", type="build", when="@6: build_system=autotools")
+    depends_on("cmake@2.6.0:", type="build", when="build_system=cmake")
+    # Builds for Windows via CMake available only
+    # after this version - builds before exist
+    # for Windows but are targeted at VS2005
+    conflicts("platform=windows", when="@:5.0.0")
 
-    def cmake_args(self):
-        return [
-            self.define("PROJ_LIB", join_path(self.stage.source_path, "nad")),
-            self.define_from_variant("ENABLE_TIFF", "tiff"),
-        ]
+    build_system("autotools", conditional("cmake", when="@5.0.0:"), default="autotools")
 
     def setup_run_environment(self, env):
         # PROJ_LIB doesn't need to be set. However, it may be set by conda.
@@ -107,3 +110,36 @@ class Proj(CMakePackage):
 
     def setup_dependent_run_environment(self, env, dependent_spec):
         self.setup_run_environment(env)
+
+
+class CMakeBuilder(cmake.CMakeBuilder):
+    def cmake_args(self):
+        args = [
+            self.define("PROJ_LIB", join_path(self.stage.source_path, "nad")),
+            self.define_from_variant("ENABLE_TIFF", "tiff"),
+            self.define_from_variant("ENABLE_CURL", "curl")
+        ]
+        if self.spec.satisfies("@6:"):
+            args.append(self.define("USE_EXTERNAL_GTEST", True))
+        return args
+
+
+class AutotoolsBuilder(autotools.AutotoolsBuilder):
+    def configure_args(self):
+        args = ["PROJ_LIB={0}".format(join_path(self.stage.source_path, "nad"))]
+
+        if self.spec.satisfies("@6:"):
+            args.append("--with-external-gtest")
+
+        if self.spec.satisfies("@7:"):
+            if "+tiff" in self.spec:
+                args.append("--enable-tiff")
+            else:
+                args.append("--disable-tiff")
+
+            if "+curl" in self.spec:
+                args.append("--with-curl=" + self.spec["curl"].prefix.bin.join("curl-config"))
+            else:
+                args.append("--without-curl")
+
+        return args
--- a/var/spack/repos/builtin/packages/proj/package.py
+++ b/var/spack/repos/builtin/packages/proj/package.py
@@ -3,8 +3,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack.package import *
 from spack.build_systems import autotools, cmake
+from spack.package import *
 
 
 class Proj(CMakePackage, AutotoolsPackage):
@@ -117,7 +117,7 @@ class CMakeBuilder(cmake.CMakeBuilder):
         args = [
             self.define("PROJ_LIB", join_path(self.stage.source_path, "nad")),
             self.define_from_variant("ENABLE_TIFF", "tiff"),
-            self.define_from_variant("ENABLE_CURL", "curl")
+            self.define_from_variant("ENABLE_CURL", "curl"),
         ]
         if self.spec.satisfies("@6:"):
             args.append(self.define("USE_EXTERNAL_GTEST", True))
--- a/var/spack/repos/builtin/packages/proj/package.py
+++ b/var/spack/repos/builtin/packages/proj/package.py
@@ -126,7 +126,7 @@ class CMakeBuilder(cmake.CMakeBuilder, Setup):
 
 class AutotoolsBuilder(autotools.AutotoolsBuilder, Setup):
     def configure_args(self):
-        args = ["PROJ_LIB={0}".format(join_path(self.pkg.stage.source_path, "nad"))]
+        args = []
 
         if self.spec.satisfies("@6:"):
             args.append("--with-external-gtest")
@@ -138,7 +138,7 @@ class AutotoolsBuilder(autotools.AutotoolsBuilder, Setup):
                 args.append("--disable-tiff")
 
             if "+curl" in self.spec:
-                args.append("--with-curl=" + self.pkg.spec["curl"].prefix.bin.join("curl-config"))
+                args.append("--with-curl=" + self.spec["curl"].prefix.bin.join("curl-config"))
             else:
                 args.append("--without-curl")
 
--- a/var/spack/repos/builtin/packages/proj/package.py
+++ b/var/spack/repos/builtin/packages/proj/package.py
@@ -91,10 +91,10 @@ class Proj(CMakePackage, AutotoolsPackage):
     depends_on("pkgconfig@0.9.0:", type="build", when="@6: build_system=autotools")
     depends_on("cmake@2.6.0:", type="build", when="build_system=cmake")
 
-    build_system("autotools", conditional("cmake", when="@5.0.0:"), default="autotools")
+    build_system("autotools", conditional("cmake", when="@5.0.0:"), default="cmake")
 
 
-class Setup(object):
+class Setup:
     def setup_run_environment(self, env):
         # PROJ_LIB doesn't need to be set. However, it may be set by conda.
         # If an incompatible version of PROJ is found in PROJ_LIB, it can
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -6,7 +6,7 @@
 from spack.package import *
 
 
-class Proj(AutotoolsPackage):
+class Proj(CMakePackage):
```
adamjstewart: Not all versions support CMake. You'll need to do a multi-build system like with libtiff.

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,30 +83,16 @@ class Proj(AutotoolsPackage):
     )
 
     # https://proj.org/install.html#build-requirements
-    depends_on("pkgconfig@0.9.0:", type="build", when="@6:")
```
adamjstewart, line 86: Need to specify required versions of cmake

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,30 +83,16 @@ class Proj(AutotoolsPackage):
...
-            if "+curl" in self.spec:
-                args.append("--with-curl=" + self.spec["curl"].prefix.bin.join("curl-config"))
-            else:
-                args.append("--without-curl")
-
-        return args
+    def cmake_args(self):
+        return [
+            self.define("PROJ_LIB", join_path(self.stage.source_path, "nad")),
+            self.define_from_variant("ENABLE_TIFF", "tiff"),
```
adamjstewart: Need to add flag for curl

adamjstewart (changes requested):

var/spack/repos/builtin/packages/proj/package.py
johnwparent: Thought I had checked all of them, thanks for catching this!

var/spack/repos/builtin/packages/proj/package.py
johnwparent: I've made the change w.r.t. to the `cmake_minimum_version` for the project but I'm curious, is this a standard convention for CMake packages or is this a more specific requirement to proj? I just ask because I've ported/am porting a number of CMake projects and have yet to notice a consideration for minimum version/consider it myself and I'll need to go and correct that.

adamjstewart: All CMake projects do this and should list the correct dependency range in `package.py`: https://spack.readthedocs.io/en/latest/build_systems/cmakepackage.html#important-files

johnwparent: Is this something we should be enforcing then? A quick grep of the packages directory reveals a not insignificant number of CMake packages that do not do this.

adamjstewart: The majority of people who review PRs and merge them don't know what they're doing lol. This kind of CMake version thing isn't the biggest issue (Spack defaults to the latest version anyway), but if you have an old version installed (maybe as an external package) then Spack will happily use it unless the correct dependency requirements are listed. Basically, there's no reason not to be more specific about the dependency, and it can prevent issues for users too.

johnwparent: I completely agree, it's a very good idea, and I definitely fall into the category of people who this helps (I think I have cmake like 2.8 somewhere on my machine for some test I did once) and I'm always a fan of being more explicit whenever possible, so I have no problem with the policy/suggestion at all, I was more just wondering if this was something we could/should add to the package audits check for cmake packages? Seems worth enforcing IMO as the minimum version is pretty important, and it would take the burden off those reviewers who aren't aware of the suggestion/policy and like you say it's not a huge deal either way and is generally trivial to add.

adamjstewart: Not sure how easy this would be to add to the audit tests. Not all packages declare a minimum version in CMakeLists.txt so we would have to download and grep the source code.

johnwparent: True, it's not mandatory, but I think all the audit test would need to report is that the package recipe does not specify a CMake version constraint for the package, and at that point it doesn't seem unreasonable to ask someone creating the package recipe to know this fairly basic piece of info about the project they're adding to spack/converting to CMake or to find out. Maybe I'm just not familiar enough with other use cases for Spack, but it seems like if you're adding/modifying a package, you're at least able to access that package's CMakeLists.txt, or if you're not familiar with CMake, set an arbitrary very low version. That last scenario isn't really different from the current setup, but at least kind of forces the consideration (which is all that's really important IMO).

johnwparent: This is also something I'd be happy to poke at myself, not trying to suggest more work for someone else lol.

adamjstewart: Not sure what that arbitrary low version would be. I guess I'm not sure how common it is for developers to set this flag.

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,12 +84,49 @@ class Proj(AutotoolsPackage):
...
+    depends_on("cmake@2.6.0:", type="build", when="build_system=cmake")
+    # Builds for Windows via CMake available only
+    # after this version - builds before exist
+    # for Windows but are targeted at VS2005
+    conflicts("platform=windows", when="@:5.0.0")
 
+    build_system("autotools", conditional("cmake", when="@5.0.0:"), default="autotools")
+
+
+class Setup(object):
```
adamjstewart:
```suggestion
class Setup:
```

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,12 +84,49 @@ class Proj(AutotoolsPackage):
...
     depends_on("googletest", when="@6:")
     depends_on("sqlite@3.11:", when="@6:")
     depends_on("libtiff@4.0:", when="@7:+tiff")
     depends_on("curl@7.29.0:", when="@7:+curl")
+    depends_on("pkgconfig@0.9.0:", type="build", when="@6: build_system=autotools")
+    depends_on("cmake@2.6.0:", type="build", when="build_system=cmake")
+    # Builds for Windows via CMake available only
+    # after this version - builds before exist
+    # for Windows but are targeted at VS2005
+    conflicts("platform=windows", when="@:5.0.0")
```
adamjstewart: So to clarify, version 4 and older don't work because the package only supports Autotools, version 5.0.0 doesn't work because it's designed for VS2005, and version 5.0.1+ works?

johnwparent: I was under the impression the upper bound was not inclusive for version constraints, so 5.0.0 does work (not that we have that version in this package) with CMake (support added in 5.0.0 release). Prior to that there are solutions for VS2005 for all versions that I checked (those supported by this recipe).

johnwparent: So the comment is a little unclear, you're right.

adamjstewart: All version ranges in Spack are inclusive, `@:5.0.0` includes `@5.0.0`. The Autotools base class has a conflict for Windows so I don't think you need to add this line.

johnwparent: Sounds good.

johnwparent: Also good to know, thanks.

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,12 +84,45 @@ class Proj(AutotoolsPackage):
...
     # https://proj.org/install.html#build-requirements
-    depends_on("pkgconfig@0.9.0:", type="build", when="@6:")
     depends_on("googletest", when="@6:")
     depends_on("sqlite@3.11:", when="@6:")
     depends_on("libtiff@4.0:", when="@7:+tiff")
     depends_on("curl@7.29.0:", when="@7:+curl")
+    depends_on("pkgconfig@0.9.0:", type="build", when="@6: build_system=autotools")
+    depends_on("cmake@2.6.0:", type="build", when="build_system=cmake")
 
+    build_system("autotools", conditional("cmake", when="@5.0.0:"), default="autotools")
```
adamjstewart: I think we should always default to cmake, there's no good reason to prefer autotools

johnwparent: I've been setting the default to autotools incase the CMake build breaks something for some *nix user unexpectedly, and to just generally keep things for non Windows users fairly consistent so they don't have to think about Windows/related changes

johnwparent: If we don't care about that I'll _happily_ change it (here and a number of other places I do the same thing).

var/spack/repos/builtin/packages/proj/package.py
johnwparent: I think just setting something like `cmake@2` would probably be sufficiently low. I personally have never seen a project fail to set the minimum CMake version, although I'm absolutely sure some projects do. It's generally considered best practice to include it, that being said, assuming best practices are ever followed is not realistic, but in this case I think it's common enough.

var/spack/repos/builtin/packages/proj/package.py
adamjstewart: I would def prefer cmake over autotools, I've found cmake builds to be more reliable at dependency finding, and they also generate cmake-specific config files that get added to the installation prefix that help downstream packages that use them. Proj is built by our CI (torchgeo -> rasterio -> gdal -> proj) so should be heavily tested to make sure these kinds of changes don't break anything. But I would be fine for doing this with all packages in general, not just proj.

johnwparent: Awesome. I may be a liiittle biased, but CMake is absolutely the way to go whenever possible, so I'm very on board with this approach. Thanks for supporting that!
I guess it would also only be changed in develop for now anyway, so on the off chance it did break things in some niche instance, that's not unreasonable.

var/spack/repos/builtin/packages/proj/package.py
johnwparent: We're allowed to use new style class definitions now? Sweet!

var/spack/repos/builtin/packages/proj/package.py
adamjstewart: I think newer versions of CMake started warning when a minimum isn't set, so it's def more common that it used to be.

var/spack/repos/builtin/packages/proj/package.py
adamjstewart: Yes, Python 3.6 is the lowest version of Python we support now!

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,12 +84,45 @@ class Proj(AutotoolsPackage):
...
     depends_on("libtiff@4.0:", when="@7:+tiff")
     depends_on("curl@7.29.0:", when="@7:+curl")
+    depends_on("pkgconfig@0.9.0:", type="build", when="@6: build_system=autotools")
+    depends_on("cmake@2.6.0:", type="build", when="build_system=cmake")
 
+    build_system("autotools", conditional("cmake", when="@5.0.0:"), default="cmake")
+
+
+class Setup:
+    def setup_run_environment(self, env):
```
adamjstewart, line 96: @alalazo can't remember, which of these belong in the package and which in the builder?

johnwparent: If it helps the package audit check explicitly complained about the `setup_dependent_build_environment` method not being in the builder.

Since they all invoke the same method I thought it would be best to have them all in the same place.

adamjstewart: I think `setup_run_environment` in particular has to be in the package, not the builder. Not sure about the rest.

zackgalbreath:
@spackbot run pipeline

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,12 +84,45 @@ class Proj(AutotoolsPackage):
...
+        self.pkg.setup_run_environment(env)
+
+    def setup_dependent_run_environment(self, env, dependent_spec):
+        self.pkg.setup_run_environment(env)
+
+
+class CMakeBuilder(cmake.CMakeBuilder, Setup):
+    def cmake_args(self):
+        args = [
+            self.define("PROJ_LIB", join_path(self.stage.source_path, "nad")),
```
adamjstewart: This doesn't seem to be a valid cmake flag. We could try setting it as an env var in `setup_build_environment` for both Autotools and CMake to see if that works.

var/spack/repos/builtin/packages/proj/package.py
adamjstewart: I think `setup_dependent_run_environment` should also be in the package, not the build. @alalazo is there some documentation or code I can look at to get a full list of what belongs where?

johnwparent: I've been working off of https://spack.readthedocs.io/en/latest/spack.html#spack.builder.Builder and whatever isn't there I'm putting in the package. Just missed the dependent version.

var/spack/repos/builtin/packages/proj/package.py
johnwparent: Whoops, good catch. Looking at the notes for this port, I grepped for that substring in the cmake options and got it, but looking more closely now the string I actually matched was "PROJ_LIB_SUBDIR"... I'll add to the env.

johnwparent: Actually on further inspection we already set that value during both the build and the run in the env. It's set to a different path of course. Seems like this argument in configure is pointing the build to the spack installed resources. I'm sure there must be a corresponding CMake arg, but I'm unsure what this data represents so its difficult to determine which is the correct one. Do you have any further insight there?

adamjstewart: We aren't setting it in `setup_build_environment`

johnwparent: ... 🤦 ... good point. I've made that change.

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -83,14 +84,49 @@ class Proj(AutotoolsPackage):
...
+        ]
+        if self.spec.satisfies("@6:"):
+            args.append(self.define("USE_EXTERNAL_GTEST", True))
+        return args
+
+
+class AutotoolsBuilder(autotools.AutotoolsBuilder, Setup):
     def configure_args(self):
-        args = ["PROJ_LIB={0}".format(join_path(self.stage.source_path, "nad"))]
+        args = ["PROJ_LIB={0}".format(join_path(self.pkg.stage.source_path, "nad"))]
```
adamjstewart: This prob isn't needed anymore with the env var, but we should check

var/spack/repos/builtin/packages/proj/package.py
```diff
@@ -102,22 +138,8 @@ def configure_args(self):
                 args.append("--disable-tiff")
 
             if "+curl" in self.spec:
-                args.append("--with-curl=" + self.spec["curl"].prefix.bin.join("curl-config"))
+                args.append("--with-curl=" + self.pkg.spec["curl"].prefix.bin.join("curl-config"))
```
adamjstewart: Does `self.spec` not work? We're still using that other places in the builder

johnwparent: Which attributes can be referenced from the builders vs which need to be piped through the package is very unclear, and the builder parent classes refer to the spec by `self.pkg.spec` so I erred on the side of "doing what is being done already". I don't actually know if it does or not.

adamjstewart: Ping @alalazo...

var/spack/repos/builtin/packages/proj/package.py
johnwparent: I think you're right, at least from what I can tell from the autotools docs, I'll pull this out.

var/spack/repos/builtin/packages/proj/package.py
johnwparent: Tried it out myself, `self.spec` works from the builder.

adamjstewart (approved):
I think that's everything! We build GDAL and friends in CI so as long as CI passes there shouldn't be any other build issues.

johnwparent:
@spackbot run pipeline

johnwparent:
@spackbot run pipeline

PR #35109, opened 2023-01-23
Expat: to CMake

johnwparent:
Convert Expat to build with CMake to support Windows

Part of #34938

### DIFF
--- a/var/spack/repos/builtin/packages/expat/package.py
+++ b/var/spack/repos/builtin/packages/expat/package.py
@@ -106,11 +106,7 @@ class Expat(CMakePackage):
         description="Use libbsd (for high quality randomness)",
     )
 
-    variant(
-        "shared",
-        default=True,
-        description="Build expat as shared if true, static if false"
-    )
+    variant("shared", default=True, description="Build expat as shared if true, static if false")
 
     depends_on("libbsd", when="@2.2.1:+libbsd")
 
@@ -119,7 +115,10 @@ class Expat(CMakePackage):
         return url.format(version.underscored, version.dotted)
 
     def cmake_args(self):
-        args = [self.define("EXPAT_BUILD_DOCS", False), self.define_from_variant("BUILD_SHARED_LIBS", "shared")]
+        args = [
+            self.define("EXPAT_BUILD_DOCS", False),
+            self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+        ]
 
         if "+libbsd" in self.spec and "@2.2.1:" in self.spec:
             args.append(self.define_from_variant("EXPAT_WITH_LIBBSD", "libbsd"))
--- a/var/spack/repos/builtin/packages/expat/package.py
+++ b/var/spack/repos/builtin/packages/expat/package.py
@@ -8,7 +8,7 @@ import sys
 from spack.package import *
 
 
-class Expat(AutotoolsPackage):
+class Expat(CMakePackage):
     """Expat is an XML parser library written in C."""
 
     homepage = "https://libexpat.github.io/"
@@ -102,7 +102,7 @@ class Expat(AutotoolsPackage):
     # `~libbsd`.
     variant(
         "libbsd",
-        default=sys.platform != "darwin",
+        default=sys.platform != "darwin" and sys.platform != "win32",
         description="Use libbsd (for high quality randomness)",
     )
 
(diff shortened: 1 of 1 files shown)


zackgalbreath:
@spackbot run pipeline

johnwparent:
@spackbot run pipeline

johnwparent:
@spackbot run pipeline

var/spack/repos/builtin/packages/expat/package.py
```diff
@@ -112,12 +112,10 @@ def url_for_version(self, version):
...
-    def configure_args(self):
-        spec = self.spec
-        args = [
-            "--without-docbook",
-            "--enable-static",
-        ]
-        if "+libbsd" in spec and "@2.2.1:" in spec:
-            args.append("--with-libbsd")
+    def cmake_args(self):
+        args = [self.define("EXPAT_BUILD_DOCS", False), self.define("BUILD_SHARED_LIBS", False)]
```
hartwork: `BUILD_SHARED_LIBS` should probably default to `True` for both backwards-compatibility and security.

hartwork (changes requested):

var/spack/repos/builtin/packages/expat/package.py
johnwparent: Honestly it should probably be a variant now. W.r.t. to a default perhaps you can enlighten me to this point as I am not particularly familiar with expat's configure script but to me the previous usage of `--enable-static` would indicate the prior use of static libs no? If we're considering backwards compat at least. IMO I agree, shared is the way to go.

hartwork: Hi @johnwparent , I misread `--enable-static` as `--disable-static`, my bad, sorry.
I just checked: Autotools is building both static and shared by default, so `--enable-static` was superflous previously:
```console
# ./configure --help
[..]
  --enable-shared[=PKGS]  build shared libraries [default=yes]
  --enable-static[=PKGS]  build static libraries [default=yes]
[..]
```
CMake however builds static or shared libraries, not both, to my best knowledge, unless the upstream build system goes extra miles to do that. So if you we need both static and shared in Spack, we need to build twice some way, I guess.

Variants seem to be a lot like Gentoo USE flags but with support for multiple values to co-exist. Maybe then to have the shared-or-static flag be a variant would allow to have both at the same time where needed without building twice from *inside* the package.
So a variant and shared for a default? What do you think?

hartwork: @johnwparent PS: I haven't checked yet, but there are probably other packages around here that also had to answer the `BUILD_SHARED_LIBS` question some way.

johnwparent: Right, unless the upstream project really goes out of their way (and arguably avoids best practice in doing so) CMake will only build static or shared, not both. 
The way I've been solving this as I port packages to use CMake that have been previously building both shared and static at once is to add the variant as discussed, so I'll take that approach. If a consuming package really need both the shared and static libs at once they can specify a dependency on both variants.

hartwork: @johnwparent sounds good to me :+1:

johnwparent:
@spackbot run pipeline

johnwparent:
@spackbot run pipeline

johnwparent:
@spackbot run pipeline

johnwparent:
Gitlab pipelines appear to be failing for reasons unrelated to this PR. Will be running until they pass.

scheibelp:
This effectively renders `cmake~ownlibs` impossible (expat is a cmake dependency). Presumably we'd want to at least offer the choice between Autotools and CMake.

johnwparent:
> This effectively renders `cmake~ownlibs` impossible (expat is a cmake dependency. Presumably we'd want to at least offer the choice between Autotools and CMake.

Good catch, I always forget about CMake's optional external deps. I've added back the autotools builder. 

I think technically users could work around this the way we work around things on Windows, by installing CMake+ownlibs and then using that CMake to drive builds that don't depend on CMake~ownlibs. That being said that's incredibly circuitous and far from ideal.

var/spack/repos/builtin/packages/expat/package.py
```diff
@@ -102,19 +105,36 @@ class Expat(AutotoolsPackage):
     # `~libbsd`.
     variant(
         "libbsd",
-        default=sys.platform != "darwin",
+        default=sys.platform != "darwin" and sys.platform != "win32",
         description="Use libbsd (for high quality randomness)",
     )
 
+    variant("shared", default=True, description="Build expat as shared if true, static if false")
```
scheibelp: I think this should be removed: if I understand correctly, the autotools build makes shared and static libs, and this variant does nothing.

If you think the cmake-based build needs this option, I recommend only allowing the variant when `build_system=cmake`.

scheibelp (changes requested):

var/spack/repos/builtin/packages/expat/package.py
johnwparent: I think removing the option is leaving some utility out (someone may want static over shared some day) so I'll leave it for CMake only. Pushed.

scheibelp (approved):

PR #35111, opened 2023-01-23
HDF5: No specific MPI provider in depends on

johnwparent:
Remove direct statement of dependence on a specific MPI provider for Windows platform.

This is not only fundamentally incorrect w.r.t. modeling, it also precludes us from using IntelMPI on Windows with HDF5, which is obviously not ideal. Thus this PR removes that line and allows Spack's virtual dependency management to provide the MPI.

Part of #34938


byrnHDF (approved):

lrknox (approved):

hyoklee (approved):

alalazo:
@spackbot run pipeline

alalazo (approved):

PR #35112, opened 2023-01-23
Windows: Update MSMPI package

johnwparent:
Proper external detection fixes.

Proper variable references within new Builder class.

Part of #34938

### DIFF
--- a/var/spack/repos/builtin/packages/msmpi/package.py
+++ b/var/spack/repos/builtin/packages/msmpi/package.py
@@ -20,7 +20,7 @@ class Msmpi(Package):
     git = "https://github.com/microsoft/Microsoft-MPI.git"
     tags = ["windows"]
 
-    executable = ["mpiexec"]
+    executables = ["mpiexec"]
 
     version("10.1.1", sha256="63c7da941fc4ffb05a0f97bd54a67968c71f63389a0d162d3182eabba1beab3d")
     version("10.0.0", sha256="cfb53cf53c3cf0d4935ab58be13f013a0f7ccb1189109a5b8eea0fcfdcaef8c1")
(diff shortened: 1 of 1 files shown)


zackgalbreath:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

var/spack/repos/builtin/packages/msmpi/package.py
```diff
@@ -41,29 +41,37 @@ def determine_version(cls, exe):
         ver_str = re.search(r"Microsoft MPI Startup Program \[Version ([0-9.]+)\]", output)
         return Version(ver_str.group(1)) if ver_str else None
 
+    def setup_dependent_package(self, module, dependent_spec):
+        spec = self.spec
+
+        spec.mpicc = spack_cc
```
scheibelp, line 49: Normally these would be assigned to whatever compiler wrappers come with the MPI installation. For `msmpi`, are there no such wrappers?

As it is, ~this is~ any dependent package is being told that the MPI compiler wrappers should just be whatever compiler that dependent package happens to be using (to be clear, that means that if X->msmpi, then mpicc will be the C compiler that is intended for the build of X). If that's the intent, IMO a comment would be useful here making that explicit (and pointing out that this is not typical).

(all the other changes make sense to me)

johnwparent: AFAICT from the docs for building a project with msmpi (and a number of projects I've ported on Windows) msmpi does not have wrappers, you just specify the requisite includes and libs to the compiler/linker manually (so just manual wrappers), which is a lot of what the wrappers do anyway. 

I'll update the comments to reflect this.

scheibelp (approved):

johnwparent:
Thanks @scheibelp!

PR #35116, opened 2023-01-23
add_padding: fix boundary case and text encoding

haampie:
There is a number of issues with how we use padding in Store paths:

1. We should guarantee that trailing characters aren't dir separators.
   This would avoid issues where `os.path.join(padded_path, subdir)` does
   not add an extra dir separator, so that we effectively have a shorter
   prefix than set in config.
2. Special case of (1) is when the input string with trailing dir
   separator dropped is of size padded length - 1. In that case we
   currently do nothing, but that breaks the guarantee that
   `config:install_tree:padded_length` gives you *at least* padded_length
   padding. So, instead we should just create a 1 character subdir.
3. We use `len(string)`, which counts code points instead of code units.
   Since we care about the bytes representation here when relocating,
   it's more correct when`padded_length` counts code units, so that
   relocation works from paths consisting of just single-byte characters
   to paths that contain multibyte characters.
4. We should probably still support UTF-16, since Python is not
   guaranteed to always deal with paths as UTF-8 on Windows, even after
   PEP 529. So, that means dealing with code units instead of bytes
   makes most sense.

This PR fixes those edge case, but is unfortunately also breaking in
three ways.

Breaking change 1: if we previously had a prefix like this with
`padded_length` set to `30`:

```
123456789012345678901234567890|1
/x/__spack_path_placeholder__/|/
                             ^
                             padded length
```

then effectively we would really pad to `29` characters. In this PR, we
*do* pad to `30` characters:

```
123456789012345678901234567890|1
/x/__spack_path_placeholder___|/
                             ^
                             padded length
```

Breaking change 2: similar to 1, if we previously had a prefix like this
with `padded_length` set to 3:

```
123|4
/x |
  ^
  padded length
```

we would stick to a 2-char prefix `/x`, with this PR we use at least 3
by taking `/x/_` instead:

```
123|4
/x/|_
  ^
  padded length
```

Breaking change 3: when a path contains multi-byte characters like
emojis, we previously counted characters, but now we count code units,
so the prefix we use gets shorter. For example, if we previously had
`padded_length` set to 14, we would get:

```
/😬/__spack_pat
```

but now we get 3 characters fewer, since the emoji is 4 units

```
/😬/__spack_
```

### DIFF
--- a/lib/spack/spack/util/path.py
+++ b/lib/spack/spack/util/path.py
@@ -15,6 +15,7 @@ import subprocess
 import sys
 import tempfile
 from datetime import date
+from typing import Callable
 from urllib.parse import urlparse
 
 import llnl.util.tty as tty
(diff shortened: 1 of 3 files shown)


lib/spack/spack/util/path.py
```diff
@@ -320,40 +321,59 @@ def substitute_path_variables(path):
...
+    units are one byte, but unfortunately UTF-8 paths are not strictly enforced on Windows.
+
+    Returns: padded path as string, without trailing separator.
+    """
+
+    # Drop trailing dir seperators
+    path = path.rstrip(os.path.sep)
+
+    # NOTE: the assumption is that "." is one unit
+    bytes_per_unit = len(fsencode("."))
```
becker33, line 340: Since we already assume `len(fsencode("_")) == bytes_per_unit` below, would it make more sense to compute it that way originally instead of making an additional assumption about `"."`?

lib/spack/spack/util/path.py
```diff
@@ -320,40 +321,59 @@ def substitute_path_variables(path):
...
+
+    # How many units is our current path, and how many should we add.
+    units_in_path = len(fsencode(path)) // bytes_per_unit
+    pad_units = length - units_in_path
```
becker33, line 344: Wouldn't `len(path)` be better for getting the current units, since `utf-8` can be variable-width?

```suggestion
    pad_units = length - len(path)
```

lib/spack/spack/util/path.py
```diff
@@ -320,40 +321,59 @@ def substitute_path_variables(path):
...
+    if pad_units <= units_per_sep:
+        return os.path.join(path, "_")
+
+    # Our padding including dir separator is how many units?
+    sep_with_dirname = os.path.sep + SPACK_PATH_PADDING_CHARS
+    bytes_per_sep_with_dirname = len(fsencode(sep_with_dirname))
+
+    # Ensure that there are no multi-unit characters (this way it's valid to truncate on any unit)
+    assert bytes_per_sep_with_dirname == len(sep_with_dirname) * bytes_per_unit
+    units_per_sep_with_dirname = bytes_per_sep_with_dirname // bytes_per_unit
```
becker33, line 361: If we're focused on units instead of bytes, can't we do

```suggestion
    # Our padding including dir separator is how many units?
    sep_with_dirname = os.path.sep + SPACK_PATH_PADDING_CHARS
    units_per_sep_with_dirname = len(sep_with_dirname)
```

since we can truncate on any unit anyway if the measure we're using is units?

becker33 (commented):
Couple questions here about bytes vs units, to go along with our conversation on slack.

lib/spack/spack/util/path.py
haampie: In general `len(path) <= units_in_path`; in utf-8 a code point can be 1 to 4 units, and every unit is one byte. So with the suggestion the effective length in bytes would be longer than prescribed in `padded_length`, which can be problematic when `padded_length` is large and the store dir contains many multibyte chars.

FYI:

```
In [1]: len("😁")
Out[1]: 1

In [2]: len("😁".encode("utf-8"))
Out[2]: 4
```

becker33: Ah got it, I wasn't picking up on `len` reporting code points instead of code units.

haampie: That's the part that makes it a pain ;p

lib/spack/spack/util/path.py
```diff
@@ -320,40 +321,59 @@ def substitute_path_variables(path):
...
+    units_per_sep = len(fsencode(os.path.sep)) // bytes_per_unit
+    if pad_units <= units_per_sep:
+        return os.path.join(path, "_")
+
+    # Our padding including dir separator is how many units?
+    sep_with_dirname = os.path.sep + SPACK_PATH_PADDING_CHARS
+    bytes_per_sep_with_dirname = len(fsencode(sep_with_dirname))
+
+    # Ensure that there are no multi-unit characters (this way it's valid to truncate on any unit)
+    assert bytes_per_sep_with_dirname == len(sep_with_dirname) * bytes_per_unit
```
becker33, line 360: This should come with a more informative error message, if we run into this we need to encourage the user to report it to us to make code changes.

becker33 (changes requested):

haampie:
Actually the boundary case is still not correct for `/x` padded to length 3. If it becomes `/x/_` then there is still no guarantee that something that was padded with length 3 can be moved to length 3. For example when creating a buildcache from `/xy` and installing to `/x`, while keeping padded length to 3

Not sure what's best here. The thing is that this is a real use case for people who dynamically set the store path to some git tag/branch, with padded length globally set to the same value.

Should we then turn `/x` into `/x_` instead? That might be unexpected. Or add a leading / st it becomes `//x`? Might not work on windows

becker33:
The leading `/` won't work on unix either if the build system runs any sort of path normalization.

haampie:
normpath in python preserves double // at the start for unix paths, but it feels too hacky

PR #35117, opened 2023-01-23
Added e4s-cl package

spoutn1k:
Added a package file for the E4S Container Launch tool developed for E4S containers in the ECP.


FrederickDeny (dismissed):

tldahlgren (dismissed):
Confirmed the branch and commits exist.

tldahlgren:
Adding @adamjstewart since this package inherits from `PythonPackage`.

tldahlgren:
We prefer PRs be based on branches off `develop`.  See the docs at https://spack.readthedocs.io/en/latest/contribution_guide.html#git-workflows.

var/spack/repos/builtin/packages/e4s-cl/package.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class E4sCl(PythonPackage):
+    """Container Launcher for E4S containers, facilitating MPI library
+    translations"""
+
+    maintainers = ["spoutn1k", "FrederickDeny"]
+    homepage = "https://e4s-cl.readthedocs.io"
+    url = "https://oaciss.uoregon.edu/e4s/e4s-cl/releases"
+    git = "https://github.com/E4S-Project/e4s-cl"
```
adamjstewart, line 16: Not sure how others feel about this but I usually prefer checksummable URLs over git commits.

var/spack/repos/builtin/packages/e4s-cl/package.py
```diff
@@ -0,0 +1,25 @@
...
+
+    tags = ["e4s"]
+
+    version("master", branch="master")
+    version("1.0.0", commit="410bb2e6601d9b90243a487ad7f7d2dabd8ba04c")
+    version("1.0.1", commit="b2c92993e0c7cb42de07f0f7cc02da3a06816192")
```
adamjstewart: Can you sort newest to oldest?

var/spack/repos/builtin/packages/e4s-cl/package.py
```diff
@@ -0,0 +1,25 @@
...
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
```
adamjstewart, line 27: Missing a bunch of dependencies:
```
termcolor>=1.1.0
PyYAML>=6.0
texttable>=1.6.2
python-ptrace>=0.9.7
pyelftools==0.27
requests>=2.26.0
tinydb==4.5.2
python-sotools==0.1.0
docker==5.0.3
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/e4s-cl/package.py
spoutn1k: I originally had it that way but I figured git commits could prove more reliable than archives manually copied to another server. I can change it if necessary.

var/spack/repos/builtin/packages/e4s-cl/package.py
spoutn1k: Are those dependencies made accessible when installing using `PythonPackage` when listed as spack dependencies ? Just to ensure they are not installed twice.

adamjstewart: Define "made accessible". Spack uses `pip install` to build and install a wheel, but if the dependencies aren't also installed, the package will be unusable. If you `spack load e4s-cl`, it will also load all dependencies so that they are all in `PYTHONPATH`.

spoutn1k: ~Yes and `pip install` also installs those dependencies, so I am asking if listing them via `spack` will make sure they do not get installed twice.~

I tested in a container and indeed, via `spack` the pip install ignores the `install_requires` value from `setup.py`. My previous tests must have picked up my existing environment. I'll push an updated list.

adamjstewart: Yep, we add the `--no-deps` flag to disable dependency installation so that Spack can manage all deps itself.

spoutn1k: This should be fixed !

alalazo:
@spackbot fix style

var/spack/repos/builtin/packages/e4s-cl/package.py
```diff
@@ -0,0 +1,33 @@
...
+    depends_on("py-texttable@1.6.1", type="run")
+    depends_on("py-pyyaml@6.0", type="run")
+    depends_on("py-termcolor@1.1.0", type="run")
+    depends_on("py-pyelftools@0.26", type="run")
+    depends_on("py-python-ptrace@0.9.8", type="run")
+    depends_on("py-python-sotools@0.1.0", type="run")
+    depends_on("py-requests@2.26.0", type="run")
+    depends_on("py-tinydb@4.7.0", type="run")
```
adamjstewart: Most of these versions are incorrect. I only looked at 1.0.1 (the version installed by default), but if the versions required differ between 1.0.0, 1.0.1, and main, you'll need to add `when=` clauses. Dep type should be `type=("build", "run")` for everything other than setuptools. `>=` versions should end with a trailing colon, exact versions should only be required for `==`. Can you sort these in the same order as `requirements/core.txt`? That makes it easier to review.

var/spack/repos/builtin/packages/py-python-sotools/package.py
```diff
@@ -0,0 +1,17 @@
...
+
+class PyPythonSotools(PythonPackage):
+    """python-sotools is a collection of tools to work with ELF shared objects"""
+
+    pypi = "python-sotools/python-sotools-0.1.0.tar.gz"
+
+    version("0.1.0", sha256="39a088f2ca384294e19a96a82d883feb729f0f2e5ae21d9785be357124ec61f2")
+
+    depends_on("py-setuptools", type="build")
+    depends_on("py-pyelftools@0.26", type="run")
```
adamjstewart:
```suggestion
    depends_on("py-pyelftools", type=("build", "run"))
```

adamjstewart (changes requested):

spoutn1k:
You will notice I omitted the `docker` dependency from the package. The module is optional for execution, and looks like it needs to be reworked in `spack`. For version 4.2.1:

https://github.com/docker/docker-py/blob/b22095f7429c492cd7210b6726c73af26e0d302d/requirements.txt#L1-L19

https://github.com/spack/spack/blob/75f1077b4beaffffaaac8b18ef98585f76a983d8/var/spack/repos/builtin/packages/py-docker/package.py#L18-L22

The version listed in `e4s-cl` (5.0.3) has a similar list.

spoutn1k:
@spackbot fix style

var/spack/repos/builtin/packages/e4s-cl/package.py
```diff
@@ -0,0 +1,34 @@
...
+    tags = ["e4s"]
+
+    version("master", branch="master")
+    version("1.0.1", commit="b2c92993e0c7cb42de07f0f7cc02da3a06816192")
+    version("1.0.0", commit="410bb2e6601d9b90243a487ad7f7d2dabd8ba04c")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
+
+    depends_on("py-termcolor@1.1.0", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-termcolor@1.1.0:", type=("build", "run"))
```

var/spack/repos/builtin/packages/e4s-cl/package.py
```diff
@@ -0,0 +1,34 @@
...
+    depends_on("py-pyyaml@6.0:", type=("build", "run"))
+    depends_on("py-texttable@1.6.2:", type=("build", "run"))
+    depends_on("py-python-ptrace@0.9.7:", type=("build", "run"))
+    depends_on("py-pyelftools@0.27", type=("build", "run"))
+    depends_on("py-requests@2.26.0:", type=("build", "run"))
+    depends_on("py-tinydb@4.7.0", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-tinydb@4.5.2", type=("build", "run"))
```

var/spack/repos/builtin/packages/e4s-cl/package.py
```diff
@@ -0,0 +1,34 @@
...
+    depends_on("py-python-sotools@0.1.0", type=("build", "run"))
```
adamjstewart, line 36: I would still add the docker dep, otherwise `pip check` will fail saying a required dep is missing

var/spack/repos/builtin/packages/py-python-sotools/package.py
```diff
@@ -0,0 +1,17 @@
...
+
+class PyPythonSotools(PythonPackage):
+    """python-sotools is a collection of tools to work with ELF shared objects"""
+
+    pypi = "python-sotools/python-sotools-0.1.0.tar.gz"
+
+    version("0.1.0", sha256="39a088f2ca384294e19a96a82d883feb729f0f2e5ae21d9785be357124ec61f2")
+
+    depends_on("py-setuptools", type="build")
+    depends_on("py-pyelftools@0.27", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-pyelftools", type=("build", "run"))
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-python-sotools/package.py
spoutn1k: Why are you removing the version on this particular one ?

var/spack/repos/builtin/packages/e4s-cl/package.py
spoutn1k: Well this PR is about to exponentially grow in size and as I am sure you noticed, packaging is not my day job, so we'll get time to collaborate !

adamjstewart: It's not mine either 😄 

You could also ask the developers not to list docker as a required dep and patch the requirements file if that sounds easier.

spoutn1k: From what I gather, that would require creating a new version and removing the older ones ?

What are the consequences of a `pip check` ? That docker module also lacks a ton of dependencies anyways.

adamjstewart: You don't have to create a new version, just add a patch file: https://spack.readthedocs.io/en/latest/packaging_guide.html#patches. But I would encourage you to also submit this patch upstream so the developers can review it and tell us whether or not docker really is required.

No consequences of `pip check` at the moment. We currently use pip to build all Python packages, and tell it to ignore dependencies. But in the future, I'm planning on switching to build/installer, which have a way of checking to make sure deps are installed without actually downloading them, so this package will break when I do that.

spoutn1k: I am the developer :) This is an MPI utility, and docker will *not* work with most MPI. This dependency is here by popular demand but not useful in production, and is used by one of the modular backends of the tool.

I will try and update the `py-docker` package, and see how annoying it proves to be.

One of the dependencies is `py-enum34`, which on my local clone of `spack` is an empty directory. Is this normal ?

adamjstewart: If you're the developer that makes things simpler. Why not move it to an extra so users can optionally install docker only when wanted?

spoutn1k: That is a good question, I had not found this option when I was looking for it. I applied a patch for the two versions defined and the next one will have `docker` in the extras.

adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

PR #35120, opened 2023-01-24
py-scipy: do not build 1.6 and older with python < 3.10.1

ggouaillardet:
py-scipy 1.6 and older come with pre cython-ized files that use the _PyGen_Send symbol that was removed from python 3.10.0.161, so do not build these old versions with python 3.10.1 and later

### DIFF
--- a/var/spack/repos/builtin/packages/py-scipy/package.py
+++ b/var/spack/repos/builtin/packages/py-scipy/package.py
@@ -100,10 +100,10 @@ class PyScipy(PythonPackage):
     depends_on("python@3.8:3.10", when="@1.8", type=("build", "link", "run"))
     depends_on("python@3.7:3.10", when="@1.7.2:1.7", type=("build", "link", "run"))
     depends_on("python@3.7:3.9", when="@1.6.2:1.7.1", type=("build", "link", "run"))
-    depends_on("python@3.7:", when="@1.6:1.6.1", type=("build", "link", "run"))
-    depends_on("python@3.6:", when="@1.5.0:1.5", type=("build", "link", "run"))
-    depends_on("python@3.5:", when="@1.3:1.4", type=("build", "link", "run"))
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@adamjstewart @rgommers 

After building `py-phonopy` on spack `v0.19`, we tried to run `pytest` and it failed with this error
```
E   ImportError: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-skylake_avx512/gcc-8.4.1/py-scipy-1.6.1-w5clqqr6p6sy7ibgvbrwjyy7qli5wdsi/lib/python3.10/site-packages/scipy/spatial/transform/rotation.cpython-310-x86_64-linux-gnu.so: undefined symbol: _PyGen_Send
```

google pointed to https://github.com/cython/cython/pull/3921

I noted that there is no dependency on `cython` for `py-scipy` 1.6 and older, so I concluded that old `py-scipy` packages come with pre cython-ized files that do not work with python 3.10.0.161 and more recent.
That is why I updated this recipe **not** to build older `py-scipy` with recent `python`.

FWIW, an other approach is to force older `py-scipy` to run `cython`.
I did not do that because
 - I do not know how to do that
 - I am not convinced there is a need to use older `py-scipy`
But if you disagree with the latter, feel free to close this PR and implement a better fix for older `py-scipy` with recent `python`.

rgommers (approved):
This approach is fine with me. There will be other problems besides cythonized files. Older SciPy versions do not specify what the highest Python version is that will work (newer ones do, via an upper bound in `python_requires` metadata), but it's easy to see from the produced wheels on PyPI. E.g., for 1.6.0 the supported Python version range is 3.7-3.9: https://pypi.org/project/scipy/1.6.0/#files

rgommers:
The one Gitlab failure is a timeout on building TensorFlow, so this should be good to go in.

adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

PR #35124, opened 2023-01-24
py-lazy: add 1.4 and 1.5

heerener:


var/spack/repos/builtin/packages/py-lazy/package.py
```diff
@@ -11,6 +11,8 @@ class PyLazy(PythonPackage):
 
     pypi = "lazy/lazy-1.2.zip"
```
adamjstewart, line 12: They switched to .tar.gz, did you actually test to make sure this installs? You'll need to add a url_for_version function

adamjstewart (changes requested):

adamjstewart (approved):

PR #35126, opened 2023-01-24
py-nptyping, py-typish: add package

heerener:


var/spack/repos/builtin/packages/py-nptyping/package.py
```diff
@@ -0,0 +1,31 @@
...
+from spack.package import *
+
+
+class PyNptyping(PythonPackage):
+    """Type hints for numpy"""
+    homepage = "https://github.com/ramonhagenaars/nptyping"
+    pypi = "nptyping/nptyping-2.4.1.tar.gz"
+
+    version("2.4.1", sha256="57ba684ee5fc5eb681ee04270ee94adb879e4372ce6b640defa08ace8e1df295")
+    version(
```
adamjstewart: Is there any reason to add these ancient versions that don't have sdists on PyPI?

var/spack/repos/builtin/packages/py-nptyping/package.py
```diff
@@ -0,0 +1,31 @@
...
+        url="https://github.com/ramonhagenaars/nptyping/archive/v1.4.1.tar.gz",
+    )
+    version(
+        "1.0.1",
+        sha256="a00e672bfdaddc99aa6b25dd1ae89d7d58d2b76e8ad099bd69577bac2598589f",
+        url="https://github.com/ramonhagenaars/nptyping/archive/v1.0.1.tar.gz",
+    )
+
+    depends_on("py-setuptools", type="build")
+    depends_on("python@3.5:", type=("build", "run"))
```
adamjstewart: 3.7 in the latest release. But Spack only supports 3.7+, so you could just remove this line

var/spack/repos/builtin/packages/py-nptyping/package.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyNptyping(PythonPackage):
+    """Type hints for numpy"""
+    homepage = "https://github.com/ramonhagenaars/nptyping"
+    pypi = "nptyping/nptyping-2.4.1.tar.gz"
```
adamjstewart: The PyPI tarball does not seem to include the `requirements.txt` file, does this package even build?

var/spack/repos/builtin/packages/py-nptyping/package.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyNptyping(PythonPackage):
```
adamjstewart, line 9: This package seems to be unrelated to py-typish, was this accidentally included in the wrong PR?

var/spack/repos/builtin/packages/py-typish/package.py
```diff
@@ -0,0 +1,23 @@
...
+
+from spack.package import *
+
+
+class PyTypish(PythonPackage):
+    """Python library for type checking
+
+    * Functions for thorough checks on types
+    * Instance checks considering generics
+    * Typesafe Duck-typing
```
adamjstewart: `spack info py-typish` removes newline characters so the formatting of this will be messed up. I would either simplify the docstring or add commas/semi-colons.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-nptyping/package.py
heerener: I've included it here because it depends on py-typish. Will update the PR title to reflect this.

var/spack/repos/builtin/packages/py-nptyping/package.py
heerener: I'm mostly upstreaming changes that have been lying around here for a long time, some of these ancient versions are, apparently, still being used here. I've added the latest version specifically for this PR (although I admit I should've tested properly).

On a related note: as I have more that I'd like to upstream, is it better to make a bunch of smaller PRs with packages that relate to one another, or reduce the PR-spam a little by making a bigger one that contains unrelated packages?

var/spack/repos/builtin/packages/py-nptyping/package.py
heerener: There's even [an issue](https://github.com/ramonhagenaars/nptyping/issues/98) about it - I've linked to the github tarball for now (and installed it, so that one does work). Will revisit once the issue is fixed.

var/spack/repos/builtin/packages/py-nptyping/package.py
adamjstewart: We try to follow a one-package-per-PR system, although it's fine to combine 2+ packages if the changes to one package depend on another.

var/spack/repos/builtin/packages/py-nptyping/package.py
adamjstewart: Can we remove `pypi` and set `url` globally for now? Keep the comment though, that's useful to refer to if it ever gets fixed.

adamjstewart (approved):

PR #35129, opened 2023-01-24
py-pyspark: new versions

heerener:


var/spack/repos/builtin/packages/py-py4j/package.py
```diff
@@ -13,6 +13,17 @@ class PyPy4j(PythonPackage):
     homepage = "https://www.py4j.org/"
     pypi = "py4j/py4j-0.10.4.zip"
 
+    version(
```
adamjstewart: Can you add a `url_for_version` so we don't need to keep copying the URL each time? Let me know if you need help.

var/spack/repos/builtin/packages/py-pyspark/package.py
```diff
@@ -12,15 +12,30 @@ class PyPyspark(PythonPackage):
...
     version("2.4.4", sha256="13655eb113b8cf5f3f85b24fd92f86c4668a723723bd68949d028fa0df2cf694")
     version("2.3.2", sha256="7fb3b4fe47edb0fb78cecec37e0f2a728590f17ef6a49eae55141a7a374c07c8")
     version("2.3.0", sha256="0b3536910e154c36a94239f0ba0a201f476aadc72006409e5787198ffd01986e")
 
+    depends_on("python@3.7:", when="@3.3.0:", type=("build", "run"))
+    depends_on("python@3.6:", when="@3:3.2.3", type=("build", "run"))
     depends_on("python@2.7:3.7", when="@:2", type=("build", "run"))
     depends_on("python@2.7:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
+    depends_on("py-py4j@0.10.9.5", when="@3.3.0", type=("build", "run"))
```
adamjstewart: Missing an entry for 3.3.1, also not sorted by version. I would just use version ranges instead of enumerating every version.

adamjstewart (changes requested):

adamjstewart (approved):

PR #35143, opened 2023-01-24
Windows: Fix spack.bat handling of env commands

danlipsa:
This PR enables the successful execution of the spack binary cache tutorial on Windows.
It assumes gnupg and file are available ( I installed them with choco).
It fixes additional double quote issues in spack.cmd

### DIFF
--- a/bin/spack.bat
+++ b/bin/spack.bat
@@ -96,20 +96,23 @@ set str_subcommand=%_sp_subcommand:"='%
 set str_flags=%_sp_flags:"='%
 set str_args=%_sp_args:"='%
 if "%str_subcommand%"=="ECHO is off." (set "_sp_subcommand=")
+if "%str_subcommand%"=="ECHO is on." (set "_sp_subcommand=")
 if "%str_flags%"=="ECHO is off." (set "_sp_flags=")
+if "%str_flags%"=="ECHO is on." (set "_sp_flags=")
 if "%str_args%"=="ECHO is off." (set "_sp_args=")
+if "%str_args%"=="ECHO is on." (set "_sp_args=")
 del subcmd
 del flags
 del args
 
 :: Filter out some commands. For any others, just run the command.
-if "%_sp_subcommand%" == "cd" (
+if %_sp_subcommand% == "cd" (
     goto :case_cd
-) else if "%_sp_subcommand%" == "env" (
+) else if %_sp_subcommand% == "env" (
     goto :case_env
-) else if "%_sp_subcommand%" == "load" (
+) else if %_sp_subcommand% == "load" (
     goto :case_load
-) else if "%_sp_subcommand%" == "unload" (
+) else if %_sp_subcommand% == "unload" (
     goto :case_load
 ) else (
     goto :default_case
@@ -141,21 +144,22 @@ goto :end_switch
 
 :case_env
 :: If no args or args contain --bat or -h/--help: just execute.
+set args_no_quote=%_sp_args:"=%
 if NOT defined _sp_args (
     goto :default_case
-)else if NOT "%_sp_args%"=="%_sp_args:--help=%" (
+)else if NOT "%args_no_quote%"=="%args_no_quote:--help=%" (
     goto :default_case
-) else if NOT "%_sp_args%"=="%_sp_args: -h=%" (
+) else if NOT "%args_no_quote%"=="%args_no_quote: -h=%" (
     goto :default_case
-) else if NOT "%_sp_args%"=="%_sp_args:--bat=%" (
+) else if NOT "%args_no_quote%"=="%args_no_quote:--bat=%" (
     goto :default_case
-) else if NOT "%_sp_args%"=="%_sp_args:deactivate=%" (
+) else if NOT "%args_no_quote%"=="%args_no_quote:deactivate=%" (
     for /f "tokens=* USEBACKQ" %%I in (
-        `call python "%spack%" %_sp_flags% env deactivate --bat %_sp_args:deactivate=%`
+        `call python %spack% %_sp_flags% env deactivate --bat %args_no_quote:deactivate=%`
     ) do %%I
-) else if NOT "%_sp_args%"=="%_sp_args:activate=%" (
+) else if NOT "%args_no_quote%"=="%args_no_quote:activate=%" (
     for /f "tokens=* USEBACKQ" %%I in (
-        `call python "%spack%" %_sp_flags% env activate --bat %_sp_args:activate=%`
+        `python %spack% %_sp_flags% env activate --bat %args_no_quote:activate=%`
     ) do %%I
 ) else (
     goto :default_case
--- a/bin/spack.bat
+++ b/bin/spack.bat
@@ -83,6 +83,16 @@ if defined _sp_flags (
         exit /B 0
     )
 )
+if not defined _sp_subcommand (
+   if not defined _sp_args (
+      if not defined _sp_flags (
+         python "%spack%" --help
+         exit /B 0
+      )
+   )
+)
+
+
 :: pass parsed variables outside of local scope. Need to do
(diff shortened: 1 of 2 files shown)


danlipsa:
@johnwparent Please review and suggest other reviewers. Thanks!

johnwparent:
@danlipsa Can you add a short description to the PR so other reviewers without the context of our internal meetings can understand these changes?

lib/spack/llnl/util/filesystem.py
```diff
@@ -119,10 +119,7 @@ def file_command(*args):
     try:
         file_cmd = which("file", required=True)
     except CommandNotFoundError as e:
-        if is_windows:
-            raise CommandNotFoundError("`file` utility is not available on Windows")
-        else:
-            raise e
+        raise e
```
johnwparent: This entire try/catch can be removed, it was only added to implement the additional error reporting on Windows.

lib/spack/llnl/util/filesystem.py
```diff
@@ -133,7 +130,11 @@ def _get_mime_type():
     """Generate method to call `file` system command to aquire mime type
     for a specified path
     """
-    return file_command("-b", "-h", "--mime-type")
+    if is_windows:
+        # -h option (no-dereference) does not exist in Windows
+        return file_command("-b", "--mime-type")
```
johnwparent, line 132: Is this always true or only for the version you installed via choco? I.e. will this be true if we have a spack derived version?

johnwparent:
This looks ok after a first pass. I'll need to checkout locally to verify the bat script changes.

lib/spack/llnl/util/filesystem.py
danlipsa: True. I removed that.

lib/spack/llnl/util/filesystem.py
danlipsa: I installed file using scoop (besides choco) and it also has the missing option
 ```
    -h, --no-dereference
             option causes symlinks not to be followed (on systems that sup‐
             port symbolic links).  This is the default if the environment
             variable POSIXLY_CORRECT is not defined.
```
I looked at the file package and it has references to MINGW libraries. So this is probably what we need to use if we want to build this in spack. I am not sure why that option is missing - maybe it has to do with how links are accessed in Unix vs Windows.

Comment (danlipsa) [reply lib/spack/llnl/util/filesystem.py:132]: Those options are only defined if S_IFLNK. I assume this is something set by configure which fails in Windows.

**Comment by danlipsa** [comment]
@johnwparent ping. Anybody else should review this? Thanks!

**Comment by johnwparent** [comment]
> @johnwparent ping. Anybody else should review this? Thanks!

Peter has already been tagged but is out this week.

**Comment by johnwparent** [comment]
@danlipsa This PR fixes pre-req issues for build cache but does not actually touch anything directly related to build caching code. Can we rename the PR/description to indicate that this PR is not actually touching Spack core, just so our Spack reviewers aren't confused without any context.

File: bin/spack.bat [hunk 3]
```diff
@@ -141,21 +144,22 @@ goto :end_switch
 
 :case_env
 :: If no args or args contain --bat or -h/--help: just execute.
+set args_no_quote=%_sp_args:"=%
 if NOT defined _sp_args (
     goto :default_case
-)else if NOT "%_sp_args%"=="%_sp_args:--help=%" (
+)else if NOT "%args_no_quote%"=="%args_no_quote:--help=%" (
```
scheibelp: I'm having trouble seeing why `args_no_quote` is needed for this particular check: couldn't `--help` be removed from `_sp_args` directly and compare that with `_sp_args`?

bin/spack.bat
```diff
@@ -96,20 +96,23 @@ set str_subcommand=%_sp_subcommand:"='%
...
+if "%str_flags%"=="ECHO is on." (set "_sp_flags=")
 if "%str_args%"=="ECHO is off." (set "_sp_args=")
+if "%str_args%"=="ECHO is on." (set "_sp_args=")
 del subcmd
 del flags
 del args
 
 :: Filter out some commands. For any others, just run the command.
-if "%_sp_subcommand%" == "cd" (
+if %_sp_subcommand% == "cd" (
```
scheibelp, line 116: Is this to avoid issues if `%_sp_subcommand%` contains a space?

scheibelp (commented):
I have a couple questions (mostly stemming from my ignorance about .bat scripts)

bin/spack.bat
danlipsa: All the issues I fixed here resulted from the fact that _sp_subcommand and _sp_args were already between double quotes. So for instance the first test lower would compare ""env"" with "env" that would fail.

bin/spack.bat
danlipsa: See my previous comment.

bin/spack.bat
scheibelp: Sorry did we add the "first" set of quotes ourselves? Or is this for cases where args are quoted explicitly by the user on the command line?

Also, given that, how does the following work:

```
set str_subcommand=%_sp_subcommand:"='%
if "%str_subcommand%"=="ECHO is on." (set "_sp_subcommand=")
```

It looks like "if" would never be true (if as you say _sp_subcommand contains " characters, then `str_subcommand` would have `'` characters and it looks like we'd be comparing `"'ECHO is on'" == "ECHO is on."`)

bin/spack.bat
scheibelp: In this case the comparison would be "weird" but not strictly incorrect even if the quotes were kept, right? In that case you'd be comparing

```
""... --help"" == ""...""
```

or if there was no `--help` option

```
""..."" == ""...""
```

(i.e. the statement would be true when we expect and false when we expect)

danlipsa: You are right, this part works with both args_no_quote and with _sp_args. The part that only works with args_no_quote is 
`python %spack% %_sp_flags% env activate --bat %args_no_quote:activate=%`
With _sp_args this becomes:
`python C:\Users\dan.lipsa\projects\spack\bin\spack env activate --bat "" "c:\Users\dan.lipsa\projects\spack\cache-env"`
which results in an error:
`Error: unrecognized arguments: c:\Users\dan.lipsa\projects\spack\cache-env`

danlipsa: We could get rid of args_no_quote everywhere else if you prefer.

bin/spack.bat
danlipsa: The ECHO is on/off is different. That comes without double quotes from doing echo on an undefined variable on line 88.

bin/spack.bat
danlipsa: Another way to fix this is to use
`python %spack% %_sp_flags% env activate --bat %_sp_args:"activate"=%`
Note I used _sp_args but I remove "activate" including quotes.

scheibelp: I think all the if checks should be consistent regarding testing the value of `_sp_args` or `args_no_quote`: they can all do it one way or all do it the other.

bin/spack.bat
scheibelp: If I understand, in that case defining `str_subcommand` is unnecessary and 

```
if "%str_subcommand%"=="ECHO is off." (set "_sp_subcommand="
```

for example could be just testing `sp_subcommand`

bin/spack.bat
danlipsa: Yes, this is why I used args_no_quote everywhere. So I'll leave it as is then.

bin/spack.bat
danlipsa: That is correct. Good catch! I pushed another commit that gets rid of those variables.

scheibelp (approved):

scheibelp:
Thanks!

PR #35149, opened 2023-01-24
py-cligj: add v0.7.2

adamjstewart:
0.5.0 doesn't support click 8, discovered using `pip check`.


tldahlgren (approved):
Confirmed new/revised sha256.  Rest LGTM

PR #35152, opened 2023-01-24
py-scikit-learn: add v1.2.1

adamjstewart:
https://scikit-learn.org/stable/whats_new/v1.2.html#version-1-2-1


tldahlgren (approved):
Confirmed sha256.

PR #35157, opened 2023-01-25
Add Orca version 5.0.3-f.1

awvwgk:


snehring:
@awvwgk where can I find this version's archive?

awvwgk:
That version is distributed by [faccts](https://faccts.de).

snehring:
Guessing that's not freely available. I'll just assume the hash is correct and this installs as expected. It may be worthwhile to note somewhere that the -f version is distributed from faccts.

snehring (approved):

tldahlgren (approved):
LGTM

PR #35161, opened 2023-01-25
Print file summary on checksum validation failure

haampie:
Currently we print "sha256 checksum failed for [file]. Expected X but
got Y".

This PR extends that message with file size and contents info:

"... but got Y. File size = 123456 bytes. Contents = b'abc...def'"

That way we can immediately see if the file was downloaded only
partially, or if we downloaded a text page instead of a binary, etc.

This is mostly relevant for CI where the filesystem is not accessible
on failure.

For example:

```
==> Installing libxc-6.0.0-tzq4pnlyej64ltbzmhlnegkuxymvoob6
==> No binary for libxc-6.0.0-tzq4pnlyej64ltbzmhlnegkuxymvoob6 found: installing from source
==> Fetching https://www.tddft.org/programs/libxc/down.php?file=6.0.0/libxc-6.0.0.tar.gz
==> Error: ChecksumError: sha256 checksum failed for /tmp/harmen/spack-stage/spack-stage-libxc-6.0.0-tzq4pnlyej64ltbzmhlnegkuxymvoob6/libxc-6.0.0.tar.gz
    Expected d2ca205a762200dfba2e6c9e8ca2061aaddc6b7cf42048859fe717a7aa07de7c but got c2ca205a762200dfba2e6c9e8ca2061aaddc6b7cf42048859fe717a7aa07de7c. File size = 62940412 bytes. Contents = b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x02\x03\xec\\{w\xda\xc6...\xbe\xfd\xfb\x8f\xff\xf7\xff\x07V\xf5P[\x00h~\x0f'
```

### DIFF
--- a/lib/spack/spack/test/fetch_strategy.py
+++ b/lib/spack/spack/test/fetch_strategy.py
@@ -5,7 +5,7 @@
 
 import pytest
 
-from spack.fetch_strategy import from_url_scheme
+import spack.fetch_strategy as fetch_strategy
 
 
 def test_fetchstrategy_bad_url_scheme():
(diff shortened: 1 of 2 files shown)


lib/spack/spack/test/fetch_strategy.py
```diff
@@ -5,12 +5,22 @@
 
 import pytest
 
-from spack.fetch_strategy import from_url_scheme
+import spack.fetch_strategy as fetch_strategy
```
alalazo:
```suggestion
from spack import fetch_strategy
```

alalazo (dismissed):

alalazo (approved):

PR #35164, opened 2023-01-25
trilinos@develop: conflict w/ cxxstd=11,14

eugeneswalker:
It looks like `trilinos@develop` will no longer work with `cxxstd=11` or `cxxstd=14`

```
  >> 4    CMake Error at CMakeLists.txt:130 (MESSAGE):
     5      CMAKE_CXX_STANDARD=14 is not in the allowed set (17|20|23)
     6    
     7    
     8    -- Configuring incomplete, errors occurred!
```

@jwillenbring @keitat @kuberry @psakievich @sethrj @wspear

### DIFF
--- a/var/spack/repos/builtin/packages/trilinos/package.py
+++ b/var/spack/repos/builtin/packages/trilinos/package.py
@@ -331,6 +331,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     )
     conflicts("+adios2", when="@:12.14.1")
     conflicts("cxxstd=11", when="@13.2:")
+    conflicts("cxxstd=14", when="@14:")
     conflicts("cxxstd=17", when="@:12")
     conflicts("cxxstd=11", when="+wrapper ^cuda@6.5.14")
     conflicts("cxxstd=14", when="+wrapper ^cuda@6.5.14:8.0.61")


var/spack/repos/builtin/packages/trilinos/package.py
```diff
@@ -217,6 +217,9 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
 
     # ###################### Conflicts ##########################
 
+    conflicts("cxxstd=11", when="@develop")
+    conflicts("cxxstd=14", when="@develop")
+
```
sethrj: The existing `cxxstd` conflicts are near line 340. The `=11` is redundant with the existing requirement for `@13.2:`. We probably will be able to write the conflict as `when="@14:"` since that will include develop and be true for the next trilinos release.

sethrj (changes requested):

var/spack/repos/builtin/packages/trilinos/package.py
eugeneswalker: Does it look OK now? Just pushed a change.

sethrj (approved):

psakievich (approved):

PR #35170, opened 2023-01-26
Solver: allow constraints on reuse parameter

becker33:
With this PR, users can specify constraints for Spack reuse.

E.g.
```
concretizer:
  reuse: ['%gcc']
```
This would allow Spack to reuse only packages that were built with gcc.

These constraints are combined by a logical "or" operator, so
```
concretizer:
  reuse: ['%gcc', '%clang']
  ```
  would allow Spack to reuse any package built with gcc or clang.

Note: A package with a non-reusable dependency is not reusable either.

TODO:
- [ ] docs

### DIFF
--- a/lib/spack/spack/schema/concretizer.py
+++ b/lib/spack/spack/schema/concretizer.py
@@ -14,7 +14,9 @@ properties = {
         "type": "object",
         "additionalProperties": False,
         "properties": {
-            "reuse": {"type": "boolean"},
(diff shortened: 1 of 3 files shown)


alalazo:
This went in sometimes ago, with a different syntax. See https://spack.readthedocs.io/en/latest/build_settings.html#reuse-already-installed-packages

PR #35173, opened 2023-01-26
buildcache create: avoid prefix copy

haampie:
Extracted from #33970, avoid a prefix copy when we're not mutating files before
pushing to a buildcache.

On a fast NVMe disk this saves about 5%, but should be more on slow disks.


alalazo (approved):
LGTM reading the code.

PR #35174, opened 2023-01-26
python allow builds with zlib-ng

haampie:
This would help a lot in our own docker images.

Python with `zlib` takes about `4m10s` to create a tarball of cuda + deps, with `zlib-ng` this is only `2m1s`, this is still single-threaded.

### DIFF
--- a/var/spack/repos/builtin/packages/python/package.py
+++ b/var/spack/repos/builtin/packages/python/package.py
@@ -139,6 +139,7 @@ class Python(Package):
     variant("dbm", default=True, description="Build dbm module")
     variant("nis", default=False, description="Build nis module")
     variant("zlib", default=True, description="Build zlib module")
+    variant("zlib_ng", default=False, when="+zlib", description="Use zlib-ng instead of zlib")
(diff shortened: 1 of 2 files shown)


haampie:
@spackbot run pipeline

adamjstewart:
I think it would be better to create a virtual package (maybe libz?) and depend on that instead. This could also be used for other packages.

haampie:
I think it's going to be hard to build zlib-ng without zlib

adamjstewart:
I see. I know @alalazo was working on allowing the same package to appear in the DAG multiple times. Does this also apply to virtual dependencies? Can some packages in the DAG resolve to `zlib-ng` while others resolve to `zlib`? How would a user specify this in Spack's spec syntax? This is complicated by the fact that Python has both direct and indirect build and link deps on zlib. I think we would have to break the DAG at build deps so that `cmake` and deps are built with `zlib` and `python` and deps are built with `zlib-ng`.

zackgalbreath:
@spackbot run pipeline

haampie:
Closing in favor of #37372

PR #35183, opened 2023-01-26
Tweak more parallel build job count for spack env depfile jobserver

Jordan474:
Fix `qt`, `r`, and `py-horovod` to avoid overriding parallel job count when running under a Makefile jobserver (`spack env depfile`).

The API is not pretty, but for now it still helps a lot, especially with qt.

### DIFF
--- a/var/spack/repos/builtin/packages/py-horovod/package.py
+++ b/var/spack/repos/builtin/packages/py-horovod/package.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 
+from spack.build_environment import get_effective_jobs
 from spack.package import *
 
 
@@ -220,7 +221,10 @@ class PyHorovod(PythonPackage, CudaPackage):
         env.set("PKG_CONFIG_EXECUTABLE", self.spec["pkgconfig"].prefix.bin.join("pkg-config"))
         if "cmake" in self.spec:
             env.set("HOROVOD_CMAKE", self.spec["cmake"].command.path)
-        env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
+
+        jobs = get_effective_jobs(make_jobs, parallel=self.parallel, supports_jobserver=True)
+        if jobs is not None:
+            env.append_flags("MAKEFLAGS", "-j{0}".format(jobs))
 
         # Frameworks
         if "frameworks=tensorflow" in self.spec:
--- a/var/spack/repos/builtin/packages/qt/package.py
+++ b/var/spack/repos/builtin/packages/qt/package.py
@@ -9,6 +9,7 @@ import sys
 
 import llnl.util.tty as tty
 
+from spack.build_environment import get_effective_jobs
 from spack.operating_systems.linux_distro import kernel_version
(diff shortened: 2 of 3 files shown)


adamjstewart:
Need more context to know if this is the right thing to do. Hundreds of packages use `make_jobs`, are you saying this is incorrect? Should we do this in `lib/spack/spack/build_environment.py` so that `make_jobs` is always set correctly?

Jordan474:
This is a continuation of https://github.com/spack/spack/pull/32862 .

Yes, There is a lot of packages that use `make_jobs`, but they usually don't support jobserver anyway, so we can't do anything for them (easily).

So I'm only looking to "fix" jobserver support for packages that by-pass the `make()` function (that supports jobserver implicitly).

(Personally, I use `spack env depfile etc.` with `SPACK_INSTALL_FLAGS="-j2"` so packages that don't  support jobserver have a fallback on `make_jobs = 2`, so they don't spawn too many processes. I've had very very good results compared to `spack install`.)

About the API, it definitively need improvement, but I haven't thought of a good alternative... So the verbose-but-explicit api seemed ok for now.

If we change `make_jobs` value to `get_effective_jobs`, then it would be `None` under jobserver, I think this could be error-prone.

And also there is `SPACK_NO_PARALLEL_MAKE`...

adamjstewart:
Gotcha, I'll let @haampie @trws @tgamblin review this one

Jordan474:
I got an idea for a simpler package API, if you're interested:
- We force `make_jobs` to `1` when  `SPACK_NO_PARALLEL_MAKE`: **it changes its meaning a bit** and now applies to all packages instead just make-based ones.
- It allows us to replace the `get_effective_jobs()` function by a simple `jobserver_enabled` module variable for explicit jobserver support

(`make()` and `ninja()` unchanged: still support jobserver internally implicitly)

So the only change in `qt` would be:

```diff
     def setup_build_environment(self, env):
-        env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
+        if not jobserver_enabled:
+            env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
         if self.version >= Version("5.11"):
```

and existing `get_effective_jobs` in `cmake` could be simplified:

```diff
         if not sys.platform == "win32":
             args.append("--prefix={0}".format(self.prefix))
 
-            jobs = spack.build_environment.get_effective_jobs(
-                make_jobs,
-                parallel=self.parallel,
-                supports_jobserver=self.generator.supports_jobserver,
-            )
-            if jobs is not None:
-                args.append("--parallel={0}".format(jobs))
+            if not jobserver_enabled:
+                args.append("--parallel={0}".format(make_jobs))
 
             if "+ownlibs" in spec:
                 # Build and link to the CMake-provided third-party libraries
```

haampie:
That looks great! The current way of doing things is rather ugly, replacing that with just two lines look good!

Jordan474:
Here is my proposal for a slightly simpler API for jobserver support: https://github.com/spack/spack/pull/35199

It would supersede this PR, if accepted.

zackgalbreath:
@spackbot run pipeline

Jordan474:
Closing in favor of #35644

PR #32478, opened 2022-09-01
Allow extra_instructions at beginning of container build and final stages

climbfuji:
## Description
This PR adds `pre_build` and `pre_final` extra instructions to `Dockerfile` and `singularity.def`. Credits go to @kgerheiser for suggesting a first version of these changes to me.

As described in https://github.com/spack/spack/issues/26036, it's sometimes necessary to have extra instructions before the build stage. For example, when adding new `apt` sources and installing additional packages.

Likewise, I found it necessary to have extra instructions before the final stage, for example to set environment variables such as `TZ` to avoid linux configuration tools like `debconf` waiting for user input.

## Testing

I tested this extensively with `Dockerfile` for both `pre_build` and `pre_final`, and recently also with `singularity.def`.

Note. I have not updated the documentation yet. I would like to get feedback from the developers first before I go and update the documentation.

## Issues

Fixes https://github.com/spack/spack/issues/26036

PR #35187, opened 2023-01-26
Improve Dockerfile recipe generation

### LINKED ISSUE #26036: Allow extra_instructions at beginning of container stages
### Summary

Some sites may need to add SSL certificates or other things to the beginning of the stages generated by "spack containerize". There are currently "extra_instructions:build" and "extra_instructions:final" options available to add extra commands at the end of each build stage. This request is asking for similar functionality for the beginning of each build stage.

For example:

```
FROM spack/centos7:latest as builder

# Beginning extra_instructions
ADD site_specific_cert.crt /etc/pki/ca-trust/source/anchors/site_specific_cert.crt
RUN update-ca-trust

# Auto-generated content by "spack containerize"

# Ending extra_instructions
RUN yum install -y libgomp

ENTRYPOINT ["/bin/bash", "--rcfile", "/etc/profile", "-l"]
```

### Rationale

Some sites require a site-specific SSL certificate to be installed in order to reach upstream repos outside of the site. Without these certs in place, the "docker build ." fails.

### Description

Add extra_instructions before the Dockerfile contents generated by Spack. Such a feature already exists for extra_instructions after the auto-generated content.

### Additional information

Using spack developer brach:

$ spack --version
0.16.2-4259-5d23638fdc

### General information

alalazo:
fixes #26036
closes #32478

Modifications:
- [x] Update default image to Ubuntu 22.04 (previously was still Ubuntu 18.04)
- [x] Optionally use depfiles to install the environment within the container
- [x] Allow extending Dockerfile Jinja2 template
- [x]  Allow extending Singularity definition file Jinja2 template
- [x] Deprecate previous options to add extra instructions


alalazo:
@climbfuji I found some time today to implement the idea suggested in #32478. I wrote some initial documentation for the feature. Can you check if it fits your use case?

share/spack/templates/container/Dockerfile
```diff
@@ -19,7 +20,7 @@ RUN mkdir {{ paths.environment }} \
 {{ manifest }} > {{ paths.environment }}/spack.yaml
 
 # Install the software, remove unnecessary deps
-RUN cd {{ paths.environment }} && spack env activate . && spack install --fail-fast && spack gc -y
+RUN cd {{ paths.environment }} && spack env activate . && spack concretize && spack env depfile -o Makefile && make -j && spack gc -y
```
haampie: I wouldn't use `env depfile` by default, cause (a) it will mark everything as explicitly installed so that `gc` doesn't help (which is a bug we need to fix), and (b) for gmake < 4.4 building `perl` currently swallows all the tokens, so that after perl is installed, things continue single-threaded.

haampie: Actually (a) was fixed. (b) is still an issue. Btw, use -j$(nproc)

lib/spack/docs/containers.rst
```diff
@@ -444,6 +444,119 @@ attribute:
...
+
+The template extension can override two blocks, named ``build_stage`` and ``final_stage``, similarly to
+the example below:
+
+.. code-block::
+   :emphasize-lines: 3,8
+
+   {% extends "container/Dockerfile" %}
+   {% block build_stage %}
+   ENV CFLAGS="-finline-functions"
```
haampie: I don't think this is a great example since Spack clears the variable in a build environment, and also the flag itself is questionable

lib/spack/docs/containers.rst
```diff
@@ -444,6 +444,119 @@ attribute:
...
+   &&   echo "  concretizer:" \
+   &&   echo "    unify: true" \
+   &&   echo "  config:" \
+   &&   echo "    template_dirs:" \
+   &&   echo "    - /tmp/environment/templates" \
+   &&   echo "    install_tree: /opt/software" \
+   &&   echo "  view: /opt/view") > /opt/spack-environment/spack.yaml
+
+   # Install the software, remove unnecessary deps
+   RUN cd /opt/spack-environment && spack env activate . && spack concretize && spack env depfile -o Makefile && make -j && spack gc -y
```
haampie: I would not use `depfile` by default

haampie (changes requested):

lib/spack/docs/containers.rst
alalazo: I put there a sample `echo` command instead. What people should take is that they have the ability to write custom instructions in this part of the template :slightly_smiling_face:

lib/spack/docs/containers.rst
alalazo: Added an option to use `depfile`s, default it to false.

haampie: That is nice 👍

haampie: Can you still turn `-j` into `-j "$(nproc)"`, without a value it is unbounded

lib/spack/docs/containers.rst
```diff
@@ -444,6 +444,120 @@ attribute:
...
+       awk -F: '{print $1}' | xargs strip -s
+
+   # Modifications to the environment that are necessary to run
+   RUN cd /opt/spack-environment && \
+       spack env activate --sh -d . >> /etc/profile.d/z10_spack_environment.sh
+
+   # Bare OS image to run the installed executables
+   FROM ubuntu:22.04
+
+   COPY --from=builder /opt/spack-environment /opt/spack-environment
```
haampie, line 549: Would it be easier to have one additional parent dir and copy that?

share/spack/templates/container/singularity.def
```diff
@@ -20,10 +21,14 @@ EOF
 
   # Install all the required software
   . /opt/spack/share/spack/setup-env.sh
-  spack env activate .
-  spack install --fail-fast
+  spack -e . concretize
+{% if depfile %}
+  spack -e . env depfile > Makefile
```
haampie:
```suggestion
  spack -e . env depfile -o Makefile
```

lib/spack/docs/containers.rst
haampie: ah, it's just the docs that aren't updated

climbfuji: I've read that because of the way Docker works it's recommended to use no more than two threads. Not sure if this is (or was and is still) true, though.

lib/spack/docs/containers.rst
climbfuji: Indeed! This is the reason I expanded the existing code in spack, and @alalazo suggested a better (--> this) approach. The flexibility to execute commands before and after the spack build, as well as before and after the copying to the final container, is very important.

lib/spack/docs/containers.rst
alalazo: Updated the docs too

lib/spack/docs/containers.rst
alalazo: Maybe, can we move that to the the next PR?

haampie: Yeah, I think it will be required soon since view logic will change

haampie (approved):

lib/spack/docs/containers.rst
haampie: Let me point that out in the relevant PR, maybe it can be done as part of it.

PR #35188, opened 2023-01-26
py-protobuf: various updates

adamjstewart:
A few changes:

- [x] Add 3.8.0 just so for-loop becomes simpler
- [x] Make sure for-loop covers 3.0 versions
- [x] Remove deprecated versions
- [x] Enable cpp by default

The last point is for performance reasons. Apparently the cpp implementation is much faster and should be used by default. Also, I noticed that Spack was installing `protobuf` and `py-protobuf~cpp` at different versions that weren't compatible, which was causing issues at run-time: https://github.com/lanpa/tensorboardX/issues/682. This doesn't prevent that, but it does make it less likely.


adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

var/spack/repos/builtin/packages/py-protobuf/package.py
```diff
@@ -85,7 +70,7 @@ class PyProtobuf(PythonPackage):
     for ver in list(range(21, 22)):
         depends_on("protobuf@3." + str(ver), when="+cpp @4." + str(ver))
     # Handle the 3.x series releases
-    for ver in list(range(1, 8)) + list(range(9, 21)):
```
tldahlgren, line 88: Why include `1`, `2`, `13`, and `14` when they aren't in the version's list for `py-protobuf`?

tldahlgren (approved):
Confirmed the two new sha256.  Just have a question about the `protobuf` dependency loop.

var/spack/repos/builtin/packages/py-protobuf/package.py
adamjstewart: Just to simplify the for-loop.

PR #35192, opened 2023-01-26
Add freesasa package

RMeli:


zackgalbreath:
@spackbot run pipeline

RMeli:
@spackbot run pipeline

var/spack/repos/builtin/packages/freesasa/package.py
```diff
@@ -0,0 +1,46 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Freesasa(AutotoolsPackage):
+    """C-library for calculating Solvent Accessible Surface Areas"""
+
+    homepage = "http://freesasa.github.io"
+    url = "https://github.com/mittinatten/freesasa/archive/refs/tags/2.1.2.tar.gz"
```
alalazo: Can we use [this artifact](https://github.com/mittinatten/freesasa/releases/download/2.1.2/freesasa-2.1.2.zip) to have a stable shasum to check?

var/spack/repos/builtin/packages/freesasa/package.py
```diff
@@ -0,0 +1,46 @@
...
+    git = "https://github.com/mittinatten/freesasa.git"
+    maintainers = [
+        "RMeli",
+    ]
```
alalazo:
```suggestion
    maintainers("RMeli")
```

var/spack/repos/builtin/packages/freesasa/package.py
```diff
@@ -0,0 +1,46 @@
...
+        autoreconf("--install", "--verbose", "--force")
+
+    def configure_args(self):
+        args = []
+        if "~json" in self.spec:
+            args.append("--disable-json")
+        if "~xml" in self.spec:
+            args.append("--disable-xml")
+        if "~threads" in self.spec:
+            args.append("--disable-threads")
```
alalazo: Would this:
```suggestion
        args.extend(self.enable_or_disable("json"))
        args.extend(self.enable_or_disable("xml"))
        args.extend(self.enable_or_disable("threads"))
```
work too?

alalazo (changes requested):

var/spack/repos/builtin/packages/freesasa/package.py
RMeli: Is there an advantage in using the artefact instead of the `Source code (tar.gz)`? I've found instances (in other projects) where such artifacts were missing, while `Source code (tar.gz)` is always there by default.

alalazo: Yes, that should give more stability to the sha256sum we use. The `.tar.gz` is generated on demand, so any upstream change from GitHub may impact the hash.  See https://github.com/orgs/community/discussions/46034

alalazo: Obviously, if there is no artifact, using the `tar.gz` is fine, but if there's a viable artifact, that should be preferred.

RMeli:
```suggestion
    url = "https://github.com/mittinatten/freesasa/releases/download/2.1.2/freesasa-2.1.2.zip"
```


I was not aware of this issue, thank you for pointing it out.

var/spack/repos/builtin/packages/freesasa/package.py
```diff
@@ -0,0 +1,41 @@
...
+
+class Freesasa(AutotoolsPackage):
+    """C-library for calculating Solvent Accessible Surface Areas"""
+
+    homepage = "http://freesasa.github.io"
+    url = "https://github.com/mittinatten/freesasa/releases/download/2.1.2/freesasa-2.1.2.zip"
+    git = "https://github.com/mittinatten/freesasa.git"
+    maintainers("RMeli")
+
+    version("2.1.2", tag="2.1.2", branch="master", submodules=True)
```
alalazo: Can we add a sha256sum here for the version (instead of a floating tag)?

alalazo (approved):

PR #35195, opened 2023-01-27
Ecp hdf5 vol

kwryankrattiger:
Reviving #33393 now that HDF5 1.14 is released and ParaView has been patched for new HDF5

### DIFF
--- a/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
+++ b/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
@@ -7,7 +7,9 @@ from spack.package import *
 
 
 class Hdf5VolAsync(CMakePackage):
-    """This package enables asynchronous IO in HDF5."""
+    """This package enables asynchronous IO in HDF5.  WARNING: Please refer to the
+    documentation.  This package will be available in HDF5_PLUGIN_PATH, but in order to
+    use it the consumer must set the HDF5_VOL_CONNECTOR environment variable."""
 
     homepage = "https://hdf5-vol-async.readthedocs.io"
     git = "https://github.com/hpc-io/vol-async.git"
@@ -27,15 +29,17 @@ class Hdf5VolAsync(CMakePackage):
     variant("memcpy", default=False, description="Enable buffer copy for dataset write")
 
     depends_on("mpi")
-    depends_on("argobots@main")
+    depends_on("argobots@1.1:")
     depends_on("hdf5 +mpi +threadsafe")
     depends_on("hdf5@1.13.0:1.13.2", when="@:1.3")
     depends_on("hdf5@1.13.3:", when="@1.4:")
 
+    # Enforce that MPI_THREAD_MULTIPLE is available.
+    depends_on("openmpi +thread_multiple", when="^openmpi")
+    depends_on("mvapich2 threads=multiple", when="^mvapich2")
+
     def setup_run_environment(self, env):
-        env.set("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
-        vol_connector = "async under_vol=0;under_info=[]"
-        env.set("HDF5_VOL_CONNECTOR", vol_connector)
+        env.prepend_path("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
         env.set("MPICH_MAX_THREAD_SAFETY", "multiple")
 
     def cmake_args(self):
--- a/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
+++ b/var/spack/repos/builtin/packages/hdf5-vol-async/package.py
@@ -7,9 +7,14 @@ from spack.package import *
 
 
 class Hdf5VolAsync(CMakePackage):
-    """This package enables asynchronous IO in HDF5.  WARNING: Please refer to the
-    documentation.  This package will be available in HDF5_PLUGIN_PATH, but in order to
-    use it the consumer must set the HDF5_VOL_CONNECTOR environment variable."""
+    """This package enables asynchronous IO in HDF5.
+
+    Usage: Set the HDF5_VOL_CONNECTOR environment variable to enable this adaptor, i.e.
+
+        $export HDF5_VOL_CONNECTOR="async under_vol=0;under_info={}"
+
+    ref. https://hdf5-vol-async.readthedocs.io/en/latest/gettingstarted.html#set-environmental-variables
+    """
 
     homepage = "https://hdf5-vol-async.readthedocs.io"
     git = "https://github.com/hpc-io/vol-async.git"
@@ -34,7 +39,7 @@ class Hdf5VolAsync(CMakePackage):
     depends_on("hdf5@1.13.0:1.13.2", when="@:1.3")
     depends_on("hdf5@1.13.3:", when="@1.4:")
 
-    # Enforce that MPI_THREAD_MULTIPLE is available.
+    # Require MPI_THREAD_MULTIPLE.
     depends_on("openmpi +thread_multiple", when="^openmpi")
     depends_on("mvapich2 threads=multiple", when="^mvapich2")
 
--- a/var/spack/repos/builtin/packages/hdf5-vol-log/package.py
+++ b/var/spack/repos/builtin/packages/hdf5-vol-log/package.py
@@ -22,11 +22,15 @@ class Hdf5VolLog(AutotoolsPackage):
     version("1.2.0", tag="logvol.1.2.0")
     version("1.1.0", tag="logvol.1.1.0")
 
-    depends_on("hdf5@1.13.2:")
+    depends_on("hdf5@1.14.0:", when="@1.4.0:")
+    depends_on("hdf5@1.13.2", when="@:1.3.0")
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool", type="build")
     depends_on("m4", type="build")
+
+    def setup_run_environment(self, env):
+        env.prepend_path("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
 
     def configure_args(self):
         args = []
--- a/var/spack/repos/builtin/packages/hdf5/package.py
+++ b/var/spack/repos/builtin/packages/hdf5/package.py
@@ -39,6 +39,8 @@ class Hdf5(CMakePackage):
     version("develop-1.8", branch="hdf5_1_8")
 
     # Odd versions are considered experimental releases
+    # Note: These are still required to build some VOL adapters, but even releases should be
+    #       preferred.
     version("1.13.3", sha256="83c7c06671f975cee6944b0b217f95005faa55f79ea5532cf4ac268989866af4")
     version("1.13.2", sha256="01643fa5b37dba7be7c4db6bbf3c5d07adf5c1fa17dbfaaa632a279b1b2f06da")
 
--- a/share/spack/gitlab/cloud_pipelines/stacks/data-vis-sdk/spack.yaml
+++ b/share/spack/gitlab/cloud_pipelines/stacks/data-vis-sdk/spack.yaml
@@ -52,6 +52,7 @@ spack:
     - matrix:
       - [$sdk_base_spec]
       - [$^paraview_specs]
+      - - ^hdf5@1.14 # Non-VisIt can build HDF5 1.14
     # Test ParaView builds with differnt GL backends
     - matrix:
       - [$sdk_base_spec]
--- a/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
+++ b/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
@@ -116,8 +116,8 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
     # When building with VisIt and CinemaSci in the same SDK environment there is a conflict
     # in the build dependency of py-setuptools which prevents building a py-h5py that is
     # compatible with 'hdf5@1.14:'. Until there is a version of VisIt with an updated VTK or Spack
-    # allows the concretization of multiple versions of the same build only dependency concretization
-    # with VisIt and Cinema variants will not allow building VOLs.
+    # allows the concretization of multiple versions of the same build only dependency
+    # concretization with VisIt and Cinema variants will not allow building VOLs.
     dav_sdk_depends_on("hdf5@1.12: +shared+mpi", when="+hdf5", propagate=["fortran"])
 
     # HDF5 VOL Adapters require hdf5@1.14:
--- a/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
+++ b/var/spack/repos/builtin/packages/ecp-data-vis-sdk/package.py
@@ -112,19 +112,31 @@ class EcpDataVisSdk(BundlePackage, CudaPackage, ROCmPackage):
 
     dav_sdk_depends_on("faodel+shared+mpi network=libfabric", when="+faodel", propagate=["hdf5"])
 
+    # HDF5 1.14 is a soft requirement for the ECP Data and Vis SDK.
+    # When building with VisIt and CinemaSci in the same SDK environment there is a conflict
+    # in the build dependency of py-setuptools which prevents building a py-h5py that is
+    # compatible with 'hdf5@1.14:'. Until there is a version of VisIt with an updated VTK or Spack
+    # allows the concretization of multiple versions of the same build only dependency concretization
+    # with VisIt and Cinema variants will not allow building VOLs.
     dav_sdk_depends_on("hdf5@1.12: +shared+mpi", when="+hdf5", propagate=["fortran"])
+
+    # HDF5 VOL Adapters require hdf5@1.14:
+
     # hdf5-vfd-gds needs cuda@11.7.1 or later, only enable when 11.7.1+ available.
-    depends_on("hdf5-vfd-gds@1.0.2:", when="+cuda+hdf5^cuda@11.7.1:")
+    depends_on("hdf5-vfd-gds@1.0.2:", when="+cuda+hdf5 ^cuda@11.7.1: ^hdf5@1.14:")
     for cuda_arch in cuda_arch_variants:
         depends_on(
             "hdf5-vfd-gds@1.0.2: {0}".format(cuda_arch),
-            when="+cuda+hdf5 {0} ^cuda@11.7.1:".format(cuda_arch),
+            when="+cuda+hdf5 {0} ^cuda@11.7.1: ^hdf5@1.14:".format(cuda_arch),
         )
     conflicts("~cuda", when="^hdf5-vfd-gds@1.0.2:")
     conflicts("~hdf5", when="^hdf5-vfd-gds@1.0.2:")
-    dav_sdk_depends_on("hdf5-vol-async", when="+hdf5")
-    dav_sdk_depends_on("hdf5-vol-cache", when="+hdf5")
-    dav_sdk_depends_on("hdf5-vol-log", when="+hdf5")
+    conflicts("~hdf5", when="^hdf5-vol-async")
+    conflicts("~hdf5", when="^hdf5-vol-cache")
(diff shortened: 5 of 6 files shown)


kwryankrattiger:
Draft until HDF5 1.14 is merged.

Sorry the for last rebase, lots of stuff in there that was not intended.

kwryankrattiger:
@lrknox @hyoklee Could you guys give this one last pass for review?

hyoklee (dismissed):

var/spack/repos/builtin/packages/hdf5/package.py
```diff
@@ -47,6 +47,11 @@ class Hdf5(CMakePackage):
 
     # Odd versions are considered experimental releases
     version("1.13.3", sha256="83c7c06671f975cee6944b0b217f95005faa55f79ea5532cf4ac268989866af4")
+    # NOTE: retain version 1.13.2 until:
+    # 1. hdf5-vol-async, hdf5-vol-cache, and hdf5-vol-log are ready to build off
+    #    hdf5@1.13.3+ (VOL struct underwent breaking changes).
```
lrknox: Probably should be "...ready to build off hdf5@1.14.0+

lrknox (dismissed):

var/spack/repos/builtin/packages/hdf5-vol-log/package.py
```diff
@@ -21,12 +21,15 @@ class Hdf5VolLog(AutotoolsPackage):
     version("1.2.0", tag="logvol.1.2.0")
     version("1.1.0", tag="logvol.1.1.0")
 
-    depends_on("hdf5@1.13.2:")
+    depends_on("hdf5@1.13:1.13.2", when="@:1.3.0")
```
lrknox: Needs a new version as soon as possible that depends_on("hdf5@1.14:" ...)

var/spack/repos/builtin/packages/hdf5/package.py
kwryankrattiger: I was going to ask if this was still needed for the VOL adapters. I would prefer to delete this entirely if possible.

I want to say we had a discussion that all of the vols were ready for 1.14, but I don't trust my memory.

var/spack/repos/builtin/packages/hdf5-vol-log/package.py
kwryankrattiger: I was going to ask if this could just use 1.14 now. It looks like somewhere between the original work by @svenevs and this update that the log vol changed it's constraint so maybe it is agnostic to the breaking changes in 1.13.3?

lrknox: No, it's not agnostic.  There is a new version of https://github.com/DataLib-ECP/vol-log-based tagged 3 weeks ago that needs to be added to the hdf5-vol-log package.  The new version should be built with hdf5@1.14.0:, the three older versions with hdf5@1.13.2.  These are the changes to hdf5-vol-log/package.py:
plus    version("1.4.0", tag="logvol.1.4.0")

minus  depends_on("hdf5@1.13.2:") 
plus     depends_on("hdf5@1.14.0:", when="@1.4.0")
plus     depends_on("hdf5@1.13.2", when="@:1.3.0")

I can create a PR for those changes unless you would rather add them to this one.  Sorry, '+' and '-' became bullets.

kwryankrattiger: I can add them here if they are ready. If I do would you mind if I removed all of the versions that used 1.13 and dropped the hdf5@1.13 versions as well? Or do we want to keep the hdf5@1.13:1.13 versions around for a little longer?

lrknox: If they aren't needed for e4s, dropping the hdf5@1.13 versions should be fine.

lrknox: I tested the changes - they should be ready.

var/spack/repos/builtin/packages/hdf5/package.py
lrknox: All are ready except hdf5-vol-external-passthrough.  I don't think that one is of great importance to anyone outside of hdf5.  I can try to update it this afternoon, so you can wait for the update, or go ahead without it.  As I recall we kept hdf51.13.2 so e4s tests would pass - probably Paraview in particular.

kwryankrattiger: I don't think that gets built via any of the spack CI stacks, so it should be okay to follow-up with that.

var/spack/repos/builtin/packages/hdf5-vol-log/package.py
kwryankrattiger: The 1.13 versions are still needed for h5bench, so I am going to leave them there for now.

kwryankrattiger:
@spackbot run pipeline

kwryankrattiger:
@spackbot run pipeline

kwryankrattiger:
@alalazo The only thing stopping this PR is CI being extremely reliably flakey when downloading and installing dependencies from the mirror. I really don't understand why this is happening, Any ideas? I thought it might be key related, but I thought all of that was fixed now.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -24,15 +26,17 @@ class Hdf5VolAsync(CMakePackage):
...
     depends_on("hdf5@1.13.3:", when="@1.4:")
 
+    # Enforce that MPI_THREAD_MULTIPLE is available.
+    depends_on("openmpi +thread_multiple", when="^openmpi")
+    depends_on("mvapich2 threads=multiple", when="^mvapich2")
+
     def setup_run_environment(self, env):
-        env.set("HDF5_PLUGIN_PATH", self.spec.prefix.lib)
-        vol_connector = "async under_vol=0;under_info=[]"
-        env.set("HDF5_VOL_CONNECTOR", vol_connector)
```
tldahlgren, line 38: Maybe I missed this, but why is adding the warning in the docstring better than removing setting `HDF5_VOL_CONNECTOR` when setting up the run environment?

tldahlgren (commented):
Confirmed the new sha256 (for `py-h5py`) though have a question about a change.

tldahlgren:
@adamjstewart Did you want to review the `py-h5py` changes?

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
kwryankrattiger: `HDF5_VOL_CONNECTOR` is an application/runtime choice, setting it for the runtime environment is too heavy handed. It is possible to have multiple VOLs activated in the same environment or package as well as potentially overwriting user configurations silently.

var/spack/repos/builtin/packages/py-h5py/package.py
```diff
@@ -60,8 +61,9 @@ class PyH5py(PythonPackage):
```
adamjstewart: build deps need to be updated:
```toml
[build-system]
requires = [
    "Cython ~=0.29",
    "oldest-supported-numpy",
    "pkgconfig",
    "setuptools >=61",
]
```

kwryankrattiger: Looks like setuptools was the only one that needed the update. Thanks!

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
tldahlgren: Hmm.  So having a "default" setting is too heavy handed?  

What about stand-alone testing?  If/when stand-alone tests are added to the package someone will have to be sure to set the environment accordingly.  Setting up the runs includes Spack automatically calling `setup_run_environment`.

var/spack/repos/builtin/packages/py-h5py/package.py
```diff
@@ -48,6 +49,7 @@ class PyH5py(PythonPackage):
     depends_on("py-cython@0.29.15:", type=("build"), when="@3: ^python@3.9.0:")
```
adamjstewart: Cython also needs to be updated. The latest version requires `@0.29:0`

var/spack/repos/builtin/packages/py-h5py/package.py
```diff
@@ -48,6 +49,7 @@ class PyH5py(PythonPackage):
...
     depends_on("py-pkgconfig", type="build")
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@61:", type="build", when="@3:")
```
adamjstewart:
```suggestion
    depends_on("py-setuptools@61:", type="build", when="@3.8:")
```

adamjstewart (dismissed):

var/spack/repos/builtin/packages/py-h5py/package.py
kwryankrattiger: Isn't that already expressed sufficiently here between the `python` and `py-cython` constraints?

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
kwryankrattiger: I could see the benefit for the stand-alone testing to have a helper function that provided the default `HDF5_VOL_CONNECTOR` string.

Not sure if this is idiomatic Spack or not..

```python
env.set("HDF5_VOL_CONNECTOR", self.spec["hdf5_vol_async"].package.default_vol_connector)
```

But even in stand-alone testing, the vol connector is a user configuration, so it should be intentionally set by the application using it so it is clear that that test is using the Async VOL and not the Cache VOL (which would also be set when this is loaded, not sure on ordering) or no VOL for one test and with a VOL for a different test.

var/spack/repos/builtin/packages/py-h5py/package.py
adamjstewart: The current constraints don't contain the upper bound, and the lower bound is different than it was in the previous release.

We may want to encode the correct Python support versions in Cython, I can do that in a different PR.

kwryankrattiger: It looks like the lower bound is already constrained to the `0.29` version from previous releases, unless I am completely misreading this that should be okay still. The rest of the lower bound games are just caveats on python version, which like you said should be fixed in another PR which is great.

I don't know what `py-cython@3:` is, but I will take your word that it is not? a compatible update from 0.29 and so it should be constrained out.

adamjstewart: Yeah, the newer version doesn't have the same Python-version-specific constraints, not sure if we should remove that for this newest version.

Yes, Cython 3 is very much not backwards compatible, so a lot of projects are slowing starting to add upper bounds.

var/spack/repos/builtin/packages/py-h5py/package.py
```diff
@@ -43,11 +44,12 @@ class PyH5py(PythonPackage):
 
     # Build dependencies
     depends_on("py-cython@0.23:", type="build", when="@:2")
```
adamjstewart:
```suggestion
    depends_on("py-cython@0.23:0", type="build", when="@:2")
```
Let's prevent this for ancient versions too.

var/spack/repos/builtin/packages/py-h5py/package.py
kwryankrattiger: I am going to leave them, I feel like the benefit is not worth the added complexity of trying to remove them right now. This update really just needs to go in.

That said, now I have another issue with concretization to figure out.

var/spack/repos/builtin/packages/py-h5py/package.py
kwryankrattiger: This check breaks concretization and I am not sure where the weak link is. In testing, the constraint doesn't seem to be a real requirement.

var/spack/repos/builtin/packages/py-h5py/package.py
```diff
@@ -42,12 +43,13 @@ class PyH5py(PythonPackage):
...
-    depends_on("py-cython@0.29:", type=("build"), when="@3: ^python@:3.7")
-    depends_on("py-cython@0.29.14:", type=("build"), when="@3: ^python@3.8.0:3.8")
-    depends_on("py-cython@0.29.15:", type=("build"), when="@3: ^python@3.9.0:")
+    depends_on("py-cython@0.23:0", type="build", when="@:2")
+    depends_on("py-cython@0.29:0", type=("build"), when="@3: ^python@:3.7")
+    depends_on("py-cython@0.29.14:0", type=("build"), when="@3: ^python@3.8.0:3.8")
+    depends_on("py-cython@0.29.15:0", type=("build"), when="@3: ^python@3.9.0:")
     depends_on("py-pkgconfig", type="build")
     depends_on("py-setuptools", type="build")
+    # depends_on("py-setuptools@61:", type="build", when="@3.8:")
```
adamjstewart: Why is this commented out? This should be right.

kwryankrattiger: If this exists, then py-h5py@3.8 is not usable. I think it just makes the selection of setuptools too heavy. I have dug through packages trying to find the path the conflicts to no avail.

Here is the pipeline that builds py-h5py, it uses `py-setuptools@63`.
https://gitlab.spack.io/spack/spack/-/jobs/5843681

adamjstewart: It seems to concretize just fine with 61+, what do you mean it isn't usable?

kwryankrattiger: In the full `ecp-data-vis-sdk` there is a conflict somewhere. For me, that is unusable. I don't know if it is a real conflict or if the solver is just not demoting some package and failing.

adamjstewart: @alalazo hitting more issues with setuptools pinning. Should we split setuptools into dozens of different packages based on version, or are we getting close to allowing multiple copies of build deps?

adamjstewart: This conflict only affects py-h5py 3.8, right? It's still possible to concretize ecp-data-vis-sdk with py-h5py 3.7?

kwryankrattiger: No, py-h5py 3.8 is the first version to support HDF5 1.14.0 which is a new requirement for the ECP Data and Vis SDK to support VOL adapters which allow HDF5 to use the Async IO, and more importantly the DAOS IO adaptors. DAOS is the optimized IO for Aurora using Intel's Optane memory, so it will be a pretty big deal there.

My current plan, barring this py-setuptools build dep issue getting fixed, is to add some icky logic into the dependencies so I can at least get what I want in reduced SDK builds and kick the can down the road on producing VOL support in HDF5 for VisIt and py-cinemasci as a Spack core limitation. It is kind of unfortunate.

adamjstewart: It may be worth going through all packages in your DAG that require older setuptools and either 1) adding a new version that supports newer setuptools, or 2) reaching out to the developers and asking them if they can support newer setuptools in the next release. I'm not sure how long it will take before we can properly model multiple versions of the same package in the DAG.

kwryankrattiger:
@adamjstewart I dropped the py-h5py from this PR in favor of the other one #35960 , so this work won't build until that goes in.

var/spack/repos/builtin/packages/py-h5py/package.py
kwryankrattiger: The core of the problem is VisIt, so long as it depends on VTK 8 this is going to be a problem. Everything else is either unrestricted or up to date.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -7,7 +7,9 @@
 
 
 class Hdf5VolAsync(CMakePackage):
-    """This package enables asynchronous IO in HDF5."""
+    """This package enables asynchronous IO in HDF5.  WARNING: Please refer to the
+    documentation.  This package will be available in HDF5_PLUGIN_PATH, but in order to
+    use it the consumer must set the HDF5_VOL_CONNECTOR environment variable."""
```
hyoklee: Please simplify the comment block:

"USAGE: Set the following environment variable [1]." 

  `$export HDF5_VOL_CONNECTOR="async under_vol=0;under_info={}"`

[1] https://hdf5-vol-async.readthedocs.io/en/latest/gettingstarted.html#set-environmental-variables
"

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -27,15 +29,17 @@ class Hdf5VolAsync(CMakePackage):
     variant("memcpy", default=False, description="Enable buffer copy for dataset write")
 
     depends_on("mpi")
-    depends_on("argobots@main")
+    depends_on("argobots@1.1:")
     depends_on("hdf5 +mpi +threadsafe")
     depends_on("hdf5@1.13.0:1.13.2", when="@:1.3")
     depends_on("hdf5@1.13.3:", when="@1.4:")
 
+    # Enforce that MPI_THREAD_MULTIPLE is available.
```
hyoklee: `# Enforce MPI_THREAD_MULTIPLE.`

var/spack/repos/builtin/packages/hdf5/package.py
```diff
@@ -39,6 +39,8 @@ class Hdf5(CMakePackage):
     version("develop-1.8", branch="hdf5_1_8")
 
     # Odd versions are considered experimental releases
+    # Note: These are still required to build some VOL adapters, but even releases should be
```
hyoklee, line 42: Since 1.14.0 is out, 1.13.x versions should be removed as well as this comment.

hyoklee (changes requested):

var/spack/repos/builtin/packages/hdf5/package.py
kwryankrattiger: I think that is true, but I don't want to do it in the PR. It will require removing some of the VOL versions that need `hdf5@:1.13.2` as well and I would rather the VOL maintainers mad that change since they will know best which versions work with 1.14 and which do not. Once those are taken care of I think it is safe to  drop all of the 1.13.x versions.

hyoklee: I think updating all VOL packages are right thing to do.

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -7,9 +7,14 @@
 
 
 class Hdf5VolAsync(CMakePackage):
-    """This package enables asynchronous IO in HDF5.  WARNING: Please refer to the
-    documentation.  This package will be available in HDF5_PLUGIN_PATH, but in order to
-    use it the consumer must set the HDF5_VOL_CONNECTOR environment variable."""
+    """This package enables asynchronous IO in HDF5.
+
+    Usage: Set the HDF5_VOL_CONNECTOR environment variable to enable this adaptor, ie.
```
hyoklee: ie. -> i.e.,

var/spack/repos/builtin/packages/hdf5-vol-async/package.py
```diff
@@ -34,7 +39,7 @@ class Hdf5VolAsync(CMakePackage):
     depends_on("hdf5@1.13.0:1.13.2", when="@:1.3")
     depends_on("hdf5@1.13.3:", when="@1.4:")
 
-    # Enforce that MPI_THREAD_MULTIPLE is available.
+    # Require MPI_THREAD_MULTIPLE
```
hyoklee: Please add period (.) at the end of the sentence.

hyoklee (changes requested):

hyoklee (approved):

kwryankrattiger:
@tldahlgren could you give this the last approval nudge?

jeanbez (approved):

tldahlgren (approved):

PR #35199, opened 2023-01-27
Simpler package API for jobserver support

Jordan474:
Replace the `get_effective_jobs()` mechanism by something a bit simpler: a `should_support_jobserver` variable in package modules.

Also fix `qt`, `r`, and `py-horovod` to support jobserver.

Supersedes #35183 if accepted.

Compared to a more "factorized" API (say a `make_j_arg` or something for example), I think `should_support_jobserver` is more "intentional" and more flexible for future yet unknown tools with jobserver support.

Quick summary:

Package that use `make()` or `ninja()` are OK as-is. Others should mind jobserver support: e.g. don't add a `-j` to make or ninja if `should_support_jobserver` is `True`.

### DIFF
--- a/var/spack/repos/builtin/packages/cmake/package.py
+++ b/var/spack/repos/builtin/packages/cmake/package.py
@@ -9,6 +9,7 @@ import sys
 
 import spack.build_environment
 from spack.package import *
+from spack.util.environment import env_flag
 
 is_windows = sys.platform == "win32"
 
@@ -327,12 +328,9 @@ class Cmake(Package):
         if not sys.platform == "win32":
             args.append("--prefix={0}".format(self.prefix))
 
-            jobs = spack.build_environment.get_effective_jobs(
-                make_jobs,
-                parallel=self.parallel,
-                supports_jobserver=self.generator.supports_jobserver,
-            )
-            if jobs is not None:
+            jobs = 1 if env_flag(spack.build_environment.SPACK_NO_PARALLEL_MAKE) else make_jobs
+            jobserver = jobs > 1 and should_support_jobserver and self.generator.supports_jobserver
+            if not jobserver:
                 args.append("--parallel={0}".format(jobs))
 
             if "+ownlibs" in spec:
--- a/var/spack/repos/builtin/packages/py-horovod/package.py
+++ b/var/spack/repos/builtin/packages/py-horovod/package.py
@@ -220,7 +220,9 @@ class PyHorovod(PythonPackage, CudaPackage):
         env.set("PKG_CONFIG_EXECUTABLE", self.spec["pkgconfig"].prefix.bin.join("pkg-config"))
         if "cmake" in self.spec:
             env.set("HOROVOD_CMAKE", self.spec["cmake"].command.path)
-        env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
+
+        if not should_support_jobserver:
+            env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
 
         # Frameworks
         if "frameworks=tensorflow" in self.spec:
--- a/var/spack/repos/builtin/packages/py-horovod/package.py
+++ b/var/spack/repos/builtin/packages/py-horovod/package.py
@@ -222,7 +222,7 @@ class PyHorovod(PythonPackage, CudaPackage):
             env.set("HOROVOD_CMAKE", self.spec["cmake"].command.path)
 
         if not should_support_jobserver:
-            env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
+            env.append_flags("MAKEFLAGS", "-j{0}".format(make_jobs))
 
         # Frameworks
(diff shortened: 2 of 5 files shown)


Jordan474:
Edit: I actually changed it to *not* change `SPACK_NO_PARALLEL_MAKE` behavior, I think.

Jordan474:
@spackbot run pipeline

var/spack/repos/builtin/packages/r/package.py
```diff
@@ -215,7 +215,8 @@ def setup_dependent_build_environment(self, env, dependent_spec):
 
         # Use the number of make_jobs set in spack. The make program will
         # determine how many jobs can actually be started.
-        env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
+        if not should_support_jobserver:
+            env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
```
haampie:
```suggestion
            env.append_flags("MAKEFLAGS", "-j{0}".format(make_jobs))
```

var/spack/repos/builtin/packages/qt/package.py
```diff
@@ -310,7 +310,8 @@ def url_for_version(self, version):
         return url
 
     def setup_build_environment(self, env):
-        env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
+        if not should_support_jobserver:
+            env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
```
haampie:
```suggestion
            env.append_flags("MAKEFLAGS", "-j{0}".format(make_jobs))
```

var/spack/repos/builtin/packages/py-horovod/package.py
```diff
@@ -220,7 +220,9 @@ def setup_build_environment(self, env):
         env.set("PKG_CONFIG_EXECUTABLE", self.spec["pkgconfig"].prefix.bin.join("pkg-config"))
         if "cmake" in self.spec:
             env.set("HOROVOD_CMAKE", self.spec["cmake"].command.path)
-        env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
+
+        if not should_support_jobserver:
+            env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
```
haampie:
```suggestion
            env.append_flags("MAKEFLAGS", "-j{0}".format(make_jobs))
```

var/spack/repos/builtin/packages/cmake/package.py
```diff
@@ -326,12 +327,9 @@ def bootstrap_args(self):
         if not sys.platform == "win32":
             args.append("--prefix={0}".format(self.prefix))
 
-            jobs = spack.build_environment.get_effective_jobs(
-                make_jobs,
-                parallel=self.parallel,
-                supports_jobserver=self.generator.supports_jobserver,
-            )
-            if jobs is not None:
+            jobs = 1 if env_flag(spack.build_environment.SPACK_NO_PARALLEL_MAKE) else make_jobs
```
haampie, line 331: Maybe easier if `SPACK_NO_PARALLEL_MAKE` implies `not should_support_jobserver`?

haampie (changes requested):

var/spack/repos/builtin/packages/cmake/package.py
Jordan474: I don't see where it simplifies: we would still need to set `jobs` to `1` if `SPACK_NO_PARALLEL_MAKE` ?

I don't know the original intent behind `SPACK_NO_PARALLEL_MAKE`, but it would simplify if we make it imply `make_jobs = 1` globally.

To me `SPACK_NO_PARALLEL_MAKE` looks like it means "disable parallel only for `make` commands", but even the current implementation/usage seems a bit broader (and [comment](https://github.com/spack/spack/blob/fce95e2efb76da3c6ee549f2a9b4e257fb42ae50/lib/spack/spack/build_environment.py#L87)).

Jordan474: Here is a version that changes `SPACK_NO_PARALLEL_MAKE` behavior to disable all parallel builds (`make_jobs = 1`), and with subsequent simplifications in recipes: [compared to this MR](https://github.com/Jordan474/spack/compare/simpler-jobserver-support...simpler-jobserver-support-2), and [compared to develop](https://github.com/spack/spack/compare/develop...Jordan474:spack:simpler-jobserver-support-2).

Does it look better to you? should I merge that in this PR?

haampie:
@spackbot fix style

var/spack/repos/builtin/packages/cmake/package.py
Jordan474: Here is the version with an even simpler API: https://github.com/spack/spack/pull/35644 .

Jordan474:
Closing in favor of #35644

PR #35202, opened 2023-01-27
wrf: add v4.2.2

stephenmsachs:


var/spack/repos/builtin/packages/wrf/package.py
```diff
@@ -143,8 +144,8 @@ class Wrf(Package):
     patch("patches/4.0/tirpc_detect.patch", when="@4.0")
     patch("patches/4.0/add_aarch64.patch", when="@4.0")
 
-    patch("patches/4.2/arch.Config.pl.patch", when="@4.2:")
-    patch("patches/4.2/arch.configure.defaults.patch", when="@4.2")
+    patch("patches/4.2/arch.Config.pl.patch", when="@4.2:4.2.1,4.3:4.3.2")
```
tldahlgren, line 147: Curious.  Why does the patch not specifically apply to `4.2.2` when it applies to `4.3:4.3.2`?

tldahlgren (commented):
Confirmed sha256 but have a question about a patch.

var/spack/repos/builtin/packages/wrf/package.py
stephenmsachs: Yes, their release schedule is a little strange. The [fix](https://github.com/wrf-model/WRF/pull/1552) went in on Jan 6,2022 and made it into 4.2.2 (released Jan 15). Version 4.3.2 (release Dec 15, 2021) did not get the patch.

haampie: It's just normal backports, right

haampie:
This has conflicts now

stephenmsachs:
Version has been added.

PR #35206, opened 2023-01-27
simplify spack install behavior

haampie:
`spack install` still has surprising behavior in the context of *environments*.

This commit simplifies things a bit, so that we have the following:

Example one:

```
spack install --add x y z
```

is equivalent to

```
spack add x y z
spack concretize
spack install --only-concrete
```

where `--only-concrete` installs without modifying spack.yaml/spack.lock

Example two:

```
spack install
```

concretizes current spack.yaml if outdated and installs all specs.

Example three:

```
spack install x y z
```

concretizes current spack.yaml if outdated and installs *only* concrete
specs in the environment that match abstract specs `x`, `y`, or `z`.

---

This is much better than the current behavior, which I have difficulty
even explaining:

- `spack install x y z` currently concretizes specs x, y, z *before*
  looking them up in the environment, so generally it's useless,
  and if it works, it's doing redundant work.
- `spack install --add x y z` can't make up its mind whether it should
  filter/restrict the concrete specs to x y z and add them as root specs
  *or* when they don't exist add them to the env -- but this fails when
  using unify: true, which is the default for environments. Very
  confusing...

### DIFF
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -411,7 +411,7 @@ def install_with_active_env(env: ev.Environment, args, install_kwargs, reporter_
 
     # `spack install x y z` without --add is installing matching specs in the env.
     else:
-        specs_to_install = env.all_matching_specs(specs)
+        specs_to_install = env.all_matching_specs(*specs)
         if not specs_to_install:
             msg = (
                 "Cannot install '{0}' because no matching specs are in the current environment."
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -23,7 +23,6 @@ import spack.paths
 import spack.report
 import spack.spec
 import spack.store
-import spack.traverse as traverse
 from spack.error import SpackError
 from spack.installer import PackageInstaller
 
@@ -369,17 +368,6 @@ def install(parser, args):
         raise
 
 
-# Install inside environments
-
-
-def _filter_specs_in_env(env: ev.Environment, abstract_specs: List[spack.spec.Spec]):
-    return [
-        s
-        for s in traverse.traverse_nodes(env.concrete_roots(), key=lambda s: s.dag_hash())
-        if any(s.satisfies(t) for t in abstract_specs)
-    ]
-
-
 def _maybe_add_and_concretize(args, env, specs):
     """Handle the overloaded spack install behavior of adding
     and automatically concretizing specs"""
@@ -423,7 +411,7 @@ def install_with_active_env(env: ev.Environment, args, install_kwargs, reporter_
 
     # `spack install x y z` without --add is installing matching specs in the env.
     else:
-        specs_to_install = _filter_specs_in_env(env, specs)
+        specs_to_install = env.all_matching_specs(specs)
         if not specs_to_install:
             msg = (
                 "Cannot install '{0}' because no matching specs are in the current environment."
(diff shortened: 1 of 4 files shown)


lib/spack/spack/cmd/install.py
```diff
@@ -489,43 +341,164 @@ def install(parser, args):
...
+            install_with_active_env(env, args, install_kwargs, reporter_factory)
+        else:
+            install_without_active_env(args, install_kwargs, reporter_factory)
+    except spack.build_environment.InstallError as e:
+        if args.show_log_on_error:
+            _dump_log_on_error(e)
+        raise
+
+
+# Install inside environments
```
alalazo: Nitpick, but I think these comments out of functions will become obsolete in time. Can we remove them?

lib/spack/spack/cmd/install.py
```diff
@@ -489,43 +341,164 @@ def install(parser, args):
...
+
+
+def _filter_specs_in_env(env: ev.Environment, abstract_specs: List[spack.spec.Spec]):
+    return [
+        s
+        for s in traverse.traverse_nodes(env.concrete_roots(), key=lambda s: s.dag_hash())
+        if any(s.satisfies(t) for t in abstract_specs)
+    ]
```
alalazo: Can we add this to `Environment`? I think it would be good to have an API to install only the specs matching some abstract requirement (i.e. a partial installation of the environment). A first application might be in the bootstrap environment to distinguish `--dev` dependencies from runtime dependencies if we add them.

lib/spack/spack/cmd/install.py
```diff
@@ -489,43 +341,164 @@ def install(parser, args):
...
+        with reporter_factory(specs_to_install):
+            env.install_specs(specs_to_install, **install_kwargs)
+    finally:
+        # TODO: this is doing way too much to trigger
+        # views and modules to be generated.
+        with env.write_transaction():
+            env.write(regenerate=True)
+
+
+# Install directly from the command line
```
alalazo: Same as above

lib/spack/spack/cmd/install.py
```diff
@@ -489,43 +341,164 @@ def install(parser, args):
...
+                s = spack.spec.Spec.from_json(f)
+
+        concretized = s.concretized()
+        if concretized.dag_hash() != s.dag_hash():
+            msg = 'skipped invalid file "{0}". '
+            msg += "The file does not contain a concrete spec."
+            tty.warn(msg.format(file))
+            continue
+        result.append(concretized)
+    return result
```
alalazo, line 473: Didn't carefully check these two since it seems they just moved around in the file. Let me know if there are any differences I didn't spot.

alalazo (commented):
I just had a first look at the code and basically LGTM. Will go over test driving the PR later. The main comment is that I think it would be worthwhile to move the "abstract spec matching" partial installation of an environment into the `Environment` class.

lib/spack/spack/cmd/install.py
haampie: Should be the same

lib/spack/spack/environment/environment.py
```diff
@@ -1863,6 +1865,15 @@ def get_one_by_hash(self, dag_hash):
...
         return hash_matches[0]
 
+    def all_matching_specs(self, *specs: spack.spec.Spec) -> List[Spec]:
+        """Returns all concretized specs in the environment satisfying any of the input specs"""
+        key = lambda s: s.dag_hash()
+        return [
+            s
+            for s in spack.traverse.traverse_nodes(self.concrete_roots(), key=key)
+            if any(s.satisfies(t) for t in specs)
+        ]
```
alalazo, line 1875: :+1:

alalazo (approved):
Code LGTM. Test driven on a couple of simple use cases (partial environment installation, automatic re-cocnretization when installing, etc.), and works as expected

PR #35208, opened 2023-01-27
py-nexusforge: add with dependencies

heerener:


var/spack/repos/builtin/packages/py-nexusforge/package.py
```diff
@@ -0,0 +1,35 @@
...
+    manage knowledge graphs.
+    """
+
+    homepage = "https://github.com/BlueBrain/nexus-forge"
+    pypi = "nexusforge/nexusforge-0.7.0.tar.gz"
+
+    version("0.8.0", sha256="4358505ead26e41c2a0c4e6113cf3a486c9661e2a3899394497a2b5a94b70424")
+    version("0.7.0", sha256="a8d2951d9ad18df9f2f4db31a4c18fcdd27bfcec929b03a3c91f133ea439413c")
+
+    depends_on("py-setuptools", type="build")
```
adamjstewart, line 23: missing build dep on setuptools-scm

var/spack/repos/builtin/packages/py-nexusforge/package.py
```diff
@@ -0,0 +1,35 @@
...
+    depends_on("py-nexus-sdk", type=("build", "run"))
+    depends_on("py-aiohttp", type=("build", "run"))
+    depends_on("py-rdflib@6.0.0:", type=("build", "run"))
+    depends_on("py-pyld", type=("build", "run"))
+    depends_on("py-pyshacl@0.17.2", type=("build", "run"))
+    depends_on("py-nest-asyncio@1.5.1:", type=("build", "run"))
+    depends_on("py-pyparsing@2.0.2:", type=("build", "run"))
+    depends_on("py-owlrl@5.2.3:", type=("build", "run"))
+    depends_on("py-elasticsearch-dsl@7.4.0", type=("build", "run"))
+    depends_on("py-scikit-learn", type="run")
```
adamjstewart: sklearn is optional, I would either remove it or add a variant (default to False) that enables it

var/spack/repos/builtin/packages/py-prettytable/package.py
```diff
@@ -18,11 +18,10 @@ class PyPrettytable(PythonPackage):
     version("3.4.1", sha256="7d7dd84d0b206f2daac4471a72f299d6907f34516064feb2838e333a4e2567bd")
     version("3.2.0", sha256="ae7d96c64100543dc61662b40a28f3b03c0f94a503ed121c6fca2782c5816f81")
     version("2.4.0", sha256="18e56447f636b447096977d468849c1e2d3cfa0af8e7b5acfcf83a64790c0aca")
+    version("2.2.1", sha256="6d465005573a5c058d4ca343449a5b28c21252b86afcdfa168cdc6a440f0b24c")
     version("0.7.2", sha256="2d5460dc9db74a32bcc8f9f67de68b2c4f4d2f01fa3bd518764c69156d9cacd9")
 
     depends_on("py-setuptools", type="build")
     depends_on("py-wcwidth", type=("build", "run"), when="@2.4.0:")
-    depends_on("py-importlib-metadata", type=("build", "run"), when="@2.4.0: ^python@:3.7")
+    depends_on("py-importlib-metadata", type=("build", "run"), when="^python@:3.7")
```
adamjstewart:
```suggestion
    depends_on("py-importlib-metadata", type=("build", "run"), when="@2: ^python@:3.7")
```
Not needed for 0.7.2

var/spack/repos/builtin/packages/py-pyshacl/package.py
```diff
@@ -0,0 +1,24 @@
...
+class PyPyshacl(PythonPackage):
+    """A Python validator for SHACL."""
+
+    homepage = "https://github.com/RDFLib/pySHACL"
+    pypi = "pyshacl/pyshacl-0.17.2.tar.gz"
+
+    version("0.20.0", sha256="47f014c52cc69167b902c89b3940dd400f7f5d2169a62f97f837f3419b4a737d")
+    version("0.17.2", sha256="46f31c7a7f7298aa5b483d92dbc850ff79a144d26f1f41e83267ed84b4d6ae23")
+
+    depends_on("py-poetry", type="build")
```
adamjstewart:
```suggestion
    depends_on("python@3.7:3", type=("build", "run"))
    depends_on("py-poetry-core@1.1:1", type="build")
```

var/spack/repos/builtin/packages/py-pyshacl/package.py
```diff
@@ -0,0 +1,24 @@
...
+    depends_on("py-rdflib@6.0.0:6", when="@0.17.2", type=("build", "run"))
+    depends_on("py-rdflib@6.2.0:6", when="@0.20.0:", type=("build", "run"))
+    depends_on("py-html5lib@1.1:1", when="@0.20.0:", type=("build", "run"))
+    depends_on("py-owlrl@5.2.3:6", when="@0.17.2", type=("build", "run"))
+    depends_on("py-owlrl@6.0.2:6", when="@0.20.0:", type=("build", "run"))
+    depends_on("py-packaging@21.3:", when="@0.20.0:", type=("build", "run"))
+    depends_on("py-prettytable@2.2.1", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-prettytable@2.2.1:2", type=("build", "run"))
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pyshacl/package.py
heerener: I notice the python dependency is often not specified (I guess it's implied by inheriting from PythonPackage). Do you add it here explicitly because of [the `Python :: 3 :: Only` specification](https://github.com/RDFLib/pySHACL/blob/v0.20.0/pyproject.toml#L26)?

adamjstewart: Spack officially only supports Python 3.7+. So I usually don't add the Python version for something like `@2.7,3.5:` or `@3.5:` because it's implied. However, some day, Python 4 will come out, and I don't want to have to retroactively update all packages. Even if this package happened to work fine in Python 4, the build system says it only supports Python 3, so the build would still fail. So in summary, I only specify required Python versions when it's a subset of 3.7+ (e.g., `@3.8:`, `@:3`).

var/spack/repos/builtin/packages/py-nexusforge/package.py
```diff
@@ -0,0 +1,38 @@
...
```
adamjstewart:
```suggestion
    depends_on("py-scikit-learn", when="+sklearn", type="run")
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pyshacl/package.py
heerener: Makes sense, thanks for clearing that up!

adamjstewart (approved):

PR #35209, opened 2023-01-27
mash: patched newer GCC versions

luke-dt:
fixed conflict with gcc >=11 with a simple patch. Confirmed builds and runs on gcc 12.2.1


tldahlgren (approved):
LGTM.  Will defer merge decision to give @marcusboden time to comment.

marcusboden (approved):
LGTM

PR #35210, opened 2023-01-27
`thrift`: add v0.17.0

thomas-bouvier:
Also, don't build kotlin and ruby libs if not explicitly enabled. My build was failing because of missing dependencies, even though I don't care about these languages (and didn't ask for their support).

### DIFF
--- a/var/spack/repos/builtin/packages/thrift/package.py
+++ b/var/spack/repos/builtin/packages/thrift/package.py
@@ -22,6 +22,7 @@ class Thrift(Package):
     list_url = "http://archive.apache.org/dist/thrift/"
     list_depth = 1
 
+    version("0.17.0", sha256="b272c1788bb165d99521a2599b31b97fa69e5931d099015d91ae107a0b0cc58f")
     version("0.16.0", sha256="f460b5c1ca30d8918ff95ea3eb6291b3951cf518553566088f3f2be8981f6209")
     version("0.13.0", sha256="7ad348b88033af46ce49148097afe354d513c1fca7c607b59c33ebb6064b5179")
     version("0.12.0", sha256="c336099532b765a6815173f62df0ed897528a9d551837d627c1f87fadad90428")
@@ -89,6 +90,8 @@ class Thrift(Package):
         options.append("--with-go=%s" % ("yes" if "+go" in spec else "no"))
         options.append("--with-lua=%s" % ("yes" if "+lua" in spec else "no"))
         options.append("--with-php=%s" % ("yes" if "+php" in spec else "no"))
+        options.append("--with-kotlin=%s" % ("yes" if "+kotlin" in spec else "no"))
+        options.append("--with-ruby=%s" % ("yes" if "+ruby" in spec else "no"))
         options.append("--with-qt4=%s" % ("yes" if "+qt4" in spec else "no"))
 
         configure(*options)
--- a/var/spack/repos/builtin/packages/thrift/package.py
+++ b/var/spack/repos/builtin/packages/thrift/package.py
@@ -31,12 +31,12 @@ class Thrift(Package):
     version("0.9.3", sha256="b0740a070ac09adde04d43e852ce4c320564a292f26521c46b78e0641564969e")
 
     # Currently only support for c-family and python
-    variant("c", default=True, description="Build support for C-family languages")
     variant("pic", default=True, description="Build position independent code")
+    variant("c", default=True, description="Build support for C-family languages")
     variant("python", default=True, description="Build support for python")
+    variant("java", default=False, description="Build support for java")
 
     depends_on("pkgconfig", type="build")
-    depends_on("java")
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool", type="build")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/thrift/package.py
```diff
@@ -89,6 +90,8 @@ def install(self, spec, prefix):
         options.append("--with-go=%s" % ("yes" if "+go" in spec else "no"))
         options.append("--with-lua=%s" % ("yes" if "+lua" in spec else "no"))
         options.append("--with-php=%s" % ("yes" if "+php" in spec else "no"))
```
tldahlgren: Weird!  There are no variants to support these options.

var/spack/repos/builtin/packages/thrift/package.py
```diff
@@ -89,6 +90,8 @@ def install(self, spec, prefix):
...
+        options.append("--with-kotlin=%s" % ("yes" if "+kotlin" in spec else "no"))
+        options.append("--with-ruby=%s" % ("yes" if "+ruby" in spec else "no"))
```
tldahlgren: These need corresponding `variant()` directives.

tldahlgren (changes requested):
Confirmed the sha256 but variants are needed.

var/spack/repos/builtin/packages/thrift/package.py
tldahlgren: One's needed for `java` as well.

var/spack/repos/builtin/packages/thrift/package.py
thomas-bouvier: I chose not to build these libs instead. Someone can do it in the future if needed. Is that ok?

tldahlgren: Fine by me.

var/spack/repos/builtin/packages/thrift/package.py
```diff
@@ -86,10 +89,12 @@ def install(self, spec, prefix):
...
-        options.append("--with-go=%s" % ("yes" if "+go" in spec else "no"))
-        options.append("--with-lua=%s" % ("yes" if "+lua" in spec else "no"))
-        options.append("--with-php=%s" % ("yes" if "+php" in spec else "no"))
-        options.append("--with-qt4=%s" % ("yes" if "+qt4" in spec else "no"))
+        options.append("--with-go=no")
+        options.append("--with-lua=no")
+        options.append("--with-php=no")
+        options.append("--with-kotlin=no")
+        options.append("--with-ruby=no")
+        options.append("--with-qt4=no")
```
tldahlgren, line 99: @tgamblin @alfredo-gimenez Any objections to disabling these options (since there appear to be no corresponding variants for them)?

var/spack/repos/builtin/packages/thrift/package.py
```diff
@@ -30,12 +31,12 @@ class Thrift(Package):
     version("0.9.3", sha256="b0740a070ac09adde04d43e852ce4c320564a292f26521c46b78e0641564969e")
 
     # Currently only support for c-family and python
-    variant("c", default=True, description="Build support for C-family languages")
     variant("pic", default=True, description="Build position independent code")
+    variant("c", default=True, description="Build support for C-family languages")
     variant("python", default=True, description="Build support for python")
+    variant("java", default=False, description="Build support for java")
```
tldahlgren: If you're grouping variants by language, perhaps put the languages in alphabetical order?  I know it's a nit though.

tldahlgren:
Containers not ready error for one stack.

@spackbot run pipeline

var/spack/repos/builtin/packages/thrift/package.py
tldahlgren: From https://thrift.apache.org/docs/install/ it looks like `java` needs to be at version `1.7`.  There's also a dependency on `ant` for `java`.

var/spack/repos/builtin/packages/thrift/package.py
thomas-bouvier: Could the variant propagation mechanism detailed [here](https://spack.readthedocs.io/en/latest/basic_usage.html#variants) cause these options to activate, even though said variant isn't explicitly defined in that package?

thomas-bouvier: @haampie do you think these changes are safe to add?

tldahlgren: IMO I think the issue isn't so much if they are safe or not to add but getting the build recipe "right" (e.g., dependencies, versions of dependencies) for those that are added.

tldahlgren (approved):
I am able to successfully build the package with `+c+java`.

Would be nice to have tests 😁 .  See https://spack.readthedocs.io/en/latest/packaging_guide.html#checking-an-installation.

tldahlgren:
Interesting.  Four stacks had "unready" container issues.

@spackbot run pipeline

thomas-bouvier:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

thomas-bouvier:
@spackbot run pipeline

thomas-bouvier:
What's happening with the CI?

@spackbot run pipeline

PR #35216, opened 2023-01-27
Bugfix: Update Petsc Python dependency

tldahlgren:
Was trying to build a package depending on `petsc` and `petsc` kept failing with:

```
*********************************************************************************************
*                     Python version 3.4+ is required to run ./configure                    *
*********************************************************************************************
```

Had the same issue when I tried depending on `petsc@3.18.0`.  This change fixes the issue for me.

I didn't check any versions before `3.18`.

### DIFF
--- a/var/spack/repos/builtin/packages/petsc/package.py
+++ b/var/spack/repos/builtin/packages/petsc/package.py
@@ -205,7 +205,8 @@ class Petsc(Package, CudaPackage, ROCmPackage):
     depends_on("rocprim", when="+rocm")
 
     # Build dependencies
-    depends_on("python@2.6:2.8,3.4:", type="build")
+    depends_on("python@2.6:2.8,3.4:3.8", when="@:3.13", type="build")
+    depends_on("python@2.6:2.8,3.4:", when="@3.14:3.17", type="build")
     depends_on("python@3.4:", when="@3.18:", type="build")
 
     # Other dependencies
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/petsc/package.py
```diff
@@ -206,6 +206,7 @@ def check_fortran_compiler(self):
 
     # Build dependencies
     depends_on("python@2.6:2.8,3.4:", type="build")
+    depends_on("python@3.4:", when="@3.18:", type="build")
```
balay:
```suggestion
    depends_on("python@2.6:2.8,3.4:3.8", when="@:3.13", type="build")
    depends_on("python@2.6:2.8,3.4:", when="@3.14:3.17", type="build")
    depends_on("python@3.4:", when="@3.18:", type="build")
```

balay:
> *                     Python version 3.4+ is required to run ./configure                    *

Somehow python < 3.4 got picked up in the build?  

How was this possible - since spack removed support for python versions 3.5 and below.

tldahlgren:
> > * ```
> >                   Python version 3.4+ is required to run ./configure                    *
> >   ```
> 
> 
> Somehow python < 3.4 got picked up in the build?
> 
> 
> How was this possible - since spack removed support for python versions 3.5 and below.

Yep. I am running on a system that still has its default `python` at `2.7.16`.  My `spack` is using a newer `python`.

tldahlgren:
@balay Thanks for a better set of `python` dependencies!

balay (approved):

alalazo:
@spackbot run pipeline

PR #35217, opened 2023-01-27
Update `py-docker` package

spoutn1k:
I recently looked at using the `py-docker` package, but it was outdated. This PR proposes an updated version.

### DIFF
--- a/var/spack/repos/builtin/packages/py-docker/package.py
+++ b/var/spack/repos/builtin/packages/py-docker/package.py
@@ -17,7 +17,6 @@ class PyDocker(PythonPackage):
     version("5.0.3", sha256="d916a26b62970e7c2f554110ed6af04c7ccff8e9f81ad17d0d40c75637e227fb")
     version("4.2.1", sha256="380a20d38fbfaa872e96ee4d0d23ad9beb0f9ed57ff1c30653cbeb0c9c0964f2")
 
-    variant("tls", default=True, description="Required to use docker-py over TLS")
     variant("ssh", default=False, description="Required to use docker-py over SSH")
 
     depends_on("py-setuptools", type="build")
--- a/var/spack/repos/builtin/packages/py-docker/package.py
+++ b/var/spack/repos/builtin/packages/py-docker/package.py
@@ -29,4 +29,5 @@ class PyDocker(PythonPackage):
     depends_on("py-packaging@14.0:", when="@6:", type=("build", "run"))
     depends_on("py-urllib3@1.26.0:", when="@6:", type=("build", "run"))
 
-    depends_on("py-paramiko@2.4.3:", when="+ssh", type=("build", "run"))
+    depends_on("py-paramiko@2.4.3:", when="@6:+ssh", type=("build", "run"))
+    depends_on("py-paramiko@2.4.2:", when="+ssh", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-docker/package.py
+++ b/var/spack/repos/builtin/packages/py-docker/package.py
@@ -29,4 +29,4 @@ class PyDocker(PythonPackage):
     depends_on("py-packaging@14.0:", when="@6:", type=("build", "run"))
     depends_on("py-urllib3@1.26.0:", when="@6:", type=("build", "run"))
 
-    depends_on("py-paramiko@2.4.2:", when="+ssh", type=("build", "run"))
+    depends_on("py-paramiko@2.4.3:", when="+ssh", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-docker/package.py
+++ b/var/spack/repos/builtin/packages/py-docker/package.py
@@ -13,10 +13,21 @@ class PyDocker(PythonPackage):
     homepage = "https://github.com/docker/docker-py"
     pypi = "docker/docker-4.2.1.tar.gz"
 
+    version("6.0.1", sha256="896c4282e5c7af5c45e8b683b0b0c33932974fe6e50fc6906a0a83616ab3da97")
+    version("5.0.3", sha256="d916a26b62970e7c2f554110ed6af04c7ccff8e9f81ad17d0d40c75637e227fb")
     version("4.2.1", sha256="380a20d38fbfaa872e96ee4d0d23ad9beb0f9ed57ff1c30653cbeb0c9c0964f2")
 
-    depends_on("python@2.7:2.8,3.5:", type=("build", "run"))
+    variant('tls', default=True, description="Required to use docker-py over TLS")
+    variant('ssh', default=False, description="Required to use docker-py over SSH")
+
     depends_on("py-setuptools", type="build")
-    depends_on("py-six@1.4.0:", type=("build", "run"))
+    depends_on("py-setuptools@45:", when="@6:", type="build")
+    depends_on("py-setuptools-scm+toml@6.2:", when="@6:", type="build")
+    depends_on("py-six@1.4.0:", when="@4.2.1", type=("build", "run"))
     depends_on("py-websocket-client@0.32.0:", type=("build", "run"))
     depends_on("py-requests@2.14.2:2.17,2.18.1:", type=("build", "run"))
+    depends_on("py-requests@2.26.0:", when="@6:", type=("build", "run"))
+    depends_on("py-packaging@14.0:", when="@6:", type=("build", "run"))
+    depends_on("py-urllib3@1.26.0:", when="@6:", type=("build", "run"))
+
+    depends_on("py-paramiko@2.4.2:", when="+ssh", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


spoutn1k:
I added the `py-enum34`, `py-ordereddict` packages that `docker@4.2.1` required, but I see the workflow rejected them on the grounds of those version not being resolved. Should I understand that `spack` does not support EOL'ed `python` versions ?

adamjstewart:
Correct, Spack only supports Python 3.7+ for installation, so you can ignore any dependencies only needed for EOL Python versions.

spoutn1k:
https://github.com/docker/docker-py/blob/aca129dd69fa068da247d37266110c3a9889a780/requirements.txt#L3

I am also ignoring the `win32` dependencies, unless there are plans to have `spack` work on Windows ?

adamjstewart:
Spack does support Windows! https://spack.readthedocs.io/en/latest/getting_started.html#spack-on-windows

I usually add Windows deps like colorama for which we already have a package, but I sometimes ignore Windows packages we haven't yet packaged since I can't actually test that they work myself.

var/spack/repos/builtin/packages/py-docker/package.py
```diff
@@ -13,10 +13,46 @@ class PyDocker(PythonPackage):
     homepage = "https://github.com/docker/docker-py"
     pypi = "docker/docker-4.2.1.tar.gz"
 
+    version("6.0.1", sha256="896c4282e5c7af5c45e8b683b0b0c33932974fe6e50fc6906a0a83616ab3da97")
+    version("6.0.0", sha256="19e330470af40167d293b0352578c1fa22d74b34d3edf5d4ff90ebc203bbb2f1")
+    version("5.0.3", sha256="d916a26b62970e7c2f554110ed6af04c7ccff8e9f81ad17d0d40c75637e227fb")
     version("4.2.1", sha256="380a20d38fbfaa872e96ee4d0d23ad9beb0f9ed57ff1c30653cbeb0c9c0964f2")
 
-    depends_on("python@2.7:2.8,3.5:", type=("build", "run"))
+    depends_on("python", type=("build", "run"))
```
adamjstewart: You can remove this, it's already added by `PythonPackage`

var/spack/repos/builtin/packages/py-docker/package.py
```diff
@@ -13,10 +13,46 @@ class PyDocker(PythonPackage):
...
     depends_on("py-setuptools", type="build")
```
adamjstewart: Needs to be updated:
```
requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2"]
```

var/spack/repos/builtin/packages/py-docker/package.py
```diff
@@ -13,10 +13,46 @@ class PyDocker(PythonPackage):
...
+
+    depends_on("py-requests@2.28.1", when="@6:", type=("build", "run"))
+    depends_on("py-requests@2.26.0", when="@5", type=("build", "run"))
+    depends_on("py-requests@2.20.0", when="@4.2.1", type=("build", "run"))
+
+    depends_on("py-six@1.10.0", when="@4.2.1", type=("build", "run"))
+
+    depends_on("py-websocket-client@1.3.3", when="@6:", type=("build", "run"))
+    depends_on("py-websocket-client@0.56.0:", when="@5.0.3", type=("build", "run"))
+    depends_on("py-websocket-client@0.32.0:", when="@4.2.1", type=("build", "run"))
```
adamjstewart: Most of these dependencies are wrong. I only checked 6.0.1, where the only required dependencies are:
```python
requirements = [                                                                         
    'packaging >= 14.0',                                                                 
    'requests >= 2.26.0',                                                                
    'urllib3 >= 1.26.0',                                                                 
    'websocket-client >= 0.32.0',                                                        
] 
```

adamjstewart (changes requested):

spoutn1k:
Oh snap ! Great work from you guys. I can't test on windows either because of allergies, and `pywin32` seems not to be implemented in `spack`, so I'll leave it be.

Any workflow I should follow to ensure my packages are valid ? I made several changes and a bunch of dependencies were dropped by `py-docker@6`, so I am looking for validity checks/dependency trees outputs for the packets. I found `spack graph`, but the output is illegible.

var/spack/repos/builtin/packages/py-docker/package.py
spoutn1k: Oh wow I was looking at the `requirements.txt` file in their project, which lists the dependencies I added, but it seems to be a red herring. I'll update that.

adamjstewart:
> Any workflow I should follow to ensure my packages are valid ?

Valid can mean many things to many people. If by valid you mean correct, I just look at the pyproject.toml/setup.py/setup.cfg to see a list of dependencies and make sure they match the package recipe. I would recommend reading https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html if you haven't already.

spoutn1k:
I mistakenly thought `requirements.txt` listed the requirements, while there was another list in `setup.py`; started from scratch and got a much cleaner commit.

I am in a bit of a pickle: for `py-docker@:5`, the `tls` variant requires `py-requests+security`, which does not exist as it ceased being a variant. How should I handle this ?

adamjstewart:
You don't _have_ to add all of these variants if you don't need them. If you _do_ need them, then you could add a no-op variant to requests, or just ignore the additional dependency entirely since it sounds like that feature is included by default now.

spoutn1k:
I don't need it, just trying to do the best possible job to learn as much as I can. I guess I will just ignore it.

While I make that last change, do you have an idea what the style checker is getting mad about ?

```diff
--- var/spack/repos/builtin/packages/py-docker/package.py	2023-01-30 21:16:20.267054 +0000
+++ var/spack/repos/builtin/packages/py-docker/package.py	2023-01-30 21:16:30.771452 +0000
@@ -15,12 +15,12 @@
 
     version("6.0.1", sha256="896c4282e5c7af5c45e8b683b0b0c33932974fe6e50fc6906a0a83616ab3da97")
     version("5.0.3", sha256="d916a26b62970e7c2f554110ed6af04c7ccff8e9f81ad17d0d40c75637e227fb")
     version("4.2.1", sha256="380a20d38fbfaa872e96ee4d0d23ad9beb0f9ed57ff1c30653cbeb0c9c0964f2")
 
-    variant('tls', default=True, description="Required to use docker-py over TLS")
-    variant('ssh', default=False, description="Required to use docker-py over SSH")
+    variant("tls", default=True, description="Required to use docker-py over TLS")
+    variant("ssh", default=False, description="Required to use docker-py over SSH")
 
     depends_on("py-setuptools", type="build")
     depends_on("py-setuptools@45:", when="@6:", type="build")
     depends_on("py-setuptools-scm+toml@6.2:", when="@6:", type="build")
     depends_on("py-six@1.4.0:", when="@4.2.1", type=("build", "run"))
```

I checked and cannot find any whitespace nor tab inconsistencies.

adamjstewart:
It's converting single quotes to double quotes. We use black now which prefers double quotes. You can run `spack style --fix` on the command line or `@spackbot fix style` on GitHub to automatically solve the problem.

var/spack/repos/builtin/packages/py-docker/package.py
```diff
@@ -13,10 +13,20 @@ class PyDocker(PythonPackage):
...
+    depends_on("py-setuptools@45:", when="@6:", type="build")
+    depends_on("py-setuptools-scm+toml@6.2:", when="@6:", type="build")
+    depends_on("py-six@1.4.0:", when="@4.2.1", type=("build", "run"))
     depends_on("py-websocket-client@0.32.0:", type=("build", "run"))
     depends_on("py-requests@2.14.2:2.17,2.18.1:", type=("build", "run"))
+    depends_on("py-requests@2.26.0:", when="@6:", type=("build", "run"))
+    depends_on("py-packaging@14.0:", when="@6:", type=("build", "run"))
+    depends_on("py-urllib3@1.26.0:", when="@6:", type=("build", "run"))
+
+    depends_on("py-paramiko@2.4.2:", when="+ssh", type=("build", "run"))
```
adamjstewart, line 33:
```suggestion
    depends_on("py-paramiko@2.4.3:", when="+ssh", type=("build", "run"))
```

adamjstewart (changes requested):

adamjstewart (dismissed):

var/spack/repos/builtin/packages/py-docker/package.py
spoutn1k: The `2.4.3` version is only for `6.0.1:`, the rest shows `2.4.2`. I guess it is easier to just put `2.4.3` ? Should I make a distinction ?

adamjstewart: In that case it would probably be better to use:
```python
depends_on("py-paramiko@2.4.3:", when="@6:+ssh", type=("build", "run"))
depends_on("py-paramiko@2.4.2:", when="+ssh", type=("build", "run"))
```

adamjstewart (approved):

PR #35220, opened 2023-01-27
montage: fixing compilation issues for newer gcc versions

snehring:
It also needs libnsl turns out.


alalazo:
@spackbot run pipeline

alalazo (approved):

PR #35224, opened 2023-01-28
gmake: Fix too many levels of symbolic links error

alecbcs:
Fix build error "too many levels of symbolic links" that started after a recent update to the way we build gmake.


adamjstewart (approved):
Wondering if we should do this by default for all AutotoolsPackages...

alalazo:
> Wondering if we should do this by default for all AutotoolsPackages...

A few years back we were wondering the same, since it was a recommended practice in CMake to build outside the source directory. If I recall correctly we reached the conclusion at the time that most autotools package would break if built that way. We should probably check, if we want to change our default build directory, if the situation changed in the meanwhile.

PR #35227, opened 2023-01-29
acts: new versions 21.1.1, 22.0.1, 23.0.0

wdconinc:
New versions:
- [major 23.0.0](https://github.com/acts-project/acts/compare/v22.0.0...v23.0.0):
  - new option `ACTS_BUILD_PLUGIN_GEANT4` -> enabled with existing variant `geant4`
  - new option `ACTS_BUILD_EXAMPLES_BINARIES`:
    - it is my understanding that the binaries for examples are deprecated (in favor of python examples); warnings to this effect have been printed for a few versions, and now the building of binaries is disabled by default,
    - rather than introducing a variant to enable deprecated behavior for only one or two versions, I propose that we just follow the default and keep this disabled. 
- [bugfix 22.0.1](https://github.com/acts-project/acts/compare/v22.0.0...v22.0.1) (no build system changes)
- [bugfix 21.1.1](https://github.com/acts-project/acts/compare/v21.1.0...v21.1.1) (no build system changes)

### DIFF
--- a/var/spack/repos/builtin/packages/acts/package.py
+++ b/var/spack/repos/builtin/packages/acts/package.py
@@ -39,7 +39,7 @@ class Acts(CMakePackage, CudaPackage):
     # Supported Acts versions
     version("main", branch="main")
     version("master", branch="main", deprecated=True)  # For compatibility
-    version("23.0.0", commit="a4ac99dd72828c5eb3fac06e146f3391958fca8c", submodules=True)
+    version("23.0.0", commit="5af1b1b5feb8ca8f4c2c69106a1b9ef612c70d9c", submodules=True)
     version("22.0.1", commit="a4ac99dd72828c5eb3fac06e146f3391958fca8c", submodules=True)
     version("22.0.0", commit="0fb6f8d2ace65338915451201e9ceb6cee11fb5e", submodules=True)
     version("21.1.1", commit="8ae825de246e8e574d05d9eaf05ba4a937c69aa9", submodules=True)
--- a/var/spack/repos/builtin/packages/acts/package.py
+++ b/var/spack/repos/builtin/packages/acts/package.py
@@ -39,7 +39,10 @@ class Acts(CMakePackage, CudaPackage):
     # Supported Acts versions
     version("main", branch="main")
     version("master", branch="main", deprecated=True)  # For compatibility
+    version("23.0.0", commit="a4ac99dd72828c5eb3fac06e146f3391958fca8c", submodules=True)
+    version("22.0.1", commit="a4ac99dd72828c5eb3fac06e146f3391958fca8c", submodules=True)
     version("22.0.0", commit="0fb6f8d2ace65338915451201e9ceb6cee11fb5e", submodules=True)
+    version("21.1.1", commit="8ae825de246e8e574d05d9eaf05ba4a937c69aa9", submodules=True)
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/acts/package.py
```diff
@@ -39,7 +39,10 @@ class Acts(CMakePackage, CudaPackage):
     # Supported Acts versions
     version("main", branch="main")
     version("master", branch="main", deprecated=True)  # For compatibility
+    version("23.0.0", commit="a4ac99dd72828c5eb3fac06e146f3391958fca8c", submodules=True)
```
HadrienG2:
```suggestion
    version("23.0.0", commit="5af1b1b5feb8ca8f4c2c69106a1b9ef612c70d9c", submodules=True)
```

HadrienG2 (changes requested):
You accidentally copypasted the 22.0.1 commit hash into the 23.0.0 release. Here's a suggestion with the right commit. :)

Concerning examples, I'm a bit torn, since even though C++ examples have been deprecated, they remain used, for example someone just provided instructions using them on the Acts chat today.

My proposal would be to follow the strategy you're proposing for now (no option for them in the spack build) and see if acts spackage users complain about not being able to build them. If so, either add them to the existing "examples" variant or create a new "deprecated_examples" variant.

var/spack/repos/builtin/packages/acts/package.py
wdconinc: Ah oops! thanks for catching that :-)

wdconinc:
> add them to the existing "examples" variant

I think that would probably the better approach, in my opinion, rather than adding a `deprecated_examples` variant. Do you know by when the compiled binary examples will be removed entirely? Acts isn't too concerned about making breaking changes on major version changes, so I don't imagine that the deprecated code will be around for long...

tldahlgren (approved):
Confirmed the three commits exist.

HadrienG2:
> > add them to the existing "examples" variant
> 
> I think that would probably the better approach, in my opinion, rather than adding a `deprecated_examples` variant. Do you know by when the compiled binary examples will be removed entirely? Acts isn't too concerned about making breaking changes on major version changes, so I don't imagine that the deprecated code will be around for long...

I have no idea, so far the attitude has been to let them decay bit by bit as new features that would require effort on their side are added, but not actively break them as long as it does not require maintenance effort to keep them around.

HadrienG2:
Also, to clarify the breaking change thing, we're following the [semantic versioning](https://semver.org/) policy where major releases are by definition about breaking changes (with minor releases being about new features and point releases being about bugfixes).

wdconinc:
> Also, to clarify the breaking change thing, we're following the [semantic versioning](https://semver.org/) policy where major releases are by definition about breaking changes (with minor releases being about new features and point releases being about bugfixes).

I didn't intend for my comment to sound as negative as it does, upon rereading. I think it's an exemplary policy thanks to the extent to which it is clearly followed. There are more than enough natural numbers and I wish some other projects were more willing to use them.

My comment was more about the fact that Acts does not seem like a project that keeps deprecated code around for long, in the interest of reducing technical debt. So, I figured these binaries might go away with a next major release soon.

wdconinc:
@HadrienG2 Does this PR look good to you? I think it's waiting for an approval.

HadrienG2 (approved):

PR #35229, opened 2023-01-30
py-igor: new package

heerener:


var/spack/repos/builtin/packages/py-igor/package.py
```diff
@@ -0,0 +1,24 @@
...
+    """igor: interface for reading binary IGOR files."""
+
+    homepage = "http://blog.tremily.us/posts/igor/"
+    pypi = "igor/igor-0.3.tar.gz"
+
+    version("0.3", sha256="b04ac38c68fb81cf3167a9981dc5a20379112d40268bb72c5a8514dc8051abba")
+
+    def patch(self):
+        setupfile = FileFilter("setup.py")
+        setupfile.filter("distutils.core", "setuptools")
```
adamjstewart: I believe pip does this automatically. Not sure about build which we're planning to switch to in the future.

var/spack/repos/builtin/packages/py-igor/package.py
```diff
@@ -0,0 +1,24 @@
...
+
+    depends_on("py-setuptools", type="build")
+    depends_on("py-numpy", type="run")
+    depends_on("py-nose", type="run")
```
adamjstewart: Nose is only needed for unit tests, which we don't run here. I would remove it.

var/spack/repos/builtin/packages/py-igor/package.py
```diff
@@ -0,0 +1,24 @@
...
+    depends_on("py-matplotlib", type="run")
```
adamjstewart: I would make numpy and matplotlib `type=("build", "run")` so the import tests pass.

adamjstewart (changes requested):

adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

heerener:
@adamjstewart is that docs build something I can fix from my end?

adamjstewart:
Probably transient, let me try kicking the tests

PR #35240, opened 2023-01-30
py-mariadb: new package

heerener:


tldahlgren (commented):
Confirmed both sha256.

tldahlgren:

adamjstewart (approved):

PR #35243, opened 2023-01-30
apptainer use the `--with-suid` configure option

simonLeary42:
https://github.com/apptainer/apptainer/commit/c56641e670e62b1386da86aaa1e7ba1486af4cfd this commit added `--with-suid`

I ran a spack install with `+suid` and this is what happened:
```
 checking: unprivileged user namespaces... enabled. Implying --without-suid
```

### DIFF
--- a/var/spack/repos/builtin/packages/apptainer/package.py
+++ b/var/spack/repos/builtin/packages/apptainer/package.py
@@ -46,3 +46,18 @@ class Apptainer(SingularityBase):
     @property
     def config_options(self):
         return []
+
+    # Hijack the edit stage to run mconfig.
+    def edit(self, spec, prefix):
+        with working_dir(self.build_directory):
+            confstring = "./mconfig --prefix=%s" % prefix
+            if "~suid" in spec:
+                confstring += " --without-suid"
+            if "+suid" in spec and spec.satisfies("@1.1.0:"):
+                confstring += " --with-suid"            
+            if "~network" in spec:
+                confstring += " --without-network"
+            configure = Executable(confstring)
+            configure()
+
+    


var/spack/repos/builtin/packages/apptainer/package.py
```diff
@@ -46,3 +46,18 @@ class Apptainer(SingularityBase):
     @property
     def config_options(self):
         return []
+
+    # Hijack the edit stage to run mconfig.
+    def edit(self, spec, prefix):
+        with working_dir(self.build_directory):
+            confstring = "./mconfig --prefix=%s" % prefix
+            if "~suid" in spec:
+                confstring += " --without-suid"
```
alalazo, line 55: Can you change this upstream, in `SingularityBase`, to be explicit on the option?

alalazo (changes requested):

var/spack/repos/builtin/packages/apptainer/package.py
simonLeary42: that configure argument didn't exist until apptainer 1.1.0, so I don't think it would work for singularity.

simonLeary42: Actually that's not true. I can make that change

simonLeary42: It's already there: https://github.com/spack/spack/blob/42a452d54c8a25f9f415fef8cf9e3a5c64b7a46a/var/spack/repos/builtin/packages/singularityce/package.py#L84

simonLeary42: Because I'm overwriting the whole `edit` method, I copy pasted what's already there and only added this:
```python
            if "+suid" in spec and spec.satisfies("@1.1.0:"):
                confstring += " --with-suid"
```

alalazo: Can't you just override `self.config_options`, instead of duplicating `edit`?

simonLeary42: How can I make sure that my override happens after `edit` happens?

simonLeary42:
I can't make any changes since I deleted the head so I made a new PR

PR #35249, opened 2023-01-30
Use setup_build_environment instead of edit in esmf/package.py

AlexanderRichert-NOAA:
This PR changes "edit" to "setup_build_environment" in the ESMF package so that the env variables set will be present when using `spack build-env esmf`.

### DIFF
--- a/var/spack/repos/builtin/packages/esmf/package.py
+++ b/var/spack/repos/builtin/packages/esmf/package.py
@@ -156,7 +156,8 @@ class Esmf(MakefilePackage):
                 version.dotted
             )
 
-    def edit(self, spec, prefix):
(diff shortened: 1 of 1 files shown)


jedwards4b (dismissed):

climbfuji (dismissed):

alalazo (changes requested):
Can you change this to modify the environment through `env` instead of `os.environ` directly?

AlexanderRichert-NOAA:
> Can you change this to modify the environment through `env` instead of `os.environ` directly?

Yep no problem-- is there a way I can iterate over all variables in the environment using `env` instead of `os.environ`?

alalazo (approved):

PR #35254, opened 2023-01-31
halide: new package

wraith1995:


var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+            self.define("Python3_EXECUTABLE", spec["python"].command.path),
+            self.define("PYBIND11_USE_FETCHCONTENT", False),
+            self.define("LLVM_DIR", llvmdir),
+            self.define_from_variant("WITH_TESTS", "tests"),
+            self.define_from_variant("WITH_TUTORIALS", "tutorials"),
+            self.define_from_variant("WITH_UTILS", "utils"),
+            self.define_from_variant("WITH_PYTHON_BINDINGS", "python"),
+            self.define("WITH_WABT", False),
+        ]
+        if "python" in spec:
```
tldahlgren:
```suggestion
        if "+python" in spec:
```

tldahlgren (commented):
Confirmed branch and version exist.

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+class Halide(CMakePackage, PythonExtension):
+    """Halide is a programming language designed to make it easier to write
+    high-performance image and array processing code on modern machines."""
+
+    homepage = "https://halide-lang.org/"
+    url = "https://github.com/halide/Halide/archive/refs/tags/v14.0.0.tar.gz"
+    git = "https://github.com/halide/Halide.git"
+    maintainers = ["wraith1995"]
+    version("14.0.0", sha256="f9fc9765217cbd10e3a3e3883a60fc8f2dbbeaac634b45c789577a8a87999a01")
+    version("main", branch="main")
```
tldahlgren: Please move the branch to above the numeric version since we prefer branch versions first then numbered versions (in reverse order) after.

tldahlgren:
@adamjstewart Do you want to review this package since it extends `PythonExtension`?

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+            self.define_from_variant("WITH_PYTHON_BINDINGS", "python"),
+            self.define("WITH_WABT", False),
+        ]
+        if "python" in spec:
+            pyspec = str(spec["python"].version[:2])
+            prefix = spec.prefix
+            args += [
+                self.define(
+                    "Halide_INSTALL_PYTHONDIR",
+                    prefix + "/lib/python{0}/site-packages/".format(pyspec),
```
adamjstewart:
```suggestion
                    python_platlib,
```
This is defined by the `python` package and points to the correct location

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+
+    depends_on("python@3.10:", type=("build", "link", "run"), when="+python")
+    depends_on("py-pybind11@2.6.2:", type="build", when="+python")
+    depends_on("py-setuptools@43:", type="build", when="+python")
+    depends_on("py-scikit-build", type="build", when="+python")
+    depends_on("py-wheel", type="build", when="+python")
+    depends_on("py-build", type="build", when="+python")
+
+    depends_on("py-imageio", type="run", when="+python")
+    depends_on("py-pillow", type="run", when="+python")
```
adamjstewart:
```suggestion
    depends_on("pil", type="run", when="+python")
```
We have a virtual dependency for this

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+    depends_on("py-scipy", type="run", when="+python")
+    depends_on("py-numpy@1.0.0:", type="run", when="+python")
```
adamjstewart: Where does this version requirement come from?

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+    depends_on("libjpeg", type=("build", "link", "run"))
+    depends_on("libpng", type=("build", "link", "run"))
+
+    depends_on("python@3.10:", type=("build", "link", "run"), when="+python")
+    depends_on("py-pybind11@2.6.2:", type="build", when="+python")
+    depends_on("py-setuptools@43:", type="build", when="+python")
+    depends_on("py-scikit-build", type="build", when="+python")
+    depends_on("py-wheel", type="build", when="+python")
+    depends_on("py-build", type="build", when="+python")
+
```
adamjstewart, line 47: Can you add a comment pointing to `python_bindings/requirements.txt` so it's easier to find the dependencies in the future?

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    depends_on("libjpeg", type=("build", "link", "run"))
+    depends_on("libpng", type=("build", "link", "run"))
+
+    depends_on("python@3.10:", type=("build", "link", "run"), when="+python")
+    depends_on("py-pybind11@2.6.2:", type="build", when="+python")
+    depends_on("py-setuptools@43:", type="build", when="+python")
+    depends_on("py-scikit-build", type="build", when="+python")
+    depends_on("py-wheel", type="build", when="+python")
+    depends_on("py-build", type="build", when="+python")
```
adamjstewart: Where do you see these dependencies listed?

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,77 @@
...
+
+    depends_on("py-imageio", type="run", when="+python")
+    depends_on("py-pillow", type="run", when="+python")
+    depends_on("py-scipy", type="run", when="+python")
+    depends_on("py-numpy@1.0.0:", type="run", when="+python")
+
+    def cmake_args(self):
+
+        spec = self.spec
+        llvmdir = spec["llvm"].prefix.lib + "/cmake/llvm"
```
adamjstewart: @trws @haampie is there a more robust way of finding this directory since lib may be lib64 on other systems?

adamjstewart (changes requested):

var/spack/repos/builtin/packages/halide/package.py
wraith1995: On consideration - I am not sure - I can drop it without issue.

var/spack/repos/builtin/packages/halide/package.py
wraith1995: https://github.com/halide/Halide/blob/main/requirements.txt

var/spack/repos/builtin/packages/halide/package.py
wraith1995: For what it is worth I copied this from: https://github.com/spack/spack/blob/fce95e2efb76da3c6ee549f2a9b4e257fb42ae50/var/spack/repos/builtin/packages/enzyme/package.py

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+    generator = "Ninja"
+    variant("python", default=False, description="Install python bindings")
+    variant("tutorials", default=False, description="Install the Halide Tutorials.")
+    variant("utils", default=False, description="Install the Halide Utilities.")
+    variant("tests", default=False, description="Build and Run Halide Tests and Apps.")
+    extends("python", when="+python")
+
+    _targets = " targets=arm,x86,nvptx,aarch64,hexagon,webassembly "
+    depends_on("cmake@3.24.3", type="build")
```
adamjstewart:
```suggestion
    depends_on("cmake@3.22:", type="build")
```

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,76 @@
...
+    depends_on("ninja", type="build")
+    depends_on(
+        "llvm@14.0.0:14+clang+lld" + _targets + " build_type=Release",
+        type=("link", "run"),
+    )
+    depends_on("libjpeg", type=("build", "link", "run"))
+    depends_on("libpng", type=("build", "link", "run"))
+
+    depends_on("python@3.10:", type=("build", "link", "run"), when="+python")
```
adamjstewart: Where does the 3.10+ requirement come from?

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,76 @@
...
+    # See https://github.com/halide/Halide/blob/main/requirements.txt
+    depends_on("py-pybind11@2.6.2:", type="build", when="+python")
```
adamjstewart:
```suggestion
    depends_on("py-pybind11@2.6.2", type="build", when="+python")
```

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,76 @@
...
+    depends_on("py-setuptools@43:", type="build", when="+python")
+    depends_on("py-scikit-build", type="build", when="+python")
+    depends_on("py-wheel", type="build", when="+python")
+    depends_on("py-build", type="build", when="+python")
+
+    depends_on("py-imageio", type="run", when="+python")
+    depends_on("pil", type="run", when="+python")
```
adamjstewart: Do they use build or pip? I doubt they're using both.

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,76 @@
...
+    depends_on("py-scipy", type="run", when="+python")
+    depends_on("py-numpy", type="run", when="+python")
```
adamjstewart: These should all be type=build/run

var/spack/repos/builtin/packages/halide/package.py
wraith1995: It is listed but I can't actually see where it is used.  I am going to test and then  nix.

var/spack/repos/builtin/packages/halide/package.py
wraith1995: Actual min is 3.8 (https://github.com/halide/Halide/blob/main/setup.py#L18); fixing.

var/spack/repos/builtin/packages/halide/package.py
wraith1995: Okay. Build is not needed, nixing.

adamjstewart: Do you see where they're using pip?

adamjstewart: I guess if the build is working for you and you don't see it downloading any dependencies then it's fine.

adamjstewart (dismissed):
Python bits look mostly okay, although I'm confused on how the package is built and installed without build or pip.

LLVM bits look sus to me but I don't know LLVM so it would be best for someone else to review them.

var/spack/repos/builtin/packages/halide/package.py
trws: The short answer is `llvm-config --libdir`.  I don't remember off top of my head if we have a `.libdir` property for that on the llvm package, but we could.

wraith1995: Looks like you can also get away with ` llvm-config --cmakedir`. I will add and resubmit soon.

tldahlgren (approved):
Thanks for making the changes I suggested.

var/spack/repos/builtin/packages/halide/package.py
```diff
@@ -0,0 +1,74 @@
...
+        values=("Release", "Debug", "RelWithDebInfo"),
+    )
+    generator = "Ninja"
+    variant("python", default=False, description="Install python bindings")
+    variant("tutorials", default=False, description="Install the Halide Tutorials.")
+    variant("utils", default=False, description="Install the Halide Utilities.")
+    variant("tests", default=False, description="Build and Run Halide Tests and Apps.")
+    extends("python", when="+python")
+
+    _targets = " targets=arm,x86,nvptx,aarch64,hexagon,webassembly "
```
adamjstewart, line 31: These are _required_ to build halide or these are what you personally _prefer_?

wraith1995: This is the suggestion in the readme and not based on my preferences. I don't know what will happen if you build without them.

wraith1995: https://github.com/halide/Halide/tree/e8e1481633b4dd24bc0f0cf21459793ba66771b5#acquiring-llvm

adamjstewart (approved):
Python stuff still looks good, although I'm still confused how the install works without pip or build. May be worth uploading the successful build log, that should contain any build/pip commands that were used.

wraith1995:
@adamjstewart Here are build logs and an explanation: the package does not need either because the build performs roughly the following steps:

1. It builds a python extension via pybind and cmake, resulting in a shared library.
2. When installed, this is moved to <install_loc>/lib/python3.whatever/site-packages/halide/
This works because Spack, surprisingly to me, adds <install_loc>/lib/ to the python path when you load halide. You don't need pip under these circumstances.
[spack-build-03-install-out.txt](https://github.com/spack/spack/files/10574530/spack-build-03-install-out.txt)
[spack-build-02-build-out.txt](https://github.com/spack/spack/files/10574531/spack-build-02-build-out.txt)
[spack-build-01-cmake-out.txt](https://github.com/spack/spack/files/10574532/spack-build-01-cmake-out.txt)

var/spack/repos/builtin/packages/halide/package.py
wraith1995: @trws Is this fine?

PR #35256, opened 2023-01-31
Score-P updates for version 8.0

wrwilliams:
* Add new tarballs for Score-P and Cube releases. OTF2 3.0 already up-to-date.
* Added self as a maintainer.
* Added `+cuda` and `+hip` variants, with version checks.
* Deprecated everything pre-7.0. Using old Score-P versions is not a great idea.
* Added MPI standard version dependency where it applies (MPI 2.2+ for Score-P 7.0+)

There's still quite a bit of cleanup to do here but this should make the package considerably more usable already.


tldahlgren (commented):
Confirmed (or re-confirmed) all but one sha256.  I'm getting the following for one of them:

```
$ spack checksum cubew 4.7.1 
==> Error: Could not find any remote versions for cube
```

wrwilliams:
> Confirmed (or re-confirmed) all but one sha256. I'm getting the following for one of them:
> 
> ```
> $ spack checksum cubew 4.7.1 
> ==> Error: Could not find any remote versions for cube
> ```

Good catch. It should be there now.

Interesting if CubeLib 4.7.1 worked, according to my JSC colleagues it shouldn't have given that base URL.

wrwilliams:
@tldahlgren Just wanted to clarify: the 4.7.1 tarballs were mistakenly not uploaded where they should have been, and are now present. The current branch should now checksum correctly for cubew/cubelib 4.7.1 without further changes.

tldahlgren (approved):
(Re)confirmed all affected (31) sha256.

lrtfm:
@wrwilliams I made some comments here: https://github.com/spack/spack/commit/72f57ffedef280dd8c043d5f1ed44c9f37ace811, could you have a look? Thank you for your attention.

PR #35258, opened 2023-01-31
Add virtual information on DAG edges

### LINKED ISSUE #34886: Annotate virtual dependencies on DAG edges

alalazo:
see #34821

fixes #34886 

Currently virtual dependencies are used only during concretization to select a provider. No annotation is kept on edges to reflect that a leaf node is providing a virtual dependency. This PR adds virtual information on edges, so we can keep track of which spec is providing what virtual both in memory and in the JSON / YAML representation.

The way edge attributes are added is generic, so to reuse the same representation later when the `use_variant` will be implemented, see https://github.com/spack/seps/pull/2. A sample representation is:
```json
        "dependencies": [
          {
            "name": "cmake",
            "hash": "iqpoju67adln3yqzvqzccrtkjmpc666m",
            "parameters": {
              "deptypes": [
                "build"
              ],
              "virtuals": []
            }
          },
          {
            "name": "openmpi",
            "hash": "p23v372ij7bomrarjrhcv4uw6ieahjt3",
            "parameters": {
              "deptypes": [
                "build",
                "link"
              ],
              "virtuals": [
                "mpi"
              ]
            }
          },
          {
            "name": "pkgconf",
            "hash": "i4avrindvhcamhurzbfdaggbj2zgsrrh",
            "parameters": {
              "deptypes": [
                "run"
              ],
              "virtuals": [
                "pkgconfig"
              ]
            }
          },
          {
            "name": "zlib",
            "hash": "nizxi5u5bbrzhzwfy2qb7hatlhuswlrz",
            "parameters": {
              "deptypes": [
                "build",
                "link"
              ],
              "virtuals": []
            }
          }
        ],
```

Modifications:
- [x] Add a `parameters` attribute to `DependencySpec`, which can hold arbitrary attributes
- [x] Modify `Spec.to_node_dict` to output information on virtuals in the `"dependencies" section
- [x] Add unit tests (round-trip specs / check annotation)
- [x] Modified `DependencySpec` to hold an arbitrary number of attributes (98b98069df9357c9b6c97e8d8b98dc48fbba7d9a)
- [x] Reconstruct virtuals from old specfile formats (0b2db308fe19cfb74d705678ea17518cee71550e)


lib/spack/spack/spec.py
```diff
@@ -668,47 +668,81 @@ class DependencySpec:
...
+        parent: "Spec",
+        spec: "Spec",
+        *,
+        deptypes: dp.DependencyArgument,
+        virtuals: Tuple[str, ...],
+    ):
         self.parent = parent
         self.spec = spec
-        self.deptypes = dp.canonical_deptype(deptypes)
+        self.parameters = {
```
haampie, line 721: Is it really necessary to have a dict here? Personally I would keep redundancy & the number of indirections small so that we don't hit even more Python slowness. (One of the bottlenecks is creating the edges when deserializing the db, so if a normal prop works that's likely better than a string and hashmap lookup -- ok, to be fair, we do go from json string -> dict -> props, so in a sense it could also be faster)

lib/spack/spack/spec.py
```diff
@@ -4732,11 +4798,25 @@ def merge_abstract_anonymous_specs(*abstract_specs: Spec):
...
     return merged_spec
 
 
+def _reconstruct_virtuals_on_edges(spec):
+    """Reconstruct virtuals on edges. Used to read from old DB
+    and reindex.
+    """
+    possible_virtuals = [x for x in spec.package.dependencies if Spec(x).virtual]
+    for vspec in possible_virtuals:
+        if vspec in spec:
```
haampie: I think you want to restrict to direct deps here?

lib/spack/spack/spec.py
alalazo: Using properties is a step towards adding custom attributes on edges. This is needed for "use variants" as proposed in https://github.com/spack/seps/pull/2 and can be implemented on top of #35322, so 2 PRs from here.

lib/spack/spack/spec.py
alalazo: That should be done already, since the loop below selects only outgoing edges from the root.

haampie: Sure, but in case vspec is not a direct dep you walk the whole dag? And what if it is a dependency of a dependency?

alalazo: Turned out the reconstruction was incomplete. See https://github.com/spack/spack/pull/35258/commits/3f1205a946536ef3cbcac332c371405023c80741

PR #35262, opened 2023-01-31
Add py netpyne, meautility, and lfpykit

heerener:


var/spack/repos/builtin/packages/py-netpyne/package.py
```diff
@@ -0,0 +1,34 @@
...
+class PyNetpyne(PythonPackage):
+    """Netpyne: A python package to facilitate the development,
+    parallel simulation, optimization and analysis of multiscale
+    biological neuronal networks in NEURON."""
+
+    homepage = "http://www.netpyne.org/"
+    url = "https://github.com/suny-downstate-medical-center/netpyne/archive/refs/tags/v1.0.3.1.tar.gz"
+    git = "https://github.com/suny-downstate-medical-center/netpyne.git"
+
+    version("master", git=git, branch="master")
```
tldahlgren: The `git` argument is not needed.

```suggestion
    version("master", branch="master")
```

tldahlgren: See example at https://spack.readthedocs.io/en/latest/packaging_guide.html#fetching-from-code-repositories.

tldahlgren (commented):
Confirmed the new sha256 and branch exist.  Have a request about removing "git=git" from one of the version directives.

var/spack/repos/builtin/packages/py-netpyne/package.py
```diff
@@ -0,0 +1,34 @@
...
+
+    depends_on("py-numpy", type="run")
+    depends_on("py-scipy", type="run")
+    depends_on("py-matplotlib@:3.5.1", type="run")
+    depends_on("py-matplotlib-scalebar", type="run")
+    depends_on("py-future", type="run")
+    depends_on("py-pandas", type="run")
+    depends_on("py-bokeh", type="run")
+    depends_on("py-schema", type="run")
+    depends_on("py-lfpykit", type="run")
```
adamjstewart: Should all be type=build/run

var/spack/repos/builtin/packages/py-netpyne/package.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+    depends_on("python@3.7")
```
adamjstewart: Why?

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-netpyne/package.py
heerener: https://github.com/suny-downstate-medical-center/netpyne/blob/v1.0.3.1/setup.py#L66 setup.py seems quite specific about their desire for 3.7 (or earlier, but you've already explained that Spack dropped support for that).

adamjstewart: These classifiers have no effect on concretization, it's just a list of versions the authors have personally tested at the time of release. Unless you know for a fact that newer versions don't work, I would remove this line.

heerener: I most certainly don't know, so I've cleaned it up :-)

adamjstewart (approved):

PR #35267, opened 2023-01-31
Add trilinos-solvers variant to nalu-wind package.

alanw0:
This allows nalu-wind to be built against a trilinos installation which doesn't have amesos2, belos, ifpack2, or muelu enabled, if the nalu-wind user provides the spec 'nalu-wind@master~trilinos-solvers'

Support for these solver-packages remains on by default.

### DIFF
--- a/var/spack/repos/builtin/packages/nalu-wind/package.py
+++ b/var/spack/repos/builtin/packages/nalu-wind/package.py
@@ -64,7 +64,9 @@ class NaluWind(CMakePackage, CudaPackage):
     depends_on("tioga@master:", when="+tioga")
     depends_on("hypre@2.18.2: ~int64+mpi~superlu-dist", when="+hypre")
     depends_on("trilinos+muelu+belos+amesos2+ifpack2", when="+trilinos-solvers")
-    conflicts("~hypre~trilinos-solvers", msg="nalu-wind: Must enable either hypre or trilinos-solvers")
+    conflicts(
+        "~hypre~trilinos-solvers", msg="nalu-wind: Must enable either hypre or trilinos-solvers"
+    )
     depends_on("kokkos-nvcc-wrapper", type="build", when="+cuda")
     for _arch in CudaPackage.cuda_arch_values:
         depends_on(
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/nalu-wind/package.py
```diff
@@ -42,7 +42,8 @@ class NaluWind(CMakePackage, CudaPackage):
     )
     variant("openfast", default=False, description="Compile with OpenFAST support")
     variant("tioga", default=False, description="Compile with Tioga support")
-    variant("hypre", default=False, description="Compile with Hypre support")
+    variant("hypre", default=True, description="Compile with Hypre support")
+    variant("trilinos-solvers", default=True, description="Compile with Trilinos Solvers support")
```
tldahlgren, line 46: According to the message in conflict at lines 67-69, it sounds like one or the other of these should be enabled but not both.  

If that is the case, then one of these should default to `False`.  

And a `conflicts` could be added to ensure both aren't enabled at the same time.

tldahlgren (commented):
Have a couple of questions about combining the new variants.

var/spack/repos/builtin/packages/nalu-wind/package.py
psakievich: The behavior we want is actually the opposite. At least one needs to be enabled at all time.

var/spack/repos/builtin/packages/nalu-wind/package.py
```diff
@@ -62,6 +63,10 @@ class NaluWind(CMakePackage, CudaPackage):
     depends_on("openfast@2.6.0: +cxx", when="+openfast")
     depends_on("tioga@master:", when="+tioga")
     depends_on("hypre@2.18.2: ~int64+mpi~superlu-dist", when="+hypre")
+    depends_on("trilinos+muelu+belos+amesos2+ifpack2", when="+trilinos-solvers")
+    conflicts(
+        "~hypre~trilinos-solvers", msg="nalu-wind: Must enable either hypre or trilinos-solvers"
```
psakievich:
```suggestion
        "~hypre~trilinos-solvers", msg="Nalu-Wind must enable at least one of the linear solvers: hypre or trilinos-solvers"
```

psakievich (changes requested):

var/spack/repos/builtin/packages/nalu-wind/package.py
alanw0: Ah yes, that would be poor wording on my part. In fact it is ok to enable either of them, or both of them, but nalu-wind won't do anything useful if neither of them is enabled. I will incorporate Phil's suggestion below. Thanks!

psakievich (approved):

jrood-nrel (approved):

tldahlgren (approved):

PR #35268, opened 2023-01-31
tau: updating python version restriction

snehring:
Newer versions play nicer with python 3.9-3.10.

### DIFF
--- a/var/spack/repos/builtin/packages/tau/package.py
+++ b/var/spack/repos/builtin/packages/tau/package.py
@@ -111,7 +111,9 @@ class Tau(Package):
     # TAU requires the ELF header support, libiberty and demangle.
     depends_on("binutils+libiberty+headers+plugins", when="+binutils")
     # Build errors with Python 3.9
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/tau/package.py
```diff
@@ -111,7 +111,9 @@ class Tau(Package):
     # TAU requires the ELF header support, libiberty and demangle.
     depends_on("binutils+libiberty+headers+plugins", when="+binutils")
     # Build errors with Python 3.9
-    depends_on("python@2.7:3.8", when="+python")
+    depends_on("python@2.7:3.8", when="@:2.31+python")
```
adamjstewart:
```suggestion
    depends_on("python@2.7:3.8", when="@:2.31.0+python")
```
In Spack, 2.31.1 is included in `:2.31`, so this constraint would also be applied to 2.31.1. The easiest way to avoid this is to add a fake patch suffix.

adamjstewart (approved):

PR #35273, opened 2023-01-31
`spack develop`: convert to config

scheibelp:
This PR converts the 'develop' section of an environment to a dedicated configuration section. This means for example that instead of having to define `develop` specs in the `spack.yaml`, the environment can `include:` another `develop.yaml` configuration which specifies which specs should be developed in the environment.

This change is not expected to be disruptive given that existing environment `spack.yaml` files will conform to the new schema.

(Update 11/28/2023) I have implemented the `develop`/`undevelop` commands in terms of more-generic modification functions added to the `config` module: `change_or_add` and `update_all`. It is assumed that the semantics added here (described in 11/18 update) would be desirable to extend to other config update actions (e.g. adding compilers, changing package requirements, adding mirrors).

(Update 11/18/2023) I have updated this such that `spack develop`, and `spack undevelop` to potentially modify all writable scopes, like https://github.com/spack/spack/pull/41147. https://github.com/spack/spack/pull/35307 will be useful for modifying included scopes, but generally speaking specifying a `--scope` will not be required for `spack develop`: `spack develop` will add new develop specs to whatever scope already has develop specs defined, or to the highest-priority writable scope (which should be the env scope).

TODOs:

- [x] If you `spack undevelop` a package which is mentioned at multiple layers of configuration, then currently this would only modify one of them. That's not technically a new issue (has always existed for configuration modification), but may be confusing to users when presented via an interface other than `spack config set`
- [x] Need to add (or confirm) the ability to modify individual config files by providing a path (rather than using a scope identifier as a key to retrieve associated config).
  - (UPDATE: 2/3/23): see https://github.com/spack/spack/pull/35307 which updates `--scope` to allow absolute file paths
  - (UPDATE: 11/18/23): this has been made less important after copying the multi-scope-update approach from #41147
- [ ] Assuming the mechanics in the PR description are agreeable, I should test the following
  - [x] `spack develop` should add new develop specs to the scope that defines them (potentially skipping higher priority scopes to e.g. augment included scope files)
    - (UPDATE 11/28) New test added in https://github.com/spack/spack/pull/35273/files#diff-1219582010f89a4cb8ca88fdb7de56d08211ff6979e0dbd39c12b0e5318b1b35
  - [ ] `spack undevelop` should be able to remove developed specs from multiple scopes (no `--scope` option required)


scheibelp:
@spackbot help

scheibelp:
@spackbot fix style

scheibelp:
@spackbot fix style

haampie:
Can we step back a bit and think about the other keys in spack.yaml? In particular `view` :upside_down_face:

scheibelp:
> Can we step back a bit and think about the other keys in spack.yaml? In particular view

Of course - do you want to discuss this more at tomorrow's meeting?

The prompt for this PR is that users might want to develop a project and test it on several platforms: i.e. for a shared filesystem across platforms, they make one source directory, tweak it once, and then want to run tests for separate platforms on that same source directory; presumably each platform would have its own associated environment (for tweaking spec details as needed for that platform), so we are encouraged to make it easy for multiple environments to refer to the same in-development package (if they tracked the in-development packages separately, then they could become inconsistent as the user updates in-development specs for one).

Pulling environment-specific properties like develop into configuration can also potentially create confusion as well, e.g. if the user accidentally modifies the user scope and starts developing a package in other environments where they don't want to. Also, this would be the first case where we have a config section that only has use in the context of an environment.

For views: I'm not clear on why we'd want multiple environments to share the same view so I don't immediately see a reason to pull it out (in general as I state above, without a use case for it I'm inclined to not pull properties out of the env config since it creates additional opportunities for confusion). I *thought* someone else on the core team had mentioned this might be of interest though so there might already be an effort to extract it.

Regarding the other sections:

* includes: could be shared although generally this is for creating entries which point outside the environment directory, so already supports sharing. This could be useful for creating additional scopes outside the context of envs, but don't think I'd want to solve most config problems in this manner.
* definitions: also could be shared, although this section is complex enough I personally think the value of extracting it is limited

tgamblin (dismissed):

tgamblin:
@spackbot fix style

lib/spack/spack/config.py
```diff
@@ -927,6 +928,84 @@ def scopes():
     return CONFIG.scopes
 
 
+def writable_scopes() -> List[ConfigScope]:
+    """
+    Return list of writable scopes
+    """
+    return list(
+        reversed(
+            list(
```
tgamblin, line 937: minor style nit: I would make this one a list comprehension instead of a generator expression passed to a list constructor.  You can't do that with the outer one (it has to be a `list()` constructor, but I think this one's more readable as `[x for x in ...]`

lib/spack/spack/config.py
```diff
@@ -927,6 +928,84 @@ def scopes():
...
+                x
+                for x in CONFIG.scopes.values()
+                if not isinstance(x, (InternalConfigScope, ImmutableConfigScope))
+            )
+        )
+    )
+
+
+def writable_scope_names() -> List[str]:
+    return list(x.name for x in writable_scopes())
```
tgamblin, line 947: same here -- prefer `[]` if it can be a list comprehension.

lib/spack/spack/config.py
```diff
@@ -927,6 +928,84 @@ def scopes():
...
+def matched_config(cfg_path):
+    return [(scope, get(cfg_path, scope=scope)) for scope in writable_scope_names()]
+
+
+def change_or_add(section_name, find_fn, update_fn):
+    """Change or add a subsection of config, with additional logic to
+    select a reasonable scope where the change is applied.
+
+    Search through config scopes starting with the highest priority:
+    the first matching a criteria (determined by ``find_fn``) is updated;
```
tgamblin, line 959:
```suggestion
    the first matching a criterion (determined by ``find_fn``) is updated;
```

tgamblin (approved):
I think this is strictly an improvement as-is. The remaining tasks in the PR description can be done in a follow-on (e.g., undeveloping all scopes).  We can think about how best to generalize the logic there, but I think think it's actually nice that most of this now pulled out of `environment.py` and lives in the `develop` command.  I think moving `spack develop` mostly out of the environment (even though it builds on the env) cleans things up a bit and separates concerns better.

PR #35274, opened 2023-01-31
gitlab ci: Provide a knob to control untouched spec pruning

scottwittenburg:
When untouched spec pruning is enabled, specs possibly affected by a change cannot be pruned from a pipeline.

Previously spack looked at all specs matching changed package names, and traversed dependents of each, all the way to the environment root, to compute the set of environment specs possibly affected by a change (and thus, not candidates for pruning).

With this PR, when untouched spec pruning is enabled, a new environment variable can control how far towards the root spack traverses to compute the set of specs possibly affected by a change.  `SPACK_UNTOUCHED_PRUNING_DEPENDENT_DEPTH` can be set to any numeric value before the `spack ci generate` command is called to control this traversal depth parameter.  Setting it to "0" traverses only touched specs, setting it to "1" traverses only touched specs and their direct dependents, and so on.  Omitting the variable restuls in the previous behavior of traversing all the way to the root.  Setting it to a negative value means no traversal is done, and always yields an empty set of possibly affected specs (which would result in the max pruning possible).

### DIFF
--- a/lib/spack/spack/ci.py
+++ b/lib/spack/spack/ci.py
@@ -513,16 +513,28 @@ def compute_affected_packages(rev1="HEAD^", rev2="HEAD"):
     return spack.repo.get_all_package_diffs("ARC", rev1=rev1, rev2=rev2)
 
 
-def get_spec_filter_list(env, affected_pkgs):
+def get_spec_filter_list(env, affected_pkgs, dependent_traverse_depth=None):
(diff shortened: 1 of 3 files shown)


scottwittenburg:
fyi @zackgalbreath @kwryankrattiger

scottwittenburg:
@spackbot help

kwryankrattiger (approved):

PR #35276, opened 2023-01-31
fastqc: Use the spack built perl

glennpj:
The fastqc script was using the system perl. This PR sets the script to use the spack built/provided perl. This PR also removes the code that adds the java path. That should be handled by module loading as far as I know.


tldahlgren (approved):
LGTM

PR #9436, opened 2018-10-04
feature: add option to retain .vcs

matz-e:
Some packages require version control information to build, e.g., python
packages depending on `setuptools-scm`.  Spack will remove any VCS
directory when caching hashable stages like tags, this patch will
provide a parameter `clean=False` to the `version` function that will
retain `.git` etc.

Should fix #8746.

PR #35284, opened 2023-02-01
feature: add option to retain .vcs

### LINKED ISSUE #8746: packages using git information during build fail when cached
During caching, the `.git` directory is excluded, effectively archiving the package without using git archive for some reason.  The result is that packages that depend on git information to determine their version at configure time, like flux and possibly rose as in #3749, break when cached but not when checked out initially.

### Expected Result

That the cached version will be the same as the non-cached version.

### Actual Result

The `.git` directory is removed.

heerener:
fixes #8746

Picking up where @matz-e left off in #9436 , I've updated the proposed changes there based on the discussion, as well as added a test.

### DIFF
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -580,6 +580,8 @@ class VCSFetchStrategy(FetchStrategy):
 
     """
 
+    vcs_info_dir = ""
+
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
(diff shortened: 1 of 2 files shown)


heerener:
@alalazo is there something I should be doing to get this PR merged? Or is it a matter of waiting for 0.20.0?

heerener:
After solving the merge conflicts I'm wondering: the diff has become extremely small. Can this be included in the near future?
And I'm wondering even more: why does a PR that doesn't do much more than adding a test _lower_ the code coverage?

@alalazo

alalazo:
> And I'm wondering even more: why does a PR that doesn't do much more than adding a test lower the code coverage?

Cause macOS unit tests failed, and macOS coverage file wasn't uploaded as a result.

RemiLacroix-IDRIS:
Any chance this will get merged? I am also facing this issue.

matz-e:
We have been running with this for ~5 years, I think this would be neat to have for all the software that's not on PyPI but uses `setuptools-scm` and/or similar shenanigans in CMake.

PR #35285, opened 2023-02-01
feature: make several packages detectable

upsj:
This makes a handful of common packages detectable by adding their libraries/executable names. I checked with `apt` that the naming scheme of the DSOs matches the version numbers, as I had to exclude a few other libraries (libffi, libpciaccess, ...) for a differing ABI versioning scheme. For those, I guess one would have to parse the config headers.

### DIFF
--- a/var/spack/repos/builtin/packages/zlib/package.py
+++ b/var/spack/repos/builtin/packages/zlib/package.py
@@ -8,6 +8,7 @@
 # The AutotoolsPackage causes zlib to fail to build with PGI
 import glob
 import os
+import re
 
 import spack.build_systems.generic
 import spack.build_systems.makefile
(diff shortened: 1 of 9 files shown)


var/spack/repos/builtin/packages/zlib/package.py
```diff
@@ -123,7 +131,7 @@ def install(self, spec, prefix):
             "bin": glob.glob(os.path.join(build_dir, "*.dll")),
             "lib": glob.glob(os.path.join(build_dir, "*.lib")),
         }
-        compose_src_path = lambda x: os.path.join(build_dir, x)
+        def compose_src_path(x): return os.path.join(build_dir, x)
```
upsj: this was my IDE, I will revert the change :)

haampie:
I appreciate having more packages support external find, but the main caveat is that we have to ensure the `<pkg>-dev` system package is installed, not just the runtime `<pkg>` package.

So, basically, detect `libsomething.<so-extension>` without trailing ABI version `.x` or API version `.a.b.c`.

Also, what about macOS? Hard-coded .so isn't very helpful there.

alalazo:

PR #35296, opened 2023-02-02
Add pika 0.12.0 and pika-algorithms 0.1.1

msimberg:
This adds new versions for pika and pika-algorithms. pika-algorithms 0.1.0 is not compatible with pika 0.12.0 so adding a constraint for that. pika algorithms 0.1.1 is compatible with pika 0.11.0 and 0.12.0.

I'm also changing the default value of the `malloc` variant in pika to `mimalloc` (this reflects a change already made in pika a few versions ago). As far as I can tell this is a perfectly safe change to do, but is there a corner case I'm not taking into account by changing the default? Existing builds should be fine, new builds may end up using a different allocator than it was using before, but otherwise I can't see any problems with changing the default.


aurianer (approved):
LGTM thanks!

alalazo (approved):

PR #35301, opened 2023-02-02
Relocate cmake build dir

### LINKED ISSUE #34938: Paraview Support on Windows

johnwparent:
NOTE: This PR is experimental support for a potential new feature and is a WIP

On Windows prior to the Fall Creators update the maximum length for a given file path was capped at 260 characters. After than point a feature was added to remove the limit if a user changes a flag in their registry, however individual programs need to opt in to supporting thing behavior. The entire Windows build toolchain Spack is currently using supports this feature until you get to the compiler. cl.exe and link.exe do not include the long path support. This means we cannot have paths generated by spack that could push file paths beyond this 260 character limit. Between Spack's spec hashing in the stage, and CMake's build directory naming conventions, this is an easy limit to exceed with larger projects like Paraview, ITK, or VTK. 

This PR looks to circumvent the issue by introducing the concept for an "external" build directory where CMake builds are transparently carried out in a (shorter) build directory external to the Spack stage, which is then returned to its proper location in the spack stage once the build is complete.

### DIFF
--- a/etc/spack/defaults/windows/config.yaml
+++ b/etc/spack/defaults/windows/config.yaml
@@ -3,4 +3,4 @@ config:
   concretizer: clingo
   build_stage::
     - '$spack/.staging'
-  cmake_ext_build: "C:/test/spack"
+  cmake_ext_build: "C:/sp-build"
--- a/etc/spack/defaults/windows/config.yaml
+++ b/etc/spack/defaults/windows/config.yaml
@@ -3,4 +3,3 @@ config:
   concretizer: clingo
   build_stage::
     - '$spack/.staging'
-  cmake_ext_build: "C:/sp-build"
--- a/etc/spack/defaults/windows/config.yaml
+++ b/etc/spack/defaults/windows/config.yaml
@@ -3,3 +3,4 @@ config:
   concretizer: clingo
   build_stage::
     - '$spack/.staging'
+  cmake_ext_build: true
--- a/etc/spack/defaults/windows/config.yaml
+++ b/etc/spack/defaults/windows/config.yaml
@@ -3,4 +3,4 @@ config:
   concretizer: clingo
   build_stage::
     - '$spack/.staging'
-  cmake_ext_build: true
+  cmake_ext_build: "C:/test/spack"
--- a/lib/spack/spack/build_systems/cmake.py
+++ b/lib/spack/spack/build_systems/cmake.py
@@ -372,7 +372,12 @@ class CMakeBuilder(BaseBuilder):
     @property
     def build_directory(self):
         """Full-path to the directory to use when building the package."""
-        return os.path.join(self.pkg.stage.path, self.build_dirname)
+        stage_path = (
+            os.path.join(self.pkg.stage.path, self.build_dirname)
+            if not self.pkg.spec.dag_hash(7) in spack.stage.CMakeBuildStage.dispatch
+            else spack.stage.CMakeBuildStage.dispatch[self.pkg.spec.dag_hash(7)]
+        )
+        return stage_path
 
     def cmake_args(self):
         """List of all the arguments that must be passed to cmake, except:
--- a/lib/spack/spack/build_systems/cmake.py
+++ b/lib/spack/spack/build_systems/cmake.py
@@ -372,12 +372,10 @@ class CMakeBuilder(BaseBuilder):
     @property
     def build_directory(self):
         """Full-path to the directory to use when building the package."""
-        stage_build_path = self.pkg.stage.build_directory
-        # default build stage is in source, if this is the case
-        # establish in stage build dir. If stage returns a non
-        # source build directory, we know it's out of source
-        # so it's an acceptable build stage for CMake
-        if stage_build_path in self.pkg.stage.source_path:
+        stage_build_path = getattr(self.pkg.stage, "build_directory", None)
+        # If build directory is not defined by the stage we are free to define our own here
+        # Note: build_directory is only defined by a CMakeBuildStage
+        if not stage_build_path:
             stage_build_path = os.path.join(self.pkg.stage.path, self.build_dirname)
         return stage_build_path
 
--- a/lib/spack/spack/build_systems/cmake.py
+++ b/lib/spack/spack/build_systems/cmake.py
@@ -372,12 +372,14 @@ class CMakeBuilder(BaseBuilder):
     @property
     def build_directory(self):
         """Full-path to the directory to use when building the package."""
-        stage_path = (
-            os.path.join(self.pkg.stage.path, self.build_dirname)
-            if not self.pkg.spec.dag_hash(7) in spack.stage.CMakeBuildStage.dispatch
-            else spack.stage.CMakeBuildStage.dispatch[self.pkg.spec.dag_hash(7)]
-        )
-        return stage_path
+        stage_build_path = self.pkg.stage.build_directory
+        # default build stage is in source, if this is the case
+        # establish in stage build dir. If stage returns a non
+        # source build directory, we know it's out of source
+        # so it's an acceptable build stage for CMake
+        if stage_build_path in self.pkg.stage.source_path:
+            stage_build_path = os.path.join(self.pkg.stage.path, self.build_dirname)
+        return stage_build_path
 
     def cmake_args(self):
         """List of all the arguments that must be passed to cmake, except:
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -51,6 +51,7 @@ def install_kwargs_from_args(args):
         "fail_fast": args.fail_fast,
         "keep_prefix": args.keep_prefix,
         "keep_stage": args.keep_stage,
+        "cmake_stage": args.cmake_external_stage,
         "restage": not args.dont_restage,
         "install_source": args.install_source,
         "verbose": args.verbose or args.install_verbose,
@@ -114,7 +115,13 @@ the dependencies""",
         action="store_true",
         help="if a partial install is detected, don't delete prior state",
     )
-
+    subparser.add_argument(
+        "--cmake-external-stage",
+        action="store",
+        dest="cmake_external_stage",
+        help="""Path to root directory where CMake should build relevant
+                projects external to Spack's stage (no impact on non Windows systems)""",
+    )
     cache_group = subparser.add_mutually_exclusive_group()
     cache_group.add_argument(
         "--use-cache",
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -115,13 +115,6 @@ the dependencies""",
         action="store_true",
         help="if a partial install is detected, don't delete prior state",
     )
-    subparser.add_argument(
-        "--cmake-external-stage",
-        action="store",
-        dest="cmake_external_stage",
-        help="""Path to root directory where CMake should build relevant
-                projects external to Spack's stage (no impact on non Windows systems)""",
-    )
     cache_group = subparser.add_mutually_exclusive_group()
     cache_group.add_argument(
         "--use-cache",
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -114,6 +114,7 @@ the dependencies""",
         action="store_true",
         help="if a partial install is detected, don't delete prior state",
     )
+
     cache_group = subparser.add_mutually_exclusive_group()
     cache_group.add_argument(
         "--use-cache",
--- a/lib/spack/spack/cmd/install.py
+++ b/lib/spack/spack/cmd/install.py
@@ -51,7 +51,6 @@ def install_kwargs_from_args(args):
         "fail_fast": args.fail_fast,
         "keep_prefix": args.keep_prefix,
         "keep_stage": args.keep_stage,
-        "cmake_stage": args.cmake_external_stage,
         "restage": not args.dont_restage,
         "install_source": args.install_source,
         "verbose": args.verbose or args.install_verbose,
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -1603,18 +1603,18 @@ def from_url_scheme(url, *args, **kwargs):
 
     url = kwargs.get("url", url)
 
-    # in order to parse Windows absolute file urls correctly,
-    # we pass them through pathlib, which has support for parsing
-    # absolute file urls out of Windows paths.
-    # If pathlib raises a value error during the parsing,
-    # we likely passed it a valid url, a relative
-    # path, or something completely invalid.
-    # In each of those cases we can reliably fall back on
-    # urlparse's behavior and this extra parsing step is a no-op
     if sys.platform == "win32":
         try:
             url = pathlib.Path(url).as_uri()
         except ValueError:
+            # in order to parse Windows absolute file urls correctly,
+            # we pass them through pathlib, which has support for parsing
+            # absolute file urls out of Windows paths.
+            # If pathlib raises a value error during the parsing,
+            # we likely passed it a valid url, a relative
+            # path, or something completely invalid.
+            # In each of those cases we can reliably fall back on
+            # urlparse's behavior and this extra parsing step is a no-op
             pass
     parsed_url = urllib.parse.urlparse(url, scheme="file")
 
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -1603,13 +1603,14 @@ def from_url_scheme(url, *args, **kwargs):
 
     url = kwargs.get("url", url)
 
-    # in order to get allow the lines below to parse
-    # Windows absolute file urls correctly, we pass them through
-    # pathlib. If pathlib raises a value error there,
-    # we likely passed it a valid url or a relative
-    # path, both of which urlparse will handly perfectly fine
-    # for all other cases, the uri call will fail and this
-    # is effectively a no-op
+    # in order to parse Windows absolute file urls correctly,
+    # we pass them through pathlib, which has support for parsing
+    # absolute file urls out of Windows paths.
+    # If pathlib raises a value error during the parsing,
+    # we likely passed it a valid url, a relative
+    # path, or something completely invalid.
+    # In each of those cases we can reliably fall back on
+    # urlparse's behavior and this extra parsing step is a no-op
     if sys.platform == "win32":
         try:
             url = pathlib.Path(url).as_uri()
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -26,6 +26,7 @@ import copy
 import functools
 import os
 import os.path
+import pathlib
 import re
 import shutil
 import urllib.parse
@@ -1600,6 +1601,19 @@ def from_url_scheme(url, *args, **kwargs):
     in the given url."""
 
     url = kwargs.get("url", url)
+
+    # in order to get allow the lines below to parse
+    # Windows absolute file urls correctly, we pass them through
+    # pathlib. If pathlib raises a value error there,
+    # we likely passed it a valid url or a relative
+    # path, both of which urlparse will handly perfectly fine
+    # for all other cases, the uri call will fail and this
+    # is effectively a no-op
+    if is_windows:
+        try:
+            url = pathlib.Path(url).as_uri()
+        except ValueError:
+            pass
     parsed_url = urllib.parse.urlparse(url, scheme="file")
 
     scheme_mapping = kwargs.get("scheme_mapping") or {
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -29,6 +29,7 @@ import os.path
 import pathlib
 import re
 import shutil
+import sys
 import urllib.parse
 from typing import List, Optional
 
@@ -1609,7 +1610,7 @@ def from_url_scheme(url, *args, **kwargs):
     # path, both of which urlparse will handly perfectly fine
     # for all other cases, the uri call will fail and this
     # is effectively a no-op
-    if is_windows:
+    if sys.platform == "win32":
         try:
             url = pathlib.Path(url).as_uri()
         except ValueError:
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -582,16 +582,13 @@ def log(pkg):
             # Check that we are trying to copy things that are
             # in the stage tree (not arbitrary files)
             abs_expr = os.path.realpath(glob_expr)
-            if os.path.realpath(pkg.stage.path) not in abs_expr and not (
-                isinstance(pkg.builder, spack.build_systems.cmake.CMakeBuilder)
-                and spack.config.get("config:cmake_ext_build")
-            ):
+            if not pkg.stage.contains(abs_expr):
                 errors.write("[OUTSIDE SOURCE PATH]: {0}\n".format(glob_expr))
                 continue
             # Now that we are sure that the path is within the correct
             # folder, make it relative and check for matches
             if os.path.isabs(glob_expr):
-                glob_expr = os.path.relpath(glob_expr, pkg.stage.path)
+                glob_expr = pkg.stage.path_rel_to_stage(glob_expr)
             files = glob.glob(glob_expr)
             for f in files:
                 try:
@@ -1877,8 +1874,6 @@ class BuildProcessInstaller(object):
         # whether to keep the build stage after installation
         self.keep_stage = install_args.get("keep_stage", False)
 
-        self.cmake_build_stage = install_args.get("cmake_stage", "")
-
         # whether to skip the patch phase
         self.skip_patch = install_args.get("skip_patch", False)
 
@@ -1907,9 +1902,6 @@ class BuildProcessInstaller(object):
         """Main entry point from ``build_process`` to kick off install in child."""
 
         self.timer.start("stage")
-
-        if self.cmake_build_stage:
-            self.pkg.cmake_stage_dir = self.cmake_build_stage
 
         if not self.fake:
             if not self.skip_patch:
@@ -2396,7 +2388,6 @@ class BuildRequest(object):
             ("package_use_cache", True),
             ("keep_prefix", False),
             ("keep_stage", False),
-            ("cmake_stage", ""),
             ("restage", False),
             ("skip_patch", False),
             ("tests", False),
--- a/lib/spack/spack/installer.py
+++ b/lib/spack/spack/installer.py
@@ -45,6 +45,7 @@ from llnl.util.tty.color import colorize
 from llnl.util.tty.log import log_output
 
 import spack.binary_distribution as binary_distribution
+import spack.build_systems.cmake
 import spack.compilers
 import spack.error
 import spack.hooks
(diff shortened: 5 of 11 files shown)


lib/spack/spack/stage.py
```diff
@@ -725,6 +728,191 @@ def _add_to_root_stage(self):
...
+    def __init__(self, hash, name, root=None, keep=False):
+        # Users can override external cmake build dir, default is %USERPROFILE%
+        # overrides can come from command line or config, command line will override all
+        self._hash = hash
+        self._path = Path(get_stage_root(), name)
+        self._remote_stage = None
+        self.keep = keep
+        if not root:
+            fallback_path = Path(os.environ["USERPROFILE"], ".sp-stage")
+            self._root = Path(spack.config.get("config:cmake_ext_build_stage_dir", fallback_path))
```
scheibelp: This doesn't match the other names

etc/spack/defaults/windows/config.yaml
```diff
@@ -3,3 +3,4 @@ config:
   concretizer: clingo
   build_stage::
     - '$spack/.staging'
+  cmake_ext_build: true
```
scheibelp: Without updating `schemas/config.py`, the validation step would fail if a user ever set this value in the configuration.

Likewise the tests should exercise the use of this config variable (in particular, I argue elsewhere that I'd prefer to remove the ability to control this on the command line, so in that case the config would be the only place to set this).

lib/spack/spack/stage.py
```diff
@@ -725,6 +728,191 @@ def _add_to_root_stage(self):
...
+        # try to create root if it doesn't exist
+        self._root.mkdir(parents=True, exist_ok=True)
+        sub_dir = self._compute_next_open_subdir()
+        full_subdir = self._root / sub_dir
+        try:
+            full_subdir.mkdir()
+        except FileExistsError:
+            # another process must have created the same directory as us
+            # try again
+            return self._setup_remote_build_stage()
```
scheibelp: This would be easier to understand without the recursion like:

```
while True:
  subdir = next_open_subdir()
  try:
    subdir.mkdir()
    break
  except FileExistsError:
    pass
```

I also suggest establishing a maximum number of tries (like 200) and failing outright if that is reached.

lib/spack/spack/installer.py
```diff
@@ -588,7 +589,10 @@ def log(pkg):
             # Check that we are trying to copy things that are
             # in the stage tree (not arbitrary files)
             abs_expr = os.path.realpath(glob_expr)
-            if os.path.realpath(pkg.stage.path) not in abs_expr:
+            if os.path.realpath(pkg.stage.path) not in abs_expr and not (
+                isinstance(pkg.builder, spack.build_systems.cmake.CMakeBuilder)
+                and spack.config.get("config:cmake_ext_build")
+            ):
```
scheibelp: Ideally we'd create a method like `Stage.contains(path)` (or a higher-level `Package.contains(path)` depending on how things are refactored) and just call that rather than performing a complex check here.

lib/spack/spack/stage.py
scheibelp: It occurred to me that if each of the prospective build directories contains a file that includes the hash of the spec being built, then I think we can match the lifetime of the stages with the lifetime of the locks on spec prefixes for the build (admittedly that doesn't exist on Windows at the moment, but point being that once locks are restored for the prefixes, the consistency of the staging directory would be maintained "for free").

lib/spack/spack/package_base.py
```diff
@@ -1059,6 +1072,17 @@ def _make_stage(self):
...
+        # if we're building a CMake package on Windows
+        # and the user set the requsite config option, setup a
+        # custom CMake build stage to relocate the cmake build dir to
+        # add here to take advantage of stage cleanup
+        if (
+            is_windows
+            and self.spec.variants["build_system"].value == "cmake"
+            and spack.config.get("config:cmake_ext_build", False)
+        ):
+            composite_stage.append(self._make_cmake_build_stage())
```
scheibelp: If I understand, the primary value of appending the relocated stage to this composite is to ensure that it is destroyed in the same way as the rest of the resources. I see the value of that but am concerned about the implications of implementing a bunch of noop functionality for the other actions that are applied to this stage. I think ideally we could just handle this with an override of `Package.do_clean`

I think an alternative is to update the root stage itself to be a `CMakeBuildStage`:

* Inherits stage
* has an extra `.build_directory` property (and the `CMakePackage` can store a reference to it via the stage).
* `destroy` e.g. destroys the stage (that means the lifetime of the build directory matches the lifetime of the package stage)
* At the same time, most methods won't mention the `. build_directory` (which reflects the fact that it doesn't participate in most stage logic)

(To clarify, encapsulating behavior like finding a suitable subdirectory of a common shared root directory is fine to place in a separate class, the issue I'm trying to resolve is the potential confusion of treating this like a stage)

lib/spack/spack/stage.py
```diff
@@ -725,6 +728,191 @@ def _add_to_root_stage(self):
...
+                    teardown(sub_item)
+                else:
+                    sub_item.unlink()
+            pth.rmdir()
+
+        if self._remote_stage:
+            teardown(self._remote_stage)
+
+    def _reclaim_remote_stage(self):
+        # another Spack process or build has taken this directory
```
scheibelp: Was this logic created in response to an encountered failure, or an anticipated failure?

> another Spack process or build has taken this directory

We don't support concurrent runs of Spack on Windows at the moment

Regarding the latter problem: I think my suggestion elsewhere to track the spec hash as a file inside the build directory would avoid conflicts.

Approaching this from another angle: this function is used by `restage`, which is supposed to completely delete and recreate everything. In that case you can create an entirely new staging directory and delete the old one.

lib/spack/spack/stage.py
```diff
@@ -725,6 +728,191 @@ def _add_to_root_stage(self):
...
+
+    def __init__(self, hash, name, root=None, keep=False):
+        # Users can override external cmake build dir, default is %USERPROFILE%
+        # overrides can come from command line or config, command line will override all
+        self._hash = hash
+        self._path = Path(get_stage_root(), name)
+        self._remote_stage = None
+        self.keep = keep
+        if not root:
+            fallback_path = Path(os.environ["USERPROFILE"], ".sp-stage")
```
scheibelp: I'd prefer not having a fallback: raise an error if the alternative root is not configured. I'm trying to limit automatic usage of `~` (or an analog of it on Windows).

lib/spack/spack/cmd/install.py
```diff
@@ -115,7 +116,13 @@ def setup_parser(subparser):
         action="store_true",
         help="if a partial install is detected, don't delete prior state",
     )
-
+    subparser.add_argument(
+        "--cmake-external-stage",
```
haampie: Can you do without adding yet another flag to `install` please.

haampie (changes requested):

lib/spack/spack/cmd/install.py
johnwparent: Not a problem. I was err-ing on the side of usability but if the install CLI is already bloated, then yes, I don't see a reason this _needs_ to be there

scheibelp: I also agree about removing the flag.

scheibelp (changes requested):
I have some higher-level requests:

* I am concerned about implementing this dynamic build directory as a stage (when it generally doesn’t need to do everything a stage does) and in particular as a part of the component stage list - I have some suggestions about that
* I agree with Harmen about removing the command line option; the config right now is broken and needs to be fixed + tested
* I’d like to abstract some of the checks (e.g. that deal with logging in installer.py)

lib/spack/spack/stage.py
johnwparent: Sounds good. In that case I'll remove the other config option and just have the behavior enabled by whether or not a user sets the external path in the config.

etc/spack/defaults/windows/config.yaml
johnwparent: I've added a test to `test/stage.py` and will push with the rest of the requested changes. Let me know if there's a better place for such a test to live.

lib/spack/spack/stage.py
johnwparent: I agree with the second comment that this method essentially doesn't need to reclaim the stage, and instead can simply remake the build stage as we do during initial creation.

etc/spack/defaults/windows/config.yaml
```diff
@@ -3,3 +3,4 @@ config:
   concretizer: clingo
   build_stage::
     - '$spack/.staging'
+  cmake_ext_build: "C:/sp-build"
```
scheibelp: Do you think it's reasonable to set a default? That is definitely convenient, but I'm not sure if we should be automatically creating directories outside of Spack without the user knowing (`~` is an exception to this, but we're also trying to minimize/eliminate that).

lib/spack/spack/fetch_strategy.py
```diff
@@ -1618,6 +1619,19 @@ def from_url_scheme(url, *args, **kwargs):
     in the given url."""
 
     url = kwargs.get("url", url)
+
+    # in order to get allow the lines below to parse
```
scheibelp: Wording is confusing here.

Also: most pathlib manipulations are located in another PR, but I assume this was necessary for this PR.

lib/spack/spack/cmd/install.py
```diff
@@ -115,7 +115,6 @@ def setup_parser(subparser):
         action="store_true",
         help="if a partial install is detected, don't delete prior state",
     )
-
```
scheibelp: Vestigial edit: undo

lib/spack/spack/build_systems/cmake.py
```diff
@@ -372,7 +372,14 @@ def build_dirname(self):
     @property
     def build_directory(self):
         """Full-path to the directory to use when building the package."""
-        return os.path.join(self.pkg.stage.path, self.build_dirname)
+        stage_build_path = self.pkg.stage.build_directory
+        # default build stage is in source, if this is the case
+        # establish in stage build dir. If stage returns a non
```
scheibelp: This comment is confusing:

* This method is supposed to return a directory that enables out-of-source builds
* When using an external build directory root, we achieve this naturally
* If not, then we place the external build directory in the stage

Possible clarifications:

* when you refer to a "default" build stage, do you mean the default chosen by CMake? also I think by "build stage" you mean "build directory"

lib/spack/spack/stage.py
```diff
@@ -415,10 +418,25 @@ def expanded(self):
...
     @property
     def source_path(self):
         """Returns the well-known source directory path."""
-        return os.path.join(self.path, _source_path_subdir)
+        return str(self._path / _source_path_subdir)
+
+    @property
+    def build_directory(self):
+        """Returns a potential build directory, defaulting to in source build"""
+        return self.source_path
```
scheibelp: I'd prefer if the base `Stage` class did not store a notion of `build_directory`, but I could be overlooking some complexities of handling this specifically in `CMakeBuildStage`

lib/spack/spack/stage.py
```diff
@@ -725,6 +743,202 @@ def _add_to_root_stage(self):
...
+    def _reclaim_remote_stage(self):
+        # another Spack process or build has taken this directory
+        # the cmake build will not work from a different dir
+        # so wait until we can take it - try five times
+        # waiting a little longer each time.
+        # This will cause a hang but this should only be called if we're trying
+        # to rebuild a pre-existing stage, so we need to get the previous
+        # build dir or CMake will error
+        ii = 0
+        while self._remote_stage.exists() and ii < 5:
```
scheibelp: Depending on where the spack process is terminated (e.g. CTRL-C), `_remote_stage` may be `None`.

Also, how can it occur that the directory would still exist at this point? Is this because you could do something like:

```
# on windows
shutil.rmtree(a_directory)
mkdirp(a_directory) # this fails because the FS on windows hasn't actually deleted the directory
```

?

scheibelp (changes requested):

etc/spack/defaults/windows/config.yaml
johnwparent: Fair enough, removing. Will add a note in the getting started guide however.

lib/spack/spack/fetch_strategy.py
johnwparent: The "refactor the entire repo at once" effort became intractably arduous for one person to do and still accomplish other tasks in a reasonable amount of time so I am performing the effort in piecemeal now. I'll refactor the wording.

lib/spack/spack/build_systems/cmake.py
johnwparent: I think the refactor clarified this behavior to the point where we don't need too much of an elaboration, so I've removed much of this comment. Will push with rest of changes.

lib/spack/spack/stage.py
```diff
@@ -725,6 +738,148 @@ def _add_to_root_stage(self):
...
+        # try to create root if it doesn't exist
+        self._remote_root.mkdir(parents=True, exist_ok=True)
+        sub_dir = self._compute_next_open_subdir()
+        full_subdir = self._remote_root / sub_dir
+        attempts = 0
+        while True and attempts < 200:
+            try:
+                full_subdir.mkdir()
+                break
+            except FileExistsError:
```
scheibelp: I recommend removing the try/except (and the loop that re-attempts creating the file)

* You would need to put `_compute_next_open_subdir()` in this loop, or else it will fail 200 times
* This would need to raise an exception or else it will end up just returning the failed directory after reaching the max number of attempts (i.e. someone else's directory)
* Since `_compute_next_open_subdir` attempts to account for already-existing directories, the only time this exception would be triggered is in a multi-process environment. AFAIK `.mkdir` is not atomic though, so we'd have to be more-careful in this case anyway. It might be easier overall to assume serial and skip the try/except altogether (also we can assume that whatever root directory we are using, spack is the only one using it).

lib/spack/spack/fetch_strategy.py
```diff
@@ -1600,6 +1602,20 @@ def from_url_scheme(url, *args, **kwargs):
     in the given url."""
 
     url = kwargs.get("url", url)
+
+    # in order to parse Windows absolute file urls correctly,
+    # we pass them through pathlib, which has support for parsing
+    # absolute file urls out of Windows paths.
+    # If pathlib raises a value error during the parsing,
```
scheibelp: (minor) can you place this part of the comment explaining the error handling inside of the `except` block?

lib/spack/spack/package_base.py
```diff
@@ -1655,6 +1672,16 @@ def content_hash(self, content=None):
...
 
+    @property
+    def cmake_stage_dir(self):
+        if not getattr(self, "_cmake_build_stage", False):
+            return ""
+        return self._cmake_build_stage
+
+    @cmake_stage_dir.setter
+    def cmake_stage_dir(self, val):
+        self._cmake_build_stage = val
```
scheibelp: These functions (the getter and setter) don't appear to be used in this PR: do you anticipate them being useful later? I'd prefer to remove them to minimize API interactions with this functionality.

lib/spack/spack/stage.py
```diff
@@ -725,6 +738,148 @@ def _add_to_root_stage(self):
...
+    when MSVC fully supports the LongPath feature on Windows.
+    """
+
+
+    def __init__(self, hash, fetcher, **kwargs):
+        # Users can override external cmake build dir, default is %USERPROFILE%
+        # overrides can come from command line or config, command line will override all
+        super(CMakeBuildStage, self).__init__(fetcher, **kwargs)
+        self._hash = hash
+        self._remote_stage = None
```
scheibelp: Can you rename this to `_remote_build_dir`? I want to avoid using the token "stage" since that might be taken to imply that it is a `Stage` object when it is in fact a string (or maybe a `Path`, but something with less functionality than a `Stage` in any case).

Side note: I don't see anything setting this variable (e.g. `_remote_stage = ...`, other than when it is set to `None`)

lib/spack/spack/stage.py
```diff
@@ -725,6 +738,148 @@ def _add_to_root_stage(self):
...
+        self._remote_root = Path(spack.config.get("config:cmake_ext_build"))
+
+    @property
+    def root_stage_context(self):
```
scheibelp: Can we keep all new properties other than `build_directory` private? I don't think anything refers to this property other than tests.

lib/spack/spack/test/stage.py
```diff
@@ -902,8 +917,91 @@ def test_diystage_preserve_file(self, tmpdir):
...
+    def test_create_cmakebuildstage(self, cmake_ext_build, mock_stage_archive):
+        """Test that cmake external build dir is created during stage creation"""
+        archive = mock_stage_archive()
+        with CMakeBuildStage("jk6dj", archive.url, name=self.stage_name) as stage:
+            # external build dir should already have been created at this point
+            # ensure symlink to build dir exists
+            assert str(stage.root_stage_context.name) == "spack-build-jk6dj"
+            assert "spack-build-jk6dj" in os.listdir(stage.path)
+            # check symlink points to the right place
+            assert str(stage.root_stage_context.resolve()) == str(cmake_ext_build / "a")
```
scheibelp: I'm inclined to say that this should be testing `.build_directory` directly (or at least, we should be checking that in addition).

lib/spack/spack/test/stage.py
```diff
@@ -902,8 +917,91 @@ def test_diystage_preserve_file(self, tmpdir):
...
+            # ensure ext build dir was set honoring config
+            assert spack.config.get("config:cmake_ext_build") == str(stage._remote_stage.parent)
```
scheibelp: A couple minor requests:

* "Canonicalize" the configuration value by constructing a `Path` with it (avoid fragility due to string comparisons with paths with/without trailing `/`)
* swap order here (in general stick with `assert the-value-I-calculated == the-thing-I-expect-it-to-be`)

lib/spack/spack/test/stage.py
```diff
@@ -902,8 +917,91 @@ def test_diystage_preserve_file(self, tmpdir):
...
+            assert "spack-build-jk6dj" in os.listdir(stage.path)
+            # ensure actual external build dir is present
+            assert str(stage._remote_stage.name) in os.listdir(populated_cmake_ext_build)
+            # ensure remote stage has correct name based on already present stages
+            # the actual name matters less here than avoiding colisions between already
+            # present build dirs
+            assert str(stage._remote_stage) == str(populated_cmake_ext_build / "b")
+
+        # ensure ext build stage is destroyed on cleanup
+        assert not "b" is os.listdir(populated_cmake_ext_build)
```
scheibelp: I recommend grabbing `build_directory` from the `Stage` object vs. inferring the name based on how you know the underling generator works (if someone meddles with that, this check will silently become useless)

scheibelp (changes requested):
I have a number of requests (which I consider minor: hopefully you agree)

lib/spack/spack/package_base.py
johnwparent: This was originally how I had the builder accessing the stage during one of my earliest passes. It's been cruft for some time, happy to remove.

lib/spack/spack/stage.py
johnwparent: This variable is set in `create`

lib/spack/spack/test/stage.py
johnwparent: Because at this point the `build_directory` attribute is `None`. When the stage exists the context manager, its ext build stage associated properties are deallocated as they don't point to anything anymore. Should we refactor that behavior?

lib/spack/spack/stage.py
```diff
@@ -725,6 +737,138 @@ def _add_to_root_stage(self):
...
+                    raise
+            else:
+                self._remote_build_dir = self._root_stage_context.resolve()
+
+    def destroy(self):
+        self._destroy_remote()
+        super(CMakeBuildStage, self).destroy()
+
+    def path_rel_to_stage(self, glob_expr):
+        return os.path.relpath(glob_expr, self._remote_build_dir)
```
scheibelp: We spoke about how this avoids strange issues given that `glob_expr` would potentially be an absolute path that resides inside the `build_directory` - in this case distinct from the stage path.

I think we also briefly discussed the possibility that a user could request archiving a file that is not inside the `build_directory` (I know of no such case, but we don't prevent it). Does this logic address that somehow? If it doesn't and you don't want to consider that case, I think it's acceptable to raise an exception if the glob_expr is not a subdir of `_remote_build_dir`.

scheibelp: Oh and this is complex enough that IMO a docstring would be useful (i.e. this function can explain why it's calculating relative to this more-specific property)

johnwparent: I've added functionality to fall back on the default (pre `CMakeBuildStage`) behavior when an archive file falls outside of the scope of the `build_directory`, so that behavior will not change, as well as a relevant docstring.

lib/spack/spack/stage.py
```diff
@@ -725,6 +739,154 @@ def _add_to_root_stage(self):
...
+        """Destroy external build tree if not keep-stage
+        Otherwise this is kept as usual"""
+
+        def teardown(pth: Path):
+            for sub_item in pth.iterdir():
+                if sub_item.is_dir():
+                    teardown(sub_item)
+                else:
+                    sub_item.unlink()
+            pth.rmdir()
```
scheibelp: Could you use `shutil.rmtree`? This function appears to reimplement it

lib/spack/spack/stage.py
```diff
@@ -725,6 +739,154 @@ def _add_to_root_stage(self):
...
+        build directory"""
+        rel_to_stage = os.path.relpath(
+            os.path.join(
+                self._root_stage_context, os.path.relpath(glob_expr, self._remote_build_dir)
+            ),
+            self.path,
+        )
+        stage_root = super(CMakeBuildStage, self)
+        if stage_root.contains(os.path.realpath(glob_expr)):
+            rel_to_stage = stage_root.path_rel_to_stage(stage_root)
```
scheibelp, line 868: (minor) can you flip the order here?

```
def path_rel_to_stage:
    if stage_root.contains(os.path.realpath(glob_expr)):
        return stage_root.path_rel_to_stage(stage_root)
    else:
        rel_to_stage = ...
```

lib/spack/spack/stage.py
```diff
@@ -725,6 +739,154 @@ def _add_to_root_stage(self):
...
+    def path_rel_to_stage(self, glob_expr):
+        """Returns path of archive files relative to the root stage dir
+        from the remote build dir transparently as if there was a normal
+        build directory"""
+        rel_to_stage = os.path.relpath(
+            os.path.join(
+                self._root_stage_context, os.path.relpath(glob_expr, self._remote_build_dir)
+            ),
+            self.path,
+        )
```
scheibelp: IMO an explanatory comment would be useful here:

* first you calculate the path of the glob_expr relative to the `_remote_build_dir` (knowing that glob_expr is a subdir of it)
* Then, knowing that `_root_stage_context` is a symlink to `_remote_build dir`, you construct the symlink-based path and then calculate the relative path of that to `stage.path`

Personally, I think it would be more straightforward to use `str.replace()` like:

```
glob_expr.replace(self._remote_build_dir, self.root_stage_context)
```

the way you did it has an advantage that it avoids possible issues where the string representations of the path differs (but are equivalent paths), so perhaps this suggestion would only be useful as an explanatory comment (or a more-generic `replace_subpath` function could be added to `llnl.util.filesystem`).

lib/spack/spack/stage.py
```diff
@@ -725,6 +739,154 @@ def _add_to_root_stage(self):
...
+    def build_directory(self):
+        return self._remote_build_dir
+
+    def _establish_context_link(self):
+        symlink(str(self._remote_build_dir), str(self._root_stage_context))
+
+    def _remove_context_link(self):
+        self._root_stage_context.unlink()
+
+    def _setup_remote_build_stage(self):
```
scheibelp: Can you rename this to `_setup_remote_build_dir` (in keeping with replacing `stage` with `dir`)?

lib/spack/spack/stage.py
```diff
@@ -725,6 +739,154 @@ def _add_to_root_stage(self):
...
+        or the remote build dir"""
+        # Check if file is in root stage
+        cont = super(CMakeBuildStage, self).contains(file_path)
+        # if not, check if it's in remote build dir
+        if not cont:
+            cont = str(self._remote_build_dir) in file_path
+        return cont
+
+    def create(self):
+        if not self.created:
```
scheibelp: Requests:

* I don't think this should condition on `if self.created`
* We should not set `self.created` to `False` if there is an exception
* (less important) I think the `_remote_build_dir` would be cleaned up as part of `destroy`, so I don't think you need the `try/except` either

lib/spack/spack/package_base.py
```diff
@@ -1024,13 +1024,31 @@ def _make_root_stage(self, fetcher):
...
+        # and the user set the requsite config option, setup a
+        # custom CMake build stage to relocate the cmake build dir to
+        # add here to take advantage of stage cleanup
+        if (
+            sys.platform == "win32"
+            and self.spec.variants["build_system"].value == "cmake"
+            and spack.config.get("config:cmake_ext_build", "")
+        ):
+            stage = CMakeBuildStage(
+                self.spec.dag_hash(7),
```
scheibelp, line 1037: I don't think it's necessary to require the user to pass a hash (apologies if you got that impression from earlier discussions we had, since I implied it might be useful): the symlink could have the same name for all stages.

scheibelp (changes requested):
Mostly minor requests.

lib/spack/spack/package_base.py
johnwparent: Originally I put the hash parameters in, in anticipation of the hash file, but I left it in order to maintain consistency with the build artifacts. In a normal stage with a CMake build system, the artifact is placed in `<spack-archive-file-root>/<build-dir-name>/CMakeCache.txt`, so in keeping the naming scheme for the build dir consistent, we maintain this `CMakeBuildStage` as effectively transparent to the end user.

lib/spack/spack/stage.py
johnwparent: I'll leave the `try/except` as if there is an exception in the remote build dir creation/link establishment, if the remote dir is created, it will not be cleaned up. The rest of the changes have been addressed.

lib/spack/spack/stage.py
johnwparent: At one point this had some extra side effect, but that's long gone at this point. Refactored.

lib/spack/spack/stage.py
johnwparent: I'll add the replace as a comment. I'd be fine with the string comp if packages didn't define their own archive files, but since they can, there's little way to protect against differently formed strings until we convert to pathlib.

lib/spack/spack/package_base.py
scheibelp: OK thanks for the explanation: it makes sense to do that to maintain the file layout and naming scheme.

johnwparent: Alternatively, the artifact dir is already scoped under the packages install prefix, so we could just refactor this and the default approach.

lib/spack/spack/stage.py
scheibelp: I assume this function can be removed now

johnwparent: 🤦‍♂️ ... you assume correctly...

johnwparent: Addressed and pushed.

scheibelp:
Closing since https://github.com/spack/spack/pull/36509 has been merged and addresses the same issue.

PR #35302, opened 2023-02-02
py-minkowskiengine: new package (sparse tensor autodiff by Nvidia)

wdconinc:
This python package (with cuda support) provides torch support for sparse tensors. The `pybind11` headers are not found without the patch to `setup.py`.

### DIFF
--- a/var/spack/repos/builtin/packages/py-minkowskiengine/package.py
+++ b/var/spack/repos/builtin/packages/py-minkowskiengine/package.py
@@ -1,0 +1,36 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyMinkowskiengine(PythonPackage, CudaPackage):
+    """The MinkowskiEngine is an auto-differentiation
+    library for sparse tensors."""
+
+    homepage = "https://nvidia.github.io/MinkowskiEngine/"
+    pypi = "MinkowskiEngine/MinkowskiEngine-0.5.4.tar.gz"
+
+    maintainers = ["wdconinc"]
+
+    version("0.5.4", sha256="b1879c00d0b0b1d30ba622cce239886a7e3c78ee9da1064cdfe2f64c2ab15f94")
+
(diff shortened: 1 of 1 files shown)


wdconinc:
@spackbot fix style

wdconinc:
@adamjstewart FYI, another ML tool in the py-torch ecosystem. I have no CUDA system to test this on, unfortunately.

var/spack/repos/builtin/packages/py-minkowskiengine/package.py
```diff
@@ -0,0 +1,38 @@
...
+    homepage = "https://nvidia.github.io/MinkowskiEngine/"
+    pypi = "MinkowskiEngine/MinkowskiEngine-0.5.4.tar.gz"
+
+    maintainers = ["wdconinc"]
+
+    version("0.5.4", sha256="b1879c00d0b0b1d30ba622cce239886a7e3c78ee9da1064cdfe2f64c2ab15f94")
+
+    depends_on("py-setuptools", type="build")
+
+    depends_on("py-torch", type=("build", "run"))
```
adamjstewart, line 24: Missing numpy dependency

var/spack/repos/builtin/packages/py-minkowskiengine/package.py
```diff
@@ -0,0 +1,38 @@
...
+    depends_on("py-pybind11", type=("build", "run"))
```
adamjstewart: If you need to link to the header files, this should be `type="link"`. Not sure if build is also needed (PATH, PYTHONPATH), but run definitely isn't. I would try link and remove the patch and see if things work.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-minkowskiengine/package.py
wdconinc: Added `"py-numpy", type=("build", "run")`.

var/spack/repos/builtin/packages/py-minkowskiengine/package.py
wdconinc: Indeed, works with just `type="link"`.

wdconinc:
@spackbot fix style

adamjstewart (approved):

PR #35305, opened 2023-02-02
Windows: Warn when extracting long paths

johnwparent:


kwryankrattiger:
Do we still need this?

johnwparent:
Yes, very much still needed. Will be until sometime after 2028 if Windows is on target.

PR #35307, opened 2023-02-02
[WIP] "spack config": allow modifying files directly

scheibelp:
The 'spack config' --scope option can now accept a full file path, which allows modifying/viewing configurations without spack being otherwise aware of them; for example this could be use to easily modify configurations that are included in an environment's spack.yaml, or modifying the environment's spack.yaml without activating it.

For example:

```
spack -d config --scope=/absolute/path/to/spack/var/spack/environments/t1/spack.yaml add config:dirty:true
```

If you have a path that matches a scope name (e.g. `user/linux`) then - in the current implementation of this feature - Spack will ignore the path if it is supplied as `user/linux`: to differentiate with a relative path in this case, you can use the absolute path or `./user/linux` (still relative, but not a scope name since no scope names start with `./`).

(EDIT 7/12/23: below is old)

~Since scope identifiers can contain `/` (e.g. `platforms/linux`), I've opted to require that path-based scopes to be absolute paths, whose syntax is always distinct from scopes:~

* ~On Mac/Linux, starts with a `/`~
* ~On Windows, starts with a drive letter and `\` like `C:\`~


scheibelp:
@spackbot fix style

haampie:
Can you avoid the abs path requirement? Notice that we have this ambiguity in tons of places, like `spack env activate <env name or path>` and `spack buildcache create <mirror name or path>` and in both cases relative paths are supported. The solution is: if it's a known name, use it as a name, otherwise interpret it as a path.

scheibelp:
> Can you avoid the abs path requirement? Notice that we have this ambiguity in tons of places, like `spack env activate <env name or path>` and `spack buildcache create <mirror name or path>`

That could work although I'll note that the exact ~semantics~ behavior slightly differs even between those two commands:

* Mirror names cannot contain `/` or `\`, so `buildcache create` recognizes any of those as paths. Anything else is assumed to be a mirror name (even if e.g. there is a directory named `foo` in your CWD, it will fail if you refer to it as `foo`, so in that sense behaves similar to this PR)
* `env activate` appears to be similar:
  * say you have a CWD with `./t1`
  * `spack env create t1` will ignore `./t1`
  * `spack env create ./t1` will fail (a regex forbids `/`)
  * `spack env create -d ./t1` succeeds (the ambiguity is resolved by adding an option)

That being said, I think we could tweak the behavior in this case to allow a more-lenient fallback scheme while minimizing ambiguities (albeit different again from `spack env` and `spack buildcache`):

1. We could assume path if it exists (relative or otherwise): if the user makes a path called `user` then they could not modify the user scope while `./user` is in their CWD
2. Reversing that, we could prioritize treating the scope as a name: in which case we can't have a directory scope called `./user` in our CWD
  * However, if the user writes `--scope=./user` (vs. just `--scope=user`) then we could fall back to looking for dirs when the provided scope isn't an identifier
  * That being said, this could still cause confusion for users (while the prior point gives users the power to select all scopes, it does not prevent the potential confusion) so it might be best to require any relative path to begin with `./`

scheibelp:
@haampie I implemented item [2] from https://github.com/spack/spack/pull/35307#issuecomment-1428538778 - I think this is ready for a review.

PR #35322, opened 2023-02-03
Cherry-picking virtual dependencies

### LINKED ISSUE #15443: Finer selection of virtual providers in a DAG
As a user I want to be able to use a package only for a subset of the virtual dependencies it provides so that I can select other providers for whatever virtual dependency is not used.

### Rationale
When a spec providing multiple virtual dependencies is present in a DAG Spack enforces the use of all of them, if required by a dependee. To make a concrete example consider  `intel-parallel-studio`: 

https://github.com/spack/spack/blob/20c77cade532844c758290c242536436cecfe922/var/spack/repos/builtin/packages/intel-parallel-studio/package.py#L176-L185

This package provides multiple virtual dependencies and it would be currently impossible to require the build of a spec which uses:
1. `intel-parallel-studio` to satisfy `mpi` 
2. `openblas` to satisfy `lapack`

This proposal aims at solving this issue by introducing a syntax to specify the binding of a virtual dependency to a particular provider.

### Description
Spack should be extended to allow specs like:
```
dealii ^[virtuals=mpi] intel-parallel-studio ^[virtuals=lapack] openblas
```
to mean that `dealii` should be compiled to use the `mpi` from `intel-parallel-studio` and the `lapack` from `openblas`. More in details:

1. The provider can be specified as an abstract spec (in which case some concretizer heuristics will kick-in) or as a concrete spec by hash.
2. Packages need a way to specify that a set of virtual dependencies needs to be used together (e.g. `openblas` provides both `lapack` and `blas` but within a unique library)

### Additional information
This modification will be instrumental to move towards compilers as regular dependencies, since compilers can be modeled as providers of languages and we want to retain the ability to mix them (e.g. `clang` for C and C++, `gcc` for Fortran):
```
hdf5 %[virtuals=cc,cxx] clang %[virtuals=fortran] gcc
```

### Task list

- [x] #34886

### General information

- [x] I have searched the issues of this repo and believe this is not a duplicate

alalazo:
depends on #39508 
fixes #15443

This PR makes it possible to select only a subset of virtual dependencies from a spec that _may_ provide more. To select providers, a syntax to specify edge attributes is introduced:
```
hdf5 ^[virtuals=mpi] mpich
```
With that syntax we can concretize specs like:
```console
$ spack spec strumpack ^[virtuals=mpi] intel-parallel-studio+mkl ^[virtuals=lapack] openblas
```
and get this concrete DAG:

strumpack_virtuals

On `develop` this would currently fail with:
```console
$ spack spec strumpack ^intel-parallel-studio+mkl ^openblas
==> Error: Spec cannot include multiple providers for virtual 'blas'
    Requested 'intel-parallel-studio' and 'openblas'
```

In package recipes, virtual specs that are declared in the same `provides` directive need to be provided _together_. This means that e.g. `openblas`, which has:
```python
provides("blas", "lapack")
```
needs to provide both `lapack` and `blas` when requested to provide at least one of them.

## Additional notes

This capability is needed to model compilers. Assuming that languages are treated like virtual dependencies, we might want e.g. to use LLVM to compile C/C++ and Gnu GCC to compile Fortran. This can be accomplished by the following[^1]:
```
hdf5 ^[virtuals=c,cxx] llvm ^[virtuals=fortran] gcc
```

[^1]: We plan to add some syntactic sugar around this syntax, and reuse the `%` sigil to avoid having a lot of boilerplate around compilers.

Modifications:
- [x] Add syntax to interact with edge attributes from spec literals
- [x] Add concretization logic to be able to cherry-pick virtual dependencies
- [x] Extend semantic of the `provides` directive to express when virtuals need to be provided together
- [x] Add unit-tests and documentation


alalazo:
@spackbot run pipeline

RMeli:
Hello. I tried this PR with the following spec:

```bash
cp2k@master +dlaf +libxc +libint +spglib build_system=cmake ^[virtuals=blas,lapack,scalapack,mkl]
    intel-mkl@2020.4.304 threads=openmp ^fftw ^cray-mpich
```

It returned this error, which is very cryptic:
```text
==> Error: 
        cp2k@master+dlaf+libint+libxc+spglib build_system=cmake ^cray-mpich ^fftw ^[virtuals=blas,lapack,mkl,scalapack] intel-mkl@2020.4.304 threads=openmp is unsatisfiable, errors are:

```

That is, no error messages are actually printed for `errors are:`.

However, by removing `mkl` I was able to concretise the environment with both `intel-mkl` and `fftw` and everything seems to work as expected:
```bash
cp2k@master +dlaf +libxc +libint +spglib build_system=cmake ^[virtuals=blas,lapack,scalapack]
    intel-mkl@2020.4.304 threads=openmp ^fftw ^cray-mpich
```

This is a really neat and useful feature.

alalazo:
@RMeli Thanks for the feedback! Error messages are a pain point when working with ASP, we'll try to improve them considerably for v0.21 (so roughly going to November 2023) :slightly_smiling_face:

lib/spack/spack/parser.py
```diff
@@ -298,6 +307,18 @@ def next_spec(
...
+    def _parse_node(self, root_spec):
+        dependency = SpecNodeParser(self.ctx).parse()
+        if dependency is None:
+            msg = (
+                "this dependency sigil needs to be followed by a package name "
+                "or a node attribute (version, variant, etc.)"
+            )
+            raise SpecParsingError(msg, self.ctx.current_token, self.literal_str)
+        if root_spec.concrete:
+            raise spack.spec.RedundantSpecError(root_spec, "^" + str(dependency))
```
becker33, line 312: Isn't this condition impossible now that we defer evaluating parsed hashes until concretization time?

lib/spack/spack/parser.py
```diff
@@ -278,7 +281,13 @@ def next_spec(
...
         root_spec = SpecNodeParser(self.ctx).parse(initial_spec)
         while True:
-            if self.ctx.accept(TokenType.DEPENDENCY):
+            if self.ctx.accept(TokenType.START_EDGE_PROPERTIES):
+                edge_properties = EdgeAttributeParser(self.ctx, self.literal_str).parse()
+                dependency = self._parse_node(root_spec)
+                edge_properties.setdefault("deptypes", ())
+                edge_properties.setdefault("virtuals", ())
+                root_spec._add_dependency(dependency, **edge_properties)
+            elif self.ctx.accept(TokenType.DEPENDENCY):
```
becker33: I think this could be rewritten for more code reuse if you first check for either `DEPENDENCY` or `START_EDGE_PROPERTIES`.

```suggestion
            if self.ctx.accept(TokenType.Dependency) or self.ctx.accept(TokenType.START_EDGE_PROPERTIES):
                edge_properties = {}
                if self.ctx.token.kind == TokenType.START_EDGE_PROPERTIES:
                    edge_properties = EdgeAttributeParser(self.ctx, self.literal_str).parse()
                edge_properties.setdefault("deptypes", ())
                edge_properties.setdefault("virtuals", ())
```
And then the line 294/303 call to `_add_dependency` would change to:
```
root_spec._add_dependency(dependency, **edge_properties)
```

And then the entire `_parse_node` method that you had to add becomes redundant.

lib/spack/spack/solver/concretize.lp
```diff
@@ -293,18 +293,42 @@ error(1, Msg) :- attr("node", Package),
...
+  :- attr("hash", Package, Hash),
+     imposed_constraint(Hash, "virtual_on_incoming_edges", Package, Virtual),
+     attr("virtual_node", Virtual).
+
+provider(Package, Virtual2)
+  :- provider(Package, Virtual1),
+     condition_holds(ID),
+     provided_together(ID, SetID, Package, Virtual1),
+     provided_together(ID, SetID, Package, Virtual2),
+     Virtual1 != Virtual2.
```
becker33: Is this important for correctness or is it just breaking symmetry for performance?

lib/spack/spack/solver/concretize.lp
```diff
@@ -323,32 +355,37 @@ error(100, "Cannot select a single provider for virtual '{0}'", Virtual)
   :- attr("virtual_node", Virtual),
      2 { provider(P, Virtual) }.
 
+error(100, "Spec cannot include multiple providers for virtual '{0}'. Requested '{1}' and '{2}'", Virtual, P1, P2)
+  :- attr("virtual_node", Virtual),
+     provider(P1, Virtual),
+     provider(P2, Virtual),
+     P1 < P2.
```
becker33: Didn't we decide that cardinality rules like below are more performant?

```suggestion
     2 { provider(Provider, Virtual) }.
```

lib/spack/spack/solver/concretize.lp
```diff
@@ -323,32 +355,37 @@ error(100, "Cannot select a single provider for virtual '{0}'", Virtual)
...
+% If a package meet the condition to provide a virtual dependency, it may be chosen as a provider
+{ provider(Package, Virtual) } :- attr("node", Package), virtual_condition_holds(Package, Virtual).
+
 % If we asked for a root package and that root provides a virtual,
 % the root is a provider for that virtual. This rule is mostly relevant
 % for environments that are concretized together (e.g. where we
 % asks to install "mpich" and "hdf5+mpi" and we want "mpich" to
 % be the mpi provider)
-provider(Package, Virtual) :- attr("node", Package), virtual_condition_holds(Package, Virtual).
+provider(Package, Virtual) :- attr("root", Package), virtual_condition_holds(Package, Virtual).
```
becker33: This seems to imply that we may need a virtual-selection mechanism in unified environments someday. But that's hopefully out of scope for this PR.

lib/spack/spack/solver/concretize.lp
```diff
@@ -323,32 +355,37 @@ error(100, "Cannot select a single provider for virtual '{0}'", Virtual)
...
 % The provider provides the virtual if some provider condition holds.
 virtual_condition_holds(Provider, Virtual) :-
    provider_condition(ID, Provider, Virtual),
    condition_holds(ID),
    virtual(Virtual).
 
-% A package cannot be the actual provider for a virtual if it does not
-% fulfill the conditions to provide that virtual
-:- provider(Package, Virtual), not virtual_condition_holds(Package, Virtual),
-   internal_error("Virtual when provides not respected").
```
becker33: This is still true, right? Just not necessary for correctness?

If so, I think it's probably better to leave it in -- IIUC integrity constraints like this will improve performance, because it can short-circuit out of directions that will never work.

lib/spack/spack/spec.py
```diff
@@ -4135,11 +4151,17 @@ def __getitem__(self, name):
                 itertools.chain(
                     # Regular specs
                     (x for x in order() if x.name == name),
+                    (
+                        x
+                        for x in order()
+                        if (not x.virtual)
+                        and any(name in edge.virtuals for edge in x.edges_from_dependents())
+                    ),
```
becker33, line 4176: Shouldn't the line after this be unreachable, assuming we've properly constructed virtuals on edges for old specs that we read in?

lib/spack/spack/spec.py
```diff
@@ -147,14 +147,14 @@
 
 #: Default format for Spec.format(). This format can be round-tripped, so that:
 #:     Spec(Spec("string").format()) == Spec("string)"
-default_format = (
+DEFAULT_FORMAT = (
```
becker33: Is there a reason for this change? It introduces a bunch of orthogonal changes into the diff for this PR

lib/spack/spack/test/build_environment.py
```diff
@@ -541,3 +541,13 @@ def test_setting_attributes(self, default_mock_concretization):
             if current_module == spack.package_base:
                 break
             assert current_module.SOME_ATTRIBUTE == 1
+
+
+def test_monkey_patching_works_across_virtual(default_mock_concretization):
```
becker33, line 647: Why not use monkeypatch if the purpose is to test monkeypatch?

var/spack/repos/builtin/packages/strumpack/package.py
```diff
@@ -67,7 +67,7 @@ class Strumpack(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("mpi", when="+mpi")
     depends_on("blas")
     depends_on("lapack")
-    depends_on("openblas threads=openmp", when="^openblas")
+    depends_on("openblas threads=openmp", when="^[virtuals=lapack,blas] openblas")
```
becker33: I don't follow from the concretizer changes why this is necessary, and it concerns me a bit. Does it matter which virtuals openblas is providing whether this requirement holds?

This brings to mind a related worry -- if `foo` depends unconditionally on `openblas` (not through the virtual interface, and `bar` depends on `foo` and `blas`, does this implementation prevent `bar ^[virtuals=blas] intel-parallel-studio`? I would think it should (since `foo` is linking `libopenblas.so`).

lib/spack/spack/test/spec_semantics.py
```diff
@@ -1003,6 +1003,103 @@ def test_spec_override(self):
...
+        for label, expected in specs_in_dag:
+            assert label in s
+            assert s[label].satisfies(expected), label
+
+    @pytest.mark.parametrize(
+        "spec_str",
+        [
+            # openblas-with-lapack needs to provide blas and lapack together
+            "netlib-scalapack ^[virtuals=blas] intel-parallel-studio ^openblas-with-lapack",
+            # intel-* provides blas and lapack together, openblas can provide blas only
```
becker33, line 1096: Isn't this description reversed?

lib/spack/spack/test/spec_semantics.py
```diff
@@ -1003,6 +1003,103 @@ def test_spec_override(self):
...
+                    ("mpi", "mpich"),
+                    ("lapack", "openblas-with-lapack"),
+                    ("blas", "openblas-with-lapack"),
+                ],
+            ),
+            # Test that we can mix dependencies that provide an overlapping
+            # sets of virtual dependencies
+            (
+                "netlib-scalapack ^[virtuals=mpi] intel-parallel-studio "
+                "^[virtuals=lapack] openblas-with-lapack",
```
becker33, line 1052: Assuming `netlib-scalapack ^intel-parallel-studio ^openblas-with-lapack` would generate an error, can we add a test confirming that?

becker33 (dismissed):

lib/spack/spack/test/build_environment.py
alalazo: Because we want to test monkey-patching a package like a user would do, not with an internal mechanism used by pytest.

lib/spack/spack/solver/concretize.lp
alalazo: Outdated by rebase

lib/spack/spack/spec.py
alalazo: I'll extract that. The reason is to make apparent where we are using global constants in the code.

lib/spack/spack/test/spec_semantics.py
alalazo: No, for some historical reason I don't have an explanation for, in our mock repository `openblas` provides blas only and we have `openblas-with-lapack` that provides both together.

haampie (dismissed):
Mostly a bike shedding comment:

I would reserve `[` and `]` for other features. Cause (a) it could be used to express constraints that aren't flat `a ^[b ^[c ^d]]`, (b) other obvious brackets need escaping (c) you can do the same with just one character instead of two

So I'd suggest ```pkg_a ^ :type=link :virtual=mpi pkg_b```. (Or maybe even `pkg_a ^pkg_b :virtuals=mpi ordinary_variant=x`)

alalazo:
> I would reserve [ and ] for other features.

Whatever we choose, either `[` and `]` or `:`, I would use that as a delimiter and group edge attributes immediately after `^`. This would slightly simplify the parser but more importantly it wouldn't allow writing things like:
```
pkg_a ^openblas +foo :+bar fee=baz :baz=fu 
```
where a spec continuously interchange between edge and node variants. Also, we don't need to give meaning to expressions like:
```
pkg_a :virtual=mpi
```
which might be allowed with a single character.

alalazo:
@spackbot run pipeline

lib/spack/spack/parser.py
alalazo: I think so, but note that this was copied verbatim from our current develop. I would rather solve this in another PR, where we can discuss if we want `/abc ^zlib` to be valid or not.

lib/spack/spack/parser.py
alalazo: I just reused `_parse_node` here (forgot to do it before)

lib/spack/spack/solver/concretize.lp
alalazo: It's important for correctness. If 2 or more variants are provided together, and one of them isn't "provided" it's an error.

alalazo: We can probably find smarter ways to write this rule, if this starts to be an issue for performance

lib/spack/spack/solver/concretize.lp
alalazo: Outdated by rebase

lib/spack/spack/solver/concretize.lp
alalazo: Oudated by rebase

var/spack/repos/builtin/packages/strumpack/package.py
alalazo: > I don't follow from the concretizer changes why this is necessary, and it concerns me a bit. Does it matter which virtuals openblas is providing whether this requirement holds?

It is not necessary, but having or not the virtuals has an effect. I added a proper test in https://github.com/spack/spack/pull/35322/commits/bfce9f8add86c667f34ef2533a12aa77e775626a to check that virtuals on edges are accounted for in conditions.

In any case, I can revert this specific change (probably this went in accidentally while I was developing).

lib/spack/spack/test/spec_semantics.py
alalazo: It doesn't generate an error. Similarly to other spec properties, clingo will try to respect the request, and will place virtuals on edges accordingly. This is a real example with:
```console
$ spack graph -d --color strumpack ^openblas ^intel-parallel-studio > strumpack.dot
```
which gives back:
strumpack

lib/spack/spack/parser.py
alalazo: @becker33 See #39574

lib/spack/spack/spec.py
alalazo: See #39573

lib/spack/docs/basic_usage.rst
```diff
@@ -1525,6 +1525,29 @@ any MPI implementation will do.  If another package depends on
...
 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Explicit binding of virtual dependencies
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are packages that provide more than just one virtual dependency. When interacting with them, users
+might want to utilize just a subset of what they could provide, and use other providers for virtuals they
+need.
+
+With Spack it is possible to bind virtual dependencies explicitly using the ``virtuals`` edge attribute:
```
tgamblin: I don't think we should mention that this is about edge attributes here, unless we're going to define edge attribute and link to it somewhere:

```
It is possible to be more explicit and tell Spack which dependency should provide which virtual, using a special syntax:
```

lib/spack/docs/packaging_guide.rst
```diff
@@ -2937,6 +2937,34 @@ The ``provides("mpi")`` call tells Spack that the ``mpich`` package
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Packages can provide more than one virtual dependency. Sometimes, due to implementation details,
+there are subsets of those virtuals that need to be provided together by the same package.
+
+A well-known example is, for instance, ``openblas``, which provides both the ``lapack`` and ``blas`` API in a single library
+called ``libopenblas``. A package needing ``lapack`` and ``blas`` can either use ``openblas`` as a provider of both, or
+not use ``openblas`` at all.
+
+To express this constraint in a package, the two virtual dependencies must be listed in the same ``provides`` directive:
```
tgamblin:
```
well-known example is ``openblas``, which provides both the ``lapack`` and ``blas`` API in a single ``libopenblas`` library. A package that needs ``lapack`` and ``blas`` must either use ``openblas``
to provide both, or not use ``openblas`` at all. It cannot pick one or the other.

To express this constraint in a package, the two virtual dependencies must be listed in the same ``provides`` directive:
```

lib/spack/docs/packaging_guide.rst
```diff
@@ -2937,6 +2937,34 @@ The ``provides("mpi")`` call tells Spack that the ``mpich`` package
...
+
+.. code-block:: python
+
+   provides('blas', 'lapack')
+
+This makes it impossible to select ``openblas`` as a provider for one of the two
+virtual dependencies and not for the other. Any request to do so results in an
+error message:
```
tgamblin: If you try to, Spack will report an error:

alalazo:
Hmm, I was able to concretize the environment on Lassen. [Here](https://gitlab.spack.io/spack/spack/-/jobs/8274292) is the error on UO-typhon for reference.

lib/spack/spack/parser.py
```diff
@@ -6,15 +6,16 @@
...
                     [name] [node_options] hash |
                     filename
 
     node          =  name [node_options] |
                      [name] [node_options] hash |
                      filename
 
-    node_options  = [@(version_list|version_pair)] [%compiler] { variant }
+    node_options    = [@(version_list|version_pair)] [%compiler] { variant }
+    edge_properties = [[ { bool_variant | key_value } ]]
```
tgamblin: you've got the `[]` specified twice -- once above under `spec` and once here.  I'd just leave them under `spec` and do single `[]` here.

lib/spack/spack/parser.py
```diff
@@ -278,23 +282,34 @@ def next_spec(
...
 
         return root_spec
 
+    def _parse_node(self, root_spec):
+        dependency = SpecNodeParser(self.ctx).parse()
+        if dependency is None:
+            msg = (
+                "this dependency sigil needs to be followed by a package name "
+                "or a node attribute (version, variant, etc.)"
+            )
```
tgamblin: The error message points to the wrong character when triggered this way:

```
> spack spec hdf5 ^[virtuals=mpi]
==> Error: this dependency sigil needs to be followed by a package name or a node attribute (version, variant, etc.)
hdf5 ^[virtuals=mpi]
                   ^
```

I think the answer is to reformulate the message to be more clear, e.g. "the dependency sigil and any optional edge attributes must be followed by a package name"

lib/spack/spack/parser.py
```diff
@@ -432,6 +447,41 @@ def parse(self, initial_spec: spack.spec.Spec) -> spack.spec.Spec:
...
+
+    def parse(self):
+        attributes = {}
+        while True:
+            if self.ctx.accept(TokenType.KEY_VALUE_PAIR):
+                name, value = self.ctx.current_token.value.split("=", maxsplit=1)
+                name = name.strip("'\" ")
+                value = value.strip("'\" ").split(",")
+                attributes[name] = value
+                if name not in ("deptypes", "virtuals"):
```
tgamblin, line 471: I am not sure these are the best names -- renaming them will change the spec format, though, right?  I think it would be better for the user to specify `type=...` and `virtual=...`... Is it worth implementing that at this point before this becomes cemented?

With use variants, we'll have more usable names for edge attributes... but these are always going to be reserved as internal ones.  We *could* implement `type` and `virtual` as aliases and reserve `type`, `deptypes`, `virtual`, and `virtuals`, w/o changing the schema.  These are unlikely to be used for anything else.

lib/spack/spack/spec.py
```diff
@@ -1463,6 +1463,26 @@ def edges_to_dependencies(self, name=None, depflag: dt.DepFlag = dt.ALL):
...
+            union.update_deptypes(edge.depflag)
+            union.update_virtuals(edge.virtuals)
+        deptypes_str = (
+            f"deptypes={','.join(dt.flag_to_tuple(union.depflag))}" if union.depflag else ""
+        )
+        virtuals_str = f"virtuals={','.join(union.virtuals)}" if union.virtuals else ""
+        if not deptypes_str and not virtuals_str:
+            return ""
+        result = f"{deptypes_str} {virtuals_str}".strip()
+        return f"[{result}]"
```
tgamblin, line 1485:
```suggestion
        attrs = " ".join(
            f"deptypes={','.join(dt.flag_to_tuple(union.depflag))}" if union.depflag else "",
            f"virtuals={','.join(union.virtuals)}" if union.virtuals else "",
        )
        return f"[{attrs}]" if attrs else ""
```

var/spack/repos/builtin/packages/strumpack/package.py
tgamblin: What's the effect here?  Is this affecting default concretizations?

tgamblin (changes requested):
Not completely done, but here is a partial review.

lib/spack/spack/spec.py
becker33: I don't think this suggestion is quite correct. It looks to me like it would print `[" "]` if nothing needs to be printed, since using `" ".join` on two empty strings will create the non-empty string with a single space.

lib/spack/spack/parser.py
alalazo: > renaming them will change the spec format, though, right?

Yes, since in the JSON representation these are basically reserved edge variant names.

> Is it worth implementing that at this point before this becomes cemented?

I would say no, but in any case I would do it in a separate PR (to come shortly after) so we can keep the two modifications (feature and renaming) separated.

> We could implement type and virtual as aliases and reserve type, deptypes, virtual, and virtuals, w/o changing the schema. These are unlikely to be used for anything else.

Yes, this could be done.

lib/spack/docs/basic_usage.rst
alalazo: Done in https://github.com/spack/spack/pull/35322/commits/f2dd4b40906b9625abd0de0dfbc8cf37f4217196

lib/spack/docs/packaging_guide.rst
alalazo: Done in https://github.com/spack/spack/pull/35322/commits/f2dd4b40906b9625abd0de0dfbc8cf37f4217196

lib/spack/docs/packaging_guide.rst
alalazo: Done in https://github.com/spack/spack/pull/35322/commits/f2dd4b40906b9625abd0de0dfbc8cf37f4217196

lib/spack/spack/parser.py
alalazo: Done in https://github.com/spack/spack/pull/35322/commits/ffd8f7d60f0d6d7e129908a979c650969f265e8b

lib/spack/spack/parser.py
alalazo: Done in https://github.com/spack/spack/pull/35322/commits/84a310178460f45bf70134906f67e7ee1fda6f90

var/spack/repos/builtin/packages/strumpack/package.py
alalazo: Reverted in https://github.com/spack/spack/pull/35322/commits/56c362f65b397b64c3b98b50548806a963a8d622

This was to test that an edge attribute matters if part of a condition, but there is also a unit test to ensure this semantic works as intended.

alalazo: To be clear, this would matter in cases where we need to specialize on a dependency that could provide more than one virtual. For instance:
```python
depends_on("foo+intel", when="^intel-parallel-studio")
```
would trigger the dependency on `foo+bar` if `intel-parallel-studio` is a dependency too. While:
```python
depends_on("foo+intel", when="^[virtuals=mpi] intel-parallel-studio")
```
is more specific and would trigger the dependency only if `intel-parallel-studio` is the `mpi` provider.

lib/spack/spack/spec.py
tgamblin: Add `.strip()` then?  I think it's still clearer.

alalazo:
> Not completely done, but here is a partial review.

@tgamblin :point_up:

tgamblin (approved):
I went over the rest of the code here and it LGTM.

PR #35323, opened 2023-02-03
Allowing use of srun from anywhere in the current PATH

G-Ragghianti:
This modifies the smoke test such that it can use any available "srun" executable.  It was previously looking only in the MPI bin directory, but srun is not usually provided by MPI but by the resource scheduler (slurm). This should resolve the testing issues on perlmutter until a more comprehensive solution to finding smoke test dependencies is available in spack.


G-Ragghianti:
https://github.com/E4S-Project/e4s/issues/114

PR #35279, opened 2023-02-01
py-black: add v23.1.0

adamjstewart:
https://github.com/psf/black/releases/tag/23.1.0

PR #35325, opened 2023-02-03
py-packaging: add constrain on python version

alalazo:
This should fix the failure we see when bootstrapping on rhel8. Was inadvertently introduced in https://github.com/spack/spack/pull/35279

### DIFF
--- a/var/spack/repos/builtin/packages/py-packaging/package.py
+++ b/var/spack/repos/builtin/packages/py-packaging/package.py
@@ -24,6 +24,7 @@ class PyPackaging(PythonPackage):
 
     depends_on("py-flit-core@3.3:", when="@22:", type="build")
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-packaging/package.py
```diff
@@ -24,6 +24,8 @@ class PyPackaging(PythonPackage):
 
     depends_on("py-flit-core@3.3:", when="@22:", type="build")
 
+    depends_on("python@3.7:", when="@22:")
```
adamjstewart: type=build/run

adamjstewart: Also can you add a comment saying why this is necessary so I don't forget and remove it?

adamjstewart (approved):

PR #35326, opened 2023-02-03
[WIP] py-build: bootstrap without pip

adamjstewart:
Spack currently uses pip to install all Python libraries. However, pip is an entire package manager, and provides many features we don't need. We would like to move towards using build/installer instead of pip to simplify the installation process. This PR is the first step in that direction. We need to be able to bootstrap build/installer without pip in order to do that.

Unfortunately, build has the worst cyclic dependency problem I've ever seen. In order to build build, build needs to be installed (this isn't that bad), but all of builds dependencies also need to be installed. All of the following packages need to be installed from wheels in order to get a working build:

* installer (build-time only)
* flit-core (build-time only)
  * toml (3.1–3.3 only)
  * pytoml (–3.0 only)
* packaging
  * pyparsing (–21 only)
  * six (–20.7 only)
  * attrs (19.1 only)
* pyproject-hooks
* colorama (Windows only)
* importlib-metadata (Python 3.7 only)
  * zipp
  * typing-extensions (Python 3.7 only)
* tomli (Python 3.7–3.10 only)

It's also worth noting that installer only supports Python 3.7+, which means none of these packages can be built for EOL Python versions. This is relevant because Spack currently supports bootstrapping its style dependencies with Python 3.6. With this PR, that will no longer be supported. Curious how @tgamblin and @alalazo feel about this.

If we really do need to be able to install these with Python 3.6, we would need to add a new `WheelPackage` base class that supports installing with pip (Python 2+) or installer (Python 3.7+). This would also require 2 new builders. I was hoping to try to tackle this in a follow-up PR, but I can turn this into one massive PR if needed.

References:

* https://github.com/pypa/build/issues/470
* https://github.com/pypa/installer/issues/150

### DIFF
--- a/var/spack/repos/builtin/packages/py-attrs/package.py
+++ b/var/spack/repos/builtin/packages/py-attrs/package.py
@@ -15,18 +15,66 @@ class PyAttrs(Package, PythonExtension):
     list_url = "https://pypi.org/simple/attrs/"
     git = "https://github.com/python-attrs/attrs"
 
-    version("22.2.0", sha256="29e95c7f6778868dbd49170f98f8818f78f3dc5e0e37c0b1f474e3561b240836", expand=False)
-    version("22.1.0", sha256="86efa402f67bf2df34f51a335487cf46b1ec130d02b8d39fd248abfd30da551c", expand=False)
-    version("21.4.0", sha256="2d27e3784d7a565d36ab851fe94887c5eccd6a463168875832a1be79c82828b4", expand=False)
-    version("21.2.0", sha256="149e90d6d8ac20db7a955ad60cf0e6881a3f20d37096140088356da6c716b0b1", expand=False)
-    version("20.3.0", sha256="31b2eced602aa8423c2aea9c76a724617ed67cf9513173fd3a4f03e3a929c7e6", expand=False)
-    version("20.2.0", sha256="fce7fc47dfc976152e82d53ff92fa0407700c21acd20886a13777a0d20e655dc", expand=False)
-    version("20.1.0", sha256="2867b7b9f8326499ab5b0e2d12801fa5c98842d2cbd22b35112ae04bf85b4dff", expand=False)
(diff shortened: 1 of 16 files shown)


adamjstewart:
@spackbot fix style

adamjstewart:
At the moment, this feels prohibitively intrusive just to install build without pip. We may want to consider waiting until build vendors its dependencies (seems unlikely given https://github.com/pypa/build/issues/470) or abandon build support until a tool with fewer dependencies inevitably comes along (given that we have 10+ build backends, why not 10+ build frontends?).

charmoniumQ:
It's frustrating that this needs 3 different toml parsers, but not much we can do about that.

Is it preferable to install from wheel's (opaque binary) or from git (transparent source code, patches work)? I don't understand all of Python packaging, but I think pure Python packages need only be copied into the installation prefix.

adamjstewart:
For downloading, we prefer URLs over git because they can be checksummed. For sdist (source) vs. bdist (wheel), we currently prefer sdist since that's what we're used to from other languages and it makes patching and development easy, but there's no reason we can't use wheels. We've talked about moving away from URLs and towards wheels for the majority of pure-Python packages but that's a lot of work and may not be trivial. We've also talked about removing pure-Python packages from Spack entirely and instead using pip to handle those, but there's a lot of work that would need to be done on PyPI's side of things to make this possible.

adamjstewart:
Other alternative is to require pip to install build/installer, and then use build/installer for everything else. This seems silly to me, but is at least less disruptive.

rgommers:
> Other alternative is to require pip to install build/installer, and then use build/installer for everything else. This seems silly to me, but is at least less disruptive.

After reading https://github.com/pypa/build/issues/470, I think the "require `pip` to install `build`/`installer`" option seems very reasonable to me.

adamjstewart:
Let's close this. I don't love this PR in its current state. I think we'll likely go with one of the following options instead:

1. build (and its deps) can only be built with pip, every other package can be built with either (easy)
2. add a WheelPackage base class and convert all pure Python packages to wheels (hard)
3. all pure Python packages should be removed from Spack and pip should be used to install them (hardest)

I may reevaluate this after Python 3.7 support is dropped.

PR #34869, opened 2023-01-10
CDashReporter: remove unused argument

alalazo:
Modifications:
- [x] Removed `ctest_parsing` as it is not used in the code
- [x] Removed `extract_ctest_test_data` as it is not used in the code

PR #35328, opened 2023-02-03
Restore our ability to submit build/test results to CDash from GitLab CI

zackgalbreath:


alalazo:
Is this related to #35338 ? If not let me know if I can help in any way.

zackgalbreath:
> Is this related to #35338 ? If not let me know if I can help in any way.

I appreciate the offer! This is a follow-on to #35338

zackgalbreath:
Here's an example of configure/build/test results getting posted to CDash from GitLab CI for this branch:
https://cdash.spack.io/build/2162441

lib/spack/spack/test/cmd/test.py
```diff
@@ -319,3 +319,26 @@ def test_test_results_status(mock_packages, mock_test_stage, status, expected):
         else:
             assert status in results
         assert expected in results
+
+
+@pytest.mark.regression("35337")
+def test_report_filename_for_cdash(install_mockery_mutable_config, mock_fetch):
+    """Test that the temporary file used to write Test.xml for CDash is not the upload URL"""
```
tldahlgren: We should be writing to `Testing.xml`, not `Test.xml`.

tldahlgren: Which also means `./share/spack/templates/reports/cdash/Test.xml` should probably be removed.

tldahlgren (commented):
Just a few comments specifically related to stand-alone testing uses the `Testing.xml`, not `Test.xml` template.  

The alternative template was added when we still supported CTest output parsing/reporting using the `Test.xml` template.  That feature was removed in https://github.com/spack/spack/pull/34869 but I didn't catch the need to remove the associated template at that time.

tldahlgren:
At this point I have tested this on a local run to confirm that `spack test run --log-format=cdash` still produces the expected output.

lib/spack/spack/test/cmd/test.py
tldahlgren: (Technically test output files are `<package>_Testing.xml` but they are generated using the `Testing.xml` template.)

zackgalbreath: 👍

kwryankrattiger (approved):
This is great, really need this

lib/spack/spack/cmd/test.py
```diff
@@ -227,22 +227,22 @@ def test_run(args):
         )
 
 
+def report_filename(args, test_suite):
+    if args.log_file:
+        if os.path.isabs(args.log_file):
+            return args.log_file
+        else:
+            log_dir = os.getcwd()
+            return os.path.join(log_dir, args.log_file)
```
haampie, line 236: This can literally be a one-liner

```
return os.path.abspath(args.log_file if args.log_file else "test-{}".format(test_suite.name))
```

PR #35329, opened 2023-02-03
`spack clean` remove build artifacts for dev specs

psakievich:
Spack clean currently doesn't do anything for develop specs. This will remove all the build artifacts that automatically get created for out of source builds.  Not sure if there is a cleaner way to do this that will work for autotools. I really only work on CMake packages and this will work well there.


psakievich:
@spackbot fix-style

psakievich:
@spackbot commands

psakievich:
@spackbot fix style

lib/spack/spack/test/cmd/clean.py
```diff
@@ -109,3 +110,35 @@ def _check_files(directory):
...
+        open(build_log, "w").close()
+
+        e = ev.Environment('.')
+        ev.activate(e)
+
+        # pre-clean state
+        assert ev.active_environment
+        assert os.path.isdir(build_dir)
+        assert os.path.isfile(build_log)
+        clean('mpich')
```
psakievich: Looks like I need to mock the installation to use this command successfully.  Any suggestions here?

PR #35332, opened 2023-02-03
Add new versions of spglib

RMeli:


var/spack/repos/builtin/packages/spglib/package.py
```diff
@@ -37,6 +45,17 @@ class Spglib(CMakePackage):
...
         return find_libraries("libsymspg", root=self.prefix, shared=True, recursive=True)
+
+    def cmake_args(self):
+        return [self.define_from_variant("USE_OMP", "openmp")]
+
+    def url_for_version(self, version):
+        if version <= Version("1.16.1"):
+            return f"https://github.com/atztogo/spglib/archive/v{version}.tar.gz"
+        else:
+            return f"https://github.com/spglib/spglib/archive/refs/tags/v{version}.tar.gz"
```
tldahlgren: The older packages are no longer available at that URL so cannot be checksummed.

It would be better to restore the `url` property, optionally updating it to the latest version if desired, since all of the listed versions are available there and with the same checksums shown here.

tldahlgren (changes requested):
Confirmed the sha256 for the new versions but the old ones are no longer available at the old URL.

var/spack/repos/builtin/packages/spglib/package.py
RMeli: Thanks @tldahlgren. I did not check older versions with `spack checksum` (and can confirm that they don't work), but I tried to install a couple of them and it worked (`spack install spglib@1.10.0`). Apparently it was fetching from a mirror. I'll change it as you suggest.

tldahlgren (approved):
LGTM. Re-confirmed (all) sha256.

PR #35334, opened 2023-02-04
Support packages for using scitokens on OSG

wdconinc:
The Open Science Grid (OSG)  encourages scitokens to provide certain services (e.g. writing to xrootd). Spack already supports this through scitokens-cpp and xrootd +scitokens-cpp.

This adds py-htgettoken, a python utility to get a scitoken from a vault through web authentication. To support htgettoken, this also adds py-gssapi.

This also adds the OSG CA cert collection which is typically at /etc/grid-security but pointed to in user installations by the X509_CERT_DIR variable.

This allows userspace support through spack for functionality that otherwise depends on installing the RPMs provided by OSG.

### DIFF
--- a/var/spack/repos/builtin/packages/py-gssapi/package.py
+++ b/var/spack/repos/builtin/packages/py-gssapi/package.py
@@ -13,7 +13,7 @@ class PyGssapi(PythonPackage):
     homepage = "https://github.com/pythongssapi/python-gssapi"
     pypi = "gssapi/gssapi-1.8.2.tar.gz"
 
-    maintainers = ["wdconinc"]
+    maintainers("wdconinc")
 
     depends_on("py-cython@0.29.29:2", type="build")
     depends_on("py-setuptools@40.6.0:", type="build")
--- a/var/spack/repos/builtin/packages/py-gssapi/package.py
+++ b/var/spack/repos/builtin/packages/py-gssapi/package.py
@@ -14,7 +14,7 @@ class PyGssapi(PythonPackage):
     pypi = "gssapi/gssapi-1.8.2.tar.gz"
 
     maintainers = ["wdconinc"]
-    
+
     depends_on("py-cython@0.29.29:2", type="build")
     depends_on("py-setuptools@40.6.0:", type="build")
 
--- a/var/spack/repos/builtin/packages/py-gssapi/package.py
+++ b/var/spack/repos/builtin/packages/py-gssapi/package.py
@@ -15,9 +15,9 @@ class PyGssapi(PythonPackage):
 
     maintainers("wdconinc")
 
+    version("1.8.2", sha256="b78e0a021cc91158660e4c5cc9263e07c719346c35a9c0f66725e914b235c89a")
+
     depends_on("py-cython@0.29.29:2", type="build")
     depends_on("py-setuptools@40.6.0:", type="build")
 
     depends_on("py-decorator")
-
-    version("1.8.2", sha256="b78e0a021cc91158660e4c5cc9263e07c719346c35a9c0f66725e914b235c89a")
--- a/var/spack/repos/builtin/packages/py-gssapi/package.py
+++ b/var/spack/repos/builtin/packages/py-gssapi/package.py
@@ -1,0 +1,23 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyGssapi(PythonPackage):
+    """Python-GSSAPI provides both low-level and high level wrappers
+    around the GSSAPI C libraries."""
+
+    homepage = "https://github.com/pythongssapi/python-gssapi"
+    pypi = "gssapi/gssapi-1.8.2.tar.gz"
+
+    maintainers = ["wdconinc"]
+    
+    depends_on("py-cython@0.29.29:2", type="build")
+    depends_on("py-setuptools@40.6.0:", type="build")
+
+    depends_on("py-decorator")
+
+    version("1.8.2", sha256="b78e0a021cc91158660e4c5cc9263e07c719346c35a9c0f66725e914b235c89a")
--- a/var/spack/repos/builtin/packages/py-gssapi/package.py
+++ b/var/spack/repos/builtin/packages/py-gssapi/package.py
@@ -20,4 +20,4 @@ class PyGssapi(PythonPackage):
     depends_on("py-cython@0.29.29:2", type="build")
     depends_on("py-setuptools@40.6.0:", type="build")
 
-    depends_on("py-decorator")
+    depends_on("py-decorator", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-htgettoken/package.py
+++ b/var/spack/repos/builtin/packages/py-htgettoken/package.py
@@ -1,0 +1,35 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyHtgettoken(PythonPackage):
+    """htgettoken gets OIDC authentication tokens for High Throughput Computing
+    via a Hashicorp vault server."""
+
+    homepage = "https://github.com/fermitools/htgettoken"
+
+    # htgettoken is not available on PyPi
+    url = "https://github.com/fermitools/htgettoken/archive/refs/tags/v1.16.tar.gz"
+    list_url = "https://github.com/fermitools/htgettoken/releases/"
+    git = "https://github.com/fermitools/htgettoken.git"
+
+    maintainers = ["wdconinc"]
+
+    version("1.16-33-g3788bb4", commit="3788bb4733e5e8f856cee51566df9a36cbfe097d")
+    version("1.16-20-g8b72f48", commit="8b72f4800ef99923dac99dbe0756a26266a27886")
+    # Older versions do not have a python build system
+    version("1.16", sha256="984fd4746082086356d968b4db536f7b35d30e65a6589fa357caa3266cb98268", deprecated=True)
+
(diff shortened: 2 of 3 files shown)


wdconinc:
@aandvalenzuela, maybe of interest, since this refers to your scitokens-cpp package and support in xrootd (for which thanks).

wdconinc:
@spackbot fix style

var/spack/repos/builtin/packages/py-gssapi/package.py
```diff
@@ -0,0 +1,23 @@
...
+
+    homepage = "https://github.com/pythongssapi/python-gssapi"
+    pypi = "gssapi/gssapi-1.8.2.tar.gz"
+
+    maintainers = ["wdconinc"]
+
+    depends_on("py-cython@0.29.29:2", type="build")
+    depends_on("py-setuptools@40.6.0:", type="build")
+
+    depends_on("py-decorator")
```
adamjstewart:
```suggestion
    depends_on("py-decorator", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-gssapi/package.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class PyGssapi(PythonPackage):
+    """Python-GSSAPI provides both low-level and high level wrappers
+    around the GSSAPI C libraries."""
+
+    homepage = "https://github.com/pythongssapi/python-gssapi"
+    pypi = "gssapi/gssapi-1.8.2.tar.gz"
+
+    maintainers = ["wdconinc"]
```
adamjstewart:
```suggestion
    maintainers("wdconinc")
```
This new formats appends instead of overwrites so I will still be listed

var/spack/repos/builtin/packages/py-gssapi/package.py
```diff
@@ -0,0 +1,23 @@
...
+
+    depends_on("py-cython@0.29.29:2", type="build")
+    depends_on("py-setuptools@40.6.0:", type="build")
+
+    depends_on("py-decorator")
+
+    version("1.8.2", sha256="b78e0a021cc91158660e4c5cc9263e07c719346c35a9c0f66725e914b235c89a")
```
adamjstewart: I usually put versions before dependencies, just a style thing, not really required

var/spack/repos/builtin/packages/py-htgettoken/package.py
```diff
@@ -0,0 +1,39 @@
...
+
+class PyHtgettoken(PythonPackage):
+    """htgettoken gets OIDC authentication tokens for High Throughput Computing
+    via a Hashicorp vault server."""
+
+    homepage = "https://github.com/fermitools/htgettoken"
+
+    # htgettoken is not available on PyPi
+    url = "https://github.com/fermitools/htgettoken/archive/refs/tags/v1.16.tar.gz"
+    list_url = "https://github.com/fermitools/htgettoken/releases/"
```
adamjstewart: This shouldn't be needed, and if it is, then something is broken

var/spack/repos/builtin/packages/py-htgettoken/package.py
```diff
@@ -0,0 +1,39 @@
...
+    git = "https://github.com/fermitools/htgettoken.git"
+
+    maintainers = ["wdconinc"]
+
+    version("1.16-33-g3788bb4", commit="3788bb4733e5e8f856cee51566df9a36cbfe097d")
+    version("1.16-20-g8b72f48", commit="8b72f4800ef99923dac99dbe0756a26266a27886")
```
adamjstewart, line 26: Are these official version numbers? Might be better to use something like `1.16-2022-02-06` that sorts better. Could also just only include master only and wait for a new release

var/spack/repos/builtin/packages/py-htgettoken/package.py
```diff
@@ -0,0 +1,39 @@
...
+    # Older versions do not have a python build system
+    version(
+        "1.16",
+        sha256="984fd4746082086356d968b4db536f7b35d30e65a6589fa357caa3266cb98268",
+        deprecated=True,
+    )
+
+    depends_on("py-setuptools", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-setuptools@30.3:", type="build")
```

var/spack/repos/builtin/packages/py-htgettoken/package.py
```diff
@@ -0,0 +1,39 @@
...
+
+    depends_on("py-gssapi", type="run")
+    depends_on("py-paramiko", type="run")
+    depends_on("py-urllib3", type="run")
```
adamjstewart: Should be type=build/run so import tests pass

var/spack/repos/builtin/packages/py-htgettoken/package.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def setup_run_environment(self, env):
+        if env.get("XDG_RUNTIME_DIR") and env.get("UID"):
+            env.set("BEARER_TOKEN", join_path(env.get("XDG_RUNTIME_DIR"), "bt_u" + env.get("UID")))
```
adamjstewart: If you're relying on these being set in the osg-ca-certs package I'm not sure if this will always be the case.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-htgettoken/package.py
wdconinc: 1.16 is an official version number, but doesn't have setuptools. `8b72f4800ef99923dac99dbe0756a26266a27886` is when setuptools support was added. `3788bb4733e5e8f856cee51566df9a36cbfe097d` is the 'natural' commit after a bunch of changes. These are reproducible lexically sortable version numbers from `git describe`. As soon as 1.17 is out, we'd just refer to that and forget about the 'special' `git describe` versions.

var/spack/repos/builtin/packages/py-htgettoken/package.py
wdconinc:
```suggestion
    depends_on("py-gssapi", type=("build", "run"))
    depends_on("py-paramiko", type=("build", "run"))
    depends_on("py-urllib3", type=("build", "run"))
```

adamjstewart (dismissed):

var/spack/repos/builtin/packages/py-htgettoken/package.py
adamjstewart: @alalazo do you have any thoughts on this?

wdconinc: Yes, you are right that this could be made more robust. Upon inspection, the package will place the token at `BEARER_TOKEN_FILE` if defined, and will read from `BEARER_TOKEN` if defined, `BEARER_TOKEN_FILE` otherwise. The default file if none of this is defined is based on `XDG_RUNTIME_DIR` or `/tmp` and `UID` or `os.getuid()`.

I think the following is more robust then (but will need `import os` at the top):
```suggestion
    def setup_run_environment(self, env):
        dir = env.get("XDG_RUNTIME_DIR") or "/tmp"
        uid = env.get("UID") or os.geteuid()
        file = join_path(dir, "bt_u" + uid)
        env.set("BEARER_TOKEN", file)
        env.set("BEARER_TOKEN_FILE", file)
```

var/spack/repos/builtin/packages/py-htgettoken/package.py
wdconinc: Added documentation to package.py.

tldahlgren (commented):
Confirmed the `versions` sha256 and commits.

var/spack/repos/builtin/packages/py-htgettoken/package.py
adamjstewart: This still relies on env vars set by other packages, so want to get @alalazo's feedback first

wdconinc: Nothing sets `XDG_RUNTIME_DIR` that is in spack, so there's nothing this package can depend on. `UID` is a bash variable, but we cannot rely on running bash either. So, maybe another way would be to ignore these locations where htgettoken itself would put the token, and just construct our own path:
```
        dir = env.get("TMPDIR") or "/tmp"
        uid = os.geteuid()
        file = join_path(dir, "bt_u" + uid)
        env.set("BEARER_TOKEN", file)
        env.set("BEARER_TOKEN_FILE", file)
```

We can't really use `tempfile.mkdtemp` since this has to persist across shell sessions through the token lifetime, say 24 hours (otherwise a user has to re-authenticate for every shell session).

var/spack/repos/builtin/packages/py-htgettoken/package.py
```diff
@@ -0,0 +1,45 @@
...
+    url = "https://github.com/fermitools/htgettoken/archive/refs/tags/v1.16.tar.gz"
+    git = "https://github.com/fermitools/htgettoken.git"
+
+    maintainers("wdconinc")
+
+    # The following versions refer to setuptools-buildable commits after 1.16;
+    # they are special reproducible version numbers from `git describe`
+    version("1.16-33-g3788bb4", commit="3788bb4733e5e8f856cee51566df9a36cbfe097d")
+    version("1.16-20-g8b72f48", commit="8b72f4800ef99923dac99dbe0756a26266a27886")
+    # Older versions do not have a python build system but included for completeness
```
adamjstewart: If the old version doesn't build I would just remove it, no reason to add a deprecated version that will be removed in the next release.

var/spack/repos/builtin/packages/py-htgettoken/package.py
```diff
@@ -0,0 +1,45 @@
...
+    version(
+        "1.16",
+        sha256="984fd4746082086356d968b4db536f7b35d30e65a6589fa357caa3266cb98268",
+        deprecated=True,
+    )
```
wdconinc: Remove older versions
```suggestion
    # Older versions do not have a python build system
```

var/spack/repos/builtin/packages/py-htgettoken/package.py
wdconinc: Good point, removed.

adamjstewart (approved):
This is probably good to go

PR #35340, opened 2023-02-04
[WIP] pip: check dependencies before install

adamjstewart:
If I'm understanding correctly, adding this flag will cause the build to fail if the dependencies are incorrect. I still need to test this, but a few gotchas I want to check for:

1. Is this only for build dependencies (setup_requires) or also run dependencies (install_requires)?
2. Will this break numpy where we relax the pinned build dep versions?
3. Will this break pillow-simd which we use as an alternative to pillow?


adamjstewart:
I still think this would be useful to add someday, but this PR can't possibly work until we're able to support multiple versions of build deps in the DAG at the same time. Will close for now.

PR #35341, opened 2023-02-04
py-installer: bootstrap without pip

adamjstewart:
Extracted from #35326. The rest of that PR seems unlikely to be merged, but this part should be good to go.

In a follow-up PR, I'll likely update most of our wheel installations to switch from pip to installer since the latter is much more lightweight and faster. Our builds will likely continue to use pip unless build resolves it's circular dependency issues.


adamjstewart:
@spackbot fix style

alalazo (approved):

alalazo:
Waiting for @pradyunsg to see if he has comments, but feel free to merge as soon as this seems ready

PR #35342, opened 2023-02-04
Add intel-gtpin@3.2.2

wyphan:
SHA256 sum computed manually after verifying the locally downloaded file using the given SHA1 sum at https://www.intel.com/content/www/us/en/download/730596/gtpin.html


rashawnLK (approved):
Thanks Wil! This looks good.

alalazo (approved):

PR #35345, opened 2023-02-05
opencascade: new variants

wdconinc:
OpenCascade has several major modules and not every application needs all of them. This adds variants for the various modules.

It also updates the 3rdparty dependency treatment.

### DIFF
--- a/var/spack/repos/builtin/packages/opencascade/package.py
+++ b/var/spack/repos/builtin/packages/opencascade/package.py
@@ -177,7 +177,7 @@ class Opencascade(CMakePackage):
         args.append("-DBUILD_DOC_Overview=OFF")
 
         # Always build the foundation classes
-        args.append(self.define("BUILD_MODULE_foundation_classes", true))
+        args.append(self.define("BUILD_MODULE_FoundationClasses", true))
         # Specify which modules to build
         for module in [
             "ApplicationFramework",
--- a/var/spack/repos/builtin/packages/opencascade/package.py
+++ b/var/spack/repos/builtin/packages/opencascade/package.py
@@ -76,13 +76,36 @@ class Opencascade(CMakePackage):
 
     # Modules, per DAG at https://dev.opencascade.org/doc/refman/html/
     variant("foundation_classes", default=True, description="Build Foundation Classes module")
-    variant("modeling_data", default=True, description="Build Modeling Data module", when="foundation_classes")
-    variant("modeling_algorithms", default=True, description="Build Modeling Algorithms module", when="+modeling_data")
-    variant("visualization", default=True, description="Bulid Visualization module", when="+modeling_algorithms")
     variant(
-        "application_framework", default=True, description="Build Application Framework module", when="+visualization"
+        "modeling_data",
+        default=True,
+        description="Build Modeling Data module",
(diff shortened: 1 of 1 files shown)


wdconinc:
@spackbot fix style

var/spack/repos/builtin/packages/opencascade/package.py
```diff
@@ -67,59 +74,134 @@ class Opencascade(CMakePackage):
         when="@:7.7.0",
     )
 
+    # Modules, per DAG at https://dev.opencascade.org/doc/refman/html/
+    variant("foundation_classes", default=True, description="Build Foundation Classes module")
+    variant(
+        "modeling_data",
+        default=True,
+        description="Build Modeling Data module",
+        when="foundation_classes",
```
wdconinc:
```suggestion
        when="+foundation_classes",
```

var/spack/repos/builtin/packages/opencascade/package.py
```diff
@@ -67,59 +74,134 @@ class Opencascade(CMakePackage):
         when="@:7.7.0",
     )
 
+    # Modules, per DAG at https://dev.opencascade.org/doc/refman/html/
+    variant("foundation_classes", default=True, description="Build Foundation Classes module")
```
alecbcs: @wdconinc is there a use case where someone might not want to build the foundation classes?

wdconinc: Yeah, there isn't. I'll remove that variant.

wdconinc:
@spackbot fix style

wdconinc:
FYI, the splitting into different modules is a first step towards making some of the dependencies dependent on the enabled modules. E.g. `+vtk` doesn't make sense unless `+visualization`, but `+tbb` does make sense for the foundation classes. I can either do all that work in this PR or in a different one. LMK.

alecbcs (dismissed):
Looks good to me. Thank you for the contribution @wdconinc! Would you like to continue work on this PR to segment out the dependencies or add it in a follow up?

var/spack/repos/builtin/packages/opencascade/package.py
```diff
@@ -67,59 +74,129 @@ class Opencascade(CMakePackage):
...
-        if "+rapidjson" in self.spec:
-            args.append("-DUSE_RAPIDJSON=ON")
-            args.append("-D3RDPARTY_RAPIDJSON_DIR=%s" % self.spec["rapidjson"].prefix)
-        else:
-            args.append("-DUSE_RAPIDJSON=OFF")
+        # Disable documentation building
+        args.append("-DBUILD_DOC_Overview=OFF")
+
+        # Always build the foundation classes
+        args.append(self.define("BUILD_MODULE_foundation_classes", true))
```
wdconinc:
```suggestion
        args.append(self.define("BUILD_MODULE_FoundationClasses", true))
```

wdconinc:
I just compared my notes and there was one issue in this PR that I just fixed, but the rest is for a future PR (i.e. adding conditionals to the e.g. `gl`  variants and dependencies). This is ready for merging.

alecbcs (approved):
Looks good to me!

PR #35282, opened 2023-02-01
Pin black to 22.12.0 to avoid spurious style changes

alalazo:
We need to sync changes on major versions with spackbot, to avoid the bot saying everything is alright and have CI failing nonetheless.

PR #35351, opened 2023-02-05
Style: black 23, skip magic trailing comma

adamjstewart:
This PR includes two changes related to black. These changes are easily separable if you would like me to submit 2 PRs instead. I included them together so we only need to ignore 1 commit in `git blame`.

- [x] Run black 23 on the entire repo
- [x] Skip magic trailing comma

Black 23 is stricter than black 22, and includes several new formatting changes. The most common you'll notice are removing newline characters after indentation (if-statement, for-loop, function-definition) and removing unnecessary tuples. It is backwards compatible in the sense that black 22 won't try to undo these changes, black 23 is just more strict. Figured we should run this on the repo instead of forcing PR authors to fix this on each file they modify.

Black has a concept of magic trailing commas, where:
```python
version(
    "1.2.3",
    sha256="4784eddd89c17a5fa030d408392992b3da3c503c872800e9d3746d985cfcc92a"
)
```
would be reformatted to:
```python
version("1.2.3", sha256="4784eddd89c17a5fa030d408392992b3da3c503c872800e9d3746d985cfcc92a")
```
but:
```python
version(
    "1.2.3",
    sha256="4784eddd89c17a5fa030d408392992b3da3c503c872800e9d3746d985cfcc92a",
)
```
wouldn't. The `skip_magic_trailing_comma` flag treats both as equivalent and reformats things to one line if possible. You can think of this as further reducing the set of possible ways to write the same thing.


alalazo:
> It is backwards compatible in the sense that black 22 won't try to undo these changes, black 23 is just more strict.

Given the above I think it's fine to have these changes, modulo some annoyance when backporting commits to point releases. If we want to actually use black23 in CI we need to:
- Revert https://github.com/spack/spack/pull/35282
- Have spackbot use black 23

adamjstewart:
@spackbot run pipeline

tgamblin (dismissed):

adamjstewart:
@spackbot run pipeline

adamjstewart:
Waiting on pfunit to be fixed in develop...

PR #35353, opened 2023-02-06
arrow: explicitly disable RE2 and UTF8PROC flags when not in dependencies

heerener:
Without ARROW_WITH_RE2 set to False, building fails with:
```
-- Checking for module 're2'
-- No package 're2' found
```

Without ARROW_WITH_UTF8PROC set to False, building fails with:
```
Could NOT find utf8proc: Found unsuitable version "", but required is at
```

### DIFF
--- a/var/spack/repos/builtin/packages/arrow/package.py
+++ b/var/spack/repos/builtin/packages/arrow/package.py
@@ -148,7 +148,7 @@ class Arrow(CMakePackage, CudaPackage):
         args.append(self.define_from_variant("ARROW_WITH_ZLIB", "zlib"))
         args.append(self.define_from_variant("ARROW_WITH_ZSTD", "zstd"))
 
-        if not self.spec.satisfies("^re2"):
+        if not self.spec.dependencies("re2"):
             args.append(self.define("ARROW_WITH_RE2", False))
         if not self.spec.satisfies("^utf8proc"):
             args.append(self.define("ARROW_WITH_UTF8PROC", False))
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/arrow/package.py
```diff
@@ -146,6 +146,11 @@ def cmake_args(self):
         args.append(self.define_from_variant("ARROW_WITH_ZLIB", "zlib"))
         args.append(self.define_from_variant("ARROW_WITH_ZSTD", "zstd"))
 
+        if not self.spec.satisfies("^re2"):
```
haampie: This can potentially match indirect deps on re2, i.e. false positives.

haampie: This is for direct deps
```suggestion
        if not self.spec.dependencies("re2"):
```

haampie (changes requested):

var/spack/repos/builtin/packages/arrow/package.py
heerener: Thanks for clarifying! I have a feeling there are more patches with this mistake waiting to be upstreamed here, I'll keep an eye out.

heerener:
@haampie now that I've seen that I was actually missing one of the suggested changes, can I bother you for a re-review and, possibly, a merge?

haampie (dismissed):

haampie (approved):
Sorry for the late merge...

PR #35358, opened 2023-02-06
wps: adding version 4.4

snehring:
This seems to build with the expected executables based on examining the last version. I'm not familiar enough with wps to speak to the correctness though, would appreciate it if a maintainer could give it a once over.


MichaelLaufer (approved):
LGTM
Thanks @snehring !

tldahlgren (approved):
Confirmed sha256.

PR #35362, opened 2023-02-06
py-mypy: add v1.0.0

adamjstewart:
Mypy switched to semver!

Successfully builds and runs on macOS 10.15.7 with Python 3.10.8 and Apple Clang 12.0.0.


tldahlgren (approved):
Confirmed the sha256.

PR #35364, opened 2023-02-07
tiramisu: new package

wraith1995:
Very similar to [the halide pull request](https://github.com/spack/spack/pull/35254) and, indeed, depends on that.


tldahlgren (commented):
Confirmed the branch and commit.

tldahlgren:
Adding @adamjstewart since this package extends Python (and because it is dependent on a package that has been approved but Adam had a question about).

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,82 @@
...
+
+    maintainers = ["wraith1995"]
+
+    generator = "Ninja"
+
+    version("master", branch="master")
+    version("2023-2-3", commit="73a9cec72e08d4dfe5e8c66da33139008124a4fa")
+
+    variant("python", default=True, description="Install python bindings.")
+    variant("cuda", default=False, description="Enable Cuda Code Generation.")
```
adamjstewart: Might be good to subclass CudaPackage so you can get all of the other nice features that base class adds. Just add CudaPackage to the list of superclasses, and remove the variant and dependency since those are already added by the base class.

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,82 @@
...
+    extends("python", when="+python")
+    variant(
+        "debug",
+        default="0",
+        description="Set the debug level.",
+        multi=False,
+        values=("0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10"),
+    )
+
+    depends_on("cmake@3.24:", type="build")
```
adamjstewart:
```suggestion
    depends_on("cmake@3.5:", type="build")
```
According to 73a9cec72e08d4dfe5e8c66da33139008124a4fa

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,82 @@
...
+    depends_on("ninja", type="build")
+
+    depends_on("halide@14.0.0", type=("build", "link", "run"))
```
adamjstewart: Does it have to be this exact version, or are there a wider range of versions supported?

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,82 @@
...
+    depends_on("isl", type=("build", "link", "run"))
+    depends_on("cuda", when="+cuda", type=("build", "link", "run"))
+
+    depends_on("python@3.8:", type=("build", "link", "run"), when="+python")
+    depends_on("py-pybind11@2.6.2:", type=("build", "link", "run"), when="+python")
+    depends_on("py-numpy", type=("build", "run"), when="+python")
+    depends_on("py-cython", type=("run"), when="+python")
```
adamjstewart: pybind11 and cython are rarely needed at run-time. Where did find the full list of deps?

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def cmake_args(self):
+        spec = self.spec
+        llvmdir = spec["llvm"].prefix.bin
+        args = [
+            self.define("HALIDE_LIB_DIRECTORY", spec["halide"].prefix.lib),
```
adamjstewart:
```suggestion
            self.define("HALIDE_LIB_DIRECTORY", spec["halide"].libs.directories[0]),
```
CMake often installs things into lib or lib64 depending on the system, so lib isn't always right. This will only work if there is a `libhalide.so` library installed. If the name is different, you'll need to add a `libs` method to halide.

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,82 @@
...
+            self.define("LLVM_CONFIG_BIN", llvmdir),
+            self.define("HALIDE_SOURCE_DIRECTORY", spec["halide"].prefix),
+            self.define("ISL_INCLUDE_DIRECTORY", spec["isl"].prefix.include),
+            self.define("ISL_LIB_DIRECTORY", spec["isl"].prefix.lib),
```
adamjstewart:
```suggestion
            self.define("ISL_INCLUDE_DIRECTORY", spec["isl"].headers.directories[0]),
            self.define("ISL_LIB_DIRECTORY", spec["isl"].libs.directories[0]),
```
Same logic here

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,82 @@
...
+            self.define("USE_FLEXNLP", False),
+        ]
+        if "+python" in spec:
+            args += [
+                self.define("Tiramisu_INSTALL_PYTHONDIR", python_platlib),
+                self.define("Python3_EXECUTABLE", spec["python"].command.path),
+            ]
+        return args
+
+    def build(self, pkg, spec):
```
adamjstewart: Is there a reason to use cmake instead of make for the build/install phases? I haven't seen this in any other package.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/tiramisu/package.py
wraith1995: Pybind is indeed only build time, but Cython is actually used to build python extensions for the code that Tiramisu generates. 

In this case, the full list of depends is only partially detailed in the readme at the moment. I can fix the readme if so desired.

var/spack/repos/builtin/packages/tiramisu/package.py
wraith1995: Sadly right now it needs to be exactly this version as we rely on a fairly precise version of the internal API. We could allow this version or higher, but things could easily break.

var/spack/repos/builtin/packages/tiramisu/package.py
wraith1995: I did this because I knew it worked and the defaults did not. I am not sure why. I can investigate further if required, but I do agree it is really weird.

var/spack/repos/builtin/packages/tiramisu/package.py
wraith1995: Done

var/spack/repos/builtin/packages/tiramisu/package.py
wraith1995: I am making a pull to update the readme with the current dependencies.

var/spack/repos/builtin/packages/tiramisu/package.py
adamjstewart: If you can get me an error message that may help

wraith1995: Sure. I'd appreciate that though I don't recall them being particularly illuminating. IIRC it tries to build more than it should, but not the dependencies for these things. Let me go erase those lines and upload some logs.

wraith1995: I realized the problem. This is quite silly. I'll push a solution in a few after a bit more verification.

wraith1995: Okay. Basically, I was not properly specifying the targets, however, now I've found a bug in the cmake files. Thanks for prompting me to look at this agian. I will fix that, get it merged in, and then ping you when done.

wraith1995: @adamjstewart It is done. 

The short short version is that a standard use of the cmake package now works.

The short version is that CMake implicitly made the install targets depend on all targets and some targets (benchmarks mainly) didn't work for our of source builds due to some hacky cmake. I've fixed the first issue and decided not to expose benchmarks here until I have time to fix them too.

var/spack/repos/builtin/packages/tiramisu/package.py
```diff
@@ -0,0 +1,71 @@
...
+class Tiramisu(CMakePackage, CudaPackage, PythonExtension):
+    """Tiramisu is a polyhedral compiler for dense and sparse deep learning
+    and data parallel algorithms.It provides a simple C++ API for expressing
+    algorithms and how these algorithms should be optimized by the compiler."""
+
+    homepage = "http://tiramisu-compiler.org"
+    url = "https://github.com/Tiramisu-Compiler/tiramisu"
+    git = "https://github.com/Tiramisu-Compiler/tiramisu.git"
+
+    maintainers = ["wraith1995"]
```
adamjstewart:
```suggestion
    maintainers("wraith1995")
```
We have a new syntax that appends instead of overwrites, so I won't be removed from the list

adamjstewart (changes requested):

adamjstewart (approved):

wraith1995:
@adamjstewart Sorry to be impatient, but is the pipelined (particularly the last remaining radiuss-pr-build) stalled for some reason?

adamjstewart:
One of them failed, prob transient. @spackbot run pipeline

Issue #8670, opened 2018-07-10
OpenMPI > 1.10.x including rpath from --with-xx statements into compiler wrapper


jrood-nrel:
So I still have an issue of OpenMPI > 1.10.x putting `/usr/lib64` in the mpi wrappers because I am using `--with-verbs` that finds IB libraries in `/usr/lib64` and the compiler I'm using isn't in `/usr/lib64` so the executables I build with OpenMPI 3.1.0, for example have GLIBC errors at runtime. I feel like we should be putting the `extra_rpaths` into the OpenMPI wrappers with rpath as well. I'm just not sure of the best approach. Hoping to find a solution to this so I can use OpenMPI > 1.10.x without hacking the wrappers myself. It really surprises me that I seem to be the only one with this problem though.

scheibelp:
Fixed by https://github.com/spack/spack/pull/8687

> It really surprises me that I seem to be the only one with this problem though.

The issue was a bit more specific than the title suggests - in order to run into the problem you have to:

* Be using a compiler outside of /usr
* Be using the OpenMPI wrappers outside of a Spack build

(the fix PR was more clear on the details)

PR #35376, opened 2023-02-07
Start using paths found in extra_rpaths in compilers.yaml when building

MatthewLieber:
When built MVAPICH will now use the paths found in extra rpaths within the compilers.yaml file

### DIFF
--- a/var/spack/repos/builtin/packages/mvapich/package.py
+++ b/var/spack/repos/builtin/packages/mvapich/package.py
@@ -6,6 +6,7 @@
 import os.path
 import re
 import sys
+import itertools
 
 from spack.package import *
 
@@ -286,7 +287,12 @@ class Mvapich(AutotoolsPackage):
             args.append("--enable-registration-cache")
         else:
             args.append("--disable-registration-cache")
-
+       
+        ld = ''
+        for path in itertools.chain(self.compiler.extra_rpaths, self.compiler.implicit_rpaths()):
+                    ld += "-Wl,-rpath," + path + " "     
+        if ld != '':   
+            args.append("LDFLAGS="+ld)
         args.extend(self.process_manager_options)
         args.extend(self.network_options)
         args.extend(self.file_system_options)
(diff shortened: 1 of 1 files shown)


natshineman (approved):
Confirming the maintainer change: Matt is now full time staff and will be working on all things Spack related, while Nick is transitioning to more of his research as a PhD student.

harisubramoni (approved):

harisubramoni:
This can be merged. I am not sure why this is not still being automatically taken in.

var/spack/repos/builtin/packages/mvapich/package.py
```diff
@@ -287,6 +288,11 @@ def configure_args(self):
         else:
             args.append("--disable-registration-cache")
 
+        ld = ""
+        for path in itertools.chain(self.compiler.extra_rpaths, self.compiler.implicit_rpaths()):
```
haampie, line 292: Why is this necessary? Spack's compiler wrapper adds those rpaths through the compiler wrapper. If not, then the compiler wrapper is not used, which is a problem.

natshineman: We received a bug report from @nhanford that the rpath's were not correctly being added to MVAPICH's compiler wrappers. This was an attempt to resolve that. Perhaps Nate can weigh in if we went about this in a method different from what he was thinking?

nhanford: @haampie We're seeing similar issues to #8670 and the proposed solution happens to resemble #8687. Perhaps there is a better way to go about this?

tgamblin: We run `filter_compiler_wrappers` on the installed `mpicc`, etc., because the Spack wrappers cannot be executed outside of Spack's build environment, so you want your `mpicc` to execute with the *real* compiler.  If the real compiler needed extra RPATHs in Spack, it likely needs them outside Spack, so it makes a lot of sense (to me) that we'd make the wrappers add these paths.

Spack-installed MPIs are *not* expected to be run just by Spack -- you want them to work with `module load`, `spack load`, envs., etc. too.

becker33 (approved):
I agree that we need these for the mvapich2 compiler wrappers to work properly when run from outside spack.

I've double checked on the compiler implicit rpaths and we already filter out system paths from that variable, so we won't be adding unreasonable rpaths to system locations.

tgamblin (approved):

PR #35382, opened 2023-02-08
py-scipy: use --config-settings

adamjstewart:
@rgommers @eli-schwartz @FFY00 If I'm understanding correctly, the latest versions of meson-python support passing options via `--config-settings`, so we can remove the hacks we had in place to build scipy. Hopefully I got the syntax correct, may take a few iterations.

### DIFF
--- a/var/spack/repos/builtin/packages/py-scipy/package.py
+++ b/var/spack/repos/builtin/packages/py-scipy/package.py
@@ -198,10 +198,10 @@ class PyScipy(PythonPackage):
             lapack += "-dynamic-lp64-seq"
 
         return {
-            "blas": blas,
-            "lapack": lapack,
-            "debug": False,
-            "optimization": 2,
+            "setup-args=-Dblas": blas,
+            "setup-args=-Dlapack": lapack,
+            "setup-args=-Ddebug": False,
+            "setup-args=-Doptimization": 2,
         }
 
     @run_after("install")
--- a/var/spack/repos/builtin/packages/py-scipy/package.py
+++ b/var/spack/repos/builtin/packages/py-scipy/package.py
@@ -3,9 +3,6 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-import glob
-import os
-
 from spack.package import *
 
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-scipy/package.py
```diff
@@ -206,29 +197,12 @@ def install(self, spec, prefix):
...
-        ]
-        python(*args)
-        args = std_pip_args + ["--prefix=" + prefix, glob.glob(join_path("dist", "scipy*.whl"))[0]]
-        pip(*args)
+        return {
+            "blas": blas,
+            "lapack": lapack,
+            "debug": False,
+            "optimization": 2,
+        }
```
eli-schwartz: https://meson-python.readthedocs.io/en/latest/reference/config-settings.html

meson-python expects config-settings for "setup-args", the value of which is a series of `-Dopt=val` arguments to Meson.

eli-schwartz (changes requested):

var/spack/repos/builtin/packages/py-scipy/package.py
adamjstewart: So the flag would look something like:
```console
$ pip install --config-settings=setup-args=-Dblas=openblas
```
? Boy that's a mouthful.

eli-schwartz: With `python -m build` it's a bit simpler because `--config-settings` becomes `-C`.

But yeah, kind of awkward.

adamjstewart: Okay, see if the new version looks better. Not sure whether multiple args would look like:
```console
$ pip install --config-settings=setup-args=-Dblas=openblas --config-settings=setup-args=-Dlapack=openblas
```
or
```console
$ pip install --config-settings=setup-args="-Dblas=openblas -Dlapack=openblas"
```

rgommers: The problem here is that `pip` still does not accept duplicate keys. We're in the process of fixing that - my edit to PEP 517 just got accepted. So you need to use `python -m build` to build a wheel, and then `pip install dist/scipy*.whl` to install it.

adamjstewart: Would my latter syntax avoid the duplicate key issue?

rgommers: No, there will be no string parsing/splitting.

adamjstewart: Alright, I'll just close this PR until pip properly supports meson.

eli-schwartz: Oh, yay... Did not realize pip still hasn't fixed this.

(I prefer to use `build` myself, and install with `installer`, because it avoids various automagic behaviors of pip when installing to environments that are *managed* by a tool other than pip.)

adamjstewart: I also like the idea of build/installer, but based on #35326, we may be a ways off from using them in Spack instead of pip.

PR #35384, opened 2023-02-08
dd4hep: new version 1.24, depends_on podio@0.16:

wdconinc:
No build system changes. Just a new requirement for podio.

### DIFF
--- a/var/spack/repos/builtin/packages/dd4hep/package.py
+++ b/var/spack/repos/builtin/packages/dd4hep/package.py
@@ -24,6 +24,7 @@ class Dd4hep(CMakePackage):
     tags = ["hep"]
 
     version("master", branch="master")
+    version("1.24", sha256="361a932b9af2479458c0759281fef0161439d8bd119da426ce462a0467adc679")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/dd4hep/package.py
```diff
@@ -152,6 +153,7 @@ class Dd4hep(CMakePackage):
     depends_on("lcio", when="+lcio")
     depends_on("edm4hep", when="+edm4hep")
     depends_on("podio", when="+edm4hep")
+    depends_on("podio@0.16:", when="@1.24:" +edm4hep")
```
vvolkl:
```suggestion
    depends_on("podio@0.16:", when="@1.24: +edm4hep")
```
I think this is a typo, otherwise looks good to me, thanks!

wdconinc: Yes, typo, auto-quotes when I'm on my phone...

tldahlgren (approved):
Confirmed sha256, rest LGTM (post-fix).

PR #35386, opened 2023-02-08
Fix build for scine-xtb

awvwgk:
Fixes upstream issue https://github.com/qcscine/xtb_wrapper/issues/4 when using meson built xtb as provided by spack.


ltalirz:
Tested as well, works for me now. Thanks for fixing @awvwgk !

alalazo (approved):

alalazo (approved):

PR #35391, opened 2023-02-08
Bugfix: fix build of xforms

stefanfechter:
This additional patch fixes the build of the now unmaintained library xforms.

### DIFF
--- a/var/spack/repos/builtin/packages/xforms/package.py
+++ b/var/spack/repos/builtin/packages/xforms/package.py
@@ -20,6 +20,7 @@ class Xforms(AutotoolsPackage):
     depends_on("libx11", type="link")
     depends_on("libxpm", type="link")
     depends_on("jpeg", type="link")
+    patch("xformsPatch.patch")
 
     def configure_args(self):
         args = ["--enable-static"]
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/xforms/package.py
```diff
@@ -20,6 +20,7 @@ class Xforms(AutotoolsPackage):
     depends_on("libx11", type="link")
     depends_on("libxpm", type="link")
     depends_on("jpeg", type="link")
+    patch("xformsPatch.patch")
```
tldahlgren, line 23: Do you know if the patch applies to all listed versions or for some version on?

See the docs at https://spack.readthedocs.io/en/latest/packaging_guide.html#patches.

tldahlgren:
@ketsubouchi As the original contributor, do you want to comment on this PR?

tldahlgren:
@spackbot run pipeline

var/spack/repos/builtin/packages/xforms/package.py
stefanfechter: That patch is needed for the last version of xforms, 1.2.4 . Not sure about the other versions though.

tldahlgren: Do you want to add the `when="@1.2.4"` argument to the patch then?

tldahlgren (approved):

PR #35244, opened 2023-01-30
[WRF] Adding new versions

t-brown:

PR #35401, opened 2023-02-08
wrf: fix v4.4.2 checksum

adamjstewart:
The WRF 4.4.2 checksum seems to have changed since it was added in #35244. Not sure if this is due to the recent GitHub hashing change or not. I have a feeling it isn't since that should only affect the auto-generated tarballs, not the manually uploaded assets. Should I open an issue with the WRF developers to investigate?

(the previous checksum does not match any of the other tar or zip, auto or manual downloads)

CI is currently failing because of this: https://gitlab.spack.io/spack/spack/-/jobs/5466213

@t-brown @tldahlgren

### DIFF
--- a/var/spack/repos/builtin/packages/wrf/package.py
+++ b/var/spack/repos/builtin/packages/wrf/package.py
@@ -73,7 +73,7 @@ class Wrf(Package):
 
     version(
         "4.4.2",
-        sha256="5d6237f1500c44a33626362936ba0a4388360c5070d9d53262e5a950c586da85",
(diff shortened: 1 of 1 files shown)


MichaelLaufer (approved):
Looking at the WRF releases page:
https://github.com/wrf-model/WRF/releases

It appears that the v4.4.2.tar.gz file was updated/reuploaded just 5 days ago (so after @t-brown 's PR, https://github.com/spack/spack/pull/35244), I assume this explains what we are seeing.
That being said it looks like a 1-time issue, and we can proceed with updating the hash on our end.

tldahlgren (approved):
Confirmed the new sha256.

adamjstewart:
@spackbot run pipeline

PR #35402, opened 2023-02-08
libgd: add libx11 dependency

frankwillmore:
adding explicit dependency on X11 (needed when there is not a system install)


tldahlgren (approved):
LGTM

tldahlgren:
@spackbot run pipeline

PR #29835, opened 2022-04-01
ASP-based solver: allow configuring target selection

alalazo:
fixes #29672 

This commit adds a new `concretizer:targets` configuration section, and two options under it.
 1.  `concretizer:targets:granularity` allows switching from considering only generic targets to consider all possible microarchitectures.
 2. `concretizer:targets:host_compatible` instead controls whether we can concretize for microarchitectures that are incompatible with the current host.

Modifications:

- [x] Add new handles to configure the solver
- [x] Add documentation
- [x] Add unit tests

Issue #17239, opened 2020-06-24
Cross-compiling detection/stored in the spec object


naromero77:
In support of large scale supercomputer and even mid-tier linux clusters, it would be desirable to store in the `spec` whether the the current spack install invocation is a cross-compiling scenario.

### Rationale

Here is a concrete use case:
https://github.com/spack/spack/pull/17180

Quantum ESPRESSO uses autoconf and does not have much of a concept of cross-compilation AFAIK. You have to manually add `--host` if you are in a cross-compile scenario. This flag suppresses autoconf from running config tests on the login node. That is all it does.

For package developers, it would be desirable if in the `spec` there was a boolean that took into account the architecture/platform and the compilation target to determine whether we are in a cross-compilation scenario.

### Description

The platform/architecture object stores information about the frontend and backend for both the os and the target. I think it would be simple if the `spec` where to take this information along with the target and determine if the package needs to be cross-compiled.

Both @becker33 and @alalazo where helpful for the aforementioned pull request. It is possible to detect cross-compilation fairly easily with `spec.satisfies(...)` for BG/Q and Cray, but it does not cover a Linux cluster with host login nodes that are different in architecture from the compute nodes. Also, cross-compilation is not as simple as just comparing target of the frontend versus the backend -- possibly also need to compare the os. We haven't even added accelerated into the mix here.

It would of been possible for me to just hack something together to get this done based on @alalazo suggestion, but it would be more desirable to have a consistent way for package developers to do this.

### General information

naromero@thetalogin6:~/spack> spack --version
0.14.2-1489-a20ba315f

- [X] I have run `spack --version` and reported the version of Spack
- [X] I have searched the issues of this repo and believe this is not a duplicate

commodo:
@naromero77 @tgamblin @alalazo @becker33 

i'm also interested in this cross-compilation aspect;
especially cross-compiling between 2 different architectures;
the use-case is in the direction of embedded devices;
while i understand that Spack is oriented towards HPC, we are evaluating Spack for the embedded use-case;

right now where we are is:
* we're able to cross-compile some packages to ARM64 (from an x86_64 host)
* the packages that don't cross-compile need a host-compatible compiler to build certain build-time tools

typically, (as you may know already) embedded build-systems make use of the `--host=` & `--target` directives (when building these build-time tools)
i think we should do something similar here;

at least for the embedded use-case, what would be sufficient:
[...]

PR #35405, opened 2023-02-08
Add docs on setting up cross-compilers

wyphan:
Specifically, this is added to the section on manual compiler configuration.


lib/spack/docs/getting_started.rst
```diff
@@ -459,6 +459,31 @@ specification. The operations available to modify the environment are ``set``, `
...
+         f77: /usr/bin/aarch64-linux-gnu-gfortran-9
+         fc: /usr/bin/aarch64-linux-gnu-gfortran-9
+       flags: {}
+       operating_system: ubuntu20.04
+       target: aarch64
+       modules: []
+       environment: {}
+       extra_rpaths: []
+
+You might also need to adjust the concretizer setting to allow building for a
```
ltalirz: Are there cases where this is not required?

At least for cross-microarchitecture compilations, this is actually the only change required, so perhaps consider moving it before the compiler spec example.

ltalirz (commented):
Thanks @wyphan for this addition!

Just one comment: it may be worth adding a sentence on cross-compiling for a different microarchitectures.

For example, on my system compilers are added with target `x86_64` by default, and in order to be able to compile for `target=zen3` on a `haswell` system, I only need to do the `spack config` change (no need to modify compiler specs).

lib/spack/docs/getting_started.rst
wyphan: Hmm, it could be moved up before the compiler spec example, or moved to `lib/spack/docs/basic_usage.rst` under "Support for specific microarchitectures". @scheibelp what do you think?

haampie:
Spack doesn't support cross-compilation, so I don't think we should imply with docs that it can.

lib/spack/docs/getting_started.rst
```diff
@@ -459,6 +459,31 @@ specification. The operations available to modify the environment are ``set``, `
          prepend_path: # Similar for append|remove_path
            LD_LIBRARY_PATH: /ld/paths/added/by/setvars/sh
 
+Finally, if you intend to cross-compile for a different system architecture,
```
scheibelp: @haampie mentions

> Spack doesn't support cross-compilation, so I don't think we should imply with docs that it can.

I think this can be addressed with more-precise language: replace

> if you intend to cross-compile for a different system architecture,

with

> if you want to compile for a different target,

lib/spack/docs/getting_started.rst
scheibelp: I agree the added content should 

* mention `spack config add concretizer:targets:host_compatible:false` first and then follow up
* eliminate the word "might" from "you might also need to" (since in fact this is necessary)

lib/spack/docs/getting_started.rst
```diff
@@ -459,6 +459,31 @@ specification. The operations available to modify the environment are ``set``, `
...
+       flags: {}
+       operating_system: ubuntu20.04
+       target: aarch64
+       modules: []
+       environment: {}
+       extra_rpaths: []
+
+You might also need to adjust the concretizer setting to allow building for a
+different architecture than the host architecture using
+``spack config add concretizer:targets:host_compatible:false``
```
scheibelp: Where possible I want to move items out of the "getting started" section, in this case I think this could be added to `_concretizer-options` (in `build_settings.rst`).

scheibelp (changes requested):
I'd like to move this section and adjust the language a bit (specifics in comments)

lib/spack/docs/getting_started.rst
wyphan: Turns out it's already documented there. I just added a link to point there.

commodo:
so, to chime in:  Spack supports cross-compilation (to an extent);
and has been supporting this for a while;
( by cross-compilation i mean cross-architecture compilation )

firstly: there's a flag (called `host_compatible` ; seems to have been added here https://github.com/spack/spack/pull/29835) which seems to be enabled by default;
this disallows any target that is incompatible with the host;
so, just configuring a compiler like mentioned here (in the docs), is insufficient; the `host_compatible` flag will stop the build early;

back in 2019 there were 2 discussions ([this one about trying to cross-compile boost](https://groups.google.com/g/spack/c/2sa-bWvixOk) and [this one about trying to cross compile a GCC compiler](https://groups.google.com/g/spack/c/aFsGIhsDKCg/m/OtzR__unCgAJ))

we're also interested in this use-case;
so, i've tested cross-compilation of a few packages, and some work; other don't work
particularly the ones that need host-compilers to build some build-time tools;

i've found this issue (https://github.com/spack/spack/issues/17239)  that starts to tackle this into a direction;
i'll maybe continue there;

PR #35406, opened 2023-02-08
Update PennyLane and PennyLane Lightning

AmintorDusko:
Hi Folks,
We (@mlxd and I) are Pennylane Lightning developers and prepared this update for PennyLane and PennyLane Lightning.
This PR also adds the py-flaky package important for our python test suite.

Thank you @marcodelapierre for your initial PR, and @adamjstewart for the quick review.

Please let me know your thoughts.


mlxd:
> @mlxd can you review this PR?
> 
> This PR modifies the following package(s), for which you are listed as a maintainer:
> 
>     * py-pennylane
> 
>     * py-pennylane-lightning

Just a note, as an author on this PR also I'll decline the review request.

var/spack/repos/builtin/packages/py-flaky/package.py
```diff
@@ -0,0 +1,19 @@
...
+
+class PyFlaky(PythonPackage):
+    """Flaky is a plugin for nose or pytest that automatically reruns flaky tests."""
+
+    homepage = "https://github.com/box/flaky"
+    pypi = "flaky/flaky-3.7.0.tar.gz"
+
+    version("3.7.0", sha256="3ad100780721a1911f57a165809b7ea265a7863305acb66708220820caf8aa0d")
+
+    depends_on("python@3.5:", type=("build", "run"))
```
adamjstewart: The setup.py says 2.7, 3.4+. Spack only supports 3.7+, so you could just remove this line and nothing would change.

var/spack/repos/builtin/packages/py-flaky/package.py
```diff
@@ -0,0 +1,19 @@
...
+    depends_on("py-setuptools", type=("build"))
+    depends_on("py-tox", type=("test"))
```
adamjstewart: This package doesn't run any tests, so there's no reason to include test dependencies, I would just remove this.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,123 @@
...
+    tag = "v0.28.2"
 
-    version("0.28.0", sha256="f5849c2affb5fb57aca20feb40ca829d171b07db2304fde0a37c2332c5b09e18")
+    maintainers = ["mlxd", "AmintorDusko"]
 
-    depends_on("python@3.8:", type=("build", "run"))
-    depends_on("py-setuptools", type="build")
-    depends_on("cmake@3.16:", type="build")
+    version("develop", branch="master")
+    version("0.28.2", sha256="c9b3afed0585681ccaf4df09fb12f2b7f09a8a1ba97a9b979139fe4a24509f31")
```
adamjstewart: Can you also include 0.28.0? We don't usually remove old versions without first deprecating them.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,123 @@
...
-    url = "https://github.com/PennyLaneAI/pennylane-lightning/archive/refs/tags/v0.28.0.tar.gz"
-    # using github for now, because pypi tarball is missing the CMakeLists.txt file
-    # pypi = "PennyLane-Lightning/PennyLane-Lightning-0.28.0.tar.gz"
+    homepage = "https://docs.pennylane.ai/projects/lightning"
+    git = "https://github.com/PennyLaneAI/pennylane-lightning.git"
+    url = "https://github.com/PennyLaneAI/pennylane-lightning/archive/refs/tags/v0.28.2.tar.gz"
+    tag = "v0.28.2"
 
-    version("0.28.0", sha256="f5849c2affb5fb57aca20feb40ca829d171b07db2304fde0a37c2332c5b09e18")
+    maintainers = ["mlxd", "AmintorDusko"]
```
adamjstewart:
```suggestion
    maintainers("mlxd", "AmintorDusko")
```
We have a new directive that appends instead of overwrites, so I will still be listed as a maintainer too.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,123 @@
...
 
-    depends_on("python@3.8:", type=("build", "run"))
-    depends_on("py-setuptools", type="build")
-    depends_on("cmake@3.16:", type="build")
+    version("develop", branch="master")
```
adamjstewart:
```suggestion
    version("master", branch="master")
```

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,123 @@
...
+        "build_type",
+        default="Release",
+        description="CMake build type",
+        values=("Debug", "Release", "RelWithDebInfo", "MinSizeRel"),
+    )
+
+    extends("python", when="+python")
+
+    # hard dependencies
+    depends_on("cmake@3.21.0:3.24.0", type="build")  # 3.21-3.24
```
adamjstewart: I don't often see upper bounds on cmake versions. Do newer versions not work?

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,123 @@
...
+
+    depends_on("python@3.8:", type=("build", "run"), when="+python")
+    depends_on("py-setuptools", type="build", when="+python")
+    depends_on("py-numpy", type=("build", "run"), when="+python")
+    depends_on("py-pybind11", type=("build"), when="+python")
+    depends_on("py-pip", type="build", when="+python")
+    depends_on("py-wheel", type="build", when="+python")
+
+
+class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
```
adamjstewart, line 70: We don't normally use separate Builders unless a package supports multiple build systems. You should be able to include all of this in the normal package class if you want.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,123 @@
...
+        if self.spec.variants["python"].value:
+            cm_args = ";".join(
+                [
+                    s[2:]
+                    for s in self.cmake_args()
+                    if s[2:] not in ["BUILD_TESTS:BOOL=ON", "BUILD_BENCHMARKS:BOOL=ON"]
+                ]
+            )
+            args = ["-i", f"--define={cm_args}"]
+            build_ext = Executable(f"{self.spec['python'].command.path}" + " setup.py build_ext")
```
adamjstewart: Direct invocation of setup.py is deprecated: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

var/spack/repos/builtin/packages/py-pennylane-lightning/v0.28-spack_support.patch
```diff
@@ -0,0 +1,1076 @@
+diff --git a/.github/CHANGELOG.md b/.github/CHANGELOG.md
```
adamjstewart: This patch is massive and contains many things that aren't required for Spack. Could this be minimized a bit?

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -10,25 +10,45 @@
...
 
-    homepage = "https://github.com/XanaduAI/pennylane"
+    homepage = "https://docs.pennylane.ai/"
+    git = "https://github.com/PennyLaneAI/pennylane.git"
     pypi = "PennyLane/PennyLane-0.28.0.tar.gz"
 
-    maintainers("marcodelapierre")
+    maintainers = ["mlxd", "AmintorDusko"]
 
+    version("develop", branch="master")
```
adamjstewart:
```suggestion
    version("master", branch="master")
```

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -10,25 +10,45 @@
 class PyPennylane(PythonPackage):
     """PennyLane is a Python quantum machine learning library by Xanadu Inc."""
 
-    homepage = "https://github.com/XanaduAI/pennylane"
+    homepage = "https://docs.pennylane.ai/"
+    git = "https://github.com/PennyLaneAI/pennylane.git"
     pypi = "PennyLane/PennyLane-0.28.0.tar.gz"
 
-    maintainers("marcodelapierre")
+    maintainers = ["mlxd", "AmintorDusko"]
```
adamjstewart:
```suggestion
    maintainers("mlxd", "AmintorDusko")
```
@marcodelapierre do you still want to be listed as a maintainer?

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -10,25 +10,45 @@
...
+    depends_on("py-autograd", type="run")
+    depends_on("py-autoray", type="run")
+    depends_on("py-cachetools", type="run")
+    depends_on("py-networkx", type="run")
+    depends_on("py-numpy@:1.23", type="run")
+    depends_on("py-pennylane-lightning+python", type="run")
+    depends_on("py-retworkx", type="run")
+    depends_on("py-requests", type="run")
+    depends_on("py-scipy", type="run")
+    depends_on("py-toml", type="run")
```
adamjstewart: These should be `type=("build", "run")`, otherwise the tests won't work because these dependencies won't exist in `PYTHONPATH`.

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -10,25 +10,45 @@
...
+
+    # Test deps
+    depends_on("py-pytest", type=("run", "test"))
+    depends_on("py-pytest-xdist", type=("run", "test"))
+    depends_on("py-pytest-mock", type=("run", "test"))
+    depends_on("py-flaky", type=("run", "test"))
```
adamjstewart: Are these required at run-time, or only at test-time?

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -10,25 +10,45 @@
...
-    depends_on("py-autograd", type=("build", "run"))
-    depends_on("py-toml", type=("build", "run"))
-    depends_on("py-appdirs", type=("build", "run"))
     depends_on("py-semantic-version@2.7:", type=("build", "run"))
-    depends_on("py-autoray@0.3.1:", type=("build", "run"))
-    depends_on("py-cachetools", type=("build", "run"))
-    depends_on("py-pennylane-lightning@0.28:", type=("build", "run"))
-    depends_on("py-requests", type=("build", "run"))
+    depends_on("py-pip", type="build")
+    depends_on("py-wheel", type="build")
```
adamjstewart: These are already added by the base class, you don't need to re-add them.

var/spack/repos/builtin/packages/py-pennylane/package.py
adamjstewart: Also, it's much quicker to review these if they are in the same order as in `setup.py` (as they were before)

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-pennylane/package.py
marcodelapierre: yes, if you guys don't mind @mlxd @AmintorDusko ? 
otherwise no worries at all!

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Hi @adamjstewart Would this be as easy as adding them, but assuming they do not get used? We do not anticipate any user to avail of 0.28.0 and 0.28.1, as 0.28.2 fixes known bugs in these releases.

Is there a process for marking these as deprecated if we add them here? To ensure they work will require additional patches which would bring them in line to the changes in 0.28.2 anyway.

What do you think is the best path?

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Recent issues we encountered with 3.25.0 and our Lightning suite caused us to hard-pin to 3.24 across our build ecosystem.

As 3.25.2 is now released, I do not see the issue. I can forbid the specific version here, rather than provide the upper bound.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: As `build_ext` is not exposed to `python -m build` (as far as I am aware), would an editable pip installation or a wheel be preferred instead?

We have no immediate plans to migrate the build system to use `pyproject.toml` and `setup.cfg`. Will this be an issue for consideration to be approved?

var/spack/repos/builtin/packages/py-pennylane/package.py
mlxd: Happy for @marcodelapierre to remain on the list

var/spack/repos/builtin/packages/py-pennylane/package.py
mlxd: We try to maintain dependencies in alphabetical order internally, but happy to go with whatever is preferred here.

Happy to make the change to ("build, "run"). Thought, the tests seem to run fine post-install for us using `spack install --verbose --test=root py-pennylane`. Is there a preferred way for us to validate these when we are testing the package internally?

var/spack/repos/builtin/packages/py-pennylane/package.py
mlxd: Test-time only.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
adamjstewart: Keeping the line would be sufficient. There is a procedure for deprecation too: https://spack.readthedocs.io/en/latest/packaging_guide.html#deprecating-old-versions

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
adamjstewart: @pradyunsg may have more ideas than me

var/spack/repos/builtin/packages/py-pennylane/package.py
adamjstewart: There are two types of tests in Spack.

1. Install tests: `spack install --test=root foo`
2. Post-install tests: `spack test run foo`

Unless you add additional tests yourself, these both attempt to import all Python files (for a pure Python package). At install-time, only dependencies listed as `type="build"` will be in `PYTHONPATH`. For post-install tests, only dependencies listed as `type="run"` will be in `PYTHONPATH`.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Great, thanks for the info -- wasn't aware of the deprecation arg.

Will add those back in and mark as `deprecated=True`.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Thanks @adamjstewart 

For note, our usual method for delivery is PyPI wheels, built through CIBuildWheel. If it is simpler to opt for that pathway, happy to do so, and run a post-build installation of the wheel.

Alternatively, we can forgo the Python builder altogether, and aim to just build the module using CMake, and manually move it after the fact. Though, I think adhering to the Python way above is probably more maintainable, at least in the interim.

var/spack/repos/builtin/packages/py-pennylane/package.py
mlxd: Ah, I understand, thanks. The above was only validated against option 1.

Will check and make the changes.

var/spack/repos/builtin/packages/py-pennylane/package.py
AmintorDusko: Hi @marcodelapierre, I'm happy to have you as a maintainer. 
Thank you for that!

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,128 @@
 from spack.package import *
 
 
-class PyPennylaneLightning(PythonPackage):
-    """PennyLane-Lightning plugin"""
+class PyPennylaneLightning(CMakePackage, PythonExtension):
```
adamjstewart, line 10: Since the package is now a CMake tool with optional Python bindings, I think it would make sense to rename it to `pennylane-lightning`. Another option would be to have both `pennylane-lightning` and `py-pennylane-lightning` packages where the former is the CMake part and the latter is only the Python bindings and depends on the former. Not sure if it's possible to build your Python bindings with a separately installed CMake part or if they have to be installed simultaneously.

mlxd: It is a good question. Right now, we have an explicitly tight coupling with the Python package, and so that is our delivery and consumption mechanism. We expect this to remain the case for most of 2023, so likely separating this out will require some decent effort on our end.

I would be keen to keep the structure as is for now, assuming this isn't a problem.

adamjstewart: I don't think `py-pennylane-lightning~python` makes sense. See https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#choosing-a-package-name for the official docs. I think it makes the most sense to rename the package. We can think about splitting them into separate packages later.

mlxd: Since the package comes from https://pypi.org/project/PennyLane-Lightning/ as the intended use, the reason to disable the Python layer is simply to speed up compilation to generate test and benchmarking binaries at the CPP layer. It is not possible to use the compiled library itself without the Python layer, so it isn't optional in this case.

The internal kernels can be compiled separately and validated, hence the `+python` flag option in this instance.
I think keeping the Python naming makes sense as it seems to fall into the same structure as listed on the link with py-genders  py-py2cairo, py-pygobject, etc.

mlxd: I'll discuss with our team internally and see if this split makes sense now rather than later.

adamjstewart: In that case I'm confused what the point of this PR is. You're adding a great deal of complexity to the package but the end goal is still to build the exact same Python library. I was under the impression that the change to the CMakePackage base class was to support greater flexibility and allow for C++-only installations, but it seems like the thing getting installed is still the same. We could remove the `+python` variant and always installs the Python portion I suppose.

mlxd: Having access to Spack in this case allows us to build the Python module with the configurability introduced by Spack.

We have tested manually compiling the package on a variety of HPC systems through different compilers, and having the variants allows us to tailor-suit the Python package for each respective system (ie without or without AVX512 support, OpenMP, Kokkos, etc).

When choosing a set of variants, the binary options give us the up-front information to understanding if the full test-suite passes on the CPP side, as well as how the kernels perform under the hood.

I'm fine dropping the +python variant and always ensuring the Python module is built, if you think this is the best situation here. We do not plan to separate out the C++ only components into a separate module yet, as this also requires plenty of retooling on the package itself. It may be a future change, but not part of our roadmap just yet. Hope that clarifies.

adamjstewart: Sounds good, let's just drop `+python` for now.

mlxd: Sure thing, and thanks for the discussion on this.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,124 @@
...
+        cm_args = ";".join(
+            [
+                s[2:]
+                for s in self.cmake_args()
+                if s[2:] not in ["BUILD_TESTS:BOOL=ON", "BUILD_BENCHMARKS:BOOL=ON"]
+            ]
+        )
+        args = ["-i", f"--define={cm_args}"]
+        build_ext = Executable(f"{self.spec['python'].command.path}" + " setup.py build_ext")
+        build_ext(*args)
```
adamjstewart: Still curious to see if @pradyunsg knows of a way to pass this option via pip instead of invoking setup.py.

mlxd: Hi @adamjstewart is there an update here? We explored the use of the Python `build` package as an alternative, but do not think this is an improvement over the `build_ext` we have above, due to the increase in verbosity: as an example 
```bash
python -m build --wheel --config-setting=--build-option='build_ext --define="CMAKE_CXX_COMPILER=clang-14;ENABLE_CLANG_TIDY=on"'
```
 would be required, instead of 
```bash
python setup.py build_ext --define="CMAKE_CXX_COMPILER=clang-14;ENABLE_CLANG_TIDY=on"
```
Since this isn't an issue for anything pre Python 3.12, which we do not aim to support until sometime in 2024, I think we can keep the above for now, as many other packages are doing likewise, and aim to identify a cleaner solution that is more maintainable long-term.

In addition, we have updated the package file to the newest release of both PennyLane and PennyLane-Lightning. Feel free to let us know if this meets the cut.

adamjstewart: I'm fine with keeping this as is for now, we can improve it at a later date as needed.

var/spack/repos/builtin/packages/py-pytest-xdist/package.py
```diff
@@ -10,8 +10,9 @@ class PyPytestXdist(PythonPackage):
     """py.test xdist plugin for distributed testing and loop-on-failing mode"""
 
     homepage = "https://github.com/pytest-dev/pytest-xdist"
-    pypi = "pytest-xdist/pytest-xdist-1.30.0.tar.gz"
+    pypi = "pytest-xdist/pytest-xdist-3.2.0.tar.gz"
 
+    version("3.2.0", sha256="fa10f95a2564cd91652f2d132725183c3b590d9fdcdec09d3677386ecf4c1ce9")
     version("1.30.0", sha256="5d1b1d4461518a6023d56dab62fb63670d6f7537f23e2708459a557329accf48")
     version("1.29.0", sha256="3489d91516d7847db5eaecff7a2e623dba68984835dbe6cedb05ae126c4fb17f")
     version("1.27.0", sha256="a96ed691705882560fa3fc95531fbd4c224896c827f4004817eb2dcac4ba41a2")
```
adamjstewart, line 18: Dependencies need to be updated:
```toml
requires = [
  "setuptools>=45.0",
  "setuptools-scm[toml]>=6.2.3",
  "wheel",
]
```
```
install_requires =
        execnet>=1.1
        pytest>=6.2.0
```

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -28,7 +40,23 @@ class PyPennylane(PythonPackage):
     depends_on("py-toml", type=("build", "run"))
     depends_on("py-appdirs", type=("build", "run"))
     depends_on("py-semantic-version@2.7:", type=("build", "run"))
-    depends_on("py-autoray@0.3.1:", type=("build", "run"))
+    depends_on("py-autoray", type=("build", "run"))
     depends_on("py-cachetools", type=("build", "run"))
-    depends_on("py-pennylane-lightning@0.28:", type=("build", "run"))
+    depends_on("py-pennylane-lightning", type=("build", "run"))
```
adamjstewart: Why were these changed? The lower bound seems correct from the setup.py

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -10,12 +10,24 @@
...
-
-    version("0.28.0", sha256="2a6100c00277c1eb59eab6856cdad7b1237e9d1fbda98b1e15020bd5a64b10a8")
+    maintainers("mlxd", "AmintorDusko", "marcodelapierre")
+
+    version("master", branch="master")
+    version("0.29.1", sha256="c5d662994b741afa69e4fdadc79a1b75840275138a8b7e0cfc5fd64b66a12eef")
+    version(
+        "0.29.0",
+        sha256="055fc22a2624cceddbb8bd25d7eda14d2dd7e0afa08ea8555c23bc75eff007a5",
+        deprecated=True,
```
adamjstewart: Why add 0.29.0 if it's deprecated?

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,131 @@
...
+    """The PennyLane-Lightning plugin provides a fast state-vector simulator written in C++."""
 
-    homepage = "https://github.com/PennyLaneAI/pennylane-lightning"
-    url = "https://github.com/PennyLaneAI/pennylane-lightning/archive/refs/tags/v0.28.0.tar.gz"
-    # using github for now, because pypi tarball is missing the CMakeLists.txt file
-    # pypi = "PennyLane-Lightning/PennyLane-Lightning-0.28.0.tar.gz"
+    homepage = "https://docs.pennylane.ai/projects/lightning"
+    git = "https://github.com/PennyLaneAI/pennylane-lightning.git"
+    url = "https://github.com/PennyLaneAI/pennylane-lightning/archive/refs/tags/v0.28.2.tar.gz"
+    tag = "v0.28.2"
```
adamjstewart: This variable doesn't seem to be used

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,131 @@
...
-    version("0.28.0", sha256="f5849c2affb5fb57aca20feb40ca829d171b07db2304fde0a37c2332c5b09e18")
+    maintainers("mlxd", "AmintorDusko")
+
+    version("master", branch="master")
+
+    version("0.29.0", sha256="da9912f0286d1a54051cc19cf8bdbdcd732795636274c95f376db72a88e52d85")
+    version(
+        "0.28.2",
+        sha256="c9b3afed0585681ccaf4df09fb12f2b7f09a8a1ba97a9b979139fe4a24509f31",
+        deprecated=True,
```
adamjstewart: Why add 0.28.2 if it's deprecated?

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,131 @@
...
+    depends_on("kokkos-kernels@3.7.00", when="+kokkos")
+    depends_on("llvm-openmp", when="+openmp %apple-clang")
 
     depends_on("python@3.8:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
-    depends_on("cmake@3.16:", type="build")
-
-    depends_on("py-ninja", type=("build", "run"))
     depends_on("py-numpy", type=("build", "run"))
-    # depends_on("py-pennylane@0.19:", type=("build", "run"))  # circular dependency
```
adamjstewart, line 26: Would be good to keep this comment so I don't ask where the dep is every time

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
adamjstewart: This comment is still unaddressed

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Sounds good, thanks!

var/spack/repos/builtin/packages/py-pennylane/package.py
mlxd: We didn't want a discontinuity in the history of versions listed, but I agree this isn't sensible since there's no historical availability of the package. Will remove it.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Same as above. Will remove this.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: We do use both Make and CMake in the package, though currently the CMake support is more mature and generalised. We will likely address the support for both in a follow-up PR, but this structure will likely be the base of said update. If you'd prefer, we can undo this and stick to the normal package structure for now, and reintroduce later.

var/spack/repos/builtin/packages/py-pennylane/package.py
mlxd: I'll reintroduce these.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,118 @@
...
+
+class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
+    build_directory = "build"
+
+    def cmake_args(self):
+        """
+        Here we specify all variant options that can be dynamicaly specified at build time
+        """
+        args = [
+            self.define_from_variant("CMAKE_BUILD_TYPE", "build_type"),
```
adamjstewart: This is already added automatically by the base class so you can remove it.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,118 @@
...
+            self.define_from_variant("ENABLE_OPENMP", "openmp"),
+            self.define_from_variant("ENABLE_NATIVE", "native"),
+            self.define_from_variant("ENABLE_BLAS", "blas"),
+            self.define_from_variant("CMAKE_VERBOSE_MAKEFILE:BOOL", "verbose"),
+            self.define_from_variant("BUILD_TESTS", "cpptests"),
+            self.define_from_variant("BUILD_BENCHMARKS", "cppbenchmarks"),
+            self.define_from_variant("ENABLE_GATE_DISPATCHER", "dispatcher"),
+        ]
+
+        if self.spec.variants["kokkos"].value:
```
adamjstewart:
```suggestion
        if "+kokkos" in self.spec:
```
Simpler way of doing this

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,118 @@
...
+        pip_args = std_pip_args + ["--prefix=" + prefix, "."]
+        pip(*pip_args)
+        super().install(pkg, spec, prefix)
+
+    @run_after("install")
+    @on_package_attributes(run_tests=True)
+    def test_lightning_build(self):
+        with working_dir(self.stage.source_path):
+            pl_runner = Executable(
+                join_path(self.prefix, "bin", "pennylane_lightning_test_runner")
```
adamjstewart:
```suggestion
                self.prefix.bin.pennylane_lightning_test_runner
```
Equivalent

var/spack/repos/builtin/packages/py-pennylane/package.py
```diff
@@ -28,7 +35,23 @@ class PyPennylane(PythonPackage):
     depends_on("py-toml", type=("build", "run"))
     depends_on("py-appdirs", type=("build", "run"))
     depends_on("py-semantic-version@2.7:", type=("build", "run"))
-    depends_on("py-autoray@0.3.1:", type=("build", "run"))
+    depends_on("py-autoray", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-autoray@0.3.1:", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-pytest-xdist/package.py
```diff
@@ -20,11 +21,15 @@ class PyPytestXdist(PythonPackage):
...
+    depends_on("py-setuptools-scm@6.2.3: +toml", type="build", when="@3.2.0:")
     depends_on("py-execnet@1.1:", type=("build", "run"))
-    depends_on("py-pytest@4.4.0:", type=("build", "run"), when="@1.28.0:")
+    depends_on("py-pytest@6.2.0:", type=("build", "run"), when="@3.2.0:")
+    depends_on("py-pytest@4.4.0:", type=("build", "run"), when="@1.28.0:1.30.0")
     depends_on("py-pytest@3.6.0:", type=("build", "run"), when="@1.25.0:1.27.0")
     depends_on("py-pytest@3.0.0:", type=("build", "run"), when="@1.18.0:1.24.0")
     depends_on("py-pytest@2.7.0:", type=("build", "run"), when="@1.16.0:1.17.0")
     depends_on("py-pytest-forked", type=("build", "run"), when="@1.19.0:")
     depends_on("py-six", type=("build", "run"), when="@1.23.0:")
```
adamjstewart:
```suggestion
    depends_on("py-six", type=("build", "run"), when="@1.23.0:1")
```

var/spack/repos/builtin/packages/py-pytest-xdist/package.py
```diff
@@ -20,11 +21,15 @@ class PyPytestXdist(PythonPackage):
...
+    depends_on("py-wheel", type="build", when="@3.20.0:")
```
adamjstewart: Technically this is already added by the base class so you don't need it.

var/spack/repos/builtin/packages/py-pytest-xdist/package.py
```diff
@@ -20,11 +21,15 @@ class PyPytestXdist(PythonPackage):
...
+    depends_on("py-setuptools", type="build", when="@:1.30.0")
+    depends_on("py-setuptools-scm@6.2.3: +toml", type="build", when="@3.2.0:")
     depends_on("py-execnet@1.1:", type=("build", "run"))
-    depends_on("py-pytest@4.4.0:", type=("build", "run"), when="@1.28.0:")
+    depends_on("py-pytest@6.2.0:", type=("build", "run"), when="@3.2.0:")
+    depends_on("py-pytest@4.4.0:", type=("build", "run"), when="@1.28.0:1.30.0")
     depends_on("py-pytest@3.6.0:", type=("build", "run"), when="@1.25.0:1.27.0")
     depends_on("py-pytest@3.0.0:", type=("build", "run"), when="@1.18.0:1.24.0")
     depends_on("py-pytest@2.7.0:", type=("build", "run"), when="@1.16.0:1.17.0")
     depends_on("py-pytest-forked", type=("build", "run"), when="@1.19.0:")
```
adamjstewart:
```suggestion
    depends_on("py-pytest-forked", type=("build", "run"), when="@1.19.0:1")
```

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Done.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Done.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Done.

var/spack/repos/builtin/packages/py-pennylane/package.py
mlxd: Done.

var/spack/repos/builtin/packages/py-pytest-xdist/package.py
mlxd: Done.

var/spack/repos/builtin/packages/py-pytest-xdist/package.py
mlxd: Good call -- removed.

var/spack/repos/builtin/packages/py-pytest-xdist/package.py
mlxd: Done.

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
```diff
@@ -7,20 +7,115 @@
...
+        cm_args = ";".join(
+            [
+                s[2:]
+                for s in self.cmake_args()
+                if s[2:] not in ["BUILD_TESTS:BOOL=ON", "BUILD_BENCHMARKS:BOOL=ON"]
+            ]
+        )
+        args = ["-i", f"--define={cm_args}"]
+        build_ext = Executable(f"{self.spec['python'].command.path}" + " setup.py build_ext")
+        build_ext(*args)
```
adamjstewart: Does this work?
```suggestion
        python("setup.py", "build_ext", *args)
```

adamjstewart (commented):
I promise this is my last review comment 😆

adamjstewart (approved):

var/spack/repos/builtin/packages/py-pennylane-lightning/package.py
mlxd: Sure, fine with me. Much cleaner too. Done.

mlxd:
Thanks for the detailed review @adamjstewart

AmintorDusko:
Thank you, @adamjstewart!

PR #35408, opened 2023-02-08
meme add perl-xml-simple dependency

simonLeary42:
Meme 4.5.0 has the first occurrence of the string
```
use XML::Simple
```
I found this by doing a binary search manually extracting tarballs until `grep` came up empty.

I would assume that this would be a runtime dependency only, but I don't know enough about perl to say that with 100% confidence.


tldahlgren (approved):
LGTM

PR #35409, opened 2023-02-08
gitlab: push public key to mirror after signing specs

scottwittenburg:
Assuming we can have the signing job export the public part of the reputational signing key, [here](https://github.com/spack/spack-infrastructure/pull/387), this change will make sure that public key is placed into the expected location in the mirror after the signing script finishes.  

Since pipelines follow up the signing job with a rebuild index job, which notably includes the `--keys` argument, that should take care of indexing the new key once it's in place.


scottwittenburg:
@kwryankrattiger I can't wait for that boiler plate reduction PR to land and put an end to PRs like this one!

haampie:
@scottwittenburg since this is not tested and won't be tested even if you drop the draft status I'm not going to backport it to 0.19.1

scottwittenburg:
> this is not tested and won't be tested even if you drop the draft status

Just making sure we're on the same page: you're right this won't be tested in the PR, and won't be tested until a protected pipeline runs with the change.



> I'm not going to backport it to 0.19.1

I think that's fine.  As I mentioned in chat, I can manually push public key anywhere we need it for the current point release.

kotfic (approved):
LGTM

haampie:
What's the status here?

Issue #22918, opened 2021-04-10
CMakePackage: Default Build Type


ax3l:
I realized the default build type for `CMakePackage` in spack is `build_type [RelWithDebInfo]`.

That's cool (I do the same for my projects) because `-g` comes with no real (runtime) cost and adds understanding on crashes; but there is an important detail: due to legacy reasons, most compilers are set up with `-O2` on that build type:
https://github.com/ECP-WarpX/WarpX/issues/1201
https://github.com/ECP-WarpX/WarpX/pull/1203

Are we aware of that and change this somewhere back to `-O3`? Otherwise `Release` might be a better default :)

More background in this thread: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/591

ax3l:
co-assigning people that did a lot of editing in the package and might be interested :)

sethrj:
I think for an HPC software package system the goal should eventually be the "most performant" build -- but since spack is still in development (no 1.0 tag yet) the extra debuggability of packages built using `RelWithDebInfo` could be justification for staying as-is. It's probably worth suggesting in the docs for building out HPC software stacks to add `build_type=Release` under the variants.

The other question is what to do about non-CMake build systems -- is it even worth worrying too much about the cmake packages when autoconf/makefiles add their own arbitrary flags? 😬

haampie:
Hehe, I just made Meson use the same as CMake: #22715. But I can see the argument in favor of Release by default. I do like having at least debug symbols in CI, but of course I could add the `build_type` variant there by hand.

jjellio:
I generally use `Release` and set `*flags=-g` - in my little world, `RelWithDebInfo` can also mean more things. (e.g., in Trilinos that means some additional 'Debug' checks can get enabled).... the dirty laundry of HPC builds (and the moral of that story, just set release + `-g` ><)

That said - I wouldn't change the default (I'd have it on par with whatever Kitware uses).

@sethrj You may want to see PR #22779 - I need to set `RelWithDebInfo` to be `-O2 -g`, but the gist is to work towards providing some kind of top-level variant that could steer all packages into a uniform direction.

ax3l:
> That said - I wouldn't change the default (I'd have it on par with whatever Kitware uses).

[...]

PR #35418, opened 2023-02-09
CMake: Change default build type to Release

fsimonis:
This PR changes the default build type of CMake packages from `RelWithDebInfo` to `Release`.

While `RelWithDebInfo` enables debug information, it drops the optimization level down to `-02`.
`Release` uses `-03`, which enables some additional optimizations.

Side effects for CMake packages will be:
* smaller installation size
* additional optimizations
* no debug information by default

Package providers relying on the default to include debug information now need to change their config to `RelWithDebInfo` or by adding `-g` as a compile flag.

Also changes the default meson build type to release for consistency.

### DIFF
--- a/lib/spack/spack/build_systems/cmake.py
+++ b/lib/spack/spack/build_systems/cmake.py
@@ -55,7 +55,7 @@ class CMakePackage(spack.package_base.PackageBase):
         # https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html
         variant(
             "build_type",
-            default="RelWithDebInfo",
+            default="Release",
             description="CMake build type",
             values=("Debug", "Release", "RelWithDebInfo", "MinSizeRel"),
         )
--- a/lib/spack/spack/build_systems/meson.py
+++ b/lib/spack/spack/build_systems/meson.py
@@ -33,7 +33,7 @@ class MesonPackage(spack.package_base.PackageBase):
     with when("build_system=meson"):
         variant(
             "buildtype",
-            default="debugoptimized",
+            default="release",
             description="Meson build type",
             values=("plain", "debug", "debugoptimized", "release", "minsize"),
         )


haampie:
- I think most autotools packages default to `-O2 -g` too.
- `meson` has `buildtype=debugoptimized`, that should then also be `release` for consistency.

It's a shame there's no `-O3 -g` default.

fsimonis:
@haampie Good point. Shall I change all of them in this PR to keep things consistent?

haampie:
All of them = meson, autotools doens't have build type presets

mwkrentel:
IMO, this is a somewhat controversial decision, either way.
There are legitimate use cases for debug symbols.

There are diverse users with diverse use cases.  And rather than
telling half of all users, "you're wrong," let me suggest that a better
solution to this is to provide a top-level knob for the default.  That
way, everyone can have it their way and there's no need to argue about
which side is "right."

haampie:
That knob is `packages.yaml` in principle.

The question is more what the default should be.

I'd personally like the non-existing `-O3 -g` as a default, so that you get speed, but also something sensible in `perf record -g`. Actually, in Meson we *could* do `-Ddebug=true -Doptimization=3`

mwkrentel:
> That knob is packages.yaml in principle.

But do I have to set this for each package?
Or, how do I set it once for all packages?

toxa81:
`RelWithDebInfo` also does not define `-DNDEBUG` wich leaves `assert()` macros active.

fsimonis:
The size difference could be a compelling argument for the default.
The debug builds of preCICE are over 100M compared to release builds which are around a few M.

Also, debug symbols alone often aren't the full story. For truly useful perf information `-fno-omit-frame-pointer` is often required too.

The real problem in my eyes is the `-02` in RelWithDebInfo, which leaves performance on the table.

Maybe this PR should really be a bug report in the CMake repo.

> RelWithDebInfo also does not define -DNDEBUG wich leaves assert() macros active.

It does for gcc and clang. You can check this with: `grep CMAKE_CXX_FLAGS_RELWITHDEBINFO CMakeCache.txt` in your binary directory (for C++).

mwkrentel:
> For truly useful perf information -fno-omit-frame-pointer is often required too.

Some debuggers or profiling tools need this. 
The one I work for (hpctoolkit) does not require this.

haampie:
There was also some discussion on the internet recently about how `-fno-omit-framepointer -g` is the better default; IIRC you _may_ get register spill, but it's really just _one_ register, which obviously is not an SIMD register, so the performance cost is 0 for typical scientific codes. The upside is you actually can get a nice call/flame graph with perf, which helps spotting bottlenecks across all packages.

If you care about debug symbol size, there's also `-gz`.

So, `-O3 -DNDEBUG -g -fno-omit-framepointer -gz` is the way ;p (edit: I mean, if we can convince build systems to do that... since in some cases (e.g. julia including llvm headers) you want targeted -DNDEBUG for dependencies only; not for the main package being compiled, which may require -UNDEBUG).

adamjstewart:
> Or, how do I set it once for all packages?

```yaml
packages:
  all:
    variants: build_type=Release
```

fsimonis:
I checked the CMake sources regarding https://github.com/spack/spack/pull/35418#issuecomment-1424470164, and found that the flags for `RelWithDebInfo` are all over the place. CMake overrides the default Release `-O2` to `-O3` for some compilers, but doesn't change `RelWithDebInfo`. `-DNDEBUG` is also not always consistent between `Release` and `RelWithDebInfo`. 

~~I'll open a bug report regarding this and propose a `Profile` build type given that it is common use-case.~~

This behaviour is known since 20 years and partially intentional
* [Inconsistent NDEBUG (fixed but later reintroduced)](https://cmake.org/Bug/view.php?id=11366)
* [Inconsistent -O2 vs -03](https://gitlab.kitware.com/cmake/cmake/-/issues/13496)
* [Optimization flags on Windows](https://gitlab.kitware.com/cmake/cmake/-/issues/20812)

The big changes regarding this seem to be tracked here
* [Add abstraction for debug info](https://gitlab.kitware.com/cmake/cmake/-/issues/23898)
* [Add abstraction for optimization level](https://gitlab.kitware.com/cmake/cmake/-/issues/23374)

I even found [a comment that mentions the default used by spack](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/591#note_1165628).

Regarding meson, [`buildtype=debugoptimized`](https://mesonbuild.com/Builtin-options.html#core-options) defaults to `-O2 -g` and it doesn't specify `NDEBUG` by default for any(!) configuration as [`b_ndebug`](https://mesonbuild.com/Builtin-options.html#base-options) is `false` by default. The mesonbuild package doesn't set `b_ndebug=if-release`, so `NDEBUG` is in the hands of project owners for meson-based packages. Is this worth opening a separate issue?

There are many inconsistencies regarding these release + debug configurations across and within build systems.

Coming back to my proposal here: at least release builds are _somewhat_ consistent. :sweat_smile:

boomanaiden154:
I am for making the default Release as while debug information can be useful for profiling in some cases, I don't think it's particularly useful to have on by default, especially on large projects (like LLVM where it can balloon the install size to 30+GB) in the face of the increased size, which is especially pertinent in some HPC environments where there might not be any disks on nodes and applications are sometimes installed to RAM disks. Having a config option to configure this would probably be needed though (as mentioned before).

However, if spack does want to do custom flags for `RelWithDebInfo` builds, setting `CMAKE_CXX_FLAGS_RELWITHDEBINFO` and `CMAKE_C_FLAGS_RELWITHDEBINFO` on the CMake command line should do the trick. This doesn't cover meson or any of the other build systems though.

tgamblin:
Implemented; superseded by #36679 and #37436.  Sorry we did not see this PR before putting those two in!

See also #22918.

This went into the 0.20 release -- see item 8 in the release notes: https://github.com/spack/spack/releases/tag/v0.20.0

PR #35425, opened 2023-02-09
Add namd2.15a2 + AOCC4.0 support

amd-toolchain-support:
Adds new NAMD version 2.15a2 - providing AVXTiles algorithm support for non-Intel compilers (AOCC, Clang, GCC)

Also adds support for building charm 6.10.x with the AOCC compiler

### DIFF
--- a/var/spack/repos/builtin/packages/charmpp/package.py
+++ b/var/spack/repos/builtin/packages/charmpp/package.py
@@ -48,6 +48,7 @@ class Charmpp(Package):
     # Patch for AOCC
     patch("charm_6.7.1_aocc.patch", when="@6.7.1 %aocc", level=1)
     patch("charm_6.8.2_aocc.patch", when="@6.8.2 %aocc", level=3)
+    patch("charm_6.10_aocc.patch", when="@6.10.0:6.10.2 %aocc")
 
     # support Fujitsu compiler
     patch("fj.patch", when="%fj")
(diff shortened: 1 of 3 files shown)


var/spack/repos/builtin/packages/charmpp/package.py
```diff
@@ -48,6 +48,7 @@ class Charmpp(Package):
     # Patch for AOCC
     patch("charm_6.7.1_aocc.patch", when="@6.7.1 %aocc", level=1)
     patch("charm_6.8.2_aocc.patch", when="@6.8.2 %aocc", level=3)
+    patch("charm_6.10_aocc.patch", when="@6.10.0:6.10.2 %aocc")
```
alalazo, line 51: Is there any submission upstream for these patches? It would be good if support for a given compiler could be contributed upstream and Spack could work with pristine sources.

amd-toolchain-support: Agreed.  At the current time we have no traction to get patches upstreamed.  Hopefully with upstream switch to CMake for Charm 7.0 onwards this becomes a non-issue.

alalazo: @amd-toolchain-support If I may, why is there no traction to get the patch upstreamed, considering it is ready and submitted to Spack? Asking generally also in reference to PRs like #35504 and #35583

amd-toolchain-support:
We will be raising separate Pull Requests for NAMD and Charmpp, and hence we are closing this Pull Request

PR #35431, opened 2023-02-10
bug fix: halide -> Halide or something.

wraith1995:
I misunderstood the meaning of class vs directory names (somehow) and things still built (somehow) despite the fact that libhalide is actually libHalide. This fixes that and I've confirmed it works on a clean spack.


wraith1995:
So that fix was not accepted, as you can see, so I implemented a temporary fix that at least works now. Are there any other packages with upper case names?

adamjstewart (approved):

PR #35436, opened 2023-02-10
Update xtb 6.6.0

awvwgk:
Released at https://github.com/grimme-lab/xtb/releases/tag/v6.6.0

### DIFF
--- a/var/spack/repos/builtin/packages/tblite/package.py
+++ b/var/spack/repos/builtin/packages/tblite/package.py
@@ -1,0 +1,50 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Tblite(MesonPackage):
+    """Light-weight tight-binding framework"""
+
+    homepage = "https://tblite.readthedocs.io"
+    url = "https://github.com/tblite/tblite/releases/download/v0.3.0/tblite-0.3.0.tar.xz"
+    git = "https://github.com/tblite/tblite.git"
+
+    maintainers = ["awvwgk"]
+
+    version("0.3.0", "46d77c120501ac55ed6a64dea8778d6593b26fb0653c591f8e8c985e35884f0a")
+
+    variant("openmp", default=True, description="Use OpenMP parallelisation")
+    variant("python", default=False, description="Build Python extension module")
+
+    depends_on("blas")
+    depends_on("dftd4@3:")
+    depends_on("lapack")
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/tblite/package.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
```
tldahlgren:
```suggestion
# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
```

var/spack/repos/builtin/packages/tblite/package.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class Tblite(MesonPackage):
+    """Light-weight tight-binding framework"""
+
+    homepage = "https://tblite.readthedocs.io"
+    url = "https://github.com/tblite/tblite/releases/download/v0.3.0/tblite-0.3.0.tar.xz"
+    git = "https://github.com/tblite/tblite.git"
+
+    maintainers = ["awvwgk"]
```
tldahlgren: The syntax for adding maintainers has changed.  See https://spack.readthedocs.io/en/latest/packaging_guide.html#maintainers

```suggestion
    maintainers("awvwgk")
```

tldahlgren (commented):
Confirmed the sha256 for both packages.  Just a couple of nits.

awvwgk (commented):
Apply suggested changes

haampie (approved):

var/spack/repos/builtin/packages/xtb/package.py
```diff
@@ -32,7 +34,9 @@ def meson_args(self):
         elif lapack != "openblas":
             lapack = "netlib"
 
+        lapack_opt = "-Dlapack={0}" if self.version > Version("6.6.0") else "-Dla_backend={0}"
```
ltalirz, line 37: Building 6.6.0 with this recipe fails with
```
>> 9    ../spack-src/meson.build:19:0: ERROR: Unknown options: "la_backend"
```


Looking at https://github.com/grimme-lab/xtb/pull/241/files it appears to me that `la_backend` is to be _introduced_ in 6.6.1.

That would mean this logic is incorrect, and it should rather be
```
lapack_opt = "-Dlapack={0}" if self.version <= Version("6.6.0") else "-Dla_backend={0}"
```

Is that correct?

awvwgk: Could you open a new issue or send a PR fixing this? Commenting in a merged PR is likely to get overlooked.

ltalirz: https://github.com/spack/spack/pull/38309

PR #35442, opened 2023-02-10
Spack develop: fix bug with delayed install of root

scheibelp:
If you have packages `X->Y`, Y is in development (i.e. with `spack develop`), and you change Y and then specifically install it, a subsequent call of `spack install` would not reinstall X. i.e. for the following commands:

```
spack install
# change source code of Y
spack install Y # this reinstalls just Y
spack install # at this point X should be reinstalled because Y was changed and reinstalled
```

`X` will not be reinstalled.

This adds a check where a dependent is now also overwritten if any in-development dependency has a newer installation date than the dependent.


scheibelp:
@spackbot fix style

lib/spack/spack/environment/environment.py
```diff
@@ -1787,6 +1787,46 @@ def _add_concrete_spec(self, spec, concrete, new=True):
...
+
+        # Perform installation-time check
+        transitive_install_times = collections.defaultdict(int)
+        for spec in traverse.traverse_nodes(
+            self.concrete_roots(), direction="children", order="post"
+        ):
+            if not spec.installed:
+                continue
+
+            _, record = spack.store.db.query_by_spec_hash(spec.dag_hash())
```
alalazo:
```suggestion
            _, record = spack.store.STORE.db.query_by_spec_hash(spec.dag_hash())
```

lib/spack/spack/environment/environment.py
```diff
@@ -1787,6 +1787,46 @@ def _add_concrete_spec(self, spec, concrete, new=True):
...
+            for parent in spec.dependents():
+                transitive_install_times[parent] = max(
+                    transitive_install_times[parent], when_this_spec_was_installed
+                )
+        specs_where_a_dependency_was_installed_more_recently = list()
+        for spec, transitive_install_time in transitive_install_times.items():
+            if not spec.installed:
+                continue
+
+            _, record = spack.store.db.query_by_spec_hash(spec.dag_hash())
```
alalazo:
```suggestion
            _, record = spack.store.STORE.db.query_by_spec_hash(spec.dag_hash())
```

alalazo (dismissed):
Requiring two changes due to #38944 being merged

haampie:
Can you implement it along the lines of

```python
def _get_overwrite_specs(
    concrete_specs: List[spack.spec.Spec], database: Optional[spack.database.Database] = None
) -> List[str]:
    """If a develop spec has been modified, we need to overwrite it and its parents.
    Also ensure that the parents of installed develop specs are overwritten if they
    are older than the installation time of the develop spec.

    Args:
        database: The database to use. If None, the default database is used.

    Returns:
        List of dag hashes of specs that need to be overwritten."""
    db = database or spack.store.STORE.db

    # Get all specs in topological order.
    specs: Iterable[spack.spec.Spec] = reversed(
        list(traverse.traverse_nodes(concrete_specs, order="topo", key=traverse.by_dag_hash))
    )

    # max_time[spec.dag_hash()] is the maximum installation time dev deps; or inf
    # if any of its deps needs to be overwritten.
    max_time: Dict[str, float] = {}
    hashes_to_overwrite = []

    for spec in specs:
        key = spec.dag_hash()
        upstream, record = db.query_by_spec_hash(key)
        dep_time = max_time.pop(key, 0.0)

        if not upstream and record:
            # Case 1: spec is installed locally, and a newer dev dep is in the subdag.
            if record.installation_time < dep_time:
                hashes_to_overwrite.append(key)
                dep_time = math.inf
            else:
                path = spec.variants.get("dev_path", None)
                if path:
                    mtime = fs.last_modification_time_recursive(path.value)
                    # Case 2: this spec is a develop spec, and it has been modified.
                    if mtime > record.installation_time:
                        # Have to reinstall this develop spec.
                        hashes_to_overwrite.append(key)
                        dep_time = math.inf
                    else:
                        # Case 3: the develop spec is already installed, dependents
                        # need to be overwritten if they are older than the installation
                        dep_time = max(dep_time, record.installation_time)

        # Propagate max dep install time to dependents
        for dependent in spec.dependents():
            dep_key = dependent.dag_hash()
            if max_time.get(dep_key, 0.0) < dep_time:
                max_time[dep_key] = dep_time

    return hashes_to_overwrite
```

I think it's more clear if you topo sort and then propagate (a) install times of up-to-date develop dependencies or (b) infinity for out of date develop deps. Maybe split it into two functions to avoid the deeply nested control flow.

That way we also don't check O(develop spec) modification times if say one common leaf node has changed, and checking one would be enough.

Also, with the above structure ^ could you refactor the test so that it takes a few us/ms instead of seconds? Construct a dev spec, mark it concrete, create a db, add it with particular install time. That way you don't need to concretize, hit the filesystem as much, or use `sleep` :laughing: 

Btw, I'm also thinking that we should maybe reverse the default order of topo sort, cause if you look at the implementation it already does a reverse, and the call site commonly re-reverses it, like ^

scheibelp:
> Also, with the above structure ^ could you refactor the test so that it takes a few us/ms instead of seconds? Construct a dev spec, mark it concrete, create a db, add it with particular install time.

I agree with this and added a new helper function in `Database` called `_update_install_time`.

> Can you implement it along the lines of

I'm having trouble seeing the value of this:

> That way we also don't check O(develop spec) modification times if say one common leaf node has changed, and checking one would be enough.

my implementation is `O(|N| + |E|)`, and ultimately there will be no algorithm which is asymptotically superior.

> I think it's more clear if you topo sort and then propagate

`traverse(order="post")` means "visit the parents after their children; `reverse(topo_sort)` does that too, but I don't see how it could be clearer, when the effect of "post" is explicitly documented for the `traverse()` function. One could complain about my logic depending on the default value of the dict being `0`, and I could certainly clarify that, but it seems semantically equivalent to propagating `math.inf`. Overall I think it's worth asking: is the approach, as implemented currently, *unclear* or *deficient*? If not, then IMO it's worth merging.

haampie:
If develop spec A depends on develop spec B, and B has changed, there's no point in checking that A has changed, you know it should be rebuilt.

Since "some dependency gets rebuilt" bubbles up exactly like your max develop spec install time, my suggestion is to combine these two. In the code above "some dependency gets rebuilt" corresponds to math.inf install time (I wish inf wasn't so ugly in python).

lib/spack/spack/database.py
```diff
@@ -1238,6 +1238,19 @@ def add(self, spec, directory_layout, explicit=False):
         with self.write_transaction():
             self._add(spec, directory_layout, explicit=explicit)
 
+    @_autospec
+    def _update_install_time(self, spec, absolute_time=None, delta=None):
```
haampie: Extending the interface of an object for something only used in tests is bad practice. I was hoping you could do something like

```python
def test_dev_build_rebuild_dependent_delayed(tmpdir):
    # Create a dag
    a, b = spack.spec.Spec("a"), spack.spec.Spec("b")
    b.variants["dev_path"] = vt.SingleValuedVariant("dev_path", str(tmpdir.join("source")))
    a.add_dependency_edge(b, deptypes=("link", "run"), virtuals=())
    a._mark_concrete()

    # Create a source dir
    tmpdir.ensure("source", dir=True)
    last_modification = fs.last_modification_time_recursive(tmpdir.join("source"))

    # Create a database in which the dev spec was installed *later* than its dependent
    db = spack.database.Database(tmpdir.join('db'), lock_cfg=spack.database.NO_LOCK)
    db.add(b, directory_layout=None, installation_time=last_modification + 1)
    db.add(a, directory_layout=None, installation_time=last_modification - 1)

    assert get_overwrite_specs([a, b], database=db) == [a.dag_hash()]
```

which is more of a unit test and actually runs in ms instead of s.

scheibelp: The functionality added in this PR accesses at least two interfaces:

* The specs (to see if they are installed)
* The database

if you pass in a test instance DB, that must be used to mock the global DB, so that specs grab from that properly. In this case, I happen to know from being familiar with all the underlying logic that assigning this DB object you create to be the store would be sufficient (at least for now), but it is subject to conceptual drift (if the notion of how the spec is installed changes, or it reaches into any other part of the global state, then there are possibilities for more-insidious bugs).

So I like the general idea of producing a more-isolated test for this, but I don't think that can be done cleanly without mocking both the Spec object as well as the Database: does that seem reasonable? Another option along those lines is to promote the concept of install-time to the spec level (at which point the only object that needs to be mocked is the Spec), although that would conflict with your no-functions-in-objects-just-for-tests preference <sup>1</sup>.

Also regarding your comment about the test taking a long time, I think that could be improved in general without specifically changing this test. Originally I was going to blame the fixtures, but actually the longest running parts appear to be the calls to `install()`, I think this could be improved by mocking the filesystem. Overall that isn't necessarily something to address here (although it may still be worthwhile having this test in addition to the more-specific one), but the integration-level tests certainly have a place somewhere in our testing modules, and speeding them up would be worthwhile.

1. That being said, while I understand the rule of thumb in play there, I don't think it applies in this case: for example I see no difference between using a private function like `_mark_concrete` and using `_update_install_time`; it's true that `_mark_concrete` has an already-existing meaningful intent, but you can use it here because you understand it also has a meaningful public-API-like interpretation, and for that matter I can say the same thing about `_update_install_time`).

haampie:
I still would say it's better to combine the two passes.

The rules are:

1. A node gets overwritten if one of its children gets overwritten
2. A node gets overwritten if one of its children is a more recently installed develop spec
3. A node gets overwritten if it is a develop spec and its mtime > install time

Those rules are order from cheap to expensive to check. Right now we call (3) for every develop spec, whereas relying on rule (1) and (2) can sometimes eliminate checking (3) altogether.

Propagating 1 and 2 is pretty much the same code, so that looks like a clear advantage to me.

lib/spack/spack/database.py
haampie: IMO if the interface of the function you're unit testing takes a database instance as argument, the meaning of "installed" is clear enough: it should be in that database.

I don't think there's a need to be worried about whether the function you're unit testing uses `spec.installed`, if it accepts a database as an argument. If it would, then it's a bug (and tests will fail).

And I also don't think we're going to redefine what it means to be installed. I think `spec.installed` is just a convenience function, not a definition of installed-ness. Also note that this wouldn't be the first place where db.query(...) is used instead of spec.installed.

In a perfect world we'd invert the spec/db dependency for install checks too: `db.installed(spec)` instead of `spec.installed(db)` (well, it's not even an argument currently...). But I guess we're not quite there yet. As you can imagine, implicit globals are not great, and it's not just in tests where we have to replace globals for mocking, we also have to do so in the bootstrapping context... (Having to replace a bunch of globals in different modules just to be able to get the correct result out of `spec.installed` in a bootstrap context is not great)

> I see no difference between using a private function like `_mark_concrete` and using `_update_install_time`

For databases it's convenient to have something to set an install time in general (e.g. when reindexing a buildcache, you might want to use the tarball creation time instead of the current time). Sounds like that could be public interface. In fact `_add(...)` already has that signature, but `add(...)` does not. Also, it's just another example of dependency injection... remove the hard-coded dependency of `def add(...)` on `time`, allow it to be injected instead. 

Regarding `_mark_concrete` I guess it's kinda like a poor man's builder pattern, which in the future can be improved to turn an assembled AbstractSpec into an immutable ConcreteSpec (@alalazo wanted something like that). Sure `_mark_concrete` is private, but I don't think it's necessarily wrong to use it in tests.

The difference is whether you're adding new methods *just* for tests or not.

--

In short:

1) db.installed(spec) >> spec.installed
2) avoid extending an interface just for tests

scheibelp: The test now refactors environment to use an (optionally) injected Database (and `Database` now has an `installed` method).

scheibelp:
> I still would say it's better to combine the two passes.

I have now done so: the method will now avoid checking timestamps on dependents if a dependency has already met the criteria for overwriting.

The test is updated now as well (and moved, since it isn't exercising the `dev-build` command).

scheibelp:
@haampie can you review again?

lib/spack/spack/environment/environment.py
```diff
@@ -422,7 +425,7 @@ def _is_dev_spec_and_has_changed(spec):
         # Not installed -> nothing to compare against
         return False
```
haampie: Above you're still using `spec.installed` from the global db, whereas below `record` is from the correct provided db.

lib/spack/spack/environment/environment.py
```diff
@@ -1794,31 +1797,73 @@ def _add_concrete_spec(self, spec, concrete, new=True):
...
+            # non-develop spec, in which case it should not be necessary to
+            # reinstall the parent)
+            transitive_dev_install_time_child = transitive_dev_install_times[spec]
+            for parent in spec.dependents():
+                transitive_dev_install_times[parent] = max(
+                    transitive_dev_install_times[parent], transitive_dev_install_time_child
+                )
+
+        for spec in traverse.traverse_nodes(
+            self.concrete_roots(), direction="children", order="post"
```
haampie, line 1837:
```suggestion
            self.concrete_roots(), order="post"
```

lib/spack/spack/environment/environment.py
```diff
@@ -1794,31 +1797,73 @@ def _add_concrete_spec(self, spec, concrete, new=True):
...
+        overwrite_specs = set()
+        transitive_dev_install_times = collections.defaultdict(int)
+
+        def update_transitive_dev_install_times(spec):
+            dev_path_var = spec.variants.get("dev_path", None)
+            if dev_path_var and db.installed(spec):
+                _, record = db.query_by_spec_hash(spec.dag_hash())
+                when_this_spec_was_installed = record.installation_time
+
+                transitive_dev_install_times[spec] = max(
```
haampie: Why don't you assign `transitive_dev_install_time_child = transitive_dev_install_times[spec]` first and then conditionally update it, instead of calling `transitive_dev_install_times[spec]` 3 times?

lib/spack/spack/environment/environment.py
```diff
@@ -1794,31 +1797,73 @@ def _add_concrete_spec(self, spec, concrete, new=True):
...
+            if any(child in overwrite_specs for child in spec.dependencies()):
+                overwrite_specs.add(spec)
+                continue
+
+            update_transitive_dev_install_times(spec)
+
+            # Say x->y->z, you force uninstall y, then you change the source
+            # of z (which is develop). x should be reinstalled, but not because
+            # of y. Therefore, we wait until now to bail on uninstalled specs.
+            if not db.installed(spec):
```
haampie: You're using `db.installed` followed by `db.query_by_spec_hash` on three occasions. Just use `_, record = db.query_by_spec_hash(...)` then and replace `if not db.installed(...)` with `if record is None or not record.installed`

lib/spack/spack/environment/environment.py
```diff
@@ -1794,31 +1797,73 @@ def _add_concrete_spec(self, spec, concrete, new=True):
...
+                continue
+
+            # If a transitive child develop spec has a greater install
+            # time, then overwrite
+            _, record = db.query_by_spec_hash(spec.dag_hash())
+            when_this_spec_was_installed = record.installation_time
+            if when_this_spec_was_installed < transitive_dev_install_times[spec]:
```
haampie: Instead of looking up this value yet again, why not return it from `update_transitive_dev_install_times`.

In fact you can also delay looping over `dependents()` until after this branch, since it's redundant if this branch is followed.

lib/spack/spack/environment/environment.py
```diff
@@ -1794,31 +1797,73 @@ def _add_concrete_spec(self, spec, concrete, new=True):
...
-                root=True,
-                order="breadth",
-                depth=True,
-                direction="parents",
-                key=traverse.by_dag_hash,
-            )
-            if depth == 0 or spec.installed
-        ]
+        overwrite_specs = set()
+        transitive_dev_install_times = collections.defaultdict(int)
```
haampie:
```suggestion
        transitive_dev_install_times = collections.defaultdict(float)
```

haampie:
Some small requests, and then let's get this pr in.

FWIW, I'm generally not a fan of using `Spec` objects as keys in dicts and sets. I guess it's OK here since all specs are concrete, but still better to reduce the number of times we call `Spec.__hash__` and `Spec.__eq__`.

scheibelp:
I have fixed the bug mentioned in https://github.com/spack/spack/pull/35442#discussion_r1299661695

> Why don't you assign transitive_dev_install_time_child = transitive_dev_install_times[spec] first and then conditionally update it, instead of calling transitive_dev_install_times[spec] 3 times?

> Instead of looking up this value yet again, why not return it from update_transitive_dev_install_times.

> In fact you can also delay looping over dependents() until after this branch, since it's redundant if this branch is followed.

I have tried to refactor the code with these requests in mind as well, but as for why I did it the way I did it:

* I consider it "free" to do O(1) operations that are O(1) any time I write a function; if intermediate values become temporarily irrelevant, I prefer to discard and recompute them, since carrying them around complicates functions. 
* I also dislike inferring state from underlying variables. For example when we used `InstallRecord.installed` vs. `Database.installed`  we are subverting the API of `Database`: we know asking `if record and record.installed` is the same as `Database.installed`, but not every person who comes along and reads this code will immediately understand that.

> FWIW, I'm generally not a fan of using Spec objects as keys in dicts and sets. I guess it's OK here since all specs are concrete, but still better to reduce the number of times we call Spec.__hash__ and Spec.__eq__.

* `Spec.__hash__` is cached for concrete specs, if you call it once, all future calls are nearly instantaneous. 
* I can see a case made for `__eq__`, although I am personally unconvinced without profiling. If it is truly a major time sink in Spack, could we make it fast by maintaining a global cache like:

```
class SpecEqCache:
  def __init__:
    self.cache = defaultdict(list)

  # every concrete spec that calls _cmp_iter can call this and set a property
  def eq_id(self, spec)
    hash_list = self.cache(spec.dag_hash)
    x = hash_list.index_of(spec)
    if not x:
      hash_list.append(spec)
      x = len(hash_list) - 1
    return x
```

?

PR #29698, opened 2022-03-24
libxpm: Add setup_dependent_build_environment and fix when gettext dose not provide libintl

tkameyama:
- Add setup_dependent_build_environment
- When gettext is external package and  the gettext configured with `--without-included-gettext` option, gettext dose not provide libintl. So if gettext dose not provide libintl, LDFLAGS environment variable is not set.

PR #35450, opened 2023-02-10
Make recipes consistent w.r. libintl, iconv, gettext

greenc-FNAL:
Following
[discussion](https://github.com/spack/spack/wiki/Telcon:-2023-02-08) re
#34541 and friends, I noted that I had a fork with comprehensive changes
purporting to solve the whole mess with `iconv`/`libiconv`/`libc`,
`libintl`/`libc`, and `gettext`, and all the dependents thereof.

This set of changes not only gives consistent build stack behavior with
respect to the above, but also works with externals.

### DIFF
--- a/var/spack/repos/builtin/packages/acl/package.py
+++ b/var/spack/repos/builtin/packages/acl/package.py
@@ -26,8 +26,10 @@ class Acl(AutotoolsPackage):
     depends_on("attr")
     depends_on("gettext")
 
-    def setup_build_environment(self, env):
-        env.append_flags("LDFLAGS", "-lintl")
+    def flag_handler(self, name, flags):
+        if name == "ldlibs" and "intl" in self.spec["gettext"].libs.names:
+            flags.append("-lintl")
+        return self.build_system_flags(name, flags)
 
     def autoreconf(self, spec, prefix):
         bash = which("bash")
--- a/var/spack/repos/builtin/packages/binutils/package.py
+++ b/var/spack/repos/builtin/packages/binutils/package.py
@@ -269,9 +269,12 @@ class AutotoolsBuilder(spack.build_systems.autotools.AutotoolsBuilder):
         # also grab the headers from the bfd directory
         install(join_path(self.build_directory, "bfd", "*.h"), extradir)
 
-    def setup_build_environment(self, env):
-        if self.spec.satisfies("%cce"):
-            env.append_flags("LDFLAGS", "-Wl,-z,muldefs")
-
-        if "+nls" in self.spec:
-            env.append_flags("LDFLAGS", "-lintl")
+    def flag_handler(self, name, flags):
+        spec = self.spec
+        if name == "ldflags":
+            if spec.satisfies("%cce"):
+                flags.append("-Wl,-z,muldefs")
+        elif name == "ldlibs":
+            if "+nls" in self.spec and "intl" in self.spec["gettext"].libs.names:
+                flags.append("-lintl")
+        return self.build_system_flags(name, flags)
--- a/var/spack/repos/builtin/packages/elfutils/package.py
+++ b/var/spack/repos/builtin/packages/elfutils/package.py
@@ -7,6 +7,7 @@ import glob
 import os.path
 
 from spack.package import *
+from spack.util.environment import is_system_path
 
 
 class Elfutils(AutotoolsPackage, SourcewarePackage):
@@ -99,7 +100,12 @@ class Elfutils(AutotoolsPackage, SourcewarePackage):
         files = glob.glob(os.path.join("*", "Makefile.in"))
         filter_file("-Werror", "", *files)
 
-    flag_handler = AutotoolsPackage.build_system_flags
+    def flag_handler(self, name, flags):
+        if name == "ldlibs":
+            spec = self.spec
+            if "+nls" in spec and "intl" in spec["gettext"].libs.names:
+                flags.append("-lintl")
+        return self.inject_flags(name, flags)
 
     def configure_args(self):
         spec = self.spec
@@ -120,9 +126,19 @@ class Elfutils(AutotoolsPackage, SourcewarePackage):
         # zlib is required
         args.append("--with-zlib=%s" % spec["zlib"].prefix)
 
+        if spec.satisfies("@0.183:"):
+            if spec["iconv"].name == "libc":
+                args.append("--without-libiconv-prefix")
+            elif not is_system_path(spec["iconv"].prefix):
+                args.append("--with-libiconv-prefix=" + format(spec["iconv"].prefix))
+
         if "+nls" in spec:
-            # configure doesn't use LIBS correctly
-            args.append("LDFLAGS=-Wl,--no-as-needed -L%s -lintl" % spec["gettext"].prefix.lib)
+            # Prior to 0.183, only msgfmt is used from gettext.
+            if spec.satisfies("@0.183:"):
+                if "intl" not in spec["gettext"].libs.names:
+                    args.append("--without-libintl-prefix")
+                elif not is_system_path(spec["gettext"].prefix):
+                    args.append("--with-libintl-prefix=" + spec["gettext"].prefix)
         else:
             args.append("--disable-nls")
 
--- a/var/spack/repos/builtin/packages/extrae/package.py
+++ b/var/spack/repos/builtin/packages/extrae/package.py
@@ -127,14 +127,20 @@ class Extrae(AutotoolsPackage):
             make.add_default_arg("CXXFLAGS=%s" % self.compiler.cxx11_flag)
             args.append("CXXFLAGS=%s" % self.compiler.cxx11_flag)
 
+        return args
+
+    def flag_handler(self, name, flags):
         # This was added due to:
         # - configure failure
         # https://www.gnu.org/software/gettext/FAQ.html#integrating_undefined
         # - linking error
         # https://github.com/bsc-performance-tools/extrae/issues/57
-        args.append("LDFLAGS=-lintl -pthread")
-
-        return args
+        if name == "ldlibs":
+            if "intl" in self.spec["gettext"].libs.names:
+                flags.append("-lintl")
+        elif name == "ldflags":
+            flags.append("-pthread")
+        return self.build_system_flags(name, flags)
 
     def install(self, spec, prefix):
         with working_dir(self.build_directory):
--- a/var/spack/repos/builtin/packages/gdal/package.py
+++ b/var/spack/repos/builtin/packages/gdal/package.py
@@ -9,7 +9,7 @@ import sys
 from spack.build_systems.autotools import AutotoolsBuilder
 from spack.build_systems.cmake import CMakeBuilder
 from spack.package import *
-from spack.util.environment import filter_system_paths
+from spack.util.environment import filter_system_paths, is_system_path
 
 
 class Gdal(CMakePackage, AutotoolsPackage, PythonExtension):
@@ -626,7 +626,6 @@ class AutotoolsBuilder(AutotoolsBuilder):
             self.with_or_without("hdf4", package="hdf"),
             self.with_or_without("hdf5", package="hdf5"),
             self.with_or_without("hdfs", package="hadoop"),
-            self.with_or_without("libiconv-prefix", variant="iconv", package="iconv"),
             self.with_or_without("idb", package="idb"),
             self.with_or_without("ingres", package="ingres"),
             self.with_or_without("jasper", package="jasper"),
@@ -680,6 +679,11 @@ class AutotoolsBuilder(AutotoolsBuilder):
             self.with_or_without("perl"),
             self.with_or_without("php"),
         ]
+        if "+iconv" in self.spec:
+            if self.spec["iconv"].name == "libc":
+                args.append("--without-libiconv-prefix")
+            elif not is_system_path(self.spec["iconv"].prefix):
+                args.append("--with-libiconv-prefix=" + self.spec["iconv"].prefix)
 
         # Renamed or modified flags
         if self.spec.satisfies("@3:"):
--- a/var/spack/repos/builtin/packages/gettext/package.py
+++ b/var/spack/repos/builtin/packages/gettext/package.py
@@ -6,6 +6,7 @@
 import re
 
 from spack.package import *
+from spack.util.environment import is_system_path
 
 
 class Gettext(AutotoolsPackage, GNUMirrorPackage):
@@ -78,7 +79,6 @@ class Gettext(AutotoolsPackage, GNUMirrorPackage):
         config_args = [
             "--disable-java",
             "--disable-csharp",
-            "--with-libiconv-prefix={0}".format(spec["iconv"].prefix),
             "--with-included-glib",
             "--with-included-gettext",
             "--with-included-libcroco",
@@ -86,6 +86,11 @@ class Gettext(AutotoolsPackage, GNUMirrorPackage):
             "--with-lispdir=%s/emacs/site-lisp/gettext" % self.prefix.share,
             "--without-cvs",
         ]
+
+        if self.spec["iconv"].name == "libc":
+            config_args.append("--without-libiconv-prefix")
+        elif not is_system_path(self.spec["iconv"].prefix):
+            config_args.append("--with-libiconv-prefix=" + self.spec["iconv"].prefix)
 
         if "+curses" in spec:
             config_args.append("--with-ncurses-prefix={0}".format(spec["ncurses"].prefix))
@@ -111,6 +116,10 @@ class Gettext(AutotoolsPackage, GNUMirrorPackage):
             config_args.append("--with-included-libunistring")
 
         return config_args
+
+    @property
+    def gettext_libs(self):
+        return self.libs
 
     @property
     def libs(self):
--- a/var/spack/repos/builtin/packages/gettext/package.py
+++ b/var/spack/repos/builtin/packages/gettext/package.py
@@ -118,10 +118,6 @@ class Gettext(AutotoolsPackage, GNUMirrorPackage):
         return config_args
 
     @property
-    def gettext_libs(self):
-        return self.libs
-
-    @property
     def libs(self):
         return find_libraries(
             ["libasprintf", "libgettextlib", "libgettextpo", "libgettextsrc", "libintl"],
--- a/var/spack/repos/builtin/packages/git/package.py
+++ b/var/spack/repos/builtin/packages/git/package.py
@@ -7,6 +7,7 @@ import os
 import re
 
 from spack.package import *
+from spack.util.environment import is_system_path
 
 
 class Git(AutotoolsPackage):
@@ -466,12 +467,18 @@ class Git(AutotoolsPackage):
         # The test avoids failures when git is an external package.
         # In that case the node in the DAG gets truncated and git DOES NOT
         # have a gettext dependency.
-        if "+nls" in self.spec:
-            if "intl" in self.spec["gettext"].libs.names:
-                env.append_flags("EXTLIBS", "-L{0} -lintl".format(self.spec["gettext"].prefix.lib))
-            env.append_flags("CFLAGS", "-I{0}".format(self.spec["gettext"].prefix.include))
+        spec = self.spec
+        if "+nls" in spec:
+            if "intl" in spec["gettext"].libs.names:
+                extlib_bits = []
+                if not is_system_path(spec["gettext"].prefix):
+                    extlib_bits.append(spec["gettext"].libs.search_flags)
+                extlib_bits.append("-lintl")
+                env.append_flags("EXTLIBS", " ".join(extlib_bits))
+            if not is_system_path(spec["gettext"].prefix):
+                env.append_flags("CFLAGS", spec["gettext"].headers.include_flags)
 
-        if "~perl" in self.spec:
+        if "~perl" in spec:
             env.append_flags("NO_PERL", "1")
 
     def configure_args(self):
@@ -480,10 +487,16 @@ class Git(AutotoolsPackage):
         configure_args = [
             "--with-curl={0}".format(spec["curl"].prefix),
             "--with-expat={0}".format(spec["expat"].prefix),
-            "--with-iconv={0}".format(spec["iconv"].prefix),
             "--with-openssl={0}".format(spec["openssl"].prefix),
             "--with-zlib={0}".format(spec["zlib"].prefix),
         ]
+
+        if not self.spec["iconv"].name == "libc":
+            configure_args.append(
+                "--with-iconv={0}".format(
+                    "yes" if is_system_path(spec["iconv"].prefix) else spec["iconv"].prefix
+                )
+            )
 
         if "+perl" in self.spec:
             configure_args.append("--with-perl={0}".format(spec["perl"].command.path))
--- a/var/spack/repos/builtin/packages/glib/package.py
+++ b/var/spack/repos/builtin/packages/glib/package.py
@@ -6,6 +6,7 @@
 import os.path
 
 from spack.package import *
+from spack.util.environment import is_system_path
 
 
 class Glib(Package):
@@ -210,7 +211,7 @@ class Glib(Package):
         if self.spec.satisfies("@:2.72"):
             args.append("-Dgettext=external")
         if self.spec.satisfies("@:2.74"):
-            if "libc" in self.spec:
+            if self.spec["iconv"].name == "libc":
                 args.append("-Diconv=libc")
             else:
                 if self.spec.satisfies("@2.61.0:"):
@@ -244,7 +245,7 @@ class Glib(Package):
             args.append(
                 "--with-python={0}".format(os.path.basename(self.spec["python"].command.path))
             )
-        if "libc" in self.spec:
+        if self.spec["iconv"].name == "libc":
             args.append("--with-libiconv=maybe")
         else:
             args.append("--with-libiconv=gnu")
@@ -350,10 +351,14 @@ class Glib(Package):
         # Packages that link to glib were also picking up -lintl from glib's
         # glib-2.0.pc file. However, packages such as py-pygobject were
         # bypassing spack's compiler wrapper for linking and thus not finding
-        # the gettext library directory. The patch below explitly adds the
+        # the gettext library directory. The patch below explicitly adds the
         # appropriate -L path.
         spec = self.spec
-        if spec.satisfies("@2.0:2"):
+        if (
+            spec.satisfies("@2.0:2")
+            and "intl" in self.spec["gettext"].libs.names
+            and not is_system_path(spec["gettext"].prefix)
+        ):
             pattern = "Libs:"
             repl = "Libs: -L{0} -Wl,-rpath={0} ".format(spec["gettext"].libs.directories[0])
             myfile = join_path(self.spec["glib"].libs.directories[0], "pkgconfig", "glib-2.0.pc")
--- a/var/spack/repos/builtin/packages/mono/package.py
+++ b/var/spack/repos/builtin/packages/mono/package.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
+from spack.util.environment import is_system_path
 
 
 class Mono(AutotoolsPackage):
@@ -74,6 +75,8 @@ class Mono(AutotoolsPackage):
 
     def configure_args(self):
         args = []
-        li = self.spec["iconv"].prefix
-        args.append("--with-libiconv-prefix={p}".format(p=li))
+        if self.spec["iconv"].name == "libc":
+            args.append("--without-libiconv-prefix")
+        elif not is_system_path(self.spec["iconv"].prefix):
+            args.append("--with-libiconv-prefix={p}".format(p=self.spec["iconv"].prefix))
         return args
--- a/var/spack/repos/builtin/packages/procps/package.py
+++ b/var/spack/repos/builtin/packages/procps/package.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
+from spack.util.environment import is_system_path
 
 
 class Procps(AutotoolsPackage):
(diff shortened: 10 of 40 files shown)


greenc-FNAL:
Closes #34541

greenc-FNAL:
@scheibelp @mwkrentel As promised.

var/spack/repos/builtin/packages/acl/package.py
```diff
@@ -27,7 +27,8 @@ class Acl(AutotoolsPackage):
     depends_on("gettext")
 
     def setup_build_environment(self, env):
-        env.append_flags("LDFLAGS", "-lintl")
+        if "intl" in self.spec["gettext"].libs.names:
+            env.append_flags("LDFLAGS", "-lintl")
```
adamjstewart: Why not use:
```suggestion
        env.append_flags("LDFLAGS", self.spec["gettext"].libs.link_flags)
```

var/spack/repos/builtin/packages/gdal/package.py
```diff
@@ -629,7 +629,6 @@ def configure_args(self):
             self.with_or_without("hdf4", package="hdf"),
             self.with_or_without("hdf5", package="hdf5"),
             self.with_or_without("hdfs", package="hadoop"),
-            self.with_or_without("libiconv-prefix", variant="iconv", package="iconv"),
```
adamjstewart, line 629: What's the build error you see without this?

greenc-FNAL:
Note: changes originally made in collaboration with @marcmengel.

var/spack/repos/builtin/packages/acl/package.py
greenc-FNAL: If `libs.link_flags` gets filled in properly for `gettext` as an external, that would be a reasonable thing to do.

var/spack/repos/builtin/packages/gdal/package.py
greenc-FNAL: If `iconv` is provided by `libc` and is an external, then this adds `/usr` as a high-priority prefix.

var/spack/repos/builtin/packages/acl/package.py
adamjstewart: It will add links for:
```python
["libasprintf", "libgettextlib", "libgettextpo", "libgettextsrc", "libintl"]
```

var/spack/repos/builtin/packages/gdal/package.py
adamjstewart: That sounds like it isn't a bug with libc but a bug with external packages. Shouldn't you instead check where iconv is installed and filter out system paths? `spack.util.environment` has `is_system_path` and `filter_system_paths` functions to handle this kind of stuff.

mwkrentel:
@greenc-FNAL Hmm, this is not exactly the patch I had in mind.

First, I was expecting the logic about `if "intl" in libs.name` to be
in gettext, not every package that uses it.

Also, in the gettext patch, I'm not so sure about the separation for
redhat, centos and scientific distros.  I checked one RH 8.x and one
Ubuntu 18.04 system and they both put the gettext functions in libc.
Neither had a separate `libintl`.

The main difference was RH put libc in `/usr/lib64` and Ubuntu put it
in `/lib/x86_64-linux-gnu`.

So, the dividing line between `-lintl` and not seems to be build
gettext from source vs find it in `/usr`, not whether it's RH/Centos
vs other distro.

Finally, there's a separate problem over the names of the functions as
seen in one of the failed tests.

https://gitlab.spack.io/spack/spack/-/jobs/5538241

It looks like some packages want the `libintl_gettext()` names and
some are more tolerant about that or `gettext()`.

Spack doesn't really have a good way to express this.  Maybe we need
to add a variant that says whether the `libintl_` names exist.  Make
that true for building gettext and false for finding it in /usr.

I'm not sure, I haven't done a careful inventory of what other
packages need.

greenc-FNAL:
> @greenc-FNAL Hmm, this is not exactly the patch I had in mind.
> 
> First, I was expecting the logic about `if "intl" in libs.name` to be in gettext, not every package that uses it.

@mwkrental Understood. This is what I had and have been using, so it's what I turned into the initial PR.

> Also, in the gettext patch, I'm not so sure about the separation for redhat, centos and scientific distros. I checked one RH 8.x and one Ubuntu 18.04 system and they both put the gettext functions in libc. Neither had a separate `libintl`.

I can look into that.

> The main difference was RH put libc in `/usr/lib64` and Ubuntu put it in `/lib/x86_64-linux-gnu`.
> So, the dividing line between `-lintl` and not seems to be build gettext from source vs find it in `/usr`, not whether it's RH/Centos vs other distro.

I think its more, "how was it built." I'm pretty sure I can build gettext from source on e.g. Scientific7 and if I use the right options and virtual provider settings it will be built without `libintl.so`

> Finally, there's a separate problem over the names of the functions as seen in one of the failed tests.
> 
> https://gitlab.spack.io/spack/spack/-/jobs/5538241
> 
> It looks like some packages want the `libintl_gettext()` names and some are more tolerant about that or `gettext()`.

I will investigate.

> Spack doesn't really have a good way to express this.

Stipulated.

> Maybe we need to add a variant that says whether the `libintl_` names exist. Make that true for building gettext and false for finding it in /usr.
> I'm not sure, I haven't done a careful inventory of what other packages need.

I've thought about the variant idea previously, but see above re gettext build options.

var/spack/repos/builtin/packages/gdal/package.py
greenc-FNAL: I need to double-check, but I'm moderately sure that using `--libiconv-prefix` tries to add `-liconv` to the library list  which will fail if we're getting it from `libc`. The fouling of the library search path is secondary to that.

With reference to, "Shouldn't you instead check where iconv is installed and filter out system paths?" however: if "you" is, the package maintainer, the unfortunate answer to that is, "no, because nobody else does." As a release manager who configures a certain set of externals all the time, I'm forever running into packages where some aspect of the recipe: library search paths, `PATH`, or other environment settings gets tripped up by using externals. It's left to individual Spack users like me to trip over the problem, fix it in every package that has the problem and submit the PR. If there were a way to have the CI test recipe builds with dependencies configured as externals in addition to having them installed in the same Spack setup, that might help. Until then however, the burden falls on the first unfortunate soul to want to build against _that_ dependency as an external and gets to watch everything explode.

adamjstewart: By "you" I mean the person writing or maintaining the `package.py` build recipes. From your initial reply, it sounded like instead of checking whether iconv is libc, we should instead check to see if iconv is installed in a system directory where many other things are installed.

We automatically filter system deps most of the time (anything automatically added to the compiler wrappers, for example). It's hard to automatically filter them the rest of the time though. The best we can do automatically would be to ensure that all link/include flags are sorted with system locations last. Can't remember if @scheibelp or @becker33 or @haampie ever managed to get that working, or who was working on that. I agree that it's annoying to have to do this on a package-by-package basis.

greenc-FNAL: > By "you" I mean the person writing or maintaining the `package.py` build recipes. From your initial reply, it sounded like instead of checking whether iconv is libc, we should instead check to see if iconv is installed in a system directory where many other things are installed.

No, sorry.

> We automatically filter system deps most of the time (anything automatically added to the compiler wrappers, for example). It's hard to automatically filter them the rest of the time though. The best we can do automatically would be to ensure that all link/include flags are sorted with system locations last. Can't remember if @scheibelp or @becker33 or @haampie ever managed to get that working, or who was working on that. I agree that it's annoying to have to do this on a package-by-package basis.

We already have environment utilities provided by me to do this:https://github.com/spack/spack/blob/9b3c4e0696a9f90ac0870c32b0ef9dad1b529583/lib/spack/spack/util/environment.py#L76-L80 and https://github.com/spack/spack/blob/9b3c4e0696a9f90ac0870c32b0ef9dad1b529583/lib/spack/spack/util/environment.py#L330-L338 but that has to be used on a package-by-package basis. I'm not sure how easy that would be to integrate into the wrapper's operations.

haampie: It's already used by default.

greenc-FNAL:
@spackbot style

lib/spack/spack/util/environment.py
```diff
@@ -62,11 +62,16 @@
...
+search_prefix = "-L"
 
 
 def is_system_path(path: Path) -> bool:
     """Returns True if the argument is a system path, False otherwise."""
-    return bool(path) and (os.path.normpath(path) in SYSTEM_DIRS)
+    return (
+        bool(path)
+        and os.path.normpath(path[path.startswith(search_prefix) and len(search_prefix) :])
+        in SYSTEM_DIRS
```
scheibelp: This function should only be called on paths: if you have `-L/path/to...` then any calling function should strip `-L` before calling this function.

var/spack/repos/builtin/packages/glib/package.py
```diff
@@ -339,10 +340,14 @@ def gettext_libdir(self):
...
         # bypassing spack's compiler wrapper for linking and thus not finding
-        # the gettext library directory. The patch below explitly adds the
+        # the gettext library directory. The patch below explicitly adds the
         # appropriate -L path.
         spec = self.spec
-        if spec.satisfies("@2.0:2"):
+        if (
+            spec.satisfies("@2.0:2")
+            and "intl" in self.spec["gettext"].libs.names
+            and not is_system_path(spec["gettext"].libs.directories[0])
```
scheibelp: (no change needed) I'll confirm this change is likely for the best: checking `.external` isn't enough.

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
```diff
@@ -77,6 +73,20 @@ class PyOnnxRuntime(CMakePackage, PythonExtension):
...
+    # https://github.com/microsoft/onnxruntime/issues/4234#issuecomment-698077636
+    @when("@1.7.2")
+    def patch(self):
+        if self.spec["iconv"].name == "libiconv":
+            patch("libiconv.patch", level=0)
+
+    @when("@1.10.0")
+    def patch(self):
+        if self.spec["iconv"].name == "libiconv":
+            patch("libiconv-1.10.patch", level=0)
```
scheibelp: Note that the original patch declaration `patch("libiconv.patch", level=0, when="@1.7.2 ^libiconv")` should work: `^libiconv` is specifically asking whether the `libiconv` implementation of `iconv` is a dependency.

(I see you did something similar to that for `fsl` though in this PR, so maybe I'm missing something)

var/spack/repos/builtin/packages/tar/package.py
```diff
@@ -65,12 +65,16 @@ def configure_args(self):
...
         # so that tar can locate them when invoked without spack load.
         args = [
-            "--with-libiconv-prefix={0}".format(self.spec["iconv"].prefix),
             "--with-xz={0}".format(self.spec["xz"].prefix.bin.xz),
             "--with-lzma={0}".format(self.spec["xz"].prefix.bin.lzma),
             "--with-bzip2={0}".format(self.spec["bzip2"].prefix.bin.bzip2),
         ]
 
+        if self.spec["iconv"].name == "libc":
+            args.append("--without-libiconv")
```
scheibelp: (question) High-level question not specific to this package, but this is a relevant example: when you say `--without-libiconv`, is it telling the configure script not to use it at all? Or is it telling the configure script that it doesn't have to look (e.g. because it's in libc). On further consideration, it's surprising to me that the configure script doesn't check for this automatically.

var/spack/repos/builtin/packages/extrae/package.py
```diff
@@ -132,7 +132,7 @@ def configure_args(self):
         # https://www.gnu.org/software/gettext/FAQ.html#integrating_undefined
         # - linking error
         # https://github.com/bsc-performance-tools/extrae/issues/57
-        args.append("LDFLAGS=-lintl -pthread")
```
scheibelp, line 135: FYI this removes `-lpthread`, which I don't think is in `gettext.libs`

var/spack/repos/builtin/packages/elfutils/package.py
```diff
@@ -121,8 +122,10 @@ def configure_args(self):
         args.append("--with-zlib=%s" % spec["zlib"].prefix)
 
         if "+nls" in spec:
-            # configure doesn't use LIBS correctly
-            args.append("LDFLAGS=-Wl,--no-as-needed -L%s -lintl" % spec["gettext"].prefix.lib)
+            args.append(
+                "LDFLAGS=-Wl,--no-as-needed "
+                + " ".join(filter_system_paths(self.spec["gettext"].libs.ld_flags.split()))
```
scheibelp: You may need the same care here that you applied in glib: avoid `-L` of system directories. It occurs to me that `gettext` could exclude system directories, e.g. `libs` can do:

```
def libs(self):
  libs = find_libraries(...)
  libs.directories = [] # you can set this after generating the object
  return libs
```

scheibelp (commented):
I have a couple requests and questions

greenc-FNAL:
@scheibelp I've done some rather extensive rewrites after tracing the `elfutils` CI issue. When the CI is successful I'll go over your comments and address those that I think remain applicable. Feel free to let me know at that time if you think I missed any.

lib/spack/spack/util/environment.py
greenc-FNAL: Bypassed the need for it with other changes.

var/spack/repos/builtin/packages/glib/package.py
greenc-FNAL: Slight tweak to use `is_system_path(spec["gettext"].prefix)` in the conditional instead of `is_system_path(spec["gettext"].libs.directories[0])` in case `spec["gettext"].libs.directories` is empty.

var/spack/repos/builtin/packages/tar/package.py
greenc-FNAL: It should have been, `--without-libiconv-prefix`: see https://github.com/spack/spack/blob/fd11c09996f1059e523a93def026cad38003d78d/var/spack/repos/builtin/packages/tar/package.py#L75-L78

var/spack/repos/builtin/packages/extrae/package.py
```diff
@@ -127,14 +127,20 @@ def configure_args(self):
...
-
-        return args
+        if name == "ldlibs":
+            if "intl" in self.spec["gettext"].libs.names:
+                flags.append("-lintl")
+        elif name == "ldflags":
+            flags.append("-pthread")
+        return self.build_system_flags(name, flags)
```
greenc-FNAL, line 143: Relevant to https://github.com/spack/spack/pull/35450/files#r1164714416

var/spack/repos/builtin/packages/extrae/package.py
greenc-FNAL: See https://github.com/spack/spack/pull/35450/files#r1172727456

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
greenc-FNAL: I was probably confused a little here: `py-onnx-runtime` is deprecated in favor of `py-onnxruntime`, and they both had a longstanding typo in a `patch()` line. Change above reverted and typo fixed.

var/spack/repos/builtin/packages/elfutils/package.py
greenc-FNAL: This has been resolved in a different way, see https://github.com/spack/spack/blob/fd11c09996f1059e523a93def026cad38003d78d/var/spack/repos/builtin/packages/elfutils/package.py#L135-L143

greenc-FNAL: Note: with respect to your suggestion to clear `libs.directories`, it is apparently habit to refer to `libs.directories[0]` without further checks. I could clear up the remaining uses of those, but it would create a fragility with respect to future recipe changes.

greenc-FNAL:
Closes #36569.
Closes #34541.

var/spack/repos/builtin/packages/acl/package.py
greenc-FNAL: It appears that, of packages requiring `libintl`, most of them require only that and not any of the others. Indeed, putting the full list in for `elfutils` actually caused an ODR violation as `color_mode` is defined in `libeu.a` and `libgettextsrc` (IIRC). In other words, using only `-lintl` when needed, rather than the full `libs.link_flags`, seems generally to be the way to go.

var/spack/repos/builtin/packages/gdal/package.py
scheibelp: @greenc-FNAL I'll note that our compiler wrappers reorder `-L` etc. so even if the configure script introduces a mixed order of system/custom paths, our wrapper should disentangle that. It would be useful to see an example failed installation and specifically the corresponding modified compiler invocations you can see in `spack-cc..out` if you run the install with `-d` like `spack -d install ...`.

var/spack/repos/builtin/packages/binutils/package.py
```diff
@@ -269,9 +269,12 @@ def install_headers(self):
...
-            env.append_flags("LDFLAGS", "-lintl")
+    def flag_handler(self, name, flags):
+        spec = self.spec
+        if name == "ldflags":
+            if spec.satisfies("%cce"):
+                flags.append("-Wl,-z,muldefs")
+        elif name == "ldlibs":
+            if "+nls" in self.spec and "intl" in self.spec["gettext"].libs.names:
+                flags.append("-lintl")
+        return self.build_system_flags(name, flags)
```
scheibelp, line 280: (question) I'll note this changes the default behavior for all flags to make them build system arguments (the default is injecting flags) - was that intentional? That's fine if so.

var/spack/repos/builtin/packages/elfutils/package.py
```diff
@@ -99,7 +100,12 @@ def patch(self):
         files = glob.glob(os.path.join("*", "Makefile.in"))
         filter_file("-Werror", "", *files)
 
-    flag_handler = AutotoolsPackage.build_system_flags
+    def flag_handler(self, name, flags):
+        if name == "ldlibs":
+            spec = self.spec
+            if "+nls" in spec and "intl" in spec["gettext"].libs.names:
+                flags.append("-lintl")
+        return self.inject_flags(name, flags)
```
scheibelp, line 108: (question) this changes the default to always inject (in this package, the flag handler was set to export as build system arguments) - was that intentional?

var/spack/repos/builtin/packages/gettext/package.py
```diff
@@ -112,6 +117,10 @@ def configure_args(self):
 
         return config_args
 
+    @property
+    def gettext_libs(self):
```
scheibelp: This `<packagename>_libs` format is typically for packages which provide multiple virtuals (e.g. if `gettext` provided a `gettext` virtual, and something else) - was something going wrong when you didn't define this?

var/spack/repos/builtin/packages/xfsprogs/package.py
```diff
@@ -25,21 +25,19 @@ class Xfsprogs(AutotoolsPackage):
...
-        iflags = []
         if name == "cflags":
             if self.spec.satisfies("@:5.4.0 %gcc@10:"):
-                iflags.append("-fcommon")
-        return (iflags, None, flags)
+                flags.append("-fcommon")
+        elif name == "ldlibs":
+            if "intl" in self.spec["gettext"].libs.names:
+                flags.append("-lintl")
+        return build_system_flags(name, iflags)
```
scheibelp: `iflags` is undefined

Also note: the original behavior split these flags across injection and build system args (fine if you intentionally made them all exported as build system args)

var/spack/repos/builtin/packages/xfsdump/package.py
```diff
@@ -25,8 +25,10 @@ class Xfsdump(MakefilePackage):
     depends_on("attr")
     depends_on("xfsprogs@:4.20.0")
 
-    def setup_build_environment(self, env):
-        env.append_flags("LDFLAGS", "-lintl")
+    def flag_handler(self, name, flags):
+        if name == "ldlibs" and "intl" in self.spec["gettext"].libs.names:
+            flags.append("-lintl")
+        return build_system_flags(name, flags)
```
scheibelp: Note that `MakefilePackage` does not implement `flags_to_build_system_args` (like `CMakePackage` does), so this may not have any effect: I think you may want to use `env_flags` here.

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
```diff
@@ -55,7 +55,7 @@ class PyOnnxRuntime(CMakePackage, PythonExtension):
     # https://github.com/cms-externals/onnxruntime/compare/5bc92df...d594f80
     patch("cms.patch", level=1, when="@1.7.2")
     # https://github.com/cms-externals/onnxruntime/compare/0d9030e...7a6355a
-    patch("cms_1_10.patch", whe="@1.10")
+    patch("cms_1_10.patch", when="@1.10")
```
scheibelp: It appears this is a duplicate package: ideally this could be split off into another PR (which removes one of the duplicates and updates dependencies accordingly).

var/spack/repos/builtin/packages/procps/package.py
```diff
@@ -17,14 +18,16 @@ class Procps(AutotoolsPackage):
...
+
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool", type="build")
     depends_on("m4", type="build")
     depends_on("pkgconfig@0.9.0:", type="build")
     depends_on("dejagnu", type="test")
     depends_on("iconv")
-    depends_on("gettext")
+    depends_on("gettext", when="+nls")
```
scheibelp, line 30: (question) Was this made optional on account of some additional issues? I don't immediately see how introducing the variant is part of addressing the problems this PR intends to solve.

scheibelp (changes requested):
Overall looking good, with a couple of typos. General notes:

* In several cases this just swaps out an explicit dep on `libiconv` for `iconv` virtual, which seems like a good thing to do in terms of linking consistency (although somewhat a separate concept to the core issue that this PR tackles)
* I'm wondering if `libintl` should be treated like `iconv` a virtual that may be provided by libc. I think that is the case but it is not modeled that way, so the libiconv/libintl handling differs 

As a side note, it would be nice to add a `configure_iconv_prefix_arg` method to all implementations of the `iconv` virtual, which would generally avoid repetition of the following code block:

```
            if spec["iconv"].name == "libc":
                args.append("--without-libiconv-prefix")
            elif not is_system_path(spec["iconv"].prefix):
                args.append("--with-libiconv-prefix={p}".format(p=spec["iconv"].prefix))
```

That doesn't have to be done in this PR, but I'll be considering how to refactor this.

var/spack/repos/builtin/packages/binutils/package.py
greenc-FNAL: Yes, that was intentional.

var/spack/repos/builtin/packages/elfutils/package.py
greenc-FNAL: Yes, that was intentional.

var/spack/repos/builtin/packages/gettext/package.py
greenc-FNAL: This may be a vestigial holdover from an earlier attempt. I can remove.

var/spack/repos/builtin/packages/xfsprogs/package.py
greenc-FNAL: The single export type was intentional. The vestigial `iflags` was not: will fix.

var/spack/repos/builtin/packages/xfsdump/package.py
greenc-FNAL: Yep, that edit was probably a mixup with `xfsprogs`: will fix.

var/spack/repos/builtin/packages/procps/package.py
greenc-FNAL: The rest of the improvements to this recipe provide the consistency: this just removes an unnecessary dependency on `gettext` when native language support is not required.

var/spack/repos/builtin/packages/py-onnx-runtime/package.py
greenc-FNAL: > Overall looking good, with a couple of typos. General notes:
> 
> * In several cases this just swaps out an explicit dep on `libiconv` for `iconv` virtual, which seems like a good thing to do in terms of linking consistency (although somewhat a separate concept to the core issue that this PR tackles)

I believe this is necessary for consistency: if `libc` is selected as the required provider for `iconv`, then packages in the DAG with explicit dependencies on `libiconv` could cause problems.

> * I'm wondering if `libintl` should be treated like `iconv` a virtual that may be provided by libc. I think that is the case but it is not modeled that way, so the libiconv/libintl handling differs

I think this was deliberately not done because `gettext` may be required for non-`libintl` reasons. It is perfectly reasonable to use `libc` for `libintl` even in the presence of `gettext` in the DAG.

> As a side note, it would be nice to add a `configure_iconv_prefix_arg` method to all implementations of the `iconv` virtual, which would generally avoid repetition of the following code block:
> 
> ```
>             if spec["iconv"].name == "libc":
>                 args.append("--without-libiconv-prefix")
>             elif not is_system_path(spec["iconv"].prefix):
>                 args.append("--with-libiconv-prefix={p}".format(p=spec["iconv"].prefix))
> ```
> 
> That doesn't have to be done in this PR, but I'll be considering how to refactor this.

This is the "modern" way for the `autotools` `m4` macros to handle this, but it is not the case for older `autotools` versions. In other words, the implementations of `configure_iconv_prefix_arg()` would have to consider `autotools` versions. Even that may not be comprehensive due to package-specific implementations of the handling, but I haven't done a survey.

scheibelp (approved):

espentrydal:
The last commit to libxpm packages.py breaks it on RHEL8. Still compiles on fedora.

Error message is that it can't find libintl:
```

  >> 305    /tmp/espen/spack-stage/spack-stage-libxpm-3.5.12-qp3nu2zggqtetamiwvrmwixheidazy7z/
            spack-src/cxpm/cxpm.c:149: undefined reference to `libintl_bindtextdomain'
  >> 306    /tmp/espen/spack-stage/spack-stage-libxpm-3.5.12-qp3nu2zggqtetamiwvrmwixheidazy7z/
            spack-src/cxpm/cxpm.c:150: undefined reference to `libintl_textdomain'
  >> 307    /tmp/espen/spack-stage/spack-stage-libxpm-3.5.12-qp3nu2zggqtetamiwvrmwixheidazy7z/
            spack-src/cxpm/cxpm.c:157: undefined reference to `libintl_gettext'

```
@scheibelp

espentrydal:
Reverting the change fixes it

scheibelp:
@espentrydal thanks for the info: I can investigate this later this week. I would have assumed that the before/after for that package were equivalent methods of conveying the libintl dependency so I want to try to reproduce your issue

bernhardkaindl:
@espentrydal @scheibelp: I guess the initial line in `libxpm/package.py` which had `-L<prefix.lib>` should help:
```py
env.append_flags("LDFLAGS", "-L{0} -lintl".format(self.spec["gettext"].prefix.lib))
```

It might also be a good idea to have a look at #29698 which might be helpful when libxpm itself is an external package from a Linux system which also does not add a libintl into the build of dependent packages.

greenc-FNAL:
> The last commit to libxpm packages.py breaks it on RHEL8. Still compiles on fedora.
> 
> Error message is that it can't find libintl:
> 
> ```
> 
>   >> 305    /tmp/espen/spack-stage/spack-stage-libxpm-3.5.12-qp3nu2zggqtetamiwvrmwixheidazy7z/
>             spack-src/cxpm/cxpm.c:149: undefined reference to `libintl_bindtextdomain'
>   >> 306    /tmp/espen/spack-stage/spack-stage-libxpm-3.5.12-qp3nu2zggqtetamiwvrmwixheidazy7z/
>             spack-src/cxpm/cxpm.c:150: undefined reference to `libintl_textdomain'
>   >> 307    /tmp/espen/spack-stage/spack-stage-libxpm-3.5.12-qp3nu2zggqtetamiwvrmwixheidazy7z/
>             spack-src/cxpm/cxpm.c:157: undefined reference to `libintl_gettext'
> ```
> 
> @scheibelp

@espentrydal Can i ask you to put in a new issue for this, please? This wasn't caught by the CI because it seems to be very platform-specific, so please provide full details of your system and minimal reproducing command(s).

jrood-nrel:
I came here to post that same error. https://github.com/spack/spack/pull/36568 works for me, but this PR broke libxpm for me.

var/spack/repos/builtin/packages/git/package.py
```diff
@@ -466,12 +467,18 @@ def setup_build_environment(self, env):
...
+        spec = self.spec
+        if "+nls" in spec:
+            if "intl" in spec["gettext"].libs.names:
+                extlib_bits = []
+                if not is_system_path(spec["gettext"].prefix):
+                    extlib_bits.append(spec["gettext"].libs.search_flags)
+                extlib_bits.append("-lintl")
+                env.append_flags("EXTLIBS", " ".join(extlib_bits))
+            if not is_system_path(spec["gettext"].prefix):
+                env.append_flags("CFLAGS", spec["gettext"].headers.include_flags)
```
haampie, line 479: This drops optimization flags :(

var/spack/repos/builtin/packages/mono/package.py
```diff
@@ -74,6 +75,8 @@ def patch(self):
 
     def configure_args(self):
         args = []
-        li = self.spec["iconv"].prefix
-        args.append("--with-libiconv-prefix={p}".format(p=li))
+        if self.spec["iconv"].name == "libc":
+            args.append("--without-libiconv-prefix")
+        elif not is_system_path(self.spec["iconv"].prefix):
+            args.append("--with-libiconv-prefix={p}".format(p=self.spec["iconv"].prefix))
```
grospelliergilles, line 81: I have looked at the configure and `mono` does not seem to use configure arguments `--without-libiconv-prefix` or `--with-libiconv-prefix`. When it searches for `libiconv` it uses default flags. If we want to change the path for `libiconv`,  we have to use `CPPFLAGS` or `LDFLAGS` environment variables.

PR #35455, opened 2023-02-13
Add fortls

RMeli:


tldahlgren (commented):
Confirmed the sha256.

tldahlgren:

var/spack/repos/builtin/packages/py-fortls/package.py
```diff
@@ -0,0 +1,26 @@
...
+    homepage = "https://fortls.fortran-lang.org"
+    pypi = "fortls/fortls-2.13.0.tar.gz"
+
+    maintainers("RMeli")
+
+    version("2.13.0", sha256="23c5013e8dd8e1d65bf07be610d0827bc48aa7331a7a7ce13612d4c646d0db31")
+
+    depends_on("python@3.7:")
+
+    depends_on("py-setuptools", type="build")
```
adamjstewart: Missing a bunch of deps:
```toml
requires = [
    "setuptools >= 45",
    "wheel",
    "setuptools_scm[toml] >= 6.2",
    "setuptools_scm_git_archive",
]
```

var/spack/repos/builtin/packages/py-fortls/package.py
```diff
@@ -0,0 +1,26 @@
...
+    depends_on("py-packaging", type="build")
```
adamjstewart: Should be type=build/run

var/spack/repos/builtin/packages/py-fortls/package.py
```diff
@@ -0,0 +1,26 @@
...
+    """A Language Server for Fortran providing code completion, diagnostics, hovering and more."""
+
+    homepage = "https://fortls.fortran-lang.org"
+    pypi = "fortls/fortls-2.13.0.tar.gz"
+
+    maintainers("RMeli")
+
+    version("2.13.0", sha256="23c5013e8dd8e1d65bf07be610d0827bc48aa7331a7a7ce13612d4c646d0db31")
+
+    depends_on("python@3.7:")
```
adamjstewart: Should probably be type=build/run. However, Spack only supports Python 3.7+, so you could also just remove this line.

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-fortls/package.py
RMeli:
```suggestion
```

RMeli:
Thanks @adamjstewart @tldahlgren

var/spack/repos/builtin/packages/py-fortls/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+    maintainers("RMeli")
+
+    version("2.13.0", sha256="23c5013e8dd8e1d65bf07be610d0827bc48aa7331a7a7ce13612d4c646d0db31")
+
+    depends_on("py-setuptools@45:", type=("build", "run"))
+    depends_on("py-wheel", type=("build", "run"))
+    depends_on("py-packaging", type=("build", "run"))
+    depends_on("py-setuptools-scm@6.2:+toml", type=("build", "run"))
+    depends_on("py-setuptools-scm-git-archive", type=("build", "run"))
```
adamjstewart: These should all be type=build except for packaging which is type=build/run

var/spack/repos/builtin/packages/py-fortls/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+    homepage = "https://fortls.fortran-lang.org"
+    pypi = "fortls/fortls-2.13.0.tar.gz"
+
+    maintainers("RMeli")
+
+    version("2.13.0", sha256="23c5013e8dd8e1d65bf07be610d0827bc48aa7331a7a7ce13612d4c646d0db31")
+
+    depends_on("py-setuptools@45:", type=("build", "run"))
+    depends_on("py-wheel", type=("build", "run"))
```
adamjstewart: wheel is also added by the base class so technically you don't need to repeat it

var/spack/repos/builtin/packages/py-fortls/package.py
```diff
@@ -0,0 +1,27 @@
...
+    depends_on("py-packaging", type=("build", "run"))
+    depends_on("py-setuptools-scm@6.2:+toml", type=("build", "run"))
+    depends_on("py-setuptools-scm-git-archive", type=("build", "run"))
+
+    depends_on("py-json5", type=("build", "run"))
+    depends_on("py-importlib-metadata", type=("build", "run"), when="python@3.7")
+    depends_on("py-typing-extensions", type=("build", "run"), when="python@3.7")
```
adamjstewart:
```suggestion
    depends_on("py-importlib-metadata", type=("build", "run"), when="^python@:3.7")
    depends_on("py-typing-extensions", type=("build", "run"), when="^python@:3.7")
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-fortls/package.py
RMeli: Sorry, I thought in the comment above also `setuptools` needed to be of type `run` for some reason. I got confused...

adamjstewart (approved):

PR #35457, opened 2023-02-13
nag: update the versioning scheme

skosukhin:
We have been ignoring the `Build` version number of the NAG compiler for too long and it's finally time to add it to the versioning scheme of the compiler. It looks like for every `<major>.<minor>` version it's possible to download only the latest `<build>` version. To be able to install other `<build>` versions, we have to rely on Spack's cache.

### DIFF
--- a/lib/spack/spack/compilers/nag.py
+++ b/lib/spack/spack/compilers/nag.py
@@ -4,7 +4,10 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import os
+import re
 from typing import List
+
+import llnl.util.lang
 
 import spack.compiler
 
@@ -32,7 +35,13 @@ class Nag(spack.compiler.Compiler):
     }
 
     version_argument = "-V"
-    version_regex = r"NAG Fortran Compiler Release ([0-9.]+)"
+
+    @classmethod
+    @llnl.util.lang.memoized
+    def extract_version_from_output(cls, output):
+        match = re.search(r"NAG Fortran Compiler Release (\d+).(\d+)\(.*\) Build (\d+)", output)
+        if match:
+            return ".".join(match.groups())
 
     @property
     def verbose_flag(self):
--- a/lib/spack/spack/test/compilers/detection.py
+++ b/lib/spack/spack/test/compilers/detection.py
@@ -281,7 +281,7 @@ def test_oneapi_version_detection(version_str, expected_version):
         (
             "NAG Fortran Compiler Release 6.0(Hibiya) Build 1037\n"
             "Product NPL6A60NA for x86-64 Linux\n",
-            "6.0",
+            "6.0.1037",
         )
     ],
 )
--- a/var/spack/repos/builtin/packages/nag/package.py
+++ b/var/spack/repos/builtin/packages/nag/package.py
@@ -17,6 +17,7 @@ class Nag(Package):
     homepage = "https://www.nag.com/nagware/np.asp"
     maintainers("skosukhin")
 
+    version("7.1.7125", sha256="738ed9ed943ebeb05d337cfdc603b9c88b8642b3d0cafea8d2872f36201adb37")
     version("7.1.7101", sha256="18640737b232cebeb532ba36187675cdaf36d5b1fc235a780fc9e588c19a3ed2")
     version("7.0.7048", sha256="6d509208533d79139e5a9f879b7b93e7b58372b78d404d51f35e491ecbaa54c7")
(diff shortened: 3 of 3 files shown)


tldahlgren (commented):
I am unable to check the sha for any of the 5 affected versions, perhaps due to licensing.

skosukhin:
@tldahlgren please, note that there is only one new sha (for the new version `7.1.7114`). And the tarball is publicly available on the [official page](https://www.nag.com/content/downloads-nag-fortran-compiler-versions) (product code NPL6A71NA). All other shas are the same (however, as I have already mentioned, tarballs for some of the versions, e.g. `7.1.7101` and `6.1.6136`, are not available on the official web page anymore and can be fetched from Spack cache only).

tldahlgren:
> @tldahlgren please, note that there is only one new sha (for the new version `7.1.7114`). And the tarball is publicly available on the [official page](https://www.nag.com/content/downloads-nag-fortran-compiler-versions) (product code NPL6A71NA). All other shas are the same (however, as I have already mentioned, tarballs for some of the versions, e.g. `7.1.7101` and `6.1.6136`, are not available on the official web page anymore and can be fetched from Spack cache only).

I need to be more careful how I describe the checks.  I have a script that automatically checks all new/modified `version` statements; hence, the 5 I mentioned.

skosukhin:
@tldahlgren three of the five ([`6.2.6252`](https://www.nag.com/downloads/impl/npl6a62na_amd64.tgz), [`7.0.7048`](https://www.nag.com/downloads/impl/npl6a70na_amd64.tgz), [`7.1.7114`](https://www.nag.com/downloads/impl/npl6a71na_amd64.tgz)) should work as they are (still) publicly available. It's strange that the script you use produces negative results for all five.

tldahlgren (commented):
Confirmed the sha256 for 7.1.7114, 7.0.7048, 6.2.6252.  Getting "Could not find any remote versions for nag" for 6.1.6136, suggesting setting it as `deprecated` (https://spack.readthedocs.io/en/latest/packaging_guide.html#deprecating-old-versions).

Getting a different sha256 for 7.1.7101:

```
    version("7.1.7101", sha256="0f173ea0e0809673db216622a7354cc2eb52bf13372a34ef85e62a115bb6f94f")
```

tldahlgren:
> @tldahlgren three of the five ([`6.2.6252`](https://www.nag.com/downloads/impl/npl6a62na_amd64.tgz), [`7.0.7048`](https://www.nag.com/downloads/impl/npl6a70na_amd64.tgz), [`7.1.7114`](https://www.nag.com/downloads/impl/npl6a71na_amd64.tgz)) should work as they are (still) publicly available. It's strange that the script you use produces negative results for all five.

Tried again after making sure I had the PR checked out.

skosukhin:
@tldahlgren thank you for double-checking.

@adamjstewart how about this? Versions that cannot be downloaded from the official site are now deprecated and for deprecated versions, we expect "manual download", i.e. their URLs start with `file://` and the `manual_download` is `True` (although e93187b is a hack, so I'm not sure about that). I also noticed that the `Build` numbers are prefixed with `<major><minor>` digits, so it might make sense to drop them (i.e. we have `7.1.14` instead of `7.1.7114`, see 61fce14). The only exception to that that I know is in the [respective test](https://github.com/spack/spack/blob/e75a07d1558e1271e561b8feddd2de2e02fd6cdc/lib/spack/spack/test/compilers/detection.py#L276), so I assume that this scheme hasn't always been in place but it is for all versions that we want or can currently support. On the other hand, the scheme might change again and we will have to rework the versioning again, which is disruptive and to be avoided. So, I'm not 100% about this either, although the result looks much nicer.

adamjstewart:
At one point we had a good relationship with NAG, and @ThemosTsikas submitted multiple PRs to improve NAG support in Spack. I seem to recall him leaving NAG, but there should still be NAG developers who know about Spack. I still think it's worth reaching out to them and seeing if stable URLs would be an option. Especially if anyone has a support contract. Mention the need for security and stable checksums for URLs on DOE and DOD supercomputers and that should get their attention.

skosukhin:
@adamjstewart Unfortunately, I haven't managed to get an answer from NAG regarding the URLs in the last several months (although they seem to be quite responsive when we ask them other questions). However, I would like to move forward with this PR. I have rebased the branch and reverted all the hacky modifications suggested in https://github.com/spack/spack/pull/35457#issuecomment-1430026098. I think we can simply deprecate all versions that cannot be downloaded from the official page and set their `url`s to `file:///...`. This way, the users can still install the deprecated versions either from Spack mirror or from the local tarball if they happen to have one.

tldahlgren (approved):

PR #35460, opened 2023-02-13
py-jax: add v0.4.3

adamjstewart:
Successfully installs on RHEL 7 with Python 3.10.8 and GCC 12.2.0.

### DIFF
--- a/var/spack/repos/builtin/packages/py-jax/package.py
+++ b/var/spack/repos/builtin/packages/py-jax/package.py
@@ -41,7 +41,7 @@ class PyJax(PythonPackage):
     }
 
     for jax, jaxlib in jax_to_jaxlib.items():
-        depends_on(f"py-jaxlib@{jaxlib}", when=f"@{jax}", type=("build", "run"))
+        depends_on(f"py-jaxlib@{jaxlib}:", when=f"@{jax}", type=("build", "run"))
 
     # Historical dependencies
     depends_on("py-absl-py", when="@:0.3", type=("build", "run"))
--- a/var/spack/repos/builtin/packages/py-jax/package.py
+++ b/var/spack/repos/builtin/packages/py-jax/package.py
@@ -7,7 +7,7 @@
 from spack.package import *
 
 
-class PyJax(PythonPackage, CudaPackage):
+class PyJax(PythonPackage):
     """JAX is Autograd and XLA, brought together for high-performance
     machine learning research. With its updated version of Autograd,
     JAX can automatically differentiate native Python and NumPy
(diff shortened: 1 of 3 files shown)


var/spack/repos/builtin/packages/py-jax/package.py
```diff
@@ -7,7 +7,7 @@
 from spack.package import *
 
 
-class PyJax(PythonPackage, CudaPackage):
+class PyJax(PythonPackage):
```
adamjstewart, line 10: JAX is pure-Python and doesn't use CUDA. If you want to control which cuda_arch is used by jaxlib, you should use:
```console
$ spack install py-jax ^py-jaxlib +cuda cuda_arch=...
```

var/spack/repos/builtin/packages/py-jaxlib/package.py
```diff
@@ -17,25 +17,52 @@ class PyJaxlib(PythonPackage, CudaPackage):
...
+
+    # .bazelversion
+    depends_on("bazel@5.1.1:", when="@0.3:", type="build")
+    # https://github.com/google/jax/issues/8440
+    depends_on("bazel@4.1:4", when="@0.1", type="build")
+
+    # README.md
+    depends_on("cuda@11.4:", when="@0.4:+cuda")
+    depends_on("cuda@11.1:", when="@0.3+cuda")
+    depends_on("cuda@11.1:11.7.0", when="@0.1+cuda")  # TODO: why the upper bound?
```
adamjstewart: Any idea where this upper bound came from?

tldahlgren (commented):
Confirmed the new `py-jax` and `py-jaxlib` sha256.

tldahlgren:
I am curious what the listed maintainer/reviewers think about removing `cuda` options.

var/spack/repos/builtin/packages/py-jax/package.py
aweits: I would think that this kind of default install would surprise most users, no? also, if we keep this, alphafold should probably be addressed.

var/spack/repos/builtin/packages/py-jaxlib/package.py
aweits: it's a headerfile change in cuda 11.7.1 that angers the old bundled tf install under the hood - i can dig out my detailed notes if necessary...

var/spack/repos/builtin/packages/py-jax/package.py
adamjstewart: What kind of default install? We can tell py-jaxlib to build +cuda by default if you want.

var/spack/repos/builtin/packages/py-jaxlib/package.py
adamjstewart: If you have a link to an issue that would be great, otherwise I'll add a comment to that affect

var/spack/repos/builtin/packages/py-jax/package.py
adamjstewart: Is it the case that alphafold requires `py-jax ^py-jaxlib+cuda`? We can add a fake dep on `py-jaxlib` if that helps. I try to avoid variant forwarding, especially now that we have `++` and `==` variants in the spec syntax.

aweits: I think jaxlib +cuda by default is reasonable...

adamjstewart: Should alphafold also default to +cuda?

var/spack/repos/builtin/packages/py-jaxlib/package.py
aweits: we've been here before https://github.com/google/jax/issues/12614

aweits (approved):

PR #35470, opened 2023-02-13
py-radiant-mlhub: add new versions

adamjstewart:
Successfully installs on macOS 10.15.7 with Python 3.10.8 and Apple Clang 12.0.0.


tldahlgren (approved):
Confirmed the three version sha256.

PR #35472, opened 2023-02-14
fix intel-tbb build with rocmcc

tukss:
rocmcc is a clang compiler. Pass that information on to the TBB build system.


rscohn2 (approved):
Thanks!

PR #34661, opened 2022-12-22
view: fix issue with non-contributing specs

haampie:
Closes #32456
Closes #35475

Specs that did not contribute any files to an env view caused a problem
where zip(specs, files grouped by prefix) got "out of sync", causing the
wrong merge map to be passed to a package's `add_files_to_view`, which
specifically caused an issue where *sometimes* bin/python ended up as a
symlink instead of a copy.

One such example is kokkos + kokkos-nvcc-wrapper, as the latter package
only provides the file bin/nvcc_wrapper, which is also added to view by
kokkos, causing kokkos-nvcc-wrapper to contribute 0 files.

The test feels a bit contrived, but it captures the problem... pkg a is
added first and has 0 files to contribute, pkg b adds a single file, and
we check if pkg b receives a merge map (and a does not).

PR #35475, opened 2023-02-14
views: specs were not being associated with the correct file group

scheibelp:
This fixes a bug which affects all packages added to views when those packages have specialized view logic (i.e. any `Package` with a specialized implementation of `add_files_to_view`, which includes all `py-*` packages). There was an incorrect mapping which meant that specs were not grouped correctly with the prefix and set-of-files associated with that spec, namely the line:

```
for (spec, (src_root, rel_paths)) in zip(specs, files_per_spec):
```

(in particular because the order of `files_per_spec` does not relate to `specs` so `zip`ping them effectively randomizes the file paths associated with each spec)

### DIFF
--- a/lib/spack/spack/filesystem_view.py
+++ b/lib/spack/spack/filesystem_view.py
@@ -690,8 +690,17 @@ class SimpleFilesystemView(FilesystemView):
         # For compatibility, we have to create a merge_map dict mapping
         # full_src => full_dst
         files_per_spec = itertools.groupby(visitor.files.items(), key=lambda item: item[1][0])
+        prefix_to_files = dict()
+        for src_root, rel_paths in files_per_spec:
+            prefix_to_files[src_root] = list(rel_paths)
 
-        for (spec, (src_root, rel_paths)) in zip(specs, files_per_spec):
+        spec_to_file_group = dict()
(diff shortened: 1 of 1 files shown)


lib/spack/spack/filesystem_view.py
```diff
@@ -690,8 +690,17 @@ def skip_list(file):
         # For compatibility, we have to create a merge_map dict mapping
         # full_src => full_dst
         files_per_spec = itertools.groupby(visitor.files.items(), key=lambda item: item[1][0])
+        prefix_to_files = dict()
+        for src_root, rel_paths in files_per_spec:
+            prefix_to_files[src_root] = list(rel_paths)
```
scheibelp, line 695: If I replace the prior 3 lines with:

```
prefix_to_files = dict(files_per_spec)
```

then nothing actually gets linked into the view. I'm not clear why, but this is more verbose than one might opt for on account of that.

haampie:
The order isn't arbitrary, it's that it doesn't account for packages not providing any new files to the view.

I've fixed the issue here, but didn't get anyone to review

https://github.com/spack/spack/pull/34661

haampie:
Closing this in favor of #34661

PR #35477, opened 2023-02-14
octave: fftw-api should be used for amdfftw and mkl

nkgh77:
fftw-api should be used instead of fftw because amdfftw and mkl can be linked as e.g. 
$ spack install octave+fftw ^amdfftw


siko1056 (approved):
LGTM 👍

tldahlgren (approved):
LGTM

PR #35481, opened 2023-02-14
Orca: Added required runtime libraries

hseara:
Several libraries are need to be present at run time so that the code can be run in parallel.  I have added them as dependencies and to LD_LIBRARY_PATH. Orca comes as a binary so  the libraries cannot be added as RPATH at compilation time.

Also, orca 5.0.3 was compiled against 4.1.1, not 4.1.2.

### DIFF
--- a/var/spack/repos/builtin/packages/orca/package.py
+++ b/var/spack/repos/builtin/packages/orca/package.py
@@ -57,7 +57,7 @@ class Orca(Package):
         "4.0.1.2": "2.0.2",
         "4.2.0": "3.1.4",
         "4.2.1": "3.1.4",
-        "5.0.3": "4.1.1",
+        "5.0.3": "4.1.2",
         "5.0.3-f.1": "4.1.2",
     }
     for orca_version, openmpi_version in openmpi_versions.items():
--- a/var/spack/repos/builtin/packages/orca/package.py
+++ b/var/spack/repos/builtin/packages/orca/package.py
@@ -49,13 +49,15 @@ class Orca(Package):
     )
 
     depends_on("zstd", when="@:4.2.1", type="build")
+    depends_on("libevent", type="run")
+    depends_on("libpciaccess", type="run")
 
     # Map Orca version with the required OpenMPI version
     openmpi_versions = {
         "4.0.1.2": "2.0.2",
         "4.2.0": "3.1.4",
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/orca/package.py
```diff
@@ -49,13 +49,15 @@ class Orca(Package):
...
+    depends_on("libevent", type="run")
+    depends_on("libpciaccess", type="run")
 
     # Map Orca version with the required OpenMPI version
     openmpi_versions = {
         "4.0.1.2": "2.0.2",
         "4.2.0": "3.1.4",
         "4.2.1": "3.1.4",
-        "5.0.3": "4.1.2",
+        "5.0.3": "4.1.1",
```
snehring: 4.1.2 was specified previously due to a build error I was seeing with 4.1.1 specifically [in the past](https://github.com/spack/spack/pull/33649#issue-1432011610)

snehring (changes requested):

var/spack/repos/builtin/packages/orca/package.py
hseara: We had some memory issues running with the previous build in parallel. After adding the libraries to the runtime environment and changing the `openmpi` version, the problems were gone. After adding the libraries to the runtime environment, I must admit that I still need to check 4.1.2. However,, guided by my past experience with `orca,` which proved to be very stubborn to the minor version of `openmpi,` I proposed the change. If it should work for any 4.1.x, should we aim for this spec?

snehring: I've confirmed that the build error still exists with openmpi@4.1.1 in my environment. If you can test that 4.1.2 works for you I'd appreciate it. I'm not opposed to switching it to 4.1.1 and dealing with it on our end, since I suspect the issue only exists in pmix environments, which probably isn't everyone.

hseara: Let's do the following. I will remove the change of version from this PR and check 4.1.2. I will come back to the issue in a different PR if the problem persists.

snehring: sounds good to me.

snehring (approved):

tldahlgren (approved):
LGTM

PR #35490, opened 2023-02-15
adding new version to bacio, including handling of shared  library builds

edwardhartnett:
@AlexanderRichert-NOAA can you take a look at what I've done with shared library builds here and tell me if it's even close to correct? ;-)


var/spack/repos/builtin/packages/bacio/package.py
```diff
@@ -37,6 +40,11 @@ class Bacio(CMakePackage):
 
     def cmake_args(self):
         args = [self.define_from_variant("CMAKE_POSITION_INDEPENDENT_CODE", "pic")]
+        if self.spec.satisfies("@2.6.0:"):
+            if "+shared" in spec:
+                args.append("-DBUILD_SHARED_LIBS=ON")
+            else:
+                args.append("-DBUILD_SHARED_LIBS=OFF")
```
AlexanderRichert-NOAA: The preferred way to do this is `args.append(self.define_from_variant("BUILD_SHARED_LIBS", "shared"))`

var/spack/repos/builtin/packages/bacio/package.py
```diff
@@ -37,6 +40,11 @@ class Bacio(CMakePackage):
```
AlexanderRichert-NOAA, line 36: We need to declare a `shared` variant here. We can put the version restriction here as well. Something to the effect of `variant("shared", default=False, description="Build shared library", when="@2.6.0:")`. Spack is smart enough that down below when it runs `define_from_variant`, it will skip it if the condition doesn't match what's declared in the variant (i.e., it'll automatically skip setting `BUILD_SHARED_LIBS` if the version is older than 2.6.0).

AlexanderRichert-NOAA (changes requested):

var/spack/repos/builtin/packages/bacio/package.py
edwardhartnett: OK, done.

var/spack/repos/builtin/packages/bacio/package.py
edwardhartnett: OK, take a look...

AlexanderRichert-NOAA: I don't see any changes

edwardhartnett: OK, my push didn't work. Now I've got it, try again...

var/spack/repos/builtin/packages/bacio/package.py
```diff
@@ -36,7 +39,9 @@ class Bacio(CMakePackage):
     variant("pic", default=True, description="Build with position-independent-code")
 
     def cmake_args(self):
+        variant("shared", default=False, description="Build shared library", when="@2.6.0:")
```
AlexanderRichert-NOAA: The `variant` call should go right after the existing one (`variant("pic"...`), in the main body of the Bacio class definition.

edwardhartnett: OK, thanks!

edwardhartnett:
@AlexanderRichert-NOAA can you take another look and we'll see if I got this right?

AlexanderRichert-NOAA (approved):

alalazo (approved):

PR #35495, opened 2023-02-15
Halide: new variants and bug fixes

wraith1995:
This pull:

- Makes sure that Halide only looks for python if python is in the spec.
- Adds an option to build from a shared llvm, which is the llvm typically provided by system package managers. 
- Adds the option to choose among different output targets, as detailed in Halide's cmake documentation.
- Propagates these requirements to llvm.

In the future, I want to make this more robust by improving the llvm package's detection of the capabilities of system llvm.


wraith1995:
Assuming it passes all checks, the latest version satisfies my own testing.

adamjstewart (approved):

PR #35497, opened 2023-02-15
py-pyscf: add v2.2.0,2.1.1,2.1.0 libcint: add v5.2.0

prckent:
Note the most recent versions of the libxc library (6.1.0, 6.0.0) that py-pyscf depends on currently have broken download links.  Older versions appear to work fine, e.g.`spack install py-pyscf ^libxc@5.2.3` Current version dependencies are listed at https://pyscf.org/install.html#compiling-from-source-code

### DIFF
--- a/var/spack/repos/builtin/packages/py-pyscf/package.py
+++ b/var/spack/repos/builtin/packages/py-pyscf/package.py
@@ -11,23 +11,28 @@ class PyPyscf(PythonPackage):
     """PySCF is a collection of electronic structure programs powered
     by Python."""
 
-    homepage = "https://sunqm.github.io/pyscf/"
+    homepage = "https://pyscf.org"
     git = "https://github.com/pyscf/pyscf"
 
     maintainers("naromero77")
 
+    version("2.1.1", tag="v2.1.1")
+    version("2.1.0", tag="v2.1.0")
     version("2.0.1", tag="v2.0.1")
     version("1.7.5", tag="v1.7.5")
     version("1.7.3", tag="v1.7.3")
 
     # dependencies
+    depends_on("cmake@3.10:", type="build", when="@2.1:")
     depends_on("cmake@2.8:", type="build")
+    depends_on("python@3.6:", type=("build", "run"), when="@2.1:")
     depends_on("python@2.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-numpy@1.8.0:", type=("build", "run"))
     depends_on("py-numpy@1.13.0:", type=("build", "run"), when="@2:")
     conflicts("py-numpy@1.16:1.17", when="@2:")
     depends_on("py-scipy@0.12:", type=("build", "run"))
+    depends_on("py-scipy@0.19:", type=("build", "run"), when="@2.1:")
     conflicts("py-scipy@1.5.0:1.5.1", when="@2:")
     depends_on("py-h5py@2.3.0:", type=("build", "run"))
     depends_on("py-h5py@2.7.0:", type=("build", "run"), when="@2:")
--- a/var/spack/repos/builtin/packages/py-pyscf/package.py
+++ b/var/spack/repos/builtin/packages/py-pyscf/package.py
@@ -35,7 +35,7 @@ class PyPyscf(PythonPackage):
     depends_on("py-scipy@0.12:", type=("build", "run"))
     depends_on("py-scipy@0.19:", type=("build", "run"), when="@2.1:")
     depends_on("py-scipy@:1.1.0", when="^python@:3.7 platform=darwin", type=("build", "run"))
-    conflicts("py-scipy@1.5.0:1.5.1", when="@2:")
+    conflicts("^py-scipy@1.5.0:1.5.1", when="@2:")
     depends_on("py-h5py@2.3.0:", type=("build", "run"))
     depends_on("py-h5py@2.7.0:", type=("build", "run"), when="@2:")
     depends_on("blas")
--- a/var/spack/repos/builtin/packages/py-pyscf/package.py
+++ b/var/spack/repos/builtin/packages/py-pyscf/package.py
@@ -12,7 +12,7 @@ class PyPyscf(PythonPackage):
     by Python."""
 
     homepage = "https://pyscf.org"
-    url = "https://pypi.io/packages/source/p/pyscf/pyscf-2.2.0.tar.gz"
+    pypi = "pyscf/pyscf-2.2.0.tar.gz"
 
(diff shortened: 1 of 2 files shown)


tldahlgren (commented):
Confirmed the affected sha256 and tags as well as the presence of the new home page.

var/spack/repos/builtin/packages/py-pyscf/package.py
```diff
@@ -11,23 +11,28 @@ class PyPyscf(PythonPackage):
     """PySCF is a collection of electronic structure programs powered
     by Python."""
 
-    homepage = "https://sunqm.github.io/pyscf/"
+    homepage = "https://pyscf.org"
     git = "https://github.com/pyscf/pyscf"
```
adamjstewart: Would it be possible to download from PyPI instead? Then each release can be checksummed.

var/spack/repos/builtin/packages/py-pyscf/package.py
```diff
@@ -11,23 +11,28 @@ class PyPyscf(PythonPackage):
...
+    depends_on("cmake@3.10:", type="build", when="@2.1:")
     depends_on("cmake@2.8:", type="build")
+    depends_on("python@3.6:", type=("build", "run"), when="@2.1:")
     depends_on("python@2.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-numpy@1.8.0:", type=("build", "run"))
     depends_on("py-numpy@1.13.0:", type=("build", "run"), when="@2:")
     conflicts("py-numpy@1.16:1.17", when="@2:")
     depends_on("py-scipy@0.12:", type=("build", "run"))
+    depends_on("py-scipy@0.19:", type=("build", "run"), when="@2.1:")
```
adamjstewart, line 36: Where do you see these scipy version requirements?

adamjstewart: Should also be a requirement like:
```python
depends_on("py-scipy@:1.1.0", when="^python@:3.7", type=("build", "run"))
```

adamjstewart (commented):
Will give the maintainers a chance to comment

var/spack/repos/builtin/packages/py-pyscf/package.py
prckent: This package follows the"installation from source" instructions from the official documentation at https://pyscf.org/install.html#compiling-from-source-code . This is the best location I could find for up to date version requirements.
```
Prerequisites for manual install are

    CMake >= 3.10

    Python >= 3.6

    Numpy >= 1.13

    Scipy >= 0.19

    h5py >= 2.7
```

I do not understand your request to add a python version dependency to py-scipy usage. Where is this indicated in the pyscf docs? Is it flagged by spack CI?

var/spack/repos/builtin/packages/py-pyscf/package.py
prckent: Agreed that checksums would be preferred. Will investigate if the devs are consistent in updating PyPI and tackle next release if so.

var/spack/repos/builtin/packages/py-pyscf/package.py
adamjstewart: Documentation is meaningless to me and usually doesn't get updated very often. I check pyproject.toml/setup.py/setup.cfg which is what pip checks. Look at the `_scipy_version` stuff in `setup.py` to see what I mean.

That's not to say that the documentation isn't useful and we shouldn't keep the lower bound, just that if there's a bug in the docs, no one notices or reports an issue because most people pip install it.

tldahlgren (dismissed):

var/spack/repos/builtin/packages/py-pyscf/package.py
prckent: Updated to use pypi

var/spack/repos/builtin/packages/py-pyscf/package.py
prckent: Updated with requested requirement. This is darwin specific in setup.py so I added that dependency

prckent:
Added v2.2.0 and addressed all requests.

tldahlgren (dismissed):
Confirmed/re-confirmed all affected version sha256.

var/spack/repos/builtin/packages/py-pyscf/package.py
```diff
@@ -11,23 +11,30 @@ class PyPyscf(PythonPackage):
     """PySCF is a collection of electronic structure programs powered
     by Python."""
 
-    homepage = "https://sunqm.github.io/pyscf/"
-    git = "https://github.com/pyscf/pyscf"
+    homepage = "https://pyscf.org"
+    url = "https://pypi.io/packages/source/p/pyscf/pyscf-2.2.0.tar.gz"
```
adamjstewart:
```suggestion
    pypi = "pyscf/pyscf-2.2.0.tar.gz"
```

var/spack/repos/builtin/packages/py-pyscf/package.py
```diff
@@ -11,23 +11,30 @@ class PyPyscf(PythonPackage):
...
 
     # dependencies
+    depends_on("cmake@3.10:", type="build", when="@2.1:")
     depends_on("cmake@2.8:", type="build")
+    depends_on("python@3.6:", type=("build", "run"), when="@2.1:")
     depends_on("python@2.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-numpy@1.8.0:", type=("build", "run"))
     depends_on("py-numpy@1.13.0:", type=("build", "run"), when="@2:")
     conflicts("py-numpy@1.16:1.17", when="@2:")
```
adamjstewart:
```suggestion
    conflicts("^py-numpy@1.16:1.17", when="@2:")
```

var/spack/repos/builtin/packages/py-pyscf/package.py
```diff
@@ -11,23 +11,30 @@ class PyPyscf(PythonPackage):
...
     depends_on("py-scipy@0.12:", type=("build", "run"))
+    depends_on("py-scipy@0.19:", type=("build", "run"), when="@2.1:")
+    depends_on("py-scipy@:1.1.0", when="^python@:3.7 platform=darwin", type=("build", "run"))
     conflicts("py-scipy@1.5.0:1.5.1", when="@2:")
```
adamjstewart:
```suggestion
    conflicts("^py-scipy@1.5.0:1.5.1", when="@2:")
```

adamjstewart (approved):

Issue #35212, opened 2023-01-27
Spack fails to generate correct modulefiles for module-specified externals


koysean:
### Steps to reproduce

Set up a Spack instance with a dependency such as `mpich` provided by external modules in `packages.yaml`. Install a package with that dependency and generate the modulefiles:
```console
$ spack install hypre^mpich
[...]
==> mpich@50.1 : has external module in [mpich/50.1, libfabric/1.2.3, ...]
[+] /path/to/mpich (external mpich-50.1-yuxbwgffkpgydwdllhfoyrrk35x52h36)
==> Installing hypre-2.26.0-zot76ybciz2jjqxn74qscbvl6z6h6abf
[...]
$ spack module tcl refresh
```

The generated `hypre` modulefile in `$spack/share/spack/modules` will have the following lines:
```
if { ![ is-loaded 'mpich/50.1-yuxbwgf' ] } {
    module load mpich/50.1-yuxbwgf
}
```
which tries to load the spack-generated mpich module. That module, in turn, only contains the environment modifications that are implied by the inferred prefix and spack's package-specific environment modifications (the same root cause as https://github.com/spack/spack/issues/35168).

In short, the generated modules for externals completely ignore the modules listed in `packages.yaml` which provide the external itself.

Expected behavior would be for the `hypre` module to either directly load the external `mpich` modules or for the spack-generated `mpich` module to directly load the external modules specified in `packages.yaml`.

### Information on your system

* **Spack:** 0.20.0.dev0 (75f1077b4beaffffaaac8b18ef98585f76a983d8)
* **Python:** 3.6.15
* **Platform:** linux-sles15-icelake
* **Concretizer:** clingo

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

shahzebsiddiqui:
@koysean after several iterations of E4S deployment we just stop generating modulefiles via spack and just force users to adopt `spack load` due to the nature of how modulefiles can interact with external modulefiles, having to deal with module conflicts and custom suffix. 

Finally the explosion of modules caused lots of headache for user. Settting up the module configurations in spack can be tricky but it is doable.

koysean:
[...]

PR #35498, opened 2023-02-15
Pull in external module dependencies when writing modulefiles

koysean:
Fix for https://github.com/spack/spack/issues/35212

This PR modifies spack's module APIs to partially support multiple modulefiles for a single spec. These changes allow Spack to handle an external package with multiple modulefiles.

The primary API changes are in `spack.modules.common.get_module()` and `spack.modules.common.BaseFileLayout.use_name`. The return types were changed from `str` to `List[str]` to accomodate a single spec having multiple module names.

In addition, modulefiles are no longer created for external packages with modules, and modulefiles which depend on external modules will now directly load the external modules (if autoload is enabled).

One possible issue is that the module index will not be updated with external modules. This doesn't seem like a problem to me since Spack isn't managing the external modules anyways, but I'm not very familiar with how Spack uses the module index. Is this an issue?


lib/spack/spack/cmd/modules/__init__.py
```diff
@@ -169,15 +169,16 @@ def loads(module_type, specs, args, out=None):
...
-                module_type,
-                spec,
-                get_full_path=False,
-                module_set_name=args.module_set_name,
-                required=False,
-            ),
+            module,
         )
         for spec in specs
+        for module in spack.modules.common.get_module(
```
becker33: If this is going to return multiple modules, it should be renamed to `get_modules`.

lib/spack/spack/cmd/modules/__init__.py
```diff
@@ -169,15 +169,16 @@ def loads(module_type, specs, args, out=None):
...
+            module_type,
+            spec,
+            get_full_path=False,
+            module_set_name=args.module_set_name,
+            required=False,
+        )
```
becker33, line 179: Oh goodness this is hard to read... if the style checker is going to format it like this we might need to have a comment explaining what's going on here.

lib/spack/spack/modules/common.py
```diff
@@ -605,10 +614,13 @@ def dirname(self):
 
     @property
     def use_name(self):
-        """Returns the 'use' name of the module i.e. the name you have to type
-        to console to use it. This implementation fits the needs of most
-        non-hierarchical layouts.
+        """Returns a list of the 'use' names of the module i.e. the names you
```
becker33, line 621: Another method that should be pluralized with the changes you're making.

becker33 (changes requested):

lib/spack/spack/cmd/modules/__init__.py
koysean: Added a comment - the style checker also changed its mind (?) and wants the tuple (spec, module) on one line now, so it's a lot easier to read.

PR #35500, opened 2023-02-16
package_qgis_add_v328

Sinan81:
add new versions and an updated patch.

### DIFF
--- a/var/spack/repos/builtin/packages/qgis/package.py
+++ b/var/spack/repos/builtin/packages/qgis/package.py
@@ -17,14 +17,16 @@ class Qgis(CMakePackage):
 
     maintainers("adamjstewart", "Sinan81")
 
-    version("3.22.0", sha256="cf0c169863f332aab67d8c4943e14b73a564f0254bf54015f5826c6427e6785b")
-    version("3.18.2", sha256="1913e4d5596bbc8b7d143f3defb18bf376f750a71f334f69d76af5deca7ecc5d")
     # Prefer latest long term release
     version(
-        "3.16.12",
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/qgis/package.py
```diff
@@ -17,14 +17,16 @@ class Qgis(CMakePackage):
 
     maintainers("adamjstewart", "Sinan81")
 
-    version("3.22.0", sha256="cf0c169863f332aab67d8c4943e14b73a564f0254bf54015f5826c6427e6785b")
-    version("3.18.2", sha256="1913e4d5596bbc8b7d143f3defb18bf376f750a71f334f69d76af5deca7ecc5d")
     # Prefer latest long term release
```
adamjstewart, line 21: Isn't 3.22 the current LTS? Should we prefer that by default? Or should we get rid of this comment and the `preferred=True`?

Sinan81: according to the roadmap, the current LTR is 3.28
https://qgis.org/en/site/getinvolved/development/roadmap.html#location-of-prereleases-nightly-builds

Sinan81: confusingly enough 3.28 isn't listed under Long Term **repo** column.

Sinan81: Let me revert the preferred version to 3.22.16 until 3.28 becomes a long term "repo".

adamjstewart (approved):

PR #35501, opened 2023-02-16
Support pre-commit for styling

blue42u:
Currently the canonical way to format and lint any Spack changes is to run `spack style`. This PR adds a slightly more convenient alternative that runs during `git commit` via [`pre-commit`](https://pre-commit.com). Just run `pre-commit install` in your checkouts and never `spack style` again!

The included configuration tries to replicate the effect of `spack style --fix` as closely as possible (`pre-commit` by design is always in `--fix` mode). The command that runs all hooks on all files, `pre-commit run -a`, is added as part of the `prechecks` phase in CI.

This new framework exposes latent linter failures in the codebase. Most (but not all) of these are simply missing type annotations, so this PR also provides simple and unobtrusive fixes to pass pre-commit.

### DIFF
--- a/lib/spack/spack/builder.py
+++ b/lib/spack/spack/builder.py
@@ -7,7 +7,7 @@ import collections.abc
 import copy
 import functools
 import inspect
-from typing import List, Optional, Tuple
+from typing import TYPE_CHECKING, List, Optional, Tuple
 
 import spack.build_environment
 
@@ -581,5 +581,14 @@ class Builder(collections.abc.Sequence, metaclass=BuilderMeta):
 
 
 # Export these names as standalone to be used in packages
-run_after = PhaseCallbacksMeta.run_after
-run_before = PhaseCallbacksMeta.run_before
+if TYPE_CHECKING:
+
+    def run_after(phase, when=None):
+        return PhaseCallbacksMeta.run_after(phase, when=when)
+
+    def run_before(phase, when=None):
+        return PhaseCallbacksMeta.run_before(phase, when=when)
+
+else:
+    run_after = PhaseCallbacksMeta.run_after
+    run_before = PhaseCallbacksMeta.run_before
--- a/.github/workflows/valid-style.yml
+++ b/.github/workflows/valid-style.yml
@@ -53,6 +53,17 @@ jobs:
     - name: Run style tests
       run: |
           share/spack/qa/run-style-tests
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # @v2
+      with:
+        fetch-depth: 0
+    - uses: actions/setup-python@d27e3f3d7c64b4bbf8e4abfb9b63b83e846e0435 # @v2
+      with:
+        python-version: '3.11'
+        cache: 'pip'
+    - uses: pre-commit/action@646c83fcd040023954eafda54b4db0192ce70507 # @v3.0.0
   audit:
     uses: ./.github/workflows/audit.yaml
     with:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,69 @@
+default_language_version:
+  python: python3
+
+# Most hooks should not operate on the external/vendored code, to avoid
+# refactoring all of it to fit the trend.
+exclude: ^lib/spack/external/
+
+repos:
+
+# ------------------------------------------------------------------------------------------------
+#  Formatting hooks:  Hooks that alter the syntax but not semantics of files
+# ------------------------------------------------------------------------------------------------
+
+- repo: https://github.com/psf/black
+  rev: 23.1.0
+  hooks:
+  # Blacken all Python code
+  - id: black
+    args: [--config, pyproject.toml]
+
+- repo: https://github.com/PyCQA/isort
+  rev: 5.12.0
+  hooks:
+  # Sort imports in all Python code
+  - id: isort
+    alias: isort-lib
+    args: [--settings-path, pyproject.toml]
+    files: ^lib/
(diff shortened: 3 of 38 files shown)


var/spack/repos/builtin/packages/wgrib2/package.py
```diff
@@ -132,7 +132,13 @@ def inject_flags(self, name, flags):
 
         return (flags, None, None)
 
-    flag_handler = inject_flags
+    @property
```
becker33: `flag_handler` is a method, so I don't think this will work to propagate compiler flags into `wgrib2`.

lib/spack/spack/builder.py
```diff
@@ -581,5 +581,9 @@ def __str__(self):
 
 
 # Export these names as standalone to be used in packages
-run_after = PhaseCallbacksMeta.run_after
-run_before = PhaseCallbacksMeta.run_before
+def run_after(phase: str, when: Optional[str] = None):
```
becker33: Is there a good reason to replace a direct call to the `run_after` method with a wrapper around it?

becker33 (changes requested):

var/spack/repos/builtin/packages/wgrib2/package.py
blue42u: `flag_handler` is defined as a `@property` in `PackageBase` ([ref](https://github.com/spack/spack/blob/develop/lib/spack/spack/package_base.py#L2176-L2184)), so this is needed to bind `self` to the method before returning it as a callable object. Otherwise MyPy errors about `flag_handler` being assigned a callable with a different signature, which I presume is a valid error.

It probably should just be a method, but making it one triggered more MyPy errors in other packages, so :shrug: 

<details>
<summary>Specific errors when `flag_handler` is defined as a method</summary>

```
var/spack/repos/builtin/packages/qmcpack/package.py:216: error: Incompatible types in assignment (expression has type "Callable[[Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]", base class "PackageBase" defined the type as "Callable[[str, Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]")  [assignment]
var/spack/repos/builtin/packages/ipopt/package.py:60: error: Incompatible types in assignment (expression has type "Callable[[Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]", base class "PackageBase" defined the type as "Callable[[str, Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]")  [assignment]
var/spack/repos/builtin/packages/hpctoolkit/package.py:192: error: Incompatible types in assignment (expression has type "Callable[[Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]", base class "PackageBase" defined the type as "Callable[[str, Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]")  [assignment]
var/spack/repos/builtin/packages/hiop/package.py:130: error: Incompatible types in assignment (expression has type "Callable[[Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]", base class "PackageBase" defined the type as "Callable[[str, Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]")  [assignment]
var/spack/repos/builtin/packages/exago/package.py:141: error: Incompatible types in assignment (expression has type "Callable[[Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]", base class "PackageBase" defined the type as "Callable[[str, Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]")  [assignment]
var/spack/repos/builtin/packages/elfutils/package.py:104: error: Incompatible types in assignment (expression has type "Callable[[Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]", base class "PackageBase" defined the type as "Callable[[str, Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]")  [assignment]
var/spack/repos/builtin/packages/upcxx/package.py:104: error: Incompatible types in assignment (expression has type "Callable[[Iterable[str]], Any]", base class "PackageBase" defined the type as "Callable[[str, Iterable[str]], Tuple[Optional[Iterable[str]], Optional[Iterable[str]], Optional[Iterable[str]]]]")  [assignment]
```

</details>

lib/spack/spack/builder.py
blue42u: I'm using the wrapper function as a means to add type annotations, without them MyPy errors with:
```
lib/spack/spack/build_systems/_checks.py:128: error: Cannot determine type of "run_after"  [has-type]
```

The alternative is marking it with some long `Callable[...]` type. Would that be preferred?

blue42u:
Rebased onto `develop`, and split the MyPy fixes into multiple commits for (hopefully) easier review.

var/spack/repos/builtin/packages/wgrib2/package.py
blue42u: New patch lifts `flag_handler` and a few related `@classmethod`'s into proper methods (instead of using Python trickery for nearly the same result).

lib/spack/spack/builder.py
blue42u: New patch guards these wrapper functions under `if TYPE_CHECKING:`, so they don't cause any slowdown at runtime.

alecbcs:
@becker33

PR #35502, opened 2023-02-16
nonexistent module concise error message

simon-Leary42:
Before:
image

After:
image

In the before case, I didn't realize that the information I was looking for (the module location) was right in front of me, because it was buried under all that crap.

### DIFF
--- a/lib/spack/spack/modules/common.py
+++ b/lib/spack/spack/modules/common.py
@@ -405,7 +405,7 @@ def get_module(module_type, spec, get_full_path, module_set_name="default", requ
             fmt_str = "{name}{@version}{/hash:7}"
             if not writer.conf.excluded:
                 raise ModuleNotFoundError(
-                    "The module for package {} should be at {} but it does not exist!".format(
+                    "The module for package {} should be at {}, but it does not exist".format(
                         spec.format(fmt_str), writer.layout.filename
                     )
                 )
--- a/lib/spack/spack/modules/common.py
+++ b/lib/spack/spack/modules/common.py
@@ -403,9 +403,7 @@ def get_module(module_type, spec, get_full_path, module_set_name="default", requ
         writer = spack.modules.module_types[module_type](spec, module_set_name)
         if not os.path.isfile(writer.layout.filename):
             if not writer.conf.excluded:
-                err_msg = "No module available for package {0} at {1}".format(
-                    spec, writer.layout.filename
-                )
+                err_msg = f"The module for package {spec.name}@{spec.version}/{spec.dag_hash()[:10]} should be at {writer.layout.filename} but it does not exist!"
                 raise ModuleNotFoundError(err_msg)
             elif required:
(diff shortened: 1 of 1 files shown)


lib/spack/spack/modules/common.py
```diff
@@ -403,8 +403,8 @@ def get_module(module_type, spec, get_full_path, module_set_name="default", requ
         writer = spack.modules.module_types[module_type](spec, module_set_name)
         if not os.path.isfile(writer.layout.filename):
             if not writer.conf.excluded:
-                err_msg = "No module available for package {0} at {1}".format(
-                    spec, writer.layout.filename
+                err_msg = "The module for package /{0} should be at {1} but it does not exist!".format(
+                    spec.dag_hash(), writer.layout.filename
```
alalazo: Can we maintain the spec name and version along with the hash? That shouldn't result in an extremely long message, and would improve readability.

alalazo: Also, now we're allowed to use f-strings, since Spack supports running Python 3.6+

alalazo (changes requested):

lib/spack/spack/modules/common.py
simon-Leary42: Done, assuming the correct syntax is `spec.name` and `spec.version`

simon-Leary42:
@spackbot fix style

lib/spack/spack/modules/common.py
simon-Leary42: Tested:
```
==> Error: The module for package gcc@9.4.0/occi2durd5 should be at /modules/spack-test-hierarchy/share/spack/modules/linux-ubuntu20.04-x86_64/gcc-9.4.0-gcc-9.4.0-occi2du but it does not exist!
```

simon-Leary42: Although that module name looks wrong, I think that's unrelated to this PR

haampie:
you have to fix the line length issue by hand

alalazo:
Still style issues (line-length)

haampie (dismissed):

lib/spack/spack/modules/common.py
```diff
@@ -402,13 +402,19 @@ def get_module(module_type, spec, get_full_path, module_set_name="default", requ
     else:
         writer = spack.modules.module_types[module_type](spec, module_set_name)
         if not os.path.isfile(writer.layout.filename):
+            fmt_str = "{name}{@version}{/hash:7}"
             if not writer.conf.excluded:
-                err_msg = "No module available for package {0} at {1}".format(
-                    spec, writer.layout.filename
+                raise ModuleNotFoundError(
+                    "The module for package {} should be at {} but it does not exist!".format(
```
haampie:
```suggestion
                    "The module for package {} should be at {}, but it does not exist".format(
```

haampie (approved):

alalazo (approved):

PR #35504, opened 2023-02-16
QE: Adding support for AOCC 4.0 and zen4

amd-toolchain-support:
This pull request adds support for AOCC compiler for QE 7.1 and fixes compatibility between AOCC and QE 7.0

### DIFF
--- a/var/spack/repos/builtin/packages/quantum-espresso/configure_qe-7.1_aocc.patch
+++ b/var/spack/repos/builtin/packages/quantum-espresso/configure_qe-7.1_aocc.patch
@@ -1,0 +1,76 @@
+diff --git a/install/configure b/install/configure_aocc
+index c8f62eb..bca4e4d 100755
+--- a/install/configure
++++ b/install/configure_aocc
+@@ -3183,6 +3183,7 @@ echo $ECHO_N "checking version of $mpif90... $ECHO_C"
+ ifort_version=`$mpif90 -V 2>&1 | grep "Intel(R)"`
+ pgf_version=`$mpif90 -V 2>&1 | grep "^pgf"`
+ nvfortran_version=`$mpif90 -V 2>&1 | grep "^nvfortran"`
++aoccflang_version=`$mpif90 -v 2>&1 | grep "AMD clang version"`
+ gfortran_version=`$mpif90 -v 2>&1 | grep "gcc version"`
+ nagfor_version=`$mpif90 -v 2>&1 | grep "NAG Fortran"`
+ xlf_version=`$mpif90 -v 2>&1 | grep "xlf"`
+@@ -3203,6 +3204,12 @@ then
+ 	f90_minor_version=`echo $version | cut -d. -f2 | cut -d- -f1`
+         echo "${ECHO_T}nvfortran $version"
+         f90_in_mpif90="nvfortran"
++elif test "$aoccflang_version" != ""
++then
++        version=`echo $aoccflang_version | cut -d" " -f 5`
++        echo "${ECHO_T}mpif90 $version"
++        f90_in_mpif90="mpif90"
++        try_foxflags="-D__PGI"	
+ elif test "$pgf_version" != ""
+ then
+         version=`echo $pgf_version | cut -d ' ' -f2`
+@@ -3374,6 +3381,9 @@ ppc64-bg*:*xlf90_r )
+ ppc64-bg*:*xlf90 )
+         try_cc="bgxlc"
+         ;;
++*:mpif90 )
++        try_cc="mpicc $try_cc"
++        ;;	
+ ppc64:*xlf* | ppc64le:*xlf* )
+         try_cc="xlc_r $try_cc"
+         ;;
+@@ -3763,11 +3773,14 @@ ppc64-bg:* )
+ ppc64-bgq:* )
+         try_cflags="-O3"
+         ;;
++x86_64:* )
++        try_cflags="-Ofast -Mstack_arrays"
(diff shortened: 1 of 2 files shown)


var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -367,7 +367,8 @@ class QuantumEspresso(CMakePackage, Package):
     )
 
     # Configure updated to work with AOCC compilers
-    patch("configure_aocc.patch", when="@6.7:6.8 %aocc")
+    patch("configure_aocc.patch", when="@6.7:7.0 %aocc")
+    patch("configure_qe-7.1_aocc.patch", when="@7.1: %aocc")
```
alalazo, line 371: Did you upstream this patch?

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -590,11 +591,19 @@ def install(self, pkg, spec, prefix):
                 zlib_libs = spec["zlib"].prefix.lib + " -lz"
                 filter_file(zlib_libs, format(spec["zlib"].libs.ld_flags), make_inc)
 
+        # Filer file must be applied for AOCC due to incorrect make.inc generation
+        if spec.satisfies("%aocc") and spec.satisfies("@6.8:"):
```
alalazo:
```suggestion
        if spec.satisfies("@6.8: %aocc"):
```

alalazo (commented):
Just a couple of minor questions. Waiting for maintainers review before merging.

var/spack/repos/builtin/packages/quantum-espresso/package.py
amd-toolchain-support: No, this patch has not been upstreamed to QE codebase yet.

alalazo: Can you please submit it, so we don't have to rely on patches on the next version?

var/spack/repos/builtin/packages/quantum-espresso/configure_qe-7.1_aocc.patch
```diff
@@ -0,0 +1,76 @@
...
+ ppc64:*xlf* | ppc64le:*xlf* )
+         try_cc="xlc_r $try_cc"
+         ;;
+@@ -3763,11 +3773,14 @@ ppc64-bg:* )
+ ppc64-bgq:* )
+         try_cflags="-O3"
+         ;;
++x86_64:* )
++        try_cflags="-Ofast -Mstack_arrays"
++        try_dflags="-D__OPENMP"
```
ye-luo, line 42: OpenMP is optional. It should not be hard coded.

var/spack/repos/builtin/packages/quantum-espresso/package.py
```diff
@@ -590,11 +591,19 @@ def install(self, pkg, spec, prefix):
...
+            filter_file("FOX_FLAGS = -D__PGI", "FOX_FLAGS = -D__PGI -cpp", make_inc)
+
         # QE 6.8 and later has parallel builds fixed
         if spec.satisfies("@:6.7"):
             parallel_build_on = False
         else:
             parallel_build_on = True
+        # Parallel build fails with 7.1
+        if spec.satisfies("@7.1"):
+            parallel_build_on = False
```
ye-luo, line 606: Why this needs to be special. Are you saying parallel build is broken with any compiler?

ye-luo (changes requested):
The necessary patch needs to proof itself by being accepted by upstream q-e at gitlab. Then we can keep a backport here for 7.1.

Q: does the default CMake work with AOCC 4.0? Older AOCC worked when I manually build QE via CMake. Wondering if AOCC 4.0 changes anything.

prckent:
As a QE and spack user I am happy to see this work on getting optimized AMD support - QE has a large userbase. Echoing ye-luo's question, since the preferred QE configuration/installation route is clearly now CMake, are any updates needed there or does the CMake route already work optimally with AMD compilers and libraries?

ye-luo:
I tried a manual build out side spack with cmake. I don't have MPI enabled.
```
cmake -DCMAKE_C_COMPILER=/soft/AMD/aocc-compiler-4.0.0/bin/clang -DCMAKE_Fortran_COMPILER=/soft/AMD/aocc-compiler-4.0.0/bin/flang -DCMAKE_PREFIX_PATH=/soft/AMD/aocl-linux-aocc/4.0/lib -DQE_ENABLE_MPI=OFF -DQE_ENABLE_OPENMP=ON -DFFTW3_ROOT=/soft/AMD/aocl-linux-aocc/4.0 ../q-e-repo
```
I noticed that OpenMP is required to be on because libflame provided by AOCL requires it.
So with cmake there is no patching needed on QE source code.

haampie:
This is yet another case where AMD's magical compiler performance comes from ... `-Ofast`.

`-Ofast` implies `-ffast-math` (at least on clang, on which aocc is based), and `-ffast-math` enables flush to zero, which can influence correctness of other, unrelated programs, because it's a toggle on the CPU (at least on Intel CPUs): https://simonbyrne.github.io/notes/fastmath/#flushing_subnormals_to_zero

Are AMD folks aware of this? Is this intended?

See https://simonbyrne.github.io/notes/fastmath/#flushing_subnormals_to_zero for example.

amd-toolchain-support:
This patch was internal and should not have made it into the queue for publication.  Our apologies for the noise.  A cleanup of our existing patching will be incoming as a new MR.

PR #35510, opened 2023-02-16
pfunit: add v4.6.3, clean up recipe

alalazo:
Modifications:
- [x] Add v4.6.3
- [x] Deprecate v3.X
- [x] Removed `shared` variant (it seems only static builds are officially supported)
- [x] Fixed a few inconsistencies in argument names and values
- [x] Don't set unneeded environment variables

### DIFF
--- a/var/spack/repos/builtin/packages/pfunit/package.py
+++ b/var/spack/repos/builtin/packages/pfunit/package.py
@@ -39,16 +39,16 @@ class Pfunit(CMakePackage):
     version("4.1.12", sha256="7d71b0fb996497fe9a20eb818d02d596cd0d3cded1033a89a9081fbd925c68f2")
     version("4.1.11", sha256="16160bac223aaa3ed2b27e30287d25fdaec3cf6f2c570ebd8d61196e6aa6180f")
     version("4.1.10", sha256="051c35ad9678002943f4a4f2ab532a6b44de86ca414751616f93e69f393f5373")
-    version("3.3.3", sha256="9f673b58d20ad23148040a100227b4f876458a9d9aee0f0d84a5f0eef209ced5")
-    version("3.3.2", sha256="b1cc2e109ba602ea71bccefaa3c4a06e7ab1330db9ce6c08db89cfde497b8ab8")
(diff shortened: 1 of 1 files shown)


alalazo:
@spackbot maintainers

mathomp4:
@alalazo This is a bit of a change! I'll try and test it out soon

alalazo:
Pipelines are failing on `develop` due to `pfunit@3.3` which is needed by `flecsi@1.4.2`, see:
- https://gitlab.spack.io/spack/spack/-/jobs/5540096

This is not appearing here because, since deprecating `pfunit@:3`  drives `clingo` to select a newer `flecsi@2.1.0` which doesn't depend on `pfunit`. Before `flecsi@1.4.2` was selected, since 1.4.2 is the preferred version.

@eugeneswalker @rspavel @ktsai7 @rbberger Is that a viable solution? Asking since `flecsi` seems quite a peculiar package, where the preferred version comes from a fork that is a couple of years behind the main repository.

eugeneswalker:
> This is not appearing here because, since deprecating `pfunit@:3` drives `clingo` to select a newer `flecsi@2.1.0` which doesn't depend on `pfunit`. Before `flecsi@1.4.2` was selected, since 1.4.2 is the preferred version.
> 
> @eugeneswalker @rspavel @ktsai7 @rbberger Is that a viable solution? Asking since `flecsi` seems quite a peculiar package, where the preferred version comes from a fork that is a couple of years behind the main repository.

I don't know much about this package, other than we need to include some version of it in E4S. Whichever version that is or should be, I have to defer to those with more knowledge about how it is used. I'm fine with whatever works. But also, we don't want this to be holding up PRs, so if using a newer version makes things work here, I like that idea.

haampie:
I'm gonna merge since it blocks other PRs, if someones doesn't agree feel free to follow up with a different solution.

haampie (approved):

alalazo:
@mathomp4 Feel free to post here any follow up changes you'd need. I'll try to submit a PR for them, in case, asap.

PR #35514, opened 2023-02-16
Enforce carriage return on Windows bat files

johnwparent:
Batch scripts in general will not function without carriage return line endings on Windows. We rely on these scripts to support cmd, so we should not allow these scripts to be converted to lf. Note: Windows 11 supports lf line endings due to the use of Windows Terminal. Once support for Windows 10 is dropped, this change can be reverted.


scheibelp (approved):

johnwparent:
@scheibelp my development workflow thanks you for merging this so quickly!

PR #35517, opened 2023-02-16
increase db timeout 3s -> 60s

haampie:
When running many concurrent spack install processes that need to write
to the db, Spack regularly times out. This is because writing to the DB
after another process has written to it requires deserialization of the
db, mutating it in memory, and serializing it again, which takes some
time. On top of that, I believe there's a 1 second retry when a write
lock cannot be obtained, so I think this means only 3 processes can
really write to the DB at the same time before timing out.

This change should help particularly with `spack env depfile` in
environments with many leaf nodes.

(Ping @iarspider)

### DIFF
--- a/etc/spack/defaults/config.yaml
+++ b/etc/spack/defaults/config.yaml
@@ -181,7 +181,7 @@ config:
   # when Spack needs to manage its own package metadata and all operations are
   # expected to complete within the default time limit. The timeout should
   # therefore generally be left untouched.
-  db_lock_timeout: 3
+  db_lock_timeout: 60
 
 
   # How long to wait when attempting to modify a package (e.g. to install it).


alalazo (approved):

iarspider:
Thanks @haampie , will try this.

tgamblin:
@haampie:
> This is because writing to the DB after another process has written to it requires deserialization of the db, mutating it in memory, and serializing it again, which takes some time.

One thing that we could do here is, for most reads, don't even hold the lock.  Right now I think we do this:

```python
with spack.store.db.read_transaction():    # reads the DB
    # do something
# lock is released after `with` block
```

That means the lock is held while:
1. DB JSON is read
2. DB JSON is converted to specs
3. whatever is in the with block

Most reads don't need all of that -- technically they should only need to hold the lock for (1) unless there's going to be a nested write transaction somewhere in the with block.

I wonder if we have a lot of places where we're holding read locks for too long.

haampie:
Yeah, probably we can optimize read transactions, but at the end of the day, when something has to be written to the db, we need `with write_transaction`, and when multiple processes write to the db, it causes a reread when the lock is acquired, it needs to construct the entire DAG in Spec objects, has to apply the change, and dump to JSON. I don't think this can be avoided.

(However, the real issue may be more trivial, cause as @iarspider reported, running `spack env depfile -o Makefile` + `make -j8` on a fresh db reliably caused timeouts. So, it's more likely that the 3 seconds db timeout with 1 second retry to acquire a write lock is generally too strict; might already fail at 3 concurrent processes -- ugh, I don't recall how I got to this 1 second, but I think it was from --debug output.) 

Edit: I think the 1 second is made up 😬 we only have some custom long retry delay when waiting for a prefix write lock, for db locks it seems frequent polling. In any case, increasing the timeout helps 😆 it's been there for a while in Gitlab CI too.

PR #35520, opened 2023-02-16
spdk: add a new package

hyoklee:
The Storage Performance Development Kit (SPDK) provides a set of tools
and libraries for writing high performance, scalable, user-mode storage
applications.

PR #35522, opened 2023-02-16
fio: add a new version

hyoklee:
The latest is 3.33. 
See also #35520.


tldahlgren (approved):
Confirmed the sha256.

Issue #34370, opened 2022-12-07
Setting `cxxflags="stdlib=libc++"` results in `Error: {'stdlib=libc++'} does not equal set()` after #34218


msimberg:
### Steps to reproduce

The `cxxflags` variant doesn't seem to accept certain types of flags:
```console
> spack spec cmake cxxflags="-stdlib=libc++"
==> Error: {'-stdlib=libc++'} does not equal set()
```
Curiously, e.g.
```console
> spack spec cmake cxxflags="-stdlib=libc+"
```
or

```console
> spack spec cmake cxxflags="-stdlib=libc+++"
```
or
```console
> spack spec cmake cxxflags="-O3"
```
do concretize (is `-stdlib=libc++` being filtered out at some point?).

### Error message

```console
> spack -d spec -lI cmake cxxflags="-stdlib=libc++"
==> [2022-12-07-13:30:07.174827] Imported spec from built-in commands
==> [2022-12-07-13:30:07.175544] Imported spec from built-in commands
==> [2022-12-07-13:30:07.183557] Reading config from file /apps/daint/SSL/pika/spack/etc/spack/defaults/config.yaml
==> [2022-12-07-13:30:07.210096] Reading config from file /apps/daint/SSL/pika/spack/etc/spack/config.yaml
==> [2022-12-07-13:30:07.215239] DATABASE LOCK TIMEOUT: 3s
==> [2022-12-07-13:30:07.215334] PACKAGE LOCK TIMEOUT: No timeout
==> [2022-12-07-13:30:07.216063] Reading config from file /apps/daint/SSL/pika/spack/etc/spack/defaults/bootstrap.yaml
==> [2022-12-07-13:30:07.228291] Reading config from file /apps/daint/SSL/pika/spack/etc/spack/defaults/repos.yaml
==> [2022-12-07-13:30:07.233727] Reading config from file /apps/daint/SSL/pika/spack/etc/spack/defaults/config.yaml
==> [2022-12-07-13:30:07.256862] Reading config from file /apps/daint/SSL/pika/spack/etc/spack/config.yaml
==> [2022-12-07-13:30:07.258273] DATABASE LOCK TIMEOUT: 3s
==> [2022-12-07-13:30:07.258328] PACKAGE LOCK TIMEOUT: No timeout
==> [2022-12-07-13:30:07.258419] [BOOTSTRAP CONFIG SCOPE] name=_builtin
[...]

PR #35524, opened 2023-02-16
add PIC flag in mpi-serial build

jedwards4b:
I ran into a [known problem](https://github.com/spack/spack/issues/34370) with building mpi-serial and found that this change solves the problem.


tldahlgren (approved):
LGTM

PR #35525, opened 2023-02-16
py-pyinstrument: add v4.4.0

adamjstewart:


tldahlgren (approved):
Confirmed the sha256.  Didn't know about `skip_modules`, cool.

PR #28783, opened 2022-02-04
Use Spec.constrain to construct spec lists for stacks

alalazo:
fixes #28749 

Before this PR the construction of a spec list for software stacks was based on string concatenation rather than constraint merging. This may cause issues like #28749 if a line in the matrix contains constrains on both the root spec and a dependency. This PR fixes the issue by using `Spec.constrain` instead of string concatenation to construct the spec list.

Nested matrices and temporarily converted back to string to allow for a sigil to be applied in front of them when they are expanded as `$^` or as `$%`.

Modifications:

- [x] Add a regression test based on #28749
- [x] Fix a bug in `Spec.constrain` that wasn't adding a name or a namespace to an anonymous spec if constrained with a named spec
- [x] Use `Spec.constrain` to merge the constraints for spec matrices

PR #35529, opened 2023-02-17
Spec matrices: preserve ordering of constraints

becker33:
Currently it is not possible to use `concretizer:unify:when_possible` configuration to minimize the dependencies among a series of packages built with a combination of MPI/compiler combinations. The matrix:

```
- matrix:
  - [foo]
  - [^mpich]
  - [ '%gcc', '%clang']
```

resolves to specs `foo%gcc ^mpich` and `foo%clang ^mpich`.

With the `when_possible` unification option, the concretizer will concretize to use a single `mpich` in both cases (because it is possible to do so), mixing compilers rather than building with each combination of compiler/mpi.

With this PR, the previous behavior is preserved for the matrix:

```
- matrix:
  - [foo]
  - [ '%gcc', '%clang']
  - [^mpich]
```

but the original matrix produces the specs `foo ^mpich%gcc` and `foo ^mpich%clang`. Spack's preference to avoid mixing compilers when not otherwise instructed will then ensure that `foo` is built with both MPI/compiler combinations.

Includes docs and test case

@tgamblin this is the PR we discussed with @nicholas-sly on Tuesday.

### DIFF
--- a/lib/spack/docs/environments.rst
+++ b/lib/spack/docs/environments.rst
@@ -346,7 +346,7 @@ the Environment and then install the concretized specs.
    (see :ref:`build-jobs`). To speed up environment builds further, independent
    packages can be installed in parallel by launching more Spack instances. For
    example, the following will build at most four packages in parallel using
-   three background jobs: 
+   three background jobs:
 
    .. code-block:: console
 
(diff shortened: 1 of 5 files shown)


alalazo:
For reference, https://github.com/spack/spack/pull/28783 was going in the opposite direction (use `Spec.constrain` instead of string concatenation). Would it be cumbersome to:

1. First define a list of MPIs in a matrix
2. Reuse the MPI matrix with applications

?

alalazo:
Looking at docs [here](https://spack.readthedocs.io/en/latest/environments.html#spec-list-references), what I am suggesting looks like:
```yaml
spack:
  definitions:
    - compilers: ['%gcc', '%clang']
    - mpis: 
      -  matrix:
        - [mpich]
        - [$compilers]

  specs:
  - matrix:
    - [foo]
    - [$^mpis]
```
I didn't check if it works out of the box, but hopefully this doesn't require to pass to string concatenation (which changes the semantics of applying constraints).

becker33:
Closing in favor of #35663

PR #35532, opened 2023-02-17
py-amici, py-python-libsbml: new packages

heerener:


var/spack/repos/builtin/packages/py-amici/package.py
```diff
@@ -0,0 +1,45 @@
...
+    homepage = "https://github.com/AMICI-dev/AMICI"
+    pypi = "amici/amici-0.11.28.tar.gz"
+
+    version("0.16.0", sha256="1a2d6633ec34241d8d8b496d18d4318482cffe125e9ddf3ca6cac5d36d235f38")
+    version("0.11.28", sha256="a8ddda70d8ebdc40600b4ad2ea02eb26e765ca0e594b957f61866b8c84255d5b")
+
+    variant("boost", default=False, description="Enable boost support")
+    variant("hdf5", default=False, description="Enable HDF5 support")
+
+    depends_on("py-setuptools@48:", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-setuptools@48:", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-amici/package.py
```diff
@@ -0,0 +1,45 @@
...
+
+    depends_on("blas", type=("build", "run"))
+    depends_on("boost", when="+boost", type=("build", "run"))
+    depends_on("hdf5", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("hdf5", when="+hdf5", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-python-libsbml/package.py
```diff
@@ -0,0 +1,23 @@
...
+    git = "https://github.com/sbmlteam/python-libsbml.git"
+    pypi = "python-libsbml/python-libsbml-5.19.7.tar.gz"
+
+    version("5.19.7", sha256="447b1fde7aceccd11a93dc9f589ffd9319ba854d7b7583f911259a8b0127ab7b")
+    version("5.19.5", tag="v5.19.5", submodules=True)
+
+    depends_on("py-setuptools", type="build")
+
+    depends_on("swig", type=("build", "run"))
+    depends_on("cmake", type=("build", "run"))
```
adamjstewart: These are rarely needed at run-time. Where did you find them listed?

heerener: https://github.com/sbmlteam/python-libsbml#preparing-the-package-build
I'm never sure whether something like this means it should or shouldn't be there at runtime, but I can remove the "run" type.

var/spack/repos/builtin/packages/py-amici/package.py
heerener: So I know what to look out for in the future: why is it a run dependency as well this time?

adamjstewart: https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#setuptools

It's listed under `install_requires`, and is imported in the code that gets installed.

var/spack/repos/builtin/packages/py-python-libsbml/package.py
adamjstewart: I think those are only needed to build, you can try using it and see if anything breaks if these aren't available in PATH if you want to confirm.

heerener: ~~I did a small experiment and things still seemed to work. I'm not exactly familiar with the software, but I'm sure if I'm wrong someone will open an issue. Feel free to tag me if that happens.~~

@matz-e is telling me I might need cmake after all... checking.

matz-e: AMICI itself needs `cmake` at runtime, as it generates a C++ model from SBML. Then it runs `setup.py` and BLAS/SWIG/HDF5/CMake to build a Python module from the C++ code.

var/spack/repos/builtin/packages/py-amici/package.py
```diff
@@ -0,0 +1,45 @@
...
+class PyAmici(PythonPackage):
+    """Advanced Multilanguage Interface for CVODES and IDAS"""
+
+    homepage = "https://github.com/AMICI-dev/AMICI"
+    pypi = "amici/amici-0.11.28.tar.gz"
+
+    version("0.16.0", sha256="1a2d6633ec34241d8d8b496d18d4318482cffe125e9ddf3ca6cac5d36d235f38")
+    version("0.11.28", sha256="a8ddda70d8ebdc40600b4ad2ea02eb26e765ca0e594b957f61866b8c84255d5b")
+
+    variant("boost", default=False, description="Enable boost support")
```
heerener: Would it be acceptable to have these two variants enabled by default?

adamjstewart: I personally prefer to default to off (matches pip behavior, faster build times for the average user) but I'll let you decide.

adamjstewart (approved):

PR #35534, opened 2023-02-17
environment.matching_spec: linear time traversal

haampie:
... and use colors in disambiguate message for clarity.

This commit avoids the worst-case O(v^2) loop

```python
for root in roots:
  for dep in deps(root):
    ...
```

instead it ensures it's O(v + e).

Also adds a small optimization when searching for concrete specs, since
we can assume uniqueness of dag hash, so it's fine to early exit.

The order of traversal is breadth first over the roots, so that we're
guaranteed to find matching specs as roots first, instead of as deps of
other roots.

### DIFF
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -1856,23 +1856,24 @@ class Environment(object):
         and multiple dependency specs match, then this raises an error
         and reports all matching specs.
         """
-        roots = [root for _, root in self.concretized_specs()]
+        env_roots = [root for _, root in self.concretized_specs()]
         by_hash = lambda s: s.dag_hash()
-        root_matches = []
-        dep_matches = []
+        root_matches, dep_matches = [], []
 
         for i, env_spec in enumerate(
-            spack.traverse.traverse_nodes(specs=roots, key=by_hash, order="breadth")
+            spack.traverse.traverse_nodes(specs=env_roots, key=by_hash, order="breadth")
         ):
             if not env_spec.satisfies(spec):
                 continue
 
-            # Early exit for concrete specs, exploit uniqueness of dag hash.
+            # If the spec is concrete, then there is no possibility of multiple matches,
+            # and we immediately return the single match
             if spec.concrete:
                 return env_spec
 
-            # Distinguish between matching roots and deps of roots.
-            if i < len(self.concretized_user_specs):
+            # Distinguish between environment roots and deps. Specs that are both
+            # are classified as environment roots.
+            if i < len(env_roots):
                 root_matches.append((env_spec, self.concretized_user_specs[i]))
             else:
                 dep_matches.append(env_spec)
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -1856,12 +1856,13 @@ class Environment(object):
         and multiple dependency specs match, then this raises an error
         and reports all matching specs.
         """
-        env_roots = [root for _, root in self.concretized_specs()]
-        by_hash = lambda s: s.dag_hash()
+        env_root_to_user = {root.dag_hash(): user for user, root in self.concretized_specs()}
         root_matches, dep_matches = [], []
 
-        for i, env_spec in enumerate(
-            spack.traverse.traverse_nodes(specs=env_roots, key=by_hash, order="breadth")
+        for env_spec in spack.traverse.traverse_nodes(
+            specs=[root for _, root in self.concretized_specs()],
+            key=lambda s: s.dag_hash(),
+            order="breadth",
         ):
             if not env_spec.satisfies(spec):
                 continue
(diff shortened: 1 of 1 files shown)


lib/spack/spack/environment/environment.py
```diff
@@ -1855,59 +1856,57 @@ def matching_spec(self, spec):
...
 
-            if concretized_user_spec.satisfies(spec):
-                matches[concretized_user_spec] = user_spec
-            for dep_spec in concretized_user_spec.traverse(root=False):
-                if dep_spec.satisfies(spec):
-                    # Don't overwrite the abstract spec if present
-                    # If not present already, set to None
-                    matches[dep_spec] = matches.get(dep_spec, None)
+            # Distinguish between matching roots and deps of roots.
+            if i < len(self.concretized_user_specs):
```
scheibelp: If some roots are deps of other roots, I'm not sure if this check will do what you want. For example if you have:

```
X
|\
Y |
|/
Z
```

and X, Z are roots (i.e. user specs) but Y is not (in that sense, calling them roots is potentially confusing)

Either way, I think it is more straightforward to say:

`if env_spec in set(roots)`

which should be O(1) (make the set before the loop)

Likewise for the subsequent:

`root_matches.append((env_spec, self.concretized_user_specs[i]))`

lib/spack/spack/environment/environment.py
```diff
@@ -1855,59 +1856,57 @@ def matching_spec(self, spec):
...
-            spec = spack.spec.Spec(spec)
+        for i, env_spec in enumerate(
+            spack.traverse.traverse_nodes(specs=roots, key=by_hash, order="breadth")
+        ):
+            if not env_spec.satisfies(spec):
+                continue
 
-        for user_spec, concretized_user_spec in self.concretized_specs():
-            # Deal with concrete specs differently
+            # Early exit for concrete specs, exploit uniqueness of dag hash.
```
scheibelp: Could this comment be expanded, something like:

> If the spec is concrete, then there is no possibility of multiple matches, and we can return the single match

seems clearer to me

scheibelp (changes requested):
I have one concern about the logic for identifying specs as being roots, and a request about clarifying a comment.

lib/spack/spack/environment/environment.py
haampie: It does what I want because of iteration order, but yeah, it can be improved. Is hash(spec) and spec_a == spec_b O(1). I never verified that. If it's by object id then yes, otherwise I hope it uses dag hash for both the hash key and the equality check, but I never verified that.

haampie: I checked: calling `set(roots)` triggers a process_hash to be calculated on the environment roots, which triggers traversal including build/test deps and computes package hash. I think I'm happy with the current implementation and don't see the point of yet another DAG traversal.

scheibelp: > calling set(roots) triggers a process_hash to be calculated on the environment roots, which triggers traversal including build/test deps and computes package hash. 

Yes: this is needed to execute `__hash__` on a `Spec`. This shouldn't incur `O(n^2)` operations though so I don't see how the request conflicts with the PR's stated intent.

If it were clear to me that the current strategy worked, I would be fine with it staying as it is, but I don't see how a breadth-first traversal is guaranteed to avoid issues with the example I listed above (in particular, there are valid breadth-first orderings that run into problems), nor anything in the documentation of the `traverse.traverse_nodes` function which guarantees that the roots are visited first (and I assume it couldn't make any such guarantees anyway when complying with a requested ordering).

haampie: I don't get why it's not clear to you. Breadth first traversal is queue based, and the queue is initialized with the environment roots, newly discovered nodes get pushed to the back. 😕 there is no need to keep track of a set of roots. The bfs behavior with multiple sources is natural to me and there are unit tests covering it

scheibelp: > Breadth first traversal is queue based, and the queue is initialized with the environment roots, newly discovered nodes get pushed to the back. 

Put that in the docstring for `traverse_nodes` (actually not exactly that, but that the roots are guaranteed to be visited first): you are saying you know this works because of underlying, undocumented details of these functions. Whoever reads this next should find it easy to reason about, they cannot (based simply on a reading of the associated function docstrings) know that this strategy would work.

> The bfs behavior with multiple sources is natural to me

Not to me, and in particular not when some nodes are children of others (which I've suggested prohibiting before). None of the tests in `test/traverse.py` explicitly state testing this aspect of behavior. `test_breadth_first_traversal_multiple_input_specs` appears to touch that - if not thoroughly - but I had to read through the entire module including all the code in each test to figure that out, and ideally someone reading this in the future doesn't have to do that.

I'd like to state on top of all of this that if there are two approaches that work, and one of them is clearer (and I have argued in this case that it is not more expensive in a big-O sense), then why not use that? Reasoning about traversals by inferring properties of elements based on their index is confusing.

haampie: Ok, I've changed it to a dict from keyed by dag hash.

I realized it's possible to have repeated concrete environments roots (for example `specs: [python, python@3]` with `unify: true`), so indexing can be wrong. I left `order="breadth"` cause it's faster than recursion in Python.

haampie:
ping @scheibelp

scheibelp (approved):

scheibelp:
Thanks!

PR #35535, opened 2023-02-17
snakemake: added versions 7.19.0-7.22.0

prs513rosewood:
Also added the corresponding dependencies. Cheers!


adamjstewart (commented):
The dependencies for the `+reports` variant have changed and need to be updated

adamjstewart (approved):

PR #35552, opened 2023-02-17
cmake build system: make "generator" a variant

alalazo:
This PR make `generator` a variant in `CMakePackage`. This allows us to declare proper dependencies on `ninja` and `gmake`, and to select generators at concretization time where it makes sense.

Modifications:
- [x] Add a cmake specific generator directive, that builds on `variants` and `conflicts`
- [x] Modify `CMakePackage`s to use that directive

### DIFF
--- a/lib/spack/spack/build_systems/cmake.py
+++ b/lib/spack/spack/build_systems/cmake.py
@@ -8,7 +8,7 @@ import os
 import platform
 import re
 import sys
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 import llnl.util.filesystem as fs
 
@@ -16,7 +16,7 @@ import spack.build_environment
 import spack.builder
 import spack.package_base
 import spack.util.path
-from spack.directives import build_system, depends_on, variant
+from spack.directives import build_system, conflicts, depends_on, variant
 from spack.multimethod import when
 
 from ._checks import BaseBuilder, execute_build_time_tests
@@ -33,6 +33,43 @@ def _extract_primary_generator(generator):
     """
     primary_generator = _primary_generator_extractor.match(generator).group(1)
     return primary_generator
+
+
+def generator(*names: str, default: Optional[str] = None):
+    """The build system generator to use.
+
+    See ``cmake --help`` for a list of valid generators.
+    Currently, "Unix Makefiles" and "Ninja" are the only generators
+    that Spack supports. Defaults to "Unix Makefiles".
+
+    See https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html
+    for more information.
+
+    Args:
+        names: allowed generators for this package
+        default: default generator
+    """
+    allowed_values = ("make", "ninja")
+    if any(x not in allowed_values for x in names):
+        msg = "only 'make' and 'ninja' are allowed for CMake's 'generator' directive"
+        raise ValueError(msg)
+
+    default = default or names[0]
+    not_used = [x for x in allowed_values if x not in names]
+
+    def _values(x):
+        return x in allowed_values
+
+    _values.__doc__ = f"{','.join(names)}"
+
+    variant(
+        "generator",
+        default=default,
+        values=_values,
+        description="the build system generator to use",
+    )
+    for x in not_used:
+        conflicts(f"generator={x}")
 
 
 class CMakePackage(spack.package_base.PackageBase):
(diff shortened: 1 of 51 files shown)


lib/spack/spack/build_systems/cmake.py
```diff
@@ -35,6 +35,43 @@ def _extract_primary_generator(generator):
...
+    _values.__doc__ = f"{','.join(names)}"
+
+    variant(
+        "generator",
+        default=default,
+        values=_values,
+        description="the build system generator to use",
+    )
+    for x in not_used:
+        conflicts(f"generator={x}")
```
alalazo, line 72: This works fine, but can probably cleaned a bit once we have the `drop` directive from #35045

adamjstewart:
The only reason we didn't do this originally was because the thing getting installed (shouldn't) change, so we didn't want to include it in the hash. But with concretizer reuse the default, maybe this doesn't matter as much.

However, the current implementation is more "make generator a directive" than "make generator a variant". The packages that set `generator = "Ninja"` aren't doing it because only Ninja works, they're doing it because it's faster. I'm in favor of a normal variant at the base class level, and packages shouldn't override the default. I think we should default to Ninja and only override the variant in packages that won't build with Ninja.

alalazo:
> However, the current implementation is more "make generator a directive" than "make generator a variant".

If we have packages that only work with `make` or only work with `ninja` we want their generator to be the default, and the other one to be disallowed. That currently means a combination of `variant` and `conflicts`, and as soon as #35405 gets in of `variant` and `drop`.

The issue are the checks we do on conditional dependencies like:
```python
depends_on("gmake", when="generator=make")
```
that would raise when only `ninja` is allowed. The `generator` wrapper is really a convenience for calling 2 directives with most arguments fixed.

---

Also forgot to mention that a added benefit is that now the dependency on `gmake` or `ninja` is explicit, so I think I can remove a few more directives in packages.

adamjstewart:
Oh, so you're saying if the base class contains:
```python
variant("generator", values=["make", "ninja"])
depends_on("ninja", when="generator=ninja")
```
and a subclass contains:
```python
variant("generator", values=["make"])
```
the subclass will raise audit issues because `generator=ninja` is invalid? That's annoying.

alalazo:
> will raise audit issues because generator=ninja is invalid? That's annoying.

Actually it won't concretize too, because we do consistency checks also during concretization :sweat_smile:

kwryankrattiger:
This is much improved over the current model! Thanks!

alalazo:
@spackbot run pipeline

adamjstewart:
In order to avoid the aforementioned concretizer/audit check issue, could we instead use:
```python
variant("generator", values=["make", "ninja"])
depends_on("ninja", when="generator=ninja")
```
in the base class and:
```python
conflicts("generator=ninja")
```
in subclasses that only support make? Then we don't need to add a whole new directive just for this.

alalazo:
@adamjstewart For reasons related to `clingo` it's better if packages that only support Ninja have `ninja` as the default generator. That means the packages also need to override the variant. Hence, having a declarative way of setting a conflict and declaring a variant while providing only the information that is necessary.

adamjstewart:
> For reasons related to `clingo`

This doesn't tell me a lot, but I trust you. Just trying to avoid adding more and more hyper-specific directives if we don't need them.

alalazo:
> This doesn't tell me a lot, but I trust you. Just trying to avoid adding more and more hyper-specific directives if we don't need them.

If we don't make `ninja` the default, the only value possible would indeed be `ninja` - but with a penalty in the optimization. That can steer clingo towards weird solutions.

alalazo:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

lib/spack/spack/test/build_systems.py
```diff
@@ -268,16 +268,18 @@ def test_cmake_std_args(self, default_mock_concretization):
...
         assert spack.build_systems.cmake.CMakeBuilder.std_args(s.package)
 
-    def test_cmake_bad_generator(self, monkeypatch, default_mock_concretization):
+    def test_cmake_bad_generator(self, default_mock_concretization):
         s = default_mock_concretization("cmake-client")
-        monkeypatch.setattr(type(s.package), "generator", "Yellow Sticky Notes", raising=False)
         with pytest.raises(spack.package_base.InstallError):
-            s.package.builder.std_cmake_args
+            spack.build_systems.cmake.CMakeBuilder.std_args(
+                s.package, generator="Yellow Sticky Notes"
```
haampie, line 275: 😆

lib/spack/spack/build_systems/cmake.py
```diff
@@ -67,8 +104,15 @@ class CMakePackage(spack.package_base.PackageBase):
...
         )
+
+        if sys.platform == "win32":
+            generator("ninja")
+        else:
+            generator("ninja", "make", default="make")
+
         depends_on("cmake", type="build")
-        depends_on("ninja", type="build", when="platform=windows")
+        depends_on("gmake", type="build", when="generator=make")
```
haampie, line 114: I approve of this!

```
$ spack spec gmake +guile | grep cmake
```

has nothing on current develop, so it shouldn't break "advanced" gmake builds. `gmake ~guile` doesn't have deps.

haampie (approved):
I grepped for gmake build deps among packages of which there's only two cmake packages that specify a bound on it, so that looks fine. Ninja I think you got right too

PR #35556, opened 2023-02-17
root: new version 6.28.00

wdconinc:
Full changelog: https://github.com/root-project/root/compare/v6-26-10...v6-28-00

Based on the CMakeLists.txt file, no major build system changes. Upgraded requirements for cmake added.

### DIFF
--- a/var/spack/repos/builtin/packages/root/package.py
+++ b/var/spack/repos/builtin/packages/root/package.py
@@ -32,6 +32,7 @@ class Root(CMakePackage):
     # Development version (when more recent than production).
 
     # Production version
+    version("6.28.00", sha256="afa1c5c06d0915411cb9492e474ea9ab12b09961a358e7e559013ed63b5d8084")
     version("6.26.10", sha256="8e56bec397104017aa54f9eb554de7a1a134474fe0b3bb0f43a70fc4fabd625f")
     version("6.26.08", sha256="4dda043e7918b40743ad0299ddd8d526b7078f0a3822fd06066df948af47940e")
     version("6.26.06", sha256="b1f73c976a580a5c56c8c8a0152582a1dfc560b4dd80e1b7545237b65e6c89cb")
@@ -197,6 +198,8 @@ class Root(CMakePackage):
 
     depends_on("cmake@3.4.3:", type="build", when="@:6.16")
     depends_on("cmake@3.9:", type="build", when="@6.18.00:")
+    depends_on("cmake@3.16:", type="build", when="@6.26.00:")
+    depends_on("cmake@3.19:", type="build", when="@6.28.00: platform=darwin")
     depends_on("pkgconfig", type="build")
 
     depends_on("blas")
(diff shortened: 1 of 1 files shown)


HadrienG2 (changes requested):
I see in the release notes that "The build options gfal, gsl_shared, jemalloc, monalisa, pyroot_legacy, tcmalloc, and xproofd have been deprecated. Please complain with root-dev@cern.ch should you still need one!".

On our side, we expose the jemalloc variant, which we'll probably want to mark as deprecated accordingly, and we internally use gsl_shared when the gsl variant is enabled, so we need to investigate why we use it and how bad losing it will be.

wdconinc:
There is no way to deprecate variants in spack, AFAIK. I propose something like:
```diff
-    variant("jemalloc", default=False, description="Enable using the jemalloc allocator")
+    variant("jemalloc", when="@:6.28", default=False, description="Enable using the jemalloc allocator (deprecated in 6.28)")
```

The `gsl_shared` build option doesn't seem to do anything at all from what I can tell... I can't find any useful references until the git history of the root5 days.

HadrienG2:
I like your proposed solution for the jemalloc variant. Concerning `gsl_shared`, if it doesn't seem to do anything, I would say let's try to disable it (at least if `spec.satisfies("@6.28:")`, but if it's been useless for so long, maybe perma-set it to `False` for all releases?) and see if users report back any problem.

wdconinc:
For the oldest version, all I can find is that `gsl_shared` may simply have been inherited from `./configure` but not implemented in CMake.
```console
12:46:00 wdconinc@menelaos ~/git/root ((tags/v6-06-02^0) %) $ grep -iIr gsl_shared
cmake/modules/RootBuildOptions.cmake:ROOT_BUILD_OPTION(gsl_shared OFF "Enable linking against shared libraries for GSL (default no)")
configure:   enable_gsl_shared         \
configure:enable_gsl_shared=no
configure:    # enable_gsl_shared=$enable_shared"
configure:        check_library "libgsl gslML gsl" "$enable_gsl_shared" "$gsllibdir" \
configure:                    "$enable_gsl_shared" "" \
configure:                if test "x$enable_gsl_shared" = "xyes"; then
configure:       test "x$enable_gsl_shared" = "xno" ; then
```

wdconinc:
gsl_shared introduced here: https://github.com/root-project/root/commit/f75abd6e6e801baddfa45bc906763c62e005c02f

and "gsl_shared is unused" per https://github.com/root-project/root/pull/12203

wdconinc:
@spackbot fix style

HadrienG2 (approved):

var/spack/repos/builtin/packages/root/package.py
```diff
@@ -505,7 +513,6 @@ def cmake_args(self):
             define_from_variant("gl2ps", "opengl"),
             define("glite", False),
             define("globus", False),
-            define_from_variant("gsl_shared", "gsl"),
```
tldahlgren, line 508: Just to confirm: removing this option for *all* version installs was deliberate, correct?

tldahlgren (approved):
Confirmed the sha256.  Just a quick question about removal of one of the options.

var/spack/repos/builtin/packages/root/package.py
wdconinc: It never did anything.

wdconinc: Not in this millenium anyway.

PR #35558, opened 2023-02-17
Add `spack git` cmd to run git within spack's install dir

alecbcs:
This PR adds a minimal `spack git` command which allows users to run git commands within the spack repository without requiring them to first cd into SPACK_ROOT. This command is philosophically very similar to git sub commands in other projects like [pass](passwordstore.org). I've personally wanted a command like this in spack for a long time so that when updating files I can,

```
spack git pull
spack edit packageX
spack git checkout -b update/packageX
spack git add .
spack git commit
spack git push
```

without needing to dig down into my spack directory. I'm curious to learn if others in the Spack community might also find this command useful.


lib/spack/spack/cmd/git.py
```diff
@@ -0,0 +1,36 @@
...
+    subparser.add_argument(
+        "command", nargs=argparse.REMAINDER, help="git command to run within spack repo"
+    )
+
+
+def git(parser, args):
+    # make sure this is a git repo
+    if not spack_is_git_repo():
+        tty.die("This spack is not a git clone. Can't use 'spack git'")
+    git = spack.util.git.git(required=True)
```
becker33: Could make this bootstrap git if we wanted to...

lib/spack/spack/cmd/git.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # execute git within the local spack repository
+    with working_dir(spack.paths.prefix):
+        output = git(*args.command, output=str, fail_on_error=False)
+        print(output.strip())
```
becker33: Why pass the output to a string, just to print it?

alecbcs: I felt like a single line combination was a bit long,
```
print(git(*args.command, output=str, fail_on_error=False).strip())
```

although rewriting as the following is still pretty understandable,
```
print(git(*args.command, output=str, fail_on_error=False), end='')
```

lib/spack/spack/cmd/git.py
alecbcs: What benefits would using the bootstrap git provide over standard git?

alecbcs:
@spackbot fix style

haampie:
Are we going to wrap everything? git, grep, cd.

Not necessarily against this feature, but at this point we might as well consolidate and do `spack run ...`, make it change working dir, followed by execv(argv[2:])

lib/spack/spack/cmd/git.py
```diff
@@ -0,0 +1,35 @@
...
+
+def git(parser, args):
+    # make sure this is a git repo
+    if not spack_is_git_repo():
+        tty.die("This spack is not a git clone. Can't use 'spack git'")
+    git = spack.util.git.git(required=True)
+
+    # execute git within the local spack repository
+    with working_dir(spack.paths.prefix):
+        print(git(*args.command, output=str, fail_on_error=False), end="")
```
haampie: how about execv?

haampie:
Actually `spack run ...` could go into the shell wrapper for spack s.t. python/spack startup time is avoided 🙃

lib/spack/spack/cmd/git.py
psakievich: I think bootstrapping git maybe a larger discussion? Maybe separate PR? One advantage is that on some systems the version of git is quite old and many modern features fail. We had an NREL platform where all spack based `git clone` operations were failing for some users and not others.  Turns out some had an updated git in their path.

alecbcs:
> Are we going to wrap everything? git, grep, cd.

I don't think we need to wrap everything. Git in particular is closely tied to our Spack workflows. For example, I'm constantly using git to keep my local spack up-to-date, try out other's branches, and push patches back up to the GitHub repository and I often find myself quickly changing into my spack install directory just to then quickly use git.

> Not necessarily against this feature, but at this point we might as well consolidate and do spack run ..., make it change working dir, followed by execv(argv[2:])

Perhaps? I think there's a fine line to balance here between functionality and complexity. `spack run git pull` doubles the number of characters we'll have to type for every spack related git command and at that point it might just be easier to create a custom `sgit` alias for most users.

> could go into the shell wrapper for spack s.t. python/spack startup time is avoided 🙃

Oh this is a fantastic idea! Should we also keep a boilerplate cmd entry so that the command shows up in the help dialogue?

haampie:
I agree this is a useful thing to have, don't get me wrong.

I'm just a tiny bit worried we get tons of slightly different wrappers for unix tools. I don't like `spack python` for example cause the arg parsing is done in spack, and is effectively a small subset of accepted python flags.
 
What would be super nice is if the wrapper is non-intrusive, and we could *also* get normal tab completion for `spack (run) git` 🙃

lib/spack/spack/cmd/git.py
tgamblin: In most other commands that wrap os commands, we find the command in path or fail, so I'd just continue that here.

lib/spack/spack/cmd/git.py
tgamblin: Why not just run git and *not* redirect the output?

lib/spack/spack/cmd/git.py
tgamblin: `execv` is a better choice here because it doesn't run git's potentially colored/formatted/tty-sensitive output through our redirection.  See, e.g., the (as yet incomplete) refactor in #34105.

The other thing you could potentially do here is re-route the original `sys.argv` arguments to `execv` more cleanly, and avoid argparse wrapper nastiness.  See [this part of #34388](https://github.com/spack/spack/pull/34388/files#diff-cde3b3e250ab1d245a3a0dcfefcb41ea14dd1d31bc2e189c1682807ca331a701R177). Yes, commands can ask for unknown args, but there are some weird parse issues, e.g. `-A5` seems to work with `spack pkg grep` but not `-A 5`.

I'm not sure what range of `git` arguments you can actually get through `argparse.REMAINDER`, I guess it might be all of them... since this command is simpler than `spack pkg grep`... but I haven't tried.

alecbcs:
I haven't figured out how to get passthrough bash autocompletion to work yet but taking @haampie's suggestion here's an initial stab at implementing a run and aliased `spack git` command within the spack shell wrapper. Feedback is hugely welcome. Modifying the shell wrapper for Spack seems like something we should do carefully so I'd appreciate any and all feedback on making this system more efficient.

bin/spack
```diff
@@ -10,6 +10,18 @@
...
+if [ -n "$1" ] && [ $1 == "run" ]; then
+  shift;
+  run_aliases=("$1" "${run_aliases[@]}")
+fi
+
+if [[ " ${run_aliases[*]} " =~ " $1 " ]]; then
+  (cd $SPACK_ROOT && exec $@)
+  exit $?
+fi
+
```
becker33:
```suggestion
# preempt python load and natively execute program in $SPACK_ROOT
if [ "$1"x == gitx ]; then
  (cd $SPACK_ROOT && exec $@)
  exit $?
fi

```

becker33 (commented):
I don't think we should take over the `spack run` name with a command that's for internal developers. Let's keep this at just `spack git` for now, and we can always add to it easily later.

alecbcs:
Over time I've changed my mind on this topic and I think it's better to leave Git out of Spack. This largely came about since I was tired of running,

```console
$ cd ~/src/spack/spack
```

I know...silly right. Anyway I solved that for myself with the relatively simple alias below.

```bash
alias cdsp="cd $SPACK_ROOT"
```

I'm closing this now in favor of #37549 since I do still think that new and casual users of Spack need a way of updating the application and repository without necessarily needing to remember where they've installed Spack and what git commands they need to run to keep everything up-to-date.

psakievich:
Do you use `spack cd -r`?  This does the same thing.  Another hack here is `git -C $SPACK_ROOT` which I sometimes do for the same purpose 🙂

PR #35560, opened 2023-02-18
Add go v1.20.x, 1.19.6 and un-deprecate previous versions without CVEs

alecbcs:
Adding Go versions v1.20.1, v1.20.0, and 1.19.6. Although I'd previously been deprecating all but the latest version of every major release, as seemed necessary due to related CVEs, I'd like to instead pivot to only deprecating Go versions as CVEs are published.

### DIFF
--- a/var/spack/repos/builtin/packages/go/package.py
+++ b/var/spack/repos/builtin/packages/go/package.py
@@ -44,19 +44,15 @@ class Go(Package):
 
     maintainers("alecbcs")
 
+    version("1.20.1", sha256="b5c1a3af52c385a6d1c76aed5361cf26459023980d0320de7658bae3915831a2")
+    version("1.20", sha256="3a29ff0421beaf6329292b8a46311c9fbf06c800077ceddef5fb7f8d5b1ace33")
+
+    version("1.19.6", sha256="d7f0013f82e6d7f862cc6cb5c8cdb48eef5f2e239b35baa97e2f1a7466043767")
     version("1.19.5", sha256="8e486e8e85a281fc5ce3f0bedc5b9d2dbf6276d7db0b25d3ec034f313da0375f")
-    version(
-        "1.19.4",
-        sha256="eda74db4ac494800a3e66ee784e495bfbb9b8e535df924a8b01b1a8028b7f368",
-        deprecated=True,
-    )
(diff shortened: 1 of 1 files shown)


haampie (approved):

vanderwb:
@alecbcs - this PR seems to have made the 1.20.x version the default, but this results (at least on our system) in a Go that cannot build, as I believe >= 1.20 cannot be bootstrapped by the C version. This is what I am seeing:

```
Building Go cmd/dist using /glade/gust/scratch/vanderwb/spack-tests/gust/23.03b/envs/build/opt/__spack_path_placeholder__/__spack_path_placeholder__/__spac/go-bootstrap/1.4-bootstrap-20171003/gcc/7.5.0. (go1.4-bootstrap-20170531 linux/amd64)
can't load package: package ./cmd/dist: found packages build.go (main) and notgo117.go (building_Go_requires_Go_1_17_13_or_later) in /glade/gust/scratch/vanderwb/temp/spack/gust/23.03b/builds/spack-stage-go-1.20.1-fzh2beqgqnzx3cywd437ofvujvyioboz/spack-src/src/cmd/dist
```

If nothing else, perhaps the 1.19 series should be kept default until the build mechanism in the package.py is modified (I guess to prefer gccgo?).

alecbcs:
Thanks for the bug report @vanderwb! This is a good push for us to look to refactor the way Go is currently bootstrapped now that bootstrapping from C is no longer possible. Does bootstrapping from GCC work for your system or is defaulting to Go 19.x a hard requirement?

vanderwb:
@alecbcs - It shouldn't be a hard requirement, and I suspect using `gcc languages=go` would work fine (we are already building GCC with Go support). That said, I haven't tested it out yet since this recipe dictates the bootstrapping approach on x86. I'll doing a quick hack and make sure our stack still looks okay.

vanderwb:
I just tested out a build of go@1.20.1 using the GCC approach and it worked well for our uses, though I needed to use a very modern version of GCC (12 series) to get the requisite Go support. I think that's a viable approach moving forward for those on x86 who don't have an external Go they can use (and perhaps older Spack builds of Go should be a viable provider option too?).

That said, I'm agnostic to the solution, since we have a number of workarounds we can use in the meantime. Thanks for looking at the problem!

PR #35564, opened 2023-02-18
Disable module generation by default

haampie:
1. It's used by site administrators, so it's niche, the rest of the Spack users does not need this;
2. If it's used by site administrators, they likely need to modify the config anyhow, so the default config only serves as an example on how to get started;
3. it's too arbitrary to enable tcl, but disable lmod
4. modules can be generated with `spack modules [type] refresh ...`, so nothing is lost.

To get the old behavior back, one can simply run

```
spack config --scope site add modules:default:enable:[tcl]
```

### DIFF
--- a/etc/spack/defaults/modules.yaml
+++ b/etc/spack/defaults/modules.yaml
@@ -40,9 +40,8 @@ modules:
     roots:
      tcl:    $spack/share/spack/modules
      lmod:   $spack/share/spack/lmod
-    # What type of modules to use
-    enable:
-      - tcl
+    # What type of modules to use ("tcl" and/or "lmod")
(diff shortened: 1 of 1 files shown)


alalazo (approved):
Completely approve.

haampie:
I didn't expect that to get merged right away :p should maybe be announced in the "upcoming high-impact changes", because people who use tcl will be wasting time figuring out why it's not getting generated.

alalazo:
> I didn't expect that to get merged right away :p should maybe be announced in the "upcoming high-impact changes", because people who use tcl will be wasting time figuring out why it's not getting generated.

Done that before merging

PR #35567, opened 2023-02-18
`py-pyarrow`: fix build environment

thomas-bouvier:
Fixes the undefined spec variable + wrong env variable passed to the build environment.

### DIFF
--- a/var/spack/repos/builtin/packages/py-pyarrow/package.py
+++ b/var/spack/repos/builtin/packages/py-pyarrow/package.py
@@ -66,13 +66,13 @@ class PyPyarrow(PythonPackage, CudaPackage):
     patch("for_aarch64.patch", when="@0 target=aarch64:")
 
     def setup_build_environment(self, env):
-        if spec.satisfies("+parquet"):
-            env.set("PYARROW_PARQUET", "1")
-        if spec.satisfies("+cuda"):
+        if self.spec.satisfies("+parquet"):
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/py-pyarrow/package.py
```diff
@@ -66,13 +66,13 @@ class PyPyarrow(PythonPackage, CudaPackage):
...
+        if self.spec.satisfies("+parquet"):
+            env.set("PYARROW_WITH_PARQUET", "1")
+        if self.spec.satisfies("+cuda"):
             env.set("PYARROW_WITH_CUDA", "1")
-        if spec.satisfies("+orc"):
+        if self.spec.satisfies("+orc"):
             env.set("PYARROW_WITH_ORC", "1")
-        if spec.satisfies("+dataset"):
+        if self.spec.satisfies("+dataset"):
             env.set("PYARROW_WITH_DATASET", "1")
```
adamjstewart: We should actually simplify this to:
```python
env.set("PYARROW_WITH_PARQUET", self.spec.satisfies("+parquet"))
...
```
Fewer lines of code and more thorough. Otherwise, it's possible that one of these environment variables could be set outside of Spack and:
```console
$ PYARROW_WITH_PARQUET=1 spack install py-pyarrow~parquet
```
would install something different.

adamjstewart (approved):

thomas-bouvier:
@spackbot run pipeline

PR #35569, opened 2023-02-18
pandoc: minimize dependencies

adamjstewart:
Texlive has a gigantic dependency tree, we should disable it by default to minimize dependencies.


michaelkuhn (approved):
Sounds sensible to me

PR #35576, opened 2023-02-18
Add pyedr and panedr

RMeli:


var/spack/repos/builtin/packages/py-pyedr/package.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class PyPyedr(PythonPackage):
+    """Reads Gromacs EDR file to populate a pandas dataframe"""
```
RMeli:
```suggestion
    """Pyedr provides a means of reading a Gromacs EDR binary XDR file and return
    its contents as a dictionary of NumPy arrays"""
```

var/spack/repos/builtin/packages/py-panedr/package.py
```diff
@@ -0,0 +1,25 @@
...
+    its contents as a pandas dataframe"""
+
+    homepage = "https://github.com/MDAnalysis/panedr"
+    pypi = "panedr/panedr-0.7.1.tar.gz"
+
+    maintainers("RMeli")
+
+    version("0.7.1", sha256="64c74863f72d51729ac5cb1e2dbb18747f7137588990c308ef8ca120fbf2ddd4")
+
+    depends_on("py-pyedr@0.7.1", type=("build", "run"), when="@0.7.1")
```
adamjstewart, line 21: Where do you see the version requirement?

var/spack/repos/builtin/packages/py-panedr/package.py
```diff
@@ -0,0 +1,25 @@
...
+
+    depends_on("py-pandas", type=("build", "run"))
+    depends_on("py-pbr", type=("build", "run"))
+
+    depends_on("py-setuptools@42:", type="build")
```
adamjstewart: Where do you see the version requirement?

var/spack/repos/builtin/packages/py-pyedr/package.py
```diff
@@ -0,0 +1,26 @@
...
+    homepage = "https://github.com/MDAnalysis/panedr"
+    pypi = "pyedr/pyedr-0.7.1.tar.gz"
+
+    maintainers("RMeli")
+
+    version("0.7.1", sha256="ad7ccdeb739399acd11a25f2d2413ebb46a54223059a2b902ac604d29fabd767")
+
+    depends_on("python@3.6:")
+
+    depends_on("py-numpy@1.19.0:", type=("build", "run"))
```
adamjstewart, line 23: Where do you see the version requirement?

var/spack/repos/builtin/packages/py-pyedr/package.py
```diff
@@ -0,0 +1,26 @@
...
+    depends_on("py-pbr", type=("build", "run"))
+    depends_on("py-tqdm", type=("build", "run"))
+
+    depends_on("py-setuptools@42:", type="build")
```
adamjstewart: Where do you see the version requirement?

var/spack/repos/builtin/packages/py-pyedr/package.py
```diff
@@ -0,0 +1,26 @@
...
+    its contents as a dictionary of NumPy arrays"""
+
+    homepage = "https://github.com/MDAnalysis/panedr"
+    pypi = "pyedr/pyedr-0.7.1.tar.gz"
+
+    maintainers("RMeli")
+
+    version("0.7.1", sha256="ad7ccdeb739399acd11a25f2d2413ebb46a54223059a2b902ac604d29fabd767")
+
+    depends_on("python@3.6:")
```
adamjstewart:
```suggestion
    depends_on("python@3.6:", type=("build", "run"))
```
Or just remove it. Spack only supports Python 3.7+ anyway

var/spack/repos/builtin/packages/py-panedr/package.py
RMeli:
```suggestion
    depends_on("py-setuptools", type="build")
```

var/spack/repos/builtin/packages/py-pyedr/package.py
RMeli:
```suggestion
    depends_on("py-setuptools", type="build")
```

var/spack/repos/builtin/packages/py-pyedr/package.py
RMeli: It it here: https://github.com/MDAnalysis/panedr/blob/7268391a60e14b7c065eb1f7c074e14c29e59515/pyedr/requirements.txt#L1 However, it did not make it to `setup.cfg`.

var/spack/repos/builtin/packages/py-panedr/package.py
RMeli: This is a bit of an assumption with some insider knowledge. Since version `0.6.0`, `panedr` has been split into `pyedr` and `panedr` (which now depends on the former). The two packages live in the same [repository](https://github.com/MDAnalysis/panedr) and are [released together](https://github.com/MDAnalysis/panedr/blob/7268391a60e14b7c065eb1f7c074e14c29e59515/.github/workflows/deploy.yaml#L42-L56), therefore I think it is safer to match their versions.

adamjstewart: Can you add a comment to that effect?

var/spack/repos/builtin/packages/py-pyedr/package.py
adamjstewart: Can you add a comment explaining where it comes from? Since it's not in a location search by pip I also don't know where to look for it.

RMeli:
```suggestion
    # Minimal NumPy version only specified in requirements.txt
    depends_on("py-numpy@1.19.0:", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-panedr/package.py
RMeli:
```suggestion
    # PyEDR is released together with PanEDR, therefore versions should match
    depends_on("py-pyedr@0.7.1", type=("build", "run"), when="@0.7.1")
```

RMeli:
Thanks @adamjstewart for the thorough review!

adamjstewart (approved):

PR #35582, opened 2023-02-19
Fix broken links in docs

adamjstewart:
Discovered using:
```console
$ cd lib/spack/docs
$ make linkcheck
```
There are still plenty of outdated (e.g., https://github.com/LLNL/spack) links that redirect to the correct place or links that use http instead of https that could be updated. Unfortunately we use too many fake URLs so we can't add this to our tests.

### DIFF
--- a/lib/spack/docs/.gitignore
+++ b/lib/spack/docs/.gitignore
@@ -5,3 +5,4 @@ llnl*.rst
 _build
 .spack-env
 spack.lock
+_spack_root
--- a/lib/spack/docs/build_systems/wafpackage.rst
+++ b/lib/spack/docs/build_systems/wafpackage.rst
@@ -58,9 +58,7 @@ Testing
 ``WafPackage`` also provides ``test`` and ``installtest`` methods,
 which are run after the ``build`` and ``install`` phases, respectively.
 By default, these phases do nothing, but you can override them to
-run package-specific unit tests. For example, the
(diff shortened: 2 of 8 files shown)


lib/spack/docs/.gitignore
```diff
@@ -5,3 +5,4 @@ llnl*.rst
 _build
 .spack-env
 spack.lock
+_spack_root
```
adamjstewart, line 8: Building the docs creates this symlink, which then ends up in git's stage directory

lib/spack/docs/build_systems/wafpackage.rst
```diff
@@ -58,9 +58,7 @@ Testing
 ``WafPackage`` also provides ``test`` and ``installtest`` methods,
 which are run after the ``build`` and ``install`` phases, respectively.
 By default, these phases do nothing, but you can override them to
-run package-specific unit tests. For example, the
-`py-py2cairo <https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-py2cairo/package.py>`_
```
adamjstewart, line 62: This package was removed (Python-2 only), the only remaining WafPackage is ns-3-dev. I think @alalazo is actually planning on removing WafPackage

haampie (approved):

lib/spack/docs/build_systems/wafpackage.rst
alalazo: I'd really like to, but there are a few concerns see https://github.com/spack/spack/pull/34084#issuecomment-1329762725

PR #35583, opened 2023-02-20
LAMMPS: Add Intel Package Support for AOCC 4.0

amd-toolchain-support:
This PR adds support to compile LAMMPS with `+intel` using AOCC 4.0


alalazo:
Can you upstream the patch too? Asking in various PRs since I see a pattern of having patches for AMD in multiple packages, while the best way to deal with compiler support imo would be to have:
1. Patches as a stopgap
2. Support for a new compiler added upstream in newer versions

amd-toolchain-support:
See https://github.com/lammps/lammps/pull/3562 for upstream MR

var/spack/repos/builtin/packages/lammps/package.py
```diff
@@ -854,6 +855,7 @@ def url_for_version(self, version):
     patch("lib.patch", when="@20170901")
     patch("660.patch", when="@20170922")
     patch("gtest_fix.patch", when="@:20210310 %aocc@3.2.0")
+    patch("intel-aocc.patch", when="@20220324: +intel %aocc")
```
rbberger: Since this was merged in upstream on 2022-12-16, it should already be part of the patch release `20221222`. That patch should therefore only be applied until the previous patch release `20221103`.

rbberger (changes requested):

var/spack/repos/builtin/packages/lammps/package.py
amd-toolchain-support: Thanks for the suggestion, this has been addressed in the latest commit

rbberger: The when condition is still wrong. It should be `@20220324:20221103`. You want it to apply until and including `20221103`, but not after.

rbberger (dismissed):

alalazo:
@amd-toolchain-support Can you resolve the conflict?

amd-toolchain-support:
We will create a new PR for LAMMPS

PR #35585, opened 2023-02-20
macos: set new ad-hoc signature with codesign after binary string replacement

haampie:
```
% /Users/harmenstoppels/lol/darwin-ventura-m1/apple-clang-14.0.0/tar-1.34-xswbvqgcxw4xeo4rpnwaj4ojqgj3qmcp/bin/tar --version   
tar (GNU tar) 1.34
```

it works again. Apple's codesigning is not a security feature.

The only difference between before and after is that the codesign flags go from "adhoc,linker-signed" to just "adhoc", but I don't see any consequences of that.


alalazo (approved):
Can't check right now on Apple M*, but code LGTM.

adamjstewart:
Seems like we're still having relocation issues: https://github.com/spack/spack/pull/36586#issuecomment-1507355312

PR #29317, opened 2022-03-03
remove activate/deactivate support in favor of environments

haampie:
Environments and environment views have taken over the role of `spack activate/deactivate`, and we should deprecate these commands for several reasons:

- Global activation is a really poor idea:
   - Install prefixes should be immutable; since they can have multiple, unrelated dependents; see below
   - Added complexity elsewhere: verification of installations, tarballs for build caches, creation of environment views of packages with unrelated extensions "globally activated"... by removing the feature, it gets easier for people to contribute, and we'd end up with fewer bugs due to edge cases.
- Environment accomplish the same thing for non-global "activation" i.e. `spack view`, but better.

Also we write in the docs:

```
However, Spack global activations have two potential drawbacks:

#. Activated packages that involve compiled C extensions may still
   need their dependencies to be loaded manually.  For example,
   ``spack load openblas`` might be required to make ``py-numpy``
   work.

#. Global activations "break" a core feature of Spack, which is that
   multiple versions of a package can co-exist side-by-side.  For example,
   suppose you wish to run a Python package in two different
   environments but the same basic Python --- one with
   ``py-numpy@1.7`` and one with ``py-numpy@1.8``.  Spack extensions
   will not support this potential debugging use case.
```

Now that environments are established and views can take over the role of activation
non-destructively, we can remove global activation/deactivation.

PR #35588, opened 2023-02-20
Remove unused `ignore` parameter of `extends()` directive

adamjstewart:
If I'm understanding correctly, the `ignore` parameter was only used for `spack activate/deactivate`, it isn't used by Spack Environments which has its own handling of file conflicts. We should remove it.

Is there somewhere I can check for this and issue a deprecation warning? Or is it harmlessly ignored anyway? This is one of those situations where we use kwargs where we really shouldn't, but have to keep doing so due to old Python version support.

### DIFF
--- a/lib/spack/docs/packaging_guide.rst
+++ b/lib/spack/docs/packaging_guide.rst
@@ -2726,29 +2726,6 @@ variant(s) are selected.  This may be accomplished with conditional
        extends("python", when="+python")
        ...
 
-Sometimes, certain files in one package will conflict with those in
-another, which means they cannot both be used in a view at the
-same time.  In this case, you can tell Spack to ignore those files:
-
-.. code-block:: python
-
-   class PySncosmo(Package):
-       ...
-       # py-sncosmo binaries are duplicates of those from py-astropy
-       extends("python", ignore=r"bin/.*")
-       depends_on("py-astropy")
-       ...
-
-The code above will prevent everything in the ``$prefix/bin/`` directory
-from being linked in a view.
-
-.. note::
-
-   You can call *either* ``depends_on`` or ``extends`` on any one
-   package, but not both.  For example you cannot both
-   ``depends_on("python")`` and ``extends("python")`` in the same
-   package.  ``extends`` implies ``depends_on``.
-
 -----
 Views
 -----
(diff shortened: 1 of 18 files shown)


haampie:
In environment views we ignore file-file conflicts right now.

Not sure if this argument is still used in `spack view ...` to make file-file conflicts non-fatal.

adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

tgamblin (approved):

tgamblin:
Everything that handles `ignore=` was removed in #29317 and included in 0.19, when we removed `spack activate` and `spack deactivate` in favor of environments.  So all of these are already being ignored by Spack -- the only place they were used was in activate/deactivate.

tgamblin:
@adamjstewart I just enabled auto-mere on this since ignore is harmlessly ignored by Spack.  I think we should get the change in before it needs another rebase.

I'll put in a follow-up PR to remove the unused `extendee_args()` method and to restructure the way `extends()` works.  I'll also add a deprecation warning for `ignore=` specifically and stop accepting other arguments to extends.

alalazo:
@spackbot run pipeline

tgamblin:
@alalazo how did the pipeline fail?

alalazo:
It was a spurious failure with a "Spack error" on a Python package. On rebuild the package was installed correctly.

alalazo:
The python package in question is `py-protobuf`, here's the rebuild that went :heavy_check_mark: : https://gitlab.spack.io/spack/spack/-/pipelines/306861

tgamblin:
do you remember what type of spack error or can you link to the broken pipeline?

alalazo:
This is the pipeline that was failing: https://gitlab.spack.io/spack/spack/-/pipelines/306620 And this is the job in question: https://gitlab.spack.io/spack/spack/-/jobs/6182731

alalazo:
It doesn't give a lot of hints other than "Failed to call Python: exit status -4"

tgamblin:
Yeah and the log is just the `tty` output showing the command line that was executed. This one's weird.

PR #35595, opened 2023-02-20
ExaGO 1.5.1 release

cameronrutherford:
Version release for ExaGO, along with some QOL life changes to the ExaGO/HiOp packages.

Since HiOp is the main package that uses RAJA within the ExaSGD stack, I decided that having RAJA version constraint logic, as well as cuda arch specification consolidated to HiOp makes the most sense.

cc @pelesh @ryandanehy @jaelynlitz

### DIFF
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -121,7 +121,7 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("umpire@6.0.0:6", when="@1.1.0:+raja")
     depends_on("camp@0.2.3:0.2", when="@1.1.0:+raja")
     # This is no longer a requirement in RAJA > 0.14
-    depends_on("umpire+cuda~shared", when="+raja+cuda")
+    depends_on("umpire+cuda~shared", when="+raja+cuda ^raja@:0.14")
 
     depends_on("petsc@3.13:3.14", when="@:1.2.99")
     depends_on("petsc@3.16.0:3.16", when="@1.3.0:1.4")
--- a/var/spack/repos/builtin/packages/exago/package.py
+++ b/var/spack/repos/builtin/packages/exago/package.py
@@ -33,7 +33,7 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
 
     # Progrmming model options
     variant("mpi", default=True, description="Enable/Disable MPI")
-    variant("raja", default=False, when="+hiop", description="Enable/Disable RAJA with HiOp")
+    variant("raja", default=False, description="Enable/Disable RAJA")
     variant("python", default=True, when="@1.4:", description="Enable/Disable Python bindings")
     variant("logging", default=True, description="Enable/Disable spdlog based logging")
     conflicts(
@@ -63,6 +63,8 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("blas")
     depends_on("ipopt~mumps", when="+ipopt")
     depends_on("cuda", when="+cuda")
+    depends_on("raja", when="+raja")
+    depends_on("umpire", when="+raja")
 
     depends_on("cmake@3.18:", type="build")
 
@@ -101,7 +103,6 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     # ^ need to depend when both hpctoolkit and tau
 
     # HiOp dependency logic
-    depends_on("hiop+raja", when="+hiop+raja")
     depends_on("hiop@0.3.99:", when="@0.99:+hiop")
     depends_on("hiop@0.5.1:", when="@1.1.0:+hiop")
     depends_on("hiop@0.5.3:", when="@1.3.0:+hiop")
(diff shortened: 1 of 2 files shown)


pelesh (changes requested):
I can see the short-term benefit in this PR, but I am not sure this is the best long-term solution. ExaGO uses Umpire and RAJA on its own and there is an obvious value of using RAJA in ExaGO without HiOp at all. Right off the bat you could use OpenMP RAJA policy in ExaGO with Ipopt optimization engine. This PR would make HiOp mandatory dependency if you want to use thread parallel model evaluation on the host.

It would make more sense to me if ExaGO Spack package builds HiOp with the same RAJA and Umpire libraries. But, I really don't see a need for one of these packages to set RAJA/Umpire dependency for the other.

cameronrutherford:
> I can see the short-term benefit in this PR, but I am not sure this is the best long-term solution. ExaGO uses Umpire and RAJA on its own and there is an obvious value of using RAJA in ExaGO without HiOp at all. Right off the bat you could use OpenMP RAJA policy in ExaGO with Ipopt optimization engine. This PR would make HiOp mandatory dependency if you want to use thread parallel model evaluation on the host.
> 
> 
> 
> It would make more sense to me if ExaGO Spack package builds HiOp with the same RAJA and Umpire libraries. But, I really don't see a need for one of these packages to set RAJA/Umpire dependency for the other. 

Perhaps this is true in the future. We will want to support ExaGO with RAJA without HiOp, and any matrix of that.

This configuration allows us to build ExaGO with and without RAJA, regardless of what HiOp has.

Perhaps I can re-introduce the version constraints of RAJA/Umpire/Camp to ExaGO then? It perhaps only makes sense to introduce RAJA versions as a dependency of ExaGO once we support that.

I'm happy with this PR apart from that. We can always edit the package again in the future to support additional functionality.

cameronrutherford:
ExaGO builds just fine using RAJA through HiOp in this configuration FWIW

pelesh:
> This configuration allows us to build ExaGO with and without RAJA, regardless of what HiOp has.

I think this is very much needed and valuable contribution of this PR. However, if RAJA is enabled in ExaGO, there should be no constraints from HiOp as to what RAJA version ExaGO should use.

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -47,29 +48,21 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
     depends_on("mpi", when="+mpi")
     depends_on("blas")
     depends_on("ipopt~mumps", when="+ipopt")
     depends_on("cuda", when="+cuda")
-    depends_on("raja", when="+raja")
-    depends_on("umpire", when="+raja")
-
-    depends_on("umpire@6.0.0", when="@1.1.0: +raja")
-    depends_on("raja@0.14.0", when="@1.1.0: +raja")
-    depends_on("camp@0.2.3", when="@1.1.0: +raja")
```
pelesh: I would keep this in ExaGO configuration instead of letting HiOp set RAJA and related dependencies.

cameronrutherford:
@pelesh please re-review

var/spack/repos/builtin/packages/exago/package.py
```diff
@@ -108,15 +103,26 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
+    # ExaGO will support +raja~hiop in the future
+    depends_on("hiop+raja", when="+hiop+raja")
+    # This is duplicated from HiOp
+    # RAJA > 0.14 and Umpire > 6.0 require c++ std 14
+    # We are working on supporting newer Umpire/RAJA versions
+    depends_on("raja@0.14.0:0.14", when="@1.1.0:+raja")
+    depends_on("umpire@6.0.0:6", when="@1.1.0:+raja")
+    depends_on("camp@0.2.3:0.2", when="@1.1.0:+raja")
+    # This is no longer a requirement in RAJA > 0.14
+    depends_on("umpire+cuda~shared", when="+raja+cuda")
```
tldahlgren: Did you want to be more specific in the `when` argument then?

I believe you can specify `when="+raja+cuda ^raja@:0.14"`.

var/spack/repos/builtin/packages/hiop/package.py
```diff
@@ -101,9 +105,20 @@ class Hiop(CMakePackage, CudaPackage, ROCmPackage):
...
-    depends_on("raja@0.14.0:0.14", when="@0.5.0:0.6.2+raja")
-    depends_on("umpire@6.0.0:", when="@0.5.0:0.6.2+raja")
+    # RAJA > 0.14 and Umpire > 6.0 require c++ std 14
+    # We are working on supporting newer Umpire/RAJA versions
+    depends_on("raja@0.14.0:0.14", when="@0.5.0:+raja")
+    depends_on("umpire@6.0.0:6", when="@0.5.0:+raja")
+    depends_on("camp@0.2.3:0.2", when="@0.5.0:+raja")
+
+    # This is no longer a requirement in RAJA > 0.14
+    depends_on("umpire+cuda~shared", when="+raja+cuda")
```
tldahlgren: You can be more specific here about the version of raja in the `when` argument.

tldahlgren (dismissed):
Confirmed the new commit.  Have a minor suggestion.

var/spack/repos/builtin/packages/exago/package.py
cameronrutherford: Great point, thanks! I have added that to both packages.

This is something that was a soft requirement of the RAJA/Umpire suite for a while but should only apply to specific versions...

var/spack/repos/builtin/packages/hiop/package.py
cameronrutherford: I fixed this, as well as another shared requirement in HiOp

pelesh (commented):
Looks good. I suggest to merge if all tests pass.

tldahlgren (approved):
LGTM

PR #35596, opened 2023-02-20
Pull in external module dependencies when writing modulefiles

koysean:
Fix for https://github.com/spack/spack/issues/35212

This PR modifies spack's module APIs to partially support multiple modulefiles for a single spec. These changes allow Spack to handle an external package with multiple modulefiles.

The primary API changes are in spack.modules.common.get_module() and spack.modules.common.BaseFileLayout.use_name. The return types were changed from str to List[str] to accomodate a single spec having multiple module names.

In addition, modulefiles are no longer created for external packages with modules, and modulefiles which depend on external modules will now directly load the external modules (if autoload is enabled).

One possible issue is that the module index will not be updated with external modules. This doesn't seem like a problem to me since Spack isn't managing the external modules anyways, but I'm not very familiar with how Spack uses the module index. Is this an issue?

(this is a re-done duplicate of https://github.com/spack/spack/pull/35498 because I completely botched the commit history on that branch...)

@becker33 I addressed the requested changes in the original PR - sorry about the confusion with the duplicate PR and such.


lib/spack/spack/modules/common.py
```diff
@@ -390,12 +393,25 @@ def get_module(module_type, spec, get_full_path, module_set_name="default", requ
...
+            if required:
+                tty.debug("The module configuration has excluded {0}: " "omitting it".format(spec))
+            else:
+                return [None]
+
+        if get_full_path:
+            err_msg = "Cannot retrieve full path to an external module"
+            raise NotImplementedError(err_msg)
+        else:
+            return spec.external_modules
```
alalazo: As far as I understand this is the main part that gets the new functionality in. Instead of changing inner classes that generate modules, can't we just:
1. Change this function to return a list?
2. Have another, separate class to deal with external modules?

If I read this correctly `*Writer` will always write a single module file for each spec, and here we are talking about interfacing with external (already existing) module files. Is that right?

koysean: Thanks for the feedback. I think making a separate class for external modules isn't actually needed - I've made some changes that minimize the external API changes (including reverting the changes to `*FileLayout.use_name`) and avoid creating a `*Writer` object for external modules.

With the new commits the only functional change to the internal module classes is in `BaseConfiguration._create_module_list_of()`, and the only external change is to `get_modules()`.

becker33 (changes requested):
I think this PR needs some additional thought to consider what happens if the user is requesting `tcl` modules but the external modules are written in `lmod`. The opposite direction is ok since lmod reads tcl modules, but that direction would cause a problem.

koysean:
Thank you for the review. I'm having trouble imagining a scenario in which that is a problem - could you elaborate? If a user is trying to read lua modules using Environment Modules, that's a case of (Spack-unrelated) user error. If the concern is that a user generates modules written in tcl but those modules load lua modules, `lmod` is capable of parsing that.

PR #35598, opened 2023-02-21
Fix Spec.satisfies semantics

haampie:
If you see Specs as sets with concrete specs as singleton sets in the extreme, then

```python
x.satisfies(y[, strict=False])
```
is commutative, and tests non-empty intersection: x ∩ y ≠ Ø.

Whereas

```python
x.satsifies(y, strict=True)
```

is not commutative, and tests set-inclusion: x ⊆ y.

Currently there's a bug in `x.satisfies(y)` where if you give it a concrete
spec y and an abstract spec x, it's no longer commutative. This PR fixes that.

### DIFF
--- a/lib/spack/spack/test/spec_semantics.py
+++ b/lib/spack/spack/test/spec_semantics.py
@@ -1298,4 +1298,11 @@ def test_package_hash_affects_dunder_and_dag_hash(mock_packages, default_mock_co
 def test_satisfies_is_commutative_with_concrete_specs(mock_packages, default_mock_concretization):
     a1 = default_mock_concretization("a@1.0")
     a2 = Spec("a@1.0")
-    assert a1.satisfies(a2) and a2.satisfies(a1)
+
+    # strict=False means non-empty intersection, which is commutative.
+    assert a1.satisfies(a2)
+    assert a2.satisfies(a1)
+
+    # strict=True means set inclusion, which is not commutative.
+    assert a1.satisfies(a2, strict=True)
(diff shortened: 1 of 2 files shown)


lib/spack/spack/test/spec_semantics.py
```diff
@@ -1293,3 +1293,16 @@ def test_package_hash_affects_dunder_and_dag_hash(mock_packages, default_mock_co
...
+    a1 = default_mock_concretization("a@1.0")
+    a2 = Spec("a@1.0")
+
+    # strict=False means non-empty intersection, which is commutative.
+    assert a1.satisfies(a2)
+    assert a2.satisfies(a1)
+
+    # strict=True means set inclusion, which is not commutative.
+    assert a1.satisfies(a2, strict=True)
+    assert not a2.satisfies(a1, strict=True)
```
alalazo, line 1308: This doesn't test the initial case we found wrong:
```
Spack version 0.20.0.dev0
Python 3.8.10, Linux x86_64
>>> s = Spec("zlib")
>>> t = Spec("zlib@1.2.13")
>>> s.satisfies(t)
True
>>> t.satisfies(s)
True
>>> t.concretize()
>>> s.satisfies(t)
False
>>> t.satisfies(s)
True
>>> 
```
where `strict=False` and one spec was concrete.

alalazo (approved):

alalazo:
Should we open an issue to explain what was wrong, or is that sufficiently self explanatory?

haampie:
I think the commit message is enough, together with the updated docstring

PR #35600, opened 2023-02-21
Update FEniCSx libraries to v0.6

garth-wells:


chrisrichardson (dismissed):
Seems a straightforward version bump.

tldahlgren (dismissed):
Confirmed the 5 new version sha256.

tldahlgren:
@adamjstewart Did you want to comment since this PR includes Python packages?

var/spack/repos/builtin/packages/py-fenics-basix/package.py
```diff
@@ -34,6 +35,7 @@ class PyFenicsBasix(PythonPackage):
     )
 
     depends_on("fenics-basix@main", type=("build", "run"), when="@main")
+    depends_on("fenics-basix@0.6.0", type=("build", "run"), when="@0.6.0")
     depends_on("fenics-basix@0.5.1", type=("build", "run"), when="@0.5.1")
     depends_on("fenics-basix@0.4.2", type=("build", "run"), when="@0.4.2")
     depends_on("fenics-basix@0.3.0", type=("build", "run"), when="@0.3.0")
```
adamjstewart, line 41: Missing deps:
```toml
requires = ["setuptools>=42", "wheel", "pybind11>=2.9.1"]
```
```python
      install_requires=["numpy>=1.21"],                                                  
```
Wheel is added by the base class, but the others need to be added or updated here.

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -41,6 +42,7 @@ class PyFenicsDolfinx(PythonPackage):
     depends_on("python@3.7:", type=("build", "run"))
```
adamjstewart, line 43: Python 3.8+ is required now

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,20 +51,23 @@ class PyFenicsDolfinx(PythonPackage):
     depends_on("fenics-dolfinx@0.1.0", when="@0.1.0")
 
     depends_on("fenics-basix@main", type=("build", "link"), when="@main")
+    depends_on("fenics-basix@0.6.0:0.6", type=("build", "link"), when="@0.6.0:0.6")
```
adamjstewart, line 55: I don't see where basix is required by this version

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,20 +51,23 @@ class PyFenicsDolfinx(PythonPackage):
...
     depends_on("fenics-basix@0.5.1:0.5", type=("build", "link"), when="@0.5.0:0.5")
     depends_on("fenics-basix@0.4.2", type=("build", "link"), when="@0.4.1")
     depends_on("fenics-basix@0.3.0", type=("build", "link"), when="@0.3.0")
     depends_on("fenics-basix@0.2.0", type=("build", "link"), when="@0.2.0")
     depends_on("fenics-basix@0.1.0", type=("build", "link"), when="@0.1.0")
 
     depends_on("py-fenics-ffcx@main", type="run", when="@main")
+    depends_on("py-fenics-ffcx@0.6.0:0.6", type="run", when="@0.6.0:0.6")
```
adamjstewart:
```suggestion
    depends_on("py-fenics-ffcx@0.6", type="run", when="@0.6")
```
Simpler version of the same thing

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,20 +51,23 @@ class PyFenicsDolfinx(PythonPackage):
...
     depends_on("py-fenics-ffcx@0.5.0.post0", type="run", when="@0.5.0:0.5")
     depends_on("py-fenics-ffcx@0.4.2", type="run", when="@0.4.1")
     depends_on("py-fenics-ffcx@0.3.0", type="run", when="@0.3.0")
     depends_on("py-fenics-ffcx@0.2.0", type="run", when="@0.2.0")
     depends_on("py-fenics-ffcx@0.1.0", type="run", when="@0.1.0")
 
     depends_on("py-fenics-ufl@main", type="run", when="@main")
+    depends_on("py-fenics-ufl@2023.1.1.post0", type="run", when="@0.6.0:0.6")
```
adamjstewart:
```suggestion
    depends_on("py-fenics-ufl@2023.1", type="run", when="@0.6")
```

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,20 +51,23 @@ class PyFenicsDolfinx(PythonPackage):
...
     depends_on("py-fenics-ufl@2022.2.0", type="run", when="@0.5.0:0.5")
     depends_on("py-fenics-ufl@2022.1.0", type="run", when="@0.4.1")
     depends_on("py-fenics-ufl@2021.1.0", type="run", when="@0.1:0.3.99")
```
adamjstewart: Where does the lower bound for pybind11 come from?

adamjstewart: All of these deps should be type=build/run

var/spack/repos/builtin/packages/py-fenics-ffcx/package.py
```diff
@@ -42,11 +43,13 @@ class PyFenicsFfcx(PythonPackage):
...
+    depends_on("py-fenics-ufl@2023.1.1.post0", type="run", when="@0.6.0:0.6")
     depends_on("py-fenics-ufl@2022.2.0", type="run", when="@0.5.0:0.5")
     depends_on("py-fenics-ufl@2022.1.0", type="run", when="@0.4.2")
     depends_on("py-fenics-ufl@2021.1.0", type="run", when="@0.1.0:0.3")
 
     depends_on("py-fenics-basix@main", type="run", when="@main")
+    depends_on("py-fenics-basix@0.6.0:0.6", type="run", when="@0.6.0:0.6")
     depends_on("py-fenics-basix@0.5.1:0.5", type="run", when="@0.5.0:0.5")
     depends_on("py-fenics-basix@0.4.2", type="run", when="@0.4.2")
     depends_on("py-fenics-basix@0.3.0", type="run", when="@0.3.0")
```
adamjstewart: All deps (including setuptools) should be type=build/run

var/spack/repos/builtin/packages/py-fenics-ffcx/package.py
```diff
@@ -42,11 +43,13 @@ class PyFenicsFfcx(PythonPackage):
     depends_on("py-numpy", type="run")
 
     depends_on("py-fenics-ufl@main", type="run", when="@main")
+    depends_on("py-fenics-ufl@2023.1.1.post0", type="run", when="@0.6.0:0.6")
```
adamjstewart:
```suggestion
    depends_on("py-fenics-ufl@2023.1", type="run", when="@0.6")
```

var/spack/repos/builtin/packages/py-fenics-ufl/package.py
```diff
@@ -19,6 +19,9 @@ class PyFenicsUfl(PythonPackage):
     maintainers("chrisrichardson", "garth-wells", "jhale")
 
     version("main", branch="main")
+    version(
+        "2023.1.1.post0", sha256="9e6e87f1447635029cec42604f62a76bba84899beb4b8822af10389d1f93a9b6"
+    )
     version("2022.2.0", sha256="d6e18e06df5d7a626c3138d49a543914d68186afb6159c4d1a7cd72b2a199b02")
     version("2022.1.0", sha256="48359903d47fb397900d105fe4a60b459c50bbf9d9da78beb9accb54e4e4acaf")
     version("2021.1.0", sha256="130fdc09bb7fcd39dcd2618426912b8a25a03431d94575711068b38c666b4337")
```
adamjstewart, line 27: Missing dep on setuptools 58+

adamjstewart (changes requested):
May help to read through https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html to see where these deps come from.

I didn't check any of the CMake packages.

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
garth-wells: Updated - version requirement in the CMakefile.txt that builds the wrapper interface.

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
garth-wells: It's in the `CMakeLists.txt` that builds the wrapper interface. Versioning is a bit sloppy in the `CMakeLists.txt` file.

garth-wells:
Thanks @tldahlgren and @adamjstewart. I think I've addressed all the comments.

tldahlgren (commented):
Confirmed the seven affected version sha256, one per package.

garth-wells:
Our spack install test is now failing after review changes, so please don’t merge yet.

garth-wells:
False alarm, CI failure was a hardware issue.

@tldahlgren, @adamjstewart anything else we need to do?

tldahlgren:
> False alarm, CI failure was a hardware issue.
> 
> @tldahlgren, @adamjstewart anything else we need to do?

@adamjstewart Will need to re-review since he needs to clear the Requested Changes status.

var/spack/repos/builtin/packages/py-fenics-basix/package.py
```diff
@@ -34,16 +35,18 @@ class PyFenicsBasix(PythonPackage):
...
     depends_on("fenics-basix@0.2.0", type=("build", "run"), when="@0.2.0")
     depends_on("fenics-basix@0.1.0", type=("build", "run"), when="@0.1.0")
 
+    depends_on("py-setuptools@42:", type=("build", "run"))
+    depends_on("py-numpy@1.2.1:", type=("build", "run"))
```
adamjstewart: 1.21, not 1.2.1. Also, these requirements are only for the latest version, not all versions. You'll need to see when these requirements were introduced and add `when=` arguments as needed.

garth-wells: NumPy version number fixed. 

`setuptools` and `numpy` have always been dependencies. Just the versioning in the Python packaging was a bit sloppy in the beta versions that were not on PyPI.

garth-wells:
@adamjstewart I've fixed the NumPy version.

var/spack/repos/builtin/packages/py-fenics-basix/package.py
adamjstewart: So you're saying setuptools 42+, numpy 1.21+, python 3.8+, and pybind11 2.9.1+ have been required since 0.1.0? And that Python 3.7 has never worked?

garth-wells: The would require trawling back over alpha and beta versions, and testing against all versions of dependencies. Hardly seems like a good use of time and effort, especially for deprecated versions.

adamjstewart: No need to test anything, I usually just use whatever was reported in pyproject.toml/setup.py/setup.cfg.

garth-wells: Earlier versions did not include dependencies in `setup.py`, etc. Dependencies were cleaned up in recent versions when creating PyPI packages. Not sure how to move this forward.

ma595:
@adamjstewart Could you advise what we should actually do here? Are you suggesting we fix dependencies (numpy, setuptools, python, pybind11) for previous versions (0.1.0, 0.2.0,...) as well?

Perhaps a compromise is to use `when` i.e. `depends_on("py-setuptools@42:", type=("build", "run"), when="0.4.0:")`, doing the same for python3.8, numpy, pybind11. i.e. fixing those versions for 0.4.0 and higher.  This needs to be done for `py-fenics-dolfinx`, `py-fenics-basix` and probably other fenics components.

Then we should mark earlier versions, 0.1.0, 0.2.0 and 0.3.0 as deprecated.

adamjstewart:
When I review PRs for Python packages, the only method I have for determining dependency requirements is to check the same places that pip does (pyproject.toml, setup.py, setup.cfg, etc.). So I tend to rely heavily on those, and flag things that look incorrect when they disagree with what those files say.

With that said, I only do this because I'm not the developer of the software. Since y'all are actually developers of FEniCS, I trust your knowledge more than what was in these files long ago.

So, as an example, _if_ you have reason to believe that py-fenics-basix requires Python 3.8 and has always required 3.8 (regardless of what the files say), then the current PR is correct. But _if_ Python 3.7 was once supported, this can easily be encoded as:
```python
depends_on("python@3.8:", when="@X.Y:", type=("build", "run"))
```
I can't ask you to test and verify all of these past versions, I'm trusting you to make the right decision. It's best practice to keep dependency requirements as loose as possible so that it doesn't prevent people from building a particular combination if they need to. For example, if someone wants to build py-fenics-basix AND another package, and that other package requires an older version of something, we would like to support that if possible.

If you do decide to keep the stricter requirements that weren't documented in older releases, it would be good to at least add a comment saying that pyproject.toml/setup.py/setup.cfg isn't trustworthy for releases older than X.Y.

ma595:
@adamjstewart, I have attempted to address your comments, could you please re-review? Thanks.

var/spack/repos/builtin/packages/py-fenics-basix/package.py
```diff
@@ -34,16 +35,22 @@ class PyFenicsBasix(PythonPackage):
...
     depends_on("fenics-basix@0.3.0", type=("build", "run"), when="@0.3.0")
     depends_on("fenics-basix@0.2.0", type=("build", "run"), when="@0.2.0")
     depends_on("fenics-basix@0.1.0", type=("build", "run"), when="@0.1.0")
 
+    depends_on("py-setuptools@42:", type=("build", "run"), when="@0.4:")
+    depends_on("py-setuptools@40:", type=("build", "run"))
+    depends_on("py-numpy@1.21:", type=("build", "run"), when="@0.4:")
+    depends_on("py-numpy@1.15:", type=("build", "run"))
     depends_on("cmake@3.19:", type="build")
+    depends_on("python@3.8:", type=("build", "run"), when="@0.4:")
```
adamjstewart: Shouldn't this be 3.7+? https://github.com/FEniCS/basix/blob/main/python/setup.py#L10

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,32 +52,39 @@ class PyFenicsDolfinx(PythonPackage):
...
     depends_on("py-petsc4py", type=("build", "run"))
-    depends_on("py-pybind11@2.6.2:", type=("build", "run"))
+    depends_on("py-pybind11@2.7.0:", type=("build", "run"), when="@0.4:")
+    depends_on("py-pybind11@2.2.4:", type=("build", "run"))
     depends_on("xtensor@0.23.10:", type="build", when="@:0.5")
 
+    depends_on("py-setuptools@42:", type=("build", "run"), when="@0.4:")
+    depends_on("py-setuptools@40:", type=("build", "run"))
+
     depends_on("py-cffi", type="run")
```
adamjstewart:
```suggestion
    depends_on("py-cffi", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -38,9 +39,11 @@ class PyFenicsDolfinx(PythonPackage):
     depends_on("cmake@3.19:", type="build")
     depends_on("hdf5", type="build")
     depends_on("pkgconfig", type=("build", "run"))
```
adamjstewart: pkgconfig isn't usually required at run-time

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,32 +52,39 @@ class PyFenicsDolfinx(PythonPackage):
...
 
     depends_on("py-numpy@1.21:", type=("build", "run"), when="@0.5.0:")
-    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-numpy@1.15:", type=("build", "run"))
 
     depends_on("py-mpi4py", type=("build", "run"))
     depends_on("py-petsc4py", type=("build", "run"))
-    depends_on("py-pybind11@2.6.2:", type=("build", "run"))
+    depends_on("py-pybind11@2.7.0:", type=("build", "run"), when="@0.4:")
+    depends_on("py-pybind11@2.2.4:", type=("build", "run"))
```
adamjstewart, line 80: Is the minimum pybind11 version documented anywhere?

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,32 +52,39 @@ class PyFenicsDolfinx(PythonPackage):
...
     depends_on("xtensor@0.23.10:", type="build", when="@:0.5")
 
+    depends_on("py-setuptools@42:", type=("build", "run"), when="@0.4:")
+    depends_on("py-setuptools@40:", type=("build", "run"))
```
adamjstewart: Is the minimum setuptools version documented anywhere?

var/spack/repos/builtin/packages/py-fenics-ffcx/package.py
```diff
@@ -36,19 +37,23 @@ class PyFenicsFfcx(PythonPackage):
...
-    depends_on("py-fenics-basix@0.4.2", type="run", when="@0.4.2")
-    depends_on("py-fenics-basix@0.3.0", type="run", when="@0.3.0")
-    depends_on("py-fenics-basix@0.2.0", type="run", when="@0.2.0")
-    depends_on("py-fenics-basix@0.1.0", type="run", when="@0.1.0")
+    depends_on("py-setuptools@58:", type=("build", "run"), when="@0.4:")
+    depends_on("py-setuptools@40:", type=("build", "run"))
+
+    depends_on("py-cffi", type=("build", "run"))
+    depends_on("py-numpy@1.21:", type=("build", "run"), when="@0.4:")
+    depends_on("py-numpy@1.15:", type=("build", "run"))
```
adamjstewart: Is the numpy min version documented anywhere?

var/spack/repos/builtin/packages/py-fenics-ufl/package.py
```diff
@@ -34,4 +37,7 @@ class PyFenicsUfl(PythonPackage):
     version("2016.2.0", tag="ufl-2016.2.0")
 
     depends_on("python@3.7:", type=("build", "run"))
-    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-setuptools@58:", type=("build", "run"), when="@2022.1.0:")
+    depends_on("py-setuptools@40:", type=("build", "run"))
+    depends_on("py-numpy@1.21:", type=("build", "run"), when="@2022.1.0:")
+    depends_on("py-numpy@1.15:", type=("build", "run"))
```
adamjstewart: Is the numpy min version documented anywhere?

adamjstewart (commented):
Few more things I saw that disagreed with pyproject.toml/setup.py/setup.cfg in the latest versions. Feel free to ignore them if these files still aren't correct in the latest release.

var/spack/repos/builtin/packages/py-fenics-ffcx/package.py
ma595: My mistake. 1.21 is the min version. Versions lower do not work with py3.7+

var/spack/repos/builtin/packages/py-fenics-ufl/package.py
ma595: My mistake. 1.21 is the min version. Versions lower do not work with py3.7+

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
ma595: I'm not aware that it is.

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
ma595: It hasn't. Advice on this would be appreciated.

adamjstewart: If you know that this is the correct min version, then it's fine to include in the package. I would suggest updating the fenics source code to correctly document all dependencies someday.

var/spack/repos/builtin/packages/py-fenics-basix/package.py
```diff
@@ -34,16 +35,20 @@ class PyFenicsBasix(PythonPackage):
...
     depends_on("fenics-basix@main", type=("build", "run"), when="@main")
+    depends_on("fenics-basix@0.6.0", type=("build", "run"), when="@0.6.0")
     depends_on("fenics-basix@0.5.1", type=("build", "run"), when="@0.5.1")
     depends_on("fenics-basix@0.4.2", type=("build", "run"), when="@0.4.2")
     depends_on("fenics-basix@0.3.0", type=("build", "run"), when="@0.3.0")
     depends_on("fenics-basix@0.2.0", type=("build", "run"), when="@0.2.0")
     depends_on("fenics-basix@0.1.0", type=("build", "run"), when="@0.1.0")
 
+    depends_on("py-setuptools@42:", type=("build", "run"), when="@0.4:")
+    depends_on("py-setuptools@40:", type=("build", "run"))
```
adamjstewart: I don't think setuptools is required at run-time

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,32 +52,38 @@ class PyFenicsDolfinx(PythonPackage):
...
     depends_on("py-mpi4py", type=("build", "run"))
     depends_on("py-petsc4py", type=("build", "run"))
-    depends_on("py-pybind11@2.6.2:", type=("build", "run"))
+    depends_on("py-pybind11@2.7.0:", type=("build", "run"), when="@0.4:")
+    depends_on("py-pybind11@2.2.4:", type=("build", "run"))
     depends_on("xtensor@0.23.10:", type="build", when="@:0.5")
 
-    depends_on("py-cffi", type="run")
+    depends_on("py-setuptools@42:", type=("build", "run"), when="@0.4:")
+    depends_on("py-setuptools@40:", type=("build", "run"))
```
adamjstewart: I don't think setuptools is required at run-time

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
```diff
@@ -49,32 +52,38 @@ class PyFenicsDolfinx(PythonPackage):
...
     depends_on("py-fenics-ffcx@0.4.2", type="run", when="@0.4.1")
     depends_on("py-fenics-ffcx@0.3.0", type="run", when="@0.3.0")
     depends_on("py-fenics-ffcx@0.2.0", type="run", when="@0.2.0")
     depends_on("py-fenics-ffcx@0.1.0", type="run", when="@0.1.0")
 
     depends_on("py-fenics-ufl@main", type="run", when="@main")
-    depends_on("py-fenics-ufl@2022.2.0", type="run", when="@0.5.0:0.5")
-    depends_on("py-fenics-ufl@2022.1.0", type="run", when="@0.4.1")
-    depends_on("py-fenics-ufl@2021.1.0", type="run", when="@0.1:0.3.99")
+    depends_on("py-fenics-ufl@2023.1", type="run", when="@0.6")
```
adamjstewart: Shouldn't these all be type=build/run?

adamjstewart (commented):
Only 2 packages left that I have questions about, some of the dependency types don't seem correct to me.

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
ma595: These are only required at runtime.

var/spack/repos/builtin/packages/py-fenics-basix/package.py
ma595: fixed

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
ma595: fixed

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
adamjstewart: Won't the import tests (`spack install --test=run py-fenics-dolfinx`) require these to be in `PYTHONPATH` in order to pass?

var/spack/repos/builtin/packages/py-fenics-ffcx/package.py
```diff
@@ -37,18 +38,21 @@ class PyFenicsFfcx(PythonPackage):
...
-    depends_on("py-fenics-ufl@2022.1.0", type="run", when="@0.4.2")
-    depends_on("py-fenics-ufl@2021.1.0", type="run", when="@0.1.0:0.3")
-
-    depends_on("py-fenics-basix@main", type="run", when="@main")
-    depends_on("py-fenics-basix@0.5.1:0.5", type="run", when="@0.5.0:0.5")
-    depends_on("py-fenics-basix@0.4.2", type="run", when="@0.4.2")
-    depends_on("py-fenics-basix@0.3.0", type="run", when="@0.3.0")
-    depends_on("py-fenics-basix@0.2.0", type="run", when="@0.2.0")
-    depends_on("py-fenics-basix@0.1.0", type="run", when="@0.1.0")
+    depends_on("py-setuptools@40:", type="build")
```
adamjstewart: Setuptools is also required at run-time

var/spack/repos/builtin/packages/py-fenics-ufl/package.py
```diff
@@ -34,4 +37,6 @@ class PyFenicsUfl(PythonPackage):
     version("2016.2.0", tag="ufl-2016.2.0")
 
     depends_on("python@3.7:", type=("build", "run"))
-    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-setuptools@58:", type="build", when="@2022.1.0:")
+    depends_on("py-setuptools@40:", type="build")
```
adamjstewart: Setuptools is required at run-time too

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-fenics-ffcx/package.py
ma595: done

var/spack/repos/builtin/packages/py-fenics-ufl/package.py
ma595: done

var/spack/repos/builtin/packages/py-fenics-dolfinx/package.py
ma595: If I understand `spack install --test=root` correctly I agree that the `build` option makes sense in this case, but I don't think any tests are currently implemented (apart from default import tests?). This is something I intend to do soon. As far as I'm aware, testing isn't doing what we want currently, so I have held back (https://github.com/spack/spack/issues/29447), but I could probably write some tests that run the C++ examples (which do need ffcx).

adamjstewart: Yes, only the import tests are implemented by default.

ma595: I will keep this in mind for a forthcoming PR. Can v0.6.0 be merged, or shall I amend this now in anticipation of future tests?

adamjstewart: The import tests already require these to be present at build time, so I don't see why we shouldn't make them build/run deps now.

adamjstewart (approved):

ma595:
Can this be merged? Some tests are failing but these are due to docker pull failing...

adamjstewart:
Keep running "@spackbot run pipeline" until it passes, might be days, might be weeks 🤷‍♂️ 

@scottwittenburg

PR #35607, opened 2023-02-21
add prefix search for CCE

lukebroskop:
add prefix search for CCE

### DIFF
--- a/lib/spack/spack/compilers/cce.py
+++ b/lib/spack/spack/compilers/cce.py
@@ -4,8 +4,8 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import os
 
-from spack.compiler import Compiler, UnsupportedCompilerFlag
 import spack.util.executable
+from spack.compiler import Compiler, UnsupportedCompilerFlag
 from spack.version import ver
 
 
--- a/lib/spack/spack/compilers/cce.py
+++ b/lib/spack/spack/compilers/cce.py
@@ -154,6 +154,7 @@ class Cce(Compiler):
             for line in cce_output.splitlines():
                 if line.startswith("programs:"):
                     cce_prefix = line.split(":")[1].strip()
+                    cce_prefix = cce_prefix.strip('=')
                     return cce_prefix
 
             raise RuntimeError(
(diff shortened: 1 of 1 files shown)


lib/spack/spack/compilers/cce.py
```diff
@@ -141,3 +142,20 @@ def stdcxx_libs(self):
...
+        # CCE craycc/crayCC (cc/CC) reports its install prefix when running `-print-search-dirs`
+        # crayftn/ftn should be installed at the same location as the C/C++ compilers
+        cxx = spack.util.executable.Executable(self.cxx)
+        with self.compiler_environment():
+            cce_output = cxx("-print-search-dirs", output=str, error=str)
+
+            for line in cce_output.splitlines():
+                if line.startswith("programs:"):
+                    cce_prefix = line.split(":")[1].strip()
+                    return cce_prefix
```
skosukhin: Ok. I haven't checked yet how this method is used and why my review was requested but here are the outputs of `CC -print-search-dirs` on two different Cray machines:

```console
programs: =/opt/cray/pe/cce/14.0.2/cce-clang/x86_64/share/../bin:/opt/cray/pe/cce/14.0.2/cce-clang/x86_64/bin
libraries: =/opt/cray/pe/cce/14.0.2/cce-clang/x86_64/lib/clang/14.0.6:/lib/../lib64:/usr/lib/../lib64:/lib:/usr/lib
```
```console
programs: =/opt/cray/pe/cce/12.0.3/cce-clang/x86_64/share/../bin:/opt/cray/pe/cce/12.0.3/cce-clang/x86_64/bin:/opt/gcc/8.1.0/snos/lib/gcc/x86_64-suse-linux/8.1.0/../../../../x86_64-suse-linux/bin
libraries: =/opt/cray/pe/cce/12.0.3/cce-clang/x86_64/lib/clang/12.0.0:/opt/gcc/8.1.0/snos/lib/gcc/x86_64-suse-linux/8.1.0:/opt/gcc/8.1.0/snos/lib/gcc/x86_64-suse-linux/8.1.0/../../../../lib64:/lib/../lib64:/usr/lib/../lib64:/opt/gcc/8.1.0/snos/lib/gcc/x86_64-suse-linux/8.1.0/../../..:/opt/cray/pe/cce/12.0.3/cce-clang/x86_64/bin/../lib:/lib:/usr/lib
```
and it looks like the current implementation will return wrong results in both cases because it does not account for the leading `=`. Also, it might make sense to normalize the path before returning it.

PR #35609, opened 2023-02-22
Add CCE support to hipfort

lukebroskop:
Add CCE support to hipfort

### DIFF
--- a/var/spack/repos/builtin/packages/hipfort/package.py
+++ b/var/spack/repos/builtin/packages/hipfort/package.py
@@ -97,6 +97,8 @@ class Hipfort(CMakePackage):
 
     depends_on("rocm-cmake@3.8.0:", type="build")
 
+    depends_on("binutils", when="%cce")
+
     for ver in [
         "3.8.0",
         "3.9.0",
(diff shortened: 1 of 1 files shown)


cgmb:
Is this a workaround for a problem in the hipfort build, or is this the normal set of variables that you have to specify to build any library with CCE?

lukebroskop:
This is a normal set of variables that are needed to override the variables set by the hipfort build. Otherwise the build will fail as the Cray fortran compiler does not accept the same flags that are defaulted for the GCC build. 

From:
https://github.com/ROCmSoftwarePlatform/hipfort/blob/6f6ae98ed45df9a66e48121ec7c0cb8f0fbd6137/CMakeLists.txt#L9-L15
```
# Default compiler gfortran, otherwise cmake Fortran package will set it to f95
# You can override the compiler with cmake -DHIPFORT_COMPILER=
set(HIPFORT_COMPILER "/usr/bin/gfortran" CACHE STRING "Fortran Compiler to build HIPFORT")
set(HIPFORT_AR "/usr/bin/gcc-ar" CACHE STRING "Static archive command")
set(HIPFORT_RANLIB "/usr/bin/gcc-ranlib" CACHE STRING "ranlib used to create Static archive")
set(HIPFORT_COMPILER_FLAGS "-ffree-form -cpp -ffree-line-length-none -fmax-errors=5" CACHE STRING "Compiler flags to build HIPFORT")
```

cgmb (approved):
LGTM. Though, it seems weird to me that hipfort uses `HIPFORT_COMPILER` rather than `CMAKE_Fortran_COMPILER`.

tldahlgren (approved):
LGTM

PR #27429, opened 2021-11-12
mpich: fix macos problem with `-flat-namespace`

albestro:
After a long period working around the problem with `flat-namespace` flag in MPICH wrappers, I've seen the light thanks to @eschnett!

https://github.com/pmodels/mpich/issues/5530

As reported in the above issue `--enable-two-level-namespace` is the solution on macOS.

PR #35611, opened 2023-02-22
mpich: fix macos problem with -flat-namespace

albestro:
See discussion in #27429, it is just applying what has been decided there.

I had to create a new one due to a mess I did with the branch on my fork.

### DIFF
--- a/var/spack/repos/builtin/packages/mpich/package.py
+++ b/var/spack/repos/builtin/packages/mpich/package.py
@@ -98,9 +98,9 @@ spack package at this time.""",
     variant("fortran", default=True, description="Enable Fortran support")
 
     variant(
-        "two_level_namespace",
-        default=False,
-        description="""Build shared libraries and programs
+        'two_level_namespace',
+        default=(sys.platform == 'darwin'),
+        description='''Build shared libraries and programs
 built with the mpicc/mpifort/etc. compiler wrappers
 with '-Wl,-commons,use_dylibs' and without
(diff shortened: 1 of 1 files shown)


alalazo:
@eschnett fyi

alalazo (dismissed):

var/spack/repos/builtin/packages/mpich/package.py
```diff
@@ -491,6 +482,10 @@ def configure_args(self):
             "--with-yaksa={0}".format(spec["yaksa"].prefix if "^yaksa" in spec else "embedded"),
         ]
 
+        # see https://github.com/pmodels/mpich/issues/5530
+        if sys.platform == "darwin":
```
haampie:
```suggestion
        if spec.platform == "darwin":
```

haampie (dismissed):

alalazo (approved):

raffenet (dismissed):

alalazo (approved):

naromero77:
Can someone check why the CI pipeline failed?

naromero77:
@spackbot commands

naromero77:
@spackbot run pipeline

alalazo (approved):

raffenet (commented):
My only concern is that this may cause issues with older versions. We only switched the default in MPICH 4.1. Maybe add a condition on the version being built?

PR #35614, opened 2023-02-22
gromacs: add sve variant on aarch64

annop-w:


junghans (dismissed):

marvinbernhardt (approved):
lgtm

Edit (sorry, not fully awake yet): We don't have variants for all the other SIMD settings. Maybe we shouldn't introduce one for this one and just enable when available.

annop-w:
> lgtm
> 
> Edit (sorry, not fully awake yet): We don't have variants for all the other SIMD settings. Maybe we shouldn't introduce one for this one and just enable when available.

My concern is that forcing SVE on SVE enabled machines might not always be the best option. Having this as a variant gives users this flexibility to test between NEON and SVE. The variant is anyway enabled only on V1 and A64FX and it should not show up with `spack info gromacs` on other machines (although testing on N1 reveals that it does which should be an incorrect behaviour, on M2 mac it does not show up). What do you think ?

marvinbernhardt:
(I have no experience with ARM NEON or SVE)

There are [more specific targets](https://spack.readthedocs.io/en/latest/packaging_guide.html#querying-the-target-microarchitecture), that could be used, e.g.:
```
ARM - aarch64
    cortex_a72  neoverse_n1  neoverse_v1

Apple - aarch64
    m1  m2
```
Since we currently don't let the user decide which SIMD to use on non-ARM it might be best to keep that pattern. Would it be possible to use the specific architecture (e.g. `m2`) to decide between NEON and SVE?

annop-w:
As I mentioned not adding it as a variant means that on SVE enabled targets, which are `neoverse_v1` and `a64fx` available in spack currently, we force users to use SVE on these two targets. In my opinion, this could mean that users do not get the best performance possible on these two targets because the SVE version might not be as optimized as the NEON version ?? Having the variant at least allows users to test them out.

marvinbernhardt:
In general, the same could be argued for some x64 processors, where sometimes the highest SIMD instruction set is not optimal. But probably this is a larger issue on ARM, justifying an own variant. So in the end, I'm ok with merging.

annop-w:
@spackbot run pipeline

annop-w:
@alalazo Could you please take a look at this one too ? I think this is ready to go in. Thank you.

alalazo (approved):

PR #35618, opened 2023-02-22
mpifileutils: add DAOS variant

koysean:
Adds variant for DAOS support (see https://mpifileutils.readthedocs.io/en/v0.11.1/build.html#daos-support).

### DIFF
--- a/var/spack/repos/builtin/packages/mpifileutils/package.py
+++ b/var/spack/repos/builtin/packages/mpifileutils/package.py
@@ -68,6 +68,8 @@ class Mpifileutils(Package):
     variant("experimental", default=False, description="Install experimental tools")
     conflicts("+experimental", when="@:0.6")
 
+    variant("daos", default=False, description="Enable DAOS support", when="@0.11:")
+
     def flag_handler(self, name, flags):
         spec = self.spec
         iflags = []
@@ -101,6 +103,9 @@ class Mpifileutils(Package):
             args.append("-DENABLE_EXPERIMENTAL=ON")
         else:
             args.append("-DENABLE_EXPERIMENTAL=OFF")
+
+        if "+daos" in self.spec:
+            args.append("-DENABLE_DAOS=ON")
 
         return args
 
--- a/var/spack/repos/builtin/packages/mpifileutils/package.py
+++ b/var/spack/repos/builtin/packages/mpifileutils/package.py
@@ -50,6 +50,8 @@ class Mpifileutils(Package):
 
     depends_on("attr", when="@0.11.1:+xattr")
 
+    depends_on("daos", when="+daos")
+
     depends_on("bzip2")
 
     depends_on("libcap")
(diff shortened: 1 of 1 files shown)


tldahlgren:
According to the linked docs, you also need to depend on `daos` so it is installed.  I don't see a `daos` package so that would need to be added so Spack can build it.  

Then you can reference the path to its installation in the required ` -DWITH_DAOS_PREFIX=</path/to/daos/>` option that is included in the docs.

tldahlgren (changes requested):
This looks good but is incomplete according to the linked documentation.  See prior comments.

kwryankrattiger:
Since this has been sitting for awhile, and I am now hitting this in the wild in an upstream I need to depend on, I have taken it into my own hands to add the missing parts.

This is getting lightly tested on a system with DAOS to ensure I didn't break anything and should be good to merge once CI is passing.

var/spack/repos/builtin/packages/mpifileutils/package.py
```diff
@@ -102,6 +106,12 @@ def cmake_args(self):
         else:
             args.append("-DENABLE_EXPERIMENTAL=OFF")
 
+        if "+daos" in self.spec:
```
tldahlgren: Is there a chance that another package that might become a dependency of this one could have a variant with this name?

We generally recommend using `self.spec.satisifies()` if the goal is to check that the variant applies to the current spec.  See the docs at https://spack.readthedocs.io/en/latest/packaging_guide.html#testing-spec-constraints:

"The difference between the two functions is that `satisfies()` tests whether spec constraints overlap at all, while `in` tests whether a spec or any of its dependencies satisfy the provided spec."

If `in` has the desired behavior, then I'll merge as is.

var/spack/repos/builtin/packages/mpifileutils/package.py
```diff
@@ -102,6 +106,12 @@ def cmake_args(self):
...
+            args.append("-DENABLE_DAOS=ON")
+            args.append("-DWITH_DAOS_PREFIX=%s" % self.spec["daos"].prefix)
+        else:
+            args.append("-DENABLE_DAOS=OFF")
```
tldahlgren, line 113: Also,  more as  an FYI in case you weren't aware, CMakePackage has some handy helper methods like `self.define_from_variant()`, described at https://spack.readthedocs.io/en/latest/build_systems/cmakepackage.html#adding-flags-to-cmake.

tldahlgren (dismissed):
LGTM.  You might want to consider a couple of things I noted below.  Let me know if you want to merge as-is.

var/spack/repos/builtin/packages/mpifileutils/package.py
kwryankrattiger: I did it this way because it is consistent with the rest of the package, I generally try to stick with whatever the maintainer was using.

tldahlgren (approved):
LGTM

koysean:
Thanks for pushing this through @kwryankrattiger!

tldahlgren:
@koysean For future reference, if you aren't getting a response to a PR in a couple of weeks, it's good to post to the #pull-requests Spack slack channel and even ping the reviewer(s) in a comment here.  Even if you have to do it weekly (since we tend to get a LOT of email from github).

PR #35623, opened 2023-02-22
buildcache create: reproducible tarballs

haampie:
Currently `spack buildcache create` creates compressed tarballs that
differ between each invocation (even when nothing is modified in the
prefix), thanks to:

1. The gzip header containing mtime set to time.time()
2. tar.add() copying stat info (uid/gid/uname/gname/mtime). 

To avoid this, construct GZipFile explicitly (since the Python API doesn't
expose the mtime arg) and update the tarinfo entries.

Notice that for distribution (spack buildcache create), there is no point
in recording the the user and group ids and names, since they don't
exist on other systems. In fact, that data is only ever used when python
is run with privileges, since only then it can try and restore the
ownership. I think it's a really bad idea to have this in Spack, in fact it
may cause problems when extracting a binary tarball in a docker
container where the user is root, and Python fails to change user ids
because the users don't exist there. (We've seen this in the past for
source tarballs actually, so we changed to --no-same-owner or smth). 

To fix the ownership issue: set uid/gid to 0, meaning that if run under
elevated privileges, the user is "changed" to root, i.e. not changed at
all.

Lastly, in binary distribution, we should only ever add dirs, files, symlinks
and hardlinks to the tarball, not character/block devices and fifos, so
this change explicitly drops those types of entries.

For file modes, normalize like `git`: `0o644` for files/links that are not
executable by user, and `0o755` otherwise (including dirs, symlinks).
This also prevents surprises when extracting a tarball with `tar xf`,
since users wouldn't have to worry (as much) about newly created
files being world-writable and executable.

A test is added that checks for bitwise reproducibility of a compressed
tarball, even when the timestamps on disk are altered. Also a test is
added to check file mode normalization.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -21,6 +21,7 @@ import urllib.parse
 import urllib.request
 import warnings
 from contextlib import closing
+from gzip import GzipFile
 from urllib.error import HTTPError, URLError
 
 import ruamel.yaml as yaml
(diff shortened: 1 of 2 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -1139,6 +1139,66 @@ def generate_key_index(key_prefix, tmpdir=None):
...
+def gzip_compressed_tarfile(path):
+    """Create a reproducible, compressed tarfile"""
+    # Create gzip compressed tarball of the install prefix
+    # 1) Use explicit empty filename and mtime 0 for gzip header reproducibility
+    # 2) On AMD Ryzen 3700X and an SSD disk, we have the following on compression speed:
+    # compresslevel=6 gzip default: llvm takes 4mins, roughly 2.1GB
+    # compresslevel=9 python default: llvm takes 12mins, roughly 2.1GB
+    # So we follow gzip.
+    with open(path, "wb") as fileobj, closing(
+        GzipFile(filename="", mode="wb", compresslevel=6, mtime=0, fileobj=fileobj)
```
alalazo, line 1154: Why do we set `filename` here? Wouldn't it be sufficient to:
```suggestion
        GzipFile(fileobj=fileobj, mode="wb", compresslevel=6, mtime=0)
```
?

alalazo (dismissed):

lib/spack/spack/binary_distribution.py
alalazo: Update: discussed in DMs. This is needed to avoid using `fileobj.name`

https://github.com/python/cpython/commit/68de37973f9b2bd01ba91b10e01fb207396b145f

alalazo: @haampie This LGTM. I'll merge as soon as we add a comment on this gotcha, to avoid forgetting why it was necessary in the future.

alalazo (approved):

PR #35632, opened 2023-02-22
buildcache push: ensure bool arguments for include_*

haampie:
`bindist.push(...)` args were not correctly forwarded after a small API change
that should've made the arguments more explicit. Instead it caused dependencies
to be pushed too, where just the root node should've been.

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -1332,7 +1332,7 @@ def nodes_to_be_packaged(specs, root=True, dependencies=True):
         return list(filter(packageable, nodes))
 
 
-def push(specs, push_url, include_root=True, include_dependencies=True, **kwargs):
+def push(specs, push_url, include_root: bool = True, include_dependencies: bool = True, **kwargs):
     """Create a binary package for each of the specs passed as input and push them
     to a given push URL.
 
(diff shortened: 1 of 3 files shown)


lib/spack/spack/binary_distribution.py
```diff
@@ -1345,6 +1345,10 @@ def push(specs, push_url, include_root=True, include_dependencies=True, **kwargs
         **kwargs: TODO
 
     """
+    # Be explicit about the arugment type
+    if type(include_root) != bool or type(include_dependencies) != bool:
+        raise ValueError("Expected include_root/include_dependencies to be True/False")
```
alalazo, line 1350: Curious if we would have caught this with an annotated function, using `mypy`.

haampie: It doesn't :(

scottwittenburg (approved):
Yes, this makes sense, thanks @haampie!  What a relief to have sanity restored.

lib/spack/spack/binary_distribution.py
tgamblin: is this needed at all now that we have annotations and `mypy`?  Or is this just a belt and suspenders thing after all that CI chaos?

tgamblin: One other thought: I'm pretty positive this would've been caught if we checked the whole codebase with `mypy`.  Right now we only check annotated functions. I think we should try to get to checking everything to prevent subtle stuff like this -- the codebase is large enough that it can save us a lot of pain.

Specifically: I think we should get to where we can turn on `--check-untyped-defs`

adamjstewart: This should prob use isinstance instead of type

PR #35640, opened 2023-02-23
windows: use `sys.platform == "win32"` instead of `is_windows`

tgamblin:
`mypy` only understands `sys.platform == "win32"`, not indirect assignments of that value to things like `is_windows`.  See https://github.com/python/mypy/issues/9242#issuecomment-667586397.

If we don't use the accepted platform checks, `mypy` registers many Windows-only symbols as not present on Linux, when it should skip the checks for platform-specific code. 

I'd like to get us to where we can enable `check_untyped_defs` in `mypy` (see #35639), so I've pulled this off of #35639 and made it a first PR.  I realize this may look uglier to some people, but I don't see a good way around it as long as `mypy` is going to be dumb.

### DIFF
--- a/lib/spack/spack/util/windows_registry.py
+++ b/lib/spack/spack/util/windows_registry.py
@@ -13,8 +13,7 @@ from contextlib import contextmanager
 
 from llnl.util import tty
 
-is_windows = sys.platform == "win32"
-if is_windows:
+if sys.platform == "win32":
     import winreg
 
 
@@ -154,7 +153,7 @@ class WindowsRegistryView(object):
                          to get an entrypoint, the HKEY constants are always open, or an already
                          open key can be used instead.
         """
-        if not is_windows:
+        if sys.platform != "win32":
             raise RuntimeError(
                 "Cannot instantiate Windows Registry class on non Windows platforms"
             )
@@ -167,7 +166,7 @@ class WindowsRegistryView(object):
         try:
             yield
         except FileNotFoundError as e:
-            if e.winerror == 2:
+            if sys.platform == "win32" and e.winerror == 2:
                 tty.debug("Key %s at position %s does not exist" % (self.key, str(self.root)))
             else:
                 raise e
@@ -182,7 +181,7 @@ class WindowsRegistryView(object):
                 winreg.OpenKeyEx(self.root.hkey, self.key, access=winreg.KEY_READ),
             )
         except FileNotFoundError as e:
-            if e.winerror == 2:
+            if sys.platform == "win32" and e.winerror == 2:
                 self._reg = -1
                 tty.debug("Key %s at position %s does not exist" % (self.key, str(self.root)))
             else:
--- a/lib/spack/llnl/util/filesystem.py
+++ b/lib/spack/llnl/util/filesystem.py
@@ -16,7 +16,6 @@ import stat
 import sys
 import tempfile
 from contextlib import contextmanager
-from sys import platform as _platform
 from typing import Callable, List, Match, Optional, Tuple, Union
 
 from llnl.util import tty
@@ -26,9 +25,7 @@ from llnl.util.symlink import islink, symlink
 from spack.util.executable import Executable, which
 from spack.util.path import path_to_os_path, system_path_filter
 
-is_windows = _platform == "win32"
-
-if not is_windows:
+if sys.platform != "win32":
     import grp
     import pwd
 else:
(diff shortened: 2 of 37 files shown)


alalazo:
For reference #34779 is loosely related to the `sys.platform =="win32"` issue[^1], and orthogonal to the change here.

[^1]: It gathers platform specific changes together into their own "private" folders.

scheibelp:
I wanted the check to be abstracted into a method in case multiple things would be "equivalently-Windows" from a Spack point of view. No such case exists yet (although the thread you linked is an example of treating "cygwin" and "win32" the same), but that could be handled later. However, when it comes to that, we run into the problem again.

I don't think it's a major problem here but I'm generally concerned about `mypy` considerations negatively affecting the API. Is there some way to give mypy extra information about this?

Side note: my concern in https://github.com/spack/spack/pull/34779#issuecomment-1442316077 does not apply to this PR.

tgamblin:
There is apparently no way to get `mypy` to understand anything but the checks documented in [PEP 484](https://peps.python.org/pep-0484/#version-and-platform-checking), which are intentionally "simple".  The `mypy` developers don't seem particularly enthused about covering more cases, and someone would have to decide which ones to support, etc. See https://github.com/python/mypy/issues/8547#issuecomment-601813024.

One issue in the current code is that `is_windows` is defined in lots of places, and it's not even used in some files, but that's not caught because it's assigned to a module-scope variable.  So, when making this, I actually removed several unneeded instances of `is_windows = sys.platform == "win32"`.  They probably became unused with refactoring.

IMO this means we need to standardize on the `sys.platform == "win32"` idiom for now -- it's honestly pretty readable and `sys.platform != "win32"` might even be more readable than `not is_windows` (at least to me).

I'd say let's switch to using this, then think about whether more stuff can be consolidated in a checker-friendly way in #34779.

johnwparent:
@tgamblin would `mypy` better handle this case if there was a singular definition for `is_windows` somewhere and each module that actually needed it just imported it? IK that would at least throw an unused import if refactoring removed the need for the `is_windows` statement.
If that would solve the problem, I have a branch implementing something along those lines that's been languishing as I focus on other efforts, but I could spin up on that and PR pretty quickly. Personally I find a lot of `sys.platform == "win32"` clutters up a stanza of code pretty easily/quickly and I've gotten feedback from other community members to that effect as well.

tgamblin:
@johnwparent: unfortunately no, there's no way to make `mypy` smarter about this.

I also find that `sys.platform == "win32"` is ugly, but OTOH I think that does a good job of making you want to find a way to not have to write it :).  I don't think `is_windows` or an `is_windows()` method is necessarily the best way to hide this (and does it really hide it?). I think it would be better if we instead came up with some abstractions to use that worked well in both places, and buried the `if` inside those.

So, really, I think we *have* to do it this way unless `mypy` changes -- I don't see where making the relatively small number of lines here slightly prettier is worth giving up the ability to `mypy` check the whole codebase.

tgamblin:
@scheibelp @johnwparent any additional thoughts?

johnwparent:
From what you're saying about `mypy` it doesn't really seem like there's another option, so not much of a surface to have thoughts on LOL. 
Seems like our hands are tied, so LGTM. 
I'll start mocking out ways to abstract logic like that in a way mypy can understand, do you have a good way to test what mypy can handle/docs? FWIW I think implementing the pathlib refactor in full will resolve a lot of these anyway.

scheibelp (dismissed):

johnwparent (dismissed):

tgamblin:
I think it's really just the stuff in PEP 484 above, with no smart data flow analysis to track the value of the expression.  I suspect it's bc `mypy` is mostly tracking types -- not values -- and they don't want to implement what you'd need to statically analyze the assignment to `is_windows`.

lib/spack/spack/util/windows_registry.py
```diff
@@ -182,7 +181,7 @@ def _load_key(self):
                 winreg.OpenKeyEx(self.root.hkey, self.key, access=winreg.KEY_READ),
             )
         except FileNotFoundError as e:
-            if e.winerror == 2:
+            if sys.platform == "win32" and e.winerror == 2:
```
scheibelp, line 184: Note this extra check is unnecessary because this object cannot be instantiated on non-windows systems

scheibelp (approved):
In all cases except two I see a direct replacement

In one case an extra is-windows check is added which is unnecessary (but does not hurt either)

lib/spack/spack/util/windows_registry.py
tgamblin: It’s another unfortunate mypy thing. The winerror attribute doesn’t exist on non-windows systems but mypy sees it statically, so without this you get an error at typecheck time.

PR #35641, opened 2023-02-23
`colify.py`: Get rid of `**kwargs`

tgamblin:
`colify` is an old module in Spack that still uses `**kwargs` liberally. We should be more explicit.  Doing this eliminates the need for many checks (can't pass the wrong arg if it isn't allowed) and makes the function documentation more clear.

I'm pulling this out of #35639 as I think it helps the code regardless.

- [x] use named arguments instead of `kwargs`


alalazo (approved):
LGTM looking through the diff

PR #35643, opened 2023-02-23
onednn: add variant to use Arm Compute Library on aarch64

annop-w:


var/spack/repos/builtin/packages/onednn/package.py
```diff
@@ -85,12 +87,14 @@ class Onednn(CMakePackage):
         values=("ocl", "none"),
         multi=False,
     )
+    variant("acl", default=False, description="Use Arm Compute Library", when="target=aarch64:")
 
     # https://github.com/oneapi-src/oneDNN#requirements-for-building-from-source
     depends_on("cmake@2.8.11:", type="build")
```
adamjstewart, line 104:
```suggestion
    depends_on("cmake@2.8.12:", when="@2.3:", type="build")
    depends_on("cmake@2.8.11:", type="build")
```

var/spack/repos/builtin/packages/onednn/package.py
```diff
@@ -85,12 +87,14 @@ class Onednn(CMakePackage):
         values=("ocl", "none"),
```
adamjstewart: Can you change `cpu_runtime`'s:
```python
("omp", "tbb", "seq")
```
to:
```python
(
    conditional("none", when="@2.3:"),
    "omp",
    "tbb",
    "seq",
    conditional("threadpool", when="@1.4:"),
    conditional("dpcpp", when="@2:"),
    conditional("sycl", when="@2:"),
)
```
and change `gpu_runtime`'s:
```python
values=("ocl", "none")
```
to:
```python
(
    "ocl",
    "none",
    conditional("dpcpp", when="@2:"),
    conditional("sycl", when="@2:"),
)
```
For future me, I found this using the following script:
```bash
fname=cmake/options.cmake

for tag in $(git tag -l | sort -Vr)
do
    echo
    echo $tag
    echo
    git checkout $tag $fname && grep -B1 'Unsupported .PU runtime' $fname && git restore --staged $fname && git restore $fname
done
```

var/spack/repos/builtin/packages/onednn/package.py
```diff
@@ -85,12 +87,14 @@ class Onednn(CMakePackage):
...
         multi=False,
     )
+    variant("acl", default=False, description="Use Arm Compute Library", when="target=aarch64:")
```
adamjstewart: Any idea what version of onednn added this feature? If not I can try grepping all versions using the script above.

var/spack/repos/builtin/packages/onednn/package.py
```diff
@@ -125,4 +129,11 @@ def cmake_args(self):
         if self.spec.satisfies("gpu_runtime=ocl"):
             args.append("-DOPENCLROOT=" + self.spec["opencl"].prefix)
 
+        if self.spec.satisfies("+acl"):
+            args.append("-DDNNL_AARCH64_USE_ACL=1")
```
adamjstewart:
```suggestion
        args.append(self.define_from_variant("DNNL_AARCH64_USE_ACL", "acl"))
```
This will also explicitly disable it when `~acl`

adamjstewart (changes requested):

var/spack/repos/builtin/packages/onednn/package.py
annop-w: Done.

var/spack/repos/builtin/packages/onednn/package.py
annop-w: I don't know from the top of my head.

adamjstewart:
```suggestion
    variant("acl", default=False, description="Use Arm Compute Library", when="@1.7: target=aarch64:")
```
Looks like 1.7 is the answer.

adamjstewart:
@spackbot fix style

var/spack/repos/builtin/packages/onednn/package.py
```diff
@@ -75,22 +77,35 @@ class Onednn(CMakePackage):
...
+    variant(
+        "acl", default=False, description="Use Arm Compute Library", when="@1.7: target=aarch64:"
+    )
 
     # https://github.com/oneapi-src/oneDNN#requirements-for-building-from-source
+    depends_on("cmake@2.8.12:", when="@2.3:", type="build")
     depends_on("cmake@2.8.11:", type="build")
     depends_on("tbb@2017:", when="cpu_runtime=tbb")
     depends_on("llvm-openmp", when="%apple-clang cpu_runtime=omp")
     depends_on("opencl@1.2:", when="gpu_runtime=ocl")
```
adamjstewart, line 107: Also need to add a few dependencies for these new runtimes:
```python
depends_on("tbb", when="cpu_runtime=sycl")
depends_on("opencl@1.2:", when="gpu_runtime=sycl")
depends_on("openapi-level-zero", when="gpu_runtime=sycl")
```
SYCL also requires `%dpcpp` but there isn't an easy way to express that right now so we can prob just ignore it.

annop-w: Done.

adamjstewart (approved):

PR #35644, opened 2023-02-23
Simpler package API for jobserver support v2

Jordan474:
Simplify packages parallel job count computation, and jobserver support:

- Replace `get_effective_jobs(...supports_jobserver...)` by a `should_support_jobserver` variable in package modules

- The environment variable `SPACK_NO_PARALLEL_MAKE` now affects all packages via `make_jobs` instead of just `MakeExecutable`

Also fix a few packages to support jobserver.

Supersedes #35183 and #35199.

Quick summary about packaging:

- You have nothing to do if you use `make()`, `ninja()`, `CMakePackage`, `MakefilePackage` , or derived
- If you can support jobserver elsewhere: use `should_support_jobserver` to test if you should
- If you don't support jobserver: `make_jobs` is fine

### DIFF
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -536,7 +536,7 @@ def determine_number_of_jobs(
         max_cpus (int or None): maximum number of CPUs available. When None, this
             value is automatically determined.
     """
-    if not parallel:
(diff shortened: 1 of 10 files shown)


alalazo:
> Supersedes https://github.com/spack/spack/pull/35183 and https://github.com/spack/spack/pull/35199

Can we close these other two PRs then?

PR #35648, opened 2023-02-23
hpcviewer: add version 2023.02

mwkrentel:


tldahlgren (dismissed):
Confirmed one of the sha256, rest I'll take on faith. 😉

mwkrentel:
@tldahlgren The reason CI failed is that it tries to build hpctoolkit with
hpcviewer as a dependency.  That fails because the url patch changed
its checksum from gitlab.

I have another PR #35662 which should "fix" the checksum,
except that it won't actually fix it, it will turn that case into a
conflict.  I'll have to see how the concretizer reacts to that.

Anyway, I think I can fix the CI, but I have to fix it by getting the
other PR merged.  But since you turned on auto-merge, there's a decent
chance I can fix it without further action from you.  yeah!

mwkrentel:
The new sha256 sums for hpctoolkit are from  #35662.
But they seem to be blocking this commit.

mwkrentel:
@tldahlgren I fixed the sha256 sums from #35662 and now it passes the CI tests.
If you could re-approve/merge, that would finish the viewer release.
Thanks!

mwkrentel:
@tldahlgren Well, Harmen converted the URL patches to actual patches
in #35711 (while I was asleep).  So, I reverted the last commit here
for updating the checksums.

So, now we're back to the original patch which I expect should pass CI.

tldahlgren (approved):

mwkrentel:
@tldahlgren Ready to merge.   Thanks.

Issue #4927, opened 2017-07-30
Spack compiler wrappers aren't compatible with SCons


adamjstewart:
I'm working on a `SConsPackage` base class and started debugging some SCons builds. I discovered that SCons ignores the `CC` environment variable. If I add `'CC={0}'.format(spack_cc)` to the `serf` build line, it actually uses Spack's compiler wrappers, but I end up seeing the following error message:
```
Spack compiler must be run from Spack! Input 'SPACK_PREFIX' is missing.
```
Maybe SCons purifies the build environment for some reason? We might need a way to work around this. For now I'll tell it to use the actual compilers and manually set `-L` and `-I`.

adamjstewart:
Dug a little deeper into this. It looks like SCons runs all build recipes in an isolated Python environment. By default, this environment doesn't pass any environment variables from the parent process to the subprocess (even `PATH`!). It is up to the package developers to manually pass these environment variables. Some packages like `cantera` have a `env_vars=all` option to pass these environment variables, but others like `serf` don't. I don't think there will be a generic way to fix this.

hyoklee:
@alalazo , any progress on this issue?

alalazo:
Not that I know of. Are you running into the same problem? What is the spec you are trying to build?

hyoklee:
Yes. I'm building [DAOS spec](https://github.com/hyoklee/spack/tree/develop/var/spack/repos/builtin/packages/daos).

adamjstewart:
We managed to patch bazel to pass in Spack's env vars. I wonder if we could do the same for SCons.

PR #35649, opened 2023-02-23
daos: add a new package

hyoklee:
Depends on #35520 

See https://github.com/daos-stack/daos for details.

See also https://github.com/spack/spack/issues/4927


var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,170 @@
...
+    distributed Non Volatile Memory (NVM)."""
+
+    homepage = "https://github.com/daos-stack/daos"
+    git = "https://github.com/daos-stack/daos.git"
+    maintainers("hyoklee")
+
+    version("master", branch="master", submodules=True)
+    version("2.2.0", tag="v2.2.0", submodules=True)
+    version("2.0.2", tag="v2.0.2", submodules=True)
+    version("1.2.0", tag="v1.2.0", submodules=True)
```
soumagne: same comment here, there's no reason I think for having all these versions, I'd just start from 2.0.2 and remove all the rest. That will allow you also to remove all the other patches :) the dependency on cart  and a whole bunch of other stuff.

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,89 @@
...
+    depends_on("openssl")
+    depends_on("pmdk")
+    depends_on("protobuf-c")
+    depends_on("py-distro")
+    depends_on("readline")
+    depends_on("scons@4.4.0:")
+    depends_on("spdk+shared+rdma")
+
+    def build_args(self, spec, prefix):
+        args = ["PREFIX={0}".format(prefix), "--build-deps=yes"]
```
soumagne: you need to remove --build-deps=yes here otherwise that would tell daos to build itself all the dependencies

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,89 @@
...
+    depends_on("libyaml")
+    depends_on("mercury+boostsys")
+    depends_on("mpich")
+    depends_on("openssl")
+    depends_on("pmdk")
+    depends_on("protobuf-c")
+    depends_on("py-distro")
+    depends_on("readline")
+    depends_on("scons@4.4.0:")
+    depends_on("spdk+shared+rdma")
```
soumagne: you probably want to pin to a specific version of spdk, see: https://github.com/daos-stack/daos/blob/v2.2.0/utils/build.config

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,89 @@
...
+    version("master", branch="master", submodules=True)
+    version("2.2.0", tag="v2.2.0", submodules=True)
+    variant(
+        "debug", default=False, description="Enable debugging info and strict compile warnings"
+    )
+
+    depends_on("argobots")
+    depends_on("boost@develop+python", type="build")
+    depends_on("cmocka", type="build")
+    depends_on("dpdk@main")
```
soumagne: same here you probably want to avoid main

var/spack/repos/builtin/packages/daos/package.py
hyoklee: @soumagne , thank you for your review.

Unfortunately, this is required due to DAOS rdb git submodule:

https://github.com/daos-stack/daos/tree/master/src/rdb

There's no Spack package for DAOS rdb. 

Do you have any idea on how to handle this?
I can think of 2 options:

1. Create a new daos-rdb package and use it as a dependent.
2. Add +build_deps variant.

Please advise!

soumagne: I don't understand, yes it's a submodule but it gets built along with daos, there's no need to pass `--build-deps` for that. Do you get an error without it?

hyoklee: Yes. I get an error without it.

soumagne: you should not have any error, which version 2.2 or master ?

hyoklee: I'm using master because of https://github.com/daos-stack/daos/pull/11393.

soumagne: ok thanks I will take a look

soumagne: I don't see any issues with building master or 2.2.0 with spack without passing `--build-deps=yes` so maybe try again or please send me the error log that you get. I would suspect that enabling `--build-deps` is potentially hiding a dependency issue that you have on your end. In any case, it should be disabled.

hyoklee: See https://github.com/hyoklee/spack/actions/runs/4297917161 for the failure.

Here's the error message:

```
2023-03-01T00:45:57.3800340Z BuildRequired: ucx needs to be built, use --build-deps=yes:
2023-03-01T00:45:57.4064304Z   File "/tmp/runner/spack-stage/spack-stage-daos-master-yvkyuhtzz2z4bz2wmeivkwzdnwf7vs2q/spack-src/SConstruct", line 501:
2023-03-01T00:45:57.4065058Z     scons()
2023-03-01T00:45:57.4066145Z   File "/tmp/runner/spack-stage/spack-stage-daos-master-yvkyuhtzz2z4bz2wmeivkwzdnwf7vs2q/spack-src/SConstruct", line 440:
2023-03-01T00:45:57.4066823Z     prereqs.run_build(opts)
2023-03-01T00:45:57.4067926Z   File "/tmp/runner/spack-stage/spack-stage-daos-master-yvkyuhtzz2z4bz2wmeivkwzdnwf7vs2q/spack-src/site_scons/prereq_tools/base.py", line 559:
2023-03-01T00:45:57.4068626Z     self.__env.Clone().require(comp)
2023-03-01T00:45:57.4069796Z   File "/home/runner/work/spack/spack/opt/spack/linux-ubuntu22.04-x86_64_v4/gcc-12.1.0/scons-4.4.0-btcyyrncqhd4g2rnail7oikfbj4hctef/lib/python3.10/site-packages/SCons/Util.py", line 737:
2023-03-01T00:45:57.4070502Z     return self.method(*nargs, **kwargs)
2023-03-01T00:45:57.4071587Z   File "/tmp/runner/spack-stage/spack-stage-daos-master-yvkyuhtzz2z4bz2wmeivkwzdnwf7vs2q/spack-src/site_scons/prereq_tools/base.py", line 807:
2023-03-01T00:45:57.4072252Z     raise error
2023-03-01T00:45:57.4073311Z   File "/tmp/runner/spack-stage/spack-stage-daos-master-yvkyuhtzz2z4bz2wmeivkwzdnwf7vs2q/spack-src/site_scons/prereq_tools/base.py", line 797:
2023-03-01T00:45:57.4074332Z     if comp_def.build(env, needed_libs):
2023-03-01T00:45:57.4075434Z   File "/tmp/runner/spack-stage/spack-stage-daos-master-yvkyuhtzz2z4bz2wmeivkwzdnwf7vs2q/spack-src/site_scons/prereq_tools/base.py", line 1407:
2023-03-01T00:45:57.4076133Z     raise BuildRequired(self.name)
2023-03-01T00:45:57.4360710Z ==> Error: ProcessError: Command exited with status 2:
```

alalazo (changes requested):
```
PKG-DIRECTIVES: 8 issues found
1. daos: unknown package 'isa-l' in 'depends_on' directive
     in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
2. daos: unknown package 'isa-l_crypto' in 'depends_on' directive
     in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
3. daos: unknown package 'spdk' in 'depends_on' directive
     in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
4. daos: dependency on scons@4.4.0: cannot be satisfied by known versions of scons
    happening in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
    known versions of scons are 4.3.0, 4.2.0, 4.1.0.post1, 4.0.1, 4.0.0, 3.1.2, 3.1.1, 3.1.0, 3.0.5, 3.0.4, 3.0.1
5. daos: dependency on dpdk@22.11: cannot be satisfied by known versions of dpdk
    happening in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
    known versions of dpdk are 20.02, [19](https://github.com/spack/spack/actions/runs/4294446033/jobs/7483488342#step:6:20).11, 19.08, 19.05, 19.02
6. daos: dependency on isa-l cannot be satisfied by known versions of isa-l
    happening in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
7. daos: dependency on isa-l_crypto cannot be satisfied by known versions of isa-l_crypto
    happening in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
8. daos: dependency on spdk@[23](https://github.com/spack/spack/actions/runs/4294446033/jobs/7483488342#step:6:24).01:+rdma+shared cannot be satisfied by known versions of spdk
    happening in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
```

hyoklee:
@alalazo I made [separate PRs for dependencies](https://github.com/spack/spack/pulls?q=is%3Apr+is%3Aopen+hyoklee+). If you merge them, there will be no dependency check error.

var/spack/repos/builtin/packages/daos/package.py
soumagne: ah thanks a lot for sending that log. Yes you're missing a dependency on `ucx` in that case (there's already a package for it in spack), eventually we'd need to add a variant for it but I don't think that's possible right now.

hyoklee: Yes, although I include ucx as dependent, Scons doesn't pick it up.

https://github.com/hyoklee/spack/actions/runs/4298976035

I mentioned https://github.com/spack/spack/issues/4927 in the first comment of this PR.
I've spent a lot of time and tried many hacks but Spack and Scons don't get along well.

Thus, I wish Intel DAOS team uses CMake only and remove Scons for DAOS as a build system.

soumagne: :) I see, don't worry I'm pretty sure Scons can pick it up. Indeed scons does not use spack compiler wrappers but instead packages are passed through the `ALT_PREFIX` so you would need to add the ucx dependency and pass `format(spec["ucx"].prefix),` also to `ALT_PREFIX` so that it gets picked up there as well. This will then tell scons about the ucx install and where to look for ucx libraries etc.

hyoklee: Right, I keep forgetting about `ALT_PREFIX`!

soumagne: Yes it's not perfect but it should work. Also you might need to pass the flag `USE_INSTALLED=all` to the args.

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,89 @@
...
+    depends_on("libyaml")
+    depends_on("mercury+boostsys")
+    depends_on("mpich")
+    depends_on("openssl")
+    depends_on("pmdk")
+    depends_on("protobuf-c")
+    depends_on("py-distro")
+    depends_on("readline")
+    depends_on("scons@4.4.0:")
+    depends_on("spdk@23.01:+shared+rdma")
```
soumagne: also just a warning that 23.01 is not officially supported by daos yet, see https://github.com/daos-stack/daos/blob/master/utils/build.config that tells you which version is supported

var/spack/repos/builtin/packages/daos/package.py
hyoklee: `hwloc` fails although I specify it in ALT_PREFIX.

```
Checking targets for component 'hwloc'
Checking for C header file hwloc.h... no
MissingTargets: Package hwloc-devel is required. Check config.log:
  File "/tmp/hyoklee/spack-stage/spack-stage-daos-master-kpcfuqimmjkxzzebduxpr7dh\
zaxfgveu/spack-src/SConstruct", line 501:
    scons()
```

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+
+    depends_on("argobots")
+    depends_on("boost@develop+python", type="build")
+    depends_on("cmocka", type="build")
+    depends_on("dpdk@22.11:")
+    depends_on("go", type="build")
+    depends_on("hwloc")
+    depends_on("isa-l")
+    depends_on("isa-l_crypto")
```
koysean: should be `isa-l-crypto` (also needs fixing under alt_prefix)

var/spack/repos/builtin/packages/daos/package.py
soumagne: could be a bug that needs to be fixed in the daos scons files

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,91 @@
...
+    maintainers("hyoklee")
+
+    version("master", branch="master", submodules=True)
+    version("2.2.0", tag="v2.2.0", submodules=True)
+    variant(
+        "debug", default=False, description="Enable debugging info and strict compile warnings"
+    )
+
+    depends_on("argobots@1.1:")
+    depends_on("boost@develop+python", type="build")
```
soumagne: I don't think we need develop here ? whatever version should be fine ?

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,91 @@
...
+    depends_on("cmocka", type="build")
+    depends_on("dpdk@22.11:")
```
soumagne: I fear it's a little risky to include dpdk separately (maybe it's fine but double check) because spdk already includes dpdk as a submodule so you'd have to make sure it's the same versions I think.

soumagne: if you use the variant I mentioned in the other spdk PR, you won't need that dependency here anymore and instead you'd just do `depends_on("spdk@22.01.1:+shared+rdma+dpdk")`

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,90 @@
...
+            format(spec["pmdk"].prefix),
+            format(spec["protobuf-c"].prefix),
+            format(spec["spdk"].prefix),
+            format(spec["ucx"].prefix),
+        ]
+        alt_prefix_clean = []
+        for i in alt_prefix:
+            if i != "/usr":
+                alt_prefix_clean.append(i)
+                alt_prefix_clean.append("/usr")
```
koysean: Should this line be unindented two levels? Right now it appends `/usr` after every path in alt_prefix.

hyoklee:
@alalazo @soumagne @tldahlgren would you please merge this?

soumagne (dismissed):

soumagne:
this one looks good to me now

tldahlgren:
Still getting audit issue below so noted dependency.

```
PKG-DIRECTIVES: 2 issues found
1. daos: unknown package 'spdk' in 'depends_on' directive
     in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
2. daos: dependency on spdk@22.01.1:+dpdk+rdma+shared cannot be satisfied by known versions of spdk
    happening in /home/runner/work/spack/spack/var/spack/repos/builtin/packages/daos/package.py
```

tldahlgren (dismissed):
Confirmed the branch and tag.  Just waiting on the dependency so this can be updated.

alalazo:
> would you please merge this?

CI must pass before we merge, this is still waiting for dependencies to be added, and CI to re-run. In general, if you have related changes, I'd advise to submit them as a single PR instead of multiple PRs with dependencies among themselves.

hyoklee:
@spackbot run pipeline

hyoklee:
@alalazo and @tldahlgren , all checks passed. Would you please merge this PR?

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,92 @@
...
+    depends_on("hwloc")
+    depends_on("isa-l@2.30.0:")
+    depends_on("isa-l-crypto@2.23.0:")
+    depends_on("libfabric@1.15.1:")
+    depends_on("libfuse@3.6.1:")
+    depends_on("libuuid")
+    depends_on("libunwind")
+    depends_on("libyaml")
+    depends_on("mercury@2.2.0:+boostsys")
+    depends_on("mpich")
```
soumagne: @hyoklee mpich is only needed for building certain tests so unless you need to build those tests, you should consider removing the dependency on mpich, it's just going to cause you more issues down the road.

soumagne (approved):

var/spack/repos/builtin/packages/daos/package.py
alalazo: Why was this conversation resolved, while `boost@develop` is still here? Am I missing some comments?

alalazo: Also, can you clarify why `boost` is a `build` dependency?

var/spack/repos/builtin/packages/daos/package.py
```diff
@@ -0,0 +1,91 @@
...
+            format(spec["pmdk"].prefix),
+            format(spec["protobuf-c"].prefix),
+            format(spec["spdk"].prefix),
+            format(spec["ucx"].prefix),
+        ]
+        alt_prefix_clean = []
+        for i in alt_prefix:
+            if i != "/usr":
+                alt_prefix_clean.append(i)
+        alt_prefix_clean.append("/usr")
```
alalazo: I assume these are some sort of dependencies. Therefore, we should traverse those dependencies, instead of hard-coding their names (which might change as the recipe evolve). Then we can:
1. Move `"/usr"` last by sorting the list with an appropriate `key` function
2. Avoid adding `/usr` to the prefixes, if it was not there before[^1]

[^1]:  Wouldn't adding `/usr` be picking dependencies from the system, thus masking modeling errors?

alalazo (changes requested):
I left a few comments, let me know if anything is unclear. Thanks!

var/spack/repos/builtin/packages/daos/package.py
soumagne: thanks @alalazo for bringing that up I had missed that in my second review pass. @hyoklee it should be `boost` only, why do you need `+python` ? It's only a `build` dependency, that part is correct because we only use the boost preprocessor macros, not any boost library.

soumagne (changes requested):

var/spack/repos/builtin/packages/daos/package.py
hyoklee: @alalazo, I removed develop & python. However, I don't understand "traverse those dependencies, instead of hard-coding their names" as well as 1. and 2. Can you give me a Python code illustration that you desire?

alalazo: You can use:
```python
for node in spec.traverse():
    node_prefix = node.prefix
    ...
```
instead of hard-coding the names of the dependencies. Doing like that is shorter, and future-proof with respect to changes in dependencies. Once you built your initial list, you can use the key function to order the list:
```
Python 3.8.10 (default, Mar 13 2023, 10:26:41) 
[GCC 9.4.0] on linux
>>> l = ["/some/path", "/usr", "/usr/lib", "/another/path"]
>>> sorted(l, key = lambda x: x.startswith("/usr"))
['/some/path', '/another/path', '/usr', '/usr/lib']
```

hyoklee: Thanks @alalazo ! I'll try your `for` loop code.

hyoklee:
@alalazo , would you please merge this PR?

hyoklee:
@tldahlgren , would you please merge this PR?

kwryankrattiger (approved):
LGTM

alalazo (approved):

PR #35652, opened 2023-02-23
scons: add a new version

hyoklee:
This will help https://github.com/spack/spack/pull/35649


tldahlgren (commented):
Confirmed the sha256.

adamjstewart (commented):
The base class adds python as a build/run dep, but the package adds it as a build/link dep. Does the package actually link to python's libs/headers? If not I would remove the python dep since it's already added by the base class and Spack only supports 3.7+ anyway

hyoklee:
> 

@adamjstewart , would you please ping the original author of this package?
 If no one maintains it, I can try removing Python dependency but 
I want to defer to someone who wrote this package first.

adamjstewart:
I'm pretty sure I'm the one who wrote it. I just checked and it doesn't link to python, so you can simply remove the dep and use the one inherited from the base class.

adamjstewart (approved):

PR #35655, opened 2023-02-23
Add LLVM_ENABLE_RTTI option to llvm-amdgpu package, required by mesa.

nicholas-sly:
This change is required to enable building mesa against this package, as is done when building the E4S full-stack in their own minimal container.


becker33 (approved):

srekolam (commented):
This flag does not seem to be set in our internal CI builds. I need to reconfirm if this is can be enabled . Just need a little time more time to consult with the team

srekolam (approved):
verified with the team internally. It is okay if this needs to be enabled for mesa. In our internal ci builds,  it is turned off though.

alalazo:
@spackbot run pipeline

becker33:
@spackbot run pipeline

PR #34701, opened 2022-12-26
Updates to symlink.py for windows unit test development

techxdave:
This PR is for changes to the symlink operation in the llnl/util module to beef up and correct the functionality on windows considering both Administrator and non-Administrator command prompts where os.symlink may or may not work. We no longer retreat to copying files when we can't symlink; and specifically, when we can't symlink on Windows because of permissions issue, we use a junction or a hardlink, which are available to users running without admin permissions.

PR #35656, opened 2023-02-23
Enable environment tests for Windows

markus-ferrell:
Slight changes to environments to enable windows environment tests. Opens up support for enviroments through batfiles to match the functionality for environments with bash files on linux.

Note: this is rebased on top of [windows symlink changes](https://github.com/spack/spack/pull/34701) which is where the symlink file changes come from

### DIFF
--- a/lib/spack/llnl/util/symlink.py
+++ b/lib/spack/llnl/util/symlink.py
@@ -5,70 +5,152 @@
 import errno
 import os
 import shutil
+import sys
 import tempfile
-from os.path import exists, join
-from sys import platform as _platform
 
-from llnl.util import lang
+from llnl.util import lang, tty
 
-is_windows = _platform == "win32"
+from spack.error import SpackError
 
-if is_windows:
+if sys.platform == "win32":
     from win32file import CreateHardLink
 
 
-def symlink(real_path, link_path):
+def symlink(real_path: str, link_path: str):
     """
-    Create a symbolic link.
+    Try to create a symbolic link.
 
-    On Windows, use junctions if os.symlink fails.
+    On non-Windows and Windows with System Administrator
+    privleges this will be a normal symbolic link via
+    os.symlink.
+
+    On Windows without privledges the link will be a
+    junction for a directory and a hardlink for a file.
+    On Windows the various link types are:
+
+    Symbolic Link: A link to a file or directory on the
+    same or different volume (drive letter) or even to
+    a remote file or directory (using UNC in its path).
+    Need System Administrator privileges to make these.
+
+    Hard Link: A link to a file on the same volume (drive
+    letter) only. Every file (file's data) has at least 1
+    hard link (file's name). But when this method creates
+    a new hard link there will be 2. Deleting all hard
+    links effectively deletes the file. Don't need System
+    Administrator privileges.
+
+    Junction (sometimes called soft link): A link to a
+    directory on the same or different volume (drive
+    letter) but not to a remote directory. Don't need
+    System Administrator privileges.
     """
-    if not is_windows:
-        os.symlink(real_path, link_path)
-    elif _win32_can_symlink():
-        # Windows requires target_is_directory=True when the target is a dir.
-        os.symlink(real_path, link_path, target_is_directory=os.path.isdir(real_path))
+    real_path = os.path.abspath(real_path)
+    link_path = os.path.abspath(link_path)
+
+    if os.path.exists(link_path):
+        raise SymlinkError(f"Link path ({link_path}) already exists. Cannot create link.")
+    elif not os.path.exists(real_path):
+        raise SymlinkError(f"Source path ({real_path}) does not exist. Cannot create link.")
+    elif sys.platform == "win32" and not _windows_can_symlink():
+        _windows_create_link(real_path, link_path)
     else:
-        try:
-            # Try to use junctions
-            _win32_junction(real_path, link_path)
-        except OSError:
-            # If all else fails, fall back to copying files
-            shutil.copyfile(real_path, link_path)
+        os.symlink(real_path, link_path, target_is_directory=os.path.isdir(real_path))
+
+    if not os.path.exists(link_path):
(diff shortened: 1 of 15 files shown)


lib/spack/llnl/util/symlink.py
```diff
@@ -57,7 +57,7 @@ def _win32_junction(path, link):
         path = os.path.join(parent, path)
         path = os.path.abspath(path)
 
-    CreateHardLink(link, path)
+    CreateHardLink(link, path) # Hard Link is for files, junctions are for directories. This will only work for files
```
markus-ferrell, line 60: The junction code path doesn't seem to use real junctions. It uses HardLinks instead which only work for files. Junctions only work for directories. This should either be renamed or reworked to use junctions for dirs and hard links for files. WDYT @johnwparent

markus-ferrell:
[The junction code path](https://github.com/markus-ferrell/spack/blob/3759b5bf3fbaad8a6809877ce3bd3595b907c09c/lib/spack/llnl/util/symlink.py#L61 ) doesn't seem to use real junctions. It uses HardLinks instead which only work for files. Junctions only work for directories. This should either be renamed or reworked to use junctions for dirs and hard links for files. WDYT @johnwparent

lib/spack/llnl/util/symlink.py
```diff
@@ -15,6 +15,7 @@
 
 if is_windows:
     from win32file import CreateHardLink
+    from pywintypes import error
```
johnwparent: What is this `error` type? I assume some abstraction around an actual Windows style error, but it's not at all clear/obvious what this error is, what could trigger it, or what it means if it's triggered, or how generic the catch is now. Can we just catch the actual error this represents? Other Windows specific errors have been referenced/caught elsewhere in the code base are able to handle this scenario (I believe in `llnl/util/filesystem.py`), so if possible could we refactor to do something like that?

johnwparent:
@spackbot fix style

johnwparent:
@markus-ferrell I believe there may be some overlap with #34701 w.r.t. to the change to `symlink.py`. With regard to the hardlinks issue, is there currently an error occurring caused by that issue? It does seem on the surface like that should raise an issue when linking to directories, as you're right, hardlinks for dirs do not exist, but that would mean we are never calling that method on a directory or that that code path is never hit or tested, which is either way an issue. I also do not see an equivalent win32file API endpoint for junctions unfortunately, which means we'll need to likely handroll that functionality. Could you investigate this a little further?

markus-ferrell:
Changes to support tests env.py and environment_modifications.py

lib/spack/spack/test/cmd/env.py
```diff
@@ -823,7 +841,10 @@ def test_env_with_included_config_missing_file(tmpdir, mutable_empty_config):
     spack_yaml = tmpdir.join("spack.yaml")
     missing_file = tmpdir.join("packages.yaml")
     with spack_yaml.open("w") as f:
-        f.write("spack:\n  include:\n    - {0}\n".format(missing_file.strpath))
+        file_seperator = "//"
+        if is_windows:
+          file_seperator = "///"
```
markus-ferrell: I'm not a fan of having to do this all over the place. Is there a helper function to create these file paths?

johnwparent: I assume the `///` is an artifact of file urls? IIRC there's no handler now, but pathlib/urllib should make the need for a handler unnecessary if you can use those tools?

lib/spack/spack/test/environment_modifications.py
```diff
@@ -295,11 +352,20 @@ def test_source_files(files_to_be_sourced):
 
     assert len(modifications["PATH_LIST"]) == 3
     assert isinstance(modifications["PATH_LIST"][0], RemovePath)
-    assert modifications["PATH_LIST"][0].value == "/path/third"
+    if is_windows:
+        assert modifications["PATH_LIST"][0].value == "\\path\\third"
+    else:
+        assert modifications["PATH_LIST"][0].value == "/path/third"
```
markus-ferrell: sections like this are going to get cleaned up by using system specific separators

lib/spack/spack/test/environment_modifications.py
```diff
@@ -70,9 +71,21 @@ def prepare_environment_for_tests(working_env):
...
+    os.environ["PATH_LIST"] = list_sep.join([os.sep.join(["","path","second"]), os.sep.join(["","path", "third"])])
+    os.environ["REMOVE_PATH_LIST"] = list_sep.join([
+        os.sep.join(["","a", "b"]),
+        os.sep.join(["","duplicate"]),
+        os.sep.join(["","a", "c"]),
+        os.sep.join(["","remove", "this"]),
+        os.sep.join(["","a","d"]),
+        os.sep.join(["","duplicate"]),
+        os.sep.join(["","f","g"])
+        ])
```
markus-ferrell, line 84: This feels like horrible syntax but at least it's the same code for both platforms? any recommendations?

lib/spack/spack/test/cmd/env.py
markus-ferrell: yup, makes sense. I cleaned it up with pathlib, lmk if it looks better now.

lib/spack/spack/util/environment.py
```diff
@@ -980,18 +980,27 @@ def environment_after_sourcing_files(*files, **kwargs):
...
     """
     # Set the shell executable that will be used to source files
-    shell_cmd = kwargs.get("shell", "/bin/bash")
-    shell_options = kwargs.get("shell_options", "-c")
-    source_command = kwargs.get("source_command", "source")
-    suppress_output = kwargs.get("suppress_output", "&> /dev/null")
+    if sys.platform == "win32":
+        shell_cmd = kwargs.get("shell", "cmd.exe")
+        shell_options = kwargs.get("shell_options", "")
+        suppress_output = kwargs.get("suppress_output", "")
```
markus-ferrell, line 986: @johnwparent This got merged in with install tests. Disregard changes for this file.

PR #35658, opened 2023-02-23
py-fiona: add upper bounds to Python versions

adamjstewart:
Discovered entirely through trial and error. Older versions don't build due to C99 implicit declaration errors, so I couldn't test farther back, but I'm guessing only Python 3.7 works, and we're deprecating 3.7 soon anyway, so I deprecated those versions too.


adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

alalazo:
@spackbot run pipeline

alalazo (approved):

PR #35662, opened 2023-02-23
hpctoolkit: add version 2023.03.01, add python variant

mwkrentel:
1. add version 2023.02.20
 2. add variant 'python' that supports unwinding python source
 3. clean up some things with the cray variant
 4. require the latest libmonitor
 5. fix sha256 checksum for url patch
 6. delete rocm 5.3 from older versions


mwkrentel:
@spackbot fix style

mwkrentel:
There is a mistake in how hpctoolkit tests for the rocm version.
That's why it fails for recent rocm.

We'll fix this in hpctoolkit and push up a new release in a few days.

mwkrentel:
@spackbot style --fix

mwkrentel:
@spackbot fix style

var/spack/repos/builtin/packages/hpctoolkit/package.py
```diff
@@ -45,17 +46,19 @@ class Hpctoolkit(AutotoolsPackage):
...
     # MPI applications.  These options add hpcprof-mpi, the MPI
-    # version of hpcprof.  Cray needs a separate option because an
-    # external MPI module in packages.yaml doesn't work.
+    # version of hpcprof.  Cray is a separate option for old systems
+    # where an external MPI module doesn't work.
     variant(
-        "cray", default=False, description="Build hpcprof-mpi for Cray systems (requires --dirty)."
+        "cray",
+        default=False,
+        description="Build hpcprof-mpi for Cray systems (may require --dirty).",
```
alalazo, line 54: Just a note: using `sticky=True` prevents clingo from accidentaly changing this variant (and require users to change it explicitly if need be). If that's useful we can add that in another PR.

var/spack/repos/builtin/packages/hpctoolkit/package.py
```diff
@@ -139,9 +145,11 @@ class Hpctoolkit(AutotoolsPackage):
     depends_on("memkind", type=("build", "run"), when="@2021.05.01:")
     depends_on("papi", when="+papi")
     depends_on("libpfm4", when="~papi")
+    depends_on("mpi", when="+cray")
```
alalazo, line 148: Does it make sense instead:
```python
conflicts("~mpi", when="+cray", reason="activating +cray implies +mpi")
```
?

alalazo (approved):
LGTM, I left two minor comments. If this seems good to go I'll merge it.

var/spack/repos/builtin/packages/hpctoolkit/package.py
mwkrentel: I don't understand.

var/spack/repos/builtin/packages/hpctoolkit/package.py
mwkrentel: Actually, our use of `+cray` is kinda unusual.  We use it as an
alternative to `+mpi` for some cases where the standard MPI doesn't
work (don't ask).  So, in this case, `+cray` depends on `mpi`
(cray-mpich), but it doesn't imply `+mpi`.

var/spack/repos/builtin/packages/hpctoolkit/package.py
alalazo: https://spack.readthedocs.io/en/latest/packaging_guide.html#sticky-variants

mwkrentel: Hmm, I'll have to think about this for a while.
I don't feel that I understand all the implications of using `sticky`.
But it's good to know, a useful tool in the toolbox.  Thanks!

PR #35663, opened 2023-02-23
matrices: broadcast key combinatorially applies to all nodes in matrix

becker33:
Supersedes #35529 

Currently it is not possible to use concretizer:unify:when_possible configuration to minimize the dependencies among a series of packages built with a combination of MPI/compiler combinations. The matrix:
```
- matrix:
  - [foo]
  - [^mpich]
  - [ '%gcc', '%clang']
```
resolves to specs `foo%gcc ^mpich` and `foo%clang ^mpich`.

With the when_possible unification option, the concretizer will concretize to use a single mpich in both cases (because it is possible to do so), mixing compilers rather than building with each combination of compiler/mpi.

With this PR, we introduce a `broadcast` key to matrices. The broadcasted constraints are applied to all nodes in the matrix. For example

```
- matrix:
  - [foo]
  - [^mpich]
  broadcast:
  - ['%gcc', '%clang']
```
resolves to two specs `foo%gcc^mpich%gcc` and `foo%clang^mpich%clang`.

The `broadcast` key accepts the same syntax as the matrix key, allowing for complex constraints.

```
- matrix:
  - [foo]
  - [^mpich]
  broadcast:
  - ['%gcc', '%clang']
  - [+debug, ~debug]
```
creates 4 specs.

### DIFF
--- a/lib/spack/spack/schema/env.py
+++ b/lib/spack/spack/schema/env.py
@@ -30,6 +30,10 @@ spec_list_schema = {
                         "type": "array",
                         "items": {"type": "array", "items": {"type": "string"}},
                     },
+                    "broadcast": {
+                        "type": "array",
+                        "items": {"type": "array", "items": {"type": "string"}},
+                    },
                     "exclude": {"type": "array", "items": {"type": "string"}},
                 },
             },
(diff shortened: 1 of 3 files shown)


becker33:
@spackbot fix style

PR #35670, opened 2023-02-24
Variant and conflict fixes on intel-mpi-benchmarks

besnardjb:
When installing the IMB you can only get `IMB-MPI1` the reason is that the variants are somehow mutually-exclusive. This fixes this issues by removing some `elif`. I have also added some conflicts and inhibited parallel build for older versions.

### DIFF
--- a/var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
+++ b/var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
@@ -78,17 +78,17 @@ class IntelMpiBenchmarks(MakefilePackage):
         targets = []
         if "+mpi1" in spec:
             targets.append("MPI1")
-        elif "+ext" in spec:
+        if "+ext" in spec:
             targets.append("EXT")
-        elif "+io" in spec:
+        if "+io" in spec:
             targets.append("IO")
-        elif "+nbc" in spec:
+        if "+nbc" in spec:
             targets.append("NBC")
-        elif "+p2p" in spec:
+        if "+p2p" in spec:
             targets.append("P2P")
-        elif "+rma" in spec:
+        if "+rma" in spec:
             targets.append("RMA")
-        elif "+mt" in spec:
+        if "+mt" in spec:
             targets.append("MT")
 
         if self.spec.satisfies("@2019:"):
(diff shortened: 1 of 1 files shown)


carsonwoods (approved):
This looks good, thanks for catching this and submitting a fix.

carsonwoods (dismissed):

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
```diff
@@ -58,6 +58,14 @@ class IntelMpiBenchmarks(MakefilePackage):
     variant("rma", default=True, description="Build MPI1 benchmark")
     variant("mt", default=True, description="Build MPI1 benchmark")
 
+    # Handle missing variants in previous versions
+    conflicts("+p2p", when="@:2019")
+    conflicts("+mt", when="@:2019")
+
+    # No parallel build before 2019
+    with when("@:2019"):
+        parallel = False
```
alalazo: This ought to be a `property` (`parallel` is not a directive, so the context manager has no effect)

alalazo (changes requested):

var/spack/repos/builtin/packages/intel-mpi-benchmarks/package.py
besnardjb: Was not sure of how to handle this conditional parallel build from other package examples. I can look further tonight if needed.

besnardjb: I had a try in 2573ef7. I had to manually pass to the individual make targets to inhibit parallel build. I also factorized the variants in the process. Feel free to comment, if it is too complex, we can revert to a single `parallel=False`.

alalazo: @besnardjb Why not keeping that as before and just:
```python
@property
def parallel(self):
    if self.spec.satisfies("@:2019"):
        return False
    return True
```
?

besnardjb: You are right it is just me being overcomplicated, AFAIK did not find another package doing such an override. Just pushed the fix as per your recommendation and checked that all versions are building.

Edit: there are many packages doing so in fact, just the wrong grep.

alalazo (approved):

PR #35672, opened 2023-02-24
Back-port a patch for filesystem logic in `gmx` executable.

eirrgang:
Reference: https://gitlab.com/gromacs/gromacs/-/issues/4724

### DIFF
--- a/var/spack/repos/builtin/packages/gromacs/package.py
+++ b/var/spack/repos/builtin/packages/gromacs/package.py
@@ -273,6 +273,14 @@ class Gromacs(CMakePackage):
                 "#include <queue>\n#include <limits>",
                 "src/gromacs/modularsimulator/modularsimulator.h",
             )
+        # Ref: https://gitlab.com/gromacs/gromacs/-/merge_requests/3504
+        if self.spec.satisfies("@2023"):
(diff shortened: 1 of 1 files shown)


eirrgang:
Originally reported by @ggouaillardet, but the upstream fix is not yet released (targeted `gromacs@2023.1`). We should let `gromacs@2023` behave better.

ggouaillardet:
I do not think the "relocatable" is appropriate in the context of this patch (it is more of a bug fix)


Anyway, you might also be interested in the last inlined patch from https://gitlab.com/gromacs/gromacs/-/issues/4728

TL;DR make it work with `libstdc++` from RHEL8

eirrgang:
> I do not think the "relocatable" is appropriate in the context of this patch (it is more of a bug fix)

updated PR description

> Anyway, you might also be interested in the last inlined patch from https://gitlab.com/gromacs/gromacs/-/issues/4728

I think that's out of scope for this PR. That issue has not yet been resolved upstream.

eirrgang:
Note: this update is needed also for https://github.com/spack/spack/pull/35738

junghans (approved):

PR #35674, opened 2023-02-24
Fix `--test` behavior for gromacs package.

eirrgang:
For `spack install --test=all gromacs`
* remove the `test` target from the `check()` call and just use the `check` target, in accordance with usual GROMACS test protocol
* build the test binaries explicitly during the build phase

Additional minor updates are necessary. This change updates the package structure to the newer format with a separate Builder class so we can override `check()`. However, note that additional modernization should be undertaken with care.


junghans (approved):

marvinbernhardt (approved):
tested locally, lgtm

eirrgang:
@junghans and @marvinbernhardt : What needs to happen for this to be merged?

junghans (approved):

PR #35599, opened 2023-02-21
Fix wrong behavior with conditional provider and anonymous satisfies

alalazo:
fixes #35597

Issue #35597, opened 2023-02-21
Spec.__contains__ produces wrong results for `intel-oneapi-mkl`


alalazo:
### Steps to reproduce

Reported by @dev-zero in #35584

```python
Spack version 0.20.0.dev0
Python 3.8.10, Linux x86_64
>>> from spack.spec import Spec
>>> s = Spec("cp2k ^intel-oneapi-mkl")
>>> c = s.concretized()
>>> "intel-oneapi-mkl" in c
True
>>> "^intel-oneapi-mkl" in c
False
>>> c["blas"]
intel-oneapi-mkl@2023.0.0%gcc@9.4.0~cluster+envmods~ilp64+shared build_system=generic arch=linux-ubuntu20.04-icelake ^intel-oneapi-tbb@2021.8.0%gcc@9.4.0+envmods build_system=generic arch=linux-ubuntu20.04-icelake
>>> c["lapack"]
intel-oneapi-mkl@2023.0.0%gcc@9.4.0~cluster+envmods~ilp64+shared build_system=generic arch=linux-ubuntu20.04-icelake ^intel-oneapi-tbb@2021.8.0%gcc@9.4.0+envmods build_system=generic arch=linux-ubuntu20.04-icelake
```

### Error message

The entry:
```python
>>> "^intel-oneapi-mkl" in c
False
```
should return `True`.

### Information on your system

* **Spack:** 0.20.0.dev0 (fbd5c5c2cdc6761cb6cd5dfd2d0c33fdfcdb1712)
* **Python:** 3.8.10
* **Platform:** linux-ubuntu20.04-icelake
* **Concretizer:** clingo

### General information

- [X] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

alalazo:
It seems for the `intel-oneapi-mkl` case we do something wrong here:

https://github.com/spack/spack/blob/5b8917188a4d84713c1037fd9464bf77503edd27/lib/spack/spack/spec.py#L3574-L3584

alalazo:
I think we should have an answer before hitting the lines above, since `"^intel-oneapi-mkl" in c` doesn't involve any virtual dependency at all :thinking:

alalazo:
Trying to investigate if this:
```diff
diff --git a/lib/spack/spack/spec.py b/lib/spack/spack/spec.py
[...]

PR #17563, opened 2020-07-16
ci: fixes for compiler bootstrapping

scottwittenburg:
The commits here developed by @haampie, along with #17536, are allowing compiler bootstrapping pipelines to work again in the wake of the distributed builds PR.

The commit changing the `package_id` format may be controversial, so I want to get some feedback on this before I spend any time fixing the tests which are failing as a result.

PR #35681, opened 2023-02-24
Split `satisfies(..., strict=True/False)` into two functions

alalazo:
closes #35599

This PR splits each `satisfies` method having a `strict=` argument into two function calls:
1. `intersects`: corresponds to `strict=False` and has "intersection" semantic. 
2. `satisfies`: corresponds to `strict=True` and has "subset" semantic.

From 1 it follows that `intersects` is commutative, so in this PR I also tried to remove most of the non-commutative behaviors we had with `strict=False`. 

Modifications:
- [x] Split `satisfies(..., strict=...)` into `satisfies` and `intersects`
- [x] Add unit tests to verify expected properties of the methods
- [x] Fixed bugs in `ArchSpec` related to targets, see https://github.com/spack/spack/pull/17563#pullrequestreview-1312243665
- [x] Fixed semantics of variant, when compared with each other see 3b89375a23f45399495978a3fc02219e81485554
- [x] Fixed semantics of flags (now they satisfy on exact match, not on set match - intersection is True with specs that don't have a flag set) see 3a5b534fa9c8cb42e71fe715ea9859a8a18d5528
- [x] Fixed semantic of constraining an abstract spec by a concrete one, see a2a916778d7410f341acac86aee4b63415b4fc65 and 3466bd29d5dfe286dbba0e4e5cd39bb5d836cd68
- [x] Make `GitVersion` consistent with the `intersects` and `satisfies` semantic (see 10c147fdc57e0dad5fb0e60e0e9fcd3f5095cc99)

### DIFF
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -1503,7 +1503,7 @@ def _from_merged_attrs(fetcher, pkg, version):
     return fetcher(**attrs)
 
 
-def for_package_version(pkg, version):
+def for_package_version(pkg, version=None):
     """Determine a fetch strategy based on the arguments supplied to
     version() in the package description."""
 
@@ -1514,8 +1514,15 @@ def for_package_version(pkg, version):
 
     check_pkg_attributes(pkg)
 
-    if not isinstance(version, spack.version.VersionBase):
-        version = spack.version.Version(version)
+    if version is not None:
+        if not isinstance(version, spack.version.VersionBase):
+            version = spack.version.Version(version)
+
+        version_list = spack.version.VersionList()
+        version_list.add(version)
+        pkg.spec.versions = version_list
+    else:
+        version = pkg.version
 
     # if it's a commit, we must use a GitFetchStrategy
     if isinstance(version, spack.version.GitVersion):
--- a/lib/spack/spack/fetch_strategy.py
+++ b/lib/spack/spack/fetch_strategy.py
@@ -1515,6 +1515,9 @@ def for_package_version(pkg, version=None):
     check_pkg_attributes(pkg)
 
     if version is not None:
+        assert not pkg.spec.concrete, "concrete specs should not pass the 'version=' argument"
+        # Specs are initialized with the universe range, if no version information is given,
+        # so here we make sure we always match the version passed as argument
         if not isinstance(version, spack.version.VersionBase):
             version = spack.version.Version(version)
 
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -778,9 +778,7 @@ class FlagMap(lang.HashableMap):
         return all(f in self and set(self[f]) == set(other[f]) for f in other)
 
     def intersects(self, other):
-        if not all(
-            set(self[f]) == set(other[f]) for f in other if (other[f] != [] and f in self)
-        ):
+        if not all(set(self[f]) == set(other[f]) for f in other if (other[f] != [] and f in self)):
             return False
 
         # Check that the propagation values match
@@ -3510,7 +3508,7 @@ class Spec(object):
         if other.concrete and self.concrete:
             return self.dag_hash() == other.dag_hash()
 
-        # FIXME (INTERSECTS): the lines below are not commutative
+        # TODO (INTERSECTS): the lines below are not commutative
         # If the names are different, we need to consider virtuals
         if self.name != other.name and self.name and other.name:
             # A provider can satisfy a virtual dependency.
@@ -3577,19 +3575,6 @@ class Spec(object):
         for name in self.common_dependencies(other):
             if not self[name].intersects(other[name], deps=False):
                 return False
-
-        # FIXME (INTERSECTS): Remove this or generalize to other.concrete ?
-        if self.concrete:
-            # Here we already checked that common dependencies are compatible, so if there are no
-            # nodes in other that are not in self return True, otherwise return False
-            try:
-                _ = [self[x.name] for x in other.traverse() if x.name is not None]
-            except KeyError:
-                # If self is concrete, and some node is not in self, then return False
-                # (since self is immutable and can't satisfy other)
-                return False
-            else:
-                return True
 
         # For virtual dependencies, we need to dig a little deeper.
         self_index = spack.provider_index.ProviderIndex(
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -193,9 +193,7 @@ def colorize_spec(spec):
 
 @lang.lazy_lexicographic_ordering
 class ArchSpec(object):
-    """Aggregate the target platform, the operating system and the target
-    microarchitecture into an architecture spec..
-    """
+    """Aggregate the target platform, the operating system and the target microarchitecture."""
 
     @staticmethod
     def _return_arch(os_tag, target_tag):
@@ -402,7 +400,7 @@ class ArchSpec(object):
 
         return self._target_satisfies(other, strict=False)
 
-    def _target_satisfies(self, other, strict):
+    def _target_satisfies(self, other: "ArchSpec", strict: bool) -> bool:
         if strict is True:
             need_to_check = bool(other.target)
         else:
@@ -417,8 +415,8 @@ class ArchSpec(object):
 
         return bool(self._target_intersection(other))
 
-    def target_constrain(self, other):
-        if not other.target_satisfies(self, strict=False):
+    def _target_constrain(self, other: "ArchSpec") -> bool:
+        if not other._target_satisfies(self, strict=False):
             raise UnsatisfiableArchitectureSpecError(self, other)
 
         if self.target_concrete:
@@ -430,8 +428,12 @@ class ArchSpec(object):
 
         # Compute the intersection of every combination of ranges in the lists
         results = self._target_intersection(other)
-        # Do we need to dedupe here?
-        self.target = ",".join(results)
+        attribute_str = ",".join(results)
+
+        if self.target == attribute_str:
+            return False
+
+        self.target = attribute_str
         return True
 
     def _target_intersection(self, other):
@@ -503,7 +505,7 @@ class ArchSpec(object):
                 setattr(self, attr, ovalue)
                 constrained = True
 
-        constrained |= self.target_constrain(other)
+        constrained |= self._target_constrain(other)
 
         return constrained
 
@@ -519,7 +521,9 @@ class ArchSpec(object):
     @property
     def target_concrete(self):
         """True if the target is not a range or list."""
-        return ":" not in str(self.target) and "," not in str(self.target)
+        return (
+            self.target is not None and ":" not in str(self.target) and "," not in str(self.target)
+        )
 
     def to_dict(self):
         d = syaml.syaml_dict(
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -364,7 +364,7 @@ class ArchSpec(object):
 
         self._target = value
 
-    def satisfies(self, other, strict):
+    def satisfies(self, other, strict=False):
         """Predicate to check if this spec satisfies a constraint.
 
         Args:
@@ -593,6 +593,9 @@ class CompilerSpec(object):
             return compiler_spec_like
         return CompilerSpec(compiler_spec_like)
 
+    def intersects(self, other):
+        return self.satisfies(other, strict=False)
+
     def satisfies(self, other, strict=False):
         other = self._autospec(other)
         return self.name == other.name and self.versions.satisfies(other.versions, strict=strict)
@@ -605,7 +608,7 @@ class CompilerSpec(object):
         other = self._autospec(other)
 
         # ensure that other will actually constrain this spec.
-        if not other.satisfies(self):
+        if not other.intersects(self):
             raise UnsatisfiableCompilerSpecError(other, self)
 
         return self.versions.intersect(other.versions)
@@ -1058,7 +1061,7 @@ def _libs_default_handler(descriptor, spec, cls):
     home = getattr(spec.package, "home")
 
     # Avoid double 'lib' for packages whose names already start with lib
-    if not name.startswith("lib") and not spec.satisfies("platform=windows"):
+    if not name.startswith("lib") and not spec.placeholder_satisfies("platform=windows"):
         name = "lib" + name
 
     # If '+shared' search only for shared library; if '~shared' search only for
@@ -2613,9 +2616,9 @@ class Spec(object):
                 # it's possible to build that configuration with Spack
                 continue
             for conflict_spec, when_list in x.package_class.conflicts.items():
-                if x.satisfies(conflict_spec, strict=True):
+                if x.placeholder_satisfies(conflict_spec):
                     for when_spec, msg in when_list:
-                        if x.satisfies(when_spec, strict=True):
+                        if x.placeholder_satisfies(when_spec):
                             when = when_spec.copy()
                             when.name = x.name
                             matches.append((x, conflict_spec, when, msg))
@@ -2625,7 +2628,7 @@ class Spec(object):
         # Check if we can produce an optimized binary (will throw if
         # there are declared inconsistencies)
         # No need on platform=cray because of the targeting modules
-        if not self.satisfies("platform=cray"):
+        if not self.placeholder_satisfies("platform=cray"):
             self.architecture.target.optimization_flags(self.compiler)
 
     def _patches_assigned(self):
@@ -2667,7 +2670,7 @@ class Spec(object):
             # Add any patches from the package to the spec.
             patches = []
             for cond, patch_list in s.package_class.patches.items():
-                if s.satisfies(cond, strict=True):
+                if s.placeholder_satisfies(cond):
                     for patch in patch_list:
                         patches.append(patch)
             if patches:
@@ -2685,7 +2688,9 @@ class Spec(object):
             patches = []
             for cond, dependency in pkg_deps[dspec.spec.name].items():
                 for pcond, patch_list in dependency.patches.items():
-                    if dspec.parent.satisfies(cond, strict=True) and dspec.spec.satisfies(pcond):
+                    if dspec.parent.placeholder_satisfies(
+                        cond
+                    ) and dspec.spec.placeholder_satisfies(pcond):
                         patches.extend(patch_list)
             if patches:
                 all_patches = spec_to_patches.setdefault(id(dspec.spec), [])
@@ -2943,7 +2948,7 @@ class Spec(object):
         # evaluate when specs to figure out constraints on the dependency.
         dep = None
         for when_spec, dependency in conditions.items():
-            if self.satisfies(when_spec, strict=True):
+            if self.placeholder_satisfies(when_spec):
                 if dep is None:
                     dep = dp.Dependency(self.name, Spec(name), type=())
                 try:
@@ -2978,7 +2983,7 @@ class Spec(object):
             # result.
             for provider in providers:
                 for spec in providers:
-                    if spec is not provider and provider.satisfies(spec):
+                    if spec is not provider and provider.intersects(spec):
                         providers.remove(spec)
             # Can't have multiple providers for the same thing in one spec.
             if len(providers) > 1:
@@ -3303,7 +3308,7 @@ class Spec(object):
         # already satisfies the constraint (and the method returns False)
         # or it raises an exception
         if self.concrete:
-            if self.satisfies(other):
+            if self.intersects(other):
                 return False
             else:
                 raise spack.error.UnsatisfiableSpecError(self, other, "constrain a concrete spec")
@@ -3389,7 +3394,7 @@ class Spec(object):
         # TODO: might want more detail than this, e.g. specific deps
         # in violation. if this becomes a priority get rid of this
         # check and be more specific about what's wrong.
-        if not other.satisfies_dependencies(self):
+        if not other._intersects_dependencies(self):
             raise UnsatisfiableDependencySpecError(other, self)
 
         if any(not d.name for d in other.traverse(root=False)):
@@ -3451,25 +3456,140 @@ class Spec(object):
             return spec_like
         return Spec(spec_like)
 
-    def satisfies(self, other, deps=True, strict=False):
-        """Determine if this spec satisfies all constraints of another.
+    def intersects(self, other: "Spec", deps: bool = True) -> bool:
+        """Return True if there exists at least one concrete spec that matches both
+        self and other, otherwise False.
 
-        There are two senses for satisfies, depending on the ``strict``
-        argument.
+        This operation is commutative, and if two specs intersect it means that one
+        can constrain the other.
 
-          * ``strict=False``: the left-hand side and right-hand side have
-            non-empty intersection. For example ``zlib`` satisfies
-            ``zlib@1.2.3`` and ``zlib@1.2.3`` satisfies ``zlib``. In this
-            sense satisfies is a commutative operation: ``x.satisfies(y)``
-            if and only if ``y.satisfies(x)``.
-
-          * ``strict=True``: the left-hand side is a subset of the right-hand
-            side. For example ``zlib@1.2.3`` satisfies ``zlib``, but ``zlib``
-            does not satisfy ``zlib@1.2.3``. In this sense satisfies is not
-            commutative: the left-hand side should be at least as constrained
-            as the right-hand side.
+        Args:
+            other: spec to be checked for compatibility
+            deps: if True check compatibility of dependency nodes too, if False only check root
         """
+        # return self._satisfies(other, strict=False, deps=deps)
+        other = self._autospec(other)
 
+        # Optimizations concrete specs
+        if other.concrete and self.concrete:
+            return self.dag_hash() == other.dag_hash()
+
+        # FIXME (INTERSECTS): the lines below are not commutative
+        # If the names are different, we need to consider virtuals
+        if self.name != other.name and self.name and other.name:
+            # A provider can satisfy a virtual dependency.
+            if not self.virtual and other.virtual:
+                try:
+                    # Here we might get an abstract spec
+                    pkg_cls = spack.repo.path.get_pkg_class(self.fullname)
+                    pkg = pkg_cls(self)
+                except spack.repo.UnknownEntityError:
+                    # If we can't get package info on this spec, don't treat
+                    # it as a provider of this vdep.
+                    return False
+
+                if pkg.provides(other.name):
+                    for provided, when_specs in pkg.provided.items():
+                        if any(self.intersects(when, deps=False) for when in when_specs):
+                            if provided.intersects(other):
+                                return True
+            return False
+
+        # namespaces either match, or other doesn't require one.
+        if (
+            other.namespace is not None
+            and self.namespace is not None
+            and self.namespace != other.namespace
+        ):
+            return False
+
+        if self.versions and other.versions:
+            if not self.versions.satisfies(other.versions, strict=False):
+                return False
+
+        # None indicates no constraints when not strict.
+        if self.compiler and other.compiler:
+            if not self.compiler.satisfies(other.compiler, strict=False):
+                return False
+
+        var_strict = False
+        if (not self.name) or (not other.name):
+            var_strict = True
+        if not self.variants.satisfies(other.variants, strict=var_strict):
+            return False
+
+        # Architecture satisfaction is currently just string equality.
+        # If not strict, None means unconstrained.
+        if self.architecture and other.architecture:
+            if not self.architecture.satisfies(other.architecture, strict=False):
+                return False
+
+        if not self.compiler_flags.satisfies(other.compiler_flags, strict=False):
+            return False
+
+        # If we need to descend into dependencies, do it, otherwise we're done.
+        if deps:
+            return self._intersects_dependencies(other)
+        else:
+            return True
+
+    def _intersects_dependencies(self, other):
+        other = self._autospec(other)
+
+        if not other._dependencies or not self._dependencies:
+            # one spec *could* eventually satisfy the other
+            return True
+
+        # Handle first-order constraints directly
+        for name in self.common_dependencies(other):
+            if not self[name].intersects(other[name], deps=False):
+                return False
+
+        # FIXME (INTERSECTS): Remove this or generalize to other.concrete ?
+        if self.concrete:
+            # Here we already checked that common dependencies are compatible, so if there are no
+            # nodes in other that are not in self return True, otherwise return False
+            try:
+                _ = [self[x.name] for x in other.traverse() if x.name is not None]
+            except KeyError:
+                # If self is concrete, and some node is not in self, then return False
+                # (since self is immutable and can't satisfy other)
+                return False
+            else:
+                return True
+
+        # For virtual dependencies, we need to dig a little deeper.
+        self_index = spack.provider_index.ProviderIndex(
+            repository=spack.repo.path, specs=self.traverse(), restrict=True
+        )
+        other_index = spack.provider_index.ProviderIndex(
+            repository=spack.repo.path, specs=other.traverse(), restrict=True
+        )
+
+        # This handles cases where there are already providers for both vpkgs
+        if not self_index.satisfies(other_index):
+            return False
+
+        # These two loops handle cases where there is an overly restrictive
+        # vpkg in one spec for a provider in the other (e.g., mpi@3: is not
+        # compatible with mpich2)
+        for spec in self.virtual_dependencies():
+            if spec.name in other_index and not other_index.providers_for(spec):
+                return False
+
+        for spec in other.virtual_dependencies():
+            if spec.name in self_index and not self_index.providers_for(spec):
+                return False
+
+        return True
+
+    def placeholder_satisfies(self, other: "Spec", deps: bool = True) -> bool:
+        """Return True if all concrete specs matching self also match other, otherwise False.
+
+        Args:
+            other: spec to be satisfied
+            deps: if True descend to dependencies, otherwise only check root node
+        """
         other = self._autospec(other)
 
         # Optimizations for right-hand side concrete:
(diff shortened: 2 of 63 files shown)


lib/spack/spack/spec.py
```diff
@@ -3451,25 +3491,123 @@ def _autospec(self, spec_like):
...
+            other: spec to be checked for compatibility
+            deps: if True check compatibility of dependency nodes too, if False only check root
         """
+        other = self._autospec(other)
+
+        # Optimizations concrete specs
+        if other.concrete and self.concrete:
+            return self.dag_hash() == other.dag_hash()
+
+        # TODO (INTERSECTS): the lines below are not commutative
```
alalazo: This is one place where I need to look what needs to be done. This is the code we had before, but it is clearly non-commutative.

lib/spack/spack/test/spec_semantics.py
```diff
@@ -1295,14 +1298,126 @@ def test_package_hash_affects_dunder_and_dag_hash(mock_packages, default_mock_co
...
+        (CompilerSpec, "gcc", "clang", (False, False, False)),
+        (CompilerSpec, "gcc", "gcc@5", (True, False, True)),
+        (CompilerSpec, "gcc@5", "gcc@5.3", (True, False, True)),
+        (CompilerSpec, "gcc@5", "gcc@5-tag", (True, False, True)),
+        # Flags (flags are a map, so for convenience we initialize a full Spec)
+        # Note: the semantic is that of sv variants, not mv variants
+        (Spec, "cppflags=-foo", "cppflags=-bar", (False, False, False)),
+        (Spec, "cppflags='-bar -foo'", "cppflags=-bar", (False, False, False)),
+        (Spec, "cppflags=-foo", "cppflags=-foo", (True, True, True)),
+        (Spec, "cppflags=-foo", "cflags=-foo", (True, False, False)),
```
alalazo, line 1255: More cases to test are welcome here, if reviewers have some.

lib/spack/spack/test/spec_syntax.py
```diff
@@ -1040,18 +1040,30 @@ def test_compare_abstract_specs():
         assert a <= b or b < a
 
 
+@pytest.mark.xfail(reason="FIXME (INTERSECTS)")
```
alalazo: This is another part I need to look over more closely.

lib/spack/spack/test/variant.py
```diff
@@ -638,11 +638,11 @@ def test_satisfies_and_constrain(self):
         b["foobar"] = SingleValuedVariant("foobar", "fee")
         b["shared"] = BoolValuedVariant("shared", True)
 
-        assert not a.satisfies(b)
-        assert b.satisfies(a)
+        assert a.intersects(b)
+        assert b.intersects(a)
```
alalazo, line 642: This is because of commutativity of `intersects`. Before the semantic imo was not that clear.

lib/spack/spack/variant.py
```diff
@@ -575,29 +578,11 @@ def substitute(self, vspec):
...
-        to_be_checked = [k for k in other]
-
-        strict_or_concrete = strict
-        if self.spec is not None:
-            strict_or_concrete |= self.spec._concrete
-
-        if not strict_or_concrete:
-            to_be_checked = filter(lambda x: x in self, to_be_checked)
+    def satisfies(self, other):
+        return all(k in self and self[k].satisfies(other[k]) for k in other)
```
haampie, line 586: yay

lib/spack/spack/variant.py
```diff
@@ -575,29 +578,11 @@ def substitute(self, vspec):
...
 
-        return all(k in self and self[k].satisfies(other[k]) for k in to_be_checked)
+    def intersects(self, other):
+        return all(self[k].intersects(other[k]) for k in other if k in self)
```
haampie, line 589: more yay ;p

hppritcha (dismissed):

lib/spack/spack/spec.py
```diff
@@ -3451,25 +3491,123 @@ def _autospec(self, spec_like):
...
+        # If the names are different, we need to consider virtuals
+        if self.name != other.name and self.name and other.name:
+            # A provider can satisfy a virtual dependency.
+            if not self.virtual and other.virtual:
```
haampie: Why not say

```python
if self.name != other.name:
    if self.virtual and other.virtual:
        if not there_are_pkgs_providing_both(self.name, other.name):
            return False
    elif self.virtual or other.virtual:
        virtual, nonvirtual = (self, other) if self.virtual else (other, self) 
        # the package lookup stuff but replace self/other with virtual/nonvirtual
    return False
```

adamjstewart:
I'm not caught up on the history of this but I do seem to recall that the following (which are all used frequently in packages) are slightly different:
```python
if "+python" in self.spec:
if self.spec.satisfies("+python"):
if self.spec.satisfies("+python", strict=True):
```
With this PR, will the first two be equivalent and correct? Can't remember what the issue was with these originally.

lib/spack/spack/test/spec_syntax.py
alalazo: Should be fixed in 10c147fdc57e0dad5fb0e60e0e9fcd3f5095cc99

alalazo:
@adamjstewart This PR tries to have a slightly more formal notion of `strict=True` and `strict=False`, and splits the behavior around that argument in two functions. One is:
```python
self.intersects(other)
```
which roughly corresponds to `strict=False` and has intersection semantic - if there exist at least one spec that matches both `self` and `other` then `self.intersects(other)` is true. Otherwise it's false. This function is commutative.

The other is:
```python
self.satisfies(other)
```
which corresponds to `strict=True` and has subset semantic, i.e. all the specs matching self also match other. 

This refactor was motivated by the bug in #35597 which, upon investigation, revealed a few inconsistencies in the current API that we are trying to fix here. In the course of preparing this PR I also discovered a few bugs in version comparison and target comparison, that are fixed here.

lib/spack/spack/spec.py
alalazo: Done in aea2da1aea8f3245f95fc40335a740ff66acfa06

lib/spack/spack/spec.py
alalazo: Fixed in aea2da1aea8f3245f95fc40335a740ff66acfa06

adamjstewart:
I guess what I'm asking is whether:

```python
if "+foo" in self.spec:
```
and:
```python
if self.spec.satisfies("+foo"):
```
are now identical, and if not, which is correct? And do all current packages use the correct one?

The idea that both are commonly used but both don't have the same meaning scares me.

alalazo:
> are now identical, and if not, which is correct? And do all current packages use the correct one?

They are now for anonymous specs. For named specs:
```python
"openmpi +foo" in spec
```
is equivalent to:
```python
spec.satisfies("^openmpi +foo")
```
when `spec.name != "openmpi"`

lib/spack/spack/fetch_strategy.py
```diff
@@ -1514,8 +1514,15 @@ def for_package_version(pkg, version):
 
     check_pkg_attributes(pkg)
 
-    if not isinstance(version, spack.version.VersionBase):
-        version = spack.version.Version(version)
+    if version is not None:
```
haampie, line 1517: I don't follow this, can you at least add a comment as to why this is mutating `pkg.spec`? Looks a bit unexpected...

lib/spack/spack/test/cmd/install.py
```diff
@@ -276,7 +276,7 @@ def test_install_commit(mock_git_version_info, install_mockery, mock_packages, m
     assert filename in installed
     with open(spec.prefix.bin.join(filename), "r") as f:
         content = f.read().strip()
-    assert content == "[]"  # contents are weird for another test
+    assert content == "[0]"  # contents are weird for another test
```
haampie, line 279: Ok, this comment already existed, but it's not particularly helpful for understanding what is tested, and why it's changed now.

lib/spack/spack/test/conftest.py
```diff
@@ -159,7 +159,9 @@ def latest_commit():
             return git("rev-list", "-n1", "HEAD", output=str, error=str).strip()
 
         # Add two commits on main branch
-        write_file(filename, "[]")
+
+        # A commit without a previous version counts as "0"
+        write_file(filename, "[0]")
```
haampie, line 164: ah, okay. weird.

lib/spack/spack/test/spec_syntax.py
```diff
@@ -1040,18 +1040,44 @@ def test_compare_abstract_specs():
...
+        ),
+        # Git tag vs. generic develop
+        (
+            "develop-branch-version@git.0.2.15=develop",
+            "develop-branch-version@develop",
+            (True, True, False),
+        ),
+    ],
+)
+def test_git_ref_spec_equivalences(mock_packages, lhs_str, rhs_str, expected):
```
haampie, line 1072: This makes sense to me, so commit shas are just a stricter specification of the version.

lib/spack/spack/test/variant.py
haampie: Looks correct now

lib/spack/spack/variant.py
```diff
@@ -349,6 +348,9 @@ def satisfies(self, other):
         # (`foo=bar` will never satisfy `baz=bar`)
         return other.name == self.name
 
+    def intersects(self, other):
+        return self.compatible(other)
+
     @implicit_variant_conversion
     def compatible(self, other):
         """Returns True if self and other are compatible, False otherwise.
```
haampie, line 358: Minor nitpick: can you make `compatible` call `intersects` and have the implementation in `intersects` for consistency? The docstring probably should also say "returns true if there are concrete specs matching both lhs and rhs"

lib/spack/spack/version.py
```diff
@@ -348,11 +348,25 @@ def isdevelop(self):
...
+    def intersects(self, other: "VersionBase") -> bool:
+        """Return True if self intersects with other, False otherwise.
+
+        Two versions intersect if one can be constrained by the other. For instance
+        @4.7 and @4.7.3 intersect (the intersection being @4.7.3).
+
+        Arg:
+            other: version to be checked for intersection
+        """
+        return self.satisfies(other) or other.satisfies(self)
```
haampie: Hm yeah, so this works because version @x interpreted as @x:x means that @x and @y have overlap if and only if one is fully contained in the other... not very pretty

haampie: I would implement it as 

```python
n = min(len(self.version), len(other.version))
return self.version[:n] == other.version[:n]
```

since it's faster and doesn't leave future reader wonder if it's possible for neither v1 in v2 nor v2 in v1 while still having non-empty intersection.

lib/spack/spack/version.py
```diff
@@ -587,14 +604,14 @@ def _cmp(self, other_lookups=None):
 
         return self.version
 
+    @coerced
+    def overlaps(self, other):
```
haampie: overlaps -> intersects?

lib/spack/spack/version.py
```diff
@@ -587,14 +604,14 @@ def _cmp(self, other_lookups=None):
...
+        return self.satisfies(other) or other.satisfies(self)
```
haampie: Same here, I personally find it a confusing implementation because for nonempty `x` and `y` if `(x ⊂ y) ∨ (x ⊃ y)` then `x ∩ y ≠ Ø` but not the other way around.

alalazo: Yes, you're correct in general. I think it's true for `VersionBase` objects but I can change that to be more explicit.

lib/spack/spack/version.py
```diff
@@ -805,26 +822,32 @@ def __contains__(self, other):
...
+    @coerced
+    def satisfies(self, other):
+        """A version range satisfies another if it is a subset of the other.
+
+        Examples:
+        - 1:2 does not satisfy 3:4, as their intersection is empty.
+        - 1:3 does not satisfy 2:4, as they overlap but neither is a subset of the other
+        - 1:3 satisfies 1:4.
+        """
+        return self.overlaps(other) and self.intersection(other) == self
```
haampie: Is computing `self.intersection(other)` too expensive to just do `return self.intersection(other) == self`?

haampie: duh, intersection calls overlaps itself

haampie: I guess the expensiveness depends on whether this returns true or false more often in practice.

lib/spack/spack/version.py
```diff
@@ -1040,31 +1065,12 @@ def from_dict(dictionary):
...
-        essentially the same algorithm as overlaps() does for
-        VersionList, but it calls satisfies() on member Versions
-        and VersionRanges.
-
-        If strict is specified, this version list must lie entirely
-        *within* the other in order to satisfy it.
-        """
-        if not other or not self:
+    def satisfies(self, other) -> bool:
+        for x in self:
```
haampie: This can be one-liner now:

```python
return all(any(lhs.satisfies(rhs) for rhs in other) for lhs in self)
```

Also, please add a comment here saying that this exploits the invariant that VersionLists are always simplified. So VersionList([1:3, 2:4]) is not possible, only VersionList([1:4]) is.

(in principle it could also exploit the order of version list, but maybe better to just stay correct now)

lib/spack/spack/test/spec_syntax.py
alalazo: Yes, that is the idea here.

lib/spack/spack/test/cmd/install.py
alalazo: I'm not really sure either if I got the spirit of this tests 100% correctly. The change here is done following:

https://github.com/spack/spack/blob/0aeaa35b39045131439e45f0b2c5df5c2c8d77a6/lib/spack/spack/version.py#L595-L596

which in turn is there to maintain the semantic of the following test:

https://github.com/spack/spack/blob/0aeaa35b39045131439e45f0b2c5df5c2c8d77a6/lib/spack/spack/test/versions.py#L614-L615

Since what we write in the mock-git repo and we test for content is rather arbitrary, I thought this change here is allowed. @becker33 can confirm if that is right.

lib/spack/spack/test/conftest.py
alalazo: See https://github.com/spack/spack/pull/35681#discussion_r1122232365 The git version was satisfying `@:0` so I tried to maintain the same high-level semantic, though I think trying to map commit hash to version ranges is not doable in general.

lib/spack/spack/fetch_strategy.py
alalazo: The issue here is the following. When we initialize a `Spec` just by name we do:

https://github.com/spack/spack/blob/18e0b893f268443d4d285bcfa7c071f4838c5557/lib/spack/spack/spec.py#L1265

There are contexts in which we want to extrapolate a version `url` from an abstract spec with just the name. Sometimes this means calling `url_for_version`, which has an unfortunate API:

https://github.com/spack/spack/blob/18e0b893f268443d4d285bcfa7c071f4838c5557/var/spack/repos/builtin/packages/cddlib/package.py#L24-L29

where we pass the version as a string. Thus for any comparison we relied on `satisfies(..., strict=False)` i.e. the current `intersects`. 

I think we don't want to have changes in packages to fit an implementation of extrapolation that can be improved later, and `satisfies` gives the same result as `intersects` for concrete specs. Thus, in contexts where we call this function with an abstract spec I currently hack the version to make it satisfy the correct branch in `url_for_version`.

lib/spack/spack/version.py
alalazo: Done in https://github.com/spack/spack/pull/35681/commits/76d4ef974669290b9d8026f8bb4fbf29c292dff2

lib/spack/spack/version.py
alalazo: Done in https://github.com/spack/spack/pull/35681/commits/76d4ef974669290b9d8026f8bb4fbf29c292dff2

lib/spack/spack/version.py
alalazo: Done in https://github.com/spack/spack/pull/35681/commits/76d4ef974669290b9d8026f8bb4fbf29c292dff2

lib/spack/spack/version.py
alalazo: Done in https://github.com/spack/spack/pull/35681/commits/76d4ef974669290b9d8026f8bb4fbf29c292dff2

lib/spack/spack/version.py
alalazo: Done in https://github.com/spack/spack/pull/35681/commits/76d4ef974669290b9d8026f8bb4fbf29c292dff2

lib/spack/spack/variant.py
alalazo: Done in https://github.com/spack/spack/pull/35681/commits/bebaf82a9fee74cb45e746ef67d27e725b2ede76

lib/spack/spack/fetch_strategy.py
alalazo: Added a comment and an assert (for safety) in https://github.com/spack/spack/pull/35681/commits/a1c1ecef7851343413af7810b8326a3e5ef95d55

lib/spack/spack/version.py
```diff
@@ -882,33 +916,32 @@ def union(self, other):
...
+            if other.start is not None:
+                if other.start > start or other.start in start:
+                    start = other.start
 
+        if self.end is None:
+            end = other.end
         else:
-            return VersionList()
+            end = self.end
+            # TODO: does this make sense?
```
haampie, line 934: This simplifies if `self.next = next_version(self.end)` is stored, then it's just intersection of half-open ranges [x, y). For the example below with ranges :1.6.5 and :1.6 you can just store 1.6.6 and 1.7 and use <. (In fact you can use an explicit value for -inf and inf instead of None, so simplify it even more)

haampie: Anyways, for later... I remember I implemented that once.

var/spack/repos/builtin/packages/macsio/package.py
```diff
@@ -21,7 +21,7 @@ class Macsio(CMakePackage):
     version("1.0", sha256="1dd0df28f9f31510329d5874c1519c745b5c6bec12e102cea3e9f4b05e5d3072")
 
     variant("mpi", default=True, description="Build MPI plugin")
-    variant("silo", default=True, description="Build with SILO plugin")
+    variant("silo", default=False, description="Build with SILO plugin")
```
haampie, line 24: Unrelated?

lib/spack/spack/test/spec_semantics.py
```diff
@@ -41,23 +43,23 @@ def _specify(spec_like):
     return Spec(spec_like)
 
 
-def check_satisfies(target_spec, constraint_spec, target_concrete=False):
+def constrain_or_raise(target_spec, constraint_spec, target_concrete=False):
```
haampie: Can this be called `check_can_constrain` instead?

lib/spack/spack/test/spec_semantics.py
```diff
@@ -41,23 +43,23 @@ def _specify(spec_like):
...
     target = make_spec(target_spec, target_concrete)
     constraint = _specify(constraint_spec)
 
     # Satisfies is one-directional.
-    assert target.satisfies(constraint)
+    assert target.intersects(constraint)
 
     # If target satisfies constraint, then we should be able to constrain
```
haampie:
```suggestion
    # If target intersects constraint, then we should be able to constrain
```

lib/spack/spack/test/spec_semantics.py
```diff
@@ -41,23 +43,23 @@ def _specify(spec_like):
     return Spec(spec_like)
 
 
-def check_satisfies(target_spec, constraint_spec, target_concrete=False):
+def constrain_or_raise(target_spec, constraint_spec, target_concrete=False):
     target = make_spec(target_spec, target_concrete)
     constraint = _specify(constraint_spec)
 
     # Satisfies is one-directional.
```
haampie:
```suggestion
```

lib/spack/spack/test/spec_semantics.py
```diff
@@ -41,23 +43,23 @@ def _specify(spec_like):
...
-    assert target.satisfies(constraint)
+    assert target.intersects(constraint)
 
     # If target satisfies constraint, then we should be able to constrain
     # constraint by target.  Reverse is not always true.
```
haampie:
```suggestion
```

lol

lib/spack/spack/test/spec_semantics.py
```diff
@@ -393,70 +394,70 @@ def test_unsatisfiable_variant_types(self):
...
-        check_unsatisfiable("mpich++foo", "mpich~~foo")
-        check_unsatisfiable("mpich foo==True", "mpich foo==False")
+        check_cannot_constrain("mpich~foo", "mpich+foo")
+        check_cannot_constrain("mpich+foo", "mpich~foo")
+        check_cannot_constrain("mpich foo=True", "mpich foo=False")
+        check_cannot_constrain("mpich~~foo", "mpich++foo")
+        check_cannot_constrain("mpich++foo", "mpich~~foo")
+        check_cannot_constrain("mpich foo==True", "mpich foo==False")
 
     def test_satisfies_matching_compiler_flag(self):
```
haampie: there's still a lot of left-over `satisfies` jargon in tests that use `intersect`

var/spack/repos/builtin/packages/macsio/package.py
alalazo: Kind of. This was just to make Gitlab pass on an unrelated issue that was blocking this PR. Not sure why this PR had to rebuild `macsio` in the first place.

lib/spack/spack/spec.py
```diff
@@ -364,65 +362,81 @@ def target_or_none(t):
...
-        # Check target
-        return self.target_satisfies(other, strict=strict)
+        return self._target_satisfies(other, strict=False)
 
-    def target_satisfies(self, other, strict):
-        need_to_check = (
-            bool(other.target) if strict or self.concrete else bool(other.target and self.target)
-        )
+    def _target_satisfies(self, other: "ArchSpec", strict: bool) -> bool:
+        if strict is True:
```
haampie, line 404: Should this `strict` stuff be dropped too?

alalazo: I think so, but since this is in a private function, I'd rather do it in a follow up PR. Would that be fine?

lib/spack/spack/spec.py
```diff
@@ -738,24 +774,22 @@ def __init__(self, spec):
         super(FlagMap, self).__init__()
         self.spec = spec
 
-    def satisfies(self, other, strict=False):
-        if strict or (self.spec and self.spec._concrete):
-            return all(f in self and set(self[f]) == set(other[f]) for f in other)
-        else:
+    def satisfies(self, other):
+        return all(f in self and set(self[f]) == set(other[f]) for f in other)
```
haampie: Why set? Flags are order dependent, `-ffeature -fno-feature` != `-fno-feature -ffeature`.

I think I would just stick to a interpretation of *flags being a variant with string value type. So, for satisfies you just drop `set` and for intersects you stick to the flags set in either an require equal values.

lib/spack/spack/spec.py
```diff
@@ -3479,10 +3626,7 @@ def satisfies(self, other, deps=True, strict=False):
         # 2. For non-empty intersection (strict=False) we only have a fast path
```
haampie: this comment is outdated

lib/spack/spack/spec.py
```diff
@@ -3514,76 +3656,72 @@ def satisfies(self, other, deps=True, strict=False):
         ):
             return False
         if self.versions and other.versions:
```
haampie: I think this if and the next elif is redundant if "no version" implies empty VersionList(). Note that

```python
>>> xs = []; ys = []
>>> all(any(x == y for y in ys) for x in xs)
True
```

as expected, since all(...) is true for the empty set.

So, simplify to

```python
if not self.versions.satisfies(other.versions):
    return False
```

is more concise.

lib/spack/spack/spec.py
```diff
@@ -3514,76 +3656,72 @@ def satisfies(self, other, deps=True, strict=False):
...
-            if not self.versions.satisfies(other.versions, strict=strict):
+            if not self.versions.satisfies(other.versions):
                 return False
-        elif strict and (self.versions or other.versions):
+        elif self.versions or other.versions:
             return False
 
         # None indicates no constraints when not strict.
```
haampie: more strict remnants.

lib/spack/spack/spec.py
```diff
@@ -3514,76 +3656,72 @@ def satisfies(self, other, deps=True, strict=False):
...
+            return False
 
-        elif not self._dependencies:
-            # if not strict, this spec *could* eventually satisfy the
-            # constraints on other.
-            return True
+        # use list to prevent double-iteration
+        selfdeps = list(self.traverse(root=False))
+        otherdeps = list(other.traverse(root=False))
+        if not all(any(d.satisfies(dep) for d in selfdeps) for dep in otherdeps):
```
haampie: Somewhere here we there should be a comment that for abstract specs the graph structure is more or less ignored (direct deps of abstract specs may concretize to deps at any depth...). As far as I understand, we only ever hit this when the rhs is abstract (`concrete.satisfies(abstract)`). Documenting these invariants makes it a bit easier to follow.

lib/spack/spack/spec.py
```diff
@@ -3514,76 +3656,72 @@ def satisfies(self, other, deps=True, strict=False):
...
+            return False
 
         # Handle first-order constraints directly
         for name in self.common_dependencies(other):
             if not self[name].satisfies(other[name], deps=False):
                 return False
 
+        if self.concrete:
```
haampie: This is dead code? You already checked that all rhs satisfy any lhs

haampie:
What is still not entirely right is concrete spec semantics. Sometimes it's interpreted as a singleton set, elsewhere it's a set with an infinite amount of elements. (I don't think it should be fixed now)

In a perfect world `spec.concrete is True` would really just signal it's a singleton set. So then `intersects(concrete zlib@1.2 +shared, abstract zlib@1.2.3 +non_existent_variant)` would be false because the lhs doesn't have `non_existing_variant` and 1.2 exactly is not in the range 1.2.3:1.2.3. Whereas `intersects(abstract zlib@1.2 +shared, abstract zlib@1.2.3 +non_existent_variant)` is true, because you could possibly construct `zlib@1.2.3 +shared +non_existent_variant`.

Distinguishing non-existence of a variant/prop/value is only possible when `concrete=True` is taken into account.

lib/spack/spack/spec.py
alalazo: No, this is the part that fixed #35597 originally.

alalazo:
> What is still not entirely right is concrete spec semantics. 

More from DM exchange:
- Versions semantic also has the issue that we cannot pinpoint a single item in a set
- Concrete spec semantic becomes relevant in practice when we consider concrete specs over time (e.g. retrieve a concrete spec installed before the variant `foo` was added to a package and intersect `package +foo` with concrete spec)
- Flags semantic is also kind of wrong and can probably be factored together with single valued variant in the future

lib/spack/spack/spec.py
alalazo: > As far as I understand, we only ever hit this when the rhs is abstract

Yes, since we have the fast way out:
```python
if other.concrete:
    # The left-hand side must be the same singleton with identical hash. Notice that
    # package hashes can be different for otherwise indistinguishable concrete Spec
    # objects.
    return self.concrete and self.dag_hash() == other.dag_hash()
```
I think we can possibly remove some redundancy between here and the lines below:
```python
for name in self.common_dependencies(other):
    if not self[name].satisfies(other[name], deps=False):
        return False
```
but I'll try to explore that in future PRs.

alalazo: See 00f68d94d8793e3a8e16680fa40dfb0b87e3934b

lib/spack/spack/spec.py
alalazo: These should have been removed. Let me know if there are others

lib/spack/spack/spec.py
alalazo: Done in 9e97a4487017a7a626eb045615ddd50876df5c18

lib/spack/spack/spec.py
alalazo: Updated

lib/spack/spack/spec.py
alalazo: Flags semantic has been changed in 3a5b534fa9c8cb42e71fe715ea9859a8a18d5528

lib/spack/spack/test/spec_semantics.py
alalazo: Tests have been refactored, jargon shouldn't be there anymore

lib/spack/spack/test/spec_semantics.py
alalazo: Done

lib/spack/spack/test/spec_semantics.py
alalazo: Refactored the tests, so this is outdated now.

lib/spack/spack/test/spec_semantics.py
```diff
@@ -9,280 +9,348 @@
...
-    spec.constrain(constraint)
-    assert exp == spec
+        assert lhs.intersects(rhs)
+        assert rhs.intersects(lhs)
 
+        c1, c2 = lhs.copy(), rhs.copy()
+        c1.constrain(rhs)
+        c2.constrain(lhs)
+        assert c1 == c2
+        assert c1 == expected
```
alalazo, line 240: Here I always check all the invariants we expect for each of the new parametrized tests.

lib/spack/spack/spec.py
haampie: I'm not a fan of this implementation to be honest. Why don't you replace it with a one-liner and leave the provider business to `satisfies`?

```python
def satisfies_deps(self, other):
    return all(any(lhs.satisfies(rhs, deps=False) for lhs in self.traverse(root=False)) for rhs in other.traverse(root=False))
```

alalazo: Fine with me, I tried to maintain most of the previous logic in this PR, but I agree that there are simpler implementations - in particular now that we split the `intersects` vs. `satisfies` semantic.

haampie: imo it's less likely to be incorrect, since this is pretty much the property we need (assuming the graph structure is irrelevant for the abstract rhs...)

alalazo: Ok, done in cf3a79a2e98a9881382a4d898125d5e9389ef9c0. We have to revisit how we handle structured DAGs as soon as Spack handles "separate concretization of build dependencies" and we can use syntax like:
```
hd5 %( cmake ^zlib~shared )
```
to have a non flat representation of an abstract DAG.

lib/spack/spack/spec.py
alalazo: Removed redundancy in cf3a79a2e98a9881382a4d898125d5e9389ef9c0 - so I think this one is done too.

alalazo:
I tried to time a few concretizations with the same setup on `develop`:

radiuss_develop

and with this branch:

radiuss_split

Here are the datasets:

[radiuss_develop.csv](https://github.com/spack/spack/files/10900429/radiuss_develop.csv)
[radiuss_split.csv](https://github.com/spack/spack/files/10900430/radiuss_split.csv)

I don't see significant differences.

lib/spack/spack/spec.py
haampie: nice

scheibelp:
FWIW it looks like outside of the old concretizer, that `x.satisfies(y)` is only ever called when `x` is concrete. IMO that may be useful to encode somehow (given that the abstract-vs.-concrete possibilities make things more difficult to reason about): for example `.satisfies` could (by default) *require* that `self` is concrete.

haampie:
The abstract vs concrete case isn't (very?) relevant for satisfies: if the rhs is concrete and has a dag hash, the lhs should have the same dag hash, and if only the lhs is concrete it doesn't matter how you interpret the lhs (as a singleton set with just the spec, or still an infinite set where elements have additional variants defined etc) for the result to make sense.

The abstract vs concrete case is relevant for intersects, since then the answer to "is there a more specific spec than a given concrete spec" determines whether intersects(concrete, abstract) returns true or false in some cases. If you see concrete specs as singletons, then intersects(concrete, abstract) with additional variants in abstract is false. If you see concrete specs sort of like very constrained abstract specs, then it could be true. [so for example `intersects(a@1.2.3, a+x)` where lhs is concrete and doesn't have `x` defined]

To me abstract specs as singletons make most sense, since they're elements picked from the solution space, so I would say `intersects(a@1.2.3, a+x) == false`. The only reason I don't insist for that to be fixed in this PR is because of Version(x) == VersionRange(x, x) rabbit holes :upside_down_face:

alalazo:
> To me abstract specs as singletons make most sense, since they're elements picked from the solution space, so I would say intersects(a@1.2.3, a+x) == false. **The only reason I don't insist for that to be fixed in this PR is because of Version(x) == VersionRange(x, x) rabbit hole**

Agreed, since they are basically the same issue.

haampie (approved):

alalazo:
develop (39adb65dc7f377662a2bbb3b4377b4627cf64d31, blue) vs. PR (d54611af2c56a04ed0990055d357b30b54b30a5d, orange)

[radiuss_develop.csv](https://github.com/spack/spack/files/11088136/radiuss_develop.csv)
[radiuss_35681.csv](https://github.com/spack/spack/files/11088137/radiuss_35681.csv)
[radiuss.txt](https://github.com/spack/spack/files/11088138/radiuss.txt)

No impact on concretization time:

totals

hppritcha (commented):
Looks good to me!  can't seem to check off.

tgamblin:
@hppritcha it's already merged :) @alalazo is just posting reports on concretization speed on PRs so we can track which ones may have impacted performance.

PR #35682, opened 2023-02-24
docbook: resolve conflict in spack env view

glennpj:
If the docbook packages
- docbook-xml
- docbook-xsl

are installed in a spack environment view the catalog files will be in conflict in the view directory. This PR resolves that by adding an appropriate prefix to each catalog name so that they are unique in the view. The resulting XML_CATALOG_FILES environment variable will then be able to point to both of them.


alalazo (approved):

alalazo:
@spackbot run pipeline

PR #35686, opened 2023-02-24
osi: add version 0.108.7

dlkuehn:
Add version 0.108.7 to package osi.  Also added pkg-config dependency. On hosts without pkg-config installed (e.g. Ubuntu 22.04 +build-essential), the OSI install cannot find the CoinUtils dependency and fails.


dlkuehn:
Errors made.  Will resubmit.

PR #35688, opened 2023-02-24
clp: add version 1.17.7

dlkuehn:
Add version 1.17.7 to package clp. 

Add pkg-config dependency.  On hosts without pkg-config/pkgconf installed (e.g. Ubuntu 22.04 +build-essential), the clpI install cannot find CoinUtils and fails.

### DIFF
--- a/var/spack/repos/builtin/packages/clp/package.py
+++ b/var/spack/repos/builtin/packages/clp/package.py
@@ -13,7 +13,7 @@ class Clp(AutotoolsPackage):
     homepage = "https://projects.coin-or.org/Clp"
     url = "https://github.com/coin-or/Clp/archive/releases/1.17.6.tar.gz"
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/clp/package.py
```diff
@@ -13,9 +13,11 @@ class Clp(AutotoolsPackage):
     homepage = "https://projects.coin-or.org/Clp"
     url = "https://github.com/coin-or/Clp/archive/releases/1.17.6.tar.gz"
 
+    depends_on("pkg-config")
```
alalazo:
```suggestion
    depends_on("pkg-config", type="build")
```

alalazo (changes requested):

alalazo (approved):

PR #35689, opened 2023-02-24
osi: add version 0.108.7

dlkuehn:
Add version 0.108.7 to package osi.

Add pkg-config dependency.  On hosts without pkg-config installed (e.g. Ubuntu 22.04 +build-essential), the OSI install cannot find the CoinUtils dependency and fails.

### DIFF
--- a/var/spack/repos/builtin/packages/osi/package.py
+++ b/var/spack/repos/builtin/packages/osi/package.py
@@ -20,8 +20,10 @@ class Osi(AutotoolsPackage):
     homepage = "https://projects.coin-or.org/Osi"
     url = "https://github.com/coin-or/Osi/archive/releases/0.108.6.tar.gz"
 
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/osi/package.py
```diff
@@ -20,8 +20,10 @@ class Osi(AutotoolsPackage):
     homepage = "https://projects.coin-or.org/Osi"
     url = "https://github.com/coin-or/Osi/archive/releases/0.108.6.tar.gz"
 
+    depends_on("pkg-config")
```
alalazo:
```suggestion
    depends_on("pkg-config", type="build")
```

alalazo (changes requested):

alalazo (approved):

PR #35703, opened 2023-02-25
Set GOMAXPROCS for go package to limit build processes

AlexanderRichert-NOAA:
This PR sets `$GOMAXPROCS` in the `go` package in order to limit the number of build processes based on `make_jobs`. On high-CPU count systems, the build can fail due to ulimits.

### DIFF
--- a/var/spack/repos/builtin/packages/go/package.py
+++ b/var/spack/repos/builtin/packages/go/package.py
@@ -116,6 +116,7 @@ class Go(Package):
         # internal Spack wrappers and fail.
         env.set("CC_FOR_TARGET", self.compiler.cc)
         env.set("CXX_FOR_TARGET", self.compiler.cxx)
+        env.set("GOMAXPROCS", make_jobs)
 
     def setup_dependent_package(self, module, dependent_spec):
         """Called before go modules' install() methods.


alecbcs (approved):
Looks good to me! @AlexanderRichert-NOAA do you have a copy of the error you got when attempting to build Go on one of these systems with a large number of CPU cores?

AlexanderRichert-NOAA:
@alecbcs Yep, I get upward of a dozen messages about `newosproc`:
```
runtime: failed to create new OS thread (have 10 already; errno=11)
runtime: may need to increase max user processes (ulimit -u)
fatal error: newosproc
```
The largest count it gave me in my most recent test was `have 120 already`. This is on a node with 256 logical cores (2 sockets x 64 cores/socket x 2 threads/core). My `ulimit -u` is 1024.

The final error message is:
```
go tool dist: FAILED: /path/to/cache/build_stage/spack-stage-go-1.19.6-emutec4z2cgutq7b4ku2iqx3cc2jfxti/spack-src/pkg/tool/linux_amd64/go_bootstrap install -a -i cmd/asm cmd/cgo cmd/compile cmd/link: exit status 2
```

alalazo (approved):

Issue #35546, opened 2023-02-17
Installation issue: amrex@23.02+hdf5


meyersbs:
### Steps to reproduce the issue

```console
$ spack spec -I amrex@23.03%gcc@9.3.0+hdf5
Input spec
--------------------------------
amrex@23.02%gcc@9.3.0+hdf5

Concretized
--------------------------------
ccifr6k  amrex@23.02%gcc@9.3.0~amrdata~cuda~eb~fortran+hdf5~hypre~ipo+linear_solvers+mpi~openmp~particles~petsc~pic~plotfile_tools~rocm~shared~sundials~sycl~tiny_profile build_system=cmake build_type=RelWithDebInfo dimensions=3 precision=double arch=linu
x-rhel7-skylake_avx512
nbe2xqn      ^cmake@3.25.2%gcc@9.3.0~doc+ncurses+ownlibs~qt build_system=generic build_type=Release arch=linux-rhel7-skylake_avx512
ex5yyfs          ^ncurses@6.4%gcc@9.3.0~symlinks+termlib abi=none build_system=autotools arch=linux-rhel7-skylake_avx512
3g6w4fu          ^openssl@1.1.1t%gcc@9.3.0~docs~shared build_system=generic certs=mozilla arch=linux-rhel7-skylake_avx512
au3vaca              ^ca-certificates-mozilla@2023-01-10%gcc@9.3.0 build_system=generic arch=linux-rhel7-skylake_avx512
vzu47cs              ^perl@5.36.0%gcc@9.3.0+cpanm+open+shared+threads build_system=generic arch=linux-rhel7-skylake_avx512
fsrsn6r                  ^berkeley-db@18.1.40%gcc@9.3.0+cxx~docs+stl build_system=autotools patches=26090f4,b231fcc arch=linux-rhel7-skylake_avx512
4cm2qmx                  ^bzip2@1.0.8%gcc@9.3.0~debug~pic+shared build_system=generic arch=linux-rhel7-skylake_avx512
fok6ki4                  ^gdbm@1.23%gcc@9.3.0 build_system=autotools arch=linux-rhel7-skylake_avx512
rcybyat                      ^readline@8.2%gcc@9.3.0 build_system=autotools patches=bbf97f1 arch=linux-rhel7-skylake_avx512
[...]

PR #35705, opened 2023-02-26
HDF5: is_enabled helper (ON)

ax3l:
Related to `determine_variants` for finding external HDF5 installs by querying `-showconfig` from HDF5 executables.

Slightly generalize the `is_enabled` helper in the HDF5 package. `ON` is the most typical CMake bool option passed, besides many other possible `true` values, and should be included as a possible check to the config.

Loosely related to #35546 and https://github.com/HDFGroup/hdf5/issues/2488

### DIFF
--- a/var/spack/repos/builtin/packages/hdf5/package.py
+++ b/var/spack/repos/builtin/packages/hdf5/package.py
@@ -489,9 +489,7 @@ class Hdf5(CMakePackage):
     @classmethod
     def determine_variants(cls, exes, version):
         def is_enabled(text):
-            if text in set(["t", "true", "enabled", "yes", "1"]):
-                return True
-            return False
+            return text.lower() in ["t", "true", "enabled", "yes", "1", "on"]
 
         results = []
         for exe in exes:


alalazo:
@spackbot run pipeline

hyoklee (approved):

var/spack/repos/builtin/packages/hdf5/package.py
```diff
@@ -466,7 +466,7 @@ def determine_version(cls, exe):
     @classmethod
     def determine_variants(cls, exes, version):
         def is_enabled(text):
-            if text in set(["t", "true", "enabled", "yes", "1"]):
+            if text.lower() in set(["t", "true", "enabled", "yes", "1", "on"]):
                 return True
             return False
```
tldahlgren: FWIW. You could simplify this further:

```suggestion
            return text.lower() in ["t", "true", "enabled", "yes", "1", "on"]
```

tldahlgren (dismissed):
LGTM.  Have a minor suggestion but let me know if you want this merged as-is.

hyoklee (dismissed):

var/spack/repos/builtin/packages/hdf5/package.py
ax3l: Good point:)

hyoklee (approved):

tldahlgren (approved):
LGTM. Thanks!

ax3l:
rebased against latest `develop` because CI somehow failed/was stuck.

PR #35711, opened 2023-02-27
hpctoolkit: fix broken patches

haampie:
The patches don't have a stable checksum:

```
7eb3a21c9bf9e1b65ef0375609665d36e5f7a74c664eb474cba728faada9baeb *411d62544717873432c49ef45c7cb99cc5de2fb8.patch
ccf7a319be0362d8a047953aa5d594b386acc7d6ea21fdc15925e926f83c7bd0 *511afd95b01d743edc5940c84e0079f462b2c23e.patch
```

### DIFF
--- a/var/spack/repos/builtin/packages/hpctoolkit/411d62544717873432c49ef45c7cb99cc5de2fb8.patch
+++ b/var/spack/repos/builtin/packages/hpctoolkit/411d62544717873432c49ef45c7cb99cc5de2fb8.patch
@@ -1,0 +1,91 @@
+From 411d62544717873432c49ef45c7cb99cc5de2fb8 Mon Sep 17 00:00:00 2001
+From: "Mark W. Krentel" <krentel@rice.edu>
+Date: Thu, 15 Dec 2022 16:43:43 -0600
+Subject: [PATCH] Add a temporary hack to allow both ROCM 5.2/5.3 to build
+ cleanly.
+
+There were some corner cases (build 5.3 pieces from spack and feed
+into autotools) that didn't work.  After the next release, I will want
+to rework ROCM configure more extensively.
+---
+ configure    | 21 +++++++++++++--------
+ configure.ac | 17 +++++++++++------
+ 2 files changed, 24 insertions(+), 14 deletions(-)
+
+diff --git a/configure b/configure
+index 1760e678e8..814376b3bd 100755
+--- a/configure
++++ b/configure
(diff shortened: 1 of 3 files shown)


alalazo (approved):

mwkrentel:
@haampie @alalazo 
Well, I was going to update the checksums in #35662 or #35648 (now
reverted), but whatever.

So, does this mean that we should never use URL patches from gitlab?

And in the future, I would prefer you not to submit, approve and merge
patches to hpctoolkit all while I (as maintainer) am asleep.

In this case, the better fix is to delete the 411 patch and replace it
with a conflict as I did in #35662.  But it's short and I'll fix it in
#35662.

alalazo:
> And in the future, I would prefer you not to submit, approve and merge patches to hpctoolkit all while I (as maintainer) am asleep.

@mwkrentel Apologies, that seemed a trivial fix for something that was reported broken since last Wednesday - so I went ahead and merged as requested by @haampie I'll keep that in mind that you prefer to review all changes to `hpctoolkit` in the future. Also, let me know if you have other preferred solutions, I'll try to prioritize reviewing them.

alalazo:
The other thing to note is that, since `hpctoolkit` is in pipelines, failure to verify the checksum of patches means failure to pass pipelines tests.

mwkrentel:
@alalazo I mainly wanted a chance to discuss updating the checksums
vs copying the patch.   Does this mean that we should never use a URL
patch from gitlab?

alalazo:
> I mainly wanted a chance to discuss updating the checksums vs copying the patch

The chance is still there, and we can revert and go back to a URL if we find a way to have a stable shasum. This was mainly to avoid failures in pipelines due to a changing shasum. I didn't check yet if Gitlab has something similar to `?full_index=1` for GitHub

mwkrentel:
When it comes to code and commits, I'm like John Roberts (Chief
Justice, US Supreme Court).  I prefer to make the smallest change that
gets the job done, unless there's a compelling reason to do otherwise.

In this case, if it were me, I would have just updated the sha256 sums
as the smaller change.  But now that they're switched to actual
patches, I prefer to leave it as is.  I'm going to delete one of them
in a couple days anyway.

haampie:
I'm sure GitLab will at some point produce a slightly different diff
with one whitespace or line of context more ore less, etc. It'd be
better if we as a general rule would not allow any dynamically
generated content in source and patch urls.

PR #35713, opened 2023-02-27
cgns: enable tools

nicolaslg:
- added a `tools` variant that activates the option `CGNS_BUILD_CGNSTOOLS` and the necessary dependencies
- added a patch for older versions of cgns (see https://bugs.gentoo.org/662210)


alalazo (approved):
LGTM. Waiting for maintainer's review.

PR #35717, opened 2023-02-27
py-wheel: add v0.38.4

Sangu-Mbekelu:
updating py-wheel package for newer versions


Sangu-Mbekelu:
@spackbot fix style

Sangu-Mbekelu:
@spackbot fix style

Sangu-Mbekelu:
Can't get around this style error

tldahlgren (dismissed):
Confirmed sha256.

var/spack/repos/builtin/packages/py-wheel/package.py
```diff
@@ -15,6 +15,12 @@ class PyWheel(Package, PythonExtension):
     )
     list_url = "https://pypi.org/simple/wheel/"
 
+    version(
+        "0.38.4",
+        url="https://files.pythonhosted.org/packages/bd/7c/d38a0b30ce22fc26ed7dbc087c6d00851fb3395e9d0dac40bec1f905030c/wheel-0.38.4-py3-none-any.whl",
```
adamjstewart: Remove this line, update the URL above to https://files.pythonhosted.org/packages/py3/w/wheel/wheel-0.38.4-py3-none-any.whl, and add the following function before install:
```python
def url_for_version(self, version):
    url = "https://files.pythonhosted.org/packages/{0}/w/wheel/wheel-{1}-{0}-none-any.whl"
    if version >= Version("0.38"):
        python = "py3"
    else:
        python = "py2.py3"
    return url.format(python, version)
```

adamjstewart (changes requested):

Sangu-Mbekelu:
@spackbot fix style

adamjstewart:
You lost the new version. I only meant to remove the url override for that version.

adamjstewart:
Now you added back the URL override and undid the base class URL change. You'll want to undo this.

Sangu-Mbekelu:
@spackbot fix style

adamjstewart (dismissed):

PR #35615, opened 2023-02-22
Update cosma recipe

mtaillefumier:
- Update cosma recipe
- Update costa recipe
- Add tiled-mm recipe (cosma hard dependency when CUDA/ROCM is used)
- Add semiprof : optional dependency

authors : @simonpintarelli , @mtaillefumier

PR #35718, opened 2023-02-27
Update cp2k recipe to use cmake or the current build system

mtaillefumier:
Offers the possibility to build cp2k with the new cmake build system. commands like this are now supported

spack install cp2k@master build_system=cmake +.....

the recipe supports the following optional functionalities

- superlu, cosma, sirius, spglib, metis, spglib, libxc, libint, cuda/rocm, mkl/openblas/sci (and others), mpi, openmp, dbcsr
- dbcsr is built separately using the currently available recipe.

Two PRs need to be merged to be fully functional (cosma update in spack + one PR in cp2k github).

NB : to be fully functional two PRs need to be merged https://github.com/spack/spack/pull/35615 and one in cp2k repo. It is a temporary limitation at that point but I would like to receive feedback about recipe independently of the two PRs.

limitations are +cosma and +plumed are not properly working: +cosma is fixed (fix by https://github.com/spack/spack/pull/35615), +plumed is a typo in cp2k cmake build system. It is also better to use the cmake build system with the current master as it includes important fix for the cmake build system.

### DIFF
--- a/var/spack/repos/builtin/packages/cp2k/package.py
+++ b/var/spack/repos/builtin/packages/cp2k/package.py
@@ -896,8 +896,7 @@ class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
         args += [
             self.define(
                 "CP2K_ENABLE_ELPA_OPENMP_SUPPORT",
-                (spec.satisfies("+openmp") and spec.satisfies("+elpa"))
-                or spec.satisfies("^elpa+openmp"),
+                ("+elpa +openmp" in spec) or ("^elpa +openmp" in spec),
             )
         ]
 
--- a/var/spack/repos/builtin/packages/cp2k/package.py
+++ b/var/spack/repos/builtin/packages/cp2k/package.py
@@ -52,13 +52,13 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
     )
     variant("plumed", default=False, description="Enable PLUMED support")
     variant(
-        "libint", default=True,  description="Use libint, required for HFX (and possibly others)",
+        "libint", default=True, description="Use libint, required for HFX (and possibly others)"
     )
-    variant("libxc", default=True,  description="Support additional functionals via libxc",)
+    variant("libxc", default=True, description="Support additional functionals via libxc")
     variant(
         "pexsi",
         default=False,
-        description="Enable the alternative PEXSI method" "for density matrix evaluation",
+        description="Enable the alternative PEXSI method for density matrix evaluation",
     )
     variant(
         "elpa",
@@ -69,24 +69,23 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
     variant(
         "sirius",
         default=False,
-        description="Enable planewave electronic structure" " calculations via SIRIUS",
+        description="Enable planewave electronic structure calculations via SIRIUS",
     )
     variant("cosma", default=False, description="Use COSMA for p?gemm")
     variant(
         "libvori",
         default=False,
-        description="Enable support for Voronoi integration" " and BQB compression",
+        description="Enable support for Voronoi integration and BQB compression",
     )
     variant("spglib", default=False, description="Enable support for spglib")
     variant(
         "spla",
         default=False,
-        description="Use SPLA off-loading functionality. Only relevant when CUDA or ROCM are enabled",
+        description="Use SPLA off-loading functionality. Only"
+        " relevant when CUDA or ROCM are enabled",
     )
     variant("pytorch", default=False, description="Enable libtorch support")
-    variant("metis", default=False, description="Enable (par)metis support")
     variant("quip", default=False, description=("Enable quip support"))
-    variant("superlu-dist", default=False, description=("Enable superlu support"))
 
     with when("+cuda"):
         variant(
@@ -168,7 +167,7 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
         depends_on("libxc@4.0.3:4", when="@7.0:8.1")
         depends_on("libxc@5.1.3:5.1", when="@8.2:8")
         depends_on("libxc@5.1.7:5.1", when="@9:2022")
-        depends_on("libxc@6:6.1", when="@2023:", type=("build", "run"))
+        depends_on("libxc@6:6.1", when="@2023:", type=("build"))
 
     with when("+mpi"):
         depends_on("mpi@2:")
@@ -176,7 +175,7 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
         depends_on("scalapack")
 
     with when("+cosma"):
-        depends_on("cosma+scalapack", type=("build", "run"))
+        depends_on("cosma+scalapack", type=("build"))
         depends_on("cosma@2.5.1:", when="@9:")
         depends_on("cosma@2.6.3:", when="@master:")
         depends_on("cosma+cuda", when="+cuda")
@@ -220,9 +219,9 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
         conflicts("@:6")
 
     with when("+libvori"):
-        depends_on("libvori@201219:", when="@8.1", type=("build", "run"))
-        depends_on("libvori@210412:", when="@8.2:", type=("build", "run"))
-        depends_on("libvori@220621:", when="@2023.1:", type=("build", "run"))
+        depends_on("libvori@201219:", when="@8.1", type=("build"))
+        depends_on("libvori@210412:", when="@8.2:", type=("build"))
+        depends_on("libvori@220621:", when="@2023.1:", type=("build"))
         # libvori support was introduced in 8+
         conflicts("@:7")
 
@@ -240,9 +239,9 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
 
     with when("build_system=cmake"):
         depends_on("dbcsr")
-        depends_on("dbcsr+openmp", when="+openmp", type=("build", "run"))
-        depends_on("dbcsr+cuda", when="+cuda", type=("build", "run"))
-        depends_on("dbcsr+rocm", when="+rocm", type=("build", "run"))
+        depends_on("dbcsr+openmp", when="+openmp", type=("build"))
+        depends_on("dbcsr+cuda", when="+cuda", type=("build"))
+        depends_on("dbcsr+rocm", when="+rocm", type=("build"))
 
     # CP2K needs compiler specific compilation flags, e.g. optflags
     conflicts("%apple-clang")
@@ -255,14 +254,26 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
     # for optimal kernels. Note that we don't override the cuda_archs property
     # from the parent class, since the parent class defines constraints for all
     # versions. Instead just mark all unsupported cuda archs as conflicting.
-    dbcsr_cuda_archs = ("35", "37", "60", "70")
-    cuda_msg = "cp2k only supports cuda_arch {0}".format(dbcsr_cuda_archs)
 
-    for arch in CudaPackage.cuda_arch_values:
-        if arch not in dbcsr_cuda_archs:
-            conflicts("+cuda", when="cuda_arch={0}".format(arch), msg=cuda_msg)
+    supported_cuda_arch_list = ("35", "37", "60", "70")
+    gpu_map = {
+        "35": "K40",
+        "37": "K80",
+        "60": "P100",
+        "70": "V100",
+        "gfx906": "Mi50",
+        "gfx908": "Mi100",
+        "gfx90a": "Mi250x",
+    }
+    cuda_msg = "cp2k only supports cuda_arch {0}".format(supported_cuda_arch_list)
 
-    conflicts("+cuda", when="cuda_arch=none", msg=cuda_msg)
+    conflicts("+cuda", when="cuda_arch=none")
+    # rocm already returns an error if amdgpu_target is not defined
+
+    with when("+cuda"):
+        for arch in CudaPackage.cuda_arch_values:
+            if arch not in supported_cuda_arch_list:
+                conflicts("+cuda", when="cuda_arch={0}".format(arch), msg=cuda_msg)
 
     # Fix 2- and 3-center integral calls to libint
     patch(
@@ -335,7 +346,6 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
         nvflags = ["-O3"]
         ldflags = []
         libs = []
-        gpuver = ""
 
         # CP2K Makefile doesn't set C standard, but the source code uses
         # C99-style for-loops with inline definition of iterating variable.
@@ -558,6 +568,7 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
             fcflags += ["-I{0}".format(sirius.prefix.include.sirius)]
             libs += list(sirius.libs)
 
+        gpuver = ""
         if spec.satisfies("+cuda"):
             libs += [
                 "-L{}".format(spec["cuda"].libs.directories[0]),
@@ -599,12 +610,25 @@ class Cp2k(MakefilePackage, CudaPackage, CMakePackage, ROCmPackage):
                     cppflags += ["-D__PW_CUDA"]
                     libs += ["-lcufft", "-lcublas"]
 
-            cuda_arch = spec.variants["cuda_arch"].value[0]
-            if cuda_arch:
-                gpuver = {"35": "K40", "37": "K80", "60": "P100", "70": "V100"}[cuda_arch]
+            gpuver = gpu_map[spec.variants["cuda_arch"].value]
+            if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
+                gpuver = "K20X"
 
-                if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
-                    gpuver = "K20X"
+        if spec.satisfies(+rocm) and spec.satisfies("@2022:"):
+            libs += [
+                "-L{}".format(spec["rocm"].libs.directories[0]),
+                "-L{}/stubs".format(spec["rocm"].libs.directories[0]),
+                "-lhipblas",
+                "-lhipfft",
+                "-lstdc++",
+            ]
+
+            cppflags += ["-D__OFFLOAD_HIP"]
+            acc_compiler_var = "hipcc"
+            acc_flags_var = "NVFLAGS"
+            cppflags += ["-D__ACC"]
+            cppflags += ["-D__DBCSR_ACC"]
+            gpuver = gpu_map[spec.variants["amdgpu_target"].value]
 
         if "smm=libsmm" in spec:
             lib_dir = join_path("lib", self.makefile_architecture, self.makefile_version)
@@ -804,21 +828,20 @@ class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
 
         args = []
 
+        if spec.satisfies("+cuda") or spec.satisfies("+rocm"):
+            if spec.satisfies("cuda_target"):
+                args += [self.define("CP2K_WITH_GPU", gpu_map[spec.variants["cuda_arch"].value])]
+
+            if spec.satisfies("amdgpu_target"):
+                args += [
+                    self.define("CP2K_WITH_GPU", gpu_map[spec.variants["amdgpu_target"].value])
+                ]
+
         if spec.satisfies("+cuda"):
-            cuda_arch = self.spec.variants["cuda_arch"].value[0]
-
-            gpu_map = {"35": "K40", "37": "K80", "60": "P100", "70": "V100", "80": "A100"}
-
-            gpuver = gpu_map[cuda_arch]
-
-            args += ["-DCP2K_WITH_GPU=%s" % gpuver, "-DCP2K_USE_ACCEL=cuda"]
+            args += ["-DCP2K_USE_ACCEL=cuda"]
 
         if spec.satisfies("+rocm"):
-            amd_arch = self.spec.variants["amdgpu_target"].value[0]
-
-            gpuver = {"gfx906": "Mi50", "gfx908": "Mi100", "gfx90a": "Mi250x"}[amd_arch]
-
-            args += ["-DCP2K_WITH_GPU={0}".format(gpuver), "-DCP2K_USE_ACCEL=hip"]
+            args += ["-DCP2K_USE_ACCEL=hip"]
 
         args += [
             self.define_from_variant("CP2K_USE_ELPA", "elpa"),
@@ -833,11 +856,11 @@ class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
             args += ["-DCP2K_USE_LIBTORCH=ON"]
 
         args += [
-            self.define_from_variant("CP2K_USE_METIS", "metis"),
+            self.define_from_variant("CP2K_USE_METIS", "pexsi"),
+            self.define_from_variant("CP2K_USE_SUPERLU", "pexsi"),
             self.define_from_variant("CP2K_USE_PLUMED", "plumed"),
             self.define_from_variant("CP2K_USE_SPGLIB", "spglib"),
             self.define_from_variant("CP2K_USE_VORI", "libvori"),
-            self.define_from_variant("CP2K_USE_SUPERLU", "superlu-dist"),
             self.define_from_variant("CP2K_USE_SPLA", "spla"),
             self.define_from_variant("CP2K_USE_QUIP", "quip"),
         ]
@@ -854,38 +877,27 @@ class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
         if "fftw" in spec:
             args += ["-DCP2K_USE_FFTW=ON"]
 
-        with when("smm=libxsmm"):
+        if spec.satisfies("smm=libxsmm"):
             args += ["-DCP2K_USE_LIBXSMM=ON"]
-
-        if spec["blas"].name in ["intel-mkl", "intel-parallel-studio", "cray-libsci", "openblas"]:
-            if spec["blas"].name in ["intel-mkl", "intel-parallel-studio"]:
-                args += ["-DCP2K_BLAS_VENDOR=MKL"]
-                args += ["-DCP2K_SCALAPACK_VENDOR=MKL"]
-
-            if "^cray-libsci" in spec:
-                args += ["-DCP2K_BLAS_VENDOR=SCI"]
-                args += ["-DCP2K_SCALAPACK_VENDOR=SCI"]
-
-            if "^openblas" in spec:
-                args += ["-DCP2K_BLAS_VENDOR=OpenBLAS"]
-                args += ["-DCP2K_SCALAPACK_VENDOR=GENERIC"]
         else:
-            lapack = spec["lapack"]
-            blas = spec["blas"]
+            args += ["-DCP2K_USE_LIBXSMM=OFF"]
 
-            args.extend(
-                [
-                    self.define("CP2K_LAPACK_FOUND", "true"),
-                    self.define("CP2K_LAPACK_LINK_LIBRARIES", lapack.libs.joined(";")),
-                    self.define("CP2K_BLAS_FOUND", "true"),
-                    self.define("CP2K_BLAS_LINK_LIBRARIES", blas.libs.joined(";")),
-                    self.define("CP2K_SCALAPACK_FOUND", "true"),
-                    self.define("CP2K_SCALAPACK_INCLUDE_DIRS", spec["scalapack"].prefix.include),
-                    self.define(
-                        "CP2K_SCALAPACK_LINK_LIBRARIES", spec["scalapack"].libs.joined(";")
-                    ),
-                ]
-            )
+        lapack = spec["lapack"]
+        blas = spec["blas"]
+
+        args.extend(
+            [
+                self.define("CP2K_LAPACK_FOUND", "true"),
+                self.define("CP2K_LAPACK_LINK_LIBRARIES", lapack.libs.joined(";")),
+                self.define("CP2K_BLAS_FOUND", "true"),
+                self.define("CP2K_BLAS_LINK_LIBRARIES", blas.libs.joined(";")),
+                self.define("CP2K_SCALAPACK_FOUND", "true"),
+                self.define("CP2K_SCALAPACK_INCLUDE_DIRS", spec["scalapack"].prefix.include),
+                self.define("CP2K_BLAS_VENDOR", "CUSTOM"),
+                self.define("CP2K_SCALAPACK_VENDOR", "GENERIC"),
+                self.define("CP2K_SCALAPACK_LINK_LIBRARIES", spec["scalapack"].libs.joined(";")),
+            ]
+        )
 
         return args
 
--- a/var/spack/repos/builtin/packages/cp2k/package.py
+++ b/var/spack/repos/builtin/packages/cp2k/package.py
@@ -893,7 +893,7 @@ class CMakeBuilder(spack.build_systems.cmake.CMakeBuilder):
         lapack = spec["lapack"]
         blas = spec["blas"]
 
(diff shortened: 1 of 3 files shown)


var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -44,15 +50,25 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+        "libxc",
+        default=True,
+        description="Support additional functionals via libxc",
     )
-    variant("libxc", default=True, description="Support additional functionals via libxc")
     variant(
         "pexsi",
         default=False,
-        description=("Enable the alternative PEXSI method" "for density matrix evaluation"),
+        description = "Enable the alternative PEXSI method" "for density matrix evaluation",
```
dev-zero:
```suggestion
        description = "Enable the alternative PEXSI method for density matrix evaluation",
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +79,48 @@ class Cp2k(MakefilePackage, CudaPackage):
     variant(
         "sirius",
         default=False,
-        description=("Enable planewave electronic structure" " calculations via SIRIUS"),
+        description = "Enable planewave electronic structure" " calculations via SIRIUS",
```
dev-zero:
```suggestion
        description = "Enable planewave electronic structure calculations via SIRIUS",
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +79,48 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+        "cosma",
+        default=False,
+        description="Use COSMA for p?gemm",
     )
-    variant("cosma", default=False, description="Use COSMA for p?gemm")
     variant(
         "libvori",
         default=False,
-        description=("Enable support for Voronoi integration" " and BQB compression"),
+        description="Enable support for Voronoi integration" " and BQB compression",
```
dev-zero:
```suggestion
        description="Enable support for Voronoi integration and BQB compression",
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +79,48 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+    variant(
+        "pytorch",
+        default=False,
+        description="Enable libtorch support",
+    )
+    variant(
+        "metis",
+        default=False,
+        description="Enable (par)metis support",
+    )
```
dev-zero: NACK. CP2K only uses (par-)metis via PEXSI, hence a variant does not make sense.

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +79,48 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+    variant(
+        "quip",
+        default=False,
+        description=("Enable quip support"),
+    )
+    variant(
+        "superlu-dist",
+        default=False,
+        description=("Enable superlu support"),
     )
```
dev-zero: NACK, same as for metis.

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -153,16 +202,17 @@ class Cp2k(MakefilePackage, CudaPackage):
         depends_on("libxc@4.0.3:4", when="@7.0:8.1")
         depends_on("libxc@5.1.3:5.1", when="@8.2:8")
         depends_on("libxc@5.1.7:5.1", when="@9:2022")
-        depends_on("libxc@6:6.1", when="@2023:")
+        depends_on("libxc@6:6.1", when="@2023:", type=("build", "run"))
```
dev-zero: Why should libxc not get added to the linker?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -153,16 +202,17 @@ class Cp2k(MakefilePackage, CudaPackage):
...
 
     with when("+mpi"):
         depends_on("mpi@2:")
         depends_on("mpi@3:", when="@2023.1:")
         depends_on("scalapack")
 
     with when("+cosma"):
-        depends_on("cosma+scalapack")
+        depends_on("cosma+scalapack", type=("build", "run"))
```
dev-zero: Why should cosma not get added to the linker?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -220,6 +272,12 @@ class Cp2k(MakefilePackage, CudaPackage):
     # in the stage of cp2k@4.1
     depends_on("wannier90", when="@3.0+mpi", type="build")
 
+    with when("build_system=cmake"):
+        depends_on("dbcsr")
+        depends_on("dbcsr+openmp", when="+openmp", type=("build", "run"))
+        depends_on("dbcsr+cuda", when="+cuda", type=("build", "run"))
+        depends_on("dbcsr+rocm", when="+rocm", type=("build", "run"))
+
```
dev-zero, line 258: same here, why the limitation to build and run but not link?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -778,3 +836,106 @@ def check(self):
...
+            gpu_map = {
+                "35": "K40",
+                "37": "K80",
+                "60": "P100",
+                "70": "V100",
+                "80": "A100",
+            }
+
+            gpuver = gpu_map[cuda_arch]
```
dev-zero: factor out and share with the makefile one

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -778,3 +836,106 @@ def check(self):
...
+        if "fftw" in spec:
+            args += [ "-DCP2K_USE_FFTW=ON" ]
+
+        with when("smm=libxsmm"):
+            args += [ "-DCP2K_USE_LIBXSMM=ON" ]
+
+        if spec["blas"].name in ["intel-mkl", "intel-parallel-studio", "cray-libsci", "openblas"]:
+            if spec["blas"].name in ["intel-mkl", "intel-parallel-studio"]:
+                args += [ "-DCP2K_BLAS_VENDOR=MKL" ]
+                args += [ "-DCP2K_SCALAPACK_VENDOR=MKL" ]
```
dev-zero: Please separate blas and scalapack here, on macOS for example there is no Intel MPI and no ScaLAPACK, hence the combination of MKL for BLAS/LAPACK and netlib-scalapack for scalapack.

dev-zero: In general I wonder whether this kind of explicit spec for the scientific libs is still necessary with CMake or whether Spack provides a restricted enough environment for CMake to auto-discover them correctly?

dev-zero (changes requested):

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +69,24 @@ class Cp2k(MakefilePackage, CudaPackage):
...
     variant("spglib", default=False, description="Enable support for spglib")
+    variant(
+        "spla",
+        default=False,
+        description="Use SPLA off-loading functionality. Only relevant when CUDA or ROCM are enabled",
+    )
+    variant("pytorch", default=False, description="Enable libtorch support")
+    variant("metis", default=False, description="Enable (par)metis support")
+    variant("quip", default=False, description=("Enable quip support"))
+    variant("superlu-dist", default=False, description=("Enable superlu support"))
```
dev-zero:
```suggestion
    variant("pytorch", default=False, description="Enable libtorch support")
    variant("quip", default=False, description=("Enable quip support"))
```

CP2K uses metis and superlu only via PEXSI, variant does not make sense.

var/spack/repos/builtin/packages/cp2k/package.py
mtaillefumier: my guess is that I do not really understand what these options are triggering. I will correct all of these.

var/spack/repos/builtin/packages/cp2k/package.py
mtaillefumier: Yes cmake autodetects them automatically. It will take the first one it finds on a system I usually specify things to help cmake finding the dependencies but in the spack case this is known. 

In some case it is necessary to specify these (essl for instance). I have for instance five different implementations of blas / lapack on my system. So specifying one version allows me to test things. 

CP2K cmake build system will still generate a warning if it find mkl but scalapack is set to its default value. So When we know that we use MKL better specify the scalapack version as well. 

macos ? Not sure to understand what's wrong here.

dev-zero: You are assuming here that if MKL is used for BLAS it should also be used for ScaLAPACK.
Which is at least on macOS not possible.

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -778,3 +802,106 @@ def check(self):
...
+        if "fftw" in spec:
+            args += ["-DCP2K_USE_FFTW=ON"]
+
+        with when("smm=libxsmm"):
+            args += ["-DCP2K_USE_LIBXSMM=ON"]
+
+        if spec["blas"].name in ["intel-mkl", "intel-parallel-studio", "cray-libsci", "openblas"]:
+            if spec["blas"].name in ["intel-mkl", "intel-parallel-studio"]:
+                args += ["-DCP2K_BLAS_VENDOR=MKL"]
+                args += ["-DCP2K_SCALAPACK_VENDOR=MKL"]
```
dev-zero:
```suggestion
            if spec["blas"].name in ["intel-mkl", "intel-parallel-studio"]:
                args += ["-DCP2K_BLAS_VENDOR=MKL"]
            if spec["scalapack"].name in ["intel-mkl", "intel-parallel-studio"]:
                args += ["-DCP2K_SCALAPACK_VENDOR=MKL"]
```

dev-zero: theoretically, this also applies to all the other cases below

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -778,3 +802,106 @@ def check(self):
...
+        if (spec.satisfies("+openmp") and spec.satisfies("+elpa")) or spec.satisfies(
+            "^elpa+openmp"
+        ):
+            args += ["-DCP2K_ENABLE_ELPA_OPENMP_SUPPORT=ON"]
+
+        if "spla" in spec and (spec.satisfies("+cuda") or spec.satisfies("+rocm")):
+            args += ["-DCP2K_USE_SPLA_GEMM_OFFLOADING=ON"]
+
+        if "fftw" in spec:
+            args += ["-DCP2K_USE_FFTW=ON"]
```
dev-zero:
```suggestion
        args += ["-DCP2K_USE_FFTW=ON"]
```

I made `fftw-api` a hard requirement for CP2K on Spack. And I guess `USE_FFTW` should be set even if the provider is `libsci` or `mkl`?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -778,3 +802,106 @@ def check(self):
...
+
+        with when("smm=libxsmm"):
+            args += ["-DCP2K_USE_LIBXSMM=ON"]
```
dev-zero: I have to check whether the following works:
```
self.define_from_variant("CP2K_USE_LIBXSMM", "smm=libxsmm"),
```

but in any case we might want to set it to OFF explicitly just to anticipate the CP2K CMake to switch the default at some point.
Also, in DBCSR CMake we have modeled that flag the same way as in in Spack: `-DUSE_SMM=...` to anticipate new/other libsmm providers.

var/spack/repos/builtin/packages/cp2k/package.py
dev-zero: Do we also have to ensure consistent settings for libxsmm?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -778,3 +826,94 @@ def check(self):
...
+
+        if spec.satisfies("+cuda"):
+            args += ["-DCP2K_USE_ACCEL=cuda"]
+
+        if spec.satisfies("+rocm"):
+            args += ["-DCP2K_USE_ACCEL=hip"]
+
+        args += [
+            self.define_from_variant("CP2K_USE_ELPA", "elpa"),
+            self.define_from_variant("CP2K_USE_LIBINT", "libint"),
```
RMeli:
```suggestion
            self.define_from_variant("CP2K_USE_LIBINT2", "libint"),
```

var/spack/repos/builtin/packages/cp2k/package.py
RMeli:
```suggestion
        args += ["-DCP2K_USE_FFTW3=ON"]
```
I think it should be `FFTW3`

mtaillefumier:
A ping to review this PR ;)

alalazo:
@dev-zero

alalazo:
@albestro Would you have time to review this PR in the next few days?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +824,104 @@ def check(self):
...
+        if "+cuda" in spec and len(spec.variants["cuda_arch"].value) > 1:
+            raise InstallError("CP2K supports only one cuda_arch at a time")
+
+        if "+rocm" in spec and len(spec.variants["amdgpu_target"].value) > 1:
+            raise InstallError("CP2K supports only one amdgpu_arch at a time")
+
+        args = []
+
+        if spec.satisfies("+cuda") or spec.satisfies("+rocm"):
+            if spec.satisfies("cuda_target"):
```
msimberg:
```suggestion
            if not spec.satisfies("cuda_arch=none"):
```
?

_But_, it might be clearer to also have separate branches for `spec.satisfies("+cuda")` and `spec.satisfies("+rocm")` (possibly even merging them with the `InstallError` checks above and the `CP2K_USE_ACCEL` branches below?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +824,104 @@ def check(self):
...
+                args += [self.define("CP2K_WITH_GPU", gpu_map[spec.variants["cuda_arch"].value])]
+
+            if spec.satisfies("amdgpu_target"):
+                args += [
+                    self.define("CP2K_WITH_GPU", gpu_map[spec.variants["amdgpu_target"].value])
```
msimberg: I think `value` is an array with one element at this point:
```suggestion
                args += [self.define("CP2K_WITH_GPU", gpu_map[spec.variants["cuda_arch"].value[0])]

            if spec.satisfies("amdgpu_target"):
                args += [
                    self.define("CP2K_WITH_GPU", gpu_map[spec.variants["amdgpu_target"].value[0])
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +824,104 @@ def check(self):
...
+
+        if spec.satisfies("smm=libxsmm"):
+            args += ["-DCP2K_USE_LIBXSMM=ON"]
+        else:
+            args += ["-DCP2K_USE_LIBXSMM=OFF"]
+
+        lapack = spec["lapack"]
+        blas = spec["blas"]
+
+        if spec["blas"].name in ["intel-mkl", "intel-parallel-studio"]:
```
msimberg:
```suggestion
        if blas.name in ["intel-mkl", "intel-parallel-studio"]:
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +824,104 @@ def check(self):
...
+            else:
+                args += ["-DCP2K_SCALAPACK_VENDOR=MKL"]
+        else:
+            args.extend(
+                [
+                    self.define("CP2K_LAPACK_FOUND", "true"),
+                    self.define("CP2K_LAPACK_LINK_LIBRARIES", lapack.libs.joined(";")),
+                    self.define("CP2K_BLAS_FOUND", "true"),
+                    self.define("CP2K_BLAS_LINK_LIBRARIES", blas.libs.joined(";")),
+                    self.define("CP2K_SCALAPACK_FOUND", "true"),
```
msimberg:
```suggestion
                    self.define("CP2K_LAPACK_FOUND", True),
                    self.define("CP2K_LAPACK_LINK_LIBRARIES", lapack.libs.joined(";")),
                    self.define("CP2K_BLAS_FOUND", True),
                    self.define("CP2K_BLAS_LINK_LIBRARIES", blas.libs.joined(";")),
                    self.define("CP2K_SCALAPACK_FOUND", True),
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +824,104 @@ def check(self):
...
+            self.define_from_variant("CP2K_USE_VORI", "libvori"),
+            self.define_from_variant("CP2K_USE_SPLA", "spla"),
+            self.define_from_variant("CP2K_USE_QUIP", "quip"),
+        ]
+
+        # we force the use elpa openmp threading support. might need to be revisited though
+        if (spec.satisfies("+openmp") and spec.satisfies("+elpa")) or spec.satisfies(
+            "^elpa+openmp"
+        ):
+            args += ["-DCP2K_ENABLE_ELPA_OPENMP_SUPPORT=ON"]
```
msimberg: This and the the other args below could be rewritten as
```suggestion
        args += [self.define("CP2K_ENABLE_ELPA_OPENMP_SUPPORT",
                 (spec.satisfies("+openmp") and spec.satisfies("+elpa")) or
                 spec.satisfies("^elpa+openmp"))]
```
but this is a matter of taste.

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -231,14 +266,27 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+    cuda_msg = "cp2k only supports cuda_arch {0}".format(supported_cuda_arch_list)
+
+    conflicts("+cuda", when="cuda_arch=none")
+    # rocm already returns an error if amdgpu_target is not defined
 
-    conflicts("+cuda", when="cuda_arch=none", msg=cuda_msg)
+    with when("+cuda"):
+        for arch in CudaPackage.cuda_arch_values:
+            if arch not in supported_cuda_arch_list:
+                conflicts("+cuda", when="cuda_arch={0}".format(arch), msg=cuda_msg)
```
msimberg, line 295: Do you want to add similar conflicts for the unsupported rocm architectures?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -575,12 +623,25 @@ def edit(self, spec, prefix):
                     cppflags += ["-D__PW_CUDA"]
                     libs += ["-lcufft", "-lcublas"]
 
-            cuda_arch = spec.variants["cuda_arch"].value[0]
-            if cuda_arch:
-                gpuver = {"35": "K40", "37": "K80", "60": "P100", "70": "V100"}[cuda_arch]
+            gpuver = gpu_map[spec.variants["cuda_arch"].value]
+            if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
```
msimberg:
```suggestion
            cuda_arch = spec.variants["cuda_arch"].value[0]
            gpuver = gpu_map[cuda_arch]
            if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
```
?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -575,12 +623,25 @@ def edit(self, spec, prefix):
...
+            ]
 
-                if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
-                    gpuver = "K20X"
+            cppflags += ["-D__OFFLOAD_HIP"]
+            acc_compiler_var = "hipcc"
+            acc_flags_var = "NVFLAGS"
+            cppflags += ["-D__ACC"]
+            cppflags += ["-D__DBCSR_ACC"]
+            gpuver = gpu_map[spec.variants["amdgpu_target"].value]
```
msimberg:
```suggestion
            gpuver = gpu_map[spec.variants["amdgpu_target"].value[0]]
```

msimberg: Also, does `gpuver` need to be used somewhere in the `+rocm` case? I can't see it being used but I may have missed it...

msimberg (changes requested):

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -231,14 +267,46 @@ class Cp2k(MakefilePackage, CudaPackage):
...
-    cuda_msg = "cp2k only supports cuda_arch {0}".format(dbcsr_cuda_archs)
 
-    for arch in CudaPackage.cuda_arch_values:
-        if arch not in dbcsr_cuda_archs:
-            conflicts("+cuda", when="cuda_arch={0}".format(arch), msg=cuda_msg)
+    supported_cuda_arch_list = ("35", "37", "60", "70", "80")
+    supported_rocm_arch_list = (
+        "60",
+        "70",
+        "80",
```
msimberg: I guess these don't belong in the rocm list?

msimberg (changes requested):
One more nit from me, looks good to me otherwise.

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +73,31 @@ class Cp2k(MakefilePackage, CudaPackage):
...
     )
     variant("spglib", default=False, description="Enable support for spglib")
+    variant(
+        "spla",
+        default=False,
+        description="Use SPLA off-loading functionality. Only"
+        " relevant when CUDA or ROCM are enabled",
+    )
+    variant("pytorch", default=False, description="Enable libtorch support")
+    variant("quip", default=False, description=("Enable quip support"))
```
albestro:
```suggestion
    variant("quip", default=False, description="Enable quip support")
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +73,31 @@ class Cp2k(MakefilePackage, CudaPackage):
...
         default=False,
-        description=("Enable support for Voronoi integration" " and BQB compression"),
+        description="Enable support for Voronoi integration and BQB compression",
     )
     variant("spglib", default=False, description="Enable support for spglib")
+    variant(
+        "spla",
+        default=False,
+        description="Use SPLA off-loading functionality. Only"
+        " relevant when CUDA or ROCM are enabled",
```
albestro:
```suggestion
        description="Use SPLA off-loading functionality. Only relevant when CUDA or ROCM are enabled",
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +73,31 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+    )
+    variant("pytorch", default=False, description="Enable libtorch support")
+    variant("quip", default=False, description=("Enable quip support"))
+
+    variant(
+        "enable_regtests",
+        default=False,
+        description="Configure cp2k to run the regtests afterwards."
+        " It build cp2k normally but put the executables in exe/cmake-build-* instead of the"
+        " conventional location. This option is only relevant when regtests need to be run.",
```
albestro, line 99: Same "style fix" about split string here?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -231,14 +267,46 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+        "70": "V100",
+        "80": "A100",
+        "gfx906": "Mi50",
+        "gfx908": "Mi100",
+        "gfx90a": "Mi250",
+        "gfx90a:xnack-": "Mi250",
+        "gfx90a:xnack+": "Mi250",
+    }
+    cuda_msg = "cp2k only supports cuda_arch {0}".format(supported_cuda_arch_list)
+    rocm_msg = "cp2k only supports andgpu_target {0}".format(supported_rocm_arch_list)
```
albestro:
```suggestion
    rocm_msg = "cp2k only supports amdgpu_target {0}".format(supported_rocm_arch_list)
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -5,23 +5,33 @@
 import copy
 import os
 import os.path
+import sys
 
+import spack.platforms
 import spack.util.environment
+import spack.util.executable
+from spack.build_environment import dso_suffix
```
albestro, line 13: What are these needed for?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -576,11 +644,25 @@ def edit(self, spec, prefix):
...
             cuda_arch = spec.variants["cuda_arch"].value[0]
-            if cuda_arch:
-                gpuver = {"35": "K40", "37": "K80", "60": "P100", "70": "V100"}[cuda_arch]
+            gpuver = gpu_map[cuda_arch]
+            if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
+                gpuver = "K20X"
 
-                if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
-                    gpuver = "K20X"
+        if spec.satisfies(+rocm) and spec.satisfies("@2022:"):
```
albestro: FIY, I think this is equivalent to something like

```suggestion
        if "@2022: +rocm" in spec:
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +845,114 @@ def check(self):
...
+            if (len(spec.variants["cuda_arch"].value) > 1) or spec.satisfies("cuda_arch=none"):
+                raise InstallError("CP2K supports only one cuda_arch at a time.")
+            else:
+                gpu_ver = gpu_map[spec.variants["cuda_arch"].value[0]]
+                args += ["-DCP2K_USE_ACCEL=CUDA"]
+                args += [self.define("CP2K_WITH_GPU", gpu_ver)]
+
+        if "+rocm" in spec:
+            if len(spec.variants["amdgpu_target"].value) > 1:
+                raise InstallError("CP2K supports only one amdgpu_target at a time")
```
albestro:
```suggestion
                raise InstallError("CP2K supports only one amdgpu_target at a time.")
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +845,114 @@ def check(self):
...
+            self.define_from_variant("CP2K_USE_SPLA", "spla"),
+            self.define_from_variant("CP2K_USE_QUIP", "quip"),
+        ]
+
+        # we force the use elpa openmp threading support. might need to be revisited though
+        args += [
+            self.define(
+                "CP2K_ENABLE_ELPA_OPENMP_SUPPORT",
+                (spec.satisfies("+openmp") and spec.satisfies("+elpa"))
+                or spec.satisfies("^elpa+openmp"),
```
albestro: This is just an example where readability might improve using the `in` operator
```suggestion
                ("+elpa +openmp" in spec) or ("^elpa +openmp" in spec),
```

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -147,13 +173,13 @@ class Cp2k(MakefilePackage, CudaPackage):
             )
 
     with when("+libxc"):
-        depends_on("pkgconfig", type="build", when="@7.0:")
-        depends_on("libxc@2.2.2:3", when="@:5", type="build")
-        depends_on("libxc@4.0.3:4", when="@6.0:6.9", type="build")
+        depends_on("pkgconfig", when="@7.0:")
```
albestro, line 176: not sure about the others (I see that a discussion on this topic has already occurred with @dev-zero), but probably `pkgconfig` is actually needed just at build time...isn't it?

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -576,11 +644,25 @@ def edit(self, spec, prefix):
...
 
-                if cuda_arch == "35" and spec.satisfies("+cuda_arch_35_k20x"):
-                    gpuver = "K20X"
+        if spec.satisfies(+rocm) and spec.satisfies("@2022:"):
+            libs += [
+                "-L{}".format(spec["rocm"].libs.directories[0]),
+                "-L{}/stubs".format(spec["rocm"].libs.directories[0]),
+                "-lhipblas",
+                "-lhipfft",
+                "-lstdc++",
```
albestro, line 648: I don't know the reason why you need to explicitly instruct the rocm dependency like this but FIY

```bash
>>> s = Spec("hwloc") # just a random package with a dependency
>>> s.concretize()
>>> dep = s["ncurses"]
>>> dep.libs.search_flags
'-L/Users/ialberto/spack/opt/spack/darwin-ventura-m2/clang-16.0.3/ncurses-6.4-pylthzrstz72hhppg4oiyr3glvziyyur/lib'
>>> dep.libs.link_flags
'-lncurses -ltinfo -lncursesw -ltinfow'
>>> dep.libs.ld_flags
'-L/Users/ialberto/spack/opt/spack/darwin-ventura-m2/clang-16.0.3/ncurses-6.4-pylthzrstz72hhppg4oiyr3glvziyyur/lib -lncurses -ltinfo -lncursesw -ltinfow'
```

albestro (commented):
Super work @mtaillefumier!

I don't have knowledge about CP2K, so from my side just minor comments and suggestions on style/available helpers.

About `"+variant in spec"` vs `spec.satisfies("+variant")` styles, I don't think there is any preferred method so no change is required, but in general I would try to stick to one (and personally I prefer the former one, more coincise and readable most of the times in my opinion).

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -763,3 +836,113 @@ def check(self):
...
+            "37": "K80",
+            "60": "P100",
+            "70": "V100",
+            "80": "A100",
+            "gfx906": "Mi50",
+            "gfx908": "Mi100",
+            "gfx90a": "Mi250",
+            "gfx90a:xnack-": "Mi250",
+            "gfx90a:xnack+": "Mi250",
+        }
```
RMeli, line 860: The same `gpu_map` is used above, is it worth defining it in one place (globally)?

msimberg (approved):

RMeli (approved):

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -198,27 +225,36 @@ class Cp2k(MakefilePackage, CudaPackage):
...
         # libvori support was introduced in 8+
         conflicts("@:7")
 
     # the bundled libcusmm uses numpy in the parameter prediction (v7+)
     # which is written using Python 3
-    depends_on("py-numpy", when="@7:+cuda", type="build")
-    depends_on("python@3.6:", when="@7:+cuda", type="build")
+    depends_on("py-numpy", when="@7:+cuda")
+    depends_on("python@3.6:", when="@7:+cuda")
+    depends_on("py-fypp")
```
lorisercole, line 244: I just noticed that the `py-fypp` dependency causes `cp2k` to depend on `python`. Is that really necessary?
Or is it just a build dependency? @mtaillefumier

RMeli: `py-fypp` should be a build dependency AFAIK.

var/spack/repos/builtin/packages/cp2k/package.py
```diff
@@ -63,15 +73,31 @@ class Cp2k(MakefilePackage, CudaPackage):
...
+        description="Enable support for Voronoi integration and BQB compression",
     )
     variant("spglib", default=False, description="Enable support for spglib")
+    variant(
+        "spla",
+        default=False,
+        description="Use SPLA off-loading functionality. Only relevant when CUDA or ROCM"
+        " are enabled",
+    )
+    variant("pytorch", default=False, description="Enable libtorch support")
```
giordano, line 91: When `pytorch` is enabled, there is no extra dependency on `py-torch`.  How is this supposed to work?  Was it an oversight?

RMeli: @giordano this looks like an oversight indeed.

PR #35720, opened 2023-02-27
fix spack load --list and spack unload on windows

danlipsa:
fixes for:
spack load --list (this printed 0 packages even if packages were loaded)
spack unload 'package' (this said that the package is not loaded even if it was).


danlipsa:
@johnwparent Please review

danlipsa:
@scheibelp Please review

scheibelp:
A couple general requests:

* I assume there are associated (but currently disabled on Windows) tests for `spack load` functionality: ideally they would be reenabled as part of this PR. Does that seem reasonable?
* Can you update the PR description with the what and why?

danlipsa:
> * I assume there are associated (but currently disabled on Windows) tests for `spack load` functionality: ideally they would be reenabled as part of this PR. Does that seem reasonable?
@scheibelp That would be great! Can you point me to some documentation on how to run the tests or give me some brief pointers. Thanks!

> * Can you update the PR description with the what and why?
How does the updated description look?

scheibelp:
>> I assume there are associated (but currently disabled on Windows) tests for spack load functionality: ideally they would be reenabled as part of this PR. Does that seem reasonable?

> @scheibelp That would be great! Can you point me to some documentation on how to run the tests or give me some brief pointers. Thanks!

[test/cmd/load.py](https://github.com/spack/spack/blob/develop/lib/spack/spack/test/cmd/load.py) is the one I see (there is a module level disabling of tests for Windows).

>> Can you update the PR description with the what and why?

> How does the updated description look?

Good: thanks!

danlipsa:
@scheibelp I enabled/updated the tests in load.py and a two tests in find.py and now they run successfully on windows. Please review.

lib/spack/spack/cmd/unit_test.py
```diff
@@ -217,6 +217,9 @@ def unit_test(parser, args, unknown_args):
             if pytest is None:
                 spack.bootstrap.ensure_environment_dependencies()
                 import pytest
+    else:
```
johnwparent: @danlipsa If you look at line 213 you can see the reason we skip this is because at the time Clingo was not supported via bootstrap on Windows. This is no longer the case, so we can actually just be rid of this entire if/else stanza.

danlipsa: Actually, I could not bootstrap clingo. I installed it using pip.

danlipsa:
@johnwparent I rebased this PR over devel and the load.py tests still pass on my machine.

johnwparent:
Thanks for the upkeep Dan! Changes in general still look good to me, so pending CI passing, this is ready for  another round of review.

danlipsa:
@scheibelp This branch is also ready.

danlipsa:
@scheibelp ping. This passes all tests

bin/spack.bat
```diff
@@ -210,13 +210,22 @@ if NOT "%_sp_args%"=="%_sp_args:--help=%" (
 goto :end_switch
 
 :case_load
+if not defined _sp_args (
+   if not defined _sp_flags (
+      python "%spack%" load --help
+      exit /B 0
+   )
```
scheibelp: `_sp_flags` are the flags passed to the `spack` binary, so for example `spack -d load` would have `_sp_flags` but no `_sp_args`: I'm inclined to say this should print help whenever `_sp_args` is empty.

Also I have a question: earlier this script does:

```
set "_sp_flags="
set "_sp_args="
```

so would it ever be the case that `_sp_args` is undefined (I thought setting to an empty value was different than un-defining)?

lib/spack/spack/cmd/unit_test.py
scheibelp: @danlipsa  can you coordinate with @johnwparent about getting this working for you if this is still a problem?

lib/spack/spack/test/cmd/load.py
```diff
@@ -28,15 +24,26 @@ def test_manpath_trailing_colon(
...
+    if sys.platform == "win32":
+        bat_out = load("--bat", "--only", "package", "mpileaks")
+        lines = bat_out.split("\n")
+        assert any(re.match(r'set "MANPATH=.*:"', ln) for ln in lines)
+
+        os.environ["MANPATH"] = "/tmp/man;"
+
+        bat_out = load("--bat", "--only", "package", "mpileaks")
+        lines = bat_out.split("\n")
+        assert any(re.match(r'set "MANPATH=.*;/tmp/man;:', ln) for ln in lines)
```
scheibelp: A few things:

* We don't need to be on `win32` to test the output of `spack load` (when the shell arg is provided). So the "if" could be removed here
* AFAIK `PATH` ending in `:` only has significance on Linux, so we only need to assert that there
* That being said, the fact that this is asserting that the line is `r'set "MANPATH=.*;/tmp/man;:'` is confusing to me: what does the `:` at the end of that string mean?

lib/spack/spack/test/cmd/load.py
```diff
@@ -51,21 +58,33 @@ def test_load(install_mockery, mock_fetch, mock_archive, mock_packages):
...
+        csh_out_test = "setenv CMAKE_PREFIX_PATH %s" % mpileaks_spec.prefix
+        assert sh_out_test in sh_out
+        assert csh_out_test in csh_out
+
+        # Test hashes recorded properly
+        hash_test_replacements = (uenv.spack_loaded_hashes_var, mpileaks_spec.dag_hash())
+        sh_hash_test = "export %s=%s" % hash_test_replacements
+        csh_hash_test = "setenv %s %s" % hash_test_replacements
+        assert sh_hash_test in sh_out
+        assert csh_hash_test in csh_out
```
scheibelp: Like `test_manpath_trailing_colon `, we don't have to only execute `--sh` on windows

Unlike the `test_manpath_trailing_colon`, I think this could be consolidated with parameterization:

```
@pytest.parameter(shell, [
  (--bat, 'set "%s=%s", X),
  (--sh, "export CMAKE_PREFIX_PATH=%s", "setenv %s %s"), 
  (--csh, ...)
])
``` 

assuming there is a sensible setting for X (I assume on Windows we also set `uenv.spack_loaded_hashes_var`).

Assuming that makes sense, can you look into whether a similar consolidation can be applied to the other tests (I don't think `test_manpath_trailing_colon` could be updated easily in this way, so that might be the case elsewhere).

scheibelp (changes requested):

bin/spack.bat
danlipsa: > _sp_flags are the flags passed to the spack binary, so for example spack -d load would have _sp_flags but no _sp_args: I'm inclined to say this should print help whenever _sp_args is empty.

Sounds good.

danlipsa: > would it ever be the case that _sp_args is undefined (I thought setting to an empty value was different than un-defining)?

Setting an empty value is the same as undefining. See bullet 6 of Remarks section:
https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/set_1

lib/spack/spack/test/cmd/load.py
danlipsa: > We don't need to be on win32 to test the output of spack load (when the shell arg is provided). So the "if" could be removed here

Good point.

danlipsa: > That being said, the fact that this is asserting that the line is r'set "MANPATH=.*;/tmp/man;:' is confusing to me: what does the : at the end of that string mean?

That was a result of terminating the manpath with colon. I change that to use os.pathsep

danlipsa:
@scheibelp Please review. Thanks!

danlipsa:
@johnwparent @scheibelp Please review. I rebased this PR on develop - all the tests in load.py pass on windows now.

bin/spack.bat
```diff
@@ -188,25 +188,28 @@ if NOT "%_sp_args%"=="%_sp_args:--help=%" (
...
-    )
+if NOT defined _sp_args (
+   python "%spack%" "%_sp_subcommand%" --help
+   exit /B 0
+)
+
+:: If args contain --bat, or -h/--help: just execute.
+if NOT "%_sp_args%"=="%_sp_args:--help=%" (
+    goto :default_case
+) else if NOT "%_sp_args%"=="%_sp_args: -h=%" (
```
johnwparent: Can you remind me why the space here before the `-h` is required?

bin/spack.bat
```diff
@@ -188,25 +188,28 @@ if NOT "%_sp_args%"=="%_sp_args:--help=%" (
...
-if defined _sp_args (
-    if NOT "%_sp_args%"=="%_sp_args:--help=%" (
-        goto :default_case
-    ) else if NOT "%_sp_args%"=="%_sp_args:-h=%" (
-        goto :default_case
-    ) else if NOT "%_sp_args%"=="%_sp_args:--bat=%" (
-        goto :default_case
-    )
+if NOT defined _sp_args (
+   python "%spack%" "%_sp_subcommand%" --help
```
johnwparent: This is not the behavior I observe on linux when `spack load` is invoked with no arguments. There, the command seems to exit silently without invoking the help. Seems like these should be the same (and IMO this approach is better)
@scheibelp thoughts?

bin/spack.bat
```diff
@@ -188,25 +188,28 @@ if NOT "%_sp_args%"=="%_sp_args:--help=%" (
...
+) else if NOT "%_sp_args%"=="%_sp_args:--bat=%" (
+    goto :default_case
+) else if NOT "%_sp_args%"=="%_sp_args:--list=%" (
+    goto :default_case
 )
 
 for /f "tokens=* USEBACKQ" %%I in (
-    `python "%spack%" %_sp_flags% %_sp_subcommand% --bat %_sp_args%`) do %%I
+    `python "%spack%" %_sp_flags% %_sp_subcommand% --bat %_sp_args%`
+    ) do %%I
```
johnwparent, line 208: Is there a reason for the formatting change? Don't have an issue with it, just trying to understand why it's necessary in case I have to change this line later.

bin/spack.bat
danlipsa: That is a typo, it should not be there. Nice catch!

bin/spack.bat
danlipsa: I don't remember why I made it. Looking at it, I like it better because it highlights the back ticks and show what it does: generates a list of commands that will be executed by the windows code. But I don't have a strong preference one way or the other. I can revert it if you prefer.

bin/spack.bat
danlipsa: @scheibelp Can we merge this as is, or should we make it consistent with linux?

danlipsa: @johnwparent I removed the help print out, so now the behavior is the same as in Linux. If we want to change that we can do it in a separate commit.

danlipsa:
@johnwparent Is this ready?

bin/spack.bat
```diff
@@ -188,25 +188,27 @@ if NOT "%_sp_args%"=="%_sp_args:--help=%" (
...
 
 goto :end_switch
 
-:case_unload
-goto :case_load
```
scheibelp, line 208: (question) how will this affect the behavior of `spack unload --help`?

lib/spack/spack/test/cmd/load.py
```diff
@@ -17,61 +18,94 @@
...
-
 
+@pytest.mark.parametrize(
+    "shell,set_command,commandsep",
+    (
+        [("--bat", 'set "%s=%s"', "\n")]
+        if sys.platform == "win32"
+        else [("--sh", "export %s=%s", ";\n")]
+    ),
+)
```
scheibelp: IMO it's a bit confusing to have a parameter list that always evaluates to 1 list item depending on platform: can you move these variables to the beginning of the function, vs. parameters passed as input?

(I think all other tests modified here have a similar issue)

scheibelp (changes requested):
I have one question and one request (although that request applies to several changes)

bin/spack.bat
danlipsa: For "unload" we go directly to :case_load so there is no need for this additional indirection. spack unload --help ends up at :default_case which prints the correct message.

lib/spack/spack/test/cmd/load.py
danlipsa: I agree that for lists that have only one item, they are better as variables instead of parameters from parametrize. There are a few functions that have two items in the list on unix, one for sh and one for csh. Should leave those as they are?

bin/spack.bat
scheibelp: OK, I think I see that above around line 128 - is that what you are referring to?

lib/spack/spack/test/cmd/load.py
scheibelp: > There are a few functions that have two items in the list on unix, one for sh and one for csh. Should leave those as they are?

That sounds good to me - thanks!

lib/spack/spack/test/cmd/load.py
```diff
@@ -17,61 +18,85 @@
...
 
-    paths_sh = extract_cmake_prefix_path(sh_out, prefix="export CMAKE_PREFIX_PATH=")
-    paths_csh = extract_cmake_prefix_path(csh_out, prefix="setenv CMAKE_PREFIX_PATH ")
+    paths_shell = extract_cmake_prefix_path(
+        shell_out, prefix=set_command_prefix % "CMAKE_PREFIX_PATH"
+    )
 
     # Shouldn't be a difference between loading csh / sh, so check they're the same.
-    assert paths_sh == paths_csh
+    # assert paths_sh == paths_csh
```
scheibelp: I think we'll want to preserve this check. I acknowledge that makes this awkward (I'd probably start by splitting into two tests and then looking for ways to consolidate); another option is to directly extract the path from the spec, and then assert something like:

```
assert paths = cmake_prefix_paths_from_spec(spec)
```

lib/spack/spack/test/cmd/load.py
```diff
@@ -17,61 +18,85 @@
...
-pytestmark = pytest.mark.not_on_windows("does not run on windows")
-
 
 def test_manpath_trailing_colon(
     install_mockery, mock_fetch, mock_archive, mock_packages, working_env
 ):
+    (shell, set_command, commandsep) = (
+        ("--bat", 'set "%s=%s"', "\n")
+        if sys.platform == "win32"
+        else ("--sh", "export %s=%s", ";\n")
```
scheibelp: I see why it works to call `commandsep=;\n` here, but I think it would be clearer to keep it as `;` and `.strip()` the resulting lines. Overall I think it would be ideal to put that in a function like `break_shell_output_into_commands`, which might be useful for simplifying `extract_cmake_prefix_path`.

lib/spack/spack/test/cmd/load.py
```diff
@@ -17,61 +18,85 @@
...
+        re.match(set_command % ("MANPATH", ".*" + os.pathsep + "/tmp/man" + os.pathsep), ln)
+        for ln in lines
+    )
 
 
-def test_load_recursive(install_mockery, mock_fetch, mock_archive, mock_packages, working_env):
+@pytest.mark.parametrize(
+    "shell,set_command_prefix,set_command_postfix,commandsep",
+    (
+        [("--bat", 'set "%s=', '%s"', "\n")]
```
scheibelp: This decomposition is hard to follow IMO, in particular the splitting into `set_command_prefix` and `set_command_postfix` (this makes it slightly more confusing than the other parameterizations): is it possible to rewrite the test in terms of a single `set_command`?

scheibelp (changes requested):
In my original review I missed that an assert was commented out, and I think it would be useful to preserve. I also have a couple extra refactoring requests.

danlipsa:
@scheibelp How does this look? Thanks!

lib/spack/spack/test/cmd/load.py
```diff
@@ -17,101 +18,128 @@
...
+
+        # Map a prefix found in CMAKE_PREFIX_PATH back to a package name in mpileaks' DAG.
+        prefix_to_pkg = lambda prefix: next(
+            s.name for s in mpileaks_spec.traverse() if s.prefix == prefix
+        )
+
+        paths_shell = extract_value(shell_out, "CMAKE_PREFIX_PATH")
+
+        # Shouldn't be a difference between loading csh / sh, so check they're the same.
+        # assert paths_sh == paths_csh
```
scheibelp: This comment can be removed now (you now handle it below).

lib/spack/spack/test/cmd/load.py
```diff
@@ -17,101 +18,128 @@
...
-
-def test_load_includes_run_env(install_mockery, mock_fetch, mock_archive, mock_packages):
+def test_load_recursive(install_mockery, mock_fetch, mock_archive, mock_packages, working_env):
+    def test_load_shell(shell, set_command):
+        """Test that `spack load` applies prefix inspections of its required runtime deps in
+        topo-order"""
+        install("mpileaks")
+        mpileaks_spec = spack.spec.Spec("mpileaks").concretized()
+
+        # Ensure our reference variable is cleed.
```
scheibelp: (minor) this was a typo from before, but can you change "cleed" -> "clean"

scheibelp (changes requested):
Minor requests: you have a commented out piece of code that you've since addressed (and also a typo from before this PR).

Other than that this looks good!

scheibelp (approved):

PR #35724, opened 2023-02-27
Package heffte: Enable smoke test to find MPI launcher

G-Ragghianti:
This fixes the smoke test by adding the path to the MPI dependency to the ~~CMAKE_PREFIX_PATH~~ MPI_HOME.


var/spack/repos/builtin/packages/heffte/package.py
```diff
@@ -165,6 +165,9 @@ def test(self):
                 "-Drocfft_DIR=" + join_path(self.spec["rocfft"].prefix, "lib", "cmake", "rocfft")
             )
 
+        # Provide the path to the MPI dependency so that the test can find an MPI launcher
+        options.append("-DCMAKE_PREFIX_PATH=" + self.spec["mpi"].prefix)
```
mkstoyanov: Are you sure will work? CMake recommend you use `MPIEXEC_EXECUTABLE` and `MPI_CXX_COMPILER` for locating MPI.

https://cmake.org/cmake/help/latest/module/FindMPI.html

G-Ragghianti: Yes, this allows the cmake process that is building the tests to find the MPI install and to set the needed variables like MPIEXEC_EXECUTABLE. It is true that the user could in theory provide MPIEXEC_EXECUTABLE to cmake, but wouldn't make sense in this case because the value of MPIEXEC_EXECUTABLE is already determined and put into the cmake files that are installed with the MPI that was set up as a dependency of heffte. We set CMAKE_PREFIX_PATH so that the cmake build of the tests can find the previously installed MPI and get the variables that it needs from there.

G-Ragghianti: Normally, spack would do this automatically for packages that are set up as a runtime dependency, but the smoke tests are not currently treated correctly within spack (by getting the runtime dependencies automatically provided).

G-Ragghianti: As an example, here is another packages doing something very similar (but for a larger number of dependencies):
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/slate/package.py

mkstoyanov: Ugh, more spack quirks with workarounds that will only break again the future.

mkstoyanov (dismissed):

var/spack/repos/builtin/packages/heffte/package.py
G-Ragghianti: Yep, my understanding from the spack devs is that there will eventually be proper injection of dependencies for the smoke test environment, but for now it is all workarounds.  I'll be happy to remove all this once the proper functionality is in place!

tldahlgren: I have this working in #35750 with the following change:

```suggestion
        options.append("-DMPI_HOME=" + self.spec["mpi"].prefix)
```

tldahlgren: Well,  technically working when I use these changes with #34236 .

tldahlgren (changes requested):
From what I can tell, you only need `MPI_HOME`.

var/spack/repos/builtin/packages/heffte/package.py
tldahlgren: > Are you sure will work? CMake recommend you use `MPIEXEC_EXECUTABLE` and `MPI_CXX_COMPILER` for locating MPI.
> 
> https://cmake.org/cmake/help/latest/module/FindMPI.html

https://cmake.org/cmake/help/latest/module/FindMPI.html#variables-for-locating-mpi lists the second variable you can use as `MPI_HOME`, which I found works for me.

mkstoyanov: @tldahlgren is right, `MPI_HOME` is the correct thing to do here.

@G-Ragghianti can you update the PR, please?

mkstoyanov (approved):
The `MPI_HOME` change is also included in #35750 

To avoid a conflict, maybe close this one or reduce it to just adding the new maintainer.

tldahlgren:
Let's see if this resolves the odd CI errors for two stacks.  If not, the branch may need to be updated.

@spackbot run pipeline

tldahlgren:
> The `MPI_HOME` change is also included in #35750
> 
> To avoid a conflict, maybe close this one or reduce it to just adding the new maintainer.

This PR is further along so I'm fine with addressing the conflict in #35750 when the time comes.

tldahlgren (approved):
Thanks for the fix.

G-Ragghianti:
@spackbot run pipeline

G-Ragghianti:
Can this be merged?

G-Ragghianti:
Can this be merged?

PR #35728, opened 2023-02-27
Fix superlu-dist package for cray fortran

lukebroskop:
The superlu-dist code developers the build such that the patch is no longer needer for `@7.2.0:`  (the patch will actually fail)


quellyn:
Just to confirm what Luke said: I tried 
```
spack patch superlu-dist@7.2.0%cce
```
and 

```
spack patch superlu-dist@8.1.2%cc
```

On RZvernal (Cray cce 15.0.1) and both rejected the patch.

alalazo (approved):

PR #35731, opened 2023-02-28
BinaryCache: Add SHA256 and ETag to spackfile

kwryankrattiger:
This PR adds the extra SHA256 attribute to uploaded spack file binaries and gets the ETag from the binary after being moved to the binary cache. Currently only `file://` and `s3://` support getting the ETag of binaries in the build cache.

This PR is a first step to adding a `spack buildcache validation` method.

FYI: @tgamblin @scottwittenburg @haampie @alalazo

### DIFF
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -821,14 +821,7 @@ def _url_etag_group(url):
     if isinstance(url, str):
         parsed_url = urllib.parse.urlparse(url)
 
-    if "http" in parsed_url.scheme:
-        # HTTP mirrors may have host specific etags
-        etag_group = parsed_url.host
-    else:
-        # s3:// and file:// use a consistent ETag method
-        etag_group = parsed_url.scheme
-
-    return etag_group
+    return parsed_url.geturl()
 
 
 def select_signing_key(key=None):
@@ -1270,7 +1263,7 @@ def _build_tarball(
 
     # push tarball to remote mirror
     # do this first in order to attempt to get an etag for the uploaded resource
-    web_util.push_to_url(
+    spackfile_etag = web_util.push_to_url(
         spackfile_path,
         remote_spackfile_path,
         keep_original=False,
@@ -1279,11 +1272,6 @@ def _build_tarball(
     )
     # Try to get the ETag of the uploaded tarball. This must happen after uploading the
     # tarball since it is not possible to pre-compute the resource etag
-    try:
-        spackfile_etag = web_util.read_etag(remote_spackfile_path)
-    except web_util.FetchError:
-        spackfile_etag = None
-
     if spackfile_etag:
         etag_group = _url_etag_group(remote_spackfile_path)
 
@@ -1299,7 +1287,6 @@ def _build_tarball(
     bchecksum["hash_algorithm"] = "sha256"
     bchecksum["hash"] = checksum
     # Store the etag if it was found
-    # Partition by netloc to avoid mismatch if copied
     if spackfile_etag:
         bchecksum["etag"] = {etag_group: spackfile_etag}
     spec_dict["binary_cache_checksum"] = bchecksum
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -821,13 +821,7 @@ def _url_etag_group(url):
     if isinstance(url, str):
         parsed_url = urllib.parse.urlparse(url)
 
-    if parsed_url.scheme == "s3":
-        return "s3"
-    if parsed_url.scheme == "file":
-        return "file"
-    else:
-        return parsed_url.host()
-
+    return "{0}://{1}".format(parsed_url.sheme, parsed_url.netloc)
 
 def select_signing_key(key=None):
     if key is None:
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -821,7 +821,12 @@ def _url_etag_group(url):
     if isinstance(url, str):
         parsed_url = urllib.parse.urlparse(url)
 
-    return parsed_url.geturl()
+    if parsed_url.scheme == "s3":
+        return "s3"
+    if parsed_url.scheme == "file":
+        return "file"
+    else:
+        return parsed_url.host()
 
 
 def select_signing_key(key=None):
--- a/lib/spack/spack/binary_distribution.py
+++ b/lib/spack/spack/binary_distribution.py
@@ -817,6 +817,20 @@ def checksum_tarball(file):
     return hasher.hexdigest()
 
 
+def _url_etag_group(url):
+    if isinstance(url, str):
+        parsed_url = urllib.parse.urlparse(url)
+
+    if "http" in parsed_url.scheme:
+        # HTTP mirrors may have host specific etags
+        etag_group = parsed_url.host
+    else:
+        # s3:// and file:// use a consistent ETag method
+        etag_group = parsed_url.scheme
+
+    return etag_group
+
+
 def select_signing_key(key=None):
     if key is None:
         keys = spack.util.gpg.signing_keys()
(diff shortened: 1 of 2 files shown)


kwryankrattiger:
simple testing done with local buildcache and everything seems to work as expected.

Tested with s3 it seems like the s3 checksum and the checksum in the spec.json don't match at a glance. Maybe I am missing something...

image

image

lib/spack/spack/binary_distribution.py
```diff
@@ -1254,8 +1268,26 @@ def _build_tarball(
...
+        spackfile_path,
+        remote_spackfile_path,
+        keep_original=False,
+        # For S3 buckets, enable the extra checksum SHA256 attribute
+        extra_args={"ChecksumAlgorithm": "SHA256"},
+    )
+    # Try to get the ETag of the uploaded tarball. This must happen after uploading the
+    # tarball since it is not possible to pre-compute the resource etag
+    try:
+        spackfile_etag = web_util.read_etag(remote_spackfile_path)
```
haampie: This has a race condition. Can you not get the etag from the response headers of a push?

lib/spack/spack/util/web.py
```diff
@@ -85,6 +86,27 @@ def handle_starttag(self, tag, attrs):
...
+        s3 = s3_util.get_s3_session(remote_url, method="push")
+        remote_path = remote_url.path
+        while remote_path.startswith("/"):
+            remote_path = remote_path[1:]
+        obj = s3.get_object(Bucket=remote_url.netloc, Key=remote_path)
+        if not obj:
+            raise FetchError("Could not obtain etag or s3 resource")
+        return obj["ETag"]
+    elif remote_url.scheme == "file":
+        return hashlib.md5(open(remote_url.path, "rb").read()).hexdigest()
```
haampie: This is wrong. Use `spack.util.url.file_url_string_to_path`

haampie: Also, avoid reading an entire file into memory; hashlib supports streaming

lib/spack/spack/binary_distribution.py
```diff
@@ -817,6 +817,20 @@ def checksum_tarball(file):
...
 
+def _url_etag_group(url):
+    if isinstance(url, str):
+        parsed_url = urllib.parse.urlparse(url)
+
+    if "http" in parsed_url.scheme:
+        # HTTP mirrors may have host specific etags
+        etag_group = parsed_url.host
+    else:
+        # s3:// and file:// use a consistent ETag method
```
haampie: Do they? minio, openstack etc have an s3-compatible API. Etags are really only useful if you keep track of the pair `(url, etag)`. They're not meant to be universally unique.

lib/spack/spack/util/web.py
```diff
@@ -85,6 +86,27 @@ def handle_starttag(self, tag, attrs):
                     self.links.append(val)
 
 
+def read_etag(url):
```
haampie: See spack.util.web.parse_etag. Can you use that to guard against garbage etag responses? Also note you don't extract the quoted value.

haampie:
This has a race condition because the etag is not obtained from the put response headers. Can you check if that's possible? Otherwise it's not very useful.

lib/spack/spack/util/web.py
kwryankrattiger: I am just cribbing this from other places that use the `urlparse.path`, I am curious how that can be wrong.

As far as the hashlib streaming, I saw that but couldn't get it to work locally, it seems like it may only exist in a later versions? Or I am just doing it wrong.

```
>>> with open("/path/to/file", "rb") as f:
...     digest = hashlib.file_digest(f, "sha256")
... 
  File "<stdin>", line 2, in <module>
AttributeError: module 'hashlib' has no attribute 'file_digest'
```

lib/spack/spack/binary_distribution.py
kwryankrattiger: So since we don't have stable tarballs for binaries (yet) this is certainly not true... I will just use the full URL as the key for now.

lib/spack/spack/util/web.py
kwryankrattiger: yeah, I saw the double quotes getting pulled in. I didn't see the `parse_etag` API, I will use that, thanks!

lib/spack/spack/binary_distribution.py
kwryankrattiger: Is it? I thought the push was blocking. If not, I will see if we can't get it from the header. The reason I didn't do that to begin with is there is, afaict, no actual standard for how an ETag is represented in a response header, or if it is even required to provide one.

kwryankrattiger: So we _ can_ get the ETag from the s3 response...but the current method we use, `upload_file` will not work. It requires using the `put_object` method instead.

I am not sure what the connotations of that are. I will keep looking into it.

kwryankrattiger: `put_object` has some issues...but I think they are solvable. I will push some changes after I verify the fix and you can let me know if you can think of a less horrible solution.

lib/spack/spack/binary_distribution.py
kwryankrattiger: okay, so I did some diggiing, and s3 does have a consistent scheme for identical objects even if they are in a different prefix (same bucket).

I was thinking about the publish mirror in spack-binaries and the way we handle those now is just by copying the `.spack` and `.json.sig` files from the develop mirrors. I did this as a test and the ETag for the original and copy were identical. In the case where they were not, we could have to update the etag in the `.json.sig`, but since they are the same I think it is worth while to keep this logic. If the ETag does end up as different, then that should be a sign that `.json.sig` and and the `.spack` came from a different source.

lib/spack/spack/util/web.py
haampie: Those other places got it wrong too, for example:

```
>>> spack.util.web.url_util.path_to_file_url("/hello world")
'file:///hello%20world'
```

and on Windows it's strictly required.

haampie: See `spack.util.crypto.checksum` regarding chunks

kwryankrattiger:
@spackbot fix style

kwryankrattiger:
So some additional complexities involved this change to use ETags for faster validation.

ETags change depending on the location of the file, so copying a build cache is more complicated
* Copying a build_cache contents will require downloading and updating the `....spec.json*` files to update ETag of binary.
  * Read ETag of the binary in the source location
  * Copy binary and get ETag from (`util.web.copy_to_url(url_source, url_dest)`)
  * Compare the source ETag with the ETag of the copied
    * If same, copy the `spec.json` file directly the the new build cache
    * If they are different
      * If URL scheme `file` -> Error, copy corruption
      * else
        *  Download `spec.json`
        * Update ETag
        * Upload to destination URL

Things that exist already in the mirrors that we want to be able to validate will need to update `spec.json` files to include ETags
* Download tarball, store ETag, compute sha256sum, delete tarball
* Download `spec.json`, read the sha256 checksum, validate, write ETag section, reupload `spec.json`

The ETags do not really provide any benefit other than the very narrow scope of fast validation. Using them as a pre-check for fetching in install or anything like that is more likely to be broken by  a bug in a build cache copy than it is to prevent anything that is not already covered by checking the sha256sum.

Using ETags through cloudfront seems to just forward the s3 Etag. I think this is pretty simple to just add an option for `buildcache validate --no-check-location`.

@haampie @scottwittenburg can either of you think of other things I am missing that would be required to use ETags for validation of a buildcache?

The last thing is it seems that the SHA256 checksum provided by AWS is completely unusable for validation without downloading and processing the entire S3 object and using base64 encoding (although I don't think it is that simple). For those reasons I think it is unusable for the purpose of fast validation. Unless someone else has an idea.

---

Propsed options for `buildcache validate`
* `--no-check-location` - Do not check the location connected with the etag when checking the etag (useful for validation via cloudfront validation)
* `--use-checksum` - allow fallback to use checksum validation, default no
* `--update-etags` - update the etags in the `spec.json` files with invalid (missing or mismatched) ETags. Mismatches with unmatched checksums are reported as errros and ETag is not updated.

haampie:
I'm sort of confused as to what problem is being solved :thinking: you want to locate tarballs with potentially wrong checksums? Notice you can conditionally fetch tarballs with `If-None-Match: <etag>`. Does that help reduce IO?

kwryankrattiger:
@haampie 
> I'm sort of confused as to what problem is being solved thinking you want to locate tarballs with potentially wrong checksums? Notice you can conditionally fetch tarballs with `If-None-Match: <etag>`. Does that help reduce IO?

My comment is a summary a meeting I had with @scottwittenburg today.

The feature I am working on adding is a way to validate if the binaries match the `spec.json` without having to download the entire binary file and compute the checksum. As it turns out, using ETags is the best option. Initially we hoped that using the SHA256 computed by AWS would be useful, but it isn't the checksum of the stored data so it ends up just being a fancy ETag.

The working plan for the solution here is to add the ETag to the `.spec.json` files. To validate binaries, download the json, inspect the ETag, and read the ETag from the binaries to see if they match. If they do, good, if not, log the URL for the `.spec.json` and the `.spack` files as an error.

Since ETags are location dependent, two questions came up.
1. How can we safely copy a build_cache contents so the ETags remain intact. 
2. How to we insert ETags safely into the existing `.spec.json` files so that validation is possible going forward.

I proposed some psuedo code above for how these could be solved.

In my comment, I mention that using the ETag to pre-check before fetching during install is a not good idea. Sure, maybe some reduced IO, but only in the case of failure. In the case of success, which we anticipate to be the common case, this would introduce more IO and an additional failure mode for installation that is more likely to be a red-herring than an actual problem. The only reason I brought this up is so that there is a record of this being thought about and why we decided it was a waste.

PR #35734, opened 2023-02-28
qwt: lift restrictions on qt version and added an opengl variant and VisIt use

nicolaslg:
- qwt prior to 6.1.5 was limited to qt@5.14, which was reflected in the recipe; this limitation is lifted for versions >= 6.1.5
https://sourceforge.net/p/qwt/code/HEAD/tree/tags/qwt-6.1.6/CHANGES-6.1
- added an `opengl` variant with a patch to disable it in the qwtconfig.pri file, similar to what has been done for designer,  that allows building against a `qt~opengl`


nicolaslg:
The `ci/gitlab-ci` fails on the `e4s-power-pr-build`. I added a modification to the `visit` recipe to depend on the new variant `qwt+opengl` in case this is the issue.

alalazo (approved):

cyrush (approved):

PR #35737, opened 2023-02-28
Improve setup build / run / test environment

haampie:
This adds a `SetupContext` class which is responsible for setting
package.py module globals, and computing the changes to environment
variables for the build, test or run context.

The class uses `effective_deptypes` which takes a list of specs (e.g. single
item of a spec to build, or a list of environment roots) and a context
(build, run, test), and outputs a flat list of specs that affect the
environment together with a flag in what way they do so. This list is
topologically ordered from root to leaf, so that one can be assured that
dependents override variables set by dependencies, not the other way
around.

This is used to replace the logic in `modifications_from_dependencies`,
which has several issues: missing calls to `setup_run_environment`, and
the order in which operations are applied.

Further, it should improve performance a bit in certain cases, since
`effective_deptypes` run in O(v + e) time, whereas `spack env activate`
currently can take up to O(v^2 + e) time due to loops over roots. Each
edge in the DAG is visited once by calling `effective_deptypes` with
`env.concrete_roots()`.

By marking and propagating flags through the DAG, this commit also fixes
a bug where Spack wouldn't call `setup_run_environment` for runtime
dependencies of link dependencies. And this PR ensures that Spack
correctly sets up the runtime environment of direct build dependencies.

Regarding test dependencies: in a build context they are are build-time
test deps, whereas in a test context they are install-time test deps.
Since there are no means to distinguish the build/install type test deps,
they're both.

Further changes:

- all `package.py` module globals are guaranteed to be set before any of the
  `setup_(dependent)_(run|build)_env` functions is called
- traversal order during setup: first the group of externals, then the group
  of non-externals, with specs in each group traversed topological (dependencies
  are setup before dependents)
- modules: only ever call `setup_dependent_run_environment` of *direct* link/run
   type deps
- the marker in `set_module_variables_for_package` is dropped, since we should
  call the method once per spec. This allows us to set only a cheap subset of
  globals on the module: for example it's not necessary to compute the expensive
  `cmake_args` and w/e if the spec under consideration is not the root node to be
  built.
- `spack load`'s `--only` is deprecated (it has no effect now), and `spack load x`
  now means: do everything that's required for `x` to work at runtime, which
  requires runtime deps to be setup -- just like `spack env activate`.
- `spack load` no longer loads build deps (of build deps) ...
- `spack env activate` on partially installed or broken environments: this is all
  or nothing now. If some spec errors during setup of its runtime env, you'll only
  get the unconditional variables + a warning that says the runtime changes for
  specs couldn't be applied.
- Remove traversal in upward direction from `setup_dependent_*` in packages.
  Upward traversal may iterate to specs that aren't children of the roots
  (e.g. zlib / python have hundreds of dependents, only a small fraction is
  reachable from the roots. Packages should only modify the direct dependent
  they receive as an argument)


haampie:
Notice that this exposes some weird things, like `pkg` depends on `gettext` as a link dep, and `gettext` has `+tar` as a default, making it depend on `tar` as (build,run), `tar` depends on compression methods gzip/pigz/xz/.. as run deps, so those compression packages are marked `RUNTIME_EXECUTABLE`.

`gettext` is a bit weird.

haampie:
The TODO is modules

haampie:
libxc failing is irrelevant, the download is returning something incomplete.

lib/spack/spack/user_environment.py
```diff
@@ -62,40 +68,57 @@ def unconditional_environment_modifications(view):
...
+
+    yield
+
+    for s in traverse.traverse_nodes(specs, key=lambda s: s.dag_hash()):
+        s.prefix = prefixes.get(s.dag_hash(), s.prefix)
+
+
+def environment_modifications_for_specs(
+    specs: Union[spack.spec.Spec, List[spack.spec.Spec]], view=None, set_package_py_globals=True
+):
```
alalazo: Can we use the following API:
```suggestion
def environment_modifications_for_specs(
    *specs: spack.spec.Spec, view=None, set_package_py_globals=True
):
```
? That will force setting the `view` and `setup_package_py_globals` argument by keyword (which is good imo) and make it unnecessary to have the `isinstance`check at the beginning of the function.

lib/spack/spack/user_environment.py
```diff
@@ -62,40 +68,57 @@ def unconditional_environment_modifications(view):
...
-    env.extend(
-        spack.build_environment.modifications_from_dependencies(
-            spec, context="run", set_package_py_globals=set_package_py_globals
-        )
-    )
-
-    if set_package_py_globals:
-        spack.build_environment.set_module_variables_for_package(spec.package)
+    if not isinstance(specs, list):
+        specs = [specs]
```
alalazo: We can remove this with the suggestion above:
```suggestion
```

lib/spack/spack/user_environment.py
```diff
@@ -62,40 +68,57 @@ def unconditional_environment_modifications(view):
     return env
 
 
-def environment_modifications_for_spec(spec, view=None, set_package_py_globals=True):
+@contextmanager
+def projected_prefix(specs, projection):
```
alalazo: This is worth a docstring to say the context manager substitutes the global prefixes with their projection in-place (and restores them on exit)

alalazo: Can we also add type hints to this function?

lib/spack/spack/user_environment.py
```diff
@@ -62,40 +68,57 @@ def unconditional_environment_modifications(view):
...
-    if set_package_py_globals:
-        spack.build_environment.set_module_variables_for_package(spec.package)
+    if not isinstance(specs, list):
+        specs = [specs]
 
-    spec.package.setup_run_environment(env)
+    env = environment.EnvironmentModifications()
+    topo_ordered = traverse.traverse_nodes(specs, root=True, deptype=("run", "link"), order="topo")
+
+    with projected_prefix(specs, view.get_projection_for_spec) if view else nullcontext():
```
alalazo: (nitpick) This is probably easier to read if split into 2 lines (assign the context manager first, use it in the next line).

lib/spack/spack/user_environment.py
```diff
@@ -62,40 +68,57 @@ def unconditional_environment_modifications(view):
...
+        # Static environment changes (prefix inspections)
+        for s in reversed(list(topo_ordered)):
+            static = environment.inspect_path(
```
alalazo, line 114: Just for discussion, would we gain something in clarity if we used a visitor pattern also here?

lib/spack/spack/user_environment.py
```diff
@@ -62,40 +68,57 @@ def unconditional_environment_modifications(view):
...
+                s.prefix, prefix_inspections(s.platform), exclude=environment.is_system_path
+            )
+            env.extend(static)
+
+        # Dynamic environment changes (setup_run_environment etc)
```
alalazo, line 119: 👍 for the high level description of the intent

lib/spack/spack/build_environment.py
```diff
@@ -563,46 +568,42 @@ def determine_number_of_jobs(
     return min(max_cpus, config_default)
 
 
-def set_module_variables_for_package(pkg):
+def set_module_variables_for_package(pkg, context="build"):
```
alalazo: Can we make `context` an `enum`?

lib/spack/spack/build_environment.py
```diff
@@ -835,154 +831,256 @@ def setup_package(pkg, dirty, context="build"):
...
+    in reverse so that dependents override dependencies, not the other way around.
+    """
+    assert context in ("build", "run", "test")
+
+    visitor = traverse.TopoVisitor(
+        EnvironmentVisitor(specs, context), key=lambda x: x.dag_hash(), root=True, all_edges=True
+    )
+    traverse.traverse_depth_first_with_visitor(traverse.with_artificial_edges(specs), visitor)
+
+    modes = defaultdict(lambda: Mode(0))
```
alalazo: (Minor) I'm in doubt whether a comment to say that this represents the zeroed flag is worth it or not. 🤔

lib/spack/spack/build_environment.py
```diff
@@ -835,154 +831,256 @@ def setup_package(pkg, dirty, context="build"):
...
+    # with out-edges. Can this be improved?
+    for spec, mode in modes.items():
+        if mode and Mode.ADDED not in mode:
+            nodes_with_type.append((spec, mode))
+
+    return nodes_with_type
+
+
+def modifications_from_dag(
+    specs: Union[spack.spec.Spec, List[spack.spec.Spec]],
```
alalazo: Same as above, we can use:
```suggestion
    *specs: spack.spec.Spec,
```

lib/spack/spack/build_environment.py
```diff
@@ -835,154 +831,256 @@ def setup_package(pkg, dirty, context="build"):
...
+            package.py files. Warning: never use this, as setup_run_environment and others
+            may depend on package.py globals.
     """
-    if context not in ["build", "run", "test"]:
-        raise ValueError(
-            "Expecting context to be one of ['build', 'run', 'test'], " "got: {0}".format(context)
-        )
+    if context not in ("build", "run", "test"):
+        raise ValueError(f"Expecting context to be one of build, run, test. Got {context}")
+
```
alalazo: This can be deleted if we use an `enum` for `context`

lib/spack/spack/build_environment.py
```diff
@@ -835,154 +831,256 @@ def setup_package(pkg, dirty, context="build"):
...
+    if context in ("run", "test"):
+        should_setup_run_env |= Mode.ROOT
+        should_be_runnable |= Mode.ROOT
+
+    # I guess everything that calls setup_run_environemnt and setup_dependent_* should have our
+    # horrible globals available too.
+    should_populate_package_py_globals = (
+        should_setup_dependent_build_env | should_setup_run_env | Mode.ROOT
+    )
```
alalazo: (Not necessarily for this PR) I agree that globals are nasty, but people rely on them. An idea to gracefully deprecated them is to:
1. Wrap the globals in some object that can emit a message before doing what the global is expected to do
2. Decide how noisy we want to be at the beginning, and how long should we support globals

We can then either go for a slow deprecation (e.g. start emitting debug only messages and turn them into warning messages in a couple of releases) or a faster one (warning message, removal in v0.21) or something in between.

lib/spack/spack/cmd/load.py
```diff
@@ -102,16 +101,14 @@ def load(parser, args):
...
-    with spack.store.db.read_transaction():
-        if "dependencies" in args.things_to_load:
-            include_roots = "package" in args.things_to_load
-            specs = [
-                dep for spec in specs for dep in spec.traverse(root=include_roots, order="post")
-            ]
+    if args.things_to_load != "package,dependencies":
+        tty.warn(
+            "The `--only` flag in spack load is deprecated and will be removed in Spack v0.21.0"
+        )
```
alalazo: 👍

alalazo (changes requested):
I started reviewing the PR and have a few initial comments. I think most package changes are worth being extracted and been given a fast track to merge. Let me know what you think on that.

lib/spack/spack/user_environment.py
haampie: It's also a footgun, cause `f(a, b)` has to be replaced with `f(a, b=b)`, but yeah, I went for this.

lib/spack/spack/build_environment.py
haampie: It's done everywhere except where it's assumed to be a string when serialized etc.

lib/spack/spack/build_environment.py
haampie: done

lib/spack/spack/build_environment.py
haampie: Yeah, let's save that for later

lib/spack/spack/user_environment.py
haampie: I don't think so, it's quite verbose. It would help avoid a double revert, though, since the default order you get out of the algorithm is reversed.

haampie:
@spackbot rebuild everything

lib/spack/spack/cmd/common/env_utility.py
```diff
@@ -41,14 +41,14 @@ def setup_parser(subparser):
...
 class AreDepsInstalledVisitor:
-    def __init__(self, context="build"):
-        if context not in ("build", "test"):
-            raise ValueError("context can only be build or test")
-
-        if context == "build":
-            self.direct_deps = ("build", "link", "run")
+    def __init__(self, context: Context = Context.BUILD):
+        if context == Context.BUILD:
+            # TODO: run deps shouldn't be required for build env.
```
trws, line 47: Does this mean we get failures if run is removed here?

haampie: Yeah, this is used in `spack build-env <spec>`. If runtime deps of the root are not installed, Spack errors.

If I remember correctly this mirrors the installer: `spack install --only=package <spec>` would also error if the root has missing run deps.

I guess that we require pure run deps to be installed s.t. the dependents can in principle register absolute paths to their executables. An example of this is `tar`: it has a runtime dep on `zstd`, but also works without setting `PATH=<zstd prefix>/bin`, because the absolute path to the run dependency registered. (In principle this still doesn't require zstd to be installed first... but if zstd is installed first, you could search for the executable in zstd's bin dir, I can imagine that configure scripts do this)

trws:
That sounds right to be honest. Do we want to have build-env silently succeed with only a partial environment?

---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>
On October 2, 2023 at 11:39:29 AM GMT+2, Harmen Stoppels ***@***.***> wrote:

@haampie commented on this pull request.

________________________________

In lib/spack/spack/cmd/common/env_utility.py<https://urldefense.us/v3/__https://github.com/spack/spack/pull/35737*discussion_r1342477426__;Iw!!G2kpM7uM-TzIFchu!xLv1jrmolrtH-Q9LNQfh31OkFKyeAT9rfstPRMNPQFCSzPDN_RAq6Qr-Sa87yv1o1vifdnEUlPi4NRLS_cSvgkWQtyA$>:

> @@ -41,14 +41,14 @@ def setup_parser(subparser):


 class AreDepsInstalledVisitor:
-    def __init__(self, context="build"):
-        if context not in ("build", "test"):
-            raise ValueError("context can only be build or test")
-
-        if context == "build":
-            self.direct_deps = ("build", "link", "run")
+    def __init__(self, context: Context = Context.BUILD):
+        if context == Context.BUILD:
+            # TODO: run deps shouldn't be required for build env.


Yeah, this is used in spack build-env <spec>. If runtime deps of the root are not installed, Spack errors.

If I remember correctly this mirrors the installer: spack install --only=package <spec> would also error if the root has missing run deps.

haampie:
So, things that I don't like but also don't see how to fix:

1. Packages can override CXX, which hip packages "exploit" to use hipcc+clang instead of Spack's %gcc.
2. `modules:prefix_inspections` being too noisy imho.

trws:
Overriding CXX seems especially unfortunate to be honest.  It's not wrong in a sense, but it's really, really wrong in another since it's supposed to be the HIP compiler and not necessarily the C or C++ compilers (though maybe the C++ compiler :facepalm:).  That's a nasty compilers as deps problem really, @tgamblin, @alalazo, @becker33, any thoughts on handling language *dialect* deps? In a sense hip is a language, the compilers and cmake treat it that way, but it's also a dialect of C++, so it should be able to provide both, and *maybe* override both, but allow C++ to come from something else as well in the same project.

haampie:
@spackbot run pipeline

lib/spack/spack/cmd/load.py
alalazo: Do we need to bump version here?

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+    for spec, mode in modes.items():
+        if mode and not (ADDED & mode):
+            nodes_with_type.append((spec, mode))
+
+    return nodes_with_type
+
+
+class SetupContext:
+    """This class encapsulates the logic to determine environment modifications, and is used as
+    well to set magic globals in modules of package.py."""
```
alalazo: (minor) Let's not talk about :magic_wand: :
```suggestion
    well to set global names in modules of package.py."""
```

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
 
 
-def modifications_from_dependencies(
-    spec, context, custom_mods_only=True, set_package_py_globals=True
-):
-    """Returns the environment modifications that are required by
-    the dependencies of a spec and also applies modifications
-    to this spec's package at module scope, if need be.
+#: Entrypoint spec (a spec to be built; an env root, etc)
+ROOT = 0b1
```
alalazo: I know we are using the same trick for deptypes (using `int` directly), but there it was due to performance reasons. I would expect setting up a context is not done with the same frequency, so it won't be a huge issue performance wise. If that is the case, can't we use `enum.Flag` ?

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
-        if os.path.isdir(bin_dir):
-            env.prepend_path("PATH", bin_dir)
+class EnvironmentVisitor:
+    def __init__(self, *roots: spack.spec.Spec, context: Context):
+        # For the roots (well, marked specs) we follow different edges
+        # than for their deps, depending on the context.
+        self.root_hashes = set(s.dag_hash() for s in roots)
+
+        if context == Context.BUILD:
+            # Drop direct run deps in build context
```
alalazo, line 812: Is this a change in behavior? I need to double check, but looking at the current code (which is quite involved) it seems we are adding to the PATH all the `<prefix>/bin` directories for packages involved in a given build, regardless of the dependency type. Am I reading this correctly? 

To be clear, my main concern here is that we don't drop having a runtime dep in PATH if that is used by configuration scripts to detect it.

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+            # We don't really distinguish between install and build time test deps,
+            # so we include them here as build-time test deps.
```
alalazo, line 814: (Only a comment) I agree that test dependencies need a complete refactor, and this has been an issue for a long time. Maybe we should target that for v0.22

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+
+    for edge in visitor.edges:
+        key = edge.spec
+
+        # Mark the starting point
+        if edge.parent is None:
+            modes[key] = ROOT
+            continue
+
+        mode = modes[edge.parent]
```
alalazo:
```suggestion
        parent_mode = modes[edge.parent]
```

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+        all_edges=True,
+    )
+    traverse.traverse_depth_first_with_visitor(traverse.with_artificial_edges(specs), visitor)
+
+    # Dictionary with "no mode" as default value, so it's easy to write modes[x] |= flag.
+    modes = defaultdict(lambda: 0)
+    nodes_with_type = []
+
+    for edge in visitor.edges:
+        key = edge.spec
```
alalazo:
```suggestion
        current_child = edge.spec
```

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+        if (RUNTIME | RUNTIME_EXECUTABLE | BUILDTIME_DIRECT) & mode:
+            if dt.LINK & deptypes:
+                modes[key] |= RUNTIME
+            if dt.RUN & deptypes:
+                modes[key] |= RUNTIME_EXECUTABLE
+
+        # Propagate BUILDTIME through link deps.
+        if BUILDTIME & mode:
+            if dt.LINK & deptypes:
+                modes[key] |= BUILDTIME
```
alalazo: (Minor) We can avoid 2 levels of indentation here

lib/spack/spack/build_environment.py
alalazo: Can we also use something like `current_parent` as a variable name, instead or repeating `edge.parent`?

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+        # before out-edges, meaning that edge.parent is done.
+        if not (ADDED & mode):
+            modes[edge.parent] |= ADDED
+            nodes_with_type.append((edge.parent, mode))
+
+    # Attach the leaf nodes, since we only added nodes
+    # with out-edges. Can this be improved?
+    for spec, mode in modes.items():
+        if mode and not (ADDED & mode):
+            nodes_with_type.append((spec, mode))
```
alalazo:
```suggestion
    # Attach the leaf nodes, since we only added nodes with out-edges.
    for spec, mode in modes.items():
        if mode and not (ADDED & mode):
            nodes_with_type.append((spec, mode))
```

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+        specs_with_type = effective_deptypes(*specs, context=context)
+
+        self.specs = specs
+        self.context = context
+        self.external: List[spack.spec.Spec]
+        self.nonexternal: List[spack.spec.Spec]
+        # Reverse so we go from leaf to root
+        self.nodes_in_subdag = set(id(s) for s, _ in specs_with_type)
+
+        # Split into non-external and extenal, maintaining topo order per group.
```
alalazo: Typo:
```suggestion
        # Split into non-external and external, maintaining topo order per group.
```

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+        self.external, self.nonexternal = stable_partition(
+            reversed(specs_with_type), lambda t: t[0].external
         )
+        self.should_be_runnable = BUILDTIME_DIRECT | RUNTIME_EXECUTABLE
+        self.should_setup_dependent_build_env = BUILDTIME | BUILDTIME_DIRECT
+        self.should_setup_run_env = RUNTIME | RUNTIME_EXECUTABLE
```
alalazo: I wonder if we can simplify this further... These values depend only on `context` :thinking:

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
-        if not dspec.external:
-            add_modifications_for_dep(dspec)
+        # Everything that calls setup_run_environemnt and setup_dependent_* needs globals set.
+        self.should_populate_package_py_globals = (
+            self.should_setup_dependent_build_env | self.should_setup_run_env | ROOT
+        )
 
-    return env
+    def set_module_globals(self):
+        """Set the magic globals in modules of package.py files."""
```
alalazo:
```suggestion
        """Set global variables in modules of package.py files."""
```

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+        for dspec, flag in chain(self.external, self.nonexternal):
+            pkg = dspec.package
+
+            if self.should_populate_package_py_globals & flag:
+                if self.context == Context.BUILD and (ROOT | BUILDTIME_DIRECT) & flag:
```
alalazo: Should/can we push this check, for consistency, into more fine grained `should_*` kind of varianbles?

lib/spack/spack/build_environment.py
```diff
@@ -813,158 +801,242 @@ def setup_package(pkg, dirty, context="build"):
...
+
+            if self.should_setup_run_env & flag:
+                # TODO: remove setup_dependent_run_environment...
+                for spec in dspec.dependents(deptype=dt.RUN):
+                    if id(spec) in self.nodes_in_subdag:
+                        pkg.setup_dependent_run_environment(env, spec)
+                pkg.setup_run_environment(env)
+        return env
+
+    def _make_buildtime_detectable(self, dep, env):
```
alalazo: Can we add type hints to modified / added functions and methods?

lib/spack/spack/environment/environment.py
```diff
@@ -1691,40 +1691,17 @@ def check_views(self):
...
-                        errors.append((msg, str(e)))
-                        continue
-
-                    all_mods.extend(mods.reversed() if reverse else mods)
-
-        return all_mods, errors
+        try:
+            mods = uenv.environment_modifications_for_specs(
+                *self.concrete_roots(), view=self.default_view
+            )
```
alalazo: Client use is kind of nice (the high-level modules are less cluttered with implementation details)

alalazo (changes requested):
First batch of comments and questions. Will follow with more comments later. I agree with the high-level approach, and I like that this simplifies client calls.

lib/spack/spack/build_environment.py
haampie: Sure, but it's consistent with the rest of the "state machine".

lib/spack/spack/environment/environment.py
haampie: I still don't like it too much :p the fact that "static" and "dynamic" variables are added separately, and the fact that static ones overlap dynamic ones.

lib/spack/spack/build_environment.py
haampie: Hm yeah, I like Python less and less now I know that trivial types always allocate. Basically every trivial x & y == z is an allocation and requires gc :/ Do we care about enums? Also using an enum requires the thing has a name

lib/spack/spack/build_environment.py
haampie: in principle context could be static & part of the function name. setup_run_context(...), setup_build_context(...) etc. But that causes duplication

lib/spack/spack/build_environment.py
haampie: Yeah, it is a change of behavior.

Do we have good reasons to put them in the PATH?

As an example:

A package like `tar` has `run` deps on compression packages like `xz` / `bzip2` / ... to hard-code paths to `bin/*`. I guess another way of looking at that is that it's a `link` type dep, since the executable path is fixed instead of found in PATH.

If the package would need `which("xz")` or so to locate the executable without guessing the path, it could even just be marked as a `build` / `link` dep, not run at all.

Notice that if e.g. `xz` itself had run-type deps, they would appear in the `PATH` regardless with my changes (run type deps of link type deps are RUNTIME_EXECUTABLE)

lib/spack/spack/cmd/load.py
alalazo: @haampie I think this :point_up: needs to be:
```suggestion
    if args.things_to_load != "package,dependencies":
        tty.warn(
            "The `--only` flag in spack load is deprecated and will be removed in Spack v0.22.0"
        )
```

alalazo:
I think we can improve the code further, but this is way better than the current state already. 

This https://github.com/spack/spack/pull/35737#discussion_r1354311703 may need some agreement on (pure runtime deps of root in a build context), and we might need to document what we decide about the meaning of build types. Documentation may happen in a following PR.

Another point is:

> Remove traversal in upward direction from setup_dependent_* in packages. Upward traversal may iterate to specs that aren't children of the roots (e.g. zlib / python have hundreds of dependents, many of which may not be in the build environment). Packages should be happy with the dependent they receive as an arg and not try to be smart traversing the wrong dag.

with which I completely agree. I wonder if we can check that traversal is not added back in those functions :thinking:

haampie:
@tgamblin do you wanna have a look too?

alalazo (dismissed):
Besides https://github.com/spack/spack/pull/35737#issuecomment-1759554440 this LGTM. Waiting for @tgamblin review before merging.

haampie:
I'm still leaning towards dropping run deps in the build env from PATH, leaving the PR as is.

If this was a cross compilation type of environment, then run deps are for the target, so makes no sense to have them in PATH.

If you wanna hard-code absolute paths to executables that you have to locate during the build, then it's much more like a link dep.

haampie:
Rebased this to trigger Gitlab CI, nothing changed wrt:

> I'm still leaning towards dropping run deps in the build env from PATH, leaving the PR as is.

> If this was a cross compilation type of environment, then run deps are for the target, so makes no sense to have them in PATH.

> If you wanna hard-code absolute paths to executables that you have to locate during the build, then it's much more like a link dep.

haampie:
Yet another rebase w/o changes.

lib/spack/spack/build_environment.py
tgamblin: I think this is a good change.  If CI passes I'm good with it, and I think we should be making packages be more explicit about their dependencies -- even ones from configure.

lib/spack/spack/cmd/load.py
```diff
@@ -108,16 +110,14 @@ def load(parser, args):
...
-    with spack.store.STORE.db.read_transaction():
-        if "dependencies" in args.things_to_load:
-            include_roots = "package" in args.things_to_load
-            specs = [
-                dep for spec in specs for dep in spec.traverse(root=include_roots, order="post")
-            ]
+    if args.things_to_load != "package,dependencies":
+        tty.warn(
+            "The `--only` flag in spack load is deprecated and will be removed in Spack v0.22"
+        )
```
tgamblin, line 116: I think we should undo this, but in a follow-on before 0.21.  I don't think the semantics of `--only dependencies` are unclear, and I think people actually do use that, e.g. with `BundlePackages`.

Can you say what's unclear about `--only package`?  I don't think it's *that* unclear... I also think it's not great that it is incomplete... I'm not sure what you'd use it for.

tgamblin (dismissed):
Approving but see note -- I don't think the deprecation should stay, at least not for all args.  Let's talk about that one in a follow-on.

lib/spack/spack/cmd/load.py
haampie: I don't see the point of `spack load --only=package py-x` if you don't get `python` in `PATH`, or `spack load --only=package tar` not putting `gzip` / `xz` / etc in `PATH`.

haampie:
Had to rebase once more

alalazo (approved):

balay:
This PR breaks petsc build with oneapi compilers

```
Executing: /software/intel/oneapi/mpi/2021.10.0/bin/mpiicc -E  -I/tmp/petsc-gmbc5xcy/config.setCompilers  /tmp/petsc-gmbc5xcy/config.setCompilers/conftest.c
Possible ERROR while running preprocessor:exit code 0
stdout:
stdout:
# 1 "/tmp/petsc-gmbc5xcy/config.setCompilers/conftest.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
<snip>
# 1156 "/usr/include/stdlib.h" 2 3 4
# 4 "/tmp/petsc-gmbc5xcy/config.setCompilers/conftest.c" 2

stderr:
icx: warning: -Wl,-rpath,/software/intel/oneapi/mkl/2023.2.0/lib/intel64: 'linker' input unused [-Wunused-command-line-argument]
Source:
#include "confdefs.h"
#include "conffix.h"
#include <stdlib.h>
```

Somehow this PR change is triggering this `STDERR` message? [this breaks this simple configure test - resulting in build failure]

ref: `spack install   petsc %oneapi@2023.2.0`

balay:
> icx: warning: -Wl,-rpath,/software/intel/oneapi/mkl/2023.2.0/lib/intel64: 'linker' input unused [-Wunused-command-line-argument]

Checking how to deal with this issue.

balay:
For now working around this issue by adding  -Wno-unused-command-line-argument to compilers.yaml
```
- compiler:
    spec: oneapi@=2023.2.0
    paths:
      cc: /software/intel/oneapi/compiler/2023.2.0/linux/bin/icx
      cxx: /software/intel/oneapi/compiler/2023.2.0/linux/bin/icpx
      f77: /software/intel/oneapi/compiler/2023.2.0/linux/bin/ifx
      fc: /software/intel/oneapi/compiler/2023.2.0/linux/bin/ifx
    #flags: {}
    flags:
      cflags: -Wno-unused-command-line-argument
      cxxflags: -Wno-unused-command-line-argument
    operating_system: fedora39
    target: x86_64
    modules: []
    environment: {}
    extra_rpaths: []

```

balay:
> For now working around this issue by adding -Wno-unused-command-line-argument to compilers.yaml

this didn't work. It turned out to be a "wack-a-mole" thingy - with different errors coming up with usage of different flags :(

haampie:
@balay looks like intel packages are running

```python
        """Adds environment variables to the generated module file.

        These environment variables come from running:

        .. code-block:: console

           $ source {prefix}/{component}/{version}/env/vars.sh
        """
        # Only if environment modifications are desired (default is +envmods)
        if "~envmods" not in self.spec:
            env.extend(
                EnvironmentModifications.from_sourcing_file(
                    join_path(self.component_prefix, "env", "vars.sh"), *self.env_script_args
                )
            )
```

and we call `setup_run_environment` in more cases than before when setting up the build environment.

Can you check if #40916 fixes your issue?

Otherwise can you check the difference between `spack build-env ...` before / after this PR?

balay:
> Can you check if https://github.com/spack/spack/pull/40916 fixes your issue?

@haampie this MR doesn't work for me.

I'm not sure why my earlier debugging of this issue pointed to this PR [reverting this PR change worked for me at that time].

Now (with my simpler build command - without xsdk)  I think the  trigger is the presence of the  following new values in env:
```
__INTEL_POST_CFLAGS=-Wl,-rpath,/opt/intel/oneapi/mkl/2022.2.0/lib/intel64
__INTEL_POST_FFLAGS=-Wl,-rpath,/opt/intel/oneapi/mkl/2022.2.0/lib/intel64
```
And the commit (from "git bisect") that makes the difference is: e5f3ffc04fb7a1fea237226210b0cafc9246d0f2

And my current work-around is at: https://github.com/spack/spack/commit/62915c25cdbccfe47f9fd31fd390407dc6d1f74a

Wrt e5f3ffc04fb7a1fea237226210b0cafc9246d0f2 - attaching build-env - before and after this commit.

``` console
xsdk@pj02:~/spack$ git checkout -q e5f3ffc04fb7a1fea237226210b0cafc9246d0f2~1
xsdk@pj02:~/spack$ ./bin/spack build-env petsc~fortran~metis~hypre~hdf5 > build-env-prev.txt
xsdk@pj02:~/spack$ git checkout -q e5f3ffc04fb7a1fea237226210b0cafc9246d0f2
xsdk@pj02:~/spack$ ./bin/spack build-env petsc~fortran~metis~hypre~hdf5 > build-env-after.txt
xsdk@pj02:~/spack$
```

[build-env-prev.txt](https://github.com/spack/spack/files/13273248/build-env-prev.txt)
[build-env-after.txt](https://github.com/spack/spack/files/13273249/build-env-after.txt)

PR #35738, opened 2023-02-28
Add a `py-gmxapi` package.

eirrgang:
This package provides the Python package for the GROMACS public API. The Python package is not strongly coupled to a specific GROMACS _version_, but its compiled extension module is strongly coupled to a specific GROMACS _installation_.


eirrgang:
Note that `py-gmxapi` cannot build against `gromacs@2023` until https://github.com/spack/spack/pull/35672 is merged.

eirrgang:
This is a new package, but maybe it could be reviewed by the [gromacs](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/gromacs/package.py) maintainers?
@junghans ? @marvinbernhardt ?

junghans (dismissed):

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,69 @@
...
+    depends_on("py-packaging", type="run")
+    depends_on("py-pybind11@2.6:", type="build")
+    depends_on("py-pytest", type="test")
+
+    conflicts("gromacs+mdrun_only", msg="gmxapi requires the full GROMACS library installation.")
+    conflicts("gromacs~shared", msg="gmxapi requires the full GROMACS library installation.")
+    conflicts("gromacs@:2022.0", when="@0.4:", msg="Use GROMACS 2022.1 or newer for gmxapi 0.4.")
+    conflicts(
+        "gromacs+mpi", when="@:0.3.2", msg="Use gmxapi 0.4 or higher for MPI-enabled GROMACS."
+    )
```
eirrgang: The installation and dependency checking seemed to work fine, but the `spack audit` complained. Are the `conflicts` supposed to be in the form of the current package spec? I.e. should I use explicit dependency syntax?
```suggestion
    conflicts("^gromacs+mdrun_only", msg="gmxapi requires the full GROMACS library installation.")
    conflicts("^gromacs~shared", msg="gmxapi requires the full GROMACS library installation.")
    conflicts("^gromacs@:2022.0", when="@0.4:", msg="Use GROMACS 2022.1 or newer for gmxapi 0.4.")
    conflicts(
        "^gromacs+mpi", when="@:0.3.2", msg="Use gmxapi 0.4 or higher for MPI-enabled GROMACS."
    )
```

junghans: @adamjstewart please help.

adamjstewart: Yes, you should add the ^ at the beginning because it is a dependency

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,69 @@
...
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+import os
+
+import spack.build_systems.python
+from spack.package import *
+
+
+class PyGmxapi(spack.build_systems.python.PythonPackage):
```
adamjstewart:
```suggestion
class PyGmxapi(PythonPackage):
```

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,69 @@
...
+
+    depends_on("gromacs")
+    depends_on("mpi")
+    depends_on("py-mpi4py")
+    depends_on("py-networkx@2.0:", type="run")
+    depends_on("py-numpy@1.7:", type="run")
+    depends_on("py-setuptools@42:", type="build")
+    depends_on("py-packaging", type="run")
+    depends_on("py-pybind11@2.6:", type="build")
+    depends_on("py-pytest", type="test")
```
adamjstewart: Missing build:
```toml
requires = ["cmake>=3.16", "mpi4py", "pybind11>=2.6", "setuptools>=42.0", "wheel"]
```
and build/run:
```cfg
install_requires =
        mpi4py
        packaging
        pybind11
        networkx>=2.0
        numpy>1.7
```
dependencies.

var/spack/repos/builtin/packages/py-gmxapi/package.py
adamjstewart: Wouldn't it be easier to say:
```python
depends_on("gromacs@2022.1:~mdrun_only+shared")
```

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,69 @@
...
+
+    conflicts("gromacs+mdrun_only", msg="gmxapi requires the full GROMACS library installation.")
+    conflicts("gromacs~shared", msg="gmxapi requires the full GROMACS library installation.")
+    conflicts("gromacs@:2022.0", when="@0.4:", msg="Use GROMACS 2022.1 or newer for gmxapi 0.4.")
+    conflicts(
+        "gromacs+mpi", when="@:0.3.2", msg="Use gmxapi 0.4 or higher for MPI-enabled GROMACS."
+    )
+
+
+class PythonPipBuilder(spack.build_systems.python.PythonPipBuilder):
```
adamjstewart: No need to create a separate builder class unless the package supports multiple build systems. You can just put this stuff in the regular package class.

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def install(self, pkg, spec, prefix):
+        global pip
+        # `pip` is a module attribute that is injected by the py-pip package
+        # via setup_dependent_package. It defines an extension of the `command`
+        # from the `python` spec, an instance of spack.util.executable.Executable.
+        pip.add_default_env("GROMACS_DIR", self.spec["gromacs"].prefix)
+        pip.add_default_env("gmxapi_ROOT", self.spec["gromacs"].prefix)
+        pip.add_default_env("Python3_ROOT", self.spec["python"].prefix)
+        super().install(pkg, spec, prefix)
```
adamjstewart: Replace this entire method with:
```python
    env.set("GROMACS_DIR", self.spec["gromacs"].prefix)
    env.set("gmxapi_ROOT", self.spec["gromacs"].prefix)
    env.set("Python3_ROOT", self.spec["python"].home)

File: var/spack/repos/builtin/packages/py-gmxapi/package.py [hunk 6]
```diff
@@ -0,0 +1,69 @@
...
+
+    @run_after("install")
+    def install_test(self):
+        if not self.pkg.run_tests:
+            return
+
```
adamjstewart:
```suggestion
    @run_after("install")
    @on_package_attributes(run_tests=True)
    def install_test(self):
```

adamjstewart (changes requested):

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: Yes. Actually, I had that locally, but I wasn't sure whether to keep the `conflicts("^gromacs+mpi", when="@:0.3.2", msg="Use gmxapi 0.4 or higher for MPI-enabled GROMACS.")` as a `conflicts` line or a second conditional `depends_on`. I figured I'd bring that up after resolving the original `audit` question. But I'll include in the next commit.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: Yes. Originally I thought I would need to add a `build` phase. But no need for the complexity, it seems. Will simplify.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: I thought I'd tried that. I will try again, but I got the impression that would only affect the `build` phase, whereas here it is needed in the `install` phase, unless I override the normal protocol for the PythonPipBuilder.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: Requires eliminating the separate Builder class, I believe. Will do in combination with the suggested restructuring.

eirrgang (commented):
Thanks! New commit coming, pending some more local testing.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: Seems to work. Added to upcoming commit.

Actually, what I had tried before was to add these through `setup_dependent_build_environment` in the `gromacs` package. I don't know if it didn't work because of different semantics or because I failed to bust the cache. In any case, though, it is appropriate to add here, I think, because this version of the package explicitly relies on these, whereas revisions to the gmxapi build system are planned.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: updated

eirrgang: Note: I did not include `pybind11` in the "run" dependencies. It was unnecessary, and will be removed in future package metadata. Ref: https://gitlab.com/gromacs/gromacs/-/merge_requests/3525

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,53 @@
...
+    depends_on("py-setuptools@42:", type="build")
+    depends_on("py-packaging", type=("build", "run"))
+    depends_on("py-pybind11@2.6:", type="build")
+    depends_on("py-pytest@4.6:", type="test")
+    depends_on("py-wheel", type="build")
+
+    conflicts("^gromacs@:2022.0", when="@0.4:", msg="Use GROMACS 2022.1 or newer for gmxapi 0.4.")
+    conflicts(
+        "^gromacs+mpi", when="@:0.3.2", msg="Use gmxapi 0.4 or higher for MPI-enabled GROMACS."
+    )
```
adamjstewart: Since this package only contains a single version, I don't see any reason to keep these conflicts. All of this can be encoded in `depends_on`

eirrgang: Do all supported versions need to be listed?

Alternatively, is there a preferred way to block installation of unsupported versions?

My preference would be to encourage people to install the latest version, or at least >=0.4, but someone might explicitly try to install an older version, in which case I wanted to provide best-effort support.

What do you suggest?

adamjstewart: I only add the versions I need (usually the latest version) and only consider the versions listed in the package. So my preference would be to pretend that 0.3 doesn't exist and write the constraint as:
```python
depends_on("gromacs@2022.1:~mdrun_only+shared")
```
No conflicts, no `when`, no complexity.

var/spack/repos/builtin/packages/py-gmxapi/package.py
adamjstewart: `setup_build_environment` affects all of `spack install`, `setup_run_environment` affects `spack load` and `module load`. The `dependent` versions of these functions affect packages that depend on `py-gmxapi`.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: That means that if someone does `spack install py-gmxapi@0.3 ^gromacs+mpi` they will experience preventable problems.

Is it normal to ignore such usage as edge cases?

I'm not complaining. Just doing a culture-check, I guess.

eirrgang: I mean, I'm just assuming that people don't look at the actual package file for every `spack install` they do.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: > `setup_build_environment` affects all of `spack install`, `setup_run_environment` affects `spack load` and `module load`.

Oh! Thanks for the clarification.

> The `dependent` versions of these functions affect packages that depend on `py-gmxapi`.

I meant that I had tried to add the `dependent` version of the function to the dependency package in `gromacs/package.py`. (I'm sorry if I wasn't clear.) I haven't double-checked whether that actually works. I may have tried it with an insufficiently clean spack state. If it would represent an important bug that should be documented, I'd be happy to follow up.

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: Per @adamjstewart : Don't bother with (second-class) support for older versions. 
```suggestion
    depends_on("gromacs@2022.1:~mdrun_only+shared")
    depends_on("mpi")
    depends_on("py-cmake@3.16:", type="build")
    depends_on("py-importlib-metadata", type="test", when="^python@:3.7")
    depends_on("py-mpi4py", type=("build", "run"))
    depends_on("py-networkx@2.0:", type=("build", "run"))
    depends_on("py-numpy@1.7:", type=("build", "run"))
    depends_on("py-setuptools@42:", type="build")
    depends_on("py-packaging", type=("build", "run"))
    depends_on("py-pybind11@2.6:", type="build")
    depends_on("py-pytest@4.6:", type="test")
    depends_on("py-wheel", type="build")
```

eirrgang: > my preference would be to pretend that 0.3 doesn't exist

Ah, I see you were already clear.

adamjstewart: Yes, the majority of people who use Spack would use `spack install py-gmxapi` without specifying a version (or at least I think). If they tried to use 0.3, Spack would complain that it isn't checksummed and warn them against using it. They can also run `spack info` or `spack versions` to see which versions are available in the package, they don't need to check the package.py.

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,48 @@
...
+    depends_on("gromacs@2022.1:~mdrun_only+shared")
+    depends_on("mpi")
+    depends_on("py-cmake@3.16:", type="build")
+    depends_on("py-importlib-metadata", type="test", when="^python@:3.7")
+    depends_on("py-mpi4py", type=("build", "run"))
+    depends_on("py-networkx@2.0:", type=("build", "run"))
+    depends_on("py-numpy@1.7:", type=("build", "run"))
+    depends_on("py-setuptools@42:", type="build")
+    depends_on("py-packaging", type=("build", "run"))
+    depends_on("py-pybind11@2.6:", type="build")
```
adamjstewart:
```suggestion
    depends_on("py-pybind11@2.6:", type=("build", "run"))
```
According to `setup.cfg`

var/spack/repos/builtin/packages/py-gmxapi/package.py
```diff
@@ -0,0 +1,48 @@
...
+    pypi = "gmxapi/gmxapi-0.4.0.tar.gz"
+    version("0.4.0", sha256="7fd58e6a4b1391043379e8ba55555ebeba255c5b394f5df9d676e6a5571d7eba")
+
+    depends_on("gromacs@2022.1:~mdrun_only+shared")
+    depends_on("mpi")
+    depends_on("py-cmake@3.16:", type="build")
+    depends_on("py-importlib-metadata", type="test", when="^python@:3.7")
+    depends_on("py-mpi4py", type=("build", "run"))
+    depends_on("py-networkx@2.0:", type=("build", "run"))
+    depends_on("py-numpy@1.7:", type=("build", "run"))
```
adamjstewart:
```suggestion
    depends_on("py-numpy@1.8:", type=("build", "run"))
```
They use >, not >=

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: True, but setup.cfg is wrong and will be fixed in 0.4.1. Is there some sort of metadata checking that will cause a problem if this change is not made for 0.4.0?

var/spack/repos/builtin/packages/py-gmxapi/package.py
eirrgang: Ah. Good catch.

var/spack/repos/builtin/packages/py-gmxapi/package.py
adamjstewart: At the moment, nothing automatic. I've been thinking about adding the `--check-build-dependencies` flag to `pip install` which might complain, and running `pip check` as a post-install test, which will definitely complain, but haven't added those yet. I usually err on the side of following the documented deps, even if they are wrong. But it's up to you.

eirrgang: `pip check` would break this if done in a different environment than the (non-isolated) build environment.

I vote for caution.

adamjstewart (approved):

PR #35740, opened 2023-02-28
Fix bufr package typo (libufr->libbufr)

AlexanderRichert-NOAA:
This package fixes a typo in the bufr package which causes problems with `spack load` (or presumably downstream dependencies).


edwardhartnett (approved):

tldahlgren (approved):
Great catch.

PR #35743, opened 2023-02-28
Add py-mpi4py as a dependency for henson

mrzv:
Adds a missing dependency for henson, when `+python` variant is chosen.

### DIFF
--- a/var/spack/repos/builtin/packages/henson/package.py
+++ b/var/spack/repos/builtin/packages/henson/package.py
@@ -14,6 +14,8 @@ class Henson(CMakePackage):
 
     version("master", branch="master")
 
+    maintainers("mrzv")
+
     depends_on("mpi")
 
     variant("python", default=False, description="Build Python bindings")
--- a/var/spack/repos/builtin/packages/henson/package.py
+++ b/var/spack/repos/builtin/packages/henson/package.py
@@ -20,7 +20,7 @@ class Henson(CMakePackage):
 
     variant("python", default=False, description="Build Python bindings")
     extends("python", when="+python")
-    depends_on("py-mpi4py", when="+python")
+    depends_on("py-mpi4py", when="+python", type=("build", "run"))
     variant("mpi-wrappers", default=False, description="Build MPI wrappers (PMPI)")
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/henson/package.py
```diff
@@ -14,10 +14,13 @@ class Henson(CMakePackage):
 
     version("master", branch="master")
 
+    maintainers("mrzv")
+
     depends_on("mpi")
 
     variant("python", default=False, description="Build Python bindings")
     extends("python", when="+python")
+    depends_on("py-mpi4py", when="+python")
```
adamjstewart: What dependency type should this be? I assume it's needed at build/run-time, not sure about link-time.

mrzv: That's an interesting question. As far as I know it's only build and run-time, there is nothing that's being linked from mpi4py.

adamjstewart: You can check the build log and see. But if it's build/run only, you can use:
```suggestion
    depends_on("py-mpi4py", when="+python", type=("build", "run"))
```

mrzv: What does that buy? I.e., why does it ever make sense to have build without link?

mrzv: (In any case, I've added this change.)

adamjstewart: These types have the following meaning:

* build: add to PATH/PYTHONPATH during `spack install`
* run: add to PATH/PYTHONPATH during `spack load` or `module load` or `spack env activate`
* link: add to link/include commands and inject RPATHs via compiler wrappers
* test: only build when `spack install --test`

adamjstewart (approved):

var/spack/repos/builtin/packages/henson/package.py
mrzv: Ah, makes perfect sense. Thanks for the explanation.

PR #35757, opened 2023-03-01
add nfft 3.2.4 (still the required version for octopus)

glaweh:
Octupus requires an older version <=3.2.4 of nfft (when compiled with nfft support).
In this branch, we add that version to the spack nfft package.


alalazo:
@spackbot fix style

alalazo (approved):

PR #35769, opened 2023-03-01
gdb: version 13.1 and debuginfod support

robertu94:


robertu94:
I've rebuilt all gdb with and without debuginfod for all effected versions.

var/spack/repos/builtin/packages/gdb/package.py
```diff
@@ -108,6 +111,9 @@ def configure_args(self):
         if "+tui" in self.spec:
             args.append("--enable-tui")
 
+        if "+debuginfod" in self.spec:
+            args.append("--with-debuginfod")
```
tldahlgren: FYI.  You could be using helper functions for these (https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#with-or-without).

tldahlgren (dismissed):
Confirmed the sha256.  Let me know if you want to leave this PR as-is or take advantage of the helper function(s).

var/spack/repos/builtin/packages/gdb/package.py
robertu94: I adopted them in most cases, but left it for python and gmp because these dependencies require more complex flags

robertu94:
@tldahlgren ping.  This one is done too.

var/spack/repos/builtin/packages/gdb/package.py
```diff
@@ -80,11 +82,20 @@ class Gdb(AutotoolsPackage, GNUMirrorPackage):
...
+    depends_on("elfutils@0.178:+debuginfod", when="@11.1:+debuginfod")
 
     build_directory = "spack-build"
 
     def configure_args(self):
-        args = ["--with-system-gdbinit={0}".format(self.prefix.etc.gdbinit)]
+        args = [
+            "--with-system-gdbinit={0}".format(self.prefix.etc.gdbinit),
+            *self.enable_or_disable("lto"),
+            *self.enable_or_disable("quad"),
```
tldahlgren: Should this be using `with_or_without()` for `quad`?

tldahlgren: I can build successfully but am not sure why the `*self...` syntax works.

alalazo: > I can build successfully but am not sure why the *self... syntax works.

`enable_or_disable` returns a list, so this is the same as:
```python
>>> l = [2, 3]
>>> [1, *l, 4]
[1, 2, 3, 4]

```

robertu94: Good catch for `with_or_without('quad')`.  Should be pushed now @tldahlgren

tldahlgren: > > I can build successfully but am not sure why the *self... syntax works.
> 
> `enable_or_disable` returns a list, so this is the same as:
> 
> ```python
> >>> l = [2, 3]
> >>> [1, *l, 4]
> [1, 2, 3, 4]
> ```

I must've done something wrong when I tried that with a list of strings and with a local method that returned a list of strings.  Both reported syntax errors.

tldahlgren (approved):

PR #35773, opened 2023-03-01
e3sm-scorpio: add e3sm-scorpio package

downloadico:
This is the Scorpio package from the e3sm.org site.

### DIFF
--- a/var/spack/repos/builtin/packages/e3sm-scorpio/package.py
+++ b/var/spack/repos/builtin/packages/e3sm-scorpio/package.py
@@ -1,0 +1,53 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class E3smScorpio(CMakePackage):
+    """The SCORPIO (Software for Caching Output and Reads for Parallel I/O) library is used by all the model components in DOE's Energy Exascale Earth System Model (E3SM) for reading input data and writing model output"""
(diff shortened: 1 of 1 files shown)


var/spack/repos/builtin/packages/e3sm-scorpio/package.py
```diff
@@ -0,0 +1,55 @@
...
+
+    homepage = "https://e3sm.org/scorpio-parallel-io-library/"
+    url = "https://github.com/E3SM-Project/scorpio/archive/refs/tags/scorpio-v1.4.1.tar.gz"
+
+    version("1.4.1", sha256="7cb4589410080d7e547ef17ddabe68f749e6af019c1d0e6ee9f11554f3ff6b1a")
+
+    variant("timing", default="False", description="Enable timing")
+    variant("mpi", default="True", description="Enable MPI")
+
+    depends_on("cmake", type="build")
```
tldahlgren, line 22: Is there a reason to not just rely on the `cmake` dependency inherited from `CMakePackage`?

tldahlgren (approved):
Confirmed the sha256 and ability to concretize the spec.  Just a minor question about the `CMake` dependency.

tldahlgren:
Thank you for your contribution.  For future reference, it would simplify the review process if a PR-specific branch is used.  See https://spack.readthedocs.io/en/latest/contribution_guide.html#branching for the process.

var/spack/repos/builtin/packages/e3sm-scorpio/package.py
downloadico: no!  OOPS!

downloadico:
https://github.com/spack/spack/pull/35794

PR #35780, opened 2023-03-02
py-rtoml: add new package

adamjstewart:
This PR adds Yet Another Build Backend (YABB): maturin

Maturin is kinda like setuptools-rust, which is kinda like setuptools, which is kinda like distutils. Not really sure what the difference is, but rtoml chose it, so we now support it.

Didn't bother adding it to the PythonPackage docs since dependency discovery is identical to all the rest and we only have one package so far that uses it. 

This is a rust/cargo package, so it downloads hundreds of dependencies during build time. We currently don't have better control of rust packages.


tldahlgren (approved):
Confirmed the sha256, rest LGTM.

PR #35786, opened 2023-03-02
py-nbstripout: add new package

adamjstewart:


tldahlgren (approved):
Confirmed sha256.

PR #32306, opened 2022-08-22
view_copy: relocate symlinks

haampie:
fixes #33606
fixes #19550

Previously `spack:view:default:link_type:copy` did not handle symlinks:
it changed symlinks into redundant file copies, errored on dangling
links, and failed at symlinked directories.

With this commit it actually creates symlinks, and relocates them if
necessary. However, it's far from perfect (doesn't properly check if a
symlink points into a prefix dir in multiple ways: relative symlinks
aren't checked whether they escape the prefix, and absolute symlinks are
not checked in realpath sense). But held to the standard of
Spack's current relocation logic nobody can complain...

Supersedes #31958

PR #23476, opened 2021-05-06
env views: make view updates atomic

becker33:
Currently, environment views blink out of existence during the view regeneration, and are slowly built back up to their new and improved state. This is not good if other processes attempt to access the view -- they can see it in an inconsistent state.

This PR fixes makes environment view updates atomic. This requires a level of indirection (via symlink, similar to nix or guix) from the view root to the underlying implementation on the filesystem. 

Now, an environment view at `/path/to/foo` is a symlink to `/path/to/._foo/<hash>`, where `<hash>` is a hash of the contents of the view.  We construct the view in its content-keyed hash directory, create a new symlink to this directory, and atomically replace the symlink with one to the new view.

This PR also future-proofs environment views so that we can implement rollback.

For background:
* there is no atomic operation in posix that allows for a non-empty directory to be replaced.
* There is an atomic `renameat2` in the linux kernel starting in version 3.15, but many filesystems don't support the system call, including NFS3 and NFS4, which makes it a poor implementation choice for an HPC tool, so we use the symlink approach that others tools like nix and guix have used successfully.

PR #35804, opened 2023-03-03
env views: use renameat2 for atomic view updates.

becker33:
This PR creates a new option for environment views, `update_method`, with values `symlink|exchange|auto`

`symlink` is the option to keep the previous behavior. We use an indirection layer of symlinks to do atomic updates of environment views. This worked, but led to corner cases in which users found paths to the underlying view implementation (you've seen this if you've ever noted a directory like `._root/<hash>`), which were not stable across updates to the environment.

`exchange` is the option to use the new behavior. `renameat2` is a new syscall in `linux@3.15:` that supports atomic swaps of non-empty directories, which was previously impossible. `renameat2` is available in `libc@2.28:`, which coincides with RHEL8 for a sense of timeframe. With `renameat2`, we create the view in an alternate location and atomically swap it into place.

`auto` is the option to use `exchange` if feasible. That means that if your view already exists, it will use the method the view was created with (assuming it is supported by the OS) and if the view does not already exist, it will use the `exchange` method if possible and fall back on the `symlink` method. This is the default.

For all methods, if the view already exists and Spack is configured to use a different method than the view was created with or the OS does not support the method the view was created with, you will receive an error with a prompt to either change your view configuration, run on a different OS, or run `spack env view regenerate --force` which will do a non-atomic update.

### DIFF
--- a/lib/spack/external/archspec/json/cpu/microarchitectures.json
+++ b/lib/spack/external/archspec/json/cpu/microarchitectures.json
@@ -2706,7 +2706,7 @@
         ],
         "clang" : [
           {
-            "versions": "9.0:12.99",
+            "versions": "9.0:12.0",
             "flags" : "-march=armv8.4-a"
           },
           {
--- a/lib/spack/external/archspec/json/cpu/microarchitectures.json
+++ b/lib/spack/external/archspec/json/cpu/microarchitectures.json
@@ -2706,7 +2706,7 @@
         ],
         "clang" : [
           {
-            "versions": "9.0:12.0",
+            "versions": "9.0:12.99",
             "flags" : "-march=armv8.4-a"
           },
           {
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -496,7 +496,7 @@ class ViewDescriptor(object):
         if not os.path.exists(self.root):
             return None
 
-        view = self.view(created_path=False)
+        view = self.view()
         orig_path = view.metadata.get("created_path", None)
         if orig_path:
             return orig_path
@@ -539,7 +539,7 @@ class ViewDescriptor(object):
         view = self.view()
         return view.get_projection_for_spec(spec)
 
-    def view(self, new=None, created_path=True):
+    def view(self, new=None, created_path=False):
         """
         Generate the FilesystemView object for this ViewDescriptor
 
@@ -552,6 +552,10 @@ class ViewDescriptor(object):
             new (str or None): If a string, create a FilesystemView
                 rooted at that path. Default None. This should only be used to
                 regenerate the view, and cannot be used to access specs.
+            created_path (bool): Pass metadata on the path the view was created in to the
+                underlying view implementation. Default False. This option should be enabled
+                when creating a view to add packages to, but not when creating a view object
+                to query an existing view.
         """
         root = new if new else self.root
 
@@ -710,7 +714,7 @@ class ViewDescriptor(object):
         if specs:
             tty.msg("Updating view at {0}".format(self.root))
 
-        view = self.view(new=new_root)
+        view = self.view(new=new_root, created_path=True)
 
         # Create a new view
         try:
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -479,7 +479,7 @@ class ViewDescriptor(object):
     @property
     def _current_root(self):
         """
-        Return the directory in which the view is currently implemented.
+        Return the directory in which the view has been constructed.
 
         If the view is using renameat2 for atomic updates, self.root is a directory and the root
         of the underlying implementation is the same as self.root.
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -39,6 +39,7 @@ import spack.store
 import spack.subprocess_context
 import spack.traverse
 import spack.user_environment as uenv
+import spack.util.atomic_update
 import spack.util.cpus
 import spack.util.environment
 import spack.util.hash
@@ -469,6 +470,9 @@ class ViewDescriptor(object):
 
     @property
     def _current_root(self):
+        if spack.util.atomic_update.use_renameat2():
+            return self.root
+
         if not os.path.islink(self.root):
             return None
 
@@ -486,6 +490,9 @@ class ViewDescriptor(object):
         return os.path.join(root_dir, "._%s" % root_name, content_hash)
 
     def content_hash(self, specs):
+        print("CONTENT_HASH")
+        print("    ", specs)
+        print("    ", self.to_dict())
         d = syaml.syaml_dict(
             [
                 ("descriptor", self.to_dict()),
@@ -493,6 +500,7 @@ class ViewDescriptor(object):
             ]
         )
         contents = sjson.dump(d)
+        print("    ", spack.util.hash.b32_hash(contents))
         return spack.util.hash.b32_hash(contents)
 
     def get_projection_for_spec(self, spec):
@@ -599,6 +607,10 @@ class ViewDescriptor(object):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        if spack.util.atomic_update.use_renameat2():
+            if os.path.isdir(new_root):
+                shutil.rmtree(new_root)
+
         _error_on_nonempty_view_dir(new_root)
 
         # construct view at new_root
@@ -607,26 +619,15 @@ class ViewDescriptor(object):
 
         view = self.view(new=new_root)
 
-        root_dirname = os.path.dirname(self.root)
-        tmp_symlink_name = os.path.join(root_dirname, "._view_link")
-
         # Create a new view
         try:
             fs.mkdirp(new_root)
             view.add_specs(*specs, with_dependencies=False)
-
-            # create symlink from tmp_symlink_name to new_root
-            if os.path.exists(tmp_symlink_name):
-                os.unlink(tmp_symlink_name)
-            symlink(new_root, tmp_symlink_name)
-
-            # mv symlink atomically over root symlink to old_root
-            fs.rename(tmp_symlink_name, self.root)
+            spack.util.atomic_update.atomic_update(new_root, self.root)
         except Exception as e:
             # Clean up new view and temporary symlink on any failure.
             try:
                 shutil.rmtree(new_root, ignore_errors=True)
-                os.unlink(tmp_symlink_name)
             except (IOError, OSError):
                 pass
 
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -481,21 +481,28 @@ class ViewDescriptor(object):
         """
         Return the directory in which the view has been constructed.
 
-        If the view is using renameat2 for atomic updates, self.root is a directory and the root
-        directory of the view is the same as self.root.
+        Query the view if it stores metadata on where it was constructed.
 
         If the view us using symlinks for atomic updates, self.root is a link and we read the link
         to find the real root directory.
 
-        If self.root does not exist or is a regular file, the view has not been
-        constructed on the filesystem.
+        If self.root is not a view with metadata and is not a link, the view has not been
+        constructed.
         """
-        if not os.path.islink(self.root):
-            if os.path.isdir(self.root):
-                return self.root
-            else:
-                return None
+        # Get the view as self.root even if it is actually a symlink
+        # We will not operate on this view object, only query metadata
+        # We don't want to pass a created_path to this view, so that we can read where it says it
+        # was created.
+        view = self.view(self.root, created_path=False)
+        orig_path = view.metadata.get("created_path", None)
+        if orig_path:
+            return orig_path
 
+        # Backwards compat only applies for symlinked views
+        if not os.path.islink(self.root):
+            return None
+
+        # For backards compat, check link for symlink views if no "created_path"
         root = os.readlink(self.root)
         if os.path.isabs(root):
             return root
@@ -531,7 +538,7 @@ class ViewDescriptor(object):
         rel_path = os.path.relpath(view_path, self._current_root)
         return os.path.join(self.root, rel_path)
 
-    def view(self, new=None):
+    def view(self, new=None, created_path=True):
         """
         Generate the FilesystemView object for this ViewDescriptor
 
@@ -553,14 +560,17 @@ class ViewDescriptor(object):
                 "View root is at %s" % self.root
             )
             raise SpackEnvironmentViewError(msg)
-        return SimpleFilesystemView(
-            root,
-            spack.store.layout,
-            ignore_conflicts=True,
-            projections=self.projections,
-            link=self.link_type,
-            final_destination=self.root,
-        )
+
+        kwargs = {
+            "ignore_conflicts": True,
+            "projections": self.projections,
+            "link": self.link_type,
+            "final_destination": self.root,
+        }
+        if created_path:
+            kwargs["metadata"] = {"created_path": root}
+
+        return SimpleFilesystemView(root, spack.store.layout, **kwargs)
 
     def __contains__(self, spec):
         """Is the spec described by the view descriptor
@@ -677,9 +687,6 @@ class ViewDescriptor(object):
         # will be /dirname/._basename_<hash>.
         # This allows for atomic swaps when we update the view
 
-        # Check which atomic update method we need
-        update_method = self.update_method_to_use(force)
-
         # cache the roots because the way we determine which is which does
         # not work while we are updating
         new_root = self._next_root(specs)
@@ -689,21 +696,10 @@ class ViewDescriptor(object):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
-        print(new_root)
-        #        print(
-        #            [
-        #                (s, os.stat(os.path.join(os.path.dirname(new_root), s)).st_mtime)
-        #                for s in os.listdir(os.path.dirname(new_root))
-        #            ]
-        #        )
-        print(specs)
+        # Check which atomic update method we need
+        update_method = self.update_method_to_use(force)
+
         if update_method == "exchange" and os.path.isdir(new_root):
-            # If new_root is the newest thing in its directory, no need to update
-            parent = os.path.dirname(new_root)
-            siblings = [os.path.join(parent, s) for s in os.listdir(parent)]
-            if max(siblings, key=lambda p: os.stat(p).st_mtime) == new_root:
-                tty.debug("View at %s does not need regeneration." % self.root)
-                return
             shutil.rmtree(new_root)
 
         _error_on_nonempty_view_dir(new_root)
@@ -719,7 +715,14 @@ class ViewDescriptor(object):
             fs.mkdirp(new_root)
             view.add_specs(*specs, with_dependencies=False)
             if update_method == "exchange":
-                spack.util.atomic_update.atomic_update_renameat2(new_root, self.root)
+                # Swap the view to the directory of the previous view if one exists so that
+                # the view that is swapped out will be named appropriately
+                if old_root:
+                    os.rename(new_root, old_root)
+                    exchange_location = old_root
+                else:
+                    exchange_location = new_root
+                spack.util.atomic_update.atomic_update_renameat2(exchange_location, self.root)
             else:
                 spack.util.atomic_update.atomic_update_symlink(new_root, self.root)
         except Exception as e:
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -490,9 +490,6 @@ class ViewDescriptor(object):
         return os.path.join(root_dir, "._%s" % root_name, content_hash)
 
     def content_hash(self, specs):
-        print("CONTENT_HASH")
-        print("    ", specs)
-        print("    ", self.to_dict())
         d = syaml.syaml_dict(
             [
                 ("descriptor", self.to_dict()),
@@ -500,7 +497,6 @@ class ViewDescriptor(object):
             ]
         )
         contents = sjson.dump(d)
-        print("    ", spack.util.hash.b32_hash(contents))
         return spack.util.hash.b32_hash(contents)
 
     def get_projection_for_spec(self, spec):
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -689,6 +689,9 @@ class ViewDescriptor(object):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        print(new_root)
+        print([(s, os.stat(os.path.join(os.path.dirname(new_root), s)).st_mtime) for s in os.listdir(os.path.dirname(new_root))])
+        print(specs)
         if update_method == "exchange" and os.path.isdir(new_root):
             # If new_root is the newest thing in its directory, no need to update
             parent = os.path.dirname(new_root)
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -628,11 +628,8 @@ class ViewDescriptor(object):
     def raise_if_exchange_before_symlink(self, force):
         if os.path.isdir(self.root) and not os.path.islink(self.root):
             if force:
-                try:
-                    shutil.rmtree(self.root)
-                    return
-                except:
-                    pass
+                shutil.rmtree(self.root, ignore_errors=True)
+                return
             msg = f"The view at {self.root} cannot be updated with the 'symlink' update method"
             msg += " because it was originally constructed with the 'exchange' method."
             msg += "\n  Either change the update method to 'exchange' or"
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -690,12 +690,12 @@ class ViewDescriptor(object):
             return
 
         print(new_root)
-#        print(
-#            [
-#                (s, os.stat(os.path.join(os.path.dirname(new_root), s)).st_mtime)
-#                for s in os.listdir(os.path.dirname(new_root))
-#            ]
-#        )
+        #        print(
+        #            [
+        #                (s, os.stat(os.path.join(os.path.dirname(new_root), s)).st_mtime)
+        #                for s in os.listdir(os.path.dirname(new_root))
+        #            ]
+        #        )
         print(specs)
         if update_method == "exchange" and os.path.isdir(new_root):
             # If new_root is the newest thing in its directory, no need to update
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -470,6 +470,18 @@ class ViewDescriptor(object):
 
     @property
     def _current_root(self):
+        """
+        Return the directory in which the view is currently implemented.
+
+        If the view is using renameat2 for atomic updates, self.root is a directory and the root
+        of the underlying implementation is the same as self.root.
+
+        If the view us using symlinks for atomic updates, self.root is a link and we read the link
+        to find the root of the underlying implementation of the view.
+
+        If self.root does not exist or is a regular file, there is no current implementation of the
+        view on the filesystem.
+        """
         if not os.path.islink(self.root):
             if os.path.isdir(self.root):
                 return self.root
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -690,12 +690,12 @@ class ViewDescriptor(object):
             return
 
         print(new_root)
-        print(
-            [
-                (s, os.stat(os.path.join(os.path.dirname(new_root), s)).st_mtime)
-                for s in os.listdir(os.path.dirname(new_root))
-            ]
-        )
+#        print(
+#            [
+#                (s, os.stat(os.path.join(os.path.dirname(new_root), s)).st_mtime)
+#                for s in os.listdir(os.path.dirname(new_root))
+#            ]
+#        )
         print(specs)
         if update_method == "exchange" and os.path.isdir(new_root):
             # If new_root is the newest thing in its directory, no need to update
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -470,11 +470,11 @@ class ViewDescriptor(object):
 
     @property
     def _current_root(self):
-        if spack.util.atomic_update.use_renameat2():
-            return self.root
-
         if not os.path.islink(self.root):
-            return None
+            if os.path.isdir(self.root):
+                return self.root
+            else:
+                return None
 
         root = os.readlink(self.root)
         if os.path.isabs(root):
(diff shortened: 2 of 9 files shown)


becker33:
@spackbot fix style

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+from contextlib import contextmanager
+from typing import Optional
+
+import llnl.util.filesystem as fs
+from llnl.util.symlink import symlink
+
+libc: Optional[ctypes.CDLL] = None
+try:
+    libc = ctypes.CDLL("/lib64/libc.so.6", 0x04)  # 0x04 is RTLD_NOLOAD
```
haampie: Uh, don't open by path. Just do CDLL(None) and you get a handle for the Python process which links against libc.

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    if use_renameat2():
+        return atomic_update_renameat2(oldpath, newpath)
+    else:
+        return atomic_update_symlink(oldpath, newpath)
+
+
+@contextmanager
+def open_safely(path):
+    fd = os.open(path, os.O_CLOEXEC | os.O_PATH)
```
haampie, line 40: > When O_PATH is specified in flags, flag bits other than O_CLOEXEC, O_DIRECTORY, and O_NOFOLLOW are ignored.

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,79 @@
...
+
+def atomic_update_renameat2(src, dest):
+    dest_exists = os.path.exists(dest)
+    if not dest_exists:
+        fs.touch(dest)
+    with open_safely(src) as srcfd:
+        with open_safely(dest) as destfd:
+            try:
+                libc.renameat2(
+                    srcfd, src.encode(), destfd, dest.encode(), 2
```
haampie: srcfd and destfd are ignored if you pass absolute paths, which is probably why it worked for you. The file descriptors should be those of the directories the paths are supposed to be relative to. Can you drop that `open_safely` business then and use use AT_FDCWD?

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,79 @@
...
+def open_safely(path):
+    fd = os.open(path, os.O_CLOEXEC | os.O_PATH)
+    try:
+        yield fd
+    finally:
+        os.close(fd)
+
+
+def atomic_update_renameat2(src, dest):
+    dest_exists = os.path.exists(dest)
```
haampie: lexists (may also need to drop trailing dir separators, otherwise non-existent symbolic links are still followed)

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,79 @@
...
+    if not dest_exists:
+        fs.touch(dest)
+    with open_safely(src) as srcfd:
+        with open_safely(dest) as destfd:
+            try:
+                libc.renameat2(
```
haampie: check return value

haampie:
General question: should you care about the input path? If dest is a symlink to a dir and it's replaced with a dir, you have to be careful not to pass `/path/to/dest/` instead of `/path/to/dest`.

lib/spack/spack/util/atomic_update.py
becker33: Did not realize I could use `None`, I was switching it to `"libc.so.6"` instead of the full path but I guess this is cleaner for some far-off future that libc changes.

I think this will also fix the code coverage issue, since the reason I was switching to not use a path was that libc lives in a different location on ubuntu and that's where our CI runs.

becker33:
@haampie I think I've addressed all review comments.

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,78 @@
...
+RENAME_EXCHANGE = 2
+AT_FDCWD = -100
+
+libc: Optional[ctypes.CDLL] = None
+try:
+    # CDLL(None) returns the python process
+    # python links against libc, so we can treat this as a libc handle
+    # we could also use CDLL("libc.so.6") but this is (irrelevantly) more future proof
+    libc = ctypes.CDLL(None)
+except BaseException:
```
tgamblin: This raises `OSError` on failure -- I think it's best to catch that here instead of `BaseException`, as catching `BaseException` can be very hard to debug if the error *happens* to be some other random thing.

Minor thought: using `"libc.so.6"` instead of `None` would make this fail earlier for platforms other than Linux, which means that you don't have to think about them later (e.g., for `use_renameat2()`).  But different distros put `libc` in different places (`/lib`, `/lib64`) so `None` is probably better from that perspective.

I don't think it's likely that other OS's would grow a `renameat2` that doesn't comply with the Linux one, so I think on balance I like `None` better.

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,78 @@
...
+    pass
+
+
+def use_renameat2():
+    return hasattr(libc, "renameat2")
```
tgamblin: I think this belongs in `environment.py`, because it's specific to the implementation of views, while atomic update is just providing ways to do atomic update.  can this be `renameat2_available`, and then something in `environment` specify whether it's actually used?

lib/spack/spack/environment/environment.py
```diff
@@ -467,6 +468,9 @@ def from_dict(base_path, d):
 
     @property
     def _current_root(self):
```
tgamblin, line 480: Can this have a docstring?  Reading this I don't know what it's *supposed* to return.  It seems like it returns `None` for non-renameat views when there isn't  a symlink... but that's going to confuse things with the new implementation. 

Consider:
1. user does a directory view with `renameat2`
2. switches to symlinks
3. this will now say there is no current root, but that's probably wrong as there is a whole directory there.

And, yes, we don't have a config option for this, but:
* We could, and we might want to
* People check out different versions of Spack all the time.
* You might have a view exposed to different machines, only one of which supports `renameat2`

I think for that last reason, this needs to be a config option for views that gets stored *in* the view directory -- i.e. a view identifies itself as a `renameat2` view and says it needs to be updated a certain way.

lib/spack/spack/environment/environment.py
```diff
@@ -467,6 +468,9 @@ def from_dict(base_path, d):
...
+        if spack.util.atomic_update.use_renameat2():
+            return self.root
```
tgamblin: See above -- I think this needs to be more careful about what *might* be there.  There aren't just two cases (symlink and renameat2), there are also cases where I used one then the other.

lib/spack/spack/environment/environment.py
```diff
@@ -467,6 +468,9 @@ def from_dict(base_path, d):
 
     @property
     def _current_root(self):
+        if spack.util.atomic_update.use_renameat2():
```
tgamblin: I think whether we *use* it should be made into a module-scope var or function *here*, and that should ask `atomic_update` whether it's actually available.

lib/spack/spack/environment/environment.py
```diff
@@ -597,6 +601,10 @@ def regenerate(self, concretized_root_specs):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        if spack.util.atomic_update.use_renameat2():
+            if os.path.isdir(new_root):
+                shutil.rmtree(new_root)
```
tgamblin: what if it's a link?

tgamblin (changes requested):
Ok I put some comments in here -- Thinking about how this is going to work in practice makes me think it really needs to be an option on the view.

The update logic here doesn't think too hard about cases where you have a view that *used* to be a symlink... and I think that's going to cause users some grief.

I also think *old* spack is going to be very confused by the new views -- and we should be clear that the two are incompatible and make it easy to understand how to transition.

The easy way to do that for now is to make this an experimental config option -- I think it's likely to only be used at our site in the near future.  But really the top-level dir linking mechanism is part of the view and affects compatibility... so I think at the very least we need to make this opt-in, and probably also start identifying views as using this scheme or not.

lib/spack/external/archspec/json/cpu/microarchitectures.json
```diff
@@ -2706,7 +2706,7 @@
         ],
         "clang" : [
           {
-            "versions": "9.0:12.0",
+            "versions": "9.0:12.99",
```
tgamblin: Is this intentional?  I think this should be a separate change?

lib/spack/spack/environment/environment.py
tgamblin: this still needs a docstring

lib/spack/spack/util/atomic_update.py
tgamblin: any thoughts on this comment?

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,76 @@
...
+libc: Optional[ctypes.CDLL] = None
+try:
+    # CDLL(None) returns the python process
+    # python links against libc, so we can treat this as a libc handle
+    # we could also use CDLL("libc.so.6") but this is (irrelevantly) more future proof
+    libc = ctypes.CDLL(None)
+except BaseException:
+    pass
+
+renameat2 = getattr(libc, "renameat2", None)
```
tgamblin: The stuff up here should be encapsulated in a method so that it only happens when needed, not on import.

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def atomic_update(oldpath, newpath):
+    """
+    atomically update newpath to contain the information at oldpath
```
tgamblin:
```suggestion
def atomic_update(source, dest):
    """
    Atomically move or swap source into the location of dest.
```

lib/spack/external/archspec/json/cpu/microarchitectures.json
becker33: I think this is related to the bug that @haampie was discussing reverting a PR for about version ranges with upper bounds. I can revert this and if it fails in CI we can look at fixing the other bug.

lib/spack/spack/util/atomic_update.py
becker33: I thought I had changed it to `OSError`, definitely agree with changing it.

lib/spack/spack/environment/environment.py
```diff
@@ -467,8 +468,23 @@ def from_dict(base_path, d):
 
     @property
     def _current_root(self):
+        """
+        Return the directory in which the view is currently implemented.
```
tgamblin:
```suggestion
        Return the directory in which the view has been constructed.
```

lib/spack/spack/environment/environment.py
```diff
@@ -467,8 +468,23 @@ def from_dict(base_path, d):
...
+
+        If the view is using renameat2 for atomic updates, self.root is a directory and the root
+        of the underlying implementation is the same as self.root.
```
tgamblin:
```suggestion
        directory of the view is the same as self.root.
```

lib/spack/spack/environment/environment.py
```diff
@@ -467,8 +468,23 @@ def from_dict(base_path, d):
...
+
+        If the view us using symlinks for atomic updates, self.root is a link and we read the link
+        to find the root of the underlying implementation of the view.
```
tgamblin:
```suggestion
        to find the real root directory.
```

lib/spack/spack/environment/environment.py
```diff
@@ -467,8 +468,23 @@ def from_dict(base_path, d):
...
+
+        If self.root does not exist or is a regular file, there is no current implementation of the
+        view on the filesystem.
```
tgamblin:
```suggestion
        If self.root does not exist or is a regular file, the view has not been
        constructed on the filesystem.
```

lib/spack/spack/environment/environment.py
```diff
@@ -576,6 +593,15 @@ def specs_for_view(self, concretized_root_specs):
...
         return specs
 
+    def use_renameat2(self):
+        if os.path.islink(self.root):
+            return False
+        elif os.path.isdir(self.root):
+            if not spack.util.atomic_update.renameat2:
+                raise Exception
+
+        return bool(spack.util.atomic_update.renameat2)
```
tgamblin: I am pretty convinced that this still needs to be a config option, not an automatic decision.  Reason:

1. There are sites that share a filesystem across OS's.  They may want to ensure that views are usable across OS's with older kernels *and* OS's with newer kernels.
2. People may actually want the symlink for whatever reason.
3. This feature isn't well tested outside LLNL, and I would rather introduce it in a way that doesn't risk breaking everyone's view until we've had time to use it in production.  Putting it behind a feature flag (config option) will do that.

So we should add something like `update_method` here:

```yaml
view:
    root: /path/to/root
    update_method: [symlink | renameat2]
    link: [roots|all|run]
    # etc...    
```

And I think the default `update_method` should stay `symlink` until we can play with this more.  With the config option I'm way less worried about some unexpected consequence of this for environments that are *not* LC's (i.e., everyone else's)

Ideally, the code here would be smart enough to look at the config option and replace the view regardless of whether it was previously done with `renameat2` or a `symlink`, but if there is a config option I don't think we need to gate the merge on this, so long as we error out gracefully.

lib/spack/spack/environment/environment.py
```diff
@@ -588,6 +614,9 @@ def regenerate(self, concretized_root_specs):
         # will be /dirname/._basename_<hash>.
         # This allows for atomic swaps when we update the view
 
+        # Check which atomic update method we need
+        use_renameat2 = self.use_renameat2()
```
tgamblin: change to config option

lib/spack/spack/environment/environment.py
```diff
@@ -597,6 +626,10 @@ def regenerate(self, concretized_root_specs):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        if use_renameat2:
+            if os.path.isdir(new_root):
+                shutil.rmtree(new_root)
+
```
tgamblin: add some comments on this one to say what happens when we're configured to use symlinks but we find a `renameat2` directory and vice/versa.

lib/spack/spack/environment/environment.py
```diff
@@ -605,26 +638,18 @@ def regenerate(self, concretized_root_specs):
...
-            if os.path.exists(tmp_symlink_name):
-                os.unlink(tmp_symlink_name)
-            symlink(new_root, tmp_symlink_name)
-
-            # mv symlink atomically over root symlink to old_root
-            fs.rename(tmp_symlink_name, self.root)
+            if use_renameat2:
+                spack.util.atomic_update.atomic_update_renameat2(new_root, self.root)
+            else:
+                spack.util.atomic_update.atomic_update_symlink(new_root, self.root)
```
tgamblin: Why isn't this calling `atomic_update()` (which abstracts the check)?

lib/spack/spack/environment/environment.py
```diff
@@ -605,26 +638,18 @@ def regenerate(self, concretized_root_specs):
...
         except Exception as e:
             # Clean up new view and temporary symlink on any failure.
             try:
                 shutil.rmtree(new_root, ignore_errors=True)
-                os.unlink(tmp_symlink_name)
```
tgamblin, line 629: isn't there still a temporary symlink when not using `renameat2`?

tgamblin (changes requested):
One larger request (but not that large) and mostly LGTM.

lib/spack/spack/environment/environment.py
becker33: That is now handled as part of the `atomic_update_symlink` method

lib/spack/spack/environment/environment.py
becker33: Because whether the ViewDescriptor uses renameat2 is a separate concern from whether it is available, and `atomic_update()` keys on whether it's available.

blue42u:
Sorry to piggyback an RFE here, but I think it deeply relates to the changes in this PR.

I don't normally use Spack environments in parallel, so the non-atomic part of this seems less important to me. But I _have_ run into issues when the current `symlink` implementation changes the canonical path to some executables, it's a problem (and a serious annoyance) for e.g. `meson` which saves the canonical path to itself in the build directory.

It sounds like the `exchange` implementation would not have this issue, but on systems where this is unavailable I would much rather fall back to non-atomic view updates (with loud warnings) than to the broken `symlink` implementation. I would even be fine if automatic view updates were disabled when they are not atomic.

So now that this PR adds multiple update methods, my request is to extend `update_method` with extra options prioritizing consistent canonical paths for files in the view (bikeshedding requested):
 - `nonatomic`: Simple implementation that sacrifices atomicity for consistent canonical paths for files in the view, e.g. uses `shutil.rmtree` and `os.symlink` directly in the view target directory, and
 - `consistent-nonatomic-auto`: Auto-chosen implementation that maintains consistent canonical paths for files in the view and may sacrifice atomicity, AFAICT this is `exchange` with a fallback to `nonatomic`.
 - `consistent-auto`: Auto-chosen implementation that maintains both consistent canonical paths for files in the view and atomicity of updates, AFAICT this is just `exchange` right now.

Alternative solutions include removing the `auto` option and instead making `update_method` a preference list:
```yaml
  update_method: [exchange, symlink]  # or e.g. [exchange, nonatomic]
```
Or auto-choosing an implementation based on the features it supports:
```yaml
  update_method:
    atomic: true
    consistent_paths: false
```

Thoughts?

becker33:
@blue42u it sounds like your second suggestion of allowing a list of allowed methods is probably better than trying to create useable/rememberable/sensible names for all the combinations we could think up.

Would it entirely satisfy your use-case if we did the following?

1. All three existing options retain their current meanings
2. We add an additional option `nonatomic` which reverts to the behavior before the symlink method was introduced
3. We allow multiple options to be specified in a list, in which case it prefers the first but falls back on others

I think point (3) is really only relevant if `exchange` is your first choice, but I could certainly imagine that changing in the future, especially if we introduce some windows-only view feature.

Since all three points are compatible with the current code, I would probably want to approach that as a follow-on PR.

blue42u:
That should work just fine for my case.

My only point of contention (with point 1) is that `auto` doesn't make much sense if the list form is available, since (right now) it's equivalent to `[exchange, symlink]`. If `auto` is kept there are edge cases like:
```yaml
update_method: [exchange, auto, nonatomic]
```
that are tricky to reason about. (Of course `auto` could only be allowed as a single value, but IMHO explicitly writing out `[exchange, symlink]` leaves less room for surprised users.)

becker33:
If you have an environment view created with the `symlink` method on a `rhel8` system (so it supports the exchange method) and you have `update_method: [exchange, symlink]`, are you imagining that the behavior would be:

1. silently use the symlink option because that's what works with the view as previously created
2. use the symlink option but warn
3. fail and require the user to say which they want?

I'm assuming it would be (1), so I agree that makes `auto` == `[exchange, symlink]`

Interestingly, `[exchange, symlink, nonatomic`]` would maybe make sense as a default at that point...

I have on internal user who is already using this but he is only using the `exchange` method, so as long as `update_method: exchange` doesn't change meanings I don't have to make him change anything (and I don't think it would for any of your suggestions.

becker33:
So now I'm leaning towards modifying this PR so that `auto` is replaced by `[exchange, symlink]` and that is the default, and add `nonatomic` as an option in a future PR (with that PR to determine whether the new default should be `[exchange, symlink, nonatomic]` or keep the default as-is).

becker33:
@spackbot fix style

becker33:
@spackbot fix style

lib/spack/spack/environment/environment.py
```diff
@@ -578,7 +603,72 @@ def specs_for_view(self, concretized_root_specs):
...
+        """When methods are already determined invalid, construct error message for methods."""
+        msg = "View cannot be updated using specified update methods:"
+        if "exchange" in methods:
+            if not spack.util.atomic_update.renameat2():
+                msg += "\n  Operating system does not support 'exchange' atomic update method."
+                msg += f"\n    If the view {self.root} does not already exist on the filesystem,"
+                msg += " change its update_method to 'symlink' or 'auto'."
+                msg += f"\n    If the view at {self.root} exists already, either change the"
+                msg += " update_method and run `spack env view regenerate --force`"
+                msg += " or run on a newer OS."
```
haampie, line 632: Can this helpful hint be removed?

lib/spack/spack/environment/environment.py
```diff
@@ -578,7 +603,72 @@ def specs_for_view(self, concretized_root_specs):
...
+
+    def valid_update_method_symlink(self, force):
+        if os.path.exists(self.root):
+            if not os.path.islink(self.root):
+                if force:
+                    shutil.rmtree(self.root)
+                    return True
+                else:
+                    return False
+        return True
```
haampie, line 670: Can you improve readability a bit? It's not necessary to have 3 levels of nested control flow, it can be 1. Also can you add some comments what's being validated for posterity?

```suggestion
        # comment 1
        if not os.path.exists(self.root) or os.path.islink(self.root):
            return True

        # comment 2
        if force:
            shutil.rmtree(self.root)
            return True

       # comment 3
        return False
```

haampie: Same applies elsewhere (note: codepilot/chatgpt tooling can do this for you)

lib/spack/spack/environment/environment.py
```diff
@@ -599,6 +692,17 @@ def regenerate(self, concretized_root_specs):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        if update_method == "exchange":
+            if os.path.isdir(new_root):
```
haampie:
```suggestion
        if update_method == "exchange" and os.path.isdir(new_root):
```

lib/spack/spack/environment/environment.py
haampie: Additionally, isn't the first branch a very complicated way of expressing `lexist`?

becker33: No, it is not the same as `lexists`. For example, this condition is `False` if the path exists and is a directory, `lexists` returns `True` in that case

becker33: I think this:

```
    def valid_update_method_exchange(self, force):                                                                                                                                                                              
        if not spack.util.atomic_update.renameat2():                                                                                                                                                                            
            return False                                                                                                                                                                                                        
                                                                                                                                                                                                                                
        if not os.path.islink(self.root):                                                                                                                                                                                       
            return True                                                                                                                                                                                                         
                                                                                                                                                                                                                                
        if force:                                                                                                                                                                                                               
            os.unlink(self.root)                                                                                                                                                                                                
            return True                                                                                                                                                                                                         
        else:                                                                                                                                                                                                                   
            # Ensure we don't swap symlink -> exchange if we have symlink as an acceptable method.                                                                                                                              
            # This is to avoid problems switching between OSs.                                                                                                                                                                  
            return "symlink" not in self.update_method                                                                                                                                                                          
                                                                                                                                                                                                                                
        return True
```

is a lot less legible than

```
    def valid_update_method_exchange(self, force):                                                                                                                                                                              
        if not spack.util.atomic_update.renameat2():                                                                                                                                                                            
            return False                                                                                                                                                                                                        
                                                                                                                                                                                                                                
        # Ensure we don't swap symlink -> exchange if we have a symlink and symlink is an                                                                                                                                       
        # acceptable method. This is to avoid problems switching between OSs.                                                                                                                                                   
        if os.path.islink(self.root):                                                                                                                                                                                           
            if force:                                                                                                                                                                                                           
                os.unlink(self.root)                                                                                                                                                                                            
                return True                                                                                                                                                                                                     
            else:                                                                                                                                                                                                               
                return "symlink" not in self.update_method                                                                                                                                                                      
                                                                                                                                                                                                                                
        return True
```

despite the latter being a level further indented.

Removed the other condition as it was vestigial.

becker33: And this seems even clearer:

```
    def valid_update_method_exchange(self, force):                                                                                                                                                                              
        if not spack.util.atomic_update.renameat2():                                                                                                                                                                            
            return False                                                                                                                                                                                                        
                                                                                                                                                                                                                                
        # Ensure we don't swap symlink -> exchange if we have a symlink and symlink is an                                                                                                                                       
        # acceptable method. This is to avoid problems switching between OSs.                                                                                                                                                   
        if os.path.islink(self.root):                                                                                                                                                                                           
            if force:                                                                                                                                                                                                           
                os.unlink(self.root)                                                                                                                                                                                            
            elif "symlink" in self.update_method:                                                                                                                                                                               
                return False                                                                                                                                                                                                    
                                                                                                                                                                                                                                
        return True
```

tgamblin: I'd have gone with this one:

```python
    def valid_update_method_exchange(self, force):
        if not spack.util.atomic_update.renameat2():
            return False

        if not os.path.islink(self.root):                                                                                                                                                                                       
            return True

        if force:
            os.unlink(self.root)
            return True

        # Ensure we don't swap symlink -> exchange if we have symlink as an acceptable method.
        # This is to avoid problems switching between OSs.
        return "symlink" not in self.update_method
```

Since the last return is unreachable.  But I don't find @becker33's particularly unreadable.  I do tend to prefer eliminating possibilities without a lot of nesting.

lib/spack/spack/environment/environment.py
haampie: Can this "or run on a newer OS." comment still be removed? It's a weird suggestion

lib/spack/spack/environment/environment.py
haampie: I would still add clear comments what the purpose of the test is exactly, otherwise it may get turned into `lexist` in a future refactor.

lib/spack/spack/environment/environment.py
```diff
@@ -599,6 +689,16 @@ def regenerate(self, concretized_root_specs):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        if update_method == "exchange" and os.path.isdir(new_root):
+            # If new_root is the newest thing in its directory, no need to update
+            parent = os.path.dirname(new_root)
+            siblings = [os.path.join(parent, s) for s in os.listdir(parent)]
+            siblings.sort(reverse=True, key=lambda p: os.stat(p).st_mtime)
+            if siblings[0] == new_root:
+                tty.debug("View at %s does not need regenration." % self.root)
```
haampie:
```suggestion
                tty.debug("View at %s does not need regeneration." % self.root)
```

lib/spack/spack/environment/environment.py
```diff
@@ -599,6 +689,16 @@ def regenerate(self, concretized_root_specs):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        if update_method == "exchange" and os.path.isdir(new_root):
+            # If new_root is the newest thing in its directory, no need to update
+            parent = os.path.dirname(new_root)
+            siblings = [os.path.join(parent, s) for s in os.listdir(parent)]
+            siblings.sort(reverse=True, key=lambda p: os.stat(p).st_mtime)
+            if siblings[0] == new_root:
```
haampie:
```suggestion
            if max(siblings, key=...) == new_root:
```

Did you mean lstat?

lib/spack/spack/environment/environment.py
```diff
@@ -599,6 +689,16 @@ def regenerate(self, concretized_root_specs):
             tty.debug("View at %s does not need regeneration." % self.root)
             return
 
+        if update_method == "exchange" and os.path.isdir(new_root):
```
haampie, line 708: Just to double-check, do you need to avoid following symlinks here (both isdir and stat are used).

lib/spack/spack/environment/environment.py
haampie: (And I still think a flat structure is easier to follow 😬)

lib/spack/spack/environment/environment.py
haampie: I'm pretty sure you don't want to call `stat` on arbitrary files, it can throw FileNotFound

lib/spack/spack/environment/environment.py
```diff
@@ -578,7 +603,69 @@ def specs_for_view(self, concretized_root_specs):
 
         return specs
 
-    def regenerate(self, concretized_root_specs):
+    def update_method_error_msg(self, methods):
+        """When methods are already determined invalid, construct error message for methods."""
+        msg = "View cannot be updated using specified update methods:"
+        if "exchange" in methods:
+            if not spack.util.atomic_update.renameat2():
+                msg += "\n  Operating system does not support 'exchange' atomic update method."
```
haampie, line 627: Indentation is off here, should be 4 spaces.

Also "Operating system does not support" is technically incorrect, cause there can be other causes such as libc not providing the symbol you're trying to call. So can you just rephrase it to something generic like 'update method is not supported'?

haampie:
@becker33 can you check what the status is of `spack containerize` with this PR? I currently can't access a linux machine with docker. It has an explicit `COPY --from=previous_stage` for the `._view` symlink. (That is problematic on its own because all files are duplicated since docker follows the symlink, but well...). I expect that this errors with `update_method: exchange` as a default. The solution is to `COPY` the parent dir of the view instead.

lib/spack/spack/environment/environment.py
becker33: No, that's already been checked if you're using the "exchange" method

becker33:
@haampie good catch I didn't know this was used directly there. I'll have to look at how to fix the template.

haampie:
regarding the last commit, you do realize #32306 exists? 😬

becker33:
@spackbot fix style

becker33:
@spackbot fix style

lib/spack/spack/util/atomic_update.py
```diff
@@ -0,0 +1,75 @@
...
+    src = src.rstrip(os.path.sep)
+    dest = dest.rstrip(os.path.sep)
+
+    dest_exists = os.path.lexists(dest)
+    if not dest_exists:
+        fs.mkdirp(dest)
+    try:
+        rc = renameat2()(AT_FDCWD, src.encode(), AT_FDCWD, dest.encode(), RENAME_EXCHANGE)
+        if rc:
+            raise OSError(f"renameat2 failed to exchange {src} and {dest}")
```
tgamblin, line 56: this needs to preserve why the error happened -- via errno

tgamblin:
Ok: I have some sad news that maybe we should've investigated first.  It looks like `renameat2`, despite being in the RHEL8 kernel and libc, is not supported by Lustre or NFS.

* Here's the Lustre issue: https://jira.whamcloud.com/browse/LU-11557
* I can't find anything saying for sure for NFS, but at least on NFSv3 servers at LLNL, `renameat2` with `RENAME_EXCHANGE` fails with the same errno 22 (invalid argument).

So this definitely should *not* be default.  It should be an option for the currently limited set of places that support it.

I am tempted to say that we should go back to raw directories as the default in Spack, as the indirection of the symlink we use as the env root will break some things.  The most notable one, at least for us, is that many builds call `realpath()`, and things like CMake's `FindMPI` will end up pointing to the installation directory instead of the view, even if run in the view.

I don't know how valuable atomic update is for our users though.  We've only had this since #23476 -- I'm curious if anyone benefits from the atomic symlink update introduced there, or if using raw directories (which avoid realpath problems) would be a better default despite not being 100% atomic.

PR #35810, opened 2023-03-03
update SPERR package

shaomeng:
This PR updates the SPERR version.

### DIFF
--- a/var/spack/repos/builtin/packages/sperr/package.py
+++ b/var/spack/repos/builtin/packages/sperr/package.py
@@ -30,7 +30,6 @@ class Sperr(CMakePackage):
     variant("openmp", description="use openmp in 3D inputs", default=True)
     variant("utilities", description="build SPERR CLI utilities", default=True)
 
-
     def cmake_args(self):
         # ensure the compiler supports OpenMP if it is used
         if "+openmp" in self.spec:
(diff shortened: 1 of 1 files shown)


robertu94 (approved):
Confirmed that it works on my machine and with LibPressio.

robertu94 (approved):

robertu94:
I think something got goofed here

var/spack/repos/builtin/packages/sperr/package.py
```diff
@@ -10,21 +10,25 @@ class Sperr(CMakePackage):
...
     depends_on("pkgconfig", type=("build"), when="+zstd")
 
     variant("shared", description="build shared libaries", default=True)
-    variant("zstd", description="use Zstd for more compression", default=True)
-    variant("openmp", description="use openmp for acceleration", default=True)
-
-    maintainers("shaomeng", "robertu94")
+    variant("zstd", description="use zstd for more compression", default=True)
+    variant("openmp", description="use openmp in 3D inputs", default=True)
+    variant("utilities", description="build SPERR CLI utilities", default=True)
```
tldahlgren, line 31: The default before this change was for this variant to essentially be `False` but now it is `True`.  Just want to confirm that is intentional.

tldahlgren (approved):
Confirmed the branch exists and the sha256 are correct.  Just a quick question about the default value of the new variant.

tldahlgren:
I was able to successfully built the package at v0.6.2.

tldahlgren:
@robertu94 Is this PR ready to merge as is?

var/spack/repos/builtin/packages/sperr/package.py
shaomeng: Yes, that's an intentional change. Also,  this PR is ready to  merge.

robertu94:
Sorry for the delayed response, but yes it was ready 😄

PR #35816, opened 2023-03-03
CMake: return nothing faster for external CMake

scheibelp:
Add explicit CMake .libs implementation that returns an empty list; same for .headers

If CMake is treated as a link dependency (e.g. `depends_on(cmake)` without specifying deptypes will assume link+build by default), and it is external, then the library search can be extremely slow, because CMake in fact has no libraries and the search is fully recursive from the root (e.g. `/usr/`).


tgamblin (dismissed):

scheibelp:
@spackbot run pipeline

scheibelp:
@spackbot run pipeline

tgamblin (approved):

tgamblin:
Approving this but I think we need a more general solution to avoid recursive search as a default for finding libs/etc.  I think we should allow packages to specify they need that, and instead make the fast case default.

Issue #35819, opened 2023-03-04
Installation issue: llvm


omsai:
### Steps to reproduce the issue

```console
$ spack spec -I llvm@11.1.0
Input spec
--------------------------------
 -   llvm@11.1.0

Concretized
--------------------------------
 -   llvm@11.1.0%apple-clang@13.1.6+clang~code_signing+compiler-rt~cuda~flang~gold+internal_unwind~ipo+libcxx~link_llvm_dylib+lld+lldb+llvm_dylib~mlir~omp_debug~omp_tsan+polly~python~split_dwarf~z3 build_system=cmake build_type=Release patches=514926d,6811ad4,7b286b1,f920173 targets=none version_suffix=none arch=darwin-ventura-skylake
[+]      ^cmake@3.25.2%apple-clang@13.1.6~doc+ncurses+ownlibs~qt build_system=generic build_type=Release arch=darwin-ventura-skylake
[+]          ^openssl@1.1.1t%apple-clang@13.1.6~docs~shared build_system=generic certs=mozilla arch=darwin-ventura-skylake
[+]              ^ca-certificates-mozilla@2023-01-10%apple-clang@13.1.6 build_system=generic arch=darwin-ventura-skylake
[+]      ^hwloc@2.9.0%apple-clang@13.1.6~cairo~cuda~gl~libudev+libxml2~netloc~nvml~oneapi-level-zero~opencl~pci~rocm build_system=autotools libs=shared,static arch=darwin-ventura-skylake
[+]          ^libxml2@2.10.3%apple-clang@13.1.6~python build_system=autotools arch=darwin-ventura-skylake
[+]              ^libiconv@1.17%apple-clang@13.1.6 build_system=autotools libs=shared,static arch=darwin-ventura-skylake
[+]      ^libedit@3.1-20210216%apple-clang@13.1.6 build_system=autotools arch=darwin-ventura-skylake
[+]      ^ncurses@6.4%apple-clang@13.1.6~symlinks+termlib abi=none build_system=autotools arch=darwin-ventura-skylake
[+]      ^ninja@1.11.1%apple-clang@13.1.6+re2c build_system=generic arch=darwin-ventura-skylake
[+]          ^re2c@2.2%apple-clang@13.1.6 build_system=generic arch=darwin-ventura-skylake
[+]      ^perl-data-dumper@2.173%apple-clang@13.1.6 build_system=perl arch=darwin-ventura-skylake
[...]

PR #35820, opened 2023-03-04
Bugfix/llvm 11 darwin disable libcxx static abi

omsai:


omsai:
Actually I'm no longer sure that disabling static linking is a good way to resolve the issue with #35819 because without static linking all the downstream uses of llvm would need adding the `clang` option `-stdlib=libc++`.

naromero77:
@omsai Can we close this PR?

PR #35827, opened 2023-03-05
Update biopython, gsd, and griddataformats

RMeli:


var/spack/repos/builtin/packages/py-griddataformats/package.py
```diff
@@ -15,12 +15,39 @@ class PyGriddataformats(PythonPackage):
...
+    )
+    version(
+        "0.4.0",
+        sha256="f81d6b75aa7ebd9e8b64e14558c2d2583a0589829382beb4ef69860110261512",
+        deprecated=True,
+    )
+    version(
+        "0.3.3",
+        sha256="938f0efcb3bc2f58ec85048b933942da8a52c134170acc97cb095f09d3698fbd",
+        deprecated=True,
```
adamjstewart, line 39: What's the motivation for deprecation? (not that a good reason is required, just curious)

var/spack/repos/builtin/packages/py-griddataformats/package.py
```diff
@@ -15,12 +15,39 @@ class PyGriddataformats(PythonPackage):
...
+    )
 
-    depends_on("python@2.7:")
     depends_on("py-setuptools", type="build")
-    depends_on("py-numpy@1.0.3:", type=("build", "run"))
-    depends_on("py-six", type=("build", "run"))
+
+    depends_on("python@3.8:", when="@1.0.1:")
```
adamjstewart: Presumably these should be type=build/run?

var/spack/repos/builtin/packages/py-gsd/package.py
```diff
@@ -18,8 +18,14 @@ class PyGsd(PythonPackage):
...
 
+    maintainers("RMeli")
+
+    version("2.8.0", sha256="f2b031a26a7a5bee5f3940dc2f36c5a5b6670307b297c526adf2e26c1f5b46ae")
     version("1.9.3", sha256="c6b37344e69020f69fda2b8d97f894cb41fd720840abeda682edd680d1cff838")
 
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@42:", type="build", when="@2.8.0:")
     depends_on("py-cython", type="build")
     depends_on("py-numpy@1.9.3:19", type=("build", "run"))
```
adamjstewart: This upper bound doesn't make sense, and no upper bound is listed in the package

var/spack/repos/builtin/packages/py-gsd/package.py
```diff
@@ -18,8 +18,14 @@ class PyGsd(PythonPackage):
...
+
+    depends_on("python@3.6:", when="@2.8.0:")
```
adamjstewart: Presumably this should be type=build/run. Also, Spack only supports Python 3.7+, so you don't actually have to list this if you don't want to.

var/spack/repos/builtin/packages/py-gsd/package.py
RMeli: It is listed in the `pyproject.toml`: https://github.com/glotzerlab/gsd/blob/bdc281ac09bf498bf87022c16798f226d41269ca/pyproject.toml#L6

var/spack/repos/builtin/packages/py-gsd/package.py
RMeli:
```suggestion
```

var/spack/repos/builtin/packages/py-griddataformats/package.py
RMeli: Version `0.5` is already 4 years old, and still supports Python `2.7`. I'm not too keen to support such an old version (there have been other versions in between) since I added myself as a maintainer.

I can remove the deprecations if they are a bit too premature, but it's my way of saying I'm not actively looking into/testing the old versions.

var/spack/repos/builtin/packages/py-griddataformats/package.py
RMeli:
```suggestion
    depends_on("python@3.8:", when="@1.0.1:", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-gsd/package.py
adamjstewart: I'm commenting on numpy, not setuptools

var/spack/repos/builtin/packages/py-griddataformats/package.py
adamjstewart: I'm fine with keeping the deprecations since you're maintaining the package now. But `2.7:` means 2.7+ which includes 3, so it's not like users _can't_ use it if they want to. But there are no packages that specifically require older versions of griddataformats, so it should be fine to deprecate them.

var/spack/repos/builtin/packages/py-griddataformats/package.py
```diff
@@ -15,12 +15,39 @@ class PyGriddataformats(PythonPackage):
...
-    depends_on("py-six", type=("build", "run"))
+
+    depends_on("python@3.8:", when="@1.0.1:", type=("build", "run"))
+
+    depends_on("py-numpy@1.19:", when="@1.0.1", type=("build", "run"))
+    depends_on("py-scipy", when="@1.0.1", type=("build", "run"))
+    depends_on("py-mrcfile", when="@1.0.1", type=("build", "run"))
+
+    # Deprecated
+    depends_on("python@2.7:", when="@:0.5.0")
```
adamjstewart:
```suggestion
    depends_on("python@2.7:", when="@:0.5.0", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-gsd/package.py
RMeli: OIC. Digging back version `1.9.3` there was the following: https://github.com/glotzerlab/gsd/blob/dbe614757001036bf9d5a634434921fe536a8ab5/setup.py#L55

I'll remove the upper bound.

var/spack/repos/builtin/packages/py-griddataformats/package.py
RMeli: > But there are no packages that specifically require older versions of griddataformats

Good point, thank you for checking!

var/spack/repos/builtin/packages/py-griddataformats/package.py
```diff
@@ -15,12 +15,39 @@ class PyGriddataformats(PythonPackage):
...
+    depends_on("python@3.8:", when="@1.0.1:", type=("build", "run"))
+
+    depends_on("py-numpy@1.19:", when="@1.0.1", type=("build", "run"))
+    depends_on("py-scipy", when="@1.0.1", type=("build", "run"))
+    depends_on("py-mrcfile", when="@1.0.1", type=("build", "run"))
+
+    # Deprecated
+    depends_on("python@2.7:", when="@:0.5.0", type=("build", "run"))
+    depends_on("py-numpy@1.0.3:", type=("build", "run"), when="@:0.5.0")
+    depends_on("py-six", type=("build", "run"), when="@:0.5.0")
```
adamjstewart:
```suggestion
    depends_on("python@3.8:", when="@1:", type=("build", "run"))

    depends_on("py-numpy@1.19:", when="@1:", type=("build", "run"))
    depends_on("py-scipy", when="@1:", type=("build", "run"))
    depends_on("py-mrcfile", when="@1:", type=("build", "run"))

    # Deprecated
    depends_on("python@2.7:", when="@0", type=("build", "run"))
    depends_on("py-numpy@1.0.3:", when="@0", type=("build", "run"))
    depends_on("py-six", when="@0", type=("build", "run"))
```
Version 1.0.0 was when these changed, presumably these will also be needed in 1.0.2+

var/spack/repos/builtin/packages/py-gsd/package.py
```diff
@@ -18,8 +18,12 @@ class PyGsd(PythonPackage):
...
+    maintainers("RMeli")
+
+    version("2.8.0", sha256="f2b031a26a7a5bee5f3940dc2f36c5a5b6670307b297c526adf2e26c1f5b46ae")
     version("1.9.3", sha256="c6b37344e69020f69fda2b8d97f894cb41fd720840abeda682edd680d1cff838")
 
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@42:", type="build", when="@2.8.0:")
     depends_on("py-cython", type="build")
-    depends_on("py-numpy@1.9.3:19", type=("build", "run"))
+    depends_on("py-numpy@1.9.3:", type=("build", "run"))
```
adamjstewart, line 29:
```suggestion
    depends_on("py-numpy@1.9.3:", type=("build", "run"))
    depends_on("py-numpy@1.9.3:1", when="@:1", type=("build", "run"))
```

var/spack/repos/builtin/packages/py-griddataformats/package.py
RMeli: Yes, the lack of `:` was an oversight, thanks for catching it.

adamjstewart (approved):

PR #35830, opened 2023-03-05
alglib: add v3.20.0

alecbcs:
Added alglib v3.20.0.

https://www.alglib.net/arcnews.php#date_19_12_2022


tldahlgren (approved):
Confirmed the sha256.

PR #35832, opened 2023-03-05
apr: add v1.7.2, apr-util 1.6.3

alecbcs:
Add APT v1.7.2 which includes multiple bug fixes.

**Changelog:**
- configure: Fix various build issues for compilers enforcing strict C99 compliance.
- configure: Prefer posix name-based shared memory over SysV IPC.
- Fix handle leak in the Win32 apr_uid_current implementation.
- Add error handling for lseek() failures in apr_file_write() and apr_file_writev().
- Avoid an overflow on 32 bit platforms.
- Add --tag=CC to libtool invocations.
- Signals: Allow handling of SIGUSR2 in apr_signal_thread.

Full changelog available [here](https://downloads.apache.org/apr/CHANGES-APR-1.7).


tldahlgren (approved):
Confirmed both sha256.

PR #35842, opened 2023-03-05
blktrace: add v1.3.0

alecbcs:
Add blktrace v1.3.0.

https://git.kernel.dk/cgit/blktrace/log/


tldahlgren (approved):
Confirmed the sha256.

PR #35847, opened 2023-03-05
colordiff: add v1.0.21

alecbcs:
Add colordiff v1.0.21.

**Changelog:**
https://www.colordiff.org/changes.html


tldahlgren (approved):
Confirmed the sha256.

PR #35850, opened 2023-03-06
dash: add v0.5.12

alecbcs:
Add dash v0.5.12.

**Changelog:**
https://git.kernel.org/pub/scm/utils/dash/dash.git/log/


tldahlgren (approved):
Confirmed the sha256.

PR #35851, opened 2023-03-06
dnsmasq: add v2.89

alecbcs:
Add dnsmasq v2.89.

**Changelog:**
- Fix bug in --dynamic-host when an interface has /16 IPv4 address. Thanks to Mark Dietzer for spotting this.
- Add --fast-dns-retry option. This gives dnsmasq the ability to originate retries for upstream DNS queries itself, rather than relying on the downstream client.
- Add --use-stale-cache option.
- Add --port-limit option which allows tuning for robustness in the face of some upstream network errors.

Full changelog can be found [here](https://thekelleys.org.uk/dnsmasq/CHANGELOG).


tldahlgren (approved):
Confirmed the sha256.

PR #35853, opened 2023-03-06
easybuild: add v4.7.0

alecbcs:
Add easybuilld v4.7.0. 

**Summarized Changelog:**
- also run unit tests with Python 3.11;
- add support for checksums specified in external checksums.json file;
- take into account custom configuration options specified in easystack files - see also [docs](https://docs.easybuild.io/en/latest/Easystack-files.html);
- add support for using --output-format=md (MarkDown);
- add support for postinstallmsgs;
- add gfbf as subtoolchain of foss;
- make iimkl toolchain aware of intel-compilers;
- add generic easyblocks for installing (bundle of) Julia packages: JuliaPackage + JuliaBundle;

Full changelog [here](https://github.com/easybuilders/easybuild/releases/tag/easybuild-v4.7.0).

### DIFF
--- a/var/spack/repos/builtin/packages/easybuild/package.py
+++ b/var/spack/repos/builtin/packages/easybuild/package.py
@@ -18,11 +18,12 @@ class Easybuild(PythonPackage):
     version("4.7.0", sha256="e68c280e3a508965aefbdf234872919f777f739cf4787cb11bb97a56e0cf74a6")
     version("4.0.0", sha256="21bcc1048525ad6219667cc97a7421b5388068c670cabba356712e474896de40")
 
-    depends_on("python@2.6:2.8,3.5:", when="@4:", type=("build", "run"))
+
+    depends_on("python@3.5:", type=("build", "run"))
     # pip silently replaces distutils with setuptools
     depends_on("py-setuptools", type="build")
 
-    for v in ["@4.0.0"]:
+    for v in ["@4.0.0", "@4.7.0"]:
         depends_on("py-easybuild-framework" + v, when=v, type="run")
         depends_on("py-easybuild-easyblocks" + v, when=v, type="run")
         depends_on("py-easybuild-easyconfigs" + v, when=v, type="run")
--- a/var/spack/repos/builtin/packages/easybuild/package.py
+++ b/var/spack/repos/builtin/packages/easybuild/package.py
@@ -18,7 +18,6 @@ class Easybuild(PythonPackage):
     version("4.7.0", sha256="e68c280e3a508965aefbdf234872919f777f739cf4787cb11bb97a56e0cf74a6")
     version("4.0.0", sha256="21bcc1048525ad6219667cc97a7421b5388068c670cabba356712e474896de40")
 
-
     depends_on("python@3.5:", type=("build", "run"))
     # pip silently replaces distutils with setuptools
     depends_on("py-setuptools", type="build")
--- a/var/spack/repos/builtin/packages/easybuild/package.py
+++ b/var/spack/repos/builtin/packages/easybuild/package.py
@@ -15,6 +15,7 @@ class Easybuild(PythonPackage):
     pypi = "easybuild/easybuild-4.0.0.tar.gz"
     maintainers("boegel")
 
+    version("4.7.0", sha256="e68c280e3a508965aefbdf234872919f777f739cf4787cb11bb97a56e0cf74a6")
     version("4.0.0", sha256="21bcc1048525ad6219667cc97a7421b5388068c670cabba356712e474896de40")
 
     depends_on("python@2.6:2.8,3.5:", when="@4:", type=("build", "run"))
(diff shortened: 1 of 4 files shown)


var/spack/repos/builtin/packages/easybuild/package.py
```diff
@@ -15,6 +15,7 @@ class Easybuild(PythonPackage):
     pypi = "easybuild/easybuild-4.0.0.tar.gz"
     maintainers("boegel")
 
+    version("4.7.0", sha256="e68c280e3a508965aefbdf234872919f777f739cf4787cb11bb97a56e0cf74a6")
     version("4.0.0", sha256="21bcc1048525ad6219667cc97a7421b5388068c670cabba356712e474896de40")
 
     depends_on("python@2.6:2.8,3.5:", when="@4:", type=("build", "run"))
```
boegel: Support for Python 2.6 was removed in EasyBuild v4.4.0 (see [release notes](https://docs.easybuild.io/release-notes/#release_notes_eb440)), so this needs an update

var/spack/repos/builtin/packages/easybuild/package.py
```diff
@@ -15,6 +15,7 @@ class Easybuild(PythonPackage):
     pypi = "easybuild/easybuild-4.0.0.tar.gz"
     maintainers("boegel")
 
+    version("4.7.0", sha256="e68c280e3a508965aefbdf234872919f777f739cf4787cb11bb97a56e0cf74a6")
```
boegel, line 18: I *think* that `@4.7.0` also needs to be added in the `for` loop below, where the dependencies on `py-easybuild-*` are declared (and hence those packages need an update too)

boegel (changes requested):

var/spack/repos/builtin/packages/easybuild/package.py
adamjstewart: Could just delete this line entirely, Spack only supports installing Python 3.7+ so there's no reason to restrict it

tldahlgren (commented):
FWIW. Confirmed the sha256.

var/spack/repos/builtin/packages/easybuild/package.py
alecbcs: I think I'm going to keep,
```python
depends_on("python@3.5:", type=("build", "run"))
```

Just in case someone is using an external `python@:3.5`. @adamjstewart are there any objections to doing so?

var/spack/repos/builtin/packages/easybuild/package.py
alecbcs: Added! Thanks for noticing this!

var/spack/repos/builtin/packages/easybuild/package.py
adamjstewart: No objections, hopefully no one is doing that though

var/spack/repos/builtin/packages/py-easybuild-easyblocks/package.py
```diff
@@ -16,10 +16,11 @@ class PyEasybuildEasyblocks(PythonPackage):
     maintainers("boegel")
 
     version("4.0.0", sha256="a0fdef6c33c786e323bde1b28bab942fd8e535c26842877d705e692e85b31b07")
+    version("4.7.0", sha256="c23e81cbaa3e4fa5ab1bb8b2db759332867d61110bf4ec34763ea170780f0655")
```
adamjstewart, line 18: Can you sort these newest to oldest?

var/spack/repos/builtin/packages/py-easybuild-framework/package.py
```diff
@@ -16,6 +16,7 @@ class PyEasybuildFramework(PythonPackage):
     maintainers("boegel")
 
     version("4.0.0", sha256="f5c40345cc8b9b5750f53263ade6c9c3a8cd3dfab488d58f76ac61a8ca7c5a77")
+    version("4.7.0", sha256="ea51c3cb88fca27daadd2fb55ee31f5f51fc60c4e3519ee9d275954540312df8")
```
adamjstewart, line 18: Same here

adamjstewart (dismissed):
Minor style thing, otherwise looks good

var/spack/repos/builtin/packages/py-easybuild-easyblocks/package.py
alecbcs: Is this new? I thought we always sorted oldest to newest?

tldahlgren: Uh oh.  I didn't check this change on your other PRs.

tldahlgren: No, this is not new.  Spack prefers newer versions be listed first.

alecbcs: Oh I just did this backwards for the most recent dependencies I added. Sorry lol I was confused because I thought @adamjstewart was suggesting I put the most recent versions at the bottom not the top.

alecbcs: Done! @tldahlgren the rest of the PRs I did should be good. I believe this was a one off this morning when I wasn't paying attention.

var/spack/repos/builtin/packages/py-easybuild-framework/package.py
alecbcs: Done!

adamjstewart (approved):

PR #35862, opened 2023-03-06
fonttosfnt: add v1.2.2

alecbcs:
Add fonttosfnt v1.2.2.

**Changelog:**
https://cgit.freedesktop.org/xorg/app/fonttosfnt/log/


tldahlgren (approved):
Confirmed the sha256.

PR #35869, opened 2023-03-06
fslsfonts: add v1.0.6

alecbcs:
Add fslsfonts v1.0.6.

**Changelog:**
https://cgit.freedesktop.org/xorg/app/fslsfonts/log/


tldahlgren (approved):
Confirmed the sha256.

PR #35875, opened 2023-03-06
grep: add v3.9

alecbcs:
Add grep v3.9.

**Changelog:**
https://lists.gnu.org/archive/html/info-gnu/2023-03/msg00001.html


tldahlgren (approved):
Confirmed the sha256.

PR #35880, opened 2023-03-06
iceauth: add v1.0.9

alecbcs:
Add iceauth v1.0.9.

**Changelog:**
https://cgit.freedesktop.org/xorg/app/iceauth/log/


tldahlgren (approved):
Confirmed the sha256.

PR #35881, opened 2023-03-06
ico: add v1.0.6

alecbcs:
Add ico v1.0.6.

**Changelog:**
https://cgit.freedesktop.org/xorg/app/ico/log/


tldahlgren (approved):
Confirmed the sha256

PR #35885, opened 2023-03-06
lammps: GPU/Kokkos package updates

rbberger:
LAMMPS has multiple GPU backends. The most actively developed one is for Kokkos, but the older GPU package also supports various backends, including HIP. 

This adds the missing `rocm` variant to lammps. By default, if kokkos is not enabled, it will build with the GPU package, as has been the case with the existing `+opencl` and `+cuda` variants.

The patch is necessary because of the hipcub dependency. It gets installed into a non-standard location when used via Spack. The current LAMMPS cmake was making some assumptions which break in that case. Since I have to choose some sort of cutoff for maintaining this, the patch is only applied beginning from the last stable release.

This and another issue in upstream `develop` should be resolved soon with https://github.com/lammps/lammps/pull/3674

I've also added a new dependent variant `gpu_precision`, which was/is actually the most significant benefit of the GPU package over Kokkos, as it allows you to choose between single, mixed, or double precision when doing the offloading.

For `+kokkos`, I've also added the propagation of the `cuda_arch` and `amdgpu_target` properties.

### DIFF
--- a/var/spack/repos/builtin/packages/lammps/hip_cmake.patch
+++ b/var/spack/repos/builtin/packages/lammps/hip_cmake.patch
@@ -1,0 +1,58 @@
+From b11049ba1a5fa22ff575f4d4afa6579973425962 Mon Sep 17 00:00:00 2001
+From: Richard Berger <richard.berger@outlook.com>
+Date: Sun, 5 Mar 2023 19:03:38 -0700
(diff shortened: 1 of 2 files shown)


alalazo (approved):
Checked package syntax and idioms and LGTM

PR #35889, opened 2023-03-06
new py-thop package

Sangu-Mbekelu:


Sangu-Mbekelu:
@spackbot fix style

tldahlgren (commented):
Confirmed the sha256.

var/spack/repos/builtin/packages/py-thop/package.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+from spack.package import *
+
+
+class PyThop(PythonPackage):
+    """A tool to count the FLOPs of PyTorch model."""
+
+    homepage = "https://github.com/Lyken17/pytorch-OpCounter/"
+    url = "https://files.pythonhosted.org/packages/bb/0f/72beeab4ff5221dc47127c80f8834b4bcd0cb36f6ba91c0b1d04a1233403/thop-0.1.1.post2209072238-py3-none-any.whl"
```
adamjstewart:
```suggestion
    url = "https://pypi.io/packages/py3/t/thop/thop-0.1.1.post2209072238-py3-none-any.whl"
```
This URL will work for all versions

var/spack/repos/builtin/packages/py-thop/package.py
```diff
@@ -0,0 +1,22 @@
...
+
+    version(
+        "0.1.1.post2209072238",
+        sha256="01473c225231927d2ad718351f78ebf7cffe6af3bed464c4f1ba1ef0f7cdda27",
+        expand=False,
+    )
+
+    depends_on("py-setuptools", type=("build", "run"))
```
adamjstewart: Setuptools isn't actually needed because we're installing a pre-built wheel, but torch is required:
```suggestion
    depends_on("py-torch", type=("build", "run"))
```

adamjstewart (changes requested):

adamjstewart (approved):

PR #35894, opened 2023-03-06
libxc: use gitlab release tarballs

haampie:
The current website is not particularly reliable, often hangs up
half-way a download.


var/spack/repos/builtin/packages/libxc/package.py
```diff
@@ -11,30 +11,36 @@ class Libxc(AutotoolsPackage, CudaPackage):
...
-    version("2.2.2", sha256="6ca1d0bb5fdc341d59960707bc67f23ad54de8a6018e19e02eee2b16ea7cc642")
-    version("2.2.1", sha256="ade61c1fa4ed238edd56408fd8ee6c2e305a3d5753e160017e2a71817c98fd00")
+
+    # The webserver at https://tddft.org/programs/libxc/download is unreliable,
+    # see https://gitlab.com/libxc/libxc/-/issues/453. For the time being use
+    # GitLab release tarballs.
+    url = "https://gitlab.com/libxc/libxc/-/archive/6.1.0/libxc-6.1.0.tar.gz"
+
+    version("6.1.0", sha256="f593745fa47ebfb9ddc467aaafdc2fa1275f0d7250c692ce9761389a90dd8eaf")
+    version("6.0.0", sha256="0c774e8e195dd92800b9adf3df5f5721e29acfe9af4b191a9937c7de4f9aa9f6")
```
alalazo, line 17: Can we change the `url` only for these two? The other ones are in our source mirror, so they should still be reliable.

var/spack/repos/builtin/packages/libxc/package.py
```diff
@@ -11,30 +11,35 @@ class Libxc(AutotoolsPackage, CudaPackage):
...
+    version("5.1.3", sha256="53cbccae4227794c4f4e042d0c20c3171894d8eacd13fd78aab376d5971a966c")
+    version("5.1.2", sha256="3fe05ccf7033622112f7d6ab28ac82301f5840b8d1bbe8b6df1022dbc732883f")
+    version("5.1.0", sha256="5af77427c20a01d6d23ce769c8f40088b6bc7682c7b76cf1f87dae63605d5150")
+    version("5.0.0", sha256="bfe8d8daa700d238bacf15ef2112f75e23d0a74c1a8d6c3e4741319561f577da")
+    version("4.3.4", sha256="83aba38dfa03f34cc74f84c14c83bf501a43493c818c797e2d0682647569b147")
+    version("4.3.2", sha256="e34700331aaffa5a3c421786863daf9a68cee4cbfbd2a79a4578fcf783205235")
+    version("4.2.3", sha256="c68f5171828e5dd9ebdaa5c1a914f91a1456bbd18fc52041d92f662e790f6750")
+    version("3.0.0", sha256="df2362351280edaf2233f3b2c8eb8e6dd6c68105f152897a4cc629fa346a7396")
+    version("2.2.2", sha256="6ffaad40505dbe8f155049448554b54ea31d31babf74ccf6b7935bfe55eeafd8")
+    version("2.2.1", sha256="c8577ba1ddd5c28fd0aa7c579ae65ab990eb7cb51ecf9f8175f9251f6deb9a06")
```
alalazo: Was this overlooked? Or am I missing something?

haampie: Ugh, I'm on my phone, can you force push a partial shasum revert and merge?

alalazo (approved):

tldahlgren:
Wow!  It's like "wack-a-mole" with the checksums.  

I encountered a pipeline that had a checksum issue with `libxc` (https://gitlab.spack.io/spack/spack/-/jobs/5945204) so thought I'd check all of them.

Just about every time I run `spack checksum` on `5.2.3` to `5.0.0` they are different from the time before.  Even `2.2.1` was "off" unless I explicitly paste the old `url` as an argument.

haampie:
Ugh, yeah... in Spack you don't see this, but if you download with `curl` it'll tell you the remote hung up, so you get partial files. On the Spack side, that error is not caught, because we use `shutil.copyfileobj` to read chunks from the stream, and somehow that doesn't check whether # consumed bytes == content length.

haampie:
Also, python doesn't seem to care about incomplete downloads :p 

https://github.com/python/cpython/blob/11a2c6ce516b24b2435cb627742a6c4df92d411c/Lib/http/client.py#L467-L468

PR #35900, opened 2023-03-06
Debugging: add log statements to debug custom modifications

scheibelp:
Add logging to help users debug where external file searches are taking a long time. If it looks like Spack is "hung", it may be that some custom package modification is taking a long time. This often occurs for external packages when they search for files recursively from their base prefix. One such example is the `tk` package: which has:

```
env.set("TK_LIBRARY", os.path.dirname(sorted(find(self.prefix, "tk.tcl"))[0]))
```

file exists but takes quite a while to locate (e.g. `/usr/share/tk8.5/tk.tcl`) and starting from a deeper directory would speed the search up. This PR doesn't address that specifically, but without the edits here it's difficult to pinpoint which package is taking a long time.

### DIFF
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -485,7 +485,13 @@ def set_wrapper_variables(pkg, env):
             query = pkg.spec[dep.name]
             dep_link_dirs = list()
             try:
+                # In some circumstances (particularly for externals) finding
+                # libraries packages can be time consuming, so indicate that
+                # we are performing this operation (and also report when it
+                # finishes).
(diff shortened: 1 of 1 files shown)


scheibelp:
Note: https://github.com/spack/spack/pull/35902 updates the `tk` package recipe to resolve the specific case of external `tk`, but this PR would be useful for detecting other similar problems.

tldahlgren (approved):
LGTM

haampie:
FWIW, this is conflicting with https://github.com/spack/spack/pull/35737. Can you keep track of that PR @scheibelp so that your changes here don't disappear due to changes there?

PR #35902, opened 2023-03-07
Tk/Tcl packages: speed up file search

scheibelp:
Search more-specific prefix first before falling back to root. For an external `tk` package, this should make it faster for dependents to locate the script and set `TK_LIBRARY` on some systems.


haampie:
Would it help to limit search depth for externals?

scheibelp:
In this case the file of interest is in `<prefix>/share/tk8.5/`. Without collecting statistics, ~I think most files are contained within this depth.~ (EDIT: https://github.com/spack/spack/pull/35902#issuecomment-1463524984 demonstrates that it is definitely not most files, but I later argue that it is enough that the result is still slow)

Even if choosing a depth would speed up the search, I worry about missing files contained in nested directories: how would this truncated search be conveyed clearly to users? I'd prefer to use something like https://github.com/spack/spack/pull/35900 to find and fix overly-broad searches on an individual basis than to miss the file (especially when it looks like we should not be missing it).

haampie:
I don't think it makes a lot of sense to search at high depths for things that are typically one or two dirs deep.

An alternative approach that doesn't set arbitrary depth limits is iterative deepening: first search at depth 0, then 1, then 2, etc instead of going depth-first into that one dir that has tons and tons of subdirs, but not the file you're looking for.

scheibelp:
> I don't think it makes a lot of sense to search at high depths for things that are typically one or two dirs deep.

This requires defining "typical" which IMO is harder than just searching everything.

> An alternative approach that doesn't set arbitrary depth limits is iterative deepening: first search at depth 0, then 1, then 2, etc instead of going depth-first into that one dir that has tons and tons of subdirs,

Do you know that depth-first is the reason this is slow? You can fit a lot of files into 3 levels of directories (i.e. `.`, `./sub`, and `./sub/sub`). Also, file search could be slow for other reasons (e.g. if a directory is a mount) and searching specific paths on a per-package basis should avoid all of them.

I could see the benefit of reorganizing the generic file search, but I'm not sure if I want to handle all of the considerations as part of addressing this issue (e.g. reimplementing `os.walk`).

haampie:
> Do you know

It's not very hard to try and implement, right? I just looked at some HPC system's /usr dir and it shows 98% of the directories are at depth > 2, so, it may be worth it:

```
$ for i in {00..15}; do echo "$i: $(find /usr/ -mindepth $i -maxdepth $i -type d 2>/dev/null | wc -l)"; done
      ┌                                        ┐ 
    0 ┤ 1                                        
    1 ┤ 12                                       
    2 ┤■■ 476                                    
    3 ┤■■■■■■■ 1 894                             
    4 ┤■■■■■■■■■■■■■ 3 565                       
    5 ┤■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 8 991   
    6 ┤■■■■■■■■■■■■■■■■■■ 4 781                  
    7 ┤■■■■■■■ 1 975                             
    8 ┤■■■ 789                                   
    9 ┤■■■ 697                                   
   10 ┤■■■ 782                                   
   11 ┤ 17                                       
   12 ┤ 9                                        
   13 ┤ 1                                        
   14 ┤ 0                                        
      └                                        ┘ 
```

Is it similar on your system?

scheibelp:
(EDIT: the assumptions in this comment leading to searching an estimated 10% of entries are later shown false, and the ~2% estimate is more likely to be the correct one)

If I take your data from https://github.com/spack/spack/pull/35902#issuecomment-1463524984

* I'm searching for something in `<prefix>/share/tk8.5/` (to be clear, locating `tk8.5` is not sufficient: I want to find a file which is usually inside of that) so that would be up to and including `03` in your listing
* The total is 23989, and the sum of all things up to (and including) `03` is 2382 (I parsed out `[12, 476, 1894, 3565, 8991, 4781, 1975, 789, 697, 782, 17, 9, 1, 0, 0]`), which is approximately 10% (`0.099`) or on average `5%` (assuming it doesn't have to search through everything at depth `03`
* My searches were taking 5-10 minutes (before I terminated them), which for these numbers places a lower bound of 15-30 seconds of search time
* The specific prefixes are much faster than that

I agree with you that we could be improving the generic file search, and I like the concept of using BFS to deal with it, but even so, IMO it's worth merging this PR.

haampie:
Yes, let's merge this PR.

> so that would be up to and including 03 in your listing

It was using `-type d`, so that'd be 02, not 03.

> I like the concept of using BFS to deal with it

Fwiw, iterative deepening is typically depth-first so it has the same memory footprint as dfs, at the cost of visiting nodes multiple times. But yeah, bfs or iterative deepening would probably both work.

(Also: `find(...)` doesn't early exit on the first match)

var/spack/repos/builtin/packages/tcl/package.py
```diff
@@ -93,6 +93,17 @@ def command(self):
             os.path.realpath(self.prefix.bin.join("tclsh{0}".format(self.version.up_to(2))))
         )
 
+    def _find_script_dir(self):
```
haampie, line 98: What about error handling? For libraries / headers an exception is thrown iirc.

scheibelp:
>> so that would be up to and including 03 in your listing

> It was using -type d, so that'd be 02, not 03.

`file -type d` from what I understand just means "only show things that are directories": it does not have implications on the depth of the search. For example for the following command:

```
find /usr/ -mindepth X -maxdepth X | grep -i "tk.tcl"
```

to return anything, X has to be 3 for me (on a system where `tk.tcl` is at `/usr/share/tk8.5/tk.tcl`)

haampie:
That's just how `find` defines depth, a dir is just a file with a list of its contents, so if you wanna find files at x/y/file you open x/y at depth 2. In any case, let's move on with this PR and move the search order discussion to #36083 🙃

var/spack/repos/builtin/packages/tcl/package.py
scheibelp: Overall think I should raise an exception here, as you say (that would generally preserve the current behavior although more explicitly). Regarding the behavior of `.libs`/`.headers`:

* `find` (and associated implementations like `find_libraries`) return `[]` and don't fail.
* The default implementation of `.libs` does fail, although I don't know if those semantics need to be copied here on that account (i.e. the user could decide just not to set the associated env variable)
* Without this PR, a failure to locate any files would fail when the code attempts to dereference the first element of an empty list

scheibelp: I now raise an exception in `_find_script_dir` if nothing is found (which restores the prior behavior, but with a clearer error message).

scheibelp:
> That's just how find defines depth, a dir is just a file with a list of its contents, so if you wanna find files at x/y/file you open x/y at depth 2.

OK I think I understand now: given that, it is entirely possible that https://github.com/spack/spack/pull/36083 could sufficiently reduce the search times in this case. I could test that before this is merged although I don't think merging this would hurt (and would cut down on unnecessary extra FS traversal either way).

scheibelp:
On the system that prompted this PR, if I use the logic in https://github.com/spack/spack/pull/36083:

```
time spack-python -c "from llnl.util.filesystem import FindFile; x = FindFile('/usr/', 'tk.tcl'); x.run()"
'/usr/share/tk8.5/tk.tcl'

real	0m8.360s
user	0m1.329s
sys	0m1.737s
```

To get that time savings, these packages would need to be updated to use `FindFile` (or likewise pass some argument to `find` to indicate that only one result is expected/needed).

haampie:
I've renamed it to `find_first` there

mdavis36:
@haampie @scheibelp @tldahlgren Will this make it into the upcoming release, we would like these changes if possible.

haampie (approved):

PR #35904, opened 2023-03-07
py-opencv-python: new package

heerener:
I really hope I didn't mess up the numpy dependency mess here... if there's a better way of translating this into Spack syntax, I'm very interested to learn.

### DIFF
--- a/var/spack/repos/builtin/packages/py-opencv-python/package.py
+++ b/var/spack/repos/builtin/packages/py-opencv-python/package.py
@@ -5,8 +5,7 @@
 
 import platform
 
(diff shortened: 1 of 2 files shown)


adamjstewart:
We've been using `opencv+python3` instead of `py-opencv-python`, not sure if it's worth duplicating the package.

heerener:
We had one internal package that depended on py-opencv-python but I've managed to replace that with your suggested dependency on `opencv+python3+python_bindings_generator+imgproc`. That means I have no reason left to advocate for this.

PR #35905, opened 2023-03-07
openblas: use "all" target since it's marked sequential

haampie:
openblas likes to concurrently write to the same archive from different
targets, and solves that through .NOTPARALLEL: all

We run `make x y z` which is not affected by `NOTPARALLEL`, running into
races.

This adds back the `tests` target unconditionally, but I don't see a better
way other than running `make x && make y && make z`.


haampie:
@spackbot run pipeline

alalazo (approved):

PR #35914, opened 2023-03-07
modules: add support for autoloading only run dependencies

michaelkuhn:
This adds support for autoloading run dependencies in module files and makes it the default since Spack already takes care of link dependencies via rpath. In contrast to `direct`, this can significantly speed up module loading.

This Git module autoloads all direct dependencies:
```
Benchmark 1: module load git
  Time (mean ± σ):     695.4 ms ±  23.4 ms    [User: 676.6 ms, System: 18.7 ms]
  Range (min … max):   658.5 ms … 736.6 ms    10 runs
```

This Git module autoloads only run dependencies:
```
Benchmark 1: module load git
  Time (mean ± σ):      38.9 ms ±   2.9 ms    [User: 31.2 ms, System: 8.1 ms]
  Range (min … max):    36.0 ms …  52.1 ms    73 runs
```

### DIFF
--- a/etc/spack/defaults/modules.yaml
+++ b/etc/spack/defaults/modules.yaml
@@ -45,11 +45,13 @@ modules:
 
     tcl:
       all:
-        autoload: direct
+        autoload: run
+      hide_implicits: true
 
     # Default configurations if lmod is enabled
     lmod:
       all:
-        autoload: direct
+        autoload: run
+      hide_implicits: true
       hierarchy:
         - mpi
--- a/lib/spack/docs/module_file_support.rst
+++ b/lib/spack/docs/module_file_support.rst
@@ -654,7 +654,7 @@ Autoloading can also be enabled conditionally:
 The configuration file above will produce module files that will
 load their direct dependencies if the package installed depends on ``python``.
 The allowed values for the ``autoload`` statement are either ``none``,
-``direct`` or ``all``.
+``run``, ``direct`` or ``all``.
 
 .. note::
   Tcl prerequisites
--- a/lib/spack/spack/build_environment.py
+++ b/lib/spack/spack/build_environment.py
@@ -926,11 +926,12 @@ class SetupContext:
     """This class encapsulates the logic to determine environment modifications, and is used as
     well to set globals in modules of package.py."""
 
-    def __init__(self, *specs: spack.spec.Spec, context: Context) -> None:
+    def __init__(self, *specs: spack.spec.Spec, context: Context, run_root: bool = True) -> None:
         """Construct a ModificationsFromDag object.
         Args:
             specs: single root spec for build/test context, possibly more for run context
-            context: build, run, or test"""
+            context: build, run, or test
+            run_root: whether root spec should be runnable"""
         if (context == Context.BUILD or context == Context.TEST) and not len(specs) == 1:
             raise ValueError("Cannot setup build environment for multiple specs")
         specs_with_type = effective_deptypes(*specs, context=context)
@@ -953,7 +954,7 @@ class SetupContext:
         self.should_setup_dependent_build_env = UseMode.BUILDTIME | UseMode.BUILDTIME_DIRECT
         self.should_setup_build_env = UseMode.ROOT if context == Context.BUILD else UseMode(0)
 
-        if context == Context.RUN or context == Context.TEST:
+        if (context == Context.RUN or context == Context.TEST) and run_root:
             self.should_be_runnable |= UseMode.ROOT
             self.should_setup_run_env |= UseMode.ROOT
 
--- a/lib/spack/spack/modules/common.py
+++ b/lib/spack/spack/modules/common.py
@@ -125,16 +125,16 @@ def dependencies(spec, request="all"):
 
     Args:
         spec: spec to be analyzed
-        request: either 'none', 'direct' or 'all'
+        request: either 'none', 'run', 'direct' or 'all'
 
     Returns:
         list of dependencies
 
         The return list will be empty if request is 'none', will contain
-        the direct dependencies if request is 'direct', or the entire DAG
-        if request is 'all'.
+        the run dependencies if request is 'run', the direct dependencies
+        if request is 'direct', or the entire DAG if request is 'all'.
     """
-    if request not in ("none", "direct", "all"):
+    if request not in ("none", "run", "direct", "all"):
         message = "Wrong value for argument 'request' : "
         message += "should be one of ('none', 'direct', 'all')"
         raise tty.error(message + " [current value is '%s']" % request)
@@ -144,6 +144,24 @@ def dependencies(spec, request="all"):
 
     if request == "direct":
         return spec.dependencies(deptype=("link", "run"))
+
+    if request == "run":
+        r_deps = []
+        deps_with_type = spack.build_environment.effective_deptypes(spec, context=Context.RUN)
+
+        for dep, dep_type in deps_with_type:
+            if dep_type & spack.build_environment.UseMode.RUNTIME_EXECUTABLE:
+                r_deps.append(dep)
+            elif dep_type & spack.build_environment.UseMode.RUNTIME:
+                context = spack.build_environment.SetupContext(
+                    dep, context=Context.RUN, run_root=False
+                )
+                context.set_all_package_py_globals()
+
+                if len(context.get_env_modifications()) > 0:
+                    r_deps.append(dep)
+
+        return r_deps
 
     # FIXME : during module file creation nodes seem to be visited multiple
     # FIXME : times even if cover='nodes' is given. This work around permits
@@ -183,7 +201,7 @@ def merge_config_rules(configuration, spec):
     # Transform keywords for dependencies or prerequisites into a list of spec
 
     # Which modulefiles we want to autoload
-    autoload_strategy = spec_configuration.get("autoload", "direct")
+    autoload_strategy = spec_configuration.get("autoload", "run")
     spec_configuration["autoload"] = dependencies(spec, autoload_strategy)
 
     # Which instead we want to mark as prerequisites
--- a/lib/spack/spack/schema/modules.py
+++ b/lib/spack/spack/schema/modules.py
@@ -32,7 +32,7 @@ array_of_strings = {"type": "array", "default": [], "items": {"type": "string"}}
 
 dictionary_of_strings = {"type": "object", "patternProperties": {r"\w[\w-]*": {"type": "string"}}}
 
-dependency_selection = {"type": "string", "enum": ["none", "direct", "all"]}
+dependency_selection = {"type": "string", "enum": ["none", "run", "direct", "all"]}
 
 module_file_configuration = {
     "type": "object",
--- a/lib/spack/spack/test/data/modules/lmod/autoload_run.yaml
+++ b/lib/spack/spack/test/data/modules/lmod/autoload_run.yaml
@@ -1,0 +1,10 @@
+enable:
+  - lmod
+lmod:
+  core_compilers:
+    - 'clang@3.3'
+  hierarchy:
+    - mpi
+
+  all:
+    autoload: run
--- a/lib/spack/spack/test/data/modules/tcl/autoload_run.yaml
+++ b/lib/spack/spack/test/data/modules/tcl/autoload_run.yaml
@@ -1,0 +1,5 @@
+enable:
+  - tcl
+tcl:
+  all:
+    autoload: run
--- a/lib/spack/spack/test/modules/lmod.py
+++ b/lib/spack/spack/test/modules/lmod.py
@@ -120,6 +120,20 @@ class TestLmod:
         assert "whatis([[Version : 3.0.4]])" in content
         assert 'family("mpi")' in content
 
+    def test_autoload_run(self, modulefile_content, module_configuration):
+        """Tests the automatic loading of direct dependencies."""
+
+        module_configuration("autoload_run")
+
+        # dtbuild1 has
+        # - 1 ('run',) dependency
+        # - 1 ('build','link') dependency
+        # - 1 ('build',) dependency
+        # Just make sure the 'build' dependency is not there
+        content = modulefile_content("dtbuild1")
+
+        assert len([x for x in content if "depends_on(" in x]) == 1
+
     def test_autoload_direct(self, modulefile_content, module_configuration):
         """Tests the automatic loading of direct dependencies."""
 
--- a/lib/spack/spack/test/modules/tcl.py
+++ b/lib/spack/spack/test/modules/tcl.py
@@ -31,6 +31,20 @@ class TestTcl:
 
         assert "module-whatis {mpich @3.0.4}" in content
 
+    def test_autoload_run(self, modulefile_content, module_configuration):
+        """Tests the automatic loading of run dependencies."""
+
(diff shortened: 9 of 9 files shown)


haampie:
Sounds good to me, but link-type deps may have libraries that do `execvp(...)` and require dirs of their run-type deps in PATH. Also they may need variables in some weird cases from `setup_run_environment` to locate config files or whatever.

So, as an option it's fine, but I wouldn't set it as a default

haampie:
Btw, is you benchmark run with tcl? I'm getting `145.3 ms ±   1.7 ms` with `direct` using lmod. I wonder if the tcl version just traverses all possible paths instead of each node once.

michaelkuhn:
> Sounds good to me, but link-type deps may have libraries that do `execvp(...)` and require dirs of their run-type deps in PATH. Also they may need variables in some weird cases from `setup_run_environment` to locate config files or whatever.
> 
> So, as an option it's fine, but I wouldn't set it as a default

Good point, I didn't think about that. Default for both tcl (which currently is none) and lmod (which currently is direct, that is, no change)?

michaelkuhn:
> Btw, is you benchmark run with tcl? I'm getting `145.3 ms ± 1.7 ms` with `direct` using lmod. I wonder if the tcl version just traverses all possible paths instead of each node once.

Yes, I'm using tcl.

adamjstewart:
> but link-type deps may have libraries that do `execvp(...)` and require dirs of their run-type deps in PATH

So... then they should be `type=("link", "run")`?

haampie:
> So... then they should be type=("link", "run")?

Not sure if I understand what you're saying, but no, a package shouldn't have to specify a run-type dep on a link-type dep that has run-type deps. Spack should just set the dependency up correctly. Also it's completely infeasible cause you can have a chain of a hundred optional link-type deps where the tail ultimately needs a run-type dep; if you want that run edge to propagate all the way to the root, you might as well stop distinguishing link/run edges.

adamjstewart:
Oh sorry, I completely misunderstood, I thought you meant direct deps, not transitive deps.

michaelkuhn:
> Btw, is you benchmark run with tcl? I'm getting `145.3 ms ± 1.7 ms` with `direct` using lmod. I wonder if the tcl version just traverses all possible paths instead of each node once.

The results also seem to depend heavily on the version of environment-modules. The ones from the initial post were measured using version 4.5.2. With 5.2.0, I get:
```
Benchmark 1: module load git-2.39.2-gcc-12.2.1-3kt4ygd
  Time (mean ± σ):     184.4 ms ±   7.3 ms    [User: 174.1 ms, System: 9.5 ms]
  Range (min … max):   172.8 ms … 194.3 ms    15 runs
```
vs.
```
Benchmark 1: module load git-2.39.2-gcc-12.2.1-3kt4ygd
  Time (mean ± σ):      32.3 ms ±   1.2 ms    [User: 25.6 ms, System: 6.6 ms]
  Range (min … max):    29.5 ms …  35.1 ms    89 runs
```

michaelkuhn:
I've pushed a new version that reverts the change to lmod's default but changes tcl's default to direct for consistency.

michaelkuhn:
After spending some more time with this, I came up with a better solution: Instead of autoloading only direct run dependencies, we now take a look at the DAG of link+run dependencies and autoload all run dependencies that appear somewhere in the DAG. This should take care of @haampie's concern above.

This Git module autoloads all dependencies:
```
Benchmark 1: module load git
  Time (mean ± σ):     199.0 ms ±   8.2 ms    [User: 188.5 ms, System: 9.9 ms]
  Range (min … max):   186.2 ms … 220.3 ms    15 runs
```

This Git module autoloads direct dependencies:
```
Benchmark 1: module load git
  Time (mean ± σ):     185.3 ms ±   7.1 ms    [User: 173.9 ms, System: 10.6 ms]
  Range (min … max):   175.1 ms … 197.2 ms    16 runs
```

This Git module autoloads only run dependencies:
```
Benchmark 1: module load git
  Time (mean ± σ):      48.2 ms ±   1.7 ms    [User: 40.8 ms, System: 7.1 ms]
  Range (min … max):    45.3 ms …  53.5 ms    60 runs
```

michaelkuhn:
@haampie ping? :slightly_smiling_face:

haampie:
Looking at it!

haampie:
the `[x for x in xs if not (x in ys or ys.add(x))]` is ... not great :p isn't it just a difficult way of expressing `list(set(xs))`?

haampie:
Hm, I still think `autoload: run` is trading correctness for speed. The point stands that `link` type deps require `setup_run_environment`. If speed is not *that* big of a deal in newer versions of tcl, I'm inclined to say we should favor correctness :p 

(That said, we put link-only deps in PATH, which is ugly)

haampie:
Also note that environment-modules 5.1 has `depends-on`, which can replace our ugly logic. (Edit: it's apparently equivalent to prereq, which Spack already supports)

haampie:
It looks like `environment-modules` spends most time running the module files instead of figuring out dependencies, since it doesn't matter if you just add direct deps or all deps to the module. So, the only way to speed it up is to load fewer modules, but we can't just skip link-only deps, that will break things.

haampie:
Okay, I think we're better off without this PR to be honest. Instead:

- Use a sane `module` command, i.e. an up-to-date `lmod`. `tcl` is just horrible in general, `lua` is infinitely better.
- If you don't care about correctness, only about speed, and you want stuff in your `PATH` there's `autoload: none`
- I don't think it's great to add a second option to `autoload` which produces technically broken runtime environments.

For me, `lmod` loads `git` with `autoload: direct` in 80ms, which feels instant, so I don't see a good reason to improve performance.

xdelaruelle:
I have tested this `module load git` and I get following performances on my laptop:
* Environment Modules 4.5.2: ~350ms
* Environment Modules 5.2.0: ~250ms
* Lmod 8.6.12: ~170ms

Environment Modules takes time to compute the dependency and ensure that loaded environment is consistent (every dependency expressed are satisfied). You do not see the immediate benefit on that on this example, but once loaded, if you try to unload a dependent module, for instance `readline`, you will see a difference.

This dependency calculation/check time could be optimized in the future. However this extra time should not be perceived by user.

xdelaruelle:
> Use a sane module command, i.e. an up-to-date lmod. tcl is just horrible in general, lua is infinitely better.

@haampie I do not share your point on this, but my judgment may be biased (as I develop Environment Modules). I would be interested if you could elaborate on this (here or in Slack DM for instance) :-)

michaelkuhn:
I also wasn't able to see a significant speedup using Lmod (similar to @xdelaruelle's numbers above).

@xdelaruelle I was also planning to give the module cache from the upcoming version 5.3.0 a try to see if that improves the situation. I had problems getting the development version to work, though (something about the version not being set, which caused problems with module compatibility checks).

@haampie How about improving the `all` setting instead to only load dependencies that actually cause a change to the run environment? Might be a bit tricky to determine since we have an empty `setup_run_environment` by default. As it stands, this makes module files more or less impossible to use for more extensive software stacks because it can take several minutes to load everything.

haampie:
@xdelaruelle maybe my reply was a *bit* colored :sweat_smile: but I will never understand the point of everything-is-a-string-type languages :grimacing: it reminds me of CMake and its horrors... (I have http://antirez.com/articoli/tclmisunderstood.html on my reading list, maybe I can get convinced it's not that bad). I also expected environment-modules load time for trivial (mostly static) `set-env ...` type module files to be dominated by IO, but that's not the case. In the small tests I did, lmod was much faster, so I'm surprised @michaelkuhn does *not* see this

> How about improving the all setting instead to only load dependencies that actually cause a change to the run environment?

That sounds sensible to me. We can run the relevant methods and see if they contribute anything meaningful to the env changes.

xdelaruelle:
> @xdelaruelle I was also planning to give the module cache from the upcoming version 5.3.0 a try to see if that improves the situation. I had problems getting the development version to work, though (something about the version not being set, which caused problems with module compatibility checks).

@michaelkuhn I made a fix earlier this month on the ability to install modules from a short-depth git clone (see cea-hpc/modules#486). If your issue is coming from something else, do not hesitate to open an issue on cea-hpc/modules project.

xdelaruelle:
> @xdelaruelle maybe my reply was a bit colored sweat_smile but I will never understand the point of everything-is-a-string-type languages grimacing it reminds me of CMake and its horrors... (I have http://antirez.com/articoli/tclmisunderstood.html on my reading list, maybe I can get convinced it's not that bad). I also expected environment-modules load time for trivial (mostly static) set-env ... type module files to be dominated by IO, but that's not the case. In the small tests I did, lmod was much faster, so I'm surprised @michaelkuhn does not see this

@haampie Many thanks for your feedback. It is strange to observe such performance difference between the 2 tools, unless you run this `module load` on a system where modulefiles are located on a busy shared filesystem. In this situation, if cache is enabled for lmod, of course it will be faster. Cache mechanism in upcoming environment-modules 5.3 will solve this issue.

It you observe this performance gap on a simple local environment, I consider this is an issue. Thus I am interested to get some details (modules/lmod version number, their configuration, example modulefile, load times you get).

michaelkuhn:
> That sounds sensible to me. We can run the relevant methods and see if they contribute anything meaningful to the env changes.

OK, I will give that a try in the next few days.

> I also expected environment-modules load time for trivial (mostly static) `set-env ...` type module files to be dominated by IO, but that's not the case. In the small tests I did, lmod was much faster, so I'm surprised @michaelkuhn does _not_ see this

Here are my concrete numbers for reference (both with `autoload: direct`):
- Modules 5.2.0
```
$ export SPACK_SKIP_MODULES=1
$ . share/spack/setup-env.sh
$ . opt/spack/linux-fedora37-x86_64/gcc-12.2.1/environment-modules-5.2.0-wwl2hbc7uqw2v5jjyfa5ww2ig5nno4lo/init/bash
$ export MODULEPATH=$PWD/share/spack/modules/linux-fedora37-zen2
$ export -f module
$ hyperfine 'module load git-2.39.2-gcc-12.2.1-ur2ncvr'
Benchmark 1: module load git-2.39.2-gcc-12.2.1-ur2ncvr
  Time (mean ± σ):     174.5 ms ±   6.7 ms    [User: 164.2 ms, System: 9.6 ms]
  Range (min … max):   162.9 ms … 185.5 ms    16 runs
```
- Modules 5.2.0+XX-g56d76ded
```
$ export SPACK_SKIP_MODULES=1
$ . share/spack/setup-env.sh
$ . opt/spack/linux-fedora37-x86_64/gcc-12.2.1/environment-modules-main-xyf2rawfbtva46l2dvljxeum26jlj2aq/init/bash
$ export MODULEPATH=$PWD/share/spack/modules/linux-fedora37-zen2
$ module cachebuild
Creating $PWD/share/spack/modules/linux-fedora37-zen2/.modulecache
$ export -f module
$ hyperfine 'module load git-2.39.2-gcc-12.2.1-ur2ncvr'
Benchmark 1: module load git-2.39.2-gcc-12.2.1-ur2ncvr
  Time (mean ± σ):     181.0 ms ±   4.5 ms    [User: 171.9 ms, System: 8.3 ms]
  Range (min … max):   173.5 ms … 190.7 ms    16 runs
```
- Lmod 8.7.20
```
$ export SPACK_SKIP_MODULES=1
$ . share/spack/setup-env.sh
$ . opt/spack/linux-fedora37-x86_64/gcc-12.2.1/lmod-8.7.20-hkvxs7ws3rqh2f22lusndrezsit67ujo/lmod/lmod/init/bash
$ export MODULEPATH=$PWD/share/spack/lmod/linux-fedora37-x86_64/Core
$ export -f module
$ hyperfine 'module load git/2.39.2-ur2ncvr'
Benchmark 1: module load git/2.39.2-ur2ncvr
  Time (mean ± σ):     105.3 ms ±   2.1 ms    [User: 91.8 ms, System: 13.1 ms]
  Range (min … max):   101.4 ms … 109.4 ms    27 runs
```
- Lmod 8.7.20 (with modulefiles, just for fun)
```
$ export SPACK_SKIP_MODULES=1
$ . share/spack/setup-env.sh
$ . opt/spack/linux-fedora37-x86_64/gcc-12.2.1/lmod-8.7.20-hkvxs7ws3rqh2f22lusndrezsit67ujo/lmod/lmod/init/bash
$ export MODULEPATH=$PWD/share/spack/modules/linux-fedora37-zen2
$ export -f module
$ hyperfine 'module load git-2.39.2-gcc-12.2.1-ur2ncvr'
Benchmark 1: module load git-2.39.2-gcc-12.2.1-ur2ncvr
  Time (mean ± σ):     493.5 ms ±   6.5 ms    [User: 459.9 ms, System: 30.8 ms]
  Range (min … max):   484.5 ms … 504.4 ms    10 runs
```

michaelkuhn:
I've pushed a new version that autoloads:
1. direct and indirect run dependencies
2. dependencies that change the environment

I've kept the new name (`run`) since, while this should be correct, it still changes behavior.

It still provides a nice speedup:
- This Graphviz module autoloads direct dependencies:
```
Benchmark 1: module load graphviz
  Time (mean ± σ):     386.1 ms ±  14.0 ms    [User: 370.6 ms, System: 14.2 ms]
  Range (min … max):   367.3 ms … 407.8 ms    10 runs
```
- This Graphviz module autoloads all dependencies:
```
Benchmark 1: module load graphviz
  Time (mean ± σ):     448.1 ms ±  10.4 ms    [User: 433.9 ms, System: 12.8 ms]
  Range (min … max):   432.2 ms … 464.6 ms    10 runs
```
- This Graphviz module autoloads only run dependencies:
```
Benchmark 1: module load graphviz
  Time (mean ± σ):      99.6 ms ±   3.8 ms    [User: 90.5 ms, System: 8.8 ms]
  Range (min … max):    93.5 ms … 107.5 ms    27 runs
```

michaelkuhn:
@haampie ping (I've been running this on our cluster for a while and it seems to work fine)

michaelkuhn:
@haampie I have updated this to make use of `effective_deptypes` and `SetupContext`. I needed to make a small change to `SetupContext` to be able to reuse it for this but it is backwards compatible. Overall, this mode still provides a nice speedup:
- This Python module autoloads all dependencies:
```
Benchmark 1: module load python
  Time (mean ± σ):     221.0 ms ±   7.2 ms    [User: 209.9 ms, System: 10.1 ms]
  Range (min … max):   208.5 ms … 230.2 ms    13 runs
```
- This Python module autoloads direct dependencies:
```
Benchmark 1: module load python
  Time (mean ± σ):     225.1 ms ±   9.2 ms    [User: 212.9 ms, System: 11.2 ms]
  Range (min … max):   209.5 ms … 239.7 ms    14 runs
```
- This Python module autoloads only run dependencies:
```
Benchmark 1: module load python
  Time (mean ± σ):      85.7 ms ±   3.8 ms    [User: 77.0 ms, System: 8.2 ms]
  Range (min … max):    80.5 ms …  96.2 ms    34 runs
```

luigi-calori:
Sorry, I would have been interested into this functionality. PR closed means is not foreseen any merge or int is no more needed as problem alreay soveld in another way?
If I correctly got the issue, this PR would avoid loading modules that do not change runtime environment? 
That would reduce the number of modules loaded for recipes with a lot of dependencies which would be good for cluster production environment usage, not only for loading time but also for providing module user experience similar to manually generated modules

